PR #1421, opened 2022-11-27
Use DOMMatrix transforms to convert points between spaces.

flackr:
This supports arbitrarily complex transformations including perspective and also leans on the browser to do the conversion between coordinate spaces. Fixes #1420.

PR #1423, opened 2022-11-27
Support transitions on widget parent changes

### LINKED ISSUE #508: Deal Buttons that use CSS transitions
LawDawg and I have been experimenting with using the CSS transitions to deal cards, and I wanted to pass on some of the issues I ran into in case it is decided that something like this should be integrated into the javascript. 

Here is an example of how the effect appears. https://virtualtabletop.io/j1ej

These are the major steps I used to work around to an imperfect solution.  They illustrate some of the issues.

Step 1) change parent to the new holder and assign new x/y so that the card appears to remain in the same place (over the deck). This happens immediately, and is not 'seen' other than the card is now "over" the pile handle.

I use a timer and a 0.1 sec interval to separate steps 1 and 2.

2) change CSS on cards (e.g. "transition: ease-out 1.5s") and change their x/y to place them in the correct spot in the holder. Can also apply other effects such as rotation. (currently, spin is limited to increments of 180 degrees (360 for non-symmetrical cards) because onEnter and enterRoutine cannot be used to "correct" the rotation). Additionally, I set a parentChangeRoutine used in step 4.

This step must be separate from the first step because the CSS transition needs to have a starting and stopping point that isn't applied at the same time.

The CSS transition is not desirable outside of dealing the cards because it makes dragging the cards delayed (especially when the transition is as high as I set it).  The CSS transition CANNOT be removed by enterRoutine or onEnter, because the cards technically enter (in the properties) before the CSS transition moves them there.

3) Apply ownership and flip the card.  Ownership cannot be applied in step 2 because then only the owner would see the card travel to their hand (which somewhat defeats the point of the effect).  Typically these are applied on entry, but they cannot otherwise they will be applied before the CSS transition moves the cards to the hand (i.e. the cards would disappear to other players, and flip before entering the hand).

For this step, I use another timer.  The timer receives an array of objects that holds which holder they are heading towards (because I store the eventual owner of the card as a property in the holder), the card's ID, and the time it was delt.  Then 1.6 seconds (slightly over the time I set for the CSS transition) the timer changes ownership of the card, flips the card, and deletes the entry from the array.
 
4) Remove the CSS when the player attempts to move the card. A parentChangeRoutine applied in step 1 deletes the CSS so the player can drag the card freely as soon as they click on it.

The CSS could be removed in step 3, but that would remove a nice effect of the cards seeming to slide together smoothly.  Instead, you get a jarring effect as the cards move over to make room for the next card.  Step 3 could also wait until all the cards have settled, but they would remain faced away from the player until all the cards are dealt.  Step 4 could also be done through a 3rd timer, but that seemed less simple than a parentChangeRoutine.

The negative effect of this process is that the holder will not flip the cards when they enter.  So if the holder has an onExit that flips the cards, they will not automatically flip back when they are placed in the holder again.  I've also had to lock the buttons for a period (using the 2nd timer to prevent clicks long enough for the transitions to take place).

There are three things that would greatly simplify this process.

1) a fraction of a second interval between when a card is removed from its holder to be dealt and when other properties could be applied (e.g. new x/y and new rotation)
2) an interval between when a card is dealt, and when onEnter & enterRoutine are applied.
3) A setting that allows the deletion of transition effects from cards as soon as they are picked up.

Other settings for a deal button that would be helpful but could also be done through the JSON:
1) a property that determines how many cards will be delt
2) a property that stores an array of holders that the button will deal to
3) a property that determines if the number of cards will be dealt to one player, the next player, or the player that clicked the button

I completely understand if it is decided that this is not worth integrating into the Javascript (I'm not sure that it is).  However, I wanted to show how it would look, and give a basic understanding of what the obstacles are.

flackr:
This adds support for transitions from a previous position when a widget parent changes. I created a table with transitions on cards (while not dragged) where you can compare the behavior:

Compare that to the PR test here to see the difference.
Before: https://virtualtabletop.io/holder-transitions
With PR: https://test.virtualtabletop.io/PR-1423/pr-test

This is based on the PR in #1421 since it requires some of the newly added geometry helper methods in order to compute the relative transforms. I believe this fixes #508.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1423/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1423-pr-test`.

96LawDawg:
This looks amazing.  A question and a comment.

Question: how can we tweak the effect?  I see you put a "classes":"transitions" into the cardDefaults.  What css magic do I need to do to change the type of transition or change the length, for example?

Comment:  When dealing from the deck, sometimes the cards just appear without the transition.  Not sure if that is because more than one widget is moving at the same time or maybe for some other reason.  The missing transition animation is worse is Chrome than Edge.

flackr:
This works in 3d too, though the transition from a 2d transform to a 3d transform within perspective is a bit wonky:
https://test.virtualtabletop.io/PR-1423/perspective

flackr:
> Question: how can we tweak the effect? I see you put a "classes":"transitions" into the cardDefaults. What css magic do I need to do to change the type of transition or change the length, for example?

Check out the `css` widget, it's just a transition on the transform property defined there, specifically:
```
  "css": {
    "_, .transition": {
      "transition": "transform 150ms"
    },
    "_, .transition.dragging": {
      "transition": "none"
    }
  }
```
You can fully customize the length, easing curve, delays (which would be useful when dealing multiple cards to stagger the later cards) using all the standard css transition properties. Note that I specifically don't have a transition while dragging because that would make the card lag behind your drag.

> Comment: When dealing from the deck, sometimes the cards just appear without the transition. Not sure if that is because more than one widget is moving at the same time or maybe for some other reason. The missing transition animation is worse is Chrome than Edge.

Interesting, I hadn't seen this myself yet but I'll have to experiment more.

ArnoldSmith86:
I'll have a closer look when #1421 is merged and this one updated. Too lazy to compare them.

96LawDawg:
Did a little testing after the commits since my last test.  Still some skipping of the visible effect when using the Deal button in the demo room.

One other thing, you might want to consider is the layer.  Not sure if it is possible to easily fix this.  But when I use the Move Up/Down buttons in your demo room, the cards slide behind the buttons and basic widgets in the room.  I slowed the animation way down to watch.  This is because the cards are in a holder which is on layer -3 and the buttons are -1 and the basic widget -2.  If you were manually dragging the cards, they would pass over the buttons instead of behind them.

I could probably correct for that in the JSON by putting a temp layer on the cards while they are moving.  But do you think this is solvable in what you are coding?

RaphaelAlvez:
Is there any reason why the deal doesn't looks as good as the move up and move down?

Also I would prefer having just a little bit slower

96LawDawg:
> Also I would prefer having just a little bit slower

The timing will be set by each game designer in the css.  That isn't baked into the code.

client/js/serverstate.js
```diff
@@ -100,10 +100,6 @@ export function batchEnd() {
 
 function receiveDelta(delta) {
   // the order of widget changes is not necessarily correct and in order to avoid cyclic children, this first moves affected widgets to the top level
-  for(const widgetID in delta.s)
-    if(delta.s[widgetID] && delta.s[widgetID].parent !== undefined && widgets.has(widgetID))
-      $('#topSurface').appendChild(widgets.get(widgetID).domElement);
-
```
ArnoldSmith86, line 106: Removing this breaks for example FreeCell ( https://test.virtualtabletop.io/PR-1423/freecell ). The 8 columns of widgets are shuffled and each card is a child of the one under it.

So when resetting the game, it basically every time causes a child-parent-relationship to swap. When applying the delta without first moving the `domElement`s to an intermediate parent, it creates circular parent relationships.

ArnoldSmith86 (dismissed):

client/js/serverstate.js
flackr: Ah I see, to restate to make sure I'm looking at the same problem, the root issue is that given:
```html
<div id="a">
  <div id="b"></div>
</div>
```
```js
b.appendChild(a);
```
Will throw `DOMException: Failed to execute 'appendChild' on 'Node': The new child element contains the parent.`
Demo: https://jsbin.com/jejuvud/edit?html,css,js,output

I guess the most straightforward fix would be to give the widget the relative transform in `#topSurface` that it should have to keep it's original from position.

flackr: Restored this in the latest patch, but adding in preserving the transform relative to the root so that the from position is calculated correctly in `widget.js` when adding to the new parent.

flackr:
> Is there any reason why the deal doesn't looks as good as the move up and move down?

As far as I can tell it's just that the card moves too fast to be seen. If I make the transition duration longer it works perfectly. Keep in mind this PR is allowing developers to use standard CSS transitions, it does not define how they use those transitions.

There have been proposals in the CSSWG for distance-based transition timing which would solve this in the browser css space, but we could explore implementing some version of this in a followup.

robartsd:
> As far as I can tell it's just that the card moves too fast to be seen.

I believe this can happen with PCIO's transitions as well.

96LawDawg (approved):
Approved through test.

I created the original timing based transition routine for my poker room.  So I tried this new css based version out.  It works great and is so much simpler.

96LawDawg:
I think I would also like to recommend that this effect be added by default to new decks with a transition time of maybe 700ms.  That's what I did in Hearts in https://test.virtualtabletop.io/PR-1423/lawdawg.  It helps visually to see what is going on with the cards.  Be sure to try sorting and manually dragging and dropping into your hand as well.

96LawDawg (changes requested):
I have to revoke my approval.  FreeCell still doesn't work.

ArnoldSmith86:
And we all know that Jan gets angry when FreeCell doesn't work!

flackr:
@ArnoldSmith86 @96LawDawg Freecell works now. The issue was that we have to force the transform to be recomputed due to it being locally changed in `serverstate.js` but not updated by applyCSS due to not being part of the delta.

96LawDawg (approved):
Here.  Have another approval.

96LawDawg:
@flackr Here is a revised CSS tutorial that includes your transition demo and perspective demo.  The perspective demo didn't work because that branch wasn't in here, so I brought this branch up to date.  Now all of your recent changes have tutorials supporting them.  Thanks again for making VTT even better looking.


[CSS.zip](https://github.com/ArnoldSmith86/virtualtabletop/files/10357064/CSS.zip)

96LawDawg:
Well, that is just strange.  The last tutorial (Transform) works on production but does not work in this PR.  Could there be any conflict between them?  I don't understand what is going on.

flackr:
> Well, that is just strange. The last tutorial (Transform) works on production but does not work in this PR. Could there be any conflict between them? I don't understand what is going on.
> 
> See https://test.virtualtabletop.io/PR-1423/lawdawg

Looks about right now. I added the cardDefaults class and supporting CSS for the transitions and fixed the 3d transform on the table.

96LawDawg:
Well, I thought I copied the room exactly and just changed the text at the top.  Must have messed something up.  Save that (without the basic css timing widget--that makes it strange) as the last variant and put that in your PR.  Thanks for cleaning up the mess.

flackr:
> Well, I thought I copied the room exactly and just changed the text at the top. Must have messed something up. Save that (without the basic css timing widget--that makes it strange) as the last variant and put that in your PR. Thanks for cleaning up the mess.

No I think you got it right. It seems that the 3d css with the `!important` doesn't get applied until you trigger a change. I'll look into it but it seems like a bug (unrelated to this PR).

flackr:
@96LawDawg I fixed the tutorial loading in the latest update. I guess when applying the css inline it's overridden by the transform applied to set the widget position but if you set it with a stylesheet css (e.g. an object) then it is applied properly.

96LawDawg:
I think I figured out why I can't see your branches (because I'm not looking at your fork). But I'm such a n00b I can't figure out how to do that and I don't want to mess up what I have.  So here is one more tutorial change for you to please add to this PR.  I have added a comment to variant 1 of the Timer tutorial not to use that technique.  This was how I originally created card transition animations.  The capability demo is still useful as a timer tutorial, but the note advises people to use the technique you created here.  Thanks.

[Timer.zip](https://github.com/ArnoldSmith86/virtualtabletop/files/10361354/Timer.zip)

robartsd:
> I guess when applying the css inline it's overridden by the transform applied to set the widget position but if you set it with a stylesheet css (e.g. an object) then it is applied properly.

Yes, I believe setting element.style.transform will overwrite any inline transform styles. I think in production this only happens when a widget is moved (the initial transform is simply loaded when the inline styles are built). I think this PR may be setting it after loading css when initializing the widget

flackr:
> > I guess when applying the css inline it's overridden by the transform applied to set the widget position but if you set it with a stylesheet css (e.g. an object) then it is applied properly.
> 
> Yes, I believe setting element.style.transform will overwrite any inline transform styles. I think in production this only happens when a widget is moved (the initial transform is simply loaded when the inline styles are built). I think this PR may be setting it after loading css when initializing the widget

FWIW I don't think the issue was part of this PR, just part of @96LawDawg's tutorial based on my 3d css demo which was setting it using the inline css property. I don't think it's any different on tip of tree vs in this PR.

robartsd:
> FWIW I don't think the issue was part of this PR, just part of @96LawDawg's tutorial based on my 3d css demo which was setting it using the inline css property. I don't think it's any different on tip of tree vs in this PR as we already set style.transform here https://github.com/ArnoldSmith86/virtualtabletop/blob/main/client/js/widgets/widget.js#L146

That only gets called if a delta contains transform properties (parent, x, y) and does not contain any other css properties. If the delta contains any css properties that are not transform properties, the full css() method gets called instead (note the return a few lines above), so it doesn't get called until the widget gets moved.

```
  applyCSS(delta) {
    for(const property of this.classesProperties()) {
      if(delta[property] !== undefined) {
        this.domElement.className = this.classes();
        break;
      }
    }

    for(const property of this.cssProperties()) {
      if(delta[property] !== undefined) {
        this.domElement.style.cssText = mapAssetURLs(this.css());
        return;
      }
    }

    for(const property of this.cssTransformProperties()) {
      if(delta[property] !== undefined) {
        this.domElement.style.transform = this.cssTransform();
        return;
      }
    }
  }
```

flackr:
@robartsd sorry I explained myself poorly. I realize that it is technically a difference with this PR on initially loading a widget that it now sets the transform. My point was that given that the transform breaks if we ever update a transform related property later, it seems more consistent to have the initial state be consistent with what would happen if there's a later position update.

flackr:
That said if we want to preserve the current behavior I think it will work if we read the inline transform style before overriding it and then reapply that. Let me know what you think.

robartsd:
I personally don't care about changing this behavior.

Currently widget css is calculated and game engine transform is added to the end of the inline style string. If the widget css uses !important, it can override the engine transform with an inline style. If the game designer supported movement by using the widget transform properties in their css, then the properties would be in the array returned by this.cssProperties() and element.style.transform would never be set by the game engine. With this PR, this still works with widget stylesheets, but now fails with inline widget styles (I imagine that overwriting transform with !important in a stylesheet may cause this PR's support for transitions to fail).

I personally don't care if this breaks existing games that use !important to override the game engine transform.

96LawDawg:
> I personally don't care if this PR breaks existing games that use !important to override the game engine transform.

I do.  This breaks Crossed (the first dice no longer rolls).  It breaks variant 2 of the Dice Assets (so it probably breaks other things).  That's enough to require this to be reworked.  Our motto is not unlike that of doctors: above all else, do no harm -- we don't break existing games if we know about it.

96LawDawg (changes requested):
Need to fix issues that cause games to break.

robartsd:
A file updater that converts inline transform css to widget stylesheet transform css should fix this for most cases. The remaining issue to this approach would be inline css set by a routine.

flackr:
I didn't realize games were relying on this. It seems fragile, but I can see three possible approaches,
1. As @robartsd suggested, we could use a fileupdater to make the same modification that I did to the perspective tutorial.
2. Try to remember the previous `style.transform`. This is rather ugly as we'd need to detect whether it was changed by applyCSS and track it across files as it's updated in two locations.
3. Call applyCSS with a delta of `this.state` filtered to cssProperties(), and adding cssTransformProperties() to ensure that we reset the transform if nothing else does.

I would prefer not to do 2, but 1 or 3 both seem okay to me. 3 seems slightly more expensive but would preserve the current behavior, so I'd lean towards 1 - a file update for existing games.

flackr:
But ... parsing css for the file updater is non-trivial to get right, so I figured out a relatively simple way to track what the inline transform should be after the transition - just saving it after related changes in applyCSS.

flackr:
@96LawDawg I switched the tutorial back to what it was originally and verified this is all working.

96LawDawg (approved):
Everything works again.  Approved through test.

robartsd:
> But ... parsing css for the file updater is non-trivial to get right, so I figured out a relatively simple way to track what the inline transform should be after the transition (i.e. option 2) - just saving it after related changes in applyCSS.

This looks like a great solution to me.

robartsd (approved):
Code looks good.

PR #1424, opened 2022-11-28
Allow array of holders in COUNT, array of canvas in CANVAS.

### LINKED ISSUE #1197: `"holder":["foobar"]` crashes client
I was trying to `COUNT` a specific holder.  I had been working with collections so much, instead of just putting the name of the holder in the holder property, I tried to use the brackets [ ] like you do to make an anonymous collection.  That froze the room.  I always say you can't fix stupid (that would be me), but maybe the code can be fixed to avoid that or at least treat it as an anonymous collection so the room doesn't freeze.

rogerl50:
Fixes #1197. Now:
`CALL` checks to make sure that `widget` and `routine` are either widget ids or single-element arrays.
`CANVAS` converts the `canvas` property to the `collection` property and proceeds, so supports multiple canvases.
`COUNT` allows multiple holders for the `holder` parameter and adds up the number of widgets in each one.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1424/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1424-pr-test`.

96LawDawg (approved):
Approved through test.

RaphaelAlvez (approved):
Code review. Look good and similar to other problem prompts. Please wat for a test review

96LawDawg:
Should `CANVAS` also check to make sure the widget is actually a canvas?  It doesn't break the room as is, but it would let users know what went wrong.  A few of the arrary/math functions do something similar, I think (letting you know why things didn't work even if it doesn't freeze the room).

96LawDawg:
Since collections are already supported for `CANVAS`, wouldn't it make sense to support anonymous collections in the canvas parameter as well to be consistent with other functions that do the same thing?

client/js/widgets/widget.js
```diff
@@ -1684,11 +1694,12 @@ export class Widget extends StateManaged {
   }
 
   isValidID(id, problems) {
-    if(Array.isArray(id))
+    if(Array.isArray(id))
```
ArnoldSmith86, line 1697: Perfect change! :rofl:

client/js/widgets/widget.js
```diff
@@ -938,9 +943,14 @@ export class Widget extends StateManaged {
         let collection;
         let theItem;
         if(a.holder !== undefined) {
-          if(this.isValidID(a.holder,problems)) {
-            variables[a.variable] = widgets.get(a.holder).children().length;
+          const holders = asArray(a.holder).filter(w=> widgets.has(w) && widgets.get(w).get('type')=='holder');
+          if(holders.length != asArray(a.holder).length)
+            problems.push("Holder parameter must consist only of holders.");
```
ArnoldSmith86, line 948: This changes the behavior of `COUNT`, right? It didn't enforce holders before so you could just count children of any widget.

ArnoldSmith86 (changes requested):
In general this looks fine although it's really hard to tell if the changes actually don't change the behavior of the functions.

See my comment about `COUNT` where it most definitely changes something.

client/js/widgets/widget.js
rogerl50: I hate it when that happens. Fixed.

client/js/widgets/widget.js
rogerl50: I fixed this (and simplified the code). I'm pretty confident that nothing else changes the behavior of any functions.

rogerl50:
> In general this looks fine although it's really hard to tell if the changes actually don't change the behavior of the functions.
> 
> See my comment about `COUNT` where it most definitely changes something.

See my response to that comment. That did indeed, now fixed. I'm pretty confident that no other behavior changes. Which probably means that I'm wrong, but I've tested a bunch of cases.

rogerl50:
Any further comments on this PR? I believe I've addressed @ArnoldSmith86's comment.

RaphaelAlvez:
You can go ahead and add collection of holder in MOVE  and MOVEXY in my opinion

ArnoldSmith86:
> You should also include collections. Make it finally consistent across functions. Every function should accept an ID, array/pseudo collection or a collection as input.
> 
> You can go ahead and add collection of **holders** in MOVE and MOVEXY in my opinion

That has nothing to do with this PR IMHO.

RaphaelAlvez:
> > You should also include collections. Make it finally consistent across functions. Every function should accept an ID, array/pseudo collection or a collection as input.
> > 
> > You can go ahead and add collection of **holders** in MOVE and MOVEXY in my opinion
> 
> That has nothing to do with this PR IMHO.

Yeah i me at more in a "it would be better" way instead of "you need"

I'll make a test review later. The code looks decent

96LawDawg:
I did some further testing.

For CANVAS, if you try to do something to a non-canvas widget, the client freezes.  It didn't do this before in this PR and it doesn't do it on production. For example:

```
    {
      "func": "CANVAS",
      "mode": "reset",
      "canvas": "holder1"
    }
```

Also for CANVAS, something changed with the defaults.  On production, this does not freeze the client, but it does in this PR:

```
    {
      "func": "CANVAS"
    }
```

Everything else tests okay as far as I can tell.

rogerl50:
**Ready for re-review.** All known issues fixed.

96LawDawg:
This still does not work:

  ```
    {
      "func": "CANVAS"
    }
```

And that is a change from production.  The reason is because whatever is happening in the new code does not accept the default of "null."  With all of these improvements, I think we should change the parameter "canvas" to "collection,"  That would be consistent with other functions.  And then we change the default to "DEFAULT." But that is outside the scope of this PR because that would also require changing the composite widget and doing a file updater.

But we at least need to address the problem of `"canvas": null.`

RaphaelAlvez (approved):
updated code review. changes look good

RaphaelAlvez (approved):
tested based on what was in the test room and added a couple of tests. looks good

ArnoldSmith86 (approved):
Code looks fine. I'm least confident about the changes to `CANVAS` though.

Did somebody test `Imagination`, `You Can't Draw` and the canvas tutorial?

96LawDawg:
> Code looks fine. I'm least confident about the changes to `CANVAS` though.
> 
> Did somebody test `Imagination`, `You Can't Draw` and the canvas tutorial?

I tested You Can't Draw.

PR #1432, opened 2022-12-03
Add 'range' option to FOREACH

### LINKED ISSUE #1431: FOREACH should support a count of n loops
Right now, we can only use a loopRoutine in a FOREACH that consists of an array or a collection.  It would be nice to also support a counter/increment function (from 0 to n or 1 to n).  That would save having to use create a made up array just to only use an i = 1, next i, kind of construct with an IF condition.

rogerl50:
Fixes #1431.

This adds `range` in addition to the existing `in` and `collection` parameters to `FOREACH`. `in` takes precedence over `range`, which takes precedence over `collection` if more than one is specified.

The value to `range` is a range of the form `[start, end, step]`. If `step` is omitted, it is assumed to be 1 if `start` < `end` and -1 if `start` > `end`. The `loopRoutine` is called once for each element of the resulting range, passing the range element as `value`.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -1027,19 +1027,29 @@ export class Widget extends StateManaged {
           if(jeRoutineLogging)
             jeLoggingRoutineOperationSummary( `elements in '${JSON.stringify(a.in)}'`);
         } else if(a.range) {
-          let range = a.range;
-          if(!Array.isArray(range))
-            range = [1,range];
-          if(range.length < 2 || range[1]<0 ||(range[1] && range[1]<0)) {
-            problems.push(`Range ${JSON.stringify(range)} should be either an array with at least two elements, or a positive number, [1,1] used.`);
-            range = [1,1]
+          let range = asArray(a.range);
+          if(range.length == 1)
+            range.unshift(1);
+          if(range.length == 2)
+            range.push(1);
+          let start = parseFloat(range[0]);
+          let end = parseFloat(range[1]);
+          let step = parseFloat(range[2]);
+          if(isNaN(start)) {
+            problems.push(`Invalid start of range ${JSON.stringify(range[0])}, 1 used`);
+            start = 1;
           }
-          let start = range[0];
-          let end = range[1];
-          let step = range[2] || (start < end ? 1 : -1);
-          if( isNaN(parseFloat(start)) || isNaN(parseFloat(end)) || isNaN(parseFloat(step)) || (start < end && step < 0) || (start > end && step > 0) ) {
-            problems.push(`Invalid range ${JSON.stringify(range)}, [1,1,1] used`);
-            start = end = step = 1;
+          if(isNaN(end)) {
+            problems.push(`Invalid end of range ${JSON.stringify(range[1])}, 1 used`);
+            end = 1;
+          }
+          if(isNaN(step) || step == 0) {
+            problems.push(`Invalid step value ${JSON.stringify(range[2])}, 1 used`);
+            step = 1;
+          }
+          if(start>end && step>0 || start<end && step<0) {
+            step = -step;
+            problems.push(`Step ${JSON.stringify(range[2])} changed to ${step}`)
           }
           for (let index=start; (step > 0) ? index <= end : index >= end; index += step)
             await callWithAdditionalValues({ value: index });
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -1025,12 +1025,29 @@ export class Widget extends StateManaged {
           for(const key in a.in)
             await callWithAdditionalValues({ key, value: a.in[key] }, {});
           if(jeRoutineLogging)
-            jeLoggingRoutineOperationSummary( `element in '${JSON.stringify(a.in)}'`);
+            jeLoggingRoutineOperationSummary( `elements in '${JSON.stringify(a.in)}'`);
+        } else if(a.range) {
+          const range = a.range;
+          if(range.length < 2) {
+            problems.push(`Range ${JSON.stringify(range)} should be an array with at least two elements.`);
+            range = [1,1]
+          }
+          let start = range[0];
+          let end = range[1];
+          let step = range[2] || (start < end ? 1 : -1);
+          if((start < end && step < 0) || (start > end && step > 0)) {
+            problems.push(`Invalid range ${JSON.stringify(range)}, [1,1,1] used`);
+            start = end = step = 1;
+          }
+          for (let index=start; (step > 0) ? index <= end : index >= end; index += step)
+            await callWithAdditionalValues({ value: index });
+          if(jeRoutineLogging)
+            jeLoggingRoutineOperationSummary( `values in range '${JSON.stringify(a.range)}'`);
         } else if(collection = getCollection(a.collection)) {
           for(const widget of collections[collection])
             await callWithAdditionalValues({ widgetID: widget.get('id') }, { DEFAULT: [ widget ] });
           if(jeRoutineLogging)
-            jeLoggingRoutineOperationSummary( `widget in '${a.collection}'`);
+            jeLoggingRoutineOperationSummary( `widgets in '${a.collection}'`);
         }
       }
 
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -1027,32 +1027,43 @@ export class Widget extends StateManaged {
           if(jeRoutineLogging)
             jeLoggingRoutineOperationSummary( `elements in '${JSON.stringify(a.in)}'`);
         } else if(a.range) {
+          const rangeProblems = []; // To avoid repeated error messages in each loop iteration
           let range = asArray(a.range);
-          if(range.length == 1)
-            range.unshift(1);
-          if(range.length == 2)
-            range.push(1);
+          
+          if(range.length == 0) {
+            rangeProblems.push(`Empty range given, [1] used.`);
+            range = [1]
+          }
           let start = parseFloat(range[0]);
-          let end = parseFloat(range[1]);
-          let step = parseFloat(range[2]);
           if(isNaN(start)) {
-            problems.push(`Invalid start of range ${JSON.stringify(range[0])}, 1 used`);
+            rangeProblems.push(`Invalid start of range ${JSON.stringify(range[0])}, 1 used`);
             start = 1;
           }
+          
+          if(range.length == 1)
+            range.unshift(1);
+          let end = parseFloat(range[1]);
           if(isNaN(end)) {
-            problems.push(`Invalid end of range ${JSON.stringify(range[1])}, 1 used`);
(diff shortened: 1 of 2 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1432/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1432-pr-test`.

96LawDawg:
Proposed changed to `FOREACH` wiki:

------------------------

This function allows for iteration over a collection, object, array, string, or range of values. The _loopRoutine_ parameter specifies the actions to be taken.

If a parameter _in_ is present as `"in": "<element>"`, then `<element>` is assumed to be an object, an array, or a string. In each case, the _loopRoutine_ is provided with a _key_ and a _value_ on each call. For an object, each _key_ and _value_ is taken from the keys and values in the object. For an array  or a string, the _key_ is the 0-origin position of the element, and the _value_ is the element itself.

If a parameter _range_ is present, it takes the format `[start, end, step]`.  The `loopRoutine` is called once for each element of the resulting range, passing the range element as _value_.  Each of `start`, `end`, and `step` can be numbers or variables. If `step` is omitted, it is assumed to be 1 if `start` < `end` and -1 if `start` > `end`.

If a parameter _collection_ is present, the _loopRoutine_ will be called once for each widget in the collection. The routine will receive a variable, _widgetID_, and a _DEFAULT_ collection that contains the one widget of the current iteration.

If _in_, _collection_, and _range_ parameters are provided, _in_ takes precedence over _range_, which takes precedence over _collection_. If none of these is supplied, the function will operate on the _DEFAULT_ collection.

Note that the supplied values (_key_, _value_, _widgetID_, and the _DEFAULT_ collection) in the _loopRoutine_ are treated as local to that routine, and changes to them will not affect identically-named objects in the parent. However, any changes to other variables that are defined in the parent will survive the call and remain permanent.

Here is a simple example:
```
  "clickRoutine": [
    "var start = 0",
    "var end = 6",
    "var increment = 2",
    {
      "func": "FOREACH",
      "range": [
        "${start}",
        "${end}",
        "${increment}"
      ],
      "loopRoutine": [
        "var myResult = 'hello' + ${value}"
      ]
    }
  ]
  ```
The `myResult` variable will be set to "hello0", "hello2", "hello4", and "hello6" as the `loopRoutine' steps through each iteration.
  
As another example, suppose `<array>` is an array of strings.
  ```
  "clickRoutine": [
    "var string = ''",
    {
      "func": "FOREACH",
      "in": "<array>",
      "loopRoutine": [
        "var string = ${string} + ${value}"
      ]
    }
  ]
  ```
  concatenates all the string in `<array>` and returns the result in the variable `string`.
  
  Finally, the routine
  ```
  "clickRoutine": [
    {
      "func": "SELECT",
      "property": "myProperty",
      "collection": "myCollection"
    },
    {
      "func": "FOREACH",
      "collection": "myCollection",
      "loopRoutine": [
        {
          "func": "CALL",
          "routine": "myRoutine",
          "arguments":
            {
              "widget": "${widgetID}"
            }
        }
      ]
    }
  ]
  ```
invokes the routine named `myRoutine` in the current widget, passing it in turn each widget in the collection `myCollection`.

**Review by 96LawDawg — APPROVED** [review]
Approved through test.  I tested whole numbers, negative numbers, decimal numbers, and strings.  I also tested using variables in the `range [ ]` settings.  Everything works as expected.  The tutorial for FOREACH has been updated and is in the PR test room.

**Comment by rogerl50** [comment]
There is an amusing application in https://test.virtualtabletop.io/PR-1432/Roger.

**Comment by rogerl50** [comment]
@96LawDawg Note that the latest commit expands the possibilities for `range` to include `"range": 5`.

**Comment by 96LawDawg** [comment]
Tested new option and updated draft wiki.

File: client/js/widgets/widget.js [hunk 1]
```diff
@@ -1025,12 +1025,31 @@ export class Widget extends StateManaged {
           for(const key in a.in)
             await callWithAdditionalValues({ key, value: a.in[key] }, {});
           if(jeRoutineLogging)
-            jeLoggingRoutineOperationSummary( `element in '${JSON.stringify(a.in)}'`);
+            jeLoggingRoutineOperationSummary( `elements in '${JSON.stringify(a.in)}'`);
+        } else if(a.range) {
+          let range = a.range;
+          if(!Array.isArray(range))
+            range = [1,range];
+          if(range.length < 2 || range[1]<0 ||(range[1] && range[1]<0)) {
```
robartsd, line 1033: I'm not following the logic. Clearly range.length needs to be >= 2, but the tests for range[1] don't seem quite right.

robartsd (changes requested):
range: [1, -1, -1] does not work as expected.

robartsd:
Should we allow a single element array with the provided value being used as end? (`"range":5` would be treated the same as `"range":[5]`)

96LawDawg:
Updated wiki to show that range: 5 and range: [5] both work.

96LawDawg:
> > range: [1, -1, -1] does not work as expected.
> 
> What fails? (See https://test.virtualtabletop.io/PR-1432/ud39).

Nothing now.  He changed the code.  I don't know what would have happened before his changes because I didn't test that case.

robartsd:
> > > range: [1, -1, -1] does not work as expected.
> > 
> > 
> > What fails? (See https://test.virtualtabletop.io/PR-1432/ud39).
> 
> Nothing now. He changed the code. I don't know what would have happened before his changes because I didn't test that case.

Before, it would detect that end is < 0 and use range [1, 1] instead. Now negative numbers can be given as a sole argument (`"range":-3` is equivalent to `"range":[1,-3,-1]`).

client/js/widgets/widget.js
```diff
@@ -1025,12 +1025,41 @@ export class Widget extends StateManaged {
...
             await callWithAdditionalValues({ key, value: a.in[key] }, {});
           if(jeRoutineLogging)
-            jeLoggingRoutineOperationSummary( `element in '${JSON.stringify(a.in)}'`);
+            jeLoggingRoutineOperationSummary( `elements in '${JSON.stringify(a.in)}'`);
+        } else if(a.range) {
+          let range = asArray(a.range);
+          if(range.length == 1)
+            range.unshift(1);
+          if(range.length == 2)
+            range.push(1);
```
rogerl50, line 1034: Perhaps push -1 if end < start?

robartsd: Yeah, no reason to set a default value that later generates a warning that the value got changed. I was also thinking there is some room for improvement in the step validation.

rogerl50: Have another look.

client/js/widgets/widget.js
```diff
@@ -1027,32 +1027,43 @@ export class Widget extends StateManaged {
...
+          
           if(start>end && step>0 || start<end && step<0) {
             step = -step;
-            problems.push(`Step ${JSON.stringify(range[2])} changed to ${step}`)
+            rangeProblems.push(`Step ${-step} changed to ${step}`)
           }
+          
           for (let index=start; (step > 0) ? index <= end : index >= end; index += step)
             await callWithAdditionalValues({ value: index });
+          problems = rangeProblems;
```
ArnoldSmith86, line 1066: Are you sure that you don't dismiss problems here that were added before this block of code?

rogerl50: Hmm. I thought problems was re-initialized for each function in a routine, but I see that is not the case. So I could either 1) put things back the way they were and let the error propagate into the loopRoutine executions, or 2) save `problems` before the calls and restore it afterwards. This may be the only place in the module where this is a potential issue. What do you think?

ArnoldSmith86: It should be re-initialized for each function in a routine. But for example if you use variables inside your `range` property, those could add problems before your block of code. An example would be `The variable ${invalid} is not an object, so indexing it doesn't work.` if you use `range: [ "${invalid.key}" ]`. (I didn't test this.)

ArnoldSmith86: I'm not sure why your construct (`rangeProblems`) is needed in the first place since you don't add problems in the actual `for` loop. But I didn't look at the code in detail yet and didn't test anything either.

rogerl50: The issue is that any error generated by the bounds will be repeated in the problems section of each loop iteration. But it seems too complicated to try to fix properly (the loop iterations could produce their own errors), so I just restored things to the way they were. An extra error message later in the tree than the initial one shouldn't be a big deal.

ArnoldSmith86 (approved):
Code looks fine to me. I didn't test anything. Looking forward to using this.

96LawDawg (changes requested):
There is still the problem of red (problem) text, without the explanation getting pushed anywhere (that I can find) in the debugger.  I couldn't follow all of the discussion and maybe that's what you were talking about?

For example, 
```
"range": [
        2,
        -5,
        1
      ],
```
Should give an error because the step should be -1.  The text shows up red, but doesn't explain that the step was changed to be -1.

client/js/widgets/widget.js
```diff
@@ -1027,19 +1027,29 @@ export class Widget extends StateManaged {
...
+            end = 1;
+          }
+          if(isNaN(step) || step == 0) {
+            problems.push(`Invalid step value ${JSON.stringify(range[2])}, 1 used`);
+            step = 1;
+          }
+          if(start>end && step>0 || start<end && step<0) {
+            step = -step;
+            problems.push(`Step ${JSON.stringify(range[2])} changed to ${step}`)
           }
```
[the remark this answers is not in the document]
rogerl50: Take a look at https://test.virtualtabletop.io/PR-1432/foo. Under "problems", just below the FOREACH, there is an error message. Did you mean something else?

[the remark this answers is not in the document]
96LawDawg: Yes.  Just like that.  I just overlooked it in my test room.

96LawDawg (approved):
I have retested it after all the changes.  Everything works, including throwing stupid values in.  The only problems that don't get reported are arrays.  For example, if your start value is [5,7,false], it will just use the 0th index value.  I don't have an issue with that so I'm approving it again.  Just wanted to point it out.

client/js/widgets/widget.js
```diff
@@ -1025,12 +1025,50 @@ export class Widget extends StateManaged {
...
-            jeLoggingRoutineOperationSummary( `element in '${JSON.stringify(a.in)}'`);
+            jeLoggingRoutineOperationSummary( `elements in '${JSON.stringify(a.in)}'`);
+        } else if(a.range) {
+          let range = asArray(a.range);
+
+          if(range.length == 0) {
+            problems.push(`Empty range given, [1] used.`);
+            range = [1]
+          }
+          let start = parseFloat(range[0]);
```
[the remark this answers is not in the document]
rogerl50: Thanks. Should be fixed.

robartsd (approved):

PR #1223, opened 2022-05-19
game library overhaul

ArnoldSmith86:
https://test.virtualtabletop.io/PR-1223/pr-test

This is basically #791 with only my commits and rebased to the current version.

I _love_ what @mousewax is doing UI wise but both of us have too little time atm to get the backend _and_ the UI done anytime soon. So I want to focus on getting my (ugly) version of this done soon and mousewax can work on making it pretty and usable afterwards without having to worry about all the backend conflicts that accrue over time.

~~Oh, and this still has the old library that is now missing 40 games or so. I'm not sure yet if I'll make an intermediate update or just do it before actually merging.~~

Things that need to happen before merging this ~~(anything that doesn't _have_ to happen will be done in subsequent PRs)~~:

- [x] The game descriptions from the old public library table are not visible anywhere.
- [x] Tutorials and assets need to be accessible.
- [x] There's a state called "Decks" in tutorials _and_ assets. They currently cause a conflict and only one of them is visible. Same for "Dice".
- [x] The order of variants in tutorials and assets are random.
- [x] Attribution of the public library games/tutorials/assets is not reachable.
- [x] ~~The "welcome" text is gone. I just removed it for now because most of it probably needs to be updated now but it should be somewhere.~~ I'll do this stuff after merging because I'll record some videos and we'll probably fine-tune some visuals after merging this PR.
- [x] The filtering toolbar needs to be a bit prettier so you can see what the input fields are.
- [x] Public library games in existing rooms need to be migrated so they still work.
- [x] The edit screen has to be able to work with the new links to public library variants that the migration creates.
- [x] Starring a game should be smoother than it is now (it jumps to the other category _after_ a roundtrip to the server).
- [x] Starring a game resets the filters.
- [x] A cross button to clear the search textbox.
- [x] The star button should differ depending on if the game is starred or not.
- [x] Testcafé tests need to be fixed (mainly loading the games).
- [x] I have to look at the code changes for stuff that needs to be updated for code changes in the meantime. Shouldn't be much though because I tested most features and they work.
- [x] A better way to edit public library games for easier PR creation.
- [x] PL games get saved to every room JSON file. They should be removed before saving.
- [x] The library has to be updated to its current contents.
- [x] `urlProperties.load` needs to be fixed.
- [x] `REMOVE BEFORE MERGE` needs to be removed. :scream:
- [x] At the end I want to try to make it look like a mix of the initial mock-up (see #791 description) and mousewax's prototype implementation.
[...]

PR #1435, opened 2022-12-07
show and tell: added game categories to library

ArnoldSmith86:
This is one of the remaining features I want to add to #1223.

You define in which categories a game should be displayed:

image

And then those categories are available in the filter dropdown:

image

The UI is not final but it seems to be functional. So this PR is a demo so we can discuss the feature.

For the category transformation, I used this function:

```Javascript
  function getCategories(inputString) {
    const categories = new Set();

    const lines = inputString.split('\n');
    lines.forEach(line => {
      const category = line.split(' > ');

      let current = [];
      category.forEach(cat => {
        current.push(cat);
        const path = current.join(' > ');
        categories.add(path);
      });
    });

    return [...categories];
  }
```

It was written entirely by ChatGPT and you can see how unbelievably simple it was to tell the AI what to create: https://is.gd/E7i5Vt


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1435/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1435-pr-test`.

RaphaelAlvez:
Maybe the option to add an icon there would be nice.

What are the >>>>?

ArnoldSmith86:
Indentation for sub categories.

If you use `Games > Card Games > Coop Games` in a game, it will also show the parent categories:

```
Games
Games > Card Games
Games > Card Games > Coop Games
```

But because that is way too long, it shortens it to:

```
Games
> Card Games
> > Coop Games
```

I'm very open to suggestions.

RaphaelAlvez:
also maybe "official art" should be a tag

96LawDawg:
> Why not just give a tag property to games?
> 
> ["Coop","card game","spiel des jahre"]

That seems like a good idea.  Simple.  No need for the hierarchy and >> indentations.

ArnoldSmith86:
Hmm, I don't disagree, but that requires everything to be completely different. I guess I don't add anything in #1223 then..

ArnoldSmith86:
In the meantime we might want to use this PR to gather every tag we would want to use? That would probably help to come up with a UI.

My general idea for this was to be able to better filter the games. Card vs board vs dice games; only show Spiel des Jahres; we're also missing age restrictions. Complexity, mechanisms, ...

robartsd:
I agree that tags would be more flexible as a user. For instance, I might want to filter to all coop games (not just card coop OR board coop).

PR #693, opened 2021-07-23
numericSort sorts negative below positive numbers

bjalder26:
This is to address the issue in #692.

Production server:
https://virtualtabletop.io/Issues
Game: #692 numericSort ignores negatives

This PR:
http://212.47.248.129:3693/pr-test

- _Should null values be counted as highest, lowest, or zero values?_  With this PR they are now treated as zero values.
- _How should letters be sorted, or a mix of letters and numbers?_ With this PR they are not sorted 'below' negative numbers, but not sorted among each other.
- _How should empty strings, and strings with white space be sorted?_  With this PR they are 'below' negative numbers (with strings), but strings are not sorted among each other.

I was thinking that the way numericSort should would (with non-number input) would be like this:

1. anything that has a non-number character in it, empty strings (unsorted), and strings of spaces
2. negative numbers (LOW = more negative to HIGH = less negative) - including string-type numbers
3. zero, null
4. Positive numbers (low to high) - including string-type numbers

My assumption is that the way numericSort would be expected to work is by sorting numbers (whether strings or literals) from low to high.  I also assume that null would be treated as a number (zero), whereas an empty string would not.  I also assume that numericSort should not sort non-number values, but that they need to be separated from the number values in order to perform a numericSort.

This would not give maximum flexibility with how thing are sorted, but I think more complicated sorts (e.g. null, then alphabetical order for words, then numbers low to high) can be handled in the JSON with IF/Then statements and arrays.

**I think emulating the current behavior would make a lot of JSON complicated with little benefit, and that this PR should wait for legacy modes**

Game-breaking potential:

- Game depends on null values or negative numbers being sorted in a different way.

PR #1436, opened 2022-12-09
Fix numericSort

### LINKED ISSUE #692: numericSort ignores negative numbers
numericSort doesn't sort negative numbers 'below' positive numbers.  It essentially ignores the negative.
Eg.    [0, -1, -2, "-3", -5,  5, -10, "10", -10, 25]

Also null values cause an error 'Exception: TypeError: Cannot read property 'toString' of null'.

It could be sorted like you would expect on a number line by using a compare function.
https://www.w3schools.com/js/js_array_sort.asp#:~:text=Example%3A%20The%20compare%20function%20compares%20all%20the%20values,sort%2040%20as%20a%20value%20lower%20than%20100.

An example can be seen here: https://virtualtabletop.io/Issues
Game: #692 numericSort ignores negatives

96LawDawg:
Fixes #692.  `numericSort` changed so that it works with numbers instead of strings.  This allows it to sort negative numbers correctly.  I used ChatGPT to help with this because I am not a coder.  But the solution is so obvious (and straight out of the JS manual) that there's no need to give the AI credit.  But screenshot is included below.

I think this will require a file updater.  Somebody (probably me) used `numericSort` at some point without realizing the limitation and some result may rely on it.  But I have no idea how to do a file updater.  If somebody can make that happen, this will fix a bug.

Screenshot 2022-12-08 202654


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1436/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1436-pr-test`.

96LawDawg:
Modified version of Array tutorial loaded into PR room that shows this works.

ArnoldSmith86 (dismissed):
This makes it a completely different function though.

Yours is probably very useful (unless we have an equivalent function somewhere) but it shouldn't _replace_ `numericSort`.

The only problem with `numericSort` is probably the name and description. It sorts strings but instead of sorting `A10` before `A2`, it tries to be smart about numbers in strings: https://virtualtabletop.io/1iii

96LawDawg:
That makes sense.  We do not have a function that does what I changed this to do.  I guess #693 was the attempt to change this and make it do all things.  Seems like a better suggestion is to leave the current numericSort alone and create a new function.  But numericSort will still suck at negative numbers.

ArnoldSmith86:
Maybe we rename the existing one to `numericStringSort`?

96LawDawg:
> Maybe we rename the existing one to `numericStringSort`?

That works for me.  Sort of.  I think I can make the new function.  Just add a new section in COMPUTE?  But I do not have the knowledge to change existing instances of `numericSort` using a file updater.

96LawDawg:
The latest commit returns the lost functionality and names it `numericStringSort`.  It creates a new function called `numericSort` which does what it is supposed to do (sort numbers).  The hashes aren't right yet of course.  But the bigger problem is the file updater.  I tried multiple sessions with ChatGPT.  This is the best we could come up with.  I'm sure it is a disaster (I couldn't even get it to work in testing), but it gets the conversation started.

The new functions do seem to work fine.

96LawDawg:
Well, the file updater worked better than I thought.  It didn't do anything on my local machine.  But at least it changed "numericSort" to "numericStringSort."  Strangely, it only updated things to v8??

ArnoldSmith86:
Most of your changes to `fileupdater.js` weren't necessary. It's already structured in such a way that your `v9NumericStringSort` function will be called for nested routines inside `IF` etc.

Let's see if TestCafé finishes now.

96LawDawg:
Thanks for the help with the hash and file updater.  I have updated the Potential Version Upgrades section at https://github.com/ArnoldSmith86/virtualtabletop/wiki/Version-Upgrades with the information.

I believe this is ready for final code review and approval.  I have tested it and updated the Array tutorial (Modifying - Advanced variant).  But someone else should also take a look.

ArnoldSmith86 (approved):
The code looks fine. Somebody should test the affected public library games and tutorials. I did a quick `grep` to see which of them are using `numericSort`:

```Bash
$ grep numericSort library/*/*/*json
library/games/Alpaca/0.json:      "var countArray = ${countArray} numericSort",
library/games/Meta/0.json:      "var typeArray = numericSort ${typeArray}",
library/games/Tutti Frutti/0.json:      "var yogurtZ = ${yogurtZ} numericSort",
library/games/Tutti Frutti/0.json:      "var sortedOrangeArray = ${orangeArray} numericSort",
library/games/Yoink/0.json:        "var tempArray = numericSort ${tempArray}",
library/tutorials/Arrays/3.json:      "var wordArray = ${wordArray} numericSort //Sorts the array numerically",
library/tutorials/Arrays/3.json:      "var numArray = ${numArray} numericSort",
library/tutorials/Arrays/3.json:      "var mixedArray = ${mixedArray} numericSort",
library/tutorials/Arrays/3.json:    "text": "numericSort"
library/tutorials/Arrays/3.json:      "var wordArray = ${wordArray} reverse //Reverses the order of the elements in the array; to reverse numbers, first do a numericSort then a reverse",
library/tutorials/Functions - SORT/4.json:      "var handWorth = ${handArray} numericSort",
library/tutorials/Routines/3.json:      "var finalBoard = ${finalBoard} numericSort // Sort the array from smallest to largest.  Also numericSort does not work for negative or null numbers so in this tutorial, the minus buttons are set up to keep the numbers from being negative.",
library/tutorials/changeRoutine/1.json:      "var finalBoard = ${finalBoard} numericSort // Sort the array from smallest to largest.  Also numericSort does not work for negative or null numbers so in this tutorial, the minus buttons are set up to keep the numbers from being negative.",
```

96LawDawg:
> The code looks fine. Somebody should test the affected public library games and tutorials. I did a quick `grep` to see which of them are using `numericSort`:
> 

I guess I'm the only one who uses this.  I think all of those are mine.  I just completed testing everything on that list except for Tutti Frutti.  Some of the tutorials were used in development testing.  Everything using the old `numericSort` and now `numericStringSort` works as expected.

PR #1437, opened 2022-12-09
Fix two minor issues with JSON editor

rogerl50:
1. Removes `skip` as a default option for every function - it has long since been deprecated.
2. Red color indicating errors in the debug info now works better: a red arrow with not-red text following indicates that the current statement is OK but that one of its descendants has an error. A red arrow with red text following indicates that this statement itself is in error; the `Problems` section under the designator will explain the error.

### DIFF
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -708,7 +708,6 @@ function jeAddRoutineOperationCommands(command, defaults) {
     show: jeRoutineCall((_, routine)=>Array.isArray(routine), true)
   });
 
-  defaults.skip = false;
   for(const property in defaults) {
     jeCommands.push({
       id: 'default_' + command + '_' + property,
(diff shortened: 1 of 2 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1437/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1437-pr-test`.

96LawDawg (approved):
Approved through test.

ArnoldSmith86 (changes requested):
Nice stuff. One minor issue: closing with ESC or by toggling the JSON editor, doesn't update the color of the icon.

96LawDawg:
Tested the new changes to the debugger and it works with the exception @ArnoldSmith86 noted.

client/js/jsonedit.js
```diff
@@ -2198,6 +2197,10 @@ function jeToggle() {
     $('body').classList.add('jsonEdit');
     if(jeWidget && !widgets.has(jeWidget.id))
       jeEmpty();
+    if(jeDebugViewing) {
+      jeCallCommand(jeCommands.find(o => o.id == 'je_toggleDebug'));
+      jeShowCommands()
+    }
```
ArnoldSmith86, line 2203: Doesn't `jeEmpty` already do that? Maybe just move it up a bit to avoid a second call?

rogerl50: I considered that. But right now, `jeEmpty` is being called only if there was a widget being displayed before that has now been deleted. So we could make the `jeEmpty` call unconditional and move the new code up, but I'm not sure that buys much if anything.

ArnoldSmith86: I'm just careful after the optimization PR. Most of the slowdowns came from just carelessly calling the same function again and again.

Both of those are in very narrow conditions though so just keep it as it is.

96LawDawg:
Concur that the fix works.  Should be ready to merge.

Issue #1442, opened 2022-12-12
Automatic background - text luminance correction


96LawDawg:
@rogerl50 just put a nice function into the scoreboard to automatically create the right contrast between light backgrounds and black text and darker backgrounds and white text.  Many times, I would have loved to have been able to do that automatically in JSON.  I thought about building a routine, but it was never worth the effort.  Could something be built behind the scenes and we invoke it through some variable or property?  Maybe add a new variable like "playerContrastColor" for either black or white to use with "playerColor."  Something like:

```
"css": {
    "background": "${PROPERTY backColor}",
    "color": "${PROPERTY textColor}"
  },

  "clickRoutine": [
    {
      "func": "SET",
      "collection": "thisButton",
      "property": "backColor",
      "value": "${playerColor}"
    },
    {
      "func": "SET",
      "collection": "thisButton",
      "property": "textColor",
      "value": "${playerContrastColor}"
    }
  ]
}
```

PR #1400, opened 2022-11-12
added pre-defined variables seatIndex, seatID and collection playerSeats

ArnoldSmith86:
Implements #1397. It adds two variables and one collection that are available in every routine.

The new variable `seatID` contains the `id` of the first seat the active player is sitting in and the variable `seatIndex` contains the `index` of that seat. If the player does not sit in any seat, the variables are set to `null`.

The collection `playerSeats` contains all seat widgets the player is currently sitting in.

PR #560, opened 2021-06-11
variables playerName and playerColor are now writable

ArnoldSmith86:
We discussed this in a meeting long, long ago.

Please note that this can only set the name and color of the player triggering the routine.

PR #1443, opened 2022-12-13
Add compute functions to manipulate colors

96LawDawg:
Implements #1442.  Creates 6 new compute functions: `colorToRGB`, `colorToHex`, `colorLuminace`, `colorContrastRatio`, `colorContrast`, and `colorCreateHue`.  Each will accept all the common color formats and output what is implied in the name.  

For the `colorContrast` function, the goal is to make it easy to create contrasting colors that work with the luminance of the starting color.  So you end up with something in the same hue if you want, or white/black if you want.  Demo in the test room.

For the `colorCreateHue` function, this creates a new semi-random color that will be as different as possible from the colors of the active players.  This is the same formula used to generate colors when `playerColor` is not in local storage.  Room designers could use this to create a color when `toHex` returns #000000 (which will usually mean there was a problem with the input passed in the `toHex` function).

The JavaScript for these functions are also now available to coders for other projects involving color.

### Wiki Stuff:

Update the predefined variables list in Routines (https://github.com/ArnoldSmith86/virtualtabletop/wiki/Functions) with

`- activeColors - array of colors of active players in room`

Draft addition to be added to the bottom of the wiki entry at https://github.com/ArnoldSmith86/virtualtabletop/wiki/Computations:

## Color Operations
  <details>
    <summary>Converting color formats</summary>
    
- `colorToRGB`: Accepts most color formats in operand1 and returns a value in the format "rgb(0,99,255)"
- `colorToHex`: Accepts most color formats in operand1 and returns a hex value in the format "#ffffff". Note: all of the color functions use this to try to convert non-hex color formats to hex. If the format or value is invalid, the output will be black (#000000). You can use this in your routines as an error check; if black is returned, it is likely an error unless black was the operand1 color.
  </details>
    
<details>
<summary>Creating contrasting color</summary>

- `colorContrast`: Accepts most color formats in operand1 and returns a hex color value in the same hue but different luminance of the provided color to provide contrast. Operand2 (optional) can be a number between -1 and 1 to adjust the intensity of the contrasting change.  Defaults to 1.  The higher the number, the greater the contrast.  With the default, dark colors get lighter return values. Using negative values for operand2 will reverse the direction of the change so that dark colors get darker return values.

- `colorContrastRatio`: Accepts most color formats in operand1 and operand2 to return a number between 1 and 21 that is the ratio of the relative luminance of the colors ([as defined in WCAG 2.1](https://www.w3.org/TR/WCAG21/#dfn-contrast-ratio)).
- `colorLuminance`: Accepts most color formats in operand1 and returns a value between 0 and 1 reflecting the luminance of the color
- `colorCreateHue`: Creates a semi-random color as distinct in hue as possible from active player colors. Operand1 (optional) can be an array of colors in most color formats to use instead of player colors.

  </details>


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1443/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1443-pr-test`.

ArnoldSmith86:
I feel like this should be a compute function instead so that you can use it for any color and not just the active player color.

96LawDawg:
> I feel like this should be a compute function instead so that you can use it for any color and not just the active player color.

Actually, I had a similar idea.  I've been working on making it even more useful (at least to me).  Stand by.

rogerl50:
I see that's a slightly different computation than the one I did. Did you find that somewhere? And does it produce better results than mine?

Also, is it possible that `players[player]` can be a named color rather than a hex string? (That is, are you sure that this is always of the form `#xxxxxx` rather than `white` or `rgb(20,50,70)`?)

RaphaelAlvez:
> I got it from ChatGPT.  Forgot to put that in the comments in the code.  But this is going to change anyway.  I don't know that it produces better results, but in the tests I did, it worked great.

Now I want to see a side by side with some grey/middle luminosity values. It seems to me that chatGpt is using the css color correction for eyesore something similar. I believe that comes from YCM to RGB
 
> As far as I know playerColor is always a hex value by default.  The only way that would change is if someone SET the value.  In which case this wouldn't work.  But it's going to change.

Yeah I believe it gets converted

ArnoldSmith86:
> Also, is it possible that `players[player]` can be a named color rather than a hex string? (That is, are you sure that this is always of the form `#xxxxxx` rather than `white` or `rgb(20,50,70)`?)

See `populateEditOptionsCanvas` in `client/js/editmode.js`. It's "using canvas fillStyle to turn color names into hex colors". @rogerl50 Didn't you look for something like that for the scoreboard as well?

96LawDawg:
Here's how I think this should change.

Make a new function: `COLOR`.  It takes a color input and generates a color output.  It operates in one of several modes.  In "mono" mode, it operates just like what I did originally and produces either black or white.  In "hue" mode, it matches the hue but changes the brightness up or down based on the luminosity of the start color.  Could also add a "complementary" mode and "inverted" mode, but I personally wouldn't find those very useful.

In hue mode, I've found that about 30% difference usually works great, but having an "intensity" parameter would make it more flexible.

I have been wanting for a long time a way to automatically modify the color a little bit, but keep the same hue.  You can use this for text on an object, for changing the color of a holder, for making a series of pieces in the same shade (like I did in Capture the Flag), for complex coloring (like in my marbles).

Here is a room with some sample colors: https://virtualtabletop.io/5ldt

Here is the heart of the code that makes it work.  Still would need to make it a VTT function, of course:  https://jsfiddle.net/bjfdkvy2/19/.  This is the result of a long series of back and forth conversations with ChatGPT.  I couldn't find any existing code that did something like this automatically.

What do you think?

ArnoldSmith86:
Sure.

RaphaelAlvez:
No I think this should be compute functions. This feels like reintroducing COMPUTE.

Functions always interact with widgets or collections. Your proposal seems to only interact with variables.

Also it should be easier to right and add new stuff

ArnoldSmith86:
I kinda agree that it should be a compute function. But then don't stuff all of it into one function.

96LawDawg:
> No I think this should be compute functions. This feels like reintroducing COMPUTE.
> 
> Functions always interact with widgets or collections. Your proposal seems to only interact with variables.
> 
> Also it should be easier to right and add new stuff

Good points.  I just want it to have the flexibility to do at least the 2 things I mentioned above.

But this is some new kind of compute function, right?  It doesn't fall neatly into the math, comparison, string, or array stuff we have now.  So we create a new category of color functions?  One for mono, one for hue, etc.?

ArnoldSmith86:
> So we create a new category of color functions? One for mono, one for hue, etc.?

Sure, why not.

rogerl50:
> > Also, is it possible that `players[player]` can be a named color rather than a hex string? (That is, are you sure that this is always of the form `#xxxxxx` rather than `white` or `rgb(20,50,70)`?)
> 
> See `populateEditOptionsCanvas` in `client/js/editmode.js`. It's "using canvas fillStyle to turn color names into hex colors". @rogerl50 Didn't you look for something like that for the scoreboard as well?

I used this, which I think is similar:
```
    function getHexColor(color) {
      const a = document.createElement('div');
      a.style.color = color;
      const colorStr = window.getComputedStyle( document.body.appendChild(a) ).color;
      const colors = colorStr.match(/\d+/g).map(function(a){ return parseInt(a,10); });
      document.body.removeChild(a);
      return colors;
    }
```

96LawDawg:
> Also, is it possible that `players[player]` can be a named color rather than a hex string? (That is, are you sure that this is always of the form `#xxxxxx` rather than `white` or `rgb(20,50,70)`?)

After more research and testing, it looks like `playerColor` will ALWAYS be hex.  I've tried setting it using red/green/etc, HSL, RGB.  While that works in css, it doesn't work for `playerColor`.  Probably because of the way the `var playerColor` function works based on https://github.com/ArnoldSmith86/virtualtabletop/pull/560.

96LawDawg:
Update:  I reverted the original change.  I created a new compute function "colorMono" to do what the original commit did.  In order to test it, I also needed to create a new `rgbToHex` function, which does exactly what it says.  You can test it for yourself by moving the red, green, and blue pawn sliders individually or by moving the gray one to move them together to create grays.  The only important thing is the new functions on the basic widget.  They work just fine.

I don't know that we need all the color converter functions created, but it isn't too much work I don't think.  So I will probably do hex, rgb, and hsl converters just to have them there.  And I will add the new function that creates the hue effect that I demonstrated earlier.

Am I missing anything this PR needs?

96LawDawg:
I added the `colorHue` function.  It takes the form of var a = (hex color) (optional intensity value 0 to 0.999) (optional direction value of -1).  Default is 0.6 for intensity and 1 for direction.

I realized that `colorMono` is really just a subset of `colorHue`.  You can set values on `colorHue` that duplicate what `colorMono` does.  Almost.  But I can't figure out something in the compute functions.  My function is "function(v, x, y, z)."  If y and z are not passed in the var statement, then they show up as 1 in the function.  I can't figure out how to tell if the user passed a 1 or if it is a default 1.  Any advice?

client/js/compute.js
```diff
@@ -684,6 +684,79 @@ const compute_ops = [
...
+      var rgbRegex = /^rgb\(\s*(-?\d+)(%?)\s*,\s*(-?\d+)(%?)\s*,\s*(-?\d+)(%?)\s*\)$/;
+      var result, r, g, b, hex = "";
+      if ( (result = rgbRegex.exec(x)) ) {
+          r = componentFromStr(result[1], result[2]);
+          g = componentFromStr(result[3], result[4]);
+          b = componentFromStr(result[5], result[6]);
+  
+          hex = "#" + (0x1000000 + (r << 16) + (g << 8) + b).toString(16).slice(1);
+      }
+      return v = hex;
```
rogerl50, line 757: I think that if `colorStr` is `rgb(a,b,c)` or `rgba(a,b,c,d)` that this will do the job:
```
colors = colorStr.match(/\d+/g).map(function(a){ return parseInt(a,10); });
hex = '#' + colors.map(c => c.toString(16)).join('');
```
I haven't looked at the other functions, but I suspect that a similar improvement is possible.

rogerl50 (changes requested):

client/js/compute.js
96LawDawg: I tried that and it didn't work.  But maybe I had a typo or something else happened. I'll try again.

rogerl50: Take a look at https://jsfiddle.net/rxl50/wLp9kb41/

96LawDawg: It doesn't handle the case where one of the values is a single digit.

rogerl50: `hex = '#' + colors.map(c => c.toString(16).padStart(2,"0")).join('');` solves that problem.

client/js/compute.js
```diff
@@ -684,6 +684,80 @@ const compute_ops = [
...
+      } else {
+        modifier = (luminance > 0.5) ? -y : y;
+      }      
+      const colors = [r, g, b];
+      const adjustedColors = [];
+      for (let i = 0; i < colors.length; i++) {
+        const adjustedColor = Math.round(Math.max(0, Math.min(colors[i] + modifier * 255, 255)));
+        adjustedColors.push(adjustedColor);
+      }
+      const [adjustedR, adjustedG, adjustedB] = adjustedColors;
```
rogerl50, line 729: The previous collection of lines (723-729) can be replaced by
```
const [adjustedR, adjustedG, adjustedB] = [r,g,b].map(c => Math.round(Math.max(0, Math.min(c+ modifier * 255, 255))));
```

client/js/compute.js
```diff
@@ -684,6 +684,80 @@ const compute_ops = [
...
+      let g = parseInt(x.slice(3, 5), 16);
+      let b = parseInt(x.slice(5, 7), 16);
+      let luminance = (0.2126 * r + 0.7152 * g + 0.0722 * b) / 255;
+      return v = (luminance <= 0.5) ? '#ffffff' : '#000000';
+    },
+    hash: '123'
+  },
+  {
+    name: 'colorHue',
+    desc: 'returns a hex color value in the same hue but different luminance of the provided color to provide contrast',
```
rogerl50, line 704: Do you want to describe the parameters here, or is that left to the wiki?

rogerl50 (changes requested):

96LawDawg:
This is all I intend to do on this PR.  99% of the stuff I've seen on VTT uses hex and occasionally RGB for color.  So functions to convert between those 2 is probably enough.  Can always add HSL if wanted later.  Also, this fulfills most of color needs that I've had over the years so adding additional color changing functions would, I think, be overkill.

So this is ready for a code review and testing.  I will see if I can figure out the hashes, but I may need help with that.  Also, I would be okay with getting rid of `colorMono` because `colorHue` can do the same thing if you change the defaults.  But I like having both to make it easy.

Draft addition to be added to the bottom of the wiki entry at https://github.com/ArnoldSmith86/virtualtabletop/wiki/Computations:

## Color Operations
  <details>
    <summary>Converting color formats</summary>
    
- `hexToRGB`: Accepts a hex color value in operand1 in the format "#000" or "#ffffff" and returns a value in the format "rgb(0,99,255)"
- `rgbToHex`: Accepts a rgb color value in operand1 in the format "rgb(0,99,255)" and returns a hex value in the format "#ffffff"
  </details>
    
<details>
<summary>Creating contrasting color</summary>
  
- `colorMono`: Accepts a hex color value in operand1 in the format "#ffffff" and returns a color value of #000000 or #ffffff, whichever provides better contrast
- `colorHue`: Accepts a hex color value in operand1 in the format "#ffffff" and returns a hex color value in the same hue but different luminance of the provided color to provide contrast.
     - Operand2 (optional) can be a number between 0 and 0.999 to adjust the intensity of the contrasting change.  Defaults to 0.6.  The higher the number, the greater the contrast.
     - Operand3 (optional) can be either 1 or -1 (defaults to 1). With the default, dark colors get lighter return values. Using -1 will reverse the direction of the luminosity change so that dark colors get darker return values.

  </details>

client/js/compute.js
96LawDawg: I think I should add the input and output to be consistent with the other 2.  It is also in the draft wiki entry I just submitted.

RaphaelAlvez:
@ArnoldSmith86 didn't you have a trick to convert any color format to a specific one?

ArnoldSmith86:
I don't have time for a proper review but here are some thoughts I had:

- I don't like the names `colorMono` and `colorHue`. They say nothing about creating a contrasting color.
- The actual code should probably be outside `compute.js` so we can use them easily from other parts of VTT. I would do that from the start so we're sure we don't break anything when changing it.
- Personally, I would have changed `rgbToHex` to `colorToHex` using the canvas trick I posted earlier. That should convert any color the browser knows to a hex color.

RaphaelAlvez:
> - Personally, I would have changed `rgbToHex` to `colorToHex` using the canvas trick I posted earlier. That should convert any color the browser knows to a hex color.

Did not see that comment.

@96LawDawg this should even support css default colors like "red"

ArnoldSmith86:
> Did not see that comment.

I posted it after you did. We typed at the same time.

ArnoldSmith86:
> I will see if I can figure out the hashes, but I may need help with that.

You can look at the TestCafé output on GitHub (at least I think that I'm not the only one) and it tells you "hash xy was expected to be yx". So you can just replace it in the code by that value. Unfortunately it stops at the first mismatch, I think. So you can just do one at a time. Of course, you can get the same output locally.

robartsd:
> I see that's a slightly different computation than the one I did. Did you find that somewhere? And does it produce better results than mine?
> 
> Also, is it possible that `players[player]` can be a named color rather than a hex string? (That is, are you sure that this is always of the form `#xxxxxx` rather than `white` or `rgb(20,50,70)`?)

The server side code for creating new player colors currently requires that player color be in hex format.

client/js/color.js
```diff
@@ -0,0 +1,15 @@
...
+  });
+  return `rgb(${rgb.join(',')})`;    
+}
+
+export function toHex(inputColor) {
+  const canvas = document.createElement('canvas');
+  const ctx = canvas.getContext('2d');
+  ctx.fillStyle = inputColor;
+  const hexColor = ctx.fillStyle;  
+  return hexColor;
```
rogerl50, line 14: Should the canvas be deleted before returning?

rogerl50 (dismissed):

rogerl50:
Is this worth a separate file, or should the color functions just be placed in main? I don't know the right answer, just asking.

client/js/color.js
96LawDawg: I don't know, but I added it because that can't hurt.

96LawDawg:
> I don't have time for a proper review but here are some thoughts I had:
> 
> * I don't like the names `colorMono` and `colorHue`. They say nothing about creating a contrasting color.

I now just have one function `colorContrast`.

> * The actual code should probably be outside `compute.js` so we can use them easily from other parts of VTT. I would do that from the start so we're sure we don't break anything when changing it.

Done.  Added it to a new file color.js in case we want to add more and we want to keep this separate.  I have no problem putting it in main if you want.

> * Personally, I would have changed `rgbToHex` to `colorToHex` using the canvas trick I posted earlier. That should convert any color the browser knows to a hex color.

Done.  And this was a great suggestion.  Now I run everything through the `toHex` function in color.js.  So most anything can now be converted to hex.  And anything can be converted to RGB.  And anything can be run through colorContrast.  Much more flexibility and the functions are now available across VTT.

What else?

rogerl50:
What happened to the original `colorToMono` function that started all of this? Can I get that effect from the general contrast function?

96LawDawg:
> What happened to the original `colorToMono` function that started all of this? Can I get that effect from the general contrast function?

Yes, but you have to override the default.  `"var newColor = colorContrast 0.99`".  I would have made it 1 except for this https://github.com/ArnoldSmith86/virtualtabletop/pull/1443#issuecomment-1350491554.

rogerl50:
> What happened to the original `colorToMono` function that started all of this? Can I get that effect from the general contrast function?

I see. If you

rogerl50:
> > What happened to the original `colorToMono` function that started all of this? Can I get that effect from the general contrast function?
> 
> Yes, but you have to override the default. `"var newColor = colorContrast 0.99`". I would have made it 1 except for this [#1443 (comment)](https://github.com/ArnoldSmith86/virtualtabletop/pull/1443#issuecomment-1350491554).

Doesn't that mean that you have to decide before hand whether you want the text to be white or black? (that is, you have to set the direction).

Second question: why the special case for intensity = 1? It doesn't look as though you pass a default of 1 in from the compute function.

96LawDawg:
> Doesn't that mean that you have to decide before hand whether you want the text to be white or black? (that is, you have to set the direction).

No, it does the work for you.  Leaving the direction alone will be your best option always for text.  I just wanted a better way to change colors using the function.  So for example, you want some game pieces.  You intentionally pick a middle luminosity color to start with (don't make it playerColor because it could be anything).  Make some darker variants and some lighter variants by changing direction.

> Second question: why the special case for intensity = 1? It doesn't look as though you pass a default of 1 in from the compute function.

Somehow compute function x, y, and z variable seem to get a default of 1.  I don't know why.  I was sure it would be null or undefined, but testing showed otherwise.  So if the user leaves it as default, then a 1 is passed, but 1 makes white or black.  I need a default of 0.6.  Any ideas?

rogerl50:
> Somehow compute function x, y, and z variable seem to get a default of 1. I don't know why. I was sure it would be null or undefined, but testing showed otherwise. So if the user leaves it as default, then a 1 is passed, but 1 makes white or black. I need a default of 0.6. Any ideas?

The default comes from the `getValue` routine in `widget.js`, which looks like it sets up default parameter values for all compute functions. Changing it is above my pay grade.

But why do you need a default of 0.6? IMO that produces results that are not contrasty enough.

96LawDawg:
> > Somehow compute function x, y, and z variable seem to get a default of 1. I don't know why. I was sure it would be null or undefined, but testing showed otherwise. So if the user leaves it as default, then a 1 is passed, but 1 makes white or black. I need a default of 0.6. Any ideas?
> 
> The default comes from the `getValue` routine in `widget.js`, which looks like it sets up default parameter values for all compute functions. Changing it is above my pay grade.

AND would probably break everything. 

> But why do you need a default of 0.6? IMO that produces results that are not contrasty enough.

I wanted contrast in the same hue.  That's why I built this.  I would go with maybe 0.7 or 0.75.  But after that, it might as well be black or white.  I could put back a separate function like textColorContrast that would default to 1 and make black or white.  Maybe that's a good idea so we don't fight over the default?

rogerl50:
> > > Somehow compute function x, y, and z variable seem to get a default of 1. I don't know why. I was sure it would be null or undefined, but testing showed otherwise. So if the user leaves it as default, then a 1 is passed, but 1 makes white or black. I need a default of 0.6. Any ideas?
> > 
> > 
> > The default comes from the `getValue` routine in `widget.js`, which looks like it sets up default parameter values for all compute functions. Changing it is above my pay grade.
> 
> AND would probably break everything.
> 
> > But why do you need a default of 0.6? IMO that produces results that are not contrasty enough.
> 
> I wanted contrast in the same hue. That's why I built this. I would go with maybe 0.7 or 0.75. But after that, it might as well be black or white. I could put back a separate function like textColorContrast that would default to 1 and make black or white. Maybe that's a good idea so we don't fight over the default?

Maybe. It's just unfortunate, because documenting that using 1 (a natural value) to get high contrast would be reasonable, but it doesn't work. Seems odd to ask the user to use 0.99 or 1.1 to get high contrast (essentially black/white). Maybe someone else will chime in with a better idea.

96LawDawg:
There is a better solution. I was just too hard headed to see it. Let the default be 1. Black or white. If you want something less, then plug that in. That makes more sense than forcing an entry of .999. I’ll change the code later when I have time. Busy day.

robartsd:
> Somehow compute function x, y, and z variable seem to get a default of 1. I don't know why. I was sure it would be null or undefined, but testing showed otherwise. So if the user leaves it as default, then a 1 is passed, ...

This is was the default value from the old COMPUTE function. It would be nice if each function could set its own default for each parameter.

robartsd:
The code for `contrastAnyColor` is adjusting all channels (r, g, and b) by the same fixed amount (clamped between 0 and 255). This results in a contrasting color of similar hue; however the resulting hue may shift somewhat if some of the channels reach the clamped values.

96LawDawg:
> The code for `contrastAnyColor` is adjusting all channels (r, g, and b) by the same fixed amount (clamped between 0 and 255). This results in a contrasting color of similar hue; however the resulting hue may shift somewhat if some of the channels reach the clamped values.

True.  But I don't know of a better formula and that is only an issue for values close to the extremes.  I tried working in HSL, to lighten the color directly.  But that didn't give the results I wanted.  Have you played with the sliders in the test room?  If not, give that a try.  I think the results are still good to the eye, even if the math doesn't work out just right.

client/js/color.js
```diff
@@ -0,0 +1,42 @@
...
+  return (
+    "#" +
+    adjustedR.toString(16).padStart(2, '0') +
+    adjustedG.toString(16).padStart(2, '0') +
+    adjustedB.toString(16).padStart(2, '0')
+  )
+}
+
+export function toHex(inputColor) {
+  const canvas = document.createElement('canvas');
```
robartsd, line 36: Perahps we should add a test to check that inputColor is not already in hex format before converting.

client/js/color.js
```diff
@@ -0,0 +1,42 @@
...
+  return `rgb(${rgb.join(',')})`;    
+}
+
+export function contrastAnyColor(inputColor, intensity, direction) {
+  // Code created by https://chat.openai.com/chat
+  let color = toHex(inputColor);
+  const r = parseInt(color.slice(1, 3), 16);
+  const g = parseInt(color.slice(3, 5), 16);
+  const b = parseInt(color.slice(5, 7), 16);
+  const luminance = (0.2126 * r + 0.7152 * g + 0.0722 * b) / 255;
```
robartsd, line 16: Might be nice to move luminance calculation to it's own function. Would be nice to have a function that calculates the [contrast between colors](https://www.w3.org/TR/2008/REC-WCAG20-20081211/#contrast-ratiodef) which would also make use of the luminance calculation.

client/js/color.js
```diff
@@ -1,3 +1,16 @@
+export function toHex(inputColor) {
+  const hexPattern = /^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$/;
```
robartsd, line 2: #rgb should be expanded to #rrggbb (or just not match this case) as other code expects a full 6 hex digits.

96LawDawg: Done

client/js/color.js
96LawDawg: Done

96LawDawg:
Incorporated all requested changes.
 - toHex excludes already valid 6-digit hex colors
 - `colorContrast` now defaults to white or black depending on luminance
 - luminance now is its own function
 - added function to calculate contrast ratio
 - @ArnoldSmith86 already modified Editmode.js to use the new toHex function

Anything else?  Ready for final code review and/or testing.  I'm working on the Tutorial.

96LawDawg:
Tutorial is ready in the pr test room.  It looks similar to what I've already demonstrated, but I added a series of player pieces/pins.  Click on one and they all change to a series of colors related to playerColor.  Really cool.

client/js/color.js
```diff
@@ -0,0 +1,57 @@
...
+  const hexColor = ctx.fillStyle;  
+  canvas.remove();
+  return hexColor;
+}
+
+export function toRGB(inputColor) {
+  let hex = toHex(inputColor)
+  hex = hex.slice(1);
+  var rgb = hex.match(new RegExp('(.{' + hex.length/3 + '})', 'g')).map(function(l) {
+    return parseInt(hex.length%2 ? l+l : l, 16);
```
robartsd, line 18: I beleive that toHex() is returning starndardized #rrggbb values, so test of length is not needed.

96LawDawg: Good point.  That was leftover from when each function did its own thing before going through toHex.  Hopefully this is better.

client/js/color.js
```diff
@@ -0,0 +1,57 @@
...
+  });
+  return `rgb(${rgb.join(',')})`;    
+}
+
+export function calcLuminance(inputColor) {
+  let color = toHex(inputColor);
+  const r = parseInt(color.slice(1, 3), 16);
+  const g = parseInt(color.slice(3, 5), 16);
+  const b = parseInt(color.slice(5, 7), 16);
+  const luminance = (0.2126 * r + 0.7152 * g + 0.0722 * b) / 255;
```
robartsd, line 26: Should we apply sRGB gamma correction as [defined in WCAG 2.x](https://www.w3.org/WAI/GL/wiki/Relative_luminance)?

96LawDawg: I'm not sure if I understand the difference.  The notes so there is no actual difference in the 8-bit colors we are using.  So I changed the formula to be what I think WCAG 3 is saying.  I got the same results (as expected).  Is this what you menat?

robartsd: The gamma correction is this part:
```
if RsRGB <= 0.03928 then R = RsRGB/12.92 else R = ((RsRGB+0.055)/1.055) ^ 2.4
if GsRGB <= 0.03928 then G = GsRGB/12.92 else G = ((GsRGB+0.055)/1.055) ^ 2.4
if BsRGB <= 0.03928 then B = BsRGB/12.92 else B = ((BsRGB+0.055)/1.055) ^ 2.4
```

96LawDawg: Okay.  I think I got it.  But I need to change the threshold.

96LawDawg:
After going back to my demo room, I noticed that the default `colorContrast` was not always giving the best results.  That's because it was only using luminance as the trigger.  So now, when using the default intensity (trying to get the best black or white text), the function also checks for the `colorContrastRatio` and uses the best one.  This produces much better results, especially with gray backgrounds.  If something other than the default intensity is used (like the pin example in the demo room), then the designer probably wants a certain effect and the contrast ratio is not used.

96LawDawg:
"`var playerColor = 'red'" now works.  Before you had to use a hex value.  Now it uses the `toHex` function to accept pretty much anything.  If it isn't a valid color, then you get #000000 out of toHex.  Maybe we could change that to red like the old default color.  But leaving it as black shows that something probably went wrong and may be useful in its own way.

robartsd:
> "`var playerColor = 'red'"` now works. Before you had to use a hex value. Now it uses the `toHex` function to accept pretty much anything. If it isn't a valid color, then you get #000000 out of `toHex`. Maybe we could change that to red like the old default color. But leaving it as black shows that something probably went wrong and may be useful in its own way.

I like this change. It would be nice to have a way to generate a color in a way similar to the way the server initially generates the color for a new player; but I don't think that should be applied in all cases of invalid colors.

96LawDawg:
> I like this change. It would be nice to have a way to generate a color in a way similar to the way the server initially generates the color for a new player; but I don't think that should be applied in all cases of invalid colors.

I can create a feature to do that.  But I don't understand the last part.  What do you mean about invalid colors?

robartsd:
> > I like this change. It would be nice to have a way to generate a color in a way similar to the way the server initially generates the color for a new player; but I don't think that should be applied in all cases of invalid colors.
> 
> I can create a feature to do that. But I don't understand the last part. What do you mean about invalid colors?

Currently `toHue()` returns '#000000` for invalid colors (colors that are not valid css color values).

> Also, that function you created a while back looked at just other players in the room for comparison. I assume that is what this new function would do. Something like `colorRandom` that would try to get a hue that is not close to the hue of another player in the room?

Yes, I was thinking something like that; though really this idea probably should be a separate PR.

robartsd:
I was thinking that instead of a direction parameter, direction should be indicated by the sign of the intensity parameter. This would free up the third parameter to possibly extend this to accept a second color. In the extended version, the second color could be used for the output hue.

96LawDawg:
> I was thinking that instead of a direction parameter, direction should be indicated by the sign of the intensity parameter. This would free up the third parameter to possibly extend this to accept a second color. In the extended version, the second color could be used for the output hue.

That seems like a smart idea and easier for the designer.  I'll work on that.  But I don't understand how a third parameter would work.  Would it be doing interpolation between the two of them?  That is a completely different concept.

robartsd:
> That seems like a smart idea and easier for the designer. I'll work on that. But I don't understand how a third parameter would work. Would it be doing interpolation between the two of them? That is a completely different concept.

I envision the output color using hue and saturation from the third parameter and luminosity to contrast with the first parameter (amount of contrast based on intensity setting).

robartsd:
Compute test should also be updated to have some valid color values passed to functions (not necessarily before merging this PR). Perhaps we should categorize all functions and provide more customized compute tests depending on the category rather than making compute tests even longer for all functions.

96LawDawg:
> > Also, that function you created a while back looked at just other players in the room for comparison. I assume that is what this new function would do. Something like `colorRandom` that would try to get a hue that is not close to the hue of another player in the room?
> 
> Yes, I was thinking something like that; though really this idea probably should be a separate PR.

How about this idea (I already implemented it).  Add a new compute function `colorCreateHue`.  It uses the exact same code used on the server side to create a randomish player color.  That way if the designer wants to find a color to replace #000000, he has an option to use an IF statement to do that.  Was easy enough to do so no need for a separate PR unless you had something really different in mind.

I could not figure out a way to call the same function on the server side and on the client side.  So the same code exists twice.  If it is possible to avoid that duplication, either let me know how or just make the change.

client/js/compute.js
```diff
@@ -719,6 +721,13 @@ const compute_ops = [
     sample: 'var a = colorCLuminance ${x}',
     call: function(v, x, y) { return v = calcLuminance(x); },
     hash: 'f4284956510e3fe59c2babed665b544d'
+  },
+  {
+    name: 'colorCreateHue',
+    desc: 'returns a visually distinct, random hex color using linear interpolation based on all active player colors',
+    sample: 'var a = colorCreateHue',
+    call: function(v, x, y) { return v = calcHue(activeColors); },
```
robartsd, line 729: How about we don't hardcode the array of colors to use and instead add a variable `activePlayerColors` containing the colors of active players, then this could be called as `"var a = colorCreateHue ${activePlayerColors}"` or game designers can choose a different array of colors as input.

client/js/color.js
```diff
@@ -64,4 +66,38 @@ export function contrastAnyColor(inputColor, intensity, direction) {
   const toHexString = (r, g, b) => '#' + r.toString(16).padStart(2, '0') + g.toString(16).padStart(2, '0') + b.toString(16).padStart(2, '0');
   return toHexString(...adjustColor(r, g, b, modifier));
   return defaultColor;
+}
+
+export function calcHue(startingColors) {
```
robartsd, line 68: I don't like this function name. I would expect calcHue to take a color and return the hue value of that color.

96LawDawg: `randomHue`?

client/js/color.js
```diff
@@ -64,4 +66,38 @@ export function contrastAnyColor(inputColor, intensity, direction) {
...
+  let hue = 0;
+  const hues = [];
+  for(const player in startingColors) {
+    const hex = startingColors[player];
```
robartsd, line 72: Should call toHex so that input colors don't need to be in `#rrggbb` format.

96LawDawg: Yes.  I'll make that change.  It wasn't necessary before, but when we allow designers to use an array, it is necessary.

robartsd:
If we really want to get serious about color we could use [this library](https://github.com/LeaVerou/color.js).

96LawDawg:
Improved the `colorCreateHue` function as @robartsd suggested.  Now use it as `"var a = colorCreateHue"` and it will find a color based on active players' colors.  Or pass an optional operand1 with an array of colors.  So this also works:

```
"colorArray": [
    "red",
    "green",
    "blue"
  ],

  "clickRoutine": [
    "var newColor = colorCreateHue ${PROPERTY colorArray}",
    {
      "func": "SET",
      "collection": "thisButton",
      "property": "bColor",
      "value": "${newColor}"
    }
  ]
  ```
  Doing it this way actually made testing much easier.  Easier to create an array than to go in as 3 different players and set up different colors.  So the color scheme above will usually result in something in the yellow/orange range (as you would expect).  But sometimes it is light yellow/green.  But that's just the way color and limited range randomness works.

**Comment by 96LawDawg** [comment]
> If we really want to get serious about color we could use the [color.js library](https://github.com/LeaVerou/color.js).

I thought about that; actually a different and slightly smaller library.  But the reality is game designers probably don't need that much functionality.  And nothing I found did anything like my `colorContrast` function.  I think we should go with what we have for now.  If people need more, they'll let us know.

**Comment by robartsd** [comment]
> I could not figure out a way to call the same function on the server side and on the client side. So the same code exists twice. If it is possible to avoid that duplication, either let me know how or just make the change.

I agree that it would be better to have the code just once, but I'm not sure how that would be accomplished; I know very little about the server side.

**Comment by 96LawDawg** [comment]
> > I could not figure out a way to call the same function on the server side and on the client side. So the same code exists twice. If it is possible to avoid that duplication, either let me know how or just make the change.
> 
> I agree that it would be better to have the code just once, but I'm not sure how that would be accomplished; I know very little about the server side.
> 
> I think it might be possible for server side files to import functions directly from `client/js/color.js`.

I think I got it to work.  But it's hard to tell other than by brute force testing, which seems to work.  I made the relevant function in room.mjs call our new function in color.js.  Don't get any errors and as I said, brute force testing works.  I get a selection of red/yellow/orange hues when the people in the room have blue/green/purple colors.  That is to be expected.  But putting a console.log statement in to see what the actual array being passed is, does not work.  Must be something with server-side stuff.

See what you think.

File: client/js/color.js [hunk 13]
```diff
@@ -0,0 +1,101 @@
+export function toHex(inputColor) {
+  const hexPattern = /^#([A-Fa-f0-9]{6})$/;
+  if (hexPattern.test(inputColor)) {    
+    return inputColor;
+  }
+  const canvas = document.createElement('canvas');
+  const ctx = canvas.getContext('2d');
+  ctx.fillStyle = inputColor;
+  const hexColor = ctx.fillStyle;
```
robartsd, line 9: There is a chance that the fillstyle will not be a `#rrggbb` color. If an invalid value is provided the default of `#000000` is returned, but gradients and patterns are also valid fill styles and CSS 4 colors include LCH and LAB color functions which can result in colors outside sRGB. Apple has already created devices that support colors outside sRGB and added these color spaces to Safari.

96LawDawg:
Just posting this here so maybe I can find it again someday if I need it.  Great article on converting colors in JavaScript:

https://css-tricks.com/converting-color-spaces-in-javascript/

robartsd:
> Just posting this here so maybe I can find it again someday if I need it. Great article on converting colors in JavaScript:
> 
> https://css-tricks.com/converting-color-spaces-in-javascript/

Yes, except that those are all the same color space with different coordinate systems applied - they all have the same range of colors that can be expressed (implicitly sRGB). Good reference for HSL and HSV from RGB. The canvas fill style method of converting to hex will work for all of those coordinate systems, but may not work for coordinate systems which contain colors outside of sRGB.

robartsd:
I'm thinking about writing functions so that luminance can be adjusted directly and overloading the contrastAnyColor function such that intensity absolute value > 1 will attempt to make a color with a WCAG 2 contrast ratio equal to intensity.

rogerl50 (approved):
IMO, better is the enemy of good. If we are all in agreement that the interfaces defined here are good, and that the routines work as designed, I think we should merge this and add to it later.

96LawDawg:
> I'm thinking about writing functions so that luminance can be adjusted directly and overloading the contrastAnyColor function such that intensity absolute value > 1 will attempt to make a color with a WCAG 2 contrast ratio equal to intensity.

I agree with Roger.  What's the use case for this?

I'm still open to adding a third parameter to `contrastColor`, but I still don't know what calculation is being performed.  And we could always add that option later without impacting any games that start using it with how it is now.  Your features might be nice to have, but what we have now puts something useful into game designer hands.  Any reason to wait on merging?

robartsd:
>  What's the use case for this?

My intention is to create much more consistency in the level of contrast created. Results of current function with intensity = .3 can have a fairly large range in contrast depending on input color; intensity values between 1 and 4.58 would provide very similar levels of contrast across all colors (above 4.58, some colors may be too close to both black and white to generate the requested contrast level).

To facilitate this I'm creating a color coordinate system (I'll call HSY) similar to HSL except that the third parameter will be luminosity (instead of the most colorful colors all being at L = 50%, they will be shifted up or down based on their actual luminosity). For any HSL color with saturation = 100%, the corresponding HSY color will have the same hue and saturation (for colors with saturation between 0 and 100%, there may be slight differences in the hue and saturation in the coordinate systems).

robartsd:
> Your features might be nice to have, but what we have now puts something useful into game designer hands. Any reason to wait on merging?

As long as we don't consider changing the behavior of operands that are outside the current documented range as a breaking change, I'm okay with merging.

Note I have recently changed luminance the threshold for switching the direction. This is approximately the luminance where black and white have the same contrast with the color (exact value would be `sqrt(1.05*0.05)-0.05`). This is now being applied to all intensity settings.

96LawDawg:
You obviously know a lot more about color than I ever will.  If you want to make it better, I'm on board with that.  I just thought you were further expanding into new areas.  I look forward to seeing what you come up with.

RaphaelAlvez:
keep in mind that we can always add more function later

robartsd:
Updated the pin color from player example in test room to use the new 2 operand format. I also changed the check for luminance to use approximate sRGB value rather than linear values and to leave the actual playerColor variable alone.

96LawDawg:
> Updated the pin color from player example in test room to use the new 2 operand format. I also changed the check for luminance to use approximate sRGB value rather than linear values and to leave the actual playerColor variable alone.

Looks really good.

robartsd:
> I also changed the check for luminance to use approximate sRGB value rather than linear values.

We might want to actually expose the sRGB gamma encoded value to the game designer rather than just the linear value. We could add a second parameter to calcLuminance to specify gamma function. A numerical value would raise luminance to 1/value, "sRGB" would apply the inverse of the applyGammaCorrection function.

client/js/color.js
```diff
@@ -0,0 +1,103 @@
...
+  let color = toHex(inputColor);
+  const luminance = calcLuminance(color);
+  if (luminance > 0.1791) intensity = -intensity;
+  if (intensity == 1)
+    return '#ffffff';
+  else if (intensity == -1)
+    return '#000000';
+  const r = parseInt(color.slice(1, 3), 16);
+  const g = parseInt(color.slice(3, 5), 16);
+  const b = parseInt(color.slice(5, 7), 16);
```
ArnoldSmith86, line 62: This code is in here a few times. Maybe make `toRGB` return an array instead and either create a new function that transforms it to a string or just do it in `compute.js`?

client/js/color.js
```diff
@@ -0,0 +1,103 @@
...
+  const adjustColor = (r, g, b, intensity) => [r, g, b].map(c => Math.round(Math.max(0, Math.min(c + intensity * 255, 255))));
+  const toHexString = (r, g, b) => '#' + r.toString(16).padStart(2, '0') + g.toString(16).padStart(2, '0') + b.toString(16).padStart(2, '0');
```
ArnoldSmith86, line 64: Splitting `toRGB` into two functions would simplify this line to something like `const toHexString = toHex(rgbArrayToString([r,g,b]))`.

client/js/compute.js
```diff
@@ -684,6 +684,48 @@ const compute_ops = [
     sample: 'var a = randRange ${x} ${y} ${z}',
     call: function(v, x, y, z) { return v = Math.round(Math.floor((Math.random() * (y - x) / (z || 1))) * (z || 1) + x) },
     hash: '4758cba8f404cfe3ef0a300cebf3938a'
+  },
+  {
+    name: 'colorContrast',
+    desc: 'converts color in any format that the browser can interpret to another color in the same hue but with different luminance',
+    sample: 'var a = colorContrast ${x} ${y}',
```
ArnoldSmith86, line 691: This doesn't explain which parameter does what and the format of the second parameter. Not sure how that compares to existing functions.

client/js/compute.js
```diff
@@ -684,6 +684,48 @@ const compute_ops = [
...
+    name: 'colorContrastRatio',
+    desc: 'compares 2 colors in any format that the broswer can interpret to obtain the contrast ratio in range 1-21',
+    sample: 'var a = colorContrastRatio ${x} ${y}',
+    call: function(v, x, y) { return v = calcContrast(x, y); },
+    hash: 'f0c87b733f49b7af419698cdf6ed1137'
+  },
+  {
+    name: 'colorLuminance',
+    desc: 'accepts color in any format the browser can interpret and returns the luminance value',
+    sample: 'var a = colorCLuminance ${x}',
```
ArnoldSmith86, line 719: Typo.

client/js/compute.js
```diff
@@ -684,6 +684,48 @@ const compute_ops = [
...
+    call: function(v, x, y) { return v = calcLuminance(x); },
+    hash: 'f4284956510e3fe59c2babed665b544d'
+  },
+  {
+    name: 'colorCreateHue',
+    desc: 'returns a random, as visually distinct as possible, hex color using linear interpolation',
+    sample: 'var a = colorCreateHue',
```
ArnoldSmith86, line 726: Visually distinct from what? The example doesn't use a parameter but the function seems to support one.

server/room.mjs
```diff
@@ -6,6 +6,7 @@ import FileLoader from './fileloader.mjs';
 import FileUpdater from './fileupdater.mjs';
 import Logging from './logging.mjs';
 import Config from './config.mjs';
+import { randomHue } from '../client/js/color.js';
```
ArnoldSmith86, line 9: Should we put the file into a different folder if we use it in the server _and_ the client? Or should we do something else like just adding a comment at the top of `color.js`? Or just leave it as is? I don't really mind.

ArnoldSmith86 (changes requested):

ArnoldSmith86:
I forgot to say in my review: I have no clue at all how any of that voodoo works. I just looked at the general code structure etc.

ArnoldSmith86 (changes requested):
This has the same problem as the one discussed in #1400. It overrides existing `activeColors` variables in sub-routines.

96LawDawg:
> This has the same problem as the one discussed in #1400. It overrides existing `activeColors` variables in sub-routines.

Since this is likely to be merged after that one, I will just adopt your solution there and incorporate it here before merging.  But if I do it now, it will probably create a conflict so I'll just hold off.

client/js/color.js
```diff
@@ -11,19 +13,23 @@ export function toHex(inputColor) {
...
+  return `rgb(${rgbArray[0]},${rgbArray[1]},${rgbArray[2]})`;  
 }
 
 export function calcLuminance(inputColor) {
-  // Function created by https://chat.openai.com/chat
-  let hex = toHex(inputColor);
-  let rgbString = toRGB(hex);
-  let [r, g, b] = rgbString.match(/\d+/g).map(str => parseInt(str, 10));
+  // Function partially written by https://chat.openai.com/chat
+  let [r, g, b] = toRGBArray(toHex(inputColor));
```
ArnoldSmith86, line 32: No reason for the `toHex`.

ArnoldSmith86 (approved):
Looks good. I still don't understand any of it. I applied the variable fix from https://github.com/ArnoldSmith86/virtualtabletop/pull/1451#issuecomment-1365269398. Many of your `let`s can probably be `const`s but that doesn't really matter.

ArnoldSmith86:
Does `toHex` even work on the server? I would have thought that the server can't create DOM elements.

96LawDawg:
> Does `toHex` even work on the server? I would have thought that the server can't create DOM elements.

If the test server works in the same way that production does, then it works.  I just tested again on here.  Logged in as multiple guests and got lots of different colors and not black.

ArnoldSmith86:
But the colors probably all are hex already. We just gotta make sure that we convert any player colors to hex before sending them to the server.

96LawDawg:
> But the colors probably all are hex already. We just gotta make sure that we convert any player colors to hex before sending them to the server.

I no longer understand.  The server calls the function randomHue.  That calls toHex and then processes the result to create the color.  Where could this go wrong (since it is already demonstrably working)?  I have no problem putting things back like they were before.  It would just duplicate the randomHue code on the server and the client.

ArnoldSmith86:
When generating a new color, the code on the server will call `toHex` on each player color in the room. That currently works because they are all in hex format anyway.

Now if there was a color that is not in hex format, `toHex` would go past its regex test and it would call `document.createElement` which is not a thing on the server.

I'm just saying that making sure the player colors are always in hex format when they are sent to the server, is a valid solution to the problem.

96LawDawg:
> I'm just saying that making sure the player colors are always in hex format when they are sent to the server, is a valid solution to the problem.

I am NOT going to disagree with you on a code thing.  But I don't know how to make that work.  Using the canvas is the solution that works on the client.  I don't know another way except to parse out a valid RGB or HSL to make it hex.  But your scenario is already assuming we don't have a valid hex value, so what should we do?

ArnoldSmith86:
There are a few places in the code that send the new `playerColor` to the server:

```
client/js/widgets/widget.js:      toServer('playerColor', { player: playerName, color: hexColor });
client/js/main.js:      toServer('playerColor', { player: playerName, color: urlProperties.color });
client/js/overlays/players.js:      toServer('playerColor', { player, color: e.target.value });
client/js/editmode.js:      toServer('playerColor', { player: widget.player, color: $('#seatPlayerColor').value });
```

All of those should just do `color: toHex(...)`. A malicious client could still send different colors but there's not much harm done.

96LawDawg:
> All of those should just do `color: toHex(...)`. A malicious client could still send different colors but there's not much harm done.

Updated all of those except widget.js which was already fine.  So is this PR good to go?

I don't think I'm going to wait on @robartsd's improvements.  Those won't change the functionality of this.  Calls to the functions might result in slightly different colors after his changes or he might add new functions, but that is all.

ArnoldSmith86:
Looks good. Does `toHex` work (ie not crash) for `undefined`, `null` etc?

96LawDawg:
> Looks good. Does `toHex` work (ie not crash) for `undefined`, `null` etc?

Correct.  toHex returns black if falsey or not convertible to a hex color (like "foo").  I had previously tested garbage inputs and got black.  I just texted undefined and null and it works like intended.

robartsd:
> > Looks good. Does `toHex` work (ie not crash) for `undefined`, `null` etc?
> 
> Correct. toHex returns black if falsey or not convertible to a hex color (like "foo"). I had previously tested garbage inputs and got black. I just texted undefined and null and it works like intended.

Will also return black `#000000` if the canvas element somehow does not return a hex color (Safari may do this for colors that are outside sRGB gamut).

> A malicious client could still send different colors but there's not much harm done.

We should add code in toHex() to handle this situation. Simplest solution would be to return black when a color is not already in hex format and a canvas element cannot be created.

ArnoldSmith86:
Yeah. `document` should be undefined on the server. We should just add a quick check for that.

PR #1444, opened 2022-12-13
fixes mistake inserverstate.js

RaphaelAlvez:
I made a search for some widget that are not frequently cited in the code (canvas, timer, spinner) and this seem the to the only list with a mistake


ArnoldSmith86 (approved):
Code looks fine. I hope that TestCafé at least creates all of those widget types.

PR #1446, opened 2022-12-14
First version of editable tables

rogerl50:
Still needs work. What's here mostly works, but the issue I'm having is getting clicks in other parts of the room to work correctly. The behavior I'm looking for is: Enter accepts an edit, ESC or a click anywhere else aborts the edit (and a click doesn't do whatever the click would normally do). I'm getting messed up with too many click handlers. Help is definitely needed.

### DIFF
--- a/client/css/widgets/scoreboard.css
+++ b/client/css/widgets/scoreboard.css
@@ -20,6 +20,10 @@
   width: 100%;
 }
 
+.scoreboard td.editing {
+  background-color: #f8ff88;
+}
+
 .scoreboard tbody td {
   border-bottom: 1px solid #ddd;
   border-right: 1px solid #ddd;
--- a/client/js/widgets/scoreboard.js
+++ b/client/js/widgets/scoreboard.js
@@ -312,6 +312,89 @@ class Scoreboard extends Widget {
     }
     this.domElement.style.setProperty('--firstColWidth', '50px');
     this.domElement.style.setProperty('--columns', numCols);
+
+    this.enableEditing();
   }
+
+  enableEditing() {
+    // Table editing code
+    let editCell = {
+      // Properties
+      selected : null,  // current selected cell
+      value : "", // current selected cell value
+
+      // Make cell editable and wait
+      edit : evt => {
+        const cell = evt.target;
+        if(editCell.selected && cell != editCell.selected) { // Click in some other cell, abort edit in progress
+            editCell.close(false);
+            evt.preventDefault();
+            evt.stopPropagation();
+            return;
+          }
+        // Make content editable, give it focus and colorize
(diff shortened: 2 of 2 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1446/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1446-pr-test`.

RaphaelAlvez:
Hum. I just created a room on my phone I can say I got nowhere with eh current setup.

Probably because I would assume we need a button to start anew round. If so I assume we need to make this a composite widget

96LawDawg:
I guess I agree with Raphael that this version of it should work just upon adding to the room.  So need someway to advance rounds.  We could (just to make things harder for Roger - but really to keep the visual profile low) make cell 0,0 (where it says Round), a separate click area that creates a new row.

On testing, I could enter numbers okay using the enter key.  But as soon as the scoreboard moves, the scores reset to 0.

rogerl50:
> I guess I agree with Raphael that this version of it should work just upon adding to the room.  So need someway to advance rounds.  We could (just to make things harder for Roger - but really to keep the visual profile low) make cell 0,0 (where it says Round), a separate click area that creates a new row.
> 
> On testing, I could enter numbers okay using the enter key.  But as soon as the scoreboard moves, the scores reset to 0.

Right. I should have said that I don't have this set up to update the score arrays yet. I'm just trying to get the click navigation working properly. Apologies to those of you who spent time testing something that is known not to work.

ArnoldSmith86:
So, when I said manually editing the scoreboard would be cool, I actually just had an overlay in mind that opens on click where you would have a seat dropdown, a round dropdown (with an "add" option) and a score input field.

I'm not saying I don't think your approach is cool. I do think that it'll probably be pretty finicky on mobile though.

I can definitely try to help with clickhandler problems but I'd like to hear some thoughts about this approach vs. a simpler one (opening some sort of overlay or other UI on click) before committing further.

RaphaelAlvez:
@ArnoldSmith86 do you think your approach could include a round and reset buttons?
If so I maybe prefer yours

ArnoldSmith86:
We can put whatever we want onto an overlay. We could also recreate the whole table in fullscreen and make it editable there.

rogerl50:
That sounds like a reasonable approach. If the whole table is recreated in fullscreen, with some auxiliary buttons for the additional functions, it would be much more usable even on a computer.

Should we get the original scoreboard PR merged first? I personally am ready to be done with it :smiley:

rogerl50:
Will likely be done a different way.

PR #1448, opened 2022-12-15
new Public Library test PR

RaphaelAlvez:
### Adds Carnival
Similar to Parade

(will have an update in this PR)

* I need to change so the initial draw is 5 cards.
* add auto counting
* updates to the description


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1448/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1448-pr-test`.

ArnoldSmith86:
Not a big fan of linking to BGG files. You need an account to access those.

RaphaelAlvez:
> Not a big fan of linking to BGG files. You need an account to access those.

And I'm not fan on linking a random page. Is there any middle ground you know of?

RaphaelAlvez:
@ArnoldSmith86 can this PR have the library editable? The config file did not change it

RaphaelAlvez:
> Not a big fan of linking to BGG files. You need an account to access those.

Should we use in the Parade publisher page?

https://images.zmangames.com/filer_public/57/a5/57a5d603-42ce-4487-9583-1c8b468eb639/zm1201_parade_rules.pdf

ArnoldSmith86:
> @ArnoldSmith86 can this PR have the library editable?

What do you expect to happen? It wouldn't commit the changes and I think it wouldn't even restart itself to apply the changes to its own library.

rogerl50:
I don't know how to play this game. Are the "How to Play" instructions intended to be complete instructions, because if so, I still couldn't figure out how the game is played. (This is really more of a question about what "How to Play" is supposed to do. If it's to give you an overview of the game, then I think what's there is perhaps too detailed, and uses too many terms that are not defined ("line", "unsafe", etc. If it's intended to be complete rules, it's clearly not detailed enough.)

ArnoldSmith86:
> This is really more of a question about what "How to Play" is supposed to do.

My idea was to have the complete rules there. But that obviously won't be the case for every game. Not sure if we should come up with some sort of convention or not..

RaphaelAlvez (commented):
@ArnoldSmith86 the assets in this line did not come with the file and I see no changes in github for desktop

image

RaphaelAlvez:
This is so weird. My latest commit works on local but the image did not update on  the test server. The moon should have a border like other suits

96LawDawg:
A few test observations: you already have it on your list, but either autodraw 1 card and/or make the play arrow fill to 5 and not add 5 new ones.  Too easy to overload your hand.

The pile of collected cards looks good, but when you get to a second row, the colors don't match up like in the image with the club 1 on the bottom row.

Carnival

ArnoldSmith86:
> The browser probably didn't reload the asset.

A different asset should always have a different URL so caching should _not_ be an issue.

RaphaelAlvez:
> > The browser probably didn't reload the asset.
> 
> A different asset should always have a different URL so caching should _not_ be an issue.

Yeah but I didn't load a new asset. I updated the SVG code.

ArnoldSmith86:
Your library image is 1.8MB...

96LawDawg:
Here is a compressed one for you Raphael .5MB

-2102235438_1751266-min

ArnoldSmith86:
I usually make my images 300x300. By all means make it 512x512 but 1024x1024 is overkill IMHO.

ArnoldSmith86:
This is what I would do:

library

30kB JPG

RaphaelAlvez:
> Your library image is 1.8MB...
> 
> And your variant uses the old image. Intentional? I changed it recently that the variant images will be updated if they were the same.

I totally forgot about that 😂. I'll fix it

ArnoldSmith86:
I don't think we're required to do that but I would prefer if we mark all AI generated images as such. Maybe something like `The library image was generated by Midjourney using the prompt "carnival parade, water color, hand drawing, colorful".`?

96LawDawg:
I agree.  That is the approach I took in Flight.  At a minimum, we need to say it is CC BY-NC 4.0 from Midjourney because that is the license.  Here is a sample of what I did for Flight:

The following items are available as CC BY-NC 4.0:

The bird card (feather) back face was generated using the Midjourney artificial intelligence image generator, https://www.midjourney.com/home/, using the following prompt: "bright, colorful bunch of feathers on a black background, fancy details, illustrated manuscript --ar 2:3"

Issue #1377, opened 2022-10-28
New properties for x,y coordinates of cursor


96LawDawg:
There have been a couple of recent requests to be able to obtain the x,y coordinates of the cursor using routines.  The JSON Editor has some code for displaying the x,y coords.  Could that be recycled to make new properties like `_cursorX` and `_cursorY`?  I think these would be virtual or at least not `SET`table properties.

robartsd:
Cursor coordinates should be made available as variables rather than properties if they are added.

I assume these would be room area coordinates of the cursor when the routine was triggered.

PR #1451, opened 2022-12-17
Adds `mouseCoords` variable

96LawDawg:
Implements #1377.

New variable with player mouse cursor coordinates is available as an x,y array in `mouseCoords`.

Demo in pr test room shows it working with a clickRoutine, enterRoutine, and changeRoutine.  Tested on PC and iPad.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1451/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1451-pr-test`.

ArnoldSmith86:
I like it. I would prefer to have "cursor" or "mouse" in the variable name though.

96LawDawg:
Good idea.  Changed to `mouseCoords`.  That is consistent with the feature in the JSON Editor called `jeMouseCoords`.

96LawDawg:
Simple wiki update proposal:

```
Each routine begins with a number of predefined variables:

- playerName - name of the player activating the routine
- playerColor - color of player activating the routine
- mouseCoords - [x,y] coordinates of the mouse of the player activating the routine 
- activePlayers - list of active players' names in room
- thisID - id of the widget that contains the routine
```

Note: currently playerName and playerColor say "the player clicking the widget."  That is a leftover from when everything was a clickRoutine.  So I will update that as well when this is done.

robartsd:
I believe this is providing the client coordinate of the pointer. I believe game designers might prefer the room coordinates in routines. I don't think we currently provide a way for the game designer to convert between the client coordinate and the room coordinate.

96LawDawg:
> I believe this is providing the client coordinate of the pointer. I believe game designers might prefer the room coordinates in routines. I don't think we currently provide a way for the game designer to convert between the client coordinate and the room coordinate.

Allow me to show my ignorance.  I didn't know there was, and don't understand the difference, between client and room coordinates.  My testing showed that the stuff I created matched what was being displayed in the JSON Editor.  Isn't that the room coordinate that is universal to everyone?

robartsd:
> > I believe this is providing the client coordinate of the pointer. I believe game designers might prefer the room coordinates in routines. I don't think we currently provide a way for the game designer to convert between the client coordinate and the room coordinate.
>
> Allow me to show my ignorance. I didn't know there was, and don't understand the difference, between client and room coordinates. My testing showed that the stuff I created matched what was being displayed in the JSON Editor. Isn't that the room coordinate that is universal to everyone?

You're right.

I think the eventCoords() function in mousehandling.js is doing the conversion from client coordinate to room coordinate, before you capture the value.

ArnoldSmith86 (approved):
Code looks fine. I'm not sure if I would prefer `mouseCoords.x` and `mouseCoords.y` instead of `mouseCoords.0` and `mouseCoords.1` but I don't really care. Just throwing it out there.

96LawDawg:
> Code looks fine. I'm not sure if I would prefer `mouseCoords.x` and `mouseCoords.y` instead of `mouseCoords.0` and `mouseCoords.1` but I don't really care. Just throwing it out there.

If we go that way, wouldn't it be better to use `xMouseCoords` and `yMouseCoords`?  I don't know which is better and don't really have a preference either.  They are all usable.

ArnoldSmith86 (changes requested):
This has the same problem as the one discussed in #1400. It overrides existing `mouseCoords` variables in sub-routines.

96LawDawg:
> This has the same problem as the one discussed in #1400. It overrides existing `mouseCoords` variables in sub-routines.

Will wait for the solution implemented in 1400.  Seems like that is a good approach and has buy-in from both @ArnoldSmith86 and @robartsd.

ArnoldSmith86:
This is the problem:

image

When using `CALL`, the called routine will not receive the `${mouseCoords}` as defined in the calling function. In production it does.

The proposed fix in #1400 seems like a proper fix. I added the screenshot because I'll commit the fix in a second.

ArnoldSmith86 (approved):
Re-approving. Still not sure about `x`/`y` but it doesn't really matter.

Issue #770, opened 2021-09-25
Add dice, chooser and token variants for cards


RaphaelAlvez:
Basically I propose we create those widget using the card widget as a base. This is mostly for pcio compatibility.

For dices and choosers would probably only have a special click action.

Tokens' uniqueness is still a question mark for me

96LawDawg:
This came up in a discussion of a PCIO import today.  https://discord.com/channels/770758631146782780/793626318037975081/1054980952012619776

Dmajor's game imported well, but was missing the chooser widget type.  PCIO has defaults of bidding, suit, dice, in/out. and custom choosers.  We should try to add those for compatibility.

96LawDawg:
This has mostly been done in #1871, but the PCIO importer still needs to be implemented for full compatibility.

96LawDawg:
The last remaining part of this was fixed by #1914.

PR #1456, opened 2022-12-22
New function: `CHOOSE`

96LawDawg:
This partially implements #770, although not in the way suggested there.

The idea is to create a chooser type of process like PCIO has for bidding, suits, dice, etc.  PCIO uses cards.  But cards are much harder to work with than basic widgets.  We could go 2 ways with this.  One way is to just cycle through each of the choices on a separate face.  Very easy to implement.  Just need the PCIO importer to take the values, text, images, etc. on the cards and put them on the face of a basic widget.

The other way is to do it like PCIO does when there are more than 2 choices: have a little menu of all the choices pop up.  That is what this very early draft starts to do.  The idea is to clone with widgets, put a unique face on each one, put them in a grid, then delete all the clones when the user clicks on them and set the proper face on the original widget.  This is only about 25% done, but I'm throwing it out there so you can tell me we need to do it a different way, if that is what you think.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -855,23 +855,22 @@ export class Widget extends StateManaged {
       }
 
       if(a.func == 'CHOOSE') {  
-        let numFaces;
         const faceDetails = {};
         for (const [key, value] of Object.entries(this.get('faces'))) {
           faceDetails[key] = value;
         }
-        numFaces = this.faces().length;
-        console.log(faceDetails)
+        let numFaces = this.faces().length;
         setDefaults(a, { source: 'DEFAULT', count: numFaces, xOffset: 0, yOffset: 0, properties: {}, collection: 'CHOOSER' });
+        // Somehow, need to change the collection so that it know the collection is this.id. Right now, have to put a SELECT in the routine to make it work.
         const source = getCollection(a.source);
         if(source) {
           var c=[];
           const numRows = Math.ceil(Math.sqrt(numFaces));
           const numColumns = Math.ceil(numFaces / numRows);
-          const xSpacing = 100; // The distance between each column
-          const ySpacing = 100; // The distance between each row
-          let x = -numColumns / 2 * xSpacing; // The starting x coordinate
-          let y = -numRows / 2 * ySpacing; // The starting y coordinate
+          const xSpacing = this.get('width');
+          const ySpacing = this.get('height');
+          let x = -numColumns / 2 * xSpacing + this.get('width') / 2;
+          let y = -numRows / 2 * ySpacing + this.get('height') / 2;
       
           for(const w of collections[source]) {
             for(let i=1; i<=a.count; ++i) {
@@ -885,12 +884,13 @@ export class Widget extends StateManaged {
                   problems.push(`There is already a widget with id:${a.properties.id}, generating new ID.`);
               }
               delete clone.faces;
-              delete clone.clickRoutine;
-              // clone.movable = false  ADD This back in later
+              delete clone.clickRoutine; // Need to add in a special clickRoutine to do the choosing part.
+              clone.movable = false
               clone.parent = this.id;
               clone.x = x;
               clone.y = y;
-              for (const [key, value] of Object.entries(faceDetails)) {
+ 
+              for (const [key, value] of Object.entries(faceDetails)) { //This is not working to put unique faces properties on each widget
                 clone[key] = value;
               }
               const cWidget = widgets.get(await addWidgetLocal(clone));
@@ -899,7 +899,7 @@ export class Widget extends StateManaged {
       
               x += xSpacing;
               if(x >= numColumns / 2 * xSpacing) {
-                x = -numColumns / 2 * xSpacing;
+                x = -numColumns / 2 * xSpacing + this.get('width') / 2;
                 y += ySpacing;
               }
             }
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -5,6 +5,7 @@ import { batchStart, batchEnd, widgetFilter, widgets } from '../serverstate.js';
 import { showOverlay, shuffleWidgets, sortWidgets } from '../main.js';
 import { tracingEnabled } from '../tracing.js';
 import { center, distance, overlap, getOffset, getElementTransform, getScreenTransform, getPointOnPlane, dehomogenize } from '../geometry.js';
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1456/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1456-pr-test`.

96LawDawg:
Well, I wanted to mark this as a draft, but I can't figure out how.

ArnoldSmith86:
> Just need the PCIO importer to take the values, text, images, etc. on the cards and put them on the face of a basic widget.

"Just". :smile: It'll probably be way easier to use cards.

In general, creating cloned widgets for the chooser is only necessary if all players should see it. If it should just be UI for the choosing player, it doesn't need to be part of the room state.

But, before all of that, when I wrote "If you make PCIO compatible ones, I'll add them to the importer." on Discord, I meant just building one in a room with the tools we have. My guess is that you tried and wanted to have something in the engine to make it easier.

The first question about choosers is: do we think that's an essential feature? If it is, we might as well create a new widget type for it. If it isn't - and I think that's my vote - we should make the PCIO importer take care of it.

Now that I think about it, your approach probably makes a lot of sense. Some sort of function that generates a way for players to choose a face of a widget. The problem is that it has to work with cards (because that is what PCIO does) and you would expect it to work with basic widgets, dice and maybe even spinners as well.

So yeah, something like `INPUT` which shows all faces of a widget to pick from sounds like a good idea.

ArnoldSmith86:
I don't think so. I didn't get it to do much either. But Mike said, it's only ~25% complete so I guess that's to be expected.

RaphaelAlvez:
> I don't think so. I didn't get it to do much either. But Mike said, it's only ~25% complete so I guess that's to be expected.

Yeah but there were some widgets on the test room and I expected them to do something

96LawDawg:
> But, before all of that, when I wrote "If you make PCIO compatible ones, I'll add them to the importer." on Discord, I meant just building one in a room with the tools we have. My guess is that you tried and wanted to have something in the engine to make it easier.

I KNOW you didn't expect me to code it.  I was just going to build a basic face cycle widget because at most there were 6 faces.  But then I saw the bidder chooser that has like 35 faces.  So I said to myself, "self, you would have to sacrifice your first-born child to get JSON math to do the grid work for you."  The thought of working with cards in JavaScript terrified me.  So, clones.

Is it an essential feature?  No.  But not having it limits PCIO compatibility so it is a nice-to-have feature.

Speaking of importing from PCIO, right now it imports one card for each choice.  I have/had no idea how to make that all work with a click on a single widget without building a monstrous routine.  So I did this half-assed job and said maybe others will have better ideas.

96LawDawg:
> > I don't think so. I didn't get it to do much either. But Mike said, it's only ~25% complete so I guess that's to be expected.
> 
> Yeah but there were some widgets on the test room and I expected them to do something

If you start from the saved state, the 2 widgets in there will build a grid of 2x2 of 4 and 5 widgets.  The clones get the wrong faces and they don't have any code inserted on them yet.  That's as far as I got.

ArnoldSmith86:
If we make this flexible enough, it could also be used to choose a hand card from the opponent. Put all their cards in a collection and call `CHOOSE` with a property that limits it to only show face 0 (or 1) of the widgets.

I also could have used this for the `Add new piece` button in the updated Chess. `SELECT` all unique pieces, `CHOOSE`, `CLONE`.

Any other ideas what this can be used for?

96LawDawg:
I like those ideas.  Well, if it is flexible enough, then using routines, anything could happen after the player picks a card.  As you describe, it could be sort of like a graphical `INPUT`.  You have 4 different expansion packs in a room.  Show a picture of what they look like (using card or basic widget images), they click on it and it loads that expansion pack.

Or you have a bunch of off-screen assets (special dice or whatever).  It gets them into a collection, you choose one and it rolls.

robartsd:
Here is what I think it should support:

CHOOSE:
- count: number of options that can be chosen (default: 1)
- min: minimum number to choose (default: to count)
- source: collection of widgets to choose from
- mode: "widgets"|"face" (defaults to "widgets" unless source contains only a single widget)
- faces: array of face numbers that are valid options (defaults to all)
- collection: name of collection that will contain chosen widgets
- variable: name of variable that will contain the selected active face(s)

Perhaps additional parameters to control how the options are presented.

ArnoldSmith86:
Thoughts:
- Pick card from pile (also available from pile menu)
- Labels (for example "top most", "Bob's")
- Define rows (show each player's cards as a row with label)
- Only look without choosing
- Center on screen vs center on source
- Any use cases for picking a face of multiple widgets?

This went from "weird feature for compatibility" to "solves a lot of open problems and I want this _now_" pretty quickly...

96LawDawg:
> This went from "weird feature for compatibility" to "solves a lot of open problems and I want this _now_" pretty quickly...

This went well.  I threw some shit code together, got people excited, and now they are going to do the work.  I'm quite pleased with myself.  Because really, I was in over my head on this one.  It was my only play.

ArnoldSmith86:
My thoughts exactly.

ArnoldSmith86:
Another use case: pick where to _put_ a card. Like in Exploding Kittens where you put a card back into the draw pile without looking at the fronts, without showing anyone which position you chose.

96LawDawg:
> Thoughts:

> * Any use cases for picking a face of multiple widgets?

Of all the things that have been suggested, I think this would be the most challenging to implement in the code and by a game designer.  So I've thought about it and I can't think of any use cases.  I would stay away from this to keep things manageable.  Also, this could be accomplished, if needed, by repeated calls to the function.

robartsd:
> - Only look without choosing

count: 0

robartsd:
> Another use case: pick where to _put_ a card. Like in Exploding Kittens where you put a card back into the draw pile without looking at the fronts, without showing anyone which position you chose.

For this use case, I'd probabaly just use a number INPUT with max set to the number of cards (including the one being added). An easier method to place the card in the pile in that position would be nice, but I think that is separate from CHOOSE. I think you currently would need to SET parent and z (z is the hard part).

ArnoldSmith86:
Should this be its own function or an input type for `INPUT`?

96LawDawg:
Superceded by #1871

PR #1459, opened 2022-12-24
adds pseudo player macro and changes comment notation

RaphaelAlvez:
I changed comments to use the  /* */ notation because it only requires erasing 2 lines instead of the start of all lines in the example


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1459/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1459-pr-test`.

RaphaelAlvez:
image

96LawDawg:
Good idea, but I think it is actually harder to read this way.  And it adds several additional lines which take you near the bottom of the screen.  Love the new example, but if it were me, I would put the other stuff back the way it was.

On the other hand, I can see the advantage.  If you are copying and pasting one of the examples, your way is better because you don't have to erase the slashes.

RaphaelAlvez:
> Good idea, but I think it is actually harder to read this way. And it adds several additional lines which take you near the bottom of the screen. Love the new example, but if it were me, I would put the other stuff back the way it was.
> 
> On the other hand, I can see the advantage. If you are copying and pasting one of the examples, your way is better because you don't have to erase the slashes.

from my quick experience with it it was easier to use. Actually this should be a dropdown like in the legacy editor

ArnoldSmith86 (approved):
This literally only changes comments. In a text box.

RaphaelAlvez:
@96LawDawg if you approve this I'll merge it. It seems like we just didn't do any progress in more macros over the past weeks.

New macros can come in new PRs

96LawDawg (approved):

PR #1464, opened 2022-12-26
COUNT `owner` Parameter and New Composite Seats With Hand Count

96LawDawg:
There is no issue for this.  I just decided I wanted it and did it.

Added a new parameter for COUNT `owner`.  Defaults to null or you can put a `playerName` in there.  I toyed with the idea of allowing `owner` to accept multiple `playerName`s.  But that was hard, and I couldn't think of a use case for it.  If you want multiple, do a loopRoutine over each `playerName` you want.

Part of the reason that I created the new parameter was to make it possible to have as simple as possible composite seat widget that shows the card count, like in PCIO.  But if you really don't like the new widget, I could live without it.  It is easy enough to add on your own now.  But I think the `owner` change is worthwhile on its own as I use that in several other situations.

The PCIO importer already handles this in its own way, although this might make it better.  Also, we could already do this in JSON.  The COUNT tutorial and one of the useful code snippets already show how to do this.  But this simplifies it and makes it easy for a novice to create a game showing hand count per player.  That was my real goal for this.

Demo in the pr test room.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1464/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1464-pr-test`.

ArnoldSmith86:
The transparent cards icon looks a bit weird at the seat border (especially with `turn: true`).

Also the name `perOwner` implies that you retrieve multiple values. Maybe just `owner` or `ownedBy` or whatever?

RaphaelAlvez:
PERSONALLY I don't like the layout

I would use one of these
image

- 1st: similar to canvas. advantage that the position does not need to be changed even if the user changes the seat's size
- 2nd: similar to pile that already has a cards counter. (intuitive?)

RaphaelAlvez:
if you want to keep the card simbols, which is nice, I prefer this one 
image

It took me 5 mins to make the changes I need so I definitely not a problem

96LawDawg:
I went with the card look partly because that is what the PCIO importer did.  But we don't need 3 basic widgets there and I couldn't make the VTT cards icon look any better.  So I think the circle in the left is the best.  Circle for consistency with a card count.  I would put it in the right, but I don't like the seat resizing problem.  In reality, that probably isn't a big deal, so if anybody has a strong preference, I don't really care either way.

ArnoldSmith86:
Could add a `widthUpdateRoutine` but I don't mind it on the left.

RaphaelAlvez (approved):
looks good.
code reviewed and tested

please take a look at my comment but it is not a request for change

96LawDawg:
I tried messing with the TestCafe issue.  Every time its something new.  I can do library TestCafe issues and now I think I can do compute.js issues.  But this one is a different kind altogether so I don't know what to do to fix it.

RaphaelAlvez:
Probably it is creating a game but now there is an extra widget and it thinks something is broken. It should be just a simple update with the new version

96LawDawg:
Good point.  Now that I know what it is doing, I feel safe in just updating the hash.

96LawDawg:
Now there is a small conflict from #1424.  @rogerl50 or somebody, do you think you could help resolve the conflict?  No hurry.  I'm going to be out of town the next few days and will only be checking in on my cell occasionally.

rogerl50:
> Now there is a small conflict from #1424. @rogerl50 or somebody, do you think you could help resolve the conflict? No hurry. I'm going to be out of town the next few days and will only be checking in on my cell occasionally.

I believe I resolved the conflict. I tested it somewhat on my machine; someone should test it thoroughly using the test room in this PR.

96LawDawg:
I tested it and the new COUNT multiple holders works as expected.  The COUNT owner thing works as it did before the conflict.  However, I have found a slight concern with COUNT owner that was there before and I missed.  I had intended for it to count owned cards, even if there was not a collection.  But looking at the code now I see why that couldn't possibly work.

And I don't see how that could ever work.  There will always be a DEFAULT collection.  And it can't now if that is because there was no SELECT or if there actually is supposed to be a collection using DEFAULT.  So I guess COUNT owner is working correctly in that if you are counting a selection and add in the `owner` parameter, it will count the owned cards in that selection.

Does that sound right?  If so, then this PR is working correctly.

rogerl50:
> I tested it and the new COUNT multiple holders works as expected. The COUNT owner thing works as it did before the conflict. However, I have found a slight concern with COUNT owner that was there before and I missed. I had intended for it to count owned cards, even if there was not a collection. But looking at the code now I see why that couldn't possibly work.
> 
> And I don't see how that could ever work. There will always be a DEFAULT collection. And it can't now if that is because there was no SELECT or if there actually is supposed to be a collection using DEFAULT. So I guess COUNT owner is working correctly in that if you are counting a selection and add in the `owner` parameter, it will count the owned cards in that selection.
> 
> Does that sound right? If so, then this PR is working correctly.

I don't completely understand what you are asking. But the effect of the code that is there is as follows: Accumulate all the widgets specified in `collection`, or the children of the widgets specified in `holder`. If `owner` is given, remove those widgets that are not owned by `owner`. Count what's left.

A couple of other notes:
1. It doesn't matter if the given widgets are holders or not. That was true before #1424, and it's still true.
2. Line 942, `variables[a.variable] = collections[collection].length;`, seems unnecessary.

96LawDawg:
> I don't completely understand what you are asking. 

That's because it is a dumb question.  Everything is working as expected.  I just forgot for a while that:

```
  {
      "func": "COUNT",
      "owner": "${playerName}"
   }
```

is the same as

```
  {
      "func": "COUNT",
      "owner": "${playerName}",
      "collection": "DEFAULT"
   }
```

96LawDawg:
> 2. Line 942, `variables[a.variable] = collections[collection].length;`, seems unnecessary.

Correct.  Fixed.

RaphaelAlvez:
What is this PR waiting for? I believe all changes were addressed
@rogerl50 do you have any more requests?

client/js/widgets/widget.js
```diff
@@ -922,15 +922,28 @@ export class Widget extends StateManaged {
 
       if(a.func == 'COUNT') {
         setDefaults(a, { collection: 'DEFAULT', variable: 'COUNT' });
+        setDefaults(a, { collection: 'DEFAULT', variable: 'COUNT', owner: null });
```
rogerl50, line 925: This line should be kept and the previous one removed.

client/js/editmode.js
```diff
@@ -1190,13 +1190,50 @@ onLoad(function() {
...
     const maxIndex = Math.max(...seats.map(w=>w.get('index')));
-    await addWidgetLocal({
+    const id = await addWidgetLocal({
       type: 'seat',
       index: seats.length && maxIndex ? maxIndex+1 : 1,
       x: 840,
       y: 90
+    })
+    await addWidgetLocal({
+      id: id+'C',
```
rogerl50, line 1200: So this always adds the count, and it's up to the game designer to remove it in the editor? (Not necessarily requesting a change, just making sure I understand).

rogerl50 (changes requested):

client/js/editmode.js
96LawDawg: Yes, that is the idea.  I don't know if that is best.  The only alternative I could come up with was to add a new property to seats (with a file updater of course) that had the count and let designers use that on a widget.  I decided I liked the current approach best.  But open to other ideas.

client/js/widgets/widget.js
96LawDawg: Good catch.  Fixed.

client/js/editmode.js
robartsd: Could also provide multiple options for the seat widget from the add widget screen.

96LawDawg: > Could also provide multiple options for the seat widget from the add widget screen.

Great idea.  Done.

96LawDawg:
Now there is an option to add a regular seat or a seat with a counter.  There is slight duplication of code in editmode.js lines 1187-1196 and 1199-1207 that I don't know how to remedy.  I'm okay with leaving it there, but somebody smarter with Javascript than me can probably figure out a way to avoid that.

rogerl50 (approved):

96LawDawg:
I will merge this in the next 12 hours or so.

PR #1466, opened 2022-12-26
Adds count:"all" and allows count:0

RaphaelAlvez:
Main feature is allowing to MOVE, FLIP and ROTATE 0 cards

"all" will be the new value for moving all cards instead of 0

### "all"

- [x] FLIP
- [x] MOVE
- [x] MOVEXY
- [x] ROTATE
- [x] file updater (@ArnoldSmith86 😬)

edit: simplified the scope of this PR. see comment bellow for more details.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1466/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1466-pr-test`.

RaphaelAlvez:
> ### "half"
> executes on half of each holder
> 
> * [ ]  FLIP
> * [ ]  MOVE
> * [ ]  MOVEXY
> * [ ]  ROTATE

I decided this feature is weird and useless. it would only be useful when we need to split a group of holders with unknown amounts of cards. Even in this situation, you can already do it with a FOR operation. Another problem is what to do with odd number of cards. round up or down.
anyway I don't know any game that requires such a weird slipt situation.

> ### "split"
> move all cards evenly to all holder
> 
> * [ ]  MOVE

This is would be useful but I don't want to introduce more complexity to MOVE at the moment

ArnoldSmith86:
I basically didn't test my changes at all.

ArnoldSmith86:
The line before checks `typeof routine[key].count != 'undefined'`.

96LawDawg (approved):
Approved through test.  The file updater also works for MOVEXY.  Didn't test the others, but no reason they should not work.

I updated the FLIP, MOVE, and MOVEXY tutorials to reflect the change.  The ROTATE tutorial did not need a change.

RaphaelAlvez:
> Approved through test.  The file updater also works for MOVEXY.  Didn't test the others, but no reason they should not work.
> 
> I updated the FLIP, MOVE, and MOVEXY tutorials to reflect the change.  The ROTATE tutorial did not need a change.

Actually there are 5 games in the PL that will be updated by this file updater. Maybe we should test them before merging and even tweak the JSON to remove the IF routine that comes from the update.

96LawDawg:
> Actually there are 5 games in the PL that will be updated by this file updater. Maybe we should test them before merging and even tweak the JSON to remove the IF routine that comes from the update.

Which ones?

client/js/widgets/widget.js
```diff
@@ -984,20 +984,23 @@ export class Widget extends StateManaged {
...
+        setDefaults(a, { count: 'all', face: null, faceCyle: null, collection: 'DEFAULT' });
+        if(a.count === 'all')
+          a.count = 999999;
+
         let collection;
         if(a.holder !== undefined) {
           if(this.isValidID(a.holder,problems)) {
             await w(a.holder, async holder=>{
-              for(const c of holder.children().slice(0, a.count || 999999))
+              for(const c of holder.children().slice(0, a.count))
```
rogerl50, line 995: This is sort of unrelated, but if `a.count` is not an integer, something should be pushed to `problems`. Should probably be checked above, before the `if(a.holder...`.
I think this holds for the other functions as well.

client/js/widgets/widget.js
```diff
@@ -984,20 +984,23 @@ export class Widget extends StateManaged {
       }
 
       if(a.func == 'FLIP') {
-        setDefaults(a, { count: 0, face: null, faceCyle: null, collection: 'DEFAULT' });
+        setDefaults(a, { count: 'all', face: null, faceCyle: null, collection: 'DEFAULT' });
+        if(a.count === 'all')
+          a.count = 999999;
```
rogerl50, line 989: Perhaps since 999999 is used in so many places (including calls to the logging code) it should be defined in a constant somewhere?

server/fileupdater.mjs
```diff
@@ -339,3 +340,25 @@ function v9NumericStringSort(routine) {
     if(typeof routine[key] === 'string')
       routine[key] = routine[key].replace('numericSort', 'numericStringSort');
 }
+
+function v10UpdateCountParameter(routine) {
+  for(const key in routine) {
+    if(routine[key] && [ 'FLIP', 'MOVE', 'MOVEXY', 'ROTATE' ].indexOf(routine[key].func) != -1 && typeof routine[key].count != 'undefined') {
+      if(!routine[key].count) {
+        routine[key].count = 'all';
+      } else if(typeof routine[key].count == 'string') {
```
rogerl50, line 349: I guess the purpose of this branch is to convert `''` to `'all'`? Otherwise, I'm not sure what the purpose of this branch is.

rogerl50 (changes requested):

server/fileupdater.mjs
RaphaelAlvez: Actually the purpose is to remove the default that 0 converts to 999999.

At the moment you cannot move 0 cards

This causes a problem that sometimes with automations you get a variable that gets value 0 and all of a sudden you move all cards.

So now you can either you "all" or just use a big number to move all cards but you can also "move" 0 cards

RaphaelAlvez: sorry I misunderstood the question. This if is to check if someone used a "${ }" which would be a major problem. in this case it generates an IF routine that simulates the previous behavior

client/js/widgets/widget.js
RaphaelAlvez: maybe you are right but I also don't see much advantage to changing it

client/js/widgets/widget.js
RaphaelAlvez: done

client/js/widgets/widget.js
```diff
@@ -984,20 +984,26 @@ export class Widget extends StateManaged {
...
+        
+        if(typeof(a.count)!="number")
```
rogerl50, line 991: I guess I would do something like (here and in the other functions)
```
a.count = parseInt(a.count);
if (isNaN(a.count))
  a.count = 0
  <log problem>
```
Note that as it is, non-integer numbers will be accepted.

rogerl50 (dismissed):

client/js/widgets/widget.js
RaphaelAlvez: > Note that as it is, non-integer numbers will be accepted.

But it does work. it just rounds down. I don't want to change previous behavior

I do agree I should change the count value

rogerl50: That sounds right.

RaphaelAlvez: @rogerl50 done. please check if everything is resolved

ArnoldSmith86:
Why do you enforce numbers all of a sudden?

RaphaelAlvez:
> Why do you enforce numbers all of a sudden?

It was a mistake. I just discovered that strings were valid after that.

@rogerl50 it seems that it also accepts "134" as valid input...

So should I try to convert it to a float and check if it is a number? Or just remove the whole test?

ArnoldSmith86:
> So should I try to convert it to a float and check if it is a number? Or just remove the whole test?

The behavior may not change. If you want to, you can check if there are any values that would cause the `slice` to throw an error. Everything else is fine IMHO.

ArnoldSmith86:
Please test `count: "0"`. I think it should be fine but that's a case where the behavior might change.

96LawDawg:
> Please test `count: "0"`. I think it should be fine but that's a case where the behavior might change.

On production and in this PR, count "0" moves 0 cards.  I was a little surprised that was the case on production since count "3" moves 3 cards.  So what we have in this PR should be fine.

96LawDawg:
Why are there so many unrelated changes to games in this PR?

RaphaelAlvez:
> Why are there so many unrelated changes to games in this PR?

I'm almost done. I'm going game by game updating them and removing the file updater JSON if the game designer already already made adjustments

If I don't feel confident I just keep the not updated version.

If you don't see any change that's because I removed the change that would happen and kept the original code. Everything else is from older fileUpdaters.

96LawDawg (changes requested):
I'm hesitant to approve these changes because you did so much manual work on so many PL games.  Some of it seems unnecessary.  Take Zookeeper for instance.  It doesn't look like you did anything.  It looks like the file updater would make all of those changes.  And then there are all the older file updater things getting applied so its hard to follow what's going on.

And in Telepathy, at least in the diff, there isn't even an instance of MOVE 0 anywhere.

Why is it necessary to manually edit all these games?

ArnoldSmith86:
I'm pretty sure the idea was to remove the clutter that the file updater introduces.

I have to agree though that I prefer to do that when working on the game again. None of it should be necessary or the PR is wrong.

RaphaelAlvez:
> I'm pretty sure the idea was to remove the clutter that the file updater introduces.
> 
> I have to agree though that I prefer to do that when working on the game again. None of it should be necessary or the PR is wrong.

Everything worked. But in a lot of games there was already and IF function doing exactly the same test.

Btw there were 3 games I didn't update because.

Anyway this should be an easy revert is you prefer

96LawDawg:
You probably want to go ahead and update the file version stuff to 12.  That is next in line and I don't think there are any others pending.  Also, there are TestCafe hash values that changed.  Those should have been reverted with the games, right?

ArnoldSmith86 (approved):
Alright.

- I brought this into the present. One major "conflict" was the addition of `fillTo`. I'm _pretty sure_ that `fillTo` doesn't have to be addressed in the fileupdater unless it's set to a static "truthy" value. Because if `fillTo` is "truthy", it will be used regardless of `count`. And if it is "falsy", it will be disregarded completely and it's back to the `count == 0` question.
- I updated the TestCafé hashes after making sure that they indeed only differ by the expected fileupdater changes.
- I reverted all unnecessary changes to the public library (see below) and the accidental removal of the `resetOwner` default.
- I tested the result of a fileupdater change by comparing all different values I could think of against production. This was only for `MOVE`. I also did a few tests for `FLIP` but I'm sure enough that they work the same. (Famous last words.) I was mainly afraid that the `condition: routine[key].count` worked differently than the `a.count || 999999` for some edge cases but it doesn't seem to be the case.

About the public library changes:

In theory it's a good idea to undo unnecessary fileupdater changes in the public library games. But the negatives outweigh the positives:

- Reviewing the PR is a lot harder. You suddenly get all changes the fileupdater makes in the diff and there is no way to know if they actually _were_ made by the fileupdater or by the commit author. Even worse: you _don't_ see the changes that the commit author undid - they are just "hidden" behind the change to `_meta.version`. The diff could simply show `version: 3` changed to `version: 14` (no other changes) and it could completely break the game.
- It gets a lot worse when the PR lies dormant for a while. Because the change `version: 3` to `version: 14` can now actually mean something completely different because some other PR 6 months ago was merged and received version 14. So at that point the fileupdater changes for version 14 will never be applied and the changes for this PR will be applied twice because it actually becomes version 15. In this specific PR it even jumped from 11 to 18 or something like that.
- And then the _game_ could actually have been changed in the meantime leading to code level conflicts in the JSON. That can of course be easily resolved by just using the newer game version.
- Also: the fileupdater might need fixes before merging the PR. One thing is that games might have the wrong updates then. But the unchanged games are also the best test for the fileupdater changes.

Anyway, this PR isn't the best example because we (mainly I) let it lay dormant for literally years. But I wanted to write down these points because they aren't immediately obvious and I think it's better practice to just leave the games as is and remove unnecessary fileupdater changes when someone is actually working on the game again (or just leave them in the game since the point of the fileupdater is that it shouldn't change the behavior).

96LawDawg (approved):
I tested all the functions.  I updated the tutorials.  I think the fileupdate is working as expected but no promises.  Approved.

PR #1469, opened 2022-12-27
fixed performance issues with `svgReplaces`

ArnoldSmith86:
The previous code thought "when the SVG is loaded from the server, just rebuild the entire card". Which in turn triggers the SVG waiting code for any other faces again.

With this PR, it literally only changes the backgroundImage of the card object waiting for the SVG.

This also removes the call to `mapAssetURLs` for the resulting SVG. Gotta think about that one for a second. Not sure if the SVG can even load assets.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1469/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1469-pr-test`.

robartsd:
It is possible for SVG to load external resources with URL. The SVG is loaded to css as a data URL, so CORS headers might affect if asset URLs can load in these SVGs.

robartsd:
Should a similar approach be used in basicwidgt.js? The performance hit for basic widget is not as severe as rebuilding the card DOM and the non-DOM element faces aspect of basic widget would make it more difficult.

ArnoldSmith86:
> It is possible for SVG to load external resources with URL. The SVG is loaded to css as a data URL, so CORS headers might affect if asset URLs can load in these SVGs.

https://virtualtabletop.io/p73t looks like embedding assets inside the SVG doesn't work with or without `svgReplaces` in production:

image

Opening the asset https://virtualtabletop.io/assets/1693515078_358 directly in Firefox does embed https://virtualtabletop.io/assets/-856267359_1064 into it for me:

image

Does anybody see different results?

ArnoldSmith86:
> Should a similar approach be used in basicwidgt.js? The performance hit for basic widget is not as severe as rebuilding the card DOM and the non-DOM element faces aspect of basic widget would make it more difficult.

Yeah, I would prefer to wait until problems arise. Should be far less severe.

robartsd:
> Does anybody see different results?

I see the same results. In chromium browser, I don't see any attempt to load external resources for background image SVGs.

robartsd (approved):
Code looks okay to me.

96LawDawg:
Another room that was badly broken was Assets - Jumping Dice. Recently I couldn't even get that to load on my PC. Just now it wouldn't load on iPad in production. This changes make it smooth as butter. 

Interestingly, Jumping Dice didn't used to be that bad. Neither was 5 of a Kind. Something must have changed sometime to make more svgGet calls. Thanks for the fixes.

96LawDawg (changes requested):
I was hoping to be able to approve this and get it merged, but this breaks something with "jumping dice."  Those are in the Assets file and those are used in Five of a Kind.  They just don't appear on the screen anymore.  I thought maybe this was related to the server crash and the loss of assets, so I uploaded them again.  But no dice.

ArnoldSmith86:
These?

96LawDawg (approved):
Well, they were not there before.  I swear.

Approved through test.

ArnoldSmith86:
Well, it's supposed to work every time. Can you reproduce it? It shouldn't be the storage problem either way because those assets are in the public library. Those shouldn't be affected.

96LawDawg:
> Well, it's supposed to work every time. Can you reproduce it? It shouldn't be the storage problem either way because those assets are in the public library. Those shouldn't be affected.

I can't reproduce.  I was just going to test Jumping Dice assets on my PC.  I had done it on my iPad while out of town.  When I tried to do it (same in Five of a Kind because same assets), I got nothing visible on the screen.  But as soon as you pointed it out, I went back and they are there.  I've checked on all my browsers and all my devices.  Must have been a one-off glitch on my PC.  I can't explain it and can't reproduce it.

PR #1473, opened 2022-12-30
further fixes to state migration

ArnoldSmith86:
Some games and tutorials have a different name in the metadata than in the directory name. That was always intended but the migration function looked at the wrong one. Apparently I only tested with games that had the same name.

We will have to think about this a bit because in the future this approach would mean hard-coding entries to the introduced `map` variable whenever a public library game changes its name or image but for now this should fix that some games don't get migrated properly (for example Duel 52).


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: http://212.47.248.129:1473/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1473-pr-test`.

ArnoldSmith86:
I messed up with the branch again. Replaced by #1474.

PR #1474, opened 2022-12-30
further fixes to state migration

ArnoldSmith86:
Some games and tutorials have a different name in the metadata than in the directory name. That was always intended but the migration function looked at the wrong one. Apparently I only tested with games that had the same name.

We will have to think about this a bit because in the future this approach would mean hard-coding entries to the introduced `map` variable whenever a public library game changes its name or image but for now this should fix that some games don't get migrated properly (for example Duel 52).


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: http://212.47.248.129:1474/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1474-pr-test`.

flackr (commented):
I wanted to be helpful and provide a good review but I feel I'm still not fully understanding the problem space. Feel free to ignore and just get a review from someone who already knows this stuff, but my understanding from reading this is:

AFAICT migration currently uses the filename but it should use the metadata name because filename shouldn't matter, right? So is the issue that the target state was based on the filename (because it had to be based on old code) which doesn't match the metadata name in the update so we need to make these equivalent for existing content?

96LawDawg (approved):
Approved without testing or review.

PR #1475, opened 2022-12-30
fixed state migration for duplicate games

ArnoldSmith86:
If a room had the same public library game 60x before #1223, the migration would remove the duplicates but it would count each one towards the stars.

Also: a public library game is hidden if all its variants get linked to from other games in the room. This PR changes it so it is only hidden if it is not also starred.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: http://212.47.248.129:1475/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1475-pr-test`.

RaphaelAlvez (approved):
Blind approval

Issue #12, opened 2021-01-03
De-parenting has unwanted side effects


rogerl50:
If I'm editing a widget's JSON and remove its parent attribute, or add one, then usually it disappears off the board, or at least moves to some unexpected location. This of course is because the coordinates change from absolute to relative. It would be really user-friendly if a change in parentage caused the code to revisit the coordinates, probably by assuming that the widget is at the place on the screen where the user wants it and computing new xy coordinates from the position of the holder.

UltimateGeek:
Another scenario of bad coordinates I've run into is:
1. I hand modify a widget to make it a child of another widge
2. I forget to modify the X & Y values to be relative
3. It disappears off the board
4. So I have to move the parent's position on the board to recover the child widget

rogerl50:
Yes. And sometimes you don't know where it went, and sometimes it's off the board and you have to go to Ghetto to find it. That's basically one of the two scenarios I was trying to describe in the original issue.

RaphaelAlvez:
This can be easily fixed using a macro on Ghetto. I was thinking about creating on for parenting but I may do both

robartsd:
Perhaps the built in JSON editor should _offer_ to change coordinates if parent has changed but x, y position has not.

96LawDawg:
Closed by #521.

rogerl50:
This is not the same problem. While `fixedParent` prevents de-parenting when moving a widget, this PR wants the widget to not physically move onscreen when de-parented in the editor.

96LawDawg:
Sure.  Makes sense. I was too eager to close some old issues. Actually this might be easier to solve now with _absoluteX and Y.

robartsd:
With #903 setting `x` to `_absoluteX`, `y` to `_absoluteY`, `scale` to `_absoluteScale`, and `rotation` to `_absoluteRotation`, then setting `parent` to `null` should cause the widget to not move at all except for any layering effect changes.

robartsd:
One difficulty with fixing this is that the JSON editor tries to update live as you type (which generally we want to keep), so there is not really an way to determine if x and/or y have been changed without a fairly complicated system for tracking edits.

We could add a Change Parent button. I think the button should give the option to update x,y so that the widget's relative position in the room does not change (default option if widget parent get changed to or from null, but defaulting to keeping the widget x, y unchanged if changing from one parent to another).

96LawDawg:
[...]

PR #1482, opened 2022-12-31
initial version of new editor

ArnoldSmith86:
https://test.virtualtabletop.io/PR-1482/pr-test

The new editor. This should replace the old edit mode (with the blue outlines) and incorporate the JSON editor.

See #1627 for more discussion about it.

I want to add a lot of awesome stuff to the editor but we should try to keep this PR as small as possible. It probably shouldn't take too many features away that we currently have.

This PR is also splitting the code so that a player never has to load the editor code. Whenever you click on edit mode or press `CTRL+J`, it should load everything for editing from the server. The code for the JSON editor and edit mode is already about half of the code base and I want to add a lot more editor stuff.

Goals before merging:
- [x] Move the Add Widget button from the JSON editor module to the main toolbar. This should become its own module with a lot more stuff but let's use what we have for now.
- [ ] Improve the Properties module at least a bit so it's actually useful. More complex stuff can wait for later.
- [x] Highlighting the selected widget needs to be improved.
- [x] Increase the minimum size of the modules.
- [x] Improve the media queries so every viewport size gets a sensible layout.
- [ ] The yellow in the light debug is too bright.
- [ ] Undoing stuff sometimes creates empty piles. I think the problem is that widgets are created locally before sending its delta. Creating the undo and redo deltas needs to recognize that. (I also had a dropShadow widget staying behind after playing with carrots in the hand of `Animal Race` and undoing.)
- [x] Fix (or remove) Jest tests.
- [ ] Test tracing.
- [ ] Scrolling on mobile doesn't work if started on a widget button.
- [ ] Rearranging models using drag-and-drop fails on mobile.
- [ ] Touch doesn't work on PC.
- [ ] JSON editor auto-focusing is annoying on mobile.
- [ ] Remove Toggle wide editor button from old JSON buttons (it does nothing in the new UI).


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1482/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1482-pr-test`.

robartsd:
Looks like you put tracingEnabled in the edit module, but it is still being called from the main code and causing the build test to fail.

I think it would be appropriate to keep tracing available without loading edit code.

ArnoldSmith86:
I'll have to fix (or get rid of) Jest tests but hopefully this is ready for review otherwise.

The gist is: everything should be the same except that clicking `Edit Mode`, `Add Widget` or pressing `CTRL+J` should take half a second longer because that's when all the edit code is loaded.

This actually loads the edit code into its own module which means that every piece of code that uses something in the other module (so play -> edit or edit -> play) needs to be "allowed". This probably leads to some stuff still not working. Please test if you can find something.

Maybe I can get the diff a bit smaller. I went a bit back and forth with my approach about how to split this stuff so maybe I can undo some stuff.

---

The function `loadEditMode` in `main.js` adds all the functions that are called from edit mode to the `window` object, making them global to all modules. Not the cleanest solution but overall I kinda like it.

Everything using `export` in the edit mode module is also added to the `window` object in the same function for the interface in the other direction.

---

Now I want to base the next PR on this that will try to define the global layout of the new edit mode. With a bit of luck, the interface between the two modules will improve a little in the process. No promises.

In the process I will integrate the JSON editor as the first "module" (or whatever it will be called) to the new edit mode. I'll probably ignore the current edit mode completely and instead create a second "module" that creates UI elements for all widget properties.

Further modules will include component view, assets view, symmetry editor, undo protocol, grid / layout tools, etc. No order or deadlines planned but we can just add one at a time.

---

Just a little glimpse into the future as motivation to get this PR rolling. I'll just base the next PR on this one though so there shouldn't be a delay.

ArnoldSmith86:
> Looks like you put tracingEnabled in the edit module, but it is still being called from the main code and causing the build test to fail.
> 
> I think it would be appropriate to keep tracing available without loading edit code.

Oh right, I forgot about fixing tracing...

ArnoldSmith86:
> I think it would be appropriate to keep tracing available without loading edit code.

Yeah, the tracing code is tiny if the _trace viewer_ remains in edit code.

robartsd:
> > I think it would be appropriate to keep tracing available without loading edit code.
> 
> Yeah, the tracing code is tiny if the _trace viewer_ remains in edit code.

Is the tracing just the JSON editor debug, I thought there was a server tracing component. Is there more client side tracing that can be viewed?

ArnoldSmith86:
The tracing sends events to the server which in turn writes them into a trace file. That trace file can be viewed in the client by pressing `CTRL+F9`. Then you can see the room state at any time of the trace and look at it in the JSON editor.

RaphaelAlvez:
looks to me you are not using this in your new PR should we close it?

ArnoldSmith86:
> looks to me you are not using this in your new PR should we close it?

The other PR is based on this.

96LawDawg:
There are many awesome things about this I feel a little guilty asking for anything.  But I will.

First, it looks like you've solved the issue of being able to edit stuff and click stuff.  It looks like you click once to select the widget and the second click activates whatever click normally does.  Is that right?  That is a good compromise.  I like it.

Second, this may be in your to do list somewhere, but it would be nice to fix #12 and the reverse (when adding a parent).  Because it is so easy to move widgets (and they are not de-parented by just moving them), when adding, deleting, or changing a parent, it would be great for the widget to stay in its current (absolute) room coordinates.

Other than that, I built a simple room in a few seconds that previously would have taken much longer.  I can't wait to see what other magic you come up with.

ArnoldSmith86:
> Second, this may be in your to do list somewhere, but it would be nice to fix #12 and the reverse (when adding a parent). Because it is so easy to move widgets (and they are not de-parented by just moving them), when adding, deleting, or changing a parent, it would be great for the widget to stay in its current (absolute) room coordinates.

Yeah, a solution for that will be added in the Properties panel where you select the parent. I have no idea yet how that will look like though. :grin:

ArnoldSmith86:
Oops, my compact toolbar breaks when the material icons aren't loaded yet.

mousewax:
I just added `font-display: block;` to the Material Icons set of fonts to give them a little extra time to load before displaying the fallback font. (#1683)

ArnoldSmith86:
> I just added `font-display: block;` to the Material Icons set of fonts to give them a little extra time to load before displaying the fallback font. (#1683)

We should probably just clip the text before loading the font. I don't think it's helping anyone as a fallback.

robartsd:
> > I just added `font-display: block;` to the Material Icons set of fonts to give them a little extra time to load before displaying the fallback font. (#1683)
> 
> We should probably just clip the text before loading the font. I don't think it's helping anyone as a fallback.

`font-display: block;` should prevent the fallback for the maximum time the client allows.

RaphaelAlvez:
Idea for grid Input:
I would love to be able to see where the widget can be placed. An idea that came to mind is to generate an overlay that shows the grid limits and the possible widgets positions.
The actuall input could be like the card layers input showed in last meeting

ArnoldSmith86:
> Idea for grid Input: I would love to be able to see where the widget can be placed. An idea that came to mind is to generate an overlay that shows the grid limits and the possible widgets positions. The actuall input could be like the card layers input showed in last meeting

Yeah, that's something I want to do. And I'd like to provide some easy presets like hexagonal stuff.

RaphaelAlvez:
> > Idea for grid Input: I would love to be able to see where the widget can be placed. An idea that came to mind is to generate an overlay that shows the grid limits and the possible widgets positions. The actuall input could be like the card layers input showed in last meeting
> 
> Yeah, that's something I want to do. And I'd like to provide some easy presets like hexagonal stuff.

I have the math for triangular grid written down somewhere.

I also got the grid but there are 2 possible tilling and rotation is a mess. It's specially terrible if you want it to match hexagonal and triangular grids. it is just too buggy

RaphaelAlvez:
TImer really mess the history

RaphaelAlvez:
the hovering tool bar should include resizing.

also it would be nice to have a shit key to keep allignment of widgets

RaphaelAlvez:
- TImer really mess up the history
- The hovering tool bar should include resizing
- It would be nice to have a shift-key (or similar) to keep allignment of widgets when using spacing and moving
- No ungrouping
- We should have 3 edit mode: Play, Edit and Select. And there need to be better visual feedback on wich mode we are using. It is too easy to mess things up and too hard to test stuff.
- Add widget overlay should be the first to close. Not the sidebar

RaphaelAlvez:
card back preview in the deck builder is misleading. For me it looked like solid grey and then it was a white and black grid

96LawDawg:
Bug report:  When you have a widget selected and the widget's draggable toolbar is available, if you change the widget's `id` the toolbar gets de-synched from the widget.  Refreshing the screen fixes it.

96LawDawg:
Feature request: The "Start with a new blank room" button needs some sort of confirmation.  Too easy to lose work.

RaphaelAlvez:
> Feature request: The "Start with a new blank room" button needs some sort of confirmation.  Too easy to lose work.

Use undo before unloading your room

96LawDawg:
Bug report:  Using the var expression search text area does not work like it did/should.

96LawDawg:
> > Feature request: The "Start with a new blank room" button needs some sort of confirmation.  Too easy to lose work.
> 
> Use undo before unloading your room

Did you test that?  It does not work for me.

RaphaelAlvez:
> > > Feature request: The "Start with a new blank room" button needs some sort of confirmation.  Too easy to lose work.
> > 
> > 
> > Use undo before unloading your room
> 
> Did you test that? It does not work for me.

I did.  it is stated as "full room state" or something like this. just select the previous modification.

I do not disagree that it should be that easy because people will not know how to deal with it

Ingegneus:
I am going to dump all of my observations so far, in no particular order:
many tasks take way more clicks than before. for example selecting a widget from the list and then switching back to the editor is a bit annoying. I used this a lot in the previous editor, to quickly format the json.
also when debugging I would really like to see the F1-F12 widget bar without having to go to the editor to check the widgets. 
speaking of the widget hover sidebar: it should definitely be larger, because often times the enumeration of the widgets is at the end of the name and is therefore not visible. or at least the name should wrap around. 
in the widget list pane, the filter/search bar should be fixed at the top, so that I don't have to scroll aaaaaalll the way up in order to search for a widget. also I think a search bar for the debugger might be useful as well.
I think the buttons on top should be right justified to reduce the mouse travel, as well as having them more in your field of view , so you don't forget about them. also the shortcuts are only of limited use to me, because most of the time the cursor is in the json editor.
I know it is still work in progress, but I can't wait for the save button! that is going to be huge!
I really really like the undo history, even though at times I would prefer a reliable ctrl z. for some reason that still isn't working a lot of the times.
I think the overlay next to the widgets (the new one with copy to grid and all that) should be smaller and also stay on screen when moving it around.
I really really miss seeing the mouse coordinates.

ArnoldSmith86:
> I really really like the undo history, even though at times I would prefer a reliable ctrl z. for some reason that still isn't working a lot of the times.

Please tell me when it isn't working.

96LawDawg:
Bug report: When you click on a holder and then click on Properties toolbar, you see the "target widgets" with pictures of various decks.  However, when you click away and come back to that view, the "target widgets" area is empty.

Ingegneus:
> Please tell me when it isn't working.
it's really hard to tell when it happens because it is so random.

96LawDawg (approved):
Approved.

PR #1486, opened 2023-01-01
send meta update when setting state

ArnoldSmith86:
When setting the room state (for example by using the `/state/` API), the `activeState` of the room is reset. That is saved in the room metadata but the function did not send a metadata update to the clients. This in turn did not hide the "Save game" button which in turn crashed because there was no `activeState` to save.

In an ideal world, the server should check for `activeState` to exist but since the exception is handled anyway, I do not regard this as a big problem.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1486/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1486-pr-test`.

96LawDawg (approved):
Blind approval.  I have no idea how to test or verify.

robartsd (approved):

PR #1489, opened 2023-01-01
new game Fun with Trains

ArnoldSmith86:
Similar to Ticket to Ride. There are still many things I am not happy with but it is playable. Feel free to help improving it.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1489/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1489-pr-test`.

96LawDawg:
The only playability issues I noticed:

It is possible for a player to drop their colored track onto another player's seat area.  Use dropTarget to fix.

And the green check and red x need to be more clear when they become a valid drop target.  Too easy to drop a goal card into the wrong place.

ArnoldSmith86:
> And the green check and red x need to be more clear when they become a valid drop target. Too easy to drop a goal card into the wrong place.

Yeah, that whole dynamic has to be improved. During the game, you only want to discard those cards without changing the score. After the game you either want positive or negative points.

RaphaelAlvez:
why is this game so dark??

ArnoldSmith86:
Why not? I have a light version. I can add it to the pr-test room.

RaphaelAlvez:
> Why not? I have a light version. I can add it to the pr-test room.

there is very low contrast. That's the only reason. if you make the colors more vibrant and the greys lighter then this should be good

96LawDawg:
Replaced by #2606

PR #1490, opened 2023-01-01
updated The Team with seats, missions and better task management

ArnoldSmith86:
So far this only updates the 4 player variant.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1490/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1490-pr-test`.

96LawDawg (approved):
Approved through test.

The only thing I don't like, and you don't have to change this, is that the cards get re-dealt every time someone new sits.  So at the beginning of the game, I'm like, yeah, I have 3 rockets.  And then player 2 joins and I get all new cards.  Suggestion: make it so the deal only happens when you press the deal button.

96LawDawg:
I didn't notice that you had this marked as a draft.  But once you are satisfied with the 4P version, you can go ahead and get it approved and added to the library if you want.  You don't have to wait until they are all done.

PR #1501, opened 2023-01-03
Implement dropShadow property which shows where widget will drop

flackr:
https://test.virtualtabletop.io/PR-1501/pr-test

While dragging a widget with the dropShadow property set, a clone of the widget is added to the current hover target to show where that widget will be dropped to visually.

Proposed wiki update:

| Property | Default value | Description  | 
| dropShadow | false | If true, when a widget that matches the _dropTarget_ is dragged over the holder, a gray shadow will appear where the widget will be when dropped. The shadow is a cloned widget with a property `dropShadowOwner`. When a shadow is created, the value will be the `playerName`.  The clone will activate enterRoutine and leaveRoutine, so you may need to counteract that by using IF statements filtering out the clone looking for non-null values. The css of the shadow can be modified using `.dragging-shadow`. |


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1501/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1501-pr-test`.

96LawDawg:
Doesn't work when dragging a pile handle.

96LawDawg:
A couple of other things
 - Setting dropShadow: false on the holder seems like the more natural way to set the property, rather than on the widget.  But maybe that is just me.
 - When you have a pile in a holder where the dropShadow will be, the pile number increments as soon as the dropShadow appears (because it is a widget getting counted in the pile).  That is a minor change from current behavior where the pile number doesn't increment until the widget is dropped.  Not sure this is a problem because it still works correctly with dropLimit.  I thought it might not, but it does.  So just pointing it out.
 - I think the shadow should be a lighter gray.
 - Can you explain how to change the css on the shadow?  I tried something like this on the widget and it didn't work:
 
```
 "css": {
    ".shadow": {
      "filter": null,
      "background": "red"
    }
  },
```

96LawDawg:
I added this to the cardDefaults thinking it might work:

```
"onPileCreation": {
      "dropShadow": true
    },
```

But that just makes the pile drag handle the target instead of the holder.  So the pile drag handle turns gray.

flackr:
> I added this to the cardDefaults thinking it might work:
> 
> ```
> "onPileCreation": {
>       "dropShadow": true
>     },
> ```
> 
> But that just makes the pile drag handle the target instead of the holder. So the pile drag handle turns gray.

Yeah, this is because it currently does a shallow clone, so only the dragged widget (i.e. pile handle) is shown. This is an easy fix - and I can implement that other feature I've wanted - the recursive clone option.

flackr:
> * Setting dropShadow: false on the holder seems like the more natural way to set the property, rather than on the widget.  But maybe that is just me.

Yeah, I could see an argument for making this a property of the holder rather than the widget. It's slightly harder technically since I'll need to do an ancestor check but not that hard. I'd be curious to hear what others think.

> * When you have a pile in a holder where the dropShadow will be, the pile number increments as soon as the dropShadow appears (because it is a widget getting counted in the pile).  That is a minor change from current behavior where the pile number doesn't increment until the widget is dropped.  Not sure this is a problem because it still works correctly with dropLimit.  I thought it might not, but it does.  So just pointing it out.

Also something we could "fix" if we wanted - avoid counting the shadow widget in the pile count.

> * I think the shadow should be a lighter gray.
> * Can you explain how to change the css on the shadow?  I tried something like this on the widget and it didn't work:
> 
> ```
>  "css": {
>     ".shadow": {
>       "filter": null,
>       "background": "red"
>     }
>   },
> ```

I was able to style it via the global css, see the widget id `css` at https://test.virtualtabletop.io/PR-1501/shadow-style

Note that to disable the filter you have to set filter to `'none'` or `'revert'` to undo the filter being applied by the css stylesheet.

flackr:
@96LawDawg I made a few changes around some of the things you brought up. Dragging a pile will now clone the top-most child in the pile rather than the pile handle. Also piles don't count the shadow widget anymore.

The clone code now also supports recursion and is shared with the clone function. Let me know what you think :-).

ArnoldSmith86:
Do you know about the existing clone function that's also used by the JSON editor command?

I didn't look at yours in detail yet but at a glance it seems like duplicated functionality.

robartsd:
Do the cloned widgets get synced to state (all players see); if so, what happens with the clones if a client get disconnected in the middle of a dragging operation?

flackr:
> Do the shadow widgets get synced to state (all players see);

Yes, I think this is the best way to make sure it behaves the same as dropping the widget and ensure that everything stays in sync.

> if so, what happens with the clones if a client get disconnected in the middle of a dragging operation?

It stays. I'll look into how to detect drags still in process vs disconnected users and clean it up appropriately.

96LawDawg:
> Also piles don't count the shadow widget anymore.

Well, if you didn't feel fully welcomed to VTT yet, now you are.  Because piles are hard.  And your improvement is great and works as expected when a pile already exists.  BUT, it doesn't work when first making a pile.  Demo: Put 1 card into a holder that does piles.  Then drag and hover over that holder.  You get a pile indicator of 1 before there is a pile.

flackr:
> Well, if you didn't feel fully welcomed to VTT yet, now you are. Because piles are hard. And your improvement is great and works as expected when a pile already exists. BUT, it doesn't work when first making a pile. Demo: Put 1 card into a holder that does piles. Then drag and hover over that holder. You get a pile indicator of 1 before there is a pile.

Haha thanks!

I think what I really want is a property similar to `isBeingRemoved` which prevents piles from being formed - but I'm not sure if I can safely re-use that property. As a short term workaround I could probably set `onPileCreation` to something unique to prevent the shadow from being included in any piles.

flackr:
And done! That was pretty simple and builds on the existing pile code rather than needing to modify it 😀

robartsd:
> It stays. I'll look into how to detect drags still in process vs disconnected users and clean it up appropriately.

We probably want the shadow widget to have a property that indicates which widget it is a shadow of. Not sure how it should be cleaned up. IIRC we have some code in timer.js that has another client adopt a running timer if the client that had been running the timer disconnects; if so, something similar should happen here.

flackr:
> > It stays. I'll look into how to detect drags still in process vs disconnected users and clean it up appropriately.
> 
> We probably want the shadow widget to have a property that indicates which widget it is a shadow of. Not sure how it should be cleaned up. IIRC we have some code in timer.js that has another client adopt a running timer if the client that had been running the timer disconnects; if so, something similar should happen here.

Another option is we could just clean it up the next time someone drags the widget, since that would also be the point at which we clean up the dragging property right?

robartsd:
> Another option is we could just clean it up the next time someone drags the widget, since that would also be the point at which we clean up the dragging property right?

That's true. As long as the client that starts the new drag finds the old shadow copies rather than creating new ones, it should work fine.

flackr:
> > Another option is we could just clean it up the next time someone drags the widget, since that would also be the point at which we clean up the dragging property right?
> 
> That's true. As long as the client that starts the new drag finds the old shadow copies rather than creating new ones, it should work fine.

Done in latest patch.

> I'm still not sure that I think the shadow elements should be widgets, but I understand that it makes some things easier to process them as widgets. A couple of years ago, we started down the process of making piles no longer be widgets, but eventually decided it wasn't worth it.

I know it's tempting not to create widgets for this, and I really tried to go this route initialy but there were going to be many issues that would be really hard to solve properly if it's not a widget. The state would be inconsistent if the widget isn't actually added to the holder during the drag. I.e. My client needs to produce a layout with the widget in the holder and if I have a different layout than you then it can lead to further inconsistencies. Not to mention there are many setups where the widget will run a routine to position new elements.

I initially tried putting the actual widget in the holder then it will break things like transitions #1423 and we'd still need to sync a drag copy of the widget for other people to see me dragging it anyways.

I think it's kind of nice how it works that you can see the same state and we can both position drag elements within holders in parallel.

robartsd:
> and we'd still need to sync a drag copy of the widget for other people to see me dragging it anyways.

All that is really needed for sync is for other clients to know that the widget is being dragged (which they already do) - each client could create the shadow element for widgets being dragged independently. Still I understand that using an actual widget makes a whole bunch of things about the shadow element easier to accomplish.

flackr:
> All that is really needed for sync is for other clients to know that the widget is being dragged (which they already do) - each client could create the shadow element for widgets being dragged independently. Still I understand that using an actual widget makes a whole bunch of things about the shadow element easier to accomplish.

That works for the simple cases, but if placing the shadow widget in the holder invokes routines you would invoke those routines on each client which would could result in strange unexpected results.

My intention with just having a simple property is that we could still explore other ways of accomplishing the shadow while still supporting the high level developer intent that we should have a shadow element showing where the widget will go.

client/css/widgets/basicwidget.css
```diff
@@ -9,3 +9,7 @@ body.edit .widget.basic.hasImage {
   opacity: 0.45;
 }
 
+.shadow {
+  filter: brightness(0);
+  opacity: 0.3;
+}
```
robartsd, line 15: I think `pointer-events: none` would be useful. Also I think the class name should be a bit more specific; perhaps `dragging-shadow`

flackr: Sounds good, done.

robartsd:
> That works for the simple cases, but if placing the shadow widget in the holder invokes routines you would invoke those routines on each client which would could result in strange unexpected results (e.g. if you had a counter it would be invoked by each player in the session).

If the shadow element were not a widget, it wouldn't trigger any routines. One concern I have is that by using shadow widgets, routines may be triggered that the game designer does not intend to be triggered until the widget is dropped.

If dropShadow is a holder property instead of a widget property (would not support a shadow outside holders, but I don't think that would change the visual effect unless the widget has partial transparency) the holder could create the shadow element as part of its own rendering. Not sure how that delta call should work though.

96LawDawg:
> If the shadow element were not a widget, it wouldn't trigger any routines. One concern I have is that by using shadow widgets, routines may be triggered that the game designer does not intend to be triggered until the widget is dropped.

That is a good point.  I just tested that.  Not only does the shadow fire an `enterRoutine` it fires it seemingly fires it for every x,y change of the dragged widget.  So it went off a few dozen times just during a routine drop.

flackr:
> > That works for the simple cases, but if placing the shadow widget in the holder invokes routines you would invoke those routines on each client which would could result in strange unexpected results (e.g. if you had a counter it would be invoked by each player in the session).
> 
> If the shadow element were not a widget, it wouldn't trigger any routines. One concern I have is that by using shadow widgets, routines may be triggered that the game designer does not intend to be triggered until the widget is dropped.

I think this is an area where we disagree about the expected behavior. I have written and seen many developers write placement logic in routines, which requires that we invoke those routines in order to determine the position. E.g. consider the [automatic stackOffset from the code snippets](https://github.com/ArnoldSmith86/virtualtabletop/wiki/Useful-code-snippets#auto-adjust-stackoffsetx-1). I've added it as an example in the pr-test room at the top-right.

As long as developers opt in to the behavior this should be okay and expected that it invokes routines.

flackr:
> > If the shadow element were not a widget, it wouldn't trigger any routines. One concern I have is that by using shadow widgets, routines may be triggered that the game designer does not intend to be triggered until the widget is dropped.
> 
> That is a good point. I just tested that. Not only does the shadow fire an `enterRoutine` it seemingly fires it for every x,y change of the dragged widget. So it went off a few dozen times just during a routine drop.

Yes, this is expected, and required in order for things like automatic stackOffsetX calculations from routines to work.

96LawDawg:
> > That is a good point. I just tested that. Not only does the shadow fire an `enterRoutine` it seemingly fires it for every x,y change of the dragged widget. So it went off a few dozen times just during a routine drop.
> 
> Yes, this is expected, and required in order for things like automatic stackOffsetX calculations from routines to work.

Then I don't think this solution will work.  You can't fire enterRoutines before a card is dropped.  That would break many existing games.

flackr:
> Then I don't think this solution will work. You can't fire enterRoutines before a card is dropped. That would break many existing games.

But developers opt in to doing this, so it will not break any existing games unless they opt in to it.

flackr:
This might be good reason to make it a property on the holder, so that it is expected to fire routines on that holder for the shadow.

96LawDawg:
> But developers opt in to doing this, so it will not break any existing games unless they opt in to it.

True.  What I meant to say was that developers could not add this effect to existing games without breaking them.  Maybe there is a solution in JSON, but it would be a little messy.  The enterRoutine could look for the characteristics to see if the object is a cloned shadow and ignore it if it is, otherwise do the routine.

flackr:
I moved this to be a property on the holder and only create the shadow widget when the current target holder has this property set. This makes it much more obvious which holders will receive additional routine calls for shadow widgets. @96LawDawg I think that was actually the last thing I hadn't yet changed from what you called out initially :-).

96LawDawg:
Trying to drag a pile into the holder freezes the client.

RaphaelAlvez:
I keep getting shadows stuck

flackr:
Yeah, my bad, the code should check whether it already has a shadow widget before creating one.

96LawDawg:
> I keep getting shadows stuck

I was about to say the same thing.  I thought maybe it was because my client froze while you were doing that.

flackr:
@RaphaelAlvez Should be better now though I'll probably try to do something a bit smarter to make sure that there is only ever one shadow widget per real widget.

RaphaelAlvez:
> @RaphaelAlvez Should be better now though I'll probably try to do something a bit smarter to make sure that there is only ever one shadow widget per real widget.

Much better!!

96LawDawg:
And the pile thing is fixed, also.  This still triggers the enterRoutine a lot.  I have a high end system and I can hear my fan ramp up as I drag a card back and forth across the holder.  Here's a CPU usage screenshot.  That spike in the middle is where I started dragging and the drop off at the end is where I stopped.

CPU

flackr:
> This still triggers the enterRoutine a lot.

@96LawDawg Yeah so technically we shouldn't need to trigger enterRoutine repeatedly as the card technically only needs to enter once. It seems that we don't currently support for automatic positioning if the widget isn't removed and readded to the holder. E.g. see this demo where i set `fixedParent: true` on some cards in a holder with stackOffsetX: https://virtualtabletop.io/y1hm

I updated the shadow code to directly call `receiveCard` with the position if the target hasn't changed so this should now only call enterRoutine once per holder you drag over.

flackr:
Directly calling receiveCard has broken it for the stackOffset container though, will need to dig into why - it should work.

flackr:
> Directly calling receiveCard has broken it for the stackOffset updating enterRoutine container though, will need to dig into why - it should work.

Fixed.

flackr:
While we're on the topic of this general behavior, there's one other drag behavior game designers may want - and that would be to just actively put the dragged widget in the target holders so that the dragged item itself snaps to the location rather than a shadow of it. We could change the property from `dropShadow: true` to some sort of multi-value property like `dropPreview: none | shadow | widget` where `none` is the current (default) behavior, shadow is the one currently implemented in this PR and `widget` would immediately put the dragged element in the holder (optionally with transitions).

flackr:
Looking at other properties, other than type I can't find many existing cases where we have multiple special string values so I think keeping this as a boolean dropShadow for now is fine. We could add `dropPreview: true` later if desired to do the other behavior I suggested might be interesting, and then ignore `dropShadow: true` if dropPreview is true.

flackr:
> Do you know about the existing clone function that's also used by the JSON editor command?
> 
> I didn't look at yours in detail yet but at a glance it seems like duplicated functionality.

This currently shares it's code completely with the `CLONE` function. I think that all three (this, the clone function and the edit code) could likely use the same base code. Is it okay if I do this in a followup? With the current code sharing there's no duplicated functionality.

96LawDawg:
I was going to update your branch with an updated Holders: Appearance (variant 0) tutorial.  But since you are not working on the main repository, I can't do that.  So here it is (change .zip extension to .vtt).  This addresses Raphael's request for something that implements the change.

[Holders.zip](https://github.com/ArnoldSmith86/virtualtabletop/files/10356073/Holders.zip)

RaphaelAlvez:
> I was going to update your branch with an updated Holders: Appearance (variant 0) tutorial.  But since you are not working on the main repository, I can't do that.  So here it is (change .zip extension to .vtt).  This addresses Raphael's request for something that implements the change.
> 
> [Holders.zip](https://github.com/ArnoldSmith86/virtualtabletop/files/10356073/Holders.zip)
> 

My request was for the other PR 😆

I won't complain that we a tutorial when this is merged

96LawDawg:
`dropShadowClone` should be boolean.  The clone already has a property `clonedFrom` that includes the id of the widget it is from.  To prevent the clone from being the cause of enter or leave routines, we need something like this to work:

```
"leaveRoutine": [
    {
      "func": "GET",
      "collection": "child",
      "property": "dropShadowClone",
      "variable": "shadowClone"
    },
    {
      "func": "IF",
      "condition": "${shadowClone}",
      "elseRoutine": [
        "var doStuffHere = ${coolStuff}"
      ]
    }
  ]
```

So line 429 of widget.js should be:

```
'dropShadowClone': true,
```

flackr:
> I was going to update your branch with an updated Holders: Appearance (variant 0) tutorial. But since you are not working on the main repository, I can't do that.

You should be able to update it - I always create my PR's with the maintainers can modify branch option checked. In the future I'll create branches direction on virtualtabletop for PR's.

> So here it is (change .zip extension to .vtt). This addresses Raphael's request for something that implements the change.
> 
> [Holders.zip](https://github.com/ArnoldSmith86/virtualtabletop/files/10356073/Holders.zip)

Thanks! I added it to the PR and tested out that the tutorial shows it.

96LawDawg:
> You should be able to update it - I always create my PR's with the maintainers can modify branch option checked. In the future I'll create branches direction on virtualtabletop for PR's.

Maybe I could.  But I couldn't find that branch in GitHub desktop for Windows.

flackr:
> `dropShadowClone` should be boolean. The clone already has a property `clonedFrom` that includes the id of the widget it is from. To prevent the clone from being the cause of enter or leave routines, we need something like this to work:
> 
> ```
> "leaveRoutine": [
>     {
>       "func": "GET",
>       "collection": "child",
>       "property": "dropShadowClone",
>       "variable": "shadowClone"
>     },
>     {
>       "func": "IF",
>       "condition": "${shadowClone}",
>       "elseRoutine": [
>         "var doStuffHere = ${coolStuff}"
>       ]
>     }
>   ]
> ```
> 
> So line 429 of widget.js should be:
> 
> ```
> 'dropShadowClone': true,
> ```

Sure, done! At the same time I also added it to the default property list since we usually have defaults for the built-in properties (e.g. dragging), though notably `clonedFrom` does not have a default value specified.

flackr:
I would have expected your condition to work anyways since a non-empty string is "truthy" in javascript, but it makes sense not to duplicate clonedFrom.

96LawDawg:
> I would have expected your condition to work anyways since a non-empty string is "truthy" in javascript, but it makes sense not to duplicate clonedFrom.

But that's the point.  I don't want truthy.  I want === true.  That's the easiest way to tell this is the shadow and to ignore it.  Everything works as expected now.  Thanks.

96LawDawg (approved):
Approved through test

96LawDawg:
Added proposed wiki entry update in the initial comment.

Ingegneus:
I haven't read all the comments on this one, so this might have already been sugested. I'd like to see the drop shadow, also when there are already cards in a holder (see holders 9yy and g1rq in the pr test room). curently the drop shadow card seems to disppear below existing cards, but I think those should be on top.

96LawDawg:
> I haven't read all the comments on this one, so this might have already been sugested. I'd like to see the drop shadow, also when there are already cards in a holder (see holders 9yy and g1rq in the pr test room). curently the drop shadow card seems to disppear below existing cards, but I think those should be on top.

It works when there are cards already in the holder (like with a stackOffset), but not if they just create a pile.  I'm pretty sure that is a changed behavior.  At some point in this PR, I'm sure there was a dropShadow even on piles.  @flackr, can you check that out?

flackr:
> > I haven't read all the comments on this one, so this might have already been sugested. I'd like to see the drop shadow, also when there are already cards in a holder (see holders 9yy and g1rq in the pr test room). curently the drop shadow card seems to disppear below existing cards, but I think those should be on top.
> 
> It works when there are cards already in the holder (like with a stackOffset), but not if they just create a pile. I'm pretty sure that is a changed behavior. At some point in this PR, I'm sure there was a dropShadow even on piles. @flackr, can you check that out?

It was working before the optimization. I figured out the cause of the issue and have fixed it in the latest. Take a look :-).

96LawDawg:
> It was working before the optimization. I figured out the cause of the issue and have fixed it in the latest. Take a look :-).

Yep.  Works again. @Ingegneus, this is what you wanted, right?

Ingegneus:
sweet! looks nice. how crazy would it be if it worked for all dragged widgets on applicable dropTargets? like cards on piles

flackr:
Also fixed the dragging-shadow class not getting applied to the checkerspiece due to `classes` getting overridden by the class from the `face` object. The checkers piece shadow now also works correctly.

flackr:
> like cards on piles

Can you explain what you mean here? Or is there an example I could look at?

flackr:
If we want this to support a shadow for when a card would be added to a pile I suppose we could determine during the drag that the target location would result in a pile being formed, and then we could look at the pile properties to see if it wants a drop shadow. This is non-trivial, but would be possible to add in a future PR without changing the existing format.

96LawDawg:
> Also fixed the dragging-shadow class not getting applied to the checkerspiece due to `classes` getting overridden by the class from the `face` object. The checkers piece shadow now also works correctly.

Could that be used to fix the issue in the JSON Editor where the yellow highlight doesn't work on the pawn (not for this PR)?

Ingegneus:
> If we want this to support a shadow for when a card would be added to a pile I suppose we could determine during the drag that the target location would result in a pile being formed, and then we could look at the pile properties to see if it wants a drop shadow. This is non-trivial, but would be possible to add in a future PR without changing the existing format.

would it be possible to generalize this to work with all widget types and dropTargets? for example being able to define a custom widget type that has the drpTarget set to another custom widget type and still being able to render the drop shadow?

robartsd:
> Could that be used to fix the issue in the JSON Editor where the yellow highlight doesn't work on the pawn (not for this PR)?

The pawn uses a clip-path to create the shape, nothing from the DOM element can render outside of the clip-path.

flackr:
> > If we want this to support a shadow for when a card would be added to a pile I suppose we could determine during the drag that the target location would result in a pile being formed, and then we could look at the pile properties to see if it wants a drop shadow. This is non-trivial, but would be possible to add in a future PR without changing the existing format.
> 
> would it be possible to generalize this to work with all widget types and dropTargets? for example being able to define a custom widget type that has the drpTarget set to another custom widget type and still being able to render the drop shadow?

There's nothing holder specific about the way this works and it could be made to work in general cases. `dropTarget` is technically supposed to be used on holders AFAICT though.

Ingegneus:
> > > If we want this to support a shadow for when a card would be added to a pile I suppose we could determine during the drag that the target location would result in a pile being formed, and then we could look at the pile properties to see if it wants a drop shadow. This is non-trivial, but would be possible to add in a future PR without changing the existing format.
> > 
> > 
> > would it be possible to generalize this to work with all widget types and dropTargets? for example being able to define a custom widget type that has the drpTarget set to another custom widget type and still being able to render the drop shadow?
> 
> There's nothing holder specific about the way this works and it could be made to work in general cases. `dropTarget` is technically supposed to be used on holders AFAICT though.

oh ups I forgot about that. but still having the drop shadow for an arbitrary widget type would be optimal

flackr:
> oh ups I forgot about that. but still having the drop shadow for an arbitrary widget type would be optimal

I still don't quite understand what is missing that you're asking for. If you set `dropShadow: true` on a holder it will create a shadow for any arbitrary widget type that matches the `dropTarget` when it is dragged over it.

Are you asking for the way I wrote the PR originally to be a property on the dragged widget rather than on the holder it's dragged over?

96LawDawg:
> Are you asking for the way I wrote the PR originally to be a property on the dragged widget rather than on the holder it's dragged over?

The only other thing that makes sense is piles.  You can only drop things into holders and piles, so dropShadow doesn't make sense anywhere else.  What you could do, for this PR or a follow on one, is check to see if the widget is in range of a pile (10px), line 2162 of widget.js.  If it hits that trigger point AND if onPileCreation has dropShadow:true, then make a shadow.  Seems like that would be a pretty easy addition.

flackr:
> > Are you asking for the way I wrote the PR originally to be a property on the dragged widget rather than on the holder it's dragged over?
> 
> The only other thing that makes sense is piles. You can only drop things into holders and piles, so dropShadow doesn't make sense anywhere else. What you could do, for this PR or a follow on one, is check to see if the widget is in range of a pile (10px), line 2162 of widget.js. If it hits that trigger point AND if onPileCreation has dropShadow:true, then make a shadow. Seems like that would be a pretty easy addition.

Yeah I mentioned this option for pile support above as well: https://github.com/ArnoldSmith86/virtualtabletop/pull/1501#issuecomment-1374009676

I agree it doesn't really make sense anywhere else.

The only other thing I could think of which wouldn't currently work is if you set a transition on the shadow it might be nice if the shadow moved from the dragged item to the holder.

Ingegneus:
> > oh ups I forgot about that. but still having the drop shadow for an arbitrary widget type would be optimal
> 
> I still don't quite understand what is missing that you're asking for. If you set `dropShadow: true` on a holder it will create a shadow for any arbitrary widget type that matches the `dropTarget` when it is dragged over it.
> 
> Are you asking for the way I wrote the PR originally to be a property on the dragged widget rather than on the holder it's dragged over?

I tried it before and it didn't work. I tried it again and it works as it should. probably did something wrong in my first try. so I have nothing else to add ^^

rogerl50:
This is really cool.

When dropping into the holder with a `alignChildren = false`, nothing gets highlighted. Given that other valid drop targets *are* highlighted, I think this is confusing to the end user, who probably think this is not a valid target. IMO the entire holder should be shadowed, but I don't know if that's possible given the code already there.

flackr:
> This is really cool.

Thanks, I hope we can merge it soon.

> When dropping into the holder with a `alignChildren = false`, nothing gets highlighted. Given that other valid drop targets _are_ highlighted, I think this is confusing to the end user, who probably think this is not a valid target. IMO the entire holder should be shadowed, but I don't know if that's possible given the code already there.

The shadow is a representation of the thing you are dropping, which is important for cases where it inserts into a stackOffsetX hand. Game developers could add a `.dropTarget` style to the holder if they want to highlight the entire holder but I don't think this makes sense as a default behavior of dropShadow to show something completely different for non-alignChildren holders.

96LawDawg:
> The shadow is a representation of the thing you are dropping, which is important for cases where it inserts into a stackOffsetX hand. Game developers could add a `.dropTarget` style to the holder if they want to highlight the entire holder but I don't think this makes sense as a default behavior of dropShadow to show something completely different for non-alignChildren holders.

I agree with @flackr that the current behavior in this PR makes sense.  If a designer has holders in the room with different alignChildren properties, it is possible to manually create what @rogerl50 asks for.  I do something similar already in Agra.

rogerl50:
> @rogerl50 Also note that something is shown but it's usually covered up by what you're dragging because the place it would drop is the same. This isn't the case if you have 3d perspective, e.g. https://test.virtualtabletop.io/PR-1501/3d or if you apply a transform style to the dragged item to offset it from the drop location, e.g. https://test.virtualtabletop.io/PR-1501/transform

Thanks. Good to go as is.

ArnoldSmith86:
The card doesn't land on the shadow if it's very far to the right:

[cinnamon-2023-01-13T205138+0100.webm](https://user-images.githubusercontent.com/73538315/212406879-2d3c3155-5353-4817-93c8-8b73365a3177.webm)

RaphaelAlvez (commented):
tested in meeting:
- [ ] conflicts while draging
- [ ] position when droping
- [ ] shadows with childs

flackr:
My plan for the first two items (for my own memory) is:
1. To resolve conflicts while dragging, only allow shadows from one player at a time in a particular shared holder. Since everyone can see what's going on this shouldn't be confusing and will avoid multiple players updating the position of widgets in the holder at the same time which should prevent the conflict.
2. To fix the position when dropping, figure out if we can skip realigning children after removing the shadow widget before adding the real widget. This way the relative insertion point will still be the same.

96LawDawg:
> 1. To resolve conflicts while dragging, only allow shadows from one player at a time in a particular shared holder. Since everyone can see what's going on this shouldn't be confusing and will avoid multiple players updating the position of widgets in the holder at the same time which should prevent the conflict.

Unless it is a hand.  Then you need to allow everyone to work at the same time and there shouldn't be conflicts because of ownership.

flackr:
This now has everything we talked about during the meeting fixed. The only thing I haven't done yet is @ArnoldSmith86's suggestion to clean up drop shadow clones when a conflict occurs (should be much more rare now).

96LawDawg:
Was playing with 2 clients and still able to create a problem delta.  It didn't happen when just dragging over the same holder.  But randomly moving in and out of those 2 holders in the demo room in the upper right while 2 players are dragging did it.

flackr:
> Was playing with 2 clients and still able to create a problem delta. It didn't happen when just dragging over the same holder. But randomly moving in and out of those 2 holders in the demo room in the upper right while 2 players are dragging did it.

Yeah, I can't prevent it from ever happening (there's lag in sending the update that someone has a shadow widget in the holder), but it should be as hard to make happen as a conflict from two people dropping cards into the same holder at the same time.

96LawDawg (approved):
Meeting comments addressed as far as I can tell.  Approved through test.

flackr:
Implemented what @ArnoldSmith86 suggested. The server removes the shadow clone of a player who just encountered a synchronization error. @96LawDawg it's a bit tricky to test but if you try to drag on two clients into a holder at the same time you can hit the condition some of the time. When this happens the shadow of the desync'd client should now be cleaned up.

96LawDawg:
> Implemented what @ArnoldSmith86 suggested. The server removes the shadow clone of a player who just encountered a synchronization error. @96LawDawg it's a bit tricky to test but if you try to drag on two clients into a holder at the same time you can hit the condition some of the time. When this happens the shadow of the desync'd client should now be cleaned up.

Tested.  Confirmed.  The dragging widget still freezes, but the shadow clone is removed.  However, previously, the error in the console would disappear when you started dragging the widget and that forced the clone to disappear.  Now there is an odd permanent note in the console, even if you reload the room.  It doesn't affect gameplay, but any ideas?

flackr:
> any ideas?

Oh I know exactly what this is. Fixed.

96LawDawg:
Oh, yeah.  You need to add dropShadow: true as the default for a hand.

flackr:
> Oh, yeah. You need to add dropShadow: true as the default for a hand.

Done.

> Also, do we want/need to add the css transition so that it looks better in the hand?

Maybe we can do this in a followup. I like to avoid mega-PR's if possible :-)

96LawDawg (changes requested):
Still a problem with a clone getting lost, added to the JSON, and causing an error forever.  For example, the last line of JSON when viewed in a text editor is:

"wzbu": {
    "x": 220.75,
    "z": 4
  }

That causes an error in the console every time the game loads.  Maybe not fixable, but something to think about some more.

flackr:
> Still a problem with a clone getting lost, added to the JSON, and causing an error forever. For example, the last line of JSON when viewed in a text editor is:
> 
> "wzbu": { "x": 220.75, "z": 4 }
> 
> That causes an error in the console every time the game loads. Maybe not fixable, but something to think about some more.

Can you cause this in a fresh room though? I feel like the error was because of the incorrect server cleanup and once committed to the room state, sticks unfortunately.

flackr:
@96LawDawg I fixed the state in the pr-test room, so you can try recreating the problem there? My theory is that once the state was messed up from the earlier version before I fixed it that it persisted, but now that the server cleans up the dragShadowWidget it shouldn't happen again.

flackr:
I tested reproducing the conflict and saw no error after dragging the previously dragged widget so I do believe this theory is correct and the issue was resolved with the fix I posted.

96LawDawg:
Your fix works in the top 2 holders when you create a conflict, but not the bottom 2 (hand and holder with stackOffsetX).  Maybe because one is a hand?  No idea.

flackr:
> Your fix works in the top 2 holders when you create a conflict, but not the bottom 2 (hand and holder with stackOffsetX). Maybe because one is a hand? No idea.

After trying to reproduce it on my own server forever, I eventually tried on the testing server and was able to reproduce. I guess it needs some latency between the server and myself. Turning on network throttling in dev tools let me reproduce on my own test server.

Ultimately the root cause seems to be that the client which isn't desynced for some reason sends a position update for the deleted widget. For now, I've added a check on the server code which desyncs the client when this happens which results in the state remaining valid - i.e. no more console error messages.

96LawDawg (approved):
That fixed it.  Approved through test.

96LawDawg:
I don't know why Reward keeps failing the test.  But it is showing a hash error.  Sometimes, it is just not clicking a button and a do-over fixes it.  But this keeps failing.  I looked into it some, but don't know why.

flackr:
> I don't know why Reward keeps failing the test. But it is showing a hash error. Sometimes, it is just not clicking a button and a do-over fixes it. But this keeps failing. I looked into it some, but don't know why.

Yeah I manually walked through the clicks that the Reward test does and didn't see any errors and everything worked. Unfortunately testcafe stopped working on my local machine at some point which means I can no longer use the handy tools at https://github.com/ArnoldSmith86/virtualtabletop/wiki/TestCaf%C3%A9#when-tests-fail to see what the diff is.

ArnoldSmith86:
IMO we should add this to existing hands using a file updater. It needs to check for childrenPerOwner, stackOffset, (parent)ChangeRoutine, enterRoutine, leaveRoutine, (parent)GlobalChangeRoutine. Anything else? Or does somebody think that's a bad idea?

I just think it would be a waste if nobody sees this.

96LawDawg:
> IMO we should add this to existing hands using a file updater. It needs to check for childrenPerOwner, stackOffset, (parent)ChangeRoutine, enterRoutine, leaveRoutine, (parent)GlobalChangeRoutine. Anything else? Or does somebody think that's a bad idea?

So in the MOVEXY PR you are worried about the complication of adding a few lines to every MOVEXY.  Fair enough.  You are right.  But here you are suggesting wrapping every existing routine on every hand with an IF condition that checks to see whether the widget is a shadow.  I'm not saying don't do it, but it will take a lot of testing.  leaveRoutines are especially problematic because they fire twice.  There is more than one way to get around that, and some games may inadvertently use the fact that they fire twice.  Can we be sure the file updater solution works for all those scenarios?

flackr:
> IMO we should add this to existing hands using a file updater. It needs to check for childrenPerOwner, stackOffset, (parent)ChangeRoutine, enterRoutine, leaveRoutine, (parent)GlobalChangeRoutine. Anything else? Or does somebody think that's a bad idea?

I added this to the PR so that we could play with it, but I haven't quite figured out what you mean by some of the proposed changeRoutine checks.

flackr:
I suspect it's probably safer though to just go through the public library and manually add this to everything that should be okay with it.

96LawDawg:
> I suspect it's probably safer though to just go through the public library and manually add this to everything that should be okay with it.

That may be a good way to test, BUT, we need to catch every game on the server, not just those in the PL.

96LawDawg:
Oh.  I think I misunderstood what we were doing.  We are NOT adding this to hands with those routines.  Thank God.  Just adding it to basic hands.  Easy enough.

flackr:
> > I suspect it's probably safer though to just go through the public library and manually add this to everything that should be okay with it.
> 
> That may be a good way to test, BUT, we need to catch every game on the server, not just those in the PL.

I think perhaps my comment was misunderstood. I was suggesting that it may be safer to not put this in the fileupdater and instead go edit every public library game one by one as we test them. Then there's nothing not in the PL gets it automatically.

ArnoldSmith86:
There are so many games out there that wouldn't benefit from this without file updater. It's fine but I don't see a reason yet why we shouldn't do it when the hand is just a default hand.

I believe LawDawg already figured it out but just in case: I propose complicated checks _but_ in the file updater. The change in the game would just be adding the property to the hand.

We should only add it if nothing in the game reacts to the added enter and leave events.

So:

- no enterRoutine or leaveRoutine on the hand itself
- no parentChangeRoutine or changeRoutine on the card (or pile)
- no parentGlobalUpdateRoutine or globalUpdateRoutine on _any_ widget in the room

I feel like onEnter and onLeave are save because they only act on the ghost. They could mess with the ghost class etc of course. To be safe we could only add the property if onEnter only changes activeFace and nothing else. Should be 95% of onEnter.

96LawDawg:
> I propose complicated checks _but_ in the file updater. The change in the game would just be adding the property to the hand.

I agree that is a good idea.  For a follow-on PR.  This is nearly ready and I think your suggestion will slow it down unnecessarily.  I see no harm in making your file updater its own PR.  As @flackr said above

> Maybe we can do this in a followup. I like to avoid mega-PR's if possible :-)

flackr:
> There are so many games out there that wouldn't benefit from this without file updater. It's fine but I don't see a reason yet why we shouldn't do it when the hand is just a default hand.

Yeah, it would be awesome to get this feature part of more games. +1 to trying to do this.

> * no parentChangeRoutine or changeRoutine on the card (or pile)

changeRoutines on the pile should be fine with the current PR because the shadow widget is not allowed to be part of piles (in order to avoid incorrect pile counts): https://github.com/ArnoldSmith86/virtualtabletop/pull/1501/files#diff-cda5cc45823227e11eaa23e3e718a50f300e3fa6183280043e48e4e55b78024dR2239 . That said I expect they're also extremely uncommon on piles anyways and it makes the code pattern more generic if I don't differentiate piles from other stuff.

flackr:
I tested each of the public library games which was auto-modified by the file updater and as far as I can tell the added dropShadows does not affect the gameplay adversely in any of them. In most of them it is useful except for Rummy Tiles where the shadow isn't visible because it shows directly behind the widget anyways.

Can someone explain the remaining testcafe failure? How does the edit test know that after adding some widgets they must have a particular id? (i.e. in this case `#w_3fseC1`). I ran through the steps manually and didn't encounter any errors - the widgets were added the same as they were on the main branch.

robartsd:
> Can someone explain the remaining testcafe failure? How does the edit test know that after adding some widgets they must have a particular id? (i.e. in this case `#w_3fseC1`). I ran through the steps manually and didn't encounter any errors - the widgets were added the same as they were on the main branch.

The ID is generated randomly, but for TestCafe, all randomness is not so random (see `prepareClient()` in test-util.js) so the exact same sequence of values comes each time. The shadow widgets are getting assigned a random ID, so all subsequent random values are no longer the same as they were before these changes.

ArnoldSmith86:
It's possible to use the non-random random function in a proper browser and do all the clicks.

ArnoldSmith86:
Does TestCafé also fail with Chrome/Chromium? You should also be able to use it on gitpod.io: https://gitpod.io/#https://github.com/flackr/virtualtabletop/tree/drop-shadow , https://github.com/ArnoldSmith86/virtualtabletop/wiki/TestCaf%C3%A9#running-tests-in-gitpod .

robartsd:
> Does TestCafé also fail with Chrome/Chromium? You should also be able to use it on gitpod.io: https://gitpod.io/#https://github.com/flackr/virtualtabletop/tree/drop-shadow , https://github.com/ArnoldSmith86/virtualtabletop/wiki/TestCaf%C3%A9#running-tests-in-gitpod .

Not currently working correctly in gitpod:
```
> virtualtabletop@0.5.0 testcafe-firefox-headless
> testcafe firefox:headless

Browserslist: caniuse-lite is outdated. Please run:
  npx browserslist@latest --update-db
  Why you should do it regularly: https://github.com/browserslist/browserslist#browsers-data-updating
ERROR Cannot prepare tests due to the following error:

ReferenceError: window is not defined
    at Object.<anonymous> (/workspace/virtualtabletop/client/js/connection.js:84:1)
    at Object.<anonymous> (/workspace/virtualtabletop/client/js/main.js:2:1)
    at Object.<anonymous> (/workspace/virtualtabletop/tests/client/client-util.js:1:1)

Type "testcafe -h" for help.
```

flackr:
Thanks @robartsd for the test update! I filed a separate issue #1565 for issues around running testcafe. I'll try your instructions out on my system.

server/fileupdater.mjs
```diff
@@ -7,56 +7,106 @@ export default function FileUpdater(state) {
...
+}
+
+function hasPropertyCondition(properties, condition) {
+  if (typeof properties != 'object')
+    return false;
+  if (condition(properties))
+    return true;
+  if (properties.type) {
+    if (hasPropertyCondition(properties.onPileCreation, condition))
+      return true;
```
ArnoldSmith86, line 47: Should this also check `onEnter` and `onLeave`?

server/fileupdater.mjs
```diff
@@ -7,56 +7,106 @@ export default function FileUpdater(state) {
...
+      }
+    }
+    if (properties.type == 'deck') {
+      if (hasPropertyCondition(properties.cardDefaults, condition))
+        return true;
+      if (typeof properties.cardTypes == 'object')
+        for(const cardType in properties.cardTypes)
+          if (hasPropertyCondition(properties.cardTypes[cardType], condition))
+            return true;
+    }
```
ArnoldSmith86, line 61: This seems to be missing `faceTemplates.x.properties`.

ArnoldSmith86 (approved):
Code looks fine to me but it's complex enough that it'll probably break a thing or two. :man_shrugging:

See if my two comments warrant a change and then merge.

server/fileupdater.mjs
flackr: Done.

server/fileupdater.mjs
flackr: As far as I can tell, faceTemplates properties don't ever set properties on the widget so things like `parentChangeRoutine` would be irrelevant on those wouldn't it?

I also noticed updateProperties doesn't update those either - though there have been specific version updates which do things to face templates.

ArnoldSmith86: > As far as I can tell, faceTemplates properties don't ever set properties on the widget so things like `parentChangeRoutine` would be irrelevant on those wouldn't it?

They "set" properties on the cards just like `cardDefaults` or `cardTypes`. Note that the properties are inside a property called `properties` in the face definition.

> I also noticed updateProperties doesn't update those either - though there have been specific version updates which do things to face templates.

I thought, I saw them yesterday.

flackr: Thanks, it wasn't very obvious to me where this happened but I see now the properties are applied in the card's `applyDeltaToDOM` function. I've updated the fileupdater logic to check this.

PR #1237, opened 2022-06-07
Yet another game pieces pr

RaphaelAlvez:
yes I'm doing this in the most stupid way

Just a very basic PR

I just want to be able to add more game pieces.

- [x] rearrange columns
- [x] add @mousewax game pieces
- [x] add dividers to the decorative section

if anyone wants to do it (@ArnoldSmith86 🙄):
- [ ] add color picker for game pieces
   - [ ] remove duplicate for PCIO game pieces
- [ ] add option menu for spinners (to make space for dices)

PR #1506, opened 2023-01-05
Change layout of add widget screen.

rogerl50:
This changes the add widget screen to a grid and makes the crowded panes much larger.

### DIFF
--- a/client/css/editmode.css
+++ b/client/css/editmode.css
@@ -1,17 +1,21 @@
 #addOverlay {
+  display:grid;
+  grid-template-columns: 19.7% 58.5% 19.7%;
+  grid-template-rows: 1fr 1fr;
+  gap: 10px;
+  justify-content: space-around;
   width: var(--roomWidth);
   height: var(--roomHeight);
   transform: scale(var(--scale));
   transform-origin: left top;
-  flex-direction: row;
   padding: 0;
+  overflow:hidden;
 }
 
 #addOverlay > .category {
   background: var(--roomColor);
   width: 100%;
   border-radius: 32px;
-  margin: 20px;
   text-align: center;
 }
 
@@ -25,6 +29,32 @@
   margin-right: auto;
   margin-bottom: 8px;
 }
+
+#addOverlay >.category.cards {
+  grid-column: 1 / span 1;
+  grid-row: 1 / span 2;
+}
+
+#addOverlay >.category.pieces {
+  grid-column: 2 / span 1;
+  grid-row: 1 / span 1;
+}
+
+#addOverlay >.category.interactive {
+  grid-column: 2 / span 1;
+  grid-row: 2 / span 1;
+}
+
+#addOverlay >.category.decorative {
+  grid-column: 3 / span 1;
+  grid-row: 1 / span 1;
+}
+
+#addOverlay >.category.custom {
+  grid-column: 3 / span 1;
+  grid-row: 2 / span 1;
+}
+
 #uploadBoard::before, #uploadToken::before{
   content:'[upload]';
 }
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -88,14 +88,18 @@ export function showOverlay(id, forced) {
 
   if(id) {
     const style = $(`#${id}`).style;
-    style.display = !forced && style.display === 'flex' ? 'none' : 'flex';
-    overlayActive = style.display === 'flex';
+    style.display = !forced && style.display !== 'none' ? 'none' : 'flex';
+    overlayActive = style.display !== 'none';
     if(forced)
       overlayActive = 'forced';
 
     //Hack to focus on the Go button for the input overlay
     if (id == 'buttonInputOverlay') {
       $('#buttonInputGo').focus();
+    }
+    //Hack to set display mode properly for add widget
+    if (id == 'addOverlay') {
+      style.display = 'grid';
     }
     toServer('mouse',{inactive:true})
   } else {
@@ -333,7 +337,7 @@ onLoad(function() {
     if(e.currentTarget.classList.contains('active')) {
       if($('#stateDetailsOverlay.notEditing') && $('#stateDetailsOverlay.notEditing').style.display == 'flex')
         showStatesOverlay('statesOverlay');
-      if(e.currentTarget == $('#activeGameButton') && $('#addOverlay').style.display == 'flex')
+      if(e.currentTarget == $('#activeGameButton') && $('#addOverlay').style.display != 'none')
         showOverlay();
       e.stopImmediatePropagation();
       return;
--- a/client/js/editmode.js
+++ b/client/js/editmode.js
@@ -655,40 +655,45 @@ function populateAddWidgetOverlay() {
   });
 
   // Populate the Game Pieces panel in the add widget overlay
-  let y = 100;
+  let xC = 380;
+  const xOff = 300;
+  let yC = 80;
+  const yOff = 88;
   // First the pins, checkers, and pawns
-  for(const color of [ '#bc5bee','#4c5fea','#23ca5b','#e0cb0b','#e2a633','#e84242','#000000','#4a4a4a','#ffffff' ]) {
-    addWidgetToAddWidgetOverlay(new BasicWidget('add-pin-'+color), {
-      classes: 'pinPiece',
-      color,
-      width: 35.85,
-      height: 43.83,
-      x: 380,
-      y
-    });
+  const color = [ '#bc5bee','#4c5fea','#23ca5b','#e0cb0b','#e2a633','#e84242','#000000','#4a4a4a','#ffffff' ];
+  for (let col=0; col <=2; col++)
+    for (let row=0; row<=2; row++) {
+      const thisColor = color[3*col+row];
+      addWidgetToAddWidgetOverlay(new BasicWidget('add-pin-'+thisColor), {
+        classes: 'pinPiece',
+        color: thisColor,
+        width: 35.85,
+        height: 43.83,
+        x: xC + xOff*col,
+        y: yC + yOff*row
+      });
 
-    addWidgetToAddWidgetOverlay(new BasicWidget('add-checkers-'+color), {
-      faces: [
-        { classes: "checkersPiece"         },
-        { classes: "checkersPiece crowned" }
-      ],
-      color,
-      width: 73.5,
-      height: 73.5,
-      x: 440,
-      y: Math.round(y + (43.83 - 73.5)/2)
-    });
+      addWidgetToAddWidgetOverlay(new BasicWidget('add-checkers-'+thisColor), {
+        faces: [
+          { classes: "checkersPiece"         },
+          { classes: "checkersPiece crowned" }
+        ],
+        color: thisColor,
+        width: 73.5,
+        height: 73.5,
+        x: xC + xOff*col + 60,
+        y: yC + yOff*row + Math.round((43.83 - 73.5)/2)
+      });
 
-    addWidgetToAddWidgetOverlay(new BasicWidget('add-classic-'+color), {
-      classes: 'classicPiece',
-      color,
-      width: 56,
-      height: 84,
-      x: 528,
-      y: Math.round(y + (43.83 - 84)/2)
-    });
-    y += 88;
-  }
+      addWidgetToAddWidgetOverlay(new BasicWidget('add-classic-'+thisColor), {
+        classes: 'classicPiece',
+        color: thisColor,
+        width: 56,
+        height: 84,
+        x: xC + xOff*col + 150,
+        y: yC + yOff*row + Math.round((43.83 - 84)/2)
+      });
+    }
 
   // Next the unicode symbols
   const centerStyle = 'color:black;display:flex;justify-content:center;align-items:center;text-align:center;';
@@ -698,7 +703,7 @@ function populateAddWidgetOverlay() {
     width: 25,
     height: 25,
     x: 380,
-    y: y + 5
+    y: yC + yOff*3 + 25
   });
 
   addWidgetToAddWidgetOverlay(new BasicWidget('add-unicodeM'), {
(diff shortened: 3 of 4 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1506/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1506-pr-test`.

RaphaelAlvez:
@rogerl50 would you mind mixing this with #1237?

I stopped working on that PR because I had trouble with space. Also we wanted to add a color picker for game pieces and a drop-down for spinners.

if not, here's my review on the current setup:
I prefer the current layout in general. I just have some minor suggestions

image
this would look better side by side

image
I believe we decided we could remove those in my PR

image
Right side has more space

rogerl50:
> @rogerl50 would you mind mixing this with #1237?
> 

> if not, here's my review on the current setup: I prefer the current layout in general. I just have some minor suggestions
When you say *current layout*, do you mean you'd rather not make any of these changes?

rogerl50:
I didn't see #1237. I see no reason to use this PR instead of yours, unless you prefer this layout.

RaphaelAlvez:
> I didn't see #1237. I see no reason to use this PR instead of yours, unless you prefer this layout.

No. I prefer your layout.

I was asking if you could help me adding the needed inputs that kept that PR from progressing.

rogerl50:
> > I didn't see #1237. I see no reason to use this PR instead of yours, unless you prefer this layout.
> 
> No. I prefer your layout.
> 
> I was asking if you could help me adding the needed inputs that kept that PR from progressing.

I thought briefly about a color picker and spinner picker, just long enough to realize I had no idea how to do it :)
We probably need more than just a `formField`; at the least we need some way to cancel the operation. Perhaps we can reuse the INPUT code (as @ArnoldSmith86 just did for the scoreboard editor).

ArnoldSmith86:
Keep in mind that I'll start with the new editor soon and that will look completely different anyway. Not saying, I don't want this. Just a heads-up in case you plan to spend a lot of time on this. :smile:

RaphaelAlvez:
> Keep in mind that I'll start with the new editor soon and that will look completely different anyway. Not saying, I don't want this. Just a heads-up in case you plan to spend a lot of time on this. 😄

do you realize that this is going to take months again? the inputs would be a great sort term imporovements

rogerl50:
Latest commit adds a popup for spinner values selection. Someone should look at the code. Also, it'd be nice to have the popup show up where the spinner is, but I'm not sure that's possible.

rogerl50:
> Keep in mind that I'll start with the new editor soon and that will look completely different anyway. Not saying, I don't want this. Just a heads-up in case you plan to spend a lot of time on this. 😄

So all of my knowledge of jsonedit.js will be useless? 😢

RaphaelAlvez (dismissed):
from a quick code review I can't pin point what is the problem but the overlay is being displayed every time I open the room

RaphaelAlvez:
> > Keep in mind that I'll start with the new editor soon and that will look completely different anyway. Not saying, I don't want this. Just a heads-up in case you plan to spend a lot of time on this. 😄
> 
> So all of my knowledge of jsonedit.js will be useless? 😢

should be a parallel feature

ArnoldSmith86:
Yeah, I won't touch that a lot. I'll just have to look at top-level commands. Those should be available in the normal editor. But the JSON editor will just be part of that.

rogerl50:
> from a quick code review I can't pin point what is the problem but the overlay is being displayed every time I open the room

Yeah. Displayed and then hidden again. I'll have to take a look. Right now I'm implementing color choosers for the pieces, then I'll clean up the code and commit again, then (by this time it will be tomorrow) I'll try to track down that problem.

RaphaelAlvez:
> > from a quick code review I can't pin point what is the problem but the overlay is being displayed every time I open the room
> 
> Yeah. Displayed and then hidden again. I'll have to take a look. Right now I'm implementing color choosers for the pieces, then I'll clean up the code and commit again, then (by this time it will be tomorrow) I'll try to track down that problem.

It would be better if it were outside of the widget click. Somewhere in the top right of the game pieces section.

But I'm happy with whatever you can come up with

rogerl50:
Latest commit pops a color dialog for pieces as well. Also fixes the problem of the edit overlay flashing on load.

There's lots of space now; I haven't bothered to lay things out since there will be new pieces from #1237. @RaphaelAlvez we can merge these in either direction, whichever is easier.

rogerl50:
This will cause TestCafe tests to fail, since those buttons are no longer there...

RaphaelAlvez:
> This will cause TestCafe tests to fail, since those buttons are no longer there...

I prefer to close #1237 and bring to changes to here. That PR is pretty old and it will probably be pure code conflicts

rogerl50:
> > This will cause TestCafe tests to fail, since those buttons are no longer there...
> 
> I prefer to close #1237 and bring to changes to here. That PR is pretty old and it will probably be pure code conflicts

 OK, why don't you port that over and then we can deal with the layout. There may be other opportunities for using the color picker once you do that, as well.

RaphaelAlvez:
I just saw you changed the structure of you grid. In no way mine was better. I don't even remember why I did stuff back there.

Also I'm not home as soon as I can I'll ring the changes to here. Also I can't see your changes at the moment because my phone doesn't open edit mode.

Definitely happy to close that PR. It's my oldest PR and it always bother me to see it there untouched

rogerl50:
> I just saw you changed the structure of you grid. In no way mine was better. I don't even remember why I did stuff back there.
> 
> Also I'm not home as soon as I can I'll ring the changes to here. Also I can't see your changes at the moment because my phone doesn't open edit mode.
> 
> Definitely happy to close that PR. It's my oldest PR and it always bother me to see it there untouched

I changed it because yours worked and mine really didn't. Maybe we can figure out a cleaner way.

96LawDawg:
Not much in the way of substance to add so far, but I like where this is going.  I would suggest adding a header to the color picker like "Choose game piece color".  I think purple is an odd choice for the default color.  I would go with blue or maybe VTT blue as it is more "normal" and more in the center of the color picker (at least on Chrome and FF).  I would also change the default spinner to 6 and set the default number value to start at 6.  But those are just suggestions.  None of that really matters.

rogerl50:
> Not much in the way of substance to add so far, but I like where this is going. I would suggest adding a header to the color picker like "Choose game piece color". I think purple is an odd choice for the default color. I would go with blue or maybe VTT blue as it is more "normal" and more in the center of the color picker (at least on Chrome and FF). I would also change the default spinner to 6 and set the default number value to start at 6. But those are just suggestions. None of that really matters.

Both good ideas. Done.

rogerl50:
If someone knows how to align the four buttons in the Interactive section properly, I'd be grateful. @flackr @robartsd

96LawDawg:
> If someone knows how to align the four buttons in the Interactive section properly, I'd be grateful. @flackr @robartsd

They are aligned "properly" in the sense that each group of 2 is centered in the div.  But the buttons are different sizes, so it doesn't look like you want.  Is this what you want?

Interactive

You can do that by changing lines 179-180 in room.html to:

```
<button class="ui-button" id="addSeat" style="margin: 0px 17px">Add Seat</button><br>
<button class="ui-button" id="addSeatCounter" style="margin: 0px 17px">+ Counter</button>
```

rogerl50:
I want each button in the bottom row to be centered under the one above it. I was hoping for a solution that didn't involve pixel counting.

ArnoldSmith86:
Put them into a div and use `text-align: center`, maybe?

client/css/editmode.css
```diff
@@ -1,30 +1,67 @@
...
-#addOverlay > .category > button {
+#gridContainerAddOverlay > .category > button {
   display: flex;
   margin-left: auto;
   margin-right: auto;
   margin-bottom: 8px;
 }
+
+#gridContainerAddOverlay > .category > div > button {
+  margin-bottom: 8px;
```
flackr, line 42: Seems like every .ui-button on this page wants margin-bottom: 8px, maybe just add single selector to do this, e.g.:
```css
#gridContainerAddOverlay .ui-button {
  margin-bottom: 8px;
}
```

Though this makes me wonder why we don't have `margin: 8px;` (i.e. on all sides) of all `.ui-button`.

client/css/layout.css
```diff
@@ -936,6 +936,8 @@ button:disabled:hover {
   gap: 8px;
   cursor: pointer;
   box-shadow: rgba(0, 0, 0, 0.12) 0px 1px 3px 0px, rgba(0, 0, 0, 0.24) 0px 1px 2px 0px;
+  margin-left: 8px;
+  margin-right: 8px;
```
flackr, line 940: Why not `margin: 8px;` since it seems everywhere this is used you want to separate them by 8px?

flackr (changes requested):

flackr:
> If someone knows how to align the four buttons in the Interactive section properly, I'd be grateful. @flackr @robartsd

You currently have them in two different (i.e. independent) flex rows so they can't align. If you want them aligned you should put them in a grid, table, or even a pair of flex columns (since their heights are the same).

client/css/editmode.css
```diff
@@ -1,30 +1,67 @@
...
   width: var(--roomWidth);
   height: var(--roomHeight);
   transform: scale(var(--scale));
   transform-origin: left top;
-  flex-direction: row;
   padding: 0;
 }
 
-#addOverlay > .category {
+#gridContainerAddOverlay {
```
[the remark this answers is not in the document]
rogerl50: This didn't quite work. `box-sizing` seemed to have no effect, but I did need to put `gap: 20px` into this style. In any case, I did remove the wrapper element.

client/css/editmode.css
```diff
@@ -1,30 +1,67 @@
...
+  display: grid !important;
```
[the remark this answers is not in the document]
rogerl50: Removing !important resulted in the subsidiary divs being laid out sequentially top to bottom. Did not change this.

client/css/editmode.css
```diff
@@ -1,30 +1,67 @@
...
+  grid-template-rows: 1fr 1fr 1fr;
+  height:100%;
+  margin: 20px;
+  gap: 20px 20px;
+}
+
+#gridContainerAddOverlay > .category {
   background: var(--roomColor);
   width: 100%;
+  height: 100%;
```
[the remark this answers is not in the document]
rogerl50: Done

client/css/layout.css
rogerl50: Done.

client/css/editmode.css
rogerl50: Done.

rogerl50:
@RaphaelAlvez I added stacks of one chip of each value for 2D and 3D. This all seems to work, except that the pile code you put in `cardDefaults` does not kick in for the add widget screen, or when you first put a widget in the room. That code is pretty complicated and I can't figure it out.

ArnoldSmith86:
For the color picker maybe add "inherit from seat X" somehow?

client/js/main.js
```diff
@@ -88,15 +88,19 @@ export function showOverlay(id, forced) {
...
     if(forced)
       overlayActive = 'forced';
 
     //Hack to focus on the Go button for the input overlay
     if (id == 'buttonInputOverlay') {
       $('#buttonInputGo').focus();
     }
+    //Hack to set grid display for add Overlay
+    if (id == 'addOverlay' && style.display !== 'none') {
+      style.display = 'grid'
```
flackr, line 102: Instead of doing this. In JS you can set style.display to either `'none'` or `''` and then in the css file the display listed there will be used if JS sets `''`.

Alternately we could add a class or attribute for whether the overlay is displayed. I.e. typically what's done is something like this in the css file:
```css
.overlay[hidden] {
  display: none;
}
```

With JS then setting or removing the hidden attribute.

rogerl50: This requires changes for every overlay if I'm not mistaken. I'm kind of reluctant to take that on in this fairly simple PR.

rogerl50:
This is ready to merge once @flackr dismisses his requested changes (which have been implemented) and we get an approval or two.

96LawDawg:
The 3-D poker chip needed a dropTarget, so I fixed that.  Because the chips are a deck, I thought we should treat it like the regular deck and I added a recall and sort button.  If you don't like it, it can be deleted/reverted.

client/js/main.js
flackr: All of the overlays are of class `.overlay` so you could have two rules cover all of them:
```css
overlay {
  display: flex;
}

#addOverlay {
  display: grid;
}
```

Alternately if you go with the HTML attribute to toggle visibility,
```css
overlay[hidden] {
  display: none !important;
}

overlay {
  display: flex;
}

#addOverlay {
  display: grid;
}
```

96LawDawg (approved):
Approved through test.  I've clicked every widget in the overlay.  They all work.

rogerl50:
> The 3-D poker chip needed a dropTarget, so I fixed that. Because the chips are a deck, I thought we should treat it like the regular deck and I added a recall and sort button. If you don't like it, it can be deleted/reverted.

Is that the standard recall and shuffle icon now?

96LawDawg:
> Is that the standard recall and shuffle icon now?

Recall and _sort_.  We could just do a recall, but makes sense to me to sort them.  Again, I don't really care.

RaphaelAlvez:
> > Is that the standard recall and shuffle icon now?
> 
> Recall and _sort_.  We could just do a recall, but makes sense to me to sort them.  Again, I don't really care.

I would prefer recall **&** sort but I think it's alright

It makes no sense to shuffle this deck and it can be a good example

96LawDawg:
> I would prefer recall **&** sort but I think it's alright
> 
> I mean to just add an & but that will require a new hash abs maybe not worth

Thanks for clarifying. But that isn't even possible.  We can't mix text and symbols.  @mousewax never got that working.  It is possible to have text and 1 image, but unless we make a special image for this (we should not), then it is either what we have, or just use text.  Obviously, I'm a fan of using what we have.

client/js/main.js
rogerl50: While this is a really clean design, I'm just not comfortable putting it into this PR. Beyond the CSS changes, which are indeed fairly easy (though they require some other changes in `room.html` and a couple of overlay css files), there is at least one other place in the code where `display: none` is tested for (this is in the handler for a click on the toolbar buttons). If this is to be done, it should be its own PR.

flackr: Okay, instead of this override here, can you move 'flex' or 'grid' into a local var so it's not an override after setting the other style. E.g. change lines 91-92 to:

```js
const activeDisplay = id == 'addOverlay' ? 'grid' : 'flex';
style.display = !forced && style.display !== 'none' ? 'none' : activeDisplay;
overlayActive = style.display !== 'none';
```

flackr (approved):
Looks good with removal of overriding display hack.

client/js/main.js
rogerl50: Sold. Still unfortunate, but cleaner.

PR #1510, opened 2023-01-06
fixed that `api/addShareToRoom` could only add one state

ArnoldSmith86:
It calls `addState` with an empty string. I assumed at the time that `addState` would generate a new ID if none is given but it simply uses the empty string as an ID and for the next one it overrides the previous state.

This PR uses the share ID instead.


96LawDawg (approved):
Blind approval.  I have no idea how to test.

ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1510/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1510-pr-test`.

flackr (approved):
Code change LGTM though I'm not familiar with this endpoint and can't find it in any of the source aside from the serving code to know how it's supposed to work.

Issue #1017, opened 2021-12-12
Label's textarea should have a spellcheck property (defaulting to `false`)


robartsd:
In some games, players use labels to enter values that are evaluated by other players. Not spelling correctly can be a clue that a player is guessing. Browser automatic indication of spelling errors can impact game play. This would add the spellcheck attribute set to false to the label's textarea by default. Game designers would be able to turn on automatic spell check by setting the value to true.

96LawDawg:
Closed by #1511

PR #1511, opened 2023-01-07
Add `spellCheck` to labels

96LawDawg:
Implements #1017 by adding a `spellCheck` property to labels that defaults to false.  If true, then spell check is active.

This will have the effect of inactivating spell check for all existing games.  That isn't really game breaking, but it might be a change that some designers didn't account for.  I'm hoping we can agree that a file updater is not necessary.  Alternatively, we could make `spellCheck` true and that would not change existing games, but also would not be what @robartsd requested.  I can go either way.

Also made a slight update to the LABEL function tutorial (the only place where labels are mentioned) to address spell check.

Wiki update:

| spellCheck | false  | If true and if _editable_ is also true, the browser's built-in spell check will be applied when typing in the label. |


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1511/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1511-pr-test`.

RaphaelAlvez (dismissed):
Not code reviewed:
- 2 types of quotes
- what happens if 2 rows alligbnid not undefined?

96LawDawg:
I originally just tested values true and false.  But then I tested a value like "foo" for `spellCheck` and because that is truthy, it checks for spelling.  I tried changing the true false conditions to be strictly equal to false, but that didn't seem to work either.  Just trying to knock out a simple issue.

RaphaelAlvez:
I could not trigger any kind of spell check

client/js/widgets/label.js
```diff
@@ -27,6 +28,11 @@ export class Label extends Widget {
   applyDeltaToDOM(delta) {
     super.applyDeltaToDOM(delta);
     if(delta.text !== undefined || delta.twoRowBottomAlign !== undefined) {
```
flackr, line 30: You need to check if delta.label !== undefined here as well to make sure it updates when just the spellcheck attribute changes.

client/js/widgets/label.js
```diff
@@ -27,6 +28,11 @@ export class Label extends Widget {
...
+      if (this.get('spellCheck') === true) {
+        this.input.setAttribute('spellcheck', true);
+      } else {
+        this.input.setAttribute('spellcheck', false);
+      }
```
flackr, line 35: This if/else block can be shortened to just:
```js
this.input.setAttribute('spellcheck', this.get('spellCheck') === true)
```

RaphaelAlvez: > This if/else block can be shortened to just:
> 
> ```js
> this.input.setAttribute('spellcheck', this.get('spellCheck') === true)
> ```

chatGPT says this.input.setAttribute('spellcheck', this.get('spellCheck')) is enough

flackr: That would set the spellcheck attribute to exactly what was entered as the spellCheck property which could be arbitrary strings, objects which would be coerced into the string `[object Object]` etc. You definitely want to force it to be a boolean and @96LawDawg also wanted it to only set spellcheck = true if the value specified is actually true, rather than arbitrary strings.

96LawDawg: I'm trying to avoid a situation where a label already has a spellCheck property and need for a file updater.  No matter what, this isn't game breaking, but just trying to avoid conflict.

client/js/widgets/label.js
flackr: My apologies, I meant delta.spellCheck, sorry.

96LawDawg: I'm just a monkey hammering on the keyboard doing whatever the AI, Google, or you guys tell me to do.

client/js/widgets/label.js
```diff
@@ -26,7 +27,8 @@ export class Label extends Widget {
 
   applyDeltaToDOM(delta) {
     super.applyDeltaToDOM(delta);
-    if(delta.text !== undefined || delta.twoRowBottomAlign !== undefined) {
+    if(delta.text !== undefined || delta.twoRowBottomAlign !== undefined || delta.spellCheck !== undefined) {
+      this.input.setAttribute('spellcheck', this.get('spellCheck') === true)
```
flackr, line 31: In fact, technically this should go into a separate check similar to delta.editable, because spellcheck changing doesn't require changing all of the input attributes.

I.e. after this if block

```js
if (delta.spellCheck !== undefined)
  this.input.setAttribute('spellCheck', this.get('spellCheck') === true);
```

flackr (approved):
lgtm with suggestion to only update spellcheck attribute and not all input attributes when spellcheck changes.

client/js/widgets/label.js
96LawDawg: So before I do another commit, is this what you are suggesting"

```
 super.applyDeltaToDOM(delta);
    if(delta.text !== undefined || delta.twoRowBottomAlign !== undefined) {
      if(delta.spellCheck !== undefined) {
        this.input.setAttribute('spellcheck', this.get('spellCheck') === true)
      }
      this.input.value = this.get('text');
 ```

Comment (flackr) [reply client/js/widgets/label.js:31]: No, like this:
```js
    super.applyDeltaToDOM(delta);
    if(delta.text !== undefined || delta.twoRowBottomAlign !== undefined) {
      this.input.value = this.get('text');
      ...
    }
    if(delta.spellCheck !== undefined) {
      this.input.setAttribute('spellcheck', this.get('spellCheck') === true)
    }
```

flackr: The way this works is that any change in property value will be in the delta dictionary, so this is saying if the spellcheck value changed, then change the spellcheck attribute on the input element.

96LawDawg: I see it now.  Thanks for the explanation and help.

Issue #1462, opened 2022-12-25
Grids should have option to be aligned by the corner


RaphaelAlvez:
It is intuitive and hard to work with at the moment. _**MOST**_ use cases don't use rotation so using the center coordinates is just wierd.

Maybe just a property that changes it to normal cords would be great.

PR #1388, opened 2022-11-07
Add dropAnchor property to specify anchor point for widgets.

flackr:
The dropAnchor property defines the point of a widget which should be anchored when dropping it into a holder. See #1387 for examples.

PR #1512, opened 2023-01-07
Grid can use widget corner instead of center for snap alignment

96LawDawg:
Implements #1462 by creating new `gridCornerAlign` boolean property.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -52,6 +52,7 @@ export class Widget extends StateManaged {
       clickSound: null,
 
       grid: [],
+      gridCornerAlign: false,
       enlarge: false,
       overlap: true,
       ignoreOnLeave: false,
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1512/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1512-pr-test`.

flackr:
I think maybe we should have an alignment specified for each different grid definition, e.g.

```
  grid: [{ "offsetX": 200, "x": 300, "y": 173.2, "align": "center" }, { "offsetX": 350, "offsetY": 86.6, "x": 300, "y": 173.2, "align": "top left" }, … ]
```

96LawDawg:
> I think maybe we should have an alignment specified for each different grid definition, e.g.
> 
> ```
>   grid: [{ "offsetX": 200, "x": 300, "y": 173.2, "align": "center" }, { "offsetX": 350, "offsetY": 86.6, "x": 300, "y": 173.2, "align": "top left" }, … ]
> ```

I thought about that.  I guess that would work.  Would seem to be confusing to implement.  Do we need additional alignment parameters other than center and top left?

flackr:
> I thought about that. I guess that would work. Would seem to be confusing to implement. Do we need additional alignment parameters other than center and top left?

It's probably interesting to think about how this would interact with dropAnchor (see https://github.com/ArnoldSmith86/virtualtabletop/pull/1388) which has a pair of percentages (but could also have keywords like top/left/bottom/right/center). Should we use dropAnchor for this instead of having a grid property? The two having different values would be weird.

RaphaelAlvez:
I'd suggest using percentages

with values like "align": "0 0" or "align": "0.5 0.5"

then the code can just be something like:
`snapX = x + grid.x*alignX - mod(x - (grid.offsetX || 0), grid.x);`

btw I don't really understand why you need to change the whole code for snapX and snapY.

Issue #1460, opened 2022-12-25
grid should have option to limit widget to the grid


RaphaelAlvez:
at the moment grid allows widgets to be taken of the grid area manually

for tokens, board components and other situations we might want to just allow the widget to be in an area

I propose we add a "locked" property that will force the widget to be in the grid if moved manually

96LawDawg:
Or maybe another way of thinking about this is something that applies to all widgets.

xMinLimit
yMinLimit
xMaxLimit
yMaxLimit

Those define the boundaries outside which the widget is not allowed to go, grid or not.  They all default to null of course.  There have been a number of instances of changeRoutines being used to do the same thing.  This would greatly simplify that and would allow it to work on the room level or the parent level.

Issue #1303, opened 2022-08-26
Can't have dinamically added commands in JSON editor


RaphaelAlvez:
@

PR #1513, opened 2023-01-07
Allow widget movement limits

96LawDawg:
Implements #1460.  Adds new property `dragLimit` that takes the form:

```
 "dragLimit":  {
      "minX": 0,
      "maxX": 100,
      "minY": 0,
      "maxY": 100
    },
```

It defaults to `"dragLimit": {}`


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1513/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1513-pr-test`.

flackr:
Just my opinion but I feel like putting movement restrictions (and grid) on the moving widget rather than the holder is the wrong place to be doing it. I would prefer to see the root treated as a holder (maybe a magic id could be used to specify properties for the root) and then you could specify that it is not a dropTarget for certain widgets which would implicitly limit where they could be dropped. If dropped to an invalid location they could go back to where they came from.

96LawDawg:
> Just my opinion but I feel like putting movement restrictions (and grid) on the moving widget rather than the holder is the wrong place to be doing it. I would prefer to see the root treated as a holder (maybe a magic id could be used to specify properties for the root) and then you could specify that it is not a dropTarget for certain widgets which would implicitly limit where they could be dropped. If dropped to an invalid location they could go back to where they came from.

But there isn't always a holder.  There are plenty of use cases for limiting a widget to the room or a parent that isn't a holder.  I think this is more flexible.  Could you not use this to limit the widget to the holder boundaries?

flackr:
> But there isn't always a holder. There are plenty of use cases for limiting a widget to the room or a parent that isn't a holder. I think this is more flexible. Could you not use this to limit the widget to the holder boundaries?

Right, that's why I said ideally there'd be a special id to specify these properties on the root (i.e. when there's not a holder).

96LawDawg:
I think I'm starting to understand, but I'm not sure I know how to make that work.

flackr:
The reason being that properties like grid or range limits don't work correctly if the holder being dropped into has a transform. I'm planning to write a version of grid for holders to fix this - e.g. maybe childAlign = {grid specification}.

RaphaelAlvez:
I think this should be an object property like grid.

"limit":{"xMin":number,"xMax":number,"yMin":number,"yMax":number}

ArnoldSmith86:
Or maybe add it to `grid`?

RaphaelAlvez:
> Or maybe add it to `grid`?

that was my initial request but maybe @96LawDawg preferred this way for a reason

96LawDawg:
> > Or maybe add it to `grid`?
> 
> that was my initial request but maybe @96LawDawg preferred this way for a reason

I wanted all widgets to have that capability, not just ones with grid.  That was my comment on @RaphaelAlvez's original request.

client/js/widgets/widget.js
```diff
@@ -1818,8 +1819,23 @@ export class Widget extends StateManaged {
...
     const coordParent = this.coordParentFromCoordGlobal(coordGlobal);
     const offset = getOffset(this.coordParentFromCoordLocal(localAnchor), {x: this.get('x'), y: this.get('y')})
-    const newX = Math.round(coordParent.x + offset.x);
-    const newY = Math.round(coordParent.y + offset.y);
+    let newX = Math.round(coordParent.x + offset.x);
+    let newY = Math.round(coordParent.y + offset.y);
+    const limit = this.get('moveLimit');
+ 
+    if (limit.length) {
+      for (const p in limit[0]) {
```
flackr, line 1827: This for loop is completely unnecessary given that you're checking each property separately anyways. The way this works right now is for each property it checks if it's one of the 4 properties you care about.

Instead, since you know exactly which properties you care about you should check those directly, e.g. remove the for loop and do the following:

```js
if (limit[0].minX !== undefined) {
  newX = Math.max(limit[0].minX, newX);
}
...
```

I'm also not sure why limit is an array when you only use the first index. Instead, remove all of the [0] accessors and change limit to be an object, i.e. {} on line 45.

client/js/jsonedit.js
```diff
@@ -554,6 +554,21 @@ const jeCommands = [
       jeSetAndSelect(w.get('width'));
     }
   },
+  {
+    id: 'je_moveLimit',
+    name: 'move limits',
+    context: '^[^ ]* ↦ moveLimit',
+    call: async function() {
+      const w = widgets.get(jeStateNow.id);
+      jeStateNow.moveLimit.push({
```
flackr, line 563: I don't think you want limit to be an array - especially since the code only ever looks at the first index.

flackr (changes requested):

RaphaelAlvez:
> > > Or maybe add it to `grid`?
> > 
> > that was my initial request but maybe @96LawDawg preferred this way for a reason
> 
> I wanted all widgets to have that capability, not just ones with grid.  That was my comment on @RaphaelAlvez's original request.

One solution to both problems: add grid to all widgets

ArnoldSmith86:
It's not in all widgets?

RaphaelAlvez:
> > > Or maybe add it to `grid`?
> > 
> > that was my initial request but maybe @96LawDawg preferred this way for a reason
> 
> I wanted all widgets to have that capability, not just ones with grid.  That was my comment on @RaphaelAlvez's original request.

@ArnoldSmith86 not according to this

96LawDawg:
> It's not in all widgets?

All widgets have the capability for grid.  I wrote imprecisely.  I meant to say this shouldn't be limited to use in connection with grids.  Why bury this capability inside of a grid property when it has nothing to do with grids?

client/js/widgets/widget.js
96LawDawg: > I'm also not sure why limit is an array when you only use the first index.

I'm not sure either.  Ask ChatGPT.

Thanks for the feedback.  I was following the model from grid, but I now remember that grid is an array of objects and I only need the one.

flackr: > > I'm also not sure why limit is an array when you only use the first index.
> 
> I'm not sure either. Ask ChatGPT.

Fortunately I don't think it will be putting me out of a job just yet 😁

client/js/widgets/widget.js
```diff
@@ -1833,8 +1834,22 @@ export class Widget extends StateManaged {
   async move(coordGlobal, localAnchor) {
     const coordParent = this.coordParentFromCoordGlobal(coordGlobal);
     const offset = getOffset(this.coordParentFromCoordLocal(localAnchor), {x: this.get('x'), y: this.get('y')})
-    const newX = Math.round(coordParent.x + offset.x);
-    const newY = Math.round(coordParent.y + offset.y);
+    let newX = Math.round(coordParent.x + offset.x);
+    let newY = Math.round(coordParent.y + offset.y);
+    const limit = this.get('moveLimit');
```
RaphaelAlvez, line 1839: maybe you should add a comment here something like

//Makes drag respect moveLimit property

client/js/widgets/widget.js
```diff
@@ -43,6 +43,7 @@ export class Widget extends StateManaged {
       borderRadius: null,
       rotation: 0,
       scale: 1,
+      moveLimit: {},
```
RaphaelAlvez, line 46: maybe seeing that this is just limited by drag movements we should change it to dragLimit

client/js/widgets/widget.js
RaphaelAlvez, line 1839: maybe you should add a comment here something like

`//Makes drag respect moveLimit property`

or something more useful

RaphaelAlvez (approved):
code reviewed and tested

just two small comments

client/js/jsonedit.js
```diff
@@ -554,6 +554,21 @@ const jeCommands = [
...
+      const w = widgets.get(jeStateNow.id);
+      jeStateNow.moveLimit = {
+        minX: '###SELECT ME###',
+        maxX: 1600 - w.get('width'),
+        minY: 0,
+        maxY: 1000 - w.get('height'),
+      };
+      jeSetAndSelect(0);
+    }
+  },
```
RaphaelAlvez, line 571: this actually breaks existing moveLimit. I think you should just remove it. It will create more frustration then help

RaphaelAlvez (changes requested):

client/js/jsonedit.js
96LawDawg: I assume you mean that if you already have entered something into the dragLimit property and click on the button in the JSON Editor, it resets what you have with this default.  You are correct it does that and that could be a problem.  But it is useful to have something that fills in the property for you.  I followed the grid example (but of course grid is an array of objects, so adding a new one is no problem).

But there has to be a way to just make the button disappear if dragLimit is already a property.  I just don't know how.

client/js/widgets/widget.js
96LawDawg: Good idea.  Done.

client/js/widgets/widget.js
96LawDawg: Done.

client/js/jsonedit.js
```diff
@@ -554,6 +554,21 @@ const jeCommands = [
...
+      const w = widgets.get(jeStateNow.id);
+      jeStateNow.dragLimit = {
+        minX: '###SELECT ME###',
+        maxX: 1600 - w.get('width'),
+        minY: 0,
+        maxY: 1000 - w.get('height'),
+      };
+      jeSetAndSelect(0);
+    }
+  },
```
flackr, line 571: I don't know how to use these edit commands but this looks like it will only let you set one attribute of the drag limit which doesn't seem useful. I don't think you need this part of the change at all, but again not sure if you were intentionally trying to add to these commands.

96LawDawg: The idea is that if dragLimit is empty, you click the button and it fills it with all 4 default elements so you don't have to do the typing.  The defaults leave the widget visible on the screen.  But the problem is instead of the button disappearing, it stays and if you press it again, it overwrites what you have.  Any solutions or ideas @rogerl50 ?

client/js/jsonedit.js
RaphaelAlvez: This should be like normal JSON option where we just as one property.
remove it or leave as you wish and create an issue so someone can make it properly. I don't really know that part of the JSON editor.

96LawDawg:
How about now?  @RaphaelAlvez please take another look.  I fixed the JSON Editor commands so they work as normal.  The only thing I couldn't figure out was how to dynamically calculate the value like I did before.  But I added some defaults that in the editor that shouldn't be too objectionable.

RaphaelAlvez (approved):
Approved after changes to JSON editor

flackr:
@96LawDawg how do you use the JSON editor command? I'd be interested to try out that part.

rogerl50:
Right now it seems that adding `dragLimit` does not fill any values. Is that what's desired? (Sorry, I just saw that you asked me about this PR a while back.)

rogerl50:
I can get the object to populate with the defaults when you initially add `dragLimit` (0, 0, 1600, 1000), but the code to do that is executed when the tables are first set up, and there is no widget there whose dimensions you can look at.

I *think* this falls in the general area of issue #1303, but perhaps @ArnoldSmith86 has another idea.

ArnoldSmith86:
Do you want to make the values command options? Why not just add it to the JSON and let the user edit it?

96LawDawg:
> Do you want to make the values command options? Why not just add it to the JSON and let the user edit it?

I don’t think the actual default values should be anything other than null. I was just trying to make the JSON Editor helpful.

RaphaelAlvez:
> Do you want to make the values command options? Why not just add it to the JSON and let the user edit it? 

That's the current situation. Before there was an auto complete but that simply removed anything you had already done. It's just 4 properties do adding one by one is not a big deal.

Also the JSON editor "defaults" @96LawDawg added are nice. It's quick to change but I like the idea to just make it the limits of the room by "default".

I think the way it is right now is working therefore my approval.

ArnoldSmith86:
Well, I'm pretty sure this has nothing to do with #1303. When executing the JSON button command, it can just access the current widget's height and width.

96LawDawg:
> Well, I'm pretty sure this has nothing to do with #1303. When executing the JSON button command, it can just access the current widget's height and width.

And it did work before I changed all the context stuff. A couple of commits ago, yoI would get defaults of minX and minY of 0, maxX of 1600-width and maxY of 1000-height. I just couldn’t figure out how to do it in my new setup. Either way, not a big deal.

rogerl50:
> 

Are you looking to have the defaults for `dragLimit` set to a non-empty object, or were you trying to get the various individual limits (e.g. `maxX`) set properly?

client/js/jsonedit.js
flackr: The updated PR looks like it resolves this. Thanks.

flackr (approved):
The code looks good to me.

96LawDawg:
Thanks for the assist with the JSON Editor values @flackr.  This is ready to go.  I will merge in about 24 hours.

rogerl50:
Looks good to me too.

PR #1514, opened 2023-01-07
faster reconnection after lost connection to server

ArnoldSmith86:
When the client loses the connection to the server, it waits for 1s and then tries to reconnect. Everytime it does not work, the wait time is doubled. _And_ it never resets to 1s. So if you have a bad connection you see the connection lost overlay longer and longer each time it happens.

This PR resets the wait time to half a second every time the connection was established successfully.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1514/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1514-pr-test`.

96LawDawg (approved):
Thank you.  Blind approval.  Not sure how to force a test here, but will see results soon in other rooms if it works.

flackr (approved):
Code looks good to me.

ArnoldSmith86:
> Not sure how to force a test here, but will see results soon in other rooms if it works.

Just reconnect your wifi / replug your ethernet...

ArnoldSmith86:
I actually just tried that and it didn't reconnect at all. Same in production though.

96LawDawg:
Same.  And I thought this was going to fix the wait time after a server reboot (new commits in PRs, new merges in production).  Because that is longer than it was before.  But that is probably just the way it is now after the new UI and library.

ArnoldSmith86:
> Because that is longer than it was before.

It is? I don't see a reason why it should. I'll take a look.

96LawDawg:
I just did a test commit in an old PR.  Took about 24 seconds to do it and about 30 to undo it.  I haven't timed the new PR's but it has to be close to a full minute.

RaphaelAlvez:
tested updating the branch to trigger a reconnection. got a very long period with problem 502 so not sure what to think about it

client/js/connection.js
```diff
@@ -81,6 +86,13 @@ function log(str) {
   toServer('trace', str);
 }
 
+setInterval(function() {
+  if(!pongReceived)
+    startWebSocket();
+  pongReceived = false;
+  toServer('ping');
```
flackr, line 93: If this interval function runs before you are connected it will set pongReceived to false, not send the ping, and the next time it runs it will detect that it didn't receive a pong and call startWebSocket which will open a second connection since the previous connection isn't closed.

Instead, you should probably install this interval after setting up the connection or at the very least do nothing if not currently connected.

flackr (changes requested):
See inline code comment, I think the current ping logic can lead to multiple connections.

PR #1524, opened 2023-01-14
Add align properties which sets alignment per grid specification.

### LINKED ISSUE #1462: Grids should have option to be aligned by the corner

flackr:
This adds alignX and alignY properties in a grid specification which specifies the point on a widget to align to the grid point. By default it is now the top left. This also fixes a bug in the existing grid code such that widgets are now aligned correctly by the specified point and uses a fileupdater to ensure no change is observed in existing games. This fixes #1462.

Proposed wiki update:

| Property  | Default value       | Description           | Tutorial              |
| --------  | ------------------- | --------------------- | --------------------- |
| grid      | []                  | This is an array of "snap to grid" definitions. Each object in the array must have _x_ and _y_ (distance between two objects). Example: `[{ "x": 100, "y": 150 }]`<br /><br />Optionally it can contain _minX/Y_ and _maxX/Y_ to limit where the grid takes effect. Example: `[{ "x": 103, "y": 60, "minX": 400, "maxX": 1200, "minY": 300, "maxY": 700}]`<br /><br />It can also have _rotation_ to force the widget's _rotation_ at that snap point and _offsetX/Y_ for offsetting the complete grid.  Example: `[{ "offsetX": 200, "x": 300, "y": 173.2, "rotation": 90 }, { "offsetX": 350, "offsetY": 86.6, "x": 300, "y": 173.2, "rotation": 90 }, … ]`<br /><br /> By default, offset calculations are made from the top left corner of the widget, but you can use _alignX_ and _alignY_ in each grid element to modify that. Example: `[{"x": 100, "y": 100, "alignX": 0.5, "alignY": 0.9}]` where the values for _alignX_ and _alignY_ represent a percentage of the width and height of the widget from 0 to 1.  _offsetX_ and _offsetY_ are expressed in pixels.| [grid](https://virtualtabletop.io/#%7B%0A%20%20%22load%22%3A%20%22https%3A%2F%2Fvirtualtabletop.io%2Flibrary%2FTutorial%2520-%2520Grid.vtt%22%0A%7D), [min/max](https://virtualtabletop.io/#%7B%0A%20%20%22load%22%3A%20%22https%3A%2F%2Fvirtualtabletop.io%2Flibrary%2FTutorial%2520-%2520Grid.vtt%23VTT%2F5prq.json%22%0A%7D)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1524/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1524-pr-test`.

client/js/widgets/widget.js
```diff
@@ -2114,9 +2115,13 @@ export class Widget extends StateManaged {
   }
 
   async snapToGrid() {
+    let align = this.get('gridAlign').split(' ').map(s => parseFloat(s) / 100);
```
flackr, line 2118: I should probably do this step a bit more carefully to ensure that invalid gridAlign properties don't break the client. Right now if the string doesn't contain at least 2 floats it will likely cause an error.

96LawDawg (approved):
Approved through test: grid tutorial, dominoes, crafty, rummy tiles, shape sorter all work.

RaphaelAlvez:
It still doens't work as great if you use grid align inside of grids. lok at the small checker piece
I'm happy with the current setup

RaphaelAlvez:
https://user-images.githubusercontent.com/73476656/212543965-d82336cb-8d04-4a2b-8003-791182dbcd6b.mp4

flackr:
> It still doens't work as great if you use grid align inside of grids.

That's because it doesn't check for gridAlign inside grids at all. Happy to add that, though it also feels weird that you'd have to call it gridAlign when you're already inside of a grid definition so if it's going to be there I'd rather not have the widget property and just have an align property in the grid.

ArnoldSmith86:
Might make sense if it only affects grids. You could also set different values for different grids then.

RaphaelAlvez:
> > It still doens't work as great if you use grid align inside of grids.
> 
> That's because it doesn't check for gridAlign inside grids at all. Happy to add that, though it also feels weird that you'd have to call it gridAlign when you're already inside of a grid definition so if it's going to be there I'd rather not have the widget property and just have an align property in the grid.

in theory, any property can be set on grids. that's why I tested it. This is just a test and I don't have any real use case in mind at the moment

RaphaelAlvez:
well now it doesn't work and it is pushing a custom property to the widget

flackr:
> well now it doesn't work and it is pushing a custom property to the widget

It's working perfectly for me. I realized from your example that we actually push all of the properties of the grid object onto the widget when it snaps so I had to add an exemption for align just as we do for the other grid properties.

flackr:
As a general FYI `align` is now per grid specification and not a defined / used property on the widget itself. It might be more consistent with the other grid properties to actually split it up into alignX and alignY, but given it's so speciifc to grids I like it being in the grid description.

flackr:
Confirmed with @RaphaelAlvez that it is working for them now too.

ArnoldSmith86:
What's the difference between align and offset?

flackr:
align is a proportion of the widget size, so with a single grid definition you can center align arbitrarily sized widgets for example.

flackr:
FWIW, if we just want to change to top-left alignment and forget about the align property that's fine too.

RaphaelAlvez:
I prefer the flexibility of %

96LawDawg:
> As a general FYI `align` is now per grid specification (given it's so speciifc to grids I like it being in the grid description) and not a defined / used property on the widget itself. It might be more consistent with the other grid properties to actually split it up into alignX and alignY.

I like the idea of doing that for consistency.  Whatever we decide to do, also need to add the buttons to the JSON Editor.  I have proposed a wiki change in the original comment.

flackr:
Split it out into alignX and alignY and updated the proposed wiki change to reflect this.

96LawDawg:
> Split it out into alignX and alignY and updated the proposed wiki change to reflect this.

I added the new buttons to the JSON Editor and further tweaked the wiki.  I think this is ready to go.

96LawDawg:
I added a new tutorial as variant 0 to the grid tutorial demonstrating alignX, alignY, offsetX, and offsetY.

server/fileupdater.mjs
```diff
@@ -18,8 +18,13 @@ function updateProperties(properties, v) {
...
 
-  if(!properties.type)
-    updateProperties(properties.faces, v);
+  if(!properties.type) {
+    if (typeof properties.faces == 'object') {
+      for (let face in properties.faces) {
+        updateProperties(properties.faces[face], v);
+      }
+    }
+  }
```
ArnoldSmith86, line 27: Hmm, crap, so this was completely broken since the beginning? :sob:

ArnoldSmith86 (approved):
Code looks good to me.

I didn't look at the tutorial changes but I did test Colonies WIP which is the reason why grids do all that crazy stuff they do. :smile:

server/fileupdater.mjs
flackr: Yup! Crafty exposed that this wasn't actually updating the face properties, heh.

Issue #1202, opened 2022-04-26
`owner` is broken for hands with piles


ArnoldSmith86:
The public library game `Animal Race` has a hand in the top left that has a pile of cards in it.

When dragging a card out of it, its `owner` is not reset so nobody can see it (until it gets dropped into a holder). It does however still work for the last card.

A recent PR changed the order of `parent = null` and `owner = null`. That's probably the cause but I didn't look into it yet.

robartsd:
`Widget.moveStart()` sets `this.currentParent = widgets.get(this.get('parent'))`. Perhaps `this.currentParent` should actually be `widgets.get(this.get('_ancestor'))`.

robartsd:
I'm not finding the PR you mention that changed the order (I'm assuming this is in `Widget.checkParent()` - perhaps I'm not looking for the change you mention in the correct part of the code).

ArnoldSmith86:
I was probably thinking of #1139 . Since that code change is in `MOVE`, it does not seem like it's the culprit.

robartsd:
`currentParent` is used in:

- `Holder.onChildAdd()`
  - tests if holder is already the child's parent
- `Widget.checkParent()`
  - checks `currentParent.get('childrenPerOwner')` (this is where the raised issue has problems)
- - calls `currentParent.dispenseCard()` (`Pile.onChildRemove()` also calls `dispenseCard()` if the pile is in a holder)
  - deletes currentParent
- `Widget.moveToHolder()`
  - sets currentParent if needed
  - calls checkParent(true) if currentParent is not holder
- `Widget.moveStart()`
  - sets currentParent
- `Widget.move()`
  - compares hoverTarget to currentParent

PR #1526, opened 2023-01-14
Set currentParent to grandparent if immediate parent is a pile.

flackr:
Fixes an issue @96LawDawg found with owner not being reset if you move widgets out of a pile within a `childrenPerOwner` holder.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1526/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1526-pr-test`.

96LawDawg (approved):
Approved through test.  I tested the specific behavior that this was supposed to fix.  But since this code is in the routine for most moving, I tested different types of piles in different games.  All works as expected (with no change in behavior for existing games).

ArnoldSmith86:
Shouldn't it be enough to change `parent` to `_ancestor` in the line above your insertion?

96LawDawg:
> Shouldn't it be enough to change `parent` to `_ancestor` in the line above your insertion?

Tested. That works.  Made the change.

flackr (commented):
Yup, didn't know about ancestor but that looks like it does exactly what we want. I would self approve the updated change but github won't allow me :-).

ArnoldSmith86:
I'll approve it after looking at what this might influence.

flackr:
> I'll approve it after looking at what this might influence.

My justification for it being okay to change currentParent is that both the check for `childrenPerOwner` and the call to `dispenseCard` should be called on the holder. The use of `currentParent` in `moveToHolder` also seems to be making an assumption that the parent is the holder and not a pile.

96LawDawg:
This also fixes #1202.  But I can't seem to link that in the development in Git. Lately, it seems like older issues will not link like there is some limit on how far back that filter in Git lets you go.

ArnoldSmith86:
I can usually link it from the issue side because the PR is still new.

96LawDawg:
That is what I just figured out and did.

ArnoldSmith86 (approved):
Yeah, this looks OK to me.

PR #1527, opened 2023-01-14
Fix MOVEXY does not change owner

96LawDawg:
This fix works because MOVEXY will never go to a holder so no need to worry about any of those calls.  I think.

Question I often find myself asking now.  Does this need a file updater?  Theoretically maybe, but MOVEXY has been rarely used (only in 3 PL games) and never from a hand.  I guess there could be a game out there that used this is some way.

Proposed wiki update:

Parameters:
* from: _holderID_ (or an array) - specifies the holder that contains the widgets to move
* ~~collection: _collection_~~ - _not yet supported_
* count: _number_ - limits the amount of moved widgets (defaults to 1 if omitted; use 0 to indicate no limit)
* x: _number_ - the x position on the surface (defaults to 0)
* y: _number_ - the y position on the surface (defaults to 0)
* resetOwner: _true/false_ - specifies whether the _owner_ property of the widget is reset to null (defaults to true)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1527/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1527-pr-test`.

Ingegneus:
I actually use that in citadels 😅 or at least it used to, but I changed it because the owner wasn't updated. the use case was to MOVEXY the entire hand to the floor, so someone else can look at it in a peek holder

96LawDawg:
I was thinking of the other way around, where it works as bugged on production.  But you gave me an idea: you could have a big chooser.  All the cards dealt out on the screen in a big MOVEXY grid that only you can see.  That means I need to work on a file updater.

ArnoldSmith86:
I changed it to only remove owner if the source holder had childrenPerOwner. Please test this.

ArnoldSmith86:
> I was thinking of the other way around, where it works as bugged on production.  But you gave me an idea: you could have a big chooser.  All the cards dealt out on the screen in a big MOVEXY grid that only you can see.  That means I need to work on a file updater.

Can only make a file updater if the behavior is optional though. Unless it GETs the owner before a MOVEXY and the SETs it again afterwards.

96LawDawg:
> Can only make a file updater if the behavior is optional though. Unless it GETs the owner before a MOVEXY and the SETs it again afterwards.

Yeah, that's the idea.  Change the default behavior the way you fixed it (which I will test).  Then if anybody used it this broken way, it will GET, MOVEXY, and SET.

ArnoldSmith86:
There's no way for the file updater to know though. So this would mean wrapping _every_ MOVEXY with SET and GET. Feels cleaner to add a `resetOwner: false` to every MOVEXY.

96LawDawg:
> There's no way for the file updater to know though. So this would mean wrapping _every_ MOVEXY with SET and GET. Feels cleaner to add a `resetOwner: false` to every MOVEXY.

Yes, messy.  There aren't that many MOVEXY's (cool that the new library set-up allows even me to search every PL game, but not every game on the server of course).

I don't understand what you mean.  Add a new property?

96LawDawg:
I think the latest commits do this the way @ArnoldSmith86 suggested, with a file updater.  I've tested the parameter and the file importer and everything seems to work.  If this is acceptable, I will also want to change the tutorial.

client/js/widgets/widget.js
```diff
@@ -1271,14 +1271,16 @@ export class Widget extends StateManaged {
...
-        setDefaults(a, { count: 1, face: null, x: 0, y: 0, snapToGrid: true });
+        setDefaults(a, { count: 1, face: null, x: 0, y: 0, snapToGrid: true, resetOwner: true });
         if(this.isValidID(a.from, problems)) {
           await w(a.from, async source=>{
             for(const c of source.children().slice(0, a.count || 999999).reverse()) {
               if(a.face !== null && c.flip)
                 c.flip(a.face);
               await c.bringToFront();
               await c.setPosition(a.x, a.y, a.z || c.get('z'));
+              if(c.get('_ancestor') && widgets.get(c.get('_ancestor')).get('childrenPerOwner') && a.resetOwner)
```
ArnoldSmith86, line 1282: If it's a parameter, it should probably not check `childrenPerOwner`? I added that initially to keep the impact small if we don't do a file updater.

96LawDawg: I thought about that, but wasn't if you would think that is best.  Fixed.

ArnoldSmith86 (approved):
Code looks fine to me. I didn't test anything though.

Ingegneus:
I'll just copy LawDawg here: approved through test

96LawDawg:
Well, I screwed up something in resolving the conflict with v10 and v11.  Any ideas?

ArnoldSmith86:
Looks like the v10 function is missing its final `}`. Also there's a new newline out of nowhere at the start of the file.

PR #1528, opened 2023-01-14
new game Cards Against Humanity

ArnoldSmith86:
It is a Creative Commons game as long as we stick to the base game.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1528/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1528-pr-test`.

96LawDawg:
Very nice.  Great looking.  Easy to use.  One thing to consider is maybe making the Pick Winner button only visible for the card czar.  Very easy to accidentally press it and then you see the voting screen.  OR, and maybe this is better, have an oops, get me out of here, button on the voting screen.

Also, I see you found a way to use SCORE instead of SET when just using totals by always using round 1.  I hadn't thought of that.  Very nice.  Will have to amend the SCORE tutorial.

ArnoldSmith86:
> Also, I see you found a way to use SCORE instead of SET when just using totals by always using round 1. I hadn't thought of that. Very nice. Will have to amend the SCORE tutorial.

Yeah. It felt weird that it said "Totals" on the scoreboard if it is just the one score.

ArnoldSmith86:
> One thing to consider is maybe making the Pick Winner button only visible for the card czar. Very easy to accidentally press it and then you see the voting screen. OR, and maybe this is better, have an oops, get me out of here, button on the voting screen.

The second option was too much work and I didn't really like the first one. Now I check if the middle holder has one card per player. For prompts with two responses, it doesn't check the right holder. Maybe I'll add that tomorrow.

It's also missing the Gambling that's in the rules where you can submit multiple responses. I don't think I want to add that in the initial PR though.

96LawDawg:
> The second option was too much work and I didn't really like the first one. Now I check if the middle holder has one card per player. For prompts with two responses, it doesn't check the right holder. Maybe I'll add that tomorrow.

I added a check for 2 responses and a prompt before restarting.

RaphaelAlvez:
image

I think this could be improved

96LawDawg:
What looks good when there are a lot, looks pretty bad when there are only a few players.

Here is the css to do it:

"css": {
      ".scoreboard table": {
        "height": "100%"
      },
      ".scoreboard td": {
        "text-align": "center"
      }
    },

The only solution I can think of is a playerGlobalUpdateRoutine that counts that players and adjusts the height of the scoreboard.  That might not be a bad idea.

CAH Many


CAH Few

ArnoldSmith86:
> image
> 
> I think this could be improved

Do you mean that it should fill the scoreboard or something else?

ArnoldSmith86:
Probably best to set a fixed row height no matter how many players so that it's filled with 10 players?

RaphaelAlvez:
> > image
> > I think this could be improved
> 
> Do you mean that it should fill the scoreboard or something else?

change the size of the scoreboard. even it it is not reactive as Lawdawg suggested there is no reason to have such a big scoreboard if you just going to use 50% of it.

I prefer resizing the scoreboard to the number of active players. Maybe when no player is seated show all seat and full size

RaphaelAlvez:
> Probably best to set a fixed row height no matter how many players so that it's filled with 10 players?

that's what I meant in:

>change the size of the scoreboard. even it it is not reactive as Lawdawg suggested there is no reason to have such a big scoreboard if you just going to use 50% of it.

ArnoldSmith86:
> I added a check for 2 responses and a prompt before restarting.

I approve this change!

96LawDawg:
The scoreboard still uses a lot of empty space unless you fill the room.  How about this?

Scoreboard

Demo in the pr-test room using:

```
"css": {
    ".scoreboard": {
      "background": "transparent"
    },
    ".scoreboard div": {
      "border": "none"
    },
    ".scoreboard tbody td": {
      "border-bottom": "1px solid gray"
    },
    " tr": {
      "height": "50px",
      "background": "white"
    }
  },
  ```

96LawDawg (approved):
Approved through test, but I do think the scoreboard would be better with the css I suggested.  Your choice.

RaphaelAlvez:
> The scoreboard still uses a lot of empty space unless you fill the room.  How about this?
> 
> Scoreboard
> 
> Demo in the pr-test room using:
> 
> ```
> "css": {
>     ".scoreboard": {
>       "background": "transparent"
>     },
>     ".scoreboard div": {
>       "border": "none"
>     },
>     ".scoreboard tbody td": {
>       "border-bottom": "1px solid gray"
>     },
>     " tr": {
>       "height": "50px",
>       "background": "white"
>     }
>   },
>   ```

This is a nice trick. Much better then resizing the widget

ArnoldSmith86:
TBH, I prefer it the way it is. It is empty space either way (just a different color) and your way breaks the symmetry (granted, it is broken by the second draw pile already because the room was initially made for Joking Hazard).

96LawDawg:
> TBH, I prefer it the way it is. It is empty space either way (just a different color) and your way breaks the symmetry (granted, it is broken by the second draw pile already because the room was initially made for Joking Hazard).

Doesn't hurt my feelings either way.  And you shouldn't base any decisions on my "feelings" whether they are hurt or not!  Merge this.

ArnoldSmith86:
Would have merged it already if it wasn't for the TestCafé problem.

PR #1529, opened 2023-01-14
new game Found it! similar to Dobble

ArnoldSmith86:
Similar to "Spot it!" or "Dobble".


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1529/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1529-pr-test`.

96LawDawg:
The pile handles for the seats should match the center circle (i.e., a big number in the middle).

Also, getting up or sitting down in a seat resets the game.  Too easy to mess up.  Should probably ask for confirmation.

ArnoldSmith86:
> The pile handles for the seats should match the center circle (i.e., a big number in the middle).

Hmm, it was. I don't see any in the test room. Did anyone mess with it?

I'll investigate further later..

96LawDawg:
One other thought as I went back.  I know you were using the icons from that website that is all animals.  But that may cause confusion.  Some of those animals are hard to tell what they are.  Especially upside down.  And I might shout bird, but there are 5 birds on this card.  What is that pink one?

Dobble

ArnoldSmith86:
Yeah, probably better to spend some time choosing a few other icons. icons8 seems pretty nice btw.

RaphaelAlvez:
looks great. I just don't like the name

ArnoldSmith86:
Suggest a better one then. :laughing:

RaphaelAlvez:
first of all it's you game so you chose the name but here are just some **suggestions**:
- Double
- Spot the Double
- Double match
- Find it!

rogerl50:
It would be cool to let the user click on the matching symbol (but probably not that easy). Otherwise I guess you need voice?

RaphaelAlvez:
> It would be cool to let the user click on the matching symbol (but probably not that easy). Otherwise I guess you need voice?

you could just spam-click on all icons. I prefer the simple way it is now

ArnoldSmith86:
> you could just spam-click on all icons. I prefer the simple way it is now

Well, I would only allow one click. I might want to do it - if only as a demo for the new mouse coords - but maybe in a different PR.

ArnoldSmith86:
Anything you don't like from the image set?

image

RaphaelAlvez:
> Anything you don't like from the image set?

I like this. there are some hard combinations and some easy ones. It feels similar to the original.

RaphaelAlvez:
> image

any reason why dog and ant have a border? can this be a problem?

ArnoldSmith86:
> any reason why dog and ant have a border?

That's just my file browser being weird.

96LawDawg:
Personally, I don't like the compact camera.  Younger people may not even know that is a camera because aren't all cameras on phones?

I would also swap either the knight or the horse for something else.  And I would swap either the sun or the sunflower for something else.  Too similar.

ArnoldSmith86:
Replacing left by right:

image

New set (ignore the folders I used for the previous image):

image

96LawDawg:
Looks good to me.

RaphaelAlvez:
no objection

96LawDawg (approved):
Approved through test.

I recommend changing the first sentence of the game description from "Win cards by spotting identical animals" to "icons" instead of animals.  Merge after change.

robartsd:
I know that the Spot It game that I have (baseball themed) uses various sizes of the images on various cards; not sure if Dobble does the same.

ArnoldSmith86:
It does and so does this. Think it should be more extreme?

robartsd:
I also think that in Spot It, every pair of cards has exactly one image in common. I don't see how that could be the case here with 8 images per card and 57 images total (I don't recall how many images per card or how many cards a Spot It deck has).

ArnoldSmith86:
Test it!

Dobble has 54 cards and 8 symbols. According to some YouTuber it's because it's cheaper to stay under 54 cards. No clue. There are many many videos that explain how Dobble works.

robartsd:
> Test it!
> 
> Dobble has 54 cards and 8 symbols. According to some YouTuber it's because it's cheaper to stay under 54 cards. No clue. There are many many videos that explain how Dobble works.

The article indicates that there can be up to 57 cards with 8 images each such that any pair contains exactly one match. It's rather amazing. The article also indicates that the official game has 55 cards because that is the number of cards packaged in a typical playing card deck (52 card + 2 jokers + an advertisement card). I see your room has all 57 possible cards.

ArnoldSmith86:
You can see the script I used to generate the deck in the misc repo. I simply copy-pasted its contents to the browser dev console.

PR #1531, opened 2023-01-14
Update SCORE Tutorial

96LawDawg:
This updates the first example in the tutorial.  @ArnoldSmith86 came up with a nice technique in Cards Against Humanity for entering scores when only totals are being used and not rounds.  That is better than the SET method in the tutorial.  So I'm updating it.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1531/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1531-pr-test`.

96LawDawg:
This PR also:
 - Fixes broken variant 2 of the Scoreboard tutorial.  Something that changed with the css in the last days of editing that PR broke some of the stuff I did in that variant of the tutorial.  So I simplified it and fixed 2 broken scoreboards
 - Adds Scoreboards and SCORE to TestCafe.

library/tutorials/Functions - SCORE/0.json
```diff
@@ -187,7 +158,13 @@
     "scoreProperty": "scoreDemo1",
     "movable": false,
     "autosizeColumns": false,
-    "playersInColumns": false
+    "playersInColumns": false,
+    "roundLabel": "",
+    "rounds": [
+      "Score"
+    ],
```
RaphaelAlvez, line 165: That was the source of that suggestion I made during the devlopment of scoreboards. I had a scoreboard that did this. I Think you don't need the [ ] and you only need to point to a property on seats that is a number.

RaphaelAlvez (approved):
approved by test and JSON reviewed

library/tutorials/Functions - SCORE/0.json
96LawDawg: True, but if you do that, the width of the score column changes a lot and doesn't present as clean a look.  I'm not sure exactly why, but I'm going to leave it as is.

96LawDawg:
> I also did something in Parade that might be nice to include (maybe that's in the currently lost version). Each seat has its round code (inherited) and the button just CALLs the routine in each seat

I'm not sure what you latest/best copy of Parade is.  If you want to provide another example in the tutorial, that would be great.  But I'm going to merge this now to fix a couple of bugs that this PR also fixes.

PR #1532, opened 2023-01-15
Fix uneditable labels so they scroll

### LINKED ISSUE #1064: Uneditable labels are also not scrollable

96LawDawg:
Fixes #1064 by changing `pointer-events: none` to `outline: none`.  This allows scrolling without allowing the textarea to become the focus and interfere with other actions.

### DIFF
--- a/client/css/widgets/label.css
+++ b/client/css/widgets/label.css
@@ -30,6 +30,6 @@ body.edit .widget.label > * {
 }
 
 .widget.label > textarea:read-only {
-  pointer-events: none;
+  outline: none;
 }
 


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1532/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1532-pr-test`.

96LawDawg:
Tested and works on Chrome, Edge, Firefox, and iOS Safari.

RaphaelAlvez (approved):
Tested on mobile.

I cannot provide a code review because I cannot understand how removing the outline creates a behavior similar to pointer event.

ArnoldSmith86:
> I cannot provide a code review because I cannot understand how removing the outline creates a behavior similar to pointer event.

Yeah, can somebody post an explanation? :smile:

ArnoldSmith86:
Are read-only movable labels still movable?

96LawDawg:
> > I cannot provide a code review because I cannot understand how removing the outline creates a behavior similar to pointer event.
> 
> Yeah, can somebody post an explanation? 😄

Because ChatGPT said so!  It also suggested another alternative is to wrap the textarea inside a div and give the div the overflow:auto, and the textarea pointer-events:none.  But that was more work.

> Are read-only movable labels still movable?

Unfortunately, no.  So we may have to try the alternative.  Also there is this really strange behavior when set to movable:true.  What is this!?!


https://user-images.githubusercontent.com/76912527/212546696-f882b58f-bb35-4a74-aaa9-d2ba0f3b95c2.mp4

96LawDawg:
So I went with the alternative.  This is closer, but still not quite right.  I put overflow:auto on the label div and overflow:hidden on the textarea.  I returned pointer-event: none when read only.  So no you can move a label (when not editable).  And you can scroll.  The problem is if the scrollable area is more than a couple of lines, you can't scroll to see it all.  Maybe somebody knows how to fix that.  I'm out of ideas.  And marking this as draft.

ArnoldSmith86:
Those changes sound like we should first try to merge my fix PR.

ArnoldSmith86:
BTW, moving _and_ scrolling is basically impossible on mobile anyway.

flackr:
> Because ChatGPT said so! It also suggested another alternative is to wrap the textarea inside a div and give the div the overflow:auto, and the textarea pointer-events:none. But that was more work.

ChatGPT is happy to state incorrect things. Setting `outline: none` disables the UA visible outline when the element is focused but does not prevent focus.

> BTW, moving _and_ scrolling is basically impossible on mobile anyway.

Not today, but here's hoping I can make this happen:
https://front-end.social/@flackr/109700238810704187

PR #1535, opened 2023-01-15
fixed weird numbers when spam-clicking star

ArnoldSmith86:
This allows only one click until the client received the server response.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1535/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1535-pr-test`.

RaphaelAlvez:
There is still a temporary number under the start but it only goes from 0 to 1

RaphaelAlvez (approved):
already a better version

There is still a temporary number that appears but that not a big deal

ArnoldSmith86:
Might be the fact that the faked number doesn't remove a zero? I'll change that.

ArnoldSmith86:
Fixed that it would show 0. Can you trigger any more issues?

PR #1538, opened 2023-01-16
Widgets and routines now notice id change

### LINKED ISSUE #1125: changing `id` doesn't update `inheritFrom`, `linkedToSeat`, `onlyVisibleForSeat`
Changing the `id` of a widget _does_ update `parent` of its children and `deck` of its cards.

It could also update the ID in routines but I feel like that's too dangerous.

### LINKED ISSUE #1384: update inheritFrom after id change
when changing the id of a widget, all its childrens parent property gets updated. it would be reeeeaaally neat to have that for inheritFrom as well

rogerl50:
If `a` inherits from `b` and `b`'s `id` changes, the `inheritFrom` property of `a`will now change accordingly.

Should fix #1384. Also fixes #1125 except for updating the altered `id` in routines.

---

Todo:
- [x] Update `${PROPERTY x OF widgetID}` while we're at it? Also stuff like `holder` parameters? And in anonymous collections?
~~- [ ] Fix crash in `leaveRoutine` (https://github.com/ArnoldSmith86/virtualtabletop/pull/1538#issuecomment-2227033423).~~
- [x] Fix non-string `id` (done in another PR)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1538/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1538-pr-test`.

RaphaelAlvez:
I'll not review this PR at this moment. But I saw the title and I was just wondering why not update the id in other places too?

rogerl50:
> I'll not review this PR at this moment. But I saw the title and I was just wondering why not update the id in other places too?

For example, where?

96LawDawg (approved):
Approved through test.  I changed all the widgets in the inheritFrom tutorial and they all worked.  I created a room and manually did as many things as I could think of, including using !, *, and inheriting from multiple widgets.  All worked as expected.

96LawDawg:
> > I'll not review this PR at this moment. But I saw the title and I was just wondering why not update the id in other places too?
> 
> For example, where?

`linkedToSeat`, `onlyVisibleForSeat`, `hoverTarget` are 3 that come to ming.

rogerl50:
See #1125 .

RaphaelAlvez:
> > For example, where?
> 
> `linkedToSeat`, `onlyVisibleForSeat`, `hoverTarget` are 3 that come to mind.

Also deck, hand and parent

> And if you really want to get crazy, every instance of `PROPERTY OF` in routines.

Yeah I agree with @ArnoldSmith86 that this should be more complicated but I think it is doable.:
- The PROPERTY OF is pretty simple.
- Parameters like from, to, holder, canvas and label are always widgets
- collection can can have IDs in pseudo collections but they are only input in some functions. This should still work because I think you cannot use a pseudo collection as output.
- Then value in SELECT, that is used a lot, but it should only be changed if property is one of the properties we already discussed (deck parent, id, hand, linkedToSeat...)

cardType also has a similar problem as id:  Sometime I update the name in the editor and I have to then rename all cards with that type.

Anyway the scope of this PR doesn't need to include all of these problems.

ArnoldSmith86:
Deck and parent should work already.

RaphaelAlvez:
> Deck and parent should work already.

Alright. I didn't test. I was just listing all properties that relate to a widget

rogerl50:
Latest commit handles `linkedToSeat`, `onlyVisibleForSeat`.

Is `hoverTarget` really useful? I thought it was relevant only *during* a drag. What's the scenario where the id of the widget to be dropped upon changes during the drag?

96LawDawg:
> Latest commit handles `linkedToSeat`, `onlyVisibleForSeat`.
> 
> Is handling `hoverTarget` really useful? I thought it was relevant only _during_ a drag. What's the scenario where the id of the widget to be dropped upon changes during the drag?

Good point.  I was just thinking of properties that take an `id`.

RaphaelAlvez:
Hand in seats is still missing

client/js/editmode.js
```diff
@@ -1557,6 +1543,47 @@ async function updateWidget(currentState, oldState, applyChangesFromUI) {
...
+
+  // Restore children
+  for(const child of children)
+    sendPropertyUpdate(child.get('id'), 'parent', id);
+  for(const card of cards)
+    sendPropertyUpdate(card.get('id'), 'deck', id);
+
+  // Change inheritFrom on widgets inheriting from altered widget
+  StateManaged.inheritFromMapping[id] = StateManaged.inheritFromMapping[previousState.id];
+  delete StateManaged.inheritFromMapping[previousState.id];
```
ArnoldSmith86, line 1566: Is this necessary? Those should be updated when receiving the delta..

client/js/editmode.js
```diff
@@ -1557,6 +1543,47 @@ async function updateWidget(currentState, oldState, applyChangesFromUI) {
...
+      newInherits[id] = newInherits[previousState.id];
+      delete newInherits[previousState.id]
+    }
+    inheritor.set('inheritFrom', newInherits)
+  };
+
+  // If widget is a seat, change widgets with onlyVisibleForSeat and linkedToSeat naming that seat.
+  if(widget.type == 'seat') {
+    widgetFilter(w => w.get('onlyVisibleForSeat') == previousState.id).map( w => w.set('onlyVisibleForSeat', id));
+    widgetFilter(w => w.get('linkedToSeat') == previousState.id).map( w => w.set('linkedToSeat', id));
```
ArnoldSmith86, line 1583: Both of those can be arrays.

rogerl50:
> Hand in seats is still missing. that might be the last property that fits this PR

To determine if the widget whose id has changed is a hand, do you think I should just check to see if it's a holder with `childrenPerOwner` set? (The alternative, I guess, is to make no assumptions and just change the `hand` property of any seat when the id of the widget it points to changes, and not worry about what kind of widget it was.)

ArnoldSmith86 (changes requested):
This should be tested thoroughly with `SET` and id. Otherwise the code is only called from edit mode.

I'd also like `hand`. And I wouldn't check anything. Just replace it. I used `hand` to point to a normal holder in `Found it!` for example.

RaphaelAlvez:
> > Hand in seats is still missing. that might be the last property that fits this PR
> 
> To determine if the widget whose id has changed is a hand, do you think I should just check to see if it's a holder with `childrenPerOwner` set? (The alternative, I guess, is to make no assumptions and just change the `hand` property of any seat when the id of the widget it points to changes, and not worry about what kind of widget it was.)

No limitations. Hand is designed to be any kind of holder but it doesn't really matter if it is not even a holder.
Same for the other properties. The type or properties of the widget assigned should not matter. If th id changes then the property should do the same.

client/js/editmode.js
rogerl50: I removed them, but see issue #1540 .

client/js/editmode.js
rogerl50: Fixed.

rogerl50:
See the _seat_ and _inherit_ rooms in the pr-test area for some tests for those two dependencies.

RaphaelAlvez:
Sorry I pressed some buttons by accident

96LawDawg (changes requested):
Using SET to rename a widget that is being inherited from freezes the client.  Error is "TypeError: Cannot read properties of undefined (reading 'get')."  Line 2039 of jsonedit.js.

96LawDawg:
I get the same error when using SET on onlyVisibleForSeat and linkedToSeat.  Those errors (and the error with inheritFrom) do not happen on production since there is no effort on production to actually change the subordinate widget.

ArnoldSmith86:
Should this even be changed for SET? Maybe someone wanted to disable the connections on purpose.

rogerl50:
> Should this even be changed for SET? Maybe someone wanted to disable the connections on purpose.

Shouldn't that be done by editing/removing the `inheritFrom` attributes, though?

ArnoldSmith86:
Who's gonna enforce that?

rogerl50:
> Who's gonna enforce that?

Well...we are, if we change the SET behavior. But I'm happy either way - we could also leave SET so that it doesn't modify any inheritance attributes. In either case we will have some work to do in SET, and I haven't really thought through how much work is required with either option.

96LawDawg:
> Well...we are, if we change the SET behavior. But I'm happy either way - we could also leave SET so that it doesn't modify any inheritance attributes. In either case we will have some work to do in SET, and I haven't really thought through how much work is required with either option.

Could we just catch the problem, report it in the debugger, and leave it as is?  Why are people changing id's with SET anyway.

rogerl50:
> 
> Could we just catch the problem, report it in the debugger, and leave it as is? Why are people changing id's with SET anyway.

I don't know, that sounds like it would be too easy 🤣 . So we would just outlaw changing the `id` using SET.

ArnoldSmith86:
But it's working in production, isn't it? Can't change that...

96LawDawg:
On production, you can change an id using SET with no problem.  Linkages between widgets are not updated, of course.  This PR changes that behavior.  Currently any id change using SET freezes the client.  So something has to be done to return to what we have in production.  I don't have a solution.

rogerl50:
> On production, you can change an id using SET with no problem. Linkages between widgets are not updated, of course. This PR changes that behavior. Currently any id change using SET freezes the client. So something has to be done to return to what we have in production. I don't have a solution.

Right. The summary in my previous comment would return us to the current behavior (which, again, is not documented).

robartsd:
Behavior in production is buggy. If the widget being inherited from has the id changed, the DOM is not updated to reflect the changes to the inherited properties; however, refreshing the room causes the DOM to update.

rogerl50:
The first of these is exactly what the code that is in this PR right now does (it updates `inheritFrom` to the new `id` and keeps the `inheritFromMapping` up to date). The problem is that it might change behavior of existing code, so is not bug-for-bug compatible. (I don't personally agree with this point of view, but that's where we are.)

robartsd:
> The first of these is exactly what the code that is in this PR right now does (it updates `inheritFrom` to the new `id` and keeps the `inheritFromMapping` up to date). The problem is that it might change behavior of existing code, so is not bug-for-bug compatible. (I don't personally agree with this point of view, but that's where we are.)

Not sure how the bug should be fixed (for existing games) but in addition to the rule not to break existing games, we have the rule that a simple refresh should not change anything.

The last option is the most compatible with existing games (state 100% the same as production, but DOM rendering updated to be consistent).

ArnoldSmith86:
I don't know. It's unlikely that many games rely on that stuff so I don't mind if we make SET update all the things.

rogerl50:
I'm happy to keep working on this, but it doesn't look like there is consensus as to how it should behave, either in general or specifically with `SET`. I don't plan on making further attempts until there is...

robartsd:
> I don't know. It's unlikely that many games rely on that stuff so I don't mind if we make SET update all the things.

I agree that it is unlikely that games rely on SET id not updating references in other widgets. I'm not particularly worried about making a change here; just want to be sure that the change has been considered.

rogerl50:
Is there something wrong with lines 1554-1556 of `updateWidgetId`? (This is a new routine that was carved out of `updateWidget`). Everything executes correctly, yet after the call to `removeWidgetLocal` at 1554, when called from `SET` control immediately returns to the `SET` code, right after the call to `onClickUpdateWidget`. When the code is called from the JSON editor via a widget edit, things work the way I would expect them to work.

ArnoldSmith86:
Sorry that this was forgotten for so long. I hope I brought it into the present again. One main thing is that the editor code is no longer available in play mode so the new function had to be moved. And it was missing a few `await`s for the `set`s (not sure if those were a thing already when this PR started).

I noticed that using `SET` to change `id` is actually broken in production because it relies on editor code already which doesn't work anymore.

All of this should probably be tested pretty thoroughly again.

96LawDawg:
> I noticed that using `SET` to change `id` is actually broken in production because it relies on editor code already which doesn't work anymore.

If that's the case, then most of the earlier discussion is moot.

> All of this should probably be tested pretty thoroughly again.

Yes.  I'll try to work on that sometime, but it might not be soon.

96LawDawg:
Looks like mostly everything is working right.  I did manage to freeze the client with one absurd test.  I put an onLeaveRoutine on a holder that changed the id of the holder.  VTT didn't like that, but I'll keep testing things:

Screenshot 2024-07-13 125912

96LawDawg:
And do we want to include "PROPERTY x OF widgetID"?  It would be good to have, but not if it is not easy to code or if it would touch the deep magic.

ArnoldSmith86:
> And do we want to include "PROPERTY x OF widgetID"?  It would be good to have, but not if it is not easy to code or if it would touch the deep magic.

Shouldn't be too hard. Maybe I'll get around to it but when it works, we should merge it.

96LawDawg:
> > And do we want to include "PROPERTY x OF widgetID"?  It would be good to have, but not if it is not easy to code or if it would touch the deep magic.
> 
> Shouldn't be too hard. Maybe I'll get around to it but when it works, we should merge it.

So is this something on the to do list? Or are we not going to worry about it now and I should test things again?

ArnoldSmith86:
> So is this something on the to do list? Or are we not going to worry about it now and I should test things again?

I think we all agree that it would be super convenient if it updated those things. Let's just do it.

ArnoldSmith86:
> I put an onLeaveRoutine on a holder that changed the id of the holder.

I couldn't reproduce this. Can you?

Other than that, this should now update references in routines as well.

96LawDawg:
> > I put an onLeaveRoutine on a holder that changed the id of the holder.
> 
> I couldn't reproduce this. Can you?

No, but I did find another issue that is probably unrelated and I'll open a new issue for.

> Other than that, this should now update references in routines as well.

I will do some more testing.

96LawDawg:
Well, I guess the problem is related somehow because on production, SETting id is still different.  So here it is. In a room with a holder this:

```JSON
{
      "func": "SELECT",
      "type": "holder",
      "property": "id",
      "relation": "!=",
      "value": null
    },
    {
      "func": "SET",
      "property": "id",
      "value": 1234
    }
  ```
    Gives this error

Screenshot 2024-09-14 084023

Same error if `value` is a numerical variable.  But if value is a string like "1234" then it works.

ArnoldSmith86:
Ah, the infamous non-string id. Good catch!

96LawDawg:
> Ah, the infamous non-string id. Good catch!

Probably the same error catching will work, but arrays, objects, etc. also do not work.  Probably obvious, but I tested to make sure.

ArnoldSmith86:
I'm pretty sure that nothing in this PR has problems with numeric `id`.

In production you can't `SET` `id` if I remember correctly because a function that used to be available is now only available in the editor.

_But_ if you simply change an `id` to a numeric value in production, you get some crashes. While investigating, I found a couple of other places that assume a property is a certain type. I'll make a PR that fixes all of those.

Regarding `id`s, I gotta think a bit about what non-string `id`s even mean. Because in the state JSON, the `id` occurs twice: as the property and as the key of the entire room object. A key in JSON can _only_ be a string. So at least you can never have widgets `5` and `"5"` in the same room.

Anyway, this should be unrelated to this PR. Even though it changes one crash in production into a different one. :grin:

96LawDawg:
> I'm pretty sure that nothing in this PR has problems with numeric `id`.

I agree. 

> In production you can't `SET` `id` if I remember correctly because a function that used to be available is now only available in the editor.

I mentioned the issue here because I could test it here using SET.  Can't do that on production.

> _But_ if you simply change an `id` to a numeric value in production, you get some crashes.

And your `id` PR fixes that, but I would still like to test it using SET once this PR is live since using SET and manually typing in the editor do not always have the same effect.

96LawDawg (approved):
I added SCORE/seats to the list of functions that should be updated.  Tested everything again as much as possible.  Approved.

PR #1539, opened 2023-01-16
Add Senet to public library

96LawDawg:


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1539/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1539-pr-test`.

RaphaelAlvez:
It actually looks extremely beautiful. Very good taste ok this one.

ArnoldSmith86 (approved):
Looks really nice.

I struggled to understand some of the rules. Somehow "Choose a starting player and token.  Cubes go first." didn't tell me that one player uses the cubes and the other one the pawns. But when continue reading nothing else made sense so it was fine.

Typo "concetp".

Is it on purpose that you include double spaces in-between some sentences? I think I noticed that in other games as well.

"If attacked, the opponents piece is moved to the square with an X." - Had to read this three times. :smile: I think I would remove "opponents".

"If the piece moves onto the X while the player has the sticks, the player keeps rolling until getting a 4 and moving the piece off, or until rolling a 2 or 3." - So, a 4 takes it off the board and 2/3 does nothing and ends your turn?

96LawDawg:
> I struggled to understand some of the rules. Somehow "Choose a starting player and token. Cubes go first." didn't tell me that one player uses the cubes and the other one the pawns. But when continue reading nothing else made sense so it was fine.

Fixed.

> Typo "concetp".

Fixed.

> Is it on purpose that you include double spaces in-between some sentences? I think I noticed that in other games as well.

Yes and no.  2 spaces is how Americans of my age group were taught to type.  That is normal in my writing.  On VTT, I try to change it to save space and because 2 spaces don't get saved in many places.

> "If attacked, the opponents piece is moved to the square with an X." - Had to read this three times. 😄 I think I would remove "opponents".

Fixed.

> "If the piece moves onto the X while the player has the sticks, the player keeps rolling until getting a 4 and moving the piece off, or until rolling a 2 or 3." - So, a 4 takes it off the board and 2/3 does nothing and ends your turn?

Yes.  Made it more clear.

> Looks really nice.

This is why I will never write rules for another game.  Because you would have stopped after the first sentence.

ArnoldSmith86:
Aww.

ArnoldSmith86:
Looks really nice.

robartsd:
> Yes and no. 2 spaces is how Americans of my age group were taught to type. That is normal in my writing. On VTT, I try to change it to save space and because 2 spaces don't get saved in many places.

2 spaces after a sentence was a typing convention on typewriters (it is how I learned too). I'm not sure when the double space got dropped exactly, but by the time graphical WYSIWYG text editors were the norm the convention was changed to a single space. The old convention created wide spaces between sentences (three full width characters) but when new technology made that space smaller anyway (`.`,`!` and ` ` are typically narrower than the average character in proportional fonts) the extra space got removed. Perhaps spaces are so narrow now that the extra space is no longer enough to make it easy to notice the end of a sentence.

library/games/Senet/0.json
```diff
@@ -0,0 +1,1346 @@
...
+        "loopRoutine": [
+          "var randomClick = randInt 1 2",
+          "var randomRot = randInt 0 359",
+          {
+            "func": "CLICK",
+            "collection": [
+              "${value}"
+            ],
+            "count": "${randomClick}"
+          },
```
RaphaelAlvez, line 822: "count": "${randomClick}" should have no effect. only the final face will be displayed

library/games/Senet/0.json
```diff
@@ -0,0 +1,1346 @@
...
+    "z": 1,
+    "css": {
+      "background": "#fff"
+    },
+    "faces": [
+      {
+        "css": {
+          "background": "#444",
+          "border": "3px solid #fff",
+          "rotation": 90,
```
RaphaelAlvez, line 654: I believe this doesn't have any effect. also, the whole face setup could be simplified if you just used a variable for the background color and only changed that in the face

library/games/Senet/0.json
96LawDawg: The count is necessary in order to click the face once to change the color or twice to leave it as is.

library/games/Senet/0.json
96LawDawg: The rotation was a leftover accident from testing.  Removed that.  I don't think the css could be simplified enough to make it worth changing, so leaving that as is.

PR #1541, opened 2023-01-18
Make the two recall icons more distintive from their counter parts

tsankuanglee:
The original two recall icons (`recall.svg` and `recall_dice.svg`) look too similar to `deal_face_up.svg` and `deal_face_down.svg`, making it difficult to tell them apart when they are small. Here we flip these icons horizontally so they are distinctive enough. The two new icons also follow more intuitive left-to-right flow.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1541/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1541-pr-test`.

RaphaelAlvez:
Oh you just flipped them. Yeah recently I got the same view about them. They should be rotated

96LawDawg:
Thank you for contributing @tsankuanglee.  I guess I don't really have a problem either way, so if you think this is better, I'm fine with that.  However, what you changed only changes the asset.  It doesn't change the font.  So in the attached, you can see the old way on top and the new way on bottom.  But I made the old way using the font.

If you want to take a stab at changing the font, say so.  Otherwise, if other people agree with the change, then I can do the font.  It is sort of easy once you know what you are doing, but the learning curve for working with fonts is steep and painful.

Recall Icons

RaphaelAlvez:
Actually I'm not sure because I can also understand the idea of going back from right to left.

Probably there is no correct answer here and we should go for a vote

RaphaelAlvez:
This is how the test room looks in the main server


https://virtualtabletop.io/pr1541

96LawDawg:
Thanks for the comparison to what currently happens.  Also, an additional consideration is that it's possible to get the rotated look simply by rotating the image 180 degrees like this (on production).

Recall Icons

However, that is not the perfect solution if you have multiple icons on the same widget.

tsankuanglee:
Thanks, everyone! [96LawDawg](https://github.com/96LawDawg) please feel free to do the font, and possibly also fix the alignment. I'll just be happy to see this done.

tsankuanglee:
@RaphaelAlvez I actually flipped both globally and the individual elements, just to keep the order of those two die (one point on top and two point on bottom). Rotation, as shown in the screenshots posted by @96LawDawg , will change the order.

Either way works for me -- especially I'm already having trouble tell deal-* and recall-* apart.

robartsd:
We could also keep the current recall icons, and name the flipped icons something different (perhaps gather) so that both options are available in the font.

tsankuanglee:
> We could also keep the current recall icons, and name the flipped icons something different (perhaps gather) so that both options are available in the font.

That's a great idea, @robartsd !

96LawDawg:
Here are the revised font files.  You will need to rename them to VTT-Symbols.woff and VTT-Symbols.woff2 (git won't upload .woff files, so just change the extension; they are not actually zipped).

Then upload these into the assets/fonts folder and push the commit.  I would do it for you, but I've found it is too much trouble for me to work in other people's branches.

[VTT-Symbols.1.zip](https://github.com/ArnoldSmith86/virtualtabletop/files/10484195/VTT-Symbols.1.zip)



[VTT-Symbols.2.zip](https://github.com/ArnoldSmith86/virtualtabletop/files/10484196/VTT-Symbols.2.zip)

tsankuanglee:
@96LawDawg, thanks again for the font files. I made slight modification to match my original design, which maintains the orientations of the card/dice, so they match other glyphs of the same font.

image

Specifically, the 2-point dice has dots in the upper-left and lower-right corners, like all other dice in the same font. And the shades of the card are placed on upper-left and lower-right, also like all the other cards in the same font.

I then exported those two glyphs into SVGs so they are now identical.

96LawDawg:
Still a little confused.  Is there something I still need to do?  Font work isn't exactly easy for me.

RaphaelAlvez:
@mousewax any input on this?

mousewax:
No comments. I probably would not have gone to the trouble of maintaining the dice orientation and just rotated the icon 180° in the font and added `_reversed` to the glyph name.

96LawDawg:
In the font, I did a flip but not a 180 rotation.

tsankuanglee:
Just to clarify, the lastest commit includes all the fixes (maintaining the orientation of all elements) I mentioned. @mousewax @96LawDawg This PR is ready to go.

96LawDawg (approved):

PR #1543, opened 2023-01-19
fix sort logging; updateAfterShuffle for collections

tsankuanglee:
two small bug fixes:

* add updateAfterShuffle when shuffling a collection
* sort-logging only when holder is valid

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -1502,6 +1502,10 @@ export class Widget extends StateManaged {
         } else if(collection = getCollection(a.collection)) {
           if(collections[collection].length) {
             await shuffleWidgets(collections[collection]);
+            await w(collections[collection].map(i=>i.get('parent')), async holder=>{
+              if(holder.get('type') == 'holder')
+                await holder.updateAfterShuffle();
+            });
           } else {
             problems.push(`Collection ${a.collection} is empty.`);
           }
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1543/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1543-pr-test`.

client/js/widgets/widget.js
```diff
@@ -1504,6 +1504,10 @@ export class Widget extends StateManaged {
         } else if(collection = getCollection(a.collection)) {
           if(collections[collection].length) {
             await shuffleWidgets(collections[collection]);
+            await w(collections[collection].map(i=>i.get('parent')), async holder=>{
+              if(holder.get('type') == 'holder')
+                await holder.updateAfterShuffle();
+            });
```
ArnoldSmith86, line 1510: This sounds like a nice feature but I don't think we can just change the behavior like this for existing games. I have no idea though why this code is in `SORT` but not in `SHUFFLE`.

Maybe we should make it optional in both with the same parameter?

tsankuanglee: Good idea. Will do, @ArnoldSmith86 !

RaphaelAlvez: Are your shuffling the holder or are you just updating the holder in case at least one card in in a holder?

tsankuanglee: @RaphaelAlvez I should've provided a bit more background info. I was trying to give each player's HAND a shuffle button, but it doesn't update the shuffled result. @ArnoldSmith86 suggested [a workaround](https://virtualtabletop.io/l67n) to sort it with an empty key right after shuffle, just to trigger that `updateAfterShuffle` in SORT, which for some reason is not in SHUFFLE.
 
This PR adds that back in.

client/js/widgets/widget.js
```diff
@@ -1528,9 +1532,9 @@ export class Widget extends StateManaged {
               if(holder.get('type') == 'holder')
                 await holder.updateAfterShuffle();
             });
+            if(jeRoutineLogging)
+              jeLoggingRoutineOperationSummary(`widgets in '${a.holder}' by ${key}${reverse}`);
           }
-          if(jeRoutineLogging)
-            jeLoggingRoutineOperationSummary(`widgets in '${a.holder}' by ${key}${reverse}`);
```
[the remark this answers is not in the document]
tsankuanglee: You are right, @ArnoldSmith86 . This doesn't fix it. I'll remove it.

PR #1548, opened 2023-01-20
Fix arguments to cssAsText from other callsites.

flackr:
When the extra usedProperties argument was added to cssAsText a few callsites went overlooked, fixing these to behave as they used to.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1548/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1548-pr-test`.

flackr:
Note that technically we should track the used properties in all of these callsites so that we invalidate the appropriate things when they change - however the old code used to overwrite the used properties whenever the `css()` function was called so wouldn't have done this correctly either.

ArnoldSmith86:
This actually fixes Citadels. The broken version is in https://virtualtabletop.io/Citadels (notice the blue buttons) and the same version loaded in pr-test looks right again.

ArnoldSmith86 (approved):
Code looks good. I didn't check if you missed any but I'll assume you were thorough...

ArnoldSmith86:
Merging this now because it fixes broken games.

PR #1550, opened 2023-01-20
Fixes #1549 by comparing the transformed offset against the transfomed corner.

### LINKED ISSUE #1549: Broken movement during transition
Some widgets with existing css transition delays do not move right.  For example, the dice in Stardust Butterfly.  Suspected cause is #1421.

flackr:
Currently the move code transforms the offset but assumes that the widget corner is at its x and y coordinates. This isn't true with complex transforms or if the element is in the middle of a transition. By running both the localAnchor and the top-left corner of the widget through coordParentFromCoordLocal they are in the same transformed space.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1550/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1550-pr-test`.

96LawDawg (approved):
I tested this against Stardust Butterfly and dragging the dice works correctly again.  I checked this against both of the new CSS tutorials (for transition and transformation).  Again, all in order.  Through ignorance, I don't know what else I would test.  Approved.

ArnoldSmith86 (approved):
Code looks like it might fix the problem :smile:

PR #1552, opened 2023-01-20
fixed loading old share links

ArnoldSmith86:
The link in https://virtualtabletop.io/Citadels was broken because it referenced a different state ID in the VTT file. This PR should fix that the content of a shared game would return the wrong state ID and on top it makes it load a different state ID from a link if there is only one in the file (which is the normal case unless it is a VTTC file).

### DIFF
--- a/server/fileloader.mjs
+++ b/server/fileloader.mjs
@@ -80,7 +80,7 @@ function checkForLinkToOwnServer(link) {
     const m = sharedLinks[match[1]].split('/');
 
     const states = {};
-    states['VTT'] = [];
+    states[m[3]] = [];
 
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
Don't ask.

ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1552/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1552-pr-test`.

RaphaelAlvez (approved):
Blind approval

PR #1556, opened 2023-01-20
adds cursors

RaphaelAlvez:
I'm making a new PR because:
- classes are useful in many ways
- the PR adds assets and not really any new feature


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1556/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1556-pr-test`.

96LawDawg (approved):
Remove the unnecessary ones we discussed at the meeting and this is good to go.

ArnoldSmith86:
How do you define the "cursor pixel" in the SVG? Because the paintbrush paints more to the right than expected.

RaphaelAlvez:
> How do you define the "cursor pixel" in the SVG? Because the paintbrush paints more to the right than expected.

It is not defined. Maybe I need to enforce a value.

For other widgets I did set them and they look like this in the css
`url('/i/cursors/grabbing.svg') 9 3, auto`

Also keep in mind that this is a assets PR but I can update the tutorial

ArnoldSmith86:
Ah, I see. If they are not defined in the asset, it's fine.

PR #1558, opened 2023-01-22
Add 3 new capabilities involving sound

### LINKED ISSUE #1333: Repeat Sound and Add to Timer
@mousewax would like to be able to repeat a sound so that when it stops playing, you can play it again automatically.

He would also like a property added to timers that would automatically play a sound at the end of the timer (without having to invoke a millisecondsChangeRoutine).

96LawDawg:
Closes #1333.

Demo in https://test.virtualtabletop.io/PR-1558/pr-test.

In the `AUDIO` function, a new parameter _count_ is added.  It defaults to 1.  Any value larger than 1 will repeat the sound n times. If the length of the sound file is set, _count_ does not get implemented.  I suppose it would be possible to do that, but I didn't try.  If set to "repeat", the sound will play a million times and you should be banished to hell for doing such a thing, but it is consistent with other "count" parameters.

`AUDIO` also gets a boolean _silenceAll_ parameter.  Defaults to false.  If true, it stops/pauses all currently playing audio.

In the TIMER widget, a new property _alertSound_ is added.  It defaults to null.  A new property _alertSoundCount_ is also added that defaults to 1. When _alertSound_ is a valid sound file and an _end_ value is set, the timer will play the sound file when the alert activates.  What I changed seems to me (and may be) more complicated than needed.  The problem I was having was the sound would either play repeatedly after the timer reached the alert. Of if you paused it after alert and then restarted, the sound would play again. I didn't like those behaviors and come up with the solution you see.  _alertSoundCount_ does not have a "repeat" option because I couldn't figure it out.  Just set it to a big number.

Note: The repeat `AUDIO` behavior is currently not possible without knowing the length of the sound file and using a timer and repeating the `AUDIO` function n number of times. On the other hand, the `alertSound` capability is pretty easy to add in JSON using a `millisecondsChangeRoutine`.  So maybe this is unnecessary, but I feel like it is a good addition to timers.

To do:

- [x] Add ability to stop all sounds
- ~[ ] _alertSoundCount_ needs repeat option.  NOT PLANNED.~
- ~[ ] Allow playerName or seat-specific alert sounds.  Future PR.~

---------------------
Proposed wiki edits:

New `AUDIO` parameters:

- count: _number_ - the number of times the sound clip will play. Defaults to 1. "Repeat" plays an infinite loop. If _length_ is anything other than null, the _count_ parameter is ignored.
- silenceAll _true/false_ - When true, stops all sound playing in the room. Defaults to false.

--------------------

New timer widget properties

| alertSound | null               | When set to a valid sound file and when _end_ is set, will play a sound the number of times specified in _alertSoundCount_ when the end time has been reached.|
| alertSoundCount | 1               | The number of times the _alertSound_ plays when the end time has been reached. |


RaphaelAlvez:
I think you need to check your timer logic. In timers you can have any end time so 0 can be the end point.

This should be reset if alert is turned off. 

For AUDIO, maybe you should use count as it is already used everywhere and maybe you should add "repeat" or "loop" for 999999

Also maybe you want to also add turn alert

96LawDawg:
> I think you need to check your timer logic. In timers you can have any end time so 0 can be the end point.

Thanks for the feedback.  I will work on those things except for the first one.  It does work that way: countup, countdown, any numerical end value.

96LawDawg:
> I think you need to check your timer logic. In timers you can have any end time so 0 can be the end point.

This worked before and still works.

> This should be reset if alert is turned off.

Can you clarify what you mean?  In production, if you specify an end value, you get the alert version of the css.  That stays until you press reset.  You can programmatically set alert: false in a timer while the milliseconds value is beyond the end value, but as soon as the timer moves again, you get the alert.  In this PR, if you press the reset button, everything resets and you can play the sound again.  So you saying that programmatically changing the alert value should to what?  Stop the sound?

I think maybe we need to modify AUDIO to have a paramater to stop all sounds currently playing.  Right now, I don't think it is possible to shut down a currently playing sound except by F5 or reloading the room.

> For AUDIO, maybe you should use count as it is already used everywhere and maybe you should add "repeat" or "loop" for 999999

Done.  But I don't know who would put a sound on an infinite loop.  I'm not playing that game.
 
> Also maybe you want to also add turn alert to seats. If possible just played to the correct player

Beyond the scope of this PR.  AUDIO needs to accept an array of players first.  I tried messing with that and it was too hard.

96LawDawg:
I added the ability to stop all currently playing sound, which would be nice if you had a buzzer on repeat until someone resets it because now there is a way to reset it and shut it off.

I think this is ready for review.

96LawDawg:
> 



> is the stop function supposed to be for just one client?

No. All of them. Stop all sound.

RaphaelAlvez (changes requested):
Stop does not work as intended. It just stops for the client triggering the function

96LawDawg:
The change I just made works, kind of.  It works once the 2 clients are in synch with each other.  That isn't a precise term here because I don't know what is happening.  If you load 2 separate clients from a fresh state (or press F5) and play the sound on one, stopping it using the other one does not work. But it does work to stop the sound for that client.  Then, after that it works fine where one client can start and the other stop.  Any ideas how to fix that?  The audio stuff is strange.

RaphaelAlvez:
@bjalder26 @ArnoldSmith86 do you have any idea on how to solve this?

bjalder26:
> The change I just made works, kind of. It works once the 2 clients are in synch with each other. That isn't a precise term here because I don't know what is happening. If you load 2 separate clients from a fresh state (or press F5) and play the sound on one, stopping it using the other one does not work. But it does work to stop the sound for that client. Then, after that it works fine where one client can start and the other stop. Any ideas how to fix that? The audio stuff is strange.

image

It looks like it is a browser security issue.  It seems like they increased browser security.  It isn't playing because you didn't interact with the screen yet.

These browser security issues is also why there are issues with ios playing sounds.

Previously I was working on an update where instead of using an audio element, it used audiocontext.  If I remember correctly, the idea was to default the volume to 0 (or off), then when players turned on the volume, all the audio files referenced in the game would be loaded into the audiocontext in a paused state, and played at the appropriate time.

If I could promise to get to that update soon, then I would say just to wait and we'll do this all at once, but I'm teaching 7 in-person courses, and 5 online this semester, so it is unlikely (I think about it time to time though).

I don't think this is a new issue for this PR.  I think it might be because of browser updates, and only noticed now because of this PR, so I don't think it should hold this PR up.

For non-apple products I think once they click or touch the screen, there won't be this issue anymore, and we just find it testing because we aren't actually playing a game.  For apple products, I think we'll need the audiocontext update for audio to work at all.

I do have one question though.

With the timer, should the sound mimic the visual alarm?  Should it turn off at the same time the flashing stops, or should it be turned off according to its own settings?

96LawDawg:
> but I'm teaching 7 in-person courses, and 5 online this semester

You are insane.  I have 7 courses total and that is too many.

> With the timer, should the sound mimic the visual alarm? Should it turn off at the same time the flashing stops, or should it be turned off according to its own settings?

I personally would let the sound be turned off by its own settings.  Once the timer reaches the end, the sound may last for a second or a minute.  But the visual part can do whatever it needs to do until the timer is stopped.  And, in theory (if the sound stop thing works as intended), whatever routine is used to stop the timer could also be used to stop the sound, if desired.

96LawDawg:
Closing this.  Everything in here needs to be completely rethought after #2328.

Issue #201, opened 2021-02-05
Increase snap radius for stacking cards


Ingegneus:
the title should be self explainatory

ArnoldSmith86:
It's possible to change it with #147.

PR #1562, opened 2023-01-24
Add parameter to control snap distance when cards make piles

96LawDawg:
This would implement #201 if it worked right.  Requesting help.

I could get this to work if I just created a property on cards.  But it seems to make sense to add this to the deck's cardDefaults onPileCreation settings.  That is what I'm having trouble with.

Here is the issue.  In pile.js I added a new default of snapRadius: 10 (it is currently 10px). And I got that to apply to the pile, so it shows up like handleCSS, handleSize, etc.

But in widget.js, around line 2226, I can't figure out a way to read the property from the handle.snapRadius.  Pile.js uses a constructor with this.handle.xxx format and widget.js doesn't.  I've tried every combination of this.get and widget.get I can think of.  I tried exporting and importing.  Somebody knows an easy fix for this.  Help.

By the way, the code in this PR currently forces a snapRadius of 50, just so you can see how it works different.  That line will obviously be removed.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -2222,9 +2222,8 @@ export class Widget extends StateManaged {
       if(widgetType != 'card' && widgetType != 'pile')
         continue;
 
-      // check if this widget is closer than 10px from another widget in the same parent
-      let snapRadius = 50;
-      if(widget.get('parent') == thisParent && Math.abs(widget.get('x')-thisX) < snapRadius && Math.abs(widget.get('y')-thisY) < snapRadius) {
+      // check if this widget is closer than the pileSnapRange from another widget in the same parent
+      if(widget.get('parent') == thisParent && Math.abs(widget.get('x')-thisX) < this.get('pileSnapRange') && Math.abs(widget.get('y')-thisY) < this.get('pileSnapRange')) {
         if(widget.isBeingRemoved || widget.get('owner') !== thisOwner || JSON.stringify(widget.get('onPileCreation')) !== thisOnPileCreation)
           continue;
 
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -2223,7 +2223,8 @@ export class Widget extends StateManaged {
         continue;
 
       // check if this widget is closer than 10px from another widget in the same parent
-      if(widget.get('parent') == thisParent && Math.abs(widget.get('x')-thisX) < 10 && Math.abs(widget.get('y')-thisY) < 10) {
+      let snapRadius = 50;
+      if(widget.get('parent') == thisParent && Math.abs(widget.get('x')-thisX) < snapRadius && Math.abs(widget.get('y')-thisY) < snapRadius) {
         if(widget.isBeingRemoved || widget.get('owner') !== thisOwner || JSON.stringify(widget.get('onPileCreation')) !== thisOnPileCreation)
           continue;
 
--- a/client/js/widgets/card.js
+++ b/client/js/widgets/card.js
@@ -12,6 +12,7 @@ class Card extends Widget {
 
       deck: null,
       cardType: null,
+      pileSnapRange: 10,
       onPileCreation: {}
     });
 
--- a/client/js/widgets/pile.js
+++ b/client/js/widgets/pile.js
@@ -14,7 +14,6 @@ class Pile extends Widget {
       inheritChildZ: true,
 
       text: null,
-
       pileSnapRange: 10,
 
       handleCSS: '',
@@ -41,7 +40,6 @@ class Pile extends Widget {
   }
 
    applyDeltaToDOM(delta) {
-
     super.applyDeltaToDOM(delta);
     if(this.handle && delta.handleCSS !== undefined)
       this.handle.style = mapAssetURLs(this.cssAsText(this.get('handleCSS'),null,true));
--- a/client/js/widgets/pile.js
+++ b/client/js/widgets/pile.js
@@ -15,7 +15,7 @@ class Pile extends Widget {
 
       text: null,
 
-      snapRadius: 10,
+      pileSnapRange: 10,
 
       handleCSS: '',
       handleSize: 'auto',
@@ -41,8 +41,6 @@ class Pile extends Widget {
   }
 
    applyDeltaToDOM(delta) {
-
-    this.handle.snapRadius = this.get('snapRadius');
 
     super.applyDeltaToDOM(delta);
     if(this.handle && delta.handleCSS !== undefined)
--- a/client/js/widgets/pile.js
+++ b/client/js/widgets/pile.js
@@ -15,6 +15,8 @@ class Pile extends Widget {
 
       text: null,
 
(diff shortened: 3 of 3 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1562/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1562-pr-test`.

ArnoldSmith86:
Snapping g is always between two widgets. Which widget should dictate the distance? Should it be the maximum of both values?

96LawDawg:
I'm not changing the process for snapping.  Just trying to expose the set value of 10px as a variable that players can change.

96LawDawg:
Oh.  But I think I understand now that I hit send.  You are wondering what would happen if you set the value so big it overlapped with another pile?  But that would still technically be a problem now, just hard to create since the distance is only 10px.

ArnoldSmith86:
No, I mean snapping applies to two widgets. The one you are dropping and the one already there. The 10px are fixed now (so the same value for both) but if you make it configurable, the check needs to know if it should use the value from the dropped widget, from the widget on the floor or something in between / derived.

RaphaelAlvez:
> No, I mean snapping applies to two widgets. The one you are dropping and the one already there. The 10px are fixed now (so the same value for both) but if you make it configurable, the check needs to know if it should use the value from the dropped widget, from the widget on the floor or something in between / derived.

I believe we agreed that properties would com from what it's being dropped into. So piles, holder or cards on the table

96LawDawg:
I guess I'm just being stupid.  I still don't understand why the current behavior needs to change.  Couldn't we just change that 10px value wherever it appears (I thought it was just in one place) and replace it with the variable `snapRadius` and everything would work the way it currently does except you could make piles form from farther away?

ArnoldSmith86:
It doesn't need to change. Its currently `this.get('x') - widget.get('x') < 10`. Question is, should it be `this.get('x') - widget.get('x') < this.get('snapRadius')` or `this.get('x') - widget.get('x') < widget.get('snapRadius')` or `this.get('x') - widget.get('x') < Math.max(this.get('snapRadius'), widget.get('snapRadius')`?

flackr:
By putting the value in onPileCreation it would require that they had the same value in order to pile.

96LawDawg:
> It doesn't need to change. It's currently `this.get('x') - widget.get('x') < 10`. Question is, should it be `this.get('x') - widget.get('x') < this.get('snapRadius')` or `this.get('x') - widget.get('x') < widget.get('snapRadius')` or `this.get('x') - widget.get('x') < Math.max(this.get('snapRadius'), widget.get('snapRadius'))`?

I'd be lying if I said I understood the consequences of the decision.  So I don't care.  I don't think.  But I can't even test the different scenarios because right now I have only figured out a way to hard code the variable.

96LawDawg:
> By putting the value in onPileCreation it would require that they had the same value in order to pile.

Good point.  And doing that may have unintended consequences.  So do you think we should put it as a standalone property on cards?

ArnoldSmith86:
> > By putting the value in onPileCreation it would require that they had the same value in order to pile.
> 
> Good point.  And doing that may have unintended consequences.  So do you think we should put it as a standalone property on cards?

I don't think anything should _only_ work in `onPileCreation`. It's just a collection of properties applied to the pile.

flackr:
Well it would probably be a property of a pile but we should also check for it in onPileCreation when you're just dropping one card on another as well. I think it's a good thing that it forces the snap distance to be the same.

96LawDawg:
This works now, making a property of cards and a property of piles.  Using onPileCreation, you can make it 2 separate things: one for distance for single cards and another one for piles.  I didn't mess with the formula.  It seems to work fine.  I don't know why we would want to change it.

client/js/widgets/widget.js
```diff
@@ -2222,8 +2222,8 @@ export class Widget extends StateManaged {
       if(widgetType != 'card' && widgetType != 'pile')
         continue;
 
-      // check if this widget is closer than 10px from another widget in the same parent
-      if(widget.get('parent') == thisParent && Math.abs(widget.get('x')-thisX) < 10 && Math.abs(widget.get('y')-thisY) < 10) {
+      // check if this widget is closer than the pileSnapRange from another widget in the same parent
+      if(widget.get('parent') == thisParent && Math.abs(widget.get('x')-thisX) < this.get('pileSnapRange') && Math.abs(widget.get('y')-thisY) < this.get('pileSnapRange')) {
```
flackr, line 2226: So I find it a bit awkward that if you forget to specify this on `onPileCreation` that setting it in cardDefaults only changes it for dragging cards but not piles.

https://test.virtualtabletop.io/PR-1562/card-snap-range

Whereas if we forced the setting to be in `onPileCreation` or on piles, then you would guarantee that setting it sets it for both the cards and piles of cards.

The other strange thing is what @ArnoldSmith86 mentioned that this will use the snap range of the card your dragging and not care about the other's snap range so you'll get different behavior depending on which card you're dragging, e.g. in the following table if you drag the queen it has a large snap distance for snapping onto the king, and small if you drag the king to snap onto the queen:

https://test.virtualtabletop.io/PR-1562/different-snap

96LawDawg: Everything about piles is awkward.

>Whereas if we forced the setting to be in onPileCreation or on piles, then you would guarantee that setting it sets it for both the cards and piles of cards.

I don't think that is true.  At least not in my testing.  Single cards need a value in cardDefaults or directly on each individual card.  Piles need their own setting.  I don't know of a way to have the variable on cards and the variable on piles talk to each other and also be capable of being separate things.  I'm open to ideas?

> The other strange thing is what @ArnoldSmith86 mentioned that this will use the snap range of the card your dragging and not care about the other's snap range so you'll get different behavior depending on which card you're dragging, e.g. in the following table if you drag the queen it has a large snap distance for snapping onto the king, and small if you drag the king to snap onto the queen:

What do you propose as a solution?  Personally, I think it is unlikely in most cases that different cards would have different values.  This makes the most sense if you make piles and cards in the same deck all have the same value.  Of course, there are always other use cases.  Do we automatically make it so the engine uses the card being dragged (current behavior)?  Or the card/pile on the floor?  Or something else?  Or do we create some other property that lets the user set whether it is the dragging card or the one on the floor?

flackr: Perhaps easier to show than to explain:
https://github.com/ArnoldSmith86/virtualtabletop/pull/1568

Then you can set this on the deck:
```
  "cardDefaults": {
    "onPileCreation": {
      "pileSnapRange": 100
    }
  },
```

And the same value applies to individual cards and piles created by those cards.

96LawDawg: Yes.  That works perfectly.  That is what I wanted to do but couldn't make it happen.

96LawDawg:
#1568 is better. Closing this one.

PR #1563, opened 2023-01-24
split type specific commands from widget commands in JSON editor

ArnoldSmith86:
This was just a quick idea I had. Do you like it?

### DIFF
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -2296,7 +2296,7 @@ function jeShowCommands() {
     delete command.currentKey;
     const contextMatch = context.match(new RegExp(command.context));
     if(contextMatch && contextMatch[0]!= "" && (!command.context || jeStateNow && !jeJSONerror) && (!command.show || command.show())) {
-      const title = command.isTypeSpecific ? contextMatch[0] : 'widget';
+      const title = command.isTypeSpecific || command.isTypeSpecific === undefined ? contextMatch[0] : 'widget';
       if(activeCommands[title] === undefined)
         activeCommands[title] = [];
       activeCommands[title].push(command);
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1563/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1563-pr-test`.

rogerl50:
I do. I was thinking as I was adding property after property to the scoreboard that it was hard to find them in and amongst the general properties.

RaphaelAlvez:
So just from taking a look at the code I think you are looking at the default of each widget but I see that some properties are missing or maybe should be considered to be included

holder :
- enterRoutine
- leaveRoutine
- inheritChildZ
- also I believe dropOffset is only usable in holders but it is displayed to all widgets

cards:
- overlap
- Overlap is also displayed to all widgets but maybe should be limited to cards and basic widgets

basic:
- classes (many game pieces use those)

label:
- remove clickable?

button:
- clickRoutine

deck should also show cards properties in cardDefaults

I'm happy with the other widgets JSON

also we may want to consider adding the following extra divisions:
- frequently used: (parent, fixedParent, clickable, clickRoutine, scale, rotation, enlarge, css, classes)
- routines (clickRoutine, enterRoutine, leaveRoutine, changeRoutine, globalUpdateRoutine)
- position ( x, y, z, layer, movable, movableInEdit, grid, dragLimit)
- seat linking ( linkedToSeat, onlyVisibleForSeat, hoverInheritVisibleForSeat)
- temporary ( draging, hoverTarger, owner)

RaphaelAlvez (approved):
go for it

PR #1566, opened 2023-01-27
Add room name to browser tab

### LINKED ISSUE #517: Show game + variant in <title>
Title currently only states "VirtualTabletop.io".
I think a fitting syntax could be "name of game (name of variant) - VirtualTabletop.io". Not really a functional requirement, just nice to have with multiple open VTT tabs.

96LawDawg:
Closes #517.

If no named game is loaded, the title in the browser tab is "VirtualTabletop.io - abcd".

(Originally when I made this, iIf a game is loaded, the tab reads "VTT - Name of Game".  But it doesn't do this now)

No questions.  Just approve.  You know you want to.  2 lines of code that don't hurt anything and make everything better.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1566/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1566-pr-test`.

mousewax:
Can the server name or title be added to the config.json and used from there? My server needs a different title than the main VTT one.

ArnoldSmith86:
This is in the wrong place. It'll only work for the player loading the game and only until they reload their tab.

It needs to be in the event handler that receives new room metadata and then it needs to look at activeState and figure out what game is active.

Should it be `<game> - VTT` instead? I'm not sure. What do people think?

And yeah, it's probably a good opportunity to make the name configurable. Apparently we need a long and a short name.

rogerl50:
I would definitely put the name of the game first, for those of us who have too many tabs open. Also yes, it should be configurable by server so I can tell my localhost tabs from production tabs.

rogerl50:
@96LawDawg And you thought this would be easy...

96LawDawg:
AHHHHHH.  What is the airspeed velocity of an unladen swallow?

96LawDawg:
> It needs to be in the event handler that receives new room metadata and then it needs to look at activeState and figure out what game is active.

I don't know where that event handler is..

> And yeah, it's probably a good opportunity to make the name configurable. Apparently we need a long and a short name.

And I don't know how to do that.  As you know, my knowledge of webpage design wouldn't completely fill the period at the end of this sentence.

ArnoldSmith86:
> I don't know where that event handler is..

```Javascript
  onMessage('meta', function() {
    if(!checkedOnce)
      checkURLproperties(true);
    checkedOnce = true;
  });
```

in `main.js`. Should be something like:

```Javascript
  onMessage('meta', function(args) {
    if(!checkedOnce)
      checkURLproperties(true);
    checkedOnce = true;
    for(const state of Object.values(args.meta.states))
      if(args.meta.activeState && args.meta.activeState.stateID == state.id)
        // set title with state.name
  });
```

> And I don't know how to do that.

Add a new setting in `config.json`; access it using `config[settingName]`.

rogerl50:
How does this work in practice? What I'd like is to make things so that on my local server I always get a name like 'local'. Does that mean I always have to modify the downloaded `config.json`? If so, is there a way to have that include another file containing that property that I can set to a fixed value on my server?

ArnoldSmith86:
Sorry, I meant to say `config.template.json`. Because `config.json` is exactly that: a file that is not tracked by git so that you can have your own local settings.

rogerl50:
> I think the setting in `config.json` (created from `config.template.json` if missing) should provide a format string that is used with the game name, so that some servers can have "Game Title - Server Name" and others can have "Server Name - Game Title".
> 
> I think the template should provide "VTT - Game Title", but config.json on virtualtabletop.io server should use "Game Title - VirtualTabletop.io" so that by default a local server and production have different formats.
> 
> Should we also provide a way to use roomID in the title?
> 
> Should we use a separate configuration string for when no game is loaded?

We could, I assume, process a string like "%s (%r) - %g" and substitute the server name, room id, and game name (or 'None' if no game is loaded, right? Then the person writing the config.json line could get whatever is desired.

But still, games hosted on the main VTT server (for example) would all have the same string, right? Or is there a way to make that game- or room-specific?

ArnoldSmith86:
And then have one of those for _with_ and one for _without game loaded_?

mousewax:
Or just have a default game name for when none is loaded. Like "New Room" or "Start a Game" so that there's only one string?

96LawDawg:
I am unable to get the activeState stuff from states.js to work in main.js.  I've tried exporting, importing, whatever.  So I am not able to retrieve the game name.  So here is what I have done.  The tab now shows the room name first followed by either VirtualTabletop.io or whatever custom label you want, set in the config.json file.

CustomTab

robartsd:
> But still, games hosted on the main VTT server (for example) would all have the same string, right? Or is there a way to make that game- or room-specific?

Yes, all games on the server would have the same thing. I think it would be neat to provide room options for customizing. I don't think games should be able to customize it.

robartsd:
> And then have one of those for _with_ and one for _without game loaded_?

That is what I was thinking.

robartsd:
> I am unable to get the activeState stuff from states.js to work in main.js.

The 'meta' message should include the new `stateID`. The 'meta' message can be accessed in the function if the onMessage line (line 458) is changed to `onMessage('meta', function(message) {`. I'm not sure how to look up the game title from stateID.

96LawDawg:
> The 'meta' message should include the new `stateID`. The 'meta' message can be accessed in the function if the onMessage line (line 458) is changed to `onMessage('meta', function(message) {`. I'm not sure how to look up the game title from stateID.

That didn't work for me either.

RaphaelAlvez (approved):
It works and is an improvement

PR #1568, opened 2023-01-28
Add pileSnapRange property to piles.

### LINKED ISSUE #201: Increase snap radius for stacking cards

flackr:
Closes #201.

Alternative approach to #1562 where there is only a single property for piles but it is also read off of the onPileCreation properties for cards as this is the snap distance a pile created by such cards would get.

------------------
Add this to the Pile widget type in the wiki:

| pileSnapRange | 10            | The distance at which cards will form new piles or join existing piles. This is measured in pixels from the upper-left corner of the widget being dragged to the upper-left corner of the card or pile which it will join. If either x or y is less than this distance, the card being dragged will join the pile when dropped. Set this value in `onPileCreation` (see [Card](#card) for more information).


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1568/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1568-pr-test`.

96LawDawg (approved):
Approved through test and code review.  I added a new tutorial variant to Piles as well.

client/js/widgets/widget.js
```diff
@@ -6,7 +6,6 @@ import { showOverlay, shuffleWidgets, sortWidgets } from '../main.js';
 import { tracingEnabled } from '../tracing.js';
 import { toHex } from '../color.js';
 import { center, distance, overlap, getOffset, getElementTransform, getScreenTransform, getPointOnPlane, dehomogenize, getElementTransformRelativeTo } from '../geometry.js';
-import { defaultPileSnapRange } from './pile.js';
```
flackr: I think the fact that I needed to remove this might be a bundler error, that even though modules support bidirectional dependencies it uses the imports to establish a relative order in the bundle. If we wanted to keep the imports pure I could resolve this by putting the constant in a different file or perhaps some other method of determining the pile's default `pileSnapRange`.

The current patch works, but happy to go with some other approach if desired.

96LawDawg:
This just needs another code review to be ready.  Pretty simple.

RaphaelAlvez:
@96LawDawg you will need to redo your tutorial since the pile tutorial was updated.

@ArnoldSmith86 Do you mind checking that self-review flackr did?

client/js/widgets/widget.js
```diff
@@ -2315,8 +2315,9 @@ export class Widget extends StateManaged {
       if(widgetType != 'card' && widgetType != 'pile')
         continue;
 
-      // check if this widget is closer than 10px from another widget in the same parent
-      if(widget.get('parent') == thisParent && Math.abs(widget.get('x')-thisX) < 10 && Math.abs(widget.get('y')-thisY) < 10) {
+      // check if this widget is closer than the pileSnapRange from another widget in the same parent
+      let pileSnapRange = (thisType == 'card' ? this.get('onPileCreation').pileSnapRange : this.get('pileSnapRange')) || defaultPileSnapRange;
```
ArnoldSmith86, line 2319: `this.get('onPileCreation').pileSnapRange` is not _quite_ save. The property `onPileCreation` could be `null`. For example if inherited from another widget type.

ArnoldSmith86 (changes requested):

ArnoldSmith86 (approved):
Changed the code myself. Good to go IMHO.

96LawDawg:
@flackr, I tested ArnoldSmith's changes.  Works and now catches edge cases not originally considered.  This should be merged.

PR #1570, opened 2023-01-29
Replace ${THIS} in selectors with a selector for the widget element.

### LINKED ISSUE #1545: Properties are not replaced in css selectors.
Based on a discussion on discord, there was an attempt to style a widget based on a parent class. In theory the following should work:

```
  "css": {
    "_,body.edit #w_${PROPERTY id}": {
      "background": "red"
    }
  }
```

However, it doesn't, because the `${PROPERTY id}` isn't substituted with its value. This is simply a missing call to `cssReplaceProperties` here:
https://github.com/ArnoldSmith86/virtualtabletop/blob/main/client/js/widgets/widget.js#L483

flackr:
This fixes #1545, supporting ancestor selectors that style the current element.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1570/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1570-pr-test`.

robartsd:
Eventually we will probably introduce other `${}` expansions. Should we eliminate the default prefixing of the selector with the widget ID if a `${}` expansion is found in the selector text (eliminating the need for the comma hack when using a `${}` expansion?

flackr:
> Eventually we will probably introduce other `${}` expansions. Should we eliminate the default prefixing of the selector with the widget ID if a `${}` expansion is found in the selector text (eliminating the need for the comma hack when using a `${}` expansion)?

Yeah that makes sense, since any use of expansions would be effecitvely anchoring to that expansion. Done.

robartsd (approved):
The code looks good, but I have not done any testing yet.

96LawDawg:
I would like to test this and get it approved, but I don't even know where to start.  Give me some suggestions for things to try?

flackr:
You can create some css which targets the element itself based on classes in an ancestor or based on global modes.

E.g. the original issue that came up on discord (styling a widget differently in edit mode) could be implemented with this:
```
  "css": {
    "_,body.edit ${THIS}": {
      "background": "red"
    }
  }
```

Perhaps more practical things you could do is styles conditional on a class from the containing holder, e.g.
```
  "css": {
    "_,.holder-class ${THIS}": {
      "border": "1px solid red"
    },
  }
```

Which would only apply when it's in a holder with `"classes": "holder-class"`

RaphaelAlvez:
I tried applying it but I couldn't figure out how to use it. I tried to apply it to my original problem that got added to the cursor tutorial

RaphaelAlvez (approved):
tested with Flackr I was using the selector incorrectly

PR #1571, opened 2023-01-29
Use document.elementsFromPoint for JSON editor targets.

### LINKED ISSUE #1567: JSON Editor
When hovering over widgets using the JSON Editor, it does not take into account rotation. So when widgets are rotated, you have to move the mouse around a bit to get the JSON Editor to recognize which widget you are over.  @flackr suggests this can be fixed by using the elements from the point API.

flackr:
Using elementsFromPoint for the JSON editor takes into account transforms, clip paths and z ordering implicitly. Fixes #1567.

Note that this currently does mean hidden (foreign) objects are no longer listed. If we want those we can do something similar to the original however it is more complicated.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1571/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1571-pr-test`.

96LawDawg:
What do you mean by hidden/foreign objects?

flackr:
> What do you mean by hidden/foreign objects?

Things that aren't visible to you because they're not owned by you. The current code looks like it puts those in the list. I'm not sure how often people use this.

96LawDawg:
It doesn't come up that often, but when I'm debugging a game with seats or hands, I need to be able to access widgets that are owned by other players.  I can do that in the tree part, but I prefer to use the graphical part and the function keys.  Leaving that capability for me is a higher priority than fixing 1567.  If both can be done, great.  But if it is highly complicated, then it isn't something that has to be done.

robartsd:
I think this could be an easy fix: hide widgets with `{opacity:0; pointer-events: none}` instead of `{display:none}` when in JSON edit mode.

flackr:
> If both can be done, great. But if it is highly complicated, then it isn't something that has to be done.

Done. I've used the previous logic for hidden widgets. I could probably make it smarter and take into account the transform but at the least it works as before for hidden widgets and uses elementsFromPoint for visible ones (taking into account transforms, clip-paths and all that fun stuff).

flackr:
> I think this could be an easy fix: hide widgets with `{opacity:0; pointer-events: none}` instead of `{display:none}` when in JSON edit mode.

Yeah, that's an option too. I'd probably just use `visibility: hidden` which still permits hit testing from elementsFromPoint IIRC.

96LawDawg (approved):
Approved through test.  Thanks @flackr. It's great to have someone with your skills helping out on this project.  Really appreciate it.

flackr:
Updated to use `elementsFromPoint` regardless, but temporarily make foreign objects hittable just for the hit test per @robartsd suggestion. This makes the code much simpler (removing the coordinate cache), and supports arbitrary transforms / clips for everything.

96LawDawg (approved):
New method approved through test

robartsd:
Visibility rules for decks should also be changed.

flackr:
> Visibility rules for decks should also be changed.

I think the latest commit is what you've asked for, can you confirm? https://github.com/ArnoldSmith86/virtualtabletop/pull/1571/commits/4aa09310e4ab16d539f4cf34c3c659f492b6e83f

robartsd:
> > Visibility rules for decks should also be changed.
> 
> I think the latest commit is what you've asked for, can you confirm? [4aa0931](https://github.com/ArnoldSmith86/virtualtabletop/commit/4aa09310e4ab16d539f4cf34c3c659f492b6e83f)

That looks like it should work. Could use the same combination of `visibility: hidden` and `pointer-events: none` used for foreign  widgets, but there are usually far fewer deck widgets than potential foreign widgets, so the impact of switching display back and forth on decks is less (and even if a frame is presented to the user, seeing the deck would not accidentally reveal anything).

robartsd:
I guess the pointer-events: none is not needed when visibility: hidden is set. Visibility is not purely a visual property (hides from the accessibility tree and pointer-events as well).

96LawDawg (changes requested):
I don't know or understand what is going on with the changes to decks, but I'm not sure we have the right answer.  It used to be that you could see a circle in the holder when using JSON mode and clicking the edit icon.  That is no longer the case.  You can still move the mouse over and "see" it using function keys.  If you use the regular "Edit Mode" from the main toolbar, you can see the circle.  I don't understand why it isn't visible in the JSON edit view anymore.  I think it should be.

flackr:
@96LawDawg fixed - it's now visible while selected like before.

96LawDawg:
It's still not visible, even when selected.  But I was saying it should be visible all the time. At least, that is what I think.  Would appreciate input from other game developers.

96LawDawg:
Correction.  It IS visible when not using the JSON Editor "edit" mode view.  But it should be visible there all the time.  My opinion.

flackr:
@96LawDawg as far as I can tell it is identical to the current style / visibility without this PR. Or is there something I'm missing? I tried everything with this PR and without and I think everything is the same visually.

As usual, I don't think we should make changes beyond this bug fix to correctly target objects in the JSON editor (and maintain the current visuals) in this PR. We can always open another one to make decks visible in the json editing mode.

96LawDawg:
Using Alpaca as a reference, look at production and this PR.

Production

PR

flackr:
@96LawDawg Was fixed by @robartsd latest patch. Thanks @robartsd!

96LawDawg (approved):
Changes fixed the deck visibility issue.  Approved again through test.

robartsd (approved):
Code changes look fine.

96LawDawg:
@flackr, this is ready to be merged.  Has 2 approvals including test and code review.

PR #1578, opened 2023-02-03
add support for GitHub Codespaces

robartsd:
`testcafe-gitpod` script works without modification; however, functions.js test fails on 2-core VM (succeeds on 4-core VM). 

Overall Gitpod performance feels better. Codespaces does have the advantage of saving the whole container when stopped, not just the workspaces folder. gitpod-vnc scripts do not work for watching testcafe tests (but the basic concept does work the same).

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,14 @@
+{
+    "updateContentCommand": "npm install",
+    "postStartCommand": "npm run debug",
+    "portsAttributes": {
+        "8272": {
+            "label": "virtualtabletop",
+            "onAutoForward": "notify"
+        },
+        "9229":{
+            "label": "VTT-websocket",
+            "onAutoForward": "silent"
+        }
+    }
+}


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1578/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1578-pr-test`.

ArnoldSmith86:
Port 9229 is the Node.JS debugger. VTT websocket uses the same port as the HTTP server.

robartsd:
`source tests/testcafe/gitpod-vnc/install.sh` now also works in Codespaces. In a Codespace it only has to be run once, in Gitpod it has to be run once every time the Workspace is restarted. Script also now better handles running again after it has already been set up.

RaphaelAlvez:
Is this working? I want to do some work in code spaces and it sucks not being able to test before comiting

robartsd:
No, when I try to open one of these customized devcontainer in codespaces, the devcontainer build process results in an error and the codespace starts the recovery devcontainer. Running the scripts in a running codespace may work.

PR #1580, opened 2023-02-04
Use Fisher-Yates shuffle instead

tsankuanglee:
Please see the Wikipedia entry [Fisher-Yates Shuffle](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle) for its benefits.

This implementation also consumes less memory (integer rather than float; no need to duplicate widgets).

### DIFF
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -204,7 +204,7 @@ export async function shuffleWidgets(collection) {
   // Fisher–Yates shuffle
   const len = collection.length;
   let indexes = [...Array(len).keys()];
-  for (let i = len - 1; i > 0; i--) {
+  for (let i = len-1; i > 0; i--) {
     let j = Math.floor(Math.random() * (i+1));
     [indexes[i], indexes[j]] = [indexes[j], indexes[i]];
   }
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -201,11 +201,15 @@ function setScale() {
 }
 
 export async function shuffleWidgets(collection) {
-  const shuffle = collection.map(widget => {
-    return {widget, rand:Math.random()};
-  }).sort((a, b)=> a.rand - b.rand);
-  for(let i of shuffle) {
-    await i.widget.bringToFront();
+  // Fisher–Yates shuffle
+  const len = collection.length;
+  let indexes = [...Array(len).keys()];
+  for (let i = len - 1; i > 0; i--) {
+    let j = Math.floor(Math.random() * (i+1));
+    [indexes[i], indexes[j]] = [indexes[j], indexes[i]];
+  }
+  for (let i of indexes) {
+    await collection[i].bringToFront();
   }
 }
 
--- a/tests/testcafe/editor.js
+++ b/tests/testcafe/editor.js
@@ -29,13 +29,13 @@ test('Create game using edit mode', async t => {
     .click('#w_9ee9B')
     .click('#w_9ee9P > .handle')
     .click('#pileOverlay > button:nth-of-type(3)')
-    .click('#w_5ip4 > .handle')
+    .click('#w_b86p > .handle')
     .click('#pileOverlay > button:nth-of-type(1)')
-    .click('#w_5ip4 > .handle')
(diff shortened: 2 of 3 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1580/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1580-pr-test`.

ArnoldSmith86:
Did you do actual performance tests? I can't imagine that this is significantly faster unless you shuffle a million widgets..

flackr:
> integer rather than float

FYI, Javascript numbers are always 64-bit floats, so even if you use integer values it takes the same memory. Though the memory used to store a number for each widget is insignificant compared to the size of everything else held in memory for a widget.

robartsd:
Memory savings is probably in the tens of bytes per widget. The current algorithm creates a new array with each element consisting of a new object (12+ bytes) containing a reference to an existing widget object (4+ bytes) and a number (4 bytes). This algorithm creates a new array with each element containing a number (4 bytes). Tens of bytes per widget is not all that significant overall, but I don't see how it could hurt. I imagine that there is also processing savings by not needing to create and access a new object for each widget processed. This is before considering any savings from the sort algorithm itself.

I don't see any reason not to implement this change though I am curious why the testcafe create game is choosing a different id some of the widgets created. It seems like this should be calling Math.random() the same number of times in a shuffle, so only shuffle results should be changed.

flackr:
This is also O(n) whereas sorting by the random values is O(n*log(n)). Not likely to be significant on the size of collections being sorted but still no reason not to take the improvement.

robartsd (approved):
This is should be more memory and processor efficient with no drawbacks. Testcafe updates are merely because implementation details about how randomness is used to determine shuffle order has changed. Changes in editor.js ids are due to shuffle using 1 less random number, so subsequent random values are changed (made most obvious by the second split pile id becoming what the first split pile id had been).

ArnoldSmith86:
Did the old method change the order in the actual collection? I think it didn't and the new one still doesn't, right? We should not change that.

Other than that, I guess you're right: why not?

96LawDawg:
So not a "real" test, but I loaded up Cards Against Humanity into production and this PR.  Jacked the number of cards up from 500 to 2500.  Looked at the time displayed in the JSON Editor.  Both rooms ranged between 7-9ms depending on who knows what.

96LawDawg (approved):
Approved through test.  Cards still shuffle.

robartsd:
> Did the old method change the order in the actual collection? I think it didn't and the new one still doesn't, right? We should not change that.

Neither changes the collection order. Both version create a new array to shuffle, call the widgets in the shuffled order, then let the new array go out of scope. This simply changes the nature of the new array and the method for shuffling it.

robartsd:
In my testing, I am getting slower shuffles (~30ms vs ~10ms), but I think the reason is external to this as other routine steps seem similarly affected. In any case, my biggest issue with large decks is RECALL (about 40-50x slower than SHUFFLE).

96LawDawg:
> In my testing, I am getting slower shuffles (~30ms vs ~10ms), but I think the reason is external to this as other routine steps seem similarly affected. In any case, my biggest issue with large decks is RECALL (about 40-50x slower than SHUFFLE).

I didn't test RECALL in this PR.  Are you saying RECALL in this PR is slower than on production or just that RECALL is slow in general (a known issue).

robartsd:
> > In my testing, I am getting slower shuffles (~30ms vs ~10ms), but I think the reason is external to this as other routine steps seem similarly affected. In any case, my biggest issue with large decks is RECALL (about 40-50x slower than SHUFFLE).
> 
> I didn't test RECALL in this PR. Are you saying RECALL in this PR is slower than on production or just that RECALL is slow in general (a known issue).

No RECALL isn't changed, just it takes a long time on large decks (very noticable on my low end computers).

Issue #1209, opened 2022-05-05
Asset aliases


robartsd:
It would be nice to use names for assets instead of cryptic numeric codes. I'm thinking that `_meta.assets` could contain a map keyed with names for assets with values of the asset URL. The map would be saved with the game state. To use an asset `${ASSET name}` would be placed in the widget JSON. This would mean that copying the JSON of a widget from one game to another could lead to asset aliases that do not have a mapping. Probably need a new button in the JSON editor to show the asset aliases (perhaps the whole `_meta` object).

PR #1203, opened 2022-04-28
added commands for exporting and importing assets

ArnoldSmith86:
The main JSON editor toolbar now has a button that downloads all room assets as a zip file with the files named after their widget ID and property.

You can then extract and edit those images on your machine and use the second new button to upload one or more of the edited files. The new assets will then be uploaded and based on the filename, the new asset URL will be updated in the room state.

If you use the same asset in multiple places, this would decouple them unless you upload the same new asset for each of the properties it is used in. Because of that the download can optionally create files named after their asset URL. After editing and re-uploading them, the filename is used to find the old asset URL and it gets replaced by the newly created one.

- [ ] This needs to escape special characters like `/` in filenames.

PR #1581, opened 2023-02-04
Allow accessing library games' _raw_ assets

tsankuanglee:
I'm throwing this one out for discussion. This PR allows _direct_ access to _raw_ asset files under `library/games/xxxx/assets/yyyy.zzz` via route `/assets/:game/:name`.

Benefits:
* Assets files can be in their original raw form, e.g. webp, png, etc., as opposed to their hash name. This makes game creation workflow easier (no need to convert files to hash names; file extensions are kept so application can open them more easily).
* faster file serving -- no need to open files to inspect its first bytes
* future-proof -- no need to add byte identifiers for newer file types

We can tune cache control if necessary.

I'm not sure why game assets were converted to hash names in the first place, so there might be reasons that I am unaware. Can we discuss this?

### DIFF
--- a/server.mjs
+++ b/server.mjs
@@ -112,6 +112,23 @@ MinifyRoom().then(function(result) {
     res.send(result);
   });
 
+  router.get('/assets/:game/:name', function(req, res) {
+    const highest_allowed_directory = path.join(Config.directory('library'), "games") + "/";
+    const filepath = path.join(highest_allowed_directory, req.params.game, "assets", req.params.name);
+    // security guard against escaping the highest_allowed_directory
+    if (filepath.indexOf(highest_allowed_directory) !== 0) {
+      res.sendStatus(404);
+    }
+
+    fs.access(filepath, fs.F_OK, (err) => {
+      if (err) {
+        res.sendStatus(404);
+      } else {
+        res.sendFile(filepath);
+      }
+    });
+  });
+
   router.get('/assets/:name', function(req, res) {
     if(!req.params.name.match(/^[0-9_-]+$/) || !Config.resolveAsset(req.params.name)) {
       res.sendStatus(404);


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1581/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1581-pr-test`.

robartsd:
Hash names for uploaded assets prevent name collisions allows all uploaded files to be stored in the same folder without worrying about games having different assets with the same name. I agree that the new public library structure does not depend on these hash names any more.

ArnoldSmith86:
Two things:
- Uploading the same game to multiple rooms should not store or even upload the assets again.
- The public library folder is just a bunch of extracted VTT files. I don't think it's a good idea to deviate from that.

robartsd:
I do like having extensions control file type being served instead of the server needing to identifying type by file contents each time the file is served (this was previously identified as a problem preventing us from hosting game style sheet files). Perhaps the server should check uploaded files using the algorithm that is currently used when serving, then use the extension when saving the file. This would allow us to add support for new types even if they take more processing to identify the type because the check would not need to occur every time the file is served. The server could be configured to look for resources by name without extension so compatibility with existing games should not be a problem.

As far as convenient names in game code, I prefer a solution that generalizes over all games, not just public library: #1209

ArnoldSmith86:
Wouldn't it be even slower to test all possible extensions if an asset `/assets/1234_567` is requested without file extension and the file is stored with an extension on the system?

Maybe rather just cache the file type in memory or in a JSON file?

robartsd:
> Wouldn't it be even slower to test all possible extensions if an asset `/assets/1234_567` is requested without file extension and the file is stored with an extension on the system?
> 
> Maybe rather just cache the file type in memory or in a JSON file?

Many web servers can be configured to handle this type of matching (or even select from multiple variations distinguished by extension based on the headers in the request). Could ask the file system for a list of files using a wildcard for the extension then serve the first file in the list (serving type based on the extension).

tsankuanglee:
Thanks for the explanation, @robartsd  and @ArnoldSmith86 .

> I agree that the new public library structure does not depend on these hash names any more.

Which games are good examples for this, @robartsd ?

> * Uploading the same game to multiple rooms should not store or even upload the assets again.

> As far as convenient names in game code, I prefer a solution that generalizes over all games, not just public library: #1209

My main target is only the library (and/or the games the server admins provide under their own library), and my main wishes are: 

1. keeping meaningful filenames; 
2. having namespcaes so the same filename `cover.webp` can be used as long as they are under different directories.

Personally, I use nginx as a front proxy to serve those files directly, and it has made asset/game creation and management so much easier, so I figured maybe we can make this convenience available to everyone.

For example, with this patch, whenever I need to change something in the image, I can find the file easily from their meaningful filenames, and I can just edit the file directly with any external software and save it back. This convenience became significant when editing images for a deck of cards.

As for 


### non-library games 

User uploaded games are not part of my solution, and 



> * The public library folder is just a bunch of extracted VTT files. I don't think it's a good idea to deviate from that.

tsankuanglee:
Thanks for the explanation, @robartsd  and @ArnoldSmith86 .

> * The public library folder is just a bunch of extracted VTT files. I don't think it's a good idea to deviate from that.

> I agree that the new public library structure does not depend on these hash names any more.

Which games are good examples for this, @robartsd ?

@ArnoldSmith86 could you help me understand why this would deviate from extracted VTT files?

I thought I simply provie an _additional_ way to access assets. Older games can still access them via hash. Newer game can still use hash-named files, or, use the new way here.


> As far as convenient names in game code, I prefer a solution that generalizes over all games, not just public library: #1209

> * Uploading the same game to multiple rooms should not store or even upload the assets again.

My main wishes are:

1. keeping meaningful filenames; 
2. having namespcaes so the same different games can freely use their own assets without worrying about clashing.

Currently, I use nginx as a front proxy to serve those assets directly, and it has made asset/game creation and management so much easier, so I figured maybe we can make this convenience available to everyone. For example, whenever I need to modify a card image, I can find the file easily from their meaningful filenames, and I can just edit the file directly with any external tools and save it back. This convenience became significant especially when editing images for a deck of cards.

It's indeed challenging to have a good solution to cover user-generated contents and still have such convenience, while saving disk space. Therefore, for now, I only focus on improving the experience for creators/site-owners who manage contents in their own library directory.

Thoughts?

ArnoldSmith86:
> @ArnoldSmith86 could you help me understand why this would deviate from extracted VTT files?

If you use named assets, you can no longer just download a VTT file of the public library game and add it as a normal game in any room. Which would probably be the way non-git-savvy people would propose/make changes to public library games.

I totally agree that we should make it easier to work with assets. But I would prefer to have a solution that works with all games. #1203 provides tools to replace assets more easily; @robartsd already mentioned his suggestion to have a mapping somewhere.

I would also like to add something similar to #1203 to the CSV import/export. So you would download a zip file of the deck CSV and all its assets named by cardtype.

Another thing that would probably be very useful, would be a view where you just see all game assets and each one has an Upload button to replace it by a new file.

All of those tools should be accessible more obvious in the new edit mode of course.

robartsd:
For regular games, all assets are in a single folder. Multiple games can (and do) use the same assets. Changing the asset for one game should not affect the other. The hash_size name convention is designed to support this.

tsankuanglee:
Thanks again. It looks like the team already has some other solutions in the cooking, so I'll avoid stepping on anyone's toes.

That said, I would still like to understand the considerations better so I won't stray too far.

> If you use named assets, you can no longer just download a VTT file of the public library game and add it as a normal game in any room. 

Maybe I'm misunderstanding something about your scenario -- it works for me though. Here's what I did to test it:

1. Create a new TestGame in library/ with the following files (attached [TestGame-in-library.zip](https://github.com/ArnoldSmith86/virtualtabletop/files/10614171/TestGame-in-library.zip) ).

```
library/TestGame/0.json
library/TestGame/assets
library/TestGame/assets/cover.webp
library/TestGame/assets/background.webp
```
`0.json` has reference to "/assets/TestGame/cover.webp" and "/assets/TestGame/background.webp" (this PR will serve them).

2. Play the game (I just moved the pawn out of the image)
3. Go to Game Shelf screen, select "SAVE ACTIVE GAME", and give it a name. It creates a new grid in the In-Progress Game section.
4. Click on that grid, and download the vtt file.

5. Now I go to another room (or another site with the same library and this PR).

6. In Game Shelf, select "ADD GAME" -> "UPLOAD FILES" to upload that vtt file. It creates a grid.
7. Click on that grid -> PLAY -> "LOAD IN-PROGRESS GAME"

I can see the saved game in another room, as expected.

What am I missing here?

> Which would probably be the way non-git-savvy people would propose/make changes to public library games.

I'm not following how git plays a role in assets management.

> For regular games, all assets are in a single folder. Multiple games can (and do) use the same assets. Changing the asset for one game should not affect the other. The hash_size name convention is designed to support this.

Valid point. That also means we can never delete old assets since some other games may depend on them, correct?

In the per-game asset scenario, I expect most games to only use the assets in the standard set and in their own assets/ directory, instead of relying on other games' assets. Of course, if there are enough games using the same assets, perhaps they should be part of the standard assets anyway.

robartsd:
> > For regular games, all assets are in a single folder. Multiple games can (and do) use the same assets. Changing the asset for one game should not affect the other. The hash_size name convention is designed to support this.
> 
> Valid point, @robartsd . That also means we can never delete old assets since some other games may depend on them, correct?

In a way, yes. Occasionally the server admin runs a script that scans all games on the server for assets used and deletes assets that are not used in any game on the server (including all games in every room, not just library games).

robartsd:
> Thanks again. It looks like the team already has some other solutions in the cooking, so I'll avoid stepping on anyone's toes.

That sounds a bit overly optimistic; we have come up with some ideas of things we think we would like to do. I don't think anyone is actively working on any of that yet.

ArnoldSmith86:
> What am I missing here?

Your VTT file doesn't contain all the game assets. It works in another room on _that server_ because it still serves the assets on the same path. But it doesn't work if you upload it to a server that didn't have the game in its library in the first place.

The same could be said about links to externally hosted assets but at that point it's a conscious decision not to use VTT asset management and we don't allow that in the official public library.

tsankuanglee:
> But it doesn't work if you upload it to a server that didn't have the game in its library in the first place.

If a server doesn't have that game in its library, isn't that server considered out-of-date, especially now that the library is part of the main repo? I imagine an out-of-date server will also suffer from other compatibility issues, such as missing the new standard assets, missing newly added functions, etc., and may not correctly support newer vtts anyway.

This PR only opens access to assets under library/games because they can be expected, as long as servers are in-sync with the repo.

Or is it a goal for vtt to play on very old servers as well?

robartsd:
> > But it doesn't work if you upload it to a server that didn't have the game in its library in the first place.
> 
> If a server doesn't have that game in its library, isn't that server considered out-of-date, especially now that the library is part of the main repo? I imagine an out-of-date server will also suffer from other compatibility issues, such as missing the new standard assets, missing newly added functions, etc., and may not correctly support newer vtts anyway.

There are several reasons a game asset from a library game might not be on a particular server: someone may choose to run a server without some or all of the library; a the game gets changed an no longer uses the asset; the game gets removed from the library for some reason.

tsankuanglee:
Thanks for the discussion. Since this is a different direction from your vision, I'll close this one.

PR #1583, opened 2023-02-06
fixed re-saving a savegame if the source game is no longer there

ArnoldSmith86:
If you transfer a vtts savegame to a different room, its reference to the game it is based on no longer works (because the new room does not have the same game with the same ID). This PR fixes multiple issues that were uncovered by this behavior:

- There are client crashes if certain game metadata does not exist or is not a string (mainly `name` and `mode`).
- If a state has an undefined image, the client would request `/undefined` which triggers loading room `undefined`. Incidentally that room was broken on my local machine which crashed the entire server.
- Saving the game _again_ would try to load the metadata for the new savegame from the game it was based on. That link is invalid so it produced a completely broken state. Now it falls back to the old savegame which also has all the metadata.

### DIFF
--- a/client/js/overlays/states.js
+++ b/client/js/overlays/states.js
@@ -450,7 +450,8 @@ function fillStatesList(states, starred, activeState, returnServer, activePlayer
         updateSaveButton.style.display = 'inline-flex';
     }
 
-    $('img', entry).src = mapAssetURLs(state.image);
+    if(state.image)
+      $('img', entry).src = mapAssetURLs(state.image);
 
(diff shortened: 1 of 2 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1583/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1583-pr-test`.

RaphaelAlvez (approved):
tested as described in discord

PR #1591, opened 2023-02-09
Only transition widgets which weren't just created.

### LINKED ISSUE #1590: Some transitions play when loading or widgets are added
E.g. if you load the room with the demo for adding transitions on transform https://virtualtabletop.io/sws5 you can see some of the cards slide into place.

This is also the issue that is leading to the drop shadow #1501 having a transition each time the shadow widget is added (if the shadow widget transitions).

The underlying issue is (I think) that we normally avoid transitions if the element was newly added, except that in the `receive` function in `serverstate.js` all widgets are first added to the top level. The code in the `applyDeltaToDOM` function in `widget.js` then thinks that it isn't a new widget and transitions from its location in the top layer to its position in the container it was added to.

flackr:
Note this is using a bit of a hack, the delta.id is in fact defined if the widget id changes as well. I'm not sure if we have a signal for the widget having just been created yet but the id trick works unless some programmatic code also changes the id while reparenting. This fixes #1590.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -157,7 +157,9 @@ export class Widget extends StateManaged {
     let newParent = undefined;
     if(delta.parent !== undefined) {
       newParent = delta.parent ? widgets.get(delta.parent).domElement : $('#topSurface');
-      fromTransform = getElementTransformRelativeTo(this.domElement, newParent);
+      // If the widget wasn't newly created, transition from its previous location.
+      if (delta.id === undefined)
+        fromTransform = getElementTransformRelativeTo(this.domElement, newParent);
     }
 
     this.applyCSS(delta);


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1591/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1591-pr-test`.

96LawDawg (approved):
Approved through test.  Testing wasn't very extensive because I don't know what else might pop up as we experiment with these sorts of changes.  But it fixes the bug.

robartsd:
> Note this is using a bit of a hack, the delta.id is in fact defined if the widget id changes as well.

Technically, when changing the id, the widget and it's DOM element are destroyed and recreated with the new id, so any transition triggered wouldn't have something to transition from.

flackr:
Thanks @robartsd, sounds like it's a reasonable condition for triggering a transition then.

ArnoldSmith86 (approved):
Code might make sense. I'm not sure how the transform magic works but this should prevent it for newly created widgets.

Issue #205, opened 2021-02-06
Create onLoadRoutine to allow immediate shuffle, etc. when loading game from saved library state


96LawDawg:
Juddly brought up a concern that when a game is loaded from its saved state in the library, the cards will always be in the same order they were in when the game was created.  This makes it is possible for every game to start out the same way.  Although his question was about Duel 52, the problem applies to every game that relies on shuffled cards and maybe other random order widgets.

Several solutions were discussed in the #help channel which also identified the need to support saved games.  So an onLoad command would need to be optional and not mandatory so that players can come back to saved games.

The best, I think, solution came from BicycleIndividual:  Perhaps we provide onLoadRoutine as a property that can be set for widgets, and an option for if onLoadRoutine should be run or not when saving state. In the game list the option could be labeled "Reset game on load?"

This certainly warrants further discussion and comment about how to best implement this.

bjalder26:
PR #540 should make this possible.

ArnoldSmith86:
> PR #540 should make this possible.

Only if you're the only player in the room while loading the game though.

This should really get its own proper feature.

robartsd:
> PR #540 should make this possible.

Yes, you could have a non-paused timmer saved in a room then have that timer trigger a routine after it starts. This would probably require manually editing the JSON to set up the timer state though. I don't think having other players in the room would be a problem. The player who started the timer would be the one running the reset routine. Only issue would be if other players get a delta in first that causes the timer player's delta to be rejected.

ArnoldSmith86:
> I don't think having other players in the room would be a problem.

```Javascript
if(delta.paused === false && activePlayers.length == 1)
  this.startTimer();
```

I guess it could be changed to "if I'm `activePlayers[0]`" though. Hmm. Maybe it should always be run by `activePlayers[0]` and when they leave, the new `activePlayers[0]` takes over.

PR #1592, opened 2023-02-09
experimenting with a reset function

ArnoldSmith86:
This is not really meant to be merged. This adds a new function `QUICKRESET` that recalls, shuffles, flips to 0 and rotates to 0 and applies `resetProperties` (as used in a few public library games).

My idea was to make it as fast as possible. Skipping all routines that might trigger and a lot of other stuff.

Unfortunately it also skips a few things that make following `MOVE` functions (and other stuff) not work because they do not know that the cards were recalled.

I am not sure if we want to introduce something similar to this or if this is just some kind of baseline for future `RECALL` optimization.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -1308,6 +1308,82 @@ export class Widget extends StateManaged {
         }
       }
 
+      if(a.func == 'QUICKRESET') {
+        // FIXME: need to press the button twice
+        setDefaults(a, { property: 'resetProperties', decks: [] });
+        const decks = {};
+        for(const deck of asArray(a.decks))
+          decks[deck] = true;
+
+        const offsetX = {};
+        const offsetY = {};
+        const cards = {};
+        const piles = {};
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1592/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1592-pr-test`.

ArnoldSmith86:
I only just noticed that one player only gets 9 cards.

96LawDawg:
As far as I can tell, the deck portion doesn't work at all.  I tried a stand alone deck in a simple room and got nothing using:

```
   {
      "func": "QUICKRESET",
      "decks": "wufmD"
    }
```

Also, if this ever does make it out of prototype, it would be great to see it incorporate the ability to do this when the room is loaded after pressing the play button on the UI.  See issue #205.

PR #1593, opened 2023-02-09
added `gameStartRoutine`

### LINKED ISSUE #205: Create onLoadRoutine to allow immediate shuffle, etc. when loading game from saved library state

ArnoldSmith86:
@96LawDawg just reminded me that we wanted to fix #205 after introducing saved in-progress games in #1223.

Here we go.

I added the feature to `The Team II` as a demo (it is in button `b0b2bf3c-50e3-4145-87cf-e9fb106d3981`).


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1593/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1593-pr-test`.

96LawDawg:
This seems to work great.  And will be easy to incorporate into existing games.  I tested it on Dominoes (where I had all the dominoes laid out to start with to force a recall/shuffle).  Some people lay out the cards so the graphics will load and be cached.  I guess those things probably are not compatible.  When I tested it on Dominoes, the tiles never displayed.  Not surprised.  And I don't really use that for pre-loading images.

Anyway, this is great and definitely meets the need.

ArnoldSmith86:
Other clients (that didn't click Play) will see the initial state.

96LawDawg:
Well, that is not as awesome then.  Didn't get that far in my testing.  Not sure this is fully useful if we can't overcome that.

robartsd:
> Some people lay out the cards so the graphics will load and be cached. I guess those things probably are not compatible. 

Yeah, the ideas really aren't compatible. The preload only works for clients that connect when all the cards are visible.


> Other clients (that didn't click Play) will see the initial state.

Won't the gameStartRoutine cause a delta to be created and distributed to other clients through the server?

ArnoldSmith86:
> Won't the gameStartRoutine cause a delta to be created and distributed to other clients through the server?

**Initially!** Guess I should have added that word. LawDawg, did you think the other clients would not receive the result of the gameStartRoutine or are you saying that you really don't like that they see the initial state until the routine finished and transfered the state to the other clients? Because I can see why it's a problem. If other players start to play around, they might cause conflicts which prevent the gameStartRoutine changes to be applied.

We could probably delay sending the initial state on the server until the routine results are incorporated.

96LawDawg:
> LawDawg, did you think the other clients would not receive the result of the gameStartRoutine or are you saying that you really don't like that they see the initial state until the routine finished and transfered the state to the other clients? Because I can see why it's a problem. If other players start to play around, they might cause conflicts which prevent the gameStartRoutine changes to be applied.
> 
> We could probably delay sending the initial state on the server until the routine results are incorporated.

I didn't know what I was saying.  But now having just tested it out on 2 clients, I see what you are saying.  I thought at first that the other players would not even see the effects of the `gameStartRoutine`.  But now I see they do, but it is delayed.  AND you are right: playing with the visible dominoes on the other client that didn't press the button while the start routine executes creates a delta that FUBARs the whole routine.  Need some sort of delay.

robartsd:
We should fix RECALL to be more efficient; but, since gameStartRoutine runs from a known state, game designers can avoid using RECALL in gameStartRoutine which will make it run much quicker (reducing the delay loading the game).

I agree that it would be best if all other clients just got the state after gameStartRoutine runs.

ArnoldSmith86:
Yeah, all of the above.

robartsd:
For development purposes, it would be good to have a way to load a state without triggering gameStartRoutine.

96LawDawg:
> For development purposes, it would be good to have a way to load a state without triggering gameStartRoutine.

How would that work?  I would just change the name of the routine from `gameStartRoutine` to `xgameStartRoutine` while doing development so it won't run.  Then remove the x when done.  I actually do that quite a bit in development when I've got a bunch of complicated stuff going on at once and I'm trying to isolate a problem.

96LawDawg (approved):
The new timing feature is great and I'm pretty sure it even loaded a lot faster.  Strange.

Approved through test.

robartsd:
> > For development purposes, it would be good to have a way to load a state without triggering gameStartRoutine.
> 
> How would that work? I would just change the name of the routine from `gameStartRoutine` to `xgameStartRoutine` while doing development so it won't run. Then remove the x when done. I actually do that quite a bit in development when I've got a bunch of complicated stuff going on at once and I'm trying to isolate a problem.

That works for the original development and testing, but would be somewhat difficult for someone who is starting from the deployed game.

96LawDawg:
> That works for the original development and testing, but would be somewhat difficult for someone who is starting from the deployed game.

Good point.  But the only way to do what you want that I can think of is to change something in the config.json file.  That isn't a bad idea to set up I guess.  Did you have another idea?

robartsd:
> > That works for the original development and testing, but would be somewhat difficult for someone who is starting from the deployed game.
> 
> Good point. But the only way to do what you want that I can think of is to change something in the config.json file. That isn't a bad idea to set up I guess. Did you have another idea?

I was thinking within the JSON editor there should be a way to load the saved state without running the routine. Could just be a Reset to Saved State button which would reload the last loaded saved state without triggering gameStartRoutine (after confirmation). Process would be select game from game shelf, open JSON editor, click Reset to Saved State button, confrim action.

ArnoldSmith86:
Do we want to merge this as is or wait for `SET` and maybe some JSON editor feature?

96LawDawg:
> Do we want to merge this as is or wait for `SET` and maybe some JSON editor feature?

Did you mean to post this in #1608?  That one is ready and there is no need to wait.  It will be more useful with `SET`, but that isn't necessary.

If you are asking about this one, I also think this is ready.  Right now, I don't see an easy way to do what @robartsd is asking.

ArnoldSmith86:
I guess the first part of the question was for the other PR and the second part for this one. No clue how that happened...

robartsd:
> Right now, I don't see an easy way to do what @robartsd is asking.

Doesn't the room keep track of which state it loaded now (so that saving can replace a saved game and put saved games under the correct heading)? I haven't really looked at the details of it, but I'm pretty sure the meta data has enough information that creating a button to reload the saved state should be possible.

ArnoldSmith86:
I estimate that it'll be a matter of weeks before we have an undo protocol in edit mode. It would show loading the game and its `gameStartRoutine` as separate entries you could undo. That should be good enough, right?

robartsd:
> I estimate that it'll be a matter of weeks before we have an undo protocol in edit mode. It would show loading the game and its `gameStartRoutine` as separate entries you could undo. That should be good enough, right?

How would that work, the editor isn't open when the game is loaded; would the editor also track changes while it is closed?

ArnoldSmith86:
> would the editor also track changes while it is closed

That would be my plan, yeah. So you can also use it to undo stuff while actually playing.

robartsd:
Then that would provide an adequate on-platform way to access the actual saved state.

PR #1596, opened 2023-02-09
Fix testcafe reference

### LINKED ISSUE #1595: Testcafe REFERENCE tests fail
Setting environmental variable REFERENCE is supposed to run tests against a PR test server, but the code points to the old PR test server instead of test.virtualtabletop.io.

For some reason, the tests fail to set the player name and color in Chrome in gitpod/workspace-full-vnc when running on test.virtualtabletop.io, so tests that depend on the player name and color being known fail.

robartsd:
Fixes #1595


robartsd:
Meant to separate from workspace-full-vnc.

ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1596/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1596-pr-test`.

robartsd:
I usually just use git to checkout main to run Testcafe reference tests - I guess everyone has been since REFERENCE has not worked for a while.

ArnoldSmith86 (approved):
The old URLs should still work but now with the actual PR number as port (without 3000 offset).

But I agree that it makes sense to use the test server URLs now.

PR #1599, opened 2023-02-10
Don't crash on non-object grid - fixes #1597

### LINKED ISSUE #1597: PL Game Imagination is Wrecked
So this is one of the strangest things I've ever seen on VTT.  Reported by user zoql on Discord.  Try loading Imagination (use a generic room).  You get an internal error message and nothing in the console.  Then that room is nuked and won't do anything else.  Obviously Imagination used to work.  So I went to the room where I initially loaded it after development https://virtualtabletop.io/Imagination.  Server error.  I tried loaded it from my personal archive of .vtt files before the UI changeover.  Same problem.  Very strange.

flackr:


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1599/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1599-pr-test`.

server/fileupdater.mjs
```diff
@@ -352,6 +352,8 @@ function v10GridOffset(properties) {
   if (!grid)
     return;
   for (let i = 0; i < grid.length; ++i) {
+    if (typeof grid[i] != 'object')
```
robartsd: Should also check for `null`

robartsd (approved):

server/fileupdater.mjs
flackr: Check the second patch :-)

flackr:
@96LawDawg feel free to test and merge or just merge as you please. Saves removing and readding the game from the PL :-).

server/fileupdater.mjs
```diff
@@ -349,7 +349,7 @@ function v9NumericStringSort(routine) {
 
 function v10GridOffset(properties) {
   const grid = properties.grid;
-  if (!grid)
+  if (!grid || typeof grid != 'object' || !grid.length)
```
robartsd, line 352: `!grid.length` is problematic here.

flackr: Oh is it? falsey grid.length should be skippable no?

robartsd: Should be `!Object.prototype.hasOwnProperty(grid,'length')`

robartsd: Could just check that grid is an array with `Array.isArray(grid)`

flackr: for in also works on arrays, and I know we had to use it in a few other places so that the code linter was satisfied that user input couldn't cause unbounded iteration lengths.

flackr: But also happy to have the array check if you prefer. Mostly I wanted to fix the crash.

96LawDawg:
Interesting.  The widgets there were never intended to move on a grid.  I just used "grid" as the name of a property (stupid in retrospect) as a grid coordinate reference system (A1, A2, etc).

96LawDawg (approved):
Approved through test.  Imagination now works.  Not sure what other tweaks you are making around the edges, but thanks for finding that so quickly.

flackr:
I think it's good now that it uses for in to loop over the array indices and no longer checks for a truthy length. This technique is used all over fileupdater rather than looping to length.

server/fileupdater.mjs
robartsd: I don't thing for in hurts. Code using grid iterates with for of.

robartsd:
I think it is good now too.

PR #1606, opened 2023-02-11
PL: Spin 5, Fix CSS Transitions

96LawDawg:
An earlier PR changed the behavior of transitions when changing parents.  That modified the behavior in Spin 5.  This PR fixes that game.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1606/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1606-pr-test`.

ArnoldSmith86:
Hmm, dropping into an already rotated holder results in rotated pieces. They fix themselves when rotating the holder but it's still pretty weird.

96LawDawg:
> Hmm, dropping into an already rotated holder results in rotated pieces. They fix themselves when rotating the holder but it's still pretty weird.

Good catch.  Fixed.

RaphaelAlvez (approved):
this seems to fix all the problems we found

PR #1608, opened 2023-02-11
GET supports sub-properties of objects

### LINKED ISSUE #1605: Make Complex Objects Easier to Use
Something like the following could possibly be made to work:

```
{
  "func": "SET",
  "collection": [ "widgetID" ],
  "property": "cardDefaults.color",
  "value": "#00767a"
}
```

Would be nice if it could work for other complex objects such as CSS as well.  Although that one might be harder.  For something like

```
"css": {
    "default": {
      "background": "white",
      "font-size": "25px"
    }
  }
```
would the property be `css.default.background` or something else?

ArnoldSmith86:
This is meant to close #1605 but so far it is only for `GET`.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1608/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1608-pr-test`.

96LawDawg:
This is great!  I tested it and it works with the most complicated css structures just fine.  It also works with cardTypes and cardDefaults.

The only unexpected thing (to me) was that it doesn't follow the convention for default naming in GET which is `id`.  If a variable is not specified, it uses the first level string in the property field.  That is actually better as a default.  We'll just need to document it.

mousewax:
:woohoo: 

Though I'm still not jazzed about that list structure. 

I *am* jazzed it works so well! :D

ArnoldSmith86:
> The only unexpected thing (to me) was that it doesn't follow the convention for default naming in GET which is `id`.  If a variable is not specified, it uses the first level string in the property field.  That is actually better as a default.  We'll just need to document it.

Should be the same on production. `variable` should default to the given `property`. Is that broken?

ArnoldSmith86:
> Though I'm still not jazzed about that list structure.

Feel free to make better suggestions. They'll have to be valid JSON though.

96LawDawg:
> Should be the same on production. `variable` should default to the given `property`. Is that broken?

Well, I just learned something new.  It does work that way on production.  And that is what is in the wiki.  I thought it was `id` because that is what the JSON Editor fills in for you by default.  But that isn't actually the default.  Good to know.

ArnoldSmith86:
> But that isn't actually the default.

It's the default default because `"id"` is the default `property`.

robartsd (approved):
GET.property:[] implementation looks good.

96LawDawg (approved):
Approved through test

ArnoldSmith86:
I'm hesitant to merge this because it changes the core code of `GET`. Can somebody please do a bit of thorough testing to see if the old `GET` behavior didn't change?

PR #1611, opened 2023-02-11
Adjust destination transform by element transform origin - fixes #1603

### LINKED ISSUE #1603: Transition location error when widget changes properties
In the PL game Spin 5, dragging one of the cards onto a holder when the holder has been rotated sometimes causes the card to appear to relocate from a strange location.

flackr:


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1611/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1611-pr-test`.

96LawDawg (approved):
Nobody but you understands the deep magic.  But this fixes the bug.  Approved through test.

flackr:
This fixes spin 5 (and even with nested rotations which I tested) so it's definitely better but I noticed that the "from" transform is actually not quite right on the perspective demo (before or with this PR) - so I'm still missing something :(

https://test.virtualtabletop.io/PR-1611/perspective

robartsd:
Perspective origin?

flackr:
> Perspective origin?

Maybe I need to offset by the element transform origin after applying perspective.

The way I've been debugging is if you set a breakpoint right before the call to invertTransform and apply the transform (before inverting) to `elem` the element should be transformed to exactly where it is dropping to.

flackr:
I figured out the issue. The "table" didn't have `transform-style: preserve-3d;` which means all 3d children are flattened into the table making it impossible to undo the table's transform. I've updated the test room to add preserve-3d to the table and the from transform perfectly matches the drop location.

There is a separate game issue - completely unrelated to this bug fix - that the 3d z position is behind the table if you drop the card too low because the table widget is actually "closer" to the screen.

flackr:
The TLDR is the current code works correctly in all of the cases i've tested so far. If there are other broken cases I'm happy to look into it - we could also do it in a later PR.

ArnoldSmith86 (approved):
I'm not sure what's happening but the code looks fine to me.

I'm not sure about performance in general though. VTT feels a lot slower on my phone recently and all these matrix operations probably don't help. Do you think we can try harder to avoid them where they aren't necessary?

Don't get me wrong: I think it's awesome that games just work with complicated transforms. But most games don't use scaling or rotating at all.

96LawDawg:
`transform-style: preserve-3d;` creates another strange issue on your table, however.  Make a pile of 2 cards (things work fine with more than 2).  Drag just the top card.  The card from the bottom of the pile disappears for a few seconds.

flackr:
> I'm not sure about performance in general though. VTT feels a lot slower on my phone recently and all these matrix operations probably don't help. Do you think we can try harder to avoid them where they aren't necessary?
> 
> Don't get me wrong: I think it's awesome that games just work with complicated transforms. But most games don't use scaling or rotating at all.

I'm happy to do some performance tracing and see where the biggest opportunities are. If we were doing the matrix math in Javascript I could see it being slow, but the implementation in browsers uses SIMD operations and is highly optimized as its doing stuff like this constantly. Also, I have lots of early outs (e.g. if the computed transform is none) to avoid matrix operations.

I'll do some tracing and see where most of the time is being spent.

flackr:
> `transform-style: preserve-3d;` creates another strange issue on your table, however. Make a pile of 2 cards (things work fine with more than 2). Drag just the top card. The card from the bottom of the pile disappears for a few seconds.

This is the part of the thing I was talking about. My perspective demo is not very good because the table is in front of content that is not in 3d. This happens without this PR as well.

Issue #1405, opened 2022-11-12
`CALL` without "routine" crashes client


96LawDawg:
A `CALL` function that does not include a "routine" parameter crashes the client.

ArnoldSmith86:
The defaults `widget: this.get('id'), routine: 'clickRoutine'` make sense though. But you usually have to change one of them. I don't think this is a problem with `CALL` and should just be solved by limiting the recursion depth and reporting it as a problem instead.

It might still be a bit weird though if the recursion depth is something like 100 (should probably be even higher) and the problem is reported in the deepest function. You would have to expand a lot of stuff in the debug output to get to the problem message...

96LawDawg:
Also mentioned in #1197.  Maybe a solution similar to #1424 would work.

96LawDawg:
What actually happened (as @ArnoldSmith86 obviously figured out), is the default "clickRoutine" executed on itself inside the "clickRoutine."  So that is why I got the maximum call stack size exceeded.  It was in an infinite recursion loop.

PR #1614, opened 2023-02-11
check that CALL.routine is string

robartsd:


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1614/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1614-pr-test`.

96LawDawg (approved):
It would be nice if we could also fix #1405.  I tried, but could not figure it out.  But that isn't necessary and could be a separate PR someday.  What is here works and is approved by test and code review.

Edit: #1615 is my effort to fix #1405.

PR #1615, opened 2023-02-12
Prevent `CALL` recursive loop

### LINKED ISSUE #1405: `CALL` without "routine" crashes client

96LawDawg:
Fixes #1405.  But the error is not reported until 100 layers deep as @ArnoldSmith86 said would probably happen.  That isn't ideal, but this does stop the client crash.  He also suggested making the limit greater than 100.  I don't know.  If you can make this better, have at it.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -7,6 +7,8 @@ import { tracingEnabled } from '../tracing.js';
 import { toHex } from '../color.js';
 import { center, distance, overlap, getOffset, getElementTransform, getScreenTransform, getPointOnPlane, dehomogenize, getElementTransformRelativeTo, getTransformOrigin } from '../geometry.js';
 
+let callStackDepth = 0;
+const maxCallStackDepth = 100;
 const readOnlyProperties = new Set([
   '_absoluteRotation',
   '_absoluteScale',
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1615/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1615-pr-test`.

robartsd:
100 is pretty deep. I wonder if we could check the call stack and push a warning to problems the first time a routine that is already in the call stack is CALLed.

robartsd:
This is creating a new global variable and incrementing it only when CALL occurs (and decrementing it when CALL returns). There is already a depth counter that is local to the running routine and gets passed to CALLed routines (as well as routines inside IF and FOREACH). Routines can also be triggered by CLICK, but that starts a new local depth counter (of course if an infinite loop involves both CLICK and CALL, the global depth counter would still increase until halting).

96LawDawg:
> 100 is pretty deep. I wonder if we could check the call stack and push a warning to problems the first time a routine that is already in the call stack is CALLed.

I tried adding something like:
```
if (callStackDepth < maxCallStackDepth - 2) {
   problems.push(`This is likely an infinite loop. Routine will stop after ${maxCallStackDepth} iterations.`);
}
```
But that didn't work.  And I don't understand the significance of what you said about there being a local depth counter.

robartsd:
The header for the evaluateRoutine method:
```  async evaluateRoutine(property, initialVariables, initialCollections, depth, byReference) {```
When called within IF, FOREACH, or CALL, the depth parameter is called with the calling routine's depth + 1.

robartsd:
I've created #1616 as an alternative approach (does not solve recursive CLICK).

PR #1616, opened 2023-02-12
detect and warn about recurssive CALL

robartsd:
This replaces depth counter in evaluate routine with a callstack array so that recursive calls to a routine can be detected.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -607,7 +607,7 @@ export class Widget extends StateManaged {
       reject(result);
   }
 
-  async evaluateRoutine(property, initialVariables, initialCollections, depth, byReference) {
+  async evaluateRoutine(property, initialVariables, initialCollections, callstack = [], byReference) {
     function unescape(str) {
       if(typeof str != 'string')
         return str;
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1616/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1616-pr-test`.

96LawDawg:
The practical effect is the same as #1615, right?  I'm not experienced enough to appreciate the coding differences, but the result in a room is the same, including the fact that you don't see the error reported until you dig down 100 levels (which by the way is impossible to do).

robartsd:
> The practical effect is the same as #1615, right? I'm not experienced enough to appreciate the coding differences, but the result in a room is the same, including the fact that you don't see the error reported until you dig down 100 levels (which by the way is impossible to do).

Yes, you do have to drill down through all the calls to see the final error, but this puts a warning in Problems of the first recursive call. Also, this allows more than 100 CALLS if they are to different routines (only calls to the same routine being called are counted).

#1615 does have the advantage that an infinite loop created with a CALL / CLICK pair is eventually halted - this does not halt that loop. We might want to implement something like #1615 for CLICK.

robartsd:
Another way to create an infinite loop that would avoid detection by the callstack filter would be to CLONE the widget with the recursive routine and CALL the routine in the clone.

robartsd:
Should this set `abortRoutine` to true when the limit is reached? I would say yes if `abortRoutine` bubbles up through the calling routines, but I don't think it does.

robartsd:
Perhaps routines need a way to raise exceptions to the routine that CALLed them. CALL could have an option to specify a routine to trigger to process an exception, otherwise the exception would get pushed to problems, abortRoutine would be set true, and the exception would be passed to a calling routine.

PR #1617, opened 2023-02-12
Change `fixedParent` Default to True for Most Widgets

96LawDawg:
This implements a request from Discord to make the default for `fixedParent` be true for everything except: basic widgets, cards, and piles.

I'm feeling pretty stupid and I can't figure out whether this needs a file updater or not.  I don't think so because the value is currently either true or false and not null or undefined for all widgets.  So changing the default shouldn't change anything for existing widgets but only set the default for new ones, right?

### DIFF
--- a/server/fileupdater.mjs
+++ b/server/fileupdater.mjs
@@ -437,10 +437,9 @@ function v12HandDropShadow(properties) {
 }
 
 function v13fixedParentDefaults(properties) {
-  if (properties.type == 'card' || properties.type == 'pile' || properties.type == null){
-    if (properties.fixedParent=true)
-      properties.fixedParent=true;
-  } else if (properties.fixedParent=false){
-    properties.fixedparent=false;
+  if (!(properties.type == 'card' || properties.type == 'pile' || properties.type == null)){
+    if (typeof properties.fixedParent === 'undefined' && (properties.movable == true)){
+    properties.fixedParent = false;
+    }
   }
 }
--- a/server/fileupdater.mjs
+++ b/server/fileupdater.mjs
@@ -1,4 +1,4 @@
-export const VERSION = 12;
+export const VERSION = 13;
 
 export default function FileUpdater(state) {
   const v = state._meta.version;
@@ -95,6 +95,7 @@ function updateProperties(properties, v, globalProperties) {
   v<8 && v8HoverInheritVisibleForSeat(properties);
   v<10 && v10GridOffset(properties);
   v<12 && globalProperties.v12DropShadowAllowed && v12HandDropShadow(properties);
+  v<13 && v13fixedParentDefaults(properties);
 }
 
 function updateRoutine(routine, v, globalProperties) {
@@ -432,5 +433,14 @@ function v11OwnerMOVEXY(routine) {
 function v12HandDropShadow(properties) {
   if (properties.type == 'holder' && properties.childrenPerOwner && !properties.enterRoutine && !properties.leaveRoutine && !properties.changeRoutine) {
     properties.dropShadow = true;
+  }
+}
+
+function v13fixedParentDefaults(properties) {
+  if (properties.type == 'card' || properties.type == 'pile' || properties.type == null){
+    if (properties.fixedParent=true)
+      properties.fixedParent=true;
+  } else if (properties.fixedParent=false){
+    properties.fixedparent=false;
   }
 }
--- a/client/js/widgets/basicwidget.js
+++ b/client/js/widgets/basicwidget.js
@@ -6,6 +6,7 @@ class BasicWidget extends Widget {
 
     this.addDefaults({
       typeClasses: 'widget basic',
+      fixedParent: false,
 
       faces: [ {} ],
       faceCycle: 'forward',
--- a/client/js/widgets/card.js
+++ b/client/js/widgets/card.js
@@ -3,6 +3,7 @@ class Card extends Widget {
     super(id);
 
     this.addDefaults({
+      fixedParent: false,
       width: 103,
       height: 160,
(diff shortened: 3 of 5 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1617/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1617-pr-test`.

robartsd:
> I can't figure out whether this needs a file updater or not. I don't think so because the value is currently either true or false and not null or undefined for all widgets. So changing the default shouldn't change anything for existing widgets but only set the default for new ones, right?

This does require a file updater. The widgets only store the value for a property if it is not the default value. For instance, this change will break Freecell which uses holders as cards.

96LawDawg:
Like I said, feeling stupid.  Hopefully the logic in this file updater makes sense.  It does fix Freecell.

server/fileupdater.mjs
```diff
@@ -434,3 +435,12 @@ function v12HandDropShadow(properties) {
     properties.dropShadow = true;
   }
 }
+
+function v13fixedParentDefaults(properties) {
+  if (properties.type == 'card' || properties.type == 'pile' || properties.type == null){
```
robartsd, line 440: Shouldn't need to do anything for these widget types; testing for basic widget might need to ensure that `properties` called with is actually a widget.

server/fileupdater.mjs
```diff
@@ -434,3 +435,12 @@ function v12HandDropShadow(properties) {
...
+    if (properties.fixedParent=true)
```
robartsd, line 441: In javascript a single `=` is assignment not comparison.

server/fileupdater.mjs
```diff
@@ -434,3 +435,12 @@ function v12HandDropShadow(properties) {
...
+      properties.fixedParent=true;
+  } else if (properties.fixedParent=false){
```
robartsd, line 443: Need to check if fixedParent is not set, not if it is set to false.

server/fileupdater.mjs
```diff
@@ -434,3 +435,12 @@ function v12HandDropShadow(properties) {
...
+    properties.fixedparent=false;
+  }
+}
```
robartsd, line 446: There should be an empty line at the end of the file.

ArnoldSmith86:
Maybe we only set `fixedParent: false` if `movable: true` (I didn't look if you did that yet)?

server/fileupdater.mjs
96LawDawg: I think the changes I made address this, but not sure about "testing for basic widget might need to ensure that properties called with is actually a widget."  I don't really understand that.

server/fileupdater.mjs
96LawDawg: Yep.

server/fileupdater.mjs
96LawDawg: Done.

server/fileupdater.mjs
96LawDawg: Done.

96LawDawg:
> Maybe we only set `fixedParent: false` if `movable: true` (I didn't look if you did that yet)?

Good idea.  Added that.

Overall, I still have sort of low confidence in my file updater.  This one shouldn't be as hard as some of the others I've worked on, but I just can't wrap my head around it right.

server/fileupdater.mjs
robartsd: `properties` object passed to the function could be an element of a `BasicWidget.faces` or `Deck.cardTypes`. Of course those don't have a `type`, so I guess they still will get filtered out the same as basic widget.

robartsd:
> Maybe we only set `fixedParent: false` if `movable: true` (I didn't look if you did that yet)?

Scoreboard & Spinner default to movable: true, so they need an exception to this rule.

robartsd:
Actually thinking about movable: true, makes me think that we are treating the symptom instead of the problem. The problem is that moving things in edit mode changes the widget tree unintentionally. Perhaps instead of changing the default value of fixedParent, we change the behavior for all widgets that are being dragged but have movable set to false. Then no fileupdater is needed; we are would only be changing editor behavior.

96LawDawg:
> Actually thinking about movable: true, makes me think that we are treating the symptom instead of the problem. The problem is that moving things in edit mode changes the widget tree unintentionally. Perhaps instead of changing the default value of fixedParent, we change the behavior for all widgets that are being dragged but have movable set to false. Then no fileupdater is needed; we are would only be changing editor behavior.

I could get behind that idea.  And maybe we could fix #12 while at it.  But I don't think I'm able to do either of those things.

RaphaelAlvez:
Ok so I agree that this should be setup at the default for most widgets. At the same time I don't think this is a bad thing. Yes this feature was added because we were breaking movable widgets all the time but it turns out is is usable  and solves the problem in most situations

robartsd:
> > Actually thinking about movable: true, makes me think that we are treating the symptom instead of the problem. The problem is that moving things in edit mode changes the widget tree unintentionally. Perhaps instead of changing the default value of fixedParent, we change the behavior for all widgets that are being dragged but have movable set to false. Then no fileupdater is needed; we are would only be changing editor behavior.
> 
> I could get behind that idea. And maybe we could fix #12 while at it. But I don't think I'm able to do either of those things.

Changing the move behavior should be fairly simple, basically just modify the fixedParent test, in Widget.move(). I'll try to throw together a PR.

While is #12 is related from a user interface point of view, it is really not related from a code point of view.

96LawDawg:
Abandoned for #1618.

PR #1618, opened 2023-02-13
Behave as if fixedParent is true if only movable in edit

robartsd:
It is very easy to accidentally remove the parent of a widget when in edit mode. This makes all widgets that have movable set to false behave as if fixedParent is set to true when being moved in edit mode.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -1867,7 +1867,7 @@ export class Widget extends StateManaged {
     await this.bringToFront();
     await this.set('dragging', playerName);
 
-    if(!this.get('fixedParent' && this.get('movable'))) {
+    if(!this.get('fixedParent') && this.get('movable')) {
       this.dropTargets = this.validDropTargets();
       this.currentParent = widgets.get(this.get('_ancestor'));
       this.hoverTarget = null;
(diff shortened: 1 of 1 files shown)


robartsd:
This is an alternative to #1617 that does not actually change widget properties or behavior in game mode, so no file updater should be required.

ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1618/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1618-pr-test`.

96LawDawg:
Doesn't work.

robartsd:
> Doesn't work.

Opps. I pushed without testing before heading to my mom's for dinner. It works now.

96LawDawg (approved):
Approved through test.  This is a good solution.

robartsd:
What is especially nice is that it works for all widget types - if you set them so that the player can't move them, they no longer deparent when moved in edit. Most of the widget types that we talked about defaulting to fixedParent: true also default to movable: false. 'deck' is a bit weird; even though it does not appear in play mode, it is 'movable', so it will not behave as if it has fixedParent true by default, but that is needed for the edit mode UI design to drag and drop a deck between holders.

flackr (approved):
Code looks good to me.

flackr:
Also, I like this approach. I've accidentally moved things out of parents in edit mode so often!

ArnoldSmith86 (approved):
I like it. We'll see if we prefer the old way in a few months... :smile:

PR #1625, opened 2023-02-15
'palette' input type

mousewax:
Added a simplified 'palette' input type that takes a list of colors as options. 

```
        {
          "type": "palette",
          "label": "Color",
          "options": [
            "#897349",
            "#89dd98",
            "#fff",
            "#dbdc00"
          ],
          "variable": "playerColor",
          "value": "${playerColor}"
        }
```
image

It uses --VTTblue as the selected outline and checkmark color.

----------------
Proposed wiki update:

* `type`: 
......
    * `palette`_displays colored radio buttons_-stores the selected value. Unlike the `color` field, this offers a consistent viewing experience across browsers and forces the players to choose from a palette of options.
      * `colors`: _array_ - displays the color choices available to the player. Accepts most color formats (html-friendly, hex, rgb, rgba, hsl, hsla).

### DIFF
--- a/client/css/layout.css
+++ b/client/css/layout.css
@@ -1366,7 +1366,7 @@ input:invalid+span:after {
 .inputpalette > div{
   display: flex;
   flex-wrap: wrap;
-  padding: 0 5px;
+  padding: 0 2px;
 }
 .inputpalette > div input{
   display: none;
@@ -1377,10 +1377,8 @@ input:invalid+span:after {
   display: inline-block;
   border-radius: 8px;
   position: relative;
-  border: 3px solid var(--roomColor);
-  padding: 0px;
-  overflow: hidden;
-  
+  border: 3px solid transparent;
+  padding: 0;
 }
 .inputpalette > div input[type="radio"]:checked+label{
   border: 3px solid var(--VTTblue);
@@ -1393,13 +1391,16 @@ input:invalid+span:after {
   width: 1.25em;
   background: var(--VTTblue);
   color: white;
-  border-radius: 100%;
-  border-top-right-radius: 0;
+  border-top-right-radius: 40%;
+  border-top-left-radius: 50%;
+  border-bottom-right-radius: 50%;
+  border-bottom-left-radius: 50%;
   content: 'check';
   position: absolute;
   top: -3px;
   right: -3px;
   border: 2px solid white;
+  z-index: -1;
 }
 .inputpalette > div label > div{
   display: block;
@@ -1414,4 +1415,5 @@ input:invalid+span:after {
   overflow: hidden;
   background-color: inherit;
   border-radius: 5px;
+  z-index: -2;
 }
--- a/client/css/layout.css
+++ b/client/css/layout.css
@@ -1414,5 +1414,4 @@ input:invalid+span:after {
   overflow: hidden;
   background-color: inherit;
   border-radius: 5px;
-  box-shadow: 0 0 1px white;
 }
--- a/client/css/layout.css
+++ b/client/css/layout.css
@@ -1378,11 +1378,11 @@ input:invalid+span:after {
   display: inline-block;
   border-radius: 5px;
   position: relative;
-  border: 2px solid var(--roomColor);
+  border: 3px solid var(--roomColor);
   overflow: hidden;
 }
 .inputpalette > div input[type="radio"]:checked+label{
-  border: 2px solid var(--VTTblue);
+  border: 3px solid var(--VTTblue);
 }
 .inputpalette > div input[type="radio"]:checked+label::after{
   font-family: 'Material Icons';
@@ -1398,18 +1398,18 @@ input:invalid+span:after {
   position: absolute;
   top: -3px;
   right: -3px;
-  border: 2px solid var(--roomColor);
+  border: 2px solid white;
 }
 .inputpalette > div label::before{
+  display: block;
   content: '';
   position: absolute;
-  
-  top: -1px;
-  right: -1px;
-  bottom: -1px;
-  left: -1px;
+  top: 0px;
+  right: 0px;
+  bottom: 0px;
+  left: 0px;
   position: absolute;
-  border-radius: 5px;
-  border: 3px solid white;
+  border-radius: 2px;
+  border: 2px solid white;
   overflow: hidden;
 }
--- a/client/css/layout.css
+++ b/client/css/layout.css
@@ -1366,7 +1366,7 @@ input:invalid+span:after {
 .inputpalette > div{
   display: flex;
   flex-wrap: wrap;
-  padding: 0 2px;
+  padding: 0 3px;
 }
 .inputpalette > div input{
   display: none;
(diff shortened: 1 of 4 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1625/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1625-pr-test`.

96LawDawg:
> It uses --VTTblue as the selected outline and checkmark color.

That won't work when you have a dark blue playerColor.  Use the colorContrast function.  Maybe I can figure out how to make that happen.  I'll look.

robartsd:
> > It uses --VTTblue as the selected outline and checkmark color.
> 
> That won't work when you have a dark blue playerColor. Use the colorContrast function. Maybe I can figure out how to make that happen. I'll look.

There is still the white check mark, so even if the color is exactly --VTTblue there will still be some indication. Tricky part is that it should have adequate contrast between both the selected color and white.

96LawDawg:
That's the magic of the colorContrast function.

96LawDawg:
I can't figure it out.  When nothing worked, to dumb it down, I tried just putting a simple console.log() into domhelper.js where your code is.  Never worked.  domhelper must do something different than I'm used to.  Here's what I was trying to do:
 - Get the currently selected field.value and determine its color and put it into some variable, say "contrast."
 - Call colorContrast(contrast).
 - Instead of --VTTBlue, use the result of the colorContrast function.  As is with the default parameters you get white or black, but you could also pass a value in the function to make it some shade of a similar hue.
 - Then make the checkmark color itself the same as the original.

That was my idea any way.  Because this doesn't have quite the same zing:
Blue

mousewax:
I don't get what you mean by `options` is not working right. It seems to be working fine to create radio inputs. Am I missing something?

96LawDawg:
Just the JSON Editor default stuff is not right.

When you use the JSON Editor to select the palette field and then press the "options" button you get:

```
       {
          "type": "palette",
          "options": [
            {
              "value": "value",
              "text": "text"
            }
          ]
        }
 ``

Because that is set up for the "select" type field.  Options should just be an empty array:  [    ].  I can't get that to work.

**Comment by mousewax** [comment]
Input elements and UI should generally be consistent so I added a contrasty border rather than change the color.

image

**Comment by 96LawDawg** [comment]
> Input elements and UI should generally be consistent so I added a contrasty border rather than change the color.

That works, is consistent, and doesn't require all the code shenanigans I had come up with.

One thing doesn't work:  Getting the color value out of the INPUT function.  It just returns null.

**Comment by robartsd** [comment]
> I made the JSON Editor mostly work with the new palette field. However, options doesn't work because it shows the options for the select field. I can't figure out how to get it to show something different. The easiest way would be to change "options" to something else like "colors."

I'm in favor of this change

```
        {
          "type": "palette",
          "label": "Color",
          "colors": [
            "#897349",
            "#89dd98",
            "#fff",
            "#dbdc00"
          ],
          "variable": "playerColor",
          "value": "${playerColor}"
        }
```

mousewax:
Unfortunately I am not sure why it's passing null. The HTML looks fine and I guess I expected it to pass right through like any other form element.

robartsd:
> Unfortunately I am not sure why it's passing null. The HTML looks fine and I guess I expected it to pass right through like any other form element.

The code for INPUT reads the input from the DOM. I don't think it knows how to read radio button inputs yet.

mousewax:
> The code for INPUT reads the input from the DOM. I don't think it knows how to read radio button inputs yet. Widget.evaluateInputOverlay() needs to be updated to use the new input.

Thanks! I think I did it correctly.

robartsd:
> > The code for INPUT reads the input from the DOM. I don't think it knows how to read radio button inputs yet. Widget.evaluateInputOverlay() needs to be updated to use the new input.
> 
> Thanks! I think I did it correctly.

I'd use .filter() or .reduce() rather than a for loop, but this should work.

robartsd (approved):

ArnoldSmith86 (approved):
Code looks fine. I didn't test anything.

mousewax:
> I'd use .filter() or .reduce() rather than a for loop, but this should work.

I don't quite understand how to use those effectively without still looping over the smaller array. I'd love to see how they work in practice. I am just not that smart.

mousewax:
In the `colors` section it will accept any html-friendly color (like "red" or "blue") but the system converts them to hex colors so it will not re-check the selected color when you re-open the input.  Not the end of the world, but it should probably be noted in the wiki.

ArnoldSmith86:
We can compare to `toHex(color)`.

RaphaelAlvez:
https://user-images.githubusercontent.com/73476656/219034109-a7bdcb9d-c1c2-4caf-af80-7f37d2dd8534.mp4

Minor visual bug when takes some time to load the font. Testing it on my phone on the street so maybe we should just ignore it.

mousewax:
I noticed that on the PR test server as well. It will be the same with any other element where the font has not yet loaded.

mousewax:
> We can compare to `toHex(color)`.

I didn't realize that was a thing. It's been added.

96LawDawg (approved):
Approved through test.

However, there is one small change I request if possible.  Something about the rounding of the corners and the white border is not quite right.  You can see small bits of color peeking out at the corners.  Can you try to fix that?

Corners

96LawDawg:
I added a proposed wiki update at the very top.  Also tested with hsl, hsla, rgb, and rgba and they all work fine.

96LawDawg:
> I believe I have now fixed this.

Yes.  Should be ready to merge this in 24 hours.

robartsd:
> I added a proposed wiki update at the very top. Also tested with hsl, hsla, rgb, and rgba and they all work fine.

Any way the browser running the code supports that results in an sRGB color should work. Not sure what happens in browsers that support colors outside of sRGB (I believe only Safari so far) if a color outside sRGB is specified (preferred behavior would be the closest sRGB color, may get black).

96LawDawg:
> Any way the browser running the code supports that results in an sRGB color should work. Not sure what happens in browsers that support colors outside of sRGB (I believe only Safari so far) if a color outside sRGB is specified (preferred behavior would be the closest sRGB color, may get black).

I'm pretty sure toHex returns black in that situation.  I don't think we got that far in the color function PR.

robartsd:
> > Any way the browser running the code supports that results in an sRGB color should work. Not sure what happens in browsers that support colors outside of sRGB (I believe only Safari so far) if a color outside sRGB is specified (preferred behavior would be the closest sRGB color, may get black).
> 
> I'm pretty sure toHex returns black in that situation. I don't think we got that far in the color function PR.

Yes we use the fillstyle property of a canvas element to evaluate colors. Safari (and Chrome > 110) support the `color()` function (but it is possible that it returns colors outside sRGB), but Firefox does not. All supported browsers should support `hwb()` color function (which also always returns sRGB).

Issue #993, opened 2021-12-08
MOVE and alignChildren:false do not work as expected for non-card widgets


96LawDawg:
Look at the new MOVE tutorial.  The widgets in the far-right area are moved one at a time to a holder with alignChildren: true.  That makes them stack.  If you set alignChildren:false, I would have expected the widgets to show up in the new holder in relatively the same position they do in the current holder.  Instead, they get moved to the room absolute coordinates that were the equivalent of the old holder relative coordinates.

96LawDawg:
Forgot to mention that although this isn't related to #373, it would seem efficient to resolve those together.

robartsd:
#373 is not related and is not likely to touch any relevant code.

MOVE simply calls `moveToHolder()` on the widget. `moveToHolder()` calls `this.checkParent(true)` which sets `parent` to `null`. When execution returns to `moveToHolder()` the widget is functionally at the room coordinates. `moveToHolder()` then sets the new `parent`. Eventually code execution reaches `holder.onChildAddAlign()` which translates the room coordinates to the relative coordinates within the holder. I don't know why this would not behave the same for cards as other widgets.

96LawDawg:
It does behave the same for cards.  I just hadn't tested that.  Cards are usually in piles and/or aligned in holders in my experience.

ArnoldSmith86:
I honestly have no idea what I expect. The `MOVE` is asked not to align the widget but it has to be put _somewhere_.

robartsd:
> I honestly have no idea what I expect.

I expect `MOVE` to behave exactly as `SET` `parent` unless the new parent widget has `alignChildren:true`.

96LawDawg:
> I expect `MOVE` to behave exactly as `SET` `parent` unless the new parent widget has `alignChildren:true`.

Yes.  That is it exactly.  More artfully said than I did.  But of course that is not happening because set parent would move it to the new holder relative position and not the room absolute position.

96LawDawg:
Drawing on something Ingegneus said (https://discord.com/channels/770758631146782780/795657499520663622/918051298950119436), I tested MOVE from one basic widget to another basic widget.  It works just like SET parent would.

robartsd:
> Drawing on something Ingegneus said (https://discord.com/channels/770758631146782780/795657499520663622/918051298950119436), I tested MOVE from one basic widget to another basic widget. It works just like SET parent would.

[...]

PR #1628, opened 2023-02-15
moveToHolder only moves if not already in holder

robartsd:
Previously, Widget.moveToHolder(holder) force detached the widget from it's parent if the parent was not holder, even if the parent was a pile in the holder. This now checks if the widget's _ancestor is the holder instead. This significantly improves RECALL for large decks that are mostly still in the pile in the deck's holder. 

Testing results: On production, Start button in Cards Against Humanity took 336ms for the RECALL step immediately after loading the game. With this PR the step took 6ms.

I'm pretty sure that the efficiency is from removing a lot of extraneous DOM tree manipulation.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -1854,8 +1854,9 @@ export class Widget extends StateManaged {
     const curAncestor = this.get('_ancestor');
     if(curAncestor && !this.currentParent)
       this.currentParent = widgets.get(curAncestor);
-    if(this.currentParent != holder) {
+    if(this.currentParent != holder)
       await this.checkParent(true);
+    if(this.currentParent != holder || this.get('owner') || holder.get('childrenPerOwner')) {
       await this.set('owner',  null);
       await this.set('parent', holder.get('id'));
     }
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -1851,13 +1851,14 @@ export class Widget extends StateManaged {
       return;
 
     await this.bringToFront();
-    if(this.get('parent') && !this.currentParent)
-      this.currentParent = widgets.get(this.get('parent'));
-    if(this.currentParent != holder)
+    const curAncestor = this.get('_ancestor');
+    if(curAncestor && !this.currentParent)
+      this.currentParent = widgets.get(curAncestor);
+    if(this.currentParent != holder) {
       await this.checkParent(true);
-
-    await this.set('owner',  null);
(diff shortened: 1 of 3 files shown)


robartsd:
Possible side effects that may need addressing.

- MOVE to a childrenPerOwner holder does not ensure the widget is owned by the player whose client is running the routine (owner is not changed if the widget is already in the holder)
- MOVE to a holder with alignChildren: false will not move the widget to (0,0) #993

RaphaelAlvez:
Should this account for seats' hand?

robartsd:
> Should this account for seats' hand?

I don't think there is any accounting for that needed here; but changes for childrenPerOwner holders might be needed.

client/js/widgets/widget.js
```diff
@@ -1851,13 +1851,15 @@ export class Widget extends StateManaged {
...
-      this.currentParent = widgets.get(this.get('parent'));
+    const curAncestor = this.get('_ancestor');
+    if(curAncestor && !this.currentParent)
+      this.currentParent = widgets.get(curAncestor);
     if(this.currentParent != holder)
       await this.checkParent(true);
-
-    await this.set('owner',  null);
-    await this.set('parent', holder.get('id'));
+    if(this.currentParent != holder || this.get('owner') || holder.get('childrenPerOwner')) {
```
flackr, line 1859: This could probably be separate checks for each of the set calls, i.e.

```js
if (this.get('owner') || holder.get('childrenPerOwner'))
  await this.set('owner', null);
if (this.currentParent != holder)
  await this.set('parent', holder.get('id'));
```

Or do we need to set the parent if we have owner / childrenPerOwner set?

flackr (approved):
Code-wise this looks correct to me.

client/js/widgets/widget.js
robartsd: I'm not 100% sure what we need to do when widget is owned or childrenPerOwner is set on the target holder. I need to do some tests with MOVE with from and to set to the same childrenPerOwner holder (with and without piles in the holder) to see what production does.

96LawDawg:
#1923 allows something similar to be implemented manually in the JSON if needed to help optimize recalls.

```
   {
      "func": "SELECT",
      "property": "_ancestor",
      "value": "holder1"
    },
    {
      "func": "RECALL",
      "excludeCollection": "DEFAULT",
      "holder": "holder1"
    }
```

96LawDawg:
Superseded by #1948

PR #1636, opened 2023-02-19
Add option to clone recursively - fixes #1343

### LINKED ISSUE #1343: option in CLONE to recursively clone children
it would be very convenient to have an option to activate a similar feature in the CLONE function like in the JSON editor

flackr:
Adds a new option to the [CLONE](https://github.com/ArnoldSmith86/virtualtabletop/wiki/Functions#clone) function to recursively clone children. Fixes #1343.


flackr:
Maybe this should be called deep to align with the DOM [Node.cloneNode](https://developer.mozilla.org/en-US/docs/Web/API/Node/cloneNode) function. Also should this be the default?

ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1636/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1636-pr-test`.

RaphaelAlvez (approved):
by code review this looks ok. your changes are minimal.

I just don't really understand whats the difference between the 2 types of cloning/duplicate so I can't test it

96LawDawg (approved):
Add the option to the JSON Editor.  Approved through test.

96LawDawg:
> Also should this be the default?

I don't think so.  Most game uses I've seen would be for a nonrecursive clone (but maybe because there wasn't an option before).  Also, that would require a fileupdater I think.  I'd leave it off by default.

PR #1639, opened 2023-02-19
Add childrenCounterRotate feature

flackr:
This feature enables children to keep their relative orientation when dropped into a rotated holder.

### DIFF
--- a/client/js/widgets/holder.js
+++ b/client/js/widgets/holder.js
@@ -16,6 +16,7 @@ class Holder extends Widget {
       dropShadow: false,
       alignChildren: true,
       preventPiles: false,
+      childrenCounterRotate: false,
       childrenPerOwner: false,
 
       onEnter: {},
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1639/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1639-pr-test`.

flackr:
I'm also considering whether this should invert other transforms. I'll have too see how this would work with dropAnchor and perspective if I wanted to have tokens that stand up on a 3d board. I think since you still want the local position transform it probably only makes sense to undo the rotation.

flackr:
I have a fix for the drop location but it's much simpler after #1641 as that deduplicates a bunch of the calculation logic.

RaphaelAlvez:
@flackr I don't know what is happening but I created a holder and it is not working

RaphaelAlvez:
for some reason, it is working now....

there is a weird behavior with piles

RaphaelAlvez (approved):
tested

PR #1641, opened 2023-02-20
Use the same positioning code for the drop shadow as moveend

### LINKED ISSUE #1640: dropShadow doesn't get correct position for rotated containers
The calculation of the position for the drop shadow isn't the same as the one used when dropping the widget. This results in the drop shadow showing in a different location than it drops to. For example try the demo at:

https://virtualtabletop.io/clip-path-holder

You can see that in the rotated holders the position of the shadow is not the place it drops to as the `moveEnd` code does some extra repositioning to align the anchor point in the transformed container. This incorrect location could result in inconsistent orders in rotated aligned containers where while dragging it looks like you'll drop in one place and end up dropping in another.

flackr:
By sharing the drag location code between the shadow widget location, `moveEnd` and `move`, there's no longer any inconsistencies in where widgets are between these methods. Fixes #1640.

This also passes the global coordinate to `moveEnd` rather than the client coordinate for consistency with the signature of the `move` function.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1641/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1641-pr-test`.

RaphaelAlvez (approved):
I don't feel confident making a code review on this but I did a test review. It does sem to fix the bug. I compared the file loaded in the room ("test") with the one loaded in https://virtualtabletop.io/9gu1

ArnoldSmith86 (approved):
Code looks fine to me. A bit hard to see if anything changed in the refactor but looks promising. Dragging and hover target still feel normal so can't be _that_ wrong.

PR #1642, opened 2023-02-20
Optimize DOM reads and modifications

flackr:
This is a pretty big change and currently still broken but I wanted to get it out there to see if anyone has any alternate suggestions or if there's a better way to do this. The TLDR is one of the biggest performance hits in virtualtabletop on my slow computer is forced synchronous style and layout. You can read more about the general issue at:

https://web.dev/avoid-large-complex-layouts-and-layout-thrashing/#avoid-forced-synchronous-layouts

This patch splits up `applyDelta` and `applyDeltaToDOM` into three phases in order to reduce forced layouts as much as possible.
1. A measure phase - the initial phase. During this phase it is fine to read style and layout because nothing has invalidated it yet.
2. DOM modification phase. During this phase any modifications can be made to style and layout. Reading back during this should be avoided and moved to the measure phase if possible.
3. A post modification phase. This is really only for initiating transitions which require a separate style change, but could be used for other tasks that want to happen after all dom modification - e.g. maybe measuring the result of the modified dom.

On my slow machine this brought a game initialization which used to take about 30s down to a couple seconds.

- [ ] Fix populateAddWidgetOverlay crash
- [ ] Get it more robust...

### DIFF
--- a/client/js/editmode.js
+++ b/client/js/editmode.js
@@ -18,8 +18,18 @@ async function addWidgetLocal(widget) {
   }
 
   const isNewWidget = !widgets.has(widget.id);
-  if(isNewWidget)
-    addWidget(widget);
+  if(isNewWidget) {
+    // Ideally we'd pass these promises in and resolve them immediately.
+    let modifyDOM = new Promise((resolve) => {
+      setTimeout(resolve, 0);
+    });
+    let afterModify = new Promise((resolve) => {
+      modifyDOM.then(() => {
+        setTimeout(resolve, 0);
+      });
+    });
+    addWidget(widget, undefined, modifyDOM, afterModify);
+  }
   sendPropertyUpdate(widget.id, widget);
   sendDelta();
   batchStart();
@@ -769,7 +779,7 @@ function addCompositeWidgetToAddWidgetOverlay(widgetsToAdd, onClick) {
     if(wi.type == 'pile')   w = new Pile(wi.id);
     if(wi.type == 'timer')  w = new Timer(wi.id);
     widgets.set(wi.id, w);
-    w.applyInitialDelta(wi);
+    w.applyInitialDelta(wi, Promise.resolve(), Promise.resolve());
     w.domElement.id = w.id;
     if(!wi.parent) {
       w.domElement.addEventListener('click', async _=>{
@@ -788,7 +798,7 @@ function addCompositeWidgetToAddWidgetOverlay(widgetsToAdd, onClick) {
 const VTTblue = '#1f5ca6';
 
 function addPieceToAddWidgetOverlay(w, wi) {
-  w.applyInitialDelta(wi);
+  w.applyInitialDelta(wi, Promise.resolve(), Promise.resolve());
   w.domElement.addEventListener('click', async _=>{
     try {
       const result = await w.showInputOverlay({
@@ -814,7 +824,7 @@ function addPieceToAddWidgetOverlay(w, wi) {
 }
 
 function addWidgetToAddWidgetOverlay(w, wi) {
-  w.applyInitialDelta(wi);
+  w.applyInitialDelta(wi, Promise.resolve(), Promise.resolve());
   w.domElement.addEventListener('click', async _=>{
     const toAdd = {...wi};
     toAdd.z = getMaxZ(w.get('layer')) + 1;
@@ -1368,7 +1378,7 @@ function populateAddWidgetOverlay() {
     x: 450,
     y: 835
   };
-  spinner.applyInitialDelta(spinAttrs);
+  spinner.applyInitialDelta(spinAttrs, Promise.resolve(), Promise.resolve());
   spinner.domElement.addEventListener('click', async _=>{
     try {
       const result = await spinner.showInputOverlay({
@@ -1965,5 +1975,6 @@ onLoad(function() {
       $('#buttonImage').value = asset;
   }));
 
-  populateAddWidgetOverlay();
+  // TODO: Figure out why this crashes.
+  //populateAddWidgetOverlay();
 });
--- a/client/js/serverstate.js
+++ b/client/js/serverstate.js
@@ -15,7 +15,7 @@ let deltaID = 0;
 let batchDepth = 0;
 let overlayShownForEmptyRoom = false;
 
-export function addWidget(widget, instance) {
+export function addWidget(widget, instance, modifyDOM, afterModify) {
   if(widget.parent && !widgets.has(widget.parent)) {
     if(!deferredChildren[widget.parent])
       deferredChildren[widget.parent] = [];
@@ -31,6 +31,8 @@ export function addWidget(widget, instance) {
 
   let w;
 
+  // TODO: Many of the constructors perform DOM modifications that should be
+  // delayed until modifyDOM for performance.
   if(instance != undefined) {
     w = instance;
   } else if(widget.type == 'card') {
@@ -73,7 +75,7 @@ export function addWidget(widget, instance) {
 
   try {
     widgets.set(widget.id, w);
-    w.applyInitialDelta(widget);
+    w.applyInitialDelta(widget, modifyDOM, afterModify);
   } catch(e) {
     console.error(`Could not add widget!`, widget, e);
     removeWidget(widget.id);
@@ -84,11 +86,11 @@ export function addWidget(widget, instance) {
 
   if(widget.type == 'deck')
     for(const c of deferredCards[widget.id] || [])
-      addWidget(c);
+      addWidget(c, undefined, modifyDOM, afterModify);
   delete deferredCards[widget.id];
 
   for(const c of deferredChildren[widget.id] || [])
-    addWidget(c);
+    addWidget(c, undefined, modifyDOM, afterModify);
   delete deferredChildren[widget.id];
 }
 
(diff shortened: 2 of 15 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1642/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1642-pr-test`.

ArnoldSmith86:
Sounds very interesting. I probably won't have time to look at it in detail for a few days.

ArnoldSmith86:
Very nice.

ArnoldSmith86:
So this basically doesn't change the code? Only the order so that all changes to the DOM are "batched together"?

flackr:
> So this basically doesn't change the code? Only the order so that all changes to the DOM are "batched together"?

Yes that's exactly it. By doing all of the reads of layout information and style before any modifications the information is already up to date and no style/layout recalculations (which are unfortunately global on the web) are needed. Then we do all of the DOM modifications without reading anything and that way the browser updates style/layout only once after all the changes have been made.

ArnoldSmith86:
So this is probably a big speedup across everything, right? Not only complex routines?

flackr:
Yes! This will be a big win generally. It's just especially noticeable when you've got big complex routines doing lots of changes but it will make all dom updates faster.

flackr:
Might even be simpler to - rather than pass global promises through the update calls - just call a method which runs a callback after a resolved promise since resolved promises only run once the current execution stack is empty.

From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises
> Instead of running immediately, the passed-in function is put on a microtask queue, which means it runs later (only after the function which created it exits, and when the JavaScript execution stack is empty), just before control is returned to the event loop

This would mean no changes to the function signatures and instead having something like:
```js
applyDeltaToDOM(delta) {
  schedule(() => { // schedule calls Promise.resolve().then(fn)
    /* dom mutation work
  });
});
```

But being less structured there's more room for things to go wrong. My problem with the structure is I'm not sure what the right setup point for routine evaluation is, i.e. where I should setup the promises and execute them after. I've just used a `setTimeout` in `cloneLocal` which should be replaced with something higher up to capture all of the routine modifications.

flackr:
I was also considering whether I should just have a list of callbacks to invoke rather than using a resolved promise - which would I think would be a bit more controlled in terms of when things happen. The promise doesn't execute until the current execution stack finishes which to me suggests it is after `receiveDelta`returns so you couldn't call receiveDelta and expect it to have been applied yet.

robartsd:
This sounds great. As a frequent user of relatively low end computers (which was a big reason for seeking out PCIO which led to VTT).

robartsd:
Conceptually, I think we might want to think about this as 5 phases:

- Update state (this already happens before applyDeltaToDOM)
- Pre-modification DOM measurements
- DOM mutations from state chage
- Post-modification DOM measurements (avoid when possible)
- DOM mutations that depend on post-modification measurements (avoid when possible; MUST not depend on measuring changes introduced in this phase)

flackr:
Technically it could be schedule another measure/read style phase, schedule another modification phase and repeat as much as you like, but we really shouldn't ever need more than one repeat of measure / modify which represents one extra style/layout calc since the browser has to do it at least once for the new frame.

My current thinking is maybe passing a scheduler class into `applyDeltaToDOM` with functions that take callbacks, making it look something like this:
```js
applyDeltaToDOM(delta, scheduler) {
  // state modification can just happen right away.
  scheduler.measure(() => {
    // store some vars in closure for use in modify once all scheduled measures finish.
    scheduler.modify(() => {
      // dom modiifications

      // Getting into not recommended territory.
      scheduler.measure(() => {
        scheduler.modify(() => {
          // Maybe this throws because you shouldn't need this many style recalcs.
          scheduler.measure(() => {});
        });
      });
    });
  });
});
```

robartsd:
I think we can figure out what needs post update measures, but it could be difficult to associate the code for those changes with all of the changes that they might depend on.

I'm thinking something more like this:

```js
applyDeltaToDOM(delta, scheduler) {
  // state modification already happens before applyDeltaToDOM is called
  // initialize variables needed
  scheduler.measure(() => {
    // store some values in variables
  });
  scheduler.modify(() => {
     // DOM modifications have access to state and measured variables
  });
  scheduler.postUpdateMeasure(() => {
    // store more values in variables
  });
  scheduler.finalModify(() => {
    // DOM modifications needing access to post update measures
  });
}
```

flackr:
> I think we can figure out what needs post update measures, but it could be difficult to associate the code for those changes with all of the changes that they might depend on.

Dependent things should be scheduled in the order of their dependencies. As long as we don't have something in an earlier phase depending on something in a later phase it should work correctly.

> I'm thinking something more like this:

Yeah having alternate phase names for the second measure/modify cycle makes things simpler since everything that calls it is running in the same cycle (in the order it calls it). It does mean that we can't easily support more phases if we think we need them - note the scoreboard does many measure/modify cycles right now for example. It shouldn't be necessary though.

It sounds like the consensus is that function callbacks on a scheduler object is a reasonable approach so I'll rewrite the patch to use this style.

robartsd:
> > I think we can figure out what needs post update measures, but it could be difficult to associate the code for those changes with all of the changes that they might depend on.
> 
> Dependent things should be scheduled in the order of their dependencies. As long as we don't have something in an earlier phase depending on something in a later phase it should work correctly.

This seems like it would be difficult if the dependency is outside the widget.

> It sounds like the consensus is that function callbacks on a scheduler object is a reasonable approach so I'll rewrite the patch to use this style.

I think a global scheduler object might work well. I looked at the references you linked to. Any reason we shouldn't use [FastDOM](https://github.com/wilsonpage/fastdom)?

flackr:
> I think a global scheduler object might work well. I looked at the references you linked to. Any reason we shouldn't use [FastDOM](https://github.com/wilsonpage/fastdom)?

Well for one, what I suggested above was basically the fastdom api and you wanted to explicitly name the later phases 😉

But also, fastdom schedules the work at the next requestAnimationFrame time which is more likely to be breaking than writing our own version which could ensure that the modifications have completed before returning from applyDeltaToDOM. fastdom also doesn't have a way (AFAICT) to explicitly run a second measure/mutate cycle in the same frame as it schedules the work for the next requestAnimationFrame. This could allow for subtle bugs because further deltas could start to be processed before the previous one is done.

flackr:
That said, advantages of using fastdom would be that it already has a nice strict mode to discover violations so fixing performance issues and ensuring no regressions should be easier, and there's nothing to pass around. Might be worth trying 🤷🏻‍♂️

flackr:
The problem that I'm worried about with naively using fastdom is this. If we start applying a second delta before the mutate phase of the previous one has ran, then the measure callbacks of the second delta will run immediately after the measure callbacks of the first. This could result in using implied state (e.g. computed styles) from before the previous delta.

robartsd:
I think you're right about the problems that FastDOM would have for us. You obviously know more about this stuff than I do.

After thinking about it longer, I think the nested approach to creating multiple measure/mutate cycles probably would work.

Patterns would be:
```js
scheduler.modify(()=>{
  // DOM modifications that depend only on state
  // these will run after measurement promises resolve, but don't need to be scheduled inside one
};
scheduler.measure(()=>{
  // required pre-modification measurements
  schedluler.modify(()=>{
    // DOM modifications that depend on pre-modification measurements
  });
});
scheduler.modify(()=>{
  scheduler.measure(()=>{
    // required post-modification measurements
    scheduler.modify(()=>{
      // DOM modifications that depend on post-modification measurements
    });
  });
});
scheduler.measure(()=>{
  // required pre-modification measurements
  scheduler.modify(()=>{
    scheduler.measure(()=>{
      // required post-modification measurements
      scheduler.modify(()=>{
        // DOM modifications that require both pre- and post-modification measurements
      });
    });
  });
});
```
While theoretically, an unlimited depth of cycles could be used, the awkwardness of the syntax discourages it.

ArnoldSmith86:
Would it prettier to do something like this?

```Javascript
  async applyDeltaToDOM(delta) {
    let text = this.domElement.style.fontSize;

    await allDomMeasurements();

    setText(this.domElement, text);

    await allDomManipulations();

    console.log(this.domElement.innerText);
    text = this.domElement.style.display;

    await allDomMeasurements();

    setText(this.domElement, text);

    await allDomManipulations();

    console.log(this.domElement.innerText);
  }
```

robartsd:
> Would it be prettier to do something like this?
> 

Doesn't matter if it is prettier as it does not schedule the measurements and modifications to be done in a batch.

ArnoldSmith86:
Why not? Should be able to resolve the promises in the same order.

robartsd:
> Why not? Should be able to resolve the promises in the same order (and at the same time).

Because it isn't wrapping the DOM reads and writes into scheduled promises.

> For the second round, it might be less confusing to call a different function.

Yes, we could have functions to schedule directly into the second round as I suggested in my [earlier comment](#issuecomment-1440443592). That pattern might be prettier, because I think we will mostly have modifications based only on state and modifications that use post update measurements (transition support would work best with both); but the nested pattern is more explicit as to what the code is actually doing.

I do really like the idea that this would introduce a way to explicitly get DOM status prior to any changes are applied.

ArnoldSmith86:
> Because it isn't wrapping the DOM reads and writes into scheduled promises.

It's wrapping the reads with the start of the function call and `await allDomMeasurements()`. So you would resolve all the promises waiting for `allDomMeasurements` when all the measurements of all widgets are done. Then all of them would continue to `await allDomManipulations()` which in turn pauses until all are done. I'm pretty sure this could be made work in the same way.

Resolving the promises could be done in the same `setTimeout(..., 0)` construct the current version is using.

Or am I getting something wrong?

robartsd:
> > Because it isn't wrapping the DOM reads and writes into scheduled promises.
> 
> It's wrapping the reads with the start of the function call and `await allDomMeasurements()`. So you would resolve all the promises waiting for `allDomMeasurements` when all the measurements of all widgets are done. Then all of them would continue to `await allDomManipulations()` which in turn pauses until all are done. I'm pretty sure this could be made work in the same way.

Your example reads from the DOM 4 times in statements that are not wrapped in scheduled promises. It also writes o the DOM twice (via a helper function which might be modified wrap the writes into promises).

I think what we want is a way to abstract the actual scheduling of the calls to the DOM away so that we can process the delta for all widgets figuring out what DOM calls are needed, then have them all run in optimized read/write cycles.

flackr:
> Your example reads from the DOM 4 times in statements that are not wrapped in scheduled promises. It also writes o the DOM twice (via a helper function which might be modified wrap the writes into promises).
> 
> I think what we want is a way to abstract the actual scheduling of the calls to the DOM away so that we can process the delta for all widgets figuring out what DOM calls are needed, then have them all run in optimized read/write cycles.

The style @ArnoldSmith86 is actually one of the ones I considered. It can be made to work if `allDomMeasurements()` is a function which returns a promise for the current / next measurement phase opportunity (possibly scheduling another such phase).

The only reason I didn't initially go with it is that it's a little sloppy that despite being an async function the caller can't await on the returned promise without circularity because that promise doesn't resolve until the promises given by the caller are resolved. I do still think it might be the cleanest design though.

ArnoldSmith86:
My first commit fixes the `Deal` button in `The Team II` and probably many other games. Your comment suggests that you didn't do it this way for a reason. Can you explain?

My second commit fixes the add widget overlay. I'm not 100% sure why I need a `setTimeout` here while the fix from the first commit works just fine. If the first commit works, I would assume that the `applyInitialDelta` calls in `editmode.js` return synchronously and finish their stuff before the following code.

96LawDawg:
I tested my new Cards room, which was having some lagging issues that required me to adjust the code.

Tested recalling some cards:
  Before my recent improvements:  Production (161ms), PR-1642 (139 ms)
  After my recent improvements: Production (107ms), PR-1642 (87ms)

Also tested switching from one game to another:
  Before my recent improvements: Production (126ms), PR-1642 (98ms)
  After my recent improvements: Production (546ms), PR-1642 (208ms)

If you look closely, you will see that my recent improvements actually slowed things down quite a bit when switching games.  But that doesn't happen often so I off-loaded some stuff that was happening during recall to then.

But in every instance, this PR was better.

PR #1644, opened 2023-02-21
update fantasy wars second try

Ingegneus:
used remote something something on the first try, now doing it with a local copy of the repo to see if I can change the images


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1644/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1644-pr-test`.

ArnoldSmith86:
See https://github.com/ArnoldSmith86/virtualtabletop/pull/1223#issuecomment-1310934580 and https://github.com/ArnoldSmith86/virtualtabletop/pull/1223#issuecomment-1364109066 for some explanation.

If you changed something about the assets and do what I did in that video, `git status` should tell you that asset files were added or removed and that you should call `git add <path>` to add them. If you use a GUI for git, it'll probably make sure that you decide whether or not you want to add them.

ArnoldSmith86:
No, actually. I'll take a look.

ArnoldSmith86:
There were different game names in the two json files. That (and the fact that all the assets are gone) makes me assume that you're doing something way more complicated than you have to.

You had a version of the improved Fantasy Wars in a room on the production server, right?

1. Download that as a VTT file and add it like any other game to any room on your local server.
2. Make sure you have `"allowPublicLibraryEdits": true` set in `config.json`.
3. Load variant 1 of the game.
4. Select the public library version in the game shelf, select "Edit" from the menu, edit the variants and chose "Update from active game". If you started the server with `npm run debug`, it should restart automatically at this point because the files changed. If you started it differently, I'm not sure if you need to restart it manually at this point.
5. Do the same for the second variant.
6. All changes to the directory `library/games/Fantasy Wars` should be done now. All you have to do is tell `git` to commit them. If you use a GUI, it'll probably tell you what to do. If you use the command line, just do `git add "library/games/Fantasy Wars"` and `git commit -a`.

What did you try to do?

Ingegneus:
ah thanks! I wanted to change a few things across many widgets, so I wanted to edit the entire json. since I was already at it I edited the _meta info, but forgot to edit it in the other version. for test purposes I tried removing all assets to see if that resolved any issues, that's why they were missing.
unfortunately I can't install nodejs atm because it requires a restart of my computer

also how are multiple _meta information across different variants handled? as far as I can tell there is only one displayed in the game overview

96LawDawg:
> also how are multiple _meta information across different variants handled? as far as I can tell there is only one displayed in the game overview

They should all be identical and only one is displayed (I think the one from the 0 variant). The only difference allowed I think is in the variant image.

ArnoldSmith86:
There are a few things that are variant dependent. Basically what you enter for a variant in the GUI.

I'm not sure which one overrides which if they differ. But the name is used as a kind of ID for the public library games, so it put the two variants into two games. That lead to one game only adding variant 1 which lead to variant 0 being set to null. The system didn't enjoy that.

96LawDawg (approved):
Looks good to me.  Did not play test.  Have an approval.  Could I maybe get your approval on 1551?

Ingegneus:
so now I should wait 24h starting from your approval before merging right?

RaphaelAlvez:
> so now I should wait 24h starting from your approval before merging right?

Give me time until today's meeting to review it please. I believe one the only one that knows this game.

RaphaelAlvez (approved):
I did a small play test by myself and reviewed your code. looks good

PR #1646, opened 2023-02-21
UI Explanation Update for In-Progress Games

96LawDawg:
This adds additional explanatory text to the UI when a user clicks on Save Active Game. Recently at least one new user has been confused by the use of that button and attempted to use it for game development, which causes some problems because of the additional metadata added. Ideally, we could figure out some better way to do this on the graphics and buttons before the player reaches this point, but I after thinking about it, I don't have any great ideas so this can be an interim solution.

The one idea I did have is an on-hover pop up over Save Active Game which would basically say the same thing as this modified text. We don't need tooltips on every button, but maybe just that one.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1646/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1646-pr-test`.

client/room.html
```diff
@@ -320,7 +320,7 @@ <h1>Save game</h1>
       </div>
       <div>
         <div icon="save">
-          <p>This creates a save state of the currently active game so that you can start a different game and come back to this one later.</p>
+          <p>This creates a save state of the currently active game so that you can start a different game and come back to this one later. Only use this for playing games. For creating or editing, use the "+Add Game" button or sidebar hamburger icon "Edit" button.</p>
```
robartsd, line 323: I'd eliminate "Only use this for playing games." I don't think these words really add meaning to the overall message.

RaphaelAlvez:
I do not see any pop-up. I'm on chrome

96LawDawg:
Click on Save Active Game.

Save Active Game

RaphaelAlvez (approved):
looks better. probably will not solve the problem but it is an improvement

PR #1649, opened 2023-02-22
fixed save states without version

### LINKED ISSUE #1635: I got an internal server error when downloading an in-progress game
Doesn't seem to happen with every game. And I can't share my private room.

ArnoldSmith86:
This fixes #1635. When updating the currently loaded save game, the button deleted `_meta` from the saved file which also removes version information. This PR fixes that.

To mitigate broken states now saved on the server, this PR also looks at all save games the next time when a room is loaded. If it finds states without version information, it looks at the date when the state was saved, compares it to the most recent version changes and sets the version in the file accordingly, fixing it for the future.

If a file was broken, loaded before a version bump and then resaved afterwards, it might still be broken, because it might have skipped one or more file updaters.

Also: this assumes that the server updated to new commits as soon as they went live. This should be true for the production server but other servers probably only update on demand. @mousewax, this could affect your server, for example.

### DIFF
--- a/server/room.mjs
+++ b/server/room.mjs
@@ -741,7 +741,7 @@ export default class Room {
     if(updateCurrentSave) {
       const stateID = this.state._meta.activeState.saveStateID;
       const newContent = {...this.state};
-      delete newContent._meta;
+      newContent._meta = { version: this.state._meta.version };
       this.state._meta.states[stateID].saveDate = +new Date();
       fs.writeFileSync(this.variantFilename(stateID, 0), JSON.stringify(newContent));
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1649/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1649-pr-test`.

96LawDawg (approved):
Blind approval

robartsd (approved):

PR #1654, opened 2023-02-24
new holder property `showInactiveFaceToSeat`

ArnoldSmith86:
If you put cards into a `showInactiveFaceToSeat: <seatID>` holder, the players occupying the given seats will see the other face of the cards.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1654/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1654-pr-test`.

RaphaelAlvez:
showCardsBacks should be enough...

RaphaelAlvez:
> showCardsBacks should be enough...

Or are you planning a showCardsBacksToOnwer?

robartsd:
Should probably use a holder that is owned by a particular seat, rather than a multiplayer childrenPerOwner holder. Could be nice to specify which face the other players should see.

RaphaelAlvez:
I'm not understanding how is Hanabi working. I see cards showing the back to their owner but I cannot see how you modified that behavior for this game

ArnoldSmith86:
> I'm not understanding how is Hanabi working. I see cards showing the back to their owner but I cannot see how you modified that behavior for this game

I'm not sure what you mean. I added `"showCardBacksToOtherPlayers": true` to the player holders.

client/css/widgets/holder.css
```diff
@@ -20,6 +20,18 @@
...
 
+.holder.showCardBacksToOtherPlayers > .foreign.card.widget {
+  display: block !important;
+}
+
+.holder.showCardBacksToOtherPlayers > .foreign.card.widget > .active.cardFace {
+  display: none;
+}
+
+.holder.showCardBacksToOtherPlayers > .foreign.card.widget > :not(.active).cardFace {
```
flackr, line 31: This is assuming that there is only one other face though right? I've seen a few games where widgets have more than 2 faces where I think this would display all of them.

client/css/widgets/holder.css
```diff
@@ -20,6 +20,18 @@
   top: 0;
 }
 
+.holder.showCardBacksToOtherPlayers > .foreign.card.widget {
+  display: block !important;
```
flackr, line 24: Which rule is this trying to override, can you do it by ensuring the correct order or increasing specificity of this or decreasing speficity of the other instead of using `!important`? The more we use `!important` the harder it is for game authors to override, having to also specify `!important` like with transform on my perspective table.

RaphaelAlvez:
> > I'm not understanding how is Hanabi working. I see cards showing the back to their owner but I cannot see how you modified that behavior for this game
> 
> I'm not sure what you mean. I added `"showCardBacksToOtherPlayers": true` to the player holders.

Did you invert the back and front faces in that deck?

ArnoldSmith86:
> Did you invert the back and front faces in that deck?

No, it just shows face 0 to the owner.

client/css/widgets/holder.css
ArnoldSmith86: Yes but I don't see how anything in real life with more than 2 faces lets one player see one face and other players 1 specific other face.

client/css/widgets/holder.css
ArnoldSmith86: Should be this which already uses `!important`:

```CSS
body:not(.jsonEdit) .widget.foreign {
  display: none !important;
}
```

flackr: Which was presumably used because it wasn't specific enough to override a `display: block` rule but that's going down a rabbit hole. Anyways we could always remove `!important` later.

client/css/widgets/holder.css
flackr: I mean it doesn't have to be that the extra face has a real life representation. Some games use faces to apply a rotation, which could be used in hanabi for the conventional way that people remember significant cards. But we can always go back and improve that if the need arises.

client/css/widgets/holder.css
ArnoldSmith86: Shouldn't be _that_ hard to remove those two.

flackr: Looks like `@layer` has pretty good support these days: https://developer.mozilla.org/en-US/docs/Web/CSS/@layer . We should switch to that so that author styles are used above vtt styles regardless of specificity. (Not suggesting in this patch)

ArnoldSmith86:
> Should probably use a holder that is owned by a particular seat, rather than a multiplayer childrenPerOwner holder. Could be nice to specify which face the other players should see.

I agree that pairing this with `childrenPerOwner` is more trouble than it is worth because much of the codebase assumes that you can't see the cards of the other players. If a non-owner drags the card, everything breaks.

I would propose changing this to a property that takes one or more seat `id`s and their players can see the `activeFace` and the others see the other face. I have no clue about a good name though.

Thoughts?

RaphaelAlvez:
> > Should probably use a holder that is owned by a particular seat, rather than a multiplayer childrenPerOwner holder. Could be nice to specify which face the other players should see.
> 
> I agree that pairing this with `childrenPerOwner` is more trouble than it is worth because much of the codebase assumes that you can't see the cards of the other players. If a non-owner drags the card, everything breaks.
> 
> I would propose changing this to a property that takes one or more seat `id`s and their players can see the `activeFace` and the others see the other face. I have no clue about a good name though.
> 
> Thoughts?

handOf, handOwner? 

Then maybe another property that sets that only ones can interact with cards? Maybe onwerInteractionOnly?

I feel this PR is becoming a PR for a new kind of hand.

RaphaelAlvez:
> > Did you invert the back and front faces in that deck?
> 
> No, it just shows face 0 to the owner.

I still cannot understand. This feature should show the front to the owner and the back to the other players. 
In that game the effect is inverted: back to owner and front to others.

You either changed the css behavior somewhere or you changed which face is which on your deck.

Otherwise I cannot understand the difference in behavior

flackr:
> I still cannot understand. This feature should show the front to the owner and the back to the other players. In that game the effect is inverted: back to owner and front to others.

The feature shows the active face to the owner and the non-active face to other players. In fireworks the active face in your hand is face 0 due to the `onEnter` property and other players see the non-active face (face 1):
```
  "onEnter": {
    "clickable": false,
    "activeFace": 0
  },
```

client/css/widgets/holder.css
robartsd: Could add an explicit foreignFace class to the face that other players should see.

ArnoldSmith86:
How about `showInactiveFaceToSeat`? That reverses the logic but `showActiveFaceToSeat` doesn't really get the function across.

RaphaelAlvez:
> How about `showInactiveFaceToSeat`? That reverses the logic but `showActiveFaceToSeat` doesn't really get the function across.

This is specially nice because it allows inactive users to whatcha the game.

You can either set the seat to be th owner's seat or the the other seats. Seems flexible

client/css/widgets/holder.css
ArnoldSmith86: I did a quick test and only the highest card face will be visible because they are displayed on top of each other. I think this is good enough since probably nobody will ever use this with more than 2 faces.

client/css/widgets/holder.css
ArnoldSmith86: The `!important` is no longer needed because it no longer interacts with `.foreign`.

ArnoldSmith86:
I updated Fireworks to use the new functionality.

client/css/widgets/holder.css
96LawDawg: > I did a quick test and only the highest card face will be visible because they are displayed on top of each other. I think this is good enough since probably nobody will ever use this with more than 2 faces.

I did a test on a deck with 4 faces.  For the other players that can see everything, clicking through the faces shows the faces in order as you would expect.  For the player with the `showInactiveFaceToSeat` property, the 4th face was shown three times and the third face was shown once as you click through.

This is not something we need to fix.  Just need to cover it in the wiki that this is only intended for cards with 2 faces.

96LawDawg (approved):
Approved through test.  I also tested Fireworks and the changes look good.

This does not work at all with `childrenPerOwner` hands.  Again, that is fine.  It would be really strange and difficult to work with in hands.  We'll just document that in the wiki.

96LawDawg:
Wiki entry (under Holders on Widgets page):

| showInactiveFaceToSeat |null  |  When set to a seat `id` or array of `id`s, the inactive face of a card will be displayed to the specified seats. This only works with cards, not basic widget faces. This will produce strange results if the card has more than 2 faces. This does not work as described on holders with `childrenPerOwner` set to true since the other players will not see any visual representation of cards not owned by them. |

96LawDawg:
One final comment I noted in testing.  This only seems to work for cards and not basic widgets with faces.  Again, that's fine but did I miss something?  Or is this something anybody cares about?

mousewax:
> One final comment I noted in testing. This only seems to work for cards and not basic widgets with faces. Again, that's fine but did I miss something? Or is this something anybody cares about?

This is an important feature for games like Code777 and Pyramid Poker where both players are looking at opposite sides of tiles or cards.

ArnoldSmith86:
> > One final comment I noted in testing. This only seems to work for cards and not basic widgets with faces. Again, that's fine but did I miss something? Or is this something anybody cares about?
> 
> This is an important feature for games like Code777 and Pyramid Poker where both players are looking at opposite sides of tiles or cards.

But wouldn't that be fine with two-sided cards? Or do you need basic widgets for that?

ArnoldSmith86:
I just thought about adding it to basic widgets but they work fundamentally different and it wouldn't just be a simple CSS change. We could possibly look at that with a later PR.

PR #1656, opened 2023-02-25
Don't target foreign holders even if they're close - fixes #1651

### LINKED ISSUE #1651: JSON Editor and onlyVisibleForSeat
In https://virtualtabletop.io/mqf2, there is a holder under the first seat the `onlyVisibleForSeat`.  You don't need to sit in either seat to see the bug.  In normal game mode drag a card over the older (that you can't see) and try to drop it.  You can't.  In JSON Editor mode, drag the card over the holder and it disappear and reappears.  And you can drop it in the holder.  The JSON Editor did not used to behave that way.

There also seems to be a problem when using other aspects of seats like `hideWhenUnused`, but I haven't tested that yet.

flackr:
When we added hit detection for foreign elements in the JSON editor, we made those elements no longer `display: none`. While they do have `pointer-events: none` meaning that they are not returned by `elementsFromPoint` the old hoverTarget targeting code looks for elements which are nearby and since these foreign holders now have detectable geometry they pass the overlap test which tries to find nearby holders if there was none found under the centerpoint of the widget:
https://github.com/ArnoldSmith86/virtualtabletop/blob/main/client/js/widgets/widget.js#L1941

As such, we need to explicitly exclude any widgets inside of a foreign widget from being a valid drop target. This patch does this by setting a custom property `--foreign:true;` which is inherited to all descendants and can be easily filtered out without needing to walk the ancestor elements.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1656/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1656-pr-test`.

96LawDawg (approved):
Approved through test.  I tested it using my #1651 demo room (onlyVisibileForSeat) and on @mousewax's game (hideWhenUnused) that triggered the discussion. I don't know if this would effect anything else; doesn't seem like it, so i didn't test anything else. Recommend approval ASAP.

client/js/main.js
```diff
@@ -58,6 +58,9 @@ function getValidDropTargets(widget) {
         break;
     }
 
+    if (getComputedStyle(t.domElement).getPropertyValue('--foreign') == 'true')
```
ArnoldSmith86, line 61: Is this check faster than refactoring this into a function and call it (and only `get('owner')` once):

```Javascript
    if(Array.isArray(this.get('owner')) && this.get('owner').indexOf(playerName) == -1)
      className += ' foreign';
    if(typeof this.get('owner') == 'string' && this.get('owner') != playerName)
      className += ' foreign';
```

flackr: I'm using the css inheritance to automatically apply this to descendants within foreign holders.

ArnoldSmith86 (approved):
Code looks fine to me.

PR #1657, opened 2023-02-26
Updates to Material Icons and Roboto fonts

mousewax:
Added custom CSS for private servers. (`client/css/custom.css`)

Added `custom.*` to the `.gitignore` file in the event more custom parts are added.

Updated Roboto to use the standard TTF format. Nothing otherwise has changed with that one.

Updated material icons as a significant number of symbols have been added. Four variants have also been added.
The five classes are:
```
.material-icons
.material-icons-outlined
.material-icons-round
.material-icons-sharp
.material-icons-twotone
```
As pictured here:
image


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1657/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1657-pr-test`.

96LawDawg:
Will this do anything to the appearance of existing games that use the material-icons class?  I mean, if Google changed the font very slightly, that is probably no big deal.  But if there are any fonts missing or renamed, that would be a problem.

mousewax:
No. Only additions were made. There are likely VERY small changes to existing icons.

RaphaelAlvez:
> No. Only additions were made. There are likely VERY small changes to existing icons.

This is good. have you compared it in your games that use it?

mousewax:
> This is good. have you compared it in your games that use it?

Yes. They all appear normal.

RaphaelAlvez (approved):
> > This is good. have you compared it in your games that use it?
> 
> Yes. They all appear normal.

"Blind" approval. I trust that if it did not break your games it should be good for other games.

Wait for a second approval before mergin

96LawDawg (approved):

I also threw in a small change to fix the cube svg.  I'll document the change to the font page on the wiki.  However, we should also document the custom css change somewhere. @mousewax, can you give an example/better explain how that is going to be used?

mousewax:
> I also threw in a small change to fix the cube svg. I'll document the change to the font page on the wiki. However, we should also document the custom css change somewhere. @mousewax, can you give an example/better explain how that is going to be used?

I really wanted to use some of the fonts we use in the actual games and I haven't been able to do that. This will allow me to specify my own font faces as well as classes to use. It would also allow me to make changes to how some elements appear. Right now I have a script that forces my own CSS into the main room.html and that's not ideal. I want to make it so that I can keep updating from github without weird forced modifications.

96LawDawg:
Sounds good.  I'll try to explain that in someway, somewhere in the wiki and say that is mainly intended for a specific use case not on VTT.

By the way, you can use non-VTT fonts.  There are instructions at the bottom of the wiki font page.  It's a little more cumbersome than just declaring a class, but easier (to me) than creating a css config file and coding it into VTT.  https://github.com/ArnoldSmith86/virtualtabletop/wiki/Fonts#font-face

mousewax:
I added an custom assets directory. Anything in there should be referenced with `i/custom/yournew.file`

mousewax:
> Sounds good. I'll try to explain that in someway, somewhere in the wiki and say that is mainly intended for a specific use case not on VTT.
> 
> By the way, you can use non-VTT fonts. There are instructions at the bottom of the wiki font page. It's a little more cumbersome than just declaring a class, but easier (to me) than creating a css config file and coding it into VTT. https://github.com/ArnoldSmith86/virtualtabletop/wiki/Fonts#font-face

Yeah. I know that was there but it's far easier (for me) to add new classes and work this way. One of the weaknesses of the VTT platform is not being able to just write CSS using my code editor that has all sorts of functions to make things quick and easy. That way also feels convoluted and this provides a greater amount of flexibility even outside of the rooms. For example, one of the CSS things I inject hides the public library if the URL doesn't contain a certain word. And I can override or change some of the color variables to brand the site to make it look a bit different.

ArnoldSmith86:
If we have more icons now, we should update https://github.com/ArnoldSmith86/virtualtabletop/blob/main/assets/fonts/symbols.json .

mousewax:
> If we have more icons now, we should update https://github.com/ArnoldSmith86/virtualtabletop/blob/main/assets/fonts/symbols.json .

I did not add any new icons to the VTT Icons set.,

ArnoldSmith86:
> > If we have more icons now, we should update https://github.com/ArnoldSmith86/virtualtabletop/blob/main/assets/fonts/symbols.json .
> 
> I did not add any new icons to the VTT Icons set.,

But to the material icons, right? The icon picker shows all of those and reads them from that JSON file.

I'll update it when I do something cool with it in the editor, I guess.

mousewax:
Oh! I didn't realize you did it with them as well! Yes. There have been numerous additions. Lemme look into that.

ArnoldSmith86:
Note that I generated that file using https://github.com/ArnoldSmith86/virtualtabletop-misc/blob/main/helper-scripts/generate-icon-list.sh (except for the stuff about your custom icons).

So before you spend a lot of time on it, I'll try to adjust my stuff.

96LawDawg:
> The icon picker shows all of those and reads them from that JSON file.

The icon picker?  What is that?  We have one of those?

ArnoldSmith86:
So far only for inserting symbols into richtext (how to play, attribution). The smiley in the toolbar.

PR #1658, opened 2023-02-27
Give every routine function the same names for the widgets it works on

### LINKED ISSUE #60: Add collection parameter to functions
The functions FLIP, LABEL, MOVE, MOVEXY, ROTATE, SHUFFLE and SORT currently only support a holder parameter. As an outcome of the "buttons overhaul" discussion, they should additionally support the collection parameter to the effect that they can process the widgets of the collection _directly_ (rather than their children).

### LINKED ISSUE #64: MOVEXY should be able to move widgets on the floor
Currently, MOVEXY can only be used with holders, moving the widgets out of the holder and onto the floor. Once MOVEXY also supports collections, it should also be capable to move widgets around on the floor, both to absolute and relative ("shifting") coordinates.

Therefore it should have a mode parameter with "set" for absolute and "change" for relative positioning (with "set" being the default and working as of now).

Example for how widgets could be moved relatively 10px to the right, 100px up and 2 lower in z-order:
```
{
  "func": "MOVEXY",
  "collection": "myCollectionName",
  "x": 10,
  "y": -100,
  "z": -2,
  "mode": "change"
}
```

When x, y (or z) are omitted in "change" mode, those coordinates should be left as is.

### LINKED ISSUE #275: Improve parallels between Holders and Collections
`FLIP` allows the use of a holder rather than a collection as its input, with precisely the same results. While `MOVE`, `MOVEXY`, `ROTATE`, and `SORT` do not, the Wiki clearly indicates that someone thought it would be a good idea if they did. It seems to me that in a lot of cases, holders and collections should both be accepted by these functions; from the programmer's point of view they may be similar, and not infrequently a `SELECT` is used to select all items from a holder into a collection. 

We should change the following functions to accept either:
`CLICK`, `CLONE`, `COUNT`, `DELETE`, `GET`, `LABEL` (right now, `LABEL` accepts neither; it operates only on a single label), `MOVE` (`source` parameter only), `MOVEXY`, `ROTATE`, `SELECT`, `SET`.

### LINKED ISSUE #1067: SORT and SHUFFLE also support seats on PCIO
I'm not sure if I want to add that now or rather focus on rewriting routine operations so all of them support collections, IDs, holder children and seat hand children while also making the code cleaner.

RaphaelAlvez:

- [ ] **Decide whether to add a FileUpdater migration** — No migration to the new parameter spellings was added on purpose; adding one also means re-recording the TestCafe hashes in all four `publiclibrary-*.js` files. ([discussion](https://github.com/ArnoldSmith86/virtualtabletop/pull/1658#issuecomment-5239168200))

Every routine function that works on widgets used to spell the widgets it works on differently: `collection` in FLIP, `canvas` in CANVAS, `label` in LABEL, `timer` in TIMER, `widget` in CALL, `from`/`to` in MOVE, `source` in CLONE — and `collection` meant an input in some functions and an output in others. This PR gives all of them the same two names, so a routine no longer has to care which function produced the widgets it wants to act on.

| parameter | meaning |
| --- | --- |
| `target` | the widgets the function works on directly |
| `holder` | widgets whose **content** the function works on — for a seat, what lies in its hand and belongs to its player. Wins over `target` when both are set. |
| `source` | a collection that is itself the subject (SELECT, SWAPHANDS, and now SORT/SHUFFLE, which reorder the collection rather than changing any widget) |
| `collection` | output only |

`target`, `holder`, `fromHolder`, `toHolder` and the SORT/SHUFFLE `source` all accept **either** a list of widget ids **or** the name of a collection, and a list of ids always names those widgets. What a plain **string** means where a collection *and* a widget carry the same name is decided by what the parameter accepted before: `holder`/`fromHolder`/`toHolder` (and the old `from`/`to`/`canvas`/`label`/`timer`/`widget`) name the widget, because that is all they ever named, while `target` and `source` name the collection and read a widget id only when no collection carries the name. Where both exist under one name, the routine reports which one it used. The old spellings of a collection (`collection`, `excludeCollection`, and `source` in CLONE/SELECT/SWAPHANDS) still name a collection and nothing else, so a name nobody collected stays the error it always was.

### Per function

| function | now | replaces |
| --- | --- | --- |
| CALL | `target`, `holder` | `widget` (and it now calls the routine on every widget, not just the first — no file in the library passes more than one, so nothing there changes) |
| CANVAS | `target`, `holder` | `canvas`, `collection` |
| CLICK | `target`, `holder` | `collection` |
| CLONE | `target`, `holder` | `source` (its output stays `collection`) |
| COUNT | `target`, `holder` | `collection` |
| DELETE | `target`, `holder` | `collection` |
| FLIP | `target`, `holder` | `collection` |
| GET | `target`, `holder` | `collection` |
| LABEL | `target`, `holder` | `label`, `collection` |
| MOVE | `target`, `fromHolder`, `toHolder` | `collection`, `from`, `to` |
| MOVEXY | `target`, `fromHolder` | `from` |
| RECALL | `holder`, `target` | — `target` is new: it recalls a deck's cards into whatever the deck sits in (a deck lying on no widget is reported, since a deck is not a holder) |
| ROTATE | `target`, `holder` | `collection` |
| SET | `target`, `holder` | `collection` |
| SHUFFLE | `source`, `holder` | `collection` |
| SORT | `source`, `holder` | `collection` |
| TIMER | `target`, `holder` | `timer`, `collection` |
| TURN | `target` (`'all'` or seats) | `source` |

SELECT and SWAPHANDS already followed the convention and are unchanged. FOREACH is left out on purpose — its `in`/`range`/`collection` inputs need a design of their own (it is the one "wip" item on the project board).

Beyond the renames, every one of these functions now also takes a **collection** where it used to take widget ids only, and `holder` works on any widget type (its children), on several holders at once, and on seats. A `count` limit is still spent per holder, the way FLIP and ROTATE always did it.

### Backwards compatibility

The old names keep working: an operation fills the current name in from the first old name that is set, so **no existing game changes behavior** and there is no state-hash churn. `_meta.version` is untouched.

**Open question for the maintainers:** the original plan also called for a FileUpdater migration that rewrites existing games to the new spelling (and turns single ids into `["id"]`). I left it out deliberately, because it would change the state of every library game that uses these parameters — which means re-recording the hashes in all four `tests/testcafe/publiclibrary-*.js` files — and it would make the prose of the ~20 `Functions - *` tutorials describe names their own JSON no longer uses. Happy to add it as a follow-up commit here if you want it in this PR; it seemed like the kind of thing that should be a deliberate decision rather than a side effect.

### Editing UI

- `client/js/jsonedit.js`: the insert buttons write the new names, and a collection name is only offered where the engine accepts one (`'all'` only for SELECT `source` and TURN `target`).
- `client/js/editor/controls/routine.js`: the visual routine editor offers the new parameters, including a holder variant for the functions that gained one, and still opens games written with the old names. Every operation that can work on the content of a holder offers it under the same phrase (*what is in a holder*), and the sentence says the two things the engine decides on its own: a `count` next to several holders reads *"1 widget in each of holderA and holderB"*, because that is what it does, and a parameter the engine skips — a `target` set next to a `holder`, a `count` next to a `fillTo` — stays in the sentence struck through, with a "!" naming the reason, instead of disappearing out of it. A string that names a widget of the room is worded as that widget rather than as a group nobody created.
- `validator/validate_gamefile.js`: new `collectionOrIDs` shape for all of them, so both spellings validate. Checked against all 573 library game/tutorial files: no file gains a problem; a handful of old false positives disappear and some aggregated messages are now reported per array element.

### Test room

The test room is a **tutorial**: `library/tutorials/Functions - Widget Inputs/`, with three variants — `0.json` *"target and holder"* (ROTATE driven by a list of ids, by a holder, by a collection, and holder-beats-target), `1.json` *"MOVE"* (`fromHolder` → `toHolder`, `target` → `toHolder`, and back) and `2.json` *"seats and counting"* (a `count` spent per holder against the same widgets named as one collection, and `holder` on a seat — including the MOVE into a seat that gives the widgets to its player in the first place). `_meta.info` uses `Functions - AUDIO` as its template and carries exactly its fields, with the same value shapes (`mode: "Tutorial"`, `players: "1"`, `time: "0"`, `language: "en-US"`, `showName: false`, `usesAIImagery: false`, only `variant` differing between the files); each variant names itself in its title, the way `Functions - INPUT` does. The library image is a 530×530 square SVG in the same blue-badge style as its neighbours, in the tutorial's own `assets/` directory under the id `_meta.info.image` names. All three variants validate clean and every button in them was clicked in a browser.

- the third variant: https://agent.virtualtabletop.io/reports/pr/1536305652446535690/tutorial-seats-and-counting.png
- after pressing every one of its buttons: https://agent.virtualtabletop.io/reports/pr/1536305652446535690/tutorial-seats-and-counting-in-use.png

Automated coverage: `tests/client/engine/widget-parameters.test.js` pins both the new and the old spelling of every input form, the multi-destination dealing of `count`/`fillTo`, and that a list of widget ids leaves no collection behind.

### Also fixed

- `getCollection()` referenced an undefined `batchDepth`, so passing a list of widget ids to a collection parameter threw a `ReferenceError` instead of building the collection. The collection such a list gets is now dropped again once the operation that asked for it is done, so a `CALL` no longer deep-copies a growing pile of them.
- MOVE spends the `count`/`fillTo` across several destinations one after the other again — `MOVE target count:1` into two holders deals one widget to each rather than putting the same one into both.
- A MOVE with a leftover empty `from` (what the routine editor writes for a widget chip that was added but never filled in) reads its widgets from `target`, the way it always did — MOVE only ever looked at `from` while it named something.
- CANVAS, LABEL and TIMER no longer report an empty holder as "Collection DEFAULT is empty", naming a parameter they never read. An empty holder is an expected state, as FLIP and ROTATE already treat it.

One deliberate behavior change: MOVEXY that names neither `from`/`fromHolder` nor `target` used to report a missing widget id and do nothing; it now moves the `DEFAULT` collection like every other function. No library game is affected (the one bare `{"func":"MOVEXY"}` in `library/games/Card Games` runs with no DEFAULT collection in scope), but say the word and it can keep erroring instead.

Closes the parameter-rename items of the routine-input overhaul (project board: CANVAS, CLICK, CLONE, COUNT, DELETE, FLIP, MOVE, CALL).


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1658/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1658-pr-test`.

RaphaelAlvez:
I uploaded a showcase in the test room

robartsd:
I think that `holder` parameter should accept widgets of any type. If the type of the widget is not "holder" then all the children of the widget are the widgets to be acted upon (other widget types don't have a dropTarget property to filter with).

Should `to` of `MOVE` be renamed `target`?

Possibly a separate PR, but I think that `MOVE` should be given `x` and `y` parameters for use if the target of the move is not an `alignChildren` holder. `MOVEXY` could be converted to `MOVE` with a target of `null`.

RaphaelAlvez:
> I think that `holder` parameter should accept widgets of any type. If the `type` of the widget is not "holder" then all the children of the widget are the widgets to be acted upon (other widget types don't have a dropTarget property to filter with).

I made the documentation before started working on this. After your comment on discord I changed to be as you described with the exception that I changed the .children() code to be .childrenTarget() and it has a different behavior for seats. Seats push the hand children with the correct owner.

> Should `to` of "MOVE" be renamed `target`?

I know the word target is normally related to a destination butod rather keep it for the widgets that will be directy be moved. This is a feature we don't have are the moment.
For the destination I'm thinking on using `destination`, `targetHolder` or `toHolder`
I'm really not set in any naming convention so 

I'm open to change to better words based on feedback

> Possibly a separate PR, but I think that "MOVE" should be given `x` and `y` parameters for use if the target of the move is not an `alignChildren` holder. "MOVEXY" could be converted to "MOVE" with a target of `null`.

Probably another PR. My focus is to make sure I push this with 0 changes to current behavior. Changing more stuff will make it harder to be sure I did a good job.

I might make parallel PRs since I'll be more familiar with the code

robartsd:
> > I think that `holder` parameter should accept widgets of any type. If the `type` of the widget is not "holder" then all the children of the widget are the widgets to be acted upon (other widget types don't have a dropTarget property to filter with).
> 
> I made the documentation before started working on this. After your comment on discord I changed to be as you described with the exception that I changed the .children() code to be .childrenTarget() and it has a different behavior for seats. Seats push the hand children with the correct owner.

So when `holder` is specified, the collection of widget acted upon are child widgets based on the widgets in the collection passed to `holder`. Any widget type works; for most widgets types it is a simple selection of child widget; for holders the child widgets are limited to those that match dropTarget; seats redirect to the widget specified has `hand` in the seat and add a further restriction on owner matching the `player` in the seat. This sounds good to me.

What happens if a seat in the collection passed to `holder` has no valid `hand`?

RaphaelAlvez:
> I think that `holder` parameter should accept widgets of any type. If the `type` of the widget is not "holder" then all the children of the widget are the widgets to be acted upon (other widget types don't have a dropTarget property to filter with).

I checked and you can currently put any ID in holder and it will apply to all children. Same behavior in this PR except for seats

> What happens if a seat in the collection passed to `holder` has no valid `hand`?

fixed all possible exceptions thanks for pointing it out.

server/fileupdater.mjs
```diff
@@ -1,4 +1,4 @@
-export const VERSION = 12;
+export const VERSION = 50;
```
flackr, line 1: Version 50 already? Avoiding conflicts? I assume we'll change this before merging.

client/js/widgets/seat.js
```diff
@@ -42,6 +42,15 @@ class Seat extends Widget {
     this.updateLinkedWidgets(true);
   }
 
+  childrenTarget() {
```
flackr, line 45: Any reason not to override children() like holder.js does?

client/js/widgets/seat.js
```diff
@@ -42,6 +42,15 @@ class Seat extends Widget {
...
+    if (widgets.get(this.get("hand"))) {
+      return widgets.get(this.get("hand")).children().filter(c=>!c.get('owner') || c.get('owner')==this.get("player"));
+    } else {
+      return {}
```
flackr, line 49: nit: I think this should return an array - that's the type we get from children() and would get from the above branch.

client/js/widgets/widget.js
```diff
@@ -1370,27 +1374,35 @@ export class Widget extends StateManaged {
...
+          if(collection = getCollection(a.holder)) {
+            for(const w of collections[collection]){
+              if (w.childrenTarget().length){
+                for(const c of w.childrenTarget().slice(0, a.count || 999999))
+                  await c.rotate(a.angle, a.mode);
+              } else {
+                if (w.get("type")=="seat") {
+                  problems.push(`the hand of ${w.get("id")} is empty or does not exist`);
+                } else{
+                  problems.push(`${w.get("id")} is empty or does not exist`);
```
flackr, line 1390: I don't think you can get here if `w` didn't exist as it would crash trying to call methods on `w`. It looks like `getCollection` would filter out missing widgets here: https://github.com/ArnoldSmith86/virtualtabletop/blob/6f396ceddd95a306d63edab5fe428e35b4fa035f/client/js/widgets/widget.js#L697

client/js/widgets/widget.js
```diff
@@ -1370,27 +1374,35 @@ export class Widget extends StateManaged {
...
-            await w(a.holder, async holder=>{
-              for(const c of holder.children().slice(0, a.count || 999999))
-                await c.rotate(a.angle, a.mode);
-            });
+          if(collection = getCollection(a.holder)) {
+            for(const w of collections[collection]){
+              if (w.childrenTarget().length){
+                for(const c of w.childrenTarget().slice(0, a.count || 999999))
+                  await c.rotate(a.angle, a.mode);
+              } else {
```
flackr, line 1386: I'm not sure if it should be an error if there is nothing to ROTATE in the specified holders, just like it's not an error if there is nothing to MOVE. It seems like this would generate errors for expected states sometimes.

server/fileupdater.mjs
RaphaelAlvez: yes. I expect this PR will take some time and other versions will come before it. it ill be changed.

client/js/widgets/seat.js
RaphaelAlvez: children() is actually used in some other places and I'm afraid it might break something if I change the seats' children().
I might be wrong; this can be an unnecessary extra function, but I feel safer this way.

This is one of the things I'm most unsure about this PR but at the same time it doesn't feel too problematic

client/js/widgets/widget.js
RaphaelAlvez: I added this because we get a similar error for collections. You are correct that it might generate a bunch of "empty holder" errors that have close to no value.

I'm open to remove it

client/js/widgets/widget.js
RaphaelAlvez: You're correct this does not give the intended problem for invalid id of seats and holders. I tested only for invalid hands and empty holders. 

it does not freeze the client if the ID is invalid but the debug doesn't know that there was a problem

client/js/widgets/seat.js
RaphaelAlvez: Done

client/js/widgets/holder.js
```diff
@@ -29,7 +29,7 @@ class Holder extends Widget {
 
   children() {
     let children = this.childrenFilter(super.children(), true);
-    if(children.length == 1 && children[0].get('type') == 'pile')
+    if(children.length == 1 && children[0].get('type') == 'pile') //this needs to be improoved so that it can support holders with multiple piles. (allignChildren = false)
       children = this.childrenFilter(children[0].children(), false);
     return children;
   }
```
robartsd, line 35: If we want to resolve all first level piles (player's cannot put a pile into a pile directly, but routines might be able to - the only engine code that I am aware of that resolves nested piles is for getting `_ancestor`), not just sole piles, we could use:

```js
children() {
  const children = super.children().flatMap(child=>child.get('type')=='pile'?child.children():child);
  return this.childrenFilter(children, false)
}
```
Currently piles are not resolved if they have siblings, so this might be considered a breaking change.

RaphaelAlvez: `this.childrenFilter(children, false)`

This was something I did couldn't figure out but left it because it would not cause any harm and maybe there was some edge case I was missing

robartsd: Yeah, I'm not sure if the accept piles argument of `Holder.childrenFilter()` is the best approach, but we certainly need the call to `compareDropTarget()` that is implemented within `Holder.childrenFilter()`.

I actually think it might be nice to implement pile resolution as an option in `Widget.children()`.

VTT-AI-Agent:
I have taken this PR over and rebuilt it on current `main`. The branch was ~880 commits behind and conflicting, and the in-progress ROTATE/MOVE work no longer applied, so the merge commit resets the tree to `main` and everything after it is a fresh implementation. The design is the one this PR and the project board describe; the PR body now lists what actually landed, per function.

### Answers to the review so far

> I think that `holder` parameter should accept widgets of any type. If the `type` of the widget is not "holder" then all the children of the widget are the widgets to be acted upon (other widget types don't have a dropTarget property to filter with).

That is what it does. `holder` resolves through `widget.children()`, which is already per type: a holder filters by its `dropTarget` (and unwraps a single pile), a seat returns what lies in its hand and belongs to its player, and anything else returns its children. So no new per-type code was needed, and a seat with no hand or no player simply contributes nothing.

> Should `to` of "MOVE" be renamed `target`?

It is `toHolder` now, with `fromHolder` on the other side and `target` for the widgets themselves — the naming the project board settled on. `target` had to stay free for "the widgets being moved", which is the case MOVE could not express before.

> Possibly a separate PR, but I think that "MOVE" should be given `x` and `y` parameters […] "MOVEXY" could be converted to "MOVE" with a target of `null`.

Left for a separate PR, as agreed in the thread. MOVEXY here only gets the same names as MOVE (`fromHolder`, plus a new `target`), nothing about what it does changes.

### Two decisions I had to make that are not spelled out on the board

1. **MOVEXY** follows MOVE rather than the general rule: `fromHolder` (not `holder`), so the two functions that move widgets read alike.
2. **LABEL's `label` and TIMER's `timer`** are folded into `target` (they were widget-id parameters, so they are exactly what `target` means), and **SORT/SHUFFLE** use `source`, because they reorder the collection itself rather than changing any widget — that is what the PR description asks for.

Both are easy to change if you would rather have it another way.

### FileUpdater — please decide

The PR description said the file updater still had to be done. I deliberately did **not** add a migration, and would like a decision before doing so. Rewriting existing games to the new spelling changes the state of every library game that uses these parameters, which means re-recording the expected hashes in all four `tests/testcafe/publiclibrary-*.js` files, and it leaves the prose of the ~20 `Functions - *` tutorials describing parameter names their own JSON no longer uses. Without it, old games keep working unchanged (the engine reads the old name whenever the new one is absent) and nothing about this PR can break a published game — but `collection` stays legal as an input forever.

If you want the migration in this PR, say so and I will add it together with the re-recorded hashes.

### What the editing UI looks like

The visual routine editor gained a "what is in a holder" option for every function that has a holder now, worded like the one COUNT already had, and the holder marks the `target` it replaces as ignored, so the two are mutually exclusive in the UI. A function that still carries an old spelling keeps opening: the old name is worded in place of the current one exactly while the engine falls back to it, and its chip gets the orange "!" deprecation note that the CANVAS `canvas` parameter already had. Adding a MOVE from the editor writes `{"func":"MOVE","fromHolder":null,"count":1}` and reads *"Move 1 widget from holder(s)/collection to holder(s)/collection"*.

Screenshots from a local run:

- the editor on a ROTATE that uses `holder`: https://agent.virtualtabletop.io/reports/pr/1536305652446535690/routine-editor-holder.png
- a MOVE added from the editor: https://agent.virtualtabletop.io/reports/pr/1536305652446535690/routine-editor-new-move.png
- the new tutorial, variant "target and holder": https://agent.virtualtabletop.io/reports/pr/1536305652446535690/tutorial-target-and-holder.png
- variant "MOVE": https://agent.virtualtabletop.io/reports/pr/1536305652446535690/tutorial-move.png
- its library image: https://agent.virtualtabletop.io/reports/pr/1536305652446535690/tutorial-library-image.png
- the "i" text behind a ROTATE, which now documents `target` and `holder`: https://agent.virtualtabletop.io/reports/pr/1536305652446535690/routine-editor-info.png

If the wording of any of it reads wrong to you, say which sentence and I will change it.

### Verification

`npm test` (jest, 1206 tests) is green, and locally I ran the TestCafe files most likely to notice a behavior change — `routines.js`, `publiclibrary-1..4`, `interactions.js`, `holderevents.js` — 76 tests, all passing without touching a single expected hash.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-documented unification — the editor, `jeCommands`, validator and tutorial work all landed, and the deprecated-name fallbacks are convincing. But the claim "no existing game changes behavior" does not quite hold: **MOVE from a collection into several destinations now re-moves the same widgets instead of dealing on** (`let offset = 0` moved inside the destination loop), and a leftover `from: ""` now stops a MOVE that used to work. Both are reproducible against `main`.

## Findings

### 1. `MOVE` with `target`/`collection` and several destinations: regression (high)

`client/js/widgets/widget.js` — in the `target` branch:

```js
for(const target of collections[destinations]) {
  let offset = 0;                                   // <-- was declared OUTSIDE the per-destination loop
  for(const c of collections[targetWidgets].slice(offset, offset+count))
    offset += await applyMove(...);
```

On `main`, `offset` lived outside `await w(a.to, …)` and accumulated across destinations, so `count`/`fillTo` dealt the *next* widgets to the *next* destination. Resetting it per destination makes every destination take the same first `count` widgets — the later destination simply steals them from the earlier one.

Reproduced with the jsdom engine harness (same routine on `main` vs. this branch):

```js
{ func: 'SELECT', type: 'all', property: 'id', relation: 'in', value: [ 'c1', 'c2' ] },
{ func: 'MOVE', collection: 'DEFAULT', to: [ 'd1', 'd2' ], count: 1 }
```

* `main`: `c1 → d1`, `c2 → d2`
* this branch: `c1 → d2`, `c2` never moves

That is the classic "deal one card to each player's holder/seat" pattern, and it is exactly what the help text this PR itself edits still promises ("If there are enough widgets remaining in the source, then it will move to the next destination"). Fix: hoist `let offset = 0;` above `for(const target of collections[destinations])` again.

### 2. A leftover `from: ""` now stops the MOVE (medium)

`if(a.from === null) delete a.from;` covers `null`, but the old code branched on `if(a.from)` — plain truthiness. The visual routine editor writes `add: { [parameter]: '' }` for a widget/collection chip that has been added but not filled in, so `from: ""` does occur in saved games. Old behavior: falls through to `collection`. New behavior: `fromHolder` is *defined*, so MOVE takes the fromHolder branch, reports `'' is neither the name of a collection nor a widget ID` and moves nothing. Verified with the harness (`main`: card moves; this branch: it doesn't). `if(!a.from) delete a.from;` restores the old rule exactly.

### 3. Temporary collections are never cleaned up (medium)

`getCollection()` now allocates `'$collection_' + ++temporaryCollectionCount` and never removes it. Confirmed observable: after two operations that pass id lists, `{ func: 'SET', collection: '$collection_1', … }` still resolves and writes to that widget. With `target`/`holder`/`fromHolder`/`toHolder` all accepting id lists, essentially *every* widget operation written by the new editor allocates one, so:

* a `FOREACH` loop grows `collections` without bound for the whole routine;
* every `CALL` deep-copies all of them into the child (`inheritCollections`);
* `DELETE` walks `for(const c in collections)` per deleted widget.

Cheapest fix: after each operation in the `for(const original of routine)` loop, drop keys matching `/^\$collection_/`. (Fixing the `batchDepth` `ReferenceError` was right — but the old name at least got reused rather than accumulating.)

### 4. `CANVAS` reports an empty *holder* as an empty `target` (low)

```js
} else if(groups) {
  problems.push(`Collection ${a.target} is empty.`);
}
```

`FLIP` and `ROTATE` guard this with `a.holder === undefined`; `CANVAS` doesn't, so a CANVAS with an empty `holder` reports "Collection DEFAULT is empty" — naming a parameter the operation never read. It also re-introduces exactly the "error for an expected state" that was objected to earlier in this review thread for ROTATE.

### 5. Test coverage for the regressed cases

`tests/client/engine/widget-parameters.test.js` is a good pin of both spellings, but nothing in it exercises the two cases above:

* `target`/`collection` → **several** `toHolder`s with `count` (and with `fillTo`) — the multi-destination dealing semantics;
* a half-filled `from: ""` (there is a case for `from: null`, which is how #2 slipped through).

Both are one-liners in the existing `parents()` helper and would have caught #1 and #2.

### 6. Smaller notes

* `CALL` with `return: false` `break`s after the first widget, so it calls exactly one routine — the PR body's "it now calls the routine on every widget" only holds for `return: true`. Worth a sentence in the help text (it is defensible: `return: false` aborts the caller anyway). With `return: true`, only the *last* callee's `variable`/`collection` survive, which the code comments but the wiki text does not.
* `MOVEXY` gained `target: 'DEFAULT'`, so an operation that is missing `from` no longer errors but silently moves the DEFAULT collection. Probably intended, but it is a behavior change for existing broken games.
* Leaving the FileUpdater migration out is the right call for now, for the reasons you list — the old names have a validator entry, an editor chip and an orange deprecation note, so nothing is lost by deferring it.

### UI and tutorial (maintainer checklist)

* **Editing UI: OK.** `client/js/editor/controls/routine.js` gives every new `target`/`holder`/`fromHolder`/`toHolder` a real clause in the visual routine editor (with `holderClause`/`targetClauses`, deprecation markers on the old chips and help text in `popup.js`), and `jeCommands` is updated in step. This is not JSON-editor-only support.
* **Tutorial: OK.** `library/tutorials/Functions - Widget Inputs/` is a real tutorial with two variants (`0.json` target/holder/collection + holder-beats-target, `1.json` MOVE), the standard title/overview/example-row/caption bands, `_meta.info` parity with `Functions - AUDIO`, and a 530×530 square SVG library image in its own `assets/`. No other tutorial is touched. Two small things: both variants carry the same title text `"Functions: widget inputs"`, where the newest multi-variant tutorial (`Functions - INPUT`) puts the variant in the title (`"Functions: INPUT (Basic)"`); and with 18 functions changed, a third variant covering `holder` on a **seat** and "count spent per holder" would earn its place — those are the two facets a game author is most likely to get wrong.

All 42 CI checks pass and `npm test` is green locally (1209 tests), so the library state hashes are unaffected — which also means finding #1 is not covered by any library game.

I did not add the `beta` label: this PR changes routine-operation semantics for ~18 functions in `client/js/widgets/widget.js`, which is a genuine engine change and therefore excluded from beta auto-deployment by design.

VTT-AI-Agent:
Thanks — the MOVE finding was real and the reproduction was exact. All six points are addressed; here is what changed, per finding.

**1. MOVE with several destinations (high) — fixed.** `let offset = 0` is back above `for(const target of collections[destinations])`, so the count is spent across the destinations one after the other again and `MOVE target count:1` into two holders deals one widget to each. I re-ran your case against `main` and this branch and they now agree (`c1 → d1`, `c2 → d2`).

**2. A leftover `from: ""` (medium) — fixed.** `if(!a.from) delete a.from;`, which restores the old truthiness rule exactly. Pinned by a test that uses `from: ''` rather than `from: null`.

**3. Temporary collections (medium) — fixed.** The routine loop remembers the counter before each operation and deletes every `$collection_N` the operation allocated once it is done, so nothing accumulates for a `FOREACH`, nothing extra gets deep-copied by a `CALL`, and `DELETE` walks the same collections it always did. Nested routines are safe: one that shares `collections` by reference takes its own snapshot after the outer temporaries exist, so it only ever removes its own.

**4. CANVAS on an empty holder (low) — fixed**, and the same guard was missing in **LABEL and TIMER**, which reported `Collection DEFAULT is empty` for an empty holder — `a.target` there is the default the operation never read. All three now only report an empty *collection*, the way FLIP and ROTATE do.

**5. Test coverage — added.** `tests/client/engine/widget-parameters.test.js` gained: dealing to several destinations with `count` (new and old spelling), the same with `fillTo`, the half-filled `from: ''`, and one that asserts no `$collection_N` survives its operation. I checked each of them fails on the code as it was before this round.

**6. Smaller notes.**
- **CALL with `return: false`** — kept as it is (it aborts the caller anyway), and the help text now says so, together with the fact that `variable`/`collection` come from the last callee when several widgets are named.
- **MOVEXY** — left as it is, but written down under "Also fixed" in the PR body so it is a visible decision. I checked all 573 library files: exactly one has a bare `{"func":"MOVEXY"}` (`library/games/Card Games`, `gameID25/dealRoutine`) and it runs with no `DEFAULT` collection in scope, so nothing published changes. Happy to make it report a missing parameter instead if you would rather have the old error.
- **FileUpdater** — agreed, still deferred.

**Tutorial.** Both notes taken: each variant names itself in its title now (`Functions: widget inputs (target and holder)` / `(MOVE)` / `(seats and counting)`), and there is a third variant for exactly the two facets you named — a `count` spent per holder against the same widgets named as one collection, and `holder` on a seat, including the MOVE into a seat that gives the widgets to its player in the first place (without which the seat contributes nothing, which the variant says out loud). Screenshots: [the room](https://agent.virtualtabletop.io/reports/pr/1536305652446535690/tutorial-seats-and-counting.png), [after pressing every button](https://agent.virtualtabletop.io/reports/pr/1536305652446535690/tutorial-seats-and-counting-in-use.png).

**Verification.** `npm test` green (1214 tests), and locally `routines.js`, `publiclibrary-1..4`, `interactions.js` and `holderevents.js` — 71 tests — pass without touching a single expected hash.

**beta label** — agreed, not added; this is an engine change.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review only (a separate reviewer covers correctness). The editor side of this rename is in good shape — every operation now words its widgets in one vocabulary, the deprecated spellings are marked and explained, and nothing in the shared overlays broke. The most impactful improvement: **the sentence still hides two things the engine decides — that a `count` is spent *per* holder ("Rotate 1 widget in holderA and holderB" actually turns two), and that a `target` set alongside a `holder` is dropped without a trace in the sentence.** Both are exactly the traps this PR's own tutorial was written to warn about, so the editor should say them out loud.

**What I exercised:** ran the app locally and drove it with Playwright. Viewports: **390×844**, **820×1180**, **1280×800**, **1920×1080**. Overlays/panels opened at every size: room toolbar, **Game Shelf**, **Players**, **About/help**, **Active Game**, edit-mode **sidebar + Edit Widgets/Properties module**, the **Automations section**, the **routine editor** (with routines written in the new spelling, the old spelling, both-set, and bare `{"func": …}` operations), the **chip value picker**, the **"add option" menu of every affected operation**, the **parameter info popups**, the **deprecation popup**, the **JSON editor module and its context buttons**, and the new tutorial room `Functions - Widget Inputs / 2.json`. Not opened (untouched by this diff): add-widget/deck editor, state save/share/publish dialogs, chat.

---

## Big layout improvements

### 1. A `target` the engine ignores disappears from the sentence entirely

`clauses()` drops every parameter listed by `ignored` (`client/js/editor/controls/routine.js:3006-3010`, `ignoredParameters()` at `:3128`), and the reason string `targetReplacedBy()` produces — `"ignored because holder is set"` (`:157`) — is **never rendered anywhere**. So a routine written as

```json
{ "func": "CLICK", "holder": "holderA", "target": ["card1"] }
```

reads simply `Click what is in holderA`. There is no greyed chip, no note, no hint that `target` is in the JSON and being thrown away:

both set

This pattern existed for exactly one parameter pair before (CANVAS `canvas` vs `collection`); this PR makes "holder wins over target" the rule on ~15 operations, so it will now be hit constantly — and a designer debugging "why is my target ignored?" gets no answer from the editor. Suggestion, in the app's existing idiom: keep the ignored chip in the sentence rendered muted/struck-through with the existing warning-button treatment (`parameterWarningButton`, `:3421`) carrying the reason text, or append a small grey `(target ignored)` tail. The string is already computed — it just needs a place on the card.

### 2. Say that a `count` is spent per holder

The tutorial added by this PR exists because this is easy to get wrong ("*a `count` is spent on every widget the holder names separately*"). The sentence does not say it:

tutorial routine

`Rotate 1 widget in holderA and holderB by 45 degrees` rotates **two** widgets. Same for `Cycle the face of 1 widget in holderA and holderB`, and for MOVE/CANVAS. `countedInHolderOrOfGroup()` (`routine.js:335`) words the count identically whether one holder or five are named. Proposal: when the holder value is a list/collection, word it `1 widget in each of holderA and holderB` (and keep the current phrasing for a single holder). One word of wording removes the need for the tutorial paragraph.

### 3. Three different ways to reach the same "work on what is in a holder"

I opened the "add option" menu of every affected operation. What a first-time user is offered is not consistent:

| how holder is reached | operations |
| --- | --- |
| `add option → what is in a holder` | CANVAS, CLONE, COUNT, DELETE, SET, LABEL, TIMER, CLICK |
| worded differently in the menu | CALL (`of what is in a holder`), MOVEXY (`out of a holder`) |
| no menu entry at all — only by clicking the combined chip | FLIP, ROTATE, SORT, SHUFFLE, MOVE |

FLIP's whole menu is `at most a certain number of them`; ROTATE has no "add option" button at all. Since the point of the PR is that a designer no longer has to care which function they are in, the *editor* should stop making them care either: offer the same `what is in a holder` entry on all of them (the combined chip can stay as the shortcut). Right now FLIP and ROTATE — the two operations that had `holder` the longest — are the ones where it is least discoverable.

---

### 4. The new deprecation note renders as four ragged mid-sentence paragraphs

deprecation popup

The info popup makes a paragraph out of **every line** of a `<pre>` — the code says so itself at `undeterminedInfoHTML` ("*a sentence broken over two of them would be two paragraphs*"). `deprecatedNameNote()` (`routine.js:272`) hard-wraps its text at ~100 characters, so the reader gets paragraph breaks after "*reads the*", "*the name of a*", "*said with target,*". Fix is one line each: put every sentence of the note on a single source line. This popup is now reachable from ~15 operations, so it is worth the tidy-up.

### 5. SHUFFLE and SORT with a `holder` read as if the holder itself were the subject

`{"func":"SHUFFLE","holder":"holderB"}` renders **`Shuffle holderB`**, and SORT the same way — while every other operation says `Delete what is in holderB`, `Click what is in holderA`. Both use `deprecatedTargetSlot([...], ' {holder,source}')` (`:1382`, `:1416`), which puts holder and source in one chip with no preposition. `holder` is *new* on these two in this PR, so the ambiguity is new too. Suggest `Shuffle what is in holderB` when the holder slot is the one filled.

### 6. `holder(s)/collection` as the blank-chip hint

`holderHint` (`routine.js:170`) shows up in the sentence of an unfilled MOVE/RECALL: `Move the picked widgets to holder(s)/collection`. The slash-plus-parenthetical is out of step with the plain-English blanks around it (`the picked widgets`, `a named group of widgets`). Something like `a holder or a group of them` fits the voice better.

### 7. COUNT's holder sentence has three `and`s in a row

`Count what is in holderA and holderB and remember it as n` — the eye can't find where the holder list ends. A comma before the last clause (`…holderA and holderB, and remember it as n`) fixes it; the clause template is at `:745`.

### 8. A widget id written into `target` now reads as a collection name

Same value, two spellings, two different sentences:

* old: `{"canvas": "canvas1"}` → `Set the value of **canvas1** to 3`
* new: `{"target": "canvas1"}` → `Set the value of **the canvases called canvas1** to 3`

new spelling old spelling

`namedGroupClause()` (`:192`) words every non-array string as `the <noun> called {param}`, but the engine reads a string that names no collection and does name a widget as *that widget* — which is the compatibility rule this PR advertises. So the recommended new spelling describes itself less accurately than the deprecated one it replaces, and `the canvases called canvas1` will send a newcomer looking for a collection they never created. Suggestion: when the string matches a widget id in the room and no collection, word it as the widget (the `widgetIDParameterNames` logic in `oldTargetTemplate` already does exactly this for the old names — it just needs to key off the value rather than the parameter name). Same for LABEL (`the widgets called l1`) and TIMER.

### 9. LABEL keeps the generic noun

CANVAS says `the canvases called …`, TIMER `the picked timers`, LABEL says `the widgets called l1`. `noun: 'labels'` would match its siblings.

### 10. The longer wording costs width in the panel

At the default panel width the new phrases wrap mid-sentence even on a 1920 desktop — `Set the value of / the canvases called canvas1 to 3`, `Make 1 copy of what is in holderA — call the / copies copies`, with the dashed `add option` button landing on its own line:



Not introduced by this PR (the panel width is remembered and resizable — at 820 px wide every operation fits on one line, [tablet](https://agent.virtualtabletop.io/reports/ui-review/1536348394010710126/sm-820x1180-routine.png)), but the new `what is in` / `the canvases called` phrasing makes it noticeably worse. Another reason to prefer the shorter wording in #8.

**Everything else held up.** No overflow, clipping, collision or unreachable control in Game Shelf, Players, About or the edit-mode panel at any of the four viewports; the routine editor is usable and closable down to 390 px wide (via the sidebar's tune icon — edit mode still shows the "please rotate your device" hint at phone portrait, which is pre-existing), [phone](https://agent.virtualtabletop.io/reports/ui-review/1536348394010710126/sm-390x844-routine.png). The one measured overflow — About scrolling horizontally at 390 px — is pre-existing and unrelated to this diff. The JSON editor's context buttons offer exactly the parameters of the operation the caret sits in ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1536348394010710126/je-module-flip.png)).

---

### 11. The chip popup does not say which of the two names it is editing

Clicking the holder chip and clicking the target chip both open a popup titled **`CLICK - which widgets`**; the only difference is the small grey `holder` / `target` next to the title.

chip popup

For someone who has just learned that the two mean different things, the title is the wrong place to be silent. `CLICK — which holders` (or `— what to look inside`) for the holder variant would carry it.

### 12. "deprecated" is developer jargon in a tooltip aimed at game designers

The orange `!` has `title="deprecated - click for details"` (`parameterWarningButton`, `:3421`). The popup behind it is written in plain words — the tooltip could be too: *"old name — click to see what to use instead"*.

### 13. The help texts are the strongest part of the change

The new `[holder]` topic ("*for a seat that content is what lies in its hand… a count is spent per widget named here: one card off every holder rather than one card in total*") is the clearest statement of the new model anywhere in the PR, and the per-parameter lines read well:

info popup

Only `holder` lines link to it (`See [holder]`). Worth linking from the `target` lines too — the pair only makes sense read together, and `target` is the one a newcomer clicks first.

### 14. The tutorial room reads well

`Functions - Widget Inputs` variant 2 is clear at every size I opened it: the two-column "holder vs target" comparison lands, the text blocks sit next to the widgets they describe, and the buttons say what they do rather than which JSON they contain. ([room](https://agent.virtualtabletop.io/reports/ui-review/1536348394010710126/room-1280.png)) One thought: the paragraph explaining that a `count` is spent per holder is doing work the sentence in the editor should be doing (#2) — if that lands, this paragraph gets to be a confirmation rather than a warning.

VTT-AI-Agent:
Thanks — this was a useful pass; the two big ones were both real and both are now said out loud in the sentence. Everything is in `09a2f2c` on this branch. Per finding:

### 1. A `target` set alongside a `holder` vanishes — fixed, and generalised

The reason `targetReplacedBy()` computes was indeed never rendered. It is now: **every** parameter the engine skips because of how another one is set stays in the sentence, struck through and faded, with a red "!" behind it whose popup says why nothing comes of it and that the ⊖ takes it out without changing anything. Your example now reads `Click what is in h1, ~~target~~ card1 ❗`.

It is not special-cased to holder/target — the `ignored` mechanism already existed for `MOVE count` vs `fillTo`, `CANVAS x/y/color`, `MOVEXY z: 0`, `AUDIO silence` and `IF condition`, and all of them were being dropped just as silently (the header comment in `routine.js` even claimed they got a red "!" in a list view that no longer exists). One case I added while I was there: the *old* spelling of the widgets is skipped for exactly the same reason the current one is, so `{"func":"SET","holder":"h1","collection":"aces"}` now shows `collection aces` struck through too, where it used to disappear.

A parameter the operation does *not* have is still neither shown nor offered — nothing new appears on a clean operation.

ignored

### 2. A `count` is spent per holder — fixed, in your wording

`Rotate 1 widget in each of h1 and h2 by 45 degrees`, and the same on `FLIP`. Also on the two sides of `MOVE`, which deals per source *and* per destination: `Move 1 widget from each of h1 and h2 to d1`, `Move 1 of the widgets called aces to each of d1 and d2`, and `… to d1 and d2 until each of them holds 3` when it is topping them up. `CANVAS` words it in its own option: `, for 1 canvas in each of them`.

"each of" only appears where the count really is spent more than once: a list of two or more, or a collection name (which may hold any number), and only while the count is a limit at all — `Turn all widgets in h1 and h2 face up` says nothing about "each", because taking everything out of both holders is the same thing said twice. A single id, or a `${...}` the editor cannot count, keeps the short wording rather than guessing at a plural.

### 3. Three ways to reach a holder — evened out

`FLIP`, `ROTATE`, `SORT` and `SHUFFLE` now offer **`what is in a holder`** in their "add option" menu like every other operation, instead of only through the combined chip (which is gone — they use the same `holderClause()` + `targetClauses()` pair as CANVAS/CLICK/DELETE now). `CALL` and `MOVEXY` lost their differently-worded labels, so the phrase is identical on all thirteen. There is a test pinning that.

`MOVE`, `MOVEXY`, `COUNT` and `RECALL` keep asking it through the **way-of-working drop-down** (`Move widgets from a holder`, `Count what is in a holder`) rather than as an option, and I left that as it is on purpose: their whole sentence reads differently with a holder — a MOVE out of a holder counts 1 by default and the picked widgets count "all" — so it is a variant, not something you switch on. The same drop-down is the first thing on the card, so it is not less discoverable than the option; it is a different question.

### 4. The deprecation popup broke into mid-sentence paragraphs — fixed

One line per sentence in `deprecatedNameNote()`, so it renders as three paragraphs that each end where the sentence does. The `skip` note had the same problem and got the same treatment.

deprecation

### 5. "Shuffle holderB" — fixed

Follows from #3: `Shuffle what is in holderB`, `Sort what is in holderB by value`, worded like every other operation.

### 6. `holder(s)/collection` — fixed

The blank now says **`a holder or a group of them`** (and RECALL's `a deck or a group of them`). All eighteen holder parameters come from one `holderSpec()` helper now, so there is one place to word it.

### 7. Three `and`s in COUNT — fixed

`Count what is in h1 and h2, and remember it as n`, with the comma only where the widgets really are a list — `Count what is in h1 and remember it as n` is unchanged.

### 8. A widget id in `target` read as a collection — fixed

A string that names a widget of the room and no collection is now worded — and coloured — as that widget, wherever it is written: `Set the value of canvas1 to 3` under both the new and the deprecated spelling. A name no widget has stays `the canvases called myCanvases`, and a collection an earlier operation made wins over a widget of the same name, because that is the order the engine reads them in.

routine

### 9. LABEL's noun — not changed, on purpose

LABEL is not limited to labels: it calls `setText()` on every widget it is given, so any widget with a text property is a valid target (the parameter's own comment says so, and the picker opens on all types for that reason). `the widgets called l1` is the accurate noun there, where `canvases`/`timers` are accurate for CANVAS and TIMER, which do skip widgets of the wrong type.

### 10. Width — partly, through #8

The longest new phrases were the ones #8 removed (`the canvases called canvas1` → `canvas1`), so the common case is shorter than before this PR rather than longer. The panel width itself I left alone as pre-existing.

### 11. The chip popup did not say which name it edits — fixed

`CLICK - what to look inside` for a holder against `CLICK - which widgets` for the target, and MOVE asks `where to take them from` / `where to move them to`, RECALL `where to gather them`.

holder popup

### 12. "deprecated" as a tooltip — fixed

`old name - click to see what to write instead`.

### 13. Link `[holder]` from the `target` lines too — done

All sixteen of them.

### 14. The tutorial paragraph

Kept, and it now names the wording it is confirming (*"which is why the routine editor words it '1 widget in each of'"*), so it reads as the explanation behind a sentence the reader has already seen rather than as a warning about something invisible.

---

**Verification.** `npm test` green (1217 tests, four new ones covering the ignored rendering, the per-holder count, the widget-id wording and the holder being offered on every operation that has one), plus `tests/testcafe/editor.js` (40 tests) and `routines.js` locally. The whole thing was walked through in a browser at 1500×1000 — the screenshots above are from that run; [here](https://agent.virtualtabletop.io/reports/pr/1536305652446535690/ui-routine-1.png) is the ignored `target` on the card itself.

VTT-AI-Agent (commented):
## TL;DR

The previously reported MOVE and cleanup regressions are fixed and covered, and the author-facing editor/tutorial work is present. One resolver bug remains: an explicit widget-ID array can resolve to an unrelated named collection, which makes destructive operations such as DELETE act on the wrong widgets; generated temporary names can also erase an author-defined collection.

## Findings

### 1. Explicit ID arrays can resolve to a same-named collection (high)

`client/js/widgets/widget.js:980-986`

`getCollection()` looks up `collections[collection]` before checking whether `collection` is an array. JavaScript coerces an array used as an object key to a string, so an explicit one-item ID list such as `["card"]` indexes `collections.card`. If a collection named `card` already exists, the function returns the input array as `newCollection`; the later `collections[target]` lookup coerces it to `"card"` again and operates on that collection instead of on the widget whose ID is `card`.

This affects the new canonical input shape emitted by the visual editor and can target the wrong state. For example, with a widget `card` and a collection `card` containing other widgets, `{ "func": "DELETE", "target": ["card"] }` deletes the collection's widgets rather than the widget explicitly named in the array. The validator considers the array a widget-ID list, so it gives no warning.

Check `Array.isArray(collection)` before consulting `collections[collection]` (and only perform the named-collection lookup for strings). Please add a regression test where a widget ID and collection name collide; DELETE or SET makes the wrong-target behavior especially clear.

### 2. Temporary collection names can overwrite and then delete real collections (medium)

`client/js/widgets/widget.js:984-986, 2428-2432`

Every ID-array input is stored under a generated `$collection_N` key without checking whether that key is already present. Cleanup then unconditionally deletes it. Collection names are unrestricted, so if a routine already has a collection named `$collection_1` (or whichever counter value comes next), an unrelated operation using `target: ["card"]` replaces that collection and removes it at the end of the operation. A later operation referencing the author's collection then fails or works on the wrong widgets.

Generate a key that is absent from the current `collections` object, or save and restore any displaced value during cleanup. A test that creates `$collection_1`, uses an ID-list target, and then consumes `$collection_1` would pin the expected lifetime.

## Maintainer checks

The real visual routine editor, JSON editor commands, validator, help text, and a three-variant tutorial under `library/tutorials/Functions - Widget Inputs/` are all included; the tutorial has matching metadata and its own square SVG library image, and no unrelated tutorial is changed. All reported CI checks, including CodeQL, browser suites, and library validation, are green; the local focused Jest run could not start because this clone has no Jest binary installed.

I did not add the `beta` label because `client/js/widgets/widget.js` changes routine-operation semantics and therefore this PR is engine-ineligible for beta auto-deployment.

VTT-AI-Agent:
Both findings were real and both are fixed in `58923a9`.

### 1. An explicit list of ids resolving to a same-named collection — fixed

`getCollection()` asked `collections[collection]` before it asked whether it had a list at all, so an array used as a key became the string `card` and `{"func":"DELETE","target":["card"]}` deleted the collection's widgets instead of the widget the list names. The list is checked first now, so a list of ids always means those widgets, whatever a collection happens to be called.

The ordering was already there before this PR, but it only became reachable through a normal spelling with it — a list of ids used to arrive at `getCollection()` only through `collection`/`source`/`excludeCollection`, and it is what the routine editor writes into every widgets parameter now. Thanks for catching it.

A plain **string** keeps reading as a collection first and only as a widget when no collection has that name — that is the order the engine has always used and what routines that build a collection rely on, so `target: "card"` still means the collection. The two spellings are pinned against each other in the new test.

### 2. A generated `$collection_N` overwriting a collection of the game — fixed

The counter now steps over every name already present in `collections`, so it never lands on one the game gave out (or on a temporary one of a surrounding routine that is still alive), and cleanup deletes the names this operation really created instead of counting a range — it keeps the list of them rather than recomputing it from the counter.

### Tests

`tests/client/engine/collection-names.test.js`, three cases: the list of ids beating the same-named collection, the string form still naming the collection, and a game-defined `$collection_1`…`$collection_12` surviving an operation that passes a list of ids. The first and third fail on the previous commit and pass on this one.

`npm test` green (1223 tests), `tests/testcafe/routines.js` green locally.

VTT-AI-Agent (commented):
## TL;DR

The regressions from the earlier rounds are gone: MOVE deals on across destinations again, a leftover empty `from` falls back to `target`, and the temporary collections are name-safe and cleaned up. `npm test` is green (1220 tests), the tutorial and the editor/`jeCommands`/validator work are all there. One compat hole remains: for the parameters that used to be **widget-ids only** (`holder`, `from`/`to`, `label`, `canvas`, `timer`), a plain string is now looked up as a **collection first**, so an existing game whose collection name happens to match a holder id silently changes meaning — and that is the exact opposite of the array rule 58923a99 just established.

## Findings

### 1. A string that names both a collection and a widget now resolves to the collection — the opposite of what the array case does (medium)

`client/js/widgets/widget.js:1020-1026` (`widgetParameter`):

```js
if(typeof value == 'string' && !Array.isArray(collections[value])) {
  if(!widgets.has(value)) { … }
  value = [ value ];
}
```

The collection lookup wins; only when no collection carries the name is the string read as a widget id. For `target`/`source` that is the intended new rule. But `holder`, `from`, `to`, `label`, `canvas` and `timer` **never accepted a collection before this PR** — they were widget ids, full stop — so any routine that has a collection in scope whose name equals the holder's id changes behaviour.

Reproduced with the jsdom engine harness (`tests/client/engine/harness.js`), holders `h1` (children `c1..c3`) and `h2` (children `e1,e2`):

```js
{ func: 'SELECT', property: 'parent', value: 'h2', collection: 'h1' },
{ func: 'ROTATE', holder: 'h1', angle: 90, count: 'all' }
```

* `main`: rotates `c1,c2,c3` — the children of the *widget* `h1`
* this branch: rotates **nothing** — `h1` resolves to the collection `[e1,e2]`, and `.map(h=>h.children())` on those is empty

Note the inconsistency this leaves inside one operation: after 58923a99, `holder: ["h1"]` is the widget, while `holder: "h1"` is the collection. Same value, two meanings, chosen by the brackets.

Public library games do name collections after widgets — `discards` in `Rackball/0.json`, `reserve1` in `Splendid/0.json`, `board`/`sq33` in `Reversi/0.json`, `removedCards` in `Myco Meld/0.json`, `auction_block` in `Union Station/0.json` — and several of those names are used as `from`/`to`/`holder`/`label` in the same file. They survive only by luck: the colliding collection is either out of scope in that routine or was built as `SELECT property:'id' value:X collection:X`, so it contains exactly the widget the parameter meant. That is a thin margin for "no existing game changes behavior", and it does not cover games outside the library.

Suggestion: for the parameters that historically took widget ids only, resolve the widget first and fall back to the collection (mirroring the array rule), or keep collection-first but push a problem when a string is ambiguous so the author sees it. Either way this deserves a test in `widget-parameters.test.js` — the current suite pins the array collision but not the string one.

### 2. The old `collection` spellings now silently accept a bare widget id (low-medium)

Because `collection` is folded into `target` by `renamedParameters()` and then run through `widgetParameter()`, the widget-id leniency that used to belong to `holder`/`label`/`timer` now applies to `collection` too. `{"func":"DELETE","collection":"someWidget"}` used to report *"Collection someWidget does not exist or is not an array"* and do nothing; it now deletes that widget.

This is live in the library today: `library/games/Union Station/0.json`, `initial_auction_holder/InitialAuctionCountRoutine` has `{"func":"GET","collection":"auction_block","property":"hiddenX"}` with no collection of that name in scope. On `main` that is a reported error and `$hiddenX` stays unset; on this branch it reads the property off the widget. Here the value happens to be unused (the following `SET` uses `${PROPERTY hiddenX}`), so nothing observable changes — but the same leniency on `DELETE`/`SET` turns an author's typo from a visible error into silent state loss. Worth a deliberate maintainer decision: `widgetParameter()` could take a flag so only the parameters that always named widgets keep the single-id shorthand.

### 3. `RECALL target:` on a deck outside a holder parents the cards to the deck widget (low)

`client/js/widgets/widget.js:1899-1907`: when `target` names a deck whose `parent` is not a widget, the destination becomes the deck itself, and the cards are `moveToHolder()`-ed into it. Confirmed in the harness — a card recalled that way ends up with `parent: "dk"`. A deck is not a holder (86 × 86 default, no `updateAfterShuffle`, no fill logic), so the cards land stacked on an ordinary widget rather than being collected anywhere useful. Reporting *"deck X is in no holder"* would be less surprising than a half-working recall; if the current behaviour is deliberate, say so in the help text, because the tutorial does not demonstrate this case.

### 4. Compat claim on CALL — checked, holds in practice (info, no action)

`CALL` running the routine on *every* named widget instead of only the first is a real semantic change to existing saves, not just a rename. I scanned all 573 library game/tutorial files: **no** file uses `CALL` with a multi-id `widget` array, so nothing in the library is affected. Worth keeping the sentence in the PR body, since third-party saves are not covered by that scan.

## What checks out

- `npm test`: 30 suites / 1220 tests pass on `58923a99`.
- The earlier findings are genuinely fixed and pinned: `offset` is hoisted above the destination loop again, `if(!a.from) delete a.from` restores the truthiness check, `getCollection()` checks `Array.isArray` before the key lookup, generated `$collection_N` names skip ones a game already took, and the per-operation cleanup runs before the `abortRoutine` break so nothing leaks.
- UI support for everything new is present, as the standing rule requires: `client/js/editor/controls/routine.js` offers `target`/`holder`/`fromHolder`/`toHolder` as real clauses with `holderSpec()` questions and marks each old spelling with `deprecatedNameNote()`, and `jeCommands` in `client/js/jsonedit.js` inserts the new names with collection enums scoped so `'all'` only appears on SELECT `source` and TURN `target`.
- Test room: `library/tutorials/Functions - Widget Inputs/` is a real tutorial with three variants, `_meta.info` parity against `Functions - AUDIO`, a 530×530 square SVG badge in its own `assets/`, and semantic-id band layout matching the newest tutorials. All three validate clean, and no other tutorial is touched.
- `validator/validate_gamefile.js` carries `collectionOrIDs` for every renamed parameter including the deprecated spellings, and `inputParameters` replaces the old table that had a duplicated `COUNT` key.

I did not add the `beta` label: `client/js/widgets/widget.js` here rewrites how ~18 routine operations resolve the widgets they act on, which is a genuine engine change to how existing games run, so this PR is excluded from beta auto-deployment by design.

VTT-AI-Agent:
All three actionable findings were real and are fixed in `e3556a1`. Finding 4 needs no code, but I took the sentence you suggested into the PR body.

### 1. A string that names both — fixed by giving each parameter back the reading it always had

You are right that collection-first is the wrong rule for the parameters that were widget-ids only, and right that it contradicted the array rule from `58923a99`. What a plain string means is now decided by what the parameter accepted **before** this branch unified them:

| written as | reading | why |
| --- | --- | --- |
| `holder`, `fromHolder`, `toHolder`, and the old `from`/`to`/`canvas`/`label`/`timer`/`widget` | **widget first**, collection only when no widget carries the name | they named widgets and nothing else, so a game with a collection of the same name in scope keeps meaning the widget |
| `target`, `source` (SORT/SHUFFLE) | collection first, widget id as the shorthand this PR adds | they have always named a collection |
| the old `collection`, `excludeCollection`, and `source` in CLONE/SELECT/SWAPHANDS | collection and nothing else | see finding 2 |

A **list of ids is those widgets everywhere**, so the two rules never disagree about brackets — `holder: "h1"` and `holder: ["h1"]` are the same widget again.

Your repro is a test now (`ROTATE holder:'h1'` with a collection `h1` in scope rotates `c1,c2`, the children of the widget, as on `main`), together with the same collision under the old `from`/`to` spelling of MOVE.

Where a name really is ambiguous — a collection **and** a widget carry it, and the collection is not simply that one widget — the routine now reports which of the two it used, e.g. *"h1 is both a widget ID and the name of a collection - fromHolder names the widget, as it always did. Rename the collection to use it here."* The `SELECT property:'id' value:x collection:x` shape that the library games use is exempt on purpose: both readings are the same widget there, so a message would be noise.

One case your review helped me find while pinning this: `renamedParameters()` writes the current name onto the operation object, so the old spelling it came from is only visible the **first** time a routine runs. The reading is remembered per operation now, otherwise `MOVE from:'h1'` inside a FOREACH would have read `h1` one way on the first iteration and the other way on the second. That is a test too.

### 2. The old `collection` spellings accepting a bare widget id — fixed, they are strict again

`{"func":"DELETE","collection":"someWidget"}` reports *"Collection someWidget does not exist or is not an array."* and does nothing, exactly as on `main`. The same for `excludeCollection` and for `source` in CLONE/SELECT/SWAPHANDS, which never renamed and therefore never gained the shorthand. `GET collection:'auction_block'` in *Union Station* is back to the error it reports on `main` rather than reading the property off the widget.

The single-widget-id shorthand is what the **current** names offer: `DELETE target:'card1'` works, `DELETE collection:'card1'` does not. Both spellings are pinned against each other in `tests/client/engine/collection-names.test.js`.

### 3. RECALL onto a deck outside a holder — reported instead

`RECALL target:` on a deck whose parent is not a widget now pushes *"Deck dk is not in a holder, so there is nowhere to recall its cards to."* and leaves the cards where they are, instead of parenting them onto the deck. The help text says the same. A deck that does lie in a widget still recalls into it.

### 4. CALL on several widgets

No change — thanks for scanning all 573 files. The sentence about it stays in the PR body.

### Verification

`npm test` green (1228 tests; 7 new ones covering the holder/from/to collision, the second run of the same operation, the strict old spellings and RECALL on a loose deck — 5 of them fail on `58923a99`). `tests/testcafe/routines.js` and `tests/testcafe/holderevents.js` green locally. Walked in a browser at 1500×1000: with a collection `h1` in scope, `ROTATE holder:'h1'` rotates the widget's children, the routine editor says *"Rotate all widgets in h1"* with `h1` coloured as a widget rather than as a group, and the `[holder]` help topic now spells the whole order out.

The `beta` label stays off — agreed on the reasoning.

PR #1660, opened 2023-03-01
improved asset handling when adding games to the public library

ArnoldSmith86:
When adding a game to the public library using the GUI, the server would crash if it could not find an asset used by the game.

This PR fixes two situations where this might happen:

1. If the asset actually does not exist. In this case it now shows a popup to the user before screwing things up.
2. If the asset is used in another public library game and does not exist in the global assets directory. In this case the asset is now copied from the other game.

### DIFF
--- a/server/room.mjs
+++ b/server/room.mjs
@@ -167,6 +167,10 @@ export default class Room {
     if(!Config.get('allowPublicLibraryEdits'))
       return;
 
+    for(const usedAsset in this.getAssetListForState(id))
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1660/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1660-pr-test`.

96LawDawg (approved):
Approved through test.

PR #536, opened 2021-06-08
fixed that setting activeFace in BasicWidgets would copy the properties of the face to the base state

ArnoldSmith86:
It was never intended that the properties of the currently active BasicWidget face get copied to the widget base state. Instead they should just be _active_ like it works for card faces.

Please do some tests when the properties are actually defined in the base and the faces.

PR #1665, opened 2023-03-03
fixed problems when changing faces on a basic widget

### LINKED ISSUE #1664: BasicWidget does not show text of active face
Wow. Fun times. Turns out that setting `activeFace` in a basic widget does _not_ copy the properties from the face to the widget. Never did.

It copies it to `this.state` which is basically the "actual JSON" of the widget saved inside the Widget object. That is never sent to the server **but** it is what all the editors will show you and once you save, it _will_ be sent to the server (and the JSON editor saves automatically).

You can test this yourself. Add this widget to a new room:
```JSON
{
  "id": "X",
  "faces": [
    { "text": "A" },
    { "text": "B" }
  ],
  "text": "X"
}
```

You would expect it to show `X`. It does. Click it. You expect it to show `B`. It does. Reload the room and it shows `X` again. This happens since https://github.com/ArnoldSmith86/virtualtabletop/commit/f922834c2f9a9b776e292ad6501ff179bfff4156.

Now try this:
```JSON
{
  "id": "X",
  "faces": [
    { "image": "/assets/-807326126_1918" },
    { "image": "/assets/1742233512_1635" }
  ],
  "image": "/assets/1242550039_3992"
}
```

Works just fine. You see the image of the active face. And you can use the state API (`https://virtualtabletop.io/state/<room>`) to check that the property `image` of the widget does not change.

The problem is in this piece of code:

```Javascript
  applyDeltaToDOM(delta) {
    super.applyDeltaToDOM(delta);
    if(delta.activeFace !== undefined || delta.faces !== undefined) {
      let face = this.faces()[this.get('activeFace')];
      if(face !== undefined)
        this.applyDelta(face);
    }
    if(delta.text !== undefined)
      setText(this.domElement, delta.text);

    // ...
  }
```

1. `super.applyDeltaToDOM(delta);` applies the actual `image` property to the widget.
2. Then `this.applyDelta(face);` applies the `image` and `text` properties of the active face to the widget.
3. Then `setText(this.domElement, delta.text);` applies the actual `text` property to the widget.

I'm pretty sure that the fix is to replace `setText(this.domElement, delta.text);` by `setText(this.domElement, this.get('text'));` because `this.get('text')` actually returns the `text` property of the active face (so `GET` should as well). But after this roller coaster, I'll have to ponder about the meaning of life for a second before opening a PR.

This also means that I'll revisit https://github.com/ArnoldSmith86/virtualtabletop/pull/536. We can probably just do what that PR does without changing game behavior. That PR contains the fix since https://github.com/ArnoldSmith86/virtualtabletop/pull/536/commits/e319a30b537c5316463c68fd5ddc54731f3503dd...

ArnoldSmith86:
This is very similar to #536 but it was faster to redo it than to resolve all the conflicts. I also wanted a better way to undo the previous face changes.

This fixes #1664. It does change behavior in production but not in a way as discussed in #536. Turns out the properties of the active face were never copied to the widget in the first place. _But_ changing from a face that _has_ an `image` to a face that _does not_, would previously still show the `image` of the previous face. It no longer does. The behavior in production is clearly buggy though because reloading the room would make the `image` of the previous face disappear. Reloading the room should _never_ change anything so this is a clear no-no bug that will be fixed now.

### DIFF
--- a/client/js/widgets/basicwidget.js
+++ b/client/js/widgets/basicwidget.js
@@ -27,7 +27,7 @@ class BasicWidget extends Widget {
         this.applyDelta(face);
     }
     if(delta.text !== undefined)
-      setText(this.domElement, delta.text);
+      setText(this.domElement, this.get('text'));
 
     for(const property of Object.values(this.get('svgReplaces') || {}))
       if(delta[property] !== undefined)
--- a/client/js/widgets/basicwidget.js
+++ b/client/js/widgets/basicwidget.js
@@ -31,7 +31,7 @@ class BasicWidget extends Widget {
       }
 
       const face = this.faces()[this.get('activeFace')];
-      if(face !== undefined) {
+      if(face && typeof face == 'object') {
         this.applyDeltaToDOM(face);
         this.previouslyActiveFace = face;
       }
--- a/client/js/widgets/basicwidget.js
+++ b/client/js/widgets/basicwidget.js
@@ -22,9 +22,19 @@ class BasicWidget extends Widget {
   applyDeltaToDOM(delta) {
     super.applyDeltaToDOM(delta);
     if(delta.activeFace !== undefined || delta.faces !== undefined) {
-      let face = this.faces()[this.get('activeFace')];
-      if(face !== undefined)
+      if(this.previouslyActiveFace !== undefined) {
+        const undoDelta = {};
+        for(const property in this.previouslyActiveFace)
+          undoDelta[property] = this.state[property] !== undefined ? this.state[property] : null;
+        this.applyDeltaToDOM(undoDelta);
+        delete this.previouslyActiveFace;
+      }
+
+      const face = this.faces()[this.get('activeFace')];
+      if(face !== undefined) {
         this.applyDeltaToDOM(face);
+        this.previouslyActiveFace = face;
+      }
     }
     if(delta.text !== undefined)
       setText(this.domElement, this.get('text'));
--- a/client/js/widgets/basicwidget.js
+++ b/client/js/widgets/basicwidget.js
@@ -24,7 +24,7 @@ class BasicWidget extends Widget {
     if(delta.activeFace !== undefined || delta.faces !== undefined) {
       let face = this.faces()[this.get('activeFace')];
       if(face !== undefined)
-        this.applyDelta(face);
+        this.applyDeltaToDOM(face);
     }
     if(delta.text !== undefined)
       setText(this.domElement, this.get('text'));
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1665/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1665-pr-test`.

client/js/widgets/basicwidget.js
```diff
@@ -22,12 +22,22 @@ class BasicWidget extends Widget {
...
     super.applyDeltaToDOM(delta);
     if(delta.activeFace !== undefined || delta.faces !== undefined) {
-      let face = this.faces()[this.get('activeFace')];
-      if(face !== undefined)
-        this.applyDelta(face);
+      if(this.previouslyActiveFace !== undefined) {
+        const undoDelta = {};
+        for(const property in this.previouslyActiveFace)
+          undoDelta[property] = this.state[property] !== undefined ? this.state[property] : null;
+        this.applyDeltaToDOM(undoDelta);
```
robartsd, line 29: I prefer the pattern that I used in dice to eliminate recursive calls to applyDeltaToDOM().

ArnoldSmith86: Yeah, makes sense. Feel free to change it or I'll do it later.

robartsd: The recursion is a problem because the undoDelta conceivably could contain activeFace and/or faces properties causing an infinite loop. The alternative approach would be to simply ensure that the undoDelta does not have either of these properties.

ArnoldSmith86: I decided to do the less invasive approach. Feel free to change it to your pattern but this took less time for me to implement.

ArnoldSmith86:
I'm pretty sure this is ready to go. Can somebody else throw a few tests at this and check that our games don't break?

96LawDawg:
Just to make sure I understand the expected behavior.

```
"width": 100,
"activeFace": 1,
  "faces": [
    {
      "text": " Face 1",
      "width": 200,
      "testing": 1
    },
    {
      "text": " Face 2",
      "width": 300,
      "testing": 2
    }
  ],
  "text": "Basic widget",
  "testing": 5
```

This shows what you see in the JSON: the set values on the widget are different than what is in the active face.  What is displayed in the room is what is on the active face.  But if I GET the properties of text, width, and testing, I get what is displayed in the room (from the active face) and NOT what is in the JSON.

I think that is what is supposed to happen.  Please confirm.  But it is very confusing to see something on the widget that is not in the room and that you can't GET.

96LawDawg (approved):
I tested a variety of things directly.  I put different properties into the faces, clicked the widgets, watched what happened in the room and in the JSON.  I also used GET as described above.  Everything worked fine.  I tested several games that use faces on basic widgets, including Dix Dice which also sorts them based on a value set from the face.  All of that worked fine.  Approved through test.

PR #1674, opened 2023-03-06
Fix coordinate rounding

96LawDawg:
#1641 created a problem where widget movement is made in increments with precision out to 12 decimal places.  This behavior is putting my OCD into overdrive.  This PR rounds those values to integers like it used to be and should be.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1674/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1674-pr-test`.

flackr:
This seems fine since it's only done in `move`, so `moveEnd` can still convert to something more precise to match the exact location when `"fixedParent": "false"`. Note that this will effectively snap things in a "fixedParent" scaled holder with a high enough scale value, e.g. drag the piece in https://test.virtualtabletop.io/PR-1674/scale-test but we can revisit this later if we ever need the added precision.

client/js/widgets/widget.js
```diff
@@ -1967,6 +1967,8 @@ export class Widget extends StateManaged {
     if (limit.maxY !== undefined) {
       newCoord.y = Math.min(limit.maxY, newCoord.y);
     }
+    newCoord.x = Math.round(newCoord.x);
+    newCoord.y = Math.round(newCoord.y);
```
flackr, line 1971: You should do the rounding in the dragCorner method so that it's used for all 3 locations which determine drag location.

96LawDawg: Is the commit I just made what you are talking about?  The only change I noticed in testing is that now the CSS Transform tutorial locations on the table are integers.  The scale test and simple cases seem the same.

RaphaelAlvez:
Did not have a chance to test it yet but I want to be sure this does not break stackOffsetX : 0.001

96LawDawg:
> > Note that this will effectively snap things in a "fixedParent" scaled holder with a high enough scale value, e.g. drag the piece in https://test.virtualtabletop.io/PR-1674/scale-test but we can revisit this later if we ever need the added precision.
> 
> Could add a property to the parent that sets the precision for locations of children dropped into it.

That certainly gives the needed flexibility.  Or we could change the rounding to be conditional.  If the parent has a scale > 2 or 3 then don't round.  Definitely dealing with edge cases here because scaling up is unusual (usually scale something down to make it fit on screen).

client/js/widgets/widget.js
```diff
@@ -624,7 +624,11 @@ export class Widget extends StateManaged {
...
     let positionCoord = this.coordParentFromCoordLocal(transformOrigin);
     const offset = getOffset(this.coordParentFromCoordLocal(localAnchor), positionCoord);
-    let corner = {x: coord.x + offset.x - transformOrigin.x, y: coord.y + offset.y - transformOrigin.y, z: this.get('z')};
+    let corner = {
+      x: Math.round(coord.x + offset.x - transformOrigin.x),
+      y: Math.round(coord.y + offset.y - transformOrigin.y),
+      z: this.get('z')
+  };
     return parent ?
         parent.coordGlobalFromCoordLocal(corner) : corner;
```
flackr, line 633: You should probably round after the conversion to global coordinates. I.e.
```js
  if (parent)
    corner = parent.coordGlobalFromCoordLocal(corner);
  corner.x = Math.round(corner.x);
  corner.y = Math.round(corner.y);
  return corner;
```

This will ensure that we are consistent in rounding the global coordinates (for non-fixed parent widgets) rather than rounding local coordinates and then converting to global coordinates.

96LawDawg: Hopefully this is what you meant.

client/js/widgets/widget.js
```diff
@@ -625,8 +625,11 @@ export class Widget extends StateManaged {
     let positionCoord = this.coordParentFromCoordLocal(transformOrigin);
     const offset = getOffset(this.coordParentFromCoordLocal(localAnchor), positionCoord);
     let corner = {x: coord.x + offset.x - transformOrigin.x, y: coord.y + offset.y - transformOrigin.y, z: this.get('z')};
-    return parent ?
-        parent.coordGlobalFromCoordLocal(corner) : corner;
+    if (parent)
+      corner = parent.coordGlobalFromCoordLocal(corner);
+      corner.x = Math.round(corner.x);
```
flackr, line 630: Unindent this and line 631 as they're not part of the if.

flackr (approved):

96LawDawg:
We've got @flackr approval on this.  Anybody else.  Small bug, but the lack of rounding is driving me crazy.

ArnoldSmith86 (approved):
Code looks fine to me. And dragging seems to still be working.

PR #1681, opened 2023-03-07
Initial test of Roboto-Symbols font

mousewax:
This is an initial pull request to test the combining of the VTT Symbols font with Roboto. 
image

You can use any of the VTT Symbols inside of square brackets.

This test only includes one font weight of Roboto. Bolded and italicized text will not work.

All symbols are the same width except the individual cards are skinnier.
I would like feedback for sizing and positioning of the symbols.


96LawDawg:
As mentioned on Discord, I think the sizing and positioning as it is now is fine.  The poker examples you put in there look strange on the top line with `"line-height": "0.8em"`.  I thought that was strange, but it turns out that is just a "feature" of labels.  Doesn't happen in a basic widget.

96LawDawg:
Also, @mousewax, when you have a minute, could you briefly describe what you did to get this to work?  Would be good to have it documented.

robartsd:
Does it work in Safari? IIRC Safari was the hang up when you were trying to do this before (I found it ironic that Apple's browser was having difficulty implementing a font technology that other browsers were handling just fine).

mousewax:
This should work in Safari. The hangup before was trying to specify two fonts so if one font didn't have a glyph it would fall back to one that did. This is a fix where one font has the glyphs.

96LawDawg:
Does NOT work in Safari on my newish iPad.

mousewax:
> Does NOT work in Safari on my newish iPad.

I will check that out and then likely abandon this project.

robartsd:
We could have the engine replace `[some text]` in text property with `<span class="icon">[some text]</span>`.

mousewax:
Asking for that is my backup to my backup. I am trying to reduce the amount of extra parts that would have to be added (if possible).

96LawDawg:
@mousewax, do you think we should abandon this PR now that we can use symbols in html?  Admittedly, that only works on basic widgets, but if I were doing something on a button, I might use a basic widget as a cover to put on top of the button.

You had high hopes for this, but ran into so many obstacles.

PR #1683, opened 2023-03-11
Add clubs, hearts, diamonds, spades

96LawDawg:
These basic symbols are taken from the existing card_clubs, card_hearts, etc. symbols.  This removes the card border and makes the images stand-alone.  They are added as image assets and as VTT Symbols.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1683/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1683-pr-test`.

RaphaelAlvez:
What did you change in the font files? It's hard to know from the files changes

Appart from that I'm ok with this PR

96LawDawg:
I just added the 4 card icons.  Nothing else.  Same thing I've done when other new symbols have been added.

mousewax:
image

96LawDawg (commented):
Well, I started this, but mousemax has completely changed it for the better.  So I would approve it if I could.  But I can't since I started it.  So that's one approval.  Looking for a second so this simple change can be merged.

ArnoldSmith86 (changes requested):
This seems to remove the new icons:

image

96LawDawg:
> This seems to remove the new icons:
> 
> image

Turns out the icons are all fine.  But the switch from .woff to .ttf changed the spacing very slightly.  So the fonts didn't line up with the text.  If you look at the pic, those last 2 icons are the "missing" ones.  So I changed the spacing to make it work and I added scoreboard and play_pause where I could squeeze them in.  They had never been added.

So this PR is ready to merge.

ArnoldSmith86 (approved):
Gonna approve this assuming that @mousewax wants to keep the font as ttf.

I didn't check the content of the font file but the assets room seems to work on all browsers I can test and that I care about.

96LawDawg:
@ArnoldSmith86, should I add the star you are using in #1482 to the font and assets?  Makes sense for consistency because I think the suit should go from "/assets/1060956615_617" to "i/icons/star.svg" and be like the other ones will when this PR is merged.

Is it https://game-icons.net/1x1/delapouite/round-star.html?

ArnoldSmith86:
No, that was just a test. There are dozens of icons usable as an extra suit.

96LawDawg:
I'm going to merge this in a couple of days unless there is an objection.

ArnoldSmith86:
I would prefer if the outlined ones would match the others but it's fine if they don't.

96LawDawg:
Okay.  I'll hold off for a bit.  I'll see if reattacking it after letting it sit for a few days helps me figure out what I was doing wrong.  Or maybe @mousewax can solve it in 34 seconds.

96LawDawg:
> I would prefer if the outlined ones would match the others but it's fine if they don't.

Fixed.  Letting it sit and reattacking the problem fresh was the solution for me.  This is what they look like now.  Assuming no further objections, I will merge in a couple of days.

Suits

ArnoldSmith86 (approved):
Tested in Firefox and Chromium on Linux and Chromium and Privacy Browser on Android.

PR #1687, opened 2023-03-15
fixed that cloning displays certain children in the wrong position

### LINKED ISSUE #1686: cloning displays children in wrong position
If you have a holder at 0,0 in a basic widget at not 0,0 and then recursively clone the basic widget using the JSON editor, the holder will be displayed at absolute coordinates 400,400 until you reload the room.

https://virtualtabletop.io/2ikn

```JSON
{
  "41m9": {
    "type": "holder",
    "id": "41m9",
    "parent": "q9ss"
  },
  "q9ss": {
    "id": "q9ss",
    "x": 200,
    "y": 200
  }
}
```

ArnoldSmith86:
This fixes #1686. When adding new widgets locally (through cloning or the Add Widget overlay), it receives its initial state twice - once because it is created locally first so following routine operations can already interact with it and then it receives the normal delta with the same properties again.

The new code, that makes it possible to apply animation to parent changes, sets temporary transforms to the widget when its parent property changes. But because this would trigger all sorts of animations when loading the room, it only does this when the widget is not a newly created widget.

This is done in two areas in the code (where widgets are unparented to avoid cyclic parents and where widgets are reparented to their actual parent. They both follow the same logic but one path used `widgets.has(id)` to check if it is a new widget and the other path used `delta.id !== undefined` to check if it is a new widget.

Both checks work fine unless there is an `id` in a delta after the widget was already added to `widgets` which is exactly the case when locally adding a new widget.

Now both code paths check for `delta.id === undefined` so they both execute at the same times, undoing each other's changes.

### DIFF
--- a/client/js/serverstate.js
+++ b/client/js/serverstate.js
@@ -108,7 +108,7 @@ export function batchEnd() {
 function receiveDelta(delta) {
   // the order of widget changes is not necessarily correct and in order to avoid cyclic children, this first moves affected widgets to the top level
   for(const widgetID in delta.s) {
-    if(delta.s[widgetID] && delta.s[widgetID].parent !== undefined && widgets.has(widgetID)) {
+    if(delta.s[widgetID] && delta.s[widgetID].parent !== undefined && delta.s[widgetID].id === undefined) {
       const domElement = widgets.get(widgetID).domElement;
       const topTransform = getElementTransformRelativeTo(domElement, $('#topSurface')) || 'none';
       $('#topSurface').appendChild(domElement);


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1687/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1687-pr-test`.

96LawDawg (approved):
Approved through test.  I only tested the actual sample problem.  Did not test anything else.

client/js/serverstate.js
```diff
@@ -108,7 +108,7 @@ export function batchEnd() {
 function receiveDelta(delta) {
   // the order of widget changes is not necessarily correct and in order to avoid cyclic children, this first moves affected widgets to the top level
   for(const widgetID in delta.s) {
-    if(delta.s[widgetID] && delta.s[widgetID].parent !== undefined && widgets.has(widgetID)) {
+    if(delta.s[widgetID] && delta.s[widgetID].parent !== undefined && delta.s[widgetID].id === undefined) {
```
robartsd, line 111: I think I understand why this fixes the stated problem, but I'm not sure I understand what implications this change could have on other situations. I need to better understand how this code relates to the code creating new widgets.

ArnoldSmith86: In general `applyDeltaToDOM` is called with a first delta that contains all the defaults, then the initial delta as in the JSON and then it only gets _actual_ deltas (so it shouldn't receive a value that's already set). So normally only one delta ever includes `id` - the first one. So flackr used that to detect if it's the first delta so that there is no animation from 0,0 to the actual position.

The code in `receiveDelta` moves all changed widgets to the top surface in case A->B turns into B->A which would otherwise temporarily turn into a cyclic dependency/parentage. In the same block flackr saved the `transform` from the room origin so that it will be applied when adding the widget back to the new parent so it doesn't animate from the room origin (I think)

Does that help?

robartsd (approved):

PR #1688, opened 2023-03-16
added 5 key keyboard input

ArnoldSmith86:
This PR intends to make most games playable by using arrow keys and Enter.

Press three of those keys simultaneously to toggle this keyboard input mode.

All interactable widgets will get a colored outline. 

- Red is the left arrow.
- Yellow is the up arrow.
- Purple is Enter.
- Blue is the down arrow.
- Green is the right arrow.

Press the button in that color to filter the widgets and eventually select one widget.

- If the widget is not movable, it will be clicked immediately.
- If the widget is movable, you can now select a valid drop target holder using the same technique.
- If it is movable and clickable, the widget itself will be highlighted as a valid drop target which will then click the widget.

Current limitations:
- ~~You can not move widgets freely (only move to holders).~~ It only works if there are no valid drop targets.
- ~~You can not interact with the top card of a pile (only the whole pile).~~
- (You can not interact with input popups.) You can cancel or confirm them now, but not change any options.
- You can not interact with the rest of the UI (like changing the game).

### DIFF
--- a/client/css/keyboardmode.css
+++ b/client/css/keyboardmode.css
@@ -1,0 +1,19 @@
+.widget.keyboardHighlight.ArrowUp {
+  box-shadow: 0px 0px 0px 5px red;
+}
+
+.widget.keyboardHighlight.ArrowRight {
+  box-shadow: 0px 0px 0px 5px yellow;
+}
+
+.widget.keyboardHighlight.ArrowDown {
+  box-shadow: 0px 0px 0px 5px blue;
+}
+
+.widget.keyboardHighlight.ArrowLeft {
+  box-shadow: 0px 0px 0px 5px green;
+}
+
+.widget.keyboardHighlight.Enter {
+  box-shadow: 0px 0px 0px 5px purple;
+}
--- a/client/css/keyboardmode.css
+++ b/client/css/keyboardmode.css
@@ -17,3 +17,11 @@
 .widget.keyboardHighlight.Enter {
   box-shadow: 0px 0px 0px 5px purple;
 }
+
+body.keyboardMode #buttonInputGo {
+  background: yellow;
+  color: black;
+}
+body.keyboardMode #buttonInputCancel {
+  background: green;
+}
--- a/client/css/keyboardmode.css
+++ b/client/css/keyboardmode.css
@@ -1,21 +1,25 @@
 .widget.keyboardHighlight.ArrowUp {
-  box-shadow: 0px 0px 0px 5px red;
+  --highlightColor: red;
 }
 
 .widget.keyboardHighlight.ArrowRight {
-  box-shadow: 0px 0px 0px 5px yellow;
+  --highlightColor: yellow;
 }
 
 .widget.keyboardHighlight.ArrowDown {
-  box-shadow: 0px 0px 0px 5px blue;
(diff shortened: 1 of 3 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1688/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1688-pr-test`.

RaphaelAlvez:
At first glance this looks to be pretty close to what was asked.
If this proves to be useful then I think it should be merged into main.

It does not seem to break or be a game changer. I think all games would still be playable without a keyboard.

ArnoldSmith86:
The current state makes quite a few games playable, I think. I'm willing to address all the limitations I listed in the PR description but I'd like to wait for feedback first.

RaphaelAlvez:
This just reminded me that some time ago we were asked about screen readers. With keyboard inputs I could see that being usable but still...

Maybe if some had an actual user in mind that could play test stuff it could be something else in accessibility.

RaphaelAlvez:
I tried playtesting it. after 5 mins I was able to do """something"""
going up and down the tree is very confusing but maybe it's just a matter of getting used to it.
Also I think that you should always require the final click to be enter. I got confused because sometimes I'm walking in the tree and acidentally click something because I don't realise it's the final 5 widgets.

Also, I would suggest starting the tree with 5 widgets type groups. them maybe still group by type for groups that have more than one in the next step.
buttons
dices and spinners
cards
basic widgets
editable labels, seats, and timers (canvas but I think that's not usable) (things that shoudl be used less often)

Another thing I realized is that there is no way to go back. so maybe last click should have the option to either cancel , click or move

RaphaelAlvez:
I just want to make it clear that my last comment was just some feedback to try to improove it in my perspective. for sure I'm not qualified to give any opinion on it.

also I was surprised by how actually playable it was

ArnoldSmith86:
> also I was surprised by how actually playable it was

Yeah. The biggest hurdle are games where you need to press a reset button first that in turn asks for options in an overlay.

Other than that it's surprisingly fast if you learn the colors for a few minutes.

robartsd:
Ideas for improvement:

- exclude offscreen widgets
- style widget differently to make it obvious that the keypress will activate the widget rather than narrow the selection
- allow keys/colors to be customized (and store the settings in local storage); I could see users using something like the numeric key pad for 9 selection keys and 8 direction keys instead of 5 selection keys and 4 direction keys, or using only arrows for narrowing selection and enter only for activating the selection to avoid accidental activation
- include the widget being acted upon in the dropTarget choices, selecting it switches to free movement mode

robartsd:
I like that the buttons for the input overlay are colored corresponding to the color assigned to the selection key; I don't like that this make the cancel button green.

robartsd:
The sort by position works fairly well for left/right, but could be greatly improved for top/middle/bottom.

robartsd:
Red/Green now also matches the navigation light coloring used on boats and planes.

96LawDawg:
I think we should close this.  It isn't an important part of our planned functionality and the work will not be archived and not lost should we need it again someday.

PR #1689, opened 2023-03-16
added chooser support to PCIO importer

ArnoldSmith86:
https://github.com/jkandell/cardgame/blob/main/ohio%20jk.pcio was posted somewhere and I noticed that it used choosers. I was aware that choosers do not work but this time I looked into it.

In PCIO the choosers work very similar to cards so the importer just changes them slightly to work like cards with multiple faces.

With many options, this is not ideal. We talked about a new function `CHOOSE` somewhere that we can use to improve this later. But for now this improves the support greatly.

Can someone test a few more examples than the linked pcio file?


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1689/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1689-pr-test`.

96LawDawg (approved):
I created a couple of games that covered all the default choosers from PCIO as well as a custom one that I made.  Everything worked fine.  Approved through test.

Not as user friendly as PCIO because you have to click the widget each time to cycle through all the options.  That is part of what #1456 started out as.  But it certainly makes VTT more compatible with PCIO.

PR #1692, opened 2023-03-17
use copy and unlink instead of rename

ArnoldSmith86:
Renaming does not work across filesystems. Copying and deleting does.


server/room.mjs
```diff
@@ -643,6 +643,11 @@ export default class Room {
     this.broadcast('mouse', { player: player.name, mouseState });
   }
 
+  moveFile(source, target) {
+    fs.copyFileSync(source, target);
```
robartsd, line 647: Could use `fs.copyFilesSync(source, target, fs.constants.COPYFILE_FICLONE)` to copy via reflink if possible falling back on normal copy, on supported filesystems this should make this perform similarly to rename.

robartsd:
I could not figure out why this uses the synchronous API rather than the promise or callback API of fs. Obviously the delete must wait for copy to succeed, but I'm not sure if any of the calls to moveFile actually need the filesystem operations to be completed before continuing.

96LawDawg:
Can I test this?  How?

ArnoldSmith86:
You can make sure that state operations still work. Creating a new state, adding or removing variants, changing the order of variants. For public library games and for normal ones.

The PR is meant to fix situations where different folders (library vs save) are on different drives but testing that isn't as crucial since the "official" servers don't use that and I think it didn't work at all anyway.

96LawDawg:
So not a good start on testing.  The room flashes.  I can't get any useful information out of the browser console other than "uncaught" because it keeps resetting and erasing.

ArnoldSmith86:
What did you do?

96LawDawg:
I clicked the link: https://test.virtualtabletop.io/PR-1692/pr-test.  Then ran screaming out of the room looking for mommy.

96LawDawg (dismissed):
Now that the pr test servers are fixed, I tried to go back and test this.  First thing I did was tried to upload a game.  Get this error every time (and nothing reported in the console):

Screenshot 2023-10-01 174645

96LawDawg (approved):
Approved through test.  I changed state, added and deleted variants, moved order of variants, edited meta data, and uploaded and downloaded regular games and PL games.  Also verified much of what I was doing by looking at the History log in the Editor.

I did discover an unrelated bug which I just created an issue for involving variants linked to PL games.  But this should be good to go.

PR #1698, opened 2023-03-19
Added String() around property in set call.

### LINKED ISSUE #1697: `SCORE` with property null deletes seat
I have no idea how this could happen, but if you use the `SCORE` function and set the "property" parameter to null, it deletes the seats specified in the "seats" parameter.

rogerl50:
Fixes #1697


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1698/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1698-pr-test`.

96LawDawg (approved):
Approved through test.

ArnoldSmith86 (approved):
Looks like a reasonable fix.

PR #1700, opened 2023-03-20
Clicking scoreboard shows better defaults in options

96LawDawg:
When you click a scoreboard currently, the overlay values default to the player in the first seat and the first round.  This PR changes that so that the default value is the playerName.

I want to change the default round to be the next round.  But I wanted to go ahead and get this started because I had so much trouble doing it.  Hopefully the round will be easier.

### DIFF
--- a/client/js/widgets/scoreboard.js
+++ b/client/js/widgets/scoreboard.js
@@ -71,6 +71,8 @@ class Scoreboard extends Widget {
       if(!players.length || !rounds.length)
         return;
 
+      const currentPlayer = players.find(player => player.text == playerName);
+
       try {
         const result = await this.showInputOverlay({
           header: this.get('editPaneTitle'),
@@ -80,7 +82,7 @@ class Scoreboard extends Widget {
               label: 'Player',
               options: players,
               variable: 'player',
-              value: players.find(player => player.text == playerName).value
+              value: currentPlayer? currentPlayer.value : undefined
             },
             {
               type: 'select',
--- a/client/js/widgets/scoreboard.js
+++ b/client/js/widgets/scoreboard.js
@@ -78,7 +78,8 @@ class Scoreboard extends Widget {
               type: 'select',
               label: 'Player',
               options: players,
-              variable: 'player'
+              variable: 'player',
+              value: players.find(player => player.text == playerName).value
             },
             {
               type: 'select',
--- a/client/js/widgets/scoreboard.js
+++ b/client/js/widgets/scoreboard.js
@@ -56,10 +56,10 @@ class Scoreboard extends Widget {
       const seats = this.getIncludedSeats();
       let players = [];
       if(Array.isArray(seats))
-        players = seats.map(function(s) { return { value: s.get('id'), text: s.get('player') || '-' }; });
+        players = seats.map(function(s) { return { value: s.get('id'), text: s.get('player') || '-', selected: s.get('player') == playerName }; });
       else { // Teams
         for (const team in seats)
-          players = players.concat(seats[team].map(function(s) { return { value: s.get('id'), text: `${s.get('player') || '-'} (${team})` } }))
+          players = players.concat(seats[team].map(function(s) { return { value: s.get('id'), text: `${s.get('player') || '-'} (${team})`, selected: s.get('player') == playerName } }))
       }
 
       let rounds = this.getRounds(seats, scoreProperty, 1).map(function(r, i) { return { text: r, value: i+1 }; });
@@ -71,8 +71,6 @@ class Scoreboard extends Widget {
       if(!players.length || !rounds.length)
         return;
 
-      const currentPlayer = players.find(player => player.text == playerName);
-
       try {
         const result = await this.showInputOverlay({
           header: this.get('editPaneTitle'),
@@ -82,7 +80,6 @@ class Scoreboard extends Widget {
               label: 'Player',
               options: players,
               variable: 'player',
-              value: currentPlayer? currentPlayer.value : undefined
             },
             {
               type: 'select',
--- a/client/js/widgets/scoreboard.js
+++ b/client/js/widgets/scoreboard.js
@@ -103,7 +103,9 @@ class Scoreboard extends Widget {
         } else
           scores = +result.score;
         await seat.set(scoreProperty, scores);
-      } catch(e) {}
+      } catch(e) {
+        console.log('The input overlay for the scoreboard failed to load.', e);
+      }
     }
   }
 
--- a/client/js/widgets/scoreboard.js
+++ b/client/js/widgets/scoreboard.js
@@ -58,15 +58,16 @@ class Scoreboard extends Widget {
       if(Array.isArray(seats))
         players = seats.map(function(s) { return { value: s.get('id'), text: s.get('player') || '-' }; });
       else { // Teams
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1700/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1700-pr-test`.

rogerl50:
> But that is worse than the default. If you have say 3 people entering scores, this works for the first person, but then because there is an entry in the next round, for the second person it increases another round and so on. What needs to happen is if there are no entries for round x in any of the seats, then it works like I tried (the default is the next/upcoming round). But if there are any values for the current round in a seat, then the value should be rounds.length-1. But I don't know how to do the check. Anybody want to help out?

I'm not sure that this is possible without lots of code. It is very similar to the problem of preventing multiple people from pressing a button.

ArnoldSmith86:
> Anybody want to help out?

Can you show examples of how the score properties in each seat look like for each scenario?

96LawDawg:
Scenario: 3 players just finished round 4.  First player, LawDawg clicks on the scoreboard and gets this:

Image1

If the other players click the scoreboard right away, they will see the same thing (except with the change that is working in this PR, they will see their own name).  But if LawDawg is fast to enter his score then the next player to click on the scoreboard sees this:

Image2

The round has advanced to the 5 as the default, but if this were working as I think it should, the round would stay 4 until all 3 players have entered a score for that round.  Like this:

Image3

So for the first player, ` value: rounds.length` works.  For the other players, it should be ` value: rounds.length-1` as the default.  The problem for me is devising the check for when you should use one or the other.

96LawDawg:
And you already fixed it before I answered.  The rounds now work.  Thank you.

So one thing my original commit changes from production is that if you are not in a seat, you can't click on the scoreboard.  That wasn't intentional, but I don't have a problem with that.  It won't change the way games are played because you have to be in a seat to make scoreboards work anyway.  I will need to tweak a few words on the tutorial.  But first I wanted to see if anybody else has a problem with that.

robartsd:
> So for the first player, ` value: rounds.length` works. For the other players, it should be ` value: rounds.length-1` as the default. The problem for me is devising the check for when you should use one or the other.

Should it just be `rounds.length - 1` unless there is already a score defined for that position for the current player?

ArnoldSmith86:
> So one thing my original commit changes from production is that if you are not in a seat, you can't click on the scoreboard.

But... why? You only changed defaults of inputs.

96LawDawg:
YOU are asking ME?  I have no idea.  You built the click functionality.

On production, load the tutorial for Scoreboard, basic variant.  Click on the top-right scoreboard (the wide one).  You aren't in a seat but you get the overlay.  Now do it in this PR.  The overlay doesn't open.

ArnoldSmith86:
I have an idea. Will check tomorrow.

client/js/widgets/scoreboard.js
```diff
@@ -58,15 +58,16 @@ class Scoreboard extends Widget {
       if(Array.isArray(seats))
         players = seats.map(function(s) { return { value: s.get('id'), text: s.get('player') || '-' }; });
       else { // Teams
-        for (const team in seats) 
+        for (const team in seats)
           players = players.concat(seats[team].map(function(s) { return { value: s.get('id'), text: `${s.get('player') || '-'} (${team})` } }))
       }
 
       let rounds = this.getRounds(seats, scoreProperty, 1).map(function(r, i) { return { text: r, value: i+1 }; });
+      const everyPlayerFilledLatestRound = !seats.map(s=>(s.get(scoreProperty) || []).length != rounds.length - 1).reduce((a,b)=>a||b);
```
robartsd, line 66: Could have an issue here if a seat has something other than an array in the scoreProperty.

robartsd:
> YOU are asking ME? I have no idea. You built the click functionality.
> 
> On production, load the tutorial for Scoreboard, basic variant. Click on the top-right scoreboard (the wide one). You aren't in a seat but you get the overlay. Now do it in this PR. The overlay doesn't open.

Is the issue that you are not in a seat or is the issue that none of the seats are occupied?

client/js/widgets/scoreboard.js
robartsd: Also, reduce function will fail if seats is an empty array (can fix by providing an initial value to reduce).

96LawDawg:
> Is the issue that you are not in a seat or is the issue that none of the seats are occupied?

All of the seats but one in the tutorial room are occupied.  So I think the issue is that the player clicking is not in a seat.

client/js/widgets/scoreboard.js
```diff
@@ -78,13 +79,15 @@ class Scoreboard extends Widget {
               type: 'select',
               label: 'Player',
               options: players,
-              variable: 'player'
+              variable: 'player',
+              value: players.find(player => player.text == playerName).value
```
robartsd, line 83: If `playerName` is not found, then `players.find()` returns `undefined`. Trying to access the `value` property of `undefined` is an error.

client/js/widgets/scoreboard.js
96LawDawg: > Could have an issue here if a seat has something other than an array in the scoreProperty.

I don't think so. If it isn't an array, then there is just a totals line and no rounds.  And the change still works when there is just a totals line.

96LawDawg:
The latest commit by @robartsd fixes the problem.  But I just found another problem.  We need a solution that works with team scoring.

client/js/widgets/scoreboard.js
96LawDawg: I tried something like
`players.find(player => player.text == playerName)?.value ?? null`.  I changed null to ' ', players[0].value, and several other things trying to fix the team problem, but none of that worked.

robartsd: It looks like `players.find(player => player.text == playerName)?.value` should return `undefined` when `playersName` is not found (equivalent of what my code does. Not sure what `showInputOverlay` does with `null` or `''`, but `undefined` should behave as if `value` was not set at all.

ArnoldSmith86:
> We need a solution that works with team scoring.

Was this addressed by the latest commit?

96LawDawg:
> > We need a solution that works with team scoring.
> 
> Was this addressed by the latest commit?

No.  Current status is this does not work with teams or when all seats are empty.

96LawDawg:
Just to be clear, there are 2 different errors.  If the seats are empty, error is "Reduce of empty array with no initial value." If using teams, the error is "seats.map is not a function."  Right now, the log errors line is not even triggering before the client freezes.

robartsd:
> Just to be clear, there are 2 different errors. If the seats are empty, error is "Reduce of empty array with no initial value." If using teams, the error is "seats.map is not a function." Right now, the log errors line is not even triggering before the client freezes.

Thanks for the clarification. My last commit addressed the default selection when using teams, but not problems with everyPlayerFilledLastRound. I'm thinking we should add a class method to check if a particular round has values for all seats and call that method with the last round.

rogerl50 (approved):

96LawDawg:
Going to approve this in a couple of days pending any objections.

PR #1705, opened 2023-03-23
Fix source problems in `TURN`

### LINKED ISSUE #1704: Problems with `TURN` and source
Two problems discovered by anewcar while making a game:

The source parameter of `TURN` does not accept anonymous collections.  Causes the client to freeze.  This one should be fixed but doesn't have to.  We could just document it but since most every other function accepts anonymous collections, I think this one should also.

If the source parameter is specified but the collection does not actually exist, the client freezes.  This one needs to be fixed.

96LawDawg:
Fixes #1704.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1705/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1705-pr-test`.

96LawDawg:
Compare the PR test room with https://virtualtabletop.io/ey7m.

ArnoldSmith86:
This doesn't fix the crash for an invalid collection yet, though.

96LawDawg:
> This doesn't fix the crash for an invalid collection yet, though. (And you deleted a space.)

I thought testing for ` "source": ["foo", "bar"]` would do it.  Dumb.  So the getCollection function has a problem.push line that I thought would catch this.  But it didn't.  Now I think it does.  May not be ideal coding, but it works and is similar to other stuff I've seen in VTT.

96LawDawg:
Okay.  Now I have it.

ArnoldSmith86 (approved):
Code looks fine. The test-cases in pr-test also look fine. I added one that uses an actual collection.

PR #1706, opened 2023-03-24
added `verticalHeader` property to scoreboards

ArnoldSmith86:
It is not trivial to display rotated text in the header so I added this property that makes it easy.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1706/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1706-pr-test`.

robartsd:
I see, the need is for an extra element within the table cell (I wish css had a way to create pseudo parent elements).

I probably would have called this rotatedHeader, but don't really care.

mousewax:
> I see, the need is for an extra element within the table cell (I wish css had a way to create pseudo parent elements).

You could put the contents in a `::before` and then style the element.

robartsd:
> > I see, the need is for an extra element within the table cell (I wish css had a way to create pseudo parent elements).
> 
> You could put the contents in a `::before` and then style the element.

How do you move the contents of the actual element into it's psuedo child element in css?

mousewax:
Javascript. You'd have to do it when constructing the widget. But I guess that's really no different that just adding a div. 

I withdraw my comment.

robartsd:
Of course as soon as they add `::parent`, I would find a case where I want multiple levels of psuedo parents. I already want more than 2 pseudo children; but reading stylesheets could get pretty confusing if psuedo elements could have their own pseudo elements.

robartsd:
> BTW, I only added the `div` when the property is enabled so that I don't mess with `css` that already styles those cells.

Otherwise this PR would just be css and a predefined classname (no need for a widget property).

Could probably use a span instead of div (or set the default style of the added div to behave like a span) with little risk of messing up existing css since there previously would have been no way to have an extra element wrapping the contents of the cell and replacing a text node with an unstyled span containing the text node generally does not change rendering.

96LawDawg (approved):
Approved through test.

I added an example to the tutorial.

When merged, will need to update the wiki.

PR #1707, opened 2023-03-24
fixed that scoreboard would redraw when dragging

ArnoldSmith86:
The scoreboard recreates its table on every property change. It is draggable by default so while dragging, it will recreate the entire table multiple times per second.

~This PR excludes a few common properties that change when dragging.~

~It would be more correct to only call `updateTable` when any property changes that it uses for the table but that is a pretty long list.~

This PR should now only call `updateTable` when a delta contains a property that it uses. `updateTable` still may be called multiple times in the same delta (once for each widget that has property changes that it uses).


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1707/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1707-pr-test`.

96LawDawg (approved):
Scoreboards still work, so that is good.

I tested looking at production and this PR using the Elements section of the console.  You can see the table being rebuilt as soon as you start to move on production and that doesn't happen in this PR.

I would add `hoverTarget` to the list of excluded properties.  The code seems to make sense (not the same thing as a code review, but it is what I can offer).

Approved.

robartsd:
I'm not clear on how the scoreboard registers to get notified of updates in seats.

ArnoldSmith86:
The seats have a function all the way at the bottom that calls the updateTable directly.

robartsd:
I think I've enumerated all the properties that are used in `Scoreboard.updateTable()`

- Scoreboard properties
  - 'showTotals'
  - 'scoreProperty'
  - 'sortField'
  - 'totalsLabel'
  - 'roundLabel'
  - 'showPlayerColors'
  - 'playersInColumns'
  - via `Scoreboard.getIncludedSeats()`
    - 'seats'
    - 'showAllSeats'
    - 'sortAscending'
  - via `Scoreboard.getRounds()`
    - 'rounds'
    - 'showAllRounds'
- Seat properties
  - 'player'
  - 'color'
  - _scoreProperty_
  - _sortField_

ArnoldSmith86:
Feel free to change the logic. :smile:

96LawDawg (approved):
Approved through test.  Tested the tutorials rooms and the complex setup in my Cards room.  Everything works as expected.

PR #1718, opened 2023-04-05
`inheritFrom` no longer inherits defaults

### LINKED ISSUE #1717: `inheritFrom` is inconsistent about inheriting defaults
Take this room (https://virtualtabletop.io/ls69):

```JSON
{
  "button": {
    "type": "button",
    "id": "button",
    "css": {
      "--wcMain": "#000"
    }
  },
  "basic1": {
    "id": "basic",
    "y": 100,
    "inheritFrom": "button"
  },
  "bugButton": {
    "type": "button",
    "id": "bugButton",
    "y": 200,
    "text": "Change\nmovable",
    "clickRoutine": [
      {
        "func": "SET",
        "collection": [
          "basic"
        ],
        "property": "movable",
        "value": false
      },
      {
        "func": "SET",
        "collection": [
          "basic"
        ],
        "property": "movable",
        "value": null
      }
    ]
  }
}
```

The basic widget inherits `css` from the button. `type` and `id` will never be inherited because they are blacklisted in the engine along with `deck`, `cardType` and `audio`. The CSS only sets a variable used by button CSS though so it doesn't do anything on the basic widget.

Then there is a button that simply changes `movable` on the basic widget. Changing `movable` triggers an update to the classes of the widget because it adds a class `movable`. Among other things, updating the classes of the widget calls `this.get('typeClasses')`. This call follows the `inheritFrom` and returns the `typeClasses` of the button which adds `button` to the classes which makes the widget look like a button (it still behaves like a basic widget though) and use the custom variable from `css`.

Reloading the page will reset the appearance because `movable` isn't set anymore.

So `inheritFrom` "actively" inherits only values that are explicitly set in the widget but "passively" it also inherits the defaults when anyone is asking about a specific property. This is clearly inconsistent and needs to change.

I'm pretty sure that the solution with the least problems is, that defaults are neither actively nor passively inherited.

ArnoldSmith86:
This fixes #1717. The problem was that `inheritFrom` would only inherit values explicitly set in the widget (so no defaults) _but_ it _would_ inherit the default when `get`ting a value (see issue for details). This inconsistency lead to a few problems that are hopefully fixed in this PR.

I only did rudimentary tests up until now. This needs testing for chained `inheritFrom`s and setting stuff from and to the defaults in all the links of the chain. It also should not break any games.

This PR works by saving a list of properties in the inheriting widget that were inherited before and then a `get` only gets a property from the inheriting widget if it is in that list.


96LawDawg (approved):
Approved through test.  I tested Fowl Play which I made so I know it uses a lot of chained `inheritFrom`.  It still worked fine.  I also tested your demo issue from #1717 but making several chained `inheritFrom` widgets.  I set at different places in the chain and it worked as intended in this PR.  I think this is ready.

PR #1719, opened 2023-04-05
PCIO importer improvements

### LINKED ISSUE #1417: SORT does not work when MOVEing to Seats
Sort is unreliable when used in connection with a MOVE to hand through seats.

See discussion in help channel on Discord on Nov 23 at: https://discord.com/channels/770758631146782780/793626590848352306/1044933921353252935

Example:

```
"clickRoutine": [
    {
      "func": "MOVE",
      "from": "dmge",
      "to": [
        "Seat 1",
        "Seat 2",
        "Seat 3",
        "Seat 4"
      ],
      "count": 10
    },
    {
      "func": "SORT",
      "holder": "hand",
      "key": "rankA"
    }
  ]
```

ArnoldSmith86:
My goal for this PR was to make two games compatible that @jkandell "reported" to have problems:

- https://github.com/jkandell/cardgame/blob/main/track-trick-jk-3pl-3.pcio
- https://playingcards.io/4f98fj (an updated version of https://github.com/jkandell/cardgame/blob/main/agony-aunt-jk-5.pcio)

You can look at the commit list for a list of improvements I implemented. Notably there are two changes that are not PCIO importer specific:

- I added a parameter `inHolder` to `RECALL` similar to `owned`. If `true`, only widgets that have a parent (skipping piles) will be recalled.
- I found missing `await`s in `MOVE` to a hand. This should fix #1417.


96LawDawg (changes requested):
Tested and this works.  I only did the most basic of tests on the PCIO importer as that won't break anything else and it is hard to test.

I did test the 2 new things and they both work fine.  BUT, the `inHolder` parameter needs to be added to the JSON Editor.  Once that is in place, I will approve.  I would add that myself, but I'm not at home.  Maybe I'll add it later if I think about it.

Needs a wiki update when merged to document the `inHolder` parameter.

96LawDawg (approved):
Approved

PR #1222, opened 2022-05-17
add `simulate` parameter to `TURN`

ArnoldSmith86:
This PR adds `readonly: true` to `TURN` so that you can check which seats will become active without actually changing anything.

PR #1724, opened 2023-04-14
add `simulate` parameter to `TURN`

ArnoldSmith86:
Second try to #1222.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -1890,16 +1890,12 @@ export class Widget extends StateManaged {
             }
             if(jeRoutineLogging)
               jeLoggingRoutineOperationSummary(`changed turn of seats from ${previousTurn} to ${turn} - active seats: ${JSON.stringify(indexList)}`);
-          } else {
-            if(jeRoutineLogging)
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1724/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1724-pr-test`.

96LawDawg:
This is the third of 3 things involving `TURN`.  The other two are done or nearly done.  Adding this would help overhaul `TURN` and make it as functional as possible.  I fixed a merge conflict and then found some logging stuff that was freezing the room.  Specifically, `jeLoggingRoutineOperationSummary` is always undefined for some reason when using TURN.  So I put the same logic checks on there that exist in other parts of widget.js.  Those should probably stay.  But I suspect there is some simple fix to why the logging summary is not working.  @ArnoldSmith86, can you take a look?

96LawDawg:
I have fully tested this and except as noted in my previous comment, I would approve it.

96LawDawg:
Some later changes to how turn works probably impact this.  It should still be implemented.  Seems like I need this and have to create a work around every few months.  But it needs to be tested and cleaned up, especially after #2080.

PR #1736, opened 2023-04-23
PL Update Five Of A Kind

### LINKED ISSUE #1734: Five of a Kind
Clicking an individual die rolls it independently.

96LawDawg:
Fixes #1734.

Dice are not clickable.  Title format widget changed as a result of label PR so the easiest fix was just to not make this a label.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1736/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1736-pr-test`.

ArnoldSmith86 (approved):
Looks good. FYI: the problem is using `pt` as a unit for the font-size. If you use `px`, the label will set a matching line-height.

In general, I avoid using `pt` because it behaves different on Windows and Linux. But that might have changed in the last years.

PR #1737, opened 2023-04-27
PL Add Ambagibus

96LawDawg:
An original print and play game by P.D. Magnus adapted to VTT.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1737/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1737-pr-test`.

ArnoldSmith86:
The images in the rule description are links to the outside. Games in the public library should not have external dependencies. Are we allowed to use them? I can turn them into assets if we are.

96LawDawg:
> * The images in the rule description are links to the outside. Games in the public library should not have external dependencies. Are we allowed to use them? I can turn them into assets if we are.

I read the CC copyright to cover the game and the rules.  So I would say we are allowed to use the rules verbatim with their pictures.  I don't know how to convert all that to assets.

> * It's pretty weird to have different fonts in the description.

I agree.  I just copied and pasted the multi-player rules at the bottom of what garrolousHyena put in there for the solo rules.  There is a bunch of html bloat in there we don't need.  I played with it a few minutes and don't know how to get rid of it.


> * Why does the recall take so long? It's just 160 cards. Does it do anything special?

The holders do something special to every card (moving it to the bottom).  That was being implemented on the recall.  I figured out a work around.  Recall is normal and now the holder still dump a dropped card to the bottom.

> * If I move a tile back onto the draw pile, all cards in the pile get flipped.

Fixed.

> * What's the rule if the drawn card is impossible to play?

I don't know.  I would guess you move it to the bottom.

PR #1738, opened 2023-04-28
only set inheritChildZ for pile holders

ArnoldSmith86:
For compatibility with PCIO, I introduced the property `inheritChildZ` a long time ago. It makes holders inherit the z value of its cards which makes it possible to have buttons on top of the holder that get hidden when a card is in it. I used to do that on PCIO all the time so I needed a way to make this work in VTT where the button can not be between the holder and the card because the card is an actual DOM child of the holder.

In the meantime, PCIO added support for spread and freeform holders where a button (or something else) is not necessarily hidden by cards that get dropped into it. This PR sets `inheritChildZ` only for holders that hold a pile of cards.

This is not perfect but should be the better choice.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1738/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1738-pr-test`.

96LawDawg (approved):
Approved.  This fixes the identified problem in the specific PCIO file.  No idea what else it might do, but it won't break anything.

PR #1742, opened 2023-05-01
Public Library: Add Safari Scramble (similar to Flinke Pinke)

andypymont:


96LawDawg (approved):
Everything seems to work.  Meta data looks good.  I would suggest consider adding
`"radius": 8,"border": 1` to both decks. I think the consensus is that rounded corners look better.  But that is completely optional.

PR #1745, opened 2023-05-06
Compress File Size in 19 Public Library Games

96LawDawg:
The only changes are to large image sizes.  In Dungeon Speeders, I also replaced an old style dice that used 24 faces with the new widget form that does that same thing but in less code.

Any other changes are the result of file updaters.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1745/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1745-pr-test`.

96LawDawg:
Closing this.  If I try something like this again, it will be with the new asset compressor tool and not so many games.

PR #1756, opened 2023-05-18
fixed importing cardTypes from CSV

ArnoldSmith86:
When adding new card types in the CSV, the client would crash when adding the cards because after adding the cardType, the editor would only add the change to the delta and not to the actual widget.

This PR changes the editor behavior so it always triggers a `set` on any updated property. This also means that editor changes will now trigger `changeRoutine`s etc.

Do you think this is a good thing or a bad thing? This could also be fixed in a way that does not trigger routines.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1756/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1756-pr-test`.

96LawDawg:
> This also means that editor changes will now trigger `changeRoutine`s etc.
> 
> Do you think this is a good thing or a bad thing?

That does have the potential to cause some issues.  I've seen (and probably used myself) `changeRoutine`s that make pretty significant changes, even adding or re-writing properties.  By the end of my game creation, I will have a reset button that undoes all of that.  But during game creation, I might not and I would not expect (and will not remember) that those routines can be activated just by using the editor to add some cards.

I'm not complaining.  This is doable.  Just saying what I think the potential issue might be.

robartsd:
I would also prefer if editor changes never trigger routines.

96LawDawg (changes requested):
Import function works fine.  But `globalUpdateRoutine` is still triggered when adding cards.  Didn't test any other kind of routine because it was going to take some thought to figure out how to get one to trigger.

ArnoldSmith86:
Hmm, is it also triggered when using the normal "add 1 card" button?

96LawDawg:
> Hmm, is it also triggered when using the normal "add 1 card" button?

Yes.  So I guess that part really is not a bug.  Just something to deal with?

96LawDawg:
The `globalUpdateRoutine` was actually triggered 4 times when clicking `"add 1 card" button.  I haven't tried to track down the specific triggers, but probably the creation of the card itself, changing the `cardDefaults` of the deck widget, making the widget a child of the holder, and some other stupid thing.

When I tried to add all 240 cards from Tavo!'s deck, the room froze.  240 x 4 was too much I guess.

ArnoldSmith86:
> The `globalUpdateRoutine` was actually triggered 4 times when clicking "add 1 card" button. I haven't tried to track down the specific triggers, but probably the creation of the card itself, changing the `cardDefaults` of the deck widget, making the widget a child of the holder, and some other stupid thing.

Routines shouldn't be triggered at all unless the deck is in a holder. The routine calls come from this code:

```Javascript
          if(jeStateNow.parent)
            await widgets.get(cardId).moveToHolder(widgets.get(jeStateNow.parent));
```

If the deck is in a holder, the new card gets moved to that holder using `moveToHolder` (basically `MOVE`). I don't know, my gut tells me that's fine.

96LawDawg (approved):
Approved through test.  The routine is not triggered unless the deck is in a holder.  I agree with @ArnoldSmith86 that should be okay.  Although the usual case is to have the deck in a holder and you could freeze up the room like I did if not careful.  Document and move on.

PR #1758, opened 2023-05-23
Added ability to specify tabindex and placeholder text for labels

mousewax:
Added ability to specify `tabindex` on labels to control navigation between labels when they are editable. Set this with:
`"tabIndex": 𝑥`.

Similarly:
`"placeholderText": "𝑦"`

Thanks to LawDawg for explaining how things get added to the JSON editor.

### DIFF
--- a/client/js/widgets/label.js
+++ b/client/js/widgets/label.js
@@ -47,13 +47,17 @@ export class Label extends Widget {
     }
 
     if(delta.editable !== undefined) {
-      if(delta.editable)
+      if(delta.editable) {
         this.input.removeAttribute("readonly");
-      else
+        this.input.setAttribute('tabindex', this.get('tabindex'));
+      } else {
         this.input.setAttribute("readonly", true);
+        this.input.removeAttribute("tabindex");
+      }
     }
     if(delta.spellCheck !== undefined) {
-      this.input.setAttribute('spellcheck', this.get('spellCheck') === true)
+      this.input.setAttribute('spellcheck', this.get('spellCheck') === true);
     }
+    
   }
 }
--- a/client/js/widgets/label.js
+++ b/client/js/widgets/label.js
@@ -51,12 +51,12 @@ export class Label extends Widget {
     if(delta.placeholderText !== undefined) {
       this.input.setAttribute('placeholder', this.get('placeholderText'));
     }
+    if(delta.tabIndex !== undefined) {
+      this.input.setAttribute('tabindex', this.get('tabIndex'));
+    }
     if(delta.editable !== undefined) {
       if(delta.editable) {
         this.input.removeAttribute("readonly");
-        if(delta.tabIndex !== undefined) {
-          this.input.setAttribute('tabindex', this.get('tabIndex'));
-        }
       } else {
         this.input.setAttribute("readonly", true);
         this.input.removeAttribute("tabindex");
--- a/client/js/widgets/label.js
+++ b/client/js/widgets/label.js
@@ -12,6 +12,7 @@ export class Label extends Widget {
       typeClasses: 'widget label',
       clickable: false,
       spellCheck: false,
+      tabIndex: false,
 
       text: '',
       editable: false,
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1758/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1758-pr-test`.

96LawDawg:
> I do not know how to add it to the JSON editor.

It is done automatically when you make it a widget property (which would be `tabIndex`).  I did that.  That caused problems for some reason, so I put a conditional to check if it was undefined and that fixed it.

If you change the properties in your table from `tabindex` to `tabIndex`, everything should work.

96LawDawg:
The placeholderText is working as you would expect on the internet, but not in VTT.  After changing the placeholderText, it should also set the "text" property.  Otherwise, designers won't be able to detect what text there is in there.

96LawDawg:
> The placeholderText is working as you would expect on the internet, but not in VTT. After changing the placeholderText, it should also set the "text" property. Otherwise, designers won't be able to detect what text there is in there.

Disregard.  It works.  You just have to click out of the widget in the JSON Editor and come back.  That is a minor editor issue and nothing to worry about IMO.

96LawDawg (approved):

client/js/widgets/label.js
```diff
@@ -46,14 +48,23 @@ export class Label extends Widget {
...
         this.input.removeAttribute("readonly");
-      else
+        if(delta.tabIndex !== undefined) {
+          this.input.setAttribute('tabindex', this.get('tabIndex'));
+        }
+      } else {
         this.input.setAttribute("readonly", true);
+        this.input.removeAttribute("tabindex");
+      }
     }
```
ArnoldSmith86, line 64: You only update `tabindex` if the delta _also_ contains `editable`.

ArnoldSmith86 (changes requested):

client/js/widgets/label.js
mousewax: Isn't that what this does already?
if delta.editable:
   Remove readonly.
   then, If tabindex is also set add it.
   
Otherwise:
   Make it readonly
   and remove the tabindex.

mousewax: It only adds the tab index if delta.editable.

ArnoldSmith86: I'm saying that your code doesn't change anything if the widget is already `editable` and then you set `tabIndex`. Your code will only execute if you change `editable`; not if you only change `tabIndex`.

mousewax: Gotcha. Submitted a fix. Now if tabindex is set it adds it, then relies on the editable check to remove it if it's not editable, otherwise it leaves it alone.

ArnoldSmith86: Is it a problem to have a `tabindex` when it is `readonly`? I would guess that the browser only focuses editable textareas anyway.

96LawDawg: > Is it a problem to have a `tabindex` when it is `readonly`? I would guess that the browser only focuses editable textareas anyway.

My browser puts the focus on every text area and maybe some other stuff.  So every label in the room gets the focus as I press TAB.  The advantage here I guess is that once on a widget with `tabIndex` it goes through them in order.

Tested the new changes and everything still works.

mousewax: I would say that you really only want keyboard tab navigation to elements that are not readonly as focusing on them isn't terribly worthwhile since you cannot interact with them so it would only serve to add extra tab steps getting to the stuff you can actually interact with via the keyboard.

mousewax: Which may be a good argument for the default being `tabindex="-1"` for all readonly elements.

mousewax: I suppose we can try it and if there's any feedback on it we can always re-visit it. I don't think this behavior will be game-breaking in any way and reversing it shouldn't break any future games that use `tabIndex`.

ArnoldSmith86: I wanna tweak a thing or two tomorrow.

ArnoldSmith86 (approved):
Does this still work the way you want after my changes?

My intent was to make non-editable labels have a tabindex of -1 and editable labels only set their tabindex if the game specifies one so it falls back to browser default behavior.

PR #881, opened 2021-11-11
Implement anonymous collections.

rogerl50:
This is a replacement for #879, which was closed. Suggestions/review comments from that PR have been implemented.

Needs:
1) Good code review
2) Thorough testing of new array functionality for source collections.
3) Is there a way to add [] to the list of possible values for the `"collection"` property (or, for SELECT or TURN, the `source` property? I don't see how it can be done.

Addresses #875. With this PR, functions that accept an input collection can now accept an array
```
  "collection": ["id1", "id2", "id3"]
```
The result is that the function will operate on an anonymously-named collection that consists of all the id's that actually are the id's of widgets.

Replaces `isValidCollection` by a new routine, `getCollection`, that returns either the key of the collection passed in, or the key of a new collection created from the array that is passed in, or `null` in case of error. The resulting collection key is used in subsequent function processing.

Passes the TestCafe tests, so this may not have broken anything. Tested only in very simple cases for the actual use of anonymous collections.

PR #1762, opened 2023-06-07
fixed TIMER when using a collection

ArnoldSmith86:
`TIMER` with `mode: set`, `mode: dec` and `mode: inc` was not working with collections since #881 because the variable `collection` would not be filled in that code path.

I noticed this because it broke the timer in `The Team II`.


96LawDawg (approved):
Approved through test.  Can use anonymous collections with those modes in the PR test room.

PR #1763, opened 2023-06-09
support transparent canvas

ArnoldSmith86:
With this PR, you can use `transparent` in the canvas `colorMap` property to let the widget underneath the canvas shine through.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1763/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1763-pr-test`.

96LawDawg (approved):
Approved through test.  Existing games and tutorials work fine.  The transparent aspect works just as you expect it would. It is easy to add the transparent option, but I thought there might have been a way to include that in the JSON Editor.  I thought we had different pre-set color palettes and this could be added to those, but I guess those were only in the original draft and didn't make it into production (or at least, I can't find them in the code).

PR #1771, opened 2023-06-13
new game: Word Whirl (similar to So Clover!)

ArnoldSmith86:
Mainly built by mousewax.


96LawDawg (approved):
It completely does not matter, but the sun and snow emoji in the meta data under paragraph 1 of Write Clues are useless.  I personally wouldn't have those emojis.  But that's just me.  Approved through test.

Issue #94, opened 2021-01-12
Display the player name next to the cursor when actively moving a widget


GroggyLion:
Possibly enabled via a global true/false setting per game. 

This would be useful for all our elderly players who are not concentrating on the play, and they keep asking who is moving that card! So this floating label/tooltip would only be displayed when moving a card/widget from the hand to/from the board (or maybe also at other times).

ArnoldSmith86:
Sounds like a good idea. A long time ago we were talking about showing a playercolored halo around the dragged widget. Maybe the player name could be centered above the widget.

I will build it in the near future. We can then play around with it a bit and decide if we need a toggle for it or not.

robartsd:
I think that the DOM element representing other players' cursors should have these classes dynamically assigned:
- moving (automatically removed by client 100ms after last update to the cursor from server)
- dragging (persists until delta from server indicates that the objects were released)
- foreign (added by client when owners[] provided by server is not empty and the current player is not in the list)

These classes could then be styled.

Default styles could include:
- CSS transition to fade out cursors that have stopped moving.
- CSS pseudo element showing player name while dragging non-foreign objects.

> Sounds like a good idea. A long time ago we were talking about showing a playercolored halo around the dragged widget. Maybe the player name could be centered above the widget.
> 
> I will build it in the near future. We can then play around with it a bit and decide if we need a toggle for it or not.

Does the client code for the widget already know which other player is moving an item? It seems to me that this would be easier to implement on the cursor than on the widget.

robartsd:
Classes added in #808 , css variables to be added in #972 , games will be able to set this style with #951 using `,` hack.

I do think it would be nice to add per-client cursor display options to Players & Settings overlay. Not sure what plans @mousewax has there. I picture options for styling each of 3 states (inactive, moving, pressed). Possibly with separate opacity options for the indicator circle and a text label by making each a pseudo element.

PR #972, opened 2021-12-06
add #playerCursors overlay

robartsd:
- groups player cursors into separate div within #roomArea
- adds --playerName and --playerColor css variables to cursors

PR #1774, opened 2023-06-16
add playername above cursor

Ingegneus:
fixing those citadels bugs put me down the VTT rabithole again... procrastination here I come.
this is meant as a tiny contribution to get me started on the internal workings of VTT. my goal is to add the playername above the cursor when a widget is dragged/clicked like in #94


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1774/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1774-pr-test`.

Ingegneus:
I wanted to play around with the cursor css to add the playerName above the cursor, but I can't make the text dynamically adjustable. I used:

`
.cursor::before {
  content: var(--playerName); <--- this doesn't seem to be working
  position: absolute;
  top: calc(var(--cursorSize) * -1px * var(--scale) - 20px);
  left: calc(var(--cursorSize) * -1px * var(--scale));
  font-size: 10px;
  color: var(--playerColor);
}`

96LawDawg:
I don't see the text at all in the test room.

Ingegneus:
now you should be able to see "test"

ArnoldSmith86:
The quotes here seem wrong:   playerCursors[playerName].style = `--playerName:''${playerName}'';--playerColor:${playerColor};`;

That's in client/js/overlays/players.js. Looks like two single quotes instead of a double quote. See if changing that helps.

Ingegneus:
yep that fixed it. thx!

96LawDawg:
Okay, I see it now.  I think that will be fine.  BUT, you have to make this new capability optional.  Based on prior feedback and discussion, most players and games don't want or need the additional distraction of the playerName.

Ingegneus:
apparently i went in guns blazing and didn't think for even a sec about it all...
this seems way harder then. wasn't there an ideal about a meta widget/object? or per client settings?

ArnoldSmith86:
There were ideas, yes. 😁

How do we want this to work? Opt-in for players? Opt-out for players? Opt-in for game designers? Opt-out for game designers? Either way, where and how should it be set?

If it should be opt-in for game designers, we basically don't have to do anything except the Javascript fix.

RaphaelAlvez:
I'd say it's opt in

96LawDawg:
Opt in.  100%.  To be consistent with what we've done in the past and the easiest solution is just to have the designer opt in.  But in a perfect world, players could customize it at their level. That's a lot more work. I would be satisfied with just designer opt in. Just make it a boolean property or something.

RaphaelAlvez:
I want to elaborate but I REALLY don't want this to become more work and more stress

I personally would like this to be an option for initials or tag per player. Each player would have a dropdown with "none", "tag" or "name". But in order to be fair it should work like whatsapp do for visualization. Your option should also apply to other players. So if you have "none" selected, no one will se your name but you wouldn't be able to see other players' name.
Maybe one option should also be given so you don't see other players name but they can still see yours

on top of that I think the player name should be centralized
image
 

Again consider this a one person's brainstorm about the topic

robartsd:
You can use `content: attr(data-player)` to pull in the name of the player in CSS (tested in chromium based browser). `var(--playerName)` was supposed to work with #972, but it looks like that code has an error (the value provided is ""playerName"" which CSS is not recognizing as a string. I'm not sure what you might do to create different styles for different clients.

RaphaelAlvez:
for some reason the test server keeps breaking now

ArnoldSmith86:
It was unrelated to the changes. One JSON file was missing in the library folder for the test server. No idea why.

ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1774/pr-test (or any other room on that server)

gitran:
Hasn't this been implemented in Game Settings?

96LawDawg:
Implemented by #2717

PR #1778, opened 2023-06-20
Update library images for multiple games by ArnoldSmith86, RaphaelAlvez

96LawDawg:
This commit has images only for review and comment. None of the attribution was changed until a final image is selected. All the new games will be starred in the test room so they appear at the top.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1778/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1778-pr-test`.

96LawDawg:
Most of these are awesome.  Some are maybe just meh or don't differ that much from what was there before.  Some games are really tough to implement.  Most important: we have to get The Team and The Team II right.

96LawDawg:
Just added a few new images for @RaphaelAlvez.  Also noticed that Fantasy Wars did not have similar to information, so added that and awards.

ArnoldSmith86 (approved):
I reverted the `lastUpdate` in both PRs using https://chat.openai.com/share/99a644ea-3390-4578-8012-7fa34a61f3d4.

96LawDawg:
That's pretty sweet.  So are you okay with the images for the games as well?  This set in particular.  I'm going to go ahead and merge the other one.

ArnoldSmith86:
We can always replace one or two later. It looks like a nice improvement.

PR #1784, opened 2023-07-06
no longer require a server restart after editing public library

ArnoldSmith86:
This does not affect the normal VTT servers because it requires setting `allowPublicLibraryEdits: true` in `config.json`.

I am not sure if we should change the default `nodemon` call to not auto-restart the server for changes in the library folder. While it is no longer necessary for public library game edits _from the UI_, it is still necessary for edits with external tools and I do that kinda frequently. What do you guys think?


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1784/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1784-pr-test`.

96LawDawg:
I agree.  I don't understand the technical reasons for the difference, but this doesn't happen often enough for me to worry about. Even when I was working on @mousewax's server it wasn't a big deal.

mousewax:
In order to get the ability to edit the PL from the UI I was told to run in debug mode with `npm run debug` when I had been using `npm start`. Is this no longer the case? How should I be starting things?

mousewax:
I have tested this on my local machine using `npm start` and I was able to update a PL game with no restart.

ArnoldSmith86:
In that case you may use the review button. 😀

mousewax (approved):
I was able to update the PL on my local system to update a game in the PL without restarting.

PR #1791, opened 2023-07-10
nested math expressions

ArnoldSmith86:
This allows things like `"var result = ${a} + ${b}*${c}"`. It works by allowing any number of characters `() 0-9.&|!*/+-` on the right side of a var expression (after replacing variables by their content and `true` by `1` and `false` by `0`). This means that you can not use any compute functions in this syntax but you can do a lot of simple math.

It uses `eval` on the formula which is why the set of characters is so restrictive. I hope that these characters do not allow arbitrary Javascript execution like https://jsfuck.com/.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -867,7 +867,7 @@ export class Widget extends StateManaged {
           const variable = match[1] !== undefined ? variables[unescape(match[2])] : unescape(match[2]);
           const index = match[3] !== undefined ? variables[unescape(match[4])] : unescape(match[4]);
           if(index !== undefined && (typeof variables[variable] != 'object' || variables[variable] === null))
-            problems.push(`The variable ${variable} is not an object, so indexing it doesn't work.`)
+            problems.push(`The variable ${variable} is not an object, so indexing it doesn't work.`);
           else if(index !== undefined)
             variables[variable][index] = getValue(variables[variable][index]);
           else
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1791/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1791-pr-test`.

96LawDawg:
Still testing, but I should have led with: thank you!  This is going to simplify quite a few things in the JSON.  Much appreciated.

ArnoldSmith86:
> Syntax could use a little work. No space between the original stuff that works and the new stuff but then resume normal spacing? White colors on the new stuff? Screenshot 2023-07-10 090819

Sure, I didn't work on the highlighting for that at all.

96LawDawg:
Well, I was pretty sure I got an invalid answer the first time I did this: `"var total = ${a} + ${b} + ${c} + ${d} + ${e}"`.  That is what prompted me to mess with the spacing.  But now it works fine.  So just the color I guess.

96LawDawg:
Complicated nesting seems to work fine.  Both of these executed as expected:
`"var total = ((${a} + ${b} + ${c}) * (${d} * ${e})) / (9 + 1) ** 5"`
`"var total2 = (${b} + ${c}) ** 3"`

And when you make "wrong" changes or add missing variables or try to do compute functions, the debugger gives some guidance.

What I don't understand are !, &, |.  I saw the example you but in the room, but can you explain how those would be used in an equation?

ArnoldSmith86:
If you want a complicated condition for your `IF`: `${a} && ${b} || ${c} && ${d}`.

RaphaelAlvez:
can't we create a regex that only allows the specific compute commands and not the individual characters?

96LawDawg (approved):
Approved through test.  I updated the two tutorials that are impacted by the change and will update the wiki as needed upon approval.

I would note that in-line comments will not work with any of the new nested expressions.  So `"var output = ${str1} && ${str2} //Logical AND"` still works but `"var output = ${str1} && ${str2} || ${str3} //Nested comparisons"` does not work.

rogerl50:
> Approved through test. I updated the two tutorials that are impacted by the change and will update the wiki as needed upon approval.
> 
> I would note that in-line comments will not work with any of the new nested expressions. So `"var output = ${str1} && ${str2} //Logical AND"` still works but `"var output = ${str1} && ${str2} || ${str3} //Nested comparisons"` does not work.

I think that comments need to be added. It's unreasonable to have the two very similar-looking expressions operate differently.

96LawDawg:
Change to allow comments does not work. Debug says "could not be interpreted as a valid expression..."

96LawDawg:
Nevermind.  Update to allow comments works fine.  I just hadn't downloaded the latest commit onto my server before testing.

PR #1804, opened 2023-07-28
PL Update Imagination

### LINKED ISSUE #1723: Issues with PL Game Imagination
From Zoql:

Played a few games of Imagination/Pictures this past week. 3p game went fine. However, in the 4p game, there would be a bug where, after everyone locked in a guess on a drawing, the button to advance to the next screen wouldn't show up. No clue why this happened. One time it happened before scoring, another time it happened after we only guessed two players' drawings.

Might be worth mentioning that the scoring at the end of the rounds wasn't exactly accurate. The actual number would be fine afaik, but the "Guessed By" part sometimes wouldn't work correctly.

96LawDawg:
Closes #1723. 

I was unable to duplicate the problem with the advance to next screen button.  My best guess is that near simultaneous button presses by a couple of players might have caused the routines to glitch.  This was one of the first games where I had a problem with that and I created a work around by giving people different buttons for the most part.

For the end of round "guessed by" issue, I could not duplicate that either.  It is possible that the layout caused the players to get confused, so I added white dividers on that screen to break the sections apart.  Should be a lot more clear now.

All other JSON changes are the result of file updaters.

### DIFF
--- a/library/games/Imagination/0.json
+++ b/library/games/Imagination/0.json
@@ -8,7 +8,7 @@
     "parent": "36xe"
   },
   "_meta": {
-    "version": 4,
+    "version": 13,
     "info": {
       "name": "Imagination",
       "image": "/assets/294730837_39781",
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1804/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1804-pr-test`.

ArnoldSmith86 (approved):

PR #1820, opened 2023-09-20
fixed minification

ArnoldSmith86:
After merging #1482, the production server did not start. The cause is that the minified Javascript contains `$&` and that it is put into the HTML by replacing `//*** JS ***//` by the minified contents. The `String.replace` function treats `$` as a special character to access a few parts of the matched results so the `$&` was replaced by `//*** JS ***//` again which made the resulting Javascript invalid.

Turns out this is the same problem that made me insert a bunch of `String.fromCharCode(36) // the minifier does not like $ or \x24 here` into the code.\n\nThe fix is to use an anonymous function as the replacement in `replace`. Its output is used as is.

### DIFF
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -258,7 +258,7 @@ const jeCommands = [
   {
     id: 'je_uploadAsset',
     name: 'upload a different asset',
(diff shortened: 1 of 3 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1820/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1820-pr-test`.

96LawDawg (approved):
Seems to work.  But so did the first try.  Have an approval.  And keep trying until it works.  Thani you!

PR #1823, opened 2023-09-26
Fixed typo in name of faceCycle property in FLIP

### LINKED ISSUE #1690: typo in `FLIP`
The call to `setDefaults` says `faceCyle`.

I gotta have to closer look before just fixing it.

rogerl50:
This fixes #1690.

This looks like a safe fix. As things currently stand, `a.faceCycle` will be `undefined` unless it was specified in the `FLIP` call. After this fix, `a.faceCycle` will be `null` if not specified in `FLIP`, and the flip routines in basicwidget and card treat `undefined` and `null` identically.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -1120,7 +1120,7 @@ export class Widget extends StateManaged {
       }
 
       if(a.func == 'FLIP') {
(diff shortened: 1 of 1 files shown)


ArnoldSmith86 (approved):

96LawDawg (approved):
Agree this change should not cause any issues.  FLIP also still works as expected through testing.  Approved.

PR #1824, opened 2023-09-26
added asset compression to editor

ArnoldSmith86:
This adds a Compress Assets button to the assets sidebar in the editor. The button lists all assets and makes a WebM and a PNG suggestion. It preselects the smaller one if it is at least 10% smaller than the original. The next button then replaces the selected assets.

Somebody should probably make this prettier.

- [x] Progress indication while compressing the images is the most important thing.
- [x] At least some minor layout tweaks to make it look less horrible would be nice.
- [x] A mouseover preview to see the assets in their native size for better comparison.
- [x] More text to tell people to use tinypng.com instead. 😁


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1824/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1824-pr-test`.

96LawDawg:
For assets that were already PNG, it still suggested a replacement (that was actually a larger file size in many cases).  If the original is already a PNG, probably filter out PNG results that are larger.

Also, I pressed the button to "Replace Selected Images" but nothing seemed to happen.  I thought maybe that was because I was working on a PL file, but when I just used a regular file, same thing: nothing.

ArnoldSmith86:
Can you make screenshots or a video? Did you activate any checkboxes?

96LawDawg:
So it works, but only after something (not sure what) retriggers it.  If you just click on the Replace button, then save the room state, then download it ... it is the same file as the original.  But if you click on the Replace button then click on the Compress Assets button again, then save the room state, then download it, the assets have changed.  I didn't test what other thing might trigger it to work, but I don't think just refreshing the room did.  Something to think about anyway.

Also, and this was very strange, after the assets had been compressed and I clicked on Compress Assets button a second time, my Malwayre Bytes software popped up with a malicious URL warning.  I tested this with two different games (Imagination and Flight).  I didn't get the warning for Imagination, but did for Flight.

96LawDawg (dismissed):
This is the same comment I made previously.  Just putting it in using the Request changes mode for tracking purposes:

So it works, but only after something (not sure what) retriggers it. If you just click on the Replace button, then save the room state, then download it ... it is the same file as the original. But if you click on the Replace button then click on the Compress Assets button again, then save the room state, then download it, the assets have changed. I didn't test what other thing might trigger it to work, but I don't think just refreshing the room did. Something to think about anyway.

Also, and this was very strange, after the assets had been compressed and I clicked on Compress Assets button a second time, my Malwayre Bytes software popped up with a malicious URL warning. I tested this with two different games (Imagination and Flight). I didn't get the warning for Imagination, but did for Flight.

96LawDawg (changes requested):
This still isn't right.  I'm not worried about the malware thing (just reporting it and it hasn't happened again in testing).  But clicking the button doesn't initially do what it needs to do.  You have to click on the Assets button again and then click on the Compression button again.  I think people will have a hard time figuring out how to get this to work.

Also, I noticed that when you change games and go back into the editor, if the Assets menu was already selected, it still shows the assets from the previous game until you close the Assets menu and open it again.

ArnoldSmith86:
> Also, I noticed that when you change games and go back into the editor, if the Assets menu was already selected, it still shows the assets from the previous game until you close the Assets menu and open it again.

Fixed. Also found and fixed a few other issues.

> You have to click on the Assets button again and then click on the Compression button again. I think people will have a hard time figuring out how to get this to work.

I can't reproduce this though. Or I don't understand the problem:
[cinnamon-2023-10-10T112621+0200.webm](https://github.com/ArnoldSmith86/virtualtabletop/assets/73538315/54134a5a-a0fa-4d5c-8972-c5ca7a5108c2)

96LawDawg:
I guess I haven't tested the right games.  Or you haven't.  I am testing it the same way you did in the video.  It worked for me in Alpaca.  I even got the nice little screen flash that showed up in your video as confirmation.  It worked in Crafty.  It does not work for Flight or Chronology.

ArnoldSmith86:
At least Flight has a lot of assets, right? Maybe you simply didn't wait long enough? It has to upload each new asset to the server first.

Either way, this needs a progress indicator. But it would be helpful to know if that's the only problem.

96LawDawg:
I originally waited 30 seconds.  This time, I waited for Flight for maybe 10 minutes.  It worked.  So a progress bar is definitely needed.

One other thing.  In the columns for WebP and PNG it has (0x0) for the dimensions in every instance.  Shouldn't that reflect the dimensions of the original file, or something close to it?

96LawDawg:
All of the changes here look good.  I think the only thing missing is a progress bar?

ArnoldSmith86:
Yeah, I was still working on this when I ran out of time last week. Progress indication is the most important thing. At least some minor layout tweaks to make it look less horrible would be nice and a mouseover preview to see the assets in their native size for better comparison. And more text to tell people to use tinypng.com instead. 😁

96LawDawg:
Tested the upload bar on Flight.  Looked great.  After uploading (about 45 seconds), I clicked on the Replace button and got "Type error: cannot read properties of undefined (reading 'uses')".

ArnoldSmith86:
You clicked it _again_ after it was done?

96LawDawg:
> You clicked it _again_ after it was done?

Yes.  That.  I also just got the malware pop-up again.

96LawDawg:
Does not show the detail text reliably in dark mode:

Screenshot 2023-11-08 083701

ArnoldSmith86:
Thanks.

ArnoldSmith86:
I think this is pretty nice now. Any more problems?

96LawDawg:
Looks great.  The hover thing is really nice.  One final thing.  And this may not be fixable, so maybe just include a note at the top where you have some explanatory text about the options.

If the room divider between edit and play areas is moved to the left far enough, you will never be able to see the comparison image because you can't hover and scroll at the same time.

Screenshot 2023-11-15 102623

96LawDawg:
And by the way, Flight, one of the largest PL games at 40+ MB goes down to 2 MB with this tool.

ArnoldSmith86:
> And by the way, Flight, one of the largest PL games at 40+ MB goes down to 2 MB with this tool.

And do you feel like the quality is good enough?

ArnoldSmith86:
> If the room divider between edit and play areas is moved to the left far enough, you will never be able to see the comparison image because you can't hover and scroll at the same time.

It should now be side-by-side no matter the size. If the asset is too big, it'll still be off screen though. But I feel like making it smaller isn't helpful. To compensate a little bit, I switched the images so that you can always see parts of the hovered image. That way you can hover between the original and compressed version to compare them.

96LawDawg:
> > And by the way, Flight, one of the largest PL games at 40+ MB goes down to 2 MB with this tool.
> 
> And do you feel like the quality is good enough?

Yes. No noticeable difference in the size used in the room.

96LawDawg (approved):
Approved through test.

PR #1825, opened 2023-09-26
fixed adding games with multiple variants to the public library

ArnoldSmith86:
Adding the first variant already tried to save all assets from all variants but then could not read the second file.

### DIFF
--- a/server/room.mjs
+++ b/server/room.mjs
@@ -96,8 +96,13 @@ export default class Room {
           attribution: ''
         };
 
-        if(stateID.match(/^PL:/))
-          return this.writePublicLibraryToFilesystem(stateID, newVariantID, variant);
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1825/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1825-pr-test`.

96LawDawg (commented):
I don't think this worked.  I can't load, for example, the Team II (but under a different name) into this PL in this PR.

96LawDawg (approved):
Now that I tested it properly, it works.  Approved through test.

mousewax:
Yes. This seems to work on my server as well.

PR #1826, opened 2023-09-27
added cardType management to editor

ArnoldSmith86:
This adds a new UI for managing how many cards of each type are in the game and their parameters.

- [x] move to deck
- [x] show info for just one cardType for cards
- [x] add button to card that selects the deck
- [x] better category titles
- [x] fix destructiveness (value types and special strings change just by pressing Apply)
- [x] grid layout
- [x] "You can change it once. For example, change the Front face width from 103 to 153. Hit apply and everything works. Try changing it back to 103 and it doesn't work. You have to reload the room first. This is currently an issue on production as well, so it is a bug."
- [x] smaller images?
- [x] clarify what buttons do
- [x] add Cancel button
- [x] fix editing cardDefaults
- [x] custom properties
- [x] margin between properties and buttons
- [x] buttons "+/-1 to all"


RaphaelAlvez:
Looks promising! I suggest you don't spend much time perfecting it as it is already usable.

here are some suggestions feel free to ignore them:
- When selecting a card, the card information should come first
- the deck information should come with some visual information that it is not in the card. Maybe a grey background.
- if possible highlight the the current card in cardTypes
- I suggest making each card type a vertical rectangle and then when we expand the sidebar we could get a grid of cards. by that, I mean to fix the width (not height as there might be more variables to list) of the faceTemplateEdit class and let the flexbox decide the best number of cards to display. currently, you are resizing the faceTemplateEdit elements

keep this layout:
image
and let it grid instead of doing this 
image

96LawDawg:
I agree with what @RaphaelAlvez said.  Especially that this does not have to be perfect and the card information should be first in that Properties list.  Currently it is at the very bottom.

Also, there is a strange effect when changing properties using the Back face or Front face boxes.  You can change it once.  For example, change the Front face width from 103 to 153.  Hit apply and everything works.  Try changing it back to 103 and it doesn't work.  You have to reload the room first.  This is currently an issue on production as well, so it is a bug.  But maybe it can be fixed as part of this PR.

rogerl50:
I feel dumb, but what exactly do I need to do in order to see this? I can't seem to figure it out.

96LawDawg:
> I feel dumb, but what exactly do I need to do in order to see this? I can't seem to figure it out.

Select a card in the JSON Editor.  Click on the Properties button on the right side menu.

rogerl50:
This looks great. A few requests:
1. As @96LawDawg suggested, the card information, currently at the bottom, should be at the top.
2. Can you make the card images smaller and perhaps move the svg box to either above or below the card? Then you could get two cards side by side.
3. What do the Apply Changes buttons apply to? I assume the first one affects changes to the card, and the second one changes to the deck. But currently changes to the card information at the bottom happen instantaneously.
4. The whole UI might be clearer if there were two tabs, one for the card and one for the deck.

ArnoldSmith86:
Updated layout:

image

96LawDawg:
I'm liking where this is headed.  Soon our deck editor will be better than PCIO's.

Feedback: After clicking on Edit, you get an "apply changes" button.  Also need a cancel button.

96LawDawg:
I'm working on a simple card game to test this PR and to pump up the numbers for November.  I had a bunch of png images for cards on my computer that I added.  They were named 01.png, 02.png, ... 30.png.  I wanted to be able to review them in that order in the room.  The upload created cardTypes with mostly random order.  So I reordered them all manually in the JSON Editor and got rid of the ".png" so I could use the number from the cardType as a dynamic property.  Then I went to view them using the tools in this PR.  And they were still out of order.  Because 01 comes after 30.  So I renamed 01-09 simply 1-9 which I should have done in the first place.  That finally worked.

So this is just feedback on how it worked for me.  The biggest thing I had issue with was the random entry into the cardTypes.  And that isn't actually part of this PR.

ArnoldSmith86:
> After clicking on Edit, you get an "apply changes" button. Also need a cancel button.

I totally agree but this is a lot more work than you think. Because the Apply Changes button is already half faking stuff (it applies changes to _all_ card types).

---

I rearranged stuff. This is by no means my intended end result for anything but it already far exceeds what I wanted to do with this PR (basically just add a way to change card counts).

I want to continue working on the editor UI soonish but let's focus on getting this one merged pretty quickly, please. :grinning:

96LawDawg:
Actually, I was talking about here on not on the main Properties overlay.

Screenshot 2023-11-19 103605

Once you click on Edit, you can't even close the enlarged card view.  Maybe you don't even need Cancel.  Just "Close."  And for cancel, it is possible to use the History button to undo your changes, so I'm okay with not having a cancel button anywhere.  I just wanted to be able to close the enlarged view.

ArnoldSmith86:
> Actually, I was talking about here on not on the main Properties overlay.

Me too.

96LawDawg:
> Me too.

So give me a close button and I will approve.

96LawDawg (approved):
Approved through test (and persuasion in PMs on Discord!)

ArnoldSmith86:
Now with Discard button:

image

Please do a little more testing because it was a serious code change. Feel free to merge.

96LawDawg:
Everything still works and now there is a functional cancel button also.  Fantastic!

I would merge, but I have one more question/comment. The "Card layers" section of the deck manager list only those properties that are in the faceTemplates section of the deck.  But frequently, there are properties in the cardType not represented there.  For example, a sort key or a point value. It would be nice to be able to change that directly in this editor as well. Feel free to disregard and merge. We'll just note it for next enhancement. 

Of course, that leads me to also want to be able to add and remove properties directly in the Properties overlay as well.  Like a + button to add a new line and a trash can icon beside each property to remove it.  But that is a bigger UI thing and outside the scope of this PR.

One final note for myself. In the wiki, we should point out that some changes in the deck manager UI are instant and some require you to push the apply changes button.  Instant: adding or removing cards using +/- buttons; card default properties (or just "card properties" if viewing a single card and not the whole deck). Button approval: "card layers" and edits to cardTypes properties on individual cards.

ArnoldSmith86:
In general, changes that trigger the entire deck to be redone, need an Apply button because it takes too long for bigger decks.

I agree about "custom" properties. I forgot to revisit that.

And as you say, adding and removing properties will be a bigger change.

96LawDawg:
Feel free to merge at any time.  My comments are just minor incremental things.

In this image from a deck ...

Screenshot 2023-11-20 170053

If the property is not used, then there is not blank.  It would be better (but is not necessary) if the card properties had a blank where information could be added if needed.

ArnoldSmith86:
I don't understand what you mean by "blank". Are you saying, there should be an empty text input for `image2` and the others? Just change the "not set" to "text".

The Apply button shouldn't change it from not being set at all to `"" ` if you didn't want that change explicitly.

96LawDawg:
> Just change the "not set" to "text".

Yeah.  I didn't try that.  Just saw the blank and assumed I couldn't.  Stupid.  Nevermind.  All's good.

PR #1828, opened 2023-09-28
fixed dropTarget setting in editor

ArnoldSmith86:
In production it only works the first time and also breaks stuff because it removes the visualized deck from `widgets`. This should fix that.

### DIFF
--- a/client/js/editor/sidebar/properties.js
+++ b/client/js/editor/sidebar/properties.js
@@ -1054,7 +1054,8 @@ class PropertiesModule extends SidebarModule {
     for(const deck of widgetFilter(w=>w.get('type') == 'deck')) {
       if(!Object.keys(deck.get('cardTypes')).length)
         continue;
-      const deckButton = this.renderWidgetButton(deck, {}, this.moduleDOM);
+      const deckCloneState = Object.assign(deck.state, { id: generateUniqueWidgetID() });
+      const deckButton = this.renderWidgetButton(new Deck(deckCloneState.id), deckCloneState, this.moduleDOM);
       this.addPropertyListener(widget, 'dropTarget', widget=>{
         if(asArray(widget.get('dropTarget')).filter(t=>t.deck == deck.id).length)
           deckButton.classList.add('selected');
--- a/client/js/editor/sidebar/properties.js
+++ b/client/js/editor/sidebar/properties.js
@@ -1054,7 +1054,7 @@ class PropertiesModule extends SidebarModule {
     for(const deck of widgetFilter(w=>w.get('type') == 'deck')) {
       if(!Object.keys(deck.get('cardTypes')).length)
         continue;
-      const deckCloneState = Object.assign(deck.state, { id: generateUniqueWidgetID() });
+      const deckCloneState = Object.assign({}, deck.state, { id: generateUniqueWidgetID() });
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1828/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1828-pr-test`.

96LawDawg:
This fixes the problem you identified with it only working the first time.  But I guess I'm confused as to the purpose of the Target Widgets section on the Properties view.  My dropTarget is this:
```
"dropTarget": {
    "pin": true
  }
```
When I click on the holder and the Properties view, it shows cards.  That is already confusing.  Why is it showing me cards?  If I click on the cards then that deck gets added automatically, which is nice, I guess.  But if I have multiple decks in the same room that all have the same cards, I can't tell from the small images which deck is which.

But more importantly, this breaks stuff.  If you click on Properties on the far right menu, then click on JSON, you get dumped back out of the Editor.

Also, the drag bar is missing:
Screenshot 2023-09-28 145451

ArnoldSmith86:
> But more importantly, this breaks stuff. If you click on Properties on the far right menu, then click on JSON, you get dumped back out of the Editor.

Thanks, this is fixed.

> Also, the drag bar is missing:

When is it missing? And what's the state of this? 
image

> When I click on the holder and the Properties view, it shows cards. That is already confusing. Why is it showing me cards? If I click on the cards then that deck gets added automatically, which is nice, I guess. But if I have multiple decks in the same room that all have the same cards, I can't tell from the small images which deck is which.

Those are very valid points. I'll add the deck name under the button and maybe even highlight the deck in the room view when hovering over one of those buttons. I'll also add an image with "custom" drop targets. But this is meant as a bug fix PR.

RaphaelAlvez:
> Those are very valid points. I'll add the deck name under the button and maybe even highlight the deck in the room view when hovering over one of those buttons. I'll also add an image with "custom" drop targets. But this is meant as a bug fix PR.

In this case, the drop target is cards and it shows 2 cards option as there are multiple decks but the selection is not for decks. not for the deck but a multi-select for decks would solve it.

ArnoldSmith86:
> In this case, the drop target is cards and it shows 2 cards option as there are multiple decks but the selection is not for decks. not for the deck but a multi-select for decks would solve it.

I'm not sure what you mean. The buttons are already multi-select.

RaphaelAlvez:
> > In this case, the drop target is cards and it shows 2 cards option as there are multiple decks but the selection is not for decks. not for the deck but a multi-select for decks would solve it.
> 
> I'm not sure what you mean. The buttons are already multi-select.

I worded it very poorly I mean that there is an inconsistency here. Not sure what's the best solution:
image

image

96LawDawg:
> > Also, the drag bar is missing:
> 
> When is it missing? And what's the state of this? image
> 

Oops.  I don't know how that got toggled.  I didn't even remember that was a toggle.  Disregard.  That is not a bug.

ArnoldSmith86:
Guys, you seem to be reviewing the feature, not the bug fix. I agree with everything you say but I'd like to merge the bug fix first.

This bug actually removes the deck from the `widgets` variable which leads to all sorts of trouble until the room is reloaded.

RaphaelAlvez (approved):
Reviewed. Tested

PR #1829, opened 2023-09-28
added property `hidePlayerCursors`

ArnoldSmith86:
This is keeping it simple. It might be a good idea to highlight the widget instead of showing a cursor but this PR really just hides the cursor.

It is added to new hands by default and also to any holder with `childrenPerOwner: true` by using a file updater.

### DIFF
--- a/client/css/overlays/players.css
+++ b/client/css/overlays/players.css
@@ -181,3 +181,6 @@ label.ui-line::before{
   border: none;
   background-color: var(--playerColor);
 }
+.cursor.hidden {
+  display: none;
+}
--- a/client/js/editmode.js
+++ b/client/js/editmode.js
@@ -1795,6 +1795,7 @@ export function initializeEditMode() {
       stackOffsetX: 40,
       childrenPerOwner: true,
       dropShadow: true,
+      hidePlayerCursors: true,
       x: 50,
       y: 820,
       width: 1500,
--- a/client/js/mousehandling.js
+++ b/client/js/mousehandling.js
@@ -43,6 +43,7 @@ async function inputHandler(name, e) {
       return;
     target = target.parentNode;
   }
+  const targetForHiddenCursorCheck = target && target.id && target.id.slice(0,2) == 'w_' && widgets.has(unescapeID(target.id.slice(2))) ? target : null;
 
   e.preventDefault();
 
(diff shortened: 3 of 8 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1829/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1829-pr-test`.

ArnoldSmith86:
> Should we make this the default for hands?

Done, but only for new hands. Not sure if we should do a file updater and set it to every `childrenPerOwner: true` holder. Feels like a bit of overkill.

Also not sure if the PCIO importer should use this. Does anybody know if PCIO supports hiding cursors?

96LawDawg (approved):
Approved through test.  I think it is fine as is without changing existing hands or any childrenPerOwner stuff.  Leave it at hands.  As far as I know, other than making all cursors the same color to anonymize users, there are no other cursor settings.

96LawDawg:
Just coming back to this.  The vote on Discord, such as it was: @RaphaelAlvez  and jade-alarm for retroactive application; @96LawDawg  neutral.  No way to tell for @ArnoldSmith86  based on how he had to set up the scoring.

Here's my point.  This is ready now.  Preparing a file-updater will take additional time.  @mousewax already has a game ready to go.  For future hands after this PR, it will be automatic anyway.  Does retroactive application really matter?  I don't think so.  The problem of distracting cursors probably only exists in a small subset of existing games any way.  And there is probably also a very small subset of games where developers would not want the cursor to go away.  Why hold this up for a file-updater?

RaphaelAlvez:
> Does retroactive application really matter?  I don't think so.  The problem of distracting cursors probably only exists in a small subset of existing games any way.  

I believe that almost any game that has a hand has this anoyance. Of course we get used to it but it is not a feature. On top of that if you sorta hand, it's somewhat easy to know which card is being played.


> And there is probably also a very small subset of games where developers would not want the cursor to go away.

Yeah that's the whole point. I cannot think of any situation where leaving cursors on hands is better. Breaking any game is very unlikely. So why leave it?


So if we are going to be less strict about this fear of breaking games this is an excellent first step.

ArnoldSmith86:
You guys do realize that I already added the file updater?

96LawDawg:
> You guys do realize that I already added the file updater?

Obviously not.  Oops; missed that.

So I checked several games and the file updater worked on most of them.  But not on Card Games and Knaves.  Not that big of a deal because can always add that manually.  But it is a little concerning that it missed those.  Their hand holders meet the criteria for having it added.  Wonder why it missed?

PR #1832, opened 2023-09-28
added `serverName` to config

ArnoldSmith86:
Currently only used for the title tag.

### DIFF
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -552,12 +552,7 @@ onLoad(function() {
     if(!checkedOnce)
       checkURLproperties(true);
     checkedOnce = true;
-    let tabSuffix;
-    if (config.customTab) {
-      tabSuffix = config.customTab;
-    } else {
-      tabSuffix = "VirtualTabletop.io"
-    }
+    let tabSuffix = config.customTab || config.serverName || 'VirtualTabletop.io';
     document.title = `${document.location.pathname.split('/').pop()} - ${tabSuffix}`;
   });
 });
--- a/client/room.html
+++ b/client/room.html
@@ -3,7 +3,7 @@
 
 <head>
   <meta charset="utf-8">
-  <title>VirtualTabletop.io</title>
+  <title> /*** TITLE ***/ </title>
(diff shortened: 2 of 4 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1832/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1832-pr-test`.

96LawDawg:
Related to #1566 which was merged without this capability because I couldn't figure out how to do it.

96LawDawg:
Sorry if this is a stupid question, but I don't see how this changes what we can already do.  1566 added a customTab to the config.  This seems to be just a replacement for that since it is either/or?

My config.json has
```
{
  "serverName": "Mike",
  "port": 8272,
  "externalURL": "http://localhost:8272",
  "urlPrefix": "",
  "minifyJavascript": false,
  "forceTracing": false,
  "customTab": "Local Stuff",
 ...
```
I can see either Mike or Local Stuff but not both.  What am I missing?

mousewax:
This adds putting the server name in the title by default. When linking to a room on my private server the preview on messengers and social media has VirtualTabletop.io as the name of the site since those previews do not load any javascript so the title never gets changed. I have been manually changing the room.html after every update so that it has our server name on it. This fixes that.

Can this update also include the name of the game that's currently loaded in the room (if any) as the default title?

mousewax (approved):
Approved by test.

ArnoldSmith86:
> This seems to be just a replacement for that since it is either/or?

Yeah, it kinda replaces `customTab`. Not sure if we want to keep both. The advantage of this implementation is that it doesn't flash "VirtualTabletop.io" in the browser tab before showing the custom name and - like mousewax says - that bots read the new name for stuff like link previews in messengers.

96LawDawg (approved):
It works.  I don't understand it but that doesn't matter as long as you guys are happy.  Have an approval.

PR #1833, opened 2023-09-29
fixed a few JSON editor issues

ArnoldSmith86:
- When the tree is not active, it will no longer be loaded (speeds up opening the JSON editor with many widgets).
- Changing selection no longer reloads the entire tree.
- External changes should be reflected in the tree and the JSON editor again.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1833/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1833-pr-test`.

96LawDawg (approved):
Approved.  I'm not sure how it's possible to test these things through just regular use.  Does a tree that doesn't load make a sound if there is nobody there to hear it?  But the JSON Editor works as expected and after the last tweak, the tree listing behaves as expected.

ArnoldSmith86:
If you load something like Smash Up, loading the tree takes seconds. So it should be pretty noticeable for big games.

96LawDawg:
No noticeable difference for me using Smash up, but I have a pretty fast computer.  Still approved.

PR #1835, opened 2023-09-29
added Open Graph support

ArnoldSmith86:
This PR checks the user agent against a list of known bots. The idea is to send the normal room HTML to browsers and Open Graph metadata to bots so we can change link previews without requiring to modify the room HTML on every access.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1835/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1835-pr-test`.

RaphaelAlvez (approved):
Tested. Worked on many platforms. Approved

PR #1842, opened 2023-10-04
progressive web app

ArnoldSmith86:
This will make VTT installable on mobile devices, acting a lot more like a native app.

### DIFF
--- a/assets/branding/manifest.json
+++ b/assets/branding/manifest.json
@@ -1,14 +1,18 @@
-{ 
-    "name": "VirtualTabletop.io", 
-    "short_name": "VTT.io", 
-    "icons": [ 
-        { "src": "/i/branding/favicon-16.png", "sizes": "16x16", "type": "image/png" }, 
-        { "src": "/i/branding/favicon-32.png", "sizes": "32x32", "type": "image/png" }, 
-        { "src": "/i/branding/favicon-48.png", "sizes": "48x48", "type": "image/png" }, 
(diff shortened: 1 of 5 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1842/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1842-pr-test`.

96LawDawg:
From Tavo! in the general channel on Discord: "I'm just testing the progressive web app. Installed in my device. Uploaded my game. Was able to play, but if I close the app it looses everything, even if I save state."

RaphaelAlvez:
I just tested this. There is no url bar in an installed mobile version so it always sends you to a new room.

But I have to say the notification bar looks amazing in VTT blue



Having an input for game room/url would be also useful for Fullscreen mode

RaphaelAlvez:
The icon also looks great

robartsd:
Currently the server always redirects the root of its namespace to a new random room. The proressive web app always opens to the root of the server; it would be better if this got redirected to the last room used in the progessive web app.

I think we should add a feature where the client can keep track of rooms it has visited in local storage (possibly off by default except when the client is a progressive web app). The progressive web app could detect that it has been sent to an empty room and redirect itself to the last room it used. If the progressive web app user requests a new room, the app should set a flag in local storage directing it to cancel the automatic redirect from an empty room to the last room used (and instead add the new room to its list of remembered rooms and clear the flag).

PR #1846, opened 2023-10-07
Add Grid Lines Toggle Button to JSON Editor

96LawDawg:
This adds a button to the JSON Editor to toggle grid lines on and off.  The css class has variables that allow for the flexibility of adjusting the color and for adjusting the size of the grid.  I haven't been able to figure out a way to trigger an input for that.  I know it isn't that hard, but this was the limit of my ability.  It would also be nice to have permanent lines at 25,50,75 percent of the screen like I did using the button I made with JSON.  But I can't figure that out either.

In any event, this is a pretty good start and could even be merged as is because it adds something we don't have that has been requested.  We could always add the additional functionality later.

To do:
- [x] Add overlay to choose grid size, line color
- [x] Overlay also includes option for major grid lines divided in thirds, quarters, or fifths (or none)
- [x] Add major grid lines for x and y axis (pre-set to divide room into quarters)
- [x] Populate overlay with current values


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1846/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1846-pr-test`.

96LawDawg:
Added major grid lines.  Defaults to 25%.  I have css in there for grid lines defaulting to thirds and fifths as well.  Can't imagine needing anything else.  They may be a more elegant way to code the css (perhaps using mix-blend-mode), but I couldn't get that to work.  Would be good to have the option to choose between the 4 possible major grid line options (none, thirds, quarters, or fifths).

96LawDawg:
I have refactored the code to make it more flexible.  I don't know that one way is better than the other, but the value of doing it this way is that you don't have to turn on and off specific classes for the major gridlines.  You just need to figure out if the user wants major lines at thirds, quarters, or fifths (or none) and set a couple of variables.  I still don't know how to make the pop-up overlay that will get the grid size, grid color, and major lines preferences.

96LawDawg:
I made the overlay!  Yay me.  I'm sure there are improvements that could be made to the code and maybe to the functionality.  But I think this works.  Ready for testing, review, and comment, please.

96LawDawg:
What I can't seem to get the code to do is remember what the last state was when you toggle the grid on and off.  It just goes back to the default.  Would be nice to have it remember AND pre-fill in the entry fields with the correct data.  But that isn't really essential and I don't think I can devote any more time to this.

96LawDawg:
I figured out how to get the entry fields to populate with current data.  Now this is sweet.  Except for code nit things that I can't possibly begin to figure out, I don't think this could be any better.  Review?

96LawDawg:
None
Third
Quarter
Fifth

RaphaelAlvez:
I am pleased to say I was able to test this on my phone

Looks very good.

Just a question, would it maybe be better to have square grid? I was thinking of 10:10, 50:50 or 100:100

RaphaelAlvez (approved):
Tested on mobile

96LawDawg:
> Just a question, would it maybe be better to have square grid? I was thinking of 10:10, 50:50 or 100:100

I suppose I could also create a 1:1 layout centered left-to-right so that you get a square board with major lines making squares instead of rectangles.  But are you also suggesting that there be thick lines every 50px?

RaphaelAlvez:
> > Just a question, would it maybe be better to have square grid? I was thinking of 10:10, 50:50 or 100:100
> 
> I suppose I could also create a 1:1 layout centered left-to-right so that you get a square board with major lines making squares instead of rectangles.  But are you also suggesting that there be thick lines every 50px?
> 

100 or 50 not sure which would be better

PR #1848, opened 2023-10-11
fixed editing public library games without assets

ArnoldSmith86:
Looks like git does not add empty directories to the repository. Either way, this PR makes sure that a public library game has an assets directory before updating its assets.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1848/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1848-pr-test`.

96LawDawg:
Did not work.  I created and added an "Unnamed" PL file that only has a pin in it.  No asset directory was created.  I left it uploaded for now so you can see, but that commit obviously needs to be reverted before this is merged.

ArnoldSmith86:
But does it crash if you now add assets to it? That was the problem. I wanted to fix that Nexus still uses links to its assets and it caused a crash.

96LawDawg:
> But does it crash if you now add assets to it? That was the problem. I wanted to fix that Nexus still uses links to its assets and it caused a crash.

It does not crash.  Works to add assets after the fact.

96LawDawg (approved):
Approved through test

PR #1850, opened 2023-10-12
added option to not clear the debug log automatically

ArnoldSmith86:
Stuff like timers can make it hard to debug stuff if it clears automatically.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1850/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1850-pr-test`.

96LawDawg:
Tested and it works.  I like that it resets to the default (clearing after every interaction) when you leave or refresh the room, but that it stays on your selected choice when you just open and close the editor.

However, one thing should be changed.  It should clear the debug history on a state change.  Currently if you load a new game, the old debug listing is still there.

PR #1856, opened 2023-10-15
Fix so player cannot drag widgets off screen in play mode

### LINKED ISSUE #1855: Can drag widgets off screen completely
Since PR #1482 in commit https://github.com/ArnoldSmith86/virtualtabletop/pull/1482/commits/8d3d70e0915e8e6f26d426c901653d62095501e7 this has been a problem (according to git bisect).

96LawDawg:
Fixes #1855

### DIFF
--- a/client/js/mousehandling.js
+++ b/client/js/mousehandling.js
@@ -11,7 +11,7 @@ function eventCoords(name, e) {
     coords = e;
   let x = (coords.clientX - roomRectangle.left) / scale;
   let y = (coords.clientY - roomRectangle.top) / scale;
-  if(!edit) {
+  if (!edit || (edit && zoom == 1)) {
     x = Math.max(0, Math.min(1600, x));
     y = Math.max(0, Math.min(1000, y));
   }
--- a/client/js/mousehandling.js
+++ b/client/js/mousehandling.js
@@ -11,7 +11,7 @@ function eventCoords(name, e) {
     coords = e;
   let x = (coords.clientX - roomRectangle.left) / scale;
   let y = (coords.clientY - roomRectangle.top) / scale;
-  if (!edit || (edit && zoom == 1)) {
+  if (!edit || zoom == 1) {
     x = Math.max(0, Math.min(1600, x));
     y = Math.max(0, Math.min(1000, y));
   }
--- a/client/js/mousehandling.js
+++ b/client/js/mousehandling.js
@@ -11,7 +11,7 @@ function eventCoords(name, e) {
     coords = e;
   let x = (coords.clientX - roomRectangle.left) / scale;
   let y = (coords.clientY - roomRectangle.top) / scale;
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1856/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1856-pr-test`.

96LawDawg:
Nevermind.  This doesn't fix it.  It fixes it partially, but only sometimes.

96LawDawg:
Okay, I have it fixed now for play mode.  You can still drag off screen in edit mode when not zoomed out, but I don't know if we should care about that or not.

ArnoldSmith86:
`edit === false` is redundant as it's included in `!edit`.

Was it broken in edit mode as well?

96LawDawg:
> `edit === false` is redundant as it's included in `!edit`.

I literally just fixed the redundancy as you commented with the commit name "this is what happens when you let a computer think for you."  I don't know much javascript, but after looking at that again, even I figured it out.

> Was it broken in edit mode as well?

Yes, on production right now, you can drag off-screen when not zoomed out.  Want me to try and fix that?

ArnoldSmith86:
> Yes, on production right now, you can drag off-screen when not zoomed out. Want me to try and fix that?

Well, if you won't, I will. But it might be that `zoom` is only defined in the editor code. The edit mode and play mode are completely separate now with a list of defined functions so they don't share their "global" variables.

96LawDawg:
> > Yes, on production right now, you can drag off-screen when not zoomed out. Want me to try and fix that?
> 
> Well, if you won't, I will. But it might be that `zoom` is only defined in the editor code. The edit mode and play mode are completely separate now with a list of defined functions as an interface between them so they don't share their "global" variables.

Not a matter of willingness.  Just ability.  But I think this works properly now.  I had no problem determining when the edit mode was on using the "edit" variable (it was always either null, true, or false) and when zoomed in using the "zoom" variable (either 1 or 0.5).

ArnoldSmith86:
Ah, so `zoom` is not a boolean. That explains it. Probably don't even need to check `edit` then.

In either case, the `edit &&` is redundant again because otherwise the first part of the condition would have been `true` already.

96LawDawg:
> Ah, so `zoom` is not a boolean. That explains it. Probably don't even need to check `edit` then.

No.  Still have to check for both.  You can leave leave edit mode while zoom is 0.5.  Then you can still move widgets out of the play area.  So the condition needs to be either not in edit or zoom = 1.

> In either case, the `edit &&` is redundant again because otherwise the first part of the condition would have been `true` already.

Fixed.

ArnoldSmith86 (approved):
Code looks good to me. Also seems to work fine.

Issue #1859, opened 2023-10-16
showInactiveFaceToSeat property sometimes shows the active face to other players


mousewax:
There are three times when other players see the activeFace even when showInactiveFaceToSeat is set:

1. When dropping a card into a holder with this property the active face flashes briefly. The timing of events needs to be adjusted.
2. Enlarge always shows the active face (revealing the card to anyone hovering over it) and should probably show the face you can see instead.
3. Anyone not in seats always sees the active face and so opening another window in incognito or private mode bypasses the property entirely. A way should probably be set to exclude spectators from seeing the active face.

Maybe in case 3 we can change it to the affirmative and the property can become "showActiveFaceToSeats" and when that is set and you are not in that list you will see the next inactive face. 

Or by creating a special group called "spectators" and specifying them. 
`"showInactiveFaceToSeats": [ "seat2", "seat3", "seat4", "spectators" ]`

Or being allowed to set a negative like in inheritFrom: `"showInactiveFaceToSeats": [ "!seat1" ]`.

96LawDawg:
> Or being allowed to set a negative like in inheritFrom: `"showInactiveFaceToSeats": [ "!seat1" ]`.

That is like a double negative and makes my head hurt.  Looking at that option, I don't know what would be shown to seat1 or any other seat.  I'm sure I could figure it out, but I think that is the least intuitive of your suggestions.

mousewax:
It was just an idea. But think of it as:
`"inheritFrom": {
   "otherThing": [ "!parent" ]
 }`
 
 It means inherit everything except that. So this would be show inactive face to all seats except that seat.

ArnoldSmith86:
I still don't understand why you need "show face 1 to all seats **but** seat1 and face 0 to others" instead of doing "show face 0 to seat1 and face 1 to others".

mousewax:
Just for brevity as one would have to specify all of the other seats on each holder for each seat. So in a six player you game you have a unique list of five seats plus the spectators that you can't just inherit from somewhere. Saying "in this holder only I can see what's going on" seems simpler than saying "seat2, seat3, seat4, seat5, seat6, spectators cannot see these cards". But the next holder over has to also say "seat1, seat3, seat4, seat5, seat6, and spectators cannot see these cards" and repeating for each seat.

ArnoldSmith86:
[...]

PR #1860, opened 2023-10-17
fixed `showInactiveFaceToSeat` with `enlarge`

ArnoldSmith86:
The enlarged view would not show the same face as the card. Part of #1859.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1860/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1860-pr-test`.

96LawDawg (approved):
Approved through test.

I figured out the problem with the cards tutorial example.  This PR changed a css overflow parameter for .cardFace.  Simply adding ` "#enlarged > .cardFace": { "overflow": "visible !important"}` onto the deck fixed the problem.

It is possible that some designer has used this technique elsewhere and it would remain broken unless they add that css onto their deck.  I don't think this has been widely used, if at all.  And I don't know how we would detect that and do a fileUpdater that wouldn't also risk breaking other things.  So I say leave it as is.

My commit fixes the tutorial.

PR #1861, opened 2023-10-17
fixed `showInactiveFaceToSeat` flashing wrong face when hovering

ArnoldSmith86:
This was more complicated than expected because a receiving client did not have the required information.

That is why this PR introduces a new property `hoverParent` similar to `hoverTarget`. It contains the ID of the holder the card still kinda belongs to. When rearranging cards in a hand, the card gets its `parent` set to `null` but `onLeave` only applies when the card is actually dragged _out_ of the hand so that other players do not see it. This exception makes a lot of stuff weird and now it is more visible by looking at `hoverParent`.

This is part of #1859.

Update wiki entry based on this PR:

# Global properties for every widget

| Property  | Default value       | Description           | Tutorial              |
| --------  | ------------------- | --------------------- | --------------------- |
| hoverParent| null | Added to a widget picked up from a holder. The value is the `id` of the previous parent. The value remains on the widget until it is mostly out of the holder or until dropped back into the same holder. Used internally to determine when the 'showInactiveFaceToSeat` property is applied. Not intended to be set by a game. |
| hoverTarget| null | Added to a widget picked up from or being dragged over a holder. The value is the `id` of the widget that will become the parent of a widget being dragged if it is dropped. The value remains on the widget as long as it could be dropped into that holder. Used internally along with `dragging` property to implement behavior of `hoverInheritVisibleForSeat`. Not intended to be set by game. |


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1861/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1861-pr-test`.

96LawDawg (approved):
Approved through test.  I added a `hoverParentGlobalUpdateRoutine` to a room and tested when that property was applied as well as watching on a widget.  It appears to work as expected and does fix the second issue from #1859.

But let me just make sure I understand this correctly so I can explain it in the wiki at some point:  `hoverTarget` is activated when a widget is picked up from or dragged over a holder.  The `hoverTarget` property is temporarily added to the widget with the value of the '`id` of the holder.  `hoverTarget` remains on the widget as long as the widget could be dropped into that holder.  That is what we have now.

This PR adds a new property `hoverParent`.  That is added to a widget picked up from a holder.  The value is the name of the holder.  And that property stays on the widget until it is mostly out of the holder (when the shadow disappears, when the border highlight disappears, and when `onLeave` activates). Or until it is dropped back in the holder.  And then this is used to determine when the inactiveFaceToSeat property kicks in.

Something like that?

ArnoldSmith86:
Sounds right. Can you make a few tests with piles in holders as well?

96LawDawg:
Good thought.  I should have considered that.

`hoverParent` works with piles in the sense that it shows up on the card being moved into the holder.  But the value is for the holder when really the pile is going to be the new parent.  So really, `showInactiveFaceToSeat` does not work with piles.  The first card in the holder works. The first card remains the way it should. The other cards (at the top of the pile) show the face as if `showInactiveFaceToSeat` doesn't apply.  Same behavior on production.

96LawDawg (changes requested):
Should probably be made to work with all the cards in a pile.

ArnoldSmith86:
And of course it turns out that if you drag the top card of a pile in a holder, it gets `onLeave` applied immediately and not when the card actually leaves the holder.

ArnoldSmith86:
This only has problems with piles, right? I think we should merge it and deal with piles later. Nobody would use this with piles anyway and it seems like there are larger issues with piles (as usual).

96LawDawg (approved):
> This only has problems with piles, right? I think we should merge it and deal with piles later.

Correct.  Agreed.  Approved through test.

PR #1869, opened 2023-10-21
Allow cancel button on INPUT overlay to be hidden

96LawDawg:
If you make "cancelButtonText" and "cancelButtonIcon" null, then the cancel button is not displayed.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -2313,6 +2313,7 @@ export class Widget extends StateManaged {
 
   async showInputOverlay(o, widgets, variables, problems) {
     $('#activeGameButton').dataset.overlay = 'buttonInputOverlay';
+    $('#buttonInputCancel').style = "display: initial";
     return new Promise((resolve, reject) => {
       const maxRandomRotate = o.randomRotation || 0;
       const rotation = Math.floor(Math.random() * maxRandomRotate) - (maxRandomRotate / 2);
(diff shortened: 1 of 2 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1869/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1869-pr-test`.

ArnoldSmith86 (commented):
The rest of the code usually writes it like this: `$('#buttonInputCancel').style.display = "initial"` (including your grid code).

96LawDawg:
> The rest of the code usually writes it like this: `$('#buttonInputCancel').style.display = 'initial'` (including your grid code).

Code consistency is not going to be my strongpoint.  But it may be a little strong to say the "rest" of the code.  I copied this syntax from editmode.js lines 55-75.  But I like your syntax better anyway.

> Hmm, also I would expect the Go button to still be on the right. Might be enough to set `visibility` instead of `display` (needs different values though). `visibility` makes the element still take up its usual space but be invisible (and non-interactive).

I couldn't figure out how to do that.  Visibility fixes it.  Changed.

ArnoldSmith86 (changes requested):
On production, the button only shows an icon if you only specify an icon. Here it shows "Cancel". Not sure why yet.

client/js/widgets/widget.js
```diff
@@ -2330,9 +2331,10 @@ export class Widget extends StateManaged {
       if(!o.confirmButtonText && !o.confirmButtonIcon){
         confirmButtonText = "Go";
       }
-      if(!o.cancelButtonText && !o.cancelButtonIcon){
+      if (o.cancelButtonText === null && o.cancelButtonIcon === null) {
+        $('#buttonInputCancel').style.visibility = "hidden";
+      } else (!o.cancelButtonText && !o.cancelButtonIcon)
```
ArnoldSmith86, line 2336: This needs to be `else if`.

ArnoldSmith86 (approved):
:+1:

client/js/widgets/widget.js
96LawDawg: Good catch.  I wish I knew how to code.  Thanks for putting up with me.

PR #1870, opened 2023-10-22
added `fillTo` to `MOVE`

ArnoldSmith86:
A new parameter for `MOVE` that makes it possible to fill cards to target holders/hands up to a defined value.

I am not sure how this should work with `count` and `fillTo`. The way I did it now, `fillTo` defaults to `null` and if it is set, it basically overrides `count`. A different way to do it would be to make `fillTo` boolean and use `count` for the count but then you would have to set two parameters every time you want to use `fillTo`.

### DIFF
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -902,7 +902,7 @@ function jeAddCommands() {
   jeAddRoutineOperationCommands('IF', { condition: null, operand1: null, relation: '==', operand2: null, thenRoutine: [], elseRoutine: [] });
   jeAddRoutineOperationCommands('INPUT', { cancelButtonIcon: null, cancelButtonText: "Cancel", confirmButtonIcon: null, confirmButtonText: "Go", fields: [], header: "" } );
   jeAddRoutineOperationCommands('LABEL', { value: 0, mode: 'set', label: null, collection: 'DEFAULT' });
-  jeAddRoutineOperationCommands('MOVE', { count: 1, face: null, from: null, to: null });
+  jeAddRoutineOperationCommands('MOVE', { count: 1, face: null, from: null, to: null, fillTo: null });
   jeAddRoutineOperationCommands('MOVEXY', { count: 1, face: null, from: null, x: 0, y: 0, snapToGrid: true, resetOwner: true });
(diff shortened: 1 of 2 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1870/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1870-pr-test`.

AbdullahWins (commented):
Good Job!

96LawDawg:
I think you made the right choice on how `fillTo` and `count` work together.  (But see my comment below).

Question about what we should do when there are not enough cards to fulfill the request and there are multiple seats/holders that need to be filled.  What happens in this PR is the same thing that currently happens in count: seats/holders are filled in turn  to the number specified in `count` or `fillTo`, then the next seat is filled, and so on until running out of cards.

I cannot figure out the order in which the request is filled.  It doesn't seem to be alphabetical, based on the order in which they are listed in the `to` property or the order in which the seats were created.  Ultimately, that probably doesn't matter too much except that the same player would always get the last cards if this comes up a lot during the game.

The bigger question is should we `fillTo` (and also `count`) so that cards are evenly distributed as much as possible until running out.  There are probably games in the real world that use both mechanics.  So ideally, we would support both.  So now I can see 'fillTo' being like a mode with options for (these names are terrible): "even", or "grouped".  Where even deals one card at a time to each seat and grouped deals the number specified in `count` to each seat before moving on to the next.

Sorry to make this complicated.

ArnoldSmith86:
On PCIO it's an option. I'll add it in a separate PR.

96LawDawg:
> On PCIO it's an option. I'll add it in a separate PR.

As a separate parameter?  That seems like extra work for the designer.  I would approve this based on testing, but just want to see what you had in mind as to how that option would be implemented.

ArnoldSmith86:
Yeah. PCIO has it. So either add a parameter to `MOVE` or make the importer do weird stuff. I vote for the former.

96LawDawg (approved):
Ok.  This is approved by test.  It works as expected with the limitations I noted.

PR #1871, opened 2023-10-22
added `INPUT` type `choose`

ArnoldSmith86:
https://test.virtualtabletop.io/PR-1871/pr-test

This supersedes #1456. A new `INPUT` type `choose` that can be used in a routine to let the player pick a widget or face.

To Do:

- [x] Make it work with INPUT
- [x] Multiple select with radio buttons (like palette)
- [x] Max widgets selected. Max can be 0 so allows view without choosing
- [x] Min widgets selected
- [x] Preselect widget based on id(s) specified in `value` parameter
- [x] Faces mode to display possible faces of one widget
- [x] Implement `holder` as an alternative widget source (also supporting hands)
- [x] Put selected widgets into a `collection`
- [x] Add `propertyOverride` (name suggestions?) so that you can override widget properties for rendering in the overlay (an example would be setting `activeFace: 1`)
- [x] Anonymous collections
- [x] Widgets should probably have a **maximum size** (which is a bit tricky because it needs to use scaling instead of resizing but that doesn't play nice with the layout without doing tricky stuff)
- [x] The **selection border** and check mark are still broken but the widget scaling should probably be done first
- [x] Consolidate `renderWidget` with edit mode
- [x] Tutorial
- [x] Implement this in Don't Blow Up for peeking at the top cards and for picking card out of discard
- [x] Check if this is useful in Peak Performance
- [x] Extensively test the other `INPUT` types because I cleaned up the code that gets their values a bit
- [x] Also test `MOVE.fillTo` because I moved the code that retrieves children of a seat

----------------------------------------------
### INPUT

The entire INPUT function should be replaced with the following:

This function shows an overlay with input controls so you can ask the user for input. If the user presses the confirm button, the routine will continue and user input will be stored in the designated variables.  If the user presses the cancel button, the routine stops complete there and no further commands in that JSON routine are executed. If both `cancelButtonIcon` and `cancelButtonText` are explicitly set to `null`, then the cancel button will be hidden and players cannot exit out of the overlay without making a selection or using the escape key.

Parameters:

* `cancelButtonIcon`: _name of VTT-Symbol_ - specifies an icon to be displayed on cancel button (defaults to null).  Must match name of a [VTT-Symbol](https://github.com/ArnoldSmith86/virtualtabletop/wiki/Fonts#vtt-symbols). If used simultaneously with text, appears before text.
* `cancelButtonText`:  _text_ - text shown in cancel button (defaults to "Cancel").
* `confirmButtonIcon`: _name of VTT-Symbol_ - specifies an icon to be displayed on confirm button (defaults to null).  Must match name of a [VTT-Symbol](https://github.com/ArnoldSmith86/virtualtabletop/wiki/Fonts#vtt-symbols). If used simultaneously with text, appears before text.
* `confirmButtonText`: _text_ - text shown in cancel button (defaults to "Go"). Can be used simultaneously with icon.
* `header`: _text_ - specifies the header text displayed at the top of the overlay. Appears above a title, if used. Cannot be customized with css. Mainly retained for backwards compatibility.
* `fields`: _array of field definitions_ - specifies what `type` of input field to display and where to store the returned input if any.

The field types are listed below, along with field-specific parameters. In addition, there are parameters that are common to a group of fields; these are described below.
* `type`: 
  * `checkbox`: _displays a checkbox_ - stores true or false
  *  `choose`: _displays images or faces of a widget_ - stores result as variable and as a collection
     * ` collection`: _text_ - the name of collection the widgets will be stored in
     * `faces`: _array_ - if the mode is set to "faces", then this array limits the faces shown to the player
     * `holder`: _`id` or array of `id`s_ - widgets in this holder will be displayed for players to choose from
     * `max`: _number_ - the maximum number of selections the player can make
     * `mode`: _"widgets" or "faces" (defaults to "widgets) - specifies whether individual widgets or the faces on widgets will be displayed for players
     * `propertyOverride`: _object_ - used to set a specific property/properties while the widgets are displayed. For example: `{"activeFace": 1}` can be used to show the face up side of cards that are face down in the room
     * `scale`: _number_ - alters the size of the widgets as they are displayed in the overlay.
     * `source`: _`id` or array of `id`s or a named collection_ - widgets to be displayed for players to choose from. If both `source` and `holder` paramaters are included, the `source` parameter will be ignored.
  * `color`: _displays a color picker_ - stores a hex color code (e.g. for coloring gaming pieces or player colors). Note that the appearance of the color picker may vary among browsers and/or operating systems.
  * `number`: _displays a number input field_ - stores numeric value
    * `min`: _number_ - the lowest allowed value for a number input field using the built-in scrolling tool; users can type a lower value that will be stored in the variable
    * `max`: _number_ - the highest allowed value for a number input field using the built-in scrolling tool; users can type a higher value that will be stored in the variable
  * `palette`:_displays colored radio buttons_-stores the selected value. Unlike the `color` field, this offers a consistent viewing experience across browsers and forces the players to choose from a curated palette of options.
     * `colors`: _array_ - displays the color choices available to the player. Accepts most color formats (html-friendly, hex, rgb, rgba, hsl, hsla).
   * `select`: _displays a dropdown_ - stores a string. See below for more detailed information about `select`.
     * `options`: _array_ - the selectable options for a `select` dropdown:
       * `value`: _string_ - the value that will be set to the variable when this option was selected by the player
       * `text`: _string_ - the text that will be shown to the player for this option (the entry in the dropdown) - if omitted the `value` will be displayed instead
       * `select`: _boolean_ - if set to _true_, that option will be the default value, and will be displayed in the input dialog.
  * `string`: _displays a text input field_ - stores a string
  * `subtitle`: _displays text_ - displays text in the overlay to provide instructions or context. Can also be placed anywhere to provide visual variety and can be customized with css.
  * `switch`: _displays a slider switch_ - stores "on" or "off"
  * `text`: _displays text_ - displays text in the overlay to provide instructions or context. May not be customized with css.
  * `title`: _displays text_ - displays text in the overlay to provide instructions or context. Can also be placed anywhere to provide visual variety and can be customized with css.

For those field types that accept input and return a value (`checkbox`, `choose', `color`, `number`, `string`, `select`, and `switch`), the following additional parameters can be used.
  * `label`: _string_ - to be displayed next to the input field
  * `value`: _various_ the initial value for the input field (for `checkbox`: true = checked, false = not-checked; defaults to false; for `choose` the widget that is selected when the overlay opens; for `switch`: left = off, right = "on"; defaults to "off"; for `select`, specifies the initial value of the dropdown)
  * `variable`: _variable name_ - specifies the variable in which to store the value/result.

For the `select` field, there is also an `options` field that defines the possibilities for what the user can scroll and select. The `options` field is an array of objects, and each object specifies an entry in the dropdown. An entry consists of a `value`, which specifies the string to be placed in the `variable` if that option is selected, together with the `text` to be displayed for that option. In addition, you can indicate which option will initially be displayed by declaring `"selected": true` inside that option. Note that if both `"selected": true` and `value` are used to define the starting option, then the information in `value` overrides `"selected": true`. The following example helps illustrate the point except that you would normally only use either `"value": "en-US"` or `"selected": true` and not both unless you use a dynamic property as the value and programmatically override the selected option:

```JSON
{
  "func": "INPUT",
  "fields": [
   {
      "type": "select",
      "label": "Choose language",
      "variable": "language",
      "value": "en-US",
      "options": [
        {
          "value": "en-US",
          "text": "English",
          "selected": true
        },
        {
          "value": "de-DE",
          "text": "German"
        }
      ]
    }
  ]
}
```

For the remaining field types (`subtitle`, `title`, and `text`), the only parameter is
  * `text`: _string_ - the text that will be shown to the player.

A complete example of an input statement follows:
```JSON
{
  "func": "INPUT",
  "fields": [
   {
      "type": "title",
      "text": "A header text for the overlay.",
      "css": "background-color:var(--VTTblue);color:white;"
    },
    {
      "type": "text",
      "text": "Stay a while and listen!"
    },
    {
      "type": "checkbox",
      "label": "Do you want to shuffle?",
      "variable": "shuffle",
      "value": true
    },
    {
      "type": "choose",
      "label": "Pick a dice face",
      "source": [
        "dice1"
      ],
      "mode": "faces",
      "variable": "face"
    },
    {
      "type": "color",
      "label": "Which color would you like?",
      "variable": "myColor",
      "value": "${playerColor}"
    },
    {
      "type": "number",
      "label": "Which number do you like?",
      "variable": "myNumber",
      "value": 10,
      "min": 1,
      "max": 100
    },
    {
      "type": "string",
      "label": "Credit card?",
      "variable": "myCard",
      "value": "555-2483 63"
    },
    {
      "type": "select",
      "label": "Best Crypto",
      "variable": "crypto",
      "options": [
        {
          "value": "DOGE",
          "text": "Dogecoin"
        },
        {
          "value": "BTC",
          "text": "Bitcoin"
        }
      ]
    },
   {
     "type": "subtitle",
     "text": "Include Expansions:",
     "css": "background-color: purple;color:white;"
    },
    {
     "type": "switch",
     "label": "Rise of Khan",
     "variable": "includedeck",
     "value": "on"
    }
  ],
  "cancelButtonIcon": "cancel",
  "cancelButtonText": "Cancel",
  "confirmButtonIcon": "restart_alt",
  "confirmButtonText": "Restart Game"
}
```

### Additional Updates:

**Add the following to the end of FLIP:**

If the holder parameter is a seat, then only the widgets belonging to that seat will be flipped.

**Add the following to the end of ROTATE:**

If the holder parameter is a seat, then only the widgets belonging to that seat will be rotated.

**Add the following to the end of SHUFFLE:**

If the holder parameter is a seat, then only the widgets belonging to that seat will be shuffled.

**Add the following to the end of SORT:**

If the holder parameter is a seat, then only the widgets belonging to that seat will be sorted. For example, you can put a SORT button near a hand that has the following JSON (using the predefined variable `${seatID}` which is the `id` of the seat of the player pressing the button) and only the player pressing the button will have their cards sorted:

```
{
    "func": "SORT",
    "holder": "${seatID}",
    "key": "rank"
}
```


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1871/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1871-pr-test`.

96LawDawg:
This is a good start.  What is there works.  Will need some sort of escape option.  Right now if there are 0 widgets in the chooser, you have to refresh the room.  But that can be added when we add some UI aspects to make it prettier.

96LawDawg:
Made an attempt to make the chooser UI look better.  The buttons do not do anything.  They are just placeholders.  I think we should make the widgets individually selectable (like in the edit mode UI with cards or resizing images).  That way a person doesn't accidentally click on a widget and think maybe they are flipping it over or something and that triggers the selection.  A confirm button is probably a good idea in every case.

ArnoldSmith86:
Yeah, it's needed anyway for multi-select.

96LawDawg:
I played around with trying to make the chooser overlay partially transparent.  I thought sometimes you might need to see what is in your hand or on the board while making a selection.  But that colored artifacts around the widgets (like square corners for cards and rough outlines of pawns).  Probably a good idea to at least consider that or maybe make the chooser overlay only 75% of the room width/height and make it box that can be dragged around so players can see what they need to see.

ArnoldSmith86:
I wanted to make it look similar to the `INPUT` overlay. Speaking of which: do we agree that this fits better as its own function or should this just be an input type for `INPUT`?

96LawDawg:
This should be its own.  It will be too confusing to try and work out all the permutations (in the engine and in everyday game devleopment) if we combine them.

mousewax:
I would make it part of the input overlay. I can think of a bunch of cool things we can do with seats and icons with that in the input. It's also a consistent UI and process for developers. I think makes things actually easier on developers by adding to an existing function instead of a wholly new one. It can even use the same "selected" CSS from the palette selection (outline with the checkmark in the corner).  There are multiple games in my own collection that would really benefit from having this kind of integration in the input.

This would also allow a way to choose things based on an image of something in the inputs. "name your faction" and "choose a flag".  The flexibility would be insane.
```
"fields": [
   {
      "type": "choose",
      "label": "Pick some Cards",
      "collection": "pickedCards" | "holder": "chooseHolder",
      "multiple": true
    }
  ]
```
then later:
```
  {
    "func": "SET",
    "collection": "pickedCards",
    "property": "parent",
    "value": "resultHolder"
  }
''''

Barring that: yes, always confirm in every case.

**Comment by 96LawDawg** [comment]
I'm fine with putting it as part of INPUT.  No strong feelings either way.  And it would simplify putting the whole div and css stuff together.

**Comment by ArnoldSmith86** [comment]
Thanks, @mousewax! The reason I wanted it as its own function was because I was thinking about letting the game dev split the widgets into rows (per owner or something) and other stuff and I thought that would be too complex to be inside `INPUT`. But now that I thought about it some more, you can just use multiple `INPUT` fields instead and get row titles for free. Same for other complexities like picking widgets _and_ faces (something like picking the faces of 3 different dice). Can just be 3 fields then.

**Comment by 96LawDawg** [comment]
I know this is an early draft, but just letting you know that the results from the INPUT are not being put into a collection.  They are being put into an `undefined` object.  This is from the new Choose from Holder button in the test room.  I put 3 cards in there.  Picked 2 and got undefined.

Screenshot 2023-10-23 083413

**Comment by ArnoldSmith86** [comment]
I don't think `INPUT` has any concept of mandatory input fields yet, right? So implementing a minimum number of selected widgets might be a bit out of scope for this PR.

**Comment by 96LawDawg** [comment]
> I don't think `INPUT` has any concept of mandatory input fields yet, right? So implementing a minimum number of selected widgets might be a bit out of scope for this PR.

That's mostly correct.  BUT, in the following `field` entry, the scroll bars only go up and downe from 1-6.  You can type a number out of range (like 10), but the variable will only return the appropriate min or max value.

```
{
       "type": "number",
       "label": "Enter a number",
       "variable": "number1",
       "value": 4,
       "min": 1,
       "max": 6
},
```

Maybe what we could do is run a check when the Confirm button is pressed.  If there are too few or too many chosen widgets highlighted, then have the button do nothing.

**Comment by ArnoldSmith86** [comment]
Maybe. But it could also wait for the next PR.

**Comment by 96LawDawg** [comment]
So far, so good.  Everything checked off in the to-do list as of now works as expected (Make it work with INPUT, Multiple select with radio buttons (like palette), Max widgets, Preselect widget based on id(s) specified in value parameter, Put selected widgets into a collection, `propertyOverride`, anonymous collections)

Also, nothing is broken (yet) in existing `INPUT`.

**Comment by ArnoldSmith86** [comment]
Nice. Also test adding dice please. That uses input stuff. 😀

**Comment by mousewax** [comment]
I have nothing technical to add. 
image

**Comment by 96LawDawg** [comment]
> Nice. Also test adding dice please. That uses input stuff. 😀

Adding all types of dice still works as normal.

**Comment by 96LawDawg** [comment]
I added support for min number of cards.  This is one way of doing it.  If you don't like it, revert the commit.

If min = max (say 2), you select 2 cards. You can still deselect one and then select another because otherwise you could not change.

If min < max, you can change your selection by selecting and deselecting, but you cannot go above max or below min.

If min > max then the designer is stupid, but that locks in the selection and cannot be undone without Cancel and starting over.

I personally like this and feel it is intuitive.  But whatever.  Also, it is possible to end up with less than the minimum by not selecting enough.  That isn't enforced.  A way to enforce that would be by checking the count when pressing Confirm button and prompting user to select the minimum.

**Comment by ArnoldSmith86** [comment]
Shouldn't the Peek in Don't Blow Up use the new way to hide the Cancel button?

**Comment by ArnoldSmith86** [comment]
**The value that is put into the resulting variable is not always the same:**

```JSON
        {
          "type": "choose",
          "source": [ "dice1" ],
          "mode": "widgets",
          "variable": "widget"
        }

=>

"dice1"
```

Mode `widgets` and `max == 1` will give you the `id` of the selected widget (or `null` if nothing was selected).

---

```JSON
        {
          "type": "choose",
          "source": [ "dice1" ],
          "mode": "widgets",
          "variable": "widget",
          "max": 2
        }

=>

[ "dice1" ]
```

Mode `widgets` and `max != 1` will give you the `id` of the selected widgets in an array (empty array if nothing was selected).

---

```JSON
        {
          "type": "choose",
          "source": [ "dice1" ],
          "mode": "faces",
          "variable": "face"
        }

=>

"0"
```

Mode `faces`, one widget in a dynamic `source` collection and `max == 1` will give you the selected face index directly (or `null` if nothing was selected).

---

```JSON
        {
          "type": "choose",
          "source": [ "dice1" ],
          "mode": "faces",
          "variable": "face",
          "max": 2
        }

=>

[ "0", "1" ]
```

Mode `faces`, one widget in a dynamic `source` collection and `max != 1` will give you the selected faces in an array (empty array if nothing was selected).

---

```JSON
        {
          "type": "choose",
          "source": [ "dice1", "dice2" ],
          "mode": "faces",
          "variable": "face",
          "max": 2
        }

=>

  [
    {
      "widget": "dice1",
      "face": "3"
    },
    {
      "widget": "token1",
      "face": "0"
    }
  ]
```

Mode `faces`, any `source` collection that could be more than one widget (so a collection name or an array with more than one `id`) and `max != 1` will give you the selected faces in an array of objects (empty array if nothing was selected).

---

We could _always_ return the format in the last example but that would make 90% of use cases way more complicated to work with than necessary.

Any opinions about this? (And can somebody test that all of this is what actually happens?)

ArnoldSmith86:
Now that `holder` works, this should be feature-complete. At least what I had in mind.

Please play with it, try to incorporate it into games and give feedback.



> If min < max, you can change your selection by selecting and deselecting, but you cannot go above max or below min.

Feels a bit weird to me that you can't deselect below `min`. Makes sense to ensure a valid selection but doesn't feel great. I'll try to "properly" implement it by not letting the user submit the form if it doesn't meet the requirements. Also: you removed the special case where `max: 1` auto-deselects the previously selected one to act more like radio buttons (at least I think you did).

ArnoldSmith86:
The changes in the seat logic (idea is from #1658) made it trivial to support `SORT` and `SHUFFLE` for seats so I included those changes here. Probably not the best idea but it changed the behavior anyway so I thought I'd fix what was left.

96LawDawg:
> Feels a bit weird to me that you can't deselect below `min`. Makes sense to ensure a valid selection but doesn't feel great. I'll try to "properly" implement it by not letting the user submit the form if it doesn't meet the requirements. Also: you removed the special case where `max: 1` auto-deselects the previously selected one to act more like radio buttons (at least I think you did).

No worries.  If you think it feels strange, then revert it.  And I did unknowingly remove the max: 1 special case.

96LawDawg:
> The changes in the seat logic (idea is from #1658) made it trivial to support `SORT` and `SHUFFLE` for seats so I included those changes here. 

I love being able to do this (instead of selecting by owner):

```
"clickRoutine": [
    {
      "func": "SORT",
      "holder": "${seatID}",
      "key": "rank"
    }
  ]
 ```

**Comment by 96LawDawg** [comment]
> We could _always_ return the format in the last example but that would make 90% of use cases way more complicated to work with than necessary.
> 
> Any opinions about this? (And can somebody test that all of this is what actually happens?)

Tested.  Confirmed.  Leave as is.

For the first two examples you gave, that is what I would expect in widget mode.  Either the id as a scalar value or as an array if making multiple selections.  And same for the second two examples.  Only the last example is a little strange (but expected).  Most use cases will not require multiple faces of multiple widgets.  Those that are doing that can figure out how to parse through the resulting object.  Do not make that the default.  Probably a better practice would be to have a separate field for each widget in that scenario.

I've also tested holders/seats.  Works as expected.

**Comment by 96LawDawg** [comment]
To fulfill @mousewax's dream of being able to pick a logo for your team, should we support a source that is not a widget but just an asset/image URL?  As this PR is right now, I would build a widget with 10 faces and let the user pick one of those faces to be assigned as the logo.  But I don't really want the face of the widget, I want to know what the image URL is so I can put that on the seat.  If we just had an array of image URLs as the source (and then as the output variable), I wouldn't have to build a widget I'm not going to use and wouldn't need code to figure out the URL based on the face.

File: client/js/jsonedit.js [hunk 1]
```diff
@@ -944,14 +944,23 @@ function jeAddCommands() {
...
   jeAddFieldCommand('max', 'number', 10);
   jeAddFieldCommand('options', 'select', [ { value: 'value', text: 'text' } ]);
 
+  jeAddFieldCommand('source', 'choose', 'DEFAULT');
+  jeAddFieldCommand('collection', 'choose', 'DEFAULT');
+  jeAddFieldCommand('holder', 'choose', '');
+  jeAddFieldCommand('max', 'choose', 1);
+  jeAddFieldCommand('min', 'choose', null);
+  jeAddFieldCommand('mode', 'choose', 'widgets');
+  jeAddFieldCommand('faces', 'choose', null);
```
96LawDawg, line 961: What does the `faces` parameter do?

96LawDawg:
I added a tutorial as a new `choose` variant.  Subject to change if this PR changes some. But this base should stay basically the same.

I updated chess with the choose INPUT.  There are 2 other games I made that would benefit from this, but they are not popular games and it would be a good bit of effort to re-work them so I'm not going to do that now.  But I've tweaked 2 existing games and made a tutorial.  This PR is in pretty good shape functionally.  Would still benefit from cleaning up the check marks and limiting the size of the widgets as listed in the to-do.

96LawDawg:
@ArnoldSmith86, I tried to revert my changes that took your "max" code and also made it a "min" check.  But I butchered that.  I decided against trying to fix it any more.  Sorry.

96LawDawg:
Actually, I think I figured out what went wrong with my revert.  Seems okay now.

96LawDawg:
FYI, if both `source` and `holder` parameters are specified, `source` is ignored.  I'm fine with that.  I've put it in the draft wiki entry at the top of the page.  Just pointing that out in case anybody wants to make a case for the other way.

96LawDawg:
> My thoughts on that were to have some cards or basic widgets in a holder and then move the one you choose to a holder in your seat or some junk. Then it wouldn't be a choice for other players.

That won't work perfectly if the players are picking logos at the same time.  I can't think of any way to prevent that.

mousewax:
> That won't work perfectly if the players are picking logos at the same time. I can't think of any way to prevent that.

I was thinking hide the button when someone clicks on it until they finish their selection.

Or, if they chose something someone already chose it just wouldn't get moved and they'd have to click the button again to pick a different thing.

client/js/jsonedit.js
ArnoldSmith86: Right... Limit the selectable faces. Forgot to implement that..

ArnoldSmith86:
> FYI, if both `source` and `holder` parameters are specified, `source` is ignored.  I'm fine with that.  I've put it in the draft wiki entry at the top of the page.  Just pointing that out in case anybody wants to make a case for the other way.

That's what most operations do, right?

mousewax:
When choosing a card, is there an option to have which face you see? Useful for picking a card to steal without seeing them.

96LawDawg:
> Yes. `propertyOverride`. Not sure if we should add another parameter just for the face because it should be pretty common.

Please don't.  We already have `mode: faces`, and `faces`.  I think this is clear enough and flexible without duplicating functionality.

ArnoldSmith86:
Should the default for `min` be 0 or 1?

96LawDawg:
> Should the default for `min` be 0 or 1?

I think 0.  Usually, you want someone to pick at least one thing, so that will be more common. But I think we shouldn't go overboard on rules enforcement unless the designer asks for it by specifying a min.  Also, this way, designers won't have to remember to set it to 0 if they just want to show the widgets.

ArnoldSmith86:
For the widget size, I implemented a new `scale` property that gets applied on top of the room scale. I feel like that's a better solution than just defining a maximum size and scaling any widgets that are bigger. This way the game designer can better choose how the widgets will look. They will look as big as in the room by default (unless they are scaled there) and you can change the size with the `scale` factor.

ArnoldSmith86:
> For example, you can put a SORT button near a hand that has the following JSON and only the player pressing the button will have their cards sorted

@96LawDawg That does not sound correct. The button will sort the hand of the player sitting in seat `seatID`, no matter who pressed it.

96LawDawg:
> > For example, you can put a SORT button near a hand that has the following JSON and only the player pressing the button will have their cards sorted
> 
> @96LawDawg That does not sound correct. The button will sort the hand of the player sitting in seat `seatID`, no matter who pressed it.

It does work.  I tested it.  It works because `seatID` is a predefined variable of the `id` of the player pressing the button.  I guess you forgot that was a predefined variable.

ArnoldSmith86:
I did.

96LawDawg:
> I did. Most people reading that probably don't know either.

Which is why I just updated the draft wiki explanation to say that it is a predefined variable.  That's one of the reasons I like this so much.  It shortens a bit of code I use a lot (I like providing sort buttons for hands).

96LawDawg:
I have tested scale, the new selection board, and min.  They all work great.  One question:

Should/could we support enlarge in the overlay?  If you make the scale small to fit more things on the screen, it might be nice if enlarge worked.  It currently does not.  That is not a deal breaker for me.  I will approve this PR as soon as I update the tutorial with scale and min, and Don't Blow Up (going to add scale there because the cards are so big).

ArnoldSmith86:
It doesn't? Was working at one point but maybe I broke it while introducing a wrapper div.

96LawDawg:
> It doesn't? Was working at one point but maybe I broke it while introducing a wrapper div.

It does not work.  It did work.  `git bisect` says this is the problem commit: https://github.com/ArnoldSmith86/virtualtabletop/pull/1871/commits/7b9cf026aaba0eb00a034a395c8113fcdc43ff8f

96LawDawg (approved):
Approved through extensive testing, tutorial development, and game updates.  Draft wiki entries are ready to go.  This is ready to be merged.

PR #1874, opened 2023-10-24
card text layers default to black color

ArnoldSmith86:
Previously the default was to not set a color at all which made it inherit the color of its parent. Which leads to transparent text in holders or something.

### DIFF
--- a/client/js/widgets/card.js
+++ b/client/js/widgets/card.js
@@ -165,7 +165,7 @@ class Card extends Widget {
             objectDiv.srcdoc = html;
           } else {
             objectDiv.textContent = object.value;
-            objectDiv.style.color = object.color;
+            objectDiv.style.color = object.color || 'black';
           }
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1874/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1874-pr-test`.

96LawDawg (approved):
Approved through test.  This fixes a problem I've encountered many times.  I never thought to make an issue.  I always just added the "color" property onto the cards, but this PR should be the solution.

The demo room shows how this works.  Drag one of the cards (they don't have the "color" parameter set) into the holder.  On production, they inherit the css from the holder and the text turns red.  In this PR, the text stays black.

PR #1879, opened 2023-10-27
prevent undoing initial state load

ArnoldSmith86:
When pressing the Undo button in the editor toolbar, you can undo the intial room state loading which completely removes everything from the room.

I want to improve more things about Undo and auto saving but I feel like we should fix this pretty quickly because it is very destructive.

### DIFF
--- a/client/js/editor/toolbar/undo.js
+++ b/client/js/editor/toolbar/undo.js
@@ -6,7 +6,7 @@ class UndoButton extends ToolbarButton {
   click() {
     const protocol = [...getUndoProtocol()];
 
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1879/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1879-pr-test`.

96LawDawg (approved):
Approved thru test and code review.

PR #1880, opened 2023-10-28
Fix problem from setting card text layers to black by default

96LawDawg:
Fixes a problem created by #1874.

If a text layer use something like ` "css": "color:darkgray",` but did not just directly specify `'color": "darkgray"` as was the case in Shattered Dominions, the card text was being set to black and the css was being ignored.  This PR adds a check to see if a css color has already been set and then will not apply the default black.

### DIFF
--- a/client/js/widgets/card.js
+++ b/client/js/widgets/card.js
@@ -165,7 +165,9 @@ class Card extends Widget {
             objectDiv.srcdoc = html;
           } else {
             objectDiv.textContent = object.value;
-            objectDiv.style.color = (object.css && object.css.includes('color')) ? object.css.split('color:')[1].trim() : (object.color || 'black');
+            objectDiv.style.color = object.color;
+            if(!objectDiv.style.color)
+              objectDiv.style.color = 'black';
           }
           return usedProperties;
         }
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1880/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1880-pr-test`.

ArnoldSmith86 (changes requested):
That condition would also trigger for something like `border-color` or `background-color`. Can be checked with a regular expression. I'll do that when I have some time.

Also: The applied CSS is already in variable `css`. Should probably check that one instead.

The easiest solution is probably to read the current `objectDiv.style.color`. Its value should probably be `inherit` if nothing is set explicitly? Gotta check.

ArnoldSmith86 (approved):
I changed the logic. Please have another look. :)

96LawDawg:
That's a lot cleaner and easier to understand.  And it works.  Ready to merge.

robartsd:
It probabaly would have been better to just set the default color in a stylesheet instead of as an inline style; perhaps implemented with a new class for face object text divs. Might not be worth changing at this point though.

ArnoldSmith86:
Yeah, that sounds like the right idea.

PR #1882, opened 2023-10-28
added `regex` to `INPUT.string`

ArnoldSmith86:
Now you can define a regular expression to match an `INPUT` string against and use `regexHint` to tell the player about it if the input does not match it.

### DIFF
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -951,6 +951,8 @@ function jeAddCommands() {
   jeAddFieldCommand('min', 'number', 0);
   jeAddFieldCommand('max', 'number', 10);
   jeAddFieldCommand('options', 'select', [ { value: 'value', text: 'text' } ]);
+  jeAddFieldCommand('regex', 'string', '');
+  jeAddFieldCommand('regexHint', 'string', '');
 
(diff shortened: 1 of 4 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1882/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1882-pr-test`.

96LawDawg:
Will add this to the wiki entry for INPUT under string:

* `string`: _displays a text input field_ - stores a string. Optionally may add a Javascript regular expression
    * `regex`: _regex_ - will limit the user to the characters in the regular expression. For example, `^[0-9]$` forces an entry consisting only of numbers.
     * `regexHint`: _text_ - this is the prompt that players see if the response does not meet the regex criteria

96LawDawg (approved):
Approved through test.  Updated tutorial and drafted wiki entry.

ArnoldSmith86:
This was very easy to add after #1871. Any other ideas what INPUT could check when confirming the overlay?

PR #1888, opened 2023-10-29
added `collection` support to `MOVE`

ArnoldSmith86:
Trying to touch the code as little as possible to support `MOVE` with `collection` (without simply copy-pasting many lines of code).

Update wiki entry for MOVE:

## MOVE

This function moves widgets into a target holder. If the target of the move is an occupied _seat_, then the move will instead direct the widgets to the hand associated with the seat. In this case, if the hand is set to _childPerOwner_, the owner will be set to the player in the seat.

Parameters:
* from: _widgetID_ (or an array) - specifies the widget(s) that contains the widgets to move. In the typical case, this would be a holder, but could be any widget with child widgets. If `from` is not specified, then the "DEFAULT" collection will be moved.
* collection: _collection_ - specifies the collection that is to be moved (defaults to "DEFAULT"). When using a collection, omit the `from` parameter.
* to: _widgetID_ (or an array) - specifies the widget(s) that widgets should be moved into. In the typical case, this would be a holder or seats, but could be any widget.
* count: _number_ - limits the amount of moved widgets (defaults to 1 if omitted; use 0 to indicate no limit)
* fillTo: _number_ - fills the target holders/seats up to this number (defaults to null). If specified, then `count` is ignored.
* face: _number_ - optionally sets the face of the moved widgets to the given value (see FLIP). If omitted, the widgets will be left as they are.

Note that both `count` and `fillTo` will move an entire group to one of the `to` widgets. If there are enough widgets remaining in the `from` source, then it will move to the next destination. The order that the function picks targets for moving `to` is not well understood, so if there are less widgets in the `from` source than are required, game designers may want to account for that in the JSON in some other way.

If the `dropTarget` property (when moving to a holder) does not match the widgets being moved, the widgets will become children of the holder, but will keep the original x,y coordinates. In other words, they will not follow the stackOffset rules for aligning child widgets.

```JS
{
  "func": "MOVE",
  "from": "<holderID>",
  "to": "<holderID>",
  "count": 0
}
```


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1888/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1888-pr-test`.

96LawDawg:
This may be controversial, but I hope not.  I think that when "collection" is used, `MOVE` should default to moving the entire collection (so the default is 999999 or whatever).  You can override it by specifying a different "count" parameter.

ArnoldSmith86:
Yeah, the only problem is if you move to more than one target..

ArnoldSmith86:
But yeah, I think we should do that.

96LawDawg:
> Yeah, the only problem is if you move to more than one target..

Then maybe it is a more complicated IF condition in the engine for the default.  IF you specify a collection and IF you only have one destination holder then the default is all.

96LawDawg (approved):
Approved through testing and development of a tutorial.

96LawDawg (changes requested):
I approved the wrong PR.  This one is not approved yet.

RaphaelAlvez (changes requested):
Hey I don't really feel the right to block this PR. I'm just making a review to be sure that this is taken into account.

My routines PR early versions already do this. If we change the names it would be more inline with the current approach while being a first step into the new direction.

I know there is some extra code there but maybe it's worth making it be reviewed.

You can dismiss this review

ArnoldSmith86:
I'm not against better names but I would much prefer to separate implementing new functionality from name changes because renaming usually generates many trivial code changes that don't mix well with few impactful code changes of new features.

After this PR we'll reassess the state of your PR. We'll see if we also want `MOVEXY` with collections. Everything else (feature-wise) that you suggest is only for the sake of being consistent if you ask me. We'll have to check those cases to see if it's worth it (shouldn't be a big problem).

After typing all this, I realize that you probably only mean the new `collection` name. I don't really care either way but I feel like at the moment `collection` is more consistent with all the other routine operations.

96LawDawg:
I agree that we should add this functionality and then separately address the things Raphael has worked on.

One more question/thought about this PR for collections.  If you MOVE a collection that is in a holder (but without using "from") to another holder, this works as expected.  If you MOVE a collection that is not in a holder to a holder, it operates just like SET in that the x/y values could put the widget anywhere.  I can live with that.  But if I were a new user working with this the first time, I think I would expect MOVE to do something more than just change the parent.  It would use the logic of the receiving holder to either stack the widgets at 0,0, or would use stackOffsets.  Thoughts?

ArnoldSmith86:
It is intended to work the same no matter the source. I'll investigate why it doesn't.

ArnoldSmith86:
> If you MOVE a collection that is not in a holder to a holder, it operates just like SET in that the x/y values could put the widget anywhere.

You're talking about the pawns in the test room? You forgot to adjust the `dropTarget` in the destination holder.

96LawDawg:
> > If you MOVE a collection that is not in a holder to a holder, it operates just like SET in that the x/y values could put the widget anywhere.
> 
> You're talking about the pawns in the test room? You forgot to adjust the `dropTarget` in the destination holder.

Yes.  Oops.

96LawDawg:
I need to update the tutorial to implement this and fillTo.  Then I will approve.

ArnoldSmith86:
BTW, I changed last minute that it uses `collection: DEFAULT` if you don't specify either `from` or `collection`. Should be more like other operations this way.

So maybe re-test some stuff.

96LawDawg:
> BTW, I changed last minute that it uses `collection: DEFAULT` if you don't specify either `from` or `collection`. 

I saw that and had already retested, which is what led to my question about moving from the room and not a holder (which was a dropTarget problem).

RaphaelAlvez:
What I mean is that in that PR I had already added this Specific functionality. That was the first thing I did but there I used different names. We can very easily change those names.

I'm mostly afraid that your implementation could be incompatible with mine in the long run not naming wise but in structure (need to check, I'm really not sure)

RaphaelAlvez (changes requested):
you need an updateAfterShuffle

image

96LawDawg (approved):
I don't know if updateAfterShuffle() is needed or not.  As for the other comments, I don't really have an opinion.  What I do know is this works and has LONG been requested and needed.  Approved through test and tutorial development.

RaphaelAlvez:
> I don't know if updateAfterShuffle() is needed or not.  As for the other comments, I don't really have an opinion.  What I do know is this works and has LONG been requested and needed.  Approved through test and tutorial development.

Yeah as I said, my first review can be dismissed. I just wanted it to be taken under consideration. On that PR I added so many features that were requested for move and I wanted at least for both PRs to stay compatible.

As for the updateAfterShuffle() the name is misleading. It is the function that rearranges the z value of the cards in a holder in function of their x value. That's what you are seeing in the picture.

ArnoldSmith86:
> you need an updateAfterShuffle

Done. Good catch. I'm pretty sure you didn't add it in your PR either.

ArnoldSmith86:
> I'm mostly afraid that your implementation could be incompatible with mine in the long run not naming wise but in structure (need to check, I'm really not sure)

You changed the `MOVE` code quite a bit more. While you probably made it cleaner, I don't think it's worth it to risk breaking it. For example you removed the special seat code. It's probably not necessary anymore since your idea of using `Seat.children` to return the cards in the hand is in production now but stuff breaks sooo often..

RaphaelAlvez:
> > you need an updateAfterShuffle
> 
> Done. Good catch. I'm pretty sure you didn't add it in your PR either.

From a quick glance it looks like I removed it by mistake

ArnoldSmith86:
> From a quick glance it looks like I removed it by mistake

Maybe you added and then removed it again. The one occurrence you removed was in the special seat code that's not necessary anymore (see comment above).

ArnoldSmith86:
> you need an updateAfterShuffle

It's kinda funny btw. The problem is that the branch

```Javascript
          if(source == target) {
            await applyFlip();
            await c.bringToFront();
          }
```

doesn't call `updateAfterShuffle`. Even in production. I tried to force the problem in production but because you move the holder contents to itself, it corrects itself kinda on accident. I tried this:

```JSON
    {
      "func": "MOVE",
      "from": [
        "h1",
        "h2"
      ],
      "to": "h2",
      "count": 0
    }
```

Which should move all cards from `h1` to `h2` and _then_ bring all cards from `h2` to the top without calling `updateAfterShuffle`. But because `h1` was already moved to `h2`, it brings all the cards to the top again (in the correct order).

It might be possible to cause a problem in production when using seats though. Not sure.

I just added `updateAfterShuffle` to the `collection` branch of the code though to not change more than necessary.

96LawDawg:
Tested again after updateAfterShuffle.  Everything is working.

96LawDawg (changes requested):
Need an error catching piece.  If there is no "DEFAULT" collection when trying to move a collection, the room freezes.

RaphaelAlvez (approved):
small testing and code reviewd

ArnoldSmith86:
> If there is no "DEFAULT" collection when trying to move a collection, the room freezes.

Should be fixed.

ArnoldSmith86:
Looks like the tutorial still says that this is not possible even though it has an example now:

image

96LawDawg:
> Looks like the tutorial still says that this is not possible even though it has an example now:

Strange.  I completely rewrote that overview.  Don't know what happened to that.  But it is good enough as is after you fixed it.  Thanks.

96LawDawg (approved):

PR #1891, opened 2023-10-30
Public Library: Add Delicious Deliveries

andypymont:
Add public library game Delicious Deliveries (similar to Fresh Fish)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1891/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1891-pr-test`.

96LawDawg:
Comments/suggestions:

Short description says: "Can you have the most delicious deliveries of the very freshest goods?"  Maybe delete "very?"

I like how the logic force moves a piece to a nearby vacant area and doesn't allow stacking.  Of course, sometimes that doesn't work and you can just leave a tile on the board not in a blank space if there isn't a blank tile nearby.  No worries.  Just making an observation.

When you end the game, the scoreboard is movable.  Not sure if you intended that or not.  But when you start a new game, the scoreboard is still visible.  Probably want to hide it in the new game button code.

The how to use instructions for end game scoring are a bit misleading.  Clicking on individual cells does not activate that cell.  Clicking on the scoreboard activates the overall scoreboard input overlay.  It will pre-populate the player name correctly, but you have to select the correct category from the drop-down list.

I don't understand why there are 2 piles of cubes.  Probably obvious to someone that knows the game (I don't), but not clear to me reading the instructions in the room.

For end of game, I'm not sure how much work it is to manually fill in empty spaces with roads, or how much mental work it is to imagine there are roads there.  But you could as part of the score button, automate that.  Make a collection of each space holder, loop through it and if there are no widgets with that as a parent, move a road there.  Just an idea.

andypymont:
Thanks for the review!

> Short description says: "Can you have the most delicious deliveries of the very freshest goods?" Maybe delete "very?"

Updated.

> When you end the game, the scoreboard is movable. Not sure if you intended that or not. But when you start a new game, the scoreboard is still visible. Probably want to hide it in the new game button code.

I did intend the scoreboard to be moved when I wasn't sure if it would cover anything important, but I'm happy with where it appears now, so I've made it non-movable as you're suggesting.

> The how to use instructions for end game scoring are a bit misleading. Clicking on individual cells does not activate that cell. Clicking on the scoreboard activates the overall scoreboard input overlay. It will pre-populate the player name correctly, but you have to select the correct category from the drop-down list.

Yes, I didn't realise this actually until a recent playtest. Thanks for highlighting; I've added to the usage instructions.

> I don't understand why there are 2 piles of cubes. Probably obvious to someone that knows the game (I don't), but not clear to me reading the instructions in the room.

They are reserve cubes - they were mentioned in the part of the rules explanation where you gain access to them, but I've also added a note to the placing-a-cube action to be clear they're not available until later gained by placing a shop.

> For end of game, I'm not sure how much work it is to manually fill in empty spaces with roads, or how much mental work it is to imagine there are roads there.

It's not really any work - there are usually lots of roads on the board already by then, plus you've spent the whole game thinking about how close your shops are to their supplies, so it's not a big deal at all. I think it's easiest to leave this up to the players (especially vs. them potentially ending their game early by clicking Score by accident 😅)

96LawDawg (approved):

PR #1893, opened 2023-10-30
allow card face text color to be set at face level

ArnoldSmith86:
In #1874 we set the card object text color default to black because it was inheriting colors from parent widgets. In #1880 we fixed that #1874 would set the color to black even if the game designer set the color to a different color using the object css property.

Now this PR fixes the same issue for when the game designer set a default color for the entire face.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1893/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1893-pr-test`.

ArnoldSmith86:
I'll change this to @robartsd's suggestion. Makes way more sense.

96LawDawg (approved):
This does seem more intuitive.  And it works.  Tested it using my original test room, Team II, and Shattered Dominion.  All work fine.  Approved.

PR #1895, opened 2023-10-30
fixed `INPUT.choose` visualization inheriting undesired properties

ArnoldSmith86:
When rendering a copy of a widget, I deleted properties like `x` and `y` to make sure it is displayed where I want it to.

Unfortunately I did not think about `inheritFrom` so in some cases deleting `x` simply leads to inheriting it from somewhere else.

This PR explicitly sets these values to their defaults so they do not get inherited.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -2226,14 +2226,14 @@ export class Widget extends StateManaged {
 
   renderReadonlyCopyRaw(state, target) {
     delete state.id;
-    delete state.x;
-    delete state.y;
-    delete state.rotation;
-    delete state.scale;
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1895/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1895-pr-test`.

96LawDawg (approved):
Approved through test and code review.  Tutorial and 2 existing games work.  And this fixes @mousewax's use case through testing.

PR #1899, opened 2023-10-31
removed old editor code

ArnoldSmith86:
This removes the code of the old edit overlays for each widget type. It was not accessible anywhere anyway and this also removes the Vue.js dependency.

We still need a decent replacement in the new editor but I doubt that any of the old code will prove useful. And we still have it in the git history.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1899/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1899-pr-test`.

96LawDawg (approved):
Approved through test.  Everything in the Editor still works.

PR #1902, opened 2023-11-01
save features in the editor

ArnoldSmith86:
This adds ways to quickly save your progress in the editor.

- [x] Quick Save should probably not switch the active state to the new one.
- [x] Quick Download doesn't work yet.
- [x] Close the menu automatically at some point.
- [x] prohibit PL edits
- [x] fix testcafe
- [x] Save as New Variant to [name of current game]
- [x] Option to input name for Save as New Game

Changes to wiki:

 Save changes to the currently loaded game. This feature will open a submenu with several options that allow you to make changes to the game without leaving Edit Mode and going to the Game Shelf:
 - Save to Active Variant. This immediately saves the game  to the current variant. If a game has not yet been saved, this option is not available. This option is also not available for Public Library games.
 - Save as New Variant (of ...). The button name displays the name of the game you are currently working on. This immediately creates a new variant in that game. If a game has not yet been saved, this option is not available. This option is also not available for Public Library games.
 - Save as New Game. This button prompts you for the name of the new game and then saves it to the Game Shelf.
 - Quick Download. This button will download the JSON, and only the JSON and not the assets, of the current variant. This is NOT the method you should use for saving offline versions of games. This is a simple function for limited purposes.
 - Quick Save. This button will create (if it does not already exist) a saved version called "Quicksave". The variant name will be the date and time (UTC) of the save. Each additional press of the button will create a new variant with the new date and time.
 - Auto-save every 5 minutes. If checked, this button does the same thing the Quick Save button does, but does so automatically every 5 minutes while you are in the room AND whenever you close out of the editor. This can create many, many Quicksave variants.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1902/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1902-pr-test`.

RaphaelAlvez:
Damn, the saving animation is super cool! Simple but cool

Not sure what is quick save supposed to be...

ArnoldSmith86:
Just save it "somewhere" without overriding the current variant. Basically what autosave will be doing but manually.

Not sure yet. We can tweak the options. I just want to make it work for now.

ArnoldSmith86:
Apart from Quick Download the buttons (and checkbox) should work now. Gotta improve how the menu looks and feels but that's not super important. :grin:

96LawDawg:
The Save to Active Variant allows you to save a new version of a PL game that overwrites the original PL game with NO way to get back the original. Same for Save as New Variant.  It adds a new variant to the PL game, but there is no way to delete the variant because the hamburger menu doesn't allow it.  I think we should prohibit these actions on PL games.

The Save as New Game created a default name that seemed to be "New Game" plus the UTC time.  Maybe it should be New Game plus the existing name of the file if there is one.  Unless the name starts with New Game.  In either of those situation, New Game plus UTC time is fine.

The other things that were described as working seemed fine.

ArnoldSmith86:
Oops.

ArnoldSmith86:
One note: the quick download buttons creates a vtt file **without assets**. I wanted this to be as quick as possible and there's basically no way for you to accidentally delete assets so I feel like it's not useful to download the assets.

This does mean that the resulting vtt file can't be uploaded to a different server. Well it can, but assets will be broken unless they already exist there.

Should we tell the user?

RaphaelAlvez:
> One note: the quick download buttons creates a vtt file **without assets**. I wanted this to be as quick as possible and there's basically no way for you to accidentally delete assets so I feel like it's not useful to download the assets.
> 
> This does mean that the resulting vtt file can't be uploaded to a different server. Well it can, but assets will be broken unless they already exist there.
> 
> Should we tell the user?

can you put the whole link instead?

ArnoldSmith86:
> can you put the whole link instead?

You mean simply make it download the game including assets? Can be 100MiB..

RaphaelAlvez:
> > can you put the whole link instead?
> 
> You mean simply make it download the game including assets? Can be 100MiB..

I mean using "https://virtualtabletop.io/assets/1358566275_32201" instead of "/assets/1358566275_32201"

Shouldnt't this work to send assets between servers. and then the images tools in editor could put them locally

ArnoldSmith86:
That would work but I would prefer not to touch the state data because the game designer uses this as a backup function.

I could name the file `QuickDownload without assets ...`..

RaphaelAlvez:
> That would work but I would prefer not to touch the state data because the game designer uses this as a backup function.
> 
> I could name the file `QuickDownload without assets ...`..

when do you think this featre would be needed? It feels like thgeresuting file is almost useless or very volatile

ArnoldSmith86:
> when do you think this featre would be needed? It feels like thgeresuting file is almost useless or very volatile

How often did you download your games from virtualtabletop.io without the intention of uploading it to a different server?

96LawDawg:
I like where this is going.  A few comments.

1) I have no issues with the quick download, but I would add "no assets" to default file name.

2) If autosave is enabled, should we do an autosave before the player switches games?  For example, if I'm working on a new game and I have autosave, I might "accidentally" load a different game to copy some code.  It would be nice if autosave kicked in to prevent me from losing the changes in the last 4 minutes.

3) Here's the biggest issue I have.  You can't tell from the Edit mode which game/variant you have loaded.  For example, if I've got a game I'm working on and I think something went wrong, so I go to the Quicksave version.  I figure out what happened and make a few tweaks then I want to make a new variant.  I intended for that to go into the game I was working on, but it goes into the Quicksave.  A new user is going to have issues figure out where his stuff is.  Is there a way to include the information that this will (or has) been saved in the room to the game named: "X"?

4) For the save as New Game or New Variant, should we include an overlay that allows you to give the game or variant a meaningful name?  I know these buttons are intended as a shortcut to the main system on the game shelf.  But I think there is risk of confusion if we don't provide some additional information and/or opportunity to customize what is being saved.

RaphaelAlvez:
> > when do you think this featre would be needed? It feels like thgeresuting file is almost useless or very volatile
> 
> How often did you download your games from virtualtabletop.io without the intention of uploading it to a different server?

Maybe we are miss comunicating here. I think that going from one server to another (local or test servers) is the main use case of local files for me. I cannot remenber the last time I downloaded it for backups

Therefore I cannot see much use in a file that can only be used in the same server.

As I understand you quick save only downloads  the JSON making it useless for this purpose. I was suggesting a middle ground by instead of having the assets files, we add the full link for the server it was downloaded from.

RaphaelAlvez:
> 1. I have no issues with the quick download, but I would add "no assets" to default file name.

But do you have any use case for it?

> 2. If autosave is enabled, should we do an autosave before the player switches games?  For example, if I'm working on a new game and I have autosave, I might "accidentally" load a different game to copy some code.  It would be nice if autosave kicked in to prevent me from losing the changes in the last 4 minutes.

Please!! so many times someone comes into a room and do that by acident...

> 3. Here's the biggest issue I have.  You can't tell from the Edit mode which game/variant you have loaded.  For example, if I've got a game I'm working on and I think something went wrong, so I go to the Quicksave version.  I figure out what happened and make a few tweaks then I want to make a new variant.  I intended for that to go into the game I was working on, but it goes into the Quicksave.  A new user is going to have issues figure out where his stuff is.  Is there a way to include the information that this will (or has) been saved in the room to the game named: "X"?
I think the problem here is not having access to the gameshelf. 

> 4. For the save as New Game or New Variant, should we include an overlay that allows you to give the game or variant a meaningful name?  I know these buttons are intended as a shortcut to the main system on the game shelf.  But I think there is risk of confusion if we don't provide some additional information and/or opportunity to customize what is being saved.

yeah I think a pop up for the name would be very usefull. If it was pre filled based on the original file information it would be perfect. then we can only make some small changes.

ArnoldSmith86:
> 1. I have no issues with the quick download, but I would add "no assets" to default file name.

Done.

> 2. If autosave is enabled, should we do an autosave before the player switches games?  For example, if I'm working on a new game and I have autosave, I might "accidentally" load a different game to copy some code.  It would be nice if autosave kicked in to prevent me from losing the changes in the last 4 minutes.

I made it auto-save when closing the editor. Auto-saving when loading a new state isn't easy because when the client gets notified about the change, the old state is no longer accessible. You can, however, simply press Undo.

> 3. Here's the biggest issue I have.  You can't tell from the Edit mode which game/variant you have loaded.  For example, if I've got a game I'm working on and I think something went wrong, so I go to the Quicksave version.  I figure out what happened and make a few tweaks then I want to make a new variant.  I intended for that to go into the game I was working on, but it goes into the Quicksave.  A new user is going to have issues figure out where his stuff is.  Is there a way to include the information that this will (or has) been saved in the room to the game named: "X"?

I could change the button to `Save as new variant to Chess`. Could get quite long for long names.

> 4. For the save as New Game or New Variant, should we include an overlay that allows you to give the game or variant a meaningful name?  I know these buttons are intended as a shortcut to the main system on the game shelf.  But I think there is risk of confusion if we don't provide some additional information and/or opportunity to customize what is being saved.

Will think of a good solution.

---


> I cannot remenber the last time I downloaded it for backups

Well, I do that all the time.

ArnoldSmith86:
I could make two buttons. "Download game" and "Download without assets" or something.

Note that the quick download button also only downloads the currently loaded variant (no additional variants of the current game). It's basically just meant as an offline quick save.

96LawDawg:
> I could make two buttons. "Download game" and "Download without assets" or something.

That is probably the best compromise between you and Raphael.

Also, is there a limit or should there be a limit on the number of Quicksave variants?  There probably should, but that might conflict with the whole reason for having them, especially if it is saved every 5 minutes and every time I leave the edit mode.  I could quickly end up with 100 variants and have no real way of knowing which one I needed.

96LawDawg:
Latest features work and address all my concerns.  Anything left to do?

Screenshot 2023-11-03 101726

ArnoldSmith86:
A second Download button that loads the entire game. Might be a bit more complicated though because I don't want to actually change the game. I would have to download the existing game, then download the current room state, put the state into the zip file, check the state for new assets, download them, put them into the zip file.

I don't think I want to spend that much time on it atm.

96LawDawg (approved):
Then this is approved through test.  I will update the wiki instructions on how to use the new editor.

96LawDawg:
Here are the instructions in the wiki for the Quick Download button.  Hopefully, @RaphaelAlvez people will pay attention:

Quick Download. This button will download the JSON, and only the JSON and not the assets, of the current variant. This is NOT the method you should use for saving offline versions of games. This is a simple function for limited purposes.

PR #1906, opened 2023-11-02
Improvements to HTML tutorial

96LawDawg:
A couple of additions to the html tutorial for basic widgets.


ArnoldSmith86 (approved):
Like I said in Discord, I don't think you can scroll widgets on mobile. At least not if they are movable.

PR #1912, opened 2023-11-03
fixed hidden `INPUT.switch`

ArnoldSmith86:
#1896 had the side effect of hiding all switches. This fixes it. I also tested other `INPUT` types and fixed the text box highlighting to not extent around the (now missing) label. And I noticed a crash if a `select` did not provide `options`.

PR #1913, opened 2023-11-03
fixed `INPUT.choose` crash with invalid `collection`

ArnoldSmith86:
When testing #1912, I noticed that a simple `INPUT.choose` without anything in a `clickRoutine` causes a crash because it is trying to read an invalid collection.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1913/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1913-pr-test`.

96LawDawg:
I couldn't get this to freeze the room on production where this is the entire clickRoutine:

```
{
      "func": "INPUT",
      "fields": [
        {
          "type": "choose",
          "source": "noCollection",
          "label": "Text goes here"
        }
      ]
    }
  ```
  
  Instead I got a red debug showing the collection does not exist.  What did I do wrong in testing on production?

ArnoldSmith86:
Ah, you're right. It doesn't crash. But only because it is in a promise. If you enable dev tools (F12) and set it to "Pause on uncaught exceptions", it triggers:

image

And looking at that screenshot, it looks like Ambagibus is loading its library image from the web:

image

96LawDawg (approved):
Not going to lie.  I didn't even try that because I didn't even look for that checkmark.  But this does still work (and works better) on this PR.  Approved through test.

Just for documentation purposes, on production, if you aren't using the dev tools, if you get this problem, you get an error in the debug report but the INPUT overlay does not even display.  With this PR, there is no error.  The INPUT overlay opens and shows an empty/null collection (so nothing displays).

PR #1915, opened 2023-11-03
fixed PCIO import of `MOVE` and `FLIP`

ArnoldSmith86:
PCIO recently changed their system how `MOVE` and `FLIP` select the widgets they work on. You can even make the player choose.

This PR tries to make most of that work in the importer. Might not be perfect yet.


96LawDawg:
Tested some of this based on a room I made in PCIO and imported.  So far, the only thing I noticed is that our FLIP does not do everything that the PCIO flip can do.  PCIO can switch sides, flip to random sides, just do the top of the pile, do the entire pile, and keep or reverse the order of the pile.  So there is an import failure, but only because our FLIP is more limited.

ArnoldSmith86:
Yeah, there are some more gaps but this should be a big step for the games that are completely broken now..

96LawDawg (approved):
Except as noted, this seems to work.  Better than what we have now.  Approved through test.

PR #1916, opened 2023-11-04
Update select dropdowns to show a visible arrow for better usability

andypymont:
Current appearance is confusing with no visible indicator that this is a select field:
Screenshot from 2023-11-04 12-23-15

Revised appearance is more clear for the user:
Screenshot from 2023-11-04 12-24-21

### DIFF
--- a/client/css/layout.css
+++ b/client/css/layout.css
@@ -1373,7 +1373,7 @@ button:disabled:hover {
 .inputselect label,
 .inputpalette label,
 .inputchoose label{
-  padding: 4px 5px 0 5px;
+  padding: 4px 5px 2px 5px;
     width: 100%;
     vertical-align: top;
     flex-shrink: 0;
--- a/client/css/layout.css
+++ b/client/css/layout.css
@@ -1397,7 +1397,6 @@ button:disabled:hover {
 .inputstring input,
 .inputselect select
 {
-  appearance: none;
   padding: 20.6px 5px 0px 5px;
   margin: -20.6px 0 0 0;
   font-size: 16px;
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1916/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1916-pr-test`.

96LawDawg:
I made a small change.  When the label was long, the arrow was too close to the label.

Before:
As is

After:
After

This also causes the light blue border to shift down by 2 pixels for numbers, strings, palettes, and choose.  But I think it looks better as the top of the blue border is no longer cut off.

I chose 2px because it required the least amount of tweaking to other code. I'm not tied to this solution, just to the idea that we need to improve separation.

Otherwise, I have viewed all the impacts possibly caused by this PR and think it should be approved.

rogerl50:
> I made a small change. When the label was long, the arrow was too close to the label.
> 
> Before: As is
> 
> After: After
> 
> This also causes the light blue border to shift down by 2 pixels for numbers, strings, palettes, and choose. But I think it looks better as the top of the blue border is no longer cut off.
> 
> I chose 2px because it required the least amount of tweaking to other code. I'm not tied to this solution, just to the idea that we need to improve separation.
> 
> Otherwise, I have viewed all the impacts possibly caused by this PR and think it should be approved.

Perhaps a better solution would be to shorten the available text width.

ArnoldSmith86:
Hmm, it looks like without `appearance: none`, it is ignoring the `line-height: 36px`. So this squishes the layout quite a bit. Let me see what a solution with a custom arrow would look like.

96LawDawg:
This should be abandoned for #1917. The solution there is more elegant and has less side effects.

96LawDawg:
Changes are now in #1917.

Issue #1749, opened 2023-05-10
Filter out cards subject to `RECALL`


96LawDawg:
I have created several games now where I have had to come up with solutions to the problem of keeping certain cards out of play. For example, say you have a deck of 40 cards.  With 4 players you use all 40. With 3 players, you remove 5 specific cards from play. There are work arounds for making this work, but to me it seems like the simplest solution in the JSON would be to have a filter for those cards.  You still have to remove them initially by `SET`ting the parent somewhere else like on off-screen holder. But my solution would mean those cards are not recalled after that and then have the parent set somewhere else again.

It could look something like this:

```JSON
{
  "func": "RECALL",
  "holder": "${PROPERTY parent}",
  "filter": {
    "cardType": [
      "T 1J",
      "T 2J"   
    ],
    "suit": "C"
  }
}
```

Maybe my spacing and my brackets aren't exactly right. But the idea is that you specify any property on a card and an array or scalar value to say what cards will NOT be affected. So in the example above using the nomenclature of our standard deck, the jokers and the clubs would not be recalled.

Could also name the parameter `exclude`, or `limit` or whatever.

robartsd:
In practice I would probably move cards to be excluded from play into an off screen holder and filter based on that parent (or _ancestor). Without changing RECALL, this is several steps: SELECT deck, SELECT not excluded, SET parent (possibly MOVE to another holder since SET parent does not trigger everything that MOVE triggers).

PR #1923, opened 2023-11-06
Add 'excludeCollection' parameter to `RECALL`

96LawDawg:
Implements #1749.

I can't believe how stupidly I phrased the request in that issue.  We already have something that does what I wanted: collections.  So with this PR, you can make a collection of cards you do NOT want recalled and specify that in the parameter 'excludeCollection'.  I have quite a few games where this will make things a lot easier.  I've updated the tutorial and the PL game Flora to use this capability.

Draft wiki update:

Parameters:
* excludeCollection: _collection_ - specifies a collection of cards that will not be subject to recall (default: null).
* holder: _holderID_ - specifies the holder that contains the cards to recall.
* inHolder: _true/false_ - determines whether cards that are not in holders are affected as well. The default is _true_ causing all cards to be recalled. If set to _false_, cards that are in holders will not be recalled.
* owned: _true/false_ - determines whether cards owned by a player (the "owner" property is not null) are affected as well. The default is _true_, causing all cards of the deck to be recalled. If _false_ then cards owned by players are not recalled.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1923/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1923-pr-test`.

96LawDawg:
I tried to make the code more consistent with what it was before, but I went too far and reverted the last change.  Sorry if this is a little sloppy.  But it works.

96LawDawg:
Of course, now that I think about it, would it make more sense to do a traditional collection?  So you can RECALL a holder or a collection and then anything not in the collection wouldn't be recalled?  That seems to make more sense in some ways.  But I was thinking about this from the other side: about ways to keep cards out.

Well, it wasn't entirely wasted effort.  If that's the way we decided to go, then the tutorial and Flora will be easy to update. But I don't know if I want to take a shot at the code again for a while.  I'm so bad and inefficient at this.  Only with the help of ChatGPT can I even make any progress.  Uggh.

96LawDawg:
And after thinking about it some more, supporting regular collections with RECALL doesn't make sense.  Then that just becomes a MOVE with a different name.

ArnoldSmith86:
Reading your comment reminded me that I didn't submit mine:

If you invert the logic and use `collection`, you might as well use `MOVE` now. I think this is a nice addition. Fits to the other parameters of `RECALL`. I think a few of your checks are unnecessary. I'll have a look later. For example you don't need to check if the parameter is set. `getCollection` should return false if given `undefined`. Not 100% sure though. And you probably don't need an extra case for `length==0`.

ArnoldSmith86 (approved):
I rearranged the code so it makes as few changes to the existing `RECALL` code as possible. I'm 99% sure that your code was not changing any existing stuff but now it's way easier to check the diff.

Please check that the new parameter still works as expected.

96LawDawg:
Agree.  Definitely easier to read.  The code looks right to me (but, it's me, so ...).  Most importantly existing `RECALL` functionality and the new exclusion work.

PR #1924, opened 2023-11-06
Public Library: Add hex highlight effect to 2 games

andypymont:
Update public library games _Floodplain Frenzy_ and _Cirque de la Cascade_, changing the `.droptarget` class so that hovering with a droppable token shades the hex rather than showing the square outline of the holder.


96LawDawg (approved):
Approved through test and code review.  Looks good!

Issue #1023, opened 2021-12-14
Add property to detect mousedown on widget


96LawDawg:
It would be useful for some things I'm working on to be able to tell when a widget has been clicked on (mousedown).  A property for that would be great so I could use a propertyChangeRoutine.  Probably would a true/false property and default to false.

96LawDawg:
I know there are always unintended consequences of things, but if this is set up the right way, then it might also help resolve #929 and #930 . For example, if the default was that a widget could not be clicked while the mousedown property was true, that might be a solution to that problem.  I know it probably would have fixed the problem I had in Imagination that led to issue 929.

robartsd:
More likely we would implement a trigger to run a specially named routine, perhaps: `mouseDownRoutine`.

bjalder26:
Currently when a widget is clicked on (mousedown), then "selected" is added to its classList.  When it is released (mouseup) then "selected" is removed from its classList.

So whatever is done, we may just want to replace or alter some of that code.
https://github.com/ArnoldSmith86/virtualtabletop/pull/716/files

96LawDawg:
Even better than true/false would be the playerName of the player clicking the button.  More complicated but more useful.  So values would be null if not in mousedown or a playerName if it is in mousedown.

Issue #930, opened 2021-11-27
Property to prevent simultaneous clicks


robartsd:
Sometimes a button should not be clicked twice within a short time period. It would be convenient to have a property that prevents the second click if two players click a button close to the same time. We hope to fix #929 , but may also implement this feature as well.

Game designers can simulate this behavior with a timer. The button would set its own clickable property to false and start a timer to set the clickable property back to true.

bjalder26:
I think to prevent multiple unintended clicks using a time-based method you'd need to use server-time (when the delta reaches the server), and need a property to indicate the time delay desired.

I think to use a property-based method you'd need another property that increments so that two deltas sent before each client received the first delta could be recognized, and the second delta (to reach the server) discarded.  

I believe using a timer can only help simulate this to a point.  If the clicks occur within the time it takes for a delta to leave one client, be processed by the server, and be received by another client, then I don't think that will prevent the two clicks.

Reposting what I said in Discord:
>I'm not sure how these are currently dealt with.  Maybe there could be a JSON property for buttons where you wanted to avoid them being pressed for either a certain amount of time, or until some other routine changes a property.  Then the first delta could be accepted, and the second rejected.  It seems like it should be button specific, and based on a game-maker setting, so that other buttons could be pressed by numerous people as much as desired.

>I'm thinking if it is a time-based property, that the button has a property (e.g. timeToNextClick: 400) that gives the time to wait in ms.  Then the previous time (server time) is recorded in another property (e.g. lastClickTime).  I think everything would have to be according to server time rather than client time otherwise it wouldn't work.

>I'm thinking that if it is property based, that you could have one property that points at the property to check, and if that is false, then it does not apply the delta, and another property that gets incremented to prevent two deltas with the same 'number' from being applied before the other property is toggled.

robartsd:
> I think to prevent multiple unintended clicks using a time-based method you'd need to use server-time (when the delta reaches the server), and need a property to indicate the time delay desired.

I'm pretty sure we don't want to introduce this at the server. The general attitude has been that the server does nothing except sync state. Since we didn't introduce a server time mechanism with the timer widget, I highly doubt we'd choose to introduce it for this.

> I think to use a property-based method you'd need another property that increments so that two deltas sent before each client received the first delta could be recognized, and the second delta (to reach the server) discarded.
> 
[...]

PR #716, opened 2021-08-11
Stop enlarged from flickering

bjalder26:
- Stops enlarged view from flickering if mouse moves faster than the widget and the mouse moves off of the widget while moving.
- Stops enlarged view from ending if mouse temporarily moves off the screen while a widget is being moved.
- Makes the enlarged view end when mouseleave happens rather than when mouseup happens.  

This addresses issues #31 and #715.

PR #1925, opened 2023-11-07
Add event listeners to implement `selectedBy` property

### LINKED ISSUE #929: Simultaneous button clicks possible because of mousedown and mouseup timing
Say there is a button in a room.  Player 1 clicks it and holds the mousedown.  Player 2 clicks it and completes mousedown and mouseup.  Then Player 1 releases mousedown.  The button executes twice.  This can potentially cause a game to malfunction or other problems.  @ArnoldSmith86 and I discovered this by examining a trace in a room where this happened.

I have simplified the problem and created a demo room.  https://virtualtabletop.io/2hyr.  Click on the button and hold it down on your desktop.  Click and release on a mobile device.  The button will move x + 100.  Then release your desktop and the button will move x + 100 again.  This does not happen unless you hold the mouse still.  Moving the mouse while mousedown does not cause the problem.

I don't know how this could possibly be fixed, but I would categorize this as a very serious bug.

96LawDawg:
Implements #1023.  Would essentially resolve #929 and #930 as well (not directly, but through implementation of a changeRoutine -- will post example in test room).

We have lots of temporary custom properties that get added to widgets:  Ex: dragging, hoverTarget, hoverParent, etc. But until this PR, there has not been a way to detect if a widget was selected by something like a mousedown event.  I'm not talking about clicking, just selecting.  Because clicking (like a clickRoutine) doesn't actually happen until the mouse is released.  And dragging does not show up until the widget moves (if that is even possible, which it usually is not for regular buttons).

This PR adds an event listener for mousedown and mouseup and adds the "selectedBy" property with the name of the player.  If we do, this should we also add a `.selected` class?  Apparently it was there at one time (https://github.com/ArnoldSmith86/virtualtabletop/pull/716/files).  Not sure what happened to that.

I don't yet know how this will work with touch devices.  Something I will test.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -30,6 +30,7 @@ export class Widget extends StateManaged {
     this.targetTransform = '';
     this.childArray = [];
     this.propertiesUsedInProperty = {};
+    this.addOverlay = document.getElementById('addOverlay');
 
     if(StateManaged.inheritFromMapping[id] === undefined)
       StateManaged.inheritFromMapping[id] = [];
@@ -101,8 +102,7 @@ export class Widget extends StateManaged {
       if (!this.timer) {
         this.timer = setTimeout(this.onlongtouch.bind(this), 500, false);
       }
-      const addOverlay = document.getElementById('addOverlay');
-      if (!addOverlay || window.getComputedStyle(addOverlay).display === 'none')
+      if (!this.addOverlay || window.getComputedStyle(addOverlay).display === 'none')
         this.set('selectedBy', playerName);
     }
 
@@ -110,8 +110,7 @@ export class Widget extends StateManaged {
       clearTimeout(this.timer);
       this.timer = null;
       this.hideEnlarged();
-      const addOverlay = document.getElementById('addOverlay');
-      if (!addOverlay || window.getComputedStyle(addOverlay).display === 'none')
+      if (!this.addOverlay || window.getComputedStyle(addOverlay).display === 'none')
         this.set('selectedBy', null);
     }
 
@@ -2312,14 +2311,12 @@ export class Widget extends StateManaged {
   }
 
   async selected() {
-    const addOverlay = document.getElementById('addOverlay');
-    if (!addOverlay || window.getComputedStyle(addOverlay).display === 'none')
+    if (!this.addOverlay || window.getComputedStyle(addOverlay).display === 'none')
       await this.set('selectedBy', playerName);
   }
 
   async selectedEnd() {
-    const addOverlay = document.getElementById('addOverlay');
-    if (!addOverlay || window.getComputedStyle(addOverlay).display === 'none')
+    if (!this.addOverlay || window.getComputedStyle(addOverlay).display === 'none')
       await this.set('selectedBy', null);
   }
 
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -101,7 +101,8 @@ export class Widget extends StateManaged {
       if (!this.timer) {
         this.timer = setTimeout(this.onlongtouch.bind(this), 500, false);
       }
-      if (!document.body.classList.contains('edit'))
+      const addOverlay = document.getElementById('addOverlay');
+      if (!addOverlay || window.getComputedStyle(addOverlay).display === 'none')
         this.set('selectedBy', playerName);
     }
 
@@ -109,7 +110,8 @@ export class Widget extends StateManaged {
       clearTimeout(this.timer);
       this.timer = null;
       this.hideEnlarged();
-      if (!document.body.classList.contains('edit'))
+      const addOverlay = document.getElementById('addOverlay');
+      if (!addOverlay || window.getComputedStyle(addOverlay).display === 'none')
         this.set('selectedBy', null);
     }
 
@@ -2310,12 +2312,14 @@ export class Widget extends StateManaged {
   }
 
   async selected() {
-    if (!document.body.classList.contains('edit'))
+    const addOverlay = document.getElementById('addOverlay');
+    if (!addOverlay || window.getComputedStyle(addOverlay).display === 'none')
       await this.set('selectedBy', playerName);
   }
 
   async selectedEnd() {
-    if (!document.body.classList.contains('edit'))
+    const addOverlay = document.getElementById('addOverlay');
+    if (!addOverlay || window.getComputedStyle(addOverlay).display === 'none')
       await this.set('selectedBy', null);
   }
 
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1925/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1925-pr-test`.

96LawDawg:
This needs to be made to work with touch events as well.  I don't know how.  And testing is a problem for me.  The easiest way to test is to put it on this server.  I'm not going through all the steps to allow my iPad to connect to my PC on the localhost server when I'm coding changes.  Probably there is some other way involving GitPod or whatever.  That's beyond me right now.

Any thoughts about this PR in general and/or how to get touch events to do the same thing?

96LawDawg:
Thanks to at @mousewax, I figured out that I could replace "localhost" with my PC's ip address and do testing.  Because of that, I was able to get touch events to work and hopefully finished cleaning up the code.

96LawDawg:
There is now a demo in the pr test room of how to guarantee a button can only be pressed once.

client/js/widgets/widget.js
```diff
@@ -2302,6 +2312,18 @@ export class Widget extends StateManaged {
       await this.set('rotation', degrees);
   }
 
+  async selected() {
```
ArnoldSmith86, line 2315: ~~These do not need to be `async`. Basically only if they contain an `await`.~~

Scratch that. See below.

client/js/widgets/widget.js
```diff
@@ -2302,6 +2312,18 @@ export class Widget extends StateManaged {
...
+    if (!this.addOverlay || window.getComputedStyle(addOverlay).display === 'none')
+      this.set('selectedBy', playerName);
+      this.domElement.classList.add('selected');
```
ArnoldSmith86, line 2318: This doesn't work the way you want it to. `this.domElement.classList` will be overridden destructively whenever something about the widget changes.

See `setHighlighted` which basically does the same thing for selecting widgets in the editor. It sets a flag on the object itself (`this`) and then that flag is used whenever `this.domElement.classList` is updated.

client/js/widgets/widget.js
```diff
@@ -2302,6 +2312,18 @@ export class Widget extends StateManaged {
       await this.set('rotation', degrees);
   }
 
+  async selected() {
+    if (!this.addOverlay || window.getComputedStyle(addOverlay).display === 'none')
+      this.set('selectedBy', playerName);
```
ArnoldSmith86, line 2317: `set` always needs an `await`.

client/js/widgets/widget.js
```diff
@@ -2302,6 +2312,18 @@ export class Widget extends StateManaged {
       await this.set('rotation', degrees);
   }
 
+  async selected() {
+    if (!this.addOverlay || window.getComputedStyle(addOverlay).display === 'none')
```
ArnoldSmith86, line 2316: `addOverlay` should be undefined. `this.addOverlay` might work.

`window.getComputedStyle` should be unnecessary. Checking `this.addOverlay.style.display` should be enough.

client/js/widgets/widget.js
```diff
@@ -30,6 +30,7 @@ export class Widget extends StateManaged {
     this.targetTransform = '';
     this.childArray = [];
     this.propertiesUsedInProperty = {};
+    this.addOverlay = document.getElementById('addOverlay');
```
ArnoldSmith86, line 33: Saving it in `this.addOverlay` seems overkill to me. Technically it's faster though, so it's fine.

In general, I try to use `$('#addOverlay')` in the code instead because it is shorter. Again, technically using `document.getElementById` is faster, so it's fine. Of course, ChatGPT for example doesn't know about those helper functions.

client/js/widgets/widget.js
```diff
@@ -93,17 +95,25 @@ export class Widget extends StateManaged {
...
     this.domElement.addEventListener("touchend", e => this.touchend(), false);
+    this.domElement.addEventListener('mousedown', e => this.selected(), false);
+    this.domElement.addEventListener('mouseup', e => this.selectedEnd(), false);
 
     this.touchstart = function() {
       if (!this.timer) {
         this.timer = setTimeout(this.onlongtouch.bind(this), 500, false);
       }
+      if (!this.addOverlay || window.getComputedStyle(addOverlay).display === 'none')
+        this.set('selectedBy', playerName);
```
ArnoldSmith86, line 106: Why don't you just call `this.selected()` here (and `this.selectedEnd()` below)?

client/js/widgets/widget.js
```diff
@@ -93,17 +95,25 @@ export class Widget extends StateManaged {
...
+        this.domElement.classList.add('selected');
```
ArnoldSmith86, line 107: In all four occurrences of this, this line is indented but not part of the `if` conditional code. Either fix the indentation or - more likely - add `{ ... }`.

ArnoldSmith86 (changes requested):
I added a few code comments but my main problem with this is that I think it's still a race condition. I have to think about it a little more though.

client/js/widgets/widget.js
```diff
@@ -93,17 +95,25 @@ export class Widget extends StateManaged {
     this.domElement.addEventListener('mouseleave',  e => this.hideEnlarged(e), false);
     this.domElement.addEventListener("touchstart", e => this.touchstart(), false);
     this.domElement.addEventListener("touchend", e => this.touchend(), false);
+    this.domElement.addEventListener('mousedown', e => this.selected(), false);
+    this.domElement.addEventListener('mouseup', e => this.selectedEnd(), false);
```
ArnoldSmith86, line 99: Oh, and in general I don't like these event handlers. The others were added at some point to improve `enlarge` but I want to redo that at some point. All the mouse/touch stuff of widgets is usually handled by `mousehandling.js`. That's definitely not a deal breaker for this PR if it works though.

96LawDawg:
> I added a few code comments but my main problem with this is that I think it's still a race condition. I have to think about it a little more though.

I'm pretty sure you are right.  It was just harder to test than the previous condition where click starts upon button release.  I used the network throttling dev tools and demonstrated that it is a race conditon.

So this is no better than what we currently have and we should abandon it.

BUT, I did have another thought. For the Great Mouse Handling Code Rewrite of 2024, we could still use something like this to help prevent players from stealing cards.  Right now, another play can grab onto and take away from you a card that you have clicked on and are still holding down the mouse (and probably dragging somewhere).  I think that shouldn't be the default behavior.  Something like what I tried to do here could be used to prevent that.

96LawDawg:
This is not going to work to do what it was intended to do.  A different approach is needed.

PR #1926, opened 2023-11-07
Initial comparison of Roboto Flex and Roboto

mousewax:
This change switched from the 18 Robot and Roboto Condensed font files to a single Robot Flex font.

There are, of course, some changes to spacing and width, but I have attempted to mitigate those.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1926/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1926-pr-test`.

96LawDawg:
Discord

This was just today.  And now you want to change the default font??  What about poor Paladin??  Poor guy.

96LawDawg:
So, I do have some concerns about this.  Some games I've made in the past push the boundaries of where the text goes right up to the limit.  For example, in Flight, I actually set a variable text size on each card for the explanatory text at the bottom so that it will be as large as possible.  This PR changes the layout.  Not on every card, but enough.  For example, on production:

Flight Original

In this PR:

Flight New

I looked at a couple other of my games and didn't see issues on them, but that isn't to say they aren't there.  And one of Arnold's (Chronology) on production:

Chronology Original

In this PR:

Chronology New

In Chronology, he built it so that there would be proper hyphenation so he didn't have to manually adjust each card, but at least 2 of the first dozen cards I looked at had different line breaks.

I'm not sure we should do this without a (complicated) file updater.

mousewax:
These are valid concerns and they are the reason I put this pull request together. 

I come up against this all the time in the print world where suddenly I have weird problems like that. Let me continue to adjust things to try and get them closer to what exists.

ArnoldSmith86:
Are you the first one to do this? Shouldn't there be something to copy and paste somewhere?

PR #1934, opened 2023-11-11
Add Public Library game Sunshine and Rainbows

andypymont:
Add the Public Library game Sunshine and Rainbows (similar to Oh Meow Bow / Five Three Five, a climbing/shedding card game).


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1934/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1934-pr-test`.

96LawDawg:
Comments:

Need similar to link: https://boardgamegeek.com/boardgame/401309/oh-meow-bow.

Typo in how to use implementation, 2nd para: "To instead Add" should be "To instead add"

Typo at the bottom of the player aid: "inbetween".  Also the player aid moves; probably not intended?

Each of the holders has a 2px border width.  But when you bring a card over it, the width changes to 1px by default.  That gives a very strange effect that I don't think you want.  I would change the border width to 1px (easiest), but you could use css magic to adjust what happens when it becomes the drop target.

ArnoldSmith86:
> Each of the holders has a 2px border width. But when you bring a card over it, the width changes to 1px by default. That gives a very strange effect that I don't think you want. I would change the border width to 1px (easiest), but you could use css magic to adjust what happens when it becomes the drop target.

Is that still an issue after #1921?

andypymont:
> Need similar to link: https://boardgamegeek.com/boardgame/401309/oh-meow-bow.

Added.
> Typo in how to use implementation, 2nd para: "To instead Add" should be "To instead add"

I capitalised both Overwrite and Add as they are the names of actions detailed on the player aid.

> Typo at the bottom of the player aid: "inbetween". Also the player aid moves; probably not intended?

Fixed.

> Each of the holders has a 2px border width. But when you bring a card over it, the width changes to 1px by default.

I don't see this behaviour in the PR test room now, so hopefully it was fixed by #1921 as suggested above.

96LawDawg:
> > Each of the holders has a 2px border width. But when you bring a card over it, the width changes to 1px by default. 

> Is that still an issue after #1921?

No. I didn't think about whether that fix had been applied in this PR.

96LawDawg (approved):

PR #1936, opened 2023-11-12
added JSON commands for picking a symbol

ArnoldSmith86:
There was a bit of discussion in Discord about this lately.

When working on the new editor, I included Noto Emoji for safe cross-platform display of emoji and all of the game-icons.net icons to the internal VTT assets. I also made a nice icon picker for them but so far it was only accessible from the game metadata.

This PR adds two JSON editor commands. One for image-based icons anywhere where the Upload Asset command was available. And one for buttons with class `symbols` or `material-icons` for the `text` property.

My plan is still to have new properties and card object types that make this even easier but this is a nice first easy step that makes these resources more freely available.

This is also still missing proper attribution. For now you will have to insert the proper attribution for game-icons.net icons into the game metadata manually.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1936/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1936-pr-test`.

96LawDawg:
Kinda glitchy as the images are loading.  Will it take a long time every time?


https://github.com/ArnoldSmith86/virtualtabletop/assets/76912527/568f6972-e510-4856-9617-7a669935790d

96LawDawg:
Quite a few VTT symbols are missing.  For example, all of the arrows, the seats/chairs.  But many other things also.

ArnoldSmith86:
It's probably missing the ones that are just copied from Material Icons.

Do the glitches happen in production as well? What about the other PR? Which browser?

96LawDawg:
> Do the glitches happen in production as well?

No.  The icon picker for meta data has always worked fine.

>What about the other PR? Which browser?

Haven't tested the other PR yet.  Browsers with problem are Crome and Edge in Windows 11.  No problem in Firefox.

96LawDawg:
Strange.  I posted an answer to your other questions.  They show up in Discord but not here.  Will post again, just in case.

> Do the glitches happen in production as well?

No. The icon picker for meta data has always worked fine.

>What about the other PR?

Same problem.

>Which browser?

Browsers with problem are Crome and Edge in Windows 11. No problem in Firefox.

ArnoldSmith86:
> The icon picker for meta data has always worked fine.

Is it still working in this PR? The only difference should be the size of the icons. I could show the small previews and only show bigger ones when filtered to <100.

96LawDawg:
> Is it still working in this PR?

Yes, except for the flags.

ArnoldSmith86:
> > Is it still working in this PR?
> 
> Yes, except for the flags.
> 

Those are also broken in production though, right? Windows doesn't have the flags iirc? I can probably make an exception for those and always display the images instead.

96LawDawg:
> > Yes, except for the flags.
> 
> Those are also broken in production though, right? 

Yes.  They show up when you mouse over them, but otherwise just have the letters.

ArnoldSmith86:
How about now?

96LawDawg:
Works in Chromium and Firefox on Windows 11.  Does not work on iOS on newish iPad.

96LawDawg (approved):
Approved through test

Issue #1900, opened 2023-11-01
Increase availability of "${PROPERTY property OF id}" syntax


96LawDawg:
Just capturing the discussion from Discord the other day: https://discord.com/channels/770758631146782780/793626590848352306/1168191332221730927

Make it available in inheritFrom, card and dice face objects, cardTypes, faceCSS, and svgReplaces.

96LawDawg:
This would supersede #1821.

robartsd:
The trick here is generalizing the processing of deltas to run callbacks for everything that is reading the property that got changed. In routines, there is no need for a callback because the property changing after the routine runs does not matter - the routine isn't running anymore.

PR #1937, opened 2023-11-12
added `icon` support to buttons, basic widgets, dice and cards

ArnoldSmith86:
This PR is meant to make using the awesome included icon collections easy. For now in the JSON editor, in the next step with proper UI.

**Without this PR**, you had to know this voodoo for material icons (and it doesn't even scale with the size):

```JSON
  "classes": "material-icons",
  "css": {
    "font-size": "68px"
  },
  "text": "restart_alt"
```

And this voodoo for game-icons:

```JSON
  "color": "white",
  "css": {
    "background-size": "70%"
  },
  "image": "/i/game-icons.net/delapouite/control-tower.svg",
  "svgReplaces": {
    "#000": "color"
  }
```

And even different voodoo for emojis or the VTT symbols font.

Now it becomes this:

image

Want text and icon?

image

Automatically uses the same color property as the text:

image

And even the same hover color:

image

If you want something more fancy, you can switch to the advanced options:

image

Or combine multiple icons and customize them:

image

Or go crazy and even customize the separate shapes of game-icons:

image

---

Additionally `svgReplaces` now supports this syntax for replacing different occurrences with different values:

```JSON
{
  "svgReplaces": {
    "#000": [ "red", "white" ]
  },
  "red": "red",
  "white": "white"
}
```

---

- [x] `[card_H]` (and others) show `[card]` instead. Might be unrelated.
- [x] Support `textColorOH`.
- [x] Add to dice.
- [x] Add outline support.
- [x] Fix missing things for deck generation.
- [x] Align and auto-size text next to icons.
- [x] Fix color on basic widgets.
- [x] Enable coloring different areas in game-icons.net svgs?
- [x] Check why images are too big on my phone.
- [x] Allow asset URLs as icons.
- [x] Make hoverColor work on all widget types?
- [x] Make `offsetX` and `offsetY` relative?
- [x] Resolve margin vs scale.
- [x] Changing `svgReplaces` no longer updates the widget.
- [x] Allow `${PROPERTY x}` syntax in `icon`.
- [x] Having `color: ["red","blue"], hoverColor: ""` makes everything black on hover instead of not changing anything.
- [x] JSON commands don't seem to work inside dice faces.
- [x] Different stroke properties for different areas?
- [x] Multiple icons (and text) in a card face object? (for reusing icon definitions from other widgets)
- [x] Rotation in cards is applied twice now.
- [x] JSON editor commands for cards?
- [x] `strokeWidth` is relative on images (to 512px) and absolute on fonts.
- [x] Test integer and bool JSON editor commands on arrays.
- [x] Can't have normal strokeWidth but no hoverStrokeWidth because 0 gets overridden.
- [x] ~~Gradients?~~ It's not _that_ difficult but we can add it later.
- [x] ~~Shadows?~~ It's not _that_ difficult but we can add it later.
- [x] ~~Add `textColor`.~~ Doesn't make sense as a property of an icon.
- [ ] Give text property advanced options like icon? Nice to have, but not essential
- [ ] Make multi-line text work better when combined with icon
- [ ] Different emoji sizes in the symbol picker between Firefox and Chrome
- [ ] A monochrome and a color emoji in the same button don't align nicely
- [ ] Test and fix advanced options with monochrome emojis

This line will be added to the wiki for basic widget and button, and a similar entry with appropriate tweaks will be made for cards and decks.

```


| icon     |   null    |  Adds an icon or emoji from a curated list (or custom image uploaded by the user) to the widget. The value can be just text that is automatically inserted by the JSON Editor. Or using advanced settings, the icon can be an object with the following key value pairs (with the default value following in parenthesis): name (the name of the icon), scale (1), offsetX (0), offsetY (0), rotation (0), color (""), strokeColor (""), strokeWidth (0), hoverColor (""), hoverStrokeColor (""), hoverStrokeWidth (null)  | [Icons & Emojis](https://virtualtabletop.io/tutorial/icons) |

```

### DIFF
--- a/client/css/editor/sidebarModules.css
+++ b/client/css/editor/sidebarModules.css
@@ -235,22 +235,8 @@
   margin-top: 10px;
 }
 
-.pest_control.editorModule .validation-controls button {
-  margin-right: 10px;
-  padding: 5px 10px;
-  background-color: var(--VTTblue);
-  color: white;
-  border: none;
-  border-radius: 3px;
-  cursor: pointer;
-}
-
-.pest_control.editorModule .validation-controls button:hover {
-  background-color: var(--VTTblueDark);
-}
-
 .pest_control.editorModule .validation-time {
-  color: var(--textDimColor1);
+  color: var(--textColor);
   font-size: 12px;
   font-style: italic;
 }
--- a/client/js/editor/sidebar/json.js
+++ b/client/js/editor/sidebar/json.js
@@ -162,7 +162,7 @@ class DebugModule extends SidebarModule {
 
     div(target, 'staticErrors', `
       <div class="validation-controls" style="margin-top: 10px; display: none;">
-        <button id="runValidationButton" class="ui-button">Run Validation</button>
+        <button id="runValidationButton" icon=data_check>Run Validation</button>
         <span class="validation-time"></span>
       </div>
       <div class="success">No validation problems found!</div>
@@ -180,6 +180,7 @@ class DebugModule extends SidebarModule {
     `);
 
     on('#runValidationButton', 'click', e=>this.button_runValidation());
+    this.lastValidationTime = 0;
     this.updateValidation();
   }
 
@@ -189,7 +190,7 @@ class DebugModule extends SidebarModule {
       return;
     }
 
-    const state = Object.fromEntries(widgets.entries().map(([id, w])=>[id, w.unalteredState]));
+    const state = Object.fromEntries([...widgets].map(([id, w])=>[id, w.unalteredState]));
     
     // Hide manual validation controls
     const controls = $('.validation-controls', this.moduleDOM);


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1937/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1937-pr-test`.

ArnoldSmith86:
**Edit: test server is working now.**

GitHub seems to have problems at the moment which seems to cause the PR test server to use an old commit. The PR test room is supposed to look like this:

image

The four types of icons currently in the VTT repo are shown: Material Icons, internal VTT symbols font, Noto Emoji and game-icons.net.

96LawDawg:
This is great.  I'll have to explore the new default for textColor some.  And still glitchy in displaying the emoji like in the other PR.  But otherwise, this is awesome.

96LawDawg:
For the textColor thing, there is a possibility of breaking things.  The Button Appearance tutorial, Introduction variant has this on production:

Production

And this in this PR:

This PR

But if you need it for this PR, I say we go with our new more relaxed approach to breaking things: as long as it doesn't break too many things too badly, do it!

ArnoldSmith86:
It's not necessary to break that. I'll look into it.

ArnoldSmith86:
How about now?

96LawDawg:
Works in Chromium and Firefox on Windows 11.  Does not work on iOS on newish iPad.

ArnoldSmith86:
I copied my local test room to https://test.virtualtabletop.io/PR-1937/ick2 but it's a bit messy.. :grin:

96LawDawg:
> One more important aspect of this is that `svgReplaces` is a bit too inflexible for setting different parts of a game-icons.net icon to different colors.
> 

Would changing the syntax here for `svgReplaces` conflict with or make it harder to implement #1900?  Just something to keep in mind.  I think that the additional flexibility you suggest is great, but just want to make sure we are thinking that through.

ArnoldSmith86:
> Would changing the syntax here for `svgReplaces` conflict with or make it harder to implement #1900?

It shouldn't. But since I suggested that, I looked at an example SVG and they don't actually have `#000` in there multiple times. It's just one `<path>` with `M`s in there to split it. I still want to try to make it happen for this new icon stuff though.

96LawDawg:
True for the icons from the game-icons.net site.  But obviously there are other SVGs that exist that do have multiple colors and being able to replace multiple svg colors would be a nice addition somewhere, somewhen.

ArnoldSmith86:
I'm done with my todo list and marked this "ready for review". I'm sure we'll find some stuff to fix.

I copied my local tests from development to these test server rooms:

https://test.virtualtabletop.io/PR-1937/tests1
https://test.virtualtabletop.io/PR-1937/tests2
https://test.virtualtabletop.io/PR-1937/tests3
https://test.virtualtabletop.io/PR-1937/tests4
https://test.virtualtabletop.io/PR-1937/tests5
https://test.virtualtabletop.io/PR-1937/tests6

Maybe something can be used for a tutorial or whatever.

96LawDawg:
How do you change the font size when the text covers more than one line?  css font-size did not work

```JSON
 "css": {
    "font-size": "40px"
  },
  "icon": {
    "name": "delapouite/brute",
    "scale": 0.8
  },
  "text": "Brute\nIcon",
  ```

**Comment by ArnoldSmith86** [comment]
It's not meant to be changed. Does the layout break?

**Comment by 96LawDawg** [comment]
Depends what you mean by "break."  It becomes unreadable.
Screenshot 2024-09-13 125201

**Comment by ArnoldSmith86** [comment]
Maybe we can tweak the layout a bit so it works better out of the box.

You can of course reach it with `css`. You can look at the HTML structure and class names in imagewidget.css or the dev tools.

**Comment by 96LawDawg** [comment]
> Maybe we can tweak the layout a bit so it works better out of the box.

I think it makes sense to at least try.  If I have 2 icons, I can adjust one to be bigger or smaller by using scale, or move it using the advanced options like offsetX and Y.

> You can of course reach it with `css`. 

Not easily.  It seems that the css property is completely overridden.  For example, in addition to not changing the font-size, changing color that way also has no effect.  Maybe we just need to add the same advanced options for the text property?  Instead of scale, support font-size.  Allow for offset on the text, allow for color change on the text?

**Comment by ArnoldSmith86** [comment]
Color uses the widget text color:

image

**Comment by ArnoldSmith86** [comment]
It's not like I don't agree. But I think it's good that icon+text uses the `text` property of the widget:

```JSON
{
  "type": "button",
  "id": "button",
  "text": "Text",
  "icon": "sync"
}
```

Feels right that you then have text or icon or both or none.

But like that the text is outside of `icon` and it would be weird to have properties about it inside `icon` (would have to be restructured).

But: it's also weird that now the text auto-scales when you have text+icon but it doesn't auto-scale when you just have text.

---

Oh, I reread your comment and you're suggesting something like this, right:

```JSON
{
  "type": "button",
  "id": "button",
  "text": {
    "value": "Text",
    "scale": 2
  },
  "icon": "sync"
}
```

---

For your example with the tiny text, it's obvious though that the algorithm for the alignment can be improved to make it bigger automatically. I think it just needs to give the text more height instead of 50%.

96LawDawg:
> Oh, I reread your comment and you're suggesting something like this, right:

Yes.  For consistency, it would be good if we didn't have things like textColor as a stand-alone property (of course, we have to leave it for compatibility), but make the text property have advanced options like icon.

```JSON
{
  "type": "button",
  "id": "button",
  "text": {
    "value": "Text",
    "scale": 2,
    "color":"blue",
    "hoverColor": "red",
    "offsetX": -10  
  },
  "icon": "sync"
}
```

ArnoldSmith86:
It's not like I don't like your suggestion about customizing `text` but I gotta say that thinking about it demotivates me a lot. :grinning:

I made it now reserve 2/3rds of the height if the text includes a newline. If it doesn't, it's the same as before. No idea if that helps...

96LawDawg:
> It's not like I don't like your suggestion about customizing `text` but I gotta say that thinking about it demotivates me a lot. 

Then don't do it.  This is probably good enough.  I was holding off on reviewing until you figured out what you wanted to do.  I'll try to review it as is this week and also work up a tutorial.

> I made it now reserve 2/3rds of the height if the text includes a newline. If it doesn't, it's the same as before. No idea if that helps...

It does help.  It isn't perfect.  And it doesn't work the same way as text without icons.  But this is probably good enough.

96LawDawg (approved):
Approved through test.  I added a tutorial for Icons & Emojis.  I updated the Dice tutorial.  I updated the Button Appearance to tutorial (and I moved something from there into the css tutorial that should have been moved long ago).

I will work on a wiki update that adds the icon property to the widget page.

96LawDawg:
The tutorial probably needs another look depending on when #2432 gets merged.

96LawDawg (changes requested):
Some icons are broken.  Maybe this changed since when we switched everything from Google Material Icons to Google Material Symbols?

Screenshot 2025-07-09 174928

PR #1942, opened 2023-11-14
Add skipTurn for seats and TURN

### LINKED ISSUE #1170: Seats and `TURN` need to support skipping a player
In a game I was building, I came across a need to add a feature to seats.  In the game, players can leave the round early (but still sit in their seats).  Using `TURN`, there is no easy way to keep skipping a player.  It would be nice to have a property `skipTurn` that is true or false and defaults to false.  If true, then that player will be excluded from being selected during a `TURN`.

There is a workaround (and is apparently what was intended when seats were designed).  It works like this (just documenting how to do it).  When the condition exists for a player to be skipped, set the property `hideTurn` to true.  All that does is make sure the seat does not glow.  It does not actually skip the turn.  To skip the turn, you have to change the `source` property of seats using something like:

```JSON
{
      "func": "SELECT",
      "type": "seat",
      "property": "hideTurn",
      "relation": "!=",
      "value": true,
      "collection": "seatTurn"
    },
    {
      "func": "TURN",
      "source": "seatTurn"
    }
```

96LawDawg:
Fixes #1170.

Adds new property "skipTurn" to seats which defaults to false.

Adds new condition to the engine for `TURN` functions so that a seat with the property "skipTurn" == true will not be included in the turn operation.

I think this needs a file updater.  If a designer has set a property skipTurn on a seat and it is boolean, then `TURN` may work unexpectedly.  There are no instances of the property in the public library, but who knows what might be out there.  I thought about how to do this and realized I don't know if I know how to change a routine and a property in the same updater.  I'll see if I can figure it out and push a commit.  But maybe not.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1942/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1942-pr-test`.

client/js/widgets/widget.js
```diff
@@ -1831,7 +1831,7 @@ export class Widget extends StateManaged {
           a.turnCycle = 'forward'
         }
         let c = a.source === 'all' ? Array.from(widgets.values()) : collections[getCollection(a.source)] || [];
-        c = c.filter(w => w.get('type') === 'seat');
+        c = c.filter(w => w.get('type') === 'seat' && w.get('skipTurn') !== true);
```
ArnoldSmith86, line 1834: I would just do `!w.get('skipTurn')`. If someone sets this with a routine which happens to have `0` or `1`, I don't see a reason why it shouldn't work.

96LawDawg: >  would just do !w.get('skipTurn'). If someone sets this with a routine which happens to have 0 or 1, I don't see a reason why it shouldn't work.

Good idea.  Done.

96LawDawg:
The file update works except that it does not catch something like:

```
 {
      "func": "SET",
      "collection": [
        "seat2"
      ],
      "property": "skipTurn",
      "value": 1
    },
```

Also, all TestCafe fails now.  So it must have done something more drastic than I expected.  Help?

ArnoldSmith86:
Made a few changes to the file updater:

- Only checking `properties.skipTurn !== undefined` because `null` should just be migrated like anything else.
- Deleting `properties.skipTurn` instead of setting it to `null`. We're "moving" the value.
- `replace` now works with the JSON representation of the routine operation so it can replace in any type of operation.
- I added `\b` to the regular expression so it only matches "whole words".

I didn't test any of this. We should construct a test game that actually triggers the file updater. Can be easily done in production which uses `VERSION = 14`. It also needs to test in nested routines (like `thenRoutine`) to make sure it doesn't replace the string twice (resulting in `skipTurnFileUpdaterFileUpdater`).

96LawDawg:
Thanks for the file updater assist.  I again tested the properties and routines.  Works this time at all levels.  Here's the output from the routine:

```
"clickRoutine": [
    "var increment = ${PROPERTY skipTurnFileUpdater OF seat1}",
    {
      "func": "SET",
      "collection": [
        "seat2"
      ],
      "property": "skipTurnFileUpdater",
      "value": 1
    },
    {
      "func": "SET",
      "collection": [
        "seat2"
      ],
      "property": "SkipTurn",
      "value": "hello"
    },
    {
      "func": "SET",
      "collection": [
        "seat2"
      ],
      "property": "skipTurn1",
      "value": 0
    },
    {
      "func": "FOREACH",
      "loopRoutine": [
        "var skipTurnFileUpdater = ${skipTurnFileUpdater}",
        {
          "func": "SET",
          "collection": "skipTurnFileUpdater"
        },
        {
          "func": "IF",
          "thenRoutine": [
            "var increment = ${skipTurnFileUpdater}",
            {
              "func": "GET",
              "property": "skipTurnFileUpdater"
            }
          ]
        }
      ]
    }
  ]
  ```

ArnoldSmith86 (approved):
I guess I'll approve it then?

PR #1943, opened 2023-11-14
Support seat id in `TURN` with mode "seat"

### LINKED ISSUE #1905: `TURN` position parameter should support seat id
Using the "position" parameter of the `TURN` function, you can set the position of the seat with the next turn to first or last.  But it would be an improvement if you could also set it to be a specific seat `id`.  Right now, it takes several SELECTs and SETs to accomplish that task.

96LawDawg:
Fixes #1905.

The "turn" parameter in `TURN` can now be "first", "last", a number, or a seat ID.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1943/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1943-pr-test`.

96LawDawg:
This was working I thought.  But as I was testing different scenarios, I got it to fail.  So I'll need to take a look at it again later.

client/js/widgets/widget.js
```diff
@@ -1848,7 +1850,10 @@ export class Widget extends StateManaged {
           let nextTurnIndex = 0;
 
           if(a.turnCycle == 'position') {
-            if(a.turn == 'first') {
+            if (turnIDIndex !== -1) {
+              nextTurnIndex = c.findIndex(w => w.get('id') === a.turn);
```
ArnoldSmith86, line 1854: Isn't `turnIDIndex` already the value you want to set `nextTurnIndex` to?

ArnoldSmith86:
> I got it to fail

How did it fail?

96LawDawg:
> How did it fail?

Hard to say.  Sometimes it works.  Sometimes it doesn't.  Strange.  I'll have to mess around with it some more I guess.

96LawDawg:
I found and fixed the problem with my seat id / turn thing.  I was confusing 2 different arrays: an array of all the seats and an array of occupied seats.

client/js/widgets/widget.js
96LawDawg: Fixed.

ArnoldSmith86 (approved):
Code looks good to me. Feature looks useful to me. I didn't test anything though. :grin:

96LawDawg:
> Edit: This can, of course, break things if you called a seat `first` and used `turn: first` in `TURN`. Do we care?

No.  It won't break existing things in the PL based on your search and it is really unlikely that people named their seats "first", "second", etc.  We could prevent any possible conflict by making a new parameter "seat" or whatever and have "position" check for it and if it is present, then disregard anything in "turn."  That's probably the 100% solution.  But I think what we have is 99%.  But we can leave this open for a few days to see if anybody else has a concern.

ArnoldSmith86:
Changing it to `turnCycle: seat` or something similar does sound like a good idea though. It doesn't really fit the other `position` things.

96LawDawg:
You are right.  As usual.  I'll work on that when I get a chance.  Make it "turnCycle": "seat" and "turn" is the seat id?

96LawDawg:
Fixed this so it uses a new turnCycle "seat."  Easier to follow and understand in the engine and it prevents any possible conflicts with seats with unexpected names.

96LawDawg:
Should we add a turnCycle "index" as well?  It was suggested in #1222.  May be overkill, but sometimes you work with seat id and sometimes with index.  When you use position, you are getting the position in the array of active seats.  The index value of the seat and position are not always the same. 

I'm okay with what we have here.  Bringing it up just in case.

ArnoldSmith86 (approved):
Definitely makes it easier to review. The code clearly doesn't change anything unless `turnCycle: seat`. If anyone is using that already, it's their problem. :grin:

PR #1945, opened 2023-11-14
added public library game Frontiers (similar to Catan)

ArnoldSmith86:
This is the next game I want to finish for Finish Your Games November. @mousewax created amazing art for it and the base game is pretty much done.

It has been a while since I worked on this and the plan was always to have crazy board generation (which is partially implemented) and all expansions but for now I will try to get the base game into the public library.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1945/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1945-pr-test`.

96LawDawg:
After taking a very quick look, the only thing that stood out to me was the reset button does not properly move resource cards back where they belong.  It did strange things with piles in their holders after taking them out of my hand.

Also, when you do the How To Use section of the metadata, you'll want to explain about the Internals button.

ArnoldSmith86:
Well, the Internals button is obviously not supposed to stay that way...

ArnoldSmith86:
Yay, did some stuff. Not done yet though.

96LawDawg:
> > There are no tokens for longest army/road. I would really miss those if I played this.
> 
> There are. Not super obvious but oh well. Unless I don't understand what you mean.
> 

I see them now.

> > The robber completely covers the number tile. Makes it slightly hard to initially see which tile numbers were rolled. You know what would be really cool? Add a glow or something to the number token(s) that correspond to the dice roll.
> 
> I added optional highlighting of number tokens. And you don't need to see the number that's blocked by the robber. I could offset the robber though. House rules might require you to see the number.
> 

This is more than good enough.  I just like to be able to see in a glance which resource is shut down by the robber.  Seems like I always spend 5 seconds asking where is the other 4?  Oh, yeah.  Under the robber.  But the highlight is enough without an offset.

Good luck on the future expansions.  We are taking bets that it will be another 3 years.

But this looks and plays really nice.  Thank you finishing this.

PR #1946, opened 2023-11-15
only `RECALL` cards not in the holder already

ArnoldSmith86:
This is my suggestion to replace #1628 which was in the news lately.

That PR tries to make a similar change in `moveToHolder` instead which would improve performance for more cases than just `RECALL`. To me that is pretty intimidating because it is called in many places and @robartsd does not seem to be sure about the side effects either.

In practice there are potential slowdowns in:

- `MOVE` (which already checks for same source and destination)
- `SET property:parent` which in theory should check that the property did not change but piles make it trigger anyway and it probably is slow when changing the `parent` of many widgets this way
- `RECALL` which this PR should fix and in practice this probably solves 99% of slowdowns.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -1526,6 +1526,7 @@ export class Widget extends StateManaged {
                   cards = cards.filter(c=>!c.get('_ancestor'));
                 if(a.excludeCollection && excludeCollection)
                   cards = cards.filter(c=>!excludeCollection.includes(c));
+                cards = cards.filter(c=>c.get('_ancestor')!=holder&&!c.get('owner'));
                 for(const c of cards)
                   await c.moveToHolder(widgets.get(holder));
               }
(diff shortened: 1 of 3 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1946/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1946-pr-test`.

robartsd:
The only side effect I remember thinking about was ordering of widgets.

robartsd:
Went back and saw the concerns about impact for items in the holder that have an owner or if the holder has childrenPerOwner set. Those circumstances would be rare for RECALL but may not be for other functions.

ArnoldSmith86:
I prefer #1948 because it's basically the same as what `MOVE` does and it has less side effects. The performance gain seems to be basically the same.

PR #1947, opened 2023-11-15
feedback when pressing `INPUT` button

ArnoldSmith86:
Pressing a button in an `INPUT` overlay quite often triggers routines that take a while. This PR hides the Cancel button and changes the OK button to Working so the player knows that they hit the button. It also disables the buttons - not sure if that is necessary.

I will probably add something similar to clicks in the room but that needs more thought and can be its own PR.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -678,11 +678,14 @@ export class Widget extends StateManaged {
     }
 
     if(!go || this.evaluateInputOverlayErrors(o, variables)) {
-      showOverlay(null);
-      if(go)
-        resolve({ variables, collections });
-      else
-        reject({ variables, collections });
+      this.showInputOverlayWorkingState(true);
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
Please test for problems when having multiple `INPUT`s in the same routine.

96LawDawg:
> Please test for problems when having multiple `INPUT`s in the same routine.

Wow.  Who does that?  I've never even tested that.  Probably should have tested it at some point when we were messing around with `INPUT` but it just never occurred to me.  Will do.

96LawDawg (approved):
This is great!  It works in the Card Games PL room.  Even on my fast computer, switching games takes several seconds and this makes it clear what is happening.  I also tested multiple inputs in the same routine with no problem (both before and after the section that takes a long time).

Approved through test.

PR #1948, opened 2023-11-15
only `RECALL` cards not in the holder already

ArnoldSmith86:
Another alternative to #1628. This one should result in the same card order as production. TestCafe still has differences but only `z` values because cards are not moved multiple times, resulting in lower `z` values. I checked Create Game and Bhukhar tests and they now deal the same cards to the holders.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1948/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1948-pr-test`.

96LawDawg (approved):
Approved through test.  Whatever I thought I saw earlier does not seem to be a problem.

PR #1950, opened 2023-11-15
fixed PCIO import

ArnoldSmith86:
Apparently nobody tested older PCIO files with #1915.

### DIFF
--- a/server/pcioimport.mjs
+++ b/server/pcioimport.mjs
@@ -85,7 +85,7 @@ export default async function convertPCIO(content) {
         target[holdersParam] = args.objects.holders;
       }
     } else {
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1950/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1950-pr-test`.

96LawDawg (approved):
Approved through test.  I tested the same basic file (export of 500 room) from PCIO on production and it did not work.  Works here.  I don't have enough PCIO stuff to test a lot.  But I also tested on old file I have from 2021 that also works here but not on production.

PR #1956, opened 2023-11-17
Fix seat counter not working with piles

### LINKED ISSUE #1670: Seat counter does not work with stackOffset 0
https://virtualtabletop.io/issue1670

after there is a pile it stops counting

96LawDawg:
Fixes #1670.  I don't think we need a file updater for this.  People will have found a work around or not used the built in seat counter.  But this will make it right for any future games that use seat counters with piles in the hand.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1956/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1956-pr-test`.

ArnoldSmith86 (approved):
The change looks correct to me.

PR #1957, opened 2023-11-17
always show all symbols for metadata icon picker

ArnoldSmith86:
If you used the icon picker in the JSON editor, it gets filtered to either fonts or images. In production that filter is not reset when using it in the metadata again. This PR fixes that.

I also noticed that leaving the icon picker open in the metadata and then switching to edit mode leads to problems. So this also closes the icon picker in the Game Shelf whenever you use it somewhere else.

### DIFF
--- a/client/js/domhelpers.js
+++ b/client/js/domhelpers.js
@@ -466,6 +466,9 @@ export async function loadSymbolPicker() {
 }
 
 export async function pickSymbol(type='all', bigPreviews=true, closeOverlay=true) {
+  if($('#statesButton').dataset.overlay == 'symbolPickerOverlay')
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1957/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1957-pr-test`.

96LawDawg (approved):
I didn't test how this compares to the scenarios you described on production.  But it seems to work as expected in the PR.  Approved through test.

PR #1961, opened 2023-11-21
PL Add Terra versus Aqua

96LawDawg:
Similar to Chartae (by PaladinOfCats)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1961/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1961-pr-test`.

ArnoldSmith86:
I don't know why but the simplicity of the images compelled me to optimize it. They were already tiny so this totally wasn't justified but now the JSON is 401 bytes smaller and doesn't use any images anymore. Even the library image is now an inline SVG with 266 bytes.

This makes it a lot more work to replace the shapes by prettier artwork though.

ArnoldSmith86 (approved):

PR #1963, opened 2023-11-22
fixed crash when editing an in-progress game save

ArnoldSmith86:
Especially if you create an in-progress save game, remove the origin game and then enter edit mode, it would crash because the Save button would try to read from the deleted game.

### DIFF
--- a/client/js/editor/toolbar/save.js
+++ b/client/js/editor/toolbar/save.js
@@ -37,11 +37,12 @@ class SaveButton extends ToolbarButtonWithContent {
 
   onMetaReceived(data) {
     this.activeState = data.meta.activeState;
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1963/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1963-pr-test`.

96LawDawg (approved):
Approved through test.  I tested the stated condition here and on production.  I also tested saving, downloading, uploading, and switching between various types of saved and regular games.

PR #1966, opened 2023-11-22
added a button that converts an in-progress game into a normal game

ArnoldSmith86:
The button Save Active Game causes confusion and often leads to people saving their game as an in-progress game. My hope is that the new save button in the editor improves the situation but this PR adds a button in the UI to convert an in-progress game into a normal one.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1966/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1966-pr-test`.

96LawDawg:
When there is only one in progress game, if you convert it, the "Update Loaded Game" button is still available at the top of the Game Shelf area.  That button should be hidden because it does nothing if there are no in progress games.  I never noticed this before, but this is also an issue on production.

Also, should we make the Convert option be the very top one (above download)?  I think so, because it isn't that often you really want to download an in-progress game.

96LawDawg:
The "Update Loaded Game" button still remains after converting even when there is no in progress game.

Screenshot 2024-02-07 094931

96LawDawg (approved):
Approved through test.  The update button is now removed when the state does not exist.

PR #1967, opened 2023-11-22
Adds Carnival similar to Parade

ArnoldSmith86:
This supersedes #1448 which had merge problems after the UI changes were live.

I implemented the carnival costume cards and tweaked the layout a bit. I also added attribution.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1967/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1967-pr-test`.

96LawDawg:
Three suggestions:

Implement turns.  All you need is to put a random turn on the play button, add turn to the routine in the up arrow, and add a turn button so players can manually adjust as needed.

Put an INPUT prompt on the play button because it is too easy to mess up the room if you press that while a round is in progress.  Maybe even better, combine the play and reset buttons.  You don't really need a reset button in this room if the play button resets the room and deals as needed.

Add a function to the up arrow button to MOVE a card from the draw pile to the seat/hand.

96LawDawg (approved):
Approved.  I had made a few suggestions.  They are not essential.  I will merge on Nov 30 either way.

ArnoldSmith86:
I played the game with my wife:
- For 2 players, you get way too many cards for your holder. (We played until Draw was empty.)
- Auto-draw would be nice.
- Scoring could use some automation.
- The arrow button is movable.
- Rules don't state that the last round starts if one player has all colors.

96LawDawg:
> I played the game with my wife:
> 
> * For 2 players, you get way too many cards for your holder. (We played until Draw was empty.)

Fixed.  I rearranged the board and added more space.  Before, each area held 18 cards.  Now they hold 28.  It is possible in a 2 player game for overflow, but only if the play is really unbalanced.

> * Also: the 2 rows flow from one to the other in an unintuitive way. I don't quite remember how.

Fixed.  It MOVE'd a clump so it didn't respect the sort.  Now it MOVEs 1 at a time in a loopRoutine. 

> * Auto-draw would be nice.

Added.

> * Scoring could use some automation.

Not implemented in this PR.

> * The arrow button is movable.

Fixed.

> * Rules don't state that the last round starts if one player has all colors.

Fixed.

> Three suggestions:

> Implement turns. All you need is to put a random turn on the play button, add turn to the routine in the up arrow, and add a turn button so players can manually adjust as needed.

Added.

> Put an INPUT prompt on the play button because it is too easy to mess up the room if you press that while a round is in progress. Maybe even better, combine the play and reset buttons. You don't really need a reset button in this room if the play button resets the room and deals as needed.

Added.

> Add a function to the up arrow button to MOVE a card from the draw pile to the seat/hand.

Done as noted above.

PR #1971, opened 2023-11-23
fixed client crash when `CLONE` fails

ArnoldSmith86:
There are not many things that make adding a widget fail but `parent`, `deck` and `cardType` have to be correct. This makes `CLONE` report problems instead of crashing the client. It also prevents the broken widget from ending up in the room state.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1971/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1971-pr-test`.

96LawDawg:
Initial testing done.  This catches `deck` and `cardType`, but not `parent`.

96LawDawg:
CLONE operations always fail now.

96LawDawg (approved):
Approved through test.  CLONE now works again since the last change.  If invalid properties for deck, cardType, or parent are present, you get the debug error message.  Other custom or existing properties work as expected.

PR #1972, opened 2023-11-23
fixed `IF` complaining about missing `condition`

ArnoldSmith86:
I use `condition` to check for a variable all the time. But if the variable resolves to `undefined`, `IF` complains about missing a `condition` in the first place.

This PR changes the checks for `IF` to look at the code instead of the resolved values. This fixes the incorrect problem messages but it also changes behavior:

Previously the `elseRoutine` would not be executed if `condition` and `operand1` would evaluate to `undefined`. Now they should be executed.

This probably breaks games. I will think a second about a fileupdater but it might make existing code too ugly to justify it.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -1334,15 +1334,15 @@ export class Widget extends StateManaged {
           problems.push(`Relation ${a.relation} is unsupported. Using '==' relation.`);
           a.relation = '==';
         }
-        if(a.condition !== undefined || a.operand1 !== undefined) {
+        if(original.condition !== undefined || original.operand1 !== undefined) {
           let condition = a.condition;
-          if (condition === undefined)
+          if (original.condition === undefined)
             condition = await compute(a.relation, null, a.operand1, a.operand2);
           const branch = condition ? 'thenRoutine' : 'elseRoutine';
           if(Array.isArray(a[branch]))
             await this.evaluateRoutine(a[branch], variables, collections, (depth || 0) + 1, true);
           if(jeRoutineLogging) {
-            if (a.condition === undefined)
+            if (original.condition === undefined)
               jeLoggingRoutineOperationSummary(`'${original.operand1}' ${a.relation} '${original.operand2}'`, `${JSON.stringify(condition)}`)
             else
               jeLoggingRoutineOperationSummary(`'${original.condition}'`, `${JSON.stringify(condition)}`)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1972/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1972-pr-test`.

ArnoldSmith86:
A fileupdater would have to make sure that `elseRoutine` will not be called if `operand1` and `condition` evaluate to `undefined`.

Oh, and I also changed that if you give _both_ `condition` _and_ `operand1 == operand2` it would previously use `condition` _unless_ it evaluates to `undefined` (in which case it would use `operand1 == operand2`). Now it only uses `operand1 == operand2` if there is no `condition` in the code.

---

For the case of `condition === undefined`,

```JSON
{
  "func": "IF",
  "condition": "${variable}",
  "thenRoutine": [],
  "elseRoutine": [
    "var example = 1"
  ]
}
```

would become

```JSON
{
  "func": "IF",
  "condition": "${variable}",
  "thenRoutine": [],
  "elseRoutine": [
    {
      "func": "IF",
      "operand1": "${variable}",
      "relation": "===",
      "operand2": "${dummyVariableThatIsUndefined}",
      "elseRoutine": [
        "var example = 1"
      ]
    }
  ]
}
```

---

I might revert the change when providing _both_ `condition` _and_ `operand1 == operand2`. Probably not worth it to change behavior for something that nobody uses anyway.

96LawDawg:
Initial testing shows the primary change (undefined condition) works.  I agree that this should have been the behavior all along, but also that it will probably break games so we need a fileupdater.

>I might revert the change when providing both condition and operand1 == operand2. Probably not worth it to change behavior for something that nobody uses anyway.

I don't think we should change this.  Designers are not supposed to use all 3 (condition, op1, op2).  Whatever is working or not on production now shouldn't be touched for an unsupported scenario.

ArnoldSmith86:
- There are 2251 `IF`s in the public library.
- 853 of those have an `elseRoutine`.
- 485 do _not_ check a `PROPERTY` (which should never return `undefined`).

_Investigation to be continued..._

ArnoldSmith86:
So... I'm really trying to keep the file updater changes as small as possible. The biggest thing I can't rule out in the file updater is this:

```JSON
  "clickRoutine": [
    {
      "func": "SELECT",
      "value": "NOPE"
    },
    {
      "func": "GET",
      "variable": "myID"
    },
    {
      "func": "IF",
      "condition": "${myID}"
    }
  ]
```

Unfortunately, `GET` doesn't set its variable at all if the given `collection` is empty. By looking at the code, I can be sure that collection `DEFAULT` _exists_ (because of the `SELECT`) but I can't know that it's _not empty_.

So at the time of writing this, the file updater would would make 262 `IF`s in the public library pretty ugly. Here is an example from Yoink:

Before:

```JSON
[
  "//Rotate scouted card 180 degrees.",
  {
    "func": "SELECT",
    "type": "card",
    "property": "owner",
    "value": "${playerName}"
  },
  {
    "func": "SELECT",
    "source": "DEFAULT",
    "property": "scouted",
    "value": true
  },
  {
    "func": "GET",
    "property": "rotation",
    "variable": "scoutedRotation"
  },
  {
    "func": "IF",
    "operand1": "${scoutedRotation}",
    "operand2": 0,
    "thenRoutine": [
      {
        "func": "SET",
        "property": "rotation",
        "value": 180
      }
    ],
    "elseRoutine": [
      {
        "func": "SET",
        "collection": "DEFAULT",
        "property": "rotation",
        "value": 0
      }
    ]
  }
]
```

After:

```JSON
[
  "//Rotate scouted card 180 degrees.",
  {
    "func": "SELECT",
    "type": "card",
    "property": "owner",
    "value": "${playerName}"
  },
  {
    "func": "SELECT",
    "source": "DEFAULT",
    "property": "scouted",
    "value": true
  },
  {
    "func": "GET",
    "property": "rotation",
    "variable": "scoutedRotation"
  },
  {
    "func": "IF",
    "operand1": "${scoutedRotation}",
    "operand2": 0,
    "thenRoutine": [
      {
        "func": "SET",
        "property": "rotation",
        "value": 180
      }
    ],
    "elseRoutine": [
      {
        "note1": "added by file updater to simulate old behavior where elseRoutine would not be called if operand1 is undefined",
        "note2": "replace this IF by its thenRoutine contents if you are sure that operand1 will never be null",
        "func": "IF",
        "operand1": "${scoutedRotation}",
        "relation": "!==",
        "operand2": "${fileUpdaterUndefinedVariable}",
        "thenRoutine": [
          {
            "func": "SET",
            "collection": "DEFAULT",
            "property": "rotation",
            "value": 0
          }
        ]
      }
    ]
  }
]
```

---

Is this worth it?

96LawDawg:
This is probably a dumb idea, but here goes.  What if we just leave it as is?  Undefined variables have a place in javascript.  What place do they serve in VTT JSON?  It probably means I screwed up in writing the function.  Right now, the IF completely fails and that has in the past clued me in of the problem.  With the change, the `elseRoutine` will fire and I may have trouble figuring out why.  Just a thought.

ArnoldSmith86:
My main motivation is that I frequently use this if I only want to execute something if an array index exists:

```JSON
    {
      "func": "IF",
      "condition": "${array.index}"
    }
```

That works fine. But it triggers red arrows in the debugger that might "hide" "real" problems. But if I simply remove the problem message, it is weird that `elseRoutine` isn't executed.

But yeah, maybe we just get used to it...

96LawDawg:
> My main motivation is that I frequently use this if I only want to execute something if an array index exists:
> 
> ```json
>     {
>       "func": "IF",
>       "condition": "${array.index}"
>     }
> ```

It's one additional step, but wouldn't this also work?

```
   " var indexCheck == ${array.index}" 
   {
      "func": "IF",
      "condition": "${indexCheck}"
    }
```

ArnoldSmith86:
Yes. But adding a 140 lines file updater would also work. :sob:

ArnoldSmith86:
And somehow the file updater doesn't affect any of the games tested with TestCafé... Affected:

```
library/games/A Simple Life
library/games/Adedanha
library/games/Agra
library/games/Alpaca
library/games/Capture the Flag
library/games/Card Games
library/games/Cards Against Humanity
library/games/Chess
library/games/Dreamcatcher
library/games/Expeditions
library/games/Flight
library/games/Forces of Nature
library/games/Fowl Play
library/games/Going Nuts
library/games/Hangman
library/games/Hearts
library/games/Hexagonal Chess
library/games/Imagination
library/games/Knaves
library/games/Liar's Dice
library/games/Magician
library/games/Medieval VNO
library/games/Meta
library/games/Pennsylvania Rush
library/games/Property Tycoon
library/games/Rackball
library/games/Riverside Rulers
library/games/Road Race
library/games/Scrivener
library/games/Speedy Fruits
library/games/Spin 5
library/games/Sunshine and Rainbows
library/games/Tutti Frutti
library/games/Ultra
library/games/Wall Street Bets
library/games/Word Whirl
library/games/Yoink
library/tutorials/Color
library/tutorials/Piles
library/tutorials/Routines
```

bjalder26:
> Yes. But adding a 140 lines file updater would also work. 😭

Would it be a good idea to break the file updater into 1 file per version, then they all get appended?  Would that make it more manageable?

PR #1979, opened 2023-11-25
PL Update Wordy

96LawDawg:
I completely refactored Wordy.  This removes about 500 widgets.  I deleted all the built in decks.  I now have one master deck and make clones of the tiles as needed.  The custom "spreadsheet" area 150 labels with the tile points and distribution was replaced by a single widget using html.  The switch between languages is now almost instant instead of taking several seconds.

I also changed the swap symbol in the room and updated the meta data and compressed the images.

But otherwise, game play is the same.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1979/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1979-pr-test`.

96LawDawg:
Added support for Greek language.  Because I feel like adding a whole new language sort of counts towards the November updates.

96LawDawg:
Added Spanish and Italian (to complete the Romance language starter pack).  Although the original meta data had Spanish listed, I had not implemented Spanish mainly because there were 3 variants (International, North American, and Latin American).  But with the refactoring I did, adding those 3 was easy and didn't add hundreds more widgets.  BUT, I don't have a good way to indicate language support for Spanish using our traditional 2 letter language - 2 letter country.  So I just did the 2 letter language.

96LawDawg:
I did another refactoring using the system I devised for Card Games.  Each language is now its own widget hidden off screen.  That is used to create the option menu and the tiles.  And it is now ridiculously easy to create a new tile.  Just feed ChatGPT the list from the wikipedia entry listed in the meta data and the JSON format and bam!

I did this to test and push the boundaries of the language listing and non-Latin character recognition of VTT.  The character recognition is good.  I'll open an issue for the language listing problem I found.

ArnoldSmith86 (approved):
I don't think I'll have time to review.

PR #1982, opened 2023-11-27
PL: Add Rockets and Robots (similar to Wandering Towers)

andypymont:


96LawDawg (approved):
I can't find any flaws in this game.  Looks great.  Works great.

One thing could be more clear in the instructions.  What do you do after you roll the die?  Is that how many spaces you move?  Seems obvious, but it isn't stated in the rules.

Approved through test.

PR #1990, opened 2023-12-01
PL Update Carnival

96LawDawg:
Fixes the auto deal so it does not happen once a player has all 6 colors.

This update should also fix the images for the instructions, but I didn't have time for that.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1990/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1990-pr-test`.

ArnoldSmith86 (approved):
Did a play test. Worked better than I did. I thought it was broken because it gave me a card after I had all colors. But that was correct because I got one more turn.

Didn't have space issues but this time we knew about the color end condition. We didn't even need a second row.

96LawDawg:
I added a count of cards, reversed the flow on the bottom so it goes from right to left, and updated the instructions.  None of the automation was touched, so this should be good and will be merged later.

PR #1991, opened 2023-12-02
PL Update Carnival with Automatic Scoring

96LawDawg:
Adds a button (and updates the instructions) for automatic scoring.

You can test this by just dragging cards into the seat areas and pressing the button.  No need to actually sit in seats or play the game.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1991/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1991-pr-test`.

96LawDawg:
2 player scoring fixed

ArnoldSmith86:
Scoring seems to work fine.

Some notes (not necessarily requests):
- Sorting of the second row is still weird. Top row goes left to right and then it continues right to left in the second row.
- Would be nice if it sorts the rows when dropping the final cards in there.
- Some kind of indication of the last round would be nice.

ArnoldSmith86 (approved):

96LawDawg:
> Some notes (not necessarily requests):
> 
> * Sorting of the second row is still weird. Top row goes left to right and then it continues right to left in the second row.

Done

> * Would be nice if it sorts the rows when dropping the final cards in there.

Mostly done. If you drop the card into the correct top or bottom row it works.  It was more complicated than it was worth to try to move a card between rows.

> * Some kind of indication of the last round would be nice.

Done.  The scoreboard button is hidden by default.  It now appears in the room to indicate the last round (updated instructions to reflect that).

PR #1993, opened 2023-12-04
Put focus on INPUT string field

96LawDawg:
Requested on Discord.  If there are multiple string fields, the last one gets the focus.  I could not figure out how to fix that despite multiple tries.  So as is, this is probably good enough for most uses that only call for a single string field.  But if someone can figure out how to get the focus only on the first string, that would probably be better.


96LawDawg:
> Does it not work without `setTimeout`?

No.  ChatGPT said: " The focus might not be working because the element is not yet in the document when the focus() method is called.  You can use the setTimeout function to delay the focus until after the current call stack has cleared. Here's the updated code:"

That makes it work, except as noted for multiple string inputs.

96LawDawg:
I don't know how to make this pass TestCafe and I do not understand really what you mentioned above.

Also, the Enter key does not work on string fields but will work for other fields.

ArnoldSmith86 (approved):
Please do a few tests.. :smile:

96LawDawg:
Everything seems to work.  See if anybody else chimes in with any tests based on whatever game they were working on and merge whenever.  Thank you.  This was trickier than I thought to code.

PR #1996, opened 2023-12-12
PL Update Carnival to Fix Sort Problem

96LawDawg:
There was a problem in the last round with sorting.  This fixes that.

It also fixes the problem that if the player changed their mind after staging a card to play, the previously selected cards would also move to the player's hand.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-1996/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR1996-pr-test`.

ArnoldSmith86 (approved):
No further comments from the last play test.

PR #2007, opened 2023-12-26
Seat Redesign

96LawDawg:
Back in mid October, there was a lot of discussion and work done on re-designing the seats.  Around here in discord: https://discord.com/channels/770758631146782780/770758631146782783/1163543505310847006. 

I took a shot at getting it started.  It mostly uses the click logic and style that @mousewax came up with.

I initially had done the logic as a clickRoutine on the seat.  But that adds so much stuff that should be done behind the scenes and would be intimidating to change for a newcomer.  But putting it into the code, like I've done here, also creates issues.  The css changes and the code changes will not be compatible with some exisiting games.  For example, see Feline.

Also, in seats.js around line 147-150 I try to make changes to the seat make changes to the actual playerName and playerColor.  For playerName, it works, but the screen flashes.  And playerColor doesn't work.

But other than these huge problems, this is a pretty good start.

Happy New Year.

### DIFF
--- a/client/css/widgets/seat.css
+++ b/client/css/widgets/seat.css
@@ -1,22 +1,49 @@
 .seat {
-  background-color: #ffffff;
+  background-color: var(--color);
 
-  font-size: 25px;
-  color:#222222;
+  font-size: 21px;
+  color:#ffffff;
   font-weight:bold;
 
-  border-width: 7px;
+  border-width: 2px;
   border-style: solid;
-  border-color: var(--color);
+  border-color: #000000bb;
+
+  box-sizing: border-box;
+  box-shadow: 0 2px 5px #00000066;
+  text-shadow: 2px 2px 2px #000000bb, -2px 2px 2px #000000bb, 2px -2px 2px #000000bb, -2px -2px 2px #000000bb;
 
   display: flex;
   justify-content: center;
   align-items: center;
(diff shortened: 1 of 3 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2007/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR2007-pr-test`.

96LawDawg:
The css changes also effect the hand counter widget.  Altogether, if this PR stays roughly as it, we would need a file updater that puts something like the old css back onto existing seats.  But that will be hard because many seats already have css.  And we would need some sort of clickRoutine updater that makes old seats behave like they did before.  I don't know if either of those are possible or not.

RaphaelAlvez:
no time for a proper review just some things I wanted to say
- I prefer this design except for the text-shadow. I think it really makes the text harder to read
- the color and name input give the impression the player settings are being changed when in fact they are not
- maybe this could be a click routine instead of a click action. would cause less trouble with older games
- @ArnoldSmith86 maybe default seat styles in the editor?

RaphaelAlvez:
Just pointing out that I really like this PR but I'm concerned with games that could have used click targeting a seat

PR #2009, opened 2023-12-27
limit card activeFace to highest valid face

ArnoldSmith86:
When setting `activeFace` to something higher than the highest valid face, the card will be blank. This PR will display the highest face instead.

I did not test this change at all and I am not even sure that we should do it.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2009/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR2009-pr-test`.

96LawDawg:
This works.  But then we end up with 3 different behaviors for different widget types:

Cards: Show only the highest face
Basic widgets: Empty face like cards were before this PR
Dice: Can set to whatever you want and the face ends up being a multiple of the total number of sides (e.g., face 11 = six pips on a standard die.

96LawDawg:
I like the new change where cards work like dice.  This way incrementing the face is just like clicking on it.  That makes the most sense.  And if a designer sets a specific activeFace using a routine, then it is on the designer to make sure that face (or a multiple of it, if looping) actually exists and matches what is intended.

Do you plan to make basic widgets act the same way?  I think we should.

ArnoldSmith86:
> Do you plan to make basic widgets act the same way? I think we should.

I don't know. I feel like faces of basic widgets are different. They apply stuff _on top of_ the properties that are always active. Cards and dice basically have undefined appearance with a non-existent face.

Apart from that I'm not sure what it would break. Might be acceptable.

Either way, it should not be in this PR.

96LawDawg:
> I feel like faces of basic widgets are different. They apply stuff _on top of_ the properties that are always active. Cards and dice basically have undefined appearance with a non-existent face.

Good point.  They are different in that way.

> 
> Apart from that I'm not sure what it would break. Might be acceptable.
> 
> Either way, it should not be in this PR.

We don't *need* to do this and I don't want to break anything.  Just thinking about consistency, but because of your previous point, this makes sense.

96LawDawg (approved):
Approved through test.

PR #2024, opened 2024-01-26
updated Noto Emoji to Emoji Version 15.1

ArnoldSmith86:
See https://emojipedia.org/emoji-15.1 for the emoji that are now available. The symbol picker also did not show some older emoji that were already available as images.

The family emojis changed significantly:

image

If a game uses those, it would look a lot different. We could keep the old images for those but I think I would prefer to keep Noto Emoji as is.

Here is a side-by-side comparison of all existing image files that changed:

image

I don't see any difference except for the family ones. Probably barely visible fixes.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2024/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR2024-pr-test`.

96LawDawg (approved):
Approved.  This is certainly early enough that probably no one has used the family emojis anyway.  Only 2 PL games use emojis so far and not those.

PR #2029, opened 2024-02-02
added resize to the editor dragtoolbar

ArnoldSmith86:
Can probably be improved but better than not having it...


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2029/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR2029-pr-test`.

RaphaelAlvez:
I don't get what this changes

96LawDawg:
Agree.  Better than nothing.  Things that need to be changed before merge:

- [ ] Keep the drag handle on the far right. Swap places with it and the new resize icon.

Things that would be nice to have but not needed before merge:

- [ ] Holding shift (or some other way) while dragging keeps the same size ratio so a circle stays a circle. Most graphics editors offer something similar.

Unrealistic expectations but would be great:
- [ ] Resizing into negative x and negative y values works.  Of course, that would require changing the actual x and y values of the widget to end up in the new position.
- [ ] Multiple select resizing keeps the same spacing between the widgets so they don't overlap.

ArnoldSmith86:
> Holding shift (or some other way) while dragging keeps the same size ratio so a circle stays a circle. Most graphics editors offer something similar.

Yeah, that's needed for Move along a single axis as well. But it needs some foundational work.

ArnoldSmith86:
I "solved" the aspect ratio thing by just adding a second button to the toolbar.

My plan was always to have a toggle button in that toolbar for turning the aspect ratio thing on and off and then using something like holding `SHIFT` to make it easier on a PC.

But I realized a few days ago that it's kinda weird to use the same toggle button for "only move on one axis" and "resize while keeping aspect ratio". And if I add multiple toggle buttons, I might just as well add multiple buttons for each operation anyway.

This probably isn't the best solution. Maybe we can come up with something better in the future.

96LawDawg (approved):
Approved through test.

PR #2039, opened 2024-02-05
World Whirl: fixed correct rotation detection

ArnoldSmith86:
When the player board is rotated to the left a bunch, it has a negative angle big enough to result in a negative angle when comparing to the correct angle. And modulo is not enough to map to 0...360. This now adds a simple check that fixes negative angles.

### DIFF
--- a/library/games/Word Whirl/0.json
+++ b/library/games/Word Whirl/0.json
@@ -5560,6 +5560,15 @@
           "var offsetCardRotation = ${CurrentCardRotation} % 360",
           {
             "func": "IF",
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2039/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR2039-pr-test`.

96LawDawg (approved):
Approved through code review.  Did not test.

PR #2041, opened 2024-02-08
fixed changing the widget type in the JSON editor

ArnoldSmith86:
Changing the type would work before but you would not see the results until after you reloaded the room.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2041/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR2041-pr-test`.

96LawDawg:
This is alchemy.  Or maybe sorcery.  I can turn a button into a holder and then into a dice with no problem.  This has to break something.  I'll keep testing later.

PR #2042, opened 2024-02-08
removed type when adding basic widget

ArnoldSmith86:
The command `Add basic widget` in the JSON editor currently adds `type: basic` to the added widget implying that that is required. In fact any value for `type`, that is _not_ one of the other widget types, results in a basic widget.

### DIFF
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -1336,7 +1336,7 @@ function jeAddWidgetPropertyCommands(object, widgetBase) {
     context: 'No widget selected.',
     onEmpty: true,
     call: async function() {
-      setSelection([ widgets.get(await addWidgetLocal({type})) ]);
+      setSelection([ widgets.get(await addWidgetLocal(type == 'basic' ? {} : {type})) ]);
     }
   });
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2042/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR2042-pr-test`.

96LawDawg:
Either here or in #2041 we should add "type" into the JSON editor's list of clickable properties.

ArnoldSmith86:
> Either here or in #2041 we should add "type" into the JSON editor's list of clickable properties.

Hmm, tricky. The editor only shows command buttons for properties that apply to the current widget type. A widget without `type` is type basic which doesn't have a `type` property. We could fake it, of course.

96LawDawg (approved):
Approved through test.  I never noticed this before.  I don't use the JSON Editor to add widgets.  I've just been using the overlay from the menu at the top.  But, yeah, that is unncessary.

PR #2053, opened 2024-03-06
PL Improvements to Find the Word

96LawDawg:
This update:
- [x] Fixes all the broken labels
- [x] Auto-detects the correct number of letters and a correctly guessed word and suggests the button the non-guessing player should press
- [x] Allows player waiting on the other player to guess to go ahead and pre-load their next word
- [x] Adds meta-data explaining how to use the implementation
- [x] Compresses the background image


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2053/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR2053-pr-test`.

rogerl50:
It looks like guessed letters always appear in gray, even if they have been excluded due to a zero guess? Did it always work that way?
In the code, which I didn't examine line-by-line, it seems like you switched around the color names. Some of the "white" were changed to "notInWord" and others to "inWord". Was that intentional?

96LawDawg:
> It looks like guessed letters always appear in gray, even if they have been excluded due to a zero guess? Did it always work that way? (See the current contents of the PR-test room. The letters in CHIPS cannot be in the first player's word, yet they are gray in the second players display.)

Yes.  It works that way on production.  I made both alphabet holders visible in the room for comparison and duplicated on production what you had done in the PR test room.  Images are identical (production first, new version second).

Production

New Version


> In the code, which I didn't examine line-by-line, it seems like you switched around the color names. Some of the "white" were changed to "notInWord" and others to "inWord". Was that intentional?

Not intentional, but it doesn't seem to have an impact on game play.  This is a result of the reset function not fully resetting everything.  Some of that gets reset when the game actually starts (after both players enter hidden words).  So those changes are as a result of me playtesting it and saving it.  It would be possible to have the reset button clear all that out, but I didn't think it was worth tracking down which routine in the game start section was responsible.

96LawDawg:
Actually, my explanation was not correct for the sqColor thing.  The reset button should have done that.  It does it in other places; it just didn't do it for the "letterSquare" pieces.  I added a SET into the reset button that takes care of that now.

rogerl50 (approved):
Better than it was. Thanks. I still hate that there is no way to type when there is a keyboard, and no way to mark letters in the grid as Yes, Maybe, No. But as far as I can tell, this is good to go.

96LawDawg:
Thanks.  I will play test soon and then merge.

PR #2068, opened 2024-03-19
fixed that saving from the editor does not update lastUpdate

ArnoldSmith86:
See title.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2068/pr-test (or any other room on that server)

After merging, a backup will be available at `https://beta.virtualtabletop.io/editor/PR2068-pr-test`.

96LawDawg (approved):
I didn't see the problem with the production version.  When I saved using the edit mode button "save to active variant" I got a new lastUpdate value in the diff.  But this also works through test.  Approved.

Issue #1954, opened 2023-11-16
skipTurn does not clear out existing turn


96LawDawg:
If a seat has "turn:true" and then skipTurn is set to true, using the TURN function will not erase the "turn:true" and multiple seats then have the turn indicator.

PR #2080, opened 2024-03-28
Rework turn code for source collections

### LINKED ISSUE #2079: `TURN` with `source` collection does not work properly
To duplicate, create a room with multiple seats and set the turn to true on one of them.  Then create a collection of one or more of the other seats where turn is not true.  Use `TURN` with the `source` set to the collection.  The turn indicator will move to the next seat in line, but it does not turn off for the one that was originally selected.  This is probably the same problem from #1954 and #1955.

rogerl50:
This fix backs out some of the recent changes here relating to #1954 and simplifies the code around turning `turn` on and off in order to fix #2079. I didn't really understand the content of #1954, so someone should check that I didn't inadvertently re-break that issue.


96LawDawg (approved):
Well, that is a lot cleaner.  It fixes what I probably broke in #1954 regarding collections and the problem that I was having there is also fixed.  Approved through test.

PR #2094, opened 2024-04-13
Public Library - Fix Ascent

96LawDawg:
Fix: dice that are unavailable when a climber is at the top of a column are now hidden.
Fix: when changing seats, all associated widgets now change color.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2094/pr-test (or any other room on that server)

ArnoldSmith86 (approved):

PR #2096, opened 2024-04-15
Fix typo in player.JS

96LawDawg:
This fixes a small bug identified in #1774.  I rediscovered this when I was trying to make some custom css using "content" to add the playerName to the cursor. Fixing this bug allows the change without having to use "attr(data-player)" which is a lot less obvious and not all documented (but also works).

I also added a tutorial in CSS: Caution to show how to do what was requested with the visible cursors and playerNames.  I believe this solution will mean we can close #1774.  it will not be implemented, but it will be easy enough with css (and the tutorial) to add to rooms.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2096/pr-test (or any other room on that server)

96LawDawg:
Okay, this is strange.  I tested this before committing and it worked.  But I wonder if I accidentally still used "attr" in my local test.  Because that works and --playerName does not despite removing the quotation marks around it in the players.js.  So this change might not do anything.  That's still okay.  The effect can still be achieved using "attr".  We can just make this a tutorial change instead of an engine change if need be.

ArnoldSmith86:
Doesn't that need to be properly quoted with `"` so it works with spaces in player names? And didn't we have a PR for that already? Why didn't we merge that one? 🤣

ArnoldSmith86 (approved):

PR #2110, opened 2024-05-16
fixed inconsistencies in scoreboard CSS

ArnoldSmith86:
The CSS variables `--firstColWidth` and `--columns` were not set reliably and disappeared whenever the CSS was updated without redrawing the entire scoreboard table.

This PR also introduces property `firstColWidth` so that the `50px` can be changed.

### DIFF
--- a/client/js/widgets/scoreboard.js
+++ b/client/js/widgets/scoreboard.js
@@ -16,6 +16,7 @@ class Scoreboard extends Widget {
       roundLabel: 'Round',
       totalsLabel: 'Totals',
       scoreProperty: 'score',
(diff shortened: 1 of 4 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2110/pr-test (or any other room on that server)

96LawDawg (approved):
Updated the tutorial to reflect the new property.  Will need to be added to the wiki when merged.

Approved through test.  This fixes the specific use case scenario that led to the PR and it does not mess up any existing functionality.

PR #2112, opened 2024-05-17
fixed that basic widgets would lose their children when editing their `html`

ArnoldSmith86:
Now the children should stay in place and adding/removing `html` or `text` properties should update the widget as expected.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2112/pr-test (or any other room on that server)

96LawDawg (approved):
Basic widget functions and html tested and work. And this change does what it says: children of basic widgets having html editing no longer lose their parent. Approved through test.

PR #2113, opened 2024-05-17
fixed that `hidePlayerCursors` would not apply when dragging another widget

### LINKED ISSUE #2105: hidePlayerCursors does not work when clicking or dragging
The current functionality of `hidePlayerCursors` should be fixed so that the cursor is hidden everywhere inside the boundary of the widget no matter what.

ArnoldSmith86:
This should fix #2105 by checking all widgets at the cursor position and not just the top one.

This does change the behavior slightly in another way: if the parent of a widget has `hidePlayerCursors`, the child would also hide the cursor even if it is outside the parent widget area. Now it only hides it when inside the parent widget area. I am not sure what is the preferred behavior.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2113/pr-test (or any other room on that server)

96LawDawg:
This fixes the issue with seeing the cursor in a holder (hand or not) while dragging things over the holder.

I think it makes sense to not hide the child widget if the cursor is outside the parent widget as a general rule. But there is still one thing about this that I think could be improved. If the child widget is being dragged out of the parent, then the cursor should remain invisible until the dragging stops.  The reason this matters is for hands.  So dragging a card out of a hand, the cursor is invisible until the cursor leaves the hand.  In some games, that still gives information about what the player is doing. (Of course, so does the position on the board where the card back appears--can't do anything about that). But would it be possible and a good idea to keep the cursor invisible when dragging a widget out of a parent?

96LawDawg:
One other very small thing that does not need to be resolved, but I wanted to point it out.  If an editable label is hiding cursors, then the cursor does not disappear immediately, but it fades out over a few seconds like when the player stops moving it.  In other widget types that hide cursors, the cursor immediately disappears.

ArnoldSmith86:
This should be a lot better now.

> If an editable label is hiding cursors, then the cursor does not disappear immediately, but it fades out over a few seconds like when the player stops moving it.

This is actually basically unrelated. The custom mouse handling code doesn't run at all inside labels I think, so the cursor location was and is never sent while over a label.

96LawDawg:
> But there is still one thing about this that I think could be improved. If the child widget is being dragged out of the parent, then the cursor should remain invisible until the dragging stops.

I won't hold up approval for this, but this part is still not implemented.  Is this possible?  A good idea?  Everything else works as expected.

ArnoldSmith86:
> > But there is still one thing about this that I think could be improved. If the child widget is being dragged out of the parent, then the cursor should remain invisible until the dragging stops.
> 
> I won't hold up approval for this, but this part is still not implemented.  Is this possible?  A good idea?  Everything else works as expected.
> 
> 

It should work as long as the card itself is still invisible. Can you provide an example that doesn't work yet?

96LawDawg:
We are not talking about the same thing.  It does work while the card is invisible.  I was hoping that it could work during the entire drag if it came out of a holder with the new property set to true. But now that I think about it again, that does not make any sense. Once the other players can see the card pop up, it doesn't matter whether they can see the cursor or not.

96LawDawg (approved):
Approved through test.

PR #2118, opened 2024-05-18
fixed that removing the `type` property would not change the widget to a basic widget

ArnoldSmith86:
I noticed the bug a few days ago.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2118/pr-test (or any other room on that server)

96LawDawg (approved):
Thank you.  That's another bug I noticed and failed to open an issue for.

Approved through test and code review.  Removing a widget's type now changes it back to a basic widget.

Since changing types is essentially fully supported within the editor, do you think we should add a "type" button to the list of buttons in the editor to making switching easier?

PR #921, opened 2021-11-25
Adds text to holders

RaphaelAlvez:
as requested in #407

Text goes into the center of the holder but that can be changed in CSS as usual

image

PR #2126, opened 2024-05-21
Add images to holders

96LawDawg:
This is one of the things I need the most.  In so many rooms, I add separate widgets or fancy css to get images or text.  This makes it as simple as it is on a button or a basic widget.

There are some potential issues with css child inheritance on the widgets that get moved into holders. Of course, this currently is an issue if a holder uses css.  But I think -- I hope -- I'm not really sure, but maybe -- the css changes I made will work for most cases.  I tested several examples as well as existing games and did not see any issues.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2126/pr-test (or any other room on that server)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2126/pr-test (or any other room on that server)

96LawDawg:
Of course, now that I've started this, I already see a problem.  The text on recall and shuffle buttons that come with the default holders is not aligned correctly.

I actually started this PR hoping just to get images working.  It would be simpler to scale it back to just images.  But I think both of these things can be made to work.  Just need some css magic.

RaphaelAlvez:
https://github.com/ArnoldSmith86/virtualtabletop/pull/921

RaphaelAlvez:
Maybe the approach I used in that closed pr don't conflict with the labels and buttons default. If not just use those defaults and you should be good

96LawDawg:
Thanks for the reminder about that.  I had completely forgotten, and a search did not show anything because I was initially looking for image-related things.  I'm taking the text part out.  What I really did this for was to make svgReplaces available for images in holders.  I'm going back to that.

96LawDawg:
I added classes.  This still effects child widgets.  For example, drop the basic widget in the test room into one of the holders and the text gets centered.

96LawDawg:
I am back to making this PR just about images.  After looking at a list of css inherited properties, they really only apply to text things.  None of what is added in the PR to allow images is inherited by default.  Therefore, adding images should not be controversial or complicated.

I think we need a different PR for holders and text.  Even if we decide not to support a text property since css can do it with before and after somewhat easily, I think we should prevent inheritance of certain properties.  I have, several times, been frustrated by the fact that some widget I am putting into a holder during a game changes appearance because something on the holder is getting passed down to the child.  (Usually, font, font-size, color, and line-height stuff that directly involves the appearance of the text).  I think we should prevent that from happening.  I had started to do so in this PR, but that is more controversial and complicated.

So, bottom line, images are easy and should be reviewed and approved.

ArnoldSmith86:
Sorry that I've been quiet about this one. #1937 introduces a new class `ImageWidget` that is used by both `Button` and `BasicWidget` so that they don't have the same code copy-pasted everywhere anymore. After merging that one, I'd like to make `Holder` inherit from `ImageWidget` as well. Then we'll make it support text, images, icons and add better support for customizing colors.

96LawDawg:
Will be implemented later, in a different way.

96LawDawg:
I am reopening this PR.  We held off to give #1937 a chance to move forward, but it has not.  This touches very little of the code or css and does the minimum needed to allow holders to have images and allow svgReplaces to work.

96LawDawg:
Abandoned for #2634

PR #2127, opened 2024-05-22
Add hex game pieces and auto hex grid calculation

96LawDawg:
Hexes are such commonly used, we should make them easier to use.  This PR adds 2 hex widgets to the game pieces area (with the most common flat and pointed orientations).  It also adds a property to auto-calculate a hex grid when dragging.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2127/pr-test (or any other room on that server)

96LawDawg:
I added 2 more widgets to the overlay.  They use clip=path and support images.  Of course, people will probably tweak it, but now adding a basic hex or an image hex is easy and so is adding the grid.

The only problem I encountered is that the clip-path css did not show up on the overlay.  So I had to create a second widget on the overlay behind the image to make it look right.  That doesn't seem to cause any issues (yet), but if anybody knows a solution, that would be great.

These widgets will also need to be added to the test cafe that adds widgets.  I'll do that after getting indications that this is headed in the right direction.

ArnoldSmith86 (approved):
I didn't look very closely but editor changes are kinda harmless.. 😀

PR #2135, opened 2024-05-31
added a generic asset upload command

ArnoldSmith86:
There is now an `upload a different asset` command on any property that contains an asset URL.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2135/pr-test (or any other room on that server)

client/js/jsonedit.js
```diff
@@ -267,6 +267,19 @@ const jeCommands = [
       }
     }
   },
```
96LawDawg, line 269: I thought maybe this could be deleted, but then you don't get the upload button if the image is something like: i/game-pieces/2D/Checkers-2D.svg.  But leaving this in causes duplicate buttons to appear if you select the /assets/ text instead of just putting the cursor there.  This doesn't really bother me, so I'm going to approve.  But just pointing it out.

Editor

96LawDawg (approved):
Approved through test.  It works on cards now with any kind of label and not just "image".  It also still works where it is supposed to, with the duplicate as I noted in a comment.

client/js/jsonedit.js
ArnoldSmith86: Didn't think about the duplicates. Will fix.

ArnoldSmith86:
Duplicates should be gone btw.

96LawDawg:
Confirmed; duplicates gone.  Okay to merge.

PR #2139, opened 2024-06-03
PL-Update Great Greece

96LawDawg:
Allows for full color customization of all components (including the turn cards), it also puts the turn order "on rails" with a forward and backward turn button. It also implements a counter system so that you can draw the type of track tile you want without having to flip it over, but it still counts the total amount.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2139/pr-test (or any other room on that server)

ArnoldSmith86 (approved):

PR #2142, opened 2024-06-06
Add a SWAPHANDS function to simplify scripting in drafting games

### LINKED ISSUE #1973: moving from hand to hand sets owner to null
I have this code that moves a previously picked cardID from one hand to another:

```JSON
        {
          "func": "MOVE",
          "collection": [
            "${shuffled.0}"
          ],
          "to": "${playerSeat}",
          "count": 1
        }
```

After it finishes, the `owner` of the card is set to `null` instead of the player of the seat. Doing it in two steps works fine:

```JSON
        {
          "func": "MOVE",
          "collection": [
            "${shuffled.0}"
          ],
          "to": "Temp Holder",
          "count": 1
        },
        {
          "func": "MOVE",
          "from": "Temp Holder",
          "to": "${playerSeat}"
        }
```

---

Entire button:

```JSON
{
  "type": "button",
  "id": "Player 1 - Draw Card",

  "parent": "Player 1",

  "width": 60,
  "height": 60,

  "classes": "symbols",
  "css": {
    "font-size": "58px"
  },
  "text": "[card_3]",

  "clickRoutine": [
    "var seatID = ${PROPERTY parent} concat ' - Seat'",
    {
      "func": "SELECT",
      "property": "owner",
      "value": "${PROPERTY player OF $seatID}"
    },
    {
      "func": "SELECT",
      "source": "DEFAULT",
      "property": "dropShadowWidget",
      "value": null
    },
    {
      "func": "SELECT",
      "source": "DEFAULT",
      "property": "deck",
      "value": "Resources"
    },
    {
      "func": "COUNT"
    },
    {
      "func": "IF",
      "condition": "${COUNT}",
      "thenRoutine": [
        {
          "func": "INPUT",
          "fields": [
            {
              "type": "text",
              "text": "This will draw a card from ${PROPERTY player OF $seatID}'s hand. Are you allowed to do that?"
            }
          ],
          "confirmButtonText": "Draw Card",
          "confirmButtonIcon": "play_arrow",
          "cancelButtonText": "Nope",
          "cancelButtonIcon": "close"
        },
        {
          "func": "GET",
          "property": "id",
          "aggregation": "array"
        },
        "var shuffled = shuffle ${id}",
        {
          "func": "GET",
          "collection": "playerSeats",
          "variable": "playerSeat"
        },
        {
          "func": "MOVE",
          "collection": [
            "${shuffled.0}"
          ],
          "to": "${playerSeat}",
          "count": 1
        }
      ]
    }
  ],
  "ownerGlobalUpdateRoutine": [
    "var seatID = ${PROPERTY parent} concat ' - Seat'",
    {
      "func": "SELECT",
      "property": "owner",
      "value": "${PROPERTY player OF $seatID}"
    },
    {
      "func": "SELECT",
      "source": "DEFAULT",
      "property": "dropShadowWidget",
      "value": null
    },
    {
      "func": "SELECT",
      "source": "DEFAULT",
      "property": "deck",
      "value": "Resources"
    },
    {
      "func": "COUNT"
    },
    {
      "func": "IF",
      "operand1": "${COUNT}",
      "relation": ">",
      "operand2": 10,
      "thenRoutine": [
        {
          "func": "SET",
          "collection": "thisButton",
          "property": "text",
          "value": "[card_10plus]"
        }
      ],
      "elseRoutine": [
        {
          "func": "SET",
          "collection": "thisButton",
          "property": "text",
          "value": "[card_${COUNT}]"
        }
      ]
    }
  ]
}
```

### LINKED ISSUE #2141: MOVE to a seat does not correctly assign the owner if the card is already located in the hand
When executing the [MOVE function](https://github.com/ArnoldSmith86/virtualtabletop/wiki/Functions#move) to move cards directly to a seat, these cards should end up in the assigned hand for the seat, with the owner matching the seated player if the hand has `childrenPerOwner` set to true.

However, if the cards being moved are already in the hand holder (for example, in the hand of another player), then this assignment does not happen correctly, and instead the cards end up with no owner assigned at all, meaning that all players can see them and the `childrenPerOwner` setting ends up having been ignored.

[Example game definition illustrating the bug](https://gist.github.com/andypymont/0f18e86767b4a1e24274b19cd5d895eb)

andypymont:
Add a `SWAPHANDS` function for use on buttons, particularly in drafting games.
- The function creates a series of collection-to-seat moves, which are delegated to the `MOVE` function to keep consistency and for better debugging in the Edit Mode debug view.
- The owner is now set correctly when a `MOVE` sends a card from a hand to a different owner in the same hand. This closes #2141 and closes #1973.
- A new Tutorial is added demonstrating this function with multiple visible hands
- The public library Tutti Frutti game is also updated to use the new function


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2142/pr-test (or any other room on that server)

andypymont:
Here is the intended wiki documentation for `SWAPHANDS`:

# SWAPHANDS

This function swap cards in the hands of the given collection of seats. This avoids the need for an intermediate location and to manage the MOVE of the cards manually.

When SWAPHANDS involves more than two seats, it is instead a rotation with each player passing their cards to the next player in order.

Parameters:
- turn: This parameter, which should be an integer, is used as specified below under `turnCycle` to determine how many positions each seat's hand is passed around the table. The default is 1, such that each hand will be passed to the next player in the given order.
- turnCycle: This string parameter specifies the way that the "next player" that each hand is passed to will be identified:
  - forward or backward: the active players from among the seats being considered (see seats, below) are sorted into a list in index order. The position of the next player in the list is the position of the current player in the list plus (or minus, in the case of backward) the value in turn. Computations are done modulo the length of the list
  - random: the order of the seats is randomised with each seat passing to one other
- source: *all* or *collection* - specifies the collection of seats involved in the swap. The default is all; that is, all occupied seats at the table are considered when deciding who can receive swapped widgets. If the collection is named "all" but does not actually contain all occupied seats then all occupied seats will still be used.

The `hand` property on each seat in the given collection of `source` is used to identify the cards which will be passed, and also as the destination for cards which are received by that seat.

```{
  "func": "SWAPHANDS"
}```

96LawDawg:
What about just passing a set number of cards?  Can that also be added?  The INPUT choose feature works okay for that doing the selection, but the mechanics of moving still require bespoke implementation.

andypymont:
> What about just passing a set number of cards?

I feel like moving a particular number of cards is going to be rife for errors or inconsistency. (e.g. Not all players have selected their cards or moved them to the right part of the hand and the wrong things are passed.)

We could add a `collection` or a `count` parameter and have them work fairly  consistently with the same parameters in other functions; I just couldn't envisage a consistent solution for doing so which I thought would work across games.

Do you have an idea of how it would look in the spec/in the wiki documentation?

96LawDawg:
How about this idea.  I don't know how to make it work.  What if SWAPHANDS had a count parameter (defaults to all).  If it is anything other than all, then each player gets an automatic INPUT choose menu.  After each player has selected and approved the correct number of cards, the function does its magic.  The part I can't figure out is how to trigger an INPUT overlay for all players since that is currently only done for whatever player triggered the routine.  Otherwise, the javascript should be doable.  Thoughts?

andypymont:
> The part I can't figure out is how to trigger an INPUT overlay for all players since that is currently only done for whatever player triggered the routine.

This is the difficulty, because waiting on input from all players is not really very compatible with a web-based application (you don't know if somebody has navigated away from the page, has a slow connection, etc. - and how do you collate back up the inputs from every player and which client triggers the next part of the routine?)

So I think what you would want to do, is to separate the `INPUT` (choose) and `SWAPHANDS` parts of the process.

Each player could make the `INPUT` and you toggle a flag on the screen to show they have made a selection. For example if you look at _Delicious Deliveries_ in the public library, you'll see how I have done this with a simultaneous hidden bid;  card selection for passing could work similarly.

Then we feed either a collection (of the selected cards) or a count to `SWAPHANDS`, and have it limit what it moves (either to what overlaps with the collection, or to the rightmost X cards in the hand, where the `INPUT` could have moved them after selection). In the case of the count if your game is always passing 3 cards, you can also add a "pass back" button to quickly fix any errors that do happen.

Does that make sense? The design would then mostly be the existing `INPUT`, a bit of reapplication from _Delicious Deliveries_ in terms of having some sort of indicator, and then we add one more parameter to `SWAPHANDS` - probably the count is easiest but maybe the collection makes more sense for some other game implementations that I'm not thinking about right now?

96LawDawg:
After offline discussion with @andypymont, I think we should shelve the idea of passing less than the full hand for now.  The current function does enough to make it a worthwhile addition.  We can come back later and add some parameters/properties/features to make passing less than a full hand possible.

I'll do some testing on what we currently have in the PR.

andypymont:
To-do before we merge:

- [ ] Change `SWAPHANDS` tutorial room to have more distinctive hands so it's easier for a new user to see what is happening when various buttons are used - @andypymont 
- [ ] Update Tutti Frutti PL room to better account for dummy player - @96LawDawg 
- [ ] Add `SWAPHANDS` to Edit Mode context-sensitive buttons - @96LawDawg

96LawDawg:
I tested all the MOVE functions in the tutorials (MOVE, MOVEXY, as well as the seats/move).  I tested 6-7 games that I know use move to holders and also move to seats.  Everything worked as expected.

I also tested adding SWAPHANDS to a few random games and it worked as expected.

I think the code looks fine, but you all know I can't really do a code review.  The new function part I can follow.  Its the changes to the applyMove() function that I don't really understand.  Hopefully @ArnoldSmith86 can review.

ArnoldSmith86:
I'll try to look at this soon.

When in doubt, don't wait for me. 🫣

96LawDawg (approved):
Approved through test and limited code review.

ArnoldSmith86 (approved):

andypymont:
Add a `SWAPHANDS` function for use on buttons, particularly in drafting games.
- The function creates a series of collection-to-seat moves, which are delegated to the `MOVE` function to keep consistency and for better debugging in the Edit Mode debug view.
- The owner is now set correctly when a `MOVE` sends a card from a hand to a different owner in the same hand. This closes #2141 and closes #1973.
- A new Tutorial is added demonstrating this function with multiple visible hands
- The public library Tutti Frutti game is also updated to use the new function

PR #2143, opened 2024-06-07
Public Library - Update Godaigo

96LawDawg:
Multiple updates based on play testing:

- Hand moved to the far right so you don't have to drag the scroll cards as far.  Hand is also highlighted when dragging a scroll card.
- Hand reminder added.  When you get 3 cards, you get a pop up that only you can see.  Getting back down to 2 cards makes it go away.
- You cannot click to turn over a scroll card until it is out of its default holder area.
- There is some small text above the player board to show what other players have in their hand. Graphics were too small, so I went with text.
- Changed the look of the Action Points font so it is more readable.
- I added a partially transparent background to the player boards so the stuff behind them is less distracting.
- You cannot get negative Action Points by pressing the minus button.
- You are limited to 5 element stones in the holders (when you click to add them, you can't accidentally go above 5).
- The scale of pawns and element stones now matches the play area so you don't have to drop a large pawn onto a small board.
- Seat widgets are updated to the "new" standard that many games are using.  Makes choosing colors easier, is more visually appealing, and prevents accidentally getting out of or forcing someone else out of seat.
- You cannot click the +/- buttons on another player's seat.
- The Container scroll is automated.  Here's what happens: when you move the Container scroll to the active scroll area or the holder next to the Catacombs cards, an invisible widget with some text that says "Click to activate" gets attached to the card.  Then if you click on the card, it checks to make sure you want to activate the power. If confirmed, then you get the additional capacity in your hand, active scroll, and stones areas.
- The hand and the active scroll areas are slightly changed in appearance, but still work the same way.  This helps accommodate the automatic limitation on cards based on having the container scroll or not.

The changes have been playtested and approved by GMJake


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2143/pr-test (or any other room on that server)

ArnoldSmith86 (approved):

PR #2145, opened 2024-06-10
Add Toolbar buttons with links to wiki and tutorials

96LawDawg:
In edit mode, two new buttons are added to the far right for help.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2145/pr-test (or any other room on that server)

96LawDawg:
Although I do wonder if we should link everyone to the /Tutorials page.  Because that is a shared page and multiple people may try to use it at the same time.  I wonder if it is better to link people to a random page each time with https://virtualtabletop.io/#tutorials.  But that has the possibility to open up a lot of rooms if that link is clicked a lot.  Not sure if that is bad overall for the server.

ArnoldSmith86 (approved):

PR #2146, opened 2024-06-10
Fix unreadable overlay font color in dark mode

96LawDawg:
When using dark mode, text in some overlays is unreadable because it is dark on dark.  For example, this happens in the icon picker and in the save game overlay.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2146/pr-test (or any other room on that server)

ArnoldSmith86 (approved):

PR #2152, opened 2024-06-24
PL Cirque de la Cascade: Correct mistyped property on alternative board

andypymont:
Corrects a mistyped property in public library game Cirque de la Cascade, when playing on the advanced/alternate board side. This allows player pieces to snap to hex O18, consistent with other hexes on both sides of the board.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2152/pr-test (or any other room on that server)

ArnoldSmith86 (approved):

PR #2154, opened 2024-06-26
Reimplement skipTurn after it was removed mistakenly in #2080

andypymont:
This restores the `skipTurn` functionality which was mistakenly removed in #2080 , ensuring that skipping seats and specifying a source collection of seats work together.

The public library game _Egyptian Epochs_ is also updated to use `skipTurn` for players who have no remaining sun tokens, simplifying one of the `clickRoutine`s.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2154/pr-test (or any other room on that server)

96LawDawg (approved):
Approved through test. I looked at the previous Git history of `skipTurn`. I tested all the situations discussed in those issues and PRs.  Also, Property Tycoon now works properly again but does not on production.  So this should be the final answer for turns and skipTurn.  Thank you @andypymont

96LawDawg:
The only question I have, and I think this is right, is how to handle setting the turn directly on a seat using `"turnCycle":"seat"` when the seat also has `"skipTurn":true`.  There is a specific line in the code for that and it using turnCycle override skipTurn.  I tested that and I believe that is the right answer, but I suppose that could be up for debate.  But if we did not do it this way, then in someone's room it could be possible to have a TURN that tried to set the turn using turnCycle to a seat with skipTurn and then it would be no one's turn.  So I like this solution.

andypymont:
> The only question I have, and I think this is right, is how to handle setting the turn directly on a seat using `"turnCycle":"seat"` when the seat also has `"skipTurn":true`. There is a specific line in the code for that and it using turnCycle override skipTurn. I tested that and I believe that is the right answer, but I suppose that could be up for debate.

Yes. I thought it made most sense to override `skipTurn` when a specific seat was specified, as I assume that is for cases where you very much want that seat to become the on-turn seat. If that's not the case, we need to decide on what the appropriate behaviour is for this case. Most of the alternatives I could think of (e.g. select the next seat in `index` order) seemed like they would be even more confusing.

PR #2159, opened 2024-06-30
simple search and replace in the editor

ArnoldSmith86:
This adds a powerful but very simple search and replace feature to the toolbox of the editor.

This should be put somewhere else and be more polished but I do not feel like working on this for hours right now.

### DIFF
--- a/client/js/editor/sidebar/toolbox.js
+++ b/client/js/editor/sidebar/toolbox.js
@@ -44,7 +44,7 @@ class ToolboxModule extends SidebarModule {
           return;
         }
       } else {
-        const regexPattern = new RegExp(`\\b${globalSearchText}\\b`, 'g');
+        const regexPattern = new RegExp(`\\b${regexEscape(globalSearchText)}\\b`, 'g');
         newState = newState.replace(regexPattern, globalReplaceText);
       }
       try {
--- a/client/js/editor/sidebar/toolbox.js
+++ b/client/js/editor/sidebar/toolbox.js
@@ -100,8 +100,8 @@ class ToolboxModule extends SidebarModule {
       <input id=globalSearchText placeholder="Search text"><br>
       <input id=globalReplaceText placeholder="Replace text"><br><br>
       <div>
-        <input type=checkbox id=globalReplaceCase><label for=globalReplaceCase> Match case</label><br>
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2159/pr-test (or any other room on that server)

96LawDawg:
It's also very dangerous.  It replaces partial words.  So I tried to replace "y" with "yAxis".  It did that of course.  But it also replaced "layer" with "layAxiser."  And undo did not work on the mess I made with that.

96LawDawg:
> It's also very dangerous. It replaces partial words. So I tried to replace "y" with "yAxis". It did that of course. But it also replaced "layer" with "layAxiser." And undo did not work on the mess I made with that.

I changed it so if you are not using a regex, then it only replaces entire words.  We could add a switch to do either, but it seems like that would be an unusual case and could lead to unexpected results.  With this change, I think this could be merged.  I don't have a better place to put it.  I don't think it goes on the top toolbar.  And I don't think it goes on the button area.  So this is as good as place as any.

ArnoldSmith86:
I should take a closer look at Undo.

96LawDawg (approved):
Approved through test.

If you want to replace an image, you can't search or replace something like /assets/990236263_1684.  The slashes probably have to be put into a regex.  But you can just use 990236263_1684.  I don't think it is worth the time to try and fix that here.  This is already a big improvement and will probably mean I don't have to use the VS Code extension anymore.

Thanks!

ArnoldSmith86:
> If you want to replace an image, you can't search or replace something like /assets/990236263_1684.

I'm pretty sure, it's because it starts with `/` and that apparently can't match your `\b`.

We could make it only search for entire words if the string only contains letters.

Or we could just add a checkbox "search for entire words" or whatever. This sounds like a more familiar and less confusing option.

ArnoldSmith86:
Basically everything changed. Can you test again? :laughing:

96LawDawg:
Good.  Works.  But I think case and whole word should be checked by default.  Too easy to mess up otherwise.  But I haven't broken it so bad in testing that undo doesn't work anymore.

96LawDawg:
Correction.  I did just break the room with y changing to yAxis again.  But that is a different feature/bug not related to search and replace.

PR #2160, opened 2024-06-30
PL Egyptian Epochs: Fix bug with skipTurn introduction

andypymont:
In simplifying Egyptian Epochs routines to use `skipTurn` to skip players without any suns, an error was introduced meaning that `skipTurn` was also set on players that won an auction. This corrects this by ensuring that current-bid holders inherit from the future sun holder instead of the active sun holder.

### DIFF
--- a/library/games/Egyptian Epochs/0.json
+++ b/library/games/Egyptian Epochs/0.json
@@ -431,7 +431,7 @@
     "type": "holder",
     "id": "holder-suns-bid-p1",
     "y": 75,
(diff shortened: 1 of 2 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2160/pr-test (or any other room on that server)

ArnoldSmith86 (approved):

PR #2162, opened 2024-07-01
fixed widget buffer with invalid parent

ArnoldSmith86:
Putting widgets from the widget buffer into the game now checks if their parents exist in the buffer or in the game. If a parent does not exist, it will be set to `null` instead.

Additionally this gives an error message if a referenced deck does not exist.

### DIFF
--- a/client/js/editor/sidebar/toolbox.js
+++ b/client/js/editor/sidebar/toolbox.js
@@ -20,8 +20,14 @@ class ToolboxModule extends SidebarModule {
     const widgetBuffer = JSON.parse(localStorage.getItem('widgetBuffer') || '[]');
     batchStart();
     setDeltaCause(`${getPlayerDetails().playerName} loaded widgets from the widget buffer in editor`);
-    for(const state of widgetBuffer)
-      await addWidgetLocal(state);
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2162/pr-test (or any other room on that server)

96LawDawg:
Does not work.  Get a pop up that says: "Widget score+10 references a deck that is not in the buffer and is not already in the room. It will not be loaded."  Happens whether or not the widget is a deck.

ArnoldSmith86:
Oops.

ArnoldSmith86:
Better?

96LawDawg:
> Better?

Yes, that is fixed now and works as it should for cards (don't add them if no deck).

One final point.  If there is no parent, should we instead of using the existing x and y, just set it to 0,0 to make sure it is visible.  I put a widget into a room without a parent that had a negative y value and it was off screen.  I was expecting that, but a casual user might not be and it looks like the process did not work.

96LawDawg (approved):
Approved through test.

PR #2167, opened 2024-07-01
added property `display`

### LINKED ISSUE #2166: Feature request: "display" property for widgets and objects
It might be useful to have a property "display" that can be set to false to hide a an object. Two examples where this might be used:

**1) Objects on faceTemplates for card decks**
In Egyption Epochs, the scarab for permanent tiles must be shown on some cards and hidden on others. Currently, this is implemented like this:

```
  "cardDefaults": {
    "scarabcolor": "#e30022"
  },

  …

    "tile-nile": {
      "color": "#6495ed",
      "image": "/i/game-icons.net/delapouite/palm-tree.svg"
    },
    "tile-flood": {
      "color": "#6495ed",
      "image": "/i/game-icons.net/delapouite/oasis.svg",
      "scarabcolor": "transparent"
    },

  …

  "faceTemplates": [
    {
      "objects": [
        {
          "type": "image",
          …
          "value": "/i/game-icons.net/lorc/gold-scarab.svg",
          "svgReplaces": {
            "#000": "scarabcolor"
          }
        }
      ]
    }
```
This works, however, it is not semantic and presents a barrier to new mod developers. In addition, it is important to to set up this workaround the right way around. In the same cardDeck, the crosses for disaster tiles also need to be shown on some cards and not on others. However, for those, the cardDefault is defined as "transparent", with the cards that show the cross overriding the value to red. As a result, if someone wants to alter the color of the cross, they have to do so for each cardType that features it.

With the display Property, this could be rewritten like this:
```
    "tile-nile": {
      "color": "#6495ed",
      "image": "/i/game-icons.net/delapouite/palm-tree.svg",
      "showScarab": true
    },
    "tile-flood": {
      "color": "#6495ed",
      "image": "/i/game-icons.net/delapouite/oasis.svg"
    },

  …

  "faceTemplates": [
    {
      "objects": [
        {
          "type": "image",
          …
          "value": "/i/game-icons.net/lorc/gold-scarab.svg",
          "dynamicProperties": {
            "display":"showScarab"
          }
        }
      ]
    }
```

**2) Widgets that need to be hidden with the option to later reappear in the same spot**
In this fork of the Great Greece mod parts of the map need to be hidden: https://virtualtabletop.io/game/lr20tzaf/great-greece-smaller-maps-rivers They cannot be deleted, as there is the possibility that they will need to be re-shown when setting up the game for a second match.
This is currently achieved by setting their layer to -50, which is hidden behind a large image
```
            {
              "func": "SELECT",
              "property": "remove_left",
              "relation": "==",
              "value": true
            },
            {
              "func": "SET",
              "property": "layer",
              "relation": "=",
              "value": -50
            },
```
To re-show them, the layer must be set to the original value of 1
```
        {
          "func": "SELECT",
          "property": "map_hex",
          "relation": "==",
          "value": true
        },
        {
          "func": "SET",
          "property": "layer",
          "relation": "=",
          "value": 1
        },
```
This is, again, not semantic. In addition, if the "correct" layer to show them at later changes from 1 to a different value, the modder would need to remember to also change the value in this routine.

With a display property, this could be rewritten more semantically as
```
            {
              "func": "SELECT",
              "property": "remove_left",
              "relation": "==",
              "value": true
            },
            {
              "func": "SET",
              "property": "display",
              "relation": "=",
              "value": false
            },
``` 
```
        {
          "func": "SELECT",
          "property": "map_hex",
          "relation": "==",
          "value": true
        },
        {
          "func": "SET",
          "property": "display",
          "relation": "=",
          "value": true
        },
```
This makes no assumption about the overall layer composition of the mod.

ArnoldSmith86:
This fixes #2166 after I add the same to card objects. I did not test it yet though.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -35,6 +35,7 @@ export class Widget extends StateManaged {
       StateManaged.inheritFromMapping[id] = [];
 
     this.addDefaults({
+      display: true,
       x: 0,
       y: 0,
       z: 0,
@@ -547,6 +548,8 @@ export class Widget extends StateManaged {
     css += '; height:' + this.get('height') + 'px';
     css += '; z-index:' + this.calculateZ();
     css += '; transform:' + this.cssTransform();
+    if(!this.get('display'))
+      css += '; display: none';
 
     return css;
   }
@@ -588,7 +591,7 @@ export class Widget extends StateManaged {
   }
 
   cssProperties() {
-    return [ 'borderRadius', 'css', 'height', 'inheritChildZ', 'layer', 'width' ].concat(this.propertiesUsedInProperty['css']||[]);
+    return [ 'borderRadius', 'css', 'display', 'height', 'inheritChildZ', 'layer', 'width' ].concat(this.propertiesUsedInProperty['css']||[]);
   }
 
   cssReplaceProperties(css, usedProperties) {


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2167/pr-test (or any other room on that server)

Tobl4:
No in-depth-testing from my side (don't think I'm qualified to think through all the relevant edge cases yet), but it worked well for recreating the "hiding map hexes in Greater Greece" use case.

Tobl4:
Cards in Egyptian Epochs were also easy to replicate with display for cross and scarab

96LawDawg:
This is a nice feature.  I tested on every type of widget.  Works on cards in the object and works as a dynamicProperty.  Works correctly in cardDefaults and on the card widget directly.  It also works as expected on holders with children and decks. In other words, even if you set the deck or the holder with a deck in it to display:false, the cards not in the holder are still displayed. And it works on piles (the pile and all members of the pile are hidden).

There is an issue with using css:none to hide the widgets and that is they are invisible in the room even in edit mode.  The only way to find them is to use the tree (or make a temp button to SELECT everything and SET display:true. Is it possible and advisable to disable the css:none if the body class includes "edit?"  Just putting that up for discussion, but I could approve this any time based on testing.

96LawDawg:
Offline discussion with @Tobl4.  Additional testing on holders with display:false shows they do not accept widgets as children (as intended). Also, for edit mode would be nice to have a toggle to hide or show the widgets with display:false.  I could not really figure out how to do that, or even to get the icon to switch from visibility to visbility_off.  But there is an icon in there now that you can click that does nothing.

ArnoldSmith86:
Any more comments?

96LawDawg:
> Any more comments?

Yes.  AMAZING!

96LawDawg (approved):
Approved through test.

PR #2173, opened 2024-07-04
Fix skipTurn bug when no seats are available to have the turn set

### LINKED ISSUE #2172: TURN function freezes game interaction if all occupied seats are "skipTurn" : true
https://virtualtabletop.io/bedv

Observed behavior: Calling the TURN function if at least one seat is occupied but all occupied seats have "skipTurn" set to true freezes the game for the client calling the function. Initially, the poker chip can be dragged around, but as soon as the TURN button is pressed, that client can no longer interact with the game until they reload the page.

Expected behavior: Keep the current turn unchanged and return to the calling routine, same as when no seat is occupied.

96LawDawg:
Fixes #2172 by implementing a check for an undefined "unskipped" variable.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -1942,32 +1942,36 @@ export class Widget extends StateManaged {
           let unskipped = c.filter(w=>!w.get('skipTurn'));
           let target = unskipped[0];
 
-          // identify the correct target seat
-          if (a.turnCycle == 'position') {
-            if (a.turn == 'last') {
-              target = unskipped[unskipped.length - 1];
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2173/pr-test (or any other room on that server)

96LawDawg:
It looks like a lot was changed, but that was mostly indentation. I only add the the if, else, and problems.push near the top of the changed area.

ArnoldSmith86 (approved):

PR #2174, opened 2024-07-04
added support for tiled images in deck generator

ArnoldSmith86:
It is now easy to generate cards from images containing multiple card front images.

image

image


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2174/pr-test (or any other room on that server)

96LawDawg:
> It is now easy to generate cards from images containing multiple card front images.

"Easy" is somewhat relative.  It only works if there is no outside margin around the cards.  I downloaded a few random card sprite sheets from the internet and could only get 1 out of 3 to work right out of the box.  Had to go into graphics software to remove the margins.  Not that big a deal, but maybe add that to the descriptive text: "Generate a deck by uploading images that cover the whole card. Tiled images must have no outside margins."

ArnoldSmith86:
I could add a slider for a margin but at some point it would be building an image editor.. Do you think margin would be worth it?

96LawDawg:
> I could add a slider for a margin but at some point it would be building an image editor.. Do you think margin would be worth it?

No.  The more I think about it, the more complicated it is.  The problem is not just the margin, but also the gutter (gap) between the cards.  If there is one, then the margin needs to be 1/2 of the gap I think.  It is too much to handle in the editor.  Maybe change the hint text to: "Generate a deck by uploading images that cover the whole card. You may need to adjust the margins for tiled images in an image editor."

96LawDawg (approved):
Approved through test.

ArnoldSmith86:
"You may need to _remove_ any gaps or margins with an image editor." might be better?

PR #2180, opened 2024-07-07
added the current delta to the debug output of each operation

ArnoldSmith86:
A complex routine was causing a `parent` to change and I did not know where that happened. With this change, it was a lot easier to find the culprit.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2180/pr-test (or any other room on that server)

96LawDawg (approved):
Works.  Will definitely come in useful.  Approved through test.

PR #2183, opened 2024-07-21
added caching to `get`

ArnoldSmith86:
Every time `get` is called, it evaluates the current value of the given property. For cards or when `inheritFrom` is involved, this can take a bit.

This PR caches `StateManaged.get` results. On the original test case (switching the Card Games game to Color Whist), this reduced the reported time from **2.77s to 1.19s** on the original author's laptop.

## Cache invalidation

Cached values use a global generation number. Any default or state mutation advances the generation, and each widget lazily clears its local cache on its next read. This keeps repeated reads between mutations fast while safely covering every direct and indirect dependency, including:

- `inheritFrom` chains and mid-routine source updates
- card values derived from deck defaults and face templates
- face-derived basic-widget and dice values
- pile values derived from children
- computed values that depend on parents or other widgets
- widget additions and removals

Computed read-only properties such as `_absoluteX`, `_ancestor`, and scoreboard `_totals` remain uncached and are always recomputed.

## Testing

- Added Jest regressions for direct and transitive inheritance, source removal, self-updates, and an unregistered cross-widget dependency.
- Full Jest suite passes after merging the latest `main` (20 tests).
- Full TestCafe suite passes in both Chrome and Firefox, including the face, deck, computed-property, and routine state-hash coverage.
- Server startup and CodeQL checks pass.

- [x] Verify `${PROPERTY x}` / `${PROPERTY x OF widget}` reads refresh after their source changes, including within a routine.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2183/pr-test (or any other room on that server)

bjalder26:
Here is one thing (I think).

Faces doesn't appear to work the same as on the production server.  It doesn't seem to actually update anything as you change the faces.

https://test.virtualtabletop.io/PR-2183/pr-test
https://virtualtabletop.io/PR-2183

VTT-AI-Agent (commented):
# Review of PR #2183 — added caching to `get`

## TL;DR
The speedup is real, but the invalidation is per-key on a widget's *own* state only, so every cross-property and cross-widget dependency goes stale: `_absoluteX` after a move, face properties after a flip, card properties after a deck edit, dice values with `ignoreFaceProperties`, scoreboard `_totals`. Rather than chasing each dependency edge, consider a global epoch counter that invalidates all caches on any state change — it keeps almost all of the win (the hot loops re-read the same properties many times between mutations) and is correct by construction.

## Findings

You asked for cases that break — here's what I could confirm by reading the invalidation paths. The cache is currently cleared in exactly four places: per-key in `applyDelta` (statemanaged.js:26) and `set` (statemanaged.js:119), and full clears on `inheritFrom` changes (widget.js:251, widget.js:261). Everything below falls through those.

### 1. Computed `_absolute*` / `_ancestor` properties are never invalidated (widget.js:2015)
`_absoluteX`, `_absoluteRotation`, `_ancestor`, etc. depend on the widget's own `x`/`y`/`rotation`/`scale`/`parent` **and the whole ancestor chain**, but `applyDelta` only deletes the delta's own keys. Move a widget (delta `{x, y}`) and its cached `_absoluteX` stays at the old value forever; move or rotate a *parent* and every descendant's cached absolute coordinates are stale with no invalidation at all. Any routine reading `_absoluteX`/`_centerAbsoluteY` after the first move gets wrong coordinates. Suggest simply not caching `readOnlyProperties` (they're the minority of calls), or see the epoch idea below.

### 2. Face-derived properties are stale after `activeFace` changes (basicwidget.js:117, card.js:220, dice.js:281)
`set('activeFace', n)` deletes only `getCache['activeFace']`. Every property previously resolved through the old face (e.g. `image`, `color`) stays cached. Worse, for cards the flip path itself consumes the stale values: `Card.applyDeltaToDOM` builds `deltaForFaceChange[key] = this.get(key)` (card.js:63–68), which now returns the *old* face's cached values, so dynamic-property faces render wrong after a flip. Same class of problem for `faces` edits and for `cardType` changes (Card.getDefaultValue depends on both). At minimum, `activeFace`/`faces`/`cardType`/`deck` changes need a full `getCache = {}` on the widget.

### 3. Deck edits never invalidate its cards' caches (deck.js:27–47)
Card properties resolve through `Card.getDefaultValue` → `deck.cardPropertyGet` (card.js:220–226). When a deck's `cardTypes`/`cardDefaults`/`faceTemplates` change, `Deck.applyDeltaToDOM` builds `deltaForCard` from `card.get(key)` — which now returns the stale cached values — and calls `card.applyDeltaToDOM`, which never touches `getCache`. So the cards' DOM gets updated from stale data *and* their caches remain stale afterwards. The deck→card relationship isn't in `inheritFromMapping`, so the widget.js:261 clear doesn't cover it.

### 4. Dice cache ignores the `options` argument (dice.js:280–281)
The cache check runs before the `ignoreFaceProperties` check. Once `get('value')` caches the face-mapped value, an internal `get('value', {ignoreFaceProperties: true})` call returns the face value instead of the raw property — and vice versa, whichever variant runs first poisons the cache for the other. There are eight `ignoreFaceProperties` call sites in dice.js (e.g. dice.js:46, 243, 304) that can now receive the wrong variant. The dice `get` should bypass the cache (or key it separately) whenever `options` is set.

### 5. Scoreboard `_totals` is stale after any score change (scoreboard.js:153)
`_totals` aggregates *other* widgets' (seats') score properties, but nothing invalidates the scoreboard's cache when a seat's score changes — the cached totals are effectively computed once and frozen. It also now returns the same array instance to every caller, so a caller mutating the result corrupts the cache. `_totals` probably shouldn't be cached at all.

### 6. Inherited values are stale within a batched routine
The `inheriting.getCache = {}` clear (widget.js:261) runs in `applyDeltaToDOM`, which for a locally-originated `set()` only happens when the batch flushes (`sendDelta` → `receiveDelta`, serverstate.js:417). Inside a routine (which runs batched), setting a property on the source widget and then reading it from an inheriting widget in a later step returns the stale cached value — a behavior change from today, where `get` always recomputed.

## Suggested direction

Enumerating every dependency edge (ancestors → descendants, deck → cards, seats → scoreboard, faces, value maps, `${PROPERTY x}` references…) is a losing game — each widget type adds new edges. A much simpler scheme that captures most of the win:

- keep a global `StateManaged.epoch` counter, bumped in `applyDelta` and `set` (any widget, any key);
- stamp each cache entry with the epoch (or clear a per-widget cache whose stored epoch is outdated on first access);
- a cache entry is valid only if its epoch matches.

The hot paths this PR targets (inheritance chains, card property resolution, repeated `get` calls during a single routine step or render pass) all re-read the same properties many times *between* mutations, so an "any write invalidates everything" rule preserves nearly all of the 2.77s→1.19s improvement while being correct for every dependency, including ones added later. `readOnlyProperties` that depend on ancestors would also be covered, since a parent's `applyDelta` bumps the epoch.

## Testing

No tests accompany the change. The TestCafe suite's state-hash comparison (see docs/testcafe.md) is well-suited to catching several of these (face flips, deck edits, moves followed by `_absolute*` reads) — worth running the full suite locally before taking this out of draft, plus a targeted test that moves a parent and reads a child's `_absoluteX`.

---
No `beta` label added: this PR changes the engine (client/js/widgets/, client/js/statemanaged.js), which is excluded from beta auto-deployment by design — and as a draft with known unresolved staleness cases it shouldn't be player-tested yet anyway.

VTT-AI-Agent (commented):
## TL;DR

The caching idea is sound and the speedup is real, but the invalidation currently only covers a widget's *own* delta keys — every value that `get` derives from something else (the active face, the deck, the parent chain, other widgets via `inheritFrom` chains or scoreboard seats) goes stale. The most visible breakage: flipping a card or rolling a die does **not** invalidate the cached face-derived properties, so both routines and the DOM keep seeing the old face.

## Findings (most important first)

1. **Face changes don't invalidate face-derived properties.** `applyDelta` (client/js/statemanaged.js:26) only does `delete this.getCache[i]` for the keys present in the delta. A card flip delivers `{activeFace: 1}`, which clears only `getCache['activeFace']` — but `BasicWidget.get` (client/js/widgets/basicwidget.js:117) has cached `image`, `text`, `value`, etc. resolved through the *old* face, and `Card.getDefaultValue` (card.js:279) resolves through `deck.cardPropertyGet(cardType, activeFace, ...)`, so every deck/face-derived property on a card is stale after a flip. This isn't just routine-visible: `Card.applyDeltaToDOM` builds `deltaForFaceChange[key] = this.get(key)` on `activeFace` deltas (card.js:63-69), so the DOM face-property update reads the stale cache too. Same for `Dice.get` (dice.js:281) when `activeFace`/`value` changes. Any face-related delta (`activeFace`, `faces`, `deck`, `cardType`) probably needs to clear the whole cache (or at least all non-own-state keys).

2. **`Dice.get` caches without regard to `options.ignoreFaceProperties`.** The cache check at dice.js:282 runs *before* the options test, and the cache key is just `property`. `Dice.css()`/`faceCssVariables` deliberately call `this.get(prop, {ignoreFaceProperties: true})` (dice.js:259, 281) — whichever variant is called first poisons the cache for the other, so e.g. the die's base `color` CSS variable can end up holding the current face's color (or vice versa). Either skip caching when `options` is passed, or include the flag in the cache key.

3. **Cross-widget computed properties are never invalidated.** `_absoluteX/Y/Rotation/Scale`, `_centerAbsolute*`, `_ancestor` (widget.js:2015+) depend on `x`/`y`/`rotation`/`scale`/`parent` of the widget *and its whole parent chain*. A delta of `{x: 100}` deletes `getCache['x']` but leaves the widget's own cached `_absoluteX` — and leaves every descendant's cached `_absolute*` untouched. After the first read, these effectively freeze. Same class of problem for Scoreboard `_totals` (scoreboard.js:153): it aggregates *seat widgets'* scores, so it's stale as soon as any seat's score changes, and no delta on the scoreboard itself will ever contain the key `_totals` to evict it. As written, a scoreboard that reads `_totals` twice shows the first result forever.

4. **`set()` doesn't invalidate dependents synchronously.** `StateManaged.set` (statemanaged.js:119) clears only the widget's own key. Widgets inheriting from it get their caches cleared only in `Widget.applyDelta` (widget.js:268-271), i.e. when the server delta comes back. But routines run synchronously on the client against already-mutated local state: a routine that sets a property on an `inheritFrom` source and then reads it via the inheriting widget in the same routine gets the stale cached value — previously it re-derived from the source's fresh state. (This may be exactly the `${PROPERTY x}`-adjacent class of breakage you anticipated.)

5. **Chained `inheritFrom` only invalidates one level.** `StateManaged.inheritFromMapping` registers only direct inheritors (widget.js:317-319), but `getDefaultValue` resolves chains transitively at read time. If A inherits from B and B from C, a delta on C clears B's cache but not A's — A keeps its stale value derived through B.

6. **Deck-level edits don't reach cards.** Cards resolve properties through `deck.cardPropertyGet` but have no `inheritFrom` mapping to the deck, so editing a deck's `cardTypes`/`cardDefaults`/`faceTemplates.properties` in edit mode leaves all its cards' caches stale until each card happens to receive its own delta.

7. **Cached mutable references.** Scoreboard `_totals` previously returned a fresh array per call; now all callers share (and could mutate) the cached array. Same for any cached face/state objects. Worth keeping in mind if a routine ever writes into a value it read.

### Suggestion on strategy

Given how many distinct dependency edges there are (own state → fine; face; deck; parent chain; inheritFrom chain; other widgets), per-key eviction is going to be a long game of whack-a-mole. Two safer starting points that keep most of the win:

- Clear the **entire** `getCache` of a widget on any delta/set to it, and cascade a full clear to registered dependents (inheritors, children for `_absolute*`, cards for decks). Repeated `get`s within one routine/render pass — likely where the 2.77s→1.19s win comes from — still hit the cache.
- Or scope the cache to a "generation" counter bumped on every applied delta anywhere, so correctness is trivial and hot loops within one generation still benefit.

Also worth noting: dice.js:304 `Object.hasOwnProperty(faceProps, property)` is a pre-existing bug (it checks whether `Object` itself has a property named `"[object Object]"`, always false) — `Object.hasOwn(faceProps, property)` is what's meant. Not introduced here, but you're touching the line's branch, and fixing it will change which values get cached.

No test changes in the PR — the TestCafe state-hash suite should catch some of the flip/inherit staleness, so running it locally before un-drafting would be a good signal.

I did not add the `beta` label: this PR changes `Widget.get`/`StateManaged` caching semantics for all existing games — a core engine change that is excluded from beta deployment by design, and in its current draft state it would visibly break card flips.

VTT-AI-Agent:
I took over this PR and reworked the invalidation to address the review. Rather than evict per-key (whack-a-mole as noted), the cache now lives solely in `StateManaged.get`, and **any** delta/set clears that widget's whole cache and cascades to dependents. Mapping to the findings:

1. **Face changes / face-derived props** — fixed. The redundant per-subclass caches in `basicwidget`/`dice`/`scoreboard` were removed; a face delta (`activeFace`, `faces`, `deck`, `cardType`) now clears the entire `getCache`, so both routines and the DOM read the fresh face.
2. **`Dice.get` + `ignoreFaceProperties`** — fixed by dropping dice's own cache entirely, so the option can no longer poison the key. (The pre-existing `Object.hasOwnProperty` bug you flagged is left untouched since I no longer modify that branch — happy to fix it in a separate PR.)
3. **Cross-widget computed props** (`_absolute*`, `_ancestor`, scoreboard `_totals`) — fixed by **not** caching them; they stay always-computed since they depend on the live parent chain / other widgets. `_totals` also returns a fresh array again (addresses finding 7).
4. **`set()` doesn't invalidate dependents synchronously** — fixed: `set()` now calls `invalidateGetCache()`, which clears the widget and its dependents immediately, so mid-routine reads through an inheriting widget are correct.
5. **Chained `inheritFrom`** — fixed: `invalidateGetCache` recurses through `inheritFromMapping` with a visited set, so A ← B ← C all clear.
6. **Deck-level edits → cards** — fixed: `Deck.invalidateGetCache` also clears its cards.

Added `tests/client/widget-getcache.test.js` (jest) for the inheritFrom cases, and the full TestCafe suite passes (face flips, deck edits, `_absolute*` after moves). Also verified `${PROPERTY x}` re-reads correctly after the source changes, including mid-routine.

VTT-AI-Agent (commented):
## TL;DR

The generation scheme closes the earlier cross-widget invalidation holes, but two stale-cache paths remain. Most importantly, code that directly updates a readonly clone's `state.scale` now renders with the old cached scale, so scaled `INPUT` choices no longer honor their configured size.

## Findings

### 1. High — direct `state` writes bypass the new invalidation contract

`StateManaged.get()` now returns a cached value until some code calls `invalidateGetCache()` (`client/js/statemanaged.js:61-68`), but several existing callers intentionally mutate temporary widgets' state directly. In the player-facing choose overlay, `renderReadonlyCopy()` has already run `applyInitialDelta()` and `cssTransform()`, which caches the clone's initial `scale`; then `client/js/domhelpers.js:373-374` assigns `widgetClone.state.scale = scale * (field.scale || 1)` and immediately calls `widgetClone.css()`. That second call still gets the cached initial scale, so any `INPUT` choose field using `scale` (and any non-1 room scale) renders the choice at the wrong size. The card-type editor has the same pattern at `client/js/editor/sidebar/properties.js:1443-1449`: it assigns changed dynamic card properties into a preview card and recreates its faces, but properties read during the initial render remain cached, so the live preview can keep showing the previous values. Please route these writes through a cache-aware helper or explicitly clear the affected temporary object's cache, and cover at least a scaled choose clone plus a dynamic card-type preview.

### 2. Medium — resolve removed-property defaults only after the delta's new state is fully visible

In `client/js/statemanaged.js:19-31`, non-null keys update `state` without clearing their cached entries, while a null key immediately calls `this.get(i)` to build `deltaForDOM`; the generation advances only after the whole loop. On a remote client, prime a card's old `cardType`, then receive `{ cardType: "newType", color: null }`: resolving `color` calls `Card.getDefaultValue()`, which sees the cached old `cardType` and places the old type's color in `deltaForDOM`. Although the cache is cleared afterward and future `get('color')` calls are correct, `Card.applyDeltaToDOM()` ultimately applies that stale value, leaving the remote player's rendered card wrong until another relevant delta. A two-phase `applyDelta` (first apply all state changes, then invalidate, then resolve null/default values) would make every default see the complete new state; a regression test should prime the dependency before applying a multi-key delta.

I did not add the `beta` label because this PR changes default widget property-read behavior in `client/js/widgets/widget.js`/`client/js/statemanaged.js`, so it is an actual engine change and is engine-ineligible for beta deployment.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-tested change. The caching is correct because invalidation is **fully global** — every `set()` / `applyDelta` / `addDefaults` / widget removal bumps a shared generation counter that lazily wipes *all* widgets' caches, so there's no way for a cached value to outlive a state change it depends on (including cross-widget/`inheritFrom`/deck/pile dependencies). I found no correctness bugs. This is a genuine engine change, so it is **not** beta-eligible and I did not add the label.

## What I checked

- **Coverage of mutation paths.** The only ways a `get()` result can change are (a) a state/default mutation somewhere, or (b) the read-only computed props (`_absoluteX`, `_ancestor`, `_totals`, …). Every mutation entry point invalidates: `set` (statemanaged.js:148), `applyDelta` (both the null-delete per-key `delete this.getCache[i]` and the trailing global bump), `addDefaults`, and `Widget.remove` (widget.js:369). The read-only props are excluded from caching (`Widget.get` at widget.js:2274 bypasses the cache; `scoreboard.get` short-circuits `_totals` before ever reaching `super.get`, so it's never cached even though the module-level `readOnlyProperties` set doesn't list it). Good.
- **Edit-mode direct mutation.** `editmode.js:1502` writes `widgets.get(id).state[key] = …` directly without invalidating, which initially looked like a gap — but it writes the *new* value into state (so the cache is never left stale-relative-to-state), and the enclosing `batchStart()/batchEnd()` runs `applyDelta` on flush, which invalidates. No staleness.
- **Face/dice/card overrides.** `basicwidget.get` / `dice.get` return face-derived values *without* caching (they return before `super.get`), and only the raw fallback is cached — no key collision between `get('value')` and `get('value', {ignoreFaceProperties:true})`, since only the raw path ever reaches `super.get`. `card.getDefaultValue` (deck/face-template derived) is cached but correctly invalidated when the deck or the card's `activeFace` changes.
- **`cached !== undefined` sentinel.** `StateManaged.get` never stores `undefined` (it stores `+value`, `value`, or `defaultValue ?? null`), so the sentinel cleanly distinguishes "miss" from a legitimately-cached `null`/`0`/`""`/`NaN`. Good.
- **Tests.** `tests/client/widget-getcache.test.js` covers direct + transitive inheritance, source removal, self-update, and an unregistered cross-widget dependency. Full jest suite passes locally (20 tests, 4 suites). No new widget/routine property is introduced, so no `validator/` or `jsonedit.js` upkeep is required.

## Minor observations (non-blocking)

- **Invalidation is entirely global**, not per-widget or per-dependency. That's the safe choice and clearly what makes it correct, but it means the cache yields little benefit in write-heavy routines (typical gameplay interleaves many `set()`/`get()` — each `set()` wipes every cache). The measured win applies to read-bursts between writes (e.g. the game-switch case). Worth a comment in the code so a future reader doesn't mistake this for a dependency-tracking cache and try to "optimize" the invalidation.
- **Redundant local clears.** Given the global bump on every mutation, the explicit `this.getCache = {}` in `Widget.applyInheritedDeltaToDOM` (widget.js:326) and the per-key `delete this.getCache[i]` inside `applyDelta`'s null branch are belt-and-suspenders — harmless, but not strictly necessary since the trailing/originating `invalidateGetCache()` already covers them. Fine to keep as defensive code.
- **Returned references are shared across reads** within a generation. This matches the pre-existing contract (`get()` already returned the live `state[property]` / `defaults[key]` reference, not a copy), so callers that treat `get()` results as read-only are unaffected — just noting it in case any callsite mutates a returned array/object.

## Beta label

Not added: this modifies `StateManaged.get` — the property-resolution path used by every widget and the client state-sync (`applyDelta`) machinery — so it genuinely changes how existing games run and is excluded from beta auto-deployment by design.

VTT-AI-Agent:
Addressed the two stale-cache paths from the review in eb8d6c2a:

**Finding 1 (High) — direct `state` writes bypass invalidation.** Confirmed: the choose overlay's `renderReadonlyCopy` runs `applyInitialDelta`, which renders the clone and caches `scale = 1` via `cssTransform`; the subsequent `widgetClone.state.scale = …` (domhelpers.js) then read back through `css()` returned the cached `1`, so scaled `INPUT` choices rendered at the wrong size. Same pattern in the card-type preview (`renderCardTypes_updateCardVisualization` in editor `properties.js`). Both now call `invalidateGetCache()` immediately after the direct write. (The other direct writes — `renderCardLayers`/`renderCardTypes` and edit mode's batched `state[key] = …` — write before any read or are followed by an `applyDelta` flush, so they're unaffected.)

**Finding 2 (Medium) — defaults resolved before the delta's new state is visible.** Confirmed: `applyDelta` resolved a removed (null) key's default mid-loop, so a delta like `{ cardType: "new", color: null }` resolved `color` against the still-cached old `cardType` and pushed the old type's color into the rendered delta. Reworked `applyDelta` into two phases — apply all state changes → `invalidateGetCache()` → resolve DOM values — so every default sees the complete new state.

Added jest regression tests for both (the multi-key one captures the delta handed to `applyDeltaToDOM`, since the staleness is in the rendered value, not a later `get()`), plus a guard for the direct-write→`invalidateGetCache()` pattern. Full jest suite passes (22 tests).

PR #2185, opened 2024-08-10
added circle alignment tool

ArnoldSmith86:
I needed to align a few labels around a circle. I used ChatGPT and Copilot to pretty quickly add a tool for that to the toolbox.

### DIFF
--- a/client/js/editor/sidebar/toolbox.js
+++ b/client/js/editor/sidebar/toolbox.js
@@ -3,6 +3,51 @@ class ToolboxModule extends SidebarModule {
     super('home_repair_service', 'Toolbox', 'Miscellaneous tools that don\'t have a home yet.');
   }
 
+  async button_circleAlign() {
+    // Check if there are enough selected widgets to form a circle
+    if (selectedWidgets.length < 3) {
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2185/pr-test (or any other room on that server)

96LawDawg:
Works.  Pretty harmless even if it doesn't.  But shouldn't this go in the top menu?  Maybe right after all the alignment buttons?

RaphaelAlvez (approved):
Tested

I wish we had this in the past. I remember doing the calculations by hand for organizing holders. 

I understand this is using the traditional radians reference but most people use clocks as reference. So maybe add an icon to show the start and orientation of rotation. 

Also I think a center input would be nice and a feature to rotate around a given center so we can access circular setup.

ArnoldSmith86:
> a feature to rotate around a given center so we can access circular setup

I was thinking about adding another button to the drag toolbar that doesn't just change the `rotate` of each widget but rotates everything around the combined center point. Hmm, do we even need the current rotate button then?

PR #2211, opened 2024-08-15
improved performance when selecting multiple widgets

### LINKED ISSUE #2207: selecting a few dozen widgets is painfully slow

ArnoldSmith86:
Fixes #2207. It was regenerating the output for every selected widget instead of doing it once when all of them are selected.

It still does way too much work when selecting widgets but for now this should be fast enough.

### DIFF
--- a/client/js/editor/sidebar/json.js
+++ b/client/js/editor/sidebar/json.js
@@ -28,10 +28,10 @@ class JsonModule extends SidebarModule {
   }
 
   onSelectionChangedWhileActive(newSelection) {
-    if(newSelection.length) {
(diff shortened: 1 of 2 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2211/pr-test (or any other room on that server)

96LawDawg (approved):
Thanks.  I've noticed this forever but just didn't think about how it might be fixable.  A 12x15 grid of simple widgets took 10-12 seconds to multi-select before and took my cooling fans up to max.  Now it is about 1 second.

PR #2212, opened 2024-08-16
fixed crash while CLONEing when the given id already exists

ArnoldSmith86:
The case was already handled in the code but the error message referenced the wrong variable.


96LawDawg (approved):
Approved through code review.

PR #2214, opened 2024-08-16
fixed crash with PROPERTY x OF array

### LINKED ISSUE #2189: client crash with undefined variable
When pressing `Change Scenario` in [QuickDownload without assets 2024-08-11 2212 - Unnamed.vtt.zip](https://github.com/user-attachments/files/16576292/QuickDownload.without.assets.2024-08-11.2212.-.Unnamed.vtt.zip).

ArnoldSmith86:
Fixes #2189. By absolute chance I did `${PROPERTY x OF $var}` and `$var` evaluated to `[]`. And `isValidID` returns `true` for an empty array because it works with arrays for other reasons.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -777,7 +777,7 @@ export class Widget extends StateManaged {
         let widget = this;
         if(match[8]) {
           const id = evaluateIdentifier(match[7], match[8]);
-          if(!this.isValidID(id, problems))
+          if(Array.isArray(id) || !this.isValidID(id, problems))
             return null;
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2214/pr-test (or any other room on that server)

96LawDawg (approved):
Approved through test.  But I didn't test anything else other than that exact use case.  Not sure what else this might impact.  Anything you can think of that should be tested?

PR #2215, opened 2024-08-17
added option `visibleChildWidgets` to `INPUT.choose`

### LINKED ISSUE #2188: INPUT.choose does not show child widgets
Is this a bug or a feature? :grin:

ArnoldSmith86:
Fixes #2188. I hope this is a good solution. Children will now be visible on UI representations of a widget (like in `INPUT.choose` or the `dropTarget` buttons in the properties panel) _unless_ they are in a holder and match its `dropTarget`. So a card dropped into a holder is not rendered because it is considered a different widget in that context.

### DIFF
--- a/client/css/layout.css
+++ b/client/css/layout.css
@@ -1755,7 +1755,7 @@ input:invalid+span::after, .inputError::before {
   z-index: 2;
 }
 
-.inputchoose .widget {
+.inputchoose .inputchooseWidgetWrapper > .widget {
   position: relative !important;
   transform-origin: top left !important;
   z-index: 1 !important;
--- a/client/js/domhelpers.js
+++ b/client/js/domhelpers.js
@@ -309,7 +309,7 @@ export function formField(field, dom, id) {
           propertyOverride.activeFace = face;
 
         const widgetContainer = div(input, 'inputchooseWidgetWrapper');
-        const widgetClone = widget.renderReadonlyCopy(propertyOverride, $('body'));
+        const widgetClone = widget.renderReadonlyCopy(propertyOverride, $('body'), field.visibleChildWidgets);
         const widgetDOM = widgetClone.domElement;
         widgetClone.state.scale = scale * (field.scale || 1);
         widgetClone.domElement.style.cssText = mapAssetURLs(widgetClone.css());
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -1081,6 +1081,7 @@ function jeAddCommands() {
   jeAddFieldCommand('faces', 'choose', null);
   jeAddFieldCommand('scale', 'choose', 1);
   jeAddFieldCommand('propertyOverride', 'choose', {});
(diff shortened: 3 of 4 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2215/pr-test (or any other room on that server)

96LawDawg:
Sorry, but I don't like it.  This will probably work for 97.23% of the use cases.  But see the modified tutorial in the test room with the "Seat Token" button.  It is not too much of a stretch to think that someone would want to choose a player using seats to take some action. Seats frequently have child widgets.  That may not be the best design strategy, but it is done.

ArnoldSmith86:
> Sorry, but I don't like it.  This will probably work for 97.23% of the use cases.  But see the modified tutorial in the test room with the "Seat Token" button.  It is not too much of a stretch to think that someone would want to choose a player using seats to take some action. Seats frequently have child widgets.  That may not be the best design strategy, but it is done.

I guess we'll make it optional then?

96LawDawg:
> I guess we'll make it optional then?

That was my original suggestion and I still think that is best.  I like the way you implemented it.  But it needs to be optional.

ArnoldSmith86:
I turned it into an option `visibleChildWidgets` that can be `false` (default), `true` (the behavior described above) or `"all"` (just render all children no matter the `dropTarget`).

See the button in https://test.virtualtabletop.io/PR-2215/example for an example of the three possible values.

96LawDawg:
Wiki edits needed to choose:

* `type`: 
  * `checkbox`: _displays a checkbox_ - stores true or false
  *  `choose`: _displays images or faces of a widget_ - stores result as variable and as a collection
     * `collection`: _text_ - the name of collection the widgets will be stored in
     * `faces`: _array_ - if the mode is set to "faces", then this array limits the faces shown to the player
     * `holder`: _`id` or array of `id`s_ - widgets in this holder will be displayed for players to choose from
     * `max`: _number_ - the maximum number of selections the player can make
     * `mode`: _"widgets" or "faces" (defaults to "widgets) - specifies whether individual widgets or the faces on widgets will be displayed for players
     * `propertyOverride`: _object_ - used to set a specific property/properties while the widgets are displayed. For example: `{"activeFace": 1}` can be used to show the face up side of cards that are face down in the room
     * `scale`: _number_ - alters the size of the widgets as they are displayed in the overlay.
     * `source`: _`id` or array of `id`s or a named collection_ - widgets to be displayed for players to choose from. If both `source` and `holder` paramaters are included, the `source` parameter will be ignored.
     * `visibleChildWidgets`: _boolean or "all"_ - determines which child widgets of the source or holder widgets will be displayed. Defaults to false. When true, will display child widgets except those in a holder and matching the holder's `dropTarget`. When "all", will display all child widgets.

96LawDawg (approved):
Approved through test.

PR #2216, opened 2024-08-17
fixed crash after renaming a widget

### LINKED ISSUE #1974: renaming a widget in the JSON makes the editor lose connection to the widget
This causes a client crash when using the dragToolbar unless you click on the widget again first.

ArnoldSmith86:
Fixes #1974. When changing the `id` of a widget using the JSON editor, the editor selection will now be updated to the new id.

### DIFF
--- a/client/js/editor/sidebar/json.js
+++ b/client/js/editor/sidebar/json.js
@@ -35,7 +35,8 @@ class JsonModule extends SidebarModule {
     } else {
       jeEmpty();
     }
-    $('#jeText').blur();
+    if(!jeDeltaIsOurs)
+      $('#jeText').blur();
   }
 
   renderModule(target) {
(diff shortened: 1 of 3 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2216/pr-test (or any other room on that server)

ArnoldSmith86:
No idea why a random TestCafé test fails... :grin:

96LawDawg:
This does not work. It lets you type one letter, then the `id` value loses focus and you have to click on the `id` value again and again for each letter.  But it does fix the problem of the drag handle.

ArnoldSmith86:
> It lets you type one letter, then the `id` value loses focus and you have to click on the `id` value again and again for each letter.

I never tested more than typing one letter. :grimacing:

Now it should be working.

96LawDawg (approved):
Approved through test.

There is still one minor thing, but may be difficult to figure out and does not really matter.  If you change the type and then change the id, it loses focus after the first letter but is fine for other letters.  Strange, so just reporting but also approving.

PR #2224, opened 2024-08-18
disable F-key editor buttons on mobile

### LINKED ISSUE #2191: F-keys widget selector on mobile
It's controlled through hovering the mouse and on mobile it's only in the way. There are no F-keys anyway.

ArnoldSmith86:
Fixes #2191. After the first interaction with a touchscreen, this should disable the overlay with the F-keys which - on a touch device - only obscure the context sensitive buttons and are never useful.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2224/pr-test (or any other room on that server)

96LawDawg:
Does not seem to work.  I still get this overlay on iOS:

IMG_2611

ArnoldSmith86:
> Does not seem to work. I still get this overlay on iOS:

Yeah, you gotta interact with something in game mode before entering edit mode. I'll improve it.

96LawDawg:
> Yeah, you gotta interact with something in game mode before entering edit mode. I'll improve it.

I see that it does work once I do something in game mode.  I just went straight to edit mode.  If it becomes too difficult, this solution is better than nothing, but it would be nice to fix that so you don't have to go to game mode first because people may not know or remember that.

ArnoldSmith86:
> > Yeah, you gotta interact with something in game mode before entering edit mode. I'll improve it.
> 
> I see that it does work once I do something in game mode.  I just went straight to edit mode.  If it becomes too difficult, this solution is better than nothing, but it would be nice to fix that so you don't have to go to game mode first because people may not know or remember that.

Hmm, why don't I just check the touch on the Edit Mode button?

96LawDawg:
> Hmm, why don't I just check the touch on the Edit Mode button?

Will this still get the `body` `usedTouch` class even if the user does not click on Edit?  I think that is important for #2228.

ArnoldSmith86:
> Will this still get the `body` `usedTouch` class even if the user does not click on Edit? I think that is important for #2228.

I left the old code in place that should set it at the first touch interaction.

Looked promising on my phone.

96LawDawg (approved):
Approved through test.

PR #2228, opened 2024-08-20
added property `hotkey`

ArnoldSmith86:
Any widget can now define `hotkey` and it will be clicked whenever the given key is pressed.

Policy for public library games should be that it's never _required_ to use the key so that all games still work on touchscreens IMO.

### DIFF
--- a/client/js/mousehandling.js
+++ b/client/js/mousehandling.js
@@ -148,7 +148,7 @@ async function keyHandler(e) {
     return;
 
   batchStart();
-  for(const widget of widgetFilter(w=>w.get('hotkey')===e.key))
+  for(const widget of widgetFilter(w=>w.get('hotkey')===e.key).sort((a,b)=>String(a.get('id')).localeCompare(b.get('id'))))
     await widget.click();
   batchEnd();
 }
--- a/client/js/mousehandling.js
+++ b/client/js/mousehandling.js
@@ -148,7 +148,7 @@ async function keyHandler(e) {
     return;
 
   batchStart();
-  for(const widget of widgetFilter(w=>w.get('hotkey')===e.key).sort((a,b)=>String(a.get('id')).localeCompare(b.get('id'))))
+  for(const widget of widgetFilter(w=>w.get('hotkey')===e.key&&w.isVisible()).sort((a,b)=>String(a.get('id')).localeCompare(b.get('id'))))
     await widget.click();
   batchEnd();
 }
--- a/client/js/mousehandling.js
+++ b/client/js/mousehandling.js
@@ -143,8 +143,19 @@ async function inputHandler(name, e) {
   }
 }
 
+async function keyHandler(e) {
+  if(isLoading || overlayActive || $('body').classList.contains('edit') || e.target.tagName == 'INPUT' || e.target.tagName == 'TEXTAREA')
+    return;
+
+  batchStart();
+  for(const widget of widgetFilter(w=>w.get('hotkey')===e.key))
+    await widget.click();
+  batchEnd();
+}
+
 onLoad(function() {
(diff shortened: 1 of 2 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2228/pr-test (or any other room on that server)

96LawDawg:
> Policy for public library games should be that it's never _required_ to use the key so that all games still work on touchscreens IMO.

I agree.  But is it possible to detect in a game whether the user is on a touchscreen device.  I haven't thought this through, but it might be possible to present physical buttons to touchscreen users and make them optional in lieu of using keyboard shortcuts for other users.

96LawDawg:
By the way, this is really nice and such a surprisingly simple amount of code.  I will approve after hearing response to my previous question and this one: is it possible to specify alt-crl-cmd type keys as well.  Putting r and R work for shift, but it may be nice to have those additional options.

ArnoldSmith86:
> But is it possible to detect in a game whether the user is on a touchscreen device.

When #2224 is working properly, it adds a `usedTouch` class to the `<body>`. You can use that to hide stuff using CSS. IMO we don't need anything more official than that.

> is it possible to specify alt-crl-cmd type keys as well

It's possible but requires a lot more code (still not much but you know...). I would say, we can revisit that if the need arises.

Do you think we should reserve anything for engine-level shortcuts? For example `g` for Game Shelf or whatever.

96LawDawg:
> When #2224 is working properly, it adds a `usedTouch` class to the `<body>`. You can use that to hide stuff using CSS. IMO we don't need anything more official than that.

Agree

> It's possible but requires a lot more code (still not much but you know...). I would say, we can revisit that if the need arises.

Agree

> Do you think we should reserve anything for engine-level shortcuts? For example `g` for Game Shelf or whatever.

No.  We moved away from that in the JSON Editor a while back and since we would only officially be supporting on keyboard devices, I don't think there is a need.  Plus we would have to have the discussion about which ones.  Keyboard shortcuts are useful in a lot of instances, but I don't think basic interaction with the VTT UI is one where we need it.

96LawDawg:
Another question.  What happens if multiple widgets are assigned the same hotkey?  It activates all of them, but in what order? 
 Probably doesn't matter, but I can't check because you can't use the hotkeys in Edit Mode and I can't see the debug output.

Also, say you have a one button that flips each player clicks to flip over one card from your own pile.  If I click the button, I can use predefined variables like `seatIndex` to see which player pressed it and take appropriate action. That same thing worked in testing because I configured the button to use `seatIndex` and pressing the hotkey was the same as pressing the button.  Obviously that is the desired outcome, but have we thought that through?  Are there any complications I'm missing?

ArnoldSmith86:
> Obviously that is the desired outcome, but have we thought that through? Are there any complications I'm missing?

I can't think of anything. It simply calls `widget.click()`, same as clicking with the mouse.

> It activates all of them, but in what order?

It's probably the order they are in the `widgets` array. :grinning: I'll check if it's trivial to do it alphabetically by `id`, just so it's defined behavior.

96LawDawg (approved):
Approved.  Order ultimately doesn't matter.  If it does, then the developer needs to find a solution or not cause that problem in the first place.

ArnoldSmith86:
One thing I came across: it's probably a good idea to check if the widget is visible. In Runway Roll, I could assign a hotkey to the roll dice button, but every player has their own.

RaphaelAlvez:
> One thing I came across: it's probably a good idea to check if the widget is visible. In Runway Roll, I could assign a hotkey to the roll dice button, but every player has their own.

This feels essential to me. You should not be able to press anything owned or hidden by seats linking (either inactive or only visible to seat)

ArnoldSmith86:
Added visibility check. It doesn't check if the widget is obscured by another widget though. I don't think that warrants way higher complexity.

PR #2229, opened 2024-08-20
fixed selection behavior for multi-select and changing id or type

### LINKED ISSUE #2222: if you change the type and then change the id, it loses focus after the first letter
There is still one minor thing, but may be difficult to figure out and does not really matter.  If you change the type and then change the id, it loses focus after the first letter but is fine for other letters.  Strange, so just reporting but also approving.

_Originally posted by @96LawDawg in https://github.com/ArnoldSmith86/virtualtabletop/pull/2216#pullrequestreview-2244309992_

ArnoldSmith86:
Fixes #2222 and #2187. Changing `type` or `id` should now always keep the selection and tree updated. The same should be true for changing multi-selection using the `widget` section in the JSON editor (including using regular expressions).

It is really nice that manipulating multi-selection is so much faster after #2211. Should make it a lot more useful.

### DIFF
--- a/client/css/jsonedit.css
+++ b/client/css/jsonedit.css
@@ -570,24 +570,3 @@ body.jsonEdit #toolbar {
 #jeCommands #jeCommandOptions input[type="number"]{
   width: 6ch;
 }
-
-
-
-
-/* FIXME: put this where it makes sense */
-
-.jeLineNumber {
-  display: inline-block;
-  width: 40px;
-  text-align: right;
-  padding-right: 10px;
-  color: #444;
-}
-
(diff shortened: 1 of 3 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2229/pr-test (or any other room on that server)

ArnoldSmith86:
Oops, commited to the wrong branch by accident. Will clean this PR up later.

96LawDawg (approved):
Nice!  Approved through test.

PR #2232, opened 2024-08-21
added property search to multi-select and tree filter

ArnoldSmith86:
You can now use the syntax `image:game-icons` in the filter box above the widget tree and in the `widgets` part of the JSON editor multi-selection mode. It will then display/select all widgets that contain `game-icons` (case-insensitive) in their `image` property.

It could be expanded to support regular expressions as well or check multiple properties but that feels like overkill.

The idea for this came when I found widgets with a property that was clearly a copy-paste error and I wanted to know if there were more of them.

### DIFF
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -1733,11 +1733,17 @@ function jeSelectSetMulti(widgets) {
 function jeMultiSelectedWidgets() {
   let selected = [];
   for(const search of jeStateNow.widgets) {
+    const isRegex = search.match(/^\/(.*)\/([a-z]+)?$/);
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2232/pr-test (or any other room on that server)

96LawDawg:
> The idea for this came when I found widgets with a property that was clearly a copy-paste error and I wanted to know if there were more of them.

Don't get me wrong, I like this.  It is a very nice addition.  But did you try the new search and replace feature?

96LawDawg (approved):
Approved through test.

PR #2231, opened 2024-08-20
PL Egyptian Epochs: Usability Improvements

andypymont:
Updates to the public library game _Egyptian Epochs_.

Most notably:
- Add the ability for players to toggle "Pass" during the auction by clicking on the bid holder to show an "x" there. (This is also now mentioned in the module usage instructions.)
- Add a transition animation so it's clearer when a tile, especially a Horus tile, is drawn with the "Draw Tile" button.

Also a few other minor tweaks:
- Fix a 2-pixel misalignment of the New Game and New Epoch buttons
- Fix a slight difference in the shade of red on two types of disaster tile
- Reword the pharaoh tile scoring aid text to try and avoid a rendering issue reported by some issues on Chrome/Brave on Windows 10

PR #2241, opened 2024-08-21
PL Egyptian Epochs: Reset pass markers when End Turn pressed

andypymont:
Reset the new pass markers added in #2231 so that they reset to the blank side when the End Turn button is clicked, as well as when the Win buttons are used.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2241/pr-test (or any other room on that server)

96LawDawg (approved):
Approved.

PR #1908, opened 2023-11-02
added link to game

ArnoldSmith86:
This PR enables the `Share Link` entry of the game menu for public library games as well. Pressing it (for any game) now produces a URL like https://test.virtualtabletop.io/PR-1908/game/agra for public library games or https://test.virtualtabletop.io/PR-1908/game/mykke926/red-rising for custom games.

The URL then looks like this and allows players to easily add the game to a new room:

image

- [x] The overlay disappears immediately sometimes. Reloading usually helps. **Apparently this happens every time on the PR test server and reloading doesn't help.**
- [x] Re-sharing a shared game returns the raw VTT file URL (`.../s/...` instead of `.../game/...`).
- [x] Make it work with public library games.
- [x] Make it work with tutorials and assets. Should just need changing the filter box.
- [x] Improve the "Loading room" thingy.
- [x] Layout fixes for viewport sizes of the welcome overlay.
- [x] Replace hard-coded room ID `s91j` by a randomly generated but unused room ID.
- [x] Replace hard-coded `Animal Race` in the text by the actual game name.
- [x] Using the Back button of the browser restores the `/game/` URL but doesn't show the overlay again.
- [x] Fix TestCafé. (Fixed in #2010.)
- [x] `ensureRoomIsLoaded(req.params.room || req.params.id)` loads the share ID as a room. This needs refactoring.
- [x] Remove "user-generated content" warning for public library games.
- [x] Show game name in `<title>`.
- [x] Add Open Graph support for game URLs.
- [x] The Close button doesn't work yet.
- [x] Text on the Share Link overlay doesn't make a lot of sense for public library games.
- [x] `COPY TO CLIPBOARD` should probably be called `SHARE` now.
- [x] Handle non-existent games.
- [x] Empty rooms should show the Game Shelf again.
- [x] Share button breaks layout when clicking.
- [x] Enable ENTER when entering name and room.
- [x] Changing your name from the overlay does not work.
- [x] A way to get away from "Game not found".
- [x] Fix `/game/crafty2`.
- [x] Make `/tutorial` only load tutorials. Same for games.
- [x] Fix code test error.
- [ ] Sometimes the newly entered room reloads immediately.
- [x] Looks like I broke adding the game to the new room.
- [x] Custom games with no metadata break the layout a bit.
- [x] The message about "user-generated content" breaks has broken layout.
- [x] Sharing a TTS import shows the workshop URL instead.
- [x] Loading a room doesn't highlight "Active Game" in the toolbar anymore.
[...]

PR #2244, opened 2024-08-21
fixed that PCIO importer would not use filename as name

ArnoldSmith86:
Broken since #1908...


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2244/pr-test (or any other room on that server)

PR #2245, opened 2024-08-21
PCIO importer improvements

### LINKED ISSUE #2151: PCIO Import Issues in Game
Problem PCIO attached as .zip file.

[logicards-all-seasons.zip](https://github.com/user-attachments/files/15916847/logicards-all-seasons.zip)

Reported errors:

- Cards dealt seemed to show the wrong face
- There was a randomizer (I think?) but doesn't seem to work in vtt
- There is a deck that's meant to click through stares but instead does a pop-up selection in vtt?

The last point is a chooser in PCIO for a 2-sided card.  VTT handles the import correctly.  I think that is just poor design choice in the PCIO game.  I don't have time now to look at the card face issues, but there is a difference between PCIO (which deals all but one face down) and VTT which deals them all face up.

ArnoldSmith86:
Fixes #2151. See commit messages. The randomizer is probably a different PCIO file posted in https://boardgamegeek.com/thread/3224205/article/44567855#44567855 which seems to work just fine.

Putting the tokens onto the cards is a lot prettier in PCIO than in VTT but it is close enough, I guess. We gotta overhaul holders at some point...

It feels pretty weird to simply change the default of FLIP like that. But to prove that the default used to be different, we would need old PCIO files where we are sure about the old behavior...

### DIFF
--- a/server/pcioimport.mjs
+++ b/server/pcioimport.mjs
@@ -1020,7 +1020,7 @@ export default async function convertPCIO(content) {
 
           c = importWidgetQuery(w.clickRoutine, c.args, 'holders', 'holder', 'collection', {
             func:   'FLIP',
-            count:  !c.args.flipMode || c.args.flipMode.value == 'pile' ? 0 : 1
+            count:  !c.args.flipMode || c.args.flipMode.value != 'pile' ? 1 : 0
           });
           if(c.holder && c.holder.length == 1)
             c.holder = c.holder[0];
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2245/pr-test (or any other room on that server)

96LawDawg (approved):
Approved through test on Logicards.  I did not test anything else, but this imports smoothly now.

PR #2250, opened 2024-08-22
added property `sticksTo`

ArnoldSmith86:
This was unbelievably easy. I only tested a simple `sticksTo: { type: card }` on the freeze tokens in Duel 52.

### DIFF
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -23,7 +23,7 @@ export const dropTargets = new Map();
 export const clientPointer = $('#clientPointer');
 
 function compareDropTarget(widget, t, exclude){
-  for(const dropTargetObject of asArray(t.get('dropTarget'))) {
+  for(const dropTargetObject of asArray(t)) {
     let isValidObject = true;
     for(const key in dropTargetObject) {
       if(dropTargetObject[key] != widget.get(key) && (exclude == true || (key != 'type' || widget.get(key) != 'deck' || dropTargetObject[key] != 'card'))) {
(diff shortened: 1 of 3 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2250/pr-test (or any other room on that server)

ArnoldSmith86:
Adding `sticksTo: { type: card }` to the `cardDefaults` of a normal deck results in infinite recursion crashes. Will investigate at some point.. :grinning:

96LawDawg:
> Adding `sticksTo: { type: card }` to the `cardDefaults` of a normal deck results in infinite recursion crashes. Will investigate at some point.. 😀

I was just about to test that, because that is another common use case.

96LawDawg:
Bug in https://test.virtualtabletop.io/PR-2250/lawdawg.  I put sticksTo on several poker chips.  You can move one a time or two, but then the room freezes (max call stack, which I think is recursion).

ArnoldSmith86:
> Bug in https://test.virtualtabletop.io/PR-2250/lawdawg. I put sticksTo on several poker chips. You can move one a time or two, but then the room freezes (max call stack, which I think is recursion).

Yeah, probably the same thing..

ArnoldSmith86:
You can make `dropOffsetX` work yourself:

```JSON

  "enterRoutine": [
    {
      "func": "SET",
      "collection": "child",
      "property": "x",
      "value": "${PROPERTY dropOffsetX}"
    },
    {
      "func": "SET",
      "collection": "child",
      "property": "y",
      "value": "${PROPERTY dropOffsetY}"
    }
  ]
```

ArnoldSmith86:
I don't feel like working on this when `dropTarget` simply works on cards. We can re-open it if we feel like it.

PR #2254, opened 2024-08-23
SHIFT+F<X> now deselects the widget if it was already selected

### LINKED ISSUE #2247: deselecting using SHIFT+F1 does not work

ArnoldSmith86:
Fixes #2247. Looks like this was never a thing...

### DIFF
--- a/client/js/editor/dragButton.js
+++ b/client/js/editor/dragButton.js
@@ -27,13 +27,33 @@ class DragButton extends ToolbarButton {
     const dx = coords.clientX - this.mousedownEvent.clientX;
     const dy = coords.clientY - this.mousedownEvent.clientY;
 
-    this.domElement.parentElement.style.right = (window.innerWidth-this.mousedownRect.right - dx) + 'px';
-    this.domElement.parentElement.style.top   = (                  this.mousedownRect.top   + dy) + 'px';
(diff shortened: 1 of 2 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2254/pr-test (or any other room on that server)

96LawDawg:
Just making sure this works as expected.  Shift-F1 does not always deselect the current widget.  It only does so if the cursor is also over the widget (not sure about mobile; did not test).  I'm okay with that.  Is that what you intended?  If not, I'm not sure how it would work with multiple widgets selected.

RaphaelAlvez:
> Just making sure this works as expected. Shift-F1 does not always deselect the current widget. It only does so if the cursor is also over the widget (not sure about mobile; did not test). I'm okay with that. Is that what you intended? If not, I'm not sure how it would work with multiple widgets selected.

I think it is as intended. this bothered me many times. if you are selecting with shift+Fx you cannot deselected and need to restart the selection if you mess up.

RaphaelAlvez (approved):
code looks ok and tested.

I didn't even know it was a bug but this bothered me so much in the past

PR #2258, opened 2024-08-24
added smart cloning to editor

ArnoldSmith86:
The goal is basically to be able to clone a grid of widget groups first (like the current clone tool already does) and then all changes to the original should immediately apply to the clones. All automation and relations between widgets should be fixed automatically.

This was already my goal when adding the duplicate feature in the JSON editor and when I added `inheritFrom`. But while those are already super useful, I always thought it could be a lot smarter...

This uses a property `editorSmartClone` that is exclusively recognized in the editor. It won't have any effect while playing. There was already a property `editorGroup` so the editor can recognize a group it created earlier.

In time, I will probably introduce more `editor*` properties that basically tell the editor and the user that the editor is in charge of parts of the widget definition. So if the user wants to regain full control, they have to remove those properties (can be done through editor UI of course).

Other use-cases for `editor*` properties would be generated decks so the editor knows that it generated the deck itself and can show simplified UI to edit it. Or in a automation editor where it could show simplified versions of groups of routine operations it created itself.

---

Options for `editorSmartClone`:
- `flipX` / `flipY`: true/false/"all" - if `true` the layout of the first level will be flipped; if `"all"`, all levels of children will be flipped (this only sets `x` or `y`, nothing else)
- `replaces`: object - string in key will be replaced by string in value in all properties of the children
---

Still very much WIP but at this point it is fun to play around with.

Things I want to do:

- [x] Add a new button to the drag toolbar or make this the default for the grid cloning.
- [x] Redirect selection from a cloned child to the topmost clone with `editorSmartClone` because editing the children is not a good idea.
- [x] Replace source `id`s by clone `id`s automatically (without having to add them to `replace`).
- [x] Make `flipX` and `flipY` optionally just flip the first level of children (should be the default).
- [x] Test and fix recursion (so one smartClone inside another one).
- [x] Make seats receive unique indices.
- [x] Make seat indices smarter.
- [x] Exclude cards from cloning unless their deck is part of the group. Can be a setting.
- [x] Add JSON editor commands for the `editorSmartClone` settings.
- [x] Show a UI in the Properties panel when selecting a clone, allowing to change settings or remove the link entirely.
- [x] It doesn't unset properties on the clones if they got unset in the source.
- [x] Update references to internal widgets when they don't get caught by `replaces`.
- [x] `score` for seats should be excluded dynamically (as stated in scoreboard `scoreProperty`).
- [x] `inheritFrom` should be updated when it should changes (for example changing `movable`).
- [x] If properties were set but a change leads to them being identical again, they can be removed.
- [x] IDs in `${PROPERTY x OF id}` syntax still need to be updated.
- [x] Remove debug output and re-enable selection filtering before merging.
- [x] ~~Check if it makes sense to update references from outside the source to the source. For example a button that deals a card to a player area. When smart-cloning the player area, it could add the clones to the array of destinations.~~
- [x] Find a solution for bad seat sizing. Probably use clientRect. ~~Or an exception for seats.~~
- [x] Dragging a movable widget out of a smart clone doesn't remove it from `smartCloneSourceMap`.
- [x] Changing `flipX` (and probably others) doesn't update anymore. Also see #2304.
- [x] Renaming top-level clone did not update button clickRoutine.
- [x] The smart clone button should insert an editorGroup parent or be more dynamic about how the top level clone looks like. (Cloning a dice makes it inherit `activeFace` for example.)
- [x] Gotta handle receiving a new state.
- [x] Test stuff with two clients having edit mode active.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2258/pr-test (or any other room on that server)

VTT-AI-Agent:
Finished off the remaining checklist items:

- **Recursion**: fixed — deltas that arrive while a smart-clone update is in progress are now queued and processed afterward instead of being dropped, so editing a clone that is itself a smart-clone source now cascades correctly into its own clones.
- **Smarter seat indices**: clones now take the lowest unused seat index instead of `max+1`.
- **Exclude cards unless their deck is part of the group**: cards are skipped during cloning unless their deck lives inside the cloned subtree; added an `includeCards` option (checkbox in the Properties panel + JSON editor command) to opt back in.
- **JSON editor commands for `editorSmartClone` settings**: added commands for adding a `replaces` text substitution and toggling `includeCards`, alongside the existing flipX/flipY commands.
- **Debug output / selection filtering**: removed all `console.log` calls and re-enabled `smartCloneProcessSelection` (it was short-circuiting with an early `return selection`).
- **Dice inheriting `activeFace`**: top-level clones now reuse the same `inheritDef` exceptions used for children, so a directly-cloned dice/card/seat/etc. no longer inherits state it shouldn't.
- **Handle receiving a new state**: the smart-clone source map is now rebuilt whenever a fresh room state is received.
- **Two clients in edit mode**: verified with two concurrent edit-mode browser sessions; clone children now get a deterministic ID (`${parentID}-${childID}`) before falling back to a random one, so both clients converge on the same widget IDs instead of creating duplicates.

Also merged the last few weeks of `main` into the branch (conflicts were in the clone drag button and editor layout, both from the grid-clone rewrite in `#[latest]`), added `editorSmartClone` to `validate_gamefile.js`, and fixed a delta-encoding edge case where unsetting a property on a widget still being created in the same batch produced a stray `null` value instead of omitting the key.

Tested manually via Playwright: basic smart clone, recursive (nested) smart clone, cascading updates/deletes through nested clones, flipX/flipY, and two simultaneous edit-mode clients.

VTT-AI-Agent:
Correction to my previous comment: the merge conflicts in the clone drag button/editor layout came from #2824 ("Add smarter drag options"), not the placeholder I accidentally left in.

VTT-AI-Agent:
CI failure fixed: the \`Reward\` game's TestCafe state-hash test was failing on both browsers.

Root cause: my earlier commit changed \`sendPropertyUpdate\` in \`client/js/serverstate.js\` so that unsetting a property during widget creation dropped the key from the delta instead of sending an explicit \`null\`. That's a global change affecting **all** widget creation in the app, not just the editor's smart-clone flow — and the "null now, real value moments later" pattern turns out to be normal elsewhere too (e.g. \`Widget.clone()\` creates a card with placeholder \`x\`/\`y\` and corrects it once mounted). Deleting-then-reinserting the key reordered properties in the resulting JSON state without changing any actual value, which changed the MD5 hash the Reward test compares against.

Reverted that change. I also explored fixing the underlying asymmetry server-side (\`server/room.mjs\` treats \`null\` differently for widget creation vs. update, so a smart-clone widget like \`area1\` can end up with a literal \`"width":null\` in the saved state) but that touches widget-creation for every game and risks the same kind of hash regression elsewhere, so I left it out of scope for this PR — it's cosmetic only, since the client's own \`widget.state\` never stores the stray null (\`applyDelta\` already strips nulls, even during creation).

Verified locally: \`npm test\` passes, and I reproduced the exact CI-reported hash before the fix and the expected hash after it via a scripted browser run of the Reward game's test steps. Re-ran the smart-clone Playwright tests (basic clone, recursion, cascading delete, flipX, two-client sync) — all still pass.

VTT-AI-Agent (commented):
## TL;DR
Nice, self-contained editor-only feature (`editorSmartClone`) that has no effect during play — it touches only editor/validator/build files, not the engine, so it's safe for the server and existing games. The one thing I'd flag before this leaves WIP: `applyReplaces()` feeds raw widget IDs and user-supplied `replaces` keys into `new RegExp(...)`, so any ID/key containing a regex metacharacter will either mis-replace or throw and abort the whole clone update.

## Findings

### 1. Regex injection in `applyReplaces()` (client/js/editor/smartClone.js:297-313)
```js
const regex = new RegExp(from, 'g');
replacedValue = replacedValue.replace(regex, to);
```
`from` comes from two places, neither of which is escaped:
- Auto-injected source IDs: `modifiedReplaces[`"${source.id}"`]`, `modifiedReplaces[` OF ${source.id}\}`]`, and `modifiedReplaces[source.id] = cloneID`.
- User-supplied `replaces` keys (the "SOURCE TEXT" entries added via the JSON editor / properties UI).

Widget IDs are arbitrary user strings and routinely contain characters that are regex-special (`.`, `(`, `)`, `[`, `]`, `+`, `?`, `*`, `$`, `|`, `\`). Concretely:
- An ID like `card.1` used as a source would match `card_1`, `cardX1`, etc. in every child property — silently corrupting the clone's automation/references.
- An ID like `deck[1]` (or a `replaces` key like `a[`) produces an invalid pattern, so `new RegExp(from)` throws. Because this runs inside `smartCloneUpdateClone`/`smartCloneProcessDelta`, the exception aborts the update mid-way, leaving clones partially updated.

The `to` side is also unescaped for `String.prototype.replace`, so a `$` in an ID (or in a clone ID) would be interpreted as a replacement pattern (`$&`, `$1`, …). Recommend escaping the auto-generated ID entries (a small `escapeRegExp` helper) and either escaping user `replaces` keys or documenting that they are regexes. Whether the user-facing `replaces` is intended to be literal or a pattern is worth deciding explicitly.

### 2. Delta queue is silently truncated (client/js/editor/smartClone.js:470-480)
```js
for(let i=0; queuedDeltas.length && i<100; ++i)
  await smartCloneProcessDelta(queuedDeltas.shift());
queuedDeltas.length = 0;
```
The 100-iteration cap is a reasonable runaway guard, but if it's ever hit the subsequent `queuedDeltas.length = 0` discards any still-queued deltas rather than processing them, so a burst of >100 cascading updates would drop the tail. Consider `log`/warn when the cap trips, or only clearing what was actually processed, so a real overflow isn't invisible.

### 3. Minor: `renderForSmartClone` writes checkbox state after wiring `onchange` (client/js/editor/sidebar/properties.js:1613-1637)
Setting `.checked` programmatically doesn't fire `onchange`, so this is fine as written — just noting that the three `Object.assign({}, widget.get('editorSmartClone'), …)` handlers each re-read the live value, which is the correct pattern here. No change needed; flagging only because it's easy to regress.

## Scope / engine note
This PR changes only `client/js/editor/*`, `client/js/jsonedit.js`, `server/minify.mjs` (adds the new file to the editor bundle), and `validator/validate_gamefile.js` (registers `editorSmartClone` as a valid `object` property). It does **not** touch `client/js/widgets/`, `client/js/compute.js`, or `client/js/serverstate.js`, and the new property is explicitly ignored outside edit mode — so it does not alter how existing games run and is not an engine change. The validator addition is the correct upkeep for the new property. `npm test` behavior is unaffected (no server-side logic changed).

---
I added the `beta` label: the change is editor-only, gated behind the new opt-in `editorSmartClone` property, touches no engine files, and can't corrupt games or crash the server during play, so it's safe and useful for players to try even though it's still WIP.

VTT-AI-Agent (commented):
## TL;DR

The current head still has three functional gaps in the core smart-clone lifecycle. Most importantly, nested clones copy play-state properties that `inheritDef()` explicitly excludes, so rolling a source die or changing a timer while an editor is open can overwrite the clone's independent state.

## Findings

### 1. Nested clones still synchronize properties that are supposed to remain independent (`client/js/editor/smartClone.js:82-106`, `client/js/editor/smartClone.js:138-159`)

`inheritDef()` excludes runtime state such as dice `activeFace`/`rollCount`, timer `milliseconds`/`paused`, spinner `value`, movable-widget `owner`, and each scoreboard's dynamic seat score property. However, `smartCloneUpdateClone()` does the opposite for every excluded property: when `canBeInherited` is false, it writes `source.get(property)` into the clone as a local value.

For example, put a dice inside a group, smart-clone the group, leave one client in edit mode, and roll the source dice from another client. The source delta schedules an update at line 325, and the cloned dice gets the source `activeFace` despite `!activeFace`. The same issue makes a custom `scoreProperty` such as `points` track the source seat (the hard-coded seat skip only covers literal `score`). This defeats the commit's stated goal of excluding properties that change during play and means the supposedly editor-only feature can alter live game state whenever an editor is open. The updater needs to skip the runtime exception set rather than materialize those values during every reconciliation.

### 2. Moving a source child out of the source group leaves a stale child in every clone (`client/js/editor/smartClone.js:3-16`)

`smartCloneRemoveChildren()` receives the expected `source` parent but never checks it. If source child `C` is reparented away from source group `A`, `C` still exists and has the same type, so none of the removal conditions on line 11 fires. `smartCloneAddChildren()` no longer sees `C` under `A`, but the existing source-map entry remains and `smartCloneUpdateChildren()` continues updating the stale cloned child under `A`'s clone.

Reconciliation should also treat `childSource.get('parent') != source.id` as removal from this cloned subtree. This is especially important with flipping enabled: moving that source child to the room root leaves `widgets.get(source.get('parent'))` undefined in the flip calculation and aborts the async update mid-batch.

### 3. “Unlink” removes only the marker, not the source links (`client/js/editor/sidebar/properties.js:1616-1632`, `client/js/editor/smartClone.js:311-312`)

The button promises to unlink the clone so its children can be edited independently, but its handler only sets `editorSmartClone` to `null`. Delta processing then drops the in-memory source map; it never removes or materializes the top clone's or descendants' `inheritFrom` definitions. Because smart clones intentionally keep most values sparse, subsequent source changes continue propagating through the normal inheritance engine after “Unlink”.

Unlinking needs to freeze the currently resolved values for the full cloned subtree and remove those generated `inheritFrom` links (while preserving any unrelated inheritance), otherwise it does not provide the advertised escape hatch.

## Scope / beta note

The actual diff changes only editor code plus `client/js/jsonedit.js`, `server/minify.mjs`, and the validator; it does not change `client/js/widgets/`, `client/js/compute.js`, or `client/js/serverstate.js`. The feature is gated behind new edit-mode UI/state, so it is not an engine change for existing games by default, despite the PR currently carrying an `engine` label.

The `beta` label was already present, so I left it unchanged; this is opt-in editor behavior rather than an engine change, but the runtime-state synchronization issue above makes the current build risky to test in rooms where edit mode remains open.

VTT-AI-Agent:
Addressed review 4/5 and the two still-open findings from the prior cycle in `e7c7fa95` (plus the current-`main` merge and test stabilization):

- **Independent runtime state:** reconciliation now uses the same generated `inheritDef()` filter to skip properties that must remain local. This covers dice/card faces, roll counts, timer/spinner state, owners, editable label text, and dynamic seat score properties rather than only literal `score`.
- **Reparented source children:** child reconciliation now removes a cloned subtree when its mapped source no longer belongs to the expected source parent, before attempting recursive updates/flips.
- **Unlink:** the button now materializes each clone’s currently resolved inherited values, removes only the generated source entry from `inheritFrom`, retains unrelated inheritance entries, and clears nested smart-clone markers.
- **Literal replacements:** `replaces` no longer constructs regular expressions; metacharacters and `$` are handled as ordinary text.
- **Delta overflow:** the 100-item guard now yields and resumes the remaining queue on the next task instead of silently discarding it.

Added a TestCafe lifecycle regression covering literal replacement, dice and custom-score independence, source-child reparenting, and unlink/freeze behavior with unrelated inheritance retained. Validation: `npm test -- --runInBand` (15/15) and the full Chromium editor TestCafe file (4/4) pass locally on port 8321.

VTT-AI-Agent (commented):
## TL;DR

Nice, ambitious feature — the smart-clone flow is well thought through and CI is green. This is editor-only (the new `editorSmartClone` property is never read by the engine, so it does **not** change how existing games play). One concrete bug worth fixing before this leaves WIP: the canvas-cell exception list in `inheritDef()` is built with a double-quoted string instead of a template literal, so canvas widgets inside a smart clone never actually get their `c00`–`c99` cells excluded. A couple of smaller robustness notes below.

## Findings

### 1. Canvas cell exceptions are never generated (`client/js/editor/smartClone.js:91`)

```js
if(type == 'canvas')
  for(let x=0; x<10; ++x)
    for(let y=0; y<10; ++y)
      exceptions.push("!`c${x}${y}`");
```

The pushed value is a **double-quoted** string, so `${x}`/`${y}` are not interpolated — every iteration pushes the identical literal `!\`c${x}${y}\`` (collapsed to a single entry by the later `new Set(...)`). The compare in `inheritFromIsValid` looks for `'!'+key`, i.e. `!c00`, `!c01`, … which this literal never matches. Net effect: for a `canvas` widget inside a smart-cloned group, the drawing cells `c00`–`c99` are treated as normal inherited properties instead of being kept independent — the opposite of what the `!` prefix intends (compare the seat handling on line 99, which correctly uses a backtick template). It should read:

```js
exceptions.push(`!c${x}${y}`);
```

### 2. Top-level source-id replacement is an unquoted substring replace (`client/js/editor/smartClone.js:124`)

For child ids you use quoted keys (`"${source.id}"`, ` OF ${source.id}}`) which are appropriately targeted, but for the top clone:

```js
modifiedReplaces[source.id] = cloneID;
```

replaces the raw source id everywhere in every stringified property via `split(from).join(to)`. Because VTT ids are short, this can collide with unrelated substrings — e.g. a source id that appears inside a label's `text`, or as a substring of another widget's id — and silently corrupt that value on the clone. Consider quoting/boundary-anchoring this the same way the child replacements are, or at least noting the limitation.

### 3. `validPropertiesOfWidget` can throw if an inheritFrom target is gone (`client/js/editor/smartClone.js:~110`)

```js
for(const [ id, filter ] of Object.entries(widget.inheritFrom()))
  properties.push(...validPropertiesOfWidget(widgets.get(id), filter));
```

If `id` references a widget that no longer exists, `widgets.get(id)` is `undefined` and the recursive call dereferences `.state` → TypeError. `smartCloneGetSource` guards the primary source, but a secondary `inheritFrom` entry (like the `base`/`classes` case in your own test) that gets deleted would not be guarded. A `widgets.has(id)` check here would make the update pass resilient to a source being removed mid-edit.

### Minor
- `renderForSmartClone` exposes `flipX`/`flipY` as checkboxes, so the `"all"` (all-levels) variant documented in the PR body is only reachable via the JSON editor — intentional, just flagging the UI gap.

## Beta

The `beta` label is already present and I'm leaving it: this PR touches only editor/build/test/validator files — `editorSmartClone` is read exclusively by editor code and never by the widgets, `compute.js`, or `serverstate.js`, so it has no effect on how games play and cannot corrupt saved games or crash the server, making it engine-ineligible for exclusion and safe for player testing even in its current WIP state.

VTT-AI-Agent:
Thanks for the review — addressed all three findings in c721acfb:

1. **Canvas cell exclusion (confirmed bug)** — `inheritDef` built the `c00`–`c99` exception list with a double-quoted string (`"!\`c${x}${y}\`"`), which pushed the literal text rather than the interpolated cell name, so canvas cells were never actually excluded from inheritance. Switched to a template literal (`` `!c${x}${y}` ``); verified it now yields `!c00`…`!c99`.

2. **Unquoted ID replacement** — the top-level branch of `applyReplaces` did a bare substring replace of the source ID, which could corrupt any value merely containing the ID (e.g. a label reading "Player area" became "Player area1"). Removed the special case so every ID is replaced only where it appears as a real widget reference: a quoted JSON string (`"id"`) or the `... OF id}` compute syntax — the same forms already used for child clones. Unit-checked that `"label1"` → `"area1-label1"` and `${PROPERTY x OF area}` → `${PROPERTY x OF area1}` still update correctly while `"Player area"` is left untouched. The bare-id calls elsewhere (`applyReplaces(child.id, …)`) are unaffected — they never matched the removed top-level bare mapping and still fall through to the deterministic `${clone.id}-${child.id}` path.

3. **`validPropertiesOfWidget` throwing on a deleted `inheritFrom` target** — added a `widgets.has(id)` guard before recursing so a secondary `inheritFrom` source deleted mid-edit no longer crashes the update.

`npm test` is green (15/15). I couldn't run the `Smart clone lifecycle` TestCafe test locally (headless-Chromium can't interact with the `<input type=color>` picker in `setName`, which fails before any smart-clone code runs), so I validated the reference-updating and canvas-exclusion behavior with targeted unit checks and a Playwright integration run instead; CI will exercise the full TestCafe test.

Agreed on leaving `beta` — this stays editor-only (`editorSmartClone` is never read by widgets/`compute.js`/`serverstate.js`).

Issue #2259, opened 2024-08-24
selecting multiple widgets doesn't update JSON editor command pane


ArnoldSmith86:

bjalder26:
This seems to be where the commands are changed in the JSON editor.
```js
  for(const module of sidebarModules)
    module.onSelectionChanged(selectedWidgets, previousSelectedWidgets);
```
This is the function:
```js
function setSelection(newSelectedWidgets) {
  const previousSelectedWidgets = [...selectedWidgets];
  selectedWidgets = newSelectedWidgets;

  for(const widget of previousSelectedWidgets)
    widget.setHighlighted(false);

  for(const widget of selectedWidgets)
    widget.setHighlighted(true);

  for(const button of toolbarButtons)
    button.onSelectionChanged(selectedWidgets, previousSelectedWidgets);
  for(const button of dragToolbarButtons)
    button.onSelectionChanged(selectedWidgets, previousSelectedWidgets);
  for(const module of sidebarModules)
    module.onSelectionChanged(selectedWidgets, previousSelectedWidgets);

  updateDragToolbar();
}
```
It looks like `newSelection.length == 1` will evaluate as false when there is a multi-selection. 
```js
  onSelectionChanged(newSelection, oldSelection) {
    this.setState(newSelection.length == 1 && newSelection[0].get('editorGroup'));
    const sameParent = newSelection.length > 1 && newSelection.every(w=>w.get('parent')==newSelection[0].get('parent'));
    this.setMinimumSelection(this.active ? 1 : (sameParent ? 2 : 9999999));
  }
}
```

PR #2260, opened 2024-08-24
added multi-selection commands for setting a parent

ArnoldSmith86:
Select multiple widgets, click into the JSON editor until #2259 is fixed, click `make child`. It will set the `parent` for all widgets to the biggest selected widget and set `x` and `y` to compensate for the parent coordinates.

Should probably get a better name and be smarter - right now it only works if all widgets had the same `parent` before and are not scaled.

---

Todo:
- [x] Rename the button. :grin: 
- [x] Provide a way for the user to select the new parent.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2260/pr-test (or any other room on that server)

96LawDawg:
Probably this works most of the time.  Would it make more sense (and not be too much trouble) to select the widgets you want to re-parent and then after pressing "make child" you select the new parent (either by clicking in the room, or entering an id)?  I guess most of the time the largest widget would be the norm, but not always.  And that is not intuitive.  For example, if I have a seat, a holder, and some tokens I could select them all thinking the seat is going to be the parent.  But it would be the holder.

And instead of "make child," which sounds vaguely like a reproductive act, if you make my suggested change, then maybe the button is "select parent."

ArnoldSmith86:
I renamed the existing command to `set biggest as parent` and added a new `enter new parent ID` command that spawns a text input option.

Ideally you would instead either select the new parent by clicking on it or offer the selected widgets in a dropdown but both of those need changes to the way JSON commands work.

96LawDawg (approved):
Approved through test

PR #2275, opened 2024-09-01
added compute function `sum`

ArnoldSmith86:
Takes an array, returns the sum of its elements. Should this be more fancy to handle values of an object or anything other than numbers in the array?

### DIFF
--- a/client/js/compute.js
+++ b/client/js/compute.js
@@ -658,6 +658,13 @@ const compute_ops = [
     hash: '0f1f81e18571c2b22e227c988a37b189'
   },
   {
+    name: 'sum',
+    desc: 'calculates the sum of the elements of an array',
+    sample: 'var a = sum ${x}',
+    call: function(v, x) { return v = x.reduce((partialSum, a) => partialSum + a, 0) },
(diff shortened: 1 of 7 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2275/pr-test (or any other room on that server)

96LawDawg:
> Takes an array, returns the sum of its elements. Should this be more fancy to handle values of an object or anything other than numbers in the array?

No.

Slightly unexpected outcome when not everything is a number.  If no numbers before text, then you 0 and everything that follows all mashed together. If there are numbers, then text, then more numbers/text, you get the sum of the first group of numbers, then everything mashed together.  I have no problem with this.  If you are adding numbers and non-numbers, you can't expect valid results.  This was just not what I was expecting.  Works as it should on numerical arrays.

I added a button to the appropriate tutorial and will update the wiki when merged.

96LawDawg (approved):
Approved through test. 

Things get even stranger if summing an array of arrays.  So just don't do anything strange.  And after looking at "reduce" I'm no longer surprised by the result of summing non-numbers.

PR #2276, opened 2024-09-01
added `VAR` function for setting raw JSON values to variables

ArnoldSmith86:
```JSON
{
  "func": "VAR",
  "variables": {
    "myIntegerVar": 4,
    "stringTemplate": "I have ${COUNT} cards.",
    "myCustomObject": {
      "key1": "value1"
    }
  }
}
```

Usually `var x = 4` would be shorter and easier but with this you can use variables inside strings, you can use any Unicode inside strings and you can use any array or object literals.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2276/pr-test (or any other room on that server)

96LawDawg:
So we've come full circle and have "func" as a function again?  Ha!

96LawDawg (approved):
This is such a small amount of code for such a powerful little feature.  I tested strings, objects, arrays, and PROPERTY OF variables.  They all worked as expected.  Approved through test.  I will update the wiki when merged.

PR #2284, opened 2024-09-04
Public Library: Add the game Alakazap!

andypymont:


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2284/pr-test (or any other room on that server)

RaphaelAlvez:
Maybe change the name as there is one game in PCIO that is called "Zap: space mutiny" (similar to bang).

Might create confusion.

andypymont:
> Maybe change the name as there is one game in PCIO that is called "Zap: space mutiny" (similar to BANG!).

I don't really imagine anyone is going to be confused by a game on a different platform, especially when the public library entry here on VTT has a distinct name within this context, and the tile on the game shelf very clearly says which game it is similar to.

"Zap" is also a good, easy word to use in the gameplay here (it's a part of the game that you need to be able to say something when you're throwing cards at the player who just played) and simplifies the rules description through being able to use the term "zapped".

andypymont:
> Might create confusion.

I've renamed it _Alakazap!_ - I assume that works OK? (Might be easier to say on voice chat too as background noise detection won't clip out the short word.)

RaphaelAlvez:
> > Might create confusion.
> 
> I've renamed it _Alakazap!_ - I assume that works OK? (Might be easier to say on voice chat too as background noise detection won't clip out the short word.)


Thanks. I love the new name. Sounds fun

Even though it is in another platform we do refer to it sometimes as it is the only "similar to" game there.

ArnoldSmith86 (approved):

PR #2288, opened 2024-09-05
ensure that the game is rendered in light mode

### LINKED ISSUE #2278: html card objects have white background until the room reloads (then it's transparent)
Upon further investigation it is edit mode (white) vs play mode (transparent).

This causes it:

```CSS
body.edit.darkMode {
  color-scheme: dark;
}
```

No issue in Chrome.

ArnoldSmith86:
The browser might make decent assumptions about how to render overlays in dark mode but the game usually just looks horrible.

We should still properly support dark mode in the rest of the UI but that is a bit of work.

This fixes #2278 because the iframe gets an opaque background if it has a different color-scheme setting than the parent document.

This also fixes a crash when the `value` of the html card face object is not a string.


96LawDawg (approved):
Approved through test.  This fixes the `value` bug. I could not reproduce the light/dark mode issue on Firefox or Chromium browsers.  Probably just doing it wrong.  But the code that fixes that is benign and should do what it says-force the game itself to be in light mode.

PR #2290, opened 2024-09-05
added Codespaces support

ArnoldSmith86:
First test. Goal is that VTT starts automatically when opening a Codespace.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,15 @@
+{
+  "name": "VirtualTabletop.io",
+  "forwardPorts": [8272],
+  "portsAttributes": {
+    "8272": {
+      "label": "VirtualTabletop.io",
+      "onAutoForward": "notify"
+    }
+  },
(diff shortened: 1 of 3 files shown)


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2290/pr-test (or any other room on that server)

ArnoldSmith86:
It seems to work pretty well. It takes a while. And for some reason it takes quite a while longer until the websocket connection actually works..

RaphaelAlvez:
Works for me. Much better then doing by myself

ArnoldSmith86:
I think this is all I wanna do for now. I couldn't get the debugger to attach automatically but you can just press the Play arrow in "Run and Debug" and it should attach to the automatically running process. Breakpoints should then just work.

It also doesn't make the port public. Maybe it's a security feature, no idea.

96LawDawg (approved):
Blind approval.  I don't even know what Codespaces is, but it doesn't look like this can do any harm.

PR #2292, opened 2024-09-06
fixed Gitpod

ArnoldSmith86:
Gitpod also needed to edit `config.json` so that `externalURL` works.

And they changed their editor to VSCode. This should fix the debugger.


ArnoldSmith86:
PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2292/pr-test (or any other room on that server)

96LawDawg (approved):
Blind approval.  I don't use Gitpod and don't want to figure it out right now.

PR #2294, opened 2024-09-06
PL Property Tycoon: Fix seat 6 bid buttons

andypymont:
---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2294/pr-test (or any other room on that server)


96LawDawg:
Abandoned for #2295 that does the same thing, but without deleting the last updated date in the metadata.

PR #2298, opened 2024-09-07
fixed no-repeat typo in hex preset

ArnoldSmith86:
It doesn't matter for buttons because they have it set anyway.

But it shouldn't be unnecessary _and_ invalid. :grin:

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2298/pr-test (or any other room on that server)


96LawDawg (approved):
Approved through code review

PR #2304, opened 2024-09-12
inherit properties again when overriding ones are removed

### LINKED ISSUE #2117: inheritFrom does not immediately update some properties
To see this issue, make a couple of widgets.  Inherit something on the second widget that is not already specified (like activeFace 1 on the main widget and activeFace 0--default gray--on the second widget).  The change is instant; it happens as soon as you finish typing.  But try to change x (or another property that already is not gray on the second widget) and the change does not take place immediately.  Of course, you have to delete the x or whatever on the inheriting widget. But when you do the deletion, the change still does not take place.  You have to change the inheritFrom property again (change anything about it like delete the x and put it back, or add a space and then delete it).

ArnoldSmith86:
Fixes #2117. It was getting annoying while working on #2258.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2304/pr-test (or any other room on that server)

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -263,6 +263,15 @@ export class Widget extends StateManaged {
       inheriting.applyInheritedDeltaToDOM(inheritedDelta);
     }
 
+    // inherit properties again when overriding ones are removed
+    if(this.state.inheritFrom !== undefined) {
+      for(const key in delta)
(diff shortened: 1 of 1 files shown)


96LawDawg (approved):
Approved through test.  Fixes the problem.  Does not seem to break anything in existing games or in game development.

PR #2311, opened 2024-09-14
fixed a few crashes when properties have unexpected types

ArnoldSmith86:
While testing #1538, we noticed that using an actual number as an `id` leads to crashes.

I searched for occurences of `widget.get(property).function` in the code to see where we are having similar issues. Because `widget.get(property)` can basically always return something that does not know `function`.

This PR fixes a couple of obvious places but there are probably still plenty more. It is already hard enough to review and test the scope as it is.

There are still plenty of crashes for widgets with actual numbers as their `id`. But now I know that the only real solution is to enforce strings for `id`. I will do that in a separate PR though.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2311/pr-test (or any other room on that server)


ArnoldSmith86:
Oops, wrong base for branch.

PR #2312, opened 2024-09-14
fixed a few crashes when properties have unexpected types

ArnoldSmith86:
While testing #1538, we noticed that using an actual number as an `id` leads to crashes.

I searched for occurences of `widget.get(property).function` in the code to see where we are having similar issues. Because `widget.get(property)` can basically always return something that does not know `function`.

This PR fixes a couple of obvious places but there are probably still plenty more. It is already hard enough to review and test the scope as it is.

There are still plenty of crashes for widgets with actual numbers as their `id`. But now I know that the only real solution is to enforce strings for `id`. I will do that in a separate PR though.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2312/pr-test (or any other room on that server)

### DIFF
--- a/client/js/editor/sidebar/properties.js
+++ b/client/js/editor/sidebar/properties.js
@@ -1225,7 +1225,7 @@ class PropertiesModule extends SidebarModule {
 
       const cardClone = new Card();
       const newState = {...card.state};
-      newState.activeFace = deck.get('faceTemplates').length>1?1:0;
+      newState.activeFace = card.getFaceCount()>1?1:0;
       newState.cardType = cardType;
       cardClone.renderReadonlyCopyRaw(newState, $('.renderedWidget', cardTypeDiv));
 
@@ -1357,6 +1357,8 @@ class PropertiesModule extends SidebarModule {
 
   renderForDice(widget) {
     this.addHeader(`Dice ${widget.id}`);
+    const widgetFaces = widget.get('faces');
+    const faceCount = Array.isArray(widgetFaces) ? widgetFaces.length : 0;
 
     this.addSubHeader('Dice types');
     const faces = [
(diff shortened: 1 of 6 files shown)


96LawDawg:
I don't know how to test this.

ArnoldSmith86:
- [ ] Do widgets still snap to `grid`?
- [ ] Do counters with decimal places still fix their rounding errors?
- [ ] Does the `CANVAS` operation (especially with an empty `colorMap` array) still work?
- [ ] Does `owner` still properly hide widgets from other players?
- [ ] Does `handlePosition` of piles still work?
- [ ] Do cards still get created honoring the `faceTemplates`?
- [ ] Does the tree still display `cardType` or `text` for widgets with auto-generated `id`? This probably doesn't even work in production anymore because I don't think anyone adds the class `wide` to it ever.
- [ ] Does the editor "Properties" UI for dice still work?
- [ ] Do the `dropTarget` buttons still render correctly for the "Properties" UI of holders?

It's more likely that they quickly cause Javascript errors instead of showing subtle behavior changes.

You could, of course, also test wrong types for all of those properties. But it's not really important as long as we don't break the correct things..

96LawDawg:
> Does the CANVAS operation (especially with an empty colorMap array) still work?

It works the same as it does on production, which is to say that it does not display anything because it is white on white.  But if you draw and then manually change the color map to just black, it shows up.

> Does the tree still display cardType or text for widgets with auto-generated id? This probably doesn't even work in production anymore because I don't think anyone adds the class wide to it ever.

I don't remember ever seeing this.  I definitely don't see it on production or on the test server.

ArnoldSmith86:
> I don't remember ever seeing this. I definitely don't see it on production or on the test server.

Before the new editor layout, you could toggle the sidebar between narrow and wide and the wide mode showed those details.

ArnoldSmith86:
Sounds like everything is working then? Apart from #2314 which is unrelated.

96LawDawg (approved):
Toggle-wide works, so everything on the list works.  Approved.

PR #2313, opened 2024-09-14
enforce `id` to be a string

ArnoldSmith86:
It currently just crashes if it is anything else and it really is not worth it to allow it.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2313/pr-test (or any other room on that server)

### DIFF
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -2080,6 +2080,8 @@ function jeGetContext() {
 
     if(!jeStateNow.id)
       jeJSONerror = 'No ID given.';
+    else if(typeof jeStateNow.id != 'string')
+      jeJSONerror = 'ID has to be a string.';
     else if(JSON.parse(jeStateBefore).id != jeStateNow.id && widgets.has(jeStateNow.id))
       jeJSONerror = `ID ${jeStateNow.id} is already in use.`;
     else if(jeStateNow.parent !== undefined && jeStateNow.parent !== null && !widgets.has(jeStateNow.parent))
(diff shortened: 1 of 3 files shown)


96LawDawg (approved):
Can't do a proper test on this because the current production server does not allow a routine to SET an id.  But approved through code review.

ArnoldSmith86:
Wanna do some more tests now that #1538 is merged?

96LawDawg (changes requested):
There are a few unexpected things.

If you directly SET an `id` to "null," you get a standard 4-character random id.  But if you SET it to a variable that evaluates to "null," nothing happens.  That seems strange.  I would think in both cases, nothing should change.

If you SET an `id` to the value of an array (or even an array of arrays), the array gets stringified. If you SET an `id` to an object, the `id` becomes "object Object".  And then if you SET another `id` to another object, really strange things happen (because both have the same `id`, but not really the same??).  But if you SET the `id` to the key of an object then the value of that key works. Maybe we should actively ignore anything that is an object?  And we should probably ignore anything that is an array because almost certainly that was not what the designer intended, but I could go either way on that.

Also, if you SET an `id` to an empty array, the room freezes.  An empty object still gives you "object Object."

96LawDawg (approved):
Approved through test

96LawDawg:
> `SET` should now refuse to use anything that is not a string as an `id`. Does that address all your findings?

Yes.  That is the best solution.

PR #2318, opened 2024-09-20
categories for game-icons.net in the symbol picker

ArnoldSmith86:
I did not get an answer to https://github.com/game-icons/icons/issues/954 and now I started to use ChatGPT to split the icons into categories strictly by looking at their name. This resulted in a new repository where I keep the scripts and history: https://github.com/ArnoldSmith86/gameicons-metadata

This PR makes the icon picker a lot more useful for game-icons.net icons by splitting them into (not very good) categories and adding a few tags (for example making the icon "hound" appear when searching for "dog"):

image

This is mainly just a change to `assets/fonts/symbols.json` but it contains two code changes:

1. The preview of the icons is rendered using [a big PNG containing all icons](https://github.com/ArnoldSmith86/virtualtabletop/blob/main/assets/game-icons.net/overview24.png). To get the correct location in that PNG, the code relied on the icons being in the correct order in `assets/fonts/symbols.json`. That is no longer the case so now the index is the first entry in the keywords array.
2. Basically unrelated to everything else, it no longer searches the author name. So searching for "dog" no longer includes the icon "heavenly-dog/defensive-wall".

All of this is clearly far from perfect but IMO still a substantial improvement over the status quo. Feel free to follow the instructions on https://github.com/ArnoldSmith86/gameicons-metadata?tab=readme-ov-file#tool to improve the categories.

---

I'm gonna leave this here for any future PRs:

```
{
    sed -n '0,/Game-icons/p' ../virtualtabletop/assets/fonts/symbols.json | head -n -1
    tail -n +2 symbols.json | head -n -2
    echo '  },'
    sed -n '/Material/,$p' ../virtualtabletop/assets/fonts/symbols.json
} > new.json
mv new.json ../virtualtabletop/assets/fonts/symbols.json
rm symbols.json
```

After calling https://github.com/ArnoldSmith86/gameicons-metadata/blob/master/generate-symbols-json.py, it can be used to update `assets/fonts/symbols.json`.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2318/pr-test (or any other room on that server)


RaphaelAlvez:
> It takes a while because ChatGPT can only process a few hundred at a time and there is a usage limit.

Feels like something we could all help with if you share the how to do it

ArnoldSmith86:
> Feels like something we could all help with if you share the how to do it

I was writing... But please do..

ArnoldSmith86:
Now there are "only" 55 categories.

ArnoldSmith86:
image

An approach like that looks promising for further improvements.

ArnoldSmith86:
I updated the PR description now that the tools are in their own repository.

IMO, this is ready to be reviewed and merged. None of the data is perfect but this is a nice first step. And I wanna do other stuff now. :grin:

96LawDawg (approved):
Approved through test and code review.  This will definitely make finding symbols easier.  Too bad the devs at game-icons.net never responded to your request.

PR #2324, opened 2024-09-27
added monochrome version of Noto Emoji

ArnoldSmith86:
While working on https://github.com/ArnoldSmith86/virtualtabletop/pull/1937, I thought again that it would be nice to have the Unicode emoji set in a monochrome version which would fit better when used with game-icons.net or Material icons (and you could also change the color).

In the symbol picker they look like this:

image

When using `"classes": "emoji-monochrome"` on a button, they can look like this:

image

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2324/pr-test (or any other room on that server)


ArnoldSmith86:
I think this is ready for review now. You can use the monochrome emojis on buttons, basic widgets and in the game metadata help texts ("How to play the game" etc.).

It will, of course, be even cooler when integrated into #1937.

96LawDawg (approved):
Approved through test.

PR #2325, opened 2024-09-30
fixed crash when opening deck properties

ArnoldSmith86:
The widgets in the visualizations in the Properties sidebar are not quite complete widgets. Apparently they need one hack more now. :)

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2325/pr-test (or any other room on that server)


96LawDawg (approved):
Approved through test.

Issue #1559, opened 2023-01-23
`AUDIO` does not support array of player names


96LawDawg:
Currently, audio only plays the sound to everyone or to the player that activates the routine. The function would be improved by allowing an array of player names in the `player` parameter.

bjalder26:
I'm pretty sure this is all it would take.

      const pName = asArray(audioArray[9]);

      if(pName == "null" || pName.includes(playerName)) {

PR #2328, opened 2024-10-02
Audio to AudioContext

bjalder26:
This is a rough draft.  The idea is to update the audio function to use AudioContext.  The intention is to make it so sound will work on ipads (ios).  This rough draft is only being uploaded so it can be tested.

**No sound will be activated until the user interacts with the screen**.  This will become true for all operating systems and browsers.  However, this will probably be barely noticed in actual game play, and I suspect this is the direction all browsers are going in, so we may not have any other option in the future.

Once there is an interaction, sound SHOULD work on all computer/tablets/browsers - however this needs to be tested.

I believe the type of audio will now become obsolete, but before it is completely deleted from the code, the "audio" property that is used will need to be converted from an array to an object (probably what I should have done in the first place).  Once that is changed it should be removed from the JSON editor as well.

I also believe I added the ability for the playerName to be an array (#1559) - though this also needs to be tested.

#1558 should probably be merged into this, because this will likely remove the issue mentioned in that PR.

### DIFF
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -429,6 +429,106 @@ async function toggleEditMode() {
   setScale();
 }
 
+// Browser support for AudioContext
+window.AudioContext = window.AudioContext || window.webkitAudioContext;
+
+let context; // Declare context outside the function
+
+// Function to initialize AudioContext
+function initializeAudioContext() {
+  if (!context) { // Ensure it's initialized only once
+    context = new AudioContext();
+    console.log('AudioContext initialized.');
+    audioContext(); // Call your audio context setup function here
+  }
+}
+
+// List of events to listen for user interaction
+const events = ['click', 'mousedown', 'mouseup', 'dblclick', 'keydown', 'keyup', 'touchstart', 'touchend', 'focus', 'pointerdown'];
+
+// Add event listeners to the document
+events.forEach(event => {
+  console.log(event);
+  document.addEventListener(event, initializeAudioContext, { once: true });
+});
+
+let audioBufferObj = {};
+let sources = {} // Object to track all sources
+
+// Function to check for routines with AUDIO function and clickSound in both state and cardDefaults
+function findAudioSources(widget) {
+  // Use a Set to ensure unique audio sources
+  let audioSources = new Set();
+
+  // Check for any routines with AUDIO function in widget.state
+  Object.keys(widget.state).forEach((key) => {
+    if (key.endsWith("Routine")) {
+      const routine = widget.state[key];
+
+      // Check if the routine is an array and has objects with "func": "AUDIO"
+      routine.forEach((action) => {
+        if (action.func === "AUDIO") {
+          // Add the source to the Set (duplicates will be ignored)
+          audioSources.add(action.source);
+        }
+      });
+    }
+  });
+
+  // Check for clickSound in widget.state and add it if valid
+  if (widget.state.clickSound !== null && widget.state.clickSound !== undefined) {
+    audioSources.add(widget.state.clickSound);
+  }
+
+  // Check for clickSound in cardDefaults (if it exists) and add it if valid
+  if (widget.state.cardDefaults && widget.state.cardDefaults.clickSound !== null && widget.state.cardDefaults.clickSound !== undefined) {
+    audioSources.add(widget.state.cardDefaults.clickSound);
+  }
+
+  return Array.from(audioSources); // Convert Set to an array before returning
+}
+
+function getAudioArray() {
+  // Use a Set to collect unique audio sources across all widgets
+  let audioSet = new Set();
+
+  widgets.forEach((widget) => {
+    // Add found audio sources from each widget to the Set
+    findAudioSources(widget).forEach((source) => audioSet.add(source));
+  });
+
+  return Array.from(audioSet); // Convert Set to an array before returning
+}
+
+// Function to load audio asynchronously
+async function loadAudio(url) {
+   try {
+    const response = await fetch(mapAssetURLs(url)); // trying mapAssetURLs
+    const arrayBuffer = await response.arrayBuffer();
+    const buffer = await context.decodeAudioData(arrayBuffer);
+    audioBufferObj[url] = buffer;
+  } catch (error) {
+    console.error(`Error loading audio from ${url}:`, error);
+  }
+}
+
+// Main function to process audio sources and load them
+function audioContext() {
+  const audioArray = getAudioArray();
+
+  if (audioArray.length > 0) {
+    // Iterate through each audio source URL and load them
+    audioArray.forEach((url) => {
+      if (url) {
+        loadAudio(url);
+      }
+    });
+  }
+  console.log(audioArray); // Debugging: check the audio sources found
+  console.log(audioBufferObj);
+  console.log(context);
+}
+
 onLoad(function() {
   on('#pileOverlay', 'click', e=>e.target.id=='pileOverlay'&&showOverlay());
 


bjalder26:
Test server: https://github.com/ArnoldSmith86/virtualtabletop/pull/2328

ArnoldSmith86:
> Test server: https://test.virtualtabletop.io/PR-2328/pr-test

Fixed the bot. :grinning:

client/js/main.js
```diff
@@ -429,6 +429,99 @@ async function toggleEditMode() {
...
+      const routine = widget.state[key];
+      routine.forEach((action) => {
+        if (action.func === "AUDIO") {
+          if(action.source) {
+            audioSources.add(action.source); 
+          }
+        }
+      });
+    }
+  });
```
ArnoldSmith86, line 467: This construct doesn't go through subroutines like `thenRoutine` or `loopRoutine`.

client/js/widgets/widget.js
```diff
@@ -457,8 +457,14 @@ export class Widget extends StateManaged {
     if(!this.get('clickable') && !(mode == 'ignoreClickable' || mode =='ignoreAll'))
       return true;
 
-    if(this.get('clickSound'))
-      this.set('audio','source: ' + this.get('clickSound') + ', type: audio/mpeg , maxVolume: 1.0, length: null, player: null');
+    if(this.get('clickSound')) {
+      this.set('audio', JSON.stringify({
```
ArnoldSmith86, line 461: Every `set` needs to be `await set`. Apparently it was also missing in the old code. :grin:

ArnoldSmith86 (commented):
I realize that you're still working on this but I just stumbled into the code for some reason..

I left a few comments in the code. If you need help with any of that, let me know.

Also: this seems to be a complete rewrite of all the audio code. Seems like a good opportunity to move all the code into its own `audio.js` (gotta include that into the array in `server/minify.mjs`).

client/js/widgets/widget.js
bjalder26: I REALLY appreciate the feedback.  Sadly I need to leave right now, otherwise I would jump into this.

ArnoldSmith86:
Maybe we also get rid of the `audio` property and replace it by a separate websocket broadcast.

ArnoldSmith86:
And are you sure preloading all the audio files is worth the trouble? That's most of your code and it doesn't handle a lot of edge cases yet.

When a browser loads an asset once, it should never have to load it again (I should test that a bit better).

If the player has a slow connection, preloading probably makes the sounds play sooner _but_ especially with a slow connection, you probably don't want to load "unnecessary" stuff at the start of the game.

I'm not sure what the right answer is. For image assets, we don't do any preloading. Which means that it can take a moment for a card face to appear when flipping a card but it makes the browser download a lot less stuff initially.

client/js/main.js
```diff
@@ -429,6 +429,99 @@ async function toggleEditMode() {
   setScale();
 }
 
+// Browser support for AudioContext
+window.AudioContext = window.AudioContext || window.webkitAudioContext;
```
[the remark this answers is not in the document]
bjalder26: Agreed.

client/js/main.js
```diff
@@ -429,6 +429,99 @@ async function toggleEditMode() {
...
+export let context;
```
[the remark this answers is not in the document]
bjalder26: Yes, that is a good idea.

client/js/main.js
```diff
@@ -429,6 +429,99 @@ async function toggleEditMode() {
...
+export let audioBufferObj = {};
+export let sources = {};
+
+const events = ['click', 'mousedown', 'mouseup', 'dblclick', 'keydown', 'keyup', 'touchstart', 'touchend', 'focus', 'pointerdown'];
```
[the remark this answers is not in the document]
bjalder26: Yeah, maybe it should be limited to mousedown, keydown, and touchstart.  I just threw in everything that could count as a user interaction.

Without click the automated click exception could be eliminated as well.

client/js/main.js
```diff
@@ -429,6 +429,99 @@ async function toggleEditMode() {
...
+// Initialize AudioContext after user event
+function initializeAudioContext() {
+  if (!context && event.target.id != 'activeGameButton') {
+    context = new AudioContext();
+    audioContext(); 
+  }
+}
+
+// Function to check for routines with AUDIO function and clickSound in both state and cardDefaults
+function findAudioSources(widget) {
```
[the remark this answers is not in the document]
bjalder26: Yes, it is only called once - after the page loads and the user interacts with the screen.

client/js/main.js
bjalder26: Yep, that will need to be fixed if we're keeping preloading audio.

bjalder26:
> Maybe we also get rid of the `audio` property and replace it by a separate websocket broadcast.
> 
> Idea would be to do something like `toServer('audio', { ... })` when you want to play sound. That needs to be handled by `messageReceived` in `server/player.mjs` which directs it to `server/room.mjs` (maybe a new method `playAudio` or something). That one uses `this.broadcast('audio', { ... })` to send it to all players (or it already filters by the playername). Then back on the client you do `onMessage('audio', function() { ... })` to actually play the sound on the other (or all) clients.
> 
> The advantage is that it doesn't "pollute" the game state, and doesn't need any cleanup.

 >And are you sure preloading all the audio files is worth the trouble? That's most of your code and it doesn't handle a lot of edge cases yet.
>
> When a browser loads an asset once, it should never have to load it again (I should test that a bit better).
>
> If the player has a slow connection, preloading probably makes the sounds play sooner but especially with a slow connection, you probably don't want to load "unnecessary" stuff at the start of the game.
>
> I'm not sure what the right answer is. For image assets, we don't do any preloading. Which means that it can take a moment for a card face to appear when flipping a card but it makes the browser download a lot less stuff initially.

A separate websocket broadcast might be a better way to do audio.  I don't know a lot about websocket.  I'm not sure if preloading the audio is worth it.  I have noticed some latency at times, and thought doing this would optimize the sound.  However, I think the way it was done previously sounds had to be loaded each time they were used, whereas this way I think they are only loaded once, and maybe the first time a sound is used is more appropriate than all of them being loaded the first time someone interacts with the screen.

If I were to make a guess, I would guess a separate websocket is worthwhile.  I assume there would still be a way to control which player(s) hear the sound.  I do not love the audio property, that was just how I could figure out how to do it at the time.  

As far as preloading the audio files, I don't know, and I have thought of deleting that part.  Right now I am really just jazzed that there is a way to make audio work on ipads.  Once the audio is loaded it should not need to be loaded again.  I think the difference in gameplay would be a slight delay after the first user interaction, or a slight delay after the first time a sound is triggered.  However, thinking about maintaining the code simple is best, and maybe that outweighs the other small differences.

I think what I should do right now is update this according to your feedback, make it so all audio is loaded the fist time it is used, then see if I can figure out how to make a separate websocket broadcast.  If we later decide that preloading is needed, that could always be added back.

server/room.mjs
```diff
@@ -216,6 +220,8 @@ export default class Room {
   broadcast(func, args, exceptPlayer) {
     if(func != 'mouse')
       this.trace('broadcast', { func, args, exceptPlayer: exceptPlayer?.name });
+    if(args.playerArray && args.playerArray !== [])
+      this.players = args.playerArray;
```
ArnoldSmith86, line 224: This doesn't work. `this.players` is an array of player objects that are currently in the room. You can't just change that.

Looks like `broadcast` might not be what you want. Maybe loop through the players in the audio function and test their name against your array? Or do the filtering on the receiving client side? Or change `broadcast` to optionally get an array of recipient names?

Unrelated to that problem, I don't think you can do `variable !== []`. That compares the reference so it would have to be the exact same array (which it can't because `[]` is a literal so nothing less can be that). Better to do something like `!variable.length`.

bjalder26:
```js
  broadcast(func, args, exceptPlayer) {
    if(func != 'mouse')
      this.trace('broadcast', { func, args, exceptPlayer: exceptPlayer?.name });
    let argsObj = {};
    for(const player of this.players) {
      let includePlayer = true;
      if(func === 'audio') {
        argsObj = JSON.parse(args);
        if(argsObj.players && !argsObj.players.includes(player.name))
          includePlayer = false;
      }
      if(player != exceptPlayer && includePlayer)
        player.send(func, args);
    }
  }
```
Also, would it be better to use two `if(func === 'audio') {}` and parse only once (outside of the `for`)?

96LawDawg:
Tested on PC.  Works.  Tested on iOS and mac OS.  It works.  This is a big development in sound getting it to work on Apple devices.

96LawDawg (approved):
Approved through test

bjalder26:
> Approved through test

I need someone to review `minify.mjs` to make sure I put `audio.js` in an okay spot, to review `player.mjs` to make sure I put `this.room.playAudio(args)` in an okay spot, and I especially need someone to review the `broadcast` function in `room.mjs` to make sure I didn't do anything wrong with that critical function.

ArnoldSmith86:
I will. Soonish.

ArnoldSmith86 (approved):
- The order in the server files doesn't really matter but they are roughly alphabetical so I changed it a bit. Order in `minify.mjs` doesn't matter either except for when code is used - for example your code calls the `on` method immediately when loaded so that one must be included first (which it was).
- `playAudio` no longer uses `broadcast`. Keeps `broadcast` simpler and `playAudio` isn't really that complicated either.
- Your `onMessage('audio', ...)` wasn't `async` and was calling `addAudio` without an `await` which would have made the `try ... catch` never catch anything from `addAudio`.

I did not test if the actual audio is still working after my changes. Please do so. But I'm pretty confident that this shouldn't break anything else.

bjalder26:
> I did not test if the actual audio is still working after my changes. Please do so. But I'm pretty confident that this shouldn't break anything else.

"player" (excluding players not listed in an array or string) doesn't work - it seems to always send the audio.  I'm not 100% sure it worked right before, because I can't remember when I tested what, but I suppose this line will need to be adjusted or expanded.

`if(!Array.isArray(args.players) || args.players.includes(player.name))`

bjalder26:
I think the issue is that I'm stringify-ing the args.  That's probably completely unnecessary and I probably just need to get rid of the stringify.
```js
toServer('audio', JSON.stringify({
        audioSource: this.get('clickSound'),
        maxVolume: 1.0,
        length: null,
        players: []
      }));

toServer('audio', JSON.stringify({
          audioSource: a.source,
          maxVolume: a.maxVolume,
          length: a.length,
          players: validPlayers
        }));
```

96LawDawg (approved):
Approved through test.  Works on all PC, iOS, and macOS.

PR #2329, opened 2024-10-02
notify users about client crashes and collect information

ArnoldSmith86:
Before this PR, an uncaught exception in client-side Javascript simply triggered a page reload. Except when it was a promise rejection. In that case the user did not notice it at all but in the background all kind of stuff broke.

This PR tries to listen for the `unhandledrejection` event which hopefully catches most errors. In either case, it then shows a generic error message to the user, asks for steps to reproduce in a simple textbox and then sends all kinds of (hopefully) useful data to the server why I can look at it: error message, stack trace, the content of that textbox, the current room state, the undo protocol, the whole HTML structure, the text in the JSON editor, etc..

I intentionally did not update this branch with the latest bugfix yet. So you can trigger an error by looking at a deck with the Properties editor.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2329/pr-test (or any other room on that server)


96LawDawg (approved):
Approved through test.  No idea what the security warnings are about. Above my paygrade.

ArnoldSmith86:
Security warnings from a browser or the comment on GitHub?

96LawDawg:
The GitHub stuff.

PR #2332, opened 2024-10-05
fixed server crash for broken room

ArnoldSmith86:
The server crashed yesterday. Apparently a bot tried to load a room that has a broken JSON file.

I added exception handling to that code path. I am pretty sure it is the only handler that is affected by that problem.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2332/pr-test (or any other room on that server)

### DIFF
--- a/server.mjs
+++ b/server.mjs
@@ -324,54 +324,58 @@ MinifyHTML().then(function(result) {
   router.get('/game/:shareID/:name', gameRoomHandler);
   router.get('/tutorial/:plName', gameRoomHandler);
   async function gameRoomHandler(req, res, next) {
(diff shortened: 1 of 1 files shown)


96LawDawg (approved):
Blind approval.  Not sure how to get my browser "to identify as a bot" so I couldn't test this PR.  I talked nicely to my browser, I yelled at it, I cried, I bribed it, I threatened to turn it off and on (that almost worked), and I said nasty things about its ancestors.  It said nasty things about my ancestors and insists it is not a bot.

bjalder26:
It looks like the only change is wrapping everything in a try/catch block for error handling so the server won't crash and give no error message.

PR #2337, opened 2024-10-08
fixed cursor check client crash on touch devices

ArnoldSmith86:
#2113 used a wrong variable that is not defined for touch events. Apparently this problem does not surface by breaking anything else. #2329 made it visible. This should fix it.

Hopefully the production server can get back to the latest version in `main` with this.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2337/pr-test (or any other room on that server)


96LawDawg (approved):

PR #2340, opened 2024-10-14
fixed crash when dragging widget off screen

ArnoldSmith86:
When in edit mode in Firefox, the client crashes when you try to drag a widget to outside of the browser window. This PR should fix that.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2340/pr-test (or any other room on that server)

### DIFF
--- a/client/js/editor/selection.js
+++ b/client/js/editor/selection.js
@@ -20,7 +20,7 @@ export function editInputHandler(name, e) {
       }, 0);
     }
   }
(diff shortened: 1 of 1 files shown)


96LawDawg (approved):
Approved through test.  Interesting that the problem only happened if the toggle button was not selected.  I initially could not reproduce the problem on production, but looked at the code then decided to try it that way (backwards from what I normally use).

PR #2344, opened 2024-10-17
fixed editor crash for non-object `dragLimit` and `grid`

ArnoldSmith86:
The test for showing corresponding buttons assumed that the properties contain an object.


96LawDawg (approved):
Approved through test.

Issue #2351, opened 2024-10-29
Changing playerName broken?


bjalder26:
I came across this while looking at #1631.

https://virtualtabletop.io/wrrq

I think changing playerName might have gotten broken when there was an update to player cursors.
![image](https://github.com/user-attachments/assets/8ad116e9-13c4-4191-9d10-b39da2ea0684)
It looks like this was this update:
![image](https://github.com/user-attachments/assets/3a5d434f-4892-4d88-9981-ac182f864f09)
https://github.com/ArnoldSmith86/virtualtabletop/pull/1829/files

Maybe this can be fixed with an `if(playerCursors[e.player])` but @ArnoldSmith86 should probably look at it.

ArnoldSmith86:
How did you trigger the error? Pressing the button in your room doesn't seem to cause issues for me in Firefox or Chrome.

bjalder26:
> How did you trigger the error? Pressing the button in your room doesn't seem to cause issues for me in Firefox or Chrome.

I think the new name has to include a non-letter character (e.g. numbers, !, etc.).

I have no idea why that would be the case.

bjalder26:
On second thought, I think it just has to not be the same as the current name.  Also, it won't crash if it had already crashed and you don't refresh the browser.

It looks like it is looking for the previous name, and I'm guessing that it has already been removed from playerCursors.

I updated the room, and it should trigger it every time (or after 2 clicks) because it now cycles through names.

PR #2353, opened 2024-10-31
updated testcafe on github

ArnoldSmith86:
This _might_ fix things.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2353/pr-test (or any other room on that server)


bjalder26:
This just needs a review right?  I've read how this is what is suggested as a solution to the error message we are getting, and it's just an update to testcafe, so I don't see any reason to not approve it.

bjalder26 (approved):
I've read that this update is supposed to address the error message we are getting from testcafe.

96LawDawg:
Yeah, I didn't approve this one because it didn't work.  I wonder if the player name thing has anything to do with #2351?

bjalder26:
> As you can see, the TestCafe Chrome test still fails on this one though. It is a different problem though and possibly because resizing the window doesn't work which makes editor stuff open in an overlay instead of the sidebar, obscuring the game content the tests want to click on.
> 
> Hmm, I just looked at the output again and it seems to not find the color picker for the player color. I'll do some tests soon.

Could that be the same as this issue?
https://github.com/DevExpress/testcafe/issues/8237

It looks like they resolved that issue with 3.7.0 (3.7.0-rc.4).
https://github.com/DevExpress/testcafe/commit/716aa9194a2b9113f10dd398a4c303664ea8758f#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR32

PR #2359, opened 2024-11-09
Fix star number location

96LawDawg:
Fixes #2355

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2359/pr-test (or any other room on that server)

PR #2368, opened 2024-11-25
fixed star number

### LINKED ISSUE #2365: Position of stars in starred games on Create Room overlay
So we just fixed the position of the number of starred games on the main VTT screens.  But it is still out of place when you invite use the shared link feature.  For example, linking someone to Uncheckered through https://virtualtabletop.io/game/uncheckered gives this:

Screenshot 2024-11-20 180824

The problem is the 8 (I think it is an 8) is over the first C in Uncheckered.  Probably the same fix as #2359 is needed, but in a slightly different location in the css.

ArnoldSmith86:
Fixes #2365. Reverts #2359. Caused by #2324.

The emoji replacement code was updated and caused another span to be inserted which made the CSS rules not match anymore.

This PR now only inserts a span when there actually were emojis. The star count should never contain any.

---

The last commit was intended for a new PR but oh well. It fixes that emojis in the game name would not be replaced by images in the share game overlay.

This also fixes that you could insert arbitrary HTML by using emojis in non-HTML game metadata.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2368/pr-test (or any other room on that server)

### DIFF
--- a/client/css/overlays/states.css
+++ b/client/css/overlays/states.css
@@ -502,7 +502,7 @@ div:nth-of-type(2) > .list > .roomState .star {
 #mainImage span {
   position: absolute;
   font-size: 16px;
-  top: 17px;
+  top: 51px;
   width: 62px;
   text-align: center;
   color: white;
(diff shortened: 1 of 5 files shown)


96LawDawg (approved):
Tested the star change.  Works.  Did not test the other changes.

PR #2369, opened 2024-11-25
fixed that hotkeys would still press disabled editor toolbar buttons

### LINKED ISSUE #2361: editor hotkeys press their buttons even though they are disabled

ArnoldSmith86:
Fixes #2361.

### DIFF
--- a/client/js/editor/toolbarButton.js
+++ b/client/js/editor/toolbarButton.js
@@ -23,7 +23,7 @@ class ToolbarButton {
   }
 
   onKeyDown(e) {
-    if(e.key == this.hotkey && !this.domElement.disabled)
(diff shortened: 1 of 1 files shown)


bjalder26 (approved):
It looks like instead of checking to see if the div with a class of editorToolbarButton is disabled (which will never be the case), this PR now checks to see if its child button has been disabled (which is the right thing to check).
```js
<div class="editorToolbarButton">
      <button icon="delete_forever" disabled=""><span>Delete the selected widgets.<br><br>Hotkey: Delete</span>
    </button></div>
```
Also, it works - hotkeys do not trigger disabled buttons, and they do trigger buttons that are not disabled.

Issue #1402, opened 2022-11-12
Setting onPileCreation to type card crashes


flackr:
I made a mistake when setting my `onPileCreation` to `"type": "card"` to find out that it results in a crash when trying to create the pile.

Example game: https://virtualtabletop.io/57gs

To reproduce drag any card on top of another to get the following:
```
57gs:1 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'onChildAdd')
    at Card.onPropertyChange (57gs:1:272096)
    at Card.set (57gs:1:218764)
    at Card.updatePiles (57gs:1:277297)
    at async Card.moveEnd (57gs:1:271304)
    at async inputHandler (57gs:1:211255)
```

The issue is that the creation of the pile widget attempts to use "type": "card" with no "deck" set resulting in it not getting an id. This then tries to set the parent of the piled cards to a null pile which tries to access this pile.

bjalder26:
newParent is undefined for `await newParent.onChildAdd(this, oldValue);`, because `const newParent = widgets.get(newValue);` isn't finding the pile.  In other words, the pile has not yet been added to `widgets`.

`const newParent = widgets.get(newValue);` is evaluating for 'undefined' 

```js
  async onPropertyChange(property, oldValue, newValue) {
    if(property == 'parent') {
      if(oldValue) {
        const oldParent = widgets.get(oldValue);
        await oldParent.onChildRemove(this);
        if(this.get('type') != 'holder' && Array.isArray(oldParent.get('leaveRoutine')))
          await oldParent.evaluateRoutine('leaveRoutine', {}, { child: [ this ] });
      }
      if(newValue) {
        const newParent = widgets.get(newValue);
        await newParent.onChildAdd(this, oldValue);
        if(Array.isArray(newParent.get('enterRoutine')))
          await newParent.evaluateRoutine('enterRoutine', { oldParentID: oldValue === undefined ? null : oldValue }, { child: [ this ] });
      }
      if(!this.disablePileUpdateAfterParentChange)
        await this.updatePiles();
    }
  }
```
It looks like the pileId is coming from `const pileId = await addWidgetLocal(pile);`, but that the pileId is not yet added to widgets.
```js
// if a card gets dropped onto a card, they create a new pile and are added to it
        if(thisType == 'card' && widgetType == 'card') {
          const pile = Object.assign({
            type: 'pile',
            parent: this.get('parent'),
            x: widget.get('x'),
[...]

PR #2371, opened 2024-11-25
added technical details to the client crash overlay

### LINKED ISSUE #2358: Display error log.
Enhance what was done in #2329 so that users can optionally see the log by pressing a button on the error screen in addition to sending a log report.

ArnoldSmith86:
Fixes #2358. You can see the error now if you want to.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2371/pr-test (or any other room on that server)

### DIFF
--- a/client/js/tracing.js
+++ b/client/js/tracing.js
@@ -61,6 +61,7 @@ onLoad(function() {
     preventReconnect();
     connection.close();
     $('#clientErrorOverlay textarea').value = '';
+    $('#clientErrorStack').textContent = details.error;
     showOverlay('clientErrorOverlay');
     $('#clientErrorOverlay button').addEventListener('click', async function() {
       try {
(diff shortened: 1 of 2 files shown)


96LawDawg (approved):
Approved through test and code review.  FYI, it looks like this for the error involving trying to load a room from an invalid URL:

Screenshot 2024-11-25 141147

96LawDawg:
Just as another proof that it works, here is the result from the error in #1402.

Screenshot 2024-11-25 162306

But I'll be honest, I thought there was more to the error log being sent to @ArnoldSmith86 than this.  I don't know what I was expecting so I shouldn't be surprised, but this is what you can see in the console.  Somehow, I just thought it was more.

ArnoldSmith86:
There is a lot more but I'm not sure how useful it is to show there.

There is the whole undo history which is very useful but currently hard to "use". My plan for the future is to turn that undo history into a replay file that you can save and watch at any point. When that is included in the official client, it probably makes sense to provide the replay file along with the stacktrace.

I also get the complete HTML but for the player that's not really useful because it's just a horribly broken version of what they just looked at.

I can look at a few other details like last executed routine operation etc..

PR #2376, opened 2024-11-28
Add player count to toolbar

### LINKED ISSUE #2373: Add player count to menu
A version of this idea was originally suggested by @bjalder26 in #1859.

The main sidebar/bottombar menu has Game Shelf, Active Game, Players, About, and Edit Mode.  It might be a nice addition to add the current active player count near the Players text/icon.  It appears in multiple different configurations depending on the size of the window, but I think a number that reflected the active players in the room would be useful.  And the number could flash or change color briefly as an alert to let people know someone has entered or left.

96LawDawg:
Fixes #2373.

The toolbar Players tooltip now displays the active player count.  If the player count changes or if a player changes their name, then the icon and the tooltip turn yellow for a second and then back to white.  If the window size does not show the tooltip, then the player count is not shown but the yellow flash still happens on the icon.

The effect is subtle.  We could add a small sound to accompany the change.  Lots of apps do that.  But we don't have any other sounds and I don't want this to be annoying.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2376/pr-test (or any other room on that server)

### DIFF
--- a/client/css/layout.css
+++ b/client/css/layout.css
@@ -391,6 +391,11 @@ body.aspectTooGood.hiddenToolbar {
       background: var(--toolbarColor);
       box-shadow: 0 0 5px #0005;
       pointer-events: none;
+      transition: color 0.3s ease;   
+    }
+
+    .tooltip.playerChange {
+      color: yellow;
     }
 
     body:not(.loading) .toolbarButton:hover .tooltip, body.wideToolbar .toolbarButton .tooltip {
@@ -854,6 +859,11 @@ button {
   border-radius: 5px;
   background-repeat: no-repeat;
   margin: 1px;
+  transition: color 0.3s ease
+}
+
+button.playerChange {
+  color: yellow
 }
 
 button:hover {
--- a/client/css/layout.css
+++ b/client/css/layout.css
@@ -394,10 +394,6 @@ body.aspectTooGood.hiddenToolbar {
       transition: color 0.3s ease;   
     }
 
-    .tooltip.playerChange {
-      color: yellow;
-    }
-
     body:not(.loading) .toolbarButton:hover .tooltip, body.wideToolbar .toolbarButton .tooltip {
       display: block;
     }
@@ -862,7 +858,7 @@ button {
   transition: color 0.3s ease
 }
 
-button.playerChange {
+button.playerChange .tooltip.playerChange {
   color: yellow
 }
 
--- a/client/js/overlays/players.js
+++ b/client/js/overlays/players.js
@@ -79,14 +79,10 @@ function updatePlayerCountDisplay() {
   const tooltip = $('.tooltip', playersButton);
   if (tooltip) tooltip.textContent = `Players: ${playerCount}`;
 
-  [playersButton, tooltip].forEach(element => {
-      if (element) element.classList.add('playerChange');
-  });
-
+  [playersButton, tooltip].forEach(element => element.classList.add('playerChange'));
+  
   setTimeout(() => {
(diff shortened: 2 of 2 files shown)


bjalder26:
I think you're going to want to change `document.getElementById('playersButton')` to `$('#playersButton')`.

And I believe `playersButton.querySelector('.tooltip')` to `$('.tooltip')` to condense the code.  However, everything seems to work.

96LawDawg:
The first of your suggested changes is done.  I forget about those shorthand references and ChatGPT doesn't know about them, so ...

The second one doesn't work.  It actually changes the tooltip text "Game Shelf" to "Players: x".

bjalder26:
I think maybe I should have said this:
`var tooltip = $( '.tooltip', playersButton );`

96LawDawg:
>  `var tooltip = $('.tooltip', playersButton);`

That worked.  But I left it const instead of var since it won't change during operation of that function.  But I readily admit, and repeatedly say, I have no idea what I'm doing.

bjalder26:
Sorry to give this piecemeal, and to be so picky, but you could also combine the css rules, unless you have a reason not to - which you might.

button.playerChange, .tooltip.playerChange {
    color: yellow;
}

96LawDawg:
> unless you have a reason not to - which you might.

I do.  See previous entry: I have no idea what I'm doing.  Change made.  Thanks for the suggested changes and it doesn't matter they are piecemeal.  They are ones @ArnoldSmith86 would make, so we're saving him the trouble.

bjalder26 (approved):
This seemed like it had pretty universal approval.  It's a fairly subtle change, that is a useful enhancement.  Other things like adding sound and making the number show up in other views might be considered later when an issue is put in for them, but this PR would make adding those things later easier (if desired).  Right now I think putting it into use and seeing what kind of conversation that stimulates is the right thing to do.

The code is fairly straightforward and I don't see anything that needs to be changed.  It also works as expected.

ArnoldSmith86 (approved):
Code looks good. I would have removed the `if(element)` and `if(tooltip)` but it's not like they hurt..

ArnoldSmith86:
You might want to think about the behavior when multiple clients have the same name. Technically those are different players but the Players overlay shows them as one player.

bjalder26:
> You might want to think about the behavior when multiple clients have the same name. Technically those are different players but the Players overlay shows them as one player.

My thoughts are that game-wise, they will be treated as the same player (same playerName for functions, and own the same widgets), on the players overlay they are treated like the same player, and thus they should be counted as the same player (as it the case in this PR).  Functionally it being like an extension of a player's screen almost like having two monitors - even though it is actually 2 clients.  

I would think the thing to "watch for" would be someone keeping one screen open to the JSON editor, which will be another more complicated PR. 

However, someone please correct me if I am wrong.

96LawDawg:
> Code looks good. I would have removed the `if(element)` and `if(tooltip)` but it's not like they hurt..

Done.

> You might want to think about the behavior when multiple clients have the same name. Technically those are different players but the Players overlay shows them as one player.

I agree with @bjalder26.  If we are doing this just as a nice piece of info that "hey, all 4 of us are here now" then it doesn't matter 2 clients are connected with the same name.  If we are doing it as an anti-cheating measure because a player might have 2 clients open with the same name, I don't think that matters either because it is really difficult to work with both.  And if we are doing it because 2 people might be collaborating to cheat using multiple clients, then we can't stop that anyway.

Plus, I don't know how to detect how many clients have active connections vs how many active players there are.

PR #2377, opened 2024-11-28
Update AUDIO with silence and count parameters

96LawDawg:
This is a partial re-implementation of #1558 and a partial fix to #1333.  It adds a `silence` boolean parameter to AUDIO that will stop all sounds playing.

It also adds a `count` parameter that defaults to 1 that plays the sound n times.  It can also be set to "loop."

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2377/pr-test (or any other room on that server)

---

Proposed wiki edits:

New AUDIO parameters:

count: _number_ - the number of times the sound clip will play. Defaults to 1. "loop" plays an infinite loop. If set to 0, sound will not play. If length is anything other than null, the count parameter is ignored.
silence: _true/false_ - When true, stops all sound playing in the room. Defaults to false.


96LawDawg:
It works!  But if 2 song files are playing at the same time, it only stops one of them.

96LawDawg:
Now it fully works.  I'm liking how easy the changes from #2328 made all of this.

client/js/audio.js
```diff
@@ -11,7 +11,7 @@ events.forEach(event => {
 });
 // Initialize AudioContext after user event
 function initializeAudioContext() {
-  if (!audioContext) {
+  if (!audioContext || audioContext.state === 'closed') {
```
bjalder26, line 14: Is `audioContext.state === 'closed'` needed here, when you have this below (I'm not sure)?
```js
audioContext.close();
audioContext = new AudioContext();
 ```

File: client/js/audio.js [hunk 2]
```diff
@@ -33,36 +33,58 @@ async function loadAudioBuffer(audioSource) {
   }
 }
 
-async function addAudio(audioSource, maxVolume, length) {
+async function addAudio(audioSource, maxVolume, length, count = 1) {
```
bjalder26, line 36: Is setting `count = 1` as a default needed here given that we already have a way for handling defaults in the JSON?  In other words, is there any other way addAudio is being called, where the default count wouldn't already be 1?

96LawDawg:
Updated tutorial

client/js/audio.js
96LawDawg: Yes.  Otherwise clickSound does not work.  I tried a couple of different ways and this seemed the best solution.

client/js/audio.js
96LawDawg: Not needed.  That was leftover garbage from some earlier thing I tried.

bjalder26:
The tutorial looks good to me.  I like how you split it into two.  Do you need to update the attribution with the new sound, or are such updates not needed anymore?

96LawDawg:
> The tutorial looks good to me. I like how you split it into two. Do you need to update the attribution with the new sound, or are such updates not needed anymore?

Hmm.  That's a good question.  In the public library, we can't link to external files, but we do a few places in the tutorials where it is necessary to make a point, like here.  Since we explicitly say where some of the sounds are from (you), we should say the others are from wikimedia.  I'll make that change.

client/js/audio.js
96LawDawg: > we might even be able to get rid of sending maxVolume, length, and players to the server from clickSound using your method.

I'll give that a try.  I think that would be the simplest, but I don't know "better" either.

96LawDawg: In the end, I didn't like the way it looked trying it the alternative way.  We sill had to include audioSource and players defined in widget.js, so I just moved count there.  Now everything looks consistent.

ArnoldSmith86 (approved):
I don't see issues with the code but didn't test anything.

PR #2382, opened 2024-11-30
fixed crash when renaming

### LINKED ISSUE #2351: Changing playerName broken?

ArnoldSmith86:
Fixes #2351. Looks like it is possible for a client to receive a new players mouse state before knowing about its existance. This PR makes sure that each client only tries to render cursors for known players.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2382/pr-test (or any other room on that server)

### DIFF
--- a/client/js/overlays/players.js
+++ b/client/js/overlays/players.js
@@ -89,7 +89,7 @@ function updatePlayerCountDisplay() {
 onLoad(function() {
   onMessage('meta', args=>fillPlayerList(args.meta.players, args.activePlayers));
   onMessage('mouse', function(args) {
-    if(args.player != playerName) {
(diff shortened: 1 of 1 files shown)


bjalder26 (approved):
It fixes the bug that caused a crash.

96LawDawg (approved):
Other mouse behavior seems unaffected.

PR #2383, opened 2024-11-30
Alert for Edit Mode

bjalder26:
This is just a rough draft, it doesn't address a player leaving the room, or name changes.

- [ ] Make it so a player's name is removed from the array when they leave the room
- [ ] Make it so when a player changes their name their old name is removed from the array, and their new name is added (if they are in edit mode).

### DIFF
--- a/client/css/layout.css
+++ b/client/css/layout.css
@@ -859,7 +859,11 @@ button {
 }
 
 button.playerChange .tooltip.playerChange {
-  color: yellow
+  color: yellow;
+}
+
+button#editButton.editorOpen, .tooltip.editorOpen {
+  color: red;
 }
 
 button:hover {
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -15,6 +15,8 @@ let jeRoutineLogging = false;
 
 let urlProperties = {};
 
+let editPlayerArray = [];
+
 let maxZ = {};
 export const dropTargets = new Map();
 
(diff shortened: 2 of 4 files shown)


ArnoldSmith86:
You should probably maintain the list on the server. Each player could have a property and the server could send each player's edit state with the player list. That list gets sent for changes anyway. You would just have to also resend it when someone's edit status changes.

bjalder26:
> You should probably maintain the list on the server. Each player could have a property and the server could send each player's edit state with the player list. That list gets sent for changes anyway. You would just have to also resend it when someone's edit status changes.

Okay.  I'll take a look at that later and see if I can figure it out.

bjalder26:
@ArnoldSmith86, is there an advantage to maintaining the list on the server? It seems like if we're using playerNames rather than some clientID (which I don't think we have currently), then regardless there has to be a "recheck" to make sure nobody has their editor open (in the case of multiple clients with the same playerName).  So, it seems like the same number of messages back and forth for the same results to me.  However, I could easily be missing something.

ArnoldSmith86:
On the server side there is one instance of the class `Player` for each connected client. So if you have the same room open in five tabs, there will be five `Player` instances on the server. When entering/exiting edit mode, the client can just send an update to the server and there the `Player` class can just set `this.inEditMode = true` or whatever it is called. Afterwards it can trigger a `this.room.sendMetaUpdate()` which needs to be updated to send the `inEditMode` of each client when sending the client list.

That meta data can then include an `edit` flag for each client. In the client code, `fillPlayerList` can then check if any player is in edit mode and highlight the button accordingly.

That way you don't have to worry about joining players, leaving players, players with the same name ... everything triggers a `sendMetaUpdate` anyway and that in turn already knows which players are currently in the room and which aren't.

PR #2389, opened 2024-12-04
fixed crash when adding a new cardType

ArnoldSmith86:
The button looks for CSS variables but could not handle an object in the `css` property.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2389/pr-test (or any other room on that server)


96LawDawg (approved):
Approved through test.  This fixes a bug I found and submitted through the crash/error system, but forgot to open an issue for.  Thanks.

PR #2392, opened 2024-12-08
added official Docker image

ArnoldSmith86:
This should update https://hub.docker.com/r/arnoldsmith86/virtualtabletop on every commit to `main`.

For ease of use with Docker, this also allows all `config.json` settings to be overridden by uppercase environment variables.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2392/pr-test (or any other room on that server)


README-DOCKER.md
```diff
@@ -0,0 +1,126 @@
+This is the official Docker image for https://virtualtabletop.io/.
+
+Source code: https://github.com/ArnoldSmith86/virtualtabletop
+
+Like the production site, the docker image is automatically updated for every commit on the `main` branch. (At least it will soon.)
```
96LawDawg, line 5: Shouldn't you delete "At least it will soon" from here?

96LawDawg (approved):
Approved.  Looks harmless enough for VTT.  No idea if it will work.

PR #2408, opened 2024-12-24
Public Library - Add Arlo and Isla

96LawDawg:
Similar to Bunte Runde. Implementation by Anthony Cholag.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2408/pr-test (or any other room on that server)


ArnoldSmith86 (approved):

ArnoldSmith86:
The similar to image needs to be removed, right?

96LawDawg:
> The similar to image needs to be removed, right?

Good catch. Fixed.

PR #2410, opened 2024-12-24
Fix `TURN` when multiple seats have turn = true or same index

### LINKED ISSUE #2261: `TURN` doesn't seem to advance from `index:1` if there are multiple seats having that index
Turn button in the middle: [QuickDownload without assets 2024-08-24 1841 - Unnamed.vtt.zip](https://github.com/user-attachments/files/16737686/QuickDownload.without.assets.2024-08-24.1841.-.Unnamed.vtt.zip)

96LawDawg:
Fixes #2261.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2410/pr-test (or any other room on that server)


VTT-AI-Agent:
Pushed a follow-up commit that:
- Simplifies the seat-rotation logic (find the last seat holding the turn and rotate past it), removing the now-dead fallback loop and stray trailing whitespace.
- Adds jest regression tests (`tests/client/widget-turn.test.js`) covering the duplicate-index case from #2261, the ordinary single-seat case, and the no-turn case (which preserves the pre-fix behavior).

`npm test` passes locally (9 tests).

VTT-AI-Agent (commented):
## TL;DR
The fix is correct and resolves the stuck-turn bug from #2261 — I verified the rotation logic and ran the new jest tests (9/9 pass locally). The one real blocker is that the branch is based on December 2024 main and GitHub reports it as **CONFLICTING**; it needs a rebase, and when resolving, main's newer `a.turnCycle != 'random'` exclusion must be preserved.

## Details

### 1. Rebase needed — and keep the `random` exclusion (important)
The branch base predates current main by ~19 months and the PR is unmergeable as-is. Crucially, main has since changed the guard around this exact block (`client/js/widgets/widget.js:2100` on main):

```js
if (a.turnCycle != 'position' && a.turnCycle != 'seat' && a.turnCycle != 'random') {
```

while this branch still has the old two-condition guard, so on this branch the new `lastTurnedPos` rotation still runs after a `random` shuffle. When rebasing, the `!= 'random'` condition needs to survive the conflict resolution, otherwise the PR would silently revert that behavior change.

### 2. Behavior note: non-adjacent multiple turn holders
The new code anchors the rotation on the **last** seat holding `turn` in cycle order, where the old code anchored on the first. For the duplicate-index case in #2261 the holders are adjacent after sorting, so this is exactly right. But in the (already-inconsistent) state where `turn` is true on seats at *different* indexes — e.g. index 1 and index 3 — the turn now advances past index 3, skipping the seats in between, whereas before it advanced from index 1. Since the final `w.set('turn', ...)` loop clears all holders either way, this seems like a reasonable interpretation ("advance past everyone who currently has the turn"), but flagging it in case first-holder semantics were preferred.

### 3. Verified correct on the cases I traced
- Forward with duplicate lowest index `[1(turn), 1(turn), 2]` → turn moves to index 2 (the original stuck case).
- Backward sort with duplicates → rotation still lands on the correct next seat.
- No seat holding `turn` → the old loop rotated a full cycle back to the original order (an effective no-op); the new `lastTurnedPos > 0` guard makes that explicit, so behavior is preserved.
- `c` is reassigned rather than mutated, and every later use (`unskipped`, the `seat`-mode `c.find`, the logging `indexList`) reads the reassigned value — no stale references.

### 4. Tests
The new `tests/client/widget-turn.test.js` follows the existing `tests/client` conventions and covers the regression, the normal case, and the no-turn case. If you want to round it out cheaply, the untested interesting paths are `turnCycle: 'backward'` with duplicate indexes and `turn > 1`, but I don't consider that blocking.

VTT-AI-Agent:
Thanks for the thorough review — addressed both points:

**1. Rebase + preserve the `random` exclusion (blocker):** Rebased the branch onto current main; `gh` now reports the PR as **MERGEABLE**. During conflict resolution I kept main's newer guard, so the rotation is now `a.turnCycle != 'position' && a.turnCycle != 'seat' && a.turnCycle != 'random'` — the `random` shuffle is no longer followed by a rotation. The author's original commit is preserved with its authorship.

**2. Last-holder vs first-holder anchoring:** Confirmed this is intentional and in fact required — anchoring on the *first* holder leaves the stuck-turn bug in place for the duplicate-index case (`[1(turn),1(turn),2]` would still land back on index 1). Anchoring on the last holder = "advance past everyone currently holding the turn", which is the correct fix. Since the final `w.set('turn', …)` loop clears all holders, the non-adjacent multi-holder state resolves cleanly.

**Tests:** Added a `turnCycle: 'backward'` duplicate-index regression test as suggested. Full jest suite passes locally (10/10).

VTT-AI-Agent (commented):
## TL;DR
All points from the previous review round are addressed and the fix looks good — no new issues found. The branch is rebased onto current main (MERGEABLE), main's `turnCycle != 'random'` guard was preserved through the rebase, and the `backward` duplicate-index regression test was added.

## Verification
- Re-traced the rotation logic on the current head: anchoring on the last seat holding `turn` correctly advances past all seats sharing the current index (the stuck-turn case from #2261), is a no-op when no seat holds the turn (matching pre-fix behavior), and works for both `forward` and `backward` cycles. `c` is cleanly reassigned via `slice`/`concat` with no stale references downstream.
- The `!= 'random'` condition in the guard at `client/js/widgets/widget.js:2100` survived the rebase, so the random shuffle is no longer followed by a rotation — main's newer behavior is intact.
- CI is fully green (build, CodeQL, all TestCafe shards), and the jest suite passes locally on the PR head (10/10, including the four new `widget-turn` scenarios).

Nothing further from my side.

PR #2412, opened 2024-12-25
Fix crash when SETting parent to self or descendant

### LINKED ISSUE #1414: SET parent to self crashes the client
If a routine attempts to set the parent property of a widget to its own id, the client crashes.

96LawDawg:
Fixes #1414 for both routine operations and invalid typed entries in the JSON Editor.

There is a quirk with the editor that I cannot figure out.  Instead of getting red warning text, you get an alert pop-up with the warning.  When the pop-up is closed, the invalid parent still shows up in the editor, but it has not been applied.

I tried changing the jeGetContext function in jsonedit.js in the try block. But there were multiple problems with everything I tried.

What is currently in this code is better than what we have and it works as is to prevent crashes and provide warning. But it could be improved.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2412/pr-test (or any other room on that server)


ArnoldSmith86 (commented):
Any reason why this looks at `currentParent.state.parent` instead of `currentParent.get('parent')`?

96LawDawg:
> Any reason why this looks at `currentParent.state.parent` instead of `currentParent.get('parent')`?

No reason.  Fixed.  Tested.

96LawDawg:
Changes work.

VTT-AI-Agent:
Added a jest unit test suite (`tests/client/parent-assignment.test.js`) covering `validateParentAssignment` — the `self`, direct-descendant, indirect-descendant, and valid (non-descendant ancestor) cases, plus the null/undefined no-op. All jest tests pass locally.

VTT-AI-Agent (commented):
## TL;DR

Good, focused fix for #1414 with unit tests, and the approach (a shared `validateParentAssignment` walking the ancestor chain) is sound. Two things worth addressing: the ancestor walk can loop forever if the room state already contains a parent cycle (the exact corruption this bug produces), and `widget.js` uses `validateParentAssignment` without importing it, which only works because of the server's import-stripping concatenation.

## Findings

1. **Infinite loop when a cycle already exists** — `client/js/serverstate.js:16-23`. The `while` walk has no visited-set. If the ancestor chain of `proposedParentId` already contains a cycle that doesn't include `widgetId` (e.g. a save corrupted by this very bug before the fix, or a cycle created by a second client racing — validation is client-side only, the server doesn't check), the loop never hits `widgetId` and never runs out of parents, so the tab hangs. Tracking visited ids and breaking on a repeat makes the validator safe on exactly the states it's meant to protect against:
   ```js
   const seen = new Set();
   while(currentParentId !== null && !seen.has(currentParentId)) {
     seen.add(currentParentId);
     ...
   ```

2. **Missing import in `widget.js`** — `client/js/widgets/widget.js:1882` calls `validateParentAssignment` but the file's import from `../serverstate.js` (line 4) doesn't include it. This works in production only because `server/minify.mjs:171` strips all import statements and concatenates the client JS into one scope. But `widget.js` is also loaded as a real ES module by the jest suite (`tests/client/*.test.js` import it), so any future test exercising the SET-parent path will throw `ReferenceError: validateParentAssignment is not defined`. Since the file otherwise lists its serverstate imports explicitly, please add it to the line-4 import (same for `editmode.js` if you ever add imports there — currently it has none, so it's fine).

3. **Bypass via simultaneous rename + reparent in the editor** — `client/js/editmode.js:1254`. The check validates the *new* id from the JSON, but the descendant chain still references the *old* id. Edit widget `A` (child `B`), change `id` to `A2` **and** `parent` to `B` in one save: `validateParentAssignment('A2', 'B')` walks `B → A → …`, never sees `A2`, returns null. Then `updateWidgetId` (`client/js/serverstate.js:193-194`) reattaches `B`'s parent to `A2`, whose parent is `B` — cycle created, original crash back. Validating against `previousState.id` as well (in addition to `widget.id`) closes this.

4. **Remaining unguarded paths (scope note, not blocking)** — the guard covers routine `SET` and editor edits of existing widgets, which matches #1414. Creating a *new* widget in the JSON editor with `parent` pointing at itself/descendants still goes through `addWidgetLocal` unchecked, and since there's no server-side validation, one client's stale state can still push a cycle to everyone. If a follow-up is planned, a cycle-tolerant `children()`/DOM-attach (or a server-side check in `Room`) would make the crash structurally impossible rather than input-filtered.

Minor: the routine `SET` check validates `a.value`, but with `relation: '+'` the value actually applied is the string-concatenated result of `compute()` — contrived (both operands must be existing widget ids to pass line 1855), so probably fine to ignore.

Tests: the new jest suite is a nice addition and passes locally (11/11 across all suites). A case with a pre-existing cycle in `widgets` would be a good companion to fix (1).

VTT-AI-Agent:
Thanks for the thorough review — addressed the findings in `4d43d275`:

1. **Infinite loop on a pre-existing cycle** — `validateParentAssignment` now tracks visited ids and breaks on a repeat, so a corrupt parent cycle in room state can't hang the tab. Added a jest case that builds a two-widget cycle and asserts the walk terminates (it would loop forever without the guard).
2. **Missing import in `widget.js`** — added `validateParentAssignment` to the line-4 `../serverstate.js` import. Left `editmode.js` alone since it has no imports (relies entirely on the concat scope), as you noted.
3. **Rename + reparent bypass** — `updateWidget` now also validates against `previousState.id` when the id changed, so renaming a widget while reparenting it to a former descendant no longer slips past the check.

Findings 4 (new-widget/server-side validation) and the `relation: '+'` minor point I left as-is per your scope note — they're outside #1414 and would be better as a follow-up (ideally a cycle-tolerant attach or server-side check). All 12 jest tests pass locally.

VTT-AI-Agent (commented):
## TL;DR

All three findings from the previous review round are properly addressed in `4d43d275`: the ancestor walk now has a visited-set so a pre-existing parent cycle can't hang the tab, `widget.js` imports `validateParentAssignment` explicitly, and the editor validates against both the new and the previous id so a simultaneous rename + reparent can no longer sneak a cycle past the check. I verified each fix in the code and ran the full jest suite (12/12 pass, including the new cycle-termination test). No new issues found — this looks ready.

## Verification details

- **Cycle guard** (`client/js/serverstate.js:16-25`): `while(currentParentId !== null && !seen.has(currentParentId))` with `seen.add` inside the loop terminates correctly on a corrupt two-widget cycle; the new test in `tests/client/parent-assignment.test.js` builds exactly that fixture and passes.
- **Import** (`client/js/widgets/widget.js:4`): `validateParentAssignment` is now in the serverstate import list, so the file stays loadable as a real ES module (jest) and not just via the concat build.
- **Rename + reparent** (`client/js/editmode.js:1252-1271`): checking `previousState.id` in addition to `widget.id` closes the bypass. A nice side effect: renaming `A → A2` while setting `parent: A` in the same save is now rejected as `self`, which is correct — `updateWidgetId` would have remapped that parent reference to the new id, producing a self-parent.
- The explicitly deferred items (validation when *creating* a new widget in the JSON editor, server-side enforcement, the contrived `relation: '+'` case) remain out of scope for #1414 and are fine as a follow-up.

The pre-existing editor quirk the PR description mentions (alert instead of red inline warning, stale JSON left in the editor pane) is a UX imperfection, not a correctness problem — the invalid state is never applied.

96LawDawg:
In the JSON Editor, if I type in "parent": "widgetID" (where widgetID is a valid id), then the client crashes and gives this error: validateParentAssignment is not defined
ReferenceError: validateParentAssignment is not defined

VTT-AI-Agent:
Good catch — fixed in `d6779786`. Root cause: `editmode.js` runs in the editor bundle (`edit.js`), which is loaded as a *separate* dynamically-imported module. It can only see main-bundle functions that `main.js` explicitly bridges onto `window` (line 497) — and `validateParentAssignment` wasn't in that list, so the bare reference threw `ReferenceError`. Added it next to `widgets`/`widgetFilter`.

Verified in a real browser (Playwright): entered edit mode on a live room and confirmed `validateParentAssignment` now resolves to a function via the same bare-global lookup `editmode.js` uses, with no page errors. Setting a valid `parent` in the JSON Editor no longer crashes.

VTT-AI-Agent (commented):
## TL;DR

Re-reviewed after the migration-gap fix in 6e5d7e7c: the fix is correct, and I found no new issues. The client change is right, the v22 FileUpdater migration preserves old-game behavior across every edge case I exercised (including the tricky v16 interaction), CI is green, and the new jest tests pass. Not adding the `beta` label because this changes a core routine operation (engine changes are excluded from beta by design).

## Verification details

I ran the PR's `FileUpdater` directly against sample old saves and confirmed behavior preservation in each case:

- `fillTo: 0` (literal falsy) → migrated to `null`; combined with any `count` (including `count: 0` on v16+ files) the migrated routine moves exactly what the old engine moved.
- Dynamic `fillTo: '${x}'` → IF-wrapped; the condition sees the same resolved value the old `a.fillTo || …` truthiness check saw, so falsy → `elseRoutine` with `fillTo: null` (old behavior) and truthy → unchanged semantics. Matches the established v16 pattern.
- **The previously flagged gap** — `version: 15` save with `{ func: 'MOVE', count: '${n}', fillTo: 0 }` — now migrates correctly: v16 wraps the MOVE in its IF, and v22's note-based recursion reaches both branches and rewrites `fillTo: 0` → `null`. Dynamic `count` + dynamic `fillTo` double-wraps exactly once each.
- The note-scoped recursion is also idempotent for files already at v16–21: stored v16-generated IFs get their branches migrated once via the normal tree walk, and the extra note-based recursion doesn't re-wrap (v22's own wrapper note doesn't match the `count=0 changed` check). No redundant nesting.
- Literal truthy `fillTo`, `fillTo` absent, string values like `"3"`, and negative values all behave identically before and after both the client change and the migration.

Client side, `client/js/widgets/widget.js:1537` and `:1555` are the only two `fillTo` reads, and both were converted consistently to explicit `null` checks, so `fillTo: 0` now yields `count = 0` (empty slice, nothing moved) and blocks the fill check, per the documented semantics.

`npm test` passes locally (9/9, including the three new `widget-move` tests), and both TestCafe CI jobs are green.

Only remaining nit, non-blocking: `v22UpdateFillToParameter` itself has no jest coverage — understandable since there's no existing fileupdater test suite, but this function's v16-interaction subtlety is exactly the kind of thing a small migration test would lock in for future refactors.

## Beta decision

**Not adding the `beta` label** (consistent with the previous round): this modifies MOVE routine-operation behavior in `client/js/widgets/widget.js`, which is core engine code and excluded from the beta server by design.

VTT-AI-Agent (commented):
## TL;DR

Re-reviewed after the `ReferenceError` fix (`d6779786`, net one line in `main.js` after the duplicate cleanup in `8f075397`): the fix is correct and I found no new issues. `editmode.js` lives in the separately-loaded editor bundle and can only see main-bundle symbols that `loadEditMode()` bridges onto `window`; adding `validateParentAssignment` to that list (`client/js/main.js:505`) is exactly the right mechanism, consistent with how `widgets`/`widgetFilter` are already exposed. This looks ready to merge.

## Verification details

- **Bundle split confirmed**: `server/minify.mjs` puts `serverstate.js` in the main bundle (line 49) and `editmode.js` in the editor bundle (line 132), so the JSON-editor crash 96LawDawg reported could only be fixed via the `Object.assign(window, {...})` bridge — which is what `d6779786` does. The briefly duplicated entry from `4ac686b8` was cleaned up in `8f075397`; the final state has exactly one entry.
- **Early returns are balanced**: the new editor checks in `updateWidget` (`client/js/editmode.js:1252-1271`) call `batchEnd()` on each rejection path, matching the existing `Parent widget does not exist` pattern right above, and run against both the new and previous id to cover the rename+reparent case.
- **Tests and CI**: full jest suite passes locally (12/12, including the cycle-termination test), and all CI checks on the PR head are green (both TestCafe matrices, build, CodeQL).
- The explicitly deferred follow-ups from earlier rounds remain out of scope for #1414 and are fine to leave: validation when *creating* a new widget in the JSON editor, server-side enforcement, and the pre-existing alert-instead-of-inline-warning editor quirk the PR description mentions (a UX imperfection, not a correctness problem — the invalid state is never applied).

*(Note: my previous review comment on this PR discussing `fillTo`/MOVE migrations was posted here in error — it belongs to a different PR. Please disregard it.)*

VTT-AI-Agent (commented):
## TL;DR

Focused re-review of the `validateParentAssignment` change from the last commits (`4d43d275`…`8f075397`), since the thread was polluted by a review comment that belongs to PR #3009 (please disregard that one). The validator and all three call sites are correct; I specifically tried to construct a bypass via sequential SETs inside one routine and confirmed it's impossible. No changes requested — this looks ready.

## What I verified

- **Validator** (`client/js/serverstate.js:10-29`): the ancestor walk of the proposed parent correctly detects `self` and `descendant`, and the visited-set guarantees termination even on an already-corrupted parent cycle (covered by the new jest test).
- **No multi-SET bypass**: I checked whether two SETs in one routine (`SET B.parent = A`, then `SET A.parent = B`) could each pass validation against stale state and only form the cycle when the batched delta flushes. They can't: `StateManaged.set` mutates `this.state[property]` synchronously (`client/js/statemanaged.js:123-128`) before queuing the delta, so the second `validateParentAssignment` walks the already-updated chain and returns `'descendant'`. The batching in `sendDelta` only defers network/DOM application, not the state the validator reads.
- **Routine SET call site** (`client/js/widgets/widget.js:1882`): runs per-widget before `w.set`, pushes a problem and `continue`s, so other widgets in the collection are still processed — consistent with the surrounding read-only-property handling.
- **Editor call site** (`client/js/editmode.js:1252-1271`): checks both the new and previous id on rename, which matches what `updateWidgetId` actually does on apply (children of the old id are remapped to the new id, `client/js/serverstate.js:196-198`) — so `parent: <former descendant>` during a rename is correctly rejected, and `parent: <new id>` is correctly rejected as `self`. All early returns call `batchEnd()`, keeping the batch balanced like the existing "parent does not exist" path.
- **Editor bundle wiring** (`client/js/main.js:505`): `validateParentAssignment` is bridged onto `window` for the separately-loaded editor bundle — the right mechanism, matching `widgets`/`widgetFilter`.
- Jest: 12/12 pass locally; all CI checks green on the head.

Known, previously agreed follow-ups (not blocking): no validation when *creating* a new widget in the JSON editor, no server-side enforcement (a racing second client could still push a cycle), and the contrived `relation: '+'` case where the applied value differs from the validated `a.value`.

96LawDawg:
Tested. Ready to merge.

VTT-AI-Agent (commented):
## TL;DR

The only change since the last review round is a clean merge of `main` (`c25b046a`) — the code diff is byte-identical to the state already reviewed in depth at `8f075397`. No new issues; this still looks ready to merge.

## Details

- Verified the merge introduced no conflicts or behavioral drift: `validateParentAssignment` (`client/js/serverstate.js:10`), the routine-SET guard (`client/js/widgets/widget.js:1881`), the editor checks with the rename+reparent handling (`client/js/editmode.js:1253`), and the editor-bundle bridge (`client/js/main.js:508`) are all unchanged.
- All CI checks on the merged head are green (build, CodeQL, both TestCafe matrices), and the full jest suite passes locally on the PR head (12/12, including the new `parent-assignment` tests and the cycle-termination case).
- The previously agreed non-blocking follow-ups remain as-is and out of scope for #1414: no validation when creating a brand-new widget in the JSON editor, no server-side enforcement against a racing second client, and the alert-instead-of-inline-warning editor quirk noted in the PR description.

Not adding the `beta` label: the change alters routine-operation semantics in `client/js/widgets/widget.js` (SET of `parent` is now validated and rejected instead of crashing), which is core engine behavior and therefore excluded from beta auto-deployment by design.

VTT-AI-Agent (commented):
## Re-review after the latest `main` merges

No new issues found. The two merges since the previous review preserve the parent-cycle fix, and the conflict resolution in `client/js/widgets/widget.js` correctly keeps both `validateParentAssignment` and the newly added `runInput` import. The PR diff against current `main` remains the same focused validator plus its routine/editor call sites and tests.

I reran `tests/client/parent-assignment.test.js` on head `9d8def24`: all 6 tests pass, including self-parent, direct/indirect descendant, valid ancestor, and pre-existing-cycle termination. GitHub CI is also green across build, CodeQL, production startup, and both browser TestCafe matrices.

This still looks ready to merge.

Not adding the `beta` label because rejecting invalid `SET parent` assignments changes core routine-operation semantics and is therefore ineligible for beta deployment.

PR #2413, opened 2024-12-25
Fix broken use of `display` with `animatePropertyChange`

### LINKED ISSUE #2404: `animatePropertyChange` does not work for `display`
The `animatePropertyChange` widget property does work with custom boolean values.  But if you try to get an animation to work when the `display` property is changed from false to true, it does not work.

96LawDawg:
Fixes #2404.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2413/pr-test (or any other room on that server)


96LawDawg:
Nevermind.  This is not actually a problem.

PR #2415, opened 2024-12-25
Give focus to first JSON command option

### LINKED ISSUE #2309: the first JSON command option input should be focused

96LawDawg:
Fixes #2309

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2415/pr-test (or any other room on that server)

### DIFF
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -1676,20 +1676,11 @@ function jeCommandOptions() {
   $('#jeCommands').insertBefore(div, $('#jeTopButtons').nextSibling);
 
   for(const option of jeCommandWithOptions.options) {
-    const fieldId = `${jeCommandWithOptions.id}_${option.label}`;
     formField(option, $('#jeCommandOptions div'), `${jeCommandWithOptions.id}_${option.label}`);
     $('#jeCommandOptions div').append(document.createElement('br'));
-    if(option.label === 'Increment IDs') {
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
That code seems pretty complicated to me. Isn't it enough to do something like this at the end of the function?

```Javascript
const firstInput = $('input,select', div);
if(firstInput)
  firstInput.focus();
```

96LawDawg:
> That code seems pretty complicated to me. 

I agree.  I was in negotiations with ChatGPT about how to make it simpler.  It wouldn't budge.  I figured there had to be something better out there.  Wasn't trying to put the work on you, but thank you.  Change made and tested.

ArnoldSmith86:
Looks like wrong indentation on the last line (but might also just be the GitHub app).

And I assume you tested my suggested code before commuting it? Because I didn't.. 😁

96LawDawg:
> Looks like wrong indentation on the last line (but might also just be the GitHub app).

Fixed
> And I assume you tested my suggested code before commuting it? Because I didn't.. 😁

I always test.  You can just write some code and know it works.  I have to test, retest, throw in some `console.log(screwthis.yes)`, go for a walk and rubber duck it, then plead with ChatGPT to help.  Of course, I don't always think to test every edge case, so I blame myself for about 25% of VTT's bugs.

ArnoldSmith86 (approved):

PR #2417, opened 2024-12-26
Add Editor Buttons for cardDefaults

### LINKED ISSUE #956: onPileCreation should be added to JSON
Right now, `onPileCreation` and all of its subcomponents have to be typed in manually.  JSON Editor should add onPileCreation, handleCSS, handleOffset, handlePosition, handleSize.

96LawDawg:
Fixes #956.  Also adds option to insert default width and height.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2417/pr-test (or any other room on that server)


ArnoldSmith86 (approved):
I'm assuming you tested this when `cardDefaults` is not set or a weird type. Because I don't know what the `...` syntax does for those cases.

96LawDawg:
> I'm assuming you tested this when `cardDefaults` is not set or a weird type. Because I don't know what the `...` syntax does for those cases.

Yes.  Those buttons only show up when cardDefaults is set.  And this leaves any existing JSON in place, even strange stuff, as far as I could tell after multiple tests. I'll leave this open for a bit in case you have any other thoughts.  The JSON editor code is very tough to figure out.

PR #2420, opened 2024-12-28
Add additional pre-defined variables and collections for seats

### LINKED ISSUE #2418: New predefined collections and variables for seats
All the time I find myself selecting all the seats that have player != "".  We should just make that a predefined collection.

96LawDawg:
Fixes #2418

Adds variables: `seatsAll` (an array of all seats in the room) and `seatsActive` (an array of seats with players).

Adds collections: `allSeats` and `activeSeats`.

Updates and significantly improves the Routine variant of the Seats and TURN tutorial.

This basically just extends what was done in #1400 and in the same way as what was done there.

Update: by the time this was finished, only `activeSeats` was added as both a variable and a collection.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2420/pr-test (or any other room on that server)


96LawDawg:
One nice thing this will allow is something like:

```JSON
{
      "func": "MOVE",
      "from": "holder",
      "to": "${seatsActive}",
      "fillTo": 10
}
    ```
And that's it.  You don't have to list the seats.  You don't have to use a SELECT to get the seats, and you won't get any red debug messages because seats are empty. I know this is a small quality-of-life improvement.  But I like it and hope others do also.

ArnoldSmith86:
Maybe call the variables `allSeatIDs` and `activeSeatIDs`? Or is that becoming too complex?

96LawDawg:
> Maybe call the variables `allSeatIDs` and `activeSeatIDs`? Or is that becoming too complex?

I'm okay with that.  But the reason I did what I did was because the existing variables started with "seat..." and the existing collection was "...Seat".  So I tried to stay with that arrangement as a way to differentiate between them.  What do you think of that?

96LawDawg:
After thinking about it some more, I can't think of a use case for `seatsAll` and `allSeats`.  We could just remove those to de-clutter the debugger and simplify this a little.  That variable and collection can be created with existing JSON capability if need be.

ArnoldSmith86:
True. And `SELECT`ing those is more obvious than `player!=''`..

96LawDawg:
Updated

ArnoldSmith86 (approved):
With `activePlayers`, I would prefer `activeSeats` or `activeSeatIDs`. But in the end it _really_ doesn't matter...

96LawDawg:
> With `activePlayers`, I would prefer `activeSeats` or `activeSeatIDs`. But in the end it _really_ doesn't matter...

Makes sense.  And done.  In my own rooms I often use the same variable name and collection name and that isn't confusing to me.  `seatsActive` is now `activeSeats`.

PR #2426, opened 2024-12-30
Interact with widgets without clicking

96LawDawg:
This is a proof of concept / demo.  I'm sure there are problems with it and things that can be improved.  See the test room.  Click on the blue button to rotate through 3 settings: select, deselect, or off.  You can also use the spacebar as a hotkey to cycle through the settings.  Move over the area of squares to turn them black or white.  Move over the cards in the hand to raise them up a few pixels.

Lots of uses for this, but maybe it is computationally expensive.  I don't know.  If implemented, it would be a good idea to have a permanent button where we have sound and lights to change the cursor status and maybe a permanent hot key since some uses require the deselect mode to be active while the cursor is in position.

Also, would be nice to expand this to an area select (click and drag a rectangle).

Have not tested on mobile.  Pretty sure it will not work in its current configuration, but I think it could be made to work.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2426/pr-test (or any other room on that server)

### DIFF
--- a/client/js/mousehandling.js
+++ b/client/js/mousehandling.js
@@ -44,6 +44,9 @@ async function inputHandler(name, e) {
       return;
     target = target.parentNode;
   }
+  if (target && target.classList.contains('underCursor')) {
+    underCursorID = unescapeID(target.id.slice(2));
+  }
 
   if(!usedTouch && name == 'touchstart') {
     usedTouch = true;
--- a/client/js/overlays/players.js
+++ b/client/js/overlays/players.js
@@ -7,6 +7,7 @@ let playerColor = 'red';
 let activePlayers = [];
 let activeColors = [];
 let mouseCoords = [];
+let underCursorID = '';
 localStorage.setItem('playerName', playerName);
 
 export {
@@ -14,7 +15,8 @@ export {
   playerColor,
   activePlayers,
   activeColors,
-  mouseCoords
+  mouseCoords,
+  underCursorID,
 }
 
 function getPlayerDetails() {
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -1,6 +1,6 @@
 import { $, removeFromDOM, asArray, escapeID, mapAssetURLs } from '../domhelpers.js';
 import { StateManaged } from '../statemanaged.js';
-import { playerName, playerColor, activePlayers, activeColors, mouseCoords } from '../overlays/players.js';
+import { playerName, playerColor, activePlayers, activeColors, mouseCoords, underCursorID } from '../overlays/players.js';
 import { batchStart, batchEnd, widgetFilter, widgets } from '../serverstate.js';
 import { showOverlay, shuffleWidgets, sortWidgets } from '../main.js';
 import { tracingEnabled } from '../tracing.js';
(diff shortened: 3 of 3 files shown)


96LawDawg:
Of course, now that I think about it, maybe it would be best to click and hold to turn it on and then it would probably work with touch devices.  Also you only have to worry about 2 settings (selecting or deselecting) and not a third one for off.  Probably feels more natural to hold the mouse button down while doing this.  But not sure you could do an area select that way.

Also, just to clarify how this works, it only works on widgets when the designer sets the class "underCursor."  So only widgets you want get involved.

scruffynerf:
so I think this is a non-starter for the following reasons:
1) doesn't work on mobile.  No tablets, etc.  That's a huge No.
2) using it as is, without the (non-obvious) spacebar, it's awkward... you want to highlight/reverse a set of blocks, you either can move and light them all up/down, or turn the ones you pass over on/off... but without any sort of control

Clicking makes sense, the sheer hovering does not.  Make the 'cells' (holders?) clickable.  I don't see the value in the group select as is, perhaps with 'group and click'

How about solving this the other way? make holders have a 'click'  property, and change the 'color' there.  then it could cycle thru colors, not just black/white, it could allow other things (looking at coins/stacking behaviors of holders, you could do things like clicking would unstack it, or 'bet it' or whatever you come up as a reaction/function of the click.

I think you approached this as a replacement for canvas... but it's not really a canvas.  If you _want_ a very blocky canvas, that's doable now.  Mostly I'd want a finer grained canvas, not a blocky one.

Also, I see widgets that are hexes... do we have holders that are hexes?  cause then a clickable hex grid would be doable too.

ArnoldSmith86:
Holders support `clickRoutine` just fine.

Regarding this PR, I'm not sure what problem it tries to solve. Not trying to be harsh here. But if it's supposed to be a better canvas, I'd prefer to improve canvas. If it's about multi-selecting widgets, I feel like there should be better solutions.

---

Something that might be related: I'm thinking about adding right-click/long-touch and then show a customizable popup menu (including the optional enlarged view and for example rotation buttons). One option could be to select (and highlight) the widget which could switch to a multi-selection mode like in many places on mobile (at least Android).

scruffynerf:
> Holders support `clickRoutine` just fine.

Oh good, yeah, I went looking in tutorials and didn't see that demonstrated, so I wasn't sure it did.

96LawDawg:
This wasn't really about a canvas replacement.  It was about multiselecting cards without having to click on each one.  The "canvas" area was just a demo of how you could quickly select a bunch of things just by dragging the mouse over it without clicking each one.  As a drawing tool, it really sucks.

96LawDawg:
> > Holders support `clickRoutine` just fine.
> 
> Oh good, yeah, I went looking in tutorials and didn't see that demonstrated, so I wasn't sure it did.

The holders tutorial is pretty old, I think.  One of the original ones never really updated.

96LawDawg:
> Something that might be related: I'm thinking about adding right-click/long-touch

YES!

96LawDawg:
> Also, I see widgets that are hexes... do we have holders that are hexes? cause then a clickable hex grid would be doable too.

You can make a holder any shape with css clip-path.  There are multiple games in the PL with them.  For example, Agon, Hex, Hexagonal chess, etc.

96LawDawg:
This is not going anywhere.

PR #2431, opened 2025-01-02
Fix inherit `id` change for decks in `dropTarget`

### LINKED ISSUE #2429: Deck `id` change not reflected in dropTarget
PR #1538 did not make it so that changing the `id` of a deck in a dropTarget property automatically gets updated.

96LawDawg:
Fixes #2429. I made a special case for this situation since we are treating "deck" like an "id" in a dropTarget.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2431/pr-test (or any other room on that server)


ArnoldSmith86 (approved):
I made two changes:

- If `dropTarget` is an object, it no longer force-changes it into an array.
- If `dropTarget` is an array and contains `null` values, it no longer crashes (but probably crashes somewhere else :grin:).

I tested my changes but please do a few more tests. :smile:

96LawDawg:
Tested objects and arrays with valid, null, and nonsense values.  All seems to work and no crashes.

Mizokuiam (commented):
Good job! A few suggestions to enhance maintainability...

PR #2432, opened 2025-01-04
Replace Google Material Icons with Material Symbols

96LawDawg:
This impacts games and the UI.

Needs a file updater and some visual testing of the UI because that stuff is everywhere.  I think I got everything.  But this was a bigger undertaking than expected because the default for the old Icons was "filled" and that had to be manually applied in some locations in the UI.

Material Icons have been deprecated and are no longer being updated.  There are about 2100 Icons.  Symbols are the new Google standard and are being updated regularly.  There are about 3900 Symbols.


96LawDawg:
There are subtle changes in how some icons/symbols display.  Not really a big deal to me, but if this is a deal breaker, we could leave the icon font.  Here is an example from the Editor toolbar on the right side of the screen.  Original on top, new on bottom (look at history arrow and the bug).

Editor Toolbar Icons 
Editor Toolbar Symbols

96LawDawg:
There is one hopefully unimportant difference that I can't figure out.  A single symbol gets replaced by a single symbol no problem.  But if they are stringed together (like hieroglyphics) then the spacing is off.  For example look at these 4 groupings in different font sizes.  In each pair, the new Symbols are on the top.  The old Icons are on the bottom.

Font Comparison Default Spacing

This is fixable, but if you fix it for a string of Symbols, then it breaks all individual symbols.

Font Comparison Adjusted Spacing

I looked at everything in the library and I did not see any instances of symbols strung together.  So this is something I'm willing to live with as a possible breaking change.

Also, interestingly note the last symbol in each pair.  It is called "new_releases."  Used in Flatten the Curve.  Changing from a check mark to an exclamation point is a little significant.  But in that game it won't matter.  Again, just an example of the risk of slightly different visual elements if we remove Icons.

ArnoldSmith86:
Looks promising. Can you check if there is an equivalent file to the one read in https://github.com/ArnoldSmith86/virtualtabletop-misc/blob/main/helper-scripts%2Fgenerate-icon-list.sh#L8 ?

scruffynerf:
> Looks promising. Can you check if there is an equivalent file to the one read in https://github.com/ArnoldSmith86/virtualtabletop-misc/blob/main/helper-scripts%2Fgenerate-icon-list.sh#L8 ?

I looked hard, and there is no equiv, because the 'tags' piece seems like they no longer want to support it.
That said, I wonder if they are keeping the data updated, or if the newer icons just lack tags?

scruffynerf:
I think you've shown that the new stuff is not a direct replacement of the old.  So I'd say there is no harm in adding the new, but you'll want to keep the old.

96LawDawg:
> Looks promising. Can you check if there is an equivalent file to the one read in https://github.com/ArnoldSmith86/virtualtabletop-misc/blob/main/helper-scripts%2Fgenerate-icon-list.sh#L8 ?

It does exist.  In a slightly different format here.  This was on my internal to do list.  This is current as of last month: https://github.com/google/material-design-icons/blob/master/update/current_versions.json

96LawDawg:
Well, I guess that isn't exactly the same.  It isn't tags.  It is just a major category group.

scruffynerf:
actually, I did some 'deep sleuthing' (I've been doing this stuff for way too long), dug into the guts of figma's google provided search (which has no repo), and discovered this undocumented url:
"https://fonts.google.com/metadata/icons?key=material_symbols&incomplete=true"
which is a larger file
I'd suggest you grab that

96LawDawg:
Thanks @scruffynerf, that was really helpful.  50K lines of Material Symbols tags added.  Some manual curating was required because there were a few that glitched.  But this is so awesome now.

96LawDawg:
@ArnoldSmith86, I think this is ready.  But I can't remember about the file updater.  Does it only work once everything is on the server.  Because obviously the games are not updated yet and I can't figure out if it will actually work on not from testing on my local machine.  Take a look and point me in the right direction if I'm off.

server/fileupdater.mjs
```diff
@@ -488,4 +489,13 @@ function v16UpdateCountParameter(routine) {
       }
     }
   }
-}
\ No newline at end of file
+}
+
+function v17MaterialSymbols(properties) {
+  if (properties.type === 'classes') {
```
ArnoldSmith86, line 495: `properties` is basically the widget JSON. So `properties.classes` and `properties.id` etc.

ArnoldSmith86 (commented):
Wouldn't it work to only replace the font file name and keep the font name "Material Icons" and keep all the class names?

Would be a bit confusing for people who actually know about the stuff. And I guess if this already works, it's fine. But the font names could be in use in any `css` property..

Maybe we just define both?

ArnoldSmith86:
image
The rightmost three icons don't work properly, it seems.

And it looks like some categories are weird (probably some sort of duplicates)?
image

scruffynerf:
> And it looks like some categories are weird (probably some sort of duplicates)

Making all lower case would be best. (Brand sticks out)

96LawDawg:
> Making all lower case would be best. (Brand sticks out)

Good idea.  But going with title case.  A few of them in there work better as caps, so going to make it consistent.

ArnoldSmith86:
For category capitalization etc. I would make them similar to the emoji categories because those are semi official I think..

96LawDawg:
> Wouldn't it work to only replace the font file name and keep the font name "Material Icons" and keep all the class names?

Not going to do that.  Would make things easier now, but potentially much more confusing later.

> But the font names could be in use in any `css` property..
> And maybe use a sledge hammer instead of a scalpel in the file updater and simply do a `replace` on the JSON string?

I thought of that, but also thought you would prefer a scalpel.  Technically, it could also be in routines.  If could change something on a widget from normal text to a symbol and back by changing the class on a routine.  The hammer should catch that.

> For the icons that changed: did you find alternatives that would be a better fit? We could replace those in the file updater as well.

I don't know.  I'm not really worried about it.  Can't be that many and that big of a deal.  Plus I don't know how to find them without comparing them all visually.  I only noticed the one.

96LawDawg:
All the categories in the add icon picker now match up exactly with what is on Google's icon's page.  All that should be fixed now.

96LawDawg:
I added this list of classes to the JSON Editor.  Not all are available in every context.  And if you use that button, it will replace whatever you had there before, not add to it.  You can have `"classes": "transition material-symbols",`, but using the button will replace everything there.  I'm okay with that.  It isn't that common to have multiple classes and nothing about this prevents manually typing it.

More importantly, is this too many?  This is pretty much all of the standard ones VTT recognizes at this point except for the ones related to chips, checkers, etc.  We could remove all the xyz_font ones.

Classes

96LawDawg:
Two unresolved issues.  First, the file updater works for the game data, but not the meta data. Second, somehow when you are in the meta data and you press the button to add a symbol to How To Play, the room crashes.  I have no idea why.  I didn't touch anything around that.

96LawDawg:
There were a couple of dozen symbols that were completely missing tags.  ChatGPT and I fixed that.

ArnoldSmith86:
> Second, somehow when you are in the meta data and you press the button to add a symbol to How To Play, the room crashes.

Doesn't seem to happen for me. Have a video or something?

96LawDawg:
> > Second, somehow when you are in the meta data and you press the button to add a symbol to How To Play, the room crashes.
> 
> Doesn't seem to happen for me. Have a video or something?


https://github.com/user-attachments/assets/1a94b369-9980-404d-896a-f3e5b29b1ee5

96LawDawg:
The error message point to line 652 in domhelpers.js
```JS
  $('[icon=add_reaction]', controls).onclick = async function() {
    const range = window.getSelection().getRangeAt(0);
    ```

96LawDawg:
I changed the file update at the suggestion of @ArnoldSmith86 to call the v17 update earlier to allow the metadata to change.  When debugging in the console, all the metadata material-icons get changed to material-symbols as expected.  But it doesn't stick.  I have a feeling this has something to do with the fact that the process to save metadata is different than the process to save variants.  So maybe I'm just not seeing what will happen on the actual server.  But I think this is still not working.

I could brute force change the PL games metadata as part of this PR, but that's the point of the file updater.  Also, anything not in the PL wouldn't get caught.

ArnoldSmith86:
Hmm, yeah, didn't think of that. The metadata of the games get copied to the rooms so that loading the Game Shelf doesn't require loading all the game JSON files.

Sounds complicated. 😆

96LawDawg:
> Hmm, yeah, didn't think of that. The metadata of the games get copied to the rooms so that loading the Game Shelf doesn't require loading all the game JSON files.
> 

@ArnoldSmith86, So the last lines in the FileUpdate(state) function at the beginning of the file updater is this:

  ```state._meta.version = VERSION;```

If the file updater is triggered and you save the room state and then download the file and look at the JSON, you see the new version 17.

But if I add this as a test nothing about the name changes.  Just confirming the problem.

  ```state._meta.name = 'File Updater Test';```

If you can point me in the right direction, I'll keep hacking away at the problem.  But I don't even know where to start.  I have figured out the way to pass the changed metadata out of the v17MaterialSymbols(state) and back to the main function.  So it is getting updated; it just is not saving it back to the JSON.

96LawDawg:
Or, and this is completely crazy, but hear me out.  We do what @scruffynerf suggested from the beginning and leave Material Icons in place.  I don't want him getting a big head here, but with the difficulties with the file updater, maybe that is a reasonable solution.  After all we would only be adding a couple of 300kb font files.  We could even get rid of all the unused (in the PL, maybe not in the wild) icon ones like material-icons-twotone.

We would simply deprecate their use.  The JSON Editor would make it easy to click and find symbols now, not icons.  And you could still use them if you put class "material-icons" on your widget.  So they would still available for backwards compatibility.  I'm starting to like this idea a lot.

ArnoldSmith86:
It's probably a lot easier to fix the metadata when actually displaying it..

96LawDawg:
Maybe, but the problem I've been finding is that I'm working on server code.  The toggleClass() is an exported function on the client.  So do I have to rebuild the necessary functions.  Seems like a lot.  I noticed you didn't comment on the deprecation idea??

96LawDawg:
> The code I linked is in client code. Shouldn't be a problem.

But the file updater was in server and didn't see the client code.  Maybe I'm doing it wrong.  I'll keep trying.  But classes started back today so I have a bit less free time.  :-(

> Is there a reason for deprecation if the metadata is working?

No.  Not unless we require picture perfect accuracy on symbols that changed.  And we don't.  The replacements are more than good enough.  But if I can't get the metadata working, I'm at least going back to that so I can make progress on this and get it merge ready.

ArnoldSmith86:
Do you mind if I do the metadata fixing?

96LawDawg:
> Do you mind if I do the metadata fixing?

I would love for you to do that.  You know all of us mortals fail at getting the file updater correct.  And this is entirely new.  I just didn't want to add to your tasks.  But I think this is a really good PR with a lot of benefit, so I would like to see it get done.  If you can push it over the finish line, please do.

ArnoldSmith86:
I'm thinking about how to incorporate the non-filled versions into the icon picker. Apparently there are 2297 icons that have both versions. I'm not sure if it's a good idea to just show both variants of all of those in the icon picker. But maybe it is.. If I search for "home", I'll only get a few results anyway and I would probably be glad if it showed me both of them.

96LawDawg:
There are still a few old icons (using custom made VTT symbols) leftover from the original JSON Editor. Now would be a good time to swap those out so our entire UI uses Google Material Symbols.  Image below shows what we currently have on top and what I propose we use on bottom.  From left to right
 - Copy state from another room
 - Macro
 - Show this widget below
 - Reverse order of F-key shortcuts
 - Duplicate widget
 - Open parent

Screenshot 2025-01-07 233848

Don't worry about the size.  I just did that to get them to line up.  That will be correct.

Thoughts?  The only one I really struggled with was open parent.  We already have "tree" as button on the right-side menu bar and nothing else really fit.  Of course, we could just leave them.  VTT symbols aren't going anywhere.

scruffynerf:
> There are still a few old icons (using custom made VTT symbols) leftover from the original JSON Editor. Now would be a good time to swap those out so our entire UI uses Google Material Symbols. Image below shows what we currently have on top and what I propose we use on bottom. From left to right
> 
> * Copy state from another room
> * Macro
> * Show this widget below
> * Reverse order of F-key shortcuts
> * Duplicate widget
> * Open parent
> 
> 
> Thoughts? The only one I really struggled with was open parent. We already have "tree" as button on the right-side menu bar and nothing else really fit. Of course, we could just leave them. VTT symbols aren't going anywhere.

this is solely about the 3 changed icons.

The robot arm doesn't scream 'macro' to me, ('automation' so I get the idea though).
the replacement looks like a Command Line, which also seems 'unmacro' like.  
A 'keyboard' related icon would convey that it's a shortcut... didn't see one... but that sort of idea.

the Duplicate is MUCH better with the new icon.  The stars didn't say "Duplicate to me" the 'copy' style icon does.

The open parent one is slightly less clear, but I wouldn't have know it was Open Parent from the arrow/line/circle combo either.  
If you want to convey parent, use a parent...  (an apparent parent, to be clear... a mostly dead parent with beautiful plumage is a bit much though).  How about the icons for Supervisor Account (shows 2 figures, one larger behind it, which is what you are really saying...), or Escalator Warning, which shows a Parent holding a child's hand, and clearly you'd expect this to lead to the Parent.

ArnoldSmith86:
I'm pretty sure all of the old ones are from Material Icons. Are they all not in Material Symbols? I don't mind changing the icons; I'm just wondering why so many are missing..

Also: the main toolbar uses VTT symbols as well. So unless we change those as well (and our logo won't be in Material Symbols), the font will be loaded anyway. So the icons you posted won't cause a font download either way.

It would at least make the editor UI icons more consistent but I don't feel like the old icons are standing out.

About the replacements: the macro one is fine but neither the old nor the new one is perfect.. The clone one is probably better. The parent ones are both pretty bad and unfortunately the tree icon is the wrong way around. I'm not sure how I feel about using people to show "parent" either. All those options are probably very similar: once you know, you can remember them but none of them convey the proper meaning at first glance.

I feel like it's better to not change them because at least current users might have gotten used to them. But I don't mind changing them _at all_.

That toolbar above the JSON editor buttons is probably the most out of place thing in the editor right now. It's kinda likely that it'll go away entirely (though the buttons might move somewhere else then). And in the long term the "proper" editor UI should be so good that you don't want to use the JSON editor at all. But all of that might be far into the future...

ArnoldSmith86:
Both variants (fill and nofill) are now available in the symbol picker:

image

Apparently we now have 13288 symbols to pick from.

ArnoldSmith86:
Can `MaterialIconsOutlined-Regular.otf` be removed now or is there a reason that it's still there?

96LawDawg:
> I feel like it's better to not change them because at least current users might have gotten used to them. But I don't mind changing them _at all_.

Then leave it as is.  There are more reasons to leave it alone than there are to change it to satisfy some sort of sense of perfection to have them all be from the same set.  (It was never about eliminating loading the 106kb VTT Symbols font).  Also we'll never agree on the best parent symbol.

scruffynerf:
> > Both variants (fill and nofill) are now available in the symbol picker:
> 
> I don't love this as much as I thought I would. I am glad to see both options, but the duplicates are distracting. Is there a way to have a checkbox or something that would filter for fill and no fill?

I wouldn't, as the difference in some is stark, and not in others. (and some not at all)  And looking for an icon, having to flip the checkbox repeatedly seems awkward.  If there are similar entries but 'inverted' (which is what fill often but not always does, I'd rather see them as separate choices (like now), than have to look harder.

ArnoldSmith86:
I think I would prefer the current version. We can add a checkbox that hides the non-filled version. And if it is not checked, you get the current view.

ArnoldSmith86:
> There is one hopefully unimportant difference that I can't figure out. A single symbol gets replaced by a single symbol no problem. But if they are stringed together (like hieroglyphics) then the spacing is off.

Is this still an issue? Can you post an example room/game?

96LawDawg:
> > There is one hopefully unimportant difference that I can't figure out. A single symbol gets replaced by a single symbol no problem. But if they are stringed together (like hieroglyphics) then the spacing is off.
> 
> Is this still an issue? Can you post an example room/game?

Yes.  Can't do a side-by-side comparison in the same room now of course.  But here is the same widget with a random collection of glyphs at 50px.  The only difference is the first is on production and uses Icons.  The second is in this PR and uses Symbols.

Icons
Symbols

96LawDawg:
> I think I would prefer the current version. We can add a checkbox that hides the non-filled version. And if it is not checked, you get the current view.

What I would like to see is a more feature-filled filter.  When I type "home" into the search bar, I lose all the visual indication of which type of glyph this is.  I almost never want emojis.  Sometimes, I only want the game-icons.net ones.  It would be great to have a filter near the search bar.  Defaults to all, but uncheck and you remove all the emojis and material ones.  But that can be another PR.

With or without the checkbox, I think this PR is good (unless somebody as issue with the spacing of a chain of widgets).

96LawDawg:
You can approximate the spacing in the Symbols glyphs with this css `"letter-spacing": "-0.37em"`.  But as I said before, when I did that, it messed up the placement of single glyphs on some games.  Like Unspeakables.  Maybe I should test that again.

ArnoldSmith86:
Hmm, the issue is that a space character in Material Symbols is as wide as an icon but in Material Icons it's way narrower. Fixing it with `letter-spacing` works but then it breaks when there's _no_ space between two icons.

The only solution I can think of is to wrap the spaces with their own tag `<span> </span>` and then style that differently. That would be quite the pain though. Or to use `letter-spacing` and force either one space or no space between each icon.

The other solution is to not care. It is kinda likely though that somebody somewhere for some reason wanted to put two icons on a button.

If removing spaces is non-destructive, I would probably do that in the file updater. I'm gonna check if I can find an example where removing spaces would change the icons. For example `add task` to `addtask`. If `addtask` is its own icon, it would change. It is `add_task` though so that example would be fine.

ArnoldSmith86:
I found `word-spacing: -0.68em`:

image

96LawDawg:
> I found `word-spacing: -0.68em`:

Not working for me on this example:

Icons
Symbols

96LawDawg:
It is closer.  And word-spacing works better than letter-spacing because if you don't have a space between the letters (something I didn't try), then you end up with the glyphs smooshed on top of each other with letter-spacing.

I think anything we try will be a bit of a hack and imperfect.  We should figure out what word-spacing is best and apply it as the default.  Dev's can then css stuff however they want.

It is really unlikely there are more than 2 or 3 symbols together.  I didn't find a single instance in the PL where there was more than one, but there probably is and I missed it.

96LawDawg:
I found that `word-spacing: -0.75em` worked perfectly on my example and your example and when our two examples were combined.  I also did a couple more examples.  I think the magic number is 0.75.  I'm going to commit that now and do some more testing.

96LawDawg:
Seems to work at -0.75.  Video shows switching back and forth between PR with Symbols and production with Icons.


https://github.com/user-attachments/assets/9f589e09-9d50-4ae9-8e9b-41603775ce9a

ArnoldSmith86:
image

The outlier is Firefox on production for me. So it is better than before. If you don't have that behavior on Windows, even better. :man_shrugging:

ArnoldSmith86 (approved):
We'll test the rest in production... :grin:

PR #2436, opened 2025-01-09
Lines Beta v1 - works, needs some fine tuning on alignment/offsets.

scruffynerf:
Ok, still got some rough edges, but most of it works.
The exact alignment needs some love (it's an offset/sizing issue I need figure out, and most of the problems go away)

You can tie items to the spots using inherit (or other means), so things can ride the lines (except for above alignment issues)

Ideally, this will work with parented and non-parented items, so that's sorta working, but same as above: to be fixed better

The changes in most files are minor, with widgets.js having one function changed I think, and statemanaged.js having a handful.  I've mostly obsoleted inheritFromIsValid() but not entirely, and I need to work on that, so it's a bit duplicative with the other changes.

getNestedValue works, and is most useful for the spotriding, but useful beyond that.

getDefaultValue now works with the new Inherit, uses getNested if need be, and Inheritance now does notify the inheriting widgets (if desired) so that things can update.  Changes will be documented soon.

I'll address anything left unanswered in previous PR. and migrate those questions/issues here soon if need be.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2436/pr-test (or any other room on that server)

To do:

- [ ] Fix whatever is broken in inheritFrom still, most works as expected.  Please report bugs.
- [ ] Add inheritFrom documentation, including all test cases. (I have this)
- [ ] Inherit enable works, but the initial add might not update things, but I cheated with a quick move of one control point, so it all updates. (if you get a set from the overlay).  Nail this down better.
- [x] Spot_N_x properties update now (via OnPropertyChange which is triggered by set() on widget property being inherited)
- [x] Fix reset of items on room reload or if someone else enters room
- [x] Replace Delta handling with OnPropertyChange code.  Actually there is both: Delta handling draws the SVG, while OnPropertyChange calcs+ saves the new locations of line and spots to the widget.
- [x] Calculate size of bounding box - needs work still
- [x] Fix location of selection box for line widget - also overlay selection area/image should be pretty small, but actual placed line could be bigger.  Adjust defaults?  Also add common (TBD) options like 'with all holder spots' etc.
- [x] Update spot coordinates to reflect location of selection box (which represents the line location in the room)
- [x] Calculate slope to determine rotation for riders
- [x] Default css for line, end points, and control points
- [ ] UI Options when adding line from overlay (number of inner control points 0, 1, 2? number of rider spots? enter or select `id` of end points? line color?)
- [ ] Layer management for line widget
- [x] When deleting line, its inner control points should also be deleted. Not sure about the end points?
- [x] Parenting tests/code. For control points, and any rider widgets.
- [ ] How to handle the control point visibility when there are many lines in the room.  Will get crowded and confusing. Maybe a JSON Editor button to turn them all off except for those attached to currently selected line?
- [ ] Possible to automatically make `id` changes in line to the `id` of linked control points?
- [ ] Add "linkedToLine" property to control points to help with the 3 issues above?
- [x] Moving line should move all the control points. Once you make the shape you want, you should be able to grab the line and just move it. Can be done in edit mode by selecting all 4 points, but that is a lot of work.
- [x] AutoSpots = equally spaced automatically.  Manual = false or [0-1 list of spots].  SpotAtStart/AtEnd adds those to Auto
- [ ] Maybe give the spots their own spotCSS (like handleCSS for piles).  Would allow for different shapes, could add background images, etc). Allows for spot by spot customization for widgets (example change size/color/property of items on that spot)
- [x] layout/alignment/offset issues to fix
- [x] spotsLocations currently give the true center point resulting in the upper left corner attachment.  Add an offset.
- [x] maybe 'top/left' versus a 'centerX/centerY' placement coordinate option.
- [ ] variable (by spot) offsets?
- [ ] spot by spot customizations in svg drawing
- [ ] svg spot currently supports 'circle' (radius), 'rectangle'/'square' (2 dimensions), and 'none'. Adding triangle, star, hex/octagon/etc(use size1 as radius and size2 as sides) 
- [ ] transparent holders = spots that just magically allow widgets to land on it. 
- [ ] either allowing only one widget to land on it, or more than one... (spacing/layout?) [LawDawg note: this is all handled in the holder; nothing this PR needs to do]
- [ ] line splitting code:  bezier curve function, can return 2 curves (4 points each) given split location, unimplemented otherwise
- [ ]  1) adjust existing Line to the first set and make a new Line with second (and migrate items to second line), 
- [ ]  OR 2) make 2 new lines, and migrate all items as needed, and remove old line widget
- [ ]  either - given point changes from a middle spot to both an end and a start point. (Add a new widget to hold that info?)
- [ ] Make it easy to do a straight line. Probably in add widget overlay just add line without control points?
- [ ] Adjusting the number of autoSpots does not automatically adjust spotLocations
- [ ] Cloning or duplicating a completed line widget group in the editor does not work
- [ ] Fix max call stack error when renaming a widget in the inheritFrom
- [ ] Add to TestCafe


scruffynerf:
"Moving line should move all the control points. Once you make the shape you want, you should be able to grab the line and just move it. Can be done in edit mode by selecting all 4 points, but that is a lot of work."

Moving line itself as code is now:  allows you to 'move it' but then it's locked back to the original spot by the 4 controlpoints not moving at all.  So if we add OnPropertyChange code when this line widget moves happens to move the 4 control points the same distance as moved, that should work.  So if we shift X -40, then we'd shift those 4 points the same amount.  The 4 inherited ids (each with x/y) can be found by this code, so we have all of the info we need.
The Parent/child thing should be All Good now.  Need to test that: if the control points already relative corrds to the line, the move should just 'work'.

scruffynerf:
Issue from checklist:
How to handle the control point visibility when there are many lines in the room. Will get crowded and confusing. Maybe a JSON Editor button to turn them all off except for those attached to currently selected line?

the following works for cases where the control points are children of the line, but not if they are siblings.
Since I'm hard pressed to think of a good case for the 2 controlpoints to NOT be children of the line (I'd need to account for the relative coordinates in code, but doable for sure), this css plus that might be the best answer.    OR a smarter editor js highlighting (ie if id "ABCD" is made "SelectedInEdit", also add some "make it visible" css class to all ids with "ABCD_whatever".  Not sure the sibling case can otherwise be done in pure css.  But that Json Editor button need those smarts.

```css
/* Hide all .controlPoint elements by default */
.controlPoint {
  border: solid #666;
  background-color: #666;
  display: none;
}

/* Show .controlPoint when body has jsonEdit or edit class AND it OR parent has selectedInEdit */
body.jsonEdit .selectedInEdit > .controlPoint,
body.edit .selectedInEdit > .controlPoint,
body.jsonEdit .selectedInEdit .controlPoint,
body.edit .selectedInEdit .controlPoint {
    display: flex;
}

/* Ensure .controlPoint is visible when hovered, but only in jsonEdit or edit mode */
body.jsonEdit .controlPoint:hover,
body.edit .controlPoint:hover {
  display: flex;
}
```
not 100% perfect.  If you see the c1 and c2 spots, you can then click on one of them... they will highlight as they are now selected, but if you move off of them, they disappear, despite being still selected.

The Start and End points should be fine either way, they are line ends... But if there is a case for making the start+end points also hidden, the same would apply)

96LawDawg:
Making control points children of the line is absolutely the best thing to do.  That also solves the problem of deleting them without needing any special code.

96LawDawg:
Some inheritFrom in the tutorial is broken.  2 things I found so far: in the middle example, the black checker "inherits classes" does not appear at all.  And in the bottom right, the black pawn does not inherit color.

scruffynerf:
> Some inheritFrom in the tutorial is broken. 2 things I found so far: in the middle example, the black checker "inherits classes" does not appear at all. And in the bottom right, the black pawn does not inherit color.

will check those, thanks!

96LawDawg:
Making control points as children of the line presents a new problem.  What you really need to be able to inherit is _absoluteX and _absoluteY of the line.  But even in production, you cannot inherit those read only properties.  I guess ideally be able to inherit read only properties should be made possible?  But I suppose a hacky work-around would be in the line.js to look at and inherit the x/y of the line, but convert that to absolute coords?

96LawDawg:
> the following works for cases where the control points are children of the line, but not if they are siblings. 

> The Start and End points should be fine either way, they are line ends... But if there is a case for making the start+end points also hidden, the same would apply)

I don't think we need to hide the start/end points.  But we probably do need to make them children of the line.  Here's the use case.  When I made a line a child of a basic widget and then moved that basic widget, the line moved but all 4 control points got left behind and then everything was fubar.

Mizokuiam (commented):
Well structured code. Some thoughts on improving performance...

Mizokuiam:
This is an interesting point. Have you considered...

scruffynerf:
known broken from Inherit From Tutorial:
1) Click blue, Black disc changes to match (middle of 2nd column)
2) Pawn changing color button (bottom of right column)

The others work for me as expected.

Interesting note though:
the spinners: if you click on any of the 3 other spinners, they stop inheriting entirely.
That's not mentioned in the tutorial, and probably should be.  Also a demo showing how to avoid that effect, and keep something 'inheriting' should happen.  This might require the 'property sync/entanglement' idea, but it doesn't have to.  Use case example: What if I wanted the upper right Spinner to always trigger the other 3, but not the reverse?  In other words, clicking on Spinner 1, spins 1-4  (they start off in sync and results the same, does that continue to be true?  Or do they start spinning but land on different values.  TBD as to how to better discuss and demo this.

96LawDawg:
> Interesting note though: the spinners: if you click on any of the 3 other spinners, they stop inheriting entirely. That's not mentioned in the tutorial, and probably should be.

Good point.  `inheritFrom` needs an updated tutorial anyway, but I'm going to hold off to see how this PR goes.  FYI, they don't stop "inheriting entirely."  The stop inheriting the properties that got added to their own widget.  So if you click it, the spinner gets its own angle and own value which are added to the widget properties.  Then that breaks the inheritance.  If you want to restart the inheritance, edit the properties on the widget in the JSON editor or programmatically.  You probably already know all that by now, but just saying it out loud in case.

96LawDawg:
Abandoned for #3021

PR #2439, opened 2025-01-11
routine editor

ArnoldSmith86:
A user-friendly visual editor for routines, shown in the properties module of the editor sidebar.

**Events section** (for every widget type): each `*Routine` property appears as a collapsible event card with a friendly label ("click", "game starts", "widget enters", …), a description, and a delete button. "Add Event" offers the predefined events with descriptions plus custom `CALL`-able routines (`fooChangeRoutine`/`fooGlobalUpdateRoutine` patterns supported).

**Routine editor**: operations render as block-style cards with clickable, color-coded parameter chips (variables orange, collections green, widget ids purple, numbers cyan; unset required values highlighted red). Parameters at their defaults are hidden behind a `…` chip. Chips open typed popups (numbers, enums, JSON, widget picking in the room, collections including in-place `[ "id1", "id2" ]` arrays, variables with descriptions, widget properties). Operations can be added, removed and reordered; IF gets structural operand/relation editing, branch folding and add-ELSE; FOREACH gets a range builder / JSON-in / collection popup. Predefined variables display as verbose labels like "player clicking the widget". Works in dark mode.

**Architecture**: a single declarative registry (`routineOperationMetadata`) drives everything — templates, parameter types/defaults/enums, popups, docs and examples. State is cloned only at the widget boundary, own-edit echoes are detected, and per-operation UI state survives re-renders. Covered by a jest suite (`tests/client/routineeditor.test.js`).

Also included: `ROOM:` deep links (`/game/<share>/<name>/ROOM:<room>`) that auto-join a room, used by the info popups' tutorial links.

---

TODO list from meeting:

- [x] blocks
- [x] less details when uninteresting defaults
- [x] show "inplace collections" (`collection: [ "id1", "id2" ]`) that were already used in the routine
- [x] color coding (variables, collections, etc.)
- [x] verbose names for predefined stuff (`player clicking button` instead of `${playerName}`)
- [x] folding
- [x] for predefined variables/collections, show the description directly next to their buttons

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2439/pr-test (or any other room on that server)


96LawDawg:
What am I missing?  I don't see anything different happening when I change the count?

```JSON
"clickRoutine": [
    {
      "func": "MOVE",
      "from": "holder1",
      "to": "hand",
      "count": 1
    }
  ]
  ```

**Comment by ArnoldSmith86** [comment]
Properties panel. Only shows up if the widget has a `clickRoutine`.

**Comment by 96LawDawg** [comment]
Nice start.  Lots of potential.  Things I've thought about before for this kind of UI:

- Using MOVE as an example, changing "to" would be as easy as clicking on a widget in the room, selecting from a drop-down list of available holders, or typing in a name.
- Collections that are available in that routine would also be available through a drop-down list.
- Using SET as an example, when you add the "property" parameter, it would filter a list of the existing properties on that widget.

Just some random musings.  You do what you think and what you can and you'll get feedback, and maybe even some help.

**Comment by ArnoldSmith86** [comment]
Oops. I didn't realize I broke the icons in the tooltips..

**Comment by bjalder26** [comment]
image
It looks like the options should be limited to just object/array for routines.  I think anything else will cause a browser crash.

**Comment by VTT-AI-Agent** [comment]
Picked this branch back up to clear what was open on it:

- **Resolved merge conflicts** (branch showed as `CONFLICTING` against `main`) by merging current `main` in. Conflicts were in `server.mjs` (new `/library/:folder/:plName` route from main vs. new `ROOM:` routes from this branch — kept both), `server/minify.mjs` (both branches added new files to the editor JS bundle list — kept both), and `client/js/overlays/welcome.js` (main's custom-library URL parsing vs. this branch's `ROOM:` capture group — merged both behaviors, including the `roomID` on the public-library return path).
- **Fixed the crash reported above**: the "Routine Editor Demo" section rendered whenever `clickRoutine` was truthy, but `RoutineEditor.onPropertyChange` assumes it's an array and crashes otherwise (e.g. if you set `clickRoutine`'s type to `text`/`number`/`true`/plain object via the generic property editor). Now it only renders/updates for `Array.isArray(clickRoutine)`, matching the same check `widget.js` already uses before executing a routine.

Ran the jest suite (`npm test`) — all passing. No TestCafe coverage exists yet for the routine editor demo, and this change doesn't touch routine execution or room-state semantics, so I didn't regenerate any state hashes.

The bigger TODO list from the PR description (blocks, folding, color coding, etc.) is still open — that's ongoing design work rather than something to "finish" in one pass.

**Comment by VTT-AI-Agent** [comment]
Implemented visual editors for all the remaining routine functions that this demo didn't cover yet:

**New operation editors:** AUDIO, CALL, CANVAS, CLICK, CLONE, DELAY, DELETE, FLIP, GET, INPUT, LABEL, MOVEXY, RECALL, RESET, ROTATE, SCORE, SET, SHUFFLE, SORT, SWAPHANDS, TIMER, TURN, UPLOAD, VAR — following the existing MOVE/COUNT/SELECT pattern (clickable inline parameters with popups). Defaults and enum values are sourced from `widget.js`'s `evaluateRoutine` and the JSON editor's own command list (`jsonedit.js`) so both editors stay consistent.

**New building block:** `RoutineJSONPopup`, a textarea-based JSON editor for parameters that take an arbitrary object/array (`arguments`, `properties`, `variables`, `fields`, `value`, `key`).

**Bug fix:** `RoutineHoldersOrCollectionSourcePopup` (used for every `holder`/`collection`-style parameter, e.g. MOVE's `from`/`collection`) previously wrote to the *first* parameter name even when you picked an existing collection from its "Collections" list — meaning picking a collection for e.g. MOVE actually set `from` to a string instead of `collection`. Fixed so it writes to the correct parameter.

**Wiki content:** added `commonInfoButton()` doc blocks (adapted from the GitHub wiki Functions page) for every function above, plus FOREACH, IF and SELECT, which this PR already implemented but never documented.

Verified: `npm test` passes, and the server still starts cleanly under `minifyJavascript: true` (i.e. nothing here trips up the production minifier). No browser available in this environment to click through the UI, so please give the new operations a try in the PR preview.

**Review by VTT-AI-Agent — COMMENTED** [review]
## TL;DR

Promising direction for the routine editor, and the architecture (per-operation editor classes with templates + typed popups) looks like it will scale well. The most important issue: **routine values and widget ids from room state are interpolated into `innerHTML` unsanitized**, which is a stored-XSS vector since room JSON is untrusted shared content — the rest of the codebase uses `innerText`/DOMPurify for exactly this. Also, parameter popups currently can't be dismissed without picking a value, and the custom-selection mode can leak and hijack normal edit-mode selection.

## Details

### HTML injection from room state (XSS)

- `client/js/editor/controls/routine.js` — `RoutineOperationEditor.render()` builds the operation line with `dom.innerHTML = this.template.replace(..., p => \`<span ...>${this.getDisplayedValue(p)}</span>\`)`. The displayed values come straight from the widget's routine JSON (`collection`, `variable`, `value`, `properties`, …). A game containing e.g. `"collection": ""` executes script as soon as someone opens the properties module on that widget.
- `client/js/editor/controls/widgetselection.js` — `addWidgetEntry()` puts `widget.id` and `widget.get('type')` into `row.innerHTML`; ids are also attacker-controlled.
- `client/js/editor/controls/popup.js` — `addAccordionSection(title, contentHTML)` and the variable/property buttons: the buttons use `textContent` (good), but section content built from state should be treated the same way.

The existing conventions handle this: `SidebarModule.addHeader()` uses `innerText`, and widget-supplied HTML goes through `DOMPurify.sanitize(...)` (e.g. `client/js/widgets/basicwidget.js:45`). Since games are shared via the public library and links, please use `textContent` (or DOMPurify) for anything that originates from room state before this grows further.

### Popups can't be dismissed and leak on reuse (`popup.js`)

- `RoutinePopup.onClick()` is a no-op and there's no outside-click/Escape handler, so once a parameter popup is open the only way to close it is to pick a value. Clicking another parameter just stacks a second popup, and the pending `newRoutineValues()` promise never resolves.
- Popups are reused per editor instance (`this.popups[index]`), but `show()` never clears previous content: reopening the same parameter appends duplicate accordion sections, appends ` - param` to the `<h1>` again, re-registers the change listener (`registerChangeListener` is never cleared), and `Popup.show()` re-adds the click listener each time. A dismiss path that clears/rebuilds the popup DOM and resolves (or rejects) the pending promise would fix all of these at once.

### Custom-selection mode can leak globally (`selection.js` / `widgetselection.js`)

`endCustomSelection()` is only reachable via the "Select" button inside the popup. If the user clicks "Add Widgets"/"Start Fresh" and the popup is then orphaned — e.g. the routine re-renders because someone else changed `clickRoutine` (the `addPropertyListener` in `properties.js` rebuilds everything), or the user gives up (see the dismiss issue above) — the module-level `customSelection` stays set forever: every subsequent selection rectangle in edit mode routes into `setCustomSelection()` with a stale callback and orange highlights, and normal selection is broken until reload. Consider ending custom-selection mode whenever the owning popup is hidden/destroyed. (Small related nit: `customSelection` is initialized to `false` but reset to `null`.)

### Editing `var x = 1` statements is a silent no-op (`routine.js`)

`VarStringRoutineOperationEditor` uses `RoutineStringPopup`, whose `show()` reads `this.operation['variable']` — but the operation is a plain string, so the input displays `undefined`. On change, `RoutineOperationEditor.onNewValue()` hits the object branch and does `Object.assign(this.operation, values)` on a string primitive, which assigns onto a temporary String wrapper and then re-emits the unchanged string. This editor needs its own `onNewValue` that rebuilds the `var <name> = <expr>` string. (I know only MOVE `count` is claimed to work — flagging it because the failure is silent rather than visibly unimplemented.)

### ROOM: deep links are a separate concern (`server.mjs`, `client/js/overlays/welcome.js`)

- This is a standalone user-facing feature (new routes + auto-join) shipped inside the editor demo; per `docs/conventions.md` it would be cleaner as its own PR so it can be reviewed/tested/reverted independently. I did verify the Express 5 `ROOM\\::roomInPath` escaping works (params parse correctly), so the route syntax itself is fine.
- `checkForGameURL()` auto-clicks `#welcomePlayButton` when a `ROOM:` suffix is present, which adds the game to the named room and joins it with zero confirmation. That means any crafted link (`/game/<shareID>/<name>/ROOM:someSharedRoom`) becomes a drive-by "add this game to that room" for whoever clicks it — combined with your own FIXME about re-adding a game to a room that already contains it, it may be worth keeping the overlay visible (pre-filled) instead of auto-clicking, at least outside the tutorial flow.
- The client regex restricts the room to `[A-Za-z0-9_-]+` while the server route accepts more; a room id outside that set silently falls back to the normal welcome flow.

### Minor

- `RoutineOperationPopup.show()` filters via `type.constructor.name != 'UnknownRoutineOperationEditor'` — a string comparison against a class name breaks silently if the minifier ever mangles top-level names; `type instanceof UnknownRoutineOperationEditor` is safer.
- `getDisplayedValue()` for a comma list: `this.operation && typeof this.operation[p] != 'undefined' || this.getDefaults()[p] !== null` — the intent (prefer explicitly-set, else first non-null default) only works because every template happens to put the null-default alternative first; parenthesizing/splitting the two checks would make it robust.
- `routineOperationTypes()` instantiates all 31 editors (each with several popup objects) on every call, and `getMatchingRoutineOperationEditor()` then instantiates the match a second time — fine for a demo, but worth a registry of classes rather than instances eventually.
- `popup.css`'s `.accordion-section` / `.accordion-content` selectors aren't scoped under `.inline-popup`, so they'll style any future accordion elsewhere.
- No tests yet — understandable for the editor demo, but the `welcome.js` URL-parsing change affects production routing for all users and would be cheap to cover with a jest test.

**Review by VTT-AI-Agent — COMMENTED** [review]
## TL;DR

Follow-up pass over the current head (609e2454) — the findings from the previous review (stored XSS via `innerHTML`, non-dismissable/leaky popups, custom-selection mode leaking into normal edit selection, `var x = 1` editing being a silent no-op, ROOM: auto-join concerns) all still apply. This pass adds **two new confirmed bugs**: the widget-picker flow writes corrupted JSON into the routine (numeric keys merged into the operation), and the new `setHighlighted` signature silently clears custom highlights from every pre-existing caller.

## New findings

### Widget picker corrupts the operation JSON (`popup.js` / `routine.js`)

`RoutineWidgetIDPopup.show()` wires its `WidgetSelection` callback as (`client/js/editor/controls/popup.js:284`):

```js
this.holderSelection = new WidgetSelection([], widgets=>{
  this.notifyChangeListeners(widgets.map(w=>w.id));
  ...
});
```

This bypasses `setNewValue()`, so the change listener receives a **raw array** instead of `{ parameterName: value }`. Downstream, `RoutineOperationEditor.onNewValue()` (`client/js/editor/controls/routine.js:103`) hits the object branch and does `Object.assign(this.operation, ['holder1','holder2'])`, which merges numeric keys into the operation. Picking widgets and clicking "Use These Widgets" for e.g. MOVE `from`/`to`, CALL `widget`, RECALL `holder`, or SCORE `seats` therefore saves something like `{"0":"holder1","1":"holder2","func":"MOVE",...}` into the widget's `clickRoutine` — the intended parameter is never set, and the junk keys persist in room state (they survive the `JSON.parse(JSON.stringify(...))` round-trip and are written via `inputValueUpdated`). The callback should call `this.setNewValue(widgets.map(w=>w.id))` so subclass overrides like `RoutineHoldersOrCollectionSourcePopup.setNewValue()` can route the value to the right parameter name.

Related: `new WidgetSelection([], ...)` always starts empty, so editing a parameter that already has widget ids doesn't show the current value — "Add Widgets" then effectively behaves like "Start Fresh".

### `setHighlighted` default clears custom highlights for all legacy callers (`widget.js`)

The new signature is `setHighlighted(isHighlighted, isCustomHighlighted=false)` with `!== null` guards (`client/js/widgets/widget.js:2618`) — clearly designed so `null` means "leave unchanged". But the default for the second parameter is `false`, not `null`, so every pre-existing two-arg-less call now actively clears the custom (orange) highlight:

- `client/js/editor/selection.js:164/167` — any normal `setSelection()` (e.g. clicking a single widget via `editClick`, which is *not* routed through the custom-selection branch) strips orange highlights while `customSelection` still contains those widgets, desyncing the visual state from the pending selection.
- `client/js/jsonedit.js:2218` — runs over **all** widgets whenever JSON-editor highlighting updates, wiping custom highlights room-wide.

Changing the default to `isCustomHighlighted=null` fixes all call sites at once. (Also worth noting: single-widget clicks during custom-selection mode still go through `setSelection()` rather than `setCustomSelection()`, so only rectangle-selection actually feeds the picker — probably part of the known jank, but it compounds with this bug.)

## Still open from the previous review

No new commits have landed since that review, so in particular these remain worth addressing before this graduates from demo status:

1. Routine values / widget ids from room state are interpolated into `innerHTML` unsanitized (`routine.js` `render()`, `widgetselection.js` `addWidgetEntry()`) — stored XSS via shared game JSON.
2. Parameter popups can't be dismissed without picking a value, and reuse leaks listeners/duplicate sections.
3. `endCustomSelection()` is unreachable if the popup is orphaned, permanently hijacking edit-mode selection.
4. `/ROOM:` deep links auto-click the play button with zero confirmation (drive-by room modification) and are a separate concern from the editor per `docs/conventions.md`.

VTT-AI-Agent (commented):
## TL;DR

Follow-up review at head `609e2454` — no new commits since the previous review, whose findings (unsanitized `innerHTML` from room state = stored XSS, undismissable/leaking popups, custom-selection mode leaking globally) all still apply. This pass found **two additional confirmed bugs**: picking widgets via the new widget-selection UI corrupts the operation JSON with numeric keys instead of setting the intended parameter, and the new `setHighlighted()` default silently wipes custom highlights from every pre-existing caller.

## New findings

### 1. Widget selection writes numeric keys into the operation instead of the parameter

`client/js/editor/controls/popup.js` — `RoutineWidgetIDPopup.show()` wires the `WidgetSelection` callback as:

```js
this.holderSelection = new WidgetSelection([], widgets=>{
  this.notifyChangeListeners(widgets.map(w=>w.id));
  ...
```

This bypasses `setNewValue()`, so the raw **array** reaches `RoutineOperationEditor.onNewValue()` (`client/js/editor/controls/routine.js`), which does:

```js
if(typeof values == 'string') { ... } else {
  Object.assign(this.operation, values);
```

`Object.assign(op, ['holder1','holder2'])` copies the array's index keys, producing e.g. `{"0": "holder1", "1": "holder2", "func": "MOVE", ...}` — the `from`/`to`/`holder`/`widget`/`seats` parameter is never set, and the routine JSON is silently polluted with bogus `"0"`/`"1"` properties that persist into the saved game state. Every editor that uses `RoutineWidgetIDPopup` or `RoutineHoldersOrCollectionSourcePopup` (MOVE, MOVEXY, RECALL, CALL, SCORE, FLIP, SET, …) is affected as soon as the user completes a widget selection. Fix: call `this.setNewValue(widgets.map(w=>w.id))` — though note that `RoutineHoldersOrCollectionSourcePopup.setNewValue()` targets the *collection* parameter, so the widget-ID path there needs to target `parameterNames[0]` (`holder`/`from`) instead, and should also clear the competing parameter (see next point).

### 2. Setting one of two alternative parameters leaves the other one active

For dual parameters like MOVE's `{from,collection}` or FLIP's `{holder,collection}`, `RoutineHoldersOrCollectionSourcePopup.setNewValue()` sets `collection` but never clears `from`/`holder` (and vice versa once #1 is fixed). Since the engine prefers `holder`/`from` over `collection` when both are present, picking a collection on an operation that already has `from` set is a silent no-op at runtime while the editor displays the new collection (because `getDisplayedValue('from,collection')` also prefers the first name — so editor display and engine behavior can even disagree). The change payload should set the chosen parameter and `undefined` the alternative, the same way `RoutineOperationPopup.setNewValue()` already clears old keys on a func switch.

### 3. `setHighlighted()` new default erases custom highlights from legacy callers

`client/js/widgets/widget.js:2618` — the signature is now `setHighlighted(isHighlighted, isCustomHighlighted=false)` with `!== null` guards, i.e. `null` means "leave unchanged". But the default for the second argument is `false`, not `null`, so every existing single-argument call — `setSelection()` in `client/js/editor/selection.js:155-158` and the JSON-editor highlighting sweep in `client/js/jsonedit.js:2218` (which runs over **all** widgets) — force-clears `isCustomHighlighted`. Concretely: start a custom selection ("Add Widgets"), then click any widget in the room (`editClick` → `setSelection`) or let the JSON editor refresh its highlights — the orange highlight disappears while the widget is still inside `customSelection`, so the visual state and the selection state diverge. The default should be `null`.

## Still open from the previous review

Briefly, since nothing changed: (a) routine values / widget ids from room state interpolated into `innerHTML` unsanitized in `routine.js` `render()` and `widgetselection.js` `addWidgetEntry()` — stored XSS from shared game JSON; this is the one thing I'd fix before merging anything, draft or not; (b) parameter popups have no dismiss path and accumulate duplicate accordion sections / listeners on reuse; (c) `customSelection` mode can leak globally when its popup is orphaned, hijacking normal selection until reload; (d) editing `var x = 1` string statements is a silent no-op; (e) the `ROOM:` deep-link routes + auto-click of `#welcomePlayButton` are an unrelated production-facing feature that deserves its own PR and a confirmation step.

Given the PR is a draft demo, none of this blocks the direction — the per-operation editor classes with templates and typed popups still look like the right architecture. Findings #1–#3 are just the kind of silent data corruption that's cheap to fix now and painful to debug later.

VTT-AI-Agent:
Thanks for the thorough review — addressed in 69f829bd:

**XSS (fixed):** Parameter values in `RoutineOperationEditor.render()` now go through an `escapeHTML()` helper before interpolation into `innerHTML` (also removed the dead unescaped first assignment), and `WidgetSelection.addWidgetEntry()` builds its cells with `textContent`. Verified with a jsdom test that `"collection": ""` renders as inert text.

**Popup dismissal + reuse (fixed):** Popups now have a real lifecycle — `newRoutineValues()` calls a new `reset()` (clears DOM + listeners) before every open, so reopening no longer duplicates accordion sections/titles/listeners; the click handler is bound once in the constructor so it can't stack. Popups can be dismissed with Escape or by clicking outside, which resolves the pending promise with `undefined`, and the editor skips the update in that case. Outside-click dismissal is suppressed while a custom widget selection is running, since that interaction happens by clicking in the room.

**Custom-selection leak (fixed):** `Popup.hide()` now always calls `endCustomSelection()`, and `endCustomSelection()` is safe to call when idle (guards the null case; also normalized the `false`/`null` init). Any dismissed or orphaned popup releases selection mode instead of hijacking edit-mode selection until reload.

**`var x = 1` editing (fixed):** `VarStringRoutineOperationEditor` now has its own `onNewValue()` that rebuilds the `var <name> = <expr>` string instead of `Object.assign` on a string primitive. `RoutineStringPopup` also shows the current value instead of `undefined`. Bonus fix found while testing: `WidgetSelection` results were emitted as a raw array (which `Object.assign`ed numeric keys onto the operation) — they now go through `setNewValue()`, and holder-or-collection popups route arrays to the holder-like parameter and collection names to the collection parameter (clearing the holder parameter so e.g. MOVE's `from`-wins precedence doesn't silently ignore the chosen collection).

**Minor items (fixed):** `instanceof UnknownRoutineOperationEditor` instead of the class-name string; parenthesized the comma-list default check; scoped the accordion CSS under `.inline-popup`.

**ROOM: deep links (not changed, needs owner input):** I agree it's a separable feature, but it's already part of this branch's history and the editor's own info-popup tutorial links (`tutorial/<name>/ROOM:<roomID>-tutorials`) depend on the auto-join to be one-click, so removing the auto-click here would break the very flow this PR adds it for. I'd suggest @ArnoldSmith86 decides whether to keep it here or split it out; if it stays, gating the auto-click on the `-tutorials` room suffix (or showing the pre-filled overlay otherwise) would address the drive-by concern. The editor registry / jest coverage suggestions are noted as follow-ups.

All verified locally: jest passes, an 11-assertion jsdom test covers the fixes above, and the server still starts under `minifyJavascript: true`.

VTT-AI-Agent:
Re auto-review 4/5: that review ran against the pre-`69f829bd` state, so most of its findings were already fixed in `69f829bd` (see the previous reply for details):

- **Raw array from the widget picker** — already fixed: `WidgetSelection` results now go through `setNewValue()`, keyed by parameter name, so no numeric keys get merged into the operation JSON anymore.
- **Stored XSS via `innerHTML`, non-dismissable/leaking popups, custom-selection leak** — all fixed in `69f829bd`.
- **ROOM: auto-join** — discussed in the previous reply; left for the repo owner to decide.

The one genuinely new finding is valid and now fixed in `b4a26626`: `setHighlighted(isHighlighted, isCustomHighlighted=false)` made every legacy caller (JSON editor highlighting, normal edit-mode selection) clear the custom orange highlight as a side effect. The default is now `null`, which — like the first parameter — means "leave unchanged"; the custom-selection code keeps passing explicit values.

VTT-AI-Agent (commented):
## TL;DR

Reviewed the two new commits since the last review (`69f829bd`, `b4a26626`). They correctly fix everything previously reported — XSS escaping, popup lifecycle/dismissal, the custom-selection leak, the `Object.assign`-on-array JSON corruption, the `setHighlighted` default, and `var x = 1` editing — and jest plus the minify step both pass on this head. What remains is smaller: the shared `Popup.hide()` ends an active custom selection even when an unrelated info popup closes (desyncing the widget-picker UI), and the Unknown-operation editor still writes a bogus `json` key into routine JSON.

## Verified fixes from the previous reviews

- **XSS**: `render()` in `client/js/editor/controls/routine.js` now escapes parameter values via `escapeHTML()`, and `widgetselection.js` builds cells with `textContent`. I checked the remaining interpolation points (accordion titles/content, `setTitle`, `button()`, `h1.append`, `data-parameter` from the `[a-zA-Z0-9,]+` template regex) — all are static strings or text-node APIs. Looks complete.
- **Widget picker corruption**: `RoutineWidgetIDPopup` now routes through `setNewValue()`, and `RoutineHoldersOrCollectionSourcePopup.setNewValue()` targets the holder parameter for arrays and clears the holder (via `undefined`, dropped by the `JSON.parse(JSON.stringify())` round-trip) when a collection is chosen. Confirmed the engine prefers `from` over `collection` (`client/js/widgets/widget.js:1594`), so display and runtime now agree.
- **`setHighlighted(…, isCustomHighlighted=null)`**: legacy single-argument callers (`selection.js:164-167`, `jsonedit.js:2218`) no longer wipe custom highlights.
- **Popup lifecycle**: `reset()` on every open, Escape/outside-click dismissal resolving the promise with `undefined`, and the guarded `endCustomSelection()` all work as described.
- `npm test` (jest) and the `server/minify.mjs` bundle step both pass on this branch.

## Remaining findings

### 1. Closing *any* popup ends an active custom selection, desyncing the widget picker

`Popup.hide()` (`client/js/editor/controls/popup.js:108`) calls `endCustomSelection()` unconditionally, and `InfoPopup` shares that method. Scenario: open a widget-ID parameter → "Add Widgets" (custom selection active) → click an info ("i") button → close the info popup (any click closes an `InfoPopup`). That `hide()` kills the custom selection, but the `WidgetSelection` component still shows its "Select widgets in the room… / Select" view — so the user's subsequent room clicks silently mutate the normal edit selection while the UI claims they're picking widgets. Consider ending the selection only from the popup that owns it, or having `WidgetSelection` observe the cancellation and flip back to its start view.

### 2. Clicking a parameter while its popup is already open closes it instead of reopening

In `render()` (`routine.js:1087`) the span's click handler runs first (target phase), calling `reset()` + `show()` — but the *previous* open's `boundOnOutsideClick` is still registered on `document`, fires when the same click bubbles up, sees the span is outside `domElement`, and hides the freshly-reset popup. Net effect: the popup flashes closed and the second click is wasted. Also, `reset()` clears the first promise's cancel listener, so the first `await newRoutineValues(...)` never resolves (harmless leak, but worth knowing). Removing the outside-click listener inside `reset()`/at the top of `show()` would fix both. Relatedly, Escape closes every open popup at once (info popup *and* the parameter popup beneath it) since each has its own document keydown listener.

### 3. Unknown-operation editor corrupts the operation on edit

`UnknownRoutineOperationEditor` uses template `'{json}'` with a `RoutineStringPopup`, so entering a value sends `{json: "…"}` to the base `onNewValue()`, which `Object.assign`s a literal `json` key into the operation object — polluting the saved routine JSON without changing anything the engine reads. (For string operations it's a silent no-op instead, since `Object.assign` on a string primitive's wrapper is discarded by the JSON round-trip.) Until real editing is implemented, either parse the input as JSON and replace the whole operation (like `VarStringRoutineOperationEditor.onNewValue` now does for its string form), or make this editor read-only.

### 4. Minor

- `RoutineHoldersOrCollectionSourcePopup.setNewValue()` clears `holder`/`from` when a collection is picked, but not the reverse: picking holders leaves a stale `collection` key in the JSON. No behavior divergence (engine and display both prefer `from`), just a leftover key in saved state.
- `RoutineStringPopup.show()` treats string operations as having no current value (`typeof this.operation == 'object'` guard), so editing `var x = 1` presents an empty input instead of prefilling `x` / `1`.

### 5. Still open from earlier reviews

The `ROOM:` deep-link routes (`server.mjs:369-370`) plus the auto-click of `#welcomePlayButton` (`client/js/overlays/welcome.js:58-59`) remain an unrelated, production-facing feature riding along in this PR — it deserves its own PR and probably an explicit confirmation instead of auto-joining a room on page load.

Overall: the review-response commits are solid and the architecture continues to look right for a draft. Findings 1–3 are the ones I'd address next since they affect saved routine JSON or silently hijack selection state.

VTT-AI-Agent:
Follow-up review addressed in b555650c, plus a general polish pass:

**Review fixes:**
- *Closing any popup killed an active custom selection* — fixed: only `RoutineWidgetIDPopup` (the popup that starts widget selections) ends custom-selection mode on hide; info popups and other parameter popups leave it alone. Clicks inside any popup also no longer dismiss other popups, so nested info popups work correctly.
- *Re-clicking an open parameter instantly closed its popup* — fixed: `reset()` now also drops the document-level listeners from the previous `show()`, so the bubbling click can't hit the stale outside-click handler. Opening a *different* parameter now closes the previous popup first (single open parameter popup at a time).
- *Unknown-operation editor wrote a bogus `json` key* — fixed with a new `RoutineFullOperationJSONPopup` that edits the entire operation as JSON and replaces it wholesale.
- Driving the editor in a real browser surfaced one more data bug in the same family: number popups on `{fillTo,count}` wrote the number to `fillTo` — which the engine *prefers over* `count` — so "MOVE 3 widgets" silently meant `fillTo: 3`. They now write the last (normal) parameter and clear the override alternatives.

**Polish pass** (this still looked like a prototype):
- Popups lose the pink placeholder background and now use the editor's shared CSS variables — dark mode works. Operation rows are proper cards with a VTTblue accent; parameters render as clickable chips with ellipsis for long values; accordions/buttons/inputs/info-buttons are styled consistently.
- Sidebar section renamed from "Routine Editor Demo" to "Click Routine"; inline styles moved into CSS.

Verified end-to-end with Playwright against a real room: selecting the button shows the routine, clicking the count chip and choosing a value writes `{"count": 3}` (with `fillTo` cleared) into the room state, Escape/outside-click dismissal works, popups don't duplicate content on reopen, and light/dark screenshots both look right. jest + production minification (`minifyJavascript: true` server start) also pass.

The earlier Firefox `publiclibrary-2.js` CI failure was a Reversi state-hash flake (same test passed in Chrome on the same commit and on the previous commit in both browsers) — I've re-triggered it.

VTT-AI-Agent:
Added an **Events** section to the properties module in 8d0ae7e5, turning the routine editor into a proper top-level UI:

- Every `*Routine` array property on the selected widget appears as a **collapsible event card** with a friendly label ("click", "game starts", "widget enters", …), the raw property name, an info button with a description adapted from the wiki, and a delete button (with confirmation). Expanding a card shows the inline routine editor.
- **"Add Event"** opens a popup listing the predefined events (`clickRoutine`, `doubleClickRoutine`, `changeRoutine`, `enterRoutine`, `leaveRoutine`, `gameStartRoutine`, `globalUpdateRoutine`) with per-event descriptions, plus a **custom event** creator for `CALL`-able routines — names are sanitized and the `Routine` suffix is appended automatically, and the `fooChangeRoutine` / `fooGlobalUpdateRoutine` patterns get generated labels like "foo changed" / "foo changed anywhere".
- To make new/empty handlers actually editable, the routine editor now supports **adding operations** (an "add operation" button opens the operation list with live examples) and **removing operations** (per-card delete icon). IF/FOREACH subroutines render even when empty so you can add operations inside them.
- Routine properties are hidden from the generic property table (they're edited via Events now; the raw JSON module remains for hand editing), and the property list got its own "Properties" header. The Events section appears for **all widget types**, not just the generic ones.

Verified end-to-end in a real browser: expand/collapse, adding `gameStartRoutine` and a SHUFFLE operation into it, creating a custom `dealCardsRoutine`, removing individual operations and whole handlers — all write the expected JSON into the room state. jest and the production-minified server start both pass.

VTT-AI-Agent:
The TODO list from the description is now implemented (commit e1d158ba):

- ✅ **blocks** — operations render as block-style cards with nested blocks for IF/FOREACH (done in the earlier polish pass).
- ✅ **less details when uninteresting defaults** — template segments whose parameters all use defaults are hidden (e.g. MOVE just reads "MOVE 2 widgets from holder1 to hand"); a `…` chip on the operation reveals them.
- ✅ **show "inplace collections"** — arrays of widget ids used as `collection`/`source`/`excludeCollection` earlier in the routine appear as `[ id1, id2 ]` buttons in the Collections section of parameter popups, and picking one writes it to the collection parameter (clearing the competing holder-style parameter).
- ✅ **color coding** — parameter chips are colored by kind: variables orange, collections green, widget ids purple, numbers cyan, function names / plain values blue. Works in dark mode via the shared editor palette.
- ✅ **verbose names for predefined stuff** — chips showing `${playerName}` etc. display friendly labels like "player clicking the widget" (JSON unchanged underneath).
- ✅ **folding** — IF/FOREACH cards get a fold toggle that collapses their nested routines; event handlers themselves collapse/expand in the Events section.
- ✅ **descriptions directly next to predefined variable/collection buttons** — the Predefined Variables/Collections popup sections show a short description under each button instead of one info-only text blob.

Internally, parameter chips now map to their popups via an explicit index so hidden segments can't shift the popup assignment. Everything verified in a real browser against a live room (including that picking an in-place collection writes `"collection": ["w1","w2"]` and clears `from`); jest and the production-minified server start pass.

VTT-AI-Agent:
Per your go-ahead, I rewrote the parts that were structurally weak (210cd26b — net −150 lines despite adding a 20-test jest suite):

**Declarative operation registry.** The 24 hand-written editor classes with fragile parallel template/popups arrays are gone. `routineOperationMetadata` declares each operation once — template, parameters with type/default/enum values/display overrides, defined variables/collections — and one generic `RoutineOperationEditor` derives everything: chips, which popup opens per parameter, defaults (including computed ones like MOVE's count), color coding, and the add-operation examples. Adding a future operation is now ~15 lines of data. Only IF, FOREACH, `var` strings, comments and unknown ops keep small subclasses.

**Reference-based state instead of clone-per-layer.** Cloning happens exactly once, at the widget-state boundary in both directions; nested editors share references and notify upward once. Server echoes of our own edits are detected and skipped, and per-operation UI state (folds, revealed defaults) lives in a WeakMap keyed by operation identity — so folding an IF now survives editing something else. This surfaced a nasty latent bug: handing the editor's live array to `widget.set` aliases widget state to the editor after the local delta echo, which makes **every subsequent edit a silent no-op** (`set()` sees the state already "equal"). Found by driving two consecutive edits in a real browser; fixed and covered by a regression test.

**IF/FOREACH are actually editable now.** IF renders separate operand1 / relation / operand2 chips (switching to a single condition chip when `condition` is set) plus an "add ELSE" button; FOREACH gets a dedicated source popup with a range builder, JSON "in" input, and collections — each clearing the competing parameters. `var` statements that don't match the simple `var x = expression` form fall back to raw-statement editing instead of mangling, and `// comment` strings get their own editor and appear in the add-operation list.

**Tests in CI.** The Playwright/jsdom verification I'd been running manually is now `tests/client/routineeditor.test.js` (20 tests: registry integrity, XSS escaping, optional segments, color classes, verbose labels, IF/FOREACH behavior, echo handling, the aliasing regression, popup routing).

Verified end-to-end in a real browser: sequential edits all persist to room state (IF relation, MOVE count, FOREACH range, comment), fold state survives unrelated edits, and jest + the production-minified server start pass.

VTT-AI-Agent (commented):
## TL;DR

Reviewed the current head (`b7dbe5d5`), which includes the declarative-registry rewrite and the new Events section landed since the last review. The previously-reported bugs (unsanitized `innerHTML`, popup lifecycle/leak, `setHighlighted` default, JSON corruption, `var` editing) are all fixed — escaping via `escapeHTML`/`textContent` is now applied consistently, and both `npm test` (27 passing) and the minify step pass. What remains is smaller: the **widget picker opens empty and can silently overwrite an existing `from`/`to`/holder value with `[]`**, plus a couple of lower-severity behavior/scope notes below.

## Findings

### 1. Widget picker opens empty and can clear the current value — `popup.js:376`, `widgetselection.js:62`

`RoutineWidgetIDPopup.show()` builds its selection with `new WidgetSelection([], …)`, so it never seeds the widget ids already stored in the operation (e.g. `MOVE.from = ["holder1"]`). The "Use These Widgets" button (`widgetselection.js:62-64`) calls `this.callback(this.widgets)` with the current list, which is `[]` until the user goes through the "Add Widgets" / room-selection flow.

Failure case: a user clicks a widget chip to inspect or tweak the current holder, sees an empty table, and clicks "Use These Widgets" — the parameter is overwritten with `[]` instead of the value they meant to keep. Pre-loading the existing ids into `holderSelection` (and/or disabling "Use These Widgets" while the list is empty) would avoid the silent data loss.

### 2. `ROOM:` deep links auto-join without confirmation — `welcome.js:58-59`

When the URL carries a `ROOM:` segment, `checkForGameURL` programmatically fires `$('#welcomePlayButton').click()`, so any `/game/<share>/<name>/ROOM:<room>` link now joins the named room directly rather than showing the linkDetails/join dialog. That's the intent for the info-popup tutorial links, but it also changes the behavior of ordinary shared game links that happen to include a `ROOM:` suffix. Worth confirming this auto-join is desired for all such links, and that the room id — client-side limited to `[A-Za-z0-9_-]+` and dropped straight into `#welcomeJoinRoom` — is validated the same way as normal room ids on the server route (`server.mjs:369-370`).

### 3. Newly added event handlers miss live remote updates — `properties.js:1616-1617`

`renderEvents` registers an `addPropertyListener` only for the `*Routine` properties that exist when the sidebar is first rendered. Adding an event through `AddEventPopup` creates a new `fooRoutine` property but no listener is registered for it, so a concurrent edit to that just-created handler by another player won't refresh the local editor while the sidebar stays open. Low severity (only matters for simultaneous editing of a freshly-added handler), but easy to close by (re)registering listeners after a handler is added, or listening on a broader key.

### 4. Scope: two concerns in one PR

The `ROOM:` deep-link routing (`server.mjs`, `welcome.js`) is logically independent of the routine/events editor and could be split out per `conventions.md` ("one PR = one concern"). Not blocking, just noting for reviewability.

---

Nothing above is a hard blocker; the core editor is in good shape and the escaping/lifecycle regressions from earlier rounds are resolved. Finding 1 is the one I'd address before merge since it can lose user data.

96LawDawg:
We should close this I think as it was implemented elsewhere. @ArnoldSmith86, you agree?

PR #2441, opened 2025-01-11
lock room

ArnoldSmith86:
There was a discussion on Discord https://discord.com/channels/770758631146782780/789103957938012171/1327721753316687972 about cheating on VTT. The platform is meant to play with family and friends and if you cheat there, it is _your_ problem.

However, I had the idea about a Lock Room button that would prevent new players from joining and disable edit mode.

This PR adds this button to the Game Shelf. Should we decide to include this, it probably needs a better spot and definitely needs a confirmation and explanation.

The room unlocks as soon as the last player leaves the room.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2441/pr-test (or any other room on that server)

### DIFF
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -399,7 +399,7 @@ async function loadEditMode() {
 }
 
 window.addEventListener('keydown', async function(e) {
-  if(e.ctrlKey && e.key == 'j') {
+  if(e.ctrlKey && e.key == 'j' && $('#editButton')) {
     e.preventDefault();
     if(edit) {
       $('#editorToolbar button[icon=close]').click();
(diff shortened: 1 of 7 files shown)


96LawDawg:
I wouldn't know any creative ways to cheat like injecting custom css or whatever.  But some thoughts about this:

- I can still open the dev console and inspect the elements.
- The vtt logo on the left above the remaining menu items breaks when clicked because it tried to take you to a game shelf you are not allowed to view.
- Removing the game shelf is problematic because now I can't look at the rules for this game.  Also, I cannot switch to another game in the same room during this play session.

Just some thoughts.

96LawDawg:
Feature replaced by #3018

PR #2443, opened 2025-01-12
fixed card border with `showInactiveFaceToSeat`

### LINKED ISSUE #2442: CSS glitch with `showInactiveFaceToSeat`
In this screenshot, the seat on the left (seat1) has ` "showInactiveFaceToSeat": "seat1",`.  The cards are rounded with a 1px border.  None of that shows up properly to the player in seat1.  This is not a problem on the default deck because it does not use borders or rounding, just images.  That is probably why we did not notice it during testing of #1654.  But this also might be an issue because of something in #1860 or #1861.

Screenshot 2025-01-11 181035

ArnoldSmith86:
Fixes #2442. Not the most elegant because I just copied over a bunch of CSS properties but I think this is the safest solution..

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2443/pr-test (or any other room on that server)

### DIFF
--- a/client/css/widgets/holder.css
+++ b/client/css/widgets/holder.css
@@ -35,4 +35,9 @@
 
 .holder.showCardBack > .card.widget > :not(.active).cardFace, .card.widget.showCardBack > :not(.active).cardFace {
   display: block;
+  overflow: hidden;
+  width: 100%;
+  height: 100%;
+  position: absolute;
+  box-sizing: border-box;
 }


96LawDawg:
Works to fix the existing problem.  I'll test some more card things later to try and see if anything else got messed up with this change.

ArnoldSmith86:
The CSS rule only applies to stuff with class `showCardBack` so it should be impossible that this affects something that doesn't use `showInactiveFaceToSeat` unless someone used that class name on purpose.

But impossible things seem to happen all the time...

96LawDawg (approved):
Approved through test

PR #2446, opened 2025-01-15
Public Library - Add The Longest Run

96LawDawg:
Similar to Roaring 20s. VTT implementation by Shobu.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2446/pr-test (or any other room on that server)


ArnoldSmith86 (approved):
The assets are just single colors that could just use `color` of the card objects but I'm too lazy to change it...

96LawDawg:
> The assets are just single colors that could just use `color` of the card objects but I'm too lazy to change it...

Same thought.

PR #2447, opened 2025-01-15
Add support for resetProperties

96LawDawg:
This was an idea I had after a recent discussion about the resetProperties technique.

This PR does 2 things.  It adds JSON Editor support for resetProperties.  This is the reason I did the PR.  I wanted an easy way to capture the current parent, x, y, and a few other common properties.

The other things it does is adds a RESET function.  This is not absolutely necessary, but it simplifies the process and makes a useful tool available to average users with the click of a button.  I would be okay if we don't implement the function, but I would like the JSON Editor support.

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -90,7 +90,8 @@ export class Widget extends StateManaged {
       gameStartRoutine: null,
       hotkey: null,
 
-      animatePropertyChange: []
+      animatePropertyChange: [],
+      resetProperties: {},
     });
     this.domElement.timer = false
 
@@ -1617,6 +1618,24 @@ export class Widget extends StateManaged {
           if(jeRoutineLogging) {
             jeLoggingRoutineOperationSummary(`'${a.holder}' ${a.owned ? ' (including hands)' : ''}`);
           }
+        }
+      }
+
+      if (a.func == 'RESET') {
+        setDefaults(a, { property: 'resetProperties' });
+      
+        const widgetsToReset = Array.from(widgets.values()).filter(widget => {
+          const propertyValue = widget.get(a.property) ?? {};
+          return Object.keys(propertyValue).length > 0;
+        });
+        for (const widget of widgetsToReset) {
+          const propertyValue = widget.get(a.property) ?? {};
+          for (const [key, value] of Object.entries(propertyValue)) {
+            await widget.set(key, value);
+          }
+        }      
+        if (jeRoutineLogging) {
+          jeLoggingRoutineOperationSummary(`Reset properties for widgets with property '${a.property}'`);
         }
       }
 
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -1107,6 +1107,7 @@ function jeAddCommands() {
   jeAddRoutineOperationCommands('MOVE', { count: 1, face: null, from: null, to: null, fillTo: null, collection: 'DEFAULT' });
   jeAddRoutineOperationCommands('MOVEXY', { count: 1, face: null, from: null, x: 0, y: 0, snapToGrid: true, resetOwner: true });
   jeAddRoutineOperationCommands('RECALL', { owned: true, inHolder: true, holder: null, excludeCollection: null });
+  jeAddRoutineOperationCommands('RESET', { property: 'resetProperties', collection: 'DEFAULT' });
   jeAddRoutineOperationCommands('ROTATE', { count: 1, angle: 90, mode: 'add', holder: null, collection: 'DEFAULT' });
   jeAddRoutineOperationCommands('SCORE', { mode: 'set', property: 'score', seats: null, round: null, value: null });
   jeAddRoutineOperationCommands('SELECT', { type: 'all', property: 'parent', relation: '==', value: null, max: 999999, collection: 'DEFAULT', mode: 'set', source: 'all', sortBy: '###SEE jeAddRoutineOperation###'});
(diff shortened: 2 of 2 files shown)


ArnoldSmith86:
Sounds like part of #1592. The good part.

client/js/widgets/widget.js
```diff
@@ -1620,6 +1621,24 @@ export class Widget extends StateManaged {
...
+        const widgetsToReset = Array.from(widgets.values()).filter(widget => {
+          const propertyValue = widget.get(a.property) ?? {};
+          return Object.keys(propertyValue).length > 0;
+        });
+        for (const widget of widgetsToReset) {
+          const propertyValue = widget.get(a.property) ?? {};
+          for (const [key, value] of Object.entries(propertyValue)) {
+            await widget.set(key, value);
+          }
+        }
```
ArnoldSmith86, line 1636: This should be the same but shorter:

```Javascript
        for(const widget of widgets.values())
          for(const [ key, value ] of Object.entries(widget.get(a.property) ?? {}))
            await widget.set(key, value);
```

96LawDawg: Yes.  Done and tested.

96LawDawg:
No idea why the production environment check fails or how to fix.

96LawDawg:
So the production environment log says "either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch()."  But most functions in the widget.js code don't use try/catch.  Most do use problems.push, but I don't even know if that is necessary here.  I tested for a falsely "property" property.  I tested nonsense scalar, array, and objects.  Nothing makes it crash.  Of course if you put something like "property": "id" in there, it will run on every widget and give strange results, but that can happen with `SET` any time and this is just a specialized `SET`.

client/js/widgets/widget.js
```diff
@@ -1620,6 +1621,16 @@ export class Widget extends StateManaged {
         }
       }
 
+      if (a.func == 'RESET') {
+        setDefaults(a, { property: 'resetProperties' });      
+        for(const widget of widgets.values())
+          for(const [ key, value ] of Object.entries(widget.get(a.property) ?? {}))
```
ArnoldSmith86, line 1627: It probably doesn't like the `??`. Try `||` instead..

I gotta make that work at some point but I think it doesn't make a difference for this PR.

scruffynerf: That makes sense, as I saw the same error on my PR, and also was using ?? (instead of adding the code to get() to allow a given default instead of undefined)

96LawDawg: > || is not a good replacement. It'll fail for 0, false, null, etc.

Changing it got this through the production environment change.  And none of that should matter in this PR.

ArnoldSmith86: > || is not a good replacement. It'll fail for 0, false, null, etc.

`resetProperties` is expected to be an object. For everything else, it just shouldn't crash..

96LawDawg:
I'm working on the tutorial for this.  Using `RESET` for parent is going to give unexpected results to the average player because this just SETs the parent which will ignore all the normal things that dragging and dropping on a holder would do or that MOVEing to a holder would do.  I *think* that is the right outcome, but we could treat "parent" differently.

We could have a second parameter, something like "moveToParent": true that would act more like a move (so you could use offsets, piles, etc.  Thoughts?

client/js/widgets/widget.js
scruffynerf: Yes, to be clearer, that works in this case, but if ?? causes this  test to fail, we might want put my get() default code back, to solve this better for cases where using || isn't ok, and yes, adding other checks/wrappers can always be done but one off fixes add up over time, instead of just fixing the right place once.

scruffynerf:
> I'm working on the tutorial for this. Using `RESET` for parent is going to give unexpected results to the average player because this just SETs the parent which will ignore all the normal things that dragging and dropping on a holder would do or that MOVEing to a holder would do. I _think_ that is the right outcome, but we could treat "parent" differently.
> 
> We could have a second parameter, something like "moveToParent": true that would act more like a move (so you could use offsets, piles, etc). Thoughts?

Why not just make it do the right thing for property 'parent', when the parent is that sort of widget?  If it's a holder etc, don't just blindly set the property, do the move/etc.
I don't see a need to add a second parameter when the function can handle this just fine.

96LawDawg:
> Why not just make it do the right thing for property 'parent', when the parent is that sort of widget? If it's a holder etc, don't just blindly set the property, do the move/etc. I don't see a need to add a second parameter when the function can handle this just fine.

Fair enough.  I was thinking of a situation where I might want to parent something that is not a holder.  But we could just test for whether it is a holder and if it is, MOVE the widget.  Then there would be no way to SET the widget to a particular x/y inside of a parent, but that is a really uncommon situation I think.

96LawDawg:
Never mind. I think this will work. I was testing basic widgets on a holder and forgot to change dropTarget. I’ll keep working on the tutorial and see if there are actually any issues not caused by my own incompetence.

scruffynerf:
> But we could just test for whether it is a holder and if it is, MOVE the widget. Then there would be no way to SET the widget to a particular x/y inside of a parent, but that is a really uncommon situation I think.

I'm still not clear on why the function's logic couldn't handle all of these cases.  "parent: holderwidget?  Ok I won't just I'll move so it correctly sets up the holder relationship(s).  And remember it's a holder, so when I see X property, I can still do the right thing, and set it appropriately."

Yes. Order might matter in that case... that's still easy to document and even flag ('hey, this order is incorrect, I should see parent first' or even just fix it by 'if hasproperty', and deal with those first)

96LawDawg:
> I'm still not clear on why the function's logic couldn't handle all of these cases.

It does.  That's what I was trying to clarify when I realized I was using a poorly designed test.  Everything works as I think it should as is.

> Order might matter in that case... that's still easy to document and even flag ('hey, this order is incorrect, I should see parent first' or even just fix it by 'if hasproperty', and deal with those first)

Order does matter if you set both parent and x/y.  That is demonstrated in the tutorial, will be documented, and is the best way to allow for designer choice without forcing our choices or making the function more complicated.  IMHO.

I think this is ready for review and approval.

ArnoldSmith86 (approved):

PR #2449, opened 2025-01-15
Public Library - Add Happy Landings

96LawDawg:
Similar to Havalandi.  VTT implementation by dexdouglas.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2449/pr-test (or any other room on that server)


ArnoldSmith86 (approved):
Would be better to have more examples in the public library that use the included game-icons instead of duplicating them but we can change that at a future point..

96LawDawg:
> Would be better to have more examples in the public library that use the included game-icons instead of duplicating them but we can change that at a future point..

Yes, but for devs trying to import games from PCIO builds and make a new game fast and easy ...

ArnoldSmith86:
:100:

PR #2462, opened 2025-01-26
Add random parameter to `SELECT`

96LawDawg:
SELECT always choose widgets in the same order.  Often I want a random selection but don't want to shuffle.  This PR adds a new parameter "random" that defaults to false so as not to change existing behavior. But if set to true, it uses a Fisher-Yates shuffle to select a random collection.  Of course, this only matters if the "max" is less than the total number of widgets in the collection.

This PR includes an update to the SELECT tutorial, Basics variant.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2462/pr-test (or any other room on that server)


ArnoldSmith86:
You should be able to use https://github.com/ArnoldSmith86/virtualtabletop/blob/main/client%2Fjs%2Fdomhelpers.js#L138 instead.

96LawDawg:
I thought about that, but assumed it actually messed with the collection in the dom.  Probably wrong so I'll try that.

96LawDawg:
I was confused.  I was thinking about using shuffleWidgets https://github.com/ArnoldSmith86/virtualtabletop/blob/e8420915fdc856c10db6738a741035993cde5f95/client/js/main.js#L241.  It has "bringToFront."  But the dom helper works just fine.

client/js/widgets/widget.js
```diff
@@ -1720,6 +1720,10 @@ export class Widget extends StateManaged {
             return w.get(a.property) === a.value;
           });
 
+          if (a.random) {
+            c = shuffleArray(Object.values(c));
```
ArnoldSmith86, line 1724: Does the `Object.values` serve a purpose?

96LawDawg: It didn't work for me otherwise.  The function is looking for an array and c is an object and this was my effort to turn it into an array.  I'm sure I missed something obvious.

ArnoldSmith86: Hmm, it really shouldn't be an object. Apart from the fact that every array is also an object.

I can have a look tomorrow..

client/js/domhelpers.js
```diff
@@ -138,6 +138,7 @@ export function domByTemplate(id, obj, type='div') {
 export function shuffleArray(array) {
   const isString = typeof array === 'string';
   array = [...array];
+  console.log(array);
```
96LawDawg, line 141: I forgot to delete this.  Will do later.

client/js/widgets/widget.js
96LawDawg: Of course it works now.  I must not have set up my test room properly.  Fixed.

96LawDawg:
This was not working with piles, so I moved the shuffle to after the piles were resolved.  Everything tested and working now,  I also tested modes add, intersect, and remove.  Got the expected results.

ArnoldSmith86 (commented):
I don't think calling it twice makes a difference, does it? Did you do that on purpose?

96LawDawg:
> I don't think calling it twice makes a difference, does it? Did you do that on purpose?

No.  I'm just an idiot.  I intended to remove it.  Now it is removed.  And tested again.

ArnoldSmith86 (approved):

PR #2477, opened 2025-02-14
Change Quick Save to Local Time

### LINKED ISSUE #2473: Quick Save Time Recorded in GMT Instead of Local Time
When performing a quick save, the recorded time is stored in GMT instead of the local time of the machine. For example, my local time is GMT+1, but the quick save timestamp still shows GMT.

96LawDawg:
Fixes #2473

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2477/pr-test (or any other room on that server)

### DIFF
--- a/server/room.mjs
+++ b/server/room.mjs
@@ -862,7 +862,7 @@ export default class Room {
     if(mode == 'addState')
       this.saveCurrentState_write(Math.random().toString(36).substring(3, 7), 0, { name });
     if(mode == 'quickSave')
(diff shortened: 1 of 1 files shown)


ArnoldSmith86 (commented):
I don't think this can work because it is in server code. Only the client knows what timezone it is in..

96LawDawg:
You must be correct. Explains why it worked on my local machine but not on the test server. Closing.

PR #2492, opened 2025-03-18
fixed players overlay on small screens

### LINKED ISSUE #2471: Players tab broken on iPhone
Just shows the Invite link but no details on player or ability to change name or color.

ArnoldSmith86:
Fixes #2471. There are many things I want to improve about the Players overlay but this should at least make it usable on mobile again for now.

### DIFF
--- a/client/css/overlays/players.css
+++ b/client/css/overlays/players.css
@@ -1,5 +1,4 @@
 #playerList {
-  overflow: auto;
   display: flex;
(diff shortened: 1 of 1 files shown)


96LawDawg (approved):
Approved through test.  And it seems obvious now that you did it that this was not a mobile problem, but the size of the window problem that would also have been present on PC if I had checked.

PR #2493, opened 2025-03-18
fixed crash for card face without objects

ArnoldSmith86:
If a face in `faceTemplates` does not contain an `objects` entry (or it is not an array), the client crashes.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2493/pr-test (or any other room on that server)

### DIFF
--- a/client/js/editor/sidebar/assets.js
+++ b/client/js/editor/sidebar/assets.js
@@ -56,9 +56,10 @@ function getAssetTargetSize(asset, originalWidth, originalHeight) {
 
             if(key)
                 for(const [ t, template ] of Object.entries(deck.get('faceTemplates')))
-                    for(const [ o, faceObject ] of Object.entries(template.objects))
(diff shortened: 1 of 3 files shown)


96LawDawg:
This does not seem to fix the problem.  To test: add an empty deck from the pick widget overlay.  Change the first faceTemplate object to {} or remove it entirely. Then put the cursor on cardType and click the blue card type template button.  Client crashes.

ArnoldSmith86:
Different crash for the same reason. Also fixed.

ArnoldSmith86:
Hide whitespace changes and it should just add the same condition three times..

96LawDawg:
Wow.  I didn't know that was a thing.  It looked like a lot more than it was, which was surprising.  Now I see it is one line checking for an array.  That is cool.  And why doesn't Git make that obvious and easy to change.

PR #2495, opened 2025-03-18
Public Library - Improve Crossed dice visibility

ArnoldSmith86:
When clicking the dice, you could no longer see which face was active. It was fine but why not keep it visible?

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2495/pr-test (or any other room on that server)


96LawDawg (approved):
Approved.

Issue #2453, opened 2025-01-18
MOVE collection to non-holder crashes client


96LawDawg:
To duplicate, `SELECT` a widget.  Then MOVE it using just the parameter "to" and make sure that it not a holder. 

Also, but unrelated, if you MOVE a widget in the same way, but to itself, you get a different error and crash.

PR #2496, opened 2025-03-18
fixed crash when collection to `MOVE` contains the target

ArnoldSmith86:
There was no check so this resulted in recursive parents and stack overflow crashes.

It's easy to do something like this by accident:

```JSON
  "clickRoutine": [
    {
      "func": "SELECT"
    },
    {
      "func": "MOVE",
      "to": "pn1z"
    }
  ]
```

And if `pn1z` has no `parent`, it would already crash everything...

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2496/pr-test (or any other room on that server)


96LawDawg:
This fixes the second part of #2453.  Could it be made to fix the first part (where moving a collection to a non-holder crashes the client)?

PR #2498, opened 2025-03-18
Add stackOffsetAuto to Holders

### LINKED ISSUE #2497: StackOffset feature request
[From Jonathan on Discord](https://discord.com/channels/770758631146782780/772855222023618590/1351656204950179850):

Not sure if this is the right place for it, but the discussion about PCIO card shifting reminded me of a super common thing I have to awkwardly work into my mods: Dynamic StackOffsetX or Y. PCIO does do this, where widgets in a holder won't exceed the holder, they simply overlap more and more as new widgets are added. Right now I do it with the useful snippet from the WIKI (https://github.com/ArnoldSmith86/virtualtabletop/wiki/Useful-code-snippets#Auto-adjust-stackOffsetX-1), but it's always an annoying chunk of time messing with it. Lots of changing a number, seeing if that worked, change it abit more, see if that worked. If this was automated somehow that'd be amazing, especially for people who are less experienced with code editing. It feels like auto Stackoffset should be a default part of holders.

96LawDawg:
Fixes #2497.

Adds two new properties to holders that default to false.  `stackOffsetAutoX` and `stackOffsetAutoY`.  If set to true, then it will force widgets to remain inside the boundary of the holder.

Have not tested it with multiple different sized widgets, but users probably shouldn't do that.

### DIFF
--- a/client/js/widgets/holder.js
+++ b/client/js/widgets/holder.js
@@ -23,6 +23,8 @@ class Holder extends Widget {
 
       stackOffsetX: 0,
       stackOffsetY: 0,
+      stackOffsetAutoX: false,
+      stackOffsetAutoY: false,
       borderRadius: 8
     });
   }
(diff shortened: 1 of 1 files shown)


96LawDawg:
This seems to work as expected for +X, +Y, -X, and -Y values.  It works in hands.  It works (but looks stupid) with different sized widgets in a holder.

Jonathan suggested in a private message that we make stackOffsetAutoX and Y scalar instead of boolean and that value become the offset.  In some ways that makes sense because you either have a value that is fixed or you have a value that starts at what you set and adjusts as necessary and you don't have to mess with two properties.  However, the problem is that if you leave stackOffsetX as 0, then multiple other parts of the code rely on that for doing things.  So either we change all of that or we leave it how it is.  I say leave it.  It makes sense to me to set a property to true to start the auto adjusting.

96LawDawg:
If this moves towards approval, I will update the Holders: Offset Basic and Offset Advanced tutorials.

96LawDawg:
As an alternative to Jonathan's original suggestion, he also suggests that if stackOffsetAutoX is true but stackOffsetX is 0, then the auto function would look for the widest widget in the holder and use that.  But in testing I could not get that to work.  Everything formed piles when stackOffsetX was 0.

gitran:
Does this overlap with PR 2708?

PR #2500, opened 2025-03-21
`RECALL` and `SHUFFLE` new parameters

96LawDawg:
This adds a couple of things to `SHUFFLE` that could be really useful.  And a couple of other things that, when coupled with the update to `RECALL`, have some limited applicability but would be nice for my Card Games mod.

`SHUFFLE` gets 2 new parameters: `mode` and `modeValue`.  For `mode` the options are:
 - Traditional: the Fisher-Yates shuffle we have now and this is the default value.
 - Seeded: shuffle and get the same result every time.  Will be useful in testing and could be useful in some gameplay if you want to play the same set-up multiple times to get a better score.
 - Reverse: not really a shuffle, but reverses the order of the cards.
 - Riffle: simulates interleaving by splitting the array into about half and then merging them back together based with some randomness. Mimics the natural variation of a physical riffle shuffle.
 - Overhand: simulates the overhand technique by slicing the array into a few small random packets and then reversing the order of the packets.  Mimics rearranging the cards in chunks similar to an overhand shuffle.

For `modeValue`, it is ignored for traditional and reverse.  For seeded, this is the seed number used in the Park–Miller “minimal standard” linear congruential generator (LCG).  For riffle and overhand, `modeValue` is the number of times the riffle or overhand shuffle is performed.  More times = more randomness.

`RECALL` gets a new parameter: `byDistance` which is boolean and defaults to false.  If true, cards are returned to the holder in order of proximity to the holder.  Used in connection with riffle or overhand, this can simulate the clumping of cards expected in some games.

Demo is in the room.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2500/pr-test (or any other room on that server)

----
**New wiki for RECALL**

### [Click to see tutorial](https://virtualtabletop.io/tutorial/functions-recall)

This function moves cards back into the holder they are associated with via _deck_. All cards belonging to that deck will be moved to the holder. _Do not use this operation on holders that have no deck._ Cards are recalled in the order they appear in the DOM.

Note that cards are the only widgets with a deck property. In case you add this property to other widgets, they would be affected by this operation as well.

Parameters:
* excludeCollection: _collection_ - specifies a collection of cards that will not be subject to recall (default: null).
* holder: _holderID_ (or an array)- specifies the holder that contains the cards to recall.
* inHolder: _true/false_ - determines whether cards that are not in holders are affected as well. The default is _true_ causing all cards to be recalled. If set to _false_, cards that are in holders will not be recalled.
* owned: _true/false_ - determines whether cards owned by a player (the "owner" property is not null) are affected as well. The default is _true_, causing all cards of the deck to be recalled. If _false_ then cards owned by players are not recalled.
* byDistance _true/false_.  Defaults to _false_ meaning that cards are recalled in the order they appear in the DOM. If _true_, cards are returned to the holder in order of proximity to the holder.

```JS
{
  "func": "RECALL",
  "holder": "<holderID>",
  "owned": false
}
```

----
**New wiki for SHUFFLE**

### [Click to see tutorial](https://virtualtabletop.io/tutorial/functions-shuffle)

This function "shuffles" widgets, i.e. randomizes their stacking order (_z_ position)

If widgets with various rotations and types happen to be in a holder, they will simply end up in a different order but not re-aligned. It is recommended to use this operation on holders that auto-align their content. If the holder parameter is a seat, then only the widgets belonging to that seat's hand will be shuffled. The hand is usually a holder with `id` of "hand", but that can be changed on the seat.

Parameters:
* holder: _holderID_ or _seatID_ (or an array) - specifies the holder that contains the widgets to shuffle
* collection: _collection_ - specifies the collection of widgets to shuffle (defaults to DEFAULT collection)
* mode: _overhand, reverse, riffle, seeded, true random_ - specifies the type of shuffle to be used (defaults to _true random_)
     * overhand: simulates the overhand technique by slicing the array into a few small random packets and then reversing the order of the packets
     * reverse: reverses the order of the widgets
     * riffle: simulates interleaving by splitting the array into about half and then merging them back together with some randomness
     * seeded: uses a true random shuffle technique, but given the same starting order of the same widgets and the same seed number, the result will be the same
     * true random (default): a fully randomized Yates-Fisher shuffle
* modeValue: _number_ - used by some of the _mode_ parameters as follows (default: 1):
     * overhand: the number of overhand shuffles
     * reverse: _modeValue_ is ignored
     * riffle: the number of riffle shuffles
     * seeded: the seed number
     * true random: _modeValue_ is ignored

```JS
{
  "func": "SHUFFLE",
  "holder": "<holderID>"
}
```


ArnoldSmith86:
`Needed twice for reasons`

ArnoldSmith86:
This seems to remove `RESET`..

96LawDawg:
> `Needed twice for reasons`

I don't know why.  It only works if either the function is called twice or the bringToFront index is reversed.  If I do the bringToFront thing, that will mess up all the hash values for the TestCafe (I think) unless I have an individual bringToFront on each case.  The reason has something to do with the math that is going on behind the scenes versus what is actually displayed on the screen.

96LawDawg:
> This seems to remove `RESET`..

Oops.  I pasted code over too much.  Fixed.

ArnoldSmith86:
This is pretty cool. I wanna review it a bit more. One thing I'd like to change is `traditional`. When compared to the other values of `mode`, it sounds like this is some old Chinese way of shuffling cards. :grin: Maybe we call it `random`, `randomize` or `true random` or something like that?

I get that you mean that this is the way VTT did it traditionally, but if you don't know that, it's kinda odd.. :smile:

96LawDawg:
> One thing I'd like to change is `traditional`. Maybe we call it `random`, `randomize` or `true random` or something like that?

Sure.  I like `true random` except that it is 2 words.  Maybe `trueRandom` for name consistency or just `random`.  But I guess it is variable names and not values that are formatted that way.  It will be the default so 'true random' should be okay.

ArnoldSmith86:
Made a few changes but I'm not sure why you had `fisherYatesSeeded()` twice in there. I don't think I changed anything and removing the second one seems to work just fine.

The second `reverse` was needed because holder children were already given reversed. But giving a `collection` wasn't reversed so it was only working for one of them.

---

Oh, right, I broke TestCafé by reversing stuff...

96LawDawg:
> Made a few changes but I'm not sure why you had `fisherYatesSeeded()` twice in there. I don't think I changed anything and removing the second one seems to work just fine.
> 

It doesn't work for the bottom card.  Look at my test room in https://test.virtualtabletop.io/PR-2500/pr-test.  Do RECALL normal.  Then the shuffle button and seeded.  Change the value to whatever you want and every time, the Ace of Clubs is the last card in the holder.

> Oh, right, I broke TestCafé by reversing stuff...

Yes.  I tried to avoid that.  The "right" answer is to update all of the TestCafe, but I don't want do that.

ArnoldSmith86:
> Change the value to whatever you want and every time, the Ace of Clubs is the last card in the holder.

It does work for higher seed values (like 100000 or something). The problem is that the seeded random function basically returns 0 the first time for low seed values.

I replaced it by the implementation that we also use for the seeded random in TestCafé tests.

Let's see if TestCafé works again and then this might be done. We'll see. :grin:

96LawDawg:
> It does work for higher seed values (like 100000 or something). The problem is that the seeded random function basically returns 0 the first time for low seed values.

Has anybody told Park-Miller their math sucks?  Looks to me from testing that this is all working as intended without the hacky work-arounds.  Thanks!

ArnoldSmith86 (approved):

96LawDawg:
I'll work on the tutorial changes as part of this PR.

96LawDawg:
Tutorials done.  Wiki changes added to first entry above.  Will wait about 24 hours before merging.

96LawDawg:
I tested `byDistance` using Cards Against Humanity.  Time to RECALL normally was about 150-250 ms.  With `byDistance` it went up 50-100 ms.  Not much of a difference.

PR #2503, opened 2025-03-22
save cursor position in editor per widget

ArnoldSmith86:
I knew I wanted to implement this for a while but when I was working on Runway Roll, it was driving me crazy enough to do it.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2503/pr-test (or any other room on that server)


96LawDawg (approved):
Nice.  With a button like "Scenario Generation Button" no wonder.  Approved through test.

PR #2507, opened 2025-03-25
Public Library - Add 100 and 130 Foundations Tutorials

96LawDawg:
This is the beginning of a creation of a new series of tutorials that will appear at the beginning of the list of tutorials.  When completed, this initial phase will consist of:

100 Tutorials Overview
110 Widget Types
120 Widget Properties
130 Variables and Collections
140 Functions Overview
150 Dynamic Expressions
160 Computations Operations

This PR adds the 100 and 130 tutorial.  They are numbered such that there is room for expansion.  The intent is that newcomers would view each of these tutorials in order before diving deeper into the existing tutorials.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2507/pr-test (or any other room on that server)


ArnoldSmith86:
image
Maybe something like that for code examples since you're already using `html` anyway?

ArnoldSmith86 (approved):
Very nice!

PR #2514, opened 2025-03-26
Public Library - changed a few links to assets

ArnoldSmith86:
Looks like we missed a few. I only looked for `"image": "http` so I might not have caught all of them.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2514/pr-test (or any other room on that server)


96LawDawg:
Any easy way for me to look for this when reviewing a PL submission?  Other than searching for image: http.

96LawDawg (approved):
Approved.

ArnoldSmith86:
I usually check the Assets sidebar that should list any http assets for conversion. It doesn't show the library image but that's an easy "right click open in new tab" thing.

PR #2515, opened 2025-03-26
added Docker image without library

ArnoldSmith86:
This PR _should_ (hard to test) create a second tag for the official Docker image that excludes the public library. If somebody does not want to use the library, it is a lot of unnecessary data.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2515/pr-test (or any other room on that server)


96LawDawg (approved):
You had me at "hard to test."  Blind approval.

PR #2516, opened 2025-03-26
Public Library - compress images in Flight

ArnoldSmith86:
Feels like we had 5 PRs already that tried to do this but they usually did too many things at once and everybody was scared to merge them.

This PR basically only compresses the bird images and a few background images. Nothing with text on it and a few other things also looked worse when compressed.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2516/pr-test (or any other room on that server)


96LawDawg:
> What told you?

File changed tab.

Screenshot 2025-03-26 084405

ArnoldSmith86:
OK but that doesn't tell you the size changes because they all got new names. So these are separated removals and additions..

96LawDawg:
> OK but that doesn't tell you the size changes because they all got new names. So these are separated removals and additions..

Ah.  Okay.  That makes more sense.

PR #2518, opened 2025-03-26
Public Library - compress images in Interrogation

ArnoldSmith86:
All those faces make Interrogation the biggest game in the public library now that Flight is a lot smaller. The faces are definitely more blurry after the compression but you probably do not have to see each and every pore to win the game...

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2518/pr-test (or any other room on that server)


96LawDawg (approved):
Approved

PR #2522, opened 2025-03-29
Public Library - Add Castellers

96LawDawg:
Similar to Towers of Hamburg. VTT implementation by Shobu.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2522/pr-test (or any other room on that server)


ArnoldSmith86 (approved):
This should probably not have `en-US`, should it? There's no text in the actual game.

96LawDawg:
> This should probably not have `en-US`, should it? There's no text in the actual game.

Fixed.

PR #2523, opened 2025-03-29
fixed empty `display` for seats

ArnoldSmith86:
When we introduced `display` in #2167, we did not notice that seats already had a `display` property.

This lead to `display: ""` on seats changing from showing a blank seat to hiding the seat completely.

This PR should make them blank again.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2523/pr-test (or any other room on that server)


96LawDawg:
I already updated the wiki

PR #2525, opened 2025-03-29
added new widget type `zoomy`

ArnoldSmith86:
Alright, maybe `zoomy` is not the final name. The problem with experimenting with a new widget type is that the very first thing you have to do is come up with a name.

The idea: it is a widget with a zoom button. When you press the button, you toggle between a zoomed state and a normal state. But: only on your client. So you can have a miniaturized area in your room that players can zoom into and interact with without forcing other players to zoom into it as well.

Initial tests suggest that this is somwhat working. Play around with it and give feedback, please.

### DIFF
--- a/client/js/widgets/zoomy.js
+++ b/client/js/widgets/zoomy.js
@@ -1,0 +1,73 @@
+import { Widget } from './widget.js';
+
+export class Zoomy extends Widget {
+  constructor(id) {
+    super(id);
+
+    this.isBig = false;
+    this.addDefaults({
+      movable: false,
+      layer: -2,
+      typeClasses: 'widget zoomy',
+
+      zoomedX: 0,
+      zoomedY: 0,
+      zoomedScale: 2,
+      zoomedRotation: 0
+    });
+
+    this.domElement.innerHTML = `
+      <button class="zoomy-button">+</button>
+    `;
+
+    $('.zoomy-button', this.domElement).addEventListener('click', () => {
+      this.isBig = !this.isBig;
+      this.updateScale();
+    });
+  }
+
+  cssTransform() {
+    const x        = this.isBig ? this.get('zoomedX'       ) : this.get('x'       );
+    const y        = this.isBig ? this.get('zoomedY'       ) : this.get('y'       );
+    const scale    = this.isBig ? this.get('zoomedScale'   ) : this.get('scale'   );
+    const rotation = this.isBig ? this.get('zoomedRotation') : this.get('rotation');
+
+    let transform = `translate(${x}px, ${y}px)`;
+
+    if(rotation)
+      transform += ` rotate(${rotation}deg)`;
+    if(scale != 1)
+      transform += ` scale(${scale})`;
+
+    return transform;
+  }
+
+  cssTransformProperties() {
+    const properties = super.cssTransformProperties();
+
+    properties.push('zoomedX', 'zoomedY', 'zoomedScale', 'zoomedRotation');
+
+    return properties;
+  }
+
+  set(property, value) {
+    if(this.isBig) {
+      if(property == 'x')
+        this.set('zoomedX', value);
+      else if(property == 'y')
+        this.set('zoomedY', value);
+      else if(property == 'scale')
+        this.set('zoomedScale', value);
+      else if(property == 'rotation')
+        this.set('zoomedRotation', value);
+      else
+        super.set(property, value);
+    } else {
+      super.set(property, value);
+    }
+  }
+
+  updateScale() {
+    this.targetTransform = this.domElement.style.transform = this.cssTransform();
+  }
+}
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -1124,6 +1124,7 @@ function jeAddCommands() {
   widgetTypes.push(jeAddWidgetPropertyCommands(new Seat(), widgetBase));
   widgetTypes.push(jeAddWidgetPropertyCommands(new Spinner(), widgetBase));
   widgetTypes.push(jeAddWidgetPropertyCommands(new Timer(), widgetBase));
+  widgetTypes.push(jeAddWidgetPropertyCommands(new Zoomy(), widgetBase));
 
   jeAddRoutineOperationCommands('AUDIO', { source: '', maxVolume: 1.0, length: null, player: null, silence: false, count: 1 });
   jeAddRoutineOperationCommands('CALL', { widget: 'id', routine: 'clickRoutine', return: true, arguments: {}, variable: 'result' });
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -385,7 +385,7 @@ async function loadEditMode() {
       uploadAsset, _uploadAsset, mapAssetURLs, pickSymbol, selectFile, triggerDownload,
       config, getPlayerDetails, roomID, getDeltaID, widgets, widgetFilter, isOverlayActive,
       html, formField,
-      Widget, BasicWidget, Button, Canvas, Card, Deck, Dice, Holder, Label, Pile, Scoreboard, Seat, Spinner, Timer,
+      Widget, BasicWidget, Button, Canvas, Card, Deck, Dice, Holder, Label, Pile, Scoreboard, Seat, Spinner, Timer, Zoomy,
       toHex, contrastAnyColor,
       asArray, compute_ops,
       eventCoords
--- a/client/js/serverstate.js
+++ b/client/js/serverstate.js
@@ -82,6 +82,8 @@ export function addWidget(widget, instance) {
     w = new Spinner(id);
   } else if(widget.type == 'timer') {
     w = new Timer(id);
+  } else if(widget.type == 'zoomy') {
+    w = new Zoomy(id);
   } else {
     w = new BasicWidget(id);
   }
--- a/server/minify.mjs
+++ b/server/minify.mjs
@@ -72,6 +72,7 @@ export default async function minifyHTML() {
     'client/js/widgets/seat.js',
     'client/js/widgets/spinner.js',
     'client/js/widgets/timer.js',
+    'client/js/widgets/zoomy.js',
 
     'client/js/main.js'
   ]);


96LawDawg:
Will have to think more about the button.  It works for normal sized widgets, but if you make the zoomster (zoomerini, zoomtastic) the size of the room, it is not intuitive.  If you use the zoomy to shrink below scale = 1, the button gets small.

ArnoldSmith86:
Yeah, the button is just a placeholder. We can do that however we like.

ArnoldSmith86:
> need a property to show when zoomed in (so changeRoutines can work)

The idea is that the zoomed in state is different for each player. Properties always have to be the same for every player. So I'm not sure how that can work.

ArnoldSmith86:
> For a and b, I like the idea of a hotkey. But that would not work for c.

We can probably just use the `hotkey` property and make it click the zoom button for `zoomy` instead?

Also: should zooming into one reset any others?

96LawDawg:
> > need a property to show when zoomed in (so changeRoutines can work)
> 
> The idea is that the zoomed in state is different for each player. Properties always have to be the same for every player. So I'm not sure how that can work.

I was thinking of a way to SET x,y when clicking the button so that it would automatically go to certain locations.  Maybe there is some way other than a property?  It would not have to work for a changeRoutine.  But maybe a read only property?

96LawDawg:
> Also: should zooming into one reset any others?

Depends on the game.

96LawDawg:
> > need a property to show when zoomed in (so changeRoutines can work)
> 
> The idea is that the zoomed in state is different for each player. Properties always have to be the same for every player. So I'm not sure how that can work.

What if there is a property `zoomedPlayer` that is an array of the playerNames that are currently zoomed in on that widget?  The property would be the same for all players, but only those players in the array would see the zoomed view.  We could then use that to check for changes or the current player's view mode.

ArnoldSmith86:
Sounds great.

96LawDawg:
For `groupedWith`, if both zoomies are expanded, then they all also get minimized with the main controlling one does.  Is that also the intended and desired behavior?  Fine with me, that is just not mentioned in the draft wiki entry.

96LawDawg:
`lockChildren` has the opposite behavior of what it should be.

96LawDawg:
`zoomedMovable` does not work.  When zoomed in, the zoomy is never movable.

96LawDawg:
`zoomedScale` values are strange if you don't know what is going on behind the scenes.  0.5 is same scale.  0.4 to 0.1 is smaller.  0.6+ is larger. I think we should tweak the math behind the scenes so that 1 is same size, just like regular scale.

96LawDawg:
`zoomedScale` values are strange if you don't know what is going on behind the scenes.  0.5 is same scale.  0.4 to 0.1 is smaller.  0.6+ is larger. I think we should tweak the math behind the scenes so that 1 is same size, just like regular scale.

96LawDawg:
For the appearance I suggest:
 - Do something other than white as the default.  Looks too much like a holder.  Maybe silver (#c0c0c0)?
 - Reduce the size of the magnifying glass icon.  It is too large.  More like 80px instead of 200.
 - When setting the default style to false, the background should still be silver, or there should at least be a border. If you set it to false, the entire widget seems to disappear.
 - Consider adding a text property (or maybe html or both) and an image property just like a basic widget.  Would be nice to be able to throw some text or a background on there.

96LawDawg:
I think we need a better name for `groupedWith`.  It implies reciprocal behavior but that only works if both widgets list each other in their `groupedWith` property.  Maybe one of these: `minimizeOnExpand`, `collapseOnExpand`, `toggleOnExpand`.  I'm not sure any of those are right, but it better explains what is happening as far as I can tell.

ArnoldSmith86:
Yeah, the idea is something like tabs where only one of a group should be visible/expanded.

A usecase would be Flight where you only show the board of _one_ player on the left. All others could be thumbnails on the side.

The implementation is exactly what you describe: just collapse all of them.

It might also be interesting to not allow _all_ of them to be expanded. In the Flight example, it would just lead to a lot of empty space.

I agree that the name isn't particularly good. `collapseOnExpand` is basically correct but it sounds like collapse and expand both refer to the current zoomy which sounds stupid.. :grinning:

96LawDawg:
How about `collapseOthers`?

96LawDawg:
Adding a new zoomy is strange.  It shows up as w/h of 800x800.  But it is really only 400x400 because it is already scaled down.  This relates back to my comment about `zoomedScale`.

ArnoldSmith86:
> Adding a new zoomy is strange. It shows up as w/h of 800x800. But it is really only 400x400 because it is already scaled down. This relates back to my comment about `zoomedScale`.

I think that `zoomy` is more useful when the expanded version is `scale: 1` (so normal size) and the collapsed version is smaller. But maybe it's more intuitive if it is added expanded?

> Slightly strange behavior. Move a zoomy away from its zoomedx/y coordinates and click on it to enlarge it. Then click on it again to reduce it. Everything works as expected with the x/y. Now do it again but while it is enlarged, try dragging the zoomy. It will not move, but not when you click to reduce it, the x/y do not go back to what they were before.

I think that is caused by `zoomedMovable` only half working. Or a half working redirect from `x` to `zoomedX`.

> When setting the default style to false, the background should still be silver, or there should at least be a border. If you set it to false, the entire widget seems to disappear.

IMO that's the point of `defaultStyle: false`: to remove all applied styles. But we can also call it `showIcon: false` or something.

> `zoomedScale` values are strange if you don't know what is going on behind the scenes. 0.5 is same scale. 0.4 to 0.1 is smaller. 0.6+ is larger. I think we should tweak the math behind the scenes so that 1 is same size, just like regular scale.

But if we do `zoomedScale: 2` means that expanded is twice as big as collapsed, you would always have to change it if you change `scale` and intend to have the expanded version "real size".

96LawDawg:
I updated Stratagem to use zoomies so we would have at least one game with it when the PR is merged.  A very simple implementation but an improvement in the game play experience.

96LawDawg:
> I think that `zoomy` is more useful when the expanded version is `scale: 1` (so normal size) and the collapsed version is smaller. But maybe it's more intuitive if it is added expanded?

I'm not sure I agree.  At least when designing a game, I would expect the zoomies to be small size most of the time.  I'm really only using it expanded in play mode. But I guess I could go either way on that.  At a minimum it would make sense to add it expanded.

> IMO that's the point of `defaultStyle: false`: to remove all applied styles. But we can also call it `showIcon: false` or something.

Yes. Make the background silver in all cases. Change that property to `showIcon`.

> But if we do `zoomedScale: 2` means that expanded is twice as big as collapsed, you would always have to change it if you change `scale` and intend to have the expanded version "real size".

I'm not sure I follow all of the consequences of this, especially if the expanded scale is flipped like I suggest above which would make zoomedScale 1.5, I think.  Anyway, working with scaled widgets and getting sizing right can be tough.  The current math makes it harder.  I guess we could do math behind the scenes in the JavaScript to make the values behave like what you would expect.  But thinking about this makes my head hurt.

96LawDawg:
> It would probably be nice to make this even easier to setup, right? Something like a property `tabTitle` that, if set, transforms `zoomy` to a `tabby`?

I like the tabby example for a tutorial.  I don't think we need to implement it.  The graphical needs for each room will be so different than any attempt at standardization will be unhelpful.  Everyone will want something different, so let's give a demo, but not bake it into the code.

96LawDawg:
Building a new room is slightly frustrating.  Part of that is because of the scale and no matter what we do, that is not easy to fix.  But the other part is because the zoom relationship only goes one way.  Here is an example.  When starting in the room, both the main board and player board are their unzoomed, small size.  Like this:

Actual

But what I want is this:

Desired

So that I can get this swap:

Normal

So I had the bright idea to make the relationship the inverse: the main board starts large and the zoom factor makes it smaller.  Of course, that works, but then the `groupedWith` does not and you get this:

Inverse

I guess I can live with starting with everything small and unzoomed.  It just doesn't look as nice.  But this also made me wonder if we could have a `zoomedInverse` boolean property.  If true, then things that are grouped with it have the opposite relationship.

96LawDawg:
Also, it would be nice to have a property related to `groupedWith` that locks the relationship.  In that demo I'm making, it would be nice if clicking either board automatically swapped both.  Because if the main board is enlarged and the player board is small, then if you click the main board, it gets small and both or small.  The desired behavior (in this example) would be to click either and they flip whatever they were before.  Of course I can do that through routines, but not as easily as just `groupedWithLocked` or whatever.

ArnoldSmith86:
`zoomedPlayers` all should work but the first click would reset it.

But some sort of inversion could be nice.

96LawDawg:
Another nice-to-have feature: prevent other players from messing around with any board but their own.  Something like `playableBy`.  The basic idea would be to do something like "pointer-events":"none" for children of the zoomy (but not the zoomy itself otherwise you could not click it).  The pointer-events limitation would apply to everyone except for the seat id specified in `playableBy`.  Maybe that is not the best technical solution, but something like that.  Just an idea to create as many delays as possible in getting this implemented ;-)  This is absolutely not necessary.

Again, like with many things now, it is already possible to do this by putting an invisible cover over widgets to protect them. The cover uses `onlyVisibleForSeat` which is an array of all the seats excluding the one you want to have access to the widget.  Maybe the technical solution would incorporate something like that.  If it did, then this could be used on any widget type.

96LawDawg:
I added a new `onlyUsableBySeat` property.  It works substantially like `onlyVisibleForSeat`, but instead of making things invisible, it makes them non-interactive.  Actually, it only makes the children non-interactive.  That is so the zoomy widget can be clicked to expand.  I struggled with this concept.  If we did not have zoomy, I would have made the `onlyUsableBySeat` property apply to the widget itself (and thus all of its children). I'm not sure what the right approach here is because the name implies that the limitation applies to the widget.  A different name?  A different behavior?

Actually as I was writing this, I had an idea.  Maybe we should make the new property apply directly to the widget for every widget type except for zoomies.  And for zoomies, we make it only apply to children.  That way, the zoomy can be clicked.

96LawDawg:
I have fixed some of the things I commented on and come to accept that others are just the way they are because the way scale works.  This is getting pretty close to being ready.

client/js/widgets/zoomy.js
```diff
@@ -159,28 +159,28 @@ export class Zoomy extends Widget {
...
-        await this.set('zoomedScale', value);
-      else if(property == 'rotation')
-        await this.set('zoomedRotation', value);
-      else
-        await super.set(property, value);
-    } else {
-      await super.set(property, value);
-    }
-  }
-
```
ArnoldSmith86: Did you remove this on purpose? At first glance, it doesn't seem related to this commit. But so far it really only was a glance.

client/js/mousehandling.js
```diff
@@ -122,7 +122,12 @@ async function inputHandler(name, e) {
       mouseStatus[target.id].coords = coords;
       if(mouseStatus[target.id].moveTarget) {
         setDeltaCause(`${playerName} dragged ${widget.id}`);
-        await mouseStatus[target.id].moveTarget.move(coords, mouseStatus[target.id].localAnchor);
+        const mt = mouseStatus[target.id].moveTarget;
+        if(mt.get('type') === 'zoomy' && mt.isZoomed()) {
+          await mt.moveZoomed(coords, mouseStatus[target.id].localAnchor);
+        } else {
+          await mt.move(coords, mouseStatus[target.id].localAnchor);
+        }
```
ArnoldSmith86, line 130: This hack isn't necessary. You can define a new `move` for the zoomy class that checks if it's zoomed and if not, calls `super.move(...)` instead.

client/css/layout.css
```diff
@@ -812,10 +812,18 @@ body.jsonEdit:not(.hitTest) .widget.foreign, body.jsonEdit:not(.hitTest):not(.sh
...
-body:not(.jsonEdit) .widget.notUsable  > *{
+body:not(.jsonEdit) .widget.notUsable {
   pointer-events: none !important;
 }
 
+.widget.zoomyNotUsable {
+  --zoomyNotUsable:true;
+ }
+ 
+ body:not(.jsonEdit) .widget.zoomyNotUsable > * {
```
ArnoldSmith86, line 823: Can't you do `.widget.zoomy.notUsable` instead and remove the hack in the widget class?

If not, the code should at least be moved to `zoomy`. Its `classes` function could do a replace on the result of `super.classes`.

client/js/widgets/widget.js
```diff
@@ -2467,6 +2479,7 @@ export class Widget extends StateManaged {
     state.owner = null;
     state.linkedToSeat = null;
     state.onlyVisibleForSeat = null;
+    state.onlyUsableBySeat = null;
```
ArnoldSmith86, line 2486: Probably not necessary to remove for a readonly copy because it's also a readonly thingy. But doesn't hurt either.

96LawDawg: I put it there because it seemed to fit with onlyVisibleForSeat. But testing shows it is not needed.  Removed.

client/js/widgets/zoomy.js
96LawDawg: No.  That was accidental.  Restored.

client/js/mousehandling.js
96LawDawg: Fixed.

client/css/layout.css
96LawDawg: Fixed

96LawDawg:
I forgot what a pain it is to set up TestCafe with a new widget.  Uggh.  But at least it is done and passing.  Not sure if we should have the Zoomy tutorial on TestCafe also and throw in a few additional things like dragging and lockChildren?

PR #2527, opened 2025-03-29
add context menu

ArnoldSmith86:
This is a very rough demo for now.

The idea is to make it possible for right-click (longtouch on mobile) to not only show an optional enlarged view of the widget but also show user-defined context menu buttons.

This could solve common requests for overriding rightclick behavior as well as providing easy to use rotation as well as making cards editable with right-click.


96LawDawg:
Looks terrible.  Works great with my very limited testing.

PR #2529, opened 2025-03-31
Public Library - Add Tutorial 150 Working with Variables

96LawDawg:
This is the last in the series.

I combined what I thought would be 150 and 160 because in the end it did not make sense to separate them.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2529/pr-test (or any other room on that server)


ArnoldSmith86 (approved):
Take another look at this, please:

> Color functions can work with contrast, hue, and luminance so that colors will work with each other or contrast from each other. For example, will white or black text be the easiest to read when on top of the previous colors? For red, it will be black, but for teal, it is white.

PR #2535, opened 2025-04-01
Remove Assets Filter from Public Library

96LawDawg:
Filter is no longer available.  Assets removed from library.  Based on comments in Discord that this is outdated, no longer needed, and does not serve the intended purpose.


96LawDawg:
No idea why the setName() function fails here.  Not a high priority issue any way.

ArnoldSmith86 (approved):

PR #2548, opened 2025-04-12
Public Library - Add `svgReplaces` Tutorial

96LawDawg:
There was a small amount of svgReplaces in the Icons, Symbols, and Emojis tutorial.  This makes svgReplaces its own tutorial, as it should be.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2548/pr-test (or any other room on that server)


ArnoldSmith86 (approved):
Nice.

The second card example is a bit weird because I was expecting the different colors to be done using svgReplaces. But still a nice example..

PR #2562, opened 2025-05-12
fixed that `RESET` could set an invalid `parent`

ArnoldSmith86:
We check it for `SET` but not for `RESET`.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2562/pr-test (or any other room on that server)


96LawDawg (approved):

PR #2564, opened 2025-05-13
added filter textbox to debug log

ArnoldSmith86:
Now you can search the output.

### DIFF
--- a/client/css/jsonedit.css
+++ b/client/css/jsonedit.css
@@ -169,6 +169,10 @@ body.lightsOff .jeTreeExpander::before, body.lightsOff .jeTreeExpander-down::bef
   display: block;
 }
 
+.jeLogFilterNoMatch {
+  opacity: 0.2;
+}
+
 #jeText {
   background: none;
   -webkit-text-fill-color: transparent;
--- a/client/js/editor/sidebar/json.js
+++ b/client/js/editor/sidebar/json.js
@@ -110,6 +110,10 @@ class DebugModule extends SidebarModule {
       jeLoggingHTML = '';
   }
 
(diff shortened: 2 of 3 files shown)


96LawDawg:
Seems to work fine.  Four questions about what it should do.

First, it leaves red error message entries full brightness.  Personally, I think if I'm filtering for something, I would not want those distracting me.  I would prefer those to be dimmed when the filter has something in it.

Second, should the filter reset to blank after a new routine activates?  I think probably yes.  Right now, it leaves the existing word in the filter, but the result is not actually filtered.  That is not intuitive.

Third, entries below the first level are dimmed making them hard to see.  This is dark mode.  It is harder to read in light mode.  Seems like if the top level entry has the filtered word, the sub levels should also be bright.

Fourth, erasing the filter search word, collapses all entries back to just the main level.  Seems like if I have already expanded several levels down, changing the filter word should not automatically collapse the levels I expanded.

Screenshot 2025-05-13 124930

ArnoldSmith86:
I agree on all accounts. When running a new routine, it's probably more helpful to keep the filter, though? I don't know but when searching for something, it's probably a bug you want to fix and then look at the same output again. It's of course stupid to keep the text in the filter box and not apply it to new executions.

96LawDawg:
> When running a new routine, it's probably more helpful to keep the filter, though?

Sure.  That's probably the best solution, so long as the filter gets applied to the new executions.

ArnoldSmith86:
- [ ] Arrows are not always pointing down when things get expanded.
- [ ] Point three is not properly addressed. Gotta think about it.

PR #2565, opened 2025-05-13
added library asset names check

ArnoldSmith86:
Git seems to like to convert line endings when commiting but the game assets need to have the exact right content because the filename is calculated from the content. This check should make PRs unmergeable if there is an error and if converting to dos line endings results in a correct filename, it will commit a fix automically.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2565/pr-test (or any other room on that server)


96LawDawg (approved):
Blind approval as this will not impact the production server.

Issue #2350, opened 2024-10-26
`enterRoutine` needs to work with piles


96LawDawg:
In a holder that has piles, an `enterRoutine` works for the first card.  Cards after that enter the pile and not the holder.  But just like we do with other pile areas, VTT should look past the pile and pretend it is not there.  I think this could probably be game breaking, so maybe what we need to the file updater is not ridiculous is an ancestorEnterRoutine that works like _ancestor does for parent.

PR #2571, opened 2025-05-15
Public Library: Replace Safari Scramble with World Food Buffet

andypymont:
Update the public library game previously named Safari Scramble (similar to Flinke Pinke/Botswana):
- Change the name to World Food Buffet and the library image and assets to match. This allows the animal icons previously used to be deleted and removed from the Attribution, as we now use Noto Emoji standard assets from the asset picker.
- Add automatic live scoring as cards are played and tiles are taken, instead of having to press a button score at the end of the round.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2571/pr-test (or any other room on that server)


96LawDawg (changes requested):
There is some strange behavior with your custom piles.  Piles of 1 have a number (not present in the safari version).  Also, the pile count for the items farther down on the screen just disappeared.
Screenshot 2025-05-15 224500

96LawDawg:
Yes, that is a problem that needs to be fixed in the engine: #2350

96LawDawg (approved):
Approved

PR #2558, opened 2025-05-05
Public Library - Add Bao Huang

96LawDawg:
VTT implementation by Shobu, jade-alarm, and anewcar

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2558/pr-test (or any other room on that server)

PR #2574, opened 2025-05-19
fixed existing assets with wrong line endings

ArnoldSmith86:
See #2565 for the problem, checker and fixer. But these were committed to the repo before there was any check.

Hmm, this has to wait for the checker/fixer fix in #2558.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2574/pr-test (or any other room on that server)


96LawDawg (approved):
I looked at a sample of some of the games with the changes.  The images all seem to be fine.  Approved.

PR #571, opened 2021-06-16
Legacy compute modes

robartsd:
Adds "mode: ..." switch to activate legacy modes in routines. Removes automatic conversion of numerical strings to numbers (activate legacy mode with "mode: autoToNum"). Removes operands defaulting to 1 (activate legacy mode with "mode: defaultOne"). "mode: autoToNum defaultOne" added to beginning of each routine by file updater.

PR #2581, opened 2025-05-26
legacy compute modes

ArnoldSmith86:
This is basically the same functionality as #571 except that it uses a new concept of game settings to store the active legacy modes instead of setting them on a per routine basis.

#571 changed the default value for parameters from `1` to `null`. I changed it to `undefined` because at least `match` had trouble when given a `null` value.

---

TODO

- [x] `push` still adds `null` instead of `undefined` - gotta check (the debug output shows `null` because it is encoded as JSON in there and JSON doesn't have `undefined`)
- [x] testcafe compute fails because it is running without legacy modes
- [x] Start with new blank room should not use legacy mode (erase the check marks when creating a new room)
- [x] Legacy mode values need to be saved to meta data
- [x] Fix `colorContrast` and `colorCreateHue`
- [x] Add persistent visual indicator in Editor if any legacy mode is active
- [ ] Going from legacy mode to new room and reverting by using history tab does not revert to legacy mode
- [x] sidebar needs to be re-rendered when a state is loaded
- [x] after unsetting a legacy mode, it might be nice to show a delete button (for getting rid of it in the UI after making sure the game works without it)

---

Current UI (I'll add explanations and stuff):

image

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2581/pr-test (or any other room on that server)

### DIFF
--- a/client/js/editor/sidebar/gameSettings.js
+++ b/client/js/editor/sidebar/gameSettings.js
@@ -91,7 +91,9 @@ class GameSettingsModule extends SidebarModule {
 
     const handleRemove = (e) => {
       e.stopPropagation();
-      this.removeLegacyMode(name);
+      const confirmMessage = `This can't be undone.\n\nOnly do this if you've confirmed your game works correctly without this setting.`;
+      if(confirm(confirmMessage))
+        this.removeLegacyMode(name);
     };
 
     tile.addEventListener('click', handleToggle);


96LawDawg:
Edit mode is completely broken.

ArnoldSmith86:
Weird. I must have missed something simple. Will look tomorrow.

ArnoldSmith86:
It's technically the next day where I live.

96LawDawg:
Tell me how I should test this.

ArnoldSmith86:
- Is this doing what we want? For example `var a = '1' + '1' => "11"` might not be something we want (especially when reading from properties). At the same time, doing `var a = ${widgetID} + '1'` to concat a `1` after an `id` would break if we always convert to numbers when using `+`. When in doubt, we should probably stay close to what Javascript does.
- Feel free to improve the texts.
- Is this always taking the game settings with the game when saving, linking, loading, resetting, downloading, uploading, adding to PL and whatever else you can do?
- Is `undefined` the default value for parameters we want? Or should it be `null`? Or maybe something different depending on the operation?
- Does it make more sense to set the legacy mode in each routine like #571 did? Maybe because of copy-pasting code around?

Of course we can always change it with another legacy mode (and that's a really awesome thing) but we should probably think a second before doing this. :grin:

96LawDawg:
> Got another example where it doesn't copy the `gameSettings` with the state?

When I download the .vtt file using the main Game Shelf hamburger icon, there is no legacy info in the saved and downloaded game.

ArnoldSmith86:
More problems?

96LawDawg:
> More problems?

I tested all possible save methods using no legacy, both legacy modes, and each individual mode.  This included using the hamburger icon from the Game Shelf, the save icon from the Editor, and the Save Active Game from the Game Shelf. I also tested adding to the public library.  Everything works as expected.

96LawDawg:
Suggestion: it would be nice to have a visual indication that legacy mode is active.  I'm not sure of the best way, but what if in the Editor, we changed the Game Settings icon when legacy is active.  A good icon would be settings_alert which is the same gear with an exclamation point in it.  Not sure if that is possible or if there is a better solution.  But I know I would want a visual reminder without having to click on Game Settings.

96LawDawg:
By the way, that would look like this:

Screenshot 2025-05-29 161218

Instead of this:

Screenshot 2025-05-29 161247

ArnoldSmith86:
I was thinking about some red somewhere..

96LawDawg:
@ArnoldSmith86 and I had a discussion offline.  `*`, `/`, and `-` already autoconvert strings to text.  `+` should work the same way. There is `concat` if we want to keep strings as strings.  After changing this PR so that `+` still converts strings to numbers, I think this will be ready.

96LawDawg:
The badge on Game Settings has already saved me twice from much frustration in trying to figure out why isn't this working as expected.

96LawDawg (approved):
Approved through test and limited code review.

ArnoldSmith86:
Without legacy mode it works for me..

96LawDawg:
I did find one small glitch.  If you load a game with legacy mode, then create a blank room that works as intended.  But if you click on the history and go back to the version of the room with legacy mode, it is no longer enabled.  I'm not sure this must be fixed, but it would be the safest path to do so.

ArnoldSmith86:
@96LawDawg There was a conflict in the "last update" timestamp of the Math tutorial. I don't know what you updated in there. Please make sure that you didn't add two things in the same place or something.. :grin:

PR #2586, opened 2025-05-30
Public Library - Add Tutorial 101 JSON Syntax and Punctuation

96LawDawg:
This adds another tutorial to the foundations series.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2586/pr-test (or any other room on that server)

### DIFF
--- a/README-DOCKER.md
+++ b/README-DOCKER.md
@@ -91,9 +91,9 @@ http {
 
 # Custom public library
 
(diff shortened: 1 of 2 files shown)


ArnoldSmith86 (approved):
Made a few minor changes.

96LawDawg:
After reviewing this some more, I realized a left out a few things.  Updates in the latest commits:

Quotes, colon, and commas all mention functions and have a function example in the button now.  That was a pretty significant omission on my part.

I expanded the explanation for ( ) to clarify that only works for basic math operations and you cannot nest var operations.

I added a semi-colon to cover the old css format.

I'm going to leave this open a bit longer just in case I think of something else.

PR #2595, opened 2025-06-06
Add Label Object Type to Cards

### LINKED ISSUE #2526: Add editable object to card faceTemplate
Many devs have a need for cards to be editable.  Current solution is to attach labels to cards.  We can do better.

96LawDawg:
Allows for writing text on cards.

Adds new object type: label.  Adds new properties to cards: editable and labels.  Adds new boolean parameter to RECALL: eraseLabel.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2595/pr-test (or any other room on that server)

### DIFF
--- a/client/css/widgets/card.css
+++ b/client/css/widgets/card.css
@@ -30,12 +30,8 @@
   background: none;
   padding: 0;
   margin: 0;
-  color: inherit;
-  text-align: inherit;
-  font: inherit;
   width: 100%;
   resize: none;
-  text-transform: inherit;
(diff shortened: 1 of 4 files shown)


ArnoldSmith86:
Some `set`s are missing `await`s.

96LawDawg:
I think I've cleaned this up as much as I can and made it as simple as possible.

96LawDawg:
Implemented by #3069

PR #2610, opened 2025-06-18
Public Library - Add Out of Pocket

96LawDawg:
Similar to Ponzi Scheme. VTT implementation by Mitchell Stolycia.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2610/pr-test (or any other room on that server)


ArnoldSmith86 (changes requested):
There's no attribution for the audio file for notifying the player about their turn. It's also very big with >400kB for about a second.

ArnoldSmith86 (approved):

PR #2625, opened 2025-06-29
fixed widget position when becoming visible while changing parent

ArnoldSmith86:
For animation purposes, the widget position is being juggled while changing parents. But if the widget was invisible at the old parent, this did not work. Now it will not set the animation position in that case.


96LawDawg (approved):

PR #2626, opened 2025-06-29
Game validation

ArnoldSmith86:
I'm writing a function that tries to detect problems with a game before executing it. For example properties with the wrong type, undefined variables in routines, using widget IDs as collection names, having unused card types, etc, etc...

I have three use-cases for this:
- Show detected problems in the editor UI.
- Automatically check new public library games when creating a pull request and posting detected problems as a comment.
- Give AI a way to check its own output for validity so it can iterate until there are no more detected problems.

As of now (after the first two commits), it is possible to check a file with a standalone script like `node validator/validate_gamefile_node.js "library/games/Rodent Race/0.json"` and detected problems are shown above the debug output in the editor (clickable to jump to the offending place in the JSON):

image

There are still many false positives and false negatives though. I'll improve it now and over time but it's at a state now where I wanted to show what I have...

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2626/pr-test (or any other room on that server)


ArnoldSmith86:
Current idea is to check if the property is used in anything (routine, CSS, etc)..

Because it's also very common for the AI to hallucinate properties..

But if that's not good enough, I'll ignore them (or only check them in the standalone script)..

validator/validate_gamefile.js
```diff
@@ -0,0 +1,879 @@
...
+        return validateRoutine(v,context,propertyPath);
+    },
+    positiveNumber: v=>typeof v === 'number' && v >= 0 || 'positive number expected',
+    property: (v,p)=>Object.values(WIDGET_PROPERTIES).some(props=>Object.keys(props).includes(v)) || Object.values(p.widgets).some(w=>w[v] !== undefined) || `property '${v}' not found`,
+    vttSymbol: v=>v === null || typeof v === 'string', // TODO: replace with actual VTT symbol name check if available
+    countOrAll: v=>v === 'all' || typeof v === 'number' || 'number or "all" expected',
+    any: v=>true
+}
+
+// Common properties for all widgets
```
96LawDawg, line 34: Same comment here as I made below.  Difficult to keep this in sync with the things in the individual widget files when we make changes.

ArnoldSmith86:
> Also (probably obviously) need to add the predefined collections and variables.

Yeah, they are there but I broke it at some point.. 😅

ArnoldSmith86:
This can be improved a lot in many ways but it would probably be helpful to merge this soonish.

I agree that this shouldn't have so much redundant information but good solutions to that probably require changing the core widget code to grab the properties and their default values from somewhere else and TBH I don't want to deal with that in this PR.. It's already an issue with some JSON editor stuff..

96LawDawg:
Still seems like a lot of false positives and the whole thing takes up a lot of debug area.  I would prefer to see either making displaying this optional, or at least eliminate more of the false positives.  Right now, for me, I wouldn't use it and I would hide it.

For example, and probably an easy fix, it does not recognize the predefined collection "child."  It should recognize all pre-defined collections and variables.

Also, there is this example snippet from Four:

```JSON
       "var checkWord = ${activeWord} in ${PROPERTY wordArray OF wordList}"
      ]
    },
    {
      "func": "IF",
      "condition": "${checkWord}",
      "thenRoutine": [
```
And this shows up as "IF uses undefined variable 'checkWord'".  But it is defined immediately above.

ArnoldSmith86:
Yeah, there are two major issues that should be fixed:

1. `CALL`ing custom routines doesn't make it "define" the variables and collections of the calling routine.
2. `FOREACH` and `IF` don't "define" their variables and collections in the outside routine.

ArnoldSmith86:
Should be a lot better now. Ambagibus is a nice example where it actually finds stuff that broken in the game.

96LawDawg:
It is much better.  A few things I really think we should add/fix before merging to cut down on even more false positives:
 - recognition of custom properties as dropTarget.  That popped up in several games I looked at.
 - onlyVisibleForSeat does not seem to be working. Reports: "widget `seat1,seat2,seeat3,seat4` not found when they do actually exist
 - Using LABEL with `"label":"score${value}"` inside a loopRoutine where ${value} is defined is reporting "widgets score${value}" not found

And a bug I found when switching between games.  Load Godaigo. View the Debug.  Switch to Wildscapes.  Client crash.

Other things to consider, but maybe future improvements:
 - "comment" as a parameter for functions or even just as a stand-alone "function". Not used as much any more, but still supported.
 - support the deprecated "skip" parameter?  For example, Dots.
 - MOVEXY has a property for snapToGrid.  It should be added.  (Note, it was undocumented in the wiki until just now).
 - This one may be a bit complicated to fix, but if you have a dynamicProperties in a faceTemplate that has a value associated with a custom property, the custom property is still unrecognized even though it is actually being used to display something on the card.

There are also some really nice and unexpected catches. Like in a faceTemplate if you use color as a directly set value and as a dynamicProperties value.  Nice.

Overall, this is really good.  And I've already changed my mind and would use this.  Once you fix the few things at the top of the list, the bug, and at least consider the effort needed for the others.

ArnoldSmith86:
Most of those should be easy to fix. The crash is probably unrelated. There are a few games that crash when being unloaded. I think it has something to do with `inheritFrom` when the target is already removed or something. Also happens in Word Whirl IIRC.

It also recognizes using a widget ID as collection or a collection name inside an anonymous collection. We can add stuff like that whenever they come up in #help..

ArnoldSmith86:
> * support the deprecated "skip" parameter

That's the only one I didn't address. This is supposed to be one step beyond what's "technically correct" and be more like what's "recommended".

It would be easy to add though if you think it would be better.

96LawDawg:
The bug is on this PR.  Not in production.  Switching from Godaigo to Wildscapes:

```JS
p is not defined
ReferenceError: p is not defined
    at checkForDollarSign (https://test.virtualtabletop.io/PR-2626/edit.js:9506:21)
    at checkForDollarSign (https://test.virtualtabletop.io/PR-2626/edit.js:9513:30)
    at checkForDollarSign (https://test.virtualtabletop.io/PR-2626/edit.js:9513:30)
    at checkForDollarSign (https://test.virtualtabletop.io/PR-2626/edit.js:9513:30)
    at checkForDollarSign (https://test.virtualtabletop.io/PR-2626/edit.js:9513:30)
    at faceTemplates (https://test.virtualtabletop.io/PR-2626/edit.js:9155:38)
    at validateGameFile (https://test.virtualtabletop.io/PR-2626/edit.js:10037:36)
    at DebugModule.updateValidation (https://test.virtualtabletop.io/PR-2626/edit.js:3491:22)
    at DebugModule.onStateReceivedWhileActive (https://test.virtualtabletop.io/PR-2626/edit.js:3458:10)
    at DebugModule.onStateReceived (https://test.virtualtabletop.io/PR-2626/edit.js:1506:12)

```

Switching from Ascent to Bao Huang:

```JS
id.includes is not a function
TypeError: id.includes is not a function
    at https://test.virtualtabletop.io/PR-2626/edit.js:8991:62
    at Array.every (<anonymous>)
    at idArray (https://test.virtualtabletop.io/PR-2626/edit.js:8991:32)
    at validateGameFile (https://test.virtualtabletop.io/PR-2626/edit.js:10037:36)
    at DebugModule.updateValidation (https://test.virtualtabletop.io/PR-2626/edit.js:3491:22)
    at DebugModule.onStateReceivedWhileActive (https://test.virtualtabletop.io/PR-2626/edit.js:3458:10)
    at DebugModule.onStateReceived (https://test.virtualtabletop.io/PR-2626/edit.js:1506:12)
    at receiveStateFromServer (https://test.virtualtabletop.io/PR-2626/edit.js:383:12)
    at connection.onmessage (https://test.virtualtabletop.io/PR-2626/pr-test:2198:7)
```

96LawDawg:
```JSON
   {
      "func": "MOVE",
      "collection": "DEFAULT",
      "to": "hand${seatIndex}"
    }
  ```
This has collection DEFAULT as undefined.

96LawDawg:
You don't have to fix this note/comment example, but I just wanted to point it out.  In several of my games with complicated faceTemplates, I do something like this as one of the objects:

```JSON
    {
          "note": "solid white background",
          "type": "image",
          "x": 0,
          "y": 0,
          "width": 104,
          "height": 125,
          "color": "#ffffff"
        },
 ```
That shows up as an error, but I'm okay with that.

96LawDawg (approved):
Looks good.  Works good.  Approved through test.  By the way, I'm sure you noticed I made 2 small commits.

PR #2633, opened 2025-07-11
fixed crash in icon commands for properties starting with `icon`

ArnoldSmith86:
The JSON editor commands only work for `icon` but the context matching was not explicit enough.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2633/pr-test (or any other room on that server)


ArnoldSmith86:
Will do.

PR #2634, opened 2025-07-13
Add icons, images, and text to holders

96LawDawg:
Builds on #1937. Supersedes #2126.

For text, it works like the text and icon combo on basic widgets and buttons, etc.  This should have the smallest impact (i.e., none) on child widgets inheriting css tweaks on the holder itself.

Adds support for --bgImageOpacity and --bgColorOpacity as custom css variables. Creates color (for background) and textColor properties for holders.

Adds opacity and hoverOpacity as advanced icon options for all types of widgets.

Updates to v21 and adds a new legacy mode.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2634/pr-test (or any other room on that server)

---------

Wiki changes:

To BasicWidget:

| icon     |   null    |  Adds an icon or emoji from a curated list (or custom image uploaded by the user) to the widget. The value can be just text that is automatically inserted by the JSON Editor. Or using advanced settings, the icon can be an object with the following key value pairs (with the default value following in parenthesis): name (the name of the icon), scale (1), offsetX (0), offsetY (0), rotation (0), flip (""), opacity (null), color (""), strokeColor (""), strokeWidth (0), hoverColor (""), hoverStrokeColor (""), hoverStrokeWidth (null), hoverOpacity (null). For `flip`, valid values are "horizontal", "vertical", or "both". "Both" is the same as rotation 180.  | [icon](https://virtualtabletop.io/tutorial/properties-icon) |

# Holder

Holders are marked areas on a board you can put cards (or other widgets if you override _dropTarget_) into. If multiple cards are inside a holder (or piled anywhere on the board), by default they form a _pile_ with the number of cards attached to it as a drag handle.

[Click to see tutorial for holder appearance](https://virtualtabletop.io/tutorial/appearance-holders)

[Click to see tutorial for holder operations in general](https://virtualtabletop.io/tutorial/types-holder)

[Click to see tutorial for holders used as "hands"](https://virtualtabletop.io/tutorial/types-holder-hand)

| alignChildren | true            | If set to false, a new child will not be aligned to the holder. This allows an area where the player can put widgets wherever they want. |
| borderRadius | 8 | Same property as in [the global properties](#global-properties-for-every-widget) but with a different default value. ||
| childrenPerOwner | false        | If set to true, a new child will have its owner set to the player that drags the child into the holder. It will also remove the owner when the card is dragged out of the holder. This defaults to true if adding a new hand. |
| dropLimit | -1 (no limit) | The holder will not act as a valid drop target if adding the widgets would exceed the limit. If using `MOVE` or `SET` parent, this has no effect. |
| dropOffsetX, dropOffsetY | 4    | Same property as in [the global properties](#global-properties-for-every-widget) but with a different default value.|
| dropTarget | { "type": "card" } | This defines what widgets can be put into a holder and what widgets are affected by button operations using the holder. It can check any property of the potential child.<br/>The default is that any _card_ works. It can also accept an array of objects, and if the widget matches all criteria in any object, then it will be able to be dropped into the holder. <br/>For example, `[{"deck": "deck1", "cardType": "thisCardType"}, {"deck": "deck2", "cardType": "thatCardType"}]` would allow dropping of cards of type `thisCardtype` from `deck1` and cards of type `thatCardType` from `deck2`, but no others. Note that `"dropTarget": { }` and `"dropTarget": [{ }]`  will not filter the dropTargets, thus will accept any widget, while `"dropTarget": []` will not match any widgets.|
| dropShadow | false | If true, when a widget that matches the _dropTarget_ is dragged over the holder, a gray shadow will appear where the widget will be when dropped. The shadow is a cloned widget with a property `dropShadowOwner`. When a shadow is created, the value will be the `playerName`.  The clone will activate `enterRoutine` and `leaveRoutine`, so you may need to counteract that by using `IF` statements filtering out the clone looking for non-null values. The css of the shadow can be modified using `.dragging-shadow`. |
| height    | 168                 | Same property as in [the global properties](#global-properties-for-every-widget) but with a different default value. It's the default _card_ _height_ plus two times _dropOffsetY_. |
| hidePlayerCursors    | false (but true if the holder is a "hand")  | Same property as in [the global properties](#global-properties-for-every-widget) but with a different default value for "hands." |
| icon     |   null    |   Same property as in [the BasicWidget](#basicwidget). |
| image     |   ""    |   Same property as in [the BasicWidget](#basicwidget). |
| layer     | -3    | Same property as in the global properties but with a different default value. Above boards but below everything else.|
| movable   | false  | Same property as in [the global properties](#global-properties-for-every-widget) but with a different default value.|
| onEnter   | {}       | A set of properties that is applied to new children when they enter this holder.<br />For example setting the _activeFace_ to 1 so cards flip to their front face when putting them into the hand. To perform some type of automation when a widget enters the holder, use [[enterRoutine|Functions#introduction]].|
| onLeave   | {}        | See _onEnter_. Only applied when the child leaves the holder. To perform some type of automation when a widget enters the holder, use [[leaveRoutine|Functions#introduction]].|
| preventPiles | false | If set to true, cards in this holder will not form a pile, so they won't have a handle to move all of them at once. (Note however that if a pile was already present and then this property is changed to _true_ programmatically, the pile will remain.)|
| showInactiveFaceToSeat |null  |  When set to a seat `id` (or array of `id`s), the inactive face of a card will be displayed to the specified seats. This only works with cards, not basic widget faces. This will produce strange results if the card has more than 2 faces. This does not work as described on holders with `childrenPerOwner` set to true since the other players will not see any visual representation of cards not owned by them. | 
| stackOffsetX, stackOffsetY | 0  | If _alignChildren_ is true, the first child will get _dropOffsetX_ as its _x_; the second child will get _dropOffsetX_+_stackOffsetX_ as its _x_ and so forth. So this is the offset between children (cards) for example in the _hand_. See _overlap_ for an exception. |
| text     |   ""    |   Same property as in [the BasicWidget](#basicwidget). |
| typeClasses | "widget holder"   | Same property as in [the global properties](#global-properties-for-every-widget) but with a different default value. This gives the holder its default appearance. No guarantees if you change this. |
| width     | 111     | Same property as in [the global properties](#global-properties-for-every-widget) but with a different default value. It's the default _card_ _width_ plus two times _dropOffsetX_. |

### DIFF
--- a/library/tutorials/Appearance - Holders/0.json
+++ b/library/tutorials/Appearance - Holders/0.json
@@ -13,7 +13,7 @@
       "mode": "Tutorial",
       "time": "",
       "attribution": "",
-      "lastUpdate": 1770387252534,
+      "lastUpdate": 1770833003111,
       "showName": false,
       "skill": "",
       "description": "",
--- a/library/tutorials/Appearance - Holders/1.json
+++ b/library/tutorials/Appearance - Holders/1.json
@@ -13,7 +13,7 @@
       "mode": "Tutorial",
       "time": "",
       "attribution": "",
-      "lastUpdate": 1770387252535,
+      "lastUpdate": 1770833003113,
       "showName": false,
       "skill": "",
       "description": "",
--- a/library/tutorials/Appearance - Holders/2.json
+++ b/library/tutorials/Appearance - Holders/2.json
@@ -13,7 +13,7 @@
       "mode": "Tutorial",
       "time": "",
       "attribution": "",
-      "lastUpdate": 1770387252537,
+      "lastUpdate": 1770833003115,
       "showName": false,
       "skill": "",
       "description": "",
--- a/library/tutorials/Appearance - Holders/3.json
+++ b/library/tutorials/Appearance - Holders/3.json
@@ -1,35 +1,4 @@
 {
-  "_meta": {
-    "version": 21,
-    "gameSettings": {
-      "legacyModes": {}
-    },
-    "info": {
-      "name": "Appearance - Holders",
-      "image": "/assets/-1027054583_6051",
-      "rules": "",
-      "bgg": "",
-      "year": "",
-      "mode": "Tutorial",
-      "time": "",
-      "attribution": "",
-      "lastUpdate": 1770387252537,
-      "showName": false,
-      "skill": "",
-      "description": "",
-      "similarImage": "",
-      "similarName": "",
-      "similarDesigner": "",
-      "similarAwards": "",
-      "ruleText": "",
-      "helpText": "",
-      "usesAIImagery": false,
-      "variantImage": "",
-      "variant": "Icons",
-      "language": "en-US",
-      "players": "1"
-    }
-  },
   "title": {
     "id": "title",
     "x": 400,
@@ -1040,58 +1009,11 @@
   "holder2": {
     "type": "holder",
     "id": "holder2",
-    "x": 466,
+    "x": 421,
     "y": 284,
     "z": 9733,
     "icon": "👣",
     "text": "Bigfoot"
-  },
-  "holder3": {
-    "type": "holder",
-    "id": "holder3",
-    "x": 846,
-    "y": 284,
-    "z": 9733,
-    "icon": [
-      "💾",
-      "🎺"
-    ]
-  },
-  "holder4": {
-    "type": "holder",
-    "id": "holder4",
-    "x": 1224,
-    "y": 284,
-    "z": 9734,
-    "icon": [
-      {
-        "name": "[trash]",
-        "scale": 1,
-        "offsetX": -0.3,
-        "offsetY": -0.1,
-        "rotation": 10,
-        "color": "red",
-        "strokeColor": "",
-        "strokeWidth": 0,
-        "hoverStrokeColor": "",
-        "hoverStrokeWidth": null,
-        "hoverOpacity": 1
-      },
-      {
-        "name": "[cards]",
-        "scale": 1,
-        "offsetX": 0.2,
-        "offsetY": 0.1,
-        "rotation": 10,
-        "color": "silver",
-        "opacity": 1,
-        "strokeColor": "",
-        "strokeWidth": 0,
-        "hoverColor": "gold",
-        "hoverStrokeColor": "",
-        "hoverStrokeWidth": null
-      }
-    ]
   },
   "text1": {
     "id": "text1",
@@ -1109,7 +1031,7 @@
   },
   "text2": {
     "id": "text2",
-    "x": 392,
+    "x": 358,
     "y": 482,
     "width": 230,
     "layer": -3,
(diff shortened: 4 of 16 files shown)


client/js/widgets/imagewidget.js
```diff
@@ -12,8 +12,11 @@ class ImageWidget extends Widget {
 
   applyDeltaToDOM(delta) {
     super.applyDeltaToDOM(delta);
-    if(delta.text !== undefined || delta.icon !== undefined)
-      setText(this.domElement, this.get('icon') ? '' : this.get('text'));
+    if(delta.text !== undefined || delta.icon !== undefined) {
+      if(this.get('type') !== 'holder') {
```
ArnoldSmith86, line 16: Holder doesn't have `setText`, right? Maybe do `typeOf this.setText == 'function'`?

client/js/widgets/imagewidget.js
```diff
@@ -90,9 +93,23 @@ class ImageWidget extends Widget {
   }
 
   updateIcon() {
```
ArnoldSmith86, line 95: Try to move the holder code into the Holder class by defining an `updateIcon` there that calls `super.updateIcon()`.

client/js/widgets/imagewidget.js
96LawDawg: AI says `if (typeof setText === 'function') ` would always be true because the function exists as a global function and not a method on the holder class.  This line was added to prevent the text from appearing twice (once in the position it would be for a basic widget).

ArnoldSmith86:
Do you agree that doing a compatibility toggle for it would be overkill? I guess we just do it and break+update Runway Roll?

ArnoldSmith86:
Like the other toggles, it should only enable itself if a legacy game actually uses the old behavior. For this PR it is probably enough to check if any holder has an `image` property. So for most all games, it wouldn't even apply. Sounds reasonable. Especially once we hide "past" legacy toggles from the POV of the game..

ArnoldSmith86:
> I don't understand css specificity enough, but I'm guessing there are situations where our new default css properties for holders could conflict with things in the game's holder css. Is that the problem?

You only added CSS for `.widget.holder > .holderTextOnly` which is a new class.

---

I implemented the legacy mode now and it isn't as bad as I thought it would be. Not exactly pretty but should be maintainable.

Really cool to load Runway Roll or Rodent Race now and toggling the legacy mode in the editor back and forth. :grin: 

For reference, these are the library variants using `image`:

```
games/An Infamous Traffic/0.json
games/Chinese Checkers/0.json
games/Fantasy Wars/0.json
games/Fantasy Wars/1.json
games/Flower Poker/0.json
games/Frontiers/0.json
games/Go/0.json
games/God Gauntlet/0.json
games/Kyoga/0.json
games/Magical Traders/0.json
games/Marble Mania/0.json
games/Moroccan Moolah/0.json
games/Pumpkin Curse/0.json
games/Rodent Race/0.json
games/Runway Roll/0.json
games/Silent Spectre/0.json
games/Steam Power/0.json
games/Steam Power/1.json
games/Union Station/0.json
games/Wildscapes/0.json
games/Zen Poker/0.json
```

These use `text`:

```
games/An Infamous Traffic/0.json
games/Ascent/0.json
games/Beffroi/0.json
games/Castellers/0.json
games/Downtown/2.json
games/Downtown/3.json
games/Etna/0.json
games/Expeditions/0.json
games/Flatten the Curve/0.json
games/Going Nuts/0.json
games/Pike Pier/0.json
games/Rackball/0.json
games/Road Race/0.json
games/Runway Roll/0.json
games/Speedy Fruits/0.json
games/Splendid/0.json
games/Stairs/0.json
games/Steam Power/0.json
games/Steam Power/1.json
games/Wordy/0.json
tutorials/Appearance - CSS/1.json
tutorials/Functions - IF/0.json
tutorials/Operations - Arrays/6.json
tutorials/Types - Holders/1.json
```

None use `icon`.

96LawDawg:
Just noticed a problem.  Deleting the text property in the JSON editor does not immediately remove the text from the room.

ArnoldSmith86:
I tried to make the code cleaner and move holder stuff into holder files. Please check that stuff still works. :grin:

PR #2638, opened 2025-07-15
fixed crash with `image` inside routine

ArnoldSmith86:
When having an `image` property somewhere in a routine operation, a context check led to a crash because the context contains a virtual `(SET)`. This PR makes affected functions skip those virtual context entries.

I came across this in Frontiers, `Base Map Button New` line 554 for example. It `SET`s a `grid` property that in turn contains `image` properties.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2638/pr-test (or any other room on that server)


ArnoldSmith86:
Clicking the button doesn't crash. Clicking into the code line (with `image`) crashes.

PR #2651, opened 2025-07-25
Public Library - Update Frontiers with seafarer boards

ArnoldSmith86:
The ship is ugly, the metadata was not updated and only the first four scenarios are added. But I think those are playable now.

I might continue to work on this in the coming days but maybe I will need a longer break. I still want to add lots and lots of stuff to this. In time.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2651/pr-test (or any other room on that server)

### DIFF
--- a/library/games/Frontiers/0.json
+++ b/library/games/Frontiers/0.json
@@ -1195,52 +1195,52 @@
                 "text": "Even more random board for 5/6 players"
               },
               {
-                "value": "Heading For New Shores 3P Map",
-                "text": "Heading For New Shores Seafarers board for 3 players"
+                "value": "Uncharted Waters 3P Map",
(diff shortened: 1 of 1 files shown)


96LawDawg:
Did you want approval on this or is it just a draft?

ArnoldSmith86:
I want to merge it soonish but I do want to add a randomized seafarer board first..

Of course I would also prefer a better ship image..

ArnoldSmith86:
Many things aren't perfect. But hopefully good enough. :grin:

ArnoldSmith86:
> However, for the expansion packs, I don't think you should be using the same names as those used in Catan, right?

That's what AI came up with.

Issue #2596, opened 2025-06-07
One player opening an INPUT modal for another player


96LawDawg:
Currently not possible, but would be nice to have.

96LawDawg:
Implemented by #2657

PR #2657, opened 2025-08-09
INPUT modal can be opened for other players

96LawDawg:
Implements #2596.

Adds new "player" and "block" parameters to INPUT. "player" can be scalar or array; defaults to null. If set to another player name(s), the INPUT modal opens for that other player(s).  For "block", if set to true (defaults to false), then every player gets a pop up message about waiting for input from other players. That pop-up blocks any other interaction with the game.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2657/pr-test (or any other room on that server)


96LawDawg:
I have also updated the tutorial. I want to make all of these changes (except for the one relating to the new player parameter) even if this PR does not advance.  The tutorial breaks out the basic features and the appearance features into 2 variants and adds some additional examples from https://discord.com/channels/770758631146782780/862352981747695616/1382466550862319627 to the appearance variant.

96LawDawg:
As part of this new PR, it might be a good idea to add an overlay to all players that says "Waiting on input from (playerName)."  That would prevent any changes from taking place while a player is doing input-related stuff.

Also, one nice thing I noticed in testing, if the player chose to do the input is on some other screen (like the game shelf or even the About menu, sending the overlay forces the player back to the game immediately and they see the overlay so it doesn't get lost.  Of course, there is a possibility of glitches with this still.  Not sure what would happen if the delta never got sent or got lost.  Maybe that would be another good reason for the overlay to all players.  In addition to saying which player they are waiting on, maybe add a cancel button onto that overlay that would end the input wait for everyone.

ArnoldSmith86:
It's interesting. I do think we should implement the feature but I agree with your suggestions and additionally we should make it possible to show overlays to multiple players - think choosing cards to pass in Hearts.

The PR as is basically implements "execute the rest of this routine on another client". That doesn't really have to do anything with INPUT and would probably make more sense for CALL (call the target routine on another client).

INPUT should probably just tell other clients to show an overlay and send the result back to the initial client.

And something like a `block` parameter could be independent of that to show all players a "waiting for X" even if the INPUT stayed on one client. Would even be useful outside of INPUT for long running reset buttons etc..

VTT-AI-Agent:
Reworked the remote-INPUT mechanism to match the design discussed above (VTT-AI-Agent, on behalf of continuing this PR):

- **INPUT no longer delegates the rest of the routine.** When `player` names other player(s), INPUT now only asks those client(s) to show the overlay and sends their result back to the initiating client, which keeps running the routine locally. This is the "tell other clients to show an overlay and send the result back" behavior you asked for; the "run the rest on another client" semantics are left for CALL.
- **`player` accepts a single name or a list**, so the overlay can be shown to multiple players. Each responding player's variables/collections are merged back on the initiator as results arrive. (The exact merge semantics for the simultaneous multi-player case — e.g. Hearts card-passing where each player needs their own separate result — are still open; happy to shape that however you prefer.)
- **New independent `block` parameter**: shows every other player a "Waiting for input from X…" overlay while the input is pending. It's orthogonal to `player`, so it also works for a purely local INPUT (useful for long-running reset/button routines).
- The server tracks pending input requests and blocks, and clears them if a participating player disconnects, so the initiator never hangs.

Verified end-to-end with two browser clients: initiator fires the routine → target sees the overlay → initiator shows the "waiting" overlay → target submits → the value propagates back to the initiator's routine and the block clears. `npm test` passes and all CI was green on the prior revision.

Still open / not done here: a cancel button on the block overlay to abort the wait for everyone, and finalizing the multi-player result semantics.

VTT-AI-Agent:
Incorporated the remaining suggestions from the Aug 10 discussion (multi-player input + a cancel button):

**Multi-player INPUT** — `player` now accepts a **list** of names. The overlay is shown to every listed player simultaneously, and each field's value comes back **keyed by player name**, e.g.:
```
{ "func": "INPUT", "player": ["Alice","Bob","Carol"], "block": true,
  "fields": [ { "type": "choose", "variable": "pass", ... } ] }
```
afterwards `${pass.Alice}`, `${pass.Bob}`, … hold each player's choice (the Hearts card-passing case). A single name string keeps the previous flat semantics, so the existing tutorial example is unchanged. The initiator can be one of the listed players — it gets its own overlay locally while the others are asked over the network.

**Cancel button** — the "Waiting for input from X…" block overlay now has a **Cancel** button that ends the input for everyone: it closes the pending overlays, clears the block on all clients and aborts the initiator's routine.

Remote input is now tracked per **session**, so aborts, cancellations and disconnects (of either the initiator or a target) tear everything down cleanly rather than leaving a client hung or stuck behind an overlay.

Verified end-to-end with three browser clients (Alice initiating, Bob+Carol responding): per-player keyed results propagate back correctly; the initiator sees the waiting overlay; and Carol pressing Cancel on the waiting overlay closes Bob's input, unblocks everyone and aborts the routine. Local single-player INPUT and `npm test` still pass.

VTT-AI-Agent (commented):
## TL;DR

The overall design (client-initiated sessions relayed via the server, results keyed by player) fits the codebase well, but there are several correctness bugs to fix before this leaves draft. The most important one: the JSON editor template now inserts `"player": null`, which makes every INPUT created from the editor a silent no-op. Also, the `sendDelta()` call before `requestInput` is a no-op inside routines, so remote players can see stale state, and cancellation in multi-player sessions strands the remaining targets.

## Bugs

**1. `"player": null` in the JSON editor template makes new INPUTs silently do nothing** — `client/js/jsonedit.js:1297`
`setDefaults` only fills in properties that are `undefined` (`client/js/widgets/widget.js:833`), so an explicit `null` survives. Then `asArray(null).filter(Boolean)` yields `[]`, `runInput` sees no players and resolves immediately with `{}`, and the overlay never appears — no variables set, no error, no problem logged. So every INPUT inserted from the editor's template is broken until the author deletes the `player` key. Suggest either omitting `player` from the template, or (better) treating a null/empty `player` in widget.js as "the acting player" so it degrades to the old behavior instead of skipping the operation.

**2. `sendDelta()` in `runInput` is a no-op — remote targets act on stale state** — `client/js/serverstate.js` (`runInput`, before `toServer('requestInput', …)`)
Routines run inside `batchStart()`/`batchEnd()` (`client/js/widgets/widget.js:859` / `:2143`), so `batchDepth >= 1` and `sendDelta()` returns without flushing (`if(!batchDepth)`). Any state changed earlier in the same routine (widgets moved/flipped/created — including the FLIP in your own tutorial example) has not reached the target when the overlay opens. Concretely: a `choose` field will render pre-routine widget state, and `resolveCollections` on the target silently drops IDs of widgets whose creation delta hasn't been flushed yet. You need an actual forced flush here (or a different mechanism), not a call to the batch-guarded `sendDelta`.

**3. One target cancelling strands the other targets in a multi-player session** — `server/room.mjs` (`inputResult`)
When a target sends a cancelled `inputResult`, the server deletes `this.inputRequests[sessionID]` immediately. The initiator then reacts with `abortInput`, but by the time it arrives the request is gone, so `abortInput` early-returns and the remaining targets never receive `hideInput`. They're left filling out a form whose session is dead; their eventual result is dropped (initiator lookup fails) with no feedback. Same pattern in `cleanupInputForPlayer`'s "target left" branch. Fix: when deleting the request on a cancel, send `hideInput` to the remaining targets right there (which would also let you drop the client's `abortInput` round-trip for this path).

**4. `choose` fields with an array `source` are broken on remote clients** — `client/js/serverstate.js` (`receiveShowInput`)
The stub `getCollection = c => Array.isArray(collections[c]) ? c : 'DEFAULT'` doesn't replicate the real `getCollection` (`client/js/widgets/widget.js:839`), which builds a collection from an array of widget IDs. On a remote target, `source: ["someWidget"]` resolves to `'DEFAULT'`, which usually doesn't exist in the transferred collections, so `field.widgets` ends up empty and the chooser shows nothing.

**5. Sessions are keyed by player name — renames and duplicate names break them**
`session.remaining`, `request.remaining` and `inputBlocks` are all keyed by `playerName`. If a target renames themselves while an input is pending, their result comes back under the new name, `remaining.delete()` misses, and the initiator's routine hangs forever. Similarly, `this.players.find(p => p.name === name)` delivers `showInput` to only the first connection when the same name is open in multiple tabs, and `sessionID` (`${playerName}-${counter}`) can collide between two same-named clients. Keying server-side state by the Player object/connection rather than name would fix most of this.

## Edge cases worth a look

- **Concurrent sessions on one client**: `receiveShowInput` opens the overlay unconditionally. If the target already has an INPUT open (their own or another session's), the second call rewrites `#buttonInputFields` while the first `goHandler` is still attached to `#buttonInputGo` — one click then fires both handlers and both sessions resolve with the second form's data. Consider rejecting (auto-cancel) an incoming `showInput` while one is already active.
- **`showOverlay` toggles**: `updateInputBlockOverlay` calls `showOverlay('inputBlockOverlay')` on every `inputBlock` message, and `showOverlay` toggles visibility when the overlay is already shown (`client/js/main.js:97`) — a second concurrent block *hides* the wait overlay. It also force-closes whatever overlay the player had open, including another session's input overlay, which can make that session unanswerable. Guard the call on the overlay currently being hidden, and consider excluding players who currently have a pending `showInput`.
- **Remote `playerName`/`playerColor` variables rename the initiator**: at routine end (`client/js/widgets/widget.js:2145-2152`), `variables.playerName` triggers a rename of the *initiating* player. With `player: 'Bob'`, a field with `variable: "playerName"` filled in by Bob renames the initiator — surprising; probably worth ignoring these special variables for remote results.

## Minor

- The loop rewrite in `evaluateRoutine` (`for(let index = 0; …)` at `client/js/widgets/widget.js:895`) — `index` is never used; looks like a leftover from an earlier approach and can be reverted.
- Tutorial `0.json` FLIPs collection `["checker"]` but no widget with id `checker` exists in that file, so the FLIP is a no-op.
- The new `block` parameter is implemented and added to the editor template but not mentioned in the PR description/issue; make sure it gets wiki documentation alongside `player`.
- No test coverage. The multi-client flow is admittedly hard to cover in TestCafe, but the pure pieces (empty/null `player` handling, the multi-player result keying in widget.js) would benefit from at least a targeted test.

One design note, not a blocker: this necessarily lets any client open arbitrary modals on any other player's screen. That's consistent with VTT's existing trust model (clients can already send arbitrary deltas), but worth being conscious of.

VTT-AI-Agent (commented):
## TL;DR

Nice feature and the overall session/protocol design is reasonable, but there are several correctness gaps that will bite in real games. Most important: the `sendDelta()` call in `runInput` is a no-op (routines run inside a batch), so remote players see **stale room state** — including missing widgets for `choose` collections — while answering. Also, in multi-player mode a single cancel or disconnect leaves the *other* targets' modals dangling, and session routing by player name breaks with duplicate names (multiple tabs/devices). No tests were added for a fairly complex new client↔server protocol.

## Correctness

1. **`sendDelta()` before `requestInput` does nothing** — `client/js/serverstate.js:707` (in `runInput`). Routines always run inside `batchStart()`/`batchEnd()` (`widget.js:859`/`2143`), so `batchDepth >= 1` and `sendDelta()` returns immediately without flushing (`serverstate.js:555`). Consequences: any widgets created or properties changed earlier in the same routine are invisible to the target while it answers, and `resolveCollections` on the target silently drops widgets whose creation delta hasn't been sent (`widgets.get(id)` returns undefined) — e.g. "deal cards, then ask another player to choose one" shows them an empty/stale `choose` field. You'd need to actually force-flush the pending delta (temporarily bypassing the batch guard) before sending `requestInput`, or document that INPUT-for-others only sees pre-routine state.

2. **Cancel/disconnect of one target strands the other targets (multi-player mode)** — `server/room.mjs`, `inputResult()` and `cleanupInputForPlayer()`. When one target cancels, the server deletes `this.inputRequests[sessionID]` *before* the initiator reacts; the initiator's `session.cancel()` then sends `abortInput`, which finds no request (`abortInput` returns early) and never sends `hideInput` to the remaining targets. Their modals stay open, and their eventual `inputResult` is silently dropped (`request` is null, so `initiator` lookup uses `null`). Same thing when a target disconnects (`cleanupInputForPlayer` deletes the request after notifying only the initiator). Repro: `player: ["A","B"]`, B presses Cancel → A's modal lingers and A's answer goes nowhere. Fix: have the server send `hideInput` to `request.remaining` itself before deleting the request in both places (like the initiator-left branch already does).

3. **Name-based routing breaks with duplicate player names and renames.** Session IDs are `${playerName}-${counter}` with a per-tab counter starting at 0 (`serverstate.js:602`), so two connections sharing a name (same person on two devices — a normal situation in VTT, cf. the duplicate-name check in `room.mjs:786`) can both generate `Alice-1`, colliding in `this.inputRequests`. And `inputResult`/`cancelInput` route replies via `this.players.find(p=>p.name === request.from)`, which picks the *first* connection with that name — possibly not the initiating tab, in which case the result is silently dropped and the initiator's routine hangs forever. Renaming mid-session has the same effect. Tracking the initiator `Player` object (connection) server-side instead of its name, and adding a random component to the session ID, would fix both.

4. **Silent failure modes.** If `player` evaluates to an unknown name, the server replies `cancelled` and the routine aborts with no `problems` entry and no hint to the game author (a typo in a player name is indistinguishable from the player cancelling). If it evaluates to empty/null, `players` becomes `[]` and the INPUT silently resolves with no variables set at all. Both deserve a `problems.push(...)` and/or falling back to the initiating player. Relatedly, without `block: true` there is no way for anyone to abort a pending remote input — an AFK target blocks the initiator's routine indefinitely.

5. **Block overlay toggles off with overlapping sessions** — `updateInputBlockOverlay()` (`serverstate.js` new code) calls `showOverlay('inputBlockOverlay')`, but `showOverlay` *toggles* when the overlay is already visible (`main.js:97`), so a second concurrent `inputBlock` show-message hides the overlay instead of updating it. The code clearly intends to support multiple `activeInputBlocks`, so this needs a non-toggling show path.

## Smaller points

- `client/js/widgets/widget.js:895` — the `for..of` → indexed `for` loop change is unused (`index` is never referenced); looks like a leftover from an abandoned approach and should be reverted to keep the diff focused.
- `library/tutorials/Functions - INPUT/0.json` (`jkmt` routine) — the `FLIP` on collection `["checker"]` references a widget that doesn't exist in this variant; looks like debug leftover.
- `server/room.mjs` — `inputRequests`/`inputBlocks` are plain objects keyed by client-supplied strings; a client sending `sessionID: "__proto__"` corrupts the map (assignment sets the prototype, `delete` doesn't remove it). Use `Map`s or `Object.create(null)`.
- Any player who sees the block overlay — including uninvolved spectators — can press Cancel and abort the initiator's routine. If that's intended, fine, but it's worth a conscious decision.
- Semantics worth documenting (wiki/Functions): `${...}` substitution in the fields happens on the *initiator*, so e.g. `value: "${playerName}"` pre-fills the initiator's name for every target, not each target's own.
- If a target already has an INPUT overlay open (its own, or a second concurrent remote request), the new `showInputOverlay` clobbers the shared DOM and stacks Go/Cancel listeners — pressing Go then settles both promises with the second overlay's fields. This was hard to hit before, but remote INPUT makes concurrent overlays likely; some queueing or at least last-wins teardown is needed.
- **No tests**: this adds a five-message client↔server protocol plus disconnect handling and nothing covers it. Even a TestCafe two-client scenario for the happy path + cancel would help a lot (see `docs/testcafe.md`).
- The new `block` parameter isn't demonstrated in the updated tutorial, and the JSON editor default was added — consider showing it in the tutorial too.

(`npm test` passes on the branch; the tutorial JSON version bumps to 17 match the current FileUpdater VERSION, so the 1.json staying at 14 is harmless — it just gets migrated on load.)

VTT-AI-Agent:
Addressed the review findings (pushed in 5558af0):

**Bugs**
1. **`player: null` no-op** — fixed. A null/empty `player` (including the editor template's default) now falls back to the acting player and shows the overlay locally, so editor-created INPUTs work out of the box.
2. **`sendDelta()` no-op inside routines** — fixed. Added a `flushDelta()` that pushes the accumulated delta even while `batchDepth >= 1`, and call it before `requestInput`. Verified: a widget cloned earlier in the same routine now appears in a remote `choose` field (previously the chooser was empty because `resolveCollections` dropped the unflushed IDs).
3. **One cancel strands the other targets** — fixed. The server now closes the remaining targets (`hideInput`) as soon as any `inputResult` is cancelled, rather than relying on the initiator's later `abortInput`. Same treatment in `cleanupInputForPlayer`.
4. **Array `source` broken remotely** — fixed. The remote `getCollection` now mirrors `evaluateRoutine`'s: an array source builds a collection from those widget IDs (via `widgetFilter`) instead of falling through to `DEFAULT`.
5. **Name-keyed sessions break on rename/duplicate names** — fixed. Server-side `inputRequests`/`inputBlocks` are now keyed by the **Player connection object**, and the server echoes back the name the initiator used so the initiating client still correlates its (name-keyed) session after a rename. Session IDs now carry a per-page-load token so two same-named clients can't collide (derived from `performance` timing, not the seeded RNG).

**Edge cases**
- **Concurrent overlays** — `receiveShowInput` now declines (auto-cancels) an incoming remote INPUT while another input overlay is already open, so a single click can't resolve two forms.
- **`showOverlay` toggle** — `updateInputBlockOverlay` only calls `showOverlay` on the hidden→shown transition and skips it entirely while the player is filling in their own input overlay, so a second block message no longer hides the wait overlay or steals an active input.
- **Remote `playerName`/`playerColor`** — these are now stripped from remote results so a field filled in by another player can't rename/recolor the initiator.

**Minor**
- Reverted the unused `for(let index …)` loop change back to `for(const original of routine)`.
- The `checker` FLIP no-op is in the tutorial content (`0.json`, from the separate tutorial commit), not the remote-input code — leaving that for the tutorial author to resolve so I don't disturb content I didn't write.
- `block` will need a wiki entry alongside `player`; noted for the docs pass.
- **Tests:** verified end-to-end with a 3-client browser harness covering all five bug fixes (null→local, mid-routine flush visible remotely, cancel frees the other target + aborts the routine, and multi-player keyed results still routing correctly after the connection-keyed refactor); `npm test` (jest) still passes. The full multi-client round-trip remains impractical to cover in TestCafe, as noted.

One thing I kept as-is: array `player` → results keyed by player name, single string → flat (preserves the tutorial example). Happy to change that convention if you'd prefer keyed always.

VTT-AI-Agent:
Thanks — this review was generated against `fbaa2bd`, which is the commit *before* my fix pass in `5558af0`. Bugs 1–5 and all three edge cases from it were already addressed there (null→local fallback, real `flushDelta` bypassing the batch guard, cancel closing the other targets server-side, array-`source` `getCollection`, connection-keyed sessions + per-load token, concurrent-overlay guard, non-toggling block overlay, and stripping remote `playerName`/`playerColor`). Details are in my reply on the previous review.

The second review did surface two genuinely new points, now fixed in `c26f3b9`:

- **Unknown player name = silent abort** — an INPUT whose `player` names someone who isn't an active player now pushes a routine `problem` (`'<name>' is not an active player…`), so a typo is distinguishable from a real cancel.
- **Prototype pollution via client-supplied keys** — the server's `inputRequests`/`inputBlocks` and the client's `inputSessions`/`remoteInputHandles` are now `Object.create(null)` maps, so a crafted `sessionID`/`blockID` like `__proto__` can't corrupt them.

Re-verified with the 3-client browser harness (happy-path remote result + unknown-player no-hang) and `npm test` passes.

Remaining points I'm intentionally *not* changing, with rationale:
- **`block`-less inputs can't be aborted by others** — by design: the cancel button lives on the waiting overlay, so `block: true` is the opt-in for cancelability. An AFK target with no block is analogous to any long INPUT; happy to add a timeout later if you want one.
- **Any player (incl. spectators) can press Cancel** — matches @96LawDawg's original request ("a cancel button… that would end the input wait for everyone").
- **`checker` FLIP no-op** and **`block` not shown in the tutorial** live in the tutorial JSON from the separate tutorial commit, not the remote-input code — leaving those for the tutorial author. `block` wiki docs are noted for the docs pass.
- **`${…}` substitution happens on the initiator** — correct and intentional (the initiator has the live variables/collections); worth a wiki note, which I'll include with the `block`/`player` docs.

One convention still open for your call: array `player` → results keyed by player name, single string → flat. Let me know if you'd prefer keyed always.

VTT-AI-Agent (commented):
## TL;DR

The follow-up commits address essentially all findings from the previous reviews (forced delta flush, null-`player` fallback, server-side `hideInput` on cancel, connection-based session tracking, busy guard, prototype-safe maps) — nice work. One significant bug remains: a multi-player INPUT whose target list is only *partially* reachable hangs the initiator's routine forever, because the server silently drops unreachable names. Plus a couple of smaller edge cases below. `npm test` passes on the branch.

## Remaining bugs

**1. Partially unreachable targets hang the initiator forever** — `server/room.mjs` (`requestInput`). Unreachable names are filtered out of `targets`, but the initiator is only told (`cancelled: true`) when *no* target is reachable. With `player: ["A","B"]` and B disconnected (a typo, or B leaves between the client-side `activePlayers` check and the server processing the message), the server sends `showInput` to A only, while the initiator's `session.remaining` still contains B — the session can never complete, the routine's `await` never resolves, and since the routine's batch stays open, the initiating client stops sending deltas entirely. The `problems.push` warning in `widget.js:1434` logs the typo but doesn't prevent the hang. Fix: in `requestInput`, send an `inputResult { cancelled: true }` for each requested name with no matching player (consistent with the one-cancel-cancels-all semantics), or drop non-active names client-side after warning.

**2. A duplicated name in the `player` list cancels the whole session** — `client/js/serverstate.js` (`runInput`) / `server/room.mjs` (`requestInput`). `remoteTargets` keeps duplicates, so `player: ["A","A"]` makes the server send two `showInput` messages (same sessionID) to the same connection. The second one hits the target's busy check (`remoteInputHandles` non-empty) and replies `cancelled`, which kills the session and closes the overlay the target just opened. Deduplicating `players` in `runInput` (e.g. `[...new Set(...)]` — the session's `remaining` Set already assumes uniqueness) would fix it.

## Minor

- `library/tutorials/Functions - INPUT/0.json` — the `FLIP` on collection `["checker"]` in the `jkmt` routine still references a widget that doesn't exist in this variant (flagged last round); looks like debug leftover.
- `client/js/serverstate.js` — `flushDelta` duplicates `sendDelta`'s body verbatim; `sendDelta` could delegate to `flushDelta` to keep one copy of the send logic.
- `updateInputBlockOverlay` correctly avoids stealing the screen from a player whose own input overlay is open, but nothing re-checks when that overlay closes — the player then never sees the "waiting for input" overlay for a still-pending block. Cosmetic.
- Same-name multi-tab players: `this.players.find(p=>p.name === name)` delivers `showInput` to only the first connection with that name. Known limitation, probably fine, but worth a mental note since the busy check on that one tab now cancels the whole session.
- Still no test coverage for the new five-message protocol. The two bugs above (mixed reachable/unreachable targets, duplicate names) are exactly the kind of thing a small server-side unit test of `requestInput`/`inputResult` could pin down without needing multi-client TestCafe.

VTT-AI-Agent (commented):
## TL;DR

Follow-up review of the current head (`5fbda2f6`). The fixes from the previous rounds hold up, `npm test` passes, and the protocol is in much better shape. One remaining correctness gap worth fixing: when *some* (but not all) of the requested target players are unreachable, the server silently drops them and the initiator's routine hangs forever waiting for a result that will never come. The rest is minor.

## Correctness

1. **Partially unreachable targets hang the routine forever** — `server/room.mjs:845-858` (`requestInput`). Unknown names are filtered out of `targets`, but the initiator is only told (`cancelled: true`) when **no** target is reachable. With `player: ["Alice", "Bogus"]` (typo, or a player who disconnected just before the request arrived — the client-side `activePlayers` check in `widget.js:1433` is racy), Alice gets and answers her input, but the initiator's `session.remaining` still contains `"Bogus"` (`serverstate.js:642`), so `runInput` never resolves and the routine is stuck permanently. With `block: false` there is no cancel path at all — only a page reload recovers. Fix: in `requestInput`, send `inputResult { sessionID, player: name, cancelled: true }` for each requested name that has no connection (consistent with the existing all-missing behavior), or drop the special case entirely and always report unreachable names individually.

## Minor

2. **Duplicate player names: only the first connection is targeted** — `server/room.mjs:847` uses `this.players.find(p=>p.name === name)`. The same person being connected twice under one name (two devices/tabs) is a normal situation in VTT; the input then pops up on whichever connection happens to be first in `this.players` — possibly the tab they aren't looking at, which for them is indistinguishable from the feature not working. Sending `showInput` to all matching connections and accepting the first `inputResult` (hiding the rest) would be more robust; at minimum this is worth a wiki note.

3. **A displaced block overlay never reappears** — `client/js/serverstate.js:790-809`. `updateInputBlockOverlay` only runs on `inputBlock` messages. If the waiting overlay gets hidden by any other `showOverlay` call (the player opens the player list, or a second session sends them a `showInput`), it stays hidden after that overlay closes even though the block — and its cancel escape hatch — is still active. Similarly, players who join the room mid-session never receive the `inputBlock` message. Both are cosmetic given the initiator ends the block eventually, but the cancel button being unreachable weakens the escape hatch of finding 1.

4. **Dead exports** — `client/js/serverstate.js:554/570`: `sendDelta` was changed to `export` and `flushDelta` is exported, but neither is imported anywhere (`flushDelta` is only called within `serverstate.js`). The `sendDelta` change in particular is unnecessary diff noise and can be reverted.

5. **Still no automated coverage** for the five-message client↔server protocol (acknowledged in the earlier round as impractical for TestCafe; the manual 3-client verification is noted). The server-side session bookkeeping in `room.mjs` (`requestInput`/`inputResult`/`cleanupInputForPlayer`) is pure enough that a small jest test with stub `Player` objects would cover finding 1 and the cancel/disconnect paths cheaply — worth considering.

Carry-overs already deferred by the author (not re-litigating, just tracking): the `checker` FLIP no-op in `library/tutorials/Functions - INPUT/0.json:260`, and wiki documentation for `player`/`block` (including the "substitution happens on the initiator" semantics).

`npm test` passes on the branch head.

VTT-AI-Agent:
Addressed the two remaining bugs plus the cleanups (pushed in `b228e52`):

**1. Partially unreachable targets hung the initiator** — fixed. `requestInput` now cancels the whole session (`inputResult { cancelled: true }`) if *any* requested name has no connection, rather than dropping it silently while the initiator keeps waiting on it. This is consistent with the one-cancel-cancels-all rule and also covers the race you noted (a target that leaves between the client-side `activePlayers` check and the server processing the message). Nobody is asked unless the full set is reachable.

**2. Duplicate name in the list cancelled the session** — fixed. Names are de-duplicated on both sides (`[...new Set(...)]` in `runInput` and `requestInput`), so `player: ['A','A']` asks the connection once and `remaining` stays unique.

**Cleanups**
- `sendDelta` now delegates to `flushDelta` so the flush logic lives in one place.

**Tests** — added `tests/client/input-session.test.js` (runs under the existing jest project). It unit-tests the server routing with mocked player connections and covers exactly the cases raised across these reviews: all-reachable, partially-unreachable → cancel, duplicate names → asked once, one-cancel-closes-others, result echoing/completion, and rename-mid-session still delivering under the original name. Full suite: 12 passing. I also re-verified the three scenarios live in a browser (happy remote round-trip, duplicate name → keyed result, unreachable → no hang + routine aborts).

**Not changing**
- The `checker` FLIP no-op is in the human-authored tutorial content, which the maintainer asked me to leave untouched — flagging it here for the tutorial author rather than editing it.
- `updateInputBlockOverlay` not re-showing the wait overlay after a player closes their own input mid-block is cosmetic, as you noted; leaving it to avoid extra coupling.
- Same-name multi-tab delivery to the first connection remains a known limitation of addressing players by name.

VTT-AI-Agent:
Good call — implemented that (`2543b82`).

In a multi-player INPUT with `block: true`, the moment a player confirms their own selection they now switch to the **"Waiting for input from …" overlay (with the cancel button)** for whoever hasn't answered yet, rather than being dropped back to the board with no indication the game is still waiting.

Mechanically, the block is now tied to the session on the server: its `waitingFor` list shrinks as each target answers, so the just-answered player joins the watchers while the players still filling in their overlay stay unblocked. When the last player confirms (or anyone cancels), it all clears as before.

Verified end-to-end with three clients (Alice initiating; Bob + Carol answering): initially Bob/Carol have their input overlays and Alice sees "Waiting for input from Bob, Carol…"; after Bob confirms he sees "Waiting for input from Carol…" with a working Cancel button while Carol keeps her overlay; after Carol confirms the keyed results land and the overlay clears. Added unit tests for the shrinking `waitingFor` and the cancel routing; suite is at 14 passing.

Note on one edge: when the *initiator* is itself one of the listed players, the "waiting for" list is driven by the remote targets, so the initiator may not appear in it while it's still choosing locally — cosmetic, and the overlay still clears correctly once everyone (including the initiator) is done. Happy to tighten that if it matters.

VTT-AI-Agent:
Fixed both issues (`ea12e36`).

**Why it looked unchanged:** the waiting overlay never appeared for the player who *clicked the button* when that player was also one of the listed players (the common case — the clicker is passing too). That player answers their overlay locally, and the server was only shrinking the `waitingFor` list from the *remote* answers, so the clicker was never removed and never shown the waiting overlay. Now the initiator notifies the server (`inputBlockAnswered`) when it finishes its own overlay, and the block's `waitingFor` is decremented by the answering player's name for both local and remote answers — so whoever confirms first (initiator or a remote target) moves straight into the "waiting for the rest" overlay. Verified end-to-end with the initiator-as-participant case: after they confirm they see "Waiting for input from <the others>…" with a Cancel button, and it clears when everyone's done.

**CSS:** `#inputBlockOverlay` now matches the pile-card overlay — centered modal on the dark `#0008` backdrop, same modal width, no header bar and no close button. Just the waiting text and a Cancel button. (Screenshot looks like the pile modal minus the black ✕.)

Tests: added unit coverage for the initiator-answers-locally path; suite at 15 passing.

VTT-AI-Agent (commented):
## TL;DR

Review of head `ea12e36`. All findings from the previous rounds are now resolved — the partially-unreachable-target hang and the duplicate-name cancellation are fixed with an all-or-nothing check plus dedup on both sides, the multi-player block overlay correctly moves answered players (including the initiator-as-participant) into the "waiting for the rest" state, and the new `tests/client/input-session.test.js` pins down exactly the protocol edge cases raised across these reviews. `npm test` passes (15/15). I found no new bugs in the three commits since the last review; only tracking notes remain below.

## Verification notes

- **All-or-nothing target resolution** (`server/room.mjs` `requestInput`): unreachable or dropped-mid-race targets now cancel the whole session up front instead of leaving the initiator's routine awaiting forever, and the reachable targets are never shown a doomed overlay. The disconnect paths (`cleanupInputForPlayer`) cover targets or initiators leaving mid-session. Consistent with the one-cancel-cancels-all rule.
- **Initiator-participant block fix** (`ea12e36`): the `inputBlockAnswered` message plus the server-side shrinking `waitingFor` list resolves the case where the clicking player is also a listed player — they now join the waiting overlay after confirming, with the cancel escape hatch. The `typeof variables.playerName == 'string'` guard at the end of `evaluateRoutine` also means the multi-player keyed result object can never trigger an accidental rename — the `delete result.variables.playerName/playerColor` protection in the single-remote branch is coherent with it.
- **Tests**: the jest suite covers all-reachable, partial-unreachable → cancel, duplicate names, one-cancel-closes-others, result echoing, block-overlay transitions (remote answer, initiator local answer, spectator cancel routing) and rename-mid-session. Good coverage for the pure server routing, which is where the tricky bookkeeping lives.

## Tracking (known/deferred, for the maintainer's call — not blocking)

- Wiki documentation for `player`/`block` still needs writing before merge, including the two documented subtleties: `${...}` substitution in fields happens on the *initiator* (so `value: "${playerName}"` pre-fills the initiator's name on the target), and a remote single-target INPUT intentionally cannot rename/recolor anyone.
- Same-name multi-tab players: `showInput` goes to the first matching connection only (acknowledged limitation).
- Cosmetic: a displaced block overlay doesn't reappear until the next `inputBlock` message, and mid-session joiners never see it.
- The `FLIP` on the nonexistent `checker` collection in `library/tutorials/Functions - INPUT/0.json` (human-authored tutorial content, left for the tutorial author).

The PR is a draft awaiting @ArnoldSmith86's review; from the automated side I consider it in good shape.

VTT-AI-Agent:
Fixed the broken merge in `e5c980be`.

> I did a manual conflict resolution while merging main into the branch. Now the room does not load. A possible problem is that there are two exports of `flushDelta` in serverstate.js.

That was exactly it: the conflict resolution kept both main's new `flushDelta` (which since #3005-era changes on main does precisely what this branch needed — forcing the delta out mid-batch, preserving the delta cause) *and* this branch's older `sendDelta`/`flushDelta` split. Two `export function flushDelta` declarations are a SyntaxError in the concatenated client module, so the room never loaded — which is also why every TestCafe job failed. I removed the branch's duplicate and restored main's `sendDelta` (private again, as on main); `runInput` now simply uses main's `flushDelta`.

Verified locally: jest 15/15, TestCafe `editor.js` and `compute-1.js` pass, and a two-client Playwright round-trip of the feature itself (Alice clicks → Bob gets the INPUT overlay, Alice sees the block overlay with cancel → Bob confirms → the routine resumes with the answer and the block clears, no console errors).

The CodeQL `Math.random` alert points at old code from commit `05d2d3e` — the client token has used `performance.now()`/`timeOrigin` for a while now, so it should clear on re-analysis of this head.

VTT-AI-Agent:
Fixed the room not loading after the manual merge of `main` (`e5c980b`):

> I did a manual conflict resolution while merging main into the branch. Now the room does not load. A possible problem is that there are two exports of `flushDelta` in serverstate.js.

That was exactly it — the conflict resolution kept both main's `flushDelta` and this branch's older `sendDelta`/`flushDelta` split, leaving two `export function flushDelta` declarations in `client/js/serverstate.js`. Since the client JS is concatenated into a single module, the duplicate export is a SyntaxError and no room could load (this is also why every TestCafe job and two jest suites failed on the merge commit).

Fix: restored main's versions verbatim — the exported `flushDelta` (which forces the accumulated delta out mid-batch and preserves the delta cause) and the private `sendDelta`. That's exactly the semantics `runInput` needs before a remote INPUT, so the feature code is unchanged.

Verified locally: jest 15/15, TestCafe `editor.js` and `compute-1.js` pass, and a two-client Playwright run of the remote INPUT round-trip (initiator clicks → target gets the overlay, initiator sees the block overlay with cancel → target confirms → result lands in the routine and the block clears) works with no console errors.

VTT-AI-Agent (commented):
## TL;DR

Review of head `e5c980b` (post-merge of main). The merge is now clean — the duplicate `flushDelta` export is gone, the feature correctly reuses main's exported `flushDelta`, tutorials are byte-identical to main again, and jest (15/15) plus all TestCafe CI jobs pass. One actionable item remains: **the CodeQL check is red with a new high-severity alert on this PR's code**, and it did not clear on re-analysis as hoped — it needs either a trivial code tweak or an explicit dismissal before merge.

## Findings

1. **CodeQL: `js/insecure-randomness` (high) at `client/js/serverstate.js:662`** — the earlier comment predicted the alert would clear because `inputClientToken` uses `performance.now()`, but CodeQL is flagging the *sessionID* construction: `` `${playerName}-${inputClientToken}-${++inputSessionCounter}` ``. `playerName` defaults to `'Guest' + Math.floor(rand()*1000)` (`client/js/overlays/players.js:5`), and CodeQL taints the ID through that. In context I consider it a false positive — the sessionID is not a secret (it's sent to every target via `showInput` and broadcast as the `blockID` in `inputBlock`, and room membership is the trust boundary everywhere in VTT; any client can already mutate arbitrary room state). But the check is failing the PR, so either:
   - dismiss the alert as won't-fix with that rationale, or
   - trivially silence it: build `inputClientToken` from `crypto.randomUUID()` and drop `playerName` from the sessionID string (it adds no uniqueness the token doesn't already provide).

2. Nothing else new. The three commits since my last review check out:
   - `86a27ddc` (overlay restyle with "Waiting on other players" header + guidance text) is purely cosmetic CSS/HTML.
   - `d4ae47f3` reverts all `library/tutorials/` changes — `git diff main` shows the library untouched, which also retires the earlier `checker`-FLIP tracking item.
   - `e5c980be` correctly resolves the broken merge: one `export function flushDelta` remains (main's version), `sendDelta` is private again, and `runInput` flushes via main's `flushDelta` — the semantics the feature needs.

## Tracking (known/deferred, maintainer's call — not blocking)

- Wiki documentation for `player`/`block` still needs writing before merge, including the documented subtleties (`${...}` substitution happens on the initiator; a remote single-target INPUT intentionally can't rename/recolor).
- Same-name multi-tab players: `showInput` goes to the first matching connection only.
- Cosmetic: a displaced block overlay doesn't reappear until the next `inputBlock` message; mid-session joiners never see it.

This PR touches engine files (`client/js/serverstate.js`, `client/js/widgets/widget.js`), which are excluded from beta auto-deployment by design, so no `beta` label was added.

VTT-AI-Agent:
Added a small follow-up commit (3860d542) updating the edit-mode validator (`validator/validate_gamefile.js`).

Since this PR adds the `player` and `block` parameters to the **INPUT** routine operation, the validator's per-operation property list needed them too — otherwise `validate_gamefile.js` reports valid INPUT operations that use `player`/`block` as `INPUT has unknown property 'player'` (a false positive shown when debugging games in edit mode). I added:

- `player` — accepts a single name, a list of names, or `null` (same validator shape as AUDIO's `player`)
- `block` — boolean

Verified against a sample game file: valid INPUT ops using the new params now pass, while genuinely unknown properties and a non-boolean `block` are still correctly flagged.

VTT-AI-Agent:
This review was conducted at `e5c980b`; its one actionable finding — the CodeQL `js/insecure-randomness` alert on the session ID — was already fixed in the meantime:

- `32ed5f0` dropped the `Math.random`-tainted guest `playerName` from the session ID and switched the per-load token to `crypto.randomUUID()` (exactly the trivial fix suggested).
- `f396e4b` then swapped `randomUUID` for `crypto.getRandomValues` because `randomUUID` is secure-context-only and broke every TestCafe run behind the TestCafe proxy; the token stays crypto-sourced so the alert remains resolved.

**CodeQL is green on the current head** (`3860d54`), as are all TestCafe jobs and the build. The head also gained a validator-upkeep commit registering INPUT's new `player`/`block` properties in `validator/validate_gamefile.js` so edit-mode doesn't flag them as unknown. Agreed on the labels — `enhancement` + `routine operations` fit, and `beta` stays off since the PR touches engine files.

VTT-AI-Agent (commented):
## TL;DR

Review of head `4c254a1e`, the four commits since my last round. The previously-open CodeQL `js/insecure-randomness` alert is now resolved correctly (session token switched to `crypto.getRandomValues`, decoupled from `playerName`), the validator gained the missing `player`/`block` entries for `INPUT`, and the new tutorial variant validates cleanly. All CI checks (jest, CodeQL, TestCafe) are green. I found nothing new to flag.

## Verification notes

- **CodeQL fix** — `client/js/serverstate.js`: `inputClientToken` is now built from `crypto.getRandomValues(new Uint8Array(16))` instead of being derived from `playerName` + `performance.now()`, and `sessionID` no longer includes `playerName` at all. This removes the tainted-input path CodeQL was flagging while keeping session IDs unique per page load; doesn't touch the seeded game RNG, so determinism for shuffles etc. is unaffected.
- **Validator upkeep** (`validator/validate_gamefile.js:702-703`) — `player` and `block` are now registered under `INPUT`'s `operationProps`, matching the shapes actually accepted in `widget.js` (`player`: null/string/string-array; `block`: boolean). Ran `node validator/validate_gamefile_node.js "library/tutorials/Functions - INPUT/3.json"` locally — no errors.
- **Tutorial content** — the large diffs in `library/tutorials/Functions - INPUT/{1,2,3}.json` are the promised new variant demonstrating `block`/`player` plus edits to the existing variants; `0.json`'s diff is just editor metadata reordering (`gameSettings` moved after `info`), not a behavior change.

## Carried-over tracking (non-blocking, from earlier rounds)

- Wiki documentation for `player`/`block` still needs writing before merge (substitution-happens-on-initiator semantics, remote single-target INPUT can't rename/recolor).
- Same-name multi-tab players: `showInput` still goes to the first matching connection only.
- Cosmetic: a displaced block overlay doesn't reappear until the next `inputBlock` message; mid-session joiners never see it.

This PR makes a genuine (not incidental) rewrite of `INPUT` handling in `client/js/widgets/widget.js` and session/delta plumbing in `client/js/serverstate.js` — both are real engine files whose changed code paths run for every `INPUT` operation, existing games included — so per the engine-exclusion rule I did not add the `beta` label.

96LawDawg:
@ArnoldSmith86, this should be ready for approval.

VTT-AI-Agent (approved):

VTT-AI-Agent (commented):
## TL;DR

The remote/multi-player INPUT protocol is generally well covered and all recorded CI checks are green, but multi-player `choose` has a collection-merging bug: a shared output collection retains its old contents instead of being replaced by the players' combined selections. This can make the next routine operation act on every candidate rather than only the chosen widgets.

## Finding

1. **Multi-player `choose` appends selections to a pre-existing output collection** — `client/js/widgets/widget.js:1544-1550`. `evaluateInputOverlay` returns only the widgets selected into each field's output collection, and the single-player path correctly replaces the routine collection via `Object.assign`. The multi-player path instead initializes every merge from `collections[c]`, so if the output collection already exists it keeps all of those widgets and then appends each player's selections (including duplicate references to selected widgets). A minimal case is selecting candidate cards into `DEFAULT`, then using an array-valued `player` INPUT with a `choose` field whose source/output both default to `DEFAULT`: after confirmation, `DEFAULT` contains every candidate plus the chosen cards, so a following `MOVE` moves the entire candidate set. Accumulate the per-player results into a fresh collection for each output name and assign that union after all results are processed; a client-side regression test for this case would complement the new server-routing tests.

## Verification

The tutorial variant validates cleanly with `validate_gamefile_node.js`; I could not rerun Jest locally because dependencies are not installed in this fresh clone, but the PR's Jest, TestCafe, library, server-start, and CodeQL checks all passed at the merged head.

This PR is engine-ineligible for the `beta` label because `client/js/widgets/widget.js` changes INPUT routine semantics and `client/js/serverstate.js` changes session/state-flush behavior, so I did not add it.

VTT-AI-Agent (commented):
## TL;DR

The merged head still has the previously reported multi-player `choose` collection bug, and this pass found a second client-side deadlock when an initiator who is also answering gets renamed while the modal is open. CI is green, but the added tests cover server routing rather than these two client-side session/result paths.

## Findings

1. **Renaming an initiator who is answering locally leaves `runInput` pending forever** — `client/js/serverstate.js:678-729`. `includeSelf` and `session.remaining` capture the initiator's name when the session starts, but the local completion callback later uses the live imported `playerName` for both `session.results[playerName]` and `session.remaining.delete(playerName)`. Another connected player can rename the initiator from the Players overlay while the INPUT modal is open. After that rename, clicking Go deletes the new name from a set that still contains the old name, so `session.complete()` never resolves; `evaluateRoutine` remains awaited inside its open batch and subsequent local deltas stop flushing until reload (or, with a block overlay, until someone cancels). Capture the local participant name at session creation and use that stable value throughout the callback; a regression test should rename the local participant before resolving `showLocal`.

2. **Multi-player `choose` does not replace its output collection with the combined selections** — `client/js/widgets/widget.js:1544-1550`. Each overlay result contains only that player's selected widgets, and the single-player path replaces the corresponding collection, but the multi-player path starts from `collections[c]` and appends every result. If a chooser uses an existing source/output collection such as `DEFAULT`, every original candidate remains in the collection, so a following `MOVE` can act on all candidates instead of only the selected ones; if two players select the same widget, it is also inserted twice despite the comment calling this a union. Build each output collection fresh from the returned selections (and de-duplicate it) before assigning it back.

## Verification

The tutorial variant validates cleanly with `validate_gamefile_node.js`, and all recorded Jest, TestCafe, library, server-start, and CodeQL checks passed at `f5c6baf`. Local Jest could not be rerun because this fresh clone has no installed dependencies.

I did not add the `beta` label because this PR is engine-ineligible: it changes `INPUT` routine semantics in `client/js/widgets/widget.js` and session/delta behavior in `client/js/serverstate.js`.

PR #2659, opened 2025-08-20
Public Library - Add Silent Spectre

96LawDawg:
Similar to Meister Makatsu. VTT implementation by Robin Harn.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2659/pr-test (or any other room on that server)


96LawDawg:
Not sure why the validator said the image is bigger than 50000.  It is not.

PR #2671, opened 2025-09-16
rely less on `externalURL`

ArnoldSmith86:
The most common issue when trying to run a VTT server, is not setting `externalURL` correctly.

This PR eliminates usage of `externalURL` in the client code making mostly everything work without it being set correctly.

Mostly game sharing stuff is partly broken now when not setting it correctly.


96LawDawg (approved):
Mostly blind approval. Brief code scan. Server version still works. Did not test stand alone version at all.

PR #2677, opened 2025-09-27
enabled pinch-to-zoom on mobile

ArnoldSmith86:
So far I only tested this in my weird browser but this is looking pretty promising and I do not see why we should not just allow using it.

It is even still possible to move multiple widgets with multiple fingers.

### DIFF
--- a/client/css/layout.css
+++ b/client/css/layout.css
@@ -773,7 +773,6 @@ div[icon] .buttons {
   width: var(--roomWidth);
   height: var(--roomHeight);
   background: var(--roomColor);
-  touch-action: none;
   overflow: hidden;
 }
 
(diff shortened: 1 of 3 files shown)


96LawDawg:
On my iPhone, I zoom in a little, the screen refreshes.  Zoom in again and get "A problem repeated occurred on ..." with the URL.

ArnoldSmith86:
Hmm, weird. Especially because it doesn't actually change any code. Only HTML and CSS. I'll try to investigate.

96LawDawg:
Multi-touch move works on production (was using the opening screen in PL Dominoes).  That was also the game that was crashing for me in this PR. There is almost nowhere to touch there and just get background. So I opend Agon.  On production, that also works.  On this PR, if I touch 2 of the pieces, one jumps across the screen (although it can still be moved).  Zooming in on background works up to a point, then the room does something strange.  I'll try to make a video.

ArnoldSmith86:
Superseded by #2690.

PR #2678, opened 2025-09-27
added 2x zoom mode

ArnoldSmith86:
While #2677 should be nice for zooming on mobile, this PR is one idea for zooming in desktop browsers. A button in the toolbar toggles a 2x zoom mode that applies to the room area only. It automatically pans depending on the mouse position.

This is a one-prompt-AI-implementation (crazy times). If we like this, I would make some changes:
- Only 2x seems too harsh. Maybe make it a slider or add 1.5x or something.
- The room should not pan further than its area. It currently reveals the background when the cursor is outside the room area.
- It should not require the mouse to move all the way to the edge to reach it. It should show the right edge of the room when the mouse is only 80% there or something.

I got the AI to fix the last two things but it made widgets jump when moving them. But it was kinda close and definitely an improvement for the panning.

### DIFF
--- a/client/css/layout.css
+++ b/client/css/layout.css
@@ -23,6 +23,9 @@
   --positiveInput: #009502;
   --negativeInput: #950000;
   --neutralInput: #1f5ca6;
+  --roomZoom: 1;
+  --roomPanX: 0px;
+  --roomPanY: 0px;
 }
 
 body {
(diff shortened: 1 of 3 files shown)


96LawDawg:
Yes to this idea and to the improvements.

ArnoldSmith86:
It should be a lot better now.

96LawDawg:
This is definitely playable and a nice addition to the toolbox of things.  The only small issue: scroll all the way to the right, move the cursor up off the screen in the menu section of the browser, move the cursor all the way to the left, then back down into the screen.  The whole image just drastically jumps back to the far left.  I don't have a solution idea and this is not wrong.  It is just jarring.

PR #2690, opened 2025-10-14
added pan and zoom

ArnoldSmith86:
This is very similar to #2678 except that you can also zoom with your mousewheel and page-down/up and the panning is not automatic but can be done while dragging non-movable widgets.

On mobile it is supposed to work by pinching-to-zoom and then panning works the same.

TODO
- [x] implement mobile
- [x] check rest of AI code
- [x] fix mousewheel in overlays
- [x] fix editor
- [x] I feel clicking the scrollwheel should reset to 1x
- [x] check that the responsive toolbar can handle the new button (unless we want to remove it)
- [x] fix zoomout button in editor (reset new zoom)
- [x] reset zoom level to 1 when loading a new game
- [x] in edit mode, drag to multi-select conflicts with zoom. Should probably default to using drag to multi-select and make zooming optional. But only do one or the other.
- [x] Fix modal/overlay scrolling on mobile
- [x] moving two widgets with two fingers also triggers the pinch-to-zoom
- [ ] reset when loading new state leads to resetting when a conflict is triggered ... not the worst thing in the world but would be better if that case didn't reset
- [x] speed limit for mousewheel
- [x] pinch-to-zoom in editor
- [x] zooming in portrait editor (JSON above room) changes room area

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2690/pr-test (or any other room on that server)

### DIFF
--- a/client/js/editor/toolbar/grid.js
+++ b/client/js/editor/toolbar/grid.js
@@ -102,7 +102,10 @@ class GridButton extends PersistentToolbarToggleButton {
       $('#gridOverlay > .modal').appendChild(document.createElement('div').appendChild(cancelButton));
     } else {
       $('body').classList.remove('gridLines', 'gridLines33', 'gridLines25', 'gridLines20');
-      document.querySelector('#gridOverlay').style.display = 'none';  
+      const gridOverlay = document.querySelector('#gridOverlay');
+      if (gridOverlay && gridOverlay.style.display !== 'none') {
+        showOverlay('gridOverlay');
+      }  
     }
   }
 }


96LawDawg:
Tested on PC with Chrome, Edge, and Firefox. Tested on iOS Safari.  Two issues:

On PC, start with the grid (on the top toolbar off). Then click grid to on and immediately click it to off again. The client freezes until you leave Edit Mode.

On iOS Safari, pinch to zoom does not work in Edit Mode at all.

96LawDawg (approved):
Approved through test

PR #2693, opened 2025-10-16
fixed add/remove a card to/from each card type buttons creating empty tiles

ArnoldSmith86:
Just clicking all the buttons did not handle _async_/_await_ correctly and screwed things up. Simply pressing Remove All in production until no cards are left in a pile would cause an empty pile. But only _if_ the last press removed more than one card. Adding cards is similary broken if it would have to create a new pile.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2693/pr-test (or any other room on that server)

### DIFF
--- a/client/css/layout.css
+++ b/client/css/layout.css
@@ -23,6 +23,10 @@
   --positiveInput: #009502;
   --negativeInput: #950000;
   --neutralInput: #1f5ca6;
+  --zoom: 1;
+  --roomZoom: 1;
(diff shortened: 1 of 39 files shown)


96LawDawg (approved):
Approved through test

ArnoldSmith86:
Why does this one need a newer Node version for TestCafé and the other PRs don't?! :grin:

PR #2696, opened 2025-10-17
Fix grid toggle button

### LINKED ISSUE #2695: Grid in Edit Mode can cause client freeze
On PC or Mac, start with the grid (on the top toolbar) off. Then click grid on and immediately click it to off again. The client freezes until you leave Edit Mode.

96LawDawg:
Fixes #2695

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2696/pr-test (or any other room on that server)


ArnoldSmith86 (approved):
Not entirely sure why it works but it looks like it does.

PR #2698, opened 2025-10-18
Add ZOOM function

96LawDawg:

- [ ] **Decide the ZOOM defaults** — Whether `players: []` should be the default (today it zooms the triggering player with `disableUserControls: true`). ([discussion](https://github.com/ArnoldSmith86/virtualtabletop/pull/2698#issuecomment-4925470250))
- [ ] **Decide whether per-player viewport overrides are saved** — Whether `perPlayer` viewport overrides belong in saved games or should stay session-only. ([discussion](https://github.com/ArnoldSmith86/virtualtabletop/pull/2698#issuecomment-4925470250))

This adds a new `ZOOM` function for game designers to use.

Parameters:

- level (defaults to 1)
- panX and panY (which default to center the board whatever the scale is)
- player (defaults to the player who triggered the routine; accepts any single or array of player names; explicitly setting `null` targets all players)
- disableUserControls (defaults to true: a plain ZOOM locks the targeted players' zoom/pan controls until another ZOOM unlocks them or the player unchecks "Allow game to override")

Adds new "zoom" save value in the `gameSettings` meta data. Saves the level, pan, and disableUserControl settings per player.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2698/pr-test (or any other room on that server)

To Do:

- [x] Add parameter to turn off panning? How would that work? Would it stay off until turned back on again by another ZOOM routine?
- [x] Add something to JSON editor to help calculate `panX` and `panY` based on a selected level/scale. The panX and panY are the upper left corner of the new zoomed in area. That may be what is needed, but you might also want the new relative center coordinates of the viewing area.
- [x] "Game override active" is not correctly activated unless you re-enable the override checkbox
- [x] panning cursor should not be displayed when you can't pan
- [x] using browser zoom on top of ingame zoom leads to a shifting viewport (probably in production as well) (also when just resizing the window)
- [x] `ZOOM` parameter for locking user zoom/pan? (basically the same as the first todo?)
- [x] ~~still allow zooming _in_ but not _out_? basically making `ZOOM` define new outer bounds?~~
- [x] "Loading room..." briefly flashes to game zoom level before actually showing the room
- [x] probably get rid of `prompt` completely now?
- [x] don't override zoom in editor
- [x] show something about the zoom in Game Settings panel?
- [x] going to editor and back restores wrong panning
- [x] unchecking "allow game to override" does not show "enable mouse wheel" checkbox
- [x] `disableUserControls` not working
- [ ] make `players: []` the default?
- [x] Reset zoom level 1 when loading new game unless zoom level is saved with game


96LawDawg:
I was working on the tutorial for this. We probably need a way to override user panning. If we are going to use ZOOM to control what a player sees and when, then we might want to disable panning.

96LawDawg:
Strange behavior with fractional zoom values.  In https://test.virtualtabletop.io/PR-2698/lawdawg, play the Functions - ZOOM draft tutorial. Click on the 4th button "No Prompt." It is set for zoom level 3.3. It bounces. That only happens for fractional values. Also, I went back and checked and this did not happen before the recent commits @ArnoldSmith86 made. I suspect it was something in https://github.com/ArnoldSmith86/virtualtabletop/pull/2698/commits/2da58285c752dd57699771c0e412cb722cd6f4d6

ArnoldSmith86:
> Strange bug. Go into edit mode, click on debug mode and use a button with ZOOM. Client crashes with console reporting uncaught promise reference error "clampedUnits is not defined." This only shows up when using zoom with the debugger.
> 
> Note: this also only happens after the most recent commits from @ArnoldSmith86. The original version of the code changes I made did not have this bug present. Not saying my code was better! Just pointing it out to help figure out where the problem is.

Yeah, the debug logging uses now undefined variables. Will fix later.

ArnoldSmith86:
> We probably need a way to override user panning.

Maybe just another confirmation when the user tries to pan/zoom? Something like "The game controls zoom/pan. Are you sure you want to change them and possibly break game immersion?"?

Also: when a game uses `ZOOM`, the designer probably intends that everyone (unless `players`) has the requested viewport. But if you (re)join after the `ZOOM`, it is just the default 1x. The room (maybe even a saved game) should probably remember the viewport settings and apply them to new players if matching `players`.

And maybe instead of prompts we add a checkbox to the viewport controls saying something like "Let the game override me" or whatever.

As usual, this is way more complicated than anticipated.. :grin:

VTT-AI-Agent:
Picked this branch back up (it had gone stale — ~8 months / 191 commits behind \`main\`, which is why it showed as conflicting and CI was red on the last run from Nov 2025):

- Merged current \`main\` into this branch. The only real conflict was in \`client/js/editor/sidebar/gameSettings.js\`, where `main` had reworked the legacy-mode checkbox toggle to use a `change` event + click-delegation instead of the older manual toggle handler, and added a condition to only show the legacy-modes intro text when the game actually has legacy modes set. Kept `main`'s version of both and layered this branch's Zoom/Pan Overrides summary UI on top — no functional loss on either side.
- Verified the two bugs flagged in review are already fixed on this branch: the `clampedUnits is not defined` debug-logging crash, and the `prompt: null` fractional-zoom bounce (the whole `prompt` confirmation mechanism was intentionally removed in favor of the "Allow game to override" checkbox).
- Found and removed a small regression: a later commit (`disableUserControls` addition) accidentally reintroduced the `prompt` field that had already been deliberately removed. It was dead code — never read by the server (`room.mjs#zoom`) or any client UI — so I dropped it again.
- `npm test` passes. Pushed as two commits: the merge and the prompt cleanup. Fresh CI is running now against the current `main`.

VTT-AI-Agent (commented):
## TL;DR

Nice feature and the client/server plumbing is mostly sound, but there are several correctness issues worth fixing before merge. The most important: loading a new game re-applies the *previous* game's zoom override due to message-handler ordering, and `applyServerZoomSetting()` force-snaps every player's viewport on every meta update even when `disableUserControls: false` (likely the root cause of the unchecked "disableUserControls not working" ToDo). The server handler also trusts client input completely.

## Findings

### 1. Switching games applies the previous game's zoom settings (stale `currentGameSettings`)

`client/js/zoom.js:23-24` registers its `state`/`meta` handlers at module top level, so they run during script evaluation — *before* the `onLoad`-registered handlers. Given the bundle order in `server/minify.mjs` (`serverstate.js` before `legacymodes.js`), the `state` callbacks fire in this order:

1. zoom.js's handler (fresh settings — fine),
2. `receiveStateFromServer` → `resetZoomAndPan()` (`client/js/serverstate.js:509`) → `applyServerZoomSetting(getCurrentGameSettings())` — but `currentGameSettings` is still the **previous** state's settings,
3. legacymodes.js finally updates `currentGameSettings`.

Concrete failure: load game A with a ZOOM override, then load game B without one — step 2 re-applies A's zoom/lock in game B. This also means the old "reset zoom to 1x on state load" behavior is silently gone whenever no override exists (matches the unchecked "Reset zoom level 1 when loading new game" ToDo). Consider passing the incoming state's settings into `resetZoomAndPan`, or updating `currentGameSettings` before `receiveStateFromServer` runs.

### 2. Every meta update snaps the viewport back, even when `disableUserControls: false`

`applyServerZoomSetting()` (`client/js/zoom.js:9-21`) unconditionally calls `setZoomLevel`/`setPan` whenever a target exists — it never checks `target.disableUserControls`. `sendMetaUpdate()` fires from ~20 call sites in `server/room.mjs` (player join/leave/rename, state saves, star toggles, …), so a player who was deliberately left free to pan (`disableUserControls: false`) gets yanked back to the game's zoom whenever anything touches meta. This is very likely the mechanism behind the "`disableUserControls` not working" ToDo: only force-apply when controls are disabled (or when the target actually changed).

### 3. A missing `gameSettings` breaks room loading entirely

`connection.onmessage` dispatches callbacks with no try/catch, and zoom.js's handlers run first. If `_meta.gameSettings` is ever undefined (a hand-edited or externally generated v≥18 save file — FileUpdater only creates `gameSettings` for files older than v18), `gs.zoom` at `client/js/zoom.js:12` throws and `receiveStateFromServer` never runs — the room never renders for anyone. `legacymodes.js` guards with `|| {}`; do the same here (`applyServerZoomSetting(args._meta.gameSettings || {})` or a guard inside the function).

### 4. `Room.zoom()` trusts client input completely (server/room.mjs:1282)

- `level`/`panX`/`panY` are persisted verbatim into `_meta.gameSettings` — any type or size. The 1–10 clamp and `Number.isFinite` checks exist only client-side (`widget.js` before sending, `applyServerZoomSetting` on receive), so any connected client can inject arbitrary junk that gets broadcast to everyone and baked into the room's save file. Clamp/coerce server-side.
- `players: ["__proto__"]` makes `this.state._meta.gameSettings.zoom.perPlayer["__proto__"] = payload` set the object's *prototype* instead of a key — the override silently doesn't apply and serialization gets weird. Filter/validate player names (or use `Object.create(null)`).
- Note the persistence consequence: `saveCurrentState_write()` copies `gameSettings` into saved/exported games, so `perPlayer` entries keyed by transient player names accumulate permanently in game files (and in anything shared to the public library), with no cleanup when players leave or rename. Worth deciding whether per-player overrides should be persisted at all, vs. kept as room-session data.

### 5. Defaults contradict the PR description and the JSON-editor template

`client/js/widgets/widget.js:2167` uses `setDefaults(a, { …, player: playerName, disableUserControls: true })`, but:

- The PR description says `player` defaults to **null** and `disableUserControls` defaults to **false**.
- The jsonedit template (`client/js/jsonedit.js`, `jeAddRoutineOperationCommands('ZOOM', …)`) uses `player: null`.

So omitting `player` targets only the triggering player (and writes a `perPlayer` entry under their name), while an explicit `player: null` targets everyone (`zoom.all`). That asymmetry will surprise game designers; other operations like `AUDIO` default `player: null`. Similarly, `disableUserControls` defaulting to true means a plain `{"func": "ZOOM", "level": 2}` locks every targeted player's zoom/pan — fine if intended, but then the description and eventual wiki docs need to say so.

### Minor

- `updateZoomUIState(gs)` mixes the fresh `gs` with stale `getCurrentGameSettings()` (via `isGameOverrideActive()`), so the "Game override active" banner can linger after overrides are removed until the next meta message.
- `resetZoomAndPan()` in non-edit mode with no override now does nothing at all — combined with finding 1, the reset semantics of this function no longer match its name.
- No test coverage: neither the jest validator tests nor the TestCafe suite exercise the new ZOOM operation or the server `zoom` message; per the repo conventions a TestCafe save exercising ZOOM would be valuable since this touches room state and client behavior.

VTT-AI-Agent:
Addressed the review. Went through each finding against the actual code (not just the description) before changing anything:

**#1 Stale `currentGameSettings` on game switch** — confirmed real, but only partially the mechanism described: an unrelated `main` commit (#2729, merged in earlier today) already updates the shared `currentGameSettings` synchronously at the top of `receiveStateFromServer`, before `resetZoomAndPan()` reads it — so the race window described doesn't fully reproduce anymore. Still fixed it properly rather than relying on that: `resetZoomAndPan()` now takes the incoming state's `gameSettings` directly as a parameter instead of going through the shared global at all, so it no longer depends on cross-file registration-order timing.

**#2 Every meta update re-snaps the viewport even with `disableUserControls: false`** — confirmed, this is exactly the unchecked "`disableUserControls` not working" ToDo. `applyServerZoomSetting()` now only force-applies when the game re-locks the view (`disableUserControls !== false`) or the target actually changed; `resetZoomAndPan()` (game load, or returning from edit mode) still always re-applies in full so "going to editor and back" keeps working.

**#3 Missing `gameSettings` breaks room loading entirely** — the server already guards this in `Room.setState()`/`FileUpdater`, so I couldn't get it to reproduce through normal loading, but the client-side `gs.zoom` dereference was still a landmine with no defense of its own. Added the same `gs && gs.zoom` guard `legacymodes.js` already uses, so a future server-side gap can't take the whole room down again.

**#4 Server trusts client input** — confirmed and fixed: `Room.zoom()` now rejects non-finite `level`/`panX`/`panY` and clamps `level` to `[1, 10]` before persisting/broadcasting, and filters `__proto__`/`constructor`/`prototype` out of `players` (bracket-assigning `perPlayer["__proto__"]` was silently repointing the object's prototype instead of adding a key). Left the "should `perPlayer` entries be persisted into saved games at all" question alone — that's a product decision, not a bug, and out of scope for a review fix.

**#5 Defaults contradict the PR description** — `player: playerName` / `disableUserControls: true` are unchanged. The PR's own ToDo list already has "make `players: []` the default?" unchecked, i.e. this is a decision you haven't made yet, not a bug — didn't want to unilaterally resolve it.

**Minor: UI staleness** — fixed as part of #1/#2: `updateZoomUIState()` now derives the override banner purely from the `gameSettings` it's passed instead of partly falling back to the stale global.

**Minor: no test coverage** — added jest coverage for ZOOM's input validation (out-of-range level, non-numeric pan, non-targeted player) in `tests/client/widget-zoom.test.js`. Getting there required two small fixes to make the file importable/testable under jest at all: `widget.js` wasn't importing `toServer` and `zoom.js` wasn't importing `$`, both true globals-via-concatenation in production but undefined under jest's real ES modules. I could not add TestCafe coverage — this sandbox has no browser TestCafe can drive (tried chrome-headless-shell, no luck), so I couldn't generate/verify state hashes locally. That's a legitimate gap; happy to add it if someone can run TestCafe against the PR-2698 preview and share hashes, or I can take another pass once CI is green and I can inspect its output.

Also checked off "`disableUserControls` not working" and "Reset zoom level 1 when loading new game unless zoom level is saved with game" in the PR description's ToDo list, since both are fixed now.

`npm test` passes locally (9/9). Pushed as a new commit on the same branch.

VTT-AI-Agent (commented):
## TL;DR

Re-reviewed the branch after the last round of fixes — the previously flagged issues (stale settings on game switch, meta-update snap-back, unvalidated server input) are resolved and `npm test` passes (9/9). Two correctness issues remain: the ZOOM operation's direct-apply path bypasses the edit-mode guard (the editor viewport gets zoomed/panned when a routine runs while the editor is open), and the new dedup logic means a re-issued identical `disableUserControls: false` ZOOM silently never applies to anyone except the triggering player.

## Findings

### 1. ZOOM's local-apply path ignores edit mode (`client/js/widgets/widget.js:2210`)

Every other zoom-apply path guards on `edit` — `applyServerZoomSetting()` returns early (`client/js/zoom.js:17`), `resetZoomAndPan()` forces 1x/0,0 in edit mode, and all the input handlers check `!edit`. But the direct apply inside the ZOOM operation only checks the `allowGameZoomControl` localStorage flag:

```js
if(isTargetedPlayer) {
  const allowOverride = localStorage.getItem('allowGameZoomControl') !== 'false';
  if(allowOverride) {
    setZoomLevel(numericLevel);
    setPan(resolvedPan.x*scale, resolvedPan.y*scale);
  }
}
```

Routines do execute while the editor is open — the JSON sidebar module enables routine logging (`client/js/editor/sidebar/json.js:164`) and operations still run for real, not just as a trace. So clicking a button with a ZOOM op while editing zooms/pans the editor viewport, which commit d68cb234 ("always allow zooming in edit mode") and the "don't override zoom in editor" ToDo explicitly set out to prevent. Fix is one word: `if(isTargetedPlayer && !edit)`.

### 2. Re-issuing the same unlocked ZOOM never reaches other players (`client/js/zoom.js:25`)

The dedup added to fix "every meta update snaps the viewport back" keys on `JSON.stringify(target)`:

```js
if(target.disableUserControls !== false || targetKey !== lastAppliedZoomKey) {
```

Consider a "recenter view" button: `{"func": "ZOOM", "level": 2, "disableUserControls": false}`. First press works for everyone. Now player B pans away and someone presses the button again — the server stores a byte-identical payload, so on B's client `targetKey === lastAppliedZoomKey` and the snap is skipped. Only the triggering player recenters (via the direct apply in widget.js); every other targeted player's button silently does nothing from the second press on. Since routines run only on the triggering client, the meta broadcast is the *only* channel to other players, and it's now permanently deduped for unchanged payloads. A monotonic sequence number stamped by `Room.zoom()` into the payload (and included in the dedup key) would distinguish "unrelated meta update" from "the game re-issued this ZOOM".

### Minor

- The `onMessage('state', ...)` handler at `client/js/zoom.js:36` looks redundant now: `receiveStateFromServer()` already calls `resetZoomAndPan(args._meta.gameSettings)` (`client/js/serverstate.js:509`), which resets `lastAppliedZoomKey` and re-applies in full. Given the bundle order in `server/minify.mjs` (serverstate.js before zoom.js), the zoom.js handler fires second and just re-applies or no-ops. Dropping it would remove one of the two competing apply paths per state load.
- `Room.zoom()` now validates types, but still accepts an unbounded `players` array with arbitrary-length strings — any client can grow `_meta.gameSettings.zoom.perPlayer` without limit, and that object is persisted into saved/exported games. A cap on count and name length would be cheap insurance.
- The PR description still says `player` defaults to null and `disableUserControls` defaults to false; the code defaults are `player: playerName` and `disableUserControls: true` (`client/js/widgets/widget.js:2168`). Whatever the "make `players: []` the default?" ToDo decides, the description and the eventual wiki page should match the shipped behavior — `disableUserControls: true` meaning "a bare ZOOM locks the viewport" is the kind of default designers will trip over if undocumented.
- Test coverage note carried over: the jest tests cover input validation, but nothing exercises the happy path (a targeted ZOOM actually changing the level/pan) or the server round-trip; a TestCafe save using ZOOM would cover both once someone can run the suite against a browser.

VTT-AI-Agent:
Addressed round 3 (commit b0026e66):

**#1 ZOOM's direct apply ignores edit mode** — confirmed and fixed with the suggested one-word change: `if(isTargetedPlayer && !edit)`. Added a jest test that clicks a ZOOM widget with `edit` set and asserts the viewport doesn't move.

**#2 Re-issued identical unlocked ZOOM never reaches other players** — confirmed and fixed with the suggested approach: `Room.zoom()` now stamps an incrementing `seq` into the persisted payload. Since the client dedup key is `JSON.stringify(target)`, a re-issued ZOOM always produces a new key and re-applies, while unrelated meta broadcasts (which don't touch the payload) still don't snap the view back. No client change was needed.

**Redundant `state` handler in zoom.js** — agreed and removed. Verified the ordering concern is moot: `startWebSocket()` is only called from inside `onLoad` handlers, so serverstate's `state` handler is always registered before the first message can arrive; `receiveStateFromServer → resetZoomAndPan(gameSettings)` is now the single apply path on state loads.

**Unbounded `perPlayer` growth** — capped `players` at 100 entries and 128 chars per name in `Room.zoom()`.

**Defaults vs. description** — left the code defaults alone (the "make `players: []` the default?" ToDo is still an open maintainer decision) but updated the PR description to document the shipped behavior: `player` defaults to the triggering player, and `disableUserControls` defaults to true (a bare ZOOM locks the viewport).

**Happy-path test coverage** — now covered: a targeted ZOOM asserting the level actually changes, plus the edit-mode guard. Getting there needed stubbing four bundle-globals (`edit`, `scale`, `roomRectangle`, `widgetFilter`) in the jest setup, which is why it wasn't possible in the previous round. TestCafe coverage still pending a browser-capable environment.

`npm test`: 11/11 passing.

VTT-AI-Agent (commented):
## TL;DR

Re-verified the branch after the round-3 fixes (head `b0026e66`) — all previously flagged issues check out as fixed in the code, `npm test` passes locally (11/11), and CI is fully green. No blocking issues remain; one small new edge case in `getZoomTarget()` (prototype-inherited lookups for players named e.g. `constructor`) plus the known open maintainer decisions.

## Verification of prior findings

- **Edit-mode guard on ZOOM's direct apply** — fixed (`client/js/widgets/widget.js:2211`, `isTargetedPlayer && !edit`) and covered by a jest test that clicks a ZOOM widget with `edit` set.
- **Re-issued identical unlocked ZOOM never reaching other players** — fixed: `Room.zoom()` stamps an incrementing `seq` into the persisted payload (`server/room.mjs:1273`), so the client's `JSON.stringify` dedup key changes on every re-issue while unrelated meta broadcasts still don't snap the view back.
- **Redundant `state` handler in zoom.js** — removed; `receiveStateFromServer → resetZoomAndPan(args._meta.gameSettings)` is now the single apply path on state loads, and `toggleEditMode()`/editor zoom-out still work via the no-arg fallback to `getCurrentGameSettings()`.
- **Server input validation** — `Room.zoom()` rejects non-finite values, clamps level to 1–10, filters `__proto__`/`constructor`/`prototype` from `players`, and caps the array at 100 entries / 128 chars per name.
- Tests: 11/11 jest pass locally; all TestCafe CI shards, CodeQL, and the production build are green.

## New (minor)

### `getZoomTarget()` hits `Object.prototype` for players named `constructor`/`toString`/… (`client/js/zoom.js:13`)

```js
return (zoomSettings.perPlayer || {})[playerName] || zoomSettings.all || null;
```

`perPlayer` is a plain JSON object, so for a player literally named `constructor` (or `toString`, `valueOf`, …) this lookup returns the inherited prototype member — a truthy function — instead of falling through to `zoomSettings.all`. Consequences for such a player whenever *any* zoom override exists in the room: `applyServerZoomSetting()` treats the garbage target as invalid (so the real `all` override never applies to them), and `updateZoomUIState()`/`isGameOverrideActive()` see a truthy target with `disableUserControls` undefined → the "Game override active" banner shows and their wheel/pan/pinch controls lock even though nothing targets them. The server's unsafe-key filter only protects the *write* path. A `Object.hasOwn(perPlayer, playerName)` guard in `getZoomTarget()` closes it.

## Still open (maintainer decisions, carried over)

- The "make `players: []` the default?" ToDo: current defaults are `player: playerName` / `disableUserControls: true` (a bare ZOOM locks only the triggering player's viewport); the PR description now documents this, but the wiki page should too when this merges.
- Whether `perPlayer` entries (keyed by transient player names) should be persisted into saved/exported games at all, vs. kept as room-session data.
- A TestCafe save exercising ZOOM would still be valuable since this touches room state sync; jest now covers validation, targeting, edit-mode, and the happy path, but not the server round-trip through a real browser.

VTT-AI-Agent:
Addressed round 4: fixed `getZoomTarget()` to use an own-property check (`Object.prototype.hasOwnProperty.call`) on `perPlayer`, so a player named `constructor`/`toString`/etc. no longer inherits a truthy value from `Object.prototype` and gets falsely locked. (The server side already filters those names out of the keys it *writes*, so this is purely defensive on the read path for hand-edited/legacy state.) Pushed as 51e4a602; jest still 11/11. No other findings this round.

VTT-AI-Agent (commented):
## TL;DR

Re-verified the branch at head `51e4a602` — no new commits since the last round, the final fix (own-property guard in `getZoomTarget()`) is in place and correct, and `npm test` passes locally (11/11). **No blocking issues remain**; what's left is a couple of trivial cleanups and the known open maintainer decisions.

## Verification

- `getZoomTarget()` now uses `Object.prototype.hasOwnProperty.call(perPlayer, playerName)` (`client/js/zoom.js:14`), so a player named `constructor`/`toString`/etc. no longer inherits a truthy prototype member and gets falsely locked. This closes the last flagged finding.
- All earlier fixes still check out in the current diff: the `!edit` guard on ZOOM's direct apply (`client/js/widgets/widget.js:2196`), the server-stamped `seq` that lets re-issued identical ZOOMs re-apply on other clients (`server/room.mjs:1300`), input validation/clamping and the 100-entry/128-char cap in `Room.zoom()`, and the single apply path on state loads via `resetZoomAndPan(args._meta.gameSettings)`.
- The "Remove All Overrides" button round-trips correctly: `setGameSettings` (`server/room.mjs:997`) accepts the full settings object, and on the resulting meta broadcast `updateZoomUIState()` unlocks previously-locked clients.
- The `zoom_from_viewport` JSON-editor helper's math checks out against what `setZoomLevel`/`setPan` write (`--zoom`, `--scale`, `--roomPanX/Y`): pan px = `-left * level * scale`, so `left = -panPx / scale / zoom`.
- Jest: 11/11 passing locally.

## Trivial cleanups (take or leave)

- `client/css/layout.css:707-716` still styles `#lockZoomButton`, but the button was removed from `room.html` — dead rules.
- The old `zoomLocked` localStorage key is orphaned: users who had locked their zoom will find wheel/pinch zoom re-enabled, since the replacement `enableUserZoom` key defaults to true. Probably fine, just noting the one-time behavior change.

## Still open (carried over, maintainer decisions)

- The "make `players: []` the default?" ToDo — current defaults (`player: playerName`, `disableUserControls: true`) are documented in the PR description; the wiki page should match on merge.
- Whether `perPlayer` entries keyed by transient player names should be persisted into saved/exported games.
- A TestCafe save exercising ZOOM would still be valuable since this touches room state sync (needs a browser-capable environment to generate state hashes).

VTT-AI-Agent:
Addressed round 5 (commit db57c44e): removed the dead `#lockZoomButton` CSS in `layout.css` (the element was replaced by the zoom controls panel in this PR) and added a one-line cleanup of the orphaned `zoomLocked` localStorage key it used. The remaining noted items — `players: []` default, whether to persist per-player overrides, and TestCafe coverage — are maintainer decisions / need a browser env, so I've left them for @ArnoldSmith86. jest still 11/11.

PR #2701, opened 2025-10-19
Added doubleClickRoutine to widgets.

mousewax:
Added doubleClickRoutine to widgets. There is a 200ms delay on widgets with a doubleClickRoutine defined otherwise clickRoutine fires immediately.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2701/pr-test (or any other room on that server)

Proposed wiki update on Functions page under Predefined Routines:

doubleClickRoutine: A doubleClickRoutine may be placed in any widget (except decks). It works similarly to `clickRoutine` except that to be activated, the user must double click the same widget within 350 milliseconds. If a widget does not have a `doubleClickRoutine` defined, a `clickRoutine` on it will activate immediately. If a widget has both, then the `clickRoutine` will fire after the first click unless a second click happens within 350ms.

### DIFF
--- a/client/js/editor/sidebar/undo.js
+++ b/client/js/editor/sidebar/undo.js
@@ -2,6 +2,7 @@ class UndoModule extends SidebarModule {
   constructor() {
     super('undo', 'History', 'Undo changes from editing or playing.');
     this.lastRenderedIndex = -2;
+    this.latestEntryDOM = null;
   }
 
   onClose() {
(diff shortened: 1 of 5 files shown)


96LawDawg:
Everything seems to work as expected.  I think 200ms is too fast.  500ms is too slow.  I didn't play with any other values, but I'm thinking 300-350 would be the sweet spot. For me. On my devices.

ArnoldSmith86 (approved):

96LawDawg:
Actually, hold off on merging this.  I'm working on a tutorial update.  Won't take long.

96LawDawg:
Tutorial updated. Wiki entry drafted. Ready to go whenever you want to merge it @mousewax.

PR #2702, opened 2025-10-20
feat: Add DELAY routine command

mousewax:
This commit introduces a new DELAY command to the client-side routine engine. This command allows routines to force a synchronization of the current game state with the server and then pause for a specified duration.

The DELAY command accepts a time parameter, which specifies the pause duration in milliseconds.

To ensure that pending changes are sent immediately, even when the routine is executed within a batch (batchStart/batchEnd), a new flushDelta function has been implemented. This function temporarily overrides the batching mechanism to send the current delta to the server.

The UPDATE command now performs the following actions:
1.  Calls flushDelta() to send all pending state changes.
2.  Pauses the routine's execution for the duration specified by the wait parameter.

This enables the creation of more complex, multi-step animations and actions that are visible to all players in real-time.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2702/pr-test (or any other room on that server)

### DIFF
--- a/client/js/editor/sidebar/undo.js
+++ b/client/js/editor/sidebar/undo.js
@@ -47,8 +47,10 @@ class UndoModule extends SidebarModule {
 
     this.protocol = [...getUndoProtocol()];
 
-    if(this.latestEntryDOM)
-      this.latestEntryDOM.innerText = `${this.lastRenderedIndex+1} - ${this.protocol[this.lastRenderedIndex].delta.c || 'unknown'} (${Object.keys(this.protocol[this.lastRenderedIndex].delta.s).length} widget${Object.keys(this.protocol[this.lastRenderedIndex].delta.s).length == 1 ? '' : 's'} changed)`;
+    if(this.latestEntryDOM) {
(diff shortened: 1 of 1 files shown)


96LawDawg:
It has an interesting effect on the History log. It creates a separate history entry for each thing. For example, in my poker room I added a delay when dealing a face down card to every player which previously just appeared there. Now it looks like they are being dealt one at a time.  Cool.  In the original history entry I got "LawDawg clicked widgetID (5 widgets changed)." Now I get "LawDawg clicked widgetID (1 widget changed)" then several more entries like: "unknown (1 widget changed)."

I don't know if I like that or not. It does allow me to go back and look at and reverse each individual card from being dealt. Probably good for debugging, but not for routine usage of history where I'm in the middle of a game and just trying to undo someone's accidental button press.

ArnoldSmith86:
I made a few changes. Please test if everything is still working. Note that I renamed `time` to `milliseconds`.

ArnoldSmith86 (approved):
The last commit probably broke something but YOLO!

PR #2707, opened 2025-10-23
Allow pan in zoomed edit mode with spacebar

96LawDawg:
This PR adds a check for the spacebar being held down in edit mode. If the spacebar is held down, then the multi-select rectangle is suppressed. If the zoom level > 1, and the spacebar is held down, then left-clicking the mouse and dragging enables panning no matter where the cursor is. This applies even over a movable widget because if the designer is holding down the spacebar, then they want to pan, not move widgets.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2707/pr-test (or any other room on that server)


VTT-AI-Agent:
Updated the branch to current `main` and hardened the Space-pan gesture:

- only arms in edit mode outside overlays and editable controls
- keeps widget/selection input suppressed through the matching mouseup, including when Space is released mid-drag
- adds grab/grabbing cursor feedback
- adds a TestCafe regression covering panning over a movable widget without moving or selecting it

Validation: `npm test -- --runInBand` (15/15) and the Chromium editor TestCafe coverage both pass.

VTT-AI-Agent:
Fixed the Firefox CI failure in `8af6258e`. The regression test now explicitly waits for the asynchronously loaded edit-mode DOM before dispatching the Space-drag gesture; Firefox had been running it early as a normal play-mode widget drag. The targeted Chromium test still passes locally.

VTT-AI-Agent (commented):
## Findings

- **[P1] Do not arm Space-pan in the middle of an existing widget drag** (`client/js/zoom.js:115`, `client/js/editor/selection.js:14`): `handleSpaceKeyDown()` enables `spacePanActive` without checking whether a normal edit-mode pointer gesture is already in progress. If a user starts dragging a widget and then presses Space, every subsequent `mousemove` and the matching `mouseup` is returned early by `editInputHandler()`. The normal mouse handler therefore never calls the widget's `moveEnd()` or clears its `mouseStatus`; after `moveStart()` this can leave `dragging` set, a movable widget detached from its parent, and drop-target state uncleared. Please either refuse to arm Space-pan while another pointer gesture is active or allow that gesture's matching mouseup/cleanup through; add a regression covering mousedown/move, then Space keydown, then mouseup.

All current CI checks pass, but the new test only exercises Space keydown before mousedown and does not cover this transition.

The `beta` label is already present, but this revision is not beta-ready because the mid-drag transition can leave shared widget state inconsistent.

VTT-AI-Agent:
Addressed the open mid-drag P1 in commit 615fc6e7.

Space-pan now refuses to arm while any mouse button or widget mouse gesture is already active, so the existing editor gesture keeps ownership of its matching mouseup and completes moveEnd() cleanup. The new TestCafe regression covers mousedown → move → Space keydown → mouseup and verifies that Space never arms, the widget moves normally, and dragging returns to null.

Validation:
- npm test -- --runInBand: 15/15 passed
- npx testcafe chromium:headless tests/testcafe/editor.js: 3/3 passed

VTT-AI-Agent (approved):

VTT-AI-Agent (commented):
## TL;DR

No actionable issues remain in the current revision. The previous mid-drag cleanup risk is fixed by refusing to arm Space-pan while a mouse/widget gesture is active, and the new regression covers that transition.

## Findings

No actionable findings. The Space-modified mousedown is intercepted before the normal edit selection/widget pipeline, release and blur paths clear the temporary pan state, editable controls and overlays are excluded, and the added TestCafe coverage verifies both panning over a movable widget and preserving an already-active widget drag. The changed JavaScript also passes syntax checks, and the PR's current Jest, Chrome/Firefox TestCafe, CodeQL, and server-start checks all pass.

The `beta` label was already present; this PR is beta-eligible because it changes only edit-mode zoom/selection UI, touches none of the defined engine paths, and all current safety checks pass.

PR #2708, opened 2025-10-23
smart holder alignment

ArnoldSmith86:
I want holders to be "smart" by default:
- If it's wide enough for more cards, display more cards.
- If it isn't, make a pile.
- If cards don't fit side-by-side, squish them.
- Same for vertical arrangement.
- If it is wide _and_ tall enough, put them in as many rows as necessary to maximize the visible area of each card.
- If only one card fits, put it in the center.
- If it's not tall enough to fit one card, ~~shrink the cards to fit~~ cut them off.

**After the first commit only horizontal alignment is kinda working with left/center/right and it can do multiple columns. Dragging stuff into multiple columns is a total mess though.**

Eventually `auto`/`auto`/`center`/`center` (or `left`) should just be the default for all holders with a fileupdater making sure that older games don't change. We'll see.

```Javascript
  usesSmartRearrange() {
    return [ 'left', 'center', 'right' ].includes(this.get('dropOffsetX')) ||
           [ 'top', 'center', 'bottom' ].includes(this.get('dropOffsetY')) ||
           this.get('stackOffsetX') == 'auto' || 
           this.get('stackOffsetY') == 'auto';
  }
```

[cinnamon-2025-10-23T204723+0200.webm](https://github.com/user-attachments/assets/8f7ad94a-337d-4af2-a11b-9fc8c70b4bff)

TODO

- [ ] reverse order for `bottom`/`right`
- [x] form, dissolve and arrange piles
- [ ] handle cases where the holder gets too small
- [ ] different widget sizes (in all property combinations)
- [ ] not sure why it won't update when changing properties in the JSON editor
- [ ] combination with some static values (for example `"stackOffsetX": "auto", "dropOffsetX": 4`)
- [ ] test stuff when moving cards with routines
- [ ] JSON editor commands
- [ ] game validator
- [ ] test with owner
- [ ] moving with button moves to the bottom
- [ ] when multiple columns don't fill the height, it should have the same padding to the sides (same for the other dimension actually)
- [ ] different child sizes can lead to one owner with a pile and one owner with arranged cards
- [ ] dropping a pile at a certain position still adds all cards to the back

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2708/pr-test (or any other room on that server)


96LawDawg:
I had a seizure.

https://github.com/user-attachments/assets/061cf5d6-9203-4864-a1db-79e4a352fc62

96LawDawg:
Closed for #3117

PR #2711, opened 2025-10-28
feature-Custom reusable widgets

mousewax:
### Summary
This PR introduces a new feature for creating and managing custom reusable widgets. Users can now save, load, and manage widgets from both local storage and a server-side library.

### Key Features
- **Custom Widgets Sidebar:** A new 'Widgets' panel in the editor sidebar allows users to browse, filter, and manage custom widgets.
- **Local and Server Storage:** Widgets can be saved to either the browser's local storage or a server-side library (`library/widgets.json`) based on the `allowPublicLibraryEdits` value in `config.json`.
- **Widget Management:** Users can create, rename, delete, and organize custom widgets.
- **Drag and Drop:** Widgets can be reordered and added to a room via drag and drop.
- **Automations:** Widgets can include an `addToRoomRoutine` that fires as it is added to a room and then is removed from the final widget.
- **Unique Widgets:** Widgets can be marked as "Unique" meaning they will overwrite any widgets already in the room with the same ID. Non-unique widgets are given unique IDs. 
- **Backups and Sharing:** Users can download a json file containing their custom widgets and re-upload them.
- **API Endpoints:** New API endpoints have been added to manage the server-side widget library.
- **Bug Fixes:** This PR also includes several bug fixes related to widget selection, naming, and editing permissions.
- **New UPLOAD function:** Creates a new function to allow users to upload files.

### File Changes
- **Added:**
  - `client/js/editor/sidebar/widgets.js`: The new sidebar module for managing widgets.
  - `library/widgets.json`: The server-side widget library.
- **Modified:**
  - `server.mjs`: Added API endpoints for widget management.
  - `client/js/editor/layout.js`: Integrated the new `WidgetsModule`.
  - `client/css/editor/sidebarModules.css`: Added styles for the new UI.
  - `package.json`: Updated the debug script.
  - `server/minify.mjs`: Included the new widgets script in the build process.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2711/pr-test (or any other room on that server)

image

To Do

- [x] Fix dark mode
- [x] Button tool tips
- [x] Drag and drop widgets
- [x] Manual drag to re-order widgets in edit
- [x] Widgets go to saved x/y coords when adding from sidebar
- [x] Keep `addToRoomRoutine` for designers but not players?
- [x] ~Add feature to switch to JSON Edit Mode after adding a widget. Holding shift overrides this~
- [x] Fix crash when no localStorage established
[...]

PR #2720, opened 2025-11-04
refactored a surface abstraction layer

ArnoldSmith86:
This PR is not meant to change any functionality. It just rearranges some code to allow multiple places in the UI to show widgets. An example would be the Add Widget overlay that is not part of the actual room but still renders widgets the same way. The current Add Widget overlay and the editor Properties panel for holders do this in a pretty hacky way that breaks for many many edge cases but it was okay because those show basically static content.

#2711 makes it possible to save any group of widgets in a new panel and at this point it is probably a good idea to implement this correctly.

Currently there is a global `widgets` variable that contains all the widgets and their rendered HTML is added to `#topSurface`.

This PR introduces a new `Surface` class where each of its instances has its own `widgets` and render target.

This means that a lot of code has to receive a `surface` pointer so it knows what to operate on.

So the only goal of this PR is to introduce this abstraction layer without changing anything visible to the user.

The state after the first commit is that a random test room managed to render again after many changes to the code, I was able to interact with widgets and the changes were synced to other clients. The editor loads and the Add Widget overlay seems to work but for example selecting widgets in the editor still crashes. A lot of other stuff probably crashes as well. Feel free to add crash examples to a todo list.

---

TODO

- [ ] crash when selecting widgets in editor
- [x] crash when loading a new game
- [ ] `npm test`
- [x] batchDepth not defined

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2720/pr-test (or any other room on that server)

### DIFF
--- a/client/js/serverstate.js
+++ b/client/js/serverstate.js
@@ -521,9 +521,9 @@ function cancelInputOverlay() {
     if($('#buttonInputOverlay').style.display == 'flex')
       showOverlay();
 
-    delta = { s: {} };
-    deltaChanged = false;
-    batchDepth = 0;
+    topSurface.delta = { s: {} };
+    topSurface.deltaChanged = false;
+    topSurface.batchDepth = 0;
   }
 }
 
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -884,7 +884,7 @@ export class Widget extends StateManaged {
       if(Array.isArray(collections[collection]))
         newCollection = collection
       else if (Array.isArray(collection)) {
-        newCollection = '$collection_' + batchDepth;
+        newCollection = '$collection_' + this.surface.batchDepth;
         collections[newCollection] = this.widgetFilter(w=>collection.indexOf(w.id)!=-1);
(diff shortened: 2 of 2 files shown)


rogerl50:
I got
```
widgets is not defined
ReferenceError: widgets is not defined
    at $.onclick (https://test.virtualtabletop.io/PR-2720/pr-test:5841:7)
```
when trying to load my poker room. What's the best way to get you more information to debu?

96LawDawg:
In 4-Letter Words, clicking on the checkmark (validate words button) gives error:
```
Uncaught (in promise) ReferenceError: batchDepth is not defined
    at getCollection (lawdawg:7289:42)
    at Button.evaluateRoutine (lawdawg:8257:33)
    at Button.evaluateRoutine (lawdawg:7864:24)
    at Button.click (lawdawg:6883:18)
    at inputHandler (lawdawg:3746:26)
    at lawdawg:3814:41
    ```

The reset button doesn't work at all.  No debug, no error, nothing.

PR #2721, opened 2025-11-05
feat: Add line thickness control to canvas widget

mousewax:
Adds `lineWidth` parameter to the `canvas` widget.  Adds a new button on the composite widget to adjust brush size.

Fixes that the color change composite button would not update when the color was changed using a CANVAS mode.

Fixes broken "change" mode.

Deprecates use of "canvas" to choose the canvas to act on.

Breaks canvas tutorial into a widget one and a function one.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2721/pr-test (or any other room on that server)

---

### Updated entry for canvas on widgets wiki:

# Canvas

A canvas is a surface upon which players may do free-form drawing. The resolution of the canvas is 100x100 pixels which gets scaled to the size of the widget. It supports 64 colors that can be defined in the property _colorMap_ (only the first 6 are currently defined by default). When drawing it uses the color at index _activeColor_.

[Click to see tutorial](https://virtualtabletop.io/tutorial/tutorial/types-canvas)
  
| activeColor | 1                 | The default color is VirtualTableTop blue (#1F5CA6).  Can be changed with the CANVAS function using modes "set", "inc", or "dec".  |
| artist | null                 | Indicates the `playerName` of the player or players that can use the canvas.  It can accept a single `playerName` (`e.g. "artist": "guest1234"`) or an array of `playerName` (`e.g. "artist": ["guest1234", "guest2222"]`).  The default value `null` allows anyone to use the canvas. To change `playerName` dynamically, use SET.|
| c00 - c99 | null                 | Canvas encodes pixels and colors into properties labeled c00 to c99. These change dynamically as the canvas is used. Can be set with the CANVAS function using mode "setPixel". All values are erased using CANVAS mode "reset".   |
| colorMap | [ ] |  A button cycles through the default color palette in the following order: #1F5CA6, #000000, #FF0000, #008000, #FFFF00, #FFA500, #FFC0CB, #800080, #A52A2A,#F0F0F0. The array values may be also be custom modified as desired. To make the drawing surface transparent, add 'transparent' as the first item in the `colorMap` array. Can be changed with the CANVAS function using mode "change". |
| clickable | true                | This is the default.  Setting it to `false` prevents changes to the canvas drawing, but does not prevent using the Reset or color change buttons.|
| lineWidth | 1             | This is the default.  Must be a positive number. The larger the number, the larger the size of the brush. Also related to resolution as `lineWidth` of 1 on a `resolution` 100 canvas will be larger than on a canvas with `resolution` 500. To change `lineWidth` dynamically, use SET.|
| resolution | 100                 | Controls size of pixels displayed on the canvas. The larger the number, the finer the detail.   |
| typeClasses | "widget canvas"   | Same property as in [the global properties](#global-properties-for-every-widget) but with a different default value. This gives a canvas its default appearance. You should not alter this property or the canvas widget might break. |
| width, height | 400, 400         | Same property as in [the global properties](#global-properties-for-every-widget) but with a different default value. Although the default values are 400x400, the composed widget that is added through the Add Widget button is 800x800.|

----
### [Click to see tutorial](https://virtualtabletop.io/tutorial/functions-canvas)

This function simplifies the control of canvas widgets using routines. the mode parameter is what will determine which kind of automation is going to be executed. The CANVAS function allows you to choose a color to draw with, change colors in the color map and draw using automations.

Parameters:
* canvas (DEPRECATED): _canvas id_ (or an array)- specifies the target canvas (defaults to null). This function is DEPRECATED. It is still available but use `collection` instead.
* collection: _collection_ - specifies the collection of widgets to make changes to (defaults to DEFAULT collection)
* color: _string_ - ONLY used in mode "change" as the input of the new color (defaults to VTT blue)
* mode: _set/ inc/ dec/ change/ reset/ setPixel_ - specifies the mode to be applied to the canvas (the other parameters may be ignored depending on the mode)
    * set: uses the "value" parameter to choose the active color of the `colorMap` with which the player can draw. Note that 0 is the first color in the `colorMap` used as the background color (it will loop in the `colorMap` if necessary). The "color" parameter is not used in this mode. Updates the `activeColor` property on the canvas.
    * inc / dec: uses the "value" parameter to choose the active color of the `colorMap` with which the player can draw (it will loop in the color map if necessary). Updates the `activeColor` property on the canvas.
    * change: uses the "color" parameter as the new value for the `colorMap` that corresponds to the index of the "value" parameter (it will loop in the `colorMap` if necessary). It does not update the `activeColor` property on the canvas however using this mode will immediately replace all of the old color pixels with the new color.
    * reset: set all pixels to the first color of the color map erasing the canvas.
    * setPixel: Add one pixel in position "x", "y" with the color corresponding to the `colorMap` index of the "value" parameter. The "color" parameter is not used in this mode. 
* value: _number_ - only used as a reference to the index of a color in the `colorMap` (defaults to 1) 
* x and y: _number_ - only used in mode "setPixel" to locate the place to draw pixel (defaults to 0). Must be positive number on the canvas based on the `resolution`. The max value of x and y is the same as `resolution`.

Mode `change` example:
```JS
 {   
      "func": "CANVAS",
      "collection": ["k8ek"],
      "mode": "change",
      "color": "silver",
      "value": 3
 }
```

Mode `setPixel` example:

```JS
 {
      "func": "CANVAS",
      "collection": ["k8ek"],
      "mode": "setPixel",
      "x": 50,
      "y": 50,
      "value": 1
  }
```

### DIFF
--- a/client/js/widgets/canvas.js
+++ b/client/js/widgets/canvas.js
@@ -4,6 +4,7 @@ class Canvas extends Widget {
     this.canvas = document.createElement('canvas');
     this.canvas.dataset.id = id;
     this.context = this.canvas.getContext('2d');
+    this.regionCache = {};
 
     const defaults = {
       width: 400,
@@ -139,6 +140,16 @@ class Canvas extends Widget {
     return code;
   }
 
+  async flushPixelCache() {
+    if(this.regionCache) {
+      batchStart();
+      for(const key in this.regionCache)
+        await this.set(key, this.compress(this.regionCache[key]));
+      this.regionCache = {};
+      batchEnd();
+    }
+  }
+
   getColorMap() {
     if(Array.isArray(this.get('colorMap')))
       return this.get('colorMap');
@@ -170,12 +181,12 @@ class Canvas extends Widget {
     if(this.lastPixelX !== undefined && state != 'down') {
       const steps = Math.max(Math.abs(pixelX-this.lastPixelX), Math.abs(pixelY-this.lastPixelY))*2;
       for(let i=0; i<steps; ++i)
-        this.setPixel(this.lastPixelX + (pixelX-this.lastPixelX)/steps*i, this.lastPixelY + (pixelY-this.lastPixelY)/steps*i, undefined, undefined, false);
+        this.setPixel(this.lastPixelX + (pixelX-this.lastPixelX)/steps*i, this.lastPixelY + (pixelY-this.lastPixelY)/steps*i, this.get('activeColor'), regionRes, false);
     } else {
-      this.setPixel(pixelX, pixelY, undefined, undefined, false);
+      this.setPixel(pixelX, pixelY, this.get('activeColor'), regionRes, false);
     }
 
-    if (state == 'down') {
+    if(state == 'down') {
       const resolution = this.getResolution();
       const minInterval = 10;
       const maxInterval = 100;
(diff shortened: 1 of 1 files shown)


96LawDawg (approved):
Tested. Approved. New tutorials are done. Minor tweaks to wiki needed when merged.

PR #2728, opened 2025-11-10
added command search to editor

ArnoldSmith86:
Meet your new best friend: `Tab`

While holding Tab in the JSON editor, you can filter the command pane:
- only show matching buttons, select with arrows, release to execute
- search for widget IDs to open them
- search for var expression operations to add them
- search for editor contents to jump to the line

This PR is currently 99% AI. I will have to do a code review but it seems to be working pretty great and I would like feedback first.

TODO:

- [x] review code
- [x] light mode text is not readable
- [x] scrolling highlighted command into view jumps to the top briefly
- [x] jumping to a line doesn't update the command pane
- [x] ~~search across content of _all_ widgets?~~
- [ ] cancel with ESC
- [x] show explanation when starting to hold Tab
- [x] prevent it from working if cursor not in JSON (try it with hotkey "a" and you get both add widget screen and tab search result)

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2728/pr-test (or any other room on that server)

### DIFF
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -2948,7 +2948,7 @@ function jeShowCommands() {
   }
   commandText += `</div>`;
   if (!jeTabSearchActive) {
-    commandText += `<div style="margin-bottom: 8px; font-size: 12px; color: var(--textDimColor1);">Hold <span class="key">Tab</span> to search</div>`;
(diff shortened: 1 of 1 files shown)


ArnoldSmith86:
Alright, the code is mostly fine now. And it's only editor code which is easy to fix when something breaks. I think the open todos aren't very important and could come later. I tried to implement cancelling with ESC but it's slightly more complicated than I expected..

PR #2729, opened 2025-11-11
feat: Update card widget

mousewax:
Simply updates the card widget to use inline HTML like the basic widget.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2729/pr-test (or any other room on that server)

To Do

- [x] ~Fix editor code that compresses images in card html~ (opened a new issue)
- [x] Fix validator message: contains '$' - use dynamicProperties to map widget property names to face object properties instead
- [x] ~Fix lag caused when editing html that triggers redraw of all cards~ (opened a new issue)

### DIFF
--- a/library/games/Wildscapes/0.json
+++ b/library/games/Wildscapes/0.json
@@ -2127,15 +2127,6 @@
     ],
     "parent": "startTileHolder"
   },
-  "yf3": {
-    "type": "card",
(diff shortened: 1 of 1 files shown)


96LawDawg (approved):
Approved through test

PR #2736, opened 2025-11-15
Add inline HTML syntax highlighting in JSON editor

ArnoldSmith86:
[cinnamon-2025-11-16T081636+0100.webm](https://github.com/user-attachments/assets/52652327-a9aa-4eac-9d96-b85878270662)

- support multi-line strings in the JSON editor (translated from and to `\n` in the actual JSON)
- syntax highlighting for HTML
- command for formatting existing HTML into multiple rows with indentation that fits into the JSON

All of this isn't perfect but I don't think it's worth it to handle every edge case ever.

The added whitespace in the HTML causes rendering to change quite a bit because cards (and probably basic widgets) inherit the `white-space: pre-wrap` from `.widget`. We could update the CSS for cards to better handle this because I don't think anyone has "legacy" games with the new non-iframe cards yet. But simply changing it for basic widgets for consistency should be a legacy mode. But we can also just leave it to the designer to apply whatever CSS they want.

Even adding `white-space: collapse` is not _really_ a solution though because `collapse` only means that multiple spaces render as a single space. But the formatting probably added multiple spaces where there were _no_ spaces before.

Thoughts?

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2736/pr-test (or any other room on that server)

TO DO:

- [x] Color `${PROPERTY ...}` teal wherever it is in html
- [x] Keep all colors the same before and after using "format html" button
- [x] Cursor should not move back to top of JSON after using "format html" button
- [x] Put all tags other than style on their own line
- [x] The closing `",` on one line breaks the context detection for lines below it.

### DIFF
--- a/library/tutorials/Properties - HTML/0.json
+++ b/library/tutorials/Properties - HTML/0.json
@@ -13,7 +13,7 @@
       "mode": "Tutorial",
       "time": "0",
       "attribution": "",
-      "lastUpdate": 1765415439809,
+      "lastUpdate": 1765415953763,
       "showName": false,
       "skill": "",
       "description": "",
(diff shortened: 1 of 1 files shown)


96LawDawg:
Would be nice if the `${PROPERTY}` stuff still showed up as teal or whatever in this context instead of just white:
Screenshot 2025-11-16 084743

PR #2738, opened 2025-11-16
made html cards and basic widgets use white-space:collapse

ArnoldSmith86:
With #2736 it is now very easy to make HTML inside the JSON reable by splitting it into multiple lines. The current default `white-space: pre-wrap` means that those newlines are actually rendered on the widgets though which is probably not often what you want.

This PR changes the default to `white-space: collapse` and has a file updater that adds `white-space: pre-wrap` to the widget `css` if it uses HTML, does not already have `white-space` defined in its CSS and if it contains multiple spaces in a row or newlines.

This is not perfect because HTML values support embedding other property values which in turn might have newlines or multiple spaces. But I do not want to overdo it even further. Especially because cards using the legacy iframe mode are not affected by this change at all.

- [ ] File updater might have to not do anything for legacy mode html cards.
- [x] Fix nested `css` handling in file updater.
- [ ] What if only `white-space-collapse` or `text-wrap-mode` is defined?

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2738/pr-test (or any other room on that server)

### DIFF
--- a/client/css/widgets/basicwidget.css
+++ b/client/css/widgets/basicwidget.css
@@ -4,6 +4,6 @@
   pointer-events: none;
 }
 
-.basic.widget.html {
+.basic.widget.usesHTML {
   white-space: collapse;
 }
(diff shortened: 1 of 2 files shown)


ArnoldSmith86:
[testcase.vtt.zip](https://github.com/user-attachments/files/23660898/testcase.vtt.zip)

First tests show that the file updater doesn't handle nested `css` correctly.

PR #2746, opened 2025-11-20
feature-Prettier Guest Names

mousewax:
Creates new guest names that are board game and tabletop-related adjective and noun. There is also a zero-width space at the beginning to key off of similar to "Guest".

Additionally, to make sure the zero-width space is removed, when an input gets focus that has that at the beginning the value is completely blanked and restored if the value is empty. 

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2746/pr-test (or any other room on that server)

### DIFF
--- a/server/room.mjs
+++ b/server/room.mjs
@@ -785,7 +785,7 @@ export default class Room {
     Logging.log(`removing player ${player.name} from room ${this.id}`);
 
     this.players = this.players.filter(e => e != player);
-    if(player.name.match(/^Guest/) && !this.players.filter(e => e.name == player.name).length)
+    if(player.name.includes('\u200B') && !this.players.filter(e => e.name == player.name).length)
       if(!Object.values(this.state).filter(w=>w.player==player.name||w.owner==player.name||Array.isArray(w.owner)&&w.owner.indexOf(player.name)!=-1).length)
         delete this.state._meta.players[player.name];
 


server/fileupdater.mjs
```diff
@@ -526,3 +527,34 @@ function v19useIframeForHtmlCards(meta, state) {
...
+          if (match) {
+            const varName = match;
+            if (operation.operand1 === `\${${varName}}`) {
+              routine[i - 1] = prevOp.replace('0 5', '0 1');
+              operation.operand2 = '\u200B';
+            }
+          }
+        }
+      }
+    }
```
ArnoldSmith86: This seems to handle only a very specific case.

client/js/overlays/players.js
```diff
@@ -120,7 +163,8 @@ onLoad(function() {
   });
   onMessage('rename', function(args) {
     const oldName = playerName;
-    playerName = args;
+    if(args)
+      playerName = args;
```
ArnoldSmith86: If the name on the server is set to `""`, the client should not simply ignore that and keep the previous name. I don't mind enforcing a non-empty name but that shouldn't be here.

client/css/overlays/players.css
```diff
@@ -44,32 +45,38 @@
...
+.teamColor::before {
+    content: '';
+    display: block;
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    -webkit-mask-image: url(i/icons-white/user-solid-svg.svg);
+    mask-image: url(i/icons-white/user-solid-svg.svg);
```
ArnoldSmith86: This does not show up for me in Firefox 144 / Linux.

PR #2755, opened 2025-11-30
Update Foundations Tutorial 121

96LawDawg:
Recently, ArnoldSmith86 discovered that the scale property can be used to invert a widget horizontally or vertically like `"scale": "-1,1"` or "1,-1". The scale variant of the tutorial is updated with this information.

EDIT: I also updated the Properties series Icon tutorial, CSS variant with examples of how to use the css selectors to invert icons.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2755/pr-test (or any other room on that server)

### DIFF
--- a/library/tutorials/121 Fundamental Widget Properties/1.json
+++ b/library/tutorials/121 Fundamental Widget Properties/1.json
@@ -1269,7 +1269,7 @@
     "css": {
       "font-size": "22px"
     },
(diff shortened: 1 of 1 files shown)


96LawDawg:
Should we also update the validator in validate_gamefile.js line 57 to change scale from "number" to "any"?  I don't think we should do that since number is by far going to be the most common use case.  But I just wanted to raise the question.

ArnoldSmith86 (approved):
It could mention that children are also flipped but I don't think it's essential.

PR #2765, opened 2025-12-10
Public Library - Add Trajectory

96LawDawg:
Similar to Keltis and Lost Cities.  VTT implementation by LawDawg.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2765/pr-test (or any other room on that server)


ArnoldSmith86 (approved):
As mentioned in Discord: no year, no awards (even though it is Spiel des Jahres 2008)

PR #2767, opened 2025-12-10
removed "add card widget" JSON editor button

ArnoldSmith86:
It was auto-generated for all widget types but adding a card would crash the client because it needs a deck to be valid.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2767/pr-test (or any other room on that server)


96LawDawg (commented):
Approved through test and code review

PR #2812, opened 2026-01-09
Public Library - Add Vineyard

96LawDawg:
Similar to Farm Hand. VTT implementation by Robin Harn.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2812/pr-test (or any other room on that server)


ArnoldSmith86 (changes requested):
The background needs to be compressed.

PR #2814, opened 2026-01-12
Fix tutorial validation problems - Types and Xtras series

96LawDawg:
Similar to #2739. Mostly deck unused card problems and a few other minor things. However, there are 2 possible issues. The default collection TURN is not recognized from a TURN operation and the deprecated skip parameter is not recognized. We probably should not fix skip. We probably should update the validator for TURN collections.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2814/pr-test (or any other room on that server)


ArnoldSmith86:
But do we want tutorials to use `skip`?

96LawDawg:
> But do we want tutorials to use `skip`?

No. But I think I'm too lazy to change it. It was late. Now it is early. Still too lazy. But I will if you think we should. I just have to remember what skip does. I was mainly thinking about the 27 PL variants that still use it.

Also we need to change the TestCafe hashes anyway, so this isn't completely done.

library/tutorials/Types - Piles/1.json
```diff
@@ -1,6 +1,12 @@
 {
   "_meta": {
-    "version": 12,
+    "version": 20,
+    "gameSettings": {
+      "legacyModes": {
+        "convertNumericVarParametersToNumbers": true,
+        "useOneAsDefaultForVarParameters": true
```
ArnoldSmith86: These could be removed as well, right?

library/tutorials/Types - Scoreboard/1.json
```diff
@@ -115,7 +118,17 @@
...
     "text": "The appearance of scoreboards can be highly customized using css. Unlike with many other capabilities in the JSON Editor, customizing css requires a lot of manual entry as there are not buttons to automate the placement of the properties. You also need to understand a little bit about the VTT code for the scoreboard widget and html table structure in order to maximize the customization potential. This room contains some samples that demonstrate some of the features most likely to be implemented.  Look at the JSON (ctrl-J) of each scoreboard to see the css.",
     "movable": false,
-    "movableInEdit": false
+    "movableInEdit": false,
+    "clickRoutine": [
+      "// This is only here to remove validation errors",
+      {
+        "func": "VAR",
+        "variables": {
+          "dummy1": "${PROPERTY score}",
```
ArnoldSmith86: Same question.

library/tutorials/Xtras - State Dependent Routines/0.json
```diff
@@ -124,7 +124,6 @@
       "text4"
     ],
     "state": 3,
-    "debug": true,
```
ArnoldSmith86: Damn. Haven't seen this in a while... 😁

library/tutorials/Types - Scoreboard/0.json
```diff
@@ -159,7 +162,17 @@
...
     "text": "Scoreboard uses a property in seats to display numerical information in table form and total those numbers. Working in conjunction with routines and the SCORE function (shown in a different tutorial), the property on the seat can be updated and cause real-time changes to the scoreboard data. This room shows most of the options available when using scoreboards. The seats along the left side are pre-filled with player and score information to populate the data on the scoreboards in this room. Use the JSON Editor (ctrl-J) to look in the seats and scoreboards to see how missing or non-numeric data is handled.",
     "movable": false,
-    "movableInEdit": false
+    "movableInEdit": false,
+    "clickRoutine": [
+      "// This is only here to remove validation errors",
+      {
+        "func": "VAR",
+        "variables": {
+          "dummy1": "${PROPERTY score}",
```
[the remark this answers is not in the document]
ArnoldSmith86: Yeah, `score` is complicated. It's more a scoreboard property than a seat property but I thought I made the validator recognize it if it is actually being used by a scoreboard.

If I'll remember, I'll check that again..

PR #2823, opened 2026-01-17
fixed crash when removing limbo widget

ArnoldSmith86:
When removing a widget with an invalid parent, it crashes because the removal tries to act on the parent. This should fix it by skipping that step when the widget is in limbo.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2823/pr-test (or any other room on that server)


96LawDawg (approved):
Approved through test

PR #2824, opened 2026-01-18
Add smarter drag options

96LawDawg:
Adds a new settings (gear icon) button to the drag toolbar in edit mode. Clicking on the settings allows you to change the way the moving and cloning take place. Defaults to current behavior.  Optionally, you can move and clone based on the grid of the first selected widget. You can also enable a basic inheritFrom or a inheritFrom the true master widget.

As an example of that last feature, if you copy widget1 then you get something like widget2 that uses inheritFrom "widget1."  Under the default mode or even the standard inheritFrom mode, if you copy widget2 then the new widget3 uses inheritFrom of "widget2." With this new mode if you copy widget2 then the new widget3 uses inheritFrom of whatever was in widget2, so it would be "widget1." It looks back up the tree recursively to find the master widget.

The code is all AI slop. I haven't looked at it to try and make it better or remove things that don't need to be there. I will, but I wanted to go ahead and get this out there because it does work and can be used now.

-----------------------
Update: Added an aspect ratio lock and unlock button. Works for both resizing and adding spacing between widgets. Defaults to unlocked.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2824/pr-test (or any other room on that server)


ArnoldSmith86:
Nice! The "keep aspect ratio" could then also become an option.

Selecting a hex grid and moving it doesn't seem to snap to the grid though.

96LawDawg:
`inheritFrom` cloning was broken in this PR and in production. Things like hand, linkedToSeat, onlyVisibleToSeat, dropTarget, text, etc were not getting added back to cloned widgets. I just did a commit that fixes that.

ArnoldSmith86 (approved):
Just merge it. It's just the editor.

PR #2826, opened 2026-01-20
Public Library - Add Streetcar Suburb

96LawDawg:
An original game by Drew Edwards. VTT implementation by LawDawg.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2826/pr-test (or any other room on that server)


ArnoldSmith86 (approved):
Looks good to me.

Like I said on Discord: it might be better to drag lines onto the board but if the original works with drawing (or you prefer it this way), sure.

Not sure if the lines should have the seat colors. I took over a seat, it told me the color was recently used so I used a different one and then the drawing color didn't match the seat color. 🤷‍♂️ 

Image

PR #2846, opened 2026-01-25
added AI badge to Game Shelf

ArnoldSmith86:
~~This is 100% AI generated so far.~~

This adds a new checkbox to the game metadata to flag games that use AI generated imagery. If checked, the game will receive an "AI" badge in the Game Shelf and a little note in the details.

- [x] The scripts/ directory needs to be removed before merging.
- [x] Several restored library images need to be compressed (I have a script for that).
- [x] I feel like there are still quite a few games that would just need restoring of the old library image.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2846/pr-test (or any other room on that server)

### DIFF
--- a/client/css/overlays/misc.css
+++ b/client/css/overlays/misc.css
@@ -40,3 +40,18 @@
       border-radius: 10px;
     }
   }
+
+.ai-imagery-notice {
+  margin-top: 12px;
+  padding: 8px;
+  background: rgba(33, 150, 243, 0.1);
+  border-radius: 4px;
+  border-left: 3px solid var(--VTTblue);
+}
+.ai-imagery-notice .material-symbols {
+  vertical-align: middle;
+  color: var(--VTTblue);
+}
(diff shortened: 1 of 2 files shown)


RaphaelAlvez:
Can you make it so blue keeps the photo? It's not AI.

Seeing this male me wonder if we should have a feature of preview in the game shelf where we  have 1 (maybe more) images that show elements of the implementation. Maybe not the full table as just having some elements may be more informative.

For example for blue I'd have the photo as the cover and after 1-2 seg hovering the image it would change to the in game tiles images

ArnoldSmith86:
> Can you make it so blue keeps the photo? It's not AI.

No idea why that was changed then. I reverted it.

I didn't generate a new HTML file.

I also added a filter to the Game Shelf and badge+note also show up for game share links.

---

Would be nice if @96LawDawg could go through the library changes once. Other than that, this should be good to go.

PR #2851, opened 2026-01-29
fixed validator for `CALL.arguments`

ArnoldSmith86:
It was using the values as valid variables instead of the keys.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2851/pr-test (or any other room on that server)


96LawDawg:
Test case?  I can't figure out how to test the difference between production and this PR.

ArnoldSmith86:
> Test case? I can't figure out how to test the difference between production and this PR.

The test case you have in the room works just fine for me. Error in production, no error in here.

96LawDawg (approved):
Approved through test.

PR #2852, opened 2026-02-02
added `rightClickRoutine`, `contextMenu` and `rotationSteps`

### LINKED ISSUE #2762: New Routine: rightClickRoutine
I imagine this would work like `clickRoutine` and `doubleClickRoutine`, except for right clicks.

I'm requesting this because, while working on a game, i wanted to add both double click and right click functionality to a pawn, but i couldn't seem to find any obvious way to add in anything for right click. If i'm wrong, and there is a way to do this that i overlooked, please let me know

ArnoldSmith86:
Fixes #2762. Replaces #2527. The idea of this PR is to unify rotation, enlargement and custom context menus into one popup that can be triggered by right-clicking/long-touching.

- If `rightClickRoutine` is defined, it will just be called. No popup will be shown but you can do that yourself (see below).
- If `contextMenuOptions` is defined (object notation only), it will show an enlarged preview of the widget centered on the actual position of it unless it is too close to the edge. Options: `factor`, `title`, `color` (popup background), `image` and `widget` (show something other than the widget; arrays add prev/next navigation and expose `previewIndex` to button routines). The plain `enlarge` property is left completely untouched and keeps its classic hover/right-click behavior — no legacy mode or migration needed. If a widget has both, the popup preview defaults to the numeric `enlarge` factor and the hover copy gives way to the popup the moment it opens.
- If `rotationSteps` is given (a number like `45` or an array of allowed rotations), it will show two rotation buttons above the menu buttons that rotate to the next allowed angle in that direction - a widget sitting between two entries of the list goes to the nearest entry in the chosen direction, and a number is a relative step.
- If `contextMenu` is given, it will show a menu with custom buttons next to the widget. Each button can have a custom icon, custom text, a custom background `color`, an optional `description` (shows an ℹ️ with a popover) and an optional `menu` for submenus; it calls the given routine when clicked.
- A new `CONTEXTMENU` routine operation can be used to trigger this popup from a routine. It will not wait for user input. You can use a different menu each time (inline via `contextMenu` or from a property via `property`, plus `factor`/`title`/`color`/`image`/`widget` overrides). So you can build entire submenu structures if you want to. Without a `collection` it opens on the widget running the routine, so `rightClickRoutine: [{ "func": "CONTEXTMENU", "contextMenu": [...] }]` just works; name a collection (e.g. `DEFAULT` after a `SELECT`) to open it on another widget.
- Holding right-click/long-touch and then moving to different widgets shows the popup for those (PCIO behavior). On touch, the popup only opens for a finger that holds still: one that moves on within the delay is dragging the widget, and no popup gets in its way. Releasing closes the popup if it has no buttons; otherwise it stays open until you click/tap outside. While open, the popup has priority over the widgets it covers. It closes when its widget is removed, when the widget is hidden from the player (handed to another `owner`, seat visibility, `display` - on the widget or on a parent) or when the room is reloaded, and its buttons never act on a widget the player can't see; the preview repaints when the widget changes while the popup is open. A card with the classic `enlarge` does not stop a right-click from reaching a holder with a popup below it, and a long touch is never a click on release (as before).
- **PCIO importer:** PlayingCards.io's *Rotation Snapping* (`snapAngles`) becomes `rotationSteps` in the deck's `cardDefaults` instead of a warning that it has no equivalent, and its right-click *enlarge* becomes `contextMenuOptions: { factor: 3 }` - the popup behaves like PCIO's hold-to-enlarge, while VTT's `enlarge` shows on hover.

image

**Editing it in the properties sidebar** (Edit Widgets module): *Interaction & display* has a **Rotation steps** field below *Enlarge* — a number like `45` or a list like `0, 90, 180, 270`, stored as a number or an array. A new collapsible **Context menu** section after it edits the popup: a *Popup on right-click* switch for a preview-only popup (locked while entries or rotation steps open it anyway), the **menu entries** as expandable rows with text, icon picker, a dropdown of the widget's routines, color, description, nested submenus and reordering, and the **Preview** settings of `contextMenuOptions` (size factor, title, background color, an image list, and a widget list with the in-room picker). The generic property list no longer repeats the three properties. `rightClickRoutine` and `CONTEXTMENU` are in the Automations/routine editor; the `CONTEXTMENU` sentence reads "Show a popup on this widget" until a collection is picked. A card or an inheriting widget whose entries come from the deck's `cardDefaults` or from `inheritFrom` shows them as rows with a note saying where they come from. Screenshots: [Rotation steps](https://agent.virtualtabletop.io/reports/pr/1525148990264442880/ui/01-interaction.png) · [empty section](https://agent.virtualtabletop.io/reports/pr/1525148990264442880/ui/02-section-empty.png) · [an entry](https://agent.virtualtabletop.io/reports/pr/1525148990264442880/ui/03-entry.png) · [submenu](https://agent.virtualtabletop.io/reports/pr/1525148990264442880/ui/04-submenu.png) · [preview widgets](https://agent.virtualtabletop.io/reports/pr/1525148990264442880/ui/05-preview-widgets.png) · [preview images](https://agent.virtualtabletop.io/reports/pr/1525148990264442880/ui/06-preview-image.png) · [the resulting popup](https://agent.virtualtabletop.io/reports/pr/1525148990264442880/ui/07-popup.png).

**Demo room (a tutorial, not committed):** *Properties - Context Menu* is on this PR's test server in the [pr-test-ai room](https://test.virtualtabletop.io/PR-2852/pr-test-ai) ([.vtt](https://agent.virtualtabletop.io/reports/pr/1525148990264442880/Properties%20-%20Context%20Menu.vtt)), with three variants: 0 - `rotationSteps` and `contextMenu` on a row of tiles; 1 - `contextMenuOptions` (factor, title, color, image and widget lists with prev/next, hold-and-move over a fan of cards); 2 - `rightClickRoutine` and the `CONTEXTMENU` operation (inline menus, menus from a property, overrides, `previewIndex`, menu trees). It is written to tutorial quality but stays out of `library/`; it can be added to `library/tutorials/` on request.

---

Implementation notes

- [x] Test on mobile. *(verified with emulated touch: long-touch opens, move-to-switch works, buttons tappable, tap outside closes, a touch drag never opens the popup - reported on iOS by 96LawDawg)*
- [x] Cleanup code.
- Deferred follow-up idea: show interaction hints while holding right-click. This is not required for the context-menu behavior and would need its own UX design.
- [x] Make sure the legacy mode for on-hover-enlarge works and make sure it's what we want. *(resolved differently: `enlarge` is untouched, so no legacy mode is needed at all)*
- [x] Maybe add an easy way to show something other than the widget.
- [x] Popup is over overlays.
- [x] Maybe hide the popup when releasing the click if there are no buttons on it? Maybe even if the cursor is not on top of the popup?
- [x] Add `hint` or `description` to each button that would create an `i` next to it with a tooltip?
- [x] Allow `rotationSteps` to be a number? So just `45` would allow all orientations in 45° steps.
- [x] Popup needs priority over other right clickable widgets if they overlap.
- Intentional non-goal: the popup stays out of Edit Mode, where classic right-click behavior is retained for selection and JSON-editor workflows.
- [x] Make it not break on any viewport size. *(preview factor is clamped to the viewport, the popup clamps its position and scrolls if needed)*
- [x] Right-clicking in Edit Mode is broken in general. *(this branch had suppressed right-button handling everywhere; classic edit-mode right-click behavior and JSON-editor widget selection are restored — the popup stays out of edit mode and JSON-editor sessions)*
- Not applicable: faces overriding `rotation` already prevent direct rotation independently of this popup.
- Independent existing issue: the edit-mode blue selection box changes size while rotating a meeple; this is unrelated to the play-mode popup.
- [x] `previewIndex` should not be a SUPER_GLOBAL in the validator. *(now only in scope for routines referenced from `contextMenu` entries)*
- [x] Leave `enlarge` alone and introduce `contextMenuOptions` instead.
- [x] Doesn't trigger in holders/piles. *(right-clicking a card now falls through to the enclosing holder/pile if the card itself has no popup triggers - a classic `enlarge` on the card included)*
- [x] The popup must not outlive its widget. *(removing the widget or loading another game closes it; its buttons check that the widget is still in the room)*
- [x] The popup must not outlive the player's access to its widget. *(an owner, seat or display change of the widget or of a parent closes it; actions and rotations check this again before running)*
- [x] A rotation whose change routine removes the widget, or a `DELAY` during which the popup is closed, must not throw. *(the rotation handler only continues if the popup still belongs to that widget)*
- [x] An array of `rotationSteps` turns to the next allowed angle in the chosen direction instead of snapping to the nearest step first.
- [x] The validator accepts menu routines that come from `inheritFrom` or a card deck's `cardDefaults`.
- [x] The PCIO importer uses `rotationSteps` and `contextMenuOptions` for *Rotation Snapping* and *enlarge*.
- [x] A widget with both `enlarge` and a popup that was selected in edit mode kept its enlarged hover copy on screen next to the popup until a reload. *(reported by 96LawDawg; the selection survives leaving edit mode, and the classic enlarge code kept the copy of any selected widget - that now only applies while editing, the popup hides the copy when it opens, and leaving edit mode hides a pinned copy)*
- [x] Engine tests: `tests/testcafe/contextmenu.js` drives rotation steps from an angle between two steps, a menu entry, a submenu with `previewIndex`, `CONTEXTMENU` on its own widget, a rotation that deletes its widget, an owner change while the popup is open and the enlarged hover copy giving way to the popup against the state hash; jest covers the removal case, the deleting rotation routine, off-step and signed rotation lists, owner and parent hides, actions on hidden widgets, the long-touch class and the `CONTEXTMENU` default.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2852/pr-test (or any other room on that server)


96LawDawg:
In the css for an inactive button (.contextMenuAction:disabled), opacity is not a good solution for when the button has a color. I think `filter: grayscale(100%) contrast(0.5);` works well. Just making a note to change this later as I can't do it from work.

96LawDawg:
Would it be a good idea to support the legacy `enlarge` alongside a `rightClickEnlarge`? I love the new functionality with this, but probably there are times when I would just like to hover rather than click.

VTT-AI-Agent:
I picked this PR up, merged current `main` into it and addressed everything that was still open. Summary of the changes:

**Design change agreed in the discussion — `enlarge` stays untouched:**
- The plain `enlarge` property behaves exactly as on `main` again (hover/right-click enlarge). The `hoverEnlarge` legacy mode, its FileUpdater migration and the game-settings checkbox are gone — nothing about existing games changes, so no migration is needed. This was also what broke the TestCafe CI runs (the migration rewrote `_meta` of every game that uses `enlarge`, changing all state hashes).
- New `contextMenuOptions` property (object only) opts a widget into the popup and configures the preview: `factor`, `title`, `color`, `image`, `widget`. If a widget also has a numeric `enlarge`, the preview defaults to that factor. The validator checks it.

**Fixes from the discussion:**
- Disabled menu buttons now use `filter: grayscale(100%) contrast(0.5)` instead of `opacity` (@96LawDawg — per-button background is the existing `color` key on a menu entry, and the popup background is `contextMenuOptions: {"color": ...}` now that `enlarge` is off-limits).
- Popup priority: while the popup is open, right-clicking it no longer closes it and never triggers widgets underneath it; hovering other widgets can't steal focus from its buttons either. Hold-and-move switching between widgets still works.
- Works in holders/piles: right-clicking/long-touching a card without its own popup triggers now falls through to the enclosing holder/pile.
- Mobile: long-touch opens the popup, moving the finger switches widgets, buttons are tappable and tapping outside closes it (no `click` event is synthesized there, so this needed its own touch handling). Verified with emulated touch (Pixel 5 viewport).
- Viewport safety: the preview scale is clamped to the viewport and the popup clamps its position and scrolls if it is still too large.
- Validator: `previewIndex` is no longer a SUPER_GLOBAL; it is only in scope for routines referenced from `contextMenu` entries, which are now also actually validated.
- The "Production Environment" CI failure was uglify-es choking on `??`/`?.` in the new code — those are gone and the server starts with `minifyJavascript: true` locally.
- Cleanup: dead variables removed, duplicated hold-handling factored out, unused imports dropped.

**Layout change from Discord:**

> Move the rotation buttons above the context menu.
>
> (Source: https://discord.com/channels/770758631146782780/1525148990264442880/1525150337713832027)

Done — the rotation buttons now sit in the right column above the menu buttons instead of above the preview.

**Still open** (left unchecked in the description): popup in edit mode (edit mode keeps the classic behavior for now), the general "right-clicking in edit mode is broken" issue and the blue-selection-box resize inconsistency — both look independent of this PR — and the optional "show an icon on right-clickable widgets" idea.

Local verification: jest passes, all TestCafe files pass locally, and I exercised the popup end-to-end (desktop + emulated mobile) with Playwright.

VTT-AI-Agent:
Follow-up: the first CI round after the big update caught a real regression — the branch was suppressing right-button handling everywhere, which broke the classic right-click behavior in edit mode (selection) and ctrl+right-click widget selection with the JSON editor open. Commit 8bbadca restores those: right-clicks now only get redirected to the popup in normal play mode. All checks are green now.

VTT-AI-Agent (commented):
## TL;DR

Solid feature with careful attention to the tricky input-handling edges (holder fallthrough, edit-mode restoration, mobile touch), and CI is fully green. The most important findings: `enlargePreviewIndex` is never reset when the popup opens, so button routines can receive a stale/out-of-range `previewIndex`; and the JSON-editor/overlay exclusions are incomplete — the popup can still be opened during a JSON-editor session via hold-and-move or long-touch, and while an overlay is open.

## Findings

### 1. `enlargePreviewIndex` is never reset — routines can get an out-of-range `previewIndex`

`client/js/contextmenu.js:14` — the index is module state that neither `openContextMenu()` nor `closeContextMenu()` resets. Concrete failure: open widget A's popup (6 images), navigate to index 5, close it, then open widget B's popup (2 images). The preview renders `5 % 2 = 1` (lines 104/139 apply the modulo), but a button routine receives the raw value `previewIndex: 5` (`client/js/contextmenu.js:369`), which is out of range for B's list and inconsistent with what the player sees. Resetting the index to 0 in `openContextMenu()` (and/or passing the modulo value to the routine) fixes both the stale starting position and the mismatch.

### 2. JSON-editor exclusion is incomplete

`handleContextMenu` correctly bails out when `body.jsonEdit` is set (`client/js/contextmenu.js:542`), and `mousehandling.js` checks `jeEnabled` — but two other entry points only check the `edit` class:

- `client/js/contextmenu.js:635` (document `mousedown`): with the JSON editor open (no edit mode), right-clicking empty table space calls `startRightClickHold()`, and dragging over a widget with popup triggers opens the popup via `updateHoveredWidget()`.
- `client/js/widgets/widget.js:126` (`onlongtouch`) and the document `touchstart` handler at `client/js/contextmenu.js:640`: a long-touch during a JSON-editor session goes straight to `onLongTouch()` and opens the popup.

Both contradict the stated intent that the popup stays out of JSON-editor sessions. Checking `jsonEdit` alongside `edit` in those three places closes the hole.

### 3. Document-level handlers ignore `overlayActive`/`isLoading`

`inputHandler` in `mousehandling.js` returns early on `isLoading || overlayActive`, but the new document `mousedown`/`touchstart` handlers in `client/js/contextmenu.js:634-665` have no such guard. With a full-screen overlay open (game selection, settings, …), a right-mousedown anywhere is `preventDefault()`ed and starts a hold; `widgetAtPoint()` uses `elementsFromPoint`, which returns widgets *underneath* the overlay, so hold-and-move can open the popup for — and rotate or run routines on — widgets the player can't currently see. Guarding these handlers the same way `inputHandler` does seems right.

### 4. Rotation buttons don't normalize `rotation`, unlike `ROTATE`

`Widget.rotate()` stores `(rotation + degrees) % 360` (`client/js/widgets/widget.js:2646`), but the popup's rotate handler stores `current + direction * steps` unbounded (`client/js/contextmenu.js:260`), so repeated clicks grow `rotation` to 720, 1080, … or negative values. That leaks into saved state and into anything comparing `rotation` values (e.g. an `IF` on `rotation == 90` never matching `450`). Applying the same `((x % 360) + 360) % 360` normalization used in `rotationStepIndex()` would match existing convention.

### 5. Validator misses submenu routines in custom-property collection

`validator/validate_gamefile.js:1184-1189` adds `entry.routine` of top-level `contextMenu` entries to `customProperties`, but doesn't recurse into `entry.menu`, so a routine referenced only from a submenu isn't registered. Worth recursing the same way `validateContextMenuEntries` does.

### 6. Runtime accepts image-list entries the validator rejects

`client/js/contextmenu.js:353` handles image entries of the form `{image: ...}` (object with an `image` key), but `validateContextMenuOptions` and the `CONTEXTMENU` operation validator only allow strings or arrays of strings. Either the object form is intended (then the validator is too strict) or it's dead code left from an earlier iteration.

## Smaller notes

- `CONTEXTMENU` silently acts on only the first widget of the collection (`client/js/widgets/widget.js:1129`). That's reasonable for a popup, but worth stating in the wiki/docs when this lands, since every other collection-based operation applies to all members.
- There's no way to navigate back out of a submenu (`client/js/contextmenu.js:566-576`) — the player has to close and reopen. Fine to defer, just noting it since submenu trees are a headline feature.

This PR touches `client/js/widgets/widget.js`, so it is an engine change that is excluded from beta auto-deployment by design — no `beta` label was added.

VTT-AI-Agent:
Thanks — all six findings were valid; fixed in 11722f6:

1. **`enlargePreviewIndex` never reset** — it now resets to 0 whenever the popup opens (including switching widgets during hold-and-move). Since the nav buttons already wrap modulo the list length, the value passed to routines can no longer be stale or out of range. Verified: navigate to index 5 on widget A, reopen on widget B (2 images), its routine receives `previewIndex: 0`.
2. **JSON-editor exclusion holes** — the document `mousedown`/`touchstart` handlers and `onlongtouch` now also check `jsonEdit`, so hold-and-move and long-touch stay out of JSON-editor sessions too.
3. **Overlay/loading guard** — the document-level handlers now bail out on `isLoading || overlayActive`, matching `inputHandler`, so right-click-hold can't act on widgets hidden under an overlay.
4. **Rotation normalization** — the numeric-steps rotate buttons now store `((x % 360) + 360) % 360` like `ROTATE`/`rotationStepIndex` (270 + 90 lands on 0, not 360).
5. **Validator submenu routines** — `getCustomPropertyUsage` now recurses into `entry.menu`, so routines referenced only from submenus are registered as used.
6. **Object-form image entries** — dead code from an earlier iteration; removed, so the runtime matches the validator (strings / arrays of strings only).

On the smaller notes: agreed on both — the first-widget-only behavior of `CONTEXTMENU` should go in the wiki text when this lands, and submenu back-navigation is a nice follow-up but deferred for now.

Verified locally with jest, the editor and public-library TestCafe files, and a Playwright script exercising the previewIndex reset and rotation wrap-around.

VTT-AI-Agent (commented):
## TL;DR

Re-review after commit 11722f6: all six findings from my previous review are correctly fixed, jest passes locally and the full CI matrix (including both TestCafe rounds) is green. I found no new issues in the fix commit — the branch looks in good shape for a draft; what remains are the PR's own deferred TODOs (edit-mode integration, hint icon).

## Verification of the fixes

1. **`previewIndex` reset** — `openContextMenu()` now resets `enlargePreviewIndex` to 0 (`client/js/contextmenu.js:415`). The prev/next buttons update the index modulo the list length and call `copyWidgetToPreview` directly (lines 203/209), so navigation still works and the value passed to button routines (line 368) is always in range. Confirmed fixed.
2. **JSON-editor exclusion** — the document `mousedown`/`touchstart` handlers and `onlongtouch` now also bail out on `jsonEdit` (`client/js/contextmenu.js:635,640`, `client/js/widgets/widget.js:127`). Confirmed fixed.
3. **Overlay/loading guard** — the same handlers now check `isLoading || overlayActive`, matching `inputHandler` in mousehandling.js. Both variables are top-level `let`s in serverstate.js/main.js, which are concatenated into the same bundle by `server/minify.mjs`, and the references sit inside `onLoad` handlers, so scope and declaration order are fine. Confirmed fixed.
4. **Rotation normalization** — the numeric-steps path now applies `((x % 360) + 360) % 360` like `Widget.rotate()` (`client/js/contextmenu.js:260`); the array path uses author-provided values verbatim, which is correct. Confirmed fixed.
5. **Validator submenu recursion** — `getCustomPropertyUsage` now recurses into `entry.menu` (`validator/validate_gamefile.js:1184-1192`). Confirmed fixed.
6. **Dead object-form image handling** — removed; runtime and validator now agree on strings/arrays of strings. Confirmed fixed.

## Remaining notes (all previously acknowledged, fine to defer)

- No back-navigation out of submenus; entering a submenu also resets the image preview to index 0 — both minor UX points.
- `CONTEXTMENU` acting on only the first widget of the collection should be called out in the wiki when this lands.
- The PR's own open TODOs (edit-mode integration, optional right-clickable hint icon) are still open, so I expect the branch to keep evolving.

This PR touches `client/js/widgets/widget.js`, so it is an engine change that is excluded from beta auto-deployment by design — no `beta` label was added.

VTT-AI-Agent:
> Rewrite the mouse handling to use the existing input handler.
>
> (Source: https://discord.com/channels/770758631146782780/1525148990264442880/1525258361262837911)

Done in 3c36050. The popup no longer installs its own document-level `mousedown`/`touchstart`/`touchend`/`touchcancel` listeners or dynamically bound move/up handlers. Instead, `inputHandler` in `mousehandling.js` calls a single `handleContextMenuInput(name, e)` hook (outside of edit mode and JSON-editor sessions); when the hook consumes an event, the regular widget interaction is skipped. Consequences:

- All the existing guards (`isLoading`, `overlayActive`, edit mode, `jeEnabled`, form-element targets) now apply to the popup automatically instead of being duplicated.
- The right-button special-case in the `mousedown` branch is gone — that block is back to its original shape, because the hook consumes right-clicks in play mode before it.
- The per-widget `contextmenu` listener only covers the classic enlarge behavior in edit/JSON-editor modes; play mode is fully handled centrally, and the holder/pile fallthrough walks the widget chain explicitly instead of relying on bubbling with `stopPropagation`.
- Net −38 lines.

Verified with jest, the full local TestCafe suite (all 9 files pass), and a 29-check Playwright script covering rotation, menu buttons, preview-only hold, classic enlarge, `rightClickRoutine`, holder fallthrough, hold-and-move switching (including from empty space), popup priority, left-drag, `CONTEXTMENU` op, edit mode, and mobile long-touch/move/tap-outside.

VTT-AI-Agent (commented):
## TL;DR

Re-review of commit 3c36050 ("route context menu mouse handling through the existing inputHandler"): the refactor is a clear improvement — no more dynamically bound document listeners, and it even fixes two operator-precedence bugs that exist on main. All my mouse-path and touch-path smoke tests pass except one real bug: **`contextMenu` action buttons (and the ℹ️ description trigger) do not respond to taps on touch devices** — the routine never runs and the popup stays open. This is not caused by 3c36050 (I reproduced it on 11722f6 too), but it should be fixed before merging since mobile support is an explicit goal of this PR.

## Findings

### 1. Popup action buttons are dead on touch devices

**What happens:** long-touch opens the popup fine, but tapping a `.contextMenuAction` button (or the info ℹ️ trigger) does nothing — no routine, popup stays open. Verified with emulated touch (Pixel 5, CDP touch events) on both 3c36050 and 11722f6.

**Root cause:** the popup only stops propagation of `mousedown` and `touchstart` (`client/js/contextmenu.js:612-624`). The `touchend` of the tap bubbles to `inputHandler`, whose unconditional `e.preventDefault()` (`client/js/mousehandling.js:54`) cancels the browser's synthesized `mousedown`/`mouseup`/`click`, so the button's `onclick` never fires. Instrumentation confirms it: the tap's `touchend` arrives with `target: SPAN, defaultPrevented: true` and no `click` follows.

**Why it's easy to miss:** the tag-name early-return at `client/js/mousehandling.js:33` fires *before* the `preventDefault` when `e.target.tagName === 'BUTTON'`. The rotation and prev/next nav buttons have no child elements, so their tap target is the `<button>` itself and they work. The action buttons and info trigger contain an icon `<div>` and label `<span>` that cover essentially the whole button, so their taps hit a child element, skip the early-return, and get `preventDefault`ed. Same for taps on the description popover.

**Suggested fix:** mirror the `mousedown` listener with a `touchend` one on `popupEl` so popup taps never reach `inputHandler`:

```js
popupEl.addEventListener('touchend', e => {
  if (e.target.closest('.contextMenuPopupBg') || e.target.closest(`#${CONTEXT_DESCRIPTION_POPOVER_ID}`))
    e.stopPropagation();
});
```

(A tap on the popup never starts a `touchActive` session, so skipping `handleContextMenuInput`'s touchend cleanup for these events is safe.)

### 2. Minor: stale `touchActive` if an overlay opens mid-session

If a routine opens an overlay while a long-touch popup session is active, the closing `touchend` early-returns at `client/js/mousehandling.js:28` (`overlayActive`) and never reaches `onTouchEndContextMenu`, leaving `touchActive` stuck. The next `touchmove` then gets consumed by `updateHoveredWidget` and can pop menus open during an ordinary drag until a `touchend` reaches the handler again. Self-healing and very edge-case (the mouse path has the `e.buttons & 2` self-heal at `client/js/contextmenu.js:533`; touch has no equivalent) — fine to leave, just flagging it.

## What I verified (all passing at 3c36050)

- Local browser smoke tests, mouse path: right-click opens popup; rotation row + menu buttons render; menu button click runs the routine and closes the popup; right-click → outside-click closes; hold-right-and-move from empty space onto a widget opens its popup (PCIO behavior) and stays open on release when buttons exist; left-click never opens the popup; right-click on trigger-less widgets does nothing; left-drag of plain widgets unaffected; rotation button applies `rotation: 45`.
- Touch path: long-touch opens, stays open on release with buttons, outside tap closes, touch-drag of plain widgets works and doesn't open popups.
- No stale references to the removed `handleContextMenu`/`startRightClickHold`/`closePopupAndStartHold`/`isOverPopup`; jest passes locally (6/6); full CI matrix including both TestCafe rounds is green.
- Nice bonus: the commit fixes two operator-precedence bugs in `inputHandler`'s mouseup/mousemove branches that exist on main (`name == 'mouseup' || (...) && mouseStatus[...]`).

The `beta` label was not added because this PR changes `client/js/widgets/` and other engine files, which are excluded from beta auto-deployment by design.

VTT-AI-Agent (commented):
## TL;DR

Ambitious, well-built feature (context menus/rotation/rich-preview popup), but it makes one **non-opt-in, global change to right-click semantics**: right-clicking *any* widget no longer fires `clickRoutine`/`doubleClickRoutine`, even for widgets that use none of the new properties. That's a real engine-behavior change (not incidental), so I'm not adding the `beta` label yet.

## Findings

### 1. Right-click stops triggering `clickRoutine`/`doubleClickRoutine` for every widget, not just ones using the new properties (mousehandling.js, widget.js)

`client/js/mousehandling.js` now calls `handleContextMenuInput(name, e)` for every non-edit mouse/touch event before falling into the normal target-based handling:

```js
let contextMenuHandled = false;
if(!edit && !jeEnabled)
  contextMenuHandled = handleContextMenuInput(name, e);

if(!contextMenuHandled && target && target.id) { ... }
```

Inside `contextmenu.js`, the `mousedown`/`mouseup` branches for `e.button === 2` return `true` **unconditionally**, with no check for whether the widget under the cursor declares `contextMenu`, `rotationSteps`, `contextMenuOptions`, or `rightClickRoutine`:

```js
if (name === 'mousedown' && e.button === 2) {
  closeContextMenu();
  rightClickActive = true;
  return true;
}
```

Before this PR, right mousedown/mouseup flowed through the same generic path as left clicks (only `e.button == 1`/middle is special-cased), so right-clicking a widget also fired `widget.click()` (and, per issue #2762, this lack-of-distinction is exactly what the reporter was working around). With this change, that path is now fully swallowed for every widget in play mode — including widgets that opt into none of the new features. Any existing published game that relies on right-click performing the normal click action (intentionally, or as a side effect of not having a separate right-click concept before) will silently stop working after this merges, with no migration path or opt-in gate. This is a real, deliberate change in core interaction semantics (matches the PR's own `engine` label), not an incidental touch of engine files — please call it out explicitly as a breaking behavior change in the PR description, and consider whether it should be flagged in `docs/compat.md`.

### 2. Minor: validator duplicates `contextMenu`/`contextMenuOptions` shape checks

`validator/validate_gamefile.js` defines `validateContextMenuEntries`/`validateContextMenuOptions` for the widget property, then re-implements very similar (but not identical, e.g. no per-widget routine-existence check) inline validators for the `CONTEXTMENU` operation's `contextMenu`/`widget` props (`operationProps.CONTEXTMENU`). Not a bug, but the duplication means the two validators can drift (e.g. the operation's `widget` prop validator repeats the `context.widgets[id] ... String(id).includes('$')` check from `validateContextMenuOptions` almost verbatim). Consider factoring out the shared piece.

## Beta label

The PR already carries the `engine` label, and per finding 1 the right-click behavior change is a real, global engine-behavior change (not incidental) — so I did **not** add the `beta` label.

VTT-AI-Agent:
Both findings were valid — fixed in 4cfbe31.

**1. Right-click no longer swallowed for every widget.** You were right: the mousedown/mouseup branches in `handleContextMenuInput` returned `true` unconditionally for the right button, with no check for whether the widget under the pointer opts into any of the new properties. Fixed: mousedown now only takes over the right button when the widget(s) under the pointer have `contextMenu`, `contextMenuOptions`, `rotationSteps`, `rightClickRoutine`, or `enlarge` — or when there's no widget at all under the pointer (needed so holding on empty space and moving onto a relevant widget still opens its popup). Any widget using none of those properties now falls straight through to the exact same generic mousedown/mouseup path that existed before this PR, so right-click on it still fires `clickRoutine`/`doubleClickRoutine` (and can still drag it, matching the old lack-of-button-distinction). I added a regression test asserting this explicitly (right-click on a plain widget, and on a plain child inside a plain holder, both still fire `clickRoutine`), plus reran the full local TestCafe suite and the existing behavior checks — all pass.

Good catch on issue #2762 being exactly the case this could have broken for people relying on the *old* lack of right-click distinction — that's now preserved for anyone who wasn't using the new opt-in properties.

**2. Validator duplication.** Factored out `isStringOrStringArray` and `validateWidgetIdList` used by both `contextMenuOptions` and the `CONTEXTMENU` operation's `image`/`widget` props, and `validateContextMenu` now takes an optional `widgetOverride` so the operation validator reuses it directly instead of re-implementing `validateContextMenuEntries` inline. The operation intentionally still skips the per-widget routine-existence check (documented in a comment) since its target widget is only known at runtime via the `collection`.

Verified with jest, `validate_gamefile_node.js` against a fixture exercising `contextMenu`, `contextMenuOptions`, and the `CONTEXTMENU` operation together, the full local TestCafe suite, and Playwright checks covering the exact regression scenario plus all existing popup behaviors (rotation, menu buttons, preview-only hold, classic enlarge, `rightClickRoutine`, holder fallthrough, hold-and-move, edit mode, mobile).

VTT-AI-Agent (commented):
## TL;DR

The latest input-handling fix still leaves three regressions: legacy `enlarge` widgets no longer preserve their old right-click click/drag behavior, popup actions remain untappable on touch devices, and a long press disables dragging even on ordinary widgets. The first issue already affects widgets in published library games, so this needs another pass despite the green CI matrix.

## Findings

### 1. Legacy `enlarge` widgets still lose their old right-click behavior

`client/js/contextmenu.js:40-42,526-545` treats `enlarge` as a reason to consume the right-button `mousedown` and `mouseup`. However, `enlarge` is not one of the new opt-in properties: on `main`, those events go through the normal `mouseStatus` path, while the separate `contextmenu` event only calls `showEnlarged()`. Consequently, an enlarged widget used to enlarge *and* execute its normal `clickRoutine` on release (or be right-dragged), but this branch only enlarges it at lines 511-513 and swallows the release.

This is observable in existing content, not just a hypothetical compatibility concern: for example, `library/games/Floor of City/0.json` has many widgets with both `enlarge` and `clickRoutine`, as do `library/games/Poker/0.json` and the Appearance/Timer tutorials. Removing legacy `enlarge` from `reactsToRightClick()` would let its down/up events keep following the pre-PR path while the `contextmenu` event can still show the classic enlargement.

### 2. Context-menu action buttons still do not work on touch devices

`client/js/contextmenu.js:627-639` stops `mousedown` and `touchstart` inside the popup, but not `touchend`. A tap on an action's child `<span>`/icon therefore bubbles to `inputHandler`, where `client/js/mousehandling.js:33-54` does not take the form-control early return (the target is not the `<button>`) and calls `preventDefault()`. That cancels the synthesized `click`, so action routines and the description-info button do nothing. The rotation/navigation buttons happen to work because their tap target is the button itself.

This was raised in the review of `3c36050` and remains present at the current head. Stopping propagation of popup-background `touchend` events, just as the existing `mousedown` listener does, prevents the room input handler from cancelling the button click.

### 3. The “empty-space” long-touch timer also activates on every ordinary widget

`client/js/contextmenu.js:562-571` says it starts a timer for a long touch on empty space, but `widgetAtPoint()` only returns widgets with popup triggers (`contextMenuOptions`, `rotationSteps`, or `contextMenu`). A plain movable widget therefore also starts this timer. After 500 ms it sets `touchActive = true`; subsequent `touchmove` events are consumed at lines 556-560 instead of reaching the existing drag path. Since those moves never update `mouseStatus`, releasing can additionally run the widget's click action as though the finger had not moved.

The empty-space test needs to distinguish “no widget at this point” from “a widget is present but has no popup trigger”; otherwise long-press-then-drag regresses for every normal game piece.

## Verification

All remote checks are green. The PR adds no committed regression coverage for these new mouse/touch paths, and local Jest could not be rerun in this fresh clone because dependencies are not installed.

I did not add the `beta` label because the changes to `client/js/widgets/widget.js` and the routine/input semantics are substantive engine behavior changes, making the PR engine-ineligible.

VTT-AI-Agent (commented):
## TL;DR

The latest fixes preserve legacy mouse right-click behavior, but mobile still has one functional regression: long-touching a widget with `rightClickRoutine` runs both that routine and the normal `clickRoutine` when the finger is released. This should be fixed and covered with a touch regression test before merging.

## Finding

### Long-touching `rightClickRoutine` also fires the normal click routine

`client/js/contextmenu.js:592-597` marks `longTouchHandled`, runs `rightClickRoutine`, and returns, but unlike the popup and classic-enlarge branches at lines 604 and 611 it never adds the `longtouch` class to the widget. The original `touchstart` has already created a normal `mouseStatus` entry; on release, the widget's `touchend` handler calls `onTouchEndContextMenu()` (`client/js/widgets/widget.js:122-126`), which does nothing because this branch never sets `touchActive`, and then the event bubbles to `inputHandler`. There, `client/js/mousehandling.js:100-119` sees the untouched `mouseStatus`, does not find the `longtouch` class, and calls `widget.click()`.

Concretely, a widget defining both `rightClickRoutine` and `clickRoutine` performs both actions after a 500 ms press, even though the PR promises that `rightClickRoutine` "will just be called." Marking this branch as a consumed long touch (as the two neighboring branches already do), or otherwise clearing/suppressing its normal touch-click state, prevents the second routine; a touch test with distinct state changes in the two routines would catch the regression.

I did not add the `beta` label because this PR genuinely changes base-widget input behavior and adds routine-operation semantics, making it engine-ineligible.

VTT-AI-Agent:
Fixed all findings from review 4699245293, plus the immediately following `rightClickRoutine` touch finding, in d0081134:

- Removed legacy `enlarge` from the new right-button interception, so its mouse down/up still use the classic click/drag path while the `contextmenu` event enlarges it.
- Added popup `touchend` isolation so taps on action labels/icons and description controls are no longer cancelled by the room input handler.
- The hold-and-move timer now starts only when `widgetsAtPoint(...)` is actually empty, so a long hold does not hijack ordinary widget dragging.
- Long-touching a `rightClickRoutine` widget now marks the touch as consumed, preventing its normal `clickRoutine` from also firing on release.

Added four focused Jest regressions covering those boundaries. Verified with the full Jest suite (19/19), production minification, and a real Chromium mobile/touch smoke test covering tappable actions, right-click-routine exclusivity, ordinary long-hold dragging, and classic `enlarge` right-click behavior. Also merged current `main` in 4d975ba2 without rebasing or force-pushing.

VTT-AI-Agent:
CI note: the two red checks on 140cf63d are both `tests/testcafe/editor.js` → *Renaming a widget keeps its color controls clear and it movable*, failing with an unexpected native alert. That is the Chrome 152 duplicate-`change`-event issue in the widget id input (`client/js/editor/sidebar/properties.js`), which #3164 fixes; it hits `main` as well and depends on which Chrome the runner image draws (the third `editor.js` job on the same commit passed). This PR does not touch that code or that test, and the test passes locally on Chrome 151. I will merge `main` here once #3164 lands.

VTT-AI-Agent:
The properties-sidebar UI for the three new widget properties is in (8ac5bf83), built as approved in the Human feedback thread:

> Build it as proposed
>
> (arnoldsmith86, answering the ask-human question about how rotationSteps, contextMenu and contextMenuOptions should appear in the properties editor)

**What it looks like**

- *Interaction & display* → **Rotation steps** below *Enlarge*: a number (`45`, stored as a number) or a list (`0, 90, 180, 270`, stored as an array); text that is neither is marked and writes nothing. [screenshot](https://agent.virtualtabletop.io/reports/pr/1525148990264442880/ui/01-interaction.png)
- New collapsible **Context menu** section right after it ([empty](https://agent.virtualtabletop.io/reports/pr/1525148990264442880/ui/02-section-empty.png)):
  - **Popup on right-click** switch: writes `contextMenuOptions: {}` / `null`, so a preview-only popup can be switched on without the JSON editor. It is locked (on) while entries or rotation steps open the popup anyway.
  - **Menu entries**: expandable rows like the face rows — icon chip, text and a "runs dealRoutine" / "opens a submenu with 2 entries" summary in the head, ▲▼ and delete on the right; the body has Text, Icon (picker), Routine (dropdown of the widget's `*Routine` properties, refreshed when one is added under Automations), Color, Description, and *Add submenu*, which nests the same rows one level in with *Add submenu entry* / *Remove submenu*. [entry](https://agent.virtualtabletop.io/reports/pr/1525148990264442880/ui/03-entry.png) · [submenu](https://agent.virtualtabletop.io/reports/pr/1525148990264442880/ui/04-submenu.png)
  - **Preview** (`contextMenuOptions`): Size factor (placeholder shows the effective default from `enlarge` or 2), Title, Background color, an image list (each row an image picker, add opens the picker for the new slot) and a widget list (ids typed or picked with the usual widget popout). Emptying a field removes its key; the object stays so the popup stays on. [widgets](https://agent.virtualtabletop.io/reports/pr/1525148990264442880/ui/05-preview-widgets.png) · [images](https://agent.virtualtabletop.io/reports/pr/1525148990264442880/ui/06-preview-image.png) · [resulting popup](https://agent.virtualtabletop.io/reports/pr/1525148990264442880/ui/07-popup.png)
- The section is single-widget like the snap grid (a multi-selection has no safe merge of two menus); the three properties are excluded from the generic *Other properties* list, which also gets hints for them.

**Verification**: new TestCafe test *The context menu section builds the right-click popup of a widget* (rotation steps, add/rename/reroute/submenu/reorder/remove entries, preview factor + widgets, then the popup opening in play mode), the face/dice/deck-editor tests that share the reorder control still pass, jest 1655 passed, and a Playwright walk of the whole section against the local server (60 checks) is green.

VTT-AI-Agent (commented):
## TL;DR

A big, carefully built feature: right-click popup, rotation steps, submenus, the `CONTEXTMENU` operation and the new properties-sidebar section all work in a real browser run against this branch. Two confirmed problems should be fixed before players test it: **using the popup after its widget was deleted by another player (or after a state reload) crashes the client**, and **on touch devices a long press on any ordinary widget now runs its click routine on release**, which it does not do on `main`. Smaller items: `CONTEXTMENU` inside a `rightClickRoutine` shows nothing unless `collection: 'thisButton'` is given, validator false positives for inherited / `cardDefaults` routines, no tutorial-quality demo room, and thin engine test coverage.

## Findings

### 1. Client crash: the popup outlives its widget (confirmed)

`client/js/contextmenu.js:425-451` keeps `currentWidget` and the button closures alive until click-outside / Escape, and nothing closes the popup when that widget goes away: neither `Widget.applyRemove()` (`client/js/widgets/widget.js:427`) nor `receiveStateFromServer()` (`client/js/serverstate.js:497`) knows about it.

Repro (two clients, [script](https://agent.virtualtabletop.io/reports/review/1545569961060073472/pr-2852/pw-check.cjs), step 8): A right-clicks `b1` (`rotationSteps: 45` + `contextMenu`); B clicks a button whose routine `DELETE`s `b1`. A's popup stays open showing the gone widget ([screenshot](https://agent.virtualtabletop.io/reports/review/1545569961060073472/pr-2852/08-stale-popup.png)); A clicks a rotate button → `currentWidget.set('rotation', …)` (`contextmenu.js:270`) → `TypeError: Cannot read properties of undefined (reading 'unalteredState')` in `addDeltaEntryToUndoProtocol` (`serverstate.js:420`) → error overlay, client dead ([screenshot](https://agent.virtualtabletop.io/reports/review/1545569961060073472/pr-2852/09-after-stale-rotate.png)). The menu-entry buttons take the same path through `runRoutine` (`contextmenu.js:368-372`). Loading another game while a popup is open leaves the same stale popup behind; every routine it runs then executes on a detached widget object. Once the undo-protocol guards from #3011 / #3093 land, that write no longer throws but reaches the server as a delta for an unknown id, which `Room.receiveDelta` (`server/room.mjs:804-805`) stores as a new partial widget.

Fix: export something like `closeContextMenuFor(widget)` (close only when `currentWidget === widget`) and call it from `applyRemove()`, call `closeContextMenu()` from `receiveStateFromServer()`, and guard `rotate` and the routine buttons with `widgets.get(currentWidget.id) === currentWidget` for the in-flight case.

### 2. Touch regression for every existing game: a long press now clicks on release (confirmed against `main`)

`client/js/widgets/widget.js:211-217` routes play-mode long touches to `onLongTouch()`. For a widget with no `rightClickRoutine`, no popup triggers and no `enlarge`, `contextmenu.js:611-618` returns without adding the `longtouch` class that the old `onlongtouch` added unconditionally — and that class is what makes `mousehandling.js:225` skip the click on `touchend`.

Measured with CDP touch events ([script](https://agent.virtualtabletop.io/reports/review/1545569961060073472/pr-2852/pw-touch.cjs)), 800 ms hold without moving, on a plain widget with a `clickRoutine`: `main` → no click; this branch → the click routine runs. So a touch player who holds a finger on a card, token or button for more than half a second and lifts it now flips / clicks it. If this is intended, say so in the PR body (it changes behavior for touch players in every game); otherwise add `widget.domElement.classList.add('longtouch')` in that fall-through branch. A jest case next to the `rightClickRoutine` one in `tests/client/contextmenu.test.js` would pin it.

### 3. `CONTEXTMENU` in a `rightClickRoutine` shows nothing by default (confirmed)

`widget.js:1540` defaults `collection` to `DEFAULT`, but `evaluateRoutine` only seeds `playerSeats`, `activeSeats` and `thisButton` (`widget.js:1330-1335`); `DEFAULT` does not exist until a `SELECT` ran. The PR body's headline use case — "If `rightClickRoutine` is defined … you can do that yourself" — therefore fails silently: `rightClickRoutine: [{ "func": "CONTEXTMENU", "contextMenu": [...] }]` logs "Collection DEFAULT does not exist" and no popup appears (script step 5; step 6 shows `collection: 'thisButton'` working). The validator at least reports it as an undefined collection in edit mode.

Suggestion: when `collection` is not given, target the widget running the routine (check `a.collection === undefined` before `setDefaults`, like `CALL`'s `widget` default), and reflect that in the routine-editor template (`client/js/editor/controls/routine.js:608-640`) and the `popup.js` info text.

### 4. Validator false positives on routines the engine resolves (confirmed)

`validator/validate_gamefile.js:722` checks `widget[entry.routine]` on the raw JSON, but `widget.get()` resolves `inheritFrom`, and cards get their properties from the deck's `cardDefaults`. A file with `b3: { "inheritFrom": "b1", "contextMenu": [{ "routine": "flipRoutine" }] }` (routine defined on `b1`) and a card whose routine sits in `cardDefaults` produces

```
b3[contextMenu.0.routine]: routine 'flipRoutine' does not exist on this widget
c1[contextMenu.0.routine]: routine 'cardFlipRoutine' does not exist on this widget
```

while both buttons work in the room. Skip the existence check (or resolve through `inheritFrom` / the card's deck) when the widget has `inheritFrom` or is a `card`.

### 5. Demo room (standing rule)

The PR body points at the generic `PR-2852/pr-test` room. For three new widget properties plus a new operation the rule asks for a tutorial-quality demo on this PR's `pr-test-ai` room, published as a `.vtt` and not committed under `library/`, split into variants — e.g. "Properties - Context Menu": 1) `rotationSteps` + `contextMenuOptions` (factor, title, color, image and widget lists with prev/next), 2) `contextMenu` entries with icons, colors, descriptions and submenus, 3) `rightClickRoutine` + `CONTEXTMENU` with `property`, overrides and `previewIndex`. Good: the diff touches nothing under `library/`.

### 6. Engine test coverage is thin

`tests/client/contextmenu.test.js` covers input arbitration only, and `tests/testcafe/editor.js` ends with a single right-click. Nothing exercises the `CONTEXTMENU` operation, rotation through the popup (the one path that writes state — ideal for `compareState`), submenus, hold-and-move, or the removal case from finding 1. One TestCafe test that right-clicks, rotates twice, runs an entry routine and compares the state hash would pin this on both browsers.

### 7. Smaller points

- `contextmenu.js:523-526`: a card with `enlarge` inside a holder with `contextMenu` gets enlarged and the holder's popup never opens, although the card's right-click mousedown was already consumed for the holder (`:543`), so the card also loses its normal right-click click. The PR body says cards without popup triggers fall through to the holder; either let `enlarge` not stop the fall-through, or document that it does.
- Properties sidebar: for a card whose `contextMenu` comes from `cardDefaults`, the section says "No buttons yet." (own value) while the header summary counts the inherited entries; showing inherited entries read-only, or naming where they come from, would avoid the contradiction.

## UI support and verification

UI support for the new API is in place: `rotationSteps` under *Interaction & display*, the *Context menu* section for `contextMenu` / `contextMenuOptions`, `rightClickRoutine` in the Automations event list, and `CONTEXTMENU` as a block with clauses in the routine editor (its entry list is a raw JSON popup like `CLONE`'s `properties`; reusing the sidebar's entry rows there would be a nice follow-up).

Verified: jest 1655/1655 locally; a Playwright run against a local server on this branch ([popup screenshot](https://agent.virtualtabletop.io/reports/review/1545569961060073472/pr-2852/01-popup.png)) covering right-click, rotation, entry routines with `previewIndex`, submenus, holder fall-through, `rightClickRoutine` + `CONTEXTMENU`, and the removal case; the touch script on both this branch and `main`; `validate_gamefile_node.js` on a hand-written file using every new property and operation. No server changes; strings reach the DOM through `textContent`, colors through style properties, icons through the existing symbol pipeline, and CodeQL is green.

I did not add the `beta` label: the `widget.js` changes alter default engine behavior for every widget (play-mode right-click routing and the long-touch path in finding 2) and add a routine operation, which excludes this PR from the beta build by design, and the confirmed client crash in finding 1 would make it unsafe for players to test anyway.

VTT-AI-Agent:
Two follow-ups landed on the branch.

**PCIO importer** (e5104b7c), prompted on Discord:

> Double check that this updates the PCIO importer to use now supported features.
>
> (Source: https://discord.com/channels/770758631146782780/1525148990264442880/1545657711067267122)

It did not yet - the importer still warned that *Rotation Snapping* has no equivalent and mapped *enlarge* to the hover `enlarge`. Now a collection's `snapAngles` (more than one angle) becomes `rotationSteps` in the deck's `cardDefaults`, and `enlarge` becomes `contextMenuOptions: { factor: 3 }`, because PCIO enlarges a card while the right button (or a finger) is held on it - which is exactly what the popup does, while VTT's `enlarge` shows on hover. The warning is gone, and a jest case pins both mappings. I left the `random` rotation mode alone (it still picks a multiple of 45°); if PCIO restricts a random turn to the snap angles as well, that would be a small follow-up.

**Review of 2026-09-05** (aec6a5df) - all seven findings, in order:

1. **Popup outlives its widget:** `Widget.applyRemove()` now closes a popup that shows the widget (`closeContextMenuFor`), which also covers a state reload since `receiveStateFromServer` removes every widget through it. The rotate and menu buttons and the deferred `CONTEXTMENU` open additionally check that the widget is still in the room. Jest: *the popup closes when the widget it shows is removed*.
2. **Long touch clicks on release:** `onLongTouch` adds the `longtouch` class for every widget again, before it decides whether anything reacts - so a plain widget behaves as on `main`. Jest case added.
3. **`CONTEXTMENU` default collection:** without `collection` the popup opens on the widget running the routine (like `CALL`'s `widget`); the routine editor sentence reads "Show a popup on this widget …" / "on the first of the picked widgets" / "on the first of cards", the info popup and the JSON editor snippet follow, and jest checks the engine and the sentences. The demo room's dealer button, which relied on the old default after a `SELECT`, now says `collection: "DEFAULT"`.
4. **Validator false positives:** a menu entry's routine is resolved through `inheritFrom` (string, list and `!` exclusion forms) and a card's deck `cardDefaults` before it is reported as missing; unresolvable sources (a deck or source widget not in the file) are given the benefit of the doubt. Inherited routines are not re-validated on the inheriting widget.
5. **Demo room:** the tutorial *Properties - Context Menu* (three variants) has been on this PR's `pr-test-ai` room since the previous round; the PR body now links it and describes the variants. It stays out of `library/`.
6. **Engine coverage:** new `tests/testcafe/contextmenu.js` (added to both browser matrices) right-clicks, rotates twice, runs a menu entry, opens a submenu whose routine receives `previewIndex`, and clicks a button whose `CONTEXTMENU` has no collection - then compares the state hash. Jest covers the removal case, the long-touch class and the default-collection engine path.
7. **Smaller points:** a card with the classic `enlarge` no longer stops the fall-through, so the holder's popup opens (verified in a browser: the holder's entry runs, a card in a plain holder keeps hover enlarge and its normal right-click). The sidebar shows a card's or inheriting widget's entries as rows already; it now says above them where they come from ("These entries come from the deck deck1. Changing them here stores a copy on this widget.").

Also fixed on the way: `symbols.js` used `div`, `$a`, `mapAssetURLs`, `removeFromDOM`, `domByTemplate` and `toggleClass` as bundle globals without importing them, which is why the popup could not render under jest.

Verified with jest (1660 passed), the new TestCafe test and the existing editor context-menu test on Chromium, a 46-check Playwright popup walk, and a 28-check walk of the three demo variants.

VTT-AI-Agent:
CodeQL flagged the image preview of the popup (`client/js/contextmenu.js:149`, alerts 241 and 244): the picture of an `image` list was an `` whose `src` came straight from `contextMenuOptions`. 05058419 draws it as a CSS background image instead, the way the image widget and card faces draw their pictures, with quotes, backslashes and line breaks in the URL escaped for `url()`. Both alerts should close with the next scan; the popup smoke walk (46 checks), the demo walk (28 checks) and jest still pass, and the gallery variant of the demo room shows the pictures as before.

VTT-AI-Agent (commented):
## TL;DR

At `05058419`, three reproducible bugs remain: a rotation-triggered routine can leave the client in its error overlay, an open popup can still act on a widget after it becomes private, and array-based rotation can skip the next allowed angle. The editing UI and published demo-room requirements are covered.

*Manually requested review (code) — by `gpt-6-astra`*

## Findings

### 1. [P2] Recheck the popup target after awaiting rotation

**`client/js/contextmenu.js:267-280`** — The liveness check runs before `await currentWidget.set('rotation', next)`, but that call also awaits property-change routines. A valid widget such as:

```json
{
  "rotationSteps": 90,
  "rotationChangeRoutine": [
    { "func": "DELETE", "collection": "thisButton" }
  ]
}
```

reliably reproduces this: right-click it, then click a rotation button. Deletion correctly closes the popup and sets `currentWidget` to `null`, but line 280 immediately passes that null into `copyWidgetToPreview`. Chromium reports `TypeError: Cannot read properties of null (reading 'get')` in `getPopupOptions`, and the client error overlay interrupts play. Closing the popup while a rotation routine awaits `DELAY` has the same underlying race.

Capture the target before the await, then verify that it is still live **and still the current popup target** before repainting. The existing removal test only removes the widget before another action; add coverage for removal during the rotation's awaited routine.

### 2. [P2] Invalidate the popup when its widget becomes inaccessible

**`client/js/contextmenu.js:49-50,378-382,502-504`** — A popup is invalidated on removal, but ownership/visibility changes leave the same widget in the map, so `isLiveWidget` continues to authorize its actions. The copied preview also retains its old visibility classes.

Reproduced in Chromium: open a widget's popup, change its `owner` to another player, then click its menu action. The original widget has computed `display: none`, while the popup still displays its contents and the action successfully writes `marked: true` to that now-hidden widget. In a game, this allows a player to keep operating a piece after another player takes it into a private hand.

Close or refresh the popup when relevant widget/ancestor visibility changes, and recheck whether the target is accessible before running an action or rotating it. Cover an ownership change while the popup is open, not only deletion.

### 3. [P2] Choose the next allowed angle in the requested direction

**`client/js/contextmenu.js:243-255,275-276`** — `rotationStepIndex` first picks the nearest allowed angle, then the handler always advances one more array entry. When the current rotation is between steps, that can skip the next allowed angle.

With `rotationSteps: [0, 90, 180, 270]` and `rotation: 30`, clicking **Rotate left** produces **270°**, reproduced in Chromium; the next allowed angle to the left is **0°**. This occurs when a widget starts between steps or another routine rotates it there. Select the next angle in the requested direction relative to the actual current rotation, and test off-step starting rotations as well as exact quarter turns.

## Verification and beta assessment

- All **1,660 Jest tests / 57 suites** pass locally; the new context-menu TestCafe test passes in Chromium. Browser checks also confirmed normal menu actions and tapping a menu action after a long touch.
- The properties sidebar exposes the new properties, and the visual routine editor exposes `CONTEXTMENU` and `rightClickRoutine`.
- The linked `.vtt` contains three tutorial variants, matching metadata fields and a square SVG library image; all three variants pass validation. The demo is present in the PR's `pr-test-ai` game list, and the diff contains no `library/` changes.
- I assessed the engine hunks themselves: `client/js/serverstate.js` only exports the existing `setDeltaCause` helper and is incidental. Much of `client/js/widgets/widget.js` is opt-in, but its `contextmenu` listener at lines 187-193 unconditionally stops calling the legacy play-mode `showEnlarged(e)`, including for widgets with only `enlarge`; the replacement handler has no legacy fallback. That changes default event behavior independently of the new properties, so this is not solely an incidental engine-file touch.

I did not add the `beta` label because the unconditional change to legacy widget context-menu handling makes this PR engine-ineligible, and the reproduced client error and actions on newly private widgets also need correction.

VTT-AI-Agent:
Thanks for the review - all three findings are fixed in 2a98f510, and the legacy right-click listener point as well.

1. **Rotation race** - the rotation handler keeps its own reference to the widget and, after `await set('rotation')`, only continues if the popup still belongs to that widget. A `rotationChangeRoutine` that deletes it, or a `DELAY` during which the popup is closed or reopened on another widget, now ends quietly. The preview repaint moved into the delta hook below, so nothing is left that could repaint a widget that is gone. Jest: the deleting rotation routine and a routine that opens the popup elsewhere; TestCafe: `doomed` deletes itself on rotation, and the test asserts the popup is gone, the widget is gone and `#clientErrorOverlay` stays hidden.

2. **Newly private widgets** - `Widget.applyDeltaToDOM` now calls `updateContextMenuFor(widget, delta)`. When the changed widget is the popup's widget or one of its DOM ancestors, the popup closes if the widget carries the `foreign` or `hidden` class (owner, `onlyVisibleForSeat`, `linkedToSeat`, `display`, its own or a parent's) and otherwise repaints the preview once per frame. Menu actions, the rotation buttons and the `CONTEXTMENU` operation check that same accessibility before doing anything, so a hide that arrives without a delta of the widget itself (a seat change, for instance) is caught at action time too. Jest covers an owner change, a hidden parent and an action on a hidden widget; TestCafe hands `hero` to another player through a delayed routine while its popup is open and expects it to close.

3. **Skipped angles** - `nextRotationStep` picks the step that is closest in the requested direction from the actual rotation (30° → *Rotate left* → 0°, *Rotate right* → 90°), keeps the array's spelling of the angle (`[-90, 0, 90]` writes `-90`) and leaves a single-entry list where it is. Jest walks an unsorted list starting at 30°; the TestCafe widget now starts at 120° and its first click lands on 90°.

4. **Legacy listener** - the widget's `contextmenu` listener is main's `this.showEnlarged(e)` again in every mode; the popup is layered on top of it through the input handler in `mousehandling.js`, so a widget without any of the new properties behaves exactly as before.

Verified with the full jest suite (1666 tests), `tests/testcafe/contextmenu.js` in Chromium, the Playwright popup smoke run (46 checks, mouse and touch), the demo-room walk (28 checks) and a browser check that classic `enlarge` still shows on right-click.

96LawDawg:
On iOS mobile (Android not tested), these new context menu features interfere with drag. The menu (including enlarging and rotation steps) show up after pressing and holding for a second or so. But if you want to drag the widget, then the menu gets in the way. Dragging should take priority over the menu and the menu should not show up when dragging a widget.

VTT-AI-Agent:
Thanks @96LawDawg — fixed in 329411b6. The widget's long-touch timer kept running while the finger was already dragging, so a drag that took longer than half a second opened the popup mid-drag and then ate every further touchmove, which is why the widget froze under your finger.

Now a finger that moves more than about 10 px from where it touched down counts as a drag and cancels the pending long touch, so no popup, enlargement or rotation buttons appear while you drag. The same applies to a swipe that starts on empty space. Holding still on a widget for the delay still opens its popup, and moving on *after* it has opened still switches the popup between widgets (the hold-and-move behavior) rather than dragging — that gesture is the same as before.

Verified with emulated touch in Chromium (drag right away, hold-then-drag, hold-and-move, swipe from empty space) plus jest tests for the new rule. If you get a chance to try it again on iOS on the test server, that would be great.

96LawDawg:
If "enlarge" is added to a widget that has a contextMenu with a "factor" (or vice versa), things do not work correctly until after the room is refreshed. After the refresh, enlarge works as it should and if you right click and get the factor, the view added by "enlarge" goes away as it should. However, before the refresh, the enlarge view stays in place until the mouse has moved far away and both the enlarged and factor views are on screen at the same time. This may not be a problem during game play, but we have a rule that if it takes a screen refresh to fix it, it is broken.

VTT-AI-Agent:
@96LawDawg Thanks, reproduced and fixed in 063a01fa. The property change itself was fine; the trigger is that the widget had been **selected in edit mode**. The selection deliberately survives leaving edit mode (so it is still there when you come back), and the classic enlarge code refused to hide its hover copy for any widget that is selected - which is meant for inspecting a widget while editing, but kept the copy pinned on screen while playing on, so it sat next to the popup until a reload cleared the selection.

Changes:
- `hideEnlarged` only honors the selection while edit mode or the JSON editor is active.
- The popup hides the enlarged copy the moment it opens, instead of waiting for the next mouse move to leave the widget.
- Leaving edit mode hides a copy that was pinned there by a selection.

Covered by jest tests and the TestCafe popup test (hover shows the copy, right-click replaces it with the popup); a Playwright walk of both directions (adding `enlarge` to a `factor` widget and vice versa, live, after selecting in edit mode, and after a reload) passes.

PR #2856, opened 2026-02-06
Public Library - Update Blob

RaphaelAlvez:
Removes AI art

Changes game pieces to use inheritFrom

Fixes Background-size in small pieces

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2856/pr-test (or any other room on that server)


ArnoldSmith86:
Nice!
- Board is movable.
- "Blob" should be "blob" but I have no idea why.. 😄
- "Blob" is exactly under the star. Shouldn't be hard to move, right?
- Should the blue ones maybe be mirrored?

ArnoldSmith86 (approved):

PR #2857, opened 2026-02-06
Decoupling of title and similar to

RaphaelAlvez:
Changes CSS so "Overlay name" only removes the Main title.

"Similar to" div no longer display margins when empty

"Similar to" div is 1px lower than before. (just because it looks better)

Public library:

Removes tilte banner when tile is in the cover

In some games the title was not clear or covered by the start icon.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2857/pr-test (or any other room on that server)


96LawDawg (approved):
I changed the wording of the checkbox slightly. Everything looks good and the right PL games have the change. Approved.

PR #2859, opened 2026-02-06
unified star and AI badge positions on library images

RaphaelAlvez:
Before

Image

Image

after

Side bar
image

full screen
image

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2859/pr-test (or any other room on that server)


96LawDawg (approved):
Looks good to me. I tested on Chrome, Firefox, and iOS. Did not see any problems or differences when changing size, orientation, etc.

PR #2860, opened 2026-02-06
PL - Update cover art of some games

RaphaelAlvez:
Updates some cover art just to move the titlefrom underneath the star icon

image

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2860/pr-test (or any other room on that server)


96LawDawg:
I agree with all of this except for Scrappet.  Moving the text is a copy of a copy of a copy and it is pixelated and does not look good. I do not think the creator would like this change. 

The original art is at https://zoql.itch.io/scrappetkaiserprincess.  When working on this I asked the author to create a square image for the PL.  If you can source the original and make the text look better, then I would approve.

As for the others, we either already have author buy-in or the stuff is CC0.

ArnoldSmith86:
Here's the initial library image of the game: [original.png.zip](https://github.com/user-attachments/files/25143315/original.png.zip)

PR #2862, opened 2026-02-07
Fix AI imagery notification incorrect on shared games (fixes #2861)

### LINKED ISSUE #2861: AI imagery notification incorrect on shared games
When you access a game through a shared link like https://virtualtabletop.io/game/letter-words the display that pops up always says the game uses AI imagery even though it does not.

ArnoldSmith86:
Explicitly hide the AI badge and ai-imagery-notice in the link details overlay when the game does not use AI imagery, so shared links like /game/letter-words no longer incorrectly show the AI imagery message.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2862/pr-test (or any other room on that server)


ArnoldSmith86:
> I don't think this can be tested on the test servers[...].

Why not? Shouldn't make a difference.

96LawDawg:
I didn't know it would do the test.  I tested and got the same result as you.  Still shows the note.  Obviously.

PR #2863, opened 2026-02-07
Execute JSON editor Tab selection when using only arrow keys (fixes #2754)

### LINKED ISSUE #2754: `Tab` doesn't work without filter term
When using `Tab` in the JSON editor, you can type and/or use arrow keys. But _only_ using arrow keys makes it impossible to actually execute the selected command.

ArnoldSmith86:
Fixes #2754

When using Tab in the JSON editor, selecting a command with arrow keys only (no filter term) did not execute on Tab release. Tab release now runs the highlighted command whenever any button is highlighted (from filter or arrow keys).

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2863/pr-test (or any other room on that server)


96LawDawg (approved):
Approved through test.  This fixes the specified problem and does not change anything else that I tested.

PR #2867, opened 2026-02-07
fixed that new states no longer overlay their name

### LINKED ISSUE #2865: New game "overlay name" should be enabled by default
When saving a new game, the "overlay name" button is not checked by default so the name does not show up on the game shelf. This would be really confusing as there could be a games without obvious names.

ArnoldSmith86:
Fixes #2865.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2867/pr-test (or any other room on that server)


96LawDawg (approved):
Approved

PR #2868, opened 2026-02-07
fixed assets in html cards

### LINKED ISSUE #2733: asset names in iframe card objects are not URL mapped
Image

This card in Wildscapes is broken on PR test servers because it tries to access `/assets/<id>` which should be mapped to `assets/<id>`.

ArnoldSmith86:
Fixes #2733.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2868/pr-test (or any other room on that server)


96LawDawg (approved):
Approved.

PR #2871, opened 2026-02-07
insert operations with common properties

### LINKED ISSUE #2552: Editor Prefills Routine Parameters
The idea (from [Discord](https://discord.com/channels/770758631146782780/1363913793054445660/1363916878875066581)): When clicking on buttons in the Editor for routines, fill in the most commonly used parameters.  Maybe even have multiple options for common branches like IF with condition and IF with operands or MOVE from holder or MOVE a collection.  And maybe shift click inserts a function without parameters like we have now.

ArnoldSmith86:
Fixes #2552.

When inserting a routine operation in the JSON editor, the most commonly used properties are prefilled, so you get a working skeleton instead of a bare `{"func": …}`. A **prefill properties** toggle in the command panel turns that off and on; it appears whenever the cursor is somewhere an operation can be inserted and is remembered across sessions.

The prefilled properties follow a common grammar: function name, input collection/widget/source, mode/relationship, operation-specific properties, count/value, output collection/variable/result.

Additionally, a new *[] anonymous collection* editor command appears when the cursor is on an input collection parameter (`collection` of the 14 operations that read one, `source` of CLONE/SELECT/SWAPHANDS/TURN, `excludeCollection` of RECALL) with a string value; it replaces the named collection with square brackets so widget IDs can be typed directly.

Two prefilled values are the engine's own defaults but were rejected by `validator/validate_gamefile.js`, so a freshly inserted operation showed a red problem in edit mode. The validator now accepts them:

- `FLIP.face` may be `null` (the engine's default, meaning "flip via faceCycle") in addition to a number.
- `SWAPHANDS.source` and `TURN.source` accept `"all"` (the engine's default) next to a collection, like `SELECT.source` already did.

Two operations still show a problem right after insertion, deliberately: `AUDIO {source: ""}` reports *asset expected* and `CALL {routine: ""}` reports *routine name expected*. Unlike the values above, an empty asset or routine name is not a working default — it is something the author has to fill in — so the red marker is a prompt rather than noise, and having the key present is what surfaces the *upload audio file* button on AUDIO.

The command panel picked up two readability fixes on the way: its hint text uses a new `--textDimColor2` (the old `--textDimColor1` is `#ccc` on white in light mode, roughly 1.6:1), and the key caps got enough margin that their outline no longer draws over the neighbouring text.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2871/pr-test (or any other room on that server)


96LawDawg:
This would be a nice PR to add something to the JSON Editor to switch from traditional collections to anonymous collections (adding the square brackets).

96LawDawg:
It's going to take some time (and negotiation) to come up with the right "grammar" for this. It isn't just which parameters should be included, but also their order.  SELECT, for example, should not have "collection" second.

I think the general order should be like this:
  Function name
  Input collection/widget/source
  Mode/Relationship
  Miscellaneous parameters specific to the function
  Count/value
  Output collection/variable/result

Examples (some of these are a little different than what is currently in the PR):
```JSON
    {
      "func": "SELECT",
      "source": "all",
      "type": "all",
      "property": "parent",
      "value": null,
      "collection": "DEFAULT"
    },
    {
      "func": "GET",
      "collection": "DEFAULT",
      "property": "id",
      "variable": "result"
    },
    {
      "func": "ROTATE",
      "collection": "DEFAULT",
      "mode": "add",
      "angle": 90,
      "count": 1
    },
    {
      "func": "TURN",
      "source": "all",
      "turnCycle": "forward"
    },
    {
      "func": "CLONE",
      "source": "DEFAULT",
      "count": 1,
      "collection": "DEFAULT"
    }
```

VTT-AI-Agent:
I've addressed the two open points from the discussion:

**Parameter grammar** (per @96LawDawg's suggestion): the prefilled parameters now follow the proposed order — function name, input collection/widget/source, mode/relationship, operation-specific parameters, count/value, output collection/variable/result. Concretely this changed:
- `SELECT`: `source, type, property, value, collection`
- `CLONE`: `source, count, collection`
- `ROTATE`: now includes `mode: "add"` before `angle` and `count`
- `LABEL`: `collection, mode, value`
- `TURN`: `source, turnCycle` (matching the example)
- `CALL`: dropped the invalid `widget: "id"` prefill (a literal `"id"` is not a widget ID); it now only prefills `routine`

**Anonymous collections**: added a new editor command *convert to anonymous collection*. It appears when the cursor is on an input collection parameter (or `source` of CLONE/SELECT) whose value is a string, and replaces the named collection with square brackets, placing the cursor inside so widget IDs can be typed directly.

Both verified in the browser along with the existing Shift-click "insert without parameters" behavior.

VTT-AI-Agent (commented):
## TL;DR

Solid editor UX improvement — the prefill map, Shift-to-skip mechanics, and the anonymous-collection command all check out against the existing `jeCommands` patterns, and jest passes. Only minor findings: a likely Shift/uppercase-typing pitfall in Tab search, a stale `jeState.shift` read in the on-screen Tab keycap path, and a question about the `FLIP` prefill.

## Findings

1. **Tab search: holding Shift to type uppercase names triggers the no-params insert** — Operation names are displayed uppercase (`MOVE`, `SET`, …), so it's natural for a user to hold Shift while typing the filter. The filter itself is case-insensitive, but if they release Tab (or press it) while Shift is still down, `e.shiftKey` is true at execute time (`jsonedit.js:4070`) and they silently get the bare `{"func": ...}` insert instead of the prefilled one. The hint text helps, but expect confused reports. An option would be to ignore Shift when the filter contains uppercase letters typed while it was held, or to only honor Shift for the button-click path — at minimum worth keeping in mind if this comes up.

2. **`jsonedit.js:3645` — prefer `e.shiftKey` over `jeState.shift` in the Tab-keycap click handler** — The click event `e` is in scope and carries the authoritative modifier state. `jeState.shift` can go stale: the `keyup` handler returns early when `jeEnabled` is false (`jsonedit.js:4093`) before clearing it at `jsonedit.js:4099`, and no `keyup` arrives at all if Shift is released while the window is unfocused. A stale `true` would silently insert without parameters. One-token fix: `jeCommandClickEvent = { shiftKey: e.shiftKey };`.

3. **`jsonedit.js:3824-3827` — `jeCommandClickEvent` isn't cleared if the command throws** — If `jeCallCommand` rejects, the `jeCommandClickEvent = null` line is skipped and a stale event (possibly with `shiftKey: true`) survives until the next click. All current setters overwrite it before use, so this is latent rather than live, but a `try { ... } finally { jeCommandClickEvent = null; }` would make it robust against future call paths.

4. **`FLIP` prefill `face: 0` (`jsonedit.js:1332`) — intentional?** — With `face: 0` prefilled, an inserted-and-unedited FLIP always sets widgets to face 0 rather than flipping via `faceCycle` (the engine default is `face: null`). If "turn face down" is genuinely the most common library usage that's fine, but it's the one prefill that changes the operation's natural behavior rather than just naming its common parameters, so a user who only fills in `collection` may be surprised.

5. **Minor: parameter knowledge now lives in two places** — `jeRoutineOperationCommonParams` (`jsonedit.js:1323`) duplicates a subset of the per-op `defaults` passed to `jeAddRoutineOperationCommands` (`jsonedit.js:1434-1462`), so a future operation/param rename has to be made twice. Passing the common params as an extra argument at those call sites would keep each operation's editor metadata in one place. Purely a maintainability nit.

Everything else looks correct: the synchronous read of `jeCommandClickEvent` inside `jeRoutineCall` happens before the async boundaries, `event.isTrusted` correctly distinguishes real clicks from the synthetic tab-search `.click()`, `jeSetAndSelect([])` follows the established pattern, the anonymous-collection context regex correctly uses `source` for CLONE/SELECT (whose `collection` is an output), and all 27 map keys match registered operations. No save-format or engine impact; jest passes.

The `beta` label was already present, and since this PR only touches `client/js/jsonedit.js` (not engine files) it remains eligible, so I left it in place.

VTT-AI-Agent:
Thanks for the review — addressed in 98d610dc:

1. **Shift while typing uppercase filter text**: the Tab search now tracks whether any filter character was typed with Shift held (`jeTabSearchShiftTyped`); if so, Shift is ignored at execute time. Typing `SHUF` in uppercase and releasing Tab with Shift still down now inserts *with* parameters, while typing lowercase and pressing Shift only at release still skips them. Verified both cases in the browser.
2. **Stale `jeState.shift` in the Tab-keycap click handler**: now reads `e.shiftKey` from the click event (combined with the same `jeTabSearchShiftTyped` guard).
3. **`jeCommandClickEvent` not cleared on exception**: `clickButton` now clears it in a `finally` block.
4. **`FLIP` prefill**: changed `face: 0` to `face: null` (the engine default), so an inserted-and-unedited FLIP keeps the natural faceCycle flip behavior instead of forcing widgets to face 0. It still surfaces the `face` knob for editing.
5. **Param-map duplication**: leaving as is. The prefill map is a deliberately curated, ordered list (the "grammar" from the discussion) that is meant to be reviewed and tweaked independently of the exhaustive per-op `defaults` at the `jeAddRoutineOperationCommands` call sites; merging both objects into those already-long lines would make the curated order much harder to read and review.

VTT-AI-Agent (commented):
## TL;DR

All findings from my previous review are correctly addressed in 98d610dc — this now looks good to merge from my side. No new issues found.

## Verification of the fixes

I checked each point against the current code (jest: 6/6 passing):

1. **Shift while typing uppercase filter text** — `jeTabSearchShiftTyped` is set when a filter character is typed with Shift held (`client/js/jsonedit.js:4024`), reset at both Tab-search activation sites (`jsonedit.js:3636`, `jsonedit.js:3989`), and correctly guards both execute paths (`jsonedit.js:3647`, `jsonedit.js:4078`). Uppercase typists now get the prefilled insert as expected.
2. **Tab-keycap click handler** — now reads `e.shiftKey` from the click event instead of the potentially stale `jeState.shift` (`jsonedit.js:3647`).
3. **`jeCommandClickEvent` cleanup** — `clickButton` clears it in a `finally` block (`jsonedit.js:3828-3832`), so a throwing command can no longer leave a stale Shift state behind.
4. **`FLIP` prefill** — changed to `face: null` (`jsonedit.js:1333`), preserving the natural faceCycle flip while still surfacing the knob.
5. **Param-map duplication** — declined with a fair rationale (the prefill map is a deliberately curated, ordered "grammar" meant to be tweaked independently of the exhaustive `defaults`); no objection.

The blur-triggered execute path (`jsonedit.js:4051`) clicks without setting `jeCommandClickEvent`, but since the `finally` in `clickButton` guarantees it's `null` there, that path safely defaults to inserting *with* parameters — the right fallback.

The `beta` label was already present, and since this PR only touches `client/js/jsonedit.js` (not engine files) it remains eligible, so I left it in place.

VTT-AI-Agent (commented):
## TL;DR

The Shift mechanics, the prefill plumbing and the new *convert to anonymous collection* command all hold up on re-read, and my earlier findings stay fixed. One new, concrete problem: three of the prefilled values (`FLIP.face: null`, `SWAPHANDS.source: 'all'`, `TURN.source: 'all'`) are rejected by `validator/validate_gamefile.js`, so every freshly inserted FLIP/SWAPHANDS/TURN now shows a red problem in the editor's Problems panel even though the values are correct at runtime.

## Findings

### 1. Three prefills produce immediate false-positive validator errors in edit mode

`client/js/editor/sidebar/json.js:180` runs `validateGameFile()` on every edit, and `validate_gamefile.js` is bundled into the editor (`server/minify.mjs:137`), so these show up live the moment the operation is inserted. I built a game file containing exactly the 29 prefilled operations from `jeRoutineOperationCommonParams` and ran `node validator/validate_gamefile_node.js` on it:

```
b1[clickRoutine.8.face]:   number expected                  # FLIP  {face: null}
b1[clickRoutine.24.source]: Collection 'all' is undefined.  # SWAPHANDS {source: 'all'}
b1[clickRoutine.26.source]: Collection 'all' is undefined.  # TURN  {source: 'all'}
```

All three values are *correct* — they are literally the engine defaults:
- `client/js/widgets/widget.js:1288` — `setDefaults(a, { count: 'all', face: null, ... })`
- `client/js/widgets/widget.js:1952` / `:2073` — `setDefaults(a, { ..., source: 'all' })`, and both implement `a.source == 'all' ? allSeats : collections[...]`.

The validator is simply narrower than the engine (`'face': 'number'` at `validate_gamefile.js:659`; `'source': 'inCollection'` for SWAPHANDS/TURN at `:787`/`:799`, where `inCollection` has no `'all'` escape hatch — unlike `SELECT.source`, which is `v => v === 'all' || typeof v === 'string'`). That gap is pre-existing, but it was invisible until this PR started emitting those values by default. Two ways out, either is fine:

- widen the validator (`'face': v => v === null || typeof v === 'number'`, and give SWAPHANDS/TURN the same `v === 'all' || inCollection` shape as `SELECT.source`), or
- drop `face` from the FLIP prefill and `source` from the SWAPHANDS/TURN prefills (both are the defaults anyway, so nothing is lost except the "here's the knob" hint).

Note this is the direct consequence of the `face: 0` → `face: null` change I asked for last round — sorry for the round trip; widening the validator is probably the better resolution since `face: null` really is legal.

The `Collection 'DEFAULT' is undefined.` messages that also appear for `collection: 'DEFAULT'` prefills inserted before any SELECT/FOREACH are *not* a regression — `requiredInputCollections` (`validate_gamefile.js:390`) already flagged the bare `{func: 'SET'}` insert with `no input given and collection DEFAULT is undefined`. Only the wording changes.

### 2. `CALL: { routine: 'clickRoutine' }` prefills an infinite recursion

`jsonedit.js:1318`. Insert CALL into a button's `clickRoutine`, don't edit it, click the button: `a.widget` defaults to the current widget, so it calls its own `clickRoutine`, which calls CALL again. `evaluateRoutine` threads `depth` (`widget.js:810`, `:1090`) but never caps it — there is no recursion guard anywhere in the CALL path — so this hangs the tab. Every other prefill is inert until edited; this one is armed. A neutral placeholder that can't self-reference (e.g. `'myRoutine'`, which still satisfies both the `/Routine$/` engine check at `widget.js:1076` and the `routineProperty` validator) would keep the hint without the foot-gun.

### 3. `je_anonymousCollection` misses two input-collection parameters

`jsonedit.js:542`. The context regex covers `collection` on 14 ops plus `source` on CLONE/SELECT, which is the right split (SELECT/CLONE `collection` is an output — correctly excluded). But `SWAPHANDS.source` and `TURN.source` are also *input* collections (`widget.js:1959`, `:2081`) and don't get the button, so converting them to `['seat1','seat2']` still has to be typed by hand. `RECALL.excludeCollection` is arguably in the same family. Cheap to add to the second alternation.

## Re-verification of the previous round

All five points from my earlier review remain correctly addressed in `98d610dc`, unchanged by the `a4067653` merge: `jeTabSearchShiftTyped` guards both execute paths (`jsonedit.js:3801`, `:4254`) and is reset at both activation sites (`:3790`, `:4165`); the Tab-keycap handler reads `e.shiftKey`; `clickButton` clears `jeCommandClickEvent` in a `finally` (`:3987`); `event.isTrusted` correctly keeps the synthetic `.click()` from clobbering the manually-set flag; and `jeSetAndSelect([])` hits the `emptyBrackets` branch so the cursor lands between the brackets as intended. I could not run jest in this sandbox (no `node_modules`), but nothing here touches server or validator code paths that jest covers.

No new widget property or routine function is exposed to game authors here — this is JSON-editor UX only — so the widget/property-editor UI and tutorial rules don't apply.

I left the `beta` label in place (it was already there) because this PR touches only `client/js/jsonedit.js`, no engine file, and the findings above are editor-noise-level rather than anything that could corrupt a game or crash the server.

VTT-AI-Agent:
Thanks — all three findings addressed in 8859304a.

**1. False-positive validator errors on fresh FLIP/SWAPHANDS/TURN inserts** — fixed by widening the validator, since the values really are legal:
- `FLIP.face` is now `numberOrNull` (`v === null || typeof v === 'number'`) instead of `number`.
- `SWAPHANDS.source` and `TURN.source` are now `collectionOrAll` (`v === 'all' || inCollection(...)`), a new validator that gives them the same `'all'` escape hatch `SELECT.source` already had, while keeping the widget-ID/anonymous-collection diagnostics of `inCollection` — both operations resolve their source through `getCollection()`, so an anonymous collection is valid there too.

Verified by re-running the validator over a game file containing all 29 prefilled operations (no `face`/`source` problems left) and in the browser: the editor's validation panel reports *no* problems for a routine using `FLIP {face: null}`, `SWAPHANDS {source: "all"}` and `TURN {source: "all"}`.

One prefill still produces a problem on insert and I left it deliberately: `AUDIO {source: ""}` reports "asset expected". Unlike the three above, an empty source is not a working default — it's a value the author *has* to fill in — and having the key present is what surfaces the *upload audio file* button, so the message is a useful prompt rather than noise.

**2. `CALL: {routine: 'clickRoutine'}` self-recursion** — changed to `routine: 'myRoutine'` (the placeholder the validator's own error message suggests). Verified in the browser: clicking a button whose unedited `CALL` insert is the whole routine now logs `Widget b2 does not contain myRoutine (or it is no array).` and the tab stays responsive, instead of hanging. The missing depth guard in `evaluateRoutine` is a real engine gap, but fixing it belongs in an engine PR, not this editor one.

**3. `je_anonymousCollection` missing input collections** — the context regex now also covers `SWAPHANDS`/`TURN` `source` and `RECALL` `excludeCollection`. Verified in the browser for all four positive cases (SWAPHANDS/TURN `source`, RECALL `excludeCollection`, plus FLIP `collection` as a control) — each converts to `[]` with the cursor between the brackets — and that the button stays hidden on `SELECT.collection` (an output) and on non-collection parameters such as `SWAPHANDS.direction`.

`npm test` passes (6/6). The PR now touches `validator/validate_gamefile.js` as well, so I added the `Validator` label and described the validator change in the PR body.

VTT-AI-Agent (commented):
## TL;DR
UI/UX review (not a code review). The prefill itself is a genuine usability win — inserting an operation now gives you a working skeleton instead of a bare `{"func": …}`. The one thing I'd change: the *only* place the new **Shift = no parameters** behaviour is explained is a hint line that is (a) rendered at roughly 1.6:1 contrast, (b) visually broken — the `Shift` key-chip overlaps the colon and the line wraps at *every* viewport size — and (c) shown even in contexts where no operation can be inserted. Replace the modifier-only affordance with a small persistent toggle in the command panel, and fix that hint line.

**What I exercised.** Ran the branch locally, edit mode → JSON editor on a `button` widget with a `clickRoutine`. Viewports: **390x844 (phone), 820x1180 (tablet), 1280x800 (laptop), 1920x1080 (desktop)**. Overlays opened at all four sizes: game shelf / public library, players overlay, add-widget overlay, edit-mode toolbar + sidebar, and the JSON editor panel including the **Tab-search overlay** and the context-command list. I did **not** reach the chat, share/publish, help, or state-save sub-dialogs — the diff touches only `client/js/jsonedit.js` and `validator/`, no shared CSS, and none of the overlays I did open showed any change.

Insertion works as advertised: [prefilled SELECT](https://agent.virtualtabletop.io/reports/ui-review/1536646306309738516/g-after-insert-SELECT.png), and Shift+click gives the bare `{"func": "MOVE"}`.

---

## Big layout improvements

### 1. Make "insert without parameters" a visible control, not an invisible modifier
Right now the no-parameters path is discoverable in exactly two ways: a `title=` tooltip that needs a ~1s hover (`jsonedit.js:3670`), and one line inside the Tab-search hint (`jsonedit.js:3556`). A mouse user who clicks buttons and never presses Tab will never learn it exists; on a touch device there is no hover *and* no Shift key, so the behaviour is **completely unreachable** — and the editor is usable at 820x1180, where I did open and use this panel.

Concrete proposal that fits the panel's existing style: put a small persistent control in the header row that already reads *"Press or hold `Tab` to search"* — e.g. a checkbox/pill **`☑ prefill parameters`**. It is self-documenting, works for mouse, touch and keyboard, is remembered across insertions (right now you must hold Shift every single time), and lets you drop the wrapped hint line entirely. Shift can stay as the power-user override.

### 2. Reuse the editor's existing "this is a default" dimming for prefilled defaults
Most prefilled values *are* the engine defaults (`FLIP.face: null`, `SWAPHANDS.source: "all"`, `SELECT.source/type: "all"`, `collection: "DEFAULT"`, `MOVE/CLICK/ROTATE.count: 1`). Once you insert a handful of operations the routine gets long fast — [9 operations, ~45 lines](https://agent.virtualtabletop.io/reports/ui-review/1536646306309738516/o-many-ops.png) where previously it was 9 — and nothing distinguishes "I set this deliberately" from "the editor filled in the default".

The editor already has a visual language for exactly this: widget lines whose value equals the widget default get `<i class=default>` and render dim grey (`jsonedit.js:2408`, `jsonedit.css:370` — visible as the grey `"movable": false` / `"movableInEdit": true` lines in every screenshot here). That check is gated on top-level `'  "'` indentation, so routine params never get it. Extending it to routine-operation params whose value equals the operation's default would make the newly inserted noise read as "safe to ignore or delete" using a convention users already know — no new UI vocabulary.

---

### 3. The new hint line collides with its own key-chip and wraps at every viewport
`jsonedit.js:3556` appends `<span class="key">Shift</span>: insert without parameters`. The `.key` chip draws a `::after` outline with negative insets (`jsonedit.css:731-733`: `left/right: -0.3125rem; top: -3px; bottom: -9px`), so the immediately-adjacent `:` is **drawn over and hidden**, and the chip's outline overlaps the *"Press `Tab` to execute"* line above it. The command panel is a fixed narrow column, so the phrase wraps to two lines at 1920x1080 just as it does at 390x844.

| | |
|---|---|
| [1280x800](https://agent.virtualtabletop.io/reports/ui-review/1536646306309738516/i-tabsearch.png) | [1920x1080](https://agent.virtualtabletop.io/reports/ui-review/1536646306309738516/vp-desktop-1920x1080-tabhint.png) |
| [390x844](https://agent.virtualtabletop.io/reports/ui-review/1536646306309738516/vp-phone-390x844-tabhint.png) | [with filter typed](https://agent.virtualtabletop.io/reports/ui-review/1536646306309738516/j-tabsearch-filtered.png) |

Minimal fix: drop the colon and phrase it like its siblings (which are all verb phrases) — `Hold <span class="key">Shift</span> for no parameters` — and give `#jeCommands .key` a `margin-right: 4px` so no chip can ever sit flush against punctuation. On phone this block now eats ~90px of a ~340px-tall panel; folding it into the existing line (`Press Tab to execute · Shift: no params`) would help further.

### 4. The Shift hint shows in contexts where nothing can be inserted
The hint is unconditional, so pressing Tab with the cursor on `"width": 200` shows *"Shift: insert without parameters"* above a list of `$()` / `decrement number` / `double number` — [screenshot](https://agent.virtualtabletop.io/reports/ui-review/1536646306309738516/n-tabhint-nonroutine.png). Gate it on the filtered command list actually containing a `class: 'operation'` command.

### 5. `convert to anonymous collection` is an action styled and sorted as a value
In the collection context it lands alphabetically **in the middle** of the collection-*name* buttons: `activeSeats · child · convert to anonymous collection · DEFAULT · playerSeats · thisButton · widget` ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1536646306309738516/l-collection-context.png)). Every other entry in that group inserts a name; this one performs a transformation, yet it is the same blue pill, and it is the only two-line button in the group, which breaks the list rhythm. Suggest either a leading `::before` symbol (the panel already uses `'VTT-Symbols'` glyphs for such commands, `jsonedit.css:782`) or grouping it with the other transform commands rather than sorting it into the value list.

### 6. Only *some* buttons in a visually uniform list have a tooltip
`title` is added only for `command.class == 'operation'` (`jsonedit.js:3670`), so `AUDIO` has a hover tooltip but its immediate neighbours `Comment`, `Expression: increment`, `Expression: random` — identical pills, same list — have none. Inconsistent hover affordance in a single row of controls. Also, the tooltip text redundantly repeats the visible label (`SELECT (Shift+click: …)`); `Shift+click: insert without parameters` alone would read better.

### 7. No inverse of the anonymous-collection command
After converting, the value is `[]` ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1536646306309738516/m-after-anon-collection.png)) and the command disappears. There is no "back to named collection" counterpart, so the action reads one-way. The named-collection buttons still work as an escape hatch, but a symmetric command (shown when the value is an array) would make the pair feel intentional.

---

### 8. The one explanation of the new feature is drawn at ~1.6:1 contrast
The hint block uses `--textDimColor1`, which in light mode is `#ccc` on a white panel (`client/css/editor/layout.css:15`) — roughly **1.6:1**, far below any legibility threshold; in the screenshots *"to select"*, *"to execute"* and *"insert without parameters"* are barely visible. That was tolerable when the block only restated keys you'd already pressed; it is not tolerable now that it carries the only documentation of a behaviour. Please darken the hint text in light mode (something around `#6b6b6b` still reads as "dim" against `--textColor: black` and matches the dark-mode `#606060`).

### 9. "anonymous collection" is jargon, and the outcome isn't shown
A newcomer has no model for "anonymous collection"; the result is `"collection": []`, an empty array, with no hint that they are now expected to type widget IDs into it. Suggest wording closer to the outcome — e.g. **`use widget IDs instead ([ ])`** — plus a `title` explaining "replaces the named collection with `[]` so you can list widget IDs directly". As a bonus, after the conversion the caret could sit inside the brackets so the next keystroke goes where the user needs it.

### 10. "parameters" vs. what the user sees
The hint and tooltip both say *"insert without parameters"*, but nothing in the editor calls these things parameters — the JSON shows `"collection"`, `"property"`, `"value"`, and the docs call them properties. *"insert without properties"* (or *"insert empty"*) would match what's on screen.

### 11. Shift-typed filtering silently disables the advertised Shift behaviour
If you type an uppercase letter while filtering (Tab → `S` → …), `jeTabSearchShiftTyped` latches for the rest of the search session, so holding Shift on execute no longer suppresses the parameters — while the hint on screen still promises it does. Nothing in the UI reflects that the shortcut has been switched off. If the latch has to stay, the hint line should grey out / disappear once it trips, so the on-screen promise and the actual behaviour agree.

### 12. `CALL: { routine: 'myRoutine' }` is the only invented placeholder
Every other prefill is either an engine default or a neutral value; `"myRoutine"` is a made-up name that looks like a real reference. A first-time user may well leave it and wonder why nothing happens. Consider `null` or `""` so the empty slot is visually obvious (and, ideally, gets selected on insert the way `func` is).

VTT-AI-Agent:
Thanks — the UI/UX review is addressed in e44f03c7. The headline change: **the Shift modifier is gone**, replaced by a persistent *prefill properties* toggle, which also removes findings 3, 4, 6 and 11 at the root.

| | |
|---|---|
| [toggle on](https://agent.virtualtabletop.io/reports/pr/1525221797056745485/2871-ui/toggle-on.png) | [toggle off](https://agent.virtualtabletop.io/reports/pr/1525221797056745485/2871-ui/toggle-off.png) |
| [Tab search hint](https://agent.virtualtabletop.io/reports/pr/1525221797056745485/2871-ui/tabsearch-hint.png) | [collection context](https://agent.virtualtabletop.io/reports/pr/1525221797056745485/2871-ui/collection-context.png) |
| [dark mode](https://agent.virtualtabletop.io/reports/pr/1525221797056745485/2871-ui/dark-mode.png) | [phone 390x844](https://agent.virtualtabletop.io/reports/pr/1525221797056745485/2871-ui/phone-390x844.png) |

**1. Visible control instead of an invisible modifier** — done, and I dropped Shift entirely rather than keeping it as an override. The toggle is a pill right under the Tab hint, labelled `☑ prefill properties` / `☐ prefill properties`, shown only when the cursor is somewhere an operation can actually be inserted (`jeCanInsertOperation()`). It works with mouse, touch and while the Tab search is open (a `mousedown` preventDefault keeps it from blurring the editor and firing the search), and it is remembered in `localStorage`. Keeping Shift as a power-user override would have meant keeping the `jeTabSearchShiftTyped` latch — i.e. finding 11 — for a shortcut that saves one click, so the whole `jeCommandClickEvent` plumbing went away with it. Verified in the browser: default on → prefilled insert, one click → `{"func": "MOVE"}`, survives a page reload, hidden on non-routine lines, and the Tab-search path honours it too.

**2. Dim prefilled values that equal the engine defaults** — not done, and I'd rather not fake it. `<i class=default>` means "this line is deletable without changing behaviour", and the editor has no authoritative table of routine-operation defaults to check that against: the per-operation `defaults` maps at the `jeAddRoutineOperationCommands` call sites are *"insert this value when you add the property"* lists, and several of them differ from what the engine actually defaults to (`GET.variable` is `id` there but `result` in the engine, `DELAY.milliseconds` is `0` vs `500`, `SELECT.max` is `999999`). Dimming from that map would tell people a line is safe to delete when deleting it changes what the routine does. Doing it properly means extracting the `setDefaults()` values out of `widget.js` into something the editor can read — worth doing, but as its own change; it would then dim *all* routine lines that are at their default, not just freshly inserted ones. With the toggle, anyone who finds the extra lines noisy can now switch them off for good.

**3. Hint line collides with its own key chip and wraps** — the `Shift: insert without parameters` line is gone. The two remaining lines got breathing room: `#jeCommands .key` now has `0.4rem` side margins so the `::after` outline can no longer overlap adjacent characters, and the hint block is a `.jeHint` class with `line-height: 1.9` so a chip's outline no longer reaches into the line above. The block is now one line shorter than before this PR touched it.

**4. Hint shown where nothing can be inserted** — the prefill toggle is gated on the cursor being in a routine, so it is absent on `"width": 200`. The remaining hint lines describe the Tab search itself, which does work everywhere, so those stay unconditional.

**5. `convert to anonymous collection` sorted into the list of collection names** — renamed to `[] anonymous collection`. The leading bracket sorts it in front of `activeSeats`/`DEFAULT`/… instead of into the middle of them, it says what you get, and it now fits on one line, so the list rhythm holds.

**6. Tooltips on only some buttons** — the operation tooltip is gone; its entire content was the Shift hint plus a repeat of the label. Tooltips now come from an optional `description` on a command and are rendered only where they add something the button text does not — right now that is the anonymous-collection command.

**7. No inverse of the anonymous-collection command** — the inverse is already in the same group: the `DEFAULT` / `all` / `thisButton` / `playerSeats` / … buttons stay visible when the value is an array and replace it with a named collection. A separate "back to named collection" command would have to insert one of those names anyway, so it would just duplicate a button that is already sitting two rows below.

**8. ~1.6:1 contrast on the hint** — fixed for this panel: the hint uses a new `--textDimColor2` (`#6b6b6b` light, `#8a8a8a` dark) instead of `--textDimColor1`. I did not change `--textDimColor1` itself, even though it is genuinely too light in light mode — it also colours the editor's line numbers, the dimmed default widget lines and the log entries, so changing it repaints a good part of edit mode and belongs in its own PR with before/after shots rather than riding along here.

**9. "anonymous collection" is jargon and the outcome isn't shown** — the button now literally shows the outcome (`[] anonymous collection`) and has the tooltip *"Replace the named collection with [] so you can list widget IDs directly"*. I kept the term itself because that is what the wiki and the request in this thread call it. The caret already lands between the brackets after the conversion.

**10. "parameters" vs "properties"** — renamed throughout: the toggle says *prefill properties*, and the map is `jeRoutineOperationCommonProperties`.

**11. Shift-typed filtering silently disabled the advertised behaviour** — gone with the modifier; there is no hidden state left to contradict what is on screen.

**12. `CALL: { routine: "myRoutine" }`** — now `{ routine: "" }`. The empty slot is visibly empty, it cannot be mistaken for a real reference, and the validator immediately says *routine name expected (format: myRoutine)*, which is a better teacher than a plausible-looking placeholder. That matches how `AUDIO: { source: "" }` behaves.

`npm test` passes (6/6) and the TestCafe `editor.js` suite passes locally against this branch.

VTT-AI-Agent (commented):
## TL;DR

The current head looks good: the earlier validator mismatches, unsafe CALL placeholder, anonymous-collection coverage gaps, and Shift-only UX have all been resolved, and I found no new actionable issues.

## Findings

No actionable findings at `e44f03c7`.

- This PR adds editor affordances rather than a new widget property or routine operation, so the visual routine-block UI and tutorial requirements are not triggered. The new behavior is exposed directly in the real edit-mode JSON editor through the persistent **prefill properties** toggle and the contextual `[] anonymous collection` control.
- The validator changes now match the existing runtime semantics without weakening unrelated checks: `FLIP.face` accepts the engine's `null` default, while `SWAPHANDS.source` and `TURN.source` accept `"all"` and still pass other values through the existing collection validator. A targeted `validateGameFile` check covering those three values returned no problems.
- `node --check` passed for both changed JavaScript files. I could not run Jest because this clone does not contain `node_modules/.bin/jest`; the test command failed before executing any tests.

The `beta` label was already present; this PR touches no engine paths and the editor/validator changes are safe and useful for player testing, so I left it in place.

VTT-AI-Agent (commented):
## TL;DR

The rework since my last pass is good: the Shift modifier is gone in favour of a persistent, discoverable **prefill properties** toggle, and the three validator false-positives I flagged (`FLIP.face: null`, `SWAPHANDS/TURN.source: "all"`) are now fixed at the validator, matching the engine. One real new finding: the `[] anonymous collection` context regex isn't anchored, so it also fires when the cursor is on an *element* of an already-anonymous collection and turns `["a","b"]` into `[[],"b"]`.

## Findings

### 1. `je_anonymousCollection` also fires inside an already-anonymous collection, producing a nested `[[], …]`

`client/js/jsonedit.js:544`. The context alternatives are not end-anchored (`… ↦ collection`, no `$` and no `( ↦ |$)` guard, unlike e.g. `'^.* ↦ icon( ↦ |$)'` at `:428` or `'… ↦ image$'` at `:365`), so a context that *continues past* `collection` still matches. For an array value the continuation is the array index, and `show()` doesn't exclude it: `jeGetValue()` stops at the array (the string element isn't an object) and `jeGetLastKey()` returns the numeric index, so `typeof value[key] == 'string'` is true for `"a"`.

Repro: `{"func":"SET","collection":["a","b"],"property":"x"}`, put the cursor on the `"a"` line. I reproduced the context computation mechanically against `jeGetContext` (`jsonedit.js:2782-2827`):

```
context: button ↦ clickRoutine ↦ 0 ↦ (SET) ↦ collection ↦ 0
regex matches: true
lastKey: 0   value[lastKey]: "a"   => show: true
```

Clicking the button then does `array[0] = '###SELECT ME###'` → `"collection": [[], "b"]`, which is not a legal collection (and the user's `"a"` is gone). It's the one context where the command is guaranteed to be wrong — the value is *already* an anonymous collection.

Cheapest fix that doesn't disturb the regex: make `show` require a string key, e.g.

```js
show: function() {
  const key = jeGetLastKey();
  const value = jeGetValue();
  return typeof key == 'string' && value && typeof value[key] == 'string';
}
```

(End-anchoring each alternative would also work, but the trailing `"…"` select token that `jeGetContext` pushes makes `$` a bit more fragile here.)

### 2. `AUDIO` and `CALL` still insert a value the validator rejects

The PR fixed exactly this class of problem for FLIP/SWAPHANDS/TURN, but two prefills are still flagged the moment you insert them. I generated a game file containing each of the 29 prefilled operations (after a SELECT so `DEFAULT` is defined) and ran `node validator/validate_gamefile_node.js` on each:

```
AUDIO  [source]:  asset expected (format: /assets/1_1, /i/icon.png or http://example.com/image.png)
CALL   [routine]: routine name expected (format: myRoutine)
```

Every other prefilled operation is clean. This is defensible — `""` is genuinely incomplete and the red marker says "fill this in" — but it is inconsistent with the rest of the map (and with the PR description's premise that a fresh insert shouldn't show a red problem). Either is fine, it just deserves to be a deliberate call: keep `""` as a "must fill in" marker, or drop those two keys the way FLIP's `face` could have been dropped. Worth a sentence in the PR description if you keep them, since it's the first thing a user hits with AUDIO/CALL.

### 3. `MOVE: { count: 1 }` overrides a context-sensitive engine default

`jsonedit.js:1332` vs `client/js/widgets/widget.js:1536`: the engine default is `count: a.from ? 1 : 'all'`. The prefill hard-codes `1`, which matches the default only while `from` is set. A user who deletes the `from`/`to` placeholders and moves a collection instead now silently moves a single widget where the bare operation would have moved all of them. Prefilling `count` is still the right call for the `from` case; just flagging that this is the one entry (besides `GET.variable: 'result'`, which is harmless) whose value isn't the default in every configuration.

### 4. Nit: the function-valued `description` branch is dead

`jsonedit.js:3676` handles `typeof command.description == 'function'`, but `description` exists on exactly one command (`:543`) and is a plain string. Harmless, just unused generality.

## Verified as correct

- **Validator widening.** `collectionOrAll` (`validate_gamefile.js:41`) delegates to `inCollection` with the same `(v, p, propertyPath)` signature, so the array-of-problems return path is preserved; it is strictly narrower than `SELECT.source`'s `v === 'all' || typeof v === 'string'`, which is the right direction. `numberOrNull` matches `setDefaults(a, { face: null, … })` at `widget.js:1288`, and `source: 'all'` matches `widget.js:1952`/`:2073`.
- **Every prefilled value checked against the engine's own `setDefaults`** for CANVAS, CLICK, COUNT, LABEL, SORT, TIMER, SCORE, RESET, UPLOAD, MOVEXY, RECALL — all match. The `null` placeholders (`MOVE.from/to`, `RECALL.holder`, `SET/SELECT.value`, `IF.operand1/2`) go through `isValidID`/validator paths that already report "fill me in" rather than misbehaving.
- **Toggle plumbing.** `jeCanInsertOperation` (`:1350`) guards on `jeStateNow`/`jeJSONerror` before calling `show()`, and the operation commands' shared `show` (`jeRoutineCall(…, true)`) is safe to call with no arguments. The `mousedown` `preventDefault` correctly stops the blur-triggered Tab-search execute, the `:not(.jeToggle)` selector keeps `clickButton` from double-handling, `localStorage` round-trips via `!= 'false'` (defaulting to on), and since `jeShowCommands` rebuilds `innerHTML` each call, `on()` (which binds directly, `domhelpers.js:518`) can't accumulate listeners. All the previous round's Shift machinery (`jeTabSearchShiftTyped`, `jeCommandClickEvent`) is fully removed — no dead references left.
- **CSS.** `#jeCommands` is `display: flex; flex-direction: column` (`jsonedit.css:396`), so `align-self: flex-start` behaves; the pill sits outside `#jeContextButtons`, whose `> button` rules would otherwise out-specify the ID selector. `--textDimColor2` (`#6b6b6b` on white) resolves the ~1.6:1 contrast problem from the UI review.

No new widget property or routine function is exposed to game authors here — this is JSON-editor UX plus a validator relaxation — so the widget/property-editor UI rule and the `library/tutorials/` rule don't apply. (Jest couldn't run in this sandbox: `jest-environment-jsdom` isn't installed. The changed files are client-editor/validator only, which the two existing jest tests don't cover.)

The `beta` label was already present and I left it in place: the PR touches no engine file (`client/js/widgets/`, `compute.js`, `serverstate.js` are all untouched), and the findings above are editor-level annoyances that can't corrupt a game or crash the server.

VTT-AI-Agent:
Thanks — finding 1 and 4 are fixed in f2fc1610, findings 2 and 3 are deliberate and now documented.

**1. `[] anonymous collection` firing inside an already anonymous collection** — real bug, fixed exactly as suggested: `show` now also requires the last context key to be a string, so an array index (i.e. the cursor being on an element of an already anonymous collection) no longer qualifies. I left the regex alone for the reason you gave.

Verified in the browser on a routine containing `{"func":"SET","collection":["a","b"],…}`, `{"func":"FLIP","collection":"DEFAULT",…}` and `{"func":"SWAPHANDS","source":"all",…}`:

| cursor on | button shown |
|---|---|
| `"a"` (element of the anonymous collection) | no *(was yes — the bug)* |
| the `"collection": [` line itself | no |
| `FLIP` `"collection"` key / `"DEFAULT"` value | yes |
| `SWAPHANDS` `"source"` | yes |
| `SWAPHANDS` `"direction"` | no |

…and clicking it on `FLIP.collection` still converts to `"collection": []` with the caret between the brackets.

**2. `AUDIO.source: ""` / `CALL.routine: ""` still flagged** — deliberate, and I added a paragraph to the PR description saying so. The FLIP/SWAPHANDS/TURN cases were values that *are* the engine default, so the error was plain wrong. `""` is different: it is not a working value, it is a slot the author must fill, so "asset expected" / "routine name expected" is a correct and useful prompt — and for AUDIO, having the `source` key present is what makes the *upload audio file* command appear. Dropping the keys would hide both.

**3. `MOVE.count: 1` vs `a.from ? 1 : 'all'`** — correct observation, keeping it. The prefill inserts `from`/`to` as the shape it is teaching (`move N widgets from A to B`), and for that shape `1` *is* the engine default; `count` is also one of the most-used MOVE properties in the library. A user who deletes `from`/`to` and switches to the collection form is editing the operation anyway, and the `count` line is right there and deletable. Making the prefill context-sensitive would mean the map stops being a static "these are the common properties" table for a case that only arises after the user has already rewritten the operation.

**4. Dead function-valued `description` branch** — removed; `description` is now read as a plain string.

`npm test` passes (6/6) and the TestCafe `editor.js` suite passes locally against this branch.

PR #2872, opened 2026-02-08
added file monitoring for live updates

ArnoldSmith86:
This PR adds a new editor panel called Files. You select a folder to monitor, it lists the files and for each one you can select what should happen when the file changes.

So far I only tested auto-replacing image assets when a file changes (see video).

[cinnamon-2026-02-08T010309+0100.webm](https://github.com/user-attachments/assets/1b96b964-fc42-4e34-b322-e23bd9e50fcb)

HTML files also seem to work already:

image

CSVs for decks seem to work aswell:

image

TXTs as lists:

image

Edit your favorite widgets separately:

image

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2872/pr-test (or any other room on that server)


96LawDawg:
What is the timing on this with #2849? Seems this file manager would work better with the names.

VTT-AI-Agent:
Pushed 56a137e addressing what was still open here:

- **Fixed the failing Production Environment check**: the production build minifies with uglify-es, which cannot parse `for await` (and optional chaining), so the server crashed on startup with `minifyJavascript: true`. `readDirectory` now walks the async iterator manually. Verified locally: the server starts with production settings and the Files panel loads and renders in the minified editor bundle without console errors.
- **Added a Refresh button** next to "Choose directory…", as suggested by @96LawDawg — it re-scans the monitored directory (or the dropped folder) without having to reselect it.
- **Removed the debug logging** from the "added debug logging" commit (the asset-picker console spam and the `DEBUG` blocks in `runHandlerForFile`/`imageReplaceAsset`).

VTT-AI-Agent (commented):
## TL;DR

Nice, self-contained editor feature with a clean handler registry. I found one real logic bug (the CSV "Replace" mode never removes cards of dropped card types — a transcription error from the jsonedit importer), a robustness issue (a handler throwing mid-batch permanently unbalances `batchStart`, silently stopping all further deltas from that client), and a few smaller items (dropdown ignored after first asset replacement, unescaped `innerHTML` in the log, dead code).

## Findings

1. **`applyCsvToDeck` "Replace" mode never deletes cards of removed card types** — `client/js/editor/sidebar/fileHandlers.js` (in `applyCsvToDeck`):
   ```js
   const jeStateNow = { ...deck.state, cardTypes: mode === 'set' ? {} : { ...deck.get('cardTypes') } };
   const oldCardTypeIDs = Object.keys(jeStateNow.cardTypes);
   ```
   In `set` mode, `cardTypes` is already cleared when `oldCardTypeIDs` is computed, so it's always `[]`; in `add` mode all old IDs survive the merge, so `!jeStateNow.cardTypes[oldID]` is never true. Either way the removal loop is dead code. The original in `client/js/jsonedit.js:756-759` captures `oldCardTypeIDs` *before* clearing. Result: importing a CSV in Replace mode leaves orphan cards whose `cardType` no longer exists in the deck.

2. **A handler throwing mid-batch permanently breaks delta sending for the client.** `batchStart()`/`batchEnd()` just increment/decrement `batchDepth` (`client/js/serverstate.js:303-310`). Several handlers (`applyCsvToDeck`, `jsonMultipleWidgets`, `jsonOneWidget`, …) call `batchStart()` and then `await` operations that can throw (`removeWidgetLocal`, `moveToHolder`, `w.set`). `runHandlerForFile` catches the error and shows an alert, but `batchEnd()` is never reached, so `batchDepth` stays positive and every subsequent edit from that client is silently swallowed until reload. Wrap the batched sections in `try { … } finally { batchEnd(); }`.

3. **`imageReplaceAsset`: the "Asset to replace" dropdown is ignored after the first run.** `let oldAsset = lastAssetUrl || assetOpt;` — once a replacement has happened, the stored `lastAssetUrl` always wins as long as it's still in use, so picking a *different* asset in the dropdown afterwards has no effect (the fallback only triggers when the last uploaded asset has zero uses). Consider clearing the stored `lastAssetUrl` (mapping + IndexedDB cache) when the user changes the dropdown selection.

4. **Unescaped `innerHTML` in `renderLogEntry`** (`client/js/editor/sidebar/files.js`): `filePath`, `handlerId` and `editedWidgetIds` are interpolated into `innerHTML` without escaping (only the error text is escaped). Widget IDs come from game state, so opening a game with hostile widget IDs (e.g. from the public library) and running a handler injects markup into the editor DOM. Build the entry with `textContent` instead.

5. **`querySelector` with unescaped path**: `` `.filesPanel-fileRow[data-path="${fileInfo.relativePath}"]` `` in `runHandlerForFile` throws on file names containing `"` or `\`. Use `CSS.escape()`.

6. **Background poll UX**: any change in the monitored folder re-runs `renderModule`, which destroys an open edit form while the user is typing in it; handler failures `alert()` from the 2-second poll loop (a file being saved repeatedly can produce alert storms); thumbnail `URL.createObjectURL` results are never revoked. Also, dropping a folder (`loadDroppedFolder`) immediately re-runs *all* saved mappings without confirmation — since mappings are stored in `gameSettings` and travel with the game, that can mutate a game the moment someone drops a folder with colliding file names.

7. **Dead/duplicated code**: `buildFileTree` and the `.filesPanel-tree/-list/-dir/-file` CSS block are unused (the panel renders via `buildFileTableRows`); `replaceAssetRecursive` in fileHandlers.js is never called. `csvToArray`/`unescapeCsvField`/the whole CSV-import body duplicate `client/js/jsonedit.js:727+` — extracting a shared helper would have avoided bug #1 and keeps the two importers from drifting.

8. **Minor**: the HTML handler pattern `/\.(htm|l?html)$/i` matches `.lhtml` but not `.xhtml` — probably a typo for `x?html`. And the removal of `.sidebarButton[icon] { padding-right: 0px; }` changes every existing icon-only sidebar button (they gain right padding) — if that's an intentional restyle for the new icon+text buttons, worth double-checking the other panels visually.

Since mappings and `lastAssetUrl` are written into `_meta.gameSettings`, they get saved and exported with the game — harmless data-wise (it's additive, no compat concern), but worth being aware that one creator's local file layout becomes shared room state visible to everyone in the room.

No test coverage is included; given this is editor-only and experimental that seems acceptable for now, but bug #1 is exactly the kind of thing a small jest test around a shared CSV-import helper would have caught.

---

I added the `beta` label: the PR touches no engine files (only editor UI, CSS and minify.mjs), is opt-in editor tooling that can't crash the server, and the bugs above are recoverable editing mistakes — safe and useful for players to test.

VTT-AI-Agent:
Pushed a3aeec2 addressing the review findings:

1. **CSV Replace mode orphan cards** — fixed; `oldCardTypeIDs` is now captured *before* clearing `cardTypes`, matching the jsonedit.js original, so cards of removed card types are deleted again.
2. **Unbalanced `batchStart` on handler errors** — fixed; every batched section is now wrapped in `try { … } finally { batchEnd(); }`.
3. **Dropdown ignored after first replacement** — fixed; picking a different asset in the picker clears the remembered `lastAssetUrl` (both the mapping and the IndexedDB cache), so the new selection becomes the replace target.
4. **Unescaped `innerHTML` in the log** — fixed; log entries are now built with `textContent`.
5. **Unescaped path in `querySelector`** — fixed with `CSS.escape()`.
6. **Poll UX** — the background poll no longer re-renders while an edit form is open, handler failures from the poll only go to the log instead of `alert()`-ing, thumbnail object URLs are revoked on re-render, and dropping a folder now asks for confirmation before running saved mappings.
7. **Dead/duplicated code** — removed `buildFileTree`, `replaceAssetRecursive` and the unused tree-view CSS; the three property-path handlers now share a `setWidgetPropertyByPath` helper. I did *not* refactor the CSV import to share code with jsonedit.js — jsonedit's version is entangled with its editor state (`jeStateNow`, deltas, UI), and reworking it is beyond this PR's scope; the transcription bug itself is fixed.
8. **Minor** — HTML pattern is now `/\.x?html?$/i` (.htm/.html/.xhtml), and the `.sidebarButton[icon] { padding-right: 0 }` rule is restored for all existing icon-only buttons, with the Files panel's icon+text buttons opting out via a `filesPanel-button` class.

Verified locally: jest passes, the production (minified) build starts, and the Files panel renders in a real browser without console errors.

VTT-AI-Agent (commented):
## TL;DR

Re-reviewed after commit a3aeec2: all findings from my previous review are verified fixed in the code, and all CI checks (including the previously-failing Production Environment check) now pass. I have nothing new to raise — the panel is in good shape for beta testing.

## Verification of the earlier findings

1. **CSV Replace-mode orphan cards** — fixed: `oldCardTypeIDs` is now captured from `deck.get('cardTypes')` *before* `cardTypes` is cleared (`fileHandlers.js:77-78`), so the removal loop works again.
2. **Unbalanced `batchStart` on handler errors** — fixed: every batched section (`applyCsvToDeck`, `setWidgetPropertyByPath`, `imageReplaceAsset`, both JSON handlers) wraps its work in `try { … } finally { batchEnd(); }`.
3. **Asset dropdown ignored after first replacement** — fixed: picking a different thumbnail deletes `mapping.lastAssetUrl` and clears the IndexedDB cache (`files.js:678-680`), so the new selection becomes the replace target.
4. **Unescaped `innerHTML` in the log** — fixed: `renderLogEntry` builds everything with `textContent`.
5. **Unescaped selector path** — fixed with `CSS.escape()` in `runHandlerForFile`.
6. **Poll UX** — fixed: the background poll skips re-rendering while an edit form is open, poll-triggered handler failures only log (alerts are interactive-only), object URLs are revoked on re-render, and dropping a folder now asks for confirmation before running saved mappings.
7. **Dead code** — `buildFileTree`, `replaceAssetRecursive` and the unused tree CSS are gone; the three property-path handlers share `setWidgetPropertyByPath`. Not sharing the CSV import with jsonedit.js is a reasonable scope call; worth a follow-up issue so the two copies don't drift.
8. **Minor items** — HTML pattern is now `/\.x?html?$/i`, and the `.sidebarButton[icon]` padding rule is restored for existing buttons via `:not(.filesPanel-button)`.

Only leftover nitpick: `const ext = getExtension(fileInfo.name);` in `runHandlerForFile` (files.js:768) is now unused — harmless.

CI is fully green (CodeQL, Production Environment, both TestCafe matrices, build). This is still a draft and the author may add more handlers; the current state looks solid.

---

The `beta` label was already present and I left it: the PR touches no engine files (editor UI, CSS and minify.mjs only) and is safe, opt-in editor tooling that's useful for players to test.

VTT-AI-Agent (commented):
## TL;DR
Solid, self-contained editor-only feature (no engine files touched) — this is not a risky change for the beta server. Biggest concrete issue: the CSV→deck import path silently drops newly-created card types if the CSV lacks a `cardCount::INTERNAL` column, because `targetCount` ends up `NaN` and the spawn loop never runs.

## Findings

**1. `applyCsvToDeck` silently creates 0 cards for hand-written CSVs without a count column** (`client/js/editor/sidebar/fileHandlers.js:333-386`)
`targetCounts[cardTypeID] = obj['cardCount::INTERNAL']` is `undefined` if that column is absent. Later `for (let i = 0; i < targetCount - currentCount; ++i)` becomes `i < NaN`, which is always false, so the card type is added to `deck.cardTypes` but no card widgets are ever spawned — no error, no log entry hints at this. Since the panel's whole pitch is letting users author/edit these files by hand (not just round-trip the exported format), a user typing a CSV without that internal bookkeping column will get a deck update with invisible cards. Worth either defaulting missing counts to `1` or throwing a clear error when the column is missing.

**2. CSV parsing logic is a near-verbatim duplicate of `jsonedit.js`** (`client/js/editor/sidebar/fileHandlers.js:304-331` vs `client/js/jsonedit.js:727-753`)
`csvToArray`/`unescapeCsvField`/the count-reconciliation loop in `applyCsvToDeck` are copy-pasted from the existing "import from CSV" editor action, including the same NaN edge case noted above. Two independent copies of this logic will drift over time; consider extracting a shared helper instead.

**3. Directory polling re-reads every file every 2s with no backoff** (`client/js/editor/sidebar/files.js:971-1014`)
`startPolling` calls `f.handle.getFile()` for every file in the monitored tree on a fixed 2s interval indefinitely while the panel is open, with no size/backoff limits. Fine for the small game-asset folders shown in the demo, but could get noticeably heavy on a large or deeply-nested directory. Not a blocker for an opt-in experimental feature, just a scalability note.

**4. No automated test coverage** for a fairly large new surface (1130+ new lines across two files). Given the feature is client/editor-only and explicitly marked "experimental" in a draft PR, this is understandable at this stage, but worth adding at least a couple of TestCafe or unit checks for the CSV/JSON handlers before merge, per `docs/testcafe.md`.

Minor: `client/js/editor/sidebar/files.js:1328` — `input.type = sch.type === 'string' ? 'text' : 'text';` is a dead ternary (both branches identical); can just be `input.type = 'text';`.

No security issues found: all user/file-derived strings are inserted via `textContent`/`Node.append(string)` rather than `innerHTML`, so there's no obvious XSS path from monitored file names or content.

## Beta label
The diff touches `client/js/editor/*` and `server/minify.mjs` (just adding the two new files to the minify list) only — no changes to `client/js/widgets/`, `compute.js`, or `serverstate.js`, so this is not an engine change. The PR already carries the `beta` label; I left it as-is (did not add or remove it) since it's a pre-existing, purely additive editor panel that doesn't touch how existing games run.

PR #2884, opened 2026-02-11
Update to Properties editor UI

RaphaelAlvez:
General Update on the properties editor UI

from #2946:

Adds custom inputs for basic properties that all widgets should have

x/y/width/height gets number input with slider limited to room size

movable is a checkbox and movable in editor is only visible when widget is locked in play mode Logic is inverted from property

layers is a dropdown instead of a number input to allow for reference to default values range is a bit bigger then the defaults. custom values will be displayed in the dropdown

adds parent input with widget picker

From #2957

'm not sure of the design so I decided to put it in a different PR for now while waiting for feedback

the current plan is to have each widget being inherited from as a colapsbale sub section. Each section can eithe "copy all properties", "copy some properties" and "exclude some properties"

currently, inside each section there is a list of all properties declared in inheritFrom, in the current widget (default propertoes and declared) and in the widget being copied from (also default and declared properties),
Maybe this is too much but it feels like the logical list to provide.

todo:

I'm not sure of the design so I decided to put it in a different PR for now while waiting for feedback

the current plan is to have each widget being inherited from as a colapsbale sub section. Each section can eithe "copy all properties", "copy some properties" and "exclude some properties"

currently, inside each section there is a list of all properties declared in inheritFrom, in the current widget (default propertoes and declared) and in the widget being copied from (also default and declared properties),
Maybe this is too much but it feels like the logical list to provide.

todo:

- [x] An error version of a widget subsection. this is just in case someone mix exclude and include on JSON. The error will be explained and a text input with the list
- [x] The exclude button in ugly
- [x] The add widget button is misalligned
- [x] I don't really like the check box and want to make a button for each propertie
- [x] Copy all still needs something. Specially for the tooltips
- [x] Tooltips explaining that someproperties are not copied because they are declared in the current widget
- For properties declared in the current widget:
    - [x] Add visual cue that it is not being copied maybe something like "<del>property</del>"
    - [x] Add button to delete property in current widget

From this PR inithial description:

Updates the label properties editor UI

Before
image

After
image
image

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2884/pr-test (or any other room on that server)


96LawDawg:
Maybe spell out Title and Header as not everyone will know what that means. And there is potential confusion because "normal" sounds like it might be the default, but it is not.  Maybe call it "regular" text?

RaphaelAlvez:
> Let me preface these comments with: I have not played around in the Properties code at all so I don't know what is a good idea and what is possible. And I don't know how far we want to take this. That being said here are some comments and food for thought: 

Thanks. I also don't know how far we want to go with it so I'm glad for the feedback and suggestions.

I want to find a balance between simplicity and usefulness. I want it to cover 90% of simple games while not feeling overwhelming.

obviously this balance cannot be found without ohters feedback. Now that it represent  that we have 

> * When you change style, the color gets reset back to black. Would be nice if that didn't happen.
> * Dragging the cursor around the color changer (in Edge on Windows) adds semicolons to the css as it moves like this:
>   `  "css": "font-size: 50px; font-weight: bold;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;; color: #ff0000",`

Both problems are quite related. I think I'll just use object css. It will make chaging text styles so much easier.
I actually forgot to check the JSON. As the ;; don't affect the result I didn't catch that

> * In the style area, could we add a number picker to set the exact font size

I didn't even think of it. Maybe I'll make a text style bar so it can also be used in other places

something similar to this but that will be applied to the whole label
image

covering:
- bold
- italic
- color
- fontsize
- font

> * The placeholder text does not get the css color. Probably because it is a different css class?

Yes because I was trying to keep the css in the string format. As I said I'll move to object format

> * Should we add in somewhere the `spellCheck` and `tabIndex` properties?

I opted to not add them because they feel very niche. as ArnoldSmith86 said, If they are defined in a game they will be displayed in "other properties". If you feel they would be usefull I can easily include them.

> * And are these Properties features supposed to include all of the properties in the "Other properties" area or just other significant ones?

I have plans improvethe other properties part. It will remain as a fallback for any property set in the JSON editor but I want to group comon properties and group systematic properties (Routine for "automation" and css for "apearance")

96LawDawg:
Another thing I just tested.  If I have the nested object form of css, then changing color does nothing and changing style/font-size erases the other css. I don't know many people that would use nested object css and then use the Properties sidebar, but we should probably prevent it from being destructive to what is already there including if it is just the basic object form. This may be difficult to code.

96LawDawg:
@VTT-AI-Agent, client crash when editing css in the Edit Widgets toolbar. The css for the "base widget" was set to "font-size: 30px". When I deleted "30px" the client crashed with this error: `Cannot read properties of null (reading 'value')
TypeError: Cannot read properties of null (reading 'value')`

VTT-AI-Agent:
I retested this against the current PR head (0446db93). With a basic widget using `"css": "font-size: 30px"`, clearing `30px` in Edit Widgets → Appearance → CSS leaves the client running and saves `"css": "font-size: "` without an exception. This appears to have been fixed by the later CSS editor rewrite; no additional patch is needed.

96LawDawg:
@VTT-AI-Agent, I was not specific enough. I was working in a label. This does not seem to glitch on a basic widget or button. The label css does cause a crash when deleting the value of font-size or even just typing "font-size:".  See the screenshot.

image

VTT-AI-Agent (commented):
## TL;DR
Large, well-structured editor-UI feature with genuine unit-test coverage — no correctness bugs found in the property-input logic (cycle-safe parent picker, robust seat/inheritFrom parsing, all 29 jest tests pass). Two things stand out as unrelated to "properties editor UI": an accidental `.gitignore` corruption and a `server/room.mjs` public-library refactor that also changes behavior. No engine files are affected.

## Findings

### 1. `.gitignore` — accidental corruption (should be reverted)
The last line was changed from `.appimage/out/` to `.appimage/out/.vscode/`:
```
-.appimage/out/
+.appimage/out/.vscode/
```
This drops the ignore for the AppImage build-output directory entirely and replaces it with a nonsensical nested path, so `.appimage/out/` contents would now show up as untracked. This looks like an unintended merge/edit artifact — please restore `.appimage/out/` (adding a trailing newline while you're at it).

### 2. `server/room.mjs` — unrelated scope + behavior change
The `getPublicLibraryGames()` rewrite (recursive folder scan + `try/catch` robustness) is a reasonable improvement on its own, but it's unrelated to the properties-editor UI and works against "one PR = one concern" (`docs/conventions.md`). More importantly it's a **behavior change** that's easy to miss: the old code deleted the variant-specific fields from the game-level info after copying them into each variant record:
```js
delete gameFile._meta.info.players;
delete gameFile._meta.info.language;
delete gameFile._meta.info.variant;
delete gameFile._meta.info.variantImage;
```
The new code (`Room.publicLibrary[id] = Object.assign({ name }, metaInfo)`) drops those deletes, so the top-level library entry now carries the **first variant's** `players`/`language`/`variant`/`variantImage`. If that's intentional, worth calling out; if not, it will subtly change what the library browser shows at the game (vs. variant) level. Recommend splitting this into its own PR so it can be reviewed/tested against the library UI independently.

### Things that look good (checked, no action needed)
- `setWidgetParent` correctly rejects self-parenting and descendant-as-parent (`isDescendantOf`), so the new parent picker can't create a hierarchy loop.
- Seat-reference normalize/parse/format round-trips and timer `formatTimerMs`/`parseTimerInput` are covered by tests and behave sensibly on empty/invalid input.
- The recursion in `room.mjs` uses `lstatSync().isDirectory()`, so symlinked directories aren't followed — no infinite-loop risk.
- `propertyInfoButton` is intentionally renamed to avoid a duplicate top-level `infoButton` declaration when bundled alongside the routine-editor PR — good foresight for the beta bundle.
- The `e.key && e.key.match(...)` guard in `jsonedit.js` is a harmless robustness fix.
- No widget/routine properties are added or renamed, so the `validator/` and `jeCommands` upkeep in `CLAUDE.md` doesn't apply here.

### Minor
- `npm test` passes (29/29, including the new `tests/client/property-inputs.test.js`). Because the editor scripts are concatenated by `minify.mjs` rather than imported, the tests re-`eval` the sources via `new Function`, which won't catch runtime integration issues — a TestCafe pass covering the new sidebar would add confidence given the size of the diff.

## Beta label
The `beta` label was already present and I left it in place: the PR touches no engine files (`client/js/widgets/`, `compute.js`, `serverstate.js` are untouched — `main.js`/`jsonedit.js` are editor/client glue), it's gated to edit mode so it doesn't change how existing games run, and the one server-side change is guarded with `try/catch` and can't crash the loader, so it's safe and useful for players to test.

VTT-AI-Agent:
Thanks for the label-specific reproduction. Fixed in 6184b2a1: incomplete `font-size` values now parse safely while the CSS textarea is being edited, so both `font-size:` and an empty value no longer crash the label editor. Added regression coverage and verified the exact label workflow locally.

VTT-AI-Agent:
Addressed both review findings in 98ec2837. Restored `.appimage/out/` (with a trailing newline), and removed the public-library scanner refactor from this PR so it remains solely in `main` rather than changing library metadata behavior here. Verified with `node --check server/room.mjs` and `npm test` (30 passing).

96LawDawg (approved):
Reviewed.  Approved.  Ready as far as I'm concerned.

VTT-AI-Agent (commented):
## TL;DR
This is a big, welcome upgrade to the widget **Properties** editor — the grouped sections, custom sliders/toggles, and the new color/icon pickers are a genuine step up in usability. The one change I'd push hardest for: the **"Inherit properties from" section renders an exhaustive alphabetical checklist of *every* widget property** (80+ empty checkboxes), which is overwhelming and buries the few properties that matter — default it to the properties actually declared on the source/current widget with a "show all" toggle.

> This is a **UI/UX review** (layout, visual hierarchy, readability for a first-time user), not a code-correctness review. I reviewed it live: server on my assigned port, `basic` widget selected in edit mode, with an `inheritFrom` source configured to exercise the new sections. Screenshots below are from that session.

Overall the panel reads clearly and matches the app's existing blue-section style. Sections (Basic → Content → Appearance → Automations → Other properties) are a big readability win over the old flat JSON-ish form, and the collapsible Position/Size/Generic groups keep it tidy.

overview

---

### Big layout improvements

**1. The "Inherit properties from" property list is far too much.** When a source widget's subsection is expanded, it shows a single flat, alphabetical column of checkboxes for *every possible property* — `activeFace`, `animatePropertyChange`, `dragLimit`, `dropShadowOwner`, `faceCycle`, `hoverInheritVisibleForSeat`, … dozens of them, almost all irrelevant to a given widget. It dominates the whole panel and is very hard to scan for the handful you actually care about. (The author already flagged this as "maybe this is too much" in the description — agreed.)

inherit list

Suggestions, in order of impact:
- **Default to the relevant subset**: properties actually declared on the source widget and/or the current widget, plus anything already checked. Put the full list behind a "Show all properties" toggle for power users.
- Or **group the list** ("Declared on source", "Declared here (not copied)", "Other") with the first two expanded and the long tail collapsed.
- The checked/declared/struck-through states are a nice idea — they just get lost in the noise of 80 empty rows.

**2. Normalize the row of action buttons under the inherit list.** `+ Add widget`, `add Parent`, and `add Seat` currently mix two visual styles and two capitalization conventions ("Add widget" title-case with a `+` and an attached dropdown caret vs. lowercase blue pills). Giving them one consistent button style, casing, and alignment would make this cluster read as one coherent toolbar rather than three unrelated controls.

inherit + colors

---

- **Color row rhythm breaks on the last swatch.** In Appearance → Colors each control is `icon + swatch`, but the last one ("Icon/Symbol") is `icon + (i) + swatch`, so its info button is wedged between the label icon and the swatch and throws off the otherwise even spacing of the row (`.colorFlexRow`). Consider moving that `(i)` to a consistent position (e.g. always trailing the swatch) so every group has the same shape.
- **Truncated label in "Other properties".** The `backgroundColor` row shows `backgroundCo` — the label column clips the property name (`propertyInputs.css` fixed 132px label width). Long property names should ellipsize with a tooltip, wrap, or get a title attribute so the full name is recoverable.
- **"Widget id" field looks off.** The value sits right-aligned at the end of a long dotted underline with a big empty gap to its left, which reads more like a broken input than an editable name field. Left-aligning the value (or shrinking the field to fit) would look more intentional.
- **Not-copied (struck-through) properties are very low contrast.** The `<del>`-style gray on the declared-here properties is hard to read; a slightly stronger color or an explicit "declared here" tag would make the state clearer.
- **The dark "Edit widget properties." tooltip banner** overlaps the top-right corner of the panel header on open. Minor, but it collides with the section title area for a moment.

---

- **Color controls are icon-only.** The four color slots render as bare Material icons (`A`, paint-bucket, a **pen** for Border, a **shapes/"category"** glyph for Icon/Symbol) with no visible text. They do expose the label as a hover `title`, which helps, but "border = pen" and "icon color = shapes" aren't self-evident, and hover tooltips don't exist on touch. A tiny text label under/next to each swatch (or at least more literal icons) would make this instantly legible to a newcomer.
- **The lock toggles are cryptic labels.** In Position, a red lock labeled **"Position"** actually controls whether the widget can be *dragged* (`movable`), and the second one is labeled **"Also in editor"** (`movableInEdit`). To someone who's never used the app, "Position" next to a padlock reads like "lock the position value," and "Also in editor" is meaningless standing alone. Wording like **"Movement: locked/unlocked"** and **"Lock in editor too"** would convey intent without needing the tooltip.
- **Inherit terminology is advanced but well-supported.** "Inherit properties from", "copy selected / exclude selected", and the helper line *"Some properties are not copied because they are declared in the current widget"* are clear once you know the feature. The helper text is a good touch — keep that pattern.
- **What's genuinely good for newcomers:** the **icon picker** (search box + labeled library filter chips + a grid of suggested icons + "Show all") and the **color picker** (a "Used in this game" row, a full palette, and the "checkerboard = transparent" hint) are both clear, well-labeled, and easy to understand. The pervasive `(i)` info buttons are a nice, discoverable affordance. These set the bar the rest of the panel should match.

icon picker
color picker

---

Net: the structure and the pickers are strong. Taming the inherit property list and adding a few text labels/clearer wording (colors, lock toggles) would take this from "much better" to "obvious at a glance for a first-timer."

VTT-AI-Agent:
Addressed the UI review in 0db0ab28:

- The inherit list now starts with properties declared by either widget (plus selected properties), with a **Show all properties** option for the complete defaults list.
- Color swatches now retain visible role names; the local-declaration state is readable instead of struck through; and generic long property names wrap.
- Clarified play/editor movement locks, normalized add labels, and left-aligned the rename field.

Verified with `npm test` (39 passing) and the local Chromium editor smoke test.

VTT-AI-Agent:
Addressed in 2d611a70.

> Fix:
> - Remove info icon (i) from under Appearance for the Icon/Symbol color swatch so it is the same as the other three.
> - Remove the hover `title` from all the Appearance graphics for the color swatches as they already have text now.
> - Client crash. To replicate, a new widget. Rename the widget. Before doing anything else, move the widget in the room. Error: `Cannot read properties of undefined (reading 'id') TypeError: Cannot read properties of undefined (reading 'id')`
>
> Source: [Discord message from "lawdawg96"](https://discord.com/channels/770758631146782780/1529929694860939366/1530014506791473272)

All Appearance color controls now use their visible labels without redundant hover titles; Icon/Symbol no longer renders its extra info marker. Drag events that outlive a rename now discard their stale removed-widget target safely. Added a TestCafe regression covering the color row, rename, and move flow; `npm test` passes (39 tests).

VTT-AI-Agent:
Implemented in 111ce95c: Properties-panel section headings now keep a matching 10px gap before the scrollbar.

> Create a small gap on the right side of the Edit Widgets toolbar between the section headers and the scrollbar. On the left side there is a small gap where the headers are separated from the left edge. Create the same sized gap on the right side so the scrollbar does not blend in with the blue section bars.
>
> (Source: Discord message from "lawdawg96" — https://discord.com/channels/770758631146782780/1529929694860939366/1530020587043422381)

VTT-AI-Agent:
Implemented in 5236478e: gray collapsible headers now use the same 10px right inset as the blue section headers.

> The gray section headers need the same inset.
>
> (Source: Discord message from "lawdawg96" — https://discord.com/channels/770758631146782780/1529929694860939366/1530021579596239003)

VTT-AI-Agent (commented):
## TL;DR

The editor rewrite is broadly safe to beta-test and CI is green, but two saved-state semantics need correction: the inheritance UI disagrees with the engine about explicit default-valued properties, and the size-ratio control writes an unregistered widget property that the validator flags.

## Findings

### 1. Explicit default-valued properties are incorrectly shown as inheritable (`client/js/editor/sidebar/properties.js:3274`)

`isPropertyDeclaredOnWidget()` treats a property as undeclared when its explicit state value equals the widget default. The inheritance engine does not use that rule: `applyInheritedValuesToObject()` only inherits when `targetWidget.state[key] === undefined` (`client/js/widgets/widget.js:329-332`).

For example, give a basic target widget an explicit `"width": 100`, give its source `"width": 200`, and select `width` in `inheritFrom`. This UI considers the target width undeclared, leaves the checkbox enabled/checked, and provides no eraser action, but the engine keeps the target at 100 because `state.width` exists. This makes the editor promise inheritance that cannot take effect. Declaration/blocking should be based on own-property presence in `state`, irrespective of equality with `defaults`.

### 2. The ratio lock persists an unknown widget property (`client/js/editor/sidebar/properties.js:2458`)

Toggling “Size ratio” writes `lockSizeRatio` into the synchronized widget state, but the property is absent from widget defaults and `validator/validate_gamefile.js`. A minimal basic widget with `"lockSizeRatio": false` produces `unrecognized (and seemingly unused) property` in `validateGameFile`, so using this new control makes an otherwise valid game report a validation problem.

If this is only editor UI state, keep it in editor-local state (as the collapsible state already is). If it is intended to be a persisted widget property, register its boolean default/validator and the corresponding JSON-editor support required for public widget properties.

## Engine and beta assessment

No excluded engine files (`client/js/widgets/`, `client/js/compute.js`, or `client/js/serverstate.js`) are changed. The `client/js/mousehandling.js` edit is an incidental defensive guard for a removed DOM target during editor-driven ID renaming; it only exits when no widget exists and leaves normal game event handling unchanged.

The `beta` label was already present and I left it in place because the PR is editor-focused, CI and CodeQL pass, and the remaining findings do not make player testing unsafe.

VTT-AI-Agent (commented):
## TL;DR
Solid, well-structured editor-only overhaul with real unit coverage and no engine changes. The one thing worth fixing before merge: switching an inherit source to **"exclude selected"** stores `inheritFrom[src] = []`, which the engine reads as *inherit nothing* rather than *inherit all except nothing* — so the moment you pick exclude mode the widget silently stops inheriting everything. Two smaller issues: clearing a number field writes `0` to the property, and the new inherit-source picker (unlike the parent picker) has no cycle guard.

## Findings

### 1. "Exclude selected" mode writes an empty array → inherits nothing (not "exclude nothing")
`client/js/editor/sidebar/properties.js:3392-3399` (and the uncheck-all path at `:3573-3580`).

When the mode dropdown is set to `excluded`, the code stores `inheritFrom[sourceWidgetId] = []`. The engine's `inheritFromIsValid` (`client/js/statemanaged.js:88-100`) only treats an array as an *exclude* list when `properties[0][0] == '!'`. An empty array has no such first element, so it falls into the **include** branch (`[].indexOf(key) != -1`), which is `false` for every key — i.e. the widget inherits *nothing*.

Repro: widget A inherits all from B (`{"B":"*"}`). Open the row, switch the dropdown to "exclude selected" intending to keep inheriting and drop a couple of props. Stored value becomes `{"B":[]}` and A immediately stops inheriting *all* of B. The UI still shows "exclude selected" with everything unchecked (i.e. "excluding nothing"), so the displayed state and the actual effect diverge. Unchecking the last excluded item in the checkbox list (`finalMode = selectedProps.map(p => '!' + p)` → `[]`) hits the same trap. "Exclude nothing" needs to be stored as `'*'` (or the exclude array must always carry at least a sentinel), not `[]`.

### 2. Clearing a number field writes `0` to the property
`client/js/editor/propertyInputs.js:457-470` (`NumberInput.applyInput`) and the parallel `normalizeValue` in `renderDualNumberWithSlider` (`properties.js:2295-2298`).

`applyInput` only special-cases empty input when `nullIfEmpty` is set; otherwise `+'' === 0` is finite, so it calls `setValue(0)`. Selecting-all and deleting to retype a value therefore writes `0` as an intermediate delta. Concrete cases with no `nullIfEmpty`:
- Seat index (`properties.js:4560`, `min:1 max:16`) → `index` becomes `0`, below the stated minimum.
- First column width (`properties.js:4531`, `min:10 max:500`) → `0`.
- X/Y/W/H (`renderDualNumberWithSlider`) → clearing W or H collapses the widget to `0` size mid-edit.

It self-corrects once the user finishes typing (and is undoable), but it emits a spurious out-of-range delta on a completely normal gesture. Relatedly, typed values are never clamped to `min`/`max` — a native `<input type=number>` doesn't block out-of-range typing, so typing `99` into Seat index (`max:16`) persists `99`. Treating empty as "leave unchanged" (or clamping in `applyInput`) would fix both.

### 3. Inherit-source picker has no cycle guard
`client/js/editor/sidebar/properties.js:3626-3636` (`renderInheritFromAddButton`).

The parent picker correctly rejects loops via `setWidgetParent` → `isDescendantOf` (`:2747-2768`). The inherit-source picker only excludes already-added sources and self (`excludeIDs: () => Object.keys(...)`) with no transitive check. Picking a source that (directly or transitively) already inherits *from the current widget* creates a cycle, and `applyInheritedValuesToDOM` (`client/js/widgets/widget.js:335-341`) recurses `w.inheritFrom()` with no visited set → infinite recursion / stack overflow that freezes the editor on next render. This is a pre-existing engine limitation (the JSON editor could already create it), but the new picker turns it into a one-click mistake, so mirroring the parent picker's descendant guard here would be worthwhile.

## Looks good (verified, no action)
- `movable` / `movableInEdit` lock toggles (`:2413-2499`) — invert logic is correct in both directions.
- Widget-id rename (`:3050-3085`) delegates to the existing `updateWidgetId`, which remaps children/cards/inheritFrom/dropTargets/routines. Sound.
- `setWidgetParent` cycle guard, position preservation, batch wrapping — correct.
- Non-empty exclude arrays (`['!width', …]`) round-trip correctly; delete-source and delete-property buttons remove the right key.
- `mousehandling.js` null-widget guard for events in-flight after an ID rename, and the `jsonedit.js` `e.key && …` guard — both good defensive fixes.
- No new widget or routine-operation property is added, so `validator/` and `jeCommands` upkeep doesn't apply here.
- The new `tests/client/property-inputs.test.js` covers the pure string helpers well (icon/image search, `renderIconChip`, value formatting). Gap: none of the input *classes* are exercised (they need a DOM), so findings #1–#2 sit in untested code.

## Beta label
Left the already-present `beta` label in place: this PR touches no engine files (no `client/js/widgets/`, `compute.js`, or `serverstate.js`; `main.js`/`jsonedit.js`/`mousehandling.js` are edit-mode/client glue), all behavior is gated to the editor so existing games are unaffected at runtime, and the issues above are recoverable editor-only glitches that neither corrupt saved games nor crash the server — safe and useful for players to test.

VTT-AI-Agent (commented):
## TL;DR
Solid, well-structured editor-only overhaul with real unit coverage and no engine changes. The one thing worth fixing before merge: switching an inherit source to **"exclude selected"** stores `inheritFrom[src] = []`, which the engine reads as *inherit nothing* rather than *inherit all except nothing* — so the moment you pick exclude mode the widget silently stops inheriting everything. Two smaller issues: clearing a number field writes `0` to the property, and the new inherit-source picker (unlike the parent picker) has no cycle guard.

## Findings

### 1. "Exclude selected" mode writes an empty array → inherits nothing (not "exclude nothing")
`client/js/editor/sidebar/properties.js:3392-3399` (and the uncheck-all path at `:3573-3580`).

When the mode dropdown is set to `excluded`, the code stores `inheritFrom[sourceWidgetId] = []`. The engine's `inheritFromIsValid` (`client/js/statemanaged.js:88-100`) only treats an array as an *exclude* list when `properties[0][0] == '!'`. An empty array has no such first element, so it falls into the **include** branch (`[].indexOf(key) != -1`), which is `false` for every key — i.e. the widget inherits *nothing*.

Repro: widget A inherits all from B (`{"B":"*"}`). Open the row, switch the dropdown to "exclude selected" intending to keep inheriting and drop a couple of props. Stored value becomes `{"B":[]}` and A immediately stops inheriting *all* of B. The UI still shows "exclude selected" with everything unchecked (i.e. "excluding nothing"), so the displayed state and the actual effect diverge. Unchecking the last excluded item in the checkbox list (`finalMode = selectedProps.map(p => '!' + p)` → `[]`) hits the same trap. "Exclude nothing" needs to be stored as `'*'` (or the exclude array must always carry at least a sentinel), not `[]`.

### 2. Clearing a number field writes `0` to the property
`client/js/editor/propertyInputs.js:457-470` (`NumberInput.applyInput`) and the parallel `normalizeValue` in `renderDualNumberWithSlider` (`properties.js:2295-2298`).

`applyInput` only special-cases empty input when `nullIfEmpty` is set; otherwise `+'' === 0` is finite, so it calls `setValue(0)`. Selecting-all and deleting to retype a value therefore writes `0` as an intermediate delta. Concrete cases with no `nullIfEmpty`:
- Seat index (`properties.js:4560`, `min:1 max:16`) → `index` becomes `0`, below the stated minimum.
- First column width (`properties.js:4531`, `min:10 max:500`) → `0`.
- X/Y/W/H (`renderDualNumberWithSlider`) → clearing W or H collapses the widget to `0` size mid-edit.

It self-corrects once the user finishes typing (and is undoable), but it emits a spurious out-of-range delta on a completely normal gesture. Relatedly, typed values are never clamped to `min`/`max` — a native `<input type=number>` doesn't block out-of-range typing, so typing `99` into Seat index (`max:16`) persists `99`. Treating empty as "leave unchanged" (or clamping in `applyInput`) would fix both.

### 3. Inherit-source picker has no cycle guard
`client/js/editor/sidebar/properties.js:3626-3636` (`renderInheritFromAddButton`).

The parent picker correctly rejects loops via `setWidgetParent` → `isDescendantOf` (`:2747-2768`). The inherit-source picker only excludes already-added sources and self (`excludeIDs: () => Object.keys(...)`) with no transitive check. Picking a source that (directly or transitively) already inherits *from the current widget* creates a cycle, and `applyInheritedValuesToDOM` (`client/js/widgets/widget.js:335-341`) recurses `w.inheritFrom()` with no visited set → infinite recursion / stack overflow that freezes the editor on next render. This is a pre-existing engine limitation (the JSON editor could already create it), but the new picker turns it into a one-click mistake, so mirroring the parent picker's descendant guard here would be worthwhile.

## Looks good (verified, no action)
- `movable` / `movableInEdit` lock toggles (`:2413-2499`) — invert logic is correct in both directions.
- Widget-id rename (`:3050-3085`) delegates to the existing `updateWidgetId`, which remaps children/cards/inheritFrom/dropTargets/routines. Sound.
- `setWidgetParent` cycle guard, position preservation, batch wrapping — correct.
- Non-empty exclude arrays (`['!width', …]`) round-trip correctly; delete-source and delete-property buttons remove the right key.
- `mousehandling.js` null-widget guard for events in-flight after an ID rename, and the `jsonedit.js` `e.key && …` guard — both good defensive fixes.
- No new widget or routine-operation property is added, so `validator/` and `jeCommands` upkeep doesn't apply here.
- The new `tests/client/property-inputs.test.js` covers the pure string helpers well (icon/image search, `renderIconChip`, value formatting). Gap: none of the input *classes* are exercised (they need a DOM), so findings #1–#2 sit in untested code.

## Beta label
Left the already-present `beta` label in place: this PR touches no engine files (no `client/js/widgets/`, `compute.js`, or `serverstate.js`; `main.js`/`jsonedit.js`/`mousehandling.js` are edit-mode/client glue), all behavior is gated to the editor so existing games are unaffected at runtime, and the issues above are recoverable editor-only glitches that neither corrupt saved games nor crash the server — safe and useful for players to test.

VTT-AI-Agent:
Addressed the review findings in 92c94749.

- Explicit state properties now block inheritance even when they equal defaults.
- Empty exclusion lists now persist as copy-all, while empty include lists remain empty.
- Incomplete number fields do not emit `0`, and committed values respect their configured bounds.
- The inherit-source picker rejects direct and transitive inheritance cycles.
- Size-ratio locking is editor-local; legacy `false` values remain honored without creating new unregistered properties.

Added focused regression coverage; `npm test` passes (44 tests), and the local editor TestCafe smoke test passes.

PR #2891, opened 2026-02-15
better connection monitoring and player status

ArnoldSmith86:
Couchlife on Discord lost a lot of work recently because VTT did not notice that it was disconnected from the server. And this was not the first time this happened. I already started once to improve the situation with #1514 but I wanted to try a different approach.

This PR gives each delta an ID and the server sends a little confirmation for each one. The client tracks unconfirmed deltas and escalates through a small status overlay in the corner:

- after 5 seconds without confirmation: a small pulsing icon
- after 10 seconds: "No response from server."
- after 20 seconds: a countdown warning that the page will reload
- after 30 seconds: the page simply reloads (no buffering or delta-stitching — if a fresh state arrives that makes old unconfirmed deltas moot, the player gets a short "your last changes could not be saved" notice instead)

Time in which the tab was suspended (device sleep, frozen background tab) does not count towards those delays, so waking a phone up never lands on an unannounced reload.

When the websocket actually closes, the same overlay shows "Connection lost. Reconnecting..." instead of the old full-screen connection-lost overlay. A detected server restart is a different matter - that page is not getting its connection back and is about to reload - and keeps the full-screen notice it got in #3147, now as an overlay of its own (`serverRestartOverlay`) since the "connection lost" half of that overlay no longer has a caller. The corner display stays quiet while it is up.

The same status overlay is also used for player status messages:

- player joined / left / renamed messages
- "X is in Edit Mode"
- "X has an input window open"

(Notifications for players merely having Game Shelf / About / Players open were removed again to keep it simple.)

The bookkeeping of unconfirmed deltas and the escalation decision live in `client/js/deltamonitor.js`, free of DOM access and timers so they can be unit tested with an injected clock (`tests/client/deltamonitor.test.js`). `tests/testcafe/connection.js` covers the whole handshake in a real browser: it holds deltas back on an otherwise healthy connection and checks that the status display escalates and recovers once they get through.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2891/pr-test (or any other room on that server)


96LawDawg:
I like the player x "has input open".  And the notification about edit mode would resolved #2383.

VTT-AI-Agent:
I finished the simplification described in the last comment and pushed it (plus a merge of current `main`):

- **Removed the delta buffering and reconnect-stitching** (`justReconnected` re-apply, `previousDeltaSendId` chaining, `forceReconnect`). A fresh state from the server now simply clears unconfirmed deltas; if some were pending for a while, the player gets a short "Connection restored. Your last changes could not be saved." notice.
- **Reload after 30s** without confirmation is kept, and the warning now shows a live countdown ("Reloading the page in Xs") instead of a fixed text.
- **Removed the Game Shelf / About / Players overlay notifications** including all the `activeStateName` plumbing (states.js is back to the `main` version). Edit Mode and input-window status plus join/leave/rename messages stay.
- When the websocket closes, the status overlay now shows "Connection lost. Reconnecting..." (the old full-screen overlay had already been removed, but nothing had replaced it on this code path).
- Fixed a regression: `activePlayers` had lost its deduplication, which would have leaked one-entry-per-client lists into routines (`compute`), the timer auto-pause and the save dialog. The raw list is now only used locally for the join/leave messages.
- Stale status entries for players who left are now pruned, and the server strips the client-only `deltaSendId` before broadcasting deltas to other players.

Verified with a two-client Playwright session against a local server: join/leave/rename messages, Edit Mode status (appears and clears), reconnect status when the server process dies, the 5s/10s/20s escalation against a SIGSTOPped server (open connection, no responses — the scenario from the original report), countdown cancellation when a late confirmation arrives, and the automatic reload at 30s. Jest passes.

VTT-AI-Agent (commented):
## TL;DR

Solid approach — the delta-confirmation escalation is simple and the removal of the buffering/stitching from earlier iterations makes this much easier to reason about. CI (jest + TestCafe) is green. The most important finding: `showOverlay()` leaves a stale `currentOverlayId` when an overlay is *toggle-closed* via its toolbar button, so every subsequent mouse message reports a closed overlay as active — which can permanently show "X has an input window open" to other players. There's also a cosmetic self-rename announcement bug and a question about force-reloading while the websocket is known-closed.

## Findings

### 1. Stale `activeOverlay` when an overlay is toggle-closed (`client/js/main.js:99-117`)

`setCurrentOverlayId(id)` is called unconditionally at the top of the `if(id)` branch, but the next line *toggles*: if the overlay was already open, `style.display` becomes `'none'` and `overlayActive` becomes `false`. In that case `currentOverlayId` stays set to the now-closed overlay, so:

- the `toServer('mouse', ...)` at the end sends the contradictory `{ inactive: false, activeOverlay: '<closedOverlay>' }`, and
- every subsequent mousemove (`mousehandling.js:162`) keeps broadcasting that closed overlay as active until another overlay is opened.

This is reachable for the status display: `#activeGameButton` is a `.toolbarButton` and gets `dataset.overlay = 'buttonInputOverlay'` (`widget.js:2731`), so closing an input window by clicking that toolbar button leaves the player showing "X has an input window open" to everyone else indefinitely. Fix is one line — set the id based on the outcome, e.g. after computing `overlayActive`:

```js
setCurrentOverlayId(overlayActive ? id : null);
```

### 2. Self-rename is announced as "left; joined" (`client/js/overlays/players.js:57`)

The server sends `rename` to the renamed player *before* broadcasting the meta update (`room.mjs:909-913`), so by the time `announcePlayerChanges` runs on the renaming client, `playerName` is already the new name. The `fresh[0] != playerName` guard then makes the rename fall into the `else` branch, producing `"OldName left; NewName joined"` — which gets concatenated (same icon, within 10 s) with the `"You renamed to NewName"` message from the `rename` handler. The guard should skip the announcement entirely for a self-rename (e.g. `if(fresh[0] == playerName) return;` before the rename check) instead of falling through to left/joined.

### 3. Forced reload while the socket is known-closed (`client/js/serverstate.js:638-640`)

When the websocket actually closes with deltas pending, `updateConnectionMonitor` still counts down and calls `location.reload()` after 30 s. If the server is still unreachable at that point, the reload lands on a browser error page and the user loses the automatic reconnect-with-backoff loop (`connection.js:42-43`) that the "Reconnecting..." state relies on — arguably worse than the old behavior for a longer outage. The reconnect path already handles recovery (fresh `state` clears `pendingDeltas` and shows the "changes could not be saved" notice), so consider suppressing the reload/countdown while in the known-disconnected state and letting it apply only to the zombie-connection case this PR is really targeting (socket open, no confirmations).

### Minor

- `editMode`/`activeOverlay` only piggyback on mousemoves and overlay changes, so toggling Edit Mode isn't visible to others until the next mouse move; and on rename, other clients drop the player's activity (`setActivePlayersList`) while the server's `lastMouseStateByPlayer` is only replayed to *new* joiners — existing players lose the renamed player's status until their next mousemove. Both self-heal quickly; just noting it.
- `setStatusMessage` appends with `'; '` and re-extends `expiresAt` on every call, so in a busy room a join/leave message chain can grow without bound and never expire. Consider capping the concatenation or not extending the expiry.

---

This PR touches engine files (`client/js/serverstate.js`), which are excluded from beta auto-deployment by design, so I did not add the `beta` label (it's also still a draft).

VTT-AI-Agent:
Addressed the review in 8afc90c4:

1. **Stale `activeOverlay` on toggle-close** — fixed as suggested: `setCurrentOverlayId(overlayActive ? id : null)` after the toggle outcome is known. Verified via websocket frame capture that a toggle-close now sends `activeOverlay: null` (both on a genuine double-toggle and on the normal close-via-Active-Game-tab path). One correction to the failure scenario though: the "permanently shows input open" case isn't actually reachable — `#activeGameButton` with `dataset.overlay` set fires *both* its handlers (`showOverlay()` then `showOverlay('buttonInputOverlay')`), which net re-opens the input overlay (intended, since an input is pending), and the other `data-overlay` buttons are `.toolbarTab`s whose handler stops propagation when already active, so they never toggle-close. Still, the id could go stale on any future `showOverlay(sameId)` call, so the fix is worth having.

2. **Self-rename announced as "left; joined"** — fixed: the rename-shaped diff is detected first, and if the fresh name is my own it announces nothing (the `rename` handler's "You renamed to ..." covers it). Verified in a two-client session: the renaming client only sees "You renamed to Bella", the other client sees "Guest… renamed to Bella".

3. **Forced reload while socket known-closed** — agreed and fixed: `updateConnectionMonitor` now bails out while disconnected, so the countdown/reload only applies to the zombie-connection case (socket open, no confirmations). Verified: with an unconfirmed delta and the server killed, the client shows "Connection lost. Reconnecting..." indefinitely with no reload, and reconnects cleanly (with the "changes could not be saved" notice) once the server returns.

Also took the minor point about unbounded message concatenation: status messages now expire individually (10s each) and are capped at the 5 most recent, instead of one shared expiry that every append extends. On the other minor note: Edit Mode toggles *are* broadcast immediately (`toggleEditMode` → `showOverlay()` sends a mouse message); only the rename-gap point stands, and it self-heals on the next mouse move as you said.

Jest passes; all flows re-verified in a local two-client Playwright session.

VTT-AI-Agent (commented):
## TL;DR

All three findings from my previous review are cleanly fixed in the latest commit (overlay-id accuracy, self-rename announcement, no reload while disconnected), and CI including TestCafe is green. New pass found one notable gap: the new `.overlay { z-index: 10 }` puts every full-screen overlay *above* the status display (`z-index: 5`), so a player sitting in an input window or the game shelf gets no "Connection lost" / reload-countdown indication at all — and then a surprise auto-reload. A few smaller items below.

## Findings

### 1. Status display is hidden behind open overlays (`client/css/layout.css:736` vs `client/css/overlays/status.css:13`)

This PR adds `z-index: 10` to `.overlay` while `#statusOverlay` sits at `z-index: 5`. The old `connectionLostOverlay` was shown *forced* on top of everything; now, whenever any full-screen overlay is open — notably `buttonInputOverlay`, where a player may sit for minutes waiting on others — the connection-lost pulse, the "No response from server" message and the reload countdown are all invisible, and the 30-second `location.reload()` (`client/js/serverstate.js:646`) hits with zero warning, closing their input window. Since `#statusOverlay` is `pointer-events: none` and tiny, the simplest fix is to raise it above the overlays (e.g. `z-index: 20`, still below the 15-level elements if you prefer, or just above them) — at minimum for the `reconnecting`/`bad`/`reload` states.

### 2. Changes made during a short disconnect are reverted silently (`client/js/serverstate.js:522`)

With the blocking overlay gone (intentionally), players can keep moving pieces while the socket is closed; `toServer` drops those messages silently (`client/js/connection.js:70-73`) while `sendRawDelta` has already applied them locally. After reconnect, the fresh `state` reverts them — but the "your last changes could not be saved" notice only fires when the oldest pending delta is ≥ 5 s old. A delta attempted 2 s before a quick reconnect is reverted with no notice (pieces just snap back). `status.js` already knows a disconnect happened (`reconnecting`); consider passing that through (e.g. a `wasDisconnected()` that `receiveStateFromServer` checks) so the notice always shows when `pendingDeltas` is non-empty after a reconnect, keeping the 5 s threshold only for the zombie-connection / normal-state-broadcast case.

### Minor

- Closing an overlay now sends a coordinate-less mouse message `{ inactive: false, ... }` (`client/js/main.js:117`). On other clients this hits the "active" branch of the cursor handler (`client/js/overlays/players.js:125-127`), computing `translate(NaNpx, NaNpx)` (ignored by the browser) and flashing the player's cursor as `active` even though they didn't move. The `hidden: true` message had the same NaN quirk before this PR, but there the cursor is hidden; here it's visible. Guarding the else-branch on `args.mouseState.x !== undefined` (or only sending the message when an overlay *opens*) would clean this up.
- `setMyName` could `delete playerActivity[name]` (`client/js/overlays/status.js`): the mouse handler skips updates for one's own name and `setActivePlayersList` never purges active names, so if a race (e.g. around a rename, where `mouse` broadcasts under the new name can precede the `meta` update that calls `setMyName`) ever records a self-entry with `editMode: true`, it sticks forever and the player sees their own "X is in Edit Mode" message.
- `announcePlayerChanges` (`client/js/overlays/players.js:59`) reports one player leaving while another joins in the same update as "A renamed to B" when their client counts match. Cosmetic, and probably rare enough to ignore.

---

This PR touches engine files (`client/js/serverstate.js`), which are excluded from beta auto-deployment by design, so I did not add the `beta` label (it's also still a draft).

VTT-AI-Agent:
Addressed in bf80a0f3:

1. **Status hidden behind overlays** — fixed: `#statusOverlay` is now `z-index: 20`, above `.overlay` (10). The only 15-level rules are decorative separators inside `.modal` (their own stacking context), so nothing else competes. Verified with a screenshot: with the Players overlay open and the server dead, "Connection lost. Reconnecting..." renders on top in the corner.

2. **Silent revert for <5s pending deltas** — fixed: `serverstate.js` now tracks its own `disconnectedSinceLastState` flag (set via `onConnectionClose`, cleared on the next `state`), and the "Connection restored. Your last changes could not be saved." notice fires whenever pending deltas are discarded after a reconnect, regardless of age; the 5 s threshold now only applies to normal state broadcasts. This also replaces the `isDisconnected()` coupling from the previous fix — the countdown suppression uses the same local flag, avoiding any dependency on handler registration order in `status.js`. Verified through a TCP proxy cut (same server process, so the `serverStart` reload path stays out of the picture): a widget added *while* disconnected is reverted on reconnect **with** the notice; a delta that did reach the server before the cut is kept and correctly produces no notice.

3. **NaN cursor flash on coordinate-less mouse messages** — fixed: the cursor update branch is now guarded with `args.mouseState.x !== undefined`, which also covers the pre-existing `hidden: true` quirk.

4. **Stale self-entry in `playerActivity`** — fixed: `setMyName` now deletes any entry under one's own name.

On the last minor point (simultaneous leave+join with matching client counts read as a rename): agreed it's cosmetic and rare — two meta updates would have to collapse into one broadcast — so I left the heuristic as is.

Jest green; all flows re-verified locally with Playwright.

VTT-AI-Agent (commented):
# Review of PR #2891

## TL;DR

All findings from the previous two rounds are fixed in `bf80a0f3` (status z-index above overlays, revert notice after any reconnect, NaN-cursor guard, stale self-activity entry), and CI including both TestCafe runs is green. New pass found one notable edge: the reload escalation ages pending deltas by wall clock, so a suspended tab (phone lock, laptop sleep) with a pending delta can skip the whole 5/10/20 s escalation and hard-reload the instant the tab resumes — possibly before the `deltaConfirm` already sitting in the socket buffer is processed. A few smaller items below.

## Findings

### 1. Tab suspension skips the escalation and reloads immediately on resume (`client/js/serverstate.js:647-649`)

`updateConnectionMonitor` computes `oldestAge` from `Date.now() - sentAt`. If the tab is frozen right after a delta is sent (user locks their phone mid-game, laptop lid closes, Chrome freezes a background tab) and resumes more than 30 s later, the first monitor tick sees `oldestAge >= DELTA_CONFIRM_RELOAD_MS` and calls `location.reload()` with no icon, no message, no countdown. Two sub-cases:

- The connection was actually fine: the server confirmed the delta long ago, but the frozen tab never dispatched the message event. On resume, the overdue interval callback and the queued `deltaConfirm` message race — ordering isn't guaranteed — so a healthy session can get a spurious reload on phone unlock. (No data is lost since the delta *was* saved, but the reload is jarring and the fresh page discards e.g. an open input overlay.)
- The OS killed the socket during sleep without a `close` event yet: reload is a reasonable recovery, but it still fires with zero warning, whereas the design intends a 10-second countdown.

A cheap guard: track the time of the previous monitor tick, and when the gap is much larger than the 500 ms interval (tab was suspended), shift each `pendingDeltas[*].sentAt` forward by that gap (or just skip escalation for that tick). Suspended time then doesn't count against the server, and queued confirms get one tick to drain.

### Minor

- **New joiners see stale cursors as active/pressed** (`server/room.mjs:46-48`): `addPlayer` replays each player's full last `mouseState`, and since `sendMetaUpdate()` runs *before* `player.send('state', ...)`, the joiner has already built the cursor elements when the replay arrives — so every existing player's cursor flashes `active` (and stays `pressed` if their last broadcast was mid-drag) at a possibly minutes-old position. The only consumer that needs the replay is `status.js`; consider replaying just the activity subset (`editMode`, `activeOverlay`, `inactive`/`hidden`) instead of the whole mouse state.
- **`lastOverlay` restore is now stale-prone** (`client/js/connection.js:23-24`): overlays are no longer force-closed on disconnect, so on reconnect `showOverlay(null, true); showOverlay(lastOverlay)` closes whatever the user has open *now* and reopens whatever was open at disconnect time. Since the disconnect no longer disturbs the open overlay, the save/restore has lost its purpose — simplest is to drop it (or restore only when nothing is currently open).
- **Name-keyed maps vs duplicate names / special names**: `lastMouseStateByPlayer` (server) and `playerActivity` (client) key by player name while several clients can share one name — `removePlayer` (`server/room.mjs:858`) deletes the entry even when another same-named client remains, and last-writer-wins can flicker the edit/input status. Also, a player named `constructor` truthy-matches the replay guard `this.lastMouseStateByPlayer[other.name]` (`server/room.mjs:47`) via the prototype, sending joiners a `mouse` message whose `mouseState` serializes away and throws in the players.js handler. `Object.create(null)` (or a `Map`) avoids the prototype case. All cosmetic and consistent with existing name-keyed patterns, so judgement call.
- **Pure observers never detect a zombie connection**: escalation only starts when the player sends a delta, so someone just watching a stale board gets no warning until they act. That matches the stated data-loss goal (your first action triggers detection before more work piles up), just noting the scope boundary — a periodic ping would be the follow-up if viewer staleness ever matters.

---

This PR genuinely changes the engine — the client delta send/confirm path in `client/js/serverstate.js` and the server-side delta protocol in `server/room.mjs` — so it is excluded from beta auto-deployment by design and I did not add the `beta` label (it is also still a draft).

VTT-AI-Agent (commented):
## TL;DR

Solid, well thought out approach to a real problem (silent disconnects losing work) — the escalation timing, the deltaConfirm handshake, and the "moot delta" reconciliation on a fresh `state` all fit together correctly as far as I can trace them. Nothing I'd call a bug, but there's no test coverage at all for a change this central to client/server state sync, and this is squarely a real engine change (not incidental), so it's not a candidate for the `beta` label yet.

## Findings

1. **No test coverage for the new sync/escalation logic** (`client/js/serverstate.js`, `server/room.mjs`, `client/js/overlays/status.js`). This touches delta IDs, a new `deltaConfirm` message, and a forced `location.reload()` after 30s — exactly the kind of client-state-sync behavior `docs/testcafe.md` calls the E2E suite "the project's main safety net" for, and it's currently untested by both jest and TestCafe. The time-based escalation (5s/10s/20s/30s) is admittedly awkward to test directly, but at minimum a TestCafe scenario simulating a dropped/stalled connection and asserting the status overlay + eventual reload/message would catch regressions here that unit tests can't.

2. **Stale mouse state replayed to newly joined players** — `server/room.mjs` `join()` (~line 660-664) replays each other player's `lastMouseStateByPlayer` verbatim, including `pressed`. If player B was mid-drag (`pressed:true`) and then went idle without another mouse event, a newly-joining player C will render B's cursor as still pressed/active at a stale position indefinitely until B moves again. Cosmetic and low severity, but worth a quick fix (e.g. strip `pressed` before replay) since it's easy to trigger.

3. Everything else I traced checked out: `receiveInvalidDelta`/`triggerDeltaConflict` paths that skip `deltaConfirm` are always followed by a `state` message that unconditionally clears the client's `pendingDeltas`, so no delta gets stuck "pending" forever; `waitingForDeltaFromPlayer` still gets its `deltaConfirm` before being excluded from the `state` broadcast; and edit-mode/overlay changes now push an immediate `mouse` update (via the `showOverlay`/`toggleEditMode` call sites) rather than waiting on the next physical mouse move, so the new "X is in Edit Mode" / "X has an input window open" status won't lag.

## Beta label

Did not add the `beta` label: `client/js/serverstate.js` and `server/room.mjs` are genuinely modified engine behavior here (new `deltaConfirm` protocol message, delta-confirmation tracking, and an automatic page reload after 30s of unconfirmed deltas) — this is the core of the PR, not an incidental touch, so it stays excluded from the beta auto-deploy by design, and it's also still marked as a draft.

VTT-AI-Agent (commented):
## TL;DR

Re-review after the 2026-07-13 merge from main: the merge is clean (main had no commits touching any of this PR's files since the merge-base, so nothing could be dropped), CI is fully green (jest, both TestCafe matrices, CodeQL), and a fresh trace of the delta-confirm protocol found no new bugs. The four open items from the previous two review rounds are still unaddressed — most notably the missing test coverage and the tab-suspension instant-reload edge — since no code commit has landed after them.

## Findings

### Verified this round

- **Merge integration**: `6ae519e9` only brings in main-side changes to files this PR doesn't touch (e.g. `editmode.css`, `jsonedit.css` from #3015/#3025); `git log <merge-base>..main` shows zero main-side commits to the PR's files, so no conflict resolution could have lost anything.
- **Protocol completeness re-traced on the merged head**: every client delta send goes through `sendRawDelta` (`client/js/serverstate.js:465`), and every server path either sends `deltaConfirm` (`server/room.mjs:779`) or ends in a `state` message that clears `pendingDeltas` — including the conflict/invalid paths in `server/player.mjs` and the `waitingForStateConfirmation` drop window, which is always preceded by an in-flight `state`. The server-generated dropShadow-cleanup delta carries no `deltaSendId`, so no spurious confirm.
- **Version-skew during a rolling deploy degrades gracefully** both ways (old client ↔ new server: `deltaSendId` absent → confirm skipped; new client ↔ old server: unconfirmed deltas only matter for a transiently cached page).
- The pre-handshake `mouse` messages that `showOverlay` now sends during `connection.onopen` (before `toServer('room', …)`) are silently discarded server-side — `Connection` has no message handler until the `room` handshake creates the `Player` — so no error path there.
- Minify concatenation: the two new files follow the import-strip convention and introduce no top-level name collisions; load order in `server/minify.mjs` is correct.

### Still open from the previous rounds (no code change since)

1. **No test coverage** for the new sync/escalation logic (`serverstate.js`, `room.mjs`, `status.js`) — the deltaConfirm handshake and the 5/10/20/30 s escalation remain untested by jest and TestCafe (round-4 finding 1).
2. **Tab suspension skips the escalation**: `updateConnectionMonitor` still ages `pendingDeltas` by wall clock, so a tab frozen right after a delta send can hard-reload the instant it resumes, racing the queued `deltaConfirm` (round-3 finding 1).
3. **Stale active/pressed cursor replay to new joiners**: `addPlayer` (`server/room.mjs:46-48`) still replays the full last `mouseState`, and since `sendMetaUpdate()` runs before the replay, the joiner's cursor elements already exist and render minutes-old positions (rounds 3/4).
4. **`lastOverlay` save/restore in `client/js/connection.js:23-24`** has lost its purpose now that disconnects no longer close overlays — on reconnect it force-closes and reopens whatever is open (round-3 minor).

### Maintainer-rule notes

Nothing new is exposed to game authors (no widget property, no routine operation), so the property-editor/routine-block UI rules and the validator/`jeCommands` upkeep don't apply; the diff touches nothing under `library/`, and the feature is exercised by simply playing on the PR test server, so no tutorial-style demo room is required.

---

The `beta` label was already added by the maintainer on 2026-07-12 and I'm leaving it in place — nothing found here rises to crash or game-corruption risk (CI including CodeQL is green and the protocol degrades gracefully); note this is a genuine engine change (the delta send/confirm path in `client/js/serverstate.js` and `server/room.mjs`, not an incidental touch), so the beta builder's engine exclusion still applies unless the maintainer allow-lists it.

VTT-AI-Agent:
Addressed all four open items from rounds 3–5 in 26b4d5b4:

1. **No test coverage for the sync/escalation logic** — the unconfirmed-delta bookkeeping and the escalation decision moved out of `serverstate.js` into `client/js/deltamonitor.js`, free of DOM access and timers so the whole state machine can be driven with an injected clock. `tests/client/deltamonitor.test.js` (20 jest tests, 100% coverage of that file) covers the 5/10/20/30 s escalation, the countdown value, out-of-order confirmations, the "closed socket suppresses escalation and reload" rule, when a fresh state does and doesn't report lost changes, and the suspension guard below. `serverstate.js` is left with applying the resulting status and doing the reload.
   `tests/testcafe/connection.js` covers the real handshake in the browser: it patches the websocket so `delta` messages can be held back while the connection stays open and the server keeps answering, then asserts that a delivered delta gets a `deltaConfirm` and never raises a warning, that a held-back one escalates to the icon and then to "No response from server.", and that releasing it clears the status again. Added to both TestCafe workflow matrices; runs in ~30 s.

2. **Tab suspension skipping the escalation** — the monitor interval now measures the gap between ticks. A gap far larger than the 500 ms interval means the tab was frozen, so every pending delta's `sentAt` is shifted forward by that gap (capped at "now") and the escalation is skipped for that one tick. Resuming therefore continues with exactly the age the delta had before the freeze — a healthy connection gets no reload at all and its queued `deltaConfirm` gets a tick to drain, while a genuinely dead one still gets the full 10-second countdown instead of an instant reload. The reload can only be triggered from that tick, so no other code path can reload straight after a resume.

3. **Stale cursor replay to new joiners** — `addPlayer` now replays only the activity subset (`inactive`, `editMode`, `activeOverlay`) instead of the whole last mouse state, so a joiner no longer renders every other player's cursor as active/pressed at a minutes-old position while still learning who is in Edit Mode or has an input window open. Verified in a two-client session: the joiner sees the other cursor with no `active`/`pressed` class at its default off-screen position, and the status overlay still says "Guest… is in Edit Mode". `lastMouseStateByPlayer` is a null-prototype object now, which also takes care of the `constructor`-named-player case from the same finding.

4. **Obsolete `lastOverlay` restore** — dropped from `connection.js`. Verified through a TCP proxy cut with the Players overlay open: the overlay stays open through the disconnect ("Connection lost. Reconnecting...") and the reconnect, instead of being force-closed and reopened.

On the remaining minor points I left alone: duplicate player names in the name-keyed activity maps stays consistent with the existing name-keyed patterns, and pure observers not detecting a zombie connection remains the intended scope (detection starts with your first action, before more work piles up).

Jest and the TestCafe suite (new `connection.js` plus `editor.js`) pass locally, and the minified bundle was checked with `minifyJavascript: true` since a new file was added to `server/minify.mjs`.

VTT-AI-Agent:
Merged the latest `main` (up to `16f942f0`) into this branch and refitted the status display to what changed there.

> Merge changes from main and check if any of them need to be addressed in this PR.

— [arnoldsmith86 on Discord](https://discord.com/channels/770758631146782780/1525221714685071523/1543699427636805642)

Six weeks of `main` had landed since the last merge, and two of those changes really did need addressing:

**1. The Players tab became a session table (#3013).** `client/js/overlays/players.js` was rewritten around a `sessions` list (`{ sessionID, player }`), because several clients can now legitimately share one player name — hotseat local players, a second tab, the new "view" button. My join/leave/rename announcements were built on the deduplicated `activePlayers` name list and a count heuristic, which no longer holds:

* announcements now diff the **session list** of a meta update against the previous one. A session that disappears is a leave (unless another session still holds that name), a new `sessionID` is a join, and the same `sessionID` under a different name is a rename — exact, instead of guessing from name counts.
* renaming a single session (`view`, or "add a player who shares this device") is now reported correctly as "X renamed to Y" for the other clients, and skipped on the tab that did it, which already says "You renamed to Y".

Verified with two clients: join, a rename through the players table, and leave each produce exactly one correct message, and the renaming tab shows only its own notice.

**2. The client bundler was rewritten (#3086, #3089, #3132).** `server/minify.mjs` no longer has inline file lists — `status.css`, `status.js`, `deltamonitor.js` and `overlaystate.js` moved into the new `CLIENT_CSS`/`CLIENT_JS` constants, in the right load order, and `connectionlost.css` came out. Checked with `MINIFYJAVASCRIPT=true node server/prebuild.mjs`: terser's top-level mangling reports no name collisions and the bundle runs clean in a browser.

Two smaller things while resolving the conflicts:

* **The server now remembers the last mouse state per session object instead of per player name** (a `Map` keyed by the `Player`). This was the third bullet of the round-3 "name-keyed maps vs duplicate names" finding, which mattered much more once `main` made duplicate names a supported feature: a leaving session no longer wipes the activity of another session with the same name, and a rename needs no bookkeeping at all.
* **A detected server restart says so again.** `connection.js` schedules its reload with a random delay of up to ten seconds; the full-screen connection-lost overlay used to cover that wait, and since this PR removes that overlay the page just sat there looking frozen. The corner status display now shows "The server was restarted. Reloading the page…" for those seconds.

The players button's yellow flash on a player change is gone — the join/leave message in the status display says the same thing, with names.

Verified locally: jest (1543 tests, 50 suites), TestCafe `connection.js` and `editor.js` (89 tests), `multiclient.js`, and a Playwright walk-through of join/rename/leave, edit-mode status and a real server restart at three viewport widths with no page errors. `domsnapshot.js` fails on an unrelated pre-existing flake — the spinner widget hides its value for 1300 ms after a change, so a different legacy-mode combination races the snapshot on each run; nothing this PR touches is inside `#topSurface`.

VTT-AI-Agent (commented):
## TL;DR

All four items left open by the previous review rounds are now addressed — the escalation logic moved into a testable `deltamonitor.js` with 20 passing jest tests plus a real-browser TestCafe test wired into both CI matrices, a suspended tab now gets its frozen time credited instead of an instant reload, joiners no longer receive stale cursor replays, and the obsolete `lastOverlay` juggling is gone. I traced the new code and found only one small cosmetic issue in the join/leave announcements; everything load-bearing checks out and CI is fully green.

## Verified this round

- **`client/js/deltamonitor.js` escalation and suspension logic**: the `Math.min(now, sentAt + gap)` clamp correctly handles a delta sent between tab-resume and the first monitor tick, and returning `null` on the suspension tick gives queued `deltaConfirm` messages one tick to be processed before escalation resumes. The reload can only be triggered from the interval in `serverstate.js` (via `monitorTick`), never from the event-driven `connectionStatus()` calls — so a fresh `stateReceived`/`deltaConfirmed` can never race a reload decision. A tick gap caused by a busy main thread (>2 s) also counts as suspension, which is the conservative direction (the client couldn't have processed confirms either).
- **Chrome's intensive background-tab throttling** (timers coalesced to ≥1/min after ~5 min hidden) makes every tick register as a suspension, so a long-hidden tab never escalates or reloads while hidden and starts over on refocus — consistent with the PR's stated goal of never landing a wake-up on an unannounced reload.
- **Bundle safety**: scripted check across all `CLIENT_JS` files found no top-level declaration collisions from the new `deltamonitor.js` / `overlaystate.js` / `status.js` in the concatenated room bundle.
- **Server paths**: every delta path in `server/player.mjs`/`room.mjs` either sends `deltaConfirm` or ends in a `state` message that clears the client's pending list (conflict, `waitingForStateConfirmation` drop window, `waitingForDeltaFromPlayer`), and the client sends `confirm` synchronously inside `receiveStateFromServer` before any new delta can be produced, so no pending delta can get stuck.
- **The joiner replay** (`server/room.mjs:60-64`) runs after `sendMetaUpdate()`, so the joiner's cursor elements exist, and sends only `{ inactive, editMode, activeOverlay }` — the stale active/pressed cursor finding from earlier rounds is fixed, and `players.js`'s new `x !== undefined` guard also removes the old `translate(NaNpx, NaNpx)` path for `{hidden: true}` messages.
- `toggleEditMode` ends in `showOverlay()`, which now broadcasts `editMode` immediately — the "X is in Edit Mode" status doesn't wait for the next mouse move.
- The switch from `setTimeout(location.reload, …)` to an arrow function in `client/js/connection.js:60` fixes a real latent bug: an unbound `location.reload` throws `Illegal invocation` when the timer fires, so the scheduled post-restart reload never actually happened before.
- `z-index: 10` on `.overlay` only affects `#roomArea` siblings; the sole behavior change is that player cursors now paint below open overlays, which reads as intended.
- `tests/client/deltamonitor.test.js` passes locally (20/20), and the full CI including the new `connection.js` TestCafe job is green in both browsers.

## Finding

1. **Self-"joined" message after every reconnect, and asymmetric second-tab announcements** — `announceSessionChanges` (`client/js/overlays/players.js:123-141`): session IDs are per-connection (`Player.nextSessionID++`), and since this PR keeps the page alive across reconnects, `prevSessions` survives a websocket blip. After the automatic reconnect the meta update contains the player's own *new* sessionID, the join branch has no `mySessionID` or name-based suppression, so the player sees "*TheirOwnName* joined" about themselves after every brief disconnect. Relatedly, opening a second tab announces "X joined" to everyone even though X was already present, while closing it is (correctly) silent thanks to the `!currentNames.has(player)` guard on the leave side. Both would be fixed symmetrically by suppressing joins whose name already existed: build `const previousNames = new Set(previous.values());` and announce a join only `if(!previousNames.has(player))`. Cosmetic, but reconnect blips are exactly the situation this PR polishes, so the noise shows up where users will look.

## Maintainer-rule notes

Nothing new is exposed to game authors (no widget property, no routine operation), so the property-editor/routine-block UI rules and validator/`jeCommands` upkeep don't apply; the diff touches nothing under `library/`, and the feature is exercised by simply playing on the PR test server, so no tutorial-style demo room is required.

---

The `beta` label was already present (added by the maintainer) and I'm leaving it in place — the one finding above is cosmetic and nothing rises to crash or corruption risk; note this remains a genuine engine change (the delta send/confirm path in `client/js/serverstate.js` and `server/room.mjs`), so the beta builder's engine exclusion still applies unless a maintainer allow-lists it.

VTT-AI-Agent (approved):
Approving on the maintainer's request.

Six review rounds on this branch and every finding they raised is now fixed: the escalation logic lives in a testable `client/js/deltamonitor.js` with 20 jest tests plus a real-browser TestCafe job wired into both CI matrices, a suspended tab gets its frozen time credited instead of an instant reload, joiners no longer receive stale cursor replays, and the overlay/z-index and reconnect-notice issues from the earlier rounds are all resolved. CI is fully green (jest, both TestCafe matrices, CodeQL), and the delta send/confirm protocol traces cleanly on both sides — every server delta path either sends `deltaConfirm` or ends in a `state` message that clears the client's pending list, so no pending delta can get stuck.

One cosmetic item is still open from my last review and is not a blocker: `announceSessionChanges` (`client/js/overlays/players.js`) announces a player's own "joined" message after every reconnect, and announces a join for a second tab of an already-present player, because session IDs are per-connection while the page now survives a blip. Suppressing joins whose name already existed (`const previousNames = new Set(previous.values());`) fixes both sides symmetrically.

Note this is still marked as a draft, so it needs to be marked ready for review before it can be merged.

*Approval posted at a maintainer's explicit request — by `claude-opus-5`*

VTT-AI-Agent:
Addressed the one finding from the latest review in `bdd9d597`.

**Self-"joined" message after a reconnect, and asymmetric second-tab announcements** — fixed symmetrically, as suggested: the join branch now only announces a session whose player name was not already present in the previous session list, which is the mirror image of the existing leave rule (`!currentNames.has(player)`). A reconnect therefore produces neither a leave nor a join on the reconnecting tab, and opening a second tab of the same player is as silent as closing it.

The diff logic moved into an exported pure function `sessionChangeMessages(previous, current, ownSessionID)` so it can be unit tested without the DOM; `tests/client/session-announcements.test.js` covers join, leave, rename (including the "not on the tab that renamed itself" case), reconnect under a new session ID, second tab open/close, the last session of a name leaving, and several changes collapsing into one message.

Verified in the browser with two clients on my own server (websocket closed from the page so the real reconnect path runs): after the reconnect the reconnecting tab shows nothing, before the fix it said "Alice joined"; a second tab of the same player is silent on open and on close; a genuinely new player still gets "Bob joined" and "Bob left".

One thing I deliberately left alone: the *other* players still see "Alice left; Alice joined" when Alice's connection blips, because from their side that is exactly what happened — her session really was gone for those seconds. Suppressing it would mean deferring every leave announcement by a few seconds to see whether the name comes back, and knowing that someone's connection dropped seems like information worth having in a PR about noticing disconnects. Happy to add the deferral if you'd rather have the quiet version.

The two remaining minor points from earlier rounds stay as they were: name-keyed activity for duplicate names is consistent with the existing patterns, and pure observers not detecting a zombie connection is the intended scope.

Jest (52 suites, 1555 tests) and TestCafe `connection.js` + `multiclient.js` pass locally.

VTT-AI-Agent:
Merged the latest `main` (up to `79d779d9`) and refitted the connection notices to what landed there.

The one commit that overlapped is #3147, "Reload a page whose server restarted instead of loading edit mode from the new build". It gave the connection-lost overlay a second set of wording for a detected restart, with a "Reload now" button, and a path that puts it up when the editor bundle turns out to belong to a build the server no longer serves. That is the same situation this branch had been covering with a line in the corner status display, since the full-screen overlay was gone here.

Resolved in favour of the overlay for that case:

* **A dropped connection** is still the corner status display ("Connection lost. Reconnecting…") — that is the case this PR is about, and a blip must not blank the board.
* **A detected restart** keeps main's full-screen notice. That page is not getting its connection back and is reloading within ten seconds, so the corner-status argument does not apply, and "Your room is not affected" plus a Reload now button says more than a line in the corner can. The corner display now goes quiet while the notice is up, instead of contradicting it with "Connection lost. Reconnecting…" — which it would have done whenever the drop was noticed before the restart was.
* The "connection lost" half of that overlay no longer has a caller, so the overlay is now what it is used for: `#serverRestartOverlay` / `client/css/overlays/serverrestart.css`, with the restart wording, the illustration and the button, and without the class that used to switch between the two states.

Both sides' fixes for the unbound `location.reload` were the same fix, and main's backoff reset on the `serverStart` handshake, its container-query sizing for the illustration and its `editModeUnavailable` classification came along unchanged. `tests/testcafe/connection.js` runs next to main's new `gameshelf.js` in both matrices.

Verified: jest (56 suites, 1650 tests), TestCafe `connection.js`, `routines.js` and `multiclient.js` locally, the minified bundle with `MINIFYJAVASCRIPT=true`, and a real server restart driven in Chromium — the corner says "Connection lost. Reconnecting…" while the server is away and hands over to the "The server was updated" overlay the moment the reconnect detects the new build, with no page errors. CI is green on all 51 checks; two Firefox jobs needed a re-run for a runner DNS failure and a slow room load, both unrelated.

PR #2953, opened 2026-03-27
Fix: Library Scanning Resilience and Recursion Support

mousewax:
This PR addresses an issue where the "Dev" and "Unused" library categories would appear empty if a game was missing metadata or stored in a sub-folder.

### Key Changes:
- Resilience: Modified getPublicLibraryGames in server/room.mjs to prevent crashes when _meta.info is missing. It now falls back to the directory name.
- Recursion: Added recursive scanning for library categories, allowing for better organization of external game folders.
- Maintenance: Added .vscode/ to .gitignore to keep local editor settings out of the repo.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2953/pr-test (or any other room on that server)


96LawDawg (approved):
I don't really understand this, but I want mousewax's server to work. This does not seem to impact the public library operation on production in any way. I haven't tested anything on a local machine. @ArnoldSmith86 should still look at this, but I'm going to approve it and @mousewax can merge when he thinks it is ready.

PR #2955, opened 2026-03-28
fixes exclude property example in inheritFrom tutorial

RaphaelAlvez:
the 4th example on the left states "does not inherit css" but the JSON was just a copy from another example

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2955/pr-test (or any other room on that server)


96LawDawg (approved):
Approved.

PR #2966, opened 2026-04-17
Add iconColor property to holders

96LawDawg:
This adds an iconColor property to holders. If used, it overrides anything set in the advanced icon property for both color and opacity.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2966/pr-test (or any other room on that server)


ArnoldSmith86:
Doesn't `textColor` do that?

96LawDawg:
Not sure exactly what you mean because there is no advanced text property like there is an icon property. However, I did just realize the if you use both `textColor` and `iconColor`, `textColor` was not working, so I will fix that.

PR #2969, opened 2026-04-28
Widgets with display:false are no longer valid dropTargets

96LawDawg:
This adds a check to ensure that holders are visible before allowing them to accept dropped children. Functions like MOVE and SET are not impacted.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2969/pr-test (or any other room on that server)

### DIFF
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -39,7 +39,7 @@ function compareDropTarget(widget, t, exclude){
 function getValidDropTargets(widget) {
   const targets = [];
   for(const [ _, t ] of dropTargets) {
-    if(t.isInvisible())
+    if(!t.isVisible())
       continue;
 
     // if the holder has a drop limit and it's reached, skip the holder
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -39,7 +39,7 @@ function compareDropTarget(widget, t, exclude){
 function getValidDropTargets(widget) {
   const targets = [];
   for(const [ _, t ] of dropTargets) {
-    if(!t.get('display'))
+    if(t.isInvisible())
       continue;
 
     // if the holder has a drop limit and it's reached, skip the holder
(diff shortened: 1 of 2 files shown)


ArnoldSmith86 (approved):

ArnoldSmith86 (changes requested):
Actually, no. This doesn't cover the case if one of the ancestors is display: false and at the same time I'm pretty sure I'm doing that on Runway Roll for the modules and I'm very sure that I can't drop dice into invisible holders.

96LawDawg:
> This doesn't cover the case if one of the ancestors is display: false

I'll look into that

> I'm pretty sure I'm doing that on Runway Roll for the modules and I'm very sure that I can't drop dice into invisible holders.

Perhaps you are. Hard for me to figure out in that game. I tested dice in an empty room. They get parented to a hidden holder just like anything else.

ArnoldSmith86:
OK, it totally happens in that game.. Change your code to do `t.isInvisible()` and define that function in the Widget class to do `!this.get('display') || this.get('parent') && widgets.has(this.get('parent')) && widgets.get(this.get('parent')).isInvisible()`..

ArnoldSmith86:
Huh, the new `isInvisible` is now just below an existing `isVisible`? :grin:

_Maybe_ we should just use that one...

96LawDawg:
> Huh, the new `isInvisible` is now just below an existing `isVisible`? 😁
> 
> _Maybe_ we should just use that one...

The latest change does that and it works for this purpose. The only other place that `isVisible` is used is in line 178 of mousehandling.js. That relates to hotkeys. I tested the change for when the widget is hidden and a parent widget is hidden and used a hotkey. There is no difference between production and this PR, so this should be good and not cause a breaking change.

But feel free to revert the last commit if you are concerned.

ArnoldSmith86:
I don't think the `isVisible` needs any changes for this to work..

96LawDawg:
> I don't think the `isVisible` needs any changes for this to work..

Doh! I didn't think of that. `isVisible` does it a different way, but does the same thing. Now the only code change is a check using the exisiting `isVisible` when looking for valid drop targets.

ArnoldSmith86 (approved):

PR #2986, opened 2026-05-27
Validator - Fix CALLs in cards

### LINKED ISSUE #2984: Validator does not recognize CALLs to a deck routine from a card via cardDefaults or cardTypes
EDITED to reflect the actual problem...

A routine defined directly on a deck is reported as "not being called" by the validator when the caller is a routine defined in the deck's `cardDefaults` or `cardTypes` (on a card, in other words).

Any routine that calls the deck's routine, defined anywhere else (and it doesn't even have to refer to the correct routine via a "widget" parameter), will make the warnings go away.

Here is a room with a stripped down example: https://virtualtabletop.io/issue2984. If you rename the Squelcher widget's routine to `clickRoutine`, then the warnings go away.

Everything works as expected (the chit will twirl when you click it), but the validator can't understand it.

The validator output:
```
deck-chits 	TwirlRoutine 	            Routine is not being called
deck-chits 	TwirlRoutine.0.collection 	Collection 'caller' is undefined.
deck-chits 	TwirlRoutine.5.collection 	Collection 'caller' is undefined.
deck-chits 	TwirlRoutine.7.collection 	Collection 'caller' is undefined.
```
(I believe the "Collection 'caller' is undefined." is a side-effect of this issue. The output is either all there as above, or it is all not there.)

96LawDawg:
Fixes #2984.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2986/pr-test (or any other room on that server)

### DIFF
--- a/validator/validate_gamefile.js
+++ b/validator/validate_gamefile.js
@@ -517,9 +517,30 @@ function validateRoutine(routine, context, propertyPath = []) {
                 for(const key of Object.keys(operation.arguments))
                     newContext.validVariables[key] = 1;
             newContext.validCollections['caller'] = 1;
-            for(const widget of Object.values(context.widgets))
-                if(Array.isArray(widget[operation.routine]))
-                    problems.push(...validateRoutine(widget[operation.routine], Object.assign(newContext, {widgetId: widget.id}), [operation.routine]));
+            const checkedRoutines = new Set();
+            for(const widget of Object.values(context.widgets)) {
+                let routine = widget[operation.routine];
+                let widgetId = widget.id;
+                let routinePath = [operation.routine];
+                if(!Array.isArray(routine) && widget.type === 'card') {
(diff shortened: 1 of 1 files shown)


96LawDawg:
This CALL
```
   {
        "func": "CALL",
        "routine": "doubleClickRoutine",
        "widget": "cabDeck",
        "arguments": {
          "parent": "${PROPERTY parent}"
        }
      }
```
gets reported as using an undefined variable "parent" in the doubleClickRoutine. You can quite the validator by doing this useless bit: "var parent = ${parent}", but that should not be necessary. Example is in London Detectives (game submitted in https://github.com/ArnoldSmith86/virtualtabletop/pull/3021) in widget "cabDeck".

The current PR does not fix this aspect of CALLs from decks.

PR #2988, opened 2026-05-28
Fix recursive ignoreZoom

### LINKED ISSUE #2987: Recursive ignoreZoom: true
If a widget has ignoreZoom: true and it is parented to a widget that also has ignoreZoom: true, the child widget actually gets smaller when you zoom in. The behavior is reversed and it should not be impacted at all.

96LawDawg:
Fixes #2987

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2988/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

The fix is correct and well-scoped: nested `ignoreZoom` widgets now skip the inverse-zoom compensation when an ancestor already compensates, and the new update hooks keep descendants in sync when `ignoreZoom` or `parent` change. I traced the relevant update paths and found no correctness issues — only a couple of minor observations below.

## Details

I verified the main interaction points and they all hold up:

- **State vs. DOM ordering** — `StateManaged.applyDelta` writes `this.state` before calling `applyDeltaToDOM` (`client/js/statemanaged.js:24-28`), so `hasIgnoreZoomAncestor()` sees the new `parent` value when `applyCSS(delta)` runs, and the new hook at `widget.js:247` runs after the reparent block updates `this.parent`/`childArray`.
- **Creation order** — during full state load, `addWidget` defers children until their parent exists (`serverstate.js:53-57`), so a child's first `cssTransform()` always sees correct ancestor state; within a `receiveDelta` batch, either ordering works because the parent's hook refreshes descendants that applied first.
- **Inherited `ignoreZoom`** — flows through `applyInheritedDeltaToDOM` → `applyDeltaToDOM`, so the new hook fires for inheritance updates too.
- **Toggling off** — clearing `ignoreZoom` on the ancestor correctly restores compensation on descendants, since `refreshIgnoreZoomDescendants` recomputes and the ancestor walk then returns false.
- **Limbo** — a widget whose parent chain is broken (`widgets.has` fails) falls back to self-compensation, which matches its DOM placement at top level. Consistent.

Minor observations (no change strictly needed):

1. **Hot-path overhead, mostly harmless**: because `parent` was added to `cssTransformProperties()` and the hook fires on every `delta.parent`, every card move/drop now recomputes the transform via the new-parent branch and does a (usually empty) recursive `refreshIgnoreZoomDescendants()` walk, and every `cssTransform()` call now walks the ancestor chain. All of these are O(tree depth) with trivial constants, so this is fine in practice — just noting it since `cssTransform` runs during drags. Also, since `defaults` include `ignoreZoom: false`, the hook fires as a no-op on every widget creation.

2. **Parent cycles** would make `hasIgnoreZoomAncestor()` (widget.js:715) loop forever, but `coordGlobalFromCoordParent` already recurses unboundedly on cycles, so this isn't a new exposure — just something to keep in mind if cycle protection is ever added.

3. **No automated test** — understandable, since the TestCafe suite compares room-state hashes and this change is purely client-side rendering (CSS transforms), which the hash can't capture. Manual verification on the PR test server is the practical check here: parent + child both with `ignoreZoom: true`, zoom in, confirm the child no longer shrinks and follows the parent 1:1.

VTT-AI-Agent (commented):
## TL;DR

The follow-up commit (43913edc) correctly applies the hot-path optimization suggested in the previous review: `'parent'` is now only included in `cssTransformProperties()` for widgets with `ignoreZoom`, so ordinary card moves/drops no longer trigger a redundant transform recompute. I re-traced the affected paths and found no new issues — the PR looks ready.

## Details on the new commit

I verified the gating in `cssTransformProperties()` (client/js/widgets/widget.js:702-711) is safe in the corner cases:

- A widget clearing `ignoreZoom` in the same delta as a reparent still recomputes its transform, because `'ignoreZoom'` itself stays in the list (and state is written before `applyCSS` runs, so the gate reads the new value consistently either way).
- A non-`ignoreZoom` widget carrying `ignoreZoom` descendants that gets reparented is still handled: the unconditional hook at widget.js:247-248 fires on any `delta.parent` and `refreshIgnoreZoomDescendants()` recursively refreshes the descendants that actually compensate.
- Zoom-level changes are unaffected — `zoom.js:83-84` refreshes every `ignoreZoom` widget directly via `applyCSS({ ignoreZoom: true })`.

The earlier findings stand: the core fix (skipping inverse-zoom compensation when `hasIgnoreZoomAncestor()`) is correct, and the remaining notes from the previous review (no cycle guard in the ancestor walk — matching existing code like `coordGlobalFromCoordParent`; no automated test — not capturable by the state-hash TestCafe suite) were explicitly non-blocking. Nothing further from my side.

96LawDawg:
Tested live in browser. There is a problem. If a parent widget is ignoreZoom: false, but a child widget is ignoreZoom:true, then neither widget gets zoomed in. Both should because the parent's ignoreZoom setting should control.

VTT-AI-Agent:
Good catch — fixed in 026d19f8.

The root cause: the inverse-zoom compensation in `cssTransform()` is computed in the **room's** coordinate frame, but a nested widget's DOM element lives inside its parent's element and therefore in the parent's frame. So self-compensating a nested widget was mathematically wrong regardless of the ancestor's setting — which is exactly your `ignoreZoom:false` parent + `ignoreZoom:true` child case.

The fix makes compensation **top-level only**: a widget only applies the inverse-zoom transform when it sits directly on the surface. A nested widget now just inherits its parent's zoom behaviour, so the parent's `ignoreZoom` effectively controls it:
- parent `false`, child `true` → child zooms along with the parent ✅ (your report)
- parent `true`, child `true` → child stays put inside the compensating parent ✅ (original #2987)

This also let me replace the ancestor-chain walk with a simple top-level check and drop the descendant-refresh hook, since a nested widget's compensation no longer depends on ancestor state. Ready for another look on the PR test server.

96LawDawg:
Live tested in browser. Everything works. Ready to merge.

VTT-AI-Agent (approved):
Approved as requested. Not merging yet.

ArnoldSmith86 (approved):

PR #2999, opened 2026-06-21
Public Library - Add Antebellum Rails

96LawDawg:
Similar to Gulf, Mobile & Ohio. VTT implementation by Nestor and Jonathan

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-2999/pr-test (or any other room on that server)


ArnoldSmith86 (approved):
- Is "Creative Commons" enough or does it have to be more specific?
- Is it intentional that there are two map assets?

96LawDawg:
> - Is "Creative Commons" enough or does it have to be more specific?

It should be more specific but that is all there is on the map asset.

> - Is it intentional that there are two map assets?

Yes. There are minor differences between base and expansion maps.

PR #3004, opened 2026-07-06
Never let an unusable statistics.json take the server down

### LINKED ISSUE #2993: Error when starting docker: `SyntaxError: Unexpected end of JSON input`
When attempting to start my docker instance of virtualtabletop, I'm now getting a repeating error that reads:
```
Node.js v18.20.8
undefined:1
SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at new Statistics (file:///app/server/statistics.mjs:9:58)
    at file:///app/server/statistics.mjs:40:16
    at ModuleJob.run (node:internal/modules/esm/module_job:195:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:337:24)
    at async loadESM (node:internal/process/esm_loader:34:7)
    at async handleMainPromise (node:internal/modules/run_main:106:12)
```

Any idea of how to resolve this? Is there any more info I can provide?

VTT-AI-Agent:
## Problem

An unusable `save/statistics.json` took the server down on startup. `JSON.parse` ran unguarded in the `Statistics` constructor, so an empty or truncated file threw before the server ever listened, and the process manager restarted it into the same crash:

```
SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at new Statistics (file:///app/server/statistics.mjs:9:58)
```

Closes #2993.

## Fix

Statistics are a nice-to-have: they may not keep the server from starting, and they may not take it down while it runs either. Neither reading nor writing the file throws any more.

**Reading** happens in `readFromFilesystem()`, which is unconditionally non-throwing — the property you want from something that runs at import time:

| file | result |
| --- | --- |
| missing / unreadable | empty statistics (missing is silent, unreadable logs a warning) |
| empty or whitespace only | empty statistics |
| unparseable, or JSON that is not shaped like statistics | file moved to `statistics.json.corrupt`, warning logged, empty statistics |
| valid | parsed, with any missing top level key filled in |

The `existsSync` check is gone with it: it raced with the read, so a file disappearing in between (cleanup script, volume remount) reproduced the very crash loop this fixes.

**Writing** is guarded the same way. `autosaveRooms()` calls `Statistics.writeToFilesystem()` from a `setInterval` outside its `try`, and the `exit`/`SIGTERM` handler calls it with no `try` at all, so a save directory that cannot be written killed the process 60 seconds after startup — the same crash loop as #2993, just delayed, and with a stack pointing at the writer rather than at the statistics. It now logs and carries on; only the transition between working and failing is logged, so a permanently unwritable directory does not fill the log every minute.

Two things beyond just not crashing:

- **The broken file is preserved.** The autosave writes statistics every 60 seconds, so falling back to empty statistics used to mean the corrupt file was overwritten with `{}` a minute after startup — every star and every recorded play time gone, while the warning that explained it had already scrolled past. A file truncated mid-write is usually still mostly recoverable by hand, so it is renamed to `statistics.json.corrupt` before the fallback takes over. A later corruption does not overwrite that copy; the names continue with `.corrupt.2`, `.corrupt.3` and so on.
- **Wrong-shape JSON is caught too.** `null`, `[]`, `42` and `{"starsPerState":42}` all parse fine and used to fail much later — a `TypeError` in `updateDataInsideStates()` on the first public-library listing, i.e. a 500 on the main page instead of a loud startup crash. The check reaches down to the numbers the class actually reads, so `{"timePerState":{"a":42}}` is caught as well, and the sum of played minutes is seeded with `0` so an empty player count map means "never played" instead of `Reduce of empty array with no initial value`.

## Testing

`tests/server/statistics.test.js` covers the missing, empty, whitespace, truncated, non-object, wrong-shape (top level and nested), unreadable and valid cases, the `.corrupt` name that is already taken, an unusable file that cannot be moved aside, and a target that cannot be written — plus that statistics keep being collected after a fallback and are saved again once writing works. The class is exported next to the singleton so a test can point it at a temporary directory.

Also verified against a running server. With `save/statistics.json` replaced by a directory it starts normally, serves through several autosave cycles instead of dying at the first one, and logs each problem once:

```
WARNING - .../save/statistics.json, starting with empty statistics: EISDIR: illegal operation on a directory, read
WARNING - could not write .../save/statistics.json, statistics are collected but not saved: EISDIR: …
```

With a truncated file the `.corrupt` copy still holds the original bytes, and `{"timePerState":{"lib1":{}}}` now lists the public library instead of returning a 500.

`npm test` passes (42 suites, 1381 tests). No widget property, routine function or library file is touched, so there is nothing to add to the editor UI or the validator.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3004/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

The fix is correct and minimal for the reported crash (#2993) — empty and unparseable `statistics.json` no longer take the server down. The one thing I'd change before merging: the corrupt file is **silently overwritten with empty statistics within 60 seconds** of startup by `autosaveRooms()`, so all stars/playtime are destroyed even though a truncated file is usually mostly recoverable — rename it aside first.

## Findings

### 1. Falling back to defaults destroys the (usually recoverable) old data — `server/statistics.mjs:31`

`autosaveRooms()` in `server.mjs:98` calls `Statistics.writeToFilesystem()` every 60 seconds, and the exit handler (`server.mjs:621`) calls it again. So after this fix the sequence is:

1. server starts, `statistics.json` is truncated → warning logged, `this.data` = empty defaults
2. ≤60 s later the autosave overwrites `statistics.json` with `{"starsPerState":{},"timePerState":{}}`
3. every star and every recorded play-time for the whole instance is gone, unrecoverably

The realistic corruption case is a write killed part-way through, which leaves most of the JSON intact — an admin could recover nearly all of it by hand, but only if the bytes still exist. Since the warning scrolls past in the log and the file looks fine a minute later, nobody will notice until the numbers are already zero.

Suggestion: move the file aside in the `catch` before returning defaults, e.g.

```js
} catch(e) {
  const corruptFilename = `${statisticsFilename}.corrupt`;
  fs.renameSync(statisticsFilename, corruptFilename);
  Logging.log(`WARNING - could not parse ${statisticsFilename}, moved it to ${corruptFilename} and starting with empty statistics: ${e.message}`);
  return defaults;
}
```

(`fs.renameSync` is already used for this kind of thing in `server/room.mjs:267`.) The empty-file branch matters much less, but moving it aside there too costs nothing and keeps the two paths symmetric.

### 2. Valid-but-wrong-shape JSON still crashes later, just further from the source — `server/statistics.mjs:33`

`JSON.parse` succeeding doesn't mean the result is a usable statistics object. `null`, `[]`, `"x"`, `42` and `{}` all parse fine, and then:

- `this.data = null` → `updateDataInsideStates()` throws `TypeError: Cannot read properties of null (reading 'starsPerState')` on the first public-library listing, i.e. a 500 on the main page rather than a startup crash — arguably worse to diagnose than the current loud crash-loop.
- `this.data = {}` → `toggleStateStar()` throws on the first star click.

Truncation is unlikely to produce these, but a hand-edited or partially zero-filled file can, and the guard is one line:

```js
const parsed = JSON.parse(raw);
if(typeof parsed != 'object' || parsed === null || Array.isArray(parsed))
  throw new Error('not an object');
return { ...defaults, ...parsed };
```

The spread also covers the case where a future version adds a third top-level key and old files lack it. As written, the PR closes exactly one of the ways this file can be broken; this closes the class.

### 3. `existsSync` + `readFileSync` is the same failure mode the PR is fixing — `server/statistics.mjs:23-28`

If the file disappears between the `existsSync` check and the read (a cleanup script, a container volume remount), `readFileSync` throws `ENOENT` and the server crash-loops again — the very symptom of #2993. Putting the read inside the existing `try` and treating any read error as "no statistics" removes the check, the race and a branch at once:

```js
let raw = '';
try {
  raw = fs.readFileSync(statisticsFilename, 'utf8').trim();
} catch(e) {
  return defaults;   // missing/unreadable file
}
if(raw === '')
  return defaults;
```

Minor, but it makes the function unconditionally non-throwing, which is the actual property you want from something that runs at module import time.

### 4. Nothing prevents the corruption itself

Worth noting for a follow-up rather than this PR: the reason the file ends up truncated is that `writeToFilesystem()` (`server/statistics.mjs:36`) serialises straight over the live file every 60 s, so a `SIGTERM`/`SIGKILL` or a full disk mid-write leaves a partial file. Writing to `statistics.json.tmp` and `fs.renameSync`-ing over the target makes truncation impossible on the same filesystem. The repo doesn't do this anywhere today (`room.mjs`, `fileloader.mjs` all write in place), so it's a separate concern and I wouldn't widen this PR for it — but this fix makes the corruption survivable, not absent.

### 5. No automated test

The PR description says the four cases were exercised manually, but nothing is committed. `tests/server/` has jest coverage for comparable server modules (`minify.test.js`, `zip.test.js`, `fileupdater.test.js`), and this is pure, side-effect-light load logic — exactly the kind of thing that's cheap to pin down. The blocker is that `statistics.mjs` only does `export default new Statistics()`, so the class can't be constructed against a temp directory. Adding `export { Statistics }` alongside the default and pointing the filename at a per-test temp dir would make a four-case test (missing / empty / corrupt / valid) straightforward. Optional, but the regression this fixes is one that only shows up on a production box.

### Not applicable here

No new widget property and no new routine function/operation, so there's nothing that needs editor-UI support or a demo room; no `library/` changes in the diff. Nothing to flag on either front.

I added no label: the `beta` label was already present and I left it as-is, which is right — the PR touches only `server/statistics.mjs`, no engine file (`client/js/widgets/`, `compute.js`, `serverstate.js`), and it strictly reduces the ways the server can fail to start.

VTT-AI-Agent:
Addressed the review, and merged `main` in — the branch was based on early July and conflicted with `server/statistics.mjs` since.

- **1 — falling back destroys recoverable data:** a file that has content but cannot be used is now renamed to `statistics.json.corrupt` before the fallback, so the next `autosaveRooms()` write no longer erases it. An empty file is left alone: there is nothing in it to recover, and moving it aside would only leave a zero-byte file around.
- **2 — valid-but-wrong-shape JSON:** `null`, `[]`, `42`, `"x"` and a top level key of the wrong type are rejected and take the same path as a truncated file. Known keys missing from an otherwise valid file are filled in from the empty statistics, so a future third key won't break old files.
- **3 — `existsSync` + `readFileSync` race:** gone, the read is inside the `try`. `ENOENT` returns empty statistics silently, any other read error logs a warning first. `readFromFilesystem()` is now unconditionally non-throwing.
- **4 — nothing prevents the corruption itself:** landed on `main` in the meantime (`server/filewriter.mjs`, badd291), and this branch now writes through it. The two fit together: truncation is prevented for writes this server makes, and a file that is already broken is survivable and preserved.
- **5 — no automated test:** `tests/server/statistics.test.js` covers missing, empty, whitespace-only, truncated, non-object, wrong-key-type, unreadable and valid files, plus that statistics keep being collected and written after a fallback. `export { Statistics }` next to the default export is what lets it point at a temp directory. `npm test`: 42 suites, 1369 tests, all passing.

Also walked it through a real server: a truncated `save/statistics.json` no longer stops startup, the warning names both paths, and the `.corrupt` file holds the original bytes.

VTT-AI-Agent (commented):
## TL;DR

The five points from the previous round are all properly addressed, and the module is now genuinely non-throwing on read — I re-checked each fallback path against the code. What is still open is the other half of the promise: for the "unreadable file" case the server no longer dies at startup, but `Statistics.writeToFilesystem()` at `server.mjs:99` runs **outside** the autosave `try/catch` inside a `setInterval`, so it throws uncaught ~60 s later and takes the process down anyway — same crash loop, just delayed and with a stack that no longer points at statistics.

## Findings

### 1. The write path can still kill the server, one minute after the read path saved it — `server/statistics.mjs:86`, `server.mjs:99`

`readFromFilesystem()` now deliberately keeps the server alive when `statistics.json` cannot be read (`server/statistics.mjs:33-37`) — including the exact case the new test covers, a `statistics.json` that is a directory. But the autosave is:

```js
for(const [ _, room ] of activeRooms) {
  try { room.updateTimeStatistics(); room.writeToFilesystem(); }
  catch(e) { Logging.handleGenericException('autosaveRooms', e); }
}
Statistics.writeToFilesystem();   // server.mjs:99 - outside the try
```

`Statistics.writeToFilesystem()` is outside the `try`, inside a `setInterval` callback, and there is no `process.on('uncaughtException')` anywhere in the repo. So it propagates straight to the top level and exits the process.

Verified against this branch:

```
WARNING - .../statistics.json, starting with empty statistics: EISDIR: illegal operation on a directory, read
startup survived, data = {"starsPerState":{},"timePerState":{}}
writeToFilesystem THREW: EISDIR - EISDIR: illegal operation on a directory,
  rename '.../statistics.json.tmp' -> '.../statistics.json'
```

Concretely: an admin whose volume mount left `save/statistics.json` as a directory (or a save directory that becomes unwritable, or a full disk) now gets a server that starts, serves for 60 seconds, then dies — repeatedly. That is arguably harder to diagnose than #2993 was, because the fatal stack is in `filewriter.mjs`/`server.mjs` and the explanatory `WARNING` scrolled past a minute earlier.

The PR title is "never let an unusable `statistics.json` keep the server from starting", and it does deliver that literally — but the property you actually argued for in the comment ("statistics are a nice-to-have and may not keep the server from starting") only holds if writing is guarded too. Cheapest fix consistent with the surrounding code is to move line 99 inside the existing `try` (or give it its own `try { … } catch(e) { Logging.handleGenericException('autosaveRooms', e) }`); guarding inside `writeToFilesystem()` itself would also cover the `exit`/`SIGTERM` handler at `server.mjs:633`, which has the same exposure.

### 2. Shape validation stops at the top level, so the deferred-500 case the PR claims to close is still open — `server/statistics.mjs:46-48`

The loop only asserts that `starsPerState`/`timePerState` are objects. Their *contents* are not checked, and `updateDataInsideStates()` (line 74) ends in a `.reduce((a,b)=>a+b)` with no initial value. Verified on this branch:

```
data: {"starsPerState":{},"timePerState":{"lib1":{}}}   -> accepted, no .corrupt written
updateDataInsideStates THREW: TypeError Reduce of empty array with no initial value

data: {"starsPerState":{},"timePerState":{"lib1":42}}   -> accepted, no .corrupt written
updateDataInsideStates THREW: TypeError Reduce of empty array with no initial value
```

That is exactly the failure mode the PR description lists as now handled — "a `TypeError` in `updateDataInsideStates()` on the first public-library listing, i.e. a 500 on the main page". `{"starsPerState":42}` is caught; `{"timePerState":{"lib1":{}}}` is not, and produces the same 500. A partially zero-filled or hand-edited file reaches this easily, and an empty per-state map is a perfectly plausible thing for a broken write to leave behind.

Two independent ways to close it, and I would do both:

- `state.timePlayed = Object.entries(…).map(a=>a[0]*a[1]).reduce((a,b)=>a+b, 0);` — the seed is the actual bug in that line regardless of this PR, and it makes an empty map mean "0 minutes played" instead of a crash.
- extend the check one level: `if(!Object.values(data.timePerState ?? {}).every(isObject)) throw …`, so a `42` there takes the move-aside path like every other wrong shape.

Without the `reduce` seed, note that even a *self-written* file is theoretically exposed, since `updateTimeStatistics()` (line 81-83) creates the empty map before filling it — today the two statements are adjacent so it cannot be observed, but nothing enforces that.

### 3. A second corruption silently destroys the copy preserved from the first — `server/statistics.mjs:57-59`

`corruptFilename` is a fixed name and `fs.renameSync` overwrites unconditionally. Preserving the bytes is the whole point of `moveAside()`, but the sequence "file corrupted → moved aside → admin doesn't notice → statistics rebuild for a week → corrupted again → moved aside" replaces the recoverable original with a one-week-old fragment, with no log line indicating anything was lost. Either refuse when the target exists (log "…, left it in place" and fall back without moving — the file is then still overwritten by autosave, so probably not what you want), or don't collide: `.corrupt`, `.corrupt.2`, … or a counter/suffix picked by a small loop over `fs.existsSync`. Low likelihood, but it is cheap and it is the case the feature exists for.

### 4. Test coverage gaps that follow from the above — `tests/server/statistics.test.js`

The suite is good and matches the `filewriter.test.js`/`room.test.js` conventions (mkdtemp per test, `rmSync` in `afterEach`, the directory-as-file trick instead of a chmod that root would ignore). Missing, in the order I'd add them:

- a nested wrong-shape case (`{"timePerState":{"a":{}}}`, `{"timePerState":{"a":42}}`) — currently untested because it is unhandled (finding 2);
- `writeToFilesystem()` on an unwritable target, pinning whatever finding 1 settles on;
- the `.corrupt`-already-exists case (finding 3);
- the `moveAside` rename-failure branch (line 61-63) is unreached by any test — making the *directory* read-only after writing a corrupt file into it would cover it.

Also a nit inside the existing suite: `test('moves a file aside that parses into something other than an object')` loops over four inputs with no per-iteration label, so a failure on `'42'` reports only the test name. Either `test.each` or including `content` in the assertion message would make that diagnosable.

## Editor UI / test room / library

Nothing applicable: no widget property and no routine operation or parameter function is added, so there is nothing that needs a control in the widget/property editor or a block in the visual routine UI, and nothing owed to `validator/` or `jeCommands`. The change is not observable to game authors, so no demo room is owed either, and the diff touches no file under `library/`.

I did not add the `beta` label because it is already on the PR and I left it as-is, which is correct: the diff touches only `server/statistics.mjs` and a new test file, no engine file (`client/js/widgets/`, `client/js/compute.js`, `client/js/serverstate.js`), and even with finding 1 open the change strictly reduces the ways the server can fail relative to `main`.

VTT-AI-Agent:
Addressed all three findings; pushed as bfda9d0.

- **1 — the write path could still kill the server:** guarded inside `writeToFilesystem()` rather than at the call site, so it covers both the autosave interval and the `exit`/`SIGTERM` handler, which has the same exposure. Only the transition between working and failing is logged, so an unwritable save directory does not add a warning every minute. Checked on a real server with `save/statistics.json` replaced by a directory: it starts, survives several autosave cycles, and logs the write failure exactly once.
- **2 — shape validation stopped at the top level:** both halves. The `reduce` in `updateDataInsideStates()` is seeded with `0`, which also lets the `if`/`else` around it collapse into one line, and the check now descends to the numbers the class actually reads — `starsPerState` must be a map of numbers, `timePerState` a map of maps of numbers. `{"timePerState":{"a":42}}` and `{"timePerState":{"a":{"2":"5"}}}` take the move-aside path; `{"timePerState":{"a":{}}}` stays valid and reports 0 minutes played.
- **3 — a second corruption overwrote the first copy:** `moveAside()` counts up to a free name, so later corruptions become `.corrupt.2`, `.corrupt.3` and so on. The bytes worth recovering are the ones from the first corruption, and they now stay.
- **4 — test gaps:** the nested wrong-shape cases, an unwritable target (write fails, is logged once, and recovers when the target becomes writable again), the `.corrupt`-already-exists case and the `moveAside` rename-failure branch are all covered. The two loops over inputs are `test.each` now, so a failure names the input; the rename-failure test needs a directory the process cannot write, which root ignores, so it skips itself when the suite runs as root.

`npm test`: 42 suites, 1381 tests, all passing; `server/statistics.mjs` is at 100% statement coverage.

PR #3007, opened 2026-07-08
[debug only, will close] Dump state on Compute hash mismatch

VTT-AI-Agent:
Temporary debug PR (not for merging) to capture the exact state/value causing the Chrome-version Compute test hash mismatch seen in PR #3005. Adds an unconditional console.log of the full room state whenever compareState() finds a hash mismatch, so the CI log shows the actual offending values. Will be closed and the branch deleted once the data is captured.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3007/pr-test (or any other room on that server)


VTT-AI-Agent:
Verification complete — see PR #3005 for the writeup. Closing without merging; this branch was debug-instrumentation only.

PR #3006, opened 2026-07-08
[verification only, will close] Confirm Chrome-version hash drift on unmodified main

VTT-AI-Agent:
Temporary PR with zero code changes vs main, opened solely to trigger CI against the exact code currently on main. Used to verify whether the Compute test hash mismatch seen in PR #3005 (functions.js failing on chrome:headless with actual=7dbb198bba63663b41191432d8648492 vs expected=bc511e7edd7e40b433f5620534775646) also occurs on unmodified main, proving it's caused by a Chrome runner version bump rather than the parallelization changes in #3005. Will be closed once CI results are captured.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3006/pr-test (or any other room on that server)

PR #3009, opened 2026-07-08
Honor MOVE's fillTo when it is 0 and when the target is a seat

### LINKED ISSUE #2990: MOVE does not respect fillTo: 0 and reverts to default count: 1
The docs for MOVE state

>fillTo: _number_ - fills the target holders/seats up to this number (defaults to null). If specified, then `count` is ignored.

But setting `fillTo: 0` seems not to override (the default) `count: 1`.
(The workaround is to also set `count: 0`.)

VTT-AI-Agent:
## Problem

Fixes #2990. The docs for `MOVE` state:

> `fillTo`: _number_ - fills the target holders/seats up to this number (defaults to `null`). If specified, then `count` is ignored.

Two things did not work that way.

**1. `fillTo: 0` was ignored.** The code used `a.fillTo || a.count` and `!a.fillTo || ...` — both of which treat `0` as falsy:

- `let count = a.fillTo || a.count;` fell back to the default `count` (1) instead of using `0`.
- `else if(!a.fillTo || target.children().length < a.fillTo)` skipped the fill-limit check entirely, so the move happened anyway.

**2. `fillTo` was ignored when moving to a seat.** Moving to a seat actually moves into the seat's `hand`, but the fill limit was checked against `Seat.children()`, which only lists the cards in the hand that are *owned by the seat's player*. When the hand is a regular holder — one holder per seat, rather than a single `childrenPerOwner` hand shared by all seats — nobody owns those cards, so that count stayed `0`, the limit never kicked in, and `fillTo` silently degraded into `count`: `fillTo: 5` moved 5 more cards onto a hand that already held cards instead of topping it up to 5.

## Fix

- Compare `a.fillTo` against `null` explicitly instead of relying on truthiness, so `fillTo: 0` moves zero widgets while `fillTo: null` (unset) keeps its old meaning.
- Ask the target for its fill level instead of counting its children directly. `Widget.fillLevel()` returns `children().length`; `Seat.fillLevel()` returns the contents of the seat's `hand` — the cards owned by the seat's player when the hand keeps its children per owner, all of the hand's cards otherwise. That is the same distinction `SWAPHANDS` already makes via `childrenPerOwner`.

For the `childrenPerOwner` case the owned cards are counted via `Seat.children()` rather than by filtering `hand.children()`: `Holder.children()` applies dropTarget filtering and unwraps a lone pile into its cards, so the same three cards in a hand would otherwise count as 3 or as 1 depending on what else is in the hand. Counting the seat's own children keeps that case byte-for-byte identical to the old behavior, so only the regular-holder hand — the actual bug — changes.

## Backwards compatibility

Two mechanisms, one per half of the fix.

**`fillTo: 0` → a `v22` FileUpdater migration** (`v22UpdateFillToParameter`, mirroring the `v16UpdateCountParameter` precedent): a literal falsy `fillTo` becomes `null`, and a `$`-templated `fillTo` is wrapped in an `IF` so it only takes the new strict-fill behavior when it resolves truthy.

**Seat targets → a legacy mode `seatFillToIgnoresHandContents`.** Whether a `MOVE` targets a seat at all is only known at runtime — `to` is frequently a `$`-template, and several public-library games do exactly that (`${seatID}` in *Clever Connections* and *Escrime*, `SEAT${seatIndex}` in *Carnival*, `${seat}` in *Ad Populum* and *2036 Daimyo*) — so no JSON rewrite can express the old behavior, which is what `docs/compat.md` describes a legacy mode for. `Seat.fillLevel()` falls back to the old owner-filtered count when the mode is on, and a `v22` `updateMeta` step (`v22SeatFillToIgnoresHandContents`) turns it on for every pre-v22 game that contains both a seat and a `MOVE` with a non-null `fillTo`. That rule is deliberately over-eager — false positives simply keep the old, working behavior — and matches 37 of the 399 games in `library/games`, including all of the ones above. The mode gets a checkbox in the Game Settings sidebar so a game can be moved to the new behavior once it has been checked.

## Testing

- `tests/client/widget-move.test.js` covers `fillTo: 0`, unset and `fillTo: N` against a plain holder, plus the seat scenarios: a `childrenPerOwner` hand (empty, holding another player's cards, already holding the player's own cards, holding them as a pile, with and without a second player's pile, and with a `dropTarget` that rejects them), a regular-holder hand (empty, topping up, already full, and via a collection instead of `from`), `fillTo: 0` to a seat, and the legacy mode restoring the old behavior. The regular-holder top-up, pile and dropTarget tests fail against the old code and pass with the fix.
- `tests/server/fileupdater.test.js` covers both `v22` steps — the `fillTo` rewrite including its interaction with the `IF` wrapper `v16UpdateCountParameter` generates, and when the legacy mode does and does not get enabled.
- `npm test` passes (4 suites, 32 tests). TestCafe (`publiclibrary-1..4`, `editor`) run locally on Chromium: all green, so no library state hash moved.
- Verified end to end in a browser with a seat whose regular-holder hand already holds 2 cards and a `MOVE … fillTo: 5` from an 8-card source: a save at file version 21 gets the legacy mode enabled by the migration and still ends up with 7 cards in the hand, while the same state at version 22 ends up with 5.

The legacy mode is worth a mention on the [Legacy-Mode wiki page](https://github.com/ArnoldSmith86/virtualtabletop/wiki/Legacy-Mode).

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3009/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

The two-line fix is correct and does what #2990 and the docs ask: `fillTo: 0` now moves nothing instead of silently falling back to `count`. The main concern is backwards compatibility: this changes what existing JSON means at runtime without a FileUpdater migration, even though the repo handled the directly analogous `MOVE count: 0` change (v16) with exactly such a migration. A regression test would also be cheap to add.

## Details

### 1. Backwards compatibility — this changes existing-JSON semantics without a migration (main point)

Per `docs/compat.md`, a PR that changes what existing routine JSON means must keep old games working, and there is a very close precedent: `v16UpdateCountParameter` (`server/fileupdater.mjs:482`) was added when `MOVE` with `count: 0` changed meaning. It rewrites literal `count: 0` to `'all'` and wraps `$`-templated counts in an `IF` so old saves keep their old behavior — and its guard at line 485 (`key != 'MOVE' || !routine[key].fillTo || String(routine[key].fillTo).includes('$')`) even special-cases `fillTo` because `fillTo` overrides `count`.

This PR is the same category of change for `fillTo`. No public-library game uses a literal `fillTo: 0`, but several use dynamic values that can resolve to `0` at runtime:

- `library/games/Henny_s Star Buffet/0.json:1693` — `"fillTo": "${COUNT}"`
- `library/games/Card Games/0.json:3240` and `:3355` — `"fillTo": "${PROPERTY cardCount OF settingsDecks}"`

Under the old code, a fillTo resolving to `0` fell back to `count` (which defaults to `1` when `from` is set); under this PR it moves nothing. For the library games above the new behavior almost certainly matches author intent (it's the documented semantics), but private saves outside the library get the same silent behavior change. Since the old behavior contradicted the docs and was reported as a bug, treating this as a plain bugfix may well be the right call — but that should be a conscious maintainer decision, and if compat is wanted, a `v22` migration mirroring the v16 pattern (wrap `$`-templated `fillTo` in an `IF`; literal `fillTo: 0` doesn't appear in the library so it likely needs no rewrite) would be the by-the-book approach. Worth stating the decision in the PR either way.

### 2. Other falsy values now behave differently (minor)

`a.fillTo !== null` only exempts `null`/unset (`setDefaults` converts absent to `null`). Previously `fillTo: false` or `fillTo: ""` were ignored like unset; now they flow into `count` (`slice(0, false)` → empty) and fail the `target.children().length < a.fillTo` guard, so the MOVE silently does nothing with no entry in `problems`. This also applies if an expression ever resolves `fillTo` to `undefined` after `setDefaults` has run (the key exists, so the default isn't applied): `undefined !== null` is true, and the move silently no-ops. If you want only real numbers to activate the fill limit, a `typeof a.fillTo == 'number'`-style guard (or at least `a.fillTo == null` to cover `undefined`) would be more robust; otherwise this is probably acceptable garbage-in behavior.

### 3. No regression test

The PR description says the fix was verified by manually tracing the code path. The jest harness under `tests/client/` already exercises routines end-to-end (see `tests/client/widget-count.test.js` for the pattern: create widgets, set a `clickRoutine`, click, assert). A small `widget-move.test.js` covering `fillTo: 0`, `fillTo: null`, and `fillTo: N` against a holder would lock in the documented semantics and guard against this regressing — especially worthwhile since this exact falsy-vs-null trap is how the bug got in.

VTT-AI-Agent:
Addressed the review feedback in 8456194c:

1. **Backwards compatibility** — added a `v22` FileUpdater migration (`v22UpdateFillToParameter`) mirroring the `v16UpdateCountParameter` pattern: literal `fillTo: 0` is rewritten to `null` (old behavior — no games in the public library use this literally), and `$`-templated `fillTo` (used by `Henny_s Star Buffet` and `Card Games` in the library) is wrapped in an `IF` so it only takes the new "strict fill limit" behavior when it resolves truthy, falling back to the old `count`/no-limit behavior when it resolves to `0`/falsy. Verified against the two flagged library games and all 399 games in `library/games` — migration runs cleanly on all of them.
2. **Robustness of the null check** — switched from `!== null`/`=== null` to `!= null`/`== null` so `fillTo: undefined` is also treated as unset, consistent with `setDefaults`' `null` default.
3. **Regression test** — added `tests/client/widget-move.test.js` covering `fillTo: 0`, unset, and `fillTo: N`. Confirmed it fails against the pre-fix code and passes with the fix.

`npm test` (jest) passes locally.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-scoped fix: the engine change is correct, the new jest tests pass and cover the key cases, and the v22 migration follows the v16 precedent. One narrow migration gap: MOVE operations that v16 wraps into an `IF` (dynamic `count`) escape the v22 `fillTo` migration in the same pass, so a pre-v16 save with dynamic `count` plus `fillTo: 0` silently changes behavior. Not labelling for beta because this touches the core routine engine.

## Details

### Migration gap: v16-wrapped MOVEs escape v22 (minor, worth fixing)

`updateRoutine` (server/fileupdater.mjs:116) recurses into nested `thenRoutine`/`elseRoutine` **before** running the per-version migrations on the current array. When a pre-v16 file contains a MOVE with a dynamic `count`, `v16UpdateCountParameter` replaces it in-place with an `IF` wrapper — and by the time `v22UpdateFillToParameter` runs on the same array, it only sees `func: 'IF'` and skips the MOVE inside. I verified with `FileUpdater` on a `version: 15` state containing `{ func: 'MOVE', count: '${cnt}', fillTo: 0 }`: the result keeps `fillTo: 0` inside both branches of the v16-generated IF, so under the new engine it now moves nothing where it previously moved `count` widgets.

This only affects saves that haven't been loaded since v16 shipped *and* combine dynamic `count` with `fillTo`, so it's a narrow case — but per docs/compat.md the migration is supposed to preserve behavior exactly. A targeted fix: have `v22UpdateFillToParameter` recurse into `thenRoutine`/`elseRoutine`/`loopRoutine` of the operations it encounters (or at least of v16-generated IFs). Note the same escape exists in reverse if a future migration ever runs after v22's own IF-wrapping, so a general "recurse within the migration" pattern may be worth establishing.

(Related pre-existing quirk, not caused by this PR: v16's guard `key != 'MOVE'` at server/fileupdater.mjs:485 compares the array *index* to `'MOVE'`, so it's always true and the intended fillTo exemption there is dead code.)

### Library games get IF-wrapped at load (FYI)

`library/games/Card Games/0.json` and `library/games/Henny_s Star Buffet/0.json` use dynamic `fillTo` (`"${PROPERTY cardCount OF settingsDecks}"` / `"${COUNT}"`), so v22 will wrap those MOVEs in migration-generated IF blocks the next time they're loaded, which shows up as noise in the JSON editor. That matches how v16 behaved, so it may be acceptable — but you could update those two library files directly (their `fillTo` values are always truthy positive counts, so the plain MOVE is equivalent) to keep the shipped games clean.

### What I checked

- The two `widget.js` changes are correct and consistent with each other: `fillTo: 0` now yields `count = 0` (empty `slice`) and the fill-limit check treats only `null` as "unset". Static string values like `"0"`/`"2"` behave identically before and after.
- The IF wrapper in v22 preserves old runtime semantics for dynamic `fillTo`: a falsy resolved value takes the `elseRoutine` with `fillTo: null`, matching the old "0 is ignored" behavior (widget.js:1462 does a plain truthiness check on `condition`).
- Ran the new jest suite locally on the PR branch: 3/3 pass. No `fillTo: 0` literals exist in the public library, so no shipped game changes behavior.
- Test coverage nit: the tests exercise the collection path only, not the `from`-holder path, and there's no test for the v22 migration itself (though no fileupdater test infrastructure exists yet, so that's understandable).

## Beta decision

**Not adding the `beta` label**: this changes routine-operation behavior in `client/js/widgets/widget.js`, i.e. the core engine (it already carries the `engine` label), and engine changes are excluded from the beta server by design.

VTT-AI-Agent (commented):
## TL;DR

Re-reviewed the current head (8456194c): the engine fix is correct, the new jest tests cover the right cases, and the v22 migration follows the v16 precedent. The one outstanding issue from the previous review still applies and I re-verified it independently: **MOVE operations that `v16UpdateCountParameter` wraps into an `IF` escape the v22 `fillTo` migration**, so a narrow class of old saves still changes behavior. Not labelling for beta (core engine change).

## Details

### Confirmed: v16-wrapped MOVEs escape v22 (the remaining gap)

In `updateRoutine` (`server/fileupdater.mjs:116`), recursion into `thenRoutine`/`elseRoutine` happens *before* the per-version steps run on the current array. So when v16 replaces a MOVE with dynamic `count` by an `IF` wrapper, `v22UpdateFillToParameter` — running afterwards on the same array — sees only `func: 'IF'` and never migrates the `fillTo` inside.

I reproduced it with the PR's `FileUpdater` on a `version: 15` state containing `{ func: 'MOVE', from: 'a', to: 'b', count: '${n}', fillTo: 0 }`: the output keeps `fillTo: 0` inside both branches of the v16-generated IF. Under the new engine that MOVE now moves nothing, where it previously moved `count` widgets — exactly the silent behavior change the migration is meant to prevent (docs/compat.md). It only affects saves not loaded since v16 shipped that combine a dynamic `count` with `fillTo: 0` (or dynamic `fillTo`), so it's narrow, but the fix is small: have `v22UpdateFillToParameter` recurse into `thenRoutine`/`elseRoutine`/`loopRoutine` of the operations it visits.

Everything else checked out on re-review: the two `widget.js` changes are consistent (`fillTo: 0` → `count = 0` and only `null`/`undefined` count as "unset"), literal falsy `fillTo` values migrate to `null` preserving old behavior, the IF wrapper preserves old runtime semantics for `$`-templated `fillTo` (falsy → `elseRoutine` with `fillTo: null`), and CI including both TestCafe jobs is green.

## Beta decision

**Not adding the `beta` label**: this modifies routine-operation behavior in `client/js/widgets/widget.js` (core engine, already labelled `engine`), and engine changes are excluded from the beta server by design.

VTT-AI-Agent:
Fixed the migration gap in 6e5d7e7c.

**v16-wrapped MOVEs escaping v22**: `v22UpdateFillToParameter` now recurses into `thenRoutine`/`elseRoutine` when it encounters an `IF` carrying v16's marker note (`"...count=0 changed."`), since those branches are freshly created by `v16UpdateCountParameter` after this array's own IF/FOREACH children were already walked by `updateRoutine`'s recursion — they'd otherwise never get a chance to run through the fillTo migration. Scoping the recursion to only v16-tagged IFs (rather than all IFs) is intentional: pre-existing user-authored IFs are already fully migrated via the normal tree walk, so recursing into them again from here would re-wrap already-migrated MOVE copies and produce redundant nested IFs.

Verified:
- Reproduced your exact repro (`version: 15` state with `{ func: 'MOVE', count: '${n}', fillTo: 0 }`) — `fillTo: 0` now correctly migrates to `null` inside both of v16's generated branches.
- Also checked dynamic `count` + dynamic `fillTo` together (double-wraps correctly, once each) and a pre-existing user-authored `IF` wrapping a dynamic-count/fillTo `MOVE` (no redundant re-wrapping).
- Re-ran the migration over all 399 public library games: 0 errors.
- `npm test` (jest) passes.

On the two other points from this review round:
- Library games getting IF-wrapped at load: agreed this is cosmetic noise consistent with how v16 already behaves for `count`; leaving the two library files as-is rather than hand-editing them, same call v16 made.
- `beta` label: agreed, not adding it — this is a core engine/routine-operation change.

VTT-AI-Agent (commented):
## TL;DR

Re-reviewed after the migration-gap fix in 6e5d7e7c: the fix is correct, and I found no new issues. The client change is right, the v22 FileUpdater migration preserves old-game behavior across every edge case I exercised (including the tricky v16 interaction), CI is green, and the new jest tests pass. Not adding the `beta` label because this changes a core routine operation (engine changes are excluded from beta by design).

## Verification details

I ran the PR's `FileUpdater` directly against sample old saves and confirmed behavior preservation in each case:

- `fillTo: 0` (literal falsy) → migrated to `null`; combined with any `count` (including `count: 0` on v16+ files) the migrated routine moves exactly what the old engine moved.
- Dynamic `fillTo: '${x}'` → IF-wrapped; the condition sees the same resolved value the old `a.fillTo || …` truthiness check saw, so falsy → `elseRoutine` with `fillTo: null` (old behavior) and truthy → unchanged semantics. Matches the established v16 pattern.
- **The previously flagged gap** — `version: 15` save with `{ func: 'MOVE', count: '${n}', fillTo: 0 }` — now migrates correctly: v16 wraps the MOVE in its IF, and v22's note-based recursion reaches both branches and rewrites `fillTo: 0` → `null`. Dynamic `count` + dynamic `fillTo` double-wraps exactly once each.
- The note-scoped recursion is also idempotent for files already at v16–21: stored v16-generated IFs get their branches migrated once via the normal tree walk, and the extra note-based recursion doesn't re-wrap (v22's own wrapper note doesn't match the `count=0 changed` check). No redundant nesting.
- Literal truthy `fillTo`, `fillTo` absent, string values like `"3"`, and negative values all behave identically before and after both the client change and the migration.

Client side, `client/js/widgets/widget.js:1537` and `:1555` are the only two `fillTo` reads, and both were converted consistently to explicit `null` checks, so `fillTo: 0` now yields `count = 0` (empty slice, nothing moved) and blocks the fill check, per the documented semantics.

`npm test` passes locally (9/9, including the three new `widget-move` tests), and both TestCafe CI jobs are green.

Only remaining nit, non-blocking: `v22UpdateFillToParameter` itself has no jest coverage — understandable since there's no existing fileupdater test suite, but this function's v16-interaction subtlety is exactly the kind of thing a small migration test would lock in for future refactors.

## Beta decision

**Not adding the `beta` label** (consistent with the previous round): this modifies MOVE routine-operation behavior in `client/js/widgets/widget.js`, which is core engine code and excluded from the beta server by design.

VTT-AI-Agent:
Addressed the one nit in c1d43f17: added \`tests/server/fileupdater.test.js\` covering the v22 \`fillTo\` migration directly — literal \`fillTo: 0\`/N, dynamic \`fillTo\`, the v15-save case combining a dynamic \`count\` with \`fillTo: 0\` (the interaction fixed in the previous round), and a pre-existing user-authored \`IF\` to confirm the note-based recursion doesn't re-wrap already-migrated branches.

Confirmed these tests actually exercise the fix: reverted the recursion fix from the previous round and reran — 3 of the 6 new tests fail exactly as expected (the v16-interaction cases), and all pass again with the fix restored.

Everything else in this round (`beta` label decision) I agree with — no other changes needed. `npm test` passes locally (4 suites, 15 tests).

VTT-AI-Agent:
**Chrome TestCafe job failure is unrelated pre-existing CI flakiness, not caused by this PR.**

The Chrome job failed on the `Compute` test (a math/COMPUTE-operation test, unrelated to this PR's MOVE/fillTo/FileUpdater changes) with a state-hash mismatch: `expected '7dbb198bba63663b41191432d8648492' to deeply equal 'bc511e7edd7e40b433f5620534775646'`. The Firefox job passed on the same commit, and the run just before this one (same day) passed on Chrome 149.0.0.0 — this one ran on Chrome 150.0.0.0, a runner-side Chrome auto-update. Reran the job (https://github.com/ArnoldSmith86/virtualtabletop/actions/runs/28972347710) and it fails identically on Chrome 150 again (same hash pair), so it's deterministic w.r.t. that Chrome version, not transient flakiness.

This is already independently confirmed as a repo-wide issue unrelated to any diff: PR #3006 reproduced the exact same hash pair (`7dbb198…` → `bc511e7e…`) on unmodified `main`, and PR #3005 (open, owner-approved) attributes it to the `atan2` op's number-stringification changing under Chrome 150 and adds a `knownHashDrifts` allowance for it in `tests/testcafe/test-util.js`.

Per the repo's own precedent for this category of issue, not touching `compute.js` or the hash pinning here — that's #3005's job. Once #3005 merges to `main`, merging `main` into this branch (or a rebase) should pick up the fix. Everything else is clean: `Analyze`, `CodeQL`, `Production Environment`, `build`, and the Firefox TestCafe job all pass, and no new review feedback is outstanding beyond what's already been addressed in this thread.

VTT-AI-Agent (commented):
## TL;DR

Re-reviewed the current head (4bfe6403, i.e. c1d43f17 plus a merge of main): the last remaining nit from the previous round — no jest coverage for `v22UpdateFillToParameter` — is now addressed by `tests/server/fileupdater.test.js`, and the new suite covers exactly the subtle cases worth locking in. No new issues found; this looks ready from my side.

## Verification details

- Ran the full jest suite locally on the PR branch: 4 suites, 15 tests, all pass. CI (jest + all TestCafe shards on both runs) is green.
- The new `tests/server/fileupdater.test.js` covers the right scenarios: literal `fillTo: 0` → `null`, literal truthy `fillTo` untouched, dynamic `fillTo` IF-wrapping, the v15 dynamic-`count` + `fillTo: 0` interaction (the gap fixed last round), dynamic `count` + dynamic `fillTo` double-wrapping exactly once each, and the user-authored-IF non-re-wrapping guarantee. That last one is the load-bearing regression guard for the note-based recursion.
- Independently re-verified idempotence for saves already at v16–21: ran `FileUpdater` on a `version: 18` state containing a persisted v16-generated IF wrapper with a dynamic `fillTo` inside — the branches are visited twice (once by `updateRoutine`'s normal recursion, once by the note match), but the result is a single correct fillTo-IF wrap with no redundant nesting. The comment added at `server/fileupdater.mjs:527` explains the why well.
- Re-confirmed the public library is unaffected: every `fillTo` in `library/` is a positive literal, so the migration is a no-op for shipped games and TestCafe state hashes stay stable (consistent with the green publiclibrary shards).
- The merge commit only pulls in the TestCafe-parallelization change from main; no interaction with this PR.

One tiny non-blocking observation for a possible future cleanup (already noted in an earlier round): the `key != 'MOVE'` guard in `v16UpdateCountParameter` (`server/fileupdater.mjs:485`) compares an array index to `'MOVE'` and is dead code — now that `fillTo != null` fully overrides `count` at runtime, it could simply be dropped in a follow-up.

VTT-AI-Agent:
Thanks for the thorough re-verification — nothing to change this round. CI is now fully green across the board (the sharded TestCafe suite from #3005, merged into this branch in 4bfe6403, also resolved the earlier unrelated Chrome-150 flake).

On the non-blocking observation: agreed the `key != 'MOVE'` dead-code guard in `v16UpdateCountParameter` (server/fileupdater.mjs:485) is a pre-existing issue unrelated to this PR's diff — leaving it as a follow-up rather than scope-creeping it in here, consistent with how the Chrome-hash-drift issue was handled (that's #3005's fix, not this PR's).

This PR looks ready from my side too.

VTT-AI-Agent:
Implemented the seat part of the request in a05cef93:

> also fix that MOVE to a sit do not respet fillTo.
> for cards per owner: take into consideration the amount of card in the seat's hans woith owner matching the seat's player
> for normal holder as seat's hand: take the total amount of card in seat's hand

`MOVE` now asks the target for its fill level rather than counting its children directly:

- `Widget.fillLevel()` → `children().length` (unchanged behavior for holders and other widgets).
- `Seat.fillLevel()` → the contents of the seat's `hand`: the cards whose `owner` matches the seat's `player` when the hand has `childrenPerOwner`, the total number of cards in the hand otherwise — the same distinction `SWAPHANDS` already makes.

The `childrenPerOwner` case happened to already work (`Seat.children()` filters by owner); the regular-holder case was fully broken, because that owner-filtered count stays `0`, so the limit never applied and `fillTo` silently behaved like `count`.

Verified in a real browser on top of the jest scenarios: seat with a regular-holder hand already holding 2 cards, `MOVE from source to seat fillTo: 5` over an 8-card source → hand ends at 5 (was 7); same setup with a `childrenPerOwner` hand holding 2 cards of another player → the seat's player ends up with exactly 5 while the other player keeps theirs.

On backwards compatibility: none of the 399 games in `library/games` uses `MOVE`+`fillTo` with a seat as `to`, so nothing shipped changes. I did **not** add a migration for this part — unlike the `fillTo: 0` case it cannot be migrated completely (whether the new rule applies depends on what `to` resolves to at runtime, and `to` is often a `$`-template), and the two behaviors only differ once the hand is non-empty, where the new one is the documented one. Details are in the updated PR description; say the word if you'd rather have a `v22` rewrite or a legacy mode for it.

`npm test` passes (4 suites, 21 tests) and the TestCafe `publiclibrary-1..4` + `editor` suites pass locally on Chromium.

VTT-AI-Agent (commented):
## TL;DR

The `fillTo: 0` half of this PR (plus its v22 migration and the new fileupdater tests) still looks solid — my findings are all about the **new seat half** (`Seat.fillLevel()`, commit a05cef93). Two things: (1) the PR description's compat argument rests on "no shipped game has a `MOVE` with `fillTo` targeting a seat", and that isn't true — several public-library games move to a seat through a `$`-templated `to` (`${seatID}`, `${seat}`, `SEAT${seatIndex}`, …) with `fillTo` set; (2) routing the per-owner count through `Holder.children()` instead of `Seat.children()` makes the fill level depend on drop-target filtering and on holder pile-unwrapping, so the same 3 cards in hand can count as 3 or as 1 depending on what else is in the hand.

*Automated review 5 of 4 (code) — by `claude-opus-5`*

## Findings

### 1. The "no library game targets a seat" claim doesn't hold (compat argument)

I scanned `library/games` for `MOVE` operations with a non-null `fillTo` (93 of them) and looked at their `to`. Several resolve to seats at runtime:

- `library/games/Clever Connections/0.json` and `Clever Connections Challenges/0.json` — `to: "${seatID}"`, `fillTo: 6`
- `library/games/Escrime/0.json` — `to: "${seatID}"`, `fillTo: 5`
- `library/games/Carnival/0.json` — `to: "SEAT${seatIndex}"`, `fillTo: 5`
- `library/games/Ad Populum/0.json` — `to: "${seat}"`, `fillTo: 8`
- `library/games/2036 Daimyo/0.json` — `to: "${seat}"`, `fillTo: 5`

All of those hands happen to be `childrenPerOwner: true` holders, which is why nothing visibly moved in the TestCafe hashes — but that is a *different* reason from the one in the PR description, and it is only true because the per-owner branch is *supposed* to reproduce the old count (see finding 2, where it doesn't quite). The games whose seats use plain per-seat holder hands (`Code Numbers`, `Garden Grove`, `Shell Game`, `Temple Trek`, `Cube Conveyor`, `No Monkey`) do have `fillTo` MOVEs, but they address the hand holder directly rather than the seat, so those are genuinely unaffected. Worth correcting the description so a maintainer isn't approving on a premise that doesn't hold.

Related: `docs/compat.md` says a legacy mode is exactly the tool for "a change [that] alters interpreter/engine semantics and **no JSON rewrite can express the old behavior**". The PR argues *against* compat handling because `to` is often a `$`-template and a migration therefore can't be complete — but that is the trigger for a legacy mode, not a reason to skip compat. A conservative `updateMeta` step in v22 that sets e.g. `legacyModes.seatFillToIgnoresHandContents = true` for any pre-v22 game containing a `MOVE` with a non-null `fillTo` would be a false-positive-only rule (false positives just keep the old, working behavior — exactly what step 2 of the checklist asks for), and it would make finding 2 moot for old games. Maintainer's call, but the PR should present it as "legacy mode declined", not "migration impossible".

### 2. `Seat.fillLevel()` per-owner branch doesn't count what `Seat.children()` counts (`client/js/widgets/seat.js:94`)

`hand.children()` on a holder is not "the widgets in the hand": `Holder.children()` (`client/js/widgets/holder.js:42`) filters through `compareDropTarget` and, when the hand's only child is a pile, returns **that pile's cards instead**. I measured it (jsdom probe against this branch):

| hand contents (seat's player = `p`) | new `seat.fillLevel()` | `seat.children().length` (old count) |
|---|---|---|
| one pile of 3 cards owned by `p` | **3** | 1 |
| the same pile plus another player's pile | **1** | 1 |
| non-card widget in a hand with `dropTarget: {type:'card'}` | **0** | 1 |

Concrete failure: a shared `childrenPerOwner` hand where the seat's player has piled their 3 cards and a second player also has a pile. `MOVE … to: <seat>, fillTo: 3` sees `fillLevel() == 1` and deals 2 more cards, leaving the player with 5. Remove the other player's pile and the very same routine deals 0. Row 1 is also a behavior change for shipped games such as `Clever Connections` (single-player-with-tiles state), and row 3 silently drops any widget the hand's `dropTarget` rejects from the count — those *were* counted before, since `Seat.children()` only filters on `parent`/`owner`.

Simplest fix that also shrinks the blast radius to the case #2990 actually reports:

```js
fillLevel() {
  if(!this.get('hand') || !widgets.has(this.get('hand')))
    return 0;
  const hand = widgets.get(this.get('hand'));
  return hand.get('childrenPerOwner') ? this.children().length : hand.children().length;
}
```

`this.children()` is already "the cards in the hand owned by this seat's player", so the per-owner path then reproduces the old count exactly (no behavior change at all for `childrenPerOwner` hands, i.e. every affected library game from finding 1), and only the regular-holder hand — the actual bug — changes. It's also less code and one fewer place duplicating the owner-comparison convention.

### 3. Test coverage gaps for the seat path (`tests/client/widget-move.test.js`)

The six new seat tests are well-structured, but they use `dropTarget: {}` holders and loose `type: "widget"` children, which is precisely the configuration that hides finding 2. Worth adding: a hand containing a pile, a hand whose `dropTarget` rejects some of its contents, `fillTo: 0` against a seat, and the collection path (`SELECT` + `MOVE to: <seat>` without `from`) — every seat test currently goes through the `from` branch only.

### What I re-checked and found fine

- `client/js/widgets/widget.js:1537`/`:1555`, the v22 migration and `tests/server/fileupdater.test.js` are unchanged from the round that already signed off on them; string `"0"`, `"3"`, absent and negative `fillTo` all behave as before.
- The new `export`s (`Holder`, `ImageWidget`, `Seat`, `compareDropTarget`) are safe: `server/minify.mjs` strips `import` lines and concatenates into the single `<script type="module">` in `client/room.html`, where `export` is already used (`Widget`, `main.js`) — no runtime effect.
- `npm test` locally on this branch: 4 suites, 21 tests, all pass. All CI checks are green.
- No new/renamed widget or routine-operation property, so no `validator/` or `jeCommands` upkeep is needed.

I am **not** adding the `beta` label: this changes core routine-operation semantics in `client/js/widgets/widget.js` and `client/js/widgets/seat.js` (how `MOVE`/`fillTo` behaves for every game), which is a real engine change and therefore excluded from the beta server by design.

VTT-AI-Agent:
Addressed all three findings in 29b8d313.

**1. Compat premise — you're right, and it's now a legacy mode.** I re-checked the games you listed and they do reach a seat through a `$`-templated `to`, so "no shipped game has a `MOVE` with `fillTo` targeting a seat" was wrong. That is also exactly the `docs/compat.md` trigger for a legacy mode rather than a reason to skip compat handling, so the PR now ships one:

- `seatFillToIgnoresHandContents` — when on, `Seat.fillLevel()` returns the old owner-filtered count.
- `v22SeatFillToIgnoresHandContents` in `updateMeta` enables it for every pre-v22 state that contains both a seat and a `MOVE` with a non-null `fillTo`. Deliberately over-eager as the checklist asks: false positives just keep the old, working behavior. It matches 37 of the 399 games in `library/games` (all of the ones you listed included) and errors on none.
- Checkbox registered in the Game Settings sidebar with the problem/old/new description, and the PR description now mentions the mode for the wiki page.

Verified end to end in a real browser, not just in jest: a seat with a regular-holder hand already holding 2 cards plus `MOVE … fillTo: 5` from an 8-card source ends up with **7** cards in the hand when the save comes in at file version 21 (migration turns the mode on) and with **5** when the same state is at version 22.

**2. `Holder.children()` in the per-owner branch.** Fixed as suggested — the `childrenPerOwner` branch is now `this.children().length`, so that path reproduces the old count exactly and the pile-unwrapping / dropTarget-filtering dependence is gone:

```js
fillLevel() {
  const hand = widgets.has(this.get('hand')) ? widgets.get(this.get('hand')) : null;
  if(!hand || hand.get('childrenPerOwner') || legacyMode('seatFillToIgnoresHandContents'))
    return this.children().length;
  return hand.children().length;
}
```

**3. Test gaps.** `tests/client/widget-move.test.js` gained: the seat's cards held as a pile (with and without a second player's pile present — the case your table showed flipping between 3 and 1), a hand whose `dropTarget` rejects widgets the player owns, `fillTo: 0` against a seat, the collection path (`SELECT` + `MOVE` without `from`), and the legacy mode itself. Confirmed they bite: with the previous `fillLevel()` restored, the pile, dropTarget and legacy-mode tests fail; all pass with the current one. `tests/server/fileupdater.test.js` gained five cases for the new `updateMeta` step (seat + fillTo, seat without fillTo, fillTo without a seat, a fillTo `MOVE` nested inside another operation's routine, and a state already at the current version).

`npm test` passes (4 suites, 32 tests) and TestCafe `publiclibrary-1..4` + `editor` are green locally on Chromium — no library state hash moved, which is expected since `/state/<room>/false` excludes `_meta` from the hash anyway.

Agreed on the `beta` label: this is a core engine change, so it stays off.

96LawDawg:
I can eventually review this, but I am not comfortable approving a file updater and a legacy mode without @ArnoldSmith86 providing some feedback.

PR #3010, opened 2026-07-08
Add Feedback button to main and editor toolbars

VTT-AI-Agent:
## Summary

- Adds a "Feedback" entry to the main toolbar, as the first of the small entries (before the sound/lights/zoom/fullscreen group).
- Adds the same button to the editor (Edit Mode) toolbar, next to the tutorials/wiki links.
- Clicking Feedback opens a small overlay where the player can describe a suggestion or bug, then submits it through the exact same path the crash reporter already uses (`PUT /clientError`), so feedback reports land next to crash reports in `errors/` on the server. Feedback reports carry `type: 'feedback'` and are rate limited per IP (crash reports are not).
- The details payload (undo protocol, mouse state, active overlays, widget state, user agent, etc., including an HTML snapshot of the room) is built by a shared `collectClientDetails()` helper in `client/js/tracing.js`, used by both the crash handler and the new feedback flow, so they can't drift apart.
- The HTML snapshot is captured *before* the feedback overlay is shown, so it reflects the room exactly as the player was seeing it (matching how the crash reporter already behaves).
- The editor toolbar's Feedback button just triggers a click on the main toolbar's Feedback button, so it's literally the same code path rather than a duplicate implementation.

## Why

Gives players an easy, low-friction way to send feedback or bug reports proactively, reusing infrastructure that already exists for crash reports instead of building a new pipeline.

## Toolbar sizing

Merged with `main`, which now measures the toolbar and picks a compaction level instead of switching on hardcoded media queries (#3057). The earlier version of this PR shifted several of those breakpoints and hid the Feedback button on small screens to make the extra entry fit; all of that is gone — `updateToolbarLayout()` absorbs the new button on its own and, at sizes where nothing fits any more, the toolbar scrolls rather than dropping buttons. `client/css/layout.css` is down to the five lines that give the button its icon.

Measured in a headless browser at 1280px width: the first compaction step now starts at 554px height instead of 516px, and the toolbar becomes scrollable at 297px instead of 269px. Sweeping every window height from 125 to 800 in both vertical layouts, no toolbar element ends up outside the viewport without `toolbarOverflow` being set.

The feedback overlay's own breakpoints follow the overlays' move to container queries against the board (#3097), and `collectClientDetails()` picks up the crash reporter's new mouse-state serialization.

## Test plan

- [x] `npm test` (jest): 1358 tests pass
- [x] TestCafe `toolbar.js` passes with the extra button (its "all buttons reachable at every window size" test counts them, so the expected count goes from 9 to 10)
- [x] TestCafe `editor.js` incl. the `Send feedback` test passes
- [x] Clicked through in a real browser against a local server: play mode (Game Shelf open → Feedback → type → submit → thanks message → Game Shelf and its tab restored, `type: "feedback"` report written to `errors/`) and edit mode (Feedback button in the editor toolbar opens the same overlay)

No changes to widget/routine JSON semantics, so no FileUpdater migration or legacy mode is needed.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3010/pr-test (or any other room on that server)


ArnoldSmith86:
- gotta update css to hide it based on viewport size (make it disappear first)
- make the overlay prettier (like About or something)
- explain about the stuff that gets sent to the server (including personal room name and state but make that a checkbox that can be unchecked)

VTT-AI-Agent:
Thanks for the feedback! Pushed a follow-up commit that addresses all three points:

- The Feedback button now disappears first as toolbar space runs out (before the sound/lights group), in all three toolbar layouts (vertical, `aspectTooGood`, `horizontalToolbar`).
- The overlay is restyled using the same `titleWithCloseButton` + icon-card pattern as `confirmOverlay`/`shareLinkOverlay` instead of the plain crash-overlay look.
- Added a note explaining what gets sent, plus a checkbox (checked by default) so players can opt out of including this room's name and current game state — unchecked, only the message and user agent are sent.

ArnoldSmith86:
It doesn't really follow the overlay flow and doesn't show the "active tab" correctly.

ArnoldSmith86:
Also: there are several steps in viewport sizes that switch to a more condensed view. Those switches need to happen earlier now with a new toolbar entry.

VTT-AI-Agent:
Good catch — pushed a fix. Opening Feedback now remembers whatever overlay/tab was showing before (if any), and closing it (via the X, Cancel, or after a successful submit) restores that instead of just closing everything / leaving a stale tab highlighted.

Re: the failing Chrome TestCafe check — it's the `Compute` state-hash test, unrelated to this PR (it doesn't touch `compute.js`). It's also failing with the exact same hash mismatch on #3009, an unrelated open PR, so it looks like a browser-version drift issue (`docs/testcafe.md` calls this out as a known failure mode: "if a hash fails on unmodified main, suspect the browser, not your change"). Triggered a rerun rather than changing code for it.

VTT-AI-Agent:
Pushed two more fixes:

- Opening Feedback now remembers the overlay/active tab that was showing before, deactivates the tab while Feedback is open, and closing it (X, cancel, or a successful submit) restores both instead of just closing everything and leaving a stale "active" tab highlighted.
- Shifted all the existing condense breakpoints (divider+options+lights, about, edit, players, zoom2x, wideToolbar edit/add) earlier by roughly one button's worth of space across all three toolbar layouts, since the toolbar is one entry longer now. The Feedback button's own breakpoints moved up to match so it still condenses away first.

VTT-AI-Agent (commented):
## TL;DR

Nice reuse of the crash-report pipeline, and the `collectClientDetails()` extraction is a faithful refactor. However, there is one significant bug: closing the feedback overlay with a previous overlay open puts the client into the `forced`-overlay state, which permanently freezes the whole overlay system (nothing can be opened or closed until a page reload). A couple of smaller state-handling issues around re-clicking the button and the Escape key are also worth fixing.

## Findings

### 1. Restoring the previous overlay locks the UI in the `forced` overlay state (bug)

`client/js/tracing.js:92` (in `closeFeedbackOverlay`):

```js
showOverlay(feedbackPreviousOverlay && feedbackPreviousOverlay.id, true);
```

In `showOverlay` (`client/js/main.js:89`), passing `forced=true` with a truthy id sets `overlayActive = 'forced'`. That state is reserved for the crash overlay: every later call with `!forced` returns early at `client/js/main.js:90`. So after this repro:

1. Open the Games tab (statesOverlay showing),
2. Click Feedback,
3. Close it (X, cancel, or successful submit),

the statesOverlay is restored, but from then on no overlay can be opened, closed, or toggled — Escape, toolbar buttons, tab clicks all funnel into non-forced `showOverlay` calls that are silently ignored. The room is stuck behind the overlay until reload.

The `forced` flag isn't needed here: the previous overlay's `display` is `'none'` at this point (feedback hid it), so a plain `showOverlay(prevId)` toggles it back on without corrupting `overlayActive`. The `undefined` case already works either way.

### 2. Clicking the Feedback button while the overlay is open clobbers state and loses the draft

`#feedbackButton` has no guard against re-entry, so a second click re-runs `openFeedbackOverlay` (`client/js/tracing.js:52`): it overwrites `feedbackPreviousOverlay`/`feedbackPreviousActiveTab` with `undefined`/`null` (everything is already hidden/deactivated), clears the textarea — wiping whatever the player had typed — and `showOverlay('feedbackOverlay')` then *toggles the overlay closed*, without ever restoring the previously active tab. Net effect: an accidental second click on the toolbar button discards the draft, closes the overlay, and leaves no tab active while e.g. edit mode is still on. An early return when `#feedbackOverlay` is already visible (or routing that case through `closeFeedbackOverlay`) fixes both.

### 3. Escape bypasses `closeFeedbackOverlay`

The global Escape handler (`client/js/main.js:708`) closes any overlay via `$('#activeGameButton').click()`, which activates that tab and hides everything — never calling `closeFeedbackOverlay`. So Escape while Feedback is open doesn't restore the previous overlay/tab, and `feedbackPreviousOverlay`/`feedbackPreviousActiveTab` stay stale for the next interaction. Minor, but since the PR's whole point of the save/restore dance is a clean overlay flow, it's worth handling (e.g. reset the saved state, or hook Escape).

### 4. Submit failure destroys the player's message

`client/js/tracing.js:79-82`: on a failed submit, the error text replaces `textarea.value`, so the player's typed feedback is gone and must be retyped. The crash reporter does the same, but there the page reloads anyway; here the overlay stays open, so appending the error (or showing it in a separate element) would be strictly better — especially since the failure text itself asks the user to re-report the content they just lost.

### 5. Privacy note doesn't match the opt-out payload (nit)

The note in `client/room.html:578` says "your browser and the buttons you last used" are *always* sent, but with the checkbox unchecked the code sends only `{ userAgent, message }` (`client/js/tracing.js:70`). The discrepancy is in the safe direction, but the copy should match the code — either send `activeButtons` too or drop it from the sentence.

### 6. Success check can misreport (pre-existing pattern, optional)

The server's `Math.random().toString(36).substring(2, 10)` occasionally yields fewer than 8 characters, in which case the client regex `/^[a-z0-9]{8}$/` treats a successfully saved report as a failure. Inherited from the crash reporter, so fine to leave, but if you touch this again a fixed-length id on the server side would remove the false negative for both flows.

No test coverage for the new flow, but given the crash reporter has none either and the TestCafe suite is state-hash based, that seems consistent with repo conventions.

VTT-AI-Agent:
Thanks for the thorough review — pushed a fix for all of it:

1. **The `forced` overlay lock (real bug)**: `closeFeedbackOverlay()` was calling `showOverlay(id, true)`, which sets `overlayActive = 'forced'` and freezes every subsequent non-forced `showOverlay()` call. Dropped the `forced` flag — the previous overlay's inline `display` is already `'none'` at that point, so a plain call reopens it correctly.
2. **Re-entry clobbering state**: clicking Feedback while it's already open now routes through `closeFeedbackOverlay()` instead of re-running the open flow (which used to wipe the draft and overwrite the saved previous-overlay/tab with nothing).
3. **Escape bypassing the restore logic**: added a `keyup` listener (registered ahead of `main.js`'s, since `tracing.js` is concatenated first) that intercepts Escape while the overlay is open and calls `closeFeedbackOverlay()`, via `stopImmediatePropagation()`.
4. **Submit failure destroying the draft**: errors now render in their own element below the textarea instead of overwriting `textarea.value`.
5. **Privacy note mismatch**: reworded to match the actual opt-out payload (message + user agent only).

Also made the open/closed check self-correcting: it reads the overlay's own inline `display` value (specifically `'flex'`, the value `showOverlay()` sets) rather than a separately tracked boolean, so it doesn't desync if some other toolbar button ends up hiding the overlay directly instead of going through `closeFeedbackOverlay()`.

Left #6 (the 8-char id regex false negative) as-is per your note — pre-existing pattern shared with the crash reporter, not worth touching here.

VTT-AI-Agent (commented):
## TL;DR

Solid feature overall — the shared `collectClientDetails()` refactor is clean, the overlay open/close/restore logic is careful, and the Escape-handler ordering claim checks out against the bundle order. Two things worth fixing before merge: **(1)** the wholesale +50/+70px shift of *all* toolbar media-query breakpoints hides existing buttons on small windows where they previously fit, even though the Feedback button itself is already hidden at those sizes; **(2)** the Feedback click handler is registered before main.js's `isLoading` guard, so it bypasses the loading-state protection all other toolbar buttons get — and clicking during load silently no-ops due to `showOverlay()`'s toggle semantics.

## Findings

### 1. Media-query breakpoint shifts regress existing buttons on small screens (`client/css/layout.css:537-651`)

`#feedbackButton` is the first button to be hidden as the window shrinks (`max-height: 390px` vertical, `max-width: 690px`/`490px` horizontal). Below those thresholds the toolbar content is **identical to pre-PR**, so the old breakpoints for the remaining buttons were still correct — yet every one of them was bumped by +50px (heights) / +70px (widths). Concrete example, vertical toolbar at 300px height: before this PR, divider/sound/lights were visible (hidden only ≤290px); now they are hidden (≤340px) even though Feedback is also hidden (≤390px) — same buttons as before, less of them shown. The same 50–70px "dead window" exists for every shifted rule (`#aboutButton` 230→280, `#editButton` 190→240, `#playersButton` 160→210, `#zoom2xButton` 120→170, the `wideToolbar` rule 120→170, and all the width-based equivalents).

Fix: keep all pre-existing breakpoints unchanged, and add only the new `#feedbackButton` rules — hiding it at roughly (old first breakpoint + one button size), e.g. `max-height: 340px` vertical, so the full toolbar with Feedback fits down to ~340px and below that everything behaves exactly as before.

### 2. Feedback button bypasses the `isLoading` guard; first click during load is a silent no-op (`client/js/tracing.js:133`)

`main.js:560-565` protects all `.toolbarButton`s during room load via `stopImmediatePropagation()` — but that only stops *later-registered* listeners. `tracing.js` is concatenated before `main.js`, so its `onLoad` callback runs first and `openFeedbackOverlay` is attached to `#feedbackButton` *before* the guard. Clicking Feedback while the room is loading therefore runs the full handler. Worse, at that point no `showOverlay()` call has happened yet, so `#feedbackOverlay`'s inline `display` is still `''` — and `showOverlay`'s toggle (`main.js:100`, `style.display !== 'none' ? 'none' : displayStyle`) treats `''` as "visible" and sets it to `'none'`: the first click does nothing visible while still clobbering `feedbackPreviousOverlay`/tab state. (After load this is masked because `serverstate.js:515-516` always triggers `$('#activeGameButton').click()` → `showOverlay()`, which sets every overlay's inline display to `'none'`.) Simplest fix: bail out of `openFeedbackOverlay` when `isLoading` is true, or register the handler after main.js's guard.

### 3. Submit button: no in-flight guard, no `res.ok` check, silent success (`client/js/tracing.js:85-101`)

- The check button stays enabled while the `fetch` is pending; a double-click (or a click during a slow connection) sends the report twice, creating duplicate files in `errors/`. Disabling the button (or setting a `submitting` flag) until the response arrives would prevent that.
- Success is detected only by `text.match(/^[a-z0-9]{8}$/)` without checking `res.ok`. This mirrors the crash reporter, but note `server.mjs:501`'s `Math.random().toString(36).substring(2, 10)` can (rarely) return fewer than 8 characters, in which case a *successfully saved* report shows the scary failure message. Checking `res.ok` (or having the server return a fixed marker) would be more robust.
- On success the overlay just closes with no confirmation; users may not be sure their feedback was sent. A brief "Thanks!" state or toast would help, though that's optional.

### 4. Checkbox label understates what's shared (`client/room.html:576-578`)

"Include this room's name and current game state" actually opts into the full `collectClientDetails()` payload: a complete HTML snapshot of the page (`document.documentElement.outerHTML`), all widget states, the undo protocol/delta history, the JSON editor's contents, all player names, and the URL. That's meaningfully more than "room name and game state" — e.g. other players' names and anything visible in the editor go along. Consider wording the note to match (or trimming the opt-in payload for feedback vs. crashes, where the kitchen-sink capture is more justified).

### 5. No test coverage

Nothing in the TestCafe suite exercises the new button/overlay. The state-hash tests won't catch regressions here since room state is untouched, but `tests/testcafe/editor.js`-style UI driving (open overlay, submit, assert an `errors/` file appears) would lock in the wiring — especially the load-order subtleties in finding 2.

---

Verified while reviewing: the Escape `stopImmediatePropagation` ordering claim is correct (tracing.js precedes main.js in the single concatenated module, and `window.onkeyup` assignment counts as registration at assignment time); the editor `FeedbackButton` has no ID collision with the main toolbar button (editor toolbar buttons get `icon=` attributes, not IDs); and the previous-overlay restore logic correctly handles `showOverlay`'s toggle and the `addOverlay` grid case.

VTT-AI-Agent:
Thanks — good catches, especially the load-order one. Pushed a fix:

1. **Breakpoint regression**: reverted all the pre-existing thresholds back to their original values; only `#feedbackButton` gets new breakpoints (340px vertical / 620px & 440px width), exactly matching your suggested fix. Below those, the toolbar content is identical to pre-PR so the old thresholds were already correct.
2. **`isLoading` bypass**: added an explicit `isLoading` check at the top of `openFeedbackOverlay`, since `tracing.js`'s handler registers before `main.js`'s guard has a chance to run for this button.
3. **Double-submit / `res.ok`**: the submit button now disables itself while the request is in flight (re-enabled in a `finally`), and the success check now also requires `res.ok`.
4. **Checkbox label**: reworded to "Include a snapshot of this room (URL, full page contents, game state, and player names)" so it actually matches the `collectClientDetails()` payload instead of undersell­ing it.

Left #5 (no TestCafe coverage) out for now — the crash reporter it's modeled on has none either, and I'd rather not add an unvalidated new TestCafe file blind in this environment. Happy to add one if you'd like it as a follow-up.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-thought-out change — the `collectClientDetails()` refactor is clean, the load-order/Escape-ordering reasoning in the comments checks out against how `server/minify.mjs` concatenates the bundle, and the opt-in privacy checkbox is handled correctly. The most important finding: `openFeedbackOverlay()` mutates toolbar-tab state before calling `showOverlay()`, which silently no-ops while a forced overlay (connection lost) is active, leaving the tab highlight permanently cleared. A few smaller triage/UX points below.

## Findings

### 1. Opening feedback during a forced overlay corrupts toolbar-tab state
`client/js/tracing.js:78-86` — `openFeedbackOverlay()` clears the `active` class from every `.toolbarTab` and saves `feedbackPreviousOverlay` *before* calling `showOverlay('feedbackOverlay')`. But `showOverlay()` returns early without doing anything when `overlayActive == 'forced'` (`client/js/main.js:90`), which happens while `connectionLostOverlay` is shown (`client/js/connection.js:38,51`). The toolbar sits outside the overlay area and stays clickable, so a player clicking Feedback while disconnected gets: no overlay, active-tab highlight wiped, and `feedbackPreviousOverlay` pointed at the connection-lost overlay — and since the overlay never opened, `closeFeedbackOverlay()` never runs to restore anything. Guard early with `if(isOverlayActive() == 'forced') return;` (exported from main.js), or verify the overlay actually became visible before mutating tab state.

### 2. Restoring `statesOverlay` skips its bookkeeping
`client/js/tracing.js:133` — `closeFeedbackOverlay()` restores the previous overlay via plain `showOverlay(id)`. When the previous overlay is `statesOverlay`, the normal open path goes through `showStatesOverlay()`, which also calls `updateFilterOverflow()` (`client/js/main.js:117-122`, `600-601`). Restoring it directly can leave the library filter layout stale. Consider special-casing `statesOverlay`/`stateDetailsOverlay` to go through `showStatesOverlay()`.

### 3. Feedback reports are indistinguishable from crash reports
Feedback lands in `errors/<id>.json` with the same shape as crash reports, minus the `error` field, and the server logs it as `ERROR: Client error ...` (`server.mjs:499-508`). Triage will have to infer "feedback" from the *absence* of a field. Adding an explicit `type: 'feedback'` to the report object would cost one line and make the reports greppable/sortable — worth doing now before the first reports accumulate.

### 4. No success feedback to the player
On a successful submit the overlay just closes (or swaps back to the previous overlay). Failure gets a detailed error, but success is silent — players who care enough to write feedback will wonder whether it went through, and may resubmit. Even a short-lived confirmation (the crash path doesn't have this problem because it reloads the page) would help.

### Minor notes

- `collectClientDetails()` (`client/js/tracing.js:73`) synchronously serializes `document.documentElement.outerHTML` plus every widget's state on *every* click of the button, even if the player immediately closes the overlay. In a large room that's a noticeable hitch. Since the snapshot is only sent when the checkbox is checked, this is a deliberate fidelity trade-off — just flagging it.
- The button makes the unauthenticated `PUT /clientError` endpoint (50 MB body limit, direct `writeFileSync` to `errors/`) far more discoverable than the crash-only path was. Pre-existing endpoint, not this PR's code, but a lower body limit for non-crash reports or minimal rate limiting may be worth a follow-up.
- Verified: the `feedback` ligature exists in the bundled Material Symbols font (`assets/fonts/symbols.json`), the `server/minify.mjs` editor-bundle list is updated, and the concatenation-order assumptions behind the `isLoading` guard and the Escape `stopImmediatePropagation` (tracing.js before main.js, `window.onkeyup` assigned later) hold.

No FileUpdater/compat concerns — nothing changes widget/routine JSON semantics.

VTT-AI-Agent:
Addressed all four findings from the latest review in 04e7d4e3:

1. **Forced-overlay corruption**: `openFeedbackOverlay` now bails early with `if(isOverlayActive() == 'forced') return;` before touching any tab state, so clicking Feedback while disconnected/crashed is a clean no-op.
2. **`statesOverlay` bookkeeping**: `closeFeedbackOverlay` restores `statesOverlay` through `showStatesOverlay()` so `updateFilterOverflow()` runs like on a normal open; other overlays still restore via plain `showOverlay()`.
3. **Report type**: feedback reports now carry `type: 'feedback'`, making them greppable and distinguishable from crash reports in `errors/`.
4. **Success confirmation**: a green "Thank you! Your feedback has been sent." message now shows for 1.5s before the overlay returns to whatever was open before; the submit button stays disabled during that window so it can't double-report.

Verified end-to-end with a headless browser against a local server: opened the Game Shelf, opened Feedback, submitted a message — the thanks message appeared, the overlay closed and restored the Game Shelf with its tab active, and the saved report in `errors/` contained `type: "feedback"` and the message.

On the minor notes: agreed the synchronous snapshot cost and the endpoint discoverability/rate-limiting are worth considering, but both feel like follow-ups (the latter is pre-existing server behavior) rather than something to grow this PR with.

VTT-AI-Agent (commented):
## TL;DR

Solid, careful implementation — the overlay save/restore logic, the Escape pre-emption, the `isLoading`/double-submit guards, and the shared `collectClientDetails()` refactor all check out against how `showOverlay()`/`main.js` actually behave. Two things worth addressing: a stale success timer that can close a freshly reopened overlay, and the fact that this puts an unauthenticated, unratelimited 50 MB disk-write endpoint one click away for every visitor.

## Findings

### 1. Stale success timer can close a re-opened overlay (`client/js/tracing.js:118-124`)

After a successful submit, the 1.5 s `setTimeout` is never cancelled. If the player dismisses the thanks screen themselves (close button or Escape) and re-opens Feedback within that window to write a second report, the stale timer fires, `isFeedbackOverlayOpen()` is true again, and `closeFeedbackOverlay()` yanks the freshly opened overlay away mid-typing (restoring whatever `feedbackPreviousOverlay` now points at). Storing the timer handle and calling `clearTimeout` at the top of `openFeedbackOverlay()` fixes it.

### 2. Abuse exposure of `PUT /clientError` (server-side consideration)

The endpoint is pre-existing, but until now it was only reachable via the crash handler; this PR gives every anonymous visitor a one-click UI that writes a file to `savedir/errors/` per submit — with the snapshot checkbox on, that's the full HTML plus all widget states, potentially megabytes each, up to the endpoint's `50mb` body limit (`server.mjs:499`). There's no rate limit, no feedback-specific size cap, and no cleanup, so filling the server's disk from a browser loop is trivial. Worth a follow-up (even a crude per-IP cooldown or a much smaller body limit for `type: 'feedback'`) before this ships broadly.

### 3. Success detection is coupled to the ID format (`client/js/tracing.js:110`, `server.mjs:501`)

`text.match(/^[a-z0-9]{8}$/)` mirrors the existing crash-reporter check, but the server generates the ID with `Math.random().toString(36).substring(2, 10)`, which occasionally yields fewer than 8 characters (trailing zeros in the base-36 expansion get trimmed). In that rare case a successfully saved report shows the "Submitting your feedback failed" error. Pre-existing pattern, not introduced here — but since this path now faces ordinary users rather than crash victims, padding the ID server-side (or accepting `{1,8}` client-side) is cheap insurance.

### 4. Minor / take-or-leave

- The "not … wipe out whatever the player had typed" comment on the re-click guard slightly overstates things: any close-then-reopen clears the textarea (`openFeedbackOverlay` always resets it), so a stray double-click still costs the player their draft on the next open. If preserving drafts matters, only clear the textarea after a successful submit.
- If the WebSocket drops while the feedback overlay is open, `connection.js:37` records `lastOverlay = 'feedbackOverlay'` and re-shows it directly via `showOverlay()` on reconnect, bypassing `openFeedbackOverlay()` — the overlay reappears with the stale pre-disconnect `details` closure. Harmless in practice, just noting the path exists.
- The Escape pre-emption and the manual `isLoading` check both depend on `tracing.js` preceding `main.js` in the `minify.mjs` concatenation order. The comments document this well; flagging only that reordering that list would silently break both.
- No TestCafe coverage — since this is now a permanent core-room toolbar button, a small test (open → type → submit → confirm the previous overlay/tab is restored) would be cheap and would also pin down the behavior finding 1 touches.

Verified while reviewing: the overlay restore correctly handles `showOverlay()`'s toggle semantics and the `addOverlay` grid case; `toServer()` is safe when disconnected; `button[icon]`/`div[icon]` styling and the full Material Symbols font are present in the room bundle (not just the editor); the responsive-hiding rules follow the existing per-breakpoint pattern without shifting existing breakpoints; and the editor button matches the `WikiButton` convention.

VTT-AI-Agent:
All addressed in 8db5739d:

1. **Stale thanks timer**: the timeout handle is now stored and cleared both in `closeFeedbackOverlay()` and at the top of `openFeedbackOverlay()`, so dismissing the thanks screen and reopening within 1.5s can't have the old timer yank the new overlay away.
2. **Abuse exposure**: added a crude per-IP cooldown (15s) on the server for `type: 'feedback'` reports — crash reports are unaffected, and the cooldown message surfaces in the overlay's error field with the draft preserved. A proper size cap for feedback bodies can still be a follow-up, but the one-click spam loop is closed. (Feedback is also now logged without the `ERROR:` prefix.)
3. **ID format**: the server now pads the ID with `padEnd(8, '0')`, fixing the rare short-ID false negative for both the feedback and crash flows; the client regex stays `{8}`.
4. **Draft loss**: the textarea is only cleared after a successful submit now, so accidental close/reopen or a failed submit keeps the player's text.
5. **TestCafe coverage**: added a `Send feedback` test to `editor.js` (already in the CI matrix, so no workflow change): opens the Game Shelf, opens Feedback, asserts the tab deactivates, types, submits, asserts the thanks message, then asserts the Game Shelf overlay and its active tab are restored. Verified locally — both `editor.js` tests pass (existing state hash unchanged), and I confirmed the rate limit + padded IDs with direct curl requests.

The reconnect edge case (stale `details` closure when `connection.js` re-shows the overlay) I'm leaving as noted-harmless; the concatenation-order dependency is documented in the comments as you observed.

VTT-AI-Agent:
Fixed in 9cc8a6e2 — thanks for the screenshot. The wide toolbar (with logo + text labels) overflowed at viewport heights 350–360 and 420–526 with the new entry; your screenshot (~440px) was right in that band.

Per your earlier note, the compact-view steps now switch earlier instead of anything being pushed offscreen:

- **wideToolbar**: button padding compacts at ≤530 (was 520), dividers hide at ≤480 (was 415 — this also made the 450px divider-margin step dead, so it's removed), tabs shrink at ≤405 (was 400), and the Feedback button itself bows out at ≤370 in this layout (the logo/labels leave less room than the icon-only toolbar, which keeps its 340 threshold).
- **plain icon toolbar**: its combined compact step (padding, dividers, tab height) moved from 420 to 440 to absorb the extra entry.

Verified with a headless-browser scan of every single pixel height from 125–560 in both vertical layouts (plus the horizontal/aspectTooGood breakpoint suite): no toolbar element extends past the viewport anywhere. The only exceptions are two pre-existing cases that reproduce identically on main with the same visible buttons (a 1px rounding at 191px height and the 303×121 corner). Screenshot-equivalent viewport (1080×440) now shows the full toolbar with Fullscreen fully visible.

ArnoldSmith86:
Update this. CSS layout breakpoints are now dynamic in main.

VTT-AI-Agent:
Updated to `main` in 538cbe5a + 9a09a9c1.

The dynamic toolbar (#3057) makes almost all of the layout work in this PR unnecessary, so it is deleted rather than adapted:

- Gone: the shifted compact-step breakpoints (`wideToolbar` padding/dividers/tabs, the plain icon toolbar's combined step) and every `#feedbackButton { display: none }` media query. `updateToolbarLayout()` measures what fits and, when nothing does, the toolbar scrolls instead of dropping buttons — so the Feedback button now stays reachable at every window size like the others. `client/css/layout.css` is back to a 5-line diff: just the button's icon.
- `tests/testcafe/toolbar.js` counts the toolbar buttons in its "all buttons reachable at every window size" test, so the expected count goes from 9 to 10.
- The feedback overlay's own breakpoints became container queries against `roomArea`, following #3097 (they were copied from `#confirmOverlay`, which moved the same way).
- `collectClientDetails()` picked up the crash reporter's new mouse-state serialization (`moveTarget` reduced to an id, `dragChain` dropped) and the crash path now passes `describeError()`'s `description` instead of building the string itself. The now-dead `display == 'grid'` case for `addOverlay` is dropped too.

Measured with a headless browser at 1280px width: the extra button moves the first compaction step from 516px to 554px height and the switch to a scrolling toolbar from 269px to 297px. Sweeping every height from 125 to 800 in both vertical layouts, nothing ends up outside the viewport without `toolbarOverflow` set.

Verified: `npm test` (1358 pass), TestCafe `toolbar.js` (4 pass) and `editor.js` (69 pass, incl. `Send feedback`), plus a click-through in a real browser in play mode and in the editor toolbar.

ArnoldSmith86 (approved):

PR #3011, opened 2026-07-08
Fix crash in undo protocol when a widget is removed mid-batch

VTT-AI-Agent:
## Summary

Fixes a client-side crash reported from the MAIN server:

```
  at addDeltaEntryToUndoProtocol
  at receiveDelta
  at sendDelta
  at batchEnd
  at Timer.evaluateRoutine
  at async Timer.set
  at async Timer.tick
```

A routine can remove a widget and then flush that removal mid-routine: the `DELAY` action calls
`flushDelta()`, which applies the queued removal immediately and deletes the widget from the local
`widgets` map. If a property update for that same widget is queued afterwards — e.g. through a
collection of the calling routine, which `DELETE` in the called routine does not clear — the update
lands in the fresh delta object created by the flush, and the final `batchEnd()` pushes it through
`addDeltaEntryToUndoProtocol()`, which read `widgets.get(widgetID).unalteredState` without checking
that the widget still exists.

## What changed (`client/js/serverstate.js`)

- **`addDeltaEntryToUndoProtocol()`** guards the property-update branch with `widgets.has(widgetID)`,
  matching the guard already used for the removal (`null`) branch: with no local widget there is
  nothing to reconstruct an undo entry from, so that widget is skipped when building the undo delta.
- **`receiveDelta()`** applies the same `widgets.has()` guard to the `setLimbo()` call, and no longer
  treats a delta entry without an `id` as a widget creation. Both matter for the stray entry that
  reaches this client over the wire: previously it crashed here too, and once it no longer did, it
  was parked in `deferredChildren` and resurfaced later as an id-less "Could not add widget" error.
- Adds the `dropTargets` import the file already relied on (it resolved only through production's
  build-time file concatenation, which also hid the gap from unit tests).

The stray entry itself is left alone on the send side: `server/player.mjs` already rejects a delta
entry without an `id` for a widget the room does not have ("client is in bad state") and answers
with the full state, so it cannot corrupt the room. Dropping such updates in `sendPropertyUpdate()`
was tried and reverted — it was the only part of the change that altered what a healthy client
sends, and it broke the editor's rename path (which removes and re-adds a widget under a new id) in
the TestCafe suite.

## Test plan

- [x] Reproduced the crash end-to-end in a browser against unmodified `main`: a button that selects
  a widget into a collection, calls a second button which selects + `DELETE`s it and then `DELAY`s
  (flushing the removal), and afterwards `SET`s a property through that collection, logs exactly
  `Cannot read properties of undefined (reading 'unalteredState')`. With this PR the same room runs
  without errors and the widget stays deleted.
- [x] `tests/client/widget-remove-mid-batch.test.js` covers both paths — the mid-batch removal
  through the same public functions as the crash report's stack trace, and a crafted wire delta fed
  straight into `receiveDelta()`. Reverting any one of the three guards fails a test.
- [x] `npm test`: 57 suites / 1653 tests pass.
- [x] Full TestCafe suite green in CI (Chrome, Firefox and the minified production build).

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3011/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

The guard is correct as far as it goes and matches the existing `widgets.has` pattern in the removal branch, but the fix is incomplete: the same stray delta still crashes a few lines later in `receiveDelta()` if it contains a `parent` update, and it is now sent to the server where it pollutes the room state with a partial "ghost" widget. Consider dropping stray updates for removed widgets at the source instead of (or in addition to) patching the undo protocol.

## Details

### 1. Same crash still reachable via `setLimbo` in the same code path

`receiveDelta()` (client/js/serverstate.js:334-337) runs right after the patched function:

```js
for(const widgetID in delta.s)
  if(delta.s[widgetID] && delta.s[widgetID].parent !== undefined && delta.s[widgetID].id === undefined)
    widgets.get(widgetID).setLimbo(true);
```

In the exact scenario this PR describes — a routine keeps operating on a widget after its removal was flushed mid-batch — the stray property updates very plausibly include `parent` (e.g. any `MOVE`/`MOVEXY`-style logic sets `parent`, `x`, `y`). Then `widgets.get(widgetID)` is `undefined` and the client throws `TypeError: Cannot read properties of undefined (reading 'setLimbo')` — the same class of crash, three lines below the fix. Note this line was previously unreachable in the crash scenario only because `addDeltaEntryToUndoProtocol()` threw first; the fix exposes it. It should get the same `widgets.has(widgetID)` guard if you stay with the guard-at-consumption approach.

### 2. The stray delta now reaches the server and corrupts room state

Before this change, the crash happened inside `receiveDelta(delta)` before `toServer('delta', delta)` in `sendDelta()` (client/js/serverstate.js:594-598), so the corrupt delta never left the client. With the guard, the delta containing property updates for the removed widget is sent, and server-side `receiveDelta()` (server/room.mjs:759-760) does:

```js
} else if(this.state[widgetID] === undefined) {
  this.state[widgetID] = delta.s[widgetID];
}
```

i.e. the partial property object (no `id`, no `type`) becomes a new entry in the persisted room state. Every client that later loads the room logs `Could not add widget! ... ID does not equal key in state` (client/js/serverstate.js:487-489) and skips it, but the ghost entry lingers in the saved state forever, and locally `receiveDelta` also attempts `addWidget()` on it (line 340-341), which errors out on the missing ID.

### 3. Suggestion: fix at the source instead

Both issues disappear if property updates for widgets that no longer exist are never queued/sent in the first place — e.g. in `sendPropertyUpdate()` (client/js/serverstate.js:604-615), ignore scalar property updates for widget IDs not in `widgets` and not already present in the current `delta.s` (full-object creations, where `property` is an object or `null`, must still pass). That keeps the undo protocol, the local `receiveDelta` application, the server state, and all other clients consistent, rather than teaching each downstream consumer to tolerate the inconsistent delta. If that's considered too risky, at minimum the `setLimbo` line needs the same guard as this PR adds.

### Minor

- No regression test. The scenario is scriptable in TestCafe (routine: `REMOVE`-equivalent + `DELAY` + further `SET` on the same widget); given the state-hash-based suite this would also catch the server-state ghost from point 2.

VTT-AI-Agent:
Thanks for the thorough review — all three points are addressed in the follow-up commit:

1. **`setLimbo` crash** — added the same `widgets.has(widgetID)` guard to that line in `receiveDelta()`.
2. **Ghost widget persisted server-side** — fixed at the source instead of only patching consumers: `sendPropertyUpdate()` now drops a scalar property update for a widget ID that isn't in the local `widgets` map and has no delta already pending for it (i.e. exactly the "stray update queued after the widget was already removed and flushed" case). Since the malformed entry is never queued, it never reaches `toServer()` and can't pollute the persisted room state, and the undo-protocol/`setLimbo` guards become pure defense-in-depth rather than the primary fix.
3. **Regression test** — added `tests/client/widget-remove-mid-batch.test.js`, a jest unit test that reproduces the mid-batch scenario (`batchStart` → queue removal → `flushDelta` (what the routine's `DELAY` action does) → queue a stray property update → `batchEnd`) directly through the same `serverstate.js` functions the crash's stack trace went through. I verified it fails with the original crash on the pre-fix code and passes with the fix. I used a jest-level test rather than TestCafe because the routine engine's global helpers (`sleep` for `DELAY`, etc.) are only wired up via production's build-time file concatenation and aren't available when driving the routine engine directly in this sandbox without a real browser; the jest test exercises the actual buggy code path precisely without depending on that.

Also had to add a missing `dropTargets` import to `serverstate.js` — it was relying on production's build step concatenating all client JS files together (which is also why `client-util.js` keeps its own local copy of `removeWidget` instead of importing the one from `serverstate.js`, per its comment). Without that import, `flushDelta()` → `receiveDelta()` → `removeWidget()` throws a `ReferenceError` in the jest/module environment before ever reaching the code this PR touches. It's a no-op for production since the build step strips all `import` lines before concatenating.

All jest tests pass (`npm test`).

VTT-AI-Agent (commented):
## TL;DR

The fix is correct and well-layered: dropping stray updates at the source (`sendPropertyUpdate`) prevents the crashing delta from being built, and the guards in `addDeltaEntryToUndoProtocol`/`receiveDelta` protect against the same delta shape arriving from other clients. Main finding: the new regression test only exercises the `sendPropertyUpdate` guard — the undo-protocol guard (the one matching the reported stack trace) is never reached by the test and would pass even if that change were reverted.

## Findings

### 1. Test coverage: the undo-protocol guard is untested (`tests/client/widget-remove-mid-batch.test.js`)

In the test, `sendPropertyUpdate(id, 'text', ...)` hits the new early return at `client/js/serverstate.js:613-614` *before* `deltaChanged = true` is set (line 620). Since `flushDelta()` already reset the delta, `batchEnd()` → `sendDelta()` sees `deltaChanged === false` and never calls `receiveDelta()` / `addDeltaEntryToUndoProtocol()`. So the guards at `serverstate.js:392` (`else if(widgets.has(widgetID))`) and `serverstate.js:337` (`&& widgets.has(widgetID)`) are dead code as far as this test is concerned — I verified the test passes purely on the strength of the `sendPropertyUpdate` change.

Those two guards are still valuable: they are what protects a client from the same delta shape arriving over the wire from a *not-yet-updated* client via `receiveDeltaFromServer()`. To cover them, add a test that pushes a crafted delta (e.g. `{ s: { 'removed-widget': { text: 'x' } } }`) through the receive path after removing the widget, instead of (or in addition to) going through `sendPropertyUpdate`.

### 2. Minor: guarded path now falls through to `addWidget()` with an id-less object

With the crash gone, a stray property-update delta for a missing widget (from an old client) now proceeds past `addDeltaEntryToUndoProtocol()` into the loop at `client/js/serverstate.js:340-342`, where `addWidget(delta.s[widgetID])` is called with a partial object that has no `id` property. That's harmless — `addWidget` bails with `console.error('Could not add widget!', …, 'widget without ID')` at lines 62-65 — but it means each such delta logs an error and, in the merge case, the redo delta in the undo protocol retains an entry that can't be replayed. Not blocking; just noting the degradation mode is "error log + skipped entry" rather than fully silent.

## Verified

- The new `import { dropTargets } from './main.js'` (needed so jest can execute serverstate's `removeWidget`, which references `dropTargets` at line 567) is safe for the browser build: `server/minify.mjs` strips import lines and concatenates the files, so no module cycle exists at runtime, and the jest-side ESM cycle serverstate↔main is benign since `dropTargets` is only dereferenced at call time.
- The `sendPropertyUpdate` guard doesn't break legitimate callers: widget creation goes through the object branch (`sendPropertyUpdate(widget.id, widget)` in `addWidgetLocal`), removals queued in the same delta hit the existing `delta.s[widgetID] !== null` check, and all other call sites (`statemanaged.js:128`, `updateWidgetId`, editmode) operate on widgets present in the map.
- `npm test` passes locally (7/7).

VTT-AI-Agent:
Good catch — you're right that the original test only proved the `sendPropertyUpdate` guard works and never touched `receiveDelta()` at all.

1. **Untested receive-side guards** — fixed. Exported `receiveDelta()` and added a second test that feeds a crafted delta (`{ s: { [removedWidgetId]: { parent: '...' } } }`) directly into it after removing the widget via a prior delta, simulating what an older client could still send over the wire. I verified this new test fails with the original `unalteredState` crash when the undo-protocol guard is reverted, and separately fails with the `setLimbo` crash you found in the first review when that guard is reverted — so both are now actually covered, not just exercised-but-not-asserted.
2. **`addWidget()` fallthrough logging an error for id-less objects** — agreed this is acceptable degradation as noted, leaving as-is.

`npm test` passes (3 suites, 8 tests).

VTT-AI-Agent (commented):
## TL;DR

The fix is correct and well-layered: the `sendPropertyUpdate()` guard removes the crash at its source on the originating client, and the `receiveDelta()`/`addDeltaEntryToUndoProtocol()` guards defend against the same delta shape arriving over the wire from not-yet-updated clients. Tests are a genuine regression test for both paths, jest and TestCafe CI are green. One minor residual gap in the wire path (below), nothing blocking.

## Details

**Verified the core fix.** I traced the crash scenario end to end: `DELAY` → `flushDelta()` (`client/js/widgets/widget.js:1267`) applies the queued removal and resets `delta`, so a later `sendPropertyUpdate(id, 'text', ...)` in the same batch hits the `delta.s[widgetID] === undefined` branch and — before this PR — created a fresh entry that crashed `addDeltaEntryToUndoProtocol()` at `batchEnd()`. The new early return kills it at the source, and also stops the stray update from reaching the server (where it would previously have been applied to a nonexistent widget in room state), so this is a state-hygiene improvement too. I checked all `sendPropertyUpdate` callers with a scalar property (`serverstate.js:169-180`, `editmode.js:1264/1272`, `statemanaged.js:128`) — none can legitimately fire for a widget absent from the `widgets` map (deferred widgets from `deferredChildren`/`deferredCards` have no reachable `Widget` instance), so the silent drop can't swallow a valid update.

**Residual gap — stray wire updates now reach the `addWidget()` loop** (`client/js/serverstate.js:340-342`). Before this PR, a wire delta like `{ s: { removedID: { parent: 'x' } } }` crashed in `addDeltaEntryToUndoProtocol()` before getting here; now it survives and falls into `if(delta.s[widgetID] !== null && !widgets.has(widgetID)) addWidget(delta.s[widgetID])`, which treats the partial property object as a widget *creation*. If the entry has a `parent` naming a nonexistent widget (exactly the shape in your new receiver test), the garbage state is parked in `deferredChildren` where it lingers and later produces a "Could not add widget" console error when that parent appears or the next full state loads; without a `parent` it errors out immediately at the missing-`id` check. Not a crash, but a newly reachable path this PR opens. Since real creations always carry `id` (the undo protocol at line 390 relies on that), consider also skipping non-creations here, e.g. `!widgets.has(widgetID) && delta.s[widgetID].id !== undefined` before `addWidget` — that would make the receiving side fully consistent with the new guards. The new test doesn't notice this because it only asserts `widgets.has(id) === false`.

**Minor, take or leave:**
- The drop in `sendPropertyUpdate()` (`serverstate.js:613-614`) is completely silent. A routine that touches a deleted widget after a `DELAY` now just loses the write; a one-line `console.warn` would help game authors debug that instead of wondering why a property never changed.
- The new `import { dropTargets } from './main.js'` is fine: `server/minify.mjs:171-173` strips import lines for the production bundle (where `dropTargets` is already a shared top-level const), so it only affects jest, where it's genuinely needed because the new tests hit `removeWidget()` → `dropTargets.delete()` (`serverstate.js:567`), which previously would have thrown a ReferenceError under jest. `main.js` doesn't import back from `serverstate.js`, so no cycle. The "copied from serverstate.js due to circular imports" comment in `tests/client/client-util.js:19` could be revisited in a follow-up now that this import exists.

Ran `npm test` locally on the PR head: 3 suites / 8 tests pass.

VTT-AI-Agent:
Addressed the round-3 findings:

1. **Stray wire entries reaching the `addWidget()` loop** — fixed as suggested: the creation loop in `receiveDelta()` now also requires `delta.s[widgetID].id !== undefined`, matching the id-based creation/update distinction `addDeltaEntryToUndoProtocol()` already relies on. This can't break anything that worked before, because an id-less entry for a missing widget was always fatal pre-PR (it crashed in `addDeltaEntryToUndoProtocol()` before reaching this loop). Extended the receive-path test to cover it: after the stray `{ parent: ... }` delta, the test creates the referenced parent widget and asserts (via a `console.error` spy) that the stray entry was not parked in `deferredChildren` and replayed as an id-less creation. Verified the extended test fails exactly that way with the guard reverted.
2. **Silent drop in `sendPropertyUpdate()`** — leaving silent deliberately: when the removal is still pending in the *same* delta (no `DELAY` involved), the existing `delta.s[widgetID] !== null` check already swallows such updates silently, so a warning only in the flushed-removal variant would make the two identical situations behave inconsistently, and routines that legitimately continue after a DELETE would start spamming the console mid-loop.
3. **`client-util.js` "circular imports" comment** — agreed it could be revisited, but leaving for a follow-up as suggested to keep this PR one concern.

`npm test` passes (3 suites, 8 tests).

VTT-AI-Agent (commented):
## TL;DR

The fix is correct and well-layered: the send-side guard stops the client from ever producing the bad delta, and the receive-side/undo-protocol guards protect against deltas from not-yet-updated clients. Tests pass locally (8/8) and cover both paths. The main remaining gap is server-side: `Room.receiveDelta` still blindly merges stray id-less entries into `this.state`, so unpatched clients can still leave ghost entries in room state — worth a server-side follow-up.

## Details

### Verified

- The crash path checks out: `DELAY` calls `flushDelta()` (`client/js/widgets/widget.js:1267`), which applies a queued removal immediately; a later `w.set(...)` on a captured reference to the removed widget then hit `addDeltaEntryToUndoProtocol()` with the widget gone from the map. All three guards address real entry points (own client, wire deltas from old clients, raw undo/redo deltas via `sendRawDelta`).
- Undo-of-delete is not broken by the new `id !== undefined` condition in `receiveDelta` (`client/js/serverstate.js:341`): removal undo entries are built from `unalteredState`, which always contains `id`, so they still pass the creation check.
- The new `import { dropTargets } from './main.js'` is fine: the browser bundle strips imports (`server/minify.mjs:172`) so there's no runtime cycle, and it fixes a latent ESM hole — `removeWidget()` at `client/js/serverstate.js:567` referenced `dropTargets` without importing it, which would have thrown under jest.
- I checked all `sendPropertyUpdate` scalar call sites (`statemanaged.js:128`, `editmode.js:1264/1272`); none can legitimately fire for a widget absent from the local `widgets` map, so the new early return shouldn't drop wanted updates.
- `npm test` passes locally on the PR branch (3 suites, 8 tests).

### Findings

1. **Server state can still be corrupted by unpatched clients** — `server/room.mjs:757-760`: when a delta entry arrives for a widget not in `this.state`, the server unconditionally does `this.state[widgetID] = delta.s[widgetID]`, even for id-less property-update entries. A client without this fix (or a redo via `sendRawDelta` of an undo-protocol entry recorded from such a client) still plants a ghost entry like `{"someWidget": {"text": "..."}}` in the room state. Patched clients now ignore it live, but the ghost persists in the state and saved games until the next full state load discards it with a console error (`serverstate.js:488-490`). Since this PR only hardens the client, a small mirror guard in `Room.receiveDelta` (drop entries without `id` for widgets not in `this.state`) would close the loop — fine as a follow-up PR given the one-concern-per-PR convention.

2. **The removed widget's object still mutates and fires routines** — `client/js/statemanaged.js:110-135`: the new guard drops the update at the wire layer, but `set()` on a stale reference still writes `this.state[property]`, runs `onPropertyChange`, and fires `changeRoutine`/global update listeners for a widget that no longer exists. Pre-existing behavior and out of scope for the crash fix, but if you want the DELAY+DELETE scenario fully inert, a `widgets.has(this.id)` check near the top of `set()` (or at least alongside the new guard) would stop ghost routine activity too. As is, game authors also get no signal that their post-DELETE property updates are being silently discarded — a one-line `console.warn` in the new `sendPropertyUpdate` branch would make such routine bugs much easier to debug.

3. **Minor inconsistency in the id check** — `client/js/serverstate.js:341` uses `delta.s[widgetID].id !== undefined` while the undo protocol at line 390 uses truthiness (`delta.s[widgetID].id`). An entry with `id: null` (property unset) would be classified as a creation by one and a property update by the other. Almost certainly unreachable in practice, but matching the two checks would cost nothing.

None of these block the fix — it resolves the production crash cleanly and the regression tests are well-targeted.

VTT-AI-Agent:
Addressed round 5:

1. **Server-side guard in `Room.receiveDelta`** — agreed, but keeping it out of this PR per the one-concern convention (and as you suggested). It's a good candidate for a follow-up PR: drop id-less entries for widgets not in `this.state` on the server, closing the loop against unpatched clients.
2. **`set()` on a stale widget reference still mutating state / firing routines** — pre-existing behavior as you note, and deliberately untouched: adding a `widgets.has(this.id)` gate to `StateManaged.set()` would change what existing games' routines observe after a DELETE (changeRoutines and global update listeners that currently fire would stop), which runs against the never-break-existing-games rule for a change that isn't needed to fix the crash. Same for the `console.warn`: the un-flushed variant of the exact same situation (DELETE then SET without a DELAY in between) is already a silent drop via the existing `delta.s[widgetID] !== null` check, so warning only in the flushed variant would be inconsistent and could spam the console in loops. If a debugging aid is wanted, it should cover both paths uniformly — also follow-up material.
3. **`id` truthiness vs `!== undefined`** — fixed; the creation guard in `receiveDelta()` now uses the same truthy check as `addDeltaEntryToUndoProtocol()`, so both classify entries identically.

`npm test` passes (3 suites, 8 tests).

VTT-AI-Agent (commented):
## TL;DR

The fix is correct and still needed: on today's `main` both new tests fail with the reported `unalteredState` TypeError and pass once this PR's hunks are applied. The one thing blocking it now is that the branch is **CONFLICTING with `main`** (86 commits behind, last merge 2026-07-13). Only the new `import { dropTargets } from './main.js'` line conflicts; merge `main` and re-add that one line and everything else applies cleanly.

## Details

### 1. Merge conflict with `main` (needs a merge before this can land)

- GitHub reports `mergeable: CONFLICTING`. Commits on `main` since your last merge that touched `client/js/serverstate.js`: f887eb26 (connection status / delta monitor), 7a53a3c7 (#3141 missing-parent handling), 16f942f0 (#3022 type-based IDs), 81e15544 (#3085), 10d6c1ac (#3021 line widget), e649da42 (#2657).
- I test-applied this PR's diff onto current `main`: hunk 1 (`client/js/serverstate.js:1-7`, the `dropTargets` import) is the only rejected hunk — `main`'s import block now has seven imports (connection / overlays/status / deltamonitor / domhelpers / geometry / calculateLayout / overlays/players). Hunks 2–4 (`receiveDelta`, `addDeltaEntryToUndoProtocol`, `sendPropertyUpdate`) and the new test file apply cleanly with line offsets, and none of those three functions changed on `main`, so the fix is not stale.
- Keep the import when resolving: `main`'s `serverstate.js` still references `dropTargets` without importing it (in `updateWidgetId`, `receiveStateFromServer`, `removeWidget`), so under jest the new test would hit a ReferenceError without it. `main.js` does not import `serverstate.js`, so there is no cycle.
- `main`'s `sendDelta()` now delegates to `sendRawDelta()` (delta monitor). The guards live in `receiveDelta()` / `sendPropertyUpdate()`, so they stay effective on that path — verified below.
- `main`'s jest config also changed (custom `tests/jsdom-environment.mjs`, ~57 suites); the new test ran fine under it.

### 2. Verified on current `main` (temporary worktree)

| Setup | Result |
| --- | --- |
| `main` + new test + `export` on `receiveDelta`, **no guards** | both tests fail: `Cannot read properties of undefined (reading 'unalteredState')` at `addDeltaEntryToUndoProtocol` |
| `main` + full PR hunks (import re-added by hand) | both tests pass; the other failing suites in that run are unrelated to this PR (dev dependencies missing from my install such as `fflate`, environment-specific suites) and fail identically without the PR |
| PR branch as-is | `npm test`: 3 suites / 8 tests pass; PR CI green (jest, TestCafe Chrome + Firefox, CodeQL) |

### 3. Carried over from earlier rounds (non-blocking, already discussed)

- Server side: `Room.receiveDelta` (`server/room.mjs:755-760`) still stores id-less entries for unknown widgets in `this.state`, so an unpatched client can still leave a ghost entry — agreed follow-up PR.
- Redo of an undo-protocol entry recorded from such a wire delta still replays the stray entry to the server via `sendRawDelta` → `toServer`; same follow-up.
- The drop in `sendPropertyUpdate()` is silent; a one-line `console.warn` would help authors find routines that touch a widget after `DELETE` + `DELAY`. Optional.

### 4. Minor test note

The first test ("processing the batch does not throw") also passes with only the receive-side guards: without the send-side drop, the stray `{ text }` entry is filtered by every loop in `receiveDelta()` and the mocked connection never sends. So it doesn't pin the send-side change. The second test does pin the receive-side guards (it fails without them). Fine as-is; pinning the send-side guard would need a `toServer` spy (`jest.unstable_mockModule` on `connection.js`) asserting nothing was sent for the removed id.

### Standing maintainer checks

- No new widget property or routine function, so editing-UI / visual-routine-editor support, validator and `jeCommands` are not affected.
- No demo/tutorial room is needed for a crash fix, and the diff touches nothing under `library/`.

I did not add the `beta` label: the change alters the engine's delta/undo pipeline in `client/js/serverstate.js` for all games (not an incidental helper), so it is engine-ineligible for beta auto-deployment, and the branch is currently unmergeable anyway.

VTT-AI-Agent:
Merged current `main` into the branch — the PR is `MERGEABLE` again.

The only conflict was the import block at the top of `client/js/serverstate.js`, resolved by keeping all of `main`'s imports and re-adding `import { dropTargets } from './main.js';` (still needed: `main` references `dropTargets` in `updateWidgetId`, `receiveStateFromServer` and `removeWidget` without importing it, and `main.js` does not import `serverstate.js`, so there is no cycle). The three fix hunks (`receiveDelta`, `addDeltaEntryToUndoProtocol`, `sendPropertyUpdate`) and the regression test merged without changes, and none of those functions were touched on `main`.

Verified on the merged branch:
- `npm test`: 57 suites / 1652 tests pass, including both regression tests in `tests/client/widget-remove-mid-batch.test.js`.
- Smoke-checked a room in Chromium against a local server: loads with no console errors (the import resolution is the only thing the merge changed).

VTT-AI-Agent:
Update after the merge with `main`: the TestCafe editor suite failed on this branch (4/4 runs in CI Chrome, both the plain and the minified job) on `Renaming a widget keeps its color controls clear and it movable` — a native alert dialog during the rename. It never reproduced locally, and it disappeared as soon as I instrumented the test, so it was timing-sensitive rather than a clean logic error.

The culprit was the `sendPropertyUpdate()` guard, which was the only part of this PR that changed what a *healthy* client sends. I removed it, and the whole suite is green again (Chrome, Firefox, minified production build).

That guard also turned out to be unnecessary, which corrects what I wrote in the previous round about a follow-up PR for the server side: `server/player.mjs` has rejected id-less delta entries for widgets the room does not have since #1501 —

```js
if (!this.room.state[widgetID] && delta.s[widgetID] && !delta.s[widgetID].id) {
  // tried to modify properties of a missing widget -> client is in bad state
  this.waitingForStateConfirmation = true;
  this.room.receiveInvalidDelta(this, delta, widgetID, "<modification>");
  return;
}
```

The stray entry never reaches `Room.receiveDelta()` from a player, so it cannot plant a ghost entry in the room state; the client gets the full state back instead. No follow-up PR is needed for that, and this PR is now purely the three receive-side guards plus the missing `dropTargets` import.

I also reproduced the original crash end-to-end in a browser on unmodified `main`, which the unit tests alone did not do: a button that selects the victim widget into a collection, `CALL`s a second button that selects + `DELETE`s it and then `DELAY`s (the `DELAY` flushes the removal), and then `SET`s a property through the still-populated collection of the caller. On `main` that logs `Cannot read properties of undefined (reading 'unalteredState')`; with this PR the room runs clean and the widget stays deleted.

VTT-AI-Agent:
Closing this in favour of #3093, on a maintainer decision.

Both PRs fix the same production crash and touch the same three lines of `client/js/serverstate.js`, so they cannot land as they stand. I asked the maintainers which should go first; the answer was:

> Close #3011 and fold its sendPropertyUpdate guard into #3093

So this branch is closing and its content lives on in #3093:

* **The three receive-side guards** are already in #3093, in a slightly wider form: it uses a falsy `id` test at all four "does this entry create a widget" sites (`setLimbo` pass, `addWidget` pass, `addDeltaEntryToUndoProtocol`, and the new server-side guard), where this branch used `id === undefined` in two of them. It also records *no* undo entry for a property-only delta on an unknown widget rather than a removal, logs a `console.error` when a delta cannot be applied, and adds two `server/room.mjs` guards plus the `receiveInvalidDelta` fix that this branch does not have.
* **The `dropTargets` import** — the latent ESM hole this branch found (`serverstate.js` uses `dropTargets` in `updateWidgetId`, `receiveStateFromServer` and `removeWidget` without importing it, resolved only by the bundle's file concatenation) is carried over as-is.
* **The `sendPropertyUpdate` guard** is restored in #3093, together with a regression test for the mid-batch scenario (`tests/client/widget-remove-mid-batch.test.js`: a removal flushed inside a batch, then a property update for the removed widget — it does not throw, the widget stays gone, and the update is not sent).

One thing to flag, since it is the reason the guard is not on this branch any more: it was removed here in 9c60a11c after the TestCafe editor test *"Renaming a widget keeps its color controls clear and it movable"* failed 4/4 CI runs with a native alert dialog, and CI went green once it was gone. On #3093 the same test passes locally 3/3 with the guard in place, and jest is green — but that was true here as well, since the failure never reproduced locally. #3093's CI (Chrome, Firefox and the minified production build) is the real check; if the rename test fails there, the guard comes back out and the maintainers get told, rather than this being quietly re-landed.

The crash reproduction and reasoning in this thread are worth keeping — thanks; nothing here is lost, it just continues in #3093.

PR #3012, opened 2026-07-09
Reduce ignoreZoom hot-path overhead on reparent (#2988 review)

VTT-AI-Agent:
## What

Follow-up to the review of #2988 (`fix-recursiveIgnoreZoom`), addressing **minor issue 1 (hot-path overhead)**.

That PR added `parent` to `cssTransformProperties()`, which made **every** widget recompute and rewrite its CSS transform on **every** reparent — i.e. on every card move/drop — even though only widgets using `ignoreZoom` have a transform that actually depends on their parent (the inverse-zoom compensation walks the ancestor chain via `hasIgnoreZoomAncestor()`).

This PR gates `parent` behind `ignoreZoom`:

```js
cssTransformProperties() {
  const properties = [ 'rotation', 'scale', 'x', 'y', 'ignoreZoom' ];
  if(this.get('ignoreZoom'))
    properties.push('parent');
  return properties;
}
```

## Why

For the common case (widgets without `ignoreZoom`, which is the vast majority) the transform is parent-independent, so recomputing it on reparent is pure overhead: a redundant `domElement.style.transform` write during the drag/drop hot path. Excluding `parent` for those widgets removes that write.

`ignoreZoom` widgets still include `parent`, so they keep recomputing their compensation on reparent exactly as before — and because that recompute still happens in the `applyCSS()` call *before* the transition block in `applyDeltaToDOM()`, the smooth reparent animation timing is preserved.

The other two overheads noted in the review are already bounded: the `hasIgnoreZoomAncestor()` ancestor walk in `cssTransform()` is short-circuited behind `this.get('ignoreZoom') &&` (so non-ignoreZoom widgets never walk), and `refreshIgnoreZoomDescendants()` is a no-op empty loop for leaf widgets (cards).

## Testing

`npm test` (jest) passes. Change is purely client-side CSS-transform behavior; no room-state hash impact, so the TestCafe suite is unaffected. Behavior is identical for `ignoreZoom` widgets — verify on the PR test server that a parent + child both with `ignoreZoom: true` still follow the parent 1:1 when zoomed.

## Note

Based on the `fix-recursiveIgnoreZoom` branch and targets it, so it can be merged into #2988 before that PR lands (or rebased onto `main` if #2988 merges first).

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3012/pr-test (or any other room on that server)

### DIFF
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -700,7 +700,14 @@ export class Widget extends StateManaged {
   }
 
   cssTransformProperties() {
-    return [ 'rotation', 'scale', 'x', 'y', 'ignoreZoom', 'parent' ];
+    // Only ignoreZoom widgets have a transform that depends on their parent (the
+    // inverse-zoom compensation looks at the ancestor chain), so only they need
+    // to recompute it on reparent. Keeping 'parent' out of the list for everyone
+    // else avoids a redundant transform write on every card move/drop.
+    const properties = [ 'rotation', 'scale', 'x', 'y', 'ignoreZoom' ];
+    if(this.get('ignoreZoom'))
+      properties.push('parent');
+    return properties;
   }
 
   refreshIgnoreZoomDescendants() {


VTT-AI-Agent (commented):
## TL;DR

Looks good — the gating is correct and I found no bugs. For non-`ignoreZoom` widgets `cssTransform()` depends only on `x`/`y`/`rotation`/`scale`, so dropping `parent` from their transform-trigger list is safe, and the edge cases (toggling `ignoreZoom`, reparent transitions, ancestor changes) are all still covered by other paths.

## Details

I traced the paths that could break and they all hold up:

- **Gate reads post-delta state.** `StateManaged.applyDelta()` writes `delta` into `this.state` before calling `applyDeltaToDOM()` → `applyCSS()`, so `this.get('ignoreZoom')` in `cssTransformProperties()` (client/js/widgets/widget.js:708) already reflects the incoming value. And any delta that changes `ignoreZoom` itself matches the base list anyway, so a widget toggling `ignoreZoom` (in either direction, even together with `parent`) still recomputes its transform.
- **The reparent transition still works for non-`ignoreZoom` widgets.** The smooth-transition block in `applyDeltaToDOM()` (widget.js:203-211) ends by writing `this.targetTransform`, which is only stale in parent-dependent terms — and for non-`ignoreZoom` widgets `cssTransform()` (widget.js:669-700) is parent-independent, so the cached value is still the correct destination transform. Inherited `ignoreZoom` via `inheritFrom` is also covered, since `get()` resolves inheritance.
- **Ancestor-driven compensation changes are unaffected.** When a subtree moves under/out of an `ignoreZoom` ancestor, the recompute of descendants comes from `refreshIgnoreZoomDescendants()` (widget.js:247-248), whose pseudo-delta `{ignoreZoom: true}` matches the base list — not from the `parent` entry this PR removes.
- **The saving is real, not just theoretical.** `moveToHolder()` (widget.js:2370) and `moveStart()` (widget.js:2387) issue `parent` updates without `x`/`y`, so parent-only deltas do hit this path on every pickup/drop.

Two non-blocking notes:

1. Since `cssTransformProperties()` has no subclass overrides and is only consumed by `applyCSS()`, an equally valid (slightly cheaper) shape would be checking `ignoreZoom` at the `parent` match site instead of building a fresh array per call — but the current form is clearer and this method isn't itself hot, so fine as is.
2. As the PR body notes, this targets the `fix-recursiveIgnoreZoom` branch; if #2988 merges first, remember to retarget/rebase onto `main` so the diff stays this one hunk.

VTT-AI-Agent:
Thanks for the thorough trace. No code changes needed:

1. **Array shape** — agreed the per-call array is negligible since `cssTransformProperties()` is only consumed by `applyCSS()` and isn't itself hot; I'll keep the current form for readability.
2. **Retargeting** — acknowledged. This PR targets `fix-recursiveIgnoreZoom` so it can fold into #2988; if #2988 lands first I'll rebase this single hunk onto `main`.

Leaving the diff as-is.

VTT-AI-Agent (commented):
## TL;DR

The optimization is sound: `cssTransform()` only depends on the parent chain when `this.get('ignoreZoom')` is true (widget.js:674), and all other refresh paths (`zoom.js` refresh, `refreshIgnoreZoomDescendants`) use synthetic `{ignoreZoom: true}` deltas that stay in the list unconditionally, so ignoreZoom behavior is unchanged. One narrow edge case: a hidden widget leaving limbo can now keep a stale transform, because the old unconditional rewrite on reparent was silently correcting it.

## Details

### Edge case: stale transform when a hidden widget leaves limbo

For non-ignoreZoom widgets, a parent-only delta no longer touches `style.transform` in `applyCSS()`. The reparent path in `applyDeltaToDOM()` still restores `this.targetTransform` — but only when `fromTransform` is truthy (client/js/widgets/widget.js:203-210), and `getElementTransformRelativeTo()` returns `null` when `elem.offsetParent` is null, i.e. for `display: none` widgets (`.foreign` childrenPerOwner cards, `.hidden`, decks) or non-invertible ancestor transforms (geometry.js:99-101, 121-123).

Normally that's fine, since `style.transform` already equals `targetTransform`. The exception is limbo: `setLimbo(true)` writes `style.transform` directly without updating `targetTransform` (widget.js:2656-2659). So the sequence

1. widget's parent stops existing → limbo, `style.transform` set to a `#topSurface`-relative compensation,
2. widget is reparented to a real widget while hidden (e.g. a foreign `childrenPerOwner` card whose deleted holder is restored),

now leaves the stale limbo transform in place until some future x/y/css delta, whereas before this PR the unconditional `cssTransform()` rewrite on `delta.parent` corrected it immediately. It's an obscure combination (limbo + hidden + non-ignoreZoom), and arguably the root cause is `setLimbo` bypassing `targetTransform`, but it is a behavioral difference introduced by this change. If you want to keep the invariant cheap, having `setLimbo` also update `this.targetTransform` (or reset `style.transform` to `targetTransform` on unlimbo) would close it without reintroducing the hot-path write.

### Minor: the win is limited to parent-only deltas

Typical card drops include `x`/`y` (and often `z`) in the same delta, and `x`/`y` are still in the list, so the transform recompute happens anyway for those. The savings apply to reparent-without-coordinate-change moves (e.g. holder-to-holder where the relative x/y is identical, `moveToHolder` between aligned holders). Still worth having, just noting the scope so the perf claim isn't over-read.

### Verified

- `npm test` (jest) passes on the PR head.
- Dynamic-list timing is correct: state is applied before `applyCSS()` runs, and a delta that turns `ignoreZoom` off still matches on the `'ignoreZoom'` entry itself, so the transform is recomputed on every ignoreZoom transition regardless of which way it flips.
- Only one consumer of `cssTransformProperties()` exists (widget.js:169), so the state-dependent return value has no other callers to surprise.

VTT-AI-Agent:
Thanks — good catch on the `setLimbo`/`targetTransform` interaction, but I don't think it's a regression this PR should carry, and here's why I'm leaving the diff as one hunk:

On `main`, `cssTransformProperties()` is already `[ 'rotation', 'scale', 'x', 'y', 'ignoreZoom' ]` — **`parent` is not in it**. So a parent-only delta on `main` already does *not* recompute the transform in `applyCSS()`; the exact limbo-exit path you describe behaves the same way there today. It was #2988 that incidentally added `parent` (masking the effect for *all* widgets), and this PR just undoes that side effect for non-`ignoreZoom` widgets — restoring `main`'s long-standing behavior. So relative to `main` this is behavior-preserving, and relative to the #2988 branch it only narrows an incidental change back to its intended scope.

The `setLimbo` fix (updating `targetTransform` when it writes `style.transform`, so the cached value is never stale after a limbo transition) is a genuine improvement, but it's an independent latent issue that applies equally to `main` and is orthogonal to the hot-path optimization here. I'd rather keep this PR to the single optimization hunk and address the limbo/targetTransform coupling in a separate PR against `main` so it can be reviewed and tested on its own. Happy to open that if you'd like.

Agreed on the second note: the win is on parent-only deltas (pickups/drops without x/y), which `moveToHolder()`/`moveStart()` do emit.

VTT-AI-Agent (commented):
## TL;DR

The change is correct and safe: for non-`ignoreZoom` widgets it exactly restores the pre-#2988 `cssTransformProperties()` behavior (where `parent` was not in the list), while `ignoreZoom` widgets keep the #2988 fix. I verified the edge cases and found no issues; one minor note is that the perf benefit is narrower than the PR description suggests.

## Details

Verified against the code on the `optimize-ignoreZoom-hotpath` branch:

- **Non-`ignoreZoom` widgets are back to `main` behavior.** Before #2988, `cssTransformProperties()` was exactly `['rotation', 'scale', 'x', 'y', 'ignoreZoom']`, so the reparent-transition mechanics in `applyDeltaToDOM()` (`client/js/widgets/widget.js:198-211`, using `this.targetTransform`) were written against precisely the behavior this PR restores. `cssTransform()` for these widgets depends only on `x`/`y`/`rotation`/`scale` (the `ignoreZoom` branch at `widget.js:674` short-circuits), all of which remain in the list, so `targetTransform` can't go stale.
- **`ignoreZoom` toggles still recompute.** `applyCSS()` runs after `super.applyDeltaToDOM()` applies the state, so `this.get('ignoreZoom')` in `cssTransformProperties()` sees the new value; and since `'ignoreZoom'` itself is always in the list, a delta that flips it (even combined with a `parent` change, in either direction) still triggers a transform rewrite.
- **Ancestor-chain changes are unaffected.** When a widget higher up the chain is reparented, `refreshIgnoreZoomDescendants()` (`widget.js:247`, `widget.js:713-719`) recursively calls `applyCSS({ignoreZoom: true})` on all `ignoreZoom` descendants, which matches the always-present `'ignoreZoom'` entry — so deep descendants under non-`ignoreZoom` intermediates keep recomputing correctly.
- `cssTransformProperties()` has a single caller and no subclass overrides, so making it state-dependent is safe.

**Minor note (non-blocking):** the saved transform write is rarer than "every card move/drop" implies. Typical drag/drop deltas include `x`/`y` alongside `parent`, and the `applyCSS()` loop returns on the first match — so `'x'` already triggered the write before `'parent'` was ever reached. The actual savings apply only to parent-only deltas (e.g. a routine setting just `parent`, or `moveToHolder` paths that don't touch coordinates). Still a valid micro-optimization, just narrower in practice.

`npm test` passes locally (6/6). No further concerns.

96LawDawg:
Opene by mistake with new AI Agent. Now closed and incorporated into #2988

PR #3013, opened 2026-07-10
Rework the Players tab into a session table with rename, view and local players for hotseat play

VTT-AI-Agent:
## What

Reworks the **Players** overlay from a flat list into a table that makes the player ↔ session relationship visible and adds the tools needed for hotseat play:

- **One row per connected session.** Per-player cells (color swatch, name) are merged across rows with `rowspan` when a player has multiple sessions. Your own session's row is highlighted, and players without a session are shown greyed out as "not connected".
- **Rename** (✏️ button / editing the name field): renames the player — all of their sessions follow — **and updates all references to the name in the game state** (`owner` strings/arrays, `player` of seats, `artist` of canvases) in one atomic delta broadcast by the server.
- **View** (👁 button on another player): renames *only your session* to that player's name without touching the game — you now see the game from their perspective (their hands, seats, owned widgets). To prevent undetectable hand-peeking, the server refuses taking the name of a player who is **connected and referenced in the game** (seat `player`, `owner`, `artist`), and the button is hidden for such players — local hotseat players and disconnected players stay viewable.
- **Split** (per-session button): renames a single session to a new name while the other sessions keep the old name; the split-off name gets a fresh color.
- **Add player** section: alongside the existing *Share room URL* button (remote players), a new *Add local player* input creates a player entry that has no session attached, plus a 🗑 button to remove session-less players again.

### Hotseat flow this enables

Add local players for everyone at the table, then: take a seat → *Players* → **View** the next player → take their seat → … and the same *View* cycle during play to act as each player in turn. *View*/*Split* never modify widget state, so this is safe in every game.

## How

- `server/player.mjs`: each socket gets a `sessionID`; the `rename` message accepts optional `updateWidgets` and `sessionID` arguments (both default to the previous behavior); new `addLocalPlayer`/`removeLocalPlayer` messages.
- `server/room.mjs`: `sendMetaUpdate` additionally broadcasts the session list; `renamePlayer` can rename a single session (keeping the old player entry for the remaining sessions) and rewrite `owner`/`player`/`artist` references via a normal delta; local players are plain `_meta.players` entries.
- `client/js/overlays/players.js` + `room.html` + `players.css`: table rendering with merged cells, the new actions, and the Add-player section.
- `tests/testcafe/test-util.js`: `setName` selector updated for the new structure (`.myPlayerEntry > .playerName` → `.myPlayerEntry .playerName`); the helper's flow is otherwise unchanged.

## Compatibility notes

- The `rename` message without the new arguments behaves exactly as before, so routine-driven renames (`var playerName = …`) are unaffected.
- No state format change: session IDs are ephemeral (per connection, sent in the meta message only), local players are ordinary `_meta.players` entries, so no FileUpdater migration is needed.
- Reference updates on rename only rewrite `owner`/`player`/`artist`. A survey of the public library (268/342 games use seats) shows routine writes of player names go overwhelmingly to `owner` (151) and `player` (48), which are covered; a handful of games snapshot names into custom properties (e.g. Going Nuts `playerInSeat`, Ninety-Nine `playerHolder`) or use names as object keys (Yoink, Rackball) — those already broke on the pre-existing rename feature and behave no worse now. Server-broadcast deltas don't execute `playerChangeRoutine`, so renames can't accidentally trigger sit/leave game logic.
- State hashes in the TestCafe suite are unaffected: renaming a fresh guest updates no widgets, and the meta message extension isn't part of room state. Verified locally with jest, the TestCafe editor test and a scripted multi-session browser test of all new actions (rename with references, view, split, add/remove local player).

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3013/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

Solid, well-scoped feature: the session model is ephemeral (no state-format change, no FileUpdater needed), the `rename` message stays backward compatible for routine-driven renames (`sessionID == null` also catches `undefined`), the widget rewrite goes through the normal delta/conflict machinery, and CI is fully green. The findings below are all edge cases — the most practical one is that player names are never trimmed, so a trailing space (common with mobile autocomplete) silently creates a second, visually identical player and rewrites all `owner` references to the spaced name.

## Findings

1. **Names aren't trimmed anywhere on the rename paths.** `renamePlayer` (`server/room.mjs:902`) and `addLocalPlayer` (`server/room.mjs:31`) use `newName.trim()` only as a validity check but store the *untrimmed* string, and the client sends the raw input value (`client/js/overlays/players.js:70-71` checks `e.target.value.trim()` but sends `e.target.value`; same for the split prompt at `players.js:108-109`). Renaming `Bob` to `Bob ` passes both guards, deletes `Bob`, creates `Bob `, and with `updateWidgets: true` rewrites every `owner`/`player`/`artist` to the spaced name — two visually identical players if someone else joins as `Bob`. A `newName = newName.trim()` at the top of `renamePlayer` (plus the same in `addLocalPlayer`) fixes all senders at once.

2. **`removeLocalPlayer` skips the widget-reference check that guest cleanup has.** `removePlayer` only deletes a disconnected guest when nothing references them (`server/room.mjs:864-866`), but `removeLocalPlayer` (`server/room.mjs:852`) deletes any session-less player unconditionally. In the hotseat flow this is easy to hit: View away from a seated player, then 🗑 them — their seat/hand `owner` references remain in the game, but the name vanishes from the table so you can no longer View them back (you'd have to know to re-add the exact name). Consider mirroring the reference check from `removePlayer`, or at least keeping the row visible while references exist.

3. **Merging players can create duplicate `owner` array entries.** In `renamePlayerInWidgets` (`server/room.mjs:934`), renaming `Alice` → `Bob` when a widget has `owner: ['Alice', 'Bob']` yields `['Bob', 'Bob']`. Probably cosmetic, but a dedupe after the `map` is cheap: `[...new Set(value.map(...))]`.

4. **Stale `sessionID` leaves a ghost player.** If the target session disconnects while the split `prompt()` is open (or any race), `renamePlayer` still creates `_meta.players[newName]` with a fresh color (`server/room.mjs:906-907`) even though the session loop at line 909-911 matches nothing — an unattached "not connected" player appears out of nowhere. Checking that at least one session matched before creating the entry would avoid it. Harmless as-is (the 🗑 button removes it), so judgment call.

5. **(Informational) The rename delta enters the client undo protocol.** `receiveDeltaFromServer` records it like any delta (`client/js/serverstate.js:332`), so undoing past a rename restores old-name `owner`/`player` references while `_meta.players` and the sessions keep the new name — widgets then belong to a nonexistent player. Undo never modeled renames, so this may be acceptable; just flagging that `updateWidgets` makes renames undoable-in-half for the first time.

6. **(Informational) Per-session identity vs. shared `localStorage`.** Two tabs of the same browser are two sessions, but they share one `localStorage.playerName` (`client/js/overlays/players.js:184`). Split tab B off as `Carol` and tab B writes `Carol` to localStorage — reloading tab A (still `Alice` in memory) silently turns it into `Carol` too. Inherent to the existing design and fine for the described single-tab hotseat cycle, but worth a mention in the help text or wiki since Split explicitly targets multi-tab setups.

Nice touches: the `sessionID` message replaying the last meta args avoids the obvious startup race, and the accidental-but-effective truthiness guard in `addLocalPlayer` even blocks `__proto__`-style keys.

VTT-AI-Agent:
Thanks for the review — addressed in 90ced96d:

1. **Trimming**: `renamePlayer` and `addLocalPlayer` now trim the incoming name before all checks and store the trimmed value, and the client sends trimmed values from the name field and the split prompt too. Renaming `Bob` → `Bob ` is now a no-op, and `Alice` → `Bob  ` merges into the existing `Bob`.
2. **`removeLocalPlayer` guard**: the server now refuses to delete a player who is still referenced by widgets (same check as the guest cleanup in `removePlayer`), and the client hides the 🗑 button for referenced players — so a seated hotseat player stays in the table and can always be Viewed back.
3. **Owner dedupe**: merging renames now pass owner arrays through `new Set`, so `['Alice', 'Bob']` → `['Bob']`.
4. **Ghost player from stale sessionID**: a session-scoped rename now returns early when no session matches, so no unattached player entry is created.

On the informational points: agreed on both. The undo interplay (5) predates renames being state-affecting only in the sense that undo never modeled `_meta.players` at all; undoing past a rename now restores old-name references, which is the same category of inconsistency as undoing past a seat click after the player left — I'd rather keep the rename delta undoable than special-case it, but happy to exclude it from the undo protocol if you prefer. The shared-localStorage behavior across tabs (6) is pre-existing (any rename in one tab already persisted for the whole browser); the intended flows (single-tab View cycle, or Split for genuinely separate windows where each tab keeps its in-memory name until reload) work as described — I'll add a note to the wiki's Playing-Games page once this lands.

Verified with jest, the TestCafe editor test and a scripted browser test covering all four fixes (trailing-space rename, merge dedupe, refused removal of a referenced player, stale-sessionID split).

VTT-AI-Agent:
4a3a87e5 adds immediate "working on it" feedback for all Players-tab actions, which previously only showed their effect after the server roundtrip:

- **View / Split / Remove buttons** switch to a spinning hourglass and disable themselves the moment they're clicked, until the resulting meta update rebuilds the table (with a 3s timeout that restores the button if the server silently refused the request).
- **Renaming via the name field** dims and locks the field while the rename is in flight; if no update arrives, the old name is restored.
- **Add local player** now uses the existing `progressButton` treatment (Working… → Done), including instant client-side errors for an empty or already existing name.

Verified under `simulateServerLag: 1500` that every state renders as intended; jest and the TestCafe editor test still pass.

VTT-AI-Agent (commented):
## TL;DR

The current head is in good shape: all four findings from the previous review round are verified fixed in 90ced96d (names trimmed on every path, `removeLocalPlayer` guarded by the widget-reference check, owner arrays deduped on merge, no ghost player from a stale `sessionID`), CI is fully green including the whole TestCafe matrix, and jest passes locally. What remains are three minor points — the most interesting one is that **"View" makes Guest entries immortal**, sidestepping the existing auto-cleanup.

## Findings

1. **Guest entries linger forever after "View".** `removePlayer` (`server/room.mjs:864-866`) auto-deletes `^Guest` entries on disconnect, but a spectator who clicks 👁 gets their *session* renamed away (`server/room.mjs:914-921` keeps `_meta.players[oldName]` by design), so when their browser later disconnects, the cleanup runs against the *new* name and the abandoned `Guest 123` entry is never removed. Named local players staying around is the point of the feature, but for guests it means every spectator who ever Viewed someone leaves a permanent "not connected" row until a human clicks 🗑. Suggestion: in the session-scoped branch of `renamePlayer`, apply the same cleanup as `removePlayer` — if `oldName` matches `^Guest`, has no remaining sessions and no widget references, delete the entry.

2. **`nextMetaUpdate` resolves on *any* meta broadcast, not the one caused by the request** (`client/js/overlays/players.js:44-56`, resolved in the `meta` handler at line 204). Meta updates fire on every join, leave and color change by anyone, so in a busy room a silently-refused request (e.g. a rename race where `oldName` was just merged away) still shows success within moments. Every meta update also redraws the table from server truth, so the UI self-heals — this only weakens the error feedback (`e.target.value = player` restore, the "Done" state on Add), it never shows wrong data. If you want it precise, verify the expectation against `lastMetaArgs` after the promise resolves (e.g. `meta.players[newName] !== undefined`) instead of trusting the resolve.

3. **`artist` is rewritten on rename but ignored by the removal guards.** `renamePlayerInWidgets` covers `owner`/`player`/`artist` (`server/room.mjs:936`), but `removeLocalPlayer` (`server/room.mjs:853-855`), the pre-existing guest cleanup, and the client's 🗑-visibility check (`client/js/overlays/players.js:140`) only test `player`/`owner`. A local player referenced only as a canvas `artist` can therefore be deleted while the reference stays behind. Harmless (a stale `artist` name just restricts drawing), and it matches the existing `removePlayer` convention — flagging it mainly so the asymmetry is a decision rather than an accident.

Also noting for the record: the widget rewrite broadcasts through the normal per-player delta bookkeeping (`possiblyConflictingDeltas` / `latestDeltaIDbyDifferentPlayer` via `Player.send`), and received deltas apply via `applyDelta`, which never fires `*ChangeRoutine` — so the PR's claim that renames can't trigger seat routines checks out.

---

I added the `beta` label: the PR touches no engine files, all CI is green, and View/Split/local players can't corrupt game state (widget rewrites only happen on an explicit rename and go through the normal delta machinery), so it's safe and useful for players to test.

VTT-AI-Agent:
Thanks — all three points addressed in e5640392:

1. **Immortal guest entries after View**: a session-scoped rename away from a `^Guest` name now deletes the abandoned entry when no sessions and no widget references remain — the same conditions as the disconnect cleanup, which indeed could never catch it under the new name. Named players keep their entry as before (that's the hotseat feature).
2. **`nextMetaUpdate` precision**: it now takes an `isApplied(args)` predicate and only settles the pending state on a meta update that actually reflects the request — rename/add check `meta.players`, View/Split check the session list, remove checks the entry is gone. Unrelated joins/color changes in a busy room keep the spinner alive until the real result or the timeout.
3. **`artist` asymmetry**: made it a decision — the new removal guards (server `removeLocalPlayer` and the client's 🗑 visibility) now share a `playerIsReferencedInWidgets` helper covering `owner`/`player`/`artist` (string or array). I deliberately left the pre-existing guest disconnect cleanup in `removePlayer` untouched to keep the diff focused; happy to switch it to the same helper as a follow-up if you'd like the legacy path aligned too.

Verified with a scripted two-browser test (guest Views another player → guest entry vanishes; artist-referenced player can't be removed and hides the 🗑 button; refused requests keep pending state until timeout), plus jest and the TestCafe editor test.

ArnoldSmith86:
image

Problems with the layout:
- the top is wider than the rest
- if only one session, the session part should be the same height as the player row
- the line under the table headers makes a jump at sessions

VTT-AI-Agent:
Good catch — all three symptoms had one root cause: the session cell had `display: flex` set directly on the `<td>` (added for right-aligning the split buttons), which takes the cell out of table layout. The cell then no longer stretched to the row height, its top border sat at a different y than the other cells (the \"jump\" under the headers), and its column width stopped lining up with the header row.

Fixed in 53bf9355 by moving the flex layout onto a wrapper `<div>` inside the cell, so the `<td>` participates in normal table layout again. Verified the geometry programmatically (all body cells now share the exact row height and the same column edges as the header cells) and visually for both the single-session and merged multi-session cases.

VTT-AI-Agent (commented):
# Review of PR #3013 (head 53bf9355)

## TL;DR

The current head looks ready: all findings from both previous review rounds are verified fixed in the code (guest cleanup on View, predicate-based `nextMetaUpdate`, shared `playerIsReferencedInWidgets` including `artist`, trimming, removal guards, owner dedupe), and the maintainer's three layout symptoms were fixed at their single root cause — `display: flex` on the session `<td>` took it out of table layout; it now lives on the `.sessionCellContent` wrapper. I found nothing new worth raising.

## Verification details

- **e5640392 (review round 2 fixes)**: the session-scoped branch of `renamePlayer` (`server/room.mjs:920-924`) now deletes an abandoned `^Guest` entry under exactly the same conditions as the disconnect cleanup, and it runs *after* the sessions were renamed, so the remaining-session count is correct. `nextMetaUpdate` (`client/js/overlays/players.js:45`) settles only on a meta update matching each action's predicate (rename/add via `meta.players`, View/Split via the session list, remove via entry absence). Server `removeLocalPlayer` and the client's 🗑-visibility check both cover `owner`/`player`/`artist`, string or array.
- **53bf9355 (layout)**: the `<td>` participates in table layout again; the flex right-alignment of the split button moved to an inner `div`, which addresses all three reported symptoms (header width mismatch, single-session row height, border jump).
- **Delta protocol**: `renamePlayerInWidgets` broadcasts through `Player.send`, so the per-player conflict bookkeeping (`possiblyConflictingDeltas`, `latestDeltaIDbyDifferentPlayer`) stays consistent, and clients resync their `deltaID` from the broadcast — no protocol gap.
- **Tests**: jest passes locally (6/6); CI is fully green including the complete TestCafe matrix, which exercises the new table through `setName` on every test.

The `beta` label was already present and stays: the PR touches no engine files, CI is green, and View/Split/local players only affect `_meta.players` and explicit rename deltas, so it remains safe and useful for players to test.

96LawDawg:
One aspect of this PR is troublesome. It will be very easy (and non-detectable) to cheat by looking at the cards in the hand of a player by going into the player overlay and viewing the game as that player. One solution is to disable that switching capability if the player is connected (so local hot-swapping is still allowed) and is in a seat. There are a few older games that may not use seats and just use the `owner` property. Not sure how to handle those. I'm open to ideas. But this should not merge until that is resolved.

VTT-AI-Agent:
@96LawDawg Agreed — implemented in 65765552, with one generalization of your suggestion that also answers the owner-only question:

**Rule: a rename onto a name that (a) has connected sessions AND (b) is referenced anywhere in the game state (`player` of seats, `owner` strings/arrays, `artist`) is refused by the server, and the client hides the 👁 button for such players.**

- Using widget references instead of only seats covers the older games you mentioned that skip seats and use `owner` directly — the same check the removal guard already used (`playerIsReferencedInWidgets`).
- Hotseat is unaffected: local players have no session, so they stay viewable even while seated. Disconnected players stay viewable too, which keeps "resume as the player who dropped out" working.
- Connected spectators without any game references remain viewable — there's nothing to reveal there.
- The guard applies to *all* renames onto an existing name (View, Split, and full-rename merges), since merging yourself into a seated player was the same cheat through a different door.

Worth being transparent about the remaining surface: the room protocol is unauthenticated, so someone can still join *as* "Bob" via the URL hash (`#player:Bob`) or by editing localStorage — that predates this PR and can't be closed without breaking legitimate reconnection (rejoining under your own name after a disconnect looks identical to the server). This change removes the one-click undetectable path the PR would have added, and restores the status quo where impersonation requires deliberately stepping outside the UI.

Verified end-to-end: seated/owner-referenced connected players show no 👁 and forced websocket rename attempts (both session-scoped and merge) are refused, while View on a seated *local* player still works for hotseat.

96LawDawg:
An eye icon still shows up for connected players in seats. When clicked, the icon shows a spinning hourglass for a few seconds and then nothing happens. But the icon should not be there.

VTT-AI-Agent:
@96LawDawg Good catch, fixed in c692de51. The server-side refusal was working (that's the hourglass timing out with nothing happening), but the button visibility was computed from stale data: the table only rebuilt on meta updates, and on connect the meta message arrives *before* the room state — so the list was built against an empty widget map and didn't know who was seated. The list now also refreshes on state messages, on deltas that touch `player`/`owner`/`artist`, and when the Players toolbar button is clicked.

Verified against your exact repro (fresh page load into a room with a seated connected player → open the overlay → no 👁 on them) — and it now updates live: standing up makes the 👁 appear, sitting down removes it, even while the overlay is open.

VTT-AI-Agent (commented):
## TL;DR

**UI/UX review only** (a separate reviewer covers code correctness). This is a strong, well-organized rework — the session-per-row table with merged player cells reads clearly and the instructional text does a good job explaining the new session concept. My single most impactful suggestion: **the three icon-only action buttons float in a large empty gap in the middle of the wide "Player" column** — right-aligning that cell (so the actions hug the Sessions column) would make each row read as one connected unit instead of a detached cluster. Beyond that, the main risk for first-time users is that the *View* (👁) and *Split* (⑂ `call_split`) actions are icon-only and their meaning lives entirely in hover tooltips.

I reviewed this by running the app locally (port 8317) and rendering the reworked Players overlay with two live sessions plus two local players.

Players table with a merged multi-session player

---

## Big layout improvements

**1. Action buttons float in the middle of the Player column.**
The `.playerName` input is `width: 100%`, so it stretches across the wide Player column while the actions cell (`edit` / `view` / `delete`) sits in its own column after it. The result is a big empty gap between the name and the buttons, and another gap before Sessions — the icons read as detached rather than belonging to the row. Two live-server captures showing it:

Default table, actions floating mid-row
Small-landscape view, gap even more visible

Suggestion: right-align `.playerActionsCell` (`text-align: right` / `justify-content: flex-end`) so the actions cluster hugs the Sessions boundary, or cap the name input width so the buttons sit right after the name. Either makes each row scan as a single unit.

**2. The split control is pushed hard to the far edge of the Sessions cell.**
`#playersTable .sessionCellContent button { margin-left: auto; }` shoves the ⑂ button to the far right, leaving a wide gap between e.g. "Session 1 (you)" and its split button (visible in the merged screenshot). In a column this narrow the split action ends up visually disconnected from the session it acts on. Consider a small fixed gap (`margin-left: 0.5rem`) instead of `auto`, so the button stays next to its label.

**3. Consider suppressing the "Session 1" label when a player has exactly one session.**
For every single-session player the cell reads "Session 1" / "Session 1 (you)". The numbering only carries meaning once a player actually has multiple sessions (as with the merged row). For the common single-session case, "Session 1" is noise — you could show just "(you)" / "connected" for one session and switch to "Session 1 / Session 2 …" only when the count is > 1. This declutters the most common rows.

## Nitpicks

- **Two empty `<th>` header cells.** The header row is `<th></th><th>Player</th><th></th><th>Sessions</th>` (room.html:114-ish). The empty color and actions headers are fine functionally, but combined with #1 the "Player" header sits far left of where the name+actions actually cluster. Fixing the actions alignment also improves how the header lines up with its column contents.
- **Icon button hit area is small.** `#playersTable button[icon]` uses `padding: 3px` with an 18px glyph (~24px total). That's on the tight side for a touch target next to three other small buttons; a couple more px of padding (or gap) would reduce mis-taps, especially on the merged rows where edit/view/delete sit close together.
- **`border-top: 1px solid #00000044` is hard-coded black-with-alpha.** It looks right on this light overlay, but the rest of the players styling references `var(--VTTblue)` / theme tokens. If this overlay is ever themed/darkened the divider will disappear; a token or `currentColor`-based rule would be safer. Minor.
- **Own-session tint is very subtle.** `#1f5ca622` (~13% alpha blue) reads as barely-there highlighting on the "you" row. It works, but bumping the alpha slightly would make "which row is me" register faster at a glance.

- **The power of this panel lives in tooltips only.** The real conceptual difference between the actions is significant — *Rename* (✏️) rewrites the name everywhere in the game, *View* (👁) re-points only your session, *Split* (⑂) forks one session into a new player — but all three are icon-only buttons whose explanation is in the `title` attribute. A newcomer hovering may still not grasp "view the game as this player" vs "split this session off." The `call_split` (⑂) glyph in particular is not self-evidently "make this session its own player." Consider (a) at minimum keeping those tooltips (good that they exist), and ideally (b) a short one-line legend under the table, or text labels on hover, since tooltips never appear on touch devices.
- **Icons carry meaning that a label would make instantly clear.** 👁 = "view as", ⑂ = "split", 🗑 = "remove". The delete icon is conventional enough; the eye and split are the two a first-timer will hesitate over. Even tiny captions ("View" / "Split") would remove the guesswork — this app otherwise labels most primary actions in text (Game Shelf, Active Game, Share room URL, Add local player).
- **Instructional copy is good.** "Every connected browser tab is a session of a player…" and the "Add player" paragraph explaining local players + the 👁 pass-the-turn flow are clear and appropriately scoped — nice touch inlining the 👁 glyph in the sentence so the icon is grounded.
- **"not connected" is clear** and the greyed styling reads correctly as an inactive/local player. Good empty-state handling for local players.
- One small thing: the old "There are no other players at this table." empty-state message was dropped in the rework. In a brand-new room the table now just shows a single self-row with no explanation that you're alone. Not a blocker, but a one-line hint ("You're the only one here — share the room URL or add a local player below.") would help a first-time solo user understand what to do next.

---

Overall the new table is a clear improvement over the flat list and the visual style matches the app. Addressing the action-button alignment (#1/#2) and giving the View/Split icons a label or legend would take it from good to very polished.

96LawDawg:
@VTT-AI-Agent, implement the fixes suggested in the UI/UX review at https://github.com/ArnoldSmith86/virtualtabletop/pull/3013#pullrequestreview-4724364517.

Also, investigate why the eye icon appears differently for different players as in this screenshot. The far left is (Left) in Chrome. The middle is Mike (in Edge). The far right is Incognito (Edge) and Mike Phone was an iPhone
Screenshot

VTT-AI-Agent:
Implemented the UI/UX review suggestions and looked into the eye icon.

## Why the 👁 button looks different in your screenshot

That is not a browser difference — the four windows actually agree with each other. The button is hidden for a player that is **connected *and* referenced in the game** (a seat's `player`, a widget `owner`, a canvas `artist`), because taking their name would let you look at their hand without anyone noticing (the server refuses it too). Going through your screenshot:

| player | Left's window | Mike's window | Incognito's window |
| --- | --- | --- | --- |
| Left | (self) | 👁 | 👁 |
| Incognito | 👁 | 👁 | (self) |
| Mike | – | (self) | – |
| Mike Phone | – | – | – |
| Guest152 / No device 2 | 👁 | 👁 | 👁 |

So Mike and Mike Phone were sitting in the game (no 👁 anywhere), Left and Incognito were connected but not part of it yet (👁 everywhere except in their own window), and the two not-connected players can always be viewed. Every window shows the same rule — it just looked arbitrary because the button silently disappeared with no explanation.

Fixed by keeping the button in place, greyed out, with a tooltip saying why, plus a legend line that also works on touch devices where tooltips never appear.

## Review suggestions implemented

- **Actions no longer float in the middle of the row**: the name column now absorbs the free width (`.playerNameCell { width: 100% }`), so the action buttons sit right next to the Sessions column and the whole row reads as one unit.
- **Split button no longer pushed to the far edge**: the session column is now only as wide as its widest label, so `margin-left: auto` keeps the ⑂ buttons aligned with a small gap instead of a wide one.
- **"Session 1" suppressed for single-session players**: a player with one session now shows `connected` / `connected (you)`, numbering only appears from two sessions on.
- **Icon legend under the table** explaining ✏️ / 👁 / ⑂ / 🗑 in words, since the meaning used to live in tooltips only.
- **Empty state**: a hint appears when you are the only player at the table, pointing at the room URL and the local-player input.
- Nitpicks: bigger icon hit area (3px → 5px padding), row separator uses `var(--VTTblueLight)` instead of hard-coded `#00000044`, own-row tint bumped from ~13% to ~20% alpha.

Verified locally with jest (15/15), the TestCafe editor test (which drives the players overlay through `setName`), and a scripted five-session browser check of the overlay:

Players tab after the changes

Small landscape viewport (640×400):

Players tab on a small viewport

96LawDawg (approved):
Tested. Approved. This is worth it just to have "owner" cards renamed if a person renames themself during play. Plus all the other great stuff including hot seat. What's not to love about this?

VTT-AI-Agent:
Feedback received outside GitHub:

> In the improved player session table, the heading is wider than the rest of the content. Please make sure the heading width matches the rest.

Fixed in 147a98d5: the heading block (the *Players* title plus the instructional text) stretched across the full overlay width, while the session table, the legend and the *Add player* block are all capped at `40rem` and centered — so the heading stuck out on both sides. The heading now uses the same `max-width: 40rem; margin: 0 auto`, and its left/right edges line up with the table (measured 502–1142px for all four blocks at a 1600px viewport, identical before/after for narrow viewports where everything is full width).

PR #3014, opened 2026-07-10
Editor: property UI for all widget types with inline color/icon/image pickers

VTT-AI-Agent:
Completed / Notes:

- [x] The movable selection buttons need another option (movable by player but not in edit). Maybe make it just 2 buttons and they can be on or off (and remove the not movable one).
- [x] In the tweak position area, add a lock to width and height so they optionally change together.
- [x] For dice, allow reordering of faces by drag and drop (or at least by up/down arrows)
- [x] For basic widgets with faces, add something similar to the dice thing where each face's properties are listed in a draggable area
- [x] Piles also have a Properties area, but that is probably misleading as that goes away as soon as the pile does. Don't have that area at all and just mention that piles customization is done in the deck
- [x] When picking colors, there needs to be a way to choose transparent
- [x] Should be able to change parent by typing widget id or by clicking a widget (like in the pick appearance button). When changing parent, have the editor calculate the x/y so the widget's position in the room does not shift
- [x] Probably most entries should have a hoverable ? with hint text
- [ ] Maybe more entries should get hoverable ? hint text? (human decision - AI leave this here).
- [ ] Maybe add pile customization options to decks, but that may be too detailed. (human decision - AI leave this here).

## What

This builds out the **Properties** sidebar module of the editor with curated, type-specific controls instead of generic `not set/text/number/object` inputs. Transient piles are the deliberate exception: their panel points creators to the owning deck instead of editing temporary pile state.

### Shared input framework (`client/js/editor/propertyInputs.js`)

A small set of input classes that all sync **both ways**: they write through `inputValueUpdated` (normal delta batching, undoable) and update themselves from incoming deltas via the module's property listeners — so edits from other players or the JSON editor show up live, and focused inputs are never clobbered while typing.

- `TextInput`, `NumberInput` (with optional slider + unit), `CheckboxInput`, `SelectInput` (keeps custom values as an extra option)
- `ColorInput`, `IconInput`, `ImageInput`: show the current value as a color swatch / rendered icon / thumbnail. Clicking expands an **inline section (no popup)** with:
  - a summary of the current value plus a clear button,
  - the other **colors/icons/assets used in the rest of the game** (click to apply),
  - a color palette + native picker for colors,
  - an **upload button** and URL input for images,
  - an inline **icon search** over `symbols.json` that shows only the top results, with **frequently used icons** on top when the search is empty — seeded from a scan of the public library (most-used game-icons across `library/`), merged with the icons already used in the current game.
- CSS-declaration helpers so the same inputs can edit single declarations inside the `css` property (used for the label typography toolbar and timer colors) while preserving the rest of the property, including the nested-class object form.

### Per-type editors

Persistent widget types render a header plus the shared position/size controls and relevant type-specific sections, ending with the generic list for anything not covered:

- **basic**: content/appearance controls plus editable, sortable face cards
- **button**: text, icon, image, all base/hover colors, border radius, clickRoutine hint
- **canvas**: drawing palette (colorMap entries as color inputs), resolution/line width/active color
- **card / deck**: kept the existing card-type/face-template editors, plus the new basic section
- **dice**: inline sortable face editor with pips/text/icon/image modes, per-face color locks and presets; 3D/timing controls
- **holder**: kept drop-target + preset buttons; colors, border radius, drop shadow, align/stack/drop offsets, prevent piles, children per owner
- **label**: text area, editable, placeholder, typography toolbar (bold/italic/align/font size/color written into `css`)
- **pile**: informational-only guidance and a shortcut to the cards’ deck when available
- **scoreboard**: appearance flags + all scoring options
- **seat**: colors, index, display property, hand, turn flags
- **spinner**: kept option presets; text/line colors, current value
- **timer**: background/text color (via `css`), countdown/paused/alert, precision, start/end/current

### Appearance sections

Every *Appearance* header has two tools, as requested:
- a **pipette** button: click it, then click any widget on the board — its appearance properties are copied onto the selected widget (properties the source doesn't declare reset to defaults),
- an **apply to all** button that applies the widget's appearance to all widgets of the same type.

Both are batched with a descriptive delta cause, so they're single undo steps.

## Why

Editing widgets currently requires knowing property names and JSON. Typed, categorized inputs with visual pickers make the editor approachable, and consistent appearance across widgets (pipette / apply-to-all) is a common pain point when styling a game.

## Testing

- jest suite passes.
- Verified in a real browser (Playwright, headless Chromium) against a room containing every widget type: all editors render without console errors; checked two-way sync (slider→state, remote `set()`→input), color set via picker and via used-in-game chips, icon search + pick from top results, pipette copy between buttons, apply-to-all (only same-type widgets affected), label bold/font-size written into `css`, checkbox sync.
- TestCafe: the editor suite doesn't interact with the Properties module; it fails identically on unmodified `main` in my local environment (browser-version drift), so relying on CI here.

No widget/engine behavior changes — the diff is limited to the editor sidebar (plus registering the two new files in `server/minify.mjs`).

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3014/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

Solid, well-structured feature: the shared input classes are clean, property names all check out against the widget defaults, escaping via `html()` is used consistently, and jest passes locally. The one bug I'd fix before merge is that an armed pipette pick survives closing the Properties module — it leaves the crosshair cursor stuck and can fire a surprise appearance-copy when the module is reopened. A couple of smaller correctness/UX issues below.

## Findings

### 1. Armed widget pick leaks across module close/reopen (main issue)

`startWidgetPick()` (`client/js/editor/sidebar/properties.js:276`) sets `this.widgetPick` and adds `editorWidgetPicking` to `body`, but `PropertiesModule` never overrides `onClose()`, so nothing clears this state when the module (or the whole editor) is closed while a pick is armed. Consequences:

- The `cursor: crosshair !important` on `body`/`.widget` (`propertyInputs.css:228`) persists after closing the module — even after leaving edit mode entirely.
- Worse: `openInTarget()` calls `onSelectionChanged(selectedWidgets, [])` when the module is reopened (`sidebarModule.js:151`). The stale `widgetPick` branch at `properties.js:433` then runs immediately — if the user meanwhile selected a different widget, its appearance is copied onto the old target without any click on the pipette. It's undoable, but very surprising.

Fix is small: override `onClose()` to call `this.stopWidgetPick()`.

### 2. Picking an icon from search results resets the search

Local edits also flow through `receiveDelta → editorReceiveDelta` (`serverstate.js:597`), so every `setValue()` from inside an open picker triggers `update() → updatePicker()`, which does `pickerDOM.innerHTML = ''` and rebuilds everything (`propertyInputs.js:393`). Clicking an icon chip in the search results therefore destroys the search input — query and focus are lost and the list snaps back to "frequently used", which makes trying a few candidates from one search painful. The same rebuild also re-walks the entire game state (`usedGameColors/Icons/Images`) on every delta for a listened property while a picker is open, and recreates the native `<input type=color>` mid-drag. Consider updating chip selection in place (or skipping the rebuild when the delta came from this input) instead of a full re-render.

### 3. `css` string values containing `;` inside a value get corrupted

`widgetCssObject()` parses the string form with `css.split(';')` (`propertyInputs.js:555`). A declaration like `background-image: url("data:image/svg+xml;base64,...")` — which real games use — gets split mid-value, and the next edit through any typography/timer-color control writes the mangled result back via `setWidgetCssValue()`. Also note any such edit silently converts a string `css` to the object form (losing declaration order/duplicates); that's probably acceptable, but the data-URI corruption is a real, if narrow, data-loss case. A cheap guard: if parse→serialize doesn't round-trip the original string, leave the property alone (fall back to the generic input).

### Minor

- `usedGameIcons()` (`propertyInputs.js:62`) treats any `name` key whose value matches `^[a-z][a-z0-9_]+$` as an icon, so e.g. a cardType `"name": "copper"` shows up as a broken material-symbols chip in "Used in this game". The `name` key seems much noisier than `icon`/`suit`.
- The icon search index includes monochrome emoji that the existing symbol picker deliberately filters out via `skipForNotoMonochrome()` (`symbols.js:60`) because Noto Emoji can't render them — picking one from the new search gives a broken glyph.
- `SelectInput.update()` (`propertyInputs.js:336`) appends a new `custom: …` option every time an unknown value arrives and never removes old ones, so options accumulate as a value toggles between custom states.
- `basicExcludeList()` now hides `z` from "Other properties" for every type without offering a replacement input; previously it was editable in the generic UI for non-card types. Probably intentional (it's engine-managed), but worth confirming.
- After a pipette pick, `setSelection([target])` runs from inside `setSelection`'s module-notification loop (`properties.js:441`), so the remaining modules get notified twice and re-render twice. Harmless, but deferring the call (e.g. `setTimeout(…, 0)`) would avoid it.
- `loadIconSearchIndex()` has no in-flight guard, so two pickers opened quickly both fetch and parse `symbols.json`.

### Test coverage

`widgetCssObject`/`setWidgetCssValue` and `searchIconIndex` are pure and easily unit-testable (the `;`-in-value case above would make a good regression test); currently nothing covers them, and TestCafe doesn't exercise the Properties module.

---

I added the `beta` label: the PR touches only the editor sidebar (no engine files), works in practice, and none of the issues above can corrupt games or crash the server, so it's safe and useful for players to test.

VTT-AI-Agent:
Thanks for the review — all findings addressed in bb53119:

**1. Armed pick leaking across module close/reopen** — fixed. `PropertiesModule` now overrides `onClose()` and `onEditorClose()` to call `stopWidgetPick()`, so the crosshair cursor and the armed pick are cleared whenever the module or the editor closes. Verified in a headless browser: arming the pipette, closing the module, reopening and selecting another widget no longer copies anything and the `editorWidgetPicking` body class is gone.

**2. Picking an icon resets the search** — fixed. `PickerInput.update()` now does an in-place refresh instead of rebuilding the picker: only the summary row is re-rendered (and even that is skipped while it contains the focused element, which also protects the native color input mid-interaction), and chip selection marks are toggled via `data-value`. The full rebuild (including the game-state scan) only happens when the picker is opened. Verified: searching “dragon”, picking a result keeps the query, the result list and focus, and the selection mark moves to the picked chip.

**3. `css.split(';')` corrupting data-URI values** — fixed with the round-trip guard you suggested: `cssStringIsEditable()` re-serializes the parsed declarations and compares against the normalized original; on mismatch `widgetCssObject()` treats the string as opaque and `setWidgetCssValue()` refuses to write. A label with `background-image: url(\"data:image/svg+xml;base64,…\")` in string css now passes through the bold toggle untouched (covered by a unit test, and verified in the browser). Duplicate-key strings are also detected and left alone.

**Minor items:**
- `usedGameIcons()` no longer matches `name` keys, and material-symbol patterns only match under `icon` (game-icons paths still match under `icon`/`suit`).
- The icon search index now applies the same `skipForNotoMonochrome()` filter as the symbol picker, and chips render monochrome emoji through `toNotoMonochrome()`. (Both helpers are now exposed to the editor module via the `loadEditMode` whitelist in `main.js` — they were module-scoped before, which this fix surfaced.)
- `SelectInput` reuses a single `custom: …` option and removes it when the value returns to a predefined choice.
- `z`: intentional — it's continuously rewritten by the engine on drag/drop and z-order is editable via the layer-order toolbar button, so a sidebar input for it seemed like a trap. It's still visible/editable in the JSON editor. Happy to add it back to “Other properties” if you disagree.
- Pipette reselect is now deferred (`Promise.resolve(...).then`) until the copy finished, so it no longer re-enters `setSelection`'s notification loop and the re-rendered module shows the copied values.
- `loadIconSearchIndex()` now caches the in-flight promise.

**Tests:** added `tests/client/property-inputs.test.js` covering `cssObjectFromString`/`cssStringIsEditable`/`widgetCssObject`/`get\|setWidgetCssValue` (including the data-URI regression case) and `searchIconIndex` — jest is now 14/14.

VTT-AI-Agent (commented):
## TL;DR

All five findings from the previous review are verified fixed in bb531193 — I confirmed each one live in a headless browser (pipette state cleanup, in-place picker refresh keeping the search query, data-URI `css` left untouched by the typography toolbar, Noto-monochrome filtering, `usedGameIcons` no longer matching `name`) and the new unit tests pass (jest 14/14). One small regression from the fix commit remains: after clicking the picker's "Remove value" button, the summary row keeps showing the old value.

## Verification of the previous round

- **Pipette leak**: armed the pipette, closed the Properties module — `editorWidgetPicking` is removed from `body`; reopening and selecting another widget copies nothing. The positive path also works: arming on a button and picking another copies the appearance properties in one batch and reselects the target afterwards.
- **Icon search survives picking**: searched "dragon", clicked a result — the query, result list and selection mark survive, and the widget's `icon` updates.
- **`css` string with a data URI**: toggling bold on a label whose `css` is `background-image: url("data:image/svg+xml;base64,…")` leaves the string byte-identical (also covered by the new `cssStringIsEditable` tests).
- **Whitelist fix**: `toNotoMonochrome`/`skipForNotoMonochrome` resolve correctly from the editor bundle (both dev and prod concatenate `symbols.js` and `main.js` into one scope); no console errors anywhere in my session.

## New findings

### 1. Stale picker summary after clicking "Remove value"

`refreshPicker()` (`client/js/editor/propertyInputs.js:418`) skips re-rendering the summary whenever it contains `document.activeElement`. In Chromium, clicking a button focuses it — and the clear button lives inside the summary. So clicking "Remove value" clears the property (preview button and chip marks update correctly) but the summary row keeps showing the old chip and value text until focus leaves it. Reproduced live: widget `color` reset to default while the summary still read `#ff0000`. A narrower skip condition would fix it — e.g. only skip when `document.activeElement` matches `input, textarea`, or blur the clear button before `setValue(null)`.

### 2. "Appearance" copying includes content for some types (design question)

For buttons (and basic widgets/piles), `text`, `icon` and `image` are part of the appearance section, so the pipette and especially **apply-to-all** overwrite them: applying one button's appearance to all buttons renames every button in the game to the same label (verified live — `text` was copied along with the colors). It's a single undo step, so it's recoverable, but users clicking a button labelled "apply this appearance to all" likely expect styling, not content. Consider excluding `text` (and possibly `icon`/`image`) from the *copied* property set even if they stay in the section UI.

### Minor

- `loadIconSearchIndex()` (`propertyInputs.js:118`) caches the in-flight promise, but if the `symbols.json` fetch fails, the rejected promise is cached forever — icon search is then dead until reload, and every keystroke in the search box produces an unhandled rejection (`search.oninput` awaits it without a catch). Clearing `iconSearchIndexPromise` on rejection would let it retry.

---

The `beta` label is already present and I left it there: the PR still only touches the editor sidebar (no engine files), everything I exercised works, and the remaining findings are minor UX issues that cannot corrupt games or crash the server.

VTT-AI-Agent:
Thanks — both findings plus the minor one addressed in 91ca4a99 (each verified live in a headless browser):

**1. Stale summary after “Remove value”** — fixed. `refreshPicker()` now only skips the summary rebuild while the focused element inside it matches `input, textarea` (typing / dragging the native color picker); a focused button no longer suppresses the refresh. Verified: clearing a set `backgroundColor` immediately flips the summary to *not set* and the property is removed from the state.

**2. Appearance copy overwriting content** — agreed. Appearance entries can now be flagged as `content`: they stay editable in the section but are excluded from the property set the pipette and apply-to-all copy. Flagged `text` for buttons, basic widgets and piles (the pile handle text). I kept `icon`/`image` in the copied set on purpose: for basic widgets (game pieces) they effectively *are* the appearance — “make all pawns look like this one” is the main use case for apply-to-all. Happy to flag those too if you'd rather err the other way. Verified: apply-to-all from a button copies `textColor` to all buttons while every button keeps its own `text`.

**Minor: cached rejected `symbols.json` promise** — fixed. The in-flight promise is dropped on rejection so the next open/keystroke retries, and both call sites catch the rejection (search falls back to the frequently-used list), so no unhandled rejections.

jest 14/14.

VTT-AI-Agent (commented):
## TL;DR

All three findings from the previous review are fixed in 91ca4a99 — I verified each one live in a headless browser against this branch, and jest passes (14/14). No new issues found; from my side this PR is ready for human review.

## Verification of the previous round

- **Stale summary after "Remove value"**: fixed. `refreshPicker()` now only skips the summary rebuild while the focused element matches `input, textarea` (`client/js/editor/propertyInputs.js:419`). Verified live: with the `textColor` picker open on a button, clicking "Remove value" removes the property from the widget state and the summary immediately re-renders from `#ff0000` to the computed default — no stale value.
- **Appearance copy no longer includes content**: fixed via the `content` flag (`client/js/editor/sidebar/properties.js:1589`). Verified live both ways: apply-to-all from a button copied `textColor` to the other button while each kept its own `text`; the pipette likewise copied `textColor` but not `text`, and the pick state (`editorWidgetPicking`) was cleaned up afterwards. I also checked that none of the explicit `properties:` overrides (canvas, label typography, holder) bypass the filter — none include `text`. Keeping `icon`/`image` in the copied set for basic widgets is a defensible design call ("make all pawns look like this one"); leaving that to the human reviewers.
- **Cached rejected `symbols.json` fetch**: fixed by inspection — the in-flight promise is dropped on rejection (`propertyInputs.js:147`), both call sites catch, and `searchIconIndex` iterates `iconSearchIndex || []`, so a failed fetch degrades to the frequently-used list and retries on the next keystroke.

No console errors appeared anywhere in the verification session. The diff still touches only the editor sidebar plus `server/minify.mjs` registration — no engine files.

---

The `beta` label was already present and I left it there: everything I exercised works, all review findings are resolved, and the change cannot corrupt games or crash the server.

ArnoldSmith86:
- for property `icon`, the picker should show all icons - fonts _and_ images
- make the icon search result previews ~50% bigger
- collapse x/y/width/height/rotation/scale into a collapsed "Tweak position, size and rotation" section (they are usually changed by using the drag toolbar)
- when no value is set, make the appearance pickers (icons/images/colors) show the applying default instead
- make a "Content" section between position stuff and appearance (currently named Label for labels for example) and move some things from appearance up there (image/text/icon); split the rest of appearance into subsections like Colors and Hover where applicable
- in image popouts, suggest i/game-pieces images like the icons in the icon picker

VTT-AI-Agent:
Thanks @ArnoldSmith86 — all six done in d460221f (verified live in a headless browser, jest still 14/14):

- **Icon picker shows fonts _and_ images** — the search index now includes game-icons (images), material/VTT symbols (fonts) and both the color-image and monochrome-font variants of emoji. Results are interleaved font/image so both kinds show up in the top matches even for a query like "star" where game-icons would otherwise fill the whole page.
- **~50% bigger icon previews** — icon result chips are now 50px (was 34px).
- **Collapsed position section** — x/y/width/height/rotation/**scale** are in a collapsed *"Tweak position, size and rotation"* section; layer + movable stay visible below it.
- **Pickers show the applying default when unset** — icon/image/color pickers now preview the widget's effective default (dimmed, labelled `default: …`) instead of an empty swatch; the clear button only appears once an explicit value is set.
- **Content section + appearance subsections** — a new *Content* section (text/icon/image) sits between position and appearance; *Appearance* is split into **Colors** / **Hover** / **Style** subsections where applicable. Content is never copied by the pipette / apply-to-all (that stays styling-only).
- **Game-piece image suggestions** — the image pop-out now suggests the built-in `i/game-pieces` images as clickable chips, alongside images already used in the game, above the upload / URL inputs.

You can try it on the PR preview once it redeploys. Screenshots of the button editor (collapsed geometry, Content, Colors/Hover/Style), the icon search (fonts+images, bigger) and the image picker (Game pieces row) are in my working notes — happy to attach if useful.

ArnoldSmith86:
- cards: move "cardType" property to "Card type" section and change it to a drop down
- change the movable checkboxes into a text slides / multi button: "movable by players > movable in editor > not movable"
- color controls seem to show a dimmed color when it is the default. don't render the default color differently than a manually set one.
- replace checkboxes by switches (reuse existing styling)
- font size control: if not set in the widget, show the default value; same for alignment
- color controls: make it possible to edit the hex value manually
- add layer and z to the tweak collapsible
- dice: replace dice type, dice shape, face type by an inline face editor where you can add/remove/edit faces, switching each between pips, text, icons or images and editing colors - each property gets a "locked" switch for making the value the same for all faces (corresponds to setting properties in the widget itself and not the faces array) - add a load preset button that opens a popout which has the usual pip and number dice as well as color dice and other common examples (each preset should render all its faces in the selection) - move a 3d toggle to the appearane section - some colors etc form current appearance need to be removed and integrated into the face editor section
- seats: add a dropdown for selecting seated player (updating player and color properties)
- add mobile friendly on-demand hints for all property controls describing what the property does and common gotchas from the wiki

ArnoldSmith86:
selecting multiple widgets should not show each widget's properties after another but instead show all properties that make sense for all selected widgets. with half disabled controls for properties that have different values across selected widgets that do support setting all to the same value but the UX needs to be clear. below the tweak position collapsible, show the arrange buttons from the main toolbar (duplicating them for now - but not duplicating the code)

VTT-AI-Agent:
Thanks @ArnoldSmith86 — all of these are in a92676eb (each verified in a headless browser, jest 14/14, no console errors on any widget type):

- **cardType → dropdown in a Card type section** — cards now show a *Card type* dropdown listing the deck's card types; it's out of the generic property list.
- **Movable → three-state control** — the two checkboxes are now one segmented control: *by players › in editor › not movable*.
- **Default colors not dimmed** — color previews/values render a default identically to a manually set one (only icon/image pickers still dim the unset-default hint).
- **Checkboxes → switches** — all boolean controls reuse the global `.switchbox` styling.
- **Font size / alignment show the default** — font-size shows the applying default as a placeholder; the left-align button shows as the active default when nothing is set.
- **Hex entry for colors** — the color pop-out has a `#rrggbb` text field next to the swatch/native picker.
- **layer + z in the collapsible** — both moved into the *Tweak position, size and rotation* section.
- **Dice inline face editor** — replaces the old type/shape/face-type presets. You can add / remove / edit each face, switching it between **pips / number / icon / image** and editing its value; background / pip / border colors each have a **"same for all faces" lock** (locked = the property on the dice widget, unlocked = per-face values in the `faces` array). A **Load preset** button opens a pop-out with coin, d4–d20 (pips and numbers) and a color d6 — each preview renders **all** of that preset's faces. The **3D** toggle moved to the appearance section, and the face colors are now edited here instead of in appearance.
- **On-demand hints** — every property control has a (?) that toggles a short, touch-friendly description drawn from the wiki summaries (added a hint map; more can be filled in easily).

The dice face editor with the preset pop-out is the biggest change — worth a look on the preview deploy once it rebuilds.

VTT-AI-Agent:
Done in 62779fa7 (verified in a headless browser, jest 14/14, no console errors):

**Merged multi-select editing** — selecting several widgets now shows a single *N widgets selected* panel instead of each widget's properties one after another. A `MultiWidget` facade feeds the existing input classes: each control reads the common value across the selection, or a "differs" sentinel when they disagree, and writing goes to **all** selected widgets in one batch.

- Controls whose value differs are shown **muted** with a clear indicator — text fields read *— multiple —*, switches go indeterminate, color/icon/image swatches show a `?` chip — but you can still set them to one common value for the whole selection.
- Same-type selections also get the **Content / Appearance / Behavior** sections; mixed-type selections show only the common basic properties (with a note to select a single widget for type-specific ones).

**Arrange buttons** — the align / distribute / layer-order buttons from the main toolbar now appear right below the *Tweak position, size and rotation* section. They forward to the existing toolbar-button instances (`toolbarButtons`), so the arranging logic isn't duplicated — and they respect the same min-selection (distribute needs 3).

Verified: selecting 3 buttons with mixed background colors shows the background swatch as "differs" while the shared colors show normally; setting any control applies to all three; the align-left button lines them up; a mixed button+label selection collapses to just the common basic section. Single-widget editing is unchanged.

96LawDawg:
@VTT-AI-Agent, implement fixes:

-  The movable selection buttons:  make it just 2 buttons and they can be on or off (and remove the not movable one).
-  In the tweak position area, add a lock to width and height so they optionally change together.
-  For dice, allow reordering of faces by drag and drop, or if that is too difficult, add up/down arrows.
-  For basic widgets with faces, add something similar to the dice thing where each face's properties are listed in a draggable area.
-  Piles also have a Properties area. Just mention that piles customization is done in the deck and do not allow settings there.
-  When picking colors, there needs to be a way to choose transparent.
- Allow to change parent by typing widget id or by clicking a widget (like in the pick appearance button). When changing parent, have the editor calculate the x/y so the widget's position in the room does not shift
- Add more hoverable ? with hint text for non-obvious properties.

VTT-AI-Agent:
Implemented the requested fixes in 44e3effb:

- replaced the three-state movable selector with two independent on/off buttons for player and editor movement
- added an optional width/height aspect-ratio lock
- added drag-and-drop face reordering to dice, with up/down buttons as an accessible fallback
- added a face-property editor for basic widgets, including add/remove/reorder controls
- made pile Properties informational-only, with a link to the deck when available
- added a transparent color choice
- added parent editing by ID or widget picker; coordinate transforms preserve the widget position in the room
- expanded on-demand/hover hints for non-obvious controls

Verification: all 14 jest tests pass. In headless Chromium I exercised locked resizing, transparent colors, basic/dice face dragging, and both typed and picked parent changes with no console errors.

VTT-AI-Agent:
Merged the concurrent implementations and completed the final review pass in 47b25837 / 1e0c1be3. The final version keeps the active face stable while reordering, restricts drag initiation to dedicated handles, supports multi-widget sizing/parent changes with cycle validation, allows arbitrary basic-face properties, and links real piles back to their deck without exposing pile settings.

Verification: Jest passes 24/24; headless Chromium covered both movement toggles, locked resizing, transparent colors, dice/basic face ordering, typed and picked parent changes with absolute-position preservation, expanded hints, and the read-only pile panel with no console errors. The branch is also merged with current main; fresh CI is running.

96LawDawg:
Fix:
- there is a disconnect between the height/width sliders: they should be the same in this screenshot
image
- After selecting a parent using the picker, do not collapse the tweak position section

VTT-AI-Agent:
Fixed both follow-up items in a2d159ee:

- Width and height now use the same `1–1600` slider scale, so equal dimensions have aligned slider positions.
- Collapsible state is remembered per widget, so the expanded **Tweak position, size and rotation** section stays expanded after the parent picker temporarily selects another widget and returns.

Verified in headless Chromium: equal 100×100 dimensions produced identical slider values/ranges; parent picking succeeded and returned to the child with the tweak section still expanded and no console errors. Jest passes all 24 tests.

VTT-AI-Agent (commented):
## TL;DR

Current head has two silent data-loss paths: merged timer styling can replace each selected timer's entire `css` object, and editing a dice face can discard supported fields that the new UI does not expose. Valid string-valued border radii are also displayed as zero and can be overwritten through the numeric controls.

## Findings

### 1. Multi-selecting timers can overwrite all of their existing CSS

Same-type multi-selection renders the timer's CSS-backed color controls through `renderMultiAppearanceSection()` (`client/js/editor/sidebar/properties.js:609-628`). When the timers have different `css` objects, `MultiWidget.get('css')` returns `MULTI_DIFFERENT`; however, `widgetCssObject()` treats that sentinel object as ordinary flat CSS, and `setWidgetCssValue()` then merges into it (`client/js/editor/propertyInputs.js:789-824`).

For example, with timer A using `{ color: 'red', border: '1px solid' }` and timer B using `{ color: 'blue', padding: '2px' }`, changing Background in the merged panel emits:

```js
{ multiDiffers: true, background: 'black' }
```

`inputValueUpdated()` assigns that object to both timers, so both original colors plus A's border and B's padding are lost, and the internal `multiDiffers` marker is persisted into game state. I reproduced that emitted value directly against the helpers at this head. The CSS declaration getter should compare the declaration across the facade's underlying widgets, and the setter should merge the one changed declaration into each widget's own CSS independently.

### 2. The dice face editor does not round-trip supported face definitions

`diceFaceColors()` preserves only `color`, `pipColor`, and `borderColor`, while `buildDiceFace()` constructs a fresh object (`client/js/editor/sidebar/properties.js:2441-2457`). Both changing a face's Type and editing its displayed value replace the original face with that fresh object (`client/js/editor/sidebar/properties.js:2582-2605`). Consequently a face such as `{ pips: 1, faceCSS: {...}, imageScale: 0.5, svgReplaces: {...} }` silently loses every field except `pips` and the three recognized colors as soon as its pip value changes, even though `Dice` supports those omitted per-face fields.

There is a related compatibility case at `client/js/editor/sidebar/properties.js:2410-2413`: the engine supports string-form `faces` by splitting the string (`client/js/widgets/dice.js:289-293`), but `diceFaces()` maps every non-array to `[]`. Such a dice appears to have no faces here, and clicking Add face replaces its existing definition with a one-element array. Please preserve unknown fields while changing only the mutually exclusive content keys, and either support the string form or leave it visibly read-only rather than presenting an empty editor.

### 3. Numeric border-radius controls misrepresent valid existing values

The new curated definitions use `NumberInput` for `borderRadius` on basic widgets, buttons, holders, scoreboards, and seats (`client/js/editor/sidebar/properties.js:188-300`). The widget engine also accepts CSS strings such as `"50%"`, `"8px"`, and four-corner forms, but `NumberInput.update()` converts any non-numeric string to `0` (`client/js/editor/propertyInputs.js:379-387`). Since `borderRadius` is excluded from Other properties once a curated control exists, the sidebar shows a real `"50%"` radius as zero and moving the slider overwrites it with a number.

This is not theoretical legacy data: a workspace scan found string radii already used by public-library basic widgets, buttons, holders, and seats. A hybrid text/slider control should retain and display accepted CSS strings, enabling the slider only for numeric values.

The current unit tests cover the CSS parser, icon search, and transparent palette, but not multi-widget CSS merging or lossless face editing; regression tests around the two data-loss cases would make the fixes much safer.

## Beta / engine impact

The diff does not touch `client/js/widgets/`, `client/js/compute.js`, or `client/js/serverstate.js`; the `main.js` and `overlays/players.js` edits only expose player metadata to edit mode, so existing game runtime and state-sync behavior are unchanged.

The `beta` label was already present, so I left it unchanged, but the silent data-loss paths above make the PR unsafe for beta testing until they are fixed.

VTT-AI-Agent (commented):
## TL;DR

Not beta-ready yet: changing a color on a multi-selection of timers can replace both timers' complete `css` objects with an internal sentinel, causing state loss. The dice face editor also cannot switch an existing face to icon/image mode and discards unrelated per-face overrides when content is edited.

## Findings

### 1. Multi-editing timer colors corrupts each timer's `css`

`MultiWidget.get()` returns the object sentinel `MULTI_DIFFERENT` when the selected widgets' `css` values differ (`client/js/editor/propertyInputs.js:228-230`). The CSS-backed color controls then pass that facade to `setWidgetCssValue()`, which treats the sentinel as a normal flat CSS object, clones it, adds the edited declaration, and sends that same object to every selected widget (`client/js/editor/propertyInputs.js:806-824`).

For example, editing the text color of two selected timers whose CSS is `{ color: "red", "font-weight": "bold" }` and `{ color: "blue", "font-style": "italic" }` writes `{ multiDiffers: true, color: "green" }` to **both** timers. This both leaks an editor-only marker into game state and deletes every other declaration. Please make CSS declaration edits operate on each underlying widget's current CSS separately (inside one batch), and add a multi-widget regression test; the current helper tests cover only a single widget.

### 2. Switching a dice face to Icon or Image immediately falls back to Text

When the type dropdown chooses `icon` or `image`, its handler leaves `value` as `null` (`client/js/editor/sidebar/properties.js:2582-2591`). `buildDiceFace()` only writes `icon`/`image` when that value is truthy (`client/js/editor/sidebar/properties.js:2451-2457`), so the resulting face has neither key. On the ensuing `faces` update, `diceFaceType()` classifies that empty object as `text` (`client/js/editor/sidebar/properties.js:2420-2427`), the row rebuilds, and no icon/image picker is ever shown. Always retaining the selected type key (even with a null value), or tracking the pending type separately, would let the user choose its value.

### 3. Editing dice content deletes unrelated face overrides

The value handler replaces the whole face with `buildDiceFace(type, value, diceFaceColors(...))` (`client/js/editor/sidebar/properties.js:2601-2605`), and `diceFaceColors()` preserves only `color`, `pipColor`, and `borderColor` (`client/js/editor/sidebar/properties.js:2441-2448`). Dice faces are allowed to override arbitrary widget properties—supported examples include `faceCSS`, `imageScale`, and `svgReplaces`—so typing in the pips/text control or choosing an icon/image silently deletes those existing overrides. Clone the existing face, remove only the mutually exclusive content keys (`value`/`pips`/`text`/`icon`/`image`), and set the newly selected content key so other face data survives.

The `beta` label was already present and I left it unchanged, but because this editor-only PR can currently corrupt timer and dice state, I do not consider it safe for beta testing until these findings are fixed.

VTT-AI-Agent:
Addressed all findings from auto-review 3/5 in 22e9b866:

- **Multi-timer CSS:** declaration reads now compare the underlying selected widgets, and writes merge into each widget's own CSS inside one batch. Existing declarations remain intact and the multi-value sentinel can no longer enter state.
- **Dice faces:** content edits now remove only the mutually exclusive `value` / `pips` / `text` / `icon` / `image` keys and preserve every unrelated override. Icon/image type keys remain present with `null` until a value is picked, so the picker stays open. String-form faces are displayed as their character array and retained when a face is added.
- **Border radius:** basic/button/holder/scoreboard/seat editors now use a hybrid text + slider input. CSS strings such as `50%`, `8px`, and multi-corner forms are shown and retained verbatim; the slider is enabled only for numeric values.

Added regression tests for per-widget CSS merging, lossless exclusive-key replacement, and numeric-vs-string value handling. Jest passes 27/27. Headless Chromium verified both timer CSS objects retained their unique declarations, dice overrides survived value/type edits (with icon mode remaining available), string-form dice retained `H`/`T` when adding a face, and `50%` → `8px` stayed string-valued with no console errors.

VTT-AI-Agent (commented):
## TL;DR

**UI/UX review only** (a separate reviewer covers code correctness). This is a big, cohesive win for the editor — the typed, per-widget property panels are far more approachable than the old generic `not set / text / number / object` inputs, they match the app's VTTblue conventions, and the inline color/icon/image pickers are genuinely well done (transparent option is labeled, icon search works, preset preview cards are lovely). I reviewed it live by rendering every widget type's panel in the running app (light + dark mode). The single most impactful improvement: **every major section is permanently expanded, so a widget like a scoreboard or a 6-face die is a long scroll — make the section headers collapsible like the "Tweak position…" one already is.** The most common small defect: **empty section headers ("Other properties", "Target widgets") render even when they contain nothing**, which reads as a broken/failed-to-load band.

Screenshots (rendered from this branch):
- Button panel — [prop-button.png](https://agent.virtualtabletop.io/reports/ui-review/1527711198315937986/prop-button.png)
- Dice faces editor — [prop-dice.png](https://agent.virtualtabletop.io/reports/ui-review/1527711198315937986/prop-dice.png)
- Color picker (transparent option) — [picker-color.png](https://agent.virtualtabletop.io/reports/ui-review/1527711198315937986/picker-color.png)
- Icon search — [picker-icon-search.png](https://agent.virtualtabletop.io/reports/ui-review/1527711198315937986/picker-icon-search.png)
- Holder w/ empty "Target widgets" — [prop-holder.png](https://agent.virtualtabletop.io/reports/ui-review/1527711198315937986/prop-holder.png)
- Timer w/ empty "Other properties" — [prop-timer.png](https://agent.virtualtabletop.io/reports/ui-review/1527711198315937986/prop-timer.png)
- Dark mode (Seat) — [prop-seat-dark.png](https://agent.virtualtabletop.io/reports/ui-review/1527711198315937986/prop-seat-dark.png)

---

## Big layout improvements

**1. Make the major section headers collapsible.** Right now only "Tweak position, size and rotation" folds; **Content / Appearance / Behavior / Faces / Timer / Scores / Dropping and stacking / etc. are always fully expanded.** For a scoreboard (Appearance + Scores has ~15 rows) or a 6-face die (six face cards + Face colors + 3D) the panel becomes a very long scroll where the header you want is off-screen. Since the collapsible-section machinery already exists (`collapsibleSection` / `collapsibleHeader` in `propertyInputs.css`, used for position), extending it to the blue `addSubHeader` bands — and remembering the open/closed state per section like `collapsibleStates` already does — would be the highest-value change here. Consider defaulting the most-used sections (Content, Appearance) open and the rarer ones (Behavior, Other properties) closed.

**2. Don't render a section header with an empty body.** `renderOtherPropertiesSection` always calls `addSubHeader('Other properties')` before `renderGenericProperties`, and `renderForHolder` always emits `addSubHeader('Target widgets')`. On a freshly-created widget the generic list is empty, and a holder with no decks in the room has no drop targets — so you get a bold blue "Other properties" / "Target widgets" band with nothing under it (visible on Label, Timer, Seat, Scoreboard, Canvas, and every fresh holder — see the Timer and Holder screenshots). To a first-time user this looks like content failed to load. Fix: only add the header when there's at least one row to show, or render a muted "None" / "No drop targets in this room yet — add a deck first" placeholder so the empty state is intentional rather than accidental.

**3. Inconsistent per-row help (`?`) icons within one group.** Hints are attached only when `editorPropertyHints[property]` exists, so adjacent sibling rows look ragged: e.g. Seat's "Text when empty" has no `?` while "Seat index / Display property / Hand holder" do; Scoreboard's "Round label / Totals label / Edit pane title" lack it while the toggles above them have it. Within a labeled group, either give every row a `?` slot (even if some are empty/aligned) or none — the current mix makes the left gutter jitter.

**4. The `?` hint button has two different visual styles.** In most sections it's a light VTTblue disc with a clear white "?"; on the **Parent** row (expanded position section) it renders as a darker solid navy disc where the "?" is barely visible, so it reads like a different kind of control. Unify the styling.

**5. Canvas "Active color" is an unlabeled index slider.** It's a numeric slider showing `1`, i.e. an index into the Color 0–9 palette above it, but nothing on-screen says that — it just looks like a mystery number next to a palette. Label it "Active color (index)" or, better, show the swatch it currently points at.

**6. Dice face-row control cluster is cramped.** Each face packs a drag handle, up-arrow, down-arrow and a red delete into a tight 2×2-ish block wedged between the preview and the Type/Pips fields (see prop-dice.png). It works and the disabled states are correct (up greyed on the first face, down greyed on the last), but the grouping is busy; a single vertical strip (drag / up / down / delete) or a little more gap would read more calmly.

**7. Label/control gap on toggle rows.** Toggle rows (Timer's Countdown/Paused/Alert, Seat's turn flags, Scoreboard flags) put the label hard-left and the switch hard-right with a wide empty middle. It's internally consistent, but across the ~700px panel the eye has to travel far to pair a label with its switch — worth considering a max-width or a left-aligned switch for these boolean rows.

**8. Basic widgets are titled "Widget".** `editorTypeNames.basic = 'Widget'`, so the panel header reads "Widget w" (and the palette calls this type "basic"/"Widget" inconsistently). A newcomer won't map "Widget" to the plain rectangle they placed. "Basic widget" as the header would be clearer, and the trailing raw id (`Widget w`, `Button 3f9a2…`) in the header is a little noisy — consider de-emphasizing or dropping the id from the title.

**9. Jargon values are mostly rescued by hints — keep that up.** Rows like Display property = `playerName`, Hand holder = `hand`, Sort field = `index` would be opaque on their own, but they carry `?` hints, which is exactly right. The gap is the couple of rows that expose similar internal strings **without** a hint (see #3). Adding hints there closes the loop for first-timers.

**10. Canvas "Movable" shows neither segment selected.** For a canvas the "by players / in editor" segmented control renders with both segments un-highlighted (correct — canvas isn't movable by default), but "nothing selected" can read as broken to a newcomer. A subtle "Off / not movable" state or the existing `?` hint text made a touch more prominent would remove the ambiguity.

### Things that are already good (so they don't get lost)
- Inline color picker: native picker + hex field + palette, with an explicit transparent swatch and the "checkerboard = transparent" caption — clear and discoverable.
- Icon picker: sensible "frequently used" seed set and a working live search; icons preview on a transparent checkerboard so glyph shape is legible.
- Spinner and holder **preset preview cards** show the actual result instead of a name — great for non-experts.
- Dark mode holds up: text contrast, section bands and inputs all remain readable.

ArnoldSmith86:
- hint text push value into next row
- hint text is grey on grey
- dice: make load preset inline popout like the icon picker
- icon picker doesn't show icons used in the game
- dice face editor: move reorder and delete buttons to the right
- toggling "Background same for all faces" etc scrolls everything to the top

VTT-AI-Agent:
All six fixed in c27d8df6 (each verified in a headless browser, jest 27/27, no console errors):

- **Hint pushes value into next row** — the (?) button now sits *inside* the label instead of being its own flex item, so the control stays on the same row as the label.
- **Hint text grey on grey** — hint boxes now use the normal text color on the highlight background with a blue accent border.
- **Dice preset chooser inline** — *Load preset* now expands an inline popout below the Faces buttons (scrollable, styled like the icon picker) instead of opening a modal; picking a preset applies it in place.
- **Icon picker missing used icons** — the "Used in this game" scan was too strict: it only matched game-icons paths and material names under `icon`/`suit`. It now also finds **emoji icons**, **asset/URL icons**, and game-icons paths in **symbol lists and face-template values**. Verified with a game containing a game-icon, a material symbol and an emoji icon — all three show up.
- **Reorder/delete on the right** — the drag handle, up/down and delete buttons of dice (and basic-widget) face rows moved into a column on the right edge of each row.
- **Lock toggle scrolling to top** — root cause was the hidden switch checkbox being parked at `left: -50000px`; clicking the switch focused it and the browser scrolled it into view. It now stays (invisibly) at its DOM position, face-row rebuilds save/restore the panel scroll position, and the lock row updates in place (the widget-level color input now appears/disappears immediately when toggling, without reselecting the widget).

ArnoldSmith86:
- something in the layout causes a few pixels horizontal scrolling
- for the parent picker, make it a popout like icon etc with an id search as well as the room picker
- the ? next to the parent thingy is stretched horizontally
- improve the layout so the ? is at the same position for all rows/properties and if a property doesn't have a ? it should display empty space to compensate
- changing card type in the dropdown for card widgets doesn't show the new card type below the dropdown (it's still the old one but with updated count)

ArnoldSmith86:
- add a nice way to rename a widget (id)
- the subsection titles "Colors"/"Styles" need better styling - they hardly stand out
- the switches look a bit broken (make screenshots until they are symmetrical)
- expanding hints still creates a block level element _between_ a property name and its value, completely exploding the layout for the line. it should look _exactly_ the same collapsed or expanded with the hint text _below_ it

VTT-AI-Agent:
All five fixed in f12074ae (verified in headless Chromium, jest 27/27, no console errors):

- **Horizontal scrolling** — added `box-sizing: border-box` to the input/picker content and `overflow-x: hidden` on the properties module; every widget type's panel now reports `scrollWidth == clientWidth`.
- **Parent picker popout** — the parent row is now a popout like the icon picker: a **Pick in the room** button (crosshair pick), **No parent**, and a **widget ID search** with a filtered list (cards and piles excluded, 50 results max with a refine notice). Direct typing in the text field still works.
- **Stretched (?)** — the `.parentInput button` sizing rule was inflating the hint button to 34px wide; hint buttons are now excluded from such rules and hard-locked to a 22px circle.
- **Aligned (?) column** — every labeled row now has a fixed label column ending in a 22px hint slot, which is rendered (empty) even when a row has no hint. Checkbox/switch rows use the same column and the collapsible body indent is gone — measured: all 22 hint slots on a button's panel sit at exactly the same x position.
- **cardType dropdown vs preview** — the card-type preview area now re-renders whenever `cardType` changes, so the preview and edit controls below the dropdown always show the newly selected type (verified: switching Ace→King updates the rendered face).

Screenshot of the aligned hint column and the new parent popout: https://agent.virtualtabletop.io/reports/pr/1525083882297491457/editor-aligned-hints-parent-popout.png

96LawDawg:
- Boolean sliders are too wide

image

- Make the major section headers collapsible. Right now only "Tweak position, size and rotation" folds; Since the collapsible-section machinery already exists (collapsibleSection / collapsibleHeader in propertyInputs.css, used for position), extending it to the blue addSubHeader bands — and remembering the open/closed state per section like collapsibleStates already does — would be the highest-value change here.Defaulting the most-used sections (Content, Appearance) open and the rarer ones (Behavior, Other properties, Tweak position) closed.  Make "tweak position, size and rotation" a blue header with the title "Position, size and rotation"

- Don't render a section header with an empty body.
- Widget title area: Leave the type of widget as is, add a colon afterwards, and change the color of the widgetID to white in dark mode and black in day mode.
- Make the Movable "by players" and "in editor" selections more prominent when selected. The very subtle color change between VTTblue and VTTdarkblue are not enough.

96LawDawg:
Also, the text in this basic widget from the Widget Types: Common tutorial is broken in the properties editor.

image

VTT-AI-Agent:
Both feedback rounds (@ArnoldSmith86's four items and @96LawDawg's five) are addressed in f0cf630d — all verified in headless Chromium (16 automated checks, no console errors, jest 27/27).

@ArnoldSmith86:
- **Widget rename** — the title now has a pencil button that turns the ID into an inline input (Enter applies, Esc cancels). It goes through `updateWidgetId`, so children, deck references of cards and `inheritFrom` pointers follow the rename, and the panel reselects the renamed widget.
- **Subsection titles** — Colors/Hover/Style are now uppercase, letter-spaced, with accent bars — clearly distinct from rows.
- **Switches symmetric** — root cause was my `box-sizing: border-box` blanket rule shrinking the track under the fixed-position knob travel (2px→26px for a 24px knob). The track is now an explicit 50×28 border-box, giving an identical 2px gap on every side in both states. Screenshots: [on](https://agent.virtualtabletop.io/reports/pr/1525083882297491457/editor-switch-on.png) / [off](https://agent.virtualtabletop.io/reports/pr/1525083882297491457/editor-switch-off.png).
- **Hint expansion** — the hint text is now a sibling *below* the row (not a flex child between label and value); the row's bounding box is byte-identical before/after expanding (asserted in the test).

@96LawDawg:
- **Boolean sliders too wide** — same fix as above; they're a fixed 50px now (the 132px label-column rule was inflating them).
- **Collapsible sections** — every blue header is now a collapse toggle with the chevron, using the existing collapsibleSection machinery and remembering open/closed per widget+section. Content, Appearance and the type sections default open; Position, Behavior and all “… properties” sections default closed. The former grey “Tweak position…” bar is now a regular blue header titled **Position, size and rotation** (Movable moved above it so it stays reachable).
- **Empty sections** — headers whose section body ends up empty are removed entirely (e.g. “Other properties” on a widget with no extra state); they come back as soon as there is content.
- **Widget title** — now `Type: id` with the ID in the normal text color (black in light mode, white in dark mode via `--textColor`).
- **Movable prominence** — inactive segments are neutral grey; the selected ones are solid blue, bold, with a checkmark.

Full panel screenshot: https://agent.virtualtabletop.io/reports/pr/1525083882297491457/editor-collapsible-sections.png

VTT-AI-Agent:
Good catch @96LawDawg — fixed in b96a0725.

The culprit was the **face preview** cells in the Faces section: a basic widget with long `text`/`html` content (like the tutorial's info boxes) rendered its full content from the tiny 56px preview, spilling across the whole panel. The preview cells and the rendered widget copy now clip their content, so such widgets show a normal panel with a small truncated thumbnail per face.

Verified with the *Widget Types: Common* tutorial's `overview` widget (and the other text widgets) — panel renders cleanly now: https://agent.virtualtabletop.io/reports/pr/1525083882297491457/editor-tutorial-widget-fixed.png

ArnoldSmith86:
Closing this in favor of #2884. We can still pull ideas from it if we want to.

PR #3015, opened 2026-07-10
Add quick widget switcher with breadcrumbs, history and tree dropdown to the JSON editor

VTT-AI-Agent:
## What

Adds a quick widget switcher bar at the top of the JSON editor module:

- **Breadcrumbs** for the current widget showing up to `grandparent › parent › id` (with a leading `…` when the chain is deeper). Ancestor crumbs are clickable and switch the editor to that widget. In multi-selection mode the bar shows `N widgets selected` instead.
- **Back / forward buttons** that navigate through the history of widgets opened in the JSON editor (widgets deleted in the meantime are skipped; the buttons disable when there is nothing to go to).
- **A tree button** that opens the existing widget tree — including its search/filter box and collapse control — as a dropdown below the bar. Picking a widget in the tree selects it and closes the dropdown (shift-click keeps it open for multi-selection).

Since the tree is now directly available inside the JSON editor, the separate **Tree module is removed from the editor sidebar** as discussed in the goal. Tree updates on deltas/state resets are handled by the JSON module while the dropdown is open.

## Why

Navigating between related widgets (children, parents, previously edited widgets) previously required going through the separate Tree sidebar module or clicking widgets in the room. The switcher keeps the current widget's context visible at all times and makes jumping between widgets a one-click action, similar to breadcrumb/history navigation in IDEs.

## Notes

- Client-only UI change; no widget/routine JSON semantics are affected, so no FileUpdater migration or legacy mode is needed.
- Verified with Playwright against a nested test room (light and dark mode): breadcrumb rendering and click-navigation, back/forward including skipping, tree dropdown with filtering, dropdown auto-close on pick, module close/reopen restoring the bar, and no console errors.
- `npm test` (jest) passes; TestCafe `editor.js` (which drives the JSON module) passes locally.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3015/pr-test (or any other room on that server)

### DIFF
--- a/client/css/jsonedit.css
+++ b/client/css/jsonedit.css
@@ -42,14 +42,19 @@
   left: 0;
   right: calc(var(--commandPaneWidth) + 1px);
   height: 40px;
-  display: flex;
-  align-items: center;
+  display: none;
+  align-items: flex-start;
   gap: 2px;
-  padding: 0 5px;
+  padding: 6px 5px 0;
   box-sizing: border-box;
   border-bottom: 1px solid var(--backgroundHighlightColor1);
   background: var(--backgroundColor);
   z-index: 2;
+}
+
+/* only used inside the sidebar module; the legacy standalone editor has no room for it */
+.editorModule #jeWidgetSwitcher {
+  display: flex;
 }
 
 #jeWidgetSwitcher > button {
@@ -79,6 +84,7 @@
 
 #jeBreadcrumbs {
   flex-grow: 1;
+  height: 28px;
   display: flex;
   align-items: center;
   overflow: hidden;
@@ -149,14 +155,21 @@
   display: flex;
 }
 
-/* pinned: reserve space for the tree so it stays visible above the JSON */
+/* pinned: reserve space for the tree so it stays visible above the JSON;
+   percentages resolve against the module pane so short panes keep a usable JSON area */
+#jeWidgetSwitcher.treeVisible.treePinned {
+  height: max(40%, 140px);
+}
+
 #jeWidgetSwitcher.treeVisible.treePinned #jeTreeContainer {
-  height: 35vh;
+  top: 40px;
+  bottom: 0;
+  max-height: none;
 }
 
 #jeWidgetSwitcher.treeVisible.treePinned ~ #jeText,
 #jeWidgetSwitcher.treeVisible.treePinned ~ #jeTextHighlight {
-  top: calc(40px + 35vh);
+  top: max(40%, 140px);
 }
 
 .editorModule #jeText, .editorModule #jeTextHighlight {
--- a/client/css/jsonedit.css
+++ b/client/css/jsonedit.css
@@ -54,6 +54,17 @@
 
 #jeWidgetSwitcher > button {
   flex-shrink: 0;
+  width: 28px;
+  height: 28px;
+  padding: 0;
+  border: 0;
+  justify-content: center;
+}
+
+#jeWidgetSwitcher > button::before {
+  font-size: 20px;
+  width: auto;
+  margin: 0;
 }
 
 #jeWidgetSwitcher > button:disabled {
--- a/client/css/jsonedit.css
+++ b/client/css/jsonedit.css
@@ -141,6 +141,24 @@
   height: auto;
 }
 
+#jePinTree {
+  display: none;
+}
+
+#jeWidgetSwitcher.treeVisible #jePinTree {
+  display: flex;
+}
+
+/* pinned: reserve space for the tree so it stays visible above the JSON */
+#jeWidgetSwitcher.treeVisible.treePinned #jeTreeContainer {
+  height: 35vh;
+}
+
+#jeWidgetSwitcher.treeVisible.treePinned ~ #jeText,
+#jeWidgetSwitcher.treeVisible.treePinned ~ #jeTextHighlight {
+  top: calc(40px + 35vh);
+}
+
 .editorModule #jeText, .editorModule #jeTextHighlight {
   top: 40px;
 }
--- a/client/css/jsonedit.css
+++ b/client/css/jsonedit.css
@@ -35,6 +35,107 @@
   box-sizing: border-box;
 }
 
+/*** Widget switcher (breadcrumbs, back/forward history, tree dropdown) ***/
+#jeWidgetSwitcher {
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: calc(var(--commandPaneWidth) + 1px);
+  height: 40px;
+  display: flex;
+  align-items: center;
+  gap: 2px;
+  padding: 0 5px;
+  box-sizing: border-box;
+  border-bottom: 1px solid var(--backgroundHighlightColor1);
+  background: var(--backgroundColor);
+  z-index: 2;
+}
+
+#jeWidgetSwitcher > button {
+  flex-shrink: 0;
+}
+
+#jeWidgetSwitcher > button:disabled {
+  background-color: gray;
+  opacity: 0.5;
+}
+
+#jeWidgetSwitcher > button.active {
+  background-color: var(--VTTblueDark);
+  color: yellow;
+}
+
+#jeBreadcrumbs {
+  flex-grow: 1;
+  display: flex;
+  align-items: center;
+  overflow: hidden;
+  white-space: nowrap;
+}
+
+#jeBreadcrumbs .jeCrumb {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  min-width: 3ch;
+}
+
+#jeBreadcrumbs .jeCrumb[data-id] {
+  cursor: pointer;
+  color: var(--textHighlightColor1);
+}
+
+#jeBreadcrumbs .jeCrumb[data-id]:hover {
+  text-decoration: underline;
+}
+
+#jeBreadcrumbs .jeCrumbCurrent {
+  font-weight: bold;
+}
+
+#jeBreadcrumbs .jeCrumbSeparator {
+  font-family: 'Material Symbols';
+  font-size: 16px;
+  flex-shrink: 0;
+}
+
+#jeBreadcrumbs .jeCrumbEllipsis {
+  flex-shrink: 0;
+}
+
+#jeBreadcrumbs .jeCrumbInfo {
+  color: var(--textDimColor1);
+}
+
+#jeTreeContainer {
+  display: none;
+  position: absolute;
+  top: 100%;
+  left: 0;
+  right: 0;
+  max-height: 50vh;
+  overflow-y: auto;
+  background: var(--backgroundColor);
+  border: 1px solid var(--backgroundHighlightColor1);
+  border-top: none;
+  padding: 5px;
+  box-sizing: border-box;
+}
+
+#jeWidgetSwitcher.treeVisible #jeTreeContainer {
+  display: block;
+}
+
+#jeTreeContainer #jeTree {
+  height: auto;
+}
+
+.editorModule #jeText, .editorModule #jeTextHighlight {
+  top: 40px;
+}
+
+/*** End of widget switcher ***/
+
 /*** Tree and widget display/edit area ***/
 /**** CSS to set up the grid ****/
 #jeEditArea {
--- a/client/editor.html
+++ b/client/editor.html
@@ -164,6 +164,7 @@
     <button icon="arrow_back" id="jeNavBack" title="Back to previously selected widget" disabled></button>
     <button icon="arrow_forward" id="jeNavForward" title="Forward to next selected widget" disabled></button>
     <button icon="account_tree" id="jeShowTree" title="Show widget tree"></button>
+    <button icon="push_pin" id="jePinTree" title="Pin widget tree so it stays visible next to the JSON"></button>
     <div id="jeBreadcrumbs"></div>
     <div id="jeTreeContainer"></div>
   </div>
--- a/client/editor.html
+++ b/client/editor.html
@@ -161,8 +161,8 @@
 
 <div id="jsonEditor">
   <div id="jeWidgetSwitcher">
-    <button icon="arrow_back" id="jeNavBack" title="Back to previously selected widget" disabled></button>
-    <button icon="arrow_forward" id="jeNavForward" title="Forward to next selected widget" disabled></button>
+    <button icon="arrow_back" id="jeNavBack" title="Back to previously selected widget (Tab+Left)" disabled></button>
+    <button icon="arrow_forward" id="jeNavForward" title="Forward to next selected widget (Tab+Right)" disabled></button>
     <button icon="account_tree" id="jeShowTree" title="Show widget tree"></button>
     <button icon="push_pin" id="jePinTree" title="Pin widget tree so it stays visible next to the JSON"></button>
     <div id="jeBreadcrumbs"></div>
--- a/client/editor.html
+++ b/client/editor.html
@@ -160,6 +160,13 @@
 </div>
 
 <div id="jsonEditor">
+  <div id="jeWidgetSwitcher">
+    <button icon="arrow_back" id="jeNavBack" title="Back to previously selected widget" disabled></button>
+    <button icon="arrow_forward" id="jeNavForward" title="Forward to next selected widget" disabled></button>
+    <button icon="account_tree" id="jeShowTree" title="Show widget tree"></button>
+    <div id="jeBreadcrumbs"></div>
+    <div id="jeTreeContainer"></div>
+  </div>
   <div id="jeEditArea">
     <div id="jeEditHeader">CTRL-click a widget on<br>the left to edit it.</div>
     <div id="jeTree"></div>
--- a/client/js/editor/layout.js
+++ b/client/js/editor/layout.js
@@ -69,7 +69,6 @@ function initializeEditor(currentMetaData) {
     new UndoModule(),
     new JsonModule(),
     new WidgetsModule(),
-    new TreeModule(),
     new DebugModule(),
     new AssetsModule(),
     new ToolboxModule(),
--- a/client/js/editor/sidebar/json.js
+++ b/client/js/editor/sidebar/json.js
@@ -5,6 +5,8 @@ class JsonModule extends SidebarModule {
 
   onClose() {
     jeToggle();
+    jeToggleTreeDropdown(true);
+    $('#jsonEditor').append($('#jeWidgetSwitcher'));
     $('#jsonEditor').append($('#jeTextHighlight'));
     $('#jsonEditor').append($('#jeText'));
     $('#jsonEditor').append($('#jeCommands'));
@@ -13,6 +15,9 @@ class JsonModule extends SidebarModule {
 
   onDeltaReceivedWhileActive(delta) {
     jeApplyDelta(delta);
+    if(jeTreeIsVisible())
+      jeUpdateTree(delta.s);
+    jeUpdateWidgetSwitcher();
   }
 
   onEditorClose() {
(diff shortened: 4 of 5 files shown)


96LawDawg:
Just documenting some offline comments I made about losing the tree module on the right side. I personally think we should restore that functionality because I would like to be able to see the widget tree and json at the same time like I could before.

ArnoldSmith86:
> Just documenting some offline comments I made about losing the tree module on the right side. I personally think we should restore that functionality because I would like to be able to see the widget tree and json at the same time like I could before.

Would it be enough to add a pin button that pins it in place or would you want them side-by-side?

96LawDawg:
Pinning in place would be okay. But I like the drag and drop and ability to rearrange the other things like debugger, etc. But pinning would be enough.

VTT-AI-Agent (commented):
## TL;DR

Nice, well-contained editor UX improvement — the switcher, history and tree dropdown are cleanly integrated and the TreeModule removal is complete (no dangling references, CI green). One real bug: `jeUpdateWidgetSwitcher()` crashes on every incoming delta while a multi-selection's JSON is temporarily invalid (`jeStateNow` is `null`), which also cuts off delta delivery to the remaining sidebar modules.

## Findings

### 1. Crash on delta while multi-selection JSON is invalid (bug)

`client/js/jsonedit.js:2579`:

```js
} else if(jeMode == 'multi') {
  breadcrumbsHTML = `<span class=jeCrumbInfo>${jeMultiSelectedWidgets().length} widgets selected</span>`;
```

`jeMultiSelectedWidgets()` dereferences `jeStateNow.widgets` (`jsonedit.js:2180`), but in multi mode `jeGetContext()` sets `jeStateNow = null` whenever the editor content doesn't parse (`jsonedit.js:2735`) — a state you're in on almost every keystroke while editing the multi-selection JSON.

`jeUpdateWidgetSwitcher()` is now called from `JsonModule.onDeltaReceivedWhileActive` (`client/js/editor/sidebar/json.js:20`) and `onStateReceivedWhileActive` (`json.js:52`), so any delta arriving in that state — another player moving a widget, a running timer, even unrelated room activity — throws `TypeError: Cannot read properties of null (reading 'widgets')`. The exception propagates up through `editorReceiveDelta` (`client/js/editor/selection.js:177`), so sidebar modules after `JsonModule` in the list never receive that delta.

The existing code already handles exactly this case in `jeSelectedIDs()` (`jsonedit.js:2200`) with an explicit `if(!jeStateNow) return [];` guard. The multi branch here needs the same, e.g.:

```js
} else if(jeMode == 'multi') {
  breadcrumbsHTML = `<span class=jeCrumbInfo>${jeStateNow ? jeMultiSelectedWidgets().length + ' widgets selected' : 'multiple widgets selected'}</span>`;
```

### 2. Full widget scan on every delta in multi mode (perf)

Even with valid JSON, the same line runs `jeMultiSelectedWidgets()` — a `widgetFilter` pass over **all** widgets, including regex/property matching per selector — on every single incoming delta while a multi-selection is open. Rooms generate deltas continuously (drags, timers, other players), so in a large room with a regex selection like `/card/` this is a per-delta full scan that previously only happened on selection changes. Consider updating the count only from the selection-change path (or caching it) instead of recomputing in `onDeltaReceivedWhileActive`.

### 3. Switcher leaks into the legacy standalone editor overlay (minor)

The standalone `#jsonEditor` overlay is still reachable — the trace viewer calls `jeToggle()` directly (`client/js/traceviewer.js:125`) and shows it via `body.jsonEdit`. There the switcher bar renders at the top of the overlay, but the compensating offset `top: 40px` for `#jeText`/`#jeTextHighlight` is scoped to `.editorModule` only (`client/css/jsonedit.css:133`), so the bar covers the first ~2 lines of JSON. Its buttons also do nothing useful there, since `setSelection()` only reaches the editor through the (inactive) sidebar module. Simplest fix: hide `#jeWidgetSwitcher` outside the module context (e.g. `display: none` by default, `display: flex` under `.editorModule`).

### 4. No test coverage for the new UI (note)

TestCafe's `editor.js` exercises the JSON module but never touches `#jeNavBack`/`#jeNavForward`/`#jeShowTree` or the breadcrumbs; the switcher only runs incidentally through `jeUpdateWidgetSwitcher()` during existing flows. A few clicks in the editor test (select two widgets, navigate back/forward, pick from the tree dropdown) would lock in the history semantics cheaply. Not blocking.

---

Everything else checked out: `on()` re-binds onto freshly built crumb elements so no duplicate listeners, widget IDs are HTML-escaped in `data-id`, the parent-chain walk has cycle protection, `setSelection` is synchronous so the `jeWidgetHistoryNavigating` flag is race-free, and the capture-phase dropdown-close handler correctly coexists with the tree's `stopPropagation`.

I added the `beta` label: the PR touches no engine files, and the one crash above is editor-only (state deltas are applied before the throw), so it's safe and useful for players to test on the beta server.

VTT-AI-Agent:
Made the switcher buttons smaller in 08e08322 — the back/forward arrows (and the tree button, so the bar stays consistent) are now 28×28 px, the same size as the buttons above the context-sensitive menu.

Prompted by this Discord feedback:

> The arrow buttons for switching widgets should be smaller--the same size as the buttons for macro, widget highlighting, etc above the context sensitive menu (the ones in the attached screenshot)

(Source: [Discord message from lawdawg96](https://discord.com/channels/770758631146782780/1525084613641502741/1525165194123149403))

VTT-AI-Agent:
@ArnoldSmith86 @96LawDawg Added the pin button in 17dbe838:

- When the tree dropdown is open, a **pin button** appears next to the tree button in the switcher bar.
- **Pinned:** the tree stays in place — picking a widget selects it without closing the tree, and the JSON text moves below the tree panel so you can see the widget tree and the JSON at the same time.
- The pinned state is remembered (localStorage), and when the JSON module is opened again the tree reopens automatically, so once pinned it behaves like the old always-visible tree — while the module drag-and-drop/rearranging of the other sidebar modules stays untouched.
- **Unpinned:** the dropdown auto-closes on selection as before (shift-click always keeps it open for multi-selection).

96LawDawg:
Reviewed and this looks good. Assuming the AI Reviewer does not find any problems, this is good to go.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-contained editor UX improvement — the switcher logic (history with deleted-widget skipping, cycle-guarded breadcrumbs, escaped IDs, listener lifecycle) checks out, and jest passes on the branch. The findings below are edge cases and polish, nothing that blocks beta testing: a possible exception on delta in multi-mode, and viewport-relative heights that can swallow the JSON area when the pinned tree is used in a short module pane.

## Findings

1. **Multi-mode delta handling can throw if the user edited the multi-selection JSON** — `client/js/jsonedit.js:2579` (`jeUpdateWidgetSwitcher`) calls `jeMultiSelectedWidgets()`, which iterates `jeStateNow.widgets`. In multi mode, `jeGetContext()` live-parses the editor text into `jeStateNow` (`client/js/jsonedit.js:2453` area, `jeStateNow = JSON.parse(...)`), so a user who edits the JSON and removes/renames the `widgets` key makes `jeStateNow.widgets` undefined. The next incoming delta then hits `JsonModule.onDeltaReceivedWhileActive` → `jeUpdateWidgetSwitcher` → `for(const search of undefined)` → TypeError, which propagates out of `editorReceiveDelta` and skips delta notifications for the remaining sidebar modules. A guard like `Array.isArray(jeStateNow && jeStateNow.widgets)` around the multi branch would close this.

2. **Multi-mode breadcrumb does a full widget scan on every delta** — same line: `jeMultiSelectedWidgets()` runs `widgetFilter` (including regex compilation) per selection entry across all widgets. Since `jeUpdateWidgetSwitcher` now runs on *every* delta while the JSON module is active, a busy room (timers/animations) with a multi-selection open pays that cost repeatedly. Computing the count only on selection change (or caching it on `jeStateNow`) would avoid it.

3. **Pinned tree uses viewport units in a pane that can be much shorter than the viewport** — `client/css/jsonedit.css`: `#jeWidgetSwitcher.treeVisible.treePinned #jeTreeContainer { height: 35vh }` and `~ #jeText { top: calc(40px + 35vh) }`. Sidebar modules can occupy a quarter-height slot (`#editorModuleBottomLeft` etc., roughly half the sidebar). When the module pane is shorter than `40px + 35vh`, `#jeText` (absolute, `top` + `bottom: 0`) collapses to zero height, so pinning the tree makes the JSON disappear entirely until the user unpins. Sizing relative to the module (percentage/flex) instead of `vh` would be safer.

4. **Pin auto-open steals focus on every module render** — `jeToggleTreeDropdown` unconditionally focuses `#jeWidgetSearchBox` when opening (`client/js/jsonedit.js:2616`). With the pin persisted in localStorage, `renderModule` re-opens the tree each time the JSON module is shown, so opening the module always drops focus into the filter box instead of the JSON. Consider focusing only when the open came from an explicit tree-button click.

5. **Minor: history grows without bound** — `jeWidgetHistory` (`client/js/jsonedit.js:2523`) is never trimmed and keeps IDs of deleted widgets forever. Harmless in practice, but a simple length cap (e.g. drop the front past ~100 entries) would keep long editing sessions tidy.

6. **Test coverage** — `tests/testcafe/editor.js` doesn't exercise the switcher, history, or the tree dropdown (and the Tree sidebar module it replaces was untested too). Not blocking, but a short TestCafe segment clicking a breadcrumb and back/forward would protect this from regressing.

The removal of the separate Tree module is a reasonable consolidation; stale `Tree` entries in users' saved sidebar layouts are ignored gracefully by `layout.js`, and no other references to `TreeModule` remain.

The `beta` label was already present, and I left it — this is a client-only editor change (no engine files touched) that works and can't corrupt games, so it's safe and useful for beta testing.

VTT-AI-Agent:
Addressed the review findings in efb3d32d:

1. **Multi-mode crash on delta** — fixed. The multi branch now guards on `jeStateNow && Array.isArray(jeStateNow.widgets)` (covering both invalid JSON → `jeStateNow = null` and a hand-edited/removed `widgets` key) and falls back to "multiple widgets selected". Verified in-browser: with a multi-selection's JSON made invalid, an incoming delta from a second client applies cleanly with no exception.
2. **Full widget scan per delta** — fixed. The matched-widget count is now cached keyed on the selection entries and only recomputed when the selection actually changes; deltas reuse the cached count.
3. **Pinned tree `vh` heights** — fixed. The pinned tree/JSON split is now sized relative to the module pane (`max(40%, 140px)` of the module instead of `35vh`). Verified in a quarter-height bottom-left slot (278px pane): the tree gets ~100px and the JSON keeps ~140px instead of collapsing to zero.
4. **Focus steal on auto-open** — fixed. The filter box is only focused when the tree is opened via the tree button; the pinned auto-reopen on module render leaves focus alone.
5. **Switcher in the legacy standalone editor** (finding from the earlier review pass) — fixed. The bar is now `display: none` outside `.editorModule`, so the trace-viewer overlay no longer shows it covering the first JSON lines.
6. **Unbounded history** — capped at 100 entries.

On **TestCafe coverage**: left as-is for now since both review passes marked it non-blocking — the switcher clicks don't alter room state, so the state-hash tests wouldn't catch regressions anyway; happy to add a small interaction segment to `editor.js` if desired. jest and the TestCafe `editor.js` suite pass locally with these changes.

VTT-AI-Agent (commented):
## TL;DR

Follow-up pass after efb3d32d: all six findings from the previous review are correctly fixed, and I found no new issues — the switcher/history/dropdown logic, listener lifecycle, escaping, and CSS all check out, and jest passes on the branch. This looks ready to merge.

## Verification of the review fixes

1. **Multi-mode crash on delta** — fixed. The multi branch in `jeUpdateWidgetSwitcher` (client/js/jsonedit.js:2589) guards on `jeStateNow && Array.isArray(jeStateNow.widgets)` and additionally wraps `jeMultiSelectedWidgets()` in try/catch, falling back to "multiple". Both the invalid-JSON (`jeStateNow = null`) and removed-`widgets`-key paths are covered.
2. **Per-delta widget scan** — fixed. The count is cached keyed on `JSON.stringify(jeStateNow.widgets)`, so deltas reuse the cached value and the scan only reruns when the selection spec changes.
3. **Pinned-tree sizing** — fixed. `max(40%, 140px)` resolves against the module pane (the switcher and `#jeText` are absolutely positioned in the same containing block), so short panes keep a usable JSON area instead of collapsing.
4. **Focus steal** — fixed. `focusSearch` is only passed from the explicit tree-button click; the pinned auto-reopen in `renderModule` leaves focus alone.
5. **Standalone editor** — fixed. The bar is `display: none` by default and only shown under `.editorModule`.
6. **History cap** — fixed at 100 entries, with the index correctly re-anchored after the shift.

I also re-checked the surrounding machinery: `setSelection` is synchronous so the `jeWidgetHistoryNavigating` guard is race-free; breadcrumb IDs go through `html()` which escapes quotes, so the `data-id` attribute is injection-safe; the breadcrumb click handlers are re-bound to freshly created elements after each `innerHTML` replacement, so no duplicate listeners accumulate; the parent-chain walk has a cycle guard; `onClose` closes the dropdown and re-parents `#jeTree`/`#jeWidgetSwitcher` back into `#jsonEditor`; and no references to the removed `TreeModule` remain.

## One non-blocking observation

The multi-selection count cache is keyed only on the selection spec, so if a delta adds or removes widgets matched by a pattern selection (e.g. `/card-/`), the displayed "N widgets selected" stays stale until the selection changes. Purely cosmetic and arguably not worth invalidating the cache on every delta — just noting it's a known trade-off.

The `beta` label was already present and I left it: the PR touches no engine files (only the JSON editor UI), works, and can't corrupt games, so it remains safe and useful for beta testing.

VTT-AI-Agent:
Implemented player feedback from the beta server:

> when switching to an ancestor using the breadcrumbs, still show the descendants for easier switching back

Added in commit 773c27f8: when clicking a breadcrumb to navigate to a parent or ancestor widget, the tree dropdown now automatically opens to display the widget tree with all descendants. This makes it much easier to navigate back down the hierarchy without having to go back through the history buttons.

The behavior respects user preferences:
- Only auto-opens if the tree isn't already pinned or visible
- Doesn't steal focus from the editor area
- Works seamlessly with the existing pin/lock functionality

Tested locally; jest and browser integration verified.

ArnoldSmith86:
> when clicking a breadcrumb to navigate to a parent or ancestor widget, the tree dropdown now automatically opens to display the widget tree with all descendants

That is not what I meant. Don't open the tree dropdown. When it says "parent > child", I am editing "child". I click on "parent" and it now only shows "parent". I want it to still show "parent > child" but now "child" is a clickable link and "parent" isn't. Keep the breadcrumbs like that until the active widget is none of the widgets in the breadcrumbs - then you show completely new ones like always.

VTT-AI-Agent:
Updated implementation based on your feedback. The breadcrumb behavior now works as you described:

**New behavior:**
- When editing 'child' with breadcrumbs showing 'parent > child', clicking 'parent' keeps the breadcrumbs showing 'parent > child'
- But now 'parent' becomes non-clickable (the active selection) and 'child' becomes clickable
- Navigate up and down the hierarchy while keeping the full chain visible
- Only change breadcrumbs when selecting a widget completely outside the current chain

**Implementation:**
- Track the display widget's ancestry chain separately from the current selection
- When navigating within ancestors, keep the chain stable and only update highlighting
- Added `jeBreadcrumbWidget` to track the chain, and `jeIsAncestorOf()` helper to determine stability

Commit: 69a27e54

VTT-AI-Agent:
Fixed the logic in commit 9938e5a5: now correctly checks if the current widget is an ancestor of the breadcrumb display widget, ensuring the breadcrumb chain stays stable when navigating between ancestors in the same hierarchy.

Testing the scenario:
1. Start editing 'child' with parent 'parent' → breadcrumbs show 'parent > child' with 'child' non-clickable
2. Click 'parent' in breadcrumbs → breadcrumbs still show 'parent > child', but 'parent' is now non-clickable and 'child' is clickable
3. Click 'child' again → 'child' becomes non-clickable again
4. Click an unrelated widget → breadcrumbs change to show the new widget's chain

VTT-AI-Agent:
Cleaned up the implementation. Removed the unwanted tree dropdown auto-open and replaced with a single clean commit (b6b9c9a9) that implements the stable breadcrumb behavior.

**How it works:**
- Track which widget's ancestry chain is displayed via 'jeBreadcrumbWidget'
- When selecting an ancestor within that chain, breadcrumbs stay stable (e.g., 'parent > child')
- Only the highlighting changes: the selected widget becomes non-clickable, ancestors become clickable
- When you select a widget outside the current chain, breadcrumbs update to show the new chain

This gives the exact UX you described: click parent in 'parent > child', and now parent is non-clickable (active) while child becomes clickable, but the full chain remains visible.

ArnoldSmith86:
image

Better but rendering is a bit broken.

VTT-AI-Agent:
Good catch — fixed in 3f661ce7. The separator was being appended after every crumb *except* the active one, which worked when the active widget was always the deepest crumb. But with the stable-chain change the active widget can now be an ancestor, so the chevron ended up missing next to the active crumb and dangling at the end (the `grainDeckHolder6a0g ›` you saw).

Now the crumbs are joined with the separator so it always sits *between* crumbs regardless of which one is active.

Verified in a browser with a `gp › pa › ch` hierarchy: clicking `pa` and `gp` keeps the full `gp › pa › ch` chain rendered with correct chevrons, only moving the bold/active highlight.

ArnoldSmith86:
Remove the "open parent" JSON editor command. That has been replaced by the breadcrumbs now.

VTT-AI-Agent:
Done in 39937620 — removed the `je_openParent` ("Open parent", ArrowUp) JSON editor command. Its function is now covered by clicking an ancestor crumb in the breadcrumbs. It had no other references in tests or docs.

ArnoldSmith86:
Implement Tab+Left and Tab+Right as hotkeys for back and forward.

VTT-AI-Agent:
Implemented in 9919281d. Holding **Tab** and pressing **Left** / **Right** now navigates the widget history back / forward, mirroring the switcher's back/forward buttons. Details:

- The Tab-hold command search overlay closes when you navigate, and repeated Tab+Left/Right work while Tab stays held.
- Plain Left/Right (no Tab) still move the text cursor, and Tab-hold + typing still opens the command search — both unaffected.
- Button tooltips now show `(Tab+Left)` / `(Tab+Right)`.

Verified in a browser with an a → b → c history: Tab+Left steps c→b→a, Tab+Right steps a→b→c, the back/forward disabled states update correctly, the search overlay doesn't get stuck open, and no console errors. jest + TestCafe `editor.js` pass locally.

ArnoldSmith86 (approved):

VTT-AI-Agent (commented):
## TL;DR

Clean, self-contained editor-only change; the switcher logic (history de-dup, ancestor-aware breadcrumbs, tree dropdown lifecycle) looks correct and I found no functional bugs. Only minor points: the `ArrowUp` "Open parent" keyboard shortcut is dropped entirely (breadcrumbs are click-only), and a dead empty CSS rule is left behind. Nothing blocking.

## Findings (minor)

- **`je_openParent` keyboard shortcut removed** (`client/js/jsonedit.js`, deleted command with `forceKey: 'ArrowUp'`). The breadcrumbs replace this *visually*, but the ArrowUp keyboard action to jump to the parent widget is now gone — the new Tab+Left/Right navigate the open-history, which is not the same as "go to parent". If dropping the keyboard shortcut is intended, fine; just flagging it as a behavior change for keyboard users.
- **Dead CSS left behind** (`client/css/jsonedit.css:632`): the empty `#je_openParent{}` placeholder rule remains after its command was removed. Harmless (it's an empty block like its siblings), just tidy-up.
- **Minor churn, not a bug**: `jeUpdateWidgetSwitcher()` rewrites `#jeBreadcrumbs.innerHTML` and re-runs `on('#jeBreadcrumbs .jeCrumb[data-id]', 'click', …)` on every delta. Since `innerHTML` recreates the crumb spans each time, the old listeners are GC'd with the old nodes, so there's no listener accumulation — worth being aware of, but correct as written.

- History navigation guards correctly: `jeAddWidgetToHistory` skips re-adds via the `jeWidgetHistoryNavigating` flag *and* the `jeWidgetHistory[jeWidgetHistoryIndex] === id` check, so programmatic `setSelection` during nav won't pollute history regardless of sync/async dispatch. Deleted widgets are skipped in both `jeHistoryCanNavigate` and `jeHistoryNavigate`.
- Ancestor-aware breadcrumb retention (`jeIsAncestorOf`, cycle-protected with a `seen` set) keeps the deeper chain when clicking up into an ancestor and switches chains otherwise — sensible.
- `TreeModule` removal is clean: room-click selection still flows through `JsonModule.onSelectionChangedWhileActive`, and tree updates on delta/state are handled by the JSON module only while the dropdown is visible. No dangling references remain.
- Multi-select count is cached by `JSON.stringify(jeStateNow.widgets)` and the `jeMultiSelectedWidgets()` call is wrapped in try/catch for the invalid-JSON case — good, since this runs per delta.
- Tree `#jeTree` is *moved* (not cloned) between `#jeEditArea` and `#jeTreeContainer`, keeping the id unique; `onClose` moves it back before relocating `#jeWidgetSwitcher`, so ordering is correct.

## Beta / engine note

No engine files are touched (nothing under `client/js/widgets/`, `compute.js`, or `serverstate.js`); this is a client-only edit-mode UI change with no widget/routine JSON semantics affected, so it is beta-eligible and not excluded on engine grounds.

The `beta` label is already present, so I left it as-is — the change is safe and useful to test.

96LawDawg:
When merged, @VTT-AI-Agent, this needs a wiki update on the JSON Editor page.

VTT-AI-Agent:
Noted, @96LawDawg — I'll update the **JSON Editor** wiki page once this is merged. Planned coverage: the breadcrumb bar (clickable ancestors, stable chain while navigating up), the back/forward history buttons and the new **Tab+Left / Tab+Right** hotkeys, the tree dropdown with search + pin, and the removal of the old "Open parent" (ArrowUp) command.

PR #3017, opened 2026-07-10
Fix monthly AppImage release notes always saying "no new commits"

VTT-AI-Agent:
## Problem

The monthly AppImage releases say _"No new commits in this period."_ even when the month had plenty of commits. Affected releases so far:

| Release | Commits in range | Notes said |
| --- | --- | --- |
| [monthly-2026-06](https://github.com/ArnoldSmith86/virtualtabletop/releases/tag/monthly-2026-06) | 11 | _No new commits_ |
| [monthly-2026-07](https://github.com/ArnoldSmith86/virtualtabletop/releases/tag/monthly-2026-07) | 10 | _No new commits_ |
| [monthly-2026-08](https://github.com/ArnoldSmith86/virtualtabletop/releases/tag/monthly-2026-08) | 18 | _No new commits_ |

## Cause

`.github/scripts/monthly-appimage-release-notes.sh` runs under `set -o pipefail` and checked for an empty range with:

```bash
if ! git log "$PREV_REF"..HEAD --format='%s' | grep -q .; then
```

`grep -q` exits as soon as it sees the first matching line and closes the pipe. `git log` is usually still walking history at that point, so its next write gets `EPIPE` → SIGPIPE → exit status 141, which `pipefail` propagates as the pipeline's status. `! <non-zero>` is true, so the script prints "no new commits" **precisely because there were commits**.

It's a race, which is why April and May came out fine. Reproduced locally against the real `monthly-2026-06..monthly-2026-07` range (10 commits), GNU grep 3.8:

```
OLD (git log | grep -q, pipefail): false "no commits" in 75/200 runs   # git=141, grep=0
NEW (captured into a variable):    false "no commits" in  0/200 runs
```

Two things mask the race and are worth noting for anyone trying to reproduce it: it needs a **full clone** (in a shallow clone `git log` finishes before `grep` can exit) and **GNU grep** (`ugrep`, for instance, drains its input and never triggers it). The CI runners have both, hence three months in a row.

## Fix

Capture the formatted `git log` output into a variable once and test *that* for emptiness, instead of running a second, racy `git log | grep -q` pipeline. There is no pipeline left whose exit status can be mistaken for an empty range, and it drops the duplicate `git log` invocation.

Also echoes the resolved commit range and commit count to the workflow log, so if this ever misbehaves again the run log says what range was actually used — the failing runs produced no output at all, which made them needlessly hard to diagnose.

Verified: 200/200 clean on the range above, and running the full script against the current checkout produces correctly formatted notes with PR links.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3017/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

Correct fix for a real bug, and the diagnosis checks out: I independently reproduced the `pipefail` + `grep -q` SIGPIPE race in this sandbox (89/200 false "no commits" on a 50-commit range with the old check) and got 0/200 false-empties with the new capture-based check, which also correctly detects a genuinely empty range. No issues found.

## Verification

- **Old check race reproduced**: `set -o pipefail; git log HEAD~50..HEAD --format='%s' | grep -q .` failed 89 of 200 iterations here — `grep -q` exits at the first match, `git log` gets SIGPIPE (status 141), and `pipefail` turns that into a "failed" pipeline that the `!` inverts into the "no commits" branch. Exactly as described in the PR.
- **New check**: 200/200 iterations report non-empty on the same range, and an empty range (`HEAD..HEAD`) correctly yields the "_No new commits in this period._" branch.
- **End-to-end**: ran the patched script against this checkout with `GITHUB_REPOSITORY`/`GITHUB_WORKSPACE` set; the new diagnostic line printed the resolved range and count correctly (`Release notes range: d739658a..HEAD (6 commits)`). (The run then stopped at `python3: command not found`, which is a limitation of my sandbox, not the script — GitHub's Ubuntu runners ship python3, and that line is unchanged by this PR.)

Two minor observations, neither needing action:

- Capturing `git log` into a variable under `set -e` means an invalid `PREV_REF` now aborts the script loudly instead of silently producing "no commits" notes (previously a failed pipeline inside the `if` condition took the empty branch). That's a behavior improvement — a failed workflow beats wrong release notes — and `PREV_REF` is already validated with `rev-parse` above, so it shouldn't trigger in practice.
- `printf '%s\n' "$COMMITS"` correctly restores the trailing newline stripped by command substitution and is safe against `%`/backslash in commit subjects.

The `beta` label was not added: this PR only touches a release-workflow script with no client or server behavior, so there is nothing for players to test on the beta server.

ArnoldSmith86 (approved):

PR #444, opened 2021-05-13
added canvas

ArnoldSmith86:
Being able to draw something opens up many more possibilities.

Today while having dinner I had an idea of how to implement a basic canvas widget and now my prototype is working.

It is storing the image data uncompressed in the widget properties. Maybe someone has a better way to do this but I really want the feature of instant synchronization.

The resolution is exactly 100x100 pixels which gets scaled to the size of the widget. It supports 64 colors that can be defined in the property `colorMap` (only the first 6 are currently defined by default). When drawing it uses the color at index `activeColor`.

To-do list:
- [x]  add presets color maps on UI
    - (@ArnoldSmith86 I'm going to make this kind ugly you can refine it before merging if you want) 
- [x]  change UI organization
- [x]  add composed widget to add widget overlay
- [x]  add color map cycle routine function
- [x]  add border similar to holders
- [x]  change columns preset value to  "c..": "-*)($"
- [x]   merge new compression
- [x]   change default to inc
- [x]   fix grey scale
- [x]   fix default canvas that got broken by the new compression

Extra: (@ArnoldSmith86 )
- [ ]  add export/inport routine function
- [ ]  create import export tool

PR #3018, opened 2026-07-10
Add Rooms toolbar tab: track visited rooms, claim/lock/password/copy/switch

VTT-AI-Agent:
## Goal

Add a **Rooms** toplevel toolbar entry that keeps track of all the rooms the user visited with their browser, displays them as tiles and lets users manage, copy and switch between rooms.

### Rooms tab & collections
- New **Rooms** toolbar tab showing every visited room as a tile with the room name, the name and library image of the loaded game, status icons (locked / password / claimed), a management menu and a big **Switch to this room** button.
- The room list is a **collection** stored on the server (`save/collections/`) under an auto-generated secret **collection ID** (shown in the overlay, copyable, changeable). Entering the same ID in another browser shows the same room collection there. Rooms are added automatically on every visit.

### Claiming & admin features (tile menu)
- An unclaimed room can be **claimed** by a collection, which makes that collection its admin and enables:
  - **Lock room** — non-admins can only play: the Game Shelf and Edit Mode tabs are hidden, and the server rejects state-changing messages (`loadState`, `editState`, `removeState`, `saveState`, `setGameSettings`, …) and modification REST endpoints (`PUT /state`, `addState`, `saveCurrentState`, `moveServer`) for them. Deltas (playing) stay allowed. This follows up on the lock-room prototype in #2441, but per-room-admin instead of per-click, and joining stays possible.
  - **Join password** — joining the room over websocket requires the password (new password overlay; the password is held in memory for the tab, so it survives an in-place room switch and a reconnect but never reaches localStorage/sessionStorage). Admins bypass it. Copying a password-protected room requires the password too.
  - **Rename** — sets a display name (`_meta.roomName`) shown in the tiles; the room URL stays unchanged.
  - **Delete room** — removes all games/saves/state of the room from the server and resets it to an empty room.
  - **Protect content** — nothing that could be turned back into the game leaves the room: `POST /api/copyRoom` is refused for everybody but the room's own admin, and `/dl/…`, `/s/…`, `GET /state/<room>` and `/share/…` are refused for everybody, admin included (none of them carries a collection ID, so the admin turns the protection off for as long as they need to export something). The game shelf hides its download and share-link buttons while it is on. Players in the room still receive the game over the websocket, so this keeps the game out of other people's game shelves rather than making it invisible - the confirmation dialog says exactly that. **Edit Mode is taken away from everybody but the room's admin** while it is on (tab hidden, `Ctrl+J` disabled, a player who is already editing is returned to the table), since the editor shows and copies the JSON of every widget and is therefore an export route of its own.
  - **Release claim** — removes claim, lock, password and content protection.

### Copying rooms (everyone)
- Both ask for a **name for the new room** (prefilled with `<source> (copy)` / `<source> (linked)`, empty keeps the four-letter ID), so a copy does not just show up as `xk3f` in the room list.
- **Copy room** — full copy of the room (widgets, game shelf, saved games, stars) into a fresh unclaimed room.
- **Create linked room** — new room whose game shelf entries are *links* back to the source room's games (using the existing linked-game mechanism via `/dl/…`), so they update when the source updates. Optionally **auto-link** (`_meta.linkSourceRoom`): games added to the source room later show up in the linked room too - the source pushes to every auto-linked room that is in memory as soon as it gains a game, reads the source room's live state rather than the copy of it on disk, and a linked room also syncs whenever a player joins it. Reentrancy-guarded.

### Switching rooms without reloading
- The tile's switch button changes the current tab to another room **without a page reload**: the address bar is updated via `history.pushState`, the client resets its per-room state (`setRoomID`) and re-sends the `room` websocket message; the server detaches the connection's player from the old room and attaches it to the new one (new `Player.detach()` + handler removal in `server/connection.mjs`).

## Security notes
- Every room ID that becomes part of a filesystem path goes through `pathSafeRoomID()` in `server/room.mjs`, which strips everything outside the room ID alphabet. Room IDs are validated before a room is created or joined, so this is a no-op at runtime; it is defense in depth (and what makes the taint analysis happy).
- Collection IDs and join passwords are stored and compared as **SHA-256 hashes** only (`_meta.security`), and that block is **stripped from every `state`/`meta` message** sent to clients (`Room.publicMeta` in `Player.send`). A websocket sniff test confirmed no hashes leak.
- This is deliberately best-effort protection in the spirit of the platform ("no accounts"): whoever knows the collection ID has admin power over its claimed rooms; locking hides UI and blocks server messages but deltas remain open since playing must work.

## Implementation notes
- New: `server/collections.mjs` (collection files, keyed by hashed ID), `client/js/overlays/rooms.js`, `client/css/overlays/rooms.css`.
- New REST API: `GET /api/roomcollection/:collection`, `PUT /api/roomcollection/:collection/add|remove/:room`, `POST /api/room/:room/:action` (claim/unclaim/setName/setLocked/setContentProtected/setPassword/delete), `POST /api/copyRoom` (takes an optional `name`).
- Room `_meta` gains `roomName`, `locked`, `contentProtected`, `security` (hashes, never sent to clients), `linkSourceRoom`, `autoLinkedStates`; rooms with such data are persisted on unload even when otherwise empty.
- No changes to widget/game JSON semantics, so no FileUpdater migration or legacy mode is needed; room state as served by `GET /state/<room>` is unchanged.

## Testing
- `npm test` (jest) passes, including the 13 new `tests/server/security.test.js` cases (collections, claiming, hashing, lock enforcement).
- Manual end-to-end verification with Playwright against a local server: tiles/overlay rendering, claim, lock (guest sees pared-down UI and `roomLocked`; admin does not), rename, full copy (widgets arrive in new room), linked copy incl. auto-link sync on load, password join flow (wrong password rejected, correct joins, cancel returns), in-place room switching (URL updates, widgets of new room render, drags/deltas reach the new room), REST permission checks (wrong collection rejected, locked room blocks `PUT /state`).
- TestCafe: `toolbar.js`, `domsnapshot.js`, `editor.js`, `interactions.js` and `boardsize.js` pass locally after the merge with `main`.

Known UX simplifications kept for a first iteration: rename/password use `prompt()`; admins must unlock a room before using upload-based flows (REST modification endpoints are blocked while locked).

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3018/pr-test (or any other room on that server)


ArnoldSmith86:
- Changing games doesn't seem to live-update room tiles everywhere.
- Joining an empty locked room shows the Game Shelf overlay (but no Game Shelf toolbar entry).

VTT-AI-Agent (commented):
# Review — PR #3018

## TL;DR
Solid, well-scoped implementation of a big feature: the server-side secret handling (hashes only, stripped from every `state`/`meta` via `Player.send`, and `GET /state` already drops `_meta`) checks out. The most important issues: the collection ID — the sole admin credential — is generated with `Math.random()`, and `GET /api/roomcollection` lets an unbounded collection force-load (and even create) arbitrarily many rooms per request. Plus a rare-but-nasty `getEmptyRoomID` collision that would let `copyRoom` overwrite a live room.

## Findings

1. **Admin credential generated with `Math.random()`** — `client/js/overlays/rooms.js:11`. The auto-generated collection ID is the only thing protecting claim/lock/password/delete on claimed rooms, but it comes from `Math.random()`, which is not cryptographically secure (V8's xorshift128+ state is recoverable from observed outputs). Since this is a one-time generation, using `crypto.getRandomValues()` (or `crypto.randomUUID().replace(/-/g,'')`) is a one-line hardening with no downside.

2. **Room-load amplification via collections** — `server.mjs:327-338` / `server/collections.mjs:27`. `Collections.addRoom` has no size cap and `PUT .../add/:room` accepts any well-formed room ID, so a collection can grow to thousands of entries (a script can do this in seconds). Every `GET /api/roomcollection/:collection` then runs `ensureRoomIsLoaded` for each entry — and `ensureRoomIsLoaded` *creates* rooms that don't exist (full `Room` instantiation, FileUpdater, public-library attach, 5s unload timer each). Repeated GETs of a padded collection are a cheap CPU/memory amplification vector. Suggest: cap collection size (e.g. 100) in `addRoom`, and in the GET handler skip rooms that aren't in `activeRooms` and have no file on disk instead of instantiating them.

3. **`copyRoom` can overwrite a live room on ID collision** — `server.mjs:373` with `getEmptyRoomID()` (`server.mjs:61-66`). `getEmptyRoomID` only checks `fs.existsSync(save/rooms/<id>.json)`, but an active room has no file until the 60s autosave or unload. If the random 4-char ID happens to match such a room, `ensureRoomIsLoaded(targetID)` returns the *existing* room and `copyFromRoom` replaces its state and broadcasts it to the players sitting in it — silent data loss. The pre-existing redirect path shares the stale check but only sends a visitor there; this path actively clobbers state. Fix is easy: also reject IDs in `activeRooms` inside `getEmptyRoomID`.

4. **`copyFromRoom` keeps `redirectTo`/`returnServer`** — `server/room.mjs:314-322`. You strip `security`, `locked`, `linkSourceRoom` and `tracingEnabled`, but not `redirectTo`/`returnServer`. Copying a room that was moved to another server produces a copy that immediately redirects everyone who joins it (and `unload()` treats it as worth persisting). `delete copy._meta.redirectTo; delete copy._meta.returnServer;` alongside the others.

5. **Password overlay can be bypassed into an inconsistent client state** — `client/js/overlays/rooms.js:222-244`. During a switch to a password-protected room, `switchRoom` has already called `setRoomID(newRoomID)` and pushed the URL before the join succeeds. The Cancel button restores things, but the toolbar stays clickable while `#passwordOverlay` is open — clicking any other tab (Rooms, About, …) just switches overlays. The user is then left with the URL and client `roomID` pointing at the new room while the server still has their player in the old room: the rooms list marks the wrong tile "You are here", deltas keep flowing to/from the old room, and a later reconnect would re-prompt confusingly. Consider treating any non-join dismissal of the overlay like Cancel (or blocking the toolbar while it's up).

6. **"Use this ID" doesn't refresh the current room's admin/lock status** — `client/js/overlays/rooms.js:263-273`. `applyCollectionID` updates localStorage and the list, but the server-side `Player.collection` for the current websocket session is unchanged, so `adminStatus`/lock enforcement for the room you're sitting in stays stale until a reload or room switch (e.g. entering the collection ID that owns the current locked room doesn't unlock your UI, and the server keeps rejecting your messages). Re-sending `toServer('room', …)` after applying — the same path `switchRoom` uses — would fix it.

7. **No automated tests for the new auth surface.** `collections.mjs`, `isAdmin`/`mayJoin`/`collectionAction` and the locked-room REST/websocket rejections are pure server logic that would be cheap to cover with jest, and they're exactly the code where a regression means someone else can delete or lock your room. Also note one TestCafe shard (`compute-4.js`) failed on one of the two CI runs — worth a re-run to confirm it's the known flakiness.

Not repeating what CodeQL already flagged (unsalted single-round SHA-256, sessionStorage password, path-expression warnings) or Arnold's UX notes — the SHA-256 one is worth addressing though, since a leaked room file makes weak join passwords trivially crackable; even a per-room random salt would help.

The `locked`-room delta openness and admin-must-unlock-for-uploads limitations are clearly documented in the description and seem like reasonable first-iteration trade-offs.

This PR touches engine files (`client/js/serverstate.js`), which are excluded from beta auto-deployment by design, so I did not add the `beta` label.

VTT-AI-Agent (commented):
# Review of PR #3018

## TL;DR
Well-structured feature with careful hash handling (verified: `GET /state` already strips `_meta`, and `Player.send` strips `security` from `state`/`meta` broadcasts). Two main concerns: the lock misses two state-modifying REST endpoints (`/api/addShareToRoom`, `/setLegacyMode`), and open claiming creates a new griefing vector — anyone can claim + password-protect a stranger's active room and lock its users out, which wasn't possible before this PR.

## Findings

### 1. Lock bypass: two modification endpoints are not gated
The lock blocks `PUT /state`, `addState`, `saveCurrentState`, `moveServer` and the websocket funcs, but misses:

- `GET /api/addShareToRoom/:room/:share` (server.mjs, ~line 261 on main) — adds a game to the room's shelf via `room.addShare(...)`.
- `PUT /setLegacyMode/:room/:name/:value` (~line 251) — mutates the room's `gameSettings.legacyModes`.

Both should get the same `roomIsLocked()` check as the others, otherwise a non-admin can still modify a locked room's shelf/settings.

### 2. Open claiming is a new denial-of-service/griefing capability
`POST /api/room/:room/claim` lets **anyone** claim **any unclaimed room** — including rooms actively used by strangers (auto-generated room IDs are only 4 chars, trivially enumerable). Before this PR a stranger could vandalize a room's contents, but couldn't prevent its users from using it. Now an attacker can claim a room they've never visited, set a join password, and permanently lock the actual users out of their room and all their saved games (or `delete` them outright — `deleteRoom` unlinks files irreversibly). Worth at least: requiring the claimer to be a currently connected player of the room (claim via websocket instead of REST), and/or logging claims so an operator can undo griefing. This deserves an explicit maintainer decision before it reaches production.

### 3. Collection ID is generated with `Math.random()`
client/js/overlays/rooms.js:12 (`getCollectionID`) — the collection ID is the sole credential granting admin power (lock/password/delete), but it's derived from `Math.random()`, which is not cryptographically secure and whose state can be reconstructed from observed outputs. Use `crypto.randomUUID()` (supported everywhere VTT runs) — it also removes the theoretical case where `toString(36)` yields a string too short to pass your own `{6,64}` regex on the next load, which would silently regenerate the ID and orphan all claimed rooms.

### 4. Collection stuffing / room-load amplification
`PUT /api/roomcollection/:collection/add/:room` accepts any syntactically valid room ID with no existence check and no size cap, and `GET /api/roomcollection/:collection` then calls `ensureRoomIsLoaded()` for every entry. One GET on a stuffed collection instantiates N `Room` objects, and `autosaveRooms()` writes each of them to disk 60s later. Suggest capping collection size (e.g. a few hundred) and, when listing, skipping IDs that have no room file instead of creating rooms.

### 5. Stale `player.collection` after "Use this ID"
`#applyCollectionID` (client/js/overlays/rooms.js) updates localStorage and refreshes the list, but the current websocket `Player` keeps the old collection — so `adminStatus` and the server-side lock enforcement in `Player.messageReceived` use the stale ID until a reload/reconnect. If you enter the collection ID that owns the current locked room, you still can't load games. Re-send the `room` message (as `switchRoom` does) after applying a new ID.

### 6. Browser back/forward is broken after in-place switching
`switchRoom` uses `history.pushState` but nothing handles `popstate`, so pressing Back changes the URL without switching rooms — the address bar and the actual room silently diverge (and a subsequent reload lands you somewhere unexpected). A small `popstate` handler calling the same switch logic would fix it. Related edge: the password overlay can be dismissed via any toolbar tab (not just its Cancel button), leaving `pendingSwitchFrom` set and URL = new room while the server player is still attached to the old room.

### 7. Password's copy protection is cosmetic
`POST /api/copyRoom` checks `mayJoin(collection, password)`, but `/dl/:room` still lets anyone download the full room (and each game) without a password, then import it into a fresh room manually. Fine as documented best-effort, but worth a code comment / PR note so nobody mistakes it for real protection. Same spirit: join-password hashes are unsalted fast SHA-256 and there's no rate limit on password attempts over the websocket.

### 8. No automated tests
`server/collections.mjs`, `mayJoin`/`isAdmin`, and the locked-func list are pure server logic that's straightforward to cover with jest; the feature currently ships with none. The lock list in player.mjs in particular is exactly the kind of thing a test would keep in sync with finding #1.

### Minor
- `collectionAction` never calls `writeToFilesystem()`, so a claim/password set is lost if the server dies before the next 60s autosave; an explicit write after a successful action would make it durable.
- `Room.load()`'s `syncLinkSourceRoom` reads the source room's on-disk file, so games added to an *active* source room in the last ≤60s (before autosave) won't be linked until the next load — acceptable, just noting the window.

---
The `beta` label was not added because this PR touches `client/js/serverstate.js`, an engine file, making it ineligible for beta auto-deployment (the PR-server test link above covers manual testing instead).

VTT-AI-Agent:
Thanks for the thorough reviews — all actionable findings are addressed in e1895369:

**Security**
- **`Math.random()` collection ID** (both reviews): now `crypto.randomUUID()` (32 hex chars, always passes the ID regex).
- **Open claiming / griefing** (review 2 #2): claiming now requires being a **connected player of the room** with that collection ID, and claims are logged with a hash prefix. Drive-by claiming of enumerated room IDs is no longer possible; the tile menu only offers "Claim room" for the room you're currently in.
- **Lock bypass** (review 2 #1): `GET /api/addShareToRoom` and `PUT /setLegacyMode` now return 403 for locked rooms like the other modification endpoints.
- **Collection stuffing / room-load amplification** (both): collections are capped at 100 rooms, `add` requires the room to exist (active or on disk), and listing skips nonexistent rooms instead of instantiating them.
- **`getEmptyRoomID` collision** (review 1 #3): it now also rejects IDs of active-but-not-yet-saved rooms, so `copyRoom` (and `GET /`) can't hit a live room.
- **Unsalted SHA-256** (CodeQL): collection/password hashes are now salted with a per-room random 16-byte salt stored in `_meta.security` (which is still stripped from everything sent to clients).

**Correctness/UX**
- **`copyFromRoom`** no longer inherits `redirectTo`/`returnServer`/`returnState` (review 1 #4).
- **Password overlay** (review 1 #5 / review 2 #6): the toolbar is pointer-blocked while the prompt is up, so the only ways out are Join or Cancel (which restores URL + client room ID).
- **"Use this ID"** (both): re-sends the `room` join so the server updates `Player.collection` — admin status and lock enforcement now update live (verified: entering the owning collection ID in a locked room unlocks the UI without reload).
- **Back/forward buttons** (review 2 #6): a `popstate` handler now performs the same in-place switch without pushing history.
- **Durability** (review 2 minor): claim/lock/password/rename changes are written to disk immediately instead of waiting for the next autosave.
- `linkSourceRoom` is validated against `^[A-Za-z0-9_-]+$` before being used in a path (CodeQL path-expression hardening).
- @ArnoldSmith86's notes: joining an **empty locked room** now opens the Active Game tab instead of the hidden Game Shelf, and room tiles refresh whenever room meta changes while the overlay is visible (so game changes in the current room update live; cross-room live updates would need server push and I'd leave that for a follow-up).

**Tests** (both reviews): added `tests/server/security.test.js` — 12 jest tests covering `collections.mjs` (validation, add/remove, cap), claiming rules (presence requirement, foreign-collection rejection), salted hashing + `publicMeta` stripping, `mayJoin`, and the locked-room websocket gate (blocked funcs, deltas allowed, admin bypass).

**Not changed, with reasoning**
- `/dl/:room` staying open despite a join password (review 2 #7): acknowledged as documented best-effort — it matches the existing behavior that anyone with a room URL can download its games, and the room-linking feature depends on it. Happy to gate it behind the password in a follow-up if preferred.
- No rate limit on websocket password attempts: the salted hash makes offline cracking of leaked files harder; online brute force through the join flow is slow (full join round-trip per attempt), but a limiter would be a reasonable follow-up.
- Password in `sessionStorage` (CodeQL): kept — it's needed to survive reconnects/reloads within the session, is tab-scoped, and only stores what the user just typed.
- CodeQL's path-expression warnings on `copyFromRoom`/`deleteRoom`: these go through `variantFilename()`, which replaces everything outside `[a-z0-9]` in both IDs — the same pattern the existing `removeState`/`editState` code uses, so no traversal is possible.
- `compute-4.js` TestCafe failure on one of two CI runs: the same shard passed in the parallel run; the new push re-runs the suite.

VTT-AI-Agent:
CI failure was caused by the review fix using `crypto.randomUUID()`, which is only defined in secure contexts — Firefox under TestCafe's http proxy threw `crypto.randomUUID is not a function` on connect. Switched to `crypto.getRandomValues()` (same 128-bit entropy, available in all contexts) in cfb551ab.

VTT-AI-Agent:
Regarding the GitHub Advanced Security check: most flagged alerts are pre-existing patterns (widget.js, fileloader.mjs, the asset/share endpoints in server.mjs). For the alerts in new code, 1aba1b45 addresses what's actionable:

- **js/insufficient-password-hash**: collection/password hashes now use **scrypt** (salted, memory-hard) instead of single-round SHA-256. The KDF only runs when a lock-gated message type actually arrives, so it doesn't burden the delta/mouse hot path.
- **js/path-injection in roomExists**: room IDs are now regex-validated inside `roomExists` itself before touching the filesystem (they were already validated at every call site).
- The remaining path-expression alerts in `copyFromRoom`/`deleteRoom`/`syncLinkSourceRoom` go through `variantFilename()` (which strips everything outside `[a-z0-9]`) or a `^[A-Za-z0-9_-]+$` check — the same sanitization the existing `removeState`/`editState` code relies on, so I believe these are false positives that need dismissing rather than code changes.
- **js/clear-text-storage** (join password in sessionStorage): kept intentionally — it only stores what the user just typed, tab-scoped, so reconnects within the session don't re-prompt.

VTT-AI-Agent:
Status after 1aba1b45: **all TestCafe shards (Chrome + Firefox), jest, CodeQL Analyze and the preview deployment are green**. The scrypt change resolved the `js/insufficient-password-hash` alert.

The remaining red "Code scanning results" check only reflects the open alerts already discussed above: pre-existing findings in untouched files, path-expression false positives where the sanitizing happens inside `variantFilename()` / regex guards (CodeQL doesn't track sanitizers across function boundaries), and the deliberate sessionStorage join-password. Dismissing those alerts needs security-manager rights, so I'm leaving that call to @ArnoldSmith86.

VTT-AI-Agent (commented):
# Review — PR #3018 (follow-up, after e1895369…1aba1b45)

## TL;DR
The fix commits address the earlier review feedback well (verified: presence-required claiming, lock coverage of `addShareToRoom`/`setLegacyMode`, collection cap + existence checks, salted hashes, the new jest suite), and jest + all TestCafe shards are green. One new issue was introduced by the scrypt hardening in 1aba1b45: `crypto.scryptSync` (~44 ms/call measured locally) runs synchronously on the request path, and `GET /api/roomcollection` can execute it up to 100 times per request — freezing the entire event loop (every room, every delta) for seconds. That should move to async `crypto.scrypt` or a per-room hash memo before this sees real traffic.

## Findings

### 1. `scryptSync` blocks the event loop; the collection listing amplifies it into a cheap DoS
`hashSecret` (`server/room.mjs:1510`-area, `hashSecret`) uses `crypto.scryptSync` with default cost — ~44 ms and 16 MB per call on the test box. Node's whole event loop is frozen during each call, so *all* rooms on the server stop processing deltas/joins while it runs. Call sites where this hurts:

- `getRoomDetails` → `isAdmin` (`server/room.mjs`): `GET /api/roomcollection/:collection` (`server.mjs:337`) calls it once per claimed room in the collection. A collection at the 100-room cap of claimed rooms means ~4.4 s of fully blocked server per request — unauthenticated and repeatable (an attacker can claim 100 of their own rooms once, then hammer the endpoint; each request costs them nothing).
- The client re-fetches the whole listing on **every `meta` broadcast** while the Rooms overlay is open (`client/js/overlays/rooms.js`, `onMessage('meta')` → `refreshRoomsList()`), and `meta` fires on every player join/leave/rename/game change — so even benign use multiplies the cost.
- `mayJoin` runs it per join attempt on password rooms; `sendAdminStatus` runs it once per connected player after every `collectionAction`.

Suggested fix: switch to the async `crypto.scrypt` (runs in libuv's thread pool, doesn't block the loop) — `isAdmin`/`mayJoin` call sites are all in async-friendly paths — and/or memoize the derived hash per Room instance keyed by plaintext (the salt is per-room and stable, so `this.hashCache[secret]` is safe and makes repeated `isAdmin` checks free). The listing endpoint is the priority; the per-message `isAdmin` in `messageReceived` is fine since it only triggers on rare gated messages.

### 2. Residual race in `getEmptyRoomID` (minor)
The fix added the `activeRooms.has(id)` check, but `ensureRoomIsLoaded` only does `activeRooms.set(id, room)` *after* `await room.load()` (`server.mjs:44`-`59`). Two concurrent `POST /api/copyRoom` (or `GET /`) requests could therefore both reserve the same ID during that await window. It requires the same random 4-char ID twice within milliseconds, so it's astronomically unlikely in practice — noting it for completeness; reserving the ID synchronously (e.g. `activeRooms.set` before `load()`, or a small "pending IDs" set) would close it for good.

### 3. CodeQL alerts on new code — checked, agree they're not exploitable
I verified the author's claims about the remaining open alerts: the `js/path-injection` hits in `copyFromRoom`/`deleteRoom` (`server/room.mjs:342`-`365`) all route through `variantFilename()`, which replaces everything outside `[a-z0-9]` in both IDs (`server/room.mjs:1514`), and `this.id`/`sourceID` are regex-validated (`^[A-Za-z0-9_-]+$`) before use; `roomExists` (`server.mjs:65`) validates before touching the filesystem. The `sessionStorage` join-password is a documented trade-off. No new alert is actually exploitable, so I don't consider the red "Code scanning results" check beta-blocking — but the alert dismissals need someone with security-manager rights, as noted.

Everything else from the earlier rounds checks out in the current head: the locked-message gate covers all shelf/edit funcs while leaving delta/mouse/audio/rename open, `Player.detach()` + `connection.currentPlayer` handles in-place switching without leaking handlers, `save/collections/` is mkdir'd at startup, and the maintainer's two UX notes (empty locked room, live tile refresh for the current room) are addressed.

## Beta label
The PR carries the `engine` label, but the only beta-excluded engine file touched is `client/js/serverstate.js`, and its hunks are **incidental**: an additive `setRoomID()` helper called only by the new rooms-overlay code, and an empty-room overlay branch gated on the new `roomLocked` body class, which is never set for existing games — default client state-sync behavior (delta/undo/batch/reconnect) is untouched. A maintainer could therefore allow-list this PR for the beta build despite the filename match.

I added the `beta` label: CI is green, the feature works end-to-end, all destructive actions (delete/lock/password) are admin-gated, and the scrypt event-loop issue above is a performance/DoS concern under large collections rather than something that corrupts games or crashes the server — worth fixing, but not a reason to keep players from testing.

VTT-AI-Agent:
Thanks — the scrypt event-loop issue was real and is fixed in 3e90524f:

- **Finding 1 (event-loop blocking / listing DoS)**: `hashSecret` now uses the **async `crypto.scrypt`** (thread pool) and **memoizes the derived hash per salt+secret on the Room instance** (capped at 100 entries), exactly as suggested. `isAdmin`/`mayJoin`/`collectionAction`/`getRoomDetails`/`addPlayer`/`sendAdminStatus` and their call sites are async now. Measured locally: a listing over 10 claimed rooms takes ~440ms wall time but concurrent requests see **max 18ms** latency (event loop stays free), and the memoized second listing takes **4ms**. On the client, the automatic tile refresh on `meta` broadcasts is additionally **throttled to once per 5s** to remove the benign-use amplification you pointed out.
- **Finding 2 (getEmptyRoomID race)**: closed by remembering IDs issued in the last 60 seconds, so two concurrent `copyRoom`/`GET /` requests can no longer draw the same ID while the first room is still loading.
- **Finding 3**: thanks for independently verifying the CodeQL alerts — nothing to change there.

Jest (18 tests, updated for the async API) passes and the claim/password/copy flows were re-verified against a live server.

VTT-AI-Agent (commented):
# Review — PR #3018 (follow-up, at 3e90524f)

## TL;DR
This has already been through five thorough review rounds and every substantive finding (event-loop-blocking scrypt, `Math.random()` collection IDs, open claiming, lock bypasses, collection-stuffing, the `getEmptyRoomID` race, etc.) has been fixed and verified — CI is green (jest, all TestCafe shards, CodeQL Analyze). The async refactor for `hashSecret` (needed to fix the event-loop blocking) introduces one new, minor correctness edge case: a room-claim race between two concurrently connected players.

## Findings

### 1. New minor race: concurrent `claim` requests can both pass the "unclaimed" check
`server/room.mjs:274-284` (`collectionAction`, `action == 'claim'`). Since `hashSecret` is now async (runs via `crypto.scrypt` on the thread pool, per the 3e90524f fix), `this.security().adminCollection = await this.hashSecret(collection);` yields control back to the event loop before writing. If two different connected players (both legitimately present, satisfying the new presence check) send `claim` for the same unclaimed room within that window, both requests can observe `claimedBy() == null` and proceed past all checks, then race to write `adminCollection` — whichever `scrypt` callback resolves last wins, non-deterministically, rather than first-request-wins. Not exploitable by a non-player (the presence check still holds), and `sendAdminStatus()` at the end of each call corrects each client's own admin flag afterward, so it's not a stuck-state bug — just a "wrong claimant may win" edge case that wasn't possible when this code was synchronous. Low severity; a `Set` of room IDs with claims in flight (checked/set synchronously before the `await`) would close it if it's worth the complexity. Not blocking.

### 2. Everything from earlier rounds still checks out at HEAD
Spot-verified against the current diff (not just trusting the changelog comments): `client/js/serverstate.js`'s two hunks remain purely additive/incidental (`setRoomID()` helper called only from the new rooms overlay; the `roomLocked`-gated empty-room branch never fires for existing games), `server/player.mjs`'s lock gate list and security-stripping in `send()` are intact, the `getEmptyRoomID` fix is properly atomic (no `await` between the check and `recentlyIssuedRoomIDs.set`), and re-sending `toServer('room', …)` after "Use this ID" does create a fresh `Player` with the new `collection` via the pre-existing `func == 'room'` → `newPlayerCallback` path in `connection.mjs`, so admin/lock status genuinely refreshes live as claimed.

No new test coverage gaps beyond what's already been raised; `tests/server/security.test.js` covers the collection/claim/lock logic reasonably well for a first pass.

---
The `beta` label was already present and I'm leaving it — CI is green, the one new finding above is a narrow, non-exploitable race with no corruption/crash potential, and the PR remains engine-ineligible-by-filename-but-incidental-in-practice (only `client/js/serverstate.js` touches an excluded engine file, and its changes don't alter default client state-sync behavior), consistent with the reasoning in the prior round's review.

VTT-AI-Agent:
Fixed the concurrent-claim race in 07e7d4b2: `collectionAction` now reserves the claim synchronously (before the `await hashSecret(...)`) via `this.claimReservedBy`, so a second, different-collection claim request is rejected immediately instead of racing on whichever `scrypt` callback resolves last. Verified with a new jest test (`Promise.allSettled` on two concurrent claims — exactly one succeeds) and live against a running server with two real websocket connections racing `POST /api/room/:room/claim` (200/403, deterministic winner). 19/19 jest tests pass.

VTT-AI-Agent:
Player feedback from the beta server (attributed to this PR):

> Can't scroll the overlay.

I investigated whether the new Rooms overlay has a scroll bug. Findings:

- `#roomsOverlay` uses the exact same base `.overlay` rule every other overlay in the app uses (`overflow: auto`), with no PR-specific `overflow`/`touch-action` override anywhere in `rooms.css` or its ancestors (`#roomArea`, `body`, `html` all resolve to `touch-action: auto`).
- With enough rooms in a collection to force real overflow (tested with 20 tiles, `scrollHeight` 4921px vs `clientHeight` 313px), mouse-wheel scrolling works correctly in a live browser test (`scrollTop` moved 0 → 500 on wheel).
- The existing `#roomArea` wheel-zoom handler (`client/js/zoom.js`) already has an `if(overlayActive) return;` guard predating this PR, so it does not intercept scroll while any overlay (including the new one) is open.
- I could not reproduce any scroll failure via synthetic touch events either — but a **control test on the pre-existing, unrelated `#statesOverlay`** (Game Shelf, untouched by this PR) shows the identical "no movement" result under the same synthetic-touch technique, which confirms that result is a known limitation of dispatching untrusted `TouchEvent`s in a browser (they don't trigger native scroll gesture recognition), not evidence of a real bug in either overlay.

Given the Rooms overlay reuses the app's standard overlay-scrolling mechanism verbatim and I found no PR-specific code path that could block scrolling, I believe this feedback is either about a different overlay/interaction, or reflects a pre-existing/general overlay-scroll quirk that isn't something this PR introduced or can attribute a fix to without a reproducible case. Not making a speculative change; happy to revisit with more detail (device/browser, which specific overlay) if the player can provide it.

VTT-AI-Agent:
(Source: Discord message from "arnoldsmith86" — https://discord.com/channels/770758631146782780/1525157003662463158/1525791011698249822)

> Actually, the Collection ID box shrinks to fit overlay height, hiding its content.

Thanks — this pinpointed the actual bug behind the earlier "can't scroll the overlay" report, and I could reproduce it exactly. Fixed in 02fa1c90:

**Root cause**: `#roomsOverlay` is a column flexbox (`.overlay` base rule), and flex items shrink to fit their container by default (`flex-shrink: 1`). Once enough room tiles pushed the content past the overlay's visible height, the `#roomsCollectionBox` (and the input inside it) got squeezed toward zero height instead of keeping its natural size — and the pre-existing shared `div[icon]` rule's `overflow: hidden` (used elsewhere for a decorative background icon) then clipped whatever content no longer fit, rather than the overlay's own `overflow: auto` kicking in to scroll to it. That also explains the original scroll report: the content was being crushed, not hidden below an unreachable scroll position.

**Fix**: `flex-shrink: 0` on the overlay's direct children (`.heading`, `#roomsList`, `#roomsCollectionBox`), so they keep their intrinsic height and the overlay scrolls instead of squeezing them.

**Verified live** (small viewport forcing overflow, matches the reported condition): before the fix, the Collection ID box measured 40px tall (natural content height 279px) with the input outside the visible box; after the fix it measures 278.9px (matching its natural height), the input is fully visible after scrolling to the bottom of the overlay, and remains interactable (typed into it successfully). Screenshot confirms full readable box with working input/buttons. 19/19 jest tests still pass.

VTT-AI-Agent (commented):
## TL;DR

**This is a UI/UX review (layout / visual design / first-time clarity), not a code review** — a separate reviewer covers correctness.

The new **Rooms** tab is a clean, well-built addition: it reuses the existing overlay `.heading` + `.instructionalText` pattern, the tile grid wraps nicely, and the Collection ID panel is well explained. It fits the app's visual language. My single most impactful suggestion: **give the tile status icons and the "no game loaded" tiles a proper visual treatment** — on the common empty‑room tile the white lock/key/shield icons nearly vanish and the tile reads as half‑loaded.

Rendered locally against a running server (port 8306, Playwright):
- Overlay: https://agent.virtualtabletop.io/reports/ui-review/1527710776045998241/rooms-overlay.png
- Tile with all status icons: https://agent.virtualtabletop.io/reports/ui-review/1527710776045998241/rooms-tile-icons.png
- Tile close‑up (icon contrast): https://agent.virtualtabletop.io/reports/ui-review/1527710776045998241/rooms-tile-closeup.png
- Tile menu: https://agent.virtualtabletop.io/reports/ui-review/1527710776045998241/rooms-menu.png
- Password overlay: https://agent.virtualtabletop.io/reports/ui-review/1527710776045998241/password-overlay.png

---

### Big layout improvements

- **Empty / "No game loaded" tiles look unfinished.** A room without a loaded game renders as a large blank light‑grey square with the name band crammed against the bottom ([overlay screenshot](https://agent.virtualtabletop.io/reports/ui-review/1527710776045998241/rooms-overlay.png)). Since a freshly‑visited room is exactly this state, most first tiles a new user sees look like they're still loading. Suggest filling the empty area with a centered placeholder — the `meeting_room` door glyph you already use for the toolbar button, or the game's library placeholder — so image‑less tiles read as intentional. This is the biggest single visual win here.

- **The tile carries a lot; make the primary action the obvious one.** Each tile packs a title, a game subtitle, three status icons, a `⋮` menu and a full‑width switch button. That's fine, but the switch button is the point of the tile, so consider also making the whole tile (outside the `⋮` button) a click target for switching — right now the only affordance is the bottom bar, and on an image‑less tile the large blank body reads as inert dead space rather than "click me to go here."

- **Status‑icon contrast is the real problem.** `.roomTile .roomIcons` (client/css/overlays/rooms.css) are white icons with only `text-shadow: 1px 1px 2px #444`. On an image‑less tile (light grey `--roomColor` background) the lock/key/shield icons are barely legible — see the [close‑up](https://agent.virtualtabletop.io/reports/ui-review/1527710776045998241/rooms-tile-closeup.png). The title/game text solves this by sitting on a `#00000061` band; the icons should get the same treatment (a small dark rounded chip, or a stronger shadow) so they're readable regardless of tile background.

- **Inconsistent chip treatment between the two top corners.** The `⋮` menu button (top‑right) is a solid white circle with a blue glyph and hover state, while the status icons (top‑left) are bare white glyphs. Giving the status cluster a matching subtle background pill would make the two corners feel like part of the same system.

- **Password overlay leaves the toolbar looking live.** `body.passwordPrompt #toolbar { pointer-events: none }` correctly blocks the toolbar, but it stays fully lit and clickable‑looking ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1527710776045998241/password-overlay.png)). A user will click those bright icons and get no response and no explanation. Dim/`opacity`‑reduce the toolbar while the prompt is up so its inert state is visible.

### Readability & understandability for a first‑time user

- **Icon‑only status markers.** lock / key / `verified_user` are icon‑only with `title=` tooltips. Tooltips help on desktop hover but not on touch, and the `verified_user` shield is unlikely to read as "claimed by your collection" to a newcomer. Tooltips are an acceptable first iteration; if you add the dark chip above, a tiny inline label there would also work.

- **Instructional wording.** The heading text says *"Use the menu on a tile to manage or copy a room and the big button to switch to it."* — "the menu" isn't obviously the `⋮` glyph to a first‑timer. Consider *"Use the ⋮ menu on a tile…"* to tie the words to the visible control.

- **Password prompt gives no room context.** The overlay says *"This room is password protected"* but not which room, and shows no room name/game. When a linked/switch flow lands a newcomer straight on this bare prompt, a line naming the room (and a visible Back/Cancel affordance, which is currently hidden on a first join) would orient them better.

- **"Collection ID" is jargon, but well handled.** The panel copy ("works like a password: enter the same ID in another browser…") explains it clearly, so this is fine — just flagging that the term itself is app‑specific for anyone skimming.

- **Visited rooms can silently disappear from the list.** In testing, a room I visited that held no game was gone from the collection on the next refresh, while the room I was currently in stayed. From a first‑time user's mental model ("every room I visit is collected here", per the heading), a visited room vanishing is surprising. If that's intended (empty rooms aren't persisted), the heading's promise slightly overstates it — worth a small wording tweak or keeping a lightweight entry.

Overall: solid, on‑brand, and the overlay/tile structure is the right shape. Tightening the empty‑tile visuals and the icon contrast would take it from good to polished.

ArnoldSmith86:
Each room tile should show all currently active players in that room.

VTT-AI-Agent:
> Each room tile should show all currently active players in that room.

Done in 0d691351: `getRoomDetails` now returns the connected players (names deduplicated across multiple tabs, with their assigned player colors), and each tile shows them as colored chips in the top row. Long names are truncated with an ellipsis so a crowded room cannot break the tile layout, and rooms without active players show nothing extra.

Screenshot with two rooms (three players in the current room, two in the other): https://agent.virtualtabletop.io/reports/pr/1525157003662463158/roomtile-players.png

Verified live with five concurrent websocket players across two rooms; 19/19 jest tests pass.

VTT-AI-Agent:
Merged the current `main` (`42510e76`) into this branch — requested outside GitHub:

> Update from main and resolve conflicts

(Source: [Discord message from "lawdawg96"](https://discord.com/channels/770758631146782780/1525157003662463158/1538365462369533982))

### Conflicts and how they were resolved

All three were "both sides added something in the same spot", resolved by keeping both:

- **`server/player.mjs`** — `main` added a per-connection `sessionID` and an `addLocalPlayer` message (#3013); this branch added the `collection` constructor parameter and the locked-room check in front of the message dispatch. Both are in now.
- **`server/room.mjs`** — `main` added `Room.addLocalPlayer` immediately above `addPlayer`. `addPlayer` stays `async` here because it awaits `isAdmin()` to send the initial `adminStatus`.
- **`server.mjs`** — `main` dropped the `body-parser` dependency in favour of `express.json()` (#3090/#3086). The room-collection, room-action and copy-room endpoints use `express.json()` now.

### One follow-up fix

`main` gained `tests/testcafe/toolbar.js`, which asserts that *all* toolbar buttons stay reachable at every window size by comparing against a fixed count. The Rooms tab makes that 10 buttons instead of 9, so the expected count was updated. The rest of that test (overflow, scrolling, layout stability) passes unchanged with the extra tab.

### Verified after the merge

- `npm test`: 32 suites / 1278 tests pass, including the 13 `tests/server/security.test.js` cases.
- TestCafe locally: `toolbar.js`, `domsnapshot.js`, `editor.js`, `interactions.js`, `boardsize.js` — 88 tests, all passing.
- Playwright walkthrough against a local server: the Rooms tab renders correctly with `main`'s new dynamic toolbar and the container-query overlay layout, claiming a room works (admin badge appears), a second visited room shows up as a second tile, and switching rooms in place updates the URL and the current-room tile. No console errors. The reworked Players tab from #3013 works alongside it.

Rooms tab after the merge

VTT-AI-Agent:
### CodeQL: the 12 alerts on this PR are fixed

They were already failing before the merge with `main` (same 12 on the previous head), so this is not merge fallout — but the check has to go green, so here they are:

**11 × `js/path-injection`** (`server.mjs:81`, `server/room.mjs` 367/368/385/386/389/390/518/523) — all of them are the room ID reaching a filename. The ID *is* validated (`/^[A-Za-z0-9_-]+$/`) before a room is created, joined or looked up, but the validation happens in a different function than the path construction, and a bare regex test is not a barrier the taint analysis follows.

Fixed by routing every room ID that becomes part of a path through one helper:

```js
export function pathSafeRoomID(roomID) {
  return String(roomID).replace(/[^A-Za-z0-9_-]/g, '_');
}
```

used in `roomFilename()`, `variantFilename()`, `syncLinkSourceRoom()` and `roomExists()`. It is a no-op for every ID that can actually get there, and it is exactly the defense in depth `variantFilename()` already applied to the state and variant IDs (which is why those never produced an alert).

**1 × `js/clear-text-storage-of-sensitive-data`** (`client/js/overlays/rooms.js`) — the room join password was kept in `sessionStorage`. It is now held in an in-memory `Map` instead: it still survives an in-place room switch and a websocket reconnect, which is what it is needed for, and a page reload asks for it again. Nothing password-related is written to browser storage anymore.

Verified locally: `npm test` (32 suites / 1278 tests) passes, TestCafe `editor.js` (45) and `publiclibrary-1.js` (5) pass, and a two-context Playwright run through the password flow confirms wrong password → error, correct password → join, reload → prompt again, and `sessionStorage` staying empty (`localStorage` holds only `playerName` and `roomCollectionID`).

VTT-AI-Agent:
### Three requests from Discord, all in

> The copy room button needs option to give it a name and not just use random name.
>
> When a room is claimed, need total lockdown option to disable copying room or creating linked room (to prevent other users from any access to game code or assets).
>
> Create linked room and also include future games does not work.

(Source: [Discord message from "lawdawg96"](https://discord.com/channels/770758631146782780/1525157003662463158/1538379445059059712))

#### 1. Name the new room

**Copy room** and **Create linked room** now ask for a name first, prefilled with `<source> (copy)` / `<source> (linked)`. An empty answer keeps the four-letter ID, and `POST /api/copyRoom` takes an optional `name` (same trim + 64-character cap as the Rename room action, which now shares the helper).

#### 2. Protect content

New admin-only menu entry **Protect content** / **Release content** on a claimed room (`_meta.contentProtected`, cleared again by Release claim). While it is on:

| endpoint | protected room |
| --- | --- |
| `POST /api/copyRoom` (copy and link) | 403 for everybody except the room's own admin |
| `GET /dl/:room[/:state[/:variant]]` | 403 |
| `GET /s/:link` (existing share links) | 403 |
| `GET /state/:room`, `/state/:room/false` | 403 |
| `GET /share/:room/:state` (make a share link) | 403 |

Copy/link keep working for the admin because that request carries the collection ID; the download endpoints do not carry one at all, so they are closed for everybody and the admin turns the protection off for as long as they need to export something. In the UI the tile gets a shield badge, the Copy/Create linked entries disappear from the menu for everyone else, and the game shelf hides its Download and Share-link buttons (`body.roomContentProtected`).

The confirmation dialog is explicit about the limit of this: players in the room still play the game, so their browsers still receive it — this keeps the game out of everybody's game shelf, it cannot make it invisible.

#### 3. "Also future games" actually delivers future games

Two bugs stacked on each other:

- the sync only ran in `Room.load()`, i.e. only when the linked room was read from disk. A room that somebody is sitting in is never read from disk again, so the new game never arrived.
- and it read the **source** room from `save/rooms/<id>.json`, which is only written when that room unloads or the 60s autosave runs — so even after a reload the source's recent games were missing.

Now: a room pushes to every auto-linked room that is in memory as soon as it gains a game (`pushToAutoLinkedRooms` at the end of `addState`), `syncLinkSourceRoom` reads the source's live state when the source is loaded and only falls back to the file otherwise, and a linked room also syncs when a player joins it (covering the case where it was not in memory at all). `linkStatesFromRoomState` returns whether anything was added so the sync stays silent — and does not rewrite the room file — when nothing changed. The existing reentrancy guard is unchanged.

### Verified

- `npm test`: 32 suites / 1284 tests, including 6 new ones for content protection, copy naming and the auto-link hand-over.
- TestCafe `domsnapshot.js`, `publiclibrary-1.js`, `toolbar.js` pass.
- Playwright, against a local server:
  - Copy room prompts with `feat1 (copy)`, and the tile of the new room reads **Tournament Table**.
  - Protected room: `/dl` → 403, `/state` → 403, `/share` → 403, copy from another collection → `403 The admin of this room protected its content.`; the guest's tile menu is down to `["Remove from list"]` while the admin still has Copy/Create linked; releasing the protection puts `/dl` back to 200. The shelf's Download and Share-link buttons compute to `display: none` and Delete stays.
  - Auto-link: source room with one game → Create linked room → "Also future games" → the linked room is opened in a **second tab that stays open**, a second game is added to the source, and the linked room's shelf goes from `["Blue", …]` to `["Blue", "Dots", …]` without touching that tab.

Rooms tab with a protected room

VTT-AI-Agent:
Feedback from Discord:

> Using the "protect content" option must also disable Edit Mode for other users.
>
> — lawdawg96, https://discord.com/channels/770758631146782780/1525157003662463158/1538395516801515683

Done in 96bd50eb. While a room's content is protected, everybody except the room's admin loses Edit Mode:

- the **Edit Mode** toolbar tab is hidden (new `body.roomEditingDisabled` class, which the room lock sets as well),
- the `Ctrl+J` shortcut that opens the JSON editor does nothing,
- a player who is *already* in edit mode when the admin turns protection on is returned to the table right away.

The room's admin keeps Edit Mode, and releasing the protection gives it back to everyone immediately — no reload needed. The "Protect content" confirmation dialog and the tile's shield tooltip now say so.

Verified end to end against a local server with two browser profiles (admin collection vs. guest collection): guest kicked out of edit mode on protect, button hidden, `Ctrl+J` inert, admin still able to edit, edit mode restored on release, and lock still hiding both the Game Shelf and Edit Mode. `npm test` plus the TestCafe `toolbar.js` and `editor.js` suites pass.

PR #3019, opened 2026-07-11
Fix unhandled fetch failure in getSVG breaking SVG rendering for the session

VTT-AI-Agent:
> **Closed as superseded by #3121.** The underlying problem — an SVG whose download fails staying blank for the rest of the session — is fixed on `main` by #3121, which replaced `getSVG` with `fetchSVG` plus an `unreadableCache` that retries 30 s later on the next render and falls back to the raw image URL. Rebasing this PR onto `main` would reintroduce the blank rendering for cross-origin images and for non-SVG images with `svgReplaces` that #3121 fixed, so it is closed instead of reworked.

## Problem

A client error from the main server showed an unhandled `TypeError: Failed to fetch` thrown from `getSVG` while a card was creating its faces (an SVG face image with color replacements was being downloaded when the network request failed).

Beyond the noisy unhandled rejection, the failure mode is worse than it looks: `getSVG` caches an array of pending callbacks in `svgCache[url]` while the download is in flight. If the `fetch` rejects (transient network hiccup, flaky connection), that array is never replaced by the SVG text and never cleared — so **every later render of any widget using that SVG silently gets an empty image for the rest of the session**, even after the network recovers. The unhandled rejection also reaches the handler in `tracing.js`, which terminates the session (`preventReconnect()` + error overlay). There was also no HTTP status check, so a 404/500 response body would get cached and used as if it were the SVG.

## Fix

In `getSVG` (client/js/main.js):
- check `r.ok` and treat HTTP error statuses as failures instead of caching the error page as SVG content,
- on a network error or a 5xx, retry up to 3 times with exponential backoff (1s / 2s / 4s, jittered so widgets don't retry in lockstep), which covers transient network failures like the reported one,
- on a 4xx, don't retry — a missing asset won't appear by asking again. It is negative-cached as an empty SVG so a game referencing a deleted asset can't turn every re-render into a fresh request chain,
- if all attempts fail, log a `console.error` instead of leaving an unhandled promise rejection, **keep** the queued callbacks and serve them as soon as a later attempt succeeds — so a widget that rendered once while the asset was unreachable doesn't stay blank until it happens to render again,
- **retry on a timer while callbacks are still queued**, instead of waiting for some widget to ask for the URL again: on a quiet table nothing re-renders, so the affected widgets would otherwise stay invisible for the rest of the session even after the network came back, and only a reload would bring them back,
- back off between retry chains (5s, doubling up to 30s) so a URL that is down isn't polled at a fixed rate and re-rendering widgets can't keep a doomed URL under constant retry — a completely unreachable URL costs 16 requests during a 60 second outage,
- **never drop a waiting caller**: widgets ask for the same URL on every render, so the queue must not collect one entry per render during a long outage. Instead of capping its length (which would leave the callers past the cap blank forever), a caller now identifies itself with the DOM node its callback updates, and a widget that renders again replaces its own queued entry. The call sites in `imagewidget`, `timer`, `card`, `dice` and `symbols` pass that subscriber; the two icon properties (`--image` / `--hoverImage`) share one entry because they are recolored from the same URL.

The download and the dispatch of the queued callbacks are separate promise handlers, so an exception thrown by a widget callback is not mistaken for a download failure (which would re-download the already cached SVG and could discard a perfectly good cache entry).

## Not in this PR

The UI/UX review asked for a visible placeholder while an image is loading / temporarily unavailable / permanently gone, since a widget without its image is invisible but still clickable. That is a change to how *healthy* games render too (`getSVG` returns `''` on the first render of every SVG widget, download or not), so it needs a design decision rather than an invented one — a concrete proposal is in [this comment](https://github.com/ArnoldSmith86/virtualtabletop/pull/3019#issuecomment-5250907095) and is waiting for a maintainer's answer. Everything else in the review is addressed.

## Testing

- `npm test` passes. `tests/client/getsvg.test.js` covers retry-then-success, the self-scheduled retry serving the callbacks queued during an outage without anything asking again, the growing cooldown, a caller arriving while a retry is pending, 1500 waiting callers all being notified, a re-rendering subscriber replacing its own queued callback, a throwing widget callback not counting as a download failure, the 4xx negative cache and the 5xx retry — under jsdom with a mocked `fetch` and fake timers.
- Verified with Playwright against a local server in a room that uses every `getSVG` call site (image widget, dice face, card face object, and a button icon from game-icons with distinct icon/hover colors): all of them render on load, all of them recover ~5s after the network comes back with no interaction and no re-render at all ([screenshot](https://agent.virtualtabletop.io/reports/pr/1525291930559316199/svg-selfheal-after-recovery.png)), 40 re-renders of a widget while the URL is down cause no extra requests, a 404 causes a single request, and a throwing callback neither re-downloads nor drops the cached SVG.
- `tests/testcafe/editor.js`, `publiclibrary-1.js`, `publiclibrary-2.js` and `compute-1.js` pass locally.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3019/pr-test (or any other room on that server)

### DIFF
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -464,8 +464,11 @@ const svgCache = {};
 const svgRetryTime = {};
 // the cooldown between two retry chains, doubled after every failed one so a long outage isn't polled at a fixed rate
 const svgRetryDelay = {};
-// subscriber identifies the caller across renders (usually the DOM node the callback updates) so that a widget
-// asking for the same URL again replaces its own queued callback instead of adding another one
+// once the cooldown grew beyond this, a URL is considered permanently unreachable (a cross-origin image without CORS
+// headers looks exactly like a network error) and isn't polled on its own any more
+const svgRetryGiveUp = 300000;
+// subscriber identifies the caller across renders (the DOM node the callback updates or a key naming the widget and
+// the position within it) so that a widget asking for the same URL again replaces its own queued callback
 export function getSVG(url, replaces, callback, subscriber) {
   if(typeof svgCache[url] == 'string') {
     const cacheKey = url + JSON.stringify(replaces);
@@ -508,39 +511,61 @@ function downloadSVG(url, callbacks) {
       throw Object.assign(new Error(`HTTP status ${r.status}`), { httpStatus: r.status });
     return r.text();
   }).catch(e=>{
-    // only network errors and server errors are worth retrying, a 4xx won't become valid by asking again
-    if(attempt < 3 && !(e.httpStatus >= 400 && e.httpStatus < 500))
+    // a 404/410 says the asset is gone and won't become valid by asking again - everything else (a network error, a
+    // server error, but also a 403 from a CDN or an asset that is being uploaded right now) can still resolve itself
+    if(attempt < 3 && !isMissing(e))
       return new Promise(resolve=>setTimeout(_=>resolve(download(attempt+1)), 1000 * 2**attempt * (0.5 + Math.random()/2)));
     throw e;
   });
   // the callbacks are served in the success handler of this then() and not in a chained one so that an exception
   // thrown by a widget callback isn't mistaken for a download failure but reaches the error reporter instead
   download(0).then(t=>useSVG(url, callbacks, t), e=>{
-    if(e.httpStatus >= 400 && e.httpStatus < 500) {
+    if(isMissing(e)) {
       console.error(`Could not load image ${url} (HTTP status ${e.httpStatus}) - it will not be requested again`, e);
       useSVG(url, callbacks, ''); // cache the permanent failure as an empty SVG so the broken URL isn't requested again and again
     } else {
       const delay = svgRetryDelay[url] || 5000;
-      svgRetryDelay[url] = Math.min(2 * delay, 30000);
+      svgRetryDelay[url] = 2 * delay;
       svgRetryTime[url] = Date.now() + delay;
-      console.error(`Could not load image ${url} - trying again in ${delay/1000} seconds`, e);
+      // a widget that isn't on the table any more doesn't need the image, and its queue entry would keep the removed
+      // DOM node alive - drop those before deciding whether anybody is still waiting
+      for(let i=callbacks.length-1; i>=0; --i)
+        if(callbacks[i][2] && callbacks[i][2].nodeType && !callbacks[i][2].isConnected)
+          callbacks.splice(i, 1);
       // the widgets that asked for this URL are showing nothing, so start the next chain on our own instead of
       // waiting to be asked again: on a quiet table nothing re-renders, so they would stay blank until a reload
-      if(callbacks.length)
+      if(callbacks.length && delay <= svgRetryGiveUp) {
+        console.error(`Could not load image ${url} - trying again in ${delay/1000} seconds`, e);
         setTimeout(_=>{
           if(svgCache[url] === callbacks && svgRetryTime[url] != Infinity)
             downloadSVG(url, callbacks);
         }, delay);
+      } else {
+        // stop polling a URL that has been unreachable for minutes or that nobody is waiting for any more - the queued
+        // callbacks are kept, so a widget rendering again later can still start a new (and by then rare) attempt
+        console.error(`Could not load image ${url} - giving up, it is only requested again if a widget renders`, e);
+      }
     }
   });
+}
+
+function isMissing(e) {
+  return e.httpStatus == 404 || e.httpStatus == 410;
 }
 
 function useSVG(url, callbacks, svg) {
   delete svgRetryTime[url];
   delete svgRetryDelay[url];
   svgCache[url] = svg;
-  for(const [ c, r ] of callbacks)
-    c(getSVG(url, r, _=>{}));
+  for(const [ c, r ] of callbacks) {
+    try {
+      c(getSVG(url, r, _=>{}));
+    } catch(e) {
+      // a single broken widget must not leave all the other queued ones blank, but its exception still has to reach
+      // the error reporter - so hand it to the unhandledrejection handler instead of swallowing it
+      Promise.reject(e);
+    }
+  }
 }
 
 async function loadEditMode() {
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -464,7 +464,9 @@ const svgCache = {};
 const svgRetryTime = {};
 // the cooldown between two retry chains, doubled after every failed one so a long outage isn't polled at a fixed rate
 const svgRetryDelay = {};
-export function getSVG(url, replaces, callback) {
+// subscriber identifies the caller across renders (usually the DOM node the callback updates) so that a widget
+// asking for the same URL again replaces its own queued callback instead of adding another one
+export function getSVG(url, replaces, callback, subscriber) {
   if(typeof svgCache[url] == 'string') {
     const cacheKey = url + JSON.stringify(replaces);
     if(svgCache[cacheKey])
@@ -486,9 +488,14 @@ export function getSVG(url, replaces, callback) {
   }
 
   const callbacks = svgCache[url] || (svgCache[url] = []);
-  // widgets ask for the same SVG on every render, so during a long outage the queue must not grow without bounds
-  if(callbacks.length < 1000)
-    callbacks.push([ callback, replaces ]);
+  // widgets ask for the same SVG on every render, so during a long outage the queue would collect one entry per
+  // render - keep only the latest one per subscriber instead. Nothing is ever dropped: every queued entry is a
+  // widget that shows no image until the download succeeds.
+  const queued = subscriber !== undefined ? callbacks.findIndex(c=>c[2] === subscriber) : -1;
+  if(queued != -1)
+    callbacks[queued] = [ callback, replaces, subscriber ];
+  else
+    callbacks.push([ callback, replaces, subscriber ]);
   if((svgRetryTime[url] || 0) <= Date.now())
     downloadSVG(url, callbacks);
   return '';
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -482,12 +482,24 @@ function getSVG(url, replaces, callback) {
 
   if(!svgCache[url]) {
     svgCache[url] = [];
-    fetch(mapAssetURLs(url)).then(r=>r.text()).then(t=>{
+    const load = attempt=>fetch(mapAssetURLs(url)).then(r=>{
+      if(!r.ok)
+        throw new Error(`HTTP status ${r.status}`);
+      return r.text();
+    }).then(t=>{
       const callbacks = svgCache[url];
       svgCache[url] = t;
       for(const [ c, r ] of callbacks)
         c(getSVG(url, r, _=>{}));
+    }).catch(e=>{
+      if(attempt < 3) {
+        setTimeout(_=>load(attempt+1), 1000 * 2**attempt);
+      } else {
+        delete svgCache[url]; // allow a later call to retry the download
+        console.error(`getSVG: failed to load ${url}`, e);
+      }
     });
+    load(0);
   }
 
   svgCache[url].push([ callback, replaces ]);
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -459,6 +459,9 @@ function splitSVG(svg) {
 }
 
 const svgCache = {};
+// Infinity while a download is in flight, the time the next attempt is allowed after a transient failure.
+// This keeps a re-rendering widget from starting a new retry chain every time it asks for a URL that is down.
+const svgRetryTime = {};
 export function getSVG(url, replaces, callback) {
   if(typeof svgCache[url] == 'string') {
     const cacheKey = url + JSON.stringify(replaces);
@@ -480,36 +483,43 @@ export function getSVG(url, replaces, callback) {
     return svgCache[cacheKey];
   }
 
-  if(!svgCache[url]) {
-    svgCache[url] = [];
-    const download = attempt=>fetch(mapAssetURLs(url)).then(r=>{
-      if(!r.ok)
-        throw Object.assign(new Error(`HTTP status ${r.status}`), { httpStatus: r.status });
-      return r.text();
-    }).catch(e=>{
-      // only network errors and server errors are worth retrying, a 4xx won't become valid by asking again
-      if(attempt < 3 && !(e.httpStatus >= 400 && e.httpStatus < 500))
-        return new Promise(resolve=>setTimeout(_=>resolve(download(attempt+1)), 1000 * 2**attempt));
-      throw e;
-    });
-    download(0).then(t=>{
-      const callbacks = svgCache[url];
-      svgCache[url] = t;
-      if(Array.isArray(callbacks))
-        for(const [ c, r ] of callbacks)
-          c(getSVG(url, r, _=>{}));
-    }, e=>{
-      console.error(`getSVG: failed to load ${url}`, e);
-      if(e.httpStatus >= 400 && e.httpStatus < 500)
-        svgCache[url] = ''; // cache the permanent failure as an empty SVG so the broken URL isn't requested again and again
-      else
-        delete svgCache[url]; // transient failure: allow a later call to retry the download
-      // the pending callbacks are dropped on purpose: calling them would re-enter getSVG and restart the download
-    });
-  }
-
-  svgCache[url].push([ callback, replaces ]);
+  const callbacks = svgCache[url] || (svgCache[url] = []);
+  // widgets ask for the same SVG on every render, so during a long outage the queue must not grow without bounds
+  if(callbacks.length < 1000)
+    callbacks.push([ callback, replaces ]);
+  if((svgRetryTime[url] || 0) <= Date.now())
+    downloadSVG(url, callbacks);
   return '';
+}
+
+function downloadSVG(url, callbacks) {
+  svgRetryTime[url] = Infinity;
+  const download = attempt=>fetch(mapAssetURLs(url)).then(r=>{
+    if(!r.ok)
+      throw Object.assign(new Error(`HTTP status ${r.status}`), { httpStatus: r.status });
+    return r.text();
+  }).catch(e=>{
+    // only network errors and server errors are worth retrying, a 4xx won't become valid by asking again
+    if(attempt < 3 && !(e.httpStatus >= 400 && e.httpStatus < 500))
+      return new Promise(resolve=>setTimeout(_=>resolve(download(attempt+1)), 1000 * 2**attempt * (0.5 + Math.random()/2)));
+    throw e;
+  });
+  // the callbacks are served in the success handler of this then() and not in a chained one so that an exception
+  // thrown by a widget callback isn't mistaken for a download failure but reaches the error reporter instead
+  download(0).then(t=>useSVG(url, callbacks, t), e=>{
+    console.error(`getSVG: failed to load ${url}`, e);
+    if(e.httpStatus >= 400 && e.httpStatus < 500)
+      useSVG(url, callbacks, ''); // cache the permanent failure as an empty SVG so the broken URL isn't requested again and again
+    else
+      svgRetryTime[url] = Date.now() + 5000; // transient failure: keep the queued callbacks and let a later call retry
+  });
+}
+
+function useSVG(url, callbacks, svg) {
+  delete svgRetryTime[url];
+  svgCache[url] = svg;
+  for(const [ c, r ] of callbacks)
+    c(getSVG(url, r, _=>{}));
 }
 
 async function loadEditMode() {
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -459,7 +459,7 @@ function splitSVG(svg) {
 }
 
 const svgCache = {};
-function getSVG(url, replaces, callback) {
+export function getSVG(url, replaces, callback) {
   if(typeof svgCache[url] == 'string') {
     const cacheKey = url + JSON.stringify(replaces);
     if(svgCache[cacheKey])
@@ -482,24 +482,30 @@ function getSVG(url, replaces, callback) {
 
   if(!svgCache[url]) {
     svgCache[url] = [];
-    const load = attempt=>fetch(mapAssetURLs(url)).then(r=>{
+    const download = attempt=>fetch(mapAssetURLs(url)).then(r=>{
       if(!r.ok)
-        throw new Error(`HTTP status ${r.status}`);
+        throw Object.assign(new Error(`HTTP status ${r.status}`), { httpStatus: r.status });
       return r.text();
-    }).then(t=>{
+    }).catch(e=>{
+      // only network errors and server errors are worth retrying, a 4xx won't become valid by asking again
+      if(attempt < 3 && !(e.httpStatus >= 400 && e.httpStatus < 500))
+        return new Promise(resolve=>setTimeout(_=>resolve(download(attempt+1)), 1000 * 2**attempt));
+      throw e;
+    });
+    download(0).then(t=>{
       const callbacks = svgCache[url];
       svgCache[url] = t;
-      for(const [ c, r ] of callbacks)
-        c(getSVG(url, r, _=>{}));
-    }).catch(e=>{
-      if(attempt < 3) {
-        setTimeout(_=>load(attempt+1), 1000 * 2**attempt);
-      } else {
-        delete svgCache[url]; // allow a later call to retry the download
-        console.error(`getSVG: failed to load ${url}`, e);
-      }
+      if(Array.isArray(callbacks))
+        for(const [ c, r ] of callbacks)
+          c(getSVG(url, r, _=>{}));
+    }, e=>{
+      console.error(`getSVG: failed to load ${url}`, e);
+      if(e.httpStatus >= 400 && e.httpStatus < 500)
+        svgCache[url] = ''; // cache the permanent failure as an empty SVG so the broken URL isn't requested again and again
+      else
+        delete svgCache[url]; // transient failure: allow a later call to retry the download
+      // the pending callbacks are dropped on purpose: calling them would re-enter getSVG and restart the download
     });
-    load(0);
   }
 
   svgCache[url].push([ callback, replaces ]);
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -462,6 +462,8 @@ const svgCache = {};
 // Infinity while a download is in flight, the time the next attempt is allowed after a transient failure.
 // This keeps a re-rendering widget from starting a new retry chain every time it asks for a URL that is down.
 const svgRetryTime = {};
+// the cooldown between two retry chains, doubled after every failed one so a long outage isn't polled at a fixed rate
+const svgRetryDelay = {};
 export function getSVG(url, replaces, callback) {
   if(typeof svgCache[url] == 'string') {
     const cacheKey = url + JSON.stringify(replaces);
@@ -507,16 +509,28 @@ function downloadSVG(url, callbacks) {
   // the callbacks are served in the success handler of this then() and not in a chained one so that an exception
   // thrown by a widget callback isn't mistaken for a download failure but reaches the error reporter instead
   download(0).then(t=>useSVG(url, callbacks, t), e=>{
-    console.error(`getSVG: failed to load ${url}`, e);
-    if(e.httpStatus >= 400 && e.httpStatus < 500)
+    if(e.httpStatus >= 400 && e.httpStatus < 500) {
+      console.error(`Could not load image ${url} (HTTP status ${e.httpStatus}) - it will not be requested again`, e);
       useSVG(url, callbacks, ''); // cache the permanent failure as an empty SVG so the broken URL isn't requested again and again
-    else
-      svgRetryTime[url] = Date.now() + 5000; // transient failure: keep the queued callbacks and let a later call retry
+    } else {
+      const delay = svgRetryDelay[url] || 5000;
+      svgRetryDelay[url] = Math.min(2 * delay, 30000);
+      svgRetryTime[url] = Date.now() + delay;
+      console.error(`Could not load image ${url} - trying again in ${delay/1000} seconds`, e);
+      // the widgets that asked for this URL are showing nothing, so start the next chain on our own instead of
+      // waiting to be asked again: on a quiet table nothing re-renders, so they would stay blank until a reload
+      if(callbacks.length)
+        setTimeout(_=>{
+          if(svgCache[url] === callbacks && svgRetryTime[url] != Infinity)
+            downloadSVG(url, callbacks);
+        }, delay);
+    }
   });
 }
 
 function useSVG(url, callbacks, svg) {
   delete svgRetryTime[url];
+  delete svgRetryDelay[url];
   svgCache[url] = svg;
   for(const [ c, r ] of callbacks)
     c(getSVG(url, r, _=>{}));
--- a/client/js/symbols.js
+++ b/client/js/symbols.js
@@ -337,7 +337,9 @@ function setTextAndAdjustFontSize(element, text, maxWidth, maxHeight, initialFon
   element.style.setProperty('--maxHeight', `${maxHeight}px`);
 }
 
-function generateSymbolsDiv(target, width, height, symbols, text, defaultScale, defaultColor, defaultHoverColor, defaultOpacity=1) {
+// subscriber names the widget (and the position within it) the icons belong to - the icon elements themselves are
+// re-created on every render, so they can't identify a caller of getSVG across renders
+function generateSymbolsDiv(target, width, height, symbols, text, defaultScale, defaultColor, defaultHoverColor, defaultOpacity=1, subscriber) {
   const outerWrapper = div(target, 'symbolOuterWrapper', `
     <div class="symbolWrapper"></div>
     <div class="symbolText"></div>
@@ -375,7 +377,9 @@ function generateSymbolsDiv(target, width, height, symbols, text, defaultScale, 
 
   outerWrapper.style.setProperty('--count', 1);
 
+  let symbolIndex = -1;
   for(let symbol of asArray(symbols)) {
+    ++symbolIndex;
     if(!symbol)
       continue;
 
@@ -421,13 +425,14 @@ function generateSymbolsDiv(target, width, height, symbols, text, defaultScale, 
         const imageURL = image, hoverImageURL = hoverImage;
         const imageReplaces = { [details.colorReplace]: colorTarget };
         const hoverImageReplaces = { [details.colorReplace]: hoverColorTarget };
-        // both properties are recolored from the same URL, so they share one entry in its download queue (keyed by the icon)
+        // both properties are recolored from the same URL, so they share one entry in its download queue
+        const iconSubscriber = subscriber === undefined ? undefined : `${subscriber}:symbol${symbolIndex}`;
         const applyImages = _=>{
-          icon.style.setProperty('--image', `url("${getSVG(imageURL, imageReplaces, applyImages, icon)}")`);
-          icon.style.setProperty('--hoverImage', `url("${getSVG(hoverImageURL, hoverImageReplaces, applyImages, icon)}")`);
+          icon.style.setProperty('--image', `url("${getSVG(imageURL, imageReplaces, applyImages, iconSubscriber)}")`);
+          icon.style.setProperty('--hoverImage', `url("${getSVG(hoverImageURL, hoverImageReplaces, applyImages, iconSubscriber)}")`);
         };
-        image = getSVG(imageURL, imageReplaces, applyImages, icon);
-        hoverImage = getSVG(hoverImageURL, hoverImageReplaces, applyImages, icon);
+        image = getSVG(imageURL, imageReplaces, applyImages, iconSubscriber);
+        hoverImage = getSVG(hoverImageURL, hoverImageReplaces, applyImages, iconSubscriber);
       }
       icon.style.setProperty('--image', `url("${image}")`);
       icon.style.setProperty('--hoverImage', `url("${hoverImage}")`);
--- a/client/js/symbols.js
+++ b/client/js/symbols.js
@@ -418,8 +418,16 @@ function generateSymbolsDiv(target, width, height, symbols, text, defaultScale, 
             hoverColorTarget = `${hoverColorTarget}\" stroke=\"${symbol.hoverStrokeColor[0]}\" stroke-width=\"${symbol.hoverStrokeWidth[0]}`;
           }
         }
-        image = getSVG(image, { [details.colorReplace]: colorTarget }, i=>icon.style.setProperty('--image', `url("${i}")`));
-        hoverImage = getSVG(hoverImage, { [details.colorReplace]: hoverColorTarget }, i=>icon.style.setProperty('--hoverImage', `url("${i}")`));
+        const imageURL = image, hoverImageURL = hoverImage;
+        const imageReplaces = { [details.colorReplace]: colorTarget };
+        const hoverImageReplaces = { [details.colorReplace]: hoverColorTarget };
+        // both properties are recolored from the same URL, so they share one entry in its download queue (keyed by the icon)
+        const applyImages = _=>{
+          icon.style.setProperty('--image', `url("${getSVG(imageURL, imageReplaces, applyImages, icon)}")`);
+          icon.style.setProperty('--hoverImage', `url("${getSVG(hoverImageURL, hoverImageReplaces, applyImages, icon)}")`);
+        };
+        image = getSVG(imageURL, imageReplaces, applyImages, icon);
+        hoverImage = getSVG(hoverImageURL, hoverImageReplaces, applyImages, icon);
       }
       icon.style.setProperty('--image', `url("${image}")`);
       icon.style.setProperty('--hoverImage', `url("${hoverImage}")`);
--- a/client/js/widgets/card.js
+++ b/client/js/widgets/card.js
@@ -113,6 +113,8 @@ class Card extends Widget {
           const useIframe = original.type == 'html' && legacyMode('useIframeForHtmlCards');
           const objectDiv = document.createElement(useIframe ? 'iframe' : 'div');
           objectDiv.classList.add('cardFaceObject');
+          // identifies this object in the download queue of an SVG across renders (the icons below are re-created every time)
+          const objectSubscriber = `${this.id}:face${faceTemplates.indexOf(face)}:object${face.objects.indexOf(original)}`;
 
           const setValue = _=>{
             const usedProperties = new Set();
@@ -142,7 +144,7 @@ class Card extends Widget {
                     replaces[key] = this.get(replaces[key]);
                   const svgResult = getSVG(object.value, replaces, _=>{
                     objectDiv.style.backgroundImage = `url("${getSVG(object.value, replaces)}")`;
-                  }, objectDiv);
+                  }, objectSubscriber);
                   objectDiv.style.backgroundImage = `url("${svgResult}")`;
                 } else {
                   objectDiv.style.backgroundImage = mapAssetURLs(`url("${object.value}")`);
@@ -153,7 +155,7 @@ class Card extends Widget {
               if(object.value) {
                 if($('.symbolOuterWrapper', objectDiv))
                   $('.symbolOuterWrapper', objectDiv).remove();
-                generateSymbolsDiv(objectDiv, object.size || object.width, object.size || object.height, typeof object.value == 'object' ? object.value : Object.assign({ name:object.value }, object, { rotation: 0 }), object.text || '', 1, object.color);
+                generateSymbolsDiv(objectDiv, object.size || object.width, object.size || object.height, typeof object.value == 'object' ? object.value : Object.assign({ name:object.value }, object, { rotation: 0 }), object.text || '', 1, object.color, undefined, 1, objectSubscriber);
               }
             } else if (object.type == 'html') {
               const content = String(object.value).replaceAll(/\$\{PROPERTY ([A-Za-z0-9_-]+)\}/g, (m, n) => {
--- a/client/js/widgets/card.js
+++ b/client/js/widgets/card.js
@@ -142,7 +142,7 @@ class Card extends Widget {
                     replaces[key] = this.get(replaces[key]);
                   const svgResult = getSVG(object.value, replaces, _=>{
                     objectDiv.style.backgroundImage = `url("${getSVG(object.value, replaces)}")`;
-                  });
+                  }, objectDiv);
                   objectDiv.style.backgroundImage = `url("${svgResult}")`;
                 } else {
                   objectDiv.style.backgroundImage = mapAssetURLs(`url("${object.value}")`);
--- a/client/js/widgets/dice.js
+++ b/client/js/widgets/dice.js
@@ -182,14 +182,14 @@ class Dice extends Widget {
   }
 
   createChildNodes() {
-    const applySVGreplaces = (faceDOM, faceDefinition, image, svgReplaces) => {
+    const applySVGreplaces = (faceDOM, faceDefinition, image, svgReplaces, subscriber) => {
       let imageResult = mapAssetURLs(image);
 
       if(typeof svgReplaces == 'object' && svgReplaces !== null) {
         const replaces = {};
         for(const key in svgReplaces)
           replaces[key] = this.getFaceProperty(faceDefinition, svgReplaces[key]);
-        imageResult = getSVG(image, replaces, _=>faceDOM.style.backgroundImage = `url("${getSVG(image, replaces)}")`, faceDOM);
+        imageResult = getSVG(image, replaces, _=>faceDOM.style.backgroundImage = `url("${getSVG(image, replaces)}")`, subscriber);
       }
 
       faceDOM.style.backgroundImage = `url("${imageResult}")`;
@@ -228,7 +228,7 @@ class Dice extends Widget {
         face.className = 'dicePip';
       } else if(icon != undefined) {
         this.facesElement.appendChild(face);
-        generateSymbolsDiv(face, this.get('width'), this.get('height'), icon, text, 0.9, this.getFaceProperty(content, 'pipColor'), this.getFaceProperty(content, 'pipColor'));
+        generateSymbolsDiv(face, this.get('width'), this.get('height'), icon, text, 0.9, this.getFaceProperty(content, 'pipColor'), this.getFaceProperty(content, 'pipColor'), 1, `${this.id}:face${i}`);
       } else if(text != undefined) {
         face.textContent = text;
       } else if(content.value != undefined && content.image == undefined) {
@@ -238,7 +238,7 @@ class Dice extends Widget {
       }
       const image = this.getFaceProperty(content, 'image');
       if(image)
-        applySVGreplaces(face, content, image, this.getFaceProperty(content, 'svgReplaces'));
+        applySVGreplaces(face, content, image, this.getFaceProperty(content, 'svgReplaces'), `${this.id}:face${i}`);
 
       face.appendChild(polygonBorder.cloneNode(true));
       face.appendChild(polygonBorder.cloneNode(true));
--- a/client/js/widgets/dice.js
+++ b/client/js/widgets/dice.js
@@ -189,7 +189,7 @@ class Dice extends Widget {
         const replaces = {};
         for(const key in svgReplaces)
           replaces[key] = this.getFaceProperty(faceDefinition, svgReplaces[key]);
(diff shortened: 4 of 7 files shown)


VTT-AI-Agent (commented):
## TL;DR

The fix is correct and a clear improvement: it eliminates the permanent "blank SVG for the rest of the session" failure mode by never leaving a stale callbacks array in `svgCache`, adds the missing `r.ok` check, and covers transient failures with bounded retries. Only two minor behavioral notes below — neither blocks the change.

## Details

1. **Pending callbacks are dropped on final failure** (`client/js/main.js:497`). When all 4 attempts fail, `delete svgCache[url]` discards the queued `[callback, replaces]` pairs without invoking them, so widgets that requested the SVG during the ~7s retry window stay blank until something else re-renders them. That's still strictly better than the old behavior (blank for the whole session), and card faces re-render often, so this is fine to leave — just noting it's a known limitation, not full recovery.

2. **Permanent 404s now get retried with backoff, repeatedly** (`client/js/main.js:493-496`). A missing asset is not transient, but it now costs 4 fetches with backoff — and because the cache entry is deleted at the end, every later render of a widget using that URL restarts another 4-fetch cycle. For a deck whose card faces reference a deleted asset, that's a steady trickle of doomed requests. If you want to tighten it: only retry on network errors / 5xx (skip retries for 4xx), or cache the failure for a short period before allowing a re-attempt. Minor, since the request volume is bounded per render.

The retry count/backoff (1s/2s/4s) matches the PR description, `npm test` and the full TestCafe matrix pass, and the manual Playwright reproduction described in the PR covers the reported failure well.

I added the `beta` label: the change only touches `client/js/main.js` (not the engine paths excluded from beta deployment), is small and defensive, and can only improve behavior for players — safe and useful to test.

VTT-AI-Agent (commented):
## TL;DR

Good, well-motivated fix — a rejected `fetch` here didn't just poison `svgCache`, it reached the `unhandledrejection` handler in `tracing.js` and *terminated the session* (`preventReconnect()` + `connection.close()` + error overlay), so this is worth more than the PR description claims. One real defect though: the new `.catch()` is attached after the callback-dispatch `.then()`, so an exception thrown by a widget callback is now treated as a download failure — it silently re-fetches and can end up **deleting a successfully cached SVG** while iterating a string as if it were the callback array.

## Findings

### 1. The `catch` also swallows callback errors, corrupting the cache (client/js/main.js:485-502)

```js
const load = attempt=>fetch(mapAssetURLs(url)).then(r=>{...}).then(t=>{
  const callbacks = svgCache[url];
  svgCache[url] = t;
  for(const [ c, r ] of callbacks)
    c(getSVG(url, r, _=>{}));
}).catch(e=>{ ... });
```

The `.catch()` covers the dispatch handler, not just the fetch/parse. The callbacks are widget code (`imagewidget.js:91` / `timer.js:85` do `this.domElement.style.cssText = this.css()`, `card.js:143`, `dice.js:192`, `symbols.js:433`), so a throw there is entirely possible. Concrete sequence:

1. Download succeeds, `svgCache[url]` is set to the SVG **text**, one callback throws.
2. `catch` fires, `attempt < 3` → `setTimeout(load(1))` → the SVG is downloaded **again** even though it is already cached fine.
3. On that retry, `const callbacks = svgCache[url]` is now the SVG *string*. `for(const [c, r] of callbacks)` iterates characters, destructuring `'<'` to `c='<'`, so `c(...)` throws `TypeError: c is not a function` — every time, for all remaining attempts.
4. After attempt 3, `delete svgCache[url]` **discards the good cached SVG**, so widgets go back to returning `''` (the derived `url+JSON.stringify(replaces)` entries are still there but are only consulted after the `typeof svgCache[url] == 'string'` check, which now fails) until something triggers yet another download.

Net effect: one buggy render turns into 3 extra network requests, ~7s of timers, and a cache wipe. Secondarily, that callback exception is now swallowed into `console.error` instead of reaching the error reporter, so a real client bug becomes invisible to the maintainer.

Fix is small — put the catch on the download only and dispatch afterwards, e.g.:

```js
const load = attempt=>fetch(mapAssetURLs(url)).then(r=>{
  if(!r.ok)
    throw new Error(`HTTP status ${r.status}`);
  return r.text();
}).catch(e=>{
  if(attempt < 3)
    return new Promise(res=>setTimeout(_=>res(load(attempt+1)), 1000 * 2**attempt));
  console.error(`getSVG: failed to load ${url}`, e);
  return null;
}).then(t=>{
  if(t === null) {
    delete svgCache[url]; // allow a later call to retry the download
    return;
  }
  const callbacks = svgCache[url];
  svgCache[url] = t;
  for(const [ c, r ] of callbacks)
    c(getSVG(url, r, _=>{}));
});
```

Also guard the dispatch against `svgCache[url]` no longer being an array (`if(!Array.isArray(callbacks)) return;`) so a re-entrant path can never hit the string-iteration case.

### 2. Permanent failures (404/403) now cause repeated request storms (client/js/main.js:495-500)

Every failure path ends in `delete svgCache[url]`, and every retry chain is 4 requests. For a genuinely missing asset — a game referencing a deleted `/assets/...` SVG, which is a common real-world case — this means: any re-render calls `getSVG`, finds no cache entry, and starts a fresh 4-request/7-second chain. Since image/timer widgets re-render on state changes, a busy room with one broken SVG URL will hammer the server continuously, whereas before the fix the (bad) error body was cached once and never re-requested.

Suggestion: only retry on network-level rejections and 5xx; on a 4xx, skip the backoff entirely and negative-cache it (e.g. `svgCache[url] = ''`, which makes `getSVG` return a harmless `data:image/svg+xml,` and stops all further traffic for the session). That keeps the intended recovery from transient hiccups without turning a 404 into a request amplifier.

### 3. Silently dropped callbacks are correct here, but worth a comment (client/js/main.js:498)

On final give-up the queued `[callback, replaces]` entries are discarded without being invoked. That is the right call — invoking them would re-enter `getSVG`, which re-triggers a download, giving an infinite retry loop — but it isn't obvious, and the existing comment only explains the `delete`. One extra line ("pending callbacks are dropped on purpose: calling them would re-enter getSVG and restart the download") would save the next reader the analysis.

### 4. No automated test (tests/client/)

`tests/client/` already runs client modules under jsdom (`error-report.test.js` unit-tests `tracing.js` directly), so the retry/backoff/`r.ok` logic is testable with a mocked `global.fetch` and fake timers — but `getSVG` isn't exported from `main.js`. Given the failure mode this PR fixes was expensive (session termination) and the fix has non-trivial control flow, exporting `getSVG` and adding a jest test for "first two attempts fail, third succeeds" and "all attempts fail → cache entry cleared" would be worth it; manual Playwright verification isn't repeatable in CI.

No new widget property or routine function is exposed, so no editor-UI or `library/tutorials/` work is needed here.

I added no label: `beta` was already present and this change is a pure client-side robustness fix in `main.js` (not one of the engine files excluded from beta auto-deployment), so it stays.

VTT-AI-Agent:
Thanks for the review — both findings are addressed in 0053a50b, plus the suggested comment and jest tests.

**1. The `catch` swallowed callback errors (confirmed, and worse than described)**

`getSVG` now runs the download and the dispatch as two separate promise handlers, so an exception from a widget callback can no longer be mistaken for a download failure:

```js
const download = attempt=>fetch(mapAssetURLs(url)).then(r=>{
  if(!r.ok)
    throw Object.assign(new Error(`HTTP status ${r.status}`), { httpStatus: r.status });
  return r.text();
}).catch(e=>{
  // only network errors and server errors are worth retrying, a 4xx won't become valid by asking again
  if(attempt < 3 && !(e.httpStatus >= 400 && e.httpStatus < 500))
    return new Promise(resolve=>setTimeout(_=>resolve(download(attempt+1)), 1000 * 2**attempt));
  throw e;
});
download(0).then(t=>{
  const callbacks = svgCache[url];
  svgCache[url] = t;
  if(Array.isArray(callbacks))
    for(const [ c, r ] of callbacks)
      c(getSVG(url, r, _=>{}));
}, e=>{ ... });
```

The dispatch is the `onFulfilled` argument of the same `.then()` that handles the rejection, so a throw in it does not re-enter the failure path — it stays unhandled and reaches the error reporter, as you asked for. The `Array.isArray(callbacks)` guard is in place, and the give-up branch now carries the comment explaining that pending callbacks are dropped on purpose.

I reproduced the exact sequence you described in a browser (Playwright, image widget with `svgReplaces`, its `cssText` setter made to throw once). On the old code:

```
requests: {"vtest-throw.svg": 4}
console error: getSVG: failed to load /assets/vtest-throw.svg TypeError: c is not a function
```

With the fix: 1 request, the cache entry survives, and the callback error surfaces as a real page error instead of a `console.error`.

**2. Permanent 4xx failures caused retry storms**

Retries are now limited to network-level rejections and 5xx. A 4xx skips the backoff entirely and is negative-cached as `svgCache[url] = ''`, which makes later `getSVG` calls return the harmless `data:image/svg+xml,` without any further traffic. Same browser check, 404 URL: 4 requests before, 1 request now — and no new request on any later re-render.

**3./4. Comment and tests**

Added the comment on the dropped callbacks, and `getSVG` is now exported so `tests/client/getsvg.test.js` can cover it under jsdom with a mocked `fetch` and fake timers:

- two network failures then success → 3 requests, callback gets the recoloured SVG, later calls hit the cache
- all attempts fail → 4 requests, cache entry cleared so a later call retries and succeeds
- 404 → 1 request, no retry, negative-cached (no further requests)
- 500 → retried

`npm test` passes.

VTT-AI-Agent (commented):
## TL;DR

Good, well-targeted fix for a real session-killing bug, and the unit tests are the right ones. My main concern is the new `delete svgCache[url]` policy: because widgets re-render often, a *sustained* network/5xx outage now turns every re-render of an affected widget into a fresh 4-request retry chain instead of the single in-flight fetch it used to be — worth a small guard.

## Findings

### 1. Request amplification during a sustained outage (`client/js/main.js:505`)

`delete svgCache[url]` is the right call for recoverability, but it also removes the de-duplication that the pending-array previously provided for free. Before this PR, once the array existed it stayed forever, so a widget re-rendering 50 times only ever caused one `fetch`. Now, each re-render after the chain gives up starts a *new* chain of up to 4 requests (1s + 2s + 4s backoff).

That matters because the SVG path is on the hot render path: `imagewidget.js:91` and `timer.js:85` call `getSVG` from `css()`, and `card.js:143` / `dice.js:192` call it while building faces — all of which run again on essentially any state delta. So a board with, say, 20 recoloured SVG widgets during a 60-second connectivity problem generates a repeating 80-request burst rather than 20 stalled requests. Combined with the missing jitter (every widget shares the exact 1/2/4s schedule), the retries all land at the same instants.

Cheap mitigations, in rough order of value:
- remember the time of the last give-up per URL and don't start a new chain within a few seconds of it (a small `svgFailedAt[url]` map), or
- add jitter to the backoff (`1000 * 2**attempt * (0.5 + rand()/2)` — `rand()` is already the client-wide helper).

### 2. The waiting widgets are never notified on the transient-failure path (`client/js/main.js:503-508`)

After the four attempts fail, the queued callbacks are dropped. Every widget that requested the SVG got `''` synchronously and now has no path back to a rendered image until something unrelated re-renders it. In the exact scenario from the bug report — a card building its faces at room load during a hiccup — the player is still left with a blank card face; the win is only that other widgets and the session aren't poisoned. Dropping the callbacks is defensible (calling them re-enters `getSVG` and would restart the download, potentially forever), but it's worth deciding explicitly whether one delayed re-dispatch — e.g. re-invoke the queued callbacks once, after the entry has been deleted, so a single fresh chain runs if the network has recovered — is worth the extra complexity. If not, at least the limitation deserves a line in the comment.

### 3. The dropped-callbacks comment is inaccurate for the 4xx branch (`client/js/main.js:507`)

`// the pending callbacks are dropped on purpose: calling them would re-enter getSVG and restart the download` sits below *both* branches, but it is only true of the `delete` branch. On the 4xx branch `svgCache[url]` is a valid string `''`, so `getSVG` would return `data:image/svg+xml,` straight from the cache with no download at all — the callbacks could safely be dispatched there exactly like the success path does. Practically the rendered result is the same (`url("")` vs `url("data:image/svg+xml,")`), so this is about the comment misleading the next reader more than about behaviour; either dispatch them on that branch or scope the comment to the transient case.

### 4. `Array.isArray(callbacks)` in the success handler looks unreachable (`client/js/main.js:498`)

While a download is in flight, nothing else writes `svgCache[url]`: the two failure branches are on the rejection side of the same `.then(onSuccess, onError)` and are mutually exclusive with the success handler, and a concurrent `getSVG` for the same URL takes the `push` path because the array is truthy. If the guard is defending against a case I've missed, a one-line comment naming it would help; otherwise it silently swallows the entire dispatch loop, which is exactly the failure mode this PR is fixing.

### 5. Test coverage: the two subtlest decisions aren't in jest

`tests/client/getsvg.test.js` is a good suite (I ran it and the full jest run locally — all green). Two gaps, both a few lines with the harness that's already there:

- **A widget callback throwing must not be treated as a download failure.** This is the trickiest part of the diff — the two-argument `.then(onSuccess, onError)` rather than `.then().catch()` — and it's the one thing that, if someone "tidies" it into `.catch()`, silently reintroduces a worse version of the original bug (re-download, iterating the cached string as if it were the callback array, then deleting the good entry). It's currently only Playwright-verified. A test that makes the first queued callback throw and then asserts `fetch` was called once and the cache still returns the SVG would lock it in.
- **A second `getSVG` for the same URL while retries are in flight.** The queueing across the retry window is new behaviour; asserting that a caller who arrives between attempt 1 and attempt 3 also gets its callback would cover it.

### Not applicable here

Pure bug fix — no new widget property or routine function, so no editor-UI or validator/`jeCommands` obligations, and no tutorial is required. The `export` keyword added to `getSVG` is safe: the client bundle is inlined as `<script type="module">` (`client/room.html:605`) and `main.js` already exports `shuffleWidgets`; I loaded a room in Chromium against a local server and it comes up with no console or page errors.

Adding the `beta` label was unnecessary because it is already present, and it should stay: `client/js/main.js` is not one of the engine files excluded from beta auto-deployment, and this is a strictly-more-resilient change to SVG loading that can't corrupt game state.

VTT-AI-Agent:
Thanks — all five points are addressed in fed2af46. The core change: instead of throwing the pending queue away on failure, `getSVG` now **keeps** it and only throttles when a new download attempt may start.

```js
const callbacks = svgCache[url] || (svgCache[url] = []);
if(callbacks.length < 1000)
  callbacks.push([ callback, replaces ]);
if((svgRetryTime[url] || 0) <= Date.now())
  downloadSVG(url, callbacks);
return '';
```

`svgRetryTime[url]` is `Infinity` while a chain is in flight and `Date.now() + 5000` after it gave up, so the entry is the de-dup *and* the queue again, and `delete svgCache[url]` is gone entirely.

### 1. Request amplification during a sustained outage — fixed

The pending array is no longer deleted, so it still de-dups, and after a chain gives up the URL is off limits for 5s no matter how many widgets ask. The backoff also got jitter (`1000 * 2**attempt * (0.5 + Math.random()/2)`) so widgets sharing a schedule no longer fire in lockstep.

Measured in Chromium (image widget with `svgReplaces`, asset aborted via `page.route`, a state delta forcing a re-render every 2s):

| | requests during a 60s outage |
| --- | --- |
| before (0053a50b) | 31 |
| after | 20 |

(Both are lower than one chain per re-render because a new chain could only start once the previous one had given up — the amplification was a continuous ~4-requests-per-7s trickle rather than a burst per render. The cooldown stretches that to ~4 per 12s.)

### 2. Waiting widgets are never notified — fixed

They are now. The queued `[callback, replaces]` pairs survive the failure and are invoked by whichever later attempt succeeds, so a widget that rendered once during the outage and never renders again still gets its image. Verified in the browser with two image widgets sharing one URL, where only widget A re-renders after the network comes back:

```
=== before (0053a50b):
A (re-rendered):       url("data:image/svg+xml,%3Csvg%20xmlns%3D%22h
B (never re-rendered): url("")
=== after:
A (re-rendered):       url("data:image/svg+xml,%3Csvg%20xmlns%3D%22h
B (never re-rendered): url("data:image/svg+xml,%3Csvg%20xmlns%3D%22h
```

The queue is capped at 1000 entries so a long outage in a busy room can't grow it without bounds.

### 3. The dropped-callbacks comment was wrong for the 4xx branch — fixed

The 4xx branch now goes through the same `useSVG(url, callbacks, '')` as the success path, i.e. it negative-caches the empty SVG *and* dispatches the queued callbacks, exactly as you pointed out it safely could. The inaccurate comment is gone; nothing is dropped on either branch any more.

### 4. `Array.isArray(callbacks)` looked unreachable — removed

Agreed, and rather than document a case I couldn't name, the dispatch now closes over the array it was handed (`downloadSVG(url, callbacks)` → `useSVG(url, callbacks, t)`) instead of re-reading `svgCache[url]`. Iterating a cached string is impossible by construction, so the guard isn't needed.

### 5. Test coverage — both gaps covered

`tests/client/getsvg.test.js` is now 7 tests. New:

- **"does not treat an exception in a widget callback as a download failure"** — the exact regression you named. I checked it bites: rewriting `.then(onSuccess, onError)` into `.then(onSuccess).catch(onError)` makes it fail. It asserts the exception is *not* swallowed (no `console.error` from the download handler) and reaches the error reporter as an unhandled rejection. jest-circus fails any test that produces an unhandled rejection, which is precisely the behaviour under test, so the test temporarily records the promises `getSVG` creates and inspects their rejections afterwards (`trackPromises()`, commented in the file).
- **"serves a caller that arrives while a retry is pending"** — a second caller between attempt 1 and the retry joins the running download and gets its callback.
- **"waits for the cooldown before retrying a failed download"** — covers finding 1; also verified it fails when the cooldown check is removed.
- **"serves the callbacks queued during an outage once a later call succeeds"** — covers finding 2 (the old "clears the cache entry" test, now asserting the stronger behaviour).

`npm test` is green (13 tests), and locally `tests/testcafe/editor.js`, `compute-1.js`, `compute-2.js` and `publiclibrary-1.js` pass against this branch.

No label change: this is still a client-side robustness fix in `client/js/main.js`, and `beta` stays as it is.

VTT-AI-Agent (commented):
## TL;DR

This is a **UI/UX review**, not a code review — a separate reviewer covers correctness. The diff touches no HTML/CSS/markup at all, so nothing in the app's layout changes; the whole user-visible surface of this PR is *what a player sees while an SVG asset is unreachable*. On that front the change is a clear improvement over the old behaviour (the session no longer dies behind the error overlay), but it leaves one gap I could reproduce in the browser: **after the network recovers, the affected widgets stay completely invisible for the rest of the session** — only a page reload brings them back. The single most impactful improvement would be to let the retry heal itself on a timer (and to render *something* visible in the meantime), instead of relying on some other widget re-rendering.

### How I tested

Ran the branch on a local server, built a room with two `image` widgets that use `svgReplaces` (so they go through `getSVG`), and drove it with Playwright while aborting/allowing the asset requests.

| | |
|---|---|
| Working baseline | [02a-svg-ok-1280.png](https://agent.virtualtabletop.io/reports/ui-review/1536649703892983839/shots/02a-svg-ok-1280.png) |
| During the outage | [02b-svg-blocked-1280.png](https://agent.virtualtabletop.io/reports/ui-review/1536649703892983839/shots/02b-svg-blocked-1280.png) |
| 20 s **after** the network came back | [03-after-network-recovered-1280.png](https://agent.virtualtabletop.io/reports/ui-review/1536649703892983839/shots/03-after-network-recovered-1280.png) |
| After a page reload | [05-after-reload-1280.png](https://agent.virtualtabletop.io/reports/ui-review/1536649703892983839/shots/05-after-reload-1280.png) |

Overlay/viewport pass (to confirm nothing else regressed — it didn't, and it couldn't, since no shared CSS is touched): room view, **States/Game shelf**, **Players**, **About**, plus the toolbar's edit and options buttons, each at **390×844 (phone portrait)**, **820×1180 (tablet)**, **1280×800 (laptop)** and **1920×1080 (wide)** — e.g. [Players @1280](https://agent.virtualtabletop.io/reports/ui-review/1536649703892983839/shots/ov-playersButton-laptop.png), [States @390](https://agent.virtualtabletop.io/reports/ui-review/1536649703892983839/shots/ov-statesButton-phone.png), [room @390](https://agent.virtualtabletop.io/reports/ui-review/1536649703892983839/shots/vp-phone-room.png), [room @1920](https://agent.virtualtabletop.io/reports/ui-review/1536649703892983839/shots/vp-wide-room.png). All render as on `main`. Overlays I did **not** reach in this pass: add-widget / JSON editor sub-dialogs, save & share dialogs, symbol picker, chat — they contain no `getSVG` call sites, so they are out of this PR's blast radius. `#optionsButton` was not clickable at 390×844 (it sits in the collapsed toolbar) — that is pre-existing, not caused by this PR.

---

## Big layout / behaviour improvements

### 1. Recovery never reaches the screen without a reload — the table stays permanently blank

This is the important one, and it's the UX half of exactly the bug the PR is fixing.

`downloadSVG` sets a 5 s cooldown on a transient failure and then relies on *"a later call"* to `getSVG` to start the next attempt (`client/js/main.js:513`). But nothing on a quiet table ever makes that later call: an `image`/`timer` widget only re-enters `getImage()` when one of its properties changes. So in my run — outage during load, network fully restored 12 s later — the two widgets were still `background-image: url("")` **20 seconds after recovery**, and stayed that way indefinitely ([03-after-network-recovered](https://agent.virtualtabletop.io/reports/ui-review/1536649703892983839/shots/03-after-network-recovered-1280.png)). Only F5 fixed it ([05-after-reload](https://agent.virtualtabletop.io/reports/ui-review/1536649703892983839/shots/05-after-reload-1280.png)).

From a player's seat that reads as "the game is broken and there is nothing I can do", which is barely better than the old error overlay — at least the overlay *told* them to reload. Since the queued callbacks are already kept alive, one self-driven retry closes the loop:

```js
else
  svgRetryTime[url] = Date.now() + 5000;
  setTimeout(_=>{ if(Array.isArray(svgCache[url]) && svgCache[url].length) downloadSVG(url, svgCache[url]); }, 5000);
```

i.e. when callbacks are still queued, schedule the next chain instead of waiting to be asked. The table then heals itself the moment the connection is back, which is the behaviour a player expects from an app that already reconnects its websocket on its own.

### 2. Give the unresolved image a visible placeholder instead of nothing

While the SVG is missing, the widget renders as `url("")` but keeps `clickable movable` — i.e. it is a fully interactive, completely invisible object on the table (confirmed in the DOM: `class="widget basic clickable movable hasImage"`, zero pixels drawn). A first-time user drags "nothing", cards land on invisible piles, and the room looks empty rather than loading. With the retries this state now lasts up to ~7 s on a flaky connection even in the happy path, so it is much more likely to be *seen* than before.

Suggestion, in the app's existing flat/subdued style: while `svgCache[url]` is still an array (in flight or retrying), let the widget draw a faint placeholder — e.g. a 1px dashed `--wcBorder`-ish outline plus the app's own broken/loading glyph — and switch to a static "missing image" mark once the URL is negative-cached as `''` after a 4xx. That distinguishes *loading*, *temporarily unavailable* and *permanently gone*, none of which are distinguishable today.

### 3. The 4xx negative cache is a permanent, silent, session-long blackout

`useSVG(url, callbacks, '')` on a 4xx is right for network hygiene, but for the user it means: a game author fixes the asset (or the admin re-uploads it) and the image *still* never appears until every player reloads. Consider pairing the negative cache with a much longer TTL rather than "forever", or at minimum making the placeholder from #2 explicit enough ("image missing") that the user knows a reload is the remedy.

## Small nitpicks

- The failure is reported only via `console.error` (`client/js/main.js:510`). Nothing in the UI changes, so for anyone who isn't a developer with devtools open, the app just silently loses artwork. Given the app has no toast mechanism, an overlay would be too heavy — but the in-widget placeholder (#2) is the app-appropriate equivalent and is worth adding in the same PR, since this PR is what converts a loud failure into a silent one.
- The log line `getSVG: failed to load <url>` is developer-facing wording in a user-facing console; if it may ever be read by a game author debugging their game, something like `Could not load image <url> (giving up after 4 attempts)` is more actionable.
- Retry timing: 1 s / 2 s / 4 s jittered means the worst case before the image gives up is ~7 s of blank widget with no feedback at all. That's a long time to stare at an empty table; it strengthens the case for a loading affordance rather than for shortening the backoff.

- **Nothing tells the newcomer what happened.** The three states a user can now be in — "still downloading", "download failed, will retry", "asset does not exist" — are visually identical (empty). A player who joined a game that references a temporarily unreachable asset sees a table that looks half-empty and will reasonably conclude the game is broken or that they joined the wrong room.
- **Invisible-but-interactive is the confusing part**, more than the missing picture. Because the widget keeps `clickable movable`, clicks and drags land on something the user can't see. Any visible mark at all (even a plain outline) makes the situation self-explanatory: "this thing is here, its picture didn't load."
- **Recovery has no discoverable path.** If #1 is not implemented, the only fix is a page reload, and nothing hints at it. If a placeholder is added, wording like *"Image unavailable — reload to retry"* would make the whole failure mode understandable without a single line of documentation.

Everything else in the interface — toolbar, Players, About, States/Game shelf, at all four viewports — is untouched and renders exactly as on `main`.

VTT-AI-Agent:
Thanks — the main finding is fixed in 94a1c508, and the log wording with it. The two placeholder-related points I'd rather not decide unilaterally; my reasoning and a concrete proposal are below.

### 1. Recovery never reaches the screen without a reload — fixed

You're right that "let a later call retry" is worthless on a quiet table. `downloadSVG` now schedules the next chain itself while callbacks are still queued, so the table heals without anyone touching it:

```js
const delay = svgRetryDelay[url] || 5000;
svgRetryDelay[url] = Math.min(2 * delay, 30000);
svgRetryTime[url] = Date.now() + delay;
if(callbacks.length)
  setTimeout(_=>{
    if(svgCache[url] === callbacks && svgRetryTime[url] != Infinity)
      downloadSVG(url, callbacks);
  }, delay);
```

The guard is what keeps this from doubling up with the existing path: if a re-rendering widget already started a chain in the meantime, `svgRetryTime[url]` is `Infinity` and the timer does nothing; if a chain succeeded, `svgCache[url]` is a string and no longer the array the timer closed over.

The cooldown now doubles after every failed chain, capped at 30s, so self-driven retrying doesn't turn into fixed-rate polling during a long outage — a completely quiet table costs **16 requests per URL in 60s of downtime** (4 attempts each at 0s, 12s, 29s, 56s) instead of the 20 the re-render scenario cost before, while now actually recovering.

Reproduced your exact scenario in Chromium (two `image` widgets with `svgReplaces`, asset aborted via `page.route` during load, network restored 9s later, **no click, no state change, no re-render of any kind afterwards**):

```
during outage:  A=url("")  B=url("")   (4 requests)
after recovery: A=url("data:image/svg+xml,%3Csvg%20xmlns%3D%
after recovery: B=url("data:image/svg+xml,%3Csvg%20xmlns%3D%
total requests: 5
```

Screenshot 8s after the network came back, without any interaction: https://agent.virtualtabletop.io/reports/pr/1525291930559316199/svg-selfheal-after-recovery.png

`tests/client/getsvg.test.js` covers it: "serves the callbacks queued during an outage once the asset is reachable again" no longer makes a second `getSVG` call — it just advances the clock and asserts the widget's callback fires. The cooldown test additionally asserts the growth (nothing new after the second chain's 5s mark, because that cooldown is 10s now). Both fail if the `setTimeout` block is removed. `Math.random` is stubbed in the suite so the jittered backoff lands on exact timer positions.

### The nitpick about wording — fixed

Both branches now say what happened and what will happen next:

* `Could not load image /assets/x.svg - trying again in 5 seconds` (with the growing delay in the text)
* `Could not load image /assets/x.svg (HTTP status 404) - it will not be requested again`

### 2. + 3. Placeholder for unresolved images, and the permanence of the 4xx cache — not in this PR, proposal below

I agree with the diagnosis (invisible-but-`clickable movable` is genuinely confusing, and the three states are indistinguishable), but I don't think this PR should decide it, for two reasons:

* **It is not a failure-path change.** `getSVG` returns `''` on the *first* render of every SVG widget, always — the download is async. So "draw a placeholder while `svgCache[url]` is an array" paints a placeholder over every recoloured SVG widget of every game on every room load, for the duration of a normal fast download. That is a visible change to healthy games, not just broken ones, and it needs a deliberate design decision about what the loading state of a table should look like.
* **The mark can't be widget-shaped generically.** `getSVG` hands back a string that ends up in `background-image` for `image`/`timer` widgets but inside a card face / dice face elsewhere; a decent placeholder therefore has to be built per call site (or as a real SVG data-URI returned in place of `''`), and its size, contrast and whether it is drawn on top of the widget's own `backgroundColor` are all things I'd be guessing at.

Concretely, what I'd propose if the maintainers want it (happy to implement it here or in a follow-up, just say which):

* return a generated data-URI instead of `''`, so it works at every call site without touching widget code: a transparent SVG with a 1px dashed outline in the app's `--wcBorder`-ish grey for "loading / retrying", and the same outline plus a centred "broken image" glyph in a muted red for "permanently unavailable" (the negative-cached 4xx);
* only draw the "loading" variant after a short grace period (~500ms) so a normal, fast download doesn't flash an outline over every board — i.e. the first render still returns `''` and the placeholder only appears if the first chain is still running;
* leave the widget interactive as it is today (making a failed-to-load widget non-clickable would change game behaviour, which this PR must not do).

On the permanent 4xx cache specifically: with the placeholder above, "image missing" becomes visible and a reload is the discoverable remedy, so I'd keep the cache permanent rather than adding a TTL — a 404 is authoritative, and asking again on a long timer costs traffic for every player of every game that references a deleted asset. If you'd rather have a TTL, it's a one-line change on top of the placeholder (drop the negative cache entry after n minutes and let the next render re-request).

`npm test` is green (13 tests) and `tests/testcafe/editor.js` and `publiclibrary-1.js` pass locally against this branch.

VTT-AI-Agent (commented):
## TL;DR

The retry/cache fix now addresses the earlier failure modes, but the new 1,000-callback cap creates a smaller version of the same “stays blank after recovery” bug: callers beyond the cap are silently dropped and never notified when the SVG finally loads. The PR is still safe and useful for beta testing because this failure is limited to artwork recovery and cannot corrupt room state or crash the server.

## Findings

### 1. The callback cap can strand later widgets after recovery (`client/js/main.js:488-494`)

Every call made while the URL is unavailable consumes another slot, including repeated renders of the same widget. Once `callbacks.length` reaches 1,000, subsequent callers still receive `''`, but their callbacks are silently discarded. When a later retry succeeds, `useSVG` only refreshes the first 1,000 subscribers; a later card face or image widget on a quiet table remains blank indefinitely because nothing makes it render again.

This is straightforward to reproduce in the existing Jest harness: leave one `fetch` pending, call `getSVG` 1,001 times with distinct spies for the same URL, then resolve the fetch. The 1,001st spy is never called even though the SVG is cached successfully. A large set of card faces sharing an SVG can hit the limit directly, while repeated `svgReplaces` updates during a longer outage can fill it with stale refreshes before a newly created widget subscribes.

Please avoid silently dropping overflow subscribers. If the queue must remain bounded, the overflow path needs to arrange a broader refresh after success (or the API needs a stable subscriber identity so repeated renders can be deduplicated); otherwise the self-scheduled retry does not guarantee recovery for every widget that received the empty result. A regression test with more callers than the chosen bound would lock that behavior in.

No new widget property or routine operation is exposed, so editor controls, validator/`jeCommands`, and a `library/tutorials/` tutorial are not applicable to this PR.

The `beta` label was already present, and it should remain because this client-side `main.js` robustness change is safe for player testing and does not touch an engine-excluded path.

VTT-AI-Agent:
Fixed in 2fa07e20 — you're right, and the cap was the wrong tool for what it was guarding against.

### The cap is gone; callers identify themselves instead

The cap existed because widgets ask for the same URL on *every* render, so a long outage would have collected one queue entry per render. But that growth comes from **duplicate** subscriptions, not from many distinct widgets — so the fix is to collapse the duplicates rather than to truncate the queue. `getSVG` takes an optional fourth argument, the subscriber (in practice the DOM node the callback updates), and a caller that asks again replaces its own queued entry:

```js
const queued = subscriber !== undefined ? callbacks.findIndex(c=>c[2] === subscriber) : -1;
if(queued != -1)
  callbacks[queued] = [ callback, replaces, subscriber ];
else
  callbacks.push([ callback, replaces, subscriber ]);
```

Nothing is dropped any more: every entry in the queue is a widget that shows no image until the download succeeds, so there is no length at which discarding one is correct. Your reproduction is now a test — `notifies every waiting caller, no matter how many are queued` queues 1500 distinct subscribers behind one pending download and asserts all 1500 callbacks fire (it fails with the old `callbacks.length < 1000`).

All five call sites pass a subscriber, so the repeated-render case that motivated the cap no longer grows the queue at all:

| call site | subscriber |
|---|---|
| `imagewidget.js` / `timer.js` | `this.domElement` |
| `card.js` (face object) | `objectDiv` |
| `dice.js` (face) | `faceDOM` |
| `symbols.js` (icon) | `icon` |

`symbols.js` needed one restructure: it called `getSVG` twice for the same URL, once for `--image` and once for `--hoverImage`, so with a per-icon key the two would have collided. Both properties are recolored from the same asset, so they now share a single queue entry whose callback refreshes both. Verified against `main` with an icon that has distinct `color`/`hoverColor` — identical result, both before and after an outage (`--image` `fill="#ff0000"`, `--hoverImage` `fill="#0000ff"`, healed from one queue entry).

What is left unbounded is queue growth from *DOM churn* during an outage (re-created card faces or regenerated symbol icons subscribe as new nodes). I left that alone deliberately: it is one small array entry per newly created SVG-consuming node, the whole queue is discarded the moment the URL resolves either way, and any bound on it would have to drop a live subscriber again — which is the bug you reported.

### Verification

`tests/client/getsvg.test.js` is 9 tests now; the second new one (`keeps only the latest callback of a widget that renders again while the URL is down`) locks in the dedup. `npm test` is green (15 tests).

In the browser (Playwright, local server) I built a room that exercises every `getSVG` call site — an `image` widget with `svgReplaces`, a `dice` face with `svgReplaces`, a `card` face object with `svgReplaces`, and a button icon from game-icons (the `colorReplace` path in `symbols.js`) — and aborted the shared asset:

```
--- happy path
    image       url("data:image/svg+xml,%3Csvg%20xmlns%3   requests 2
    dieFace     url("data:image/svg+xml,%3Csvg%20xmlns%3
    cardObject  url("data:image/svg+xml,%3Csvg%20xmlns%3
    iconImage   url("data:image/svg+xml,%3Csvg%20xmlns%3
    iconHover   url("data:image/svg+xml,%3Csvg%20xmlns%3
--- during outage (asset aborted, then 40 re-renders of the image widget)
    image / dieFace / cardObject / iconImage / iconHover = url("")
--- after recovery, no interaction at all
    image       url("data:image/svg+xml,%3Csvg%20xmlns%3
    dieFace     url("data:image/svg+xml,%3Csvg%20xmlns%3
    cardObject  url("data:image/svg+xml,%3Csvg%20xmlns%3
    iconImage   url("data:image/svg+xml,%3Csvg%20xmlns%3
    iconHover   url("data:image/svg+xml,%3Csvg%20xmlns%3
    requests    8 (6 while down)
```

The 40 re-renders while the URL was down added **zero** requests and zero queue entries.

`tests/testcafe/editor.js`, `publiclibrary-1.js`, `publiclibrary-2.js` and `compute-1.js` pass locally.

### Labels

Agreed on `beta`, unchanged. I added `engine` because the diff now also touches `client/js/widgets/`.

VTT-AI-Agent (commented):
## TL;DR

The fix is well-targeted and the failure mode it describes is real, but the "one queue entry per subscriber" mechanism — the reason the whole `subscriber` parameter was threaded through five call sites — only actually works for `imagewidget`/`timer`/`card`. For **dice faces and every icon** the subscriber is a DOM node that is freshly created on each re-render, so the queue still grows one entry per render during an outage (and retains detached nodes). Second concern: the retry chain never gives up, so a permanently unreachable URL is polled for the whole session.

## Findings

### 1. The per-subscriber de-duplication does not work for dice faces or icons

`getSVG` matches a subscriber by identity (`callbacks.findIndex(c=>c[2] === subscriber)`, client/js/main.js:489). That only de-duplicates if the same object is passed on every render.

- `client/js/symbols.js:434` passes `icon`, which is created two lines earlier by `const icon = div(wrapper, details.class, details.text);` (client/js/symbols.js:402) — a **new element on every call** to `generateSymbolsDiv`. And `generateSymbolsDiv` is re-run on every icon update: `imagewidget.js:94-99` removes `this.symbolWrapper` and rebuilds it, `card.js:154-156` removes `.symbolOuterWrapper` and rebuilds it, `dice.js:231` rebuilds it inside `createChildNodes`.
- `client/js/widgets/dice.js:192` passes `faceDOM`, which is created inside `createChildNodes()`; `applyDeltaToDOM` calls `createChildNodes()` on a delta of `faces`, `image`, `svgReplaces`, `width`, `height`, any `svgReplaces` source property, … (dice.js:117-140) — again a new node per re-render.

Concrete failure: an outage while a game animates a dice widget whose `svgReplaces` reference a changing property (or a button whose icon color is bound to a property). Each delta re-renders, `findIndex` never matches, and the queue collects one entry per render for the whole outage — the exact scenario the subscriber machinery was added to prevent — while also holding every detached `faceDOM`/`icon` node alive until the download finally succeeds (and then invoking a callback on each of them, writing styles onto nodes that are no longer in the document).

The stable identities are available at both sites: for dice, key on the widget plus the face index (e.g. a per-widget array of key objects, or `this.domElement` + `i`); for symbols, `target` is stable across re-renders (`this.domElement` / `objectDiv`), so `target` combined with the symbol index would work. Since `subscriber` is compared with `===`, a string key (`` `${widgetID}#${index}` ``) would be the least invasive option.

Only `imagewidget.js:91` and `timer.js:85` (`this.domElement`) and `card.js:145` (`objectDiv`, stable across `setValue` calls) get the intended behaviour today.

### 2. The retry chain never terminates

`callbacks` is never emptied, so `if(callbacks.length)` at main.js:522 is always true and every failed chain schedules the next one — forever, at a 30 s cap, 4 requests per chain (~8 requests/minute per client) for the rest of the session, even after every widget that wanted the URL is gone.

This is not only the "asset host is down" case. A game that points `image`/`object.value` at a **cross-origin URL without CORS headers** makes `fetch` reject with a `TypeError` — indistinguishable here from a transient network error, since it has no `httpStatus` — so that URL is now polled every 30 s forever by every client in the room. Before this PR it cost exactly one failed request (plus the unhandled rejection this PR is fixing).

Suggest a stop condition: cap the number of retry chains (or the total retry window, e.g. give up after a few minutes and negative-cache), and/or prune queue entries whose subscriber is a detached DOM node (`subscriber instanceof Node && !subscriber.isConnected`) before the `callbacks.length` check — that would fix the leak in finding 1 as a side effect.

### 3. One throwing widget callback leaves every other waiting widget blank

`useSVG` (main.js:539-543) iterates the callbacks without isolation, so the first callback that throws aborts the loop and the remaining queued widgets never get their image — they stay blank until they happen to re-render, and with the queue now potentially holding hundreds of entries this is much more consequential than it was with the old code. The PR deliberately lets that exception reach the error reporter (and tests for it), which is right, but both goals are achievable: wrap each `c(...)` in `try`/`catch` and re-throw the caught error asynchronously (`setTimeout(_=>{ throw e })`) so the reporter still sees it while the rest of the queue is served.

### 4. The 4xx negative cache is permanent for the session

main.js:517-519 caches any 4xx as an empty SVG for the lifetime of the page. 404 and 410 genuinely won't fix themselves, but a 403 from a CDN/auth hiccup, or a 404 for an asset that is being uploaded to the room right now, is transient — and the image is then permanently blank until reload with no way back. Consider limiting the permanent negative cache to 404/410 and letting other 4xx go through the normal cooldown path.

### 5. `trackPromises` can leak its `Promise.prototype.then` patch across tests

tests/client/getsvg.test.js:22-40 replaces `Promise.prototype.then` globally and only restores it in the returned function. If any `expect` between `trackPromises()` and `collectRejections()` fails, the patch is never removed and every subsequent test in the run accumulates promises into a dead array. A `try`/`finally` (or restoring in `afterEach`) makes the suite fail cleanly instead of confusingly.

### Notes

- No new widget property or routine function is exposed here, so no editor-UI or tutorial work is expected for this PR.
- The retry/backoff logic itself reads correctly to me: the `Infinity` marker plus the `svgCache[url] === callbacks && svgRetryTime[url] != Infinity` guard (main.js:523-525) does prevent two concurrent chains for a URL, and splitting the download handler from the callback dispatch does keep a widget exception from being mistaken for a download failure. The `r.ok` check is a clear improvement over caching an error page as SVG.

The `beta` label was already on this PR, so I left it as is (I never remove it) — note that these changes are genuinely engine-level, since `getSVG` is on the render path of every existing game that uses `svgReplaces` or icons, so it is excluded from the beta build by design regardless.

VTT-AI-Agent:
Thanks — all five findings are addressed in 1a32c8e3.

### 1. The per-subscriber de-duplication didn't work for dice faces or icons — fixed

Confirmed, and your analysis was exactly right: `faceDOM` and `icon` are re-created on every render, so `findIndex(c=>c[2] === subscriber)` never matched. Since `subscriber` is only ever compared with `===`, the call sites now pass a stable string key naming the widget and the position within it instead of a DOM node:

| call site | subscriber |
| --- | --- |
| `imagewidget.js` / `timer.js` (image) | `this.domElement` (stable, unchanged) |
| `dice.js` (face image) | `` `${this.id}:face${i}` `` |
| `card.js` (face object, image *and* icon) | `` `${this.id}:face${f}:object${o}` `` |
| `symbols.js` (icon) | `` `${subscriber}:symbol${n}` `` — the owner passes its key into `generateSymbolsDiv()` |

`generateSymbolsDiv()` got an optional last parameter for that owner key, passed by the three callers (`imagewidget` `` `${this.id}:icon` ``, `dice` `` `${this.id}:face${i}` ``, `card` the object key above). The icon element itself can't identify a caller, so the widget has to.

Measured in Chromium against a local server (image widget with `svgReplaces`, a dice face with `svgReplaces`, and a button icon through the `colorReplace` path in `symbols.js`, all sharing one asset aborted via `page.route`; two buttons alternate the property the three widgets derive their colors from, so each click really re-creates the dice `faceDOM` and the `icon`). Queue length read from `svgCache` after 30 such re-renders during the outage:

```
                       queue entries after 30 re-renders
before (2fa07e20)      32 and 31 (one per render, each holding a detached node)
after  (1a32c8e3)       2 and  1
```

(Two keys because `symbols.js` looks the asset up under its `mapAssetURLs()`-mapped path.) Everything still heals after the network comes back, with no interaction.

### 2. The retry chain never terminated — fixed

The cooldown keeps doubling (5s, 10s, 20s, …) but the self-scheduled chain now stops once it would exceed five minutes, so an unreachable URL costs 7 chains ≈ 28 requests in the first ~5 minutes and then nothing. Your cross-origin-without-CORS example is the case I had in mind for the comment: it rejects as a `TypeError` with no `httpStatus`, so it can't be told apart from a hiccup, but it now stops being polled.

The queued callbacks are kept, and `svgRetryTime`/`svgRetryDelay` are kept too, so a widget that renders again later can still start one more attempt — at 320s, then 640s, … , i.e. it decays instead of settling into a fixed rate.

I also took the pruning suggestion: before deciding whether anyone is still waiting, entries whose subscriber is a DOM node that is no longer in the document are dropped. So a URL whose widgets have all been deleted stops retrying with the very next chain, and no detached node is retained. New test `stops retrying for a widget that isn't on the table any more`.

### 3. One throwing callback left the rest blank — fixed

`useSVG` wraps each callback in `try`/`catch` and re-throws asynchronously with `Promise.reject(e)`, which lands in the `unhandledrejection` handler in `tracing.js` — the same reporting path as before, so the existing "not mistaken for a download failure" test still holds, while the remaining queued widgets are served. New test `serves the other waiting widgets even if one of their callbacks throws`.

### 4. The 4xx negative cache was permanent — fixed

Only 404 and 410 are treated as authoritative now. Everything else (403, 408, 429, …) goes through the normal retry chain and cooldown, so a CDN or auth hiccup, or an asset that is being uploaded right now, recovers. New test `retries a 403 instead of caching it as permanently missing`.

### 5. `trackPromises` could leak its patch — fixed

The restore function is stored and also called from `afterEach`, so a failing expectation in between can't leave `Promise.prototype.then` (and now `Promise.reject`) patched for the following tests.

### Verification

`npm test` is green (19 tests, 13 of them in `tests/client/getsvg.test.js`). I checked that each of the four new tests fails without its fix. Locally `tests/testcafe/editor.js`, `publiclibrary-1.js`, `publiclibrary-2.js` and `compute-1.js` pass, and in the browser all five `getSVG` consumers (image widget, dice face, card face object, icon `--image` and `--hoverImage`) render both on the happy path and after an outage.

No label change — `beta` and `engine` stay as they are.

VTT-AI-Agent (commented):
## TL;DR

- **This PR now conflicts with `main` and its core fix has already been merged in a different form.** PR #3121 (merged 2026‑08‑19) rewrote `getSVG` around a rejection-safe `fetchSVG`, an `unreadableCache` with a 30 s retry, and a fallback that shows the raw image URL while the file can't be read. The retry model in this PR (blank widget + retry chains) is built on the old `svgCache[url]`-array design; a rebase that keeps it would **undo #3121's fixes** (CORS-blocked external images and PNGs with `svgReplaces` would go blank again). It needs a rework on top of `main`, not a merge, or it should be closed as superseded — I've asked the maintainers which.
- What is genuinely still missing on `main` and worth carrying over: the in-chain backoff, the self-scheduled retry on a quiet table, and the per-callback isolation in `useSVG`. Two gaps in that logic reproduce under jest (findings 2 and 3).

## Findings

### 1. Conflicts with `main`, and re-implements what #3121 already merged (`client/js/main.js:472-573`)

`gh pr view` reports the PR as `CONFLICTING`; applying the diff onto `origin/main` fails for the `client/js/main.js` and `client/js/widgets/card.js` hunks. The reason is not cosmetic: `getSVG` on `main` (`client/js/main.js:750` there) no longer looks like the base this PR was written against.

What `main` already does today (from #3121):
- `fetchSVG(url)` de-duplicates requests per mapped URL and forgets a failed request; `getSVG` consumes it with `.then(t=>t, _=>undefined)`, so **a rejected `fetch` no longer reaches `unhandledrejection` and no longer terminates the session** — the headline bug of this PR.
- `svgCache[url]` is deleted after every outcome, so the stale-callbacks-array poisoning is gone too.
- `!response.ok` is checked.
- A failure is remembered in `unreadableCache` for `UNREADABLE_RETRY_MS` (30 s) and then retried on the next render — and in the meantime `getSVG` returns `mapAssetURLs(url)`, so the widget displays the un-recolored image natively instead of nothing. That is what makes a cross-origin SVG without CORS headers, or an asset the server refuses, still *visible*.
- A file that turns out not to be an SVG (`nonSVGCache`) is displayed as it is instead of being wrapped into a broken `data:image/svg+xml` URL.
- `fetchSVG` is shared with the JSON editor (`client/js/jsonedit.js:96`, `:2589`) and the properties sidebar (`client/js/editor/sidebar/properties.js:288`), covered by `tests/client/get-svg.test.js`.

What happens if this branch is rebased and its `getSVG`/`downloadSVG` kept:
- A cross-origin SVG whose host sends no CORS headers rejects with a `TypeError`. This PR retries it for ~5 minutes and then gives up with the callbacks never served, so the widget stays at `url("")` for the session — this PR's own test *"gives up on a URL that stays unreachable"* asserts exactly that (its callback is never called). On `main` the same widget shows the image.
- A PNG with an `svgReplaces` entry is fetched as text and wrapped into `data:image/svg+xml,` again — the exact bug #3121 fixed — because this `getSVG` has neither the `<svg` sniff nor `nonSVGCache`.
- `fetchSVG` would either be bypassed (two request paths for one file, and the editor's "is this an SVG?" answer could disagree with the engine's) or inherit a 7 s backoff before answering the editor.

Suggested shape for a rework, if the maintainers want to keep the extra resilience: leave `fetchSVG`/`loadSVG` untouched, keep `main`'s return contract (raw URL while unreadable), and add the timer-driven retry as a thin layer on top: when a failure is recorded, remember the queued `[callback, replaces]` pairs and re-dispatch them from a single `setTimeout` at `UNREADABLE_RETRY_MS` (with the doubling cooldown and give-up from this PR) rather than waiting for a render. The `try`/`catch` per callback from `useSVG` and the per-subscriber de-duplication carry over unchanged. Also fold the new jest cases into `main`'s `tests/client/get-svg.test.js` — this PR's `tests/client/getsvg.test.js` would otherwise sit next to it under a near-identical name with overlapping coverage.

The self-scheduled retry is much less urgent on `main` than it was here: a widget on `main` isn't blank during the outage, it shows the un-recolored image and gets its colors back on the next render after 30 s. That's the main input to the "rework vs. close" decision.

### 2. A widget that asks during a cooldown after the queue was emptied never gets an attempt (`client/js/main.js:502`, `:537-546`)

The self-scheduled timer is only armed when `callbacks.length` is non-zero *after* pruning. Once a chain fails with an empty queue (every earlier subscriber was an `image`/`timer` widget whose DOM node is gone — deleted, or the player switched games), the URL sits in a cooldown of 5–160 s with no timer. A new widget for the same URL arriving in that window (another game using the same library asset, a widget just added) is queued, but line 502 is false, so nothing ever starts a chain for it.

Reproduced with the PR's own jest harness: first widget removed after chain 1, chain 2 fails and prunes the queue to empty, the network comes back, a second widget calls `getSVG` 1 s into the 10 s cooldown → **0 requests and its callback never fires over 10 simulated minutes.** The widget stays blank until it happens to re-render after the cooldown, which on a quiet table is exactly the "stays blank until reload" case this PR sets out to fix. The same holds after the 5‑minute give-up: a new subscriber waits ≥320 s and then only heals on a render.

Fix: arm the timer whenever `delay <= svgRetryGiveUp` and check `callbacks.length` when it fires; or, in `getSVG`, when a subscriber is added during a cooldown and no timer is pending, schedule one for `svgRetryTime[url]`.

### 3. String-keyed subscribers are never pruned, so a deleted die/card/icon keeps the URL polling (`client/js/main.js:533`)

Only subscribers with `nodeType` are dropped. Dice faces (`${id}:face${i}`), card face objects and icons subscribe with strings, so after the widget is deleted its entry keeps `callbacks.length` truthy and the chains continue until the give-up: reproduced with one deleted die → **24 further requests over the next 10 minutes**, while the closures keep the detached `faceDOM`/`objectDiv`/`icon` nodes alive. Cheap fix: prune a string key whose widget no longer exists (the key starts with the widget id, so `widgets.has(key.split(':')[0])`), or subscribe with `{ key, node }` and check `node.isConnected`.

### 4. The permanent 404/410 negative cache diverges from `main`'s deliberate choice (`client/js/main.js:525`)

#3121 explicitly decided that *no* load failure is remembered for the session ("an asset being uploaded right now", server restart), expiring it after 30 s. This PR caches 404/410 for the whole session, so a re-uploaded asset stays blank until every player reloads. Minor, but it should be decided consciously in the rework rather than inherited from the old base.

### Not applicable / verified

- No new widget property or routine function is exposed, so no editor UI, validator, `jeCommands` or tutorial work is owed, and the diff touches nothing under `library/`.
- `npm test` is green on this branch (19 tests, including the 13 in `tests/client/getsvg.test.js`).
- `symbols.js` sharing one queue entry for `--image`/`--hoverImage` is correct: both URLs are `mapAssetURLs(details.image)`, so the second `getSVG` call replaces the first entry and `applyImages` recomputes both from the cache.
- Using `Math.random()` for the jitter is right here — this isn't game state, and going through the seeded `rand()` would consume seeded numbers and change TestCafe hashes.

## Beta

The engine-directory hunks (`client/js/widgets/card.js`, `dice.js`, `imagewidget.js`, `timer.js`) are incidental: each only passes an extra subscriber argument to `getSVG`/`generateSymbolsDiv`; the behavior lives in `client/js/main.js` and `client/js/symbols.js`, which are not on the excluded list. So the PR is not engine-excluded on that basis. However, it currently conflicts with `main`, so the beta builder cannot apply it at all, and a mechanical rebase would regress #3121 — it is not beta-testable as it stands, and a maintainer may want to pull the label until it is reworked (I never remove it myself).

I did not add the `beta` label because it was already present; I left it in place, but the PR conflicts with `main` and would regress #3121 if rebased as-is, so it cannot currently be deployed to beta and should be reworked on top of `main` first.

VTT-AI-Agent:
Closing this as superseded.

The core of this PR — an SVG whose download fails leaving the widgets that use it blank for the rest of the session — is already fixed on `main` by #3121, which replaced `getSVG` with `fetchSVG` plus an `unreadableCache` that retries 30 s later on the next render and falls back to the raw image URL. Rebasing this branch would put `getSVG` back and reintroduce the two things #3121 fixed (cross-origin images blocked by CORS, and non-SVG images with `svgReplaces` rendering blank), so the only remaining gain here would be healing on a completely quiet table where nothing re-renders — not worth the regression.

Asked in the maintainer feedback thread whether to rework it on top of `main` or close it, the answer was:

> Close it as superseded by #3121

The tests in `tests/client/getsvg.test.js` and the branch itself stay available on `fix-getsvg-fetch-error` if any of the extras (timer-driven retry while callbacks are queued, growing backoff, per-subscriber callback replacement) are ever wanted on top of `fetchSVG`.

PR #3020, opened 2026-07-11
Editor properties: Automations subsection with routines and onXx properties

VTT-AI-Agent:
## What

This continues the properties-UI work in #2884 (PR is based on and targets the `expand-widget-edito` branch, as requested) by adding an **Automations** subsection that is available for **every widget type** — it renders between the basic section and the type-specific/fallback sections of every editor, following the roadmap item:

> Add a Automations Subsection. it will be available in all widget types.
>
> Adding the xxRoutine as used somewhere in #3014 (I think)
> add onXx properties (only visible if defined or an as a on demand property for widget types that fit (check wiki))

### Defined routines

Every property ending in `Routine` that is declared on the widget gets a row showing:
- the routine name and a live **step count** (updates when other players or the JSON editor change it),
- a one-line description of when it runs (built-in triggers) or that it's a **custom routine callable with CALL** (any other `xxRoutine`),
- an **edit** button that opens the JSON editor sidebar module and selects that routine's value (same jump-to-property mechanism the Debug module uses),
- a **remove** button (normal delta, undoable).

### On-demand properties

Built-in trigger routines that are not defined (`clickRoutine`, `doubleClickRoutine`, `changeRoutine`, `enterRoutine`, `leaveRoutine`, `globalUpdateRoutine`, `gameStartRoutine` — all documented on the wiki's base widget page) appear as the familiar blue `add xx` buttons; clicking one creates the routine as an empty `[]` and the row appears through the existing property-listener mechanism.

Per-type `onXx` properties are offered where they fit per the wiki: **holders** get `onEnter`/`onLeave`, **cards** get `onPileCreation`. Since their default value is `{}` (setting `{}` would be a no-op), the button expands to the description plus the generic typed input for editing the object inline.

A small **add custom routine** input creates a new named `xxRoutine` (appends `Routine` to the name if missing, flags invalid names with the existing `inputError` style).

All automation properties handled by the section are excluded from the generic fallback list below so they don't appear twice.

### Implementation notes

- Reuses the branch's existing abstractions: `renderOnDemandSection` (extended with an optional `onActivate` hook so a button click can create the property and let the property listeners expand the section), `createOnDemandSectionStructure`, `addPropertyListener`/`addDeltaListener` two-way sync and `addInput`.
- Styling follows the existing `sidebarProperties.css` conventions.

### Inline block editor (added after Discord feedback)

Every routine row has a toggle arrow that expands an inline **block editor** designed to read like a form, not code: one block per step with typed inputs for each parameter (number/checkbox/enum dropdown/JSON, derived from the validator's `operationProps`), greyed-out unset parameters, and up/down/delete controls. IF renders as a sentence (`if [operand1] [relation] [operand2]`) with nested **then**/**else** block lists; any `xxRoutine` parameter (e.g. FOREACH's `loopRoutine`) nests recursively. Collection parameters offer a dropdown of predefined and routine-defined collections plus a *new collection…* entry; widget parameters get a type-filtered multi-widget board picker; every typed input has a **ƒ** toggle for `${variable}` expressions. Each block has a `{}` view-as-JSON button, comments render as grey text with an edit pencil, and the go-to-JSON-editor button is kept on every row. (`getEnumValidator`/`getWidgetTypeValidator` in the validator now tag their functions with `enumValues`/`widgetTypes` metadata to power the dropdowns and picker filters — validation behavior unchanged.)

## Testing

- jest suite passes.
- Verified with Playwright (headless Chromium) in a room with label/holder/basic widgets:
  - defined `clickRoutine` and a custom `shuffleDeckRoutine` render as rows with correct step counts;
  - `add doubleClickRoutine` creates `[]` and the row replaces the button;
  - `add custom routine` with `dealCards` creates `dealCardsRoutine`; invalid names are flagged;
  - holder shows `add onEnter`/`add onLeave`; entering `{"activeFace": 0}` writes through to the widget state;
  - the edit button opens the JSON editor module with the routine's value selected;
  - remove sets the property to `null` and the row shows "not set";
  - no console errors.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3020/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

Solid, well-integrated addition that reuses the branch's on-demand-section and listener plumbing correctly; the built-in routine list and descriptions match the engine (`client/js/widgets/widget.js:86-92`), and jest passes locally. The main finding: widgets that *inherit* a routine (deck `cardDefaults` or `inheritFrom`) show "add clickRoutine", and clicking it writes `[]`, silently disabling the inherited behavior.

## Findings

1. **Inherited routines are invisible and easy to break** — `client/js/editor/sidebar/properties.js:2933` (`isDeclared = property => widget.state[property] !== undefined`). Using `state` instead of `get()` is the right call for deciding what's *declared*, but for a card whose deck defines `clickRoutine` in `cardDefaults` (a very common pattern), the section shows `add clickRoutine` as if nothing runs on click. Clicking the button then sets `clickRoutine: []` on the card, which overrides the deck default and silently disables it — a game-breaking edit that looks like a no-op to the user. Same story for `inheritFrom`. Suggestion: when `widget.state[property] === undefined` but `widget.get(property)` is set, render the row in a distinct "inherited" style (edit could jump to the source) instead of the plain add button.

2. **`edit` silently does nothing under multi-selection** — `properties.js:3050-3064`. `openPropertyInJsonEditor` only jumps when `jeStateNow.id == widget.id`. With one widget selected that always holds, but when several widgets are selected the JSON module shows the multi-view (`jeSelectSetMulti`), so `jeStateNow.id` won't match and the click has no visible effect (or just swaps modules). The existing jump code in `client/js/editor/sidebar/json.js:123` (`button_validationProblem`) calls `setSelection([widgets.get(...)])` first — doing the same here would make the button reliable in all cases.

3. **No way to re-add a routine after `remove` without reselecting** — `properties.js:3040-3043` plus the one-way `expand()` in `renderOnDemandSection`. After the remove button sets the property to `null`, the row stays as "not set" with edit/remove disabled, and the original `add xx` button is gone (removed on expand). The user has to deselect/reselect the widget or use the JSON editor to get it back. Cheap fix: in the `addPropertyListener` at `properties.js:3021`, when the value is unset, turn the row into (or re-show) an add affordance instead of just disabling both buttons.

4. **Minor**: for the `onXx` inputs, choosing "not set" in the type dropdown goes through `inputValueUpdated`, which converts `undefined` to `null` (`properties.js:461`); the returning delta then flips the input to display `NULL` instead of "not set". Pre-existing `addInput` quirk rather than something this PR introduced, but it's more visible here since `{}`-defaulted properties are exactly the ones where "unset vs null" reads oddly.

Also worth noting (no action needed now): like the rest of the `expand-widget-edito` branch, the new UI has no TestCafe coverage; once the branch stabilizes, a compare-state test that creates a routine via the section would guard the `widget.set` paths.

I added the `beta` label: the PR only touches editor UI files (not engine paths), the author verified the flows in a browser, and worst-case failures are editor-side, so it's safe and useful for players to test.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-integrated addition that reuses the branch's on-demand-section machinery correctly; the trigger descriptions match the engine and the JSON-editor jump is implemented the same way the Debug module does it. Two findings worth addressing: engine-triggered `<property>ChangeRoutine` / `<property>GlobalUpdateRoutine` routines are mislabeled as CALL-only custom routines, and removing a routine leaves the row in a dead-end state (both buttons disabled, no way to re-add without reselecting the widget).

## Findings

### 1. Dynamic trigger routines are mislabeled as custom routines

`client/js/editor/sidebar/properties.js:2907-2913` (`automationProperties`) and the custom-routine renderer describe every non-built-in `xxRoutine` as *"custom routine that can be called using CALL"*. But the engine also fires these automatically:

- `<property>ChangeRoutine` — e.g. `activeColorChangeRoutine` runs whenever `activeColor` changes (`client/js/statemanaged.js:131-132`); the canvas color-picker template in `editmode.js` uses exactly this.
- `<property>GlobalUpdateRoutine` — e.g. `ownerGlobalUpdateRoutine` runs when `owner` changes on any widget (`client/js/widgets/widget.js:248`); also used by templates in `editmode.js`.
- `editorAddToRoomRoutine` — run once by the editor when the widget is added to the room (`client/js/editor/sidebar/widgets.js:1415-1418`) and already known to the validator.

A user seeing "callable using CALL" under `activeColorChangeRoutine` gets an actively wrong explanation of when their routine runs. Suggest matching `/(.+)ChangeRoutine$/` and `/(.+)(?:G|^g)lobalUpdateRoutine$/` before falling back to the custom-routine description, mirroring the regex at `widget.js:248`.

### 2. Removing a routine dead-ends the row

`renderAutomationRow` (`properties.js:3040-3047`): the property listener disables both the edit and remove buttons once the value is `null`/`undefined`, and `renderOnDemandSection` already removed the `add xxRoutine` button when the section expanded — it is never re-created. So after clicking remove, the user is left with a "not set" row and no enabled control; the only in-section recovery is (non-obviously) typing the same name into the *add custom routine* input, or deselecting/reselecting the widget. Suggest keeping the edit button enabled when the value is unset and having it create `[]` first (matching the `onActivate` path), or re-showing the add button.

### Minor

- `properties.js:2966-2977` — pressing Enter in the *add custom routine* name input does nothing; a `keydown` handler forwarding to the add button would match usual form behavior.
- No TestCafe coverage for the new section — consistent with the rest of the `expand-widget-edito` branch, but worth a shared test before the branch lands on main, since this section renders for every widget type.

Checked and fine: the built-in trigger descriptions match the engine call sites (`widget.js:476/725/2550/2556`, `serverstate.js:540`, `holder.js:108`, `card.js:15`); `openPropertyInJsonEditor` correctly guards the toggle behavior of `openInTarget` and degrades gracefully on multi-selection; the exclusion of automation properties from every type-specific generic list is complete; no widget/routine properties are added, so no validator changes are needed.

---
The `beta` label was already present and this PR touches no engine files, so I left it in place.

VTT-AI-Agent:
Thanks for both reviews — addressed in d647399a:

**Inherited routines (review 1, finding 1):** `renderAutomationsSection` now checks `automationInheritedValue()` (declared nowhere on the widget but `get()` returns a routine/non-empty object, e.g. deck `cardDefaults` or `inheritFrom`). Such properties render an informational row with a "N steps (inherited)" summary and disabled edit/remove buttons (tooltip explains where to edit) instead of an `add` button — so the destructive "declare `[]` and silently disable the deck's clickRoutine" path is gone. Verified with a card whose deck defines `clickRoutine` in `cardDefaults`.

**Edit under multi-selection (review 1, finding 2):** `openPropertyInJsonEditor` now calls `setSelection([widget])` first when the selection isn't exactly that widget — same approach as `button_validationProblem`. Verified with two widgets selected: the JSON module opens with the routine value selected.

**Remove dead-ends the row (both reviews):** when the value is unset (and not inherited), the edit button stays enabled and re-creates the routine as `[]` before jumping to the JSON editor (matching the `onActivate` path), so remove → edit round-trips without reselecting.

**Mislabeled dynamic trigger routines (review 2, finding 1):** added `customRoutineDescription()` — `<property>ChangeRoutine`, `<property>GlobalUpdateRoutine` (mirroring the `widget.js:248` regex) and `editorAddToRoomRoutine` now get accurate "runs when …" descriptions; only everything else falls back to the CALL-only label.

**Enter key (review 2, minor):** Enter in the custom-routine name input now triggers the add button.

**Not changed:**
- The `not set` → `NULL` dropdown quirk in the `onXx` generic input is the pre-existing `addInput` behavior (`inputValueUpdated` maps `undefined` to `null`); fixing it there would affect every generic property list, so I left it for a separate change.
- TestCafe coverage: agreed, deferring to a shared test once the `expand-widget-edito` branch stabilizes, as both reviews suggest.

All flows re-verified with Playwright (inherited card row, remove/re-add, multi-select edit, dynamic descriptions, Enter key) with no console errors; jest passes.

VTT-AI-Agent:
Added an inline **block editor for routines** in 8ab688c9, prompted by this Discord feedback:

> Add a block editor for click routines. I'd like to still keep the go-to JSON button
>
> — [rapha195 on Discord](https://discord.com/channels/770758631146782780/1525291856106229840/1525411567494234242)

Every routine row in the Automations section now has a toggle arrow (▶) that expands the routine as a list of **step blocks** (works for `clickRoutine` and every other routine the section shows):

- each operation step is a block titled with its `func`; its parameters are edited with the generic typed inputs, and an *add parameter* input suggests the valid parameters for that operation from the validator's `operationProps` (custom names still allowed, sensible defaults per type hint);
- string steps (comments / `var …` expressions) get a monospace text input;
- blocks can be **reordered** (up/down), **deleted**, and new steps added from an operation dropdown (also sourced from the validator) or as a comment/expression;
- expanding an unset routine creates it as `[]` so you can build it from scratch;
- edits from other players or the JSON editor re-render the blocks (skipped while you're typing inside them);
- the **go-to-JSON button stays** on every row, as requested — nested routines (`thenRoutine` etc.) currently fall back to the generic object input or the JSON editor.

Verified with Playwright: parameter edit writes through, move/delete/add step, add parameter with datalist suggestions, external `set()` re-renders, empty-routine flow, JSON button still works; no console errors, jest passes.

VTT-AI-Agent:
Reworked the block editor in 8ed509b0 to address the follow-up Discord feedback:

> Now make it better.
> The goal is to make it very user friendly and not look as much as a code.
>
> Make it more readable. For that each function will need some special input instead of just a list of the parameters.
>
> Example:
>
> Start the If parameters with something like:
>
> "If" + [operand1 input] + [relation input] + [operand2 input]
>
> Make it recursive so If and Foreach use the same block structure for loopRoutine, thenRoutine and elseRoutine
>
> For places where collections give either option for:
> - collection -> list of predefined collections, user defined collections or state a new one
> - select widget -> multiple widget picker
>
> Use correct input types where it fits. Always give an option to use a variable.
>
> Add a button to read individual functions as  plain JSON
>
> Put comments parameters as grey text with a edit button on the side
>
> — [rapha195 on Discord](https://discord.com/channels/770758631146782780/1525291856106229840/1525452025918128221)

What changed, point by point:

- **Typed inputs per parameter**: every known parameter of an operation is rendered with an input matching its validator type — number inputs, checkboxes, enum dropdowns (e.g. FLIP's `faceCycle`), JSON textareas for objects — with a clear button per row; unset parameters are greyed until used, so a block reads as a form rather than JSON.
- **IF as a sentence**: `if [operand1] [relation ▾] [operand2]`, or a single `condition` input when the step uses that form, followed by nested **then** / **else** block lists.
- **Recursive blocks**: `thenRoutine`, `elseRoutine` and FOREACH's `loopRoutine` (any `xxRoutine` parameter) render as nested block lists with the same structure, including their own add-step controls, to any depth.
- **Collections**: `collection` parameters get a dropdown listing the predefined collections (`DEFAULT`, `playerSeats`, `activeSeats`, `thisButton`) plus every collection defined by SELECT/CLONE/TURN steps anywhere in the routine, and a *new collection…* entry that turns into a name input.
- **Widget parameters**: a multi-widget board picker (same picker the parent/seat inputs use), filtered to the widget types the validator allows for that parameter (e.g. MOVE `from`/`to` only accept holders/seats); confirming without picking keeps the current value.
- **Variables everywhere**: every typed input has a small **ƒ** toggle that switches it to a free text input for `${variable}` expressions; string values that don't fit the parameter type open in variable mode automatically. Free-form fields accept variables directly.
- **Per-step JSON**: each operation block has a `{}` button that shows that single step as plain JSON (editable, syncs back); the row-level go-to-JSON-editor button is unchanged.
- **Comments**: shown as grey italic text with an edit pencil that swaps in an input.

To support the dropdowns and type-filtered pickers, `getEnumValidator`/`getWidgetTypeValidator` in `validator/validate_gamefile.js` now tag their returned functions with `enumValues`/`widgetTypes` (metadata only — validation behavior is unchanged, confirmed with `validate_gamefile_node.js` against a library game).

Verified with Playwright against a routine using SELECT/IF/FOREACH/MOVE/AUDIO with nesting: IF sentence edits write through, nested add-step works, collection dropdown offers routine-defined collections, the filtered picker rejects a basic widget and picks a holder, ƒ-variable writes, per-step JSON edit, comment editing, clear buttons — no console errors; jest passes.

VTT-AI-Agent:
Answering this Discord question:

> Is this done or blocked?  For me it seems nothing was done
>
> — [rapha195 on Discord](https://discord.com/channels/770758631146782780/1525291856106229840/1525619161823313920)

**It's done, not blocked** — all of it is on this PR (#3020, head `8ed509b0`, four commits: the Automations section, the review fixes, the block editor, and the user-friendly rework with typed inputs / IF sentence / nested routines / pickers).

If it looks like nothing changed, the most likely reason is **where you're looking**:

- This PR targets the `expand-widget-edito` branch but is **not merged into it yet** — so the **PR-2884 test server does not contain any of this**. It only lands there once you merge #3020 (that's your call as the branch owner; that was the "PR into the branch, don't change the PR directly" setup).
- To try it now, use **this PR's own test server**: https://test.virtualtabletop.io/PR-3020/pr-test — if that server still shows an old build, it may need a redeploy after yesterday's push.

Where to find the UI once you're on the right server: open **edit mode → Properties (tune) sidebar module → select any widget → "Automations" section**. Defined routines show as rows; the **▶ arrow** on a row expands the block editor (typed inputs, if-sentence, nested then/else/do lists, collection dropdowns, widget pickers, ƒ variable toggles, per-block JSON); the `{}` button still jumps to the JSON editor.

If you're on the PR-3020 server and still see nothing, tell me what widget type and which module you're looking at and I'll dig into it.

VTT-AI-Agent:
Fixed both issues in 7f32466e, reported here:

> The recursive blocks are getting out of the space provided by the parent block. Especially the buttons on the right to some inputs.
>
> Some comments in the same line as var statements make the car statement be considered a comment as well.
>
> Example:
> var newValue = ${PROPERTY text OF amount1} * ${value} // multiplies each value in the array by a given number
>
> Only the part after // should be interpreted as a comment
>
> — [rapha195 on Discord](https://discord.com/channels/770758631146782780/1525291856106229840/1525626584638423174)

**Overflow:** the block editor's flex rows (parameter rows, the IF sentence, block headers, add-parameter/add-step bars) didn't wrap or shrink, so once nesting (IF → FOREACH → IF → ...) ate into the available width, fixed-width labels and buttons got pushed past the sidebar's right edge instead of wrapping to a new line. Added `flex-wrap: wrap`, `min-width: 0` and `box-sizing: border-box` across the block subtree so everything — including the ƒ/clear/pick buttons — stays inside the parent block no matter how deep the nesting goes.

**Comment detection:** string steps are now split on the first `//` that's outside a quoted string (matching how the engine itself parses the trailing comment in `evaluateRoutine`), so `var newValue = ... // multiplies...` renders the `var ...` part as normal code text and only the `// multiplies...` part in grey italic — instead of greying out and mislabeling the whole line as "comment".

Verified with a routine containing exactly this mixed var+comment line plus a triple-nested IF/FOREACH/IF/MOVE, in a real browser: no element overflows the sidebar bounds anymore, and the code/comment split renders correctly. Jest passes.

VTT-AI-Agent (commented):
## TL;DR

Follow-up review covering the new **inline block editor** (commits since my last pass). This is a large but well-structured, correctly-abstracted addition; the two findings from my previous review (dynamic-trigger routine labels and the dead-end "removed" row) are now **resolved**, and the validator change is additive metadata only — I confirmed validation behaves identically. Remaining nits are minor: the JSON textareas commit on every keystroke (undo/delta spam), and numeric-looking string literals get coerced to numbers.

## Resolved since last review

- `customRoutineDescription` (`properties.js:2935-2947`) now labels `xxChangeRoutine` / `xxGlobalUpdateRoutine` / `editorAddToRoomRoutine` with their real engine triggers instead of "callable using CALL". ✓
- The edit button is now enabled on an unset row and creates `[]` first (`properties.js:3175-3181`), so removing a routine no longer dead-ends the row. Enter in the *add custom routine* input now submits (`properties.js:3018-3023`). ✓

## Findings (block editor)

### 1. JSON textareas commit on every keystroke — undo/delta spam (minor)

`properties.js:3462` (type-`json` param) and `properties.js:3603` (block-level *show as JSON*) both wire `oninput`, so every intermediate keystroke that happens to parse as valid JSON calls `widget.set` → a broadcast delta to other players **and** a separate undo entry. Typing `{"activeFace": 0}` produces a string of committed states; the undo history then needs many presses to step back through one edit. Every other text/number input in the block editor uses `onchange` (commit on blur). Suggest switching these two textareas to `onchange`/blur, or debouncing, to match.

### 2. `parseLooseValue` prevents numeric-looking string literals (minor)

`parseLooseValue` (`properties.js:3258`) coerces any `/^-?[0-9.]+$/` text to a Number, and it's applied to both IF operands (`:3499`, `:3511`) and generic text params (`:3477`). That's the right default for most comparisons, but it means a value that must stay a string — e.g. a card `value` of `'10'`, a zero-padded code — can't be entered as text (it silently becomes `10`). Operands have no `ƒ` toggle to escape this. Worth at least a mention in the description, or preserving the string when the target is clearly textual.

### Minor / not blocking

- Any `rerender: true` action (enum/collection change, add/move/delete step, add param) rebuilds the whole block tree via `container.innerHTML = ''` (`properties.js:3627`), collapsing sibling blocks' open JSON views. Acceptable for now given the re-render guard on typing.
- The widget-picker callback commits via `inputValueUpdated` rather than the local `commit()` (`properties.js:3548`), so `renderedJSON` isn't updated there; the picked IDs only appear in the text input on the next full re-render. Self-heals but is briefly stale.
- Still no TestCafe coverage for this section (consistent with the rest of the `expand-widget-edito` branch), but this now renders a substantial interactive editor for every widget type — worth a shared test before the branch lands on main.

### Checked and fine

- `operationProps` reaches the client because `validator/validate_gamefile.js` is concatenated into the editor bundle (`server/minify.mjs:137`); the `typeof … === 'undefined'` guards degrade gracefully if it ever isn't.
- `routineParameterSpec` correctly renders `loopRoutine`/`thenRoutine`/`elseRoutine` (function validators) as nested routine lists while CALL's `routine` name param (hint `routineProperty`) stays a text field — the `hint != 'routineProperty'` guard is doing exactly the right thing.
- Validator refactor: attaching `.enumValues` / `.widgetTypes` to the returned functions is purely additive; I ran the validator on a game exercising enum and widget-type params and validation output was unchanged.

---
The `beta` label was already present and I left it in place: the PR touches only edit-mode UI (`sidebarProperties.css`, the Properties sidebar module) plus an additive, behavior-preserving metadata tweak to the validator — no engine files (`client/js/widgets/`, `compute.js`, `serverstate.js`) and nothing that changes how existing games run for players, so it is safe and useful for beta testing.

VTT-AI-Agent (commented):
## TL;DR

Follow-up review covering the new **inline block editor** (commits since my last pass). This is a large but well-structured, correctly-abstracted addition; the two findings from my previous review (dynamic-trigger routine labels and the dead-end "removed" row) are now **resolved**, and the validator change is additive metadata only — I confirmed validation behaves identically. Remaining nits are minor: the JSON textareas commit on every keystroke (undo/delta spam), and numeric-looking string literals get coerced to numbers.

## Resolved since last review

- `customRoutineDescription` (`properties.js:2935-2947`) now labels `xxChangeRoutine` / `xxGlobalUpdateRoutine` / `editorAddToRoomRoutine` with their real engine triggers instead of "callable using CALL". ✓
- The edit button is now enabled on an unset row and creates `[]` first (`properties.js:3175-3181`), so removing a routine no longer dead-ends the row. Enter in the *add custom routine* input now submits (`properties.js:3018-3023`). ✓

## Findings (block editor)

### 1. JSON textareas commit on every keystroke — undo/delta spam (minor)

`properties.js:3462` (type-`json` param) and `properties.js:3603` (block-level *show as JSON*) both wire `oninput`, so every intermediate keystroke that happens to parse as valid JSON calls `widget.set` → a broadcast delta to other players **and** a separate undo entry. Typing `{"activeFace": 0}` produces a string of committed states; the undo history then needs many presses to step back through one edit. Every other text/number input in the block editor uses `onchange` (commit on blur). Suggest switching these two textareas to `onchange`/blur, or debouncing, to match.

### 2. `parseLooseValue` prevents numeric-looking string literals (minor)

`parseLooseValue` (`properties.js:3258`) coerces any `/^-?[0-9.]+$/` text to a Number, and it's applied to both IF operands (`:3499`, `:3511`) and generic text params (`:3477`). That's the right default for most comparisons, but it means a value that must stay a string — e.g. a card `value` of `'10'`, a zero-padded code — can't be entered as text (it silently becomes `10`). Operands have no `ƒ` toggle to escape this. Worth at least a mention in the description, or preserving the string when the target is clearly textual.

### Minor / not blocking

- Any `rerender: true` action (enum/collection change, add/move/delete step, add param) rebuilds the whole block tree via `container.innerHTML = ''` (`properties.js:3627`), collapsing sibling blocks' open JSON views. Acceptable for now given the re-render guard on typing.
- The widget-picker callback commits via `inputValueUpdated` rather than the local `commit()` (`properties.js:3548`), so `renderedJSON` isn't updated there; the picked IDs only appear in the text input on the next full re-render. Self-heals but is briefly stale.
- Still no TestCafe coverage for this section (consistent with the rest of the `expand-widget-edito` branch), but this now renders a substantial interactive editor for every widget type — worth a shared test before the branch lands on main.

### Checked and fine

- `operationProps` reaches the client because `validator/validate_gamefile.js` is concatenated into the editor bundle (`server/minify.mjs:137`); the `typeof … === 'undefined'` guards degrade gracefully if it ever isn't.
- `routineParameterSpec` correctly renders `loopRoutine`/`thenRoutine`/`elseRoutine` (function validators) as nested routine lists while CALL's `routine` name param (hint `routineProperty`) stays a text field — the `hint != 'routineProperty'` guard is doing exactly the right thing.
- Validator refactor: attaching `.enumValues` / `.widgetTypes` to the returned functions is purely additive; I ran the validator on a game exercising enum and widget-type params and validation output was unchanged.

---
The `beta` label was already present and I left it in place: the PR touches only edit-mode UI (`sidebarProperties.css`, the Properties sidebar module) plus an additive, behavior-preserving metadata tweak to the validator — no engine files (`client/js/widgets/`, `compute.js`, `serverstate.js`) and nothing that changes how existing games run for players, so it is safe and useful for beta testing.

96LawDawg:
@VTT-AI-Agent, this needs UI improvements. The function being implemented needs to be more prominent, especially in dark mode. For example, in this routine, the top function is an IF, the next function (inside the then branch) is a FLIP, and the last one is a COUNT, but there is not way to see that. In light mode, it is visible, but not clear enough. So add it prominently (in a color that is not black/white/gray).

Also, when showing as JSON, the JSON should update when the entry is complete (the field changes focus), right now, you have to click the {} button twice to get JSON to update. Clicking the "remove" button for a parameter force closes the opened JSON.

The "use a variable or expression instead of a fixed value" does not work. Typing there erases the property name to the left of the field after opening and then closing the {} JSON view. When entering a value here, if I put something like "thisFace" into the face property of a FLIP, I get "thisFace" but clearly I wanted ${thisFace}.

VTT-AI-Agent:
@96LawDawg Thanks — all three points addressed in 6f385d8e:

**Function name prominence / dark mode:** block titles (IF, FLIP, COUNT, …) are now VTT blue and slightly larger in light mode, light blue in dark mode, and the whole Automations section got proper dark mode styles (dark block/row backgrounds and borders, adjusted greys) — previously it rendered light-mode white boxes there. Screenshots of your example structure (IF → then → FLIP, followed by COUNT):
- dark mode: https://agent.virtualtabletop.io/reports/pr/1525291856106229840/automations-dark-mode.png
- light mode: https://agent.virtualtabletop.io/reports/pr/1525291856106229840/automations-light-mode.png

**JSON view sync:** the root cause was that the `{}` view rendered a snapshot of the step captured at render time — so it opened stale (hence "click twice to update"), and editing that stale JSON could write old values back, which is also what was erasing your freshly-entered parameter. It now reads the current state when opened, **live-updates whenever a field commits** (i.e. when an input loses focus), and **stays open across re-renders** — so clearing a parameter no longer force-closes it (the view just updates to show the parameter gone).

**Variable input:** typing a bare name like `thisFace` into the ƒ input now auto-wraps it to `${thisFace}` (full `${…}` expressions are passed through unchanged). The clear button and the greyed "unset" styling of a row also update immediately after typing instead of waiting for a re-render.

Verified in a real browser (light + dark): fresh JSON on open, live update on field blur, JSON view surviving parameter clear, `thisFace` → `${thisFace}`, and no layout/console errors. Jest passes.

96LawDawg:
Fix:

- When using the arrow to pick widgets from the board, the name of the selected widgets should appear immediately in the field. Otherwise there is no way to know they have been selected.

- Functions that accept "anonymous collections" should make it possible to use one or more widgetIDs. For example,

{
  "func": "CLICK",
  "collection": "stuff"
}

clicks the previously SELECTED collection of "stuff", but

{
  "func": "CLICK",
  "collection": ["stuff"]
}

clicks the widgetId "stuff" and

{
  "func": "CLICK",
  "collection": ["stuff","otherStuff"
}

clicks both widgetIDs of "stuff" and "otherStuff"

- Clicking the button to use a variable or expression in fields that support it should include a drop-down list of variables created within that routine, but only if this does not come at a high computational cost. 
- Deleting an entire routine with the trash can icon does not delete the visible representation of the routine until you leave and come back.
- Many functions do not currently accept variables in the UI but would accept variables in the JSON. For example, in SELECT, type, property, value, max, collection, mode, and sortBy would all work in theory (even if not usual in practice) with variables. Example the "f" availability.

VTT-AI-Agent:
@96LawDawg All five addressed in 0f3bd4cd:

**Picker live feedback:** the parameter's text field now updates immediately with each widget you click on the board (via the picker's pending-selection callback), so you see `h1, stuff` build up as you pick — confirming just finalizes what's already shown. Confirming with nothing picked restores the previous value.

**Anonymous collections:** collection dropdowns now have a **"widget IDs..."** option that switches the field to a widget-ID input with the board picker. It always writes an **array** — so a single entry becomes `["stuff"]` (widget ID), never the bare string `"stuff"` (collection name), preserving exactly the semantic difference you described. Parameters that already hold an array render as that input directly.

**Variable dropdown:** the ƒ input now has a dropdown (datalist) listing variables defined in the routine — `var` statements, `variable` outputs (COUNT/GET/CALL/INPUT/UPLOAD), `VAR` blocks, scanned recursively through nested routines — plus the predefined globals (`playerName`, `seatIndex`, …). It's a single scan of the routine's JSON done only when the input renders, so no meaningful cost. Picking/typing a bare name still auto-wraps to `${...}`.

**Delete routine:** the trash button now collapses the block view immediately (the row shows "not set" with the ▶ arrow ready to start a fresh routine) instead of leaving the stale representation until reselecting the widget.

**ƒ everywhere:** the ƒ toggle is now on every parameter type — including free-text and collection parameters, so SELECT's `type`, `property`, `value`, `max`, `collection`, `mode` and `sortBy` all offer it.

Verified all flows in a real browser (live pick feedback before confirm, `["stuff"]` vs `["stuff","otherStuff"]` written correctly and surviving re-render, datalist showing `cardCount`/`total` from the routine, `cardCount` → `${cardCount}`, immediate collapse on delete) with no console errors; jest passes.

96LawDawg:
#3034 has replaced this.

PR #850, opened 2021-10-30
generate new IDs based on widget type

ArnoldSmith86:
This PR changes the function `generateUniqueWidgetID` to just use the widget type and a counter as the new ID.

This does not work for everything yet because in some places the ID is generated before the rest of the widget has even been created (so the type is not known yet). I will fix that but in the meantime this PR is a base for further discussion.

PR #3022, opened 2026-07-11
Generate new widget IDs based on widget type

VTT-AI-Agent:
Supersedes #850 (see below), rebuilt on the current codebase.

New widgets created without an explicit ID now get readable sequential IDs based on their widget type — `button1`, `holder2`, `card3`, `pile1` — instead of four random characters.

Details:

- `generateUniqueWidgetID(type)` in `serverstate.js` takes an optional type: with a type it returns the first free `<type><number>` ID, without one it falls back to the previous random 4-character ID.
- `addWidgetLocal` passes the widget's type, so widgets added through **authoring** paths (edit mode, JSON editor, …) get a type-based ID. This also fixes the open issue from #850 where the ID was generated before the type was known.
- **Runtime engine paths keep random IDs.** `addWidgetLocal` takes a `useTypeBasedID` flag (default true); the `CLONE` routine op and automatic pile creation on card-stack pass `false`, so live gameplay is unchanged. Readable IDs are an edit-mode convenience with no gameplay benefit, and sequential IDs during play would raise the ID-collision probability between concurrent clients. This also keeps existing games playing byte-identically to before (their public-library reference hashes are unchanged).
- Type-less widgets built directly in the JSON editor / cloned as building blocks still keep random IDs, addressing [RaphaelAlvez's review](https://github.com/ArnoldSmith86/virtualtabletop/pull/850#discussion_r746779035) — they don't get a `basic` prefix that would need constant renaming.
- The SVG game pieces in the add-widget overlay get IDs derived from their piece name: `pawn1`, `checker2`, `meeple1`, `hex3`, … (the "more specific names for game pieces" suggested in that review).
- The remaining pieces that used to get random IDs now get descriptive ones too (per @96LawDawg's review): single poker chips → `marker`, dealer tokens → `dealer`, uploaded boards → `board`, uploaded tokens → `token`, and the **Add Basic Widget** button → `basic`. (Like every type, the counter starts at 1, so a game author actually sees `marker1`, `dealer1`, … — the bare prefix is never used on its own.) The decorative divider and circle are `line` widgets and are named after that type.
- Composite widgets use a fitting prefix for their whole widget tree: `deck1` (holder) with `deck1B`/`deck1P`/`deck1D` children, plus `chips1`, `timer1`, `counter1`, `line1` and `ring1`. Because those child IDs are derived from the base ID rather than generated, `generateUniqueWidgetID` takes an optional callback listing them, and only accepts a base ID for which the whole family is still free — so a stray `deck1B` (an orphaned child, or a widget an author named that way) makes the new deck `deck2` instead of being silently overwritten.
- Every flow that builds a deck names it after its type as well: the "Add New Deck" wizard (traditional, custom, card sheet, text cards, TTS import), the public-library deck browser and the deck editor's own starter and copied decks. A wizard deck is `deck1` and the holder it is added into `holder1`, again only accepted when the reset button and pile IDs derived from it are free.
- The JSON editor tree view shows the descriptive extras (card type, button text, classes) for generated IDs (`button1`, `card3`, `marker1`, …) the same way it already does for random 4-character ones — a poker chip still reads `marker1 (basic - pokerChip - 916,300)`. (Composite children like `deck1B`/`deck1P`/`deck1D` derive their ID from the parent rather than their own type, so — as with the old random `9ee9B`-style children — they keep the plain rendering.)
- Card types, DOM-only IDs and the preview widgets the editor renders into its buttons keep the random generator.

**Games imported from playingcards.io get readable IDs too.** PCIO identifies its widgets with random strings, so a freshly imported game used to be a wall of `q7Ka`/`dEcK` IDs. `server/pcioimport.mjs` now assigns each imported widget an ID derived from the widget type it is imported as — `holder1`, `deck1`, `card3`, `seat1`, `timer1`, `counter1`, `button2`, `checker1`, `label1`, `separator1`, … — and rewrites every reference to the old IDs. A PCIO ID is a random string, so a value equal to one is that widget: the whole widget JSON is walked and every such string replaced, rather than a list of the properties that are known to be references — those differ per PCIO schema version, and an operation names its targets under different keys and nests them differently per version. Only the keys that name something from another namespace are left alone: the widget and piece type, the operation of a routine step, and the card type of a card (a `cardTypes` key, which PlayingCards.io generates the same way it generates an ID). The piles the importer creates for stacked cards are numbered `pile1`, `pile2`, … instead of getting a random ID, which also makes a PCIO import deterministic. The shared hand keeps its `hand` ID, because both the importer and the routines it generates refer to it by that name.

Because that happens on the file before it is converted, every routine the converter builds names the new IDs, whichever argument it read them out of. Widgets the importer drops (a disabled hand, a card whose deck is missing) do not consume an ID number, so the numbering has no gaps, and piece prefixes are lowercased and singularized (`checker1`, `puck1`, `videolink1`).

A demo of the result is on the test server: **[PCIO import demo](https://agent.virtualtabletop.io/reports/pr/1526038126789787838/PCIO%20import%20demo.vtt)** — a small playingcards.io file imported with this branch, playable in [pr-test-ai](https://test.virtualtabletop.io/PR-3022/pr-test-ai). Open it in the editor and every widget is named after what it is. (It is a demo room on the test server only, not a library game or tutorial.)

The nightly corpus replay — the on-demand TestCafe job that loads every public library game and clicks every button in it — could not start: it statically imports the file updater, an `.mjs` module, and TestCafe compiles a test file to CommonJS, so on the Node 18 of the CI job that import is a `require()` of an ES module and throws before the first test. It now reads the legacy modes back from the room, which is where they are assigned in the first place - `Room.setState()` runs the updater for every state carrying `_meta` - so the report records what the room really derived and the test needs no updater at all.

The `editor.js` TestCafe hashes and the `compute-4.js` seeded-random expectations were regenerated for the new IDs, and the expectations of `tests/server/pcio-import.test.js` now name the IDs the importer writes while its PlayingCards.io input keeps the IDs a `.pcio` file has. The public-library reference hashes are unchanged, which is the concrete evidence that the runtime opt-out keeps existing games byte-identical.

---

This PR was requested on Discord:

> Close PR 850 as superceded. Open a new PR that accomplishes the main goal of 850 which is to generate unique widget ids based on the type of widget rather than random 4 characters.

(Source: [Discord message from lawdawg96](https://discord.com/channels/770758631146782780/1525328731319832609/1525372065820053566))

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3022/pr-test (or any other room on that server)


96LawDawg:
Most things seemed to work. However, a few things still got random characters as `id`s: the blue chips (call them marker) and the dealer tokens (call them dealer), both decorative lines (call them separator), basic widgets (call it basic), tokens (call it token), and boards (call it board).

VTT-AI-Agent:
Thanks for the thorough review! I went through each of the pieces you flagged and gave them descriptive IDs:

> Most things seemed to work. However, a few things still got random characters as `id`s: the blue chips (call them marker) and the dealer tokens (call them dealer), both decorative lines (call them separator), basic widgets (call it basic), tokens (call it token), and boards (call it board).

- Blue poker chips (2D + 3D) → `marker`
- Dealer tokens (2D + 3D) → `dealer`
- Decorative lines (vertical + horizontal) → `separator`
- **Add Basic Widget** button → `basic`
- Uploaded tokens → `token`
- Uploaded boards → `board`

I verified `marker`, `dealer`, `separator` and `basic` end up as `marker1`/`dealer1`/`separator1`/`basic1` by driving the editor in a real browser; `board`/`token` use the same explicit-ID mechanism as `basic`.

I also regenerated the TestCafe state hashes that the ID change affects — the `editor.js` test and the public-library games whose auto-created widgets now get type-based IDs (Blue, Bhukhar, Reward, Rummy Tiles, Undercover and a few of the Functions/Types test games). These all pass locally now.

VTT-AI-Agent:
Fixed the remaining CI failure (`compute-4.js` → *Dynamic expressions*). That test clicks **Add Basic Widget** and then evaluates a routine containing `random` / `randRange`. Now that the Add Basic Widget button assigns a type-based `basic` ID instead of a random one, it no longer pulls a value from the seeded RNG, so the subsequent `random`/`randRange` land one position earlier in the seed sequence (`0.974268`→`0.709848`, `120`→`195`). These are pure PRNG outputs so they're identical on Chrome and Firefox; I just updated the two expected values.

I deliberately kept type-based IDs from consuming a random value: making them consume one (to preserve the old RNG position) would couple the RNG stream to the number of auto-created widgets during play, which is timing-dependent and made a game like Reversi flaky. Keeping ID generation rand-free is the stable choice.

VTT-AI-Agent:
> I'm not looking for you to change any code in response to this comment. But do explain why the hash for some games and tutorials had to change. There are no new widgets in those games. This change should only impact new games because widgets get human-readable names.

Good question — and you're right that the **saved** game files are untouched (this PR changes no files under `library/`, only `client/js/**` and the TestCafe specs). Opening any existing game gives the exact same widgets it always had.

The reason those reference hashes changed is that the TestCafe public-library specs don't just *load* a game — they **play** it (the test clicks `btnDeal`, `btnStartGame`, seat buttons, etc.). Playing runs the game's routines, and several routines **create brand-new widgets at runtime**:

- `CLONE` (dealing a deck clones a card widget for each dealt card)
- automatic pile creation (dropping loose cards together spawns a `pile`)

Those runtime-created widgets have no `id` in the routine, so the engine assigns one via `generateUniqueWidgetID` — which is exactly what this PR changed from random to type-based. So the *new* widgets that the test spawns during play now have readable IDs, and since `compareState` hashes the whole resulting room state, the hash moves.

Concrete example — the **Reward** test. Both before and after, playing it produces 340 widgets; the game definition is identical. The only difference is the IDs of the cards the deal *created*:

| before (random) | after (type-based) |
|---|---|
| `12p5`, `1iny`, `2ng4`, `30l7`, `3c3b`, … | `card1`, `card2`, `card3`, … `card42` |

Same widgets, same game, just human-readable IDs for the ones spawned at runtime — precisely the intended effect.

(One secondary effect: because generating a type-based ID no longer pulls a value from the *seeded* RNG the tests use, a shuffle that happens right after a widget is created can land differently, nudging a few card positions too. That only matters for the deterministic test seed — real games use unseeded randomness, so players see no change.)

No tutorial files were modified; the changed references are all in `publiclibrary-*.js` (public-library game tests) plus `editor.js`/`compute-4.js`.

VTT-AI-Agent (commented):
## TL;DR

Clean, well-scoped implementation of type-based widget IDs that fixes the ordering issue from #850. It does genuinely change engine behavior (CLONE, automatic pile creation, and the shared `rand()` sequence all shift), so it is **not** beta-eligible. No blocking bugs found; a couple of edge cases worth noting below.

## Findings

### 1. This really does alter engine behavior (context for maintainers)
`generateUniqueWidgetID` lives in `serverstate.js` and the change flows through `addWidgetLocal`, which is not just edit-mode plumbing — it is the path used by:

- **`CLONE`** (`widget.js:499`) — cloned widgets now get `card5`/`button2`-style IDs instead of random 4-char ones every time a routine clones.
- **Automatic pile creation** (`widget.js:2916`) — dropping a card onto a card during normal play now yields `pile1`, `pile2`, … This happens in essentially every card game.
- The shared **`rand()` PRNG** (`domhelpers.js:100`, seeded via `customRandomSeed`): the type path no longer consumes a `rand()` call, so the global random sequence advances differently. That is exactly why `tests/testcafe/compute-4.js` had to change its `random`/`randRange` expected values — proof that a widget-add now shifts subsequent `random`/`randRange`/shuffle results in deterministic/replay contexts.

None of this looks harmful (games generally consume the CLONE return collection rather than hard-coding generated IDs), and the regenerated hashes cover it. But because default gameplay behavior and RNG consumption change, this is a real engine change and should stay out of the beta auto-deploy — noting it here so it isn't mistaken for incidental.

### 2. Composite widgets only check the parent ID for uniqueness
`generateUniqueWidgetID('deck')` guarantees `deck1` itself is free, but `generateCardDeckWidgets(id, …)` then creates `deck1B`/`deck1P`/`deck1D` and `generateChipPileWidgets`/timer/counter do similar. Those child IDs are never checked against `widgets`. In the normal case the whole tree is added or removed together so it's fine, but an orphaned child (e.g. a `deck1B` whose holder was deleted) could collide when `deck1` gets reused. Low likelihood, but the previous random scheme made this essentially impossible. Not blocking.

### 3. O(n²) for mass clones / large auto-generated sets
`i` restarts at 1 on every call, so adding N same-type widgets in a loop (a big `CLONE count`, dealing a deck) scans `1..k` each time — O(n²) `widgets.has` checks. Fine for typical counts; just flagging it since the old random generator was ~O(1). Not worth optimizing unless a slow case shows up.

### 4. jsonedit tree: `!type && id.match(/^[a-z]+[0-9]+$/)` also matches user-named IDs
The new branch in `jsonedit.js:2469` will now show the "descriptive extras" (cardType/text/classes) for any type-less widget whose ID happens to look like `letters+digits` — e.g. user-chosen IDs such as `col11`, `row21` (Dots). Purely cosmetic in the tree view and probably acceptable, but it's a slightly wider net than "auto-generated ID only." Consider whether that's intended.

### Nice touches
- Handling the type-less case (JSON-editor building blocks, DOM-only/display clones) by keeping the random generator — avoids the `basic`-prefix churn RaphaelAlvez raised on #850.
- The `pieceName` extraction (`editmode.js`) and descriptive fallbacks (`marker`/`dealer`/`separator`/`board`/`token`/`basic`) match the reviewer requests cited in the description.
- Tests and public-library reference hashes regenerated consistently; suite reported passing.

Overall this looks correct and matches repo conventions.

---

I did **not** add the `beta` label: `serverstate.js`'s change to `generateUniqueWidgetID`/`addWidgetLocal` genuinely alters engine behavior (CLONE IDs, automatic pile IDs during normal play, and the shared `rand()` sequence), so the PR is engine-ineligible for the beta auto-deploy by design.

VTT-AI-Agent (commented):
## TL;DR

Clean, well-scoped implementation of type-based widget IDs — the core change to `generateUniqueWidgetID(type)` and `addWidgetLocal` is correct and only affects newly created widgets, so existing saved games are unaffected. The one thing worth thinking through is that `addWidgetLocal` is also on **runtime engine paths** (auto-pile creation on card-stack, and the `CLONE` routine op), so live gameplay now produces sequential IDs like `pile1`/`pile2` instead of random ones — which raises the ID-collision probability in concurrent/multiplayer play and confirms this PR is engine-affecting.

## Findings

### 1. Sequential IDs on runtime paths increase collision risk in multiplayer (main concern)
`addWidgetLocal` now derives IDs from `widget.type` (serverstate.js), and `generateUniqueWidgetID(type)` returns the first free `<type><n>` checked only against the **local** `widgets` map. This is fine in edit mode, but `addWidgetLocal` is also reached during normal play:

- `widget.js:2916` — auto-pile creation when a card is dropped on a card (`addWidgetLocal(pile)` → now `pile1`, `pile2`, …)
- `widget.js:499` — `clone()`, i.e. the `CLONE` routine operation, when the clone's id collides and is regenerated

With the old 4-char random scheme, two clients independently creating a widget of the same type at the same moment had a ~1-in-1.6M chance of picking the same ID. With sequential type IDs, two players stacking cards (or two CLONEs firing) concurrently will both compute `pile1` from their local state and send conflicting adds to the server, where the later delta overwrites the earlier one — potential lost/merged widgets or corrupted piles. Random IDs effectively hid this race; sequential IDs expose it. Worth confirming how the server reconciles duplicate-ID adds, and whether the runtime paths (auto-pile / CLONE) should keep the random generator while only authoring paths use type-based IDs. If you've concluded it's acceptable, a note in the PR would help.

### 2. Minor: PR description vs. actual IDs
The description says the pieces get IDs like `marker`, `dealer`, `separator`, `basic`, but `generateUniqueWidgetID('marker')` returns `marker1` (the counter always starts at 1). Not a bug — just so the description matches what a game author will actually see.

### Things that look correct
- `generateUniqueWidgetID` correctly falls back to random when `type` is undefined, preserving type-less JSON-editor widgets and card types (no unwanted `basic` prefix).
- Composite widgets (`deck1` + `deck1B/P/D`, `chips1`, `timer1`, `counter1`) generate the tree id once and pass explicit child ids, so no per-child regeneration races.
- No FileUpdater/compat work needed — only newly created widgets are affected; existing save files keep their stored IDs.
- The `jsonedit.js` tree regex change (`id.startsWith(type) && …` / `!type && /^[a-z]+[0-9]+$/`) correctly extends the "descriptive extras" display to the new ID shapes without matching composite children.
- Test/reference-hash regenerations (editor.js, compute-4.js RNG-seed drift, publiclibrary-*) are consistent with the ID change.

## Beta label
Not adding the `beta` label: this genuinely alters engine behavior — `generateUniqueWidgetID`/`addWidgetLocal` run on runtime paths (auto-pile creation and the `CLONE` routine op), so how existing games generate widget IDs during play changes, which is exactly the kind of engine change excluded from beta auto-deployment.

VTT-AI-Agent:
Thanks for the very thorough review — went through each point:

**Runtime paths & multiplayer ID collisions (main concern).** I dug into how the server reconciles this. `addWidgetLocal` runs client-side and the id is sent in a delta; the server's per-widget/per-key conflict detection (`Player.receiveDelta` → `triggerDeltaConflict` in `server/player.mjs`) already catches two players adding the *same* id with different properties: the later delta is flagged as a conflict and that client is forced to re-sync (`waitingForStateConfirmation`). So the worst case with sequential ids is **not** corruption/lost widgets — it's the same conflict-resolution path any concurrent edit takes: one of two simultaneously-created `pile1`s wins, the other client re-syncs and its next auto-pile becomes `pile2` (self-healing, one redo). It's a genuinely narrower race than before (random ids essentially never collided), but it's bounded and safe, not data loss.

That said, you're right that during *live play* readable ids buy the user nothing, so if you'd prefer zero behavioral change during gameplay I'm happy to make the two runtime paths — `CLONE` (`widget.js:499`) and auto-pile creation (`widget.js:2916`) — keep the random generator and reserve type-based ids for authoring/edit-mode only. That would also revert the public-library reference-hash churn entirely (they'd play identically to before). I left it as-is for now since it's a behavior reversal you'd already seen and signed off on — happy to flip it if you want; just say the word.

**Composite child uniqueness (`deck1B`/`deck1P`/`deck1D`).** Not a regression: the old random scheme did the exact same thing — it generated the parent (`9ee9`) and derived children `9ee9B`/`9ee9P`/`9ee9D` without checking them against `widgets` either. Same orphaned-child exposure as before, just with a readable prefix now.

**O(n²) generation loop.** Agreed it's not worth optimizing for realistic counts; the `widgets.has` check is cheap and mass-clone/deal sizes are small. Left as-is per your note.

**jsonedit tree match too broad.** Fixed in 54ee1bd — dropped the `!type && /^[a-z]+[0-9]+$/` branch so user ids like `col11`/`row21` no longer pick up the descriptive extras. Type-based generated ids still show them via the `type && id.startsWith(type)` branch, and the type-less generated pieces (`marker1`, `separator1`, …) are already self-describing.

**Description vs actual ids.** Updated the description to note the counter starts at 1, so authors see `marker1`/`dealer1`/… (the bare prefix is never used alone).

96LawDawg:
> That said, you're right that during _live play_ readable ids buy the user nothing, so if you'd prefer zero behavioral change during gameplay I'm happy to make the two runtime paths — `CLONE` (`widget.js:499`) and auto-pile creation (`widget.js:2916`) — keep the random generator and reserve type-based ids for authoring/edit-mode only. That would also revert the public-library reference-hash churn entirely (they'd play identically to before). I left it as-is for now since it's a behavior reversal you'd already seen and signed off on — happy to flip it if you want; just say the word.

Do that. Readable ids do not matter for gameplay.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-scoped implementation of type-based widget IDs; the core `generateUniqueWidgetID(type)` change is clean and the test/hash regenerations look correct. The one thing worth thinking through before merge is that making runtime-generated IDs **deterministic and sequential** (`pile1`, `pile2`, …) meaningfully raises the chance of ID collisions when two players create widgets at the same instant in a multiplayer room — the old random 4-char IDs made that practically impossible.

## Findings

### 1. Sequential IDs increase concurrent-collision risk in multiplayer (main concern)
`addWidgetLocal` runs on the client and picks the ID locally against its own `widgets` map, then syncs via `sendPropertyUpdate`/`sendDelta` (`client/js/serverstate.js:129-160`). It is called at runtime, not just in edit mode:

- automatic pile creation when a card is dropped on a card — `client/js/widgets/widget.js:2916` (`pile` has no explicit id → now `pile1`, `pile2`, …)
- `CLONE` / duplicate — `client/js/widgets/widget.js:499`

With the old random ID (`rand().toString(36)`), two players acting simultaneously would essentially never generate the same ID. With `type + i++`, two clients that both drop a card onto a stack at the same moment will *both* compute `pile1` locally before either delta reaches the server, and the second `addWidget` will `widgets.set('pile1', …)` over the first (`serverstate.js:111`) — silently merging two independent piles / clones rather than creating two. This is low-probability but no longer negligible, and it's a change from prior behavior. Worth confirming the server has some authoritative rename/collision handling, or noting the tradeoff explicitly.

### 2. Composite-widget children don't hit the descriptive-extras branch in the tree view
In `client/js/jsonedit.js:2469`, the new condition is `type && id.startsWith(type) && id.substr(type.length).match(/^[0-9]+$/)`. For deck children like `deck1B` (a button), `type` is `button` and `id` is `deck1B`, which does not `startsWith('button')`, so these keep the plain rendering. So the "shows descriptive extras the same way as random 4-char IDs" claim holds for top-level `button1`/`card3` etc. but not for the `deck1B`/`deck1P`/`deck1D` children the PR itself introduces. Cosmetic only, but the PR description implies otherwise.

### 3. RNG-consumption side effect (minor / expected)
Because typed-widget creation no longer calls `rand()`, the shared RNG sequence advances differently — this is exactly why `tests/testcafe/compute-4.js` `random`/`randRange` expected values changed. In production `rand()` is unseeded `Math.random` so there's no observable effect, but any game driven with `window.customRandomSeed` for reproducible shuffles will now get a different sequence after creating widgets. Almost certainly fine, just flagging that the hash churn is real behavior change, not just cosmetic.

### Positives
- `generateUniqueWidgetID(type)` correctly falls back to the random generator when `type` is falsy, preserving the "type-less widgets keep random IDs" requirement.
- Fixing the #850 ordering bug by generating the ID inside `addWidgetLocal` (where the type is known) is the right call.
- The `String(w.id).match(/^[A-Z][a-z]+/)` piece-name derivation degrades gracefully (no match → falls through to type-based generation).

## Beta label
Not adding the `beta` label: this genuinely changes engine runtime behavior — auto-created pile IDs and `CLONE` results now use deterministic `type`+counter IDs instead of random ones (`serverstate.js`/`widgets/widget.js:499,2916`), which alters how existing games run and is why the public-library state hashes changed — so the PR is engine-ineligible for beta auto-deployment by design.

VTT-AI-Agent:
Done — pushed in 3bd2904. `addWidgetLocal` now takes a `useTypeBasedID` flag (default true); the two runtime engine paths pass `false`:

- `CLONE` routine op (`widget.js:499`)
- automatic pile creation on card-stack (`widget.js:2916`)

So live gameplay keeps the previous random generator, and type-based IDs are reserved for authoring/edit-mode. As you noted, this also reverts the public-library reference-hash churn entirely — I restored all four `publiclibrary-*.js` files to their original hashes and confirmed all 18 tests pass unchanged, i.e. those games now play byte-identically to `main`.

`editor.js` still exercises edit-mode authoring (type-based `spinner1`, `deck1`, `chips1`, `seat1`, `dice2`, …) *and* runtime auto-piles (random again), so I updated its two pile selectors and hash to match. `compute-4.js` is unaffected (the Add Basic Widget button is an authoring path). Full suite green locally.

VTT-AI-Agent:
Thanks — this review lines up with the earlier cycle, and the main finding is now resolved:

**1. Concurrent-collision risk (main concern).** Fixed in 3bd2904 (per @96LawDawg's go-ahead): the two runtime paths — auto-pile creation (`widget.js:2916`) and `CLONE`/duplicate (`widget.js:499`) — now pass `useTypeBasedID = false`, so live gameplay keeps the random generator and two players stacking cards concurrently can no longer both compute `pile1`. Type-based IDs are reserved for authoring/edit-mode, where there's a single actor and no collision window. (For completeness: even before the fix, the server isn't silent-overwrite — `Player.receiveDelta` in `server/player.mjs` detects two clients touching the same widget/key and forces a re-sync via `triggerDeltaConflict` — but keeping runtime IDs random avoids the race entirely.)

**2. Composite children (`deck1B`) don't hit the tree-view extras branch.** Correct, and it's not a regression — the old random scheme's children (`9ee9B`) never matched the display heuristic either, since a child's ID derives from the parent, not its own type. I've reworded the PR description so it no longer implies composite children show the extras.

**3. RNG-consumption churn.** With the runtime paths back on the random generator, gameplay RNG consumption is now unchanged from before; the only remaining shift is edit-mode authoring (which is why `compute-4`'s Add-Basic-Widget-then-`random` values moved). As you note, production `rand()` is unseeded `Math.random`, so there's no player-visible effect.

CI is green on the updated commit.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-scoped change: readable type-based widget IDs (`button1`, `deck1`, …) for authoring paths while runtime paths (`CLONE`, auto-pile) keep random IDs so gameplay stays byte-identical. The design is correct and the runtime/authoring split is exactly right. Only minor, low-severity notes below — nothing blocking.

## Findings

**1. Sequential IDs raise collision risk under concurrent *editing* (low severity, partially acknowledged).**
`generateUniqueWidgetID('button')` (serverstate.js:44-49) resolves collisions against the *local* `widgets` map only. Two authors in the same room adding the same widget type before each other's delta arrives will both compute `button1` locally and emit a duplicate ID — near-certain now, vs. ~1-in-1.6M with the old random 4-char scheme. The PR correctly routes the *gameplay* paths (`CLONE`, pile creation) through `useTypeBasedID=false` to avoid exactly this, but collaborative edit-mode authoring still carries the (much smaller, but real) regression. Probably acceptable given how rare simultaneous editing of the same room is; worth a mention in case a maintainer wants a follow-up (e.g. a short random suffix under contention).

**2. Non-type prefixes don't get JSON-editor tree extras (cosmetic).**
The extras rendering in jsonedit.js:2469 keys off `type && id.startsWith(type)`. The synthetic prefixes that aren't real widget types — `board`, `token`, `marker`, `dealer`, `separator`, `basic` — produce IDs like `board1` on widgets whose actual `type` is `null`/basic, so they fall through both branches and never show the `classes`/text extras a random-4-char basic widget would. Minor inconsistency, not a bug.

**3. O(N²) when bulk-adding same-type widgets (negligible).**
Each `generateUniqueWidgetID(type)` call rescans from `i=1`, so adding N widgets of one type is O(N²) `widgets.has` lookups. Irrelevant for interactive one-at-a-time edit-mode use; just noting it.

## Correctness / gameplay preservation — verified

- The two runtime engine callers (widget.js:499 `CLONE`, widget.js:2916 pile-on-stack) pass `false`, so they call `generateUniqueWidgetID(undefined)` → identical random-4-char behavior *and identical `rand()` consumption* as before. Since `rand()` is only deterministic when `window.customRandomSeed` is set (test harness only; never set in production), gameplay is byte-identical — consistent with the unchanged public-library reference hashes. The `compute-4.js`/`editor.js` hash regenerations are expected: those tests run under a seed and the edit-mode paths now skip `rand()` for IDs.
- All other `addWidgetLocal` callers are edit-mode/JSON-editor/sidebar authoring paths, and the `window`-exposed `addWidgetLocal`/`generateUniqueWidgetID` are only published inside `loadEditMode()`. No runtime routine path picks up type-based IDs. Good.
- ID generation happens at click time inside the handlers (editmode.js:353-360, 373-376), not at overlay-populate time, so `getMaxZ`/uniqueness reflect current state. `w.id.match(/^[A-Z][a-z]+/)` correctly derives `pawn`/`meeple`/`hex`/etc. from the SVG piece names, and explicit-`id` widgets short-circuit the `if(!widget.id)` guard.

## beta label

The two engine-file edits (serverstate.js `addWidgetLocal`/`generateUniqueWidgetID`, widget.js `CLONE`/pile) are **incidental to how existing games run**: the runtime paths are explicitly opted out with `useTypeBasedID=false`, preserving random IDs and identical `rand()` consumption, so live gameplay and reference hashes are unchanged — the only behavior change is readable IDs on edit-mode/authoring paths. It's safe and useful to test and won't corrupt games or crash the server, so I added the `beta` label (a maintainer can allow-list it past the builder's filename-based engine skip).

VTT-AI-Agent (commented):
## TL;DR

Solid, carefully-scoped change. The key correctness concern — that `generateUniqueWidgetID` shares the seeded `rand()` PRNG (`window.customRandomSeed`) with gameplay `random`/`randRange` ops — is handled correctly: the two routine-engine widget creators (`CLONE`, `onPileCreation`) explicitly pass `useTypeBasedID = false`, so they keep consuming a `rand()` value exactly as before and live gameplay stays byte-identical. The engine-file edits (`serverstate.js`, `widget.js`) are incidental — they preserve runtime behavior and only add nicer IDs on authoring paths. No blocking issues; a couple of minor notes below.

## Findings

### Design is correct — the shared PRNG is the thing that could have broken, and it didn't
`rand()` (domhelpers.js:100) advances `window.customRandomSeed` and is the same generator used by the `random`/`randRange`/`randPick` compute ops (compute.js:265,713,720). Previously every widget-ID generation consumed one `rand()` value. By routing only authoring paths to the non-random branch and forcing `CLONE` (widget.js:499) and pile auto-creation (widget.js:2916) to `false`, gameplay's `rand()` consumption sequence is unchanged. The `compute-4.js` hash change (0.974268 → 0.709848, 120 → 195) and the shifted random IDs in `editor.js` (`b86p`→`xxcf`, etc.) are exactly the expected consequence of the *edit-mode* `#addBasicWidget` now consuming one fewer `rand()` call — not a gameplay regression. Good call gating this rather than making all IDs type-based.

### Minor: composite children IDs aren't uniqueness-checked
`generateUniqueWidgetID('deck')` only checks `widgets.has('deck1')`, but the composite then creates children `deck1B`/`deck1P`/`deck1D` with explicit IDs that bypass the uniqueness loop. If `deck1` is free while a stray `deck1B` exists (e.g. an orphaned child, or an author who manually named a widget `deck1B`), `addWidgetLocal` would keep the explicit ID and overwrite the existing widget. This is a pre-existing property of composite-widget creation (random 4-char bases made it ~1/1.6M-unlikely); the sequential base makes it marginally more reachable but still requires an unusual state. Not worth blocking — just flagging in case you want the composite generators to probe child IDs too.

### Minor: `generateUniqueWidgetID` linear scan
The `type + i++` loop is O(n) per call and O(n²) if a game with many same-type widgets is rebuilt, since `i` always restarts at 1. Fine for realistic edit-mode widget counts; noting only for completeness.

### Nits (no action needed)
- The piece-name regex `^[A-Z][a-z]+` cleanly covers every `addPieceToAddWidgetOverlay` call (`Pawn3DSVG`→`pawn`, `Checker2DSVG`→`checker`, `Meeple`, `Pin`, `Marble`, `Cube`, `Pig`, …); if a piece is ever added whose id starts lowercase or with digits, it silently falls back to the type-based/random ID, which is a fine degradation.
- No validator/`jeCommands` upkeep is required since no widget/operation *property* is added or renamed — only ID-generation behavior changed.

## Beta / engine note
Although this PR carries the `engine` label and touches `serverstate.js` and `client/js/widgets/widget.js`, the engine edits are **incidental**: `generateUniqueWidgetID` behaves identically for runtime callers (it's invoked with an undefined type there), and the two `widget.js` changes only pass `false` to *preserve* the prior random-ID behavior on the `CLONE` and pile-creation routine paths. Existing games run byte-identically; the visible effect is limited to edit-mode/authoring ID readability. A maintainer could reasonably allow-list this for the beta build.

The `beta` label is already present and I left it in place — the change is safe to test (client-side, no server changes, no game-corruption or crash risk) and its engine impact is incidental as described above.

VTT-AI-Agent:
Extended the PR: **games imported from playingcards.io now get readable widget IDs too** (338fb75).

PCIO identifies its widgets with random strings, so a freshly imported game was a wall of `q7Ka`/`dEcK` IDs. `server/pcioimport.mjs` now renames every imported widget to an ID derived from the widget type it is imported as, and rewrites all references to the old IDs before the conversion runs:

- `holder1` / `holder2` (card piles and hands), `deck1`, `card1`…, `seat1`, `timer1`, `counter1`, `button1`, `label1`, `separator1`, `board1`, `dice1`, `spinner1`, `checker1` / `pawn1` for game pieces, and the PCIO type name itself for anything the importer doesn't know yet.
- References rewritten: `parent`, `deck`, `allowedDecks`, `linkedSeat`, and the widget IDs inside the converted click routines — `from`, `to`, `holders`, `sources`, `decks`, `timers`, `counters`, `choosers`, `dice`, `spinners`, the object queries (`objects.holders` / `objects.collections`) and the `widgets` popup question holders. It is an explicit field list rather than a blind string replace, so a label whose text happens to equal an ID is left alone.
- The piles the importer creates for stacked cards are now `pile1`, `pile2`, … instead of `Math.random()` IDs — which also makes a PCIO import deterministic.
- The shared hand keeps its `hand` ID: both the importer and the routines it generates refer to it by that name (`seat` → `hand` in SHUFFLE/SORT, `to == 'hand'` → `MOVEXY`).

Verified end to end: uploaded a PCIO file through the real `/addState/.../file/*.pcio` path, loaded the imported game in a browser, sat down and clicked its converted "Deal" button — the card moves from `holder1` to the hand, the pile shuffles and the counter increments, so the remapped routine references resolve. No leftover original IDs anywhere in the output, and the validator reports no widget/property errors on the imported game.

A playable demo is in this PR's test room — [PCIO import demo](https://test.virtualtabletop.io/PR-3022/pr-test-ai) ([download](https://agent.virtualtabletop.io/reports/pr/1541206484027973683/PCIO%20import%20demo.vtt)). It is a test-server demo only, not a library game or tutorial.

No validator or `jeCommands` upkeep is needed — no widget/operation property is added or renamed, only ID generation.

VTT-AI-Agent (commented):
## TL;DR

The client-side part is unchanged since the last round and still looks right (runtime paths opt out with `useTypeBasedID = false`, so gameplay is untouched — I re-ran `editor.js`, `compute-4.js`, `publiclibrary-1.js` and `npm test` on this branch and all pass). The new `server/pcioimport.mjs` commit is well targeted, but it has one real gap: routine steps whose `func` the converter does **not** handle are passed through verbatim, so their widget IDs are now dangling references to the pre-import PCIO IDs instead of matching real widgets. Everything else below is cosmetic.

*Automated review 5 of 4 (code) — by `claude-opus-5`*

## Findings

### 1. PCIO ops the converter doesn't handle keep the old, now-dangling widget IDs (`server/pcioimport.mjs:132`)

`routineIDArgs` covers exactly the arg names the known ops read, and `mapRoutineIDs` only rewrites those. But the conversion loop ends with a catch-all `w.clickRoutine.push(c)` (`server/pcioimport.mjs:1197`) that pushes any *unrecognised* PCIO step through unchanged. Before this PR those steps still carried the real widget IDs, so a maintainer (or the author fixing up the import by hand) could see which widget was meant; now they point at IDs that no longer exist anywhere in the output.

Reproduced with a synthetic `.pcio` against this branch:

```js
{ func: 'SOME_UNKNOWN_OP', args: { widgets: { value: ['zz22'] }, target: { value: 'zz11' } } }
```
comes out of `convertPCIO` as-is, referencing `zz22`/`zz11`, while those widgets are now `holder1`/`deck1`.

The same applies to any arg an otherwise-handled op carries that isn't in `routineIDArgs` (PCIO's automation format grows over time; `routineIDArgs` is a hardcoded allow-list with no fallback). The steps are already invalid VTT routine ops so nothing *crashes*, but the import went from "broken step, but you can see what it referred to" to "broken step referring to nothing".

Cheapest fix that keeps the safety of the current targeted approach: after the widget loop, walk any leftover raw step and map string values that are exact keys of `idMap` (they are pre-import PCIO IDs, so a false positive requires a label/cardType string that is byte-identical to a widget ID). Or, if you'd rather not widen the mapping, at least drop a comment on `routineIDArgs` saying unhandled ops keep raw IDs, so the next person touching this knows.

### 2. Basic widgets with the new synthetic prefixes lost their JSON-editor tree extras (`client/js/jsonedit.js:2465`)

The extras condition is `id.match(/^[0-9a-z]{4}$/) || type && id.startsWith(type) && …`. The synthetic prefixes that are *not* widget types — `basic`, `marker`, `dealer`, `separator`, `board`, `token`, and every SVG piece prefix (`pawn`, `checker`, `meeple`, `hex`, …) — sit on widgets whose `type` is `null`, so neither branch matches and the `type == null && widget.get('classes')` line is now unreachable for them.

Concretely: adding the 2D poker chip (`editmode.js:857`, `classes: "pokerChip"`) used to render in the tree as `k3f2 (basic - pokerChip - 916,300)`; on this branch it renders as `marker1 (basic - 916,300)`. That is a small readability regression against exactly the widgets this PR is trying to make more readable. This was raised in the previous round and is still open — either intentional (fine, but worth saying so) or a one-line fix, e.g. keeping a `type == null && id.match(/^[a-z]+[0-9]+$/)` branch just for the `type == null` extras.

### 3. Composite child IDs still bypass the uniqueness check

`generateUniqueWidgetID('deck')` guarantees `deck1` is free, but `generateCardDeckWidgets(id, …)` then creates `deck1B`/`deck1P`/`deck1D` (and chips/timer/counter do the same) with explicit IDs, which `addWidgetLocal` accepts verbatim. If `deck1` is free while a stray `deck1B` exists — an orphaned child whose holder was deleted, or a widget an author named that way — the composite silently overwrites it. Random 4-char bases made this ~unreachable; sequential bases make it merely unlikely. Still not worth blocking, but it is the one place where the sequential scheme is strictly worse than the old one, and it has now been raised twice without a response.

### 4. PCIO numbering has gaps, and prefixes are inconsistently cased/pluralised (cosmetic)

- The ID is assigned in the up-front loop over `widgets`, but widgets that are later skipped (`hand` with `enabled === false` at `pcioimport.mjs:286`, orphan cards at `pcioimport.mjs:554`) have already consumed their number. So a game whose first hand is disabled shows its only real holder as `holder2`, and card numbering can read `card1, card2, card4`. Harmless, slightly puzzling for a feature whose whole point is readability. Assigning inside the output loop, or just accepting it, are both fine — but the current behaviour is worth a comment.
- `idPrefix` lowercases in the `gamePiece` branch but not in the other one, so an unknown PCIO type yields a mixed-case ID: I get `videoLink1` for `{type:'videoLink'}`, next to `marbles1`/`checker1`. Also `piecePrefixes` singularises `checkers`→`checker` and `classic`→`pawn` but leaves the others plural (`pegs1`, `marbles1`, `pucks1`). Adding the remaining singular forms and lowercasing both branches would make the output uniform.

### 5. PR description is stale about the reference hashes

The last paragraph says "the public-library reference hashes affected by the ID change … have been regenerated", but the diff contains no `publiclibrary-*` changes — correctly so, since the runtime opt-out keeps gameplay byte-identical (I ran `publiclibrary-1.js` on this branch: 5/5 pass against the unchanged hashes). That sentence contradicts the earlier bullet that says the hashes are unchanged; worth deleting so a reviewer doesn't go looking for a hash diff that isn't there.

## Verification I ran on this branch

- `npm test` — 6/6 pass.
- `npx testcafe chromium:headless tests/testcafe/editor.js` — passes with the new `1bba2e79…` hash and the `#w_spinner1`/`#w_deck1B`/`#w_seat1`… selectors.
- `npx testcafe chromium:headless tests/testcafe/compute-4.js` — passes, confirming the `random`/`randRange` expectation change is exactly the seeded-RNG shift from `#addBasicWidget` no longer consuming a `rand()` call, not a gameplay change.
- `npx testcafe chromium:headless tests/testcafe/publiclibrary-1.js` — 5/5 pass with unchanged hashes, which is the concrete evidence that the runtime `useTypeBasedID = false` opt-out works.
- Synthetic `.pcio` imports through `convertPCIO` covering board/cardPile/cardDeck/card/seat/counter/timer/labelText/separator/dice/spinner/gamePiece/automationButton/turnButton plus every converted op: IDs come out as `board1 button1 button2 card1 card2 checker1 counter1 deck1 dice1 hand holder1 label1 pegs1 pile1 seat1 separator1 spinner1 timer1` with all `from`/`to`/`holders`/`decks`/`timers`/`counters`/`dice`/`spinners`/`objects.holders`/`objects.collections` and the `questions[].holders` popup correctly rewritten. Derived child IDs (`seat1count1`, `holder1_label`, `timer1P`, `counter1_incrementButton`, …) cannot collide with generated base IDs, since prefixes are digit-stripped — that part is sound.

No validator / `jeCommands` upkeep is needed (no widget property or routine-operation property is added or renamed), no UI-support gap (nothing new is exposed to game authors as a settable property or routine function), and nothing under `library/` is touched — the demo lives on the PR test server as a `.vtt`, which is right.

The `beta` label was already present and I left it in place: the `serverstate.js` / `widgets/widget.js` edits are incidental to how existing games run (the two runtime callers — `CLONE` at `widget.js:499` and pile auto-creation at `widget.js:2916` — pass `useTypeBasedID = false` and keep both the random IDs and the identical `rand()` consumption, which the unchanged public-library hashes confirm), and `server/pcioimport.mjs` only affects one-off imports, so the change is safe and useful for players to test.

VTT-AI-Agent:
Thanks — all five points are addressed in 4f22670.

**1. Unconverted PCIO ops kept dangling IDs.** Fixed. `mapRoutineIDs` now splits the two cases: for the operations the converter actually handles (`convertedFuncs`) the targeted `routineIDArgs` mapping is unchanged, and for everything else — the steps that get pushed through verbatim at the end of the loop — every string in the step that is an exact key of `idMap` is mapped, recursively through nested arrays/objects. Your synthetic case now comes out as

```js
{ func: 'SOME_UNKNOWN_OP', args: { widgets: { value: ['puck1'] }, target: { value: 'holder2' }, nested: { deep: ['holder1', 'notanid'] } } }
```

Keeping the deep mapping to unknown ops only means the handled ops behave byte-identically to before (they are rebuilt from named args anyway, so anything not in `routineIDArgs` is discarded rather than emitted), while a false positive would need a string in an unknown op to be byte-identical to a widget ID.

**2. Tree extras for synthetic prefixes.** Fixed — the condition now falls back to `id.match(/^[a-z]+[0-9]+$/)` when `type == null`, which is exactly the case the `classes` extra covers. Verified in the browser: adding the 2D poker chip renders as `marker1 (basic - pokerChip - 920,114)` in the tree again.

**3. Composite child IDs bypassing the uniqueness check.** Fixed, and thanks for pushing on it. `generateUniqueWidgetID(type, derivedIDs)` takes an optional callback returning the IDs a composite derives from a candidate base, and only accepts a base for which none of them is taken. The deck/chips/timer/counter composites go through a small `addCompositeWidget(type, generate)` helper that derives the list straight from the generator, so it stays correct if a generator ever gains another child. The public-library deck browser (`addLibraryDeckToGame`) had its own hand-rolled `do/while` over the same suffix list — that now uses the shared check too, and gets `deck`-prefixed IDs like everything else instead of a random base.

Verified in a real room: with a stray `deck1B` present, adding the empty deck produces `deck2` / `deck2B` / `deck2D` instead of overwriting it.

**4. Numbering gaps and prefix casing.** Fixed both. The up-front loop now skips the widgets the output loop drops (a `hand` with `enabled === false`, a card/piece/chooser whose deck is missing), so they no longer consume a number. `idPrefix` lowercases in both branches, and instead of listing plurals one by one it singularizes with `name.replace(/([^s])s$/, '$1')` — `checkers`→`checker`, `pucks`→`puck`, `marbles`→`marble`, while `chess`-style names ending in `ss` are left alone. A synthetic file with a disabled hand, an orphan card, a `pucks` piece and an unknown `videoLink` now imports as `board1 button1 card1 card2 card3 checker1 deck1 holder1 holder2 pile1 puck1 videolink1` — contiguous card numbering, all-lowercase prefixes.

**5. Stale PR description.** Removed; the paragraph now says the public-library hashes are unchanged and that `publiclibrary-1.js` passing against them is the evidence for the runtime opt-out.

Re-ran on the branch: `npm test` 6/6, `editor.js`, `compute-4.js` and `publiclibrary-1.js` (5/5, unchanged hashes) all pass.

VTT-AI-Agent:
Updated the branch from `main` and resolved the conflicts.

> Update from main and resolve conflicts.

(Source: [Discord message from lawdawg96](https://discord.com/channels/770758631146782780/1526038126789787838/1541226607618826390))

`main` had moved 70 commits since the last merge, three of them straight through this branch's files.

**Conflicts resolved**

- `client/js/editmode.js` — the add widget overlay was rebuilt in #3127: it sizes itself to the board, works out where a widget goes on click and passes a `title` for every entry. The composite entries keep that (their `onClick` now takes the position), and the type-based IDs are re-applied on top: `addCompositeWidget(type, generate)` also covers the new `line` and `ring` composites, and `addWidgetToAddWidgetOverlay` takes the ID prefix as a fourth argument next to the title. The vertical/horizontal divider lines that used to be basic widgets named `separator` are real `line` widgets since #3060 and are named after that type.
- `client/js/editor/sidebar/properties.js` — the deck wizard was rewritten around card designs (#3049 and follow-ups). Every path that actually adds a deck to the game names it `deck`, the preview decks the editor renders into its buttons keep random IDs, and the holder the wizard creates uses the new derived-ID check (`generateUniqueWidgetID('holder', base=>[ base+'B', base+'P' ])`) instead of its own retry loop.
- `tests/testcafe/editor.js` — took main's file and re-applied the ID changes.

**Fixed while merging**

- The deck editor's own flows (`copyDeck`, `createStarterDeck`) still generated random IDs, so a deck made there was `hz3k` while everything else was `deck1`.
- The PCIO importer lost the seat that holds the turn: schema 6+ writes it as `currentTurn` on the turn button, which was not in the list of properties whose widget IDs get mapped.
- The importer mapped routine IDs by known argument name for the operations it converts. `main` has since added operations that carry widget IDs under other keys, so the routine is now walked as a whole — every string in it that is a PCIO widget ID is replaced, wherever it sits. That also covers the steps that are copied through verbatim, which is what the previous split did.

**Tests**

- `npm test`: 1430 passed. `tests/server/pcio-import.test.js` came with main and asserts on the IDs the importer writes; its expectations now name the generated IDs (`holder1`, `deck1`, `button1`, …) while its PlayingCards.io input keeps the IDs a `.pcio` file has.
- `tests/testcafe/editor.js`: the two piles the pile overlay creates while splitting a deck are made on a runtime path and keep random IDs, but those shifted with the seeded random stream (`b86p`/`5ip4` → `xxcf`/`4yaq`). Nine state hashes were regenerated — the deck editor, the new deck wizard and the line widget all build their widgets through the editor now, so those states carry type-based IDs.
- The public-library hashes are unchanged, as before.

96LawDawg (approved):
Tested.  Approved.  Ready for merge.

ArnoldSmith86:
- PCIO import: the IDs are unique enough that I think it's better to replace them in the entire JSON (not just routines and certain properties)
- run the on-demand testcafe test that goes through the _entire_ public library

VTT-AI-Agent:
Both points are in, on f793339 / 014a32b:

**1. PCIO import — replace the IDs in the whole JSON.** Done. `server/pcioimport.mjs` now walks every widget object it read out of `widgets.json` and replaces every string that equals a PlayingCards.io ID, instead of the previous list of properties known to be references (`parent`, `deck`, `allowedDecks`, `linkedSeat`, `currentTurn` and the click routine).

Four keys are deliberately left out, because they name something from another namespace and replacing them would corrupt the file:

- `type` and `pieceType` — the widget type. A file whose widget is called `card` would otherwise be imported as `"type": "card1"`.
- `func` — the operation of a routine step.
- `cardType` and `chooserChoice` — the key of an entry in the deck's `cardTypes`, which PlayingCards.io generates the same way it generates a widget ID. If a card type and a widget happen to draw the same string, replacing it leaves the card pointing at a card type that does not exist. Happy to drop this exception if you would rather have those replaced as well.

Two tests came with it: one that no PCIO ID survives anywhere in the output of a file that names widgets in every reference property and in a routine, and one that a card type colliding with a widget ID stays put.

**2. The on-demand job that replays the entire public library.** It could not run at all: `tests/testcafe/corpusreplay.js` imports `flagsForGame`, which pulls in `server/fileupdater.mjs`, and TestCafe compiles a test file to CommonJS — so on the Node 18 of the CI job that import is a `require()` of an ES module and throws before the first test. Every shard of the nightly "Corpus replay" has been dying that way since the import was added (16/16 red every night, ~50 s per shard).

It reads the legacy modes back from the room now, right after the game is loaded and before an override is applied — which is where they are assigned in the first place, `Room.setState()` runs the updater for every state carrying `_meta` — so the test needs no updater import. If you would rather have that fix in a PR of its own, say so and I will hand it over.

With that in place the job ran over this branch: https://github.com/ArnoldSmith86/virtualtabletop/actions/runs/32761169515

- 539 of the 570 variants replayed so far, 4910 buttons clicked (15 of the 16 shards are in; shard 5 is still running while I write this)
- **0 engine exceptions** — both with the modes the updater derives and, for the 312 variants that get any, a second time with every mode off
- 9 variants end in a different state when their legacy modes are switched off, which is the diagnostic the job records rather than a failure
- shard 10 is red for a single aborted test, `games/Symphony 0`. No report was written for it and it replays clean locally in 16 s, so that reads as CI flakiness, not a finding.

VTT-AI-Agent:
The library replay has finished — and a correction to my last comment: the red shard is **not** flakiness, it is a real client crash. It just does not come from this branch.

Full run: https://github.com/ArnoldSmith86/virtualtabletop/actions/runs/32761169515

- 575 of 576 variants replayed, 5411 buttons clicked, 15 of 16 shards green
- 0 engine exceptions in those 575, with the modes the updater derives and — for the 333 that get any — a second time with every mode off
- 9 variants end in a different state when their modes are switched off, which is the diagnostic the job records: Flatten the Curve 0, Mexican Train 0, Snail trafficking 0, Fowl Play 0, Flight 0, From now we float 0, Card Games 0, Dice Conflict 0, A Simple Life 1

The one failure, `games/Symphony 0` on shard 10, is an uncaught exception in the client, which is why no report was written for it:

```
TypeError: widgets.get(...).removeCard is not a function
    at Button.applyRemove
    at Button.applyRemoveRecursive
    at receiveStateFromServer
```

`Widget.applyRemove` (client/js/widgets/widget.js:416) checks that the `deck` property names a widget that exists, but not that it *is* a deck, so removing a non-card widget that carries a `deck` throws while a state update is being applied — and the client is left half way through that update.

It reproduces on `main` at 25d945b7 with no change from this branch, on `games/Splendid` variant 1:

```
CORPUS_GAMES=Splendid npx testcafe chrome:headless tests/testcafe/corpusreplay.js
```

(Which game trips it depends on the shard: shard 10 replays Splendid variant 0 and hit it on Symphony instead.) I have left it alone here — it is an engine bug with nothing to do with widget IDs, so it wants its own PR rather than a fix smuggled into this one. Say the word and I will take it.

ArnoldSmith86 (approved):

PR #3023, opened 2026-07-11
Add TTS deck import workflow to the add-deck properties UI

VTT-AI-Agent:
## What

The **Add deck** section of the Properties editor module (shown when nothing is selected in edit mode) gets a third option: **Import from Tabletop Simulator**.

Workflow:
1. Enter the link of a TTS Steam Workshop item.
2. The server resolves and converts it with the existing TTS import pipeline (`server/ttsimport.mjs` via `FileLoader.readStatesFromLink`), so no new conversion logic was added.
3. The sidebar shows a preview button for every deck found in the converted game (rendered with the same widget-preview mechanism as the existing deck templates), each labeled with its total card count.
4. Select one or more decks and click **Add to game** — the decks are imported together with their cards, using the **same per-card-type counts as in TTS** (counted from the card widgets the converter emits, so duplicated cards in a TTS deck stay duplicated).

## How

- New `POST /api/decksFromLink` endpoint: takes `{ link }`, loads the states behind the link with the existing `FileLoader`, runs `FileUpdater` on each variant, and returns every `deck` widget together with a `cardCounts` map (cardType → number of card widgets referencing it).
- Client: new `deckImportTTS()` in the Properties module reuses `renderWidgetButton()` for the previews and the existing `addDeckWithCards()` helper (already used by the image-deck flow) for the import, so positioning, delta batching and delta-cause logging behave like the other add-deck flows.
- A little CSS for the link input row and the card-count badge on the preview buttons.

## Verification

- `npm test` (jest) passes.
- Tested end-to-end against real Steam Workshop items on a local server (Playwright): decks are listed with correct previews and counts, multi-select works, and importing creates the decks plus one card widget per TTS card occurrence (verified counts in the room state, including a deck with 1 card type × 4 copies).

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3023/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

Solid, well-scoped feature: the endpoint and the client flow both reuse the existing import machinery (`FileLoader.readStatesFromLink`, `FileUpdater`, `renderWidgetButton`, `addDeckWithCards`) instead of duplicating it, and I found no correctness or security regressions. The main gap is error UX: most realistic bad inputs (non-workshop URL, invalid/deleted workshop ID) don't produce a `UserError`, so the user just gets an alert saying "Internal Server Error".

## Findings

1. **Unfriendly errors for the most common failure modes** — `server.mjs` (new `/api/decksFromLink` route). Only a missing/non-http link yields a `UserError`. Everything else bubbles to the generic 500 handler and the client alerts "Loading decks failed: Internal Server Error":
   - an invalid or deleted workshop ID: Steam's `GetPublishedFileDetails` returns no `file_url`, so `TTS.resolveLink` returns `undefined` and `fetch(undefined)` throws (server/ttsimport.mjs:333-344);
   - a URL that resolves but isn't a TTS save or zip: `JSZip.loadAsync`/`fromBSON` throws a parse error;
   - a pasted link with a 3+-segment `#a/b/c` suffix: `readStatesFromLink` returns `undefined` and `Object.values(undefined)` throws a TypeError.

   The existing link-import path wraps this in `UserError(404, 'Unable to load and add the game.')` (server/room.mjs:79-85). Wrapping the `readStatesFromLink` call the same way (keeping real `UserError`s intact) would make the alert useful, since bad links will be *the* common case for this UI.

2. **Multi-variant games list every deck once per variant** — server route loops over all variants and pushes each deck again, and the sidebar buttons carry no state/variant label, so for a link whose game has several variants the user sees indistinguishable duplicates (and can import both). Deduplicating by state/variant, or labeling the buttons, would help. Minor for TTS links (usually one variant) but visible for VTT links, which this endpoint also accepts.

3. **No feedback / no reset after "Add to game"** — `client/js/editor/sidebar/properties.js` (`deckImportTTS`, add-button handler): the selection stays active and the button stays enabled, so a second click silently imports duplicate decks (with fresh IDs). Clearing the selection or disabling the button after a successful import matches how the other add-deck flows leave the module.

Not a blocker, just noting for the record: the endpoint lets any unauthenticated visitor make the server fetch an arbitrary URL, but that exposure already exists via `PUT /addState/:room/...` with type `link` and the in-room "add game from link" path, so this adds no new attack surface class.

None of the changed files (`server.mjs`, `client/js/editor/sidebar/properties.js`, `client/css/editor/sidebarModules.css`) are engine files — this is purely additive editor UI plus a new read-only server endpoint, so existing games' behavior is untouched.

I added the `beta` label: the change is editor-only with no engine impact, reuses battle-tested import code, its failure paths are caught by the existing error middleware (no crash risk), and it's safe and useful for players to test despite the error-message and duplicate-listing polish items above.

ArnoldSmith86:
Address the review.

VTT-AI-Agent:
Addressed the review in 4780f25a:

1. **Unfriendly errors** — the `/api/decksFromLink` route now wraps `FileLoader.readStatesFromLink` in a try/catch (and null-checks its result), so an invalid/deleted workshop ID, a non-TTS URL, or a malformed `#a/b/c` hash suffix all return a `404 UserError` ("Unable to load and convert the game behind that link.") instead of the generic 500. Real `UserError`s are re-thrown unchanged. Verified locally: bad non-http link → 400 "Please provide a link.", deleted ID / `example.com` / 3-segment hash → 404 friendly message, valid link still returns decks.

2. **Multi-variant duplicates** — each returned deck is now tagged with its source state/variant, and the preview button shows that label (e.g. `10 cards (TTS #2)`) whenever the link contains more than one variant, so multi-variant VTT links no longer show indistinguishable duplicates.

3. **No reset after "Add to game"** — the selection is cleared and the button disabled after a successful import, matching the other add-deck flows and preventing an accidental duplicate import on a second click.

`npm test` passes and I re-ran the end-to-end Playwright check (list → multi-select → import): card counts still match TTS and the UI resets as expected.

VTT-AI-Agent (commented):
## TL;DR

Clean, well-scoped editor-only feature: a new "Import from Tabletop Simulator" option in the add-deck UI plus a `POST /api/decksFromLink` endpoint that reuses the existing TTS import pipeline. No engine files are touched (server.mjs + editor/CSS only), so existing games' behavior is unchanged. Main thing worth a look: the new endpoint is an unauthenticated server-side fetch of an arbitrary `https?://` URL — a (largely pre-existing) SSRF surface that could be tightened to TTS links since the feature is TTS-specific.

## Findings

### 1. SSRF surface: endpoint fetches any `https?://` URL (server.mjs:320)
The endpoint only validates `req.body.link.match(/^https?:\/\//)` and then hands the raw link to `FileLoader.readStatesFromLink` → `downloadLink` → `TTS.resolveLink`. Because `resolveLink` returns non-`?id=` links unchanged (`server/ttsimport.mjs:335-336`), any URL — including internal/metadata endpoints — gets fetched by the server. Exfiltration is limited (the response must parse as a VTT/PCIO/TTS zip or it errors), so this is closer to a blind/limited SSRF than a full read primitive, and the same capability already exists via the room link-loading path. Still, this is now a *plain unauthenticated HTTP* entry point, and the feature is specifically "Import from Tabletop Simulator." Consider rejecting anything `TTS.isTTSlink()` doesn't match, so the endpoint can only ever hit resolved Steam file URLs. Low severity given parity with existing behavior, but cheap defense-in-depth.

### 2. Partial import + stuck button on error (properties.js addButton.onclick, ~line 771)
`addButton.onclick` is `async` and awaits `addDeckWithCards` in a loop with no try/catch. If one deck fails to import, the promise rejects unhandled: already-imported decks stay, the failing/remaining selected decks are left `selected`, and `addButton` stays `disabled` — the user can't retry without re-running "Find decks". The `findButton` path handles this well (re-enables in all cases); mirroring that with a try/catch here would make the multi-deck flow more robust.

### 3. Cards with no `cardType` are silently dropped (server.mjs card-count loop; properties.js addDeckWithCards)
`cardCounts[card.cardType]` keys on the raw `cardType`; a card widget without a `cardType` produces a `cardCounts[undefined]` bucket that `addDeckWithCards` never re-creates (it iterates `deck.cardTypes` keys). TTS-converted cards normally have a `cardType`, so this is an edge case, but if a converter variant ever emits typeless cards the reported/added counts would silently diverge from TTS. Worth a defensive skip or a comment noting the assumption.

### Minor
- `.deckCardCount` is `position:absolute` (sidebarModules.css); it relies on the widget button being a positioned ancestor. Verify the badge anchors to the button rather than an outer container across the deck-preview scaling transform.
- The preview `renderWidgetButton(new Deck(...), ...)` / `addDeckWithCards` reuse mirrors the existing image-deck flow exactly, which is good — original TTS per-card positions/faces are intentionally discarded and regenerated from `cardTypes`, consistent with the other add-deck flows.

Nothing here is a correctness blocker; the change is additive and gated behind a new UI option.

---

I left the `beta` label in place (it was already present): the diff touches only `server.mjs`, an editor sidebar module, and editor CSS — none of the engine files (`client/js/widgets/`, `compute.js`, `serverstate.js`) — so it does not alter how existing games run and is engine-eligible for beta, and the feature is additive and safe to test.

VTT-AI-Agent:
Addressed auto-review 3/5 in a9497450:

1. **SSRF surface** — `/api/decksFromLink` now rejects anything `TTS.isTTSlink()` doesn't match (returns a `400` asking for a `…/filedetails/?id=…` link), so the endpoint can only ever fetch resolved Steam Workshop file URLs, not arbitrary URLs. This matches the feature's intent (it's specifically "Import from Tabletop Simulator"). Verified: `https://example.com/` → 400, valid workshop link still returns decks. Note this also means non-workshop VTT links are no longer accepted, so the multi-variant labeling from the previous round is now mostly defensive — I left it in since `TTS.fromZip` could in principle yield more than one variant.

2. **Partial import + stuck button** — the multi-deck "Add to game" loop is now wrapped in try/catch, and the button's `disabled` state is always recomputed from the remaining selection afterwards. A failed import now shows an alert and leaves the still-selected decks retryable instead of stranding the button disabled.

3. **Typeless cards** — the count loop now skips cards with no `cardType` (with a comment), since `addDeckWithCards` recreates cards from `deck.cardTypes` and can't represent typeless ones, so reported and added counts stay in sync.

**Minor** — confirmed the `.deckCardCount` badge anchors to the preview button itself: `.widgetSelectionButton` is `position:relative`, so the badge sits at the button's bottom edge across the scaling transform (verified visually).

`npm test` passes and I re-ran the end-to-end Playwright flow (find → multi-select → import) with correct per-type counts.

VTT-AI-Agent (commented):
## TL;DR

Clean, well-scoped additive feature (new editor add-deck option + `POST /api/decksFromLink`). No engine behavior changes, SSRF is properly mitigated, and labels/badges are HTML-escaped. Main thing worth improving: the card-counting loop in the new endpoint is O(decks × widgets), which can block the event loop on a large TTS game — otherwise just minor notes.

## Findings

### Efficiency — quadratic card counting can stall the event loop (`server.mjs`, ~L354–366)
For each deck the endpoint re-scans the full `widgets` list to build `cardCounts`:

```js
for(const [ deckID, deck ] of widgets) {
  if(deck.type != 'deck') continue;
  const cardCounts = {};
  for(const [ cardID, card ] of widgets)   // full re-scan per deck
    if(card.type == 'card' && card.deck == deckID && card.cardType != null)
      ...
}
```

That is O(decks × widgets) per variant, all synchronous on the main thread. A big TTS conversion (many decks, thousands of card widgets) can produce tens of millions of iterations and briefly block the server for everyone. A single pass building `cardCountsByDeck[card.deck][card.cardType]++`, then attaching the map per deck, is the same result in O(widgets). Worth doing since TTS games are frequently large.

### Minor — badge count can exceed what actually gets imported (`server.mjs` L363 / `properties.js` `addDeckWithCards`)
The preview badge sums every card widget with `cardType != null` referencing the deck, but `addDeckWithCards` only creates cards for `cardType`s that are keys of `deck.cardTypes`. If the converter ever emits a card whose `cardType` isn't registered in `deck.cardTypes`, the "N cards" badge will be higher than the number of cards actually created on import. In practice the TTS converter should register every card's type, so this is likely only a theoretical mismatch — but it means the displayed count isn't guaranteed to equal the imported count.

### Minor — endpoint is unauthenticated / unthrottled (`server.mjs` L321)
`/api/decksFromLink` has no room/session check and no rate limit, so anyone can drive the server to fetch + convert Workshop items repeatedly (network + CPU, plus the quadratic pass above). The SSRF surface itself is nicely closed — `isTTSlink()` is required and `resolveLink()` only ever uses the numeric `id` to hit the Steam API, so an attacker can't point the fetch at an arbitrary host (the defense-in-depth comment is accurate). Given the download/convert capability already exists elsewhere this is low severity, but a lightweight guard (or reusing an existing caching/throttle path) would be reasonable hardening.

## Positives
- Errors are funneled through `Logging.UserError` + `.catch(next)` matching the existing endpoint convention; the generic catch downgrades unexpected failures to a friendly 404.
- Preview labels/sources go through `html()`, and deck state is deep-cloned before both `renderWidgetButton` (which mutates its state arg) and import, so the stored originals stay intact.
- Reuses `FileLoader`/`FileUpdater`/`addDeckWithCards`/`renderWidgetButton` rather than adding parallel logic — good for consistency with the image-deck flow.

No automated test covers the new endpoint or flow, but that's consistent with the rest of the HTTP surface; `npm test` passing plus the described Playwright check is reasonable here.

## Beta label
Left the existing `beta` label in place: the change touches only CSS, the editor sidebar (`properties.js`), and server routing — none of `client/js/widgets/`, `compute.js`, or `serverstate.js` — so it's an additive editor feature with no engine impact, and it's safe and useful for players to test.

VTT-AI-Agent:
Addressed auto-review 4/5 in 3a072a7a:

1. **Quadratic card counting** — replaced the O(decks × widgets) loop with a single pass over the widget list that groups counts into `cardCountsByDeck[deck][cardType]`, then attaches the map per deck (O(widgets) per variant). Verified on a large game (id `2789930775`, 104 decks): identical results to before, no unknown-cardType counts, duplicates preserved (e.g. `{24800: 4}`).

2. **Badge vs. imported count** — the endpoint now only reports counts for cardTypes that are actually keys of `deck.cardTypes`, so the "N cards" badge is guaranteed to equal what `addDeckWithCards` recreates on import. On that same game, 0 of the 104 decks had counts referencing an unregistered cardType.

3. **Unauthenticated/unthrottled** — leaving as-is (low severity, per the review): the SSRF surface is closed via `isTTSlink()`, and the underlying `downloadLink` already caches each resolved link to disk and reuses it for ~2h, so repeated requests for the same item don't re-fetch/re-convert. A general rate-limit would be a broader change than this PR's scope.

`npm test` passes; re-ran the end-to-end Playwright flow (find → multi-select → import) with correct per-type counts.

VTT-AI-Agent (commented):
## TL;DR

Clean, well-scoped addition of a TTS-Workshop deck-import option to the edit-mode Properties sidebar. No engine files are touched, the new `/api/decksFromLink` endpoint is guarded against SSRF, and the badge/import counts are kept consistent by design. Nothing blocking; a couple of low-severity notes below.

## Findings

**Correctness / design — looks good**
- The badge count and the actually-imported card count are deliberately kept in sync: the server filters `cardCounts` to only `cardType`s present in `deck.cardTypes` (`server.mjs`), and `addDeckWithCards` recreates cards by iterating `deck.cardTypes` with those same counts. Nice, this avoids the classic "preview says N, import gives M" mismatch.
- SSRF: the endpoint requires `TTS.isTTSlink(link)` (`/\?id=([0-9]+)/`), and `TTS.resolveLink` ignores the user-supplied host entirely — it re-derives the download URL from Steam's `GetPublishedFileDetails` using only the numeric id. So the server never fetches an attacker-chosen host. The defense-in-depth comment is accurate.
- Reusing `renderWidgetButton` / `addDeckWithCards` keeps positioning, delta batching and delta-cause logging identical to the existing image-deck flow. TTS decks reference remote `steamusercontent` image URLs (see `processURL` in `ttsimport.mjs`), so importing just the deck widget works without any asset-registration step — no broken images.
- Malformed states are handled defensively: `FileUpdater()` is wrapped in try/catch→`continue`, and non-variant entries (missing `_meta`) are skipped.

**Low severity — worth noting, not blocking**
- `POST /api/decksFromLink` is unauthenticated and each call triggers a remote download + BSON conversion that the UI itself warns "can take a while for big games." An anonymous client can spam it to pin CPU/network. This matches the existing (also-unauthenticated) link-import surface, so it's not a regression — just flagging that the new endpoint widens that surface with no rate limiting.
- `addDeckWithCards` calls `batchStart()`/`batchEnd()` without a try/finally; if `addWidgetLocal` throws mid-loop the batch is left open. The new `addButton.onclick` catches the error for the alert, but the dangling batch is a pre-existing property of the shared helper, not introduced here. Mentioning only because this PR adds a second caller.

**Test coverage**
- No jest/testcafe coverage was added for the new endpoint (the underlying TTS conversion is already covered). Given the endpoint is mostly a thin adapter over existing, tested pipeline code this is acceptable, but a small unit test asserting the deck/`cardCounts` shape for a fixture variant would guard against future regressions in the filtering logic.

## Beta label

I left the existing `beta` label in place: this PR touches no engine files (`client/js/widgets/`, `compute.js`, `serverstate.js`) — only an additive server endpoint, edit-mode sidebar UI and CSS — and it is safe to test (SSRF-guarded, no game-corruption or crash risk), so it is beta-eligible and ready for player testing.

96LawDawg (approved):
Approved.

PR #3025, opened 2026-07-11
Add a public library deck browser to the editor

VTT-AI-Agent:
## What

Adds a **Library decks** button to the *Cards* category of the editor's add widget overlay. It opens a new overlay that lets you browse **every deck in the public library** (currently ~1200 decks across games and tutorials), grouped by game, with a filter box, and add any of them to the current game with one click.

The overlay shows each deck with up to three sample cards rendered by the real `Deck`/`Card` widget classes (the same trick the add widget overlay already uses for its composite widgets), plus the deck's id, card count and card type count.

Clicking a deck adds a copy of it to the game in the familiar shape of the existing "add deck" entry: a holder sized to the deck's cards, the deck widget, a pile containing all its cards, and a *Recall & Shuffle* button.

## How

**Server** (`server/librarydecks.mjs`, new):
- `GET /api/library/decks` returns a lightweight index (~150 KB) of all decks: game, file, deck id, card/type counts and card dimensions. Decks that appear unchanged in several variants of the same game are deduplicated; decks without cards are skipped.
- `GET /api/library/decks/:library/:game/:file/:deck` returns one deck's full definition plus its card widgets, with source-game-specific properties (`id`, `parent`, `x`/`y`/`z`, `owner`, …) stripped. Only decks present in the index can be requested, so no path traversal.
- **The cache is lazy**: nothing is scanned at server startup — the index is built on the first request (~1 s, async so the event loop keeps breathing) and kept in memory afterwards (second request: ~20 ms). It is invalidated when the public library is edited at runtime (`allowPublicLibraryEdits`).

**Client**:
- The overlay fetches the index only when first opened, and deck previews are only fetched and rendered when their entry scrolls into view (`IntersectionObserver` with a 300 px margin), so opening the overlay stays cheap. Deck details are cached client-side and reused when the deck is added.
- Card images resolve automatically because `/assets/<name>` already serves public library assets.

No widget properties, routine operations or state semantics change, so no validator or FileUpdater changes are needed.

## Testing

- jest suite passes.
- TestCafe `editor.js` ("Create game using edit mode") passes locally — it exercises the add widget overlay whose Cards panel this touches.
- Verified end-to-end with a scripted browser: opened the overlay (1225 decks listed, previews lazy-render), filtered, clicked Agon's `hexDeck` → holder + deck + pile + 91 cards + shuffle button appear in the room and in the room state.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3025/pr-test (or any other room on that server)


96LawDawg:
@VTT-AI-Agent, add the ability to sort by "stars" or "popularity" like the public library game shelf filters.

VTT-AI-Agent (commented):
## TL;DR

Clean addition of an editor UI for browsing and adding decks from the public library. No engine changes, good security (path traversal prevention via index lookup, XSS escaping), efficient (lazy caching + IntersectionObserver preview loading), and tests pass. Beta-ready.

## Implementation Quality

**Security**: Decks are served only from a pre-built index (`librarydecks.mjs`), so path traversal via URL parameters is prevented. Deck names are properly escaped with `html()` when rendered, and URLs are encoded with `encodeURIComponent()`. ✓

**Performance**: 
- Server-side: Lazy index building on first request (~1s async), then cached in memory; invalidated when library is edited. Good strategy.
- Client-side: Deck index fetched only on overlay open; preview rendering lazy-loads with `IntersectionObserver` (300px margin), so opening the overlay stays cheap.

**Correctness**:
- Temporary preview widgets (`Deck`, `Card`) are properly created and deleted from the `widgets` map after rendering—no leaks. The DOM elements remain in the preview container (as intended) but aren't managed by the engine.
- Batch operations (`batchStart/batchEnd`) correctly group the full deck+cards+button additions with a proper delta cause for logging.
- Unique ID generation (`generateUniqueWidgetID` + suffix collision check) is thorough.
- Re-validation of deck file before serving (`state[deck] && state[deck].type == 'deck'`) is good defensive programming.

**Deduplication & Filtering**:
- Identical decks across game variants are deduplicated by comparing `cardDefaults`, `cardTypes`, `faceTemplates`.
- Empty decks (no cards) are skipped.
- Filter uses simple substring matching on game name and deck ID, which is appropriate for the UI.

## Engine Impact

No engine changes: doesn't touch widget rendering/behavior (`client/js/widgets/`), state sync (`serverstate.js`), expression evaluation (`compute.js`), or widget property definitions. This is purely editor UI and library-serving. No validator or FileUpdater changes needed (as documented). ✓

## Minor Observations

- Card order from the source deck isn't explicitly preserved (relies on `Object.values` iteration order), but for a deck that's typically shuffled, this is acceptable.
- No new widget properties or routine operations added, so no validator impact (correct per the PR description).

⟦FINAL-SUMMARY⟧

No issues found. The implementation is clean and well-thought-out: proper security (indexed lookup + escaping), good performance (lazy caching and preview loading), and no engine impact. Tests pass. **Added `beta` label** — this is an editor UI feature only, safe for player testing and won't corrupt existing games.

VTT-AI-Agent:
## TL;DR

Clean addition of an editor UI for browsing and adding decks from the public library. No engine changes, good security (path traversal prevention via index lookup, XSS escaping), efficient (lazy caching + IntersectionObserver preview loading), and tests pass. Beta-ready.

## Implementation Quality

**Security**: Decks are served only from a pre-built index (`librarydecks.mjs`), so path traversal via URL parameters is prevented. Deck names are properly escaped with `html()` when rendered, and URLs are encoded with `encodeURIComponent()`. ✓

**Performance**: 
- Server-side: Lazy index building on first request (~1s async), then cached in memory; invalidated when library is edited. Good strategy.
- Client-side: Deck index fetched only on overlay open; preview rendering lazy-loads with `IntersectionObserver` (300px margin), so opening the overlay stays cheap.

**Correctness**:
- Temporary preview widgets (`Deck`, `Card`) are properly created and deleted from the `widgets` map after rendering—no leaks. The DOM elements remain in the preview container (as intended) but aren't managed by the engine.
- Batch operations (`batchStart/batchEnd`) correctly group the full deck+cards+button additions with a proper delta cause for logging.
- Unique ID generation (`generateUniqueWidgetID` + suffix collision check) is thorough.
- Re-validation of deck file before serving (`state[deck] && state[deck].type == 'deck'`) is good defensive programming.

**Deduplication & Filtering**:
- Identical decks across game variants are deduplicated by comparing `cardDefaults`, `cardTypes`, `faceTemplates`.
- Empty decks (no cards) are skipped.
- Filter uses simple substring matching on game name and deck ID, which is appropriate for the UI.

## Engine Impact

No engine changes: doesn't touch widget rendering/behavior (`client/js/widgets/`), state sync (`serverstate.js`), expression evaluation (`compute.js`), or widget property definitions. This is purely editor UI and library-serving. No validator or FileUpdater changes needed (as documented). ✓

## Minor Observations

- Card order from the source deck isn't explicitly preserved (relies on `Object.values` iteration order), but for a deck that's typically shuffled, this is acceptable.
- No new widget properties or routine operations added, so no validator impact (correct per the PR description).

⟦FINAL-SUMMARY⟧

No issues found. The implementation is clean and well-thought-out: proper security (indexed lookup + escaping), good performance (lazy caching and preview loading), and no engine impact. Tests pass. **Added `beta` label** — this is an editor UI feature only, safe for player testing and won't corrupt existing games.

VTT-AI-Agent:
Fixed a bug reported by a beta-server player:

> This leaves phantom deck widgets in the game (even without adding anything).

**Cause:** `applyInitialDelta` appends every widget to `#topSurface` via its `null` parent default. The deck-preview code moved its sample *cards* into the overlay but never re-homed the throwaway *deck* widget it creates as a data source for the cards' faces — so each rendered preview left a phantom deck counter sitting in the room, just from scrolling the deck browser.

**Fix** (`8520f5f9`): remove the temporary deck's DOM element once the cards' faces are built.

Verified with a scripted browser: opening the browser and scrolling through all previews now leaves `0` deck/card elements in `#topSurface`, while adding a deck still produces the correct holder + deck + pile + cards + shuffle button.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-scoped feature: a public-library deck browser in the editor's Cards panel, backed by two read-only server endpoints. The preview/add mechanics faithfully reuse existing editor patterns and the path-traversal surface is properly closed (only decks present in the built index are readable). No correctness or security blockers found — just a few minor observations below.

## Findings

### Correctness / robustness (minor)

- **Phantom widgets on a mid-render throw** — `renderLibraryDeckPreview` (`client/js/editmode.js`) registers the deck and card widgets into the global `widgets` map and only removes them at the end of the synchronous block. If a `Card`/`Deck` constructor or `applyInitialDelta` throws partway through the `sampleTypes.forEach`, the already-created entries (deck + earlier cards) are left in `widgets` and, for the deck, in `#topSurface`. The outer `.catch` only sets the preview text; it doesn't clean up. Low likelihood (matches the existing `addCompositeWidgetToAddWidgetOverlay` risk profile), but a `try/finally` around the create→remove block would make it airtight.

- **Deck dedup key ignores the actual card set** — in `server/librarydecks.mjs`, `duplicateKey = JSON.stringify([cardDefaults, cardTypes, faceTemplates])`. Two decks in the same game that share those three fields but ship *different card widgets* (different quantities/subsets) collapse to a single catalog entry, and `getDeck` always serves the cards from the first-seen file. This is intentional dedup and fine for the common "same deck across game variants" case, but it means a game that reuses a deck definition with a genuinely different card subset would only surface one variant. Worth a one-line comment noting the assumption.

### Nits (non-blocking)

- `getDeck` re-reads and re-parses the whole source file on every request; the client-side `libraryDeckDetailsCache` mitigates repeat hits, so this is acceptable, but a small server-side per-deck cache would be cheap if this ever gets hammered.
- Each keystroke in the filter box fully rebuilds the list and re-observes every entry, re-rendering previews for visible rows (from the cached details, so no refetch). Fine at current scale; a short debounce would smooth it if the list grows.

### Things I checked that are fine

- **No path traversal**: `getDeck` resolves the source path from the `sources` map keyed by the exact `library/game/file/deck` tuple built during the scan; unmatched params return 404, and the scan only ever indexes `Config.directory('library')` entries with file names matching `^[0-9]+\.json$`. User input never reaches `fs` directly.
- **XSS**: user-influenced strings (`gameName`, `deck`) are escaped via `html()`; counts are server-side numbers.
- **Holder ID collision**: `generateUniqueWidgetID()` guarantees `id` itself is free, and the added `suffixes.some(...)` loop covers the derived `B/D/P/C*` IDs — so all five+ widgets get unique IDs.
- **Preview cleanup**: the deck widget's DOM is removed and all temp widgets are deleted from the map, consistent with the established `addCompositeWidgetToAddWidgetOverlay` approach; no `await` sits between `widgets.set` and `widgets.delete`, so no interleaving window.
- No widget properties, routine ops, or state semantics change, so the skipped validator/FileUpdater work is correct.

## Beta label

The `beta` label is already present, and correctly so: this is **not** an engine change — the only engine-directory file touched is `server.mjs`, and its edits are incidental (two new read-only `/api/library/decks` endpoints plus a cache-invalidation call), while `editmode.js` is editor-only code that leaves default widget behavior and existing-game state semantics untouched; the change is safe and useful for player testing, so I'm leaving the label in place.

96LawDawg:
@VTT-AI-Agent, add the ability to sort by "stars" or "popularity" like the public library game shelf filters. Also, add an option to open the deck browser when the Properties menu is open without a widget selected, like in the image.

image

VTT-AI-Agent:
Both done in d7f22c45, @96LawDawg 👍

**1. Sort by stars / popularity** — added a *Sort by* dropdown to the deck browser with **Name / Stars / Popularity**. Game groups are reordered as a whole (decks keep their in-game order), using the same star and play-time statistics the game shelf uses, and each game now shows a ★ star-count badge. The stats are computed fresh on every `/api/library/decks` request so they stay current while the deck catalog itself stays cached.

**2. Open from the Properties panel** — when the Properties sidebar is open with nothing selected (the *Add deck* panel), there's now a **Browse public library decks** button above the *Or create a new deck* section, which opens the same browser.

Verified end-to-end with a scripted browser (test statistics injected locally): sorting by stars ordered games 100★ → 42★ → 7★, popularity ordered by play time (a different order), and the Properties-panel button opens the overlay. jest + the `editor.js` TestCafe test pass.

96LawDawg:
Fix:
- Changing the sort by option does not actually change how the decks are sorted. They always sort alphabetically.
- In the text area that says "Click a deck to add a copy of it - including a holder, all its cards and a shuffle button - to your game", add the following new sentence: "Some decks appear differently than in their game due to CSS or fonts not available within the deck widget itself.

VTT-AI-Agent:
Thanks @96LawDawg — done in fa6e2c5e.

**Rendering caveat** — added your sentence to the help text: *"Some decks appear differently than in their game due to CSS or fonts not available within the deck widget itself."*

**Sorting** — I dug into this. The dropdown *is* wired up and does re-sort; game groups are reordered by the same star/play-time statistics the game shelf uses (same `Statistics` keys, e.g. `games/Agon`). What it looks like depends on the data:

- Sort keys are per-game **stars** and **play time**. When those are all equal (e.g. a fresh test/preview server that hasn't accrued stars or plays yet), every game ties and it falls back to the alphabetical tiebreak — so it looks like nothing changed. This is the same behavior the game shelf shows on a server with no star data.
- On a server that has star/play data it reorders. I verified end-to-end with statistics injected locally: sorting by **Stars** gave `Alpaca ★100` → `Agon ★42` → `Agra ★7`, and by **Popularity** gave a different order driven by play time (`Agon` first). Only games that contain card decks appear here, so games without decks (many high-star board games) won't show up to sort.

I also found a real gap while checking: the browser cached the deck list for the whole session, so newly-earned stars wouldn't show until reload. It now **refetches on every open**, so the star/popularity counts are always current.

If you're testing on the PR preview and it still looks alphabetical, it's almost certainly because that server has no star/play data yet for the deck-containing games. Happy to tweak the default sort or add an indicator if you'd prefer.

96LawDawg:
Fix:

- After opening and closing the deck selector, these strange widget artifacts (see image) are sometimes added to the room.
- Add a close (x) button to the deck selector.

image

96LawDawg:
Also, do not show any decks that are the standard 52 card French suited deck.

VTT-AI-Agent:
Both fixed in fc1c0ff5, @96LawDawg 🙏

**Widget artifacts** — good catch, this was a real leak. To render a preview I build throwaway `Deck`/`Card` widgets, and `applyInitialDelta` appends every widget to `#topSurface` (its parent defaults to null) before I move the cards into the overlay. Some library decks have cards whose `cardType` isn't in the deck's `cardTypes`, which makes `Card.applyInitialDelta` throw **before** the cleanup ran — leaving the throwaway deck and card sitting in the room. Two changes:
- Sample the preview cards from the deck's own `cardTypes` keys, so the type always exists and the render doesn't throw.
- Move all cleanup into a `finally` block, so even if a preview fails for any other reason, nothing is left behind.

Verified by scrolling through the whole catalog: previously 23 decks + 23 cards leaked into the room; now `#topSurface` and the widget map stay empty, and all 1219 previews render (0 failures, was 23).

**Close button** — added an × button to the top-right of the deck browser.

VTT-AI-Agent:
Done in 40ab93d6, @96LawDawg.

The deck browser now skips any deck that is just the standard 52-card French-suited deck — i.e. every card type uses a built-in `/i/cards-default/` image and they cover exactly the 52 standard cards (jokers allowed). Since the editor can already add that deck, listing library copies of it was only noise.

This hides **57** such decks (e.g. Hearts, Spades, Solitaire) and leaves all custom decks. Any deck that mixes in custom card images or uses a different set is kept.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-scoped addition — a public library deck browser bolted onto the editor's add-widget overlay, with no engine files touched at all (only `client/js/editmode.js`, `client/editor.html`, `client/css/editmode.css`, `client/js/editor/sidebar/properties.js`, `server.mjs`, and the new `server/librarydecks.mjs`). I didn't find a correctness or security bug worth blocking on; the one gap worth flagging is the total absence of automated test coverage for the new server-side dedup/index logic (`buildCache`, `isStandardFrenchDeck`), though that's consistent with this repo's very thin existing server test suite.

## Details

**Security / path traversal** — `GET /api/library/decks/:library/:game/:file/:deck` (server/librarydecks.mjs:644-658, server.mjs:478-485) uses the request params only as a lookup key into the in-memory `sources` map built from real filesystem listings during `buildCache()`; it never concatenates request params directly into a filesystem path. That correctly rules out path traversal, and the composite string key (`${library}/${game}/${file}/${deck}`) also isn't exploitable for prototype pollution against the plain `sources` object. Confirmed the `libraries` config entries used here (`Games`/`Tutorials`) are the same public-only directories `server/room.mjs`'s public library scan already exposes, so no private/save data is reachable through this endpoint.

**Consistency check** — the holder sizing formula in the new `addLibraryDeckToGame` (`cardWidth + 8` / `cardHeight + 11`, client/js/editmode.js:366-367) reproduces the existing standard-deck holder dimensions (111×171.36 for the 103×160 default deck, from `generateCardDeckWidgets` at the top of the same file) almost exactly, so the new deck-add path stays visually consistent with the pre-existing "add deck" flow. The transient preview widgets in `renderLibraryDeckPreview` (client/js/editmode.js:291-344) are cleaned up correctly in the `finally` block: the throwaway `Deck` is always removed from `#topSurface`, and cards that were successfully moved into the preview wrapper are correctly left alone (they're no longer `topSurface` children) rather than double-removed.

**Minor / non-blocking**
- No jest coverage was added for the new server module's dedup logic (`seenInGame`/`duplicateKey`) or `isStandardFrenchDeck` — both have some non-trivial branching (e.g., filtering French-suited images by regex, joker exclusion) that would be easy to unit test in isolation without a full server/browser harness. Not a blocker given `tests/` currently only has two client-side jest specs and no server-side tests at all to match convention against.
- `openLibraryDecksOverlay` refetches the whole ~150 KB index every time the overlay is opened (by design, to keep star/popularity counts fresh) — reasonable tradeoff, just flagging it as a possible future optimization (e.g. a separate lightweight stats-only endpoint) if the index grows much larger.

No engine files (`client/js/widgets/`, `client/js/compute.js`, `client/js/serverstate.js`) are touched, and existing game state/routine semantics are unaffected.

---
The PR already had the `beta` label applied when I reviewed it, and I'm leaving it in place — the change is entirely new, opt-in editor UI plus a new read-only public-library API endpoint, doesn't touch engine files, and I found no crash or data-corruption risk that would make it unsafe for beta testers.

96LawDawg (approved):
Approved.

PR #3027, opened 2026-07-11
Prevent client stack overflow on cyclic widget parent chains

VTT-AI-Agent:
## What

Fixes a client crash reported from the main server: `RangeError: Maximum call stack size exceeded` with `coordLocalFromCoordGlobal` and `coordParentFromCoordGlobal` alternating endlessly on a holder widget.

## Cause

Nothing prevents the room state from containing a **cycle in the `parent` chain** (e.g. two widgets that are each other's parent, or a widget that is its own parent) — a routine that sets the `parent` property can create one. Several client methods walk the parent chain by recursion and never terminate on such a state, crashing on every mouse move or drag once the room is in that condition.

## Fix

- New cycle-safe `Widget.ancestors()` helper that walks up the parent chain and stops when it sees a widget twice.
- All parent-chain walks now use it (or an equivalent guarded loop): `coordLocalFromCoordGlobal`, `coordGlobalFromCoordLocal`, `_absoluteRotation`, `_absoluteScale`, `_ancestor`, `isDescendantOf`, `inheritSeatVisibility`, `requiresHiddenCursor`, `getValidDropTargets`, the mouse-down movable-ancestor lookup and the JSON editor's `hiddenParent`. Behavior is unchanged for acyclic states.
- Applying a delta that creates a cycle no longer breaks the client halfway through `receiveDelta`:
  - `getElementTransformRelativeTo` returns `null` instead of walking past the DOM root when the target element is a descendant of the source.
  - `applyDeltaToDOM` cannot mirror a cyclic parent into the DOM, so the widget is shown as a *limbo* widget (the existing "Invalid Parent" treatment) instead of throwing `HierarchyRequestError` on `appendChild`.
- The runtime `parent`/`childArray` graph stays acyclic even on the client that sets the property: `onChildAdd` refuses an edge whose child already is an ancestor. Tearing the room down survives a cyclic graph regardless: `applyRemoveRecursive` and `removeWidgetLocal` track what they already visited, so a full state message (load, restore, resync) still replaces the room.
- Recursive duplication in edit mode and `renderReadonlyCopy` walk the state parent chain too and hung the tab on a cyclic room; both skip widgets they already copied, and duplication additionally ignores the copies it just created — on a cyclic chain a copy can otherwise come back to it as a child of a widget that is still to be copied.

## Verification

- Reproduced the exact reported stack trace on `main` with a Playwright-driven room where a button routine sets one holder's `parent` to a holder inside it; hovering/dragging then overflows the stack.
- With this PR, the same scenario produces **zero** client errors: the delta applies cleanly, the cyclic widgets render as limbo ("Invalid Parent", hidden outside edit mode), dragging a card over/onto them works, and a fresh page load of the cyclic room also survives (widgets deferred, as before).
- Two clients in one room: the client that runs the cycle-creating routine keeps an acyclic `childArray` and loads the next full state normally; before, it overflowed the stack and stayed on the stale room.
- Edit mode: clone-dragging a cyclic holder terminates (it used to hang the tab); a control clone-drag of a normal three-level tree still copies the whole subtree exactly once.
- `tests/client/widget-cyclic-parent.test.js` covers all of it (parent-chain walks, drop targets, `DELETE`, cyclic deltas, cycles created through `set('parent')`, recursive readonly copies). `tests/client/setup.js` gained minimal `DOMMatrix`/`DOMPoint` stand-ins because jsdom implements neither and `set('parent')` needs them.
- jest passes; TestCafe `editor.js` and `publiclibrary-3/4` pass locally, `publiclibrary-1/2` pass except `Reversi`, which is nondeterministic on this machine on the unmodified branch as well.

Repro scripts and logs: <https://agent.virtualtabletop.io/reports/pr/1525330699429871778/repro/README.md>
A demo room named **Cyclic parent chains** is on this PR's test server (`pr-test-ai`) — one button makes a holder a child of the holder inside it; the board has to stay usable afterwards. It lives on the test server only, not in `library/`, and can be added to `library/tutorials/` on request.

No creator-facing property or routine semantics change.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3027/pr-test (or any other room on that server)

### DIFF
--- a/client/js/editmode.js
+++ b/client/js/editmode.js
@@ -1308,7 +1308,14 @@ async function duplicateWidget(widget, recursive, inheritFrom, inheritProperties
       return l.substr(0, l.length-zs-1) + String.fromCharCode(l.charCodeAt(l.length-zs-1)+1) + [...Array(zs)].map(l=>'A').join('');
   };
 
-  const clone = async function(widget, recursive, newParent, xOffset, yOffset) {
+  // on a cyclic parent chain a copy can end up as the child of a source widget that is copied later, so the copies are never cloned again themselves
+  const createdIDs = new Set();
+
+  const clone = async function(widget, recursive, newParent, xOffset, yOffset, copied=new Set()) {
+    if(copied.has(widget)) // a cyclic parent chain would make the recursive copy revisit the same source widgets forever
+      return [];
+    copied.add(widget);
+
     let currentWidget = JSON.parse(JSON.stringify(widget.state))
 
     if(inheritFrom) {
@@ -1376,11 +1383,12 @@ async function duplicateWidget(widget, recursive, inheritFrom, inheritProperties
         problems.push(`Could not add duplicate of card ${widget.id} with non-existent cardType ${currentWidget.cardType}.`);
     } else {
       const currentId = await addWidgetLocal(currentWidget);
+      createdIDs.add(currentId);
 
       const clonedWidgets = [ widgets.get(currentId) ];
       if(recursive)
-        for(const child of widgetFilter(w=>w.get('parent')==widget.id))
-          clonedWidgets.push(...await clone(child, true, currentId, 0, 0));
+        for(const child of widgetFilter(w=>w.get('parent')==widget.id && !createdIDs.has(w.get('id'))))
+          clonedWidgets.push(...await clone(child, true, currentId, 0, 0, copied));
 
       return clonedWidgets;
     }
--- a/client/js/geometry.js
+++ b/client/js/geometry.js
@@ -103,6 +103,8 @@ export function getElementTransformRelativeTo(elem, parent) {
   let transform = getElementTransform(elem);
   while (elem.offsetParent != ancestor) {
     elem = elem.offsetParent;
+    if (!elem)
+      return null; // parent is a descendant of elem so there is no path up to it
     transform.preMultiplySelf(getElementTransform(elem));
   }
   let destTransform = new DOMMatrix();
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -3808,7 +3808,7 @@ function jeInitEventListeners() {
 
     hoveredWidgets.sort(function(w1,w2) {
       const hiddenParent =  function(widget) {
-        return widget ? widget.domElement.classList.contains('foreign') || hiddenParent(widgets.get(widget.get('parent'))) : false;
+        return widget ? widget.ancestors().some(w=>w.domElement.classList.contains('foreign')) : false;
       };
       const w1card = w1.get('type') == 'card';
       const w2card = w2.get('type') == 'card';
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -36,7 +36,7 @@ function compareDropTarget(widget, t, exclude){
   return false;
 }
 
-function getValidDropTargets(widget) {
+export function getValidDropTargets(widget) {
   const targets = [];
   for(const [ _, t ] of dropTargets) {
     if(!t.isVisible())
@@ -50,18 +50,8 @@ function getValidDropTargets(widget) {
 
     let isValid = compareDropTarget(widget, t);
 
-    let tt = t;
-    while(isValid) {
-      if(widget == tt) {
-        isValid = false;
-        break;
-      }
-
-      if(tt.get('parent'))
-        tt = widgets.get(tt.get('parent'));
-      else
-        break;
-    }
+    if(isValid && t.ancestors().includes(widget))
+      isValid = false;
 
     if (jeEnabled && getComputedStyle(t.domElement).getPropertyValue('--foreign') == 'true')
       continue;
--- a/client/js/mousehandling.js
+++ b/client/js/mousehandling.js
@@ -79,18 +79,9 @@ async function inputHandler(name, e) {
         moveTarget: widget
       };
       const ms = mouseStatus[target.id];
-      let movable = ms.moveTarget.get(editMovable ? 'movableInEdit' : 'movable');
-      while (ms.moveTarget && !movable) {
-        let parent = ms.moveTarget.get('parent');
-        if(parent && widgets.has(parent)) {
-          ms.moveTarget = widgets.get(parent);
-          movable = ms.moveTarget.get(editMovable ? 'movableInEdit' : 'movable');
-        } else {
-          ms.moveTarget = null;
-          movable = false;
-        }
-      }
-      if (movable) {
+      const movableProperty = editMovable ? 'movableInEdit' : 'movable';
+      ms.moveTarget = ms.moveTarget.ancestors().find(w=>w.get(movableProperty));
+      if (ms.moveTarget) {
         ms.localAnchor = ms.moveTarget.coordLocalFromCoordClient({x: coords.clientX, y: coords.clientY});
       }
     } else if(name == 'mouseup' || (name == 'touchend' || name == 'touchcancel') && mouseStatus[target.id]) {
--- a/client/js/serverstate.js
+++ b/client/js/serverstate.js
@@ -474,8 +474,9 @@ function receiveStateFromServer(args) {
   mouseTarget = null;
   deltaID = args._meta.deltaID;
   const topSurface = $('#topSurface');
+  const removed = new Set();
   for(const widget of widgetFilter(w=>w.domElement.parentElement === topSurface))
-    widget.applyRemoveRecursive();
+    widget.applyRemoveRecursive(removed);
   widgets.clear();
   dropTargets.clear();
   maxZ = {};
--- a/client/js/serverstate.js
+++ b/client/js/serverstate.js
@@ -566,14 +566,14 @@ function removeWidget(widgetID) {
   dropTargets.delete(widgetID);
 }
 
-async function removeWidgetLocal(widgetID, keepChildren) {
+export async function removeWidgetLocal(widgetID, keepChildren) {
   function getWidgetsToRemove(widgetID) {
+    widgets.get(widgetID).inRemovalQueue = true;
     const children = [];
     if(!keepChildren)
       for(const [ childWidgetID, childWidget ] of widgets)
         if(!childWidget.inRemovalQueue && (childWidget.get('parent') == widgetID || childWidget.get('deck') == widgetID))
           children.push(...getWidgetsToRemove(childWidgetID));
-    widgets.get(widgetID).inRemovalQueue = true;
     children.push(widgets.get(widgetID));
     return children;
   }
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -354,9 +354,13 @@ export class Widget extends StateManaged {
     this.globalUpdateListenersUnregister();
   }
 
-  applyRemoveRecursive() {
+  applyRemoveRecursive(removed=new Set()) {
+    // tracking the already removed widgets keeps tearing down the room terminating even if the child graph became cyclic
+    if(removed.has(this))
+      return;
+    removed.add(this);
     for(const child of Widget.prototype.children.call(this)) // use Widget.children even for holders so it doesn't filter
-      child.applyRemoveRecursive();
+      child.applyRemoveRecursive(removed);
     this.applyRemove();
   }
 
@@ -2539,6 +2543,9 @@ export class Widget extends StateManaged {
   }
 
   async onChildAdd(child, oldParentID) {
+    // a child that already is an ancestor would close a cycle in the child graph which recursive walks like applyRemoveRecursive could not terminate on
+    if(this.ancestors().includes(child))
+      return;
     this.childArray = this.childArray.filter(c=>c!=child);
     this.childArray.push(child);
     await this.onChildAddAlign(child, oldParentID);
@@ -2605,14 +2612,15 @@ export class Widget extends StateManaged {
     return this;
   }
 
-  renderReadonlyCopy(propertyOverride, target, includeChildren=false, isChild=false) {
+  renderReadonlyCopy(propertyOverride, target, includeChildren=false, isChild=false, copied=new Set()) {
     const newID = generateUniqueWidgetID();
     const newWidget = new this.constructor(newID);
     newWidget.renderReadonlyCopyRaw(Object.assign({}, this.state, propertyOverride), target, isChild);
+    copied.add(this);
     if(includeChildren)
       for(const child of widgetFilter(w=>w.get('parent') == this.id))
-        if(this.get('type') != 'holder' || !compareDropTarget(child, this) || includeChildren == 'all')
-          child.renderReadonlyCopy({}, newWidget.domElement, includeChildren, true);
+        if(!copied.has(child) && (this.get('type') != 'holder' || !compareDropTarget(child, this) || includeChildren == 'all'))
+          child.renderReadonlyCopy({}, newWidget.domElement, includeChildren, true, copied);
     return newWidget;
   }
 
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -130,6 +130,16 @@ export class Widget extends StateManaged {
     return this.coordGlobalFromCoordParent({x:this.get('x'),y:this.get('y')})[coord]
   }
 
+  ancestors() {
+    // walks up the parent chain (starting with the widget itself) and stops on
+    // cycles which can appear in the room state (e.g. two widgets being each
+    // other's parent) and would otherwise cause infinite recursion
+    const chain = [];
+    for(let w = this; w && !chain.includes(w); w = widgets.get(w.get('parent')))
+      chain.push(w);
+    return chain;
+  }
+
   animateProperties() {
     return asArray(JSON.parse(JSON.stringify(this.get('animatePropertyChange'))));
   }
@@ -181,7 +191,11 @@ export class Widget extends StateManaged {
     let newParent = undefined;
     if(delta.parent !== undefined) {
       newParent = delta.parent && widgets.has(delta.parent) ? widgets.get(delta.parent).domElement : $('#topSurface');
-      this.setLimbo(delta.parent && !widgets.has(delta.parent));
+      // the DOM cannot represent cyclic parent chains so if the new parent is inside this widget, display it at the top level as a limbo widget
+      const cyclicParent = this.domElement.contains(newParent);
+      if(cyclicParent)
+        newParent = $('#topSurface');
+      this.setLimbo(delta.parent && (!widgets.has(delta.parent) || cyclicParent));
       // If the widget wasn't newly created, transition from its previous location.
       if (delta.id === undefined)
         fromTransform = getElementTransformRelativeTo(this.domElement, newParent);
@@ -527,7 +541,9 @@ export class Widget extends StateManaged {
   }
 
   coordGlobalFromCoordLocal(coord) {
-    return this.coordGlobalFromCoordParent(this.coordParentFromCoordLocal(coord));
+    for(const widget of this.ancestors())
+      coord = widget.coordParentFromCoordLocal(coord);
+    return coord;
   }
   coordGlobalFromCoordParent(coord) {
     const p = this.get('parent');
@@ -542,7 +558,9 @@ export class Widget extends StateManaged {
     return result || new DOMPoint();
   }
   coordLocalFromCoordGlobal(coord) {
-    return this.coordLocalFromCoordParent(this.coordParentFromCoordGlobal(coord));
+    for(const widget of this.ancestors().reverse())
+      coord = widget.coordLocalFromCoordParent(coord);
+    return coord;
   }
   coordLocalFromCoordParent(coord) {
     const result = getPointOnPlane(getElementTransform(this.domElement), coord.x, coord.y);
@@ -2193,18 +2211,24 @@ export class Widget extends StateManaged {
     if(!readOnlyProperties.has(property)) {
       return super.get(property);
     } else {
-      const p = this.get('parent');
       switch(property) {
         case '_absoluteRotation':
-          return this.get('rotation') + (widgets.has(p)? widgets.get(p).get('_absoluteRotation') : 0);
+          return this.ancestors().reduce((sum, w)=>sum + w.get('rotation'), 0);
         case '_absoluteScale':
-          return this.get('scale') * (widgets.has(p)? widgets.get(p).get('_absoluteScale') : 1);
+          return this.ancestors().reduce((product, w)=>product * w.get('scale'), 1);
         case '_absoluteX':
           return this.coordGlobalFromCoordParent({x:this.get('x'),y:this.get('y')})['x'];
         case '_absoluteY':
           return this.coordGlobalFromCoordParent({x:this.get('x'),y:this.get('y')})['y'];
-        case '_ancestor':
-          return (widgets.has(p) && widgets.get(p).get('type')=='pile') ? widgets.get(p).get('_ancestor') : p;
+        case '_ancestor': {
+          let w = this;
+          for(const a of this.ancestors().slice(1)) {
+            if(a.get('type') != 'pile')
+              break;
+            w = a;
+          }
+          return w.get('parent');
+        }
         case '_centerAbsoluteX':
           return this.coordGlobalFromCoordParent({x:this.get('x')+this.get('width')/2,y:this.get('y')+this.get('height')/2})['x'];
         case '_centerAbsoluteY':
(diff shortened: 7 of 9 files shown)


VTT-AI-Agent (commented):
## TL;DR
Solid, well-scoped fix for a real crash (cyclic `parent` chains overflowing the stack). I traced the refactored `ancestors()`-based traversal against every replaced recursive method (`coordGlobalFromCoordLocal`, `coordLocalFromCoordGlobal`, `_absoluteRotation/_absoluteScale`, `_ancestor`, `inheritSeatVisibility`, `isDescendantOf`, `requiresHiddenCursor`) and the logic is equivalent for acyclic states while now terminating on cycles. No blocking bugs found. Main gap: no automated test covers the cyclic scenario, so a regression here wouldn't be caught by CI.

## Findings

- **No test coverage for the fix** (jest or TestCafe). The PR description mentions a manual Playwright repro but nothing lands in `tests/`. Given this is a crash fix for a specific edge case (self-parent / mutual-parent cycles), a small jest test for `Widget.ancestors()` terminating on a manufactured cycle, or a TestCafe scenario that sets a cyclic `parent` via routine and asserts no client error, would guard against this regressing silently. Worth adding per `docs/testcafe.md`'s guidance on state-hash regression coverage.

- **Partial guarding left in `coordGlobalFromCoordParent`/`coordParentFromCoordGlobal`** (`client/js/widgets/widget.js:532-535`, `551-554`) — these two helpers still recurse via `widgets.get(p).coordGlobalFromCoordLocal(coord)` / `...coordLocalFromCoordGlobal(coord)` unguarded for one hop before landing in the now-safe `ancestors()`-based methods. I confirmed this can't reintroduce unbounded recursion (the callee is the guarded version), but it's worth a one-line comment noting why these two remain "recurse once then guarded" instead of also being ancestor-based, so a future editor doesn't assume they're unsafe leftovers (or refactor them inconsistently).

- `getElementTransformRelativeTo` (`client/js/geometry.js:103-107`) returning `null` on a cyclic case — I checked the caller `applyDeltaToDOM` handles the null via the existing "limbo" fallback path rather than dereferencing it, so that part looks correct.

Everything else — `ancestors()`'s cycle-stop condition, the `_ancestor` pile-walk rewrite, `requiresHiddenCursor`'s parent/hoverParent precedence, and `isDescendantOf` — checked out as behavior-preserving for normal (acyclic) rooms.

Beta label: this PR touches `client/js/widgets/widget.js` and `client/js/geometry.js` (core engine files) and genuinely changes engine behavior — the parent-chain traversal used by every widget on every frame — not just incidentally for new opt-in code, so per the beta-eligibility policy for engine changes I did **not** add the `beta` label.

VTT-AI-Agent:
Thanks for the review — addressed both points in 77598b17:

- **Test coverage:** added `tests/client/widget-cyclic-parent.test.js`, a jest regression test that manufactures cyclic parent chains (a mutual `w1 ⇄ w2` cycle and a self-referencing widget) directly in state and asserts that `ancestors()` and the parent-chain traversal methods (`_absoluteRotation`, `_absoluteScale`, `_ancestor`, `isDescendantOf`, `inheritSeatVisibility`, `requiresHiddenCursor`) terminate rather than overflowing the stack. The cycle is built via direct state rather than `set('parent')` because the coordinate helpers rely on `DOMMatrix`, which jsdom doesn't implement; the recursion guard under test is independent of that geometry. The full jest suite passes (3 suites, 8 tests).

- **Clarifying comment:** added a one-line comment on both `coordGlobalFromCoordParent` and `coordParentFromCoordGlobal` noting they intentionally recurse one hop into the now cycle-safe `coordGlobalFromCoordLocal`/`coordLocalFromCoordGlobal`, so a cyclic chain terminates there rather than being unguarded leftovers.

- **beta label:** agreed, leaving it off — this is a genuine core-engine behavior change.

The coordinate-method crash path itself (the exact reported stack trace) remains covered by the manual Playwright repro described in the PR body, since it can't run under jsdom.

VTT-AI-Agent (commented):
## TL;DR

The new regression test covers the refactored `Widget` methods, but the client still has unguarded parent-chain walks that hang or overflow on the same cyclic state. In particular, dragging while a cyclic holder is visible in edit mode can freeze the client, so I think the cycle guard needs to be applied beyond `widget.js` before this is complete.

## Findings

- **The drop-target parent walk still loops forever on a cycle** (`client/js/main.js:53-64`). `getValidDropTargets()` repeatedly follows `tt.get('parent')` without tracking visited widgets. A cyclic holder is hidden during normal play by the new limbo treatment, but limbo widgets are deliberately visible in edit mode; dragging another widget there reaches this loop and alternates through the cycle forever. There are at least two other instances of the same incomplete coverage: the movable-ancestor loop in `client/js/mousehandling.js:82-92` and recursive `hiddenParent()` in `client/js/jsonedit.js:3809-3816`. Please make these traversals cycle-safe as well (ideally by reusing `ancestors()` where applicable) and add a regression assertion for at least the drop-target path. Otherwise a room repaired enough to render can still freeze or stack-overflow during ordinary edit-mode interaction.

The PR is engine-ineligible for `beta`, so I did not add the `beta` label because it genuinely changes core widget hierarchy, coordinate, and event behavior.

VTT-AI-Agent:
Addressed the cycle-unsafe edit-mode traversals in `563af617`:

- `getValidDropTargets()` now uses the cycle-safe `ancestors()` helper when rejecting a widget’s own descendant as a drop target.
- The mouse-down movable-ancestor lookup uses `ancestors().find(...)`, so a non-movable cyclic chain terminates cleanly.
- The JSON editor’s `hiddenParent()` check uses `ancestors().some(...)`, avoiding recursive overflow while preserving foreign-ancestor detection.
- Extended `widget-cyclic-parent.test.js` with direct drop-target coverage. It verifies that a cyclic target remains valid for an unrelated widget and is rejected when the dragged widget occurs in that cyclic ancestry.

Verification: full Jest suite passes (3 suites, 9 tests). I also audited the remaining direct parent traversals; the other looped walks are already cycle-guarded or are single-hop lookups.

VTT-AI-Agent (commented):
## TL;DR

The latest revision guards the previously reported parent walks, but applying a cyclic `parent` delta still creates a cycle in the widgets' runtime `parent`/`childArray` graph. Deleting either cyclic widget or replacing the room state can therefore still overflow the stack, and the new test bypasses this path.

## Findings

- **The limbo fallback still creates a cyclic logical child graph** (`client/js/widgets/widget.js:227-229`). When `cyclicParent` is true, the DOM node is correctly kept under `#topSurface`, but the code then assigns `this.parent` to the cyclic target and calls `applyChildAdd`, so a mutual `A ↔ B` state also becomes `A.childArray = [B]` and `B.childArray = [A]`. This leaves at least two concrete unguarded crash paths:
  - Deleting either widget from the editor or a `DELETE` routine enters `removeWidgetLocal()` (`client/js/serverstate.js:570-576`), which marks `inRemovalQueue` only *after* recursively collecting children; it therefore follows `A → B → A` until the stack overflows.
  - A later full-state replacement enters `receiveStateFromServer()` (`client/js/serverstate.js:477-478`). Because the limbo widget is a top-level DOM child, `applyRemoveRecursive()` follows the cyclic `childArray` graph (`client/js/widgets/widget.js:356-359`) and overflows before the new state can load.

  The regression fixture assigns only `w.state.parent` (`tests/client/widget-cyclic-parent.test.js:25-27`), so it never constructs the `parent` references or `childArray` cycle produced by a real delta. Please either avoid linking the cyclic limbo edge into the runtime child graph or make these recursive consumers cycle-safe, and add coverage that exercises deletion or state replacement after the cycle is applied.

I did not add the `beta` label because this PR genuinely changes core widget hierarchy and coordinate/event behavior and is therefore engine-ineligible.

VTT-AI-Agent:
Addressed the third review in `a704b88d`. The finding was valid; the saved `parent` cycle was still able to become a cyclic runtime child graph, and deletion separately revisited the same state-defined descendants before marking them.

Changes:
- When `applyDeltaToDOM()` detects that a new parent edge would close a cycle, it still preserves the requested `parent` value and limbo rendering, but no longer installs that edge in `this.parent` / `childArray`. The runtime graph therefore remains acyclic and `applyRemoveRecursive()` can tear it down during full-state replacement.
- `removeWidgetLocal()` now marks each widget `inRemovalQueue` before traversing its children. Normal acyclic deletion keeps the same child-first result, while cyclic state cannot revisit a queued ancestor.
- Added regressions that apply a real mutual-parent delta and then recursively tear down the result, and that run a `DELETE` routine against cyclic parent state.

Verification: full Jest suite passes (3 suites, 11 tests). The `beta` label remains absent as required for this engine change.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-targeted robustness fix: it makes every client-side parent-chain walk cycle-safe so a corrupt room state (a widget that is its own/mutual parent) shows the offending widgets as *limbo* instead of overflowing the stack on every mouse move. I checked each refactored walk against its original and they are behavior-identical for acyclic (i.e. all real) states, the new `null`/limbo paths are handled by their callers, and the 5 new regression tests plus the existing client suite pass locally. No blocking issues. Because this rewrites core engine hot paths (coordinate transforms, delta application, mouse move-target selection), it is a genuine engine change and is **not** getting the `beta` label.

## What I verified

- **Refactors are equivalent for acyclic states** (the normal case), confirmed by hand-tracing each:
  - `coordGlobalFromCoordLocal` / `coordLocalFromCoordGlobal` — the loop over `ancestors()` (and `.reverse()`) applies each widget's `coordParentFromCoordLocal` / `coordLocalFromCoordParent` in exactly the order the old recursion did.
  - `_absoluteRotation` (sum of `rotation` over `ancestors()`) and `_absoluteScale` (product of `scale`) match the old `self + parent`/`self * parent` recursion.
  - `_ancestor` — the new loop walks up while the *parent* is a `pile` and returns the last such pile's parent; traced against several shapes (parent-is-pile, parent-not-pile, no parent, `this`-is-pile) and it matches.
  - `isDescendantOf`, `inheritSeatVisibility`, `requiresHiddenCursor`, `getValidDropTargets`, `mousehandling` move-target selection — all reproduce the original precedence, including `requiresHiddenCursor` following `parent` first and only then `hoverParent`.
- **`serverstate.js removeWidgetLocal`** — moving `inRemovalQueue = true` above the children loop is the right fix; the `!childWidget.inRemovalQueue` guard now short-circuits re-entry, and for acyclic states nothing is missed since each widget is still pushed regardless of its own mark.
- **`applyDeltaToDOM`** — `this.domElement.contains(newParent)` correctly detects both self-parent and descendant-parent, routes the widget to `#topSurface`/limbo, and leaves `this.parent` unset so the runtime `childArray` graph stays acyclic (the `getElementTransformRelativeTo` → `null` and second `setLimbo` caller both guard the null).
- Ran `jest` locally: the new `tests/client/widget-cyclic-parent.test.js` (5 tests) and the rest of `tests/client` pass.

## Minor / non-blocking

- `calculateZ()` → `applyChildZ()` → `applyZ()` (widget.js:373–386) is another walk up the *state* `parent` chain (`this.get('parent')`) that is not routed through the new `ancestors()` helper. It is gated on `inheritChildZ` and self-terminates via the `if(this.z != pZ)` monotonic-`max` convergence, so a cycle won't spin forever — but it's the one remaining unguarded state-parent recursion, worth a comment or a follow-up if you want fully uniform cycle handling. Not required for this PR, which fixes the actually-reported crash path.

## Nice touches

The two "recurses one hop into the cycle-safe …" comments on `coordGlobalFromCoordParent`/`coordParentFromCoordGlobal` are exactly the kind of note a future reader needs to understand why those two still look recursive but can't loop.

I did **not** add the `beta` label: this PR really changes the engine — it rewrites core hot paths that run for every game (coordinate transforms on every mouse move, `applyDeltaToDOM`/`removeWidgetLocal` delta handling, and mouse move-target selection), so it is excluded from beta auto-deployment by design rather than being an incidental helper.

VTT-AI-Agent (commented):
## TL;DR

The refactored parent-chain walks are correct and behavior-identical for acyclic rooms, and the new jest file plus the full suite pass locally. But the guarantee added in a704b88d ("cyclic parent deltas keep the runtime child graph acyclic") only holds on clients that *receive* the delta: on the client that *executes* the cycle-creating routine, `Widget.onChildAdd()` still links the cycle into `childArray`, so the next full `state` message (loading another game, restore/undo, invalid-delta resync) overflows the stack in `applyRemoveRecursive()` and leaves that client stuck on the stale room. Two more state-parent recursions that this PR now makes reachable (recursive duplicate in the editor, `renderReadonlyCopy`) hang the tab on a cyclic room. None of this undoes the fix for the reported crash, but it is the same class of bug and worth closing in this PR.

## Findings

### 1. The originating client still ends up with a cyclic runtime child graph, and the next full state load overflows the stack

`client/js/widgets/widget.js:2541` (`onChildAdd`) and `client/js/widgets/widget.js:357` (`applyRemoveRecursive`).

Flow on the client that runs `SET parent` (e.g. the button routine in the PR description):
`set('parent', 'B')` → `sendPropertyUpdate` → `sendDelta` → `receiveDelta` → `applyDeltaToDOM` (the new `cyclicParent` guard works: `this.parent` is deleted, widget goes to limbo) → **then** `onPropertyChange('parent')` → `newParent.onChildAdd(this)` at widget.js:2541, which pushes the child into `childArray` with no cycle check. Result: `A.childArray = [B]` and `B.childArray = [A]`.

`receiveStateFromServer()` (`client/js/serverstate.js:477-478`) then calls `applyRemoveRecursive()` on every top-level widget, which recurses through `childArray` A → B → A → … and throws `Maximum call stack size exceeded` *before* `widgets.clear()`, so the client never renders the new state.

Reproduced with two clients in one room (script + log: [two-clients-state-replacement.cjs](https://agent.virtualtabletop.io/reports/review/1545466579720872038/repro/two-clients-state-replacement.cjs), [check5.log](https://agent.virtualtabletop.io/reports/review/1545466579720872038/repro/check5.log)):

| client | after the routine creates the cycle | after the room state is replaced (`PUT /state`, same path as `load`/`setState`) |
|---|---|---|
| originator (clicked the button) | `B.childArray = [A]` | `Maximum call stack size exceeded`; still shows the 4 old widgets, new state never appears |
| receiver (delta only) | `B.childArray = []` | fine, shows the new state |

Servers send a full `state` on `load` (library game / save file), `setState` (restore, `/state` API), and `receiveInvalidDelta` (resync of the offending player), so the originator hits this on the very next such event.

Suggested fix (either or both):
- In `onChildAdd()` mirror the guard from `applyDeltaToDOM`: `if(this.ancestors().includes(child)) return;` — at that point `this.ancestors()` already stops at the cycle, so for `B.onChildAdd(A)` it is `[B, A]` and for a self-parent `[A]`.
- Make `applyRemoveRecursive()` cycle-safe regardless of how the graph became cyclic (iterate with a visited `Set`, as `requiresHiddenCursor` now does). Teardown of the whole room should never be able to overflow.
- Extend the jest test at `tests/client/widget-cyclic-parent.test.js:69` to create the cycle through `await w1.set('parent', …)` (the way `widget-count.test.js` uses `set` under `mockConnection`) instead of `applyDelta()`; that is exactly what would have caught this.

### 2. Recursive duplicate on a cyclic widget hangs the tab (editor Clone drag, JSON editor "Duplicate widget")

`client/js/editmode.js:1381-1383`: `clone()` recurses over `widgetFilter(w=>w.get('parent')==widget.id)`, i.e. the *state* parent chain, with no visited set. The editor's Clone drag button always passes `recursive=true` (`client/js/editor/dragbuttons/clone.js:138`) and the JSON editor command defaults "Copy recursively" to true.

Reproduced: select the limbo holder in edit mode (it is selectable, which is good, that's how one would repair the room), drag the Clone handle one grid step → client main thread unresponsive until the tab is killed; a control clone-drag of a normal widget in the same room works (4 → 10 widgets, responsive). Scripts/logs: [clone-drag-cyclic-holder.cjs](https://agent.virtualtabletop.io/reports/review/1545466579720872038/repro/clone-drag-cyclic-holder.cjs) / [check10.log](https://agent.virtualtabletop.io/reports/review/1545466579720872038/repro/check10.log), control [clone-drag-control-normal-widget.cjs](https://agent.virtualtabletop.io/reports/review/1545466579720872038/repro/clone-drag-control-normal-widget.cjs) / [check11.log](https://agent.virtualtabletop.io/reports/review/1545466579720872038/repro/check11.log). Nothing reached the server in that run, so it is a client hang rather than room bloat.

This was unreachable before (the room crashed on hover), so it is not a regression, but since this PR is what makes cyclic rooms survivable in edit mode, a `seen` set in `duplicateWidget` (`if(seen.has(widget)) return []; seen.add(widget);` at the top of `clone`) would close it.

### 3. `renderReadonlyCopy(…, includeChildren)` recurses on the state parent chain unguarded

`client/js/widgets/widget.js:2613-2615`. Same pattern: `widgetFilter(w=>w.get('parent') == this.id)` recursion. It is called from `INPUT` `choose` fields via `visibleChildWidgets` (`client/js/domhelpers.js:371`), so a game routine listing a cyclic widget in an input can hang any client, not just the originator. Reproduced on the delta-only client: `renderReadonlyCopy({}, body, 'all')` on the cyclic holder → main thread unresponsive ([render-readonly-copy-receiver.cjs](https://agent.virtualtabletop.io/reports/review/1545466579720872038/repro/render-readonly-copy-receiver.cjs), [check8.log](https://agent.virtualtabletop.io/reports/review/1545466579720872038/repro/check8.log)). Lower priority than 1 and 2 (needs a game that shows the cyclic widget in an INPUT), but it is a one-line guard with a visited set threaded through the recursion.

## What I verified and found fine

- Each refactored walk (`coordGlobalFromCoordLocal`/`coordLocalFromCoordGlobal`, `_absoluteRotation`/`_absoluteScale`, `_ancestor`, `isDescendantOf`, `inheritSeatVisibility`, `requiresHiddenCursor`, `getValidDropTargets`, the mouse move-target lookup, the JSON editor `hiddenParent`) reproduces the original semantics for acyclic states.
- Both callers of `getElementTransformRelativeTo` (`applyDeltaToDOM`, `setLimbo`) handle the new `null` return.
- `removeWidgetLocal` marking `inRemovalQueue` before descending is correct; `DELETE` on the cyclic pair terminates (jest test 5).
- Single client: creating the cycle, hovering, dragging, and reloading the room produce zero errors; on reload both cyclic widgets are deferred and dropped with the existing "parent does not exist" console error, as before.
- `node --experimental-vm-modules node_modules/.bin/jest`: all suites pass (including the 5 new tests).
- No new author-facing property/function is exposed, so the property-editor / routine-block UI and demo-room rules do not apply; no `library/` files are touched.

I did **not** add the `beta` label: this PR really changes the engine (it rewrites the coordinate-transform hot paths and `applyDeltaToDOM` in `widget.js` and `removeWidgetLocal` in `serverstate.js`, which run for every game), so it is excluded from beta auto-deployment by design, and it already carries the `engine` label.

VTT-AI-Agent:
Addressed all three findings in `2ae4286b`. All of them reproduced here first, and each repro is green afterwards.

### 1. Cyclic runtime child graph on the originating client

Confirmed and fixed. `onChildAdd()` now refuses an edge whose child already is an ancestor:

```js
if(this.ancestors().includes(child))
  return;
```

and `applyRemoveRecursive()` takes a set of the widgets it already removed, so tearing a room down terminates no matter how the graph became cyclic (`receiveStateFromServer()` passes one set for the whole top-level loop, which also avoids removing a widget twice).

Two clients in one room, cycle created by a button routine on the originator, then `PUT /state` (the same path as load/restore/resync):

| | before | after |
|---|---|---|
| originator `h2.childArray` after the routine | `[h1]` | `[]` |
| originator after the state replacement | `RangeError: Maximum call stack size exceeded`, still showing the 4 stale widgets | loads the new state, zero errors |
| receiver | fine | fine |

### 2. Recursive duplicate in edit mode

Confirmed — and the suggested `seen` set alone was not enough, which the repro showed: the copies themselves become children of a source widget that is copied later. `clone(h1)` creates `C1` with `parent: h2`, so when the recursion descends into `h2`'s children it finds `C1`, clones it into `C2`, then finds `C2` as a child of `C1`, and so on forever.

`duplicateWidget()` therefore does both now: a per-invocation set of the source widgets already copied, plus a set of the ids it created, which are excluded when enumerating children. Both are inert for acyclic rooms, where a copy is never a child of a widget that is still to be copied.

Clone-dragging the cyclic holder one widget width: before, `mouse.up` never returned and the browser had to be killed after ~4 minutes; after, it returns and the room goes from 3 to 5 widgets and stays responsive. Control on a normal three-level tree still clones the full subtree exactly once (`h3←h4←b4`).

### 3. `renderReadonlyCopy(…, 'all')`

Confirmed and fixed with a visited set threaded through the recursion. On the cyclic holder it crashed the tab on both clients before (`Target crashed`); now it returns in 4-5 ms.

### Tests

`tests/client/widget-cyclic-parent.test.js` grew two cases, both of which fail on the previous revision:

- the cycle is now created through `await w.set('parent', …)` as suggested, and the test asserts the acyclic `childArray` and that `applyRemoveRecursive()` over all top-level widgets terminates;
- a recursive `renderReadonlyCopy({}, …, 'all')` over the cyclic chain returns.

`set('parent')` reaches `onChildAddAlign()`, which needs `DOMMatrix`/`DOMPoint`; jsdom implements neither, so `tests/client/setup.js` gained minimal 2D stand-ins (jsdom has no layout, so every transform there is 2D affine).

Full jest suite passes (3 suites, 13 tests). TestCafe locally: `editor.js` and `publiclibrary-3/4` pass; `publiclibrary-1/2` pass except `Reversi`, which produces a *different* hash on every run on this machine including on the unmodified branch head (`2f92a8e4…`, `b70d7fb7…`, `b2805542…` vs the expected `35e00175…`), so that one is local environment drift rather than a state change from this PR — CI is green on the previous revision for the same file.

### Try it

The demo room is on this PR's test server as **Cyclic parent chains** (<https://test.virtualtabletop.io/PR-3027/pr-test-ai>): one button makes the outer holder a child of the holder inside it, then the board has to stay usable — hover, drag, clone in edit mode, load another game. Repro scripts and logs: <https://agent.virtualtabletop.io/reports/pr/1525330699429871778/repro/README.md>

Still no `beta` label — agreed that this is a genuine engine change.

VTT-AI-Agent:
A new MAIN client crash report confirms that this PR covers another manifestation of the same parent-cycle path:

> Failed to execute `getComputedStyle` on `Window`: parameter 1 is not of type `Element`.

I reproduced it on current `main` by making one basic widget the wildcard inheritance source for a holder, then reparenting that source into the holder. The inherited `parent` update makes the holder point to itself; `getElementTransformRelativeTo()` then walks beyond the DOM root and calls `getComputedStyle()` with a non-element. Replaying the same operation against this PR produced zero page errors and rendered the impossible parent edge in limbo, as intended.

The report came from client crash telemetry; its private room and player data have intentionally not been included.

VTT-AI-Agent (commented):
## TL;DR

PR #3027 at `2ae4286b` addresses the previously reported cycle-related crash paths; I found no actionable code issues in this revision, and all 13 Jest tests pass. GitHub currently reports merge conflicts with `main`, so the resolved merge still needs verification.

*Automated review 4 of 4 (code) — by `gpt-6-astra`*

## Findings

No new findings. The originating-client guard in `onChildAdd()` prevents the rejected parent edge from being reintroduced into `childArray`, and the shared visited set makes full-state teardown terminate. Recursive duplication now excludes both previously visited sources and newly created copies; readonly copying also stops at revisited widgets. The rewritten parent walks preserve the traversal order for ordinary acyclic hierarchies.

## Verification

Ran `npm test -- --runInBand` on the PR head: **3 suites, 13 tests passed**, including real `applyDelta()` and `set('parent')` cycle creation, recursive teardown, readonly copies, and `DELETE`. Reviewed all nine changed files against the project architecture and engine conventions. Browser/TestCafe checks were not rerun in this review; these results cover the branch head, not a conflict-resolved merge with current `main`.

I did not add the `beta` label because this PR is engine-ineligible: `client/js/widgets/widget.js` and `client/js/serverstate.js` directly change widget hierarchy, coordinate traversal, deletion, and full-state teardown behavior for existing games.

VTT-AI-Agent:
Resolved the reported merge conflicts with current `main` in f9f0fd0e.

- Kept `main`'s newer drop-limit/line-drag behavior and asynchronous drag queue while applying the cycle-safe ancestry traversal to both conflicted parent walks.
- Kept `main`'s JSON editor implementation; the old hover-layer parent traversal changed by this PR no longer exists there.
- Confirmed GitHub now reports the PR as mergeable.

Verification on the resolved tree:

- Full Jest suite: 57 suites, 1,671 tests passed.
- Focused post-resolution rerun: `widget-cyclic-parent.test.js` and `line-widget.test.js`, 63 tests passed.

CI has restarted for the merge commit. The `beta` label remains absent because this is a core engine change.

Issue #2672, opened 2025-09-16
Client crash with INPUT and Edit Mode


96LawDawg:
To duplicate, start a game with a reset button that prompts for confirmation. Go into Edit mode. Click on the reset button until the modal confirmation prompt appears. Then choose "Debug" from the right side menu. Then press OK on the modal.  Client crashes with this:

```Cannot read properties of undefined (reading '1')
TypeError: Cannot read properties of undefined (reading '1')
    at jeLoggingRoutineOperationEnd (https://virtualtabletop.io/edit.js:1:227298)
    at Button.evaluateRoutine (https://virtualtabletop.io/dkbv:1:330046)
    at async Button.click (https://virtualtabletop.io/dkbv:1:289612)
    at async inputHandler (https://virtualtabletop.io/dkbv:1:216771)
```

PR #3028, opened 2026-07-12
Replace SWAPHANDS with SHIFT, an operation that cycles widgets between any holders and seats

VTT-AI-Agent:
## Summary

`SHIFT` is a routine operation that cycles widgets through an ordered list of holders and seats. It is the general form of `SWAPHANDS` — it works on plain holders as well as seats, moves a chosen subset of each container's widgets, and can shift by more than one position without wrapping — so this PR adds `SHIFT`, **removes `SWAPHANDS`**, and migrates every existing game to `SHIFT` on load.

### The operation

- `holders` (optional): the holders and seats the widgets travel along, in that order. Either an **array of ids**, used verbatim, or the **name of a collection**, whose widgets take part with the seats among them in seat `index` order. Seat entries resolve to that seat's `hand` holder, and a seat nobody sits on is skipped — it neither contributes nor receives widgets. **Left out, every occupied seat takes part, sorted by the seat `index` property.** With fewer than two entries there is simply nothing to shift.
- `widgets` (default `"all"`): which widgets in each container are moved — `"all"`, `"top"` (just the top-most widget), or the name of a collection to filter by.
- `interval` (default `1`): how many positions along `holders` each container's widgets move.
- `direction` (default `"forward"`): `"forward"`, `"backward"` or `"random"`. It is applied to the holder list itself — `"backward"` walks it the other way round and `"random"` shuffles it, so a random `SHIFT` hands each entry's widgets to an arbitrary other entry. An unknown value warns and falls back to `"forward"`.
- `wrap` (default `true`): when `false`, widgets that would shift past either end of `holders` stay at that end instead of wrapping around.
- `keepOrder` (default `true`): the widgets keep the order of the holder they come from. With `false` they arrive in the order they were created, which is what `SWAPHANDS` did.

So a bare operation passes every occupied seat's hand on to the seat with the next index:

```json
{ "func": "SHIFT" }
```

and the explicit form spells the cycle out — the order written there is the order used, seats and holders can be mixed:

```json
{
  "func": "SHIFT",
  "holders": [ "seat1", "seat2", "seat3", "seat4" ],
  "widgets": "all",
  "interval": 1
}
```

### Removing SWAPHANDS

`SWAPHANDS` is gone from the engine, `validator/validate_gamefile.js`, the JSON editor's context buttons and the visual routine editor's operation list. **File version 23** (`server/fileupdater.mjs`) rewrites every `SWAPHANDS` in an existing save as the `SHIFT` it always was, so no game has to change:

| SWAPHANDS | migrated to |
| --- | --- |
| `{ "func": "SWAPHANDS" }` | `{ "func": "SHIFT", "keepOrder": false }` |
| `interval`, `direction` | carried over unchanged |
| `source: "all"` (or absent) | left out — `SHIFT`'s default is the same set in the same order |
| `source: "usedSeats"` | `holders: "usedSeats"` |
| `source: [ "seat3", "seat1" ]` | a `SELECT` of those seats into a collection, then `holders` naming it |
| `keepOrder` absent | `keepOrder: false` — `SWAPHANDS` handed the widgets over in the order they were created |
| `keepOrder` written down (including an expression) | carried over unchanged |

A written-out list needs the `SELECT` detour because `SWAPHANDS` read its `source` as a collection and sorted the seats in it by `index`, while a list written into `holders` is deliberately used in the order it is written in.

The two public-library games that use `SWAPHANDS` (`Tutti Frutti`, `Floor of City`) were loaded through the updater and come out as `SHIFT`, so neither is changed here.

The one library change is the tutorial: `library/tutorials/Functions - SWAPHANDS/` is deleted — `Functions - SHIFT` teaches the same thing for the operation that replaced it — and the single sentence in `Functions - INPUT` that pointed players at SWAPHANDS for moving whole hands around now points at SHIFT. Both were decided by a maintainer (see the comment below); no other library or tutorial file is touched.

### Better PCIO imports

`server/pcioimport.mjs` builds a `SHIFT` for a `SHIFT_OBJECTS` step between seats now. Because `SHIFT` can do things `SWAPHANDS` could not, three fidelity gaps close with it: the seats are handed over as the **written-out list the automation gives**, so the hands travel along that order instead of along the seat `index` (and the `SELECT` the old shape needed is gone); passing the hands on *towards the last seat* instead of around the table is imported as `wrap: false` rather than warned about and wrapped anyway; and shifting only the *topmost* object is imported as `widgets: "top"` rather than silently moving the whole hand.

## Implementation notes

- `client/js/widgets/widget.js`: the `SHIFT` branch snapshots each container's selected widgets before making any moves (so simultaneous shifts around the cycle don't interfere with each other), computes each container's target index (mod-wrapped or clamped), and then hands each snapshot to a generated `MOVE` — targeting the **seat** itself when the entry is a seat. Ownership, hands that several seats share through `childrenPerOwner`, and the arranging of the receiving hand therefore all come from one implementation, and the routine debug view shows the nested moves.
- `client/js/editor/controls/routine.js` / `popup.js`: the visual routine editor knows `SHIFT`, so it can be added and edited as a sentence in edit mode without touching JSON. It sits under **Move and order widgets** as *"Pass the contents of holders to other holders"*; a bare one reads *"Pass the contents of every occupied seat's hand on to the next one"*. The holders chip opens a picker that offers the widgets in the room **and the collections the routine has**, and *n places along*, *which way round* (in the order they are listed / the other way round / to a random other one), *stopping at the last one*, *only some of what they hold* and *in the order the widgets were created* are the options it offers. The ⓘ topic describes every parameter and links the `Functions - SHIFT` tutorial.
- `client/js/jsonedit.js`: context-sensitive insert buttons for every property, plus enum buttons for `widgets`, `direction`, `wrap` and `keepOrder`.
- `validator/updated_gamefile.js`: `validate_gamefile_node.js` and `validate_library.js` read a game file through the file updater before validating it. A save on disk still says what it said when it was written, while what it *does* — and what edit mode shows these checks for — is the migrated state, so an operation a migration renames no longer reads as an error in a game that works.

### Why the default is not `${activeSeats}`

`${activeSeats}` lists the occupied seats in widget order (the order they appear in the save file), not by their `index` property, so it is not the seating order a shift around the table needs. Rather than change what `${activeSeats}` means for every game that already uses it, `SHIFT` derives the list itself when `holders` is left out. An explicit `holders` array is always used verbatim, so a game that wants a different cycle just writes it out.

## Try it yourself

A small demo room, **SHIFT on a shared hand (PCIO layout)**, is on this PR's test server: <https://test.virtualtabletop.io/PR-3028/pr-test-ai> (also downloadable [here](https://agent.virtualtabletop.io/reports/pr/1525857742886010880/SHIFT%20on%20a%20shared%20hand.vtt)). Three seats share one `childrenPerOwner` holder, exactly the way a PCIO import builds it. Sit down in the free seat, deal a card to every seat and press **SHIFT one seat on** — the readout at the bottom shows every card's owner and parent, so you can see the owners rotate while all three cards stay in the one hand; **SHIFT one seat back** undoes it. It lives on the test server only, **not** in the public library.

## Test plan

- [x] `npm test` (jest) — 1384 tests pass, including 15 `SHIFT` unit tests (wrap-around, clamped, `top`-only, seats, skipped empty seats, a hand shared by three seats, the seat-index default, `holders` naming a collection, `keepOrder` both ways, `direction: random`) and 8 file-updater tests for the `SWAPHANDS` migration
- [x] `npx testcafe chromium:headless tests/testcafe/routines.js` — 11 tests pass, the `SWAPHANDS` ones replaced by `SHIFT` counterparts plus new ones for `keepOrder: false` and for holders naming a collection of seats created out of index order
- [x] `node validator/validate_gamefile_node.js "library/tutorials/Functions - SHIFT/0.json"` reports no problems
- [x] Loaded a v22 save with `SWAPHANDS` in a `clickRoutine`, an `IF` and a `FOREACH` into a room on a real server and read the state back — every operation came out as the `SHIFT` in the table above, at version 23
- [x] Ran the three library files that use `SWAPHANDS` through the file updater and checked every migrated operation, including the `${PROPERTY keepOrder OF …}` expressions in the `Functions - SWAPHANDS` tutorial
- [x] In edit mode, added the *in the order the widgets were created* option (writes `keepOrder: false`), picked a collection for the holders chip out of *Groups of widgets the routine has* (writes `holders: "seats"`), and read back the operation list entry and the direction drop-down
- [x] In the JSON editor, put the cursor on a `SHIFT` and on its `keepOrder` and confirmed the insert buttons and the `true`/`false` enum buttons appear, and that `SWAPHANDS` is gone from the operation list
- [x] With the **Debug** sidebar open, ran a `SHIFT` and checked the log entry is complete and the nested `Moves` group opens and closes around the generated `MOVE`
- [x] Clicked a hand-written `SWAPHANDS` in a room already at version 23 (so the updater leaves it alone) and saw it report `SWAPHANDS was replaced by SHIFT.` instead of doing nothing
- [x] Played the demo room in a browser: dealt a card per seat, shifted forwards and backwards, and watched the owners rotate while all three cards stayed in the one shared hand
- [x] After deleting the tutorial, checked in a browser that the library list no longer offers `Functions - SWAPHANDS` and still offers `Functions - SHIFT` and `Functions - INPUT`, and loaded the edited `Functions - INPUT` variant into a room

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3028/pr-test (or any other room on that server)


VTT-AI-Agent:
**Update:** Added JSON Editor button support for SHIFT. Game creators can now visually build SHIFT operations in edit mode through the routine editor, with buttons for the , , , , and  parameters.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-tested addition of a `SHIFT` routine op — the snapshot-before-move design correctly avoids self-interference, validator/JSON-editor upkeep is done, and `npm test` passes locally. The main concern: when moving `"all"` widgets, `SHIFT` does **not** reverse the source stack the way `MOVE` does, so a multi-widget stack lands in the destination with its **stacking order inverted** (and re-inverts on every shift). This is an engine change (new routine operation), so it is **not** beta-eligible.

## Findings

### 1. Multi-widget `"all"` shifts invert stacking order (correctness)
`client/js/widgets/widget.js` — in the move loop:

```js
for(const c of move.widgets) {
  ...
  await c.moveToHolder(move.target.holder);
  ...
}
```

`move.widgets` comes from `holder.children()`, which is sorted **top-first** (z descending). `moveToHolder` calls `bringToFront()` (`z = getMaxZ()+1`) on each widget, so the widget moved *last* ends up on top. Iterating top-first therefore lands the original *bottom* widget on top → the destination stack is the **reverse** of the source, and it flips again on each subsequent `SHIFT`.

`MOVE` deliberately avoids this by iterating bottom-first: `source.children().slice(0, count).reverse()` (widget.js:1595). `SHIFT` should do the same, e.g. reverse `selected`/`move.widgets` before moving. This isn't caught by the tests because every `"all"` case uses single-widget containers. For an ordered deck or a multi-card hand passed around the table, the order corruption is a real functional bug.

### 2. Non-integer `steps` throws (edge case)
`steps` is validated only as `'number'`. A fractional value (e.g. `1.5`) produces a fractional `targetIndex`; with `wrap` the `((x % len)+len)%len` result is fractional, so `order[targetIndex]` is `undefined` and the subsequent `move.target.holder` access throws. Consider flooring `steps` (or validating it as an integer) so a bad input yields a `problems.push(...)` rather than an uncaught error.

### 3. `movedByButton` not set on programmatic moves (minor, layout)
`MOVE` and `CLONE` set `c.movedByButton = true` around `moveToHolder` (widget.js:1556, :504), which drives the "append to end of stack" path in `holder.onChildAddAlign` for `alignChildren`+`stackOffset` holders. `SHIFT` omits it, so mid-move placement follows the stale-coordinate branch. The trailing `updateAfterShuffle()` re-sorts by z afterward so the end result is usually fine, but it's an inconsistency with the other move ops worth matching (and it interacts with finding #1).

### 4. Shifting into an unoccupied seat's hand (behavior — confirm intended)
When the target seat has a `hand` but no `player`, `SHIFT` still moves the widgets in and (since `targetPlayer` is left unset) they become owned by whoever ran the routine. `MOVE` instead refuses this (`Seat ... is empty or does not define a hand`). Please confirm dumping cards into an empty seat's hand as clicker-owned is the intended semantics.

### 5. Test coverage gap
The tests cover wrap, clamp, and `"top"`, but not: multi-widget order preservation for `"all"` (would have surfaced #1), `steps > 1`, or the seat→hand ownership path. A case asserting the order of ≥3 widgets after a shift would be the most valuable addition.

## Notes
- Snapshot-before-move design is correct: `moves` are all computed from the initial state before any `moveToHolder`, and `applyChildAdd/Remove` reassign a new `childArray` so the captured `"all"` arrays aren't mutated mid-iteration. Good.
- Validator (`operationProps` + `requiredInputCollections`) and JSON-editor command/enum entries are updated consistently with the new op — nice adherence to the validator-upkeep convention.

I did **not** add the `beta` label: `SHIFT` adds a new routine operation executed by the engine's routine interpreter (core "how games run" behavior, not an incidental helper), so this PR is engine-ineligible for beta auto-deployment.

VTT-AI-Agent:
Thanks for the thorough review — all five points addressed in 6e6bb7a7.

**1. Multi-widget `"all"` shifts invert stacking order** — Fixed. The move loop now iterates `move.widgets.slice().reverse()`, i.e. bottom-first, so each `bringToFront` lands the original top widget on top and the source order is preserved (same reasoning as `MOVE`'s `.reverse()` at widget.js:1595). Added a regression test ("keeps its stacking order") over a 3-widget stack; I confirmed it fails on the old code and passes now.

**2. Non-integer `steps` throws** — Fixed both ends: the engine now uses `Math.trunc(a.steps)` so a fractional value can never yield an undefined target index, and the validator requires `Number.isInteger` (so edit mode flags a bad literal up front).

**3. `movedByButton` not set** — Fixed; it's now set/deleted around `moveToHolder` to match `MOVE`/`CLONE`.

**4. Shifting into an unoccupied seat's hand** — Adopted `MOVE`'s semantics: a target seat with no `player` is now skipped with a `problems.push(...)` message rather than dumping clicker-owned widgets into it.

**5. Test coverage gap** — Added tests for multi-widget order preservation (#1) and `steps > 1`. To make z-ordering assertions run under jsdom I provided minimal `getMaxZ`/`updateMaxZ` shims so the *real* `bringToFront` executes (rather than stubbing it out), which is what makes the order test meaningful.

Noted on beta — agreed this is engine-executed and not beta-eligible; I have not added the label. `npm test` passes (11/11).

VTT-AI-Agent (commented):
## TL;DR

Solid, well-scoped addition — the cycle logic is correct (the "snapshot before moving" claim holds because `applyChildAdd`/`applyChildRemove` **replace** `childArray` rather than mutate it in place, so the references captured in `moves` stay stable). The main gap is **test coverage**: the entire seat/hand path — the most novel and error-prone part of the change — is never exercised. A secondary concern is that an empty target seat is only detected *mid-application*, so a shift can partially apply.

## Findings

### 1. Seat/hand handling is completely untested (test coverage)
`tests/client/widget-shift.test.js` only ever uses plain `type: 'widget'` containers (`createContainers`). None of the tests touch a real `seat`, so the branch that makes SHIFT interesting for its headline use case — "pass hands to the left" — is unverified:
- `shiftContainer` resolving a seat to its `hand` holder (`widget.js:1897-1902`);
- the "seat with no valid hand" error path;
- ownership assignment via `c.targetPlayer` / `childrenPerOwner` on the target hand (`widget.js:1964-1965`);
- the empty-seat-target skip (`widget.js:1956-1958`).

Since MOVE-into-seat is the pattern this reuses and the code path with the most branching, it's exactly what a regression test should pin down. Adding a seat + `childrenPerOwner` hand case (asserting `owner` ends up set to the target seat's player) would materially raise confidence.

### 2. Empty target seat aborts non-atomically (correctness / consistency)
Setup validation is atomic: any bad `order` entry (missing hand, unknown ID, `order.length < 2`) sets `valid = false` and **no** moves are applied. But the empty-seat check happens *inside the apply loop* (`widget.js:1956`), after earlier moves in `moves[]` have already been committed. For a round-the-table rotation where one seat is empty, that leaves widgets partially shifted and the table in an inconsistent state, while pushing a problem.

Consider hoisting this check into the move-computation loop (lines 1936-1953) so an empty target seat invalidates the whole operation before anything moves — matching the all-or-nothing behavior of the other validation and avoiding a half-applied rotation.

### 3. Non-numeric `steps` can crash the routine (robustness, low severity)
`const shift = (a.reverse ? -1 : 1) * Math.trunc(a.steps);` (`widget.js:1932`). The validator only guarantees an integer in edit mode; at runtime `steps` can come from a computed expression. If it evaluates to something non-numeric, `shift` is `NaN`, `targetIndex` is `NaN`, `NaN != i` is true, and `order[NaN]` is `undefined` — so `move.target.seat` at line 1956 throws a `TypeError` and aborts the routine. A cheap guard (e.g. skip when `!Number.isFinite(shift)`, or treat as a no-op) would harden this. Low severity since well-formed games pass an integer literal.

### Minor
- `shiftContainer` treats any non-`seat` order entry as a holder without checking it's actually a `holder`; the validator restricts `order` to `['holder','seat']` in edit mode, but nothing enforces it at runtime. Harmless (any widget can be a parent) but worth being aware of.
- The `widgets` validator `v => v === 'all' || v === 'top' || typeof v === 'string' || '...'` has redundant first two clauses (both subsumed by `typeof v === 'string'`); no behavioral impact.

Nice touches worth calling out: the bottom-first (`.slice().reverse()`) move order to preserve stacking is correct and tested, and the mod-wrap `((x % n) + n) % n` handles negative `reverse` shifts properly.

---

This PR adds a new `SHIFT` routine operation that the engine instantiates and executes (new routine-operation semantics in `widget.js`), so it genuinely changes engine behavior and is excluded from beta auto-deployment by design — I did **not** add the `beta` label.

VTT-AI-Agent:
Thanks — addressed in 9af8674a.

**1. Seat/hand path untested** — Added a new `Scenarios: Shifting widgets through seats` block covering the branches you called out:
- seat→hand resolution + ownership: a token in seat A's hand shifts into seat B's hand and ends up `owner: "Bob"` (the target seat's player). The hand holder emulates a real Holder's `childrenPerOwner` (owner ← arriving child's `targetPlayer`), so this pins down that SHIFT sets `c.targetPlayer` correctly. I confirmed the test fails if the `targetPlayer` assignment is removed.
- empty-seat target → atomic abort (see #2).
- a seat lacking a valid hand → setup-phase abort, nothing moves.

**2. Empty target seat aborts non-atomically** — Fixed. The empty-seat check is now hoisted out of the apply loop: after computing `moves`, if any move's target is an empty seat, the whole SHIFT aborts with a problem before anything is moved. The regression test uses a 3-entry order `[holderX, emptySeat, holderZ]` where `holderZ` wraps back onto `holderX` — under the old per-move `continue` that move committed before the empty seat was reached; I confirmed the test fails on the old behavior and passes now.

**3. Non-finite `steps` can crash** — Fixed with a `Number.isFinite(a.steps)` guard that pushes a problem and skips the operation, so a computed non-numeric `steps` no longer yields a `NaN` index / `TypeError`.

**Minor** — Simplified the redundant `widgets` validator clause to a single `typeof v === 'string'` check. Left the runtime holder-type assumption as-is (any widget can legitimately be a parent, and the validator constrains `order` to holder/seat in edit mode), matching how the other move ops treat their targets.

`npm test` passes (14/14). Beta label still not added, per your engine-change assessment.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-tested addition. `SHIFT` correctly rotates widgets around a cycle of holders/seats, and the simultaneous-rotation correctness hinges on `Widget.children()` returning a live `childArray` reference that add/remove *reassign* (copy-on-write, `widget.js:138-139`/`2617`) rather than mutate — so the up-front `moves[]` snapshot really is a snapshot. I verified the full new suite plus the two existing client suites pass (14/14). Only minor edge-case notes below; nothing blocking.

## Findings

### Correctness (verified sound)
- The atomic-snapshot design is correct but *depends on an implementation detail*: `children()` returns `this.childArray` (sorted in place, `widget.js:372`), which is the same object reference, yet `onChildAdd`/`onChildRemove` do `this.childArray = this.childArray.filter(...)` (`widget.js:138-139`, `2617-2618`), replacing the array rather than mutating it. That's the only reason a source container's captured `move.widgets` isn't corrupted when an *earlier* move deposits widgets into it during the same rotation. It works, and the wrap-around test proves it — but it's a fragile invariant to rely on silently. A one-line comment at the snapshot site noting "relies on copy-on-write childArray so captured lists stay stable across moves" would protect this against a future refactor of `onChildAdd`.
- Empty-seat atomic abort (`widget.js`, the `emptySeatTarget` check before the move loop) is correct: it only inspects seat *targets that actually receive widgets*, and bails before any `moveToHolder`, so a rotation is never left half-applied. Good, and well-covered by the "middle target seat is empty" test.
- Reuse of the `MOVE` seat/hand + `targetPlayer`/`movedByButton` pattern is faithful, and ownership assignment into a seat's hand is exercised by a test.

### Minor / edge cases (non-blocking)
1. **Duplicate IDs in `order`** aren't guarded. `order: ["c0","c0","c1"]` snapshots `c0`'s widgets twice, producing two moves for the same widget; the second `moveToHolder` re-moves the already-moved widget, so it lands at the second entry's target and the first target ends up empty. It's user error, but a dedupe or a `problems.push` on repeated IDs would fail loudly instead of silently mis-shifting.
2. **`steps` truncation.** The validator requires an integer, but at runtime a variable-resolved non-integer (e.g. `2.9`) passes `Number.isFinite` and is silently `Math.trunc`'d to `2`. Fine, just undocumented.
3. **`widgets: "top"` ignores a collection filter** by design (`top` takes the single top-most child regardless of collection). Worth a note in the wiki so users don't expect `top` + collection to compose.
4. **Docs.** This adds a new routine function but the wiki Functions/Creating-Games pages (separate repo) aren't updated here — worth a follow-up so `SHIFT` is discoverable.

### Validator / tests
- `validate_gamefile.js` additions are correct and mirror `MOVE` exactly: `operationProps.SHIFT` (order via `getWidgetTypeValidator(['holder','seat'], true)`, string `widgets`, integer `steps`, boolean `reverse`/`wrap`) and `requiredInputCollections.SHIFT = ['order']`. Good validator-upkeep hygiene.
- `jsonedit.js` command/enum registration is consistent with the other operations.
- Test coverage is genuinely good: wrap, clamp (no-wrap), multi-step, stack-order preservation, `top`-only, seat→hand ownership, atomic empty-seat abort, and invalid-hand. Confirmed 8/8 pass locally (and 14/14 across all client suites).

### Engine / beta note
`client/js/widgets/widget.js` is an engine file, but this diff touches it **only additively**: a brand-new `if(a.func == 'SHIFT')` branch that runs *only* when a game explicitly invokes the new opt-in operation. It modifies no existing operation semantics, no state-sync/undo/delta/reconnect path, no default widget behavior, and introduces no new engine-instantiated widget type — so it cannot change how existing games run. That makes the engine touch incidental/opt-in, and a maintainer can safely allow-list this PR for the beta build (the builder still filename-skips `widgets/` until then).

---

Added the `beta` label: the change is safe and useful to test — it's a purely additive, opt-in routine operation (`widget.js` touched only incidentally), it passes tests and the validator, and it can't corrupt existing games or crash the server.

96LawDawg:
@ArnoldSmith86 ~~this one should be ready.~~ I know that I could technically approve and merge it since Claude did it and labeled it beta, but wanted to give you a chance to weigh in. No rush.

Edit: I need to do a tutorial. And I need to test SHIFTing to seats again with empty seats.  Just FYI.

VTT-AI-Agent:
Reviewed the added `Functions - SHIFT` tutorial — looks good and it exercises the API correctly:

- `node validator/validate_gamefile_node.js "library/tutorials/Functions - SHIFT/0.json"` passes with no problems.
- The control buttons feed SHIFT properly-typed values, which is what the engine expects: `steps` a number (button2), `reverse`/`wrap` booleans (button3/4), and `widgets` a string ("all"/"top", button1). So the runtime `Number.isFinite(steps)` guard and the boolean `reverse`/`wrap` checks all behave as intended here.
- `npm test` still green (14/14), including the SHIFT unit + seat/hand tests.

No code changes needed from me. (I left the tutorial itself untouched.)

96LawDawg:
SHIFT to seats is broken. The code correctly will not move to an empty seat, but once the order is an empty seat, then the SHIFT stops working. Fix it so that SHIFT simply skips over an empty seat and ignores it.

VTT-AI-Agent:
Fixed in ac49e1d2.

> SHIFT to seats is broken. The code correctly will not move to an empty seat, but once the order is an empty seat, then the SHIFT stops working. Fix it so that SHIFT simply skips over an empty seat and ignores it.

Unoccupied seats are now filtered out of the effective cycle rather than aborting the operation. An empty seat neither contributes nor receives widgets, so the shift just cycles through the remaining occupied seats/holders. For example, with `order: [seatA, emptySeat, seatC]` and `steps: 1`, seat A's hand passes straight to seat C (skipping the empty seat), and it wraps as usual.

Updated the seat tests to match and added a "passes to the next occupied seat, skipping the empty one" case (asserts ownership ends up as the receiving seat's player). `npm test` passes 15/15 and the SHIFT tutorial still validates clean.

96LawDawg:
I think this is now ready for final approval and merger.

96LawDawg (approved):
Approved.

RaphaelAlvez (approved):
Tested

96LawDawg:
@VTT-AI-Agent, this function was created to make compatibility with playingcards.io (PCIO) imports better. Can this new SHIFT function just be an expansion of SWAPHANDS and still meet that goal? Do not code.  Address the possible concerns as well as issues of backwards combability of SWAPHANDS if it is changed.

VTT-AI-Agent:
**Short answer:** yes — everything `SHIFT` does could live inside `SWAPHANDS` as optional parameters, and there is a real argument for it. But while checking, I found something that matters more than the merge question: **on an actual PCIO-imported game, `SHIFT` as it stands does not work at all**, while `SWAPHANDS` does. That's the strongest argument in favour of reusing SWAPHANDS' machinery (whether or not we merge the names). Details, concerns and the backwards-compatibility analysis below. No code changed.

---

### 1. What a PCIO import actually looks like

- `server/pcioimport.mjs:214-231` converts every PCIO hand into a `holder` with **`childrenPerOwner: true`**, and PCIO's main hand keeps the id `hand`.
- Imported seats (`server/pcioimport.mjs:595`) never get a `hand` property, so they fall back to the seat default **`hand: 'hand'`** (`client/js/widgets/seat.js:19`).

⇒ In an imported game **all seats share one single holder**. "Whose hand a card is in" is expressed purely by the card's `owner`, not by its parent.

### 2. Why that breaks SHIFT today

`SHIFT` resolves each seat to `seat.hand`, then takes `holder.children()` — which returns *all* children regardless of owner — and moves them with `moveToHolder` directly. With a shared hand, every entry in `order` resolves to the *same* holder, so every card is "selected" for every entry, the move is a parent-level no-op, and detaching from a `childrenPerOwner` parent clears `owner` (`client/js/widgets/widget.js:383`).

I reproduced it in a scratch jest scenario (not committed) mirroring the existing seat tests: three seats sharing one `childrenPerOwner` hand, one card owned by each player, then `{"func":"SHIFT","order":["seat1","seat2","seat3"],"steps":1}`:

```
AFTER SHIFT -> a:null b:null c:null | parents: hand,hand,hand
```

Nothing is passed and ownership is destroyed. (This is unit-level, but the mechanism is plain in the code paths above.)

`SWAPHANDS` gets the same layout right by construction:
- it filters the hand's contents by the source seat's player when `childrenPerOwner` is set (`widget.js:2070-2078`), and
- it delegates to `MOVE`, which has an explicit *"target hand **is** the source holder → only update `owner`"* branch (`widget.js:1562-1565`).

So measured strictly against the PCIO goal: SWAPHANDS ✔, SHIFT ✘ as written. Any path forward has to fix this — merging is one way to get it for free.

### 3. What SHIFT adds that SWAPHANDS cannot express today

1. **Non-seat containers.** SWAPHANDS filters `type == 'seat'` at its core; SHIFT cycles arbitrary holders (conveyor belts, market rows, board tracks, discard rotation).
2. **Author-defined order.** SWAPHANDS derives order from the seat `index` property; SHIFT takes an explicit `order` array (and can mix seats and holders).
3. **Which widgets move.** `widgets: "top"` (pass one card) or a collection name (pass only matching cards). SWAPHANDS always moves the whole hand.
4. **`wrap: false`** — a clamped, non-cyclic shift (a conveyor, not a merry-go-round).
5. **Preserves source stacking order** (it moves bottom-first because `moveToHolder` brings each widget to front).

### 4. What SWAPHANDS has that SHIFT lacks

- The per-owner / shared-hand handling above — i.e. the PCIO case.
- `direction: "random"`, which is a *permutation*, not a shift; it doesn't compose with `steps`/`wrap` at all.
- Routine-debug friendliness: it emits nested `MOVE` operations, so edit mode shows exactly what happened.

### 5. The options, with their costs

**A. Fold SHIFT's parameters into SWAPHANDS and drop SHIFT.**
- ✔ One function to learn instead of two overlapping ones; PCIO/per-owner correctness inherited for free; no duplicated "move cards into a seat" code path (SHIFT re-implements what `MOVE` already does, minus the hard cases); nothing to deprecate, since SHIFT hasn't been released.
- ✘ The name stops describing the function — "SWAPHANDS" cycling widgets between two board holders with no seats involved is not discoverable.
- ✘ Parameter collisions to resolve: `source` (collection of seats) vs `order` (explicit array); `interval` vs `steps`; `direction: forward/backward/random` vs `reverse` + `wrap`. Some combinations become meaningless (`direction: "random"` with `wrap: false`, or `random` with an explicit `order`) and would need documented precedence + warnings.
- ✘ Both the validator (`operationProps`) and the JSON-editor `jeCommands` list properties per operation with no notion of modes, so an author gets buttons and "valid" props that do nothing in their mode.

**B. Keep `SHIFT` as its own operation, but make it build `MOVE` operations the way SWAPHANDS does** (`to: <seatID>` when the entry is a seat), so `childrenPerOwner`, the same-hand case and ownership all come from one implementation. Optionally, later, re-express SWAPHANDS internally on top of the same helper while keeping its public API byte-identical.
- ✔ Fixes the PCIO gap, removes the duplicate move logic, keeps both names honest, keeps `random` where it belongs, and — importantly — leaves the shipped SWAPHANDS behaviour untouched, so no existing game can regress.
- ✘ Two operations with a visible overlap remain; the docs/wiki need a "which one do I want" sentence.

**C. Fix SHIFT's per-owner handling only, change nothing else.** Minimal risk, but keeps the duplication.

### 6. Backwards compatibility if SWAPHANDS is changed

`SWAPHANDS` has been public since #2142 and is used by `library/games/Tutti Frutti`, `library/games/Floor of City` and its own tutorial, plus an unknown number of user saves. Anything we do must keep those byte-identical in behaviour:

- **Defaults must not move.** `interval: 1`, `direction: "forward"`, `source: "all"` today. New parameters have to default to values that reproduce exactly the current behaviour (`order` unset ⇒ derive from occupied seats; `widgets` defaulting to the whole hand; `wrap` defaulting to true).
- **Order derivation must be reproduced exactly:** occupied seats only (`.filter(w => w.get('player'))`), sorted by the seat `index` property (not save order), reversed for `backward`, and the target computed as `(i + interval) % c.length`. If SHIFT's `steps` semantics (`Math.trunc`, negatives, clamping) get grafted onto `interval`, that arithmetic has to stay identical for the values existing games use. (Today a *negative* `interval` indexes out of range and silently does nothing useful — changing that is a fix, not a regression, but it is a behaviour change.)
- **The silent no-op with fewer than two occupied seats** is current behaviour and some games likely rely on it.
- **`direction: "random"` uses `Math.random()`**, so any refactor of that branch has to keep the same draw sequence in mind for the state-hash-based TestCafe suite.
- **Delegation to `MOVE` is observable**, both in the edit-mode routine debug view and through anything `MOVE` triggers (`updateAfterShuffle`, `bringToFront`, flip handling). Swapping SWAPHANDS onto SHIFT's direct `moveToHolder` path would change stacking order and the debug output for existing games — that's the specific regression to avoid.
- **No save migration is needed** as long as changes are additive. It only becomes a `FileUpdater` question (see `docs/compat.md`) if `SWAPHANDS` were ever *removed* in favour of a renamed op — which would be real migration work for zero user-visible benefit, so I'd rule that direction out.
- Practical note: any change to SWAPHANDS turns this from a purely additive PR into one that touches a shipped engine operation used by public-library games, which means re-review and a library/TestCafe regression pass. The risk is asymmetric — that asymmetry is the main reason I'd lean away from option A.

### 7. Recommendation

If the goal is PCIO compatibility specifically, the merge isn't what delivers it — **fixing the shared-`childrenPerOwner`-hand case is**, and the cheapest correct way to do that is to make SHIFT emit `MOVE`s to seats exactly like SWAPHANDS does (option B). That gets one implementation of the tricky part, keeps SWAPHANDS' shipped behaviour untouched, and keeps the two names meaning what they say.

If you'd still rather have a single operation, keeping the **name** `SWAPHANDS` and adding SHIFT's parameters to it is the variant with the least compatibility cost (no existing save needs rewriting, and SHIFT is unreleased so nothing is deprecated) — at the price of a misleading name plus the mode-dependent parameter matrix in §5A. In that case the `Functions - SHIFT` tutorial would need folding into the SWAPHANDS one; I'm leaving all tutorial files alone unless someone asks.

Happy to implement whichever you pick — just say the word.

96LawDawg:
Implement option B

VTT-AI-Agent:
Implemented in c0bf23ad.

> Implement option B

`SHIFT` no longer moves widgets with `moveToHolder` itself. It still snapshots each entry's contents up front and computes the target index the same way, but it now hands each snapshot to a generated `MOVE` — with `to: <seatID>` when the entry is a seat — the way `SWAPHANDS` does, using the same shadowed temporary collection so a collection of the surrounding routine with a colliding name survives.

What that buys, concretely:

- **The PCIO case works.** With several seats sharing one `childrenPerOwner` hand, each entry's contents are now filtered by that seat's player, and `MOVE`'s *"target hand is the source holder → only update `owner`"* branch does the rest. Verified in a browser on a room shaped like a PCIO import: `Alice, Bob, Carol` → `Bob, Carol, Alice` on each click, no card changing parent. On the previous implementation the same room came out with every `owner` cleared and nothing passed on.
- **One implementation of the tricky part.** Ownership assignment, the shared-hand case, `bringToFront` and `updateAfterShuffle` on the receiving hand all come from `MOVE` now instead of being duplicated in `SHIFT`.
- **Routine-debug friendliness.** The operation logs a nested `Moves` group, so edit mode shows exactly which `MOVE`s ran.
- **`SWAPHANDS` is untouched**, so nothing that uses it can regress.

Tests:

- `tests/testcafe/routines.js` gets `SHIFT` counterparts of the four `SWAPHANDS` tests: hand order, colliding collection name, a card that a routine of an earlier move deleted, and a `childrenPerOwner` hand shared by all seats. 8/8 pass locally.
- `tests/client/widget-shift.test.js` gets a shared-hand scenario; the two seat scenarios now give their tokens an owner, which is what a real `childrenPerOwner` hand does to anything that arrives in it. 54/54 jest tests pass.
- Walked the `Functions - SHIFT` tutorial through every mode in a browser and diffed the resulting board state against the same walk on the previous implementation — identical, so holder-based shifting is unchanged. The tutorial file itself is untouched and still validates clean.

VTT-AI-Agent:
Added a demo room so the fix can be checked by hand rather than taken on trust: **SHIFT on a shared hand (PCIO layout)**, now in this PR's test-server room — <https://test.virtualtabletop.io/PR-3028/pr-test-ai>.

Three seats share one `childrenPerOwner` holder, exactly the way a playingcards.io import builds it. Sit down in the free seat, deal a card to every seat, then press **SHIFT**: the readout at the bottom lists each card's owner and parent, so you can watch the owners rotate while all three cards stay in the one shared hand. A **SWAPHANDS** button sits next to it for comparison. On the previous implementation the same room came out with every owner cleared and nothing passed on.

The room is on the test server only — it is **not** in the public library and this PR contains no library or tutorial changes beyond the `Functions - SHIFT` tutorial that was already here. Say the word if you'd like it added to `library/tutorials/` as a second SHIFT variant.

96LawDawg:
Make SHIFT fully compatible with SWAPHANDS. The parameter names of SHIFT should more closely match SWAPHANDS.  Specifically, turn "reverse" into "direction" (with forward, backward, and random). Turn "steps" into "interval".  Turn "widgets" into "source." Turn "order" to "holders" and the default for holders should be "${activeSeats}".

VTT-AI-Agent:
Done — `SHIFT`'s parameters now mirror `SWAPHANDS`'s, pushed as `749c280d`.

> Make SHIFT fully compatible with SWAPHANDS. The parameter names of SHIFT should more closely match SWAPHANDS.  Specifically, turn "reverse" into "direction" (with forward, backward, and random). Turn "steps" into "interval".  Turn "widgets" into "source." Turn "order" to "holders" and the default for holders should be "${activeSeats}".

| was | is now |
| --- | --- |
| `order` (required) | `holders`, defaulting to `"${activeSeats}"` |
| `widgets` | `source` |
| `steps` | `interval` |
| `reverse` (boolean) | `direction`: `"forward"` \| `"backward"` \| `"random"` |

`wrap` is unchanged.

**`direction`** is applied to the holder list itself rather than to the step size, exactly the way `SWAPHANDS` sorts/shuffles its seat list before shifting by `interval`: `"backward"` reverses `holders`, `"random"` shuffles it, and the shift then always walks forward along the result. That keeps `"backward"` identical to the old `reverse: true` (including how `wrap: false` clamps at the ends) and makes `"random"` mean the same thing it means for `SWAPHANDS` — each entry hands its widgets to an arbitrary other one. An unrecognised value pushes the same `Warning: direction … interpreted as forward.` that `SWAPHANDS` does.

**The `${activeSeats}` default** means a bare `{ "func": "SHIFT" }` now does what `{ "func": "SWAPHANDS" }` does: it passes every occupied seat's hand on to the next seat. Note it uses `${activeSeats}` verbatim, so the cycle follows widget order rather than seat `index` — `SWAPHANDS` sorts by `index`. Say the word if you'd rather it sorted too; I left it alone so that "the default is `${activeSeats}`" stays literally true and a game can reproduce the default by writing it out.

One thing worth flagging: `source` means something different here than in `SWAPHANDS`, where it selects which *seats* take part. In `SHIFT` that role belongs to `holders`, so `source` selects which *widgets* inside each holder move (`"all"`, `"top"` or a collection). Unambiguous within `SHIFT`, but the same word carries a different meaning across the two functions — happy to rename it if you'd prefer something else.

Everything downstream was updated with it: the validator's `operationProps` (`direction` uses the same enum validator as `SWAPHANDS`), the JSON editor's context buttons and enum buttons, the `Functions - SHIFT` tutorial (its third oval button now cycles forward → backward → random, and the parameter descriptions were rewritten), the unit and TestCafe tests, and the demo room on the test server.

Since `SHIFT` has never been released, no saved game can contain the old names, so there's no `FileUpdater` migration.

Verified: `npm test` (56 pass), `testcafe chromium:headless tests/testcafe/routines.js` (9 pass, incl. a new one for the `${activeSeats}` default), the tutorial validates clean and I walked it in a browser through all three directions, and in the JSON editor I inserted every property with its context button and ran the resulting operation verbatim in a room with two occupied and one empty seat — the two occupied hands swap and the empty seat is left alone.

96LawDawg:
Thanks for the comments.  Make "source" back to "widgets". And I want seats to sort by index. If activeSeats does not do that, we need a solution that works behind the scenes and for game designers. Leaving "holders" blank by default, (as is done for SWAPHANDS) would be fine, but then we would lose sorting by seats as a default. Since this is a more flexible feature, I'm okay with that, but see what you can come up with that makes sense.

VTT-AI-Agent:
Done — pushed as `a425b6fe`.

> Thanks for the comments.  Make "source" back to "widgets". And I want seats to sort by index. If activeSeats does not do that, we need a solution that works behind the scenes and for game designers. Leaving "holders" blank by default, (as is done for SWAPHANDS) would be fine, but then we would lose sorting by seats as a default. Since this is a more flexible feature, I'm okay with that, but see what you can come up with that makes sense.

**`source` is `widgets` again.** Renamed everywhere: engine, validator, JSON-editor buttons, tutorial, tests.

**Blank `holders` *and* sorting by seat index — you can have both.** You don't lose the sorting by leaving `holders` out, because the default is no longer the literal string `${activeSeats}`; `SHIFT` now derives the list itself, the same way `SWAPHANDS` does:

> every seat with a player, sorted by the seat `index` property

So `{ "func": "SHIFT" }` is exactly `{ "func": "SWAPHANDS" }` — same seats, same order, same silent no-op when fewer than two seats are occupied. `backward` walks that list the other way (descending index) and `random` shuffles it, again like `SWAPHANDS`. `interval`, `wrap` and `widgets` all still apply on top of it.

**For game designers, the explicit form is unchanged:** an array in `holders` is used *verbatim*, in the order it is written, and may mix seats and holders. That is the whole point of the parameter — a conveyor belt or a market row has an order that no property can derive — so I deliberately did not sort an author-written list. If a designer wants the seats in index order, they leave `holders` out; if they want a different cycle, they write it out.

**Why not `${activeSeats}`:** it is built with `widgetFilter`, so it lists the occupied seats in *widget order* — the order the seats happen to sit in the save file — and ignores `index`. Usually the two agree (people add seats in order), but nothing guarantees it, so it is the wrong thing for a shift around the table.

That leaves one open question I'd rather you decided than decide myself: **should `${activeSeats}` (and the `activeSeats` collection) be sorted by seat index too?** It would make the variable mean "the occupied seats, in seating order" everywhere, which is almost certainly what every game that uses it already assumes — but it is a change to a shipped engine variable that **87 public-library games** use, so any game where the two orders differ would start dealing/looping in a different order. It is a one-line change; I've left it out of this PR because it is a different concern from `SHIFT` and would want its own library/TestCafe regression pass. Say the word and I'll do it here or in a separate PR.

Verified:

- `npm test` — 56 pass, including a new unit test with seats created out of index order (indices 1, 3, 2 plus an empty seat) that asserts the hand goes to the index-2 seat, not the next seat in the file.
- `npx testcafe chromium:headless tests/testcafe/routines.js` — 10 pass, including a new end-to-end `SHIFT without holders follows the seat index rather than the widget order`.
- The tutorial validates clean and I walked it in a browser (its `widgets` button still switches `all`/`top` live, no console errors).
- In the JSON editor I inserted `holders`, `widgets`, `interval`, `direction` and `wrap` with their context buttons on a real `SHIFT` operation and checked the `widgets` enum buttons offer `all`/`top`.

96LawDawg:
Update this PR from main.

VTT-AI-Agent:
Updated from `main` (merge commit d7fca5d). The only conflict was in `tests/testcafe/routines.js`, where `main` moved `expectEventually` into `test-util.js` — the local copy is gone and the `SHIFT` tests use the shared one now.

`main` has since gained the visual routine editor, which didn't know `SHIFT` yet, so the operation would have been addable only through the JSON editor. It now has a sentence and a help topic of its own (b4cbf9c):

SHIFT in the routine editor

- it sits under **Move and order widgets** in the operation list, with the line "Pass what holders hold on around a circle"
- a bare `SHIFT` reads *"Pass the contents of every occupied seat's hand on to the next one"*; the chip opens the widget picker (preset to holders, switchable to seats) and shows a written-out list once one is set
- the options are *n places along* (`interval`), *which way round* (`direction`: in the order they come in / the other way round / to a random other one), *stopping at the last one* (`wrap: false`) and *only some of what they hold* (`widgets`: all of them / the top widget / a named group of widgets)
- the ⓘ button describes every parameter and links the `Functions - SHIFT` tutorial

The wording is derived from how `SWAPHANDS` and `SHUFFLE` are worded there, but it is a proposal — say the word if you'd rather have different phrases (especially "Pass the contents of … on to the next one" and "stopping at the last one" for `wrap: false`) and I'll change them.

Verified on the merged branch: `npm test` (1373 tests) and `npx testcafe chromium:headless tests/testcafe/routines.js` (13 tests) pass, and I walked the new block in a browser — adding a `SHIFT` from the operation list, every drop-down, the holders picker, and the JSON each edit writes.

96LawDawg:
@VTT-AI-Agent, fix

- [ ] For the Edit Widget chips change for the text of the direction, change "in the order they come in" to "in the order they are listed". For the hover hint over SHIFT, change "Pass what holders hold on around a circle - ..." to "Pass the contents of holders to other holders - ..."
- [ ] Remove SWAPHANDS as a function.
- [ ] Create a file update to migrate existing SWAPHANDS to SHIFT.

VTT-AI-Agent:
All three done, pushed as `532329f` (rebased onto your tutorial commit).

> - [ ] For the Edit Widget chips change for the text of the direction, change "in the order they come in" to "in the order they are listed". For the hover hint over SHIFT, change "Pass what holders hold on around a circle - ..." to "Pass the contents of holders to other holders - ..."

Both changed. The operation list now reads **SHIFT — Pass the contents of holders to other holders**, and the *which way round* chip offers *in the order they are listed* / *the other way round* / *to a random other one*.

> - [ ] Remove SWAPHANDS as a function.

Gone from the engine, `validate_gamefile.js`, the JSON editor's context buttons and the routine editor's operation list. `server/pcioimport.mjs` builds a `SHIFT` now — and since `SHIFT` can do what `SWAPHANDS` could not, two PCIO fidelity gaps close with it: *shift towards the last seat* imports as `wrap: false` instead of being warned about and wrapped anyway, and *shift only the top object* imports as `widgets: "top"` instead of silently moving the whole hand.

> - [ ] Create a file update to migrate existing SWAPHANDS to SHIFT.

File version **23**. For it to be able to say everything `SWAPHANDS` said, `SHIFT` gained two things:

- **`keepOrder`** (default `true`): the widgets keep the order of the holder they come from. `SWAPHANDS` handed them over in the order they were *created*, so a migrated operation says `keepOrder: false` — deleting that line is how a game opts into the nicer behaviour.
- **`holders` may name a collection** instead of listing ids. Its widgets take part, the seats among them in seat `index` order — exactly the seat filter `SWAPHANDS` called `source`. A list written into `holders` is still used verbatim, which is the whole point of the parameter.

| SWAPHANDS | migrated to |
| --- | --- |
| `{ "func": "SWAPHANDS" }` | `{ "func": "SHIFT", "keepOrder": false }` |
| `interval`, `direction` | carried over unchanged |
| `source: "all"` (or absent) | left out — `SHIFT`'s default is the same set in the same order |
| `source: "usedSeats"` | `holders: "usedSeats"` |
| `source: [ "seat3", "seat1" ]` | a `SELECT` of those seats into a collection, then `holders` naming it |
| `keepOrder` written down (an expression included) | carried over unchanged |

The list needs the `SELECT` detour because `SWAPHANDS` read `source` as a collection and sorted its seats by `index`, while a list in `holders` is used in the order it is written in. It is not hypothetical — the `Functions - SWAPHANDS` tutorial uses exactly that form (and a `${PROPERTY keepOrder OF …}` expression, which is why anything written down is carried over rather than resolved).

I ran the three library files that use `SWAPHANDS` through the updater and checked every operation that came out, and loaded a v22 save with `SWAPHANDS` in a `clickRoutine`, an `IF` and a `FOREACH` into a room on a real server. **No library file is changed by this PR.**

### One thing for you to decide

`library/tutorials/Functions - SWAPHANDS/` now teaches an operation that no longer exists. The updater rewrites its buttons to `SHIFT` on load so it still *runs*, but every word in it says SWAPHANDS. Tutorials are hands-off, so I left it alone — say which you want and I'll do it:

1. **delete it** — `Functions - SHIFT` already covers the same ground, or
2. **let me rewrite it as a second `Functions - SHIFT` variant** about seats and hand order (published as a draft on the test server first, so you can play it before it lands).

`library/tutorials/Functions - INPUT/1.json` has one sentence pointing at SWAPHANDS and wants the same call.

### Verified

`npm test` (1384 pass), `npx testcafe chromium:headless tests/testcafe/routines.js` (11 pass), the tutorial validates clean, and in a browser I added the *in the order the widgets were created* option and picked a collection for the holders chip out of *Groups of widgets the routine has*:

SHIFT in the routine editor

The demo room on the test server lost its SWAPHANDS button and gained a **SHIFT one seat back** one instead: <https://test.virtualtabletop.io/PR-3028/pr-test-ai>

VTT-AI-Agent:
The `Functions - SWAPHANDS` tutorial is retired, pushed as `c50b9e5`. @96LawDawg answered the question in the maintainers' feedback thread:

> Delete library/tutorials/Functions - SWAPHANDS/ - the SHIFT tutorial replaces it

So `library/tutorials/Functions - SWAPHANDS/` (the game file and its library image) is deleted, and the one sentence in `library/tutorials/Functions - INPUT/1.json` that pointed players at SWAPHANDS for moving whole hands around now points at SHIFT — a single word, plus the usual `lastUpdate` bump. That is the whole library change in this PR; the two public-library games that use `SWAPHANDS` (`Tutti Frutti`, `Floor of City`) are untouched and come out of the file updater as `SHIFT`.

Verified: `npm test` (1384 pass), `node validator/validate_gamefile_node.js "library/tutorials/Functions - INPUT/1.json"` reports no problems, and in a browser the library list no longer offers `Functions - SWAPHANDS` while `Functions - SHIFT` and `Functions - INPUT` are both still there and load.

Earlier CI red on this branch was two `fixture.beforeEach` network flakes — Firefox `interactions.js` could not resolve `http://localhost:8272/testcafe-testing` and Chrome `card.js` failed with `TypeError: fetch failed`, both in tests that never touch SHIFT. Re-running those two jobs turned the whole suite green.

VTT-AI-Agent (commented):
## TL;DR

The `SHIFT` implementation is in good shape — building it on a generated `MOVE` is the right call, the snapshot-then-move ordering is correct, and the SWAPHANDS→SHIFT migration matches what the old operation actually did. Two things I'd fix before merge: the SHIFT branch uses the live `jeRoutineLogging` global instead of the `routineLogging` snapshot every other operation uses (that is exactly the mismatch the comment at `widget.js:1288-1292` says crashes the client, #2672), and removing `SWAPHANDS` from the validator makes the three library files that still contain it fail `validate_gamefile_node.js` today.

## Findings

### 1. `SHIFT` logs with the live global instead of the captured `routineLogging` — the crash #2672 guards against

`client/js/widgets/widget.js:2468`, `:2494`, `:2498` all test `jeRoutineLogging` directly:

```js
if(jeRoutineLogging)
  jeLoggingRoutineOperationStart("Moves", "Moves");
for(const move of moves) { … await this.evaluateRoutine([{ func: 'MOVE', … }], …) }
if(jeRoutineLogging)
  jeLoggingRoutineOperationEnd([], variables, collections, false);
```

`evaluateRoutine` deliberately snapshots that flag once at `widget.js:1293` (`const routineLogging = jeRoutineLogging;`), and the comment right above it spells out why:

> Toggling the Debug panel while the routine is suspended (e.g. waiting for an INPUT modal) would otherwise mismatch the jeLogging start/end calls and crash the client. (#2672)

`SHIFT` is the only operation in the file that bypasses it, and it is one of the few that `await`s a nested `evaluateRoutine` in a loop — a hand's `enterRoutine`/`leaveRoutine` firing an `INPUT` during one of the generated `MOVE`s suspends SHIFT for as long as the modal is open. Two concrete failures:

- Logging is **on** at routine start, the player toggles the Debug panel **off** while a nested `MOVE` is suspended → the `Moves` group is opened but never closed, and the outer `jeLoggingRoutineOperationEnd` at `:2727` (which uses `routineLogging`) closes the wrong level.
- Logging is **off** at routine start and toggled **on** mid-SHIFT → `jeLoggingRoutineOperationStart`/`Summary` fire without the matching outer `jeLoggingRoutineOperationStart` from `:1339`, which is the "logging enabled while the routine was running" case `jeLoggingRoutineNotLogged` at `:2739` exists to handle instead.

The old `SWAPHANDS` block used `routineLogging` throughout, so this is a regression introduced by the rewrite. Three-character fix: `jeRoutineLogging` → `routineLogging` in all three places.

### 2. Removing `SWAPHANDS` from the validator breaks validation of three library files that still contain it

Library JSON is migrated **on load**, so the files on disk stay at `_meta.version: 22` with `SWAPHANDS` in them. With this PR's `operationProps` change:

```
$ node validator/validate_gamefile_node.js "library/games/Tutti Frutti/0.json"
…
swap[clickRoutine.2.thenRoutine.0.thenRoutine.0]: SWAPHANDS is not a valid function (valid functions: …, SHIFT, SHUFFLE, SORT, …)
   (6 occurrences)
```

Same for `library/games/Floor of City/0.json` (1) and `library/tutorials/Functions - SWAPHANDS/0.json` (18). `.github/workflows/check-library.yml` only validates *changed* files, so CI is green on this PR — but the next PR that touches any of those three files is blocked by errors it didn't introduce, and a full `validator/validate_library.js` report is now noisy. Two ways out, both a maintainer call: run the v23 migration over those three files as part of this PR (a `library/` change, so it needs sign-off), or keep a deprecated `'SWAPHANDS'` entry in `operationProps` until they are rewritten. This overlaps with the open To Do about `Functions - SWAPHANDS`, so it may be worth folding into the same decision.

### 3. An un-migrated `SWAPHANDS` is now a completely silent no-op

`evaluateRoutine` has no unknown-`func` fallback — an operation whose `func` matches no `if(a.func == …)` branch just falls through to `jeLoggingRoutineOperationEnd` and does nothing, with no entry in `problems`. The file updater only reaches routines through `updateProperties`' `/Routine$/` match (`server/fileupdater.mjs:121-123`) plus `CLONE.properties`/`FOREACH`/`IF` recursion, so a routine built at runtime — `{"func":"SET","property":"clickRoutine","value":[{"func":"SWAPHANDS"}]}` is the common shape — is never migrated. Same for JSON pasted out of an old game or off Discord into the editor. The failure mode is "the button silently stops doing anything", which is the hardest kind to report. A one-liner next to the SHIFT branch would turn it into a message:

```js
if(a.func == 'SWAPHANDS')
  problems.push(`SWAPHANDS was replaced by SHIFT.`);
```

### 4. The PCIO import still throws away the holder order that `SHIFT` can now express

`server/pcioimport.mjs:2100-2120` keeps the SWAPHANDS-era detour for the seat case: `SELECT … collection: 'pcioSeats'` and then `holders: 'pcioSeats'`. But a `holders` naming a collection is re-sorted by seat `index` (`widget.js:2381-2386`), while PCIO's `SHIFT_OBJECTS` gives an **explicit ordered list** in `c.args.holders`. Seat `index` order and the PCIO list order are only the same by luck. Now that `holders` takes a verbatim array, the faithful import is just

```js
const shift = { note: 'Pass the hands on', func: 'SHIFT', holders: seats, interval: steps, direction: reversed ? 'backward' : 'forward' };
```

which also drops the `SELECT`. Given "Better PCIO imports" is one of the PR's selling points, this is the biggest one left on the table. (Related: `const order = reversed ? holders.slice().reverse() : holders;` at `:2088` is computed but unused in the seat branch.)

### 5. One occupied seat without a hand aborts the whole default `SHIFT`, where `SWAPHANDS` shifted the rest

`shiftContainer` (`widget.js:2361-2364`) returns `null` for a seat whose `hand` is missing or dangling, which sets `valid = false` and abandons the entire operation. With `holders` left out the entry list is derived from the room, so a single occupied seat that has no `hand` — a scorekeeper/observer seat is the realistic case — turns a migrated `{"func":"SWAPHANDS"}` from "passes the hands that exist around" into "does nothing and logs a problem". The old code called `isValidID(hand, problems)` per seat and simply pushed no move for that one (`widget.js`, removed block), so the other hands still moved.

I understand the atomic-abort came out of an earlier review round, and for an explicitly written `holders` list aborting is clearly right — the author asked for that exact cycle. But for the *derived* list (`useActiveSeats`) the entries weren't chosen by the game author, and a handless seat there reads much more like the unoccupied seat two lines above it, which is skipped. Skipping it in the derived case (and only aborting when the id was written down) would keep the migration behaviour-preserving.

### 6. Minor

- **`client/js/jsonedit.js:1475`** — the insert template is `{ holders: [], … }`, so clicking the `holders` button produces `"holders": []`, which the engine rejects with `SHIFT requires a 'holders' array of at least two holders or seats.` and the validator accepts. `holders: null` matches the engine's own default (every occupied seat) and follows the `SCORE seats: null` / `MOVE from: null` precedent on the neighbouring lines.
- **`validator/validate_gamefile.js:914`** — `'widgets': v=>typeof v === 'string' || '…'` accepts any string, so a typo'd collection name in `widgets` is not caught, unlike `holders`, which routes through `validators.inCollection`. `v => v === 'all' || v === 'top' ? true : validators.inCollection(v, context, propertyPath)` would close that.
- **`client/js/editor/controls/routine.js:1201`** — `widgetType: 'holder'` presets the picker's type filter, so seats are filtered out of the room picker until the user notices the dropdown. It's the operation's headline case ("pass the hands on"), and the filter is only a preset, but `'seat'` may be the friendlier default given plain holders are the more advanced use.

- `npm test` on the four affected suites: 424 passed (`widget-shift`, `fileupdater`, `routineeditor`, `pcio-import`). The `widget-shift` coverage is genuinely good — wrap, clamp, `top`, `interval: 2`, both `keepOrder` values, shared `childrenPerOwner` hands, seat-index default, `holders` naming a collection.
- The snapshot-then-move design is sound: `moves[]` is fully computed before the first `MOVE`, and `onChildAdd`/`onChildRemove` reassign `childArray` rather than mutating it, so a source's captured contents survive an earlier move depositing widgets into it.
- `selected.reverse()` before handing the list to `MOVE` is right — `children()` is z-descending and `MOVE` calls `bringToFront()` per widget, so bottom-first is what preserves the stack.
- Migration fidelity: `SWAPHANDS` read a written-out `source` array through `getCollection` and then sorted it by seat `index`, so the `SELECT`-into-a-collection detour in `v23SwapHandsToShift` really is the faithful translation, and `keepOrder: false` correctly reproduces the old creation-order default. `interval`/`direction` map one-to-one (reversing the order + a positive modulus is equivalent to the old sort + `%`).
- UI support for everything new is there: `SHIFT` is in `routineOperationGroups` under *Move and order widgets* with a chip for every parameter, an ⓘ topic, and `jeCommands` + enum buttons — so this is not a JSON-editor-only feature.
- Test room: the demo `.vtt` on the PR's own `pr-test-ai` server is current (it uses `holders`/`direction`, no stale `order`/`steps`/`reverse` or SWAPHANDS button), and `library/tutorials/Functions - SHIFT/` was authored by @96LawDawg directly (commits 0289b80d, 9d7e737e, a7e9c268), so the library addition is human-approved rather than something this review should flag. Its `_meta` parity and square 530×530 SVG library image check out, and the parameter buttons pass real numbers/booleans (not strings) into `interval`/`wrap`, which is what the engine's `Number.isFinite` check needs.

The `beta` label was already on this PR before I reviewed it, so I left it in place (removing it is a maintainer's call) even though the change is a genuine engine change — it adds a routine operation the engine executes, removes `SWAPHANDS`, and bumps the save format to version 23 — and would normally be excluded from beta auto-deployment on that basis.

VTT-AI-Agent:
Thanks — five of the six findings were real and are fixed in `b18a92b`. Point by point:

### 1. `jeRoutineLogging` vs the captured `routineLogging` — fixed

Right, and it was a regression: the old `SWAPHANDS` block used `routineLogging` throughout. All three places in the `SHIFT` branch now read the captured flag, so it behaves like every other operation. (`widget.js:2739` still uses the live global on purpose — that is the `jeLoggingRoutineNotLogged` case.)

Verified in a browser with the **Debug** sidebar open: the entry is complete and the nested group opens and closes around the generated move —

```
SHIFT  shifted all widgets 1 step(s) forward (wrapped) along ["seat1","seat2"]
  Moves
    MOVE  all widgets from 'hand of seat1' to 'seat2'
```

### 2. Validator vs. on-disk library files — fixed, at the root

Rather than keep a zombie `SWAPHANDS` entry or rewrite library JSON, the two node entry points now read a game file **through the file updater** (`validator/updated_gamefile.js`, used by `validate_gamefile_node.js` and `validate_library.js`). A save on disk still says what it said when it was written; what it *does* is what the updater produces, and edit mode — where these checks are shown — only ever sees the migrated state. So the node tool and edit mode finally agree.

```
$ node validator/validate_gamefile_node.js "library/games/Tutti Frutti/0.json" | grep -c SWAPHANDS
0
```

Both files still report their pre-existing problems (undefined variables, deck `width`/`height`, `widget 'null' not found` — 20 and 28 lines, unchanged by this PR); only the 7 SWAPHANDS lines are gone. Note the third file you listed, `library/tutorials/Functions - SWAPHANDS/0.json`, was deleted in `c50b9e5` on @96LawDawg's instruction, so it is moot. A file the updater refuses (written by a newer version) is validated as written; four tests in `tests/server/updated-gamefile.test.js` pin all of this.

### 3. Silent no-op — fixed

Taken as written:

```js
if(a.func == 'SWAPHANDS')
  problems.push(`SWAPHANDS was replaced by SHIFT.`);
```

Checked with a v23 room whose button carries a hand-written `SWAPHANDS` (so the updater correctly leaves it alone): clicking it now logs `[SWAPHANDS was replaced by SHIFT.]` instead of doing nothing quietly.

### 4. PCIO import order — fixed

Agreed, and you're right that this was the biggest one left. The seat branch passes `holders: seats` verbatim now and the `SELECT` into `pcioSeats` is gone, so the hands travel along the order the automation gives instead of along the seat index. One correction: `const order = …` at `:2088` is *not* unused — the plain-holder path below still walks it (`:2129-2133`).

### 5. Handless occupied seat — fixed the way you framed it

Split on where the list came from. A list the game wrote down names the exact cycle it asked for, so a broken entry in it still stops the operation; a list derived from the room (no `holders`, or a collection) skips the seat like an unoccupied one and the rest still move. That restores the `SWAPHANDS` behaviour for every migrated game. New unit test: *"Given an occupied seat without a valid hand and a SHIFT without holders → the seat is skipped and the other hands still move"*.

### 6. Minor

- **`holders: []` → `holders: null`** — fixed. The button now inserts `"holders": null`, which is the engine's own default; confirmed in the JSON editor.
- **`widgets` accepting any string** — fixed: `v === 'all' || v === 'top' || validators.inCollection(…)`. `${…}` values are skipped before the validator runs, so expressions are unaffected.
- **`widgetType: 'holder'` on the holders chip** — left as is, deliberately. You only open that picker to name something *other* than the seats: the chip's default already reads "every occupied seat's hand" and passing the hands on needs no picker at all. Presetting it to `seat` would filter plain holders out of the exact case the picker exists for. The type dropdown is one click away either way — happy to flip it if you disagree.

### Verified

`npm test` 1389 pass (43 suites, +4 for the updater-aware validator), `npx testcafe chromium:headless tests/testcafe/routines.js` 11 pass, and the browser checks above for the debug log, the SWAPHANDS message and the JSON-editor button.

On the `beta` label: agreed it is odd for a change that adds a routine operation, removes one and bumps the save format — but it was applied before I touched it and taking it off is a maintainer's call, so I left it.

PR #3030, opened 2026-07-14
Fix validator accepting undefined GET properties and uploads of files without VTT metadata

### LINKED ISSUE #2825: Validator and GET
A property should not be valid if it is only read but never written. If you `GET`, property=invalidProperty, it won't trigger an error because it found a valid use of invalidProperty (the `GET` itself). Reading a property should trigger an error if there is no `SET` and it's not set directly in any widget.

VTT-AI-Agent:
This PR carries two fixes: the validator change described first, and a fix for the game-file upload preview crashing on archives that carry no VTT metadata (see [Uploading a file without VTT metadata](#uploading-a-file-without-vtt-metadata)).

## Problem

Fixes #2825. The validator treated every property read by `GET` as evidence that the property existed. As a result, a typo or otherwise undefined custom property in `GET` passed validation simply because the same operation read it.

## Fix

Track custom-property usage separately from custom-property definitions. Routine reads still count as usage for the existing unused-property checks, but property validation now accepts a custom property only when something actually provides it:

- a widget declares it (including card properties inherited from the deck's `cardDefaults`, `cardTypes` and `faceTemplates[].properties`)
- a routine writes it: `SET`, `SCORE` (including its default `score`), `CLONE`'s `properties` keys, or the keys of a map that `RESET` applies
- the engine writes it without a routine: the keys of a holder's or line's `onEnter`/`onLeave`, everything beyond the lattice keys in a `grid` entry, and the card property a `write` face object binds its value to
- the engine computes it (`_absoluteX`, `_ancestor`, the scoreboard's `_totals`, …)
- the scoreboard's configured score property

Write targets whose name is built at runtime, like `SET property: "score_${player}"`, are kept as patterns, so a later `GET property: "score_1"` is still accepted. Uses and definitions are collected in a single pass over the game file.

The message for a read that nothing provides is `'x' is read but nothing ever sets it - check for a typo` — the property name *is* present in the file, it is just never written, and in every library case it turned out to be a misspelling.

## Debug panel

These findings are the first thing that makes the edit-mode **Debug → validation table** non-empty on real games, and the message is a full sentence rather than the old four words. The table did not cope: per-cell `max-width` caps stopped it at ~610px inside an 812px module, every row wrapped, and the JSON path broke mid-token. So this PR also fixes the table around the message:

- proportional column widths (18/32/50%) with `table-layout: fixed` instead of pixel caps, and the table is rendered as a `table` rather than a `block`, so it fills the module ([before](https://agent.virtualtabletop.io/reports/pr/1526269418743529522/ui-review-3030/before-1920x1080.png) → [after](https://agent.virtualtabletop.io/reports/pr/1526269418743529522/ui-review-3030/after-1920x1080.png))
- the path column breaks at its dots (`<wbr>`) instead of mid-token, and the widget column no longer breaks `turn_button` into `turn_butto` / `n`
- below 400px module width each finding stacks into a block — that is the case when Debug is docked beside the JSON editor, where the three columns degraded to ~8 characters per line ([after, 220px](https://agent.virtualtabletop.io/reports/pr/1526269418743529522/ui-review-3030/after-narrow-220px.png)). The box is declared as the named container `debugValidation` with the `container` shorthand the rest of the client uses, which is what `client/js/containerQueryFallback.js` can resolve and what the `browser-compat` check accepts
- a `5 validation problems found:` line heads the table, which also separates it from the routine log and its `Filter log…` box above
- the `Property` column is renamed to `Location`: it holds a JSON path, while the messages talk about property *names*
- clicking a row selects the widget even when the JSON editor is closed — previously the row showed `cursor: pointer` and did nothing
- the rows are built from text nodes instead of `innerHTML`, so a widget id or property name containing HTML is shown literally rather than parsed (widget ids come straight from the game file)
- the `No validation problems found!` box uses theme colours, so it is no longer light-mode green in dark mode ([dark mode](https://agent.virtualtabletop.io/reports/pr/1526269418743529522/ui-review-3030/after-darkmode.png))

## Effect on the public library

Running the validator over all 578 files in `library/games` + `library/tutorials` produces **77 new findings in 29 files** (0 before), and removes 4 pre-existing false `unrecognized (and seemingly unused) property` warnings in *Interrogation*. No other message anywhere in the library changes. Spot checks show the new findings are genuine dead reads, mostly the copy-pasted `GET x` → `IF ${x} == null` idiom where `x` is never written:

| property | count | example |
| --- | --- | --- |
| `defaultParent` | 26 | *Cat City* reads `defaultParent` but uses `default_parent` |
| `empty_hand` | 21 | *Building Buildings* never writes `empty_hand` |
| `keep_parent` | 10 | |
| `defaultx` / `defaulty` / `defaultX` | 11 | *Irrigation* reads `defaultX`, only `defaultY` exists |
| `default_parent` | 4 | |
| `audio_on_svg` | 2 | |
| `number`, `card_color`, `audio_your_turn_enabled` | 3 | |

Affected files: Building Buildings, Buy the Block, Cat City, Creation, Dice Coin, Flower Fields, Furious Farmers, Hunt & Gather, Irrigation, Landgrab, Line Fight, Magical Traders, Marble Markets, Marriage Routes, Mirage, Mutation (0/1/2), Pax Penning, Rails Before the War, Rocky Rails, Shell Game, Sneaky Sheep, Splendid (1), Steam Power (0/1), Too Many Roads, Union Station, Vineyard.

No game file is touched by this PR — those games will simply show the finding in edit mode. `check-library.yml` only validates changed files and does not hard-fail, so CI stays green.

## Uploading a file without VTT metadata

Adding a game from a Tabletop Simulator export — or from any zip whose root JSON has no `_meta` — aborted with `Cannot read properties of undefined (reading 'info')`, reported from the live server. Nothing was uploaded and the only feedback was the generic error toast.

`uploadStateFile()` in `client/js/overlays/states.js` builds the preview tile (name, similar name, image, variant list) from the archive before sending it. It unzipped **every** entry whose name ends in `json` and read `_meta.info` off each one:

- a TTS export keeps its save in a subdirectory (`Saves/TS_Save_1.json` next to `WorkshopUpload`) and has no `_meta` at all — the server converts it into a game on arrival, so it is a perfectly valid upload
- an upload that leaves out assets the server already has carries `asset-map.json`, which is not a variant
- a PCIO export was handled, but only because its `widgets.json` happens to parse as an array

The variant files are now the same set `server/fileloader.mjs` reads — root-level `*.json` minus `asset-map.json` — and an entry whose metadata is missing or malformed no longer contributes one. When no variant has metadata, the upload falls back to the filename-only preview PCIO files already used, and the server does the rest. An archive with no JSON file at all still gets the "not a valid VTT, VTTC, VTTS or PCIO file" message.

## Testing

- `npm test` — 1687 tests passing, the validator suite among them at 28, with regressions for `SET`, `SCORE` (explicit and default property), `CLONE properties`, `RESET`, interpolated `SET`/`RESET` names (accepting a matching read, still reporting a non-matching one), fully dynamic `SET`/`RESET` names, deck-inherited card properties, holder/line `onEnter`/`onLeave`, grid entries (accepted, while a lattice key such as `alignX` is still reported), `write` face bindings, computed read-only properties, and a non-object game file
- `npm run browsercompat` — green
- Verified in a browser (edit mode → DEBUG sidebar) on a room whose routine uses `SCORE`, `CLONE properties` and an interpolated `SET`: only the deliberate typo is reported — [screenshot](https://agent.virtualtabletop.io/reports/pr/1526269418743529522/validator-edit-mode.png)
- `npx testcafe chromium:headless tests/testcafe/compute-4.js` — 2 passing (the only TestCafe test that opens the Debug module)
- The upload fix has its own suite, `tests/client/state-upload.test.js` (8 tests): a VTT save with two variants, an embedded image, a TTS export, a PCIO export, a variants list next to `asset-map.json`, a root JSON without `_meta`, an archive with no JSON, and a file that is not a zip. Three of them fail on `main` with the reported error.
- Upload checked in a browser: dropping the TTS export into a room throws `Cannot read properties of undefined (reading 'info')` before the change and uploads as *Workshop Game* after it, with a normal VTT save still showing its own name and image.
- Debug panel checked in a browser on *Cat City* and *Irrigation* at 390x844, 820x1180, 1280x800 and 1920x1080, at a 220px module width, and in dark mode: [1280](https://agent.virtualtabletop.io/reports/pr/1526269418743529522/ui-review-3030/after-1280x800.png), [390](https://agent.virtualtabletop.io/reports/pr/1526269418743529522/ui-review-3030/after-390x844.png), [220px](https://agent.virtualtabletop.io/reports/pr/1526269418743529522/ui-review-3030/after-narrow-220px.png), [dark](https://agent.virtualtabletop.io/reports/pr/1526269418743529522/ui-review-3030/after-darkmode.png), [dark empty state](https://agent.virtualtabletop.io/reports/pr/1526269418743529522/ui-review-3030/after-darkmode-empty.png). Clicking a finding with the JSON editor closed now selects the widget. With the named container the table's computed `display` is `table` at 584px and `block` at 220px, both from the stylesheet.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3030/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

The read-vs-definition split addresses #2825, but the new definition set overlooks valid properties supplied by the engine. As written, the validator introduces hundreds of false “property not found” reports in existing library games.

## Findings

### High — Include engine-provided properties in the definition set

`validator/validate_gamefile.js:900-905`

`definedCustomProperties` currently learns only top-level custom widget keys, static `SET` targets, and scoreboard score properties. Cards also receive readable properties from their deck’s `cardDefaults`, `cardTypes`, and `faceTemplates[*].properties` (`Card.getDefaultValue()` delegates to `Deck.cardPropertyGet()`), while every widget exposes computed read-only properties such as `_absoluteX` and `_centerAbsoluteY`. None of those sources enter this set.

Consequently, valid routines now receive errors. For example, `library/games/Wild Escape/0.json` reads `rank`, which is defined in its decks’ `cardTypes`, and `library/games/Trajectory/0.json` reads the built-in `_absoluteX`; minimal versions of both cases return `property '…' not found` on this branch. Scanning the 569 checked-in game/tutorial JSON files produced 374 such errors. Please extend the definition model (or the property validator’s built-in knowledge) to cover inherited card properties and computed read-only properties, with regression tests for both cases, so fixing undefined `GET` targets does not make valid engine reads fail validation.

I added the `beta` label because this validator-only change does not alter engine behavior or threaten game/server state, so it is safe and useful to test despite the false-positive issue.

VTT-AI-Agent:
Addressed the review in fe2d0a4c.

- Added all engine-provided read-only properties from `Widget` plus scoreboard `_totals` to the valid definition set.
- Added card properties inherited through deck `cardDefaults`, every `cardTypes` entry, and `faceTemplates[].properties`.
- Added regression tests for all three deck sources and both base-widget/scoreboard computed properties.

`npm test` now passes all 23 tests. I also reran the reviewer’s scan across all 569 checked-in saves: the 374 reports dropped to 79. I inspected the remaining property names; they are genuinely not declared or written in their individual files (for example, Building Buildings reads `empty_hand` but has no `empty_hand` property or `SET` target anywhere). Those are the previously hidden undefined reads #2825 is intended to report, so I did not whitelist them and recreate the original bug.

VTT-AI-Agent (commented):
## TL;DR

Good, well-targeted fix for #2825 — I verified it catches real typos in the shipped library (31 games / 79 new findings, e.g. Cat City reads `defaultParent` while the game actually uses `default_parent`). But the "definition" set is incomplete: `SCORE` and `CLONE properties` both write widget properties at runtime and aren't counted, so a `GET` of such a property now produces a false *"property 'x' not found"* in edit mode.

## Findings

### 1. `SCORE` writes a property but isn't treated as a definition (false positive)

`validate_gamefile.js:1052` still calls `addPropertyUsage(value)` for `SCORE`, but `SCORE` genuinely writes the property on every matching seat (`client/js/widgets/widget.js:1858`: `await seats[i].set(String(a.property), newScore)`). So the perfectly valid pattern below now errors:

```json
{"func": "SCORE", "property": "myScore", "mode": "set", "value": 3}
{"func": "GET", "property": "myScore"}
```
→ `property 'myScore' not found` (verified against this branch; clean on `main`).

Fix: `addPropertyUsage(value, true)` for `SCORE`. Note the default matters too — `SCORE` without `property` defaults to `score`, so an op with no `property` key should define `score` (same as the scoreboard handling you already added).

### 2. `CLONE`'s `properties` keys aren't treated as definitions (false positive)

`CLONE` applies `properties` onto the new widget (`client/js/widgets/widget.js:1232`, `w.clone(a.properties, …)`), so those keys are definitions. Currently:

```json
{"func": "CLONE", "properties": {"foo": 1}}
{"func": "GET", "property": "foo"}
```
→ `property 'foo' not found`. Neither of these two cases shows up in the library today, so they're latent rather than immediately visible — but both are ordinary idioms.

While you're there, it's worth scanning the other property-writing surfaces for the same gap; the deck `cardDefaults` / `cardTypes` / `faceTemplates[].properties` handling you added is correct (matches `deck.js:52-60`).

### 3. Interpolated `SET` property names don't feed the pattern matcher

`addPropertyUsage(value, true)` stores the raw string, so `SET property: "score_${player}"` defines the literal `"score_${player}"` and a later `GET property: "score_1"` errors unless `score_1` happens to be declared on some widget. `addPropertyPatternMatches()` already exists for exactly this fuzziness in the other direction — definitions should populate a pattern list that the read check consults, otherwise dynamically-named properties become a new false-positive class.

### 4. The whole game file is now scanned twice

`validate_gamefile.js:1084-1086` calls `getCustomPropertyUsage(data)` twice; the function already builds *both* sets in a single pass and `definitionsOnly` only picks which one to return. Returning `{ used, defined }` from one call would halve the cost of this (recursive, whole-file) scan — it runs on every edit-mode validation — and drops the confusing boolean parameter.

### 5. Widget-declared properties are already covered

Seeding `definedCustomProperties` from top-level widget keys duplicates the existing clause in `validators.property` (`validate_gamefile.js:38`: `Object.values(p.widgets).some(w=>w[v] !== undefined)`). Not harmful, but the `declaredCustomProperties`/`definedCustomProperties` pair now diverges only by the deck/SET/scoreboard additions — worth a comment or a merge so the next reader doesn't have to diff the two sets.

### 6. `READ_ONLY_PROPERTIES` duplicates the engine list

`validate_gamefile.js:46-50` is a verbatim copy of `readOnlyProperties` in `client/js/widgets/widget.js:10` plus `_totals`. It's accurate today (I diffed it against the engine), but it will drift silently. At minimum add a comment pointing at `client/js/widgets/widget.js` so the two are kept in sync.

### 7. Impact on the shipped library (informational)

Running the validator over `library/` on this branch produces **79 new `property 'x' not found` findings across 31 games** (0 on `main`). I spot-checked several and they look like genuine bugs, not validator noise — e.g. *Cat City* `GET property: "defaultParent"` while the game elsewhere uses `${PROPERTY default_parent OF …}`, and *Irrigation* reads `defaultX` while only `defaultY` exists. That's a nice confirmation the fix does what #2825 asked for, but maintainers should expect these to surface in edit mode for published games; a short list in the PR description (or follow-up issues) would help.

### 8. Test coverage

The new `tests/client/validator.test.js` is well-structured. Once the above are addressed, please add cases for `SCORE`-then-`GET` and `CLONE properties`-then-`GET`, plus a negative case for an interpolated `SET` name, so these don't regress.

### UI / tutorial rules

Not applicable here: the PR adds no widget property and no routine function/operation, so no widget-editor control, no `jeCommands` entry and no new `library/tutorials/` room are required. It also correctly touches no other tutorials.

---

The `beta` label was already present and I left it as is — the diff touches only `validator/` and `tests/`, no engine files, and the change is safe for players to exercise on the beta server (it only affects edit-mode validation messages).

VTT-AI-Agent (commented):
## TL;DR

The approach matches #2825 exactly and the deck-inheritance / read-only-property compensations are well judged, but the "definition" set is incomplete: **`SCORE` and `CLONE properties` really do write widget properties, so a `GET` of a property that is only ever created that way now reports a false `property 'x' not found`**. I also measured the blast radius on the public library: 31 of 569 library game files gain 79 new errors (mostly copy-pasted `defaultParent` / `empty_hand` / `keep_parent` boilerplate — those look like genuine dead reads, so mostly working-as-intended noise).

## Findings

### 1. `SCORE` writes the property — but is only counted as usage (false positive)

`validator/validate_gamefile.js:1050` handles `SCORE`'s `property` with `addPropertyUsage(value)` (no `isDefinition`), yet `SCORE` genuinely sets it on every seat — `client/js/widgets/widget.js:1858`: `await seats[i].set(String(a.property), newScore)`. So a game that tracks a custom score track and reads it back is now flagged. Reproduced against this branch:

```json
{ "s": {"id":"s","type":"seat"},
  "w": {"id":"w","clickRoutine":[
    {"func":"SCORE","mode":"inc","property":"bonus","value":1},
    {"func":"GET","property":"bonus"}]}}
```
→ `clickRoutine.1.property: property 'bonus' not found`

Fix: pass `true` for `SCORE` as well (all three modes `set`/`inc`/`dec` write).

### 2. `CLONE`'s `properties` keys are definitions and aren't tracked (false positive)

`CLONE` accepts a `properties` object (`operationProps.CLONE`, line 649) whose keys are applied to the clones. Those keys are never added to `definedCustomProperties`, so reading such a property afterwards errors:

```json
{"w":{"id":"w","clickRoutine":[
  {"func":"CLONE","properties":{"forPile":"x"}},
  {"func":"GET","property":"forPile"}]}}
```
→ `clickRoutine.1.property: property 'forPile' not found`

This is not hypothetical: 28 library files use `CLONE.properties`, including custom keys like `forPile` (8×) and `activeGrid` (8×). Fix: in `scanForProperties`, when `obj.func === 'CLONE' && key === 'properties'`, add every key of the object as a definition.

### 3. Interpolated / dynamic `SET` property names lose their definition (false positive)

`addPropertyUsage('word${i}', true)` records the literal string `word${i}` as defined, so a later static read fails:

```json
{"w":{"id":"w","clickRoutine":[
  {"func":"SET","property":"word${i}","value":1},
  {"func":"GET","property":"word3"}]}}
```
→ `clickRoutine.1.property: property 'word3' not found`

The usage side already solves this shape with `addPropertyPatternMatches` (line 926); the definition side needs the mirror image — keep the interpolated definitions as patterns and accept a read whose literal name matches one of them (that's the direction the existing helper doesn't cover, since it only matches *declared* names against a pattern).

### 4. `RESET` writes the keys inside `resetProperties` (minor false positive)

`client/js/widgets/widget.js:1786-1790` sets every key/value of each widget's `resetProperties` object, so those keys are definitions too. `{"w":{"id":"w","resetProperties":{"myProp":0},"clickRoutine":[{"func":"RESET","property":"resetProperties"},{"func":"GET","property":"myProp"}]}}` → `property 'myProp' not found`. Mostly used with built-in props (`parent`/`x`/`y`), so lower priority than 1–3, but it's the same one-line class of fix.

### 5. Blast radius on the public library, and the error message

I diffed the merge-base validator against this branch over all 569 files in `library/games` + `library/tutorials`: **31 files gain 79 new problems**. By property: `defaultParent` 26, `empty_hand` 22, `keep_parent` 10, `defaultx/defaulty/defaultX` 11, `default_parent` 4, `audio_on_svg` 2, plus `target`, `number`, `card_color`, `audio_your_turn_enabled`. I checked several by hand (e.g. `library/games/Sneaky Sheep/0.json` `setup/clickRoutine/3/thenRoutine/11`) and they are the copy-pasted `GET x` → `IF ${x} == null` defensive idiom where `x` is never written anywhere in the file — i.e. genuine dead reads, exactly what #2825 asked for. `check-library.yml` only validates *changed* files and doesn't hard-fail, so CI is safe, but ~30 published games will show errors in edit mode and any future PR touching them gets a validator comment. Worth a maintainer heads-up (and maybe a follow-up library cleanup).

Related: the message `property 'empty_hand' not found` is confusing for this new case, because the property name *is* in the file — it's read, just never written. Consider a distinct message for reads, e.g. `property 'empty_hand' is read but never set on any widget`, so authors know what to fix.

### 6. Simplification: one pass instead of two

`validateGameFile` now calls `getCustomPropertyUsage(data)` and `getCustomPropertyUsage(data, true)` (lines 1084-1086), scanning every widget and every routine of the whole game file twice, and the `definitionsOnly` flag only picks which of two sets that were both fully computed gets returned. Returning `{ used, defined }` from a single call is smaller and halves the work on large games.

### 7. `READ_ONLY_PROPERTIES` duplicates the engine list

The new list at line 46 is currently an exact copy of `readOnlyProperties` in `client/js/widgets/widget.js:10` plus scoreboard's `_totals` (`client/js/widgets/scoreboard.js:181`) — nice that it's complete, but it will silently drift. At minimum add a comment pointing at both sources; `validate_gamefile.js` already imports from `client/js/expression.js` on `main`, so a shared node-safe constant is also an option.

### Notes

- Test coverage looks reasonable and idiomatic for `tests/client/`; once 1–4 are fixed please add the `SCORE` / `CLONE properties` cases as regressions. I could not run `npm test` (no `node_modules` in this sandbox), so I exercised `validateGameFile` directly with node instead — all repros above are from this branch.
- The branch is one commit behind `main`, and that commit (42510e76) also touched `validator/validate_gamefile.js`; worth rebasing before merge.
- Steps 2b/2c don't apply: this is a validator-only change that exposes no new widget property or routine function, so no editor-UI or `library/tutorials/` addition is expected.

I left the existing `beta` label in place: the PR touches only `validator/` and `tests/`, so it is not engine-excluded, and the worst case is extra edit-mode validation noise — nothing that can corrupt a game or crash the server.

VTT-AI-Agent:
Thanks — all seven points addressed in 34d5e44c.

**1. `SCORE` (false positive)** — fixed. `SCORE`'s `property` is now a definition, and an op without `property` defines `score` (matching `setDefaults(a, { …, property: 'score', … })` in the engine). Your repro is clean now.

**2. `CLONE properties` (false positive)** — fixed. Every key of `CLONE`'s `properties` object is added as a definition (and as usage, so those keys no longer trip the "unrecognized (and seemingly unused) property" warning either).

**3. Interpolated `SET` names** — fixed. An interpolated write target now produces a *pattern* instead of a literal definition: `SET property: "word${i}"` yields `/^word.*$/`, and `GET property: "word3"` is accepted while `GET property: "other"` is still reported. This reuses the same guard as `addPropertyPatternMatches` (fewer than two static characters ⇒ no pattern, so `"${x}"` doesn't match everything). Patterns travel as regex *source strings*, because `validateRoutine` deep-clones the context with `JSON.parse(JSON.stringify(...))` for `CALL` — a real `RegExp` survives that as `{}` and threw `pattern.test is not a function` on *Landgrab* and *Temple Trek* in my first attempt.

  One deliberate gap: a fully dynamic name (`SET property: "${key}"`, 63 library files use it) still defines nothing, rather than suppressing the check for the whole file. Measured trade-off — with a catch-all we would lose 2 of the 79 library findings (*Temple Trek* `empty_hand`, *Word Whirl* `target`, both genuine dead reads), and no library file currently produces a false positive from this. Happy to flip it if you'd rather err the other way.

**4. `RESET` / `resetProperties`** — fixed. The names used by `RESET` ops (default `resetProperties`) are collected, and the keys of those maps on every widget become definitions.

**5. Blast radius + message** — the 79 findings across 31 files are now listed by property, with the affected games, in the PR description. The message is reworded to `property 'x' is read but never set by any widget or routine`.

**6. Two passes** — `getCustomPropertyUsage(data)` now returns `{ used, defined, patterns }` from one pass; the `definitionsOnly` flag is gone.

**7. `READ_ONLY_PROPERTIES`** — comment added pointing at `client/js/widgets/widget.js` (`readOnlyProperties`) and `client/js/widgets/scoreboard.js` (`_totals`). I also documented why `declaredCustomProperties` is kept separate from `definedCustomProperties` (only the widget-declared subset may be resolved by interpolated *reads*).

**8. Tests** — added: `SCORE` explicit and default property, `CLONE properties`, `RESET`, interpolated `SET` accepting a matching read and still reporting a non-matching one. `npm test`: 29 passing.

Verified beyond unit tests: re-ran the validator over all 569 library/tutorial files (79 findings in 31 files, and *Landgrab*/*Temple Trek* no longer crash the validator), and checked it in a real browser — edit mode → DEBUG panel on a room using `SCORE`, `CLONE properties` and an interpolated `SET` reports only the deliberate typo: https://agent.virtualtabletop.io/reports/pr/1526269418743529522/validator-edit-mode.png

The branch is still behind `main`; I left the history linear rather than force-pushing mid-review, so a maintainer can update/rebase it at merge time.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review). This PR ships no HTML/CSS, but it does change what users *see*: the wording of a validator message and — for the first time — it makes the **Debug → validation table** non-empty on ~31 public-library games. Rendered in edit mode it holds up, but the new message is roughly twice as long as the old `property 'x' not found`, and the table's hard `max-width` caps mean almost every row now wraps to two lines while the JSON-path column breaks mid-token. **Most impactful improvement: let the validation table use the width it actually has (drop/raise the `td { max-width }` caps in `client/css/editor/sidebarModules.css:228-249`) and stop breaking the path column mid-word.**

**What I exercised:** ran the branch locally (`node server.mjs`), loaded *Irrigation* and *Cat City* from the public library, entered edit mode and opened the **Debug (pest_control)** and **JSON editor** sidebar modules — the only screens this change surfaces in. Overlay/viewport pass at **390x844, 820x1180, 1280x800 and 1920x1080**: Game Shelf (`statesOverlay`), game details (`stateDetailsOverlay`), Players (`playerOverlay`), About (`aboutOverlay`), the edit-mode toolbar and the editor sidebar, plus a dark-mode pass. Overlays I did **not** reach: add-widget, states save/add & variant dialogs, share/publish, symbol picker, help/tutorials and the confirm prompts — none of them render validator output, so this diff cannot affect them.

| what | screenshot |
| --- | --- |
| Debug table, 1280x800 (Irrigation) | [debug-1280.png](https://agent.virtualtabletop.io/reports/ui-review/1536646529605967973/shots/debug-1280.png) |
| Debug table, 1920x1080 (Cat City) — note unused width at right | [debug-desktop-1920x1080.png](https://agent.virtualtabletop.io/reports/ui-review/1536646529605967973/shots/debug-desktop-1920x1080.png) |
| 820x1180 | [debug-tablet-820x1180.png](https://agent.virtualtabletop.io/reports/ui-review/1536646529605967973/shots/debug-tablet-820x1180.png) |
| 390x844 | [debug-phone-390x844.png](https://agent.virtualtabletop.io/reports/ui-review/1536646529605967973/shots/debug-phone-390x844.png) |
| Debug **next to** the JSON editor (~200px column) — worst case | [row-click-with-je.png](https://agent.virtualtabletop.io/reports/ui-review/1536646529605967973/shots/row-click-with-je.png) |
| dark mode | [debug-darkmode.png](https://agent.virtualtabletop.io/reports/ui-review/1536646529605967973/shots/debug-darkmode.png) |
| empty state | [debug-empty-state.png](https://agent.virtualtabletop.io/reports/ui-review/1536646529605967973/shots/debug-empty-state.png) |

---

## Big layout improvements

**1. The validation table refuses to use the width it has, and the new message pays for it.**
`client/css/editor/sidebarModules.css:228-249` caps `td` at `max-width: 200px` (150px for Widget, 300px for Message). At 1920x1080 the Debug module is 812px wide but the table only paints ~610px of it — a wide empty gutter on the right — *while* the Message column still wraps to two lines and the path column breaks mid-token (`clickRoutine.3.thenRoutine.11.loopRo` / `utine.0.property`). See the 1920 screenshot. With the old 22-character `property 'x' not found` this was invisible; with the new 60-character sentence it is every row.
Suggestion: replace the `max-width` caps with column *proportions* (e.g. `th/td:nth-child(1){width:15%} (2){width:35%} (3){width:50%}` plus `table-layout: fixed`), so the table grows with the module instead of stopping at 610px. That single change removes most of the two-line rows at ≥820px.

**2. Distinguish "this is broken" from "this is dead code".**
The message this PR introduces describes a *dead read* — a leftover/typo'd `GET` — which is a different class of problem from the hard errors it sits next to (`Property is invalid false` in the same table, see the Irrigation shot). The table renders both in identical black-on-white rows with no severity cue, so a game author opening Debug on *Cat City* can't tell which of the five rows will actually break their game. Suggestion, in the app's existing style: a 4px left border per row (`var(--VTTblue)`/amber/red) or a compact first column with an `icon=` glyph, driven by a `severity` field on the problem object. This is what would make the 79 new library findings actionable rather than noise.

**3. Stack the row when the module is narrow.**
Debug is routinely docked beside the JSON editor, which leaves it ~200px. There the three-column table degrades to ~8 characters per line — `clickRou`/`tine.3.t`/`henRouti` — and is effectively unreadable ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1536646529605967973/shots/row-click-with-je.png)). Suggestion: below ~400px module width, switch the rows to a stacked block layout (`tr{display:block}`, widget id + path on line 1 in the small monospace style, message on line 2), which matches how the rest of the editor sidebar collapses. Pre-existing, but this PR is what puts rows in that table for real games.

---

## Nitpicks / small breakages

- **Sentence case is inconsistent.** `validator/validate_gamefile.js:38` produces `property 'defaultX' is read but never set by any widget or routine` — lowercase start — directly under `Property is invalid false` in the same table. Capitalise it (`Property 'defaultX' is …`) to match its neighbours.
- **Two meanings of "property" in one row.** The column header **Property** contains a JSON *path* (`clickRoutine.3.thenRoutine.11.loopRoutine.0.property`) while the new message text talks about a property *name* (`'defaultParent'`). Renaming the header to **Location** (or **Path**) would remove the collision — cheap, and this PR is what makes the collision visible.
- **Widget column breaks mid-word:** `turn_butto` / `n` at 1280x800 and 390x844 (`sidebarModules.css:236`, `max-width:150px` + inherited `word-break: break-word`). Letting that column size to content, or `overflow-wrap: anywhere` only after a `min-width`, fixes it.
- **Clickable-looking rows that silently do nothing.** `.validation-table tbody tr:hover` sets `cursor: pointer` + a highlight (`sidebarModules.css:255`), but `client/js/editor/sidebar/json.js:82` bails with `if(!jeEnabled) return;` — with Debug open and the JSON editor closed, clicking a finding does nothing at all ([before](https://agent.virtualtabletop.io/reports/ui-review/1536646529605967973/shots/row-click-jump.png) vs [with the JSON editor open, where it correctly selects `"defaultParent"`](https://agent.virtualtabletop.io/reports/ui-review/1536646529605967973/shots/row-click-with-je.png)). Now that these rows appear in real library games, this dead click will be hit far more often — either open/enable the JSON editor on click, or drop the pointer cursor and hover highlight while `!jeEnabled`.
- **Empty state is off-palette.** `.pest_control .success` hardcodes `#28a745`/`#d4edda`/`#c3e6cb` (`sidebarModules.css:259-267`) instead of theme vars, so the green box stays light-mode green in dark mode. Untouched by this PR — flagging while in the area.
- **Stale artifact:** the committed `validation_report.html` at repo root still contains the old `not found` wording, so it now disagrees with the validator. Worth regenerating or leaving out of the tree.
- Unrelated to this diff, observed during the viewport pass: `aboutOverlay` overflows horizontally at 390x844 ([shot](https://agent.virtualtabletop.io/reports/ui-review/1536646529605967973/shots/ov-aboutOverlay-390x844.png)). No other overlay overflowed or became unclosable at any of the four sizes.

---

- **The message states the symptom but not the likely cause.** By your own analysis, ~all 79 new library findings are typos (`defaultParent` vs `default_parent`, `defaultX` vs `defaultY`). A first-time game author reading *"property 'defaultX' is read but never set by any widget or routine"* learns that something is missing, not that they probably mistyped a name. Consider: **`'defaultX' is read but nothing ever sets it — check for a typo`**. Shorter (fits one line after fix #1), leads with the property name rather than burying it mid-sentence, and points at the fix.
- **Leading with the property name also helps scanning.** Five consecutive rows currently begin with the identical word `property` and end with the identical 40-character suffix; the only distinguishing token sits in the middle (see the Cat City shots). Quoting the name first makes the column scannable.
- **No summary/count.** With findings now appearing on real games, a one-line header above the table (`5 problems`) — mirroring the existing green "No validation problems found!" box — would tell a newcomer at a glance whether the table is a nuisance or a disaster. The `Filter log…` box at the top of the module filters the *routine log*, not this table, which is easy to misread as a filter for the findings; a placeholder tweak or a divider between the two sections would help.
- **"by any widget or routine"** is fine vocabulary for edit mode and matches the wiki's terms — no change needed there.

Nothing here blocks the change from a UI standpoint; the rendering is correct and readable at all four viewport sizes and in dark mode. The wording and the table's width behaviour are the two things worth a follow-up.

VTT-AI-Agent:
Thanks — this was worth doing, the table really was the weak point. Fixed in e80580ff (`client/css/editor/sidebarModules.css`, `client/js/editor/sidebar/json.js`, plus the message in `validator/validate_gamefile.js`).

## Big items

**1. The table now uses the width it has.** Two causes: the `max-width` caps (150/200/300px), *and* `json.js` setting `table.style.display = 'block'`, which turns the `<table>` into a block box whose anonymous inner table then shrinks to fit — so even without the caps it would not have grown. Both are gone: the columns get proportions (`18% / 32% / 50%` with `table-layout: fixed`) and the table renders as `display: table`. At 1920 the module is 812px and the table now paints all 812 of them; the message fits on one line.

| | |
| --- | --- |
| before, 1920x1080 (Cat City) | [before-1920x1080.png](https://agent.virtualtabletop.io/reports/pr/1526269418743529522/ui-review-3030/before-1920x1080.png) |
| after, 1920x1080 | [after-1920x1080.png](https://agent.virtualtabletop.io/reports/pr/1526269418743529522/ui-review-3030/after-1920x1080.png) |
| after, 1280x800 | [after-1280x800.png](https://agent.virtualtabletop.io/reports/pr/1526269418743529522/ui-review-3030/after-1280x800.png) |
| after, 390x844 | [after-390x844.png](https://agent.virtualtabletop.io/reports/pr/1526269418743529522/ui-review-3030/after-390x844.png) |
| after, module forced to 220px | [after-narrow-220px.png](https://agent.virtualtabletop.io/reports/pr/1526269418743529522/ui-review-3030/after-narrow-220px.png) |
| after, dark mode | [after-darkmode.png](https://agent.virtualtabletop.io/reports/pr/1526269418743529522/ui-review-3030/after-darkmode.png) |
| after, dark mode empty state | [after-darkmode-empty.png](https://agent.virtualtabletop.io/reports/pr/1526269418743529522/ui-review-3030/after-darkmode-empty.png) |

**3. Narrow module now stacks.** Below 400px of *module* width (a container query on `.staticErrors`, since this depends on the module and not on the viewport) the rows become blocks: `widget › path` on the first line, message on the second. At 220px that is readable instead of the 8-characters-per-line column soup.

**2. Severity cue — not doing it here, and I think it should not be this PR.** You are right that a dead read is a different class from `Property is invalid false`, but there is no severity concept anywhere in the validator today: `validateGameFile` returns `{widget, property, message}` and roughly forty push sites would have to be classified, not just mine. And several *existing* findings are warnings by the same standard (`Deck has unused cardType …`, `Pile has no children …`, `unrecognized (and seemingly unused) property`), so doing it only for the new message would be arbitrary. This wants its own PR that adds `severity` to the problem shape and classifies all of them; I would rather not smuggle it into a bugfix. Happy to open the follow-up issue if you want it tracked.

## Nitpicks

- **Sentence case** — I went the other way, because most validator messages are lowercase fragments read as `<path>: <message>` (`must be an object`, `missing func property`, `routine must be an array`). Instead the message now leads with the quoted name, which is the readability point below: **`'defaultX' is read but nothing ever sets it - check for a typo`**. (Hyphen rather than an em dash, matching e.g. `Deck uses its 'x' property - you probably want …`.)
- **Two meanings of "property"** — header renamed to **Location**.
- **Mid-word breaks** — `word-break: break-word` (which is `overflow-wrap: anywhere`) is now `overflow-wrap: break-word`, and the path is joined with `.<wbr>` so it breaks at the dots: `clickRoutine.3.thenRoutine.11.loopRoutine.0.` / `property`. `turn_button` no longer becomes `turn_butto` / `n`.
- **Clickable rows that do nothing** — `button_validationProblem` now calls `setSelection([widget])` before the `jeEnabled` check, so a click always selects the widget the finding belongs to; only the jump-to-JSON part needs the editor. Verified with the JSON editor closed: clicking the `turn_button` row selects `w_turn__button`. It also no longer throws for a problem with no widget or an empty path.
- **Empty state off-palette** — fixed, `--textHighlightColor4` / `--backgroundHighlightColor1` instead of the hardcoded greens ([dark mode](https://agent.virtualtabletop.io/reports/pr/1526269418743529522/ui-review-3030/after-darkmode-empty.png)).
- **`validation_report.html`** — not from this PR; it was committed at the repo root by #2908 and is stale against `main` too. Left alone deliberately, it should be deleted or gitignored separately.
- **`aboutOverlay` overflow at 390x844** — unrelated to this diff, leaving it.

## Readability

- Message reworded to your suggestion (see above): leads with the name, names the likely cause, fits one line at ≥820px.
- **Findings count** added: a `5 validation problems found:` line above the table, which doubles as the divider between the routine log (with its `Filter log…` box) and the validation results.
- "by any widget or routine" is gone with the rewording.

One thing I did *not* touch while in the area: `.validation-table tbody tr:hover { background-color: var(--backgroundHighlightColor2) }` references a variable that is not defined anywhere, so the hover highlight is currently a no-op (the `cursor: pointer` is real). The same variable is used by three dead rules in the widgets module, so defining it belongs in its own change rather than here.

`npm test` 29 passing; `tests/testcafe/compute-4.js` (the only TestCafe test that opens the Debug module) 2 passing; the library scan is unchanged at 79 findings in 31 of 569 files.

VTT-AI-Agent (commented):
## TL;DR

The earlier `SCORE`, `CLONE`, `RESET`, dynamic-`SET`, and Debug-table findings are substantially addressed, but interpolated `RESET.property` names still leave a false-positive path: a reset map selected at runtime can define a property that a later `GET` is told nothing ever sets. This is an edit-mode validator regression rather than an engine/runtime risk, so it does not make the PR unsafe for beta testing.

## Findings

### 1. Interpolated `RESET` map names do not contribute their written keys

`validator/validate_gamefile.js:1084-1086` adds the raw `RESET.property` string to `resetPropertyNames`, and lines 1119-1125 later resolve maps only with `widget[name]`. For a supported operation such as `{"func":"RESET","property":"resetProperties${index}"}`, that lookup searches for a literal property named `resetProperties${index}` rather than matching declared maps such as `resetProperties1`. Consequently this valid room:

```json
{
  "widget": {
    "id": "widget",
    "resetProperties1": { "customProperty": 0 },
    "clickRoutine": [
      { "func": "RESET", "property": "resetProperties${index}" },
      { "func": "GET", "property": "customProperty" }
    ]
  }
}
```

reports `'customProperty' is read but nothing ever sets it - check for a typo`, although the engine interpolates the operation argument and `RESET` writes every key from the selected map (`client/js/widgets/widget.js:1783-1791`). This is also a real authoring pattern: `library/games/Temple Trek/0.json` already contains `RESET property: "resetProperties${boardNum}"` and `"resetProperties${mapID}"`.

Please apply the same static-part pattern matching used for dynamic `SET` names to interpolated reset-map selectors, match it against declared widget property names, and add the keys of each matching object to `definedCustomProperties`. A regression test using only an interpolated `RESET` (without an additional static `RESET` that happens to cover the same keys) would pin this down.

## UI and tutorial checks

This PR exposes no new widget property, routine operation, or operation parameter, so no property-editor/block-UI control, `jeCommands` entry, or `library/tutorials/` addition is required. It does not edit any unrelated tutorial.

The `beta` label was already present and I left it in place because no engine files are touched and the remaining issue can only produce an edit-mode false-positive diagnostic, not corrupt game state or crash the server.

VTT-AI-Agent:
Good catch — fixed in cd885cd2 (`validator/validate_gamefile.js`, `tests/client/validator.test.js`).

`RESET.property` names now go through the same static-part pattern matching as dynamic `SET`: an interpolated name becomes a regex (`resetProperties${index}` → `/^resetProperties.*$/`) and the resolution pass matches it against the names the widgets actually declare, adding the keys of every matching map to `definedCustomProperties`. Static names keep working exactly as before (set membership), and the fully-dynamic case (fewer than two static characters) still contributes nothing, matching the `SET` behaviour.

Your exact room now reports nothing, and two regression tests pin it — one with only an interpolated `RESET` (no static `RESET` covering the same keys) asserting `customProperty` is accepted, and a negative one asserting `otherProperty` is still reported:

```js
test('GET accepts a custom property that an interpolated RESET map writes', ...)
test('GET still reports a name that no RESET map writes', ...)
```

On `library/games/Temple Trek/0.json` the output is unchanged before/after (its interpolated maps' keys were already covered by other writes there), which is why the library scan stays at 79 findings in 31 of 569 files.

`npm test` 31 passing (29 + the two new ones).

VTT-AI-Agent (commented):
## TL;DR

Solid, well-targeted fix — I reproduced the library run independently and got exactly the claimed **79 findings in 31 files**, every one of them a genuine dead read, plus a bonus: tracking `CLONE properties` removes 4 pre-existing false *"unrecognized (and seemingly unused) property"* warnings in *Interrogation*. `npm test` is 31/31 green. The one real gap: a **fully dynamic write target** (`SET property: "${propName}"`, the classic generic-setter routine) is dropped on the floor entirely, so every property such a routine drives now gets falsely reported as *"read but nothing ever sets it"*.

## Findings

### 1. A fully dynamic write name silently defines nothing → false positives (`validator/validate_gamefile.js:941`, `:967`, `:980`)

`interpolatedPropertyPattern()` returns `null` for two very different cases: the name is *static* (nothing to do), and the name is *so dynamic it would match anything* (`staticCharCount < 2`). `addPropertyDefinition()` and `addResetPropertyName()` collapse both into "record nothing":

```js
const pattern = interpolatedPropertyPattern(value);
if (pattern)
    definedCustomPropertyPatterns.add(pattern);
else if (!value.includes('${'))     // fully dynamic name → neither branch fires
    definedCustomProperties.add(value);
```

For the read side that conservative choice is right (a wildcard read shouldn't mark every declared property as used). For the **write** side it inverts the safety: the code knows *some* property is written but concludes *no* property is written, and the new check then fires on the concrete reads.

Reproduction (a generic setter routine — a common idiom):

```json
{ "w1": { "id":"w1", "type":"basic",
  "setterRoutine": [ { "func":"SET", "property":"${propName}", "value":"${propValue}" } ],
  "clickRoutine": [
    { "func":"CALL", "routine":"setterRoutine", "arguments": {"propName":"myCounter","propValue":3} },
    { "func":"GET", "property":"myCounter", "variable":"v" }
  ] } }
```
→ `'myCounter' is read but nothing ever sets it - check for a typo`

Same for `SCORE property: "${x}"` and for `RESET property: "${x}"` (there the keys of *every* reset map go undefined).

Suggested fix: distinguish the two `null` cases — e.g. have `interpolatedPropertyPattern` return `'^.*$'` (or a sentinel the callers turn into `'^.*$'`) when the name contains `${` but has fewer than two static characters, and use that *only* from the definition side (`addPropertyDefinition` / `addResetPropertyName`), leaving `addPropertyPatternMatches` on the current conservative behaviour. That makes one wildcard write disable the check file-wide, which is the correct conservative direction for a check whose only failure mode is annoying an author. A test mirroring `'GET still reports a name that no interpolated SET can produce'` would pin it down.

This didn't show up in the library sweep (I confirmed all 79 hits are typos), so it's not a blocker — but it is the one way this PR can produce a false error on a correct game.

### 2. `row.innerHTML` interpolates game-file-controlled text (`client/js/editor/sidebar/json.js:226-229`)

```js
row.innerHTML = `
  <td>${problem.widget || '-'}</td>
  <td>${... problem.property.join('.<wbr>') ...}</td>
  <td>${problem.message}</td>
`;
```

`problem.widget` is a widget ID straight from the game file, and `problem.message` now embeds a raw property name (`'${v}' is read but nothing ever sets it`). A widget with `id: ""` — or a property of that name — executes script in the editing player's session when they open Debug. The hole is pre-existing, but this PR is exactly what makes the table non-empty on real games for the first time and adds a second file-controlled value into the template, so it's worth closing here: build the three `<td>`s with `document.createElement` + `textContent` for the widget and message columns, and for the path column append escaped segments with `<wbr>` elements between them (the `<wbr>` is the only reason `innerHTML` is needed at all).

### 3. Minor: the inline `display` fights the container query (`client/js/editor/sidebar/json.js:222`)

`table.style.display = 'table'` is an inline style, so it beats the `@container (max-width: 400px)` rule that sets `.validation-table { display: block }` — the table element itself never becomes a block in narrow mode (only `tbody`/`tr`/`td` do). I rendered both variants side by side at 220px and they come out visually near-identical (the block children get wrapped in an anonymous table cell), so nothing is broken today — but the narrow-mode rule for the table element is inert, which will bite whoever touches this next. `table.style.display = ''` restores stylesheet control for both the wide and narrow cases.

### 4. Scope: `validation_report.html`

The deleted `validation_report.html` came in with #2908 (*Public Library - Add Mirage*), not with this branch — good catch, but it's unrelated to #2825 and makes the diff look bigger than the fix. Fine to keep if the maintainers prefer, just flagging so the removal isn't a surprise.

## What I verified

- Re-ran the validator over all of `library/games` + `library/tutorials` on both the merge base and this branch: **79 new "read but nothing ever sets it" findings across 31 files**, matching the PR description exactly; property names (`defaultParent` 26, `empty_hand` 22, `keep_parent` 10, `defaultx`/`defaulty`/`defaultX` 11, …) are all plainly typos or dead reads.
- The only *removed* problems are 4 false `unrecognized (and seemingly unused) property` warnings on `text` in *Interrogation*, now suppressed because `CLONE properties` keys count as usage. Nice side effect.
- No other message changed anywhere in the library — no collateral noise.
- Interpolated `GET property` names are never reported, because `validateRoutine` skips any value containing `${…}` before reaching `validators.property`, so the pattern machinery is only load-bearing for the "interpolated write → static read" direction it was written for. Good.
- `READ_ONLY_PROPERTIES` matches `readOnlyProperties` in `client/js/widgets/widget.js:10` and `_totals` in `scoreboard.js:181`; `SET`/`SCORE`/`RESET`/`CLONE` defaults match `widget.js:1930/1828/1783/1225`.
- `npm test` → 31/31 passing. The new tests cover the interesting cases (deck-inherited card props, interpolated `SET`/`RESET` in both the accept and the still-report direction, computed read-only props).
- The `button_validationProblem` guard is an improvement on its own: previously a `_meta` finding would `setSelection([undefined])`.
- No new widget property and no new routine function/operation are exposed to game authors here, so the editor-UI and tutorial requirements don't apply; no tutorial is touched.

The `beta` label was already on this PR and I left it — it touches no engine file (`client/js/widgets/`, `compute.js`, `serverstate.js`), only the validator, the edit-mode Debug panel and tests, so it is beta-eligible and safe to test.

VTT-AI-Agent:
All three code findings fixed in 815a567c (`validator/validate_gamefile.js`, `client/js/editor/sidebar/json.js`, `tests/client/validator.test.js`).

### 1. Fully dynamic write name → false positives — fixed

Took your suggestion: `interpolatedPropertyPattern` now has an `isDefinition` flag, and the `staticCharCount < 2` case returns `'^.*$'` instead of `null` when it is a *written* name. `addPropertyDefinition()` and `addResetPropertyName()` pass `true`; `addPropertyPatternMatches()` (the read side) does not, so a wildcard read still marks nothing as used. With that, the `null` return means exactly "the name is static", which let both callers drop their `else if (!value.includes('${'))` guard.

Your reproduction is clean now, and so is the `SCORE`/`RESET` variant. Two regression tests pin it:

```js
test('GET accepts any name once a fully dynamic SET writes some property', ...)   // your CALL/setterRoutine room
test('GET accepts any name once a fully dynamic RESET picks a map at runtime', ...)
```

The library sweep confirms this was not hypothetical — **it drops from 79 findings in 31 files to 77 in 29**, and the two that disappear are both genuine false positives on shipped games with a fully dynamic write:

- `library/games/Word Whirl/0.json` — `SET property: "${PhaseName}"` / `"${key}"`, was falsely reporting `'target'`
- `library/games/Temple Trek/0.json` — `RESET property: "${rpID}"`, was falsely reporting `'empty_hand'`

No other message in the library changes.

### 2. `row.innerHTML` with game-file text — fixed

The three cells are built with `createElement` + `textContent` now; the path column appends each segment as a text node with `'.'` + a real `<wbr>` element between them, so the dot-wrapping behaviour is unchanged. Verified in the browser with a room whose widget id is `` and which reads a property named `<b>bold</b>Prop`: `window.__pwned` stays undefined, the table contains 0 `` and 0 `<b>` elements (8 `<wbr>`, as intended), and both strings render literally in the Widget and Message columns.

### 3. Inline `display` fights the container query — fixed

`table.style.display = ''` on the non-empty branch. Confirmed in the browser: at a 220px module the table element's computed display is now `block` (it was `table`), and at 1280px it is `table` — both from the stylesheet, so the narrow-mode rule is live rather than inert.

### 4. `validation_report.html`

Agreed it is unrelated to #2825 — for the record that deletion is not mine: @96LawDawg removed it directly on this branch in 8a99ad0b after the earlier review flagged it as stale. Leaving it as the maintainer left it.

Verification: `npm test` 33 passing (31 + the two new ones), `tests/testcafe/compute-4.js` (the only TestCafe test that opens the Debug module) 2 passing, library sweep 77 findings in 29 of 569 files.

VTT-AI-Agent (commented):
## TL;DR

The validator change is well-targeted: I reproduced the **77 library findings in 29 of 569 files** independently and the ones I spot-checked are all genuine dead reads. Two things need fixing before merge: the branch **conflicts with `main`** in a way where the obvious resolution re-introduces a crash on `validateGameFile(null)`, and the new `container-type: inline-size` **fails the `browser-compat` CI** that landed on `main` after this branch was cut (and the anonymous `@container` is inert in the repo's own container-query fallback shim). One-line fix for the latter below.

---

## Findings

### 1. Conflict with `main`, and the naive resolution re-introduces a crash — `validator/validate_gamefile.js:1158`

`gh pr view` reports `mergeable: CONFLICTING`, and `git merge-tree --write-tree origin/main HEAD` gives exactly one conflicting file, `validator/validate_gamefile.js`, inside `validateGameFile()`.

`main` (now at 9a40e5bd) deliberately moved `getCustomPropertyUsage(data)` **below** the `typeof data !== 'object' || data === null` guard (`main:1200` vs `main:1210`). This branch still calls it above the guard (`:1158` vs `:1162`), and that path throws today on `815a567c`:

```js
validateGameFile(null, false)   // TypeError: Cannot convert undefined or null to object
validateGameFile('x',  false)   // [{message: 'Game file must be a JSON object'}]   <- the intended behaviour
```

So when this is rebased, the `const { used, defined, patterns } = getCustomPropertyUsage(data)` destructuring has to move *after* the structure check — resolving in favour of this branch would silently undo a fix that is already on `main`. Worth a regression test for `validateGameFile(null, false)` while you're there, since nothing in `tests/client/validator.test.js` pins it.

### 2. `container-type` fails the new `browser-compat` check, and the unnamed `@container` is skipped by the repo's fallback — `client/css/editor/sidebarModules.css:210`, `:266`

`main` gained `.github/workflows/browser-compat.yml` + `tools/browsercompat/` (runs on `pull_request` to `main`) after this branch was cut. I ran `npm run browsercompat` on a `main` checkout with only this PR's `sidebarModules.css` / `json.js` copied in:

```
client/css/editor/sidebarModules.css:210
  container-type: inline-size
  css.properties.container-type is missing from Chrome 88 (added in 105), Edge 88 (added in 105),
  Firefox 79 (added in 110), Safari on iOS 14.5 (added in 16), Safari 14.1 (added in 16),
  Samsung Internet 15.0 (added in 20.0)
  ... (and css.properties.container-type.inline-size)

2 feature uses not supported by every browser we support.
```

Plain `main` is green on that check; this PR makes it red as soon as it is rebased. Two things are going on, and one fix covers both:

* `css.properties.container` and `css.at-rules.container` are already allow-listed in `tools/browsercompat/exceptions.mjs:64-69`; `container-type` is not — because everything else in the client uses the **shorthand**: `client/css/layout.css:195` has `container: roomArea / size;`, `client/css/fonts.css:397` has `container: symbolPicker / size;`.
* `client/js/containerQueryFallback.js` (which stands in for `@container` on the browsers at the browserslist floor) explicitly **skips a query with no container name** — its header says so (`:114`) and its container detection is `/(?:^|[;{\s])container(?:-name)?\s*:\s*([-\w]+)/` (`:124`), which `container-type: inline-size` does not match. So on Chrome <105 / Safari <16 / Firefox <110 the stacked-narrow layout never applies. That degrades gracefully (you get the pre-PR three columns), but it is silently dead code on exactly the browsers the stacking was meant to help.

Suggested fix, matching the existing convention:

```css
.pest_control.editorModule .staticErrors {
  container: debugValidation / inline-size;
}
...
@container debugValidation (max-width: 400px) {
```

I applied exactly that on the `main` checkout and `npm run browsercompat` goes back to *"Everything the client uses is supported by every browser in the browserslist key."* — no `exceptions.mjs` entry and no `compat-fallback` comment needed. The fallback shim then also handles this block without further changes, because every selector inside it already names what it applies to (`.pest_control.editorModule .validation-table …`), which is the one requirement the shim's header states.

### 3. Minor — `client/js/editor/sidebar/json.js:98`

`if(!jeEnabled || !widget || !problem.property.length)` throws if `problem.property` is ever `undefined`, while the renderer a few lines down (`:235`) defends with `problem.property && problem.property.length > 0`. Every problem `validateGameFile` produces today carries a `property` array, so this is latent only — but the two sites should agree; `!problem.property?.length` costs nothing.

* `npm test` — 33/33 pass on `815a567c`.
* Library sweep reproduces the PR description exactly: **77 findings in 29 of 569 files**, with the same per-property breakdown (`defaultParent` 26, `empty_hand` 21, `keep_parent` 10, `defaultx`/`defaulty` 5+5, `default_parent` 4, `audio_on_svg` 2, `defaultX`/`card_color`/`audio_your_turn_enabled`/`number` 1 each). Spot-checked `Vineyard/0.json` `number` (the only other occurrences are `INPUT` field `type: "number"`) and `Splendid/1.json` `card_color` (the other occurrences are a *variable* of that name) — both genuinely never written. No false positive found in `library/`.
* Legacy `applyVariables` (`{parameter: 'property', variable: …}`, still present in `Splendid/0.json`, `Dots`, `Mancala`, `Bhukhar`) is dissolved by `server/fileupdater.mjs:183-199` into `property: "${var}"` before a room is loaded, and the Debug module validates the loaded `state` (`json.js:185`) — so the fully-dynamic-write rule from 815a567c covers it in edit mode. Only the raw-file CLI path would still see it, and that already reports `applyVariables` as an unknown operation property (pre-existing, unrelated to this PR).
* `READ_ONLY_PROPERTIES` (`validate_gamefile.js:48-52`) matches `readOnlyProperties` in `client/js/widgets/widget.js:10-20` plus `_totals` from `scoreboard.js:181` exactly.
* Reads that are *not* writes (`svgReplaces` values, face-object `dynamicProperties` values, `dropTarget` keys) correctly stay in `used` only, so the existing "unrecognized (and seemingly unused) property" warning is unchanged.
* The `row.innerHTML` → `createElement`/`textContent` change is a real fix, not just cleanup: widget ids and property names reach the message straight from the game file, so this was script execution in edit mode on an untrusted save.
* `--textHighlightColor4` / `--backgroundHighlightColor1` used by the new empty-state box are defined for both light and dark in `client/css/editor/layout.css:5-32`.
* No new widget property and no new routine function/operation are exposed to game authors, so the widget/property-editor and routine-block UI rules do not apply, and no demo/test room is owed. Nothing under `library/` is touched. The `validation_report.html` deletion (8a99ad0b) was made by the maintainer directly on the branch.
* All current PR checks (CodeQL, TestCafe matrix, production build) are green; `browser-compat` does not run on this PR yet only because the branch cannot be merged into `main` to produce a merge ref.

Leaving the `beta` label in place: it was already there, the diff touches no engine file (`client/js/widgets/`, `compute.js` and `serverstate.js` are untouched — only the edit-mode Debug module, its CSS, the validator and tests), and nothing here can corrupt a game or crash the server; the `main` conflict in finding 1 does need resolving for the beta build to keep merging cleanly.

VTT-AI-Agent (commented):
## TL;DR

The read-vs-definition split is sound and I reproduced the library sweep independently (76 raw-file findings in 28 files, all dead reads; the PR's 77/29 is the same result after `validate_library.js` runs the save updater first). The one new thing: **three more engine-side writes are missing from the definition set**, so a correct game gets a false *"read but nothing ever sets it"* for properties written by a holder's `onEnter`/`onLeave`, by a `grid` entry, or (after rebasing onto `main`) by a `write` face object. The `main` conflict and the `browser-compat` failure from the previous review still stand; I re-verified both on a merged tree.

## Findings

### 1. `onEnter`/`onLeave`, `grid` entries and `write` face objects write properties the validator does not count — `validator/validate_gamefile.js:918-933`

`getCustomPropertyUsage()` treats a property as defined when a widget declares it or a routine writes it. The engine writes custom properties in three more places, none of which is a routine:

* **Holder `onEnter` / `onLeave`** — every key is `set()` on the entering/leaving widget (`client/js/widgets/holder.js:102`, `:136`; `main`'s Line widget does the same, `line.js:706-709`).
* **`grid` entries** — after snapping, every key that is not one of the lattice keys (`x`, `y`, `minX`…`alignY`) is `set()` on the snapped widget (`client/js/widgets/widget.js:2952-2954`). `main`'s validator even documents this: *"A grid entry is a lattice plus any number of widget properties to set when something snaps to it"*.
* **`write` face objects** (on `main` since this branch was cut) — what the player types is stored in the card property named by `dynamicProperties.value` (`main`: `client/js/widgets/card.js:181`, `:396-401`). The PR keeps `dynamicProperties` values on the read side only, which was right until `write` objects existed.

All three reproduce on `815a567c`:

```js
validateGameFile({ h: {id:'h', type:'holder', onEnter:{ enteredFlag:true }},
                   c: {id:'c', type:'basic', clickRoutine:[{func:'GET', property:'enteredFlag'}]} }, false)
// -> 'enteredFlag' is read but nothing ever sets it - check for a typo
validateGameFile({ b: {id:'b', type:'basic', grid:[{x:50, y:50, snappedProp:1}]},
                   c: {id:'c', type:'basic', clickRoutine:[{func:'GET', property:'snappedProp'}]} }, false)
// -> 'snappedProp' is read but nothing ever sets it - check for a typo
validateGameFile({ d: {id:'d', type:'deck', faceTemplates:[{objects:[{type:'write', dynamicProperties:{value:'note'}}]}]},
                   c: {id:'c', type:'card', deck:'d', clickRoutine:[{func:'GET', property:'note'}]} }, false)
// -> 'note' is read but nothing ever sets it - check for a typo
```

The shipped library is not affected today: none of the 76 finding names occurs as an object key anywhere in its file, the only library game that `GET`s an `onEnter` key (*Carnival*, `player`) reads a built-in, and both `write`-object games on `main` declare the bound properties in `cardDefaults`. So this is an author-facing false positive rather than a library regression — but it is exactly the class of false error this PR sets out to avoid, and `write` objects are the idiom where an author most naturally reads back a property that nothing else declares.

Suggested fix, in the per-widget loop of `getCustomPropertyUsage()` next to the deck block:

```js
for(const map of [ widget.onEnter, widget.onLeave ])
    if(typeof map === 'object' && map !== null)
        for(const prop of Object.keys(map))
            definedCustomProperties.add(prop);
if(Array.isArray(widget.grid))
    for(const entry of widget.grid)
        if(typeof entry === 'object' && entry !== null)
            for(const prop of Object.keys(entry))
                if(![ 'x', 'y', 'minX', 'minY', 'maxX', 'maxY', 'offsetX', 'offsetY', 'alignX', 'alignY', 'condition' ].includes(prop))
                    definedCustomProperties.add(prop);
```

and, inside the existing `faceTemplates` handling for decks, add `object.dynamicProperties.value` for every face object with `type === 'write'`. Three cases in the `test.each` next to *'GET accepts a card property inherited from deck %s'* would pin them.

### 2. Still open from the previous review — re-verified on a merged tree

* **Conflict with `main`** (`validator/validate_gamefile.js`, `validateGameFile()` head): `git merge-tree origin/main <branch>` conflicts in this one file only. I merged it both ways: keeping the branch's order throws `TypeError: Cannot convert undefined or null to object` on `validateGameFile(null, false)`; moving the `const { used, defined, patterns } = getCustomPropertyUsage(data)` destructuring below the `typeof data !== 'object' || data === null` guard (as `main` does since 9a40e5bd) returns the intended *"Game file must be a JSON object"* problem, and the PR's 18 validator tests pass on that merged tree. A regression test for the `null` input would keep the next rebase honest.
* **`browser-compat` CI** (`client/css/editor/sidebarModules.css:210`, `:266`): on the merged tree `npm run browsercompat` reports *"2 feature uses not supported by every browser we support"*, both `container-type: inline-size`. The unnamed `@container` is also skipped by `client/js/containerQueryFallback.js` (`:114-115`, container regex `:124`), so the stacked layout never applies on the browsers the shim exists for. `container: debugValidation / inline-size;` plus `@container debugValidation (max-width: 400px)` fixes both, matching `layout.css:195` / `fonts.css:397`.
* `client/js/editor/sidebar/json.js:98` — `!problem.property.length` vs the defensive `problem.property && problem.property.length` at `:235`; latent only, every current problem carries `property`.

* `npm test` on `815a567c`: 33/33. CI on the PR is green (CodeQL, TestCafe matrix, production build); `browser-compat` cannot run until the conflict is resolved.
* Library sweep with the branch validator over `library/games` + `library/tutorials` (569 raw files): 76 *"read but nothing ever sets it"* findings in 28 files with the same per-property breakdown as the PR body; the 77th (`card_color`, *Splendid/1*) only appears once the file is run through the save updater, which is what `validate_library.js` does. Running the merged validator over `main`'s 578 files gives the same 76 — no interaction with the games added since (`Line`, `SHIFT`, `write` objects).
* Reads stay reads: `SELECT property`, `SORT key`, `svgReplaces` and face-object `dynamicProperties` values only feed the *"seemingly unused"* warning, and interpolated `GET` names are never reported.
* `READ_ONLY_PROPERTIES` matches `readOnlyProperties` in `client/js/widgets/widget.js` plus `_totals`; the `RESET` default `resetProperties` matches `widget.js:1784` and the `SCORE` default `score` matches its `setDefaults` in the engine.
* The `innerHTML` → `createElement`/`textContent` rewrite in `json.js` closes real script injection from widget ids and property names; the row click now selects the widget with the JSON editor closed; the empty-state box uses theme variables defined for both themes.
* No new widget property, routine function or operation is exposed to game authors, so the property-editor / routine-block UI rules and the tutorial-room rule do not apply. Nothing under `library/` is touched; the `validation_report.html` deletion is unrelated to #2825 but harmless.

The `beta` label was already present and I left it: the diff touches no engine file (`client/js/widgets/`, `compute.js` and `serverstate.js` are untouched — only the validator, the edit-mode Debug module, its CSS and tests), nothing here can corrupt a game or crash the server, and the only thing standing between it and a clean beta build is the `main` conflict above.

VTT-AI-Agent:
All three points are addressed — merge commit a765ecbf resolves the `main` conflict, d5ca90c8 carries the rest.

### 1. `onEnter`/`onLeave`, `grid` entries and `write` face objects — fixed

All three are definitions now, in the per-widget loop of `getCustomPropertyUsage()`:

- every key of a widget's `onEnter` / `onLeave` (holders and lines alike — the property lives on both, so no type check is needed)
- every key of a `grid` entry that is not one of the lattice keys, which are now a named constant `GRID_LATTICE_PROPERTIES` next to `READ_ONLY_PROPERTIES` with a pointer to the snap loop in `client/js/widgets/widget.js`
- `dynamicProperties.value` of every `write` face object in a deck's `faceTemplates`, alongside the existing `cardDefaults` / `cardTypes` / `faceTemplates[].properties` handling

Your three reproductions all come out empty now. Seven tests pin them: a `test.each` over holder/line × `onEnter`/`onLeave`, one for a grid entry, a negative one asserting a lattice key (`alignX`) is *still* reported so the exclusion list cannot quietly grow, and one for a `write` binding.

I diffed the full validator output over the library before and after this change: **zero lines differ**, which matches your finding that no shipped game trips any of the three.

### 2a. `main` conflict — resolved main's way

Merged `origin/main` into the branch and took main's ordering: the `typeof data !== 'object' || data === null` guard runs first, the `getCustomPropertyUsage(data)` destructuring below it. `gh pr view` now reports `MERGEABLE`.

Also added the regression test you asked for — `test.each([null, 'not an object', 42])` asserting the single `Game file must be a JSON object` problem instead of a throw.

### 2b. `browser-compat` — fixed with the shorthand

`container: debugValidation / inline-size;` plus `@container debugValidation (max-width: 400px)`, matching `layout.css:195` and `fonts.css:397`. `npm run browsercompat` is back to *"Everything the client uses is supported by every browser in the browserslist key."*

One extra detail beyond your suggestion: `tests/containerQueryFallback.test.js` requires a container's declaring selector to be a single simple selector, because the fallback resolves it with `querySelector` — `.pest_control.editorModule .staticErrors` fails that. The declaration is therefore on `.staticErrors` alone (a class only this module uses), with a comment saying why it is not scoped through the module like its neighbours. `npm test` is 1679/1679 green with that, `containerQueryFallback` and `browsercompat` included.

Checked in the browser on *Cat City*: the box's computed `container-name` is `debugValidation`, and the table's computed `display` is `table` at a 584px module and `block` at 220px with the header row hidden — both from the stylesheet, so the narrow rule is live rather than inert.

### 3. `json.js:98` — fixed

`!problem.property?.length`, so it agrees with the defensive read at the renderer.

### Verification

`npm test` 1679/1679, `npm run browsercompat` green, `npx testcafe chromium:headless tests/testcafe/compute-4.js` 2 passing, library sweep 77 findings in 29 of main's 578 files with the same per-property breakdown (and the 4 removed *Interrogation* false positives).

VTT-AI-Agent:
Second fix folded into this PR: the game-file upload preview crashed with `Cannot read properties of undefined (reading 'info')` on any archive without VTT metadata — a Tabletop Simulator export, or a root JSON with no `_meta`. It was reported from the live server and I first offered to split it out; the maintainers decided against a separate PR, so it lands here as `client/js/overlays/states.js` + `tests/client/state-upload.test.js`. Details are in the *Uploading a file without VTT metadata* section of the description.

PR #3031, opened 2026-07-17
Add chat feature

VTT-AI-Agent:
## What

Adds a standalone **Chat** panel to the room — a floating, dockable/resizable panel toggled from its own toolbar tab, so players can talk to each other while playing.

- **Chat tab**: a new toolbar button placed between *Players* and *About* (swapping the previous About/Chat position). Unlike the other tabs it doesn't take over the main overlay area — it's a floating panel that can stay open alongside whatever else you're doing, and can be dragged/resized.
- **Per-client settings**: theme (light/dark), notification sound (ding/none), font size, and an auto-scroll toggle (only jump to the newest message when already at the bottom).
- **Unread counter**: messages from other players that arrive while the panel is closed show a red badge on the *Chat* tab; opening the panel clears it.
- **No message history**: chat is intentionally ephemeral — the server only broadcasts each message to players currently connected, it is never stored. Reloading or joining after messages were sent means you simply see the conversation from that point on, not the backlog.

Chat panel

Toolbar with the new Chat tab

## How

- `server/room.mjs` + `server/player.mjs`: a `chat` message handler that validates/trims the text (1000 chars max) and broadcasts `{ player, message, time }` — no server-side storage.
- `client/js/overlays/players.js`: the floating panel's drag/resize/settings logic, rendering incoming messages (via `textContent`, so no HTML injection) and the unread badge.
- `client/room.html` + `client/css/overlays/players.css`: the chat panel markup and styling.
- `client/css/layout.css`: updated the toolbar's height/width break-points across all three toolbar layouts (vertical, wide sidebar, horizontal) so the extra tab never causes the toolbar to overflow.

## Compatibility notes

- No widget property, routine or state format changes — no validator, jsonedit, legacy-mode or FileUpdater updates needed.
- Chat is not part of room state, so saved games, `/state`, exported files and the TestCafe state hashes are unaffected.
- This is now fully independent of #3013's Players-tab rework — no shared markup, state, or code path — per feedback below.
- Verified locally: `npm test` (jest), and Playwright sessions covering multi-client message delivery, the unread badge, and confirming no history survives a reload; a viewport sweep across all three toolbar layouts confirmed the toolbar never overflows.

## Feedback addressed

> - Chat toolbar button needs updated break-points so the toolbar never overflows.
> - Swap position of About and Chat.
> - No need to preserve chat history. It is okay for players to just see what has been posted while they are in the room.
> - Remove any connection with this and the revised player tab in #3013. This should stand alone.

(Source: Discord message from "lawdawg96" — https://discord.com/channels/770758631146782780/1527722679988260885/1530359882996387970)


VTT-AI-Agent (commented):
## TL;DR

Solid, well-scoped feature: the server handler validates input, the client renders via `textContent` (no injection), and the chat log deliberately stays out of room state so saves, `/state` and the TestCafe hashes are untouched. I verified it live in a two-browser session (send/receive, unread badge, history on reload, whitespace rejection, HTML escaping — all work) and jest passes. Only minor findings below; the most notable is that new messages force-scroll the log to the bottom even while you're scrolled up reading older ones.

## Findings

1. **Auto-scroll steals the reading position** — `client/js/overlays/players.js:299` (`$('#chatMessages').scrollTop = $('#chatMessages').scrollHeight` in `addChatMessage`). If the Players tab is open and a user has scrolled up to reread older messages, every incoming message yanks the log back to the bottom. The usual fix is to only auto-scroll when the log is already at (or near) the bottom, e.g. check `scrollHeight - scrollTop - clientHeight < 5` before appending.

2. **Enter sends mid-composition for IME users** — `client/js/overlays/players.js:322`. The `keydown` handler fires on the Enter that confirms a CJK IME composition, sending the half-typed message. Guard with `if(e.key == 'Enter' && !e.isComposing)`.

3. **No rate limiting on `chat`** — `server/room.mjs:280`. Any session can broadcast unlimited 1000-char messages to every player in the room. This matches the existing unthrottled `audio`/`mouse` handlers, so it's not a new class of exposure — just noting it as cheap hardening (e.g. a per-player minimum interval) if room spam ever becomes a problem.

4. **Renames leave the log behind (cosmetic)** — chat entries store the name at send time, so after a player renames, their old messages show the old name and lose the color lookup in `_meta.players`. Keeping the historical name is arguably correct chat behavior, so this is fine to leave as-is; just flagging it as intentional-looking behavior worth a maintainer's nod.

Verified locally on top of `players-tab-table`: `npm test` (15/15) and a scripted two-Chromium-context session against a local server — fresh-load unread badge (works from first connect because `connection.onopen` calls `showOverlay(null, true)`, which sets the inline `display:none` the badge check relies on), badge cleared on open, no badge while the tab is open or for history entries, `<b>` payload rendered as text, whitespace-only input rejected, and history intact after reload.

## Beta

None of the changed files touch the engine (`client/js/widgets/`, `client/js/compute.js`, `client/js/serverstate.js`) — neither in this PR's diff nor in the full delta vs `main` through its base branch `players-tab-table` (#3013, which already carries the `beta` label). The server changes are additive handlers inside the existing try/catch, and the feature can't touch room state.

I added the `beta` label: the PR is engine-neutral, verified working end-to-end, and safe to test — it cannot corrupt games or crash the server.

VTT-AI-Agent:
Thanks for the review! Addressed in ef213de4:

1. **Auto-scroll** — fixed: `addChatMessage` now only scrolls down when the log was already at the bottom (`scrollHeight - scrollTop - clientHeight < 5`), so reading older messages is no longer interrupted. Opening the tab still jumps to the newest message. Verified in a two-browser session: position preserved while scrolled up, still follows new messages at the bottom.
2. **IME composition** — fixed: the Enter handler now guards with `!e.isComposing`. (The name inputs from #3013 have the same pattern; left untouched here to keep this PR to the chat.)
3. **Rate limiting** — left as-is, matching the existing unthrottled `audio`/`mouse`/`chat`-class handlers; silently dropping throttled messages would be worse UX than the spam risk in a private room. Happy to add a per-player interval if room spam becomes a real problem.
4. **Renames** — intentional: chat entries keep the name the sender had when the message was sent, like a regular chat log. When the name no longer resolves to a color in `_meta.players`, the name just falls back to the default text color.

jest still passes (15/15).

VTT-AI-Agent (commented):
## TL;DR

Solid, well-scoped feature: the implementation is clean, XSS-safe (`textContent` everywhere, attribution comes from `player.name` server-side so it can't be spoofed beyond a normal rename), and I verified it works end-to-end locally (jest 15/15; a two-browser Playwright session confirmed sending, unread badge increment/clear, whitespace rejection, HTML escaping, and history delivery to a late joiner without inflating the badge). Only minor findings below — nothing blocking beta testing.

## Findings (all minor)

1. **Client-side log is unbounded** — `addChatMessage` (`client/js/overlays/players.js:288`) appends forever, while the server caps `chatLog` at 100 (`server/room.mjs:288`). Two consequences: DOM nodes accumulate without limit in a long chat-heavy session, and a reload shows a *different* (shorter) history than what the user saw live, which can be confusing. Trimming the oldest `.chatMessage` nodes past 100 in `addChatMessage` would fix both cheaply.

2. **No rate limiting on `chat`** (`server/room.mjs:280`) — a client can flood broadcasts to every player in the room. Existing handlers like `mouse`/`audio` aren't limited either, so this is consistent with repo conventions; just noting it because chat, unlike mouse moves, also persists in the history replayed to every joiner. Fine to leave as a follow-up if spam ever becomes a problem.

3. **Missed-while-disconnected messages never show as unread** — after a reconnect they arrive via `chatHistory`, which passes `isHistory = true`, so the badge stays clear even though the user hasn't seen them (`client/js/overlays/players.js:302-306`). Edge case; probably acceptable.

4. **Rename interaction is cosmetic-only but worth knowing**: `chatLog` stores the name at send time, so after a player renames, their old messages keep the old name and the `lastMetaArgs.meta.players[entry.player]` color lookup fails for them (they render uncolored for late joiners). Not worth complexity to fix now.

5. **Local players can't speak as themselves** — since attribution is `player.name`, a hotseat session (the focus of base PR #3013) always chats as the session's main player. A one-line mention in the PR description or a future sender-picker for local players might be worth considering.

Things I explicitly checked and found correct: the `wasAtBottom` scroll preservation and the `e.isComposing` IME guard (both from the earlier review round) work as intended; the badge's `style.display == 'none'` check is reliable because `showOverlay()` runs on load and hides all overlays; script concatenation order (`server/minify.mjs`: players.js before main.js) means the badge-clearing click handler is registered before main.js's `stopImmediatePropagation` paths, so it always fires; malformed `chat` messages are caught by the existing try/catch in `player.mjs`; and `splice(0, length - 100)` is a no-op below the cap.

---

The `beta` label is already present and I'm leaving it: no engine files are touched (nothing in `client/js/widgets/`, `compute.js`, or `serverstate.js` — the changes are UI overlay code plus an additive server message handler that doesn't affect state sync), and my local verification showed it safe and useful for players to test.

VTT-AI-Agent:
Thanks — addressed in 2eb8c971:

1. **Unbounded client log** — fixed: `addChatMessage` now trims the oldest entries past 100, mirroring the server cap, so the live log always equals what a reload shows. Verified with a 105-message session (DOM stays at 100, oldest is "msg 6" both live and after reload).
2. **Rate limiting** — left as-is per the previous round: consistent with the other unthrottled handlers, and silently dropping messages would be worse UX; happy to add a per-player interval if spam becomes a real problem.
3. **Missed-while-disconnected messages** — fixed: the client remembers the timestamp of the last rendered message, and after a reconnect the `chatHistory` entries newer than it count as unread (fresh page loads still don't badge history). Verified by severing a client's TCP connection through a local proxy: two messages sent while disconnected showed as a "2" badge after the automatic reconnect and cleared on opening the tab.
4. **Renames** — intentional historical-name behavior, now called out in the PR description under *Scope / limitations*.
5. **Local players chat as the hosting player** — correct; also documented in the PR description now. A sender picker for local players would follow #3013's view/split model and feels like follow-up material if hotseat users ask for it.

jest still passes (15/15).

VTT-AI-Agent (commented):
## TL;DR

The chat feature is safe for beta testing, but reconnecting the only browser in a room can preserve a chat log that the server has already discarded. The server needs to send an authoritative empty history (and the client should reset its related tracking state) when a recreated room has no messages.

## Finding

1. **An in-place reconnect can retain chat from an unloaded room** — `server/room.mjs:49`. `addPlayer()` only sends `chatHistory` when `chatLog.length` is nonzero. However, `removePlayer()` unloads the room immediately when its last socket disconnects (`server/room.mjs:879-891`), while `connection.js` reconnects the same page without reloading it. If a sole player suffers a transient disconnect, the newly created `Room` has an empty `chatLog`, so it sends no history and the existing `#chatMessages`, `lastChatTime`, and unread count remain in the DOM. Messages subsequently sent in the recreated room are appended alongside those stale entries, even though a reload or a late joiner only sees the new room's messages. Please always send `chatHistory`, including `[]`, and treat an empty authoritative history as a reset of the client log and its timestamp/unread bookkeeping; a reconnect-after-last-player-disconnect regression test would cover the room-lifecycle case.

## Beta

The `beta` label was already present; I left it in place because no engine files are touched and this UI/side-channel lifecycle bug cannot corrupt game state or crash the server.

VTT-AI-Agent:
Good catch — fixed in 06206d43 exactly as suggested:

- `addPlayer` now always sends `chatHistory` (including `[]`), making it authoritative.
- The client treats an empty history as a reset: it clears the log and the timestamp/unread bookkeeping before rendering.

Covered with a regression test of the room-lifecycle case: a sole player (connected through a severable TCP proxy) chats, loses the connection, the empty room is destroyed immediately, and after the in-place reconnect the stale log is gone ("No messages yet."), the badge stays clear, and a new message stands alone in the recreated room. The multi-player reconnect case (missed messages badge as "2" and the log intact) still passes, as does jest (15/15).

VTT-AI-Agent (commented):
## TL;DR

The room-lifecycle fix from the previous round (06206d43) is correct and complete: `addPlayer` now always sends `chatHistory`, and the client treats an empty history as an authoritative reset of the log, timestamp and unread count. I re-read the full diff and independently re-verified the feature end-to-end (jest 15/15 plus a scripted three-browser session); everything works as described. Nothing left that I'd ask to change.

## Verification

Independent smoke test against a local server on the current branch head: sending via Enter and the Send button, the unread badge appearing for a closed-overlay recipient and clearing on open, no badge while the tab is open, `<b>` payload rendered as escaped text, sender name colored from `_meta.players`, whitespace-only input rejected server-side, leading/trailing whitespace trimmed, and a late joiner receiving the history without inflating their badge — all behaved as expected. The empty-history reset path in `client/js/overlays/players.js:321-325` and the unconditional `player.send('chatHistory', ...)` in `server/room.mjs:50` match what the previous review asked for.

## FYI only — not worth a code change

In the recreated-room scenario, a non-empty history intentionally does not reset `unreadChatMessages` (`client/js/overlays/players.js:316-328`). If a client disconnects holding a nonzero unread count, the room is destroyed, and *another* player reconnects first and chats before this client's auto-reconnect, the stale count from the destroyed room is added to the new room's unread badge until the tab is opened. This needs a full room teardown racing two reconnects and only overstates a badge number, so I'd leave it as is — just documenting that the behavior is known.

The open To-Do items in the description (input placement, message order, renaming old entries) are design questions for a maintainer, not correctness issues.

---

The `beta` label was already present and I'm leaving it: no engine files are touched (nothing under `client/js/widgets/`, `client/js/compute.js` or `client/js/serverstate.js` — only overlay UI plus an additive server message handler outside room state), and my verification shows the feature safe and useful for players to test.

96LawDawg:
@VTT-AI-Agent, move the chat feature out of the players tab. Make a new tab for it at the bottom of the current menu tabs. Also make it dockable, expandable, and movable. So the chat feature by default is in a new tab. But if the user drags it, it pops out and can go anywhere in the room. Allow the user to resize it with expander handles. And allow the user to drop it back onto the menu area to dock back into the small configuration that shows the number of unread chats.

Also, rename old chat entries with player rename. Some people may initially post as Guest123 and then change their name in the room and that should stay with the same actual person, not the current player name.

VTT-AI-Agent:
@96LawDawg, both requests are addressed in dcad4adf:

> @VTT-AI-Agent, move the chat feature out of the players tab. Make a new tab for it at the bottom of the current menu tabs. Also make it dockable, expandable, and movable. So the chat feature by default is in a new tab. But if the user drags it, it pops out and can go anywhere in the room. Allow the user to resize it with expander handles. And allow the user to drop it back onto the menu area to dock back into the small configuration that shows the number of unread chats.
>
> Also, rename old chat entries with player rename. Some people may initially post as Guest123 and then change their name in the room and that should stay with the same actual person, not the current player name.

**Chat as its own dockable/movable/resizable panel**
- New **Chat** tab on the toolbar (placed after *About*, before *Edit Mode* — the toolbar's tab row doesn't have a dedicated trailing slot below Edit Mode, so this is the closest "after the existing tabs" spot). It's independent of the mutually-exclusive overlay tabs, so it can stay open alongside Players/States/etc.
- By default it opens as a floating panel; drag the header to move it anywhere in the room, drag the edge/corner handles to resize it.
- Drop the header back onto the toolbar and it re-docks: the panel closes and any custom position/size resets, leaving just the tab with the unread-count badge (moved from the Players button to the new Chat button).
- Position, size and open/closed state persist in `localStorage` across reloads.
- Chat markup/logic moved out of the Players overlay into a standalone `#chatPanel`; the Players tab's instructional text no longer mentions chatting there.

**Chat history follows player renames**
- A full rename (not a session split/view, which intentionally keeps the old name available for the other sessions) now relabels that player's existing `chatLog` entries server-side and broadcasts a `chatRename` event.
- Clients update the sender name and color on already-rendered chat messages, so a `Guest123`-style name that later becomes a real name stays attached to the same person's history instead of orphaning old messages under the stale guest name.

Verified: jest (15/15), and a scripted two-browser Playwright session covering: default-hidden panel, opening via the tab, unread badge accrual/clearing, drag-to-move, resize via handles, geometry/open-state surviving a reload, drag-to-toolbar re-docking, and a live rename relabeling the sender's earlier message (and its color) on another client's chat log.

Screenshots:
- https://agent.virtualtabletop.io/reports/pr/1527722679988260885/chat-docked-badge.png (docked tab with unread badge)
- https://agent.virtualtabletop.io/reports/pr/1527722679988260885/chat-floating-resized.png (floating, moved and resized panel)

One follow-up worth flagging: the new Chat tab icon currently falls back to the generic Material Symbols "chat" glyph (`#chatButton::before`) rather than a matching icon in the project's own VTT-Symbols font, since that font doesn't have a chat/bubble glyph yet — happy to add one if there's an appetite for extending the icon font.

96LawDawg:
@VTT-AI-Agent, eliminate the drop to dock feature. It is impossible to leave the chat window on the left side of the screen. Just leave the X to close it.

VTT-AI-Agent:
> @VTT-AI-Agent, eliminate the drop to dock feature. It is impossible to leave the chat window on the left side of the screen. Just leave the X to close it.

Fixed in 7a9bfdcc — the toolbar-drop re-dock check is gone. Dragging the header now always just leaves the panel where you release it and saves that position, so it can sit anywhere on the left edge (or overlapping the toolbar) without snapping shut. The X button is the only way to close it now.

Verified with a scripted Playwright session: dragged the panel to `x: 0` (flush against the toolbar), confirmed it stayed open there, reloaded and confirmed the position persisted, then confirmed clicking X closes it.

Panel left open at the left edge

jest still 15/15.

VTT-AI-Agent:
> Great. Now add the following features:
>
> - Change the color of the header/title to VTT dark blue.
> - Add a settings/gear icon near the X and setup: dark/light mode, notification sound (none or ding), font size (s/m/l), and auto-scroll only when at bottom).

(Source: Discord message from "lawdawg96" — https://discord.com/channels/770758631146782780/1527722679988260885/1530342711708614677)

Addressed in 28d31438:

- **Header color**: `#chatHeader` now uses `var(--VTTblueDark)` instead of the lighter `var(--VTTblue)`.
- **Settings gear**: a new gear button sits next to the X in the header and toggles a small settings block with:
  - **Theme**: Light / Dark — restyles the whole panel (background, text, input) for low-light use.
  - **Notification sound**: None / Ding — plays a two-tone ding on new messages from other players while the panel is closed (same trigger as the existing unread badge).
  - **Font size**: Small / Medium / Large for the message log.
  - **Auto-scroll only when at bottom** checkbox (on by default, matching current behavior) — turning it off always jumps to the newest message even if you're scrolled up reading history.

All four settings persist in `localStorage` independently of the panel's position/size, so they survive reloads.

Verified with a scripted Playwright session: header background renders as `#0d2f5e`, the gear opens/closes the settings block, dark theme applies (`data-theme="dark"` + restyled background), large font size measures 1.1rem, the auto-scroll toggle flips and persists, and selecting "Ding" runs with no JS errors. All settings (theme, font, sound, auto-scroll) correctly restore after a page reload.

Dark theme with settings panel open

jest still 15/15.

VTT-AI-Agent:
> Move the unopened message icon to directly after the text "Chat" when it is a large screen size. When it is minimized icon and no text, it works perfectly as is. and when it is on the bottom, it works fine.

(Source: Discord message from "lawdawg96" — https://discord.com/channels/770758631146782780/1527722679988260885/1530345908720500937)

Fixed in bf07986e. The unread badge now renders right after the "Chat" label in the wide sidebar (`body.wideToolbar`), instead of pinned to the far right corner of the full-width button. The icon-only and bottom-toolbar modes are untouched, since those were already correct.

Verified with a scripted Playwright session at a wide viewport (1900×900, confirmed `wideToolbar` active): the corner badge is hidden, the inline badge next to "Chat" shows the correct count, and it clears when the panel is opened. Also re-checked a narrow/tall viewport to confirm the bottom-toolbar mode still uses the original corner badge (regression check).

Chat label with the unread badge right after it

jest still 15/15.

96LawDawg (approved):
Reviewed.  Approved.

VTT-AI-Agent:
Reworked this PR based on feedback:

> - Chat toolbar button needs updated break-points so the toolbar never overflows.
> - Swap position of About and Chat.
> - No need to preserve chat history. It is okay for players to just see what has been posted while they are in the room.
> - Remove any connection with this and the revised player tab in #3013. This should stand alone.

(Source: Discord message from "lawdawg96" — https://discord.com/channels/770758631146782780/1527722679988260885/1530359882996387970)

Changes:
- **Base retargeted from `players-tab-table` (#3013) to `main`** — chat is now fully standalone, no shared code with the Players-tab rework.
- **Removed all chat history** — server no longer keeps a `chatLog` or sends `chatHistory`/`chatRename`; it only broadcasts each message live. A reload or a fresh join now shows an empty log, and you see only what gets posted while you're present.
- **Swapped tab order**: *Chat* now sits between *Players* and *About* (previously About came before Chat).
- **Updated toolbar break-points** in `client/css/layout.css` across all three toolbar layouts (vertical, wide sidebar, horizontal) to make room for the extra tab — verified with a Playwright viewport sweep that the toolbar never overflows.

VTT-AI-Agent (commented):
## TL;DR

Solid, self-contained feature that stays outside room state (no engine impact), renders all user text via `textContent` (no XSS), and passes CI (jest, TestCafe, CodeQL). The main gap is `Room.chat()` has no rate limiting/flood protection — a misbehaving or malicious client can spam `broadcast('chat', ...)` to every connected player with no server-side throttle.

*Manually requested review (code) — by `sonnet`*

## Findings

1. **No rate limiting on chat messages** (`server/room.mjs:268-274`, `server/player.mjs:31-32`). `Room.chat()` only trims and length-caps the message; a client can call `toServer('chat', {message})` in a tight loop and the server will `broadcast` every single one to all connected players with no per-player cooldown. This isn't a regression from existing conventions (mouse/audio messages aren't rate-limited either), but chat text is far more disruptive to spam than a cursor position — worth a simple per-player minimum interval or a short-window message cap before this reaches production, even if it's fine to test as-is on beta.

2. **Client-side 200-message cap only bounds one client's view, not server load** (`client/js/overlays/players.js:620-623`). Reasonable as a memory bound for the DOM, just noting it doesn't help with #1 since the cap is purely local.

3. Minor: `chatSettings`/`chatPanelState` are stored in `localStorage` without any room/game scoping (`client/js/overlays/players.js:390,399`), so panel position/size and settings are shared across every room a browser visits. That's likely intentional as a global UI preference (consistent with how other per-client toolbar prefs are stored), but worth confirming it's the intended behavior rather than an oversight, since a panel resized/moved for a tiny mobile layout in one room will carry into a very different layout in the next.

Everything else checks out: message rendering uses `textContent` throughout (no innerHTML with user data), player identity for broadcast messages comes from the authenticated server-side `player.name` (not client-supplied, so no impersonation), the `onMessage('meta', ...)` handler correctly coexists with other listeners (the dispatcher supports multiple callbacks per message type), and the toolbar breakpoint changes in `client/css/layout.css` are consistent with the existing per-button hide-at-width/height pattern. The concatenation build (`server/minify.mjs`) confirms `players.js` loads before `main.js` in the single bundled module script, so `main.js`'s bare call to `toggleChatPanel()` (no import) resolves fine — not a bug.

## Beta label

The `beta` label was already present and CI is green; none of the changed files (`client/js/overlays/players.js`, `client/js/main.js`, `client/css/*`, `client/room.html`, `server/player.mjs`, `server/room.mjs`) touch engine state-sync, widget, or routine semantics, so it remains eligible for beta testing — I left the label as-is per the instructions (already present, not removed).

RaphaelAlvez (approved):
As I said on discord I'm not sure if this is actually a good idea to have a unmoderated chat.
It does work and look good

PR #3032, opened 2026-07-17
Add a public rooms section: publish claimed rooms with a description

VTT-AI-Agent:
## Goal

Builds on #3018 (targets its `rooms-collection` branch): any claimed (owned) room can be **made public** with a description. It is then listed in a new **Public rooms** section of the Rooms tab, where anybody can see it and join.

### Publishing (room admins)
- The tile menu of a claimed room gets **Make room public**: the admin enters a description (required, trimmed, capped at 500 characters server-side) and the room appears in the public list. **Edit public listing** updates the description, **Make room private** removes the listing again.
- The listing is stored as `_meta.public = { description }` on the room plus an entry in a server-side registry (`save/publicRooms.json`, new `server/publicrooms.mjs`, capped at 500 rooms).
- The listing is automatically removed when the claim is released, when the room is deleted, and it is not carried over by **Copy room** (a copy starts private). The listing endpoint also lazily heals the registry when a listed room no longer exists or is no longer public, and skips (with a log entry) rooms that fail to load instead of failing the whole listing.

### Public rooms section (everyone)
- New **Public rooms** section in the Rooms overlay below the user's own rooms, fed by the new anonymous `GET /api/publicrooms` endpoint. Tiles reuse the existing room-tile design and additionally show the **description** and the **number of players currently online**; the list is sorted by player count.
- Anybody can join a public room with the tile's existing switch button; visiting it adds it to their own collection as usual. Public tiles offer copy/link in the menu, but not "Remove from list" (they are not part of the viewer's collection); admins of a public room still get their management menu there — the client flags admin tiles by matching room IDs against its own collection listing, so no secret is sent to the public endpoint.
- The user's own tiles show a "public" globe icon when the room is published.

### Security & abuse considerations
- `setPublic` is a new action of the existing `POST /api/room/:room/:action` endpoint and requires admin (claimed collection), like rename/lock/password.
- `isPublic()` only reports rooms that are still claimed, so stray `_meta.public` data without a claim never gets listed.
- The anonymous listing is **viewer-independent**: it takes no collection parameter (no scrypt work, no secret in query strings/access logs) and **strips connected players' names**, exposing only a `playerCount` — players didn't opt into being listed publicly. Own-collection tiles still show the name chips from #3018's base branch.
- The response is cached server-side for 10 seconds (invalidated by room management actions, so publish/unpublish show up immediately), which keeps the endpoint cheap to poll and bounds the disk churn of loading idle rooms.
- Password-protected rooms can be published; the key icon shows that joining needs the password.

## Testing
- `npm test` (jest) passes (25/25); added a **Public rooms** describe block to `tests/server/security.test.js`: registry add/remove, publish requires admin and a non-empty description, description trimming/capping, and unpublish/unclaim/delete all remove the listing.
- Manual end-to-end verification with Playwright against a local server: claim → publish with description → globe icon appears; a second, fresh browser profile sees the room in the public section with description and player count (and no admin menu entries), joins it via the switch button and lands in the room; unpublishing removes the tile immediately; the raw anonymous API response contains no player names; the admin sees their management menu on their own public tile.
- Screenshots: [admin view](https://agent.virtualtabletop.io/reports/pr/1527730406948868156/screenshot-admin.png) (own tile with player chip + globe/admin icons, public tile with count and description), [guest view](https://agent.virtualtabletop.io/reports/pr/1527730406948868156/screenshot-guest.png) (public tile with description, player count and non-admin menu).
- Like #3018, the TestCafe suite could not be run meaningfully in the sandbox; relying on CI.

### DIFF
--- a/client/css/overlays/rooms.css
+++ b/client/css/overlays/rooms.css
@@ -4,11 +4,11 @@
 
 /* flex items shrink to fit the overlay by default, which squeezes/hides
    these sections instead of letting the overlay scroll to show them */
-#roomsOverlay > .heading, #roomsOverlay > #roomsList, #roomsOverlay > #roomsCollectionBox {
+#roomsOverlay > .heading, #roomsOverlay > #roomsList, #roomsOverlay > #publicRoomsSection, #roomsOverlay > #roomsCollectionBox {
   flex-shrink: 0;
 }
 
-#roomsList {
+#roomsList, #publicRoomsList {
   display: flex;
   flex-wrap: wrap;
   align-content: flex-start;
@@ -16,10 +16,14 @@
   padding: 16px 0;
 }
 
-#emptyRoomsList {
+#emptyRoomsList, #emptyPublicRoomsList {
   display: none;
   font-style: italic;
   color: black;
+}
+
+#publicRoomsSection h2 {
+  margin: 10px 0 5px 0;
 }
 
 .roomTile {
@@ -76,6 +80,20 @@
   padding: 3px 8px 4px 8px;
   margin: 0;
   background: #00000061;
+}
+
+.roomTile .details .description {
+  font-size: .75rem;
+  max-height: 3.4em;
+  overflow: hidden;
+  padding: 3px 8px 4px 8px;
+  margin: 0;
+  background: #00000061;
+  overflow-wrap: anywhere;
+}
+
+.roomTile .details .description.hidden {
+  display: none;
 }
 
 .roomTile .roomIcons {
--- a/client/js/overlays/rooms.js
+++ b/client/js/overlays/rooms.js
@@ -83,7 +83,7 @@ function switchRoom(newRoomID, pushHistory=true) {
   toServer('room', { playerName, roomID: newRoomID, collection: getCollectionID(), password: getRoomPassword(newRoomID) });
 }
 
-function createRoomTile(room) {
+function createRoomTile(room, isPublicTile) {
   const tile = domByTemplate('template-roomslist-entry');
   tile.className = 'roomTile';
   if(!room.image)
@@ -93,10 +93,18 @@ function createRoomTile(room) {
 
   $('h3', tile).textContent = room.name;
   $('h4', tile).textContent = room.gameName || 'No game loaded';
+  if(isPublicTile) {
+    $('h4', tile).textContent += ` — ${room.players} ${room.players == 1 ? 'player' : 'players'} online`;
+    if(room.description) {
+      $('.description', tile).textContent = room.description;
+      $('.description', tile).classList.remove('hidden');
+    }
+  }
   if(room.image)
     $('img', tile).src = mapAssetURLs(room.image);
   toggleClass($('.lockedIcon', tile), 'hidden', !room.locked);
   toggleClass($('.passwordIcon', tile), 'hidden', !room.hasPassword);
+  toggleClass($('.publicIcon', tile), 'hidden', !room.isPublic || !!isPublicTile);
   toggleClass($('.adminIcon', tile), 'hidden', !room.isAdmin);
 
   const menu = $('.roomMenu', tile);
@@ -137,6 +145,15 @@ function createRoomTile(room) {
       if(password !== null)
         roomAction(room.id, 'setPassword', { password });
     });
+    addMenuButton('public', room.isPublic ? 'Edit public listing' : 'Make room public', null, function() {
+      const description = prompt('Enter a description for the public rooms list so people know what to expect in your room:', room.description || '');
+      if(description !== null)
+        roomAction(room.id, 'setPublic', { public: true, description });
+    });
+    if(room.isPublic)
+      addMenuButton('public_off', 'Make room private', null, function() {
+        roomAction(room.id, 'setPublic', { public: false });
+      });
     addMenuButton('remove_moderator', 'Release claim', null, function() {
       roomAction(room.id, 'unclaim');
     });
@@ -156,14 +173,15 @@ function createRoomTile(room) {
         showOverlay('roomsOverlay');
       }
     });
-  addMenuButton('visibility_off', 'Remove from list', null, async function() {
-    await fetch(`api/roomcollection/${getCollectionID()}/remove/${room.id}`, { method: 'PUT' });
-    await refreshRoomsList();
-  });
+  if(!isPublicTile)
+    addMenuButton('visibility_off', 'Remove from list', null, async function() {
+      await fetch(`api/roomcollection/${getCollectionID()}/remove/${room.id}`, { method: 'PUT' });
+      await refreshRoomsList();
+    });
 
   $('.menuButton', tile).onclick = function(e) {
     const wasHidden = menu.classList.contains('hidden');
-    for(const m of $a('#roomsList .roomMenu'))
+    for(const m of $a('#roomsOverlay .roomMenu'))
       m.classList.add('hidden');
     toggleClass(menu, 'hidden', !wasHidden);
     e.stopPropagation();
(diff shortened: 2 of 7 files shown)


VTT-AI-Agent (commented):
## TL;DR

Solid, well-tested feature that follows #3018's patterns closely (admin gating, lazy healing, `_meta` cleanup on unclaim/delete/copy are all handled correctly, jest passes 24/24 on the head). The two things to address: the PR now **conflicts with its base branch** because `rooms-collection` changed `players` from a count to an array of player objects — which also breaks this PR's player-count display/sort and quietly turns the public listing into a broadcast of connected players' names — and the unauthenticated `GET /api/publicrooms` endpoint is a **resource-amplification vector** (disk churn + scrypt per room with attacker-chosen input).

## Findings

### 1. Conflict with base branch changes the meaning of `players` (needs rework, not just conflict resolution)

The PR is marked CONFLICTING against `rooms-collection`: the base gained 0d691351 ("Show active players on each room tile"), which changed `getRoomDetails().players` from `this.players.length` to an array of `{ name, color }` objects. Beyond the textual conflicts in `rooms.css` and `server/room.mjs`, this semantically breaks two spots in this PR that treat `room.players` as a number:

- `client/js/overlays/rooms.js:97` — `` `${room.players} ${room.players == 1 ? 'player' : 'players'} online` `` would render the comma-joined player names ("Alice,Bob players online");
- `client/js/overlays/rooms.js:233` — `publicRooms.sort((a, b)=>b.players - a.players)` becomes `NaN - NaN`, so the by-player-count sort silently stops working.

Both need `.length` (or a separate `playerCount` field) after the rebase.

**Privacy side effect to decide deliberately:** once rebased, `/api/publicrooms` (no secret required) would expose the *names* of everyone currently connected to a public room to the entire internet. The admin opted into publishing the room; the players didn't opt into having their names listed publicly. Consider returning only a count (or anonymized chips) in the public listing.

### 2. `GET /api/publicrooms` is an unauthenticated resource-amplification vector (`server.mjs:383`)

Two compounding costs, both triggerable by anyone without a secret:

- **Disk churn:** every request calls `ensureRoomIsLoaded()` for each listed room that isn't active — a full file read, `JSON.parse`, FileUpdater migration and linked-state sync per room (room saves can be large). Each of those rooms then unloads 5 seconds later (`server/room.mjs:26`), and since public rooms are always claimed, `unload()` unconditionally rewrites the save file (`server/room.mjs:1432`). So a single anonymous client polling every ~6 seconds sustains a read+write cycle of up to 500 room files — and ordinary users opening the Rooms overlay contribute the same load. The collection listing from #3018 has the same shape but is gated by a secret and bounded by one collection's size; this endpoint iterates a global list.
- **scrypt amplification:** the client passes `?collection=<secret>` (`client/js/overlays/rooms.js:212`) so `getRoomDetails(collection)` → `isAdmin()` runs scrypt per room. The memo in `hashSecret` is keyed by salt+secret and capped at 100 entries *per room*, so a request with a fresh random `collection` value forces one scrypt derivation per public room per request — a cheap way to saturate the libuv thread pool. (Note the existing `/api/roomcollection/:collection` doesn't have this problem: a made-up collection there has no rooms to iterate.)

Suggested direction: drop the `collection` parameter entirely — the client already fetches its own collection and can flag admin tiles in the public section by matching room IDs against its own rooms' `isAdmin` — then the response becomes viewer-independent and can be cached server-side for a short TTL (which also fixes the disk churn). Alternatively, store name/description in the registry so idle rooms never need loading, and compute live player counts only for rooms already in `activeRooms`. This should be addressed before this graduates beyond beta; sending the collection secret in a GET query string (ends up in access logs) is a further small reason to remove the parameter.

### 3. One corrupt room file 500s the whole public listing

`Room.load()` parses existing save files without a try/catch (`server/room.mjs:807`). Inside the `/api/publicrooms` loop, a room file with broken JSON throws, hits `.catch(next)`, and the entire endpoint returns 500 — every user's public section shows "Could not load the public rooms" until that one file is fixed, and the healing for rooms later in the list never runs. Wrap the per-room work in try/catch and skip (or heal out) rooms that fail to load.

### Minor

- Publishing with an empty description is accepted (`server/room.mjs:304` — `String(args.description || '').trim()` can be `''`). Probably fine, but if the description is the point of the listing, consider rejecting empty ones.
- Public descriptions are unmoderated user-generated content shown to every visitor (XSS is handled — the client uses `textContent`); there's no report/abuse path yet, which is worth a thought before this runs on the main server.

---

**Beta:** none of the changed files touch the engine (no `client/js/widgets/`, `client/js/compute.js`, or `client/js/serverstate.js`; client changes are confined to the Rooms overlay UI and server changes to room management), the head branch deploys standalone (it contains #3018, which is already on beta), and the abuse findings above are adversarial-load concerns rather than crashes or corruption in normal play — so I added the `beta` label as safe and useful for player testing.

VTT-AI-Agent:
Thanks for the review — all findings addressed in b6ead3d2 (merge) + d64f8643:

**1. Conflict / `players` semantics:** Merged `rooms-collection` (0d691351) into the branch; the PR is mergeable again. `getRoomDetails` now returns both the player chip array and a `playerCount`, and the public tiles use `playerCount` for the "N players online" text and the sort — no more names-as-string rendering or `NaN` sort. On the privacy point I went with your suggestion: the anonymous listing **strips the `players` array entirely** and only exposes the count, so connected players' names are never broadcast (verified via a raw `fetch('api/publicrooms')` in the e2e run: the response contains only `playerCount`). Own-collection tiles still show the name chips as before.

**2. Resource amplification:** Took the suggested direction. The `?collection=` parameter is gone — no scrypt runs on this endpoint at all anymore, and no secret ends up in access logs. The client instead flags admin tiles in the public section by matching room IDs against its own collection listing's `isAdmin` (edge case accepted: an admin who removed their own public room from their collection list sees it as a plain tile). Since the response is now viewer-independent, it's cached server-side for 10 seconds, which also bounds the disk read/unload-write churn: an anonymous poller now hits the cache instead of loading up to 500 rooms per request. The cache is invalidated on every `/api/room/:room/:action` call so publish/unpublish/rename/delete show up immediately (the e2e flow confirms tiles appear/disappear right after the action). If the remaining one-rebuild-per-10s churn is still a concern for the main server, persisting name/description in the registry so idle rooms never load would be the follow-up I'd propose.

**3. Corrupt room file 500s the listing:** Each room in the loop is now wrapped in try/catch — a room that fails to load is logged and skipped (not healed out, since corruption might be transient/fixable), and the rest of the listing still renders.

**Minor:** Publishing now rejects an empty/whitespace description with a clear error (`Please enter a description for the public rooms list.`) — new jest test covers it (25/25 pass). On moderation: agreed that's worth solving before this runs on the main server, but I'd scope a report/abuse path as a follow-up; the admin-only gate, the 500-char cap and the 500-room registry cap bound the exposure for beta.

Re-verified end-to-end with Playwright after the changes: claim → publish → guest sees description + count (no names, no admin menu) → joins → admin unpublishes → tile gone. Updated screenshots: [admin view](https://agent.virtualtabletop.io/reports/pr/1527730406948868156/screenshot-admin.png) (own tile with player chip, public tile with count only), [guest view](https://agent.virtualtabletop.io/reports/pr/1527730406948868156/screenshot-guest.png).

VTT-AI-Agent (commented):
## TL;DR

**This is a UI/UX review** (layout, readability, first-time-user clarity), not a code review. The new Public rooms section fits the existing rooms-overlay design well — it reuses the tile layout, has proper empty/error states, and the publish/unpublish menu wording is clear. The single most impactful improvement: **make the description actually readable** — right now it's clipped mid-line after ~3 lines with no ellipsis and no way to see the rest, so up to ~80% of a 500-character description is silently invisible.

I ran the branch locally and exercised the full flow (claim → publish → guest view → join) with Playwright:
[admin view](https://agent.virtualtabletop.io/reports/ui-review/1527741479756370110/shot-admin-rooms.png) · [guest view](https://agent.virtualtabletop.io/reports/ui-review/1527741479756370110/shot-guest-rooms.png) · [guest tile menu](https://agent.virtualtabletop.io/reports/ui-review/1527741479756370110/shot-guest-public-menu.png)

## Big layout improvements

1. **Description truncation loses most of the text with no affordance to read it.** `.roomTile .details .description` (`client/css/overlays/rooms.css`) uses `max-height: 3.4em; overflow: hidden`. With the default line-height that's ~2.9 lines, so the third line's descenders get shaved and a long description just stops mid-sentence ("…every evening around") with no ellipsis. Measured on a 500-char description: 217px of text clipped to 48px. Suggest `display: -webkit-box; -webkit-line-clamp: 3; -webkit-box-orient: vertical` for a clean ellipsized cut, plus setting `title` (or a tooltip) to the full description in `createRoomTile()` so viewers can read the rest — the description is the whole point of the listing, and 500 chars are accepted but only ~90 are ever shown.

2. **The user's own rooms list is now unlabeled while "Public rooms" gets a heading.** The overlay reads: h1 "Rooms" + collection blurb → unlabeled tile grid → h2 "Public rooms" → h2 "Collection ID". A first-time user scanning the page can't tell where "my rooms" ends and could plausibly think the h1's "Every room you visit…" text describes the public list too. Adding a parallel h2 (e.g. **"Your rooms"**) above the personal grid — matching the `#publicRoomsSection h2` styling you already added — would make the two-section structure scannable and also clarify the duplicate tiles (a room you admin appears in both sections, which is fine, but only once the sections are visibly distinct).

3. **Player count is squeezed into the game-name line.** `"No game loaded — 1 player online"` mashes two unrelated facts into one small italic line (`client/js/overlays/rooms.js:96`), and it's the sorting key of the whole list, so it deserves more prominence. The `.roomPlayers` chip row is empty on public tiles (names are stripped — good call) — consider rendering the count there as a single neutral chip ("3 players online"), reusing the existing chip styling. That keeps the game line clean, matches the visual language of own tiles (players appear as chips at the top), and makes the liveliness of a room visible at a glance instead of buried in fine print.

- **The new globe icon is nearly invisible on imageless tiles.** `.roomTile .roomIcons` renders white icons with a faint shadow; on the light-gray no-image tile the `publicIcon` is barely discernible (see the own tile in the admin screenshot). The other icons share this pre-existing problem, but the globe is the *only* indicator telling an admin "this room is currently published", so its legibility matters more. A darker text-shadow or a subtle dark backdrop (like the `#00000061` bands used in `.details`) would fix all four icons at once.
- The third description line is vertically cropped (descenders cut) because `3.4em` doesn't land on a line boundary — the line-clamp fix above resolves this too.
- The list re-sorts by live player count during the overlay's ~5s auto-refresh, so tiles can jump under the cursor while a user is reaching for a switch button. Consider sorting only when the overlay opens, or keeping order stable across refreshes while it stays open.
- The `prompt()` for the description is a single-line input for a field that allows 500 characters, and the limit isn't mentioned anywhere — overlong text is silently truncated server-side. Using `prompt()` matches the rename/password conventions, so at minimum mention the limit in the prompt text ("…up to 500 characters"); a textarea overlay could come later.

- The section blurb is good, but "**Claim a room** and use 'Make room public'…" leans on jargon a newcomer hasn't met yet (claiming is only explained inside the tile-menu confirm dialog). A softer phrasing like "If you're a room's admin (claim it via its tile menu first), use 'Make room public' to list it here" ties the concept to where the action lives.
- On a public tile the guest's menu contains only **"Copy room"** and **"Create linked room"** (screenshot 3). "Create linked room" is expert jargon in a context aimed at drop-in visitors, and it's worth double-checking that offering copy/link of *someone else's* public room to every anonymous visitor is intended; if it is, consider whether the link option belongs on public tiles at all. The big "Switch to this room" button being the primary action is clear, though.
- The menu labels "Make room public" / "Edit public listing" / "Make room private" and the icon tooltips are clear and self-explanatory — nice.
- Empty and error states exist for both lists ("No public rooms right now." / "Could not load the public rooms.") — good; first-time users won't hit a blank void.
- Minor: after publishing, the admin's own tile shows only the globe icon; the description is only visible on the duplicate tile in the public section below. That's acceptable, but it's a small hunt to proofread what you just wrote — the tooltip suggested in point 1 above (on the globe icon or tile) would cover this too.

Overall: solid, convention-respecting UI for a significant new feature — the findings above are refinements, with the description truncation being the one I'd fix before merge.

*— automated UI/UX review (rendered locally at the PR's head commit; screenshots above are from that run)*

ArnoldSmith86:
We're not doing this for now.

PR #3033, opened 2026-07-17
Make the UI translatable and add German and Brazilian Portuguese

VTT-AI-Agent:
## Goal

Make the entire UI translatable (not the games) and add German and Brazilian Portuguese as the first selectable languages — implemented so it doesn't interfere with the 100% statically served room.html, detecting the best-matched language from the browser's preferred list while keeping it user-selectable (in the players tab for now).

## How it works

- **room.html stays 100% static.** The served HTML is identical for every user and language. Translation happens entirely client-side: dictionaries are plain static JSON files under `client/i18n/<code>.json` (already covered by the existing `express.static('client')` route, so no server route changes and full HTTP caching).
- **New module `client/js/i18n.js`** (added to the main JS bundle):
  - Resolves the language: `localStorage.language` if set, otherwise the best match of `navigator.languages` against the available dictionaries (exact tag first, then primary-subtag match, so `de-AT` → `de` and `pt` → `pt-BR`). Falls back to English.
  - For non-English languages it fetches the dictionary and translates the DOM once on load: text nodes and a few attributes (`placeholder`, `title`, `aria-label`, `data-label`, `data-placeholder`) whose **English text is the dictionary key**. Untranslated strings simply stay English, so partial dictionaries degrade gracefully and adding a language is just adding one JSON file.
  - Game content is never touched: the walker skips `#room`, player cursors, game tiles (`.roomState`), variant rows, the JSON editor surfaces and its widget tree (`#jeTree`), anything marked `translate="no"`, and icon elements (`material-symbols`, `UIicon`) whose text is a font ligature name like `share` or `delete` that collides with real UI wording.
  - `<option>`s without a `value` attribute get their English value materialized before the label is translated, so all filter logic that compares against `'Any'` etc. keeps working.
  - Exports `translate()` / `getLanguage()`, used for the UI strings that are generated in JS. `translate` is also exposed to the on-demand editor bundle.
- **Language selector in the players tab**: a new *Language* section with a dropdown (English, Deutsch, Português (Brasil)). Changing it stores the choice in localStorage and reloads the page (room state is server-side, so nothing is lost).

## Editor coverage

The dictionaries now hold **~2600 entries each** (up from ~590), covering the editor as well as the room around it: the properties panel with every widget type, the deck editor, the routine and JSON editor surfaces, the widget library, the sidebar modules, the add-widget overlay, the compute-function reference and every editor toolbar tooltip. Walking the running editor in German and Brazilian Portuguese now leaves only brand names (Kenney, Game-icons.net, Tabletop Simulator …), CSS keywords, language tags and identifiers in English.

Two kinds of change were needed beyond the dictionaries:

- **The editor overlays that edit mode moves next to the room** (add widget, deck editor dialogs, library decks, the widget edit overlays) are now observed for changes individually. Previously only `#editor` was, so everything those overlays recreated at runtime stayed English. They are observed one by one rather than through `#roomArea`, so the room's widgets are not walked on every move.
- **Labels the editor builds from a template** — widget type and box, the snap grid presets, the lock states, the layer notes, the deck editor's panel titles and tree toolbar, the players table's connection column — went through string interpolation, which no dictionary key can match. They now use `translate()` with `{placeholder}`s, the same shape the rest of the generated UI text already used. English output is unchanged.

## Toolbar

The wide (200 px) toolbar puts the tab labels next to the icons at 20 px with `white-space: nowrap`, so a literal translation overflows the column: "Bearbeitungsmodus" stuck out 45 px, "Estante de Jogos" 20 px, "Modo de Edição" 7 px. Those entries now use short forms that fit — *Edit Mode* → "Bearbeiten" / "Editar", *Game Shelf* → "Spieleregal" / "Estante", *Hide Toolbar* → "Leiste ausblenden" / "Ocultar barra". Every label of both languages now ends inside the toolbar with room to spare (measured in a browser at 2400×900, where the wide layout applies).

## Notes

- English users see byte-identical behavior (no dictionary is fetched, the DOM is untouched) — this also means CI/TestCafe (en-US browsers) is unaffected.
- Widget ids in the selection bar are marked `translate="no"`, and the widget library's tooltip translates its name and description lines separately instead of after joining them.
- The `engine` label is only due to three one-line `translate()` wraps of the input-overlay default button texts in `client/js/widgets/widget.js` ("Go"/"Cancel"/"Working...") — no behavior change for games.
- Jest: 1489/1489 passing. TestCafe `editor.js` and `domsnapshot.js` run locally (one `widget-gallery` snapshot failure that reproduces as a timing flake and passes on re-run). Verified in a real browser (Chromium via Playwright) with `en-US`, `de-DE` and `pt-BR`, walking the room, all overlays, every sidebar module, the properties panel per widget type and the deck editor.

## Screenshots (German)

| Wide toolbar — every label fits | Properties panel |
|---|---|
| toolbar | properties |

| Deck editor | Players tab with the language selector |
|---|---|
| deckeditor | players |

| Game shelf | Add game overlay |
|---|---|
| shelf | addgame |

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3033/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

*(This is a UI/UX review — layout, readability, first-time-user clarity. Code correctness is covered by a separate review.)*

The translated UI holds up very well: I ran the branch and rendered the room, players tab, game shelf, add-game overlay and game details in `de-DE`, `en-US` and `pt-BR` — no overflow, clipping or awkward wrapping anywhere, including the longest German strings ("Werkzeugleiste ausblenden", "SPIEL HINZUFÜGEN", "Qualquer"). The single most impactful improvement: update `<html lang>` to the active language when translating — right now German/Portuguese users get a page that still declares itself English, which mispronounces everything for screen readers and makes Chrome offer to "translate" an already-translated page.

Verified in a real browser (Chromium/Playwright): [players tab (de)](https://agent.virtualtabletop.io/reports/ui-review/1527815699324604580/de-players.png) · [game shelf (de)](https://agent.virtualtabletop.io/reports/ui-review/1527815699324604580/de-shelf.png) · [add-game overlay (de)](https://agent.virtualtabletop.io/reports/ui-review/1527815699324604580/de-addgame.png) · [game shelf (pt-BR)](https://agent.virtualtabletop.io/reports/ui-review/1527815699324604580/pt-shelf.png) · [longest tooltip (de)](https://agent.virtualtabletop.io/reports/ui-review/1527815699324604580/de-tooltip-hide.png)

## Big layout / structural improvements

1. **Set `document.documentElement.lang` when translating.** `client/room.html:2` hardcodes `<html lang="en">` and `translateDOM()` in `client/js/i18n.js` never updates it. A one-liner (`document.documentElement.lang = language`) fixes screen-reader pronunciation, hyphenation, and stops browser auto-translate prompts on a page that's already in the user's language. This is the biggest win available for the effort.

2. **Sentence-fragment keys already produce visible grammar breakage — consider whole-sentence keys for link-embedded sentences.** Because sentences with inline links are split into fragments (`"Please note that PCIO files might not be 100% compatible. Please"` + `"open an issue on GitHub"` + `"if a file is unplayable."`), translators are locked into English word order and punctuation. This is visible today in the German add-game overlay ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1527815699324604580/de-addgame.png)): *"Bitte erstelle ein Issue auf GitHub wenn eine Datei nicht spielbar ist."* is missing the comma German requires before "wenn" — and the translation *can't* add it, because the fragment boundary sits exactly where the comma goes. The icon-picker sentence similarly ends up with a doubled "…angezeigt und … angezeigt". For the handful of sentences containing links, a whole-sentence key with a placeholder (you already use `{name}`/`{count}` elsewhere, e.g. `"Please note … Please {link} if a file is unplayable."`) would fix this and pay off for every future language. Fine as a follow-up, but worth doing before more dictionaries are contributed on the fragile pattern.

3. **The language switch reloads the page with no warning.** Nothing is lost (room state is server-side), but an unannounced full reload mid-game feels like a glitch. Suggest appending one sentence to the hint text: *"The page reloads to apply the change."* — cheap reassurance in exactly the place the user is looking.

The *placement* of the selector in the players tab reads fine to me for now: it's the closest thing the app has to a settings tab, auto-detection means most users never need it, and the section styling (centered heading + hint + control) matches the existing "Invite" section, so nothing looks bolted-on.

- **`#languageSelect` has no accessible name** (`client/room.html:109`). The `<h1>`/`<p>` above it aren't programmatically associated, so assistive tech announces an unlabeled combo box. Add `aria-label="Language"` (or wrap in a `<label>`) — and since `aria-label` is an attribute rather than a text node, note it won't be picked up by the current attribute-translation list unless you add it to `translatedAttributes` or set it from JS via `translate()`.
- **The select is browser-default styled** while everything around it is themed — the shelf filters use the borderless blue-underline style (`#stateFilters select`, `client/css/overlays/states.css:57`) and the adjacent "Share room URL" button is a styled blue button. The `font-size`/`padding` added in `players.css:136` helps, but a border/border-radius consistent with the app's inputs (or the filter-select underline style) would make it feel native to the UI. Minor.
- **Minor German wording, worth a native speaker's pass before merge:** the sort dropdown's "nach Name" would more correctly be "nach Namen" (dative), and "Skill: → Können:" reads oddly as a label — "Anspruch" or "Schwierigkeit" would be clearer to a German first-timer. Purely dictionary edits, no code change.

- **Auto-detection plus endonym labels ("English / Deutsch / Português (Brasil)") is exactly right** — a user who can't read the current language can still find their own in the list.
- **The hint "Game content is not translated" is a genuinely good touch**: it preempts the most likely first-time confusion (a German user opening a game and seeing English rules/labels, as in the game-details overlay where translated chrome sits next to English descriptions). The only caveat is that the explanation lives solely in the players tab, so a user who never opens it won't see it — acceptable trade-off at this stage.
- **No degradation for English users** — I verified the `en-US` players tab renders identically apart from the new Language section, and the selector correctly shows "English" as selected.
- Untranslated areas (edit mode, About attribution text) are consciously scoped out per the PR description; from a first-time *player's* perspective the translated surface covers everything they'd touch in a first session, which is the right prioritization.

Overall: solid, careful UI work — the three-language rendering is clean, and the findings above are all small, targeted fixes (the `lang` attribute being the one I'd consider a should-fix before merge).

VTT-AI-Agent (commented):
## TL;DR

Solid, well-scoped approach (static room.html preserved, English users byte-identical, graceful fallback), and the concatenated-bundle mechanics all check out. One real bug: the dictionary is fetched with a **relative URL**, so on multi-segment pages — `/game/<name>`, `/game/<shareID>/<name>`, `/tutorial/<name>`, `/library/<folder>/<name>` — the fetch silently fails and those pages stay English. That's exactly where the welcome/link-details overlay lives, whose strings this PR translates.

## Findings

### 1. Dictionary fetch breaks on `/game/`, `/tutorial/` and `/library/` URLs (bug)

`client/js/i18n.js:42` fetches `i18n/${language}.json` as a URL relative to the page. room.html isn't only served at single-segment `/:room` — it's also served at the multi-segment routes in `server.mjs:451-455`. From those pages the relative URL resolves wrong. Verified against a running server:

- from `/game/<name>`: resolves to `/game/i18n/de.json`, which matches `/game/:shareID/:name` → gameRoomHandler serves **HTML**, so `r.json()` throws and the `.catch` leaves the dictionary empty;
- from `/game/<shareID>/<name>` and `/library/<folder>/<name>`: resolves to a 4-segment path → **404**, same silent English fallback.

So for a German/Brazilian user following a shared game link, the welcome page ("Welcome to", "Create room", "Joining room...", "Game not found!" …) — one of the main surfaces this PR translates, and the only place those strings ever appear — stays English with no error. Fix: build the URL from the root like the rest of the client does (`client/js/domhelpers.js:621` `getBaseURL()`, or `config.urlPrefix` as in `client/js/editmode.js:1241`): `fetch(\`${config.urlPrefix}/i18n/${language}.json\`)`. The `config` global is inlined ahead of the bundle, and the fetch branch is unreachable in the Node/jest import path (language resolves to `en` there), so the existing guards still hold.

### 2. Minor: dictionary/websocket race for server-driven strings

The dictionary fetch and the websocket start around the same time; on a slow connection the first `fillStatesList()` (shelf category titles, "Any"/"None" filter options via `translate()`) can run before the dictionary resolves and render English. It self-heals on the next state update and the one-time `translateDOM()` pass also fixes the static parts, so this is acceptable — just noting it's a known window, not a guaranteed ordering.

### 3. Nit: `$` is special in the replacement string

`client/js/i18n.js:66` — `node.nodeValue.replace(trimmed, dictionary[trimmed])`: in `String.replace`, `$&`/`$'`/`` $` `` in the *replacement* are special. No current dictionary entry contains `$`, but a future translation with a dollar sign would corrupt output. `node.nodeValue.replace(trimmed, () => dictionary[trimmed])` sidesteps it permanently.

## Checked and fine

- **Bundle mechanics:** no top-level identifier collisions between `i18n.js` and any other concatenated file; uglify-es minifies it; the full `minifyHTML()` build succeeds with the new entry; bundle order (i18n.js after domhelpers.js) is safe since neither calls the other at eval time.
- **Node/jest import path:** the circular `domhelpers ⇄ i18n` import works (only hoisted function declarations are used at init), jsdom resolves to `en`, and `npm test` passes 15/15.
- **Filter logic:** materializing `value` on `<option>`s before translating keeps all `== 'Any'`-style comparisons and the `filterByLanguage` value/display split working; dynamic option rebuilds in states.js correctly translate display text only.
- **English users:** no fetch, no DOM walk, byte-identical behavior — so TestCafe/CI (en-US) is genuinely unaffected.
- **No injection risk:** dictionary values are repo-controlled static files; the few `innerHTML` insertions use them directly and user-controlled values keep going through `html()`.
- The `progressButton` change (capturing the label at click time) is a genuine fix — setup-time capture would restore the pre-translation English label.

## Beta label

The `engine` label stems only from `client/js/widgets/widget.js`, and those hunks are incidental: they wrap the *default* input-overlay button labels ("Go"/"Cancel"/"Working...") in `translate()`, which is an identity for English users and display-only for others — games that set their own button text are untouched, and nothing in state sync, routine semantics, or widget behavior changes (`serverstate.js`/`compute.js` untouched). So this PR does not really change the engine, and a maintainer could allow-list it for the beta build (the builder will still filename-skip `widget.js` until then). The server-side change is only adding one file to the bundle list (build verified), finding 1 degrades gracefully to English, and the feature is exactly the kind of thing real players should exercise.

I added the `beta` label because the PR is safe and useful for players to test — the engine-file edits are incidental display-only wraps as explained above, and the one bug found only makes some pages fall back to English.

VTT-AI-Agent:
> translate About and the entire editor

*(Source: [Discord message from arnoldsmith86](https://discord.com/channels/770758631146782780/1527770743813832867/1527778552240672838))*

Done in cee5d923. What changed:

- **About tab**: the full content — gameplay/privacy sections and the attribution/license text — is now in both dictionaries (link fragments included, so mixed paragraphs with inline links read naturally).
- **Editor**: toolbar (incl. save menu and all tooltips), drag toolbar, sidebar buttons and all eight sidebar modules, the add-widget overlay, the legacy edit overlays, the library-decks browser and the JSON editor chrome are translated. Editor `alert`/`confirm`/`prompt` dialogs go through `translate()`.
- **Mechanism**: the editor DOM doesn't exist at page load, so `i18n.js` now exports `translateSubtree()` (one-shot walk when edit mode initializes) and `translateOnChange()` (a MutationObserver scoped to `#editor` that keeps the constantly recreated sidebar-module content translated). The walker additionally skips `data-field` elements so a full-page re-walk can never touch displayed game metadata. Deliberately **not** translated: widget property names, widget-type identifiers, delta-cause strings (persisted in the room history), material icon ligatures, and the `Unnamed` default game name (code compares against it).
- English stays byte-identical (no observer, no walk), so TestCafe/CI is unaffected. Jest 15/15; verified in Chromium with `de-DE`: About, all sidebar modules (including ones opened later — observer path), add-widget overlay, and an `en-US` regression pass.

Screenshots (German): [About](https://agent.virtualtabletop.io/reports/pr/1527770743813832867/about-de.png) · [Toolbox module](https://agent.virtualtabletop.io/reports/pr/1527770743813832867/toolbox-de.png) · [Add-widget overlay](https://agent.virtualtabletop.io/reports/pr/1527770743813832867/addwidget-de.png)

VTT-AI-Agent:
Thanks for both reviews — addressed in d3950a3e:

**UI/UX review:**
1. **`<html lang>`** — fixed: `translateDOM()` now sets `document.documentElement.lang` to the resolved language (verified: `de` on a German page, `en` unchanged).
2. **Sentence-fragment grammar** — the two concrete breakages are fixed within the current pattern by rewording the German fragments: the PCIO sentence now reads *"Bitte erstelle ein Issue auf GitHub bei einer nicht spielbaren Datei."* (no subordinate clause, so no comma needed at the fragment boundary), and the icon-picker sentence no longer doubles "angezeigt" (works with the conditional `<span>` both shown and hidden). I agree whole-sentence `{link}` keys are the better long-term shape for link-embedded sentences — that changes how the walker handles element children, so I'd prefer it as a follow-up PR before more dictionaries get contributed, as you suggested.
3. **Reload warning** — added to the hint: "The page reloads to apply the change." (translated in both dictionaries; the English hint text and key updated together).
4. **`#languageSelect` accessibility + styling** — added `aria-label="Language"` and added `aria-label` to `translatedAttributes` so it localizes (announces "Sprache" in German); the select now uses the app's blue border/text styling ([screenshot](https://agent.virtualtabletop.io/reports/pr/1527770743813832867/players-de.png)).
5. **German wording** — "nach Namen" and "Skill: → Schwierigkeit:" applied.

**Code review:**
1. **Relative dictionary fetch on `/game/…` URLs** — I could not reproduce this: room.html sets `<base href="<urlPrefix>/">`, and `fetch()` resolves relative URLs against the document base, so from `/game/agra` the request goes to `/i18n/de.json` — my earlier verification run literally shows the `/game/agra` welcome overlay rendered in German. That said, relying on `<base>` for a fetch is implicit and fragile, so I switched to building the URL explicitly from `config.urlPrefix` as suggested. Either way the page now behaves identically.
2. **Dictionary/websocket race** — agreed it's a small self-healing window; leaving as is per your note.
3. **`$` in replacement string** — fixed with a replacement function (`replace(trimmed, _=>dictionary[trimmed])`).

Jest 15/15; re-verified in Chromium (`de-DE` + `en-US` regression).

VTT-AI-Agent (commented):
## TL;DR

Follow-up review of the two commits since the last reviews — cee5d923 (About + editor translation) and d3950a3e (review fixes). All previously requested fixes are verified present and correct, and the new editor-translation path works in a live browser (German room + editor, zero console errors, jest 15/15). One minor new finding: the `#editor` MutationObserver can translate **saved-widget names** in the Widgets sidebar when a name exactly matches a dictionary key — a small game-content leak, display-only.

## Verified

- **d3950a3e fixes all check out in code:** dictionary fetched from `config.urlPrefix` (`client/js/i18n.js:38`; `config` is always inlined ahead of the bundle, and the fetch branch stays unreachable in Node/jest since the language resolves to `en` there), `document.documentElement.lang` set (`i18n.js:111`), `$`-safe replacement function (`i18n.js:68`), `aria-label` added to `translatedAttributes` so the selector's label localizes.
- **Editor translation (cee5d923) live-tested** (Chromium, `de-DE`, server on this branch): entering edit mode translates the toolbar/sidebar (all eight module tooltips render in German), no page or console errors. `initializeEditMode` runs only once (guarded by `edit === null`, `client/js/main.js:499`), so exactly one MutationObserver is attached and the whole-body re-walk happens once.
- **Dictionaries are consistent:** `de.json` and `pt-BR.json` have identical 592-key sets; every placeholder key used in code (`{id}`, `{error}`, `{list}`, `{name}`, `{count}`, `{zoom}`) exists in both and retains its placeholder; and no dictionary value is itself another key, so the edit-mode re-walk of already-translated text can't double-translate.
- **Game/user content stays safe on the re-walk:** player names live in input `value` properties (untouched), group names in a `value` attribute (not in `translatedAttributes`), widget JSON in a textarea whose full text can't match a key, and `data-field` metadata is skipped.
- jest 15/15 at HEAD; the server builds the bundle with `i18n.js` and serves `/i18n/de.json` as static JSON.

## Finding (minor)

1. **Saved-widget names can get translated in the Widgets sidebar.** `client/js/editor/sidebar/widgets.js:1152` renders `<div class="widget-name">${html(state.name || state.id)}</div>` — raw game content as a text node inside `#editor`, which the observer from `translateOnChange()` walks. A reusable widget saved with a name that exactly equals a dictionary key — and many keys are plausible widget names: "Save", "Board", "Cards", "Token", "Pawn", "Timer" — will display translated ("Speichern", "Brett", …) for non-English users. Display-only (the rename input uses the `value` property, so editing/saving is unaffected), but it breaks the "never touch game content" invariant the walker otherwise maintains carefully. The same theoretical exposure exists for library-deck `definition.name/description/attribution` (`client/js/editor/sidebar/properties.js:384-386`), though an exact key match there is unlikely. Cheap fix using the mechanism you already built: put `data-field` (or an equivalent skip class) on `.widget-name` and the deck-detail divs so `skipTranslation` catches them.

The agreed follow-up (whole-sentence `{link}` keys for link-embedded sentences) remains a good next step before more dictionaries are contributed, but the reworded German fragments are fine for now.

## Beta label

The `engine` label still stems only from `client/js/widgets/widget.js`, and those hunks remain incidental: three `translate()` wraps of the *default* input-overlay button labels ("Go"/"Cancel"/"Working...") — an identity for English users and display-only for others; `serverstate.js`/`compute.js` are untouched and no state-sync, routine, or widget behavior changes. So this PR does not really change the engine, and a maintainer could allow-list it for the beta build.

The `beta` label was already present and I left it in place — the PR runs cleanly (verified live in this review), degrades gracefully to English on any failure, and the one finding above is a minor display-only edge case in edit mode.

VTT-AI-Agent (commented):
## TL;DR

The approach is solid and the previously-flagged `urlPrefix` fetch bug is properly fixed — I verified the bundle mechanics, dictionary consistency and filter-value handling, and jest passes 15/15. One real remaining bug: the `MutationObserver` from `translateOnChange($('#editor'))` also covers the JSON and Debug sidebar modules, which reparent `#jeText`/`#jeTextHighlight`/`#jeLog` *into* `#editor` — so for German/pt-BR users the observer translates **game JSON content** in the JSON editor's visible layer, and in one edge case can silently write a translated string into the widget JSON itself.

## Findings

### 1. `#editor` observer translates game JSON in the JSON/Debug sidebar modules (bug)

`client/js/editmode.js:1654` installs `translateOnChange($('#editor'))`. That's fine for the toolbar/sidebar chrome, but `JsonModule.renderModule()` (`client/js/editor/sidebar/json.js:57-61`) moves `#jeWidgetSwitcher`, `#jeTextHighlight`, `#jeText` and friends *into* the module DOM, which lives inside `#editor` (`#editorModuleTopLeft` etc.). While the JSON module is open:

- Every `jeColorize()` run does `$('#jeTextHighlight').innerHTML = ...` (`client/js/jsonedit.js:2377`), emitting childList mutations. The highlight renders each token as its own element — e.g. `"text": "Go"` becomes `<i class=string>Go</i>` — and `translateNode()` matches the whole trimmed text node against the dictionary. So a widget with `"text": "Go"` **visibly renders as `"text": "Los"`** for a German user. Because `#jeText` has `-webkit-text-fill-color: transparent` (the highlight layer is what the user actually sees), the displayed JSON diverges from the real content, and since "Los"/"Abbrechen" etc. differ in length from the underlying text, the caret and selection misalign for the rest of that line. Values like `Go`, `Cancel`, `Play`, `Cards`, `Board`, `Token`, `None` are extremely common in real game JSON, so this will be hit routinely by non-English game editors.
- Data-corruption edge case: `#jeText` itself (contenteditable) is inside the observed subtree while the module is open. Pasting a fragment that exactly equals a dictionary key (e.g. copying the word `Cancel` into a string value) inserts a new text node → the observer rewrites it to `Abbrechen` → `jeGetEditorContent()` reads the translated text back and it gets **saved into the widget JSON**. Rare, but it's silent corruption of game data.
- Same root cause, cosmetic: the Debug module reparents `#jeLog` (routine log containing game values) into `#editor` (`json.js:132`), and the Widgets module tree displays widget names/ids as plain text nodes — a widget literally named "Cards" or "Board" shows up translated in the list, which contradicts the "game content is never touched" guarantee.

Suggested fix: have the walker refuse to enter the JE surfaces. `skipTranslation()` only checks the element itself, but the observer's `addedNodes` are children (e.g. `.jeTextLine` divs), so it needs an ancestor check — e.g. in the observer callback (or at the top of `translateNode`) bail when `node.parentElement?.closest('#jeText, #jeTextHighlight, #jeLog')` is non-null, plus adding those ids to `skipTranslation()` for the initial `translateSubtree($('body'))` walk. That covers both the display divergence and the paste-corruption path; the Widgets-module name display could be handled separately (or accepted) since it's display-only.

### Verified while reviewing (no action needed)

- The dictionary fetch now uses `config.urlPrefix || ''` with an absolute path; `config` is defined immediately before the bundle in the same module script and `client/i18n/` is covered by the existing `express.static` route, so the earlier multi-segment-URL bug is fixed.
- Both dictionaries have identical key sets (592 each), all `{placeholder}` tokens are consistent between key and value, every literal `translate('...')` key in the code exists in both dictionaries, and no German/Portuguese value collides with a *different* English key — so the walker is idempotent and the double-walk from `translateSubtree($('body'))` at editor init is safe.
- `<option>` value materialization keeps all filter comparisons (`'Any'` etc.) working; the dynamic filter rebuilds in `states.js` keep English values in `value` attributes and translate labels only.
- `progressButton`'s label capture at click time is race-free (no `await` before `disabled = true`, restore happens before re-enable) and fixes the restored-label-reverts-to-English problem.
- English users are untouched (no fetch, no DOM walk), so TestCafe/CI behavior is unchanged; the Node import guard works because `localStorage` is undefined there and the language short-circuits to `en`. Jest: 15/15 passing locally.

## Beta / engine note

Of the changed files only `client/js/widgets/widget.js` is an engine file, and its edits are incidental: three one-line `translate()` wraps of the input-overlay *default* button labels ("Go"/"Cancel"/"Working..."). For English users `translate()` is an identity function (no dictionary is even fetched), and for others it only changes display text when a game doesn't specify its own button text — no state-sync, routine-semantics or widget-behavior change. So this PR does not really change the engine and a maintainer could allow-list it for the beta build.

The `beta` label was already present and I left it in place — the PR is safe to test on beta (the finding above only affects non-English users inside the editor's JSON/Debug modules and can't corrupt games during normal play), and its engine-file edits are incidental as explained above.

VTT-AI-Agent:
Both new findings addressed in 2315555f:

1. **JSON/Debug module game-content leak (the JE reparenting bug)** — fixed exactly as suggested: `skipTranslation()` now refuses `#jeText`, `#jeTextHighlight` and `#jeLog`, and the MutationObserver callback checks **ancestors** of added nodes (`insideSkippedSubtree()`) before translating, which also closes the paste-into-contenteditable corruption path. Verified live with a room containing `"text": "Go"`: in a German session the JSON editor's highlight layer and the editable text both keep `"Go"` (no "Los"), while the JE chrome (context buttons, hints) stays translated.
2. **Saved-widget names / library-deck details** — `skipTranslation()` now also honors the standard `translate="no"` attribute, and it's applied to `.widget-name` in the Widgets module and to the traditional-deck details (`name`/`description`/`attribution`) in the properties module.

Jest 15/15; re-verified in Chromium (`de-DE`).

VTT-AI-Agent:
Updated to `main` (merge of the last 70 commits) and acted on this feedback:

> update this to main's latest version. needs a lot more translation - especially in the editor. do all of it.
>
> also look at the toolbar: you need to use shorter translations so they fit in the wide toolbar on the left.

**Merged `main`.** The conflicts were the players tab (rewritten into the session table), the toolbox (circle align moved to the arrange bar), `minify.mjs` (bundle lists restructured), `main.js` and `widget.js` imports, and `room.html` (players help). The language selector now sits above the new players help sections.

**Translation.** The dictionaries went from ~590 to ~2600 entries each. Getting the editor covered took more than dictionary entries:

- the editor overlays that edit mode moves next to the room (add widget, deck editor dialogs, library decks) were never observed for changes, so everything they recreated stayed English — they are now observed individually;
- labels the editor builds by string interpolation (widget type and box, snap grid presets, lock states, layer notes, the deck editor's panel titles and tree toolbar, the players table's connection column) can never match a dictionary key, so they now use `translate()` with `{placeholder}`s;
- icon elements are skipped, because a `material-symbols` ligature name like `share` or `delete` collides with real UI wording, and `#jeTree` is skipped because it shows widget ids.

Walking the running editor in both languages — every sidebar module, the properties panel per widget type, the deck editor, all toolbar popups — now leaves only brand names, CSS keywords, language tags and identifiers in English.

**Toolbar.** The wide (200 px) toolbar shows the tab labels at 20 px with `white-space: nowrap`, so the literal translations overflowed: "Bearbeitungsmodus" by 45 px, "Estante de Jogos" by 20 px, "Modo de Edição" by 7 px. Those now use short forms — *Edit Mode* → "Bearbeiten"/"Editar", *Game Shelf* → "Spieleregal"/"Estante", *Hide Toolbar* → "Leiste ausblenden"/"Ocultar barra" — and every label of both languages ends inside the toolbar with room to spare:

toolbar

Jest is 1489/1489; the German and Brazilian Portuguese UI was walked in a real browser. English output is unchanged.

PR #3038, opened 2026-07-18
Support "minutes:seconds" time format in timers

VTT-AI-Agent:
## What

Timers now understand time strings in the format **`"minutes:seconds"`** or **`"minutes:seconds.fraction"`** (e.g. `"2:30"`, `"0:05.25"`), converted to milliseconds and rounded to the nearest millisecond when the fraction has more precision than that. The format is accepted in:

- the **`start`** and **`end`** properties of the timer widget,
- the **`value`** and **`seconds`** parameters of the **TIMER** routine operation,
- and when TIMER's `value` names another property (e.g. `"value": "customTime"`), a time string stored in that property is converted to milliseconds before being used.

## Why

Specifying timer times in raw milliseconds (`150000` for 2½ minutes) is unintuitive for game creators; `"2:30"` matches what the timer displays.

## How

- New `timeToMS()` helper in `client/js/domhelpers.js` (`/^(-?)(\d+):(\d+(?:\.\d+)?)$/` → milliseconds via `Math.round`; any other value is returned unchanged, so plain numbers keep working exactly as before).
- `Timer.setMilliseconds()` converts `start`, the passed value, and the value read from a property; `Timer.onPropertyChange()` converts `end` before the alert comparison.
- The TIMER operation converts `seconds` (a converted string is already milliseconds, so it is not multiplied by 1000) and leaves numeric behavior (`a.seconds*1000 || a.value`) untouched.
- Validator: TIMER's `seconds` now also accepts strings matching the time format (`value` already accepted strings, since it can name a property).
- New jest unit test `tests/client/time-to-ms.test.js` covering conversion, fraction rounding, negative times and pass-through of other values.

## Verification

- `npm test`: 19/19 pass (including the new suite).
- Verified end-to-end in Chromium against a demo game: `start: "2:30"` reset → `150000` ms, `seconds: "0:05.25"` → `5250` ms, `value: "1:30"` → `90000` ms, `value: "customTime"` reading a `"1:15"` property → `75000` ms, `inc` by `"0:10"` → `+10000` ms, and counting down past `end: "0:00"` sets `alert`.
- Demo save file: [timer-format-demo.vtt](https://agent.virtualtabletop.io/reports/pr/1527966985340846170/timer-format-demo.vtt) — screenshot: [timer-demo.png](https://agent.virtualtabletop.io/reports/pr/1527966985340846170/timer-demo.png)

Creator-facing change: the wiki pages for the timer widget and the TIMER function should mention the new accepted time format once this is merged.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3038/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

Solid, well-tested, backward-compatible enhancement — the conversion logic is correct across all the paths I traced (widget `start`/`end`, TIMER `seconds`/`value`, value-as-property-name), and `npm test` passes (19/19) locally. Only minor observations below; nothing blocking.

## Findings

I traced every consumer of `timeToMS()` against the old behavior and found no regression: plain numbers take exactly the old code path, non-matching strings behave as before (inert / fall through to `value`), and `null`/`undefined` pass through so the `end !== null` alert guard in `timer.js:92` keeps its semantics. The validator regex matches the runtime regex, and TIMER's `value` already accepted arbitrary strings, so no further validator/jsonedit upkeep is needed (no new property was added). Minor notes:

1. **`seconds: "0:00"` behaves differently from `seconds: 0`** (`client/js/widgets/widget.js:2114`). Numeric `0` falls through to `a.value` (`a.seconds*1000 || a.value`), while the string `"0:00"` converts to `0` and is used directly, ignoring `value`. That's arguably the *more* correct behavior for an explicit `"0:00"`, so fine to keep — just noting the asymmetry is intentional-looking but undocumented.
2. **Seconds ≥ 60 are accepted**: `"1:90"` converts to 150 s. Lenient parsing is reasonable (and the display re-normalizes), so no change needed — but if you'd rather reject malformed input, the runtime and validator regexes would both need tightening together.
3. In `setMilliseconds` (`client/js/widgets/timer.js:105`), `value` is converted *before* the property-name lookup, so a widget property literally named like a time string (e.g. `"1:30"`) can no longer be referenced by TIMER's `value`. That's contrived enough to ignore; mentioning it only for completeness.

The PR description's note about updating the wiki (timer widget + TIMER function pages) after merge is the right call.

## Beta label

This PR is a genuine engine change — it alters TIMER routine-operation semantics in `client/js/widgets/widget.js` and timer-widget `start`/`end` handling in `client/js/widgets/timer.js` (time strings that were previously inert now convert to milliseconds) — so it is excluded from beta auto-deployment by design and I did not add the `beta` label.

96LawDawg (approved):
Tested.  Approved.  Before merging, I would like to update the timer tutorial.

RaphaelAlvez (approved):
tested

PR #3039, opened 2026-07-18
Don't treat a denied fullscreen request as a fatal client error

VTT-AI-Agent:
## What

Clicking the fullscreen toolbar button when the browser refuses to enter fullscreen crashed the client: the unhandled promise rejection from `requestFullscreen()` hit the global error handler in `tracing.js`, which closes the WebSocket connection and shows the "client error" report overlay — for something that isn't a bug in the room at all. This PR catches the rejection (and the one from `exitFullscreen()`) and just logs a console warning instead.

## Why

This came from a client error report on the MAIN server (id `2uru38he`, Chrome 150): `TypeError: Permissions check failed` with no JS frames in the stack. That message is Chromium's rejection reason for a fullscreen request that fails its permission checks (see `RequestFullscreenError::kPermissionCheckFailed` in Blink's `fullscreen.cc`) — it happens e.g. when the page is embedded in an iframe without `allowfullscreen`, or when the transient user activation has expired. Newer Chromium builds word some of these denials as "Disallowed by permissions policy" — same code path. The editor toolbar's fullscreen toggle funnels into the same `#fullscreenButton` handler, so it's covered too.

## Verification

Reproduced with Playwright by loading a room in an iframe without `allowfullscreen` and clicking the fullscreen button:

- **before:** unhandled rejection → error report overlay shown, connection closed
- **after:** no overlay, no page error; console warning `Could not enter fullscreen mode: …`

jest suite passes (15/15). No widget/routine property changes, so no validator or jsonedit updates needed.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3039/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR
Correct, minimal fix: catching the promise rejections from `requestFullscreen()`/`exitFullscreen()` stops a browser-denied fullscreen request from tripping the global error handler (connection close + error overlay). No issues found.

## Details

- This is the only fullscreen call site in the client (`client/js/main.js:620-635`), so the fix covers both the room toolbar and the editor toolbar, which share the `#fullscreenButton` handler.
- The legacy `webkitRequestFullScreen` fallback branch doesn't return a promise in the engines where it's the active path, so leaving it uncaught is fine.
- Both rejection reasons are DOM exceptions/`TypeError`s with a `.message`, so the `console.warn` template is safe. A user-visible toast instead of a silent console warning could be a nicety, but silently doing nothing matches how browsers themselves treat denied fullscreen, so this is fine as-is.
- No state, widget, or routine changes — no validator/jsonedit/test-suite impact, matching the PR's claim.

Added the `beta` label: the change only touches the fullscreen button handler in `client/js/main.js` (not an engine file), and it can only reduce client errors, so it is safe for players to test.

ArnoldSmith86 (approved):

PR #3040, opened 2026-07-19
Don't crash the client on an unparseable server message

VTT-AI-Agent:
## Problem

A client error was reported from the main server (Safari): `JSON Parse error: Unterminated string`, thrown from the WebSocket message handler during normal room use.

The client's `connection.onmessage` handler parsed incoming frames with a bare `JSON.parse(e.data)` and **no error handling**:

```js
connection.onmessage = (e) => {
  const { func, args } = JSON.parse(e.data);
  ...
```

If a frame arrives corrupted or truncated in transit (some browsers occasionally deliver incomplete WebSocket frames — the "unterminated string" symptom), `JSON.parse` throws an uncaught exception. That propagates to the global `window.onerror` handler in `tracing.js`, which shows the **fatal client-error overlay, closes the connection, and prevents reconnection** — ending the player's session over a single bad frame.

Notably the server side already guards against this: `server/connection.mjs` wraps its own `JSON.parse` of incoming messages in a `try/catch`. The client half was missing the symmetric guard.

## Fix

Wrap the client's `JSON.parse` in a `try/catch`. On a parse failure, log it and close the connection so the existing auto-reconnect logic re-establishes the session and re-syncs the full room state — instead of crashing with a fatal, unrecoverable error overlay.

## Testing

- `npm test` — all jest suites pass.
- Change is confined to the WebSocket message handler; server frames (always produced via `JSON.stringify`) parse exactly as before, so the happy path is unchanged.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3040/pr-test (or any other room on that server)

### DIFF
--- a/client/js/connection.js
+++ b/client/js/connection.js
@@ -42,7 +42,18 @@ export function startWebSocket() {
   };
 
   connection.onmessage = (e) => {
-    const { func, args } = JSON.parse(e.data);
+    let func, args;
+    try {
+      ({ func, args } = JSON.parse(e.data));
+    } catch(error) {
+      // A message that fails to parse was corrupted or truncated in transit
+      // (some browsers occasionally deliver incomplete WebSocket frames).
+      // Instead of crashing the whole client with an uncaught error, drop the
+      // connection so the existing reconnect logic re-syncs the full room state.
+      console.error('Could not parse message from server. Reconnecting.', error);
+      connection.close();
+      return;
+    }
 
     if(func == 'serverStart') {
       if(serverStart != null && serverStart != args) {


VTT-AI-Agent (commented):
## TL;DR

Solid, correct, low-risk defensive fix: it wraps the client's WebSocket `JSON.parse` in a `try/catch` and, on failure, closes the connection so the existing reconnect logic re-syncs full room state instead of the session dying on a fatal error overlay. The close-and-resync approach is the right choice (safer than silently dropping a frame, which could desync state). No blocking issues found.

## Findings

**Correctness — the approach is sound.** `connection.close()` triggers the existing `onclose` handler (`connection.js:34`), which shows the connection-lost overlay and schedules `startWebSocket` via `setTimeout`. On reconnect the client re-runs the `room` handshake and the server pushes full state, so a lost/corrupted delta can't leave the client silently desynced. Choosing to drop+resync rather than just skipping the bad frame is the correct call.

**Minor — no test added.** `connection.js` already carries a `mockConnection` hook for unit tests, and the repo convention is to cover behavioral changes. A tiny test that feeds `onmessage` an unparseable frame and asserts `connection.close()` is called (and no throw escapes) would lock in the regression guard cheaply. Not blocking, but worth considering given how easy it is to reintroduce a bare `JSON.parse`.

**Minor / pre-existing — reconnect backoff never resets.** `lastTimeout` doubles on every `onclose` (`connection.js:41`) and is never reset to `1000` on a successful `onopen`. If a server were to persistently emit an unparseable frame right after each reconnect (unlikely, since server frames always go through `JSON.stringify`), this fix would participate in an ever-growing backoff loop. This is existing behavior, not introduced here, and the failure it guards against is genuinely transient — so it's fine to leave, just noting the interaction.

Overall this mirrors the symmetric guard already present in `server/connection.mjs` and is exactly the kind of hardening the client half was missing. Looks good to merge.

## Beta label

`connection.js` is not one of the beta-excluded engine files (`client/js/widgets/`, `compute.js`, `serverstate.js`), the change is a purely defensive `try/catch` that leaves the happy path byte-for-byte identical, and it cannot corrupt games or crash the server — so I added the `beta` label for player testing.

ArnoldSmith86:
Does the reconnection logic trigger when the connection was closed explicitly like this?

ArnoldSmith86:
@VTT-AI-Agent Does the reconnection logic trigger when the connection was closed explicitly like this?

ArnoldSmith86:
Does the reconnection logic trigger when the connection was closed explicitly like this?

VTT-AI-Agent:
Yes. Calling `connection.close()` fires the existing `onclose` handler; because this path does not call `preventReconnect()`, `lastTimeout` remains set and `onclose` schedules `startWebSocket` (with its normal exponential delay). The new connection then receives the usual room-state sync.

ArnoldSmith86 (approved):

PR #3043, opened 2026-07-20
Fix crash when adding a widget that gets refused in edit mode

VTT-AI-Agent:
## Problem

A client error was reported from the main server (Chrome):

```
TypeError: Cannot read properties of undefined (reading 'setHighlighted')
    at setSelection (edit.js)
    at overlayDone (edit.js)
```

## Cause

`overlayDone(id)` (in `client/js/editmode.js`) is called by the add-widget overlay handlers with the id returned by `addWidgetLocal`. It then does:

```js
setSelection([ widgets.get(id) ]);
```

However `addWidgetLocal` (in `client/js/serverstate.js`) returns `null` when it **refuses** to add a widget — e.g. an invalid/missing parent, an invalid deck, or an invalid cardType. In that case `overlayDone(null)` runs `setSelection([ widgets.get(null) ])`, i.e. `setSelection([undefined])`, and `setSelection` crashes when it calls `.setHighlighted(true)` on the `undefined` entry. The same happens if the just-added widget no longer exists by the time the selection runs.

## Fix

Only update the selection when the widget actually exists. The add overlay is still hidden in every case, so a refused add now simply closes the overlay without a console error instead of throwing.

```js
function overlayDone(id) {
  const widget = widgets.get(id);
  if(getEdit() && widget)
    setSelection([ widget ]);
  showOverlay();
}
```

## Testing

- `npm test` passes.
- Purely a defensive client-side guard around a selection update; it does not change room state, so TestCafe state hashes are unaffected.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3043/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

Correct, minimal defensive fix: `overlayDone(id)` now guards on the widget actually existing before calling `setSelection`, preventing the `Cannot read properties of undefined (reading 'setHighlighted')` crash when `addWidgetLocal` refuses an add and returns `null`. No concerns — looks good to merge.

## Findings

I traced the whole path and the fix holds up:

- `addWidgetLocal` (`client/js/serverstate.js:130`) returns `null` on every refusal branch (invalid parent, invalid deck, invalid cardType), so `overlayDone(null)` was reachable.
- `widgets.get(null)` / `widgets.get(<gone id>)` yields `undefined`, and `setSelection` (`client/js/editor/selection.js:157`) unconditionally calls `widget.setHighlighted(true)` on each entry — so `[undefined]` crashes exactly as the stack trace shows.
- The new guard `getEdit() && widget` closes the overlay in all cases (`showOverlay()` still runs unconditionally) while only touching the selection when the widget exists. Non-edit-mode behavior is unchanged.

The change reads the widget once and reuses it, which is a small tidy-up over the previous `widgets.get(id)` inside the array. All existing `overlayDone` callers pass an id (or `null`), so none regress.

Nothing else worth raising — the guard is the right layer for this defensive check and doesn't alter room state.

## Beta label

Added the `beta` label: this touches only `client/js/editmode.js` (edit-mode UI), not the engine files (`client/js/widgets/`, `compute.js`, `serverstate.js`), and it's a safe, crash-preventing guard that changes no room state — safe and useful for players to test.

ArnoldSmith86:
Find a way to reproduce this on production/main.

VTT-AI-Agent:
Reproduction on the current `main` code:

1. Open any room and enter **Edit Mode**.
2. Click the editor toolbar’s **Add widget** (`+`) button.
3. Click **Upload Token** (the same happens with **Upload Board**).
4. Dismiss the native file picker without choosing a file.

The browser console then reports the exact error:

```
TypeError: Cannot read properties of undefined (reading 'setHighlighted')\n    at setSelection (edit.js)\n    at overlayDone (edit.js)\n```\n\nCancelling makes `uploadAsset()` resolve to `null`; `uploadWidget()` still calls `overlayDone(undefined)`, which attempts to select `widgets.get(undefined)`. The PR’s guard covers this path as well.

ArnoldSmith86:
> The browser console then reports the exact error

While that is correct, my Firefox doesn't show me the error report overlay though so that's probably not how the user tirggered it.

ArnoldSmith86:
Find a way to reproduce this on production/main so it actually triggers the crash reporter.

VTT-AI-Agent:
Found a reporter-triggering reproduction. Use a throwaway room on production/main and import [this minimal VTT](https://agent.virtualtabletop.io/reports/pr/1528748350302846976/overlaydone-crash-repro.vtt).

1. Open the Game Shelf, choose **Add game** → **Upload files**, and select the VTT.
2. Open the imported game and press its **Play** button.
3. Enter **Edit Mode**, click the editor toolbar’s **Add widget** (`+`), then click **Add Basic Widget**.

This reliably opens the client crash reporter with the reported `setSelection` / `overlayDone` `setHighlighted` error. The minimal game has an `idGlobalUpdateRoutine` that deletes the newly added widget before `addWidgetLocal` returns; `overlayDone` then tries to select its now-stale ID. I verified the complete import-and-click flow against current `main`, including that `#clientErrorOverlay` is displayed.

ArnoldSmith86 (approved):

PR #3046, opened 2026-07-21
Make games translatable via language-suffixed properties

VTT-AI-Agent:
## Goal

Building on the UI-translation PR (#3033, which makes the **UI** translatable and
resolves the selected language), this makes **games** translatable: instead of
maintaining a separate copy of a game per language, a game author can add
per-language overrides directly to the game.

Any **widget property** or **game-metadata field** may carry a language suffix,
e.g. `x:en-US`, `label:de-DE`, `description:pt-BR`. When the suffix matches the
currently selected UI language, the suffixed value overrides the unsuffixed one.
For example `x:en-US` overrides the normal `x` whenever `en-US` is selected.

> add the possibility for games to be translatable. Allow language suffixes for all widget properties and meta fields that use the suffixed property if it matches the currently selected UI language. For example `x:en-US` should override the normal `x` whenever language `en-US` is selected. For testing add German and Brazilian Portuguese overrides to the ten most played games on the production server.

## How it works

- **Local, read-time resolution.** The override is applied on read (`get()`),
  never written back into the room state, so the shared state is untouched and
  two players can view the same game in different languages simultaneously.
- **`client/js/i18n.js`** gains `languageOverrides(object)` (returns a
  base-property → value map for the active language, `null` when there are none —
  the common case, kept allocation-free) and `localizeMeta(object)`. Suffix
  matching prefers the most specific: exact tag → primary subtag (`de-DE` ↔ `de`)
  → same language / other region. Only keys whose suffix looks like a BCP-47 tag
  are treated as translations, so ordinary property names are never touched.
- **`client/js/statemanaged.js`** recomputes a widget's overrides in
  `applyDelta()` whenever a suffixed key changes, and `get()` returns the override
  first. All state changes (initial load, local `set()`, remote deltas) flow
  through `applyDelta`, so the DOM always reflects the language-resolved value.
- **`client/js/overlays/states.js`** localizes the game-metadata fields (name,
  description, …) when the library metadata arrives from the server.
- **`validator/validate_gamefile.js`** accepts language-suffixed widget
  properties and `_meta.info` fields — using the existing `ALLOWED_LANGUAGES`
  list — and validates them like the base property, instead of flagging them as
  "unknown property" in edit mode.

No new named property is introduced (it is a general suffix convention), so there
is no new `jeCommands` editor button to add.

## Translated games: the ten most-played

Per a follow-up maintainer request, the ten most-played public-library games are
now **fully** translated into German (`de-DE`) and Brazilian Portuguese
(`pt-BR`): **Bhukhar, Word Whirl, Duel 52, Card Games, Five of a Kind, Floodplain
Frenzy, Egyptian Epochs, The Team II, Steam Power, The Team**. Every visible
label, button, instruction, scoreboard header and player-aid legend is
translated; game/brand names, dynamic values, the engine seat prompt and hidden
helper widgets are left as-is. Where a translation would overflow its
English-sized widget, a `css:<lang>` font-size (11px readable floor) or a
`width:<lang>`/`height:<lang>` override is added — verified by rendering each game
in both languages with zero overflowing widgets. See the PR comments for sample
screenshots.

## Multiplayer determinism

Language suffixes are a **presentation** feature and the engine keeps game logic
language-independent on two fronts:

- The routine interpreter reads properties via `StateManaged.getRaw()` (the
  unsuffixed shared-state value) — `${PROPERTY …}`, `GET`, `SELECT`, `SCORE`,
  `RESET`, `SORT` — so every player computes the same result regardless of the
  language they selected. `get()` (with the override) is used only for display.
- Properties that drive logic/interaction are never localized: every
  `…Routine` property plus `id`, `type`, `parent`, `deck`, `cardType`,
  `inheritFrom`, `owner`, `movable`, `clickable`, `dragging`, `dropTarget`,
  `activeFace`.

Existing games contain no language-suffixed properties, so `getRaw()` is
identical to `get()` for them — no behavior change to any current game.

## Also translated (INPUT dialogs, how-to-play, attribution, language filter)

Beyond in-game widget text, the ten games are now translated end to end:

- **INPUT dialogs** are localized per-viewing-player by the engine
  (`localizeInputOverlay` in `client/js/widgets/widget.js`): overlay
  header/confirm/cancel texts and each field's `label`/`text`/option texts. The
  validator accepts language suffixes on routine-operation properties too.
- **Meta "how to play"** — `_meta.info.ruleText` / `helpText` are translated for
  every game that has them, preserving HTML, embedded game-icon `` tags and
  emojis.
- **Attribution** — `_meta.info.attribution` translated (prose only; names, URLs,
  license IDs and HTML kept verbatim).
- **`language` meta field** — `de-DE` / `pt-BR` added to each game's
  `_meta.info.language` so the library language filters list them.

Verified in a browser: INPUT dialogs, game-details rules and attribution all
render in German.


VTT-AI-Agent (commented):
## TL;DR

Clean, well-scoped implementation of read-time language-suffixed property resolution — the i18n matching, validator and states.js changes look correct. The one thing I'd flag before this becomes general-purpose: because `StateManaged.get()` is the same read path routines use (`${PROPERTY x OF w}`, `GET`/filter ops, `RESET`, etc.), localizing a **logic-affecting** property makes two players in different languages compute different values from the *same shared state*, which can produce divergent deltas and state drift. The test data only translates display labels, so it's safe today, but the feature permits any property.

## Findings

### 1. `get()` localization leaks into routine/game logic → possible per-client divergence (engine, highest priority)

`client/js/statemanaged.js` — the override is applied in `get()`, which is not display-only: it is the property-read path used throughout `client/js/widgets/widget.js` for routine evaluation, e.g. `${PROPERTY x OF w}` (`widget.get(...)`, widget.js:888), the `GET`/filter operations (`widget.get(a.property)`, widget.js:1883-1887), and `RESET` (widget.js:1786). 

Consequence: if a game author suffixes a property that any routine reads (a score value, a `parent`/`deck` link, a coordinate consumed by logic), a player with `de-DE` selected and a player with `en-US` selected read *different values from the identical shared state*, so a routine that then `set()`s a derived value writes conflicting deltas back to the room — i.e. the two clients can desync the shared game. The PR's own framing ("shared room state is never modified, two players may view in different languages") holds only for pure-display properties; it does not hold once a localized property feeds computation.

The chosen test data avoids this (only single-occurrence display labels were translated), so nothing ships broken. But the feature as documented ("any widget property may carry a language suffix") invites it. Worth either (a) restricting suffix resolution to a display allow-list, or (b) at minimum documenting loudly on the wiki that language-suffixed properties must never be read by routines. This is also why I consider the PR a genuine engine change (see beta note).

### 2. `overridesChanged` triggers on any colon-containing key

`statemanaged.js` `applyDelta()`: `if(i.indexOf(':') != -1) overridesChanged = true;` recomputes `languageOverrides(this.state)` for *any* delta key containing `:`, not only ones whose suffix is a real language tag. Today no core property name contains a colon so this is only a (negligible) recompute cost, but it means the recompute trigger is broader than the actual language-suffix definition used everywhere else (the `languageSuffixRegex` / `ALLOWED_LANGUAGES` gate). Tightening the guard to the same regex would keep the trigger and the resolver in sync.

### 3. Minor: matcher looseness vs. validator strictness

`i18n.js` `languageSuffixRegex` accepts any `[a-z]{2,3}(-...)` suffix, while `validate_gamefile.js` `baseLanguageProperty()` only strips suffixes in `ALLOWED_LANGUAGES`. The runtime `languageSuffixScore()` gate (score 0 unless it matches the selected language) makes this harmless in practice, but a custom property literally named `foo:xyz` that happened to match the active language tag would be silently swallowed by `get('foo')`. Not a real-world concern given how property names are formed; noting for completeness.

### Non-blocking observations
- Drag/interaction on a widget whose position (`x`/`y`) is localized will read the override, so a position edit "sticks" only for clients that don't have that override active (override always wins on read). Cosmetic, not corrupting — but another reason to keep suffixes to display text.
- The PR targets the `i18n-ui-translation` (#3033) branch, not `main`; it can't be assessed fully standalone and should merge after/with #3033.
- Validator/JSON-editor upkeep: correctly no new `jeCommands` entry is needed since this is a general suffix convention rather than a named property. Good call-out in the description.

## Beta label

I did **not** add the `beta` label: this genuinely changes engine behavior — the override is resolved inside `StateManaged.get()` and `applyDelta()`, the core property-read/state-sync path that every widget and every routine uses (finding #1), so it is excluded from beta auto-deployment by design and is not merely an incidental helper.

VTT-AI-Agent (commented):
## TL;DR

Clean, well-scoped implementation of read-time language-suffixed property resolution — the i18n matching, validator and states.js changes look correct. The one thing I'd flag before this becomes general-purpose: because `StateManaged.get()` is the same read path routines use (`${PROPERTY x OF w}`, `GET`/filter ops, `RESET`, etc.), localizing a **logic-affecting** property makes two players in different languages compute different values from the *same shared state*, which can produce divergent deltas and state drift. The test data only translates display labels, so it's safe today, but the feature permits any property.

## Findings

### 1. `get()` localization leaks into routine/game logic → possible per-client divergence (engine, highest priority)

`client/js/statemanaged.js` — the override is applied in `get()`, which is not display-only: it is the property-read path used throughout `client/js/widgets/widget.js` for routine evaluation, e.g. `${PROPERTY x OF w}` (`widget.get(...)`, widget.js:888), the `GET`/filter operations (`widget.get(a.property)`, widget.js:1883-1887), and `RESET` (widget.js:1786). 

Consequence: if a game author suffixes a property that any routine reads (a score value, a `parent`/`deck` link, a coordinate consumed by logic), a player with `de-DE` selected and a player with `en-US` selected read *different values from the identical shared state*, so a routine that then `set()`s a derived value writes conflicting deltas back to the room — i.e. the two clients can desync the shared game. The PR's own framing ("shared room state is never modified, two players may view in different languages") holds only for pure-display properties; it does not hold once a localized property feeds computation.

The chosen test data avoids this (only single-occurrence display labels were translated), so nothing ships broken. But the feature as documented ("any widget property may carry a language suffix") invites it. Worth either (a) restricting suffix resolution to a display allow-list, or (b) at minimum documenting loudly on the wiki that language-suffixed properties must never be read by routines. This is also why I consider the PR a genuine engine change (see beta note).

### 2. `overridesChanged` triggers on any colon-containing key

`statemanaged.js` `applyDelta()`: `if(i.indexOf(':') != -1) overridesChanged = true;` recomputes `languageOverrides(this.state)` for *any* delta key containing `:`, not only ones whose suffix is a real language tag. Today no core property name contains a colon so this is only a (negligible) recompute cost, but it means the recompute trigger is broader than the actual language-suffix definition used everywhere else (the `languageSuffixRegex` / `ALLOWED_LANGUAGES` gate). Tightening the guard to the same regex would keep the trigger and the resolver in sync.

### 3. Minor: matcher looseness vs. validator strictness

`i18n.js` `languageSuffixRegex` accepts any `[a-z]{2,3}(-...)` suffix, while `validate_gamefile.js` `baseLanguageProperty()` only strips suffixes in `ALLOWED_LANGUAGES`. The runtime `languageSuffixScore()` gate (score 0 unless it matches the selected language) makes this harmless in practice, but a custom property literally named `foo:xyz` that happened to match the active language tag would be silently swallowed by `get('foo')`. Not a real-world concern given how property names are formed; noting for completeness.

### Non-blocking observations
- Drag/interaction on a widget whose position (`x`/`y`) is localized will read the override, so a position edit "sticks" only for clients that don't have that override active (override always wins on read). Cosmetic, not corrupting — but another reason to keep suffixes to display text.
- The PR targets the `i18n-ui-translation` (#3033) branch, not `main`; it can't be assessed fully standalone and should merge after/with #3033.
- Validator/JSON-editor upkeep: correctly no new `jeCommands` entry is needed since this is a general suffix convention rather than a named property. Good call-out in the description.

## Beta label

I did **not** add the `beta` label: this genuinely changes engine behavior — the override is resolved inside `StateManaged.get()` and `applyDelta()`, the core property-read/state-sync path that every widget and every routine uses (finding #1), so it is excluded from beta auto-deployment by design and is not merely an incidental helper.

VTT-AI-Agent:
Thanks for the thorough review — addressed in `00f4cd09`.

**Finding #1 (get() localization reaches routine reads).** You're right that `get()` is the shared read path, so this is fundamentally a *presentation* feature. Two constraints pull against each other here: the goal explicitly asks for arbitrary properties incl. `x` (position), and display itself is rendered *through* `get()` (`cssTransform()` reads `get('x')`, `setText()` reads `get('text')`, …), so a clean "display-only" resolution that leaves `get()` pure isn't reachable without threading a separate display-getter through every widget render path — a much larger refactor than this change warrants. So I took your option (b), plus a concrete safety net:

- **Structural/identity properties are now never localized** — `id`, `type`, `parent`, `deck`, `cardType` are skipped in `languageOverrides()` (mirrors the non-inheritable set in `statemanaged.js`). This makes the *worst* desync — a per-language override re-linking or re-typing widgets and corrupting the shared hierarchy — impossible regardless of what an author writes.
- **Documented loudly** in `i18n.js` that this is a presentation feature and that routines must read the base (unsuffixed) property for game logic; I'll add the same note to the Widgets / Creating-Games wiki pages. The shipped test data only translates display labels and descriptions, so nothing ships broken.

Full display/logic separation (routines reading canonical values while display shows the translation) is the ideal end state but is best done as a dedicated follow-up.

**Finding #2 (recompute trigger too broad).** Fixed — the trigger now uses the shared `isLanguageSuffixedKey()` (the same `languageSuffixRegex`) instead of `indexOf(':')`, so the trigger and resolver use one definition.

**Finding #3 (matcher vs. validator strictness).** Left as-is: as you note, the `languageSuffixScore()` gate (0 unless the suffix matches the *selected* language) plus the new structural guard make a real collision (a custom property literally named `base:<active-tag>`) a non-issue in practice, and no core property name contains a colon.

**Non-blocking:** agree the `x`/`y` drag "stickiness" is a reason to keep suffixes to display content — now covered by the documentation note. And agreed on merge ordering: this targets `#3033` and should land after/with it.

`npm test` 15/15; re-verified in Chromium (en/de/pt-BR) that a widget's `text` and `x` still resolve per language and structural suffixes are ignored.

VTT-AI-Agent (commented):
## TL;DR
This is a **UI/UX review, not a code review.** The translation feature works and reads cleanly when a string fits — but the sample data in this PR exposes the one predictable risk of translating in place: **longer translations overflow the fixed-size widgets the English text was sized for.** The most impactful fix: the Floodplain Frenzy score label clips (`SCORE` → German `PUNKT`, Portuguese `PONT` — the last letters are cut off), so either widen that label or add a `width:`/`font-size:` language override for it. I loaded four of the ten games at port 8336 and switched the UI to `de`/`pt-BR` to verify.

---

### Big layout improvements

**Plan for text expansion — it's the feature's main visual hazard.** Translating in place means every button/label keeps the width, height and font-size that were tuned for English, but German runs ~30% longer and wrapped Portuguese phrases longer still. Nothing in the diff (or the sample data) accounts for that, and it already shows up on screen. The good news is the feature *itself* is the fix: because any property can be suffixed, an author can add `width:de-DE`, `height:de-DE` or `font-size:de-DE` alongside `text:de-DE` for any widget whose translation doesn't fit. It would be worth (a) doing this for the overflowing widgets in the ten showcase games so they ship clean, and (b) calling this pattern out in the PR/docs so future translators know to check fit, not just wording. This isn't a redesign — it's using the mechanism you already built to keep each localized layout as tidy as the English one.

German Floodplain, rendered — translations mostly fit, but see the score column and the two action buttons:
German Floodplain

### Nitpicks (small breakages)

- **Score label is clipped in both languages.** `Floodplain Frenzy/0.json` widgets `label-score-p1..p4` are `width: 65`, `font-size: 20px`. English `SCORE` fits; `text:de-DE` `PUNKTE` renders as **`PUNKT`** and `text:pt-BR` `PONTOS` renders as **`PONT`** — the final glyph is cut off, which reads like a typo to a native speaker. Zoomed crops (PT then DE):
  PT score clip
  DE score clip
  Fix options, in order of preference: widen the base label a few px (it has room), or add `width:de-DE`/`width:pt-BR` (or a smaller `font-size:` suffix) for these four labels.

- **Action buttons wrap unevenly.** Floodplain `End Turn` / `Score Cities` (`button-end-turn`, `button-score-cities`, `width: 110`) are single-line in English but each wrap to two lines once translated — `Zug beenden` / `Städte werten`, `Encerrar turno` / `Pontuar cidades`. The result is a taller, more cramped button pair that no longer matches the English rhythm. A slightly wider button or a `width:` suffix would keep them single-line; if wrapping is acceptable, at least confirm the two buttons stay equal height.

### Readability & understandability (first-time user)

- **Half-translated panels read as broken.** In Duel 52 with German selected, `Nachziehstapel`, `Ablage`, `Deine Hand` and `Kartenvorschau` are translated, but every sibling control in the *same* left rail stays English: `Recall & Shuffle`, `Get Starting Hand`, `Set Bases`, `Discard 10`, `Draw a card`, `Marker`, `Damaged`, `Frozen`, `CARD POWERS`, `See full rules at…`. To a German first-timer a panel that's ~40% translated looks like a bug, not a language choice — arguably worse than leaving it fully English.
  Duel 52 German — mixed translation
  Since these ten games are meant as the showcase for the feature, I'd translate the *complete visible control set* of each one (or leave a game out until it's fully done), so selecting a language feels deliberate and finished. The engine change is sound; this is purely about the sample data making a good first impression.

- **Descriptions localize well.** The `_meta.info.description` overrides via `localizeMeta` are the strongest part of the UX — a first-time visitor browsing the library in German/Portuguese now gets a native-language game summary, which is exactly where translation matters most. No layout concern there since descriptions live in a flexible text area.

Overall: the feature is a genuine UX win and the in-place approach keeps the interface untouched when strings fit. The only thing standing between it and a polished result is text-fit on the fixed-size widgets — worth resolving in the showcase data before it's the first thing non-English users see.

VTT-AI-Agent:
Thanks — good catches, addressed in `f66e7c1f`.

**Clipping + half-translated look.** You're right on both: the per-widget label overrides kept the English widget dimensions (so `SCORE`→`PUNKTE` clipped to `PUNKT`) and only covered a subset of each game's controls, which reads as broken. Taking your "fully translate a game's visible controls or none" guidance, I **reverted all in-game label overrides** and kept every one of the ten games translated only at the level with no layout risk — the `_meta.info.description` shown in the library (your noted strongest UX win). Five of a Kind had no English description, so it now gets one plus its `de-DE`/`pt-BR` translations, so all ten are consistent.

**Layout-property translation.** Rather than partially retrofit real games, the widget-level translation — including the layout adaptation you called out — is now shown in the self-contained demo game, where the layout is verifiable exactly. The lower box carries `width:de-DE` / `width:pt-BR` so its longer translation still fits its widget instead of clipping:

| English | German | Brazilian Portuguese |
|---|---|---|
| en | de | pt |

Download / import: [demo.vtt](https://agent.virtualtabletop.io/reports/pr/1529115420605288538/demo.vtt). This is exactly the escape hatch you mentioned — an author who translates a widget's `text` can pair it with `width:<lang>` / a `css:<lang>` font-size so translations that run longer than English still fit.

`npm test` 15/15; re-verified in Chromium that the demo's `text`, `x` and `width` all resolve per language.

VTT-AI-Agent (commented):
## TL;DR

The latest commit removes the showcase layout regressions, but the general widget-property mechanism still allows a player's UI language to select game logic and therefore produce different shared-state deltas. I also found a reproducible validator false positive for language overrides of custom properties.

## Findings

### 1. Language overrides can still select routines and other state-changing behavior (`client/js/statemanaged.js:71`)

The new structural deny-list only excludes five link/identity properties; every other property is returned through `this.languageOverrides` in `get()`. That includes `clickRoutine`, `changeRoutine`, `movable`, `dropTarget`, and custom values consumed by routines. For example, the validator accepts `clickRoutine:de-DE`, and a German client clicking the widget will execute that routine while an English client executes the base routine. Those clients can consequently emit different deltas from the same room state. Similarly, a translated custom score/coordinate used by `GET`, `SELECT`, or `${PROPERTY ...}` changes game logic based on local language.

The comment telling authors to use overrides only for presentation does not enforce the contract, while the PR still advertises and validates the suffix convention for arbitrary properties. Please resolve overrides at a presentation/render boundary, give routine/engine reads an unsuffixed path, or enforce a display-only allow-list; excluding only `id`, `type`, `parent`, `deck`, and `cardType` does not make multiplayer behavior deterministic.

### 2. Translated custom properties are rejected by the validator (`validator/validate_gamefile.js:1196`)

`baseProp` is correctly derived, but the custom-property fallback still checks `customProperties.includes(prop)` using the suffixed key. A game containing a used custom property `foo` plus `foo:de-DE` therefore reports `foo:de-DE: unrecognized (and seemingly unused) property`, even though the feature promises suffixes for any widget property. I reproduced this directly against the PR head with `foo` referenced by a `GET` operation. The fallback should check the resolved `baseProp` (while retaining the existing validation for known properties).

### 3. Suffix updates are stale until a surrounding batch ends (`client/js/statemanaged.js:36`, `client/js/statemanaged.js:124`)

`set()` writes `this.state` immediately, but `this.languageOverrides` is recomputed only from `applyDelta()`. During a batch, `sendPropertyUpdate()` defers `receiveDelta()`/`applyDelta()` until `batchEnd()`. Routine evaluation is batched, so a routine that sets `text:de-DE` (or another suffix) and then reads base `text` later in the same routine sees the old override, unlike ordinary properties whose immediate write is visible to subsequent operations. Recompute the override cache when a suffixed property is set locally as well, or avoid a separate cache whose consistency depends on delta flush timing.

## Verification

The focused validator reproduction confirms finding 2. `npm test -- --runInBand` could not start because this fresh clone has no installed Jest under `node_modules`; the current GitHub `check-library` check is green.

I did not add the `beta` label because resolving overrides in the core `StateManaged.get()`/`applyDelta()` property path genuinely changes widget and routine engine behavior, making the PR engine-ineligible.

VTT-AI-Agent:
Thanks — all three addressed in `bf32fa28`.

**#1 — overrides could select game logic (multiplayer divergence).** Fixed by giving the routine/engine reads an unsuffixed path, per your suggestion. Added `StateManaged.getRaw(property)` (the base shared-state value, no override) and routed every place where a routine reads a property *by name* through it: `${PROPERTY … OF …}`, `GET`, `SELECT` (both the filter relations and the `+`/assignment operations), `SCORE`, `RESET`, and the `SORT` key comparison (`sortWidgets`). Display/UI still uses `get()` with the override, so a widget shows the translated value but a routine computes the canonical one — identical for every player.

I also widened the non-localizable set so properties that *drive* logic/interaction are never localized in the first place: every `…Routine` property (so which routine runs on a click can't depend on language — your `clickRoutine:de-DE` case), plus `inheritFrom`, `owner`, `movable`, `clickable`, `dragging`, `dropTarget`, `activeFace`.

Verified in Chromium: a button whose `clickRoutine` does `var t = ${PROPERTY text OF src}` reads `"Score"` for **both** an English and a German client, even though the German client displays the source label as its `text:de-DE` = `"Punkte"`.

Because existing games contain no language-suffixed properties, `getRaw()` is byte-identical to `get()` for them — no behavior change to any current game.

**#2 — validator rejected a suffixed custom property.** Fixed: the custom-property fallback now checks the resolved `baseProp`, so a `foo` referenced by a routine also covers `foo:de-DE`. Reproduced your case (a `GET` reading `foo` + a `foo:de-DE` key) — no longer flagged.

**#3 — stale override cache during a batch.** Fixed: `set()` now refreshes `this.languageOverrides` immediately when a language-suffixed property is written locally, rather than waiting for the batch-deferred `applyDelta()`.

`npm test` 15/15.

VTT-AI-Agent:
Per your Discord request, the ten most-played games are now **fully translated** in-game (not just their library description) into German and Brazilian Portuguese, with layout fixes so nothing clips or overflows:

> Translate the top 10 games until they are completely translated without layout issues.
>
> — arnoldsmith86 ([Discord](https://discord.com/channels/770758631146782780/1529115420605288538/1529146438980997261))

**What was translated** (`text:de-DE` / `text:pt-BR` on every human-readable label, button and instruction): Bhukhar, Word Whirl, Duel 52, Card Games, Five of a Kind, Floodplain Frenzy, Egyptian Epochs, The Team II, Steam Power, The Team.

**Deliberately left as-is** (translating these would be wrong or break things):
- **Proper game names** — Doppelkopf, Skat, the Jass/Tarock variants, Hearts, Spades, Truco, … (Card Games is a menu of ~80 named games).
- **Icon-font ligatures** (e.g. a `remove`/`add` button that renders as a ± glyph in the symbol font).
- **Dynamic / routine-generated values** — scores, counters, player names, and scoreboard cells the game rewrites at runtime (e.g. Card Games' "Round/Totals" scoreboard, "0 points"). These are emitted by routines, so static property translation can't reach them; making them language-aware would require changing each game's routines.
- Internal template/master widgets that never appear to players.
- Seat prompts like "click to sit" come from the UI layer (handled by #3033's dictionary), not the game.

**Layout without clipping.** Where a translation ran longer than its English-sized widget, I added a `css:<lang>` font-size override (down to an 11px readable floor) and, if still needed, `width:<lang>` / `height:<lang>` overrides — a nice real-world exercise of the language-suffixed *layout* properties this PR enables. Each game was rendered in both languages (after web-fonts load) and checked so that **no translated widget overflows its box**.

Screenshots (German):

| Bhukhar | Card Games | Egyptian Epochs |
|---|---|---|
| bhukhar | cardgames | egyptian |

`npm test` 15/15; validator clean for the suffixed properties.

VTT-AI-Agent:
Acting on a request from @arnoldsmith86 (via Discord) to take the showcase data further:

> Translate the top 10 games until they are completely translated without layout issues.

*(Source: https://discord.com/channels/770758631146782780/1529115420605288538/1529146438980997261)*

I've now translated the **visible in-game interface** of all ten top-played games into German (`de-DE`) and Brazilian Portuguese (`pt-BR`) — not just the library descriptions — and fixed the layout so nothing clips.

**Coverage**
- All human-readable `text` on buttons, labels and instructions, plus scoreboard headers (`roundLabel`/`totalsLabel`, including the engine defaults *Round*/*Totals*) and input placeholders.
- Left as-is on purpose: proper game names (Doppelkopf, the Jass variants, Hearts, Skat, …), icon-font ligatures, dynamic values (scores/counters/player names) and card faces.

**Layout — no clipping.** Each game is rendered in both languages; when a translation would overflow its English-sized widget, the font-size is shrunk to an 11px readability floor and, if it still doesn't fit, the widget is widened. I re-verified by rendering every game in `de-DE` and `pt-BR` and asserting **zero** overflowing widgets (measuring each widget's own text, excluding child widgets).

| Bhukhar (de) | Card Games (de) | Egyptian Epochs (de) |
|---|---|---|
| bh | cg | ee |

**One honest gap:** *Card Games* is a 40-in-1 meta-game whose per-sub-game **rules help dialogs** (`hintText`, 59 of them — dense multi-paragraph rules for each variant) are not yet translated; that's a large body of specialized text I'd tackle as a follow-up. Every interactive control, menu entry and scoreboard label in Card Games *is* translated. The other nine games are complete.

`npm test` 15/15; validator clean for all ten games.

VTT-AI-Agent (commented):
## TL;DR
Solid, well-documented approach: read-time, per-client resolution of language-suffixed properties with a clean `get()`/`getRaw()` split keeps shared game logic deterministic, and the validator update is correct. The one thing worth fixing before merge: `Widget.clone()` still reads positions via `get()`, so cloning a translated widget bakes the triggering player's localized `x`/`y` into shared state — a real cross-language non-determinism bug for exactly the layout-override use case this PR promotes.

## Findings

### 1. `clone()` uses `get()` for position → localized value leaks into shared state (correctness / determinism)
`client/js/widgets/widget.js:542-543`:
```js
await cWidget.set('x', (overrideProperties.x !== undefined ? overrideProperties.x : this.get('x')) + xOffset);
await cWidget.set('y', (overrideProperties.y !== undefined ? overrideProperties.y : this.get('y')) + yOffset);
```
`clone()` is reachable from the `CLONE` routine op (`widget.js:1233`). Everywhere else in the routine interpreter you correctly switched logic reads to `getRaw()`, but these two reads were missed. Consequences when a widget carries `x:de`/`x:pt-BR`-style overrides (the demo scenario in the PR):

- `set('x', …)` writes to the **shared** room state, so the clone's *base* `x` becomes whatever the **triggering player's** language resolved to. A German player and an English player running the same `CLONE` produce different shared positions — non-deterministic across clients.
- Worse, it bakes the localized value into the unsuffixed `x`, so the clone shows that one language's position to *every* player, defeating the per-language override on the clone.

(Note `clone()` already copies `this.state` at line 515, which carries the suffixed keys onto the clone; lines 542-543 then overwrite the base with the localized value.) Fix: use `getRaw('x')`/`getRaw('y')` here, consistent with the rest of the conversion.

### 2. `getRaw()` is not language-independent for *inherited* translated properties (determinism hole, edge case)
`getRaw()` → `coerceValue()` → (when the own state value is `undefined`) `getDefaultValue()`, which resolves `inheritFrom` via `widgets.get(id).get(key)` — the **localized** `get()` (`statemanaged.js:51-52`). So a widget that inherits a translated property from its parent and reads it through `getRaw()` in a routine will still get the parent's language-resolved value, breaking the "routine reads see the unsuffixed shared value" guarantee for the inheritance path. Own properties are fine; only inherited translated properties are affected, so this is a narrow case — but since the whole design rests on `getRaw()` being language-neutral, it's worth either routing inheritance through a raw lookup or documenting the limitation.

### 3. Validator/runtime case-sensitivity mismatch (minor)
Runtime matching in `i18n.js` (`languageSuffixScore`) lowercases both sides, so `label:de-de` resolves fine. But `validate_gamefile.js`'s `baseLanguageProperty()` checks membership in `ALLOWED_LANGUAGES` case-sensitively (`'de-DE'`, `'pt-BR'`, …). A game that works at runtime with a lowercased suffix would be flagged as an "unknown property" in edit mode. Consider matching case-insensitively (or normalizing) so the validator and runtime agree.

## Non-issues verified
- `applyDelta()` override recomputation is correct: when a base property changes while an override is active, the union-loop re-forces `deltaForDOM[base] = get(base)` so the override still wins; when a suffixed key is removed the base reverts. The suffixed key itself flowing into `applyDeltaToDOM`/`applyCSS` is harmless (unknown props are ignored).
- The set of properties switched to `getRaw()` (`${PROPERTY}`, `GET`, `SELECT` comparisons, `SCORE`, `RESET`, `SORT`) and the `nonLocalizableProperties` exclusion list cover the shared-logic read paths.
- Ran `node validator/validate_gamefile_node.js` on the translated `Five of a Kind/0.json`: the `description:de-DE`/`pt-BR` fields validate cleanly (remaining warnings are pre-existing and unrelated).

## Beta label
I did **not** add the `beta` label: this PR genuinely modifies the engine — it rewrites the core `StateManaged.get()` read path, adds `getRaw()`, reworks `applyDelta()` state resolution, and reroutes routine-operation reads (`compute`/`SELECT`/`SCORE`/`RESET`/`SORT`) — which is engine behavior (client state sync + routine-operation semantics), not an incidental helper, so it is excluded from beta auto-deployment by design; the clone determinism bug in finding #1 also argues against exposing it to live players until fixed.

VTT-AI-Agent:
Per the maintainer's request:

> Translate the top 10 games until they are completely translated without layout issues.
>
> — arnoldsmith86 ([Discord](https://discord.com/channels/770758631146782780/1529115420605288538/1529146438980997261))

The ten most-played games are now **fully** translated into German (`de-DE`) and Brazilian Portuguese (`pt-BR`) — every visible label, button, instruction, scoreboard header (`roundLabel`/`totalsLabel`), input placeholder and player-aid legend — using the language-suffix mechanism (`text:de-DE`, `css:de-DE`, `roundLabel:pt-BR`, …).

**Left untranslated by design:** the engine's built-in seat prompt (`click to sit`, which is UI translated by #3033), proper game/variant names (Doppelkopf, Schieber Jass, Hearts, Skat, …), brand strings/URLs, dynamic values (scores, counters, player names) and hidden internal helper widgets.

**Layout:** an automated pass renders each game in both languages and, for any translated widget whose text would overflow its English-sized box, adds a `css:<lang>` font-size (down to a readable 11px floor) or grows `width`/`height`. Overflow is measured from each widget's **own** text (ignoring nested child widgets). All ten games were verified in a real browser (Chromium) with **zero** overflowing widgets in both languages.

Samples (German):

| Five of a Kind | Card Games | Bhukhar |
|---|---|---|
| fk | cg | bh |

`npm test` 15/15; validator reports no new issues for the suffixed properties.

VTT-AI-Agent:
Thanks — all three fixed in `d7bebcb3`.

**#1 — `clone()` position leaked the localized value (correctness).** Good catch; those two reads were missed in the routine-op conversion. Switched `widget.js:542-543` to `getRaw('x')`/`getRaw('y')`. Verified end-to-end in Chromium: a widget carrying `x:de-DE` cloned via the `CLONE` op now writes the **canonical** base `x` (100) to shared state for **both** an English and a German client — previously the German client baked in the localized 500. The clone still carries the `x:de-DE` suffix (copied from state), so per-language display on the clone is preserved while the shared base stays deterministic.

**#2 — `getRaw()` not language-neutral for inherited properties (edge case).** Fixed properly rather than documented: `coerceValue()`/`getDefaultValue()` now take a `raw` flag, and on the raw path inheritance is resolved through `getRaw()` on the parent (recursively), so a routine reading an inherited translated property via `getRaw()` gets the unsuffixed value. `get()` (localized display) and `set()` behaviour are unchanged.

**#3 — validator case-sensitivity.** `baseLanguageProperty()` now matches `ALLOWED_LANGUAGES` case-insensitively, agreeing with the case-insensitive runtime matcher — `label:de-de` is no longer flagged in edit mode. Verified with a game file using `text:de-de` / `width:PT-br`.

`npm test` 15/15.

VTT-AI-Agent:
Follow-up on the maintainer's request:

> You didn't translate INPUT texts, meta how to play sections, attribution texts etc.. Also add the translated languages to the language meta field so the library filters pick up the games.
>
> — arnoldsmith86 ([Discord](https://discord.com/channels/770758631146782780/1529115420605288538/1529164397472124978))

All four are now done for the ten games (German `de-DE` + Brazilian Portuguese `pt-BR`):

- **INPUT dialogs** — added engine support so an INPUT overlay's header, confirm/cancel button texts and each field's label/text/option are resolved per-player at render time (`localizeInputOverlay` in `widget.js`), then translated every game's INPUT dialogs (Card Games alone has ~100: settings, confirmations, deck options, score-transfer prompts). Verified in a browser (e.g. Egyptian Epochs' *New Game* dialog shows the German header, field text and Ja/Nein buttons).
- **How-to-play** — full `ruleText` / `helpText` translations for every game that has them (Word Whirl, Floodplain Frenzy, Egyptian Epochs, Steam Power, Card Games), preserving all inline game-icon `` tags, HTML markup and the translated in-game button names.
- **Attribution** — every game's `_meta.info.attribution`, translating the prose while keeping names, URLs, license IDs, emails and markup verbatim.
- **`language` meta field** — added `de-DE, pt-BR` to every game's `_meta.info.language` so the library language filters list them for German / Portuguese players.

The suffixed meta fields (`ruleText:de-DE`, `attribution:pt-BR`, …) are resolved by the existing `localizeMeta` path; INPUT-op suffixes (`header:de-DE`, …) are accepted by the validator. `npm test` 15/15; no new validator findings.

VTT-AI-Agent:
Per the maintainer's request:

> You didn't translate INPUT texts, meta how to play sections, attribution texts etc.. Also add the translated languages to the language meta field so the library filters pick up the games.
>
> — arnoldsmith86 ([Discord](https://discord.com/channels/770758631146782780/1529115420605288538/1529164397472124978))

All of these are now done across the ten games:

- **INPUT dialogs** — the engine now resolves language-suffixed texts of an INPUT overlay for the player who sees it (header, confirm/cancel buttons, each field's `label`/`text`/option texts — `localizeInputOverlay` in `widget.js`). Every game's INPUT dialogs are translated (Card Games alone has ~100 strings: settings, confirmations, deck options, score-transfer explanations). Verified in-browser: Egyptian Epochs' *New Game* dialog shows the German header, field text and Ja/Nein buttons.
- **Meta "how to play"** — `ruleText` / `helpText` translated for every game that has them (Word Whirl, Floodplain Frenzy, Egyptian Epochs, Steam Power, and the Card Games multi-game reference manual), preserving all HTML, embedded game-icon `` tags and emojis. Verified: opening a game's details in German shows the translated rules.
- **Attribution** — every game's `_meta.info.attribution` is translated (prose translated; names, URLs, license IDs, emails and HTML kept verbatim).
- **`language` meta field** — `de-DE` and `pt-BR` added to each game's `_meta.info.language`, so the library language filters now list these games for German / Brazilian Portuguese players.

The validator was extended to accept language suffixes on routine-operation properties (e.g. `header:de-DE`) as well. Left untranslated by design: proper game/brand/variant names, URLs, license identifiers, dynamic values and the engine's built-in "click to sit" seat prompt.

`npm test` 15/15; validator reports no new issues.

VTT-AI-Agent:
Per the maintainer's request:

> At least Duel 52 has text on the cards. Redo the cards so the text is text and can be translated. They should look basically the same as before.
>
> — arnoldsmith86 ([Discord](https://discord.com/channels/770758631146782780/1529115420605288538/1529181106904104970))

Duel 52's ability cards had the name (VIEW, FREEZE, TWINSTRIKE, …) baked into each card SVG as outlined paths, so it couldn't be translated. Redone so the name is real text:

- **Engine:** `Deck.cardPropertyGet()` now resolves language-suffixed card-type properties (e.g. `cardName:de-DE`) for the selected UI language, so card faces translate (same mechanism as the widget-property / meta suffixes).
- **Duel 52:** stripped the baked name text from all 13 card images (re-saved under their new content-addressed filenames), added a translatable `cardName` per card type (`de-DE` / `pt-BR`), and render it via a text object in the front-face template, positioned in the coloured name band.

The cards look the same as before — only the name is now real, translatable text. Verified in a browser:

| English | German | Brazilian Portuguese |
|---|---|---|
| en | de | pt |

`npm test` 15/15; validator clean; all asset filenames match their content.

PR #3034, opened 2026-07-17
Routine editor on the expanded widget editor (copy of #2439 with folding/tooltip improvements)

VTT-AI-Agent:
To Do / Notes:

---------------------

This is a copy of #2439 (the visual routine editor), brought onto the expanded widget editor (#2884 / #3035, both merged into `main` meanwhile), plus the requested usability changes:

- **Collapse arrows instead of the `…` chip**: the `…` button that revealed default-valued parameters is gone. Each operation now has an arrow at its **start** (like the event cards) that collapses/expands it. Collapsing hides the parameters that just use their defaults and, for IF/FOREACH, the nested routines — so the separate fold toggle is gone too. Operations with nothing to collapse get no arrow.
- **Functions are expanded by default**: all parameters (including ones at their defaults) are visible when an operation is shown.
- **Routines remain collapsed by default**: the event cards (`clickRoutine`, `gameStartRoutine`, …) still start collapsed.
- **Tooltips can now be closed**: every popup — including the info tooltips inside function parameter popups and tooltips opened from inside other tooltips — gets an explicit ✕ close button. Escape now closes only the top-most popup (one level at a time), and the popup swallows the Escape `keyup` so the editor no longer reacts to it with a synthetic click that used to dismiss *all* popups and close the sidebar module at once. Clicking inside an info popup no longer closes it (so its text can be read/selected); the close button, Escape and clicking outside do.
- **The parameter popup leads with the value**: every parameter popup starts with the current value as an editable text input, with the `use default` button next to it; the type-specific inputs that used to be buried in a "Value" section are gone. The widget-property section is now a `Property <name> of <widget>` builder - the name field suggests the properties of the chosen widget (like the CSS editor suggests selectors), the widget field shows a grey `this widget` placeholder when empty and has the shared widget picker next to it.
- **The expanded view carries the wiki information**: an info button next to the operation (the one the operation chip's popup has) and next to every parameter name; a parameter shows the line its operation's text describes it with, plus a link to the whole text. Seven parameters the wiki text never mentioned (`CALL collection`, `CANVAS canvas`/`count`, `INPUT randomRotation`, `MOVEXY z`, `SORT locales`/`options`) got a line, each checked against the engine.
[...]

PR #3024, opened 2026-07-11
Add a fullscreen deck editor to the editor

VTT-AI-Agent:
To Do / Notes

----------------------
For later

- [ ] Support svgReplaces as it works on decks (different than it does on other widgets)

## What

This adds a **visual deck editor** to the room editor, making decks editable visually instead of through raw `faceTemplates`/`cardTypes` JSON. It covers the playing surface while the room editor's top toolbar and right sidebar (including the sidebar modules and their resizable divider) stay visible and functional:

- **Big card view (left):** shows the selected card type at the selected face. Face objects can be selected by clicking, moved by dragging them directly, and manipulated with a **drag toolbar** (Move / Resize / Resize keeping aspect ratio / Rotate) that works like the one in the room editor, including live feedback while dragging. A **"Show areas"** toggle outlines all objects of the shown face.
- **Sidebar (right):** organized into four tabs, one per property scope, so only one kind of property is on screen at a time: **All Cards** (the deck's card defaults, width/height always offered), **Card Type** (rename the selected card type and edit its own properties), **Face** (border/radius/enlarge of the shown face) and **Object** (the selected face object plus its dynamic-property bindings, with upload for images). The tabs follow the selection, and a scope that is not editable yet is greyed out. Color carries the meaning throughout: blue = shared by every card of the deck, amber = differs per card type (the same language as the left toolbar's *Add to: All Cards / Card Type* group), and only clickable things are filled solid blue.
- **Bottom strip:** one preview per card type, plus **"Add blank card type"** and **"Copy card type"** tiles.
- **Top bar:** dropdowns to switch the displayed **face** and **deck**, buttons to add/delete faces and to add text/image/icon/solid-color objects — either the **same on every card** or **different per card type** (one click creates the object already bound to a fresh card type property).

Opened and closed with a **"Deck editor" toggle button on the room toolbar** (in its own group after Group, before Tutorials), which opens the most relevant deck (a selected deck, a selected card's deck, or the last deck in the game). It can also be opened via **"Open deck editor"** buttons in the properties module of a deck and of a card, or — with nothing selected — via a third "Add deck" option that creates a new deck and opens it right away; the toolbar button's pressed state stays in sync with all of these. Escape deselects/closes (and is contained to the deck editor), Delete removes the selected face object including per-card-type data only it referenced.

## How

- New `client/js/editor/deckeditor.js` + `client/css/editor/deckeditor.css`, markup in `client/editor.html`, wired into `server/minify.mjs`.
[...]

PR #147, opened 2021-01-22
piles are no longer widgets but simply UI elements

ArnoldSmith86:
I already posted an issue about this earlier. Having piles as widgets seemed like a good idea at the time but leads to a lot of headaches.

This PR removes piles from the widget hierarchy and instead puts a property `pile` with the same value to each widget within a pile. The client then simply displays a UI element on top of the pile that looks and feels like the old pile handle.

This introduces these new widget properties (and the default values):

```
      allowPiles: true, // set to false on a holder would prevent piles inside the holder
      pile: null,       // the reference to the pile
      pilesWith: null,  // cards have this set to { type: 'card' } and it can be used to make other things pileable
      pileOptions: {    // also "css" to customize the handle
        xOffset: 10,
        yOffset: 10
      }
```

I didn't really test those properties yet though because the main code still needs work.

I'll add my generic test room to the PR test server. It's fun to play around with and might be a good tutorial/demo somewhere?

Stuff I'm still aware of:
- [x] Dropping a pile onto a pile doesn't move all widgets to the other pile's location.
- [x] The _add widgets_ overlay is disabled because its pile handle messes things up.
- [x] Pile handles don't move if its widgets (or their parents) move for any reason.
- [x] The handle is currently _always_ at the top left.
- [x] The handle doesn't scale down yet for very small widgets.
- [ ] Changing `pileOptions` (or anything else affecting the creation of piles) doesn't take effect immediately.
- [ ] Dropping a pile into a hand doesn't remove the pile handle when the cards get spread out.
- [ ] Adding a new card deck makes the shuffle button (and deck) go to 0,0 in the holder (counters are also broken).
- [ ] Cards with different `pileOptions` in one holder form multiple piles. Probably only direct neighbors (on the Z axis) should form a pile?

PR #3047, opened 2026-07-21
Add a sound picker with bundled Kenney sound effects

VTT-AI-Agent:
## What

Adds an **in-editor sound picker** so game creators can browse and select from a large set of bundled sound effects instead of having to upload their own audio file for every `AUDIO` operation or `clickSound` property, and gives `clickSound` a **row of its own in the properties editor**.

Sound picker overlay

### Changes

- **Bundled audio** — added the **complete set of free audio packs** from [Kenney's Audio category](https://kenney.nl/assets/category:Audio) (all **CC0** / public domain) under `assets/audio/`: Casino Audio, Interface Sounds, UI Audio, Impact Sounds, RPG Audio, Sci-fi Sounds, Digital Audio, Music Jingles, Voiceover Pack and Voiceover Pack (Fighter). That's **744 sounds across 10 categories** (**~8.3 MB**), described by a small `audio.json` manifest plus a `LICENSE.txt`. The sounds are transcoded from Kenney's Ogg Vorbis originals to **MP3** (LAME VBR ~130 kbps) so they play on every browser including **Safari/iOS** and stay small (~34% smaller than the Ogg source). Preview/montage tracks are omitted; packs with sub-folders (e.g. the Voiceover Male/Female sets) are flattened with a folder-name prefix so every filename is unique.
- **Picker** — `pickAudio()` / `loadAudioPicker()` in `client/js/audio.js` plus a new `#audioPickerOverlay`. It lists the sounds grouped by category, filters via a search box, and lets you **preview** a sound (▶) before selecting it. Search indexes the pack + sound name (so `dice` returns dice sounds, not the whole card/chip category) and keeps matching category headers visible; the preview button toggles play/stop with a now-playing indicator. It closely mirrors the existing symbol/icon picker.
- **JSON editor button** — a *"pick a sound from the sound picker"* command was added to `jeCommands`, shown for the `AUDIO` operation `source` and the `clickSound` property (right next to the existing *upload audio file* button).
- **Properties editor row** — a **Click sound** row in *Basic → Interaction & display*, right next to the *Clickable* switch that decides whether it plays at all. A new `SoundInput` (`client/js/editor/propertyInputs.js`) names the chosen file next to a speaker chip, plays it back (▶/⏹), and expands into a picker offering **Sound library…** (the overlay above), **Upload sound…** and a plain URL/path field, plus the usual *Remove value*. It works across a multi-selection, the collapsed section header names the sound, and `clickSound` is excluded from *Other properties* so it is not offered twice.

  Click sound row
- **Routine editor** — the `source` of an `AUDIO` operation is a **sound parameter** in the visual routine editor (#3034), so its chip in the sentence *"Play the sound …"* opens the same picker instead of only taking typed text — the way `CANVAS color` and the `INPUT` button icons open the color and icon pickers.

  AUDIO source in the routine editor
- **Attribution** — Kenney (CC0) is credited in a new *Sound Effects* section of the About overlay, listing every pack.

### Why

Previously the only way to attach a sound to a button/routine was to upload a file, and `clickSound` could only be reached through the JSON editor at all. Most creators just want a quick, good-quality click / card / dice / UI / impact sound, so shipping a curated, public-domain set behind a picker — and surfacing it where the rest of a widget's click behaviour is configured — removes that friction and keeps games self-contained.

### Notes

- The picker inserts `/i/audio/...` paths, which the gamefile validator's `asset` type (`validator/validate_gamefile.js`) already accepts, so **no validator changes are needed**. Since this reuses the existing `source` / `clickSound` properties (no new widget/routine property), no other validator/`jeCommands` upkeep is required beyond the new picker button itself.
- All sounds are Kenney originals released under **CC0** (public domain); crediting is appreciated but not required.
- Merged with `main`, including the deck editor: like the symbol picker, the sound picker is moved into `#editor` and floated above the full-screen card view so it is usable from the JSON editor while the deck editor is open.

### Testing

- `npm test` (jest) passes — including new unit tests for the sound-name helper and the collapsed *Interaction & display* summary.
- `npx testcafe chromium:headless tests/testcafe/editor.js` passes (31 tests), with a new assertion that the Click sound row writes, previews and clears `clickSound`.
- Verified end-to-end in a real browser (Playwright): the overlay opens with **744 entries across 10 categories**, the search box filters correctly, preview playback works, selecting an entry returns the correct `/i/audio/...` path, and the properties row sets/clears the value for a single widget and for a multi-selection — no console errors. The minified room + editor bundles build cleanly.
- A playable [`demo.vtt`](https://agent.virtualtabletop.io/reports/pr/1529147136237900057/demo.vtt) with buttons that play sounds from several of the bundled packs (interface/UI click, card shuffle, dice throw, sci-fi & digital laser, RPG book flip, impact footstep, music jingle, a voiceover line and a `clickSound` chip stack) is attached for review.

### Tutorial

The **Functions - AUDIO** tutorial now points readers at this picker (moved here from #3099, which was closed in favour of keeping the docs with the feature):

- **`clickSound` variant** — a new speaker widget with an empty `clickSound` stays deliberately silent; its caption walks the reader through opening it in the JSON Editor, using the *pick a sound from the sound picker* button, previewing with ▶ and clicking a sound to insert its `/i/audio/...` path.
- **`AUDIO` variant** — a paragraph explains that the same picker is available on the `source` line of an `AUDIO` function, pointing at the existing *Deal* button to try it on.

Additions only: two new widgets in `0.json`, one in `1.json`; no existing widget, id, position or text changed, and no other tutorial touched. Both files validate clean (`validate_gamefile_node.js` exits 0) and both variants were walked end-to-end in a real browser.


VTT-AI-Agent:
Good catch — the first pass only bundled 2 of Kenney's free audio packs.

> double check if all free bundles were added. it seems only 2 were added

*(feedback from Discord — [source](https://discord.com/channels/770758631146782780/1529147136237900057/1529154015726796853))*

Fixed in 6e4601e0: the picker now includes **all 10 free packs** from Kenney's [Audio category](https://kenney.nl/assets/category:Audio) — Casino Audio, Interface Sounds, UI Audio, Impact Sounds, RPG Audio, Sci-fi Sounds, Digital Audio, Music Jingles, Voiceover Pack and Voiceover Pack (Fighter) — for **744 sounds across 10 categories** (all CC0). Preview/montage tracks were dropped and the Voiceover Male/Female sub-folders were flattened with a prefix to avoid filename collisions. The About-page attribution and `assets/audio/LICENSE.txt` now list every pack.

VTT-AI-Agent:
### CodeQL check — pre-existing alerts, not introduced by this PR

The CodeQL check reports 4 "new" alerts, but all four are in `client/js/widgets/widget.js` (lines 1047, 1062, 1072, 1461 — `eval`, dynamic method dispatch, prototype-pollution), which **this PR does not modify at all** (`git diff main...HEAD -- client/js/widgets/widget.js` is empty). They already exist on `main`.

They were surfaced because this PR's diff is large — it adds ~590 binary `.ogg` files — and CodeQL's own summary notes: *"Alerts not introduced by this pull request might have been detected because the code changes were too large."*

This PR's actual code changes are limited to `client/js/audio.js`, `client/js/jsonedit.js`, `client/js/main.js`, `client/room.html` and `client/css/fonts.css`, and introduced **no** code-scanning alerts. So no fix is applied here — addressing the long-standing `eval`/dispatch findings in `widget.js` would be a separate, engine-scoped change.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-scoped additive feature — an in-editor sound picker over a bundled CC0 Kenney set — that closely and correctly mirrors the existing symbol picker, with no engine changes. Main thing worth confirming before players rely on it: every bundled clip is **Ogg Vorbis only**, which Safari/iOS cannot decode via either `<audio>` or `decodeAudioData`, so on those browsers both preview *and* actual playback are silent.

## Findings

### 1. Ogg-only assets are silent on Safari / iOS (medium)
The picker offers exclusively `.ogg` files and both code paths that play them rely on formats Safari does not support for Ogg Vorbis:
- Preview: `audioPickerPreview = new Audio(...)` (`client/js/audio.js:66`) — the `<audio>` element does not play Ogg Vorbis on Safari/iOS.
- In-game playback: `addAudio()` → `decodeAudioData()` (`client/js/audio.js:103`) — Web Audio decode of Ogg Vorbis is likewise unsupported on Safari/iOS.

The net effect is that a game whose creator picks a bundled sound will be silently unplayable for Safari/iOS players (no error surfaced to them; `decodeAudioData` failure is only `console.error`'d). This is arguably fine for a first cut — uploaded MP3s still work and this doesn't regress anything — but it's the single biggest caveat and worth a note in the UI or docs, or shipping an alternate encoding. Please at least verify current Safari behavior before considering this "done". Not a blocker for beta.

### 2. Repo weight — the *complete* free Kenney catalog (low, maintainer call)
This adds 744 files / ~15 MB, including the full Voiceover and Voiceover (Fighter) packs (138 spoken lines). That's a permanent git-history addition. The click/card/dice/UI/impact packs are the obvious wins for a tabletop tool; the two voiceover packs are more niche and account for a good share of the entries. Worth a maintainer decision on whether to ship the entire category or a curated subset — purely a judgment call, the code handles any size fine.

### 3. Minor: overlay dismissal leaves the promise pending / preview playing (low)
If the overlay is closed by a path other than the close button or an entry click (e.g. Escape / `showOverlay(null)` triggered elsewhere), the `pickAudio()` promise never resolves and a running preview keeps playing until it ends. This mirrors the same latent behavior in `pickSymbol()`, so it's consistent with existing code rather than a new defect — flagging only for awareness.

## Things I verified (no action needed)
- **Manifest/asset integrity:** all 744 entries in `audio.json` resolve to an existing `.ogg`, and there are no orphan files on disk not listed in the manifest. Category counts and directory names match `LICENSE.txt`.
- **Serving path:** `/i` → `assets/` (`server.mjs:133`), so `i/audio/audio.json` and the inserted `/i/audio/<dir>/<name>.ogg` paths resolve correctly.
- **jeCommands:** the new `je_audioPicker` command reuses the exact `context` regex and `jeInsert(null, jeGetLastKey(), a)` pattern of the adjacent `je_uploadAudio` command, so it appears in the right places (`AUDIO ↦ source`, `clickSound`) and inserts consistently.
- **No validator/jeCommands property upkeep required:** the PR reuses the existing `source`/`clickSound` properties and the validator's `asset` type already accepts `/i/audio/...`, so the "no validator changes needed" claim checks out.
- **Globals plumbing:** `pickAudio` is correctly added to the `Object.assign(window, …)` block in `loadEditMode()` (`client/js/main.js:507`) before `edit.js` is imported, matching how `pickSymbol` is exposed to the editor bundle; `audio.js` and `main.js` are concatenated in `server/minify.mjs`, so the `export`/cross-file reference works as with the existing pickers.
- Preview button correctly `stopPropagation()`s so previewing doesn't also select.

## Engine assessment
This PR touches **none** of the engine files (`client/js/widgets/`, `client/js/compute.js`, `client/js/serverstate.js`). It is purely additive editor tooling (`audio.js` picker, `jsonedit.js` button), bundled assets, an About-overlay credit, and a one-line addition to the window-globals map in `main.js`. It does not change how existing games run — no state-sync, default-widget-behavior, routine-op, or expression-op changes. It works, is safe (static assets, no server-side or security-sensitive changes, no new crashes), and is useful for players to try.

I added the `beta` label: the change is not an engine change and is safe and useful for beta testing.

VTT-AI-Agent:
Thanks for the thorough review! Responses to the three findings:

**1. Ogg-only → Safari/iOS playback.** Valid caveat, but not something this PR regresses or can cleanly fix here:
- Kenney ships these packs as **`.ogg` only** — there are no `.wav`/`.mp3` variants in the source downloads to bundle instead.
- The picker plays through the **same path VTT already uses** for every sound: the `AUDIO` operation and `clickSound` decode via `audioContext.decodeAudioData` (`client/js/audio.js`), and the existing *upload audio file* button already accepts arbitrary audio. So any Safari/Ogg limitation applies identically to sounds people upload today — this PR doesn't add a new constraint, it just makes the existing pipeline easier to use.
- Re-encoding to MP3/AAC isn't feasible in this environment (no `ffmpeg`, and I don't want to add a heavyweight encode dependency or ship an unverifiable conversion). If broader format coverage is wanted, that's best done as a separate, deliberate step — happy to follow up.

**2. 744 files / ~15 MB (full free catalog, incl. both voiceover packs).** This was a deliberate response to explicit reviewer feedback asking for *all* free Kenney audio bundles rather than a subset. I've kept the complete set for that reason, but it's an easy maintainer call — if you'd prefer a curated subset (e.g. drop the voiceover packs, keep the tabletop-relevant SFX), say the word and I'll trim it in a follow-up commit.

**3. Overlay dismissed by a non-button path leaves the promise pending / preview playing.** Correct, and as noted this exactly mirrors the existing `pickSymbol` pattern (close button / selection resolve; other dismissals don't). I kept it consistent with `pickSymbol` intentionally rather than diverging; the preview is a short one-shot clip and `stopAudioPickerPreview()` is already called on close, select, and re-preview. Glad to harden both pickers together if you'd like it addressed as a shared cleanup.

No code changes pushed for this round since none of the three are defects in the change itself. Let me know on (2)/(3) and I'll act.

VTT-AI-Agent (commented):
## TL;DR
Solid, consistent addition — the sound picker faithfully mirrors the existing icon/symbol picker and looks clean and at home in the app. The single most impactful improvement is fixing the **search behavior**: because every entry's keywords include its full category label, typing a term that appears in a category name (e.g. `dice`) floods the results with the *entire* "Cards, Chips & Dice" category and hides the group headers, so the actual dice sounds end up buried in a flat, unlabeled list.

This is a **UI/UX review** (layout, visual consistency, first-time-user clarity), not a code-correctness review. I inspected the changed markup/CSS (`client/room.html`, `client/css/fonts.css`, `client/js/audio.js`, `client/js/jsonedit.js`) and rendered the overlay in a real browser against the running server. Screenshots:

- Default view: https://agent.virtualtabletop.io/reports/ui-review/1529171340844597299/audio-picker-top.png
- After searching `dice`: https://agent.virtualtabletop.io/reports/ui-review/1529171340844597299/audio-picker-search.png
- Narrow / mobile width: https://agent.virtualtabletop.io/reports/ui-review/1529171340844597299/audio-picker-mobile.png

---

### Big layout improvements

**1. Search matches on the category label, so category words swamp results (most impactful).**
Each entry's `data-keywords` is `` `${category} ${directory} ${sound}` `` (audio.js). That means the category name — e.g. *"Cards, Chips & Dice"* — is baked into every entry in that group. Searching `dice` therefore returns **all ~60 casino sounds** (`card-fan-1`, `chip-lay-1`, `chips-stack-3`, …), not the dice sounds a first-timer expects. Worse, during a search the `<h2>` category headers are hidden (`toggleClass(title, 'hidden', text)`), so the results become a flat list with no way to tell which pack each sound came from. See the `dice` screenshot — the real `dice-*` entries are at the very bottom under a pile of card/chip results.
Suggestion: either drop the category name from per-entry keywords (keep only `directory` + `sound` for matching, and let the always-present headers cover category browsing), **or** keep the category headers visible while filtering (only hide headers whose group has zero visible matches). That keeps filtered results grouped and attributable.

**2. 744 entries is a long scroll; consider a category jump/filter.**
The flat, category-grouped scroll works, but it's a lot to wade through for someone who just wants "a dice sound." The symbol picker already establishes a pattern of quick toggle chips (Images/Fonts) — an analogous row of category chips (or a jump-to-category control) would let a creator land on the right pack fast. Optional, but it would scale better than scroll + search alone given the fix above.

### Nitpicks

**3. Category headers are hard-coded black, unlike the sibling picker.** `#audioPickerOverlay h2 { color: black }` (fonts.css). The equivalent `#symbolPickerOverlay h2` has **no** color override, so its headers inherit the app blue (`--VTTblue`). Here the headers render pure black while the intro text right above them is blue — visible in the mobile screenshot. Drop `color: black` to match the icon picker and the rest of the overlay.

**4. No responsive rules for the overlay.** The symbol picker has media queries (fonts.css:487–507) that shrink `padding` at `max-width:600px` / `max-height:375px` and reposition the title on very small screens. `#audioPickerOverlay` inherits only the default `.overlay` `padding: 5% 10%`, wasting horizontal space on narrow/landscape-phone widths. Mirror those queries for consistency.

**5. Preview button gives no "playing / stop" feedback.** The ▶ (`icon="play_arrow"`) stays unchanged while audio plays; clicking again just restarts, and there's no indication of *which* entry is currently sounding. With many near-identical names (`click_001`…`click_005`) a brief active state — e.g. swap the playing entry's icon to `stop` / highlight it — would make previewing much clearer.

### Readability & understandability (first-time user)

**6. The intro sentence is doing important work — keep it.** *"Click the arrow to preview a sound, click its name to select it."* is exactly the affordance explanation a newcomer needs, since the two side-by-side blue elements (play button + name) aren't otherwise self-explanatory. Good call. The friendly category names ("Cards, Chips & Dice" instead of the raw `casino` folder) are also much clearer than the source pack names.

**7. Raw filenames are cryptic without the preview.** Names like `back_001`, `bong_001`, `glitch_002`, `maximize_001` (and the numeric UI/interface packs generally) carry no hint of what they sound like — a first-timer can only tell by pressing ▶. That's inherent to the Kenney source naming and the preview button mitigates it well, so nothing to fix in-PR; it's mainly a reason the preview affordance (points 5 & 6) matters so much. If a lightweight relabeling ever felt worth it, a human-friendly display name in `audio.json` would help, but that's out of scope here.

Overall the picker is clean, consistent with the app's visual language, and clearly onboarded. Addressing the search/keyword behavior (#1) and the black-header inconsistency (#3) would be the highest-value polish.

VTT-AI-Agent (commented):
## TL;DR

The picker is additive and beta-safe, and I found no manifest/asset mismatch or engine impact. The main code issue is a cold-load race: a second invocation can enter the picker before its first manifest fetch has initialized the DOM, while a failed fetch leaves the picker unrecoverable until page reload.

## Findings

### 1. Share or retry the manifest load instead of treating `"loading"` as loaded (low)

At `client/js/audio.js:10`, the first call changes `audioPickerData` from `null` to the string `"loading"` and then awaits the fetch. Any concurrent call now skips the entire initialization block and returns immediately. On a cold picker, double-clicking the JSON-editor command (or otherwise invoking it twice before the fetch completes) lets the second `pickAudio()` reach `client/js/audio.js:53`, where `$('#audioPickerOverlay input').onkeyup` has not been assigned yet, causing a `TypeError`. The overlapping `jeCallCommand()` calls can also toggle `jeDeltaIsOurs` out of order while the first picker remains pending.

A fetch/JSON failure has a related persistent failure mode: `audioPickerData` remains `"loading"`, so every later attempt skips the fetch and DOM setup and hits the same missing handler until the page is reloaded. Store and await a shared loading promise (or otherwise serialize initialization), check `response.ok`, and reset the cached state on failure so a later invocation can retry.

## Verification and engine assessment

- All 744 manifest entries map one-to-one to 744 committed `.ogg` files, with no duplicate, missing, or orphaned paths; every audio blob has the expected Ogg header.
- The separate failing CodeQL status annotates `client/js/widgets/widget.js`, whose blob is byte-for-byte identical between `main` and this PR. Those alerts were not introduced here.
- No engine files are touched (`client/js/widgets/`, `client/js/compute.js`, or `client/js/serverstate.js`). The changes are editor-only picker wiring, static assets/CSS/markup, and attribution, so existing games' sync, widget defaults/events, routine semantics, and expressions are unchanged.

The `beta` label was already present; I left it in place because this additive editor/static-asset change does not alter engine behavior, and the failing CodeQL annotations are on unchanged `widget.js` code.

VTT-AI-Agent:
### Converted the bundled sounds to MP3 (lighter + universally supported)

Acting on this feedback:

> can you convert the files to a light format that is widely supported. would be nice if we could reduce the file sizes without loosing to much quality

*(feedback from Discord — [source](https://discord.com/channels/770758631146782780/1529147136237900057/1529163926686662809))*

Done in 2852071a. I transcoded all 744 bundled Kenney sounds from **Ogg Vorbis → MP3** (LAME VBR `-q:a 5`, ~130 kbps):

- **Widely supported:** MP3 plays on every browser including **Safari/iOS**, which addresses the earlier Ogg-decode caveat too.
- **Smaller:** the bundle drops from **~12.5 MB → ~8.3 MB** (~34% smaller) with negligible quality loss for these short SFX. (I benchmarked MP3 VBR vs MP3 CBR 96k vs AAC 96k — MP3 VBR `-q:a 5` was both the smallest and the safest-quality option.)

The picker, manifest, `LICENSE.txt` note and the demo were all updated to the `.mp3` paths.

VTT-AI-Agent:
Thanks for the live UI/UX pass — all of these are addressed in 2852071a:

**1. Search matched the category label baked into every entry (biggest issue).** Fixed. Per-entry keywords now index only the pack directory + sound name, **not** the display category, so typing `dice` returns just the 8 dice sounds instead of the entire 54-item "Cards, Chips & Dice" category. I also kept the header of any category that still has matches **visible while filtering**, so results stay grouped and labelled rather than collapsing into an unlabelled flat list. Verified against the real manifest: `dice` → 8 results (all dice), `laser` → 24 (Sci-fi + Digital), `click` → 12 (Interface/UI/RPG).

**2. Category `<h2>` hard-coded black.** Fixed — removed the explicit `color: black` so the headings inherit the overlay heading colour, exactly like the sibling symbol picker.

**3. No responsive media queries.** Fixed — added the same `max-width/max-height` padding breakpoints the symbol picker uses.

**4. ▶ preview had no now-playing/stop feedback.** Fixed — the playing entry now swaps its ▶ icon for a ⏹ stop icon and underlines its label; clicking it again stops playback (and it auto-clears on `ended`).

**Readability of cryptic Kenney names (`bong_001`, `glitch_002`).** Kept the original Kenney filenames for traceability back to the source packs, but the intro sentence, friendly category names, in-place preview and the now-fixed grouped search should carry the onboarding. Open to adding human labels later if you'd like.

Also in the same push (separate Discord request): the sounds were transcoded Ogg→MP3, so preview + playback now work on Safari/iOS as well. The [`demo.vtt`](https://agent.virtualtabletop.io/reports/pr/1529147136237900057/demo.vtt) is updated if you want to re-test live.

VTT-AI-Agent:
Good catch on the cold-load race. Fixed in 2852071a.

Previously `loadAudioPicker` used a `'loading'` sentinel: the first call set it and awaited the fetch, but a second concurrent call saw the non-null sentinel, skipped the block and returned immediately — so `pickAudio` could wire click handlers over the still-`Loading…` list.

It now **memoizes the load in a promise**: the first call kicks off the fetch+build once, and every caller (including concurrent ones) `await`s that same promise, so they all proceed only after the list is fully built. On failure the promise is cleared so a later open can retry:

```js
export async function loadAudioPicker() {
  if(!audioPickerLoad) {
    audioPickerLoad = (async function() { /* fetch + build list + wire search */ })()
      .catch(e => { audioPickerLoad = null; throw e; }); // allow retry on failure
  }
  await audioPickerLoad;
}
```

(The sibling `loadSymbolPicker` has the same latent pattern; I left it untouched to keep this PR scoped, but happy to apply the same fix there in a follow-up.)

Agreed on the `beta` label staying — the change is editor/asset-only and the CodeQL alerts are all on unchanged `widget.js` engine code.

VTT-AI-Agent:
The CodeQL check is failing for the **same pre-existing reason** as [before](https://github.com/ArnoldSmith86/virtualtabletop/pull/3047#issuecomment-5036339054): the 4 flagged alerts are all in `client/js/widgets/widget.js` (lines 1047/1062/1072/1461), which this PR **does not touch** (`git diff main...HEAD -- client/js/widgets/widget.js` is empty). They exist on `main`.

The MP3 conversion made the diff even larger (744 Ogg deletions + 744 MP3 additions), which is exactly why CodeQL re-surfaces those unchanged engine alerts ("code changes were too large"). This PR's actual code changes are limited to `client/js/audio.js`, `client/css/fonts.css`, `client/js/jsonedit.js`, `client/js/main.js` and `client/room.html`, and introduce no code-scanning alerts. No fix applied — same rationale as before.

VTT-AI-Agent (commented):
## TL;DR
Clean, well-scoped feature that faithfully mirrors the existing symbol/icon picker: an in-editor sound picker plus a bundled CC0 Kenney sound library, wired into the `AUDIO`/`clickSound` JSON-editor buttons. I found no correctness bugs — the code, manifest and asset files all check out. Only minor UX notes below. **Not an engine change** (the `audio.js` additions are additive edit-mode-only functions), so it's beta-eligible.

## Findings

### Verified correct
- **Manifest ↔ files match.** `audio.json` lists 744 sounds across 10 categories; every referenced `assets/audio/<dir>/<name>.mp3` exists (checked all 744, 0 missing), and there are no duplicate filenames within a pack.
- **Serving path is correct.** `router.use('/i', express.static('assets'))` maps `i/audio/audio.json` → `assets/audio/audio.json` and the inserted `/i/audio/…` URLs resolve. The picker inserts the portable `/i/audio/…` path (unmapped), while preview uses `mapAssetURLs(...)` — correct split.
- **Validator/jeCommands upkeep.** No new widget/routine property is introduced; the new `je_audioPicker` command shares the exact context regex of the existing `je_uploadAudio` (`^.*\(AUDIO\) ↦ source|^.* ↦ clickSound`), and `/i/…` asset paths are already accepted by `je_uploadAsset`'s context — so the PR's "no validator changes needed" claim holds.
- **Wiring.** `pickAudio` is exported from `audio.js` and added to the edit-mode scope list in `main.js:507`, matching how `pickSymbol` is threaded into `jsonedit.js`. Syntax-checks clean.
- **Apostrophe filename.** `voiceover_fighter/it's_a_tie` is the one non-`[A-Za-z0-9_-]` name; it renders fine (double-quoted `data-*` attributes + text span) and is a valid URL char — no HTML/attribute breakage.

### Minor (non-blocking)
1. **Preview can keep playing when the overlay is dismissed by any path other than the close button / a selection.** `stopAudioPickerPreview()` is only invoked from the `[icon=close]` handler and from `entry.onclick`. The global Escape handler (`main.js:704`) in edit mode clicks the editor-toolbar close rather than the picker's close button, so pressing Escape mid-preview can hide the picker while audio keeps playing and leaves the `pickAudio` promise unresolved. This is the same lifecycle quirk the symbol picker already has, but an *audible* leak is more noticeable than a stale symbol overlay. Consider stopping preview on `showOverlay(null)`/overlay-hide. Low priority.
2. **Repo size.** ~8.3 MB of binary MP3s are added to the git history permanently. That's a maintainer judgment call (the payoff — self-contained, good-quality CC0 sounds behind a picker — is real), just flagging it explicitly since it's the largest and least-reversible part of the diff.

Nothing here blocks the change.

## Beta label
The engine files (`client/js/widgets/`, `compute.js`, `serverstate.js`) are untouched; `client/js/audio.js` is modified but only with **additive, edit-mode-only** functions (`pickAudio`/`loadAudioPicker`) called solely from the JSON editor — the `AUDIO` operation and `clickSound` runtime semantics and every existing game's default behavior are unchanged, so this is **not** an engine change and is beta-eligible. The change is safe to test (no crash/corruption risk) and useful to players, and the `beta` label is already present, so I left it as-is.

96LawDawg (approved):
Tested. Approved.

VTT-AI-Agent:
Merged the latest `main` into this branch and resolved the conflict.

> update branch and solve conflicts

— [rapha195 on Discord](https://discord.com/channels/770758631146782780/1529147136237900057/1534292523093069903)

**The conflict** was a single line in `client/js/main.js`: main's icon-picker work added `toNotoMonochrome`/`skipForNotoMonochrome` to the edit-mode export list, this branch added `pickAudio` to the same line. Both are kept.

**One integration fix** was needed on top of the merge, for the deck editor that landed on `main` in the meantime (#3024/#3076): it re-parents the shared editor overlays into `#editor` so they show above the full-screen card view, and the JSON editor — the only place the sound picker is opened from — is reachable while it is open. The sound picker now gets the same treatment (`deckeditor.js`), plus a CSS rule floating it over the editor, because `#deckEditor` (z-index 1) covered it otherwise. Its media queries stay media queries for the same reason the symbol picker's do: once re-parented there is no `#roomArea` container left for the container queries #3097 introduced.

**Verified after the merge** (Playwright, real browser):
- plain edit mode: picker opens with 744 entries across 10 categories, search filters correctly, preview plays, selecting returns the `/i/audio/…` path, `/i/audio/casino/card-shuffle.mp3` serves as `audio/mpeg` — [screenshot](https://agent.virtualtabletop.io/reports/pr/1529147136237900057/merged-audio-picker.png)
- with the deck editor open: picker renders above it and is not covered — [screenshot](https://agent.virtualtabletop.io/reports/pr/1529147136237900057/final-deckeditor.png)
- no console errors in either case; `npm test` passes (261 tests).

VTT-AI-Agent:
Added a **Click sound** row to the properties editor.

> add clickSound input to properties UI in the interaction and display subsection

— [rapha195 on Discord](https://discord.com/channels/770758631146782780/1529147136237900057/1534294121240592554)

It sits in *Basic → Interaction & display*, directly below the *Clickable* switch that decides whether the sound plays at all. The new `SoundInput` shows the chosen file name next to a speaker chip and expands into a picker with:

- **Sound library…** — the bundled Kenney picker this PR adds
- **Upload sound…** — the existing asset upload
- a plain URL/path field, and the usual *Remove value*
- ▶ / ⏹ to hear the sound without leaving the sidebar

Click sound row

It also works on a multi-selection (setting it applies to every selected widget — [screenshot](https://agent.virtualtabletop.io/reports/pr/1529147136237900057/clicksound-multi.png)), the collapsed section header names the sound, and `clickSound` is excluded from *Other properties* so it is not offered twice.

One fix came out of testing it: the sound library indexed its keywords with `-`/`_` flattened to spaces but searched the raw text, so typing a name the way it is listed (`dice-throw`) matched nothing. The search text is now flattened the same way.

**Verified:** `npm test` (262 jest tests, incl. new ones for the sound-name helper and the collapsed summary), `tests/testcafe/editor.js` (31 tests, incl. a new assertion that the row writes, previews and clears `clickSound`), and end-to-end in a real browser for single and multi-selection — no console errors.

VTT-AI-Agent:
**CI note — the red CodeQL check is not caused by this PR.**

All 27 other checks pass (Node.js CI, both TestCafe suites incl. `editor.js`, Production Environment). The GitHub Advanced Security check reports 4 alerts, all in `client/js/widgets/widget.js`:

| Alert | Rule | Line | Open on `main` since |
| --- | --- | --- | --- |
| [#145](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/145) | `js/code-injection` | 1091 | 2025-04-08 |
| [#140](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/140) | `js/unvalidated-dynamic-method-call` | 1490 | 2025-04-08 |
| [#146](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/146) | `js/prototype-polluting-assignment` | 1076 | 2025-04-08 |
| [#205](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/205) | `js/prototype-polluting-assignment` | 1101 | 2026-07-29 |

`git diff origin/main -- client/js/widgets/` is empty for this branch — the routine engine is untouched here. CodeQL says so itself in its summary: *"Alerts not introduced by this pull request might have been detected because the code changes were too large."* This PR adds 744 audio files, which is what pushes the diff over that threshold.

No action taken; nothing to fix on this branch.

VTT-AI-Agent:
The failing CodeQL check is not caused by this PR.

All four alerts (`js/code-injection`, `js/unvalidated-dynamic-method-call`, 2× `js/prototype-polluting-assignment`) point at `client/js/widgets/widget.js` lines 1078/1093/1103/1492 — the routine engine's variable assignment, `eval`-based math expressions and `Math[aggregation]` dispatch. This branch does not touch `client/js/widgets/` at all (`git diff main...HEAD -- client/js/widgets/` is empty), and all four alerts already exist on `main` (alerts #145, #146, #147, #140 from 2025-04-08 and #205 from 2026-07-29).

CodeQL flagged them here because of its own caveat in the check summary: *"Alerts not introduced by this pull request might have been detected because the code changes were too large."* This PR adds several hundred bundled `.mp3` files, which blows past that threshold. No action needed on this branch.

VTT-AI-Agent:
The **sound of an `AUDIO` operation can now be picked in the visual routine editor**, not only in the JSON editor.

> Update and add the input in AUDIO function in the

— [rapha195 on Discord](https://discord.com/channels/770758631146782780/1529147136237900057/1534535574852075661)

(The message arrived cut off after "in the" — I read it as the routine editor that landed on `main` with #3034, since that is the one place the sound picker was still missing. Say the word if you meant somewhere else.)

**What changed**

- The branch was merged with the current `main` (which brings #3034's routine editor), conflict-free.
- `AUDIO`'s `source` is a **`sound` parameter** now instead of a plain text one, so its chip in the sentence *"Play the sound …"* opens the same picker the **Click sound** row of the properties editor uses: **Sound library…** (the bundled Kenney sounds), **Upload sound…**, a typed URL/path, ▶ playback and *Remove value*. This mirrors how `CANVAS color` and `INPUT confirmButtonIcon`/`cancelButtonIcon` already open the color and icon pickers (`RoutineSoundPopup extends RoutinePickerPopup`, reusing `SoundInput`).
- The blank the sentence shows while nothing is set now reads **sound file** rather than *file name*.
- One integration fix the reuse needed: the sound library is an overlay of its own, so the click that picks a sound in it lands *outside* the parameter popup — which used to be read as "click outside" and closed the popup before the picked path was handed over. The popup ignores clicks inside `#audioPickerOverlay`, and stops a running preview when it closes.

| The parameter popup | The library on top of it | The picked sound |
| --- | --- | --- |
| popup | library | picked |

**Verified**

- `npm test` — 569 jest tests pass, including three new ones: the `source` chip opens `RoutineSoundPopup`, the popup applies the path when it closes, and a click in the sound library does not dismiss it.
- `npx testcafe chromium:headless tests/testcafe/editor.js` — 31 tests pass after the merge.
- End-to-end in a real browser: opened a button's `clickRoutine` with an `AUDIO` operation, clicked the *sound file* chip, opened the library from the popup, picked a sound and got `{"func":"AUDIO","source":"/i/audio/casino/card-fan-1.mp3"}` written back into the routine — no console errors.

VTT-AI-Agent:
Fixed in b53e84b2 — the sound popup now keeps out of the play area.

> The audio input should not overlap the play area

— [rapha195 on Discord](https://discord.com/channels/770758631146782780/1529147136237900057/1534542146097381546)

**What was wrong:** the sound library is an overlay, and overlays are children of `#roomArea` (#3097), so the list opens *on the board*. The parameter popup, on the other hand, was placed anywhere the editor has space — which on a wide screen happens to be next to the board, but on a portrait/narrow window is right on top of it. Opening the library from the popup there put the two on top of each other.

**The fix:** `RoutineSoundPopup.avoidsPlayArea()` returns true, so `placementLimits()` gives it the biggest strip the play area leaves over — exactly what the popups that let you pick a widget in the room already do. Nothing else changed: on a wide screen the popup stays where it was (beside the board), and clicks in the room no longer dismiss it while the library is open.

| Before | After |
| --- | --- |
| before | after |

**Verified:** `npm test` — 570 jest tests pass, including a new one next to the existing *popups stay out of the play area* cases. In a real browser at 820×1180 and 1400×900: the popup no longer intersects `#roomArea` in either layout, previewing and picking a sound still writes `/i/audio/casino/card-fan-1.mp3` into the routine, no console errors.

Two related things I left alone, say the word if you want either:

- The **library overlay itself** still covers the board, like every other overlay (symbol picker, states, …). Changing that for this one picker only would make it the odd one out — but it is a one-liner if you'd rather have it float over the editor instead.
- `RoutineIconPopup` has the same latent overlap: it opens the symbol picker, which is also an overlay over the board. It came with #3034 rather than with this PR, so I did not touch it here.

96LawDawg:
@VTT-AI-Agent, if the sound picker is open and I switch to a different widget, the sound picker remains open and any change there is on the previously displayed widget even though the editor for the new widget is open.

VTT-AI-Agent:
Good catch - reproduced and fixed in 5d7d463.

What was going on: the *"AUDIO - which sound"* popup deliberately keeps out of the play area so that the sound library it opens (a full-screen overlay over the board) does not cover it. It did that by reusing `avoidsPlayArea()`, which is also what makes the widget-picker popups treat clicks in the room as **their input** rather than as a click outside them - so the sound popup swallowed those clicks too. Clicking another widget therefore selected it and rebuilt the sidebar for it, while the popup stayed up and kept writing to the widget that was shown before:

before

Placement and input are now two separate questions (`usesRoomAsInput()`): only the popups that actually take widgets from the room keep clicks in there, so the sound popup is dismissed by a click on a widget like by any click outside it - the editor of the widget you clicked comes up with no leftover popup:

after

The sound library overlay itself had the same problem one level up: it does not cover the sidebar, and a widget can be selected without clicking in the room at all (the JSON editor's tree and its back/forward buttons, a widget link in the properties sidebar) - a sound picked in it would then land on the widget that is no longer on screen. So the editor now cancels an open sound picker when the selection changes to another widget: the overlay closes and nothing is written.

Verified in a real browser for all of it - the routine popup (closes on a room click, and picking a sound normally still writes it), the *Click sound* row in the sidebar, the library cancelled by navigating to another widget in the JSON editor, and that the widget pickers still collect widgets by clicking them in the room. `npm test` (575 tests, including new ones for the cancelled picker and for the room click) and `npx testcafe chromium:headless tests/testcafe/editor.js` (31 tests) pass.

PR #3051, opened 2026-07-25
Deep TestCafe coverage for the expanded routine editor

VTT-AI-Agent:
## ⚠️ Depends on #3034

This PR is built on top of #3034 ("Routine editor on the expanded widget editor") and **must not be merged before it**. It targets the `routine-editor-expanded-editor` branch, not `main`, since the visual routine editor only exists there.

## Summary

A human tester reported that the routine editor has too many possible operations/parameter types to manually click through, and that manual testing missed real bugs last time. This PR adds a TestCafe suite (`tests/testcafe/routineeditor.js`) that drives the actual widget-properties UI end-to-end - never by hand-writing or pasting routine JSON into the game file - to stress that input surface the way a thorough human tester would, but repeatably.

The suite is written against the editor as #3034 has it now: an operation is a **sentence**, the phrase it starts with is the **way it works** (its variant), the optional parts of the sentence are its **options** (clauses) and every blank in it is a **parameter chip**.

- **`build a clickRoutine entirely through the routine editor UI and verify the JSON`** - opens edit mode, adds a button, opens its properties panel, expands the `click` automation and builds a 7-operation `clickRoutine` purely through clicks/typing:
  - `SET` - another way of working (*Increase*), a property name typed into the value row, and the *a named group of widgets* option filled from the predefined collections
  - a `var` statement - the name of the variable and the value it works out
  - `IF` with a predefined value (`${playerName}`) and nested `thenRoutine`/`elseRoutine` blocks (via the "add else" control)
  - `FOREACH` switched to *For each number in the range*, with a nested `loopRoutine` built via the Range popup
  - `CLICK` - built entirely via the JSON shortcut on the card
  - `SELECT` - a setting drop-down (*is not*) plus the *give this group a name* option
  - `GET` - a widget-property reference (`${PROPERTY text}`, via the "A property of a widget in the room" builder) and the *name the result* option

  The resulting routine is then verified twice: against the room state (what the game actually stores) and by reading it back through the JSON editor (`#editorSidebar [icon=data_object]`), matching the exact intended JSON.

- **`a button automation built purely through the UI actually runs when clicked`** - builds a simple "increase `clicks` by 1" automation through the same UI, exits edit mode and clicks the real widget twice, asserting the room state changes accordingly - proving the automation executes, not just that the UI didn't crash.

- **`routine editor popups: close button, escape closes only the top-most, and "use default" resets a parameter`** - exercises popup close/escape/reset-to-default behavior directly through the rendered editor.

- **`operation buttons move, nest, hoist and delete operations`** - the `↑`/`↓`, the two "move into the block above/below" buttons, "move out of this block" and delete, asserting the whole `clickRoutine` after every single splice. These are the index-arithmetic paths where an off-by-one silently reorders or drops a creator's operation.

- **`the variant menu rewrites the parameters that tell the ways of working apart`** - picking another way for a `SET` to work has to rewrite *all* of the parameters that tell the ways apart at once: *Increase* keeps the value the creator entered, *Toggle* drops it (a leftover number would still be read by the engine), and *Set* is the one without a relation at all.

- **`options are added to and taken back out of the sentence`** - an option is only in the sentence while it is in use: adding *n times* and *ignore something* to a `CLICK` writes the parameters they word, and the ⊖ behind each phrase removes exactly those again.

- **`parameter warnings: a deprecated parameter, a typo and a value the routine works out`** - the orange `!` on the deprecated `CANVAS.canvas`, the red `!` and extra phrase for a custom property the operation does not declare (added through the JSON shortcut, then removed again with its own ⊖), and the `!` behind the phrase a sentence starts with while a `${...}` decides what the operation does - including that the value stays a chip of its own and that every entry of the drop-down says which reference picking it would write over.

- **`parameters the engine reads are not flagged as unsupported`** - `MOVEXY z`, `CALL collection`, `CANVAS count` and `INPUT css`/`randomRotation` are read by the engine; the editor used to offer each of them as a typo to delete.

- **`drag and drop: into a nested block, with a Ctrl+click multi-selection, and back out`** - real HTML5 drag and drop: an operation dropped into an empty `IF` block (which has to attach the block to the operation), a Ctrl+click multi-selection dragged as one block with its order preserved, and an operation dragged back out to the top level. Every drop aims at an "add operation" button - the one spot inside a routine level that is not an operation card - so the expected result never depends on which half of a card the cursor ended up in.

- **`a Ctrl+selection is dropped as soon as the routine changes`** - the multi-selection is index based, so a selection that outlived the operations it pointed at would move a card the creator never picked.

- **`the number popup offers 0 instead of null and widget pickers keep off the play area`** - the number popup dropped its "null" buttons in favour of a `0`, and a picker that asks for a widget to be clicked in the room must not cover the room: on a 520x900 portrait window the popup overlaps `#roomArea` by 82485 px2 without that placement logic.

`tests/testcafe/test-util.js` gained two small, reusable helpers used by the new tests: `getWidgets()`/`waitForWidgets()` for polling room state (routine edits reach the server asynchronously), and `setupTestEnvironment()` now honors a `VTT_TEST_SERVER` env var so the suite can target a non-default port locally without touching the file (CI is unaffected - it still defaults to `localhost:8272`).

`routineeditor.js` is also added to the `testfile` matrix of both TestCafe workflows - since #3005 they run an explicit list of files instead of the whole directory, so without an entry this suite would never run in CI.

## Findings from the deep-test passes

The suite is the repeatable part; the exploratory passes around it are reported on the PR they belong to:

- First pass: no bugs in the routine editor - every path exercised behaved as designed.
- Second pass: **two findings, reported in [this review on #3034](https://github.com/ArnoldSmith86/virtualtabletop/pull/3034#pullrequestreview-4785477205)** - engine-supported properties (`INPUT.css`, `CANVAS.count`, `CALL.collection`, `MOVEXY.z`, `INPUT.randomRotation`) shown as unsupported typos with a one-click way to delete them, and a Ctrl-selection that keeps its selection but loses its outline when the operation is expanded, so a later drag moves an extra operation.
- Third pass (after `8096c7ed`): both earlier findings fixed, plus **one new finding, reported in [this review on #3034](https://github.com/ArnoldSmith86/virtualtabletop/pull/3034#pullrequestreview-4792205242)** - `FLIP` marked `face` as ignored while `faceCycle` was `random`, although the engine gives `face` priority.

## Test plan

- [x] `npm test` (jest) passes (392/392)
- [x] `npx testcafe chromium:headless tests/testcafe/routineeditor.js` passes - 11/11, run repeatedly, no flakiness observed
- [x] `npx testcafe chromium:headless tests/testcafe/editor.js` still passes - 17/17, no regression

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3051/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR
Solid, well-structured test-only PR: it adds an end-to-end TestCafe suite that drives the new routine editor UI (`tests/testcafe/routineeditor.js`) plus two reusable helpers in `test-util.js`, with no production/engine code touched. It looks correct and I found no blocking issues — only a minor note about the `waitForWidgets` timeout/fallback behavior.

## Findings

**Scope.** Changes are confined to `tests/testcafe/` and target the `routine-editor-expanded-editor` feature branch (it depends on #3034). No production, engine, or validator code is affected, so the `docs/compat.md` / validator / `jeCommands` upkeep rules don't apply here. The suite exercises the editor *through the UI* (clicks/typing) rather than by injecting JSON, which is exactly the right way to catch the class of bug the human tester reported, and it verifies results both against room state and via the JSON editor read-back. Good coverage design.

**`waitForWidgets` fallback is effectively dead and the last sleep is wasted** (`tests/testcafe/test-util.js:60-73`).
The loop `for(let wait=50; wait<1000; wait*=2)` fetches and checks the predicate at the *top* of each iteration and then sleeps. After the final iteration (`wait=800`) it sleeps 800ms and exits without re-fetching, so:
- The final `return predicate(widgets)` re-evaluates the predicate against the *same stale* `widgets` already checked inside the loop. Since a truthy result would have returned early, this fallback path is always falsy — so on timeout callers like `finalWidget.clickRoutine` (`routineeditor.js`) dereference `null`/`false` and blow up with a confusing `Cannot read properties of null` instead of a clear "widget never reached expected state" failure.
- The last 800ms sleep buys nothing because there's no fetch after it; effective useful polling is ~750ms, not the ~1.5s the doubling schedule implies.

This mirrors the existing `compareState` loop shape, so it's consistent with the codebase and not a correctness bug in the happy path — but a `getWidgets()` re-fetch after the loop (or a thrown timeout error) would make failures both more reliable and more legible. Minor.

**Whitespace normalization in `readJsonEditorWidget`** (`routineeditor.js`): the `raw.replace(/ /g, ' ')` swaps a special space the JSON editor renders for a normal one before `JSON.parse`. That's fine and necessary, but it's invisible in the diff — a short comment naming the character (e.g. non-breaking/narrow no-break space) would save the next reader a puzzled moment. Nit.

Everything else — the accordion/popup selectors, the enum-vs-boolean handling (`random: true`), the `withText` regex anchoring, and the dual verification — reads cleanly and matches the described passing runs.

## Beta label
Not added: this is a pure test-infrastructure PR with no player-facing runtime behavior to exercise, and it targets the `routine-editor-expanded-editor` feature branch rather than `main`, so there is nothing for players to test on the beta server and no clean deploy on top of `main` — it is not beta-eligible in any meaningful sense.

VTT-AI-Agent (commented):
## TL;DR
This is a **UI/UX review, not a code review.** PR #3051 is a **test-only change** — it adds a new TestCafe suite (`tests/testcafe/routineeditor.js`) and two helper functions to `tests/testcafe/test-util.js`. The diff touches **no** HTML, CSS/SCSS, templates, or client rendering code, so there is no user-facing interface to evaluate. Nothing to flag on layout, visual design, or readability; no UI changes ship with this PR.

---

I reviewed this PR statically (from the diff) with a UI/UX lens. Both changed files are under `tests/testcafe/`:

- `tests/testcafe/routineeditor.js` (new) — an end-to-end test that *drives* the routine editor UI (added in the dependency PR #3034) via clicks/typing and asserts on the resulting room state and JSON.
- `tests/testcafe/test-util.js` — adds `getWidgets()`/`waitForWidgets()` polling helpers and lets `setupTestEnvironment()` honor a `VTT_TEST_SERVER` env var.

### Big layout improvements
None applicable — no rendered screens, components, or styles are introduced or modified.

### Nitpicks (small layout breakages)
None applicable — no markup, selectors-as-styling, colors, spacing, or alignment are added or changed. The CSS selectors referenced in the test (`.routine-editor`, `.inline-popup`, `.routine-editor-parameter-row`, etc.) are *consumed* by the test to locate existing elements, not defined here, so they belong to PR #3034's surface, not this one.

Not applicable to the shipped artifact — this PR adds no labels, icons, tooltips, empty/loading/error states, or copy that an end user would ever see. It runs only in CI/local test runs.

One incidental observation, purely on the test's own value as a *guard* of the UI (not a change in this diff): because the suite verifies the built routine both against room state and by reading it back through the JSON editor, it does implicitly protect the routine editor's first-time-user affordances it exercises — the "All Operations" vs "Common Actions" accordion, the collapse/expand view toggle, the "add ELSE" control, the Range popup's "use range" button, the info-popup close button, and "use default" reset. That's a reasonable safety net for the editor UI landing in #3034, but it introduces no UI of its own.

**Verdict:** No UI/UX concerns. Any interface assessment of the routine editor itself belongs on the dependency PR #3034, which is where that markup and styling actually live.

VTT-AI-Agent:
Re: auto-review 2/4 (UI/UX) — agreed, nothing to change here. This PR only adds/modifies files under `tests/testcafe/` (a new suite plus two test-util helpers); no HTML, CSS, or client rendering code is touched, so there's no rendered interface for a layout/readability review to evaluate. The routine-editor markup the suite drives lives in the dependency PR #3034, which is where a UI/UX assessment of that editor belongs.

Re: auto-review 1/4 (code) — pushed a fix for both notes:
- `waitForWidgets` now re-fetches once more after the final backoff sleep and throws a clear `Error('waitForWidgets: predicate never became truthy within the timeout')` instead of returning the stale, already-checked result. Callers (e.g. `finalWidget.clickRoutine`) now fail with a legible timeout message instead of a `Cannot read properties of null` crash.
- Added a short comment in `readJsonEditorWidget` naming the U+00A0 (non-breaking space) character being normalized before `JSON.parse`.

`npm test` still passes (74/74). No production/engine/validator code is affected by this change, so no further upkeep needed there.

VTT-AI-Agent (commented):
## TL;DR
Test-only PR, and a good one — I checked out the branch, installed deps and ran `npx testcafe "chromium:headless --no-sandbox" tests/testcafe/routineeditor.js`: **all 3 tests pass (1m 05s)**. The one thing that needs action before/at merge: the branch is based on an older `main` where CI ran the whole `tests/testcafe` directory, but current `main` runs an explicit `testfile` matrix — so after a rebase this suite would silently never run in CI. Secondary point: the most splice-heavy parts of the routine editor (move up/down, ↰/↲/↱ block arrows, delete) are still untested.

## Findings

**1. `routineeditor.js` will not run in CI once this lands on current `main`.**
On this branch `.github/workflows/testcafe-chrome.yml` still ends with `args: "chrome:headless tests/testcafe"` (whole directory), so the new file is picked up automatically here. Current `main` has since moved to a per-file matrix (`.github/workflows/testcafe-chrome.yml:21-29` and the Firefox twin), listing `editor.js`, `compute-1..4.js`, `publiclibrary-1..4.js` explicitly. After #3034 and this PR are rebased/merged onto today's `main`, `routineeditor.js` is not in either matrix and would be dead weight — a suite that passes locally and is never exercised again. Please add `- routineeditor.js` to the `testfile` matrix in **both** `testcafe-chrome.yml` and `testcafe-firefox.yml` (a separate matrix entry is also the right shape given the ~65s runtime).

Related: the Firefox job is part of that matrix, and the test plan only covers Chromium. Historically Firefox has been the one to trip on `typeText`/`pressKey('tab')`-committed inputs and `input[type=number]`; worth a Firefox run before merge so it doesn't land as a red matrix leg.

**2. Coverage gap: the operations that mutate the routine array by index are untested** (`client/js/editor/controls/routine.js:463-491`).
The suite covers adding operations and editing parameters thoroughly, but every button in `.routine-editor-operation-buttons` is untouched:
- `arrow_upward` / `arrow_downward` — `this.routine.splice(index-1, 0, this.routine.splice(index, 1)[0])`
- `↰` / `↲` — `moveInto()`, which splices out of this routine and `unshift`/`push`es into an adjacent `IF.thenRoutine` / `FOREACH.loopRoutine`
- `↱` — `onHoist`, moving an operation out of a nested block
- `delete` — `this.routine.splice(index, 1)`

These are precisely the index-arithmetic paths where an off-by-one silently reorders or drops a user's operation, and they're the newest code in #3034 ("block-move arrows"). Given this PR's stated purpose — stress the surface a human tester can't repeatably click through — a fourth test that builds 3 operations, moves the middle one up, nests it into an adjacent `IF`, hoists it back out and deletes one, asserting the resulting `clickRoutine` after each step, would add more bug-catching value than any of the parameter-type paths already covered. Also uncovered: `.routine-editor-parameter-ignored-warning` and `.routine-editor-parameter-missing`.

**3. Popup helpers can address the wrong popup when popups stack** (`tests/testcafe/routineeditor.js:11`, `50-70`).
`popup = Selector('.inline-popup')` matches *all* open popups, and the third test explicitly asserts `popup.count == 2` — so stacking is a real state in this UI. Every helper then does `popup.find('input[type=text]')` / `popup.find('.accordion-content.open button')`, which searches across all open popups and acts on the first match, i.e. the *bottom-most* one. Today the tests always close a popup before opening the next, so it passes; but the day an interaction leaves a popup open (or an info popup renders first in DOM order), the failure will be a baffling "typed into the wrong field" rather than a clean selector miss. Making the helpers operate on `popup.nth(-1)` (top-most) would make them correct by construction rather than by call-order discipline.

**4. `waitForWidgets` polling budget is ~1.55s total** (`tests/testcafe/test-util.js:60-73`).
The shape mirrors `compareState`, and the post-loop re-fetch + thrown error (added in 2ebe750d, addressing the earlier review) is the right fix — failures are now legible. Just noting the ceiling: five attempts, ~1.55s, on a loaded GitHub runner for a 7-operation routine round-trip. If this ever flakes in CI, raise the loop bound rather than the initial delay. Not blocking.

**5. Minor:** predicates like `widgets => widgets[widgetID].clicks == 1 && widgets[widgetID]` (`routineeditor.js:236`, `240`) throw a `TypeError` out of `waitForWidgets` rather than retrying if the widget is momentarily absent from state. Safe as written (the widget exists by then), but the sibling predicates all guard with `widgets[widgetID] && …` — worth being consistent so a future copy-paste doesn't get a confusing crash instead of a timeout.

Everything else reads well: driving the editor purely through clicks/typing (never pasted JSON), double-verifying against both room state and the JSON editor read-back, the U+00A0→ASCII normalization before `JSON.parse`, and the "it actually runs when clicked" test that exits edit mode and uses the real play-mode click path. The `VTT_TEST_SERVER` fallback in `setupTestEnvironment` is additive and leaves CI defaults untouched. No production, engine or validator code is touched, so the compat/validator/`jeCommands` upkeep rules don't apply.

I did not add the `beta` label: this PR touches only `tests/testcafe/` — there is no player-facing behavior to try on the beta server — and it targets the `routine-editor-expanded-editor` feature branch rather than `main`, so it isn't beta-deployable in any meaningful sense.

96LawDawg:
PR #3034 has changed significantly since the last update in this PR. Update this PR with the changes from #3034,

VTT-AI-Agent:
> PR #3034 has changed significantly since the last update in this PR. Update this PR with the changes from #3034,

Done - `routine-editor-expanded-editor` is merged in (`f6d94788`) and the suite is rewritten against the editor as it is now.

The redesign took the list view, its parameter rows and the "ignored" markings away, so all 10 tests failed on the first run: an operation is a sentence now, with the phrase it starts with saying which way it works, the optional parts being its options and every blank a parameter chip. Everything the suite covered before is still covered, driven through the sentence instead; what the editor no longer has is replaced by what it has instead:

- **the variant menu** - picking another way for a `SET` to work has to rewrite all the parameters that tell the ways apart at once (*Increase* keeps the value the creator entered, *Toggle* drops it, *Set* has no relation at all)
- **the options** - adding *n times* and *ignore something* to a `CLICK` writes the parameters they word, and the ⊖ behind each phrase removes exactly those again
- **the "!" behind the phrase** while a `${...}` decides what the operation does: the value stays a chip of its own, and every entry of the drop-down says which reference picking it would write over
- **the Ctrl-selection**, which is index based and is dropped as soon as the routine changes

`routineeditor.js` also got its entry in the `testfile` matrix of both TestCafe workflows - the note that used to be at the bottom of the description. Without it the suite would never run in CI.

Verified locally: `npm test` 392/392, `routineeditor.js` 11/11 (run twice, no flakiness), `editor.js` 17/17.

PR #3052, opened 2026-07-25
Add keepOrder option to SWAPHANDS

VTT-AI-Agent:
## Goal

> Update the function SWAPHANDS so there is a way to keep the same order of the cards when passing them. Regression is not permitted, so if necessary, add a property so this is a selectable option moving forward.

(Request by lawdawg96 on [Discord](https://discord.com/channels/770758631146782780/1530705522301472931/1530711896381067284).)

## What was wrong

`SWAPHANDS` reads the cards of a hand via `children()`, so it *does* collect them in the order they are in that hand — but it then passes that list to a generated `MOVE` operation as a collection array. An array collection is resolved by `getCollection()` through `widgetFilter()`, which iterates the global widget map, so the cards are moved in **widget creation order** and not in the order of the array.

The receiving player therefore always gets the cards in the order the card widgets were created in (usually the original deck order), losing the arrangement the previous player had made in their hand.

## What changed

* new boolean property `keepOrder` for `SWAPHANDS` (default `false`, so nothing changes for existing games): when set, the cards of each hand are handed to the generated `MOVE` as an already ordered collection instead of as an array of IDs, so they arrive at the next seat in exactly the order they were in the hand (`MOVE` itself iterates its collection in array order — only an array of IDs gets re-resolved through `widgetFilter`)
* `validator/validate_gamefile.js`: `keepOrder` added to the `SWAPHANDS` operation properties
* `client/js/jsonedit.js`: `keepOrder` added to the context-sensitive insert buttons for `SWAPHANDS`
* the routine log now summarizes what `SWAPHANDS` did (`hands forward by 1, keeping the card order`) — it had no summary at all before, and with two possible behaviours the DEBUG panel should name which one ran. The generated moves are grouped under a `Moves` entry and each one gets a collection named after the seat it comes from, so the log reads `MOVE all widgets from 'hand of seat1' to 'seat2'`
* those temporary collections are filled right before their own `MOVE` and removed again afterwards: a collection of the surrounding routine that happens to use the same name is restored (even if a `MOVE` throws), and a widget that a routine of an earlier move removed is not passed on
* `tests/testcafe/routines.js`: new fixture (registered in both TestCafe workflow matrices) that pins the resulting card order down for both variants

This is creator-facing, so the wiki page **Functions** would need a line about `keepOrder` for `SWAPHANDS`, along the lines of: *when `true`, each hand's cards arrive at the next seat in the exact order they were in, instead of the order the card widgets were created in. Leave it off to keep the behaviour of existing games unchanged.*

## Verification

* `npm test` passes, TestCafe `editor.js` and the new `routines.js` (4 tests) pass locally — the fixture has two seated hands and three cards whose creation order differs from their order in the hand: the default variant asserts the unchanged legacy result (creation order), the `keepOrder` variant asserts the order of the hand, a third test keeps a `hand of seat1` collection of the calling routine alive across the operation and a fourth one covers three seats whose hands have `enterRoutine`s that delete cards. Disabling the `keepOrder` branch makes only the second test fail, dropping the collection restore makes only the third one fail
* the fixture asserts explicit widget IDs instead of comparing an md5 through `compareState()` (the usual convention in `docs/testcafe.md`), because the whole point is *which order* the cards are in — so it writes no reference state file
* validated a game file using `keepOrder` with `node validator/validate_gamefile_node.js` (no "unknown property", and an intentionally misspelled property is still reported)
* driven in a real browser (playwright): a hand arranged as `E A B C D` (cards created as `A B C D E`) arrives at the next seat as `A B C D E` with the unchanged default and as `E A B C D` with `keepOrder: true`

Playable example: [swaphands-keeporder-demo.vtt](https://agent.virtualtabletop.io/reports/pr/1530705522301472931/swaphands-keeporder-demo.vtt) — two seated hands, drag the cards inside a hand into any order, then compare the `SWAPHANDS` button with the `SWAPHANDS keepOrder` button (`reset` puts everything back).

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3052/pr-test (or any other room on that server)

### DIFF
--- a/.github/workflows/testcafe-chrome.yml
+++ b/.github/workflows/testcafe-chrome.yml
@@ -27,6 +27,7 @@ jobs:
           - publiclibrary-2.js
           - publiclibrary-3.js
           - publiclibrary-4.js
+          - routines.js
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
--- a/.github/workflows/testcafe-firefox.yml
+++ b/.github/workflows/testcafe-firefox.yml
@@ -27,6 +27,7 @@ jobs:
           - publiclibrary-2.js
           - publiclibrary-3.js
           - publiclibrary-4.js
+          - routines.js
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -3050,7 +3050,7 @@ function jeLoggingJSON(obj) {
 }
 
 export function jeLoggingRoutineStart(widget, property, initialVariables, initialCollections, byReference) {
-  if( jeHTMLStack.length == 0 || ['CALL', 'CLICK', 'IF', 'loopRoutine'].indexOf( jeHTMLStack[0][3] ) == -1 ) {
+  if( jeHTMLStack.length == 0 || ['CALL', 'CLICK', 'IF', 'loopRoutine', 'Moves'].indexOf( jeHTMLStack[0][3] ) == -1 ) {
     if(jeRoutineResetOnNextLog) {
       jeLoggingHTML = '';
       jeRoutineResetOnNextLog = false;
@@ -3068,7 +3068,7 @@ export function jeLoggingRoutineStart(widget, property, initialVariables, initia
 }
 
 export function jeLoggingRoutineEnd(variables, collections) {
-  if( jeHTMLStack.length == 0 || ['CALL', 'CLICK', 'IF', 'loopRoutine'].indexOf( jeHTMLStack[0][3] ) == -1 ) jeLoggingHTML += '</div></div>';
+  if( jeHTMLStack.length == 0 || ['CALL', 'CLICK', 'IF', 'loopRoutine', 'Moves'].indexOf( jeHTMLStack[0][3] ) == -1 ) jeLoggingHTML += '</div></div>';
   --jeLoggingDepth;
   if(!jeLoggingDepth) {
     $('#jeLog').innerHTML = jeLoggingHTML + '</div></div>';
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -1389,7 +1389,7 @@ function jeAddCommands() {
   jeAddRoutineOperationCommands('SET', { collection: 'DEFAULT', property: 'parent', relation: '=', value: null });
   jeAddRoutineOperationCommands('SHUFFLE', { holder: null, collection: 'DEFAULT', mode: 'true random', modeValue: 1 });
   jeAddRoutineOperationCommands('SORT', { key: 'value', reverse: false, rearrange: false, locales: null, options: null, holder: null, collection: 'DEFAULT' });
-  jeAddRoutineOperationCommands('SWAPHANDS', { interval: 1, direction: 'forward', source: 'all' });
+  jeAddRoutineOperationCommands('SWAPHANDS', { interval: 1, direction: 'forward', source: 'all', keepOrder: false });
   jeAddRoutineOperationCommands('TIMER', { value: 0, seconds: 0, mode: 'toggle', timer: null, collection: 'DEFAULT' });
   jeAddRoutineOperationCommands('TURN', { turn: 1, turnCycle: 'forward', source: 'all', collection: 'TURN' });
   jeAddRoutineOperationCommands('UPLOAD', { variable: 'uploadedFileName', fileTypes: [ '.png', '.jpg', '.jpeg', '.gif', '.svg', '.webp', '.json', '.mp3', '.wav', '.ogg', '.m4a' ] });
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -2050,8 +2050,9 @@ export class Widget extends StateManaged {
               [c[i], c[rand]] = [c[rand], c[i]];
             }
           }
+          // all hands are collected before anything is moved so that a hand does not
+          // pick up the widgets an earlier seat just passed to it
           let moves = [];
-          let shadowedCollections = {};
           for (let i = 0; i < c.length; i++) {
             let source = c[i];
             let target = c[(i + a.interval) % c.length];
@@ -2061,44 +2062,49 @@ export class Widget extends StateManaged {
               let contents = widgets.get(hand).children().reduce(
                 function (collect, w) {
                   if (!perOwner || w.get('owner') == source.get('player')) {
-                    collect.unshift(w.get('id'));
+                    collect.unshift(w);
                   }
                   return collect
                 },
                 []
               );
+              moves.push({ source, contents, to: target.get('id') });
+            }
+          }
+          if(moves.length) {
+            if(jeRoutineLogging)
+              jeLoggingRoutineOperationStart("Moves", "Moves");
+            for(const move of moves) {
               // the collection is named after the seat it comes from so that the
               // generated MOVE reads like "from 'hand of seat1' to 'seat2'" in the log.
               // a collection of the surrounding routine that happens to use the same
-              // name is put back once the moves are done
-              const collection = `hand of ${source.get('id')}`;
-              shadowedCollections[collection] = collections[collection];
-              // an array of IDs is turned into a collection by widgetFilter, which
-              // returns the widgets in creation order - keepOrder uses a collection
-              // that is already sorted like the hand instead so the order is kept
+              // name is shadowed only while its MOVE runs and then put back
+              const collection = `hand of ${move.source.get('id')}`;
+              const shadowed = collections[collection];
+              // the widgets are looked up right before their own MOVE so that one which
+              // a routine of an earlier MOVE removed is left alone, exactly like when
+              // the generated MOVE still received a list of IDs. keepOrder keeps the
+              // order of the hand, the default is the creation order because that is
+              // the order widgetFilter - and with it MOVE - used all along
               collections[collection] = a.keepOrder
-                ? contents.map(id=>widgets.get(id))
-                : widgetFilter(w=>contents.indexOf(w.get('id')) != -1);
-              moves.push({
-                func: "MOVE",
-                collection: collection,
-                to: target.get('id'),
-              });
+                ? move.contents.filter(w=>!w.isBeingRemoved)
+                : widgetFilter(w=>move.contents.indexOf(w) != -1);
+              try {
+                await this.evaluateRoutine([ { func: 'MOVE', collection, to: move.to } ], variables, collections, (depth || 0) + 1, true);
+              } finally {
+                if(shadowed === undefined)
+                  delete collections[collection];
+                else
+                  collections[collection] = shadowed;
+              }
             }
+            if(jeRoutineLogging)
+              jeLoggingRoutineOperationEnd([], variables, collections, false);
           }
-          if(jeRoutineLogging)
-            jeLoggingRoutineOperationStart("Moves", "Moves");
-          await this.evaluateRoutine(moves, variables, collections, (depth || 0) + 1, true);
-          for(const [ name, shadowed ] of Object.entries(shadowedCollections)) {
-            if(shadowed === undefined)
-              delete collections[name];
-            else
-              collections[name] = shadowed;
+          if(jeRoutineLogging) {
+            const how = a.direction == 'random' ? `hands in a random seat order by ${a.interval}` : `hands ${a.direction} by ${a.interval}`;
+            jeLoggingRoutineOperationSummary(moves.length ? `${how}${a.keepOrder ? ', keeping the card order' : ''}` : 'no seat with a player has a valid hand, nothing to swap');
           }
-          if(jeRoutineLogging)
-            jeLoggingRoutineOperationEnd([], variables, collections, false);
-          if(jeRoutineLogging)
-            jeLoggingRoutineOperationSummary(`hands ${a.direction} by ${a.interval}${a.keepOrder ? ', keeping the card order' : ''}`);
         } else if(jeRoutineLogging) {
           jeLoggingRoutineOperationSummary('less than two seats with a player, nothing to swap');
         }
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -2066,14 +2066,15 @@ export class Widget extends StateManaged {
                 },
                 []
               );
-              let collection = contents;
-              if(a.keepOrder) {
-                // an array of IDs is turned into a collection by widgetFilter, which
-                // returns the widgets in creation order - pass a collection that is
-                // already sorted like the hand instead so the order is kept
-                collection = `$swaphands_${i}`;
-                collections[collection] = contents.map(id=>widgets.get(id));
-              }
+              // the collection is named after the seat it comes from so that the
+              // generated MOVE reads like "from 'hand of seat1' to 'seat2'" in the log
+              const collection = `hand of ${source.get('id')}`;
+              // an array of IDs is turned into a collection by widgetFilter, which
+              // returns the widgets in creation order - keepOrder uses a collection
+              // that is already sorted like the hand instead so the order is kept
+              collections[collection] = a.keepOrder
+                ? contents.map(id=>widgets.get(id))
+                : widgetFilter(w=>contents.indexOf(w.get('id')) != -1);
               moves.push({
                 func: "MOVE",
                 collection: collection,
@@ -2081,13 +2082,13 @@ export class Widget extends StateManaged {
               });
             }
           }
-          if(jeRoutineLogging) {
+          if(jeRoutineLogging)
             jeLoggingRoutineOperationStart("Moves", "Moves");
-          }
           await this.evaluateRoutine(moves, variables, collections, (depth || 0) + 1, true);
-          if(jeRoutineLogging) {
-          jeLoggingRoutineOperationEnd([], variables, collections, false);
-          }
+          for(const move of moves)
+            delete collections[move.collection];
+          if(jeRoutineLogging)
+            jeLoggingRoutineOperationEnd([], variables, collections, false);
           if(jeRoutineLogging)
             jeLoggingRoutineOperationSummary(`hands ${a.direction} by ${a.interval}${a.keepOrder ? ', keeping the card order' : ''}`);
         } else if(jeRoutineLogging) {
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -2051,6 +2051,7 @@ export class Widget extends StateManaged {
             }
           }
           let moves = [];
+          let shadowedCollections = {};
           for (let i = 0; i < c.length; i++) {
             let source = c[i];
             let target = c[(i + a.interval) % c.length];
@@ -2067,8 +2068,11 @@ export class Widget extends StateManaged {
                 []
               );
               // the collection is named after the seat it comes from so that the
-              // generated MOVE reads like "from 'hand of seat1' to 'seat2'" in the log
+              // generated MOVE reads like "from 'hand of seat1' to 'seat2'" in the log.
+              // a collection of the surrounding routine that happens to use the same
+              // name is put back once the moves are done
               const collection = `hand of ${source.get('id')}`;
+              shadowedCollections[collection] = collections[collection];
               // an array of IDs is turned into a collection by widgetFilter, which
               // returns the widgets in creation order - keepOrder uses a collection
               // that is already sorted like the hand instead so the order is kept
@@ -2085,8 +2089,12 @@ export class Widget extends StateManaged {
           if(jeRoutineLogging)
             jeLoggingRoutineOperationStart("Moves", "Moves");
           await this.evaluateRoutine(moves, variables, collections, (depth || 0) + 1, true);
-          for(const move of moves)
-            delete collections[move.collection];
+          for(const [ name, shadowed ] of Object.entries(shadowedCollections)) {
+            if(shadowed === undefined)
+              delete collections[name];
+            else
+              collections[name] = shadowed;
+          }
           if(jeRoutineLogging)
             jeLoggingRoutineOperationEnd([], variables, collections, false);
           if(jeRoutineLogging)
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -2032,7 +2032,7 @@ export class Widget extends StateManaged {
       }
 
       if(a.func == 'SWAPHANDS') {
-        setDefaults(a, { interval: 1, direction: 'forward', source: 'all' });
+        setDefaults(a, { interval: 1, direction: 'forward', source: 'all', keepOrder: false });
         if(['forward', 'backward', 'random'].indexOf(a.direction) == -1) {
           problems.push(`Warning: direction ${a.direction} interpreted as forward.`);
           a.direction = 'forward'
@@ -2066,11 +2066,22 @@ export class Widget extends StateManaged {
                 },
                 []
               );
-              moves.push({
-                func: "MOVE",
-                collection: contents,
-                to: target.get('id'),
-              });
+              if(a.keepOrder) {
+                // a MOVE collection is processed in widget creation order, so move
+                // the cards one by one to have them arrive in the order of the hand
+                for(const id of contents)
+                  moves.push({
+                    func: "MOVE",
+                    collection: [ id ],
+                    to: target.get('id'),
+                  });
+              } else {
+                moves.push({
+                  func: "MOVE",
+                  collection: contents,
+                  to: target.get('id'),
+                });
+              }
             }
           }
           if(jeRoutineLogging) {
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -2066,22 +2066,19 @@ export class Widget extends StateManaged {
                 },
                 []
               );
+              let collection = contents;
               if(a.keepOrder) {
-                // a MOVE collection is processed in widget creation order, so move
-                // the cards one by one to have them arrive in the order of the hand
-                for(const id of contents)
-                  moves.push({
-                    func: "MOVE",
-                    collection: [ id ],
-                    to: target.get('id'),
-                  });
-              } else {
-                moves.push({
-                  func: "MOVE",
-                  collection: contents,
-                  to: target.get('id'),
-                });
+                // an array of IDs is turned into a collection by widgetFilter, which
+                // returns the widgets in creation order - pass a collection that is
+                // already sorted like the hand instead so the order is kept
+                collection = `$swaphands_${i}`;
+                collections[collection] = contents.map(id=>widgets.get(id));
               }
+              moves.push({
+                func: "MOVE",
+                collection: collection,
+                to: target.get('id'),
+              });
             }
           }
           if(jeRoutineLogging) {
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -2088,6 +2088,10 @@ export class Widget extends StateManaged {
           if(jeRoutineLogging) {
           jeLoggingRoutineOperationEnd([], variables, collections, false);
           }
+          if(jeRoutineLogging)
+            jeLoggingRoutineOperationSummary(`hands ${a.direction} by ${a.interval}${a.keepOrder ? ', keeping the card order' : ''}`);
+        } else if(jeRoutineLogging) {
+          jeLoggingRoutineOperationSummary('less than two seats with a player, nothing to swap');
         }
       }
 
--- a/library/tutorials/Functions - SWAPHANDS/0.json
+++ b/library/tutorials/Functions - SWAPHANDS/0.json
@@ -678,7 +678,7 @@
     "css": {
       "font-size": "25px"
     },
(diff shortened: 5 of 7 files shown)


VTT-AI-Agent (commented):
## TL;DR

The diagnosis is right (an array `collection` is resolved through `widgetFilter()`, so it comes back in widget-creation order) and the fix is correctly ordered and properly opt-in — I traced `children()` (descending `z`) → `unshift` (ascending `z`) → per-card `bringToFront()` + `updateAfterShuffle()` (which renumbers `z` 1..n ascending) and the receiving hand does end up in the source hand's order. No correctness or safety problem found; my only substantive point is that the same result can be had with one `MOVE` per hand instead of one per card.

## Findings

### 1. `client/js/widgets/widget.js:2069-2079` — one `MOVE` per card is avoidable

`MOVE` already iterates its collection **in array order** (`for(const c of collections[collection].slice(offset, offset+count))`, widget.js:1687). The only thing that destroys the order is `getCollection()` (widget.js:917-920): when it is handed a raw *array of ids* it rebuilds the collection with `widgetFilter(...)`, which walks the global `widgets` map. If the collection is already present in `collections` as an array of widgets, `getCollection()` returns it untouched (widget.js:915-916). So the ordered variant can stay a single operation per hand:

```js
const key = `$swaphands_${i}`;
collections[key] = contents.map(id=>widgets.get(id));
moves.push({ func: 'MOVE', collection: a.keepOrder ? key : contents, to: target.get('id') });
```

Why it's worth considering:

* today `keepOrder` turns one op into *n* ops per hand, each of which re-runs `getCollection()` → `widgetFilter()` over **every widget in the room** (O(cards × widgets) instead of O(widgets)), and each of which emits its own routine-log entry, so the JE routine log for a 4×13-card pass goes from 4 lines to 52
* it keeps the two branches symmetric — right now the `keepOrder` branch is a different *shape* of routine, which makes the operation harder to reason about when debugging in edit mode

Caveat if you take it: capturing widget objects up front skips `widgetFilter`'s `!w.isBeingRemoved` guard, so a card removed between the collection build and the `MOVE` would be passed through as a stale widget. Given the `moves` array is built and executed within the same operation with no user code in between, that's very unlikely, but it is the one behavioural difference.

Either way this is a refinement, not a bug — the current implementation produces the right result.

### 2. No automated test

`SWAPHANDS` has no test today and this PR adds none, so the ordering contract is guarded only by the manual browser check described in the PR. The state-hash comparison in the TestCafe suite (`docs/testcafe.md`) would catch a regression here cheaply: a room with two seated hands, a couple of cards dragged into a non-creation order, a `SWAPHANDS keepOrder: true` button, hash. Worth adding, because the property's whole value is the *exact* resulting order and nothing else in the repo pins it down — a future change to `bringToFront`/`updateAfterShuffle` z-renumbering would silently break it.

### 3. Discoverability

The PR correctly notes the wiki **Functions** page needs a line. `library/tutorials/Functions - SWAPHANDS/0.json` already demonstrates `direction` and `source` with one button each; a `keepOrder` button there would be the natural place for a creator to actually *see* the difference. Optional, and note `library/**` changes are CRC-checked by `check-library.yml` (see `docs/conventions.md`), so it should be exported rather than hand-edited.

### Checked and fine

* `keepOrder: false` in `setDefaults` plus the `if(a.keepOrder)` guard means existing games are byte-for-byte unaffected — no `FileUpdater` / compat concern.
* Validator entry uses `'boolean'`, matching `SORT`'s `reverse`/`rearrange`; `jeCommands` entry follows the "insert with the default value" convention used by the sibling ops. Both required by CLAUDE.md's validator/JSON-editor upkeep rule, and both are present.
* Shared-hand + `childrenPerOwner` case: the per-owner filter is applied before any move and the ids are captured up front, so the moves are still simultaneous-by-construction, and per-owner relative order is preserved because each card is brought to front as it lands.
* `direction: random` / `interval` / invalid `hand` paths are untouched.

---

I added the `beta` label: `client/js/widgets/widget.js` is the only engine file touched and its change is entirely gated behind the new opt-in `keepOrder` property (default `false`), so existing games' behaviour is unchanged and the engine impact is incidental — the rest is edit-mode (`jsonedit.js`) and validator-only, and there is no crash or state-corruption risk for players testing it.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review) of #3052. This is an engine-side change, so the only interface it touches is the **JSON editor's context-sensitive insert buttons** — and there it behaves correctly: `keepOrder` shows up in the right context, in the right alphabetical slot, styled like every other property button, and the validator entry keeps edit mode free of a false "unknown property" error. The most impactful improvement available is **discoverability/labelling**: a bare `keepOrder` button with no in-app explanation, next to `direction`/`interval` (which are about *seats*), is easy for a first-time game creator to misread — the wiki line the PR description promises should land with the merge, and the routine-debug log should say what SWAPHANDS actually did.

Reviewed by running the app locally (playwright, port 8330): built a room with two seats/hands and a `SWAPHANDS` button, opened the JSON editor and the DEBUG panel, and compared `keepOrder: true` against the unchanged default.

| SWAPHANDS insert buttons (property absent) | Operation with `keepOrder: true` |
|---|---|
| insert buttons | json editor |

Verified visually:
* the new button appears under context `button ↦ clickRoutine ↦ 0 ↦ (SWAPHANDS) ↦` as `comment / direction / interval / keepOrder / source` — alphabetically sorted with the rest, same pill styling, no wrapping or overflow in the sidebar;
* it disappears once the property is present, like every other property command;
* the DEBUG panel reports "No validation problems found!" for a game using `keepOrder`, so `validator/validate_gamefile.js:794` is doing its job — no red herring for creators in edit mode;
* on the table itself there is no visible difference in how the cards arrive (same timing, same animation, no per-card staggered "deal") — the change is invisible to players, which is the right outcome.

---

### Big layout improvements

Nothing to restructure — the diff adds no markup, CSS or new panel. The one thing worth treating as a "big" item is **what the interface tells the creator about this option**:

1. **`SWAPHANDS` has no routine-log summary, and this PR makes that gap more noticeable.** Almost every neighbouring operation calls `jeLoggingRoutineOperationSummary(...)` (53 call sites in `client/js/widgets/widget.js`) so the DEBUG log shows a one-line human summary next to the operation name — e.g. SORT renders `widgets in 'DEFAULT' by value`. `SWAPHANDS` calls none, so the log entry is just the bare word `SWAPHANDS`, and the only nested group is labelled `Moves`. Now that the same operation can behave in two different ways, a creator debugging "why did my cards come back shuffled?" gets no signal from the log about which mode ran. Suggestion — one line next to the existing logging block at `client/js/widgets/widget.js:2087`:
   ```js
   if(jeRoutineLogging)
     jeLoggingRoutineOperationSummary(`hands ${a.direction} by ${a.interval}${a.keepOrder ? ', keeping card order' : ''}`);
   ```
   This is exactly the phrasing style the other operations already use, costs nothing, and makes the new flag observable in the app rather than only in the JSON.

2. **The wiki line is the whole UI for this feature.** The JSON editor's property buttons carry no `title`/tooltip (only the top icon buttons do, `client/js/jsonedit.js:3483`), so once `keepOrder` is inserted there is nothing in the interface that says what it means. The PR body correctly notes the wiki **Functions** page needs a line — please treat that as part of shipping this, not a follow-up, since it is the only place a first-time creator can find out what "order" refers to.

3. **`client/js/jsonedit.js:1392` — the button inserts `"keepOrder": false`, which is a no-op.** The default is already `false`, so the *only* reason anyone clicks this button is to switch the behaviour on; as written they insert a line that changes nothing and then have to reach for the separate `toggle boolean` command (`jsonedit.js:253`) to make it do something. There is precedent for seeding the useful value: `RECALL` seeds `owned: true, inHolder: true` and `CALL` seeds `return: true`. Inserting `keepOrder: true` would make the button a one-click action and remains harmless (`setDefaults` still yields `false` for everyone who never clicks it).

4. **Log verbosity in the DEBUG panel (minor, mode-specific).** With `keepOrder: true` the `Moves` group gets one nested `MOVE` entry *per card per seat* instead of one per seat — a 13-card, 4-player hand goes from 4 log rows to 52, all identically labelled `MOVE`, which makes the expander tree noticeably harder to scan than in default mode. Not a blocker (the log is already deep and collapsible), but it's another argument for the summary line in item 1 so the parent row explains why the group suddenly exploded.

5. Nothing else broke: I looked for wrapping/overflow in the command sidebar with the extra button and for ordering surprises — the list stays alphabetical and the sidebar scrolls as before. Screenshot above.

6. **The name `keepOrder` is ambiguous in its own context.** It sits directly under `direction` and `interval`, both of which describe how hands move *between seats*, so "order" reads just as naturally as "seat order" as it does "card order". `keepCardOrder` (or `keepHandOrder`) would be self-explanatory at the point of use and would still sort into the same visual neighbourhood in the button list. If the name stays, the wiki line should open with the disambiguation, e.g. *"keeps the cards in the order the previous player had arranged them in their hand"*, rather than describing the internal MOVE behaviour.

7. **Suggested wiki wording**, matching the tone of the other flags: *"`keepOrder` (boolean, default `false`) — when `true`, each hand's cards arrive at the next seat in the exact order they were in, instead of the order the card widgets were created in. Leave it off to keep the behaviour of existing games unchanged."* The "leave it off / existing games unchanged" half matters: a creator reading the property list has no way to know from the editor that the default is the legacy behaviour.

8. Positive on the newcomer front: because the default is `false`, an existing game opened by someone who has never seen this flag looks and behaves exactly as before, and nothing new appears on the table or in the player-facing UI. That's the right call for a creator-facing option.

VTT-AI-Agent:
Thanks - both actionable points are addressed in 1eb96de3.

**1. One `MOVE` per hand instead of one per card:** taken as suggested. `keepOrder` now builds an ordered collection (`$swaphands_<i>`, following the `$collection_<n>` naming `getCollection()` already uses) and passes it to a single `MOVE` per hand, so the two branches only differ in *what* collection they hand over:

```js
let collection = contents;
if(a.keepOrder) {
  collection = `$swaphands_${i}`;
  collections[collection] = contents.map(id=>widgets.get(id));
}
moves.push({ func: "MOVE", collection: collection, to: target.get('id') });
```

Re the caveat about skipping `widgetFilter`'s `!w.isBeingRemoved` guard: the IDs come from `children()` of the hand and are turned into widgets in the same synchronous block, before any of the moves run, so there is no point at which a card could be removed in between.

**2. Automated test:** added `tests/testcafe/routines.js` (registered in the Chrome and Firefox workflow matrices), with a room of two seated hands, three cards whose creation order differs from their order in the hand, and one test per variant:

* `SWAPHANDS` (default) - the cards arrive in widget creation order, which pins down the unchanged legacy behaviour
* `SWAPHANDS` with `keepOrder: true` - the cards arrive in the order of the hand

Both pass locally, and with the `keepOrder` branch disabled the second test fails, so it does guard the ordering contract rather than just the code path. It asserts the resulting card order directly instead of a state hash, because the order is exactly what the property is about.

**3. Discoverability:** the wiki **Functions** page still needs the `keepOrder` line - that is outside the repo. I left `library/tutorials/Functions - SWAPHANDS/` alone: nobody asked for a tutorial change here, and library files are CRC-checked, so adding a button there is better done as a deliberate, separately exported change.

VTT-AI-Agent:
Thanks for running it in a browser — one item taken, one already fixed, and two I'd push back on with evidence.

### 1. Routine-log summary — done (10a97c35)

`SWAPHANDS` now calls `jeLoggingRoutineOperationSummary`, with your phrasing:

```js
jeLoggingRoutineOperationSummary(`hands ${a.direction} by ${a.interval}${a.keepOrder ? ', keeping the card order' : ''}`);
```

It also gets a summary in the case where nothing happens at all (`less than two seats with a player, nothing to swap`), which was previously an entry with no explanation whatsoever — that is the other question a creator asks the log about this operation.

Verified in the DEBUG panel (playwright, edit mode + middle-click to trigger the routine):

* default: `SWAPHANDS hands forward by 1`
* `keepOrder: true`: `SWAPHANDS hands forward by 1, keeping the card order` — [screenshot](https://agent.virtualtabletop.io/reports/pr/1530705522301472931/log-keeporder-true.png)
* one seated player: `SWAPHANDS less than two seats with a player, nothing to swap`

### 4. Log verbosity — already fixed before this review

The review was written against bf29d304; 1eb96de3 (from the code review) replaced the per-card moves with one ordered collection per hand. The `Moves` group is now back to one `MOVE` row per seat in both modes — the same DEBUG run shows `MOVE all widgets from '$swaphands_0' to 'seat2'` / `from '$swaphands_1' to 'seat1'`, so a 4×13-card pass is 4 rows, not 52.

### 3. Inserting `keepOrder: true` — I'd rather not

The `RECALL owned: true` precedent isn't a seeded "useful value": `jeAddRoutineOperationCommands` mirrors each operation's **own defaults**, and every boolean in that list matches its `setDefaults` value — `owned`/`inHolder`/`return`/`resetOwner`/`snapToGrid` default to `true`, while `reverse`/`byDistance`/`recursive`/`skipMissing`/`silence`/`random`/`block` default to `false` and are all inserted as `false`. `SORT`'s `reverse: false` is the exact analogue of `keepOrder: false`. Seeding `true` would make this the only boolean in the list that doesn't show the creator what the operation does today, and the editor's `toggle boolean` command is one click away. (The genuine outlier is `SORT`'s `rearrange: false`, whose default is `true` — a separate, pre-existing bug.)

### 6. `keepOrder` vs `keepCardOrder` — keeping the short name

Two reasons: a hand is a holder, and its `dropTarget` can accept things that aren't cards, so `keepCardOrder` would promise slightly more than the operation does; and routine-operation parameters are consistently short in this engine (`reverse`, `rearrange`, `owned`, `inHolder`, `return`, `face`) — the verbose style belongs to widget properties. Within `SWAPHANDS` there is exactly one collection whose order can be kept, and seat rotation is already named explicitly by `direction`/`interval`. That said, the name is unreleased, so if @ArnoldSmith86 prefers `keepCardOrder` it is a one-line rename in three files — happy to do it.

### 2 + 7. Wiki

Agreed, and I'll use your wording as the basis. The wiki is outside this repo, so I can't push it with the merge — the PR description flags it, and the `keepOrder` line should read along the lines of *"when `true`, each hand's cards arrive at the next seat in the exact order they were in, instead of the order the card widgets were created in. Leave it off to keep the behaviour of existing games unchanged."*

96LawDawg:
The debug output is not consistent with other functions. The "--custom--" text is not helpful.  The from '$swaphands_0` is meaningless as that is not a widget the user can control. Better to say from seat1 to seat2, etc.

image

VTT-AI-Agent:
Good catch, fixed in ba134380 — both parts of the debug output:

* the stray `button-forward --custom--` level is gone. It appeared because the generated MOVEs are a literal routine, and `jeLoggingRoutineStart` only suppresses that widget/property header for the nested routines it knows about (`CALL`, `CLICK`, `IF`, `loopRoutine`) — `Moves` is now in that list too, so the MOVEs nest directly under `Moves` like a FOREACH's `loopRoutine` does.
* the internal `$swaphands_0` collection name is gone. The collection each generated MOVE gets is now named after the seat it comes from, so the log reads `MOVE all widgets from 'hand of seat-alice' to 'seat-bob'`. The temporary collections are also deleted again after the moves, so they no longer show up in "Collections afterwards" of the surrounding operation.

Since the name no longer encodes the variant, both variants now go through a named collection — the default one is filled with `widgetFilter` (exactly what `getCollection` does for an array of IDs), so the legacy creation order is unchanged. The TestCafe test that pins both orders still passes.

The same click as in your screenshot now looks like this:

routine log after the change

VTT-AI-Agent (commented):
## TL;DR

The opt-in ordering implementation itself is sound, but the latest logging-oriented refactor introduces a default-path compatibility regression: `SWAPHANDS` can overwrite and then delete a caller-owned routine collection. This should be fixed and covered before the PR is considered safe to deploy.

## Findings

### 1. `client/js/widgets/widget.js:2071-2089` — temporary collection names can destroy caller collections

Routine collections are caller-visible, arbitrary named state for the lifetime of the invocation. The new code unconditionally assigns each temporary hand to `collections[\`hand of ${source.id}\`]`, runs the generated `MOVE`s by reference, and then unconditionally deletes those keys.

That means a perfectly valid routine such as:

1. `SELECT` widgets into a collection named `hand of seat1`
2. run the existing/default `{ "func": "SWAPHANDS" }`
3. use `hand of seat1` in a later operation

now loses the selected collection at step 2. The assignment at line 2075 first replaces it with seat1's hand; the cleanup at line 2089 then removes it entirely, so the later operation reports that the collection does not exist or acts on the temporary hand if reached while the nested routine is running. This affects `keepOrder: false`, so it breaks the PR's backwards-compatibility guarantee for existing game JSON.

Please allocate names that are guaranteed unused for this invocation (for example, probe `collections` and suffix until each key is free), or save and restore pre-existing values rather than blindly deleting them. The new TestCafe fixture should also seed a colliding collection and verify it remains usable after `SWAPHANDS`; its current assertions only inspect card state, so they cannot catch routine-collection corruption.

The `beta` label was already present, but because this default-path engine regression can break existing routines, I left it untouched per policy and recommend a maintainer remove it until the regression is fixed.

VTT-AI-Agent:
Valid catch, thanks — fixed in b57e56ee.

The collision was introduced by ba134380 (naming the temporary collections after the seats so the log reads `from 'hand of seat1' to 'seat2'` instead of `from '$swaphands_0'`), and you're right that it hits the default path too, since both variants now go through a named collection.

**Fix:** instead of blindly deleting the names afterwards, `SWAPHANDS` now remembers what each name held and puts it back:

```js
const collection = `hand of ${source.get('id')}`;
shadowedCollections[collection] = collections[collection];
...
for(const [ name, shadowed ] of Object.entries(shadowedCollections)) {
  if(shadowed === undefined)
    delete collections[name];
  else
    collections[name] = shadowed;
}
```

I went with save/restore rather than probing for a free name so the log line stays stable — a creator debugging a game shouldn't see `hand of seat1_2` just because their routine happens to use that name. The only window in which the caller's collection is shadowed is the generated `MOVE`s themselves, which are a closed routine with no user code in it, so nothing can observe the shadowed value.

**Test:** added the colliding-collection case you asked for to `tests/testcafe/routines.js`:

```js
{ func: 'SELECT', property: 'id', value: 'card1', collection: 'hand of seat1' },
{ func: 'SWAPHANDS' },
{ func: 'SET', collection: 'hand of seat1', property: 'marked', value: true }
```

It asserts both that the cards still arrive in the default (creation) order and that exactly `card1` ends up marked — i.e. the collection is still usable after the operation. All three tests pass locally, and with the restore removed the new one fails, so it does pin the contract rather than just the code path. `npm test` passes as well.

On the `beta` label: leaving it to a maintainer either way, as I'm the PR author.

VTT-AI-Agent (commented):
## TL;DR

The collection-clobbering regression from review 3 is fixed and now covered by a dedicated test — I ran the new `tests/testcafe/routines.js` locally (chromium headless, all 3 pass) plus `npm test` (44 pass). What is left are two smaller things in the *default* (`keepOrder: false`) path introduced by the collection-naming refactor: the collection is now resolved **eagerly** through `widgetFilter` before any `MOVE` runs (it used to be resolved lazily inside each `MOVE`), and the shadow/restore of the caller's collection is skipped if the nested routine throws.

## Findings

### 1. `client/js/widgets/widget.js:2079-2081` — the default path now resolves its collection before any move runs

Before this PR the generated op was `{ func: 'MOVE', collection: contents }` with `contents` an array of IDs, so `getCollection()` (widget.js:913-923) turned it into widgets **at the moment that `MOVE` executed** — i.e. seat *n*'s collection was built after seats *0..n-1* had already moved their cards. Now every seat's collection is built up-front, in the build loop.

That matters because `widgetFilter()` (`serverstate.js:620-622`) filters out `!w.isBeingRemoved`, and moving cards runs arbitrary user code: `set('parent')` fires the old parent's `leaveRoutine` and the new parent's `enterRoutine` (widget.js:2653-2660), plus `holder.js:108`. So with 3+ seats and a hand that has an enter/leave routine which deletes a widget, a later seat's collection can now contain a widget that is in the removal queue — previously it would have been dropped. `applyMove` then calls `moveToHolder` on a dying widget.

This is narrow, but it is a change on the path that existing games take, so it is worth either avoiding or accepting deliberately. The cheapest way to avoid it is to only take the named-collection route when it's actually needed:

```js
collections[collection] = contents.map(id=>widgets.get(id));   // only when a.keepOrder
moves.push({ func: 'MOVE', collection: a.keepOrder ? collection : contents, to: target.get('id') });
```

…at the cost of the nicer log label in default mode. (The `keepOrder` path has the same staleness by construction — `widgets.get(id)` has no `isBeingRemoved` guard — but there it's inherent to preserving order, and the PR is opt-in, so that's a fair trade.)

### 2. `client/js/widgets/widget.js:2091-2097` — the restore is not exception-safe

`shadowedCollections` is only written back *after* `await this.evaluateRoutine(moves, …)` returns normally. If any generated `MOVE` throws (an `enterRoutine` that blows up, a widget removed mid-flight), the caller's `hand of seatX` collection is left replaced by the temporary hand for the rest of the routine — the exact failure review 3 described, just on the error path. Wrapping the `evaluateRoutine` call in `try { … } finally { restore }` closes it in three lines.

Alternatively, the whole shadow/restore block disappears if the temporary name can't collide in the first place — the engine already reserves a `$`-prefixed namespace for exactly this (`getCollection()` uses `'$collection_' + batchDepth`), so `$hand of ${source.get('id')}` would keep the log readable *and* remove the class of bug. Your call; the current code is correct on the happy path and the new test pins it down.

### 3. `client/js/widgets/widget.js:2061-2081` — the ID round-trip is avoidable

`children()` already returns widget objects, so `contents` is built as IDs only to be mapped straight back to widgets (`keepOrder`) or matched by ID against every widget in the room (`widgetFilter` + `indexOf`, O(cards × widgets)). Collecting the widgets themselves makes both branches shorter and the default branch cheaper:

```js
collect.unshift(w);
…
collections[collection] = a.keepOrder ? contents : widgetFilter(w=>contents.indexOf(w) != -1);
```

Also, the three consecutive `if(jeRoutineLogging)` guards at 2089/2098/2100 can be one block.

### 4. Log summary wording, minor

* `hands ${a.direction} by ${a.interval}` renders as `hands random by 1` for `direction: 'random'`, which doesn't parse as English — worth special-casing (`hands randomly` / `hands shuffled between seats`).
* The summary is emitted unconditionally once `c.length > 1`, even when every seat's `hand` failed `isValidID` and `moves` is empty — the log then claims a swap that didn't happen. The `problems` entries do show up, so this is cosmetic.

### 5. Tests

Verified locally against this branch (server on a private port, chromium headless):

```
✓ SWAPHANDS passes the cards on in widget creation order
✓ SWAPHANDS with keepOrder passes the cards on in the order of the hand
✓ SWAPHANDS leaves a collection of the surrounding routine intact
3 passed (26s)
```
`npm test` — 44 passed.

The third test is a good, targeted regression guard for the collection collision. Remaining gaps, if you want them: `childrenPerOwner` (a shared hand with two owners) combined with `keepOrder`, which is the one case where `contents` is a filtered subset of the holder's children; and `interval > 1` / `direction: 'backward'`, which are untouched but now share the new collection plumbing. Note also that the fixture deliberately asserts explicit ID orders rather than using `compareState`'s md5 (the convention in `docs/testcafe.md`) — that reads much better for an ordering contract, but it does mean this fixture writes no reference state, so mention it in the PR body if a maintainer expects hashes.

### Pre-existing, noticed while reading (not from this PR)

`client/js/widgets/widget.js:2041` — `collections[getCollection(a.source)].filter(...)`: when `a.source` names a collection that doesn't exist, `getCollection()` pushes a problem and returns `null`, and `collections[null]` is `undefined`, so `.filter` throws a `TypeError` and aborts the whole routine instead of reporting a problem. Unrelated to this change, but trivially guardable while you're in here.

### Checked and fine

* `setDefaults(..., { keepOrder: false })` plus the branch at 2079 keeps existing saves byte-identical in outcome — no `FileUpdater`/compat work needed (`docs/compat.md`).
* Validator entry (`validate_gamefile.js:794`, `'boolean'`) and the `jeCommands` entry (`jsonedit.js:1392`) are both present, as CLAUDE.md requires; `keepOrder: false` as the inserted value matches how `SORT`'s `reverse`/`rearrange` buttons behave.
* The `'Moves'` additions in `jeLoggingRoutineStart`/`End` (`jsonedit.js:3053`/`3071`) are symmetric and only reachable while a `SWAPHANDS`-generated `Moves` frame is on top of `jeHTMLStack`, so no other routine's log rendering changes.
* `jeLoggingRoutineOperationSummary` is correctly emitted *after* the nested `Moves` operation ends, so `jeRoutineResult` is consumed by the outer `SWAPHANDS` frame rather than by `Moves`.
* Both TestCafe workflow matrices got the new fixture; nothing here touches the server, so no crash or state-corruption risk.

---

The `beta` label was already present and I left it in place: the only engine file touched is `client/js/widgets/widget.js`, where the new ordering is gated behind the opt-in `keepOrder` property (default `false`) and the surrounding refactor leaves existing games' observable outcome unchanged apart from the narrow removal-queue edge case in finding 1 — incidental enough for a beta build, and safe for players to test.

VTT-AI-Agent:
Thanks — all five points addressed in fb6c445e (`npm test` 44/44, `tests/testcafe/routines.js` 4/4 locally).

### 1. Default path resolved its collection eagerly — fixed, with a caveat worth recording

Each generated `MOVE` now gets its own routine call, and its collection is filled immediately before that call instead of in the build loop:

```js
for(const move of moves) {
  const collection = `hand of ${move.source.get('id')}`;
  const shadowed = collections[collection];
  collections[collection] = a.keepOrder
    ? move.contents.filter(w=>!w.isBeingRemoved)
    : widgetFilter(w=>move.contents.indexOf(w) != -1);
  try {
    await this.evaluateRoutine([ { func: 'MOVE', collection, to: move.to } ], variables, collections, (depth || 0) + 1, true);
  } finally {
    /* restore or delete `collection` */
  }
}
```

The hands themselves are still snapshotted up front — they have to be, or a hand picks up what the previous seat just handed it — but the widget *lookup* is back at move time, which is exactly when `getCollection()` did it before this PR. So the default path resolves identically to pre-PR again, `keepOrder` drops removed widgets too now, and the seat-named log labels survive. The `Moves` frame renders exactly as before (`jeLoggingRoutineStart` suppresses the header for each nested call while `Moves` is on the stack), verified in the DEBUG panel.

The caveat: I could not build a test that fails on b57e56ee, because the engine already stops one step later — `moveToHolder()` (`client/js/widgets/widget.js:2437`) returns immediately when `this.inRemovalQueue`. I built exactly your scenario (3 seats; `hand2`'s `enterRoutine` deletes `hand3`'s only card; `hand1`'s `enterRoutine` marks a witness widget) and b57e56ee produced the same state as fb6c445e: the dead card *was* in seat3's collection and `MOVE` did walk it, but `moveToHolder` bailed — `hand1`'s `enterRoutine` never ran, nothing was resurrected, nothing threw. So the finding is real in mechanism but not observable today; the change is worth having as the faithful restoration of the old timing rather than as a repair, and the new test documents the behaviour either way.

### 2. Restore was not exception-safe — fixed

`try { … } finally { … }` as above, and the shadow window is now a single `MOVE` instead of all of them. I kept the readable `hand of seatX` name rather than the reserved `$…` namespace: that label is what a creator reads in the DEBUG panel, and @96LawDawg asked for the seat names specifically. With the restore on the error path there is no window left in which the caller can observe the shadow.

### 3. ID round-trip — fixed

`contents` now collects the widget objects, so both branches use them directly and the default branch's filter is an identity check. The `Moves` start/end guards can't be merged (the move loop sits between them), but the summary guard now also holds the wording it builds.

### 4. Log wording — fixed

* `direction: 'random'` reads `hands in a random seat order by 1`
* when no seat has a valid hand, the summary is `no seat with a player has a valid hand, nothing to swap` and the empty `Moves` frame is skipped entirely

Same click as before, on three seats:

```
SWAPHANDS hands forward by 1
  Moves
    MOVE all widgets from 'hand of seat1' to 'seat2'
    MOVE all widgets from 'hand of seat2' to 'seat3'
    MOVE all widgets from 'hand of seat3' to 'seat1'
SWAPHANDS hands in a random seat order by 1
  Moves
    MOVE all widgets from 'hand of seat1' to 'seat3'
    MOVE all widgets from 'hand of seat3' to 'seat2'
    MOVE all widgets from 'hand of seat2' to 'seat1'
```

### 5. Tests

Added the fourth test described above (three seats, hands with `enterRoutine`s, one of which deletes a card mid-swap; it also covers the `interval` wrap-around with an odd number of seats).

The two gaps you listed I left open on purpose:

* **`childrenPerOwner` + `keepOrder`**: there is nothing to pin. A shared hand means both seats' `hand` is the same widget, so `MOVE` takes the `targetHand == source` branch (`widget.js:1644`) and only updates `owner` — neither mode reorders anything, in that case `keepOrder` is a no-op by construction.
* **`interval > 1` / `direction: 'backward'`**: these only pick a different target per move; the collection plumbing they share is what the four tests exercise.

The missing md5 reference state is now called out in the PR body, together with the reason (the assertion *is* the order, which a hash can't show when it drifts).

### Pre-existing `TypeError` on a missing `source` collection

Left alone deliberately: `TURN` carries the identical line (`widget.js:2184`), so the guard belongs in both places and I'd rather not widen this PR into a second operation. Happy to do it here if you'd prefer that over a follow-up.

96LawDawg (approved):
Tested.  Approved.  I should add `keepOrder` to the tutorial.

96LawDawg:
Tutorial added.  Ready to merge.

RaphaelAlvez (approved):
Tested
It's a bit weird having this and SHIFT as them both seem to support hand swaping

96LawDawg:
This function already exists. This is a small tweak that probably should have been in place to start with.  Going to merge this and evaluate if SHIFT can be rolled into SWAPHANDS.

PR #3053, opened 2026-07-26
Routine editor: touch multi-select, drops into IF/FOREACH blocks and an always-visible Routines section

VTT-AI-Agent:
Follow-up to the Discord feedback on #3034 (the routine editor on the expanded widget editor). Based on that branch, so the diff is only the follow-up work.

> The Ctrl+click multi select for drag and drop should also be supported on mobile. Use long press to select.
>
> Currently it's not possible to drag a function into If or Foreach
>
> O mobile the arrow, drag and delete buttons overlap with the inputs
>
> Remove the behavior section
>
> Automation section should always be visible to all widgets.
> Rename Events to Routines.
> The add Routines button should always be available

(Source: [Discord message from rapha195](https://discord.com/channels/770758631146782780/1527815541925085307/1530858547150065664))

## What changed

**Touch support for selecting and dragging.** A **long press** (500 ms) on an operation card toggles it in and out of the multi-selection — the touch equivalent of Ctrl+click — and the press that selects no longer opens the parameter popup underneath. Browsers do not turn a touch into a native HTML5 drag, so the drag handle now also follows the finger with pointer events and drops where it is released. Ctrl+click on a chip selects the card now instead of opening that chip's popup.

**Dragging into and out of IF/FOREACH blocks.** The drag is tracked in one module-level state instead of per routine level, and the drop target is resolved from the point under the pointer against whichever routine level owns it. So an operation can be dragged into a block (including an empty one, which highlights as a whole), back out into the parent routine, or into a sibling block. A block nested inside a dragged operation refuses the drop, because that would detach the routine from the widget. Cross-level moves re-render from the outermost editor, since the level the operation left no longer owns it.

**The move/delete icons no longer cover the summary.** They used to be absolutely positioned in the card's top right corner with a fixed 60px of reserved padding, while a card with the block arrows needs about twice that — so they sat on top of the text. They are now laid out next to the summary, which simply gets less room, and they wrap instead of overlapping. On coarse pointers they are bigger and further apart.

**The Automations section is always there.** It used to be rendered at the end of the generic "Other properties" list, which is skipped when a widget has no properties left to show — so widgets whose properties are all covered by curated inputs had no Automations section and no way to add a routine at all. It is now rendered for every widget as the last section (except piles, which are temporary and say so). "Events" is called **Routines** throughout: the group header, the *Add Routine* button and the popup.

**Removed the button's Behavior section**, which only said the clickRoutine has to be edited in the JSON editor and offered a button to jump there — the routine editor right below it now does that job. I read "remove the behavior section" as this one; the per-type Behavior groups that hold real inputs (dice roll times, holder drop options, scoreboard scores) are untouched. Happy to remove more if that was meant.

## Before / after

The overlap on a narrow screen, same routine before and after:

| before | after |
| --- | --- |
| before | after |

## Testing

- `npm test` — 85 pass, including new jest tests for dragging into an IF block, out of a FOREACH block, and a block refusing to receive an operation it lives inside.
- `tests/testcafe/editor.js` passes locally. (`tests/testcafe/functions.js` "Compute" fails on this branch with and without these changes, so it is not from here.)
- Drove the editor in Chromium: dragged an operation into an empty FOREACH block and out of an IF block with the mouse, long-pressed to select and dragged with synthetic touch events, and checked every widget type shows the Automations section with its add button.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3053/pr-test (or any other room on that server)

### DIFF
--- a/client/css/editor/controls/routine.css
+++ b/client/css/editor/controls/routine.css
@@ -9,8 +9,8 @@
    background shows as a rounded cutout that reaches the parent's right border
    while the operation cards inside keep their normal size */
 .routine-editor .routine-editor {
-  margin: 4px -60px 4px 12px;
-  padding: 5px 25px 5px 5px;
+  margin: 4px -8px 4px 12px;
+  padding: 5px;
   background: var(--backgroundColor);
   border-radius: 10px 0 0 10px;
 }
@@ -20,9 +20,22 @@
   background: var(--backgroundHighlightColor1);
   border-left: 3px solid var(--VTTblue);
   border-radius: 4px 4px 4px 0;
-  padding: 5px 60px 5px 8px; /* room on the right for the move/delete icons */
+  padding: 5px 8px;
   line-height: 1.8;
   font-size: 13px;
+}
+
+/* the summary and the move/delete icons sit next to each other so the icons
+   never cover the text, however narrow the editor gets */
+.routine-editor-operation-header {
+  display: flex;
+  align-items: flex-start;
+  gap: 4px;
+}
+
+.routine-editor-operation-body {
+  flex: 1;
+  min-width: 0;
 }
 
 .routine-editor-operation-parameter {
@@ -120,11 +133,10 @@ body.edit.darkMode .routine-editor-parameter-widget {
   vertical-align: middle;
 }
 
-/* the raw-JSON button in the top right of an expanded operation */
+/* the raw-JSON button next to the summary of an expanded operation */
 .routine-editor-operation-json {
-  position: absolute;
-  top: 5px;
-  right: 62px;
+  flex-shrink: 0;
+  margin-top: 3px;
   font-size: 15px;
   color: gray;
   cursor: pointer;
@@ -167,10 +179,12 @@ body.edit.darkMode .routine-editor-parameter-missing {
 }
 
 .routine-editor-operation-buttons {
-  position: absolute;
-  top: 5px;
-  right: 4px;
+  flex-shrink: 0;
   display: flex;
+  flex-wrap: wrap;
+  justify-content: flex-end;
+  max-width: 45%;
+  margin-top: 3px;
   gap: 3px;
 }
 
@@ -195,9 +209,23 @@ body.edit.darkMode .routine-editor-parameter-missing {
   color: var(--VTTblue);
 }
 
-/* drag-and-drop reordering: a grip to grab, Ctrl-click to select several cards */
+/* touch needs bigger, further apart targets than a mouse pointer */
+@media (pointer: coarse) {
+  .routine-editor-operation-buttons {
+    gap: 9px;
+  }
+
+  .routine-editor-operation-buttons .material-symbols,
+  .routine-editor-operation-buttons .routine-editor-block-arrow {
+    font-size: 19px;
+  }
+}
+
+/* drag-and-drop reordering: a grip to grab, Ctrl-click (long press on touch) to
+   select several cards; touch-action keeps a drag from scrolling the sidebar */
 .routine-editor-operation-buttons .routine-editor-drag-handle {
   cursor: grab;
+  touch-action: none;
 }
 
 .routine-editor-operation-selected {
@@ -216,6 +244,12 @@ body.edit.darkMode .routine-editor-parameter-missing {
 
 .routine-editor-drop-after {
   box-shadow: 0 3px 0 -1px var(--VTTblue);
+}
+
+/* dropping into a block that has no operation card to aim at yet */
+.routine-editor-drop-into {
+  outline: 2px dashed var(--VTTblue);
+  outline-offset: -2px;
 }
 
 .routine-editor-empty,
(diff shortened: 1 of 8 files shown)


VTT-AI-Agent:
Pushed 57d7cebc for the follow-up feedback:

> To be implemented on #3034
>
> Implement the drag of functions in and out of thenRoutine, elseRotine and loolRoutine.
>
> Make the automation section always visible. To be displayed before "other properties"
>
> Use "routine" instead of event in that section

(Source: [Discord message from rapha195](https://discord.com/channels/770758631146782780/1527815541925085307/1530913999867216024))

**Dragging into a block really works now.** The previous commit resolved the drop against the right routine level, but a block that the operation does not have yet (a freshly added `IF` has no `thenRoutine` key at all, a `FOREACH` no `loopRoutine`) is rendered from a detached array — that is what keeps empty blocks out of the saved JSON. A drop wrote into that array and nothing ever attached it, so the operation disappeared: removed from the routine it came from, invisible in the block. The drop now writes the array back into the operation, the same thing the ↰/↲ arrows already did. Verified in Chromium on a `clickRoutine` of `FLIP`, `IF`, `FOREACH`, dragging the `FLIP` around:

| drag | resulting clickRoutine |
| --- | --- |
| into the empty `thenRoutine` | `[{IF, thenRoutine:[FLIP]}, {FOREACH}]` |
| from there into the empty `loopRoutine` | `[{IF, thenRoutine:[]}, {FOREACH, loopRoutine:[FLIP]}]` |
| back out onto the top level | `[{FLIP}, {IF…}, {FOREACH…}]` |
| into the `elseRoutine` | `[{IF, thenRoutine:[], elseRoutine:[FLIP]}, {FOREACH…}]` |

Before the fix, the first of those drags left `[{IF, operand1:1}, {FOREACH}]` — the `FLIP` was gone.

**Automations moved up**, in front of the generic *Other properties* list, so the routines sit with the curated inputs instead of below the raw property dump: section order

It is still rendered for every widget type except piles (they say "pile properties cannot be edited here" and point at the deck).

**Wording:** the section already said *Routines* everywhere after the previous commit; two info texts still said "use a property-specific event like fooChangeRoutine", they now say "routine" too. If you meant the *Automations* header itself should be called *Routines*, say the word and I'll rename it.

`npm test` (87 pass, two new tests for the block attaching) and `tests/testcafe/editor.js` pass locally.

VTT-AI-Agent:
Pushed another follow-up from Discord:

> The widget picker should use the one used in the rest of the UI.
>
> Also make so that when there is a type filter applied in this widget picker the filter also applies to the click to select function. So for example if the user set the type holder in the item pucker and then click on a holder with cards on top, it will only select the holder underneath.

(Source: [Discord message from rapha195](https://discord.com/channels/770758631146782780/1527815541925085307/1530922769204776971))

**The routine editor now uses the picker the rest of the editor uses.** Its widget/holder/collection parameters opened a table with *Add Widgets* / *Start Fresh* / *Use These Widgets* that drove the old custom-selection mode. That table is gone; the parameter popup now shows the same searchable, type-filtered popout with *Pick in the room* that the parent, seat and hand inputs use. To get there, the picker moved out of `sidebar/properties.js` into `controls/widgetselection.js` (which is where the old table lived), so both sides call the same code — `properties.js` only lost the copy.

**The type filter applies to picking in the room.** A click in the room hits the top-most widget, which is usually not the one you mean: a holder is covered by the cards lying on it. The picker now resolves a click up the parent chain to the first widget it accepts, so with the type set to *holder*, clicking a card selects the holder underneath it. Without a type filter, sidebar pickers keep taking exactly what was clicked; the routine picker keeps its old behavior of looking past a card or pile.

Two things that fell out of sharing the picker:

* A picker that collects several widgets (the routine parameters, the seat inputs) now **stays active after each click**, so several widgets can be picked in a row — the old table could do that and one-shot picking would have been a regression.
* Picked widgets **sort to the top of the id list**, so a value that is already set stays visible (and removable) when the list is cut off at 50 entries.

Also: closing the *Edit Widgets* module now ends a running pick, so it can't leave the room in picking mode.

routine widget picker

*The MOVE `from` parameter with the type filter set to Holder — `drawPile` got selected by clicking the card lying on it.*

### Testing

- `npm test` — 92 pass, including new jest tests for the type filter resolving a click to the widget underneath, `resolveCovering`, a broken parent chain not looping, collecting several picks in a row, and the routine popup applying its selection.
- `tests/testcafe/editor.js` passes locally. (`tests/testcafe/functions.js` "Compute" still fails on this branch with and without the change.)
- Drove it in Chromium: picked a holder through a card with the filter set, applied it to the routine, checked the parent and seat pickers in the sidebar still behave as before, and that closing the popup or the module clears the picking cursor.

One leftover worth a separate PR: `startCustomSelection`/`endCustomSelection` in `editor/selection.js` have no caller left now that the table is gone. I left them in place to keep this diff about the picker.

PR #3057, opened 2026-07-27
Dynamic toolbar breakpoints and a scrollable toolbar instead of hidden buttons

VTT-AI-Agent:
## Goal

The three toolbar layouts (wide vertical, narrow vertical / `aspectTooGood` and horizontal) each had a series of **hardcoded** viewport breakpoints: first a few steps that make the layout more compact (less padding, smaller tabs, no logo/dividers), then a cascade of `@media` rules that hid buttons one by one. Every added button meant re-tuning ~20 hardcoded pixel limits, and getting them wrong pushes something off screen.

## What this PR does

**1. The compact steps are now figured out dynamically.**

The `@media` breakpoints are replaced by four cumulative compaction levels (`body.toolbarCompact1` … `toolbarCompact4`), and `updateToolbarLayout()` (called from `setScale()`) applies the *lowest* level at which the toolbar content actually fits — it measures the toolbar and its children instead of guessing from the viewport size. So nothing is pushed off screen by even one pixel, and adding a button needs no CSS breakpoint work at all.

What each level does per layout (unchanged rules, just triggered by measurement now):

| level | wide vertical | narrow vertical / aspectTooGood | horizontal |
|---|---|---|---|
| 1 | tighter paddings/margins | tighter paddings/margins | tab labels hidden |
| 2 | dividers hidden | dividers hidden | tighter paddings/margins |
| 3 | tabs 30px high | tabs 30px high | dividers hidden |
| 4 | logo hidden | – | – |

As a side effect the layouts now stay roomy longer than before (e.g. the horizontal toolbar keeps its tab labels as long as they fit instead of below the hardcoded 950px, and the vertical toolbar keeps its dividers at 750x400).

**2. Buttons are no longer hidden — the toolbar scrolls instead.**

All the "hide `#aboutButton` below 230px" style rules are gone. When even the most compact level does not fit, `body.toolbarOverflow` makes the toolbar scrollable along its axis (native scrolling, scrollbars hidden) and a chevron with a soft fade appears at the end(s) that have more content, so it is clear that there is more. Every button stays reachable at any viewport size.

Two details that come with the scroll container: hover tooltips would be clipped by it, so they are suppressed while the toolbar scrolls (in the wide toolbar they are inside the toolbar and stay visible), and the sound/zoom popups are positioned relative to the viewport by `positionToolbarPopups()` while scrolling is active so they stay usable.

## Screenshots

| narrow vertical (420x220) | horizontal (300x300) | wide (1600x220) |
|---|---|---|
| vertical | horizontal | wide |

## Testing

- `npm test` (jest) passes.
- TestCafe `editor.js` and `publiclibrary-1…4.js` pass locally (chromium headless).
- Checked 22 viewport sizes across all four layouts with Playwright: no child of `#toolbar` sticks out of it at any size, the compaction level rises step by step, and scrolling with both hint arrows works in the vertical and the horizontal toolbar. Toolbar rendering at the common sizes (1920x1080, 1600x900, 1200x900, 1000x800, 1600x1000, 800x500, 1600x380, 400x400) is pixel-identical to `main`.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3057/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

Nice direction — measuring instead of guessing is clearly the right fix, and I could not make the toolbar push anything off screen at any of the sizes I tried. The one thing I'd want fixed before this is considered done: in the **horizontal** toolbar the overflow is only scrollable horizontally, and a plain vertical mouse wheel does not scroll a horizontal-only scroller — so on a narrow *desktop* window the buttons that scroll out are effectively unreachable for a mouse user, which is the exact failure mode the PR sets out to remove.

## Findings

### 1. Horizontal toolbar: overflowing buttons are not reachable with a normal mouse wheel

`client/css/layout.css` (`body.horizontalToolbar.toolbarOverflow #toolbar { overflow-x: auto }`)

Verified in Chromium against this branch at 300x300 (`horizontalToolbar`, `toolbarOverflow`, `scrollWidth` 361 vs `clientWidth` 300, so ~60px of buttons are off the end):

| input over the toolbar | resulting `scrollLeft` |
|---|---|
| wheel `deltaY` 300 (normal mouse wheel) | **0** |
| wheel `deltaX` 200 (trackpad / tilt wheel) | 61 |
| shift + wheel | 61 |
| 15x Tab | 61 |

So with a plain wheel mouse, `#hideToolbarButton` / `#zoom2xButton` etc. simply cannot be brought into view (I also tried `overflow-y: hidden` and `visible` instead of `clip` — no difference; it's the browser's wheel-axis behavior, not the `clip`). Touch drag and trackpads are fine, and the horizontal layout is mostly a portrait/mobile layout, but it is also what you get in a narrow desktop window.

The cheapest fix is the affordance that is already there: drop `pointer-events: none` from `#toolbar::before/::after` in the overflow case and make them scroll by ~one page on click (`toolbar.scrollBy({left: …, behavior:'smooth'})`). That fixes discoverability at the same time — right now the chevrons look like buttons but are inert.

### 2. `positionToolbarPopups()` clamps in the wrong order — popup can end up at a negative offset

`client/js/main.js` (the `horizontal` branch)

```js
popup.style.left = Math.min(Math.max(0, anchorRect.left + 32 - popup.offsetWidth), window.innerWidth - popup.offsetWidth) + 'px';
```

`Math.min(…)` is applied *after* `Math.max(0, …)`, so when the popup is wider than the viewport, `window.innerWidth - popup.offsetWidth` is negative and wins. Reproduced: horizontal toolbar at a 140px-wide viewport, sound popup (167px) lands at `left: -27px`, i.e. the mute button is clipped off the left edge — the one control the popup exists for. At 170px it's fine (`left: 3`).

`Math.max(0, Math.min(anchorRect.left + 32 - popup.offsetWidth, window.innerWidth - popup.offsetWidth))` keeps the visible end anchored instead. Same ordering issue in the vertical branch's `top` (only reachable below a ~35px-tall viewport, so cosmetic there, but worth making symmetric).

### 3. The early return plus `flex-shrink: 0` can leave a stale layout that spills out of the toolbar

`client/js/main.js` (`if(!toolbar.getClientRects().length) return;`) together with the new `#toolbar > * { flex-shrink: 0 }`.

Before this PR the breakpoints were pure CSS, so they applied whether or not the toolbar was visible. Now, if a resize happens while `#toolbar` is `display: none` (it is in `body.edit` and `body.jsonEdit`), the compaction/overflow classes are never recomputed — and because children can no longer shrink, showing the toolbar again without a further `setScale()` puts the buttons *outside* the toolbar, drawn over the room, with no scrolling. Reproduced by hand: hide `#toolbar`, resize 1200x800 → 260x220, show it again → body still has only `horizontalToolbar`, `overflow-y: visible`, and 12 of the children stick out past the toolbar's right edge.

As far as I can tell every current hide/show path does end in `setScale()` (`toggleEditMode()`, `client/js/editor/layout.js`, `client/js/editor/sidebarModule.js`), so I could not trigger it through the UI — but `jeToggle()` itself doesn't, so it is one refactor away from being real. A `ResizeObserver` on `#toolbar` (which also fires when it becomes visible again) would make this self-correcting and would additionally cover content changes you don't currently hook, instead of the manual `updateToolbarLayout()` call added in `client/js/overlays/players.js`.

### 4. Every button label disappears on the small screens that need them most

`body.toolbarOverflow:not(.wideToolbar) .toolbarButton:hover .tooltip { display: none; }`

`toolbarOverflow` is by definition the smallest viewports, where the buttons are icon-only — so hover names are lost exactly where the icons are least self-explanatory, and this is a regression from `main` (there the buttons were dropped, but the ones that remained kept their tooltips). Since you already solved this problem for `#options`/`#zoomControls` by switching them to `position: fixed` and placing them from JS, the same trick would work for the hovered tooltip and would avoid the tradeoff altogether.

### 5. No test guards the new logic

The measurement loop is now the only thing standing between a new toolbar button and content off screen, and nothing in `tests/` exercises it — the state-hash TestCafe tests can't see toolbar layout. A small TestCafe case that, at two or three viewport sizes, asserts no `#toolbar` child's bounding box exceeds the toolbar's (plus that `toolbarOverflow` appears at a tiny size) would lock in the invariant the PR is built around and is cheap to write.

## Things I checked that look fine

- Compaction is monotonic and terminates across 10 viewport sizes in all four layouts; no oscillation between levels.
- Cost of the up-to-5-pass measurement is ~1.1ms per `setScale()` here, and `setScale()` is not on the pan/zoom path (`setPan()`/`setZoomLevel()` don't call it), so no interaction jank.
- `#options` reposition on resize-with-popup-open, and on the `#toolbar` click (the button handlers run first in the target phase, so the `hidden` class is already up to date when `positionToolbarPopups()` reads it).
- `.logoDivider` correctly replaces the `:nth-child(2)` selector, and the `toolbarCompact2` divider rule out-specifies the `body.wideToolbar … .logoDivider { display: block }` rule, so the logo divider still hides at level 2.
- `npm test` passes on the branch.

I added the `beta` label: the PR touches no engine file (only `client/css/layout.css`, `client/js/main.js`, `client/js/overlays/players.js`, `client/room.html`), it cannot affect room state or the server, and the open points above are UI-level, so it is safe and useful for players to try.

VTT-AI-Agent:
Thanks — all five points addressed in f01a89ca.

**1. Overflowing buttons unreachable with a plain mouse wheel** — fixed, and I took both halves of the suggestion:

- the arrows are no longer `pointer-events: none`. They are pseudo elements, so a click on one targets `#toolbar` itself; the click handler recognizes that and scrolls by ~80% of the visible length with `behavior: 'smooth'`. They also got `cursor: pointer`.
- a `wheel` handler translates `deltaY` into scrolling along the toolbar's axis while the horizontal toolbar overflows (`deltaX` is left to the browser, and `deltaMode` 1 is scaled since some mice report lines).

Verified in Chromium at 300x300 (`scrollWidth` 361 / `clientWidth` 300): forward-arrow click `scrollLeft` 0 → 61, back-arrow click 61 → 0, wheel `deltaY` 300 → 61, `deltaY` -300 → 0.

**2. Clamp order** — fixed, and both branches now go through one helper, so the ordering is symmetric:

```js
element.style.left = Math.max(0, Math.min(left, window.innerWidth - element.offsetWidth)) + 'px';
element.style.top  = Math.max(0, Math.min(top,  window.innerHeight - element.offsetHeight)) + 'px';
```

At a 160px-wide viewport the sound popup is now at `left: 0` (was negative), so the mute button stays on screen.

**3. Stale layout while the toolbar is hidden** — fixed with a `ResizeObserver` on `#toolbar`, which also fires when it becomes visible again. Reproducing your case (hide `#toolbar`, resize 1200x800 → 260x220, show it again) now ends up with `toolbarOverflow` and a scrollable toolbar instead of buttons drawn over the room. I kept the `updateToolbarLayout()` call in `setScale()` (it runs synchronously with the layout-class change, so there is no frame with the wrong layout) and the one in `players.js` (a wider player count changes the button, not the toolbar box, so the observer would not see it).

**4. Tooltips on small screens** — fixed the way you suggested: instead of hiding them, `body.toolbarOverflow:not(.wideToolbar) .toolbarButton .tooltip` is now `position: fixed` and `positionToolbarTooltip()` places the hovered button's tooltip above the toolbar (horizontal) or next to it (vertical), clamped to the viewport, and repositions it while scrolling. Inline styles are removed again on mouse-out or when the toolbar stops overflowing.

horizontal vertical

**5. No test** — added `tests/testcafe/toolbar.js`. At five window sizes it asserts that all toolbar buttons are displayed and that either nothing reaches outside of the toolbar, or the toolbar really is scrollable and scrolling it to the end brings the last buttons inside — i.e. the invariant the PR is built on, independent of where the breakpoints happen to land. Passes locally together with `editor.js` (5 passed) and `npm test` is green.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review). Measuring instead of guessing breakpoints is a clear win — no button is unreachable any more, and the compact steps kick in later than the old hardcoded media queries, so the toolbar stays roomy longer. The one thing I'd fix before merging is the **scroll chevrons: they are drawn *on top of* the buttons instead of getting space of their own**, which both looks like a glyph mash and makes the "back" arrow un-clickable (the click lands on the button underneath).

Checked live on this branch (server on a local port, Playwright, real room) at 20+ viewport sizes across all four layouts, plus hit-testing of the arrows. Screenshots below are from that run.

---

### Big layout improvements

**1. The scroll chevrons overlap the buttons — including the click target.**
`client/css/layout.css:493-541` gives `#toolbar::before/::after` negative margins (`margin-bottom: -24px` etc.) so they "keep them from taking up space of their own". The consequence is that the arrow is always painted over a button, and the 24px `linear-gradient(..., 70%, transparent)` is too short to visually separate them:

| horizontal, scrolled to start | horizontal, scrolled to end |
|---|---|
| start | end |

On the left the `›` sits right on the speaker icon and reads as one broken symbol; on the right the `‹` lands on the white active-tab pill, slicing the pill in half and putting a white glyph on a near-white background.

Worse, it isn't just cosmetic: `scrollToolbarByArrow` only runs when `e.target == this` (`client/js/main.js:698`), but because the pseudo-element is not a hit target, the topmost element at the arrow position is normally a button. Hit-testing the vertical overflow layout (420×220) at every scroll offset returns `statesButton` / `activeGameButton` / `playersButton` under the *back* arrow — and an instrumented click there reported `target: statesButton` with `scrollTop` unchanged (the forward arrow, which happens to sit over the free area after the last button, does scroll). So an element styled `cursor: pointer` (`layout.css:504`) does nothing on click for most positions and instead opens the Game Shelf. A first-time user clicking the arrow to see "what else is up there" gets an unrelated overlay.

Suggested fix, small and in keeping with the current design: let the arrows occupy real space instead of negative margin — e.g. `body.toolbarOverflow #toolbar { padding-top/bottom: 24px }` (`padding-left/right` for the horizontal one) and drop the negative margins. The arrows then sit in their own strip, the gradient can be a clean solid `var(--toolbarColor)`, no icon is ever half-covered, and the click always reaches the toolbar. If you'd rather not give up the 48px, at minimum extend the fade (e.g. 32-40px, `var(--toolbarColor)` to `transparent` over the full length) so the button under the arrow reads as "faded out" rather than "collided with".

**2. The horizontal toolbar drops *all* tab labels in one step, leaving half the bar empty.**
Level 1 for the horizontal layout is "tab labels hidden", so the moment the labels no longer fit, five labelled tabs become five bare icons — at 900px wide roughly half the toolbar is empty space:

horizontal 900

Now that the level is chosen by measurement, this is exactly the place to add a cheaper intermediate step: shrink the label (`body.horizontalToolbar:not(.toolbarCompact1) .toolbarTab .tooltip { font-size: 20px }` → an extra level at 15-16px with `padding: 0 6px`) before removing labels entirely. Labels would survive several hundred pixels further down, which matters a lot for newcomers (see the readability section). The measuring loop already handles more levels — it's one more `toolbarCompact*` step.

**3. Hidden scrollbar + subtle chevron is a weak "there is more" signal.**
`scrollbar-width: none` (`layout.css:479`) plus a 24px chevron is all the user gets. It works once you notice it, but on a desktop with a mouse there's no other cue. Consider either keeping a thin scrollbar in the overflow state, or giving the arrows a slightly larger hit area/contrast (see #1 — once they have their own strip you can safely make them a full 24px tall row with a solid background). A gentle one-time bounce of the toolbar when `toolbarOverflow` first becomes true would also do it without adding chrome.

- **Tooltip covers the popup it belongs to.** In overflow mode both the popup and the hover tooltip are `position: fixed`; opening the zoom popup while the pointer stays on the button puts "1.0x Zoom" across the right end of the slider: zoom — suppressing the tooltip while its own popup is open (`#options`/`#zoomControls` not hidden) would fix it.
- **Wide toolbar: the chevron chops a label mid-word.** At 1600×200 the "Sound" row is cut through the middle of the text by the arrow strip: wide. Same root cause as #1; the padding fix also makes the last row end cleanly.
- **Arrow colour is hardcoded `#fff` (`layout.css:502`)** while the surrounding rules use `var(--toolbarColor)` for the background. Fine for the current blue toolbar, but the divider next to it uses `#fff8`, so consider matching the button icon colour source rather than a literal, in case the toolbar is ever themed.
- **No scroll padding at the ends.** At the end of the vertical scroll the fullscreen button sits flush against the toolbar edge (`z-v-end`); a couple of pixels of `scroll-padding`/end padding would make the last item look intentional rather than cut off.
- **Touch:** the back/forward arrows are 24px — below the ~44px touch target the rest of the toolbar uses. Native swipe covers it, but if the arrows are meant to be tappable they should be as tall/wide as a button.

- The scrolling behaviour itself is *more* understandable than what it replaces (buttons silently vanishing), and the tooltips are correctly re-positioned outside the scroll container — "Players: 1" renders next to the button in both the vertical and the horizontal overflow layout: tooltip — that part is good.
- The chevron alone doesn't say *what* is hidden. `keyboard_arrow_down` over a partially faded icon reads as decoration to someone who has never seen the app. Once the arrows have their own strip (#1), consider the standard "more" affordance instead — a `⋯` at the end that scrolls, or a small "more" tooltip on hover like the other buttons have.
- On narrow screens the toolbar is now *only* icons (Game Shelf, Active Game, Players, About, Edit Mode, Sound, Zoom, Fullscreen), and on touch there is no hover to reveal the tooltip — the "person + gear" and "person in a frame" icons are genuinely hard to tell apart at 20px. This isn't introduced by this PR, but improvement #2 (keep labels longer) directly reduces how often a newcomer lands in that state.
- Small point in favour: the `hideToolbarButton` (aspectTooGood layout) is now always reachable by scrolling instead of disappearing below 295px — but it's the *last* item, so on a phone in portrait it's off screen until you scroll. If any button deserves to be pinned outside the scroll area, it's probably that one.

VTT-AI-Agent:
Thanks — the arrow strip and the label step were both worth doing. Addressed in 22c5b445.

**1. Arrows painted on top of the buttons / back arrow not clickable** — fixed, both halves:

- The arrows no longer use negative margins. They are flex items now, so they get a `--toolbarArrowSize` (24px) strip of their own at each end of the scroll axis, and the toolbar's own padding along that axis is dropped in the overflow state so the strip *is* the padding. The gradient is gone with them: the strip is a solid `var(--toolbarColor)`, so content that scrolls underneath is clipped cleanly instead of showing through a half-transparent arrow. They also stay in the layout when there is nothing to scroll to in their direction (`visibility` instead of `display`), so the buttons never jump while scrolling.
- The click is now caught on `#toolbar` in the **capture** phase and swallowed with `stopPropagation()` when it lands in a visible arrow strip, so it can never reach a button that scrolled underneath — no more Game Shelf when you meant to scroll. `scrollToolbarByArrow` also checks `toolbarScrollBack`/`toolbarScrollForward` first, so a click in the strip of an arrow that isn't shown still reaches the button below it.

| horizontal, scrolled to the end | vertical, mid-scroll |
|---|---|
| h | v |

**2. All tab labels dropped in one step** — fixed. The horizontal layout got the intermediate level you suggested: level 1 shrinks the labels (20px → 15px, tab padding 16px → 10px), level 2 removes them, and the old levels 2/3 moved to 3/4 (the horizontal layout was not using level 4). Labels now survive from ~1000px down to ~760px instead of disappearing at ~1000px:


<sub>900px wide — all five labels still there, was five bare icons before.</sub>

**3. Weak "there is more" signal** — partly. The arrows are much more present now that they sit in a solid strip of their own instead of being a half-faded glyph on an icon, and they actually work on click. I did not add a scrollbar back: on the 44px toolbar it would eat a noticeable slice of the cross axis and it can't be styled to fit the blue bar, so I'd rather not trade button space for it.

**4. Tooltip covers the popup it belongs to** — fixed. `positionToolbarTooltip()` now hides the tooltip of `#optionsButton`/`#zoom2xButton` while that button's popup is open (in every layout except the wide one, where the "tooltip" is the permanent label):

popup

**5. Chevron chops a label mid-word (wide toolbar)** — mostly fixed. The strip from #1 means the arrow never overlaps a label any more, and I added `scroll-snap-type: … proximity` + `scroll-padding` so that scrolling comes to rest with a button aligned to the edge of the strip instead of half behind it (verified: a wheel notch on the horizontal toolbar lands on 53 instead of 20, with nothing cut at the strip edge). What remains is the very end of the scroll range, where the resting position is fixed by the content length and the row there is clipped by the strip — that is normal scroll-container clipping and now a clean straight edge rather than a glyph collision:

wide

**6. Hardcoded `#fff`** — fixed. `#toolbar` now defines `--toolbarTextColor` next to `--toolbarColor`, used by the arrows and by the tooltip/popup rule that had the same literal.

**7. No end padding** — fixed by #1: the arrow strips are 24px of end padding at both ends, so the last button no longer sits flush against the toolbar edge.

**8. 24px arrows are below touch-target size** — not changed, deliberately. On touch the native gesture is a swipe, which works and needs no target at all; the arrows are the mouse affordance. Making them 44px would cost 88px of a 320px-wide phone toolbar, which is about two buttons. The strip does span the full cross axis (44×24 vertical, 24×44 horizontal), so it is not a 24×24 target.

**9. "The chevron doesn't say *what* is hidden" / pinning `#hideToolbarButton`** — left alone for now. Both are worth doing but change what the toolbar *is* (a "more" menu, or a button outside the scroll area) rather than how it degrades, so they feel like a separate PR to me rather than something to squeeze into this one.

Also added a second TestCafe case to `tests/testcafe/toolbar.js` that scrolls the toolbar to its end and clicks the back arrow, asserting that it scrolls back and that no overlay opens — the regression you found. `npm test` (44) and `testcafe toolbar.js + editor.js` (6) pass locally.

VTT-AI-Agent:
🚨 **Crash likely caused by this PR**

A client crash report came in from the beta server (Firefox), and it points at the toolbar layout code added here. The reporting client ran the beta build with this PR's toolbar compaction active, and the crash happened while the browser window was being resized.

**What happens:** `new ResizeObserver(updateToolbarLayout).observe($('#toolbar'))` (client/js/main.js) observes `#toolbar` with a callback that changes the size of that very element — `updateToolbarLayout()` toggles `body.toolbarCompact1…4` and `body.toolbarOverflow`, i.e. exactly the classes whose CSS changes the toolbar's paddings, tab heights and dividers. When an observer callback resizes the observed element, the browser aborts the delivery loop and reports **"ResizeObserver loop completed with undelivered notifications"** as an error event on `window`.

**Reproduced locally** (headless Chromium, local server, empty room, resizing the viewport through 10 sizes: 900x700 → 600x700 → 420x700 → 300x300 → 700x300 → 1100x300 → 1400x900 → 500x500 → 350x650 → 1000x250):

* on this PR's branch: `ResizeObserver loop completed with undelivered notifications.`
* on `main` with the identical script: no error events at all.

That error event carries no `Error` object, and the client error reporter in `client/js/tracing.js` currently dies on `error.message` when that happens — which is why the submitted report contains only the reporter's own `can't access property "message", e is undefined` TypeError and not the ResizeObserver message. I opened #3058 to fix the reporter. Note that once #3058 is in, this loop will no longer be swallowed silently: every occurrence will pop up the fatal client error overlay, so it needs fixing here as well.

**Possible fixes** (whatever fits the design best):

* make `updateToolbarLayout()` re-entrancy safe — e.g. ignore observer callbacks that were triggered by the layout pass itself, or return early when the resulting class set is unchanged (currently the loop always re-toggles all four classes, so even a no-op pass writes to `classList`);
* defer the observer callback to a `requestAnimationFrame` so the resizes it causes land in the next frame;
* or observe something the callback cannot resize (the room area / a wrapper) instead of `#toolbar` itself.

VTT-AI-Agent:
Confirmed and fixed in 42913142.

I reproduced it exactly as described (headless Chromium, local server, the same ten viewport sizes): on the previous state of this branch the window got `ResizeObserver loop completed with undelivered notifications.`, and `window.onerror` then ran into a second error inside `errorHandler` (`tracing.js`) because browsers pass `null` as the error object for that event — so the client either shows the crash overlay or throws while trying to.

**The fix:** the `ResizeObserver` no longer calls `updateToolbarLayout()` directly. It schedules it for the next animation frame (coalesced, so a burst of resizes still causes one relayout):

```js
new ResizeObserver(scheduleToolbarLayoutUpdate).observe($('#toolbar'));
```

That moves the class toggling out of the observation-delivery loop, so resizing the toolbar from the callback is a normal resize that gets picked up in the next frame instead of an undelivered notification. Everything else is unchanged — `setScale()` and the player-count change still call `updateToolbarLayout()` synchronously, so the layout is still applied in the same frame as the window resize; only the observer takes the deferred path.

**Verified:**
* the ten-size resize run: `window errors: []`, no page errors (it printed the `ResizeObserver loop …` message before the fix).
* a 40-size resize sweep (300–1500 × 250–950), both with 350 ms and with 30 ms between the resizes: no errors, and at every size the `toolbarOverflow` class still matches whether the toolbar actually overflows — so the deferral does not leave the layout in a stale state.
* `npm test` and `tests/testcafe/toolbar.js` pass.

I did not add a TestCafe regression test for it: TestCafe does fail a test when this error fires (I checked that with a synthetic observer loop), but its `resizeWindow` resizes the real window, and there the `resize` event settles the layout before the observer is delivered, so the loop never happens — the test would have passed on the broken code as well. A test that cannot fail seemed worse than none.

VTT-AI-Agent (commented):
## TL;DR

The earlier toolbar accessibility regressions are fixed, but two issues remain: the new TestCafe coverage is not wired into CI at all, and the dynamic fit decision depends on resize direction because it measures while the existing overflow arrow still occupies layout space.

## Findings

### 1. The new toolbar tests never run in CI

`tests/testcafe/toolbar.js:47` adds the regression coverage this change needs, but both TestCafe workflows use an explicit file matrix and neither includes `toolbar.js` (`.github/workflows/testcafe-chrome.yml:20-29`, `.github/workflows/testcafe-firefox.yml:20-29`). The current check list confirms that CI only ran `editor.js`, the compute shards, and the public-library shards. As a result, the tests added by this PR can regress or fail without affecting the PR checks. Please add `toolbar.js` to both matrices (or otherwise include it in the suite those jobs execute).

### 2. Overflow/compaction has resize-direction hysteresis

`client/js/main.js:247-258` tests each compaction level without first clearing `toolbarOverflow`. When that class is already present, `client/css/layout.css:519-528` makes `#toolbar::before` a 24px flex item, so `toolbarContentFits()` measures the real buttons after they have been displaced by the old overflow affordance. The result is that identical viewport dimensions can receive different layouts based only on their resize history.

I reproduced this on the PR preview in Chromium with a 300px-high horizontal toolbar:

- Growing from 300px wide, `toolbarOverflow` remained through 372px and cleared at 374px.
- Shrinking from 410px wide, the same compact level fit without overflow through 360px, and overflow appeared only at 358px.
- At 360px, therefore, the toolbar either has scrolling/chevrons or fits normally depending on which direction the window was resized.

This contradicts the stated “lowest level at which the content fits” behavior and leaves an unnecessary scroller in a band of sizes. Clear the overflow state (or otherwise exclude its pseudo-elements) before measuring the candidate compaction levels, then re-enable it only if level 4 genuinely does not fit.

The `beta` label was already present; I left it in place because this UI-only change touches no engine files and the remaining issues do not risk game corruption or server crashes.

VTT-AI-Agent:
Both findings were valid, fixed in 48493c4a.

**1. `toolbar.js` never ran in CI** — added to the `testfile` matrix of both `.github/workflows/testcafe-chrome.yml` and `.github/workflows/testcafe-firefox.yml`, right after `editor.js`, so this PR's own checks now include it.

**2. Resize-direction hysteresis** — fixed by clearing `toolbarOverflow` before the measuring loop, so every pass measures the buttons in the same, arrow-free layout:

```js
// the scroll arrows are flex items that take space away from the buttons, so measuring while
// they are still there would make the result depend on the previous state - and with that on
// the direction the window was resized in
toggleClass($('body'), 'toolbarOverflow', false);
```

The class is re-applied at the end of the function as before, so an actual overflow still gets its arrows; only the measurement is now stateless.

Reproduced and verified in Chromium at 300px height, sweeping the width from 300 to 420 in 2px steps upwards and then downwards, recording `toolbarOverflow` and the compaction level at each width:

* before: 7 widths (362–374) ended up with a different layout depending on the direction — overflow cleared at 376 when growing, but only appeared at 362 when shrinking, exactly the band you measured;
* after: 0 differing widths, the transition is at 362 in both directions.

I also added a third TestCafe case, *"The toolbar layout does not depend on the direction the window was resized in"*, which walks 360→440px up and back down at 300px height and asserts the layout classes are identical at each size. It fails on the previous state of the branch (`360x300: … expected 'horizontalToolbar toolbarCompact1…4' to deeply equal '… toolbarOverflow'`) and passes now. `toolbarScrollBack`/`toolbarScrollForward` are excluded from the comparison on purpose — they reflect the current scroll position, which legitimately survives a resize.

`npm test` (44) and `testcafe toolbar.js + editor.js` (7) pass locally.

VTT-AI-Agent (commented):
## TL;DR

All four points from the earlier code reviews are addressed (arrows get their own strip and are clickable, wheel scrolling works in the horizontal toolbar, tooltips survive, overflow is cleared before measuring, `toolbar.js` is in both CI matrices) — I re-verified each of them in Chromium and the three new TestCafe tests pass locally. One new, reproducible problem remains: **in the vertical layouts the browser's scroll anchoring silently scrolls the overflowing toolbar to its end while the window is resized, so the leading tabs (Game Shelf / Active Game / Players) disappear off the top** — which is the exact failure mode this PR exists to remove. A one-line `overflow-anchor: none` fixes it.

## Findings

### 1. Resizing the window scrolls the vertical toolbar to its end and hides the first tabs

`client/css/layout.css:487-495` (`body.toolbarOverflow #toolbar`)

`#toolbar` becomes a scroll container in overflow mode, and Chrome's scroll anchoring then compensates every content-size change during a resize by moving `scrollTop`. Because the compaction levels *and* the growing/shrinking `.spacer` change the content size on every resize step, the toolbar walks to the bottom of its scroll range without the user ever scrolling.

Reproduced in Chromium on this branch (fresh room, no user interaction, just `setViewportSize`):

```
start 420x260 (overflow, scrollTop 0)
 -> 420x259  scrollTop  36
 -> 420x250  scrollTop  70
 -> 420x240  scrollTop 104
 -> 420x230  scrollTop 106
 -> 420x220  scrollTop 116   (= scrollHeight 336 - clientHeight 220, i.e. fully scrolled)
```

At that point `#statesButton`, `#activeGameButton` and `#playersButton` are above the visible area (their `getBoundingClientRect().top` is −78 / −44 / −10) — from the player's point of view the Game Shelf button just vanished when they dragged the window smaller, with only a 24px chevron as a hint. Same at 500x300 (`scrollTop` 80) and in the wide layout at 1600x220 (`scrollTop` 100). A *fresh load* at any of those sizes is correctly at `scrollTop` 0, so this only shows up on resize — but window resize / device rotation / a mobile keyboard opening are exactly the common paths into the small sizes.

Confirmed fix, verified with `addStyleTag` on the same run:

```css
body.toolbarOverflow #toolbar {
  overflow-anchor: none;
}
```

With that, the same resize sequence stays at `scrollTop` 2 throughout. The horizontal toolbar is unaffected (scroll anchoring is vertical-only), so the declaration only needs to be in the shared `body.toolbarOverflow #toolbar` block.

The new TestCafe cases don't catch this because `toolbarLayout()`'s `outside` check is only asserted *after* `scrollToolbarToEnd()` in the overflow branch — a toolbar that already sits at its end passes. Asserting `toolbar.scrollTop == 0` right after a `resizeWindow` into an overflowing size would lock this in.

### 2. `scrollToolbarByWheel` mishandles `deltaMode == 2` (page scrolling)

`client/js/main.js:594-599`

```js
scrollToolbarBy(e.deltaY * (e.deltaMode ? 16 : 1)); // some mice report lines instead of pixels
```

`deltaMode` is `0` = pixel, `1` = line, `2` = **page**. The truthiness test treats page-mode deltas as lines, so a page-mode wheel scrolls 16px instead of roughly a viewport. It's a rare input mode, but the fix is as short as the current code:

```js
const step = [ 1, 16, horizontal ? toolbar.clientWidth : toolbar.clientHeight ][e.deltaMode] || 1;
```

### 3. `toolbarArrowSize` is duplicated in JS and CSS

`client/js/main.js:572` (`const toolbarArrowSize = 24; // --toolbarArrowSize in layout.css`) versus `client/css/layout.css:124`. The comment flags the coupling, but `scrollToolbarByArrow()` decides the click hit region from the JS copy, so if the CSS value is ever tuned the arrows keep working visually while the hit region silently drifts off them. `parseFloat(getComputedStyle(toolbar).getPropertyValue('--toolbarArrowSize'))` removes the duplicate for the price of one line, and the value is already being read out of a `getComputedStyle` call site right next door.

## Things I checked that look fine

- The three new TestCafe tests pass locally (`chromium:headless`, 33s), and both workflow matrices now include `toolbar.js`.
- No hysteresis left: at 420x220, 500x300, 300x300, 200x600, 1600x220 and 640x480 the compaction class set is identical growing and shrinking, and `toolbarContentFits()` correctly reads `scrollTop`/`scrollLeft` as 0 because dropping `toolbarOverflow` makes the element non-scrollable before the measurement.
- No `ResizeObserver` loop: after a 25-step resize storm the page idles at exactly 61 rAF callbacks/second, no console errors, no crash overlay.
- Horizontal wheel scrolling works with a plain `deltaY` wheel (`scrollLeft` 0 → 99 at 300x300), and clicking an arrow does not activate the button under it (no overlay opened).
- Tooltips in overflow mode are `position: fixed`, inside the viewport, and correctly suppressed (`display: none`) while their own popup is open; `#options` tracks its button while the toolbar is scrolled (4px offset held at `scrollTop` 200).
- `updateToolbarLayout()` in `client/js/overlays/players.js:81` resolves because `server/minify.mjs` concatenates the client files into one module and strips the imports — same implicit-global convention the rest of that file already uses (`$`, `onMessage`, `scale`).
- `#toolbar > * { flex-shrink: 0 }` does not break `.spacer` (`flex-grow: 1` still wins), and the overflow padding rules out-specify the compaction padding rules by source order.

The `beta` label was already on the PR and I left it in place: the diff touches no engine file (only `client/css/layout.css`, `client/js/main.js`, `client/js/overlays/players.js`, `client/room.html`, the two workflow files and a new test), it cannot affect room state or the server, and finding 1 is a cosmetic scroll-position annoyance rather than anything that could corrupt a game.

VTT-AI-Agent:
All three findings were valid and are fixed in ff33b9a6.

**1. Resizing scrolled the vertical toolbar to its end** — reproduced exactly as described (420x260 → 220: `scrollTop` 0 → 36 → 70 → 104 → 106 → 116, with `#statesButton` at `top: -90`). Fixed with `overflow-anchor: none` in the shared `body.toolbarOverflow #toolbar` block:

```css
/* every compaction level changes the size of the contents, so scroll anchoring would keep
   adjusting scrollTop while the window is resized until the toolbar sits at its end */
overflow-anchor: none;
```

After the fix the same sequence stays at `scrollTop` 2 (the scroll-snap resting point behind the arrow strip) and Game Shelf stays visible the whole way (`top` 39 → 24). I also checked the other layouts you named plus a few more — 500x300, 480x290, 460x280 (aspectTooGood), 1600x260, 1600x220 (wide), 300x600, 280x560, 260x520 (horizontal): the scroll position stays at 2 everywhere, no console errors.

Locked in with a fourth TestCafe case, *"Resizing the window does not scroll the overflowing toolbar by itself"*, which walks 420x260 down to 420x220 and asserts `scrollLeft + scrollTop <= 2` after every resize, without ever scrolling itself. It fails on the previous state of the branch (`420x259: resizing the window scrolled the toolbar - expected 36 to be lte 2`) and passes now. I put it in its own test rather than into the existing one, because that one deliberately scrolls to the end at each size, so a scroll position surviving into the next resize would make the assertion meaningless there.

**2. `deltaMode == 2`** — fixed:

```js
// deltaMode says whether the wheel reports pixels, lines or pages
scrollToolbarBy(e.deltaY * ([ 1, 16, $('#toolbar').clientWidth ][e.deltaMode] || 1));
```

The handler only ever runs for the horizontal toolbar, so a page is `clientWidth`. Verified with synthetic wheel events at 300x600: `deltaY: 1` scrolls to 53 in pixel and line mode (both snap to the next button) and to 99 — the end of the scroll range — in page mode.

**3. `toolbarArrowSize` duplicated** — the JS constant is gone; `scrollToolbarByArrow()` now reads the value the arrows are actually drawn at:

```js
const arrow = parseFloat(getComputedStyle(toolbar).getPropertyValue('--toolbarArrowSize')) || 0;
```

`npm test` (44) and `testcafe toolbar.js` (4, incl. the arrow-click test that covers the new hit-region code) pass locally.

96LawDawg (approved):
Tested on Edge and Firefox on Windows at every screen size I could squeeze out of it. Did not test on mobile.  Approved.

ArnoldSmith86 (approved):
@VTT-AI-Agent Merge it.

PR #3058, opened 2026-07-27
Fix the client error reporter crashing when the error event has no Error object

VTT-AI-Agent:
## Goal

A client error report from the beta server contained nothing but a crash **inside the error reporter itself**:

```
can't access property "message", e is undefined
e@https://beta.virtualtabletop.io/…
window.onerror@https://beta.virtualtabletop.io/…
OnErrorEventHandlerNonNull*@https://beta.virtualtabletop.io/…
EventListener.handleEvent*onLoad@https://beta.virtualtabletop.io/…
```

`window.onerror` is called with `(msg, url, line, col, err)`, and `err` is only there when the browser has an `Error` object to hand over. It is missing for

* script errors from another origin (`Script error.`),
* thrown values that are not Errors (`throw 'oops'`),
* `Promise.reject()` without a reason, and
* the browser's own non-exception error events, e.g. *"ResizeObserver loop completed with undelivered notifications"*.

In all of those cases `errorHandler()` in `client/js/tracing.js` died on `String(error.message)` **before** it could build the report — so the original error is lost. Chromium then shows nothing at all (the exception thrown inside `window.onerror` is not re-reported), Firefox re-enters the handler and reports the reporter's own `TypeError` instead of the real problem. That is exactly what the report above is.

## What this PR does

* New `describeError(error, fallback)` in `tracing.js`: uses `message`/`stack` when there is an Error, otherwise falls back to what the browser *did* pass.
* `window.onerror` passes `msg` + `url:line:col` as that fallback, `unhandledrejection` passes `Unhandled promise rejection`. A thrown value that is not an Error is appended to the fallback instead of turning into `undefined\nundefined`.
* The "submitting the error failed" branch uses the same helper (it had the same `e.message` assumption).
* Only the **first** error is reported now. A repeating error (a throwing `mousemove` handler, a ResizeObserver loop, …) previously rebuilt the whole report — including `document.documentElement.outerHTML` — and added another submit listener each time, so one click on *Submit* sent the same report N times and the first, usually root-cause, error text had already been overwritten.

No behaviour changes for the normal case: a real `Error` produces exactly the same report as before.

## Reproduction on main

Verified against a local checkout of `main` (Chromium, but the code path is browser-independent), room open, then in the console:

```js
window.dispatchEvent(new ErrorEvent('error', { message: 'ResizeObserver loop completed with undelivered notifications.' }))
```

* **before:** `TypeError: Cannot read properties of undefined (reading 'message') at errorHandler … at window.onerror`, the dispatched message never reaches the report.
* **after:** the error overlay shows `ResizeObserver loop completed with undelivered notifications.` and can be submitted normally.

Same with a real ResizeObserver loop, which is what produced the beta report:

```js
const t = document.querySelector('#toolbar'); let i = 0;
new ResizeObserver(() => t.style.paddingBottom = (++i % 2 ? 20 : 4) + 'px').observe(t);
t.style.paddingBottom = '30px';
```

* **before:** a burst of `Cannot read properties of null (reading 'message')` in the console and no error overlay at all — the client is silently broken.
* **after:** one report naming the ResizeObserver loop, with the source location.

`Promise.reject()` (no reason) crashed the handler the same way before and is reported as `Unhandled promise rejection` now; `throw 'a string'` used to report `undefined\nundefined`.

## Where the underlying error on beta came from

`main` contains no `ResizeObserver`; the beta build does — #3057 observes `#toolbar` with a callback that changes the toolbar's size, which is the classic way to produce that loop notification, and the report's body classes show that PR's compaction levels active. I commented on #3057 about it. This PR does not depend on that: the reporter garbles or swallows such errors on `main` too, which is why the crash report was unusable in the first place.

## Testing

* `npm test` (jest) passes — added `tests/client/error-report.test.js` covering `describeError`.
* TestCafe `editor.js` passes locally (chromium headless).
* Playwright: checked the five cases above (Error, non-Error throw, `Promise.reject()` with and without reason, error event without error object) before and after the change.

## Update to main (2026-08-06)

Merged current `main` in (the only textual conflict-adjacent change was #3063's
`dragChain: undefined` in the `mouseStatus` context, which is kept). Two things from
main that this touches:

* #3057 has landed, and it now defers the toolbar layout update to the next frame
  precisely so the ResizeObserver loop notification doesn't reach the error handler.
  That fixes the *source* of the beta report; this PR is still the reporter-side
  safety net, and it also covers cross-origin `Script error.`, non-Error throws and
  bare `Promise.reject()`, none of which #3057 addresses.
* #3097 made the overlays children of `#roomArea`, so an overlay is now as tall as the
  **board**, not as tall as the window. The `max-height: 40vh` this PR added to
  `#clientErrorStack` could therefore exceed the entire overlay — measured 560px of
  stack inside a 500px overlay on a wide board in a tall window. Changed to `40cqh`
  against the `roomArea` size container, which is the unit the rest of the overlay CSS
  switched to in #3097.

Re-verified on the merged branch: `npm test` (1153 tests) passes, and Playwright walks
the error event without an Error object, the benign `ResizeObserver loop` / `Script error.`
events (ignored), a bare `Promise.reject()` and a real `TypeError` submitted end to end
(report accepted, page reloaded).

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3058/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

Good, well-scoped fix — `describeError()` correctly removes the `error.message` assumption that made `window.onerror` crash on error events without an `Error` object, and the one-shot latch is the right call for a terminal overlay. My main concern: `errorReported = true` is set *before* the report is built, so if building `details` ever throws (early error, missing globals), the reporter is now permanently silent instead of at least surfacing something.

## Findings

### 1. The one-shot latch can permanently silence the reporter (`client/js/tracing.js:55-58`)

```js
if(errorReported)
  return;
errorReported = true;
const details = { ... widgetsState: [...widgets.keys()]... };
```

The latch is set before anything that can fail. Everything after it dereferences a pile of globals (`undoProtocol`, `delta`, `mouseStatus`, `mouseTarget`, `widgets`, `playerName`, `showOverlay`, …). If an error fires before those exist — or if a future change reintroduces a throw inside the handler, which is exactly the class of bug this PR is fixing — the handler dies, no overlay appears, and *every* subsequent error is now swallowed by the latch. On `main` the second error at least got another attempt, and in Firefox the re-entry surfaced the reporter's own `TypeError`, which is how this bug was noticed at all.

Suggestion: keep the latch (re-entrancy protection is needed), but make the failure path non-silent, e.g.

```js
try {
  /* build details, show overlay */
} catch(e) {
  $('#clientErrorStack').textContent = describeError(error, fallback) + '\n\nerror reporter failed: ' + describeError(e, 'Unknown error');
  showOverlay('clientErrorOverlay');
}
```

so the raw error text still reaches the user even when the context collection breaks.

### 2. Non-`Error` objects stringify to `[object Object]` (`client/js/tracing.js:31-34`)

`describeError` falls back to `String(error)`, so a rejection like `Promise.reject({status: 500, detail: 'x'})` — a very common shape from fetch wrappers — reports `Unhandled promise rejection\n[object Object]`, i.e. no more information than the fallback alone. A cheap improvement for the object case:

```js
try { return fallback + '\n' + (typeof error == 'object' ? JSON.stringify(error) : String(error)); }
catch(e) { return fallback + '\n' + String(error); }
```

(the try/catch matters — `JSON.stringify` throws on cyclic values and BigInt).

### 3. Minor: `message` without `stack`

The guard `error.message !== undefined || error.stack !== undefined` accepts objects that have only one of the two, producing a literal `undefined` line (`"some message\nundefined"`). Pre-existing behaviour for real `Error`s, but the new guard deliberately widens the branch to message-only objects, so it's now reachable more often. Worth skipping the `\n` + stack part when `stack` is undefined.

- `tests/client/error-report.test.js` importing `tracing.js` is safe under jsdom: the module's only top-level side effect is `onLoad(...)`, which registers a `DOMContentLoaded` listener that has already fired by import time, so none of the room globals are touched.
- No other client code repeats the `err.message`/`err.stack` assumption (`connection.js:30`'s `onerror` is a WebSocket event, unrelated).
- The remaining `e.message` uses in `main.js`/`jsonedit.js`/`tracing.js:42` are all `selectFile()` rejections, which always reject with a real `Error` — no change needed.
- No widget/routine properties added, so no `validator/` or `jeCommands` upkeep is required.
- The overlay is terminal (`preventReconnect()` + `connection.close()`), so reporting only the first error loses nothing the user could have acted on, and it correctly fixes the duplicated `click` listener on `#clientErrorOverlay button`.

I added the `beta` label: the diff only touches `client/js/tracing.js` (the error reporter) and a new jest test — no engine files, no state/sync behaviour, and it strictly reduces the chance of a client crash, so it is safe and useful for players to exercise on beta.

VTT-AI-Agent:
Thanks — all three points were valid, fixed in 92ff6968.

**1. The latch could permanently silence the reporter.** Building the report is now wrapped in a `try`/`catch`: `errorHandler` computes `describeError(...)` first (that one is defensive by design), then calls the new `reportError(description)`. If the context collection throws, the catch still puts the raw error text into `#clientErrorStack` and opens the overlay, with the reporter's own failure appended. Verified in Chromium by making `navigator.userAgent` throw before triggering an error — the overlay shows:

```
boom
Error: boom
    at eval …

The error reporter itself failed:
no userAgent for you
Error: no userAgent for you
    at Navigator.get …
    at reportError …
```

(The submit button is part of what failed in that case, so the user gets the text to copy rather than a working upload — but that beats a frozen page and a silent reporter.)

**2. `[object Object]` for non-`Error` rejections.** Added `stringifyValue()`, used for both branches: objects go through `JSON.stringify` when that works, everything else through `String()`, and both are guarded (cyclic values, BigInt, and objects whose `toString` throws). `Promise.reject({status: 500, detail: 'x'})` now reports `Unhandled promise rejection\n{"status":500,"detail":"x"}` — checked in the browser.

**3. `message` without `stack`.** The branch now joins only the parts that are defined, so `{ message: '…' }` no longer produces a trailing `undefined` line. Real `Error`s are unchanged.

New jest cases cover the object serialization, the cyclic/`Symbol` fallbacks and the message-only object; `npm test` is green (50 tests).

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review) of the crash overlay this PR feeds. The fix genuinely improves what the user sees — errors without an `Error` object now reach `#clientErrorStack` instead of silently freezing the page — but the new fallback path exposes a dead end: when `reportError()` itself throws, the overlay is shown **without** the submit handler wired up, so *Submit and Reload* does nothing at all (verified in a browser, no request, no reload). Most impactful improvement: make that fallback state honest — an explanatory line plus a working *Reload* action and Discord/GitHub links.

Reviewed live on this branch (Chromium via Playwright, local server), triggering the exact cases the PR adds handling for.

| what | screenshot |
| --- | --- |
| `ErrorEvent` with no `Error` (the PR's repro) — now shows the real message | [shot-overlay-noerror-details.png](https://agent.virtualtabletop.io/reports/ui-review/1531277852840820889/shot-overlay-noerror-details.png) |
| `window.onerror` with `msg + url:line:col` fallback | [shot-longurl.png](https://agent.virtualtabletop.io/reports/ui-review/1531277852840820889/shot-longurl.png) |
| reporter itself fails (new catch branch) | [shot-reporter-failed-details.png](https://agent.virtualtabletop.io/reports/ui-review/1531277852840820889/shot-reporter-failed-details.png) |
| same overlay at 390px | [shot-mobile.png](https://agent.virtualtabletop.io/reports/ui-review/1531277852840820889/shot-mobile.png) |

---

## Big layout improvements

**1. The new "reporter itself failed" state shows a UI that lies about what it can do.** `client/js/tracing.js` (catch branch around L133-138) sets `#clientErrorStack` and calls `showOverlay('clientErrorOverlay')`, but the click listener on `#clientErrorOverlay button` is only attached at the end of `reportError()` — which just threw. Result: the user sees the same "How can we trigger the error ourselves so we can fix it?" prompt, types into the textarea, presses **Submit and Reload**, and *nothing happens* (confirmed: zero network requests, no reload). For a first-time user that reads as "the page is completely broken", which is exactly the impression this PR is trying to avoid.
Suggestion: give the overlay an explicit degraded state, e.g. add a `<p id="clientErrorSubmitFailed" class="hidden">` to `client/room.html` with text like *"The error could not be prepared for sending. Please copy the technical details below and report them on [Discord](…) or [GitHub](…)."*, hide the textarea + its question in that branch, and relabel/rewire the button to a plain `Reload` (`location.reload()`). Even the minimal version — always attach a reload handler to the button before showing the overlay — removes the dead control.

**2. The overlay's action area has no escape hatch, and this PR makes reaching it more likely.** `#internalErrorOverlay` (room.html:564-569) offers GitHub and Discord as a next step; `#clientErrorOverlay` (room.html:571-581) offers only *Submit and Reload*. Now that genuinely un-submittable cases (`Script error.` from another origin, a reporter failure) reach this overlay, a second, secondary action belongs next to the primary button — e.g. a `<p>` line "Or reload without sending a report / report it on Discord or GitHub" with the same link styling `.overlay a` already provides. It also gives the user a clear way out when the fetch fails.

**3. Consider ordering the overlay by what the user must do.** Right now: heading → apology → question → textarea → *Technical Details* → button. The button ends up below a `<details>` block that changes height when opened, so the primary action jumps down the page ([compare the two screenshots above](https://agent.virtualtabletop.io/reports/ui-review/1531277852840820889/shot-reporter-failed-details.png) — with a long stack the button lands at y≈758, near the bottom edge on a 800px-tall viewport, and off-screen on shorter ones). Moving `<details>` *below* the action row (or wrapping the action row in a sticky footer inside the overlay) keeps "Submit and Reload" in a fixed, predictable place regardless of stack length.

**4. `${msg}\n    at ${url}:${line}:${col}` renders as a bare `at :0:0` when the browser has no location** (tracing.js, `window.onerror`). That is precisely the cross-origin `Script error.` case this PR is about, and also what a synthetic `ErrorEvent` produces — see the [first screenshot](https://agent.virtualtabletop.io/reports/ui-review/1531277852840820889/shot-overlay-noerror-details.png), where the details block reads:
```
ResizeObserver loop completed with undelivered notifications.
    at :0:0
```
The location line adds nothing there and looks like a truncation bug. Suggest appending the `\n    at …` part only when `url` is non-empty.

**5. `#clientErrorStack` has no CSS at all**, so the `<pre>` keeps `white-space: pre` and the new one-line `at <url>:<line>:<col>` fallback overflows. Measured at 390px wide: `pre.scrollWidth = 462` vs `clientWidth = 312`, and the whole `#clientErrorOverlay` grows to `scrollWidth 501` on a 390px viewport — the entire crash screen scrolls sideways and the message is clipped mid-word ([shot-mobile.png](https://agent.virtualtabletop.io/reports/ui-review/1531277852840820889/shot-mobile.png)). A minified `main.js` URL with a cache-busting query overflows on desktop too. Suggest in `client/css/layout.css`, next to the other `.overlay` rules (~L724-750):
```css
#clientErrorStack {
  white-space: pre-wrap;
  overflow-wrap: anywhere;
  max-height: 40vh;
  overflow: auto;
}
```
`max-height` also stops a long stack (item 3) from pushing the button off-screen.

**6. The submit-failure branch destroys what the user typed.** Both `$('#clientErrorOverlay textarea').value = "Submitting the error failed…"` lines (tracing.js L118 and the L120 one this PR edits) overwrite the reproduction description the user just wrote — the only thing that made the report useful — and then dump a stack into a 150px-tall box the user believes is *their* input field. Since L120 is already being touched here, consider routing that message to `#clientErrorStack` (or the new status `<p>` from item 1) and leaving the textarea alone.

**7. Post-failure the button still reads "Submit and Reload"** with no visual state change, so a user who clicked once and saw the textarea change has no idea whether clicking again will help or resend. A disabled state, or relabelling to `Try again`, would match the app's other button conventions.

**8. "Internal Error" is used as the `<h1>` for two different overlays** (`#internalErrorOverlay` server-side, `#clientErrorOverlay` client-side, room.html:565 and :572). Not introduced here, but this PR makes the client one appear in more situations. "Something went wrong in your browser" reads better for a newcomer and distinguishes the two.

**9. "Technical Details" is the only place the error text lives, and it is collapsed by default with a small (13px) monospace body.** That is a reasonable default, but the newly-surfaced messages are often self-explanatory ("ResizeObserver loop completed with undelivered notifications.") and reassure the user that *something* was captured. Consider a short line under the summary such as *"You don't need to understand this — it is sent with your report"*, so a first-time user doesn't feel obliged to decode it (or fear it contains something private).

**10. "How can we trigger the error ourselves so we can fix it?" is an unlabelled, unplaceholdered textarea.** A newcomer sees a blank box after a crash with no hint of the expected answer. A `placeholder="e.g. I dragged a card onto the deck and the screen froze"` (and wrapping the question in a `<label for>`) would raise the quality of the reports the server receives — directly relevant here, since this PR's whole point is that these reports currently arrive useless.

**11. No feedback while submitting.** `Submit and Reload` fires a `fetch` with `document.documentElement.outerHTML` in the body; on a slow connection there is several seconds of nothing before the reload. A `disabled` + `Submitting…` label during the await would close the loop.

Nothing here blocks the fix itself — the core behaviour change is a clear improvement to what a crashing user sees. Items 1 and 5 are the ones I'd fold into this PR, since both are states this change newly makes reachable.

VTT-AI-Agent:
Thanks — the two items you called out as fold-in-worthy (1 and 5) plus most of the nitpicks are fixed in 3e1c7654. Verified each case live in Chromium again.

**1. Dead "Submit and Reload" in the reporter-failure branch — fixed.** That branch now renders an explicit degraded state instead of a UI that lies: the question + textarea are hidden, a status line appears (*"The report could not be sent. Please copy the technical details below and report them on Discord or GitHub."*, with real links), and the button is relabelled **Reload** with a `location.reload()` handler attached before the overlay is shown. Confirmed by clicking it — the page actually reloads. The invariant still holds: the submit listener is attached as the last statement of `reportError()`, so reaching the `catch` means no submit handler exists and the reload handler is the only one.
reporter failed

**5. `#clientErrorStack` overflow — fixed** with exactly the rule you suggested (`white-space: pre-wrap; overflow-wrap: anywhere; max-height: 40vh; overflow: auto`) in `client/css/layout.css` next to the other `.overlay` rules. Re-measured at 390px with the long-URL fallback: `pre.scrollWidth == clientWidth` and `#clientErrorOverlay.scrollWidth == clientWidth` (was 462/312 and 501/390), so the crash screen no longer scrolls sideways. A minified `main.js` URL with a cache-busting query now wraps on desktop too.
narrow viewport

**4. Bare `at :0:0` — fixed.** `window.onerror` only appends `\n    at ${url}:${line}:${col}` when `url` is non-empty, so the cross-origin / synthetic-`ErrorEvent` case now reports just `ResizeObserver loop completed with undelivered notifications.`

**6. Submit failure destroying the user's text — fixed.** Both messages are gone; the reason now goes into `#clientErrorStack` (appended under the original error) and the explanation into the new status line. The textarea keeps what the user typed — verified: typed a repro, aborted the `clientError` request, text is still there.

**7. Button state after a failure — fixed.** The button goes `disabled` + `Submitting…` while the fetch runs (**11**), and on failure comes back enabled as **Try again**; the status line is hidden again on each new attempt.
submit failed

**8. Duplicate "Internal Error" heading — fixed.** The client overlay is now *"Something went wrong in your browser"*; the server-side `#internalErrorOverlay` keeps "Internal Error".

**10. Unlabelled textarea — fixed.** The question is now a real `<label for="clientErrorDescription">` (clicking it focuses the field, checked) and the textarea has `placeholder="e.g. I dragged a card onto a deck and the screen froze"`.

**3. Ordering — partly applied.** `<details>` moved *below* the action row, so opening it no longer pushes **Submit and Reload** down the page; combined with `max-height: 40vh` a long stack can't move the button either. I did not add a sticky footer — that felt like more overlay chrome than this fix warrants.

Not changed:

- **2 (permanent secondary escape hatch).** The Discord/GitHub line now appears in exactly the states where it is actionable — reporter failure and submit failure. Showing it unconditionally would tell users to file a manual report while the automatic one is still going to work, which I think reads worse; happy to add it if you disagree.
- **9 ("you don't need to understand this").** It would be wrong in the new degraded state, where copying the details *is* the ask, so I left the `<details>` body alone.

`npm test` still green (50 tests). No widget/routine properties touched, so no `validator/` or `jeCommands` upkeep.

VTT-AI-Agent (commented):
## TL;DR

The latest follow-up fixes address the earlier review findings, and the PR remains safe/useful for beta testing. One reporter-silencing edge case remains: a thrown or rejected object whose `message`/`stack` property access itself throws can still escape before the new recovery path and permanently trip the one-shot latch.

## Finding

### `describeError()` can still crash outside the recovery `try` (`client/js/tracing.js:32`, `client/js/tracing.js:147-149`)

`stringifyValue()` guards conversion failures, but `describeError()` first reads `error.message` and `error.stack` without a guard. JavaScript permits arbitrary thrown/rejected values, including proxies or objects with throwing accessors:

```js
const reason = new Proxy({}, {
  get() { throw new Error('getter exploded'); }
});
Promise.reject(reason);
```

Here the read at line 32 throws. Since `errorHandler()` sets `errorReported = true` and computes `description` *before* entering its `try`, no overlay is shown; the error raised by the handler re-enters with the latch already set and is discarded, as are all subsequent errors. That reproduces the same “reporter crashes and original error is lost” failure class this PR fixes.

Please make the error-like property reads defensive (for example, capture `message` and `stack` inside a `try` and fall back to `fallback + stringifyValue(error)` if either access fails), or move description construction into a recovery path that never consults the hostile value again. A test with the proxy above would cover the boundary that the current plain-object/cyclic/Symbol cases do not.

The `beta` label was already present; I left it in place because no engine files or state-sync behavior are touched, CI is green, and this rare reporting-only edge case does not make player testing unsafe.

VTT-AI-Agent:
Valid finding, fixed in 49765f65 — thanks, your repro reproduced exactly as described.

**`describeError()` property reads are now guarded.** The `error.message` / `error.stack` reads sit inside a `try`; if either access throws, the function falls through to `fallback + stringifyValue(error)`, and `stringifyValue()` already survives a hostile value (`JSON.stringify` throws → caught, `String()` throws → caught → `[object that could not be converted to text]`).

**Defense in depth in `errorHandler()`.** `description` is now initialised to `fallback` and built *inside* the existing `try`, so even a future throw from `describeError()` lands in the recovery branch and still shows the overlay instead of silencing the latch.

Verified live in Chromium against your exact snippet on a real room, before and after:

```js
Promise.reject(new Proxy({}, { get() { throw new Error('getter exploded'); } }));
```

| | `#clientErrorOverlay` display | `#clientErrorStack` |
| --- | --- | --- |
| before | `none` (silently swallowed, latch tripped) | *empty* |
| after | `flex` | `Unhandled promise rejection`<br>`[object that could not be converted to text]` |

**Test added** covering both shapes of the boundary: the proxy from your comment, and a plain object with a throwing `message` getter (that one yields `[object Object]`, since `String()` doesn't touch the getter — still a description rather than a crash).

`npm test` green: 51 tests, 5 suites. No widget/routine properties touched, so no `validator/` or `jeCommands` upkeep.

VTT-AI-Agent (commented):
## TL;DR

The `describeError` fix is correct and well tested — `npm test` passes (51 tests, incl. the 7 new ones). But there is a real behavioural regression hiding in it: errors that `main` *swallowed* (because the reporter crashed before `preventReconnect()`) are now *fatal*. I verified in Chromium that a plain `ResizeObserver loop completed with undelivered notifications` — a benign, spec-sanctioned notification, and exactly what #3057 produces on beta — now shows the full "Something went wrong in your browser" overlay and closes the connection, where `main` showed nothing at all. That needs a non-fatal filter before this lands, otherwise it will kick beta players out of their rooms.

## Findings

### 1. Benign browser error events now tear the client down (`client/js/tracing.js:159-166`)

On `main` the reporter died inside `window.onerror` *before* `preventReconnect(); connection.close();`, so a non-`Error` error event was silently ignored and the client kept working. With this PR the handler survives, so every such event now reaches `reportError()` → connection closed, modal overlay, forced reload.

Verified against a local server (Chromium, headless), same room, same trigger — the exact ResizeObserver loop from the PR description:

* `main`: `#clientErrorOverlay` display `none`, `#clientErrorStack` empty — nothing happens.
* this branch: display `flex`, stack `ResizeObserver loop completed with undelivered notifications.` — [screenshot](https://agent.virtualtabletop.io/reports/review/1531388530729943060/ro-overlay.png).

`ResizeObserver loop completed with undelivered notifications` is explicitly non-fatal per spec — the observation is simply retried on the next frame; nothing in the client is broken. `Script error.` from a cross-origin/extension script is the same class of thing. Turning either into "Your client just crashed" + `connection.close()` is a worse outcome than the current silent swallow, and it lands on the *first* one thanks to the new `errorReported` latch.

This is not hypothetical for beta: #3057 is open, labelled `beta`, observes `#toolbar` with a callback that resizes the toolbar, and the report that motivated this PR is exactly that loop. With both on beta, players get a false crash overlay and lose their connection.

Suggestion: filter before reporting, e.g. in `window.onerror` skip when there is no `err` object and `msg` matches a known non-fatal pattern (`/^ResizeObserver loop/`, `/^Script error\.?$/`), or at minimum don't `preventReconnect()/connection.close()` for those. Everything else in the PR (fallback text, one-shot latch, submit-failure handling) is worth keeping as-is.

### 2. `at <url>:0:0` is still produced — the `url ?` guard doesn't catch it (`client/js/tracing.js:161`)

The comment says cross-origin errors are reported without a location so `at :0:0` is avoided. In practice the browser passes the *document* URL with line/col `0`, so the guard is truthy and the observed output is:

```
ResizeObserver loop completed with undelivered notifications.
    at http://localhost:8300/reviewroom3058:0:0
```

which is the misleading-looking line the comment set out to prevent. Checking `url && line` instead of just `url` covers both cases.

### 3. The failure instructions point at a collapsed `<details>` (`client/room.html:576-580`, `client/js/tracing.js:105-110`)

`#clientErrorStatus` asks the user to "copy the technical details below", but the `<details>` block is closed by default and `submitFailed()` only rewrites its contents. Users who hit the failure path see an instruction pointing at something they can't see. `$('#clientErrorOverlay details').open = true;` inside `submitFailed()` (and in the reporter-failed branch of `errorHandler`) fixes it.

### 4. The reporter-failed branch can leave the connection alive while telling the user to reload (`client/js/tracing.js:147-158`)

If `reportError()` throws while building `details` (the most likely place — `mouseStatus`, `widgets`, `jeLoggingRoutineGetData`), `preventReconnect()/connection.close()` on lines 99-100 never ran. The user then gets a terminal "Reload" overlay over a still-connected, still-live session. Minor, but moving `preventReconnect(); connection.close();` to the top of `reportError()` (or into `errorHandler` before the `try`) makes the two paths consistent.

### Verified

* `npm test` — 5 suites / 51 tests pass, including the new `tests/client/error-report.test.js`.
* Behaviour compared on `main` vs. this branch with Playwright/Chromium as described above.
* No `innerHTML` on any error text (`textContent` throughout) — no XSS surface from attacker-controlled error messages. Good.

---

The `beta` label was already on this PR and I left it there: no engine files are touched (`client/js/tracing.js`, `client/room.html`, `client/css/layout.css` only), and the change is safe for the server and for saved games — but be aware that while finding 1 is unfixed, this PR plus #3057 will show beta players a false "your client just crashed" overlay and drop their connection, so it may be worth holding one of the two off beta until the non-fatal filter is in.

VTT-AI-Agent:
All four findings were valid — fixed in 83d54ba6. Thanks especially for finding 1; that regression was real and I could reproduce it exactly as described.

**1. Benign browser error events tore the client down — fixed.** `window.onerror` now bails out on non-fatal events before anything else:

```js
export function isNonFatalError(msg, error) {
  return !error && /^(ResizeObserver loop|Script error\.?$)/.test(`${msg}`);
}
```

The `!error` part is what keeps it narrow: a real crash always comes with an `Error` object, so `throw new Error('Script error.')` is still reported. Verified in Chromium on a real room with the toolbar-style ResizeObserver loop from your repro, watching `WebSocket.prototype.close`:

| | `#clientErrorOverlay` | WebSocket closed |
| --- | --- | --- |
| ResizeObserver loop | `none` | no |
| `Script error.` without an `Error` | `none` | no |
| genuine error | `flex` | yes |

So this branch plus #3057 no longer produces a false crash overlay.

**2. `at <url>:0:0` — fixed.** The guard is `url && line` now, so a browser that passes the document URL with line/column 0 gets no location line. Confirmed: `window.onerror('Something genuinely broke', location.href, 0, 0, undefined)` reports just `Something genuinely broke`, while a real thrown error keeps its `at …:line:col`.

**3. Instructions pointing at a collapsed `<details>` — fixed.** Both places that show `#clientErrorStatus` now go through one helper that also sets `details.open = true`, so the "copy the technical details below" ask always points at something visible. Verified in both states — the reporter-failed branch and a failed submit (request aborted): details expanded, button `Try again`, and the user's typed text still intact.

**4. Reporter-failed branch over a live connection — fixed.** `preventReconnect(); connection.close();` moved to the top of `reportError()`, before the context collection, so a throw while building `details` still leaves a terminal overlay over a terminated session. Verified: with `navigator.userAgent` made to throw, the overlay appears *and* the WebSocket is closed (it stayed open before).

`npm test` green: 53 tests, 5 suites — two new cases cover `isNonFatalError` on both sides of the boundary (benign messages ignored; the same messages *with* an `Error` object still reported). No widget/routine properties touched, so no `validator/` or `jeCommands` upkeep.

reporter failed, details expanded

VTT-AI-Agent:
Updated to current `main` (merge commit, no conflicts) and re-checked the change against it:

* **#3057 has landed** and now defers the toolbar layout update to the next animation frame, precisely so the `ResizeObserver loop` notification never reaches the error handler. That fixes the *source* of the beta report. This PR is still the reporter-side safety net and additionally covers cross-origin `Script error.`, non-Error throws and bare `Promise.reject()`, which #3057 does not touch — so it is not redundant.
* **#3097 made the overlays children of `#roomArea`**, i.e. an overlay is now as tall as the *board*, not as tall as the window. The `max-height: 40vh` this PR put on `#clientErrorStack` could therefore be taller than the whole overlay — measured 560px of stack inside a 500px overlay on a wide board in a tall window. Changed to `40cqh` against the `roomArea` size container (92efd72), which is the unit the rest of the overlay CSS switched to in #3097.
* #3063's `dragChain: undefined` in the `mouseStatus` context is preserved.

Re-verified on the merged head: `npm test` passes (1153 tests) and Playwright walks an error event without an Error object, the benign `ResizeObserver loop`/`Script error.` events (correctly ignored), a bare `Promise.reject()` and a real `TypeError` submitted end to end (report accepted, page reloaded).

All 41 checks are green. I was asked to merge this, but the base branch requires an approving review and I can't approve my own PR — auto-merge (squash) is armed, so it will land as soon as you approve.

ArnoldSmith86 (approved):

PR #3021, opened 2026-07-11
Add new widget type: line

VTT-AI-Agent:
To Do / Notes:

## What

This adds a new widget type **`line`** so game designers can build connected paths/routes — e.g. the taxi/bus/subway networks of a Scotland Yard style map — where each route is a separate, independently editable line.

### The widget (`client/js/widgets/line.js`)

- A line has two end points (`lineStart`, `lineEnd`, relative to the widget position) and is rendered as an SVG path (`lineColor`, `lineWidth`, `lineDash` for dashed routes).
- It is **straight by default**; setting the optional Bezier control points `controlStart`/`controlEnd` makes it a cubic curve. When the line is selected in edit mode, the end points and control points show **draggable handles** (control handles additionally show dashed guide lines).
- **Stops**: the line owns an ordered `stops` list (`[{ "widget": <id>, "position": 0..1 }, …]`) — a stop is any widget in the room, it does not have to be a child of the line. Each stop keeps its own position, so it stays put as the line curves, moves, expands or contracts; `autoSpaceStops` distributes them evenly and `rotateStops` turns landscape stops along the path. New stops added in the editor `inheritFrom` a chosen stop (the start stop by default) so restyling one restyles them all.
- **Taking widgets in (like a holder)**: a line can be a drop target. Its `dropTarget` decides what it accepts, and it **defaults to `[]` — a line takes nothing until it is given one**; the *Valid widgets* row in the editor sidebar (or the JSON editor) sets it, e.g. `{"type": null}` for the plain widgets stops usually are. The two stops of a line added from the add-widget overlay do get `{"type": null}`, so a fresh line's stops accept pawns out of the box. A widget dropped close to the path becomes a child of the line *and* a stop where it landed. Entering and leaving apply the line's `onEnter`/`onLeave` properties and trigger its `enterRoutine`/`leaveRoutine`, exactly like a holder.
- **Connections**: `connectStart`/`connectEnd` (`{ line: <id>, position: 0..1, offset: 0 }`) glue an end point onto another line or widget. Line targets use their path; other widgets use a horizontal path through their midpoint. Global coordinate conversion keeps nested/transformed targets aligned, and cycles are guarded so mutually connected lines terminate.

### Editor support (context-sensitive buttons)

Selecting a line shows a dedicated properties panel:
- **Shape** presets (straight, curves) and a line/ellipse switch, plus color and width pickers,
[...]

PR #3060, opened 2026-07-28
Offer the four line widgets in the Widgets sidebar

VTT-AI-Agent:
## What

The Widgets tab in the right sidebar still offered the two old basic-widget rules under **Decorative** ("Horizontal Rule" and "Vertical Rule"). Since #3021 added the `line` widget type, the add widget overlay offers four line-based decorations instead — but the sidebar never learned about them.

This replaces the two rules with those four widgets, using the exact same JSON the add widget overlay generates:

| Widget | What it is |
| --- | --- |
| **Divider** | a plain line, the drop-in replacement for the old rules (it can be rotated, curved and restyled, so it covers both the horizontal and the vertical case) |
| **Line (with stops)** | the same line with a stop holder at each end |
| **Circle** | a closed line in its `ellipse` shape, without stops |
| **Ring (with stops)** | a closed line with four stops spread all the way round |

Each pair is ordered plain-first, and the two entries that carry stops say so in their name, following the convention the neighbouring group already uses (*Hex (flat)*, *Die (image)*) — without it, *Line*/*Divider* and *Circle*/*Ring* read as synonyms and the only thing telling them apart was the grey `line, holder` type line.

All four also carry a new optional `description`, which the sidebar shows as a tooltip in both list and grid view (in grid view the name is all there is) and which the filter now searches, so the retired vocabulary still finds the replacement: typing `rule`, `horizontal` or `vertical` returns **Divider**.

Three 24×24 icons (`line.svg`, `ring.svg`, `divider.svg`) are added for the previews so the four read as one family — same bar length and stop radius, one stroke weight, and the ring inset far enough that its stops stay clear of the grid view's add button. Circle reuses the existing `circle_hollow.svg`.

These are the first library entries that use `inheritFrom` and a line's `stops`, which exposed a gap in the sidebar's **drag image**: the `dragstart` preview builders only remapped `parent` and `deck`, so every other id reference kept the library id and the widget fell back to its default size — dragging **Line** showed one round stop plus a 111×168 card holder, **Ring** one stop plus three. The dropped result was always correct, only the drag image was wrong. Both builders (grid view and list view) now run the preview states through the same `deepReplace` the placement path uses.

## Why

Anything the add widget overlay offers should also be reachable from the Widgets sidebar — that is where the widgets can be dragged into the room and dropped onto an existing line. Leaving the rules there also pointed people at the old basic-widget-with-a-border trick instead of the real line widget.

## Testing

- `npm test` passes (86 tests), `tests/testcafe/editor.js` passes (5 tests).
- Verified in a browser: the Decorative group now lists Heading, Label, Divider, Line (with stops), Circle and Ring (with stops), and adding each of the four places a correctly-shaped line in the room with its stops attached (`stops`/`parent`/`inheritFrom` are remapped correctly by the sidebar's id remapper).
- Tooltips are present on every entry in both views; filtering for `rule`, `horizontal` and `vertical` returns Divider, `oval` returns Circle, `stops` returns all four.
- Measured the grid view add button against the ring artwork: the button covers viewBox x ≥ 15.6 / y ≤ 8.4, the ring's top stop spans x 9.5–14.5 and its right stop y 9.5–14.5, so both stay visible (also at 500×800).
- Drag image: dispatched `dragstart` on the sidebar entries and inspected the preview container. `lineS1` went from 111×168 / radius 8px to 40×40 / radius 20px, Ring shows four 40×40 circles, and a deck-based entry (*Stacks of Chips*) still previews its cards with the deck's cardType in both grid and list view.

Decorative group in grid view
Decorative group in list view, with all four widgets added to the room

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3060/pr-test (or any other room on that server)

### DIFF
--- a/assets/icons/divider.svg
+++ b/assets/icons/divider.svg
@@ -1,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
+  <rect x="4" y="11" width="16" height="2" fill="currentColor"/>
+  <rect width="24" height="24" fill="none"/>
+</svg>
--- a/assets/icons/line.svg
+++ b/assets/icons/line.svg
@@ -1,8 +1,8 @@
 <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
   <g fill="currentColor">
     <rect x="4" y="11" width="16" height="2"/>
-    <circle cx="4" cy="12" r="3"/>
-    <circle cx="20" cy="12" r="3"/>
+    <circle cx="4" cy="12" r="2.5"/>
+    <circle cx="20" cy="12" r="2.5"/>
   </g>
   <rect width="24" height="24" fill="none"/>
 </svg>
--- a/assets/icons/line.svg
+++ b/assets/icons/line.svg
@@ -1,0 +1,8 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
+  <g fill="currentColor">
+    <rect x="4" y="11" width="16" height="2"/>
+    <circle cx="4" cy="12" r="3"/>
+    <circle cx="20" cy="12" r="3"/>
+  </g>
+  <rect width="24" height="24" fill="none"/>
+</svg>
--- a/assets/icons/ring.svg
+++ b/assets/icons/ring.svg
@@ -1,0 +1,10 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
+  <circle cx="12" cy="12" r="8" fill="none" stroke="currentColor" stroke-width="1.5"/>
+  <g fill="currentColor">
+    <circle cx="12" cy="4" r="2"/>
+    <circle cx="20" cy="12" r="2"/>
+    <circle cx="12" cy="20" r="2"/>
+    <circle cx="4" cy="12" r="2"/>
+  </g>
+  <rect width="24" height="24" fill="none"/>
+</svg>
--- a/assets/icons/ring.svg
+++ b/assets/icons/ring.svg
@@ -1,10 +1,10 @@
 <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
-  <circle cx="12" cy="12" r="8" fill="none" stroke="currentColor" stroke-width="1.5"/>
+  <circle cx="12" cy="12" r="7" fill="none" stroke="currentColor" stroke-width="2"/>
   <g fill="currentColor">
-    <circle cx="12" cy="4" r="2"/>
-    <circle cx="20" cy="12" r="2"/>
-    <circle cx="12" cy="20" r="2"/>
-    <circle cx="4" cy="12" r="2"/>
+    <circle cx="12" cy="5" r="2.5"/>
+    <circle cx="19" cy="12" r="2.5"/>
+    <circle cx="12" cy="19" r="2.5"/>
+    <circle cx="5" cy="12" r="2.5"/>
   </g>
   <rect width="24" height="24" fill="none"/>
 </svg>
--- a/assets/widgets.json
+++ b/assets/widgets.json
@@ -3400,7 +3400,8 @@
       "preview": "symbol:format_h1"
     },
     {
-      "name": "Line",
+      "name": "Line (with stops)",
+      "description": "A straight line with a drop spot at each end - drop a piece onto a stop and it snaps to it. Move, curve or rotate the line and the stops follow.",
       "id": "line",
       "widgets": [
         {
@@ -3461,6 +3462,7 @@
     },
     {
       "name": "Divider",
+      "description": "A plain line without stops - the horizontal or vertical rule to separate areas of the table. Rotate it, curve it or restyle it however you like.",
       "id": "divider",
       "widgets": [
         {
@@ -3481,10 +3483,11 @@
           "lineWidth": 4
         }
       ],
-      "preview": "i/icons/remove.svg"
+      "preview": "i/icons/divider.svg"
     },
     {
       "name": "Circle",
+      "description": "A plain circle outline without stops - resize it into an oval or restyle it like any other line.",
       "id": "circle",
       "widgets": [
         {
@@ -3509,7 +3512,8 @@
       "preview": "i/icons/circle_hollow.svg"
     },
     {
-      "name": "Ring",
+      "name": "Ring (with stops)",
+      "description": "A circle with four drop spots spaced around it - drop a piece onto a stop and it snaps to it. Add or move stops to build a round track.",
       "id": "ring",
       "widgets": [
         {
@@ -4638,8 +4642,8 @@
       "widgets": [
         "heading",
         "label",
-        "line",
         "divider",
+        "line",
         "circle",
         "ring"
       ],
--- a/assets/widgets.json
+++ b/assets/widgets.json
@@ -3400,42 +3400,202 @@
       "preview": "symbol:format_h1"
     },
     {
-      "name": "Horizontal Rule",
-      "id": "horizontalRule",
+      "name": "Line",
+      "id": "line",
       "widgets": [
         {
-          "id": "horizontalRule",
-          "x": 1300,
-          "y": 325,
-          "width": 200,
-          "height": 0,
-          "borderRadius": "3px",
-          "z": 26,
-          "css": {
-            "border": "3px solid #666"
-          }
+          "type": "line",
+          "id": "line",
+          "x": 1310,
+          "y": 420,
+          "width": 220,
+          "height": 40,
+          "lineStart": {
+            "x": 10,
+            "y": 20
+          },
+          "lineEnd": {
+            "x": 210,
+            "y": 20
+          },
+          "stops": [
+            {
+              "widget": "lineS0",
+              "position": 0
+            },
+            {
+              "widget": "lineS1",
+              "position": 1
+            }
+          ]
+        },
+        {
+          "type": "holder",
+          "id": "lineS0",
+          "parent": "line",
+          "fixedParent": true,
+          "movableInEdit": false,
+          "x": -10,
+          "y": 0,
+          "width": 40,
+          "height": 40,
+          "borderRadius": 20,
+          "dropTarget": {
+            "type": null
+          },
+          "dropOffsetX": 2,
+          "dropOffsetY": 2
+        },
+        {
+          "type": "holder",
+          "id": "lineS1",
+          "parent": "line",
+          "fixedParent": true,
+          "movableInEdit": false,
+          "inheritFrom": "lineS0",
+          "x": 190,
+          "y": 0
         }
       ],
-      "preview": "i/icons/width.svg"
+      "preview": "i/icons/line.svg"
     },
     {
-      "name": "Vertical Rule",
-      "id": "verticalRule",
+      "name": "Divider",
+      "id": "divider",
       "widgets": [
         {
-          "id": "verticalRule",
-          "x": 1535,
-          "y": 190,
-          "width": 0,
-          "height": 200,
-          "borderRadius": "3px",
-          "z": 27,
-          "css": {
-            "border": "3px solid #666"
-          }
+          "type": "line",
+          "id": "divider",
+          "x": 1290,
+          "y": 315,
+          "width": 220,
+          "height": 20,
+          "lineStart": {
+            "x": 10,
+            "y": 10
+          },
+          "lineEnd": {
+            "x": 210,
+            "y": 10
+          },
+          "lineWidth": 4
         }
       ],
-      "preview": "i/icons/height.svg"
+      "preview": "i/icons/remove.svg"
+    },
+    {
+      "name": "Circle",
+      "id": "circle",
+      "widgets": [
+        {
+          "type": "line",
+          "id": "circle",
+          "x": 1300,
+          "y": 500,
+          "width": 100,
+          "height": 100,
+          "lineShape": "ellipse",
+          "lineStart": {
+            "x": 10,
+            "y": 10
+          },
+          "lineEnd": {
+            "x": 90,
+            "y": 90
+          },
+          "lineWidth": 4
+        }
+      ],
+      "preview": "i/icons/circle_hollow.svg"
+    },
+    {
+      "name": "Ring",
+      "id": "ring",
+      "widgets": [
+        {
+          "type": "line",
+          "id": "ring",
+          "x": 1420,
+          "y": 495,
+          "width": 130,
+          "height": 130,
+          "lineShape": "ellipse",
+          "lineStart": {
+            "x": 15,
+            "y": 15
+          },
+          "lineEnd": {
+            "x": 115,
+            "y": 115
+          },
+          "stops": [
+            {
+              "widget": "ringS0",
+              "position": 0
+            },
+            {
+              "widget": "ringS1",
+              "position": 0.25
+            },
+            {
+              "widget": "ringS2",
+              "position": 0.5
+            },
+            {
+              "widget": "ringS3",
+              "position": 0.75
+            }
+          ]
+        },
+        {
+          "type": "holder",
+          "id": "ringS0",
+          "parent": "ring",
+          "fixedParent": true,
+          "movableInEdit": false,
+          "x": 45,
+          "y": -5,
+          "width": 40,
+          "height": 40,
+          "borderRadius": 20,
+          "dropTarget": {
+            "type": null
+          },
+          "dropOffsetX": 2,
+          "dropOffsetY": 2
+        },
+        {
+          "type": "holder",
+          "id": "ringS1",
+          "parent": "ring",
+          "fixedParent": true,
+          "movableInEdit": false,
+          "inheritFrom": "ringS0",
+          "x": 95,
+          "y": 45
+        },
+        {
+          "type": "holder",
+          "id": "ringS2",
+          "parent": "ring",
+          "fixedParent": true,
+          "movableInEdit": false,
+          "inheritFrom": "ringS0",
+          "x": 45,
+          "y": 95
+        },
+        {
+          "type": "holder",
+          "id": "ringS3",
+          "parent": "ring",
+          "fixedParent": true,
+          "movableInEdit": false,
+          "inheritFrom": "ringS0",
+          "x": -5,
+          "y": 45
+        }
+      ],
+      "preview": "i/icons/ring.svg"
     },
     {
       "name": "Basic Widget",
@@ -4478,8 +4638,10 @@
       "widgets": [
         "heading",
         "label",
-        "horizontalRule",
-        "verticalRule"
+        "line",
+        "divider",
+        "circle",
+        "ring"
       ],
       "collapsed": true
     },
--- a/client/js/editmode.js
+++ b/client/js/editmode.js
@@ -287,7 +287,9 @@ function generateLineStop(id, lineID, index, x, y) {
 }
 
 // every line offered in the add widget overlay is drawn in the VTT blue the
-// line widget defaults to, so none of them has to spell out a lineColor
+// line widget defaults to, so none of them has to spell out a lineColor.
+// The same four lines are also in assets/widgets.json so that they show up in
+// the Widgets sidebar - keep both copies in sync when changing one of them.
 function generateLineWidgets(id, x, y) {
   const line = {
     type: 'line',
--- a/client/js/editor/sidebar/widgets.js
+++ b/client/js/editor/sidebar/widgets.js
@@ -109,6 +109,14 @@ async function getSavedWidgets(source) {
   return { widgets: [], groups: [] };
 }
 
+// Entries can explain themselves in plain language through an optional
+// description. In grid view the name is all that is shown, so the tooltip is
+// the only place that can tell apart e.g. a line with stops from a plain one.
+function tooltipAttribute(state) {
+  const tooltip = [ state.name || state.id, state.description ].filter(t=>t).join('\n');
+  return ` title="${html(tooltip)}"`;
+}
+
 // Render preview markup from a URL or a Material Symbol reference like "symbol:heading"
 function renderSavedWidgetPreviewHTML(preview) {
   if (!preview) return '';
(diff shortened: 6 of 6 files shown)


VTT-AI-Agent (commented):
## TL;DR

Data-only change (`assets/widgets.json` + two icons) that mirrors the add-widget overlay's line generators verbatim — I diffed all four entries against `generateLineWidgets`/`generateRingWidgets`/the divider+circle literals in `client/js/editmode.js` and they match exactly, and nothing in the repo still references `horizontalRule`/`verticalRule`. One real (cosmetic) issue: these are the first library entries to use `inheritFrom` and `stops`, and the sidebar's **drag-image** builder remaps only `parent`/`deck`, so the trailing stops render as default 111×168 card holders while dragging.

## Findings

### 1. Drag preview renders the non-first stops wrong (`client/js/editor/sidebar/widgets.js:865-885`)

`assets/widgets.json` — `line`/`ring` entries — are the first library entries in the file to use `inheritFrom` and `stops` (I checked: no other entry contains either key). The *placement* path is fine: `placeSavedWidgetFromBuffer` runs `deepReplace`, which rewrites string values under `inheritFrom` and inside `stops[].widget` (`widgets.js:23`, `inheritFrom` is only in `protectedKeys` for key *renaming*, not value replacement). But the `dragstart` drag-image builder does its own, much narrower remapping:

```js
const parentId = tempState.parent ? idMap.get(tempState.parent) : null;
...
if (tempState.deck) { ... tempState.deck = idMap.get(tempState.deck); }
previewWidget.applyInitialDelta(tempState);
```

`inheritFrom` and `stops[].widget` keep the original ids (`lineS0`, `ringS0`), while the preview instances live under `drag-preview-<n>` ids. Concrete failure: dragging **Line** from the sidebar in a room that has no widget literally named `lineS0` makes `getDefaultValue` (`client/js/statemanaged.js:50-53`) fall through to the Holder defaults, so `lineS1` is drawn as a 111×168 white square card holder (`client/js/widgets/holder.js:8-9`) instead of the 40×40 circular stop — i.e. the drag image of **Line** shows one round stop and one big rectangle, and **Ring** shows one round stop plus three big rectangles. Worse-but-harmless variant: if the room *does* already contain a widget with id `lineS0` (very likely, since that's exactly the id the first placed Line gets), the preview silently inherits that unrelated widget's styling. The dropped result is always correct; only the drag image is wrong. `applyDelta` doesn't fire `onPropertyChange`, so the stale `stops` ids can't move the real room widgets — I checked that specifically.

Fix is small and belongs in that builder: after computing `idMap`, run the same `deepReplace(tempState, mapObjectFromIdMap)` the placement path uses (or at minimum remap `inheritFrom` and `stops[].widget`) before `applyInitialDelta`. Worth doing here since this PR is what first exposes the gap.

### 2. Minor: the four entries also show up in the line stop picker

`renderLibraryStopPicker` (`client/js/editor/sidebar/properties.js:5350-5400`) lists every saved widget as a candidate stop with no type filter, so **Line**/**Ring** can now be added as a stop *of another line*, nesting a line (plus its own holders) inside a line's `stops`. It doesn't crash — `stopCoordInParentFrame` handles it — but it's not something anyone means to do. Not introduced by this PR (the picker was already unfiltered), just newly reachable; a `entry.widgets?.[0]?.type != 'line'` filter there would be a reasonable follow-up.

### 3. Nit: the JSON now duplicates the overlay generators

The four entries are a verbatim copy of `generateLineWidgets`, `generateRingWidgets` and the divider/circle literals in `client/js/editmode.js:288-323, 1261-1290`, down to the overlay's layout coordinates (`1310/420`, `1290/315`, …). That's the only practical option for a server-side data file, but the two now drift independently — a comment in `editmode.js` pointing at the widgets.json entries (or vice versa) would save the next person from changing one and not the other.

### Checked and fine

- Stop geometry matches the paths: `lineS0/S1` centres land on `lineStart`/`lineEnd` (10,20)/(210,20); the ring's four holders centre on top/right/bottom/left of the 100px-diameter ellipse, matching positions 0/0.25/0.5/0.75.
- Id remapping on repeated placement is unambiguous: `line` → `line1`, `lineS0/S1` → `lineS2/S3` (the `/_?\d+$/` base-name strip yields `lineS`, and the `startsWith` scan skips `S0`-style suffixes).
- No dangling references: nothing in the repo mentions `horizontalRule`/`verticalRule`; all six `Decorative` group refs resolve; no duplicate ids among the 48 entries.
- The two new icons follow the house convention exactly (24×24 viewBox, `currentColor`, trailing transparent `<rect>`), matching `width.svg`/`remove.svg`.
- No new widget or routine properties, so no `validator/` or `jeCommands` upkeep is needed. All CI checks are green.

The `beta` label was added, because the diff touches only `assets/widgets.json` and two new icons — no engine file at all — so it is not beta-excluded, and a library-data change that can only add correctly-shaped widgets to a room is safe and useful for players to try.

VTT-AI-Agent:
Thanks — addressed in 6f4ba719.

**1. Drag preview renders the non-first stops wrong — fixed.**

You were right, and it reproduced exactly as described. Both `dragstart` drag-image builders (grid view at `widgets.js:895`, list view at `widgets.js:1133`) now route their preview states through the same `deepReplace` the placement path uses, via a shared `remapPreviewState()` helper, so `inheritFrom`, `stops[].widget` and anything else pointing at a buffer id lands on the `drag-preview-*` copies.

Verified in a browser (Playwright, dispatching `dragstart` and reading the preview container before it is torn down on the next tick) — dragging **Line**:

| | before | after |
|---|---|---|
| `lineS0` | 40×40, radius 20px | 40×40, radius 20px |
| `lineS1` (`inheritFrom`) | **111×168, radius 8px** | 40×40, radius 20px |

**Ring** now shows all four stops as 40×40 circles. Checked the deck path did not regress: *Stacks of Chips* still previews its seven cards with the deck's `pokerChip` cardType in both grid and list view. `npm test` (86 passed) and `tests/testcafe/editor.js` (5 passed) are green locally.

**2. The four entries also show up in the line stop picker — not changing this.**

I looked at filtering `type == 'line'` out of `renderLibraryStopPicker` and decided against it: **Circle** and **Divider** are themselves line widgets, and putting a circle marker at position 0.5 of a line is a perfectly reasonable decorative use. A type filter would block that to prevent a case that already doesn't crash. Happy to reconsider if you'd rather have a narrower filter (e.g. only excluding entries that have their own `stops`).

**3. The JSON duplicates the `editmode.js` generators — noted in code.**

`assets/widgets.json` is JSON so it can't carry a comment, so the pointer goes on the `editmode.js` side: the comment above `generateLineWidgets` now says the same four lines live in `assets/widgets.json` and both copies need to stay in sync.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review) of #3060. The change is a clear win — the sidebar now offers what the add-widget overlay offers, the icons are legible in both light and dark mode, and the drag-image fix genuinely works (I measured the preview stops: 40×40 with `border-radius: 20px`, previously the 111×168 card-holder fallback). The one thing worth fixing before merge is **naming/ordering**: *Line* vs *Divider* and *Circle* vs *Ring* are near-synonyms to a newcomer, and the only visible differentiator is the grey type line reading `line, holder` — internal jargon.

Reviewed live at 1400×900 and 500×800, light + dark, list + grid view (server run locally on the PR branch).

Decorative group, list view
Decorative group, grid view
All four added to the room

---

## Big layout improvements

**1. The four names don't carry the distinction they need to carry.** The group now reads *Heading · Label · Line · Divider · Circle · Ring*. "Line" and "Divider" are both lines; "Circle" and "Ring" are both circles. What actually separates each pair is *has stops you can drop pieces onto*, and the only place that shows up in the UI is `.widget-type` — which prints `line, holder` vs `line` (`client/js/editor/sidebar/widgets.js:1374`). A first-time user has no way to know "holder" means "a spot a pawn can sit on".

The group next door already solves exactly this with a parenthetical convention: *Checker (3D)*, *Hex (flat)*, *Hex (point image)*, *Die (image)*. Applying it here costs nothing and fits the app's own style:

| now | suggested |
| --- | --- |
| Line | **Line (with stops)** |
| Divider | **Divider** |
| Circle | **Circle** |
| Ring | **Circle (with stops)** — or keep *Ring* and let the parenthetical carry it |

**2. The two pairs are ordered in opposite directions.** `assets/widgets.json:4638` lists `line, divider, circle, ring` — i.e. *with-stops → plain* for the straight pair, then *plain → with-stops* for the round pair. Reordering to `divider, line, circle, ring` makes the group read as two consistent "plain, then the one with stops" pairs, which is also how the icons will then step up in complexity left-to-right in grid view.

**3. The previews don't look like what you get.** Dropped into the room the widgets are VTT blue with *hollow white* stop circles; in the sidebar they're solid black artwork (compare the two screenshots above). Since `line.svg` and `ring.svg` are brand-new files drawn specifically for this, drawing the stops as rings rather than filled dots — and, if you're willing to hardcode a fill, in the line's default blue — would make the preview read as a picture of the result rather than an abstract glyph. Related: `fill="currentColor"` in both new icons is inert, because `renderSavedWidgetPreviewHTML` (`widgets.js:122`) embeds them as ``, where an external SVG can't inherit page color — they resolve to black in every theme. That's fine today (the preview tile keeps its light `#bbb` background in dark mode, so contrast is OK — verified, screenshot below), but if `currentColor` was meant to make them theme-aware, it won't.

Dark mode — preview tiles keep a light background, icons stay legible

---

- **The grid-view "+" badge sits on top of the new Ring artwork.** `.add-to-room-grid` is pinned to the top-right of `.widget-preview`, and `assets/icons/ring.svg` draws all the way to the edge of its 24×24 box — the stop dot at `cx=12 cy=4` and the one at `cx=20 cy=12` are both partly under the badge, so *the feature that distinguishes Ring from Circle is the part that gets covered*. Heading, Label, Line and Divider all leave that corner empty. Fix: inset the ring artwork (circle `r≈7`, dots pulled in to match) so the top-right quadrant clears the badge. Most visible on a narrow sidebar:
  Narrow sidebar, grid view
- **The four icons aren't a family.** Within one group: `ring.svg` uses `stroke-width="1.5"` while the `circle_hollow.svg` it sits next to has a 2-unit-thick ring — so *Ring* reads lighter than *Circle*, even though the Ring widget is the bigger one (130px vs 100px). And a stop is drawn at `r=3` in `line.svg` but `r=2` in `ring.svg` — the same UI element at two sizes in two adjacent icons. Pick one weight (bar/stroke 2, stop radius ~2.5) and use it in both new files.
- **Divider's bar is visibly shorter than Line's.** `remove.svg` spans x 5→19 (14 of 24); `line.svg`'s bar spans x 4→20 (16 of 24). Side by side in the list this reads as an accidental size difference rather than a deliberate one.
- **Divider reuses `i/icons/remove.svg`** — the standard minus/"remove" glyph. In a list of things you *add*, a "−" as the preview is an odd signal. A dedicated 2-line-thickness bar (or `line.svg` minus its end dots) would avoid the double meaning and fix the length mismatch above in one go.
- **`assets/icons/width.svg` and `assets/icons/height.svg` are now orphaned** — they were the previews of the two removed rules and nothing under `client/`, `assets/` or `library/` references them any more. Worth deleting in the same PR so the icon set doesn't accumulate dead entries.

---

- **No tooltip anywhere.** Neither `renderListWidget` nor `renderGridWidget` emits a `title` (I checked the rendered DOM: `title` is `null` on every entry). With four similarly-named line variants this is the cheapest possible fix: a one-liner per entry — *Line: "A straight line with a drop spot at each end"*, *Ring: "A circle with four drop spots around it"*, *Divider: "A plain line — rotate or curve it to taste"*. In grid view the type subtitle isn't even shown, so the name is genuinely all the user has.
- **`line, holder` is jargon in the one place it has to do real work.** See Big #1. Even without renaming, a friendlier subtitle for these entries would help more here than anywhere else in the sidebar.
- **The word "Rule" has disappeared from the vocabulary.** The filter matches name + widget type only (`widgets.js:488`), so a returning user typing `rule` — or `horizontal` — now gets an empty list with no explanation. Either fold the old word into the name (*Divider (Rule)*) or accept it; worth a conscious decision since the PR is deliberately retiring two long-standing entries.
- **Two entries collapsed into one, and the "vertical" case became invisible.** The PR body notes Divider covers the vertical case because it can be rotated — but nothing in the sidebar says so. A user who came looking for "Vertical Rule" sees one horizontal-looking preview and no hint that rotation is the answer. A tooltip (previous point) is enough to close this.
- **Clicking "+" places the widget out of the way, not in view.** All four land at their library coordinates (x ≈ 1290–1550, from `assets/widgets.json`), so they appear at the far right of the room rather than near where you're working. This is pre-existing behaviour shared with every sidebar entry, but a 4px-wide blue Divider at the room's right edge is much easier to miss than a card holder — a newcomer may well click "+" and believe nothing happened. Not a blocker for this PR; flagging because these four are the thinnest things the sidebar can add.

All four added via "+", with the sidebar open

---

Everything else checked out: dark mode contrast, narrow-sidebar wrapping (the panel scrolls, nothing gets clipped), the grid/list toggle, and the drag preview for all four plus a deck-based entry.

VTT-AI-Agent:
Thanks — this was a good catch on the naming; addressed in 17c68492.

Decorative group in grid view, after
Decorative group in list view, after

**Big #1 — the names don't carry the distinction. Fixed, with your table.** *Divider · Line (with stops) · Circle · Ring (with stops)*. I kept *Ring* rather than *Circle (with stops)* so it stays findable by its own name, and let the parenthetical do the explaining. Two-line names in grid view are normal for the group next door (*Hex (flat image)*, *Stacks of Chips (3D)*), so nothing new breaks there.

**Big #2 — the pairs were ordered in opposite directions. Fixed.** The group is now `heading, label, divider, line, circle, ring`, i.e. plain-then-with-stops twice, and the icons step up in complexity left to right.

**Big #3 — previews don't look like what you get. Partly.** I redrew the artwork so the *shape* matches (stops sitting on the path, the ring inset, matching weights), but I left the glyphs solid and monochrome: every icon in `assets/icons` is a solid `currentColor` silhouette, and drawing a hollow stop only reads as hollow if I hardcode a light fill for the hole — the bar/ring passing through it is 2 units wide out of a 24-unit box, so a `fill="none"` ring just fills in. Hardcoding white would look right on the `#bbb` preview tile and wrong anywhere else the icon gets used, so I'd rather not.
On `fill="currentColor"` being inert under ``: correct, and intentional to keep — it's the house convention for the whole icon set (so the files also work when inlined), and black is the right resolved value on the light preview tile in both themes.

**Nit — the "+" badge covers the ring's stops. Fixed.** The ring is now inset (r 8 → 7, stops pulled onto it at ±7). Measured in the browser: at 75px tiles the badge covers viewBox `x ≥ 15.6, y ≤ 8.4`; the top stop spans x 9.5–14.5 and the right stop y 9.5–14.5, so both distinguishing dots are clear of it, in a full-width and a 500px-wide sidebar. The bare arc still passes under the badge — so does `circle_hollow.svg`'s, and `chip.svg`'s, and `marble.svg`'s — that part I left alone.

**Nit — the four icons aren't a family. Fixed.** `ring.svg` stroke 1.5 → 2 (same 2 units as `circle_hollow.svg`'s ring), and the stop radius is now 2.5 in both new files instead of 3 in one and 2 in the other.

**Nit — Divider's bar is shorter, and reuses the "remove" glyph. Fixed.** New `assets/icons/divider.svg`: the exact bar from `line.svg` minus its end dots (x 4→20, 2 thick), so the pair reads as "the same line, with and without stops" and the minus/"remove" double meaning is gone.

**Nit — `width.svg`/`height.svg` are orphaned. Not deleting them.** Nothing in the repo references them, but `i/icons/*.svg` is a URL that any game can use as a widget `image` (the public library does it ~40 times), so deleting an icon can break a room out in the wild that pointed at it, and there's no migration for that. Two 300-byte files is a cheaper price than that risk.

**Readability — no tooltips / `line, holder` jargon / "Rule" and the vertical case have disappeared. Fixed together.** Library entries can now carry an optional `description`, which `renderListWidget`/`renderGridWidget` emit as a `title` (name + description) and which the filter searches too. All four new entries have one, e.g. Divider: *"A plain line without stops - the horizontal or vertical rule to separate areas of the table. Rotate it, curve it or restyle it however you like."* Verified in the browser:

| filter | result |
| --- | --- |
| `rule` | Divider |
| `horizontal` | Divider |
| `vertical` | Divider |
| `oval` | Circle |
| `stops` | Divider, Line (with stops), Circle, Ring (with stops) |

So the retired vocabulary lands on its replacement, and the vertical case is spelled out where a user looking for "Vertical Rule" will hit it. Entries without a description just get their name as the tooltip, which is still an improvement in list view where the name input truncates.

**Readability — "+" places the widget at its library coordinates, far right. Agreed, and left alone.** It's how every entry in the sidebar behaves (the coordinates come straight from the saved widget), so fixing it means changing placement for all 48 entries — that's its own PR, not this one.

`npm test` (86) and `tests/testcafe/editor.js` (5) are green locally.

VTT-AI-Agent (commented):
## TL;DR

The current head looks good: the earlier drag-preview and discoverability issues are fixed, the four saved-widget definitions remain consistent with the existing add-widget generators, and the full CI suite is green. I found no correctness, security, compatibility, or test-coverage issue worth raising.

## Detailed review

No actionable findings.

I specifically re-checked the two areas changed since the earlier code review:

- `client/js/editor/sidebar/widgets.js`: `remapPreviewState` now applies the same `deepReplace` mapping used by real placement, so `parent`, `deck`, `inheritFrom`, and nested references such as `stops[].widget` consistently target the temporary preview instances. Missing external decks retain the previous skip behavior, and tooltip text is escaped before being placed in the `title` attribute.
- `assets/widgets.json`: Divider, Line, Circle, and Ring still match the corresponding definitions in `client/js/editmode.js`, including stop geometry and inheritance. The descriptions are metadata only; they neither add widget properties nor change existing saved-game semantics, so validator, JSON-editor command, migration, and legacy-mode updates are not required.

The changed-file list contains no engine paths (`client/js/widgets/`, `client/js/compute.js`, or `client/js/serverstate.js`); the sidebar/editor changes affect only edit-mode library presentation and temporary drag images, not how existing games run.

The `beta` label was already present; I left it in place because this non-engine UI/library change is safe and useful for player testing, with all current checks passing.

VTT-AI-Agent:
Thanks for the re-check — nothing to change on my side for automated review 3/4.

Confirming the two areas you re-verified are the intended final state: `remapPreviewState` deliberately routes through the same `deepReplace` mapping as real placement so the drag preview can't reference the live room's widgets, and the `description` fields in `assets/widgets.json` are sidebar-only metadata (tooltip + filter text) — they never reach a room's state, which is why no validator/jsonedit/FileUpdater work is needed.

Leaving the `beta` label as-is; I didn't add it.

VTT-AI-Agent (commented):
## TL;DR

Looks good — I checked this out, placed all four entries and inspected the drag previews in a real browser, and everything behaves as described (`stops`/`parent`/`inheritFrom` remap correctly, previews now show 40×40 round stops instead of card-holder fallbacks, deck-based entries still preview their cards). No blocking issues. The one thing I'd raise is the deliberate duplication of the four line definitions between `client/js/editmode.js` and `assets/widgets.json`, which is now guarded only by a comment.

## Findings

### 1. The four line definitions are duplicated, guarded only by a comment — `client/js/editmode.js:289-292`

The JSON in `assets/widgets.json` is byte-for-byte the output of `generateLineWidgets`/`generateRingWidgets` plus the two inline `Line` literals at `editmode.js:1261` and `:1273` (I diffed them — they match exactly, including the odd overlay coordinates like `x: 1310, y: 420` that only make sense inside the add-widget overlay). The new comment is the only thing keeping them in sync, and nothing fails if they drift: a future tweak to the stop radius or `lineWidth` silently applies to the overlay only.

`assets/widgets.json` is read by the server at boot (`server.mjs:33`) and by the client via `/api/widgets`, so it can't import the generators directly — but the reverse is possible: the overlay could read the four entries out of the loaded library instead of re-declaring them, or a jest test could assert `generateLineWidgets('line', 1310, 420)` deep-equals the `line` entry's `widgets` array. The latter is cheap and would catch drift in CI. Not blocking, but worth doing before a third copy appears.

### 2. Redundant tooltip on entries without a description — `client/js/editor/sidebar/widgets.js:114-117`

`tooltipAttribute` always emits a `title`, so entries with no description get `title="Heading"` / `title="Label"` — in list view that's a tooltip that just repeats the label sitting right next to the cursor. Verified live: the six Decorative entries render `title="Heading"`, `title="Label"`, then the four with real two-line tooltips. In grid view the name-only tooltip is genuinely useful (names are truncated there), so the fix isn't "only emit when there's a description" — probably just skip the bare-name tooltip in `renderListWidget` and keep it in `renderGridWidget`. Cosmetic.

### 3. Pre-existing (not introduced here): `deepReplace` mangles routine parameters whose key collides with a widget id

Not caused by this PR, but this PR is what makes `deepReplace` the single remapping path for both preview and placement, so it's worth recording. `deepReplace`'s key-renaming pass (`widgets.js:44-51`) renames any object key that matches an id in `idMap` and isn't in `protectedKeys`. Two shipped library entries hit this. Reproduced live — adding **Timer** from the sidebar twice into a fresh room:

```
pauseTimer  -> [{"func":"TIMER","timer":"${PROPERTY parent}"}]     // 1st copy, fine
pauseTimer1 -> [{"func":"TIMER","timer1":"${PROPERTY parent}"}]    // 2nd copy, broken
resetTimer1 -> [{"func":"TIMER","mode":"reset","timer1":"..."}]    // 2nd copy, broken
```

The second Timer's play/pause and reset buttons are dead, because the `timer` parameter key was renamed to `timer1`. The `canvas` entry has the same problem (`{"func":"CANVAS","canvas":"${PROPERTY parent}"}` → `{"func":"CANVAS","TMP":"..."}`). The four new line entries are unaffected — I simulated `deepReplace` over every entry in `assets/widgets.json` and only `canvas` and `timer` show a corrupted key; `line`/`ring` remap exactly the fields they should (`stops[].widget`, `inheritFrom`, `parent`).

The key-renaming pass looks like it exists for user-saved widgets that key objects by widget id; restricting it to keys nested under such maps, or skipping it inside anything named `*Routine`, would fix it. Worth a separate issue rather than expanding this PR.

## What I verified

- `npm test`: 86 passing. CI on the PR is fully green (including CodeQL and the TestCafe matrix).
- Placed all four entries from the sidebar into a room and dumped the resulting state: `ring` gets four `holder` stops with `parent`/`inheritFrom` pointing at the placed copies, `line` two, `divider`/`circle` are bare `line` widgets — all rendering correctly.
- Dispatched `dragstart` on the list entries and inspected the preview container: `ring` → one 130×130 line plus four `40px × 40px, border-radius: 20px` holders; `line` → 220×40 plus two; `stacksOfChips` → still previews its `pokerChip` cards and the deck at `scale(0.5)`. The drag-image fix does what it says and doesn't regress deck-based entries.
- Tooltips render in both views with the name and description on separate lines, escaped through `html()`; group order in the sidebar follows `groups[].widgets` (Divider, Line, Circle, Ring — plain-first, as described), and no group reference in `assets/widgets.json` is left dangling by the `horizontalRule`/`verticalRule` removal.

I left the existing `beta` label in place: the PR touches no engine file (nothing under `client/js/widgets/`, `client/js/compute.js` or `client/js/serverstate.js` — `editmode.js` and the sidebar are edit-mode UI), and it is safe and useful for players to try.

96LawDawg (approved):
Approved.  Going to merge.  This is very minor, but should have been included with the line PR.

PR #3061, opened 2026-07-28
Perfect Draw!: update to the author's styled HTML player decks

VTT-AI-Agent:
## What

Public-library update for the existing game **Perfect Draw!** (`library/games/Perfect Draw_/`).

The game's authors sent a new `.vtt` export, forwarded by **rapha195**. It replaces the three default player decks with the game's **styled HTML cards**: instead of a plain spritesheet image, each card face is now built from a background, an icon and `${PROPERTY …}` text fields (`cardName`, `cardText`, `cardFooter`, `cardImage`), styled with the game's Urbanist font. That makes the starter cards directly fillable/editable by players rather than fixed artwork, and the decks grew from 4 to 8 card types.

Only `library/games/Perfect Draw_/0.json` changes — every asset in `assets/` is byte-identical to what is already in the repo and all 14 of them are still referenced, so nothing was added or removed there.

## Metadata

The uploaded export was based on an **older revision of the library metadata**, so its `_meta.info` was *not* taken over. The `_meta.info` currently on `main` is kept verbatim — the cleaned-up attribution with the rights-holder permission note, `bgg` pointing at the itch.io page, `year: "2024"`, and the tidied description / `similarDesigner` / `helpText`. The only metadata field changed is `lastUpdate`, bumped because the game content itself changed.

Two cosmetic export artifacts were normalised back to what `main` had, because they are new validator errors and have no effect on the game: `epl1` gained a `"parent": null`, and the three `xi9bD` cards gained `"x": null, "y": null` instead of `0`.

## Verification

- `node validator/validate_gamefile_node.js "library/games/Perfect Draw_/0.json"` reports exactly the same single pre-existing note as `main` (`_meta.info.bgg` is the itch.io URL, not a BGG one) — no new findings.
- `npm test` passes.
- Loaded the game from the library on a local server: the styled faces render correctly and are usable — flipping a deck's top card shows the HTML face, and it can be dragged onto a Battlefield.

Card close-up and full table after flipping/dragging:

styled card close-up

game table

Credit to **rapha195** for submitting the updated file on the authors' behalf.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3061/pr-test (or any other room on that server)

### DIFF
--- a/library/games/Perfect Draw_/0.json
+++ b/library/games/Perfect Draw_/0.json
@@ -287,7 +287,11 @@
                 "position": "absolute",
                 "left": "12px",
                 "top": "76px",
-                "width": "76px"
+                "width": "76px",
+                "height": "41px",
+                "line-height": "7px",
+                "overflow": "hidden",
+                "overflow-wrap": "break-word"
               },
               ".footer": {
                 "font-size": "6px",
@@ -919,7 +923,11 @@
                 "position": "absolute",
                 "left": "12px",
                 "top": "76px",
-                "width": "76px"
+                "width": "76px",
+                "height": "41px",
+                "line-height": "7px",
+                "overflow": "hidden",
+                "overflow-wrap": "break-word"
               },
               ".footer": {
                 "font-size": "6px",
@@ -2297,7 +2305,11 @@
                 "position": "absolute",
                 "left": "12px",
                 "top": "76px",
-                "width": "76px"
+                "width": "76px",
+                "height": "41px",
+                "line-height": "7px",
+                "overflow": "hidden",
+                "overflow-wrap": "break-word"
               },
               ".footer": {
                 "font-size": "6px",
@@ -2504,7 +2516,11 @@
                 "position": "absolute",
                 "left": "12px",
                 "top": "76px",
-                "width": "76px"
+                "width": "76px",
+                "height": "41px",
+                "line-height": "7px",
+                "overflow": "hidden",
+                "overflow-wrap": "break-word"
               },
               ".footer": {
                 "font-size": "6px",
--- a/library/games/Perfect Draw_/0.json
+++ b/library/games/Perfect Draw_/0.json
@@ -10,7 +10,7 @@
       "mode": "vs",
       "time": "120",
       "attribution": "<div>Game name, design, text, and graphics by Nora Jean Haynes, Iris Cassandra Saintclaire, and Thunderderder. Published by Double Summon Games: https://doublesummon.itch.io/perfect-draw-ttrpg and https://www.fiverr.com/thunderderder. Approved for use on Virtualtabletop.io with permission from the rights holders.</div><div>\n</div><div>Additional graphics from https://game-icons.net/ available as listed on the About overlay.</div><div>\n</div><div>Urbanist Font by Corey Hu, available under SIL Open Font License, Version 1.1.</div><div><br></div><div>Room Layout by Littlemitten and released to the Public Domain under CC0.</div>",
-      "lastUpdate": 1775692826812,
+      "lastUpdate": 1785221852470,
       "skill": "",
       "description": "\"Perfect Draw! is a tabletop role-playing game based on the Powered By The Apocalypse framework that combines Trading Card Games with collaborative storytelling - allowing you to tell stories similar to card game anime like Yu-Gi-Oh!, Duel Masters, and CardFight! Vanguard\"",
       "similarImage": "",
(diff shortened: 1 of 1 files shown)


VTT-AI-Agent (commented):
## TL;DR
Metadata is complete, consistent with the library baseline, and unchanged from the version already on `main` except for the `lastUpdate` bump — nothing to fix here.

*Automated review 1 of 4 (metadata) — by `sonnet`*

## Details

- **Attribution**: unchanged and thorough — credits the designers (Nora Jean Haynes, Iris Cassandra Saintclaire, Thunderderder), the publisher (Double Summon Games) with a rights-holder permission note, the icon source (game-icons.net), the font license (Urbanist, SIL OFL 1.1), and the room-layout author (Littlemitten, CC0). This is a good example of a complete attribution block.
- **Other fields**: name, image, players (2-4), time, language (en-US), year (2024), description, similarDesigner, and helpText are all populated and were already in good shape on `main`; this PR intentionally keeps them verbatim per the PR description rather than taking the older metadata from the uploaded export.
- **bgg field**: points to the itch.io page rather than a literal BoardGameGeek URL. This is a pre-existing, previously-flagged validator note (not introduced by this PR) and is a reasonable choice since the game isn't on BGG — no action needed.
- **lastUpdate**: correctly bumped to reflect the underlying content change (new card templates), which is the only metadata field this PR touches.

No spelling or grammar issues found in the attribution/description text.

VTT-AI-Agent (commented):
## TL;DR
The new HTML-styled starter cards are a clear upgrade over the old plain-spritesheet cards — the layout (icon, name, text, footer) is clean and on-brand with the game's other components. The most impactful further improvement would be giving `cardText` a bounded/overflow-safe height so longer player-entered text can't visually collide with the footer line. This is a public-library game-content submission, not app code, so this review covers the table/game experience only.

*Automated review 2 of 4 (UI/UX) — by `sonnet`*

I started the app locally, added **Perfect Draw!** from the public library, sat down at a seat, and dragged/flipped cards from all three of the updated starter decks.

Table overview: https://agent.virtualtabletop.io/reports/ui-review/1531564370839207958/perfectdraw-table-overview.png
Deck close-up (styled card, blank/generic placeholder, and old-style plain back side by side): https://agent.virtualtabletop.io/reports/ui-review/1531564370839207958/perfectdraw-deck-closeup.png
Drawn card close-up: https://agent.virtualtabletop.io/reports/ui-review/1531564370839207958/perfectdraw-styled-card-closeup.png

### Big layout improvements
- None needed structurally — the change only replaces the visual template of the three starter/player decks, and the new template's internal layout (icon top, name, body text, footer) reads clearly and matches the card style used elsewhere in the game (e.g., the "Establish a Plan to Win" and "Power Card Moves" reference cards on the table use the same font/spacing language). No redesign of the table itself is warranted here.

- `cardText`'s box (`.text`, `top: 76px`, `width: 76px`, no `height`/`max-height`/`overflow`) sits only ~49px above the footer (`top: 125px`) on a 140px-tall card. Since the whole point of this update is that players can type their own name/text/footer into these blank cards, a player who writes more than a line or two of `cardText` will have it visually run into (or under) the `.footer` line instead of wrapping/clipping cleanly. Consider adding `height`/`overflow:hidden` (or a smaller max font-size) to `.text` so it can't collide with the footer.
- The `.image` box allows `max-width:84px / max-height:55px` for the icon, which is roomly proportioned relative to the 100×140 card, but is positioned starting right under the name (`top:16px`) while `.text` starts at `top:76px` — with the current default dragon icon this leaves a good gap, but a very tall/tightly-cropped custom icon (an SVG with unusual aspect ratio) could visually push into the name if `cardImage` is swapped by the game master. Not a blocker, just worth knowing for anyone customizing `cardImage`.
- Base body text is `font-size: 6px` on a 100px-wide card — at 1:1 scale this is unreadable, but the deck's `enlarge: 4` setting (inherited, unchanged by this PR) makes it legible on hover/click, so this isn't a regression, just noting it relies entirely on the enlarge interaction rather than being legible on the table itself, same as before.

- The new card layout is a genuine readability win over the old version: a first-time player immediately sees "Card Name" / icon / "Card Text" / footer as clearly labeled placeholder slots inviting them to type their own values, versus the previous version's static, non-obvious spritesheet artwork. This directly supports the stated goal ("card creation: the blank cards ... have text boxes you can edit").
- The three decks read distinctly at a glance (styled front card, generic-icon "blank" card, and the still-plain-backed deck pile), so a newcomer scanning the table can tell these are different starter piles without needing to open each one.
- One small clarity gap: nothing on the card face itself hints that the text is editable (no placeholder styling like italics/gray "click to edit" affordance) — a first-time player might assume "Card Name" and "Card Text" are literal printed values rather than prompts, until they try clicking. This isn't new to this PR (it's inherent to the property-text-field pattern used elsewhere in the game), so it's a minor observation rather than something this update needs to fix.

Overall this is a solid, well-executed visual upgrade for a public-library content update — nice work matching the existing Urbanist-font styling used elsewhere in the game.

VTT-AI-Agent:
Thanks — addressed the top finding in 
the latest commit. Details below.

### ✅ Fixed: `cardText` could collide with the footer

`.text` on all four styled decks is now bounded to exactly six lines:

```json
".text": {
  "position": "absolute",
  "left": "12px",
  "top": "76px",
  "width": "76px",
  "height": "41px",
  "line-height": "7px",
  "overflow": "hidden",
  "overflow-wrap": "break-word"
}
```

- `line-height: 7px` makes the line box deterministic instead of relying on the font's default leading, so `height` lands exactly on a line boundary — the clip happens in the gap between lines rather than slicing one in half. I checked 41/42/43px against text whose last visible line ends in descenders (`jygq gypq gjyp`); 41px keeps the sixth line's descenders fully visible while nothing of the seventh line shows.
- `overflow-wrap: break-word` covers the other way this box could escape its column: a single long unbroken word (a URL, a long compound) would otherwise stick out past the 76px width.
- With `top: 76px` + `height: 41px` the box ends at 117px, leaving the footer at 125px untouched.

Same long `cardText` on the same deck, **left = with this fix, right = before**:

cardText overflow: before/after

Short/default card text renders identically to before — the box only starts doing anything once the text exceeds six lines.

### The other two points

- **`.image` box vs. the name** — I don't think a tall custom `cardImage` can reach the name: `.image` is `position: absolute; top: 16px`, so it's top-anchored and grows *downward* to at most `16 + 55 = 71px`, always below the name at `top: 4px`. The inverse *is* possible (a long `cardName` wraps down onto the icon), but bounding `.name` to one line would clip perfectly ordinary two-line card names, which felt like a worse trade than the rare overlap with the icon's mostly-empty top area. Left as-is deliberately.
- **6px base font / reliance on `enlarge: 4`** — agreed this is inherited and not a regression; the card is designed to be read enlarged, and changing the base size would reflow every card face the authors laid out. Out of scope for this update.

`npm test` still passes and `validate_gamefile_node.js` reports the same single pre-existing `_meta.info.bgg` note as `main`.

96LawDawg:
Abandoned for #3062

Issue #3059, opened 2026-07-27
Six engine bugs found while auditing routine operation parameters


VTT-AI-Agent:
While going through every `if(a.func == …)` block in `client/js/widgets/widget.js` for the routine editor's "this parameter has no effect" warnings (#3034), I ran into six engine problems that are not editor annotations but real bugs. @RaphaelAlvez asked in https://github.com/ArnoldSmith86/virtualtabletop/pull/3034#issuecomment-5096831157 to open them as one issue and tag @ArnoldSmith86.

None of them is fixed anywhere yet - line numbers are against `main` at the time of writing.

### 1. `FLIP` never applies its own `faceCycle` default

`widget.js:1290`

```js
setDefaults(a, { count: 'all', face: null, faceCyle: null, collection: 'DEFAULT' });
```

Note **`faceCyle`** - the default is set on a misspelled key, so `a.faceCycle` stays `undefined` when the operation omits it. `card.js:265` then falls back to the *widget's* own `faceCycle` property, which may be `backward` or `random`. Documentation (and the routine editor) say an omitted `faceCycle` means `forward`.

Fixing the typo changes behaviour for games that rely on the current fallback, so this may need a legacy/migration decision rather than a one-character fix.

### 2. `SELECT` with `max: "all"` empties the collection

`widget.js:1828`

```js
c = c.slice(0, a.max);
```

`slice(0, "all")` is `slice(0, NaN)` - i.e. nothing is selected. `FLIP`, `MOVE`, `MOVEXY` and `ROTATE` all normalise `"all"` to `999999` before slicing; `SELECT` does not. `all` is offered as a value for `max` in the wiki and in the routine editor, so this is easy to hit.

### 3. `SET` with `relation: "+"` degrades to `=` for the rest of the collection

`widget.js:1883`

```js
for(const w of collections[collection]) {
  …
  if(a.relation == '+' && w.get(String(a.property)) == null)
    a.relation = '=';        // <- mutates the operation, not a per-widget variable
```

The assignment happens inside the per-widget loop, so as soon as one widget has a null value, every *later* widget in the same collection is overwritten instead of appended to.

### 4. `SCORE` ignores its own sanitised `round`

`widget.js:1763` validates `round` and warns `round … must be null or a positive integer, assuming null`, but `1774`-`1776` use the raw `a.round`:

```js
const seatRound = a.round === null ? newScore.length + 1 : a.round;
```

[...]

Issue #2854, opened 2026-02-05
removing property from `inheritFrom` doesn't update


ArnoldSmith86:
I removed `css` from the `inheritFrom` in Runway Roll and it didn't update immediately:

```JSON
{
  "type": "holder",
  "id": "Radio Dice Holder 1",

  "x": 418,
  "y": 225,
  "width": 58,
  "height": 58,

  "z": 6,

  "dropLimit": 1,
  "image": "/i/game-icons.net/delapouite/pocket-radio.svg",
  "inheritFrom": {
    "Engine Dice Holder 1": [
      "css",
      "dropTarget"
    ]
  },

  "background": "#add8e680"
}
```

Probably already reported somewhere.

PR #3063, opened 2026-07-28
Engine regression testing: legacy-mode registry, flag matrix and headless op tests

VTT-AI-Agent:
Implements the [engine regression-testing plan](https://agent.virtualtabletop.io/reports/engine-regression-testing.html) — phases 1 to 6 plus the first phase-7 fixture — and replays the new probes against dated revisions of the repository to find behaviour that already changed.

**Findings reports:** [phases 1–3 audit](https://agent.virtualtabletop.io/reports/pr/1531627140448129076/engine-regression-findings.html) · [phases 4–7 audit](https://agent.virtualtabletop.io/reports/pr/1531627140448129076/engine-regression-phase4-7.html)

## The goal

Four legacy modes exist so that fixing an engine bug doesn't break the games that encode the old behaviour — but nothing currently proves *both halves* still work. The Compute sweep, the project's best engine test, pins two modes **on** and never runs the modern path at all, which is the exact opposite of what the Game Settings panel tells authors to do. This PR makes the flag configuration a dimension the suite varies, and makes a per-operation test cheap enough that a bug fix can leave a regression test behind.

## 1 — Legacy-mode registry

`client/js/legacymoderegistry.js` declares each mode once: `since`, `pr`, `interactsWith`, `detect`, `label`, `description`. Everything derives from it:

- **`client/js/editor/sidebar/gameSettings.js`** renders its checkboxes in a loop — four hand-written `addCheckbox` calls with inlined HTML are gone.
- **`server/fileupdater.mjs`** replaces `v18RoutineLegacyModes` / `v19useIframeForHtmlCards` / `v21DisableHolderImageWidget` with one pass over the registry (`v < mode.since && mode.detect(state)`). Same classification for every save — pinned by tests, see below.
- **the test matrix** gets its combinations from `interactsWith`, so it stays linear in the number of modes.

Adding a mode is now one registry entry; the sidebar, the migration and the matrix pick it up. `legacyMode()` also complains about a name the registry doesn't know, which catches typos that used to silently read `undefined`.

## 2 — Flag matrix in TestCafe

`setLegacyMode()` only ever set modes to `true`, and `setRoomState()` keeps a room's `gameSettings` (see `Room.setState`) — so a mode switched on by one test leaked into every test that ran after it in the shared `testcafe-testing` room. `applyLegacy(combo)` sets **every** mode explicitly, false ones included.

`tests/testcafe/legacymatrix.js` runs one probe game through all seven combinations the registry declares (modern, legacy-all, four one-hot, one declared pair) and asserts **per property** rather than by hash, so a failure reads `pushedString in combination only-…` instead of "hash differs". Expectations are written as functions of the combination, so both halves of each mode are pinned:

| probe | mode off | mode on |
|---|---|---|
| `var a = push '1'` | `["4711"]` | `[4711]` |
| `SELECT` by a stored all-digits id | 1 found | 0 found |
| `var sum = +` (no operands) | `0` | `2` |
| `colorContrast '#101010' '0.5'` | `#ffffff` | `#909090` ← only modern when **both** var modes are off, i.e. the declared interaction |

The Compute sweep keeps its recorded hashes and its cost (it still runs in the var-pair combination), but now names that combination instead of inheriting two modes from whatever ran before. Deliberately **not** multiplied by six: 110 ops × 15³ operands × 6 combinations would mean 660 new reference hashes in `compute.js` and a 6× CI bill, and four of those six combinations differ in nothing Compute can observe. `legacymatrix.js` covers the axis at a cost the per-PR job can afford.

## 3 — Layer A: headless engine tests

`tests/client/engine/harness.js` builds a real widget graph in jsdom, runs a routine and returns a **plain object**. `runRoutine(state, routine, { legacy })` + `forEachLegacy` from `matrix.js`; a case costs a few lines instead of a TestCafe fixture, and assertions are per field so an unrelated serialisation change doesn't invalidate anything.

The P0 operations — `SELECT` `SET` `MOVE` `COUNT` `VAR` `IF` `FOREACH` `CALL` — are covered under all seven combinations, plus `falsy-vs-omitted.test.js`, which runs every listed parameter with `0`, `""`, `false`, `null` and omitted and asserts which of those the engine distinguishes. Two currently can't, both recorded with their issue:

- `MOVE.fillTo` — `a.fillTo || a.count`, so `fillTo: 0` moves everything instead of refusing (#2990)
- `FOREACH.in` — `if(a.in)`, so `in: 0` and `in: ''` silently iterate `DEFAULT` instead of nothing (#3059.6)

Known-wrong answers are pinned **as they are today** with an issue reference (`SELECT max:'all'` selects nothing, #3059.2; `SET relation:'+'` leaking its `'='` fallback to later widgets in the collection, #3059.3). Fixing one flips one number in one place, which is the review signal.

## 4 — Layer C: file-updater tests

The updater runs on **every load**, so its heuristics are live code for the whole installed base. `tests/server/fileupdater.test.js` pins which save gets which mode: the base cases, the v18 `/"var |COMPUTE/` regex firing on ordinary label text (a known false positive, now asserted so tightening it has to be deliberate), idempotence, a future-version save being refused, and a frozen fixture set whose classification may only change on purpose.

`flagsForGame()` in `tests/server/fileupdater-util.js` is the only supported way to ask what modes a game has: a game file's checked-in `legacyModes` is pre-updater and therefore always an undercount.

## Audit: behaviour that already changed

The Layer-A probe was replayed against **15 dated revisions** (2023-Q1 → 2026-Q3) via git worktrees, twice each (all modes off / all on). 80 cases; 59 answers identical across 3.5 years. Full write-up in the [findings report](https://agent.virtualtabletop.io/reports/pr/1531627140448129076/engine-regression-findings.html).

**Two changes reached every game with no flag guarding them:**

1. **`count: 0` flipped from "all" to "none"** on `MOVE` / `FLIP` / `ROTATE` — `433efc38`, 2024-09-13, #1466. Every count used to be read as `a.count || 999999`, so a falsy count meant *all*, and `0` wasn't an exotic spelling for it — it was the **default**. #1466 gave `'all'` its own spelling and made `0` mean zero. A game written before that date saying `count: 0` because that was how you asked for everything now asks for nothing. The `'all'` half is unambiguously a fix; the `0` half is the one that altered published games, and its detector would be precise (a pre-#1466 save with an explicit `count: 0` on one of those ops) if a legacy mode is judged warranted. Flagging for a maintainer decision rather than proposing one.

2. **`MOVE` onto anything that isn't a holder threw for 19 months** — #1888 (`29070a36`, 2023-11-01) ended the new collection branch with an unconditional `await target.updateAfterShuffle()`, which only holders have; #2496 (`2a3d6011`, 2025-05-29) added the holder check as a side effect of an unrelated fix. Three different answers to one input — no-op → `TypeError` that aborts the routine → works — with nothing asserting the break or the repair. No flag warranted here (nobody depends on a crash), just the missing test.

Both are pinned in `tests/client/engine/historical-regressions.test.js`. The audit also confirmed the v18 var-parameter change (#2581) behaves exactly as designed: the answer moved for new games and stayed put under `legacy-all` on all five affected probes — the first end-to-end demonstration that the mechanism works. And it confirmed `SELECT max:'all'` has been returning nothing for the entire probed history: #1466 taught `MOVE`/`FLIP`/`ROTATE`/`MOVEXY` what `'all'` means and `SELECT` was left out.

## Smaller changes worth a look

- `onMessage` in `connection.js` and `getMaxZ` / `updateMaxZ` / `resetMaxZ` in `main.js` gained an `export`. These are bundle globals that `widget.js`, `statemanaged.js` and `legacymodes.js` already depend on; the export makes the dependency explicit and is what lets the engine run outside the browser bundle. No behaviour change — the bundle concatenates these files into one module scope either way.
- The file updater no longer throws on a post-v18 save that has no `gameSettings` object (hand-written saves and importers can produce one). Previously the v19/v21 detectors assumed the v18 step had created it.

## 4 — Layer E: the interactions that are not a click

`tests/testcafe/interactions.js` is the first test in the repository that reaches the engine through something other than a button. Free drag (assert the coordinate, the z, and that nothing re-parented), `movable: false`, `dragLimit`, dropping a card into a holder (parent, `dropOffset`, `onEnter`, `enterRoutine`) and out of one (`onLeave`, `leaveRoutine`), appending to a stacked hand (`z: 1,2,3` at `x: 4,44,84`), forming a pile by dropping a card on a card, and seven typing cases. 16 cases, each run with every legacy mode off and again with every one on — 32 tests.

This is the only route to the risk surface of `disableHolderImageWidget`: a holder's prototype swap is invisible to a routine, so no click test can see it.

Two engine behaviours it pins for the first time, both in the falsy-vs-omitted family:

- **Emptying a label removes the property.** `''` is a label's default, and a property equal to its default is deleted — so a routine reading `text` afterwards gets `undefined`, not `''`. The `changeRoutine` is handed `null`, which `SET` turns into a removal of whatever it was writing to.
- **A numeric entry is rewritten under the cursor.** Typing `007` stores the number `7` and the label writes that back into the field, so the next keystroke continues from `7` — typing `007a` leaves `7a`.

## 5 — Layer F: rendering

`tests/testcafe/domsnapshot.js` records, per widget node, the tag, the class list, the box in board units and a whitelist of computed style properties, and compares it with a baseline in `tests/testcafe/dom-snapshots/`. Deterministic by construction: no image tooling, no perceptual threshold, no flake budget. Pixel diffing can be layered on later.

Three fixtures, each captured in both the modern and the `legacy-all` combination:

| fixture | claim |
|---|---|
| `widget-gallery` — one of every state-determined widget type | the two trees must be **byte-identical**, i.e. no legacy mode changes what a player sees |
| `holder-image` | modern: `hasImage`, a `.holderTextOnly` child, `--bgColor`/`--bgImage`. Legacy: none of them |
| `html-card` | modern: `<div class="cardFaceObject html-object-…">`. Legacy: `<iframe>` with the content in `srcdoc` |

`WRITE_DOM_SNAPSHOTS=1` re-records after an intended change; the diff is printed with `json-diff`, so a failure names the node and the property. It runs in Chrome only — computed styles and rounded boxes are browser specific, and per-browser baselines are a follow-up rather than a guess.

## 6 — Layer D: the library as a corpus

`tests/server/corpus.test.js` runs the file updater over **all 571 library variants** on every PR. It asserts invariants, not counts — the library changes in almost every release, so pinning "275 variants get the var modes" would fail on a game submission: the updater terminates, is deterministic, is idempotent, invents no mode the registry does not declare, and adds no mode to a game saved after that mode's behaviour change.

`npm run corpus-census` answers the question the plan says nobody can: what the library looks like **after** the updater has run.

```
library/games: 343 games, 400 variants
  games with at least one legacy mode at load time: 249
  legacy mode                             on disk  at load
  convertNumericVarParametersToNumbers         37      275
  useOneAsDefaultForVarParameters              37      275
  useIframeForHtmlCards                         0        0
  disableHolderImageWidget                     18       69
```

- **73% of the public library is played in legacy mode** (249 of 343 games), not the 30 the files say. The number is a function of the updater's code, so any change to a detector moves it.
- **`useIframeForHtmlCards` really is at zero**, which the plan expected to be "almost certainly non-zero at load time". 28 variants do contain an `html` face object, but every one was saved at v19 or later — after the change — so the detector correctly declines. The corpus therefore provides *no* coverage for this mode, which makes the Layer-F fixture the only thing that exercises it.
- **The over-broad v18 regex costs nothing measurable.** It flags 244 pre-v18 variants and every one of them really does contain a var expression or a `COMPUTE`. The false positive is real (a label reading `var iations` trips it — pinned in `fileupdater.test.js`) but occurs in no published game.

`tests/testcafe/corpusreplay.js` is the browser half: each game is loaded **through the updater**, the modes it was actually given are recorded as part of the output, every button is clicked in a fixed order with a seeded RNG, and engine exceptions are asserted while content errors (a game shipping a card whose deck is missing) are recorded and diffed instead of gating. `CORPUS_FLAGS_OFF=1` adds the flags-off diagnostic. New nightly workflow (`corpus-replay.yml`, 16 shards, `workflow_dispatch`-able, **non-gating** — a biased sample probed by clicking is a floor, not a certificate). Without an explicit selection it samples the library instead of replaying it, so `testcafe-headless-all` stays usable.

## 7 — First cross-feature fixture

`tests/client/engine/inheritfrom.test.js`: a three-level `inheritFrom` chain mutated at each level and read at the bottom, under the full matrix. The harness gained two more bundle globals (`widgets`, `dropTargets`) to make it reachable.

## Second audit: the new probes against 14 dated revisions

The Layer-E and Layer-F probes were replayed against **14 dated revisions, 2023-04 → 2026-07**, each in its own worktree with its own `node_modules` and its own server, in both flag states where the flags existed. 15 probes per revision. Full write-up: **https://agent.virtualtabletop.io/reports/pr/1531627140448129076/engine-regression-phase4-7.html**

**Every interaction answer is unchanged since 2023-04.** Free drag lands on the same coordinate; `movable: false` still refuses; `dragLimit` still clamps to the same corner; a card dropped into a holder still lands on `dropOffset` and still fires `onEnter` before `enterRoutine`; leaving still fires `onLeave` and `leaveRoutine`; a stacked hand still renumbers `z: 1,2,3` at `x: 4,44,84`; a card on a card still makes one pile at the stationary card's position; typing still converts a digit string to a number, still leaves `1e3` a string, and still deletes the property when the field is emptied. Unlike the phase 1–3 audit, this one found **no unguarded behaviour change** on these axes.

**Rendering changed twice, and both changes are behind a legacy mode:**

| probe | 2023-04 … 2025-10 | 2026-01 | 2026-04 … today | guarded by |
|---|---|---|---|---|
| html card face object | `<iframe>` | `<div>` | `<div>` | `useIframeForHtmlCards` (#2729, `fdf7dd0e`, 2025-11-15) |
| holder `hasImage` class | absent | absent | present | `disableHolderImageWidget` (#2634, `10d3fa2b`, 2026-02-11) |
| holder `.holderTextOnly` child | 0 | 0 | 1 | ″ |
| holder `--bgColor` / `--bgImage` | absent | absent | present | ″ |

The result worth having: **the rendering snapshot taken today with `legacy-all` is byte-identical to the snapshot taken at every revision before the change** — all seven recorded fields. The two rendering legacy modes do not merely exist; they measurably restore the rendering of 2023, and `domsnapshot.js` is what keeps that true.

## 8 — Holder events: the order #1212 wrote down, as it actually is

`tests/testcafe/holderevents.js`, 22 cases. #1212 described what dragging a widget in and out of a holder *should* do and nothing has ever asserted what it *does*, so this pins the status quo — including the parts that disagree with the description — and names the difference at each case.

Every holder routine appends one entry to a log widget, and the entry carries the properties of the widget the event is about, read at that moment. So an entry says both what fired and how much of the event had already happened when it fired: `mark=null` means "this ran before the property half", `parent=handB` means "the parent was already written".

| what the player does | what fires today | what #1212 asks for |
|---|---|---|
| drag out of a holder | `leave`(mark=null), `leave`(mark=leave-A) | `beforeLeaveRoutine`, then `leaveRoutine` |
| drag A → B | the two leaves, then `enter B` after parent, onEnter and alignment | + `afterLeaveRoutine` after the arrival |
| drag within a holder | `leave`, `enter` — and **onEnter is not applied**, only the routine runs | `beforeLeaveRoutine` + `leaveCancelledRoutine` |
| drag out and back | the two leaves, then `enter` — indistinguishable from arriving from elsewhere | `leaveCancelledRoutine` with a flag |
| `MOVE` A → B | the two leaves, then `enter` | onLeave+leave, arrival, afterLeave, enter |
| `MOVEXY` out | **one** leave, and **onLeave is never applied** | onLeave, leave, afterLeave |
| `SET parent` | one leave, which reads `parent=handB` — the *new* parent | — |
| `DELETE` in a holder | one leave, which still reads `parent=handA` | — |

The doubled `leaveRoutine` is the issue's headline and it has two distinct sources: `onPropertyChange('parent')` calls the old parent's `leaveRoutine` directly when the drag detaches the widget (`widget.js:2766`), before `onLeave` is applied; `checkParent()` → `dispenseCard()` applies `onLeave` and calls it again once the widget stops overlapping the holder. Splitting those into `beforeLeaveRoutine` and `leaveRoutine` is exactly what the issue proposes, and this file is what will show, per path, what such a change moves.

Also covered because they are part of the same event: `childrenPerOwner` (the owner is written before `onEnter`), `alignChildren`, `stackOffsetX`, `dropLimit`, `ignoreOnLeave`, `dropShadow` (a holder runs its enter *and* leave events for the shadow widget, for something the player never dropped), a pile dropped into a holder (`onEnter` reaches every card inside it, `enterRoutine` fires once for the pile), `RECALL`, the `child` collection and `oldParentID` both routines are handed, and the whole holder-to-holder and `MOVE` order again with every legacy mode on.

## 9 — Everything that was listed as not done

- **Canvas drawing** — `tests/testcafe/canvas.js`, 10 cases. The raw pointer stream a canvas takes instead of clicks and drags, asserted on both sides of the round trip: the pixel the 2D context ends up showing (decode) and the run-length encoded region string in the room state (encode, i.e. what a save file holds). Plus `artist`, `clickable`, `lineWidth`, the resolution clamp, and `CANVAS` `setPixel`/`set`/`reset`.
- **Multi-client** — `tests/testcafe/multiclient.js`, 5 cases, two windows in one room. A change made in one arrives in the other; two clients acting on different widgets converge on the same board; `onlyVisibleForSeat` and `owner` hide a widget from the other player but not from its own; an `INPUT` overlay opens on the asking client only, and the other one is still connected after the answer. Needs `--disable-native-automation`, which the Chrome matrix now passes per file.
- **Keyboard / shortcuts / zoom** — `tests/testcafe/keyboardzoom.js`, 12 cases. Hotkeys including which widget wins when two share a key and the three situations the room ignores the key in, then zoom: `PageUp`/`PageDown`, `ignoreZoom`, and — the point of the section — a drag and a drop into a holder at **2x**, where the coordinate transform between the pointer and the engine can be wrong, rather than only at 1x, where it cannot.
- **The object form of `inheritFrom`** — `tests/testcafe/inheritfrom.js`, 9 cases: a property list, an exclusion list, `'*'`, two sources (the first listed wins), a change at the top of a chain, and deletion of the source. That last one found that the two halves of a widget disagree: the DOM keeps rendering the inherited value while `get()` already returns the widget's own default, so the player reads one thing off the table and a routine reads another (#2854).
- **Pixel diffing** — `tests/testcafe/pixelsnapshot.js` + `png-util.js`. Not as a checked-in golden PNG: that baseline would be recorded on one machine and compared on a runner whose Chrome version, fonts and GPU stack move without notice, which is how screenshot jobs get switched off. Instead it compares two screenshots **from the same run** of the same board in the two legacy-mode combinations. The widget gallery has to come out **pixel identical** (it does), and the two rendering fixtures have to come out different by more than an anti-aliased edge — 46187 and 710 pixels today, which is the first assertion anywhere in this repository that `useIframeForHtmlCards` and `disableHolderImageWidget` change anything a player can *see*. A fourth case shoots the same board twice and requires the two images to be identical, so a fixture that starts animating fails as itself. PNG decoding is ~60 lines against node's own `zlib` rather than two new dependencies.

## Not in this PR

- **A Firefox DOM baseline.** The baselines are now keyed by browser (`widget-gallery.chrome.json`), which is the mechanism a second one needs, and a missing baseline fails with the browser it has to be recorded on. The file itself has to be recorded on a machine that can run Firefox; this sandbox has no GTK, so Firefox does not start here and a hand-written baseline would be a guess. `pixelsnapshot.js` needs no baseline at all, so **that** half of Layer F is in the Firefox matrix now.
- **Canvas rendering as a golden image**, for the same reason the pixel layer compares within a run rather than against a recording.

## Verification

- `npm test` — **640 tests, 16 suites, all green** (595 at the previous head, 86 before this PR).
- `npx testcafe chromium:headless tests/testcafe/interactions.js` — 32/32.
- `npx testcafe chromium:headless tests/testcafe/domsnapshot.js` — 3/3, and stable across repeated runs.
- `npx testcafe chromium:headless tests/testcafe/legacymatrix.js` — 7/7.
- `npx testcafe chromium:headless tests/testcafe/corpusreplay.js` — the 18-variant default sample, all green; plus `CORPUS_GAMES=Blue,Agon CORPUS_FLAGS_OFF=1` for the diagnostic path.
- `npm run corpus-census` — 3s over 571 variants.
- `holderevents.js` 22/22, run twice for stability · `canvas.js` 10/10 · `keyboardzoom.js` 12/12 · `inheritfrom.js` 9/9 · `multiclient.js` 5/5 · `pixelsnapshot.js` 4/4 · `domsnapshot.js` 3/3 after the rename.
- Workflow matrices: the five new files in Chrome (`multiclient.js` with `--disable-native-automation`), `inheritfrom.js` and `pixelsnapshot.js` in Firefox as well. `holderevents.js`, `canvas.js` and `keyboardzoom.js` stay in Chrome for now — their drags are synthetic mouse events whose timing against the drop animation has only been measured there, and Firefox cannot be run in the sandbox to check.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3063/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

Solid, well-motivated refactor + a large amount of genuinely new regression coverage; I verified the file-updater consolidation is behaviour-equivalent on the cases I could construct, the concatenated room bundle still parses, and `npm test` gives 595/595 green. The one thing I'd fix before merge: `legacymatrix.js` turns **all four** legacy modes on and never turns them back off, so in a full local `testcafe-all` run it leaks `legacy-all` into every fixture that follows it — the exact cross-test contamination this PR was written to eliminate.

## Findings

### 1. `legacymatrix.js` leaks its legacy modes into every later fixture (test infrastructure)

`tests/testcafe/legacymatrix.js:76-77` runs seven combinations, the last of which sets modes on, and nothing resets them. `setupTestEnvironment()`'s `.after(_=>setRoomState())` only clears widgets: `Room.setState` at `server/room.mjs:1208-1215` deliberately carries `gameSettings` over when the incoming state has no `_meta`, which is exactly the leak the PR documents in `test-util.js:60-62`.

In CI this is invisible (one file per matrix job), but `npm run testcafe-chromium-headless-all` globs `tests/testcafe` alphabetically: `compute-1..4`, **`legacymatrix`**, `publiclibrary-1..4`, `routines`. So `publiclibrary-*.js` and `routines.js` now run with `useIframeForHtmlCards` and `disableHolderImageWidget` on — two modes that change `card.js` face rendering and the whole of `holder.js` — against hashes recorded with them off. That's a new local-only false failure, and it's the same bug class the PR fixes for the Compute sweep.

Cheapest fix that also removes the leak for everyone: reset to the modern combination in the shared setup, e.g. in `setupTestEnvironment()`'s `beforeEach` (`_=>{ await applyLegacy('modern'); await setRoomState(); }`), or at minimum add `.after(_=>applyLegacy('modern'))` to the `legacymatrix` fixture.

### 2. `flagsForGame()` throws on a save that is already at the current version

`tests/server/fileupdater-util.js:9` does `FileUpdater(...)._meta.gameSettings.legacyModes`, but `FileUpdater` returns early at `server/fileupdater.mjs:5-6` when `v == VERSION`, without creating `gameSettings`:

```
flagsForGame({_meta:{version:21}, l:{type:'label'}})
→ TypeError: Cannot read properties of undefined (reading 'legacyModes')
```

The PR text bills this as "the only supported way to ask what modes a game has", and `flagsForGameFile()` is aimed at checked-in game files — a v21 file saved without a `gameSettings` object (hand-written / importer output, the same case the updater fix at `fileupdater.mjs:85` exists for) crashes the helper instead of returning `{}`. `return (FileUpdater(...)._meta.gameSettings || {}).legacyModes || {}` covers it, and it's worth a test since the suite already has one for the post-v18-without-gameSettings case.

### 3. `fullLegacyCombination()` is dead code, and its logic is duplicated twice

`client/js/legacymoderegistry.js:132-138` is exported and documented but has no callers. Meanwhile `tests/testcafe/test-util.js:69-70` and `tests/client/engine/harness.js:53-56` each open-code the same "iterate `ALL_LEGACY_MODES`, coerce to boolean" loop it describes. Either use it in both places or drop it — as written it ships in the room bundle for every player and is never called.

### 4. Behaviour change worth being explicit about: `_meta.gameSettings` is now always created

Old `updateMeta` only created `gameSettings` for `v < 18`; the v19/v21 steps mutated it in place (and would have thrown if absent — the crash fix here is good). New `updateLegacyModes` (`fileupdater.mjs:84-88`) creates `{ legacyModes: {} }` for *every* below-current-version save. So a v18–v20 save that previously came out of the updater with no `gameSettings` key now gains an empty one. I traced it to `room.mjs:1211-1215` and `fileloader.mjs:113-116` and it collapses to the same value, so I believe it's inert — but it changes the serialised `_meta` shape of a class of saves and isn't mentioned in the PR description. Worth a sentence there, and it's the kind of thing the TestCafe reference hashes would catch if it weren't inert (CI will confirm).

Related, and unchanged-but-now-centralised: `if(v < 18 ...)` still *replaces* `meta.gameSettings` wholesale, discarding any sibling keys (`globalCss`, `cursorVisibility`) a pre-v18 save might carry. Unreachable for genuine v17 files, but now that all four modes funnel through one line it'd cost nothing to merge instead of replace.

### 5. Minor: `docs/compat.md` doesn't exist in this repo

`client/js/legacymoderegistry.js:5` — "see docs/compat.md for the checklist". There is no `docs/` directory in the repository; that file lives only in the agent workspace. It's the sole reference of its kind in the codebase, so a human contributor following the comment finds nothing. Either point at the wiki's Legacy-Mode page (which the descriptions already link) or inline the checklist into the comment block above.

### 6. Minor: ~2 KB of editor-only HTML moved into the room bundle

The four `description` fields (1,981 bytes of markup) previously lived in `client/js/editor/sidebar/gameSettings.js`, which is only in the editor bundle. `minify.mjs:51` puts the whole registry into the room bundle, where `legacymodes.js` needs `ALL_LEGACY_MODES`/`LEGACY_MODES` — so every player now downloads the sidebar copy. Small, and the consolidation is worth more than 2 KB, but if you want it back: keep `since`/`pr`/`interactsWith`/`detect` in the registry and a `LEGACY_MODE_DESCRIPTIONS` map in the editor bundle keyed by the same names.

## What I checked and found clean

- The concatenated room bundle still parses after the four new `export` keywords — no duplicate-export collision for `onMessage`/`getMaxZ`/`resetMaxZ`/`updateMaxZ`, and `removeImportStatements`' `/^import\s+[^;]+;/` strips the three new `legacymodes.js` imports cleanly.
- `npm test` → 14 suites, 595 tests, all passing, no jsdom warnings or open handles.
- Spot-checked the updater against v17/v18/v19/v20 states with holder/deck/var shapes; classification matches the old per-version functions in every case I tried, and the null-widget guards in the new `detect` functions are strictly more robust than the old `widget.type` / `face.objects` dereferences.
- `legacyMode()`'s new unknown-name `console.error` can't fire from any current call site (all seven are registry names), and `getEnabledLegacyModes()`'s reordering only feeds a badge count.
- Registry-order iteration in `gameSettings.js:346` preserves the previous checkbox order.
- Pinning known-wrong answers (`SELECT max:'all'`, `MOVE fillTo:0`, `FOREACH in:0`) with issue references rather than asserting the desired behaviour is the right call for a regression harness — it makes a fix a one-line visible diff.

I added the `beta` label: no engine path is touched (`client/js/widgets/`, `client/js/compute.js` and `client/js/serverstate.js` are all untouched, and the `main.js`/`connection.js` edits are `export` keywords that are no-ops inside the concatenated bundle), the only player-visible changes are the registry-driven Game Settings sidebar and a file-updater consolidation I verified is classification-equivalent, and my one substantive finding is confined to the TestCafe suite where it cannot reach a beta player.

VTT-AI-Agent:
Thanks - all five actionable points are addressed in 5ca974d.

**1. `legacymatrix.js` leaks its modes (confirmed, fixed).** You are right that this is the same leak the PR set out to remove, just one level up. I took the shared-setup option rather than a `.after()` on the one fixture: `setupTestEnvironment()` now runs `setRoomState(); applyLegacy('modern')` both before each test and after the fixture, so *every* fixture starts from an empty room in a known combination and no test can hand a mode to the next one - within a file or across files. Verified locally on port 8324: `legacymatrix.js` + `routines.js` in one run (11 passed - that combination failed before the fix), `publiclibrary-1.js` (5 passed, so games that get their modes from `_meta` are unaffected by the reset) and `editor.js` (14 passed).

**2. `flagsForGame()` throws on a current-version save (confirmed, fixed).** Now `(… .gameSettings || {}).legacyModes || {}`, with a test for the v21-without-`gameSettings` case next to the existing post-v18 one.

**3. `fullLegacyCombination()` dead code (confirmed, fixed).** Used it in both places instead of dropping it - `test-util.js:applyLegacy()` and `tests/client/engine/harness.js:setLegacyModes()` now iterate its result, and neither imports `ALL_LEGACY_MODES` any more.

**4. `_meta.gameSettings` always created (confirmed, reverted to the old shape).** Rather than documenting the change I removed it: `gameSettings` is created on demand, so a v18-v20 save that trips no detector now comes out of the updater with the `_meta` it went in with, exactly as before this PR. The pre-v18 branch merges instead of replacing, so a `globalCss`/`cursorVisibility` sibling survives. Three tests in a new `_meta shape` block pin all of it, including that an existing `legacyModes` object is added to rather than replaced.

**5. `docs/compat.md` (confirmed, fixed).** That path only exists in my workspace. The comment now spells out the checklist inline (add the entry, call `legacyMode('name')` at the site whose behavior it switches) instead of pointing at a file a contributor cannot open.

**6. ~2 KB of descriptions in the room bundle - not changing.** Splitting `description` back into an editor-only map keyed by mode name re-creates exactly the two-places-to-edit situation the registry removes, and the drift it invites (a mode with no description, or a description for a mode that no longer exists) costs more than 2 KB of gzipped markup. Happy to do it if you would rather have the bytes.

`npm test` is 599/599 green (four new tests).

VTT-AI-Agent (commented):
## TL;DR

**UI/UX review, not a code review.** This PR is overwhelmingly test/infra work; its only user-visible surface is the **Game Settings → Legacy Modes** panel, which is now rendered in a loop from `LEGACY_MODES` instead of four hand-written `addCheckbox` calls. I drove it in a real browser (light + dark, all four modes on, one off, and a game with none) and the rendered result is **pixel-equivalent to `main`** — no regression. The most impactful improvement now available *because* of this PR: the registry makes the panel data-driven, so give each mode a short one-line `summary` and collapse the long description behind a disclosure — today four checked modes produce ~900px of unbroken prose before you reach anything actionable.

Rendered against this branch on a local server:

| all four modes on (light) | one mode off (Remove section) | dark mode | game with no legacy modes |
|---|---|---|---|
| [](https://agent.virtualtabletop.io/reports/ui-review/1531650592018010213/gs-panel-top.png) | [](https://agent.virtualtabletop.io/reports/ui-review/1531650592018010213/gs-tile-unchecked.png) | [](https://agent.virtualtabletop.io/reports/ui-review/1531650592018010213/gs-dark-tile.png) | [](https://agent.virtualtabletop.io/reports/ui-review/1531650592018010213/gs-fresh.png) |

Full-panel shot with all four tiles: [gs-legacy-panel-full.png](https://agent.virtualtabletop.io/reports/ui-review/1531650592018010213/gs-legacy-panel-full.png)

Verified as rendering unchanged: labels, order (registry key order matches the old hand-written order), checked state, tile highlight, PR links, wiki link, the `Remove` section, and the sidebar badge (`4`). The badge is now computed off `ALL_LEGACY_MODES`, so a stale key left in a save's `legacyModes` no longer inflates the count — a small, correct behaviour change worth calling out because it *is* visible.

Everything below is a suggestion on the panel this PR now owns, not an objection to merging.

---

## Big layout improvements

### 1. Collapse the descriptions — the panel is a wall of text

With all four modes active the tiles measure 308 / 206 / 172 / 223 px, so ~910px of prose sits between the section header and the next thing (`UI Settings`), and the whole panel scrolls (`#editorModuleTopLeft` scrollHeight 1601 vs 1267 visible at 900px viewport). A user opening this panel wants to answer *"which modes are on, and can I turn them off yet?"* — that answer requires scrolling past three multi-paragraph essays.

Proposal, cheap now that `client/js/legacymoderegistry.js` is the single source: add a one-line `summary` field per mode and render

```
[x] Convert numeric var parameters to numbers            [Active]
    Digit-only strings in var expressions became numbers.   ▸ Details
```

with the existing `description` inside a `<details>` (collapsed by default). Four scannable rows instead of a 900px scroll, and the full explanation is one click away for anyone who needs it. The registry already has exactly the right shape to hold `summary` next to `label`/`description`.

### 2. Same treatment for the three intro paragraphs

`gameSettings.js:334-344` prints three paragraphs (p3 alone is five lines at panel width) *above* every checkbox, every time the panel is opened. Once a game author has read them once they are pure scroll cost. Suggest collapsing them into a `<details><summary>What are legacy modes?</summary>` block, or keeping only p1 inline and moving p2+p3 behind it. That alone lifts the first checkbox ~170px up the panel.

### 3. Sort active modes first, and state the summary explicitly

The loop at `gameSettings.js:346` renders in registry declaration order. The actionable ones are the *checked* ones. Consider `Object.keys(LEGACY_MODES).sort((a,b) => legacyMode(b) - legacyMode(a))` so active modes lead, plus a single line under the header — `3 of 4 legacy modes active for this game` — which is the number the sidebar badge is already showing and which currently has no textual equivalent anywhere in the panel.

---

- **Clicking anywhere in a tile flips a game-behaviour flag, including while selecting text.** `gameSettings.js:102-107` toggles the checkbox for any click whose target isn't `A` / `LABEL` / the Remove button. I reproduced this: a mouse-down → drag → mouse-up inside the description text of the first tile flipped `convertNumericVarParametersToNumbers` from `true` to `false` and selected nothing. Combined with `cursor: pointer` on the whole tile (`:25`), the description text never even shows an I-beam, so the tile reads as "not selectable" while being the very thing a user wants to copy (`var a = push '1'`). Suggest limiting the click target to the header row (`header`), or bailing out when `getSelection().toString()` is non-empty on mouseup, and setting `cursor: default` on `desc`.
- **`desc.querySelectorAll('a').forEach(link => link.style.pointerEvents = 'auto')` (`:56-58`) is a no-op.** Nothing sets `pointer-events: none` anywhere in the tile; the comment says "Prevent link clicks from toggling the checkbox" but the thing that actually does that is the `A` check in the click handler. Since this function is now the single rendering path for all four tiles, it's a good moment to delete the misleading two lines.
- **Checked vs unchecked is carried almost entirely by background colour.** Measured: `rgb(204,204,204)` checked vs `rgb(255,255,255)` unchecked in light mode — a difference that is easy to miss, and the only other signal is a 13px native checkbox. A left accent border (`border-left: 4px solid var(--primaryColor)`) or an explicit `Active` chip next to the label would make the state readable at a glance and survive greyscale/low-vision use. (Dark mode is fine: `rgb(68,68,68)` vs `rgb(8,9,10)` reads clearly.)
- **Alignment: card text is indented 12px relative to the intro prose.** The `p1/p2/p3` paragraphs and the `Legacy Modes` subheader bar start flush at the panel edge, while the tiles add `padding: 12px` inside a 1px border — so the left text edge steps in and out down the panel. Giving the intro paragraphs the same 12px horizontal inset would tidy the column.
- **The Remove button is the loudest element in the panel.** A full-width-height dark-red `REMOVE` button per tile (`:78-81`, `className = 'red'`) appears on *every* unchecked mode, so a well-behaved game that has disabled all four modes shows four red destructive buttons. Given it's already behind a `confirm()`, a subdued text-style link ("Remove this setting") would match the low-stakes nature of the action and stop the panel from looking alarming.
- **Dark mode, light mode and the link colour (`rgb(158,158,255)` on `rgb(8,9,10)`) all check out** — no contrast problems found in either theme, and no console errors during render.

---

- **The four descriptions use three different vocabularies for the same concept.** `Problem` / `Old result` / `New result` (modes 1–2), `Legacy Behavior` / `Default Behavior` (iframes), `Old behavior` / `New behavior` (holders) — see `legacymoderegistry.js:32-40`, `:57-64`, `:85-88`, `:105-112`. Now that all four strings live side by side in one file, normalising to one pair (`Old behavior` / `New behavior`, plus an optional `Example`) is a two-minute edit and makes the panel read as one document instead of four.
- **Dead / self-contradicting copy at `gameSettings.js:342`:** "If there are no checkboxes below, legacy mode is not applicable for your game." The whole intro block is inside `if(Object.keys(...legacyModes).length > 0)`, so in the normal case the reader only ever sees this sentence when there *are* checkboxes. The one state where it does show — a save carrying only a key the registry doesn't know — renders a header plus three paragraphs and zero tiles, which is exactly the confusing outcome. Suggest gating the header/intro on the number of modes that will actually render (you can compute it from `LEGACY_MODES` now) and dropping the sentence.
- **Typo in the same paragraph:** "We highly **recommended** you build and test…" → "recommend".
- **The empty state says nothing.** For a game with no legacy modes the panel jumps straight to `UI Settings` (screenshot 4) — a first-time author gets no confirmation that "your game uses no legacy modes", only silence. A single muted line, `No legacy modes are active for this game.`, turns an ambiguous absence into a positive all-clear and costs one `<p>`.
- **The sidebar badge is an unlabelled yellow number over a gear icon.** The button tooltip is `'Settings like legacy modes and global game options.'` (`:3`). Since `getEnabledLegacyModes()` is right there, appending the count to the tooltip — "3 legacy modes active" — would explain what the badge counts without opening the panel.
- **Jargon check for a newcomer:** "var expression", "`SELECT` it using the stored `id`", "card face objects with `type: 'html'`", "holder" all appear before any of them is defined. The wiki link at the end of each description carries that load, but it is the *last* thing in a 300px block. Moving `See … Legacy Mode wiki page` up next to the label (or into the proposed `Details` disclosure header) puts the escape hatch where a confused reader is, rather than where they'd arrive after giving up.

VTT-AI-Agent:
Thanks — the panel is much better for it. Everything actionable is in e25f3ad8, except the re-sort (one reason below). Verified in a real browser on a local server, light + dark, with three modes on / one off and with a game that has none.

| collapsed (3 on, 1 off) | one tile expanded | dark mode | game with no legacy modes |
|---|---|---|---|
| [](https://agent.virtualtabletop.io/reports/pr/1531627140448129076/gs-panel-collapsed.png) | [](https://agent.virtualtabletop.io/reports/pr/1531627140448129076/gs-details-open.png) | [](https://agent.virtualtabletop.io/reports/pr/1531627140448129076/gs-dark.png) | [](https://agent.virtualtabletop.io/reports/pr/1531627140448129076/gs-empty.png) |

**1. Collapse the descriptions (done).** Each registry entry now has a `summary` — one line saying what changed — rendered under the label, with the full `description` plus the PR/wiki links inside a `<details>Details</details>`. An active tile went from 308 / 206 / 172 px to **88 px each**; the whole Legacy Modes section is now four scannable rows.

Since `renderModule()` runs again on every `meta`/`state` message, an open disclosure would have snapped shut every time you ticked a box, so the module remembers which ones are open (`this.expandedModes`) and restores them. Confirmed: expand a tile, toggle a mode, the tile is still open.

**2. Intro paragraphs (done).** p1 stays inline (one sentence), p2 + p3 moved behind `<details>What are legacy modes?</details>`, also with its open state remembered.

**3. Count line (done) — re-sort (not done).** `3 of 4 legacy modes active for this game.` sits directly under the section header, in bold. I left the order alone on purpose: because the panel re-renders on the meta update that a toggle itself causes, sorting by active would make the tile you just clicked jump to the other end of the list under your cursor. The accent border and the `Active` chip (below) make the active ones findable without moving them.

**4. Click-anywhere toggling (done).** The listener moved from the tile to the header row, so the summary/description text is now selectable and dragging across it no longer flips a flag — reproduced your drag on the description before and after: it toggled `convertNumericVarParametersToNumbers` before, it doesn't now. `cursor: pointer` moved to the header with it, so the prose shows an I-beam.

**5. `pointerEvents = 'auto'` no-op (done).** Deleted, along with the comment. The `A` guard in the click handler went with the tile-level listener; links live in the description, which is outside the header, so nothing intercepts them.

**6. Checked vs unchecked (done).** Active tiles get a 4px `--textHighlightColor2` (orange) left border and an `Active` chip next to the label; the background difference is now the third signal rather than the only one. The border slot is reserved as `transparent` on inactive tiles so nothing shifts when you toggle.

**7. Alignment (done).** The intro paragraphs, the count line and the disclosure get the same 12px horizontal inset as the tile padding, and the summary/description inside a tile are indented to the checkbox's text column.

**8. The red REMOVE button (done).** Dropped `className = 'red'`; it's a normal icon button now, still behind the `confirm()`.

**9. Vocabulary (done).** All four descriptions are `Old behavior` / `New behavior` (+ `Example` where there is one), the "Problem" / "Legacy Behavior" / "Default Behavior" variants are gone, and the field comment in the registry asks the next contributor to keep it that way. The holder entry also now says *why* the new behavior breaks those games (both drawn at once) rather than just that it might.

**10. `gameSettings.js:342` (done).** The header + intro are now gated on the modes that will actually render (`Object.keys(LEGACY_MODES).filter(name => legacyMode(name) !== undefined)`), so the "header + three paragraphs + zero tiles" state is gone and the sentence with it. `recommended` → `recommend`.

**11. Empty state (done).** A game with no legacy modes now gets the `Legacy Modes` header and `No legacy modes are active for this game.` instead of silence.

**12. Badge tooltip (done).** `updateBadge()` appends `N legacy modes active.` to the sidebar button's tooltip, and drops it again when the count is zero.

**Jargon** — partly. The wiki link is still at the end of the description rather than next to the label, but the description is now one click below the label instead of 300px down it, so the escape hatch is much closer to a confused reader. Moving the link into the header row as well felt like it would compete with the checkbox for the most important click in the tile.

`npm test` 599/599 green; `tests/testcafe/editor.js` 14/14 passing locally on port 8324.

VTT-AI-Agent (commented):
## TL;DR

The latest head looks good: the earlier cross-test legacy-mode leak and file-updater edge cases are fixed, the new regression layers exercise both modern and legacy behavior, and all current CI checks are green. I found no additional correctness, security, or regression issue worth raising.

## Findings

No actionable findings.

I re-checked the registry-driven migration against the previous per-version detectors, including pre-v18 settings preservation, current-version saves without `gameSettings`, and existing `legacyModes`; the classifications and default game behavior remain intact. The shared TestCafe reset now explicitly applies the modern combination before every test and after the fixture, so later tests cannot inherit the matrix's final flags.

The engine-adjacent changes in `client/js/connection.js` and `client/js/main.js` only export existing bundle globals for the headless harness; they do not change runtime behavior. The PR does not modify `client/js/widgets/`, `client/js/compute.js`, or `client/js/serverstate.js`, and the legacy registry/sidebar consolidation does not alter how existing games execute.

The `beta` label was already present; it remains appropriate because the runtime changes preserve existing game behavior, the new UI is safe to exercise, and all TestCafe, build, CodeQL, and deployment checks are green.

VTT-AI-Agent:
Thanks for the re-check — nothing to change on my side for automated review 3/4.

Confirming the three points you verified, since they were the ones the earlier rounds flagged:

- **Cross-test leak** — `tests/testcafe/test-util.js` now applies the modern combination (all legacy modes off) before every test *and* in the fixture teardown, so the flag matrix cannot bleed into later tests regardless of ordering.
- **File-updater migration** — the registry-driven classification is behaviour-identical to the previous per-version detectors for pre-v18 saves with settings, current-version saves without `gameSettings`, and saves that already carry `legacyModes`. Covered by the jest cases in `tests/fileupdater.test.js`.
- **Engine surface** — the only edits under `client/js/` outside the settings panel are the two bundle-global exports in `connection.js` / `main.js` for the headless harness; no widget, `compute.js` or `serverstate.js` behaviour changes. The `engine` label is there because the diff touches engine-adjacent client files, not because runtime semantics moved.

On the `beta` label: it was applied by a reviewer, not by me, and I have left it as-is.

Head is still `e25f3ad8`; no new commits for this round. `npm test` 599/599 green.

96LawDawg (commented):
The new Legacy Mode side panel UI change is a good improvement and looks and works great. I have no idea about the rest of any of this.

VTT-AI-Agent:
Pushed `57939e79` — the rest of the plan, and the new probes replayed against history.

> Implement the rest of the plan as well. Do your best to execute _all_ of it against historical revisions.

*(Discord, [link](https://discord.com/channels/770758631146782780/1531627140448129076/1531721061442256986))*

**Phases 4–7.** Layer E (`tests/testcafe/interactions.js`) drives drag, holder enter/leave, stacked-hand ordering, pile formation and typing — 16 cases in both flag states, and the only route to the risk surface of `disableHolderImageWidget`. Layer F (`tests/testcafe/domsnapshot.js`) is the first rendering assertion in the repo: a computed-DOM baseline per fixture, with the widget gallery required to render *identically* with every mode on and off. Layer D is the library run through the updater — a per-PR jest test of invariants plus `npm run corpus-census` and a nightly, non-gating browser replay. `inheritfrom.test.js` is the first cross-feature fixture.

**The historical replay.** 14 dated revisions (2023-04 → 2026-07), each in its own worktree with its own server, 15 probes, both flag states where they existed. Write-up: https://agent.virtualtabletop.io/reports/pr/1531627140448129076/engine-regression-phase4-7.html

- Every interaction answer is **unchanged since 2023-04** — coordinates, holder alignment, routine firing order, stack renumbering, pile formation, typing coercion. No unguarded change on these axes, unlike the phase 1–3 audit.
- Rendering changed twice, both behind a flag: html card faces stopped being iframes in #2729 (2025-11-15), holders started drawing image/icon/text in #2634 (2026-02-11). And the thing worth having: **today's `legacy-all` rendering snapshot is byte-identical to the snapshot from every revision before those changes**, all seven recorded fields. The rendering modes measurably restore 2023, which is what `domsnapshot.js` now keeps true.

**Three corpus findings**, all from `npm run corpus-census`:

- **73% of the public library is played in legacy mode** — 249 of 343 games at load time, against the 30 the files say.
- **`useIframeForHtmlCards` is at zero even after the updater.** The plan expected it to be "almost certainly non-zero". 28 variants use an `html` face object, but all of them were saved at v19 or later, so the detector correctly declines. The library gives that mode *no* coverage at all, which makes the Layer-F fixture the only thing exercising it.
- **The over-broad v18 regex costs nothing measurable**: it flags 244 pre-v18 variants and every single one really does contain a var expression or a `COMPUTE`.

`npm test` is 640/640 green (595 before). `interactions.js` runs in both TestCafe workflows; `domsnapshot.js` in the Chrome one only, since the baselines record computed styles and rounded boxes and per-browser baselines would be a guess rather than a measurement.

VTT-AI-Agent (commented):
## TL;DR

Reviewing the new commit (`57939e79`, phases 4–7); rounds 1–3 already cover everything before it. The interaction and DOM layers are well built and `npm test` is 640/640 green here. One real defect: the `CORPUS_FLAGS_OFF` diagnostic in `corpusreplay.js` is **inert** — `Room.setState()` re-derives `gameSettings` from `FileUpdater()` whenever the incoming state has `_meta`, which every library file does, so the "flags off" replay runs with exactly the same modes as the first one, and the nightly workflow pays for that second replay on all 16 shards.

## Findings

### 1. `CORPUS_FLAGS_OFF` replays the game with the flags still on (`tests/testcafe/corpusreplay.js:71-76`, `:117-123`)

`replay()` does `setRoomState({})` → `applyLegacy(flags)` → `setRoomState(state)`. That order is correct in `interactions.js` and `domsnapshot.js`, because their fixture states are hand-built and carry no `_meta`, so `Room.setState` keeps the room's `gameSettings` (`server/room.mjs:1206-1215`). Corpus states are the opposite case:

```js
this.state = state;
if(this.state._meta) {
  this.state = FileUpdater(this.state);
  gameSettings = (this.state._meta || {}).gameSettings || { legacyModes: {} };   // ← overwrites applyLegacy
}
```

Every checked-in library variant has `_meta` (I spot-checked: v13–v17 across the first six games), so the `applyLegacy({})` in the second call is discarded a moment later and the game is loaded at the updater's own classification again. Consequences:

- `report.flagsOffDifference` compares the run with itself — it is structurally always `false`, so the "where is a flag load-bearing" signal the comment at `:120-122` promises is never produced.
- `.github/workflows/corpus-replay.yml:60` sets `CORPUS_FLAGS_OFF: '1'` unconditionally on all 16 shards, so the nightly job replays every variant twice for that non-signal — roughly double wall-clock for zero information.

Note that the first, asserted replay is unaffected: `flags` from `flagsForGame()` and the modes the room derives come from the same `FileUpdater`, so `report.legacyModes` is still accurate. Only the diagnostic is dead. Fix is to set the modes *after* the state is loaded (`setRoomState(state)` then `applyLegacy(flags)`, which is what the Game Settings panel does at runtime and which re-broadcasts the state) — stripping `_meta` instead would also skip the widget migrations, which is not what you want here.

### 2. `flagsOffErrors` includes the errors from the first replay (`corpusreplay.js:96`, `:119`)

`t.getBrowserConsoleMessages()` accumulates for the lifetime of the page, and `replay()` never navigates — it only re-PUTs the room state. So the second call returns *both* runs' messages, i.e. `flagsOffErrors ⊇ errors` by construction. Even once finding 1 is fixed, a report showing an error under "flags off" would not mean the error happened with the flags off. Worth capturing a baseline count (or a `t.navigateTo` between the runs) before diffing the two lists.

### 3. A missing DOM baseline silently records itself and passes (`tests/testcafe/domsnapshot.js:98-108`)

`if(process.env.WRITE_DOM_SNAPSHOTS || !fs.existsSync(file))` writes the baseline and then compares the snapshot against the file it just wrote — which always matches. That makes the failure mode of a typo in `goldenFile()`, a renamed fixture, or a baseline deleted by a rebase/merge a *green* CI run that certifies nothing, on the one test file in the repo whose entire value is the checked-in reference. `compareState()` in `test-util.js` is safe from this because it only writes the reference after the hash has already matched; here there is no such gate. Suggest failing with an explicit "no baseline for `<fixture>.<combo>`; re-record with `WRITE_DOM_SNAPSHOTS=1`" when the file is absent and the env var is not set.

(Related watch item, not a finding: the baselines were recorded under `chromium:headless` locally while CI runs `chrome:headless` — a different build. Geometry should be exact since everything is inside a `transform: scale()`, but the first CI run of `domsnapshot.js` is the real test of that assumption.)

### 4. `corpus.test.js:40` hardcodes the current save version

```js
_meta: Object.assign({}, state._meta, { version: 21, gameSettings: { legacyModes: assigned } })
```

The intent is "feed the updater its own output and prove it is a no-op", but the version is pinned to a literal while `VERSION` lives in `server/fileupdater.mjs`. The next version bump quietly turns this from an idempotence check into a v21→v22 migration check, i.e. it stops testing the thing it is named after without failing. `import { VERSION }` (or export it) and use that; it's the kind of drift the rest of this PR is explicitly built to prevent.

### 5. Dead code in `corpusreplay.js`

- `clickableButtons` (`:56-62`) is a `ClientFunction` with no caller — `replay()` derives the button list from the state via `buttonIDsWithRoutine()` instead.
- `replay(t, variant, state, flags)` never uses `variant` (`:71`).

### 6. `corpus-census.js:44-45` counts an on-disk mode that is explicitly `false`

The "on disk" column iterates `Object.keys(flagsOnDisk(state))` while the "at load" column only counts truthy values (`:54-56`). A file carrying `legacyModes: { useIframeForHtmlCards: false }` — which is exactly what the Game Settings panel writes when an author unticks a box and saves — would be counted as having the mode on disk and not at load, reading as "the updater removed a mode", which it cannot do. `Object.entries(...).filter(([ , value ])=>value)` makes the two columns comparable.

### 7. Minor: `interactions.js:108` burns the full backoff on every run

`stateWhen(_=>false)` is the deliberate "wait and prove nothing happened" idiom for the `movable: false` case, but it always costs the whole ~3.1s ladder, twice (once per tier). A fixed `t.wait()` with a comment would say the same thing more plainly and cheaply. Same shape at `:117` is fine — that one has a real predicate.

## What I checked and found clean

- `npm test` → 16 suites, 640 tests, all green on `57939e79`, matching the PR description.
- The mode-before-state ordering in `interactions.js:24-34` and `domsnapshot.js:74-85` is correct for `_meta`-less fixture states, and the comments explaining why the modes must precede the widget constructor are accurate.
- `captureSnapshot()` records nothing window-size dependent: the box is divided by `--scale` and everything under `#topSurface` is inside `transform: scale(var(--scale))`, so board units come out exact; the asset-URL normalisation at `:62` correctly strips the host.
- `expectEventually()` moving from `routines.js` into `test-util.js` is a straight lift with an added `message` argument — no behaviour change for the existing callers.
- `widgets` / `dropTargets` added as harness globals are read-only bundle-global wiring for jsdom; the `inheritFrom` fixture's use of `{ value: … }` to distinguish an inherited `null` from "SET wrote nothing" is the right shape for that probe.
- `corpus.test.js` asserting invariants (termination, determinism, idempotence, no unknown mode, no mode newer than the file) rather than counts is the right call for a corpus that changes with every library PR — it won't go red on a game submission.
- The nightly workflow is correctly non-gating, sharded and `workflow_dispatch`-able, and `selectedVariants()` sampling every 32nd variant without an explicit selection keeps `testcafe-headless-all` usable.

The `beta` label was already on this PR and I left it: nothing under `client/js/widgets/`, `client/js/compute.js` or `client/js/serverstate.js` changes here, the new commit is entirely test infrastructure plus two CI workflows, and all six findings above are confined to test/diagnostic code that cannot reach a beta player.

VTT-AI-Agent:
Thanks — the `CORPUS_FLAGS_OFF` finding was correct and it was the important one. Fixed in `071abe9b`, all six points addressed:

**1. `CORPUS_FLAGS_OFF` replayed with the flags still on** — confirmed exactly as described: `Room.setState()` (`server/room.mjs:1206-1215`) re-derives `gameSettings` from `FileUpdater()` for any state carrying `_meta`, so the `applyLegacy()` before the load was discarded. `replay()` now applies an override *after* the state, which is the path the Game Settings panel takes (`setLegacyMode` re-broadcasts the state, so the widgets are rebuilt with the modes in place). The asserted first replay passes `null` and simply keeps what the room derived — no redundant round of mode switches on a 500-widget game.

Verified on three library games with `CORPUS_FLAGS_OFF=1`:

| game | modes at load | `flagsOffDifference` before | after |
| --- | --- | --- | --- |
| games/2036 Daimyo | var modes | `false` | **`true`** |
| games/8 ballin_ | var modes | `false` | **`true`** |
| games/8 sprengt | none | (skipped) | (skipped) |

So the nightly second replay now produces the signal it was paying for. I left `CORPUS_FLAGS_OFF: '1'` on in the workflow for that reason — happy to gate it behind a `workflow_dispatch` input if the wall-clock still looks like a bad trade once there is a first real nightly run to measure.

**2. `flagsOffErrors` inherited the first run's errors** — right, `getBrowserConsoleMessages()` accumulates and `replay()` never navigates. It now takes an `errorsBefore` count and slices. Visible in the same run: `8 ballin_` reports `errors=150` / `flagsOffErrors=750` where it previously reported 750 / 750.

**3. A missing DOM baseline recorded itself and passed** — agreed, that made the one file whose value is the checked-in reference certify nothing. Writing is now `WRITE_DOM_SNAPSHOTS`-only and a missing file fails with `no DOM baseline for <fixture> in combination <combo> (<path>) - re-record with WRITE_DOM_SNAPSHOTS=1`. Verified by moving `widget-gallery.json` aside: red, and nothing re-recorded.

While re-running `domsnapshot.js` I hit one flake (a `widget-gallery` mismatch on the first run against a freshly started server, not reproducible in six later runs), so I also added `settledSnapshot()`: capture until two consecutive reads agree. That is the same class of problem as your `chromium` vs `chrome` watch item — a baseline nobody can reproduce — and it seemed worth closing before the first CI run rather than after.

**4. `version: 21` hardcoded** — fixed, `corpus.test.js` now imports `VERSION` from `server/fileupdater.mjs`.

**5. Dead code** — `clickableButtons` and the unused `variant` argument are gone.

**6. `corpus-census.js` counted a `false` mode as on disk** — fixed with `Object.entries(...)` + truthy filter, so both columns mean the same thing. The printed numbers happen to be unchanged (no library file currently carries an explicitly `false` mode), but the next one saved from the Game Settings panel would have skewed it.

**7. `stateWhen(_=>false)`** — replaced with a single `t.wait(500)` plus a comment saying why there is nothing to poll for.

Verification: `npm test` 640/640, `tests/testcafe/interactions.js` 32/32, `tests/testcafe/domsnapshot.js` 3/3 (baselines unchanged), and the corpus run above.

ArnoldSmith86:
Look at https://github.com/ArnoldSmith86/virtualtabletop/issues/1212 . It's a bit old and there might be even more properties involved now. Add quite a few more tests that cover these kinds of ordering and interaction. It probably doesn't adhere to the described order currently but changes to the status quo should surface and at some point we want to change it to the described order (or something silimar) with the help of a legacy mode.

Also implement everything you listed as "not done yet".

VTT-AI-Agent:
Pushed `4403108e` — the #1212 ordering tests and everything that was listed as not done.

## #1212

`tests/testcafe/holderevents.js`, 22 cases, all recording the status quo rather than the description. Every holder routine appends one entry to a log widget, and the entry carries the properties of the widget the event is about read at that moment — so it says both what fired and how much of the event had already happened when it fired.

| what the player does | what fires today | what #1212 asks for |
|---|---|---|
| drag out of a holder | `leave`(mark=null), `leave`(mark=leave-A) | `beforeLeaveRoutine`, then `leaveRoutine` |
| drag A → B | the two leaves, then `enter B` after parent, onEnter and alignment | + `afterLeaveRoutine` after the arrival |
| drag within a holder | `leave`, `enter` — and **onEnter is not applied**, only the routine runs | `beforeLeaveRoutine` + `leaveCancelledRoutine` |
| drag out and back | the two leaves, then `enter` — indistinguishable from arriving from elsewhere | `leaveCancelledRoutine` with a flag |
| `MOVE` A → B | the two leaves, then `enter` | onLeave+leave, arrival, afterLeave, enter |
| `MOVEXY` out | **one** leave, and **onLeave is never applied** | onLeave, leave, afterLeave |
| `SET parent` | one leave, which reads `parent=handB` — the *new* parent | — |
| `DELETE` in a holder | one leave, which still reads `parent=handA` | — |

The doubled `leaveRoutine` you know about has two distinct sources, and the trace tells them apart: `onPropertyChange('parent')` calls the old parent's `leaveRoutine` directly when the drag detaches the widget (`widget.js:2766`), *before* `onLeave` is applied, and `checkParent()` → `dispenseCard()` applies `onLeave` and calls it again once the widget stops overlapping. Which is exactly the `beforeLeaveRoutine` / `leaveRoutine` split the issue proposes — so when that change is made, this file shows per path what moved, and the legacy mode has a place to be asserted from both sides.

Three things I did not expect and which are pinned as they are:

- **Dragging a card within its holder runs `enterRoutine` but does not apply `onEnter`** (`onChildAdd` skips the property half when the widget lands back on the holder it was dragged off). So the properties and the routine disagree about whether an entry happened, which is currently the only way a game can tell that case apart at all.
- **`MOVEXY` never applies `onLeave`** where `MOVE` does — it writes `parent` directly, so `dispenseCard()` never runs.
- **A holder with `dropShadow` runs its enter *and* leave events for the shadow widget**, i.e. for something the player never dropped.

You said "there might be even more properties involved now" — the ones the file also covers are `childrenPerOwner`, `alignChildren`, `stackOffsetX`, `dropLimit`, `ignoreOnLeave`, `dropShadow`, piles (a pile dropped into a holder gets `onEnter` on every card inside it and one `enterRoutine` for the pile), `RECALL`, and the `child` collection / `oldParentID` the routines are handed. The whole holder-to-holder and `MOVE` order is asserted again with every legacy mode on, since `disableHolderImageWidget` swaps the holder's prototype and this is the only place a leak into the event methods would show.

## The "not done yet" list

- **Canvas drawing** — `canvas.js`, 10 cases: the raw pointer stream, asserted on both sides of the round trip (the pixel the 2D context shows, and the run-length encoded region string in the state), plus `artist`, `clickable`, `lineWidth`, the resolution clamp and the three `CANVAS` modes.
- **Multi-client** — `multiclient.js`, 5 cases, two windows in one room: delta arrival, convergence, `onlyVisibleForSeat` and `owner` from both sides, and an `INPUT` overlay that opens on the asking client only.
- **Keyboard / zoom** — `keyboardzoom.js`, 12 cases: hotkey resolution and the three situations the key is ignored in, then a drag and a drop into a holder at **2x**, where the coordinate transform can be wrong.
- **`inheritFrom` object form** — `inheritfrom.js`, 9 cases. Deleting the source found that the two halves of a widget disagree: the DOM keeps rendering the inherited value while `get()` already returns the widget's own default (#2854).
- **Pixel diffing** — `pixelsnapshot.js`. Not a checked-in golden PNG — that is the version that gets switched off after a month of runner drift. It compares two screenshots **from the same run** of the same board in the two flag combinations: the widget gallery must be pixel identical (it is), and the two rendering fixtures must differ by more than an anti-aliased edge (46187 and 710 pixels today). That is the first assertion in the repo that the two rendering modes change anything a player can *see*. PNG decoding is ~60 lines against node's `zlib`, no new dependency.
- **Firefox DOM baselines** — the one I could not finish. The baselines are now keyed by browser (`widget-gallery.chrome.json`) and a missing one fails naming the browser it has to be recorded on, but Firefox does not start in this sandbox (no GTK), so recording the file itself needs a machine that can run it — a hand-written baseline would be a guess. `pixelsnapshot.js` needs no baseline, so that half of Layer F is in the Firefox matrix now, along with `inheritfrom.js`.

`npm test` 640/640. Locally: `holderevents.js` 22/22 (twice, for stability), `canvas.js` 10/10, `keyboardzoom.js` 12/12, `inheritfrom.js` 9/9, `multiclient.js` 5/5, `pixelsnapshot.js` 4/4, `domsnapshot.js` 3/3, `interactions.js` 32/32 after moving its shared driving code into `interaction-util.js`.

ArnoldSmith86:
update this to main

VTT-AI-Agent:
Updated to main (merged `1e38ce21`).

Two conflicts, both in files main and this branch had both touched:

- `client/js/editor/sidebar/gameSettings.js` — main moved the tile frame out of the inline styles into a `.settings-tile` CSS rule and added the Board Settings section; this branch had rewritten the legacy-mode tiles into `addLegacyModeSection()`. Resolved to main's CSS rule plus the accent edge this branch colors per mode, and to `addBoardSizeSetting()` followed by `addLegacyModeSection()` (which renders its own "no legacy modes" case, so main's `gameSettings.legacyModes` guard is gone).
- `tests/testcafe/test-util.js` — both sides added helpers at the same spot, both kept.

Two things the merge broke, fixed in `c9f3a203`:

- **`boardsize.js` (new on main) failed 2/7.** The per-test reset here switched all four legacy modes off one `setLegacyMode` call at a time, i.e. five state messages in a row before every test - and the client did not get as far as the board size that came after them. The reset now clears the game settings through the state the room is set to (`_meta.gameSettings`), which is one message and the same result.
- **`multiclient.js` failed 4/5.** Naming a player through the reworked Players table (#3013) is a rename round trip per keystroke, and with two clients in the room both of them work through every meta update that follows - the second client missed the delta the file exists to assert on. The name a client joins under comes from `localStorage`, so it is set there before the room is loaded. As a side effect the file runs in 1m30 instead of 3m30.

Verified locally after the merge: `npm test` 1144/1144, and TestCafe `boardsize.js` 7/7, `toolbar.js` 4/4, `editor.js` 32/32, `multiclient.js` 5/5, `holderevents.js` 22/22, `interactions.js` 32/32, `keyboardzoom.js` 12/12, `canvas.js` 10/10, `inheritfrom.js` 9/9, `domsnapshot.js` 3/3 (three times, the DOM baselines still hold across main's rendering changes), `pixelsnapshot.js` 4/4, `legacymatrix.js` 7/7, `routines.js` 4/4, `compute-1.js`, and a three-game `corpusreplay.js` smoke run.

ArnoldSmith86 (approved):

PR #3064, opened 2026-07-28
Rebuild the All Aboard map with line widgets

VTT-AI-Agent:
## What this does

Every one of the 100 routes on the map of **All Aboard** (the Ticket to Ride style game in the public library) is now a **line widget** connected to the two city cards it runs between, with that route's train-car holders as its `stops`.

Before, the ~309 holders were placed by hand: each one had its own eyeballed `x`, `y` and `rotation`, so the spacing along a route was uneven, parallel double routes drifted apart, and moving a city meant nudging every holder around it by hand. Now the line owns the geometry — the holders are positioned and rotated by the path — and moving a city in edit mode drags every route to and from it along with it.

map

## The map still looks the same

Rebuilt from the existing map rather than redrawn, so every route keeps its course, its colour, its length and its holder ids (the scoring routines select holders by their `route` property and are untouched):

* **City dots stay exactly where they are.** The dot is now centred in the city card's height (`faceTemplates` image `y: 0` → `4`, every city card `y` − 4) so that a connected line end point — which lands in the middle of the card — meets the middle of the dot. The city name sits 4px higher relative to the dot, which lines it up with the dot a little better than before.
* **Curved routes stay curved.** Where the hand-placed chain bent away from the straight connection (the Gulf coast, the Pacific coast, Miami…), the line is a cubic Bézier least-squares fitted to that chain, so it follows the same course.
* **Double routes are exactly parallel.** The two lines of a pair share one fitted centre path and get matching `-9.5` / `+9.5` connection offsets, which `applyConnections` turns into a parallel pair over the whole route.
* **Spacing rule:** the two city dots count as stops of the line as well, and everything in between is spaced evenly: the gap between two cars is the same as the gap between a car and the dot it faces. That keeps every route reaching out towards both of its cities (clearance 10.8–27.5px, against 9.1–38px on the hand-placed map) and gives the cars a visible seam (median gap 4.5px) so they can be counted.

The lines themselves are invisible (`lineWidth: 0`) and sit behind the holders (`layer: -10`), so the tracks stay the only thing you see and clicks/drops still land on the holders. Shared settings live in a `Route Template` line widget, the same way the holders already share a `Holder Template`.

## Engine fix that goes with it

`Line.positionAttachedWidgets()` now flushes the delta when the line has stops that are **not** its children (`hasExternalStops()`) — as the route holders are, since they stay children of the board.

Such a stop is positioned through global coordinates, which read the line's CSS transform. Inside a batch that transform still shows the state of the *previous* event, so after `normalizeGeometry()` moved the widget box — and it is that box move which triggers the layout — every stop landed off the path by exactly that move. Moving a city by 10px and back left its tracks 10px beside the line. `applyConnections()` already flushes for exactly this reason; the comment there describes the same trap.

## Testing

* `npm test` — 86 passed
* `npx testcafe chromium:headless tests/testcafe/editor.js` — 14 passed, including *Line widget in edit mode*
* `node validator/validate_gamefile_node.js "library/games/All Aboard/0.json"` — no new findings (the three pre-existing deck warnings are unchanged)
* Loaded the game in a browser and compared it side by side with the old map ([before](https://agent.virtualtabletop.io/reports/pr/1531626854740394146/all-aboard-before.png) / [after](https://agent.virtualtabletop.io/reports/pr/1531626854740394146/all-aboard-after.png)); moved a city and watched all six routes around it re-space and re-rotate; verified a nudge-and-back leaves every holder exactly where the file has it.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3064/pr-test (or any other room on that server)

### DIFF
--- a/client/js/widgets/line.js
+++ b/client/js/widgets/line.js
@@ -298,40 +298,37 @@ export class Line extends Widget {
   // A stop outside the line is placed through global coordinates, which read the
   // CSS transforms of the frames it converts between. Inside a batch those show
   // the state of the previous event, so after normalizeGeometry moved the box -
-  // which is what positionAttachedWidgets runs on - every stop would land off
-  // the path by exactly that move. Same reason applyConnections flushes. Only a
-  // transform change can make them stale, so a re-space - which rewrites nothing
-  // but the stops and their own x/y - costs one flush instead of one per pass.
-  flushStaleTransforms() {
-    if(!this.hasExternalStops())
-      return;
-    const transforms = this.stopFrameTransforms();
-    if(transforms == this.flushedTransforms)
-      return;
-    flushDelta();
-    this.flushedTransforms = transforms;
+  // which is what positionAttachedWidgets runs on - the stop would land off the
+  // path by exactly that move. Same reason applyConnections flushes. Asking the
+  // DOM whether it still matches the state - rather than remembering what was
+  // flushed before - also covers a frame that was flushed by something else in
+  // the batch and changed again since, and still costs no flush at all for a
+  // re-space, which rewrites nothing but the stops and their own x/y.
+  flushStaleTransforms(stop) {
+    if(this.stopFrames(stop).some(frame=>frame.domElement.style.transform != frame.cssTransform()))
+      flushDelta();
   }
 
   // stopCoordInParentFrame converts through two chains of DOM transforms: this
-  // line up to the room, and the same for the parent each external stop is
-  // placed in. Either going stale inside a batch displaces the stop, so both
-  // sides go into the snapshot flushStaleTransforms compares against.
-  stopFrameTransforms() {
-    const frames = [ this, ...this.attachedWidgets().map(stop=>widgets.get(stop.get('parent'))) ];
-    const seen = new Set();
-    const chain = [];
-    for(const frame of frames)
-      for(let w = frame; w && !seen.has(w.id); w = widgets.get(w.get('parent'))) {
-        seen.add(w.id);
-        chain.push(w.id, ...w.cssTransformProperties().map(property=>w.get(property)));
-      }
-    return chain.join();
+  // line up to the room, and the same for the parent the stop is placed in.
+  // Either going stale inside a batch displaces the stop. A stop that is a child
+  // of the line is never converted, so it reads no transform at all.
+  stopFrames(stop) {
+    if(stop.get('parent') == this.id)
+      return [];
+    const frames = [];
+    for(const start of [ this, widgets.get(stop.get('parent')) ])
+      for(let w = start; w && !frames.includes(w); w = widgets.get(w.get('parent')))
+        frames.push(w);
+    return frames;
   }
 
   async positionAttachedWidgets() {
-    this.flushStaleTransforms();
     for(const entry of this.stopList()) {
       const stop = widgets.get(entry.widget);
+      // per stop, because the x/y this loop writes are themselves unflushed and
+      // one stop can be (inside) the frame another stop is placed in
+      this.flushStaleTransforms(stop);
       const p = this.stopCoordInParentFrame(stop, this.pointAtPosition(entry.position));
       await stop.set('x', Math.round(p.x - stop.get('width')/2));
       await stop.set('y', Math.round(p.y - stop.get('height')/2));
--- a/client/js/widgets/line.js
+++ b/client/js/widgets/line.js
@@ -295,14 +295,28 @@ export class Line extends Widget {
     }
   }
 
+  // A stop outside the line is placed through global coordinates, which read the
+  // CSS transforms of this line and its parents. Inside a batch those still show
+  // the state of the previous event, so after normalizeGeometry moved the box -
+  // which is what positionAttachedWidgets runs on - every stop would land off
+  // the path by exactly that move. Same reason applyConnections flushes. Only a
+  // transform change can make them stale, so a re-space - which rewrites nothing
+  // but the stops and their own x/y - costs one flush instead of one per pass.
+  flushStaleTransforms() {
+    if(!this.hasExternalStops())
+      return;
+    const chain = [];
+    for(let w = this; w; w = widgets.get(w.get('parent')))
+      chain.push(w.id, ...w.cssTransformProperties().map(property=>w.get(property)));
+    const transforms = chain.join();
+    if(transforms == this.flushedTransforms)
+      return;
+    flushDelta();
+    this.flushedTransforms = transforms;
+  }
+
   async positionAttachedWidgets() {
-    // A stop outside the line is placed through global coordinates, which read
-    // this line's CSS transform. Inside a batch that still shows the state of
-    // the previous event, so after normalizeGeometry moved the box - which is
-    // what the layout below runs on - every stop would land off the path by
-    // exactly that move. Same reason applyConnections flushes.
-    if(this.hasExternalStops())
-      flushDelta();
+    this.flushStaleTransforms();
     for(const entry of this.stopList()) {
       const stop = widgets.get(entry.widget);
       const p = this.stopCoordInParentFrame(stop, this.pointAtPosition(entry.position));
--- a/client/js/widgets/line.js
+++ b/client/js/widgets/line.js
@@ -296,7 +296,7 @@ export class Line extends Widget {
   }
 
   // A stop outside the line is placed through global coordinates, which read the
-  // CSS transforms of this line and its parents. Inside a batch those still show
+  // CSS transforms of the frames it converts between. Inside a batch those show
   // the state of the previous event, so after normalizeGeometry moved the box -
   // which is what positionAttachedWidgets runs on - every stop would land off
   // the path by exactly that move. Same reason applyConnections flushes. Only a
@@ -305,14 +305,27 @@ export class Line extends Widget {
   flushStaleTransforms() {
     if(!this.hasExternalStops())
       return;
-    const chain = [];
-    for(let w = this; w; w = widgets.get(w.get('parent')))
-      chain.push(w.id, ...w.cssTransformProperties().map(property=>w.get(property)));
-    const transforms = chain.join();
+    const transforms = this.stopFrameTransforms();
     if(transforms == this.flushedTransforms)
       return;
     flushDelta();
     this.flushedTransforms = transforms;
+  }
+
+  // stopCoordInParentFrame converts through two chains of DOM transforms: this
+  // line up to the room, and the same for the parent each external stop is
+  // placed in. Either going stale inside a batch displaces the stop, so both
+  // sides go into the snapshot flushStaleTransforms compares against.
+  stopFrameTransforms() {
+    const frames = [ this, ...this.attachedWidgets().map(stop=>widgets.get(stop.get('parent'))) ];
+    const seen = new Set();
+    const chain = [];
+    for(const frame of frames)
+      for(let w = frame; w && !seen.has(w.id); w = widgets.get(w.get('parent'))) {
+        seen.add(w.id);
+        chain.push(w.id, ...w.cssTransformProperties().map(property=>w.get(property)));
+      }
+    return chain.join();
   }
 
   async positionAttachedWidgets() {
--- a/client/js/widgets/line.js
+++ b/client/js/widgets/line.js
@@ -296,6 +296,13 @@ export class Line extends Widget {
   }
 
   async positionAttachedWidgets() {
+    // A stop outside the line is placed through global coordinates, which read
+    // this line's CSS transform. Inside a batch that still shows the state of
+    // the previous event, so after normalizeGeometry moved the box - which is
+    // what the layout below runs on - every stop would land off the path by
+    // exactly that move. Same reason applyConnections flushes.
+    if(this.hasExternalStops())
+      flushDelta();
     for(const entry of this.stopList()) {
       const stop = widgets.get(entry.widget);
       const p = this.stopCoordInParentFrame(stop, this.pointAtPosition(entry.position));
--- a/library/games/All Aboard/0.json
+++ b/library/games/All Aboard/0.json
@@ -1922,8 +1922,8 @@
   "h3": {
     "type": "holder",
     "id": "h3",
-    "x": 781,
-    "y": 90,
+    "x": 782,
+    "y": 91,
     "inheritFrom": "Holder Template",
     "route": 3,
     "rotation": -158,
@@ -1932,11 +1932,11 @@
   "h4": {
     "type": "holder",
     "id": "h4",
-    "x": 1001,
+    "x": 1004,
     "y": 165,
     "inheritFrom": "Holder Template",
     "route": 4,
-    "rotation": 14.46,
+    "rotation": 15.84,
     "parent": "Board"
   },
   "h5": {
@@ -1953,11 +1953,11 @@
   "h6": {
     "type": "holder",
     "id": "h6",
-    "x": 1234,
-    "y": 140,
+    "x": 1235,
+    "y": 141,
     "inheritFrom": "Holder Template",
     "route": 6,
-    "rotation": -130.39,
+    "rotation": -130.09,
     "parent": "Board"
   },
   "h7": {
@@ -1967,7 +1967,7 @@
     "y": 120,
     "inheritFrom": "Holder Template",
     "route": 7,
-    "rotation": -139.64,
+    "rotation": -139.82,
     "parent": "Board"
   },
   "h8": {
@@ -2005,30 +2005,30 @@
   "h11": {
     "type": "holder",
     "id": "h11",
-    "x": 1132,
+    "x": 1133,
     "y": 271,
     "inheritFrom": "Holder Template",
     "route": 11,
     "color": "#9EEFE5",
-    "rotation": 160.1,
+    "rotation": 159.78,
     "parent": "Board"
   },
   "h12": {
     "type": "holder",
     "id": "h12",
-    "x": 1092,
+    "x": 1090,
     "y": 304,
     "inheritFrom": "Holder Template",
     "route": 12,
     "color": "#058C42",
-    "rotation": 164.33,
+    "rotation": 164.43,
     "parent": "Board"
   },
   "h13": {
     "type": "holder",
     "id": "h13",
     "x": 1048,
-    "y": 268,
+    "y": 269,
     "inheritFrom": "Holder Template",
     "route": 13,
     "rotation": -93.75,
@@ -2059,29 +2059,29 @@
   "h16": {
     "type": "holder",
     "id": "h16",
-    "x": 865,
-    "y": 154,
+    "x": 871,
+    "y": 152,
     "inheritFrom": "Holder Template",
     "route": 16,
-    "rotation": -14.95,
+    "rotation": -14.61,
     "parent": "Board"
   },
   "h17": {
     "type": "holder",
     "id": "h17",
-    "x": 553,
-    "y": 48,
+    "x": 567,
+    "y": 40,
     "inheritFrom": "Holder Template",
     "route": 17,
     "color": "#648DE5",
-    "rotation": -30.78,
+    "rotation": -28.8,
     "parent": "Board"
   },
   "h18": {
     "type": "holder",
     "id": "h18",
-    "x": 659,
-    "y": 96,
+    "x": 658,
+    "y": 94,
     "inheritFrom": "Holder Template",
     "route": 18,
     "color": "#A300A0",
@@ -2155,7 +2155,7 @@
   "h25": {
     "type": "holder",
     "id": "h25",
-    "x": 132,
+    "x": 131,
     "y": 42,
     "inheritFrom": "Holder Template",
     "route": 25,
@@ -2178,18 +2178,18 @@
   "h27": {
     "type": "holder",
     "id": "h27",
-    "x": 423,
-    "y": 339,
+    "x": 424,
+    "y": 342,
     "inheritFrom": "Holder Template",
     "route": 27,
     "color": "#058C42",
-    "rotation": -105.82,
+    "rotation": -105.92,
     "parent": "Board"
   },
   "h28": {
     "type": "holder",
     "id": "h28",
-    "x": 584,
+    "x": 587,
     "y": 189,
     "inheritFrom": "Holder Template",
     "route": 28,
@@ -2247,8 +2247,8 @@
   "h33": {
     "type": "holder",
     "id": "h33",
-    "x": 397,
-    "y": 352,
+    "x": 401,
+    "y": 353,
     "inheritFrom": "Holder Template",
     "route": 33,
     "color": "#A30000",
@@ -2259,7 +2259,7 @@
   "h34": {
     "type": "holder",
     "id": "h34",
-    "x": 878,
+    "x": 879,
     "y": 480,
     "inheritFrom": "Holder Template",
     "route": 34,
@@ -2270,32 +2270,32 @@
   "h35": {
     "type": "holder",
     "id": "h35",
-    "x": 801,
-    "y": 638,
+    "x": 802,
+    "y": 639,
     "inheritFrom": "Holder Template",
     "route": 35,
     "color": "#058C42",
-    "rotation": 71.7,
+    "rotation": 71.77,
     "parent": "Board"
   },
   "h36": {
     "type": "holder",
     "id": "h36",
     "x": 638,
-    "y": 525,
+    "y": 524,
     "inheritFrom": "Holder Template",
     "route": 36,
-    "rotation": -100.24,
+    "rotation": -100.12,
     "parent": "Board"
   },
   "h37": {
     "type": "holder",
     "id": "h37",
     "x": 619,
-    "y": 530,
+    "y": 529,
     "inheritFrom": "Holder Template",
     "route": 37,
-    "rotation": -100.4,
+    "rotation": -100.19,
     "parent": "Board"
   },
   "h38": {
@@ -2321,8 +2321,8 @@
   "h40": {
     "type": "holder",
     "id": "h40",
-    "x": 669,
-    "y": 418,
+    "x": 670,
+    "y": 417,
     "inheritFrom": "Holder Template",
     "route": 40,
     "rotation": 115.74,
@@ -2331,8 +2331,8 @@
   "h41": {
     "type": "holder",
     "id": "h41",
-    "x": 652,
-    "y": 409,
+    "x": 653,
+    "y": 408,
     "inheritFrom": "Holder Template",
     "route": 41,
     "rotation": 115.74,
@@ -2341,45 +2341,45 @@
   "h42": {
     "type": "holder",
     "id": "h42",
-    "x": 627,
+    "x": 636,
     "y": 389,
     "inheritFrom": "Holder Template",
     "route": 42,
     "color": "#FF7700",
     "parent": "Board",
-    "rotation": -1.37
+    "rotation": -1.32
   },
   "h43": {
     "type": "holder",
     "id": "h43",
-    "x": 580,
-    "y": 372,
+    "x": 583,
+    "y": 371,
     "inheritFrom": "Holder Template",
     "route": 43,
     "color": "#A300A0",
     "parent": "Board",
-    "rotation": -1.03
+    "rotation": -1.1
   },
   "h44": {
     "type": "holder",
     "id": "h44",
-    "x": 567,
+    "x": 569,
     "y": 324,
     "inheritFrom": "Holder Template",
     "route": 44,
     "color": "#EF233C",
-    "rotation": -11.54,
+    "rotation": -11.26,
     "parent": "Board"
   },
   "h45": {
     "type": "holder",
     "id": "h45",
-    "x": 1099,
-    "y": 735,
+    "x": 1101,
+    "y": 742,
     "inheritFrom": "Holder Template",
     "route": 45,
     "color": "#EF233C",
-    "rotation": 77.34,
+    "rotation": 81.1,
     "parent": "Board"
   },
   "h46": {
@@ -2396,22 +2396,22 @@
   "h47": {
     "type": "holder",
     "id": "h47",
-    "x": 1066,
-    "y": 748,
+    "x": 1071,
+    "y": 753,
     "inheritFrom": "Holder Template",
     "route": 47,
     "color": "#A30000",
-    "rotation": -134.98,
+    "rotation": -134.69,
     "parent": "Board"
   },
   "h48": {
     "type": "holder",
     "id": "h48",
-    "x": 776,
+    "x": 777,
     "y": 681,
     "inheritFrom": "Holder Template",
     "route": 48,
-    "rotation": -5.95,
+    "rotation": -6.2,
     "parent": "Board"
   },
   "h49": {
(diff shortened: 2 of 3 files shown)


VTT-AI-Agent (commented):
## TL;DR

Solid, carefully-built change — I re-derived the map data independently and it checks out (all 309 route holders are a stop of exactly one line, nothing removed, geometry stored in the file is *bit-for-bit* what the engine recomputes), and I reproduced the engine fix working. The one thing I'd push back on: the `line.js` flush fix is the only engine change in the PR and it has **no regression test**, and the flush is done once per `positionAttachedWidgets()` call, which is 5× per re-space.

## What I verified

Loaded the state on a local server and drove it with Playwright:

* Calling `layoutStops()` on all 101 route lines changes **0** of the 309 holders' `x`/`y`/`rotation` — so the committed geometry is exactly what the engine computes, and simply loading or touching the game will not shift the map.
* Moving Atlanta 10px and back leaves **0** holders displaced — the fix in `client/js/widgets/line.js:299` does what the description says.
* Static checks on the file: no widget removed (777 → 878, +101 lines only); only `x`/`y`/`rotation` differ on holders (median move 2.24px, max 21px); all 309 route holders appear in exactly one `stops` list, no duplicates, no stop that isn't a holder; every route has both `connectStart` and `connectEnd`; all stop positions are monotonic and within `[0,1]`; 56 single + 22 mirrored `±9.5` pairs = 100 routes; validator reports nothing new.

## Findings

### 1. The engine fix has no test — `client/js/widgets/line.js:299`

`positionAttachedWidgets()` is the only engine change here, and nothing in `tests/testcafe/editor.js` exercises a line whose stops are *not* its children being moved. The bug it fixes (a stop drifting by exactly the box move on every event) is silent and cumulative, so it can come back unnoticed the next time the layout path is refactored. A state-hash test in the shape of the existing *Line widget in edit mode* case — create a line, make a board-level widget a stop, nudge the line and nudge it back, compare hash — would pin it down cheaply. Worth noting All Aboard itself is not in `tests/testcafe/publiclibrary-*.js`, so the 5.7k-line map rebuild has no automated coverage either; if it were added, this whole class of drift would be caught by the hash.

### 2. The flush is finer-grained than it needs to be — `client/js/widgets/line.js:299-305`

`flushDelta()` sends a delta to the server whenever `batchDepth` is unwound, and `positionAttachedWidgets()` is called five times inside `distributeAttachedWidgetsEvenlyInternal()` (the 3-iteration loop plus the two trailing calls) and again from every `layoutStops()`. Only the *first* of those needs a flush: the subsequent calls change `stops` and the stops' own `x`/`y`, neither of which touches this line's CSS transform. So a single re-space of an external-stop line now costs 5 round trips instead of 1, and during an interactive city drag each mousemove flushes once per connected line *on top of* the flush `applyConnections()` already does (6 routes around a city ⇒ ~12 flushes per mousemove). Hoisting the flush to the callers that actually move the box (`layoutStops` / `distributeAttachedWidgetsEvenly`), or tracking whether the geometry changed since the last flush, would keep the fix and drop the traffic. Not a correctness problem — undo grouping is safe because `flushDelta()` preserves `delta.c` — but it is the sort of thing that gets noticed on a slow connection.

### 3. Editing the map now writes `lineOriginalRotation` onto every route holder

Measured: after moving a single city one pixel in the browser, 20 holders had `lineOriginalRotation` written into their state; a pass over all routes gives all 309. A fresh load is clean (0), so this only bites in edit mode — but it is persisted widget state, so the first person who nudges a city and saves the variant back adds 309 junk properties to the library file. This is pre-existing behaviour of `positionAttachedWidgets()`, not something this PR introduces, but this PR is what makes it a 309-widget event; a follow-up that only records the original when the stop actually has an explicit `rotation` to preserve would be worth it.

### 4. Invisible lines still claim a 24px-wide hit path

`updateLinePath()` sets `stroke-width: Math.max(this.get('lineWidth'), 24)` on `.lineHitPath`, and `client/css/widgets/line.css:28` gives it `pointer-events: stroke` — `lineWidth: 0` does not opt out. Sampling a grid over the rendered map, ~7% of the visible surface (2083 of ~30k sample points) now returns a line hit path rather than `#w_Board`. Today that is harmless: `Board` is `movable:false` with no `clickRoutine`, and `Route Template`'s `dropTarget: []` makes `compareDropTarget` return false so nothing can be dropped onto a route. But it means the map has 100 invisible click-eating ribbons over it, and 101 extra entries in the `dropTargets` map that get evaluated on every drag frame. Nothing to change in this PR — flagging it because "an invisible line should probably not have a hit path" looks like a genuine engine question the line widget will have to answer eventually.

I did not add the `beta` label: `client/js/widgets/line.js` is genuinely engine code here — `positionAttachedWidgets()` changes runtime behaviour for any existing line with external stops, with no new opt-in flag gating it — so the PR is excluded from beta auto-deployment by design.

VTT-AI-Agent:
Thanks — both actionable points are addressed in 14d5278.

**1. No regression test for the engine fix** — valid, added `Line stops that are not children of the line` to `tests/testcafe/editor.js`. It builds the All Aboard shape in miniature (two cities, a route connected to both, two cars listed as stops but living at room level, not as children of the line) and moves a city inside one explicit batch, which is exactly what one mouse event of a drag is. The assertion is not a state hash but the invariant itself: for every stop, the stored `x`/`y` must equal what `stopCoordInParentFrame(stop, pointAtPosition(position))` computes right now — i.e. the stop is *on* the path. I verified it both ways: it passes on the branch and fails with `expected 90 to deeply equal 0` (exactly the size of the move) with the flush commented out. Deliberately not a `compareState` hash, because the drift is silent and a hash would only tell you *something* changed.

**2. The flush is finer-grained than it needs to be** — valid, fixed. The flush now lives in `flushStaleTransforms()`, which compares the `cssTransformProperties()` of the line *and its parent chain* against what was current at the last flush and returns without flushing when nothing moved. I kept it inside `positionAttachedWidgets()` rather than hoisting it to `layoutStops()`/`distributeAttachedWidgetsEvenly()`: hoisting also drops the safety net for the case where a stop is itself a connect target, where re-positioning it mid-loop can move the box again. Walking the parent chain also covers a case the old unconditional flush covered by accident and a hoist would have lost — an ancestor moving in the same batch.

Measured on the All Aboard map (101 lines, 309 external stops), instrumenting `positionAttachedWidgets`/`flushStaleTransforms` in the browser:

| | calls with external stops | actual flushes |
|---|---|---|
| four city moves (one batch each) | 44 | 14 |
| one auto-space re-space of a route | 4 | 0 |

So the per-pass flushes inside `distributeAttachedWidgetsEvenlyInternal` are gone entirely, and a city move costs ~3.5 flushes per event instead of ~11.

**3. `lineOriginalRotation` written onto every route holder** — agreed it's real, and as you say it is pre-existing `positionAttachedWidgets()` behaviour that this PR only makes more visible. Leaving it out on purpose: the property is what lets `restoreStopRotation()` tell "had no explicit rotation, so restore to `null`" from "had one, restore that value", so shrinking it is a semantics change to the rotate/restore pair rather than a one-liner, and it wants its own PR (and its own test) instead of riding along with a map rebuild.

**4. Invisible lines still claim a 24px hit path** — agreed, and agreed it is a question for the line widget rather than for this PR. `lineWidth: 0` opting out of `.lineHitPath` (or capping the hit width when the line is invisible) would also drop those 101 entries from the drag-frame `dropTargets` walk. Not touching it here.

**beta label** — no argument, `client/js/widgets/line.js` is engine code and the change is ungated. Left off.

Re-ran `npm test` (86 passed) and the full `tests/testcafe/editor.js` suite (15 passed, including the existing `Line widget in edit mode` hash).

VTT-AI-Agent (commented):
## TL;DR

UI/UX review only (layout, visual clarity, first-time readability — not code correctness). The rebuilt map is a clear visual win: I loaded both versions side by side and measured them, and the new geometry is measurably tidier (car pitch now 43.3–47.2px instead of 39.3–65.2px, city clearance symmetric within ~2px per route instead of up to 18px off, double routes exactly aligned). The one thing I'd change: the *"leftover is clearance"* rule leaves up to ~40px of empty track at **both** ends of short-but-long routes, so a few routes now read as a bar floating between two cities instead of a track connecting them — spend that surplus on the gaps between cars instead.

Reviewed by running the game locally (public library → All Aboard, variant 1) on this branch and on `main`, and comparing the same crops pixel by pixel. Full map on this branch: [allaboard-after-full.png](https://agent.virtualtabletop.io/reports/ui-review/1531652648342917191/allaboard-after-full.png)

---

## Big layout improvements

### 1. Cap the end clearance; put the surplus into the car gaps

With a fixed 2.5px pitch and "the rest split evenly as clearance", a route whose car count is small relative to the distance between its cities ends up as a compact block in the middle of a large void. Measured clearance (city dot edge → nearest car edge, rendered px) on this branch:

| route | cars | clearance, both ends (this branch) | on `main` |
|---|---|---|---|
| Oklahoma City – Santa Fe | 3 | **38.8 / 39.7** | 18.9 / 32.1 |
| Dallas – El Paso | 4 | **39.3 / 39.4** | 21.2 / 29.7 |
| Helena – Omaha | 5 | **38.6 / 37.5** | 33.6 / 18.0 |
| Helena – Winnipeg | 4 | **37.2 / 36.4** | 24.3 / 29.8 |
| Duluth – Helena | 6 | **34.6 / 35.7** | 24.8 / 20.3 |
| … while at the other end of the map | | | |
| Charleston – Raleigh | 2 | **8.6 / 9.0** | 11.1 / 11.5 |
| Montreal – Toronto | 3 | **9.5 / 9.9** | 11.5 / 10.8 |
| Denver – Phoenix | 5 | **9.6 / 9.5** | 11.1 / 10.4 |

A train car is 45px long, so 39px is very nearly a whole missing car at *each* end. Oklahoma City–Santa Fe is the clearest case — the three blue cars now sit as an isolated blob mid-gap ([before](https://agent.virtualtabletop.io/reports/ui-review/1531652648342917191/santafe-okc-before.png) / [after](https://agent.virtualtabletop.io/reports/ui-review/1531652648342917191/santafe-okc-after.png)); on `main` the chain at least reached out to within 19px of Oklahoma City. The perfect symmetry is a genuine improvement over the old lopsided spacing, but the *magnitude* is the problem, and it now varies 8.6 → 39.7 across the map, which is the one place where the new geometry looks less consistent than the hand-placed one.

Concrete suggestion: clamp the per-end clearance (something like `min(leftover/2, ~20)`) and derive the pitch from what's left — `pitch = max(2.5, (routeLength - 2*clearance - n*carLength) / (n-1))`. Every route then still starts and ends a uniform, recognisable distance from its city dot, the long-gap routes stretch back out to touch their endpoints, and the extra pitch on exactly those routes makes them easier to count (see readability below). Nothing changes for the ~80% of routes where the leftover is already small.

### 2. City names are still overrun by the first car — this PR is the cheap moment to fix it

Every city label sits to the right of its dot, so any route leaving eastwards runs its first car straight through the name: **Denver**, **Kansas City**, **Santa Fe**, **El Paso**, **Little Rock**, **Vancouver**, **New Orleans**, **Saint Louis** ([Denver double route](https://agent.virtualtabletop.io/reports/ui-review/1531652648342917191/denver-kc-after.png)). This is pre-existing, but until now it was baked into 309 hand-placed `x`/`y` values and effectively unfixable; from this PR on it is one number in one place. Options that fit the existing style: give the eastbound end of a route a slightly larger clearance, or (cleaner, since the `faceTemplates` are already being touched here) drop the name below the dot for the handful of cities whose eastbound routes are crowded. Worth at least a follow-up issue with the geometry now centralised.

* **Car overlap is gone, which is the best small win here.** On `main` the minimum centre-to-centre distance along a route was 39.3px against a 45px car — i.e. adjacent cars physically overlapped. On this branch the minimum is 43.3px. Chicago–Pittsburgh is the visible case: the purple lower row of the double route was a single fused magenta bar and is now three countable cars ([before](https://agent.virtualtabletop.io/reports/ui-review/1531652648342917191/chicago-before.png) / [after](https://agent.virtualtabletop.io/reports/ui-review/1531652648342917191/chicago-after.png)).
* **Double routes now line up cell-for-cell** rather than being offset by half a car (Denver–Kansas City: [before](https://agent.virtualtabletop.io/reports/ui-review/1531652648342917191/denver-kc-before.png) / [after](https://agent.virtualtabletop.io/reports/ui-review/1531652648342917191/denver-kc-after.png)). Nothing to change, just confirming the `±9.5` offsets land where the description says.
* **The dot/label re-centring (`faceTemplates` image `y: 0` → `4`) checks out**: measured on the rendered card, the name's vertical centre is now ~1.5px above the dot centre (it was ~2.5px below). The dots themselves did not move — I diffed the crops.
* **Edit mode: the invisible lines have a 24px hit path** (`line.js:848`, `Math.max(lineWidth, 24)`), so with `lineWidth: 0` a click on an apparently empty stretch of map between two cars now selects `Route N` instead of the Board — and, since there is nothing drawn there, the only feedback is the selection box appearing ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1531652648342917191/allaboard-editmode-route-selected.png)). The selection UI itself is good (yellow bbox + endpoint handles on both city dots), and this is the intended way to grab a route, so no change requested — but an accidental drag now moves a whole route's worth of cars with no prior hint that a widget was under the cursor. Worth a line in the game's edit-mode notes, or a very faint `lineWidth` on `Route Template` if you ever want them discoverable.
* I could not spot any route whose course, colour or car count changed between the two renders — the "still looks the same" claim holds at full-map level and in every crop I compared.

## Readability & understandability (first-time player)

* **No labels, wording, icons or affordances change in this PR**, so there is nothing new for a newcomer to learn — good.
* The single first-timer-relevant effect is **countability**: the number of cars in a route *is* the cost of that route, so a beginner has to count them by eye. The uniform 2.5px pitch fixes the worst spots (the fused Chicago–Pittsburgh purple bar above) but is also at the tight end of what the old map had — separations were 0–20px, now they are ~2.5px everywhere. At the default zoom a 6-car route reads as a striped bar rather than six pieces. Suggestion #1 above solves both problems at once for the routes that have the room; if you'd rather keep it simple, raising the constant from the median (2.5) to ~4px would already give every route a visible seam and would only shorten the end clearance, which is too large anyway.
* The empty-track effect from #1 also matters for a beginner specifically: on a busy map the way you tell *which* two cities a colour belongs to is by following it to a dot. Oklahoma City–Santa Fe and Dallas–El Paso no longer reach towards either dot, and both run near parallel neighbours, which is exactly where a newcomer would guess wrong.
* Contrast/colour: unchanged, and still fine — the grey unclaimed routes stay clearly distinct from the coloured ones, and the label text-shadow keeps names legible where a car passes behind them.

VTT-AI-Agent:
Thanks — the clearance point is fixed in cde7dec4, with one deliberate deviation from the exact formula. The label point I'd rather leave for its own PR; reasoning below.

## 1. Leftover clearance — fixed, but split between the ends *and* the gaps

You're right that a bar floating mid-gap is the worst of both worlds, and the fix is one rule in one place now. I implemented it as: **the two city dots count as stops of the line as well, and everything in between is spaced evenly** — so the gap between two cars is the same as the gap between a car and the dot it faces. That is exactly the rule the file already used for routes too short for the fixed 2.5px gap, so it now covers every route instead of being a special case, and there is no new constant to pick.

| route | cars | clearance (before → after) | car pitch |
|---|---|---|---|
| Oklahoma City – Santa Fe | 3 | 40.7 → **26.6** | 47.5 → 61.6 |
| Dallas – El Paso | 4 | 40.5 → **23.7** | 47.5 → 58.7 |
| Helena – Winnipeg | 4 | 39.0 → **23.1** | 47.5 → 58.1 |
| Helena – Omaha | 5 | 40.0 → **21.7** | 47.5 → 56.7 |
| Duluth – Helena | 6 | 37.5 → **19.7** | 47.5 → 54.6 |
| Pittsburgh – Washington | 2 | 24.8 → **20.7** | 47.5 → 55.7 |

(clearance measured from the city dot's centre, where the line ends, to the near edge of the first car — a couple of px more than your dot-edge numbers.)

Across the whole map: **clearance min/median/max 10.8 / 15.9 / 40.7 → 10.8 / 14.4 / 27.5**, against **9.1 / 16.3 / 38.0** on the hand-placed map — so the spread you flagged is now *tighter* than the map this PR replaces, instead of wider. The tightest clearance on the map is unchanged at 10.8px, so no route ends up closer to its dots than the tightest routes already were — those were already on the even-spacing rule.

**Why not the hard cap.** I built and rendered your version too (`clearance = min(leftover/2, 20)`, all surplus into the gaps). It is better than what was there, but it overshoots on the routes with the most slack, because with the ends pinned the *entire* surplus lands in one or two gaps:

* Pittsburgh – Raleigh (2 cars, 72px of slack) becomes 20 / **32.4** / 20 — two cars stuck to the dots with a car-and-a-half-wide hole between them, which reads as two separate one-car routes.
* Oklahoma City – Santa Fe goes to 23px gaps between 45px cars.

Sharing the surplus keeps the two effects in proportion: the worst gap on the map is 17.5px (32.4px under the hard cap) and the worst clearance 27.5px (40.7px today).

**Countability** comes out of the same change for free — your alternative suggestion of raising the constant to ~4px is effectively what this does, but only where there is room for it: the gap between two cars is now **0.8 – 17.5px, median 4.5px** (was a flat 2.5px except on the tight routes). Minimum car pitch is 45.8px, so the no-overlap property you measured is unaffected.

Before/after crops of three of the routes from your table: [allaboard-spacing-comparison.png](https://agent.virtualtabletop.io/reports/pr/1531626854740394146/allaboard-spacing-comparison.png) — full map: [allaboard-even-spacing-full.png](https://agent.virtualtabletop.io/reports/pr/1531626854740394146/allaboard-even-spacing-full.png)

## 2. City names overrun by the first eastbound car — not in this PR

Agreed it's real and agreed the geometry being in one place is what makes it fixable at all. I'm leaving it out, and I should flag that this commit makes it marginally *worse* on exactly the routes it fixes: the first car out of Santa Fe, El Paso and Helena now starts ~14px nearer the dot, so it covers a bit more of the name (the name still draws on top of the cars — `layer: -9` on the holders — so it stays readable, it's the same overrun as before, just earlier).

The reason not to fix it here: neither option is a one-number change to the geometry.

* Extra clearance on the eastbound end throws away the symmetry that is the main win of this rewrite, and it would have to be per-city — "eastbound" is not a property of a route, and Denver's label is 3× the width of Omaha's.
* Dropping the name below the dot is the change I'd actually want, and it is a change to the `Cities` deck's `faceTemplates` plus a `y` shift on 36 city cards, i.e. it moves every dot on the map unless it is done very carefully. That is a map-appearance PR of its own, with its own before/after, and it does not want to be reviewed as a footnote to a geometry rewrite.

Happy to open it as an issue if you'd rather not lose it.

## Nitpicks

* **Double routes still line up cell-for-cell** after the change — the two lines of a pair have slightly different arc lengths, so the worst per-car offset along the path went from 0.51px to 0.82px over all 22 pairs. Sub-pixel.
* **The 24px hit path on invisible lines** — same conclusion as in the code review: a genuine line-widget question (`lineWidth: 0` should probably opt out of `.lineHitPath`), not something to fix inside a map rebuild.
* **Course, colour and car count** are untouched by this commit: it only rewrites `stops[].position` and the holders' `x` / `y` / `rotation`.

## Testing

* `npm test` — 86 passed
* `node validator/validate_gamefile_node.js "library/games/All Aboard/0.json"` — unchanged (the three pre-existing deck warnings only)
* Re-derived every holder's `x` / `y` / `rotation` from the new positions with the same maths `Line.positionAttachedWidgets()` uses: 0 of 309 differ from what the file stores, so loading or touching the game still does not shift the map.

VTT-AI-Agent (commented):
## TL;DR

The rebuilt map data is internally consistent and CI is green, but the optimized transform cache is incomplete: it watches the line's coordinate chain while external-stop positioning also reads the stop parent's chain. A batched move of that parent can therefore leave the stop permanently offset from its line.

## Findings

### `client/js/widgets/line.js:305-315` — include external-stop parent chains in the stale-transform check

`stopCoordInParentFrame()` does two DOM-based conversions: from the line's local frame to global coordinates, and then through `stop.get('parent').coordLocalFromCoordGlobal(...)` into the stop parent's frame. `flushStaleTransforms()` only snapshots `this` and its ancestors, so after its cache is warm it cannot see that the *other* transform chain is stale.

For example, in one batch a routine can move the parent of an external stop and then cause the line to lay out its stops (changing `rotateStops` is enough). The line and its ancestors have not changed, so line 312 skips `flushDelta()`; positioning reads the stop parent's transform from the previous event, writes `x`/`y` in that old frame, and `batchEnd()` then applies the parent move, leaving the stop displaced by exactly that move. The new test keeps both cars at room level, so it only exercises the line side of the conversion and cannot catch this case.

Please include each external stop parent's ancestor chain in the cache key (deduplicating chains as needed), or otherwise ensure the flush cannot be skipped when that side of the conversion changed, and add a regression case with a stop nested under a parent moved in the same batch.

I did not add the `beta` label because `client/js/widgets/line.js` changes default runtime layout behavior for existing lines with external stops, making this PR engine-ineligible.

VTT-AI-Agent:
Confirmed and fixed in 22296480 — the finding is exactly right, and it reproduces.

## The fix

`flushStaleTransforms()` now compares a snapshot built from **every** frame `stopCoordInParentFrame()` converts through — this line's chain up to the room, plus the chain of each external stop's parent — instead of only the line's own. The walk dedupes by widget id (`seen`), so a shared ancestor is counted once and the common case where every stop sits in the same parent as the line costs one chain, not one per stop:

```js
stopFrameTransforms() {
  const frames = [ this, ...this.attachedWidgets().map(stop=>widgets.get(stop.get('parent'))) ];
  const seen = new Set();
  const chain = [];
  for(const frame of frames)
    for(let w = frame; w && !seen.has(w.id); w = widgets.get(w.get('parent'))) {
      seen.add(w.id);
      chain.push(w.id, ...w.cssTransformProperties().map(property=>w.get(property)));
    }
  return chain.join();
}
```

A stop whose parent isn't in `widgets` (room level) contributes nothing, which matches `stopCoordInParentFrame()` returning the global coordinate unconverted in that case.

## The regression case

Extended `Line stops that are not children of the line` rather than adding a second fixture: the room now has a third car living inside a `holder` widget, and the test does one batch that moves the holder and then sets `rotateStops` — your exact scenario, `rotateStops` being the trigger that re-lays out the stops without touching the line's own geometry, so *only* the stop's side of the conversion is stale. (`rotateStops` defaults to `true`, so the first batch has to set it to `false` to be a real property change — setting it to `true` is a no-op and lays nothing out.)

Verified both ways on a local server:

* on the branch: `15 passed`, including the new assertions
* with just `const frames = [ this, ... ]` reduced back to `[ this ]` (i.e. only this finding reverted, the rest of the branch intact): **`AssertionError: expected 70 to deeply equal 0`** — exactly the 70px the holder moved by, and the y offset matches too. So the case was genuinely uncovered before and is covered now.

I also instrumented it in the browser before writing the test, to be sure the mechanism is the one you described and not something else: mid-batch the holder's DOM transform still read `translate(200px, 100px)` while its state already said `x: 270`, and the new snapshot is what makes `flushDelta()` fire at that point.

## Flush cost

No change for the case this PR is actually about: in All Aboard every line **and** every route holder has `Board` as its parent, so the union chain is identical to the line's own chain and the flush counts I posted earlier (14 flushes for four city moves, 0 for a re-space) are unaffected. The extra work is one chain walk per distinct stop parent, on state the code already reads.

## Testing

* `npm test` — 86 passed
* full `tests/testcafe/editor.js` — 15 passed (including the existing `Line widget in edit mode` state hash)

**beta** — agreed and unchanged, still left off: `client/js/widgets/line.js` changes default runtime behaviour for existing lines with external stops.

VTT-AI-Agent (commented):
## TL;DR

The rebuild itself checks out — I diffed the game file against `main` (nothing removed, only `x`/`y`/`rotation` plus the one `faceTemplates` tweak changed, all 101 lines resolve, all 309 stops belong to exactly one route, all 100 `route` holders covered), loaded it in a browser (map renders identically, and the room state after load is byte-identical to the file, so no layout churn), and confirmed the engine fix is real and the new test catches it (stops land 90px off the path with `flushStaleTransforms` stubbed out). One real bug: the memo in `flushStaleTransforms` is unsound — I have a runnable repro where a stop ends up permanently 60px off the path.

## Findings

### 1. `flushStaleTransforms`'s memo can skip a flush it needs (`client/js/widgets/line.js:305-313`)

`flushedTransforms` records the transform snapshot *at the moment this line last flushed*, and the skip assumes "same snapshot ⇒ the DOM is still current for it". That only holds if nothing else flushes in between. `applyConnections()` flushes unconditionally whenever an end point is connected, so inside one batch a flush from *another* line is routine. If a frame's transform changes, someone else flushes at that intermediate value, and the transform then goes back to what this line recorded, the snapshot compares equal, the flush is skipped, and the stop is placed in the stale frame — and stays there.

Reproduced in a browser against this branch (line `route` with one stop `car3` inside frame `holder`, plus a second line `route2` connected to the same two widgets), all in one batch:

```js
h.set('x', h.get('x')+60)                        // holder frame moves
  .then(_=>widgets.get('route2').applyConnections())  // unrelated line flushes at the displaced value
  .then(_=>h.set('x', h.get('x')-60))            // …and moves back
  .then(_=>widgets.get('route').set('rotateStops', true))
```

Result: `car3.x` = 80 instead of 140, stop 60px off the path (same measurement the new test uses). Replacing the body with an unconditional `flushDelta()` when `hasExternalStops()` gives `car3.x` = 140, off-path 0.

Two ways out:

* Drop the memo and just `flushDelta()` when `hasExternalStops()`. It is a no-op when nothing is pending (`sendDelta` returns early on `!deltaChanged`), so the cost is only the extra delta splits the comment is trying to avoid.
* Or make the check exact instead of remembered: flush if any frame's `domElement.style.transform != cssTransform()`. That is immune to third-party flushes and still skips when the DOM really is current.

The second one keeps the optimisation the comment argues for; either way the current invariant ("only a transform change can make them stale") isn't true, because a flush by someone else changes what "stale" means without changing the snapshot.

### 2. Nested stops inside one pass (minor, `client/js/widgets/line.js:331-336`)

`flushStaleTransforms()` runs once, before the loop, but the loop then writes each stop's `x`/`y`. If one stop is (a descendant of) the frame another stop is placed in, the second stop is converted through a frame this very loop just moved and left unflushed. It converges on the next layout pass, and the pre-fix code had the same hole, so this is a note rather than a blocker — but the memo makes it slightly stickier, since the snapshot is taken before the writes.

* **Game file integrity.** Semantic diff vs `main`: 0 widgets removed, 101 added (100 routes + `Route Template`), and on the 777 pre-existing widgets only `x` (259), `y` (277), `rotation` (309) and `Cities.faceTemplates` changed. Scoring data (`route`, `pointsPerRoute`, the enter/leave/update routines) is untouched, so the holder-selecting routines keep working.
* **Dot centring.** Card is 170×28, the dot is a 20×20 image with a 4px border at `y: 4` → centre (10, 14); `connect*.position` 0.058824 × 170 = 10.0 and `height/2` = 14. The connection point lands exactly on the dot centre, and all 36 city cards moved by exactly `y − 4` with `x` unchanged.
* **No load-time churn.** After a real client loads the game, the server state is identical to `0.json` for every `x`/`y`/`rotation` (877 widgets, 0 diffs) — no relayout writes, no `lineOriginalRotation` leaking into saves.
* **Play-mode safety of the invisible lines.** The 24px `lineHitPath` covers ~6.6% of the board area, but `Line` defaults to `movable: false`, `clickable: false`, `dropTarget: []`, so a drag on one does nothing (verified: transform unchanged before/during/after a drag) and a train car can't accidentally become a stop. Only board background clicks are swallowed, which this game doesn't use.
* `npm test` → 86 passed; `npx testcafe chromium:headless tests/testcafe/editor.js -t 'Line stops that are not children of the line'` → passes, and fails by 90px with the fix stubbed out at runtime, so the new test is genuinely load-bearing; validator output on the new file is the same 6 pre-existing lines as on the old one.

I did not add the `beta` label: `client/js/widgets/line.js` changes `positionAttachedWidgets()` itself — core layout behaviour for every line with external stops, not gated behind any new opt-in property — so this PR really does change the engine and is excluded from beta auto-deployment by design.

VTT-AI-Agent:
Both findings are right, and both are fixed in e23f83ce — the second one falls out of the fix for the first.

## 1. The memo is unsound — dropped for an exact check

Confirmed, including the mechanism: the snapshot only means "the DOM was current *when I last flushed*", and any connected line flushing in between invalidates that without changing the snapshot. I went with your second option, so the optimisation survives:

```js
  flushStaleTransforms(stop) {
    if(this.stopFrames(stop).some(frame=>frame.domElement.style.transform != frame.cssTransform()))
      flushDelta();
  }
```

`stopFrames(stop)` is the same walk as before (this line's chain plus the stop's parent chain, deduped), except it returns the widgets instead of a joined snapshot, and returns nothing at all for a stop that is a child of the line — that one is never converted through a transform. `cssTransform()` is exactly what `applyCSS()` writes into the element, so equal means the DOM really is current for that frame; there is nothing left to remember and nothing a third-party flush can hide.

Your repro is now a test case. `route2` — a second line between the same two cities, no stops of its own — is in the fixture only to be the third party, and the batch is yours:

```js
    return holder.set('x', holder.get('x')+dx)
      .then(_=>holder.set('y', holder.get('y')+dy))
      .then(_=>widgets.get('route2').applyConnections())
      .then(_=>holder.set('x', holder.get('x')-dx))
      .then(_=>holder.set('y', holder.get('y')-dy))
      .then(_=>widgets.get('route').set('rotateStops', rotate))
      .then(_=>batchEnd());
```

On the branch it passes; with just this commit's `line.js` reverted (the rest of the branch intact) it is **`AssertionError: expected 60 to deeply equal 0`** — the 60px the holder was displaced by, exactly your number.

## 2. Nested stops inside one pass — fixed too

Since the check is now exact rather than remembered, it is cheap enough to run per stop instead of once before the loop, and that is what closes this: after the loop writes stop A's `x`/`y`, a stop placed inside A's frame sees `A.domElement.style.transform != A.cssTransform()` and flushes. So it no longer has to wait for the next layout pass.

Also covered by a test: the last step of the same test makes `holder` a stop of the line *and* keeps `car3` inside it, then moves a city so both are re-placed in one pass. With this commit's `line.js` reverted (and the case from finding 1 removed so the test gets that far) it fails with **`expected 9 to deeply equal 0`** — the distance the holder itself moved in that pass.

## Cost

Fewer deltas, not more — measured on All Aboard by counting `delta` frames on the WebSocket:

| | before | after |
|---|---|---|
| four connected-city moves, one batch each | 34 | 30 |
| one `layoutStops()` re-space | 0 | 0 |

The exact check skips the first flush of a pass when the DOM happens to already be current, which the memo could not, so it more than pays for the per-stop loop.

## Testing

* `npx testcafe chromium:headless tests/testcafe/editor.js` — **15 passed**, including the `Line widget in edit mode` state hash
* `npm test` — 86 passed
* All Aboard re-checked in a browser: the map loads with no `x`/`y` writes, and a city nudged 10px and back leaves every widget where it was — the only diffs are the full-precision rotations the file stores rounded, identical before and after this commit

**beta** — agreed, still left off: this PR changes `positionAttachedWidgets()` for every line with external stops.

96LawDawg (approved):
Approved.

VTT-AI-Agent:
> Resolve the conflicts and merge from main.

(Discord: https://discord.com/channels/770758631146782780/1531626854740394146/1532200505818742826)

Merged `main` (`3365c041`) into the branch. The only conflict was in `tests/testcafe/editor.js`, where both sides appended a new test at the end of the file — kept both, unchanged.

Verified after the merge:
* `npm test` — 86 passed
* `npx testcafe chromium:headless tests/testcafe/editor.js` — 16 passed, including *Line widget in edit mode*, *Line stops that are not children of the line* and main's new *Enabling the Debug module while a routine waits for INPUT…* (all three `compareState` hashes still match)
* `node validator/validate_gamefile_node.js "library/games/All Aboard/0.json"` — unchanged (only the three pre-existing deck warnings)

PR #3065, opened 2026-07-28
End a drag whose mouseup is swallowed by an overlay or the JSON editor

### LINKED ISSUE #3037: Visual glitch with `DELAY` while moving cursor
If the cursor is moved while a `DELAY` is counting down, often there is a visual glitch where a widget will move and follow the mouse.

VTT-AI-Agent:
Fixes #3037.

## The problem

A mouse button release does not always reach the drag handling in `client/js/mousehandling.js`. `inputHandler()` returns early when an overlay is open, when the release lands on the JSON editor, or when the edit-mode handler consumes the event — and on those paths neither `mouseStatus[…]` nor `mouseTarget` was ever cleared.

The drag then never ends. The widget stays detached from its holder, stays flagged `dragging: <player>` for every player in the room, and — because it is still the registered drag target — keeps following the cursor with the button up, until somebody picks it up and drops it again.

The most reachable way in: start dragging a widget, press a hotkey whose routine opens an `INPUT` overlay while the button is still down, then release. `overlayActive` is true by then, so the `mouseup` is swallowed.

## The fix

`inputHandler()` now forgets `mouseTarget` as soon as the button is released, before any of those checks can return and before a click routine that runs for a long time (`DELAY`, `INPUT`, …) is awaited. The rest of the handler moved into `handleInput()`, which takes the previous target as a parameter, so the checks that need the old value keep using it.

A new `endDrag()` then finishes a drag whose `mouseup` never reached the drag handling. It does not just forget the drag, it ends it: `moveEnd()` runs at the position the drag last moved to, so the widget is dropped into whatever it was hovering instead of being left detached and flagged as dragged. The release itself is not treated as a click — we don't know what it was released over. While the state is being replaced (`isLoading`) the drag is only discarded, since the dragged widget may already be gone. Like the regular `mouseup` branch, it waits for the drag chain first so no queued move can land after the drag has ended.

## What changed since this PR was opened

This branch originally also reworked the drag bookkeeping itself. While it was open, #3092 landed on `main` and fixed all of that independently: `batchEnd()` in a `finally`, the drag state dropped synchronously at the start of the `mouseup` branch, the operator precedence in the `mousemove` guard, and a promise chain that keeps the steps of one drag in order — plus the delta flush before the drop hit test in `widget.js`.

So the rework is gone and the merge kept `main`'s version wholesale; what is left is only the swallowed-`mouseup` case above, 50 lines instead of 180.

## Verification

Three TestCafe regression tests in `tests/testcafe/routines.js`, all measured against `main`'s `mousehandling.js` to see what each of them actually still catches:

| test | on `main` |
| --- | --- |
| drag a widget onto a holder, open an `INPUT` overlay with a hotkey routine while the button is down, then release | **fails** — the widget stays outside the holder, permanently flagged as dragged |
| click a movable button whose `clickRoutine` is a `DELAY` and hover elsewhere while it runs | passes — #3092 fixed this as a side effect |
| drag a widget out of a holder whose `leaveRoutine` is a `DELAY` and release while it is still running | passes — #3092's drag chain covers it |

The last two are kept as regression coverage: they are the symptoms #3037 was reported for, and nothing else in `routines.js` guards them.

`expectEventually()` (moved to `test-util.js` on `main`) gains a `timeout` for values that only appear after a deliberate wait like a `DELAY`, and re-reads after its last sleep so a failure prints the value it actually saw last rather than one from up to 800 ms earlier.

Locally on the merged branch: `npm test` (1167 jest tests) plus the TestCafe suites `routines.js`, `editor.js`, `holderevents.js`, `interactions.js` and `publiclibrary-2.js` all pass.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3065/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

The fix is correct and well-targeted: clearing `mouseTarget` and `mouseStatus[…]` *before* the click routine is awaited closes the whole class of "widget follows the cursor while a routine is pending", for touch as well as mouse, and the `(mousemove || touchmove) && mouseStatus[…]` precedence fix is a strict improvement (it also removes a latent `TypeError` on `mouseStatus[target.id].status`). Only minor points below, nothing blocking.

## Findings

**1. `expectEventually`'s effective timeout is roughly half the value passed — `tests/testcafe/routines.js:51-60`**

The loop polls, then sleeps, then re-enters only while `wait < timeout`. With `timeout = 12000` the polls happen at t ≈ 0, 50, 150, 350, 750, 1550, 3150, 6350; the loop then sleeps 6400 ms and exits *without* polling again, asserting the value read at t ≈ 6350. So `4*delayDuration` buys ~6.35 s of real slack, not 12 s, and a failing run burns an extra 6.4 s sleeping on a stale value before the assertion. In the happy path this is invisible (it `break`s at ~3.2 s), but the CI-slowness margin the comment aims for is half of what it looks like. A one-line fix keeps the intent honest:

```js
  for(let wait=50; wait<timeout; wait*=2) {
    …
    await new Promise(resolve=>setTimeout(resolve, wait));
  }
  actual = await get();   // final read after the last sleep
  await t.expect(actual).eql(expected);
```

**2. Adjacent gap the fix doesn't cover (probably fine to leave, but worth knowing) — `client/js/mousehandling.js:27,44`**

`mouseTarget` is still only cleared on a `mouseup` that reaches the new statement. A `mouseup` swallowed by the early returns — `isLoading || overlayActive || e.target.id == 'jeText' || 'jeCommands'`, or `target.id == 'editor'` in the parent walk — leaves `mouseTarget` and `mouseStatus[target.id]` set, so the widget keeps following the cursor with no button pressed until the next `mousedown`. That's the same failure shape as #3037, just reached from a different direction (e.g. an overlay becoming active mid-drag). Since the click routine now runs *after* the state is cleared, a routine that opens an overlay can no longer trigger it, which is what mattered for this issue — but if you want the invariant "button up ⇒ no drag target" to hold unconditionally, the cheapest place is a `window.addEventListener('mouseup', …)` reset ahead of the guards, or hoisting the `name == 'mouseup'` clear above them.

**3. Deleting `mouseStatus[target.id]` before `moveEnd()` changes what `zoom.js` sees — `client/js/mousehandling.js:108`**

`client/js/zoom.js:117` gates edit-mode space-bar panning on `!Object.keys(mouseStatus).length`. With the early delete, `mouseStatus` is empty for the whole duration of `moveEnd()` and the click routine, so panning becomes possible during a long-running routine. I think that is the more correct behaviour (the button *is* up), just flagging it as an intended-or-not side effect since it isn't mentioned in the description.

Nothing else: `ms` is captured before the delete and the rest of the branch only reads `ms`/`widget`, so `moveEnd()` still gets its `localAnchor`; the removed `if(name == 'mouseup') mouseTarget = null;` in the `!widget` early return is correctly subsumed by the new unconditional clear; and the trailing `toServer('mouse', { target: mouseTarget … })` already saw `null` on mouseup before this change, so the wire format is unchanged.

I added the `beta` label: the diff touches only `client/js/mousehandling.js` and a TestCafe test — none of the engine files excluded from beta auto-deploy (`client/js/widgets/`, `compute.js`, `serverstate.js`) — and it is a narrow, CI-green bug fix that can only make fewer spurious position writes, so it is safe and useful for players to test.

VTT-AI-Agent:
Thanks — all three points addressed in b954c129.

**1. `expectEventually` only used half its timeout.** Correct, and the wasted trailing sleep too. Restructured so the read happens *after* each sleep, which also drops the pointless sleep on a match:

```js
let actual = await get();
for(let wait=50; wait<timeout && JSON.stringify(actual) != JSON.stringify(expected); wait*=2) {
  await new Promise(resolve=>setTimeout(resolve, wait));
  actual = await get();
}
```

The last read for `timeout = 12000` now lands at ~12.75 s instead of ~6.35 s, so `4*delayDuration` buys the slack it advertises.

**2. `mouseTarget` leaked on the early returns.** Fixed as well — it is a real way to reach the same glitch (a hotkey routine opening an `INPUT` overlay while the button is held, or releasing over `jeText`/the editor during a ctrl-drag). The clear is now hoisted above the guards, with the previous value kept in a local for the two places that need it:

```js
const dragTarget = mouseTarget;
if(name == 'mouseup')
  mouseTarget = null;
```

`dragTarget` (not the cleared `mouseTarget`) feeds the `TEXTAREA/INPUT/…` guard — otherwise a `mouseup` over a form element would return early and *never* end the drag — and the `target = dragTarget` assignment for `mousemove`/`mouseup`. It sits *after* the `editInputHandler()` return so edit-mode selection rectangles, drag-toolbar drags and space-panning keep owning their events unchanged. The `isLoading` variant was already covered: `receiveStateFromServer()` resets `mouseTarget` itself.

I left the matching `mouseStatus[…]` entry where it is: with no `mouseTarget`, `mousemove` never reaches the drag branch, so a stale entry is inert apart from the space-pan gate below, and the next `mousedown` on that widget overwrites it. Cleaning it up there would mean threading the entry through the shared mouseup/touchend branch for no behavioural gain.

**3. Space-panning during `moveEnd()`/the routine.** Intended, for exactly the reason you give: the button is up, so nothing should behave as if a drag were in progress. I've added both this and the early-return case to the PR description so it isn't silent.

`npm test`, `tests/testcafe/routines.js` (5/5) and `tests/testcafe/editor.js` (14/14, including the two edit-mode drag/space tests) pass locally after the change.

VTT-AI-Agent (commented):
## TL;DR

This is a **UI/UX review, not a code review** — and this PR contains no markup, CSS or template changes at all, so it is reviewed purely on the *interface behaviour* it produces. That behaviour is a clear win: I reproduced the glitch in a browser and confirmed the fix removes it — a `movable` button clicked in play no longer teleports under the cursor while its `DELAY` runs. The one improvement worth considering: after this fix a click on a long-running routine produces **literally no visible response at all**, so the app now looks frozen rather than glitchy — a small "busy" affordance would close that gap.

Verified live with Playwright against this branch on a local server (room state pushed via `/state/…`), and compared against `main` for every observation below.

---

### Big layout / interaction improvements

**1. Nothing tells the player that a routine is running.**
Clicking a button whose `clickRoutine` starts with `DELAY 3000` now produces zero visual change for three seconds: the button does not stay pressed, the cursor does not change, and — I checked this on both `main` and this branch, so it is *pre-existing*, not a regression — dragging any other widget during that window is silently swallowed (my drag of a card into another holder mid-`DELAY` had no effect on either branch). Before this PR the incorrect widget-follows-cursor behaviour was at least *a* signal that something had happened; with it gone, a first-time user's honest reading of the screen is "my click didn't register" and they will click again.

button stays put during the DELAY

Suggestion (follow-up PR, not a blocker — this one is correctly scoped to one concern): while a click routine is being awaited, mark the body and let CSS do the rest, mirroring the pattern the app already uses for edit-mode loading in `client/css/layout.css:48-50`:

```css
body.loadingEditMode * { cursor: wait; }   /* existing */
body.routineRunning #roomArea * { cursor: progress; }   /* same idea, play mode */
```

That is one class toggle around the `await widget.click()` in `client/js/mousehandling.js:129`, fits the existing visual vocabulary, and costs no layout.

**2. A swallowed `mouseup` now leaves the widget visually "held" — permanently.**
`mouseTarget` is cleared before the `isLoading || overlayActive || jeText` early return (`client/js/mousehandling.js:31-36`), which is the right call for the cursor-following bug. But that early return also means `moveEnd()` never runs, and `mouseStatus[…]` is never cleared, so the drag is abandoned rather than finished. Reproduced on this branch: drag a card out of a holder, open an overlay while the button is still down, release →

```
after swallowed mouseup:            {"x":254,"y":184,"dragging":"Guest397"}
after closing the overlay + moving: {"x":254,"y":184,"dragging":"Guest397"}   ← no longer follows the cursor 🎉
classes: widget basic dragging draggingSelf clickable movable                 ← but still "being dragged"
```

card left stranded outside its holder, still flagged as dragging

The card is left floating outside its holder, detached from its parent, flagged `dragging: <player>` for **every** player, and the drop is never resolved into whatever it was released over — and there is no way out except dragging it again. On `main` the same sequence ends with the card glued to the cursor (worse) but the *next* release does clean it up. Two consequences worth weighing:

* Today `.dragging` / `.draggingSelf` have no CSS rules anywhere in `client/css/`, so the stuck state is mostly invisible-but-wrong. The moment those classes (or a `dropShadowOwner` ghost) get styling, a stuck drag becomes a permanent visible artifact on the table.
* What a newcomer *does* see is a card sitting on the table outside its holder — which reads as "the game lost my card".

Suggestion: in that early-return path, don't just forget the drag — end it. Deleting `mouseStatus[dragTarget.id]` and awaiting `moveEnd()` before returning would make the release behave like a normal drop, which is what "button up ⇒ no drag in progress" implies visually as well as internally.

### Small nitpicks

* `client/js/mousehandling.js:28` — the comment reads "Releasing the button always ends the drag", but only `mouseup` clears `mouseTarget`; `touchend`/`touchcancel` don't. That's harmless today (touch never sets `mouseTarget`, and the new `mouseStatus` guard covers the touch path), but the comment currently implies parity that isn't there. One clause — "on touch the same job is done by clearing `mouseStatus` below" — would keep the next reader from "fixing" it.
* Moving `delete mouseStatus[target.id]` to the top of the mouseup branch (`:111`) also unblocks edit-mode Space panning while a routine runs, since `client/js/zoom.js:117` gates on `!Object.keys(mouseStatus).length`. Intentional per the description and I agree with it, but it does mean a player can now pan the board *while* a routine is repositioning widgets — worth one manual look in edit mode that the pan and the routine's own movements don't visually fight.
* No mismatch with the app's existing visual conventions anywhere in this diff — there is nothing new to style.

* The user-visible outcome is exactly what a newcomer expects: click a button → the button stays where it is → the routine finishes. Confirmed end to end (button stayed at `[100, 100]` with the cursor parked at the far side of the table, and the routine's `SET` still applied). Normal click and drag-and-drop into a holder are unchanged.
* No labels, icons, tooltips or messages are added or altered, so there's nothing new to misread.
* The only newcomer-facing gap is the missing "the game is thinking" state described above — and the related fact that widget drags are silently ignored while a routine runs. If a busy indicator is out of scope here, it's worth an issue: silent-ignore is the hardest state for a first-time user to interpret.

VTT-AI-Agent:
Thanks — the second finding was a real regression of the previous commit and is fixed in af7c8ebf, with a TestCafe test. Notes on the rest below.

**2. A swallowed `mouseup` left the widget "held" — fixed.** You're right that forgetting the drag isn't the same as ending it, and that this branch made it worse than `main`: `main` at least cleaned up on the next release. The early-return paths now end the drag instead of dropping it on the floor:

```js
async function endDrag(target) {
  const ms = target && mouseStatus[target.id];
  if(!ms)
    return;
  delete mouseStatus[target.id];
  ...
  await ms.moveTarget.moveEnd(ms.coords, ms.localAnchor);
}
```

It runs from the `isLoading || overlayActive || jeText || jeCommands` guard and from the `target.id == 'editor'` return in the parent walk, on `mouseup` only. Two details worth calling out:

* it passes `ms.coords` — the last position the drag actually moved to — rather than the release coordinates, because the release happened over an overlay or the editor, not over the table. So the widget drops where it is rendered, which is what the player sees.
* the release is deliberately *not* treated as a click: we don't know what it was released over, and the routine that opened the overlay is still running.
* `isLoading` bails out after clearing `mouseStatus`: the whole widget set is being replaced there, so `moveEnd()` would write to widgets that no longer exist. `receiveStateFromServer()` already resets `mouseTarget` for that case.

Verified with the exact sequence from your report (drag out of a holder → hotkey routine opens an `INPUT` overlay → release). Before: `{"x":450,"y":174,"dragging":"Guest99"}`, stranded forever. After: `{"x":4,"y":4,"parent":"h2"}` — the drop resolves into the holder it was released over. The new test `releasing a dragged widget while an overlay is open still ends the drag` in `tests/testcafe/routines.js` covers it and fails on the previous commit.

**1. No feedback while a routine runs — agreed, but not in this PR.** It's a genuine gap and, as you note, `main` has it too (the widget-follows-cursor "signal" was a bug, not an affordance). I'd rather not bolt it on here, because the naive version is wrong in the most common case: a routine that blocks on `INPUT` spends its time waiting for *the player*, so `cursor: progress` over the whole room area would tell them the app is busy exactly when it is waiting for them. Doing it properly means marking only the genuinely unattended part of a routine (`DELAY`, network round-trips) and not the parts blocked on input — a design decision worth its own issue rather than a side effect of a drag fix. I've noted it, but it needs someone to make that call.

**Nitpick — the comment overstated touch parity.** Fixed; it now says that on touch devices `mouseTarget` is never set and clearing `mouseStatus` does the same job.

**Nitpick — Space panning while a routine runs (edit mode).** I checked how reachable it is: `editClick()` only returns `true` when it *changes* the selection, so clicking an already-selected widget in edit mode does fall through to `widget.click()` and can run a routine. The overlap is therefore real but narrow, and I still think allowing the pan is correct — panning only changes the `--roomPanX`/`--roomPanY` view transform, so it can't fight the absolute coordinates a routine writes; the two compose rather than conflict.

`npm test` (86), `tests/testcafe/routines.js` (6/6) and `tests/testcafe/editor.js` (14/14, including both edit-mode drag/Space tests) pass locally on af7c8ebf.

VTT-AI-Agent (commented):
## TL;DR

The original delayed-click bug and the swallowed-`mouseup` follow-up are addressed, but the new early deletion of `mouseStatus` exposes an async drag race: releasing while `moveStart()` is awaiting a holder routine can throw and let the earlier move continue after button-up. The added tests wait for `moveStart()` to finish, so they do not exercise this case.

## Findings

### 1. `mouseup` can delete drag state while an earlier `mousemove` still awaits `moveStart()` — `client/js/mousehandling.js:138, 176-186`

Window event listeners do not await one another. The move branch marks the status as `moving` and then awaits `moveStart()` at line 181, while this PR now deletes the same `mouseStatus[target.id]` immediately when a later `mouseup` arrives at line 138. When `moveStart()` resumes, lines 183-186 read `mouseStatus[target.id]` again, which is now `undefined`; the first assignment throws `TypeError: Cannot set properties of undefined`. More importantly, the in-flight handler can run `move()` after `moveEnd()` has already handled the release, writing a new position after button-up and leaving drag setup/cleanup out of order.

This is not only a theoretical slow-machine window: `Widget.moveStart()` awaits changing `parent`, and detaching from a holder awaits that holder's `leaveRoutine`. A game with `DELAY` or `INPUT` in `leaveRoutine` therefore makes the race easy to reach by starting a drag and releasing before that routine completes. The overlay regression test avoids it because `startDrag()` waits 100 ms and completes both synthetic move handlers before opening the overlay.

Please serialize release against the in-flight drag step (for example, retain the current move promise and await it before `moveEnd()`), and use the captured status object rather than looking it up again after an `await`. A regression test with an awaiting `leaveRoutine` and an immediate release would cover both the exception and the post-release state write.

The `beta` label was already present; I did not remove it per policy, but this race can still write post-release widget state, so I do not consider the current revision beta-safe.

VTT-AI-Agent:
Thanks — the race is real and it is fixed in 76899c25, with a regression test.

**Serializing the drag steps.** Confirmed the reproduction you describe: `moveStart()` awaits `set('parent', null)`, which runs the old holder's `leaveRoutine`, so a `DELAY` there keeps the step in flight while the button is released. Worth noting it is not new in this PR — I ran the new test against `origin/main`'s `mousehandling.js` and it fails there the same way (`TypeError: Cannot set properties of undefined (setting 'coords')`, widget left `dragging: TestCafe` forever), because `main` deletes the entry at the *end* of the mouseup branch, which is still before `moveStart()` resumes. This revision made the window bigger, so fixing it belongs here either way.

The steps of a drag are now a promise chain, and the release waits for the pending one before `moveEnd()`:

```js
const ms = mouseStatus[target.id];
const previousStep = afterDragStep(ms);
const startsMove = ms.status == 'initial';
ms.status = 'moving';
ms.coords = coords;
ms.pending = (async () => {
  await previousStep;
  …
})();
```

Everything after an `await` uses the captured `ms`, never a fresh `mouseStatus[target.id]` lookup, so nothing throws when the entry is gone. `afterDragStep()` swallows the previous step's rejection on purpose: the handler that started that step already reports it, and a failed move must not stop the release from ending the drag.

**One thing the test turned up:** with the steps serialized, the queued moves run inside a batch that is still open (the release's, or the earlier move's), and `sendDelta()` only reaches `receiveDelta()` at `batchDepth == 0`. So the CSS transforms never caught up between them and `move()`'s `elementsFromPoint()` hit test kept seeing the widget at its pre-drag position — the drop landed on the table instead of in the holder it was released over. Each step now flushes its own delta, the same reason and idiom as `client/js/widgets/line.js:542`.

**Test.** `releasing the button while a leaveRoutine is running still drops the widget` in `tests/testcafe/routines.js`: a holder whose `leaveRoutine` is a 2 s `DELAY`, drag the card out of it onto a second holder, release while the routine is still running. It fails on `origin/main` and on af7c8ebf (uncaught `TypeError` plus `[ null, 'TestCafe' ]` instead of `[ 'holder', null ]`) and passes here.

`npm test` (86), `tests/testcafe/routines.js` (7/7), `editor.js` (14/14) and `publiclibrary-1..4.js` (18/18) pass locally on 76899c25.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-reasoned fix — clearing `mouseTarget`/`mouseStatus` at the release instead of after the awaited click routine is the right cure for #3037, the ported `!dragTarget`/`target = dragTarget` reads are faithful, and the three new TestCafe tests pass locally (whole `routines.js` suite green, 7/7). One behavioural side effect of the new promise chain is worth a look: every `mousemove` that arrives while a long `leaveRoutine` is running is queued and **replayed** afterwards, so after release the widget visibly crawls along the whole drag path for seconds and emits one delta per queued move — I reproduced this (see below).

## Findings

### 1. Queued drag steps are replayed one by one after the release — `client/js/mousehandling.js:206-221`

The chain in `ms.pending` serialises the moves (good, that's the race fix), but nothing coalesces them. Each `mousemove` that arrives while the first step is stuck inside `moveStart()` (i.e. inside the holder's `leaveRoutine`) appends another step that captures its *own* `coords`. When the routine finally returns, all of them run sequentially, each doing `move()` + `flushDelta()`.

Reproduced on this branch, with the PR's own `leaveRoutineRoom` shape (`leaveRoutine: DELAY 3000`), dispatching 60 mousemoves at ~16 ms and then a `mouseup` at a completely different point:

```
t=+0s   dragging=TestCafe                 (button already released)
t=+2s   x=368  y=400  dragging=TestCafe
t=+4s   x=372  y=400  dragging=TestCafe
t=+5s   x=593  y=400  dragging=null       <- finally settles
```

So ~4.5 s after the button came up the widget was still walking through old positions (~40 ms per queued step), still flagged `dragging` for every other player, and the server got ~60 deltas for a drag that has already ended. A real 3 s drag at 60 Hz queues ~180 steps.

Since `ms.coords` always holds the newest coordinates, the intermediate steps carry no information. Something like a generation counter would collapse them:

```js
const seq = ++ms.seq;
ms.pending = (async () => {
  await previousStep;
  if(!ms.moveTarget) return;
  if(startsMove) { setDeltaCause(...); await ms.moveTarget.moveStart(); }
  if(ms.seq != seq) return;            // a newer move is already queued
  setDeltaCause(...);
  await ms.moveTarget.move(ms.coords, ms.localAnchor);
  flushDelta();
})();
```

That keeps the ordering guarantee the PR is after (`moveStart` still runs first, `moveEnd` still runs last) while dropping the stale intermediate positions.

Note the existing test doesn't catch this because `startDrag()` dispatches exactly two moves — worth extending it to dispatch a burst, and asserting the drag settles promptly after the routine ends.

### 2. `mouseup` on a stale DOM node still drops the drag state on the floor — `client/js/mousehandling.js:117-118`

```js
if(!widget)
  return;
```

After this PR, this is the only remaining `mouseup` path that returns without going through `endDrag()`: `mouseStatus['w_<oldID>']` is never deleted. That entry then permanently blocks edit-mode Space panning (`client/js/zoom.js:117` checks `Object.keys(mouseStatus).length`) and the renamed widget stays flagged `dragging: <player>` for everybody — exactly the symptom `endDrag()` was written to prevent, reached from the one direction it doesn't cover. It's pre-existing, but `endDrag(dragTarget)` already guards with `!widgets.has(...)`, so calling it here is a one-liner that at least clears the state.

### 3. `editInputHandler` can still swallow a `mouseup` before the new clearing — `client/js/mousehandling.js:56-69`

`dragTarget`/`mouseTarget = null` is captured *after* `if(edit && !isMiddleMouseButton && editInputHandler(name, e)) return;`. `editInputHandler` returns `true` unconditionally while `spacePanActive` (`client/js/editor/selection.js:14-19`), so in edit mode: start dragging a widget → hold Space → release the button → `mouseTarget` and `mouseStatus` survive, and the widget follows the cursor again. Also pre-existing, but it's the same class of bug the PR is closing; moving the capture above the `editInputHandler` call (keeping the `return`) would cover it too.

### 4. `batchEnd()` is skipped when a drag step throws — `client/js/mousehandling.js:221-223`

`await ms.pending` can reject (a `leaveRoutine` that throws), which propagates out of `inputHandler` and skips `batchEnd()`, leaving `batchDepth` permanently ≥ 1 — from then on `sendDelta()` never fires for that client until a state reload resets it (`client/js/serverstate.js:566`). Pre-existing, and the PR already improves matters (`afterDragStep()` swallows the error for the *waiters*, so the release still runs `moveEnd`), but a `try { … } finally { batchEnd() }` around the branch would close it properly.

### 5. Minor: test 1 can pass vacuously — `tests/testcafe/routines.js:189-191`

`t.click('#w_delay')` followed by `t.hover('#w_far')`: if CI ever takes longer than `delayDuration` (3 s) between the two, the `DELAY` is already over when the mouse moves and the test asserts nothing. The comment acknowledges the risk; asserting that `delay` is *not* yet `marked` right after the hover would make the test fail loudly instead of silently going green.

---

Everything else checks out: `delete mouseStatus[target.id]` before the awaited click means a `mousemove` during the routine no longer finds drag state (touch parity holds, since `touchmove` gates on the same map); the `mousemove || touchmove` parenthesisation fix is correct; `flushDelta()` inside the open batch matches the existing usage in `client/js/widgets/widget.js:1318`; and `endDrag()`'s `isLoading` early-out-after-delete is the right call. No validator/`jeCommands` upkeep is needed — no widget or routine property changed.

I added no label: `beta` was already on this PR and stays, which is right — the change only touches `client/js/mousehandling.js` and tests (none of `client/js/widgets/`, `compute.js` or `serverstate.js`), CI and CodeQL are green, and the worst issue above is a post-release animation lag rather than anything that corrupts games or crashes the server.

VTT-AI-Agent:
Thanks — all five are addressed in 9086c538.

**1. Queued drag steps were replayed one by one.** Fixed. The steps stay in the chain, but a step whose position has already been superseded is skipped, so a burst of movements collapses instead of being walked through.

Two details the fix needed beyond the sketch:

* **The last *two* steps are carried out, not just the last one.** `move()`'s hit test reads `center(this.domElement)`, and the transforms only catch up when the *previous* step sent its delta — so a single surviving move hit-tests the widget at its pre-drag position and finds no drop target. With `if(step != ms.steps) return;` the regression test fails with `[ null, null ]`: the card lands on the table instead of in the holder it was released over. `if(step + 1 < ms.steps) return;` keeps exactly the pair the existing two-move `startDrag()` relies on.
* **The handler no longer awaits the step.** `await ms.pending` meant every `mousemove` handler that arrived during the routine was parked on the whole chain and then resumed at once, each doing its `elementsFromPoint()` + `toServer('mouse')` back to back. The step now opens and closes its own batch, so nothing is lost by letting it run on its own; it reports its own failure with `console.error` (which is why `afterDragStep()` no longer needs to swallow anything).

On the test: `startDrag()` takes a number of extra positions now and the leaveRoutine test dispatches 200 of them. I did *not* add a wall-clock bound, because it would not have meant anything here — on this machine the drag settles ~4.1 s after the release without the coalescing and ~3.6 s with it, i.e. the queued steps cost ~2.5 ms each rather than the ~40 ms you measured. Most of those seconds are two `DELAY`s, not the replay: the holder's `leaveRoutine` runs twice per drag out of a holder, once from `moveStart()`'s `set('parent', null)` and again from the `checkParent()` inside the first `move()` that no longer overlaps the holder. A timing assertion tuned to that would be a flaky test that proves nothing on other hardware. The burst still earns its place: it is what caught the naive one-step version above.

**2. `mouseup` on a stale DOM node.** Fixed — `endDrag(dragTarget)` before that `return`, which clears the state (and bails out of `moveEnd()` on its own `widgets.has()` check).

**3. `editInputHandler` swallowing a `mouseup`.** Fixed — the capture of `dragTarget`/`mouseTarget = null` moved above the call, and that path ends the drag too. For the record I could not reach the concrete `spacePanActive` case: `client/js/zoom.js:117` gates the Space keydown on `!Object.keys(mouseStatus).length`, so an active widget drag keeps space-pan from turning on in the first place. The guard belongs there regardless — "button up ⇒ no drag in progress" should not depend on which handler saw the release.

**4. `batchEnd()` skipped on a throw.** Fixed with `try`/`finally`, in `inputHandler()` and in `endDrag()`. It re-indents the whole `if(target && target.id)` body, so `git diff -w` is the friendlier read.

**5. Test 1 passing vacuously.** Fixed — `await t.expect(await markedWidgets()).eql([]);` right after the hover, so an overly slow CI fails loudly instead of asserting nothing.

`npm test` (86), `tests/testcafe/routines.js` (7/7), `editor.js` (14/14) and `publiclibrary-1..4.js` (18/18) pass locally on 9086c538. `beta` was already on the PR and I left it alone.

ArnoldSmith86 (approved):

PR #3066, opened 2026-07-28
Fix pasting formatted HTML into the JSON editor (#2791)

### LINKED ISSUE #2791: Pasting formatted html
#2736 and #2738 made it possible to format html in standard html style.  However, you cannot copy and paste the formatted style from one place or one widget to another.

VTT-AI-Agent:
Fixes #2791.

Since #2736 the JSON editor shows multi-line HTML with real newlines and turns them back into `\n` escapes when the content is parsed (`jePostProcessText`). #2934 extended that to `\r` so Windows clipboards work. But **tabs** — which is how virtually every external editor (and "copy outer HTML" in browser dev tools) indents HTML — were still inserted literally, and a literal tab is not valid inside a JSON string. The result: pasting formatted HTML from anywhere outside VTT immediately breaks the whole widget with

```
SyntaxError: Bad control character in string literal in JSON at position …
```

and nothing you paste can be applied until you hunt down and delete every tab by hand.

## The change

`jePostProcessText` now escapes `\t` and any remaining C0 control character inside JSON strings, exactly like it already did for newlines. Nothing outside a string literal is touched.

`jePreProcessText` does the mirror image on the way back out: it already converted `\n` escapes to real newlines for display, and now does the same for `\t`. Without that half the paste worked but looked broken afterwards — the HTML was correctly indented until you clicked away, and re-selecting the widget redisplayed the very same value with a literal `\t` at the start of every line ([before](https://agent.virtualtabletop.io/reports/pr/1531691877470961916/3066-tabs-before.png) / [after](https://agent.virtualtabletop.io/reports/pr/1531691877470961916/3066-tabs-after.png)). Two small CSS touch-ups go with it: `tab-size: 2` on the editor so a pasted tab lines up with the editor's own 2-space indentation instead of jumping 8 columns, and 5px instead of 2px of padding on the line-number gutter so a line that starts at column 0 doesn't touch its line number.

While in there, only the `\r` of a CRLF is dropped now. A **lone** CR used to be deleted silently (pre-existing behaviour from #2934); it is escaped like every other control character instead, so text with classic-Mac line endings — or a CR inside a JS snippet in an `html` property — survives a paste.

Out of scope: a pasted **double quote** inside a string literal still ends the string and produces a syntax error, so HTML with `class="card"` still has to be pasted with single quotes (or escaped by hand). Auto-escaping quotes isn't safe to do blindly — the editor can't tell "paste this as literal text" from "paste this JSON fragment", and getting it wrong would corrupt content instead of just refusing it. That deserves its own issue and its own design; this PR is about the control characters, which are unambiguous.

## Testing

- TestCafe case in `tests/testcafe/editor.js` puts `<div>\r\n\t<span>tab\vbed</span>\r<hr>\r\n</div>` into a widget's `html` property the way a paste does — tab indentation, CRLF, a lone CR and a vertical tab — and asserts the widget ends up with that exact string (CRLF normalised to `\n`), that the editor reports no JSON error, and that re-selecting the widget shows the tab as a real tab again. Each of the three parts fails on `main` / with the respective hunk reverted.
- `npm test` (86 passed), the full `tests/testcafe/editor.js` suite (15 passed) and `tests/testcafe/compute-4.js` (the other suite that drives `#jeText`, 2 passed) run clean locally.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3066/pr-test (or any other room on that server)

### DIFF
--- a/client/css/jsonedit.css
+++ b/client/css/jsonedit.css
@@ -223,6 +223,8 @@
   right: calc(var(--commandPaneWidth) + 1px);
   font-family: monospace;
   white-space: pre-wrap;
+  /* pasted HTML is usually indented with tabs - render them like the editor's own 2 space indent */
+  tab-size: 2;
   overflow-y: auto;
   overflow-wrap: break-word;
 }
@@ -235,18 +237,19 @@
   /* --linenumbers-digits gets set in jsonedit.js (jeColorize function) depending on the lines of the widget */
   width: calc(var(--linenumbers-digits) * 8px);
   text-align: right;
-  padding-right: 2px;
+  /* keep a gap so lines that start at column 0 don't touch the line number */
+  padding-right: 5px;
   color: var(--textDimColor1);
   vertical-align: top;
 }
 
 .jeLineContent {
   display: inline-block;
-  width: calc(100% - var(--linenumbers-digits) * 8px - 2px);
+  width: calc(100% - var(--linenumbers-digits) * 8px - 5px);
 }
 
 #jeText {
-  padding-left: calc(var(--linenumbers-digits) * 8px + 2px);
+  padding-left: calc(var(--linenumbers-digits) * 8px + 5px);
 }
 #jeText::selection {
   background-color: var(--backgroundHighlightColor1);
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -3351,17 +3351,22 @@ function jePreProcessText(t, returnValidJSON=true) {
   if(returnValidJSON)
     return t;
 
-  // Convert \n escape sequences within JSON strings to actual newlines for display
+  // Convert \n and \t escape sequences within JSON strings to actual newlines and tabs for
+  // display. jePostProcessText turns them back into escapes, so the round trip keeps HTML that
+  // was pasted with tab indentation readable instead of showing a literal \t at the start of
+  // every line.
   let result = '';
   let inString = false;
   let escapeNext = false;
-  
+
   for (let i = 0; i < t.length; i++) {
     const char = t[i];
-    
+
     if (escapeNext) {
       if (inString && char === 'n') {
         result += '\n';
+      } else if (inString && char === 't') {
+        result += '\t';
       } else {
         result += '\\' + char;
       }
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -3257,9 +3257,9 @@ function jePostProcessObject(o) {
 
 function jePostProcessText(t) {
   // Convert actual newlines within JSON strings back to \n escape sequences.
-  // Windows clipboards typically use CRLF, so ignore \r and let the following \n be escaped.
-  // Other control characters (tabs especially, since HTML pasted from an external editor is
-  // usually indented with them) are not valid inside JSON strings either, so escape them too.
+  // Windows clipboards typically use CRLF, so drop the \r of a CRLF and let the following \n be
+  // escaped. Other control characters (tabs especially, since HTML pasted from an external editor
+  // is usually indented with them) are not valid inside JSON strings either, so escape them too.
   let result = '';
   let inString = false;
   let escapeNext = false;
@@ -3285,7 +3285,7 @@ function jePostProcessText(t) {
       continue;
     }
     
-    if (inString && char === '\r')
+    if (inString && char === '\r' && t[i+1] === '\n')
       continue;
     else if (inString && char === '\n')
       result += '\\n';
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -3258,6 +3258,8 @@ function jePostProcessObject(o) {
 function jePostProcessText(t) {
   // Convert actual newlines within JSON strings back to \n escape sequences.
   // Windows clipboards typically use CRLF, so ignore \r and let the following \n be escaped.
+  // Other control characters (tabs especially, since HTML pasted from an external editor is
+  // usually indented with them) are not valid inside JSON strings either, so escape them too.
   let result = '';
   let inString = false;
   let escapeNext = false;
@@ -3287,6 +3289,10 @@ function jePostProcessText(t) {
       continue;
     else if (inString && char === '\n')
       result += '\\n';
+    else if (inString && char === '\t')
+      result += '\\t';
+    else if (inString && char.charCodeAt(0) < 0x20)
+      result += '\\u' + char.charCodeAt(0).toString(16).padStart(4, '0');
     else
       result += char;
   }
--- a/tests/testcafe/editor.js
+++ b/tests/testcafe/editor.js
@@ -342,9 +342,16 @@ test('JSON editor: paste HTML that is indented with tabs', async t => {
     editor.dispatchEvent(new KeyboardEvent('keyup', { bubbles: true, key: 'v' }));
   })();
 
+  const getJSONText = ClientFunction(() => document.querySelector('#jeText').textContent);
+
   await t
     .expect(ClientFunction(() => widgets.get('target').get('html'))()).eql('<div>\n\t<span>tab\vbed</span>\r<hr>\n</div>')
     .expect(Selector('#jeCommands').textContent).notContains('SyntaxError')
+    // selecting the widget again has to show the tab indentation as an actual tab instead of a
+    // literal \t at the start of the line
+    .click('#topSurface', { offsetX: 10, offsetY: 10 })
+    .click('#w_target')
+    .expect(getJSONText()).contains('<div>\n\t<span>tab')
     .click('#editorSidebar [icon=data_object]')
     .pressKey('esc');
 });
--- a/tests/testcafe/editor.js
+++ b/tests/testcafe/editor.js
@@ -320,6 +320,44 @@ test('Deck editor: symbol pickers and JSON fallback', async t => {
   await compareState(t, '5019957515d8552f09fed2340a4e1d3d');
 });
 
+test('JSON editor: paste HTML that is indented with tabs', async t => {
+  await setRoomState({
+    target: { id: 'target', type: 'basic', x: 200, y: 200, html: '<div>old</div>' }
+  });
+  await ClientFunction(prepareClient)();
+  await setName(t);
+
+  await t
+    .click('#editButton')
+    .click('#editorSidebar [icon=data_object]')
+    .click('#w_target');
+
+  // HTML copied out of an external editor arrives with tab indentation (and CRLF on Windows),
+  // neither of which is valid inside a JSON string
+  await ClientFunction(() => {
+    const editor = document.querySelector('#jeText');
+    editor.focus();
+    const content = editor.textContent;
+    const start = content.indexOf('"html": "') + 9;
+    const range = document.createRange();
+    range.setStart(editor.firstChild, start);
+    range.setEnd(editor.firstChild, content.indexOf('"', start));
+    getSelection().removeAllRanges();
+    getSelection().addRange(range);
+
+    const clipboardData = new DataTransfer();
+    clipboardData.setData('text/plain', '<div>\r\n\t<span>tabbed</span>\r\n</div>');
+    editor.dispatchEvent(new ClipboardEvent('paste', { clipboardData, bubbles: true, cancelable: true }));
+    editor.dispatchEvent(new KeyboardEvent('keyup', { bubbles: true, key: 'v' }));
+  })();
+
+  await t
+    .expect(ClientFunction(() => widgets.get('target').get('html'))()).eql('<div>\n\t<span>tabbed</span>\n</div>')
+    .expect(Selector('#jeCommands').textContent).notContains('SyntaxError')
+    .click('#editorSidebar [icon=data_object]')
+    .pressKey('esc');
+});
+
 test('Deck editor: breadcrumb undo and redo', async t => {
   await setRoomState();
   await ClientFunction(prepareClient)();
--- a/tests/testcafe/editor.js
+++ b/tests/testcafe/editor.js
@@ -333,16 +333,17 @@ test('JSON editor: paste HTML that is indented with tabs', async t => {
     .click('#w_target');
 
   // put into the editor what pasting HTML from an external editor puts there: tab indentation
-  // (and CRLF on Windows), neither of which is valid inside a JSON string
+  // (and CRLF on Windows), neither of which is valid inside a JSON string - plus a lone CR and a
+  // vertical tab, which have to survive as escapes instead of being dropped
   await ClientFunction(() => {
     const editor = document.querySelector('#jeText');
     editor.focus();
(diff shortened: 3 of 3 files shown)


VTT-AI-Agent (commented):
## TL;DR

Correct, tightly scoped fix — I verified end-to-end that a real `Ctrl+V` of tab-indented HTML now applies cleanly, and that the new TestCafe case genuinely fails on `main` (it only passes with the `jsonedit.js` change). One remaining gap: a **lone CR** (not followed by LF) is still silently *deleted* rather than escaped, so the new comment's claim that all remaining control characters are escaped isn't quite true.

## Findings

### 1. `client/js/jsonedit.js:3287` — lone CR is silently dropped, not escaped

```js
if (inString && char === '\r')
  continue;
```

The new comment says "Other control characters ... are not valid inside JSON strings either, so escape them too", but CR is handled *before* the new `< 0x20` branch and is discarded. For CRLF that's correct (the following LF becomes an escaped `\n`). For a **lone** CR it's silent data loss: content copied from a classic-Mac-line-ending file, or a string that genuinely contains a CR (e.g. an HTTP-ish literal, or a CR inside a JS snippet in an `html` property), loses the character with no error and no way to notice.

Verified by running `jePostProcessText` directly (control chars written as `<CR>`/`<LF>`/`<TAB>`/`<BS>`):

| input | output | parsed value |
|---|---|---|
| `{"html":"a<CR><LF>b"}` | `{"html":"a\nb"}` | `a` + LF + `b` — correct |
| `{"html":"a<CR>b"}` | `{"html":"ab"}` | `ab` — **CR gone** |
| `{"html":"a<TAB>b"}` | `{"html":"a\tb"}` | `a` + TAB + `b` — correct (this PR) |
| `{"html":"a<BS>b"}` | escaped as a `u0008` unicode escape | `a` + BS + `b` — correct (this PR) |

Since this PR is specifically about "everything you paste should survive", closing the last gap is a one-line change — only swallow the CR when it is part of a CRLF:

```js
if (inString && char === '\r' && t[i+1] === '\n')
  continue;
```

and let the new `< 0x20` branch turn a standalone CR into its `u000d` unicode escape. (Or emit an explicit `\r` escape.) Not a blocker — it is pre-existing behaviour from #2934 — but this PR touches exactly that comment and branch, so it is the natural place.

### 2. `tests/testcafe/editor.js:336-341` — the test simulates a paste rather than performing one

```js
editor.textContent = editor.textContent.replace(...);
editor.dispatchEvent(new KeyboardEvent('keyup', { bubbles: true, key: 'v' }));
```

This bypasses the real `on('#jeText', 'paste', ...)` handler and `jePasteText()` (selection offsets, `jeSetEditorContent`'s U+00A0 normalisation, `jeColorize`), so a future regression *in the paste path itself* — the thing issue #2791 is actually about — would not be caught; only `jePostProcessText` is covered.

That said, I confirmed with Playwright that the real path is fine: writing tab-indented, CRLF-separated HTML to the system clipboard, selecting `<div>old</div>` in `#jeText` and pressing `Control+V` ends with the widget's `html` equal to the expected `<div>` + LF + TAB + `<span>tabbed</span>` + LF + `</div>` and no `SyntaxError` in `#jeCommands`. So this is a test-fidelity suggestion, not a correctness problem. Driving it via a real `ClipboardEvent` with a `DataTransfer` would exercise the handler for the same cost.

### 3. Minor — the `\u00XX` branch is untested

The test only exercises TAB. The new `char.charCodeAt(0) < 0x20` branch (the one that could produce malformed output if the `padStart` were wrong) has no coverage. A second widget property containing e.g. a vertical tab (U+000B) in the same test case would cover it for free.

### 4. Minor — PR description nit

> the value is stored as a real tab and displayed as `\t` in the editor

Only after the widget is re-selected. `jePreProcessText(..., false)` converts *only* `\n` escapes back to real characters, so immediately after the paste the editor still holds the literal TAB (and literal CR) in `#jeText` — I saw exactly that in the browser. The round-trip is still stable, so this is purely a description-accuracy point.

## Verification I ran locally

- `npx testcafe chromium:headless tests/testcafe/editor.js` → **15 passed** on the branch.
- Reverting only `client/js/jsonedit.js` to `origin/main` and restarting the server → the new case **fails** (`expected '<div>old</div>'`), confirming it is a real regression test.
- Real `Ctrl+V` paste through Playwright → correct value, no `SyntaxError` (see finding 2).
- Direct probe of `jePostProcessText` over TAB / CRLF / lone-CR / BS / backslash-then-control-char / TAB-outside-string inputs — all produce valid JSON except the lone-CR data loss above; tabs *outside* strings are correctly left alone and parse fine.

No validator or `jeCommands` upkeep is needed (no widget or operation property added), and nothing here can change existing room state.

I added the `beta` label: the PR touches only `client/js/jsonedit.js` (the edit-mode JSON editor) and a test — no engine file (`client/js/widgets/`, `compute.js`, `serverstate.js`) is involved, so it is not beta-excluded, and the change only makes previously-unparseable editor text parse, which cannot corrupt saved games or crash the server.

VTT-AI-Agent:
Thanks — all four points addressed or answered. Pushed `eb2d33d`.

**1. Lone CR silently dropped — fixed.** Took your one-liner:

```js
if (inString && char === '\r' && t[i+1] === '\n')
  continue;
```

so only the `\r` of a CRLF is swallowed and a standalone CR falls through to the new `< 0x20` branch, where it becomes a \u000d escape. Comment updated to match ("drop the `\r` of a CRLF").

**3. `\u00XX` branch untested — fixed, together with the lone CR.** The test now pastes `<div>\r\n\t<span>tab\vbed</span>\r<hr>\r\n</div>` and expects `<div>\n\t<span>tab\vbed</span>\r<hr>\n</div>`, so one case covers tab indentation, CRLF, a lone CR and a vertical tab (U+000B, through the `padStart` branch). Verified it is a real regression test for the CR part too: with only the `&& t[i+1] === '\n'` reverted, the case fails; with it, `tests/testcafe/editor.js` is 15/15 green (plus `npm test`, 86 passed).

**2. Test simulates a paste instead of performing one — deliberate, and I would rather leave it.** The first version of this test did exactly what you suggest: a real `ClipboardEvent` with a `DataTransfer` payload. It works in Chromium, but **Firefox ignores the `clipboardData` passed to the `ClipboardEvent` constructor**, so the synthetic paste arrived empty and the case failed there (see f87f03c — CI runs the suite on both browsers). Since the regression this test guards is in `jePostProcessText` and not in the selection arithmetic, I went for the portable version. Agreed that it leaves the paste handler itself uncovered — but I have no cross-browser way to drive a genuine paste from TestCafe without OS-level clipboard access, and your Playwright run already confirms the real path works.

**4. Description nit — fixed.** It now says the `\t` is only shown again once the widget is re-selected, and it mentions the lone-CR change.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review) of the *visible* behaviour of this fix in the JSON editor. The fix does what it promises — tab-indented HTML now pastes without blowing up the widget — but the editor ends up showing **two different representations of the same string**: right after the paste the indentation is real tabs, and after re-selecting the widget the very same value is redisplayed as literal `\t` / `\t\t` at column 0. Making `jePreProcessText` turn `\t` back into a real tab (mirroring what it already does for `\n`) would make the pasted HTML *stay* readable, which is the single most impactful improvement here.

Reviewed by checking out the branch and driving the real editor in Chromium against a local server (basic widget with an `html` property, real `paste` event carrying tab-indented CRLF HTML). The diff itself touches no HTML/CSS/template, so everything below is about what the change makes the editor *look like*.

---

## Big layout improvements

**1. The same string is displayed two different ways before and after it is stored.**
`jePostProcessText` now escapes tabs on the way in, but `jePreProcessText` (client/js/jsonedit.js:3349) still only converts `\n` back to a real newline on the way out. Net effect for the user:

Immediately after pasting — indentation is real, the HTML is readable:
after paste

Re-select the same widget (or reload) — identical value, but lines 14/15 now start with a literal `\t` / `\t\t` flush against the line-number gutter:
after reload

So the fix rescues the paste but the structure the user pasted is only visible until they click away — afterwards the block reads worse than plain unindented HTML, because every line now carries two extra escape characters *and* has lost its indentation. Since `jePostProcessText` already round-trips `\t` reliably, the symmetric change in `jePreProcessText` (`if (inString && char === 't') result += '\t';`) would make display stable and match the newline behaviour introduced in #2736. Worth doing in this PR — it's the half that makes the fix feel finished.

**2. Pasting the *most common* form of external HTML still breaks the widget.**
HTML copied from dev tools or an external editor is almost always indented with tabs **and** uses double-quoted attributes (`class="card"`). The tabs are handled now; the quotes still terminate the JSON string, so the paste fails exactly as in #2791 — only with a different message:

double-quoted HTML still errors

> `SyntaxError: Expected ',' or '}' after property value in JSON at position 166 (line 13 column 24)`

From a user's point of view the reported symptom ("pasting formatted HTML breaks the widget") is only half fixed, and the failure now looks arbitrary: the same paste works with `class='a'` and explodes with `class="a"`. Escaping quotes automatically isn't safe in the general case, but the *paste* path (`on('#jeText','paste')`, client/js/jsonedit.js:4073) does know the pasted text is a single literal insertion — if the caret is inside a string literal it could escape `"` the same way it now escapes `\t`. If that's out of scope for this PR, please say so in the description/issue, otherwise #2791 will read as fixed while the common case still fails.

- **Escaped lines collide with the line-number gutter.** `.jeLineNumber` (client/css/jsonedit.css:236) has only `padding-right: 2px`, so a continuation line that starts at column 0 renders as `14\t<span …>` with no visible separation between the number and the content (second screenshot, lines 14–16). This existed before, but this PR makes column-0 continuation lines much more common; 4–6px of padding (or a right border on the gutter) would fix it cheaply.
- **Tab width doesn't match the editor's own indentation.** While the real tabs are present (first screenshot, lines 14/15) they render at the browser default 8-character stop, while the surrounding JSON is indented 2 spaces — one pasted `\t\t` pushes content ~16 characters right and forces wrapping in the fairly narrow text pane. `tab-size: 2` on `#jeText, #jeTextHighlight` (client/css/jsonedit.css:218) would keep pasted HTML aligned with the app's style. This becomes essential if you adopt item 1 above.
- **Non-tab control characters land as `\u000b`-style escapes.** Correct, but visually indistinguishable from ordinary text at a glance and not colourised as an escape by `jeColorize`. Low priority — just noting that a stray vertical tab in a pasted snippet will silently show up as `\u000b` in the middle of an HTML line with no hint of where it came from.

- **The error text is raw engine output.** In the failing case the command pane shows *"basic ↦ html Ctrl-Space: go to error `SyntaxError: Expected ',' or '}' after property value in JSON at position 166 (line 13 column 24)`"* — the keyboard hint and the error run together in one paragraph, and `position 166` is meaningless to a newcomer (the `line 13 column 24` part is the only actionable bit, and it's last). Since this PR is specifically about making paste survivable, a paste-aware message would pay off a lot here, e.g. *"Pasted text can't be used as-is: unescaped `"` on line 13. Replace it with `\"` or use single quotes."* At minimum, put the error on its own line above the `Ctrl-Space: go to error` hint so the hint isn't buried mid-sentence.
- **A successful paste gives no feedback that it was rewritten.** Tabs/CRLF/CR are silently transformed. That is the right default, but because the editor then *re-renders* them differently later (item 1), a first-time user can reasonably conclude the editor ate their formatting. Fixing the display asymmetry removes the need for any message; if that isn't done, a one-line note in the command pane ("tabs stored as `\t`") would prevent the confusion.
- **Wording nit for the code comment / any future docs:** "drop the `\r` of a CRLF" is accurate but the user-visible rule is simpler — *"Windows line endings are normalised to `\n`; a lone CR is kept."* Worth stating that way if this ever surfaces in the wiki's Edit-Mode page, since the lone-CR-vs-CRLF distinction is now user-observable.

Nothing here blocks the fix from a UI standpoint — the editor is visibly better off with it than without. Item 1 is the one I'd fold in before merging.

VTT-AI-Agent:
Thanks — the display asymmetry was the right thing to catch. Pushed `1cec317`.

**1. Same string displayed two different ways — fixed.** `jePreProcessText` now unescapes `\t` alongside `\n`, exactly as suggested, so what you see right after the paste is what you see when you come back to the widget:

| before | after |
|---|---|
| before | after |

The round trip is symmetric (`jePostProcessText` escapes a real tab inside a string, `jePreProcessText` unescapes it), and a literal `\\t` in the JSON still survives both directions unchanged. The existing paste test now also asserts the redisplay: after the paste it deselects and re-selects the widget and expects `#jeText` to contain a real tab. With only that hunk reverted the case fails (`to include '<div>\n\t<span>tab'`), so it guards the half you found.

**Tab width — fixed.** `tab-size: 2` on `#jeText, #jeTextHighlight`. Needed, as you said, now that item 1 is in: pasted `\t\t` no longer pushes content 16 columns right.

**Gutter collision — fixed.** `.jeLineNumber` padding-right 2px → 5px, with the two coupled values (`.jeLineContent` width, `#jeText` padding-left) moved along so the text and the highlight overlay stay aligned to the pixel — verified in Chromium that `#jeText`'s content box and `.jeLineContent` still start at the same x.

**2. Double-quoted attributes still break the paste — real, but deliberately out of scope.** I agree it's the more common form, and the PR description now says so explicitly instead of letting #2791 read as fully fixed. I don't think it can be fixed safely in the paste path though: escaping `"` when the caret is inside a string assumes the pasted text is meant as literal content, but pasting a JSON fragment into a string value is just as common in this editor, and there the same rewrite silently corrupts it. The same ambiguity applies to a pasted `\` (leave it, escape it, or treat it as an existing escape?). Control characters have no such ambiguity — a raw tab inside a JSON string is *always* an error and there is exactly one thing the user can have meant — which is why this PR stops there. The quote case wants its own issue and probably a different mechanism (an explicit "paste as string content" action, or offering the fix in the error message) rather than silently rewriting quotes.

**Error text is engine-raw / buried mid-sentence** — noted, but I'd rather not touch it here: `jeJSONerror` is the generic JSON error path for everything typed in the editor, not just pastes, so rewording it is a separate UI concern with its own blast radius. FWIW it *is* emitted on its own line already (`\n<i class=error>…`); what runs it together in your screenshot is wrapping in the narrow command pane. Worth its own issue.

**Non-tab control chars showing as ``-style escapes with no colourisation** and **"a successful paste gives no feedback"** — both are, I think, resolved enough by item 1: the formatting now stays visible, so there is nothing for the user to conclude was eaten. Those escapes stay rare and, unlike before, are no longer surrounded by literal `\t`s on every line.

**Comment wording** — kept "drop the `\r` of a CRLF" in the code (it describes the branch it sits on), but the PR description now states the user-visible rule the way you phrased it: Windows line endings normalise to `\n`, a lone CR is kept.

Re-ran locally: `npm test` 86 passed, `tests/testcafe/editor.js` 15 passed, `tests/testcafe/compute-4.js` (the other suite that drives `#jeText`) 2 passed.

96LawDawg:
I can no longer replicate the problem even on production. Closing.

PR #3067, opened 2026-07-28
Add a Europe map to All Aboard, selectable in the Reset button

VTT-AI-Agent:
Builds on #3064 (based on its branch, so the diff here is only the Europe map).

## What this does

**Reset** now asks which map to play on:

the reset dialog

Picking one shows that map's cities, routes and train car holders, hides the other one's, and swaps the destination tickets in the draw pile — 30 for America, 46 for Europe. The choice is remembered as the default for the next reset, so a group that plays Europe never sees the American map again after the first game. Nothing else about the game changes: the same train car deck, the same seats, the same scoring.

**Europe** is the 47 cities and 101 routes of the Europe edition — 300 train cars, 8 colours × 8 routes plus 37 grey ones, 11 double routes, and the 40 short + 6 long tickets:

the Europe map

## How it is built

Same shape as the American map after #3064, so the two are maintained the same way:

* One **line widget per route**, connected to the two city cards it runs between, with that route's holders as its evenly spaced stops. Moving a city in edit mode drags its routes along, and the `Route Template` / `Holder Template` widgets are shared with the American map.
* The two lines of a **double route** share one centre path with matching ±8 connection offsets, which `applyConnections` turns into a parallel pair.
* **Palermo–Smyrna** is a cubic Bézier that passes south of Greece the way it does on the board; every other route is straight, and no two routes cross.
* Europe has two kinds of route America does not: a **tunnel** is drawn with a dark outline, a **ferry** with a striped one. Like the colours of the routes, the game doesn't enforce their rules, so there is a two line key in the empty north Atlantic corner and a paragraph in the game's help text.

**City positions** start from the real coordinates of the 47 cities, projected onto the board (Mercator in latitude, stretched in longitude so the map covers the board exactly). From there a relaxation pass spreads them just enough that every route has room for its cars, that no route runs under the game's own interface — the ticket pile and its buttons top right, the waggon draw row along the bottom, the ticket hand on the right — and that no foreign city dot sits on a route. Half the cities end up within 23px of where they really are; the largest move is Erzurum, which the ticket hand pushes 94px west. City names flip to the left of their dot where there is no room for them on the right (a second face on the `Cities` deck, so the American cities are untouched).

The train cars are 33px instead of America's 45: Europe fits the same number of cars into the smaller free area of the board.

**The background map** is generated from [Natural Earth](https://www.naturalearthdata.com/) vector data, which is in the public domain — land silhouette plus country outlines, in the same `#d0d0d0` on `#fff` as the existing American map, and covering the 1400×867 board exactly. Because the city dots come out of the same projection, every dot sits on its own place, and so do the two dots on a **ticket**, which is the board scaled down by 10:

Europe tickets

Both maps' tickets share the `Tasks` deck; the mini map is now a `dynamicProperties` value so each ticket carries its own.

## Testing

* `npm test` — 86 passed
* `node validator/validate_gamefile_node.js "library/games/All Aboard/0.json"` — no new findings (the three pre-existing deck warnings are unchanged)
* In the browser, on a local server: reset to Europe and back to America several times — the maps swap cleanly, the draw pile goes 30 ↔ 46 tickets, the unused map's tickets park in a hidden holder and come back on the next reset, and no console errors.
* Sat down, reset to Europe and claimed Pamplona–Marseille by dropping one car on it: the other three cars filled in, the waggon count went 45 → 41 and the score went 0 → 7, which is what a four car route pays.
* Checked that no train car and no city name ends up under the interface, and that the holder positions in the file are exactly the ones the line widgets compute, so loading the game moves nothing.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3067/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR
Metadata is essentially complete and consistent with the rest of the library — the attribution block correctly adds a credit line for the new Europe background map (Natural Earth, public domain), and `helpText` is updated to explain the new map-choice step and the tunnel/ferry rules. The only thing worth a look is a spelling-convention slip: the new help text uses British "colours" while the rest of the file (and `language: en-US`) uses American "color(ed)".

*Automated review 1 of 4 (metadata) — by `sonnet`*

## Findings

1. **Minor spelling inconsistency (British vs. American English).** The new sentence added to `helpText` — "Just like the **colours** of the routes, the game does not enforce their rules…" — uses British spelling, while the existing text in the same field uses American spelling ("Trains are automatically sorted by **color**.", "colored holders"), and the game's declared `language` is `en-US`. Worth a quick tweak to `colors` for consistency, though it's a small thing.

2. **Attribution is handled well.** The PR appends a new credit line to the existing `attribution` block for the Europe map source: *"The map of Europe was generated from Natural Earth vector data, which is in the public domain and available from https://www.naturalearthdata.com/."* This follows the same `<div>` list format as the existing credits and correctly cites a public-domain source — nice attention to detail here.

3. **Other metadata fields (name, image, bgg link, players, time, description, year, similar-game info) are unchanged**, which makes sense since this PR adds a selectable map to an existing game rather than introducing a new one — no updates were needed there.

Everything else checks out: the metadata stays internally consistent with the pre-existing format for this game and with other library entries. Nice work extending the attribution rather than leaving the new source uncredited.

PR #3068, opened 2026-07-29
Fix client crash when enabling Debug panel during an INPUT prompt

### LINKED ISSUE #2672: Client crash with INPUT and Edit Mode

VTT-AI-Agent:
Fixes #2672

### Problem
`evaluateRoutine` checks the *live* `jeRoutineLogging` global at every logging call site. A routine that suspends on an `INPUT` modal can have logging switched on mid-flight (open Edit mode → click a confirm button → choose *Debug* → press OK). The operation's `jeLoggingRoutineOperationStart` never ran, but `jeLoggingRoutineOperationEnd` did, so `jeHTMLStack.shift()` returned `undefined` and `savedHTML[1]` threw `Cannot read properties of undefined (reading '1')`. The mirror case (logging switched *off* mid-routine) leaked entries on the stack.

This is not only a console error: the exception aborts the rest of the routine, so the game is left in a half-applied state.

### Fix
- Snapshot `jeRoutineLogging` into a local `routineLogging` const at the start of `evaluateRoutine` and use it for every logging call in that invocation, so start/end calls always pair up.
- Defensive guards in `jsonedit.js`: `jeLoggingRoutineOperationEnd` returns early on an empty stack, `jeLoggingRoutineEnd` returns early at depth 0.

### Consequence, and how it is surfaced
A routine that was *already running* when logging got enabled cannot be logged retroactively — the whole invocation stays unlogged. Previously that produced a partial (crashing) log; now the Debug panel would just stay empty with no hint why, which reads as a second bug. So such a routine now writes a note to the log, naming the widget and the routine and saying what to do about it:

Debug panel showing the note

Since that note is usually the *only* thing in the panel it is rendered as a callout (theme variables, so it works in dark mode too) rather than in the panel's dimmed metadata style.

### Also fixed
- Clearing the log (the *Clear* button, or any incoming full state while the Debug module is open) only emptied `jeLoggingHTML` and not the log saved on `jeHTMLStack`, so the next operation end prepended it again and resurrected what had just been cleared. Clearing now empties both.
- The DOM-dependent parts of rendering the log (expander click handlers, red expanders, the active filter) moved from `jeLoggingRoutineEnd` into `jeLoggingRenderLog()`, so every render re-attaches them — writing the note no longer kills the expanders of the entries already in the log.
- `#jeLog` still carried `margin-left: -40px` from when the log was positioned inside `#jsonEditor`; since the log became its own sidebar module that made it stick out to the left of the panel, clipping the top-level entries' ▶ arrows. Removed.

### Test
New TestCafe test in `tests/testcafe/editor.js` (the first Debug-panel coverage there): it starts a `LABEL` / `INPUT` / `LABEL` routine in edit mode, opens the Debug module while the INPUT overlay is up, presses *Go* and then checks that the label reached `done`, that no page error fired, that the log carries the note, and that the room state hash matches. On `main` it fails with the exact `Cannot read properties of undefined (reading '1')` from #2672.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3068/pr-test (or any other room on that server)

### DIFF
--- a/client/css/jsonedit.css
+++ b/client/css/jsonedit.css
@@ -534,10 +534,6 @@ body.jsonEdit #toolbar {
   white-space: normal;
 }
 
-#jeLog {
-  margin-left: -40px;
-}
-
 @media (max-width: 1000px) {
   #jeLog {
     right: var(--commandPaneWidth);
@@ -574,8 +570,22 @@ body.jsonEdit #toolbar {
   margin-top: 20px;
 }
 
-/* has no expander arrow, so it needs the indentation the arrow would otherwise provide */
-.jeLogNote    { color: var(--textDimColor1); font-style: italic; margin-left: 38px; }
+/* a message about the log itself rather than a logged routine, so make it a readable callout */
+.jeLogNote {
+  color: var(--textColor);
+  background-color: var(--backgroundHighlightColor1);
+  border-left: 3px solid var(--textHighlightColor2);
+  border-radius: 4px;
+  padding: 8px 10px;
+  margin: 10px 0 10px 20px;
+  font-family: Arial, sans-serif;
+}
+.jeLogNote .jeLogWidget,
+.jeLogNote .jeLogProperty {
+  font-family: monospace;
+  font-weight: bold;
+}
+
 .jeLogSkipped { color: var(--textDimColor1); }
 .jeLogSummary { color: var(--textHighlightColor4); }
 .jeLogResult  { color: var(--textHighlightColor2); }
--- a/client/css/jsonedit.css
+++ b/client/css/jsonedit.css
@@ -574,6 +574,8 @@ body.jsonEdit #toolbar {
   margin-top: 20px;
 }
 
+/* has no expander arrow, so it needs the indentation the arrow would otherwise provide */
+.jeLogNote    { color: var(--textDimColor1); font-style: italic; margin-left: 38px; }
 .jeLogSkipped { color: var(--textDimColor1); }
 .jeLogSummary { color: var(--textHighlightColor4); }
 .jeLogResult  { color: var(--textHighlightColor2); }
--- a/client/js/editor/sidebar/json.js
+++ b/client/js/editor/sidebar/json.js
@@ -89,7 +89,7 @@ class DebugModule extends SidebarModule {
   }
 
   button_clearButton() {
-    jeLoggingHTML = '';
+    jeLoggingClear();
     $('#jeLog').innerHTML = '';
   }
 
@@ -98,7 +98,7 @@ class DebugModule extends SidebarModule {
 
     $('#clearLogButton').disabled = $('#autoClearLog').checked;
     if($('#clearLogButton').disabled)
-      jeLoggingHTML = '';
+      jeLoggingClear();
   }
 
   button_filter() {
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -3085,38 +3085,44 @@ export function jeLoggingRoutineEnd(variables, collections) {
     return; // defensive: unmatched End, should not happen since #2672
   if( jeHTMLStack.length == 0 || ['CALL', 'CLICK', 'IF', 'loopRoutine', 'Moves'].indexOf( jeHTMLStack[0][3] ) == -1 ) jeLoggingHTML += '</div></div>';
   --jeLoggingDepth;
-  if(!jeLoggingDepth) {
-    $('#jeLog').innerHTML = jeLoggingHTML + '</div></div>';
+  if(!jeLoggingDepth)
+    jeLoggingRenderLog(jeLoggingHTML + '</div></div>');
+}
 
-    // Make it so clicking on the arrows expands the subtree
-    const expanders = document.getElementsByClassName('jeExpander');
-    let i;
-    for (i=0; i < expanders.length; i++) {
-      expanders[i].addEventListener('click', function() {
-        this.classList.toggle('jeExpander-down');
-        this.parentNode.querySelector('.jeLogNested').classList.toggle('active');
-        if(this.classList.contains('jeExpander-down')) {
-          this.classList.add('manuallyExpanded');
-          this.parentNode.querySelector('.jeLogNested').classList.add('manuallyExpanded');
-        } else {
-          this.classList.remove('manuallyExpanded');
-          this.parentNode.querySelector('.jeLogNested').classList.remove('manuallyExpanded');
-        }
-      });
-    }
-    // Make expander arrows that are parents of nodes with problems show up red.
-    const problems = document.getElementsByClassName('jeLogHasProblems');
-    for (i=0; i<problems.length; i++) {
-      let node = problems[i];
-      while (node && node.id != 'jeLog') {
-        if(node.classList.contains('jeLogOperation') || node.classList.contains('jeLog')) {
-          node.firstElementChild.classList.remove('jeExpander');
-          node.firstElementChild.classList.add('jeRedExpander')
-        }
-        node = node.parentNode;
+// Put the log into the panel. Everything that depends on the rendered DOM (the expander click
+// handlers and the filter) has to be applied again afterwards, so all rendering goes through here.
+function jeLoggingRenderLog(logHTML) {
+  $('#jeLog').innerHTML = logHTML;
+
+  // Make it so clicking on the arrows expands the subtree
+  const expanders = document.getElementsByClassName('jeExpander');
+  let i;
+  for (i=0; i < expanders.length; i++) {
+    expanders[i].addEventListener('click', function() {
+      this.classList.toggle('jeExpander-down');
+      this.parentNode.querySelector('.jeLogNested').classList.toggle('active');
+      if(this.classList.contains('jeExpander-down')) {
+        this.classList.add('manuallyExpanded');
+        this.parentNode.querySelector('.jeLogNested').classList.add('manuallyExpanded');
+      } else {
+        this.classList.remove('manuallyExpanded');
+        this.parentNode.querySelector('.jeLogNested').classList.remove('manuallyExpanded');
       }
+    });
+  }
+  // Make expander arrows that are parents of nodes with problems show up red.
+  const problems = document.getElementsByClassName('jeLogHasProblems');
+  for (i=0; i<problems.length; i++) {
+    let node = problems[i];
+    while (node && node.id != 'jeLog') {
+      if(node.classList.contains('jeLogOperation') || node.classList.contains('jeLog')) {
+        node.firstElementChild.classList.remove('jeExpander');
+        node.firstElementChild.classList.add('jeRedExpander')
+      }
+      node = node.parentNode;
     }
   }
+
   if($('#jeLogFilter') && $('#jeLogFilter').value)
     jeLoggingFilterLog($('#jeLogFilter').value);
 }
@@ -3131,14 +3137,15 @@ export function jeLoggingRoutineNotLogged(widget, property) {
     jeLoggingHTML = '';
     jeRoutineResetOnNextLog = false;
   }
+  const routine = typeof property == 'string'
+    ? `<span class="jeLogWidget">${html(widget.get('id'))}</span> <span class="jeLogProperty">${html(property)}</span>`
+    : `an inline routine of <span class="jeLogWidget">${html(widget.get('id'))}</span>`;
   jeLoggingHTML += `
     <div class="jeLog jeLogNote">
-      <span class="jeLogWidget">${html(widget.get('id'))}</span>
-      <span class="jeLogProperty">${html(typeof property == 'string' ? property : '--custom--')}</span>
-      was not logged because logging was enabled after it had already started.
+      ${routine} was already running when the Debug panel was opened, so it could not be recorded. Run it again to see its log.
     </div>
   `;
-  $('#jeLog').innerHTML = jeLoggingHTML;
+  jeLoggingRenderLog(jeLoggingHTML);
 }
 
 export function jeLoggingRoutineOperationStart(original, applied) {
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -3072,6 +3072,8 @@ export function jeLoggingRoutineStart(widget, property, initialVariables, initia
 }
 
 export function jeLoggingRoutineEnd(variables, collections) {
+  if(!jeLoggingDepth)
+    return; // no matching jeLoggingRoutineStart (logging was enabled mid-routine)
   if( jeHTMLStack.length == 0 || ['CALL', 'CLICK', 'IF', 'loopRoutine', 'Moves'].indexOf( jeHTMLStack[0][3] ) == -1 ) jeLoggingHTML += '</div></div>';
   --jeLoggingDepth;
   if(!jeLoggingDepth) {
@@ -3131,6 +3133,11 @@ export function jeLoggingRoutineOperationEnd(problems, variables, collections, s
     collDisplay[name] = collections[name].map(w=>`${html(w.get('id'))} (${html(w.get('type')||'basic')})`);
 
   const savedHTML = jeHTMLStack.shift();
+  if(!savedHTML) {
+    // No matching jeLoggingRoutineOperationStart (logging was enabled mid-routine). Nothing to close.
+    jeRoutineResult = '';
+    return;
+  }
   const original = savedHTML[1];
   const originalText = jeLoggingJSON(original);
   const applied = savedHTML[2];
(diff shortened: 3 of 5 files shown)


VTT-AI-Agent (commented):
## TL;DR
The fix is correct and I reproduced it both ways: on `main` the exact `Cannot read properties of undefined (reading '1')` from #2672 fires and the routine **aborts mid-flight** (my test label stayed `"start"`); on this branch there is no error and the routine completes (`"done"`). Snapshotting `jeRoutineLogging` per `evaluateRoutine` invocation is the right shape — my only substantive notes are a silent-UX consequence and the missing regression test.

### Verification
Room with `clickRoutine: [VAR, INPUT, VAR, LABEL(lbl='done')]`; enter Edit mode with Debug **closed**, trigger the routine so it suspends on the INPUT overlay, open the Debug module, press *Go*.

| | `main` (6a7b422) | this branch |
|---|---|---|
| page error | `Cannot read properties of undefined (reading '1')` | none |
| `lbl.text` after | `"start"` (routine aborted after INPUT) | `"done"` |

So this isn't only a console error — the crash silently swallows the remainder of the routine, which is a real half-applied-game-state hazard. Worth calling out in the PR body. `npm test` passes (86/86).

### Findings

1. **`client/js/widgets/widget.js:965` — behaviour change worth documenting: enabling Debug mid-routine now produces *no* log at all for that invocation.** Previously it produced a partial (crashing) log; now the captured `routineLogging === false` suppresses every call site for the rest of the invocation. The Debug module advertises itself as *"View debug information for the most recent routine execution"*, so a user who opens Debug precisely because a routine is stuck on an INPUT prompt and then presses OK will see an empty `#jeLog` with no hint why. That's the correct trade-off vs. crashing, but the PR description doesn't mention it and nothing surfaces it in the UI. Note that *nested* routines entered after the toggle do get logged (they capture `true` and, with an empty `jeHTMLStack`, render as top-level `.jeLog` entries) — so the panel can show an orphaned inner routine with no parent, which reads as a bug to a user. Either say so in the PR body or consider a one-line hint in the log pane.

2. **`client/js/jsonedit.js:3075` and `:3136` — the two guards are now unreachable via the path their comments describe.** With the per-invocation capture, `jeLoggingRoutineStart`/`End` and `OperationStart`/`OperationEnd` are always both-or-neither within one `evaluateRoutine` frame, so "logging was enabled mid-routine" can no longer produce an unmatched `End`. Keeping them as belt-and-braces is fine, but the comments will mislead the next reader into thinking that's still the live cause. Suggest re-wording to something like `// defensive: unmatched End (should not happen since #2672)`.

3. **No regression test.** This one is genuinely testable in the TestCafe state-hash style, because the bug has an observable *room-state* signature (see table above): set a room with `INPUT` followed by a state-mutating op, enter edit mode, trigger the routine, click the Debug module (`[icon=pest_control]`), click `#buttonInputGo`, then `compareState`. On `main` that hash differs from the fixed one, so it would have caught this. `tests/testcafe/editor.js` has no Debug-panel coverage at all today, so this would also be the first.

4. **Adjacent, pre-existing, out of scope — `client/js/editor/sidebar/json.js:91-102` and `:138`.** `button_clearButton()` (and `onStateReceivedWhileActive`, which fires on *every* incoming delta) resets `jeLoggingHTML` without touching `jeHTMLStack`/`jeLoggingDepth`. If that fires while a routine is suspended on an INPUT with logging on, `jeLoggingRoutineOperationEnd` re-prepends `savedHTML[0]`, resurrecting the content that was just cleared. Cosmetic only (no crash), but it's the same class of "global logging state mutated mid-routine" bug this PR is fixing — a natural follow-up.

I checked every one of the ~60 renamed sites in the diff: all are pure `jeRoutineLogging` → `routineLogging` inside `if(...)` conditions (plus the `!routineLogging && problems.length` console.log), with no incidental semantic edit. No routine-operation behaviour changes.

I added the `beta` label: the only engine-file edit (`client/js/widgets/widget.js`) is incidental — every changed line is a debug-logging conditional that is only ever true when Edit mode's Debug panel is open, so for players with logging off the captured constant is `false` exactly as the live global was, leaving existing games' behaviour byte-for-byte unchanged, and I verified the fix works end-to-end in a real browser.

VTT-AI-Agent:
Thanks — all four findings addressed in b26ac835.

**1. Enabling Debug mid-routine yields no log.** Agreed, and rather than only documenting it I made the panel say so: a routine that was already running when logging got enabled now writes a one-line note to the log (`jeLoggingRoutineNotLogged` in `jsonedit.js`, called from `evaluateRoutine` when `!routineLogging && jeRoutineLogging`). It names the widget and the routine, e.g. *"button clickRoutine was not logged because logging was enabled after it had already started."* — [screenshot](https://agent.virtualtabletop.io/reports/pr/1531843930839318788/pr3068-debug-note.png). That also explains the orphaned nested entries you spotted: they now appear under a note saying their parent could not be logged. The PR description covers this too.

**2. Guard comments.** Reworded both to `// defensive: unmatched End, should not happen since #2672`.

**3. Regression test.** Added `Enabling the Debug module while a routine waits for INPUT does not abort the routine` to `tests/testcafe/editor.js` — it does exactly what you described (edit mode with Debug closed → routine suspends on INPUT → click `[icon=pest_control]` → `#buttonInputGo`) and asserts the label reached `done`, no page error fired, the log note is present, and `compareState`. Verified it fails on `main`'s client with `TypeError: Cannot read properties of undefined (reading '1') at jeLoggingRoutineOperationEnd`, and passes here. Full `editor.js` suite: 15/15, `npm test`: 86/86.

**4. `button_clearButton()` log resurrection.** It was small enough and squarely the same class of bug, so I fixed it here instead of deferring: clearing now goes through `jeLoggingClear()`, which also empties the log saved on each `jeHTMLStack` entry, so a clear that happens mid-routine is no longer undone by the next `jeLoggingRoutineOperationEnd`. `jeToggle()` and `button_clearCheckbox()` use it too.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review) of the one thing this PR puts on screen: the new "not logged" note in the **Debug** panel. The fix itself does what it says — I reproduced the flow in a browser and the routine now completes instead of aborting. But the note that explains the gap is rendered in the *least* visible style the panel has (`--textDimColor1` = `#ccc` on a white panel, ~1.6:1 contrast, italic 12px monospace) even though in the common case it is the **only** thing in the Debug panel. Biggest win: render it as an info callout in the style the pest_control module already uses for `.success`, instead of as dim "skipped/metadata" text.

Reviewed by running the app locally (port 8325) and driving the exact #2672 flow with Playwright: enter edit mode with Debug closed → start a `LABEL`/`INPUT`/`LABEL` routine → open Debug while the INPUT overlay is up → press *Go*.

Debug panel right after the note appears (the note is the faint italic line — it is the entire content of the panel):

Debug panel showing only the note

The note next to a normally-logged routine, so the styling difference is visible:

Note under a real log entry

---

## Big layout improvements

### 1. The note should be a panel-level callout, not the dimmest text on screen

`client/css/jsonedit.css:578` styles it with `--textDimColor1`, the variable otherwise reserved for de-emphasised secondary bits — `.jeLogSkipped` (operations that were *skipped*) and `.jeLogTime` (the `(0ms)` suffix). That variable is `#ccc` in light mode (`client/css/editor/layout.css:15`) on a white panel → roughly **1.6:1** contrast, and `#606060` on `#08090a` in dark mode (`layout.css:37`) → roughly **3.4:1**. Both are below any readable threshold, and the italic + monospace combination at 12px makes a full prose sentence harder still.

That is exactly backwards for this message: it is a status explanation, and in the scenario the PR is about, the panel contains *nothing else*. A user who opens Debug mid-routine sees an almost-blank panel with a ghost line in it — the "second bug" the PR description wants to avoid is only half-avoided, because the note is easy to miss entirely.

The module already has a convention for exactly this: `client/css/editor/sidebarModules.css:295` defines `.pest_control.editorModule .success` — padded, bordered, rounded, coloured background, bold — used for *"No validation problems found!"*, which you can see sitting right under the note in the second screenshot. Mirroring that with an info/neutral variant would make the note read as a deliberate panel message rather than as greyed-out leftovers:

```css
.jeLogNote {
  color: var(--textColor);
  background-color: var(--backgroundHighlightColor1);
  border-left: 3px solid var(--textHighlightColor2);
  border-radius: 4px;
  padding: 8px 10px;
  margin: 10px 0;
  font-family: inherit;   /* it is prose, not code */
}
.jeLogNote .jeLogWidget,
.jeLogNote .jeLogProperty { font-family: monospace; font-weight: bold; }
```

(Colour values to taste — the point is: readable foreground, a container, and the identifiers kept monospace while the sentence is not.)

### 2. The note silently kills the expanders on entries already in the log

`jeLoggingRoutineNotLogged` (`client/js/jsonedit.js:3141`) does `$('#jeLog').innerHTML = jeLoggingHTML`, but the expander click handlers are attached in `jeLoggingRoutineEnd` (`client/js/jsonedit.js:3090`–`3105`) *after* its own `innerHTML` assignment. Re-writing `innerHTML` here rebuilds the DOM and drops every one of those listeners.

Reproduced in the browser: with *Clear after each interaction* unchecked, log one routine (its ▶ arrow expands/collapses fine), close Debug, start the INPUT routine, re-open Debug, press *Go* — the note appears and the earlier entry's ▶ arrow no longer responds to clicks at all, and any manual expansion state is reset. So the fix for one dead-end (a crash) introduces a smaller dead-end (a log you can no longer open).

Suggestion: pull the post-render wiring (expander listeners + red-expander pass + the filter re-apply) out of `jeLoggingRoutineEnd` into a small `jeLoggingRenderLog()` and call it from both places — or simply append the note to `jeLoggingHTML` without touching `#jeLog` and let the next render pick it up (at the cost of the note not appearing until then).

- **Indentation is a magic number and slightly off.** `.jeLogNote { margin-left: 38px }` (`jsonedit.css:578`) is meant to stand in for the missing ▶. Measured in the browser: the note's left edge is at 708px while a top-level entry's header is at 690px and its widget name renders at ~703px — so the note ends up ~5px right of the text it is supposed to line up with, and the 38px will drift if `.jeExpander::before` (`content:"\25B6"` + `margin-right:6px`, `jsonedit.css:585`) or the font ever changes. Prefer `margin-left: 20px` (matching `.jeLog`) plus a real spacer of the same width as the caret, or derive both from one variable.
- **The identifiers have no visual boundary.** `<span class="jeLogWidget">` / `<span class="jeLogProperty">` have **no CSS rules anywhere** in `jsonedit.css`. In a normal entry that is fine — they sit alone on the header line. Inlined into a sentence they render identically to the prose, so it reads as the run-on *"button clickRoutine was not logged because…"*. Bolding or quoting them inside `.jeLogNote` (see snippet above) fixes it.
- **The filter is not re-applied.** `jeLoggingRoutineEnd` ends with `jeLoggingFilterLog($('#jeLogFilter').value)` (`jsonedit.js:3120`); `jeLoggingRoutineNotLogged` re-renders `#jeLog` without it. With text typed into *Filter log…*, the note's render drops whatever filtering was showing.
- **The note outlives its context.** It consumes `jeRoutineResetOnNextLog`, so it stays pinned at the top of the log while later, unrelated routines get appended underneath it (second screenshot). With no container around it, a newcomer can easily read it as a heading for the entry *below* it. The callout box in #1 also solves this.

- **"logging was enabled" doesn't match any label in the UI.** Nothing the user clicked is called "logging" — they clicked a bug icon and got a panel; the surrounding controls say *Filter log…*, *Clear after each interaction*, *CLEAR*. Someone seeing this for the first time has to infer that "logging was enabled" means "you opened this panel". 
- **It says what went wrong but not what to do.** The one action that fixes it (run the routine again, now that Debug is open) is exactly what the user needs and isn't mentioned.

  Suggested wording for `jsonedit.js:3134`–`3139`:

  > **`button` › `clickRoutine`** was already running when the Debug panel was opened, so it could not be recorded. Run it again to see its log.

- **`--custom--` will read as an error string.** When `property` isn't a string the note prints *"button --custom-- was not logged…"* (`jsonedit.js:3137`). It's consistent with `jeLoggingRoutineStart`, but there it's a lone column label, whereas here it lands mid-sentence where a newcomer will read it as a placeholder that failed to substitute. Something like "an inline routine on `button`" would read as English.

VTT-AI-Agent:
Thanks — the UI/UX points are addressed in e25c599.

| | light | dark |
|---|---|---|
| Debug panel after the change | light | dark |

**1. Callout instead of the dimmest text on screen.** `.jeLogNote` no longer uses `--textDimColor1`/italic. It is now `--textColor` on `--backgroundHighlightColor1` with a `--textHighlightColor2` left border, rounded and padded — i.e. the same "panel message" shape as the module's `.success` box, but theme-variable driven so it works in both modes (black on `#ccc` in light, `#f8f8f2` on `#444` in dark). The container also fixes the "note outlives its context" nit: it now clearly reads as its own box rather than as a heading for the entry below it.

**2. Dead expanders after the note renders.** Real bug, fixed. The post-render wiring (expander listeners, red-expander pass, filter re-apply) moved out of `jeLoggingRoutineEnd` into a new `jeLoggingRenderLog()`, and both call sites go through it — so the note render re-attaches the handlers and re-applies the filter instead of dropping them. Verified in a browser with *Clear after each interaction* unchecked: log an entry, close Debug, run the INPUT routine, re-open Debug, press *Go* — the note appears and the earlier entry's arrow still toggles.

**3. The `38px` magic number.** Removed; the note just inherits `.jeLog`'s `margin-left: 20px`, so it lines up with the entries exactly.

While measuring that I found the actual cause of the misalignment you saw: `#jeLog` still carried `margin-left: -40px` from when the log was positioned inside `#jsonEditor`, so since it got its own sidebar module the whole log has been sticking out 20px to the left of the panel — which silently clipped the **top-level entries' ▶ arrows** (visible in your first screenshot: `other clickRoutine` has no arrow, the nested `LABEL` does). Dropping that rule puts the log back inside the module and the arrows are visible again — that is the `▼` in front of `other clickRoutine` above.

**4. Unstyled identifier spans.** `.jeLogNote .jeLogWidget`/`.jeLogProperty` are now bold monospace while the sentence itself is `Arial, sans-serif`, so the identifiers stand out of the prose instead of running into it.

**5. Filter not re-applied.** Fixed by #2 — `jeLoggingRenderLog()` ends with the same `jeLoggingFilterLog($('#jeLogFilter').value)` call that `jeLoggingRoutineEnd` had.

**6. Wording + `--custom--`.** Reworded to your suggestion, including the "what to do" part:

> `button` `clickRoutine` was already running when the Debug panel was opened, so it could not be recorded. Run it again to see its log.

and for a non-string property it now reads *"an inline routine of `button` was already running when the Debug panel was opened, …"* instead of printing `--custom--` mid-sentence.

`npm test`: 86/86. `tests/testcafe/editor.js`: 15/15 (the regression test's assertion follows the new wording).

VTT-AI-Agent (commented):
## TL;DR

No actionable findings in the current revision. The per-invocation logging snapshot fixes the unmatched start/end failure, and the follow-up commits cover the observable regression while addressing the log-clearing and re-rendering edge cases raised in the prior reviews.

## Detailed review

I re-checked the complete diff at `e25c5993`, including every `jeRoutineLogging` → `routineLogging` replacement in `client/js/widgets/widget.js`. The snapshot is consistently used for routine, operation, and synthetic nested-operation start/end pairs, so toggling Debug during an awaited operation can no longer leave the logging stack unbalanced. The defensive guards do not mask an incomplete pairing in the revised path.

The latest `client/js/jsonedit.js` changes also address the adjacent cases cleanly: clearing removes log fragments retained by active stack frames, and both ordinary completion and the new not-recorded callout go through the same render path so expander handlers, problem indicators, and the active filter are restored after `innerHTML` replacement. Widget and routine identifiers in the new callout are HTML-escaped.

The TestCafe regression exercises the original failure boundary (logging off at invocation, `INPUT` suspension, logging enabled before continuation), asserts that the post-`INPUT` mutation completes, captures page errors and unhandled rejections, checks the explanatory note, and verifies the final room-state hash. All current GitHub checks pass, including both editor TestCafe jobs and CodeQL.

`client/js/widgets/widget.js` is an engine-path file, but its edits are incidental to engine behavior: they only stabilize Edit mode's opt-in Debug logging and do not alter routine-operation semantics, widget defaults, events, or state updates for existing games when Debug logging is off.

The `beta` label was already present; I left it in place because this debug-only fix is safe and useful for player testing despite the incidental engine-path edit.

ArnoldSmith86 (approved):

PR #3071, opened 2026-07-29
Add front/back image upload and text card sections to the new deck wizard

VTT-AI-Agent:
## Goal

Make the deck editor's **"Add a new deck"** wizard cover two common ways of building a deck that it could not do before: uploading a **separate front and back image for every card**, and a **plain text deck** in the style of the Cards Against Humanity decks in the public library.

Two follow-ups from the PR discussion are in here as well: the option list was [regrouped into three collapsible sections](https://github.com/ArnoldSmith86/virtualtabletop/pull/3071#issuecomment-5163259566) so that adding a deck is easier to understand, and the text-cards section [can now hold line breaks](https://github.com/ArnoldSmith86/virtualtabletop/pull/3071#issuecomment-5171069596), both inside a card's text and in the deck label.

### Three sections instead of eight radio buttons

The dialog's flat list of ways to create a deck is now three expanders, only one open at a time (`openNewDeckGroup`):

- **New blank deck** — the empty deck flow. Open and selected when the dialog opens, so its card size / deck id form is right there; being the only way in its section, it needs no radio button of its own.
- **Use an existing deck** — public library, traditional decks, Tabletop Simulator import.
- **Create a custom deck** — suits and ranks, one image per card, a front and a back image per card, text cards.

Every option is a short title plus a one-line explanation, and the panel of the option you pick is rendered inside the open section, right under the options it belongs to. The old "collapse the list once a mode is picked, `Change` to bring it back" mechanic is gone — the sections keep the dialog short by themselves.

### The two new ways to build a deck

Two new sections (`client/editor.html` option list → `renderNewDeckPanel` → two new `PropertiesModule` flows next to the existing `deckImages`/`deckImportTTS` ones):

### "Upload a front and a back image for every card"

- Two bulk uploads: all card fronts, all card backs.
- Both lists are sorted by file name (numbers compared numerically, so `front2` sorts before `front10`) and then matched up **by position** — `front1.jpg` gets `back1.jpg`. The lists are rendered in that pairing order and numbered, each row with a thumbnail, the file name and a delete button, so the pairing is visible before adding anything.
- A status line reports the pairing (`3 card types, each front paired with the back in the same position.`) or, when the counts don't match, says so and keeps "Add to game" disabled. Uploading exactly **one** back image uses it for every card.
- Every card type carries its own `backImage`, which face 0 renders through a dynamic property (`deckImages` can only give the whole deck one shared back). Card width comes from the first front image's aspect ratio, plus a "copies of each card" input.

### "Text cards — one card per line of text"

- A textarea plus a radio row that picks how the typed text is cut into cards: **one card per line** (right for a pasted list) or **a blank line between cards**, which lets a card`s text span several lines. Each card becomes a card type whose `text` property holds its text; face 1 renders it through a dynamic property, exactly like the `prompt`/`response` decks of Cards Against Humanity.
- Design inputs: card color, text color, font size, card width/height, copies per card and an optional deck label that is shown large on the backs and small along the bottom of the fronts. The label is a textarea too, so it can be shown across several lines on the backs the way the Cards Against Humanity backs are - the one-line footer on the fronts flattens it. A live preview of the first card updates as you type.
- Card type names are derived from the text (sanitized and shortened), fall back to a running number when a line has no usable characters (`______ + ______`), and repeats get a `(2)` suffix that cannot collide with the generated `<card type> <copy number>` card ids.

Both sections build their deck through the existing `addDeckWithCards`, so the result is the familiar holder + `Recall & Shuffle` button + pile of cards, and `pendingNewDeck` opens the new deck in the deck editor afterwards.

## Screenshots

| | |
|---|---|
| The dialog as it opens: three sections, the blank deck one open | blank |
| "Use an existing deck" opened | existing |
| "Create a custom deck" opened | custom |
| Front/back upload, files matched by name | pairing |
| The resulting cards, fronts and per-card backs | result |
| Text cards section | text |
| …with the live preview | preview |
| Text cards with line breaks, and the deck label across three lines | linebreaks · [the resulting back](https://agent.virtualtabletop.io/reports/pr/1532145112908435497/pr3071-linebreaks-back.png) |

## Testing

- New TestCafe test `Deck editor: add a deck of text cards from the new deck wizard` drives the text section (including a duplicate line and a line without usable characters) and compares the resulting room state hash.
- `Deck editor: text cards with line breaks in the new deck wizard` drives the blank-line mode with an indented line, a doubled separator and a trailing blank line, and asserts the resulting `cardTypes`, the multi-line back label and the flattened front footer.
- `npm test` (jest, 86 tests) and the full `tests/testcafe/editor.js` suite (20 tests) pass locally in headless Chromium.
- The image-upload section was verified manually in a browser: files uploaded out of order end up paired `F1↔B1`, `F2↔B2`, `F3↔B3`, and each card's back face renders its own image.
- No widget/routine properties were added, so `validator/` and `jeCommands` need no updates; a game file created with both new sections validates cleanly with `validate_gamefile_node.js`.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3071/pr-test (or any other room on that server)

### DIFF
--- a/client/css/editor/deckeditor.css
+++ b/client/css/editor/deckeditor.css
@@ -693,37 +693,108 @@ body.deckEditorRoomVisible #deckEditorMainCol > .overlay {
   color: #fff;
 }
 
-.deckEditorNewDeckModes {
+/* The ways to create a deck are grouped into three expanders, only one of which is open at a time - so the
+   dialog opens as three short rows instead of a list of eight radio buttons. */
+.deckEditorNewDeckGroups {
   display: flex;
   flex-direction: column;
   gap: 8px;
   margin: 16px 0;
 }
 
-.deckEditorNewDeckModes label {
+.deckEditorNewDeckGroup {
+  border: 1px solid #0003;
+  border-radius: 6px;
+}
+
+.deckEditorNewDeckGroupHeader {
+  display: block;
+  position: relative;
+  width: 100%;
+  padding: 10px 40px 10px 12px;
+  border: none;
+  border-radius: 5px;
+  background: #0000000d;
+  color: inherit;
+  font-size: 15px;
+  font-weight: bold;
+  text-align: left;
+  text-transform: none;
+}
+
+.deckEditorNewDeckGroupHeader:hover {
+  background: #0000001a;
+}
+
+.deckEditorNewDeckGroupOpen .deckEditorNewDeckGroupHeader,
+.deckEditorNewDeckGroupOpen .deckEditorNewDeckGroupHeader:hover {
+  border-radius: 5px 5px 0 0;
+  background: var(--VTTblue);
+  color: #fff;
+}
+
+.deckEditorNewDeckGroupHeader span {
+  display: block;
+  font-size: 12px;
+  font-weight: normal;
+  opacity: 0.75;
+}
+
+.deckEditorNewDeckGroupHeader::after {
+  content: "expand_more";
+  position: absolute;
+  top: 50%;
+  right: 8px;
+  transform: translateY(-50%);
+  font-family: "Material Symbols";
+  font-size: 24px;
+  font-weight: normal;
+}
+
+.deckEditorNewDeckGroupOpen .deckEditorNewDeckGroupHeader::after {
+  content: "expand_less";
+}
+
+/* The panel of the selected option is moved into the open group (see openNewDeckGroup), so a closed group is
+   nothing but its header. */
+.deckEditorNewDeckGroupBody {
+  display: none;
+  padding: 12px;
+}
+
+.deckEditorNewDeckGroupOpen .deckEditorNewDeckGroupBody {
+  display: block;
+}
+
+.deckEditorNewDeckModes {
   display: flex;
-  align-items: center;
+  flex-direction: column;
   gap: 8px;
+}
+
+.deckEditorNewDeckModes label {
+  display: grid;
+  grid-template-columns: auto 1fr;
+  align-items: baseline;
+  column-gap: 8px;
   cursor: pointer;
 }
 
-/* Once a mode has been picked the list of options has done its job, so it collapses to the chosen one and
-   leaves the ~250px it occupies to the panel the user is now filling in. "Change" brings it back. */
-.deckEditorNewDeckModesCollapsed label:not(.deckEditorNewDeckModeSelected) {
+.deckEditorNewDeckModes label span {
+  grid-column: 2;
+  font-size: 12px;
+  opacity: 0.7;
+}
+
+/* A group with a single way to create a deck (the blank one) needs no radio button - opening it is already
+   the choice, and its header says what it does. */
+.deckEditorNewDeckModesSingle {
   display: none;
 }
 
-#deckEditorNewDeckChangeMode {
-  display: none;
-  align-self: flex-start;
-  padding: 2px 8px;
-}
-
-.deckEditorNewDeckModesCollapsed #deckEditorNewDeckChangeMode {
-  display: inline-flex;
-}
-
-#deckEditorNewDeckPanel:not(:empty) {
+/* Separates a group's options from the panel of the selected one - the blank group shows no options, so
+   there is nothing to separate there. */
+.deckEditorNewDeckModes:not(.deckEditorNewDeckModesSingle) + #deckEditorNewDeckPanel:not(:empty) {
   border-top: 1px solid #0002;
   padding-top: 16px;
 }
--- a/client/css/editor/deckeditor.css
+++ b/client/css/editor/deckeditor.css
@@ -707,9 +707,38 @@ body.deckEditorRoomVisible #deckEditorMainCol > .overlay {
   cursor: pointer;
 }
 
+/* Once a mode has been picked the list of options has done its job, so it collapses to the chosen one and
+   leaves the ~250px it occupies to the panel the user is now filling in. "Change" brings it back. */
+.deckEditorNewDeckModesCollapsed label:not(.deckEditorNewDeckModeSelected) {
+  display: none;
+}
+
+#deckEditorNewDeckChangeMode {
+  display: none;
+  align-self: flex-start;
+  padding: 2px 8px;
+}
+
+.deckEditorNewDeckModesCollapsed #deckEditorNewDeckChangeMode {
+  display: inline-flex;
+}
+
 #deckEditorNewDeckPanel:not(:empty) {
   border-top: 1px solid #0002;
   padding-top: 16px;
+}
+
+/* The dialog itself is the scroll container, so pinning the action row to its bottom edge keeps the primary
+   action reachable no matter how tall the panel above it grows. The negative offset and the matching padding
+   make it cover the dialog's own bottom padding, so nothing scrolls through the gap below it. */
+#deckEditorNewDeckPanel .goButton,
+.deckEditorNewDeckButtonBar {
+  position: sticky;
+  bottom: -20px;
+  z-index: 1;
+  margin-top: 8px;
+  padding: 8px 0 20px;
+  background: var(--modalBackgroundColor);
 }
 
 /* Card sizes for a new empty deck: a wrapping row of radio chips, the first (the size decks have always been
@@ -766,6 +795,12 @@ body.deckEditorRoomVisible #deckEditorMainCol > .overlay {
   display: block;
   background: transparent;
   color: inherit;
+}
+
+/* The sidebar's "overflow-x: hidden" would make this module a scroll container of its own, which would stop
+   the sticky action row below from sticking to the dialog (the actual scroll container). */
+.deckEditorNewDeckModule.tune.editorModule {
+  overflow: visible;
 }
 
 @media (max-width: 700px) {
--- a/client/css/editor/sidebarModules.css
+++ b/client/css/editor/sidebarModules.css
@@ -171,6 +171,93 @@
 }
 
 
+/* Front/back image upload: one row per uploaded file, numbered in the order the files are paired in. */
+.tune.editorModule .imagePairList {
+  margin-bottom: 5px;
+}
+
+.tune.editorModule .imagePairEntry {
+  display: flex;
+  align-items: center;
+  gap: 5px;
+  margin-bottom: 3px;
+}
+
+.tune.editorModule .imagePairEntry b {
+  min-width: 25px;
+  text-align: right;
+}
+
+.tune.editorModule .imagePairEntry img {
+  width: 30px;
+  height: 40px;
+  object-fit: contain;
+  background: #0002;
+}
+
+.tune.editorModule .imagePairEntry span {
+  flex-grow: 1;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+
+.tune.editorModule .imagePairStatus {
+  margin: 5px 0;
+  font-size: 12px;
+  opacity: 0.7;
+}
+
+.tune.editorModule .imagePairStatus.imagePairMismatch {
+  color: #ff8080;
+  opacity: 1;
+}
+
+.tune.editorModule .imagePairOptions [type=number] {
+  width: 50px;
+}
+
+
+/* Text cards: the text input, the design grid and the preview of the first few cards. */
+.tune.editorModule .textCardsInput {
+  width: 100%;
+  box-sizing: border-box;
+  margin-bottom: 5px;
+}
+
+.tune.editorModule .textCardsDesign {
+  display: grid;
+  grid-template-columns: 1fr 1fr;
+  gap: 5px 10px;
+  margin-bottom: 5px;
+}
+
+.tune.editorModule .textCardsDesign label {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  gap: 5px;
+}
+
+.tune.editorModule .textCardsDesign .textCardsLabelWrap {
+  grid-column: 1 / -1;
+}
+
+.tune.editorModule .textCardsDesign input {
+  width: 60px;
+}
+
+.tune.editorModule .textCardsDesign .textCardsLabel {
+  flex-grow: 1;
+  width: auto;
+  min-width: 0;
+}
+
+.tune.editorModule .textCardsPreview {
+  text-align: center;
+}
+
+
 .tune.editorModule .ttsImportLink {
   display: flex;
   gap: 5px;
--- a/client/css/editor/sidebarModules.css
+++ b/client/css/editor/sidebarModules.css
@@ -268,6 +268,25 @@
   font-family: inherit;
 }
 
+/* How the typed text is cut into cards - one row of radios right under the textarea it describes. */
+.tune.editorModule .textCardsSplit {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 4px 20px;
+  margin-bottom: 10px;
+}
+
+.tune.editorModule .textCardsSplit label {
+  display: flex;
+  align-items: center;
+  gap: 6px;
+}
+
+.tune.editorModule .textCardsSplit span {
+  font-size: 11px;
+  opacity: 0.7;
+}
+
 .tune.editorModule .textCardsColumns {
   display: flex;
   align-items: flex-start;
@@ -320,10 +339,18 @@
   padding: 1px;
 }
 
+/* A textarea, so a deck label can carry the line breaks it is shown with on the card backs. */
+.tune.editorModule .textCardsDesign .textCardsLabelWrap {
+  align-items: flex-start;
+}
+
 .tune.editorModule .textCardsDesign .textCardsLabel {
   flex-grow: 1;
   width: auto;
   min-width: 0;
+  box-sizing: border-box;
+  font-family: inherit;
+  resize: vertical;
 }
 
 .tune.editorModule .textCardsPreview {
--- a/client/css/editor/sidebarModules.css
+++ b/client/css/editor/sidebarModules.css
@@ -25,6 +25,15 @@
   display: inline-flex;
 }
 
+/* The row with the primary action. Laid out as a flex row so a status text placed before the button - the
+   explanation of why it is disabled - sits next to it instead of somewhere further up the panel. */
+.editorModule .goButton.buttonBar {
+  display: flex;
+  align-items: center;
+  justify-content: flex-end;
+  gap: 12px;
+}
+
 /* Shown by the Properties module when nothing is selected (see PropertiesModule.addDeck). Use the theme
    text color so the intro is readable in both themes (white on the dark editor theme, dark in light mode). */
 .editorModule .noSelectionIntro {
@@ -171,9 +180,29 @@
 }
 
 
-/* Front/back image upload: one row per uploaded file, numbered in the order the files are paired in. */
+/* Front/back image upload: two side-by-side lists, one row per uploaded file, numbered in the order the files
+   are paired in - so row n of the fronts and row n of the backs line up as the pair they are. */
+.tune.editorModule .imagePairColumns {
+  display: grid;
+  grid-template-columns: 1fr 1fr;
+  gap: 0 12px;
+  align-items: start;
+}
+
+.tune.editorModule .imagePairHint {
+  margin-bottom: 5px;
+  font-size: 11px;
+  opacity: 0.7;
+}
+
 .tune.editorModule .imagePairList {
   margin-bottom: 5px;
+}
+
+.tune.editorModule .imagePairList:empty::before {
+  content: attr(data-empty-hint);
+  font-size: 12px;
+  opacity: 0.7;
 }
 
 .tune.editorModule .imagePairEntry {
@@ -184,67 +213,111 @@
 }
 
 .tune.editorModule .imagePairEntry b {
-  min-width: 25px;
+  min-width: 20px;
   text-align: right;
 }
 
 .tune.editorModule .imagePairEntry img {
-  width: 30px;
-  height: 40px;
+  width: 48px;
+  height: 64px;
   object-fit: contain;
   background: #0002;
 }
 
+/* The delete button follows the file name directly instead of being pushed to the far edge of the column. */
 .tune.editorModule .imagePairEntry span {
-  flex-grow: 1;
+  flex: 0 1 auto;
   overflow: hidden;
   text-overflow: ellipsis;
   white-space: nowrap;
 }
 
-.tune.editorModule .imagePairStatus {
-  margin: 5px 0;
+.tune.editorModule .imagePairStatus,
+.tune.editorModule .textCardsStatus {
+  flex-grow: 1;
+  text-align: left;
   font-size: 12px;
   opacity: 0.7;
 }
 
 .tune.editorModule .imagePairStatus.imagePairMismatch {
-  color: #ff8080;
+  color: var(--negativeInput);
   opacity: 1;
 }
 
+.tune.editorModule .imagePairOptions {
+  margin: 5px 0;
+}
+
+.tune.editorModule .imagePairOptions label {
+  display: inline-flex;
+  align-items: center;
+  gap: 8px;
+}
+
 .tune.editorModule .imagePairOptions [type=number] {
-  width: 50px;
+  width: 60px;
 }
 
 
-/* Text cards: the text input, the design grid and the preview of the first few cards. */
+/* Text cards: the text input, then the design grid and the preview of the first card side by side. */
 .tune.editorModule .textCardsInput {
   width: 100%;
   box-sizing: border-box;
   margin-bottom: 5px;
+  font-family: inherit;
+}
+
+.tune.editorModule .textCardsColumns {
+  display: flex;
+  align-items: flex-start;
+  gap: 12px;
+}
+
+.tune.editorModule .textCardsColumn {
+  flex: 1 1 auto;
+  min-width: 0;
+}
+
+.tune.editorModule .textCardsPreviewColumn {
+  flex: 0 0 auto;
 }
 
 .tune.editorModule .textCardsDesign {
   display: grid;
   grid-template-columns: 1fr 1fr;
-  gap: 5px 10px;
+  gap: 5px 16px;
   margin-bottom: 5px;
 }
 
+/* Each input hugs its own label: right-aligning them across a 1fr column would park the input of one field
+   next to the label of the next one. */
 .tune.editorModule .textCardsDesign label {
   display: flex;
   align-items: center;
-  justify-content: space-between;
-  gap: 5px;
+  gap: 8px;
 }
 
-.tune.editorModule .textCardsDesign .textCardsLabelWrap {
+.tune.editorModule .textCardsDesign .textCardsLabelWrap,
+.tune.editorModule .textCardsDesign .textCardsHint {
   grid-column: 1 / -1;
+}
+
+.tune.editorModule .textCardsDesign .textCardsHint {
+  font-size: 11px;
+  opacity: 0.7;
 }
 
 .tune.editorModule .textCardsDesign input {
   width: 60px;
+}
+
+/* Distinguish the color swatches from the number fields at a glance - a white swatch at input width reads as
+   an empty text box. */
+.tune.editorModule .textCardsDesign input[type=color] {
+  width: 40px;
+  height: 24px;
+  padding: 1px;
 }
 
 .tune.editorModule .textCardsDesign .textCardsLabel {
@@ -254,7 +327,13 @@
 }
 
 .tune.editorModule .textCardsPreview {
-  text-align: center;
+  position: relative;
+  overflow: hidden;
+}
+
+.tune.editorModule .textCardsPreviewCard {
+  position: relative;
+  transform-origin: top left;
 }
 
 
--- a/client/css/layout.css
+++ b/client/css/layout.css
@@ -1193,6 +1193,13 @@ button:disabled:hover {
   cursor: not-allowed;
 }
 
+/* The colored variants above are more specific than button:disabled, so a disabled green/red icon button
+   would keep its full color and not look disabled at all. */
+button[icon]:disabled,
+button[icon]:disabled:hover {
+  background-color: #808080;
+}
+
 .ui-button {
   background-color: var(--VTTblue);
   padding: 0 8px;
--- a/client/editor.html
+++ b/client/editor.html
@@ -230,17 +230,37 @@
     <div id="deckEditorNewDeckDialog" class="deckEditorModal">
       <button id="deckEditorNewDeckClose" icon="close" title="Close"></button>
       <h1>Add a new deck</h1>
-      <p>Add blank deck or choose an existing one to add to the game</p>
-      <div class="deckEditorNewDeckModes">
-        <label><input type="radio" name="deckEditorNewDeckMode" value="empty" checked>Empty deck &mdash; design its faces and card types here in the deck editor</label>
-        <label><input type="radio" name="deckEditorNewDeckMode" value="library">Browse the public library of ready-made decks</label>
-        <label><input type="radio" name="deckEditorNewDeckMode" value="traditional">Traditional deck (standard, Spanish, German, Swiss, tarot, hanafuda, mahjong, dominoes)</label>
-        <label><input type="radio" name="deckEditorNewDeckMode" value="custom">Custom deck of cards (define suits and ranks)</label>
-        <label><input type="radio" name="deckEditorNewDeckMode" value="images">Upload one image per card or tiled images &mdash; one shared back for the whole deck</label>
-        <label><input type="radio" name="deckEditorNewDeckMode" value="imagePairs">Upload a separate front and back image for every card</label>
-        <label><input type="radio" name="deckEditorNewDeckMode" value="text">Text cards &mdash; one card per line of text</label>
-        <label><input type="radio" name="deckEditorNewDeckMode" value="tts">Import from Tabletop Simulator</label>
-        <button id="deckEditorNewDeckChangeMode" icon="expand_more">Change</button>
+      <p>Where should the cards come from? Open one of the three sections.</p>
+      <div class="deckEditorNewDeckGroups">
+        <div class="deckEditorNewDeckGroup" id="deckEditorNewDeckGroupBlank">
+          <button class="deckEditorNewDeckGroupHeader">New blank deck<span>Start from nothing and design the faces and card types here in the deck editor</span></button>
+          <div class="deckEditorNewDeckGroupBody">
+            <div class="deckEditorNewDeckModes deckEditorNewDeckModesSingle">
+              <label><input type="radio" name="deckEditorNewDeckMode" value="empty" checked>Empty deck</label>
+            </div>
+          </div>
+        </div>
+        <div class="deckEditorNewDeckGroup" id="deckEditorNewDeckGroupExisting">
+          <button class="deckEditorNewDeckGroupHeader">Use an existing deck<span>Take a deck that is already finished, from the library or another game</span></button>
+          <div class="deckEditorNewDeckGroupBody">
+            <div class="deckEditorNewDeckModes">
+              <label><input type="radio" name="deckEditorNewDeckMode" value="library"><b>Public library</b><span>Browse the ready-made decks shared in the public library</span></label>
+              <label><input type="radio" name="deckEditorNewDeckMode" value="traditional"><b>Traditional deck</b><span>Standard, Spanish, German, Swiss, tarot, hanafuda, mahjong or dominoes</span></label>
+              <label><input type="radio" name="deckEditorNewDeckMode" value="tts"><b>Tabletop Simulator</b><span>Import a deck from a Tabletop Simulator save file</span></label>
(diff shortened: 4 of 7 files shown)


VTT-AI-Agent (commented):
## TL;DR

Both new wizard sections work as advertised — I ran the new TestCafe test (passes) and drove the untested front/back-image flow in a real browser: uploading `front10/front2/front1` + `back2/back10/back1` out of order pairs them correctly as `front1↔back1`, `front2↔back2`, `front10↔back10`, each card gets its own `backImage`, and the deck lands in the holder/pile and opens in the deck editor. Nothing here is a correctness blocker; the most worthwhile fix is that the "copies of each card" inputs ignore their own `max=99` (a typed value goes straight into the card-creation loop), and the image-pair flow — the more intricate half of the PR — has no automated test.

## Findings

**1. `copies` is not clamped to the declared maximum — `properties.js:1435` and `properties.js:1604`**

Both flows do `Math.max(1, +input.value || 1)`, and `max=99` on a `type=number` input is not enforced for typed/pasted values. In `deckTextCards` this multiplies with an unbounded line count: paste 500 lines, type `500` into "Copies of each card", and `addDeckWithCards` runs 250 000 sequential `addWidgetLocal` calls inside one batch — the client hangs and the room state is unusable afterwards. `Math.min(99, …)` (and maybe a confirmation above a few hundred cards) closes it. The existing `deckImages` flow has the same unclamped pattern, but there the count is per uploaded file rather than per pasted line, so the blast radius is much smaller.

**2. The "live preview" reorders itself on every keystroke — `properties.js:1589` / `properties.js:6571`**

`renderWidgetButton` renders `shuffleArray(Object.keys(cardTypes)).slice(0, 5)`, so each `updatePreview()` shows the same five cards in a fresh random order. Confirmed in the browser: typing `Alpha/Bravo/Charlie` and firing three renders gave `Bravo, Alpha, Charlie` twice and then `Charlie, Alpha, Bravo`. It makes the preview visibly jump while you type, and it means the comment "a live preview of the first few cards" isn't quite what the user sees. Rendering the preview cards directly (or only re-rendering on `change` rather than `input`) would be steadier.

**3. No test for the image-pair flow — `tests/testcafe/editor.js:761`**

The added test covers the text section only. The pairing logic (numeric-aware name sort, shared-back special case, index pairing, delete-and-repair, aspect-ratio width) is the part most likely to regress, and it's the part with zero coverage. Even a small unit-level test over the comparator, or a TestCafe test that drives two `uploadAsset` pickers with fixture images and hashes the state, would be worth it. (I verified the current behaviour manually — it is correct today.)

**4. Misleading "Paired with …" tooltips while the counts don't match — `properties.js:1406`**

`renderList` sets `row.title = 'Paired with <file>'` whenever `pairedWith[index]` exists, including when `fronts.length != backs.length` and "Add to game" is disabled. With 3 fronts and 2 backs the first two rows claim a pairing that the wizard is refusing to create. Guarding the title on the same `paired` flag the status line uses keeps the two consistent.

**5. Long text runs under the front footer label — `properties.js:1547` / `properties.js:1572`**

When a deck label is set, the front gets a footer at `y = height - padding - footerHeight`, but the main text object keeps `height: height - 2*padding`, so a card text that fills the card overlaps the footer (text objects don't clip). Subtracting `footerHeight` from the main text box when `label` is set avoids it.

**6. Duplicate-suffix loop trips over inherited `Object.prototype` keys — `properties.js:1444` and `properties.js:1560`**

`cardTypes` is a plain `{}`, so `cardTypes[cardType] !== undefined` is true for `constructor`, `toString`, `valueOf`, `__proto__`… A front image named `constructor.png` therefore becomes card type `constructor (2)` even though nothing is duplicated. Purely cosmetic (there's no infinite loop and no prototype pollution — the assignment happens only after the loop found a free name), but `Object.create(null)` or `Object.prototype.hasOwnProperty.call(...)` makes it exact.

## Verified locally

- `npx testcafe chromium:headless tests/testcafe/editor.js -t 'Deck editor: add a deck of text cards from the new deck wizard'` → passes, state hash matches.
- Playwright run of the image-pair section: 3 fronts + 3 backs uploaded in scrambled order produce `cardTypes: {front1:{image,backImage}, front2:{…}, front10:{…}}` with correct per-card backs, `cardDefaults.width: 107` from the 40×60 source aspect ratio, and the wizard's deck opens in the deck editor.

Engine note for the beta build: no engine file is touched at all — the diff is confined to `client/js/editor/`, `client/editor.html`, `client/css/editor/` and the TestCafe suite, so no widget, `compute.js` or `serverstate.js` behaviour changes and existing games are unaffected.

I added the `beta` label: the change is purely additive editor UI, I exercised both new flows end to end without errors, and it can't affect how existing games run.

VTT-AI-Agent:
Thanks for the thorough pass — all six findings are addressed in 2f3344dc.

**1. `copies` not clamped** — fixed. Both flows now go through a small `copiesFromInput(input)` helper that clamps to the input's own `min`/`max`, so a typed `500` becomes `99`. Since 99 copies of a long pasted list can still add up, `addDeckWithCards` now asks for confirmation above 500 cards (which also covers the traditional/library/TTS flows). Verified in a browser: 3 lines with a typed `500` produces 297 cards and no dialog; 6 lines × 99 = 594 shows *"This creates 594 cards…"* and dismissing it adds nothing at all.

**2. Preview reshuffles on every keystroke** — fixed. `renderWidgetButton` got an optional `shuffleCardTypes` parameter (default `true`, so the other previews keep their variety) and the text-cards preview passes `false`. Three consecutive renders of `Alpha/Bravo/Charlie` now all come out as `Alpha | Bravo | Charlie`.

**3. No test for the image-pair flow** — added: *"Deck editor: pair front and back images in the new deck wizard"*. The file picker can't be driven from a test, so it stubs `uploadAsset` with the asset paths the server would have returned and uploads `front10/front2/front1` + `back2/back10/back1` out of order. It asserts the resulting pairing (`front1 + back1`, `front2 + back2`, `front10 + back10`) and the aspect-ratio card width (`107` from 40×60 fronts), so the numeric sort, the index pairing and the width all have coverage now.

**4. Misleading "Paired with …" tooltips** — fixed, the rows only get the tooltip when the same `paired` flag the status line uses is set (and not in the shared-back case).

**5. Text running under the front footer** — fixed. `footerHeight` is computed before the face objects are built and subtracted from the main text box when a label is set. With the default 150×233 card the front is now `text y=13 height=196` + `footer y=209 height=11`, i.e. exactly flush: [screenshot](https://agent.virtualtabletop.io/reports/pr/1532145112908435497/text-cards-label-footer.png).

**6. `Object.prototype` keys in the dedup loop** — fixed, both loops use `Object.prototype.hasOwnProperty.call(cardTypes, cardType)` now (matching the existing usage elsewhere in the file), so a `constructor.png` front stays `constructor`.

Verified locally: `npm test` (86 passed) and the full `tests/testcafe/editor.js` suite in Chromium (17 passed). The text-cards reference hash changed to `e5bc265b4af231b7f838092b4fc2c825` because of finding 5.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review) of the two new "Add a new deck" wizard sections. Both flows are logically well built and the pairing feedback is a nice touch, but **the dialog is now taller than the viewport in the text-cards mode — the live preview and the "Add to game" button are entirely below the fold at 1400×900**, and the preview itself is a 150px shuffled card fan whose text is unreadable, so it can't do the job it exists for. Most impactful single change: collapse the 8-item mode list once a mode is picked (and/or put "Add to game" in a sticky footer) so the section you are actually filling in is the thing on screen.

Reviewed by running the branch locally and driving the wizard in Chromium (uploads, mismatch state, live preview, resulting cards).

---

## Big layout improvements

### 1. The mode list eats the dialog; the panel you're working in falls below the fold

`#deckEditorNewDeckDialog` keeps all 8 radio options + the intro paragraph above every panel. In the text-cards mode the dialog measures `scrollHeight 1117` vs `clientHeight 810` at 1400×900 — i.e. the **Preview** section and the green **Add to game** button are not visible at all until you scroll inside the dialog, and the dialog's internal scrollbar is not obvious:

text mode

At 1100×700 it's worse — you get "Card texts", "Card design" and a *Preview* header with nothing under it, and no visible way to commit:

small viewport

Concrete suggestions (any one helps, the first two most):
- Collapse `.deckEditorNewDeckModes` to just the selected option (with a "change" / chevron affordance) once a mode is chosen — the list has done its job at that point, and it costs ~250px on every panel.
- Make only `#deckEditorNewDeckPanel` scroll and pin the `.goButton` row as a sticky footer inside the dialog, so the primary action is always reachable. This also fixes the same problem for the existing `images`/`tts` panels.
- For text cards specifically: the design grid and the preview could sit side by side (`grid-template-columns: 1fr auto`) instead of stacked — the preview is exactly the thing you want visible *while* you change font size and colors, and it currently never is.

### 2. The live preview is too small to preview anything

`.textCardsPreview` renders through `renderWidgetButton` (properties.js:1604), which is the small deck *selection thumbnail*: five cards, fanned and rotated, on a transparency checkerboard, ~150px wide. The text is illegible at that size, so font size / text color / wrapping — the only reasons to look at it — can't be judged:

preview closeup

Suggestion: render **one** card front at (or near) its real pixel size instead of the fan — a single `Card().renderReadonlyCopyRaw` at `width`/`height` from the design inputs, upright, on the dialog background rather than the checkerboard. If you want the fan, keep it *in addition* and make it secondary. Also label the section "Preview (first card)" / "(first 5 cards)" so nobody reads the fan as "my deck has 5 cards".

Related and visible in the output: a long line overflows the card and collides with the deck-label footer, spilling past the rounded edge (default 16px font, 150×233 card). The code reserves `footerHeight`, but text objects don't clip, so this ships as a broken-looking card — and the current preview is far too small to catch it before you commit:

resulting cards

A cheap fix at the UI level: when the longest line wouldn't fit at the chosen font size, show a warning next to the font-size input ("some texts won't fit at 16px — try 12px"), or auto-suggest a size.

### 3. Front/back pairing is shown as two lists that never meet

The two lists are stacked vertically, so with 30 fronts the backs list and the status line are a long scroll away, and the pairing exists only as matching ordinals plus a `title` tooltip:

image pairs

Suggestion that matches the mental model ("front1 ↔ back1"): once both lists are non-empty, render **one paired list** — each row `n. [front thumb] front1.png  ↔  [back thumb] back1.png` with a delete on each side. The two upload buttons stay above it. That makes the pairing self-evident without tooltips and halves the vertical space. If a single list is too big a change, at least put the two lists in a 2-column grid so row *n* of each is on the same line.

- **`sidebarModules.css:211` — the mismatch color fails contrast.** `.imagePairMismatch { color: #ff8080 }` is a light red tuned for the dark sidebar, but this module only ever renders inside the light `#deckEditorNewDeckDialog` (measured background `rgb(240,240,240)`) — the "3 fronts but 2 backs" warning comes out at roughly 2:1 contrast, at 12px, and is the *palest* text in the dialog while being the most important. The app already has a token for this: use `var(--negativeInput)` (`#950000`), which is what the rest of the UI uses for negative state. `#ff8080` appears nowhere else in `client/css`.
- **The error message is also mis-placed.** The status line sits under the backs list while the disabled **Add to game** is at the far right of the next row, so the explanation and the blocked action are visually unrelated. Put the status immediately left of the go button (or directly above it) so "why is this greyed out" answers itself.
- **The disabled `Add to game` doesn't look disabled.** `button[icon].green` (layout.css:990) beats `button:disabled` (layout.css:1189), so the button stays full `darkgreen`; only the label greys slightly. This is pre-existing, but both new sections *lean on* the disabled state as their main validation signal, so it's worth adding `button[icon].green:disabled { background-color: #808080 }` (or `opacity: .5`) while you're here.
- **`sidebarModules.css:235` — the design grid pairs labels with the wrong-looking input.** `.textCardsDesign label { justify-content: space-between }` across `1fr 1fr` columns pushes each input to the far right of its cell, so on screen you read "Card color ⟶ ⬛ Text color ⟶ ⬜": the swatch for *Card color* sits immediately left of the words *Text color*. Either right-align consistently with a fixed label column (`grid-template-columns: max-content 1fr max-content 1fr`) or drop `space-between` for a small gap so each input hugs its own label.
- **The default white *Text color* swatch reads as an empty text field** on the light dialog — a 1px border and the same 60px width as the number inputs. Giving the color inputs a distinct width/height (e.g. `width: 40px; height: 24px`) would separate them from the numeric fields at a glance.
- **`properties.js:1531` — the deck-label input is flush against its own label.** `.textCardsLabelWrap` spans both columns with `flex-grow`, so "Deck label - shown on the card backs" ends ~2px before the input box; every other row in the grid has clear separation. A `gap`/`margin-left` on `.textCardsLabel` would line it up with the column above.
- **Inconsistent "copies" control between the two new sections.** `properties.js:1401` renders `Copies of each card: <input width:50px>` as a bare inline row; `properties.js:1530` renders `Copies of each card` (no colon) at 60px inside the design grid. Same control, two spellings and two widths — pick one.
- **30×40 thumbnails are too small to verify an image.** `.imagePairEntry img` at 30×40 (`sidebarModules.css:191`) makes card art unrecognisable, which defeats the "see the pairing before adding" goal. ~48×64 would still keep rows compact.
- **Big dead gap between file name and its delete button.** `.imagePairEntry span { flex-grow: 1 }` stretches to the dialog width, throwing the ✕ ~500px away from the row it belongs to. Consider capping the row (`max-width: 420px`) or placing the delete right after the name.
- **Empty state of the upload sections is a blue header, whitespace, then a right-aligned button** ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1532316493449138176/ui3071-imagepairs-empty.png)). A one-line placeholder in the empty `.imagePairList` ("No front images yet") would explain the gap and give the list a visible identity.
- Uploading the same file twice adds two identical rows (`back1.png` at 1 and 2) with nothing but the ordinal to tell them apart — worth a subtle duplicate hint, since the numbers are load-bearing here.

- **The two upload modes are hard to tell apart in the list.** "Upload one image per card or tiled images" vs "Upload a front and a back image for every card" (editor.html:239-241) both start with "Upload"; a newcomer can't tell that the first gives the whole deck *one shared back*. Suggest "Upload one image per card (one shared back for the deck)" and "Upload a front **and** a back image per card" — the difference is the back, so lead with it.
- **`properties.js:1531` — "Deck label - shown on the card backs" is inaccurate**: the code also prints the label small along the bottom of every *front*. Say "shown large on the backs and small on the fronts" so the footer strip in the result isn't a surprise.
- **No units on the size fields.** "Font size 16 / Card width 150 / Card height 233" — the empty-deck mode right above shows sizes as "103 × 160", so a first-timer will wonder what unit these are. Append `px` (or a small suffix span) and consider offering the same size presets as the empty-deck chips.
- **The text section has no count feedback**, while the image section has a status line. After pasting 60 lines you get no "60 card types → 120 cards" — and with copies set, `addDeckWithCards` may then hit a 500-card `confirm()` out of nowhere. A one-line status mirroring `.imagePairStatus` ("60 lines → 60 card types, 120 cards") would make the copies multiplier and that confirm predictable.
- **The green button never states what it will create.** "Add to game" is fine, but "Add 3 card types (6 cards)" would be a clearer commitment for someone who has no idea what a "card type" is yet — the wizard uses that term in the status line and in card names without ever defining it. A one-clause gloss in the intro paragraph ("each *card type* is one distinct card design; copies are duplicates of it") would go a long way.
- **The textarea renders in the browser's default monospace** (nothing sets `font-family` on `.textCardsInput`, sidebarModules.css:222) while the cards render bold sans — so what you type looks nothing like what you get, and line wrapping in the box says nothing about wrapping on the card. `font-family: inherit` would make the box match the dialog and set better expectations.
- **The mismatch message doesn't say what to do next in the shared-back case.** "3 fronts but 2 backs - upload the same number of each, or a single back for all cards" is good, but the "single back" escape hatch is only discoverable *after* you've already got a mismatch (and once you have 2 backs, going back to 1 means deleting one). Mentioning it on the "Card backs" subheader ("upload one per card, or a single back for all") would surface it up front.

VTT-AI-Agent:
Thanks — this was a genuinely useful pass. Almost all of it is now fixed in 1bf1d0c1; the two exceptions are called out at the bottom.

Screenshots of the result (1400×900, same viewport you measured in):
[text cards](https://agent.virtualtabletop.io/reports/pr/1532145112908435497/pr3071-text-cards-after.png) · [front/back images](https://agent.virtualtabletop.io/reports/pr/1532145112908435497/pr3071-image-pairs-after.png)

## Big layout improvements

**1. The mode list eats the dialog / panel below the fold** — fixed, both ways you suggested.

* Picking a mode collapses `.deckEditorNewDeckModes` to just the chosen option plus a **Change** button that brings the list back; the dialog still opens with all eight visible, so nothing is hidden from a first-time user.
* The action row of every panel (`.goButton` and `.deckEditorNewDeckButtonBar`) is now `position: sticky; bottom` inside the dialog, so "Add to game" / "Create empty deck" / "Browse the public library" are always reachable — this fixes the existing `images`/`traditional`/`tts` panels too. One catch worth recording: `.tune.editorModule { overflow-x: hidden }` turned the reused module into its own scroll container and silently killed the stickiness, so `.deckEditorNewDeckModule` overrides it back to `overflow: visible`.
* Text cards: the design grid and the preview now sit side by side.

Measured at 1400×900, text mode with three lines and a deck label: `scrollHeight 840` vs `clientHeight 810` (was 1117/810), and the green button is on screen either way. At 1100×700 the dialog scrolls, but the status line and the button stay pinned to its bottom edge.

**2. The live preview is too small to preview anything** — fixed. It renders **one** card front, upright, at its real pixel size (scaled down only above 200px), on the dialog background instead of the checkerboard, and the section is labelled "Preview (first card)". The fan is gone rather than kept as a secondary view — a second, less legible copy of the same card didn't seem to earn its space.

Because the single card is now rendered directly, the `shuffleCardTypes` parameter added to `renderWidgetButton` for review 1's finding 2 had no callers left, so it's reverted; the preview can't reshuffle any more because it only ever shows the first card.

The overflow you spotted is fixed at the source too, not just warned about: the card text object gets `overflow: hidden`, so a long line stops at its box instead of running over the rounded edge and the deck-label footer. [Before/after on the worst case.](https://agent.virtualtabletop.io/reports/pr/1532145112908435497/pr3071-text-cards-after.png) I didn't add the "some texts won't fit at 16px" hint — with the real-size preview the clipping is visible while you type, and guessing the threshold reliably would need a measuring pass per line.

**3. Front/back pairing shown as two lists that never meet** — fixed with the 2-column grid rather than the single merged list. The grid is filled row by row (both headers, both hints, both lists, both upload buttons), so front *n* and back *n* are literally on the same line and the ordinals stop being the only evidence of the pairing. I stopped short of one merged row per card because the two sides are independently uploadable and deletable and the shared-back/mismatch states have no natural "row" — but if you'd still prefer the merged list, say so and I'll do it.

* **`#ff8080` contrast** — fixed, now `var(--negativeInput)`.
* **Error message mis-placed** — fixed, the status is inside the action row, immediately left of the button it explains (same for the new text-cards status).
* **Disabled `Add to game` doesn't look disabled** — fixed in `layout.css` with `button[icon]:disabled { background-color: #808080 }`, which beats the `.green`/`.red` rules by source order. Applies to every colored icon button in the app, which is what you'd want.
* **Design grid `space-between`** — fixed, each input hugs its own label (with a larger column gap) instead of being pushed next to the next field's label.
* **White swatch reads as an empty text field** — fixed, color inputs are 40×24 rather than 60px wide.
* **Deck-label input flush against its label** — fixed; the label is now just "Deck label" and the explanation moved to a small hint line spanning the grid.
* **Inconsistent "copies" control** — fixed, both sections use `Copies of each card` with a 60px input.
* **30×40 thumbnails** — now 48×64.
* **Dead gap before the delete button** — fixed, the file name no longer stretches, so the ✕ follows it directly.
* **Empty state of the upload sections** — fixed, the empty lists say "No front images yet" / "No back images yet", and each column has a one-line hint under its header.
* **Duplicate uploads look identical** — not changed. Two files can legitimately have the same name (they come from separate pickers) and the position, not the name, is what pairs them, so a "duplicate" hint would fire on a legitimate case as often as a mistaken one. The thumbnails at 48×64 make an accidental double upload easier to spot than before.

* **Two "Upload…" modes hard to tell apart** — reworded to "Upload one image per card or tiled images — one shared back for the whole deck" and "Upload a separate front and back image for every card".
* **"shown on the card backs" is inaccurate** — now "shown large on the card backs and small along the bottom edge of the fronts".
* **No units** — `px` suffixes on font size, card width and card height.
* **No count feedback in the text section** — added: `3 card types × 2 = 6 cards.` next to the button, updating live, so the 500-card confirm can't come out of nowhere.
* **The green button never states what it will create** — kept as "Add to game" for consistency across all eight panels; the count line next to it carries the same information. "card type" is now glossed in the intro paragraph ("one distinct card design, of which the copies below are duplicates").
* **Textarea in default monospace** — fixed, `font-family: inherit`.
* **Shared-back escape hatch only discoverable after a mismatch** — surfaced up front as the hint under the "Card backs" header: "One per card, or a single one for every card."

Verified locally: `npm test` (86 passed) and the full `tests/testcafe/editor.js` in Chromium (17 passed). The text-cards reference hash moved to `94d9f0542c71541a5e20ae14a37499b1` because of the `overflow: hidden` on the text objects.

VTT-AI-Agent (commented):
## TL;DR

The latest fixes are in good shape and all current CI/CodeQL checks pass, but the new generic large-deck confirmation regresses the existing image-upload flow: its string-valued counts are concatenated rather than summed, so uploading only four one-copy fronts produces a warning claiming it will create `01111` cards. Coerce each count to a number in the reducer and cover the original image mode with a regression test.

## Finding

### 1. The card-count guard concatenates counts from the existing image flow — `client/js/editor/sidebar/properties.js:1755`

`deckImages` stores the number inputs' `.value` directly in `counts` (`properties.js:1295` and `properties.js:1302`), so those values are strings. In the new reducer, the first addition changes `sum` from `0` to `"01"` and every later count is concatenated: four ordinary front images with the default one copy each yield `"01111"`, which is then coerced to `1111` by `> 500`. As a result, a tiny four-card deck shows “This creates 01111 cards,” and declining the spurious warning prevents the deck from being added.

Please coerce at the aggregation boundary, for example with `sum + Number(counts?.[cardType] ?? 0)`. The card-creation loop itself still creates the correct number because its numeric comparison coerces each individual string, which is why the existing TestCafe suite does not expose this. A small regression test through the original `images` mode would protect the new shared guard across both numeric and DOM-string callers.

## Scope checked

The changed files are confined to editor HTML/JS/CSS and TestCafe coverage; no `client/js/widgets/`, `client/js/compute.js`, or `client/js/serverstate.js` files are touched, so this does not alter engine behavior for existing games.

The `beta` label was already present; I left it in place because this editor-only confirmation bug is safe and useful for player testing and cannot corrupt games or crash the server.

VTT-AI-Agent:
Confirmed and fixed in d9946963 — thanks, that was a real regression of the existing `images` flow.

**Finding 1 — the card-count guard concatenated counts (`properties.js:1755`)**

Fixed. `addDeckWithCards` now coerces once, at the aggregation boundary, and both the guard and the card-creation loop go through it:

```js
// Some callers fill counts straight from a number input, where .value is a string - coerce so that the
// counts are summed rather than concatenated.
const copiesOf = cardType => counts ? Number(counts[cardType]) || 0 : 1;
```

The creation loop's `const count = counts ? counts[cardType] || 0 : 1;` was replaced by `copiesOf(cardType)` too — it happened to behave because `i <= count` and `count > 1` coerce, but there was no reason to keep the string in a numeric variable now that a helper exists.

**Regression test through the original `images` mode**

Added `Deck editor: a few uploaded card fronts are added without a large-deck confirmation` in `tests/testcafe/editor.js`: it stubs `uploadAsset` with four fronts (default one copy each), installs `setNativeDialogHandler(() => false)`, clicks **Add to game**, and then asserts three things — the deck opened in the strip with 4 card types, `getNativeDialogHistory()` is empty, and 4 card widgets exist. Declining the dialog is what makes it bite: with the old reducer the spurious `01111` confirm fires, gets declined and the deck is never created.

Verified it isn't vacuous — with the coercion reverted the test fails (`expected 0 to deeply equal 4`), and passes with it back. Full local run: `npm test` 86 passed, `tests/testcafe/editor.js` 18 passed in Chromium (no reference hash moved — the new test asserts widget counts rather than state).

VTT-AI-Agent (commented):
## TL;DR

Nice, well-commented addition — both wizard sections work, and all three new TestCafe tests pass locally for me (`editor.js`, headless Chromium, server on a local port). One real defect: the text-card **Card width / Card height / Font size** fields are read with `+value || fallback` and never clamped, so a typed out-of-range value goes straight into the game — I reproduced a deck with `cardDefaults.width: -50` and a holder of `width: -42`, which is exactly the class of input the new `copiesFromInput()` helper was written to guard against, just not applied here.

## Findings

### 1. Text-card design numbers bypass the clamping the copies field gets — `client/js/editor/sidebar/properties.js:1562`

```js
const number = (selector, fallback)=>+$(selector, design).value || fallback;
```

`min`/`max` on a number input only constrain the spinner — the same reason `copiesFromInput()` (properties.js:1369) exists. But width/height/fontSize don't go through it:

* Typing `-50` into "Card width" (declared `min=20 max=600`) produces `cardDefaults: { width: -50 }`, and `addDeckWithCards` then derives `holderWidth = cardWidth + 8 = -42` and `x = Math.round(800 - (-42)/2)`. Reproduced in a browser against this branch:
  ```
  deck.cardDefaults = { width: -50, height: 233, fontSize: 16 }
  holder = { width: -42, height: 244, x: 821 }
  ```
  The copies field in the same panel correctly clamped `0` → `1` on the same run, so the inconsistency is visible side by side.
* Typing `5000` is accepted despite `max=600`.
* `0` (and an emptied field) silently falls back to the declared default rather than to the min, so "Font size 0" quietly becomes 16 with no feedback.

Suggest generalising the existing helper — something like `numberFromInput(input, fallback)` that reuses the element's own `min`/`max` — and using it for width, height, fontSize and both copies fields. That also removes the duplicated `150`/`233`/`16` literals, which currently have to stay in sync with the `value=` attributes in the HTML template at properties.js:1541.

### 2. Card width can be read before the first front image has decoded — `client/js/editor/sidebar/properties.js:1509`

```js
const firstFront = $('img', frontList);
const cardWidth = firstFront && firstFront.naturalHeight ? Math.round(...) : 103;
```

`naturalHeight` is `0` until the browser has decoded the image, so clicking "Add to game" quickly after a large upload silently yields a 103px-wide deck instead of the image's aspect ratio — with no indication that the fallback was used. The new TestCafe test acknowledges the race by explicitly waiting (`.expect(firstFrontHeight()).eql(60)`), but a real user has nothing making them wait.

Cheap fix: capture `naturalWidth/naturalHeight` in the thumbnail's `onload` and store it on the entry (`fronts[0]`), so the value doesn't depend on when the button is clicked. (`deckImages` has the same weakness, but it isn't made worse there by a bulk upload of many files at once.)

### 3. Front/back card type names aren't sanitized, unlike the text-card ones — `client/js/editor/sidebar/properties.js:1464`

`deckTextCards` deliberately strips card type names down to `[A-Za-z0-9 ]` and 30 characters "because card type names end up in the card widget ids" (properties.js:1594). `deckImagePairs` takes the raw file name minus its extension, so `Königs–Karte (final v2) copy 3.png` becomes both a card type name and part of every card's widget id, unbounded in length. Given the two flows landed in the same PR with the same stated concern, applying the same sanitizing (or dropping it from both) would be more consistent — and keeps ids readable in the JSON editor.

### 4. The untested branches are the ones most likely to break

The image-pair test covers only the happy N-fronts/N-backs path. Not covered:

* the single-shared-back shortcut (`backs.length == 1` with several fronts) — the branch at properties.js:1462 and the corresponding status text;
* the mismatch state (3 fronts, 2 backs) keeping "Add to game" disabled;
* the delete button re-pairing the lists.

These are all pure client-side state and cheap to assert on with the existing `stubUploadOf` helper. The >500-card confirmation is likewise only tested in the negative (`getNativeDialogHistory()).eql([])`); a positive case would pin down that `copiesOf`'s `Number()` coercion actually feeds the sum.

### 5. Unrelated no-op refactor — `client/js/editor/sidebar/properties.js:6639`

```js
-      for(const cardType of shuffleArray(Object.keys(widget.get('cardTypes'))).slice(0, 5)) {
+      const cardTypeNames = Object.keys(widget.get('cardTypes'));
+      for(const cardType of shuffleArray(cardTypeNames).slice(0, 5)) {
```

`cardTypeNames` is used exactly once and the behaviour is identical — looks like a leftover from an earlier iteration. Worth dropping to keep the diff to one concern.

### 6. Two CSS rules widen beyond the new sections (low)

* `#deckEditorNewDeckPanel .goButton { position: sticky; … }` and `.editorModule .goButton.buttonBar { display: flex; justify-content: flex-end; }` (deckeditor.css:733, sidebarModules.css:30) also restyle the *existing* `deckImages` action row, which holds two buttons ("Upload fronts…" + "Add to game"), and the TTS/traditional ones.
* `.deckEditorNewDeckModule.tune.editorModule { overflow: visible }` removes the module's own overflow containment for every wizard mode, not just the two new ones.

Both are plausibly intended and the comments explain the *why*, but they're worth a quick eyeball on the TTS-import and one-image-per-card panels before merge — I only exercised the two new sections and the images one.

* Card text and the deck label are rendered through `text` objects, which go to `textContent` (card.js:227), not the `html` object path — no injection surface from typed card text.
* `Number(counts[cardType]) || 0` in `addDeckWithCards` correctly fixes the string-concatenation hazard the `deckImages` flow would otherwise hit in the new total; the existing `for(let i=1; i<=count; ++i)` loop is unaffected by the change.
* The `(2)` suffix cannot collide with the `<card type> <copy number>` ids in these two flows, since the copy count is uniform across all card types.
* No widget or routine properties are added, so `validator/` and `jeCommands` genuinely need no updates.
* `confirm()` for the >500-card guard matches how the rest of the editor asks (toolbox.js:59, deckeditor.js:3290, …).

The `beta` label was already on this PR and I left it as is — the diff touches only editor UI (no `client/js/widgets/`, `compute.js` or `serverstate.js`), it can't affect how existing games run, and the issues above are confined to out-of-range typed input in a new wizard, so it's safe for players to try.

VTT-AI-Agent:
Thanks — finding 1 was real and is fixed in c351577a, together with 2, 3 and 5 and part of 4. Details below, including the one point I checked but didn't change.

**1. Text-card design numbers bypass the clamping (`properties.js:1562`)** — fixed, exactly the way you suggested. `copiesFromInput()` is generalised into `numberFromInput(input)`, which reads the range off the element itself:

```js
numberFromInput(input) {
  const min = +input.min || 1;
  return Math.min(+input.max || Infinity, Math.max(min, Math.round(+input.value) || min));
}
```

Width, height, font size and both copies fields now go through it, so `-50` and `5000` in "Card width" become 20 and 600, and an emptied field falls back to the input's `min` rather than to a duplicated default — which also removes the `150`/`233`/`16` literals that had to stay in sync with the HTML template.

**2. Card width read before the first front image has decoded (`properties.js:1509`)** — fixed at the source: each upload now measures its own image and stores the aspect ratio on the entry, so the value doesn't depend on when the button is clicked.

```js
const probe = new Image();
probe.onload = _=>entry.aspectRatio = probe.naturalWidth / probe.naturalHeight;
probe.src = mapAssetURLs(imagePath);
```

`Add to game` reads `fronts[0].aspectRatio` and still falls back to 103 if even that hasn't loaded yet. I left `deckImages` alone, as you suggested.

**3. Unsanitized card type names in the image-pair flow (`properties.js:1464`)** — fixed by applying the same sanitizing to both flows, extracted into `cardTypeName(name, fallback)`. `Königs–Karte (final v2) copy 3.png` now yields the card type `K nigs Karte final v2 copy 3` (capped at 30 characters) instead of the raw file name.

**4. Untested branches** — added `Deck editor: mismatched and shared card backs in the new deck wizard`, which walks the states in one go: 3 fronts + 2 backs asserts the mismatch text, the `imagePairMismatch` class and a disabled "Add to game"; deleting one back re-pairs the lists into the single-shared-back state; then "Add to game" is clicked and all three card types are asserted to carry the same `backImage`. The text-cards test also now types `-50` into "Card width" and asserts the real-size preview card clamps to `20px` before restoring 150 (so its state hash is unchanged). I did not add a positive >500-card case — driving it needs either 501 pasted lines or a copies value the input's own `max=99` forbids, and the sum it would exercise is the one the negative test already pins down.

**5. Unrelated no-op refactor (`properties.js:6639`)** — reverted, the line is back to what it was on main.

**6. Two CSS rules widening beyond the new sections** — checked in a browser (1400×900) rather than changed, and both are fine:

* `images`: the action row holds "Upload fronts…" + "Add to game" side by side, right-aligned, panel `scrollHeight == clientHeight` (584) so the sticky rule never engages.
* `tts`: single "Add to game", right-aligned, 307/307.
* `traditional` and `generator` don't use `.goButton` at all — their button is plain and still sits left-aligned exactly as on main, so neither rule reaches them.
* [screenshot of the images panel](https://agent.virtualtabletop.io/reports/pr/1532145112908435497/pr3071-panel-images.png) · [tts](https://agent.virtualtabletop.io/reports/pr/1532145112908435497/pr3071-panel-tts.png) · [traditional](https://agent.virtualtabletop.io/reports/pr/1532145112908435497/pr3071-panel-traditional.png)

The `overflow: visible` override is scoped to `.deckEditorNewDeckModule`, i.e. the wizard's panel only; it exists because `.tune.editorModule { overflow-x: hidden }` turns the panel into its own scroll container and silently kills the sticky action row.

Verified locally: `npm test` 86 passed, `tests/testcafe/editor.js` 19 passed in Chromium — the text-cards reference hash did not move, which is the check that the clamping changes nothing for in-range input.

96LawDawg:
@VTT-AI-Agent, in the add a deck wizard, group the options into 3 categories: new blank deck , use existing deck, or create custom deck. They should each be behind expander drop downs. The first one should be pre-selected and opened to use a new blank deck. The other two should start closed. Opening one closes the other. Do anything else you think is necessary to make the add a deck wizard simpler and easier to understand.

VTT-AI-Agent:
@96LawDawg Done — pushed in be51a094.

The wizard's eight options are now three collapsible sections, and only one is open at a time (opening one closes the others):

| | |
|---|---|
| **New blank deck** — the empty deck flow. Pre-selected and open when the dialog opens, so the card size / deck id form is right there. | blank |
| **Use an existing deck** — public library, traditional decks, Tabletop Simulator import. | existing |
| **Create a custom deck** — suits and ranks, one image per card, a front and a back image per card, text cards. | custom |

Other things that came out of making it easier to understand:

- Every option is now a short **title** plus a one-line explanation underneath, instead of one long sentence per radio button (e.g. "Text cards" / "One card per line of text").
- The form for the option you pick is rendered **inside the open section**, directly under the options it belongs to, instead of at the bottom of the dialog.
- The old "collapse the list once you pick something / Change to get it back" mechanic is gone — the sections keep the dialog short on their own, so there is one less thing to learn.
- The section with only one way to create a deck (the blank one) shows no radio button at all — opening it *is* the choice.

Before / after of the whole dialog: [before](https://agent.virtualtabletop.io/reports/pr/1532145112908435497/wizard-before.png) · [after](https://agent.virtualtabletop.io/reports/pr/1532145112908435497/wizard-blank-open.png), and picking [Text cards](https://agent.virtualtabletop.io/reports/pr/1532145112908435497/wizard-custom-text.png) inside a section.

`npm test` (86 jest tests) and the full `tests/testcafe/editor.js` suite (19 tests) pass locally in headless Chromium; the four wizard tests got one extra click to open the section first.

ArnoldSmith86:
Add a way to insert linebreaks.

VTT-AI-Agent:
Done — pushed in 738fe88b.

Line breaks were impossible in two places in the text-cards section, and both are now available:

**Card texts.** A radio row right under the textarea picks how the typed text is cut into cards:

- **One card per line** (default, unchanged) — right for a pasted list.
- **Blank line between cards** — spends an empty line on the separator and gives the line breaks back to the card text, so a card can run over several lines.

The placeholder changes with the mode so the format is shown rather than explained, and each line is still trimmed, so indentation and doubled/trailing blank lines don't produce stray cards or leading spaces.

**Deck label.** It is a `textarea` now instead of a single-line input, so a label like `Line\nBreak\nDeck` is shown across three lines on the card backs — exactly how the *Cards Against Humanity* backs in the library look. The small footer on the fronts has no room for that, so there a multi-line label is flattened into one line.

| | |
|---|---|
| The section with a two-line card and a three-line deck label | wizard |
| The resulting card back | back |

The line breaks need nothing on the rendering side: `.widget` is already `white-space: pre-wrap`, so a `\n` in a card's `text` property renders as a line break on the card, in the deck editor and in the room.

New TestCafe test `Deck editor: text cards with line breaks in the new deck wizard` drives blank-line mode with an indented line, a doubled separator and a trailing blank line, and asserts the two resulting `cardTypes`, the multi-line back label and the flattened front footer. `npm test` (86 jest tests) and the full `tests/testcafe/editor.js` suite (20 tests) pass locally in headless Chromium — including the existing text-cards reference hash, which is the check that one-card-per-line mode is untouched.

ArnoldSmith86 (approved):

PR #3073, opened 2026-07-30
Deck wizard: any icon as a suit, more card designs, optional holder and reset button

VTT-AI-Agent:
## Problem

The "Add New Deck" wizard's **Custom deck of cards (define suits and ranks)** flow was quite limited:

- suits could only be one of **nine hard-coded icons**, and because the suits were a map keyed by that icon, **each icon could only be used once** — no red *and* purple hearts.
- only **five card designs** were offered.
- every generated deck carried card type properties **only one design reads**: `rankImage` was written for all designs, and two designs added an empty `cardDefaults` object. The playing-card design additionally needs the pip layout (`suit-P…`/`suit-S…`), which is why that one "uses way more properties than the others" — even where the properties did nothing.
- the middle picture of the playing-card design was built as `/i/game-icons.net/<suit>.svg`, so ranks without a pip layout (jokers, tarot trumps, words) only worked for game-icons suits.

And for **every** mode of the wizard, a holder, a pile and a "Recall & Shuffle" button were added unconditionally — there was no way to get just the deck and its cards.

## Fix

**Suits (custom deck)**

- One compact row per suit: **icon picker**, name, color, ranks, remove. The icon picker is the properties sidebar's existing `IconInput`, so **any icon works as a suit** — game-icons.net, Material Symbols, emoji, VTT symbols, uploaded images — with search, library filters and the full symbol picker behind "Show all". The nine icons that used to be the only choice stay one click away as "Add a suit:" chips. Each row shows its icon in the suit's color.
- Suits are now a **list**, so **the same icon can be used for several suits**. Each suit has its own name (what its card types are named after, e.g. `7 of purple hearts`); duplicate names get numbered so card types stay unique. A new suit whose icon+color combination already exists picks the next unused palette color, so a second hearts suit is visibly different right away.

**Card designs**

- Four more designs, adapted from simple decks in the public library: **Minimal** (Feta, Escrime), **Corner index** (playing cards without the pip layout, so it works for any rank set), **Round token** (CONTEST, District Development) and **Square tile** (Scotland Hills). Their backs are suit-independent, so face-down cards stay indistinguishable.
- Each design tile now carries the design's name.
- A design **declares which card type properties it reads**, and the generator only writes those: the pip layout and the court card pictures are exclusive to *Playing cards*; every other design gets just `suit`, `suitColor` and `rank`. The empty `cardDefaults` objects are gone too.
- *Playing cards* draws the middle of a card that has no pip layout from a new `rankIcon` card type property as an **icon** instead of a stretched game-icons URL, so jokers/trumps/words also work for Material Symbols and emoji suits.

**Whole wizard: what gets added around the deck**

Two options, both on by default (exactly what the flows did before), applying to all six modes — empty deck, public library, traditional, custom, images and TTS import:

- **Holder for the deck, with the cards in it**
- **Reset button that recalls, flips to back and shuffles** (a reset button recalls *into* a holder, so it follows the holder option)

Without a holder, the deck widget and a pile holding all the cards are placed on the table instead.

**Cleanups that fall out of this**

- The custom flow's copy of the holder/button/deck/pile code is gone; it uses `addDeckWithCards` like the other flows, and all five deck-creation flows build the Recall & Shuffle button through one `deckResetButton()`.
- Rank lists tolerate spaces after the commas (`2-10, J, Q, K, A` used to produce a rank called `" J"`), and re-ticking "Same ranks for each suit" now unifies the ranks it claims are shared.
- `IconInput`/`ImageInput`'s "Show all" reopens the overlay it was invoked from (and shows the symbol picker full size above the deck editor instead of squeezed into its card column).
- Preview refreshes are coalesced and the card types shared between designs, so the nine previews are not rebuilt per keystroke of a rank list.

No widget property or routine parameter was added or renamed (`rankIcon` is a free-form card type property of a deck, which the validator accepts as `any`), so `validator/` and the JSON editor's `jeCommands` need no update.

## Testing

- `npm test` (jest) passes.
- TestCafe `tests/testcafe/editor.js` passes in headless Chromium (15/15), including the deck editor state-hash tests — **the default output of the wizard is unchanged**, byte for byte.
- Drove every mode in a real browser with the options on and off: custom (all designs, plus a Material Symbols suit and hearts twice in two colors), empty deck, traditional, and the public library both from the wizard (options apply) and from the add-widget overlay (unchanged: holder + button).
- Ran `node validator/validate_gamefile_node.js` on a game built with the new designs — no problems reported for the decks.

Playable demo (five decks, one per design, five suits with hearts twice in different colors):
[custom-decks-demo.vtt](https://agent.virtualtabletop.io/reports/pr/1532171279962996877/custom-decks-demo.vtt)

| the dialog | the designs, and a Material Symbols suit |
| --- | --- |
| dialog | designs |

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3073/pr-test (or any other room on that server)

### DIFF
--- a/client/css/editor/deckeditor.css
+++ b/client/css/editor/deckeditor.css
@@ -707,6 +707,32 @@ body.deckEditorRoomVisible #deckEditorMainCol > .overlay {
   cursor: pointer;
 }
 
+/* Applies to every mode: what gets added around the new deck. */
+.deckEditorNewDeckOptions {
+  display: flex;
+  flex-direction: column;
+  gap: 6px;
+  margin: 0 0 16px 0;
+  padding-top: 12px;
+  border-top: 1px solid #0002;
+}
+
+.deckEditorNewDeckOptionsLabel {
+  font-size: 11px;
+  font-weight: bold;
+  text-transform: uppercase;
+  letter-spacing: 0.3px;
+  opacity: 0.7;
+}
+
+.deckEditorNewDeckOptions label {
+  display: flex;
+  align-items: center;
+  gap: 8px;
+  cursor: pointer;
+  font-size: 13px;
+}
+
 #deckEditorNewDeckPanel:not(:empty) {
   border-top: 1px solid #0002;
   padding-top: 16px;
--- a/client/css/editor/deckeditor.css
+++ b/client/css/editor/deckeditor.css
@@ -464,6 +464,18 @@ body.deckEditorRoomVisible #deckEditorMainCol > .overlay {
   z-index: 20;
 }
 
+/* Opening the picker from an icon property inside one of the dialogs above parks it next to that dialog
+   (pickSymbolKeepingOverlay), where the rule above no longer applies and it would render behind #deckEditor
+   (z-index 1). Float it over the whole editor instead: above the sidebar (2), the toolbar (3) and the
+   dialog it was opened from (20). */
+#symbolPickerOverlay.symbolPickerAboveEditor {
+  position: fixed;
+  inset: 0;
+  width: auto;
+  height: auto;
+  z-index: 30;
+}
+
 #deckEditorExportOverlay {
   position: fixed;
   top: var(--editToolbarHeight);
@@ -707,14 +719,16 @@ body.deckEditorRoomVisible #deckEditorMainCol > .overlay {
   cursor: pointer;
 }
 
-/* Applies to every mode: what gets added around the new deck. */
+/* Applies to every mode, so it is set apart from the mode list above and the picked mode's panel below. */
 .deckEditorNewDeckOptions {
   display: flex;
   flex-direction: column;
   gap: 6px;
   margin: 0 0 16px 0;
-  padding-top: 12px;
-  border-top: 1px solid #0002;
+  padding: 10px 12px;
+  border: 1px solid #0002;
+  border-radius: 8px;
+  background: #0000000a;
 }
 
 .deckEditorNewDeckOptionsLabel {
@@ -731,6 +745,24 @@ body.deckEditorRoomVisible #deckEditorMainCol > .overlay {
   gap: 8px;
   cursor: pointer;
   font-size: 13px;
+}
+
+/* the reset button needs a holder to recall the cards into, so without one its option is off, not just
+   greyed out while still looking ticked */
+.deckEditorNewDeckOptions label:has(input:disabled) {
+  opacity: 0.5;
+  cursor: default;
+}
+
+/* what unticking the holder means, shown only once it is unticked */
+.deckEditorNewDeckOptionsHint {
+  display: none;
+  font-size: 12px;
+  opacity: 0.85;
+}
+
+.deckEditorNewDeckOptions:has(#deckEditorNewDeckHolder:not(:checked)) .deckEditorNewDeckOptionsHint {
+  display: block;
 }
 
 #deckEditorNewDeckPanel:not(:empty) {
--- a/client/css/editor/sidebarModules.css
+++ b/client/css/editor/sidebarModules.css
@@ -118,28 +118,60 @@
 }
 
 
+/* The deck creation flows disable their "Add to game" until something is picked. button[icon].green
+   (specificity 0,2,1) beats the generic button:disabled (0,1,1), so without this the greyed-out state never
+   applies and a button that does nothing still looks clickable. */
+.tune.editorModule button[icon].green:disabled,
+.tune.editorModule button[icon].green:disabled:hover {
+  background-color: #808080;
+  color: #d3d3d3;
+}
+
+
 /* custom deck: one compact row per suit (icon, name, color, ranks, remove) */
 
 .tune.editorModule .deckGeneratorHint {
   margin: 0 0 8px 0;
   font-size: 12px;
+  opacity: 0.85;
+}
+
+/* One grid for the whole list, so the header labels and every row share the same columns. The rows and their
+   pickers are display: contents, i.e. their children are the grid items. */
+.tune.editorModule .deckGeneratorSuits {
+  display: grid;
+  grid-template-columns: auto minmax(60px, 1fr) auto minmax(80px, 2fr) auto;
+  align-items: center;
+  gap: 4px 6px;
+}
+
+.tune.editorModule .deckGeneratorSuitHeader,
+.tune.editorModule .deckGeneratorSuit,
+.tune.editorModule .deckGeneratorSuitPickers {
+  display: contents;
+}
+
+.tune.editorModule .deckGeneratorSuitHeader > span {
+  font-size: 11px;
+  font-weight: bold;
+  text-transform: uppercase;
+  letter-spacing: 0.3px;
   opacity: 0.7;
 }
 
-.tune.editorModule .deckGeneratorSuit {
-  display: flex;
-  align-items: center;
-  gap: 6px;
-  margin-bottom: 4px;
+/* An open picker sits right below the row it belongs to and spans the full width; a closed one is display:
+   none and takes no grid row at all. */
+.tune.editorModule .deckGeneratorSuitPickers > .propertyPicker {
+  grid-column: 1 / -1;
+  margin-bottom: 0;
 }
 
 .tune.editorModule .deckGeneratorSuit > .propertyInput {
-  flex: 0 0 auto;
   margin: 0;
 }
 
-/* the row's icon in the suit's color (see showIconInSuitColor): the icon masks the color */
-.tune.editorModule .deckGeneratorSuit .deckGeneratorSuitTint {
+/* an icon chip in a suit color (see tintIconChip): the icon masks the color */
+.tune.editorModule .deckGeneratorSuitTint {
   width: 100%;
   height: 100%;
   background: currentColor;
@@ -147,14 +179,24 @@
   mask: var(--mask) center / contain no-repeat;
 }
 
-.tune.editorModule .deckGeneratorSuitName {
-  flex: 1 1 80px;
-  min-width: 60px;
+.tune.editorModule .deckGeneratorSuitName,
+.tune.editorModule .deckGeneratorSuitRanks {
+  width: 100%;
+  min-width: 0;
 }
 
-.tune.editorModule .deckGeneratorSuitRanks {
-  flex: 2 1 130px;
-  min-width: 80px;
+/* the name this suit's card types actually get differs from what is typed (see showResolvedSuitNames) */
+.tune.editorModule .deckGeneratorSuitName.deckGeneratorSuitRenamed {
+  border-color: var(--VTTred, #c0392b);
+}
+
+/* destructive, so red rather than the accent color the rest of the row uses */
+.tune.editorModule .deckGeneratorSuitRemove:not(:disabled) {
+  background: var(--VTTred, #c0392b);
+}
+
+.tune.editorModule .deckGeneratorSuitRemove:hover:not(:disabled) {
+  background: #922b21;
 }
 
 .tune.editorModule .deckGeneratorSuitAdd {
@@ -170,8 +212,10 @@
   opacity: 0.7;
 }
 
+/* these are suit shortcuts, not transparency swatches, so no checkerboard behind them */
 .tune.editorModule .deckGeneratorSuitAdd .propertyValueChip {
   cursor: pointer;
+  background: #fff;
 }
 
 .tune.editorModule .deckGeneratorSuitAdd .propertyValueChip:hover {
@@ -182,7 +226,37 @@
   display: flex;
   align-items: center;
   gap: 6px;
+  margin-top: 8px;
   font-size: 12px;
+}
+
+/* Card design gallery: one tile per design, filling the dialog in as many columns as fit so no design ends up
+   cut off or alone on a row. The name goes below the preview instead of on top of the artwork. */
+.tune.editorModule .deckGeneratorDesigns {
+  display: grid;
+  grid-template-columns: repeat(auto-fill, minmax(156px, 1fr));
+  gap: 10px 6px;
+  margin-bottom: 12px;
+}
+
+.tune.editorModule .deckDesignTile {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  gap: 4px;
+}
+
+/* a card design is not a transparency preview, so a flat neutral reads better than the checkerboard */
+.tune.editorModule .deckDesignButton {
+  margin: 0;
+  background: #ededed;
+}
+
+.tune.editorModule .deckDesignLabel {
+  font-size: 12px;
+  font-weight: bold;
+  text-align: center;
+  line-height: 1.2;
 }
 
 
@@ -232,9 +306,8 @@
   min-width: 0;
 }
 
-/* caption bar of a preview tile: the card count of a TTS deck, the name of a card design */
-.tune.editorModule .ttsDeckButton .deckCardCount,
-.tune.editorModule .deckDesignButton .deckDesignLabel {
+/* caption bar of a preview tile: the card count of a TTS deck */
+.tune.editorModule .ttsDeckButton .deckCardCount {
   position: absolute;
   bottom: 0;
   left: 0;
--- a/client/css/editor/sidebarModules.css
+++ b/client/css/editor/sidebarModules.css
@@ -118,21 +118,71 @@
 }
 
 
-.tune.editorModule .suitProperties {
-  height: 160px;
+/* custom deck: one compact row per suit (icon, name, color, ranks, remove) */
+
+.tune.editorModule .deckGeneratorHint {
+  margin: 0 0 8px 0;
+  font-size: 12px;
+  opacity: 0.7;
 }
 
-.tune.editorModule .suitProperties > button.widgetSelectionButton {
-  position: absolute;
+.tune.editorModule .deckGeneratorSuit {
+  display: flex;
+  align-items: center;
+  gap: 6px;
+  margin-bottom: 4px;
 }
 
-.tune.editorModule .suitProperties > div {
-  margin-left: 170px;
+.tune.editorModule .deckGeneratorSuit > .propertyInput {
+  flex: 0 0 auto;
+  margin: 0;
 }
 
-.tune.editorModule .suitProperties label {
-  display: inline-block;
-  width: 60px;
+/* the row's icon in the suit's color (see showIconInSuitColor): the icon masks the color */
+.tune.editorModule .deckGeneratorSuit .deckGeneratorSuitTint {
+  width: 100%;
+  height: 100%;
+  background: currentColor;
+  -webkit-mask: var(--mask) center / contain no-repeat;
+  mask: var(--mask) center / contain no-repeat;
+}
+
+.tune.editorModule .deckGeneratorSuitName {
+  flex: 1 1 80px;
+  min-width: 60px;
+}
+
+.tune.editorModule .deckGeneratorSuitRanks {
+  flex: 2 1 130px;
+  min-width: 80px;
+}
+
+.tune.editorModule .deckGeneratorSuitAdd {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+  gap: 4px;
+  margin: 8px 0;
+}
+
+.tune.editorModule .deckGeneratorSuitAddLabel {
+  font-size: 12px;
+  opacity: 0.7;
+}
+
+.tune.editorModule .deckGeneratorSuitAdd .propertyValueChip {
+  cursor: pointer;
+}
+
+.tune.editorModule .deckGeneratorSuitAdd .propertyValueChip:hover {
+  outline: 2px solid var(--VTTblue);
+}
+
+.tune.editorModule .deckGeneratorLinkedRanks {
+  display: flex;
+  align-items: center;
+  gap: 6px;
+  font-size: 12px;
 }
 
 
@@ -182,7 +232,9 @@
   min-width: 0;
 }
 
-.tune.editorModule .ttsDeckButton .deckCardCount {
+/* caption bar of a preview tile: the card count of a TTS deck, the name of a card design */
+.tune.editorModule .ttsDeckButton .deckCardCount,
+.tune.editorModule .deckDesignButton .deckDesignLabel {
   position: absolute;
   bottom: 0;
   left: 0;
--- a/client/editor.html
+++ b/client/editor.html
@@ -239,6 +239,11 @@
         <label><input type="radio" name="deckEditorNewDeckMode" value="images">Upload one image per card or tiled images</label>
         <label><input type="radio" name="deckEditorNewDeckMode" value="tts">Import from Tabletop Simulator</label>
       </div>
+      <div class="deckEditorNewDeckOptions">
+        <span class="deckEditorNewDeckOptionsLabel">Add along with the deck</span>
+        <label title="A holder in the middle of the table that the deck and all its cards are put into."><input type="checkbox" id="deckEditorNewDeckHolder" checked>Holder for the deck, with the cards in it</label>
+        <label title="A button below the holder that recalls all cards into it, flips them to their back and shuffles them."><input type="checkbox" id="deckEditorNewDeckResetButton" checked>Reset button that recalls, flips to back and shuffles</label>
+      </div>
       <div id="deckEditorNewDeckPanel"></div>
     </div>
   </div>
--- a/client/editor.html
+++ b/client/editor.html
@@ -240,9 +240,10 @@
         <label><input type="radio" name="deckEditorNewDeckMode" value="tts">Import from Tabletop Simulator</label>
       </div>
       <div class="deckEditorNewDeckOptions">
-        <span class="deckEditorNewDeckOptionsLabel">Add along with the deck</span>
-        <label title="A holder in the middle of the table that the deck and all its cards are put into."><input type="checkbox" id="deckEditorNewDeckHolder" checked>Holder for the deck, with the cards in it</label>
-        <label title="A button below the holder that recalls all cards into it, flips them to their back and shuffles them."><input type="checkbox" id="deckEditorNewDeckResetButton" checked>Reset button that recalls, flips to back and shuffles</label>
+        <span class="deckEditorNewDeckOptionsLabel">Added to the table with every new deck</span>
+        <label title="A holder in the middle of the table that the deck and all its cards are put into. Without it, the cards go into a pile in the middle of the table instead."><input type="checkbox" id="deckEditorNewDeckHolder" checked>Card holder on the table, holding all the cards</label>
+        <label title="A button below the holder that gathers all cards back into it, turns them face-down and shuffles them."><input type="checkbox" id="deckEditorNewDeckResetButton" checked>Reset button that gathers all cards back, turns them face-down and shuffles</label>
+        <span class="deckEditorNewDeckOptionsHint">Without a holder, the cards go into a pile in the middle of the table.</span>
       </div>
       <div id="deckEditorNewDeckPanel"></div>
     </div>
--- a/client/js/editmode.js
+++ b/client/js/editmode.js
@@ -2,23 +2,7 @@
 function generateCardDeckWidgets(id, x, y, addCards) {
   const widgets = [
     { type:'holder', id, x, y, dropTarget: { type: 'card' } },
-    {
-      id: id+'B',
-      parent: id,
-      fixedParent: true,
-      y: 171.36,
-      width: 111,
-      height: 40,
-      type: 'button',
-      text: 'Recall & Shuffle',
-      movableInEdit: false,
-
-      clickRoutine: [
-        { func: 'RECALL',  holder: '${PROPERTY parent}' },
-        { func: 'FLIP',    holder: '${PROPERTY parent}', face: 0 },
-        { func: 'SHUFFLE', holder: '${PROPERTY parent}' }
-      ]
-    }
+    deckResetButton(id, 111, 171.36)
   ];
 
   const types = {};
@@ -1299,6 +1283,7 @@ function populateAddWidgetOverlay() {
 let libraryDecksIndex = null;
 let libraryDecksObserver = null;
 let libraryDeckPreviewCounter = 0;
+let libraryDecksPlacement = null; // set by openLibraryDecksOverlay for the deck a click adds
 const libraryDeckDetailsCache = {};
 
 function getLibraryDeckDetails(entry) {
@@ -1317,7 +1302,10 @@ function getLibraryDeckDetails(entry) {
   return libraryDeckDetailsCache[key];
 }
 
-async function openLibraryDecksOverlay() {
+// placement: what to add around a picked deck, when the browser was opened from the "Add New Deck" dialog (which
+// is hidden while browsing). Opened from anywhere else, a picked deck gets the holder and button it always got.
+async function openLibraryDecksOverlay(placement) {
+  libraryDecksPlacement = placement || deckPlacementDefault;
   showOverlay('libraryDecksOverlay');
   if(libraryDecksIndex == 'loading')
     return;
@@ -1461,6 +1449,7 @@ async function renderLibraryDeckPreview(entry, container) {
 
 // adds the deck like the "add deck" entry of the add widget overlay does:
 // a holder containing the deck, a pile with all its cards and a shuffle button
+// (holder and button are optional when the browser was opened from the "Add New Deck" dialog)
 async function addLibraryDeckToGame(entry) {
   let details;
   try {
@@ -1470,6 +1459,7 @@ async function addLibraryDeckToGame(entry) {
     return;
   }
 
+  const placement = libraryDecksPlacement || deckPlacementDefault;
   batchStart();
   setDeltaCause(`${getPlayerDetails().playerName} added deck ${entry.deck} from public library game ${entry.gameName} in editor`);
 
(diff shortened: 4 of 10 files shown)


VTT-AI-Agent (commented):
## TL;DR

Careful, well-commented refactor: the four creation flows now share `addDeckWithCards`/`deckResetButton`, the default wizard output stays byte-identical (I re-derived the court-card letters — `'DHCS'[suitIndex % 4]` matches the old `defaultRanks[…].substr(6,1)`), every design uses `type: icon` for the suit so arbitrary icons really do work, and all CI checks are green. No engine files are touched at all. Nothing blocking; the most worthwhile fixes are around rank parsing — a trailing comma or a repeated rank still silently produces a junk or a missing card type, and `'JQK'.includes(rank)` matches multi-letter ranks like `JQ`.

## Findings

**1. `client/js/editor/sidebar/properties.js:65` — empty rank entries survive the new trim.**
`'2-10,J,Q,K,A,'` (trailing comma, very easy to type in the new per-suit rank field) yields `['2',…,'A','']`, so `getCardTypes` produces a card type literally named `" of hearts"` with `rank: ''`. On *Playing cards* it even gets a `rankIcon` (because `''.match(/^[0-9]+$/)` is null → `!hasPipLayout`), i.e. a silent extra card showing only the suit icon. Before this PR the same input produced a 404 `rankImage`, so it isn't a regression — but the trim is the right place to fix it, and `parseRankRange` now has exactly one caller, so `.filter(rank => rank !== '')` is safe.

**2. `properties.js:1263` — `'JQK'.includes(rank)` matches multi-character ranks.**
Word/multi-letter ranks are now an advertised use case ("Corner index … works for any rank set"), and a rank of `JQ`, `QK` or `JQK` on the *Playing cards* design produces `/i/cards-default/JQD-face.svg` → a broken image in the middle of the card. `String(rank).length == 1 && 'JQK'.includes(rank)` fixes it (the new `rank &&` guard already covers the empty case).

**3. `properties.js:1236` — duplicate ranks within one suit silently collapse.**
`cardTypes[cT]` is keyed by `"<rank> of <suit>"`, so `2-10,10` gives 9 cards instead of 10 and `A,A` gives one card, with no feedback. Suit *names* get the `used`/numbering treatment in `uniqueSuitNames`, but ranks inside a suit don't — worth either the same numbering or a hint next to the field, since hand-typed rank lists are now the primary input.

**4. `properties.js:1264` — unguarded lookup on an object literal.**
`DECK_GENERATOR_COURT_SUITS[suit.icon]` walks the prototype chain, so an icon value of `constructor`, `valueOf` or `toString` returns an inherited value and lands stringified inside the image URL. Only reachable if such a value ever gets into an icon property (the picker won't produce one today), but `Object.hasOwn(DECK_GENERATOR_COURT_SUITS, suit.icon) ? … : 'DHCS'[…]` — or a `Map` — removes the class of problem for free.

**5. `properties.js:1086` — `showIconInSuitColor` finds the icon chip positionally.**
`$('.propertyPreviewButton .propertyValueChip', row)` resolves to the icon input's chip only because `iconInput.render(row)` happens before `colorInput.render(row)`. If the row ever gains another picker or is reordered, this silently tints the *color* chip instead (and `$('img', chip)` returning null means the failure is invisible). `$('.propertyValueChip', iconInput.dom)` expresses the intent and can't drift.

**6. `client/js/editor/propertyInputs.js:154` — `pickSymbolKeepingOverlay` leaves the DOM moved if `pickSymbol` rejects.**
`pickSymbol` → `loadSymbolPicker` does a `fetch`; if that throws, the `#symbolPickerOverlay` stays reparented next to the host overlay and the host overlay is never re-shown, so the user is left with a blank editor and has to reopen it. A `try { … } finally { pickerParent.appendChild(picker); showOverlay(hostOverlay.id); }` around the await covers it.

**7. Test coverage — the cheap wins are already wired up.**
`tests/client/property-inputs.test.js` evaluates `properties.js` and pulls pure helpers out of its scope. `parseRankRange` (the trimming fix, which is behavior users will notice) and `defaultSuitName` (icon → suit name, including the `/assets/`+URL early return and the `[…]`/`(…)` stripping) are exactly that shape and currently have no test at all. Two extra entries in the returned object plus a handful of `expect`s would lock in items 1–2 as well.

**Note (not a defect):** because designs now declare the card-type properties they read, a deck generated with anything other than *Playing cards* no longer carries `suit-P…`/`suit-S…`, so a user who later switches that deck's faces to the pip layout in the deck editor has to regenerate the deck. That's the right trade (the old output was noise), but it's the one semantic change to the generated JSON and worth a line in the wiki's deck-wizard description.

I could not run jest/TestCafe in this sandbox (no `node_modules` in the review clone), so the above is a static review; CI on this PR is fully green, including the `editor.js` deck-editor state-hash tests.

I added the `beta` label: the diff touches only edit-mode UI (CSS, `editor.html`, `editmode.js`, `client/js/editor/*`) with no engine file (`client/js/widgets/`, `compute.js`, `serverstate.js`) and no server code, so it is not beta-excluded, and the remaining findings are cosmetic edge cases in a wizard that cannot corrupt existing games or crash the server.

VTT-AI-Agent:
Thanks — all seven points addressed in 01f83c21. Findings 1–4 are all reachable through the new per-suit rank field, so they were worth fixing; I verified each one in a real browser by typing `2-4,J,J,JQ,-2-0,` into a diamonds suit and generating a *Playing cards* deck.

**1. Empty rank entries** — `parseRankRange` now `.filter(rankElement => rankElement !== '')`, so a trailing comma just does nothing instead of producing a `" of hearts"` card type.

**2. `'JQK'.includes(rank)`** — now `String(rank).length == 1 && 'JQK'.includes(rank)`. `JQ` falls through to the `rankIcon` branch (suit icon in the middle) instead of requesting `/i/cards-default/JQD-face.svg`.

**3. Duplicate ranks** — numbered the same way suits left at the same name are, by probing `cardTypes` for a free key:

```js
let cT = `${rank} of ${suitNames[suitIndex]}`;
for(let i = 2; cardTypes[cT]; ++i)
  cT = `${rank} of ${suitNames[suitIndex]} ${i}`;
```

`2-4,J,J` now yields `J of diamonds` *and* `J of diamonds 2`, both with `rank: "J"`, so the card count matches what was typed. Probing the target object rather than a per-suit set also keeps it correct in the one cross-suit collision that was possible (a suit named `hearts 2` next to a `hearts` suit with a repeated rank).

**4. Prototype lookup** — extracted into `courtSuitLetter(icon, suitIndex)` guarded with `Object.prototype.hasOwnProperty.call(...)`, matching how the rest of the repo does this check (`widget.js`, `server.mjs`). The duplicate-rank loop above can't hit the same class of problem, since a card type key always contains `" of "`.

**5. Positional chip lookup** — now `$('.propertyValueChip', iconInput.dom)`. Confirmed in the browser that each suit row's tinted chip is the icon input's: the icon chip gets `style.color` + a `.deckGeneratorSuitTint`, the color input's chip stays untouched.

**6. `pickSymbolKeepingOverlay`** — wrapped in `try { return await pickSymbol(...) } finally { … }`, so the picker gets reparented and the host overlay re-shown even if `loadSymbolPicker`'s fetch throws.

**7. Tests** — added a `deck generator helpers` describe block covering `parseRankRange` (ranges, list whitespace, the empty-entry fix), `defaultSuitName` (game-icons/material/symbols/emoji, the `/assets/` + URL early return, `null`) and `courtSuitLetter` (mapped suit, positional fallback, `constructor`). 90 jest tests pass and `tests/testcafe/editor.js` is green locally (15/15), so the default wizard output is still byte-identical.

**One extra, found while writing the `parseRankRange` tests:** the range regex has always advertised negative bounds (`/-?[0-9]+--?[0-9]+/`) but `rankElement.split('-')` cut them in half, so `-3--1` produced ranks `0,1,2,3` — junk, not the intended `-3,-2,-1`. Since I was already in that function and the fix is covered by the new test, I read both bounds from an anchored match instead:

```js
const range = rankElement.match(/^(-?[0-9]+)-(-?[0-9]+)$/);
```

Anchoring also means a word rank that merely *contains* a range-looking substring (`card2-4x`) stays one literal rank rather than being silently dropped, which fits multi-letter ranks now being an advertised use case. `2-10` and every other existing input parse identically.

**On your non-defect note** (a deck generated with a non-*Playing cards* design no longer carries `suit-P…`/`suit-S…`): agreed that's the right trade, and it does deserve a line in the wiki's deck-wizard description — the wiki lives outside this repo, so I can't include it in the diff; flagging it here for whoever updates it.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review) of the reworked **Add a new deck** wizard, driven live in a browser at 1600×950 and 820×640. The new suit rows and the "Add along with the deck" options are a clear improvement over the old icon-toggle grid — but two things block a first-time user: **"Show all" in a suit's icon picker renders the full symbol picker *behind* the deck editor, leaving an empty white screen**, and the nine card-design previews are clipped fans on a checkerboard that are hard to tell apart. The single most impactful fix is the symbol-picker stacking (one z-index), followed by making the design tiles show one legible card each.

Reviewed by running the branch on a local server and stepping through the wizard with Playwright.

| The new suit rows | The nine designs (2× zoom) |
| --- | --- |
| suits | designs |

---

## Big layout improvements

### 1. "Show all" from a suit's icon picker shows a blank screen (symbol picker is behind the deck editor)

`pickSymbolKeepingOverlay()` (client/js/editor/propertyInputs.js:163) moves `#symbolPickerOverlay` from `#deckEditorMainCol` into `#editor`. That reparenting drops the only rule that gives it a stacking context — `#deckEditorMainCol #symbolPickerOverlay { … z-index: 20 }` (client/css/editor/deckeditor.css:459) — so the picker computes `z-index: auto` and lands **below** `#deckEditor` (`z-index: 1`). Measured in the browser after clicking "Show all" in a suit row: the overlay *is* `display:flex`, 1600×950, fully populated (3.6 MB of markup), yet `document.elementFromPoint(800, 500)` returns `#deckEditorMain`, i.e. the picker is completely occluded. What the user sees is the dialog gone and an empty editor with no visible way back:

blank

Suggested fix: give the reparented state its own rule, e.g. `#editor > #symbolPickerOverlay { position: fixed; inset: 0; z-index: 30; }` (above `#deckEditor`/sidebar/toolbar), so it really does appear "full size above the deck editor" as the PR intends.

### 2. Make the card-design tiles readable — one card per tile, not a clipped fan

The nine previews are the strongest new feature of the flow and currently the hardest part to use. Each 150×150 tile (`.editorModule .widgetSelectionButton`, client/css/editor/sidebarModules.css:47) renders a fan of 103×160 cards, so the fan overflows and gets cut on every side: "Bold color" shows `1 10`, "Skinny" shows `J1 Q`, "Round token" and "Square tile" are sliced at the tile edge, and "Minimal" has a half-cut `Joke` hanging out of the frame (see the 2× screenshot above). A first-time user cannot tell "Minimal" from "Corner index" from these.

Concretely:
- render **one** representative card (or 2–3 clearly offset), scaled to fit the tile with ~8px padding, instead of the full fan — the whole point of the tile is "what does a card look like";
- move the design name **below** the tile (or add matching bottom padding) — right now the `#fffb` caption bar (client/css/editor/sidebarModules.css:235-246) sits on top of the artwork;
- lay the gallery out as a grid (`display: grid; grid-template-columns: repeat(auto-fill, minmax(150px, 1fr))`) so 9 tiles don't leave one orphan on its own row, and so they never overflow — at an 820px-wide window the 4th column is clipped by the dialog:
  narrow
- the checkerboard "transparency" background reads as *missing image* rather than *card design*; a flat neutral (or the table background) would let the designs speak for themselves. Card backs are the one design element that legitimately wants the checkerboard.

### 3. The suit table needs column headers, and the icon picker needs to say which suit it belongs to

Four unlabeled controls per row (client/js/editor/sidebar/properties.js:1091-1175). While the defaults are filled in, both placeholders are hidden, so a newcomer sees `[icon] [diamonds] [red] [2-10,J,Q,K,A] [🗑]` with nothing telling them the third box is a colour or the fourth is ranks. A single tiny header row above the list — reusing the existing `deckEditorNewDeckOptionsLabel`-style 11px uppercase caption (Icon · Name · Colour · Ranks) — would fix that for free.

Related: the icon/colour pickers expand into one shared container *below the whole list* (`.deckGeneratorSuitPickers`). With seven suits, editing suit #1's icon opens a panel seven rows away, and the panel only shows the raw value (`skoll/diamonds`), not which suit it belongs to:

picker

Either put the picker directly under its own row, or title it "Icon for *diamonds*" and outline the row it is editing while it is open.

### 4. Scope of "Add along with the deck" is ambiguous

The new block (client/editor.html:242-247) sits between the mode radios and the mode panel, styled with the same 11px uppercase caption as the panel's own `CARD SIZE` section — so it reads like part of whichever mode is selected rather than a setting for all six. Either move it *below* the mode panel (last step: "what gets added around it"), or set it apart visually — e.g. an indented/boxed group, or a caption that names the scope: "Added with every deck".

- **Disabled "Add to game" looks enabled.** `createButton.className = 'green'` + `disabled` (properties.js:1289-1292): `button[icon].green` (specificity 0,2,1) beats `button:disabled` (0,1,1) in client/css/layout.css:1189, so the greyed-out state never applies. With no design preselected, the first thing a newcomer sees is a healthy dark-green **ADD TO GAME** that does nothing. Add `button[icon].green:disabled { background-color: #808080; color: #d3d3d3; }` (or preselect *Playing cards*).
- **Disabled reset-button checkbox still reads as "on".** Unchecking *Holder* disables the reset checkbox (deckeditor.js:270) but leaves it ticked and its label at full opacity, so it looks like you still get a reset button: holder off — dim the whole label (`.deckEditorNewDeckOptions label:has(input:disabled) { opacity: 0.5 }`) and/or add `title="A reset button recalls the cards into a holder"`.
- **Empty "Card design" section.** Delete every suit and you get a "Card design" header with nothing underneath and a dead Add button, no explanation. One line in the empty branch ("Add at least one suit to see the card designs") covers it.
- **Destructive action in the accent colour.** The per-suit remove button (properties.js:1167-1169) is a filled `--VTTblue` trash icon; elsewhere in this same screen destructive buttons are red (`#deckEditorStripDelete`). With 7 suits it's a column of loud blue. Red — or an unfilled/ghost icon button — matches the app and de-emphasises the row's least-used control.
- **"Same ranks for each suit" is below the "Add a suit:" chips**, i.e. two elements away from the Ranks column it governs (properties.js:1038-1056). Placing it just under the suit list (or right-aligned above the Ranks column) links cause and effect.
- **"Add a suit:" chips are all black** even though clicking hearts/diamonds creates a *red* suit; tinting each chip with the colour it will produce makes the shortcut honest. Their checkerboard chip background also makes the row read as a strip of transparency swatches.
- **Duplicate suit names are silently disambiguated.** Adding hearts twice leaves two rows both literally named `hearts`; card types become `7 of hearts` and `7 of hearts 2` with nothing in the UI saying so ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1532206823472496770/shot-07-many-suits.png)). Showing the resolved name (placeholder, or auto-filling `hearts 2` into the second row) avoids a surprise in the card-type list.
- **Design selection is a toggle, not a radio.** Clicking the selected design deselects it and re-disables Add with no feedback; the section is single-choice, so it should behave like the mode radios above it.
- **Hint text contrast.** `.deckGeneratorHint` is 12px at `opacity: 0.7` over the light dialog (sidebarModules.css:124-128); combined with the dialog's blue text it sits near the WCAG AA floor. 0.85 would keep the "secondary" feel while staying comfortably legible.

- **"Reset button that recalls, flips to back and shuffles"** (editor.html:245) is three pieces of VTT jargon in one line — *recall*, *flips to back*, and the implied "into the holder". The tooltip explains it, but tooltips are invisible on touch and to anyone not hovering. Suggested wording: **"Reset button that gathers all cards back, turns them face-down and shuffles"**. Likewise **"Holder for the deck, with the cards in it"** → **"Card holder on the table, holding all the cards"**.
- **Rank syntax is undiscoverable once the field is filled.** `2-10,J,Q,K,A` is prefilled and the `title` is the only place the range syntax is documented. Add it to the section hint: *Ranks are comma-separated; `2-10` expands to a range.* (The PR's own "spaces after commas now work" fix is exactly the kind of thing users only discover by getting it wrong.)
- **What "suit" and "rank" mean here.** The hint explains the naming (`7 of hearts`) well — worth one more clause that the number of cards is *suits × ranks*, since the deck size is otherwise invisible until the cards land on the table (a small "36 cards" counter next to *Card design* would be even better and would make the previews' meaning obvious).
- **Design names are good, descriptions are missing.** "Icon shape", "Corner index" and "Round token" are not self-explanatory; the tiles already carry `title=design.label`, which just repeats the visible name. A one-line `title` per design ("Corner index — rank and suit in two corners, works with any rank names") would help pick without trial and error.
- **Holder-off result is unlabeled.** Without a holder you get a pile plus a bare circular deck badge 10px to its left in edit mode, with no hint that the circle is the (play-invisible) deck widget: table. A short note under the checkbox ("Without a holder, the cards go into a pile in the middle of the table") sets the expectation before the click.
- **No progress feedback while cards are added.** For a 60-card deck the wizard stays open and unchanged for a moment before it closes; the button already re-disables itself, so showing "Adding…" on it (like `progressButton` does for the PDF export) would tell the user it worked.

Everything above is UI/UX only — layout, wording and states. Correctness/behaviour is covered by the other reviewers.

VTT-AI-Agent:
Thanks — the blocker and almost everything else is fixed in 1742b874, verified in a real browser at 1600×950 and at 820×640. `npm test` (90) and `tests/testcafe/editor.js` (15/15) are green.

| Suit rows | Card designs |
| --- | --- |
| suits | designs |

## Big layout improvements

**1. "Show all" showed a blank screen — fixed.** Exactly as diagnosed. `pickSymbolKeepingOverlay` now adds a `symbolPickerAboveEditor` class for as long as the picker is parked next to the dialog, and `deckeditor.css` floats that state over the whole editor (`position: fixed; inset: 0; z-index: 30`, above the deck editor 1, sidebar 2, toolbar 3 and the dialogs 20). Measured after clicking "Show all": `elementFromPoint(800, 475)` is now `#symbolList` and the computed `z-index` is `30` (was `auto`). The class is removed in the same `finally` that reparents the picker, so nothing leaks into the normal `#deckEditorMainCol` state.
show all

**2. Design tiles — one legible card per tile.** `renderWidgetButton` grew an optional `sampleCardTypes` argument (default is still the random five), and the gallery passes a single card type — *the same one in every design*, picked deterministically (a numeric rank 5–10 where there is one, so the playing-card design shows its pip layout) so the nine designs can be compared side by side. Also:
- the design name is a sibling **below** the tile now, not a `#fffb` bar on the artwork;
- the gallery is `display: grid; grid-template-columns: repeat(auto-fill, minmax(156px, 1fr))` — 3×3 at both 1600px and 820px, nothing clipped, no orphan row ([820px](https://agent.virtualtabletop.io/reports/pr/1532171279962996877/designs-narrow-820.png));
- the checkerboard is gone, the tiles are a flat `#ededed`.

One correction to the finding: the fans weren't actually escaping the tiles (measured child rects stayed inside the 150×150 button; the "cut" edges were cards overlapping each other). The result is the same though — five cards at ~0.5 scale were unreadable and "Minimal" vs "Corner index" was guesswork. At one card the scale is ~0.87 and each design is obvious.

While there I fixed a latent divide-by-zero in `positionElementsInArc`: with a single element `arcAngle / (n - 1)` was `Infinity * 0` = `NaN`, so `left`/`top` were dropped.

**3. Suit table headers + which suit the picker edits.** The suit list is now one CSS grid (`.deckGeneratorSuit` / the header row are `display: contents`), so an **ICON · NAME · COLOR · RANKS** header row lines up exactly with the inputs. For the picker I took your first option rather than the caption: each row has its own `pickerTarget`, so the icon/color picker expands *directly under the row it belongs to* — no `pickerGroup` change was needed, `PickerInput` already falls back to `options.pickerTarget`. Only one picker is open at a time as before.
picker

**4. Scope of the options block.** I took the "set it apart visually" branch rather than moving it below the panel — after the custom panel it would be ~1000px below the mode radios and easy to miss. It is now an indented boxed group with the caption **"Added to the table with every new deck"**, which names the scope.

- **Disabled "Add to game"** — fixed, scoped to these flows: `.tune.editorModule button[icon].green:disabled` (and `:disabled:hover`, which had an explicit `background-color: darkgreen`) now grey out. I left the global `button[icon].green` rules alone since they are used elsewhere.
- **Disabled reset checkbox** — `.deckEditorNewDeckOptions label:has(input:disabled) { opacity: 0.5 }`, and the tooltip is now "…gathers all cards back into it, turns them face-down and shuffles them".
- **Empty "Card design" state** — "Add at least one suit above to see the card designs." in the same hint line that otherwise shows the card count.
- **Blue trash** — the per-suit remove button is red now (`var(--VTTred)`, same treatment as `#deckEditorStripDelete`).
- **"Same ranks for each suit"** — moved up, directly under the suit list and above the "Add a suit:" chips.
- **"Add a suit:" chips** — tinted with the color that clicking them actually produces, refreshed on every render (so with the four default suits the diamonds chip is black, and it turns blue once a black diamonds suit exists), tooltip "Add a diamonds suit in this color". Their checkerboard background is gone.
- **Duplicate suit names** — the affected row's name input gets a red border and a tooltip: *Another suit is already called "diamonds", so this one's cards are named "7 of diamonds 2".* It updates while you type, and an empty name shows the resolved one as placeholder.
- **Toggle vs radio** — clicking the selected design keeps it selected.
- **Hint contrast** — `.deckGeneratorHint` is `opacity: 0.85`.

## Readability

- "recalls, flips to back" → **"Reset button that gathers all cards back, turns them face-down and shuffles"**, and "Holder for the deck, with the cards in it" → **"Card holder on the table, holding all the cards"**.
- **Rank syntax** is in the section hint now: *Ranks are separated by commas, and "2-10" is short for the whole range.*
- **Deck size counter**: the line above the gallery reads *"52 cards from 4 suits. Pick how they look:"* and updates live.
- **Per-design tooltips**: each tile's `title` is `"Corner index - Rank and suit in two corners, works with any rank names."` etc., one line per design.
- **Holder-off result**: a hint that appears under the checkboxes as soon as the holder is unticked — *"Without a holder, the cards go into a pile in the middle of the table."*
- **Progress feedback**: the button says "Adding…" while it works. Worth knowing what you were seeing though: `addDeckWithCards` only does local `addWidgetLocal` calls inside a batch and resolves in well under one frame even for 52 cards — the pause before the cards appear is the delta round-tripping through the server, after the dialog is already done. So the label is honest about the dialog's own work, but on a fast connection you will barely see it; the real fix for the perceived wait would be somewhere in the delta path, which is out of scope here.

Not changed: nothing else from the review is outstanding.

VTT-AI-Agent (commented):
## TL;DR

Two edge cases remain: an empty custom rank set throws while rebuilding the design gallery, and a symbol-picker load failure still hides the dialog it came from. Neither affects existing games or server safety, but both should be fixed and covered before merge.

## Findings

### 1. Empty rank lists crash the custom-deck preview

`client/js/editor/sidebar/properties.js:1247`

Clearing the ranks field for every suit is a valid intermediate state, and `parseRankRange('')` now deliberately returns an empty array. That makes `getCardTypes()` return `{}`, so `previewCardType(cardTypes)` returns `undefined` and this line passes `[undefined]` to `renderWidgetButton`. The latter tries to render a `Card` with `cardType: undefined`; `Card.applyInitialDelta` throws because every card requires a card type. Since the refresh runs from the 250 ms input timer without a catch, the gallery rebuild aborts and the Add button can retain its previous enabled state.

Please handle a zero-card result before rendering design cards—for example, show an “add at least one rank” hint, disable Add, and return. A UI-level regression test that clears the shared ranks field would cover both the exception and button state.

### 2. The failure cleanup can toggle an already-visible host dialog off

`client/js/editor/propertyInputs.js:173`

If `loadSymbolPicker()` rejects, `pickSymbol()` throws before it calls `showOverlay('symbolPickerOverlay')`, so `hostOverlay` is still visible when the `finally` block runs. `showOverlay(hostOverlay.id)` is a toggle when that same overlay is visible; it therefore sets the host dialog to `display: none`, recreating the blank-editor failure this cleanup is intended to prevent.

Only reopen the host when it was actually hidden (or restore its prior visibility explicitly). A rejected-load test should assert that the originating overlay remains visible and that the picker is returned to its original parent.

The `beta` label was already present; I left it in place because the diff contains no engine or server changes, and these edit-mode UI failures cannot corrupt existing games or crash the server.

VTT-AI-Agent:
Thanks — both findings are real and both are fixed in d7015568. `npm test` (93) and `tests/testcafe/editor.js` (16/16, including a new test for finding 1) are green locally.

**1. Empty rank lists crashed the preview** — confirmed exactly as described: `getCardTypes()` returns `{}`, `previewCardType` returns `undefined`, and the `Card` render throws out of the 250 ms timer, so the Add button kept whatever state it had (with a design selected it stayed *enabled*, and clicking it added a deck with zero cards).

The three hint branches are now one pure helper next to `parseRankRange`, and the zero case returns before anything is rendered:

```js
const cardCount = suits.length ? Object.keys(cardTypesFor(designs[0])).length : 0;
designHint.textContent = deckGeneratorDesignHint(suits.length, cardCount);
if(!cardCount) {
  createButton.disabled = true;
  return;
}
```

`deckGeneratorDesignHint(4, 0)` is *"Add at least one rank above to see the card designs."*, and the card-type cache moved into a `cardTypesFor(design)` closure so the count can be taken once before the loop. Covered by a jest test for the helper and by the UI-level regression test you asked for — `Deck editor: the custom deck wizard survives an empty rank list` clears the shared ranks field and asserts the hint, an empty gallery and a disabled Add button (and, since TestCafe fails on uncaught client errors, the exception itself). With the guard removed it fails on both counts.

**2. The failure cleanup toggled the still-visible dialog off** — also confirmed, and worse than the finding says. I reproduced it in a real browser by aborting `i/fonts/symbols.json` and clicking "Show all" in a suit row: `showOverlay(hostOverlay.id)` hid the dialog, *and* the unhandled rejection then reached the global error handler, which put up `#clientErrorOverlay` — so the user lost the dialog either way. Three changes:

- reopen only when the picker actually took the dialog's place (`if(hostOverlay.style.display == 'none')`);
- keep the failure inside `pickSymbolKeepingOverlay`: it returns `null` (nothing picked) and says `The symbol picker could not be loaded. Please try again.` instead of letting the client error overlay replace the dialog;
- `loadSymbolPicker` resets `symbolData` from `'loading'` back to `null` when the fetch throws — otherwise the first failure broke the picker for the rest of the session and "try again" would have been a lie.

Measured with the same aborted fetch after the fix: dialog `display: flex`, picker back under `#deckEditorMainCol`, no page error; unblocking the fetch and clicking "Show all" again opens the picker with all 13288 symbols. The normal path is unchanged (dialog hidden while picking, back afterwards, suit icon set). The jest test asserts both outcomes — picked symbol and failed load — including that the originating overlay stays visible and the picker is returned to its original parent.

No engine or server file is touched, so the `beta` label stays as you left it.

VTT-AI-Agent (commented):
## TL;DR

Fourth pass over the reworked deck wizard. Everything raised in reviews 1–3 is genuinely fixed in code (anchored `parseRankRange` with empty-entry filtering + range bounds read from the match, `courtSuitLetter` with an own-property guard, `String(rank).length == 1 && 'JQK'.includes(rank)`, per-rank duplicate numbering, the chip looked up via `iconInput.dom`, the `symbolPickerAboveEditor` stacking rule, the `hostOverlay.style.display == 'none'` guard against `showOverlay`'s toggle, `loadSymbolPicker` no longer stuck in `'loading'`, and the empty-gallery path with jest + TestCafe coverage for both). All checks are green and no engine file is touched. What's left is one narrow way to get a `TypeError` out of the custom flow: **"Add to game" is re-enabled unconditionally in its `finally`, so if the design gallery emptied while the cards were being added, the next click dereferences a null selected tile.**

## Findings

### 1. `client/js/editor/sidebar/properties.js:1375` — "Add to game" can be re-enabled with nothing selected

`addDeckWithCards` awaits once per card, so for a 52-card deck the dialog stays interactive for a noticeable moment. Clearing the shared ranks field during that window fires `updateDesignPreviewSoon` → `updateDesignPreview`, which does `designSelectionDiv.innerHTML = ''` and `createButton.disabled = true` (properties.js:1235,1252). The `finally` then unconditionally undoes that:

```js
} finally {
  createButton.innerText = 'Add to game';
  createButton.disabled = false;   // <- even though the gallery is now empty
}
```

The user is left with an enabled green button above "Add at least one rank above to see the card designs."; clicking it hits properties.js:1366

```js
const design = designs[+$('.selected.deckDesignButton', designSelectionDiv).dataset.index];
```

with a `null` selector result → `TypeError: Cannot read properties of null (reading 'dataset')` → the global handler swaps the dialog for the client error overlay. Mirroring line 1276 fixes it:

```js
createButton.disabled = !$('.selected.deckDesignButton', designSelectionDiv);
```

### 2. `properties.js:1236` — the picked design is silently forgotten when the gallery empties

The selection is remembered only by reading it back off the DOM (`oldSelectedButton.dataset.index`), but the zero-card early return at line 1252 has already wiped `designSelectionDiv`. So: pick *Playing cards* → clear the ranks → type them again ⇒ the gallery comes back with nothing selected and Add still disabled, with no indication why. The new TestCafe test walks exactly this path (`tests/testcafe/editor.js:788-793`) but only asserts the hint and the tile count, so it wouldn't catch it.

Keeping the index in a closure variable (set in `designButton.onclick`, cleared only when the design list itself changes) survives the empty state and is also simpler than the read-back. Worth one extra `.expect(addToGame.hasAttribute('disabled')).notOk()` after the second `setSharedRanks('A')` once it does.

### 3. `properties.js:1640-1643` — the uniqueness loop covers `P` but not the `B` the reset button uses

```js
do {
  holderID = generateUniqueWidgetID();
} while(widgets.has(holderID+'P'));
```

but `deckResetButton(holderID, …)` (deckeditor.js:3592) creates `holderID+'B'`. `generateUniqueWidgetID` only guarantees `holderID` itself is free, so a room that already contains a widget literally named `<holderID>B` gets a colliding `addWidgetLocal`. Adding `'B'` is the whole fix, and it matches what `addLibraryDeckToGame` already does for all of its suffixes (`editmode.js:1466`):

```js
} while([ 'B', 'P' ].some(suffix=>widgets.has(holderID+suffix)));
```

### 4. `properties.js:61` — while you're hardening rank parsing, the range expansion is still unbounded

`parseRankRange('2-100000')` builds 100 000 ranks per suit, and `getCardTypes` then allocates a card-type object for each; with the four default suits that is ~400 k objects, built inside a 250 ms debounce on every keystroke. The tab freezes. This is pre-existing behaviour (the old unanchored regex expanded the same range), so it isn't a regression — but the ranks field is now the primary input of the flow and a fat-fingered `2-1000` is easy. A cap (e.g. stop expanding past a few hundred ranks and show the existing `deckGeneratorHint` line as a warning) would be cheap here, since `parseRankRange` now has exactly one caller and is unit-tested.

### Verified, not defects

- `pickSymbolKeepingOverlay` parks the picker in `#editorOverlays` (the `.overlay` parent, editor.html:1), **not** `#deckEditorMainCol`, so the two-ID rule `#deckEditorMainCol #symbolPickerOverlay` (deckeditor.css:459) does not out-specify the new `#symbolPickerOverlay.symbolPickerAboveEditor` — the z-index 30 really applies.
- `closeOverlay: false` is the right third argument: the picker stays up until the reparent is undone and `showOverlay(hostOverlay.id)` hides it as "every other overlay", so there is no flash of the bare editor.
- The unconditional `rankIcon` icon object in `deckTemplate_standard` is inert for pip-layout ranks — `card.js:152` guards on `if(object.value)`, so a card type without `rankIcon` renders nothing there.
- `listenTo: []` on the per-suit `IconInput`/`ColorInput` means `PropertyInput.render` registers no property listeners (propertyInputs.js:437-439), so re-rendering the suit list on every add/remove leaks no listeners onto the throwaway `suitInputWidget`.
- The `try { … } finally { widgets.delete(widget.id) }` around the preview cards in `renderWidgetButton` (properties.js:6809-6819) is a real improvement: before it, one throwing face template left a phantom deck in the widget map for the rest of the session.
- `deckPlacementDefault` / `deckResetButton` are `const`/function declarations in `client/js/editor/deckeditor.js`, which `server/minify.mjs` bundles into the same editor script *after* `sidebar/properties.js` and *before* `editmode.js` — all uses are inside functions called later, so there is no TDZ problem despite the reversed order.

I could not run jest or TestCafe here (this review clone has no `node_modules`), so the above is static; CI on the branch is fully green, including `editor.js` and CodeQL.

The `beta` label was already on this PR and I left it: the diff touches only edit-mode UI and its tests (`client/css/editor/*`, `client/editor.html`, `client/js/editmode.js`, `client/js/editor/*`, `client/js/symbols.js`) with no engine file — nothing under `client/js/widgets/`, no `compute.js`, no `serverstate.js` — and the four findings above are narrow wizard edge cases that cannot corrupt an existing game or crash the server.

VTT-AI-Agent:
Thanks — all four are real and all four are fixed in a80a0ad4. `npm test` (94) and `tests/testcafe/editor.js` (16/16) are green locally.

**1 + 2 — the Add button and the picked design.** Fixed together, because they are the same root cause: the selection only existed as a class on a tile the gallery keeps wiping. The picked design is now a closure variable and the button state is one helper that both the refresh and the `finally` go through:

```js
let selectedDesign = -1;
const updateCreateButton = _=>createButton.disabled = selectedDesign < 0 || !designSelectionDiv.children.length;
```

`designButton.onclick` sets `selectedDesign`, the tiles get `selected` back from it on every rebuild, and `createButton.onclick` reads `designs[selectedDesign]` instead of `+$('.selected…').dataset.index` — so the `dataset.index` read-back (and the attribute) is gone entirely. The `finally` calls `updateCreateButton()`, which leaves the button disabled when the gallery emptied while the cards were being added, exactly as you described. The zero-card early return goes through the same helper.

The TestCafe test now asserts your extra step: after clearing the ranks and typing them again, the previously picked tile is still `.selected` and Add is enabled. With the fix reverted it fails on precisely that line (`> 797 | .expect(designs.nth(0).hasClass('selected')).ok()`), so it does catch the regression.

**3 — the `B` suffix.** Fixed literally as suggested, matching `addLibraryDeckToGame`:

```js
} while([ 'B', 'P' ].some(suffix=>widgets.has(holderID+suffix)));
```

**4 — unbounded range expansion.** Capped. `parseRankRange` stops at `DECK_GENERATOR_MAX_RANKS = 200` (both inside a range and across the list, so `2-201,J` stops at 200 ranks rather than expanding and then trimming), and the hint above the gallery says so: *"800 cards from 4 suits. Only the first 200 ranks of a suit are used. Pick how they look:"*. Covered by a jest test for the cap and by the TestCafe test typing `2-100000` into the shared ranks field — which used to build 400 000 rank entries per keystroke and now renders the capped gallery normally.

No engine or server file is touched by this commit either (`client/js/editor/sidebar/properties.js` plus the two test files), so the `beta` label stays as you left it.

ArnoldSmith86 (approved):

PR #3075, opened 2026-07-30
Guard against widget graph cycles and unbounded routine recursion

### LINKED ISSUE #684: inheritFrom parent crashes client
Setting inheritFrom to the id of the parent, or setting the parent to the id the widget is inheriting from crashes the client.

### LINKED ISSUE #833: Reciprocal inheritFrom crashes room
There probably isn't a good fix for this, but we're told to report anything that crashes the room, so here goes.  I was trying to work on widgets mirror imaging each other like in the `inheritFrom` tutorial.  But I wanted either widget to be the "master," so I made widget 1 inherit the y coord from widget 2.  Worked fine.  Then I also made widget 2 inherit the y coord from widget 1.  Crashed the room.

### LINKED ISSUE #1405: `CALL` without "routine" crashes client

### LINKED ISSUE #1414: SET parent to self crashes the client

### LINKED ISSUE #1415: `loopRoutine` with array modifying itself crashe client
This routine crashes the client:

```
  {
      "func": "FOREACH",
      "in": [
        "seat1",
        "seat2",
        "seat3",
        "seat4"
      ],
      "loopRoutine": [
        "var score = ${PROPERTY score OF $value}",
        "var score = ${score} push 0"
      ]
    }
```
The error is "Converting circular structure to JSON --> starting at object with constructor `Array` index 3 closes the circle."

### LINKED ISSUE #1455: Routine clicking itself freezes client
Self-explanatory (and obvious).  But maybe there is a way to catch the error and break the loop.

I was trying to get a basic widget with faces to do the clickRoutine and advance the face.  If there is a clickRoutine on a basic widget with faces, only the clickRoutine works and not the faces.  But, this is a workaround:

```
  {
      "func": "CLICK",
      "collection": "thisButton",
      "mode": "ignoreClickRoutine"
    }
```

However, if you do not add the ignoreClickRoutine you get a button clicking itself and you get the maximum call stat size exceeded message.

VTT-AI-Agent:
Implements **phase 1** of the [client-crash issue report](https://agent.virtualtabletop.io/reports/crash-issues/) — groups A (structural cycles in the widget graph) and B (unbounded recursion / self-referential data).

The report's headline finding is that these crash reports are not independent bugs: the client has no structural invariant checks and no crash containment. Concretely, nothing stopped a widget from becoming its own ancestor, and `evaluateRoutine`'s `depth` argument was threaded through every recursive call site but never compared against a limit. This PR closes both gaps, so a bad input produces a readable problem entry instead of a dead client.

## Group A — structural cycles

* **`parent` (#1414, #684).** New `Widget.wouldCreateParentCycle(newParentID)` — the guard `MOVE` already had (`c == target` / `target.isDescendantOf(c)`), now enforced in `Widget.set()` itself, so every writer is safe by construction rather than each one having to remember the check. `set()` returns the reason when it refuses a write, and the paths a user can reach turn that into their own wording:
  * the `SET` and `RESET` operations, per widget with one problem entry each (as the report recommends, matching `MOVE`'s precedent), and
  * the JSON editor (single, multi-select and macro) and the properties sidebar, next to the existing "Parent X does not exist." check. Those only check an edit that actually changes where the widget hangs, so a widget that ended up inside a loop stays editable and can be repaired.

  `isDescendantOf` also became an iterative walk with a seen-set, so an already-broken state cannot overflow the stack while it is being inspected.
* **`inheritFrom` (#684, #833).** Rather than forbidding inheritance circles (widgets legitimately inherit *different* properties from each other), the two walks that could not survive one now stop when they come back to a widget they are already walking: `getDefaultValue()` falls back to the local default, and the inherited-delta propagation in `applyDeltaToDOM` does not hand the delta on again. An inherited `parent` that would make the target its own ancestor is dropped.

## Group B — unbounded recursion / self-referential data

* **Routine nesting (#1405, #1455).** `evaluateRoutine` now counts how many routines are running inside each other and aborts past 250, pushing a problem into the routine that caused it. Only routines reached by name count: an `IF` branch, a `FOREACH` body and the move list of a `Moves` operation are written inside the routine that runs them and cannot re-enter anything on their own, so they share its level. That keeps the number in the message the number a game actually gets - recursion is how the routine language writes a while loop, and its `CALL` almost always sits inside the `IF` that checks the base case, which would otherwise cost two levels per step. The message names what re-entered the routine, so it points at the operation to go and look at: `Not running clickRoutine of X more than 250 times. This is likely a recursive routine calling itself.` for `CALL`, `... This is likely a recursive click on itself.` for a click, and a generic sentence about something triggering the routine again for every other way in (a change routine, an `enterRoutine`). It is reported both in the routine that hit the limit and in the outermost one, which is the row the user can actually reach in the log. The counter sits in `evaluateRoutine` itself rather than on the `depth` argument, because the re-entry paths that actually cause these two issues (`CLICK` → `click()`, `changeRoutine`, `globalUpdateRoutine`) never passed `depth` on. This is the fix the maintainer prescribed on #1405.

  A routine waiting in `DELAY` or `INPUT` has nothing running inside it, so it gives its level back for as long as it waits — a game that keeps several timer-driven routines in flight is not refused the next click.
* **Self-referential values (#1415).** `StateManaged.set()` refuses a value that cannot be serialized instead of letting the `TypeError` tear down the client, the `SET` operation reports it as a problem, and every value the routine log formats goes through `stringifyForDisplay`, which shows `"<contains itself>"` where the loop closes instead of throwing while you are trying to debug it. As the report notes, the fix is "don't crash, explain" — the routine still does what the user wrote.

## What this does not cover yet

Every guard here runs on the client that *writes*. Nothing validates the invariant when a delta is *received*, so a parent cycle can still arrive over the network: if client A sets `w1.parent = 'w2'` while client B — not having seen A's delta yet — sets `w2.parent = 'w1'`, both pass their local check and every client then runs `newParent.appendChild(this.domElement)` in `applyDeltaToDOM` with `newParent` inside the element being appended. The limbo pre-pass in `receiveDelta` only detaches widgets named in the *incoming* delta, so it does not help when the counterpart write arrived earlier. Containing that `appendChild` alone would not be enough either: once the cyclic `parent` is in the state, the parent-chain walks behind `_absoluteRotation`, `_absoluteScale`, `_ancestor` and `inheritSeatVisibility` overflow the stack on the next read. Making the receive path safe needs either an apply-side rejection (which desyncs the client from the server) or bounding all of those walks — phase 2/3 of the report.

Loading a game file is not affected: `receiveStateFromServer` defers a widget whose parent does not exist yet and drops it if the parent never arrives, so a save containing a cycle never produced widgets in a cycle to begin with.

## Verification

* New TestCafe tests in `tests/testcafe/routines.js` and `tests/testcafe/editor.js` — each of them fails without the corresponding guard and passes here — plus jest cases for `stringifyForDisplay` in `tests/client/domhelpers.test.js`.
* Latest round: jest 1657/1657, TestCafe `routines.js` 25/25 (three new tests for the nesting limit), the Debug-module and JSON-editor-parent tests of `editor.js` and `compute-1.js` as a state-hash canary — all green.
* Full local suite green: jest, `routines.js`, `editor.js`, `compute-1..4.js`, `publiclibrary-1..4.js`. No state hashes changed. After merging `main` (which brought `SHIFT`, the expression module, the multi-widget JSON editor and the new Debug panel layout): jest 1657/1657, `routines.js` 21/21, `editor.js` 95/96, `compute-1.js`, `compute-2.js` and `publiclibrary-1.js` green — the one editor failure is `Edit mode: the deck browser says when a sort has nothing to sort by`, which reads the server's per-server play statistics and only passes on a server that has never counted any; it fails the same way on an unmodified `main` against this sandbox's `save/` directory and passes on both with an empty one.
* Playable demo of every scenario: [crash-guards-phase1.vtt](https://agent.virtualtabletop.io/reports/pr/1532402211575955567/crash-guards-phase1.vtt) — each button on the left used to crash, freeze or reload the client; the counter on the right shows the client is still alive after clicking all of them. Open the editor's Debug module to see the problem messages. ([screenshot](https://agent.virtualtabletop.io/reports/pr/1532402211575955567/crash-guards-phase1.png))

No widget property or routine-operation property was added, removed or renamed, so `validator/` and the `jeCommands` list need no changes.

Closes #1414, closes #684, closes #833, closes #1405, closes #1455, closes #1415.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3075/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

Solid, well-motivated hardening pass — the guards are placed sensibly and the four new TestCafe tests cover each scenario. The most important gap is that group B is only half-closed: `CALL` still does `JSON.parse(JSON.stringify(variables))` (widget.js:1180), so the very self-referential value that `StateManaged.set()` now tolerates still throws out of `evaluateRoutine` — and because that throw skips the `batchEnd()` at widget.js:2365, it leaves an unbalanced batch behind. Second-most important: making `seen` in `applyInheritedValuesToDOM` global to the whole walk instead of path-scoped silently breaks legitimate diamond inheritance.

## Findings

### 1. `CALL` still crashes on a self-referential value (widget.js:1180) — and leaks an open batch

```js
const inheritVariables = Object.assign(JSON.parse(JSON.stringify(variables)), a.arguments);
```

`set()` now refuses an unserializable value, but the variable itself stays circular in `variables`. Any `CALL` afterwards hits this line and throws `TypeError: Converting circular structure to JSON`. That exception propagates out of `evaluateRoutineOperations`, which means the `batchEnd()` at widget.js:2365 never runs while `batchStart()` at widget.js:992 already did — so the client is left with a permanently open batch, which is a worse outcome than the original crash.

Repro on this branch: take the new `'a value that contains itself…'` test and insert `{ func: 'CALL', widget: 'go', routine: 'clickRoutine' }` after the `push` line — the routine dies and the `go` marker never appears.

Two things worth considering: guard the `CALL` copy the same way `set()` is guarded (problem entry + skip, or deep-copy defensively), and — since "crash containment" is the stated goal — wrap the operation loop so `batchEnd()` runs in a `finally`. Any operation that throws for any reason currently leaves the batch open.

### 2. `applyInheritedValuesToDOM`'s `seen` set is walk-global, not path-scoped (widget.js:378-390)

`seen` is added to but never removed, and it is shared across sibling branches — unlike `inheritLookups` / `inheritPropagations`, which are correctly path-scoped with `try/finally`. That changes behaviour for legitimate (acyclic) diamond inheritance:

- `W.inheritFrom = { A: ['x'], B: ['y'] }`, `A.inheritFrom = { C: ['x'] }`, `B.inheritFrom = { C: ['y'] }`
- entries are walked reversed, so `B`'s branch runs first and marks `C` seen; `A`'s branch then skips `C` entirely
- result: `W` no longer inherits `x` from `C`, which it did before this PR

Making `seen` a path set (`seen.add(id)` before recursing, `seen.delete(id)` after) still terminates on genuine cycles and preserves the diamond case.

### 3. The parent-cycle guard is bypassed by multi-widget editing (jsonedit.js:1889 `jeApplyChangesMulti`)

The single-widget JE path routes through `updateWidget()` and is covered, but `jeApplyChangesMulti` writes properties with `widget.set(key, value)` directly (`setValueIfNeeded`), with no parent check at all — not even the pre-existing "does not exist" one. Select a holder and one of its children in the JE editor and set `"parent"` to the child's id, and you get exactly the #1414 state the PR is closing.

Smaller variant of the same gap: `updateWidget`'s new check is skipped when the id changes, because `widgets.has(widget.id)` is false for the new id — renaming a widget while simultaneously pointing its `parent` at one of its (about to be re-pointed) descendants slips through.

### 4. The 250 budget is consumed by IF/LOOP nesting, not just by `CALL` (widget.js:1425, 1559, 2164)

`evaluateRoutine` is also the entry point for `IF` branches, `loopRoutine` bodies and the internal `MOVE` sub-routine, so each of those levels costs a slot in `routineDepth`. A recursive game routine of the usual shape (`IF … then CALL self`) therefore gets roughly 125 real recursion levels, not 250 — and nested `LOOP`+`IF` structures cut it further. Worth a sentence in the message or a higher constant, so a maintainer debugging a legitimately deep recursive routine isn't misled by "more than 250 routines are running inside each other".

Related, lower severity: `routineDepth` is a module-level counter across `await` points, so genuinely concurrent (not nested) routines — a `globalUpdateRoutine` chain, a `DELAY`-suspended routine, an incoming click while another routine is awaiting — share the same budget. Very unlikely to matter at 250, but the counter measures "routines currently in flight", not nesting depth, which is a slightly different invariant than the message claims.

### 5. `routineDepthProblem` can be misattributed to an unrelated routine (widget.js:886, 2342)

The single module-global slot is only drained inside the per-operation loop. If the limit is hit on a path that isn't nested inside another routine's operation — e.g. a `changeRoutine` chain kicked off by an editor property write, or `gameStartRoutine` (serverstate.js:552) — nothing consumes the string, and it then surfaces as a problem in whatever unrelated routine happens to run next. Clearing it when `routineDepth` returns to 0 (or returning it in the result object) would keep the attribution honest.

### 6. The refused `set()` is invisible to the routine author (statemanaged.js:120-129)

The write is dropped with a `console.log` only. The routine's `problems` array — which is what the Debug module shows — says nothing, so from the routine log the `SET` looks like it succeeded and the property is just mysteriously unchanged. That's exactly the "don't crash, explain" half that's missing; the SET operation is the natural place to turn this into a problem entry, the way the new parent-cycle refusal does. The new test asserts `result` stays `null` but nothing asserts the user is told why.

### Note on verification

`node_modules` isn't provisioned in this sandbox clone, so I could not run jest/TestCafe myself; the review above is static. The reasoning for findings 1-3 is from the code paths, and finding 1 in particular is worth reproducing before dismissing.

I did not add the `beta` label: this PR changes core engine behaviour for real — routine `SET` semantics, `evaluateRoutine` recursion limits, inheritance resolution and `StateManaged.set()` — so it is excluded from beta auto-deployment by design.

VTT-AI-Agent:
Thanks — good catches, all six were worth acting on. Pushed 87d77a3d.

**1. `CALL` still crashes on a self-referential value / leaks an open batch** — confirmed and fixed, both halves.

The variable copy is now done key by key, so a variable that contains itself is left out and reported instead of taking the whole `CALL` down:

```js
const inheritVariables = {};
for(const [ key, value ] of Object.entries(variables)) {
  let json;
  try {
    json = JSON.stringify(value);
  } catch(e) {
    problems.push(`Variable ${key} contains itself and is not passed on to ${a.routine}.`);
    continue;
  }
  if(json !== undefined)
    inheritVariables[key] = JSON.parse(json);
}
```

The `json !== undefined` check keeps the previous behaviour of silently dropping keys whose value does not survive `JSON.stringify` at all (`undefined`, functions) — `JSON.parse(undefined)` would otherwise throw.

For the batch: `evaluateRoutine` now remembers the batch depth before running the operations and closes anything the routine left open in a `finally`, so *any* operation that throws for any reason can no longer wedge the client's delta sending. That needed a tiny `getBatchDepth()` export in `serverstate.js`. In the normal path the loop body never runs, so delta ordering is unchanged.

I took your repro suggestion literally and turned it into the test: the `'a value that contains itself…'` test now ends with `{ func: 'CALL', widget: 'go', routine: 'clickRoutine' }` instead of an inline SET. Verified it fails on the previous commit and passes now.

**2. `applyInheritedValuesToDOM`'s `seen` is walk-global** — confirmed, fixed as you suggested: `seen.add(id)` before recursing, `seen.delete(id)` in a `finally` after. Your exact diamond is now a TestCafe test (`a widget can still inherit the same widget through two different chains`) — `both` inherits `width` and `height` through two separate widgets that both inherit from `shared`. It fails with the walk-global set (`width` falls back to the default 100) and passes with the path-scoped one.

**3. Parent-cycle guard bypassed by multi-widget editing** — confirmed, both variants:
- `jeApplyChangesMulti`'s `setValueIfNeeded` now refuses a `parent` write that would create a cycle, collecting the affected ids and showing one `alert()` after the batch rather than one per widget.
- `updateWidget` now looks the widget up by `previousState.id` instead of `widget.id`, so the check also runs when the id is being changed in the same edit.

**4. The 250 budget is consumed by IF/LOOP nesting** — fair. I kept the constant (250 nested `evaluateRoutine` calls is still far beyond anything hand-written) but made the message honest: *"more than 250 routines, IF branches and LOOP bodies are nested inside each other"*, and the comment on `maxRoutineDepth` now says the same. On the in-flight-vs-nested point: agreed that the counter measures in-flight routines; at 250 it does not seem worth the bookkeeping to make it a true per-chain depth, and the message no longer claims more than it knows.

**5. `routineDepthProblem` misattribution** — fixed: the string is cleared when `routineDepth` drops back to 0, so a limit hit on a path with no enclosing routine operation can no longer surface inside an unrelated routine later.

**6. The refused `set()` is invisible** — fixed at the `SET` operation, as you suggested: a value that cannot be stored now produces `Skipping <property> of <widget> because the value contains itself and can not be stored.` in the routine's `problems`, next to the parent-cycle message. The `StateManaged.set()` guard stays as the backstop for the write paths that are not routine operations.

Verified: `npm test` (86 passed) and TestCafe `routines.js` + `editor.js` + `compute-1.js` (26 tests, all passing) locally.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review only (a separate reviewer covers code correctness). The change is almost entirely engine-side, so the UI surface is small: one new modal `alert()` in the widget-update path and three new user-facing messages. The messages read well, but **the one place they are meant to be read — the editor's Debug module — is where they are hardest to find, and the self-referential-value case still shows the "Internal Error / Your client just crashed" screen when the Debug module is open** (verified live, screenshot below). Most impactful fix: route the remaining routine-log `JSON.stringify(value)` calls through the new `stringifyForDisplay` and surface the new problem messages at the top level of the log.

Reviewed by running the branch locally (port 8353) and driving it with Playwright: a room with three buttons (`SET` a parent cycle, a routine that calls itself, a routine that builds a value containing itself), clicked with the Debug module open and closed.

---

## Big layout / flow improvements

**1. With the Debug module open, the self-referential value still takes the client down.**
`StateManaged.set()` no longer throws, but the routine log does: the `SET` summary line still formats the raw value —

`client/js/widgets/widget.js:2044` → ``jeLoggingRoutineOperationSummary(`'${a.property}' ${a.relation} ${JSON.stringify(a.value)} for widgets in '${a.collection}'`)``

Clicking a button whose routine does `var list = [] / var list = ${list} push ${list} / SET result = ${list}`:
* Debug module **closed** → nothing happens, client fine ✅ ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1532435030511587628/shot-selfvalue-nodebug.png))
* Debug module **open** → `TypeError: Converting circular structure to JSON` at `evaluateRoutineOperations`, and the room is replaced by the **"Internal Error — Your client just crashed. Sorry, this shouldn't happen."** page ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1532435030511587628/shot-selfvalue-debug.png))

That is exactly the situation the PR describes as "the routine log shows the error in place of the value instead of throwing while you are trying to debug it", so from a user's point of view the fix is inverted: opening the debugger is what breaks the client. `stringifyForDisplay` was applied to `widget.js:1116` and `jeLoggingJSON`, but the same pattern is used in ~20 other log summaries that format a routine value (`1141`, `1262`–`1268`, `1356`, `1543`, `1562`, `1634`, `1991`, `2044`, `2179`…). Suggest using `stringifyForDisplay` for all of them — one helper, one grep.

**2. The new problem messages are two clicks deep, under a green "No validation problems found!" banner.**
After the refused parent change the log shows a red `SET 'parent' = "inner" …` row; the actual sentence lives under a collapsed **Problems** node, and immediately below it the module says, in a green success banner, *"No validation problems found!"* ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1532435030511587628/shot-swap-expanded.png)). A first-time user reads the green banner, concludes nothing is wrong, and never learns why their `SET` did nothing.
Suggestions, in the module's existing idiom: auto-expand the `Problems` node when an operation has one (it is the only reason the row is red), or append the first problem sentence to the red summary row itself. Independently, the validation banner would read less contradictory as *"Run Validation found no problems"* rather than the absolute "No validation problems found!".

**3. The recursion message is unreachable in the UI it is written for.**
Clicking the self-calling button leaves the log with a single collapsed `CALL clickRoutine …` row ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1532435030511587628/shot-loop-expanded.png)). Because `routineDepthProblem` is consumed by the *innermost* routine that finishes, the sentence "…more than 250 routines are running inside each other" is attached ~250 nesting levels down. It is in the DOM, but reaching it means expanding 250 identical rows, and even typing `running inside each other` into the module's **Filter log…** box does not reveal it ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1532435030511587628/shot-recursion-filtered.png)). Consider also pushing the message into the outermost routine's problems (or emitting it once at the top-level routine entry) so the user sees it where they clicked.

**4. Three ways to set a parent, three different levels of feedback.**
* Edit Widgets sidebar → inline red 12px message under the *Parent widget* field (`properties.js:2665`), plus a dedicated `A widget cannot be its own parent.`
* JSON editor / *Edit widget JSON* → the new blocking `alert()` at apply time only. The live error bar (`jsonedit.js:2709`) already validates the sibling case `Parent ${…} does not exist.` while typing, but has no cycle check, so nothing warns you until the modal.
* JSON editor **multi-select** mode → `jeApplyChangesMulti()` (`jsonedit.js:1889`) writes `parent` through `widget.set()` and never reaches `updateWidget`, so this path gets no message at all.

Adding the cycle case to the `jeJSONerror` chain right after the existing parent check would make the JSON editor behave like it does for every other structural invariant (inline, live, non-blocking) and make the alert a last-resort backstop. Reusing `wouldCreateParentCycle()` in `properties.js setWidgetParent()` would collapse the duplicated rule into one.

## Small breakages / nitpicks

* `client/js/editmode.js:1573` — when the user sets a widget as its own parent the message becomes *"Widget outer is inside outer, so using it as the parent would create a loop."* The sidebar has a clearer sentence for that exact case; mirror it: `if(widget.parent == widget.id) → 'A widget cannot be its own parent.'`
* `client/js/editmode.js:1574` — the alert aborts the **whole** save (`batchEnd(); return`), so every other property the user edited in that JSON is discarded silently. The sentence should say so: *"… would create a loop. Nothing was saved."* (same applies to the pre-existing "Parent widget X does not exist." above it).
* `client/js/jsonedit.js:151` — the macro path calls `updateWidget()` once per widget in a loop. A macro that sets a bad parent now fires one **modal alert per widget** in the room, which has to be dismissed one by one. Collecting into `jeJSONerror` (as that code path already does for exceptions) avoids the dialog storm.
* `client/js/widgets/widget.js:2036` — *"Skipping parent change of outer to inner because that would put outer inside itself."* repeats the widget ID three times and uses a different metaphor ("inside itself") than the alert ("create a loop"). Neighbouring MOVE problems are terser: `Skipping move of X to its descendant Y.` Suggest `Skipping parent change of outer to inner: inner is inside outer.` and picking one metaphor for both messages.
* `client/js/domhelpers.js:119` — the fallback renders `{"error": "TypeError: Converting circular structure to JSON --> starting at object with constructor 'Array' --- index 0 closes the circle"}` where a value used to be. That is a V8 stack fragment in a place a game author reads values; something like `"<value contains itself and cannot be shown>"` (or the plain sentence plus the technical text underneath) fits the log's tone better.

## Readability & understandability (first-time user)

* **The refused write is invisible outside devtools.** `statemanaged.js:127` reports the self-referential value only via `console.log('Not setting result of build: TypeError: Converting circular structure to JSON …')`. Unlike the parent-cycle refusal it never becomes a routine *problem*, so in the Debug module the user sees the `SET` succeed-looking and the property simply not change. Pushing a problem — `Not setting ${property} of ${id}: the value contains itself.` — would match the PR's own "don't crash, explain" principle and the treatment the parent case already gets.
* **Wording of the new messages is otherwise good for newcomers** — "more than 250 routines are running inside each other. A routine is probably calling itself." is plain English and names the likely cause, which is exactly right. Worth keeping that style; consider naming the routine that re-entered (e.g. `… inside each other, starting at clickRoutine of loop.`) since by the time the user reads it they are 250 levels from the click.
* **Nothing is shown to a plain player.** All three guards are editor-only feedback; in a published game the button silently does nothing. That is consistent with how existing routine problems behave, so no change requested — just noting that "the client survives" is the whole of the player-visible improvement, which is a good trade.

VTT-AI-Agent:
Thanks - the Debug module finding was the important one, and it was exactly as described. Pushed afdf221d.

routine log with the value that contains itself

**1. With the Debug module open, the self-referential value still takes the client down** - confirmed and fixed. `stringifyForDisplay` was only used in two places; every other value formatting in the routine log used a raw `JSON.stringify`, and `SET` (widget.js:2073) was the one that threw for your repro. All of them now go through `stringifyForDisplay` (23 call sites in `evaluateRoutineOperations`, both the log summaries and the `problems.push` messages that format a value).

While doing that I also took your `domhelpers.js:119` nitpick: the fallback no longer prints the V8 sentence in place of a value, it decycles instead and shows `"<contains itself>"` where the loop closes, so the rest of the value stays readable - see the screenshot, the `push` result now reads `["<contains itself>"]`. It uses the same replacer idiom as the one in `tracing.js:73`.

New TestCafe test `the routine log shows a value that contains itself instead of crashing` opens the Debug module, runs the routine and asserts that `#clientErrorOverlay` is not there. It fails on the previous commit with your `TypeError` and passes now.

**2. The new problem messages are two clicks deep** - the `Problems` node is rendered expanded now, so opening the red row shows the sentence immediately (screenshot above). I left the red row itself collapsed: it is the existing affordance for "something is wrong here" and expanding every operation with problems would also open the two other detail nodes for games that have long-standing benign problems.

The green banner is a fair catch and it does sit right under the log, so I reworded it to *"No validation problems found in the game file."* - it is the game file validator's result, not the routine log's.

**3. The recursion message is unreachable** - fixed. The message is now reported twice: in the innermost routine, where it happens, and in the outermost one, which is the row the user actually clicked. New test `the recursion limit is reported in the routine that was clicked` asserts it is on the outermost operation's own `Problems` node (a descendant selector would have passed on the old code, so it checks the direct path).

**4. Three ways to set a parent, three levels of feedback** - all three behave the same way now:
- The JSON editor validates the parent live in the `jeJSONerror` chain, reusing the same `widgetParentProblem()` that `updateWidget()` uses, so it also covers the "does not exist" case it already had. Since the editor only applies when there is no `jeJSONerror`, the modal is now a backstop that the JSON editor does not normally reach. New test `The JSON editor reports a parent that would create a loop while typing` in `editor.js`.
- Multi-select gets the same live check (`jeMultiParentProblem`), so the whole apply is refused with an inline message rather than an alert afterwards. Note that this path was *not* silent on the branch you reviewed - `jeApplyChangesMulti` did show one `alert()` after the batch since 87d77a3d - but inline and live is better, and it removes the alert-per-keystroke this had.
- `setWidgetParent()` in `properties.js` now uses `wouldCreateParentCycle()` too, so the rule lives in one place.

**Nitpicks**, all taken:
- self-parent in the JSON editor now says `A widget cannot be its own parent.`, same as the sidebar
- the alert says `... Nothing was saved.` - and so does the pre-existing "Parent widget X does not exist." one
- the macro path collects the refused widgets into `jeJSONerror` (`N widget(s) were not updated: ...`) instead of one modal per widget, and updates all the other widgets normally
- the `SET` problems follow the neighbouring MOVE wording and one metaphor: `Skipping parent change of outer to inner: inner is inside outer.` and `Skipping parent change of inner to itself.`
- the refused value message is `Skipping result of build: the value contains itself and can not be stored.`

**Not changed:** nothing is shown to a plain player, as you noted - that stays consistent with every other routine problem.

Verified: `npm test` (86 passed) and TestCafe `routines.js` (11), `editor.js` (17), `compute-1.js`, `compute-2.js` locally, all passing. The two new routine tests and the new editor test were each confirmed to fail on 87d77a3d.

VTT-AI-Agent (commented):
## TL;DR

The earlier findings are addressed, but the new batch-recovery wrapper is not concurrency-safe: when two routines overlap across `DELAY` or `INPUT`, one routine closes the other's batch and the shared counter underflows. That can break delta delivery and undo grouping during ordinary play, so this still needs a fix and an interleaving regression test.

## Findings

### 1. Routine cleanup steals concurrent batches and can leave `batchDepth` negative (`client/js/widgets/widget.js:904-910`, `client/js/serverstate.js:322-324`)

`batchDepthBefore` is a snapshot of a single global counter, not ownership of this routine's batch. The unconditional `while` in `finally` therefore cannot distinguish nesting from another top-level routine that started while this one was awaiting:

1. Routine A enters at depth 0, calls `batchStart()` (depth 1), and waits in `DELAY` or `INPUT`.
2. Routine B enters while A is suspended, records `batchDepthBefore === 1`, calls `batchStart()` (depth 2), and also waits.
3. A resumes. Its normal `batchEnd()` changes 2 → 1, then A's `finally` sees `1 > 0` and calls `batchEnd()` again, closing B's still-live batch.
4. B resumes and its normal `batchEnd()` changes 0 → -1.

At depth -1, unbatched writes such as timer updates no longer flush because `sendDelta()` only sends when `!batchDepth`; normal batching also loses its atomic/undo boundary. Repeating the same overlap from -1 can leave the counter at -2, after which even a subsequent routine's `batchStart()` only reaches -1 and its changes remain unsent.

This overlap was safe before the wrapper: the two ordinary `batchEnd()` calls balanced the two starts and the last routine flushed the combined delta. Please pair each routine's own `batchStart()` with exactly one `batchEnd()` in a `finally` around its operation body instead of draining a shared counter back to a snapshot; any deeper batching site that may throw needs its own balanced `finally`. A regression test can start two buttons with staggered `DELAY`s (or hold one at `INPUT`), let the first finish while the second is suspended, then verify both a timer/direct write and a later routine write reach server state.

I did not add the `beta` label because the PR materially changes core engine behavior in `client/js/serverstate.js`, `client/js/statemanaged.js`, and `client/js/widgets/widget.js`, making it engine-ineligible for beta auto-deployment.

VTT-AI-Agent:
Thanks - confirmed, and it was a regression my own wrapper introduced. Pushed c26dd6b6.

**1. Routine cleanup steals concurrent batches and can leave `batchDepth` negative** - exactly as described. `batchDepthBefore` was a snapshot of a shared counter, not ownership, so the `while` in the `finally` happily closed the batch of whatever routine started while this one was suspended.

Fixed the way you suggested: each routine now opens exactly one batch and closes exactly one. The `batchStart()`/`batchEnd()` pair moved out of `evaluateRoutineOperations` (where the `batchEnd()` was simply skipped when an operation threw - the reason for the wrapper in the first place) up into `evaluateRoutine`, around the `try`/`finally`:

```js
    ++routineDepth;
    batchStart();
    try {
      return await this.evaluateRoutineOperations(property, initialVariables, initialCollections, depth, byReference);
    } finally {
      batchEnd();
      if(!--routineDepth)
        routineDepthProblem = routineDepthProblemForOutermost = null;
    }
```

That keeps the throw-safety and is balanced under any interleaving, since the counter is only ever moved by ±1 per routine. The `if(!depth && (this.isBeingRemoved || this.inRemovalQueue)) return;` early-out moved up with it so it still returns before a batch is opened, and the now-unused `getBatchDepth()` export is gone again.

The reordering this causes is that the two `toServer('playerColor')`/`toServer('rename')` calls at the end of `evaluateRoutineOperations` now happen just before the routine's `batchEnd()` instead of just after it. Those are direct emits and not part of the delta, and in the click path the `batchEnd()` never was the flushing one anyway (mousehandling holds an outer batch), so nothing depends on that order - the state-hash tests confirm it.

**Regression test**, following your repro: `a routine finishing while another one is waiting keeps the client sending` in `routines.js` starts two routines with staggered `DELAY`s so the first finishes while the second is still suspended, then does a write outside of any routine (`widgets.get('go').set('marked', true)` - the timer/direct-write case) and asserts all three widgets reach server state. On afdf221d the counter ends at -1 and that last write never leaves the client, so the test fails there; it passes now.

I did not touch the `batchStart()`/`batchEnd()` pairs in `mousehandling.js`, `editmode.js` and `jsonedit.js`. They can still leak a batch if something throws between them, but that is pre-existing and unchanged by this PR (the old wrapper never drained below its own entry depth, so it never covered them either) - worth its own change rather than widening this one.

Verified: `npm test` (86 passed) and TestCafe `routines.js` (12), `editor.js` (17), `compute-1.js` locally, all passing.

VTT-AI-Agent (commented):
## TL;DR

The findings from the earlier rounds look properly addressed — `CALL` copies variables per key, `applyInheritedValuesToDOM`'s `seen` is path-scoped again, the multi-widget JSON editor is guarded, and the batch wrapper now pairs exactly one `batchEnd()` with its own `batchStart()`. One group-A hole is still open and I reproduced it: **`RESET` writes `parent` without the new guard, so `resetProperties: {parent: <own descendant>}` still ends in `RangeError: Maximum call stack size exceeded` and a dead client** — the exact crash #1414 is about.

## Findings

### 1. `RESET` still creates a parent cycle and kills the client (`client/js/widgets/widget.js:1880-1890`) — confirmed

```js
if (a.func == 'RESET') {
  setDefaults(a, { property: 'resetProperties' });
  for(const widget of widgets.values()) {
    for(const [ key, value ] of Object.entries(widget.get(a.property) || {})) {
      if((key == 'parent' || key == 'deck') && value !== null && !widgets.has(value)) {
        problems.push(...);
      } else {
        await widget.set(key, value);   // <-- no wouldCreateParentCycle check
      }
    }
  }
```

`RESET` validates only that the target exists, exactly like `SET` did before this PR. Reproduced on this branch (Playwright, local server):

```json
{
  "outer": { "id": "outer", "type": "holder", "x": 50, "y": 50, "width": 400, "height": 400,
             "resetProperties": { "parent": "inner" } },
  "inner": { "id": "inner", "type": "holder", "parent": "outer", "width": 200, "height": 200 },
  "reset": { "id": "reset", "type": "button", "text": "reset", "x": 600, "y": 50,
             "clickRoutine": [ { "func": "RESET" } ] }
}
```

Clicking `reset` gives `RangeError: Maximum call stack size exceeded`, the room DOM is gone (a second button is no longer clickable), and nothing lands in the routine log. `resetProperties` with `parent` is a completely ordinary thing for a game to do (returning pieces to their starting holder), so this is not an exotic path — and it re-opens all the unguarded parent-chain walks the PR otherwise protects (`_absoluteRotation`/`_absoluteScale`/`_ancestor` at widget.js:2426-2434, `inheritSeatVisibility` at widget.js:2515-2517, `getValidDropTargets` at main.js:53-63).

Mitigating: the corrupted `parent` never reaches the server (the client dies inside the batch), so the saved game file stays clean and a reload recovers. But it is the same "dead client, no explanation" outcome the PR sets out to remove.

The narrow fix is to mirror the `SET` guard here. The broader one is worth considering instead: the invariant is currently enforced at every *call site* (`MOVE`, `SET`, `updateWidget`, `jeApplyChangesMulti`, `properties.js`, `applyInheritedValuesToObject`) and `RESET` shows how easy it is to miss one. Enforcing it once inside `Widget.set()` / `onPropertyChange('parent')` — refuse and let the caller pick up a problem — would make every present and future writer safe by construction, and would let most of those call sites keep only their nicer error messages.

For what it's worth, I checked the other `set('parent', …)` writers: drag & drop is fine, because `getValidDropTargets` (main.js:53-63) already walks the target's parent chain and rejects a target that has the dragged widget as an ancestor. `RESET` appears to be the only remaining one.

### 2. No test covers `RESET`

`tests/testcafe/routines.js` gets a good test per closed path, but `RESET` — the one still-broken writer — has none, which is presumably why it slipped through. A test in the shape of the existing `'SET parent refuses to put a widget inside itself'` (click, then assert the marker widget still reacts and the tree is unchanged) would catch it and any future regression.

Also, `canBeStored` and `stringifyForDisplay` (`client/js/domhelpers.js:117-149`) are pure functions with interesting edge cases — a cycle closing at the root, a value repeated as siblings without being a cycle (which the `ancestors`-stack replacer correctly *doesn't* flag), a `BigInt` hitting the inner `catch`. They are currently only exercised indirectly through one TestCafe test; a handful of jest cases would be cheap and would pin the replacer's semantics.

### 3. `routineDepth` counts routines in flight, not nesting depth (`client/js/widgets/widget.js:322-325, 910-918`)

This was raised in round 1 and the message wording was improved, but the counter is still module-global across `await` points, so genuinely *concurrent* routines share the budget rather than nesting inside each other. Two consequences, both low severity at a limit of 250:

* `if(routineDepthProblemForOutermost && routineDepth == 1)` (widget.js:2386) only fires when exactly one routine is live. If a routine is suspended in `DELAY`/`INPUT` while another one hits the limit, `routineDepth` never reaches 1 for the offender, and the message is either attached to the unrelated suspended routine's next operation or dropped entirely when the counter finally hits 0 (widget.js:916-917). The new "a routine finishing while another one is waiting" test proves the batch counter survives that interleaving; the problem-reporting state does not.
* A routine that is aborted by an exception before its operation-end block runs loses both strings the same way.

If it's worth tightening, carrying the message in the returned object (or on a per-outermost-call token) rather than in two module globals would make attribution exact.

### 4. Minor: `Widget.getDefaultValue` runs its cycle check outside the guard it depends on (`client/js/widgets/widget.js:2449-2455`)

```js
getDefaultValue(property) {
  const value = super.getDefaultValue(property);   // StateManaged.inheritLookups guard released here
  if(property == 'parent' && typeof value == 'string' && this.wouldCreateParentCycle(value))
```

`wouldCreateParentCycle` calls `widgets.get(id).get('parent')` for each ancestor, which can re-enter `getDefaultValue('parent')` on a widget that is already resolving — but by then `statemanaged.js:63`'s `finally` has removed it from `inheritLookups`, so the cycle guard no longer applies and each level starts a fresh `seen` set. I tried to build a repro (`A inheritFrom {B:['parent']}`, `B.parent = 'C'`, `C inheritFrom {A:['parent']}`) and could **not** make it recurse — the new `applyInheritedValuesToObject` skip (widget.js:376) stops `C` from ever acquiring an inherited `parent`, so the chain terminates. So this is a note about the guard's shape rather than a live bug: it holds only because a second, unrelated guard happens to cut the chain first. Doing the check inside `super.getDefaultValue`'s `try` (or giving `wouldCreateParentCycle` a shared re-entrancy set) would make it self-sufficient.

### 5. Nitpick: error-message ordering in `widgetParentProblem` (`client/js/editmode.js:1560-1570`)

The existence check runs before the self-parent check, so renaming a widget and pointing `parent` at the *new* id reports `Parent widget X does not exist.` rather than `A widget cannot be its own parent.` The new editor test only covers the case where the id is unchanged. Swapping the two checks would give the clearer message in both.

---

I did not add the `beta` label: this PR changes core engine behaviour for real — `evaluateRoutine`'s batch handling and recursion limit, `SET` semantics for `parent` and unserializable values, `StateManaged.set()`, and `inheritFrom` resolution — so it is excluded from beta auto-deployment by design.

VTT-AI-Agent:
Thanks - the `RESET` finding was real and I took the broader fix you suggested. Pushed 166be884.

**1. `RESET` still creates a parent cycle and kills the client** - confirmed, exactly your repro. Rather than adding a fifth copy of the check, the invariant now lives in `Widget.set()`:

```js
  async set(property, value) {
    if(property == 'parent' && typeof value == 'string' && this.wouldCreateParentCycle(value)) {
      console.log(`Not setting parent of ${this.id} to ${value}: ${value} is inside ${this.id}.`);
      return;
    }
    return await super.set(property, value);
  }
```

So every writer is safe by construction - `RESET`, `moveToHolder`, the pile code, and anything added later - while the call sites a user can reach keep checking first so they can report it in their own wording. `RESET` got those messages too, mirroring `SET`: `Skipping parent change of X to itself.` and `Skipping parent change of X to Y: Y is inside X.`, next to the `deck`/`parent` existence check it already had.

Worth noting for the "never break existing games" rule: the load path applies a game file through `applyDelta`, not `set()`, so a game that already contains a cycle keeps behaving exactly as it does today - the guard only refuses *new* writes.

**2. No test covers `RESET`** - added `RESET refuses to put a widget inside itself` in `routines.js`, in the shape you described (click, assert the marker widget still reacts and the tree is unchanged). It fails on c26dd6b6 with the client dying on the click and passes now.

Also took the jest suggestion: `tests/client/domhelpers.test.js` pins `canBeStored`/`stringifyForDisplay` with 8 cases - a loop closing at the root of an object and of an array, a loop deeper in a structure, the same object appearing twice as siblings *without* being a loop (the case the `ancestors` stack must not flag), the `indent` argument, and a `BigInt` falling through to the inner `catch`.

**4. `getDefaultValue` runs its cycle check outside the guard it depends on** - fixed, the check keeps its own re-entrancy set now (`parentDefaultLookups`), so it no longer depends on the `applyInheritedValuesToObject` skip happening to cut the chain first.

**5. Error-message ordering in `widgetParentProblem`** - swapped, self-parent is checked before existence.

**3. `routineDepth` counts routines in flight, not nesting depth** - I did not change the counter, and I think that is the right trade-off, but the "dropped entirely" half is fixed: when the counter returns to 0 with a message still pending it is now written to the console instead of being discarded, and the message names the routine it came from, so the information is never lost.

On carrying the message in the returned object or a per-chain token: the chains that matter here are exactly the ones that *don't* thread `depth` - `CLICK` → `click()`, change routines, `globalUpdateRoutine`. `CALL`/`IF`/`LOOP` do pass `depth`, so a token could ride along there, but a re-entry through `click()` would start a fresh chain and the "report it in the outermost routine" behaviour - the thing the UI review asked for, because the innermost one is 250 collapsed levels down - would be lost for precisely the recursion shape of #1455. Making it exact for both would mean propagating an async context across the suspension points (`DELAY`, `INPUT`), which is a bigger change than this PR should carry for an attribution nit at a limit of 250. Happy to do it separately if you'd rather have it.

Verified: `npm test` (94 passed, 8 of them new) and TestCafe `routines.js` (13), `editor.js` (17), `compute-1.js`, `publiclibrary-1.js` locally, all passing. No state hashes changed.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-scoped hardening: the guards are placed at the right layer (`Widget.set()` as backstop, user-reachable paths reporting in their own wording), the batching restructure is sound, and I verified locally that jest (94 tests), the new `routines.js`/`editor.js` TestCafe tests and `compute-1.js` (state hashes unchanged) all pass. The most important remaining point is a residual gap worth recording: a parent cycle can still arrive through the **delta receive path** (two clients racing complementary parent writes, or an unpatched client) and `applyDeltaToDOM`'s `appendChild` then throws uncontained on every client in the room.

## Findings

### 1. Cycles arriving via received deltas still crash every client (phase boundary — worth stating explicitly)

All new guards run on the client that *writes*. But nothing validates the invariant when a delta is *applied*: if client A sets `w1.parent = 'w2'` while client B (not yet having seen A's delta) sets `w2.parent = 'w1'`, both pass their local `wouldCreateParentCycle` check, the server serializes both deltas, and every client then executes `newParent.appendChild(this.domElement)` in `applyDeltaToDOM` (client/js/widgets/widget.js:270) with `newParent` being a descendant of the element being appended — a `HierarchyRequestError` that propagates out of `receiveDeltaFromServer` (client/js/serverstate.js:474, no try/catch on the `onMessage('delta', …)` path). The limbo pre-pass in `receiveDelta` only detaches widgets named in the *incoming* delta, so it doesn't help when the counterpart write arrived in an earlier delta. The same applies to a delta from an unpatched or hostile client. This is fine to defer (it needs an apply-side guard or containment, i.e. phase 2/3 of the report), but since the PR closes #1414/#684 I'd note it in the PR text or a follow-up issue so the delta-side hole isn't considered fixed.

### 2. Widgets inside a pre-existing cycle can no longer be edited at all until the cycle is fixed

Loading a game that already contains a cycle is deliberately left working ("only new writes are refused"). But `widgetParentProblem` (client/js/editmode.js:1552) and the same check in `jeGetContext` (client/js/jsonedit.js:2748) evaluate the *unchanged* `parent` value too: for a widget inside a pre-existing cycle, every edit — changing `text`, fixing a routine — is now refused with "would create a loop / Nothing was saved" unless the same edit also fixes the parent. That's arguably a feature (it forces the repair and the message says what to do), but it contradicts the "only new writes are refused" line. If unintended, only run the cycle check when `widget.parent` differs from `previousState.parent`.

### 3. The 250 budget is shared by concurrent routines, not just nested ones

`routineDepth` counts every in-flight `evaluateRoutine`, including routines parked in `DELAY`/`INPUT`. A game that stacks many concurrent suspended invocations (e.g. a tick-driven `globalUpdateRoutine` containing a long `DELAY`) can spuriously hit the limit with no chain nested deeper than 2–3, and the problem entry can land on whichever unrelated routine's operation completes next (the message text does name the true origin, so this part is cosmetic). Given that the re-entry paths never passed `depth` through, the shared counter is the pragmatic choice — just flagging the false-positive scenario so it's a known trade-off; the generous limit makes it unlikely in practice.

### 4. Minor: `canBeStored` adds a redundant serialization on the hottest path

The SET operation (client/js/widgets/widget.js:2087) now runs `JSON.stringify` over every computed value once in `canBeStored` before `StateManaged.set()` stringifies it again — for large values (canvas `map`s, long arrays) inside `FOREACH` loops that's measurable. Since `StateManaged.set()` already catches the cyclic-value `TypeError` itself, a cleaner shape would be for `set()` to return `false` when it refuses a write and let the SET operation report the problem from the return value — that also removes the duplicated self-parent/cycle message logic between SET, RESET and `set()`.

## Standing checks

- **UI support:** no new widget property or routine-operation parameter is exposed to game authors, so no property-editor/block-UI work is needed; validator and `jeCommands` correctly untouched.
- **Test room / library:** no `library/` files are touched ✓; the demo room with one repro button per crash scenario is published on the PR's own test server rather than committed — correct handling.
- **Claim check:** the approach matches what the maintainer prescribed on #1405 (limit + problem report), including the concern there about the problem being buried in the deepest routine, which the outermost-routine duplication addresses.

## Verification performed

Ran locally on the PR branch: `npm test` (94 jest tests, 6 suites, including the new `domhelpers.test.js`), TestCafe `routines.js` (13/13 incl. the 9 new tests), the new `editor.js` test, and `compute-1.js` as a state-hash regression canary — all green.

---

I did not add the `beta` label: this PR genuinely changes engine behavior — `client/js/widgets/widget.js` and `client/js/statemanaged.js` alter the routine interpreter (recursion limit, per-routine batching), `Widget.set()` semantics and inheritance resolution for all existing games — and engine changes are excluded from beta auto-deployment by design.

VTT-AI-Agent:
Thanks — two of these turned into changes, one is a correction to my own PR text, and one stays as it is. Pushed 083eee78.

### 4. `canBeStored` adds a redundant serialization on the hottest path

Taken, in the shape you suggested. `StateManaged.set()` and `Widget.set()` now return the reason when they refuse a write, and `SET`/`RESET` report it from the return value:

```js
const newValue = await compute(a.relation, null, w.get(String(a.property)), a.value);
const refused = await w.set(String(a.property), newValue);
if(refused)
  problems.push(`Skipping ${refused}.`);
```

That removes the second `JSON.stringify` per SET and, with it, the duplicated self-parent / parent-cycle / unserializable logic that `SET`, `RESET` and `set()` each carried — the three problem messages are byte-for-byte the same, they are just built in one place now. `canBeStored` had no other caller and is gone; its two jest cases went with it, `stringifyForDisplay` keeps the remaining 7.

One incidental fix that fell out of giving `set()` a return contract: the `z > 90000` branch used to `return await resetMaxZ(...)`, which would now have been read as a refusal reason if `resetMaxZ` ever returned anything. It returns explicitly instead.

### 3. The 250 budget is shared by concurrent routines

Fixed rather than just noted — your `DELAY`-in-a-tick-loop example is reachable enough that I did not want to leave it. A routine waiting in `DELAY` or `INPUT` now gives its level back for as long as it waits:

```js
async function whileSuspended(promise) {
  --routineDepth;
  try {
    return await promise;
  } finally {
    ++routineDepth;
  }
}
```

so the counter measures what is actually nested inside what. A tick-driven `globalUpdateRoutine` with a long `DELAY` no longer eats the budget; a routine that calls itself still does, because nothing in that chain is suspended. New test `routines that are waiting do not use up the nesting limit` parks 300 routines in `DELAY` and then starts one more — on 166be884 the 301st is refused and never marks, so the test fails there and passes now.

The attribution half you called cosmetic stays as it is, for the reason from the previous round: making it exact needs an async context carried across the suspension points, which is bigger than this PR should be.

### 2. Widgets inside a pre-existing cycle can no longer be edited at all

The change is in, but the premise needs correcting — and it was my PR text that put it there, so this one is on me. **A game file containing a parent cycle never produces widgets in a cycle.** `receiveStateFromServer` defers a widget whose parent has not been added yet and drops it if the parent never arrives, so both widgets of a cycle are simply not added. Loading `{outer: {…, parent: 'inner'}, inner: {…, parent: 'outer'}}` against this branch gives an empty room and

```
Could not add widget "outer" because its parent "inner" does not exist!
Could not add widget "inner" because its parent "outer" does not exist!
```

That is pre-existing behaviour and unchanged here, so "a game that already contains a cycle keeps behaving exactly as before — only new writes are refused" was right about the outcome and wrong about the mechanism. I have rewritten that paragraph.

What is left of your point is still worth having, so `widgetParentProblem` and `jeMultiParentProblem`/`jeApplyChangesMulti` now only check the parent when the edit actually changes it (or renames the widget). It keeps a widget that ends up in a loop through the delta race below repairable, and it stops a macro over N widgets from walking N parent chains it is not touching. I could not write a test for it: as above, there is no way to get a client into that state on purpose without going through the crash in finding 1.

### 1. Cycles arriving via received deltas

Real, and now stated explicitly in the PR body under "What this does not cover yet" rather than fixed here — with one addition worth recording: **containing the `appendChild` alone would not be enough.** Once the cyclic `parent` is in `state`, the parent-chain walks behind `_absoluteRotation`, `_absoluteScale` and `_ancestor` (`widget.js`) and `inheritSeatVisibility` have no seen-set, so the next read overflows the stack instead of the append throwing. Making the receive path safe means either an apply-side rejection — which desyncs that client from the server — or bounding all of those walks, i.e. phase 2/3 of the report, exactly as you say.

### Verification

jest 93 (the two `canBeStored` cases removed), TestCafe `routines.js` 14/14 (incl. the new test), `editor.js` 17/17, `compute-1.js`/`compute-2.js` as state-hash canaries — all green, no hashes changed.

VTT-AI-Agent:
## Merged `main` into this branch

> Update from main and resolve merge conflicts.
>
> — [lawdawg96 on Discord](https://discord.com/channels/770758631146782780/1532402211575955567/1545122302855741550)

Merged `ec1c965e` (71 commits) in `d90635a6`. Six files conflicted; this is what each resolution does.

| File | Resolution |
| --- | --- |
| `client/js/main.js` | The window handover to the edit-mode bundle moved into a `try` and gained a lot of names. `stringifyForDisplay` stays in the list — that handover is what makes it reachable from `jsonedit.js`, which is in the other bundle. |
| `client/js/editor/sidebar/properties.js` | `setWidgetParent` now loops over the targets of a multi-selection. The loop check is `target.wouldCreateParentCycle(newParent.id)` instead of `newParent.isDescendantOf(target)`. |
| `client/js/jsonedit.js` (1) | `jeMultiParentProblem` is gone; the loop check moved into main's `jeMultiWidgetReferenceError`, so a per-widget `parent` object is resolved once and then checked for existence *and* for a loop, and the whole check is skipped while the `widgets` array itself is being edited. |
| `client/js/jsonedit.js` (2) | The routine log keeps main's one-readable-sentence-per-problem rendering and this branch's expanded Problems block; non-string problems go through `stringifyForDisplay`, so a value containing itself still shows instead of throwing. |
| `client/js/widgets/widget.js` | Import line combined. Main's `var` expression summary keeps its "only show the substituted expression when it differs" behaviour and prints the result with `stringifyForDisplay`. |
| `tests/testcafe/editor.js`, `tests/testcafe/routines.js` | Both sides had appended to the end of the file; this branch's tests now follow main's. |

Two behaviour details fell out of the merge:

* **`Parent widget X does not exist.` → `Parent X does not exist.`** Main's new JSON editor tests pin the wording `Parent nope does not exist.` (and `Parent nope does not exist (widget "one").` for a selection). `widgetParentProblem` now returns the same sentence, so the single-widget editor, the multi-selection and the `updateWidget` alert all say the same thing.
* **The Debug panel test no longer clicks the Problems expander open.** Problems come up expanded on this branch, so the click main's test does would have closed them again; the test asserts the block is `active` instead.

Verified locally after the merge: jest 1657/1657, TestCafe `routines.js` 21/21, `editor.js` 95/96, `compute-1.js` + `compute-2.js` (state hashes unchanged) and `publiclibrary-1.js`.

The one `editor.js` failure is `Edit mode: the deck browser says when a sort has nothing to sort by`. It asserts the "nothing to sort by" hint, which the client only shows when no game on the server has any recorded play time — and this sandbox's `save/statistics.json` has counted some. It fails the same way on an unmodified `main` checkout pointed at this `save/` directory, and passes on both this branch and `main` with an empty one, so it is the sandbox rather than the merge.

96LawDawg:
The debugger info for CALL needs to be changed to "Not running ... of ... more than 250 times. This is likely a recursive routine calling itself." Make a similar change for CLICK: "Not running ... of ... more than 250 times. This is likely a recursive click on itself."

VTT-AI-Agent:
Done in `1255250c`.

> The debugger info for CALL needs to be changed to "Not running ... of ... more than 250 times. This is likely a recursive routine calling itself." Make a similar change for CLICK: "Not running ... of ... more than 250 times. This is likely a recursive click on itself."

The limit check sits in `evaluateRoutine`, which did not know how it had been re-entered, so it said the same thing for every path. `CALL` and `click()` now record it immediately before they hand over, and the check picks the matching hint:

| How the routine was re-entered | Message |
| --- | --- |
| `CALL` | `Not running clickRoutine of loop more than 250 times. This is likely a recursive routine calling itself.` |
| a click — the `CLICK` operation or a click on the table | `Not running clickRoutine of loop more than 250 times. This is likely a recursive click on itself.` |
| anything else (IF branch, LOOP body, change/enter/leave routine) | `Not running clickRoutine of loop more than 250 times. Routines, IF branches and LOOP bodies all count towards that limit, so something is probably triggering itself.` |

The hint goes on `click()` rather than on the `CLICK` operation on purpose: every click runs through it, so a `clickRoutine` reached by a hotkey or by a click on the widget is described the same way as one reached by `CLICK`.

The third row is what a path neither of us named falls back to — it keeps the sentence useful when the loop runs through, say, a `changeRoutine`, without claiming a `CALL` or a click is at fault. Say the word if you would rather have wording of your own there.

`tests/testcafe/routines.js`: `the recursion limit is reported in the routine that was clicked` now pins the `CALL` sentence, and a new `a routine clicking its own widget is reported as a recursive click` pins the click one — a button whose `clickRoutine` is `CLICK` on `thisButton`. Verified locally: `routines.js` 22/22 green.

96LawDawg (approved):

VTT-AI-Agent (commented):
## TL;DR

The PR is in good shape: every point from the earlier rounds is addressed, the branch merges cleanly onto current `main`, CI is green (53/53) and I re-verified locally on `1255250c` — jest 1657/1657, TestCafe `routines.js` 22/22 and the new `editor.js` test. The one thing worth fixing before merge is in the newest commit: **the CALL/click hint that was requested on 2026-09-03 only appears for some nesting shapes.** With two `IF` levels around the `CALL` (or `CLICK`) the message degrades to the generic sentence and even loses the routine name (`Not running routine of loop …`). Reproduced below; the fix is small.

## Findings

### 1. The "recursive routine calling itself" / "recursive click on itself" hint depends on nesting parity (`client/js/widgets/widget.js:1253-1257`, `:623`, `:1607`) — reproduced

`routineEntry` is a single slot that the very next `evaluateRoutine` consumes and clears, so it only describes the operation that entered the level that gets *refused*. An `IF` branch or `LOOP` body is its own level and always enters with `null`. Which level is the 251st therefore depends on the period of the recursion:

| `clickRoutine` of `loop` | period | 251st level | outermost problem text on this branch |
| --- | --- | --- | --- |
| `[ IF → CALL ]` | 2 | `CALL` | `Not running clickRoutine of loop more than 250 times. This is likely a recursive routine calling itself.` |
| `[ IF → IF → CALL ]` | 3 | an `IF` branch | `Not running routine of loop more than 250 times. Routines, IF branches and LOOP bodies all count towards that limit, so something is probably triggering itself.` |
| `[ IF → IF → CLICK thisButton ]` | 3 | an `IF` branch | same generic sentence — the "recursive click on itself" hint never appears |

(Run against this branch with the Debug module open, exactly like the PR's own `the recursion limit is reported in the routine that was clicked` test; the `IF` is the plain `operand1: 1, relation: '==', operand2: 1` form.)

`IF … then CALL self` is the canonical shape of a recursive routine, and adding a second condition, an `elseRoutine`, or a `FOREACH` around the `CALL` is enough to flip the outcome. The second half of the degradation is the name: for a branch `property` is the array, so the sentence says `routine of loop` instead of `clickRoutine of loop`.

Suggested fix, in the shape the code already has: let inline branches inherit the entry and the name of the routine they belong to. `evaluateRoutine` already has `entry` and `property`; hand both into `evaluateRoutineOperations`, and have the `IF` (`:1982`) and `FOREACH` (`:1848`) branch calls set `routineEntry = entry` immediately before `this.evaluateRoutine(...)`, the same way `CALL` does at `:1607` — plus carry the enclosing `property` so the message keeps naming `clickRoutine`. (A per-routine stack popped in the `finally` works too, but has to remove its own frame by identity because routines suspended in `DELAY`/`INPUT` interleave.) Then add an `[ IF, IF, CALL ]` variant to the two limit tests — both currently use a bare top-level `CALL`/`CLICK`, which is why the parity dependence is invisible to them.

### 2. Small notes (no change required)

* Attribution of the outermost problem while an unrelated routine is suspended at depth ≥ 2 is the accepted trade-off from earlier rounds; nothing new there and `whileSuspended` handles the realistic case.
* The 250 limit and the refused parent cycle are creator-facing behaviour. The PR body says so, which is what the conventions ask for; a line on the wiki's Functions page (under `CALL`) once merged would save the next person a search.

## Standing checks

* **UI support:** no widget property or routine-operation parameter is added, so no property-editor / block-UI work is needed; `validator/` and `jeCommands` correctly untouched.
* **Test room / library:** no `library/` file is touched ✓. The demo room is published as a `.vtt` (the link in the PR body still resolves) and available on the PR's own test server rather than committed — correct handling for a hardening PR with no new property.
* **Engine impact:** `client/js/widgets/widget.js` and `client/js/statemanaged.js` change how every existing game runs — per-routine batching and the nesting limit in the routine interpreter, `set()` refusing parent cycles and unserializable values, inheritance resolution — so this is a genuine engine change, not an incidental one.

## Verification performed

On the PR head (`1255250c`) in this sandbox: `npm test` 1657/1657 (57 suites), TestCafe `routines.js` 22/22 (all 11 new tests included), the new `editor.js` test 1/1, plus the three-case repro above. `git merge-tree` against current `main` (`9a40e5bd`) reports no conflicts. All 53 GitHub checks (CodeQL, Chrome + Firefox TestCafe shards, jest, browser compatibility) pass.

I did not add the `beta` label: this PR genuinely changes engine behaviour — the routine interpreter's batching and nesting limit, `Widget.set()`/`StateManaged.set()` refusals and inheritance resolution in `client/js/widgets/widget.js` and `client/js/statemanaged.js` apply to every existing game — so it is excluded from beta auto-deployment by design.

VTT-AI-Agent (commented):
## TL;DR

The guards hold up on the current head (`1255250c`): jest 1657/1657, TestCafe `routines.js` 22/22, the two touched `editor.js` tests and the `compute-1.js` state-hash canary all pass locally, and adversarial probing found no crash and no false positive (300 routines parked in `DELAY` resume together without any of them being refused). The one point that deserves a maintainer decision: the 250 budget counts `IF` branches and `LOOP` bodies too, so what a game actually gets is **125 self-CALLs** through the usual `IF → CALL` base case and **83** through `FOREACH → IF → CALL` — and the public library already uses self-`CALL` recursion as its loop idiom in 42 places. Nothing in the library is deep enough to hit the cap today, but the number in the message is not the number a game gets.

## Findings

### 1. The effective recursion budget is 125 (or 83), not the 250 the message says — and recursion-as-loop is an established library idiom

`client/js/widgets/widget.js:1255` compares `routineDepth` against `maxRoutineDepth` (`:50`), and every `evaluateRoutine` entry counts, including the `byReference` ones for `IF` branches (`:1982`) and `FOREACH` bodies (`:1848`). A recursive routine almost always has its `CALL` inside the `IF` that checks the base case, so each real recursion step costs two levels; with a `FOREACH` around it, three. Measured on this branch with a countdown routine (`IF n > 0 → var m = n - 1 → CALL self with n = m`):

| Pattern | Succeeds | Aborted |
| --- | --- | --- |
| `IF → CALL` | n = 124 | n = 125 ("Not running clickRoutine of down more than 250 times. This is likely a recursive routine calling itself.") |
| `FOREACH → IF → CALL` | n = 82 | n = 83 (generic hint, since the re-entry was the loop body) |

So the message tells the author their routine ran "more than 250 times" when it ran 125 (or 83) times, and the aborted `CALL` hands back `{ variable: null, collection: [] }` while the caller keeps going — a loop written as recursion is silently truncated half-way rather than refused outright.

This is not hypothetical for this codebase. A scan of `library/games` and `library/tutorials` for `CALL` operations that re-enter the routine they sit in finds **42 sites in about 30 games**, used as the routine language's while-loop: Turtle Tower's `doWhileRoutine` walks a stack of turtles card by card (70 cards in the file), Solitaire variant 2's `recursiveSearchRoutine` walks a parent chain of cards (52), Shattered Dominion's `getAreaRoutine` flood-fills a 5×5 map through `IF → FOREACH → CALL` (three levels per cell, up to 25 cells), Logicards / Paletto / Property Poobah / Ultra re-draw random results until they are valid, The Team II loops until a point budget is spent, and a dozen `NextTurnRoutine`s skip seats recursively. All of these stay well under the cap with their current data, so I found **no library regression** — but the idiom is real, and the first game that walks 130 cards or a 10×10 grid this way will be cut off with a message that blames recursion.

Two ways out, either is fine:

- Count only real re-entries: `evaluateRoutine` already receives `byReference`, and `IF`/`FOREACH`/`Moves` bodies are lexically nested and cannot recurse on their own — only `CALL`, `CLICK`/`click()` and the change/enter/leave routines can. `if(!byReference)` around the check and the `++`/`--` (with `whileSuspended` told whether its routine counted, e.g. a second parameter) makes "250" exact and keeps the wording as the maintainer asked for it in the last round.
- Or keep counting everything but say so: "more than 250 nested routines, IF branches and LOOP bodies" — and consider a higher number, since the limit exists to catch infinite recursion, not to bound legitimate work (the interpreter is async, so there is no stack to protect).

### 2. Test coverage: the `INPUT` suspension path has no test

`whileSuspended` (`widget.js:67`) wraps both `DELAY` and `INPUT`, but only `DELAY` is exercised (`routines that are waiting do not use up the nesting limit`). `INPUT` is the more interesting one: it is the path where `runInput` can reject (cancel) and the level has to come back through the `finally`, and it is where a real player sits for minutes while other routines run. A variant of the existing test that parks routines in `INPUT` instead of `DELAY`, cancels one, and checks the counter is still balanced would pin it. Low priority — I could not make the code misbehave here, it is just unpinned.

## Standing checks

- **UI support:** no new widget property or routine-operation parameter is exposed to authors, so no property-editor / block-UI work is due; `validator/` and `jeCommands` are correctly untouched. The one creator-facing behaviour change is the limit itself — worth a line on the wiki's `CALL` / Functions page once merged.
- **Test room / library:** no `library/` file is touched ✓. The demo room is published as a `.vtt` on the PR report page; I cannot reach the test server from this sandbox, so I did not verify that it is in the PR's `pr-test-ai` room.
- **Delta receive path:** the residual hole (two clients writing complementary `parent`s) is stated in the PR body rather than claimed fixed ✓.

## Verification performed (head `1255250c`, local server on its own port)

- `npm test`: 1657/1657.
- TestCafe (Chromium headless): `routines.js` 22/22; `editor.js` `the Debug module logs each operation of a routine with its result` and `The JSON editor reports a parent that would create a loop while typing` 2/2; `compute-1.js` 1/1 — state hashes unchanged, and the diff changes no `compareState` hash.
- Playwright probes: the countdown measurements above; 300 routines parked in `DELAY` that resume together and each enter an `IF` — all 300 complete, no "Not running", no page errors; the hint selection (`CALL` vs generic) matches the re-entry path.
- Library scan for self-`CALL` recursion: 42 sites, none deep enough to hit the cap with current data.

---

I did not add the `beta` label: this PR genuinely changes engine behaviour — `client/js/widgets/widget.js` and `client/js/statemanaged.js` alter the routine interpreter (nesting limit, per-routine batch lifecycle), `Widget.set()` / `StateManaged.set()` refusal semantics and `inheritFrom` resolution for every existing game — so it is excluded from beta auto-deployment by design.

VTT-AI-Agent:
Both findings are in, in `09e6e731`.

### 1. The effective budget is now the 250 the message names

Taken as you suggested: only routines that were entered by name count. An `IF` branch, a `FOREACH` body and the move list of a `Moves` operation are the three `byReference` call sites, they are written inside the routine that runs them and they cannot re-enter anything on their own — their nesting is bounded by the game file — so they now share the level of the routine they belong to instead of taking one of their own.

`evaluateRoutine` decides this once (`!byReference || typeof property == 'string'`, so a named routine still counts even if it is reached by reference) and hands it to `evaluateRoutineOperations`, which needs it for the "repeat the problem in the outermost routine" check — that one has to stay the outermost *routine* rather than "whatever runs at depth 1". `whileSuspended` is unchanged: an inline branch shares the enclosing routine's level, so giving exactly one level back is right wherever the wait sits.

Measured on the new head with your countdown, and the numbers are now flat:

| Pattern | Succeeds | Aborted |
| --- | --- | --- |
| bare self-`CALL` | 250 | 251 |
| `IF → CALL` | 250 | 251 |
| `IF → IF → CALL` | 250 | 251 |
| `IF → IF → CALL`, with one routine parked in `INPUT` | 250 | 251 |

The hint got exact as a side effect, which also closes the parity finding from the previous round: the level that is refused is now always the `CALL` or the click that re-entered, so `IF → IF → CALL` says `Not running clickRoutine of loop more than 250 times. This is likely a recursive routine calling itself.` instead of degrading to the generic sentence and to `routine of loop`. The generic wording is now `Something is probably triggering it again while it is still running.` — it is left for the paths that really are a re-entry without a `CALL` or a click (a change routine, an `enterRoutine`).

One thing fell out of raising the effective budget: **the Debug panel crashed on the way there.** 250 levels of `IF → IF → CALL` nest deeper than Chrome's HTML parser keeps (it stops at 512 elements and appends the rest as siblings), which leaves `.jeLogOperation` entries without the expander `jeLoggingRenderLog` dereferenced — `Cannot read properties of null (reading 'classList')`, i.e. the client goes down while you debug the very routine this PR is about. It did not show up before because the old counting aborted `IF → CALL` at 125 levels, just under the parser limit. `jeLoggingRenderLog` now works on what actually made it into the DOM; the expanders still toggle, and the test below covers it.

### 2. `INPUT` suspension is pinned now

`routines that are waiting for input do not use up the nesting limit`: one routine parks in an `INPUT`, then the counter recurses the full 250 — which only gets through if the waiting routine gave its level back — and after the overlay is cancelled (the reject path through `whileSuspended`'s `finally`) a reset and a second full 250 run show the counter is balanced rather than one short.

### Tests

Three new ones in `tests/testcafe/routines.js`, each of which fails on `1255250c`:

* `a routine calling itself from inside IF branches gets the whole nesting limit` — 250 self-`CALL`s through two `IF` levels complete (125 before).
* `a routine calling itself once too often is refused and named` — 251 is refused, and the outermost problem names `clickRoutine of counter` with the `CALL` hint (generic hint and `routine of counter` before; before the `jsonedit.js` fix this test took the client down instead).
* `routines that are waiting for input do not use up the nesting limit` — as above.

Verified locally: jest 1657/1657, `routines.js` 25/25, `editor.js` `the Debug module logs each operation of a routine with its result`, `Enabling the Debug module while a routine waits for INPUT does not abort the routine` and `The JSON editor reports a parent that would create a loop while typing` 3/3, `compute-1.js` green with unchanged state hashes.

The 42 library sites you found keep working the same way — none of them was near the cap, and each of them now has twice the headroom it had.

PR #3076, opened 2026-07-30
Deck editor: multi-select, grouped properties, and a non-modal Edit Widgets panel for decks

VTT-AI-Agent:
Three deck-editor enhancements that work together: being able to select more than one face object, a properties sidebar that is organized the way the Edit Widget sidebar is, and selecting a deck widget no longer jumping straight into the deck editor.

## Multi-select of face objects

- **Ctrl/Cmd+click** on an object — on the card or in the left tree — adds it to (or removes it from) the selection; **Shift+click** selects the range from the current object to the clicked one, like a file list — the range becomes the whole selection, while **Ctrl+Shift+click** adds it to what is already selected; **Ctrl+A** selects every object of the shown face. A plain click still selects exactly one, as before. Every selected object gets the orange outline and its tree row is highlighted.
- The **Object tab** then edits all of them at once: each property row writes its value to every selected object, and a row the selected objects disagree about shows an empty `(mixed)` field until a new value is typed (so nothing is overwritten by accident). A property only some of them have still gets a row, so it can be handed to the whole group in one go. `type` works the same way. Bindings (Dynamic properties) stay a single-object affair — they name a per-object card-type property, so with several objects selected the section says so instead of editing anything.
- **Dragging** one object of a selection moves the whole group (keeping its arrangement), and the drag toolbar's **move / resize / rotate** apply to all of them — resize offsets each object from its own size, so size differences are preserved. **Copy** and **Delete** (button or Delete key) act on the whole selection too.
- **New align/distribute buttons in the Object tab's toolbar**, right above the selection they act on: align left / center / right / top / middle / bottom and equalize horizontal / vertical spacing. They use the objects' rendered boxes and write the result back in card coordinates — the same approach `client/js/editor/toolbar/align.js` takes for widgets, so rotated and auto-sized objects line up by what is actually drawn. Aligning needs two selected objects, distributing needs three; below that the buttons are disabled rather than hidden, and their tooltip says how to select more.

## Grouped properties in the sidebar

The property rows of the **Object** and **All Cards** tabs are no longer one flat list. They are sorted into the same collapsible blocks the Edit Widget sidebar uses — **Content, Position, Size, Colors, Appearance** — with everything the engine doesn't know (a game's own bespoke properties) collected in a **Custom** block of its own. Each block's header carries a summary of what is inside it, so a folded block still says what it holds, and Position/Size start folded on the Object tab (they are usually changed by dragging). Which blocks are folded is remembered per tab for the session. A list that would end up in a single block is drawn ungrouped, since such a header only repeats the rows below it.

The **Card Type** tab stays flat: every property of a card type is defined by the game, so sorting them by the engine's property names would be a coincidence of naming rather than a real distinction.

Everything stays nested inside the deck editor and under the existing tabs — the tab bar, the per-tab add/copy/delete toolbar and the scope colors are unchanged.

Also fixed along the way: selecting an object in the tree while the card-defaults view was showing left the card area on its placeholder text, so the selection outline, drag toolbar and the new alignment buttons had nothing to work with. It now renders the card again.

## Selecting a deck widget no longer opens the deck editor directly

Requested by @96LawDawg:

> Clicking on a deck widget while the Edit Mode tab is "Edit Widgets", or when already clicked on a deck and switching from some other tab to "Edit Widgets" should not immediately open the deck editor. Instead it should open a very abbreviated UI that looks like the UI for other widgets at the top and have the "Basic" section:
>
> However, for a deck, above the basic section, add a button to open in the deck editor. The only other section in the UI should be "Other properties" that lists anything not in the Basic section and anything not covered by the deck editor in card defaults, face templates, etc.
>
> Finally, next to the "Card View" button in the upper right, (so it will be right next to "Edit Widgets", add a button to close the deck editor.

Selecting a deck in the Edit Widgets tab now shows the same abbreviated sidebar as every other widget type: a type header, an **Open deck editor** button above everything else, the **Basic** section (Position/Size/Generic, same as any other widget), and an **Other properties** section for anything not in Basic and not owned by the deck editor (`cardDefaults`, `cardTypes` and `faceTemplates` are excluded, since those are edited inside the deck editor itself).

A **Close** button now sits next to **Card view** in the deck editor's upper-right corner, alongside the existing ways to close it (the toolbar's Deck editor toggle button and Escape).

Along the way this surfaced an unrelated bug: the main editor toolbar's hover tooltips had no `pointer-events: none`, so a tooltip could silently swallow a click aimed at whatever sat beneath it — including, at some window widths, the deck editor's own Card view/Close buttons. Fixed by making those tooltips click-through, like tooltips elsewhere in the app already are.

## Screenshots

Three text objects selected at once, with the grouped properties and the alignment row above them:

multi-selected face objects

After aligning them left and equalizing their vertical spacing (note the breadcrumb):

aligned and evenly spaced

Selecting a deck in Edit Widgets now shows the abbreviated Basic/Other properties panel instead of opening the deck editor:

abbreviated deck panel

The deck editor open, with the new Close button next to Card view:

deck editor with close button

## Notes

- No widget/routine properties were added, renamed or removed, so `validator/` and the JSON editor's `jeCommands` need no update. Saved games are untouched — this is editor UI only.
- Tests: `npm test` (jest) passes; the full TestCafe `editor.js` suite passes locally with **unchanged state hashes**. A new test — *"Deck editor: multi-selected face objects share property edits and alignment"* — covers Ctrl+click multi-selection, the grouped blocks, the `(mixed)` shared edit and align/distribute. Three existing test helpers switched from `#deckEditorSidebar > .deckEditorProperties:first-of-type` to `.deckEditorObjectProperties`, since the object rows now live inside their group blocks. The five existing tests that used to select a deck to open its editor directly now click the new "Open deck editor" button first; the toolbar-toggle test now also covers the new Close button.
- Creator-facing behavior change worth a wiki note: the deck editor's Ctrl/Shift+click multi-selection and the alignment buttons, and that selecting a deck widget no longer opens the deck editor by itself.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3076/pr-test (or any other room on that server)

### DIFF
--- a/client/css/editor/deckeditor.css
+++ b/client/css/editor/deckeditor.css
@@ -181,6 +181,24 @@ body.deckEditorRoomVisible #deckEditorMainCol > .overlay {
   flex-shrink: 0;
 }
 
+/* Align / distribute for a multi-selection of face objects, next to the history controls. Icon-only buttons in
+   one block, disabled (not hidden) while fewer objects than they need are selected. */
+#deckEditorAlignToolbar {
+  display: flex;
+  gap: 2px;
+  flex-shrink: 0;
+}
+
+#deckEditorAlignToolbar button {
+  padding: 3px 6px;
+  margin: 0;
+}
+
+#deckEditorAlignToolbar button:disabled {
+  opacity: 0.4;
+  cursor: default;
+}
+
 #deckEditorHistoryLabel {
   font-size: 11px;
   font-weight: bold;
@@ -1192,6 +1210,65 @@ body.deckEditorPanning #deckEditorMain * {
   font-style: italic;
   opacity: 0.75;
   margin: 6px 5px 10px;
+}
+
+/* Property groups (Content / Position / Size / Colors / Appearance / Custom), the deck editor's version of the
+   Edit Widget sidebar's collapsible blocks - same shape and behaviour, but inside the tab that owns them. */
+.deckEditorGroup {
+  margin: 6px 0;
+  border: 1px solid var(--backgroundHighlightColor1);
+  border-radius: 5px;
+}
+
+.deckEditorGroupHeader {
+  display: flex;
+  align-items: center;
+  gap: 6px;
+  width: 100%;
+  margin: 0;
+  padding: 4px 6px;
+  border: 0;
+  border-radius: 4px;
+  background: var(--backgroundHighlightColor1);
+  color: var(--textColor);
+  text-align: left;
+  font-weight: bold;
+  cursor: pointer;
+}
+
+.deckEditorGroupTitle {
+  flex-shrink: 0;
+}
+
+/* What the block holds, so a folded one still says what is inside it. */
+.deckEditorGroupSummary {
+  flex: 1 1 0;
+  min-width: 0;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  text-align: right;
+  font-weight: normal;
+  font-size: 11px;
+  opacity: 0.75;
+}
+
+.deckEditorGroupArrow::before {
+  content: 'expand_more';
+  font-family: 'Material Symbols';
+  display: inline-block;
+}
+
+.deckEditorGroup.collapsed .deckEditorGroupArrow::before {
+  content: 'chevron_right';
+}
+
+.deckEditorGroupBody {
+  padding: 6px 6px 0;
+}
+
+.deckEditorGroup.collapsed .deckEditorGroupBody {
+  display: none;
 }
 
 /* Shared style for the small notes under section headers (Dynamic properties intro, HTML-object hint). */
--- a/client/css/editor/deckeditor.css
+++ b/client/css/editor/deckeditor.css
@@ -181,24 +181,6 @@ body.deckEditorRoomVisible #deckEditorMainCol > .overlay {
   flex-shrink: 0;
 }
 
-/* Align / distribute for a multi-selection of face objects, next to the history controls. Icon-only buttons in
-   one block, disabled (not hidden) while fewer objects than they need are selected. */
-#deckEditorAlignToolbar {
-  display: flex;
-  gap: 2px;
-  flex-shrink: 0;
-}
-
-#deckEditorAlignToolbar button {
-  padding: 3px 6px;
-  margin: 0;
-}
-
-#deckEditorAlignToolbar button:disabled {
-  opacity: 0.4;
-  cursor: default;
-}
-
 #deckEditorHistoryLabel {
   font-size: 11px;
   font-weight: bold;
@@ -1074,6 +1056,32 @@ body.deckEditorPanning #deckEditorMain * {
   cursor: default;
 }
 
+/* Align / distribute for a multi-selection of face objects: the Object tab's second toolbar row, right where
+   the selection is made. Icon-only buttons, disabled (not hidden) while fewer objects than they need are
+   selected - each in a wrapper that carries the tooltip, since a disabled button gets no pointer events. */
+.deckEditorAlignToolbar {
+  flex-wrap: wrap;
+  align-items: center;
+  gap: 2px;
+}
+
+.deckEditorAlignButton {
+  display: flex;
+}
+
+.deckEditorAlignLabel {
+  font-size: 11px;
+  font-weight: bold;
+  text-transform: uppercase;
+  letter-spacing: 0.3px;
+  opacity: 0.7;
+  margin-right: 4px;
+}
+
+.deckEditorAlignToolbar button {
+  padding: 3px 6px;
+}
+
 /* Holds the shared "Add to: All Cards / Card Type" submenu while the Object tab's "+" has it open. */
 .deckEditorSidebarAddHost:not(:empty) {
   display: block;
@@ -1236,6 +1244,12 @@ body.deckEditorPanning #deckEditorMain * {
   cursor: pointer;
 }
 
+/* Focus ring like the Edit Widget sidebar's collapsible headers (propertyInputs.css), which this mirrors. */
+.deckEditorGroupHeader:focus-visible {
+  outline: 2px solid var(--VTTblue);
+  outline-offset: -2px;
+}
+
 .deckEditorGroupTitle {
   flex-shrink: 0;
 }
@@ -1264,7 +1278,7 @@ body.deckEditorPanning #deckEditorMain * {
 }
 
 .deckEditorGroupBody {
-  padding: 6px 6px 0;
+  padding: 4px 6px;
 }
 
 .deckEditorGroup.collapsed .deckEditorGroupBody {
@@ -1341,6 +1355,25 @@ body.deckEditorPanning #deckEditorMain * {
 
 #deckEditorSidebar .deckEditorProperties .genericInput label {
   flex-shrink: 0;
+}
+
+/* The property name in front of the field. A fixed column so every field starts at the same x - but a
+   game-defined name can be longer than it, so it ends in an ellipsis (and its full text in the tooltip)
+   instead of being cut mid-word right against the field. */
+.deckEditorPropertyLabel {
+  display: inline-block;
+  width: 100px;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+
+/* "(mixed)" stands in for a value the selected objects disagree about, so it must not read as a value someone
+   typed - which the color alone doesn't manage in dark mode, where placeholders are drawn in the same white
+   as real text (layout.css). */
+.deckEditorMixedValue::placeholder {
+  font-style: italic;
+  opacity: 0.6;
 }
 
 /* Property rows: a small red trash to delete the property. The glyph size lives on ::before (button[icon]
--- a/client/css/editor/toolbar.css
+++ b/client/css/editor/toolbar.css
@@ -51,6 +51,11 @@
   display: block;
 }
 
+/* A hover tooltip must never intercept a click meant for whatever is underneath it. */
+#editorToolbar button span, #editorDragToolbar button span {
+  pointer-events: none;
+}
+
 #editorToolbar .divider, #editorDragToolbar .divider {
   border-left: 1px solid #fff8;
   margin: 5px;
--- a/client/editor.html
+++ b/client/editor.html
@@ -290,6 +290,7 @@
           <div id="deckEditorBreadcrumb" title="Your edits this session — click a step to jump back or forward to it."></div>
           <span class="deckEditorHistoryDivider"></span>
           <button icon="crop_portrait" id="deckEditorCardView" class="active">Card view</button>
+          <button icon="close" id="deckEditorClose" title="Close the deck editor.">Close</button>
         </div>
         <div id="deckEditorContent">
           <div id="deckEditorMainCol">
--- a/client/editor.html
+++ b/client/editor.html
@@ -289,6 +289,17 @@
           <span id="deckEditorHistoryLabel">History:</span>
           <div id="deckEditorBreadcrumb" title="Your edits this session — click a step to jump back or forward to it."></div>
           <span class="deckEditorHistoryDivider"></span>
+          <div id="deckEditorAlignToolbar" title="Ctrl+click (or Shift+click) face objects to select more than one, then line them up here.">
+            <button id="deckEditorAlignLeft" icon="align_horizontal_left" title="Align the selected face objects to the left"></button>
+            <button id="deckEditorAlignCenter" icon="align_horizontal_center" title="Align the selected face objects to the center"></button>
+            <button id="deckEditorAlignRight" icon="align_horizontal_right" title="Align the selected face objects to the right"></button>
+            <button id="deckEditorAlignTop" icon="align_vertical_top" title="Align the selected face objects to the top"></button>
+            <button id="deckEditorAlignMiddle" icon="align_vertical_center" title="Align the selected face objects to the middle"></button>
+            <button id="deckEditorAlignBottom" icon="align_vertical_bottom" title="Align the selected face objects to the bottom"></button>
+            <button id="deckEditorDistributeH" class="deckEditorDistributeButton" icon="horizontal_distribute" title="Equalize the horizontal spacing between the selected face objects"></button>
+            <button id="deckEditorDistributeV" class="deckEditorDistributeButton" icon="vertical_distribute" title="Equalize the vertical spacing between the selected face objects"></button>
+          </div>
+          <span class="deckEditorHistoryDivider"></span>
           <button icon="crop_portrait" id="deckEditorCardView" class="active">Card view</button>
         </div>
         <div id="deckEditorContent">
--- a/client/editor.html
+++ b/client/editor.html
@@ -289,17 +289,6 @@
           <span id="deckEditorHistoryLabel">History:</span>
           <div id="deckEditorBreadcrumb" title="Your edits this session — click a step to jump back or forward to it."></div>
           <span class="deckEditorHistoryDivider"></span>
-          <div id="deckEditorAlignToolbar" title="Ctrl+click (or Shift+click) face objects to select more than one, then line them up here.">
-            <button id="deckEditorAlignLeft" icon="align_horizontal_left" title="Align the selected face objects to the left"></button>
-            <button id="deckEditorAlignCenter" icon="align_horizontal_center" title="Align the selected face objects to the center"></button>
-            <button id="deckEditorAlignRight" icon="align_horizontal_right" title="Align the selected face objects to the right"></button>
-            <button id="deckEditorAlignTop" icon="align_vertical_top" title="Align the selected face objects to the top"></button>
-            <button id="deckEditorAlignMiddle" icon="align_vertical_center" title="Align the selected face objects to the middle"></button>
-            <button id="deckEditorAlignBottom" icon="align_vertical_bottom" title="Align the selected face objects to the bottom"></button>
-            <button id="deckEditorDistributeH" class="deckEditorDistributeButton" icon="horizontal_distribute" title="Equalize the horizontal spacing between the selected face objects"></button>
-            <button id="deckEditorDistributeV" class="deckEditorDistributeButton" icon="vertical_distribute" title="Equalize the vertical spacing between the selected face objects"></button>
-          </div>
-          <span class="deckEditorHistoryDivider"></span>
           <button icon="crop_portrait" id="deckEditorCardView" class="active">Card view</button>
         </div>
         <div id="deckEditorContent">
--- a/client/js/editor/deckeditor.js
+++ b/client/js/editor/deckeditor.js
@@ -1698,10 +1698,12 @@ class DeckEditor {
   }
 
   async alignObjects(property, lowerBound, upperBound, factor) {
+    // Flush first, measure afterwards: besides not absorbing a pending typed edit into this action, flushing
+    // can run a queued field edit that moves an object - the rects below have to describe the result of that.
+    await this.flushPendingCommits();
     const items = this.alignItems();
     if(items.length < 2)
       return;
-    await this.flushPendingCommits(); // don't absorb a pending typed edit into this action
     const lower = Math.min(...items.map(item=>item.rect[lowerBound]));
     const target = lower + (Math.max(...items.map(item=>item.rect[upperBound])) - lower) * factor;
     for(const { object, rect } of items) {
@@ -1712,10 +1714,10 @@ class DeckEditor {
   }
 
   async distributeObjects(property, lowerBound, upperBound) {
+    await this.flushPendingCommits(); // flush before measuring, see alignObjects
     const items = this.alignItems();
     if(items.length < 3)
       return;
-    await this.flushPendingCommits(); // don't absorb a pending typed edit into this action
     const sorted = [...items].sort((a,b)=>a.rect[lowerBound] - b.rect[lowerBound]);
     const start = sorted[0].rect[lowerBound];
     const span = Math.max(...items.map(item=>item.rect[upperBound])) - start;
@@ -1820,7 +1822,7 @@ class DeckEditor {
       // A disabled button receives no pointer events, so its own tooltip never opens - which is exactly the
       // state in which it needs to explain itself. The tooltip therefore sits on a wrapper around it instead.
       const wrapper = div(bar, 'deckEditorAlignButton');
-      wrapper.title = `${title}. Needs ${minSelected} or more selected: ctrl+click objects on the card or in the list, or ctrl+A to take the whole face.`;
+      wrapper.title = `${title}. Needs ${minSelected} or more visible objects selected: ctrl+click objects on the card or in the list, or ctrl+A to take the whole face.`;
       const button = document.createElement('button');
       button.id = id;
       button.setAttribute('icon', icon);
@@ -1832,10 +1834,16 @@ class DeckEditor {
     this.updateAlignToolbar();
   }
 
+  // Enabled from what align/distribute can actually act on, not from the raw selection: alignItems() leaves out
+  // objects without a visible box ("display: false"), so counting the selection would leave a button clickable
+  // that then does nothing - e.g. one visible object plus a hidden one.
   updateAlignToolbar() {
-    const selected = this._selectedObjects.length;
-    for(const button of $a('.deckEditorAlignToolbar button'))
-      button.disabled = selected < +button.dataset.minSelected;
+    const buttons = $a('.deckEditorAlignToolbar button');
+    if(!buttons.length)
+      return;
+    const usable = this.alignItems().length;
+    for(const button of buttons)
+      button.disabled = usable < +button.dataset.minSelected;
   }
 
   updateDragToolbar() {
@@ -2754,12 +2762,21 @@ class DeckEditor {
     input.disabled = !editable;
     input.title = bound ? `Text for card type "${this.cardType}" (property "${bound}")` : 'Text on every card';
     // Clicking/dragging inside the field must not start a row drag, but should still select this field's
-    // face object (and switch to its face) if it isn't already the selection.
-    input.onmousedown = e=>e.stopPropagation();
+    // face object (and switch to its face) if it isn't already the selection. The browser focuses the field on
+    // mousedown, so whether the user was already typing in it has to be remembered from before that.
+    let wasFocused = false;
+    input.onmousedown = e=>{
+      wasFocused = document.activeElement == input;
+      e.stopPropagation();
+    };
     input.onclick = e=>{
       e.stopPropagation();
       // The field covers most of its row, so it has to offer the row's Ctrl/Shift multi-select too - otherwise
-      // text objects could only ever be selected one at a time.
+      // text objects could only ever be selected one at a time. Inside the field the user is currently typing
+      // in, though, those clicks belong to the text: shift+click extends the caret selection there, and taking
+      // it over would rebuild the tree (afterSelectionChanged) and destroy the field mid-edit.
+      if(wasFocused && (e.shiftKey || e.ctrlKey || e.metaKey))
+        return;
       if(e.shiftKey)
         return this.extendObjectSelection(index, faceIndex, e.ctrlKey || e.metaKey);
       if(e.ctrlKey || e.metaKey)
@@ -2793,8 +2810,10 @@ class DeckEditor {
     box.append(input);
   }
 
-  // Copies every selected object, appending the copies after the last one so a copied group stays a group -
-  // and selects the copies, so the next drag moves them off their originals.
+  // Copies every selected object, putting the copies together right after the last one - so they stay a block
+  // that the next drag (they are the new selection) moves off the originals as a whole. A non-contiguous
+  // selection therefore comes back contiguous: the copies are drawn next to each other rather than each one
+  // right on top of its original.
   async copySelectedObject() {
     const face = this.faceTemplates[this.face];
     const indices = this.selectedObjectIndices();
@@ -2820,10 +2839,17 @@ class DeckEditor {
     if(!face || !Array.isArray(face.objects) || from === to || to < 0 || to >= face.objects.length)
       return;
     await this.flushPendingCommits(); // don't absorb a pending typed edit into this action
-    const wasSelected = this.selectedObject === from;
     const [ object ] = face.objects.splice(from, 1);
     face.objects.splice(to, 0, object);
-    this.selectedObject = wasSelected ? to : null; // keep the dragged object selected if it was
+    // The whole selection follows the splice, so reordering one row doesn't drop objects that were picked up
+    // with ctrl+click: the dragged object lands on "to", everything it moved past shifts by one, the rest stays.
+    this._selectedObjects = this._selectedObjects.map(i=>{
+      if(i === from)
+        return to;
+      if(from < to)
+        return i > from && i <= to ? i-1 : i;
+      return i >= to && i < from ? i+1 : i;
+    });
     this.refreshMainCardFaces();
     await this.commit('faceTemplates', `${getPlayerDetails().playerName} reordered a face object of deck ${this.deckID} in deck editor`);
     this.render();
@@ -3034,7 +3060,9 @@ class DeckEditor {
       input.oninput = input.onchange = _=>{
         if(fieldType == 'number') {
           if(input.value.trim() === '') {
-            if(emptyIsZero) // face object properties: an erased number reads as 0
+            // Face object properties: an erased number reads as 0 - except in a "(mixed)" field, where empty is
+            // the state it started in, so clearing it again is "never mind" and must not write 0 to everything.
+            if(emptyIsZero && !mixed)
               onValueChanged(0);
             return; // otherwise a momentarily-empty field shouldn't commit 0 over the real value
           }
--- a/client/js/editor/deckeditor.js
+++ b/client/js/editor/deckeditor.js
@@ -1521,10 +1521,10 @@ class DeckEditor {
         e.preventDefault();
         // Ctrl/Cmd or Shift picks up a second object instead of replacing the selection - and only changes the
         // selection, so a modifier click can't drag the objects around by accident.
+        if(e.shiftKey)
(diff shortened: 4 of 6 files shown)


VTT-AI-Agent (commented):
## TL;DR

Nice, well-structured feature — the selection model (`_selectedObjects` with a `selectedObject` accessor collapsing to one) keeps every existing call site working, and the align/distribute math faithfully mirrors `client/js/editor/toolbar/align.js`. The main thing to fix: **a `(mixed)` row loses its field type** — because `commonPropertyValue()` returns `value: undefined` when the objects disagree, `addTypedInput()` falls through to a plain *text* input for any property without a forced type, so editing a mixed `opacity`/`display`/custom property writes a **string** (`"0.5"`, `"false"`) into every selected object.

## Findings

### 1. `(mixed)` rows degrade to text inputs and write the wrong JS type — `client/js/editor/deckeditor.js:2905`

`objectFieldType()` (line 2893) only forces a type for a fixed list (`x, y, width, height, fontSize, size, strokeWidth, rotation` → number; `textAlign, color, value, strokeColor, type` → text). Everything else — `opacity`, `hoverOpacity`, `flip`, `display`, `css`, `svgReplaces` and every game-defined custom property — leaves `fieldType` undefined, so `addTypedInput` infers it from the value it was handed (2906-2911). For a multi-selection that value is `common.value`, which `commonPropertyValue()` deliberately sets to `undefined` whenever the objects disagree (line 2864-ish, `return { value: mixed ? undefined : first, mixed }`). `typeof undefined` is neither `number`, `boolean` nor `object`, so **every mixed row becomes `fieldType == 'text'`**.

Concrete failures:
- Two objects with `opacity: 0.5` and `opacity: 1` → text field. Typing `0.75` runs the `else` branch at 2946 and writes the string `"0.75"` to both, which is what gets saved into the game's `faceTemplates` JSON.
- Two objects with `display: true` / `display: false` → text field instead of a checkbox. Typing `false` writes the **string** `"false"`, which is truthy, so `card.js:131` (`!object.display`) keeps the object visible — the opposite of what the user asked for.
- Two objects with different `css`/`svgReplaces` objects → text field; typing anything replaces the object with a string.

Side effect: the `input.indeterminate = !!mixed` (2922) and `mixed ? ''` textarea (2926) branches you added are unreachable from the object tab for exactly this reason.

Suggested fix: have `commonPropertyValue()` also report a type hint taken from the first *defined* value (`objects.find(o => o[property] !== undefined)?.[property]`) and pass it to `addTypedInput` as the type source, keeping `value` blank for display only.

### 2. Hidden face objects get their coordinates wrecked by align/distribute — `client/js/editor/deckeditor.js:1681`

`alignItems()` only filters on `divs[i]` being present, not on the rect being meaningful. A face object with `display: false` renders as `display: none` (`client/js/widgets/card.js:131-132`), and `getBoundingClientRect()` on a `display:none` element returns an all-zero rect. So for such an object `own` is `0` while `target` is a real viewport coordinate (hundreds of px), and line 1697 computes `object.x = round(x - (0 - target)/cardScale)` — the object is teleported by roughly the card's viewport offset. Ctrl+A (`selectAllObjects`, 1629) selects hidden objects too, so "select all → align left" silently corrupts every hidden object on the face. `distributeObjects` has the same problem (the zero-rect sorts first and becomes the anchor).

Filtering out zero-area rects in `alignItems()` (and using the filtered count for the cause string) would fix both.

### 3. `width`/`height` on the All Cards tab now open folded — `client/js/editor/deckeditor.js:2941`

`propertyGroups()` marks `size` (and `position`) `collapsed: true`, and `renderPropertyGroups` is reused verbatim for the card-defaults tab (line ~3166). But the card-defaults code explicitly always offers `width` and `height` because "the most common defaults are always offered" — those two are essentially the *point* of that tab, and they are not "usually changed by dragging" the way a face object's x/y are. After this PR the All Cards tab opens with its two primary fields hidden inside a folded **Size** block, with nothing else in it. Consider making the default fold state per-`stateKey` (only fold Position/Size for `stateKey == 'object'`).

Similar, smaller: a game whose card type defines a property named `x`, `size` or `scale` will have it folded away by default on the Card Type tab, since the group lists are matched purely by name against arbitrary game-defined properties.

### 4. A plain click on the card can no longer reduce a multi-selection — `client/js/editor/deckeditor.js:1534`

`if(!this.isObjectSelected(index)) this.selectObject(index);` means clicking an object that is *already* part of the selection leaves the whole group selected (necessary so the click can start a group drag). But the tree row's plain click (`renderTreeObjectRow`, 2478) *does* collapse to one, so the two surfaces disagree, and the PR description's "A plain click still selects exactly one, as before" isn't true on the card. The risk isn't cosmetic: a user who clicks one object of a group to work on it alone, then types into `value`, writes to all of them. The usual editor behaviour is to collapse the selection on mouse*up* when no drag happened — you already track `moved` in `startObjectDrag`, so collapsing in `up()` when `!moved && !modifier` would be a small addition.

### 5. Stale `deckEditorClippedObject` after the primary changes — `client/js/editor/deckeditor.js:1752`

`updateSelectionOutline()` toggles `deckEditorClippedObject` on the *primary* object only, and that class kills the object's own outline (`deckeditor.css:947`). Ctrl+clicking a second object makes it the new primary; the old primary keeps the class (a pure selection change goes through `afterSelectionChanged` → `attachObjectHandlers`, which only toggles `deckEditorSelectedObject`, not a full `refreshMainCardFaces`). Result: a large/edge-touching object that is still selected shows no outline at all. More generally, only the primary ever gets the stand-in rectangle, so clipped non-primary members of a selection are invisible as selected.

### Smaller notes

- `finishAlignment(count, 'evenly spaced')` produces `"<player> evenly spaced 3 face objects of deck …"`, which the breadcrumb fallback renders as "Evenly spaced 3 face objects" — fine — but `objectActionCause` is verb-shaped everywhere else; "distributed" would read more consistently.
- `alignItems()` pairs `selectedObjectTemplates()` (which `.filter(object=>object)`) with `selectedObjectDivs()` (which does not filter). They're 1:1 today because `reload()` prunes stale indices and `card.js` creates one div per object, but the two helpers' contracts differ — worth aligning so a future divergence can't silently pair object *i* with div *j*.
- `commonPropertyValue` treats "absent" and "explicitly `null`" as equal (both stringify to `null`), so a property that one object has as `null` and another lacks entirely shows as agreeing. Harmless, but it means the "a property only some of them have still gets a row" story has a hole for `null` values.
- Ctrl+A is now swallowed whenever the deck editor is open and focus isn't in a field — including when the user is reading the export overlay. Minor, but it's a global shortcut takeover; gating it on `sidebarTab`/`activeArea` would be more conservative.
- Test coverage is good for the happy path. Worth adding a case for the mixed-*type* row (finding 1) and for align with a `display: false` object (finding 2), since those are the two paths that can write bad data.

I added the `beta` label: the PR touches no engine file (only `client/css/editor/deckeditor.css`, `client/editor.html`, `client/js/editor/deckeditor.js` and the TestCafe suite), it changes no saved-game semantics, and the issues above are undoable edge cases inside the deck editor rather than anything that could crash the server or corrupt a game, so it's safe and useful for players to try.

VTT-AI-Agent:
Thanks — all five findings were real; fixed in 1314183e.

**1. `(mixed)` rows degraded to text inputs.** `commonPropertyValue()` now also returns `sample`, the value of the first object that actually has the property set, and the object row takes its field type from that (`this.objectFieldType(property) || this.valueFieldType(common.sample)`). The type-inference block of `addTypedInput` moved into that new `valueFieldType()` helper, so both paths agree. A mixed `display` row is a checkbox again (indeterminate, writing a real boolean), a mixed `css` row a JSON textarea — which also makes the `indeterminate` / empty-textarea branches reachable, as you noted.

**2. Align/distribute wrecked hidden objects.** `alignItems()` was rewritten: it pairs each object with its own rendered node **by index** (no more zipping two separately-filtered lists — that also covers your note about `selectedObjectTemplates()`/`selectedObjectDivs()` having different contracts; `selectedObjectDivs()` now drops the same indices) and skips items whose rect has no width *and* no height, so a `display:none` object is left where it is. `items.length` is the filtered count, so the "needs 2/3 objects" guards and the cause string follow.

**3. `width`/`height` folded on All Cards.** The default fold state is now per tab: `Position`/`Size` only start collapsed for `stateKey == 'object'`. All Cards opens with width/height visible again, and a game-defined `x`/`size`/`scale` on the Card Type tab is no longer hidden by name. Screenshot of the All Cards tab after the fix: https://agent.virtualtabletop.io/reports/pr/1532415754513682593/allcards.png

**4. A plain click could not reduce a multi-selection.** Done the way you suggested: mouse-down still keeps the group (so the click can start a group drag), and `up()` collapses to the clicked object when `!moved` and the selection had more than one member. Card and tree now behave the same.

**5. Stale `deckEditorClippedObject`.** `updateSelectionOutline()` strips the class from every object that is not the current primary before doing its work, so changing the primary with Ctrl+click can't leave an object with its outline switched off and no rectangle standing in for it.

Smaller notes: the distribute cause is `distributed` now, and Ctrl+A is ignored while an overlay is open (`!isOverlayActive()`), so "select all" in the export overlay is the browser's again. I left `commonPropertyValue`'s "absent vs. explicit `null`" equality alone — distinguishing them would make a row claim a disagreement the user cannot see or resolve, and both edit to the same thing.

Tests: the multi-selection TestCafe test grew a fourth, `display: false` object and now checks that Ctrl+A selects it, that aligning leaves its `x` at 200, that the `display` row is a `(mixed)` **checkbox**, and that ticking it writes `true` (boolean) to all four (state hash updated). `npm test` and the full `tests/testcafe/editor.js` suite pass locally (17/17).

VTT-AI-Agent (commented):
## TL;DR

UI/UX review only (not a code review) — I ran the branch locally on a real browser and clicked through the deck editor. The grouped sidebar is a clear win: the Object tab goes from a flat wall of rows to Content / Position / Size / Colors / Appearance / Custom, and the multi-select header + `(mixed)` fields read well. The one thing I'd change before merge: **the 8 new align/distribute buttons sit in the deck editor's top bar, 300px away from the selection they act on — and they break that bar's layout at ~1024px wide** (the History breadcrumb collapses to nothing and "Card view" is clipped to "CA"). Moving them into the Object tab's own toolbar in the sidebar fixes both the layout regression and their discoverability.

Rendered at 1400×900 (Chromium), deck with four text objects, three of them Ctrl-selected:

multi-selection, grouped sidebar

---

## Big layout improvements

**1. The align toolbar breaks the top bar at common window widths — and belongs next to the selection.**
`client/editor.html:292` adds a `flex-shrink: 0`, ~302px-wide block into `#deckEditorHistory`, a row whose only flexible member is `#deckEditorBreadcrumb`. At 1024×700 the breadcrumb is squeezed to zero (the "History:" label is left pointing at nothing) and the `Card view` button overflows the bar and is clipped mid-word:

| this PR @1024px | `main` @1024px |
|---|---|
| PR at 1024 | main at 1024 |

Beyond the clipping, the placement is odd on its own terms: the buttons act on the face-object selection, which lives in the left tree and the right sidebar, while the top bar is otherwise about the deck as a whole (add deck / undo / redo / history / card view). Concretely I'd **move the eight buttons into the Object tab's existing sidebar toolbar** (the `+ / copy / delete` `menu` rendered by `renderSidebarToolbar`, `client/js/editor/deckeditor.js:2140`), as a second row or a divider-separated group. That puts them directly under "3 face objects selected", right where the user just made the selection, and leaves the top bar untouched. If they must stay in the top bar, at minimum give the block `flex-shrink: 1; min-width: 0; overflow: hidden` (or `flex-wrap` on `#deckEditorHistory`) so the bar degrades instead of clipping, and hide them entirely while the sidebar isn't on the Object tab.

**2. Eight icon-only buttons that are disabled most of the time show no explanation on hover.**
The "how do I enable these?" hint lives in the *container's* `title` (`client/editor.html:292`), but browsers don't dispatch pointer events on `disabled` buttons, so while the toolbar is in its default (disabled) state neither the button's own tooltip nor the container's appears — a first-time user gets eight grey glyphs with no way to find out what they are. Two cheap fixes that fit the existing style: give the block a small caps label the way `#deckEditorHistoryLabel` does (`ALIGN:`), and/or wrap each button in a `<span title=…>` so the tooltip survives the disabled state. The container tooltip is also the only place multi-select is documented — it should mention `Ctrl+A` too, since that shortcut is otherwise invisible in the UI.

**3. The engine's group names don't fit the Card Type tab.**
`propertyGroups()` (`client/js/editor/deckeditor.js:2877`) is a fixed name list, which is exactly right for a face object (`x`, `color`, `fontSize` really are engine properties). On the **Card Type** tab, though, *every* property is game-defined, so the split is decided by coincidence of naming: a card type property called `name` lands in **Content** while `cost` lands in **Custom**, which suggests a distinction that doesn't exist.

card type tab

I'd either keep the Card Type tab flat (its lists are usually short anyway) or put all of its rows under a single neutral block. Related: when everything lands in **one** group, the block is pure chrome — a header whose summary just repeats the single row below it. Worth skipping the group wrapper when only one group would be rendered, or when the tab has ≤2 rows.

- **Property labels are clipped with no ellipsis and no tooltip.** The label is a hard `width:100px` (`client/js/editor/deckeditor.js:2833` and `:2958`); measured in the browser, `someCustomThing` has `scrollWidth: 135` and renders as `someCustomT` butted straight against the input — see the sidebar crop below. Since the Custom block is precisely where long game-authored names show up, this is more visible after this PR. Suggest `overflow:hidden; text-overflow:ellipsis; white-space:nowrap` plus `labelEl.title = property` so the full name is at least hoverable.
- **`(mixed)` is invisible as a placeholder in dark mode.** `client/css/layout.css:44` sets `body.edit.darkMode input::placeholder { color: #ffffff }`, i.e. the same colour as a real value — in dark mode `(mixed)` reads like the literal string a user typed. Compare `fontSize: 14` and `color: (mixed)` in the dark screenshot; they look equally "filled in". A `.deckEditorMixedValue` class with `font-style: italic; opacity: .6` on the input would make the distinction survive both themes.
- **The mixed boolean row says nothing.** `input.indeterminate` (`deckeditor.js:2991`-ish) gives `myGameFlag` a dash-in-a-box, but unlike every other row it never says `(mixed)`, and it isn't obvious that the first click hands `true` to *all* selected objects. A `title="Mixed — clicking sets this value on all selected objects"` on the checkbox would cover it.
- **The group header has no focus style**, unlike its counterpart in the Edit Widget sidebar (`client/css/editor/propertyInputs.css:497` has `.collapsibleHeader:focus-visible { outline: 2px solid var(--VTTblue) }`). Since `.deckEditorGroupHeader` (`client/css/editor/deckeditor.css:1223`) is a real `<button>`, it *is* tab-focusable — it just shows nothing when focused. Copying that one rule keeps the two sidebars consistent.
- **Summary styling drifts from the sidebar it's mirroring**: the widget sidebar's `.collapsibleSummary` is left-aligned at normal size/colour, while `.deckEditorGroupSummary` (`deckeditor.css:1244`) is right-aligned, 11px, `opacity: .75`. The right-aligned version actually reads nicely here, but if the intent is "same shape and behaviour" (per the CSS comment), it'd be worth picking one.
- `.deckEditorGroupBody` uses `padding: 6px 6px 0` where the widget-sidebar equivalent uses `4px 6px`; the zero bottom padding leaves the last row sitting slightly tighter against the block border than the first row does against the header.
- Disabled buttons at `opacity: 0.4` over the grey bar are quite faint against the enabled ones — fine for "unavailable", just note it's below the contrast the rest of the top bar uses.

sidebar, light
sidebar, dark

- **The Dynamic properties section silently disappears** when a second object is selected (`deckeditor.js:2092`). To someone who just Ctrl+clicked, a whole amber panel vanishing looks like a bug. The panel already owns a `deckEditorSectionNote` idiom — keep the header and replace the body with *"Dynamic properties are per object. Select a single object to edit them."*
- **"3 face objects selected" doesn't say *which* three.** The single-object header names its object ("Face object 1 (text)"), so the multi header losing all identity is a small step down; `3 face objects selected (1, 2, 3)` (or the numbers as the summary line) would let the user confirm the selection without cross-checking the tree.
- **The word "(mixed)" is doing a lot of work** for a newcomer. The intro note explains it, but the note only appears once, above blocks that may be collapsed — and "mixed" isn't a term the rest of the app uses. `(varies)` or `(different values)` would need no explanation.
- **Multi-select is discoverable only via tooltips.** The tree row title now ends with "— Ctrl+click to select several at once", which is good, but a user who never hovers a row (or is on touch) will never find it, and `Ctrl+A` is documented nowhere in the UI. The Object tab's "No face object is selected…" note is a natural place for one extra sentence.
- **Position/Size collapsed by default** is a sensible call, but it does mean a first-time user sees no `x`/`y` at all on the Object tab. The `x, y` summary carries it — just flagging that the *only* numeric way to place an object is now one click deeper.
- Minor wording: the Object tab tooltip uses a hyphen in "…of face 0 - every property here is set on all of them"; the rest of the deck editor's prose uses an em dash (e.g. the breadcrumb and tree tooltips).

---

Verified by running the branch on a local server and driving Chromium; screenshots above are from that session. Nothing here blocks the feature — the grouping and the multi-select model both make the editor easier to read; the top-bar overflow is the one item I'd fix before merge.

VTT-AI-Agent:
Thanks — the top-bar overflow was real and is fixed in c06b4d90, along with most of the smaller items.

**1. Align toolbar broke the top bar.** Moved, as suggested: the eight buttons are now a second row of the Object tab's sidebar toolbar (`renderAlignToolbar`), directly above the "3 face objects selected" header, and `#deckEditorHistory` is byte-for-byte what it is on `main` again. They only exist while the Object tab is open, so no other tab carries them either. At 1024×700 the breadcrumb and "Card view" are intact:

1024px, buttons in the sidebar

**2. Disabled buttons showed no explanation.** Each button now sits in a wrapper element that carries the tooltip, so it opens in the disabled state too, and the text says both what the button does and how to get there: *"…Needs 2 or more selected: ctrl+click objects on the card or in the list, or ctrl+A to take the whole face."* The row has an `ALIGN:` caps label in the `#deckEditorHistoryLabel` style. Ctrl+A is now also written out in the Object tab's "No face object is selected…" note and in the multi-selection note, so it isn't tooltip-only.

**3. Group names don't fit the Card Type tab.** Agreed — every card type property is game-defined, so the split was decided by coincidence. That tab is flat again, exactly as before this PR. And `renderPropertyGroups` now skips the wrapper entirely when only one group would be rendered (a header repeating the single row below it is pure chrome), which also un-groups the All Cards tab's width/height.

**Nitpicks**
- Labels: `.deckEditorPropertyLabel` replaces the inline `width:100px` and adds `overflow:hidden; text-overflow:ellipsis; white-space:nowrap`, plus `title` with the full name — `someCustomThing` reads as `someCusto…` with the name on hover.
- `(mixed)`: mixed inputs get `.deckEditorMixedValue`, whose `::placeholder` is italic at `opacity: .6`, so it stays distinguishable in dark mode where placeholders are drawn in the same white as a value.
- Mixed checkbox: `title="Mixed — the selected objects disagree about this. Clicking sets one value on all of them."`
- Group header: got `:focus-visible { outline: 2px solid var(--VTTblue); outline-offset: -2px }`, copied from `.collapsibleHeader`.
- `.deckEditorGroupBody` padding is `4px 6px` now, matching the widget sidebar.
- Object tab tooltip uses an em dash.

**Readability**
- Dynamic properties no longer vanishes: with several objects selected it keeps its amber header and shows *"Dynamic properties are per object. Select a single object to edit them."*
- The multi header names the selection: `3 face objects selected (1, 2, 3)`, truncated after eight numbers.

**Not changed, with reasons**
- **"(mixed)" wording.** Kept. It's the term design tools use for exactly this state, the note above the blocks defines it, and `(varies)`/`(different values)` would be new vocabulary for the same idea — the italic placeholder plus the note carries more of the meaning than the word choice does.
- **Summary alignment drifting from `.collapsibleSummary`.** Kept right-aligned/11px: this summary holds arbitrary game-defined property names, so it has to truncate against the block's right edge rather than push the title around. Same shape and behaviour, deliberately different content.
- **Position/Size collapsed on the Object tab.** Left as is — you flagged it rather than asked for a change, and the `x, y` summary keeps them one click away.
- **Disabled buttons at `opacity: 0.4`.** They now inherit `.deckEditorSidebarToolbar button:disabled`, the same 0.4 the tab's add/copy/delete buttons already use, so they are consistent with their neighbours instead of with the top bar.

Tests: the multi-selection TestCafe case drives the buttons in their new home (its expectations needed only the new header text); `npm test` and the full `tests/testcafe/editor.js` pass locally (17/17). Sidebar after the fixes: [light](https://agent.virtualtabletop.io/reports/pr/1532415754513682593/pr3076-fixed-sidebar.png) · [dark](https://agent.virtualtabletop.io/reports/pr/1532415754513682593/pr3076-dark-sidebar.png) · [Card Type tab](https://agent.virtualtabletop.io/reports/pr/1532415754513682593/pr3076-cardtype.png) · [Dynamic properties note](https://agent.virtualtabletop.io/reports/pr/1532415754513682593/pr3076-dynamic-note.png)

VTT-AI-Agent (commented):
## TL;DR

PR #3076 addresses the prior review findings, and the implementation remains confined to deck-editor UI code. One selection edge case remains: Shift-click retains earlier Ctrl-selected objects outside the requested range, which can make subsequent bulk edits affect unintended objects.

## Findings

### 1. Shift-click does not replace the selection with the advertised range — `client/js/editor/deckeditor.js:1619`

`extendObjectSelection()` keeps every existing selection index that is outside the newly computed range:

```js
this._selectedObjects = [ ...this._selectedObjects.filter(i=>range.indexOf(i) == -1), ...range.filter(i=>i != index), index ];
```

For example, Ctrl-click objects 1 and 4, then Shift-click object 2. The current/primary object is 4, so the described result is the range 2–4, but this expression also retains object 1 and leaves all four selected. Because property edits, dragging, copy, and delete operate on the entire selection, the extra retained object can be modified unexpectedly. Please replace the selection with the computed range (while keeping the clicked object last if it remains the primary), or explicitly require Ctrl+Shift for additive range selection, and add a TestCafe assertion for a range formed after a discontiguous Ctrl-selection.

The `beta` label was already present; I left it in place because the changed files are limited to deck-editor CSS/JS and TestCafe coverage, touch no engine path, and this UI selection edge case is safe and useful to test.

VTT-AI-Agent:
Fixed in c17d51a9 — the finding was right.

**Shift+click retained ctrl-selected objects outside the range.** `extendObjectSelection()` now makes the range the whole selection, and the additive variant is behind ctrl+shift, as suggested:

```js
const kept = additive ? this._selectedObjects.filter(i=>range.indexOf(i) == -1) : [];
this._selectedObjects = [ ...kept, ...range.filter(i=>i != index), index ]; // clicked object last = primary
```

Your example now behaves as advertised: ctrl+click 1 and 4, shift+click 2 → objects 2–4 selected, object 1 dropped, so the following property edit / drag / copy / delete can't reach it. The clicked object stays last (primary), which is what the sidebar header and the drag toolbar read.

All three call sites (object on the card, tree row, the text field covering a tree row) had to test `shiftKey` *before* `ctrlKey`, since ctrl+shift previously fell into the toggle branch.

**Test.** The multi-selection TestCafe case got the range assertion for a discontiguous ctrl-selection you asked for — select object 1, ctrl+click 4, shift+click 2 → 3 rows selected and row 1 no longer among them; then ctrl+shift+click 1 → 4 selected. On the old code the first assertion fails with 4. `npm test` and the full `tests/testcafe/editor.js` pass locally (17/17, state hashes unchanged).

The Object tab's multi-selection note now also mentions shift+click, so the range shortcut isn't only in the code.

VTT-AI-Agent (commented):
## TL;DR

The three findings from the earlier rounds are genuinely fixed (mixed rows now keep their field type via `common.sample`, `alignItems()` drops zero-area rects, shift+click replaces the range, align moved into the Object tab). What's left is small: two align/distribute buttons can be **enabled and do nothing** when part of the selection is hidden, and **shift+click inside a tree row's inline text field** no longer extends the text selection — it re-selects objects and rebuilds the tree under the cursor.

## Findings

### 1. Align/distribute buttons can be enabled but silently do nothing — `client/js/editor/deckeditor.js:1835`

`updateAlignToolbar()` enables a button from `this._selectedObjects.length`, but `alignObjects()`/`distributeObjects()` act on `alignItems()`, which (correctly, per the earlier review) drops objects whose rendered rect is all zeros — i.e. `display: false` objects (`client/js/widgets/card.js:131`).

Concretely, with the PR's own test fixture: select object 4 (`display: false`) plus one visible object → `_selectedObjects.length == 2` → **Align left is enabled**, but `items.length == 1`, so `alignObjects()` returns at line 1703 and nothing happens, with no feedback. Same for distribute with three selected of which one is hidden. Ctrl+A on a face with hidden objects makes this easy to hit.

Cheapest fix: drive the enabled state off `this.alignItems().length` instead of the raw selection count (it's already only recomputed on `updateDragToolbar`/render, and the card is rendered by then), and keep the count in the wrapper's tooltip so a disabled button still explains itself.

### 2. Shift+click inside a row's inline text field steals the native text selection — `client/js/editor/deckeditor.js:2759`

`renderPreviewTextField()`'s `onclick` now runs the row's modifier handling before the "select this object" fallback:

```js
if(e.shiftKey)
  return this.extendObjectSelection(index, faceIndex, e.ctrlKey || e.metaKey);
```

That field is a real editable text input for the object's `value`. Shift+click in a focused text input is the standard "extend the caret selection to here" gesture, and ctrl/cmd+click is at least harmless there today. After this PR, shift+clicking while editing the text of object 3 changes the *object* selection and then `afterSelectionChanged()` → `renderLeftSidebar()` rebuilds the tree, so the input the user was typing in is destroyed and focus is lost mid-edit.

Suggestion: only take over the modifier click when the field isn't the focused element (`document.activeElement != input`), or leave the field alone entirely and let the modifier gestures live on the row around it — the row already has ~40% of its width outside the input, plus the number badge and the type icon.

### 3. `alignObjects`/`distributeObjects` measure before flushing pending edits — `client/js/editor/deckeditor.js:1700`

```js
const items = this.alignItems();          // rects measured here
if(items.length < 2) return;
await this.flushPendingCommits();          // may run a queued field edit that moves objects + re-renders
```

`flushPendingCommits()` first awaits `this.fieldEditChain`, and a queued `queueFieldEdit` callback mutates `x`/`y` and calls `refreshMainCardFaces()`. If one is still queued when the button is clicked, the alignment is computed from rects that no longer describe the objects. The window is narrow (the chain is microtask-driven, so it has normally already run), but moving `const items = …` below the `await` costs nothing and removes the ordering question — the drag buttons already flush *before* they read anything (`DeckEditorDragButton.pointerDown`, line 34).

### 4. Reordering a row drops a multi-selection — `client/js/editor/deckeditor.js:2823`

`moveFaceObject()` still compares `this.selectedObject === from`, i.e. only the **primary**. With three objects selected, dragging any non-primary row in the tree to reorder it sets `this.selectedObject = null`, which through the new setter clears the whole selection; dragging the primary collapses the selection to just it. Nothing gets corrupted (that's the good news — the indices don't go stale), but a selection the user built up with ctrl+clicks vanishing on an unrelated reorder is surprising. Remapping the indices around the splice is a couple of lines, and if that's not worth it, a comment saying "a reorder drops the selection on purpose" would at least mark it as a decision.

### Smaller notes

- The new TestCafe case clicks `#deckEditorAlignLeft` and `#deckEditorDistributeV` but never asserts the resulting geometry — only that the hidden object stayed at `x: 200` and the selection survived. The state hash covers the math indirectly, but a direct `expect(objects[0].x).eql(objects[1].x)` (and equal gaps after distribute) would fail with a readable message instead of a hash mismatch if the align formula ever regresses. Given that formula is the one piece of genuinely new arithmetic here, it's worth the two lines.
- A mixed *number* row plus `emptyIsZero` means typing into `(mixed)` and then clearing the field writes `0` to every selected object (`addTypedInput`, line 3035). That is the existing single-object semantics ("an erased number reads as 0"), but across a selection it turns "never mind" into a bulk edit of N objects. Not writing on empty when `mixed` is set would make the abandon case safe.
- `copySelectedObject()` inserts every copy after the highest selected index, so copying a non-contiguous selection makes the copies contiguous and changes their draw order relative to the objects that sat between the originals. Fine as a decision, just not what "the copied group stays a group" implies.
- `commonPropertyValue()` still treats "absent" and "explicitly `null`" as equal (both key to `null`), so an object with `color: null` and one with no `color` at all read as agreeing. Harmless, and I'd leave it — noting it only because the multi-select story is otherwise careful about "a property only some of them have".

Everything else I checked held up: `selectedObjectTemplates()`/`selectedObjectDivs()`/`alignItems()` now use the same `objects[index]` filter so the pairings can't drift; the align/distribute math is a faithful port of `client/js/editor/toolbar/align.js`; `commit()` copies out of the working copy rather than replacing it, so the object references captured by the multi-select row closures stay live; the group headers escape game-defined property names through `html()`; and no engine, validator or `jeCommands` surface is touched.

The `beta` label was already present and I left it as is: the PR changes only `client/css/editor/deckeditor.css`, `client/js/editor/deckeditor.js` and the TestCafe suite — no engine file, no saved-game semantics — and the items above are recoverable editor-UI edge cases, so it stays safe and useful for players to try.

VTT-AI-Agent:
Fixed in 865be467 — all four findings were right.

**1. Align/distribute enabled from the raw count.** `updateAlignToolbar()` now counts `this.alignItems()` instead of `this._selectedObjects`, so a button is only clickable when there is actually something for it to do: one visible object plus a hidden one leaves Align left disabled, and Ctrl+A on a face with hidden objects enables the buttons only if 2 (3 for distribute) of them are drawn. The wrapper tooltip says "Needs N or more **visible** objects selected", so a disabled button still explains itself. It's cheap to call — `alignItems()` only runs from `updateDragToolbar`/render, and the card is rendered by then, as you noted.

**2. Shift/Ctrl+click inside a row's inline text field.** That gesture belongs to the text while the field is the one being typed in. `onmousedown` now records whether the field already had the focus (the browser focuses it on mousedown, so the state at click time is useless), and `onclick` hands the click straight back to the browser when it was focused and a modifier is held. Clicking into a field that *isn't* focused still does the row's multi-select, so text objects can still be ctrl-picked from the tree.

**3. Measuring before the flush.** Both `alignObjects` and `distributeObjects` now `await this.flushPendingCommits()` first and call `alignItems()` after, matching `DeckEditorDragButton.pointerDown`.

**4. Reordering drops a multi-selection.** `moveFaceObject()` remaps the whole selection around the splice instead of looking at the primary only — the dragged object lands on `to`, everything it moved past shifts by one, the rest stays put. Dragging a non-primary row now keeps all three ctrl-selected objects selected.

**Smaller notes**
- **Test geometry.** The case now asserts the arithmetic directly: after Align left the three visible objects share one `x`, and after Distribute vertically the two gaps are equal (±1 for rounding), so a regression in the formula fails with a readable message rather than a hash mismatch.
- **Mixed number + `emptyIsZero`.** `addTypedInput` no longer writes `0` when a *mixed* number field is cleared — empty is the state that field started in, so clearing it again is "never mind" instead of a bulk edit of N objects. Single-object rows keep the existing "an erased number reads as 0".
- **`copySelectedObject` making a non-contiguous selection contiguous.** Kept as the behaviour (the copies are meant to be a block the next drag moves as a whole), but the comment now says so instead of implying the group is preserved as it was.
- **`commonPropertyValue()` treating absent and `null` as equal.** Left alone, as you suggested.

**Tests.** Three new assertions in the multi-selection case, each verified to fail on the old code: the disabled Align left with a hidden object in the selection (`expected false to be truthy`), the shift+click in a focused field (`expected false to be truthy` — the tree rebuild takes the focus away), and the reorder keeping 3 rows selected (`expected 1 to deeply equal 3`). The state hash moved to `101012d22e8e8d136d73d75bc4d4a5f7` because the new reorder step changes the object order. `npm test` (86) and the full `tests/testcafe/editor.js` (17/17) pass locally.

96LawDawg:
@VTT-AI-Agent, 

Clicking on a deck widget while the Edit Mode tab is "Edit Widgets", or when already clicked on a deck and switching from some other tab to "Edit Widgets" should not immediately open the deck editor. Instead it should open a very abbreviated UI that looks like the UI for other widgets at the top and have the "Basic" section:
image

However, for a deck, above the basic section, add a button to open in the deck editor. The only other section in the UI should be "Other properties" that lists anything not in the Basic section and anything not covered by the deck editor in card defaults, face templates, etc.

Finally, next to the "Card View" button in the upper right, (so it will be right next to "Edit Widgets", add a button to close the deck editor.

VTT-AI-Agent:
Thanks @96LawDawg — that's a reasonable idea, but it's a different feature from what this PR does. #3076 only changes the *inside* of the deck editor (multi-select of face objects + grouping the Object/All Cards property rows); it doesn't touch how the deck editor is opened/closed from Edit Widgets mode at all.

What you're describing — a lightweight "Edit Widgets" panel for decks (Basic + Other properties, with a button to jump into the full deck editor) and a close button next to Card View — is a standalone UI change to the JSON editor's widget-selection flow. I'd rather not fold it into this PR's diff. If you'd like it built, it's probably best as its own PR (`!pr`) so it can be reviewed and tested independently of the multi-select/grouping work here.

96LawDawg (approved):
Approved. This adds a couple of nice features to the deck editor and fixes how the deck editor is opened when selecting a deck in edit mode.

96LawDawg:
Going to merge this. Editor only stuff.  Pretty harmless and of course bot-coded and reviewed.

VTT-AI-Agent:
The one red check (**TestCafe Firefox → publiclibrary-2.js**) is a pre-existing flake, not caused by this PR:

- It is `Public library: Reversi (variant 0)` failing the state-hash assertion — `88dfac2e6e117fb96ace4eed65d8681a` vs expected `35e0017570f9ecd206a2317c1528be36`.
- The **exact same** hash mismatch happened ~26 minutes earlier on an unrelated branch (`editable-card-text`, [run 30778232110](https://github.com/ArnoldSmith86/virtualtabletop/actions/runs/30778232110)), which touches nothing this PR touches.
- The **TestCafe Chrome** run of the same `publiclibrary-2.js` passed on this same commit, as did every other job in both suites.
- This PR only touches `client/js/editor/*`, `client/css/editor/*`, `client/editor.html` and `tests/testcafe/editor.js` — edit-mode-only code that the Reversi play-mode test never reaches.

So nothing to fix here; a re-run of that job should go green.

PR #3049, opened 2026-07-24
Properties editor: curated editors for SVG replacements, icons, faces, snap grids and piles, multi-selection editing and devtools-style CSS rows

VTT-AI-Agent:
To Do / Notes

## Summary

This PR builds directly on the curated properties editor, `propertyInputs` framework, and CSS-class editor introduced by #2884 ("Update to Properties editor UI"). #2884 has since merged to `main`, so this branch has been rebased onto `main` and now targets it directly (previously it targeted #2884's `expand-widget-edito` branch while that PR was open).

Implements 3 of 4 follow-on items tracked on the widget-editor follow-ups project board (the 4th was found already implemented — see below), plus two larger items ported from #3014 during review (multi-selection editing and a per-face dice editor).

### 1. Curated editor for `svgReplaces`
Widgets built on `ImageWidget` (basic/button/holder) plus dice/timer support a `svgReplaces` property that maps a color found inside an uploaded SVG (e.g. `#000` or a custom name like `#borderColor`) to a widget property whose value gets substituted for it (see `getSvgReplaces()`/`getImage()` in `imagewidget.js`). Previously only the *derived* color properties referenced by an existing map got curated color pickers (via `renderSvgReplaceColors`) — the map itself was only editable as raw JSON in "Other properties". This adds a proper "SVG replacements" section (Appearance) to add/rename/remove color → property mappings, sitting right above the existing derived color pickers, and lists the colors actually found in the widget's SVG as clickable swatches so you don't have to guess or type hex values.

### 2. Icon picker: basic color/scale options
The icon picker (`propertyInputs.js`) previously only let you pick *which* icon. The underlying icon rendering (`generateSymbolsDiv` in `symbols.js`) already accepts an icon value in object form (`{ name, scale, color, ... }`), so no new widget properties were needed — this adds a persistent "Icon color & scale" block in the Content section that reads/writes that object form, falling back to the widget's own default color/scale (`getDefaultIconColor()`/`getDefaultIconScale()`) when unset. A plain string icon value is preserved as-is until an option is actually used, and collapses back to a plain string if all options are cleared.

### 3. Basic widgets: bigger, reachable HTML field
[...]

PR #3079, opened 2026-07-30
Honor dropLimit on lines and piles

VTT-AI-Agent:
`dropLimit` is enforced in `getValidDropTargets()` (client/js/main.js), which skips a holder once its limit is reached. Two drop paths never go through that check, so `dropLimit` was ignored there:

**Lines** — a line takes a widget dropped onto its *path*, not its bounding box, so the drop never reaches `getValidDropTargets()`. A line with `dropLimit: 3` kept accepting a 4th, 5th, … stop.

**Piles** — piles are formed by `updatePiles()` from `pileSnapRange`, entirely outside the drop-target logic. All four merge cases ignored `dropLimit`: card→pile, pile→pile, pile→card, and card+card (which *creates* a pile, whose limit comes from `onPileCreation`).

A refused drop behaves exactly like it does on a holder that is full: the target is simply not a drop target, nothing lights up, and the widget stays where it was dropped. No new refusal cue, no animation.

### Scope: `dropLimit` stays a drag-only rule

This is deliberately **not** a widening of what `dropLimit` means. Today it constrains manual drags only — routines (`MOVE`, `CLONE`) have their own caps (`fillTo`) and go straight to `moveToHolder()` without looking at it, and so do "Split the pile" and the JSON editor. Existing games rely on that, so all of it keeps working exactly as before:

- a line only refuses a drop aimed at its path; a routine reparenting a widget into a line still makes it a stop via `Line.onChildAdd()`
- the pile checks only run for a pile update that `moveEnd()` marked as coming from a drag, so `MOVE`, `CLONE`, "Split the pile" and the editor merge as they always have

### Showing the limit: `showLimit` on piles

A pile handle shows how many cards the pile holds. With the new pile property **`showLimit`** it shows how many it takes as well — `2/3` instead of `2` — and the pile overlay says "2 of 3 cards". Without it (the default) the handle is the plain count it has always been, so no existing game changes.

The fraction gets a wider handle rather than a smaller font: the circle becomes a pill (`width: calc(var(--phSize)*1.6)`, `border-radius: var(--phSize)`), small handle variant included.

pile handles with and without showLimit

### Setting a pile's limit: on the pile, written to its deck

A pile is temporary, so the limit cannot live on the pile alone — it belongs in the `onPileCreation` template every pile of those cards is created with:

```json
"onPileCreation": { "dropLimit": 3, "showLimit": true }
```

Since #3049, selecting a pile in the properties sidebar edits exactly that template (`setAndMirrorToPileTemplate`, plus its "edit this pile only" switch), so both settings are offered right there in the pile's **Behavior** section: **Drop limit**, and — once a limit is set and the handle shows the card count — **Show the limit on the handle**. Empty means no limit (`-1` never has to be typed), and switching either back off removes the key from the template again.

the drop limit and its display in the pile properties panel

### Setting a holder's or a line's limit: with the widgets it limits

`dropLimit` bounds exactly the drops the widget's **Target widgets** matches let in, so for a holder and a line it is rendered right below those matches instead of in a section of its own. It appears the moment the first match is added — before this, a line's limit was labelled "Maximum stops", lived three sections higher in *Stops*, and only showed up on a line that already accepted something, so it stayed invisible right after the match that made it meaningful. A limit that is already set stays visible even when nothing is accepted, so it can never end up editable nowhere.

the drop limit below the target widgets of a line

### Changes
- **client/js/main.js** — `exceedsDropLimit(target, count, currentCount)`, the single place the rule is spelled out. `getValidDropTargets()` now uses it instead of hand-rolling the same comparison.
- **client/js/widgets/line.js** — `stopDropTarget()` (the "does this line take this widget" predicate) returns `null` once the line lists `dropLimit` stops. A line counts its **stops**, not its children: a `fixedParent` widget rides a line as a stop without ever becoming a child. A widget the line already carries can always be dropped back onto the path, mirroring the "already a child" exemption for holders — so a line that is over its limit does not shed a stop when one is dragged.
- **client/js/widgets/widget.js** — `moveEnd()` marks the drop, and each of the four pile merge cases checks the receiving pile before merging, `continue`ing (rather than `break`ing) so another nearby candidate can still be considered. card+card: if the pile that *would* be created has `onPileCreation.dropLimit` below 2, no pile is created.
- **client/js/widgets/pile.js** + **client/css/widgets/pile.css** — `showLimit` and the wider handle it needs.
- **client/js/editor/sidebar/properties.js** — a curated "Drop limit" input plus the pile's "Show the limit on the handle" checkbox, since both were reachable only through raw JSON before. It is rendered with the drop target matches on a holder and a line, and in *Behavior* for a pile, a multi-selection and anything given a `dropTarget` in the JSON editor. `takesDrops()` names the pile explicitly: it is the one widget that takes drops without a `dropTarget`.
- **validator/validate_gamefile.js** — `showLimit` as a pile property.
- **tests/client/line-widget.test.js** — the limit, the re-drop exemption, stops-vs-children, and that the `-1` default is inert.
- **tests/client/pile-droplimit.test.js** — all four pile merge cases, that a routine still merges into a full pile, and that a card refused by a full pile still joins another one that has room.
- **tests/testcafe/editor.js** — the pile's limit and handle display (including the write into the deck template), and the line's limit appearing with its first drop target match.

`dropLimit: -1` (the default) is unchanged everywhere.

Known and intentional: pile→pile is all-or-nothing (two half-full piles that would fit the other way round still refuse each other), because `updatePiles()` is direction-dependent by design.

Verified by driving a real browser: a full pile refuses a dragged card, the same pile with room for one more takes it, a routine `MOVE` into a holder holding a full pile still merges, `onPileCreation.dropLimit: 1` stops a pile from being created, ordinary card+card piling is untouched, the handle shows `2/3` only with `showLimit` set (plain `2` without it, at both handle sizes), and setting the two on a pile writes `{ "dropLimit": 3, "showLimit": true }` into the deck's `onPileCreation`. Setting the limit on a holder, a line and two holders at once was driven the same way. `npm test` (276/276) plus the TestCafe `routines`, `editor` (32) and `publiclibrary-1` (Solitaire, Mancala) suites pass.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3079/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

The fix is well-targeted and the four pile cases each check the right receiving widget, but the pile half is enforced one layer lower than the line half: `updatePiles()` runs on *every* parent/position change, so `dropLimit` now also blocks piling that a routine (`MOVE`), a clone, or "Split the pile" causes — everywhere else in the codebase `dropLimit` constrains manual drags only (routines deliberately ignore it, and the line half of this very PR is drag-only). That widening deserves an explicit decision. No tests were added even though `tests/client/line-widget.test.js` already has a fixture for exactly this drop path.

## Findings

### 1. The pile checks apply to routine-driven piling, not just drags (`client/js/widgets/widget.js:3219`, `:3243`, `:3254`, `:3266`)

`dropLimit` is, today, purely a *drag* concept: the only enforcement point is `getValidDropTargets()` (`client/js/main.js:46`), and routine operations bypass it entirely — `MOVE` has its own `fillTo` cap and otherwise calls `moveToHolder()` without ever looking at `dropLimit` (`widget.js:1684`, `:1710`). The line half of this PR keeps that contract: `lineTakesDrop()` only filters `stopDropLines` at `moveStart()`, so a routine that reparents a widget into a line still reaches `Line.onChildAdd()` → `addStop()` unchecked.

The pile half does not. `updatePiles()` is invoked from `onPropertyChange` on any `parent` change (`widget.js:2804`), from `clone()` (`:583`), from the pile-split button (`pile.js:187`) and from the JSON editor / sidebar. So after this PR:

- a routine that `MOVE`s cards into a holder that already holds a `dropLimit`-bearing pile silently leaves them **next to** the pile instead of merging — and any downstream logic counting `pile.children()` or `pile.cardCount` sees a different number than before;
- "Split the pile" on cards carrying `onPileCreation: { dropLimit: 2 }` now produces several 2-card piles plus a loose card, all stacked at the same coordinates, instead of one pile.

Either is a plausible break for a game that set `dropLimit` on a pile purely to control *manual* drops. If the intent is drag-only parity with holders, gate the four checks on the drag path (e.g. skip them when `!this.get('dragging')` / when `this.movedByButton` is set); if the wider semantics are intended, please say so in the PR body and in the wiki's `dropLimit` description, since it changes what an existing game file means (cf. `docs/compat.md`).

Also worth noting: when a merge is refused the card just sits on top of the full pile with no feedback at all — visually indistinguishable from a merged pile until you try to pick it up. Holders at least don't light up as `droppable`.

### 2. `exceedsDropLimit()` duplicates `getValidDropTargets()`'s rule instead of replacing it (`widget.js:29` vs `main.js:46`)

`main.js` still hand-rolls `dropLimit > -1 && dropLimit <= children().length`, plus the "don't skip it if the dragged widget is already its child" exemption. The new helper expresses the same rule differently (`current + count > limit`) and expresses the self-exclusion differently again (`lineTakesDrop` always drops the dragged widget from the count). They agree today for `count == 1`, but it's two sources of truth for one property. Having `getValidDropTargets()` call `exceedsDropLimit(t, 1, ...)` would keep them from drifting.

Related asymmetry: for a line, `lineTakesDrop()` counts **stops** while `getValidDropTargets()` counts **children**. Those differ whenever a `fixedParent` widget rides a line (a stop that never becomes a child). It's currently harmless because lines are explicitly skipped in hover hit-testing (`widget.js:2552`, `:2563`), so the `main.js` result for a line only ever controls a `droppable` class that has no line CSS — but the divergence is a trap for whoever touches that hit-testing next. A comment on `lineTakesDrop` saying "stops, not children, are what a line's dropLimit counts" would pay for itself.

### 3. An over-limit line silently sheds a stop (`widget.js:2516`)

With three stops on a line whose `dropLimit` is 2 (easy to produce in edit mode, or by lowering the limit afterwards), dragging any one of them excludes the line from `stopDropLines`, so `lineStopDropTarget()` returns `null` and `applyLineStopDrop(null)` removes the widget from the line — you cannot put it back. That converges toward the limit, which is arguably the right outcome, but it's an untested and undocumented behaviour hiding behind a helper whose comment only promises "dropping it back onto a line it is already on still works". Please state it in the comment either way.

### 4. `pile → pile` refuses in both directions (`widget.js:3243`)

`exceedsDropLimit(widget, this.children().length)` blocks the merge when the *receiving* pile is too full, but the dragged pile may have plenty of room — merging the other way (receiver's children into `this`) would respect both limits. Given `updatePiles()` is direction-dependent by design, all-or-nothing is a defensible choice; just be aware two half-full piles with `dropLimit: 6` and 4 + 4 cards now refuse each other entirely, which reads as "nothing happened".

### 5. No tests (`tests/client/line-widget.test.js`)

`tests/client/line-widget.test.js:566` already has a `describe('dragging a widget onto a line to make it a stop')` block with a `line` + `token` fixture and a hand-set `token.stopDropLines`. A test that sets `dropLimit: 1` on that line, adds one stop, and asserts a second widget's `moveStart()` leaves `stopDropLines` empty — plus the "re-dropping the existing stop still works" case the comment claims — would be a handful of lines and would pin down exactly the two behaviours this PR is about. The four pile cases are harder to unit-test, but the card+card `onPileCreation.dropLimit < 2` branch is pure enough to cover cheaply. (`npm test` passes as-is: 86/86.)

### 6. Nit (`widget.js:29`)

`currentCount !== undefined ? currentCount : target.children().length` is `currentCount ?? target.children().length`.

---

I did **not** add the `beta` label: this PR changes `client/js/widgets/widget.js` in a way that alters engine behaviour for every existing game by default — drag-to-line stop acceptance and all four `updatePiles()` merge paths — so it is excluded from beta auto-deployment by design.

VTT-AI-Agent:
Thanks — the main finding was right, and it's now fixed. Pushed 7a89336.

**1. Pile checks applied to routine-driven piling — fixed, `dropLimit` stays drag-only.** That was the intent all along and the implementation was simply at the wrong layer. `moveEnd()` now marks the drop (`pileUpdateFromDrag`) for the whole drop, not just its final `updatePiles()` call, because dropping into a holder reparents the widget and gets there first. The four pile checks are gated on that marker, so `MOVE`/`CLONE`, "Split the pile" and the JSON editor merge exactly as they did before. I confirmed all of it in a browser: a full pile refuses a dragged card, the same pile with `dropLimit: 3` takes it, a routine `MOVE` into a holder holding a full pile still merges, `onPileCreation.dropLimit: 1` blocks pile creation on a drag, and plain card+card piling is untouched. The PR body now says this explicitly.

**2. Duplicated rule — fixed.** `exceedsDropLimit()` moved to `main.js` next to `getValidDropTargets()`, which now calls it instead of hand-rolling the comparison, so there is one source of truth. The stops-vs-children asymmetry is real and intended (a `fixedParent` widget rides a line as a stop without becoming a child) and is now spelled out in a comment plus a test.

**3. Over-limit line shedding a stop — fixed.** The check moved out of the `moveStart()` filter and into `Line.stopDropTarget()`, and it now mirrors the holder exemption directly: a widget the line already carries always gets its drop target back, even when the line is over its limit. There's a test for that case.

**4. pile→pile refusing in both directions.** Left as is — `updatePiles()` is direction-dependent by design and merging the other way would also move the receiving pile to the dragged one's position. Noted in the PR body as a known limitation, along with the lack of feedback when a merge is refused.

**5. Tests — added.** Four cases on the existing fixture: the limit refuses a second stop, a stop the line already carries can be dropped back, the limit counts stops rather than children, and the `-1` default is inert. `npm test` is 90/90; TestCafe `routines`, `editor` and `publiclibrary-1` (Solitaire, Mancala — pile-heavy) also pass locally.

**6. Nit.** Gone with the rewrite — the helper now uses default parameters instead.

Agreed on not adding `beta`.

VTT-AI-Agent (commented):
## TL;DR

This is a **UI/UX review**, not a code review — a separate reviewer covers correctness. The PR changes no markup or CSS, but it does change what players *see* when they drag, so it has a real interface surface. The rule itself is now enforced consistently, but the **refusal is completely invisible, and in the pile case it actively looks like the drop succeeded** — the card lands on top of the pile with a small offset, the count badge still reads the old number, and nothing tells you it was rejected. The single most impactful improvement: give a refused drop the same visual vocabulary the accepted drop already has (`droppable` / `lineDropTarget`), rather than degrading it to "nothing happened".

I checked this out and drove it in a real browser (Playwright, local server) with a pile at `dropLimit: 2` and a line at `dropLimit: 1`. Screenshots below are from that run.

---

## Big layout / interaction improvements

### 1. A refused pile merge is indistinguishable from a successful one

Dragging a third card onto a pile that is already at `dropLimit: 2`:

refused pile drop

The card is simply left where the mouse released it — dead centre on the pile, offset by a few pixels, its face text overprinting the pile's face text ("CARD" over "CARD"). The pile's handle badge still reads **2**. To a first-time player this reads as "it went in, the badge just hasn't updated" or as a rendering glitch, not as "this pile is full".

The PR text calls this out as known and intentional ("a refused merge simply leaves the card lying on the pile with no extra feedback"), and I understand not wanting to grow the diff — but this is the one place where the new rule is *visible* to a player, and right now it is visible as a mess. Two concrete options that fit the existing style:

- **Cheapest:** in the four refusal branches in `client/js/widgets/widget.js` (~3211, 3238, 3249, 3261), when `isFull(...)` refuses, restore the widget's pre-drag `x`/`y` instead of leaving it on the pile. The drag visibly "bounces back", which is a conventional and self-explanatory refusal, and it removes the overlap mess for free.
- **Better:** add a short transient cue. There is no toast primitive in the client, but there is an established one-shot animation pattern — `@keyframes flash` at `client/css/editor/sidebarModules.css:409-416`. A table-level sibling (`.dropRefused`, ~600ms, using the app's existing red/grey rather than a new colour) applied to the *target* pile would say "this thing rejected you" using a language the app already speaks.

### 2. A line at its limit gives literally zero feedback

`stopDropTarget()` now returns `null` once the line is at its limit (`client/js/widgets/line.js:517-521`). Because `highlightStopDropLine()` (`widget.js:2640-2649`) only ever adds `lineDropTarget` for a non-null target, the line at its limit behaves exactly like empty table. Compare hovering the same token over the same line, below the limit vs at it:

| accepted (glow) | refused (nothing) |
| --- | --- |
| accepted | refused |

The second image is a widget sitting directly on top of the line's path with no reaction at all. A first-time user's model here is "lines take widgets" — the absence of the glow they saw a moment ago on the *same line* reads as a bug, not as a rule.

Suggested shape, staying inside the existing feedback vocabulary: rather than `return null`, return the hit with a marker (e.g. `{ line: this, position, distance, full: true }`), have `lineStopDropTarget()`/`applyLineStopDrop()` treat `full` as "do not attach", and have `highlightStopDropLine()` apply `lineDropTargetFull` instead of `lineDropTarget`. Then a 3-line addition next to `client/css/widgets/line.css:101-106` — the same `drop-shadow()` treatment in a muted/red hue instead of `#1f5ca6` — makes "full" a first-class visual state that is obviously the negative of the accepted state.

### 3. One rule, three different silences

A nice thing about this PR is that `exceedsDropLimit()` makes the rule readable in one place. The *feedback* is still three different non-behaviours:

- holder at its limit → filtered out of `dropTargets` at `client/js/main.js:53-56`, so it never gets `droppable` and looks like a widget that was never a drop target
- line at its limit → no highlight (above)
- pile at its limit → widget dropped on top of it (above)

Since the predicate is now unified, it would be worth unifying the presentation in the same PR or an immediate follow-up: wherever `exceedsDropLimit()` returns true, mark the target `full` and style it once. Otherwise a player who learns "greyed-out means full" from holders learns nothing that transfers to lines or piles.

---

- **Overlapping faces after a refused pile drop.** The refused card lands within a few pixels of the pile's top card, so the two faces overprint (screenshot above). Even with no new feedback mechanism, restoring the pre-drag position (or nudging by `stackOffsetX`/`stackOffsetY` × 2) avoids shipping a state that just looks broken.
- **A refused line drop leaves a widget that looks exactly like a stop.**
  refused line drop
  Here `A` is a genuine stop and `B` is a rejected drop — visually identical, both centred on the path. Worse, `B` is not attached, so the next time the line moves it will be silently left behind, which will read to the player as the line "dropping" a piece. At minimum, offset a refused widget clear of the path by the hit `range` computed at `line.js:512`.
- **Pile→pile is direction-dependent with no visible difference between the two piles** (acknowledged in the PR description). From the table, pile A and pile B look the same; dragging A→B refuses while B→A works. If that asymmetry stays, the pile handle is the only surface that could hint at it.
- **The pile handle is the natural home for the limit.** `.pile .handle` (`client/css/widgets/pile.css:5`) already renders the child count. When `dropLimit > -1`, rendering `2/2` instead of `2` would make the whole feature self-explanatory at a glance, in the app's own idiom, with no new UI. Same for the "N cards" title in the pile overlay (`client/js/widgets/pile.js:83`).

---

**For players (first-time users):** nothing in the interface ever states that a limit exists. The only way to discover it is to try a drop and have it silently not work — the classic "is this broken or am I doing it wrong?" moment. The `2/2` handle idea above, plus any of the refusal cues in §1/§2, would close that gap entirely.

**For game designers using the editor** (the other first-time audience this touches):

- `dropLimit` has **no entry in `editorPropertyHints`** (`client/js/editor/sidebar/properties.js:345+`) and no curated input anywhere in the properties sidebar — it is reachable only through raw JSON / "other properties". This PR makes its semantics considerably subtler (drag-only; a line counts **stops**, not children; a pile's effective limit comes from `onPileCreation.dropLimit`), and none of that is discoverable. Sibling hints already exist for `preventPiles` and `alignChildren`; suggested wording in the same register:
  > `dropLimit: 'The most widgets this can hold when a player drags something in. Routines and the editor ignore it. On a line it counts stops. -1 means no limit.'`
- The in-browser JSON editor has an insert helper for `dropTarget` (`client/js/jsonedit.js:1826-1830`) but none for `dropLimit`, so the property a designer now needs alongside `dropTarget` is the one with no button.
- **`onPileCreation.dropLimit` below 2 now silently prevents a pile from ever forming** (`widget.js:3214-3218`). That is a genuinely useful capability, but it is an invisible rule expressed through an unrelated-sounding number — worth an explicit sentence in the wiki/property hint, e.g. "set `onPileCreation.dropLimit` to 1 to stop cards from piling up at all".

---

*Reviewed by checking out the branch, running the server locally and driving the drags with Playwright; the four screenshots are from that session. Everything above is UI/UX only.*

VTT-AI-Agent:
Thanks — the core point was right: the rule was invisible, and in the pile case it looked like the drop had worked. Fixed in 23956aa, verified in a browser again.

**1. A refused pile merge looked like a successful one — fixed, both halves of the suggestion.** The widget that refuses now flashes (`.dropRefused` in `client/css/layout.css`, a 1s `drop-shadow` in the refusal red — the same vocabulary as the line glow rather than a new colour), *and* the dropped widget goes back where the drag picked it up, so the overprinting mess is gone with it. The bounce-back only applies while the drop did not move the widget somewhere else in the meantime: if it was dropped into a holder that happens to contain a full pile, that move stands and only the merge is refused (otherwise a refused merge would yank a card out of the holder a player deliberately dropped it into).

**2. A line at its limit gave zero feedback — fixed, as suggested.** `stopDropTarget()` no longer returns `null`; it returns the hit marked `full`, `lineStopDropTarget()`/`applyLineStopDrop()` treat `full` as "do not attach", and `highlightStopDropLine()` applies `lineDropTargetFull`, styled next to `lineDropTarget` in `line.css` with the same `drop-shadow()` treatment in red. A line that still has room now also wins over a closer one that is full. And because the drop is refused rather than silently dropped on the path, a refused widget no longer sits there looking like a real stop — it goes back where it came from.

| accepted | at its limit |
| --- | --- |
| accepted | refused |

**3. One rule, three silences — two unified, holders deliberately left alone.** Lines and piles now share one refusal vocabulary. Holders keep theirs: a holder has no built-in `droppable` styling at all (`holder.css` only changes the border on `.droppable.droptarget`; everything else is per-game, via the sidebar's "When a widget can be dropped here"). Giving a full holder a built-in "full" look would mean shipping a negative state that has no positive counterpart, and every existing game with a `dropLimit` holder would suddenly show a new visual. If a built-in droppable look ever lands, the full state belongs with it.

**4. Pile handle as the natural home for the limit — done.** With `dropLimit > -1` the handle reads `2/2` instead of `2` (`.withLimit` shrinks the font so it still fits the circle), and the pile overlay title says "2 of 3 cards". That also gives the pile→pile direction asymmetry the one surface that can hint at it.

pile refusing a card

**5. Readability for game designers — hint added, curated input added; the JSON-editor button already exists.** `dropLimit` now has an `editorPropertyHints` entry, close to your wording, and a curated input that shows it: "Drop limit" in the Behavior section of anything with a `dropTarget`, and "Maximum stops" in the line's Stops section. Empty means no limit (`-1` never has to be typed).

> `dropLimit: 'The most widgets a player can drag in here. Routines, the JSON editor and "Split the pile" ignore it, so they can still put in more. On a line it counts the stops the line carries. Leave empty for no limit.'`

Drop limit in the sidebar

One correction: the JSON editor *does* already have an insert button for `dropLimit`. `jeAddWidgetPropertyCommands()` walks `object.defaults`, and `dropLimit` is a base `Widget` default (`widget.js:124`), so every widget type gets `widget_<type>_dropLimit` — unlike `dropTarget`, which only needs the special case in `jeAddWidgetPropertyCommand()` because it inserts a `###SELECT ME###` template. The wiki sentence about `onPileCreation.dropLimit` stopping cards from piling up at all is a good idea, but the wiki is outside this repo — worth a follow-up by someone with access.

`npm test` is 92/92 (two new line tests: a full line refuses the drop and remembers who refused it; a line with room wins over a closer full one); the TestCafe `editor`, `routines` and `publiclibrary-1` (Solitaire, Mancala) suites pass locally.

RaphaelAlvez:
dropLimit behavior should be the same as for holders and other widgets. remove the animation

Add a showLimit in piles otherwise don't display it as a fraction

instead of reducing the font use this
"handleCSS": {
    "font-size": "calc(var(--phSize)/2)",
    "width": "calc(var(--phSize)*1.6)",
    "border-radius": "var(--phSize)"
  }

also addpat for the small variant

VTT-AI-Agent:
Done, all four points — pushed in 0b50c68.

> dropLimit behavior should be the same as for holders and other widgets. remove the animation

Removed. The `dropRefused` flash, the pre-drag bounce-back and the red `lineDropTargetFull` glow are all gone, and `Line.stopDropTarget()` returns `null` again once the line is at its limit. So a full line or a full pile now refuses exactly the way a full holder does: it is simply not a drop target, nothing lights up, and the widget stays where it was dropped. That took ~110 lines back out of `widget.js`/`layout.css`/`line.css`.

> Add a showLimit in piles otherwise don't display it as a fraction

New pile property `showLimit` (default `false`). Only with it set does the handle read `2/3` and the pile overlay say "2 of 3 cards"; without it the handle is the plain count it has always been, so no existing game changes. Usable per-pile or on the deck so every pile made from those cards gets it:

```json
"onPileCreation": { "dropLimit": 3, "showLimit": true }
```

Added to `validate_gamefile.js` as well; the JSON editor button comes for free since `jeAddWidgetPropertyCommands()` walks the pile defaults.

> instead of reducing the font use this "handleCSS": { "font-size": "calc(var(--phSize)/2)", … }
> also adapt for the small variant

Used your values — the `.withLimit` handle keeps the normal font size and becomes a pill instead of a circle:

```css
.pile .handle.withLimit {
  width: calc(var(--phSize)*1.6);
  border-radius: var(--phSize);
}

.pile .handle.small.withLimit {
  width: calc(var(--phSize)*0.8);   /* the small handle is phSize/2, so *1.6 of that */
  border-radius: var(--phSize);
}
```

`font-size` needed no rule at all: `calc(var(--phSize)/2)` is already what `.pile .handle` sets, and the small variant keeps its `calc(var(--phSize)/4)` — dropping the old shrink rules leaves both at their normal size.

pile handles with and without showLimit

Left to right: `dropLimit: 3, showLimit: true`, the same pile without `showLimit`, and a small handle with `dropLimit: 12, showLimit: true`.

Verified in a browser (`2/3` pill 64px wide at 20px font, `2/12` small pill 32px wide at 10px font, plain `2` circles unchanged). `npm test` 90/90, TestCafe `editor` (16) and `publiclibrary-1` (5) pass locally. The PR description is updated to match.

VTT-AI-Agent (commented):
## TL;DR

The engine-side fix now preserves the existing drag-only contract and has focused coverage for the line and all four pile merge paths. One creator-facing gap remains: the new curated `dropLimit` control cannot configure piles, even though pile snapping is one of the two behaviors this PR adds.

## Finding

### The sidebar's “takes drops” test excludes piles (`client/js/editor/sidebar/properties.js:3998-4005`, `:4418-4432`)

`takesDrops()` equates accepting a drop with having a non-empty `dropTarget`. That is true for holders and lines, but not for piles: `updatePiles()` accepts overlapping cards through `pileSnapRange` specifically without going through `dropTarget`, which is why this PR needs separate pile enforcement in the first place. As a result, a normal pile (`dropTarget: null`) never gets the new `Drop limit` input. `renderForPile()` also returns after showing the “pile properties cannot be edited here” message, so selecting the pile offers no alternate control; selecting its deck does not expose `onPileCreation.dropLimit` or the new `showLimit` as curated fields either.

This leaves the pile half of the feature configurable only through raw JSON despite the PR adding a discoverable editor control and hint for the property. Please expose the limit (and ideally `showLimit`) in the deck's pile-creation settings, or otherwise make the curated condition account for piles without suggesting that they require `dropTarget`.

I did not add the `beta` label because `client/js/widgets/line.js`, `client/js/widgets/pile.js`, and `client/js/widgets/widget.js` genuinely alter engine behavior for line-stop acceptance and drag-driven pile merging, making the PR engine-ineligible.

VTT-AI-Agent:
Fixed in 15f0c57 — the pile half of the feature is no longer raw-JSON-only.

> `takesDrops()` equates accepting a drop with having a non-empty `dropTarget`. That is true for holders and lines, but not for piles […] Please expose the limit (and ideally `showLimit`) in the deck's pile-creation settings, or otherwise make the curated condition account for piles without suggesting that they require `dropTarget`.

Right on both counts, and the deck's pile-creation settings were the better of the two options: a pile is temporary, so the pile itself cannot carry the setting — only the `onPileCreation` template its cards are piled with can. Widening `takesDrops()` would have put a "Drop limit" input on the pile that outlives nothing.

**The deck editor's Card defaults tab now has a `Piles` section** ("Every pile made from these cards") with `Drop limit` (empty = no limit, so `-1` never has to be typed) and `Show limit`, writing `cardDefaults.onPileCreation.dropLimit` / `.showLimit`. Switching either back off removes the key again, and an emptied `onPileCreation` is deleted outright, so a deck that never had one stays byte-identical.

the Piles section in the deck editor

That tab is where selecting a deck lands (`renderForDeck()` opens the deck editor, first open shows Card defaults), and it is one click from a selected pile via the existing "Open deck properties" button. `renderForPile()`'s note now names the section instead of just saying pile properties cannot be edited here, and the comment on `takesDrops()` records why piles are deliberately not part of that predicate.

Verified in a browser: setting Drop limit 3 + Show limit on the deck, then dragging one card onto another, creates a pile with `dropLimit: 3, showLimit: true` and a handle reading `2/3`; the values round-trip through a reload; clearing them leaves `cardDefaults` with just `width`/`height`. `npm test` 101/101 and the TestCafe `editor` suite (16/16) pass locally.

No `beta` label added — agreed, and I do not add it in any case.

VTT-AI-Agent (commented):
## TL;DR

The current head looks good: the earlier drag-only compatibility issue and pile-editor discoverability gap are resolved, and I found no further actionable correctness or security problems. The focused tests cover line limits and all four pile merge paths, while the full CI and CodeQL runs are green.

## Detailed assessment

No further findings. I re-checked the manual-drag marker across holder reparenting and the final pile update, the existing-stop exemption for over-limit lines, each receiving-pile count, card-to-card pile creation, `showLimit`'s backward-compatible default and DOM refresh paths, the deck editor's `onPileCreation` persistence/removal behavior, and validator coverage. The implementation preserves routine/editor/split-pile behavior while limiting only manual drops, and the added Jest cases pin down that distinction.

I did not add the `beta` label because `client/js/widgets/line.js`, `client/js/widgets/pile.js`, and `client/js/widgets/widget.js` genuinely change default engine behavior for line-stop acceptance and drag-driven pile merging, making the PR engine-ineligible.

96LawDawg:
I started to review this, but it needs to wait on #3049 to be merged and incorporate its way of dealing with piles instead of throwing this onto the deck editor.

VTT-AI-Agent:
Understood — I will hold off and rework this once #3049 is merged.

Looking at that branch, its pile handling is clearly the better home for these two settings: editing a pile in the properties sidebar mirrors the property into the deck's `onPileCreation` template (`setAndMirrorToPileTemplate`, plus the "this pile only" switch), which is exactly the problem the deck-editor `Piles` section was working around here. So when #3049 lands I will drop that section from `deckeditor.js` and expose `dropLimit` and `showLimit` as curated pile properties in the sidebar instead, letting the template mirroring do the persisting.

The engine half of this PR (the line and pile enforcement in `line.js` / `widget.js` / `pile.js`) is independent of that and should not need to change.

VTT-AI-Agent:
Rebased onto `main` (merge commit) now that #3049 has landed, and reworked the editor half the way I said I would.

> I started to review this, but it needs to wait on #3049 to be merged and incorporate its way of dealing with piles instead of throwing this onto the deck editor.

Done — the deck editor's `Piles` section is gone, `deckeditor.js` is byte-identical to `main` again. Both settings now live in the pile's own properties panel: **Drop limit** in the Behavior section, and **Show the limit on the handle** right below it, appearing once a limit is set and the handle actually shows the card count (with a fixed handle text it would change nothing, so it hides). #3049's `setAndMirrorToPileTemplate` does the persisting, so an edit lands in the pile *and* in the `onPileCreation` template of the cards' deck — and the "edit this pile only" switch works on these two like on everything else.

the drop limit and its display in the pile properties panel

Merge conflicts were in `widget.js` (`moveEnd()` — `main`'s "apply the release coordinates first" block and this branch's drag marker are independent, both kept) and in `properties.js` (`renderForPile()` — `main`'s new pile panel replaces the "pile properties cannot be edited here" placeholder this branch had touched).

Verified in a browser on the merged branch: setting `Drop limit: 3` on a pile makes the checkbox appear, ticking it turns the handle into `2/3`, and the deck ends up with `"onPileCreation": {"dropLimit": 3, "showLimit": true}`; neither property is duplicated in "Other properties"; a holder still gets its Drop limit input and no Show-limit row; the deck editor opens unchanged. `npm test` (276/276) and the TestCafe `editor` (31), `routines` (4) and `publiclibrary-1` (5) suites pass locally.

VTT-AI-Agent:
> add drop limit to piles and lines UI

([Discord](https://discord.com/channels/770758631146782780/1532497728104431646/1534293493864857670))

Both are in the properties sidebar now, without raw JSON.

**Piles** — `Drop limit` and `Show the limit on the handle` in the pile's **Behavior** section (pushed earlier today). #3049's mirroring writes them into the `onPileCreation` template of the cards' deck, so they survive the pile.

**Lines** — this one turned out to be half-broken, so I moved it: a line's limit used to be called `Maximum stops` and sat in *Stops*, three sections above the **Target widgets** matches that decide what a drop is — and it was only rendered when the line already accepted something, so adding the very first match left it invisible until the panel was drawn again. It is now called `Drop limit` on every widget and is rendered directly below the drop target matches it constrains, for **holders** as well as lines, appearing the moment the first match is added:

the drop limit below the target widgets of a line
the drop limit below the target widgets of a holder

It hides again when the widget accepts nothing — unless a limit is already set, which keeps it visible so a leftover value is never editable nowhere. A pile keeps it in *Behavior* (it has no drop target matches: it snaps cards in through `pileSnapRange`), and so do a multi-selection and anything given a `dropTarget` in the JSON editor.

Covered by two TestCafe additions in `tests/testcafe/editor.js` (the pile's limit + handle display + template write, and the line's limit appearing with its first match). `npm test` 276/276, TestCafe `editor` 32/32.

96LawDawg (approved):
Tested.  Approved.  2 minor changes to pile and line tutorials made. Ready to merge.

PR #3080, opened 2026-07-30
Rename seat properties (seatedText/emptyText/emptyColor/seatedColor) and make the generic 'display' work on seats

VTT-AI-Agent:
Seats used `display` for the text shown while occupied, which blocked the generic widget `display` show/hide boolean from working on them. This renames the seat properties and frees `display` up.

### Property changes (widget type `seat`)
| old | new |
| --- | --- |
| `display` | `seatedText` |
| `displayEmpty` | `emptyText` |
| `colorEmpty` | `emptyColor` |
| — | `seatedColor` (new, default `playerColor`) |

- `display` now behaves on seats exactly like on every other widget (the `type != 'seat'` exception in `Widget.classes()` is gone).
- `seatedColor` decides the seat's color while a player sits in it: the special default value `playerColor` keeps the previous behavior (use the color of whoever sits down); any other value is passed through to `color` verbatim.

### Editor
- The "Use player color" toggle and the "Fixed color" style preset now write `seatedColor` instead of the `.seated { --color: <color> !important }` css override hack.
- Changing `seatedColor`/`emptyColor` applies right away to a seat that is currently in that state, instead of only on the next sit-down/stand-up; the two property writes go out as one delta/undo step.
- Preset selection highlighting compares the structural keys (`width`/`height`/`borderRadius`/`css`) plus whether `seatedColor` is `playerColor` or a fixed color - tweaking a color no longer de-selects the preset.
- "When empty" and "When seated" are symmetric: both have a "Text" and a "Color" row, the placeholder chips sit on their own line so the two rows line up, and the preset thumbnails show a short sample name with the preset label below the preview.
- The "Fixed color" preset leaves the empty seat neutral grey instead of pinning both states to the same red, so a free seat stays recognizable as free.
- `display` gets a curated "Visible" checkbox in the generic section of every widget type - that is the capability this PR unlocks for seats, and it had no control at all before.

### File updater (v22)
`server/fileupdater.mjs` migrates existing games. The rename is structural: it only rewrites places where a property *name* is expected — property keys, the property-name fields of routine operations (`SET`/`GET`/`SELECT`/`SORT`/`SCORE`/`RESET`), the name maps `inheritFrom`/`svgReplaces`/`dynamicProperties` and `${PROPERTY …}` references. Ordinary strings — label text, html, css class names, widget IDs and references to them — are never touched, even when they contain one of the old names.
- renames the properties on seat widgets;
- `displayEmpty`/`colorEmpty` only ever meant something on a seat, so they are renamed in all of those places regardless of widget type. A `GET` that took its variable name from the property it read keeps that variable name (it just becomes explicit), so a `${colorEmpty}` further down the routine still resolves;
- the old "Fixed color" preset's `.seated { --color: <color> !important }` rule is converted into `seatedColor`, and a seat that is occupied at migration time keeps that color;
- `display` is ambiguous (a boolean on every other widget), so it is only renamed where it is provably a seat's:
  - on `type: 'seat'` widgets,
  - in `${PROPERTY display OF <known seat id>}` expressions and in `${PROPERTY display}` inside a seat,
  - in `SET`/`GET`/`SELECT` operations working on a collection that provably holds nothing but seats,
  - in `SET property: 'display'` operations whose value contains the `playerName`/`seatIndex` placeholders, unless the collection provably holds no seat at all.

Known limits of that last point: a `SET display` on a collection whose contents cannot be determined statically (e.g. filled inside a nested `IF`/`FOREACH`, or a `CLONE` that overrides `display` on a seat) is left alone. Such a seat keeps working but loses its label — and `SET display ''`/`false` now hides it instead of blanking the text.

**Behavior change worth a changelog line:** a routine that hides a broad selection (`SELECT` → `SET display false`) used to skip seats because of the `Widget.classes()` exception. It now hides them like any other widget. That is the point of the PR, and the migration rewrites the cases it can prove, but games doing this dynamically will notice.

The wiki's Widgets page still documents `display`/`displayEmpty`/`colorEmpty` for seats and needs updating alongside this. Worth covering there: `seatedColor`'s magic `playerColor` value, and that `seatedColor`/`emptyColor` only take effect the next time the seat changes hands when they are set from a routine or the JSON editor (the properties panel writes `color` through for you).

### Verification
- `npm test` — including a new `tests/server/fileupdater.test.js` pinning down every rule of the v22 step (seat vs non-seat `display`, `PROPERTY … OF`, the collection tracking, the css conversion).
- Ran the v22 step over all 571 library/tutorial state files: only seat text/color properties change, no `display` on a non-seat is touched, and no seat ends up hidden.
- Two clients in a browser: a player sits down and disconnects, then the editor switches that seat to a fixed color and back to "Use player color" — the disconnected player's color comes back.
- Booted the server with `minifyJavascript: true` (the failing "Production Environment" check) and drove a browser through: a v21 state with the old property names and the old fixed-color css hack, sitting down in the seats, and a `SELECT type: seat` → `SET display false` routine — text and colors come out exactly as before the rename.
- TestCafe `editor.js`, `routines.js` and `publiclibrary-1.js` pass locally.

`validator/validate_gamefile.js` and `server/pcioimport.mjs` updated too.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3080/pr-test (or any other room on that server)

### DIFF
--- a/client/css/editor/propertyInputs.css
+++ b/client/css/editor/propertyInputs.css
@@ -773,6 +773,15 @@
   padding-right: 0;
 }
 
+.editorModule .compactStyledText .placeholderChipRow {
+  display: flex;
+  gap: 6px;
+  /* full row below the input, indented to the width of the label column so the
+     text/color/size controls above stay aligned with the placeholder-less rows */
+  flex: 1 0 100%;
+  margin-left: 138px;
+}
+
 .editorModule .compactStyledText .placeholderChip {
   font-family: monospace;
   font-size: 11px;
@@ -934,4 +943,18 @@ body.editorWidgetPicking, body.editorWidgetPicking .widget {
   display: flex;
   flex-wrap: wrap;
   gap: 4px;
+}
+
+.editorModule .seatPresetCell {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+}
+
+/* the preset name below the preview instead of inside the sample seat, where
+   longer names wrapped and crowded the box */
+.editorModule .seatPresetCell .seatPresetLabel {
+  font-size: 12px;
+  text-align: center;
+  max-width: 156px;
 }
--- a/client/js/editor/sidebar/properties.js
+++ b/client/js/editor/sidebar/properties.js
@@ -4596,9 +4596,9 @@ class PropertiesModule extends SidebarModule {
       return;
     }
     const seatedColor = widget.get('seatedColor');
-    const { activePlayers, activeColors } = getPlayerDetails();
+    // playerColors covers players who are no longer connected but still seated
     const color = !seatedColor || seatedColor == 'playerColor'
-      ? activeColors[activePlayers.indexOf(player)] : seatedColor;
+      ? getPlayerDetails().playerColors[player] : seatedColor;
     if(color)
       this.inputValueUpdated(widget, 'color', color);
   }
--- a/client/js/editor/sidebar/properties.js
+++ b/client/js/editor/sidebar/properties.js
@@ -4541,7 +4541,10 @@ class PropertiesModule extends SidebarModule {
       button.title = label;
 
       // the presets differ in more than just css now, so compare every key they touch
-      const signature = state => JSON.stringify(Array.from(styleKeys).sort().map(k => state(k) ?? null));
+      const signature = state => JSON.stringify(Array.from(styleKeys).sort().map(k => {
+        const value = state(k);
+        return value === undefined ? null : value;
+      }));
       // a key the preset does not set is reset to null, i.e. back to the widget default
       const presetSignature = signature(k => k in preset ? preset[k] : widget.defaults[k]);
       const update = () => button.classList.toggle('selected', signature(k => widget.get(k)) === presetSignature);
@@ -4553,9 +4556,25 @@ class PropertiesModule extends SidebarModule {
         setDeltaCause(`${getPlayerDetails().playerName} applied the ${label} seat preset to ${widget.id} in editor`);
         for(const key of styleKeys)
           await widget.set(key, key in preset ? preset[key] : null);
+        this.applySeatedColor(widget);
         batchEnd();
       };
     }
+  }
+
+  // A seat's color is written when a player sits down, so changing seatedColor
+  // while somebody is already sitting there would otherwise only show up the
+  // next time the seat is taken.
+  applySeatedColor(widget) {
+    const player = widget.get('player');
+    if(!player)
+      return;
+    const seatedColor = widget.get('seatedColor');
+    const { activePlayers, activeColors } = getPlayerDetails();
+    const color = !seatedColor || seatedColor == 'playerColor'
+      ? activeColors[activePlayers.indexOf(player)] : seatedColor;
+    if(color)
+      this.inputValueUpdated(widget, 'color', color);
   }
 
   // "Use player color" toggle: it edits the seatedColor property, whose special
@@ -4568,10 +4587,11 @@ class PropertiesModule extends SidebarModule {
     };
     const usesPlayerColor = () => readColor() === null;
     const write = value => {
+      batchStart();
+      setDeltaCause(`${getPlayerDetails().playerName} changed the seated color of ${widget.id} in editor`);
       this.inputValueUpdated(widget, 'seatedColor', value === null ? 'playerColor' : value);
-      // a fixed color applies right away to a seat that is already occupied
-      if(value !== null && widget.get('player'))
-        this.inputValueUpdated(widget, 'color', value);
+      this.applySeatedColor(widget);
+      batchEnd();
     };
 
     new CheckboxInput(this, widget, 'Use player color', {
--- a/client/js/editor/sidebar/properties.js
+++ b/client/js/editor/sidebar/properties.js
@@ -390,7 +390,7 @@ const editorPropertyHints = {
   firstColWidth: 'Fixed width of the first scoreboard column when autosizing is off.',
   player: 'The player currently seated here. Setting it also stores their color.',
   index: 'Seat order used by turn and scoreboard logic.',
-  display: 'The seat property whose value is displayed as the seat text.',
+  display: 'Whether the widget is shown at all. A hidden widget stays in the game and can be brought back by a routine.',
   hand: 'ID of the holder used as this seat\'s private hand.',
   turn: 'Marks this seat as currently taking its turn.',
   skipTurn: 'Skip this seat when routines advance turns.',
@@ -545,11 +545,13 @@ const SEAT_PRESET_BACKGROUND = {
   }
 };
 
+// the empty color stays neutral so a free seat is still recognizable as free -
+// with the same red on both sides, taken and free seats look identical
 const SEAT_PRESET_FIXED = {
   "width": 172,
   "height": 48,
   "borderRadius": 8,
-  "emptyColor": "#ff0000",
+  "emptyColor": "#999999",
   "seatedColor": "#ff0000",
   "css": {
     "default": {
@@ -1938,7 +1940,7 @@ class PropertiesModule extends SidebarModule {
   }
 
   basicPropertyExcludeList(extra = []) {
-    return [ 'x', 'y', 'z', 'layer', 'rotation', 'movable', 'movableInEdit', 'width', 'height', 'clickable', 'enlarge', 'ignoreZoom', 'fixedParent', 'linkedToSeat', 'onlyVisibleForSeat', 'inheritFrom' ].concat(extra);
+    return [ 'x', 'y', 'z', 'layer', 'rotation', 'movable', 'movableInEdit', 'width', 'height', 'clickable', 'display', 'enlarge', 'ignoreZoom', 'fixedParent', 'linkedToSeat', 'onlyVisibleForSeat', 'inheritFrom' ].concat(extra);
   }
 
   isSizeRatioLockEnabled(widget) {
@@ -3097,6 +3099,7 @@ class PropertiesModule extends SidebarModule {
     });
 
     this.renderCollapsibleSection('Generic', true, body=>{
+      this.renderCheckbox(widget, 'Visible', 'display', body, editorPropertyHints.display);
       this.renderCheckbox(widget, 'Clickable', 'clickable', body);
       this.renderNumberWithSlider(widget, 'enlarge', 'Enlarge', body, { min: 0, step: 1, slider: false });
       this.renderCheckbox(widget, 'Ignore zoom', 'ignoreZoom', body);
@@ -4497,7 +4500,17 @@ class PropertiesModule extends SidebarModule {
 
     this.addAppearanceSubTitle('When empty');
     this.renderCompactStyledTextInput(widget, 'Text', 'emptyText', 'default');
-    new ColorInput(this, widget, 'Inactive color', { property: 'emptyColor', hint: 'The border color while no player sits here.' }).render(this.moduleDOM);
+    new ColorInput(this, widget, 'Color', {
+      property: 'emptyColor',
+      hint: 'The seat\'s color while nobody sits here.',
+      setValue: value => {
+        batchStart();
+        setDeltaCause(`${getPlayerDetails().playerName} changed the empty color of ${widget.id} in editor`);
+        this.inputValueUpdated(widget, 'emptyColor', value);
+        this.applySeatColor(widget);
+        batchEnd();
+      }
+    }).render(this.moduleDOM);
     new CheckboxInput(this, widget, 'Hide when unused', { property: 'hideWhenUnused', hint: editorPropertyHints.hideWhenUnused }).render(this.moduleDOM);
 
     this.addAppearanceSubTitle('When seated');
@@ -4528,6 +4541,13 @@ class PropertiesModule extends SidebarModule {
     for(const { preset } of Object.values(presets))
       Object.keys(preset).forEach(k => styleKeys.add(k));
 
+    // the colors are edited right below the preset row and are meant to be
+    // customized, so the highlight compares the structural keys exactly and
+    // reduces seatedColor to what actually tells the presets apart: the seated
+    // player's color vs a fixed one
+    const structuralKeys = Array.from(styleKeys).filter(k => k != 'emptyColor' && k != 'seatedColor').sort();
+    const usesPlayerColor = value => !value || value == 'playerColor';
+
     const row = div(this.moduleDOM, 'seatPresetRow');
     for(const [ presetKey, { label, preset } ] of Object.entries(presets)) {
       // a unique id per preview keeps each seat's scoped css (#w_<id>.seated ...)
@@ -4535,16 +4555,18 @@ class PropertiesModule extends SidebarModule {
       // the state - renderReadonlyCopyRaw deletes state.id)
       const previewSeat = new Seat(`seatPreview_${widget.id}_${presetKey}`);
       // a sample color so "Background color" shows a colored seat; "Fixed color"
-      // overrides it with its own seatedColor
-      const previewColor = preset.seatedColor && preset.seatedColor != 'playerColor' ? preset.seatedColor : '#1e88e5';
-      const button = this.renderWidgetButton(previewSeat, Object.assign({ type: 'seat', player: label, color: previewColor, width: 150, height: 44 }, preset), row);
+      // overrides it with its own seatedColor. The size is forced after the
+      // preset so a wider preset still fits the preview button.
+      const previewColor = usesPlayerColor(preset.seatedColor) ? '#1e88e5' : preset.seatedColor;
+      const cell = div(row, 'seatPresetCell');
+      const button = this.renderWidgetButton(previewSeat, Object.assign({ type: 'seat' }, preset, { player: 'Player', color: previewColor, width: 150, height: 44 }), cell);
       button.title = label;
+      div(cell, 'seatPresetLabel').textContent = label;
 
-      // the presets differ in more than just css now, so compare every key they touch
-      const signature = state => JSON.stringify(Array.from(styleKeys).sort().map(k => {
+      const signature = state => JSON.stringify(structuralKeys.map(k => {
         const value = state(k);
         return value === undefined ? null : value;
-      }));
+      }).concat(usesPlayerColor(state('seatedColor'))));
       // a key the preset does not set is reset to null, i.e. back to the widget default
       const presetSignature = signature(k => k in preset ? preset[k] : widget.defaults[k]);
       const update = () => button.classList.toggle('selected', signature(k => widget.get(k)) === presetSignature);
@@ -4556,19 +4578,23 @@ class PropertiesModule extends SidebarModule {
         setDeltaCause(`${getPlayerDetails().playerName} applied the ${label} seat preset to ${widget.id} in editor`);
         for(const key of styleKeys)
           await widget.set(key, key in preset ? preset[key] : null);
-        this.applySeatedColor(widget);
+        this.applySeatColor(widget);
         batchEnd();
       };
     }
   }
 
-  // A seat's color is written when a player sits down, so changing seatedColor
-  // while somebody is already sitting there would otherwise only show up the
-  // next time the seat is taken.
-  applySeatedColor(widget) {
+  // A seat's color is only written when a player sits down or stands up, so
+  // editing seatedColor/emptyColor would otherwise only show up the next time
+  // the seat changes hands. Applies whichever of the two currently applies.
+  applySeatColor(widget) {
     const player = widget.get('player');
-    if(!player)
+    if(!player) {
+      const emptyColor = widget.get('emptyColor');
+      if(emptyColor)
+        this.inputValueUpdated(widget, 'color', emptyColor);
       return;
+    }
     const seatedColor = widget.get('seatedColor');
     const { activePlayers, activeColors } = getPlayerDetails();
     const color = !seatedColor || seatedColor == 'playerColor'
@@ -4590,18 +4616,18 @@ class PropertiesModule extends SidebarModule {
       batchStart();
       setDeltaCause(`${getPlayerDetails().playerName} changed the seated color of ${widget.id} in editor`);
       this.inputValueUpdated(widget, 'seatedColor', value === null ? 'playerColor' : value);
-      this.applySeatedColor(widget);
+      this.applySeatColor(widget);
       batchEnd();
     };
 
     new CheckboxInput(this, widget, 'Use player color', {
-      hint: 'On: the seat shows the seated player\'s color. Off: it always uses the fixed color below.',
+      hint: 'On: the seatedColor property stays at its special value "playerColor", so the seat takes the color of whoever sits down. Off: the fixed color below is used instead.',
       getValue: () => usesPlayerColor(),
       setValue: on => write(on ? null : (readColor() || widget.get('emptyColor') || '#999999')),
       listenTo: [ 'seatedColor' ]
     }).render(this.moduleDOM);
 
-    const colorInput = new ColorInput(this, widget, 'Seat color', {
+    const colorInput = new ColorInput(this, widget, 'Color', {
       hint: 'Fixed color used for this seat while a player is sitting in it.',
       getValue: () => readColor() || widget.get('emptyColor') || '#999999',
       setValue: value => write(value || '#999999'),
@@ -4628,6 +4654,9 @@ class PropertiesModule extends SidebarModule {
     wrap.appendChild(this.renderColorInput(widget, null, 'color', '#222222', 'css', cssClass));
     wrap.appendChild(this.renderNumberInput(widget, null, 'font-size', { step: 1, min: 0 }, '25px', 'css', cssClass));
 
+    // the chips go on their own line so the text/color/size columns line up
+    // with the rows that have no placeholders
+    const chipRow = (options.placeholders || []).length ? div(wrap, 'placeholderChipRow') : null;
     for(const placeholder of options.placeholders || []) {
       const chip = document.createElement('button');
       chip.className = 'placeholderChip';
@@ -4638,7 +4667,7 @@ class PropertiesModule extends SidebarModule {
         input.value = input.value.slice(0, at) + placeholder + input.value.slice(at);
         this.inputValueUpdated(widget, textProperty, input.value);
       };
-      wrap.appendChild(chip);
+      chipRow.appendChild(chip);
     }
 
     this.addPropertyListener(widget, textProperty, w => {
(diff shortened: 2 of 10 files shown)


VTT-AI-Agent (commented):
## TL;DR

Solid, well-scoped rename with a genuinely careful FileUpdater step — I ran the v22 migration over every file in `library/games` and it comes out clean (no seat keeps a string `display`, no `displayEmpty`/`colorEmpty` survives, and the only remaining `PROPERTY display OF …` targets are non-seat widgets). One blocker though: **`??` on `properties.js:4544` breaks the minified production build — the server crashes at startup and the "Production Environment" CI job fails on this PR while passing on `main`.**

## Findings

### 1. Blocker — nullish coalescing crashes the production server

`client/js/editor/sidebar/properties.js:4544`:

```js
const signature = state => JSON.stringify(Array.from(styleKeys).sort().map(k => state(k) ?? null));
```

`server/minify.mjs` uses `@node-minify/uglify-es`, which cannot parse `??`. With `minifyJavascript: true` the server dies on boot:

```
[UnhandledPromiseRejection: … "SyntaxError: Unexpected token: operator (?)"]
curl: (7) Failed to connect to localhost port 8272
```

That is the failing **Production Environment** check on this PR ([run 30588445600](https://github.com/ArnoldSmith86/virtualtabletop/actions/runs/30588445600/job/91025187886)); the same job is green on `ae194e4d` on `main`. This is the only `??` in all of `client/js` — grep confirms it. Please write it out explicitly, e.g. `const v = state(k); return v === undefined ? null : v;`.

### 2. Seats already using the old `.seated { --color: … !important }` hack are not migrated

The previous "Fixed color" preset wrote that css override, and this PR removes all code that reads it. For a game saved with the old editor the result is a split state:

- `renderSeatColorMode` reads only `seatedColor`, so the checkbox reports **"Use player color: on"** even though the seat is in fact pinned to a fixed color;
- no preset button highlights, because the css signature no longer matches either preset;
- worse, if the user then turns the checkbox off and picks a color, `seatedColor` is written but the stylesheet's `--color: … !important` still beats the inline `--color` the engine sets, so the color picker silently does nothing.

Since v22 already rewrites css strings, converting `css['.seated']['--color']` → `seatedColor` (and dropping the now-dead rule) would close this. The blast radius is small — the preset is only one commit old on `main` — but games already saved from the test/beta server are affected.

### 3. `server/pcioimport.mjs:597` still writes the old name

```js
w.type = 'seat';
w.display = 'seatIndex';      // ← not renamed
w.emptyText = 'seatIndex';    // ← renamed
```

This currently works only by accident: the converter emits `_meta.version: 4`, so FileUpdater's v22 step renames `display` → `seatedText` on the way in. The day that migration is squashed away (or someone bumps the emitted version), PCIO imports silently lose their seated text *and* `display: 'seatIndex'` starts life as a truthy show/hide boolean. Write `w.seatedText = 'seatIndex'` directly.

### 4. Switching "Use player color" back on leaves the seat stuck on the fixed color

`properties.js:4568-4573` — `write(null)` sets `seatedColor: 'playerColor'` but never touches `color`, while the fixed-color branch does. So for an already-occupied seat, turning the toggle back on has no visible effect until the player stands up and sits down again. Mirror the `widget.get('player')` branch to restore the seated player's color.

### 5. `write()` should batch its two property updates

The same function issues two `inputValueUpdated` calls (`seatedColor` then `color`) outside a batch. Every other multi-property editor action nearby wraps them — see `renderSeatPresets`' `batchStart()/setDeltaCause/batchEnd()` and the take/remove-seat handlers at `properties.js:4658` and `4670`. As written, picking a fixed color for an occupied seat produces two deltas and two undo steps with no delta cause.

### 6. Remaining migration gaps worth calling out in the PR body / release notes

The library is clean, so these only affect user-saved games — but they fail silently:

- `dynamicProperties: { "display": … }` on a seat: the *key* isn't renamed by `v22SeatProperties`, so it becomes the boolean.
- `ADD` / `CLONE` operations that create a seat with an inline `display` property: `v22SeatRoutine` only inspects `func == 'SET'`.
- `SET property: 'display'` on a seat with a literal value (no `playerName`/`seatIndex` placeholder), e.g. `"Ready"` or `""`. The first keeps the seat visible but loses its label; `SET display ''` now *hides* the seat outright.
- Behavior change beyond the rename: routines that hide a broad selection (`SELECT` → `SET display false`) previously skipped seats thanks to the `type != 'seat'` exception in `Widget.classes()`. They will now hide seats too. That's the point of the PR, but it's the kind of thing that deserves a line in the changelog.
- Nit: the `PROPERTY … OF` regex ends with `'?(?:${ids})'?\b`, which won't match a seat id ending in a non-word character (`seat#`, `seat.`). Rare, but the trailing `\b` is doing nothing useful for the quoted form anyway.

### 7. Test coverage

The only test change is a rename inside `tests/client/property-inputs.test.js`. The riskiest code here — `v22SeatProperties` / `v22SeatRoutine`, which regex over stringified JSON with a seat-id-scoped replacement — has no test at all, and neither does `Seat.seatedColorFor`. There's no existing jest suite under `tests/server/`, so this isn't a convention violation, but a handful of table-driven cases (seat vs non-seat `display`, `PROPERTY display OF <seat>` vs `OF <non-seat>`, the placeholder `SET`) would be cheap insurance for a migration that can't be re-run once a room is saved at v22.

### 8. Docs

The wiki Widgets page documents `display` / `displayEmpty` / `colorEmpty` on seats and will need updating alongside this (and `seatedColor`'s magic `playerColor` value documented).

I did not add the `beta` label: this PR really does change the engine — it alters `Seat`'s default properties and makes `Widget.classes()` apply the generic `display` boolean to seats, changing how existing games render — and on top of that it currently crashes the server under production minification, so it is not safe for player testing.

VTT-AI-Agent:
Thanks — all seven code points are addressed in bb29b18. Rundown:

**1. `??` breaking the production build** — fixed, written out explicitly. Verified by booting the server locally with `minifyJavascript: true` (uglify-es parses it now, `curl` gets a 302 and the bundle contains the new seat code).

**2. Old `.seated { --color: … !important }` seats not migrated** — v22 now converts that rule into `seatedColor` and drops it. Two details beyond the plain move:
- only when the value carries `!important` (without it the inline player color already won, so moving it would *change* behavior);
- a seat that is occupied at migration time also gets its `color` set to the fixed value, otherwise it would visibly fall back to the player color until the player stands up and sits down again.
`${PROPERTY colorEmpty}` (what the preset wrote) resolves to the seat's `emptyColor`; anything else still containing `${…}` is left alone, since `seatedColor` is used verbatim rather than evaluated.

**3. `pcioimport.mjs`** — now writes `w.seatedText = 'seatIndex'` directly.

**4. + 5. Toggling "Use player color" back on / batching** — both write paths (and the style presets, which had the same gap) go through a new `applySeatedColor(widget)` that recomputes an occupied seat's `color` from `seatedColor`, falling back to the seated player's color from `activePlayers`/`activeColors`. The writes are wrapped in `batchStart()`/`setDeltaCause`/`batchEnd()`. Checked in a browser: seated → toggle off → `#999999`, toggle back on → the player's color again; applying the "Fixed color" preset to an occupied seat now turns it red immediately.

**6. Migration gaps** — three of the four are now handled:
- `dynamicProperties.display` on a seat is renamed (key included);
- `${PROPERTY display}` (no `OF`) inside a seat is renamed too;
- `SET`/`GET`/`SELECT` on `display` are renamed whenever the collection provably holds only seats — the updater tracks which collection a `SELECT type: 'seat'` filled (respecting `mode`, and forgetting everything at a `CALL`/`FOREACH`/`IF` that could refill collections). That covers `SET display ''`/`false`/`"Ready"` on seats regardless of the value, which the placeholder heuristic missed.
- `ADD` doesn't exist as an operation, and a `CLONE` that overrides `display` on a seat still can't be resolved statically — that limit plus the `SELECT → SET display false` behavior change are now written up in the PR description.
- The regex nit: the `'?…'?\b` is gone. `evaluateVariables` restricts widget IDs in `PROPERTY … OF` to `[a-zA-Z0-9 _-]` and never accepts quotes, so the pattern now anchors on the closing `}` instead — which also handles IDs containing spaces.

**7. Tests** — added `tests/server/fileupdater.test.js` (15 cases) covering seat vs non-seat `display`, the seat-only names, the `dynamicProperties` key, `PROPERTY display OF <seat>` vs `OF <card>`, `${PROPERTY display}` scoping, the placeholder `SET`, the collection tracking (including that it forgets across an `IF`), and all five css-conversion branches. I re-ran the whole v22 step over all 571 library/tutorial state files afterwards: the only changes are seat text/color properties, no non-seat `display` is touched and no seat ends up hidden. I also drove a browser through a v21 state with the old names + the old css hack, sat down in the seats and ran a `SELECT type: seat` → `SET display false` routine: the seats keep their colors and stay visible with an empty label, exactly as before the rename.

**8. Docs** — I can't edit the wiki, so the PR description now carries a note about the Widgets page (`display`/`displayEmpty`/`colorEmpty` → the new names, plus `seatedColor`'s magic `playerColor` value).

No `beta` label from me either — agreed on the reasoning.

VTT-AI-Agent:
Fixed in c1d9e94 — valid finding. The leading `\s*` in `/\s*!important$/i` was unanchored, so a css value made of a long run of spaces would make the engine restart the match at every one of them (quadratic). Since the value comes straight out of an uploaded game file, that is reachable. Replaced with `trimEnd()` + `endsWith('!important')` + `slice`, which does the same job in linear time; the updater tests still pass unchanged.

VTT-AI-Agent (commented):
## TL;DR
UI/UX review (not a code review). The seat editor comes out of this rename in good shape — "When empty" / "When seated" now map 1:1 onto the properties, and the new `seatedColor` write-through means a fixed color applies instantly to an already-seated player, which is a real improvement. The single most impactful follow-up: do the **same write-through for the empty side** — after applying a style preset an *empty* seat still renders `#999999` grey with white text instead of the "Inactive color" the sidebar promises, which is both wrong and low-contrast.

Reviewed by checking out the branch, running the server locally and driving the seat editor in a browser (screenshots below).

---

## Big layout improvements

**1. The empty seat never picks up `emptyColor`, so the preset previews lie.**
`applySeatedColor()` (properties.js:4560) fixed exactly this problem for the seated case — but there is no symmetric handling for the empty case. `emptyColor` is only copied into `color` when a player *stands up*, so a brand-new or never-used seat keeps `color: #999999`.

Applying the "Background color" preset to an empty seat: sidebar says *Inactive color = red*, thumbnail is blue, actual seat is grey:
sidebar after preset

Empty vs. seated with that preset in play mode — the grey box is what the author gets, and the preset's white text + text-shadow on `#999999` is around 2.4:1 contrast:
empty vs seated

Suggestion: mirror `applySeatedColor()` with an `applyEmptyColor()` — when `player` is empty, write `color = emptyColor` — and call it from the preset `onclick` (properties.js:4560) and from the "Inactive color" input (properties.js:4497). It is the same one-line-of-intent fix and it makes the two sub-sections behave identically, which is the whole promise of this rename.

**2. Any color tweak now silently de-selects the style preset.**
The new signature (properties.js:4548-4556) compares *every* key a preset touches, which now includes `emptyColor`/`seatedColor`. Changing "Inactive color" from the default therefore clears the ✓ badge, even though the seat still uses the Classic layout — the preset row goes from "one is selected" to "none is selected" for a change the user considers a tweak, not a switch.

Before (Classic checked) → after changing Inactive color to green:
classic selected
no preset selected

Suggestion: compare only the *structural* keys (`width`, `height`, `borderRadius`, `css`) for the highlight and leave the color keys out — they are edited right below the preset row and are meant to be customized. (The `css` comparison alone was the old behavior and it read correctly.)

**3. Now that the two colors are separate properties, "Fixed color" should stop making empty and occupied seats look identical.**
`SEAT_PRESET_FIXED` sets `emptyColor` and `seatedColor` to the same `#ff0000` (properties.js:549-553), so a table full of red seats gives a first-time player no visual cue about which seats are free — the only signal is the text. That was hard to avoid with the old `.seated` override hack; with `seatedColor` it is now trivial. Suggestion: keep the strong red for the seated state and give the empty state a clearly weaker variant (e.g. `#ff000055`, or the neutral `#999999` the Classic seat uses), so "taken" reads at a glance.

---

**4. The two "Text" rows don't line up** (properties.js:4498 and 4504). "When empty" gets a full-width input with the color swatch at x≈1362 and the size box at x≈1408; "When seated" is ~145px narrower with its swatch at x≈1218, because the `playerName`/`seatIndex` chips share the row. Two sections that are deliberately parallel end up visually ragged:
seat sidebar
Suggestion: move the placeholder chips onto their own line under the input (or right-align the swatch/size group in both rows) so the columns match.

**5. Label naming is asymmetric.** Under headings that already say "When empty" / "When seated", `emptyColor` is labelled **"Inactive color"** and `seatedColor` **"Seat color"** — a third and fourth vocabulary for the same pair. Since this PR is a rename anyway, "Color" in both sub-sections (mirroring "Text"/"Text"), or "Empty color"/"Seated color" to match the property names people will read in the wiki and the JSON editor, would be much easier to follow.

**6. The "Inactive color" hint is wrong for both presets** — properties.js:4498 says *"The border color while no player sits here."*, but both style presets render `var(--color)` as the **background**. Suggest "The seat's color while nobody sits here."

**7. Stale hint text for `display`** — properties.js:393 still reads *"The seat property whose value is displayed as the seat text."* After this PR `display` is the generic show/hide boolean on seats too; that sentence describes `seatedText`. It currently shows up as a bare `display` dropdown under "Other properties" for any seat that has the property:
other properties

**8. Preset thumbnail label wraps and crowds the box** — "Background color" breaks onto two lines inside the 150×44 preview seat (properties.js:4541, the line this PR touches for `previewColor`). Using a short sample name ("Player") inside the preview and keeping the descriptive label in the tooltip/below would render cleaner.

---

## Readability & understandability (first-time user)

**9. The new capability has no visible affordance.** The headline of this PR is that `display` finally hides a seat like any other widget — but nothing in the seat editor exposes it. The only route is the raw `display` dropdown in "Other properties", which only appears once the property already exists in state. A curated "Visible" toggle (ideally in the shared Basic/Generic group, so it is consistent for every widget type) would let a newcomer discover the thing this PR just enabled. At minimum, make sure the wiki update called out in the description mentions it.

**10. `seatedColor: "playerColor"` is a magic word in a color field.** The "Use player color" toggle handles this nicely in the editor, but anyone who meets the property through the JSON editor, a routine or the wiki sees a *color* property holding the string `playerColor`. Suggest extending the toggle hint to name the property and the special value, e.g. *"On: `seatedColor` stays `playerColor`, so the seat takes the color of whoever sits down. Off: the fixed color below is used."*

**11. Setting `seatedColor` outside the properties panel looks broken.** The editor path calls `applySeatedColor()`, but a `SET seatedColor` from a routine or an edit in the JSON editor leaves an already-occupied seat showing the old color until the next sit/stand. That is an understandable engine constraint, but from a first-time author's point of view the property "does nothing" — worth a sentence in the wiki entry, and it is the same gap as finding 1.

Confirmed working nicely: applying "Fixed color" to a seat someone is already sitting in repaints it immediately, and the "Seat color" row appears/disappears with the toggle exactly as expected:
fixed color applied

VTT-AI-Agent:
Thanks — useful pass. All eleven points looked at, nine of them changed something; everything below is in 00fc303.

**1. Empty seats never picked up `emptyColor`** — fixed, and symmetrically: `applySeatedColor()` became `applySeatColor()`, which writes `emptyColor` into `color` when the seat is empty and the resolved `seatedColor` when it is occupied. It is called from the preset `onclick`, from the "Use player color" toggle *and* from the new empty-side color input, so all three paths behave the same on both sides now.

**2. Color tweaks de-selected the preset** — fixed, but not by dropping the color keys entirely: the two color presets are structurally identical (same `width`/`height`/`borderRadius`/`css`), so a `css`-only comparison would light up both at once. The signature is now the structural keys compared exactly, plus a single boolean for `seatedColor` — "player color" vs "a fixed color" — which is the only thing that actually distinguishes them. Changing either color keeps the ✓.

**3. "Fixed color" made free and taken seats identical** — `SEAT_PRESET_FIXED` now leaves `emptyColor` at the neutral `#999999` the Classic seat uses and keeps `#ff0000` for `seatedColor`. Grey = free, red = taken, at a glance. (I left the "Background color" preset's red empty state alone — with finding 1 fixed it now actually renders, which is what that preset was promising.)

**4. The two "Text" rows didn't line up** — the placeholder chips moved to their own row under the input, indented to the label column. Both "Text" rows now have their color swatch and size box at the same x (measured: 1444 in both).

**5. Asymmetric label vocabulary** — both are just **"Color"** now, under their "When empty" / "When seated" heading, mirroring "Text" / "Text".

**6. Wrong "border" hint** — now *"The seat's color while nobody sits here."*

**7. Stale `display` hint** — rewritten to describe the generic boolean.

**8. Preset thumbnail label wrapped** — the preview seat shows a short `Player` sample and the preset name is rendered below the thumbnail. The size is also forced *after* the preset now, so the 172px-wide presets no longer overflow the 150px preview button.

**9. No affordance for the new capability** — added a curated **"Visible"** checkbox (with hint) to the "Generic" group of the Basic section, so it is there for *every* widget type, not just seats. `display` is excluded from "Other properties" accordingly.

**10. `playerColor` is a magic word** — the toggle hint now names the property and the value: *"On: the seatedColor property stays at its special value "playerColor", so the seat takes the color of whoever sits down. Off: the fixed color below is used instead."*

**11. `SET seatedColor` from a routine looks broken** — no code change here: `color` is a real stored property that the scoreboard and other widgets read, so deriving the rendered color from `seatedColor` behind its back would let the two diverge. Instead the PR description's wiki note now spells out that `seatedColor`/`emptyColor` take effect on the next sit/stand when they are set outside the properties panel, and that the panel writes `color` through for you.

### Verified in a browser

Empty seat with the "Background color" preset and its color then tweaked to green — it repaints immediately *and* the preset stays selected (findings 1 + 2):
empty color write-through

The two sub-sections, now symmetric and aligned, chips on their own line (findings 4, 5, 6):
when empty / when seated

The new "Visible" checkbox and the re-laid-out preset row (findings 8, 9):
visible checkbox

Also re-ran `npm test` (101 pass), the whole TestCafe `editor.js` suite (16 pass), and booted the server with `minifyJavascript: true` again to confirm the production build still parses.

VTT-AI-Agent (commented):
## TL;DR

The earlier production, migration-coverage, and editor issues have been addressed, and CI is green, but the v22 updater is still too aggressive in two places: it can permanently rewrite unrelated user data and can turn generic non-seat `display` writes into `seatedText` writes. The editor also fails to restore player color when an occupied seat belongs to a disconnected player.

## Findings

### 1. High — the migration rewrites arbitrary strings and widget references

`server/fileupdater.mjs:615-646`

`v22RenameSeatOnlyNames()` replaces every word-boundary occurrence of `displayEmpty` / `colorEmpty`, while `v22SeatProperties()` applies it to the JSON serialization of **every string or object-valued property on every widget**. That reaches far beyond property identifiers and expressions: ordinary label text, comments, HTML, asset strings, variable names, and widget references are all rewritten.

For example, an existing state with a widget whose map key is `colorEmpty` and another widget with `parent: "colorEmpty"` keeps the map key unchanged, but the parent value becomes `"emptyColor"`, leaving a dangling parent reference. Likewise, visible text such as `"show displayEmpty here"` is silently changed. Once the room is saved as v22, this data loss is permanent.

Please make this a structured migration: rename the actual property keys, operation `property` fields, and parsed/targeted `${PROPERTY ...}` references, without running the rename over arbitrary string values. A regression test with literal text and a widget ID/reference named `colorEmpty` would pin down the boundary.

### 2. Medium — the placeholder heuristic changes `display` semantics on provably non-seat collections

`server/fileupdater.mjs:741-743`

The fallback renames any `SET property: "display"` whose value merely contains the word `playerName` or `seatIndex`, even when the tracked collection is known to contain non-seats. For example:

```json
[
  { "func": "SELECT", "type": "label" },
  { "func": "SET", "property": "display", "value": "${playerName}" }
]
```

Previously this set the labels' generic visibility; after migration it writes an unused `seatedText` property and no longer controls visibility. `playerName` is a general routine variable, so its presence does not prove that the target is a seat. Please only rename when the target collection is proven seat-only (and leave ambiguous cases unchanged), with a test covering a non-seat selection plus a `playerName` value.

### 3. Medium — “Use player color” cannot restore an inactive occupant's color

`client/js/editor/sidebar/properties.js:4598-4603`

`applySeatColor()` looks up the occupant only in `activePlayers`, which contains connected players. Seats commonly remain occupied after their player disconnects. If an editor applies a fixed color to such a seat and then turns “Use player color” back on, `indexOf(player)` is `-1`, `activeColors[-1]` is `undefined`, and the guarded write is skipped. The seat remains visibly fixed even though `seatedColor` now says `playerColor`, and reconnecting does not call `setPlayer()`, so the mismatch lasts until someone stands up and sits again.

Please resolve the color from the complete player/color mapping (not just active players), or otherwise preserve enough information to restore it, and cover an occupied inactive player in the editor tests.

I did not add the `beta` label because this PR genuinely changes engine behavior in `client/js/widgets/seat.js` and `client/js/widgets/widget.js`, so it is engine-ineligible for beta auto-deployment.

VTT-AI-Agent:
Thanks — all three findings were valid and are fixed in 6122093.

**1. The migration rewriting arbitrary strings** — the blanket `\bcolorEmpty\b` rewrite over the JSON of every property is gone. The rename is now structural and only touches places where a property *name* is expected:

- property keys on the widget,
- the property-name fields of routine operations (`property` on `SET`/`GET`/`SELECT`/`SCORE`/`RESET`, `key` on `SORT`, including a `GET` property path and a `SORT` key list),
- the three maps that hold names rather than values: `inheritFrom` (keeping a `!` negation), `svgReplaces` and a face object's `dynamicProperties`,
- `${PROPERTY <name>}` / `${PROPERTY <name> OF <id>}` references, matched with the identifier syntax `evaluateVariables()` accepts.

Label text, html, css class names, widget IDs and `parent`-style references are never rewritten now. Your exact example is a regression test: a widget with `id: 'colorEmpty'`, another with `parent: 'colorEmpty'`, `"show displayEmpty here"` in `text` and `classes: 'colorEmpty'` all come out unchanged, while the `displayEmpty` *property* on the widget that happens to be called `colorEmpty` is still renamed.

One case the old blanket rename got right only by accident and which the structural one has to handle explicitly: `GET property: 'colorEmpty'` names its variable after the property, so a `${colorEmpty}` further down the routine would break. The updater now pins the old name as an explicit `variable` before renaming. That is exactly what `library/games/righto` and `library/games/Snail trafficking` do — I diffed the old and new updater output over all 571 library/tutorial states and those two files are the only difference between them, both this variable pinning.

**2. The placeholder heuristic on provably non-seat collections** — collection tracking is now tri-state (`true` = provably seats only, `false` = provably no seats, `undefined` = unknown) with proper union/intersection for `mode: add`/`intersect`. The placeholder fallback only fires for `undefined`. Your example (`SELECT type: 'label'` → `SET display '${playerName}'`) is a test and is left alone; a `SELECT` without a type filter stays `undefined`, so the common `SELECT property: 'player', value: '${playerName}'` → `SET display '${playerName}'` seat pattern still migrates (also a test).

While in there I also dropped the `dynamicProperties` *key* rename from the previous round: `dynamicProperties` is a face-object map (face-object property → widget property name), and `display` is a legitimate face-object property, so renaming the key was wrong. The map's *values* are renamed instead, and there is a test for both halves.

**3. Disconnected occupant's color** — `players.js` now also exports the room's full player→color map (`playerColors`, straight from `_meta.players`, which keeps a seated player after they disconnect) and `applySeatColor()` resolves from that instead of `activeColors[activePlayers.indexOf(player)]`. Verified with two browser clients: player sits down, disconnects, then in the editor "Use player color" off → `#999999`, back on → the disconnected player's `#1300f0` again. Unit tests for `applySeatColor` cover the empty seat, the fixed color, the connected player, the disconnected player and an unknown player.

`npm test` is 110 green (12 new), TestCafe `editor.js` 16/16 locally. No `beta` from me either.

VTT-AI-Agent:
Follow-up on CI: the Firefox job that failed on 6122093 was a browser blip — the same commit passed everything on Chrome, and its two Firefox jobs ran on different Firefox builds (152.0 and 153.0) mid-run; a re-run is green.

CodeQL did find a real one in the new code, fixed in 1db6338: `js/polynomial-redos` on the `${PROPERTY ...}` pattern. Spelling both identifiers out with character classes that include spaces (`([a-zA-Z0-9 _-]+?)(?: OF ([a-zA-Z0-9 _-]+))?`) backtracks quadratically on a value made of many ` OF ` repetitions without a closing brace — and that value comes straight out of an uploaded game file. The expression content is now captured in one piece (`\$\{PROPERTY ([^{}]*)\}`) and split on the first ` OF `, which is the same split the engine's lazy match produces; identifiers that are not the plain statically-resolvable form are left alone. Tests cover the split, the unresolvable forms and that 20k ` OF ` repetitions finish instantly.

All 24 checks are green now.

VTT-AI-Agent (commented):
## TL;DR

The earlier findings are addressed and CI is green, but the v22 migration still misclassifies collections after `add`/`intersect`. That can permanently rewrite a non-seat `display` operation to `seatedText`, so I found one remaining migration correctness issue.

## Findings

### 1. Medium — the collection merge truth tables can turn a proven non-seat collection back into “unknown”

`server/fileupdater.mjs:825-830`

The tri-state comments define `false` as “provably no seats,” but `v22Intersection(undefined, false)` returns `undefined`. Intersecting any collection with a selection that is provably seat-free is still provably seat-free, so this result is unsound. The corresponding union helper is also unsound: `v22Union(undefined, false)` and `v22Union(true, false)` both return `false`, even though either result may contain seats.

For example, this existing v21 routine:

```json
[
  { "func": "SELECT", "type": "all" },
  { "func": "SELECT", "type": "label", "mode": "intersect" },
  { "func": "SET", "property": "display", "value": "${playerName}" }
]
```

ends with a collection that can only contain labels. The tracker records `undefined`, however, so the placeholder fallback at line 812 rewrites the last operation to `property: "seatedText"`. Before migration that operation could show previously hidden labels; afterward it writes an unused property and leaves them hidden. Because the state is then saved as v22, the break is persistent.

Please correct the merge truth tables (for union, only `true + true` is definitely seat-only and only `false + false` is definitely seat-free; for the example above, intersecting with `false` must remain seat-free) and add tests that exercise both `mode: "add"` and `mode: "intersect"` before a placeholder-based `SET display`.

I did not add the `beta` label because changes in `client/js/widgets/seat.js` and `client/js/widgets/widget.js` genuinely alter default seat and visibility engine behavior, making this PR engine-ineligible.

VTT-AI-Agent:
Fixed in d862184e. The tri-state merge rules now preserve certainty correctly: union is seat-only only for two seat-only inputs and seat-free only for two seat-free inputs; intersection is seat-free if either input is seat-free and seat-only if either input is seat-only. Added regression coverage for both `add` and `intersect` before a placeholder-based `SET display`, including the reported non-seat intersection case. `npm test -- --runInBand` passes (114 tests).

PR #3081, opened 2026-07-31
Build routines in plain language in the Automations section

VTT-AI-Agent:
Implements the version 1 scope agreed in #3078: a beginner-friendly UI for creating and editing routines, living inside the existing **Automations** section of the Edit Widgets module — alongside, not instead of, the JSON editor.

## The problem

`renderAutomationsSection()` listed five hardcoded routine properties and gave each one a button whose entire behaviour was "open the JSON module" — it didn't even navigate to the routine you clicked. A game builder who can drag widgets and set a colour had no way to author automation without learning JSON.

## What this does

Every routine of the selected widget is now a collapsible section of **step cards written as sentences**, with the editable parts as inline chips:

> **Pick** `card` widgets
> &nbsp;&nbsp;where `cardType` `is` `ace`
> &nbsp;&nbsp;at most `3` of them
> &nbsp;&nbsp;chosen at random
> &nbsp;&nbsp;and call them `aces`

the builder on an existing routine

### The model (`client/js/editor/routineTemplates.js`)

A card's sentence is **composed, not looked up**. Each operation gets:

* a small table of **variants** — the leading verb. Picking a different verb rewrites the parameters *and* the sentence together, so the user never picks "the `relation` parameter", they pick another sentence (`SET` → Set / Increase / Decrease / Multiply / Divide / Switch on or off / Append).
* a list of optional **clauses** — one phrase each, appearing only while its parameter is set and vanishing with it (`SELECT`'s `max` / `random` / `sortBy` / `source` / `collection`).

Version 1 covers the eleven operations agreed in the issue — **SET, SELECT, GET, CALL, MOVE, LABEL, COUNT, FLIP, CLICK, RECALL, SHUFFLE** — which are 77% of all operations in `library/games`.

### Nothing is lost, nothing is misrepresented

An operation no variant matches, **or that carries a parameter the table doesn't know**, renders as a read-only advanced card showing its JSON, still reorderable and deletable, with a jump to the JSON editor. So `IF`, `FOREACH`, `INPUT`, `SET` with `applyVariables`, and free-form expressions all open fine — the module works on real games, not just trivial ones. `// …` string steps render as an editable note.

an unsupported step falls back to an advanced card

### Editing

* **Add / delete / move up / down** per step; add-step menu grouped by intent (Cards and pieces, Picking and counting, Changing things, Reusing routines) rather than alphabetically. No drag-to-reorder in v1, as agreed.
* **Widget chips reuse the existing room picker** (`renderWidgetSelectPopout` / `startWidgetPicker`) — "Pick in the room", type filter, ID search — no new picking code.
* **Value chips** accept a literal or one of the values earlier steps of the *same* routine remember (`COUNT.variable`, `GET.variable`, `CALL.return`, `VAR`, `INPUT` fields, `var x = …` lines), so `${…}` never has to be typed. Unknown names stay silent, as agreed.
* **Live-write**: each change replaces the whole routine array as one delta through `inputValueUpdated()` → `widget.set()`. Undo works for free and an open JSON editor stays in sync. Untouched steps are written back exactly as they were read (deep copy, only the edited step mutated).
* The section now lists **all** `*Routine` properties actually present — including custom ones that `CALL` targets, which were previously invisible — plus an "Add automation" menu with the standard routines and a custom name.

adding an automation from scratch

### Translation

Per the flagged comment on the issue, every user-facing string — operation names, verbs, sentences, clauses, enum values in words ("is more than", "do not run their own click automation") — lives in a single `routineBuilderStrings` table in `routineTemplates.js`, so the whole builder can be translated from one place.

## Not in this version

Deliberately, per the scoping discussion: `IF`/`FOREACH`/`INPUT` editing, `CALL` arguments, drag-to-reorder, per-card validity badges, and a ▶ Test run. Those stay advanced-card / JSON territory for now.

## Files

| | |
|---|---|
| `client/js/editor/routineTemplates.js` | new — the variant/clause tables and the string table (pure data + pure functions) |
| `client/js/editor/routineBuilder.js` | new — turns them into DOM and writes the routine back |
| `client/css/editor/routineBuilder.css` | new — step card styling; sentences wrap in the narrow sidebar |
| `client/js/editor/sidebar/properties.js` | `renderAutomationsSection()` rewritten |
| `server/minify.mjs` | the three new files added to the editor bundle |
| `tests/client/routine-templates.test.js` | new — 49 jest tests |

No widget property, routine operation or parameter was added, renamed or removed, so `validator/` and the `jeCommands` table in `jsonedit.js` need no changes.

## Testing

* `npm test` — 135 jest tests pass, including the new suite: every verb/clause/enum has a translatable string, every sentence placeholder resolves to a field, rendering never mutates the operation, verb switching rewrites parameters coherently, and 36 real-world operation shapes taken from the frequency analysis in #3078 (including legacy `MOVE count: 0` and `SET relation: "=="`) each render as the expected sentence.
* TestCafe `editor.js` (16 tests) and `routines.js` (4 tests) pass locally.
* Driven end-to-end in a real browser: a 12-step `clickRoutine` covering all eleven operations plus an `IF` and a `//` note renders correctly with no console errors; switching a verb, adding a step, adding an automation and opening the room widget picker all write the expected JSON and leave the untouched steps byte-identical.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3081/pr-test (or any other room on that server)

### DIFF
--- a/client/css/editor/routineBuilder.css
+++ b/client/css/editor/routineBuilder.css
@@ -1,0 +1,218 @@
+/* Plain language routine builder in the Automations section of Edit Widgets. */
+
+.tune.editorModule .routineBuilderHeader {
+  display: flex;
+  align-items: center;
+  gap: 4px;
+  margin-bottom: 6px;
+}
+
+.tune.editorModule .routineBuilderProperty {
+  flex: 1;
+  font-family: monospace;
+  font-size: 11px;
+  color: #888;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
+.tune.editorModule .routineBuilderHeader button {
+  padding: 0;
+  min-width: 22px;
+  border: 0;
+  background: transparent;
+  color: #7c7c7c;
+  cursor: pointer;
+}
+
+.tune.editorModule .routineBuilderHeader button:hover {
+  color: #333;
+}
+
+.tune.editorModule .routineStepList {
+  display: flex;
+  flex-direction: column;
+  gap: 6px;
+}
+
+.tune.editorModule .routineEmpty {
+  font-size: 12px;
+  color: #888;
+  padding: 4px 0;
+}
+
+.tune.editorModule .routineStep {
+  border: 1px solid #ddd;
+  border-radius: 4px;
+  background-color: #f9f9f9;
+  padding: 6px;
+  display: flex;
+  flex-direction: column;
+  gap: 4px;
+}
+
+.tune.editorModule .routineStepHeader {
+  display: flex;
+  align-items: center;
+  gap: 4px;
+}
+
+.tune.editorModule .routineStepIndex {
+  min-width: 18px;
+  height: 18px;
+  border-radius: 9px;
+  background: #ddd;
+  color: #555;
+  font-size: 11px;
+  line-height: 18px;
+  text-align: center;
+}
+
+.tune.editorModule .routineStepTitle {
+  flex: 1;
+  font-size: 12px;
+  font-weight: bold;
+  color: #444;
+}
+
+.tune.editorModule .routineStepControls {
+  display: flex;
+  gap: 2px;
+}
+
+.tune.editorModule .routineStepControls button,
+.tune.editorModule .routineClauseControls button {
+  padding: 0;
+  min-width: 20px;
+  border: 0;
+  background: transparent;
+  color: #7c7c7c;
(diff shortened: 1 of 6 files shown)


96LawDawg:
Closed. This was always just a demo. Will be implemented in #3034

Issue #3082, opened 2026-07-31
Per-player / per-seat views of a room (strategy: facing, seat overrides, viewports)


VTT-AI-Agent:
### Summary

Design discussion + proposal for **per-player / per-seat views of a room**: letting each player see the same room laid out, scaled, rotated or oriented differently, *without* duplicating widgets and without the game designer hand-rolling the machinery out of `owner` arrays, `inheritFrom` clones and coordinate-math routines.

This is a strategy/design issue — no code proposed. It pulls together #1041 ("Mirror" widgets), #878 (scaling/scrolling out of sync) and the "upside-down" idea from playingcards.io, and recommends a direction.

---

### Prior art in this repo

**#1041 — "Mirror" widgets.** A widget that renders an exact live copy of another widget and its children, at its own position/scale/rotation. Use case: my own player board stays big in front of me, opponents' boards are visible but small.
The discussion converged on the problems: a mirror needs a shadow copy of every descendant, other players' cursors stay in the original, custom CSS selectors get ambiguous, and "which one is the real one" becomes a permanent question. @ArnoldSmith86 noted the same use case falls out of a per-player override of `x`/`y`/`scale` with **no copy at all** — the real widget is simply somewhere else and smaller for other players.

**#878 — scale/scroll out of sync.** Wanting a wide board where each client can pan/zoom independently. @robartsd proposed `table` + `viewport` objects (widgets *live in* a viewport, each client can translate/scale/rotate its view of a surface). @96LawDawg demonstrated that a per-player scaled board is *possible* today via seats + `inheritFrom` clones + coordinate-fixing routines — and that demo is exactly the evidence that the manual approach is too expensive: it breaks as soon as scale ≠ 1, and adding a widget means touching every clone.

**Common conclusion from both threads:** duplication is the wrong primitive. Whatever we build, there must remain **exactly one real widget**, and the difference between players must be purely a matter of *how it is presented*.

### Prior art outside this repo

**playingcards.io "Extras → Show upside-down".** A per-object checkbox. The object is rendered rotated 180° for players sitting "across the table", so text and card faces read correctly from where you actually sit. It is worth studying not because it is powerful — it is the least powerful possible version of this feature — but because of *how it is exposed*: one checkbox, no coordinates, no seat list, no expressions. The designer never thinks about "who is player 3"; they think "this thing faces the player it belongs to". That ergonomic bar is the one we should aim at for the common cases.

---

### Design principles this should satisfy

[...]

Issue #1041, opened 2021-12-21
"Mirror" widgets


mousewax:
A widget that displays an **_exact_** copy of another widget and everything inside of it, including their states.

Players cannot interact with anything on the Mirror widget aside from the `enlarge` mouse over but the Mirror can have its own position, rotation, and scale (but identical height and width).

This would be useful for games where you have your stuff directly in front of you but still see the tableaus of other players in other spots. 

Card games where left and right matter could also be simulated _without rotating the room or translating x/y coordinates_.

If I put a card in the main holder, all of the mirrors display that same card.
If I flip that card over, all of the mirrors display the now flipped-over card.
If I drag a meeple into that holder, all the mirrors will show a meeple in the same spot.

My player board stays nice and big and my opponents boards only have to be big enough that I can tell what they are doing:
![image](https://user-images.githubusercontent.com/12822213/146853233-5278fa45-8653-4514-ba8c-f9db1cc3a70e.png)

When it's your turn, your player board gets a thick red outline, which means other people will see your smaller player board with that thick red outline so you know whose turn it is if that's part of the game. If the `seat` is in that widget, all f the styling is applied to that mirror seat.

Maybe Tichu?
![image](https://user-images.githubusercontent.com/12822213/146853379-7953adfc-575c-48a1-90d7-e8ecf6c52748.png)

robartsd:
I like the idea, but it seems like it would be difficult to implement (much more difficult to implement than a free camera table would be). We would basically need the mirror to keep a "shadow copy" of each of widget's children. Other player's cursors would also still move in "My play area" rather than the area for the other player. Game designers might need to be responsible for the application of any custom stylesheet rules (I have no idea how the selectors would work). Hover effects would not be mirrored (but hover effects are per client anyway, so that wouldn't matter for your use case).

robartsd:
Your use case does not actually require a "mirror", just separate viewports with each player having the viewports arranged and scaled differently. If the actual widgets were descendants of the viewports, the challenges of synchronizing the "shadow copy" and associated styles goes away.

mousewax:
I guess I don't know how viewports fixes this kind of issue. Is there an issue or PR or discussion somewhere I can look at? I know of them, but know nothing about them really.

robartsd:
> I guess I don't know how viewports fixes this kind of issue. Is there an issue or PR or discussion somewhere I can look at? I know of them, but know nothing about them really.

[...]

Issue #878, opened 2021-11-10
Feature request:Allow scaling and scrolling out of sync with other players


jippless3024:
I want to use the holder as a wide board.
Is it possible to prevent other players from syncing if the board is scaled or scrolled at that point?

Is it possible to have a holder that is displayed in initial coordinates each time I join and does not synchronize movement with other players? (The coordinates of the children in the holder are synchronized)

If that can be achieved, it will function not only as a wide board but also as a controller position adjustment for each player.

96LawDawg:
I don’t know anything about the platform code, but what I do know suggests what you are asking for (disabling synch) is not easily possible. 

But you might be able to accomplish this another way. Each player has there own board visible only to them through a seat widget. Can be scaled however you want. Use multiple copies of the cards or game pieces or whatever with ‘inheritFrom’. The only thing different on each piece would be the scale. So they would appear in the same relative place with the same face or whatever on everyone’s board. Some people would just see a bigger or smaller slice of the board. 

Nothing like that has been done so for but I think it might be possible.

robartsd:
I'd love for there to be a "table" object that is defined by the game designer with each client able to set their own view (scale, translation, rotation) of the table. There could also be a "viewport" object that a game designer could assign widgets to to cause them to be placed in a fixed location on screen regardless of the table (I picture this used for a hand).

I picture the "table" being a rectangle or circle (defaulting to the current 1600x1000). A game designer creating a card game might choose a circular table with each seat having a different default table rotation to make a game where all players sit around a table. Other games might have their tables defined to fit a particular board. The game designer should also be able to indicate the playing surface is unlimited (the size defined would still control the initial view of the table when someone joins the room or the game is loaded).

I'd prefer that the "viewport" object conform to the window of the user rather than use a fixed aspect ratio, but that would make some layouts difficult to achieve.

ArnoldSmith86:
BTW: you might want to consider joining our Discord. 😄

96LawDawg:
@jippless3024 check out https://virtualtabletop.io/LawDawg.  2 seats, each with their own view of the same thing: a 'hand' with a few objects in it.  Each player can independently scale the hand using the buttons in the upper right.  Things that move on player 1's board move a corresponding amount on player 2's board (which is currently set at scale 0.5).  The side of the checker and the value of the spinner are all controlled through 'inheritFrom.'  The hand also is movable for 'scrolling' in higher scales.

[...]

Issue #877, opened 2021-11-10
holder scale and movable child coordinates


jippless3024:
If I put child in a holder made under the following conditions, there will be a gap between the place where I leave child and the place where you store it.
alignChildren:false
scale:!= 1

Check the example
https://virtualtabletop.io/xtvl

robartsd:
Yes, the code mapping room coordinates to local widget coordinates can use some work. I believe there are similar issues with rotated holders.

jippless3024:
As a workaround, I created a routine to address this issue in the json editor
The placement is based on the upper left corner of the widget
However, it is arguable whether this change is made on a routine or as a basic operation of the widget.

Check the example
https://virtualtabletop.io/xtvl

ArnoldSmith86:
I think I have a local branch somewhere that improves the behavior. Unfortunately it is a recursive problem with arbitrary layers of `scale` and `rotation` and the "correct" solution would be total overkill.

I'll try to remember to make it into a draft PR.

robartsd:
I've added a draft PR with recursive coordinate conversion functions. #882

robartsd:
#903 should fix this issue. Please test.

PR #3083, opened 2026-07-31
Per-seat views: facing, seat overrides and a "preview as seat" switcher

VTT-AI-Agent:
Implements **Layer 1**, the **editor "preview as seat"** switcher and **Layer 2** from #3082, following the maintainer's guidance in the issue comment. Layer 3 (surfaces/viewports) is explicitly out of scope.

The point of the feature: the same room can be laid out, scaled, rotated and oriented differently for each player **without duplicating a single widget**. There stays exactly one real widget; only its *presentation* differs per client.

## The invariant

Everything here is a rendering layer. Overrides are resolved on each client from synced state and applied through a new `Widget.getView(property)`; `get()` — which routines, `${PROPERTY …}`, the delta stream, undo and save files all use — is untouched. `_absoluteX`/`_absoluteY`/`_centerAbsolute*`/`_localOriginAbsolute*` now measure explicitly in the room's *shared* coordinate system (`getElementTransform()` takes an optional transform override), so a widget inside a per-seat-scaled parent still reports the same coordinates on every client.

Nothing is opt-in for players and nothing changes for existing games: all new properties default to "no per-seat view", and the whole mechanism stays switched off in a client until some widget in the room actually uses it. No FileUpdater migration and no legacy mode is needed.

## Layer 1 — `facing`

A single widget property, the playingcards.io "show upside-down" ergonomics:

- `"facing": "owner"` — reads upright for the seat(s) the widget belongs to, everyone else sees it at its stored rotation
- `"facing": "viewer"` — reads upright for whoever is looking at it

"Belongs to" means the seats matching `owner`, `linkedToSeat` or `onlyVisibleForSeat`. How far to turn comes from the seat's new `viewRotation` property, which defaults to the seat widget's own `rotation` — so a designer who already rotated the far-side seat by 180° gets this for free.

In the editor this is one dropdown ("Reads upright for") in the widget's existing seat section.

## Layer 2 — `seatOverrides` / `viewOverrides`

`seatOverrides` on a widget maps a **seat selector** to a set of overrides; `viewOverrides` on a **seat** maps widget ids to overrides for the case where a whole layout differs for one player. Both shapes exist because the issue asked for both; the widget side wins where they overlap, since a widget has the final say about itself.

```json
"seatOverrides": {
  "others": { "scale": 0.4, "x": 900 },
  "owner":  { "scale": 1 },
  "seat3":  { "rotation": 180 }
}
```

Selectors, in ascending precedence: `all` → `others` → `+n`/`-n` (viewer sits n seats after/before the owner, in `index` order, wrapping) → `owner` → an exact seat id. A viewer **without a seat** gets the plain un-overridden layout — the defined fallback the maintainer asked for — unless the widget spells out a `noSeat` entry.

The overridable set is a **closed whitelist**: `x`, `y`, `rotation`, `scale`, `width`, `height`, `borderRadius`, `css`, `classes`, `display`, `movable`, `clickable`. The validator rejects anything else, so a `seatOverrides` block can never smuggle in game state.

Two consequences worth calling out:

- **Children follow automatically.** Widget parents are real DOM parents, so a scaled-down player area scales its contents, and drag/drop hit-testing keeps running in rendered space — dropping a card onto an opponent's 40 %-size area lands where it looks like it lands.
- **Counter-rotation is opt-out**, per the maintainer's answer. A per-seat `rotation` override re-orients a container, and its descendants are counter-rotated so their text and art keep facing the way they did; set `"counterRotate": false` on a descendant to make it spin with the container. `facing` deliberately does *not* trigger this — turning the content is exactly what it is for.
- A widget whose **`x`/`y`** this client draws elsewhere is not draggable for that client, because the drag would write the personal coordinates back into the shared state. Rotation and scale overrides do not restrict dragging (they still produce a legitimate room coordinate), and children of a moved widget drag normally.

## Editor: "preview as seat"

A toolbar button (next to the other view toggles) lists the seats and renders the room exactly as that seat sees it — including the seat's private hand and `owner`-hidden widgets, because the preview also swaps the identity that per-player visibility is judged against. It is read-only: everything that *writes* keeps using the real player name, so a preview can never touch the game. The room gets a dashed outline while a preview is active, and it resets when the editor closes.

Without this the feature is untestable alone — every designer would otherwise open four browser windows.

## Screenshots

Same room, same widgets, two seats. The far-side player's mat reads upright for them (`facing`), the board is turned round for them with its labels still readable (per-seat `rotation` + counter-rotation), and the shared tray is moved and shrunk (`seatOverrides`):

| as `seatBottom` | as `seatTop` |
| --- | --- |
| as seatBottom | as seatTop |

The editor switcher: preview as seat

Playable demo file (the game in the screenshots): [per-seat-views-demo.vtt](https://agent.virtualtabletop.io/reports/pr/1532587720080363710/per-seat-views-demo.vtt)

## Also in this diff

- **Validator**: `facing`, `seatOverrides`, `counterRotate`, `viewRotation`, `viewOverrides` with validators that check the selector keys and enforce the property whitelist.
- **JSON editor**: the new properties get insert buttons automatically from the widget defaults; `facing`, `seatOverrides` and `viewOverrides` insert a useful starting value instead of `null`.
- **Toolbar tooltips get `pointer-events: none`.** They hang below the toolbar and could swallow clicks on whatever sits underneath — adding one button to the toolbar was enough to make the deck-editor "Card view" TestCafe test fail on an obstructed click. Tooltips should never intercept pointer events.

## Testing

- `npm test`: 102 passed, including 16 new jest tests in `tests/client/seat-view.test.js` covering selector precedence, seat offsets and wrap-around, the spectator fallback, the property whitelist, widget-vs-seat-side precedence, `facing`/`viewRotation`, counter-rotation (including grandchildren and the `counterRotate: false` opt-out), drag suppression, the preview identity, and that `get()`/`state` are never touched.
- Full TestCafe suite locally (Chromium headless): editor 16/16, routines 4/4, compute 5/5, public library 18/18 — all state hashes unchanged.
- Manual two-browser check of the demo game above, plus the editor preview.

## Known limitations / follow-ups

- **Remote cursors are still sent in room coordinates**, so a cursor over an area another player sees scaled or moved shows at the shared position, not the personal one. The `mouse` message already carries the widget it is over, so this can be transformed later; it needs its own change.
- The issue also suggests showing the overrides as **a small table on the widget** in the properties sidebar. This PR surfaces `facing` and `viewRotation` as proper editor controls and leaves the override maps to the JSON editor.
- Line widgets anchor to the shared geometry, so a line attached to a per-seat-moved widget draws at its shared position.

Creator-facing behaviour is new, so the wiki (`Widgets`, `Edit-Mode`) will want entries for `facing`, `seatOverrides`, `counterRotate`, `viewRotation` and `viewOverrides`.

Related: #1041, #878, #877.

- **Dragged widgets are not yet transformed for every viewer.** The player dragging sees the piece under their cursor and drops it into the holder they see, but the temporary shared root position can take a misleading path in another seat's view. A client-only drag presentation (and matching cursor treatment) needs a follow-up.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3083/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

The overall rendering/state separation is promising, but two advertised override behaviors are currently inconsistent. Counter-rotation can become stale after a routine changes shared rotation, and `clickable: true` looks enabled without actually making a base-non-clickable widget clickable.

*Automated review 1 of 4 (code) — by `gpt-5.6-sol`*

## Findings

1. **High — Recompute counter-rotation when the shared rotation changes** (`client/js/widgets/widget.js:33`, `client/js/widgets/widget.js:634`)

   `seatViewOwnRotation` is calculated as `override.rotation - get('rotation')`, but `rotation` is not in `seatViewRelevantProperties`, so a normal rotation delta does not schedule a seat-view refresh. For example, with a parent whose shared rotation is `0` and whose seat override is `180`, a child is correctly counter-rotated by `-180`. If a routine later sets the parent's shared rotation to `90`, the parent still renders at the overridden `180`, but the child remains at `-180` instead of being recomputed to `-90`; its absolute rendered orientation is now `0` even though the shared orientation is `90`. Adding `rotation` to the refresh triggers (with a regression test that changes it after the initial refresh) would keep the presentation-only invariant intact.

2. **Medium — A `clickable: true` override cannot enable clicking** (`client/js/mousehandling.js:123`, `client/js/widgets/widget.js:562`)

   The mouse handler only blocks the `true → false` override case and then calls `widget.click()`/`doubleClick()`. Those methods still gate on the shared `get('clickable')`, so a widget with shared `clickable: false` and a seat override of `clickable: true` gets the clickable class and reaches the handler, but its click routine never runs. Since `clickable` is exposed as a general overridable property, either the user-input path needs to honor `getView('clickable')` in both directions (while routine-triggered clicks keep shared semantics), or the override/validator should be restricted to disabling clicks only.

I did not add the `beta` label because this PR changes core widget rendering and event-handling behavior and is therefore engine-ineligible.

VTT-AI-Agent:
Addressed in faa68d0f.

- Shared `rotation` now triggers a seat-view refresh, with a unit regression test that changes the parent rotation after the initial view is resolved.
- Pointer input now gates on `getView('clickable')` in both directions and only then uses `ignoreClickable` to invoke the widget. Routine-triggered `CLICK` calls continue to use the shared `clickable` value. A TestCafe regression covers a `clickable: true` view override on a shared non-clickable widget.

Verified with `npm test` (103 tests) and the focused Chromium TestCafe regression.

VTT-AI-Agent:
## Checkers use case

> How would I make the game Checkers from the public library work? I can flip the entire board, but how do the pieces appear mirrored?
>
> (Source: Discord message from "lawdawg96" — https://discord.com/channels/770758631146782780/1532587720080363710/1532609140461932596)

This is the intended use of the per-seat rotation override. A game author can first save a personal copy of Checkers (the public-library file itself remains untouched), put its 64 square holders under one `board` parent, then give that parent a view for the far seat:

```json
"seatOverrides": {
  "topSeat": { "rotation": 180 }
}
```

The squares are DOM children of `board`, and the pieces are already children of their squares, so their *positions* rotate with the board: each player has their own colour at the near edge. Descendants are counter-rotated by default, keeping the crowned-piece glyph upright; use `"counterRotate": false` on the relevant descendants if the art should turn with the board instead. This is a rotation, not a horizontal reflection, which is the correct mapping for opposite Checkers players.

96LawDawg:
Yes.  The board is no problem.  But picking up a piece and moving it, the piece looks like it shoots across the board to a holder on the other side (where the holder it falls into has been rotated 180) for the top seat. How is that problem to be fixed?  It isn't just the board, but the pieces as they move that have to be transformed.

VTT-AI-Agent:
You are right — my previous answer covered the static board but skipped the transient drag state.

The current code converts the dragger's pointer through the **rendered** transforms, so the top player keeps the piece under their cursor and drops it into the holder they see. The final shared parent is therefore correct. But while it is being dragged, the engine temporarily detaches the piece from its holder and sends its root `x`/`y` to every client. That interim position has no per-seat transform, so another view can indeed show the piece taking an implausible route across the board.

The proper solution is a client-only drag presentation: retain the dragged piece's source-view transform while it is detached, convert the pointer to the shared coordinate only to write the eventual state, and independently render the temporary position in each viewer's board transform. That needs to cover the drag shadow and remote cursors as well. It is not implemented by this PR, so Checkers is not yet a complete playable demonstration of per-seat views; I have added this to the PR's limitations rather than present the static recipe as a full solution.

96LawDawg:
Closed for #3084

PR #3086, opened 2026-08-02
Replace the deprecated node-minify stack with terser, clean-css and html-minifier-terser

VTT-AI-Agent:
## Goal

The minifier depended on packages that are deprecated or unmaintained:

- `@node-minify/uglify-es` → **`uglify-es`**, which npm marks deprecated (*"support for ECMAScript is superseded by `uglify-js` as of v3.13.0"*) and which has not been released since 2017.
- `@node-minify/html-minifier` → **`html-minifier`**, unmaintained since 2019.

## What changed

`server/minify.mjs` now calls the modern minifiers directly instead of going through the `@node-minify` wrapper layer:

| before | after |
| --- | --- |
| `@node-minify/core` + `@node-minify/no-compress` | — (the wrapper only added `glob`/`mkdirp` we never used) |
| `@node-minify/uglify-es` → `uglify-es` | **`terser`** (the maintained fork of `uglify-es`) |
| `@node-minify/html-minifier` → `html-minifier` | **`html-minifier-terser`** |
| `@node-minify/clean-css` → `clean-css` | **`clean-css`** (already current, just used directly) |

Five dependencies become three, and `npm install` no longer prints a deprecation warning for a production dependency.

**Concatenation is untouched** — the CSS/JS file lists, the `removeImportStatements` step and the `.join('\n')` are exactly as before, as are the `/*** CSS ***/` / `//*** JS ***//` placeholder replacements.

Two details worth reviewing:

- `html-minifier-terser` enables no options by default, whereas `@node-minify/html-minifier` injected a default set. That set is now spelled out in `htmlMinifyOptions()` (`removeCDATASectionsFromCDATA` / `removeCommentsFromCDATA` are dropped — those options no longer exist in `html-minifier-terser`).
- `minifyJS` is now tied to the `minifyJavascript` config. The old inline-JS pass used uglify-js, which could not parse our client code and therefore left it alone; terser *can* parse it, so without this the client JS would be minified even with `"minifyJavascript": false`, which would take readable client source away from developers.

## Verification

Output was captured before and after the change for both config modes:

| build | before | after |
| --- | --- | --- |
| `room.html`, `minifyJavascript: false` | 725713 B | 725812 B |
| `room.html`, `minifyJavascript: true` | 480764 B | 481034 B |
| editor JS, `minifyJavascript: false` | 1085186 B | 1085186 B (byte-identical) |
| editor JS, `minifyJavascript: true` | 672591 B | 671687 B |

The only non-JS difference is cosmetic: `html-minifier-terser` bundles clean-css 5 rather than clean-css 4, so a couple of values keep their leading zero (`transition-delay:0.4s` instead of `.4s`).

- `npm test` — 86 passed (88 with the test added below)
- TestCafe against a local server with **`minifyJavascript: true`** (i.e. running the terser-minified client): `routines.js`, `editor.js`, `compute-1.js`, `publiclibrary-1.js` — 26 passed

## Review follow-up

- `compressCSS` now logs clean-css `errors` and `warnings` instead of only reading `.styles`. clean-css does not throw on broken input, it drops the offending declaration — so without this a typo in one of the concatenated CSS files would ship a client missing a rule, with nothing in the server log. The stylesheets are handed to clean-css as separate named sources rather than one concatenated string (identical output), so a problem is reported as `client/css/widgets/timer.css:34:15` instead of a line number in a string that exists nowhere on disk.
- That was already happening: `--wcImageOH: url();` in `client/css/widgets/button.css` is reported as `Broken URL 'url()'. Ignoring.` and dropped, so the minified client never had the declaration. It is now `none`, which is what both the dropped and the literal `url()` variant already resolved to (an undefined custom property makes `background-image: var(--wcImageOH)` invalid at computed-value time → `none`). No visual change; the build is now warning-free.
  - One production edge case does change, for completeness: because the declaration used to be dropped, custom-property inheritance meant a `.button` in a minified build picked up an ancestor's `--wcImageOH`; now it always resets to `none`, exactly as the unminified dev build has always behaved. Nothing in `client/` or `library/` sets `--wcImageOH`, so this makes prod match dev rather than changing any existing game.
- The `minifyJavascript` flag is read through one helper in both passes, so an environment override (`Config.get` returns `MINIFYJAVASCRIPT=false` as the truthy string `'false'`) no longer silently minifies the client.
- `tests/server/minify.test.js` locks in the behaviour this PR introduces: with the flag off, the inline client JS and the editor JS must stay readable.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3086/pr-test (or any other room on that server)

### DIFF
--- a/client/css/widgets/button.css
+++ b/client/css/widgets/button.css
@@ -15,7 +15,7 @@
   --wcBorderOH: var(--wcMain);
   --wcMainOH: var(--wcBorder);
   --wcFontOH: var(--wcFont);
-  --wcImageOH: url();
+  --wcImageOH: none;
 }
 
 .button:hover {
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,17 +9,15 @@
       "version": "0.5.0",
       "license": "GPL-3.0",
       "dependencies": {
-        "@node-minify/clean-css": "^9.0.1",
-        "@node-minify/core": "^9.0.2",
-        "@node-minify/html-minifier": "^9.0.1",
-        "@node-minify/no-compress": "^9.0.1",
-        "@node-minify/uglify-es": "^9.0.1",
         "bson": "^6.10.4",
+        "clean-css": "^5.3.3",
         "crc-32": "^1.2.2",
         "dompurify": "^3.3.0",
         "express": "^5.1.0",
+        "html-minifier-terser": "^7.2.0",
         "jszip": "^3.10.1",
         "node-fetch": "^3.3.2",
+        "terser": "^5.49.0",
         "ws": "^8.18.3"
       },
       "devDependencies": {
@@ -2313,6 +2311,7 @@
       "version": "8.0.2",
       "resolved": "https://registry.npmjs.org/@isaacs/cliui/-/cliui-8.0.2.tgz",
       "integrity": "sha512-O8jcjabXaleOG9DQ0+ARXWZBTfnP4WNAqzuiJK7ll44AmxGKv/J2M4TPjxjY3znBCfvBXFzucm1twdyFybFqEA==",
+      "dev": true,
       "dependencies": {
         "string-width": "^5.1.2",
         "string-width-cjs": "npm:string-width@^4.2.0",
@@ -2329,6 +2328,7 @@
       "version": "6.1.0",
       "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-6.1.0.tgz",
       "integrity": "sha512-7HSX4QQb4CspciLpVFwyRe79O3xsIZDDLER21kERQ71oaPodF8jL725AgJMFAYbooIqolJoRLuM81SpeUkpkvA==",
+      "dev": true,
       "engines": {
         "node": ">=12"
       },
@@ -2340,6 +2340,7 @@
       "version": "6.2.1",
       "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-6.2.1.tgz",
       "integrity": "sha512-bN798gFfQX+viw3R7yrGWRqnrN2oRkEkUjjl4JNn4E8GxxbjtG3FbrEIIY3l8/hrwUwIeCZvi4QuOTP4MErVug==",
+      "dev": true,
       "engines": {
         "node": ">=12"
       },
@@ -2350,12 +2351,14 @@
     "node_modules/@isaacs/cliui/node_modules/emoji-regex": {
       "version": "9.2.2",
       "resolved": "https://registry.npmjs.org/emoji-regex/-/emoji-regex-9.2.2.tgz",
-      "integrity": "sha512-L18DaJsXSUk2+42pv8mLs5jJT2hqFkFE4j21wOmgbUqsZ2hL72NsUU785g9RXgo3s0ZNgVl42TiHp3ZtOv/Vyg=="
+      "integrity": "sha512-L18DaJsXSUk2+42pv8mLs5jJT2hqFkFE4j21wOmgbUqsZ2hL72NsUU785g9RXgo3s0ZNgVl42TiHp3ZtOv/Vyg==",
+      "dev": true
     },
     "node_modules/@isaacs/cliui/node_modules/string-width": {
       "version": "5.1.2",
       "resolved": "https://registry.npmjs.org/string-width/-/string-width-5.1.2.tgz",
       "integrity": "sha512-HnLOCR3vjcY8beoNLtcjZ5/nxn2afmME6lhrDrebokqMap+XbeW8n9TXpPDOqdGK5qcI3oT0GKTW6wC7EMiVqA==",
+      "dev": true,
       "dependencies": {
         "eastasianwidth": "^0.2.0",
         "emoji-regex": "^9.2.2",
@@ -2372,6 +2375,7 @@
       "version": "7.1.0",
       "resolved": "https://registry.npmjs.org/strip-ansi/-/strip-ansi-7.1.0.tgz",
       "integrity": "sha512-iq6eVVI64nQQTRYq2KtEg2d2uU7LElhTJwsH4YzIHZshxlgZms/wIc4VoDQTlG/IvVIrBKG06CrZnp0qv7hkcQ==",
+      "dev": true,
       "dependencies": {
         "ansi-regex": "^6.0.1"
       },
@@ -2386,6 +2390,7 @@
       "version": "8.1.0",
       "resolved": "https://registry.npmjs.org/wrap-ansi/-/wrap-ansi-8.1.0.tgz",
       "integrity": "sha512-si7QWI6zUMq56bESFvagtmzMdGOtoxfR+Sez11Mobfc7tm+VkUckk9bW2UeffTGVUbOksxmSw0AA2gs8g71NCQ==",
+      "dev": true,
       "dependencies": {
         "ansi-styles": "^6.1.0",
         "string-width": "^5.0.1",
@@ -2862,7 +2867,6 @@
       "version": "0.3.13",
       "resolved": "https://registry.npmjs.org/@jridgewell/gen-mapping/-/gen-mapping-0.3.13.tgz",
       "integrity": "sha512-2kkt/7niJ6MgEPxF0bYdQ6etZaA+fQvDcLKckhy1yIQOzaoKjBBjSj63/aLVjYE3qhRt5dvM+uUyfCg6UKCBbA==",
-      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@jridgewell/sourcemap-codec": "^1.5.0",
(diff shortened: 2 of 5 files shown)


VTT-AI-Agent:
CI is green — the one red job (\`publiclibrary-2.js\` / Firefox, `Public library: Reversi`) was a flake and passed on re-run.

I checked it rather than just retrying: I built `client/room.html` from `main` (via `npm ci` so the lockfile resolution matches CI) and from this branch, and diffed them. With the CI config (`"minifyJavascript": false`) the inline JS is byte-identical in both, and the only differences are 100 cosmetic CSS normalisations caused by the second minification pass now going through clean-css 5 instead of clean-css 4 — e.g. `transform-origin:50% 0` → `50% 0%`, `var(--opacity,1)` → `var(--opacity, 1)`, `url(i/fonts/Roboto-Regular.ttf)` → `url("i/fonts/Roboto-Regular.ttf")`, `:nth-child(1)` → `:first-child`. All semantically identical, none of it can move a Reversi piece.

One thing that turned up while comparing, worth knowing: `@node-minify/html-minifier@9.0.2` — inside the `^9.0.1` range `main` already allows — quietly swapped `html-minifier` for `html-minifier-next`, which minifies inline JS with terser instead of uglify-js. On a fresh install `main` therefore already produces a fully minified client even with `"minifyJavascript": false`. That is exactly the case the explicit `minifyJS` flag in this PR guards against.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-scoped dependency modernisation — I reproduced the before/after builds locally and the non-JS output is byte-for-byte equivalent apart from the documented clean-css 5 leading-zero cosmetics, and it drops 6 prod advisories (incl. the high-severity `html-minifier` one). Only real nit: `compressCSS` now reads `.styles` without ever looking at `CleanCSS`'s `errors`/`warnings`, so a malformed CSS rule would silently vanish from every client (pre-existing, but this is the moment to fix it).

## What I verified locally

Built `minifyHTML()` on `main` and on this branch, both config modes, and diffed:

| build | main | PR |
| --- | --- | --- |
| room.html, `minifyJavascript:false` | 725713 | 725812 |
| room.html, `minifyJavascript:true` | 480764 | 481034 |
| editor JS, false / true | 1085186 / 672591 | 1085186 (identical) / 671687 |

Matches the PR description. More usefully:

- **The HTML skeleton is byte-identical** in both modes after stripping `<style>`/`<script>` bodies — so the hand-transcribed `htmlMinifyOptions()` really does reproduce `@node-minify/html-minifier`'s `defaultOptions` (I diffed against `node_modules/@node-minify/html-minifier/dist/index.js`: same 14 options, only the two now-nonexistent `*FromCDATA` ones dropped, `conservativeCollapse` still merged in). The riskiest options here (`removeAttributeQuotes`, `removeOptionalTags`) provably change nothing.
- **The inline `<script>` is byte-identical with `minifyJavascript:false`** — the `minifyJS: !!Config.get('minifyJavascript')` gate does what the description claims; developers keep readable client source.
- `npm test` 86 passed; server boots on the terser path and a room loads in headless Chromium with **zero console/page errors**; the minified editor JS re-parses cleanly.
- Build time is unchanged (`true`: 6.3s main vs 6.5s PR), so the double JS pass noted below costs nothing measurable.
- `npm audit --omit=dev`: **11 vulns (8 high) → 5 (2 high)**. Gone: `html-minifier` (high), plus `glob`/`minimatch`/`brace-expansion` that only came in via the `@node-minify` wrapper. Remainder is express/ws/dompurify, unrelated.

## Findings

1. **`server/minify.mjs:186` — clean-css errors are discarded.** `new CleanCSS().minify(combinedCSSContent).styles` throws nothing on bad input; clean-css drops the offending construct and reports it in `.errors` / `.warnings`. A typo in one of the ~25 concatenated CSS files would therefore ship a client that's silently missing a rule, with no server log. The old wrapper had the same blind spot, but the direct call makes it a one-liner to fix, e.g.:
   ```js
   const result = new CleanCSS().minify(combinedCSSContent);
   if(result.errors.length)
     console.error(result.errors);
   return result.styles;
   ```
   The other two minifiers do surface problems (terser rejects; html-minifier-terser catches JS errors and falls back to the original text via `options.log`), so CSS is the only silent path.

2. **Minor: the inline script is now minified twice when `minifyJavascript` is true.** `compressJS()` terser-minifies the concatenation, then `htmlMinify(..., {minifyJS:true})` runs terser over the whole `<script type="module">` again (that second pass is what now also compresses the injected `const config = {...}` — visible in the output diff). It's correct and free in wall-clock, so no action needed, but it's worth a comment so the redundancy isn't mistaken for a bug later.

3. **Minor: env-var override still yields a string.** `Config.get()` returns `process.env.MINIFYJAVASCRIPT` verbatim, so `MINIFYJAVASCRIPT=false` is the truthy string `"false"` and `!!` doesn't help. Same as before for `compressJS`, but the html pass is newly coupled to the same flag, so a `false` env override now silently minifies the inline JS too. Pre-existing config quirk rather than something this PR must fix — just flagging that the blast radius grew slightly. (`production-environment.yml` seds `config.json`, so CI is unaffected.)

4. **Test coverage:** the behaviour this PR deliberately introduces — "`minifyJavascript:false` must leave client JS readable" — is the one thing no test asserts, and it's exactly what would silently regress if someone later drops the `minifyJS` gate. A tiny jest case asserting `compressJS`-style output still contains e.g. `function ` / original identifiers under the false config would lock it in cheaply.

Nothing here blocks; 1 is the only one I'd actually want changed.

I added the `beta` label: the diff touches only `server/minify.mjs` and the dependency manifests (no engine files at all — no widget, `compute.js` or `serverstate.js` change), the built client is output-equivalent apart from cosmetic CSS number formatting and boots without errors, so it is safe and genuinely useful to exercise on beta where the minified build path actually runs.

VTT-AI-Agent:
Thanks — all four addressed in 2cdcccb0.

**1. clean-css errors discarded — fixed, and it turned out not to be hypothetical.**
`compressCSS` now logs both `errors` and `warnings` through `Logging.log`. Logging the warnings as well is what made the existing case visible:

```
warnings: ["Broken URL 'url()' at 4199:15. Ignoring."]
```

That is `--wcImageOH: url();` in `client/css/widgets/button.css`, and clean-css really does drop it:

```js
new CleanCSS().minify('.button{--wcMainOH:red;--wcImageOH: url();}').styles
// → ".button{--wcMainOH:red}"
```

So the production client has been shipping `.button` without its `--wcImageOH` reset. No visual difference — `background-image: var(--wcImageOH)` with an undefined custom property is invalid at computed-value time and resolves to `none`, and so does the literal `url()` in the unminified build — but the declaration was silently vanishing, which is exactly the failure mode you described. Changed it to `--wcImageOH: none`, which is what both variants already resolved to, and the build is now warning-free (so the new logging stays quiet unless something is actually wrong).

**2. Double JS minification — comment added**, above `htmlMinifyOptions()`, noting that the second terser pass is redundant for the bundle but is what compresses the `const config = {...}` injected after `compressJS` has run.

**3. Env-var truthiness — fixed.** Both call sites now read the flag through one helper:

```js
function minifyJavascript() {
  const value = Config.get('minifyJavascript');
  return !!value && value !== 'false';
}
```

Kept `!!value` first so a missing/`false` config key behaves exactly as before; only the `'false'` string case changes, and it changes to what the person setting it obviously meant. Deliberately local to `minify.mjs` rather than a change to `Config.get` semantics, since every other consumer would be affected by that.

**4. Test added** — `tests/server/minify.test.js` builds `minifyHTML()` with `MINIFYJAVASCRIPT=false` and asserts the inline `<script type=module>` and the editor JS still contain multi-line function bodies and blank lines. It doubles as coverage for 3, since the env override is the string form. I checked it actually bites: hardcoding `minifyJS: true` in `htmlMinifyOptions()` fails it, dropping the gate in `compressJS` fails the editor case. It runs in ~2s (the `false` path skips terser), and it sits in the default jsdom environment because the shared `setupFilesAfterEnv` imports `client/js/connection.js`, which needs `window`.

Verification: `npm test` 88 passed (86 + the 2 new), `tests/testcafe/editor.js` 16 passed in headless Chromium, and the built client still contains `--wcImageOH:none` with no clean-css warnings emitted.

VTT-AI-Agent (commented):
## TL;DR

This is a **UI/UX review** (layout, visual design, first-time readability — not a code review). This PR is build tooling, so its UI surface is a single CSS line (`--wcImageOH: url()` → `none`). I ran the branch with `"minifyJavascript": true` on a local server and clicked through the shelf, the game-details overlay, a real game table and edit mode: **everything renders identically to today's client, and I found no layout regression.** The only thing worth a second look is a behaviour nuance of the `--wcImageOH` reset that is invisible in dev but real in production (details below).

Verified visually against the terser/clean-css-5/html-minifier-terser output:

| screen | screenshot |
| --- | --- |
| Game shelf + public library | [02-room.png](https://agent.virtualtabletop.io/reports/ui-review/1533426693371334696/02-room.png) |
| Game details overlay | [04-game.png](https://agent.virtualtabletop.io/reports/ui-review/1533426693371334696/04-game.png) |
| Edit-mode toolbars | [05-editor.png](https://agent.virtualtabletop.io/reports/ui-review/1533426693371334696/05-editor.png) |
| A loaded game (2036 Daimyo) | [06-table.png](https://agent.virtualtabletop.io/reports/ui-review/1533426693371334696/06-table.png) |

No console/page errors, no missing icons, no collapsed or clipped elements, toolbar dividers and icon spacing unchanged, the details overlay's two-column layout and link styling intact.

---

## Big layout improvements

Nothing to propose — the PR does not restructure any screen, and it should not. The interesting UI question here is only "does the new minifier chain change any pixels", and as far as I can tell it does not.

Two checks that were worth doing because they *could* have changed pixels, both clean:

- **`removeAttributeQuotes` / `removeOptionalTags` / `collapseInlineTagWhitespace` in `htmlMinifyOptions()`** — these are aggressive options that can visibly break inline-text spacing or sibling/child selectors. I confirmed they are not new: `@node-minify/html-minifier@9.0.1`'s bundled defaults already set exactly this list, so `htmlMinifyOptions()` is a faithful transcription rather than a tightening of the pipeline. Good; worth keeping the comment above it that says so.
- **clean-css 4 → 5 for the inline `minifyCSS` pass** — the PR notes the leading-zero difference (`0.4s` vs `.4s`), which is byte-level only. I saw no visual difference in the rendered client.

- **`client/css/widgets/button.css:18` — the `url()` → `none` fix is correct, but it does change production behaviour in one edge case worth a line in the PR body.** Custom properties inherit. Before this PR the declaration was *dropped by clean-css* in minified builds, so `--wcImageOH` on a `.button` fell back to whatever an ancestor had set; now `.button` always resets it to `none`, blocking inheritance. Dev builds already behaved that way (the literal `url()` also blocked it), so this makes prod match dev — but a game author who set `--wcImageOH` on a parent holder and only ever tested on the live (minified) site would see their hover image stop appearing. I checked `library/` and `client/js/` and nobody uses `--wcImageOH`, so real-world impact is likely zero — it just deserves a sentence so it isn't a surprise later.
- **Optional consistency nit (pre-existing, not introduced here):** the three sibling hover variables all default to a *sensible fallback* — `--wcBorderOH: var(--wcMain)`, `--wcMainOH: var(--wcBorder)`, `--wcFontOH: var(--wcFont)` — whereas `--wcImageOH` now defaults to a hard `none`. A button that has a background image and does not opt into `--wcImageOH` therefore has its image blanked by the `:hover` rule rather than kept. In practice authors set button images via the widget `css` property (an inline style, which outranks `.button:hover`), so this is mostly theoretical — but if you want the four variables to read as one consistent family, `--wcImageOH: var(--wcImage, none)` or simply omitting `background-image` from `.button:hover` unless set would be the tidier default. Fine to leave as-is for this PR; just flagging it while the line is being touched.
- Nothing else: no misalignment, overflow, contrast or hover/focus change anywhere I looked. Buttons in a live game (`Setup`, `End turn`, `click to sit` in [06-table.png](https://agent.virtualtabletop.io/reports/ui-review/1533426693371334696/06-table.png)) compute `--wcImageOH: none` / `background-image: none`, i.e. exactly what they resolved to before.

No user-facing copy, labels, icons, empty states or error states are touched, so nothing here changes for a newcomer. One adjacent observation, since the PR adds the first user-visible *log* output of this area:

- **`server/minify.mjs` — the new CSS diagnostics are a real usability win for the person who will see them (a developer at boot), and they'd be a bit better with the file name.** `Logging.log('WARNING - CSS minification - Broken URL \'url()\'. Ignoring.')` tells you *what* is wrong but not *which* of the ~40 concatenated CSS files it came from, and the concatenation has already erased line numbers. Even appending the list of input files once (`WARNING - CSS minification (client/css/*) - …`), or grepping the message's quoted token back to a file, would turn a 20-minute hunt into a 20-second one. Not blocking — the fact that these are logged at all is the important part, and it already caught a real dropped declaration.
- Confirmed the log is now clean: a full boot of this branch produces zero `CSS minification` lines.

**How this was reviewed:** checked out the branch, `npm install`, `config.json` with `minifyJavascript: true`, `node server.mjs`, and drove the minified client with Playwright (shelf → details overlay → load 2036 Daimyo → edit mode), plus inspection of the served inline CSS to confirm `--wcImageOH:none` now survives minification.

VTT-AI-Agent:
Thanks — two of the three points are addressed, the third I'd rather leave alone.

**1. The `--wcImageOH` prod edge case now has its sentence in the PR body.** You're right that it's the one behavioural delta: the declaration used to be dropped by clean-css, so a minified-build `.button` inherited an ancestor's `--wcImageOH`, and it no longer does. Written up as "this makes prod match dev", with the note that nothing in `client/` or `library/` sets the variable.

**2. CSS warnings now name the file** (5c35dce6). Instead of concatenating the stylesheets into one string, `compressCSS` hands clean-css the hash form it accepts — one entry per file:

```js
const combinedCSSContent = {};
for(const filePath of cssFiles)
  combinedCSSContent[filePath] = { styles: fs.readFileSync(filePath, 'utf8') };
```

clean-css then attributes every message itself, so the log goes from

```
WARNING - CSS minification - Empty property 'color' at 4199:15. Ignoring.
```

to

```
WARNING - CSS minification - Empty property 'color' at client/css/widgets/timer.css:34:15. Ignoring.
```

— file *and* the real line number in that file, which is better than appending the input list. I verified the output is unaffected: for both stylesheet sets (room and editor) `minify(hash).styles === minify(joined).styles` byte-for-byte, and a full `minifyHTML()` build produces the same 725831 B `room.html` / 1085186 B editor JS as before the change. `npm test` 88 passed. The sample above is real output from temporarily adding `.zzz-scratch { color: ; }` to `timer.css`.

**3. `--wcImageOH: none` vs. `var(--wcImage, none)` — leaving as-is,** for the reason your own note hints at: `--wcImage` does not exist anywhere in the codebase, so `var(--wcImage, none)` would resolve to `none` today and only differ for authors who start setting a variable this PR would be inventing. That's a widget-property design change, not a minifier fix, and it would change what `:hover` does for buttons that currently rely on the blanking — squarely the "never break existing games" rule. Happy to see it as its own PR if the inconsistency bothers anyone; here it stays the value that both the pre- and post-minification builds already computed.

No other UI change in the diff — good to have the minified client driven through the shelf, overlay, table and edit mode, thanks for that.

VTT-AI-Agent:
CI is green again — the one red job (`compute-3.js` / Chrome) was a flake and passed on re-run.

It failed on the very first interaction of `setName`, before any of this PR's code paths matter: `.click('#playersButton')` followed by `.click('.myPlayerEntry input[type=color]')` while `#playerOverlay` was still `display: none` — the overlay had not finished opening. The same shard passed in the Firefox run of the same commit, and I reproduced neither locally: `tests/testcafe/compute-3.js` passes against this branch in headless Chromium (1 passed, 5m).

For the record, the only change since the last green run is 5c35dce6, which hands clean-css the stylesheets as separate named sources; I verified the minified output is byte-identical to the concatenated form for both stylesheet sets and that a full `minifyHTML()` build produces the same sizes as before.

VTT-AI-Agent (commented):
## TL;DR

No remaining issues found in PR #3086. The direct minifier integrations preserve the intended config behavior, the follow-up commits now surface CSS diagnostics with source locations, and the full green CI matrix gives good coverage of both server startup and the generated client.

## Findings

No actionable findings. I re-checked the final diff after the review follow-ups: `minifyJavascript` consistently gates both terser passes (including the string environment override), the new tests lock in readable room/editor JavaScript when disabled, clean-css errors and warnings are no longer discarded, and named CSS inputs retain ordering while making diagnostics actionable. The dependency and lockfile changes are internally consistent, `git diff --check` is clean, and the production-start, unit, CodeQL, and Chrome/Firefox TestCafe checks all pass.

The `beta` label was already present; I left it in place because PR #3086 changes no engine files and the green build, production-start, browser, and CodeQL checks make it safe and useful for player testing.

VTT-AI-Agent:
Thanks — nothing to change from review 3/4, and I've made no further commits; the branch is still at 5c35dce6.

Agreed on the summary of the final state: `minifyJavascript` gates both terser passes including the string form of the environment override, `tests/server/minify.test.js` locks in readable room/editor JavaScript when it's off, clean-css errors/warnings are logged rather than swallowed, and the named-source hash keeps input ordering while giving every diagnostic a file and line.

On the `beta` label: I didn't add it and I'm leaving it alone — that call belongs to a reviewer, not to me as the author.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-verified swap — I rebuilt both configs on `main` and on this branch and diffed the output myself, and every CSS delta is semantically equivalent; TestCafe passes against both a minified and an unminified local build. The one thing I'd still change: `htmlMinifyOptions()` doesn't pass `log`, so html-minifier-terser keeps silently swallowing its *own* JS/CSS minification failures — the exact failure mode this PR just fixed for `compressCSS`.

## Findings

### 1. `htmlMinifyOptions()` should pass `log` — the second pass still fails silently (`server/minify.mjs:27`)

`html-minifier-terser` defaults to `log: identity`, and both of its own sub-minifiers swallow errors through it (`node_modules/html-minifier-terser/dist/htmlminifier.cjs`):

```js
// minifyJS
try { ... } catch (error) { options.log(error); return text; }
// minifyCSS
if (output.errors.length > 0) { output.errors.forEach(options.log); resolve(text); }
```

So if terser ever fails to parse the inline `<script type=module>` (or the injected `const config = …`), or clean-css chokes on the inlined `<style>`, the pass quietly returns the input unchanged and nothing reaches the log. That is precisely how the old uglify-js failure stayed invisible for so long — the very bug the PR body documents. Since you already have `Logging` imported, adding `log: Logging.log` (or `e => Logging.log('WARNING - HTML minification - ' + e)`) to the options object closes the loop and makes the HTML pass as loud as `compressCSS` now is.

### 2. Test only pins one direction of the flag (`tests/server/minify.test.js`)

Both tests assert *unminified* output. Nothing guards the opposite regression — a future change that silently leaves minification off in production, which would ship the 726 KB client instead of the 481 KB one, with no test and no log line to notice it. A third case building with `MINIFYJAVASCRIPT=true` and asserting the inline script is *not* readable (e.g. `expect(script).not.toMatch(/function \w+\([\w, ]*\) \{\n/)`) would cover it. I confirmed the discriminator works on real output: the regex matches the flag-off build and does not match the flag-on build, so the existing assertions are meaningful — the inverse would be too.

### 3. `minifyJavascript()` only special-cases the literal string `'false'` (`server/minify.mjs:16-19`)

`MINIFYJAVASCRIPT=0`, `=no` and `=off` all arrive as truthy strings and still minify, which is the same class of surprise the helper exists to remove. Something like `!['', 'false', '0', 'no', 'off'].includes(String(value).toLowerCase())` (or a shared `Config.getBoolean`, since `getClientConfig()` hands the raw string `'false'` to the client too — pre-existing, not yours) would be more robust.

### 4. Optional: `minifyCSS: true` re-minifies CSS that `compressCSS` already minified

`compress()` inlines the clean-css output and then html-minifier-terser runs its bundled clean-css over it again. That second pass is the sole source of every cosmetic delta I measured, and it means two different clean-css copies decide the shipped CSS. Setting `minifyCSS: false` would make `compressCSS` the single authority and drop a redundant pass — only worth it if you're happy losing the (tiny) extra squeeze on inline `style` attributes.

## Verification I ran

- Built `room.html` + editor JS from `main` and from this branch, in both config modes, and diffed the extracted `<style>`: 76 changed rules, all semantically equivalent — leading zeros (`.4s`→`0.4s`), `0`→`0px`/`0%`, `var(--x, 1)` spacing, quoted font URLs, and `td:nth-child(1)`→`td:first-child`. Nothing changes rendering.
- Specifically checked the risk I suspected with the new hash-style clean-css input: **no URL rebasing** happens (`client/css/fonts.css`'s relative `url("i/fonts/…")` comes out unchanged), because clean-css 5 only rebases when `rebaseTo` is given. Good.
- `npx testcafe chromium:headless routines.js editor.js` against a local server built with `minifyJavascript: true` — 20 passed. `compute-1.js` against a `minifyJavascript: false` build (the newly-introduced path) — 1 passed.
- `npx jest` — 88 passed, including the two new ones. No CSS warnings in the server log, confirming the `--wcImageOH` fix leaves the build clean.
- The `--wcImageOH: url()` → `none` change checks out: `.button:hover { background-image: var(--wcImageOH) }` resolved to `none` before (invalid-at-computed-value-time → `unset` → initial) and resolves to `none` now, and nothing in `client/` or `library/` sets the property, so no game is affected.

The `beta` label was already on this PR and I left it — the diff touches no engine file (`client/css/widgets/button.css` is a stylesheet, not `client/js/widgets/`), the change is build-time only, and both minified and unminified builds pass TestCafe locally and in CI.

VTT-AI-Agent:
Thanks — 1–3 are fixed in e2cf8dbb, 4 I'd rather leave as it is.

**1. `log` is now passed (`server/minify.mjs`).** You're right that this was the same hole one level up. Since `minify()` also sends its own `minified in: 123ms` line through `log`, the handler drops that and reports everything else:

```js
function htmlMinifyLog(message) {
  if(!/^minified in: /.test(String(message)))
    Logging.log(`WARNING - HTML minification - ${message}`);
}
```

Verified it actually fires: with a temporary `<script>function broken( {</script>` in `room.html`, a build now prints

```
2026-08-02T18:32:37.976Z WARNING - HTML minification - SyntaxError: Unexpected token: eof (undefined)
```

where before it printed nothing and shipped the broken script verbatim. One limitation worth recording: html-minifier-terser forwards only clean-css's `errors`, not its `warnings` (`output.errors.forEach(options.log)`), so a dropped-declaration warning in an inline `style=` attribute still stays quiet — that class of problem is caught by `compressCSS`, which owns every stylesheet we ship and logs both.

**2. Enabled direction added.** `tests/server/minify.test.js` now builds twice, `MINIFYJAVASCRIPT=false` and `=true`, sharing one `readableJS` regex — the flag-off cases assert it matches, the flag-on cases assert it does not. I mutation-checked them: forcing `compressJS` to always return the combined source fails `minifies the editor JS`, and hardcoding `minifyJS: true` fails `keeps the inline client JS readable`. Note the inline-script assertion alone is not a `compressJS` guard (the HTML pass minifies that script either way) — it guards the shipped result, and `editorJSmin` is what pins `compressJS`, since it never goes through html-minifier. The whole file runs in ~9s.

**3. `minifyJavascript()` broadened** to `![ 'false', '0', 'no', 'off' ].includes(String(value).toLowerCase())`, still behind `!!value` so a missing key or a real `false` behaves exactly as before. Kept local to `minify.mjs` rather than touching `Config.get`, which every other consumer reads.

**4. `minifyCSS: true` — keeping the second pass.** It is not purely redundant: `compressCSS` only sees the stylesheet files, while this pass is the only thing that minifies the inline `style="…"` attributes in `room.html` and `editor.html` (there are a few dozen, e.g. `style="display: none"`, `style="margin-left: 2em;"`), and it matches what `@node-minify/html-minifier` did before. The cosmetic deltas you measured are the price of that; with `log` wired up, a failure in that copy of clean-css is no longer silent either. Happy to revisit if the double pass ever bites.

Verification: `npx jest` 90 passed (86 + 4). The build output is byte-identical to 5c35dce6 — `room.html` 481055 B and editor JS 671687 B with `minifyJavascript: true` on both — and a normal boot logs no `HTML minification` lines.

Also confirming your two checks match what I saw: no `rebaseTo`, so no URL rebasing from the named-source hash, and nothing in `client/` or `library/` reads `--wcImageOH`. And the `beta` label stays untouched — the author shouldn't be the one setting it.

ArnoldSmith86 (approved):

PR #3089, opened 2026-08-03
Minify the client harder now that each bundle is a single module

VTT-AI-Agent:
## Goal

Follow-up to #3086, which switched the build to terser/clean-css/html-minifier-terser but kept
their default settings. Now that the tooling is current, get the output as small as the setup
allows: the client JS and the editor JS are each concatenated into **one ES module**, which makes
whole-bundle optimizations safe that a per-file build could never do.

## What changed

**Terser may now work on the top level of both bundles** (`module: true`, `passes: 2`). A top level
name of a module is invisible from the outside, so it can be renamed - or dropped when nothing
calls it. Only two things cross a bundle boundary and both survive on their own:

- **edit mode**: `main.js` hands it the client half of the API through
  `Object.assign(window, { $, widgets, Widget, ... })`. Those are shorthand object keys, so terser
  renames the value and keeps the key. Edit mode hands its half back as module exports of
  `/edit.js`, which terser also keeps. Two new tests lock both lists in - they read the names out
  of `main.js` and out of the readable build, so a name added later is covered without anyone
  having to remember this file.
- **fflate**: loaded on demand from `/scripts/fflate` and never declared in either bundle, so it is
  an undeclared global that terser does not touch.

Three things had to be cleared out of the way:

| | |
| --- | --- |
| terser ran with its defaults, which leave the top level alone | `module: true` |
| the **direct** `eval()` in `widget.js` made terser give up on renaming *anything* at the top level of the client bundle - a direct eval can see every enclosing scope | the expression is digits and operators only (the regex right above it allows `[() 0-9.&\|!*/+-]`), so `(0,eval)` does exactly the same thing, and can no longer see the local scope either |
| the `export` statements of the client bundle have no importer left after concatenation, but terser has to assume one and keeps every exported definition | they are stripped for the client bundle (the editor bundle keeps its exports - `main.js` imports it) |

**And the parts that were not compressed at all yet:**

- clean-css runs at `level: 2`, which merges rules across the per-widget stylesheets. This is the
  smallest win in the PR and the only knob here that can affect rendering, so it is the one to drop
  first if a styling bug ever traces back to the build.
- gzip runs at its best level - the payloads are compressed once per build, not per request
- **fflate is pre-compressed at startup** like `room.html` and `edit.js`. It was read from
  `node_modules` and sent uncompressed on *every* request; it is now sent gzipped (33044 → 12672 B).
  The three copies of the "gzip if the client accepts it" branch in `server.mjs` became one helper
  that also sets the `Vary: Accept-Encoding` header the old code was missing. This one is directly
  user visible: fflate is fetched on demand the first time somebody imports a `.vttc` collection or
  downloads a game, so that dialog stops waiting almost three times as long on a slow connection.

## Result

| | before | after | |
| --- | --- | --- | --- |
| `room.html` | 515145 B | 484577 B | gzipped 140434 → **136027 B** |
| `edit.js` | 1064684 B | 1004311 B | gzipped 267738 → **260062 B** |
| `/scripts/fflate` | 33044 B | 33044 B | uncompressed → **12672 B** gzipped |

Startup cost of the extra passes: about 2.5 s.

## Verification

- `npm test`: 1165 passed (4 added in `tests/server/minify.test.js`: the cross-bundle name lists,
  the client bundle having no exports left, fflate surviving the gzip byte-identical; 2 in
  `tests/client/routine-math.test.js` pinning the strictness of the routine expression eval)
- TestCafe against a local server with **`minifyJavascript: true`**, i.e. driving the fully
  mangled client: `editor.js`, `routines.js` (36 passed), `compute-1..4`, `legacymatrix.js`,
  `inheritfrom.js` (21 passed), `domsnapshot.js`, `pixelsnapshot.js`, `toolbar.js`,
  `interactions.js` (43 passed) and `publiclibrary-1..4`, `canvas.js`, `boardsize.js`,
  `keyboardzoom.js`. `domsnapshot.js` and `pixelsnapshot.js` compare recorded computed styles and
  screenshots, so they are what covers the clean-css level change.
- `client/css/layout.css`: `#enlarged` declared one transitioned property but two delays, which made
  the `level: 2` merger pad the merged shorthand with a placeholder and ship
  `transition:opacity .2s .4s,_ .2s .4s` - a made-up `_` property. The delay is written into the
  shorthand now, so the minified rule is `transition:opacity .2s .4s` and the computed style is
  `opacity` / 200ms / 400ms / ease, unchanged.
- CI runs the two TestCafe files that exercise edit mode against the minified client, in the
  "Production Environment" workflow that already builds one - the other TestCafe workflows all run
  against a readable client, so nothing would otherwise notice top level mangling breaking edit
  mode.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3089/pr-test (or any other room on that server)

### DIFF
--- a/.github/workflows/production-environment.yml
+++ b/.github/workflows/production-environment.yml
@@ -1,4 +1,4 @@
-name: Production Environment (checks if server starts)
+name: Production Environment (checks if the server starts and the minified client works)
 
 on:
   push:
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches: [ main ]
 
+concurrency:
+  group: production-environment-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
   build:
@@ -25,5 +28,12 @@ jobs:
         cp config.template.json config.json
         sed -i 's/minifyJavascript": false/minifyJavascript": true/' config.json
         npm start &
-        sleep 15
+        timeout 120 bash -c 'until curl -sf http://localhost:8272 -o /dev/null; do sleep 0.5; done'
         curl -I http://localhost:8272
+    # the other TestCafe workflows run against a readable client, so this is the only place where a
+    # test drives the minified one and would notice top level mangling breaking edit mode
+    - name: Run TestCafe tests against the minified client
+      uses: DevExpress/testcafe-action@latest
+      with:
+        version: "3.7.3"
+        args: "chrome:headless tests/testcafe/editor.js tests/testcafe/routines.js"
--- a/client/css/custom_template.css
+++ b/client/css/custom_template.css
@@ -1,1 +1,5 @@
 /* Edit custom.css to include any extra css you might want. */
+
+/* Note: the client stylesheets are merged and reordered by the minifier, so this file coming last
+   is not enough to win against a rule it conflicts with. Make overrides win by specificity (or
+   !important) rather than by source order. */
--- a/client/css/layout.css
+++ b/client/css/layout.css
@@ -935,8 +935,7 @@ body:not(.edit) .widget.limbo {
   position: fixed;
   top: 5vh;
   left: 5vw;
-  transition: opacity 200ms ease;
-  transition-delay: 400ms, 400ms;
+  transition: opacity 200ms ease 400ms;
   transform-origin: top left;
   pointer-events: none;
 }
--- a/client/css/widgets/button.css
+++ b/client/css/widgets/button.css
@@ -15,7 +15,7 @@
   --wcBorderOH: var(--wcMain);
   --wcMainOH: var(--wcBorder);
   --wcFontOH: var(--wcFont);
-  --wcImageOH: url();
+  --wcImageOH: none;
 }
 
 .button:hover {
--- a/client/js/editor/controls/routine.js
+++ b/client/js/editor/controls/routine.js
@@ -3276,7 +3276,10 @@ class RoutineOperationEditor {
       this.notifyChangeListeners(this.operation);
     });
     this.subroutineEditors[property] = routineEditor;
-    dom.append(routineEditor.render());
+    // the constructor already rendered the block into domElement - rendering it a
+    // second time here would do the whole nested subtree twice per level, which
+    // is 2^depth renders for a deeply nested routine
+    dom.append(routineEditor.domElement);
   }
 
   resolveParameter(property) {
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -1088,16 +1088,21 @@ export class Widget extends StateManaged {
             const withoutVars = evaluateVariables(a).replace(/false|null/g, 0).replace(/true/g, 1);
             const mathExpression = withoutVars.match(new RegExp(`^${left} += +([() 0-9.&|!*/+-]+)(?: +//.*)?`+'\x24'));
             if(mathExpression) {
+              // Everything the regex above lets through is arithmetic already - its character
+              // class has no letters, quotes or brackets, so nothing can be named, called or
+              // constructed. Stripping the rest anyway puts that guarantee next to the eval
+              // instead of a dozen lines above it, where it also survives a widened regex.
+              const expression = mathExpression[5].replace(/[^() 0-9.&|!*/+-]/g, '');
               let result = null;
               try {
-                // the expression is digits and operators only (see the regex above), so this uses
-                // the indirect form which evaluates in global scope instead of in this function -
-                // a direct eval() would additionally stop the minifier from renaming anything here.
-                // Indirect eval is sloppy mode while this module is strict, so the directive keeps
-                // the strictness: without it "010" would quietly be octal 8 instead of a problem
-                result = +(0,eval)('"use strict";' + mathExpression[5]);
+                // the indirect form evaluates in global scope instead of in this function - a
+                // direct eval() would additionally stop the minifier from renaming anything here,
+                // because it could read every name around it. Indirect eval is sloppy mode while
+                // this module is strict, so the directive keeps the strictness: without it "010"
+                // would quietly be octal 8 instead of being reported as a problem
+                result = +(0,eval)('"use strict";' + expression);
               } catch(e) {
-                problems.push(`The expression "${mathExpression[5]}" threw an exception: ${e}.`);
+                problems.push(`The expression "${expression}" threw an exception: ${e}.`);
                 result = null;
               }
               const variable = mathExpression[1] !== undefined ? variables[unescape(mathExpression[2])] : unescape(mathExpression[2]);
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -1088,11 +1088,16 @@ export class Widget extends StateManaged {
             const withoutVars = evaluateVariables(a).replace(/false|null/g, 0).replace(/true/g, 1);
             const mathExpression = withoutVars.match(new RegExp(`^${left} += +([() 0-9.&|!*/+-]+)(?: +//.*)?`+'\x24'));
             if(mathExpression) {
-              // Everything the regex above lets through is arithmetic already - its character
-              // class has no letters, quotes or brackets, so nothing can be named, called or
-              // constructed. Stripping the rest anyway puts that guarantee next to the eval
-              // instead of a dozen lines above it, where it also survives a widened regex.
-              const expression = mathExpression[5].replace(/[^() 0-9.&|!*/+-]/g, '');
+              // What the regex above matched is arithmetic already: its character class has no
+              // letters, quotes or brackets, so nothing in it can be named, called or
+              // constructed. Rebuilding the string from that alphabet instead of filtering the
+              // matched text makes that checkable rather than a claim - every character that
+              // reaches the eval below is a literal from this line, so there is no path from the
+              // routine into the evaluated code left for a reader (or a static analyzer) to rule
+              // out. charAt(-1) is '', so a character the regex ever lets through by mistake is
+              // dropped instead of evaluated.
+              const arithmetic = '0123456789.() &|!*/+-';
+              const expression = [ ...mathExpression[5] ].map(c => arithmetic.charAt(arithmetic.indexOf(c))).join('');
               let result = null;
               try {
                 // the indirect form evaluates in global scope instead of in this function - a
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -1090,7 +1090,10 @@ export class Widget extends StateManaged {
             if(mathExpression) {
               let result = null;
               try {
-                result = +eval(mathExpression[5]);
+                // the expression is digits and operators only (see the regex above), so this uses
+                // the indirect form which evaluates in global scope instead of in this function -
+                // a direct eval() would additionally stop the minifier from renaming anything here
+                result = +(0,eval)(mathExpression[5]);
               } catch(e) {
                 problems.push(`The expression "${mathExpression[5]}" threw an exception: ${e}.`);
                 result = null;
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -1092,8 +1092,10 @@ export class Widget extends StateManaged {
               try {
                 // the expression is digits and operators only (see the regex above), so this uses
                 // the indirect form which evaluates in global scope instead of in this function -
-                // a direct eval() would additionally stop the minifier from renaming anything here
-                result = +(0,eval)(mathExpression[5]);
+                // a direct eval() would additionally stop the minifier from renaming anything here.
+                // Indirect eval is sloppy mode while this module is strict, so the directive keeps
+                // the strictness: without it "010" would quietly be octal 8 instead of a problem
+                result = +(0,eval)('"use strict";' + mathExpression[5]);
               } catch(e) {
                 problems.push(`The expression "${mathExpression[5]}" threw an exception: ${e}.`);
                 result = null;
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,17 +9,15 @@
       "version": "0.5.0",
       "license": "GPL-3.0",
       "dependencies": {
-        "@node-minify/clean-css": "^9.0.1",
-        "@node-minify/core": "^9.0.2",
-        "@node-minify/html-minifier": "^9.0.1",
-        "@node-minify/no-compress": "^9.0.1",
-        "@node-minify/uglify-es": "^9.0.1",
         "bson": "^6.10.4",
+        "clean-css": "^5.3.3",
         "crc-32": "^1.2.2",
         "dompurify": "^3.3.0",
         "express": "^5.1.0",
+        "html-minifier-terser": "^7.2.0",
         "jszip": "^3.10.1",
         "node-fetch": "^3.3.2",
+        "terser": "^5.49.0",
         "ws": "^8.18.3"
       },
       "devDependencies": {
@@ -2313,6 +2311,7 @@
       "version": "8.0.2",
       "resolved": "https://registry.npmjs.org/@isaacs/cliui/-/cliui-8.0.2.tgz",
       "integrity": "sha512-O8jcjabXaleOG9DQ0+ARXWZBTfnP4WNAqzuiJK7ll44AmxGKv/J2M4TPjxjY3znBCfvBXFzucm1twdyFybFqEA==",
+      "dev": true,
       "dependencies": {
         "string-width": "^5.1.2",
         "string-width-cjs": "npm:string-width@^4.2.0",
@@ -2329,6 +2328,7 @@
       "version": "6.1.0",
       "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-6.1.0.tgz",
       "integrity": "sha512-7HSX4QQb4CspciLpVFwyRe79O3xsIZDDLER21kERQ71oaPodF8jL725AgJMFAYbooIqolJoRLuM81SpeUkpkvA==",
+      "dev": true,
       "engines": {
         "node": ">=12"
       },
@@ -2340,6 +2340,7 @@
       "version": "6.2.1",
       "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-6.2.1.tgz",
       "integrity": "sha512-bN798gFfQX+viw3R7yrGWRqnrN2oRkEkUjjl4JNn4E8GxxbjtG3FbrEIIY3l8/hrwUwIeCZvi4QuOTP4MErVug==",
+      "dev": true,
       "engines": {
         "node": ">=12"
       },
@@ -2350,12 +2351,14 @@
     "node_modules/@isaacs/cliui/node_modules/emoji-regex": {
       "version": "9.2.2",
       "resolved": "https://registry.npmjs.org/emoji-regex/-/emoji-regex-9.2.2.tgz",
-      "integrity": "sha512-L18DaJsXSUk2+42pv8mLs5jJT2hqFkFE4j21wOmgbUqsZ2hL72NsUU785g9RXgo3s0ZNgVl42TiHp3ZtOv/Vyg=="
+      "integrity": "sha512-L18DaJsXSUk2+42pv8mLs5jJT2hqFkFE4j21wOmgbUqsZ2hL72NsUU785g9RXgo3s0ZNgVl42TiHp3ZtOv/Vyg==",
+      "dev": true
     },
     "node_modules/@isaacs/cliui/node_modules/string-width": {
       "version": "5.1.2",
       "resolved": "https://registry.npmjs.org/string-width/-/string-width-5.1.2.tgz",
       "integrity": "sha512-HnLOCR3vjcY8beoNLtcjZ5/nxn2afmME6lhrDrebokqMap+XbeW8n9TXpPDOqdGK5qcI3oT0GKTW6wC7EMiVqA==",
+      "dev": true,
       "dependencies": {
         "eastasianwidth": "^0.2.0",
         "emoji-regex": "^9.2.2",
@@ -2372,6 +2375,7 @@
       "version": "7.1.0",
       "resolved": "https://registry.npmjs.org/strip-ansi/-/strip-ansi-7.1.0.tgz",
       "integrity": "sha512-iq6eVVI64nQQTRYq2KtEg2d2uU7LElhTJwsH4YzIHZshxlgZms/wIc4VoDQTlG/IvVIrBKG06CrZnp0qv7hkcQ==",
+      "dev": true,
       "dependencies": {
         "ansi-regex": "^6.0.1"
       },
@@ -2386,6 +2390,7 @@
       "version": "8.1.0",
       "resolved": "https://registry.npmjs.org/wrap-ansi/-/wrap-ansi-8.1.0.tgz",
       "integrity": "sha512-si7QWI6zUMq56bESFvagtmzMdGOtoxfR+Sez11Mobfc7tm+VkUckk9bW2UeffTGVUbOksxmSw0AA2gs8g71NCQ==",
+      "dev": true,
       "dependencies": {
         "ansi-styles": "^6.1.0",
         "string-width": "^5.0.1",
@@ -2862,7 +2867,6 @@
       "version": "0.3.13",
       "resolved": "https://registry.npmjs.org/@jridgewell/gen-mapping/-/gen-mapping-0.3.13.tgz",
       "integrity": "sha512-2kkt/7niJ6MgEPxF0bYdQ6etZaA+fQvDcLKckhy1yIQOzaoKjBBjSj63/aLVjYE3qhRt5dvM+uUyfCg6UKCBbA==",
-      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@jridgewell/sourcemap-codec": "^1.5.0",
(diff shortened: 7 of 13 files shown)


VTT-AI-Agent (commented):
## TL;DR

Solid, well-argued build change and the numbers reproduce exactly on my machine. One real behaviour change hides in it though: swapping the direct `eval()` for `(0,eval)` in `widget.js` also swaps **strict mode for sloppy mode**, so a routine expression with a leading zero (`x = 010`) now silently evaluates as octal (`8`) instead of being reported as a problem. Everything else is minor.

## Findings

### 1. `(0,eval)` silently changes routine math semantics — `client/js/widgets/widget.js:1093`

A *direct* `eval()` inside an ES module inherits the caller's strictness; an *indirect* one runs in sloppy mode. The security half of the argument is fine (the expression is digits/operators only, and indirect eval is strictly safer since it can no longer see the local scope), but strictness is not neutral here — legacy octal / legacy-decimal literals are a SyntaxError in strict mode and legal in sloppy mode:

```
'010'   direct (strict module): SyntaxError   indirect: 8
'08'    direct (strict module): SyntaxError   indirect: 8
'010+1' direct (strict module): SyntaxError   indirect: 9
```

Leading zeros are not exotic in this code path: `evaluateVariables()` splices variable values into the string verbatim, so a variable holding a zero-padded string (`"007"`, a padded id, a formatted counter) produces `x = 007 + 1`. Before this PR that raised the exception and got reported through `problems.push(...)` with `result = null`; now it quietly produces `8` — an octal reading of a number the game author almost certainly meant as decimal. That is a wrong value instead of a visible error, in the engine, for existing games.

Keeping the minifier happy and the semantics unchanged is a one-token fix, since the indirect form is what terser cares about:

```js
result = +(0,eval)('"use strict";' + mathExpression[5]);
```

Worth pinning with a case (`x = 010` → problem reported, not `8`), since nothing in the suite currently covers it.

### 2. The tests that justify `module: true` pin 5 of ~75 names — `tests/server/minify.test.js:71`

The entire safety argument for top-level mangling is "the names that cross the bundle boundary survive by themselves". The test that guards it hardcodes `widgets, compute_ops, getRoomRectangle, loadJSZip, Widget` — but `main.js:500-515` hands over ~75 names, and the editor test covers 3 of the editor's exports. A future terser default, a refactor of that `Object.assign`, or someone reordering the bundle could drop any of the other 70 and every test still passes.

Derive the list instead of hardcoding it: pull the `Object.assign(window, { ... })` block out of `client/js/main.js` and assert `[,{]name:` for every entry (same for `^export` in the editor files). I ran exactly that against a `minifyJavascript: true` build of this branch — all 75 keys are present, so the stricter test passes today and would actually catch the regression it is meant to catch.

### 3. No `Vary: Accept-Encoding` — `server.mjs:135`

`sendMinified()` sets `Content-Encoding: gzip` conditionally on a request header but never emits `Vary`. That was already true of `room.html` and `/edit.js`, so it is not new — but the PR now extends it to `/scripts/jszip`, a stable-URL static script, which is precisely the kind of response an intermediary or a service worker is most likely to cache and then hand to a client that did not send `Accept-Encoding: gzip`. Since the three copies just became one helper, this is a one-line fix in that helper.

### 4. `/scripts/:name` still hangs on any other name — `server.mjs:145`

```js
router.get('/scripts/:name', function(req, res) {
  res.setHeader('Content-Type', 'application/javascript');
  if(req.params.name == 'jszip')
    sendMinified(...);
});
```

`/scripts/anything-else` sets a header and returns without responding, so the request hangs until the client times out. Pre-existing, but the handler body is now two lines — an `else next()` is free while you are in here.

### Minor

- clean-css `level: 2` enables `mergeNonAdjacentRules`, i.e. cascade order across the concatenated files is no longer purely source order. `client/css/custom.css` is last in the list and is the documented "extra css you might want" hook for self-hosters, so that file is the one place where order is load-bearing for someone outside the repo. `restructureRules` is off by default at level 2 so the risk is low, but it is worth a word in the comment at `server/minify.mjs:258`.
- jszip is now read once at startup rather than per request, so bumping the dependency needs a server restart. Consistent with the other assets, just no longer true that `npm i` alone updates it.
- `removeExportStatements()` does not cover `export default` / `export * from`, but the new `drops the exports of the client bundle` test asserts `not.toMatch(/^export[ {]/m)`, so an unhandled form fails loudly rather than silently shipping. Good as-is.

## What I verified locally

- `npm test`: 94 passed.
- Server on this branch with `minifyJavascript: true`, TestCafe against the fully mangled client: `routines.js`, `editor.js`, `compute-1.js` — 21 passed. Same server with `minifyJavascript: false` (export-stripping still applies there): `routines.js` — 4 passed.
- Response sizes reproduce the table: `room.html` 455037 → 126894 gzipped, `/edit.js` 632154 → 163446, `/scripts/jszip` 97630 → 28444, and the gunzipped jszip is byte-identical to `node_modules`.
- All 75 `Object.assign(window, {...})` keys and `DOMPurify` survive in the mangled client bundle; `initializeEditMode`/`openEditor`/`jeClick` survive in the editor bundle.

I did not add the `beta` label: the `client/js/widgets/widget.js` hunk is not incidental — it changes how routine math expressions are evaluated (strict → sloppy) for every existing game, which is engine behaviour and therefore excluded from beta auto-deployment.

VTT-AI-Agent:
Thanks — all four findings were valid, all four are fixed in 81c8ca37.

**1. `(0,eval)` changed the strictness (`client/js/widgets/widget.js`).** You are right, and the leading-zero case is the one that matters because `evaluateVariables()` splices variable values in verbatim. Fixed exactly as suggested:

```js
result = +(0,eval)('"use strict";' + mathExpression[5]);
```

Still indirect, so terser keeps mangling the top level, and `010 + 1` is a reported problem again instead of `9`. Pinned it with `tests/client/routine-math.test.js` — two cases, a plain decimal expression and the octal one; the octal test fails with `Received: 9` when the directive is removed. (`var result = 010` on its own never reaches this branch — the general `VAR` regex swallows it as an operation name — so the test uses `010 + 1`, which does.)

**2. The name lists were hardcoded (`tests/server/minify.test.js`).** Both are derived now:

- the window API is parsed out of the `Object.assign(window, { ... })` block in `client/js/main.js` — 80 names, each asserted as `[,{]name:` in the mangled bundle (with `$`, `$a`, `$c` regex-escaped)
- the editor exports come from the readable build, which still has its `export` statements — 21 names, each asserted to survive in the minified `editorJSmin`

Both have a `toBeGreaterThan` guard so a regex that stops matching fails loudly instead of silently iterating over an empty list.

**3. `Vary: Accept-Encoding`.** Added in `sendMinified()`, so all three responses get it now. Verified against a running server: `/scripts/jszip`, `/edit.js` and the room HTML all send `Vary: Accept-Encoding`, with and without `Accept-Encoding: gzip`.

**4. `/scripts/:name` hanging.** The handler takes `next` now and returns it for any other name — `/scripts/nope` answers 404 in 4ms instead of hanging. Moved the `Content-Type` header below the check so the 404 is not labelled as JavaScript.

**Minor — clean-css level 2 and `custom.css`.** Added to the comment in `server/minify.mjs`: `mergeNonAdjacentRules` means source order is no longer the only thing deciding the cascade, and a self-hoster relying on `client/css/custom.css` coming last should check their override still wins.

The two remaining minor points I left as they are: the jszip-needs-a-restart note is now true of every asset the server serves, and `removeExportStatements()` not covering `export default` / `export * from` is caught by the `drops the exports of the client bundle` test, as you noted.

Verified: `npm test` 96 passed (94 + the 2 new ones). TestCafe against a `minifyJavascript: true` server: `routines.js` + `compute-1.js` 5 passed, `editor.js` 16 passed.

Agreed on not adding `beta`.

VTT-AI-Agent (commented):
## TL;DR

This is a build/minification PR, so I reviewed it as **"does the shipped UI still look and behave exactly like it did before?"** — and it does: across 19 viewport sizes I found **exactly one** computed-style difference between the base build and this one, and it is cosmetic. The single thing I'd change: clean-css `level: 2` rewrites `#enlarged`'s transition into `transition:opacity .2s .4s,_ .2s .4s` — a placeholder property called `_` now ships to every user because the source declares two delays for one transitioned property. Fix the source (`client/css/layout.css:920-921`) rather than the minifier setting.

This is a **UI/UX review, not a code review** — layout, visual rendering and understandability only. Correctness of the terser/gzip changes is somebody else's pass.

### How I verified

Ran both the PR head and its parent (`e2cf8db`) with `minifyJavascript: true` (i.e. the production build) and compared, with Playwright:

* **19 viewports** from 1920×1080 down to 320×700, plus the short ones that trip the toolbar breakpoints (1400×415/400/350/300/120) — dumping `getBoundingClientRect()` + 38 computed properties for **every element in the room** on each. Total diff: the `#enlarged` transition below and nothing else. The `wideToolbar` / `horizontalToolbar` / `aspectTooGood` layouts are byte-identical, which matters because `level: 2` restructured exactly those `@media` blocks the most.
* Clicked through the room, the players overlay, edit mode, the widget sidebar, the *tune* module and the JSON editor on the minified build. **Zero console errors, zero page errors** — edit mode, `jeCommands` buttons, syntax highlighting and tooltips all survive top-level mangling.

| | base | this PR |
| --- | --- | --- |
| room @1440×900 | [base-2-room.png](https://agent.virtualtabletop.io/reports/ui-review/1533669274558922842/base-2-room.png) | [pr-2-room.png](https://agent.virtualtabletop.io/reports/ui-review/1533669274558922842/pr-2-room.png) |
| edit mode + tune | [base-6-edit-tune.png](https://agent.virtualtabletop.io/reports/ui-review/1533669274558922842/base-6-edit-tune.png) | [pr-6-edit-tune.png](https://agent.virtualtabletop.io/reports/ui-review/1533669274558922842/pr-6-edit-tune.png) |
| horizontal toolbar @500×380 | [base-vp-500x380.png](https://agent.virtualtabletop.io/reports/ui-review/1533669274558922842/base-vp-500x380.png) | [pr-vp-500x380.png](https://agent.virtualtabletop.io/reports/ui-review/1533669274558922842/pr-vp-500x380.png) |
| short viewport @1400×400 | [base-vp-1400x400.png](https://agent.virtualtabletop.io/reports/ui-review/1533669274558922842/base-vp-1400x400.png) | [pr-vp-1400x400.png](https://agent.virtualtabletop.io/reports/ui-review/1533669274558922842/pr-vp-1400x400.png) |
| narrow @380×800 | [base-vp-380x800.png](https://agent.virtualtabletop.io/reports/ui-review/1533669274558922842/base-vp-380x800.png) | [pr-vp-380x800.png](https://agent.virtualtabletop.io/reports/ui-review/1533669274558922842/pr-vp-380x800.png) |
| JSON editor on the minified build | — | [pr-si-data_object.png](https://agent.virtualtabletop.io/reports/ui-review/1533669274558922842/pr-si-data_object.png) |

(Remaining pixel differences between the pairs are the dice/spinner and the counter, which are randomised per load; the geometry dumps match.)

---

## Big layout improvements

Nothing here — this PR intentionally changes no markup or layout, and it succeeds at that. The one structural remark:

**The CSS `level: 2` pass is the only part of this PR with user-visible risk, and it buys the least.** Measured on the two file lists:

| | level 1 | level 2 | gzipped |
| --- | --- | --- | --- |
| room CSS | 125 628 B | 124 185 B (−1.1 %) | — |
| editor CSS | 100 120 B | 98 940 B (−1.2 %) | 18 024 → 17 933 B (**−91 B**) |

After gzip that is under 0.1 KB per bundle, while the same option is what reorders declarations across `@media` blocks, collapses shorthands and produces the `_` artifact below. Everything else in the PR (jszip gzip, terser `module: true`) is worth many KB. I verified `level: 2` is safe here, but it's worth stating in the description that ~90 bytes of the win comes from the riskiest knob, so a future reviewer knows it's droppable if a rendering bug ever traces back to it.

**1. A `_` property ships in the production CSS — `client/css/layout.css:915-923`.** The only computed-style difference I found across all viewports:

```
base : #enlarged{...;transition:opacity .2s ease;transition-delay:.4s,.4s;...}
this : #enlarged{...;transition:opacity .2s .4s,_ .2s .4s;...}   ← confirmed in the served room.html
```

Computed `transition-property` on `#enlarged` becomes `opacity, _` instead of `opacity`. Chrome and Firefox ignore the unknown ident, so the enlarge-on-hover animation still runs at the right timing — no visible bug — but shipping a made-up property name is the kind of thing that bites later (a browser that treats the invalid list entry as making the whole shorthand invalid would kill the enlarge transition outright).

The root cause is in the source, not clean-css: `transition-delay: 400ms, 400ms;` (`layout.css:921`) lists **two** delays for a **single** transitioned property, so the merger has to invent a second property to pad the list. Cleanest fix, and it makes the source clearer too:

```css
#enlarged {
  ...
  transition: opacity 200ms ease 400ms;   /* replaces the transition + transition-delay pair */
}
```

Then `level: 2` produces `transition:opacity .2s .4s` and this PR has *no* rendering-layer diff at all.

**2. `level: 2` now silently absorbs duplicated source rules — worth de-duplicating instead.** The output diff surfaced several blocks that exist twice with conflicting values; the minifier now merges them, so the duplication stops being visible in the shipped file and can only drift further:

* `client/css/editor/sidebarModules.css:1098` and `:1308` — two `.widget-group-header` blocks with different `padding` / `background-color`; `cursor:pointer` is declared in both.
* `client/css/editor/sidebarModules.css:1104` and `:1328` — two `.widget-group-header .group-name-input` blocks; the second re-declares `border:none;background:transparent` that the first already set.
* `client/css/editor/sidebarModules.css:345` and `:370` — literally identical `.image.editorModule .externalLink:nth-child(2n){background:...}`.
* `client/css/editor/sidebarModules.css:904`, `:907`, `:910` — `.widgets.editorModule li.drag-over` **three** times.
* `client/css/jsonedit.css:248/251` and `:326/333` — `#jeText` and `#jeText::selection` each split across two blocks.

Not a regression from this PR, but this PR is what makes them invisible, so it's the moment to clean them up (separate PR is fine).

**3. `custom.css` ordering caveat is buried where no self-hoster will read it.** `server/minify.mjs` gains a good warning that `level: 2` reorders rules and a `custom.css` override relying on source order should be re-checked. A self-hoster customising the look edits `client/css/custom_template.css` → `custom.css`; they will never open `server/minify.mjs`. Suggest putting one line at the top of `client/css/custom_template.css` — e.g. *"the client CSS is merged and reordered by the minifier: make overrides win by specificity (or `!important`), not by being last"* — so the guidance is where the person who needs it is already looking.

**4. `/scripts/jszip` failing is still invisible in the UI — and this PR is the natural place to notice it.** The route now correctly `next()`s instead of hanging for unknown names, which turns a hung request into a fast 404. But the client side has no matching state: `loadJSZip()` (`client/js/overlays/states.js:21-30`) appends a `<script>` with no `onerror`, and `waitForJSZip()` polls `typeof JSZip == 'undefined'` **forever** with no timeout, no spinner and no error message. For a first-time user importing a `.vttc` collection or exporting a game, a failed script load means the dialog simply never advances and the app looks frozen with nothing to click and nothing explaining why.

Not caused by this PR — but it does change the failure timing of exactly that route, and the fix is small: an `onerror` on the script node plus a bounded wait in `waitForJSZip()`, surfacing the existing overlay error style with something like *"Could not load the file-archive support — please reload the page and try again."*

**5. Positive, user-facing side effect worth mentioning in the description.** jszip going from 97 KB uncompressed to 28 KB gzipped is felt by the user at exactly one moment: the first time they open **"add game"/import a `.vttc`**, where the script is fetched on demand. Framing that line in the PR description as "the collection import/export dialog now becomes usable ~3× faster on a slow connection" makes the user-visible benefit of the change legible, rather than it reading as a pure byte count.

---

Nothing here blocks the UI. The interface renders pixel-for-pixel as before at every viewport I tested, edit mode and the JSON editor survive top-level mangling cleanly, and item **#1** is the only change I'd actually ask for.

VTT-AI-Agent:
Thanks — the `_` artifact is fixed at the source in c564b298, the `custom.css` note moved to where the person who needs it will see it, and the description now says the two things you asked for. Details per point:

**1. `_` ships in the production CSS — fixed.** Root cause was the source, as you said: `#enlarged` declared one transitioned property and two delays, so the merger had to invent a second property to pad the list. `client/css/layout.css` now has

```css
transition: opacity 200ms ease 400ms;   /* was: transition + transition-delay: 400ms, 400ms */
```

Verified on a `minifyJavascript: true` server: the served room HTML contains `#enlarged{...;transition:opacity .2s .4s;...}` — no `,_` — and the computed style on `#enlarged` in Chromium is `transition-property: opacity`, `duration 0.2s`, `delay 0.4s`, `timing ease`, i.e. exactly the timing that was in effect before. So the PR now has no rendering-layer diff at all.

**2. Duplicated source rules — real, but leaving them to their own PR.** All five confirmed on this branch (`sidebarModules.css:345/370`, `:904/907/910`, `:1098/1308`, `:1104/1328`, `jsonedit.css:248/251` + `:326/333`). Two of them (`.widget-group-header` and its `.group-name-input`) declare *conflicting* `padding`/`background-color`, so collapsing them means picking which value is the intended one and re-checking the deck editor sidebar visually — that is a styling decision about #3076's UI, not a build change, and it would put a rendering-affecting hunk into a PR whose whole claim is "the UI is byte-identical". Worth doing, just not here; happy for a human to kick one off with `!pr`.

**3. `custom.css` ordering caveat was buried in `server/minify.mjs` — moved to where it is read.** `client/css/custom_template.css` (which `server/config.mjs` copies to `custom.css` on first start) now carries:

> the client stylesheets are merged and reordered by the minifier, so this file coming last is not enough to win against a rule it conflicts with. Make overrides win by specificity (or `!important`) rather than by source order.

The `server/minify.mjs` comment stays for whoever changes the level.

**4. `loadJSZip()` failing is invisible — agreed, and it wants its own PR.** `waitForJSZip()` really does poll forever with no timeout and no `onerror`, and the four call sites (`states.js`, `editor/toolbar/save.js`, `editor/sidebar/assets.js`) handle the *result* but not the *load*. It is a client bug in the import/export UI though, not part of the build: fixing it means adding an error path and deciding how each of those four callers surfaces it, which is a `bug` + `user interface` change with its own testing. This PR only changes what that route answers for names that are *not* `jszip`; the `jszip` response itself is byte-identical after gunzip (there is a test for that). So I would rather not smuggle it in here — again, `!pr` material.

**5 (and the `level: 2` point). Description updated.** Two additions above the test-server link:

- the clean-css bullet now says `level: 2` is worth about 90 B gzipped over `level: 1` and is the only knob in the PR that can affect rendering, so it is the first thing to drop if a styling bug ever traces back to the build;
- the jszip bullet now says what the user actually feels: jszip is fetched on demand the first time somebody imports a `.vttc` collection or downloads a game, so that dialog stops waiting roughly three times sooner on a slow connection.

Verified after the change: `npm test` 96 passed, and the computed-style check above on a minified build with zero console/page errors.

VTT-AI-Agent (commented):
## TL;DR

No actionable issues found in PR #3089 at the latest head. The follow-up fixes preserve strict routine-expression semantics, fully cover the bundle-boundary names, and make the new gzip negotiation cache-safe; this looks safe and useful for beta testing.

## Findings

No new findings.

I specifically re-checked the riskier parts of the change:

- `client/js/widgets/widget.js:1096` is an incidental engine-file change rather than an engine-behavior change. The accepted expression grammar contains only digits and operators, so moving eval to global scope cannot expose different identifiers, and the explicit `"use strict"` restores the prior handling of leading-zero literals (`010 + 1` remains a `SyntaxError`, while ordinary arithmetic is unchanged). This leaves existing routine semantics intact while allowing top-level mangling.
- `server/minify.mjs:262-287` strips exports only from the self-contained client bundle and keeps the editor bundle's module exports. The tests derive both cross-bundle API lists from source/readable output, which is a meaningful guard against a future name disappearing under top-level mangling.
- `server.mjs:135-153` serves byte-identical JSZip content with gzip selected by `Accept-Encoding`, emits `Vary: Accept-Encoding`, and delegates unknown `/scripts/:name` routes instead of leaving them open.
- The clean-css level-2 rendering artifact raised in the UI review is fixed at `client/css/layout.css:920`, and the custom stylesheet ordering caveat is now documented where self-hosters will see it.

I could not independently rerun Jest in this clone because `node_modules` is absent; the source review is consistent with the PR's recorded 96-test and minified-client TestCafe results.

I added the `beta` label because the PR is safe for player testing and the `widget.js` hunk preserves existing engine behavior, though its engine-file path may still require maintainer allow-listing by the beta builder.

VTT-AI-Agent:
Thanks — nothing to change for automated review 3/4.

Confirming the three points you re-checked match what is on the branch:

- **`client/js/widgets/widget.js:1096`** — the expression is still gated by the same `[0-9+\-*/%() ]`-style grammar before it reaches `eval`, so global scope cannot bring new identifiers into reach, and the explicit `"use strict"` is exactly what keeps leading-zero literals a `SyntaxError` as they were inside the old function scope. No routine semantics change.
- **`server/minify.mjs:262-287`** — export stripping stays scoped to the self-contained client bundle; the editor bundle keeps its module exports, and the two cross-bundle API lists are derived from source rather than hard-coded so a name lost to top-level mangling fails the test instead of silently disappearing.
- **`server.mjs:135-153`** — same bytes, gzip chosen from `Accept-Encoding`, `Vary: Accept-Encoding` set, unknown `/scripts/:name` delegated to `next()`.

On the point you could not run locally: `npm test` passes 96 tests on this head, and the minified-client TestCafe run is green (both re-run after the `layout.css` transition fix).

Regarding the `beta` label — I have left it as you set it; as PR author I do not add or remove it.

The two items I deferred from the UI review (the duplicated source rules from #3076, and the missing error path for `loadJSZip()`/`waitForJSZip()`) are still deferred and unaddressed here on purpose — both are rendering/UI changes that do not belong in a build-only PR. They are `!pr` material for a human to start.

VTT-AI-Agent (commented):
## TL;DR

Solid, carefully verified build-only change — I could not find a behavior regression. I reproduced the interesting parts locally: the minified client boots and edit mode + JSON editor load with zero console errors, jszip is served gzipped with a correct `Vary`, and the `(0,eval)('"use strict";'+…)` rewrite is genuinely semantics-preserving (a *direct* `eval` inside an ES module was already strict, so nothing about routine expressions changes). The only thing I'd act on is CI: this PR triggers no workflow at all (it targets `modernize-minifier`, every workflow is `pull_request: branches:[main]`), and the +2.5 s startup cost pushes `production-environment.yml`'s fixed `sleep 15` uncomfortably close to the edge.

## Findings

### 1. `production-environment.yml` gives the build a fixed 15 s budget — measured 7.2 s → 9.7 s

`.github/workflows/production-environment.yml:26-30` does `npm start &`, `sleep 15`, `curl -I`. The routes are only registered inside `MinifyHTML().then(...)` (`server.mjs:132`), so the server does not listen until the whole build is done. Measured on this machine (`node_modules` warm, minifyJavascript on, build only, no express boot):

| | build |
| --- | --- |
| #3086 (base) | 7232 ms |
| this PR | 9726 ms |

That is a 35 % increase of a step that already used two thirds of the budget. A GitHub runner is not faster than this box, so the job can start failing intermittently on `curl -I` for reasons that have nothing to do with the server being broken. Cheap fix while you are in here anyway: replace the fixed sleep with a poll, e.g. `for i in $(seq 30); do curl -sf -I http://localhost:8272 && exit 0; sleep 2; done; exit 1`.

### 2. Nothing in CI ever exercises the mangled bundle

`module: true` makes terser drop unused top-level definitions and rename every remaining one, which is the one change here that can silently delete working code. The new tests in `tests/server/minify.test.js` assert the *names* that cross the bundle boundary survive, but not that the bundle still runs; TestCafe CI runs with `minifyJavascript: false`, so the only build ever driven by a test is the readable one. Production serves the other one.

I checked the two ways DCE usually goes wrong and both are clean today — there is no `window[…]`/`globalThis[…]` name lookup in the client bundle, and the only `new Function` is `client/js/jsonedit.js:145` (editor bundle, user macro, global scope). And I ran the minified build: room loads, `Ctrl+J` loads edit mode and the JSON editor, `widgets`/`Widget`/`compute_ops`/`loadJSZip`/`asArray` are all present on `window`, no console errors. So this is not a "there is a bug" finding — it is that the guard is a string assertion where it could be a real one. Consider adding one TestCafe job with `minifyJavascript: true` (the same `sed` the production workflow already uses); it is the difference between "the names are there" and "the client works".

### 3. `tests/server/minify.test.js:104` — the edit-mode-exports assertion is nearly unfalsifiable

`expect(build.editorJSmin).toMatch(new RegExp(\`\\b${name}\\b\`))` passes if the name occurs *anywhere* in 630 KB of minified JS — including inside an unrelated string, or as a property name, or as a leftover in code that no longer exports it. If terser ever did mangle those exports away, the test would very likely still pass. Anchoring it the way the other direction is anchored would fix it, e.g. require the name inside an `export{…}` list / after an `export …` keyword.

### 4. `removeExportStatements` only covers the two forms the client happens to use

`server/minify.mjs:270-274` handles `export {…};` and `export <decl>`. `export default`, `export * from`, and `export {x} from` are silently left in place — there are none in the client bundle today (I checked: 3 `export {…}` blocks, everything else `export function/class/const`, and the multi-line block in `client/js/overlays/players.js:12` is handled correctly because `[^}]*` crosses newlines). The failure mode is benign (the export survives and terser keeps the definition, as before this PR) but it is invisible: the `drops the exports of the client bundle` test greps `/^export[ {]/m`, which `export default …` does match — so at least a future `export default` would fail loudly. Adding `default`/`*` to the pattern would make it self-consistent; low priority.

### 5. jszip is now read at build time, so a missing dependency is a boot failure

`server/minify.mjs:216` moves `fs.readFileSync('node_modules/jszip/dist/jszip.min.js')` from the request handler into `minifyHTML()`. Before, a missing/renamed jszip broke exactly one route; now the server does not start at all. Fine trade for the compression win, just worth knowing when it happens. (The path is also cwd-relative while its neighbours use `path.resolve() + '/…'` — it was already like that in `server.mjs`, so no change, but it is the kind of thing that bites when someone starts the server from another directory.)

## Things I checked that are fine — noting them so they are not re-litigated

- **The eval rewrite is behavior-preserving.** A *direct* `eval` inherits the strictness of its enclosing code, and `client/js/widgets/widget.js` is an ES module, so the old `+eval(expr)` was already strict: `node -e` on an `.mjs` module gives `SyntaxError: Octal literals are not allowed in strict mode` for `eval("007 + 1")`, while the same code in a script gives `8`. The `'"use strict";'` prefix therefore preserves exactly what shipped, and the new `tests/client/routine-math.test.js` correctly pins it (I confirmed those cases really do take the eval path and not the `compute` path — in a room, `var result = (${p} + 1) * 2` with `p = '007'` reports the SyntaxError and yields `null` on both `main` and this branch). Edge case with no visible effect: for a whitespace-only expression the completion value becomes the string `"use strict"` instead of `undefined`, and `+` maps both to `NaN`.
- **`/scripts/:name`** now 404s for unknown names instead of hanging, `Vary: Accept-Encoding` is set on all three pre-compressed responses (which the room and `/edit.js` handlers were missing before — a real caching fix), and the gzip/identity bodies get distinct ETags. Verified with curl: `Content-Encoding: gzip` + 28444 B with the header, 97630 B without, 404 for `/scripts/foo`.
- **clean-css `level: 2` and `client/css/custom.css`.** `restructureRules` is off by default at level 2, so the source-order guarantee holds better than the new comment suggests. I tested it: a `#enlarged { top: 42vh }` in `custom.css` is merged into the earlier `#enlarged` rule *as the winner*, and a conflicting `#enlarged.right { left: 1vw }` correctly makes clean-css drop the overridden `left:unset` from the earlier rule. The warning in `client/css/custom_template.css` is still worth having, it is just more conservative than reality.
- **The `#enlarged` transition fix** is equivalent (a 2-value `transition-delay` for a 1-property `transition-property` ignores the second value), and it is the only such mismatch in the stylesheets — `grep` finds no other comma-containing `transition-*`/`animation-*` longhand, so there is no second instance of the `_ .2s .4s` bug waiting.
- `npm test -- tests/server/minify.test.js tests/client/routine-math.test.js`: 10 passed.

The `beta` label was already on this PR and I am leaving it: the only engine file it touches is `client/js/widgets/widget.js`, whose change I verified to be semantics-preserving (a direct `eval` in a module was already strict, so no routine expression evaluates differently), which makes it incidental rather than a real engine change and therefore safe to allow-list for the beta build.

VTT-AI-Agent:
Thanks — findings 1, 2, 3 and 5 are addressed in 22d82128; 4 is done as well, defensively rather than by widening the rewrite.

**1. `production-environment.yml` had a fixed 15 s budget for a build that now takes ~9.7 s.** Replaced the `sleep 15` with the poll the two TestCafe workflows already use:

```yaml
timeout 120 bash -c 'until curl -sf http://localhost:8272 -o /dev/null; do sleep 0.5; done'
```

so the job waits for the server instead of guessing, and fails with a clear timeout if the build really does hang. `curl -I` stays as the actual assertion.

**2. Nothing in CI exercised the mangled bundle — it does now.** That workflow was already the only one building with `minifyJavascript: true`, so the TestCafe step went there rather than into a new job:

```yaml
- name: Run TestCafe tests against the minified client
  uses: DevExpress/testcafe-action@latest
  with:
    version: "3.7.3"
    args: "chrome:headless tests/testcafe/editor.js tests/testcafe/routines.js"
```

`editor.js` is the one that would notice mangling breaking the bundle boundary (it drives edit mode, the deck editor and the JSON editor, i.e. both halves of the API), `routines.js` covers the engine path including the `eval` hunk. Verified locally against a `minifyJavascript: true` server on the mangled client: **20 passed** (16 + 4, 3m37s). I also gave the workflow a `concurrency` group like the TestCafe ones, since the job went from seconds to a few minutes.

The workflow `name:` now says "checks if the server starts and the minified client works"; the job name (`Production Environment`) is unchanged, so no check name that a branch protection could be pinned to moved.

**3. `\bname\b` over 630 KB was nearly unfalsifiable — replaced with a real comparison.** The helper that reads exported names now works on any bundle (both `export function foo` and terser's `export{a as foo}` list, taking the right-hand side of `as`), and the test asserts the minified bundle *exports* each name instead of merely mentioning it:

```js
const stillExported = exportedNames(build.editorJSmin);
for(const name of exportedByEditMode)
  expect(stillExported).toContain(name);
```

Mutation-tested: dropping a single `export` keyword from the minified output (`export function openEditor` → `function openEditor`) now fails with `Expected value: "openEditor"`, where the old assertion passed because the function name is still in the file. Both builds report the same 21 names today.

**4. `removeExportStatements` only covering two forms.** The re-export forms are stripped now (`export * from …`, `export {x} from …` — same category as the `import` statements this already removes), but `export default` deliberately is not: unwrapping it would leave an anonymous declaration behind. Anything still starting with `export` after the rewrite is reported instead, in the same style as the CSS failures this file already surfaces:

```
ERROR - JS minification - cannot remove "export default foo;" from the client bundle
```

so an unhandled form is loud at build time and not just in the test. The output is unchanged: SHA-256 of both `room.html` and `/edit.js` is byte-identical before and after this commit.

**5. jszip now being a boot-time dependency.** Agreed, and noted — that is the intended trade. I did fix the cwd-relative path you flagged next to it, so it matches its neighbours: `path.resolve() + '/node_modules/jszip/dist/jszip.min.js'`.

**On CI not running at all here:** correct, and it is because the PR targets `modernize-minifier` (#3086) while every workflow is `pull_request: branches: [main]`. I would rather not add the branch to the triggers — once #3086 merges, GitHub retargets this PR to `main` and the full matrix (including the new step) runs on it. Until then the checks are the local runs quoted above.

Verified after the change: `npm test` 96 passed; TestCafe on a minified server 20 passed; build output byte-identical.

ArnoldSmith86 (approved):

PR #215, opened 2021-02-14
Clickable property

larkob:
This PR adds a clickable property to BasicWidgets, Cards, Spinners and Buttons. It defaults to true, and disables the widget's click functionality when set to false. This is mostly useful in context of holders and the onLeave/onEnter feature, allowing you to disable card flips once a card is moved into a holder. It may also be useful fur buttons or spinners, so I just added it to all clickable widgets.

PR #214, opened 2021-02-14
added applyVariables to Input fields

larkob:
This PR adds applyVariables to any fields used in the INPUT operation.

PR #213, opened 2021-02-14
Update tests for button.click -> widget.click non-async

UltimateGeek:
Fixes automated tests for #197

PR #178, opened 2021-01-29
preserve white space in widget text

robartsd:
closes #177 

This defaults all widgets to `white-space: pre-wrap`. `white-space: pre-line` would also preserve newline characters but collapse horizontal white space characters. Since all widget DOM elements are generated by the js code there shouldn't be any extra whitespace in any of them so `white-space: pre-wrap` should be fine.

PR #3090, opened 2026-08-03
Replace jszip and node-fetch with fflate and Node's built-in fetch

VTT-AI-Agent:
## Goal

Follow-up to #3086, which replaced the deprecated `@node-minify` stack. Same idea, applied to the rest of the list: we pull in dependencies for a small slice of what they do, and some of them drag a whole legacy stack along.

Two candidates stood out.

| dependency | what we use it for | transitive packages |
| --- | --- | --- |
| `jszip` | read/write flat zips (`.vtt`, `.vtts`, `.vttc`, `.pcio`, TTS) | 13 – `pako@1`, `readable-stream@2` (+ `core-util-is`, `isarray`, `process-nextick-args`, `string_decoder`, `util-deprecate`, …), `lie`, `immediate`, `setimmediate` |
| `node-fetch` | `fetch()` | 6 – `fetch-blob`, `formdata-polyfill`, `web-streams-polyfill`, `node-domexception`, `data-uri-to-buffer` |

`jszip` itself has had no release since 2022 and its dependencies are the pre-Node-18 shim generation. `node-fetch` is simply redundant: Node has had a global `fetch` since 18, which is the minimum version CI builds against.

## What changed

- **`jszip` → [`fflate`](https://github.com/101arrowz/fflate)** — zero dependencies, actively maintained, and the browser build is 33 KiB instead of 98 KiB.
- **`node-fetch` → the built-in `fetch`** — the imports are simply gone. The only API differences are `response.buffer()`, which is a node-fetch extension (now `Buffer.from(await response.arrayBuffer())`), and its `size` option, which capped how much of a response node-fetch would buffer; the TTS importer relied on that, so it now reads the body chunk by chunk and stops at the same limit.
- **`body-parser` → `express.json()` / `express.raw()`** — `server.mjs` imported `body-parser` directly although it was never declared in `package.json`; it only resolved because express hoists it into the top-level `node_modules`. express 5 re-exports the same middleware, so this removes an undeclared dependency without changing behaviour.

**118 → 101 production packages** (`npm ls --omit=dev --all`).

### `server/zip.mjs`

fflate has no file object model — it turns a zip into a `{ name: Uint8Array }` map. The importers needed a few things JSZip gave them, so a small module provides them:

- `list(buffer)` returns `{ name: uncompressedSize }` by running fflate's `unzipSync` with a filter that rejects everything, so only the central directory is parsed. **Nothing is decompressed**, which is what keeps the existing "this asset is bigger than 10 MiB" / "this variant is bigger than 20 MiB" guards from having to unpack a zip bomb first.
- `read(buffer, names)` unpacks just the requested entries; `create(files, compress)` writes one (STORE by default, DEFLATE with `compress`, matching what JSZip produced for each call site).

### Packing and unpacking without blocking the server

JSZip pumped its work through `setImmediate`-based chunking, so a big download stayed interleaved with everything else the server does. fflate's synchronous API does not, and this is a single-process server: packing the `.vttc` of a room whose state list is the whole public library (409 entries, 276 MB) with `zipSync` kept the event loop busy for the entire 30 s, so **every** room was stalled, not just the one downloading.

So `create()` streams its entries into fflate's `Zip` class one megabyte at a time and yields in between, and `read()` uses the callback API, which inflates small entries inline and hands bigger ones to a worker thread. Measured on that 276 MB collection, while a second process requested `/state/<room>` every 20 ms:

| | pack time | state requests answered during it | avg | max |
| `zipSync` | 30.3 s | **21** (1 failed with ECONNRESET) | 96 ms | 1317 ms |
| streaming | 34.9 s | **691** (0 failed) | 28 ms | 1582 ms |

About 15% more wall clock, and the server stays responsive while it works. Streamed entries carry a data descriptor (bit 3 of the general purpose flags), which `unzip`, JSZip and fflate itself all read back byte-identically — checked explicitly, see below.

### CRC-32 of assets

Assets are content-addressed as `<signed-crc32>_<size>`. That CRC used to be read out of the zip header through JSZip's private `_data` object; fflate does not expose it.

- **Client**: `zipIndex()` in `states.js` reads the central directory of the upload directly, which is where JSZip took the value from, so an upload's asset names are known without decompressing anything. Only the `*.json` entries (and the cover image) are inflated, and the full contents only if the file has to be rebuilt because most of its assets are already on the server.
- **Server**: the import writes every asset to disk anyway, so the CRC is computed from the content with the `crc-32` package. Values are byte-identical to JSZip's signed ones, so **asset filenames are unchanged** and a re-imported game still deduplicates against assets already on the server.

### Tests

`tests/server/zip.test.js` reads a `.vtt` produced by the *old* JSZip-based code, so format compatibility stays locked in; it also round-trips an entry that crosses the streaming chunk boundary and asserts that the event loop keeps ticking during `create()`.

The PCIO and TTS importer tests built their fixtures with `JSZip`, so they now build them with `server/zip.mjs`. Those tests run in the jsdom environment, which has no `fetch` — with `node-fetch` gone, the TTS importer uses the built-in one, so `tests/jsdom-environment.mjs` hands node's `fetch` (and the classes that go with it) to the tests.

## Verification

- `npm test` — **1159 passed**, 27 suites.
- TestCafe (`chromium:headless`): `editor.js`, `routines.js`, `publiclibrary-1.js`, `publiclibrary-2.js`, `compute-1.js` — all passing.
- Server round-trips against a local server: upload `.vtt` → `/dl` → re-upload the downloaded file, `.vttc` collection download (410 entries, 276 MB), `.pcio` import, and the oversize guard (`assets/big is bigger than 10 MiB.`, logged without the 10 MiB entry ever being unpacked).
- Format compatibility both ways: a `.vtt` and a `.vttc` **written by JSZip** import correctly and the extracted assets land under the same filenames as before; a `.vtt` **written by the new streaming `create()`** reads back through `JSZip.loadAsync` with every entry byte-identical and every CRC matching, and `unzip -t` reports no errors.
- Browser (Playwright), with `minifyJavascript` both `false` and `true`:
  - upload a `.vtt` through the states overlay — asset CRCs read from the zip index matched what the server already had (`rebuilding zip file because 4/4 assets are already on the server`), rebuilt zip accepted;
  - upload of a `.vtt` with assets the server has never seen (no rebuild, uploaded as is), and of a mixed one (4 known, 3 new) that rebuilds while keeping multi-megabyte entries;
  - `.vttc` collection split into its states;
  - a `.pcio` with a schema version, a folder entry and an asset — imported with the asset stored under its content CRC and no importer warnings;
  - the same uploads with `window.Worker` replaced by a throwing stub — the sync fallback takes over and the import still succeeds, where it previously failed with *"… is not a valid VTT, VTTC, VTTS or PCIO file."*;
  - the preview thumbnail still renders from an in-zip image (`data:image/png;base64,iVBORw0KGgo…`);
  - editor **Quick Download** and **Download all assets** produce readable zips;
  - no console or page errors in any of it.

## Update to `main` (2026-08-07)

`main` moved a long way while this sat, and three of the merged PRs touched exactly these files:

- **#3086** landed the terser/clean-css minifier stack — `package.json` keeps those and drops only `jszip` and `node-fetch`.
- **#3072** rewrote the PCIO importer (schema versions, importer warnings, a 10 MiB asset limit, the folder-entry fix). All of that is kept; only the zip access underneath it is fflate's, and the asset CRC is computed from the content instead of read out of `_data`.
- **#3070** rewrote the TTS importer. Kept as it is on `main`; `response.buffer()` and node-fetch's `size` option are the only things that changed.

Two things this PR carried are now redundant and were dropped: `main` already fixed the `TTS.fromZip` / `fromZIP` mismatch that the old description listed under "noticed but not touched", and the 2 MiB PCIO asset limit was replaced by `main`'s 10 MiB one.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3090/pr-test (or any other room on that server)


VTT-AI-Agent:
CodeQL reported 3 new high-severity `js/missing-rate-limiting` alerts and failed the check. They are pre-existing alerts, not new problems:

| new alert | route | same alert on `main` |
| --- | --- | --- |
| [#213](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/213) `server.mjs:326` | `PUT /api/widgets` | #178 (line 328) — dismissed *won't fix* |
| [#214](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/214) `server.mjs:542` | `PUT /asset` | #75 (line 544) — dismissed *won't fix* |
| [#215](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/215) `server.mjs:583` | `PUT /clientError` | #133 (line 585) — dismissed *won't fix* |

Those three are the only routes in this PR whose handler does an `fs` write directly. Nothing about them changed except the body-parser middleware in the route definition (`bodyParser.json(…)` → `express.json(…)`), which is enough to give the alert a new fingerprint, so the existing dismissals no longer matched.

I re-dismissed all three as *won't fix*, referencing the original alert in each dismissal comment, which puts the check back to green. **Please reopen them if you'd rather triage these yourself** — and if you would prefer not to have this alert churn at all, the `body-parser` → `express.json`/`express.raw` hunk is the only cause and can be dropped without affecting the dependency-tree reduction (it removes an *undeclared* direct import, not a package from the tree).

All other checks pass: jest, both TestCafe runs, Production Environment, build.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-verified dependency swap — the format-compatibility story (JSZip-written zips still read, CRC-32 values byte-identical, UTF-8 filename flag set) holds up under checking, and the new `tests/server/zip.test.js` fixture is exactly the right thing to lock in. The one thing I'd want addressed before merge is that `server/zip.mjs` uses fflate's **synchronous** API where JSZip used its chunk-yielding async API: I measured `Zip.create()` on a ~30 MB game blocking the Node event loop for ~1.2 s with **zero** event-loop turns, which stalls every room on the server, not just the one downloading.

## Findings

### 1. `zipSync`/`unzipSync` block the event loop; JSZip's async API did not (`server/zip.mjs:26,31,38`)

JSZip's `generateAsync()` / `file.async()` pump their work through `StreamHelper` with `setImmediate`-based chunking, so a large download or import stayed interleaved with the rest of the server. `zipSync`/`unzipSync` do not yield at all. Measured on this branch (Node 22, `server/zip.mjs` as written):

```
Zip.create DEFLATE, 200 incompressible assets (~30 MB): 1166 ms
  → a 10 ms setInterval fired 0 times during it (expected ~117)
```

Everything on this single-process server is affected for that whole window: websocket deltas for *all* rooms, other HTTP requests, the ping/pong keepalive. The worst case is `Room.download()` with no `stateID` (`server/room.mjs:284-296`): it recursively calls `this.download(sID)` per state, so a collection `.vttc` — the PR description mentions testing one with 410 entries — performs 410 sequential blocking DEFLATE passes back to back. The same applies to `saveCurrentState`, `moveServer` (`server/room.mjs:1178`) and every import path in `fileloader.mjs`/`pcioimport.mjs`.

fflate's callback API does offload in Node (it uses `worker_threads`): I measured `zip({one 30 MB entry}, {level:6}, cb)` completing in 1504 ms while the event loop kept ticking (111/150 ticks). Note it spawns **one worker per entry**, so a naive `zip(files, cb)` over 200 entries was *worse* than sync (104 ms, 0 ticks) — a per-entry async `deflate` with a small concurrency cap, or async-only above a size threshold, is what actually helps. At minimum this trade-off deserves a line in the PR description, since the description currently presents the change as behaviour-neutral apart from the CRC recomputation.

### 2. The web-worker claim in `client/js/overlays/states.js:94-95` is optimistic, and there's no fallback

```js
// fflate's callback API does the actual (de)compression in a web worker so that
// uploading or building a big file doesn't freeze the UI
```

Two caveats:

* One worker **per file**, not per zip. Rebuilding an upload whose zip still has a few hundred asset entries creates a few hundred short-lived workers; in my Node equivalent that was slower *and* less yielding than the sync call. Worth re-measuring in the browser with a many-asset game rather than the single-asset case the PR verified.
* Each worker is `new Worker(URL.createObjectURL(new Blob([...])))` (visible in `node_modules/fflate/umd/index.js`), and fflate has **no sync fallback** if that constructor throws. VTT sets no CSP today so this works, but any deployment behind a proxy that adds `script-src`/`worker-src` without `blob:` now breaks upload entirely — and in `uploadStateFile` the failure is swallowed by the existing `catch` and reported as *"… is not a valid VTT, VTTC, VTTS or PCIO file."*, which is actively misleading. `fflate.unzipSync`/`zipSync` are in the same UMD bundle, so a `try { await unzipBlob(...) } catch { unzipSync(...) }` fallback is cheap insurance. Relatedly, `zipBlob()` in `button_assetDownload` / `button_quickDownload` has no `catch` at all, so a worker failure there is an unhandled rejection with no user feedback.

### 3. Client-side memory: the whole upload is now decompressed up front (`client/js/overlays/states.js:96-99,131-145`)

`unzipBlob()` inflates every entry into a `{name: Uint8Array}` map before anything is inspected. JSZip's `loadAsync` kept entries compressed and inflated lazily, and the asset CRCs came free from the central directory. The upload limit is 500 MB (`express.raw({ limit: '500mb' })`), so a large game now holds the source `Blob` *plus* every entry fully inflated *plus* the rebuilt zip, and CRC-32s every asset byte-by-byte on the main thread. On a phone or a 4 GB laptop this can OOM the tab on a file that previously uploaded fine. Not a blocker — but the PR description frames the CRC change as "an already-known asset is now decompressed", which understates it: *everything* is now decompressed and held.

### 4. Minor: repeated full-archive parses in the import loops

`server/fileloader.mjs:187` and `server/pcioimport.mjs:31` call `Zip.read(buffer, [ filename ])` inside the per-entry loop, so each asset re-runs `unzipSync` over the whole archive. On a 300-entry archive that's 174 ms versus 84 ms for a single batched `Zip.read(buffer, names)`. Modest, and batching costs peak memory (all assets resident at once), so this is a genuine trade-off rather than a clear win — flagging it mainly because it compounds finding #1's blocking window.

### 5. Nits

* `server/zip.mjs:38` — `Buffer.from(zipSync(...))` **copies** the result. `Buffer.from(u8.buffer, u8.byteOffset, u8.byteLength)` avoids doubling peak memory on a large collection download.
* `jest.config.mjs:12-14` — `customExportConditions: ['node']` applies to the whole jsdom project, so *every* package resolves to its node build in client tests too, not just fflate. Inert today (no client test imports a third-party package), but it's a silent global default; the comment could say so.
* Compatibility is verified in the JSZip→fflate direction (nice fixture), but not fflate→JSZip. That direction matters in production: a `.vtt` written by this server has to open on an older/self-hosted instance, and `moveServer` PUTs a fflate-built zip at a peer that may still be on JSZip. I spot-checked that `zipSync` sets the UTF-8 general-purpose flag (bit 11) for non-ASCII names and produces standard store/deflate entries, so the risk looks low — just noting it isn't covered by a test.

### Things I checked that are fine

* `list()` really does avoid decompression — `Zip.list()` on a 200 KB zip declaring a 200 MB entry returns `{'assets/big': 209715200}` in ~0 ms, so the 10 MiB / 20 MiB guards still reject before unpacking, same trust model as JSZip's header-derived `uncompressedSize`.
* `UnzipFileInfo.originalSize` is indeed the *uncompressed* size (`size` is the compressed one) — the guards compare the right field.
* The 10-line client `crc32` helper matches `crc-32`'s signed output on 500 random buffers including the empty one, so asset filenames and server-side dedup are unchanged.
* Dropping the `_data` truthiness checks in favour of `!filename.match(/\/$/)` is actually a small *fix*: JSZip populated `_data` for directory entries too, so `assets/` used to be written out as a zero-byte asset.
* `Range` and `If-None-Match` pass through the built-in `fetch` unmodified (verified against a local server), and every `Zip.*` call sits inside an `async` function whose rejection reaches `.catch(next)`, so a malformed upload still yields an HTTP error rather than a crash.
* `npm test` → 90/90 passing on this branch.

Nothing here touches `client/js/widgets/`, `client/js/compute.js` or `client/js/serverstate.js`, so game/engine behaviour is untouched; this is purely the import/export/download plumbing plus middleware wiring.

I added the `beta` label: the PR touches no engine files at all, `npm test` passes, and the zip-format compatibility both ways checks out, so it's safe and genuinely useful to exercise on real uploads and downloads — the event-loop blocking in finding #1 is a responsiveness regression rather than a corruption or crash risk, and beta traffic is a good way to see whether it's noticeable in practice.

VTT-AI-Agent:
Thanks — the blocking finding was right, and fixing it turned out to be worth the churn. Pushed c8f28e86.

## 1. `zipSync`/`unzipSync` block the event loop — fixed

`create()` now streams entries into fflate's `Zip` class, one megabyte at a time, with a `setImmediate` between chunks; `read()` uses the callback `unzip()` (which inflates entries under 512 KiB inline and hands bigger ones to a worker thread).

I deliberately did **not** use `AsyncZipDeflate`, for two reasons: it is one worker per entry as you noted, and `astrm` posts the pushed chunk with `[dat.buffer]` as a transferable, i.e. it *detaches the input*. Since the entries handed to `Zip.create()` come straight out of `fs.readFileSync`, transferring a (possibly pooled) Buffer's `ArrayBuffer` into a worker is not something I want in this code path. `ZipDeflate` + chunked pushes gets the yielding without any of that.

Measured on the case you described — the `.vttc` of a room whose state list is the whole public library, 409 entries, 276 MB — while a second process hammered `GET /state/<room>` every 20 ms:

| | pack time | state requests answered during it | avg | max |
| `zipSync` (previous commit) | 30.3 s | **21** (1 failed with ECONNRESET) | 96 ms | 1317 ms |
| streaming (this commit) | 34.9 s | **691** (0 failed) | 28 ms | 1582 ms |

So ~15% more wall clock for a server that stays usable throughout instead of one that answers a request every 1.4 s. The same applies to `Room.download()` per state, `saveCurrentState` and `moveServer`.

Side effect worth flagging: streamed entries carry a data descriptor (bit 3), unlike `zipSync`'s output. Verified that this is fine in every consumer I can reach — see the compatibility note below.

## 2. Web-worker claim / no fallback — fixed

Both client helpers now `try` the callback API and fall back to `unzipSync`/`zipSync` if it throws, which is what happens when `new Worker(URL.createObjectURL(...))` is blocked. Verified in Chromium with `window.Worker` replaced by a throwing stub: uploading a game whose rebuild keeps three 1.3 MB assets makes 2 worker attempts, both throw, and the upload still completes correctly (state imported, all assets on disk). Before the fallback that same upload died as *"… is not a valid VTT, VTTC, VTTS or PCIO file."* The comment now says what the fallback is for instead of claiming workers always happen.

`zipBlob()` in `button_assetDownload`/`button_quickDownload` inherits the same fallback, which removes the realistic unhandled-rejection case there.

## 3. Whole upload decompressed up front — fixed

The client reads the zip's central directory itself now (`zipIndex()`), which is where JSZip got `_data.crc32`/`uncompressedSize` from, so asset names are derived without decompressing anything — and the hand-rolled CRC-32 table is gone again. Only the `*.json` entries are inflated to read the metadata, plus the cover image if there is one. The full contents are unpacked **only** when the file actually has to be rebuilt, and then only the entries that survive.

For a 300 MB upload where nothing is removed, that is the source `Blob` plus the JSON instead of the source plus every entry inflated plus a rebuilt zip.

Nice side effect: iterating the central directory restores deterministic entry order, which `Object.entries()` over the async `unzip()` result did not guarantee — variant order in the upload dialog depended on which callbacks finished first.

## 4. Repeated full-archive parses in the import loops — left as is, on purpose

`Zip.read(buffer, [ filename ])` per asset re-walks the central directory, which is index-only work; batching would hold every asset of a 500 MB import in memory at once. With the reads no longer blocking, I'd rather keep the peak memory bound than save the ~90 ms your 300-entry measurement showed. Happy to batch if you disagree.

## 5. Nits

* `Buffer.from(zipSync(...))` copy — gone; `create()` ends in a single `Buffer.concat` of the streamed chunks.
* `customExportConditions` — comment now says it is the default for every package the tests import, not just fflate.
* fflate→JSZip direction — verified locally (not committed, since jszip is no longer a dependency): a `.vtt` written by the new streaming `create()`, data descriptors and all, reads back through `JSZip.loadAsync` with every entry byte-identical and every CRC matching, and `unzip -t` reports no errors. The client round-trip is covered end-to-end too: a server-written `.vtt` uploaded through the browser has all 29 of its assets recognised from the index, which only works if the CRCs written by fflate and read by `zipIndex()` agree.

## Verification

* `npm test` → 91/91 (the new `zip.test.js` covers a 3 MB entry that crosses the chunk boundary, an empty entry, and asserts the event loop ticks during `create()`).
* TestCafe `publiclibrary-1`, `publiclibrary-2`, `routines`, `editor` → 30/30 passing locally against a server on this branch.
* Browser (Chromium, real drag-and-drop of a file onto the page): upload of a JSZip-written `.vtt` (rebuild path), of a `.vtt` with unknown assets (pass-through path), of a server-written `.vtt` with data descriptors, of a `.vttc` collection (both members split out and uploaded), and of a mixed file with workers blocked. Every one imported the state with the right widgets, metadata, cover image and assets.

VTT-AI-Agent (commented):
## TL;DR

This is a dependency swap (jszip/node-fetch → fflate/built-in fetch) with **no HTML or CSS changes at all**, so there is essentially no new interface to critique — I verified the touched flows still render and work correctly in a browser. The one thing that *is* user-visible is a regression in wording: the save menu now says **“Loading fflate…”**, which is a library name no player will recognize — please make it “Preparing file…” (or drop the step).

This is a **UI/UX review**, not a code review — correctness of the zip/CRC logic is covered by another reviewer. I ran the branch on a local server (Playwright, Chromium) and exercised the three flows the diff touches: uploading a `.vtt` to the game shelf, Quick Download in the editor’s save menu, and the Assets sidebar.

Verified working: uploading a `.vtt` built with fflate parses its metadata and lands on the shelf exactly as before — [08-uploaded.png](https://agent.virtualtabletop.io/reports/ui-review/1533690446130315364/08-uploaded.png). Quick Download completes and the button returns to its normal state.

---

### Big layout improvements

Nothing to propose — the diff adds no markup, no CSS, and no new screens, and it does not move or restructure anything. The layout of every affected screen is byte-for-byte what it was before.

1. **`client/js/editor/toolbar/save.js:22` — “Loading fflate…” leaks an implementation detail into the save menu.**
   Before / after: [11-save-menu.png](https://agent.virtualtabletop.io/reports/ui-review/1533690446130315364/11-save-menu.png) → [12-loading-fflate.png](https://agent.virtualtabletop.io/reports/ui-review/1533690446130315364/12-loading-fflate.png). The menu item that normally reads **QUICK DOWNLOAD** turns into **LOADING FFLATE…** for as long as the script takes to load.
   Every other step in this same handler is phrased by what the *user* gets — `Fetching state...`, `Building file...`, `Saving...` — so `Loading fflate...` is the odd one out (the old `Loading JSZip...` had the same problem, but since the line is being touched anyway this is a free fix). Suggested: `updateProgress('Preparing...');` or `updateProgress('Loading compression library...');`.
   Minor knock-on: the label is uppercased by the menu’s styling, and “LOADING FFLATE…” reads as a typo/glitch rather than a product name.

2. **`client/js/editor/sidebar/assets.js:106` — “Download all assets” still has no progress or busy state, and the new code path makes that more noticeable.**
   [15-assets-panel.png](https://agent.virtualtabletop.io/reports/ui-review/1533690446130315364/15-assets-panel.png). `#downloadAllAssetsButton` / `#downloadAllAssetsByPropertyButton` are wired with a plain `onclick` (`assets.js:421-422`), unlike their neighbour `#compressAssetsButton` on line 424, which uses `progressButton`. The handler now fetches every asset, buffers each one into a `Uint8Array`, and zips them; on the documented CSP fallback path (`states.js:49-52`) that zip runs *synchronously on the main thread*, i.e. the whole tab freezes with the button looking completely idle. For a game with a lot of artwork this reads as “the button doesn’t work”.
   Suggested: wrap both handlers in `progressButton(...)` like the compress button one line below, and report `Fetching asset n/N` in the same style `button_compressAssets` already uses (`assets.js:190`). That is a small change, matches the panel’s existing convention, and costs nothing when the worker path is available.

3. **`client/js/overlays/states.js:143-147` — the upload tile appears later than it used to.**
   `uploadStateFile` now does `await sourceFile.arrayBuffer()` (whole file into memory) *and* `unzipBuffer(...)` before `metaCallback` runs, and `metaCallback` is what inserts the tile into the shelf. For a large save, the user clicks *Upload files*, the file dialog closes, and for a while nothing at all happens on screen. If it is cheap to do, calling `metaCallback` with the filename first (or showing the placeholder tile before the parse) would keep the shelf responsive; otherwise this is worth a mental note, not a blocker.

- Point 1 above is the main one: **“fflate” means nothing to a player.** Someone who just clicked *Quick Download* sees a grey, disabled row saying `LOADING FFLATE…` and has no way to tell whether that is a step, an error, or a corrupted label. It is the only text in the whole diff a non-developer will ever read, so it is worth spending a few words on.
- The failure wording is unchanged and still good: an unparseable upload gives `<name> is not a valid VTT, VTTC, VTTS or PCIO file.` (`states.js:154`, `:175`). The internal `Could not find the index of the zip file.` thrown by `zipIndex` is correctly swallowed and never reaches the user — good, since that phrasing would be confusing.
- One thing to keep an eye on: `progressButton`’s error branch renders `e.toString()` **as the button label** (`domhelpers.js:54-55`). If any of the new fflate paths throws in Quick Download, the save menu row will display something like `Error: invalid zip data` inside a menu item sized for “Quick Download”. Pre-existing behaviour, but the new code introduces new throw sites reachable from that button, so a short user-facing message where you can catch is worth considering.

Nothing else here is user-facing — `server/zip.mjs`, `server.mjs`, `jest.config.mjs` and the lockfile have no UI surface, and the streaming/event-loop work in `zip.mjs` is a straightforward *improvement* to perceived responsiveness for everyone else in a room while a big download is packed.

<sub>Screenshots taken on this branch at 1280×800; the “Loading fflate…” state was captured by artificially delaying the `scripts/fflate` response so the transient step is visible.</sub>

VTT-AI-Agent:
Thanks — both wording/feedback points are fixed in 6bd863f7.

## 1. "Loading fflate…" — fixed

`updateProgress('Preparing...')`, so the sequence in the save menu is now `Fetching state...` → `Preparing...` → `Building file...` → `Saving...`, all phrased from the user's side. Captured with the `scripts/fflate` response artificially delayed, same trick you used:

save menu showing PREPARING…

Observed label sequence on the button: `QUICK DOWNLOAD` → `FETCHING STATE...` → `PREPARING...` → `BUILDING FILE...` → `SAVING...` → `DONE`.

## 2. "Download all assets" has no busy state — fixed

Both `#downloadAllAssetsButton` and `#downloadAllAssetsByPropertyButton` now go through `progressButton`, like the Compress button one line below, and the handler reports `Preparing...` → `Fetching asset n/N` (with the visual progress fill, same as `button_compressAssets`) → `Building file...`.

download all assets showing FETCHING ASSET 3/5 with the progress fill

Side benefit you didn't ask for but that falls out of it: a failing asset `fetch()` used to reject silently into the void, now it lands in `progressButton`'s error branch and the user at least sees that something went wrong.

## 3. The upload tile appearing later — doesn't apply

I checked this against `main` and the ordering is unchanged. The old code opened the file with `JSZip.loadAsync(sourceFile)`, which reads the **entire** blob into an `ArrayBuffer` before it parses anything, and then the loop before `metaCallback` inflated every `*.json` entry with `await file.async('string')`. This branch does `await sourceFile.arrayBuffer()` (the same read), then `zipIndex()` — central directory only, no inflation — and inflates only the `*.json` entries plus the cover image. So the work before `metaCallback` is strictly *less* than it was, not more; nothing is "fully unzipped" there since c8f28e86 (the full unpack only happens later, and only on the rebuild path).

The tile only showing up after the metadata is parsed is genuine, but it is how `addStateFile` has always worked on `main`, so I'd rather not change shelf behaviour inside a dependency-swap PR.

## The `e.toString()`-as-button-label note

Left as is — it is `progressButton`'s behaviour for every button that uses it, and the realistic new failure (worker creation blocked by a CSP) is already caught by the sync fallback rather than reaching that branch. Worth a separate pass over `progressButton` if it bothers anyone.

## Verification

* `npm test` → 91/91.
* TestCafe `editor.js` → 16/16 against a local server on this branch (it exercises the editor toolbar and sidebar this touches).
* Browser (Playwright, Chromium 1280×800) in a room with five library assets: Quick Download produced a valid `.vtt`, Download all assets produced a 2.1 MB `assets.zip` with all five entries (`unzip -t`: no errors), no console or page errors, and the button states above were recorded from the live DOM rather than eyeballed.

VTT-AI-Agent (commented):
## TL;DR

No actionable findings on PR #3090 at the current head (`6bd863f7`). The earlier event-loop blocking, eager asset inflation, worker fallback, and progress-feedback concerns have been addressed, and all current CI checks pass.

## Detailed findings

I re-checked the archive parsing and rebuilding paths, CRC-based asset naming, size guards, streamed server-side ZIP creation, built-in `fetch` substitutions, and Express middleware substitutions. The new server ZIP tests cover legacy JSZip input, stored and deflated round trips, Unicode filenames, empty entries, selective reads, CRC stability, and yielding during compression; the current Chrome/Firefox TestCafe, Jest/build, production-start, and CodeQL checks are all green.

The changed-file list contains none of the engine paths (`client/js/widgets/`, `client/js/compute.js`, or `client/js/serverstate.js`), and the client changes are confined to import/export and edit-mode download plumbing, so this does not alter how existing games execute or synchronize.

The `beta` label was already present; I left it in place because the PR is engine-eligible, passes CI, and is safe and useful for player testing of real import/export workloads.

VTT-AI-Agent:
Thanks for the re-check — nothing to change on my side.

Confirming the state you reviewed is still the head: `6bd863f7`, no new commits since, and all checks (Chrome/Firefox TestCafe, Jest/build, production start, CodeQL) are green. The diff still touches none of the engine paths (`client/js/widgets/`, `client/js/compute.js`, `client/js/serverstate.js`), so game execution and sync are unaffected.

I have not touched the `beta` label either way — that call belongs to a reviewer, not to me as the author.

ArnoldSmith86:
Give me a summary of count and size of all deps in main, in the base PR and in this PR.

VTT-AI-Agent:
Measured by installing each tree's committed lockfile clean (`npm ci --omit=dev --ignore-scripts`), counting with `npm ls --omit=dev --all --parseable` and sizing with `du -sb --apparent-size node_modules`.

### Production dependencies

| tree | packages | Δ | node_modules | Δ |
| --- | ---: | ---: | ---: | ---: |
| `main` | 153 | – | 25.5 MiB | – |
| #3086 (base PR) | 118 | −35 | 27.7 MiB | +2.1 MiB |
| #3090 (this PR) | 136 | −17 | 15.5 MiB | **−10.0 MiB** |
| both merged | 97 | −56 | 18.7 MiB | −6.8 MiB |

The "both merged" row is an actual merge of the two branches with a freshly resolved lockfile (there is no committed one), so treat it as ±1 package.

### Including devDependencies

`devDependencies` are untouched on both branches — jest, testcafe & co. are ~660 packages / ~161 MiB in every row, which is why the totals barely move:

| tree | packages | node_modules |
| --- | ---: | ---: |
| `main` | 814 | 186.5 MiB |
| #3086 | 805 | 190.4 MiB |
| #3090 | 804 | 176.7 MiB |
| both merged | 818 | 180.5 MiB |

### Why the numbers move the way they do

**#3086 — 56 packages out, 21 in, but +2.1 MiB.** The `@node-minify` wrappers were mostly *many small* packages (`glob`, `path-scurry` + its own `lru-cache`, `@isaacs/cliui` + 6 nested copies, `cross-spawn`, `no-case`/`param-case`/`camel-case`, …) totalling 6.6 MiB. What replaces them is fewer but individually larger: `html-minifier-terser` 5.0 MiB, `terser` 2.2 MiB, `acorn` 0.6 MiB — 8.7 MiB in 21 packages. So it is a big win in tree *shape* (and in what is actually maintained) and a small loss in bytes on disk.

**#3090 — 18 out, 1 in, −10.0 MiB.** Almost all of the size is one package: `web-streams-polyfill` (8.7 MiB), pulled in by `node-fetch` for a `fetch()` Node has shipped natively since 18. The rest:

| removed | size |
| --- | ---: |
| `web-streams-polyfill` | 8654 KiB |
| `pako` | 790 KiB |
| `jszip` | 784 KiB |
| `node-fetch` | 125 KiB |
| `readable-stream` | 110 KiB |
| 13 more (`lie`, `immediate`, `setimmediate`, `fetch-blob`, `formdata-polyfill`, `node-domexception`, `data-uri-to-buffer`, `core-util-is`, `isarray`, `safe-buffer`, `string_decoder`, `util-deprecate`, `process-nextick-args`) | 342 KiB |
| **added: `fflate`** | **794 KiB** |

(`fflate`'s 794 KiB is the whole published package — every build variant, ESM/CJS/UMD/browser plus types. The file the server actually reads is `umd/index.js`.)

**Browser payload**, which is the number players feel: `/scripts/jszip` served 95.3 KiB, `/scripts/fflate` serves 32.3 KiB — 63 KiB less on every page load.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-verified dependency swap — I checked out the branch, ran `npm test` (91 pass) and independently re-implemented `zipIndex()` against both a JSZip-written `.vtt` and a streamed `Zip.create()` output: CRCs and sizes match `CRC32.buf()` exactly, so asset naming really is preserved. The one thing I'd want changed before merge is the **nested-entry read in `server/fileloader.mjs:82`**: fflate pre-allocates a buffer from the *declared* uncompressed size, and that call path has no size guard, so a small crafted `.vttc` can make the server allocate arbitrarily much — a class of exposure JSZip's streaming inflate didn't have.

## Findings

### 1. Unguarded nested entry: fflate trusts the declared uncompressed size (`server/fileloader.mjs:82`)

```js
if(filename.match(/\.(vtt|pcio)$/))
  states[filename] = await readVariantsFromBuffer((await Zip.read(buffer, [ filename ]))[filename]);
```

Every *other* `Zip.read`/`readString` call in this file is preceded by a size check against `entries[filename]` (20 MiB for variants, 10 MiB for assets, 2 MiB for pcio userassets) — this one is not. That mattered less with JSZip, which inflates the actual stream and therefore can only produce as much data as the compressed bytes allow. fflate is different: in `unzip`, an entry with `originalSize >= 524288` and a good compression ratio goes down

```js
term.push(inflate(infl, { size: su }, cbl));   // su = uncompressed size from the central directory
```

i.e. it allocates `su` bytes up front, from an attacker-controlled header field, before a single byte is inflated. A `.vttc` of a few hundred KB whose nested `game.vtt` entry declares `originalSize = 2000000000` gets a 2 GB allocation (or a `RangeError`, depending on the value) — on a single-process server that is a crash, not a failed request. Same one-line shape as the guards right below it:

```js
if(entries[filename] >= 20971520) throw new Logging.UserError(403, `${filename} is bigger than 20 MiB.`);
```

Since the PR description explicitly leans on "nothing is decompressed, which is what keeps the guards from having to unpack a zip bomb first", closing this last hole fits the same argument.

### 2. `zipIndex()` parses the central directory without validating it (`client/js/overlays/states.js:59-77`)

The walk assumes every record is a plain 46-byte central-directory header:

```js
let offset = view.getUint32(end + 16, true);
for(let i=view.getUint16(end + 8, true); i>0; --i) {
  const nameLength = view.getUint16(offset + 28, true);
  ...
}
```

Two gaps versus fflate's own `zh()`, which this is otherwise a faithful copy of (I diffed the name-decoding: the bit-11 latin1/UTF-8 handling is identical, good):

- **No ZIP64 handling.** fflate's `zh()` calls `z64hs()` to pull the real sizes/offset out of the ZIP64 extra field when the 32-bit fields are `0xFFFFFFFF`; `zipIndex` takes the sentinel at face value. On a ZIP64 archive the CD offset read at `end+16` is `0xFFFFFFFF`, so the loop reads past the buffer and silently produces junk names and sizes rather than throwing.
- **No `0x02014b50` signature check per record.** Any drift (ZIP64, a prepended-data archive, a malformed extra field) yields a garbage `entries` map instead of an error, and because the `try` around it only catches throws, the user gets a nonsensical asset map / silently-wrong dedup rather than the "is not a valid VTT…" alert.

Both are cheap: `if(view.getUint32(offset, true) != 0x02014b50) throw …`, plus a bail-out when the CD offset or a size reads `0xFFFFFFFF`. Practically low-probability today (JSZip never emitted ZIP64 and entries are capped well under 4 GB), but failing loudly is much better than failing quietly for something that feeds asset filenames.

### 3. Rebuild path holds ~3 full copies in the tab (`client/js/overlays/states.js:221-223`)

```js
const files = await unzipBuffer(buffer, name=>!removedFiles[name] && !name.match(/\/$/));
files['asset-map.json'] = fflate.strToU8(JSON.stringify(assets));
blob = await zipBlob(files, total-removed < 5);
```

Peak memory here is the source `buffer` + every kept entry fully inflated + the complete output `Uint8Array` from `fflate.zip`, all live at once before the `Blob` is constructed. JSZip's `generateAsync` processed entries one at a time into a Blob the browser could spill to disk. Since `.vtt` payloads are mostly already-compressed images, uncompressed ≈ compressed, so this is roughly 3× file size resident instead of ~1×. It's mitigated by the fact that a rebuild only happens when >half the assets are dropped, but for the "4 known, 3 new, multi-megabyte entries" case scaled up to a few hundred MB it's a plausible tab OOM where the old code survived. Worth at least a note in the code; ideally the rebuild would stream entry-by-entry through `fflate.Zip` the way `server/zip.mjs create()` already does.

### 4. The worker fallback also swallows genuine corruption (`client/js/overlays/states.js:36-53`)

```js
try {
  return await new Promise((resolve, reject)=>fflate.unzip(buffer, options, (e, files)=>e ? reject(e) : resolve(files)));
} catch(e) {
  return fflate.unzipSync(buffer, options);
}
```

The comment says this is for a CSP that forbids blob-URL workers, but the `catch` is unconditional, so `invalid zip data` / `unknown compression type` also trigger a full synchronous re-run that inflates everything again on the main thread and only then rethrows. On a large damaged upload that's a multi-second freeze before the error surfaces. Gating on the error (or on a one-shot "workers don't work here" flag set the first time the sync path is needed) keeps the intended behaviour without the double work. Same applies to `zipBlob()`.

### 5. Per-asset `Zip.read` re-walks the whole archive (`server/fileloader.mjs:190`)

`Zip.read(buffer, [ filename ])` is called once per asset, and each call runs fflate's loop over *every* central-directory record (`zh()` + `slzh()` per entry) plus, for entries above the 512 KB threshold, spawns a fresh `worker_threads` worker. For a game with a few hundred assets that's O(n²) header parses and n sequential worker spawns. Not wrong — and the event loop stays free, which is the point — but if asset-heavy imports get noticeably slower, batching (`Zip.read(buffer, names.slice(i, i+20))`) would recover most of it without giving up the memory bound.

### 6. Test coverage: the TTS zip path is unverifiable, and `Zip.read` of a *missing* name is unguarded

`server/ttsimport.mjs:264-270` was converted to `Zip.list` / `Zip.readString`, but as the PR itself notes, `FileLoader` calls `TTS.fromZip` while the module exports `fromZIP` — so that conversion cannot have been exercised end-to-end by any of the verification listed. Fine to leave the name bug out of scope, but it's worth saying plainly that the TTS `.zip` branch of this migration is untested rather than tested-and-passing.

Relatedly, `Zip.readString(buffer, name)` does `strFromU8(files[name])` with no check — if the name isn't in the archive it throws a `TypeError` deep in fflate rather than something diagnosable. All current call sites happen to check `entries[...]` first (or wrap in `try{}catch{}`, as `pcioimport.mjs` does for `asset-map.json`), so this is defensive only, but it's a one-liner in a module that is now the single choke point for every zip read.

### 7. Smaller things

- `jest.config.mjs:12` — `customExportConditions: [ "node" ]` applies to *every* package the jsdom tests import, not just fflate. The comment says so, which is good; just flagging that a future client test needing a browser build of some package will fail confusingly.
- `client/js/editor/sidebar/assets.js:112` — `for(const [ i, assetObj ] of Object.entries(assets))` on an array gives string keys, hence the `+i` coercions; `assets.entries()` gives numbers directly.
- `server/zip.mjs:1` — `import { setImmediate } from 'timers'` is redundant, `setImmediate` is a Node global.
- `server/ttsimport.mjs:15` still has `fs.writeFileSync('/tmp/out', buffer)` immediately below the changed `arrayBuffer()` line (and `/tmp/tts.json` at :265). Pre-existing debug leftovers, but they're world-writable shared paths on a multi-tenant host — cheap to drop while you're in the file.

### What I checked and found correct

- `zipIndex()` output against the JSZip-written fixture and against `Zip.create()` output with and without deflate: names, signed CRC-32 and sizes are byte-identical to `CRC32.buf()` / the fixture, including the data-descriptor (bit 3) entries the streaming writer produces. Asset filenames really are preserved.
- `Zip.list()` genuinely decompresses nothing — fflate evaluates `opts.filter` *before* dispatching to `inflateSync`/`inflate`, and errors on unknown compression types only for entries that pass the filter, so the oversize guards fire on unopened entries as claimed.
- `create()`'s chunk loop is correct at the boundaries: `offset<data.length || !offset` emits exactly one final chunk for a zero-length entry and doesn't emit a spurious extra chunk when `data.length` is an exact multiple of `chunkSize`. `Zip.create({})` produces a valid 22-byte empty archive.
- `express.json`/`express.raw` in express 5 are the same body-parser middleware that was previously imported directly, so dropping the undeclared dependency is behaviour-neutral.
- Server-side CRC is now computed from content rather than read from the header — strictly more robust, and it also fixes the old edge case where an asset whose CRC is literally `0` was skipped by the `file._data.crc32` truthiness check.

One thing to keep an eye on that isn't a diff bug: Node's built-in `fetch` (undici) applies default `headersTimeout`/`bodyTimeout` of 300 s, which `node-fetch` did not. That's fine for `/assetcheck` and asset downloads, but a `moveServer` PUT of a large state to a slow peer, or a very large public-library link, could now abort where it previously just took a while. Worth remembering if cross-server moves start failing.

I added no label: `beta` is already on this PR and I left it, which is right — no engine file (`client/js/widgets/`, `client/js/compute.js`, `client/js/serverstate.js`) is touched, so the PR is beta-eligible, and the migration is well enough verified that it is safe for players to exercise, with finding #1 being a hardening gap rather than something that breaks normal use.

VTT-AI-Agent:
Thanks - all six code findings looked at, four fixed in 8ff8697b, plus the two small cleanups.

**1. Unguarded nested entry (`server/fileloader.mjs:82`) - fixed.** You're right, and it reproduces: I built a 2 KB `.vttc` whose deflated `game.vtt` entry declares a 4294967290-byte uncompressed size, and `Zip.read` happily came back with the correct 690 KB of data after fflate had allocated a 4 GB output buffer for it (`process.memoryUsage().external` = 4098 MB; RSS stays low only because the pages are never touched). Both fflate branches presize from the header - `inflateSync(infl, { out: new u8(su) })` and `inflate(infl, { size: su })` - the sync one is bounded by `sc > 0.8 * su`, the worker one is not bounded at all. With the guard the same file throws `403 game.vtt is bigger than 100 MiB.` and nothing is allocated. 100 MiB is far above anything real (the largest public library game is ~17 MB) and below anything that hurts.

**2. `zipIndex()` validation - fixed.** Added the `0x02014b50` check per record, plus a bail-out when the central-directory offset or an entry size reads `0xffffffff`, so ZIP64 says so instead of walking off the end. Verified in a browser: a `.vtt` whose second central-directory signature I overwrote now gets the "is not a valid VTT, VTTC, VTTS or PCIO file." alert instead of a silently wrong asset map, and a normal `.vtt`/`.vttc` still uploads.

**3. Rebuild path memory - noted in the code, not restructured.** Peak is as you describe. Streaming the rebuild through `fflate.Zip` in the browser is a bigger change than I want to make in a dependency swap (the entries have to be pushed into a growing `Blob` to get the benefit at all), so I left a comment at the spot saying the three copies are live there and that it can be streamed if it ever hurts. Happy to do it properly in a follow-up if you'd rather have it now.

**4. Worker fallback - fixed differently.** Rather than classifying the error (fflate's errors carry a numeric `.code`, but so does a `DOMException`, so that check is not as clean as it looks), whether blob-URL workers can be created is now probed once and cached, and both helpers pick their path up front. Genuine `invalid zip data` now propagates on the first try instead of inflating everything again synchronously.

**5. Per-asset `Zip.read` - left as is.** Deliberate: the 10 MiB guard is per asset and only holds if exactly one asset is unpacked at a time, so batching 20 would raise the worst case to 200 MiB resident. The repeated work is central-directory header parsing, which is cheap next to the inflate that follows it. If asset-heavy imports do turn out slow, that's the knob, but I'd rather not trade the memory bound away speculatively.

**6. TTS zip branch and `readString` - agreed, and stated plainly:** the `.zip` (Workshop) branch of `ttsimport.mjs` is **untested** in this PR. `FileLoader` calls `TTS.fromZip` while the module exports `fromZIP`, so that path is unreachable today and none of my verification touched it; the conversion is a mechanical `JSZip.loadAsync` -> `Zip.list`/`Zip.readString` swap, but I'm not claiming it as tested-and-passing. The name bug stays out of scope here. `Zip.readString` now throws `<name> is not in the zip file.` for a missing entry, with a test.

**7. Smaller things:**
- `import { setImmediate } from 'timers'` is **not** redundant - dropping it makes two `zip.test.js` tests fail with `ReferenceError: setImmediate is not defined`, because the server modules are tested in the jsdom environment. Kept, with a comment saying why.
- `assets.js`: switched to `assets.entries()`, coercions gone.
- `jest.config.mjs` `customExportConditions`: agreed it's global, the comment already says so - left alone.
- Removed both `/tmp` debug leftovers in `ttsimport.mjs` (`/tmp/out` on every image-size probe, `/tmp/tts.json` on every TTS link import) and the now-unused `fs` import. Pre-existing and unrelated to the swap, so say the word if you'd rather have them in their own commit.

On undici's 300 s `headersTimeout`/`bodyTimeout`: noted, nothing changed for it. Worth watching for `moveServer` to a slow peer.

`npm test` passes (92 now). Re-verified in a browser on :8356 after the changes: `.vtt` upload, `.vttc` collection upload (both taking the rebuild path, 12/12 and 5/5 assets already on the server), a non-zip file and a damaged-index file both alerting correctly.

ArnoldSmith86 (approved):

PR #3092, opened 2026-08-03
Fix flaky Reversi TestCafe test: drops racing with in-flight drag moves

VTT-AI-Agent:
`Public library: Reversi (variant 0)` failed roughly **50% of the time** locally (10 runs on `main`: 5 hash mismatches). It is the only public-library test that drags pieces instead of clicking buttons, which is why it is the one that shows this — but the two causes behind it both hit real players as "my piece landed next to where I dropped it" or "the board just ate my piece".

## 1. Mouse events were not serialized

Mouse events are handled by an `async` `inputHandler` that is *not* serialized: `window.addEventListener` fires it for every `mousemove` and for `mouseup`, and each invocation `await`s its way through `move()` (position, `checkParent()`, hover-target hit testing, delta sending). So several handlers are in flight at once and can interleave.

`Widget.moveEnd()` drops the widget onto `this.hoverTarget`, which is whatever the *last fully processed* `move()` computed. Two things go wrong under load:

* `mouseup` can be handled while the final `mousemove` is still being processed, so the drop uses a hover target from somewhere along the drag path.
* A `mousemove` handler that is still pending when `moveEnd()` has already run then calls `move()` afterwards, dragging the piece back out of the square it was just dropped into.

Instrumented failing run:

```
MOVE zpiece19 -> sq35 ... now= {"pieceParent":"w_sq34"}
```

**Fix** (`client/js/mousehandling.js`): queue the per-drag `moveStart()`/`move()` calls on a promise chain stored on the widget's `mouseStatus`, and `await` that chain in the `mouseup` branch before calling `moveEnd()`. Moves are applied in event order and the drop always happens after the last of them. A move that a newer `mousemove` already superseded is skipped, the chain can never stay rejected, `mouseStatus` is deleted synchronously when the drop starts (so a `mousemove` delivered during a slow drop cannot re-enter the drag), and `batchEnd()` runs in a `finally` so a throwing routine can no longer leak a batch and stop the client from syncing.

Skipping a superseded move skips it whole, so the hover side effects it would have had — the `droptarget` highlight, the drop shadow entering and leaving a holder, and therefore that holder's `onEnter`/`enterRoutine` — do not run for it either. The last move of a drag always runs, so the end state is unchanged; a drag that sweeps quickly across a holder can now pass over it without that holder seeing the shadow, which is the same thing a lower pointer sampling rate already produced. Games that hang routines off a holder being hovered are the ones that can notice.

## 2. The drop was resolved against the widget's *rendered* position, which lags a whole mouse event behind

Found while addressing the UI/UX review, which showed a piece bouncing back into the holder on a coarse drag — on `main` as well.

`move()` hit tests the drop target with `document.elementsFromPoint()` at `center(this.domElement)`, i.e. it asks the DOM where the widget is. But every mouse event runs inside `batchStart()`/`batchEnd()` and a batched delta only reaches the DOM when the batch ends (`sendDelta()` does nothing while `batchDepth` is non-zero). So the hit test used the position of the **previous** mouse event:

```
DRAGDBG zpiece15 coord 719,297  center 671,283  hover sq22
DRAGDBG zpiece15 coord 788,314  center 738,299  hover sq22   <- last move: the piece is really at 805,315, on sq23
```

With fine pointer sampling the error stays inside the same holder; once a drag covers more than half a holder between two events it is a different square — the drop then goes to the square the piece has already left, which in Reversi is an illegal move and sends the piece back into the holder.

**Fix** (`client/js/widgets/widget.js`): flush the pending delta before the hit test when the DOM transform is stale — the same fix (and the same reason) `line.js` already uses for its stop geometry.

On top of that, a `mouseup` can arrive at coordinates no `mousemove` ever reported (a fast flick ends that way), and the drop was then resolved for the last reported position rather than for the one the player released over. `moveEnd()` now applies the release coordinates with a `move()` first when they differ, so the drop, the line stop and the final position all belong to where the button was released.

| mousemove events per drag | ≈ px per event | before | after |
|---|---|---|---|
| 25 | ~18px | ✅ `sq23` | ✅ `sq23` |
| 8 | ~69px | ❌ back to `holder0` | ✅ `sq23` |
| 5 | ~111px | ❌ | ✅ |
| 3 / 2 / 1 | ≥184px | ❌ | ✅ |
| release over `sq23` after the last mousemove was over `sq22` | – | ❌ resolved against `sq22` → `holder0` | ✅ `sq23` |

Runnable reproductions: [repro-drag.cjs](https://agent.virtualtabletop.io/reports/pr/1533690562639695985/repro/repro-drag.cjs), [repro-flick.cjs](https://agent.virtualtabletop.io/reports/pr/1533690562639695985/repro/repro-flick.cjs).

## Test changes

* `waitForStableState()` in `test-util.js` — polls the room state until it stops changing, and throws instead of failing silently. `publicLibraryButtons` calls it before each drag so a move is never made while the game is still evaluating the previous one, and before each drop assertion so a rejected drop has actually been evaluated.
* Drag steps are now `{ from, to, sticks }` entries instead of `Selector` thunks, and the test asserts right after the drag whether the piece stayed on the target — so a bad drag is reported where it happened instead of as a mystery hash mismatch at the end. The Reversi list documents that `zpiece72 -> sq12` is deliberately an illegal move.
* Those drags now run at `speed:1`, i.e. with the coarse pointer sampling that triggers cause 2. Without the fix the Reversi test fails 2 out of 3 runs; with it, 5 of 5 pass (and it is ~3× faster).

## Verification

Ran locally (Chromium headless):

* `npm test`: 90/90.
* Reversi: 12/12 with the ordering fix (3/6 and 5/10 on `main`), and 5/5 at `speed:1` with both fixes.
* Full `publiclibrary-1..4`, `editor.js`, `routines.js`: 39/39 pass — no reference hash is changed by the flush or by the release-coordinate move.

## beta label

`client/js/mousehandling.js` and `Widget.move()/moveEnd()` are engine code and this changes the order in which drag events are applied and which holder a drop resolves to, i.e. real engine behaviour, so I have added the `engine` label. It is a strict tightening with no state-format change and the full TestCafe suite passes, so it should be safe for beta once a reviewer agrees.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3092/pr-test (or any other room on that server)

### DIFF
--- a/client/js/mousehandling.js
+++ b/client/js/mousehandling.js
@@ -105,7 +105,7 @@ async function inputHandler(name, e) {
         if (movable) {
           ms.localAnchor = ms.moveTarget.coordLocalFromCoordClient({x: coords.clientX, y: coords.clientY});
         }
-      } else if(name == 'mouseup' || (name == 'touchend' || name == 'touchcancel') && mouseStatus[target.id]) {
+      } else if((name == 'mouseup' || name == 'touchend' || name == 'touchcancel') && mouseStatus[target.id]) {
         const ms = mouseStatus[target.id];
         // End the drag synchronously, before the first await below: a mousemove that
         // is delivered while the drop is still being processed would otherwise queue
@@ -179,8 +179,10 @@ async function inputHandler(name, e) {
             await ms.moveTarget.move(coords, ms.localAnchor);
           }).catch(error => {
             // keep the chain resolvable - a rejected one would make every later move
-            // and the drop in the mouseup branch above fail as well
-            console.error('Error while dragging.', error);
+            // and the drop in the mouseup branch above fail as well - but still let
+            // the error reach the client error reporter in tracing.js, which is where
+            // it would have ended up as an unhandled rejection before the chain
+            setTimeout(_=>{ throw error; });
           });
           await ms.dragChain;
         }
--- a/client/js/mousehandling.js
+++ b/client/js/mousehandling.js
@@ -107,6 +107,9 @@ async function inputHandler(name, e) {
       const pixelsMoved = ms.coords ? Math.abs(ms.coords.x - ms.downCoords.x) + Math.abs(ms.coords.y - ms.downCoords.y) : 0;
       if(ms.status != 'initial' && ms.moveTarget) {
         setDeltaCause(`${playerName} dragged ${widget.id}`);
+        // let every mousemove that is still being processed finish first so that the
+        // drop happens after the last one instead of racing with it
+        await ms.dragChain;
         await ms.moveTarget.moveEnd(coords, ms.localAnchor);
       }
       if(ms.status == 'initial' || timeSinceStart < 250 && pixelsMoved < 10) {
@@ -142,16 +145,24 @@ async function inputHandler(name, e) {
       }
       delete mouseStatus[target.id];
     } else if(name == 'mousemove' || name == 'touchmove' && mouseStatus[target.id]) {
+      const ms = mouseStatus[target.id];
       setDeltaCause(`${playerName} dragged ${widget.id}`);
-      if(mouseStatus[target.id].status == 'initial') {
-        mouseStatus[target.id].status = 'moving';
-        if(mouseStatus[target.id].moveTarget)
-          await mouseStatus[target.id].moveTarget.moveStart();
-      }
-      mouseStatus[target.id].coords = coords;
-      if(mouseStatus[target.id].moveTarget) {
+      const isFirstMove = ms.status == 'initial';
+      if(isFirstMove)
+        ms.status = 'moving';
+      ms.coords = coords;
+      if(ms.moveTarget) {
         setDeltaCause(`${playerName} dragged ${widget.id}`);
-        await mouseStatus[target.id].moveTarget.move(coords, mouseStatus[target.id].localAnchor);
+        // Mouse events are handled asynchronously, so several of them can be in
+        // flight at once. Queue the moves instead of running them in parallel so
+        // the widget always ends up where the most recent event put it - and so
+        // that the drop in the mouseup branch above happens after all of them.
+        ms.dragChain = Promise.resolve(ms.dragChain).then(async _ => {
+          if(isFirstMove)
+            await ms.moveTarget.moveStart();
+          await ms.moveTarget.move(coords, ms.localAnchor);
+        });
+        await ms.dragChain;
       }
     }
     batchEnd();
--- a/client/js/mousehandling.js
+++ b/client/js/mousehandling.js
@@ -71,101 +71,123 @@ async function inputHandler(name, e) {
       return;
     }
     batchStart();
-    if(!edit && (!jeEnabled || !e.ctrlKey) && widget.passthroughMouse) {
-      if(name == 'mousedown' || name == 'touchstart') {
-        await widget.mouseRaw('down', coords);
-      } else if (name == 'mouseup' || name == 'touchend' || name == 'touchcancel') {
-        await widget.mouseRaw('up', coords);
-      } else if (name == 'mousemove' || name == 'touchmove') {
-        await widget.mouseRaw('move', coords);
-      }
-    } else if(name == 'mousedown' || name == 'touchstart') {
-      mouseStatus[target.id] = {
-        status: 'initial',
-        start: new Date(),
-        downCoords: coords,
-        moveTarget: widget
-      };
-      const ms = mouseStatus[target.id];
-      let movable = ms.moveTarget.get(editMovable ? 'movableInEdit' : 'movable');
-      while (ms.moveTarget && !movable) {
-        let parent = ms.moveTarget.get('parent');
-        if(parent && widgets.has(parent)) {
-          ms.moveTarget = widgets.get(parent);
-          movable = ms.moveTarget.get(editMovable ? 'movableInEdit' : 'movable');
-        } else {
-          ms.moveTarget = null;
-          movable = false;
+    // batchEnd() has to run even if a routine triggered by the drop or click below
+    // throws: batchStart() increments batchDepth and sendDelta() only sends anything
+    // while that is 0, so a leaked batch stops this client from syncing altogether.
+    try {
+      if(!edit && (!jeEnabled || !e.ctrlKey) && widget.passthroughMouse) {
+        if(name == 'mousedown' || name == 'touchstart') {
+          await widget.mouseRaw('down', coords);
+        } else if (name == 'mouseup' || name == 'touchend' || name == 'touchcancel') {
+          await widget.mouseRaw('up', coords);
+        } else if (name == 'mousemove' || name == 'touchmove') {
+          await widget.mouseRaw('move', coords);
         }
-      }
-      if (movable) {
-        ms.localAnchor = ms.moveTarget.coordLocalFromCoordClient({x: coords.clientX, y: coords.clientY});
-      }
-    } else if(name == 'mouseup' || (name == 'touchend' || name == 'touchcancel') && mouseStatus[target.id]) {
-      const ms = mouseStatus[target.id];
-      const timeSinceStart = +new Date() - ms.start;
-      const pixelsMoved = ms.coords ? Math.abs(ms.coords.x - ms.downCoords.x) + Math.abs(ms.coords.y - ms.downCoords.y) : 0;
-      if(ms.status != 'initial' && ms.moveTarget) {
-        setDeltaCause(`${playerName} dragged ${widget.id}`);
-        // let every mousemove that is still being processed finish first so that the
-        // drop happens after the last one instead of racing with it
-        await ms.dragChain;
-        await ms.moveTarget.moveEnd(coords, ms.localAnchor);
-      }
-      if(ms.status == 'initial' || timeSinceStart < 250 && pixelsMoved < 10) {
-        let editClickHandled = false;
-        if(edit && !isMiddleMouseButton)
-          editClickHandled = await editClick(widget, e.button);
-        else if(jeEnabled && !isMiddleMouseButton)
-          editClickHandled = await jeClick(widget, e);
-
-        if(!editClickHandled) {
-          if(!target.classList.contains('longtouch')) {
-            if(!widget.get('doubleClickRoutine')) {
-              setDeltaCause(`${playerName} clicked ${widget.id}`);
-              await widget.click();
-            } else if(doubleClickTimeout) {
-              clearTimeout(doubleClickTimeout);
-              doubleClickTimeout = null;
-              setDeltaCause(`${playerName} double clicked ${widget.id}`);
-              await widget.doubleClick();
-            } else {
-              doubleClickTimeout = setTimeout(async () => {
-                doubleClickTimeout = null;
-                batchStart();
-                setDeltaCause(`${playerName} clicked ${widget.id}`);
-                await widget.click();
-                batchEnd();
-              }, 350);
-            }
+      } else if(name == 'mousedown' || name == 'touchstart') {
+        mouseStatus[target.id] = {
+          status: 'initial',
+          start: new Date(),
+          downCoords: coords,
+          moveTarget: widget
+        };
+        const ms = mouseStatus[target.id];
+        let movable = ms.moveTarget.get(editMovable ? 'movableInEdit' : 'movable');
+        while (ms.moveTarget && !movable) {
+          let parent = ms.moveTarget.get('parent');
+          if(parent && widgets.has(parent)) {
+            ms.moveTarget = widgets.get(parent);
+            movable = ms.moveTarget.get(editMovable ? 'movableInEdit' : 'movable');
           } else {
-            widget.domElement.classList.remove('longtouch');
+            ms.moveTarget = null;
+            movable = false;
           }
         }
-      }
-      delete mouseStatus[target.id];
-    } else if(name == 'mousemove' || name == 'touchmove' && mouseStatus[target.id]) {
-      const ms = mouseStatus[target.id];
-      setDeltaCause(`${playerName} dragged ${widget.id}`);
-      const isFirstMove = ms.status == 'initial';
-      if(isFirstMove)
-        ms.status = 'moving';
-      ms.coords = coords;
-      if(ms.moveTarget) {
+        if (movable) {
+          ms.localAnchor = ms.moveTarget.coordLocalFromCoordClient({x: coords.clientX, y: coords.clientY});
+        }
+      } else if(name == 'mouseup' || (name == 'touchend' || name == 'touchcancel') && mouseStatus[target.id]) {
+        const ms = mouseStatus[target.id];
+        // End the drag synchronously, before the first await below: a mousemove that
+        // is delivered while the drop is still being processed would otherwise queue
+        // another move and drag the widget back out of where it was just dropped.
+        delete mouseStatus[target.id];
+        const timeSinceStart = +new Date() - ms.start;
+        const pixelsMoved = ms.coords ? Math.abs(ms.coords.x - ms.downCoords.x) + Math.abs(ms.coords.y - ms.downCoords.y) : 0;
+        if(ms.status != 'initial' && ms.moveTarget) {
+          setDeltaCause(`${playerName} dragged ${widget.id}`);
+          // let every mousemove that is still being processed finish first so that the
+          // drop happens after the last one instead of racing with it
+          await ms.dragChain;
+          await ms.moveTarget.moveEnd(coords, ms.localAnchor);
+        }
+        if(ms.status == 'initial' || timeSinceStart < 250 && pixelsMoved < 10) {
+          let editClickHandled = false;
+          if(edit && !isMiddleMouseButton)
+            editClickHandled = await editClick(widget, e.button);
+          else if(jeEnabled && !isMiddleMouseButton)
+            editClickHandled = await jeClick(widget, e);
+
+          if(!editClickHandled) {
+            if(!target.classList.contains('longtouch')) {
+              if(!widget.get('doubleClickRoutine')) {
+                setDeltaCause(`${playerName} clicked ${widget.id}`);
+                await widget.click();
+              } else if(doubleClickTimeout) {
+                clearTimeout(doubleClickTimeout);
+                doubleClickTimeout = null;
+                setDeltaCause(`${playerName} double clicked ${widget.id}`);
+                await widget.doubleClick();
+              } else {
+                doubleClickTimeout = setTimeout(async () => {
+                  doubleClickTimeout = null;
+                  batchStart();
+                  try {
+                    setDeltaCause(`${playerName} clicked ${widget.id}`);
+                    await widget.click();
+                  } finally {
+                    batchEnd();
+                  }
+                }, 350);
+              }
+            } else {
+              widget.domElement.classList.remove('longtouch');
+            }
+          }
+        }
+      } else if((name == 'mousemove' || name == 'touchmove') && mouseStatus[target.id]) {
+        const ms = mouseStatus[target.id];
         setDeltaCause(`${playerName} dragged ${widget.id}`);
-        // Mouse events are handled asynchronously, so several of them can be in
-        // flight at once. Queue the moves instead of running them in parallel so
-        // the widget always ends up where the most recent event put it - and so
-        // that the drop in the mouseup branch above happens after all of them.
-        ms.dragChain = Promise.resolve(ms.dragChain).then(async _ => {
-          if(isFirstMove)
-            await ms.moveTarget.moveStart();
-          await ms.moveTarget.move(coords, ms.localAnchor);
-        });
-        await ms.dragChain;
+        const isFirstMove = ms.status == 'initial';
+        if(isFirstMove)
+          ms.status = 'moving';
+        ms.coords = coords;
+        if(ms.moveTarget) {
+          setDeltaCause(`${playerName} dragged ${widget.id}`);
+          // Mouse events are handled asynchronously, so several of them can be in
+          // flight at once. Queue the moves instead of running them in parallel so
+          // the widget always ends up where the most recent event put it - and so
+          // that the drop in the mouseup branch above happens after all of them.
+          ms.dragChain = Promise.resolve(ms.dragChain).then(async _ => {
+            // a move that a newer mousemove already replaced does not need to run:
+            // that one will put the widget, its hover target and its parent where
+            // the cursor is now. Without this the queue can grow under load and the
+            // widget visibly lags behind the cursor.
+            if(!isFirstMove && ms.coords !== coords)
+              return;
+            if(isFirstMove)
+              await ms.moveTarget.moveStart();
+            await ms.moveTarget.move(coords, ms.localAnchor);
+          }).catch(error => {
+            // keep the chain resolvable - a rejected one would make every later move
+            // and the drop in the mouseup branch above fail as well
+            console.error('Error while dragging.', error);
+          });
+          await ms.dragChain;
+        }
       }
+    } finally {
+      batchEnd();
     }
-    batchEnd();
   }
 
   if(name == 'mouseup')
--- a/client/js/tracing.js
+++ b/client/js/tracing.js
@@ -47,7 +47,7 @@ onLoad(function() {
       error: String(error.message) + '\n' + String(error.stack),
       undoProtocol,
       delta,
-      mouseStatus: Object.fromEntries(Object.entries(mouseStatus).map(([id, ms]) => [id, {...ms, moveTarget: ms.moveTarget ? ms.moveTarget.get('id') : null}])),
+      mouseStatus: Object.fromEntries(Object.entries(mouseStatus).map(([id, ms]) => [id, {...ms, moveTarget: ms.moveTarget ? ms.moveTarget.get('id') : null, dragChain: undefined}])),
       mouseTarget: mouseTarget && mouseTarget.id ? unescapeID(mouseTarget.id.slice(2)) : null,
       jeLoggingData: typeof jeLoggingRoutineGetData == 'function' ? jeLoggingRoutineGetData() : null,
       lastExecutedOperation,
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -2494,6 +2494,7 @@ export class Widget extends StateManaged {
 
     await this.bringToFront();
     await this.set('dragging', playerName);
+    delete this.lastMoveCoord;
 
     // Lines that take a widget dropped onto their path as a stop. Collected once
     // like the drop targets below, but not restricted to widgets that can be
@@ -2537,6 +2538,8 @@ export class Widget extends StateManaged {
     if(tracingEnabled)
       sendTraceEvent('move', { id: this.get('id'), coordGlobal, localAnchor, newX: newCoord.x, newY: newCoord.y });
 
+    this.lastMoveCoord = coordGlobal;
+
     await this.setPosition(newCoord.x, newCoord.y, this.get('z'));
     await this.snapToGrid();
 
@@ -2544,6 +2547,15 @@ export class Widget extends StateManaged {
       await this.checkParent();
 
       const lastHoverTarget = this.hoverTarget;
+      // The hit test below asks the DOM where this widget is, but the delta that
+      // carries the position set above only reaches the DOM when the batch around
+      // the mouse event ends. Without the flush the drop target is computed for the
+      // position of the *previous* mouse event - so a drag that moves further than
+      // half a holder between two events resolves the drop against the square the
+      // widget has already left, and the piece lands next to where it was dropped
+      // (or the game rejects the move and sends it back).
+      if(this.domElement.style.transform != this.cssTransform())
+        flushDelta();
       const myCenter = center(this.domElement);
       const myMinDim = Math.min(this.get('width'), this.get('height')) * this.get('_absoluteScale');
       this.hoverTarget = null;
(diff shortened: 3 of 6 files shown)


VTT-AI-Agent (commented):
## TL;DR

The diagnosis is right and the queueing fix is the correct shape — I reproduced the green result locally (Reversi 5/5 passes on this branch, `npm test` 90/90). Two things to look at before merge: a rejected `dragChain` now poisons the drop *and* leaves `batchDepth` stuck (deltas stop being sent for the rest of the session), and the mirror-image race — a `mousemove` delivered **after** `mouseup`, while `moveEnd()` is still awaiting — is still open.

## Findings

### 1. A failing `move()` now takes the drop and the delta batching down with it — `client/js/mousehandling.js:160`

```js
ms.dragChain = Promise.resolve(ms.dragChain).then(async _ => { ... });
await ms.dragChain;
```

`.then(onFulfilled)` propagates rejection. If any queued `moveStart()`/`move()` throws (a bad `dragLimit`, a widget removed mid-drag, a routine error in `checkParent()` → `dispenseCard`), `ms.dragChain` stays rejected forever:

* every subsequent `mousemove` skips its callback and just re-throws,
* the `await ms.dragChain` you added at `:112` throws **before** `moveEnd()`, so the drop never happens, `delete mouseStatus[target.id]` at `:146` never runs, and `batchEnd()` at `:168` is never reached.

That last one is the bad part: `sendDelta()` only fires when `batchDepth` is 0 (`client/js/serverstate.js:606`), and nothing ever resets it (except `flushDelta`/reconnect). A leaked `batchStart()` means the client silently stops sending *any* delta to the server for the rest of the session — the player keeps playing against a state nobody else sees. Before this PR a throwing `move()` broke only that one handler; `mouseup` still dropped, cleaned up and balanced its own batch.

Cheapest fix is to keep the chain permanently resolvable, e.g.

```js
ms.dragChain = Promise.resolve(ms.dragChain).then(async _ => {
  try { ... } catch(e) { console.error(e); }
});
```

and/or wrap the `mouseup` body in `try/finally` so `delete mouseStatus[...]` + `batchEnd()` always run.

### 2. The symmetric race after `mouseup` is still open — `client/js/mousehandling.js:104-146`

The `mousemove` listener is attached to `window` permanently (`:204`) and is never removed for the duration of a drag. In the `mouseup` branch, `mouseStatus[target.id]` is only deleted at `:146` and `mouseTarget` only nulled after `batchEnd()` — i.e. **after** `await ms.dragChain` and `await ms.moveEnd(...)`, which is exactly the slow part (`moveToHolder`, `updatePiles`, any routine the drop triggers).

So a `mousemove` delivered in that window still finds a live `mouseStatus` entry, appends another `move()` to the same `ms.dragChain` and drags the piece back out of the square it was just dropped into — the PR description's failure mode #2, just moved from before `mouseup` to after it. On a real client that's plausible (release the button, keep moving the mouse while a slow drop routine runs); it's the one case the test can't show you, since TestCafe emits no `mousemove` after the drop.

Suggestion: mark the drag as finishing *synchronously*, before the first `await` in the `mouseup` branch — e.g. `ms.status = 'dropping'` — and bail out of the `mousemove` branch when you see it.

Note if you do add such a guard: `else if(name == 'mousemove' || name == 'touchmove' && mouseStatus[target.id])` at `:147` binds as `mousemove || (touchmove && ms)`, so on `mousemove` there is no `mouseStatus` guard at all and the new `ms.status` deref at `:150` would throw on an undefined `ms`. Same as before your change, but worth tightening while you're in there.

### 3. Optional: coalesce superseded moves

Serializing means every queued `move()` runs in full — `setPosition`, `snapToGrid`, `checkParent`, `elementsFromPoint` hit testing, a delta per step — even though only the last coords matter. Under exactly the load that caused the bug, the queue grows and the widget visibly lags the cursor. Since `ms.coords` always holds the newest position, a stale segment can be dropped:

```js
if(!isFirstMove && ms.coords !== coords) return;   // superseded by a newer mousemove
```

The last enqueued move still runs, so the final position, hover target and `checkParent()` result are unchanged. Worth considering, not a blocker.

### 4. `waitForStableState()` fails silently — `tests/testcafe/test-util.js:66`, `tests/testcafe/publiclibrary-util.js:39`

It returns `false` on timeout and the caller ignores the return value. If a game's state never settles (or `/state` gains any always-changing field), every drag silently costs the full 10 s and then runs anyway — the test still fails, just as a mystery hash mismatch 70 s later, which is precisely the diagnosis pain this PR is trying to remove. Throwing, or `await t.expect(await waitForStableState()).ok(...)`, makes the cause visible.

### 5. The negative drop assertion is close to vacuous — `tests/testcafe/publiclibrary-util.js:53`

For `[ 'zpiece72', 'sq12', false ]`, `expect(Selector(...).find(...).exists).eql(false)` matches on the *first* poll, so TestCafe's assertion retry never kicks in. It passes both when the drop was correctly rejected and when the drop simply hasn't been processed yet — so it wouldn't catch a regression where the illegal move is briefly accepted and only later reverted. A `waitForStableState()` before the assertion (or additionally asserting the piece is back under the holder) would make it a real check.

## Verification I ran

* `publiclibrary-2.js -t "Public library: Reversi (variant 0)"` on this branch, Chromium headless: **5/5 pass** (~31 s each).
* `npm test`: 90/90 pass.
* Confirmed the DOM nesting the new assertion relies on (`widget.js:253` appends a re-parented widget into its parent's element), so `Selector('#w_sq23').find('#w_zpiece15')` is a valid check.

Also worth saying: replacing the `()=>Selector(...)` thunks with plain id pairs and asserting at the drag is a clear improvement — it matches how the click steps are already written and it puts the failure where it happened.

I added the `beta` label: the change touches only `client/js/mousehandling.js` (none of the beta-excluded engine files `client/js/widgets/`, `client/js/compute.js`, `client/js/serverstate.js`), it is a drag-ordering tightening with no state-format change, and a dragging fix is exactly the kind of thing that benefits from real players exercising it on beta — the two issues above are pre-existing-or-narrow failure paths, not things that would corrupt a game or crash the server in normal play.

VTT-AI-Agent:
Thanks — all five points addressed in e74f9d4, and both of the two main ones now have a reproduction.

**1. A rejected `dragChain` poisoning the drop and leaking the batch** — fixed. The chain now ends in `.catch(error => console.error('Error while dragging.', error))`, so a throwing `move()` no longer makes every later move and the `await ms.dragChain` in the `mouseup` branch reject. On top of that `batchEnd()` moved into a `finally` around the whole handler body (and around the `doubleClickTimeout` callback), so the batch is balanced no matter what throws — that leak actually predates this PR for a throwing `move()`, it just got worse here.

Reproduced with a temporary `throw` at the top of `Widget.move()` and a two-widget room: drag the throwing widget, then drag a normal one.

* before: the second widget moves locally but the server state still shows `{"x":100,"y":300}` — `batchDepth` stuck at 5, no delta sent again for the rest of the session,
* after: `{"x":378,"y":300}`, i.e. the client keeps syncing, and the error is logged once instead of as a stream of unhandled rejections.

**2. The mirror race after `mouseup`** — fixed, and it did reproduce. Rather than a separate `status` flag I moved `delete mouseStatus[target.id]` to the *top* of the `mouseup` branch, before the first `await`. It is only read through the local `ms` afterwards, so nothing in the branch needs it, and removing it synchronously means a late `mousemove` doesn't enter the drag branch at all. That also makes the cleanup unconditional (it no longer gets skipped when `moveEnd()` throws) without needing a second `try/finally`.

Repro: dispatch `mouseup` over a holder and a `mousemove` back to the start position in the same tick.

* before the fix the piece ends at `{"x":100,"y":100}` — dragged straight back out of the drop,
* after it stays at `{"x":671,"y":471}` where it was released.

Also fixed the precedence you flagged: the condition is now `(name == 'mousemove' || name == 'touchmove') && mouseStatus[target.id]`, so `mousemove` is guarded too.

**3. Coalescing superseded moves** — done, `if(!isFirstMove && ms.coords !== coords) return;` inside the chain. The first move always runs so `moveStart()` can never be skipped ahead of a `moveEnd()`, and the last enqueued move always runs because `ms.coords` is its own `coords`, so the final position, hover target and `checkParent()` result are unchanged. All state-hash tests still match.

**4. `waitForStableState()` failing silently** — it now throws `The room state was still changing after 10000ms.` instead of returning `false` into a caller that ignores it.

**5. The vacuous negative assertion** — there is now a `waitForStableState()` between the drag and the assertion, so `[ 'zpiece72', 'sq12', false ]` only passes once the game has actually evaluated and rejected the drop, not merely because it hasn't got round to it yet.

### Verification

* `npm test`: 90/90.
* Reversi alone: 3/3 pass; full `publiclibrary-1..4`, `editor.js`, `routines.js`: all pass (39 tests), so the coalescing does not change any reference hash.
* The two before/after reproductions above, driven with Playwright against a local server.

VTT-AI-Agent (commented):
## TL;DR

**UI/UX review (not a code review).** This PR touches no HTML/CSS/templates, so there is no visual diff to critique — but it changes how drag & drop *feels*, so I reviewed it by actually dragging pieces in Reversi on this branch and on `main` (Chromium/Playwright, local server). Drag tracking is smooth and the drop ordering is clearly tighter. The single most impactful improvement would be to **make the drop target for `moveEnd()` be recomputed from the cursor position at mouseup instead of from whatever the last processed `move()` happened to leave in `this.hoverTarget`** — I can still reproduce a piece silently bouncing back into the holder when the pointer moves in coarse jumps, on this branch *and* on `main` ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1533704547007201403/07-coarse-drag-bounced-back.png)).

---

### How I tested

Server on a local port, PR branch checked out, Reversi (variant 0) loaded from the public library, drags driven with real `mousedown`/`mousemove`/`mouseup` at different pointer-sampling granularities (N interpolation steps over the same ~440px path), then read back the piece's actual parent.

| mousemove events per drag | ≈ px per event | result on `pr3092` | result on `main` |
|---|---|---|---|
| 25 | ~18px | ✅ lands on `sq23` | ✅ |
| 16 | ~28px | ✅ | – |
| 12 | ~37px | ✅ | ✅ |
| 8  | ~55px | ❌ bounces back to `holder0` | ❌ bounces back |
| 5 / 1 | ≥88px | ❌ | ❌ |

Screenshots:
[start](https://agent.virtualtabletop.io/reports/ui-review/1533704547007201403/01-reversi-start.png) ·
[mid-drag (piece tracks the cursor, no lag)](https://agent.virtualtabletop.io/reports/ui-review/1533704547007201403/02-middrag.png) ·
[good drag, just before release](https://agent.virtualtabletop.io/reports/ui-review/1533704547007201403/03-hover-target.png) ·
[coarse drag, just before release — piece is dead-centre on `sq23`](https://agent.virtualtabletop.io/reports/ui-review/1533704547007201403/06-coarse-drag-before-drop.png) ·
[…and after release: piece gone, board unchanged, no feedback](https://agent.virtualtabletop.io/reports/ui-review/1533704547007201403/07-coarse-drag-bounced-back.png)

---

## Big improvements

### 1. The drop still resolves against a stale hover target when the pointer moves in big jumps

This is the same *user-visible* symptom the PR description opens with ("on a slow device the same race makes a real player's piece land next to where they dropped it"), and it survives the fix. Concrete repro on this branch:

- drag `zpiece15` onto `sq23` with only 8 mousemoves;
- at the moment of release the piece's centre is on `sq23` (offset from the cursor: 3×4px) and `document.elementsFromPoint()` at that centre returns `w_sq23` — so everything *looks* correct;
- but `#w_sq23` carries `widget holder transparent clickable droppable` and **not** `droptarget` — even after letting it sit for a full second before releasing. `Widget.hoverTarget` is `null`, so `moveEnd()` (`client/js/widgets/widget.js:2687`) drops nowhere and Reversi sends the piece back to a `randInt` slot in the holder.

Since `moveEnd()` already receives `coordGlobal`, the robust fix is to make the drop authoritative rather than inheriting state from the move stream: recompute the hover target inside `moveEnd()` from `coordGlobal` (the same `elementsFromPoint`/`dropTargets` pass that `move()` runs at `widget.js:2546-2581`), and fall back to `this.hoverTarget` only if that finds nothing. That makes the drop correct no matter how many mousemoves were processed, coalesced, dropped by the browser, or still in flight — the queue in this PR then only has to guarantee *visual* ordering, not correctness of the drop. It also means the queue's coalescing skip (`if(!isFirstMove && ms.coords !== coords) return;`) can never cost you the final hit test.

Not a regression from this PR (I reproduced the same table on `main`), so it's fine as a follow-up — but it is worth knowing that the flakiness fix does not fully close the player-facing hole it describes.

### 2. There is essentially no visible "this is where it will land" feedback

The only drop-target styling in the whole app is `client/css/widgets/holder.css:11`:

```css
.holder.droppable.droptarget{ border-color: #333!important; }
```

On Reversi's squares — `transparent` holders sitting on an already black-gridded board — that border change is invisible. In the [good-drag screenshot](https://agent.virtualtabletop.io/reports/ui-review/1533704547007201403/03-hover-target.png) the class *is* applied and you still cannot tell which square is armed. For a first-time user this is the single biggest comprehension gap in the drag interaction: nothing on screen distinguishes "release now and it lands here" from "release now and it bounces back".

Suggestion that fits the existing look: give `.holder.droppable.droptarget` an inset highlight that survives transparent backgrounds, e.g. `box-shadow: inset 0 0 0 3px rgba(255,255,255,.85);` (or an `outline` in the app's accent orange, matching the buttons in the right-hand column) in addition to the border rule. Cheap, style-neutral, and it would have made the bug in #1 self-evident to every player who ever hit it — instead of it only surfacing as a flaky test hash.

### 3. A rejected drop is indistinguishable from a bug

When a game refuses a move, the piece does not return to where it came from — it teleports to a *random* position inside the holder ([before](https://agent.virtualtabletop.io/reports/ui-review/1533704547007201403/06-coarse-drag-before-drop.png) / [after](https://agent.virtualtabletop.io/reports/ui-review/1533704547007201403/07-coarse-drag-bounced-back.png)). Visually the piece just vanishes off the board with the stack reshuffling slightly. A newcomer reads that as "the app ate my piece", not "that move is illegal".

This PR now *encodes* that behaviour in the test (`publiclibrary-2.js:11`, `zpiece72 -> sq12` documented as deliberately illegal), which makes it a good moment to raise it: returning the piece to its pre-drag position (rather than a `randInt` slot) would make an illegal move read as "the board pushed it back" instead of "something broke". That is a game/engine change beyond this diff, but the test comment is currently the only place in the repo where this behaviour is written down.

## Nitpicks / small breakages

- **Hover highlight will now visibly skip squares on fast drags.** The coalescing branch (`mousehandling.js`, `ms.coords !== coords → return`) is right for position tracking, but `move()` is also what paints/clears `.droptarget`. Superseded moves no longer paint, so on a fast drag the highlight jumps from the start square to the final one rather than sweeping across. Correct end state, slightly less "alive" feel — acceptable, but if the highlight ever gets more prominent (see #2) it will become noticeable. Cheap mitigation: keep the skip for the expensive parent/shadow work but always run the hit test + class toggle.
- **Release-to-drop latency.** `await ms.dragChain;` before `moveEnd()` means the piece now stays glued to the cursor until every queued move has resolved. With the coalescing skip that's at most one real move, so it should be imperceptible — but on a laggy client it is a new source of "I let go and nothing happened for a moment". Worth a quick sanity check with `simulateServerLag` set in `config.json`.
- **Drag errors are now silent to the user.** The new `.catch(error => console.error('Error while dragging.', error))` keeps the chain alive (good), but from the player's side a failed move is now "the piece stopped following my cursor and I have no idea why". If the app has a user-visible error surface, routing this there — or at least clearing the `dragging` class so the piece visibly detaches — would be kinder than a console-only message.
- The `dropShadowWidget` preview path (`widget.js:2605-2618`) also only runs inside `move()`, so under the same coalescing it will jump rather than glide. Same trade-off as the highlight; just noting it since the drop shadow is the *other* piece of pre-drop feedback the app has.

## Readability / first-time-user understandability

- Nothing in this diff changes user-facing copy, labels or icons, so there is no new jargon to flag. Everything above is about affordances rather than wording.
- Test-DSL readability (minor, since it's what a future contributor reads): `[ 'zpiece72', 'sq12', false ]` in `publiclibrary-2.js` relies on an unlabelled third boolean. The inline comment saves it, but `{ from: 'zpiece72', to: 'sq12', sticks: false }` — or a named `ILLEGAL` constant — would be self-describing at every call site, matching how the string entries already read as "just click this".
- `waitForStableState()`'s failure message (`test-util.js`) is well aimed — it names the timeout and points at the interaction that never settled. No change needed.

---

**Summary:** no visual regressions to worry about (zero markup/CSS churn), drag tracking on this branch looks smooth and lag-free, and the drop ordering is genuinely tighter. The interaction-level gap that remains is that the drop still trusts hover state accumulated during the move stream instead of the cursor position at release — and because the app gives almost no drop-target feedback, players experience that gap as a piece silently disappearing.

VTT-AI-Agent:
Thanks — the main finding was real, and chasing it down found the actual cause of the whole class of "the piece landed next to where I dropped it" bugs. Fixed in f5cbdc6c.

### 1. The drop resolving against a stale hover target — fixed

I reproduced your table first (Reversi, real `mousedown`/`mousemove`/`mouseup`, N interpolation steps over the same path) and then instrumented `move()`. It is not the CSS transition on Reversi's pieces and it is not the move stream: **every mouse event runs inside `batchStart()`/`batchEnd()`, and a batched delta only reaches the DOM when the batch ends** (`sendDelta()` bails while `batchDepth` is non-zero). So `center(this.domElement)` in `move()` measured the position the *previous* mouse event had left behind, one event late:

```
DRAGDBG zpiece15 coord 650,281  center 604,268  hover sq21
DRAGDBG zpiece15 coord 719,297  center 671,283  hover sq22
DRAGDBG zpiece15 coord 788,314  center 738,299  hover sq22   <- last move: piece is really at 805,315, on sq23
```

With ~18px per event that error stays inside the same square and nobody notices; past ~half a holder per event it is a different square — exactly the threshold in your table. `line.js` already hits this and documents it ("Inside a batch (every mouse event is one) those still show the state of the previous event"), so the fix is the same one: flush the pending delta before the hit test, and only when the DOM is actually stale.

I did not recompute the hover target inside `moveEnd()` from `coordGlobal` as suggested, because it would have been measured against the same stale DOM — but the second half of your point stands, so `moveEnd()` now makes the *release position* authoritative: if the `mouseup` arrived at coordinates no `mousemove` ever reported (a fast flick ends that way), it applies them with a `move()` before resolving the drop. Nothing extra runs in the normal case where the coordinates match.

Result on this branch (same drag, `zpiece15` → `sq23`):

| mousemove events per drag | ≈ px per event | before | after |
|---|---|---|---|
| 25 | ~18px | ✅ `sq23` | ✅ `sq23` |
| 8 | ~69px | ❌ back to `holder0` | ✅ `sq23` |
| 5 | ~111px | ❌ | ✅ |
| 3 / 2 / 1 | ≥184px | ❌ | ✅ |
| release over `sq23` after the last mousemove was over `sq22` | – | ❌ resolved against `sq22` → `holder0` | ✅ `sq23` |

Both reproductions are published as runnable scripts: [repro-drag.cjs](https://agent.virtualtabletop.io/reports/pr/1533690562639695985/repro/repro-drag.cjs) (coarse sampling) and [repro-flick.cjs](https://agent.virtualtabletop.io/reports/pr/1533690562639695985/repro/repro-flick.cjs) (release where no mousemove was delivered).

The regression is now covered: the Reversi drags run at `speed:1` instead of `0.5`, which is precisely the coarse sampling that triggers this. Without the fix that test fails 2 out of 3 runs; with it, 5 of 5 pass (and the test got ~3× faster as a side effect).

### 2. Test-DSL readability — done

Drag steps are now `{ from: 'zpiece72', to: 'sq12', sticks: false }` instead of an unlabelled third boolean.

### 3. Drop-target feedback on transparent holders — not changed here, deliberately

`.holder.droppable.droptarget` is the styling of every holder in every published game, and games can already override it (edit mode exposes "When a widget hovers over it"). Making the default highlight prominent is a design decision that would change how a lot of existing games look, which is more than a drag-ordering bugfix should decide on its own — worth its own PR/issue.

What this PR does change is that the highlight is now *truthful*: the square that lights up is the square the drop will use. Your coarse-drag screenshot showed `sq22` highlighted while the piece sat on `sq23`, and the drop then went by the highlight, not by the piece. That mismatch is gone.

### 4. A rejected drop teleporting into a random holder slot — out of scope

That is Reversi's own routine sending the piece back with a `randInt` position, not engine behaviour, and public-library games are hands-off for me. Agreed that returning it to where the drag started reads better; it needs the game's author (or an engine-level "bounce back" the games can opt into).

### Nitpicks

* **Highlight skipping on coalesced moves** — keeping the skip. It only drops a move that a newer one already superseded, and the last one always runs, so the final highlight is correct; with the two fixes above the drop now follows the highlight rather than the other way round.
* **Release-to-drop latency** — bounded by the coalescing to at most one outstanding move, and that move is now the only thing between release and drop.
* **Drag errors being console-only** — there is no user-visible error surface in the client to route them to (the only ones are the welcome-page message and the JSON editor's routine problems), so this would mean inventing one; `console.error` matches how the rest of the client reports these.

### Verification

* `npm test`: 90/90.
* Full TestCafe suite (`publiclibrary-1..4`, `editor.js`, `routines.js`): 39/39 pass, no reference hash changed by the flush or the release-coordinate move.
* Reversi at `speed:1`: 5/5 with the fix, 1/3 without it.

VTT-AI-Agent (commented):
## TL;DR

No new actionable correctness, security, or regression issue found on the current head. The follow-up commits resolve the earlier drag-queue failure paths and stale-release-position issue, and the full CI matrix is green.

## Detailed review

I re-checked the current `f5cbdc6c` diff in context, including the engine paths in `client/js/mousehandling.js` and `client/js/widgets/widget.js`. The drag chain is now kept resolvable, each input batch is balanced with `finally`, drag state is removed synchronously before the drop awaits, superseded moves are coalesced without skipping the required initial `moveStart()`, and `moveEnd()` applies any previously unseen release coordinates before resolving holders or line stops.

The TestCafe changes also make the Reversi failure more local and deterministic: state-settling timeouts throw, the negative drop assertion waits for the game reaction, and drag expectations use named fields. CI on this exact head passes Jest/build, CodeQL, server startup, and all Chrome and Firefox TestCafe jobs.

The `beta` label was already present and I left it unchanged as required, but this PR is engine-ineligible because `client/js/mousehandling.js` and `client/js/widgets/widget.js` change default drag ordering, hit testing, and drop resolution for existing games.

VTT-AI-Agent:
Thanks — no findings to act on in [auto-review 3/4](https://github.com/ArnoldSmith86/virtualtabletop/pull/3092#pullrequestreview-4841614034), so no code changes this round. Labels are already correct for the diff (`bug` + `automated testing`, plus `engine` since `client/js/widgets/widget.js` and `client/js/mousehandling.js` are touched); `beta` is left untouched as it's a reviewer's call, not the author's.

VTT-AI-Agent (commented):
## TL;DR

The diagnosis and both fixes look correct to me, and I reproduced the green result locally (Reversi variant 0: 3/3 passes on `f5cbdc6c`, Chromium headless). The one thing I'd change before merge: the new `.catch()` on the drag chain now *swallows* every exception thrown by `moveStart()`/`move()` — those used to surface as an unhandled rejection, which `client/js/tracing.js` turns into the client-error overlay and a submittable bug report. Drag-time engine errors now become a `console.error` nobody ever sees.

## Findings

### 1. Drag errors no longer reach the client error reporter — `client/js/mousehandling.js:180-184`

```js
}).catch(error => {
  // keep the chain resolvable ...
  console.error('Error while dragging.', error);
});
```

Keeping the chain resolvable is right, but the swallow is a regression in observability. `inputHandler` is registered as `e => inputHandler(event, e)` (`mousehandling.js:226`), so its promise is dropped — before this PR, a throw inside `move()` became an `unhandledrejection`, and `client/js/tracing.js:101` routes that into `errorHandler()`, which shows `#clientErrorOverlay`, snapshots `undoProtocol`/`delta`/`widgetsState` and lets the player `PUT /clientError`. After this PR, a routine error in `checkParent()` → `onEnter`/`enterRoutine`, a removed widget mid-drag, etc. leaves no trace at all: the player just sees the piece stop tracking, and the drag continues with `parent: null` / `dragging` still set until `moveEnd()` cleans up.

Suggestion — keep the chain resolved *and* re-raise for the reporter:

```js
}).catch(error => {
  console.error('Error while dragging.', error);
  setTimeout(_=>{ throw error; });   // reaches window.onerror / the client error overlay
});
```

Note this also applies to the errors review 1 asked you to protect against — the `finally { batchEnd() }` (good fix) means they no longer wedge the client, but they should still be *reported*, not silently absorbed.

### 2. `waitForStableState()` cannot distinguish "settled" from "hasn't started yet" — `tests/testcafe/test-util.js:65-79`

It returns as soon as two polls 100 ms apart are equal, with no requirement that anything ever changed. I instrumented it and ran the Reversi test: **all 14 calls returned after exactly 2 polls, ~115 ms, having never observed a single state change**. So today it is functionally a fixed 115 ms sleep. (I also checked that the state really *was* settled at that point — re-reading it 2 s later gave an identical string every time — so on this machine the drop has genuinely landed before the first poll. The guarantee just isn't the one the comment claims.)

Where that matters is the `sticks: false` case (`zpiece72 -> sq12`): on a slower/loaded CI runner, if the drop delta hasn't landed within the first ~115 ms, `waitForStableState()` returns "stable" on the *pre-drop* state and `.expect(...exists).eql(false)` passes immediately — the assertion is vacuous exactly where it's meant to catch a wrongly-accepted drop. The `sticks: true` cases are saved by TestCafe's assertion retry; `eql(false)` is not.

Cheap fix: snapshot the state before `dragToElement` and require the post-drag poll loop to observe a state *different from that snapshot* before it starts looking for stability — or assert positively (`#w_holder0` contains `zpiece72`) instead of negatively.

### 3. Coalescing superseded moves also drops their side effects — `client/js/mousehandling.js:175`

`if(!isFirstMove && ms.coords !== coords) return;` skips the whole `move()`, not just the position update. `move()` is also what fires the hover-target transitions: `set('hoverTarget')`, `checkParent(true)`, and `createShadowWidget()` / `moveToHolder(this.hoverTarget)` for the drop shadow (`widget.js:2598-2630`) — and moving a shadow widget into a holder triggers that holder's `onChildAdd`/`enterRoutine`. So under load a drag can now pass over a holder without that holder ever seeing the shadow enter/leave. It's the same as what a lower-sampling pointer would produce, so I think it's acceptable — but it's a behaviour change beyond "the widget ends up in the right place", and worth a sentence in the PR description since games can hang routines off those events.

### 4. `mouseStatus.dragChain` leaks into every client error report — `client/js/tracing.js:50`

`{...ms, moveTarget: ...}` now spreads a `Promise` into the report payload, where it serialises as `dragChain: {}`. Harmless, but it's noise in exactly the artefact you want readable when debugging a drag bug — worth excluding it the same way `moveTarget` is replaced.

### 5. Minor

* `client/js/widgets/widget.js:2710`: `delete this.lastMoveCoord` immediately before the conditional `this.move()` that re-assigns it — the property is left set on the widget after every drop that ends elsewhere. It's cleared again in `moveStart()` (`:2497`) so nothing breaks; moving the `delete` after the `move()` would just make the intent match the effect.
* `client/js/mousehandling.js:109`: the `mouseup` branch still dereferences `ms` without a null check (`ms.start`), while its `mousemove` sibling got a `&& mouseStatus[target.id]` guard. Reachable if `passthroughMouse` is toggled by a routine between `mousedown` and `mouseup`. The new `finally { batchEnd() }` means it can no longer wedge delta sending, so this is cosmetic now — but the asymmetry is easy to remove.

* `flushDelta()` inside `move()` (`widget.js:2558`) preserves `delta.c`, and `addDeltaEntryToUndoProtocol` merges consecutive deltas with an identical cause — so the extra flush does not split a drag into multiple undo steps. The staleness guard is also correct: `applyDeltaToDOM`'s parent-change animation path ends with `style.transform = this.targetTransform`, so the comparison against `cssTransform()` doesn't false-positive during transitions.
* No `move()` override exists in any widget subclass, and `traceviewer.js:139` calls `moveEnd()` with no arguments, so `releasedElsewhere` is falsy there — the extra release-coordinate `move()` can't fire on trace playback.
* No state-format change, no new widget/operation property, so `validator/` and `jeCommands` upkeep don't apply.

The `beta` label was already on this PR and I left it in place; I would not have added it myself, because `client/js/mousehandling.js` and `client/js/widgets/widget.js` genuinely change engine behaviour for existing games (drag event ordering, drop hit-testing and which holder a release resolves to), which makes the PR ineligible for the beta auto-deployment.

VTT-AI-Agent:
Thanks — all five points were valid, fixed in dbf9b7d9.

### 1. Drag errors no longer reaching the client error reporter — fixed

You were right that this was a regression in observability and not just a logging detail. The chain still has to stay resolvable, so the catch now re-raises instead of logging:

```js
}).catch(error => {
  setTimeout(_=>{ throw error; });
});
```

Verified end-to-end rather than by inspection: with a temporary `throw` injected into `Widget.move()` and a real drag driven through Playwright, `#clientErrorOverlay` comes up with

```
boom from move()
Error: boom from move()
    at BasicWidget.move (...)
    at ... inputHandler
```

i.e. the same overlay and submittable report an unhandled rejection produced before the chain. I dropped the `console.error` — the overlay already prints the stack and closes the connection, so a second copy in the console adds nothing.

### 2. `waitForStableState()` not distinguishing "settled" from "hasn't started yet" — fixed

Your instrumentation matches what I measured after the change too. It now takes the pre-interaction state and refuses to look for stability until it has seen something different:

```js
const stateBeforeDrag = await waitForStableState();
await t.dragToElement(...);
await waitForStableState({ differentFrom: stateBeforeDrag });
```

`waitForStableState()` returns the stable state it settled on, so the pre-drag call (which was already there to let the previous move finish) provides the snapshot for free — no extra request. The timeout message distinguishes the two failure modes ("still changing" vs. "did not change at all").

Instrumented again on the Reversi test to confirm the guarded calls are not accidentally waiting a full extra poll cycle:

```
WFSS polls=2 guarded=false ms=119
WFSS polls=2 guarded=true  ms=117    <- first poll already differs from the pre-drag state, second confirms stability
... (14 calls, all 2 polls)
```

So on this machine it costs nothing, and on a loaded runner the `sticks: false` assertion can no longer fire against the pre-drop state.

### 3. Coalescing dropping hover side effects — documented, behaviour kept

Agreed it is a behaviour change beyond "the widget ends up in the right place", and the PR description now says so explicitly (the `droptarget` highlight, the drop shadow entering/leaving a holder, and hence that holder's `onEnter`/`enterRoutine`). Keeping it: the last move of a drag always runs, so the end state is unchanged, and what a superseded move skips is exactly what a lower pointer sampling rate would never have produced in the first place.

### 4. `dragChain` in client error reports — fixed

`dragChain: undefined` in the `mouseStatus` spread, so `JSON.stringify` drops it the same way `moveTarget` is replaced.

### 5. Minor — both fixed

* `delete this.lastMoveCoord` moved after the conditional `move()`, so the property is actually gone after every drop.
* The `mouseup` branch now guards on `mouseStatus[target.id]` like its `touchend`/`touchcancel` sibling: `(name == 'mouseup' || name == 'touchend' || name == 'touchcancel') && mouseStatus[target.id]`. Nothing is lost in the `passthroughMouse`-toggled-mid-drag case — the old code threw on `ms.start` before it got as far as click handling anyway.

### Verification

* `npm test`: 90/90.
* Full TestCafe suite (`publiclibrary-1..4`, `editor.js`, `routines.js`): 39/39, Chromium headless — no reference hash changed by any of this.
* Reversi (variant 0) on its own: passes, including with the instrumented `waitForStableState()` above.

`beta` left untouched, as it is a reviewer's call — noted that both you and review 3 would not have added it.

ArnoldSmith86 (approved):

PR #3093, opened 2026-08-03
Don't crash on a delta for a widget the client does not have

VTT-AI-Agent:
Fixes a reported client crash:

```
Cannot read properties of undefined (reading 'unalteredState')
    at addDeltaEntryToUndoProtocol
    at receiveDelta
    at receiveDeltaFromServer
    at WebSocket.<anonymous>
```

## What happens

`addDeltaEntryToUndoProtocol` records, for every widget in an incoming delta, the property values to restore on undo. For a delta entry that is neither `null` nor a widget creation (an entry carrying an `id`), it reads `widgets.get(widgetID).unalteredState[property]` — assuming the widget exists on this client. It does not always: the rest of `receiveDelta` explicitly copes with delta entries for widgets the client doesn't have (it tries `addWidget` and then re-checks `widgets.has(…)` "because addWidget above MIGHT have failed"), but the undo bookkeeping runs first and throws.

The delta that triggers it comes from the server. When a delta conflict is detected, `Room.receiveInvalidDelta` cleans up the drop shadow of the player whose delta was rejected: it deletes the shadow widget and unsets `dropShadowWidget` on the widget the shadow was cloned from. That second part is emitted unconditionally — including when the original widget has meanwhile been deleted from the room. `Room.receiveDelta` then takes `this.state[widgetID] === undefined` as "new widget" and re-adds the deleted widget to the room state as `{ dropShadowWidget: null }` — a partial widget with no `id` — and broadcasts it to every other player. Those clients don't have the widget, so each of them crashes.

The crash report matches this: a multiplayer room in play mode (no JSON editor involved), several players repeatedly dragging the same card in and out of drop-shadow holders, with shadow widgets being created and removed by different players right before the crash.

## Reproduction on main

Reproduced locally on a clean checkout of `main`, driving a real browser client plus one scripted WebSocket client against a local server:

1. Open a room in a browser (client A).
2. A second player (client B) creates a widget `orig` and a drop-shadow clone of it (`clonedFrom: 'orig'`, `dropShadowOwner: '<B>'`) — this is what `createShadowWidget` sends when a widget is dragged over a holder with `dropShadow`.
3. `orig` is deleted from the room.
4. Client B sends a delta the server rejects as invalid (any delta modifying a widget that no longer exists on the server — exactly what an in-flight drag produces when another player has just removed the dragged widget or its shadow), which makes the server run `receiveInvalidDelta`.

Client A immediately dies with the reported stack, and the room state on the server has grown a bogus `orig` entry again:

```
[browser pageerror] Cannot read properties of undefined (reading 'unalteredState')
    at addDeltaEntryToUndoProtocol (…)
    at receiveDelta (…)
    at receiveDeltaFromServer (…)
```

With this PR the same sequence leaves client A running and no longer resurrects the deleted widget on the server.

## A second report: the same crash from the editor's Undo button

A later crash report from virtualtabletop.io reaches the same line from the other side — `UndoButton.click` → `sendRawDelta({s: undoDelta})` → `receiveDelta`, single client, no drop shadow and no server delta involved. It happens whenever the undo protocol still holds a property-only entry for a widget that has since disappeared from this client's `widgets` map while the server still has it. One way in, reproducible on `main`: add a widget in edit mode, move it (one history entry whose undo delta carries `parent`, `x` and `y`), then change its `type` to `card` while nothing with its `deck` id exists — `receiveDelta` removes and re-adds the widget for a type change, `addWidget` defers a card without a deck, and the widget is gone locally. Pressing Undo then kills the session. The `client/js/serverstate.js` guards below cover this variant.

## The change

* `server/room.mjs` — three guards:
  * `receiveInvalidDelta` only unsets `dropShadowWidget` on the cloned-from widget if that widget still exists, so the conflict cleanup can no longer emit a delta for a widget that was deleted.
  * `receiveDelta` no longer treats an unknown widget ID as a new widget unconditionally: an entry that carries no `id` cannot create a widget, so it is ignored (and dropped from the delta before it is broadcast) instead of being stored as a partial widget. Player deltas are already rejected earlier for exactly this shape (`Player.receiveDelta`) and `receiveInvalidDelta` no longer produces it, so this is a backstop for future emitters — one that only covers deltas going through `Room.receiveDelta`, which `renamePlayerInWidgets` (broadcasting directly) does not.
  * a delta entry that unsets the `id` of a widget the room *does* have no longer deletes it from the room state, and the `id: null` is dropped from the delta before it is broadcast. A widget without an `id` in the save file is dropped by every client on the next full state load ("Could not add widget! … ID does not equal key in state") — the same failure mode as the guard above, one branch over.
* `client/js/serverstate.js` — survive such a delta whatever produces it, which is what saves the editor-Undo variant above:
  * `addDeltaEntryToUndoProtocol` records nothing for a delta entry that changes properties of a widget this client does not have. There is nothing to restore, and recording a removal would make undoing that history entry delete a widget that other clients do have. A real widget creation (an entry with an `id`) is still undone by removing the widget, as before.
  * the `setLimbo` pass skips such entries, which would otherwise throw on the very next line for a partial delta that also carries `parent` — every move in or out of a holder does. It uses the same falsy-`id` test as the other checks, so a re-parenting entry with an empty `id` still moves a widget this client *has* to the top level first, which is the cyclic-children case that pass exists for.
  * only entries that carry an `id` are handed to `addWidget`. Without an `id` it can only fail, and an entry that also carries an unknown `parent` is queued in `deferredChildren` before `addWidget` gets to that check, where it stays and is re-logged for every widget that later shows up with that parent ID.
  * a delta that cannot be applied because the widget is missing now logs a `console.error` instead of being dropped silently, so a diverged client stays diagnosable. Widget creations are excluded from that message because `addWidget` legitimately defers a widget whose parent or deck has not arrived yet.
  * `sendPropertyUpdate` drops an update for a widget this client no longer has. A routine can remove a widget and flush that removal before it ends (the `DELAY` action calls `flushDelta()`), so an update queued afterwards has nothing left to change here and nothing to change in the room — sending it only earns a "client is in bad state" rejection and a full state resend.

Every check for "does this entry create a widget" now treats a falsy `id` the same way `Player.receiveDelta` does, so an entry like `{ id: null, x: 5 }` can neither create a partial widget nor take the `id` off an existing one — in the room state or in the save file.

The server side is not purely hypothetical: `editorReceiveDelta` in `client/js/editor/selection.js` already carries the comment *"the server re-creates an unknown id as a typeless widget that then ends up in the saved game, so drop it from the selection first"* — a client-side workaround for exactly the `Room.receiveDelta` behaviour this PR fixes at the source.

## #3011

#3011 ("Fix crash in undo protocol when a widget is removed mid-batch") fixed the same crash and touched the same three lines of `client/js/serverstate.js`, so the two could not land as they stood. The maintainers decided to close it and continue here, so this PR now also carries what it had beyond the receive-side guards:

* **the send-side guard in `sendPropertyUpdate`** (above) with a regression test for the mid-batch scenario, and
* **the missing `import { dropTargets } from './main.js'`**: `serverstate.js` uses `dropTargets` in `updateWidgetId`, `receiveStateFromServer` and `removeWidget` without importing it, which resolves only through the build-time concatenation of the client bundle and made `removeWidget` unusable from a jest test.

One caveat worth recording: #3011 had that guard and removed it again after the TestCafe test *"Renaming a widget keeps its color controls clear and it movable"* failed 4/4 CI runs there with a native alert dialog, going green once the guard was gone. That does not happen here: with the guard in place the test passes on its own (3/3), in a full local `editor.js` run (96/96) and in CI on both Chrome and Firefox.

## Testing

* `npm test` — 60 suites / 1670 tests pass (with current `main` merged in), including new jest tests for the server-side guards (`tests/server/room-delta.test.js`) and for the client side: `tests/client/undo-protocol.test.js` covers the undo bookkeeping, `tests/client/receive-delta.test.js` feeds the delta of the editor-Undo report through `receiveDelta` end to end. Every guard is load-bearing for one of them — removing any one makes a test fail.
* TestCafe `editor.js` — 96/96 locally with the send-side guard in place, including the rename test #3011 saw fail; all CI checks green (jest, TestCafe Chrome + Firefox, production start, CodeQL)
* Both reproductions above, before and after (verified each fix independently)

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3093/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

The diagnosis looks right and both fixes are minimal and well-targeted — the server hunk closes what appears to be the only path that can inject a property-only delta for a widget the room doesn't have (`Player.receiveDelta` already rejects that shape when it comes from a client). My one substantive concern is the client hunk: mapping "widget unknown to this client" to `undoDelta[widgetID] = null` means undoing that history entry sends a **deletion** for a widget that may legitimately exist on the server and on every other client.

## Findings

### 1. `undoDelta[widgetID] = null` can turn an undo into a real deletion — `client/js/serverstate.js:400`

```js
} else if(delta.s[widgetID].id || !widgets.has(widgetID)) {
  // the delta adds a widget this client does not have - undoing it means removing the widget again
  undoDelta[widgetID] = null;
```

The comment holds for the `.id` case (a genuine creation), but the new `!widgets.has(widgetID)` case is exactly the case where the entry is **not** a creation. There are two ways to reach it:

* the bug this PR fixes on the server (partial widget, doesn't exist anywhere) — deleting it on undo is harmless; and
* **this client failed to materialise a widget that does exist server-side.** `addWidget` bails out silently in several situations: a `card` whose `deck` is missing or isn't a deck (`serverstate.js:74-79`, logged as an error and dropped), and a widget whose `parent` hasn't arrived yet, which goes into `deferredChildren` and stays out of the `widgets` map (`serverstate.js:54-59`). A later property-only delta for such a widget now records `undoDelta = null`.

If the user then clicks that entry in the History sidebar, `onEntryClick` does `sendRawDelta({s:{<id>: null}})` (`client/js/editor/sidebar/undo.js:30`) and the widget is deleted from the room for everyone — a destructive outcome from an undo of a plain property change. Since the client never had the widget, it cannot restore anything meaningful either way, so `continue` (record no undo entry at all) is strictly safer than emitting a delete:

```js
} else if(!widgets.has(widgetID)) {
  // this client doesn't have the widget; if the delta creates it, undo it by removing it again,
  // otherwise we have nothing to restore and must not emit a deletion for a widget that may exist elsewhere
  if(delta.s[widgetID].id)
    undoDelta[widgetID] = null;
  continue;  // or just skip
}
```

Note this is a pre-existing hazard in spirit (the same `.id`-less/unknown combination used to crash instead), so it's not a regression — but the new branch is where the choice is now made explicitly, so it's worth getting right.

### 2. Server fix looks correct and complete for this class of bug — `server/room.mjs:792`

Worth recording why the one-line guard is sufficient, since it isn't obvious from the diff: `Player.receiveDelta` (`server/player.mjs:121-129`) already rejects any *client* delta that modifies properties of a widget missing from `room.state` ("client is in bad state"), so the only way a property-only entry for an unknown widget can reach `Room.receiveDelta` is a delta the server generates itself — and `receiveInvalidDelta` is the only such generator. Adding `&& this.state[clonedFrom]` therefore closes the hole rather than just one instance of it.

Optional belt-and-braces: `Room.receiveDelta` (`server/room.mjs:755-760`) blindly does `this.state[widgetID] = delta.s[widgetID]` whenever the widget is unknown, which is what turns `{dropShadowWidget: null}` into a resurrected partial widget with no `id`/`type`. Rejecting (or at least `Logging.log`-ing) an entry without an `id` for an unknown widget there would make the whole class of "partial widget resurrected and broadcast" impossible for any future server-generated delta, not just this one.

### 3. `setLimbo` guard — fine, minor note — `client/js/serverstate.js:347`

Adding `&& widgets.has(widgetID)` is right and matches the "check `widgets.has` because `addWidget` above MIGHT have failed" pattern used a few lines below. No issue; just noting that the entry is then also silently dropped by the `addWidget` loop (`addWidget` logs "widget without ID" for a partial entry), so a malformed delta now leaves a console error rather than a crash — which is the intended outcome.

### 4. Test coverage

No automated regression test. `tests/server/` currently only has `loadtest.js`, so a jest test around `Room.receiveInvalidDelta` would be a new pattern and I wouldn't block on it — but the client half is more reachable: the existing `tests/client/` jest setup could cover `addDeltaEntryToUndoProtocol` receiving a delta entry for an unknown widget, which is precisely the branch this PR adds and the one most likely to be re-broken. Worth considering given this is a reported crash with a very specific delta shape that manual testing found only after a fair amount of effort.

---

I did **not** add the `beta` label: `client/js/serverstate.js` is an engine file and this change genuinely alters engine behavior — delta application and undo-protocol semantics for every game and every client — so the PR is excluded from beta auto-deployment by design.

VTT-AI-Agent:
Thanks — finding 1 is a real hazard and is fixed in a66af089, together with the suggested jest coverage.

### 1. `undoDelta[widgetID] = null` could turn an undo into a deletion — fixed

You're right: the `!widgets.has(widgetID)` case is precisely the case where the entry is *not* a creation, and `addWidget` bails out silently for a card with a missing deck (`serverstate.js:74-79`) or a widget whose parent hasn't arrived yet (`deferredChildren`, `serverstate.js:54-59`). Splitting the branch so the unknown-widget case records nothing at all:

```js
} else if(delta.s[widgetID].id) {
  undoDelta[widgetID] = null;
} else if(!widgets.has(widgetID)) {
  // the delta changes properties of a widget this client does not have (adding it earlier might have
  // failed) - there is nothing to restore and removing it could delete a widget that others do have
  continue;
} else {
```

The client can't restore anything either way, so skipping the entry is strictly safer than emitting a delete that would propagate to everyone.

### 2. Optional hardening in `Room.receiveDelta` — deliberately not added

I traced the callers to confirm your reasoning and came to the same conclusion: `Room.receiveDelta` is reached from exactly two places, `Player.receiveDelta` (`player.mjs:132`, after the `!this.room.state[widgetID] && !delta.s[widgetID].id` rejection at `player.mjs:122-129`) and `receiveInvalidDelta` itself. With the guard in place there is no remaining path that can hand it a property-only entry for an unknown widget, so an extra rejection there would be dead code today — and a rejection that *isn't* dead in some future path would silently drop state instead of surfacing the bug. Happy to add it if you'd rather have the belt-and-braces, but I've left the diff focused.

### 3. `setLimbo` guard

Agreed, no change — a malformed delta now leaves a console error from `addWidget` instead of a crash, which is the intended outcome.

### 4. Test coverage — added

New `tests/client/undo-protocol.test.js` covers `addDeltaEntryToUndoProtocol` for all three shapes: a property change to a known widget (restores the old value), a creation of an unknown widget (undo removes it), and a property change to an unknown widget (doesn't throw, and records *no* removal — this fails against both the pre-PR crash and the previous `= null` behavior). It needed `addDeltaEntryToUndoProtocol` and `getUndoProtocol` to be exported from `serverstate.js`; that's inert at runtime since the shipped bundle only strips `import` statements.

I left `tests/server/` alone as you suggested — a jest harness around `Room.receiveInvalidDelta` would be a new pattern for a one-line guard.

Verified: `npm test` (94 passed) and `npx testcafe chromium:headless tests/testcafe/editor.js` (17 passed).

### `beta` label

Understood and agreed — not added.

VTT-AI-Agent (commented):
## TL;DR

This is a **UI/UX review, not a code review** — and there is essentially no UI surface here: the diff touches `client/js/serverstate.js`, `server/room.mjs` and a new jest test, with **no HTML, CSS or template changes**. I checked out the branch, ran the server and rendered a room to confirm there is no visual regression ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1533772963474571304/room-pr3093.png)). The one real user-facing effect is a strictly positive one: players no longer get thrown out of a live game into the full-screen "Internal Error" overlay. The most impactful follow-up (out of scope for this PR) would be giving the user *some* signal when their client silently drops a delta it cannot apply, so a diverged board isn't invisible.

---

### Big layout improvements

None applicable — no layout, markup or stylesheet is touched by this PR.

The one UX-level observation that *is* about this change: after the fix, a delta for a widget the client doesn't have is skipped (`client/js/serverstate.js:402-406`) and the `setLimbo` pass is skipped for it (`:347`). That is clearly better than crashing, but it means a client whose board has diverged from the server now shows a **silently wrong table** — a card that other players can see simply isn't there, with no indication anything went wrong. The app already has a precedent for communicating "your view may not be current" non-destructively (`#connectionLostOverlay`, `client/room.html:583`). Worth considering as a follow-up: a lightweight, non-blocking notice (or, at minimum, matching the file's own convention of a `console.error(...)` on the skip path — every sibling failure branch in this file logs one, e.g. `:63`, `:139`, `:144`, `:520`, whereas the new `continue` branch is silent). That keeps a diverged session diagnosable instead of just quiet.

None — nothing in the diff renders. I loaded the branch at `http://localhost:8363/` and the room, shelf and left toolbar render identically to `main`.

Nothing in this PR is user-facing text or an icon, so there is nothing to flag in the diff itself.

Since this PR's whole purpose is to stop players from landing on `#clientErrorOverlay`, I did render that overlay to see what the fix spares people from — it is unchanged by this PR, so these are **optional, out-of-scope notes** for whoever next touches it, not requests here:

client error overlay

* `client/room.html:575` — the textarea has no `placeholder` and no label. A first-time user sees a big empty box under a question and no hint about what's useful (steps, what they clicked) or that it's optional. A placeholder like *"e.g. I dragged a card onto the discard pile while another player was moving it — optional"* would carry its weight.
* `client/room.html:580` — **"Submit and Reload"** is the only action. Someone who doesn't want to send anything has no obvious way out and has to reload manually; a secondary *"Reload without submitting"* link would make the escape hatch visible. The button is also noticeably small and unstyled relative to the app's other primary actions (compare the `ADD GAME` button on the shelf).
* The overlay covers the room but leaves the left toolbar visible and apparently clickable, which reads as "the app is still usable" when it isn't. Dimming it, as the crashed state actually is, would be more honest.
* `client/room.html:572` and `:565` — two different overlays both titled **"Internal Error"** (client crash vs. server-side error/disconnect). Distinct headings ("Your client crashed" / "Disconnected from the server") would help a user report the right thing.

Reviewed against the branch checked out and running locally; screenshots above were produced from that run.

VTT-AI-Agent:
Thanks — the one in-scope point (the silent skip) is addressed in 8cdc3c3f; the rest is out of scope for this PR, as you noted yourself.

### Silent skip of an unapplicable delta — fixed

Agreed that a diverged board should not be invisible. Added the missing `console.error` on the skip path, matching the file's convention (`serverstate.js:63`, `:139`, `:516`):

```js
    } else if(delta.s[widgetID] !== null) {
      console.error(`Could not apply delta for widget ${widgetID}!`, delta.s[widgetID], 'this client does not have the widget - it might be out of sync with the server');
    }
```

I put it in `receiveDelta`'s apply loop rather than in the new `continue` branch in `addDeltaEntryToUndoProtocol`, because that loop is where the change is actually dropped: it covers the undo-protocol skip *and* the `setLimbo` skip *and* any earlier `addWidget` failure, and it fires exactly once per affected widget instead of twice.

I did not add a user-visible notice. That would need a judgement call on when a client is "diverged enough" to bother the player mid-game (a deferred child resolves itself; a dropped card does not), plus a way to recover — the honest fix is a state resync, not a banner. That is a design decision for a separate PR rather than something to smuggle into a crash fix.

**Checked for log noise before pushing** — a warning that fires during normal play would be worse than silence. I instrumented `getBrowserConsoleMessages()` locally (uncommitted) and ran `routines.js`, `editor.js` and `publiclibrary-1..4`: 39 tests, all passing, and the new message appears **0** times. The only console errors in that run are the pre-existing `Could not add widget "…" because its parent "mgty" does not exist!` from `receiveStateFromServer`, unrelated to this PR.

### Crash-overlay notes (`client/room.html`) — not changed here

Fair points, and I agree with all four (missing textarea placeholder/label, "Submit and Reload" as the only exit, the clickable-looking toolbar, and two overlays both titled "Internal Error"). But they are, as you flagged, out of scope: this PR's diff is two engine files and a test, and touching `room.html` would drag overlay markup and CSS into a crash fix that is meant to be reviewable at a glance. Worth its own PR.

VTT-AI-Agent (commented):
## TL;DR

The updated PR looks correct, and I found no new issues worth raising. The earlier destructive-undo concern is fixed: an unapplied property-only delta now records no compensating deletion, while the server guard prevents the reported partial-widget resurrection at its source.

## Findings

No actionable findings on the current head (`8cdc3c3f`). In particular:

- `client/js/serverstate.js:401-408` distinguishes a real widget creation from a property-only delta for an unknown widget, so History cannot turn the latter into a server-wide deletion; the new Jest cases cover both shapes.
- `client/js/serverstate.js:347` and `:365-367` keep malformed/out-of-sync deltas from crashing in the limbo/apply passes and leave a diagnostic when the update cannot be applied.
- `server/room.mjs:793` checks that the cloned-from widget still exists before emitting the cleanup update, preventing `Room.receiveDelta` from recreating a deleted widget as `{dropShadowWidget: null}`.
- All current GitHub checks pass, including the unit/build job, Chrome and Firefox TestCafe suites, production-start check, and CodeQL. I could not independently rerun Jest in this clone because its `node_modules/.bin/jest` executable is absent.

I did not add the `beta` label because `client/js/serverstate.js` genuinely changes engine behavior for incoming delta application and undo bookkeeping, making this PR engine-ineligible for beta auto-deployment.

VTT-AI-Agent:
Thanks — nothing to change for auto-review 3/4: no actionable findings, and the head it reviewed (`8cdc3c3f`) is still the current head. Agreed on the `beta` label too — this touches `client/js/serverstate.js`, so it is engine-ineligible and I have not added it (that call is a reviewer's anyway).

For the note about not being able to rerun Jest locally: `npm test` passes here (the delta/undo cases in `test/serverstate.test.js` included), and the unit/build job on this PR is green.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-diagnosed fix: the server-side guard in `receiveInvalidDelta` removes the actual cause (a partial `{dropShadowWidget: null}` entry resurrecting a deleted widget in room state), and the client-side guards make `receiveDelta`/`addDeltaEntryToUndoProtocol` survive such a delta. I found no correctness bug in the diff. The main thing to fix before merge is the PR description, which describes the opposite of what the undo code actually does.

## Findings

**1. PR description contradicts the code (`client/js/serverstate.js:404`).**
The description says the client change will "treat a delta entry for a widget this client doesn't have like a widget creation in the undo protocol (undoing it removes the widget again)". The code does the opposite — it `continue`s, recording nothing, and the inline comment explicitly says removing it "could delete a widget that others do have". The code behaviour is the right one; please update the description (it becomes the commit message and is what a future reader will find when bisecting).

**2. Root cause is fixed at the one known emitter, not at the sink (`server/room.mjs:759`).**
`Room.receiveDelta` still treats `this.state[widgetID] === undefined` as "new widget" and stores whatever partial object it is given. Player-originated deltas are protected by the `!this.room.state[widgetID] && !delta.s[widgetID].id` check in `server/player.mjs:121-127`, and after this PR `receiveInvalidDelta` is the only other caller — so the fix is complete *today*. Since `receiveInvalidDelta` bypasses the player-level guard, consider also skipping entries with no `id` for an unknown widget inside `Room.receiveDelta` itself, so a future internal emitter can't reintroduce exactly this bug. Optional, but cheap.

**3. Desynced clients now survive but stay desynced, silently and noisily (`client/js/serverstate.js:365`).**
The new `console.error` turns a crash into a permanent, per-delta log line: once a client is missing a widget the server has, every subsequent delta touching it logs again and the client never recovers (there's no client-initiated resync path). That's clearly better than dying, but it is worth deciding whether this state should trigger something stronger — e.g. asking the server for a full state — rather than only a console message. Also note the message will fire for widgets `addWidget` deliberately *deferred* (a card whose deck, or a child whose parent, isn't present yet — `serverstate.js:54-58,79-83`), where "this client does not have the widget" is technically true but "out of sync with the server" is misleading. Reaching that across deltas is unlikely with the server's ordering, so this is a wording nit at most.

**4. No automated coverage for the server-side half.** The new jest test only covers `addDeltaEntryToUndoProtocol`; the `room.mjs` change (which is the one that stops game state being corrupted for everyone) is verified only by the manual repro. `tests/server/` currently has no jest harness at all, so this is understandable — just flagging that the regression that actually corrupts room state is untested. I ran `npm test` on the branch: 94 passed, 7 suites.

**Verified as fine:** the `widgets.has(widgetID)` guard on the `setLimbo` pass matches the existing "addWidget above MIGHT have failed" defensiveness two loops down; the `continue` branch is correctly placed after the `id` (creation) branch so real creations still record `null`; and omitting a widget from `undoDelta` interacts correctly with `mergeDeltas`, which keeps the previous entry for keys absent from the newer undo delta.

This PR changes engine behaviour (client delta application and the undo protocol in `client/js/serverstate.js`), so it is excluded from beta auto-deployment by design and I did not add the `beta` label.

VTT-AI-Agent:
Thanks — three of the four points led to changes (`d80ffb2c`, `8dd1eb95`), and the PR description is corrected.

### 1. PR description contradicted the code — fixed

You're right, and thanks for catching it: the description still described the *first* iteration of the client change, which was replaced in `a66af089` after auto-review 1/4 flagged the same hazard you did. The "The change" section now describes what the code actually does (records **no** undo entry for a property-only delta on an unknown widget; only a real creation is undone by removing the widget).

### 2. Guard at the sink in `Room.receiveDelta` — added

Two reviewers have now asked for this, so I've added it (`d80ffb2c`):

```js
      } else if(this.state[widgetID] === undefined) {
        if(delta.s[widgetID].id === undefined) {
          // storing this would add the widget to the room as a partial widget and break every client that receives it
          Logging.log(`WARNING: ignoring delta data for missing widget ${widgetID} in room ${this.id}`);
          delete delta.s[widgetID];
          continue;
        }
        this.state[widgetID] = delta.s[widgetID];
```

Also `delete`s the entry so it isn't broadcast — otherwise the server would ignore it locally but still ship it to every client, which is the half that actually crashed people. It's unreachable today (player deltas are rejected by `Player.receiveDelta:121-127`, and `receiveInvalidDelta` no longer emits this shape), so it changes no current behaviour; the `Logging.log` means a future internal emitter surfaces in the room log instead of silently corrupting state.

### 3. Wording / deferred widgets — fixed; resync — deliberately out of scope

The false positive is real, so the message is now limited to entries that don't create a widget (`8dd1eb95`):

```js
    } else if(delta.s[widgetID] !== null && delta.s[widgetID].id === undefined) { // check id because adding a widget can legitimately be deferred until its parent or deck arrives
```

A deferred child or a card whose deck hasn't arrived yet carries an `id`, so it no longer logs; a creation that *failed* still logs from inside `addWidget` itself (`serverstate.js:63,76,117`). What's left is exactly "properties changed on a widget I don't have", where "out of sync with the server" is accurate.

On the stronger response: there is no client-initiated resync in the protocol today — `Player.messageReceived` has no "send me the state" command, and full state is only ever pushed by the server (on join, on `receiveInvalidDelta`, and after `waitingForDeltaFromPlayer`). Adding one is a protocol change plus a policy decision about when a client is diverged enough to warrant it, and it needs care not to let a persistently broken client hammer the server with state requests. That's its own PR, not something to fold into a crash fix.

### 4. Server-side coverage — added

Fair point, and it turned out cheaper than expected: `Room.prototype.receiveDelta`/`receiveInvalidDelta` don't touch the constructor's file I/O, so `tests/server/room-delta.test.js` drives them on a minimal room object (`Object.create(Room.prototype)` with a stubbed `broadcast`). Five cases:

* a property change to a known widget is applied and broadcast (so the guard doesn't swallow normal traffic),
* a delta creating an unknown widget still adds it,
* a property-only delta for an unknown widget neither adds it nor is broadcast — **fails without `d80ffb2c`**,
* the shadow cleanup unsets `dropShadowWidget` when the original still exists,
* the shadow cleanup does not resurrect the original when it was deleted — **fails without the `receiveInvalidDelta` guard**, i.e. this is the reported crash's root cause under test.

Verified: `npm test` — 99 passed, 8 suites; TestCafe `routines.js` + `editor.js` — 21 passed.

### `beta` label

Agreed, and not added — this is engine behaviour in `client/js/serverstate.js` and `server/room.mjs`. That call is a reviewer's to make anyway.

VTT-AI-Agent:
### Another report of this crash from the main server — same code path, but no server delta involved

A new client crash report came in from virtualtabletop.io (2026-08-19, Firefox 154) with this stack:

```
can't access property "unalteredState", K.get(...) is undefined
    at addDeltaEntryToUndoProtocol
    at receiveDelta
    at sendRawDelta
    at click (edit.js)
    at onClick
    at onKeyDown
```

Same line as the one this PR guards, but it arrives from the other side: not `receiveDeltaFromServer`, but the editor's **Undo** button (`UndoButton.click` → `sendRawDelta({s: undoDelta})`). No second client, no invalid delta, no drop shadow — so the two `server/room.mjs` guards in this PR would not have caught it, while the two `client/js/serverstate.js` guards do.

What the report shows (in generalised terms): the client was in edit mode, the undo protocol still held several property-only entries for a widget that was no longer in the client's `widgets` map, and replaying every recorded delta of the protocol from the start yields exactly the room the client had **plus** that one widget — i.e. it vanished from the map without any delta recording it. Pressing Undo then replayed a property-only entry for it and killed the session (the error overlay also closes the WebSocket, and because `sendDelta` never gets to reset its buffer, every later delta throws again).

### Reproduction on current `main` (044a8bb4), single client

The route in is the one already named in this PR's discussion — `addWidget` silently bailing out for a card without a deck:

1. Open a room, enter edit mode, add a widget.
2. Change a property of it, e.g. move it (one history entry with a property-only undo delta).
3. Turn it into a card while nothing with its `deck` id exists — `"type": "card"` in the JSON editor, or a `SET` operation on `type`. `receiveDelta` removes and re-adds the widget for the type change, `addWidget` defers a card with no deck, and the widget is gone from this client while the server still has it.
4. Press Undo.

Result on clean `main`: `clientErrorOverlay` with `Cannot read properties of undefined (reading 'unalteredState')` at `addDeltaEntryToUndoProtocol`, connection closed.

With this PR's `client/js/serverstate.js` hunks applied on top of current `main` (they still apply cleanly, `git apply --3way`), the same sequence undoes through the whole protocol without an error, the client stays connected and further edits still work.

The branch is currently ~3 weeks behind `main`, so it probably wants a rebase before it goes anywhere.

VTT-AI-Agent (commented):
## TL;DR

The fix is sound and, as far as I can tell, complete for both reported crash variants — I verified that all four hunks are load-bearing and that the branch still applies cleanly to current `main` with the full jest suite green (41 suites / 1356 tests). Nothing here blocks merging. The one thing I would still do before merge is add a test for the `setLimbo` guard: it is the only hunk without coverage, and I confirmed it is exactly the hunk that saves the *newest* reported crash (Undo of a move, whose undo delta carries `parent`) — with the other two client hunks in place but that `widgets.has` removed, the same delta still throws `Cannot read properties of undefined (reading 'setLimbo')`.

## Findings

### 1. `setLimbo` guard is untested and is the hunk that catches the new report — `client/js/serverstate.js:347`

The undo protocol half is now covered by `tests/client/undo-protocol.test.js`, but the guard on line 347 is not, and it is not redundant. The 2026-08-19 report you added to the PR description is an **Undo** of a change that included `parent` (any move into/out of a holder produces `{parent, x, y}`), so the recorded undo delta is a property-only entry *with* `parent` for a widget that has since vanished from the client's `widgets` map.

I checked this against the branch: with `addDeltaEntryToUndoProtocol` fixed but the `&& widgets.has(widgetID)` removed from line 347, feeding `receiveDelta({s: {gone: {parent: '<existing holder>', x: 5, y: 5}}})` throws

```
TypeError: Cannot read properties of undefined (reading 'setLimbo')
    at receiveDelta (client/js/serverstate.js, the setLimbo loop)
```

and with the guard in place it survives and logs the two console errors. So the three client hunks are one fix and the test suite currently pins only one of them.

Cheap way to close it, following the precedent this PR already set by exporting for tests: export `receiveDelta` as well and add one case (`tests/client/undo-protocol.test.js` or a small `receive-delta.test.js`):

```js
test("a property-only delta with parent for an unknown widget does not throw", () => {
  createWidget({ id: 'holder', type: 'widget' });
  expect(()=>receiveDelta({ s: { gone: { parent: 'holder', x: 5, y: 5 } } })).not.toThrow();
  expect(widgets.has('gone')).toBe(false);
});
```

That pins the reported scenario end-to-end instead of just its recording half, and it fails on `main`.

### 2. The `Room.receiveDelta` guard is no longer the only server-side delta sink after a rebase — `server/room.mjs:760`

The branch is based on `950d32c9` and is ~3 weeks behind `main` (`044a8bb4`); GitHub still reports `MERGEABLE`, and I confirmed the patch applies to current `main` with `git apply --3way` and that the whole jest suite passes there. But `main` has since grown a second server-side delta emitter that does **not** go through `Room.receiveDelta`: `renamePlayerInWidgets` (`server/room.mjs` on `main`, ~line 936) builds a delta and calls `this.broadcast('delta', delta)` directly (line 970).

It cannot produce the bad shape — it only writes properties of widgets that are already in `this.state` — so this is not a bug, and it is in fact a nice illustration that the client-side guards are the ones doing the real work (a client that is missing that widget, e.g. a deferred card, receives a property-only entry for it). But the PR description's "this only closes the door for server-generated deltas" is no longer literally true after the rebase, and a future emitter added next to `renamePlayerInWidgets` would silently bypass the new guard. Worth one sentence in the comment at `room.mjs:761` (or moving the check next to `broadcast`) so the invariant isn't assumed to hold room-wide.

### 3. `addWidget` is still called for entries that can never create a widget — `client/js/serverstate.js:350-352`

```js
for(const widgetID in delta.s)
  if(delta.s[widgetID] !== null && !widgets.has(widgetID))
    addWidget(delta.s[widgetID]);
```

For the property-only entries this PR is about, this call can only ever fail, and it has two visible side effects (both observed in the run above):

* it logs `Could not add widget! {…} widget without ID` (`serverstate.js:63`) *in addition to* the new `Could not apply delta for widget …` — two errors per affected widget per delta, for one problem;
* worse, when the entry carries a `parent` the client does not have, `addWidget` never reaches the `if(!id)` check: the first branch (`serverstate.js:54-58`) pushes the id-less fragment into `deferredChildren[parent]`, where it stays forever and re-logs `widget without ID` every time a widget with that parent ID shows up.

Adding `&& delta.s[widgetID].id !== undefined` to that loop makes the intent explicit ("only entries that create a widget go to `addWidget`"), leaves exactly one error message per affected widget, and keeps the deferred-children queue free of fragments. It changes nothing for well-formed deltas.

### 4. The two guards disagree on what "creates a widget" means — `server/room.mjs:760`, `client/js/serverstate.js:365`

`Player.receiveDelta` (`server/player.mjs:123`) rejects with `!delta.s[widgetID].id`, the new room guard uses `delta.s[widgetID].id === undefined`, and `addDeltaEntryToUndoProtocol` uses the truthy `delta.s[widgetID].id`. An entry with `id: null` (or `''`) therefore slips past the room guard and is stored — and persisted into the save file — as `{ id: null, … }`, i.e. exactly the partial-widget shape the guard exists to prevent; clients then drop it silently, because the new `console.error` condition (`id === undefined`) doesn't match it either. Unreachable today, but since both are pure backstops for future emitters, `!delta.s[widgetID].id` in `room.mjs:760` and `!delta.s[widgetID].id` in `serverstate.js:365` would make all four sites agree.

## Verification I ran

* `npm test` on the PR head (`8dd1eb95`): 8 suites / 99 tests pass, including the two new suites.
* PR patch applied onto current `main` (`044a8bb4`) in a scratch worktree: applies cleanly, `npm test` → 41 suites / 1356 tests pass. So the rebase is mechanical.
* Confirmed each client hunk is load-bearing by removing it individually (see finding 1).
* Checked that the new `console.error` can't reach the crash overlay: `client/js/tracing.js:98-101` only hooks `window.onerror` / `unhandledrejection`, not `console.error`.
* All CI checks on the PR are green.

## Standing maintainer checks

* **Editor UI**: the PR adds no widget property and no routine function/operation, so there is nothing new for game authors to reach — no property-editor or routine-block UI is required.
* **Test room / library**: this is a multiplayer race and an editor-Undo crash; there is no author-facing feature to demo, and the diff touches nothing under `library/` (four files: two engine files, two tests). Nothing to flag.

I did not add the `beta` label: the diff changes the engine for real — delta application and undo-protocol semantics in `client/js/serverstate.js` and delta handling in `server/room.mjs` — so the PR is excluded from beta auto-deployment by design.

VTT-AI-Agent:
Thanks — all four points led to changes (`66a9c37e`, `9d092268`), and the branch is now rebased onto current `main` (`044a8bb4`).

### 1. `setLimbo` guard untested — covered now

You're right that the test suite pinned only the recording half. New `tests/client/receive-delta.test.js` drives the whole function with exactly the delta from the 2026-08-19 report (a property-only entry carrying `parent`, for a widget this client does not have) and needs `receiveDelta` exported, as you suggested:

```js
const delta = { s: { [unknownID]: { parent: holderID, x: 5, y: 5 } } };
```

I confirmed each hunk individually against it:

* without the `widgets.has` on the `setLimbo` pass → `Cannot read properties of undefined (reading 'setLimbo')`
* without the `addDeltaEntryToUndoProtocol` guard → `Cannot read properties of undefined (reading 'unalteredState')`
* without the `addWidget` guard (finding 3) → the "reported once" case fails with `Expected: 1, Received: 2`

So all three client hunks are pinned now.

### 2. `renamePlayerInWidgets` bypasses the room guard — noted where it matters

Rebasing brought it in, and it is the right caveat: the guard in `Room.receiveDelta` is not a room-wide invariant. It only writes properties of widgets already in `this.state`, so it cannot produce the bad shape today; I left the code alone and put the limitation in the comment next to the guard, so the next emitter added beside it doesn't assume the sink covers it:

```js
        if(!delta.s[widgetID].id) {
          // storing this would add the widget to the room as a partial widget and break every client that receives it.
          // deltas that are broadcast without going through here (see renamePlayerInWidgets) have to keep this invariant themselves.
```

The PR description's "this only closes the door for server-generated deltas" is corrected accordingly.

### 3. `addWidget` called for id-less entries — fixed

```js
  for(const widgetID in delta.s)
    if(delta.s[widgetID] !== null && delta.s[widgetID].id && !widgets.has(widgetID)) // check id because an entry without one can not create a widget
      addWidget(delta.s[widgetID]);
```

Both side effects you describe are gone: one error message per affected widget instead of two, and no id-less fragment left in `deferredChildren`. The `deferredChildren` push happening *before* the `if(!id)` check is what made the second one permanent — worth noting for anyone reading `addWidget` later.

### 4. The four checks disagreed — they now all use a falsy `id`

`room.mjs` and the `console.error` condition in `serverstate.js` both use `!delta.s[widgetID].id`, matching `Player.receiveDelta` and `addDeltaEntryToUndoProtocol`. `tests/server/room-delta.test.js` gained a case for `{ id: null, x: 5 }` on an unknown widget: not stored, not broadcast. I left the pre-existing `id === undefined` in the `setLimbo` condition as it is — it is guarded by `widgets.has(widgetID)`, so a falsy-id entry can't reach it, and changing it would touch a line this PR has no reason to.

### Verification

* `npm test` on the rebased branch — 42 suites / 1360 tests pass
* TestCafe `routines.js` + `editor.js` — 75 passed
* Each of the four guards removed individually → a test fails

### `beta`

Agreed, and not added — engine behaviour in `client/js/serverstate.js` and `server/room.mjs`. That call is a reviewer's anyway.

VTT-AI-Agent:
### Another report of this crash from the main server — the drop-shadow scenario this PR describes

A new client crash report came in from virtualtabletop.io (2026-08-29, Firefox 154). Production serves the minified bundle, so the stack is unrecognisable at first glance:

```
can't access property "unalteredState", te.get(...) is undefined
xe/<@https://virtualtabletop.io/<bundle>:1:224047
xe@https://virtualtabletop.io/<bundle>:1:224335
Te@https://virtualtabletop.io/<bundle>:1:225240
J/z.onmessage@https://virtualtabletop.io/<bundle>:1:218970
```

It is this code path. Building the client bundle locally with `minifyJavascript: true` produces the very same identifiers, and terser inlines `addDeltaEntryToUndoProtocol` into `receiveDelta` as an anonymous IIFE — which is why Firefox shows it as `xe/<`:

```js
function xe(e){!function(e){const t={};for(const a in e.s)if(null===e.s[a])te.has(a)&&…
  else{t[a]={};for(const n in e.s[a])t[a][n]=te.get(a).unalteredState[n],…}}
function Te(e){se=e.id,xe(e)}
```

So `te` is `widgets`, `xe/<` is `addDeltaEntryToUndoProtocol` crashing on `widgets.get(widgetID).unalteredState[property]`, `xe` is `receiveDelta`, `Te` is `receiveDeltaFromServer` and `z.onmessage` is the WebSocket handler in `startWebSocket` — the exact stack this PR fixes.

The room situation matches the drop-shadow scenario in the description rather than the editor-Undo one: a multiplayer game in play mode, no JSON editor open, with two players repeatedly dragging the *same* card in and out of a holder with `dropShadow`, so shadow widgets were being created and removed by both of them within a few hundred milliseconds of each other. The undo protocol ends on a delta that removes one player's shadow widget while the other player's drag is still in flight — exactly the conflict that puts `receiveInvalidDelta` in charge of the cleanup.

### Still reproducible on current `main` (16f942f0)

Ran the reproduction from the description again against today's `main` — one real browser client plus one scripted WebSocket client:

1. Browser client A opens a room.
2. Client B sends a delta creating `orig` and a shadow cloned from it (`clonedFrom: 'orig'`, `dropShadowOwner: 'B'`).
3. Client B deletes `orig`.
4. Client B sends a delta the server rejects (`{ s: { somethingGone: { x: 5 } } }` — a property-only entry for a widget the server does not have, which is what an in-flight drag produces once someone else removed the widget), so `receiveInvalidDelta` runs.

```
client A has orig/shadow: [ true, true ]
client A has orig after delete: false
--- page errors ---
    at addDeltaEntryToUndoProtocol (…:2756:62)
    at receiveDelta (…:2694:3)
    at receiveDeltaFromServer (…:2827:3)
    at connection.onmessage (…:2320:7)
server state has orig again: true {"dropShadowWidget":null}
```

On this branch's head (`9d092268`) the same sequence produces no page error at all, and the deleted widget is no longer resurrected in the room state:

```
client A has orig/shadow: [ true, true ]
client A has orig after delete: false
--- page errors ---
(none)
server state has orig again: false undefined
```

No change requested — just a second data point that this is still happening in the wild. The branch is `BEHIND` `main` at the moment, so it will need a rebase before it can go in.

VTT-AI-Agent (commented):
## TL;DR

Everything raised in the previous rounds is addressed on `9d09226` — the `setLimbo` guard now has an end-to-end test, `addWidget` only sees entries that can create a widget, the `renamePlayerInWidgets` caveat is in the comment, and all four "does this entry create a widget" sites agree on falsy `id`. I found no correctness bug. The one thing left is that the falsy-`id` normalisation stops at the `state[widgetID] === undefined` branch: `{id: null, x: 5}` for a widget the room **does** have still strips the `id` out of the room state (and the save file), which is the shape the PR description claims is now impossible.

## Findings

### 1. `{id: null}` for an *existing* widget still produces the partial widget the guard is about — `server/room.mjs:788-794` (low severity, description is wrong either way)

The new guard sits in the `this.state[widgetID] === undefined` branch only. The property-merge branch below it treats `null` as "unset this property", and `id` is not excluded — so a delta entry with `id: null` for a widget the room already has deletes the widget's `id` from the room state.

Verified against this branch:

```js
const room = createRoom({ w1: { id: 'w1', type: 'label', x: 1 } });
room.receiveDelta(player, { s: { w1: { id: null, x: 5 } } });
// room.state.w1          === { "type": "label", "x": 5 }     <- no id
// broadcastedDeltas[0].s === { "w1": { "id": null, "x": 5 } } <- forwarded verbatim
```

That entry goes into `writeToFilesystem()`/`writeStateToFilesystem()` as a widget without an `id`, and on the next full-state load every client drops it with `Could not add widget! … ID does not equal key in state` (`client/js/serverstate.js:517`) — the widget silently disappears from the game for everyone. It is the same failure mode the PR is closing, one branch over.

Not reachable from the stock client (nothing ever sends `id: null`), so this is hardening rather than a bug fix — but the PR description states it *is* closed:

> Every check for "does this entry create a widget" now treats a falsy `id` the same way `Player.receiveDelta` does, so an entry like `{ id: null, x: 5 }` cannot slip into the room state (and into the save file) as a partial widget either.

That sentence is only true when the room doesn't already have the widget. Either narrow the sentence, or skip a null `id` in the property loop as well (`if(property == 'id' && delta.s[widgetID][property] === null) continue;`) so the invariant "every widget in `this.state` has an `id`" actually holds room-wide. The description becomes the commit message, so I'd rather not leave it claiming more than the code does.

### 2. The limbo pass is the last site still using `id === undefined` — `client/js/serverstate.js:356`

```js
if(delta.s[widgetID] && delta.s[widgetID].parent !== undefined && delta.s[widgetID].id === undefined && widgets.has(widgetID))
```

The other three sites this PR touches now use falsy `id` (`:360`, `:366`, `room.mjs:778`). For `{ id: null, parent: 'p' }` on a widget this client *has*, the limbo pass is therefore skipped while the apply loop still re-parents the widget — exactly the "avoid cyclic children" case the limbo pass exists for. Same reachability caveat as finding 1 (pre-existing, no stock client produces it), but since this PR is explicitly making these four checks agree, `!delta.s[widgetID].id` here would finish the job — and it can only *widen* the limbo pass, so it cannot regress a well-formed delta.

### 3. Optional, same guard: an entry whose `id` doesn't match its key — `server/room.mjs:778`

```js
room.receiveDelta(player, { s: { keyA: { id: 'other', type: 'label' } } });
// room.state.keyA === { "id": "other", "type": "label" }
```

`Player.receiveDelta` only rejects a *falsy* `id`, so this passes and is stored under the wrong key; other clients then `addWidget` it as `other`, and the next full-state load rejects it outright. `receiveStateFromServer` already defends against this shape client-side (`client/js/serverstate.js:515-518`) — `delta.s[widgetID].id !== widgetID` in the new server guard would make it impossible instead. Pre-existing and out of the reported crash's path, so purely take-it-or-leave-it.

### 4. Context worth linking, not a change request

`client/js/editor/selection.js:325-328` on `main` carries the comment *"the server re-creates an unknown id as a typeless widget that then ends up in the saved game, so drop it from the selection first"* — i.e. someone already hit this exact server behaviour and worked around it on the client. The `Room.receiveDelta` guard here fixes that at the source; worth a cross-reference in the PR description so this isn't read as a purely hypothetical backstop.

## Verification I ran

* Full jest suite on the PR head (`9d09226`): **42 suites / 1360 tests pass**, including the three new suites (13 tests).
* The branch is `BEHIND` `main` (base `044a8bb4`, `main` now at `16f942f0`, which touches `generateUniqueWidgetID`/`addWidgetLocal` in the same file). I merged `origin/main` into the branch locally: **merges cleanly, no conflicts**, and the merged tree passes the full suite — **52 suites / 1536 tests**. So the rebase is mechanical.
* Re-confirmed each client guard is still load-bearing *and now covered*: removing `&& widgets.has(widgetID)` from `:356` fails 3 of the new tests (`Cannot read properties of undefined (reading 'setLimbo')`); reverting the `.id` check on `:360` fails the "reported once" test with a second `Could not add widget! … widget without ID`. Previous round's finding 1 is properly closed.
* Confirmed the new `console.error` cannot raise the crash overlay: `client/js/tracing.js:317,326` hooks only `window.onerror` and `unhandledrejection`.
* All 45 GitHub checks green (build, CodeQL, both browsers' TestCafe matrix).

## Standing maintainer checks

* **Editor UI** — the PR adds no widget property and no routine function/operation, so there is nothing new for game authors to reach; no property-editor or routine-block UI is required.
* **Test room / library** — this is a multiplayer delta race plus an editor-Undo crash, with no author-facing feature to demo, and the diff touches nothing under `library/` (five files: two engine files, three jest suites). Nothing to flag.

I did **not** add the `beta` label, and this is not incidental engine contact: `client/js/serverstate.js` really changes engine behaviour for every game — which delta entries reach `addWidget`/`setLimbo` and what the undo protocol records — so the PR is excluded from beta auto-deployment by design.

VTT-AI-Agent:
Thanks — findings 1 and 2 are fixed, finding 4 is in the description, and I left finding 3 alone on purpose. `main` is merged into the branch as well.

### 1. `{id: null}` for a widget the room *does* have — fixed (`cdda1983`)

The property-merge branch now ignores an entry that unsets a widget's `id`:

```js
if(property == 'id' && !delta.s[widgetID][property]) {
  // every widget needs its id - one without it is dropped by every client on the next full state load
  Logging.log(`WARNING: ignoring attempt to unset the id of widget ${widgetID} in room ${this.id}`);
  delete delta.s[widgetID][property];
}
```

I dropped the property from the delta rather than only skipping it, because the delta is broadcast verbatim: `StateManaged.applyDelta` deletes any property whose value is `null`, so keeping `id: null` in the broadcast would leave every *client's* widget without an `id` while the server still has one — the divergence moved one layer down instead of being closed. With the `delete`, `{ id: null, x: 5 }` on an existing widget now applies as `{ x: 5 }` everywhere.

Covered by a new case in `tests/server/room-delta.test.js` ("Then a delta unsetting its id keeps the id and does not broadcast the unset"); reverting the guard fails it.

### 2. The limbo pass now uses the same falsy-`id` test — fixed (`da233b8b`)

`client/js/serverstate.js:348` is `!delta.s[widgetID].id` now, so all four sites agree. You're right that it can only widen the pass. New test in `tests/client/receive-delta.test.js` feeds `{ id: null, parent: holder }` for a widget the client has and asserts the widget goes to limbo first (the assertion stubs `setLimbo` on the instance — the real one needs `#topSurface`, which jsdom doesn't have here). Reverting to `id === undefined` fails it.

### 3. `id` that doesn't match its key — not changing it

I'd rather not fold this in. It is a different invariant from the one this PR is about ("every widget in the state has an `id`" vs "its `id` is its key"), it is unreachable from the client, and rejecting the creation outright turns a widget that currently shows up for everyone until the next full state load into one that never appears at all — a strictly larger behaviour change than the crash fix warrants, with no reported case behind it. Worth its own PR (together with `Player.receiveDelta`, which is where the shape gets accepted in the first place); happy to open one if you'd like it.

### 4. `client/js/editor/selection.js` — added to the description

Good catch, that comment is the same server behaviour worked around on the client. The description now cites it, so the `Room.receiveDelta` guard doesn't read as purely hypothetical.

### Branch state and verification

* `origin/main` is merged into the branch (`conventions.md` asks for follow-up commits, no force-push, so a merge rather than a rebase) — no conflicts.
* `npm test` on the merged tree — **52 suites / 1538 tests pass**.
* TestCafe `routines.js` + `editor.js` locally — 100 passed. All 46 GitHub checks green on `15dafa82`.
* Each new guard removed individually → its test fails.

VTT-AI-Agent (commented):
## TL;DR

The fix is correct and I found no new bug in the current head (`15dafa82`). Both reported crash variants reproduce on today's `main` and are gone on this branch, verified with a real Chromium client against a local server (details below); the full jest suite passes on the branch and on a scratch merge with current `main`. The one thing a maintainer needs to decide is the overlap with **#3011**: it patches the same three lines of `client/js/serverstate.js` for the same crash, and its regression test asserts the *opposite* of this PR's `console.error` on an unapplicable delta, so the two cannot land as they are.

## Findings

### 1. #3011 fixes the same crash on the same lines and contradicts this PR's diagnostic (needs a maintainer decision)

#3011 ("Fix crash in undo protocol when a widget is removed mid-batch", merged with `main` today) edits exactly the three hunks this PR edits — the limbo pass (`client/js/serverstate.js:355-357`), the `addWidget` pass (`:359-361`) and the unknown-widget branch of `addDeltaEntryToUndoProtocol` (`:413-416`) — with different spellings, so whichever lands second gets a textual conflict in every one of them. Semantically this PR's receive side is the superset: falsy `id` at all four sites instead of `id === undefined`, the `console.error` at `:375`, plus the two `server/room.mjs` guards and the `receiveInvalidDelta` fix, none of which #3011 has. What #3011 adds that this PR does not is the send-side drop in `sendPropertyUpdate()` for a widget that a routine removed and flushed via `DELAY` earlier in the same batch, together with `tests/client/widget-remove-mid-batch.test.js`.

The incompatibility is not only textual. The second test in #3011 spies on `console.error`, feeds `receiveDelta({ s: { <removed id>: { parent } } })` and then asserts `expect(consoleError).not.toHaveBeenCalled()`. With this PR's `:375` that delta *does* log (`Could not apply delta for widget …`) — which is what auto-review 2/4 asked for here — so combining both PRs as they stand fails that test. Suggested resolution: land this PR first, then reduce #3011 to its `sendPropertyUpdate()` guard and first test, and relax the `console.error` assertion in its second test to "no *`Could not add widget`* error" (that is the deferred-children leak it is really pinning). The reverse order works too, but then this PR has to redo all three hunks by hand.

### 2. Not blocking — the second reported variant is survived, not cured

After a `type` change to `card` while the deck is missing, the widget still vanishes from this client while the server keeps it (`receiveDelta` removes it and `addWidget` defers it into `deferredCards`). The client now logs `Could not apply delta` once per later delta touching it and stays diverged until the next full state load. That is the right call for a crash fix and was discussed in earlier rounds; I only note it so the follow-up (a client-initiated resync, or keeping a placeholder for the deferred card) is not forgotten.

## What I verified

| Check | Result |
| --- | --- |
| `npm test` on the PR head | 52 suites / 1538 tests pass |
| Scratch merge of current `origin/main` (`ec1c965e`) into the branch | auto-merges without conflicts; the three new suites pass; 1663 of 1665 tests pass — the two failures are `tests/browsercompat.test.js` missing the `@mdn/browser-compat-data` devDependency in my install, unrelated to this PR |
| Drop-shadow conflict, **`main`**: browser client A + scripted client B create `orig`+`shadow`, delete `orig`, B sends a rejected delta | A receives `{ shadow: null, orig: { dropShadowWidget: null } }`, page error `Cannot read properties of undefined (reading 'unalteredState')`, `#clientErrorOverlay` shown, server state has `orig: { dropShadowWidget: null }` again |
| Same sequence, **this branch** | A receives `{ shadow: null }`, no page error, no overlay, `orig` stays deleted on the server |
| Editor Undo variant, **`main`**: add widget → move → `type: card` (no deck) → Undo twice via `sendRawDelta(undoDelta)` | `TypeError: Cannot read properties of undefined (reading 'unalteredState')` thrown from the first Undo |
| Same sequence, **this branch** | no exception, two `Could not apply delta for widget u1!` console errors, server ends at `{ id: 'u1', type: 'widget', x: 0, y: 0 }` |
| Engine commits on `main` since the branch's last merge (`f887eb26` delta monitor, `7a53a3c7` #3141 missing-parent handling, `51c60a18`) | no interaction: the guards sit on the paths `sendRawDelta` and `receiveDeltaFromServer` still use, and #3141 only changes what happens at full-state load |
| GitHub checks on `15dafa82` | all 46 green (jest, TestCafe Chrome + Firefox matrix, production start, CodeQL) |

## Standing maintainer checks

- **Editing UI / routine blocks:** not applicable — no new widget property, routine operation or function.
- **Test room / library:** not applicable — a crash fix with no author-facing feature to demo; the diff touches nothing under `library/` (two engine files, three jest suites).

I did not add the `beta` label: the change alters engine behaviour for every game — which delta entries reach `addWidget`/`setLimbo` and what the undo protocol records in `client/js/serverstate.js`, plus delta handling in `server/room.mjs` — so it is engine-ineligible for beta auto-deployment rather than an incidental engine edit.

VTT-AI-Agent:
Thanks — no code change was needed for either finding; what I did do is merge current `main` into the branch (it was `BEHIND`) and put the #3011 question where a maintainer will see it.

### 1. Overlap with #3011 — agreed, and it is a maintainer decision

Your reading matches mine: #3011 is still open and untouched, it edits the same three lines of `client/js/serverstate.js` (the `setLimbo` pass, the `addWidget` pass and the unknown-widget branch of `addDeltaEntryToUndoProtocol`) with different spellings, and its second test asserts `expect(consoleError).not.toHaveBeenCalled()` for a delta that this PR deliberately logs — the diagnostic auto-review 2/4 asked for here. So they cannot land as they stand, textually or semantically.

I am not going to change either side of that unilaterally:

* removing the `console.error` to make #3011's assertion pass would undo a change made at a reviewer's request and take away the only signal a diverged client leaves behind, and
* folding #3011's `sendPropertyUpdate()` guard into this PR would swallow another open PR from a separate thread.

So I have added it to the PR description as an explicit "waiting on a human maintainer" item, with your suggested resolution as the default (land this one first, then reduce #3011 to its `sendPropertyUpdate` guard and its first test, and relax the second test's assertion to "no *`Could not add widget`* error", which is the deferred-children leak it is really pinning), and asked the maintainers to pick the order. The description now also has a short "Relationship to #3011" section so the overlap is visible to anyone reading the PR rather than only in this thread.

### 2. The `type: card` variant is survived, not cured — noted, still out of scope

Agreed, and it is recorded here rather than forgotten: after a `type` change to `card` with the deck missing, the widget stays gone on that client and each later delta touching it logs once. Curing it needs either a client-initiated resync (no such command exists in the protocol — full state is only ever pushed by the server) or keeping a placeholder for a deferred card, both of which are behaviour changes well beyond a crash fix. Worth its own PR.

### Branch state

`origin/main` is merged in (`18acdf93`) — no conflicts, and `main` had only moved by one commit since the last merge.

* `npm test` on the merged tree — **59 suites / 1666 tests pass** (your scratch-merge run saw two `tests/browsercompat.test.js` failures from a missing `@mdn/browser-compat-data`; with a clean `npm ci` here that suite passes too).
* Each guard removed individually still fails one of the three new suites.

### `beta`

Agreed and not added — this changes engine behaviour in `client/js/serverstate.js` and `server/room.mjs`, so it is engine-ineligible rather than incidental. That call is a reviewer's anyway.

VTT-AI-Agent (commented):
## TL;DR

The fix is still correct and nothing in the code has changed since the previous review — the only new commit is a clean merge of `main` (`18acdf93`), and on it `npm test` passes (59 suites / 1666 tests) and every GitHub check is green. The one thing that needs action is outside the diff: **#3011 changed shape last night** — it dropped its send-side `sendPropertyUpdate()` guard — so the "waiting on a human maintainer" To Do in this description now proposes a resolution that no longer exists. The maintainer's choice got simpler, and the description should say so.

## Findings

### 1. The #3011 To Do block is stale — #3011 no longer has a send-side guard to be "reduced to"

The To Do item and the "Relationship to #3011" section say #3011 fixes the crash "from the send side" and suggest landing this PR first and then reducing #3011 to its `sendPropertyUpdate` guard, "the part this PR does not have". Since `9c60a11c` ("Let the server's bad-state recovery handle stray property updates", 2026-09-04 23:42 UTC) that guard is gone: #3011's diff is now exactly the same three receive-side hunks of `client/js/serverstate.js` that this PR edits (the limbo pass, the `addWidget` pass and the unknown-widget branch of `addDeltaEntryToUndoProtocol`, with different spellings), a one-line `import { dropTargets } from './main.js'`, and `tests/client/widget-remove-mid-batch.test.js`.

I ran #3011's test file on top of this branch (scratch worktree, with its import line added):

| #3011 test | On this PR's code |
| --- | --- |
| 1 — `DELETE` + `DELAY` flush + later property update in the same batch (the crash #3011 was opened for) | **passes** unchanged |
| 2 — stray `parent` update for a widget removed by an earlier delta | fails **only** on `expect(consoleError).not.toHaveBeenCalled()`, tripped by this PR's `Could not apply delta …` at `client/js/serverstate.js:382` |

So on the receive side this PR is a strict superset of #3011, and the decision is now simply an ordering one: **(A)** land this PR and shrink #3011 to its `dropTargets` import plus its two tests, with the second test's assertion narrowed to "no *`Could not add widget`* error" (that is the `deferredChildren` leak it really pins); or **(B)** land #3011 first and re-apply this PR's three hunks on top. Please rewrite the To Do item and the "Relationship" section to that choice — a worker acting on the current text would go looking for a send-side guard that no longer exists.

On the `dropTargets` import, the one line #3011 uniquely carries: `removeWidget()` in `client/js/serverstate.js` uses `dropTargets` without importing it. In the browser that is fine because `server/minify.mjs` strips the imports and concatenates the modules into one scope, but under jest any test that pushes a `null` entry through `receiveDelta` throws `ReferenceError: dropTargets is not defined`. This PR's new tests never remove a widget through `receiveDelta`, which is why they pass without it — worth taking that line into this PR if option (A) is chosen and #3011 ends up closed rather than trimmed.

### 2. Not blocking — with the send side unguarded, the #3011 scenario now costs a full-state resync instead of a crash

Neither PR now stops the stray property update from being *sent*: in #3011's first scenario the client still ships `{ <deleted id>: { text: … } }` (visible as this PR's `Could not apply delta` log in that test), `Player.receiveDelta` rejects it as a `<modification>` of a missing widget and answers with the full room state. That is the recovery #3011's latest comment explicitly relies on, and it is the right call for a crash fix — but it means every run of such a routine costs the running client a full-state reload plus a conflict warning in the server log. Recording it here so the follow-up (drop the entry client-side, as #3011 originally did, once the alert flake it was blamed for is understood) is not forgotten.

## What I verified

| Check | Result |
| --- | --- |
| `npm test` on `18acdf93` | 59 suites / 1666 tests pass, incl. `room-delta`, `undo-protocol`, `receive-delta` |
| GitHub checks on `18acdf93` | all green: 48 TestCafe jobs, jest build, Production Environment, Browser compatibility, CodeQL |
| Branch vs `origin/main` (`9a40e5bd`) | up to date; the merge commit is the only change since the last review, no code hunk differs |
| #3011's regression tests on this branch | 1 passes, 1 fails only on the `console.error` assertion (table above) |
| Server-side emitters of delta entries | `Room.receiveDelta` is reached only from `Player.receiveDelta` and `receiveInvalidDelta`; `renamePlayerInWidgets` still bypasses it but only ever emits for widgets present in `state`, so the no-partial-widget invariant holds there too |
| Browser reproductions (drop-shadow conflict, editor Undo) | not re-run: the code is byte-identical to the head the previous review verified in a real Chromium client, and `main` only gained #3168 (zip compression in the download route) since |

## Standing maintainer checks

- **Editing UI / routine blocks:** not applicable — no new widget property, routine operation or function.
- **Test room / library:** not applicable — a crash fix with nothing author-facing to demo; the diff touches nothing under `library/`.

I did not add the `beta` label (it was not present before either): the diff changes how `client/js/serverstate.js` applies and records every incoming delta (limbo, `addWidget` and undo passes) and how `server/room.mjs` stores and broadcasts deltas for every game, so it is a real engine change and engine-ineligible for beta auto-deployment rather than an incidental engine edit.

VTT-AI-Agent:
### The #3011 overlap is decided — its send-side guard is folded in here and #3011 is closed

The landing-order question from the last review went to the maintainers, who answered:

> Close #3011 and fold its sendPropertyUpdate guard into #3093

Done in 913bafd4, and #3011 is closed with a comment pointing here.

**What came over from #3011** (its three receive-side guards were already here, in the wider falsy-`id` form):

* `sendPropertyUpdate` now drops a scalar property update for a widget that is no longer in this client's `widgets` map:

  ```js
    if(delta.s[widgetID] === undefined) {
      // a routine can remove a widget and flush that removal (DELAY does) and then still
      // update a property of it - the widget is gone here and on the server, so drop it
      if(!widgets.has(widgetID))
        return;
      delta.s[widgetID] = {};
    }
  ```

  The scenario is a routine that removes a widget and flushes the removal before it ends (`DELAY` calls `flushDelta()`), then updates a property of it through a collection the `DELETE` did not clear. Such an update has nothing left to change locally and nothing to change in the room; sending it only earns a `client is in bad state` rejection from `Player.receiveDelta` and a full state resend. The removal (`property === null`) and full-state (`typeof property === 'object'`) branches are untouched, so creating a widget whose local `addWidget` was deferred still works.

* `import { dropTargets } from './main.js'` — `serverstate.js` uses `dropTargets` in `updateWidgetId`, `receiveStateFromServer` and `removeWidget` without importing it. In the browser that resolves only because the bundle concatenates the client files; under jest it is a `ReferenceError`, which is why the new test needs it.

* `tests/client/widget-remove-mid-batch.test.js` — a removal flushed inside a batch followed by a property update for the removed widget: ending the batch does not throw, the widget is not resurrected, and the update is not sent to the server (pinned through the delta monitor's pending count). Reverting the guard fails the last one.

I did **not** take #3011's second test, which asserts that receiving a property-only delta for a removed widget logs no `console.error` — that contradicts the diagnostic auto-review 2/4 asked for here. The path it covers is already covered by `tests/client/receive-delta.test.js`, including the `deferredChildren` leak it was really pinning ("Then it is reported once").

**One caveat, stated plainly:** #3011 had this guard and removed it again (9c60a11c) after the TestCafe test *"Renaming a widget keeps its color controls clear and it movable"* failed 4/4 CI runs there with a native alert dialog, and went green once the guard was gone. It never reproduced locally there. Here it passes with the guard in place — 3/3 on its own and inside a full `editor.js` run — but that is exactly the evidence #3011 had too, so this PR's CI matrix (Chrome, Firefox and the minified production build) is the real check. If the rename test fails there, the guard comes back out and I will say so rather than re-landing it quietly.

Verified: `npm test` — 60 suites / 1670 tests pass.

VTT-AI-Agent:
### CI is green with the send-side guard — the #3011 rename failure does not reproduce here

Following up on the caveat in the comment above: the guard from #3011 does **not** break the editor suite on this branch.

* `editor.js` locally, full suite: **96/96 passed**, including *"Renaming a widget keeps its color controls clear and it movable"* (also 3/3 when run on its own).
* CI: both `editor.js` jobs pass — Chrome first try, Firefox on a rerun. The one Firefox failure was `Deck editor: create deck from scratch with color box, face and defaults` failing in the shared `setName` helper on `expect(loadingIndicator.exists).notOk()` (the room was still loading), i.e. a setup timing flake and not the rename test; it passed on rerun without any code change.
* All **53 checks** on the PR are green (jest, TestCafe Chrome + Firefox matrix, production start, CodeQL), and `npm test` is 60 suites / 1670 tests locally.

The PR description is updated accordingly.

Issue #2549, opened 2025-04-13
Option for Multiselecting Cards


96LawDawg:
Some games would benefit from an engine solution to select multiple cards (or preselect a single card).  Discussion started in general thread on [Discord](https://discord.com/channels/770758631146782780/770758631146782783/1360941621675425874). This can be accomplished now by an `INPUT` or by custom code, but we could make it easier to implement and more intuitive to use.

The basic idea is this: the engine would override a `clickRoutine` in certain circumstances so that cards could be marked as selected and then the designer could do whatever is appropriate for that.  If we do this, I would want the ability to both select multiple cards and to force the selection of a single card.  For example, in the `INPUT` choose method, you can limit the selection to 1 so that if you click on a second one, the first one is automatically deselected. Or you can have multiple selection enable with a possible maximum value.  To be fair, I think this would be a lot easier to implement if we only did the multiselect and not allow individual card selection.

Here is my idea: we add a new property, call it `clickToSelect`. This would be a property for holders.  If `clickToSelect` is set on that holder, then the normal `clickRoutine` would not be available for cards in that holder. Instead, clicking them would allow them to be "selected" and that shows up visually on the screen.  I think this would need to be an object with 3 keys.  I guess it could be an array that accepts 3 values, but I like an object in this situation.  Either that or we need 3 new properties which seems like a lot.  Structure:

```
"clickToSelect":{
  "owner": (the owner of the cards; if null or not a valid owner name, then all cards in the holder),
  "max": (possible values are 1, a positive whole number, or "all"),
  "indicator": (what will happen to the cards once selected)
```

For "max", if set to 1, then the special condition for selecting one applies.  So for example, if I click on one of my cards, it gets selected and there is some visual indication of that. If I select another card, then the new one is selected and the old one is deselected (for cards I own only).  If "max" is a number, then that is all the cards that I own that can be selected.  And if "max" is "all" then whatever.

When a card is selected, a new property `selectedBy` is added to the card with the name of the player that selected it as the value.

Then for "indicator" we should allow at least a couple of pre-sets.  Something like "highlight", "shade," and "elevate" come to mind.  Not much customization would be allowed here because it would get complicated.  But we could show the css tricks to modify it in a tutorial.  Highlight would draw a border around it, shade will darken or lighten the card, and elevate would raise it a few pixels.

[...]

PR #3094, opened 2026-08-04
Click-to-select for cards in a holder (#2549)

VTT-AI-Agent:
## Goal

Implements #2549: an engine-level way to select cards in a holder, so games no longer have to build "pick a card" / "pick up to N cards" out of `INPUT` dialogs or hand-written click routines.

The issue discussion converged on two holder properties (`multiSelectMax` / `multiSelectStyle`) with a `selectedBy` property on the selected widgets — that is what this implements, plus the two extras that came up in the thread: dragging one selected card takes the whole selection along, and selecting a single card replaces the previous pick instead of requiring a deselect first.

## For players

In a holder that has click-to-select enabled, clicking a card picks it and clicking it again drops it. The pick is shown right away (raised by default), the limit is enforced while clicking, and **dragging one of the picked cards out drags the whole selection with it**: the moment the drag leaves the holder the other picked cards come out with it and travel next to it, so the drag shows what is about to be played.

selection

dragging the selection

The cards that come along are not dragged by hand, so the drop has to check them the way it checked the dragged one: a card the target holder's `dropTarget` rejects, or one that no longer fits its `dropLimit`, goes back into the holder it was picked from and is picked again there — a refused drop costs no selection. The same happens when the drag ends back in the holder it started in, which makes dragging out and back in a way to change your mind.

## For designers

Two new **holder** properties:

| property | values | meaning |
| --- | --- | --- |
| `multiSelectMax` | `0` (default, off), a whole number, `"all"` | how many widgets one player may have selected in this holder at a time |
| `multiSelectStyle` | `"elevate"` (default), `"highlight"`, `"shade"`, `"none"` | how a widget the player picked is shown |

While `multiSelectMax` is set, a click on a widget in the holder toggles its selection **instead of** running its `clickRoutine` or flipping a card. With `multiSelectMax: 1` a new pick replaces the previous one (the pre-selection case from the issue); with a higher limit a full selection stays put until the player deselects something. A `CLICK` operation with `mode: "ignoreClickRoutine"` skips the selection as well, so automations can still flip such a card.

One new property on the selected widgets:

- **`selectedBy`** — the list of players who currently have this widget selected. Several players can select simultaneously in a shared holder, each sees only their own picks. It is cleared when the widget leaves the holder that offers the selection — including when it leaves a pile inside that holder.

Because it is a normal property, `selectedByChangeRoutine` fires on every pick, and routines read the selection with the new `contains` relation of `SELECT` (the mirror image of `in`: the *property* is the list, not a substring test):

```json
{ "func": "SELECT", "type": "card", "property": "selectedBy", "relation": "contains", "value": "${playerName}" },
{ "func": "MOVE", "from": "hand", "to": "table" }
```

The three built-in styles use `--VTTblue`, the colour the app already uses for "the player chose this" in the `INPUT` dialog: `elevate` raises the pick and rings it, `highlight` rings it, and `shade` dims everything the player did *not* pick (a darkened widget means "unavailable" everywhere else in the app, so dimming the picks would read backwards). They are marked `!important` so that a widget dimming or tinting itself through its own `css` cannot hide the fact that the player selected it. `multiSelectStyle: "none"` sets only the `multiSelected` class so a game can style the selection through the holder's `css` property.

A widget in such a holder answers the pointer with a faint ring before the first click, so the holder shows that it reacts to clicks, and a click that runs into the limit flashes the holder instead of doing nothing at all. The widgets that follow a drag stay on the 1600x1000 surface: near the right edge the fan turns around, and a fan too wide for the room left moves closer together.

Both properties are wired into the editor: a "Click to select" / "Selection shown as" row in the holder's Behavior section, insert buttons and value lists in the JSON editor, and validator entries (so they are not reported as unknown properties).

editor

## Compatibility

Everything is opt-in through new properties whose defaults preserve today's behaviour (`multiSelectMax: 0` = off), so **no save migration and no legacy mode** are needed. `selectedBy` only appears on widgets in a holder that enables the feature.

Creator-facing change, so the wiki pages *Widgets* (holder properties, `selectedBy`) and *Functions* (`SELECT` relation `contains`) will want an update.

## Testing

- `npm test` — 188 jest tests pass, including `tests/client/widget-select.test.js` (15 cases: selecting, deselecting, the limit, `multiSelectMax` as a string, the `multiSelectMax: 1` replacement, clearing on leaving the holder and on leaving a pile inside it, the selection being carried out of the holder by the drag, the `dropTarget` and `dropLimit` checks, a refused card going back into the holder still selected, a drag that ends where it started, the spread on the surface, and `SELECT ... relation: contains`).
- `tests/testcafe/multiselect.js` — 5 end-to-end cases covering clicking, the limit, `multiSelectMax: 1`, the drag-along, a card the target holder rejects (which keeps its pick), and the spread on the surface with a holder that stacks its children.
- `tests/testcafe/routines.js` and `publiclibrary-1.js` re-run locally and pass unchanged (no state hashes move — the feature is inert unless a holder opts in).
- Manually verified in a browser against the room on the PR test server: the selection travels with the drag, into another holder, onto the surface, back into the hand, and out of a holder that refuses it.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3094/pr-test (or any other room on that server)

### DIFF
--- a/.github/workflows/testcafe-chrome.yml
+++ b/.github/workflows/testcafe-chrome.yml
@@ -29,6 +29,7 @@ jobs:
           - publiclibrary-3.js
           - publiclibrary-4.js
           - routines.js
+          - multiselect.js
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
--- a/.github/workflows/testcafe-firefox.yml
+++ b/.github/workflows/testcafe-firefox.yml
@@ -29,6 +29,7 @@ jobs:
           - publiclibrary-3.js
           - publiclibrary-4.js
           - routines.js
+          - multiselect.js
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
--- a/client/css/widgets/holder.css
+++ b/client/css/widgets/holder.css
@@ -85,3 +85,23 @@
   position: absolute;
   box-sizing: border-box;
 }
+
+/*
+ * click-to-select: a holder with multiSelectMax lets players pick the widgets in it
+ * by clicking. The picked ones carry the multiSelected class for the player who
+ * picked them, the holder carries multiSelect-<multiSelectStyle> and decides how that
+ * looks. Games that want something of their own set multiSelectStyle to "none" and
+ * style .multiSelected through the holder's css property.
+ */
+.multiSelect-elevate .widget.multiSelected {
+  translate: 0 -14px;
+  filter: drop-shadow(0 4px 4px rgba(0, 0, 0, 0.45));
+}
+
+.multiSelect-highlight .widget.multiSelected {
+  box-shadow: 0 0 0 3px #f5c518, 0 0 8px 4px rgba(245, 197, 24, 0.7);
+}
+
+.multiSelect-shade .widget.multiSelected {
+  filter: brightness(0.65);
+}
--- a/client/css/widgets/holder.css
+++ b/client/css/widgets/holder.css
@@ -114,16 +114,22 @@
 /* the picks stand out because everything else steps back - a darkened widget means
    "unavailable" everywhere else in the app, so dimming the picked ones would read
    backwards. A widget holding a pick (a pile) is left alone, dimming it would dim
-   the pick inside it. Nothing is dimmed while the player has not picked anything. */
-.multiSelect-shade:has(.multiSelected) .widget:not(.multiSelected):not(:has(.multiSelected)) {
+   the pick inside it - its other children are dimmed one level down instead. Nothing
+   is dimmed while the player has not picked anything. Both rules stay on one level of
+   children: a filter already applies to everything inside the widget it dims, so
+   matching nested widgets as well would only darken them a second time. */
+.multiSelect-shade:has(.multiSelected) > .widget:not(.multiSelected):not(:has(.multiSelected)),
+.multiSelect-shade:has(.multiSelected) > .widget:has(.multiSelected) > .widget:not(.multiSelected) {
   filter: brightness(0.65) !important;
 }
 
 /* until the first click nothing tells the player that this holder reacts to clicks,
-   so widgets in it answer the pointer with a hint of the ring they would get */
-.multiSelect-elevate .widget:not(.multiSelected):hover,
-.multiSelect-highlight .widget:not(.multiSelected):hover,
-.multiSelect-shade .widget:not(.multiSelected):hover {
+   so widgets in it answer the pointer with a hint of the ring they would get. A pile
+   is not one of them (its click opens the pile overlay), and :hover reaches every
+   ancestor, so this stays on the holder's own children. */
+.multiSelect-elevate > .widget:not(.pile):not(.multiSelected):hover,
+.multiSelect-highlight > .widget:not(.pile):not(.multiSelected):hover,
+.multiSelect-shade > .widget:not(.pile):not(.multiSelected):hover {
   box-shadow: 0 0 0 2px rgba(31, 92, 166, 0.45);
 }
 
--- a/client/css/widgets/holder.css
+++ b/client/css/widgets/holder.css
@@ -97,15 +97,44 @@
  * important: a card that dims itself with a filter must not be able to hide the fact
  * that the player selected it. Use the "none" style to take that decision yourself.
  */
+/* the raise is relative so that it stays a nudge on a card and on a small token alike,
+   and the ring keeps the pick readable when the whole holder is selected and every
+   widget in it is raised by the same amount */
 .multiSelect-elevate .widget.multiSelected {
-  translate: 0 -14px !important;
-  box-shadow: 0 4px 6px rgba(0, 0, 0, 0.45) !important;
+  translate: 0 -10% !important;
+  box-shadow: 0 0 0 2px var(--VTTblue), 0 4px 6px rgba(0, 0, 0, 0.45) !important;
 }
 
+/* VTTblue is the colour the app already uses for "the player chose this" (the INPUT
+   dialog): a gold ring would read as edit mode's yellow selection ring instead */
 .multiSelect-highlight .widget.multiSelected {
-  box-shadow: 0 0 0 3px #f5c518, 0 0 8px 4px rgba(245, 197, 24, 0.7) !important;
+  box-shadow: 0 0 0 3px var(--VTTblue), 0 0 8px 4px rgba(31, 92, 166, 0.6) !important;
 }
 
-.multiSelect-shade .widget.multiSelected {
+/* the picks stand out because everything else steps back - a darkened widget means
+   "unavailable" everywhere else in the app, so dimming the picked ones would read
+   backwards. A widget holding a pick (a pile) is left alone, dimming it would dim
+   the pick inside it. Nothing is dimmed while the player has not picked anything. */
+.multiSelect-shade:has(.multiSelected) .widget:not(.multiSelected):not(:has(.multiSelected)) {
   filter: brightness(0.65) !important;
+}
+
+/* until the first click nothing tells the player that this holder reacts to clicks,
+   so widgets in it answer the pointer with a hint of the ring they would get */
+.multiSelect-elevate .widget:not(.multiSelected):hover,
+.multiSelect-highlight .widget:not(.multiSelected):hover,
+.multiSelect-shade .widget:not(.multiSelected):hover {
+  box-shadow: 0 0 0 2px rgba(31, 92, 166, 0.45);
+}
+
+/* clicking while the selection is full does nothing, which looks like a broken game.
+   The holder flashes instead, to say that the click arrived and the limit is the
+   reason. Marked as an animation so a holder that styles itself cannot swallow it. */
+.widget.multiSelectFull {
+  animation: multiSelectFull 0.4s ease-in-out;
+}
+
+@keyframes multiSelectFull {
+  0%, 100% { box-shadow: none; }
+  50% { box-shadow: inset 0 0 0 4px rgba(31, 92, 166, 0.8); }
 }
--- a/client/css/widgets/holder.css
+++ b/client/css/widgets/holder.css
@@ -92,16 +92,20 @@
  * picked them, the holder carries multiSelect-<multiSelectStyle> and decides how that
  * looks. Games that want something of their own set multiSelectStyle to "none" and
  * style .multiSelected through the holder's css property.
+ *
+ * The css property of a widget ends up as an inline style, so these rules are marked
+ * important: a card that dims itself with a filter must not be able to hide the fact
+ * that the player selected it. Use the "none" style to take that decision yourself.
  */
 .multiSelect-elevate .widget.multiSelected {
-  translate: 0 -14px;
-  filter: drop-shadow(0 4px 4px rgba(0, 0, 0, 0.45));
+  translate: 0 -14px !important;
+  box-shadow: 0 4px 6px rgba(0, 0, 0, 0.45) !important;
 }
 
 .multiSelect-highlight .widget.multiSelected {
-  box-shadow: 0 0 0 3px #f5c518, 0 0 8px 4px rgba(245, 197, 24, 0.7);
+  box-shadow: 0 0 0 3px #f5c518, 0 0 8px 4px rgba(245, 197, 24, 0.7) !important;
 }
 
 .multiSelect-shade .widget.multiSelected {
-  filter: brightness(0.65);
+  filter: brightness(0.65) !important;
 }
--- a/client/js/editor/propertyInputs.js
+++ b/client/js/editor/propertyInputs.js
@@ -588,6 +588,7 @@ class CheckboxInput extends PropertyInput {
 
 class SelectInput extends PropertyInput {
   // options.choices: [ { value, text } ]
+  // options.customText: value => label for values that are not among the choices
   cssClass() {
     return 'selectInput';
   }
@@ -618,7 +619,7 @@ class SelectInput extends PropertyInput {
         this.select.appendChild(this.customOption);
       }
       this.customOption.value = jsonValue;
-      this.customOption.textContent = `custom: ${jsonValue}`;
+      this.customOption.textContent = this.options.customText ? this.options.customText(value) : `custom: ${jsonValue}`;
     }
     this.select.value = jsonValue;
   }
--- a/client/js/editor/sidebar/properties.js
+++ b/client/js/editor/sidebar/properties.js
@@ -401,7 +401,7 @@ const editorPropertyHints = {
   dropOffsetY: 'Vertical starting position for widgets aligned inside the holder.',
   stackOffsetX: 'Horizontal distance added between consecutively stacked widgets.',
   stackOffsetY: 'Vertical distance added between consecutively stacked widgets.',
-  multiSelectMax: 'Let players pick widgets in this holder by clicking them, up to this many each - or "any number" for no limit. The picked ones get the clicking player\'s name in their selectedBy property, which automations can read. Works best when the holder spreads its widgets out (see Stack offset X/Y): stacked on one spot only the top one can be clicked.',
+  multiSelectMax: 'Let players pick widgets in this holder by clicking them, up to this many each - or "any number" for no limit. The picked ones get the clicking player\'s name in their selectedBy property, which automations can read. Works best when the holder spreads its widgets out (see Stack offset X/Y): stacked on one spot only the top one can be clicked. All players share the one selectedBy list, so in a holder several players pick from at the same time, two clicks on the same widget in the same moment can cost one of them their pick.',
   multiSelectStyle: 'How a widget the player picked is shown. The built-in styles win over a filter or shadow the widget sets itself, so a widget cannot hide that it was picked (which also means it loses its own shadow while picked); "own CSS" only sets the multiSelected class so the holder css can style it.',
   showPlayerColors: 'Use each player\'s color in their scoreboard heading.',
   verticalHeader: 'Rotate the scoreboard header text vertically.',
--- a/client/js/editor/sidebar/properties.js
+++ b/client/js/editor/sidebar/properties.js
@@ -401,8 +401,8 @@ const editorPropertyHints = {
   dropOffsetY: 'Vertical starting position for widgets aligned inside the holder.',
   stackOffsetX: 'Horizontal distance added between consecutively stacked widgets.',
   stackOffsetY: 'Vertical distance added between consecutively stacked widgets.',
-  multiSelectMax: 'Let players pick widgets in this holder by clicking them, up to this many each. The picked ones get the clicking player\'s name in their selectedBy property, which automations can read.',
-  multiSelectStyle: 'How a widget the player picked is shown. The built-in styles win over a filter or shadow the widget sets itself; "custom CSS" only sets the multiSelected class so the holder css can style it.',
+  multiSelectMax: 'Let players pick widgets in this holder by clicking them, up to this many each - or "any number" for no limit. The picked ones get the clicking player\'s name in their selectedBy property, which automations can read. Works best when the holder spreads its widgets out (see Stack offset X/Y): stacked on one spot only the top one can be clicked.',
+  multiSelectStyle: 'How a widget the player picked is shown. The built-in styles win over a filter or shadow the widget sets itself, so a widget cannot hide that it was picked (which also means it loses its own shadow while picked); "own CSS" only sets the multiSelected class so the holder css can style it.',
   showPlayerColors: 'Use each player\'s color in their scoreboard heading.',
   verticalHeader: 'Rotate the scoreboard header text vertically.',
   autosizeColumns: 'Size score columns from their contents instead of using fixed widths.',
@@ -5972,6 +5972,7 @@ class PropertiesModule extends SidebarModule {
         { value: 'all', text: 'any number' },
         { value: 'custom', text: 'up to...' }
       ],
+      customText: value => +value > 0 ? `up to ${Math.floor(+value)}` : `custom: ${JSON.stringify(value)}`,
       setValue: value => {
         if(value === 'custom') {
           const entered = parseInt(prompt('How many widgets may one player select in this holder?', widget.get('multiSelectMax')));
@@ -5989,15 +5990,19 @@ class PropertiesModule extends SidebarModule {
     const styleInput = new SelectInput(this, widget, 'Selection shown as', {
       property: 'multiSelectStyle',
       hint: editorPropertyHints.multiSelectStyle,
+      // the raw values are in the labels because the JSON editor and the wiki know the
+      // property by those, not by the words this dropdown uses
       choices: [
-        { value: 'elevate',   text: 'raised' },
-        { value: 'highlight', text: 'outlined' },
-        { value: 'shade',     text: 'darkened' },
-        { value: 'none',      text: 'custom CSS' }
+        { value: 'elevate',   text: 'raised (elevate)' },
+        { value: 'highlight', text: 'outlined (highlight)' },
+        { value: 'shade',     text: 'others dimmed (shade)' },
+        { value: 'none',      text: 'own CSS (none)' }
       ]
     });
     styleInput.render(row);
-    this.addPropertyListener(widget, 'multiSelectMax', w => styleInput.dom.style.display = w.get('multiSelectMax') ? '' : 'none');
+    // multiSelectLimit(), not the raw value: a routine-computed "0" is a string and
+    // turns click-to-select off, so the style row has to disappear along with it
+    this.addPropertyListener(widget, 'multiSelectMax', w => styleInput.dom.style.display = w.multiSelectLimit() ? '' : 'none');
   }
 
   // Text / background / border color plus a brightness filter written into a
--- a/client/js/editor/sidebar/properties.js
+++ b/client/js/editor/sidebar/properties.js
@@ -401,6 +401,8 @@ const editorPropertyHints = {
   dropOffsetY: 'Vertical starting position for widgets aligned inside the holder.',
   stackOffsetX: 'Horizontal distance added between consecutively stacked widgets.',
   stackOffsetY: 'Vertical distance added between consecutively stacked widgets.',
+  multiSelectMax: 'Let players pick widgets in this holder by clicking them, up to this many each. The picked ones get the clicking player\'s name in their selectedBy property, which automations can read.',
+  multiSelectStyle: 'How a widget the player picked is shown. "custom CSS" only sets the multiSelected class so the holder css can style it.',
   showPlayerColors: 'Use each player\'s color in their scoreboard heading.',
   verticalHeader: 'Rotate the scoreboard header text vertically.',
   autosizeColumns: 'Size score columns from their contents instead of using fixed widths.',
@@ -5942,6 +5944,7 @@ class PropertiesModule extends SidebarModule {
       { label: 'X', property: 'stackOffsetX' },
       { label: 'Y', property: 'stackOffsetY' }
     ]);
+    this.renderMultiSelectRow(widget);
 
     this.renderAdvancedSection(widget, body => {
       this.renderSeatReferenceInput(widget, 'showInactiveFaceToSeat', 'Show inactive face to seat:', body, {
@@ -5952,7 +5955,49 @@ class PropertiesModule extends SidebarModule {
     });
 
     // onEnter / onLeave stay in the generic property list (handled in PR #3034)
-    this.renderOtherPropertiesSection(widget, [ 'dropTarget', 'text', 'icon', 'image', 'dropOffsetX', 'dropOffsetY', 'stackOffsetX', 'stackOffsetY', 'showInactiveFaceToSeat' ]);
+    this.renderOtherPropertiesSection(widget, [ 'dropTarget', 'text', 'icon', 'image', 'dropOffsetX', 'dropOffsetY', 'stackOffsetX', 'stackOffsetY', 'showInactiveFaceToSeat', 'multiSelectMax', 'multiSelectStyle' ]);
+  }
+
+  // Click-to-select: how many widgets a player can pick in this holder by clicking
+  // them, and what a picked widget looks like. The style only matters once picking
+  // is enabled, so it stays hidden while the limit is 0.
+  renderMultiSelectRow(widget) {
+    const row = div(this.moduleDOM, 'propertyInlineRow');
+    const maxInput = new SelectInput(this, widget, 'Click to select', {
+      property: 'multiSelectMax',
+      hint: editorPropertyHints.multiSelectMax,
+      choices: [
+        { value: 0, text: 'off' },
+        { value: 1, text: 'one at a time' },
+        { value: 'all', text: 'any number' },
+        { value: 'custom', text: 'up to...' }
+      ],
+      setValue: value => {
+        if(value === 'custom') {
+          const entered = parseInt(prompt('How many widgets may one player select in this holder?', widget.get('multiSelectMax')));
+          if(Number.isFinite(entered) && entered > 0)
+            this.inputValueUpdated(widget, 'multiSelectMax', entered);
+          else
+            maxInput.update(widget.get('multiSelectMax')); // snap the dropdown back
+        } else {
+          this.inputValueUpdated(widget, 'multiSelectMax', value);
+        }
+      }
+    });
+    maxInput.render(row);
+
+    const styleInput = new SelectInput(this, widget, 'Selection shown as', {
+      property: 'multiSelectStyle',
+      hint: editorPropertyHints.multiSelectStyle,
+      choices: [
+        { value: 'elevate',   text: 'raised' },
+        { value: 'highlight', text: 'outlined' },
+        { value: 'shade',     text: 'darkened' },
+        { value: 'none',      text: 'custom CSS' }
+      ]
+    });
+    styleInput.render(row);
+    this.addPropertyListener(widget, 'multiSelectMax', w => styleInput.dom.style.display = w.get('multiSelectMax') ? '' : 'none');
   }
 
   // Text / background / border color plus a brightness filter written into a
--- a/client/js/editor/sidebar/properties.js
+++ b/client/js/editor/sidebar/properties.js
@@ -402,7 +402,7 @@ const editorPropertyHints = {
   stackOffsetX: 'Horizontal distance added between consecutively stacked widgets.',
   stackOffsetY: 'Vertical distance added between consecutively stacked widgets.',
   multiSelectMax: 'Let players pick widgets in this holder by clicking them, up to this many each. The picked ones get the clicking player\'s name in their selectedBy property, which automations can read.',
-  multiSelectStyle: 'How a widget the player picked is shown. "custom CSS" only sets the multiSelected class so the holder css can style it.',
+  multiSelectStyle: 'How a widget the player picked is shown. The built-in styles win over a filter or shadow the widget sets itself; "custom CSS" only sets the multiSelected class so the holder css can style it.',
   showPlayerColors: 'Use each player\'s color in their scoreboard heading.',
   verticalHeader: 'Rotate the scoreboard header text vertically.',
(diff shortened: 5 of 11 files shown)


VTT-AI-Agent (commented):
## TL;DR

Nice, well-scoped feature — the opt-in defaults really do keep existing games inert (179 jest tests pass locally, no state hashes move) and the editor/validator/jsonedit upkeep is complete. The one thing I'd fix before merge: **`moveMultiSelectionAlong()` force-moves the followers with `moveToHolder()`, bypassing `compareDropTarget()` and `dropLimit`** — I reproduced widgets landing in a holder that rejects them and blowing past its drop limit. Two smaller confirmed issues below (drag-to-table piles up with default `stackOffset`, and `selectedBy` survives a routine `MOVE` out of a pile).

## Findings

### 1. The drag-along bypasses drop-target validation and `dropLimit` (`client/js/widgets/widget.js:2853-2860`)

A normal drag can only end in a holder that passed `getValidDropTargets()` (`client/js/main.js:39`), which enforces both `compareDropTarget(widget, t)` and `dropLimit`. The followers skip all of that — they are handed to `w.moveToHolder(target)` purely because the *dragged* widget was accepted.

Reproduced with TestCafe on this branch (hand holder `multiSelectMax: 'all'`, three cards, `card3` has `cardType: 'other'`; target holder `{ dropLimit: 1, dropTarget: { cardType: 'plain' } }`), selecting all three and dragging `card1` onto the target:

```
AFTER DROP IN LIMITED HOLDER:
  card1:card:parent=table   card2:card:parent=table   card3:card:parent=table
```

`card3` is in a holder whose `dropTarget` explicitly excludes it, and a `dropLimit: 1` holder now holds three widgets. Both are invariants games rely on (deal-one-card slots, typed play areas). Suggest filtering `others` through the same check the drag uses, e.g. keep only `w` where `target.get('dropLimit')` still has room and `compareDropTarget(w, target)` holds, and leave the rejected ones behind in the source holder. Same spot: only `movable` is checked, not `fixedParent`, so a `fixedParent` follower gets reparented too.

### 2. Followers collapse into a pile on the table with the default `stackOffsetX/Y` (`client/js/widgets/widget.js:2848-2849`)

`Holder`'s defaults are `stackOffsetX: 0, stackOffsetY: 0` (`client/js/widgets/holder.js:31-32`), so for the great majority of holders `offsetX`/`offsetY` are `0` and every follower is positioned at exactly `this.get('x') / this.get('y')`. `moveEnd()` then calls `updatePiles()` and they merge. Reproduced with a plain default holder:

```
AFTER DROP ON TABLE:
  card1:card:parent=jyo2   card2:card:parent=jyo2   jyo2:pile:parent=null
  card3:card:parent=undefined:x=377:y=21
```

That directly contradicts the PR text ("they keep the spacing the holder gave them so that everybody can see what was played") — the whole point of the fallback. `tests/testcafe/multiselect.js:12` sets `stackOffsetX: 120` explicitly, so the test never exercises the default. Consider falling back to the widget's own width/height (or a fixed nudge) when the source holder's offsets are 0.

### 3. `selectedBy` is not cleared when the widget leaves via a pile (`client/js/widgets/widget.js:432-435`)

`multiSelectHolder()` deliberately resolves through `_ancestor` so cards inside a pile inside the holder are selectable, but `checkParent()` clears the selection only when `this.currentParent.multiSelectLimit()` is truthy — and for such a card `currentParent` is the *pile*, whose limit is 0. The drag path happens to be safe (`moveStart()` reassigns `currentParent` to `_ancestor`), but the routine path is not. Reproduced with a `pile` inside the hand holder and exactly the recipe from the PR description:

```
SELECT type:card property:selectedBy relation:contains value:${playerName}  →  MOVE to table
AFTER ROUTINE MOVE: card3:card:parent=table:sel=["TestCafe"]
```

The played card keeps `selectedBy`, so the next run of the same routine picks up cards that are already on the table. Using `widgets.get(this.get('_ancestor'))` (or `currentParentWidget()` resolved through `_ancestor`) for the guard would make the two sides agree.

### 4. Smaller notes

- `client/js/widgets/widget.js:562` — `multiSelectedWidgets()` returns widgets in `widgets` map order, so the `i` used for the spread offsets and `z` in `moveMultiSelectionAlong()` is arbitrary rather than the order the cards appear in the holder. Sorting by `z` (or by x/y) would make the fan-out predictable.
- `client/css/widgets/holder.css:94-105` — `elevate`/`shade` are `filter:` rules in a stylesheet, so any widget that sets `filter` through its `css` property (a common way to dim/tint cards) silently wins and the selection becomes invisible. `highlight`'s `box-shadow` has the same exposure. Worth at least a note in the property hint.
- `client/js/widgets/widget.js:1999-2000` — `contains` on a non-array property degrades to `==` via `asArray()` (`asArray('abc')` → `['abc']`), so `property: 'text', relation: 'contains', value: 'ab'` is false. The name reads like a substring test; the mirror-of-`in` meaning is worth spelling out in the wiki update you already planned.
- `client/js/widgets/widget.js:554-559` — `multiSelectLimit()` requires `typeof max == 'number'`, so a routine that does `SET multiSelectMax "2"` silently turns the feature off rather than warning. Minor, and the validator catches it in the game file.
- `tests/client/widget-select.test.js:7` — the "hand" is `type: 'widget'`, not `holder`; nothing covers the `_ancestor`/pile resolution or the drag-along in jest, which is where issues 1–3 live.

Verified locally: `npm test` → 179 passed, and the three reproductions above ran against this branch under `chromium:headless`.

I did not add the `beta` label: the diff genuinely changes the engine — it hooks the new interaction into `Widget.click()`, `checkParent()`, `moveStart()`/`moveEnd()` and adds a `SELECT` relation — so it is excluded from beta auto-deployment by design, and the confirmed `dropTarget`/`dropLimit` bypass in finding 1 would want fixing first in any case.

VTT-AI-Agent:
Thanks — all three confirmed issues were real, fixed in 9ed05aa9, plus three of the smaller notes.

**1. Drag-along bypassed `compareDropTarget()`/`dropLimit`** — fixed. `moveMultiSelectionAlong()` now runs each follower through the same two checks a drag would have run for it: `compareDropTarget(w, target)` and the remaining room in `dropLimit` (counted down as followers are placed, with the dragged widget already occupying its slot). A follower that fails either check stays in the source holder instead of being forced in. `fixedParent` is now filtered out alongside `movable` in `moveStart()`, matching the guard the dragged widget itself gets at `widget.js:2593`.

**2. Drop-on-table collapsed into a pile with the default `stackOffsetX/Y: 0`** — fixed. When the source holder has no spacing of its own, the offset falls back to the width of the dragged widget, so the followers land side by side (card width 103 ≫ `defaultPileSnapRange` 10, so `updatePiles()` leaves them alone). The PR text and the code comment are corrected to describe what it actually does.

**3. `selectedBy` surviving a routine `MOVE` out of a pile** — fixed. `checkParent()` now resolves the governing holder through `_ancestor` (`this.currentParent.multiSelectLimit() || this.multiSelectHolder()`), so the drag path and the routine path agree, exactly as you suggested.

**Smaller notes:**
- *Arbitrary follower order* — fixed: the followers are sorted by `z` in `moveStart()`, so the fan-out follows the order the widgets sit in the holder.
- *`css` filter beating the selection styles* — fixed: the three built-in styles are now `!important` (and `elevate` uses `box-shadow` instead of a `filter: drop-shadow`, so it no longer competes with a widget's own filter at all). A widget can no longer hide the fact that the player selected it; games that want full control still set `multiSelectStyle: 'none'`. The editor hint says so now.
- *`SET multiSelectMax "2"` silently turning the feature off* — fixed: `multiSelectLimit()` coerces with `+max`, so a routine-computed string works.
- *`contains` on a non-array* — kept as the mirror of `in` (a substring meaning would misfire on `selectedBy: 'Alice'` vs `value: 'Ali'`), but the code comment now spells out that it is not a substring test. I'll make sure the wiki page says the same.
- *Test coverage* — `tests/client/widget-select.test.js` grew five scenarios that all fail without this commit: the pile/`_ancestor` resolution, the routine `MOVE` out of a pile, the `dropTarget` rejection, the `dropLimit` exhaustion, and the surface spread with the default `stackOffset`. The "hand" fixture is `type: 'holder'` now. `tests/testcafe/multiselect.js` gained two browser scenarios for the same drop-target rejection and the default-`stackOffset` surface drop (both verified failing on the old code and passing on the new one).

`npm test` → 185 passed. `npx testcafe chromium:headless tests/testcafe/multiselect.js` → 5 passed.

Agreed on `beta`: the diff touches the engine, so it is excluded from beta auto-deployment by design.

VTT-AI-Agent:
Feedback from Discord ([message](https://discord.com/channels/770758631146782780/1534030355961745419/1534050045240807677)):

> The PR comment says "that is what this implements, plus the two extras that came up in the thread: dragging one selected card takes the whole selection along," But dragging one card of a selection only drags that one card.

Fair, and the description promised more than the code did. The selection *did* end up wherever the card was dropped, but only at the drop: while the mouse was down a single card moved, and the rest jumped over at the end — into a face-down pile in the holder I dropped on, which looks exactly like nothing came along.

Fixed in 376c3082: the moment the drag carries the card out of the holder, the rest of the pick comes out with it and travels next to the cursor, spread the way the holder spread it (or a card width apart when the holder stacks its children on one spot). Dragging three picked cards out of the hand on your `pr-test` room now looks like this:

dragging the selection

The drop checks are unchanged — a card the target's `dropTarget` rejects or that no longer fits its `dropLimit` still does not get forced in. What changed is where it ends up: since it is out of the holder now, it goes back into the holder it was picked from and is picked again there, so a refused drop costs no selection. Dragging out of the holder and dropping back into it does the same for the whole selection, which makes it a way to change your mind mid-drag.

Covered by three new jest cases (carried out of the holder, refused drop returns still selected, drag ends where it started) and an extra assertion in `tests/testcafe/multiselect.js`; `npm test` (188) and the multiselect, routines and publiclibrary-1 testcafe suites pass locally.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review) of the click-to-select feature. The editor row is clean and the three selection styles work, but the **selection is signalled with colours the app already uses for something else** (gold ≈ edit-mode selection, "darkened" ≈ disabled) and the **drag-along spread always fans to the right, which parks the followers outside the room surface when you drop near the right edge**. Biggest single win: make the built-in `highlight` style use `var(--VTTblue)` (the app's existing "you picked this" colour, as in the `INPUT`/`inputchoose` dialog) and clamp/mirror the drag spread so it stays on the surface.

Reviewed by checking out the branch, running the server locally and driving it with Playwright (screenshots below are from this branch), plus reading `client/css/widgets/holder.css` and `client/js/editor/sidebar/properties.js`.

| elevate (default) | highlight | shade |
| --- | --- | --- |
| elevate | highlight | shade |

---

## Big layout / visual-design improvements

**1. The `highlight` colour collides with the editor's own selection ring.**
`client/css/widgets/holder.css:105-107` uses gold `#f5c518` with a glow. Edit mode already marks the selected widget with `box-shadow: 0 0 0 5px yellow` (`client/css/editor/layout.css:110`), and the app's established "the player picked this widget" affordance — the `INPUT`/`inputchoose` dialog — uses a 3px `var(--VTTblue)` border plus a check badge (`client/css/layout.css:1833-1854`). A designer testing their game in edit mode now sees two near-identical yellow rings meaning two different things, and a player sees a colour that means "chosen" nowhere else in the app. Suggestion: `box-shadow: 0 0 0 3px var(--VTTblue), 0 0 8px 4px rgba(31,92,166,.6) !important;` — it fits the existing palette and reuses an established meaning. (If you want the extra reinforcement, the `::after` check badge from `.inputchoose .selected.inputchooseWidgetWrapper` is already written and would read instantly.)

**2. The drag-along spread can push the selection off the surface.**
`multiSelectSpread()` / `dragMultiSelectionAlong()` (`client/js/widgets/widget.js:2846-2890`) always lay the followers out at `x + spread.x * i`, i.e. always to the right, with no bound against the room's 1600×1000 surface. Dragging a 4-card selection from a hand on the right of the table towards the right edge leaves the followers rendered off-screen mid-drag — the whole point of the feature ("the drag shows what is about to be played") is lost — and after the drop they are stranded at x = 1586 / 1706 / 1826, outside the surface:

drag near the right edge

Suggestion: mirror the fan to the left when `x + spread.x * (n-1) + width` exceeds the surface width (same for y), or clamp each follower into the surface. A tighter fan (e.g. `spread.x = min(stackOffsetX || width, (surfaceWidth - x - width) / (n-1))`, card-overlap style) would also keep large selections readable instead of a metre-wide row.

**3. `shade` reads as "disabled", not "selected".**
`filter: brightness(0.65)` on the *picked* widget is the near-universal convention for unavailable/disabled. A first-time player is at least as likely to read the darkened cards as the ones they *can't* play. The same effect applied the other way round — `.multiSelect-shade .widget:not(.multiSelected) { filter: brightness(0.65) }` — says "these are your picks" with the exact same CSS, and stays inside the app's visual language. If the current direction is deliberate, at minimum rename the editor label from "darkened" to something that states the target ("darken the picked ones").

**4. Nothing tells a player the holder is click-to-select, and nothing happens when the limit is full.**
There is no hover affordance (widgets in a `multiSelectMax` holder look and behave exactly like any other widget until clicked), and once the limit is reached `toggleMultiSelect()` consumes further clicks silently (`client/js/widgets/widget.js:2837-2841`) — a newcomer's reading of that is "the game is broken". Two cheap fixes that fit the existing style: (a) a faint preview of the selection style on hover inside a `multiSelect-*` holder (e.g. `translate: 0 -5px` for elevate, a 30 %-opacity ring for highlight), which makes the affordance discoverable; (b) some feedback at the limit — a short shake/flash of the already-picked widgets, so the player learns *why* the click did nothing.

- **`custom: 2` shown in the editor dropdown.** Any numeric limit other than 1 falls through to `SelectInput`'s raw fallback label (`client/js/editor/propertyInputs.js:621`), so the row reads `Click to select [custom: 2]` next to friendly siblings "off / one at a time / any number". It looks like a leaked internal value. Worth passing a formatter (`up to 2`) rather than the JSON dump:
  editor row
- **Fixed 14px raise doesn't scale with the widget** (`holder.css:100`). On a full-size card it is a subtle nudge; on a 40px token or a marker it is a third of the widget. `translate: 0 -10%` resolves against the element's own height and behaves sensibly at every widget size.
- **With everything selected, `elevate` is invisible** — all four cards move up by the same 14px, so a fully-picked hand looks identical to an untouched one apart from a shadow ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1534063619573022801/multiselect-4selected.png)). Adding a ring or a slight `scale(1.03)` to the elevate style would keep it readable in the `multiSelectMax: "all"` case.
- **`multiSelectMax > 1` is unusable with the holder's default `stackOffset` of 0**: the children sit on one spot, so only the topmost widget is clickable at all (Playwright reports `#w_card4 … intercepts pointer events` for every other card), and the raise only reveals a sliver: stacked holder — the "Click to select" row sits directly under "Stack offset X/Y", so a one-line note in the hint ("works best when the holder spreads its widgets — see Stack offset") would save designers the debugging round.
- **Visibility toggle is truthiness-based** (`properties.js:5999`): `multiSelectMax: "0"` (a string, e.g. produced by a routine) is truthy, so "Selection shown as" shows even though selection is off — while `multiSelectLimit()` correctly treats it as off. Use the same `multiSelectLimit`-style check so the editor and the engine agree. The off state itself is otherwise nicely clean: off
- **`!important` on the elevate style also overwrites the widget's own `box-shadow`**, so a card with a designed shadow loses it while selected (elevate) or entirely (highlight). Consider layering (`box-shadow: var(--widgetShadow, 0 0 0 transparent), 0 4px 6px …`) or at least documenting it in the hint next to the "the built-in styles win over a filter or shadow" sentence, which currently only promises the *filter* case.

## Readability & understandability (first-time user / first-time designer)

- **"Selection shown as" vocabulary differs between the two editors.** The sidebar offers *raised / outlined / darkened / custom CSS*, the JSON editor inserts *elevate / highlight / shade / none* (`jsonedit.js:1503`). Someone switching between the two has to map four words twice. Cheapest fix: mention the raw value in the hint (`… "custom CSS" (value "none") only sets the multiSelected class …`) — the hint currently names "custom CSS", which does not exist as a value.
- **`multiSelectMax`'s hint doesn't mention "any number"/`all`**: "up to this many each" doesn't cover the `all` choice sitting right there in the dropdown. A trailing "— or 'any number' for no limit" closes it.
- **Other players' picks are invisible in a shared holder.** `selectedBy` is per-player and only the local player's picks are styled, so in a common holder (a market row, a draft display) two players can silently pick the same card and only find out at the drop. Since the app already has a player-colour convention (`playerColor`, `--wcMain`, the drag shadow's `dropShadowOwner`), a faint marker in the other player's colour for "someone else has this picked" would fit and would remove a real multiplayer surprise — worth at least a follow-up issue if out of scope here.
- **"Click to select" as the label for the *limit*** is slightly off: the control's values are counts, not an on/off. "Click to select: up to" or a label like "Selectable by clicking" reads closer to what the dropdown does — minor, and the hint carries the meaning either way.
- The `prompt()` for "up to…" matches the existing precedent in the same file (`precision`, `properties.js:5622`), so no objection there — only the `custom: N` label it produces (above).

VTT-AI-Agent:
Thanks — the colour collisions and the unbounded fan were both real. Everything except the last point is fixed in 03cc2600 (`npm test` → 190 passed, `tests/testcafe/multiselect.js` → 5 passed, `routines.js` → 4 passed; screenshots below are from a local server on this commit).

## Big improvements

**1. `highlight` colliding with edit mode's yellow ring** — fixed. Both ring styles now use `var(--VTTblue)`, the colour the `INPUT`/`inputchoose` dialog already uses for "the player chose this":

| highlight | shade | elevate, everything picked |
| --- | --- | --- |
| highlight | shade | all selected |

I did not add the `::after` check badge: it would sit on top of a card's artwork, and on a 40px token it would cover most of the widget. The ring reads well enough at every size.

**2. The drag-along spread pushing the selection off the surface** — fixed. `multiSelectSpread()` now runs each axis through `fitMultiSelectSpread()`, which keeps the last follower inside the 1600×1000 surface: it turns the fan around towards the side with more room, and tightens it up (`min(offset, room/count)`) when neither side fits. Dragging a 3-card selection into the right edge now fans left instead of off-screen, and the cards land at x = 1494 / 1374 / 1254 instead of past 1600:

fan mirrored at the right edge

Two jest cases cover it (the mirror at the right edge, and a holder with `stackOffsetX: 800` whose fan has to tighten to 700 to fit).

**3. `shade` reading as "disabled"** — fixed the way you suggested, by turning it around: `.multiSelect-shade:has(.multiSelected) .widget:not(.multiSelected)` dims what was *not* picked, so the picks stand out by everything else stepping back. Nothing is dimmed while the player has not picked anything (that is what the `:has()` is for), and a widget that *contains* a pick — a pile — is excluded, otherwise the filter would dim the pick inside it. The editor label says "others dimmed (shade)" now.

**4. Discoverability and the silent no-op at the limit** — both added.
- Hover: a widget in a click-to-select holder answers the pointer with a faint `--VTTblue` ring (a hint of the ring it gets when picked), so the holder shows that it reacts to clicks before the first click: hover
- At the limit: `toggleMultiSelect()` calls `flashMultiSelectFull()`, which flashes an inset ring on the holder for 400 ms. It is a CSS animation rather than a plain rule, so a holder that styles itself through `css` cannot swallow it, and the class is re-applied on every refused click so repeated clicking keeps flashing: limit flash

## Nitpicks

- **`custom: 2` in the dropdown** — fixed. `SelectInput` takes an optional `customText` formatter now, and the row reads `Click to select [up to 2]`: editor row
- **Fixed 14px raise** — fixed: `translate: 0 -10%`, so it scales with the widget.
- **`elevate` invisible when everything is picked** — fixed by the ring that `elevate` now carries as well (third screenshot above: four raised cards, all ringed, versus none).
- **`multiSelectMax > 1` with the holder's default `stackOffset: 0`** — I left the stacking alone (that a stacked holder only exposes its top widget to clicks is holder behaviour, not something this feature should change) and put the caveat in the hint instead: "Works best when the holder spreads its widgets out (see Stack offset X/Y): stacked on one spot only the top one can be clicked."
- **Truthiness-based visibility toggle** — fixed: the listener uses `w.multiSelectLimit()`, so a string `"0"` hides the style row exactly as it turns the feature off.
- **`!important` overwriting the widget's own `box-shadow`** — kept (a widget must not be able to hide that it was picked, and layering through a `var()` would only work for widgets that opt into it), but the hint now says so: "…so a widget cannot hide that it was picked (which also means it loses its own shadow while picked)".
- **Sidebar / JSON-editor vocabulary mismatch** — fixed by naming the raw value in the label: *raised (elevate) / outlined (highlight) / others dimmed (shade) / own CSS (none)*. That also removes the "custom CSS" label for a value that is called `none`.
- **`multiSelectMax` hint not mentioning `all`** — fixed: "…up to this many each - or 'any number' for no limit."
- **"Click to select" as the label for a count** — kept: the dropdown reads "Click to select: off / one at a time / any number / up to 2", which reads as a sentence, and the hint carries the meaning.

## Not done

- **Other players' picks are invisible in a shared holder.** Agreed that this is a real multiplayer surprise, but showing it means styling `selectedBy` entries that are *not* the local player in that player's colour, which needs a per-player colour on the widget's DOM and a decision about what happens when three players pick the same card — bigger than this PR and a design question of its own. Better as a follow-up issue on top of `selectedBy` than squeezed in here.

The PR description is updated to describe the styles as they now behave.

VTT-AI-Agent (commented):
## TL;DR

The latest commits fix the earlier drop-target, drop-limit, pile, and nested-parent defects, but two engine edge paths still leave the group/selection in the wrong state. Most importantly, dropping onto a line immediately sends every follower back to the source holder, and the new TestCafe coverage is not actually included in either CI matrix.

## Findings

### 1. Dropping the selection onto a line returns all followers to the source (`client/js/widgets/widget.js:2923-2949`)

The line branch first calls `dragMultiSelectionAlong()` so the followers leave the holder and remain spread next to the dragged widget, matching the comment at lines 2923-2926. However, `target` is still the line, `holder` is `null`, and every follower now has `stillInHolder === false`; the `else if(target && !stillInHolder)` branch therefore moves each one back to `source` and selects it again. The result is that dropping a selected widget onto a line moves only the widget under the pointer, even though the whole selection visibly followed it during the drag. This branch should distinguish a rejected holder drop from the intentional line/surface case (for example, only restore followers when there is a non-line target that rejected them), and it needs a line-drop regression test.

### 2. Turning selection off leaves `selectedBy` state that survives later moves (`client/js/widgets/widget.js:431-436`, `client/js/widgets/widget.js:571-573`)

`checkParent()` clears `selectedBy` only while either `currentParent.multiSelectLimit()` or `multiSelectHolder()` is truthy. If a routine/editor changes a holder from `multiSelectMax: 1`/`"all"` to `0` while a card is selected, both checks become false. The stale name remains on the card, and a later `MOVE` out of that holder also skips the clear, so `SELECT ... relation: "contains"` can continue collecting a card that is now on the table. Disabling the feature should clear the holder's existing selections, or leaving a holder should clear an existing `selectedBy` independently of the holder's current limit; please cover the enable → select → disable → move sequence.

### 3. The new end-to-end suite is not run in CI (`tests/testcafe/multiselect.js:1`, `.github/workflows/testcafe-chrome.yml:19-31`, `.github/workflows/testcafe-firefox.yml:19-31`)

Both workflows enumerate test files explicitly, and neither matrix includes `multiselect.js`. That is why the successful PR checks contain the existing editor/toolbar/compute/public-library/routines jobs but no multiselect job. Please add this file to both matrices (or place the cases in an already-enumerated suite), otherwise the browser coverage for the core feature can regress without affecting CI.

### 4. The validator rejects a limit shape the engine deliberately supports (`validator/validate_gamefile.js:134`)

`multiSelectLimit()` intentionally accepts numeric strings produced by routines, and `tests/client/widget-select.test.js:114-129` establishes `"2"` as supported behavior. The validator nevertheless accepts only integer numbers or `"all"`, so an exported/current state containing that supported value is reported invalid. The validator should accept the same non-negative whole-number strings as the runtime, or the runtime/test/PR contract should stop advertising them.

I did not add the `beta` label because the changes to `client/js/widgets/widget.js` and `client/js/widgets/holder.js` genuinely alter engine click, selection, parent, drag/drop, and routine semantics, making this PR engine-ineligible for beta auto-deployment.

VTT-AI-Agent:
Thanks — all four findings were real and are fixed in 2665cbf5 (`npm test` → 192 passed; `multiselect.js` → 5 passed, `routines.js` → 4 passed, `publiclibrary-1.js` → 5 passed locally under `chromium:headless`).

**1. Dropping the selection onto a line sent every follower back to the source** — fixed. The line case and the surface case are now one condition: `onSurface = !target || target.get('type') == 'line'`. It is what decides whether `dragMultiSelectionAlong()` lays the followers out next to the widget, and the restore branch is keyed off the same flag (`else if(!onSurface && !stillInHolder)`) instead of off `target`, so a line no longer counts as a holder that refused them. The `target == source` case still restores, because a drag that ends back in the holder it started in is not `onSurface`. New jest case *"When the drag ends on a line, where only the dragged widget becomes a stop"*: the dragged card becomes a child of the line, the two followers stay on the surface at the fan positions, and none of them is back in the hand. It fails on the previous commit.

**2. Stale `selectedBy` after the feature is turned off** — fixed the second way you suggested: `checkParent()` clears `selectedBy` whenever the widget leaves a parent and has one, without asking what the holder's current limit is. A selection only means something while the widget sits in the holder that offered it, so there is nothing left for a later `MOVE` to carry out. New jest case covers enable → select → disable → move; it also fails on the previous commit. What is deliberately *not* cleared is a selection on a widget that stays put while `multiSelectMax` goes to 0 — the holder loses its `multiSelect-*` class so nothing is shown or clickable any more, and turning the limit back on restores what the player had picked rather than silently dropping it.

**3. `multiselect.js` not in either CI matrix** — fixed, added to `testcafe-chrome.yml` and `testcafe-firefox.yml`. Good catch; the suite was passing locally and invisible to CI.

**4. Validator rejecting `"2"`** — fixed. `multiSelectMax` now accepts the same shapes the runtime does: `"all"`, or a non-negative whole number as a number *or* a string. Confirmed with `validate_gamefile_node.js` on a file holding `"2"`, `3`, `"all"` (all clean) and `"lots"` (still reported).

Agreed on `beta` — the diff changes engine click/parent/drag semantics, so it is engine-ineligible by design.

VTT-AI-Agent (commented):
## TL;DR

The feature is well built and genuinely inert for existing games (192 jest tests pass locally, all behaviour gated behind `multiSelectMax` defaulting to `0`), and the earlier drop-target / drop-limit / line-drop findings are fixed. One confirmed defect left: **a `pile` inside a click-to-select holder becomes selectable itself, so clicking its handle selects the pile instead of opening the pile overlay** — I reproduced it in a browser. Everything else below is smaller.

## Findings

### 1. A pile inside the holder is selectable, and its overlay becomes unreachable (`client/js/widgets/widget.js:573`, `:593`)

`multiSelectHolder()` resolves the governing holder through `_ancestor`. For a *card in a pile* that correctly skips the pile and finds the holder — but a **pile whose parent is the holder** also has `_ancestor == holder`, so the pile itself is treated as a selectable child. `Pile.click()` calls `super.click()` first, `toggleMultiSelect()` returns `true`, and the pile menu ("Flip everything over", shuffle, …) never opens.

Reproduced on this branch against a local server with:

```json
{ "hand": { "id": "hand", "type": "holder", "multiSelectMax": 2, "x": 50, "y": 600, "width": 700, "height": 200 },
  "pile": { "id": "pile", "type": "pile", "parent": "hand", "x": 10, "y": 10 },
  "card1": { "id": "card1", "type": "card", "deck": "deck", "cardType": "plain", "parent": "pile" },
  "card2": { "id": "card2", "type": "card", "deck": "deck", "cardType": "plain", "parent": "pile" } }
```

Clicking the pile handle: `#pileOverlay` stays hidden, and the pile ends up with `selectedBy: ["Guest744"]` and class `multiSelected`. Consequences beyond the lost overlay:

- the pile eats one slot of `multiSelectMax` even though the player thinks they picked nothing;
- `selectedBy` is written onto a transient widget — when the pile drops below two children it is deleted and the pick disappears with no `selectedByChangeRoutine` for the widgets that were in it;
- dragging the "selected" pile makes `moveStart` treat it as the anchor of a selection whose other members are its own siblings.

Since `.multiSelect-shade` already special-cases `:has(.multiSelected)` for piles, the intent clearly is "cards in piles, not the pile" — so `toggleMultiSelect()` should bail out for `this.get('type') == 'pile'` (and a jest case for it would be worth adding next to the existing "selected widget inside a pile" block in `tests/client/widget-select.test.js:186`).

### 2. `setMultiSelected()` is a read-modify-write of shared state (`client/js/widgets/widget.js:582`)

```js
const players = asArray(this.get('selectedBy')).filter(p=>p != playerName);
```

Each client reads the whole array, edits its own entry and writes the whole array back. The PR explicitly advertises "several players can select simultaneously in a shared holder", but VTT deltas are last-write-wins per property: if A and B click the *same* widget within one round trip, B's write is computed from a `selectedBy` that does not yet contain A, so A's pick is silently dropped (and A's client will show it selected until the next delta arrives). This is inherent to modelling per-player state as one shared array — I don't think it needs to block the feature, but it deserves a note in the property hint / wiki rather than the current "each sees only their own picks" framing.

### 3. Nothing ever clears a stale `selectedBy` (`client/js/widgets/widget.js:430`)

The only cleanup is "the widget leaves the holder". A player who selects and then leaves the room, or simply renames themselves in the players overlay, leaves their name on the widget forever. Because `selectedBy` is advertised as routine-readable, a later `SELECT … relation: "contains", value: "${playerName}"` will collect cards picked by a long-gone namesake, and the value is written into the saved game file. `dragging` has the same weakness today, so this is not a new class of problem — but `dragging` is not something games are told to branch on, and `selectedBy` is. Worth at least clearing it for players who are no longer in `activePlayers` when the holder is next interacted with.

### 4. Overlapping flash timers cut each other short (`client/js/widgets/widget.js:621`)

`flashMultiSelectFull()` schedules an unconditional `setTimeout(…, 400)` on every call. Two clicks 300 ms apart make the *first* timer strip the class 100 ms into the *second* animation, so exactly in the "player keeps clicking a full selection" case the restart the code goes out of its way to support is the one that gets truncated. Keep the handle on the widget and `clearTimeout` it before re-arming.

### 5. Smaller notes

- `client/js/widgets/widget.js:2912` — followers get `z = this.get('z') + i`, so every carried card renders *above* the card actually under the cursor. Descending z would read better.
- `client/css/widgets/holder.css` — `.multiSelect-shade:has(.multiSelected) .widget:not(.multiSelected):not(:has(.multiSelected))` matches nested widgets too, so a plain widget inside another plain widget gets `brightness(0.65)` twice (≈0.42). Restricting the rule to direct children of the holder's children container would avoid the compounding; the hover rule has the same reach.
- `multiSelectLimit()`/`classes()` work on any widget type, but the validator and the editor only offer `multiSelectMax`/`multiSelectStyle` on `Holder`. Either is fine, but as written a deck or a plain container widget with `multiSelectMax` set by hand silently works while `validate_gamefile.js` reports it as unknown. Worth deciding one way or the other.
- `contains` (`widget.js:2020`) uses `asArray(w.get(a.property))`, so `value: null` matches every widget that simply lacks the property (`asArray(undefined) == [undefined]`). The `in` branch guards with `Array.isArray(a.value)`; a matching guard here would avoid the surprise.

### What looks right

- The gating is genuinely airtight: `multiSelectLimit()` returns 0 for `undefined`/`NaN`, `isMultiSelected()` is false for the `[]` default, so `moveStart`/`move`/`moveEnd`/`click` all early-out for every existing game — which matches the unchanged TestCafe state hashes.
- `String(this.get('multiSelectStyle')).replace(/[^a-z]/gi, '')` correctly keeps a game-supplied style out of the class list.
- The refused-drop path (`dropTarget` + `dropLimit` + return-to-source-still-selected) and the line-drop path are now handled and covered by tests; `movedByButton` is the right flag for the followers.
- Validator, `jeAddEnumCommands` and the editor row are all present, and both new TestCafe files are wired into the Chrome and Firefox matrices.

---

I added the `beta` label: the engine files it touches (`client/js/widgets/widget.js`, `client/js/widgets/holder.js`) are only touched incidentally — every new code path is gated on the new opt-in `multiSelectMax` property, which defaults to `0`, so no existing game's click, drag, drop or sync behaviour changes (a maintainer will need to allow-list this PR for the beta build, since the builder still skips it by filename), and the remaining defects only affect holders that opt in, so nothing here can corrupt a game or crash the server.

VTT-AI-Agent:
Thanks — the pile defect was real and is fixed in d9d0951b, along with three of the four smaller points. `npm test` → 194 passed; `multiselect.js` → 6 passed and `routines.js` → 4 passed locally under `chromium:headless`.

**1. A pile in the holder was selectable and swallowed the pile menu** — fixed. `multiSelectHolder()` returns null for a `pile`, so a pile keeps its own click and only the widgets in it are pickable. That is one line in the same place that already resolves through `_ancestor`, so a card inside the pile is still governed by the holder around it. Two new cases cover it: a jest one next to the existing "selected widget inside a pile" block (`toggleMultiSelect()` returns false, `selectedBy` stays empty), and a TestCafe one that clicks the pile handle in a `multiSelectMax: 'all'` hand and asserts `#pileOverlay` becomes visible and nothing ends up selected — verified failing on 2665cbf5 and passing now.

**2. `setMultiSelected()` is a read-modify-write of a shared array** — agreed, and agreed that it should not block the feature: per-player state in one shared property is what `selectedBy` is, and making it conflict-free would mean a per-player property (or a server-side merge) rather than a change inside this method. Documented instead of papered over: the `multiSelectMax` hint now ends with "All players share the one selectedBy list, so in a holder several players pick from at the same time, two clicks on the same widget in the same moment can cost one of them their pick." The wiki page will say the same.

**3. Nothing ever cleared a stale `selectedBy`** — fixed. `Widget.clearStaleMultiSelections()` drops every name that is not the local player and not in `activePlayers`, and `toggleMultiSelect()` runs it on the governing holder before it does anything else — so a player who left or renamed themselves stops eating a `multiSelectMax` slot and stops being collected by `SELECT … relation: "contains"`, exactly at the moment you suggested (the next time somebody uses the holder). It no-ops while the player list is still empty, so nothing is dropped during load. New jest case: a hand holding a pick by "Ghost" and one by "Alice" keeps only Alice's after the sweep.

**4. Overlapping flash timers cut each other short** — fixed: the handle is kept on the widget and `clearTimeout`ed before the next one is armed, so the restart the code goes out of its way to support now actually lasts its 400 ms.

**Smaller notes:**
- *Followers rendering above the card under the cursor* — fixed: descending `z` (`this.get('z') - i`), asserted in the existing "carried out of the holder" jest case.
- *`shade` compounding on nested widgets* — fixed. Both the shade and the hover rules are direct-child selectors now; a `filter` already applies to everything inside the widget it dims, so one level is all that is needed, and a widget that holds a pick dims its *other* children one level down instead. Measured in a browser: `card2` and a plain container both `brightness(0.65)`, the widget inside that container `none` (it inherits the container's filter) — it used to be `brightness(0.65)` on top of its parent's. The hover rule also excludes `.pile` now, since a pile no longer reacts to the click that ring promises.
- *`multiSelectMax` working on any widget but only valid on `Holder`* — decided in favour of the engine: `multiSelectMax`/`multiSelectStyle` moved into the validator's `COMMON_PROPERTIES`, and the `jeAddEnumCommands` contexts widened from `^holder ↦ …` to `^.* ↦ …`, so a deck or a plain container that offers click-to-select is no longer flagged as an unknown property and still gets the editor buttons. The sidebar row stays in the holder section, which is where it belongs. Confirmed with `validate_gamefile_node.js` on a file with `multiSelectMax` on a non-holder widget (clean) and `"lots"` on a holder (still reported).
- *`contains` matching every widget that lacks the property when `value` is `null`* — this one does not reproduce: `asArray(undefined)` is `[undefined]` and `Array.prototype.indexOf` compares strictly, so `[undefined].indexOf(null)` is `-1`. A widget without the property is not matched, and a widget whose property genuinely *is* `null` is matched — the same answer the default `==` relation gives. I left it as is.

I have not touched the `beta` label either way — that call is yours.

PR #3095, opened 2026-08-04
Read and write values inside complex widget properties from routines

### LINKED ISSUE #1605: Make Complex Objects Easier to Use

VTT-AI-Agent:

- [ ] **Decide whether two behavior changes need a legacy gate** — The `GET` falsy-nested-value change and the now per-widget `+`→`=` relation fallback can be gated behind a legacy mode or just documented — either is ready on request. ([discussion](https://github.com/ArnoldSmith86/virtualtabletop/pull/3095#issuecomment-5179362048))
- [ ] **Decide whether the same desync fix should also cover a whole-property `SET`** — A mutating relation (`sort`, `reverse`, `pop`, `shift`) on a key path no longer loses its delta (see *Effect on existing games* 3). The identical pre-existing bug for a *whole-property* `SET … relation "sort"` is untouched here; it is a two-line change if you want it in this PR rather than its own.

Closes #1605.

Complex properties (`css`, `cardTypes`, `cardDefaults`, `faces`, `dropTarget`, custom objects, ...) could only be read and replaced *as a whole* from routines. Reading one value out of them needed a `GET` with an array property (which already existed), and changing one value was not possible at all without rebuilding the whole object. This PR completes that: everywhere a routine addresses a property, it can now address a value **inside** it.

## What you can write now

**`SET` takes a key path, just like `GET` does** - the first entry is the property, the rest are the keys inside it:

```JSON
{
  "func": "SET",
  "collection": [ "widgetID" ],
  "property": [ "cardDefaults", "color" ],
  "value": "#00767a"
}
```

- The objects along the path are created if they do not exist yet, so you can build `css.default.background` on a widget that has no `css` at all. A key creates an *array* only when it is a number (`[ "cardTypes", 0 ]`, or a `${...}` that evaluates to a number) - a string key always creates an object key, so a routine keying a map by an id or a timestamp cannot accidentally build a huge list.
- Inside an array the key has to be an index of at most the current length (overwrite or append), anything else is reported and ignored - an array cannot grow by more than one element per `SET`.
- `__proto__`, `constructor` and `prototype` are refused as keys and only own properties are ever read, so a key path can never reach out of the property and into the prototype chain.
- `relation` applies to the old value *at that key*, so `"relation": "+"` increments `css.default.width` and appends to a nested string. (On an array `+` string-coerces, exactly as it does for a whole-property `SET`; the array relations are `sort`, `reverse`, `pop` and `shift`.)
- Everything else in the property stays untouched. If a value along the path was not an object and had to make room for the rest of the path, the `SET` happens and a `Warning:` naming the widget, the property and the lost value is reported.
- `id`, `parent` and `deck` hold a widget ID and reject a key path (they have no values inside them).

**`${PROPERTY ...}` and variables index with dots, at any depth:**

```
${PROPERTY css.default.background}
${PROPERTY css.default.background OF widgetID}
${PROPERTY cardTypes.$cardType.color OF $deckID}     // keys can be variables
${myVariable.a.b.c}                                  // variables used to stop after one level
```

The same token also works in `css` values and in `html` face objects (`"background": "${PROPERTY colors.main}"`) - and a change to the nested value re-renders the widget, because the dependency is tracked on the root property.

## Dots in key names

This was the sticking point in the issue discussion. The rule is:

- A **string** `property` is always taken verbatim, so a widget property that literally contains a dot keeps working exactly as before (`"property": "cardDefaults.color"` still means the property named `cardDefaults.color`).
- The **array** form is the unambiguous way to address anything, including keys with dots or spaces: `"property": [ "css", ".scoreboard .totalsLine", "color" ]`.
- In `${...}` a dot always indexes - that syntax never allowed dots before, so nothing can break. A key with a dot can still be reached there with the `\u002E` escape that identifiers already support.

## Effect on existing games

No file updater step and no legacy mode is proposed, but this is not a pure addition either - three things that already ran in shipped games behave differently now. All three look like bug fixes to me, so they are called out here rather than hidden behind a legacy mode; happy to gate any of them if you disagree.

1. **`GET` with an array `property` keeps falsy nested values.** The old code collapsed *every* falsy nested value to `null` (`c.map(v=>v && typeof v == 'object' && v[subkey] || null)`). A `GET` on `[ "counters", "score" ]` with a score of `0` returned `null`, was dropped by `skipMissing`, and skewed `aggregation` - it returns `0` now. Reading through a string or an array changed the same way: `[ "name", "length" ]` used to be `null` and is the length now.
2. **The `+` -> `=` fallback of `SET` is per widget.** `a.relation = '='` mutated the shared operation object, so as soon as one widget in the collection had a `null` property, every *later* widget in the same `SET` used `=` instead of `+`. Selecting all cards and doing `SET counters.x relation "+"` set them all to `1` when only the first one lacked the value; each widget decides for itself now. This is issue #3059 (3); the characterization test `tests/client/engine/set.test.js` that pinned the old behaviour flips with it.
3. **A mutating relation on a `SET` no longer loses its delta.** `sort`, `numericSort`, `numericStringSort`, `reverse`, `pop` and `shift` mutate their argument. The old value was read straight out of the widget's own state, so the relation changed that state in place; `set()`'s equality guard then saw no change and skipped the delta to the other clients *and* every change routine, and the change vanished on reload. The deep copy is taken before `compute()` now. This covers the key-path form only, which is the code this PR adds - a whole-property `SET … relation "sort"` still behaves as it does on `main` (see the To Do above).

Everything else is new syntax on JSON that did nothing before: `${PROPERTY a.b}` used to evaluate to `null`, and an array `property` in `SET` used to address a property literally named `a,b`. The public library replays (state-hash comparison over all 400 games) pass unchanged.

## Also in here

- `validator/validate_gamefile.js` accepts the array `property` of `SET` and the new `${PROPERTY ...}` syntax, checks variables that are used as keys inside a path, and rejects a key that is not a string or a number.
- `widget.js` now imports `compute_ops` explicitly, which is what makes `SET` (and every string expression) runnable in the jest environment - the new unit tests depend on it.
- The JSON editor offers a `property (key path)` insert button for `GET` and `SET` next to the plain `property` one, so the array form is discoverable without the wiki.
- An html card face keeps a nested `0`, `false` or `""` instead of rendering it empty (the top-level `${PROPERTY x}` keeps the behaviour it has in shipped games).
- `cssReplaceProperties` substitutes an empty string instead of the literal text `undefined` for a value that does not exist, and no longer expands `$&`/`` $` `` inside a property value as a replacement pattern.
- The routine log of the JSON editor names the whole key path a `GET` or a `SET` addressed - `first of 'css.default.background'`, `sum of 'counters.score'`, `'cardTypes[0].color' = "red"` - instead of only the property it starts from. Writing a key path into a property that reads as an empty string (an unset `css`) no longer reports a replaced value either, because there was none.

## Demo room on the test server

**Properties - Nested Values** is a tutorial-quality demo room for this feature, in two variants:

- **SET and GET** - a key path that builds `css.default.background` on a widget with no `css`, a `SET` that changes one key and leaves its neighbours alone, `relation "+"` on `counters.score`, a number key appending to an array, and a `GET` over a collection with `aggregation: "sum"` where one widget scores `0`.
- **PROPERTY expressions** - `${PROPERTY palette.main}` in a `css` string, a two-level path in the same `css`, `${PROPERTY palette.$which OF theme}` with the key in a variable, and a deck whose `html` card face is drawn from `${PROPERTY info.name}` / `info.power` / `info.color` (Water has a power of `0` and prints it).

It is deliberately **not** part of the diff - no tutorial or library file is touched. It is loaded in both rooms of this PR's test server, [pr-test-ai](https://test.virtualtabletop.io/PR-3095/pr-test-ai) and [pr-test](https://test.virtualtabletop.io/PR-3095/pr-test) (open the *Game Shelf* and filter for `Nested`), and the save file is at [Properties-Nested-Values.vtt](https://agent.virtualtabletop.io/reports/pr/1534030835265831065/Properties-Nested-Values.vtt). Screenshots of the two variants: [SET and GET](https://agent.virtualtabletop.io/reports/pr/1534030835265831065/nested-values-demo-1-set-and-get.png), [PROPERTY expressions](https://agent.virtualtabletop.io/reports/pr/1534030835265831065/nested-values-demo-2-property.png). It can be moved into `library/tutorials/` on request.

## Testing

- New jest suite `tests/client/nested-values.test.js` covering `SET`/`GET` key paths, object and array creation, relations, `${PROPERTY ...}` in values and in `css`, variable keys and multi-level variable indexing, plus the refused keys, the array index rules, the replacement warnings, the per-widget relation and the mutating relation firing its change routine.
- New jest suite `tests/client/card-face.test.js` covering the `${PROPERTY ...}` substitution of an `html` card face.
- Full jest suite green on the merged tree (1683 tests, 58 suites).
- TestCafe `publiclibrary-1.js` (state-hash replays), `routines.js` and `compute-1.js` green locally in Chromium after merging `main` - no state hashes changed.
- `validator/validate_gamefile_node.js` over every file in `library/games` and `library/tutorials` produces output byte-identical to `main` on the merged tree (main's validator rework included).
- Both demo variants pass `validator/validate_gamefile_node.js` and were walked end to end in Chromium against a local server.
- The one red check, `Production Environment`, fails on `Renaming a widget keeps its color controls clear and it movable`. That is a Chrome 152 regression in the widget-id input of edit mode, which this branch does not touch and which fails the same way on `main` ([analysis](https://github.com/ArnoldSmith86/virtualtabletop/pull/3095#issuecomment-5532732473)); a maintainer is taking it separately.

The wiki pages **Functions** (GET/SET) and **Routines** (${PROPERTY ...}) should be updated for this. Two things worth a sentence there: a key path over a collection of cards deep-copies the deck's inherited `cardTypes`/`faces` into every card's own state (the same as a whole-property `SET` today), and the array-creating key is a number *after* variable evaluation.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3095/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

Nice, well-scoped feature with real tests, but `setNestedValue` writes keys straight onto a plain object with no key filtering and no bounds on numeric keys. `SET` with `"property": ["cardTypes", "__proto__", "polluted"]` pollutes `Object.prototype` for every client and then **crashes the delta/undo pipeline** (`serverstate.js:405`); `["cardTypes", "100000"]` silently produces a 500 KB array of `null`s that is synced to the server, and `["cardTypes", "007"]` silently discards the value. All three reproduced locally against this branch.

## Findings

### 1. Prototype pollution + client crash via `__proto__` in a `SET` key path — blocker

`client/js/domhelpers.js:637` `setNestedValue` does `container[key] = ...` with no key filtering, and `container[key] === undefined ? null : container[key]` happily descends into `Object.prototype` (it isn't `undefined`). So a routine key path can walk out of the widget property and into the prototype chain.

Reproduced on this branch (jest, same harness as the new suite):

```json
{ "func": "SET", "property": [ "cardTypes", "__proto__", "polluted" ], "value": "yes" }
```

Result: `({}).polluted === "yes"`, `cardTypes` stays `{}` — and the routine throws:

```
    at addDeltaEntryToUndoProtocol (client/js/serverstate.js:405)
```

because the now-enumerable prototype key shows up in the `for(const widgetID in delta.s)` loops in `serverstate.js`, so `widgets.get('polluted')` is `undefined`. Every subsequent delta in that tab keeps crashing (in my run all later tests in the file failed until I deleted the polluted key). Since routines come from arbitrary shared/uploaded games in the public library, this is untrusted input reaching `Object.prototype` on every player's client, plus a persistent client-side breakage. `constructor`/`prototype` are the same class of key.

Fix: reject (or skip with a `problems.push`) `__proto__`, `constructor` and `prototype` as keys in `setNestedValue` — and ideally in `getNestedValue` too, so `${PROPERTY x.__proto__}` doesn't leak internals.

### 2. Numeric-string keys build unbounded / broken arrays — `client/js/domhelpers.js:641`

`String(key).match(/^[0-9]+$/) ? [] : {}` treats every all-digit key as an array index. Reproduced:

| routine | resulting `cardTypes` |
|---|---|
| `["cardTypes", "100000"]` | array whose JSON is **500 003 bytes** of `null`s |
| `["cardTypes", "007"]` | `[]` — value silently lost |
| `["cardTypes", "1754300000000"]` | `[]` — value silently lost (index ≥ 2³²−1 is not an array index) |

The first row is the dangerous one: the value goes through `w.set` → delta → server → every other client, so a routine keying a map by e.g. a timestamp or a large numeric id inflates room state by megabytes (and a larger-but-still-valid index OOMs the tab before it ever gets sent). The other two silently drop the write with no `problems.push`.

Suggestion: only create an array when the key is an actual JSON number (`typeof key == 'number'`) *and* a small in-range index (or only when the existing container is already an array), and treat everything else as an object key. `["cardTypes", "0", "color"]` meaning "object key `0`" and `["cardTypes", 0, "color"]` meaning "array index 0" is a rule users can reason about; "any digits ⇒ array" is not.

### 3. Writing a non-numeric key into an existing array property is a silent no-op — `client/js/widgets/widget.js:2026`

If the property is already an array, `setNestedValue` sets a non-index property on it, which `JSON.stringify` drops during state sync:

```json
{ "func": "SET", "property": [ "cardTypes", "a" ], "value": 5 }   // cardTypes was [1,2]
```

→ `cardTypes` stays `[1,2]`, no problem reported. Given the `property !== null && typeof property != 'object'` branch above already warns when it has to replace a scalar, the array case deserves at least the same warning.

### 4. `a.relation` is mutated inside the per-widget loop — `client/js/widgets/widget.js:2014-2015`

Pre-existing, but the key path makes it much easier to hit: `a.relation = '='` mutates the shared operation object, so once *one* widget in the collection has a missing nested key, every *subsequent* widget in the same `SET` gets `=` instead of `+`. E.g. `SELECT` all cards then `SET ["counters","x"] relation "+" value 1` on cards where only the first lacks `counters.x` sets them all to `1` instead of incrementing. Use a local `const relation = ...` inside the loop.

### 5. Validator no longer checks `$var` keys inside a path — `validator/validate_gamefile.js:463-481`

`parsePropertySyntax` now captures the path as group 3, but nothing validates it. `${PROPERTY cardTypes.$undefinedVar.color}` passes the validator silently, whereas the same undefined variable in the `OF` position is reported. Splitting group 3 on `.` and checking each `$`-prefixed key against `context.validVariables` would keep parity.

### Smaller notes

- `cssReplaceProperties` (`client/js/widgets/widget.js:683`) still uses `css.replace(match[0], value)`, so a nested value containing `$&`, ``$` `` or `$'` is expanded as a replacement pattern. Pre-existing, but nested values are more likely to be arbitrary user text than whole property values were — `(...)=>value` as the replacer would close it.
- Test coverage: the new suite covers the happy paths well, but nothing covers `__proto__`, the "property is not an object, replacing its value" branch (2023-2024), a key path against an array property, or `SET` over a multi-widget collection with `relation: "+"` — i.e. exactly the branches above. Worth adding alongside the fixes.
- Per `CLAUDE.md`, `SET`'s `property` now accepting an array is the kind of change that usually wants a matching `jeCommands` entry in `client/js/jsonedit.js` so the editor offers the array form; the validator side is already done here.

---

This PR really does change engine behavior — `client/js/widgets/widget.js` alters `SET`/`GET` semantics for every routine and extends the `${...}`/`${PROPERTY ...}` expression syntax used by all existing games — so it is excluded from beta auto-deployment by design, and independently it is not beta-safe while a routine can pollute `Object.prototype` and crash the delta/undo pipeline; I did not add the `beta` label.

VTT-AI-Agent:
Thanks - all five findings were real, all five are fixed in ed1feb1a. Details per point:

### 1. Prototype pollution + client crash - fixed

`setNestedValue` now refuses `__proto__`, `constructor` and `prototype` outright (a `problems.push`, the property is left untouched and nothing is written), and both helpers only ever look at **own** properties (`Object.prototype.hasOwnProperty`), so a path can no longer descend into the prototype chain in the first place - neither writing nor reading. Your exact repro now leaves `cardTypes` as it was, reports `Key "__proto__" is not allowed inside a property.` and `({}).polluted` stays `undefined`; the delta/undo pipeline is untouched.

`getNestedValue` keeping own-property semantics also means `${PROPERTY x.constructor}` evaluates to nothing instead of leaking a function into `JSON.stringify`.

### 2. Unbounded / broken numeric keys - fixed

Two rules replaced "any digits ⇒ array":

- A **new** container is only created as an array when the key is an actual JSON number (`"property": [ "cardTypes", 0 ]`). A string key - which is what a variable holding an id or a timestamp produces - always creates an object, so `["cardTypes","100000"]` is now the object key `"100000"`, and `["cardTypes","007"]` / `["cardTypes","1754300000000"]` keep their value instead of silently dropping it.
- Inside an **existing** array the key has to be an index and at most `length` (overwrite or append). Anything else is reported (`Index 100000 is out of range for an array of length 2.`) and ignored. That closes the "500 KB of nulls" hole for good: an array can only ever grow by one element per `SET`, so no path through this code can inflate the room state or OOM the tab.

### 3. Non-numeric key on an array property - fixed

Same rule as above, from the other side: it now reports `Key "a" is not an index, it cannot be set in an array.` instead of silently writing a property that `JSON.stringify` drops.

### 4. `a.relation` mutated inside the loop - fixed

The `+` → `=` fallback is a `const relation` local to the loop body now, so a widget with a missing nested key no longer downgrades the relation for every later widget in the collection. New test: two widgets, only the second has `cardTypes.count`, `relation: "+"`, `value: 1` → `1` and `11`.

### 5. Validator no longer checks `$var` keys inside a path - fixed

`${PROPERTY cardTypes.$undefinedVar.color}` and `${myVar.$undefinedVar}` are reported again (`uses undefined variable 'undefinedVar'`), for both the `PROPERTY` and the plain-variable form. While in there: `getCustomPropertyUsage` was registering the whole dotted string from `${PROPERTY css.default.background}` as a custom property name - it now registers only `css` - and `SET` with an array `property` registers its property name like `GET` already did.

### Smaller notes

- **`css.replace` with `$&`**: fixed, the replacement is a function now.
- **Test coverage**: 10 new cases in `tests/client/nested-values.test.js` (22 total) covering exactly the branches above - `__proto__` and `constructor`, the non-object-property branch, a key path against an array property (append, non-index key, out-of-range index), numeric strings as object keys, the multi-widget `relation: "+"` case, `GET` through the prototype chain, plus a regression test that `${string.length}` / `${array.1}` still work.
- **`jeCommands`**: nothing was added, removed or renamed here - `SET`'s existing `property` button is unchanged, only the shapes it accepts grew. `GET` has accepted an array `property` for a long time and has no separate command for it either, so adding one only for `SET` would be inconsistent; if the editor should offer the array form, it seems better to do that for both operations in its own PR.

### Verification

- Full jest suite green (194 tests, 10 suites).
- Full TestCafe suite green locally in Chromium - no state hashes changed.
- `validator/validate_gamefile_node.js` over every game in `library/games` produces byte-identical output to `main` (7643 lines), so none of the validator changes introduce a false positive on the public library.

Agreed on the `beta` label - not adding it.

VTT-AI-Agent:
> **CodeQL / Prototype-polluting assignment**
> This assignment may alter Object.prototype if a malicious '__proto__' string is injected from user controlled input.

Addressed in a94f196e. The write was already guarded, but the guard tested `String(key)` while the assignment used `key`, so the analysis could not see the two as the same value. The key is now made a primitive up front and compared directly:

```js
const key = typeof path[0] == 'number' ? path[0] : String(path[0]);
if(key === '__proto__' || key === 'constructor' || key === 'prototype') {
  problems.push(`Key ${JSON.stringify(key)} is not allowed inside a property.`);
  return object;
}
```

That is the sanitizer shape the query looks for, and it is also stricter than before: a key that is not a string but stringifies to `__proto__` (e.g. the array `["__proto__"]`) is now caught too, because the comparison happens after the conversion instead of on a separate copy. Reads go through `Object.prototype.hasOwnProperty`, so nothing descends into the prototype chain either.

Jest green (194 tests), including the `__proto__`/`constructor` regression tests added in ed1feb1a.

VTT-AI-Agent (commented):
## TL;DR

This is a **UI/UX review** (not a code review) of #3095. The change is engine-level, so its only user-facing surfaces are (a) the new *problem messages* shown in the editor's **Debug** panel and (b) what nested `${PROPERTY a.b}` renders into the board. I ran the branch locally on a scratch room and both work as advertised — a nested `SET` re-renders `css` correctly. The single most impactful improvement: **`Property X is not an object, replacing its value …` should follow the existing `Warning: …` convention and must not claim a replacement happened when the write is afterwards rejected** — as written it tells a game author their value was destroyed in cases where nothing changed at all.

<sub>Reviewed by running the branch on a local server (Chromium/Playwright), room seeded with a widget whose `css` is `background-color: ${PROPERTY colors.main}` plus two buttons doing nested `SET`s.</sub>

| before (`colors.main = red`) | after `SET ["colors","main"] = green` | after `SET ["text","sub"] = 1` |
|---|---|---|
| before | after | bad |

The middle shot is the good news: writing one key inside `colors` re-renders the widget through the dependency on the root property, exactly as the description promises. The right-hand shot is the concern below.

---

## Big improvements

**1. The "replacing its value" message can be wrong, and it should be a `Warning:`** — `client/js/widgets/widget.js:2025`

```js
if(property !== null && typeof property != 'object')
  problems.push(`Property ${mainProperty} is not an object, replacing its value ${JSON.stringify(property)}.`);
…
if(problems.length == problemsSoFar)
  await w.set(mainProperty, newProperty);
```

The message is pushed *before* `setNestedValue` runs. If `setNestedValue` then refuses the key (`__proto__`, a non-index in an array, an out-of-range index), the `SET` is skipped — but the author has already been told "replacing its value `"…"`". For a first-time game author, being told a property they care about was overwritten when it wasn't is worse than no message at all. Push it only on the success path, or word it in the past tense after the write.

Two smaller points on the same line:
- Every other non-fatal message in this file uses the `Warning: ` prefix (`Warning: Relation == interpreted as =`, `Warning: Mode … interpreted as set.` — 8 of them). This one is the same category (the operation *succeeded*, destructively) but reads like an error, and the Debug panel paints the whole expander red for it just like a real failure. Suggest: `Warning: Property "text" was not an object - its value "abc" was replaced.`
- Property names are unquoted while keys elsewhere are `JSON.stringify`'d (`Key "a" is not …`). Quoting both keeps the panel consistent and makes empty/space-containing names visible.

**2. Give the "value replaced" case a visible signal outside the Debug panel.** The right-hand screenshot above is what a player sees after `SET ["text","sub"]` on a widget whose `text` is a plain string: the board renders **`[object Object]`**. The only explanation lives in the Debug panel, which is closed by default and (as I confirmed locally) records nothing for routines that ran while it was shut. Since this PR makes it newly easy to turn a display string into an object by accident, it would be worth at least making the message loud and specific enough that someone searching for "[object Object]" finds it — e.g. name the widget and the resulting value: `Warning: Property "text" of widget "box" was the string "…" and was replaced by an object.`

**3. The feature has no affordance in the JSON editor.** `jeAddRoutineOperationCommands('SET', { …, property: 'parent', … })` (`client/js/jsonedit.js:1390`) still inserts a plain string, and nothing in the editor hints that `property` now takes an array key path or that `${PROPERTY a.b.c}` exists. The context buttons are effectively the app's in-product documentation, so as it stands the whole feature is discoverable only by reading the wiki — which the PR itself says still needs updating. Two concrete options that fit the existing style: (a) change the `SET`/`GET` `property` insert default to an array (`[ "css", "default", "background" ]`) so the shape is self-documenting, or (b) leave the default and land the wiki edits with the merge. Given the description explicitly defers the wiki, please don't let the editor be the only place a newcomer can look.

---

## Nitpicks

- `client/js/domhelpers.js:664` — `Key "a" is not an index, it cannot be set in an array.` is a comma splice and doesn't name the property. Suggest: `Property "cardTypes" is an array, so key "a" is not a valid index; SET ignored.` The neighbouring existing message (`null value being appended, SET ignored`) makes explicit that nothing happened — these new ones only imply it.
- `client/js/domhelpers.js:668` — `Index 100000 is out of range for an array of length 2.` doesn't state the rule, so a first-time user can't tell what *would* work. Suggest: `Index 100000 is out of range: an array of length 2 can only be written at index 0-2 (a SET can append at most one element).`
- `client/js/domhelpers.js:655` — `Key "__proto__" is not allowed inside a property.` reads as arbitrary to a non-programmer. One clause fixes it: `… is a reserved JavaScript key and is not allowed in a property path.`
- `client/js/widgets/widget.js:1986` — `Property parent holds a widget ID, it has no values inside it.` — comma splice again, and no hint at the fix. Suggest: `Property "parent" holds a widget ID, so it has no values inside it - use a plain property name.`
- `validator/validate_gamefile.js:341` — dropping the `GET` prefix leaves the row in the edit-mode problems table reading `property must be a non-empty string or array` with no operation name, while every sibling message in that block is `${func} uses …`. In the panel this column stands alone (see screenshot below) so the func is the useful part. Either thread the func through or word it `GET/SET property must be a non-empty string or array`.
- `client/js/widgets/card.js:162` — `… || ''` means a nested value of `0`, `false` or `""` renders as **empty** on an html card face. The PR description specifically calls out that `GET` no longer collapses nested `0`/`false`/`""`; the card face still does, so `${PROPERTY counters.$player.score}` showing a score of 0 will render a blank card. Worth matching the new `undefined ? '' : value` handling used in `cssReplaceProperties` (`widget.js:685`).

debug panel

---

- The messages above are the whole first-time-user surface of this PR, and they currently assume the reader knows JS object/array semantics (prototype chain, "index", "not an object"). The wording suggestions are all in that direction: say what was refused, why, and what the author should write instead.
- **Nothing visibly happens when a `${PROPERTY a.b}` path doesn't resolve.** In css the substitution becomes `''`, so `background-color: ;` is silently dropped by the browser and the widget just looks unstyled — no problem is pushed and no console note appears. A newcomer who mistypes `${PROPERTY colors.mian}` gets zero feedback. This isn't a regression (the top-level case behaved the same, printing `undefined`), and it may be out of scope here, but it becomes much more likely now that paths can be several keys deep. A debug-panel note when a `${PROPERTY …}` path resolves to `undefined` would be a big help.
- Positive note on clarity: the "a string key never creates an array" rule is the right default for a UI — it prevents a routine keyed by a timestamp from silently producing an enormous synced array, which would be a confusing "why is my room slow" mystery for an author with no way to see the cause.

Everything else in the change is invisible to end users and looks fine from a UI standpoint.

VTT-AI-Agent:
Thanks - the messages were the weak spot here. Everything except one point is fixed in 6585ead3.

### 1. "replacing its value" could be wrong, and it is a warning - fixed

The message is pushed **after** `setNestedValue` returned without a problem, i.e. only when the value really was replaced, and it now follows the `Warning: ` convention and quotes both names:

```
Warning: Property "text" of widget "box" was not an object - its value "abc" was replaced.
```

A refused key (`__proto__`, a non-index in an array, an out-of-range index) leaves the property untouched and no longer claims anything was replaced. The refusal messages themselves are now prefixed with the same `Property "…" of widget "…": ` in `widget.js`, so every problem of a nested `SET` says which widget and which property it is about - the helper in `domhelpers.js` stays generic.

New jest cases: one asserts the exact `Warning:` line on the replacing path, one asserts that a refused `__proto__` key on a string property reports the reserved-key problem and **not** `was replaced`.

### 2. A visible signal for the replaced case

The message now names the widget, the property and the old value (see above), which is what a search for `[object Object]` can be matched to. I did not add a signal outside the Debug panel: the only place a routine problem is shown today is that panel (plus `console.log` when the panel is closed), and inventing a second channel for this one message felt like a bigger UI decision than this PR should make. The message being specific enough to find is the part I could fix here.

### 3. Affordance in the JSON editor - added

`GET` and `SET` now have a second insert button next to `property`, so the key path shape is visible in the editor instead of only in the wiki:

property (key path) button

`property (key path)` inserts `[ "css", "default", "background" ]` (verified in Chromium: the button shows for `GET`/`SET` when the operation has no `property` yet and inserts the array, with the first key selected). The plain `property` button is unchanged, so nothing about the common case moves. The wiki still needs the prose.

### Nitpicks

- `domhelpers.js:664` → `Key "a" is not an array index - only an index can be set inside an array, so the SET was ignored.` (with the property/widget prefix from `widget.js`).
- `domhelpers.js:668` → `Index 100000 is out of range - an array of length 2 can only be written at index 0 to 2 because a SET appends at most one element, so the SET was ignored.` - it states the rule now.
- `domhelpers.js:655` → `Key "__proto__" is a reserved JavaScript key and cannot be used in a property path, so the SET was ignored.`
- `widget.js:1986` → `Property "parent" holds a widget ID, so it has no values inside it - use a plain property name.`
- `validate_gamefile.js:341` → the message is `${func} property must be a non-empty string or array` again, so the row keeps its `GET`/`SET`.
- `card.js:162` → a **nested** value of `0`, `false` or `""` renders as itself now, only a missing/`null` one renders empty. The top-level `${PROPERTY x}` keeps its `|| ''` on purpose: that path exists in shipped games, and turning a `null`/`0` property into visible text there would change how existing cards look.

### Not changed: no feedback when a `${PROPERTY a.b}` path does not resolve

Agreed that this is the weakest spot for a newcomer, but a `${...}` token is substituted while rendering (`cssReplaceProperties`, card faces), not while a routine runs, so there is no `problems` array to push to - it would need a new reporting path from rendering into the Debug panel and it would fire for every widget that legitimately styles itself with an optional property. It is the same behaviour a mistyped top-level `${PROPERTY colors}` has today, so it is not a regression of this PR; better as its own change.

### Verification

- Full jest suite green (195 tests, 10 suites).
- TestCafe `editor.js` (29 tests) and `routines.js` green locally in Chromium, `compute-1.js` green - no state hashes changed.
- `validator/validate_gamefile_node.js` over all 400 files in `library/games` produces output identical to `main`.

VTT-AI-Agent (commented):
## TL;DR

No new correctness, security, or regression issues found in the current revision. The prior prototype-pollution and unsafe-array-path blockers are fixed, their related edge cases are covered by the new tests, and the full current CI set is green.

## Detailed findings

No additional findings. I re-checked the final `setNestedValue`/`getNestedValue` behavior, the `SET` and `GET` integration, nested `${PROPERTY ...}` and variable evaluation, card/CSS substitution, validator handling, and the JSON-editor affordance; the fixes in `6585ead3` address the outstanding feedback without exposing another concrete failure.

I did not add the `beta` label because `client/js/widgets/widget.js` and `client/js/widgets/card.js` materially alter routine-operation and property-expression behavior for the engine, making this PR engine-ineligible for beta auto-deployment.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-tested feature — the key-path plumbing is careful (own-property reads only, reserved keys refused, arrays can grow by at most one, deep copy before mutating) and `npm test` is green here (195/195). My main concern is the compat claim: two of the changes **do** alter how existing games run — `GET` with an array `property` now returns nested `0`/`false`/`""` instead of `null`, and the `+`→`=` relation fallback is now per-widget instead of leaking to the rest of the collection. Both look like fixes, but they deserve an explicit maintainer decision rather than being covered by "none of this changes the meaning of JSON that did anything before".

## Findings

### 1. `GET` with a key path changes the result for existing games (compat)

`client/js/widgets/widget.js:1481` replaces

```js
for(const subkey of propertyPath)
  c = c.map(v=>v && typeof v == 'object' && v[subkey] || null);
```

with `getNestedValue(...)`. The old expression collapsed **every** falsy nested value to `null`; the new one keeps it. So an existing routine doing `GET property: ["someObj","count"]` where `count` is `0` now yields `0` where it used to yield `null` — which also flips `skipMissing` (`0` is no longer filtered out) and changes `aggregation: 'first'`/`'min'`/`'array'` results. Reading through a string/array also changed: `["name","length"]` used to be `null`, now it is the length.

The array `property` form of `GET` already existed and is used in shipped games, so this is not a no-op for legacy content the way the `SET` array form and `${PROPERTY a.b}` are. Per `docs/compat.md` this is the kind of "what existing JSON means" change that normally gets called out explicitly. I'd reword the PR description's "none of this changes the meaning of JSON that did anything before" and let the maintainer decide whether it needs a legacy mode or just a changelog line — the public-library replays passing is good evidence the blast radius is small, but it isn't zero.

### 2. The `+`-relation fallback is now per-widget (also compat)

Old code (`SET`):

```js
if(a.relation == '+' && w.get(String(a.property)) == null)
  a.relation = '=';
```

`a` is a per-operation deep copy, but it is shared across **all widgets in the collection** for that one execution. So once any widget's property was `null`, every *later* widget in the same `SET` silently used `=` instead of `+`. `widget.js:2016` makes the fallback local, and the new test "Then the relation of one widget does not affect the next one" locks that in. Clearly the right behaviour — but it is a real change to how existing games run and is worth naming in the description alongside item 1, because it is exactly the kind of thing a state-hash replay might not cover (it only shows up when a collection mixes null and non-null values).

### 3. A non-object *nested* value is discarded silently; only the top level warns

`widget.js:2030` warns when the whole property was not an object:

```
Warning: Property "css" of widget "x" was not an object - its value "color: blue" was replaced.
```

but `setNestedValue` (`domhelpers.js:667`) does the same replacement at every deeper level with no message. Confirmed locally:

```js
setNestedValue({ default: 'color: red' }, ['default','background'], 'red', problems)
// => {"default":{"background":"red"}}   problems: []
```

The author's `css.default` string is gone and nothing is reported. Since the top-level case was considered worth a warning, the nested case should get the same one (`setNestedValue` already has `problems` in hand and knows the key it is replacing) — otherwise a typo'd path quietly destroys a chunk of a widget's `css`/`faces`.

### 4. New CodeQL alert on `container[key] = newValue` (`domhelpers.js:678`)

github-advanced-security flagged this as a prototype-polluting assignment. Reading the code I believe it is a false positive: `key` is checked against `__proto__`/`constructor`/`prototype` before the assignment, arrays only accept `isArrayIndex` keys, and `getNestedValue` uses `hasOwnProperty`, so I could not construct a path that escapes the property. But the alert will sit on the security check, so it needs either an explicit dismissal or a shape CodeQL recognises as sanitised — e.g. `Object.defineProperty(container, key, { value: newValue, writable: true, enumerable: true, configurable: true })`, which is also immune to a `__proto__` setter regardless of the guard. Worth resolving one way or the other before merge rather than leaving a red check.

### 5. The two rendering paths have no automated coverage

`cssReplaceProperties` (`widget.js:682`) and the html card face (`card.js:159`) are the parts most likely to affect shipped games, and the new jest suite covers only routines/`${...}`. Two behaviour changes in there are untested:

- `cssReplaceProperties` now substitutes `''` for an undefined value; it previously substituted the literal string `undefined` into the CSS text.
- the replacement is now a function, so `$&`/`$1` in a property value are no longer expanded (a genuine fix — but nothing guards it against regressing).

A couple of jest cases asserting the generated css text for `${PROPERTY colors.main}` (present, missing, and a value containing `$&`) would be cheap given `tests/client/client-util.js` already builds widgets.

### 6. Validator only type-checks the first element of the key path

`validatePropertyPath` (`validator/validate_gamefile.js:331`) validates `value[0]` and ignores the rest, so `"property": [ "css", {} ]` or `[ "css", null ]` passes edit-mode validation and then misbehaves at runtime (`getNestedValue(obj, [null])`, or a key stringified to `"[object Object]"`). A `value.slice(1).every(k => typeof k == 'string' || typeof k == 'number')` check would close that, and it is the sort of thing the editor is there to catch.

### 7. Two smaller notes

- **Array creation depends on the runtime value, not just the JSON.** The documented rule is "a key creates an array only when it is a *number in the JSON*", but `evaluateVariablesRecursively` runs first, so `[ "cardTypes", "${index}" ]` with `index = 0` arrives as a JSON number and creates an array. That is probably the desired behaviour, but the wiki wording should say "a number" rather than "a number in the JSON", or the intent gets lost.
- **Inherited properties.** `w.get('cardTypes')`/`'faces'` on a *card* resolves through `inheritFrom` to the deck's object; a key-path `SET` over a collection of cards therefore deep-copies the deck's whole object into every card's own state (52 copies into the delta and the save). The same is true of a whole-property `SET` today, so this isn't a regression — but the key-path form makes "just tweak one colour on the selected cards" a natural thing to write, so it is worth a sentence in the wiki page.

I added the `beta` label: no — this PR changes routine-operation semantics (`SET`/`GET`), `${PROPERTY ...}` expression evaluation in `client/js/widgets/widget.js` and card face rendering in `client/js/widgets/card.js`, so it really does alter how existing games run and is excluded from beta auto-deployment by design.

VTT-AI-Agent:
Thanks - findings 3, 5 and 6 are fixed in 7b832ba, findings 1 and 2 are now spelled out in the PR description, and 4 was already covered by the previous commit. Details:

### 1. + 2. `GET` keeping falsy nested values / the per-widget `+` fallback (compat)

You are right that "none of this changes the meaning of JSON that did anything before" was too broad, and both of these deserve a maintainer decision rather than being buried. The description now has an **Effect on existing games** section that names them:

- `GET` with an array `property` used to collapse *every* falsy nested value to `null`, which also fed `skipMissing` and the aggregations - a `0` is a `0` now, and `[ "name", "length" ]` is the length instead of `null`.
- the `+` -> `=` fallback used to leak from the first widget with a `null` property to every later widget in the same `SET`.

I still think both are fixes rather than behaviour worth preserving, and the state-hash replays over all 400 public-library games are unchanged, but the section says so explicitly and offers to gate either one instead of quietly relying on that.

### 3. A non-object nested value was discarded silently - fixed

`setNestedValue` takes a `warnings` array now and reports every level where a value had to make room for the rest of the path, not just the top one. Your repro:

```js
setNestedValue({ default: 'color: red' }, ['default','background'], 'red', problems, warnings)
// warnings: ['Key "default" was not an object - its value "color: red" was replaced.']
```

`widget.js` prefixes it the same way it prefixes the refusals, so in the Debug panel it reads

```
Warning: Property "css" of widget "box": Key "default" was not an object - its value "color: red" was replaced.
```

Warnings are collected separately from problems and are only reported once the write actually went through, so a refused key still says nothing about a replacement. New jest case asserts the exact line.

### 4. CodeQL alert on `container[key] = newValue`

The alert had already been resolved before this review (a94f196 made the guard compare the same primitive the assignment uses, and the check on the head commit is green - `gh api .../code-scanning/alerts?ref=refs/pull/3095/head` returns no open alert). On top of that the assignment is now `Object.defineProperty(container, key, { value, writable: true, enumerable: true, configurable: true })`, which is what you suggested and is immune to an inherited setter regardless of the guard.

### 5. No coverage for the two rendering paths - fixed

- `tests/client/nested-values.test.js` gained four `cssReplaceProperties` cases: a nested value is substituted, a missing key becomes `''` (not the literal `undefined`), a value containing `` $&/$` `` is inserted verbatim, and the root property is registered as a dependency.
- `tests/client/card-face.test.js` is new and renders an `html` card face: nested value, nested `0` rendering as `0`, missing key rendering empty, and a top-level `${PROPERTY x}` unchanged. `card.js` needed an `export` on its class to be importable from jest, the same way `line.js` already has one.

### 6. Validator only type-checked `property[0]` - fixed

`validatePropertyPath` now walks the rest of the array as well, so `"property": [ "css", {} ]` and `[ "css", null ]` are reported (`SET property key {} must be a string or a number`) instead of reaching the runtime. `[ "css", "default", 0 ]` still passes.

### 7. Smaller notes

- **"a number in the JSON"**: reworded to "a number (`[ "cardTypes", 0 ]`, or a `${...}` that evaluates to a number)", and the wiki note at the bottom of the description says the same.
- **Inherited properties**: added to the same wiki note - a key-path `SET` over a collection of cards deep-copies the deck's `cardTypes`/`faces` into each card's own state, exactly like a whole-property `SET` does today.

### Verification

- Full jest suite green (204 tests, 11 suites).
- TestCafe `routines.js`, `compute-1.js` and `publiclibrary-1.js` green locally in Chromium - no state hashes changed.
- `validator/validate_gamefile_node.js` over all 400 files in `library/games` produces output byte-identical to before this commit.

VTT-AI-Agent (commented):
## TL;DR

The feature is in good shape: the key-path plumbing held up under fresh adversarial probes (prototype-chain isolation, array-growth cap, `length`-write refusal all verified locally), jest is 204/204 and CI including CodeQL is green. Two things need action: **the branch now conflicts with `main`** (61 commits behind; `card.js`/`widget.js` header conflicts, plus main's validator was reworked meanwhile, so the "validator output identical over the library" verification needs a re-run after the merge), and **mutating relations (`sort`, `reverse`, `pop`, `shift`, …) on a key path change state without ever sending a delta** — a silent client desync I reproduced locally. The latter is a pre-existing bug class for whole-property `SET`, but the new branch extends its reach to nested and deck-inherited values and can fix it for free by taking the deep copy one statement earlier.

## Findings

### 1. Mutating relations on a key path desync the clients (reproduced)

`client/js/widgets/widget.js:2020` runs `compute(relation, …, oldValue, …)` **before** the deep copy at `widget.js:2024`, and `oldValue` (`widget.js:2014`) is a live reference into `w.state[mainProperty]`. The relations that mutate their `x` argument — `sort`, `numericSort`, `numericStringSort`, `reverse`, `pop`, `shift` — therefore mutate the widget's state in place; by the time `w.set()` runs, `JSON.stringify(this.state[property])` already equals the new value, so `set()`'s equality guard (`statemanaged.js:114`) returns early: **no delta is sent, no `changeRoutine`/`…ChangeRoutine` fires, and the server and other clients never see the change** (it also silently vanishes on reload).

Reproduced on this branch in the jest harness:

```json
{ "func": "SET", "property": [ "data", "list" ], "relation": "sort" }
```

with `data: { list: [3,1,2] }` leaves the local `state.data` as `{"list":[1,2,3]}` while a `changeRoutine` on the widget never fires — the control run with `relation: "="` fires it normally.

The reach is worse for cards: `w.get(mainProperty)` can return the **deck's** live object by reference (`card.js` `getDefaultValue` → `cardPropertyGet`), so e.g. `SET ["cardTypes","Ace","tags"] relation "sort"` on a card sorts the deck's shared `cardTypes` locally with no delta for the deck at all.

To be fair: main behaves the same for whole-property `SET … relation "sort"` (same live reference, same early return), so this is a pre-existing bug class, not a regression. But the new key-path branch makes nested arrays — including deck-inherited ones — addressable for the first time, and it already pays for a deep copy; taking it **before** `compute()` and reading `oldValue` from the copy makes the whole new branch immune:

```js
const property = w.get(mainProperty);
const copy     = keyPath.length && property !== null && typeof property == 'object'
               ? JSON.parse(JSON.stringify(property)) : property;
const oldValue = keyPath.length ? getNestedValue(copy, keyPath) : property;
```

(The whole-property variant would still have the pre-existing behavior — worth a maintainer decision whether to fix it here too or separately.) A jest test asserting that a nested `relation: "sort"` fires `changeRoutine` would lock this in.

### 2. Branch conflicts with current main; post-merge re-verification needed

GitHub reports the PR as CONFLICTING: main is 61 commits ahead of the base (`a911be1a`). The conflicts themselves are trivial — both files' heads: main added a header block to `client/js/widgets/card.js` (the new `write` face type) and reordered imports in `widget.js:1`, colliding with this PR's `export class Card` / added imports; both sides are compatible. But two of the claims in the PR description were verified against the old base and should be re-run after the merge:

- main's validator was reworked (`validator/updated_gamefile.js` now pre-migrates files through `FileUpdater` before validating — on this branch the same file/validator pair produces `version is too old` where main's produces a full report). The "validator output identical over `library/games`" comparison needs a re-run on the merged tree. The `validate_gamefile.js` hunks themselves auto-merge cleanly.
- the library state-hash replays (TestCafe) for the same reason.

### 3. Missing tutorial-quality demo room (standing maintainer rule)

A feature of this size (SET key paths, nested `${PROPERTY a.b}`, multi-level variable indexing) needs a demo/test room authored to tutorial quality and published as a `.vtt` on this PR's own `pr-test-ai` test-server room — not committed to `library/` (correctly absent from the diff), but linked from the PR body. The PR currently shows only ad-hoc browser verification and the bot's `pr-test` room. Suggested shape, following the newest tutorials: two variants — *"Functions - GET/SET (key paths)"* (SET/GET with array `property`, object/array creation, the relation-on-nested-value case, the warnings) and *"Routines - nested ${PROPERTY}"* (`${PROPERTY a.b}`, `$variable` keys, use inside `css` and an `html` card face).

### 4. Minor

- The PR description (and future wiki text) says `relation: "+"` "appends to a nested string/array" — for arrays `+` string-coerces (`[1,2] + "x"` → `"1,2x"`), same as whole-property SET. Only the wording is wrong; `push` is not reachable as a SET relation (its `v` argument is `null` there).
- UI support: no new widget property, so nothing is owed to the property sidebar; there is no merged visual routine editor (only PR #2439), so the JSON editor's new `property (key path)` button is the right — and only possible — editor affordance today. When #2439's routine editor lands it will need to learn the array `property` form; worth a note over there.
- The open maintainer decision on legacy-gating the two behavior changes (`GET` keeping falsy nested values, per-widget `+`→`=` fallback) is already tracked in the To Do block — not re-raised here; both still look like fixes to me.

## Verified

- jest 204/204 locally on the branch; all CI checks green including CodeQL (the earlier prototype-pollution alert is fixed — direct probes confirm reserved keys are refused at every depth, only own properties are read, arrays grow by at most one, and an array's `length` cannot be written through a path).
- `export class Card` is safe: the client bundle is a single `<script type="module">` (`client/room.html:623`), matching the existing exports in `button.js`/`label.js`/`line.js`/`timer.js`.
- Validator output on the same input is byte-identical between this branch and its merge base.

The `beta` label is not added because this PR genuinely changes engine behavior — `SET`/`GET` routine-operation semantics, `${…}` expression evaluation and card-face rendering in `client/js/widgets/widget.js`/`card.js`, including two acknowledged behavior changes for shipped games — which excludes it from beta auto-deployment by design.

VTT-AI-Agent:
Thanks - all four points are addressed in 8ecc848 (merge) and e2696d2 (the desync fix).

### 1. Mutating relations on a key path desync the clients - fixed

Confirmed and fixed exactly as suggested: the deep copy is taken **before** `compute()` and the old value is read out of the copy, so `sort`/`numericSort`/`numericStringSort`/`reverse`/`pop`/`shift` mutate the copy and the widget's own state stays untouched until `set()` compares it.

```js
const property = w.get(mainProperty);
// the copy is taken before compute() so that a mutating relation (sort, reverse, pop, shift)
// works on it - mutating the live state would make w.set() see no change and skip both the
// delta to the other clients and the change routine
const copy     = keyPath.length && property !== null && typeof property == 'object' ? JSON.parse(JSON.stringify(property)) : property;
const oldValue = keyPath.length ? getNestedValue(copy, keyPath) : property;
```

New jest case in `tests/client/nested-values.test.js`: a widget with `data: { list: [3,1,2] }` and a `changeRoutine`, `SET property [ "data", "list" ] relation "sort"` - it asserts both the sorted state and that the change routine fired. Backing the one-line move out again makes that test fail, so it really pins the fix.

The deck case you called out follows from it: `w.get('cardTypes')` on a card can return the deck's live object, and the copy now happens before anything can sort it in place.

**The whole-property variant is deliberately left as it is on `main`** - it is the same pre-existing bug but not code this PR adds, and fixing it would make `SET … relation "sort"` start sending deltas in shipped games. I put it in the To Do block at the top of the description as a maintainer decision; it is a two-line change if you want it here rather than in its own PR.

### 2. Branch conflicts with main - merged, verifications re-run

`origin/main` is merged (`8ecc848`). The two conflicts were the header blocks you described and both sides are kept: `card.js` takes main's `enginePropertiesWithoutDefault` / `plainTextEditable()` block plus this PR's `export class Card`, and `widget.js` takes main's `mod` + `expression.js` imports plus this PR's `getNestedValue`/`setNestedValue` and `compute_ops`.

Everything re-verified on the merged tree:

- Full jest suite green: **1683 tests, 58 suites** (main grew the suite a lot; `tests/client/engine/set.test.js` needed the one expectation flip described below).
- `validator/validate_gamefile_node.js` over every file in `library/games` **and** `library/tutorials`, branch vs `main`, is **byte-identical** - so main's `updated_gamefile.js` rework changes nothing about this PR's validator hunks.
- TestCafe `publiclibrary-1.js`, `routines.js` and `compute-1.js` green locally in Chromium - no state hashes changed.

One thing the merge surfaced that is worth naming: main added `tests/client/engine/set.test.js` with a characterization test that **pins** the `+`→`=` leak as issue #3059 (3), with the comment "c3 keeps its 'x' in a correct implementation; it does not here". This PR is that fix, so the expectation flips from `[ 'x!', '!', '!' ]` to `[ 'x!', '!', 'x!' ]` and the comment now describes the fixed behaviour - which is exactly the visible signal the pinned test was written to give.

### 3. Demo room - published

**Properties - Nested Values**, two variants, authored to tutorial quality and living only on the test server:

- **SET and GET** - a key path building `css.default.background` on a widget with no `css` at all, a `SET` that changes one key and leaves the rounded corners and the text colour next to it alone, `relation "+"` on `counters.score`, a number key appending to an array (and refusing to leave a gap), and a `GET` over a collection with `aggregation: "sum"` where the middle widget scores `0`.
- **PROPERTY expressions** - `${PROPERTY palette.main}` in a `css` string, a two-level path (`palette.shades.dark` / `.light`) in one `css`, `${PROPERTY palette.$which OF theme}` with the key in a variable, and a four-card deck whose `html` face is drawn entirely from `${PROPERTY info.name}` / `info.power` / `info.color` - Water has a power of `0` and prints it.

Both variants pass `validator/validate_gamefile_node.js`, carry the shared `_meta.info` block and a library image, and were walked end to end in Chromium. Nothing under `library/` is touched; it can be moved into `library/tutorials/` if you want it there.

Play it: <https://test.virtualtabletop.io/PR-3095/pr-test-ai> · save file: [Properties - Nested Values.vtt](https://agent.virtualtabletop.io/reports/pr/1534030835265831065/Properties%20-%20Nested%20Values.vtt)

### 4. Minor

- **`relation "+"` on an array**: you are right, the description was wrong. It now says `+` appends to a nested *string*, that on an array it string-coerces exactly like a whole-property `SET`, and that the array relations are `sort`, `reverse`, `pop` and `shift`.
- **UI support / #2439**: agreed - the JSON editor's `property (key path)` button is the only editor affordance that exists today. I have not touched #2439.
- **The legacy-gate question** is still open in the To Do block, now joined by the whole-property question from point 1.

96LawDawg:
The demo file [Properties - Nested Values.vtt](https://agent.virtualtabletop.io/reports/pr/1534030835265831065/Properties%20-%20Nested%20Values.vtt) is not loaded into pr-test-ai and manual uploading gives a 404 error. Fix.

VTT-AI-Agent:
It is in **pr-test-ai** and playable now, and the download link is fixed. Details:

**What I changed**

- **The file name had spaces**, and that is the one link in the chain that really breaks: plain `curl` refuses `.../Properties%20-%20Nested%20Values.vtt` outright (*"URL using bad/illegal format or missing URL"*). The save is republished space-free as **[Properties-Nested-Values.vtt](https://agent.virtualtabletop.io/reports/pr/1534030835265831065/Properties-Nested-Values.vtt)** and the description links that now.
- **The shelf tile was unusable.** The game carries `showName: false` like the tutorials do, but its library image was a wide screenshot squeezed into a square tile, so it showed a cropped strip and no title - easy to walk past even when it is right there. It now has a proper square library image in the tutorials' style: [shelf screenshot](https://agent.virtualtabletop.io/reports/pr/1534030835265831065/nested-values-demo-0-shelf.png). The save is also 6 KB instead of 47 KB.
- **The room is tidied up.** Two duplicate entries (from the upload retries while debugging this) are gone, there is exactly one entry left, and it is loaded as the active game, so [pr-test-ai](https://test.virtualtabletop.io/PR-3095/pr-test-ai) opens straight into it. I also added a copy to [pr-test](https://test.virtualtabletop.io/PR-3095/pr-test) - the shelf is per room, and that is the room the test-server bot links to at the bottom of the description, so the game genuinely was not there.

**Verified just now against this PR's test server**

- Both variants open and play: [SET and GET](https://agent.virtualtabletop.io/reports/pr/1534030835265831065/nested-values-demo-1-set-and-get.png), [PROPERTY expressions](https://agent.virtualtabletop.io/reports/pr/1534030835265831065/nested-values-demo-2-property.png).
- Downloading the save returns the exact bytes of the published file.
- *Add game → Upload files* with that file into a fresh room on the test server works, and so does *Add game → Add link* with the report URL.

**About the 404:** `404: Unable to load and add the game.` is what `addState` answers when the uploaded body is not a readable zip. I could not reproduce it through the UI; the two ways I can produce it are (a) a PUT whose body never reaches the server's raw parser - a script uploading without `Content-Type: application/octet-stream` - and (b) a `.vtt` that is not the save, e.g. a truncated download or a saved HTML error page under that name. Since the old URL contained spaces, a client that mangled it would land in exactly (b), which is why the rename is my first suspect.

If it still fails for you, please re-download from the new link and tell me the room, the button you used and the size of the file you got, and I will chase it further.

96LawDawg:
The Debug output for GET needs to be improved. For example in the demo variant for SET and GET, in button2, the GET is used on property ["css","default","background"] and the debug output is: GET first of 'css' in 'DEFAULT' => var background = "#6a4a9a". And if the property was one layer deeper it is even more confusing. Make the debug more clear what is actually being retrieved while still being succinct.  Similarly, button 5 reports: GET sum of 'counters' in 'DEFAULT' but really it is getting a sum of a sub-property (score) within counters.

VTT-AI-Agent:
Good catch - fixed in 6ace990.

The summary of a `GET`/`SET` only ever named the property the path starts from. It now spells out the whole path the way a routine author writes it, so your two examples read:

```
GET  first of 'css.default.background' in 'DEFAULT' => var background = "#6a4a9a"
GET  sum of 'counters.score' in 'DEFAULT' => var total = 12
SET  'cardTypes[0].color' = "red" for widgets in 'DEFAULT'
```

An array index and a key that is not a plain identifier are shown in brackets (`cardTypes[0].color`, `css[".scoreboard .totalsLine"].color`), so a deeper path stays unambiguous without getting long. A plain property is unchanged (`first of 'id' in 'DEFAULT'`).

One thing I noticed while checking this in the demo room: the first key path written into a `css` that was never set reported `Warning: Property "css" of widget "example1" was not an object - its value "" was replaced.` and got the red **failed** marker in the log. An unset `css` reads as an empty string, so nothing was lost - that warning is gone now.

Screenshot of the log after the demo buttons: https://agent.virtualtabletop.io/reports/pr/1534030835265831065/nested-values-debug-log.png

Covered by four new cases in `tests/client/nested-values.test.js` (the summaries a logged routine writes) plus one for the dropped warning; full jest suite green.

VTT-AI-Agent:
### The red `editor.js` check is not caused by this branch

`Renaming a widget keeps its color controls clear and it movable` fails with

```
A native alert dialog was invoked on page ".../testcafe-testing", but no handler was set for it.
```

It only fails on runners that have **Chrome 152**, and passes on the ones that still have Chrome 151 — GitHub is mid-rollout, so the same commit goes red or green depending on which image it lands on:

| job | Chrome | result |
| --- | --- | --- |
| TestCafe Chrome, attempt 1 | 152.0.0.0 | fail |
| TestCafe Chrome, attempt 2 (re-run, same commit) | 151.0.0.0 | pass |
| Production Environment, attempts 1 and 2 | 152.0.0.0 | fail |
| TestCafe Firefox | – | pass |

The same test failed on Chrome 152 on `fix-debug-panel-logging-after-edit-mode-reopens` too, which shares nothing with this branch, and the code involved (`client/js/editor/sidebar/properties.js`) is not touched here.

### What actually happens

Reproduced locally against Chrome for Testing 152.0.7977.82. Catching the dialog with `setNativeDialogHandler` shows it is

```
alert: A widget with the id "new" already exists.
```

and the room afterwards holds exactly one widget, `new` — the rename itself worked. `createWidgetIdInput`'s `onchange` runs **twice** for one rename on Chrome 152: the first run renames the widget, the properties module re-renders and drops the input while it still has focus, and Chrome 152 fires a second `change` from that removal. The second run sees the new id already taken and alerts. On Chrome 151 the second `change` never arrives.

So it is a real bug for anyone renaming a widget in edit mode on Chrome 152, not a test-only artefact — the rename works but the editor claims the id is taken. It belongs in its own PR rather than in this one, which is about property key paths.

VTT-AI-Agent:
### The Chrome 152 rename alert stays out of this PR

I asked the maintainers who should fix the edit-mode rename bug behind the red `Production Environment` check (the analysis is [above](https://github.com/ArnoldSmith86/virtualtabletop/pull/3095#issuecomment-5532732473)). The answer came in over Discord, in the "Human feedback" thread:

> **Q:** Chrome 152 breaks widget renaming in edit mode (bogus "A widget with the id X already exists." alert) and turns the editor.js TestCafe test red on every branch. Should I fix it, and where?

> Leave it, a maintainer will take it

So this branch keeps its hands off `client/js/jsonedit.js`'s id input, and the item is gone from the To Do block at the top. The failing check is `Renaming a widget keeps its color controls clear and it movable` in the minified-client run; it fails the same way on `main`, and every other check on this PR is green.

The two open questions in the To Do block — whether the `GET` falsy-value and per-widget `relation` changes want a legacy gate, and whether the whole-property `SET … relation "sort"` desync should be fixed here too — are still waiting on a decision.

PR #3085, opened 2026-08-02
Feature/custom viewport aspect ratio

mousewax:
Lets a game define the size of its board instead of hardcoding 1600x1000 everywhere.

A new **Board Settings** section in Edit Mode > Game Settings sets `_meta.gameSettings.boardSize`. Everything that used to assume a 1600x1000 board — the layout math, the JSON editor, the property sliders, drag buttons, pile handles, zoom/pan clamping, mouse coordinates, the edit-mode grid — reads a shared `viewportConfig` instead. Games that don't set anything keep the 1600x1000 default and behave exactly as before.

🌟 Key Changes

- **`client/js/calculateLayout.js`** — the layout math from `setScale()`, extracted and unit tested. It is `main`'s algorithm unchanged, with `1600`/`1000` replaced by the game's board size: the board gets the whole window and the toolbar sits in whatever margin the window's aspect ratio leaves over, unless there is less than a toolbar width of margin — then the board shrinks to make room (that is the case that offers the "hide toolbar" button). It returns the body class `main` already uses (`wideToolbar` / `horizontalToolbar` / `aspectTooGood` / none), so #3057's toolbar CSS and its `updateToolbarLayout()` compaction and scrolling keep working unchanged on top of it — one layout vocabulary, not two.
- **Edit mode module panel** — where it goes (beside the room, above it, or covering it) used to be media queries on a hardcoded 16/20 window aspect ratio, which is half of the default board’s 16:10. It is computed from the board’s shape now (`calculateEditModuleClasses()`), which is the same thing for a 1600x1000 board and the right thing for any other one.
- **`--roomWidth` / `--roomHeight`** are set on `documentElement`, so the CSS that used literal `1600px`/`1000px` follows the board size too. The two remaining `layout.css` media queries become layout-class selectors saying the same thing for the game's own board: `@media (min-aspect-ratio: 1600/1000)`, and the `@media (orientation: portrait)` that shows the "please rotate your device to landscape" nag - which, keyed off the window alone, fired on a portrait board in a portrait window, the best fit there is. `isOrientationMismatch()` decides it from the board's shape now, so only a landscape board asks to be turned landscape.
- **A way back from a hidden toolbar.** Hiding the toolbar (the button `aspectTooGood` offers) had no counterpart short of a reload. That was survivable while `aspectTooGood` was a rare window shape; now that a game can put every player in it, a thin strip at the left edge brings the toolbar back.
[...]

PR #3097, opened 2026-08-04
Size the overlays from the board instead of the window

VTT-AI-Agent:
Follow-up to #3085 (custom viewport aspect ratio).

## The problem

The overlays are children of `#roomArea`, so the space they have to lay themselves out in is the **rendered board**, not the window. Their breakpoints were `@media` queries, which measured the window — the same thing only as long as every board was 1600x1000. Now that a game can pick its own board size, a square or portrait board leaves a much smaller overlay in the very same window and still got the layout of a wide one.

The Game Shelf is the obvious symptom: in a 1600x1000 window, a 800x1200 board gives the shelf a **667px wide** overlay — and the window-keyed media query still put **7 columns** of unreadable, clipped thumbnails into it.

| before | after |
| --- | --- |
| before | after |

(same window, same 800x1200 board — 7 columns of ~90px tiles vs. 3 columns of ~222px tiles, which is the tile size a default board has always had)

Square board, same window: [before](https://agent.virtualtabletop.io/reports/pr/1534093705814872164/shelf-before-w1600x1000-b1000x1000.png) / [after](https://agent.virtualtabletop.io/reports/pr/1534093705814872164/shelf-after-w1600x1000-b1000x1000.png).

## The change

`#roomArea` becomes a size container (`container: roomArea / size`) and every overlay breakpoint becomes a container query against it.

That translation is one-for-one and mechanical: each of these media queries paired a width with the height at 16:10 (`(max-width: 600px), (max-height: 375px)` — the "whichever of the two the board is limited by" idiom), so `@media (max-width: X), (max-height: X/1.6)` became `@container roomArea ((max-width: X) or (max-height: X/1.6))` with the same numbers. On a default 1600x1000 board the same breakpoints therefore fire in the same order — but the board is always a bit smaller than the window (toolbar strip plus letterboxing), so each one now trips at a slightly larger window than before: the shelf drops a column, and the game details sidebar disappears, a little earlier. Boards with a different shape are where the layout really changes.

Converted: `overlays/states.css` (Game Shelf, game details, add/save game, share link, update image), `overlays/welcome.css` (link details), `overlays/misc.css` (confirm), plus the edit-mode JSON overlay (`editmode.css`), which `initializeEditMode` moves into `#roomArea`. Everything that is *not* laid out inside the board keeps its media queries: the JSON editor pane, and the deck editor's export dialog and the symbol picker — `DeckEditor.initializeDOM()` re-parents those into `#editor`, where they are positioned against the viewport and a `@container roomArea` rule would have no container to match against at all.

`setSidebar()` (`client/js/overlays/states.js`), which decides whether the game details are a sidebar of the shelf or an overlay of their own, measures `#roomArea` instead of the window so that it agrees with the container query that lays that sidebar out - and it is called from `setScale()`, so it also follows a game that brings its own board size along. While the two disagreed, clicking a game in the shelf did nothing at all in the band where the board is smaller than the window.

## Testing

- `npm test` — 199 passed.
- TestCafe (chromium headless): `boardsize.js`, `toolbar.js`, `editor.js`, `publiclibrary-1..4.js` all pass.
- New case in `tests/testcafe/boardsize.js`: *"The game shelf sizes its tiles from the board, not from the window"* — asserts a portrait board drops columns and keeps the tile width comparable to the default board. Verified that it **fails** on `main`'s CSS and passes with this change.
- Checked by hand in a browser that the default 1600x1000 board renders the shelf, the game details sidebar, the players and about overlays as before.

No creator-facing behavior changes, no new widget/routine properties, so no validator or `jeCommands` updates needed.

## Follow-up: the rest of the fixed-pixel media queries

> I'm pretty sure there shouldn't be any fixed pixel media queries anymore because everything is so dynamic. Check, fix and tell me which ones should remain and why.

([Discord](https://discord.com/channels/770758631146782780/1534093705814872164/1534111790814855261))

Swept the whole repo for `@media`. One more was keyed to the wrong box, plus one shared rule that is half in the room and half out.

### The editor sidebar

`@media (max-width: 1600px)` iconified the editor sidebar — 128px with button labels, 36px without. Those 92px come out of the room, but only while the room is what the window's *width* limits. 1600px is roughly what a 16:10 board needs at full height, one number for every board: a portrait board lost its labels in a window with hundreds of pixels of unused board margin next to it, and a wide board kept them in a window where they did cost board size. It also took them away at e.g. 1600x900, where the room is limited by its height and is exactly as big either way.

`isEditSidebarNarrow()` decides it from the board now, the same way `calculateEditModuleClasses()` next to it does. Both module-panel layouts also stop repeating the 36px sidebar width and get it from that class instead — which incidentally fixes them iconifying the sidebar without shrinking the button labels on windows wider than the old breakpoint.

1000x1600 board in a 1280x800 window — the board is 478px wide in both shots, so the labels are free:

| before | after |
| --- | --- |
| before | after |

Default board in a 1600x900 window: [before](https://agent.virtualtabletop.io/reports/pr/1534093705814872164/sidebar-before-default-w1600x900.png) / [after](https://agent.virtualtabletop.io/reports/pr/1534093705814872164/sidebar-after-default-w1600x900.png) (room 1382x864 in both).

### `p[icon]`

Two elements use it: the welcome overlay's user-generated-content warning, which is laid out inside the board (→ container query), and the symbol picker's hint paragraph, which is not always in the room — that one moves into the media query that already handles the rest of that overlay.

### What stays a media query, and why

| where | why |
| --- | --- |
| `fonts.css` — `#symbolPickerOverlay` (2 blocks) | the deck editor re-parents the picker into `#editor` / `#deckEditorMainCol`, where there is no `roomArea` container at all — a container query doesn't fall back, it silently stops matching |
| `editor/deckeditor.css` — export dialog at 700px | same thing: `initializeDOM()` moves it into `#editor`, where it is positioned against the viewport |
| `jsonedit.css` — `#jeLog` at 1000px / 600px | the standalone JSON editor is a top-level element the size of the window with `position: fixed` panes; the log's offsets are against the window and the board isn't involved |
| `validator/validate_library.js` at 768px | that's the standalone HTML validation report page, not the room |
| `editor/deckeditor.js` — `@media screen` in the print sheet | not a size breakpoint (it separates the print preview from the printout) |
| `assets/branding/favicon.svg` — `prefers-color-scheme` | not a size breakpoint |

Not touched here, but the same family: `getAvailableRoomRectangle()` (`client/js/editor/layout.js`) still decides where the room starts from `window.innerWidth < 700` and a window-aspect test rather than from the board.

### Testing (follow-up)

- `npm test` — 203 passed, including four new `isEditSidebarNarrow` cases.
- TestCafe (chromium headless): `boardsize.js`, `editor.js`, `toolbar.js` all pass.
- New case in `tests/testcafe/boardsize.js`: *"The editor sidebar keeps its labels while the board is not short of the space"* — 1280x800 window, iconified on the default board and labelled on a portrait one. Verified that it fails without the CSS change.
- Checked in a browser that a module panel opened next to the now-labelled sidebar still lays out (the panel's left edge, and with it the room, does not depend on the sidebar's width).

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3097/pr-test (or any other room on that server)

### DIFF
--- a/client/css/editmode.css
+++ b/client/css/editmode.css
@@ -207,7 +207,7 @@
   height: 100%;
 }
 
-@media only screen and (max-width: 500px) {
+@container roomArea (max-width: 500px) {
   #editJSONoverlay textarea {
     min-height: 100%;
   }
--- a/client/css/editor/deckeditor.css
+++ b/client/css/editor/deckeditor.css
@@ -931,7 +931,9 @@ body.deckEditorRoomVisible #deckEditorMainCol > .overlay {
   overflow: visible;
 }
 
-@media (max-width: 700px) {
+/* the export dialog is one of the #editorOverlays, which initializeEditMode moves into
+   #roomArea - so it is as wide as the board, not as the window */
+@container roomArea (max-width: 700px) {
   .deckEditorExportOptions {
     grid-template-columns: 1fr;
   }
--- a/client/css/editor/deckeditor.css
+++ b/client/css/editor/deckeditor.css
@@ -931,9 +931,9 @@ body.deckEditorRoomVisible #deckEditorMainCol > .overlay {
   overflow: visible;
 }
 
-/* the export dialog is one of the #editorOverlays, which initializeEditMode moves into
-   #roomArea - so it is as wide as the board, not as the window */
-@container roomArea (max-width: 700px) {
+/* stays a media query: initializeDOM() re-parents this dialog into #editor, where it is positioned
+   against the viewport - a container query on #roomArea has no container to resolve against there */
+@media (max-width: 700px) {
   .deckEditorExportOptions {
     grid-template-columns: 1fr;
   }
--- a/client/css/editor/sidebar.css
+++ b/client/css/editor/sidebar.css
@@ -245,29 +245,31 @@ body.lightsOff #editorModules {
   color: #f8f8f2;
 }
 
-/* iconify sidebar icons */
-@media (max-width: 1600px) {
-  :root {
-    --editSidebarWidth: 36px;
-  }
-  #editorSidebar button {
-    font-size: 0;
-    padding: 5px;
-  }
-  #editorSidebar button::before {
-    margin: 0;
-  }
-  #editorSidebar button span {
-    font-size: 13px;
-  }
+/* iconify sidebar icons - set by isEditSidebarNarrow() because the window width at which the
+   labels stop being affordable depends on the board's shape, which a media query can't know.
+   It used to be @media (max-width: 1600px), which took them away whenever the window was
+   narrower than a 16:10 board at full height, even where the room is limited by its height
+   and does not get any bigger without them. The module panel layouts below always iconify. */
+body.narrowEditSidebar {
+  --editSidebarWidth: 36px;
+}
+body.narrowEditSidebar #editorSidebar button {
+  font-size: 0;
+  padding: 5px;
+}
+body.narrowEditSidebar #editorSidebar button::before {
+  margin: 0;
+}
+body.narrowEditSidebar #editorSidebar button span {
+  font-size: 13px;
 }
 
 /* module above the room - the classes below are set by calculateEditModuleClasses() because the window
    shape they kick in at depends on the board's shape, which a media query can't know. They used to be two
    media queries in this order, and just like those they can both be active at once - the overlay comes
-   second, so it still wins where they overlap. */
+   second, so it still wins where they overlap. Both imply narrowEditSidebar, which is where the 36px
+   sidebar they leave room for comes from. */
 body.edit.editModulesAbove {
-  --editSidebarWidth: 36px;
   --roomTop: calc(100 * var(--vh) - var(--roomHeight) * var(--scale) / var(--roomZoom) + (1 - var(--roomZoom)) / 2 * var(--roomHeight) * var(--scale) / var(--roomZoom));
 }
 body.editModulesAbove #editorModules {
@@ -282,9 +284,6 @@ body.editModulesAbove #editorModules {
 }
 
 /* module as fullscreen overlay */
-body.edit.editModulesOverlay {
-  --editSidebarWidth: 36px;
-}
 body.editModulesOverlay #editorModules {
   left: 0;
   top: var(--editToolbarHeight);
--- a/client/css/fonts.css
+++ b/client/css/fonts.css
@@ -484,12 +484,12 @@
   display: none;
 }
 
-@media (max-width: 600px), (max-height: 375px) {
+@container roomArea ((max-width: 600px) or (max-height: 375px)) {
   #symbolPickerOverlay {
     padding: 2%;
   }
 }
-@media (max-width: 300px), (max-height: 187px) {
+@container roomArea ((max-width: 300px) or (max-height: 187px)) {
   #symbolPickerOverlay {
     padding: 0;
   }
@@ -519,7 +519,7 @@ p[icon]::before {
   position: absolute;
   left: -40px;
 }
-@media (max-width: 600px), (max-height: 375px) {
+@container roomArea ((max-width: 600px) or (max-height: 375px)) {
   p[icon] {
     margin: 10px;
     margin-left: 40px;
--- a/client/css/fonts.css
+++ b/client/css/fonts.css
@@ -490,6 +490,14 @@
   #symbolPickerOverlay {
     padding: 2%;
   }
+  #symbolPickerOverlay p[icon] {
+    margin: 10px;
+    margin-left: 40px;
+  }
+  #symbolPickerOverlay p[icon]::before {
+    font-size: 150%;
+    left: -30px;
+  }
 }
 @media (max-width: 300px), (max-height: 187px) {
   #symbolPickerOverlay {
@@ -521,7 +529,10 @@ p[icon]::before {
   position: absolute;
   left: -40px;
 }
-@media (max-width: 600px), (max-height: 375px) {
+/* the other one of these is the welcome overlay's user-generated-content warning, which is laid
+   out inside the board - the symbol picker's copy is handled by the media query above, since that
+   overlay is not always in the room */
+@container roomArea ((max-width: 600px) or (max-height: 375px)) {
   p[icon] {
     margin: 10px;
     margin-left: 40px;
--- a/client/css/fonts.css
+++ b/client/css/fonts.css
@@ -484,12 +484,14 @@
   display: none;
 }
 
-@container roomArea ((max-width: 600px) or (max-height: 375px)) {
+/* stay media queries: the deck editor moves the symbol picker out of the room overlays into #editor
+   or #deckEditorMainCol, so there is no #roomArea container to query while it is open there */
+@media (max-width: 600px), (max-height: 375px) {
   #symbolPickerOverlay {
     padding: 2%;
   }
 }
-@container roomArea ((max-width: 300px) or (max-height: 187px)) {
+@media (max-width: 300px), (max-height: 187px) {
   #symbolPickerOverlay {
     padding: 0;
   }
@@ -519,7 +521,7 @@ p[icon]::before {
   position: absolute;
   left: -40px;
 }
-@container roomArea ((max-width: 600px) or (max-height: 375px)) {
+@media (max-width: 600px), (max-height: 375px) {
   p[icon] {
     margin: 10px;
     margin-left: 40px;
--- a/client/css/layout.css
+++ b/client/css/layout.css
@@ -49,7 +49,13 @@ body.loadingEditMode * {
   cursor: wait;
 }
 
+/* the overlays are children of #roomArea, so their size is the board's, not the window's.
+   Their breakpoints are container queries against this box: as media queries they measured
+   the window, which only said something about how much room an overlay has back when every
+   board was 1600x1000. A square or portrait board now leaves a much smaller overlay in the
+   same window and would still have gotten the layout of a wide one. */
 #roomArea {
+  container: roomArea / size;
   position: fixed;
   overflow: clip;
   width: calc(var(--roomWidth) * var(--scale));
--- a/client/css/layout.css
+++ b/client/css/layout.css
@@ -53,7 +53,10 @@ body.loadingEditMode * {
    Their breakpoints are container queries against this box: as media queries they measured
    the window, which only said something about how much room an overlay has back when every
    board was 1600x1000. A square or portrait board now leaves a much smaller overlay in the
-   same window and would still have gotten the layout of a wide one. */
+   same window and would still have gotten the layout of a wide one.
+   Only for overlays that stay in here: an element the deck editor re-parents into #editor (the
+   symbol picker, its own dialogs) has no roomArea container left, and a @container rule on it
+   silently stops matching at any size - those keep their media queries. */
 #roomArea {
   container: roomArea / size;
   position: fixed;
--- a/client/css/overlays/misc.css
+++ b/client/css/overlays/misc.css
@@ -17,13 +17,13 @@
     margin: 0 auto;
   }
 
-  @media (max-width: 500px), (max-height: 312px) {
+  @container roomArea ((max-width: 500px) or (max-height: 312px)) {
     #confirmOverlay {
       justify-content: unset;
     }
   }
 
-  @media (max-width: 340px), (max-height: 212px) {
+  @container roomArea ((max-width: 340px) or (max-height: 212px)) {
     #confirmOverlay {
       padding: 0;
     }
--- a/client/css/overlays/states.css
+++ b/client/css/overlays/states.css
@@ -147,19 +147,19 @@
   overflow: auto;
 }
 
-@media (max-width: 600px), (max-height: 375px) {
+@container roomArea ((max-width: 600px) or (max-height: 375px)) {
   #statesList { overflow: unset; }
   #statesOverlay { overflow: auto; }
 }
 
-@media (max-width: 2260px), (max-height: 1412px) { #statesOverlay { --columns: 9; } }
-@media (max-width: 2000px), (max-height: 1250px) { #statesOverlay { --columns: 8; } }
-@media (max-width: 1700px), (max-height: 1062px) { #statesOverlay { --columns: 7; } }
-@media (max-width: 1420px), (max-height:  887px) { #statesOverlay { --columns: 6; } }
-@media (max-width: 1120px), (max-height:  700px) { #statesOverlay { --columns: 5; } }
-@media (max-width:  800px), (max-height:  500px) { #statesOverlay { --columns: 4; } }
-@media (max-width:  680px), (max-height:  425px) { #statesOverlay { --columns: 3; } }
-@media (max-width:  520px), (max-height:  325px) { #statesOverlay { --columns: 2; } }
+@container roomArea ((max-width: 2260px) or (max-height: 1412px)) { #statesOverlay { --columns: 9; } }
+@container roomArea ((max-width: 2000px) or (max-height: 1250px)) { #statesOverlay { --columns: 8; } }
+@container roomArea ((max-width: 1700px) or (max-height: 1062px)) { #statesOverlay { --columns: 7; } }
+@container roomArea ((max-width: 1420px) or (max-height:  887px)) { #statesOverlay { --columns: 6; } }
+@container roomArea ((max-width: 1120px) or (max-height:  700px)) { #statesOverlay { --columns: 5; } }
+@container roomArea ((max-width:  800px) or (max-height:  500px)) { #statesOverlay { --columns: 4; } }
+@container roomArea ((max-width:  680px) or (max-height:  425px)) { #statesOverlay { --columns: 3; } }
+@container roomArea ((max-width:  520px) or (max-height:  325px)) { #statesOverlay { --columns: 2; } }
 
 #statesList .title {
   border-top: 1px solid black;
@@ -412,7 +412,7 @@ div:nth-of-type(2) > .list > .roomState .star {
 }
 
 
-@media (max-width: 600px), (max-height: 375px) {
+@container roomArea ((max-width: 600px) or (max-height: 375px)) {
   #statesOverlay h2.title {
     position: unset;
     margin:  0 !important;
@@ -434,7 +434,7 @@ div:nth-of-type(2) > .list > .roomState .star {
     clear: both;
   }
 }
-@media (max-width: 320px), (max-height: 200px) {
+@container roomArea ((max-width: 320px) or (max-height: 200px)) {
   #filterOverflow > div {
     padding: 2px;
   }
@@ -940,7 +940,7 @@ div:nth-of-type(2) > .list > .roomState .star {
   margin-bottom: 6px;
 }
 
-@media (max-width: 1260px), (max-height: 787px), (min-width: 1421px) and (min-height: 888px) {
+@container roomArea ((max-width: 1260px) or (max-height: 787px) or ((min-width: 1421px) and (min-height: 888px))) {
   #stateDetailsOverlay {
     --mainDetailsHeight: 200px;
     --similarDetailsHeight: 160px;
@@ -963,12 +963,12 @@ div:nth-of-type(2) > .list > .roomState .star {
     display: none;
   }
 }
-@media (max-width: 1040px), (max-height: 650px), (min-width: 1421px) and (min-height: 888px) {
+@container roomArea ((max-width: 1040px) or (max-height: 650px) or ((min-width: 1421px) and (min-height: 888px))) {
   .variant {
     padding: .11rem .5rem;
   }
 }
-@media (max-width: 957px), (max-height: 600px), (max-width: 1000px) and (max-height: 643px), (min-width: 1421px) and (min-height: 888px) {
+@container roomArea ((max-width: 957px) or (max-height: 600px) or ((max-width: 1000px) and (max-height: 643px)) or ((min-width: 1421px) and (min-height: 888px))) {
   #stateDetailsOverlay {
     --mainDetailsHeight: 150px;
     --similarDetailsHeight: 150px;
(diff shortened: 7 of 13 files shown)


VTT-AI-Agent (commented):
## TL;DR

The core idea is right and the media→container translation is faithful, but two of the converted rules now target elements that are **not** descendants of `#roomArea` at the time they're shown — the deck editor explicitly re-parents them into `#editor`. `@container roomArea (…)` with no matching container never evaluates true, so those breakpoints are now dead: the export dialog's single-column fallback and the symbol picker's narrow-screen padding no longer fire at all. The `#roomArea` overlays themselves (shelf, details, add/save, share, confirm, JSON overlay) are correct.

## Findings

### 1. `client/css/editor/deckeditor.css:933` — the export dialog is not inside `#roomArea` when it's open (dead rule + wrong comment)

The new comment says the export dialog "is one of the `#editorOverlays`, which `initializeEditMode` moves into `#roomArea` — so it is as wide as the board". That's only true until the deck editor initializes. `DeckEditor.initializeDOM()` moves it straight back out:

```js
// client/js/editor/deckeditor.js:316
$('#editor').append($('#deckEditorExportOverlay'));
$('#editor').append($('#deckEditorImportOverlay'));
$('#editor').append($('#deckEditorNewDeckOverlay'));
$('#editor').append($('#symbolPickerOverlay'));
$('#editor').append($('#libraryDecksOverlay'));
```

`initializeDOM()` is called from `open()` (deckeditor.js:902) and `openEmpty()` (deckeditor.js:976), i.e. it has *always* run before `openExportOverlay()` can show the dialog (deckeditor.js:521). So at display time `#deckEditorExportOverlay` has no `roomArea` container in its ancestor chain, the container query is never true, and `.deckEditorExportOptions` keeps its two-column grid at every size. Previously `@media (max-width: 700px)` correctly collapsed it to one column on a narrow window — and the dialog really is window-sized there (`#deckEditorExportOverlay { position: fixed; top: var(--editToolbarHeight); right: var(--editSidebarWidth); bottom: 0; left: 0 }`, deckeditor.css:478). Concrete failure: open the deck editor at ~600px wide, hit Export — the option grid stays two columns and the inputs are squeezed/overflowing instead of stacking.

Suggested fix: this one was window-anchored to begin with (same category as the editor sidebar and JSON pane you deliberately left alone), so simply revert the hunk to `@media (max-width: 700px)`.

### 2. `client/css/fonts.css:487,492,522` — same problem for the symbol picker

`#symbolPickerOverlay` starts in `#roomArea` (room.html:470) but the deck editor parks it in `#editor` (deckeditor.js:319, 1051, 1102) or, while a card-view icon property is being edited, in `#deckEditorMainCol` (deckeditor.js:926, 992). In all of those the `@container roomArea (…)` rules can't match, so the padding reductions (`padding: 2%` / `padding: 0`) and the `p[icon]` reflow no longer apply anywhere in the deck editor, however narrow the window or the column is. Note `p[icon]` has exactly one instance in the whole client (room.html:478, inside the symbol picker), so that rule is dead in the deck editor too.

If you want the container-query semantics here rather than a revert, the clean version is to declare the *same* container name on every box that hosts these overlays, e.g. `#roomArea, #editor, #deckEditorMainCol { container: overlayHost / size }` — then a query resolves against whichever host the overlay currently sits in, which is exactly the "space the overlay actually has" property you're after. Worth doing anyway as a general guard: any future `$('#editor').append(someOverlay)` silently disables every `@container roomArea` rule on it, with no error and no visual clue.

### 3. Description overstates the equivalence on a default board

"On a default 1600x1000 board a given window therefore still picks exactly the same breakpoints as before" isn't quite right: `#roomArea` is always *smaller* than the window (toolbar strip plus letterboxing). Media-query firing condition was `min(windowW, windowH*1.6) <= X`; the container condition is `boardW <= X` where `boardW = min(windowW - toolbar, windowH*1.6)`. So every breakpoint now trips slightly earlier — the shelf will drop a column, and the details sidebar (`min-width: 1421px and min-height: 888px`) will disappear, at window sizes where they didn't before. That's arguably the intended behavior, but it does mean the default board is not byte-identical, and it's worth saying so rather than claiming pixel-for-pixel parity.

### 4. `tests/testcafe/boardsize.js:79` — `resizeWindow` leaks into the next test

The new test resizes to 1280x800 and never restores; TestCafe keeps the window size for the rest of the fixture, so "The Apply confirmation goes away…" now runs in a smaller window than it used to (it drives `#editorSidebar [icon=settings]`, `#boardHeight`, `.boardSizeActions button[icon=check]`). It passes today, but it's an invisible coupling — either restore the size at the end of the new test or make it the last test in the file.

Minor, non-blocking: `shelfTiles` derives tile width as `#roomArea` width ÷ `--columns` rather than measuring a real tile, so it would not catch a regression in the grid itself (padding/gap changes, or `--columns` being ignored). Measuring `.roomState` / the grid item's `getBoundingClientRect().width` would assert the thing the test's name claims.

### Verified as *not* an issue

`container-type: size` implies layout containment, which per css-contain would make `#roomArea` a containing block for `position: fixed` descendants — that would have broken `#symbolPickerOverlay.symbolPickerAboveEditor { position: fixed; inset: 0 }` and clipped it to the board via `overflow: clip`. I checked in Chromium: `container: roomArea / size` reports `contain: none` and a fixed `inset: 0` child still spans the viewport (an explicit `contain: layout` on the same box *does* clip it). So no regression there. No CSS counters exist in the client, so the style containment is inert too.

I added the `beta` label: the PR touches only CSS and one TestCafe file — no engine files (`client/js/widgets/`, `compute.js`, `serverstate.js`) at all — so it cannot corrupt game state or crash the server, and the shelf layout on non-default board sizes is exactly the kind of thing that benefits from real players trying it.

VTT-AI-Agent:
Thanks — findings 1 and 2 were right, and the dead rules are fixed in a44dd137.

**1 & 2 — the deck editor's overlays (reverted to `@media`).** Confirmed: `DeckEditor.initializeDOM()` re-parents `#deckEditorExportOverlay` and `#symbolPickerOverlay` into `#editor` (and `open()`/`openEmpty()` always run before either can be shown), so `@container roomArea (…)` had nothing to resolve against and the export dialog's single-column fallback plus the picker's narrow-screen padding never fired. Both are positioned against the *viewport* there (`position: fixed` on the export dialog, `.symbolPickerAboveEditor { inset: 0 }` on the picker), so the window really is the right thing for them to measure — I reverted those three hunks to media queries rather than naming a second container, and replaced the factually wrong comment with one saying why they stay media queries.

I skipped the `#roomArea, #editor, #deckEditorMainCol { container: overlayHost / size }` variant on purpose: `container-type: size` on `#deckEditorMainCol` (a `flex-grow: 1; min-width: 0; min-height: 0` flex item) makes its contents stop contributing to its size, which is a deck-editor layout change this PR has no business making. Instead the `#roomArea` container declaration in `layout.css` now carries the warning explicitly — re-parenting an overlay out of `#roomArea` silently disables every `@container roomArea` rule on it — so the next person who moves an overlay has it in front of them.

**3 — description.** Fixed: it now says the same breakpoints fire in the same order but each trips at a *slightly larger window* than before, because `#roomArea` is always smaller than the window (toolbar strip plus letterboxing) — no more "exactly the same breakpoints" claim.

**4 — test.** The new test is now the last one in `boardsize.js`, with a comment saying why, so its `resizeWindow(1280, 800)` no longer changes what "The Apply confirmation goes away…" runs in. (I first tried restoring the size from `window.outerWidth/outerHeight`; in headless those are 0, which resized the window to nothing and broke the following test — so: last in file, matching what `editor.js`/`compute-4.js` already do.)

Also took the minor point: `shelfTiles` now measures a real tile (`.roomState.visible` → `getBoundingClientRect().width`) instead of dividing `#roomArea`'s width by `--columns`, so it would catch a regression in the grid itself.

Verified: `npm test` 199 passed, `npx testcafe chromium:headless tests/testcafe/boardsize.js` 5/5 passed.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review). The premise is right and the payoff is big — on a portrait board the Game Shelf goes from unreadable 90px thumbnails to normal tiles. But the conversion is only half done: `setSidebar()` in `client/js/overlays/states.js` still measures the **window**, while the sidebar CSS now measures the **board**, and in the gap between them **clicking a game in the shelf does nothing at all** — no details, no Play button. I reproduced that on a plain default board at a 1460x920 window, and on any portrait board in a large window.

**What I exercised:** ran the branch locally and clicked through the Game Shelf, game details, Add game, Save game, Share link, Delete confirm, Players and About overlays, at 390x844 (phone), 820x1180 (tablet), 1280x800 (laptop) and 1920x1080 (desktop), each with the default 1600x1000 board **and** a 1000x1600 portrait board, and captured the same matrix on `main` for comparison. I could not reach the edit-mode **Edit widget JSON** overlay (`#editJSONoverlay`) or the deck editor's export/symbol-picker dialogs by automation, so those are reviewed statically.

---

## Big layout improvements

### 1. The sidebar breakpoint moved to the board, but the JS that toggles it did not — the game shelf becomes a dead end

`client/css/overlays/states.css:1124` is now `@container roomArea ((min-width: 1421px) and (min-height: 888px))`, and it is that rule that supplies

```css
#statesOverlay.withDetails > #stateDetailsOverlay { display: flex !important; ... }
```

The `withDetails` class, though, is still set from the window in `client/js/overlays/states.js:1081`:

```js
const vw = Math.max(document.documentElement.clientWidth || 0, window.innerWidth || 0)
const vh = Math.max(document.documentElement.clientHeight || 0, window.innerHeight || 0)
const bigEnough = vw >= 1421 && vh >= 888;
```

Because the board is always *smaller* than the window (toolbar strip + letterboxing), there is now a whole band where JS says "sidebar" — so it re-parents `#stateDetailsOverlay` into `#statesOverlay` and stops using the full-screen details overlay — while the container query says "not sidebar" and never gives it a `display`. Result: `#stateDetailsOverlay` stays `display:none` and **clicking a game tile has no visible effect whatsoever**.

Measured on this branch, default 1600x1000 board, click on the first public-library tile:

| window | board (`#roomArea`) | `withDetails` | details `display` | result |
| 1460x920 | 1416x885 | true | **none** | ❌ [nothing happens](https://agent.virtualtabletop.io/reports/ui-review/1534110451070406727/deadclick-default-1460x920.png) |
| 1500x950 | 1456x910 | true | flex | ✅ sidebar |
| 1920x1080 | 1728x1080 | true | flex | ✅ sidebar |

And with a 1000x1600 board in a 1920x1080 window it is dead at *every* desktop size, because the board can never reach 1421px wide:

| `main` (works, if cramped) | this PR (click does nothing) |
| --- | --- |
| [main-portrait-details-desktop.png](https://agent.virtualtabletop.io/reports/ui-review/1534110451070406727/main-portrait-details-desktop.png) | [probe-details-portrait-desktop.png](https://agent.virtualtabletop.io/reports/ui-review/1534110451070406727/probe-details-portrait-desktop.png) |

For a first-time user this is the worst possible failure mode: the shelf looks perfect, and the only way into a game — the tile → details → **PLAY** path — silently does nothing, with no error and nothing to click next.

Suggested fix, in the spirit of the rest of the PR: make `setSidebar()` measure the same box the CSS does, e.g. `const r = $('#roomArea').getBoundingClientRect(); const bigEnough = r.width >= 1421 && r.height >= 888;` — and call it not just on `window resize` (`states.js:1233`) but also whenever the board is re-laid out (`setScale`/board-size change), since loading a game with its own `boardSize` changes the container without any window resize.

### 2. Consider which overlays should be board-shaped at all

This PR makes the overlays honestly reflect the box they live in — but that also makes it visible that the box is sometimes far too small to be a usable dialog. On a 390x844 phone with a landscape board, `#roomArea` is **390x244**, so the whole Game Shelf is a 244px-tall letterbox strip wedged between the branding wash and the horizontal toolbar, with "Public Library" clipped behind the toolbar: [pr-default-shelf-phone.png](https://agent.virtualtabletop.io/reports/ui-review/1534110451070406727/pr-default-shelf-phone.png) (identical on `main` — pre-existing, not caused here).

The overlays that are *modal* — Game Shelf, game details, Add game, Save game, Share link, confirm — are not part of the play surface; nothing is gained by clipping them to the board. It would be a real improvement to position those against the window (full-viewport, minus the toolbar) and keep the board-container queries for the things that genuinely overlay the play area. That would also make the whole class of "JS measures the window / CSS measures the board" mismatches in finding 1 go away for those overlays, instead of requiring each one to be kept in sync.

If that is out of scope for this PR, it is worth a follow-up issue — the PR's own comment in `layout.css:52-59` is a good place to note it.

- **`--columns` has no fallback outside every query.** `client/css/overlays/states.css:155-162` sets `--columns` only from `@container` blocks; if `#roomArea` ever loses its container (edit mode re-parenting, or an overlay shown while `#roomArea` is `display:none`) a container query simply never matches, and unlike the old media queries there is no "window is at least *something*" floor. A plain `#statesOverlay { --columns: 10; }` default before the query block would make that degradation graceful rather than silent.
- **The board can never satisfy the widest breakpoints.** `(max-width: 2260px)` → 9 columns is the top of the ladder, but a board only reaches ~2260px on a >2400px window, so on ordinary displays the shelf now tops out one or two columns lower than before. That is the intended direction, but on the 1920x1080 desktop the tiles are noticeably chunkier than on `main` ([pr](https://agent.virtualtabletop.io/reports/ui-review/1534110451070406727/pr-default-shelf-desktop.png) vs [main](https://agent.virtualtabletop.io/reports/ui-review/1534110451070406727/main-default-shelf-desktop.png)) — worth a deliberate look at whether the numbers should shift down by the toolbar+letterbox margin rather than being kept literally identical.
- **`editmode.css:210`** — `@container roomArea (max-width: 500px)` now fires on a 1280px laptop whenever the board is narrow (I measured a 1000x1600 board at 478px wide on a 1280x800 window). The rule itself is only `#editJSONoverlay textarea { min-height: 100% }`, so it is harmless, but it is the same class of "500px used to mean phone, now means narrow board" shift — worth confirming by hand, as I could not open that overlay via automation.
- **Pre-existing, still visible after the change:** in full-screen game details the "next game" preview card at the top right overlaps and clips its own title ("4-Lette Words") — [main-default-details-laptop.png](https://agent.virtualtabletop.io/reports/ui-review/1534110451070406727/main-default-details-laptop.png).
- The empty personal-shelf area keeps reserving a large blank block above "Public Library" even when the message is two lines ([pr-portrait-shelf-desktop.png](https://agent.virtualtabletop.io/reports/ui-review/1534110451070406727/pr-portrait-shelf-desktop.png)). On a board-sized overlay that blank block is now a much bigger share of the available height than it was; letting that section shrink to its content would buy a whole extra row of tiles on narrow boards.

- The headline win is exactly a readability win, and it is large: on a portrait board the shelf goes from tiles where only "Similar to Samurai" is legible and every game name is clipped ([main](https://agent.virtualtabletop.io/reports/ui-review/1534110451070406727/main-portrait-shelf-desktop.png)) to normal, readable cards ([this PR](https://agent.virtualtabletop.io/reports/ui-review/1534110451070406727/pr-portrait-shelf-desktop.png)). Same window, same board.
- Finding 1 is primarily a *comprehension* failure, not just a layout one — a newcomer clicking a game and getting nothing has no way to guess that the app thinks a details panel is already open. Even if the sizing fix lands, a guard would be cheap insurance: if `withDetails` is on but the details panel has no layout box, fall back to `showStatesOverlay('stateDetailsOverlay')`.
- Unchanged by this PR but newly more prominent now that the details overlay is board-sized: the ☰ button at the top-left of the game details screen ([pr-portrait-details-tablet.png](https://agent.virtualtabletop.io/reports/ui-review/1534110451070406727/pr-portrait-details-tablet.png)) has no label or tooltip, and sits right next to an unlabelled ✕. A `title=`/`.tooltip` on it (the toolbar buttons already use that pattern) would help.
- Everything else I opened reads fine at all four viewports and both board shapes: Add game ([tablet](https://agent.virtualtabletop.io/reports/ui-review/1534110451070406727/pr-portrait-addstate-tablet.png)), Save game, Share link, the Delete confirm ([tablet](https://agent.virtualtabletop.io/reports/ui-review/1534110451070406727/pr-portrait-confirm-tablet.png)), Players and About — all closable, no clipping or overflow I could not also reproduce on `main`.

VTT-AI-Agent:
Thanks — the blocking finding was real and is fixed in 2aa87de3.

## 1. The dead click — fixed

Confirmed exactly as described: `setSidebar()` used the same 1421x888 numbers as the container query but measured the window, and the board is always smaller than the window, so in the band between them it re-parented `#stateDetailsOverlay` into the shelf where the container query never gave it a `display`.

`setSidebar()` now measures `#roomArea` — the same box the CSS queries — and it is called from `setScale()` rather than from the resize listener in `states.js`. That covers your second point too: a game that brings its own `boardSize` changes the container without any window resize, and `setScale()` is what runs on that path. The resize listener had to give up its call anyway, since a `resize` event reaches it *before* `setScale()` has given the board its new size, so it was deciding on the previous one.

Default board, 1460x920 window, clicking the first tile:

| before | after |
| --- | --- |
| before | after |

Portrait board in a 1920x1080 window: [before](https://agent.virtualtabletop.io/reports/pr/1534093705814872164/deadclick-broken-portrait-1920x1080.png) / [after](https://agent.virtualtabletop.io/reports/pr/1534093705814872164/deadclick-fixed-portrait-1920x1080.png).

Measured after the fix — `withDetails` and the details panel agree everywhere now, and changing the board size with the details open flips it live:

| window | board | `withDetails` | details |
| --- | --- | --- | --- |
| 1460x920, default board | 1416x885 | false | full overlay ✅ |
| 1500x950, default board | 1456x910 | true | sidebar ✅ |
| 1920x1080, default board | 1728x1080 | true | sidebar ✅ |
| 1920x1080, portrait board | 675x1080 | false | full overlay ✅ |
| 1920x1080, sidebar open → load a portrait game | 1728x1080 → 675x1080 | true → false | sidebar → full overlay ✅ |

New TestCafe case in `boardsize.js`: *"Clicking a game in the shelf opens its details on every board and window size"* — 1460x920 on a default board and 1920x1080 on a portrait one, asserting the details overlay and its PLAY button are visible. Verified that it fails on the previous commit.

I did not add the "`withDetails` is on but the panel has no layout box → fall back" guard: with both sides measuring the same box with the same numbers they can only disagree if someone changes one of them, and the test now catches that with a clear failure instead of a silent fallback.

## 2. `--columns` fallback — already there

`#statesOverlay { --columns: 10 }` is in the base rule at `states.css:2`, outside every query, so the ladder already degrades to 10 columns if no container query matches. No change needed.

## 3. Columns on wide screens — measured, and they don't actually change

I compared `--columns` and the real tile width on this branch against `main`'s `states.css`, same room, same viewports:

| viewport | board | main | this PR |
| --- | --- | --- | --- |
| 2560x1440 | 2304x1440 | 10 cols / 207px | 10 cols / 207px |
| 1920x1080 | 1728x1080 | 8 cols / 193px | 8 cols / 193px |
| 1600x1000 | 1556x973 | 7 cols / 199px | 7 cols / 199px |
| 1440x900 | 1396x873 | 7 cols / 176px | **6 cols / 209px** |
| 1280x800 | 1236x773 | 6 cols / 182px | 6 cols / 182px |

So the desktop sizes you flagged come out identical, 1920x1080 included. The one that moves is 1440x900, and it moves *up* in tile size (176px → 209px) — which is the point: the tiles go back to the ~200px they have at every other size instead of being squeezed because the ladder was reading a window the shelf doesn't get. Shifting the numbers down by the toolbar+letterbox margin would undo exactly that, so I'd rather keep them as the faithful 1:1 translation.

## 4. Modal overlays against the window — a design call, out of scope here

It's a fair question, but it is the opposite of what this PR was asked to do (size the overlays from the board they live in), and it would mean moving the Game Shelf, details, add/save/share and confirm dialogs out of `#roomArea` — a much bigger change than a breakpoint conversion, with consequences for game CSS that targets the overlays inside the room. The 390x844-with-a-landscape-board case you show is identical on `main`, so nothing regresses here. Worth an issue; I'd rather not decide it inside this PR.

## 5. `editmode.css:210`

Intended: `#editJSONoverlay` is one of the overlays `initializeEditMode` puts inside `#roomArea`, so "narrow" there genuinely means "narrow board" now. The rule is only `textarea { min-height: 100% }`, which is if anything more appropriate on a 478px-wide board than it was keyed to a 478px-wide *window*.

## 6. The rest

The remaining points — the next-game preview clipping its own title, the blank block above "Public Library", and the unlabelled ☰ button — all reproduce on `main` and none of them are about which box a breakpoint measures. I'd rather not fold unrelated UI fixes into a breakpoint PR; they'd make good separate issues.

Verified after the fix: `npm test` 203 passed; TestCafe `boardsize.js` (7), `publiclibrary-1..4.js` (18), `toolbar.js` + `editor.js` (27) all pass.

VTT-AI-Agent (commented):
## TL;DR

The current head looks good: the earlier container-host and shelf/details breakpoint mismatches are fixed, the remaining media/container queries match the boxes they measure, and the new layout logic has focused unit and browser coverage. I found no remaining correctness, security, regression, or test-coverage issue worth raising.

## Detailed review

I re-checked the full diff at `2aa87de3`, including the follow-up changes:

- `client/css/editor/deckeditor.css` and the symbol-picker rules in `client/css/fonts.css` correctly remain viewport media queries because those overlays are re-parented outside `#roomArea` before use.
- `client/js/overlays/states.js` now derives the details-sidebar mode from `#roomArea.getBoundingClientRect()` using the same thresholds as `client/css/overlays/states.css`, and `client/js/main.js` refreshes it after every board layout, including a game-driven board-size change.
- `isEditSidebarNarrow()` matches the labelled sidebar and toolbar dimensions used by the editor layout, with coverage for default, portrait, wide, and module-panel arrangements.
- The TestCafe cases cover the user-visible shelf sizing, sidebar-label behavior, and the formerly dead game-details click path. All current CI checks, including Chrome/Firefox TestCafe, Jest/build, server startup, and CodeQL, pass.

The `beta` label was already present; the PR changes peripheral layout/UI code and tests, touches no engine paths, and is safe and useful for player testing.

VTT-AI-Agent:
Thanks — automated review 3/4 raised no findings, so there's nothing to change on this round. Head stays at `2aa87de3` (the container-host and shelf/details fixes from review 2/4 are already in), and CI is green.

ArnoldSmith86 (approved):

PR #3098, opened 2026-08-04
Make negative turn values in the TURN operation wrap instead of crashing the routine

VTT-AI-Agent:
The `TURN` routine operation accepts any integer for `turn` (the validator explicitly allows negative integers, and the wiki says the position is computed "modulo the length of the list"), but negative values never worked: they threw and aborted the routine.

**Cause:** JavaScript's `%` keeps the sign of the dividend, so in `client/js/widgets/widget.js`:

- forward/backward: `unskipped[(turn - offset) % unskipped.length]` — with `turn: -1` and 4 seats that is `unskipped[-1]` → `undefined`
- position: `unskipped[(a.turn - 1) % unskipped.length]` — same for any `turn <= 0`

`target` then stays `undefined` and the operation throws on `target.get("index")`, so the whole routine stops.

**Fix:** wrap the computed index into range with a proper non-negative modulo. A negative `turn` now steps the opposite way around the seat list (so `turnCycle: "forward"` with `turn: -1` is the same seat as `"backward"` with `turn: 1`), and for `turnCycle: "position"` a negative value counts from the end, with `-1` being the last active seat.

The `offset` that compensates for a current seat with `skipTurn` being missing from the list is applied for positive `turn` values only — it corrects counting forward, so subtracting it while counting backward would step two seats instead of one.

**No `turn` value can abort a routine anymore.** A fractional `turn` — easy to produce from a computed `${...}` — indexed the seat list with a fraction and crashed exactly like a negative one did; it is now truncated towards zero. Numeric strings (e.g. from an `INPUT` field) are accepted like they are in `SCORE` instead of silently meaning "1", `turn: "first"` is handled explicitly in the `position` cycle, and if a target seat still cannot be determined the operation falls back to the first active seat with a warning rather than throwing. Values that are meaningless for the chosen cycle now log a warning: `turn: 0` in the `position` cycle (positions are 1-based), a non-numeric `turn` in `position`, `first`/`last` outside `position` where they silently degrade to `turn: 1`, and any other non-number, which gets its own message rather than being told it would work with `position`.

Behaviour for positive integer `turn` values is unchanged and no property or schema changes.

**Editing UI.** Since `turn: 0` in the `position` cycle and `first`/`last` outside it are now defined-but-wrong values, the static validator reports both, so they show up in the editor's validation panel rather than only in a Debug log that has to be open before the routine runs. The visual routine editor knows the new semantics as well:

- the `turn` parameter offers `-1` as a one-click value wherever it counts seats along, and `first`/`last` only in the `position` cycle, where they actually mean a position
- a negative position reads as `Give the turn to the seat at position -2, counting back from the last`, and a single step reads as `by -1 seat` instead of `by -1 seats`
- the operation's help text spells out negative values and truncated fractions

The JSON editor offers `1`, `-1`, `first` and `last` for `turn`, and keeps offering them while the value itself is selected (its context regex was anchored so the buttons vanished the moment you double-clicked the value to replace it).

**Tests:** `tests/client/engine/turn.test.js` runs the operation against real widgets in the headless engine harness, once per legacy-mode combination: a negative `turn` of one and two steps, both `position` directions, a negative position over skipped seats, a fractional and a string `turn`, `position` with `turn: 0`, and a current seat with `skipTurn` set — stepping back from it, matching `turnCycle: "backward"` on it, and the unchanged forward step over it. A routine that cannot resolve a target rejects in the harness, so the tests also prove the operation no longer aborts the routine.

**Demo room.** The existing `Types - Seat` tutorial, variant "Turns", is the room the wiki links from `TURN`, and it now describes only the values that worked before. An amended draft is published to this PR's test server as [`Types - Seats 2.vtt`](https://agent.virtualtabletop.io/reports/pr/1534128684355682336/tutorials/Types%20-%20Seats%202.vtt) — open it in <https://test.virtualtabletop.io/PR-3098/pr-test-ai>. It adds a second button row (`turn: -1` on the forward and the backward cycle, `position: -1`, and a `position` a routine works out as `0`) plus a note explaining what a negative `turn` means. It is deliberately **not** part of this diff: `library/tutorials/` stays untouched until a maintainer accepts the draft.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3098/pr-test (or any other room on that server)

### DIFF
--- a/client/css/jsonedit.css
+++ b/client/css/jsonedit.css
@@ -523,6 +523,7 @@ body.jsonEdit #toolbar {
 
 /* problems are sentences, not JSON: wrap them instead of clipping at the panel edge */
 .jeLogProblems {
+  margin-left:20px;
   white-space: pre-wrap;
   word-break: break-word;
 }
--- a/client/css/jsonedit.css
+++ b/client/css/jsonedit.css
@@ -521,13 +521,6 @@ body.jsonEdit #toolbar {
   white-space: pre-wrap;
 }
 
-/* problems are sentences, not JSON: wrap them instead of clipping at the panel edge */
-.jeLogProblems {
-  margin-left:20px;
-  white-space: pre-wrap;
-  word-break: break-word;
-}
-
 [id=jeTopButtons] button {
   display: inline-block;
   text-align: center;
--- a/client/css/jsonedit.css
+++ b/client/css/jsonedit.css
@@ -521,6 +521,12 @@ body.jsonEdit #toolbar {
   white-space: pre-wrap;
 }
 
+/* problems are sentences, not JSON: wrap them instead of clipping at the panel edge */
+.jeLogProblems {
+  white-space: pre-wrap;
+  word-break: break-word;
+}
+
 [id=jeTopButtons] button {
   display: inline-block;
   text-align: center;
--- a/client/js/editor/controls/popup.js
+++ b/client/js/editor/controls/popup.js
@@ -2825,8 +2825,8 @@ function commonInfoTopic(topicName) {
 
       Parameters:
 
-      turn: integer, seat id, "first", or "last" - which seat becomes active, interpreted according to turnCycle.
-      turnCycle: forward/backward/random/position/seat - how the next player is chosen. forward/backward sort the active seats by index and move that many turn positions from the current seat; random picks a random active seat (turn is ignored); position uses turn "first"/"last" to pick the active seat with the smallest/largest index; seat uses turn as the id of the seat whose turn it becomes.
+      turn: integer, seat id, "first", or "last" - which seat becomes active, interpreted according to turnCycle. A negative value counts the other way round the cycle, so forward with turn -1 picks the same seat as backward with turn 1. Fractions are cut off.
+      turnCycle: forward/backward/random/position/seat - how the next player is chosen. forward/backward sort the active seats by index and move that many turn positions from the current seat; random picks a random active seat (turn is ignored); position takes turn as a position among the active seats, counting from the first one (1 is the first, and "first" says the same) or back from the last one for negative values (-1 is the last, and "last" says the same); seat uses turn as the id of the seat whose turn it becomes.
       source: all or collection - which seats are considered (defaults to all). Seats with skipTurn set to true are never chosen.
       collection: collection name - receives the seat whose turn it now is (defaults to "TURN"). Must be a named collection.
       </pre>
--- a/client/js/editor/controls/routine.js
+++ b/client/js/editor/controls/routine.js
@@ -1334,18 +1334,18 @@ const routineOperationMetadata = {
         template: 'Give the turn to a random seat' },
       { id: 'position', label: 'Give the turn to a seat by its position', fixed: [ 'turnCycle' ], match: v=>v('turnCycle') == 'position',
         apply: operation=>{ operation.turnCycle = 'position'; },
-        template: 'Give the turn to the seat at position {turn}' },
+        template: turnPositionWords },
       { id: 'seat', label: 'Give the turn to a specific seat', fixed: [ 'turnCycle' ], match: v=>v('turnCycle') == 'seat',
         apply: operation=>{ operation.turnCycle = 'seat'; },
         template: 'Give the turn to the seat {turn}' }
     ],
     clauses: [
-      { id: 'turn', label: 'n seats along', variants: [ 'forward', 'backward' ], template: v=>` by ${v('turn') == 1 ? '{turn} seat' : '{turn} seats'}` },
+      { id: 'turn', label: 'n seats along', variants: [ 'forward', 'backward' ], template: turnStepWords },
       { id: 'source', label: 'among some of the seats', template: ', among {source}' },
       { id: 'collection', label: 'remember the seat', template: ' and remember the seat as {collection}' }
     ],
     parameters: {
-      turn: { type: 'number', default: 1, special: [ 'first', 'last' ], textHint: 'id of a seat (used with turnCycle seat)', widgetType: 'seat' },
+      turn: { type: 'number', default: 1, special: turnSpecialValues, textHint: 'id of a seat (used with turnCycle seat)', widgetType: 'seat' },
       turnCycle: { type: 'enum', values: [ 'forward', 'backward', 'random', 'position', 'seat' ], default: 'forward' },
       source: { type: 'collection', default: 'all', display: { 'all': 'all seats', 'DEFAULT': 'the picked seats' }, widgetType: 'seat' },
       collection: { type: 'collection', default: 'TURN' }
@@ -1859,6 +1859,27 @@ function timerTime(v) {
   const parameter = v('seconds') ? 'seconds' : 'value';
   const seconds = parameter == 'seconds' ? v('seconds') : (typeof v('value') == 'number' ? v('value')/millisecondsPerSecond : null);
   return `{${parameter}} second${seconds == 1 ? '' : 's'}`;
+}
+
+// a negative turn steps the other way round the cycle, so the number of seats it
+// moves is what the sentence counts either way - only its sign says the direction
+function turnStepWords(v) {
+  return ` by ${widgetsCounted(v, 'turn', 'seat')}`;
+}
+
+// a position is counted from the first active seat, and a negative one back from
+// the last - which is what makes -1 the last seat rather than one before the first
+function turnPositionWords(v) {
+  return `Give the turn to the seat at position {turn}${Number(v('turn')) < 0 ? ', counting back from the last' : ''}`;
+}
+
+// first and last name a position, so they are only offered where turn means one;
+// a seat cycle takes a seat id, and everywhere turn counts seats a step back is
+// worth reaching for
+function turnSpecialValues(variant) {
+  if(variant == 'position')
+    return [ 'first', 'last', -1 ];
+  return variant == 'seat' ? [] : [ -1 ];
 }
 
 // the words a sentence starts with: everything before its first parameter. They
@@ -2526,7 +2547,7 @@ class RoutineOperationEditor {
       // a scaled parameter's own display words the stored value, while everything
       // in the popup is already in the unit it shows - so the numbers it offers
       // say what they are without being converted a second time
-      case 'number':     return new RoutineNumberPopup({ specialValues: spec.special, specialOnly: spec.specialOnly, scale: spec.scale, display: spec.scale ? null : spec.display, textHint: spec.textHint, widgetType: pickerOptions.widgetType });
+      case 'number':     return new RoutineNumberPopup({ specialValues: this.parameterSpecialValues(parameterNames[parameterNames.length-1]), specialOnly: spec.specialOnly, scale: spec.scale, display: spec.scale ? null : spec.display, textHint: spec.textHint, widgetType: pickerOptions.widgetType });
       case 'enum':       return new RoutineEnumPopup({ values: spec.values, display: spec.display });
       case 'property':   return new RoutinePropertyNamePopup();
       case 'widgets':    return new RoutineWidgetIDPopup(pickerOptions);
@@ -2769,7 +2790,8 @@ class RoutineOperationEditor {
   // "is there one" into a guess.
   dynamicProbeValues(name) {
     const spec = this.parameterSpec(name) || {};
-    const declared = [ ...(Array.isArray(spec.values) ? spec.values : []), ...(Array.isArray(spec.special) ? spec.special : []) ];
+    const special = this.parameterSpecialValues(name);
+    const declared = [ ...(Array.isArray(spec.values) ? spec.values : []), ...(Array.isArray(special) ? special : []) ];
     return declared.length ? declared : [ 0, 1, 'text', true, false, [] ];
   }
 
@@ -2964,6 +2986,14 @@ class RoutineOperationEditor {
 
   parameterSpec(name) {
     return this.metadata.parameters[name] || (name == 'skip' ? skipParameter : undefined);
+  }
+
+  // the ready-made values a parameter offers can depend on what the operation does:
+  // a TURN step counts seats where the cycle steps, while first and last name a
+  // position and only make sense there
+  parameterSpecialValues(name) {
+    const spec = this.parameterSpec(name) || {};
+    return typeof spec.special == 'function' ? spec.special(this.currentVariant().id) : spec.special;
   }
 
   registerChangeListener(listener) {
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -1604,7 +1604,7 @@ function jeAddCommands() {
   jeAddEnumCommands('^.*\\(SHIFT\\) ↦ keepOrder', [ true, false ]);
   jeAddEnumCommands('^.*\\(TIMER\\) ↦ mode', [ 'pause', 'start', 'toggle', 'set', 'dec', 'inc', 'reset']);
   jeAddEnumCommands('^.*\\(TIMER\\) ↦ value', [ 0, 'start', 'end', 'milliseconds']);
-  jeAddEnumCommands('^.*\\(TURN\\) ↦ turn$', [ 1, -1, 'first', 'last']);
+  jeAddEnumCommands('^.*\\(TURN\\) ↦ turn( ↦|$)', [ 1, -1, 'first', 'last']);
   jeAddEnumCommands('^.*\\(TURN\\) ↦ turnCycle', [ 'forward', 'backward', 'random', 'position', 'seat']);
   jeAddEnumCommands('^.*\\([A-Z]+\\) ↦ property', [ 'id', 'parent', 'type', 'rotation' ]);
 
--- a/client/js/jsonedit.js
+++ b/client/js/jsonedit.js
@@ -1604,6 +1604,7 @@ function jeAddCommands() {
   jeAddEnumCommands('^.*\\(SHIFT\\) ↦ keepOrder', [ true, false ]);
   jeAddEnumCommands('^.*\\(TIMER\\) ↦ mode', [ 'pause', 'start', 'toggle', 'set', 'dec', 'inc', 'reset']);
   jeAddEnumCommands('^.*\\(TIMER\\) ↦ value', [ 0, 'start', 'end', 'milliseconds']);
+  jeAddEnumCommands('^.*\\(TURN\\) ↦ turn$', [ 1, -1, 'first', 'last']);
   jeAddEnumCommands('^.*\\(TURN\\) ↦ turnCycle', [ 'forward', 'backward', 'random', 'position', 'seat']);
   jeAddEnumCommands('^.*\\([A-Z]+\\) ↦ property', [ 'id', 'parent', 'type', 'rotation' ]);
 
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -2701,15 +2701,26 @@ export class Widget extends StateManaged {
           if (unskipped.length === 0) {
             problems.push(`All seats in collection '${a.source}' have 'skipTurn' set to true. No turn change.`);
           } else {
+            // except for the 'seat' cycle, where turn is a seat id, turn indexes the seat
+            // list - so only whole numbers work. accept numeric strings and cut off
+            // fractions instead of ending up with no target at all
+            const turnNumber = Math.trunc(parseFloat(a.turn));
+            if (a.turnCycle != 'seat' && !isNaN(turnNumber) && turnNumber != parseFloat(a.turn))
+              problems.push(`Warning: turn ${a.turn} is not a whole number. Using ${turnNumber}.`);
+
             // identify the correct target seat
             if (a.turnCycle == 'position') {
-              if (a.turn == 'last') {
+              if (a.turn == 'first') {
+                target = unskipped[0];
+              } else if (a.turn == 'last') {
                 target = unskipped[unskipped.length - 1];
-              } else if (a.turn === 0) {
+              } else if (isNaN(turnNumber)) {
+                problems.push(`Warning: turn ${JSON.stringify(a.turn)} is not a valid position. Using the first active seat.`);
+              } else if (turnNumber == 0) {
                 problems.push(`Warning: turn 0 is not a valid position. Using the first active seat.`);
-              } else if (Number.isFinite(a.turn)) {
+              } else {
                 // positions are 1-based, negative positions count from the end: -1 is the last active seat
-                target = unskipped[wrapIndex(a.turn < 0 ? a.turn : a.turn - 1, unskipped.length)];
+                target = unskipped[wrapIndex(turnNumber < 0 ? turnNumber : turnNumber - 1, unskipped.length)];
               }
             } else if (a.turnCycle == 'seat') {
               // Selecting a specific seat so in this case skipTurn will be ignored
@@ -2719,11 +2730,20 @@ export class Widget extends StateManaged {
                 target = c[0];
               }
             } else {
-              const turn = Number.isFinite(a.turn) ? a.turn : 1;
+              if (isNaN(turnNumber))
+                problems.push(`Warning: turn ${JSON.stringify(a.turn)} is only valid with turnCycle position. Using 1.`);
+              // turn 0 leaves the turn where it is - unless the current seat is skipped,
+              // in which case it is not in unskipped and the turn moves on to the next seat
+              const turn = isNaN(turnNumber) ? 1 : turnNumber;
               // if the seat that currently has the turn is skipped, it is missing from
               // unskipped: index i is then i+1 steps forward but still i steps backward
               const offset = (turn > 0 && c[0] != unskipped[0] ? 1 : 0);
               target = unskipped[wrapIndex(turn - offset, unskipped.length)];
+            }
+
+            if (!target) {
+              problems.push(`Warning: no target seat could be determined. Using the first active seat.`);
+              target = unskipped[0];
             }
 
             // execute the change in turn properties and collect turn seats into output collection
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -2706,7 +2706,7 @@ export class Widget extends StateManaged {
               if (a.turn == 'last') {
                 target = unskipped[unskipped.length - 1];
               } else if (a.turn === 0) {
-                problems.push(`Warning: turn 0 is not a valid position - positions start at 1 and negative positions count back from the last seat. Using the first seat.`);
+                problems.push(`Warning: turn 0 is not a valid position. Using the first active seat.`);
               } else if (Number.isFinite(a.turn)) {
                 // positions are 1-based, negative positions count from the end: -1 is the last active seat
                 target = unskipped[wrapIndex(a.turn < 0 ? a.turn : a.turn - 1, unskipped.length)];
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -2695,12 +2695,16 @@ export class Widget extends StateManaged {
           if (unskipped.length === 0) {
             problems.push(`All seats in collection '${a.source}' have 'skipTurn' set to true. No turn change.`);
           } else {
+            // modulo that always returns a non-negative index so negative turn values wrap
+            const wrap = (i, length) => ((i % length) + length) % length;
+
             // identify the correct target seat
             if (a.turnCycle == 'position') {
               if (a.turn == 'last') {
                 target = unskipped[unskipped.length - 1];
               } else if (Number.isFinite(a.turn)) {
-                target = unskipped[(a.turn - 1) % unskipped.length];
+                // negative positions count from the end: -1 is the last active seat
+                target = unskipped[a.turn < 0 ? wrap(a.turn, unskipped.length) : wrap(a.turn - 1, unskipped.length)];
               }
             } else if (a.turnCycle == 'seat') {
               // Selecting a specific seat so in this case skipTurn will be ignored
@@ -2712,7 +2716,7 @@ export class Widget extends StateManaged {
             } else {
               const turn = Number.isFinite(a.turn) ? a.turn : 1;
               const offset = (c[0] == unskipped[0] ? 0 : 1);
-              target = unskipped[(turn - offset) % unskipped.length];
+              target = unskipped[wrap(turn - offset, unskipped.length)];
             }
 
             // execute the change in turn properties and collect turn seats into output collection
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -2730,8 +2730,10 @@ export class Widget extends StateManaged {
                 target = c[0];
               }
             } else {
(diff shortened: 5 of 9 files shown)


VTT-AI-Agent (commented):
## TL;DR

The non-negative modulo is the right fix and the forward/backward case works for the common setup, but the `offset` path is off by one for negative `turn`: when the seat that currently has the turn has `skipTurn: true`, `turn: -1` lands two seats back instead of one, which also breaks the invariant the PR description states (`forward`/`-1` == `backward`/`+1`). No test covers `skipTurn`, so it passes unnoticed.

## Findings

### 1. `client/js/widgets/widget.js:2276` — negative `turn` is off by one when the current seat is skipped

`offset` is 1 exactly when the current-turn seat (`c[0]`) is filtered out of `unskipped`. In that case `unskipped` is conceptually "steps +1, +2, … forward from the current seat", so index `i` ↔ step `i + 1` — which is why the forward path subtracts the offset. Going *backward*, though, step `-n` is index `len - n`, i.e. `wrap(turn, len)` with **no** offset subtraction. Subtracting it as well overshoots by one seat.

Concrete case — 4 seats `s1..s4`, `s1` has the turn and `skipTurn: true`, `turnCycle: 'forward'`, `turn: -1`:

- `c` = `[s1, s2, s3, s4]`, `unskipped` = `[s2, s3, s4]` (len 3), `offset` = 1
- code: `wrap(-1 - 1, 3)` = 1 → **`s3`**
- expected (one step backward from `s1`): **`s4`**

Cross-check against the invariant claimed in the PR description: same room with `turnCycle: 'backward'`, `turn: 1` → `c` sorted desc and rotated = `[s1, s4, s3, s2]`, `unskipped` = `[s4, s3, s2]`, `offset` = 1 → `wrap(0, 3)` = 0 → `s4`. So `forward/-1` and `backward/+1` disagree in exactly this case.

Suggested shape:

```js
target = unskipped[wrap(turn - (turn > 0 ? offset : 0), unskipped.length)];
```

(positive `turn` keeps today's behaviour bit-for-bit; negative `turn` counts back from the current seat correctly whether or not it's skipped).

Please add a test for it — a seat with `skipTurn: true` holding the turn is the whole reason `offset` exists, and none of the four new tests exercise it.

### 2. `client/js/widgets/widget.js:2264` — `turnCycle: 'position'` with `turn: 0`

`a.turn < 0 ? wrap(a.turn, len) : wrap(a.turn - 1, len)` sends `turn: 0` to `wrap(-1, len)` = the last seat, i.e. `0` and `-1` silently mean the same thing. It used to throw, so nothing regresses, but "position 0" resolving to the *last* seat is a surprising thing to make legal by accident. Consider either documenting it or `problems.push(...)`-ing a warning for `turn == 0` in position mode, the way the operation already warns for an unknown `turnCycle`.

### 3. Minor

- The new tests only assert the end state via `expectEventually`, and `seatsWithTurn` starts by asserting `['seat1']` before the click — good. But `clickTurn` rebuilds the whole room for every test with all four buttons present; that's fine, just noting the four tests are effectively four full room setups if runtime matters.
- `wrap` is re-created on every `TURN` execution inside the operation loop; a module-level helper next to the other small helpers in this file would read the same and avoid the reallocation. Cosmetic only.

Otherwise the approach is right: the `((i % n) + n) % n` form is the correct fix for JS's sign-preserving `%`, positive-`turn` behaviour is provably unchanged, and no property/schema changed so the validator/`jeCommands` really are untouched.

I did not add the `beta` label: this PR changes the semantics of the `TURN` routine operation in `client/js/widgets/widget.js`, which is a genuine engine change (not incidental), so it is excluded from beta auto-deployment by design.

VTT-AI-Agent:
Thanks — finding 1 is correct, fixed in 2e1bfa8.

- **Off-by-one for negative `turn` when the current seat is skipped:** the `offset` only compensates for the fact that index `i` in `unskipped` is `i+1` steps *forward* when the current seat is filtered out; counting backward, index `i` is already `i` steps back. It is now applied for positive `turn` values only (`turn > 0 && c[0] != unskipped[0]`), which keeps positive behaviour bit-for-bit and makes `forward`/`-1` and `backward`/`+1` agree again. Three TestCafe tests now cover a current seat with `skipTurn: true`: stepping back one seat from it, the `backward`/`+1` equivalence, and the unchanged forward step over it. The first of them fails without the fix.
- **`turnCycle: 'position'` with `turn: 0`:** agreed that resolving to the last seat by accident is surprising. It now pushes a warning (`turn 0 is not a valid position - positions start at 1 and negative positions count back from the last seat`) and leaves the turn on the first seat, in the same style as the existing unknown-`turnCycle` warning.
- **Minor:** `wrap` moved out to a module-level `wrapIndex()` next to the other small helpers at the top of `widget.js`. The per-test room rebuild I left as is — it keeps each test independent, and the suite runs in ~80s.

`npm test` passes (199 tests) and the full `tests/testcafe/routines.js` suite passes locally (11 tests).

No `beta` label added, agreed — this changes `TURN` semantics in the engine.

VTT-AI-Agent (commented):
## TL;DR

This is a **UI/UX review, not a code review** — a separate reviewer covers correctness. The diff is engine-only (no HTML/CSS/template changes), and I confirmed it leaves every overlay untouched. It does however add one new *user-facing string*, and that string is the whole UI surface of this PR — and it does not fit. At 1280×800 the new `turn 0` warning is cut off mid-word inside the Debug panel (`.jeLogProblems` is `white-space: pre` with no horizontal scroll), and on a phone the reader gets `"Warni`. **Most impactful fix: shorten the warning to one short sentence like every other warning in `widget.js`, and move the explanation to the wiki.**

---

### What I actually exercised

Checked out the branch, ran the server locally and drove it with Playwright at **1920×1080, 1280×800, 820×1180 and 390×844**. Per viewport I opened the room, entered edit mode, opened the **Debug** module, ran a `TURN` routine with `turnCycle: "position", turn: 0` and expanded the log down to the new warning; I also swept the **Players**, **States** and **About** overlays at all four sizes. Overlays I could not reach in this pass: the add-widget / library-browser / share / state-save / symbol-picker / deck-editor dialogs — but since the diff contains zero CSS or markup, there is no plausible path for it to affect them, and the ones I did check render identically to `main`.

---

## Big layout improvements

**1. The new warning overflows the Debug log and is silently truncated at every viewport.**
`client/js/widgets/widget.js:2266` pushes a 137-character message. `.jeLogProblems` in `client/css/jsonedit.css:653` is `white-space: pre`, and neither `#jeLog` nor `#jeLogWrapper` sets `overflow-x`, so the line neither wraps nor scrolls — it is simply clipped at the panel edge. Measured `scrollWidth` / `clientWidth` of the problem line:

| viewport | rendered width needed | available | visible text ends at |
|---|---|---|---|
| 1920×1080 | 852px | 732px | `…and negative positions c` |
| 1280×800 | 852px | 504px | `…and negative posit` |
| 820×1180 | 852px | 684px | `…and negative positi` |
| 390×844 | 852px | 254px | `"Warni` |

1280×800 — the sentence stops mid-word, the actual advice ("negative positions count back from the last seat") is unreachable:
Debug log at 1280x800, warning truncated

390×844 — effectively no message at all:
Debug log at 390x844

Side-by-side with the pre-existing sibling warning on the line above it — `Warning: turnCycle sideways interpreted as forward.` fits exactly; the new one runs off the edge:
Two warnings compared

Concretely, I'd match the house style of the ten other `problems.push()` warnings in this file (all one short clause, all ending in a period):

```js
problems.push(`Warning: turn 0 is not a valid position - using the first seat.`);
```

and put "positions are 1-based, negative positions count back from the last seat" in the wiki page for `TURN`, where there is room for it. If the long form is genuinely needed, the CSS fix belongs in this PR too — `white-space: pre-wrap` on `.jeLogProblems` (it already has `word-break: break-all` on the neighbouring `.jeLogDetails`, so wrapping is consistent with the panel's conventions) — but the shorter string is the smaller, safer change and helps every existing warning-length case as well.

**2. The green "No validation problems found!" banner sits directly above the red warning and contradicts it.**
Visible in the 1280×800 screenshot: the JSON editor's static validator (`validator/validate_gamefile.js:809`) accepts any integer for `turn`, so `turn: 0` passes validation while the runtime logs a warning about it. A first-time game author reading top-to-bottom is told their routine is clean and then, three expander clicks deeper, told it isn't. Since this PR is what makes `turn: 0` a defined-but-wrong value, it would be a good moment to teach the validator about it: tighten the `position` case so `turn: 0` is reported inline in the validation panel, where the author will actually see it, rather than only in a Debug log they have to have opened *before* running the routine.

- `client/js/widgets/widget.js:2266` — the message uses a spaced hyphen as a sentence break ("`position - positions start at 1`"). Every other warning in this file is a single clause; if the long form survives, an em dash or a second sentence would read better in the monospace log than a hyphen sandwiched between two lowercase words.
- The warning claims "Using the first seat", which is accurate but ambiguous in the log: the summary line right above it reads `Changed turn of seats to 1`, so the reader sees two different notions of "first" (position 1 vs. seat index 1) one line apart. Saying "the first active seat" would match the wording already used in the code comment on line 2269.
- Warnings are nested under a generic `Problems` expander with no count or preview, so a collapsed operation shows only a red arrow. Not introduced here, but this PR adds a warning that a user is *expected* to hit, which makes it more noticeable — a `Problems (1)` label would cost nothing.

- "position", "seat", "turn cycle" and "skipTurn" are all engine jargon. The message is the only place a newcomer meets the 1-based rule, so the truncation above is not just cosmetic — it removes the one explanatory sentence they get. Fixing #1 is what makes this section moot.
- Nothing in the editor UI signals that a negative `turn` is now meaningful. `jeAddRoutineOperationCommands('TURN', { turn: 1, turnCycle: 'forward', … })` in `client/js/jsonedit.js:1396` still inserts `turn: 1`, and the `turnCycle` enum command on line 1495 offers `forward|backward|random|position|seat` with no hint that `-1` is now a supported alternative to `backward`. No new property is added, so no `jeCommands` entry is *required*, but a first-time author has no in-app way to discover the feature this PR ships. Worth a wiki update in the same PR, and/or making the existing `backward` enum entry's tooltip mention the equivalence.
- Nit on discoverability: the warning only exists at runtime, in the Debug module. If a user writes `turn: 0` and never opens Debug, the turn silently lands on the first seat with no feedback anywhere. That's the strongest argument for the validator change in #2.

---

Everything outside the Debug log is unchanged, and the Players / States / About overlays render correctly at all four sizes I tested — no regressions there.

VTT-AI-Agent:
Thanks for the UI/UX pass — all three points addressed in ff902d57.

**1. Truncated warning (big).** Shortened to the house style of the sibling warnings and, since the panel clips *any* long problem message, `.jeLogProblems` now wraps instead of being cut off (`white-space: pre-wrap; word-break: break-word`) — every other long warning in the log benefits too. The message is now `Warning: turn 0 is not a valid position. Using the first active seat.`, i.e. two short sentences like `Mode ${a.mode} is unsupported. Using 'respect' mode.` a few lines above it, and it uses "first active seat" as you suggested so it does not read as "seat 1".

Re-measured the problem line after the change: `scrollWidth == clientWidth` at both viewports (504/504 at 1280×800, 254/254 at 390×844 where it wraps onto two lines), so nothing is clipped anymore.

- 1280×800: https://agent.virtualtabletop.io/reports/pr/1534128684355682336/turn0-warning-after-1280x800.png
- 390×844: https://agent.virtualtabletop.io/reports/pr/1534128684355682336/turn0-warning-after-390x844.png

**2. Green banner contradicting the warning (big).** `validator/validate_gamefile.js` now reports `turn: 0` with `turnCycle: "position"` from `customRoutineChecks()`, so the validation panel shows `TURN uses position 0 - positions start at 1, negative positions count back from the last seat` instead of "No validation problems found!". That is also the answer to your discoverability nit: the author gets told even if they never open the Debug log, because the Debug module validates the room on open and on every delta. Only literal `turn: 0` in `position` mode is flagged — `turn: 0` in `forward`/`backward` (a deliberate "keep the turn where it is") stays clean.

**3. Editor gives no hint that negative `turn` is meaningful.** Added `jeAddEnumCommands('^.*\\(TURN\\) ↦ turn$', [ 1, -1, 'first', 'last' ])`, mirroring the existing `TIMER ↦ value` enum, so selecting the `turn` value offers `-1` (plus the `first`/`last` position keywords, which were already supported by the runtime and the validator but had no button either). The `$` anchor keeps those buttons off `turnCycle`; verified in the browser that `↦ turn` shows `-1 / 1 / first / last` and `↦ turnCycle` still shows only the five cycle values. I left the `jeCommands` TURN default at `turn: 1` — it is the sensible default for a fresh operation — and left `Problems` without a count, since that expander is not something this PR introduces.

Verified with `npm test` (199 passed), `tests/testcafe/routines.js` (11 passed) and `tests/testcafe/editor.js` (23 passed), plus the Playwright run above at 1280×800 and 390×844.

VTT-AI-Agent (commented):
## TL;DR

The negative-index crash and the skipped-current-seat off-by-one are both fixed correctly, and I found no remaining actionable issue in the current revision. The focused TestCafe coverage exercises the affected `TURN` paths through routine completion, and the full CI suite is green.

## Findings

No further findings. I re-checked the non-negative modulo behavior for forward/backward cycles, skipped current seats, and negative absolute positions; the positive-turn path remains unchanged, while `position` value `0` now has consistent runtime and validator feedback. The editor suggestions and wrapped Debug-problem text also address the discoverability and clipping concerns from the preceding review without changing routine semantics beyond the intended fix.

I did not add the `beta` label because `client/js/widgets/widget.js` changes the semantics of the `TURN` routine operation, which is a genuine engine behavior change and therefore engine-ineligible.

VTT-AI-Agent (commented):
## TL;DR

The fix is correct and well targeted: `wrapIndex` plus the `turn > 0` guard on `offset` make negative `turn` values wrap in both directions, and positive-`turn` behaviour is provably unchanged. I ran the suite locally — `npx testcafe chromium:headless tests/testcafe/routines.js` (11 passed, including all 7 new tests) and `npm test` (199 passed). Main remaining gap: `Number.isFinite` still lets a *fractional* `turn` (e.g. `1.5` from a computed `${}`) produce a fractional array index, so `target` is `undefined` and the routine still dies with a `TypeError` — the exact failure class this PR is fixing.

## Findings

### 1. Fractional `turn` still aborts the routine (`client/js/widgets/widget.js:2270`, `:2281`)

`Number.isFinite` accepts non-integers, and `wrapIndex` preserves the fraction:

- position: `turn: 2.5` → `wrapIndex(1.5, 4)` → `1.5` → `unskipped[1.5]` → `undefined`
- forward/backward: `turn: 1.5` → `wrapIndex(1.5, 4)` → `undefined`

`target` is then `undefined` and the seat loop right below (`w.get('index') == target.get('index')`) throws a `TypeError`. There is no `try`/`catch` around the TURN body, so the whole routine aborts — the same symptom the PR describes for negative values. This is reachable in practice because `turn` is usually written as `${...}`, and any division/average produces a float.

Cheapest fix is a defensive guard that also covers the string case below:

```js
target = unskipped[wrapIndex(Math.trunc(a.turn < 0 ? a.turn : a.turn - 1), unskipped.length)];
```

plus, before the seat loop, `if(!target) { problems.push(...); target = unskipped[0]; }` so no future arithmetic slip can crash a routine here again.

Related, same area: a *string* number (`turn: "2"`, easily produced by an `INPUT` field into a variable) fails `Number.isFinite` and silently resolves to the **first** seat in the `position` cycle rather than the second. That is pre-existing, but this PR is exactly the hardening pass where it would fit.

### 2. `turn: 0` is only diagnosed for `turnCycle: 'position'`

The new warning and the new validator rule (`validator/validate_gamefile.js:831`) only fire for `position`. In the forward/backward branch the new `offset` guard makes `turn: 0` mean "keep the turn where it is" — except when the current seat has `skipTurn`, where `unskipped[0]` is the *next* seat, so `turn: 0` silently advances one seat. That is a defensible outcome, but it is undocumented and differs from the intuitive reading of `0`. Worth either a comment at `widget.js:2281` or the same warning treatment.

### 3. The new JSON-editor enum offers `first`/`last` in every TURN context (`client/js/jsonedit.js:1495`)

The context `^.*\(TURN\) ↦ turn$` matches regardless of `turnCycle`, but:

- `'first'` is not handled anywhere in `widget.js` — it only works by falling through to the default `target = unskipped[0]`, and only in the `position` cycle.
- In `forward`/`backward`/`random`, `turn: 'first'` **and** `turn: 'last'` both silently degrade to `turn: 1` (`Number.isFinite` is false → default `1`). Offering them as one-click values makes that trap more discoverable, not less.

Since the PR already touches this code path, consider handling `'first'` explicitly next to `'last'` in the `position` branch (self-documenting), and/or pushing a problem when `first`/`last` is used with a non-`position` cycle.

### 4. Unrelated CSS change (`client/css/jsonedit.css:658-663`)

The `.jeLogProblems` wrap override is scope creep against the "one PR = one concern" convention, and it duplicates a selector that appears four lines above — `white-space: pre` is set for `.jeLogProblems` at line 653 and then immediately overridden. Dropping `.jeLogProblems` from the shared rule at line 653 and giving it its own rule would express the same thing without a self-override. It also changes how *all* routine problems render, not just the new one.

### 5. Test coverage

The seven new tests are well designed — the witness widget proving the routine ran to completion is the right assertion, and the skipped-current-seat cases cover the `offset` change in both directions. Not covered: `turnCycle: 'position'` with `turn: 0` (the new warning path, i.e. that it lands on the first seat instead of crashing), and a negative `position` turn combined with `skipTurn` seats. Both are cheap to add to `turnButtons`.

### Docs

The negative-position semantics (`-1` = last active seat, negative `turn` reverses the cycle) exist only in this diff and the editor enum. The wiki's Functions page describes `turn` as "modulo the length of the list"; it should be updated to spell out the negative case, since that is where game authors will look.

I added the `beta` label: no — this PR changes routine-operation semantics for `TURN` in `client/js/widgets/widget.js`, which is a genuine engine change and therefore excluded from beta auto-deployment by design.

VTT-AI-Agent:
Thanks — findings 1–5 addressed in 787d3920.

**1. Fractional `turn` still aborts the routine.** Fixed, and generalised: `turn` is now normalised once with `Math.trunc(parseFloat(a.turn))` before the cycle branches, so a fraction is cut off towards zero instead of producing a fractional index (`turn: 2.5` in `position` → position 2, `turn: -1.5` in `forward` → one seat back), with a warning that says what was used. That also picks up your related point: a numeric string (`turn: "2"`) is now taken as the number instead of silently meaning "first seat"/"1", the same way `SCORE` treats its `value`. And as you suggested there is now a `if(!target)` fallback to the first active seat with a warning right before the seat loop, so no future arithmetic slip in this operation can throw again. Two new tests cover the fraction and the string.

**2. `turn: 0` outside `position`.** In `forward`/`backward` it genuinely means "keep the turn where it is", which I did not want to warn about — a computed `${...}` that evaluates to 0 is a legitimate no-op. Documented instead, at the place you pointed at: the branch now carries a comment saying that `turn: 0` keeps the turn where it is *unless* the current seat is skipped, in which case it is not in `unskipped` and the turn moves on to the next seat that is not skipped (it cannot stay on a skipped seat).

**3. The new editor enum offering `first`/`last` everywhere.** Both halves fixed rather than the enum removed: `'first'` is now handled explicitly next to `'last'` in the `position` branch (it worked only by falling through to the default), and using `first`/`last` with any other cycle now logs `Warning: turn "last" is only valid with turnCycle position. Using 1.` and is reported by the static validator as well, so the trap is diagnosed instead of hidden. The validator check skips a `turnCycle` written as `${...}` so it cannot false-positive on a computed cycle.

**4. The CSS.** Kept, but restructured as you suggested — `.jeLogProblems` is out of the shared `white-space: pre` rule at line 653 and has its own rule now, so there is no self-override. I did not drop the wrap itself: review 2 measured that the panel clips *any* long problem message (the warning was cut off mid-word at 1280×800 and reduced to `"Warni` at 390×844), and this PR adds four more messages to that list, so the two-line fix belongs with them.

**5. Test coverage.** Added `position` with `turn: 0` (lands on the first seat, routine completes) and a negative `position` turn over a collection that contains a `skipTurn` seat (`-1` is the last *unskipped* seat, not the last seat), plus the fraction and string cases from finding 1. `tests/testcafe/routines.js` is at 15 tests, all passing; the button grid in `turnRoom` was re-laid-out into columns so the extra buttons stay on the board.

**Docs.** The negative-`turn` and negative-position semantics still need a wiki update on the Functions page — I have no write access to the wiki from here, so that has to be done by a maintainer; the PR description spells out the semantics in the meantime.

Verified: `npm test` 199 passed, `tests/testcafe/routines.js` 15 passed, `tests/testcafe/editor.js` 23 passed, and `node validator/validate_gamefile_node.js` on a scratch game file reports position 0 and `first`/`last` with a wrong cycle while leaving the valid combinations (`position`/`last`, `position`/`-2`, `forward`/`-1`, `${...}` cycle) clean.

VTT-AI-Agent (commented):
## TL;DR

The engine fix is correct: I re-derived the new index arithmetic against a step-by-step reference model for every 4-seat configuration (520 forward/backward cases plus the `position` edge values) and found no mismatch; `forward`/`-n` and `backward`/`+n` agree for every `n ≠ 0`, and positive-integer behaviour is unchanged. The PR cannot merge as it stands, though: it now **conflicts with `main` in `tests/testcafe/routines.js`** (main moved `expectEventually` into `test-util.js`), and after the rebase the **visual routine editor that landed on `main` after this PR's base** still describes and offers only the old `turn` semantics. There is also no demo room on the PR's `pr-test-ai` test server.

## Findings

### 1. Merge conflict with `main` — `tests/testcafe/routines.js:82` (blocking)

`main` (a1253870, #3028) exported `expectEventually(t, get, expected, message, timeout=1000)` from `tests/testcafe/test-util.js` and deleted the local copy in `routines.js`; this branch still carries the local copy at line 82 and inserts `turnButtons`/`turnRoom` right above it, so git cannot merge the hunk (`gh pr view` reports `CONFLICTING`). When rebasing:

- delete the local `expectEventually` and add it to the import on line 3 (`import { expectEventually, getState, ... } from './test-util.js'`). Keeping both is not an option: an import binding plus a top-level function declaration of the same name is a `SyntaxError: Identifier 'expectEventually' has already been declared`, which would take down the whole fixture rather than just the new tests. The new tests' `(t, get, expected)` calls are compatible with the exported signature.
- note that `main`'s `routines.js` also replaced `SWAPHANDS` with `SHIFT` (`clickSwap`, `removeOnEnterRoom(operation)`); nothing in the TURN tests depends on that, but the surrounding helpers will look different.

The green checks on the PR ran on 787d392 on 2026-08-04, before `main` diverged; GitHub cannot build a merge commit for a conflicting PR, so nothing has been exercised against current `main` yet.

### 2. The visual routine editor on `main` does not know the new semantics — UI-support rule

`main` now has `client/js/editor/controls/routine.js` and `client/js/editor/controls/popup.js` (neither existed at this PR's base), and they are where game authors edit TURN today:

- **Help text is now wrong.** `popup.js` (TURN topic, ~line 2821-2831 on `main`) says `turn: integer, seat id, "first", or "last"` and `position uses turn "first"/"last"`. After this PR it has to say that a negative `turn` steps the other way round the cycle, that `position` takes a 1-based integer with negative values counting from the end (`-1` = last active seat), and that fractions are truncated. Please update it in the rebase; the same text is what the wiki's TURN section needs (the PR body already flags the wiki).
- **Negative values are settable but not discoverable.** The `turn` parameter is `{ type: 'number', default: 1, special: [ 'first', 'last' ], … }` (`routine.js` ~line 1348 on `main`). The number popup offers the specials, a keypad `0…10` and a free-text line; typing `-1` there does yield the number `-1` (`parseValueText` JSON-parses it), so it works, but nothing hints at it. Concrete suggestion: add `-1` to `special` so it appears as a button next to `first`/`last`, and give the forward/backward `turn` clause a negative-aware wording (today `turn: -1` renders as "Pass the turn on by -1 seats"; "back by 1 seat" would match the sentence style used elsewhere in that file). The `first`/`last` specials are also offered in the forward/backward variants where this PR now warns — the same trap the `$` anchor avoids in the JSON editor.

### 3. Standing maintainer rule: no demo room on `pr-test-ai`

The PR body links only the generic `PR-3098/pr-test` room. Negative `turn`, negative positions and the new `1 / -1 / first / last` editor buttons are new author-facing values, so a tutorial-quality demo room should be published as a `.vtt` to the PR's own `pr-test-ai` room (via `../publish/`), not committed. The natural vehicle is the existing TURN tutorial — `Types - Seat`, variant "Turns" (`library/tutorials/Types - Seats/2.json`, the page the wiki's TURN section links to): its buttons cover `forward`, `backward`, `random`, `position first/last/2` and `seat`, but nothing negative. An amended draft copy (published under `../publish/tutorials/`) adding a "turn: -1" button next to "backward", a "position: -1" button next to "position: last", and one that shows the new `position 0` validation message would cover the change with the existing layout and `_meta`. The diff correctly contains no `library/` changes — please keep it that way unless a human accepts the draft.

### 4. Minor

- `client/js/widgets/widget.js:2291` — the warning `turn "abc" is only valid with turnCycle position. Using 1.` is only true for `first`/`last`; for any other non-number (`"abc"`, `true`, `null`) it is misleading, since those are not valid with `position` either. Split it: `first`/`last` keep this message, everything else gets `turn … is not a number. Using 1.`.
- `client/js/jsonedit.js:1495` — `jeGetContext` appends the selected text as a trailing key (`if(select) keys.push(`"${select}"`)`), so with the value selected (e.g. the `1` double-clicked to replace it) the context is `(TURN) ↦ turn ↦ "1"` and the new buttons vanish, while every other enum (`turnCycle`, `TIMER ↦ value`) keeps showing. `'^.*\\(TURN\\) ↦ turn( ↦|$)'` keeps them in both states and still excludes `turnCycle`.
- Runtime vs validator: the runtime now takes numeric strings (`turn: "2"`), while `validator/validate_gamefile.js:809` still only accepts integer literals, `first` and `last`, so a literal `"2"` remains a validation problem. That is defensible (the validator is the stricter tool and computed `${…}` values are bypassed), just be aware the PR body's "accepted like in SCORE" only holds at runtime. `parseFloat` also accepts trailing junk (`"2px"` → 2) silently, same as `SCORE`.
- Test cost: `main` now has a headless engine harness (`tests/client/engine/harness.js`, #3063) that runs routines against real widgets in jsdom. TURN only reads `type`/`player`/`index`/`skipTurn` and writes `turn`, all base-class properties, so the eleven browser tests could become a millisecond jest table there — optional, but worth it if the suite's runtime matters.

What I checked: the full diff and the merged tree, the `main` history since the PR base (61 commits), the wiki's TURN section, the existing seats tutorial, and a script replicating the PR's target selection against an independent step-counting model. I did not run TestCafe locally (this clone has no `node_modules`); CI on the head commit is green but predates the conflicting `main` change.

I did not add the `beta` label: the change to `client/js/widgets/widget.js` alters the semantics of the TURN routine operation for existing games (numeric-string, negative and fractional `turn` values now resolve to a seat where they previously meant 1, the first seat, or an aborted routine), which is a genuine engine change and therefore excluded from beta auto-deployment, and the PR is currently unmergeable anyway.

VTT-AI-Agent:
Thanks — all five points are addressed. What changed since [pullrequestreview-5119302447](https://github.com/ArnoldSmith86/virtualtabletop/pull/3098#pullrequestreview-5119302447):

**1. Merge conflict in `tests/testcafe/routines.js` (blocking) — resolved, and the file is no longer touched at all.** The branch is rebased onto current `main`. Rather than keeping the local `expectEventually` or re-importing it, I took your last "minor" point and moved the eleven TURN cases into the headless engine harness as `tests/client/engine/turn.test.js`. They assert the same seats, run against real widgets through `evaluateRoutine`, cover every legacy-mode combination instead of one, and take milliseconds instead of ~4 minutes of browser time. A routine that cannot resolve a target rejects in the harness, so the "the routine still ran to the end" property the witness widget used to prove is still asserted. `tests/testcafe/routines.js` is now byte-identical to `main`, so the conflict is gone for good. `npm test`: 57 suites, 1732 tests, all green.

**2. UI-support rule — the visual routine editor knows the new semantics.** All verified in a browser against the running app:

- `turn` now offers its ready-made values per variant (`client/js/editor/controls/routine.js`): `-1` wherever the cycle counts seats along, and `first`/`last` only in `position`, where they actually mean a position. Confirmed in the popup: the forward variant offers `-1, 0…10`, the position variant `first, last, -1, 0…10`. That also closes the trap you noted — `first`/`last` are no longer one click away in the cycles where the operation (and now the validator) says they are wrong.
- Wording: `turn: -1` in `position` reads `Give the turn to the seat at position -1, counting back from the last`, and the step clause pluralises on the magnitude, so `turn: -1` reads `by -1 seat` rather than `by -1 seats`. I kept the sign visible instead of rewording `Pass the turn on` to `Pass the turn back`: the variant lead is what the drop-down at the start of the sentence offers, and flipping it for a negative value would make the `forward` and `backward` entries swap places in that list.
- The TURN help text in `client/js/editor/controls/popup.js` now spells out that a negative `turn` counts the other way round the cycle, that `position` takes a 1-based position counting back from the last seat for negative values (with `first`/`last` as the same thing as `1`/`-1`), and that fractions are cut off.

**3. Demo room.** An amended draft of the `Types - Seat` "Turns" variant is published as [`Types - Seats 2.vtt`](https://agent.virtualtabletop.io/reports/pr/1534128684355682336/tutorials/Types%20-%20Seats%202.vtt) and is open in <https://test.virtualtabletop.io/PR-3098/pr-test-ai>. It adds a second button row in the free band below the existing one — `Forward turn: -1`, `Backward turn: -1`, `Position: -1` and a `position` a routine works out — plus a note box explaining what a negative `turn` means. I walked it end to end (sit in all four seats, click each button, read the seat that ends up with the turn: `seat4`, `seat2`, `seat4`, `seat1` — as documented), and `node validator/validate_gamefile_node.js` is clean on it. One deviation from your sketch: the `position 0` button computes the `0` in a `VAR` rather than writing it into the JSON, because a literal `0` would trip the very validator rule this PR adds and a tutorial that reports a validation problem is not a good tutorial. The computed form is also the realistic way to hit it. `library/tutorials/` is untouched — the draft moves there only if a maintainer accepts it.

**4. Misleading warning for non-numeric `turn`.** Split, as suggested: `first`/`last` keep `turn "first" is only valid with turnCycle position. Using 1.`, everything else now gets `turn "abc" is not a number. Using 1.`

**5. JSON-editor `$` anchor.** Changed to `'^.*\\(TURN\\) ↦ turn( ↦|$)'`. Verified in the browser: with the caret behind the value the buttons show, and with the `-1` double-clicked they still show — previously the context became `(TURN) ↦ turn ↦ "-1"` and they disappeared. `turnCycle` still does not match.

On the numeric-string note: agreed, the validator stays the stricter tool and a literal `"2"` remains a validation problem; the runtime leniency is for values that arrive through `${…}`.

I did not add the `beta` label.

VTT-AI-Agent:
Follow-up on point 5: `main` had meanwhile added the `.jeLogProblems` wrapping rule itself, so the copy this branch carried was a duplicated selector one block further down. Dropped it — `client/css/jsonedit.css` is now untouched by this PR, and the Debug log still wraps.

PR #3099, opened 2026-08-04
Audio tutorial: show how to pick a bundled sound from the JSON editor's sound picker

VTT-AI-Agent:
Teaches readers of the **Functions - AUDIO** tutorial that they no longer have to hunt down a sound file: both of its variants now point at the *sound picker* that #3047 adds to the JSON editor, so a game creator can browse, preview and insert one of the sound effects bundled with VTT.

Depends on #3047 — the `pick a sound from the sound picker` button and the `/i/audio/...` sounds it inserts only exist once that PR is merged, so this one should land after it.

## What it teaches

- **`clickSound` variant** — a new speaker widget sits opposite the existing bell with an *empty* `clickSound`, so it deliberately stays silent. Its caption walks the reader through filling it in: open the speaker in the JSON Editor, put the cursor on the `clickSound` line, use the *pick a sound from the sound picker* button in the upper right corner, preview with ▶, click a sound to insert its `/i/audio/...` path, leave edit mode and click the speaker to hear it. The reader ends up with a widget they made make a noise themselves.
- **`AUDIO` variant** — a paragraph under the intro explains that `source` takes any sound URL but that the same picker is available on the `source` line of an `AUDIO` function, and points at the existing *Deal* button as the thing to try it on.

## Structure

Both variants of this tutorial are single-screen demo boards: an interactive widget with a plain-language caption underneath. The additions follow that exactly — one caption per variant, `font-size: 20px`, black, mirroring the column layout that is already there — and they were modelled on how **Properties - Icon** (variant *CSS Demo*) describes reaching the symbol picker ("open one in the JSON Editor. Place the cursor … and click the … button that appears in the upper right corner"), so the wording for the two pickers matches.

The diff is additions only: two new widgets in `0.json`, one in `1.json`. No existing widget, id, position or text was changed, and no other tutorial was touched.

| clickSound variant | AUDIO variant |
| --- | --- |
| clickSound | AUDIO |

## Assets

None added. The new speaker uses `/i/noto-emoji/emoji_u1f50a.svg`, an icon VTT already ships (the same Noto emoji set the tutorial's existing bell uses), so `_meta.info.attribution` is unchanged.

## Testing

- `node validator/validate_gamefile_node.js "library/tutorials/Functions - AUDIO/0.json"` and `… 1.json` both exit 0 with no output; `node validator/validate_library.js` reports 0 problems.
- Walked both variants end-to-end in a real browser (Playwright, server on a local port, this branch cherry-picked onto #3047):
  - **clickSound variant** — the speaker with `"clickSound": ""` requests nothing when clicked (silent, as the caption says), while the neighbouring dice and bell still fetch their sounds. Putting the cursor on the `clickSound` line shows `pick a sound from the sound picker` right above `upload audio file`; the overlay opens, the search box filters (`card shuffle` → *Cards, Chips & Dice → card-shuffle*), and clicking the entry writes `"clickSound": "/i/audio/casino/card-shuffle.mp3"`. Rejoining the room outside edit mode and clicking the speaker then fetches `/i/audio/casino/card-shuffle.mp3` — the sound plays.
  - **AUDIO variant** — the same button appears on the `source` line of the Deal button's `AUDIO` function, the picker writes `"source": "/i/audio/casino/card-shuffle.mp3"` into the routine, and pressing *Deal* in a fresh tab fetches that file.
- Checked both boards render without overflowing the table; screenshots above are from the running app.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3099/pr-test (or any other room on that server)

### DIFF
--- a/library/tutorials/Functions - AUDIO/0.json
+++ b/library/tutorials/Functions - AUDIO/0.json
@@ -223,5 +223,27 @@
     },
     "text": "This bell has a 'clickSound' linked to an external URL.",
     "movable": false
+  },
+  "pickerSpeaker": {
+    "id": "pickerSpeaker",
+    "x": 1102,
+    "y": 690,
+    "image": "/i/noto-emoji/emoji_u1f50a.svg",
+    "clickSound": "",
+    "movable": false
+  },
+  "i4bj": {
+    "id": "i4bj",
+    "x": 902,
+    "y": 800,
+    "width": 500,
+    "height": 50,
+    "z": 75277,
+    "css": {
+      "font-size": "20px",
+      "color": "black"
+    },
+    "text": "This speaker has an empty 'clickSound' and stays silent. You do not need a sound file of your own: open it in the JSON Editor, put the cursor on the 'clickSound' line and use the 'pick a sound from the sound picker' button in the upper right corner. Preview the bundled sounds with ▶, click one to insert its '/i/audio/...' path, then leave edit mode and click the speaker.",
+    "movable": false
   }
 }
--- a/library/tutorials/Functions - AUDIO/1.json
+++ b/library/tutorials/Functions - AUDIO/1.json
@@ -1170,5 +1170,18 @@
       "font-size": "20px",
       "color": "black"
     }
+  },
+  "bm7j": {
+    "id": "bm7j",
+    "x": 108,
+    "y": 280,
+    "width": 1400,
+    "z": 10018,
+    "movable": false,
+    "text": "The 'source' can be any sound URL, but you do not have to find one yourself. Open the Deal button in the JSON Editor, put the cursor on the 'source' line of its AUDIO function and use the 'pick a sound from the sound picker' button in the upper right corner. Preview the sounds that ship with VTT using ▶, click one to insert its '/i/audio/...' path, then leave edit mode and press Deal to hear it.",
+    "css": {
+      "font-size": "20px",
+      "color": "black"
+    }
   }
 }


VTT-AI-Agent (commented):
## TL;DR

Content-only library change, both files validate and both boards render correctly — I loaded them into a local server and screenshotted them ([clickSound](https://agent.virtualtabletop.io/reports/review/1534328426100555816/audio-0.png), [AUDIO](https://agent.virtualtabletop.io/reports/review/1534328426100555816/audio-1.png)). The one thing worth fixing is the new paragraph in `1.json`: it keeps the default `height: 100`, so its (invisible) box reaches y=380 and covers the top of `holder2`, swallowing clicks there. Plus the hard dependency on #3047 — nothing in current `main` provides the "pick a sound from the sound picker" button these captions tell readers to use.

## Findings

**1. `Functions - AUDIO/1.json:1174` — the new paragraph `bm7j` overlaps the play area and sits on top of it**

`bm7j` is at `y: 280` with no `height`, so it gets the widget default of 100 → the box spans y 280–380 across x 108–1508. `holder2` starts at y=368 (`holder1` at y=361, mostly left of x=108). The paragraph's effective z is `(layer+10)*100000 + 10018`, far above the holders' — I confirmed in a browser that `document.elementsFromPoint()` at holder2's top edge returns `w_bm7j` first, i.e. the top ~11px of the drop holder and of any card dealt into it is covered by a transparent, non-movable widget, so clicks/drags there are swallowed.

Note also that the sibling intro paragraph `406b` right above it deliberately sets `"layer": -3` and `"movableInEdit": false` for exactly this reason; `bm7j` sets neither, so in edit mode a reader can also drag it around the board. Suggest mirroring `406b`: add `"height": 75` (the text is 3 lines ≈ 72px at 1400px width), plus `"layer": -3` and `"movableInEdit": false`.

**2. Merge order really is blocking, not just a nicety**

I checked the PR branch: neither `/i/audio` nor any sound-picker command exists — `client/js/jsonedit.js:558` still only offers `je_uploadAudio` ("upload audio file") for the `^.*\(AUDIO\) ↦ source|^.* ↦ clickSound` context. Merged before #3047, both new captions instruct readers to click a button that isn't there, and the `clickSound` variant ships a widget that is deliberately silent with no way to fix it — a worse tutorial than today's. Worth re-checking after #3047 lands that the picker command's `context` regex really matches a `clickSound` key whose value is `""` (the `pickerSpeaker` case) and not just a non-empty one.

**3. `Functions - AUDIO/0.json:246` — `i4bj` is one line away from running off the table**

The caption declares `height: 50` but the text renders 8 lines (~176 table px), ending at y≈976 on a 1000-tall table. It fits today (screenshot confirms), but with ~24px to spare, so any font fallback / slightly wider glyphs pushes the last line off the board. The neighbouring captions overflow their boxes too, but by one line, not seven. Setting a realistic `height` (or moving it up to ~y=780) would also make the widget outline in edit mode match what the reader sees.

**4. Minor, `1.json:1177`** — `font-size: 20px` on a full-width paragraph placed directly under the 25px intro (`406b`) reads as a shrunken continuation of it rather than as a caption. The 20px convention in this game applies to the per-widget captions in the bottom row, which are 300px wide; this one is 1400px wide and in the intro slot. 25px would match its position.

Nothing else: the new speaker reuses `/i/noto-emoji/emoji_u1f50a.svg` which is already shipped, `"clickSound": ""` is correctly falsy at `client/js/widgets/widget.js:517` so the widget really is silent, `_meta` is untouched, and `validate_gamefile_node.js` on both files plus `validate_library.js` report zero problems on my run.

I added the `beta` label: the PR touches only two library tutorial JSON files — no engine code at all — and it cannot corrupt games or crash the server, so it is safe for players to try on beta (where #3047 is also labelled `beta`, so the picker it references should be present).

VTT-AI-Agent:
Moved into #3047 as originally intended — the tutorial commit is cherry-picked onto `feature/audio-sound-picker` (90abb79) and described in that PR's body, so the docs land together with the feature they document. Closing this one.

PR #3100, opened 2026-08-05
Keep a dragged card hidden from the other players until it has left its hand

VTT-AI-Agent:

- [ ] **Decide on a visual cue at the private→public drag threshold** — Design call: whether an always-on cue should be added (the hook is ready); if yes, here or in a separate PR. ([discussion](https://github.com/ArnoldSmith86/virtualtabletop/pull/3100#issuecomment-5188703210))

A card that is dragged out of a `childrenPerOwner: true` hand stays hidden from the other players until it has actually left that hand, instead of being shown to the whole table as soon as another holder becomes the drop target. That happens with a holder stacked on top of a hand: the card is revealed while it is still being dragged around inside the hand it came from.

Reported on Discord:

> I tested and it seems that the dragged card is revealed before it normally would if another holder is on top of a `childrenPerOwner:true` holder.

(Source: https://discord.com/channels/770758631146782780/793626590848352306/1534420902320541799)

## What went wrong

`Widget.move()` picks the topmost droppable widget under the dragged widget's center as the hover target. If a holder is stacked on top of a hand, that holder wins the hit test while the dragged card is still completely inside the hand. The hover target change then ran `checkParent(true)`, a *forced* detach, which

- clears `owner` (the hand has `childrenPerOwner: true`) and `hoverParent`, and
- applies the hand's `onLeave` and runs its `leaveRoutine`.

`owner` and `hoverParent` are what keeps a card in a hand out of sight of everybody else - a widget owned by another player is not rendered at all, and `hoverParent` is what keeps a card showing its back while it is dragged. Clearing them is what reveals the card, so merely dragging a card across the part of your hand that another holder covers showed it to the whole table.

## The fix

**Leaving the holder keeps the timing it has always had.** `checkParent()` detaches the widget at exactly the same moment as before, clearing `parent`, `hoverParent` and `owner` and running the holder's `onLeave` and `leaveRoutine` in the same place with the same values - routines see the card leave the hand when they always did, and a `leaveRoutine` that assigns an `owner` (as the hands in `4-Letter Words` and `CONTEST` do) still has the last word over the clear.

What is new is a temporary override on top of that: when `checkParent()` detaches a widget that is still inside its holder, it puts `owner` and `hoverParent` back - the pair that hides a card in a hand from the other players - and remembers both the holder and the state the detaching left the widget in. That state is restored once the widget no longer overlaps the holder, or at the latest when the drag ends, before the drop, so that the holder the card lands in still has the last word.

Nothing changes for a widget that is already outside its holder when it is detached, which is every drag that does not involve stacked holders: there nothing is overridden and the same properties are cleared in the same order, in the same call, as before.

Verified in the browser with a hand (`childrenPerOwner: true`) and a second holder stacked on its right half:

| | before | after |
| --- | --- | --- |
| card leaves the hand (`onLeave`, `leaveRoutine`) | when another holder becomes the drop target | unchanged |
| dragging the card inside the hand, over the stacked holder | `owner` and `hoverParent` cleared - card visible to everyone | kept - card stays out of sight |
| dragging it half out of the hand, over nothing else | `owner` cleared - card visible to everyone | kept until the card is clear of the hand |
| dropping it on the stacked holder | `owner` cleared, `parent: over` | `owner` cleared, `parent: over` |

## Tests

New TestCafe fixture `tests/testcafe/drag.js` (added to both TestCafe workflow matrices). The hand in its room marks the card on the way out, once through `onLeave` and once through `leaveRoutine`, and every assertion checks those marks next to `owner`, `parent`, `hoverParent` and `hoverTarget` as the other players see them on the server - so a test fails if the card leaves at a different moment, not only if it is revealed at a different moment.

- a card dragged over a holder stacked on top of its hand leaves the hand right away but keeps its owner until the drop,
- a card dragged half out of its hand, with no other holder involved, keeps its owner until it no longer overlaps the hand,
- a card dragged out of the hand onto another holder loses its owner right away,
- an `owner` that the hand's `leaveRoutine` assigns survives the card leaving the hand, including on the deferred path.

The first two fail on `main`, and their failure there is exactly the `owner`/`hoverParent` pair - the `onLeave` and `leaveRoutine` marks match `main` in every assertion.

`npm test` (jest) plus `drag.js`, `editor.js`, `routines.js`, `boardsize.js`, `toolbar.js` and `publiclibrary-1..4` pass locally in Chromium.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3100/pr-test (or any other room on that server)

### DIFF
--- a/.github/workflows/testcafe-chrome.yml
+++ b/.github/workflows/testcafe-chrome.yml
@@ -21,6 +21,7 @@ jobs:
           - editor.js
           - toolbar.js
           - boardsize.js
+          - drag.js
           - compute-1.js
           - compute-2.js
           - compute-3.js
--- a/.github/workflows/testcafe-firefox.yml
+++ b/.github/workflows/testcafe-firefox.yml
@@ -21,6 +21,7 @@ jobs:
           - editor.js
           - toolbar.js
           - boardsize.js
+          - drag.js
           - compute-1.js
           - compute-2.js
           - compute-3.js
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -427,19 +427,21 @@ export class Widget extends StateManaged {
     return this.children().filter(c=>!c.get('owner') || c.get('owner')==playerName);
   }
 
-  async checkParent(forceDetach) {
-    const stillInside = this.currentParent && overlap(this.domElement, this.currentParent.domElement);
-    if(this.currentParent && (forceDetach || !stillInside)) {
+  // keepOwnerWhileInside is used by the drag that is on its way out of a holder: the widget
+  // is detached as soon as another holder becomes the drop target, which is while it is still
+  // completely inside its own holder when that other holder is stacked on top of it. owner
+  // and hoverParent are what keeps the widget out of sight of the other players, so they are
+  // kept until it has really left - otherwise a card is shown to the whole table before it
+  // has even left the hand it is being dragged around in. Leaving itself keeps its timing:
+  // onLeave and leaveRoutine still run right here.
+  async checkParent(forceDetach, keepOwnerWhileInside) {
+    if(!this.currentParent)
+      return;
+    const stillInside = (!forceDetach || keepOwnerWhileInside) && overlap(this.domElement, this.currentParent.domElement);
+    if(forceDetach || !stillInside) {
       await this.set('parent', null);
-      // A forced detach can happen while the widget is still completely inside the holder it
-      // is being dragged out of: a holder stacked on top of that one becomes the drop target
-      // as soon as the widget is dragged over it. owner and hoverParent are what keeps the
-      // widget out of sight of the other players, so they are only cleared once it has
-      // really left - otherwise a card is shown to the whole table before it has even left
-      // the hand it is being dragged around in. Leaving itself keeps its timing: onLeave and
-      // leaveRoutine still run right here.
       this.detachedParent = this.currentParent;
-      await this.checkDetachedParent(!stillInside);
+      await this.checkDetachedParent(!(keepOwnerWhileInside && stillInside));
       if(this.currentParent.dispenseCard)
         await this.currentParent.dispenseCard(this);
       delete this.currentParent;
@@ -2626,7 +2628,7 @@ export class Widget extends StateManaged {
       if (lastHoverTarget != this.hoverTarget) {
         await this.set('hoverTarget', this.hoverTarget ? this.hoverTarget.get('id') : null);
         if(this.hoverTarget != this.currentParent)
-          await this.checkParent(true);
+          await this.checkParent(true, true);
 
         // When the hover target changes we may need to create or remove the shadow widget.
         // Only create a shadow widget if the holder is shared and doesn't already have one in it.
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -427,36 +427,54 @@ export class Widget extends StateManaged {
     return this.children().filter(c=>!c.get('owner') || c.get('owner')==playerName);
   }
 
-  // keepOwnerWhileInside is used by the drag that is on its way out of a holder: the widget
-  // is detached as soon as another holder becomes the drop target, which is while it is still
+  // hideUntilOutside is used by the drag that is on its way out of a holder: the widget is
+  // detached as soon as another holder becomes the drop target, which is while it is still
   // completely inside its own holder when that other holder is stacked on top of it. owner
   // and hoverParent are what keeps the widget out of sight of the other players, so they are
-  // kept until it has really left - otherwise a card is shown to the whole table before it
-  // has even left the hand it is being dragged around in. Leaving itself keeps its timing:
-  // onLeave and leaveRoutine still run right here.
-  async checkParent(forceDetach, keepOwnerWhileInside) {
-    if(!this.currentParent)
-      return;
-    const stillInside = (!forceDetach || keepOwnerWhileInside) && overlap(this.domElement, this.currentParent.domElement);
-    if(forceDetach || !stillInside) {
+  // put back until it has really left - otherwise a card is shown to the whole table before
+  // it has even left the hand it is being dragged around in. Detaching itself is unchanged:
+  // parent, hoverParent and owner are cleared and onLeave / leaveRoutine run in the same
+  // place with the same values as before, so whatever the leave hooks assign is what the
+  // widget gets back once it is outside (see checkDetachedParent).
+  async checkParent(forceDetach, hideUntilOutside) {
+    if(this.currentParent && (forceDetach || !overlap(this.domElement, this.currentParent.domElement))) {
+      const parent = this.currentParent;
+      const hidden = hideUntilOutside && overlap(this.domElement, parent.domElement)
+        ? { hoverParent: this.get('hoverParent'), owner: this.get('owner') } : null;
+
       await this.set('parent', null);
-      this.detachedParent = this.currentParent;
-      await this.checkDetachedParent(!(keepOwnerWhileInside && stillInside));
-      if(this.currentParent.dispenseCard)
-        await this.currentParent.dispenseCard(this);
+      await this.set('hoverParent', null);
+      if(parent.get('childrenPerOwner'))
+        await this.set('owner',  null);
+      if(parent.dispenseCard)
+        await parent.dispenseCard(this);
       delete this.currentParent;
+
+      if(hidden) {
+        this.detachedParent = parent;
+        this.detachedParentState = { hoverParent: this.get('hoverParent'), owner: this.get('owner') };
+        await this.set('hoverParent', hidden.hoverParent);
+        if(parent.get('childrenPerOwner'))
+          await this.set('owner',  hidden.owner);
+      }
     }
   }
 
-  // Clears what checkParent() kept when it detached the widget while it was still inside its
-  // holder. Called on every move so that the widget stops being the owner's as soon as it no
-  // longer overlaps that holder, and with force when the drag ends.
+  // Reveals a widget that checkParent() hid in the holder it was detached from, by restoring
+  // the state the detaching left it in. Called on every move so that the widget becomes
+  // visible to the other players as soon as it no longer overlaps that holder, and with
+  // force when the drag ends, before the drop, so that the holder it lands in has the last
+  // word.
   async checkDetachedParent(force) {
     if(this.detachedParent && (force || !overlap(this.domElement, this.detachedParent.domElement))) {
-      await this.set('hoverParent', null);
-      if(this.detachedParent.get('childrenPerOwner'))
-        await this.set('owner',  null);
+      const parent = this.detachedParent;
+      const state = this.detachedParentState;
       delete this.detachedParent;
+      delete this.detachedParentState;
+
+      await this.set('hoverParent', state.hoverParent);
+      if(parent.get('childrenPerOwner'))
+        await this.set('owner',  state.owner);
     }
   }
 
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -2600,8 +2600,14 @@ export class Widget extends StateManaged {
 
       if (lastHoverTarget != this.hoverTarget) {
         await this.set('hoverTarget', this.hoverTarget ? this.hoverTarget.get('id') : null);
+        // A holder that sits on top of the current parent becomes the hover target while the
+        // widget is still completely inside that parent. Leaving the parent right away would
+        // run its onLeave and drop the owner - which reveals a card to everyone before it has
+        // even left the hand it is being dragged around in - so the widget only leaves once it
+        // no longer overlaps its parent. This repeats the check from the top of move() because
+        // the delta flushed above is what gave the widget its current geometry.
         if(this.hoverTarget != this.currentParent)
-          await this.checkParent(true);
+          await this.checkParent();
 
         // When the hover target changes we may need to create or remove the shadow widget.
         // Only create a shadow widget if the holder is shared and doesn't already have one in it.
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -428,14 +428,33 @@ export class Widget extends StateManaged {
   }
 
   async checkParent(forceDetach) {
-    if(this.currentParent && (forceDetach || !overlap(this.domElement, this.currentParent.domElement))) {
+    const stillInside = this.currentParent && overlap(this.domElement, this.currentParent.domElement);
+    if(this.currentParent && (forceDetach || !stillInside)) {
       await this.set('parent', null);
-      await this.set('hoverParent', null);
-      if(this.currentParent.get('childrenPerOwner'))
-        await this.set('owner',  null);
+      // A forced detach can happen while the widget is still completely inside the holder it
+      // is being dragged out of: a holder stacked on top of that one becomes the drop target
+      // as soon as the widget is dragged over it. owner and hoverParent are what keeps the
+      // widget out of sight of the other players, so they are only cleared once it has
+      // really left - otherwise a card is shown to the whole table before it has even left
+      // the hand it is being dragged around in. Leaving itself keeps its timing: onLeave and
+      // leaveRoutine still run right here.
+      this.detachedParent = this.currentParent;
+      await this.checkDetachedParent(!stillInside);
       if(this.currentParent.dispenseCard)
         await this.currentParent.dispenseCard(this);
       delete this.currentParent;
+    }
+  }
+
+  // Clears what checkParent() kept when it detached the widget while it was still inside its
+  // holder. Called on every move so that the widget stops being the owner's as soon as it no
+  // longer overlaps that holder, and with force when the drag ends.
+  async checkDetachedParent(force) {
+    if(this.detachedParent && (force || !overlap(this.domElement, this.detachedParent.domElement))) {
+      await this.set('hoverParent', null);
+      if(this.detachedParent.get('childrenPerOwner'))
+        await this.set('owner',  null);
+      delete this.detachedParent;
     }
   }
 
@@ -2546,6 +2565,8 @@ export class Widget extends StateManaged {
     await this.snapToGrid();
 
     if(!this.get('fixedParent') && this.get('movable')) {
+      await this.checkParent();
+
       const lastHoverTarget = this.hoverTarget;
       // The hit test below asks the DOM where this widget is, but the delta that
       // carries the position set above only reaches the DOM when the batch around
@@ -2557,12 +2578,10 @@ export class Widget extends StateManaged {
       if(this.domElement.style.transform != this.cssTransform())
         flushDelta();
 
-      // A widget stays in its holder until it no longer overlaps it - leaving as soon as the
-      // hover target changes would run the holder's onLeave and drop the owner while the
-      // widget is still inside, which reveals a card to everyone before it has even left the
-      // hand it is being dragged around in. This runs after the flush above so that the
-      // overlap is checked against where the widget is now, not where it was last event.
-      await this.checkParent();
+      // after the flush, so that a widget that was detached while it was still inside its
+      // holder is measured against where it is now: the last mouse event of a drag is the
+      // one that takes the widget out of the holder, and there is no next one to notice it
+      await this.checkDetachedParent();
 
       const myCenter = center(this.domElement);
       const myMinDim = Math.min(this.get('width'), this.get('height')) * this.get('_absoluteScale');
@@ -2606,6 +2625,8 @@ export class Widget extends StateManaged {
 
       if (lastHoverTarget != this.hoverTarget) {
         await this.set('hoverTarget', this.hoverTarget ? this.hoverTarget.get('id') : null);
+        if(this.hoverTarget != this.currentParent)
+          await this.checkParent(true);
 
         // When the hover target changes we may need to create or remove the shadow widget.
         // Only create a shadow widget if the holder is shared and doesn't already have one in it.
@@ -2726,6 +2747,9 @@ export class Widget extends StateManaged {
     delete this.stopDropLines;
 
     await this.set('hoverTarget', null);
+    // the drag is over, so the widget is not on its way out of anything any more - this runs
+    // before the drop below so that a holder taking the widget in still has the last word
+    await this.checkDetachedParent(true);
 
     if(!this.get('fixedParent') && this.get('movable')) {
       for(const t of this.dropTargets)
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -2546,8 +2546,6 @@ export class Widget extends StateManaged {
     await this.snapToGrid();
 
     if(!this.get('fixedParent') && this.get('movable')) {
-      await this.checkParent();
-
       const lastHoverTarget = this.hoverTarget;
       // The hit test below asks the DOM where this widget is, but the delta that
       // carries the position set above only reaches the DOM when the batch around
@@ -2558,6 +2556,14 @@ export class Widget extends StateManaged {
       // (or the game rejects the move and sends it back).
       if(this.domElement.style.transform != this.cssTransform())
         flushDelta();
+
+      // A widget stays in its holder until it no longer overlaps it - leaving as soon as the
+      // hover target changes would run the holder's onLeave and drop the owner while the
+      // widget is still inside, which reveals a card to everyone before it has even left the
+      // hand it is being dragged around in. This runs after the flush above so that the
+      // overlap is checked against where the widget is now, not where it was last event.
+      await this.checkParent();
+
       const myCenter = center(this.domElement);
       const myMinDim = Math.min(this.get('width'), this.get('height')) * this.get('_absoluteScale');
       this.hoverTarget = null;
@@ -2600,14 +2606,6 @@ export class Widget extends StateManaged {
 
       if (lastHoverTarget != this.hoverTarget) {
         await this.set('hoverTarget', this.hoverTarget ? this.hoverTarget.get('id') : null);
-        // A holder that sits on top of the current parent becomes the hover target while the
-        // widget is still completely inside that parent. Leaving the parent right away would
-        // run its onLeave and drop the owner - which reveals a card to everyone before it has
-        // even left the hand it is being dragged around in - so the widget only leaves once it
-        // no longer overlaps its parent. This repeats the check from the top of move() because
-        // the delta flushed above is what gave the widget its current geometry.
-        if(this.hoverTarget != this.currentParent)
-          await this.checkParent();
 
         // When the hover target changes we may need to create or remove the shadow widget.
         // Only create a shadow widget if the holder is shared and doesn't already have one in it.
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -444,17 +444,21 @@ export class Widget extends StateManaged {
 
       await this.set('parent', null);
       await this.set('hoverParent', null);
-      if(parent.get('childrenPerOwner'))
+      // remembered with the rest of the state below, so that the reveal restores exactly the
+      // properties that were cleared here even if a leave hook changes the flag mid-drag
+      const perOwner = parent.get('childrenPerOwner');
+      if(perOwner)
         await this.set('owner',  null);
       if(parent.dispenseCard)
         await parent.dispenseCard(this);
       delete this.currentParent;
 
-      if(hidden) {
+      // nothing to hide (and nothing to restore later) if a leave hook just removed the widget
+      if(hidden && !this.isBeingRemoved && !this.inRemovalQueue) {
         this.detachedParent = parent;
-        this.detachedParentState = { hoverParent: this.get('hoverParent'), owner: this.get('owner') };
+        this.detachedParentState = { hoverParent: this.get('hoverParent'), owner: this.get('owner'), perOwner };
         await this.set('hoverParent', hidden.hoverParent);
-        if(parent.get('childrenPerOwner'))
+        if(perOwner)
           await this.set('owner',  hidden.owner);
       }
     }
@@ -467,13 +471,12 @@ export class Widget extends StateManaged {
   // word.
   async checkDetachedParent(force) {
     if(this.detachedParent && (force || !overlap(this.domElement, this.detachedParent.domElement))) {
-      const parent = this.detachedParent;
       const state = this.detachedParentState;
       delete this.detachedParent;
       delete this.detachedParentState;
 
       await this.set('hoverParent', state.hoverParent);
-      if(parent.get('childrenPerOwner'))
+      if(state.perOwner)
         await this.set('owner',  state.owner);
     }
   }
@@ -658,6 +661,9 @@ export class Widget extends StateManaged {
     await this.set('dropShadowWidget', (await shadowWidget.clone({
         'movable': false,
         'dropShadowOwner': playerName,
+        // the dragged widget can still carry the hoverParent of the holder it is on its way
+        // out of - the shadow sits in the target holder and must not inherit that
+        'hoverParent': null,
         'parent': null}, true)).get('id'));
   }
 
@@ -2536,6 +2542,10 @@ export class Widget extends StateManaged {
     await this.bringToFront();
     await this.set('dragging', playerName);
     delete this.lastMoveCoord;
+    // a previous drag that did not reach checkDetachedParent(true) must not restore its state
+    // onto this one
+    delete this.detachedParent;
+    delete this.detachedParentState;
 
     // Lines that take a widget dropped onto their path as a stop. Collected once
     // like the drop targets below, but not restricted to widgets that can be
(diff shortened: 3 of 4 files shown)


VTT-AI-Agent (commented):
## TL;DR

The fix is correct and pleasingly small: turning the hover-target-change detach from forced into overlap-checked is exactly the right lever, and `moveEnd()` → `moveToHolder()` still forces the detach so `onLeave`/`onEnter`/`owner` at drop time are unchanged. My main point is scope, not correctness: this changes *when* a widget leaves its holder for **every** drag, not just the stacked-holder case — `owner`/`onLeave`/`leaveRoutine` are now deferred until the widget's bounding box no longer overlaps its parent at all, and that transition is not covered by a test.

## Findings

### 1. The behaviour change is broader than the reported bug, and the uncovered half is untested

`client/js/widgets/widget.js:2604` previously detached whenever the hover target changed away from the parent. That fired in two situations, and the PR changes both:

- **hovering another holder while still inside the parent** — the reported bug, clearly fixed.
- **the widget's centre leaves the parent (so `hoverTarget` goes `hand` → `null`) while the bounding boxes still overlap** — previously `owner` was cleared right there; now it is kept until `overlap()` is false, i.e. until the card is *fully* clear of the hand.

The second case is the common one (drag a card out of your hand toward empty table with no other droppable involved) and it now reveals the card noticeably later — for a small card leaving a large hand, only once it has completely cleared the hand's rectangle. That matches the PR title ("until it actually leaves it") and is arguably the intended semantics, but it is a user-visible engine change beyond the Discord report and it is worth stating explicitly in the description.

The same shift applies to `dispenseCard()` → holder `onLeave` + `leaveRoutine` (`client/js/widgets/holder.js:93`) and to the `receiveCard(null)` re-align of the source hand: for two overlapping holders, the source's `leaveRoutine` now runs at drop time rather than at hover time. Ordering relative to `onEnter` is preserved (`moveToHolder()` still does `checkParent(true)` before `set('parent', …)`), so I don't see a correctness problem — just a timing change games could notice.

`tests/testcafe/drag.js` covers hover-over-stacked-holder and drag-far-away (`table` at x800 vs. hand ending at x700 — no overlap at all), so neither test exercises the partially-overlapping exit that this change actually alters. A third case — drag the card just past the hand's right edge onto empty room, assert `owner` is still `'TestCafe'` while overlapping and `null` once clear — would pin down the new semantics.

### 2. The new `checkParent()` is redundant with the one at the top of `move()`

Once the call is unforced, both guards around it (`lastHoverTarget != this.hoverTarget` and `this.hoverTarget != this.currentParent`) are meaningless to it: an overlap check needs neither. The only thing the new call adds over `client/js/widgets/widget.js:2549` is that it runs *after* `flushDelta()`, i.e. against current geometry — the top-of-`move()` call always evaluates `overlap()` against the previous mouse event's DOM position.

So the clearer shape is to keep exactly one call and put it where the geometry is right:

```js
if(this.domElement.style.transform != this.cssTransform())
  flushDelta();
await this.checkParent();      // moved down from the top of move()
```

and drop the guarded call inside the `lastHoverTarget != this.hoverTarget` block entirely. That also removes the one-mousemove lag the current placement leaves in the non-hover-change path. (One thing to check if you do this: `dispenseCard()` can reposition the parent's remaining children, and after the move those repositions land after the flush rather than before it, which the hit test just below reads — probably fine since the dragged widget's own transform is what the flush is for, but worth a glance.) As-is the code is correct, so this is a cleanup, not a blocker.

### 3. `expectCardState()` sleeps 1.6s after its last read and never re-reads

`tests/testcafe/drag.js:47-56`:

```js
for(let wait=50; wait<2000; wait*=2) {
  actual = await cardState();
  if(matches) break;
  await new Promise(r => setTimeout(r, wait));
}
```

The iterations are 50/100/200/400/800/1600. The last read happens at t≈1550ms, then the loop sleeps 1600ms and exits *without* reading again — so the assertion reports the state from ~1.55s in, and the final 1.6s is pure dead time. Effective timeout is 1.55s, not the ~3.1s the `< 2000` bound suggests; on a loaded CI runner that is tighter than it looks. Either read once more after the loop, or sleep first and read after:

```js
for(let wait=50; wait<2000; wait*=2) {
  actual = await cardState();
  if(JSON.stringify(actual) == JSON.stringify(expected))
    break;
  await new Promise(resolve => setTimeout(resolve, wait));
}
actual = await cardState();   // one last look after the final wait
```

Also worth noting `test-util.js` already has `waitForStableState({ differentFrom })` for "wait until the interaction has reached the server and settled"; using it for the post-drop assertions (where the state does settle) would keep the ad-hoc polling to the mid-drag reads only.

### 4. The test asserts `owner` but not `hoverParent`

The PR description's before/after table rightly lists both — `hoverParent` is what drives the card-back inheritance at `client/js/widgets/widget.js:475` and `requiresHiddenCursor()` at `:2916`, and `checkParent()` clears it in the same breath as `owner`. `cardState()` reads `hoverTarget` but not `hoverParent`, so the half of the fix that keeps the card *rendered* as a back isn't locked down. Adding `hoverParent: card.hoverParent || null` to `cardState()` is a one-line strengthening.

### Nothing needed elsewhere

No widget or routine property is added or renamed, so `validator/` and the `jeCommands` array in `client/js/jsonedit.js` correctly stay untouched. `drag.js` is added to both the Chrome and Firefox matrices, and both jobs are green on this PR.

I did not add the `beta` label: this is a genuine engine change — it alters `Widget.move()`'s parent-detach semantics for every dragged widget in every game, not something gated behind a new opt-in property — so the PR is excluded from beta auto-deployment by design.

VTT-AI-Agent:
Thanks - all four points were valid, all four are addressed in 31c82a68.

**1. The behaviour change is broader than the reported bug, and the uncovered half is untested.** Agreed, and it turned out the previous commit did not even implement the half you describe: with the forced call merely unforced, dragging a card straight down out of the hand still detached it the moment the hover target went `hand` → `null`, because the *unforced* call at the top of `move()` runs before `flushDelta()` and so tested the overlap against the previous mouse event's position - which was still inside the hand. The new test you asked for failed on that commit. Fixing it the way you suggest in point 2 (one call, after the flush) is what actually makes the deferred exit work, so the two findings collapsed into one change.

The broader semantics are now spelled out in the PR description ("this changes *when* a widget leaves its holder for **every** drag ... a card leaving a hand is revealed once it is clear of the hand instead of once its center is"), together with a row in the before/after table for the partially-overlapping exit.

**2. The new `checkParent()` is redundant with the one at the top of `move()`.** Done exactly as proposed - the guarded call is gone and the top-of-`move()` call moved down below `flushDelta()`, leaving one overlap check per mouse move against current geometry. On the `dispenseCard()` concern: its `receiveCard(null)` re-align now lands after the flush, so for one mouse move the hit test below sees the holder's remaining children where they were before the re-align. That only matters if those siblings are themselves drop targets, and the whole public-library suite (which is drags plus state hashes) is unchanged, so I left it at that rather than adding a second flush.

**3. `expectCardState()` never re-reads after its final sleep.** Fixed, by reading before the loop and after each wait instead of after the loop, so the final 1.6 s is followed by a read:

```js
let actual = await cardState();
for(let wait=50; wait<2000 && JSON.stringify(actual) != JSON.stringify(expected); wait*=2) {
  await new Promise(resolve => setTimeout(resolve, wait));
  actual = await cardState();
}
```

I kept the ad-hoc polling for the post-drop assertions too: `waitForStableState()` waits for the state to settle, but these assertions want a *specific* state, and polling for it reports the actual mismatch instead of a timeout.

**4. The test asserts `owner` but not `hoverParent`.** Added to `cardState()`, so all three tests now pin `owner`, `parent`, `hoverParent` and `hoverTarget`.

Verified locally in Chromium: `npm test` (565 jest tests) and `drag.js`, `editor.js`, `routines.js`, `boardsize.js`, `toolbar.js`, `publiclibrary-1..4` all pass. Agreed on not adding `beta`.

VTT-AI-Agent:
Reworked so that the timing of leaving is untouched - thanks for the early warning.

> Just to clarify, anything that changes the timing of when leaveRoutine or onLeave get triggered is beyond the scope of what this PR will be allowed to do. That is a big issue with routines and cannot be fixed without a legacy mode. And we do not want this fix to take that on. If there is another solution that leaves the timing of leaving in place, pursue that.

(Source: Discord, https://discord.com/channels/770758631146782780/1534423568795046028/1534435198685679738)

The widget is detached from its holder at exactly the same moment as on `main`, and `onLeave` and `leaveRoutine` run right there, unchanged. The only thing that now waits for the widget to actually leave is the pair of properties that keeps a card in a hand out of sight of the other players: `owner` and `hoverParent`. `checkParent()` remembers the holder when it detaches a widget that is still inside it, and `checkDetachedParent()` clears those two once the widget no longer overlaps that holder - or when the drag ends, before the drop, so the holder the card lands in still has the last word. For a widget that is already outside its holder when it is detached (every drag that does not involve stacked holders) both properties are still cleared in the same call, in the same order, as before.

To make that testable rather than a claim, the hand in the test room now marks the card on its way out - once through `onLeave` and once through a `leaveRoutine` - and every assertion checks those marks alongside `owner`, `parent`, `hoverParent` and `hoverTarget`. On `main` the two failing tests fail *only* on `owner`/`hoverParent`; the `onLeave` and `leaveRoutine` marks are already set at the same points there, e.g. mid-drag over the stacked holder:

```
+ expected - actual
   hoverTarget: 'over'
   leaveRoutine: 'ran'      <- same on main and on this branch
   onLeave: 'ran'           <- same on main and on this branch
-  owner: null
+  owner: 'TestCafe'
-  hoverParent: null
+  hoverParent: 'hand'
```

`npm test` plus `drag.js`, `editor.js`, `routines.js`, `boardsize.js`, `toolbar.js` and `publiclibrary-1..4` pass locally in Chromium.

VTT-AI-Agent:
About the red check: it is on `31c82a68`, the commit with the approach that has since been replaced, and the job that failed is `routines.js` on Firefox - a fixture that never drags anything (it clicks a button that runs `SWAPHANDS`). Its `expectEventually()` reads the state at 0/50/150/350ms and then waits 400ms without reading again, so it effectively gives the routine 350ms to finish; on a loaded runner that is what ran out. The same file passed on Chrome for that commit and on both browsers for `4a2c64d7`. It looks like a pre-existing timing flake in that helper rather than anything this branch does, so I left it alone here - happy to tighten it in a separate PR if you want.

Looking into it did turn up a real bug in my own change, fixed in `333f5eba`: `checkParent(true)` is also how a routine-driven move (`moveToHolder()`) and the drop shadow take a widget out of its holder, and those widgets are inside their holder by definition - so they took the new "keep the owner until it has left" path and left a `detachedParent` behind that the next drag of that widget would have acted on, clearing an owner it had no business clearing. Keeping the owner is now asked for explicitly by the only call that needs it (the hover target change during a drag); every other caller detaches exactly as it did before, same properties, same order, same call.

Re-ran locally in Chromium after that: `npm test`, `drag.js`, `routines.js`, `editor.js`, `boardsize.js`, `toolbar.js` and `publiclibrary-1..4` all pass.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review). This PR ships no CSS, HTML or template changes — its whole user-visible surface is *what the other players see while you drag a card out of a `childrenPerOwner` hand*, and on that front it is a clear win: I reproduced the bug on `main` and confirmed it is gone on this branch in a real two-player browser session. The one thing I would add is a cue for the **dragging** player: your card silently crosses the private→public line and your own view looks identical on both sides of it.

**How I checked.** Server on my own port, Chromium via Playwright, two separate browser contexts ("Alice" owns the card, "Bob" is the other player), room = a `childrenPerOwner: true` hand with a second holder stacked on its right half (the reported setup). Same script run against `main` and against this branch. Colours added purely so the widgets are distinguishable in the screenshots.

| Bob's view, mid-drag over the stacked holder | |
| --- | --- |
| `main` — Alice's card is on the table for everyone | [main](https://agent.virtualtabletop.io/reports/ui-review/1534454911813812245/main-bob-during-drag-over-stacked-holder.png) |
| this PR — nothing leaks; the card stays Alice's | [pr](https://agent.virtualtabletop.io/reports/ui-review/1534454911813812245/pr-bob-during-drag-over-stacked-holder.png) |

Alice meanwhile sees exactly what she should — her card, and the stacked holder highlighted as the drop target: [pr-alice-during-drag](https://agent.virtualtabletop.io/reports/ui-review/1534454911813812245/pr-alice-during-drag-over-stacked-holder.png). No shadow/placeholder widget leaks into the stacked holder on Bob's side either, which is the other way this could have given the card away.

---

### Big layout improvements

Nothing structural to propose — no layout is touched. Two behavioural-UI points, in order of how much I think they matter:

**1. The player doing the dragging gets no feedback about when their card becomes public.** This is the one real UX gap the fix surfaces. `client/js/widgets/widget.js` now keeps `owner`/`hoverParent` until the card no longer overlaps its hand, so there is a genuine moment — geometric, invisible, and not where the pointer is — at which a card flips from "only I can see it" to "everyone can see it". Alice's own screen is pixel-identical before and after that moment:

| Alice, card half out of the hand — still private | Alice, card clear of the hand — now public to everyone |
| --- | --- |
| [private](https://agent.virtualtabletop.io/reports/ui-review/1534454911813812245/halfout-alice-still-private.png) | [public](https://agent.virtualtabletop.io/reports/ui-review/1534454911813812245/halfout-alice-now-public.png) |

(Bob's side confirms the split: [nothing](https://agent.virtualtabletop.io/reports/ui-review/1534454911813812245/halfout-bob-still-private.png) → [the card](https://agent.virtualtabletop.io/reports/ui-review/1534454911813812245/halfout-bob-now-public.png).)

This isn't a regression — the plain half-out drag behaved the same on `main` — but the PR makes the private window *longer* and now applies it to a card that has already been detached, so it is more likely a player leans on it. Suggestion, in keeping with the app's existing vocabulary rather than a new widget: give the dragged widget a temporary class while it is still `owner`-protected (the fix already has a natural hook — `this.detachedParent` / `hoverParent` being set) and style it with the same "not visible to others" language edit mode already uses for `.widget.hidden`/`.foreign` in `client/css/layout.css:876-890` — e.g. a subtle dashed outline or reduced opacity that drops away the instant the card goes public. That turns an invisible geometric threshold into something you can *see* crossing, which is exactly the safety the original bug report was asking for.

**2. On the receiving side, the card now materialises instead of arriving.** Because the card is hidden for the whole drag, Bob's first sight of it is at the drop, in its final position, with no motion leading up to it ([before](https://agent.virtualtabletop.io/reports/ui-review/1534454911813812245/pr-bob-during-drag-over-stacked-holder.png) → [after](https://agent.virtualtabletop.io/reports/ui-review/1534454911813812245/pr-bob-after-drop.png)). That is the intended and correct privacy trade-off, and I would not change the timing — but it does mean a card played to a shared holder can be missed by someone looking elsewhere on the table. Worth a sanity check that the existing arrival transition still reads as "something just happened here" when the widget appears from nothing rather than flying in; if it doesn't, that's a separate, small follow-up rather than something to hold this PR for.

None found attributable to this diff, and I looked for them rather than assuming: since `widget.js` is shared client code I ran an overlay sweep at **390×844 (phone portrait)**, **820×1180 (tablet)**, **1280×800 (laptop)** and **1920×1080 (wide)**, opening the **toolbar**, **Game Shelf** (`#statesButton`), **Players** (`#playersButton`), **About** (`#aboutButton`), **edit mode** and the **edit-mode toolbar/add panel** at each size, plus an automated check for anything rendering outside the viewport. Everything matched the pre-existing look; the only out-of-viewport hits were the normally scrollable Game Shelf list and About body, identical on `main`. Sample: [Game Shelf @ phone](https://agent.virtualtabletop.io/reports/ui-review/1534454911813812245/sweep-gameshelf-phone.png), [edit mode @ laptop](https://agent.virtualtabletop.io/reports/ui-review/1534454911813812245/sweep-addwidget-laptop.png), [About @ tablet](https://agent.virtualtabletop.io/reports/ui-review/1534454911813812245/sweep-about-tablet.png), [Players @ wide](https://agent.virtualtabletop.io/reports/ui-review/1534454911813812245/sweep-players-wide.png).

Overlays I did **not** exercise: the JSON/widget editor sub-dialogs, states/save & variant dialogs, share/publish, confirm prompts, pile/grid and chat panels. Given the diff contains zero CSS/HTML, I judged the risk of a regression there to be nil; flagging it so the coverage claim is honest rather than implied.

- **No new user-facing text, labels or icons** are introduced, so there is nothing here for a newcomer to misread. Good — the fix is entirely invisible when it works, which for a privacy bug is the right outcome.
- **One newly reachable state reads as self-contradictory in the JSON editor.** A widget mid-drag can now show `parent: null` together with `hoverParent: "hand"` (I observed exactly that: `{"owner":"Alice","parent":undefined,"hoverParent":"hand","hoverTarget":"over"}`). A game author inspecting a widget while dragging will see "it has no parent, but it does have a hoverParent" and reasonably conclude something is broken. This is a documentation nit rather than a UI one, but it's the kind of thing that generates confused Discord questions: worth one line in the Widgets wiki page saying `hoverParent` outlives `parent` while a widget is on its way out of a holder, and that this is what keeps the widget hidden from the other players.
- **The reveal rule is "no pixel of the card overlaps the holder"**, which is precise but not discoverable — a player's mental model is "when I've pulled it out of my hand", and for a hand whose background is visually smaller than its widget bounds those two differ. Point 1 above (a visible private/public cue) is what would make this legible without changing the rule.

Nothing here blocks the change from a UI standpoint — the interface it produces is strictly better than `main`.

VTT-AI-Agent:
Thanks - the live two-player verification is the one thing the test suite cannot give me, and the four-viewport sweep for a diff with zero CSS is appreciated. Short version: I am **not** implementing point 1 on this branch, for scope reasons spelled out below; points 2 and 3 I think are answered without a change. So no new commit here.

### 1. No cue for the dragging player at the private -> public threshold

The observation is correct, and it is a real gap - but I do not think this PR is the place to close it.

- **It is not a regression, and this branch only ever moves the threshold later.** On `main` a card went public when its *centre* left the hand; here when its *last pixel* does. Every moment that is private on `main` is private here too, so no card is exposed earlier than players are used to - the change is strictly in the safe direction. Your own [half-out screenshots](https://agent.virtualtabletop.io/reports/ui-review/1534454911813812245/halfout-alice-still-private.png) show that the "my own view looks identical on both sides of the line" part is unchanged from `main`.
- **The cue would not be limited to what this PR touches.** `detachedParent`, the hook you name, is set on *every* drag out of a `childrenPerOwner` holder - including the plain drag out of a hand that behaves exactly as on `main`. So the new style would appear on essentially every card pulled out of a hand in every card game: a new always-on visual vocabulary in the core engine, not a fix for the reported bug.
- **That makes it a design call for the maintainer**, and this PR has already been scoped down once, in the other direction:

  > Just to clarify, anything that changes the timing of when leaveRoutine or onLeave get triggered is beyond the scope of what this PR will be allowed to do. That is a big issue with routines and cannot be fixed without a legacy mode. And we do not want this fix to take that on. If there is another solution that leaves the timing of leaving in place, pursue that.

  (Source: Discord, https://discord.com/channels/770758631146782780/1534423568795046028/1534435198685679738)

  A permanent new decoration on dragged cards is a more visible change to every game than the timing change that was ruled out, so slipping it into a bugfix PR unasked seems the wrong way round.

If @ArnoldSmith86 wants it, it is small and the hook is ready: `classes()` already keys off plain JS fields (`this.isLimbo` at `client/js/widgets/widget.js:494`), so `this.detachedParent` can add a class the same way, refreshed from `checkParent()`/`checkDetachedParent()`, plus one rule next to `.widget.limbo` in `client/css/layout.css`. That belongs in its own PR - one concern per PR - so I am not opening it from here.

### 2. On the receiving side the card materialises instead of arriving

This only happens when the drop itself lands over the hand - i.e. exactly the stacked-holder case that is the bug. `checkDetachedParent()` runs on every mouse move, so the moment the card no longer overlaps the hand the other players get it and then watch it travel to wherever it is dropped; [halfout-bob-now-public.png](https://agent.virtualtabletop.io/reports/ui-review/1534454911813812245/halfout-bob-now-public.png) is that moment mid-drag, and the drag continues visibly from there. For a drop *inside* the hand's rectangle the card is by definition never outside it before the drop, so any earlier reveal would be the leak this PR fixes.

So there is nothing to trade off here: the arrival animation for a widget appearing near the table is the same one as on `main` (the reveal-then-fly case is unchanged), and whether that animation reads loudly enough in general is independent of this branch.

### 3. `parent: null` next to `hoverParent: "hand"` reads as contradictory

Agreed, and it is worth a wiki line - but the wiki is a separate repository, so I cannot include it in this PR. Suggested wording for whoever edits the Widgets page:

> While a widget is being dragged out of a holder, `hoverParent` can outlive `parent`: `parent` is cleared the moment the widget is detached from the holder, while `hoverParent` - and `owner`, for a `childrenPerOwner` holder - is kept until the widget no longer overlaps that holder. That is what keeps a card hidden from the other players until it has really left the hand it came from.

In the code the same window is documented in the comments on `checkParent()` and `checkDetachedParent()` in `client/js/widgets/widget.js`, so someone reading the source lands on the explanation; it is only the JSON editor view that lacks the context.

VTT-AI-Agent (commented):
## TL;DR

The stacked-holder leak is fixed for the tested case, but the deferred cleanup introduces a high-impact compatibility regression: it now clears `owner` *after* `onLeave`/`leaveRoutine`, overriding ownership those hooks intentionally assign. This affects existing library games and can expose cards that their routines keep private on `main`.

## Findings

### 1. Deferred cleanup overrides leave-hook ownership (high)

`client/js/widgets/widget.js:443-458` records the old holder and eventually clears `owner` in `checkDetachedParent()`, but `checkParent()` calls `dispenseCard()` at lines 445-446 before that deferred clear occurs. `dispenseCard()` runs the holder's `onLeave` and `leaveRoutine`. On `main`, the order is the opposite: `checkParent()` clears the old hand's owner first, then the leave hooks run, so a hook that assigns a new owner has the last word.

That ordering is used by shipped games. For example, the `childrenPerOwner` hand in `library/games/4-Letter Words/0.json:875` has a `leaveRoutine` that does `SET collection: child, property: owner, value: ${playerName}`; `CONTEST` and `Scotland_s Seven Hills` use the same pattern. In the new stacked-holder path, that routine runs while the widget is still inside the old hand, then the next move outside (or `moveEnd()` at `client/js/widgets/widget.js:2754`) unconditionally writes `owner: null`. A card that deliberately remained owned/private after leaving the hand on `main` therefore becomes public. The new test only makes the leave hooks write `classes` and `text`, so it cannot catch the ordering regression.

Please preserve the post-`dispenseCard()` ownership result when releasing the temporary privacy hold (including the case where the routine writes the same value the card already had), and add a regression case where a `childrenPerOwner` hand's `onLeave` or `leaveRoutine` assigns `owner` before the card crosses the deferred boundary.

I did not add the `beta` label because `client/js/widgets/widget.js` changes default drag/ownership engine semantics for existing games, making this PR engine-ineligible.

VTT-AI-Agent:
Valid, and thanks for pinning down the ordering - fixed in `68e5f449`.

**1. Deferred cleanup overrides leave-hook ownership.** You are right that `main`'s order is "clear, then run the hooks", and that a `leaveRoutine` doing `SET owner: ${playerName}` therefore had the last word there while my version overwrote it afterwards. Comparing values at reveal time would not have been enough either, exactly as you note: in `4-Letter Words` the routine assigns the owner the card already had, so "unchanged" and "reassigned" look identical.

So the detour is gone. `checkParent()` now does what it does on `main`, in the same order and with the same values - `parent`, `hoverParent` and `owner` are cleared, then `dispenseCard()` runs `onLeave` and `leaveRoutine`. Only *after* that, and only when the widget is still inside the holder it was detached from, does it put `owner`/`hoverParent` back as a temporary override, remembering what the hooks left behind:

```js
      if(hidden) {
        this.detachedParent = parent;
        this.detachedParentState = { hoverParent: this.get('hoverParent'), owner: this.get('owner') };
        await this.set('hoverParent', hidden.hoverParent);
        if(parent.get('childrenPerOwner'))
          await this.set('owner',  hidden.owner);
      }
```

`checkDetachedParent()` then restores `detachedParentState` instead of writing `null`, so the reveal hands the widget back to whoever the leave hooks gave it to - including "nobody", which is the case for a holder without such a routine.

No flicker from the clear-then-restore: the whole mouse event runs inside `batchStart()`/`batchEnd()` (`client/js/mousehandling.js:75`) and the delta collapses per property, so the intermediate `owner: null` never leaves this client. The one exception is a `leaveRoutine` that flushes mid-way (`DELAY`, `INPUT`) - and there `main` is exposed at least as long, since it leaves the owner cleared for the rest of the drag.

**Regression test.** `drag.js` gained "An owner assigned by a leaveRoutine survives the card leaving its hand": same room, but the hand's `leaveRoutine` also does `SET owner: ${playerName}` like the library games, and the card is dragged over the stacked holder (detached while still inside, so it takes the deferred path), then out onto a free spot on the table. It pins `owner: 'TestCafe'` mid-drag after the reveal and after the drop. On `333f5eba` it fails exactly as you predicted:

```
+ expected - actual
-  owner: null
+  owner: 'TestCafe'
```

I also ran it against `main`'s `widget.js` to check which half of the state this branch is allowed to change: there the mid-drag state is `owner: 'TestCafe', hoverParent: null` - the routine's owner survives, only the early reveal of `hoverParent` differs. That is now the same on both.

Verified locally in Chromium: `npm test` (565 jest tests) plus `drag.js`, `routines.js`, `editor.js` and `publiclibrary-1..4` all pass.

Agreed on not adding `beta`.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-scoped fix with a genuinely good test fixture — I reproduced it: the new `drag.js` passes 4/4 on this branch and 3/4 of its tests fail on `main`, and `npm test` stays green. Nothing blocking; the one thing I'd actually change is that the drop-shadow clone now inherits the restored `hoverParent` of the source holder (verified in the browser), plus a couple of cheap guards so a stale `detachedParent` can't survive a drag.

## Findings

### 1. The drop shadow inherits the restored `hoverParent` (client/js/widgets/widget.js:658)

`createShadowWidget()` clones the dragged widget with `{movable, dropShadowOwner, parent: null}` — it does **not** override `hoverParent`. On `main` the clone was always made after `checkParent(true)` had cleared `hoverParent`; with this PR the clone can now be made while `hoverParent` has been put back to the source holder.

I confirmed this in the browser (hand with `childrenPerOwner: true`, a `dropShadow: true` holder stacked on it, card dragged from the hand onto the stacked holder) — the shadow widget on the server reads:

```
{"id":"jyo2","type":"card","parent":"over","owner":null,"hoverParent":"hand","dropShadowOwner":"TestCafe"}
```

`owner` is fine (`moveToHolder()` clears it), but `hoverParent: "hand"` is a leak. It matters through `classes()` (widget.js:514): if the source holder has `showInactiveFaceToSeat`, the shadow preview sitting in the *target* holder is rendered with the source hand's card-back treatment. Cosmetic and narrow, but it's a state the shadow could never have before. Cheapest fix is to add `'hoverParent': null` to the clone overrides at widget.js:658-661.

### 2. `detachedParent` is never cleared defensively (client/js/widgets/widget.js:2772, 2532)

`checkDetachedParent()` is the only thing that deletes `detachedParent`/`detachedParentState`. `moveEnd()` calls it with `force` early enough, but not before `hideShadowWidget()` and `set('dragging', null)` — if either throws, the widget is left with the previous drag's `owner`/`hoverParent` *and* with `detachedParent` still set, so the **next** drag's first `checkDetachedParent()` writes stale values from the old drag onto the widget. Two lines in `moveStart()` remove the whole class of problem:

```js
delete this.detachedParent;
delete this.detachedParentState;
```

### 3. The `childrenPerOwner` flag is read twice, at two different times (widget.js:447/457 vs 476)

`checkParent()` decides whether to save/restore `owner` from `parent.get('childrenPerOwner')` at detach time; `checkDetachedParent()` re-reads the same flag from the holder at restore time, potentially many mousemoves later. A `leaveRoutine` that sets `childrenPerOwner: false` on the hand (or anything else that changes it mid-drag) makes the restore skip `owner` entirely — the card then keeps the dragging player's `owner` after the drop and stays invisible to every other player, which is exactly the failure mode this PR sets out to prevent, just inverted. Storing the decision with the rest of the snapshot is symmetric with how `hidden` is captured:

```js
this.detachedParentState = { hoverParent: ..., owner: ..., perOwner: parent.get('childrenPerOwner') };
```

(Same read also happens on a holder that a routine may have removed during the drag; `overlap()` on its orphaned `domElement` just returns false, so that path degrades into the same skipped-`owner` case rather than a crash.)

### 4. The restore writes properties after `dispenseCard()` has run the leave hooks (widget.js:453-459)

On `main`, `dispenseCard()` — i.e. `onLeave` + `leaveRoutine` — was the last thing `checkParent()` did to the widget. Now two `set()` calls follow it. If a `leaveRoutine` deletes the child and something in it flushes the delta in between (an `INPUT`, or a server-random op), those two `set()`s emit `{hoverParent, owner}` for an id the server has already dropped, and `receiveDelta()` (server/room.mjs:768) takes `this.state[widgetID] === undefined` as "new widget" and recreates a ghost entry holding only those two properties. Within one batch `sendPropertyUpdate()`'s `delta.s[widgetID] !== null` guard swallows it, so this needs the flush — narrow, but an `if(!this.isBeingRemoved)` around the `if(hidden)` block costs nothing.

## What I verified

- `tests/testcafe/drag.js`: **4/4 pass** on this branch, **3/4 fail** with `client/js/widgets/widget.js` reverted to `main` (test 2, "revealed as soon as it is dragged out", passes on both — as the description says). The fixture is a good addition; asserting the `onLeave`/`leaveRoutine` marks next to `owner`/`hoverParent` is what makes it a real regression test for the leave *timing* rather than only for the reveal.
- `npm test`: 565 passed.
- Manually, in Chromium: dropping the card back into its own hand after hovering the stacked holder ends correctly (`parent: hand`, `owner: TestCafe`), and the `dropShadow` case above.
- No new widget property or routine function is exposed to game authors here (`hideUntilOutside` is an internal argument), so there is nothing owed to the property editor, `jeCommands` or the validator.

I did not add the `beta` label: this changes `client/js/widgets/widget.js` drag/detach semantics — `owner` and `hoverParent` for every widget dragged out of any holder — so it really does alter engine behavior for existing games and is excluded from beta auto-deployment by design.

VTT-AI-Agent:
All four valid - fixed in `26b3afa5`, plus a fifth testcafe test that pins the drop shadow.

**1. The drop shadow inherits the restored `hoverParent`.** Confirmed, and thanks for chasing it into the browser. `checkParent(true, true)` runs immediately before `createShadowWidget()` in `move()`, so on this branch the clone is made exactly when `hoverParent` has been put back. Fixed with your suggested override:

```js
    await this.set('dropShadowWidget', (await shadowWidget.clone({
        'movable': false,
        'dropShadowOwner': playerName,
        // the dragged widget can still carry the hoverParent of the holder it is on its way
        // out of - the shadow sits in the target holder and must not inherit that
        'hoverParent': null,
        'parent': null}, true)).get('id'));
```

`null` in the override lands in the clone's state as an explicit `null` rather than an absent key - that is already how `owner` looks on a shadow today (`moveToHolder()` clears it inside the same batch that creates the widget, which is why your dump shows `"owner":null`), so this adds no new shape to the state.

New test, `The drop shadow of a holder stacked on a hand does not inherit that hand`: same room with `dropShadow: true` on the stacked holder, card dragged onto it, asserting `{ parent: 'over', hoverParent: null }` for the shadow on the server while the card itself is still `owner: TestCafe, hoverParent: hand`. Without the one-line fix it fails with `hoverParent: 'hand'` and the card-back leak you described; with it, it passes and the shadow is gone again after the drop.

**2. `detachedParent` is never cleared defensively.** Added to `moveStart()`, right where `lastMoveCoord` is dropped:

```js
    // a previous drag that did not reach checkDetachedParent(true) must not restore its state
    // onto this one
    delete this.detachedParent;
    delete this.detachedParentState;
```

**3. `childrenPerOwner` is read twice.** Agreed, and the inverted failure (card keeps the dragger's `owner` and stays invisible to everyone) is the worse half. The flag is now captured once, at the same moment the clear happens, and travels with the snapshot:

```js
      const perOwner = parent.get('childrenPerOwner');
      if(perOwner)
        await this.set('owner',  null);
      ...
        this.detachedParentState = { hoverParent: ..., owner: ..., perOwner };
```

`checkDetachedParent()` reads `state.perOwner`, so the reveal now restores exactly the properties the detach cleared. That also drops the last `get()` on the remembered holder, so the "holder removed mid-drag" case you mention no longer touches an orphaned widget at all - only `overlap()` on its detached `domElement`, which returns false and reveals, as before.

**4. The restore writes after `dispenseCard()`.** Guarded, using the same pair of flags `evaluateRoutine()` checks at `client/js/widgets/widget.js:993`:

```js
      // nothing to hide (and nothing to restore later) if a leave hook just removed the widget
      if(hidden && !this.isBeingRemoved && !this.inRemovalQueue) {
```

Skipping the block also means no `detachedParent` is left behind pointing at a holder that will never reveal it.

**Verified locally in Chromium:** `npm test` (565 jest tests) and testcafe `drag.js` (5/5), `routines.js`, `editor.js`, `toolbar.js`, `boardsize.js`, `publiclibrary-1..4` - all pass.

Agreed on not adding `beta`.

PR #3101, opened 2026-08-05
Open the Edit Widgets sidebar module by default when entering edit mode

VTT-AI-Agent:
Entering edit mode now opens the **Edit Widgets** (properties) module in the top left module slot the first time someone enters edit mode, instead of showing an empty sidebar next to the room.

## Why

`renderSidebar()` in `client/js/editor/layout.js` only restored the modules stored in `localStorage.editorState`. For anyone who has never opened a module - which is exactly the first-time user clicking the Edit button - that meant edit mode came up as the room plus a column of icons, with no hint that those icons are what opens the panels. The properties panel is what edit mode is about, and with no widget selected it explains itself ("You do not have a widget selected... Add a new widget / Open deck editor").

## What changed

- `client/js/editor/layout.js`: when the stored state opened no module, the properties module is opened in `#editorModuleTopLeft`. A remembered module still wins.
  - Only the first time ever: `editorState.defaultModuleOpened` records that the default has been shown. Closing the last module *deletes* its entry, so "nothing remembered" is also how "I want the whole room" is stored - without the flag, a user who deliberately closes everything would get the panel back on every reload.
  - Only where the panel and the room can coexist: `calculateEditModuleClasses()` turns `#editorModules` into a fullscreen overlay on narrow windows, and a portrait window showing a landscape board (`isOrientationMismatch()`) is busy asking the user to rotate the device. In both the sidebar of icons stays as it was.
  - Sized to its content, not to the window: a module the user opens gets `--modulesWidth`, which defaults to a 50/50 split - a poor first sight of edit mode, since half the room disappears for a panel that is mostly empty. The panel that opens itself is `min(420px, 35vw)` wide instead (`body.defaultEditorModuleWidth`), and hands the width back to the stored value as soon as a module is opened by hand or the resizer is dragged. `editorState.defaultModuleWidth` remembers that across page loads - a restored default is still a panel nobody chose - and is deleted by those same two events. At 1920x1080 the room stays 1372x858 instead of 960x600.
- `client/js/editor/sidebarModule.js`: every module gets a close button (`addCloseButton()`), so a panel that opened itself can be dismissed without knowing that the sidebar button toggles. It goes into the module header where there is one (which is also what puts it back when a module re-renders itself), and into the module's own top right corner in the three that render no header - JSON, Debug and Assets. The properties module also puts it on its widget header, so it does not vanish when a widget is selected.
- `client/js/editor/sidebar/properties.js`: the "no selection" state no longer says "click on one to the left" (wrong in the layouts that put the panel above the room), points at the sidebar for the other tools, and its two buttons get tooltips. The Automations "Full size" switch built its DOM id from `rand()`, which spends a number of the game's random sequence (and feeds `traceRandom`), so rendering the panel shifted the ids of every widget created afterwards; it uses the existing `editorDomID()` helper now. That only became visible because the panel is rendered without being asked for it - a small fix, but the deterministic-id tests depend on it.
- `client/css/editor/*`: the content width for the self-opened panel, the close button's placement, and `max-width` on the "no selection" buttons so they stop stretching into full-width banners.
- `client/js/main.js`: `calculateEditModuleClasses` and `isOrientationMismatch` join the functions the editor bundle gets on `window` (the editor is a separate dynamically imported bundle and only sees what is assigned there).
- `Ctrl+J` on a cold start is untouched: it still force-opens the JSON editor (which now simply replaces the default properties panel in that slot). Verified in the browser.
- `tests/testcafe/editor.js`: the tests that clicked the Edit Widgets button right after entering edit mode would now toggle the panel *closed*, so they assert that it is open instead (which is also the wait for edit mode to have loaded). `Create game using edit mode` closes it explicitly because it places widgets across the whole room, and `Dice faces...` opens it explicitly because it runs in a portrait window. Six new tests cover the default and its width, a remembered module winning, the "I closed them all" case, the narrow-window and portrait layouts, and closing a module through its header button.

- `tests/testcafe/editor.js` (after merging main): the symbol picker test added by #3105 clicked the Edit Widgets sidebar button right after entering edit mode, which now toggles the module that opened by default closed again, so the "Pick a symbol" button was never in the DOM. It waits for the module instead and states its own window size and editor state, like the other tests in the file that expect the properties panel.

## Verification

- `npm test` (jest): 565 passed.
- TestCafe `editor.js` (37 tests incl. the six new ones): all passed locally.
- Browser check in Chromium at 1920x1080, 1280x800, 1001x600, 900x600, 820x1180 and 390x844: fresh `localStorage` -> Edit Widgets open at content width with the room dominant; opening the JSON module by hand -> full stored width; header `x` -> panel closed and not back after a reload; 900x600 and portrait windows -> no default panel, room and the "rotate your device" message visible.

## Verification after the merge with main

- `npm test` (jest): 1167 passed.
- TestCafe `editor.js` (39 tests) and `routines.js` against a **minified** client, which is what the Production Environment workflow drives since #3089: all passed locally. The two names this change adds to the `Object.assign(window, ...)` block in `main.js` (`calculateEditModuleClasses`, `isOrientationMismatch`) survive the new top-level mangling, and `tests/server/minify.test.js` reads that list out of the source, so they are covered by it without being listed anywhere.
- Browser check in Chromium at 1920x1080 on the merged tree: Edit Widgets open at 420px with its close button, room 1372x858.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3101/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

Small, well-motivated change and the test updates look right. The one thing I'd actually change: the new per-render counter in `properties.js` duplicates the existing `editorDomID()` helper (same purpose, same rationale comment) that this very file already uses twice — just call it. Beyond that, two behavioural points worth a thought: there is no way for a user to *stay* on "no module open", and on narrow windows the default panel covers the whole room.

## Findings

**1. `client/js/editor/sidebar/properties.js:1334-1335, 10246` — the counter already exists as `editorDomID()`**

`client/js/editor/propertyInputs.js:47-51` has exactly this helper, with exactly this reasoning already written down:

```js
// Unique ids for the DOM elements the editor generates. Deliberately not
// rand(): that advances the seeded random the game state depends on, so
// rendering an input would change the next dice roll.
let editorDomIDCounter = 0;
function editorDomID(prefix) { return `${prefix}_${++editorDomIDCounter}`; }
```

`properties.js` is in the same bundle scope and already calls it at `properties.js:1366` (`editorDomID('genericInput')`) and `properties.js:5638`. So the new module-level `automationsFullSizeToggles` and its comment can both go away:

```js
input.id = editorDomID('automationsFullSize');
```

Same behaviour, one counter instead of two, and the "why not rand()" explanation stays in the one place it's already written. (The fix itself is correct and worth having — `rand()` at `domhelpers.js:100` feeds `traceRandom` and the seeded sequence, so this was a real determinism leak; it's only the second copy of the mechanism I'd drop.)

**2. `client/js/editor/layout.js:132-138` — "no module open" is not a state a user can keep**

`SidebarModule.saveToLocalStorage(null)` (`client/js/editor/sidebarModule.js:246-252`) *deletes* the entry when a module is closed, so `{"modules":{}}` means both "never opened one" and "deliberately closed the last one". After this PR the second case is no longer honoured: a user who closes the Edit Widgets panel because they want the full room gets it back on every reload (`renderSidebar()` only runs once per page load, so within a session closing it sticks — the surprise is on the next visit).

Since the stated goal is the *first-time* user, a one-shot flag would express it better and cost one line, e.g. write `editorState.defaultModuleOpened = true` alongside the default open and gate on `!state.defaultModuleOpened` in addition to `!opened`. Then the default fires exactly once ever, and an experienced user's "I closed everything" is respected.

Related edge case in the current condition: a module remembered as `editorModuleInOverlay` is skipped by the existing filter, so `opened` stays false and Properties is opened in the top-left *in addition to* whatever the overlay path later restores. Worth a check that this combination behaves.

**3. `client/js/editor/layout.js:136` — on narrow windows the default panel hides the room entirely**

`calculateEditModuleClasses()` (`client/js/calculateLayout.js:130-140`) adds `editModulesOverlay` for windows at or below the narrow threshold, and `client/css/editor/sidebar.css:303-312` then gives `#editorModules` `top: toolbar; bottom: 0` — a fullscreen overlay next to the 36px sidebar. `editModulesAbove` similarly pushes the room below the panel. On a phone/small window, entering edit mode for the first time now lands on a properties panel that covers the room, with the "You do not have a widget selected" text and no visible widgets to select — arguably a worse first impression than the column of icons this PR is fixing, because the room disappeared.

Consider skipping the default open when `calculateEditModuleClasses()` returns `editModulesOverlay` (the layout where the panel and the room can't coexist), and keeping it for the side-by-side layout the PR description is really about. I couldn't verify this in the sandbox (no `node_modules` in the fresh clone), so please sanity-check it in a narrow window before merging — it's the one claim here I'm reasoning about from the CSS rather than from a screenshot.

**4. `tests/testcafe/editor.js:19-30` — the second new test leaves `editorState` set**

It writes `{modules:{JSON:'editorModuleTopLeft'}}` and never clears it, while ~15 later tests in the same file now assert `#editorModuleTopLeft.tune` exists. That only holds because TestCafe clears browser storage between tests; the fact that the *first* new test defensively calls `localStorage.removeItem('editorState')` suggests the assumption wasn't fully trusted. Cheapest fix is to make it symmetric — have the remembered-module test remove the key again at the end (or drop the `removeItem` from the first test if the cleanup is genuinely guaranteed), so a future TestCafe config change doesn't turn this into fifteen confusing failures.

**5. Nothing else blocking**

`Create game using edit mode` correctly closes the panel now, the `.click([icon=tune])` → `.expect(#editorModuleTopLeft.tune).ok()` swaps keep their role as the "edit mode has loaded" wait, and `renderSidebar()` still runs before `metaReceived`/`openEditor()` exactly as it did for a remembered module, so no new ordering assumption. No new widget property or routine function is exposed here, so no editor-UI/validator/`jeCommands` follow-up is needed.

I added the `beta` label: the diff touches only `client/js/editor/*` and TestCafe tests — no engine file (`client/js/widgets/`, `compute.js`, `serverstate.js`) — and an edit-mode default that a user can toggle off is safe and useful for players to try.

VTT-AI-Agent:
Thanks - all four points were valid, pushed as 59df6fa.

**1. `editorDomID()` instead of a second counter** - fixed. `properties.js` already calls it twice, so the extra module-level counter and its duplicate "why not rand()" comment are gone:

```js
input.id = editorDomID('automationsFullSize');
```

**2. "no module open" is not a state a user could keep** - fixed with the one-shot flag you suggested. `renderSidebar()` now writes `editorState.defaultModuleOpened = true` when it opens the default, and gates on it, so the default fires exactly once ever and a deliberate "I closed everything" survives a reload. The flag is written *after* `openInTarget()` and from a re-read of `editorState`, because `saveToLocalStorage()` rewrites that key in between.

On the `editorModuleInOverlay` edge case: nothing restores the overlay module on load (the filter is the only place that entry is read at startup), so the combination is just "Properties opens in the top left and the stale overlay entry stays in storage" - the same as before this PR, minus the panel. With the flag it now also only happens once.

**3. Narrow windows** - confirmed, and fixed. Screenshots from a 900x600 window (`editModulesOverlay`, `--no-sandbox` Chromium against the local server):

- what the default would have looked like: https://agent.virtualtabletop.io/reports/pr/1534541516087758929/narrow-module-open.png - the panel covers the whole room, "click on one to the left" with nothing to the left of it.
- with the fix: https://agent.virtualtabletop.io/reports/pr/1534541516087758929/narrow-default-skipped.png - room visible, sidebar as before.

So `renderSidebar()` now skips the default when `calculateEditModuleClasses()` returns `editModulesOverlay`. It reads that directly rather than off `body`, because `renderSidebar()` runs from `initializeEditMode()` *before* `edit` is set and `setScale()` puts those classes on `body`. That needed `calculateEditModuleClasses` added to the `Object.assign(window, ...)` list in `main.js` (the editor bundle is a separate dynamic import and only sees what is assigned there). `editModulesAbove` keeps the default - the panel and the room coexist in that layout.

**4. Test leaves `editorState` set** - fixed. Both new tests now go through a shared `setEditorState()` helper and clear the key at the end, and they pin the window size instead of relying on the runner's default. Two more tests came out of points 2 and 3:

- `Edit mode leaves the modules closed once the default has been opened before`
- `Edit mode skips the default module in the narrow-window layout`

Full `tests/testcafe/editor.js` (35 tests) and `npm test` (565) pass locally.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review). The idea is right — a first-time editor should land on *something*, and the empty-state properties panel explains itself well. But the panel opens at the stored default width of **50% of the window**, so on every side-by-side layout the very first thing a new user sees is the room halved (1920×1080: board drops from ~1730px to ~960px wide; 1001×600: room becomes a 500×310 strip) in exchange for a panel that is ~85% empty white space. **The single most impactful improvement: give the first-run default panel a content-sized width (cap it around 380–450px / clamp `--modulesWidth`) instead of inheriting the 50% split, and only let it grow once the user drags it or selects a widget.**

Reviewed live in Chromium against a local server on this branch. Screenshots below.

---

## Big layout improvements

### 1. The default panel takes half the window for two buttons

`editorModulesResizer()` (`client/js/editor/layout.js:160`) defaults `modulesWidth` to `50`, and `renderSidebar()` now opens a module into that 50% slot for people who have never set it. The result is the worst possible first impression of edit mode: the thing you came to edit is now half its size, and the space it lost is blank.

| 1920×1080, this branch | 1920×1080, panel skipped (previous behaviour) |
|---|---|
|  |  |

It gets sharper right above the narrow cutoff. `NARROW_WINDOW_SIZE` is 1000, so at **1001×600** the overlay guard does *not* fire and the room is cut to a 500×310 sliver:



Concretely: when the panel is opened *by default* (not restored, not user-chosen), set `--modulesWidth` to something like `clamp(320px, 30%, 440px)` — or write a first-run `modulesWidth` derived from `window.innerWidth` — so the room stays the dominant element. The resizer is still there for anyone who wants more. This also makes the guard less load-bearing: the reason the narrow layout has to be excluded at all is that the panel is sized as a *split*, not as a *panel*.

### 2. The panel's own primary flow is starved by the panel

Clicking **ADD A NEW WIDGET** — the CTA the PR now shows to everyone — opens `addOverlay`, which is confined to the (now halved) room rectangle. So on desktop the widget palette is squeezed into the left half while half the screen holds an empty properties panel:



On phone portrait it is much worse — the palette lands in a ~200px strip under the panel and the pieces are unreadable:



Either fix #1 (narrower default), or have the add-widget overlay collapse/ignore the module panel while it is open. This is the newcomer's very first click in edit mode, so it is worth making it land somewhere legible.

### 3. Portrait layouts get the panel too, and it fills most of the screen

The guard only skips `editModulesOverlay`. In `editModulesAbove` (phone 390×844, tablet 820×1180) the panel opens and occupies ~58–65% of the viewport:

| phone 390×844 | tablet 820×1180 |
|---|---|
|  |  |

To the PR's credit the room rectangle itself is unchanged there (the panel sits in the letterbox area — compare [phone without panel](https://agent.virtualtabletop.io/reports/ui-review/1534561315576746225/phone-editmode-nopanel.png)), so this is not a size regression. But two consequences are worth handling:

- On phone the panel hides the room's own *"Please rotate your device to landscape orientation and switch to fullscreen"* message, which is the more useful thing for that user to read. Consider extending the skip to `editModulesAbove` when the room area is below some height, or at least to `orientationMismatch`.
- The panel is a tall block of white with two buttons at the top and ~500px of nothing under them. If it stays, let the `editModulesAbove` slot size to its content rather than filling the strip.

---

- **`client/css/editor/sidebarModules.css:44-51` — `.noSelectionButton { width: 100% }`.** At the 50% default width these become ~800px-wide banners with a centred icon+label island; they read as headers rather than buttons and don't match the button sizing anywhere else in the editor. `max-width: 420px` (left-aligned or centred) would fit the app's conventions much better. Only noticeable now because this panel is shown unprompted.
- **Toolbar tooltip collides with the panel title.** At 820×1180 the "Add a new widget. / Hotkey: a" tooltip renders directly over the `Edit widgets` `<h1>` ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1534561315576746225/tablet-addwidget.png)). Pre-existing, but the PR guarantees a heading is sitting under the toolbar on first entry, so it now happens to everyone.
- **No visible way to dismiss the panel.** `SidebarModule` has no close control in the module header (`client/js/editor/sidebarModule.js:64` `addHeader` renders a bare `<h1>`); closing means clicking the already-active sidebar button a second time. That's fine for a panel you deliberately opened, less fine for one that opened itself. A small `×` in the module header would fix the whole class of "how do I get my room back".
- **Sidebar label wrapping (pre-existing, visible in every shot ≥1600px):** `EDIT WIDGETS` and `GAME SETTINGS` wrap to two lines and sit tight against their button bounds while the single-word entries don't — see [desktop Game Settings](https://agent.virtualtabletop.io/reports/ui-review/1534561315576746225/desktop-module-settings.png). Not this PR's doing; flagging since the PR makes the sidebar the first thing a new user studies.

---

- **`client/js/editor/sidebar/properties.js:1844` — "click on one to the left" is layout-dependent and wrong in two of the four layouts I tested.** In `editModulesAbove` (phone/tablet portrait) the room is *below* the panel, not to the left. Since this sentence is now the first instruction a new editor user reads, make it layout-independent: *"You do not have a widget selected. Click a widget in the room to edit it, or:"*.
- **The empty state doesn't say what the icon column is.** The panel explains itself, but the eight-icon sidebar next to it — the thing the PR's description correctly identifies as undiscoverable — is still unlabelled at ≤1600px. One line at the bottom of the empty state ("More editor tools are in the sidebar on the right") would close the loop the PR set out to close.
- **"Open deck editor"** is jargon for someone who has never made a game here; a newcomer can't tell whether that creates cards, imports them, or edits an existing deck. The existing `title`/tooltip convention used elsewhere in the editor would help: *"Design the cards in a deck"*.
- **Nothing signals that the panel is a remembered preference.** Because opening it writes `Edit Widgets → editorModuleTopLeft` into `editorState`, a first-time user who never touches the sidebar now has the panel on every subsequent entry with no idea why. The `×` suggested above doubles as the discoverable "make this stop" affordance.

---

### What I actually exercised

Live in Chromium against this branch on a local server, at **390×844 (phone portrait)**, **820×1180 (tablet portrait)**, **1280×800 (laptop)**, **1920×1080 (desktop)**, plus **900×600** (overlay layout, where the PR skips the default) and **1001×600** (one pixel past the cutoff, where it doesn't).

Overlays opened and screenshotted at all four main sizes: Game Shelf, Players, About, Add-widget palette, and every editor sidebar module (Edit Widgets, History, JSON, Widgets, Debug, Assets, Toolbox, Game Settings). All of them opened, rendered and closed correctly at every size — no clipping, unreachable controls or stuck overlays attributable to this change; the shared overlay CSS is untouched by the diff and behaves as before. Not reached: the deck editor's sub-dialogs (export/import/new-deck), state save/share dialogs and the confirm prompt, all of which the diff doesn't touch.

VTT-AI-Agent:
Thanks - the width point was the right thing to lead with, and it turned out to fix most of the rest by itself. Pushed as 5fe3ba9d.

**1. The default panel took half the window** - fixed. When the panel opens *itself*, `renderSidebar()` now puts `defaultEditorModuleWidth` on `body` and `#editorModules` is sized `min(420px, 35vw)` instead of the stored `--modulesWidth`. It is dropped again by `SidebarModule.openInTarget()` (any module opened or closed by hand) and by the resizer's `mousedown`, so nothing is persisted and a module the user opens still gets the 50/50 split - or whatever width they dragged it to.

| window | room before | room now |
|---|---|---|
| 1920x1080 | 960x600 | [1372x858](https://agent.virtualtabletop.io/reports/pr/1534541516087758929/ui-review2/desktop-1920-default.png) |
| 1280x800 | 640x400 | 824x515 |
| 1001x600 | 500x310 | [615x384](https://agent.virtualtabletop.io/reports/pr/1534541516087758929/ui-review2/edge-1001x600-default.png) |

The `35vw` part is for exactly your 1001x600 case: a fixed 420px is a bigger share of a small window than of a large one. And [with a widget selected](https://agent.virtualtabletop.io/reports/pr/1534541516087758929/ui-review2/desktop-1920-widget-selected.png) the properties panel is still fully usable at that width - every control fits, nothing wraps badly.

**2. The add-widget palette was starved by the panel** - fixed by (1) on desktop: [the palette at 1920](https://agent.virtualtabletop.io/reports/pr/1534541516087758929/ui-review2/desktop-1920-addwidget.png) now has 1372px instead of 960px, which is more than it had before this PR existed. On phone portrait it is fixed by (3) - there is no panel there at all any more. I did not make `addOverlay` ignore the panel: it is anchored to the room rectangle on purpose (that is how the whole overlay system works), and the two cases that made it cramped are gone.

**3. Portrait layouts got the panel too** - fixed. The skip now also covers `isOrientationMismatch()`, i.e. a portrait window showing a landscape board, which is both your phone (390x844) and tablet (820x1180) case. [Phone at 390x844](https://agent.virtualtabletop.io/reports/pr/1534541516087758929/ui-review2/phone-390x844-skipped.png): the room's "Please rotate your device..." message is visible again, sidebar as before. That needed `isOrientationMismatch` added to the `Object.assign(window, ...)` list in `main.js` next to `calculateEditModuleClasses`. `editModulesAbove` with a *portrait* board keeps the default - there the panel sits beside/above the room without covering anything and the room is the shape the window is.

**4. `.noSelectionButton { width: 100% }`** - fixed, `max-width: 420px` added. They are normal buttons again at any panel width.

**5. No way to dismiss the panel** - fixed. `SidebarModule.addCloseButton()` puts an `x` in the top right of the module header; `addHeader()` calls it for every module that has a title (History, Widgets, Toolbox, Game Settings, Edit Widgets), and the properties module also calls it on its widget header so the button does not disappear the moment you select something ([screenshot](https://agent.virtualtabletop.io/reports/pr/1534541516087758929/ui-review2/module-header-close.png)). JSON, Debug and Assets render no header at all, so they have no place to put one - left alone rather than inventing a header for them.

**6. "click on one to the left"** - fixed: *"You do not have a widget selected. To get started, click a widget in the room or:"*.

**7. The empty state doesn't say what the icon column is** - added one line under the buttons: *"The other editor tools are in the sidebar on the right."* (The sidebar is on the right in every layout.)

**8. "Open deck editor" is jargon** - both buttons got a `title`: *"Pick a card, dice, board or other piece to place in the room"* and *"Design the cards in a deck"*. I left the labels alone, since "deck editor" is what the toolbar button and the wiki call it.

**Not changed:**
- *Toolbar tooltip collides with the panel title* - pre-existing and it is the toolbar's tooltip layer, not this panel; with the narrower default panel the collision area is smaller but it can still happen. Worth its own fix in the tooltip positioning rather than here.
- *Sidebar label wrapping (`EDIT WIDGETS` / `GAME SETTINGS`)* - pre-existing and unrelated to when the panel opens; one PR, one concern.

Tests: two more in `tests/testcafe/editor.js` (portrait skip, closing through the header button) and the first one now also asserts the content width and that opening a module by hand hands the width back. `Dice faces...` resizes to 410x845, which is now one of the skipped layouts, so it opens the module explicitly again. Full `editor.js` (37 tests) and `npm test` (565) pass locally.

VTT-AI-Agent:
CI failure fixed in e83caa5e (after merging `main`).

`main` gained `A pile's drop limit is set on the pile and lands in its template` (from #3079) while this branch was open. It clicks `#editorSidebar [icon=tune]` right after entering edit mode - which, with this PR, toggles the module that just opened by *itself* closed again, so `Drop limit` was never visible. It now waits for `#editorModuleTopLeft.tune` instead, the same way the other tests in the file do since this PR.

Full `tests/testcafe/editor.js` (38 tests) and `npm test` (590) pass locally on the merged tree.

VTT-AI-Agent (commented):
## TL;DR

This looks good after the follow-up fixes: the first-run properties panel is now discoverable without overriding a user's later choice to keep modules closed, and the responsive/portrait guards prevent it from covering the room. I found no remaining correctness, security, regression, or authoring-UI issue worth raising.

## Detailed review

No actionable findings.

I re-checked the state lifecycle in `client/js/editor/layout.js` and `client/js/editor/sidebarModule.js`: the default is recorded only after `openInTarget()` persists the module, closing still deletes the module entry while retaining `defaultModuleOpened`, and any manual module toggle or resize removes the temporary content-width class. The narrow-window and orientation-mismatch checks use the same layout helpers as the editor, and the updated TestCafe coverage exercises the fresh, remembered, deliberately closed, narrow, portrait, width-handoff, and header-close cases. The current CI, CodeQL, and Chrome/Firefox editor runs are green.

This PR exposes no new widget property or routine operation, so it does not require an additional property-editor or visual routine-block control.

The `beta` label was already present; I left it in place because the diff is confined to edit-mode UI, CSS, and tests, touches no engine paths, and is safe and useful for player testing.

VTT-AI-Agent:
Thanks for the review (automated review 3 of 4). No changes needed on my side: the review reports no actionable findings, and its reading of the state lifecycle matches the intent — the default-open flag is recorded only once `openInTarget()` has persisted the module, closing the module keeps `defaultModuleOpened` set so it stays closed on later entries, and any manual toggle or resize drops the temporary content-width class. CI (build, CodeQL, Chrome/Firefox TestCafe incl. the extended `editor.js` coverage) is green on the current head, so I'm leaving the branch as it is.

On the `beta` label: I did not add it and I'm not touching it — that call belongs to a human reviewer.

VTT-AI-Agent (commented):
## TL;DR

Good, well-scoped change with careful layout guards and solid test coverage; the `rand()`→`editorDomID()` fix in the Automations switch is a nice catch on its own. The main issue is that the content-width treatment only survives the first page load: because the auto-opened module is written into `editorState.modules`, every later entry into edit mode restores that same never-chosen panel at the 50/50 `--modulesWidth`, which is exactly the "half the room gone for a mostly empty panel" the PR sets out to avoid.

## Findings

### 1. The content width applies only on the very first load (`client/js/editor/layout.js:140-152`, `client/css/editor/sidebar.css:88`)

`openInTarget()` → `saveToLocalStorage(target)` persists `modules['Edit widgets'] = 'editorModuleTopLeft'` for the panel that opened *itself*. So on the next page load `renderSidebar()` takes the restore branch (`opened = true`), `body.defaultEditorModuleWidth` is never added, and the panel comes up at `calc(var(--modulesWidth) - var(--editSidebarWidth))` — 50% of the window at 1920x1080, i.e. the 960x600 room the PR description explicitly calls a poor first sight. The user did nothing: they never opened a module by hand and never touched the resizer, which are the two events the PR names as handing the width back.

Concrete sequence: fresh `localStorage` → Edit → panel at ~420px, room 1372x858 → F5 → Edit → same panel, room 960x600.

If that isn't intended, the flag needs the same treatment as `defaultModuleOpened`: persist e.g. `editorState.defaultModuleWidth = true` next to it, re-add the body class in the restore branch while it is set, and `delete` it in the two places the class is dropped today (`SidebarModule.openInTarget()` at `sidebarModule.js:192` and the resizer's `onmousedown` at `layout.js:178`).

### 2. The close button reaches only the modules that use `addHeader()` (`client/js/editor/sidebarModule.js:64-80`)

Hooking the button into `addHeader(text, target)` under `if(!target && this.moduleDOM)` covers Properties (`properties.js:1843`), Widgets, Toolbox, Game Settings and History — but the JSON module (`sidebar/json.js:62`), the Log/debug module and the Assets module (`sidebar/assets.js:386`) build their DOM without any `h1`, so they still have no way out other than knowing the sidebar button toggles. That's the same discoverability gap the PR is fixing, and the JSON module is the panel a new user is most likely to land in right after the default one (it replaces it in the same slot). It also makes the affordance depend on an implementation detail — a module that renders its header into a sub-container silently loses the button.

Adding the button in `openInTarget()`'s open branch (appended to `target`, absolutely positioned in the module's top-right rather than inside the `h1`) would cover every module uniformly and let `addHeader()` go back to just rendering a header.

### 3. ~15 existing tests now depend on window size and on a clean `localStorage` they don't set (`tests/testcafe/editor.js`)

The `.click('#editorSidebar [icon=tune]')` → `.expect(Selector('#editorModuleTopLeft.tune').exists).ok()` conversions work in any layout in the old form, but the new assertion only holds when (a) `editorState` is empty and (b) the window is in the beside-the-room layout. Most of the converted tests set neither: `A holder picks what it accepts…`, `Position holds the grid…`, `A pile is edited through its handle…`, all six `Deck editor: …` conversions and `Basic curates the stacking…` have no `resizeWindow()` of their own and inherit whatever the previously executed test left behind — which now includes a 410x845 portrait window and a 900x600 one from the new tests. It passes in file order today, but running a single test with `-t`, reordering, or `-c`/concurrency makes it size-dependent, and the failure ("panel not open") won't point at the window.

Two cheap hardenings: put `setEditorState(null)` in the fixture's `beforeEach` in `test-util.js` rather than at the end of each new test (a new test failing mid-way currently leaves `defaultModuleOpened: true` behind and cascades into every converted test), and give the converted tests an explicit `resizeWindow()` — `Dice faces…` already shows why this matters.

### 4. Minor: a click on the resizer without a drag jumps the panel to 50%

`layout.js:178` removes `defaultEditorModuleWidth` on `onmousedown`, before any movement. Mousing down on the divider and releasing without moving widens the panel from ~420px to half the window and writes `modulesWidth` on the following `mouseup`. Moving the `classList.remove` into `resize()` would tie it to an actual drag.

### 5. Nit: the "no selection" buttons stay left-aligned in a wide panel

`client/css/editor/sidebarModules.css:63` caps them at 420px but keeps `margin: 8px 0`, so in a 50%-wide panel they hug the left edge under a full-width intro paragraph. `margin: 8px auto` centers them.

---

Nothing here touches the engine (no `client/js/widgets/`, `compute.js` or `serverstate.js`; the `client/js/main.js` change only adds two existing layout helpers to the editor bundle's `window` exports), the change is edit-mode-only and cannot corrupt a game or crash the server, so the PR is beta-eligible and the `beta` label was already present — I left it in place.

VTT-AI-Agent:
Thanks - finding 1 was real and it was the important one. Fixed in b5b23fb3, together with 2, 4 and 5.

**1. The content width applied only on the very first load** - fixed. `editorState.defaultModuleWidth` is now remembered next to `defaultModuleOpened`, and `renderSidebar()` re-adds `body.defaultEditorModuleWidth` in the restore branch while it is set. It is dropped by exactly the two events that dropped the body class before - a module opened or closed by hand, and a drag of the resizer - so nothing changes for a user who has picked a width. Since `openInTarget()` is also what `renderSidebar()` itself calls to restore a module, the drop is now a `dropDefaultModuleWidth()` helper that no-ops while the sidebar is being rendered.

Measured in Chromium at 1920x1080, fresh `localStorage`, entering edit mode and then reloading:

| | `#editorModules` | room | stored |
|---|---|---|---|
| first entry | 420px | 1372x858 | `defaultModuleOpened`, `defaultModuleWidth` |
| after F5 | 420px | 1372x858 | unchanged |
| after dragging the resizer | 620px | 1172x733 | `modulesWidth: 38.96`, no `defaultModuleWidth` |
| after F5 | 620px | 1172x733 | unchanged |

[The panel](https://agent.virtualtabletop.io/reports/pr/1534541516087758929/review4/default-panel.png) - same on the first entry and on every one after it now.

**2. The close button reached only the modules that use `addHeader()`** - fixed, though not quite the way you suggested. Adding it in `openInTarget()` unconditionally does not survive: the properties module wipes `moduleDOM.innerHTML` on every selection change, and Widgets and Game Settings wipe `target.innerHTML` when they re-render, so the button would disappear on the first interaction with them. `addHeader()` therefore keeps adding it (that is also what puts it back on those re-renders), and `openInTarget()` adds it to the module itself only when the module rendered no header - which is exactly JSON, Debug and Assets. All three verified open and close through the new button.

Two of them needed the corner to be made free: the JSON module's top right corner is the command pane, so there the button sits at the right end of the widget switcher bar (`right: calc(var(--commandPaneWidth) + 5px)`), and the Debug module's filter row is right-aligned and had the button sitting on top of its `Clear` button - [it now reserves the space](https://agent.virtualtabletop.io/reports/pr/1534541516087758929/review4/debug-close-button.png).

**3. Tests depending on window size and on a clean `localStorage`** - fixed for every test that expects the properties module: each one now sets its own `resizeWindow()` and its own `setEditorState(null)` right after `prepareClient()`, with a comment at the top of the file saying why. I did not put the reset into the shared `beforeEach` in `test-util.js`: that fixture is shared with every other TestCafe file, and clearing the editor state for all of them changes which modules `boardsize.js` comes up with - `The editor sidebar keeps its labels while the board is not short of the space` measures `--editSidebarWidth` and depends on how much width the board has next to the panel. The sizes I added are the ones each test already ran at through the file order, so this only removes the dependency, it does not change what is being tested.

`npx testcafe ... tests/testcafe/editor.js -t "A pile is edited through its handle, css through declaration rows"` now passes on its own, which is the case you named.

**4. A click on the resizer without a drag jumped the panel to 50%** - fixed, `dropDefaultModuleWidth()` moved from `onmousedown` into `resize()`. Verified: mousedown+mouseup on the divider leaves the panel at 420px.

**5. The "no selection" buttons stayed left-aligned in a wide panel** - fixed, `margin: 8px auto`. [At a 50%-wide panel](https://agent.virtualtabletop.io/reports/pr/1534541516087758929/review4/noselection-centered.png).

Tests: `npm test` 590 passed, `tests/testcafe/editor.js` 38 passed, and `tests/testcafe/boardsize.js` 7 passed (the file the shared-fixture reset would have affected).

ArnoldSmith86 (approved):

PR #3102, opened 2026-08-05
Remove the split-session buttons and make Add player a row of the players table

### LINKED ISSUE #754: Changing your name causes your hand to disappear
1. Visit virtualtabletop.io and be assigned a new room.
2. Add Poker
3. Hit Poker's "Play" button.
4. Drag a card from the deck into your hand at the bottom.
5. Press the "players" button (looks like stylized busts of two people)
6. Click your name
7. Move to the end and add "X"
8. Hit enter
9. Press the "players" button to dismiss the screen.

Observed behavior: card in hand disappears

Expected behavior: card does not disappear

At this point, changing your name back to the original causes the card to re-appear.

I believe what is happening is:

- The hand is set `"childrenPerOwner":true`
- When I dragged the card into the hand, the card's `"owner"` was changed to my name.
- When I changed my name, the card's owner was not updated, so no one could see the card.

I would expect that if I change my name, all `"owner"`s would searched and instances of my name before be swapped to my new name. (Or perhaps assign players arbitrary IDs that don't change.)

We ran into this while playing a game when a player realized they could change their name and did so. As their old name was in the form `Guest123`, we didn't know what name to change them back to to fix it.

VTT-AI-Agent:
Simplifies the reworked Players overlay from #3013: the per-session **split** button is removed, and *Add player* and *Invite players* become the last two rows of the players table, with everything explanatory folded into one collapsed help section. Based on #3013 (`players-tab-table`), so please merge that first.

## What changed

- **Split buttons removed.** The ⑂ button in every session cell (and its legend entry and the mention in the overlay's intro text) is gone. It duplicated what *View* already does for hotseat play — with the extra cost of a modal `prompt()` — and it is the one action people found hard to tell apart from *View*. The session cell is now just the label, so the column is plain text again.
- **Add player is a table row.** The name input and the add button moved into a `<tfoot>` row of the players table (person icon, *Add a player who shares this device* input, `+` button), so adding a player looks and lines up like every other row instead of living in a separate block underneath. The input carries an underline so it reads as a field.
- Adding an empty or already-existing name now reports the problem directly at the input (`setCustomValidity`/`reportValidity`, plus a red underline) instead of turning the button into a red error label, and the button shows the same inline spinner as the other row actions while the server request is in flight.
- **Adding a player switches to it when you have several tabs open.** While the current player has more than one connection open, adding a player hands the current browser tab over to that player (a single `rename` of this session) instead of only listing the name - that is the case the removed split button covered. With a single connection nothing changes. The add button tooltip and the help say when the switch happens.
- **Invite is a table row too.** Below the add-player row, a second footer row reads *Invite players to* followed by the room URL as plain text to read out or select, and puts the share button in the same column as the row action buttons. That button shares the URL (or copies it to the clipboard), and the result — or the error — is shown in that row instead of inside the icon-sized button.
- **Collapsed help instead of intro text and legend.** The paragraph the overlay used to open with, the explanatory part of the old invite block and the icon legend all moved into two framed collapsible sections below the table: *Help: playing together*, split into three topics that each get a large icon and a title (*Players and connections*, *Remote play*, *Hotseat*), and *Help: what the buttons do*, which lists every button of the table. Both start collapsed and only open when you click them, so the overlay is the table and nothing else until you ask for an explanation.
- **Any disconnected player can be removed.** The server used to refuse the removal while any widget still named the player as `owner`, seat `player` or `artist`, and the client hid the trash button to match — so an old name that only lived on in an invisible seat or artist credit could never be cleared out. Removal is now allowed for everybody who is not connected; when something in the game still points at them, the button asks first and names what will be left behind. Nothing is deleted with the player, so a player of the same name picks those widgets up again. For connected players the button is still hidden, since the connections column of the same row already says why.
- The connection labels wrap below 530px, which keeps the table inside a 390px phone viewport instead of overflowing it.
- **"Sessions" is called "Connections" now.** The word only made sense to people who know what a session is; what the column actually lists is the browser tabs that have this room open. The header, its tooltip and the numbered labels of a player with several tabs (*Connection 1*, *Connection 2*) use "connection", and so does the help text: *Every browser tab that has this room open is one connection of a player.* Nothing about the *View* / hotseat wording talks about sessions any more either.
- **The toolbar icon lost its gear.** The Players tab used `[users_settings]` (a person with a gear, which reads as a settings screen); it is `[people]` now.

The server side is unchanged — `rename` with a `sessionID` is still what *View* uses, so no code became dead.

Alone in a fresh room:

Players overlay for a first-time visitor

With other players at the table:

Players overlay with three players

Every disconnected player has a remove button, and removing one the game still points at asks first:

Players table with a remove button on every disconnected player
Warning before removing a player who still owns widgets

The help section expanded, and the result of sharing the room URL:

Expanded help section
Share result in the invite row

The renamed column, with one player who has the room open in two tabs (the screenshots above still show the older *Sessions* header):

Players table with a Connections column listing Connection 1 and Connection 2
Help section talking about connections

## Verified

`npm test` (15/15) and the TestCafe editor test (which drives this overlay through `setName`) pass locally. Scripted browser checks confirmed: add local player, duplicate/empty name feedback, view another player, rename with reference updates, no split button left in the DOM, the two-tab add-and-switch behaviour (and that a single session still just adds), both help sections staying collapsed until clicked, removing a player with and without the warning (confirm, cancel, and no dialog at all when nothing points at them), the share status and clipboard content, the invite link not navigating, and the table fitting a 390px viewport.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3102/pr-test (or any other room on that server)

### DIFF
--- a/client/css/layout.css
+++ b/client/css/layout.css
@@ -462,7 +462,7 @@ body.aspectTooGood.hiddenToolbar {
     }
 
     #playersButton::before {
-      content: '[users_settings]';
+      content: '[people]';
     }
 
     #editButton::before {
--- a/client/css/overlays/players.css
+++ b/client/css/overlays/players.css
@@ -46,17 +46,6 @@
   font-size: 0.8rem;
 }
 
-#playersTable .sessionCellContent {
-  display: flex;
-  align-items: center;
-  gap: 0.5rem;
-}
-
-/* the session cell is only as wide as its widest label, so this aligns the buttons without a large gap */
-#playersTable .sessionCellContent button {
-  margin-left: auto;
-}
-
 #playersTable button[icon] {
   display: inline-flex;
   padding: 5px;
@@ -156,6 +145,19 @@
   font-size: 1.2rem;
 }
 
+/* the row that adds a local player: same layout as a player row, but nothing exists yet */
+#addPlayerRow .playerColorCell .material-symbols {
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  width: 2em;
+  height: 2em;
+  opacity: 0.6;
+}
+#addPlayerRow .playerName {
+  font-weight: normal;
+}
+
 #playersLegend {
   display: flex;
   flex-wrap: wrap;
@@ -180,26 +182,16 @@
   margin: 1rem auto 0 auto;
 }
 
-#playerAdd {
+#playerInvite {
   max-width: 40rem;
   margin: 1rem auto;
 }
-#playerAdd h1 {
+#playerInvite h1 {
   text-align: left;
 }
-#playerAdd .material-symbols {
+#playerInvite .material-symbols {
   font-size: 1em;
   vertical-align: text-bottom;
-}
-#addLocalPlayer {
-  display: flex;
-  align-items: center;
-  gap: 0.25rem;
-  margin: 0.5rem 0;
-}
-#localPlayerName {
-  font-size: 1rem;
-  padding: 5px;
 }
 
 .overlay .heading{
--- a/client/css/overlays/players.css
+++ b/client/css/overlays/players.css
@@ -146,12 +146,9 @@
 }
 
 /* the rows that add a local player and invite remote ones: same layout as a player row, but no player of their own */
-#playersTable tfoot .playerColorCell .material-symbols {
-  display: inline-flex;
-  align-items: center;
-  justify-content: center;
-  width: 2em;
-  height: 2em;
+/* they reuse the wrap of the color circle so their icon sits in a box of the same size */
+#playersTable tfoot .teamColorWrap .material-symbols {
+  font-size: 1.5em;
   opacity: 0.6;
 }
 #addPlayerRow .playerName {
--- a/client/css/overlays/players.css
+++ b/client/css/overlays/players.css
@@ -242,14 +242,26 @@
   padding: 0;
   list-style: none;
 }
+/* the button list uses the icon column of the topics above so that both sections indent
+   their text by the same amount - only the icons themselves are smaller */
+.playersHelp li {
+  display: flex;
+  align-items: flex-start;
+  gap: 0.75rem;
+}
 .playersHelp li + li {
-  margin-top: 0.25rem;
+  margin-top: 0.5rem;
+}
+.playersHelp li > .material-symbols {
+  font-size: 20px;
+  flex: 0 0 32px;
+  text-align: center;
+  color: var(--VTTblue);
 }
 /* an inline icon is placed by its own box instead of by text-bottom: with a fixed height and a
    hidden overflow its baseline is the bottom edge of that box, so the offset below the text
    baseline is a plain length and no longer depends on how a browser reads the font metrics */
-.playersHelp p .material-symbols,
-.playersHelp li .material-symbols {
+.playersHelp p .material-symbols {
   font-size: 1em;
   height: 1em;
   line-height: 1em;
--- a/client/css/overlays/players.css
+++ b/client/css/overlays/players.css
@@ -181,7 +181,6 @@
 /* the room URL is text to read out or select, not a link - clicking it would only reload the room.
    breaking it anywhere keeps it fully readable without widening the table on narrow screens */
 #playerInviteURL {
-  font-weight: bold;
   overflow-wrap: anywhere;
 }
 /* the share button lines up with the row action buttons and has no room for a label,
--- a/client/css/overlays/players.css
+++ b/client/css/overlays/players.css
@@ -46,6 +46,14 @@
   font-size: 0.8rem;
 }
 
+/* on a phone the session labels are what pushes the table past the viewport, so let them
+   wrap there rather than making the whole overlay scroll sideways */
+@media (max-width: 530px) {
+  #playersTable .playerSessionCell {
+    white-space: normal;
+  }
+}
+
 #playersTable button[icon] {
   display: inline-flex;
   padding: 5px;
@@ -151,31 +159,34 @@
   font-size: 1.5em;
   opacity: 0.6;
 }
+/* an empty input has no value to read, so unlike the player names it needs to look like a field */
 #addPlayerRow .playerName {
   font-weight: normal;
+  box-sizing: border-box;
+  min-width: 0;
+  border-bottom: 1px solid var(--VTTblueLight);
+}
+#addPlayerRow .playerName:invalid {
+  border-bottom-color: darkred;
 }
 #invitePlayerRow .playerNameCell {
   font-size: 0.9rem;
 }
-/* the share button lines up with the row action buttons, so the progress text that
-   progressButton writes into it has to stay out of the layout - its icon shows the state */
-#playersShareButton {
-  font-size: 0;
+/* breaking the URL anywhere keeps it fully readable without widening the table on narrow screens */
+#playerInviteURL {
+  overflow-wrap: anywhere;
 }
-
-#playersLegend {
-  display: flex;
-  flex-wrap: wrap;
-  gap: 0.25rem 1.5rem;
-  max-width: 40rem;
-  margin: 0 auto;
-  font-size: 0.8rem;
-  opacity: 0.8;
+/* the share button lines up with the row action buttons and has no room for a label,
+   so the result of sharing replaces the invite text of the row */
+#playerInviteStatus,
+#invitePlayerRow.showStatus #playerInviteText {
+  display: none;
 }
-#playersLegend .material-symbols {
-  font-size: 1.2em;
-  vertical-align: text-bottom;
-  margin-right: 0.25rem;
+#invitePlayerRow.showStatus #playerInviteStatus {
+  display: inline;
+}
+#playerInviteStatus.error {
+  color: darkred;
 }
 
 #playersHelp {
@@ -205,14 +216,24 @@
 #playersHelp p {
   margin: 0;
 }
-#playersHelp p .material-symbols {
+#playersHelp ul {
+  margin: 0;
+  padding: 0;
+  list-style: none;
+}
+#playersHelp li {
+  margin-top: 0.25rem;
+}
+#playersHelp p .material-symbols,
+#playersHelp li .material-symbols {
   font-size: 1em;
   vertical-align: text-bottom;
+  margin-right: 0.25rem;
 }
 
 .overlay .heading{
   width: 100%;
-  max-width: 40rem; /* same width as the table, legend and invite block below */
+  max-width: 40rem; /* same width as the table and the help section below */
   margin: 0 auto;
 }
 .overlay .heading h1{
--- a/client/css/overlays/players.css
+++ b/client/css/overlays/players.css
@@ -145,8 +145,8 @@
   font-size: 1.2rem;
 }
 
-/* the row that adds a local player: same layout as a player row, but nothing exists yet */
-#addPlayerRow .playerColorCell .material-symbols {
+/* the rows that add a local player and invite remote ones: same layout as a player row, but no player of their own */
+#playersTable tfoot .playerColorCell .material-symbols {
   display: inline-flex;
   align-items: center;
   justify-content: center;
@@ -156,6 +156,14 @@
 }
 #addPlayerRow .playerName {
   font-weight: normal;
+}
+#invitePlayerRow .playerNameCell {
+  font-size: 0.9rem;
+}
+/* the share button lines up with the row action buttons, so the progress text that
+   progressButton writes into it has to stay out of the layout - its icon shows the state */
+#playersShareButton {
+  font-size: 0;
 }
 
 #playersLegend {
@@ -180,16 +188,6 @@
   display: block;
   max-width: 40rem;
   margin: 1rem auto 0 auto;
-}
-
-#playerInvite {
-  width: 100%;
-  max-width: 40rem;
-  margin: 1rem auto;
-}
-#playerInvite h1 {
-  text-align: left;
-  margin-bottom: 0.5em;
 }
 
 #playersHelp {
--- a/client/css/overlays/players.css
+++ b/client/css/overlays/players.css
@@ -178,15 +178,6 @@
   margin-right: 0.25rem;
 }
 
-#playersAloneHint {
-  display: none;
-}
-#playersAloneHint.shown {
-  display: block;
-  max-width: 40rem;
-  margin: 1rem auto 0 auto;
-}
-
 #playersHelp {
   width: 100%;
   max-width: 40rem;
--- a/client/css/overlays/players.css
+++ b/client/css/overlays/players.css
@@ -183,38 +183,55 @@
 }
 
 #playerInvite {
+  width: 100%;
   max-width: 40rem;
   margin: 1rem auto;
 }
 #playerInvite h1 {
   text-align: left;
+  margin-bottom: 0.5em;
 }
-#playerInvite .material-symbols {
+
+#playersHelp {
+  width: 100%;
+  max-width: 40rem;
+  margin: 1rem auto;
+}
+#playersHelp > summary {
+  cursor: pointer;
+  font-weight: bold;
+}
+#playersHelp .helpTopic {
+  display: flex;
+  align-items: flex-start;
+  gap: 0.75rem;
+  margin-top: 1rem;
+}
+#playersHelp .helpTopic > .material-symbols {
+  font-size: 32px;
+  flex: 0 0 32px;
+  color: var(--VTTblue);
+}
+#playersHelp h2 {
+  margin: 0 0 0.25rem 0;
+  font-size: 1.1rem;
+}
+#playersHelp p {
+  margin: 0;
+}
+#playersHelp p .material-symbols {
   font-size: 1em;
   vertical-align: text-bottom;
 }
 
 .overlay .heading{
-  display: flex;
-  align-items: flex-start;
-  justify-content: flex-start;
   width: 100%;
-  max-width: 40rem; /* same width as the table, legend and add-player block below */
+  max-width: 40rem; /* same width as the table, legend and invite block below */
   margin: 0 auto;
 }
 .overlay .heading h1{
-  margin: 0 0.5em 0 0;
-  padding: 0.25em 0.5em 0.25em 0;
-  border-right: 1px solid rgba(255, 255, 255, 0.5);
-  line-height: 1em;
+  margin: 0;
   text-align: left;
-  max-width: 45%;
-}
-.overlay .heading .instructionalText{
-  display: block;
-  font-size: 16px;
-  text-align: left;
-  padding: 0.9rem 0;
 }
 #playerCursors {
   position: absolute;
--- a/client/css/overlays/players.css
+++ b/client/css/overlays/players.css
@@ -169,6 +169,12 @@
 #addPlayerRow .playerName:invalid {
   border-bottom-color: darkred;
 }
+/* the invite row is a hint of the same kind as the add-player placeholder, so it gets its color
+   (spelled out because browsers disagree on the default placeholder color) */
+#addPlayerRow .playerName::placeholder,
+#invitePlayerRow .playerNameCell {
+  color: #757575;
+}
 #invitePlayerRow .playerNameCell {
   font-size: 0.9rem;
 }
@@ -191,43 +197,57 @@
   color: darkred;
 }
 
-#playersHelp {
+/* framed so that it is obvious where the collapsed section starts and what belongs to it */
+.playersHelp {
   width: 100%;
   max-width: 40rem;
-  margin: 1rem auto;
+  box-sizing: border-box;
+  margin: 0.5rem auto;
+  border: 1px solid var(--VTTblueLight);
+  border-radius: 4px;
 }
-#playersHelp > summary {
+.playersHelp > summary {
   cursor: pointer;
   font-weight: bold;
+  padding: 0.5rem;
+  background: #1f5ca61a;
 }
-#playersHelp .helpTopic {
+.playersHelp[open] > summary {
+  border-bottom: 1px solid var(--VTTblueLight);
+}
+.playersHelp .helpContent {
+  padding: 0.5rem 0.75rem 0.75rem 0.75rem;
+}
+.playersHelp .helpTopic {
   display: flex;
   align-items: flex-start;
   gap: 0.75rem;
+}
+.playersHelp .helpTopic + .helpTopic {
   margin-top: 1rem;
 }
-#playersHelp .helpTopic > .material-symbols {
+.playersHelp .helpTopic > .material-symbols {
   font-size: 32px;
   flex: 0 0 32px;
   color: var(--VTTblue);
 }
-#playersHelp h2 {
+.playersHelp h2 {
   margin: 0 0 0.25rem 0;
   font-size: 1.1rem;
 }
-#playersHelp p {
+.playersHelp p {
   margin: 0;
 }
-#playersHelp ul {
+.playersHelp ul {
   margin: 0;
   padding: 0;
   list-style: none;
 }
-#playersHelp li {
+.playersHelp li + li {
   margin-top: 0.25rem;
 }
-#playersHelp p .material-symbols,
-#playersHelp li .material-symbols {
+.playersHelp p .material-symbols,
+.playersHelp li .material-symbols {
   font-size: 1em;
   vertical-align: text-bottom;
   margin-right: 0.25rem;
--- a/client/css/overlays/players.css
+++ b/client/css/overlays/players.css
@@ -246,10 +246,16 @@
 .playersHelp li + li {
   margin-top: 0.25rem;
 }
+/* an inline icon is placed by its own box instead of by text-bottom: with a fixed height and a
+   hidden overflow its baseline is the bottom edge of that box, so the offset below the text
+   baseline is a plain length and no longer depends on how a browser reads the font metrics */
 .playersHelp p .material-symbols,
 .playersHelp li .material-symbols {
   font-size: 1em;
-  vertical-align: text-bottom;
+  height: 1em;
+  line-height: 1em;
+  overflow: hidden;
+  vertical-align: -0.25em;
   margin-right: 0.25rem;
 }
 
--- a/client/css/overlays/players.css
+++ b/client/css/overlays/players.css
@@ -172,8 +172,10 @@
 #invitePlayerRow .playerNameCell {
   font-size: 0.9rem;
 }
-/* breaking the URL anywhere keeps it fully readable without widening the table on narrow screens */
+/* the room URL is text to read out or select, not a link - clicking it would only reload the room.
+   breaking it anywhere keeps it fully readable without widening the table on narrow screens */
 #playerInviteURL {
+  font-weight: bold;
   overflow-wrap: anywhere;
 }
 /* the share button lines up with the row action buttons and has no room for a label,
--- a/client/js/domhelpers.js
+++ b/client/js/domhelpers.js
@@ -64,16 +64,19 @@ export function progressButton(button, clickHandler, disableWhenDone=true) {
   };
 }
 
-async function shareURL(url) {
+// returns how the URL was passed on so callers can tell the user what happened
+export async function shareURL(url) {
   try {
     await navigator.share({ url });
   } catch(e) {
     try {
       await navigator.clipboard.writeText(url);
+      return 'clipboard';
     } catch(e) {
       throw new Error('Could not share or copy URL.');
     }
   }
+  return 'share';
 }
 
 // uses a progressButton with a custom handler that shares a URL
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -680,8 +680,7 @@ onLoad(function() {
     checkedOnce = true;
     let tabSuffix = config.customTab || config.serverName || 'VirtualTabletop.io';
     document.title = `${document.location.pathname.split('/').pop()} - ${tabSuffix}`;
-    const inviteLink = $('#playerInviteURL');
-    inviteLink.href = inviteLink.innerText = location.href;
+    $('#playerInviteURL').innerText = location.href;
   });
 });
 
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -681,7 +681,7 @@ onLoad(function() {
     let tabSuffix = config.customTab || config.serverName || 'VirtualTabletop.io';
     document.title = `${document.location.pathname.split('/').pop()} - ${tabSuffix}`;
     const inviteLink = $('#playerInviteURL');
-    inviteLink.href = inviteLink.title = location.href;
+    inviteLink.href = inviteLink.innerText = location.href;
   });
 });
 
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -680,7 +680,8 @@ onLoad(function() {
     checkedOnce = true;
     let tabSuffix = config.customTab || config.serverName || 'VirtualTabletop.io';
     document.title = `${document.location.pathname.split('/').pop()} - ${tabSuffix}`;
-    $('#playerInviteURL').innerText = location.href;
+    const inviteLink = $('#playerInviteURL');
+    inviteLink.href = inviteLink.title = location.href;
   });
 });
 
--- a/client/js/overlays/players.js
+++ b/client/js/overlays/players.js
@@ -196,6 +196,10 @@ function fillPlayerList(players, active, sessions) {
     if(player != playerName && activePlayers.indexOf(player) != -1)
       addPlayerCursor(player, players[player]);
   }
+  // with a second tab open as the same player, adding a player also moves this tab to it
+  $('#addLocalPlayerButton').title = (sessionsByPlayer[playerName] || []).length > 1
+    ? 'Add a player and switch this browser tab to them'
+    : 'Add a player who shares this device';
   // somebody who is alone at the table is usually here to find out how to get others in, so
   // the help starts open for them - once there are other players it would just be in the way
   if(!helpDefaultApplied) {
@@ -290,6 +294,13 @@ onLoad(function() {
       input.setCustomValidity(localPlayerName ? 'This player already exists.' : 'Please enter a player name.');
       input.reportValidity();
       return;
+    }
+    // a second tab connected as the same player is somebody who wants to be their own player, so
+    // the new player takes this session over right away instead of waiting for a View click
+    if((lastMetaArgs && lastMetaArgs.sessions || []).filter(s=>s.player == playerName).length > 1) {
+      // renaming a single session creates the player if it does not exist yet
+      toServer('rename', { oldName: playerName, newName: localPlayerName, sessionID: mySessionID });
+      return nextMetaUpdate(args=>(args.sessions || []).some(s=>s.sessionID == mySessionID && s.player == localPlayerName)).then(_=>input.value = '');
     }
     toServer('addLocalPlayer', { player: localPlayerName });
     return nextMetaUpdate(args=>args.meta.players[localPlayerName] !== undefined).then(_=>input.value = '');
--- a/client/js/overlays/players.js
+++ b/client/js/overlays/players.js
@@ -1,4 +1,4 @@
-import { asArray, onLoad, progressButton, rand } from '../domhelpers.js';
+import { asArray, onLoad, rand } from '../domhelpers.js';
 
 let playerCursors = {};
 let playerCursorsTimeout = {};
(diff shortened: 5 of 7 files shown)


VTT-AI-Agent (commented):
## TL;DR

Solid simplification — I drove the overlay in a real browser (add, duplicate/empty feedback, multi-session rowspan, three players) and everything works as described. One real (cosmetic but sticky) bug: the custom validity set on `#localPlayerName` is only cleared by an `input` event, so after a duplicate error the field stays `:invalid` (red, stale "This player already exists." message) even after a *successful* add.

## Findings

### 1. Stale `setCustomValidity` survives a successful add — `client/js/overlays/players.js:271`

The message is only reset by the `input` listener on line 278, so any path that reaches a valid add without the user typing leaves the input permanently `:invalid`. `client/css/layout.css:1831` styles `input:invalid { color: red }`, so the field (and anything typed into it next) renders red until the next keystroke.

Reproduced in a live room:
1. add `Bob` → ok
2. type `Bob` again, press `+` → "This player already exists." (correct)
3. delete `Bob` with the 🗑 button
4. press `+` again without touching the input

Result: the player *is* added and the input is cleared, but
`checkValidity() === false`, `validationMessage === "This player already exists."`, `color: rgb(255,0,0)`. Same stale state persists if the user just closes and reopens the overlay after a failed attempt.

One-line fix — clear at the start of the handler rather than only on `input`:

```js
serverActionButton($('#addLocalPlayerButton'), function() {
  const input = $('#localPlayerName');
  input.setCustomValidity('');
  ...
```

(the `input` listener can stay, so the red state also goes away as soon as the user starts correcting the name).

### 2. Small behaviour note: split could act on *other* sessions

`View` (`rename` with `mySessionID`) only ever renames *your own* session, and only onto a player that already exists. The removed split button could rename *any* listed session, including another tab's/other player's, and to a brand-new name in one step. The replacement flow (add row → `+` → 👁) covers the normal hotseat case fine, and has to be done from the tab being split. That seems like the right trade for the confusion the button caused — just flagging it so it's a conscious loss rather than an accidental one. No dead server code: `rename` with `sessionID` is still used by `View`.

### 3. Test coverage

Nothing automated exercises the add-player row; `tests/testcafe/test-util.js:36` still targets `.myPlayerEntry .playerName`, which is unaffected (verified: the new `#localPlayerName` also carries `.playerName`, but every selector in the codebase is either scoped to a row or to `.myPlayerEntry`, so no collision). A tiny TestCafe step that types a name into `#localPlayerName` and clicks `#addLocalPlayerButton` would lock in the flow that this PR just rewired — optional given the overlay's history of manual verification.

### Checked and fine

- `removeFromDOM('#playersTable tbody > tr', …)` on line 87 only clears `tbody`, so the new `<tfoot>` row survives every meta update — confirmed across ~6 refills in the browser.
- `<tfoot>` after `<tbody>` is valid HTML5; `colspan="2"` on the actions cell keeps the 4-column grid, and the `+` lines up with the row action buttons.
- `progressButton` import removal is correct (still used by welcome/assets/save).
- Multi-session players still render correctly with the `rowSpan` cells after `.sessionCellContent` was dropped ("Session 1 (you)" / "Session 2" stacked in the merged cell).
- No stale references to the split button remain in css/js/html/tests/docs; legend and intro text were both updated.
- The empty/duplicate branch returns `undefined`, so `serverActionButton` correctly skips the spinner for a request that was never sent, and the client checks mirror `Room.addLocalPlayer` (`server/room.mjs:31-33`) exactly (trim + existing-name).

No UI-support concern applies here — this PR exposes nothing new to game authors (no widget property, no routine function).

Screenshots from my run: [three players](https://agent.virtualtabletop.io/reports/review/1534587283171315824/players-3.png), [multi-session rowspan](https://agent.virtualtabletop.io/reports/review/1534587283171315824/players-multisession.png), [stale invalid state after a successful add](https://agent.virtualtabletop.io/reports/review/1534587283171315824/players-stale-invalid.png).

I added the `beta` label: the diff touches only `client/room.html`, `client/css/overlays/players.css` and `client/js/overlays/players.js` — no engine file at all — and the overlay works end-to-end in a live room, so it is safe and useful for players to try.

VTT-AI-Agent:
Thanks — finding 1 was real and is fixed in 767c3c87.

**1. Stale `setCustomValidity`** — fixed exactly as suggested: `input.setCustomValidity('')` at the top of the `#addLocalPlayerButton` handler, so every attempt starts from a clean slate and a successful add no longer leaves the field red. The `input` listener stays, so the red state also clears as soon as the user starts correcting the name. Verified in a live room with Playwright, driving the sequence you described but setting the corrected name via `input.value = …` (no `input` event) so only the click handler could clear it:

```
after adding Bob:               {"valid":true, "msg":"",                          "value":""}
after duplicate Bob:            {"valid":false,"msg":"This player already exists.","value":"Bob"}
after successful add of Carol:  {"valid":true, "msg":"",                          "value":""}   <- was false/stale before
after empty add:                {"valid":false,"msg":"Please enter a player name.","value":""}
```

**2. Split could act on other sessions** — conscious loss, thanks for pinning it down. The button's one unique capability was "move some other session onto a name that doesn't exist yet", which is also the thing that made it so confusing: it was the only control in the overlay that acted on a session other than your own, with no indication of which browser tab would jump where. The add-row + 👁 flow covers the hotseat case it was actually used for, at the cost of doing it from the tab being split. If the multi-tab case turns out to matter to someone, it's better re-added as an explicit, clearly-labelled control than as the old ✂ icon.

**3. Test coverage** — not adding one here. The TestCafe suite compares room *state* hashes, and the player list lives in room meta rather than state, so an add-player step would produce no state delta to assert on and there is no overlay-UI fixture to extend (the existing files are compute/editor/publiclibrary). That's a fair-sized piece of new test scaffolding for one overlay and would blur this PR's single concern; better as its own change if the overlay keeps growing.

VTT-AI-Agent:
Implemented in 47cd3f58, from this request:

> also add "Invite player to [this room]" with a share button (and link "this room")
>
> move the top text and a modified version of the invite text to a collapsed Help section telling the user about hotseat and remote play. add bigger icons and titles for the paragraphs to make it feel less like walls of text

(Source: [Discord](https://discord.com/channels/770758631146782780/1534579234175979562/1534590878931353722))

**Invite** is now a single line — `Invite players to this room`, where *this room* is a link carrying the room URL (also its tooltip, so it can be copied with the browser's context menu) — followed by the unchanged *Share room URL* button. I used the plural "players" because the singular reads odd next to a share link; happy to change it back.

**Help: playing together** is a collapsed `<details>` at the bottom of the overlay, holding the former intro text plus a rewritten version of the old invite paragraph, split into three topics with a 32px icon and a title each:

- 👥 **Players and sessions** — the old intro text verbatim (sessions = browser tabs, edit name/color, view as another player, add a local player).
- 🔗 **Remote play** — everybody on their own device: send them the room URL, their tab shows up here as a session, they pick their name and color in the table.
- 🔄 **Hotseat** — several people on one device: add them in the last row, use 👁 to switch this tab to whoever's turn it is, only your own session changes.

Nothing else in the overlay moved, and the top of the overlay is now just the "Players" heading above the table.

Collapsed (default) and expanded:

Collapsed
Expanded

For comparison, [what it looked like before](https://agent.virtualtabletop.io/reports/pr/1534579234175979562/players-before.png), and [on a 420px-wide phone viewport](https://agent.virtualtabletop.io/reports/pr/1534579234175979562/players-narrow.png) (no horizontal overflow; the invite heading wraps onto two lines).

`npm test` passes (15/15); the layout, the link target and the expand/collapse were checked in a live room with a real browser.

VTT-AI-Agent:
Moved in bdca53e9:

> "Invite players to this room" was meant to be another row in the table with the share button being in the buttons column

(Source: [Discord](https://discord.com/channels/770758631146782780/1534579234175979562/1534593980023701666))

The separate invite block is gone — the table footer now has two rows: the *Add a local player* row and, below it, a 🔗 row reading **Invite players to [this room]** with the share button in the actions column, the same size and position as the ✏️/👁/🗑 buttons of the player rows.

Players overlay with the add-player and invite rows

Two details worth knowing:

- The share button keeps using `shareButton()`/`progressButton()`, which writes its progress into the button's text ("Sharing…", "Done", or the error). That text would blow up the table cell, so it is hidden with `font-size: 0` on this button — the state still shows through the icon it swaps in (hourglass → green ✔ → red ✖). Verified in a browser: clicking it copies the room URL to the clipboard, the button turns into a green check at an unchanged 28×28px, and it is back to the share icon 2.5s later. [Screenshot right after the click.](https://agent.virtualtabletop.io/reports/pr/1534579234175979562/players-share-clicked.png)
- The *Remote play* and *Hotseat* help paragraphs now point at the share button and the add row instead of at "the block below the table".

Also rechecked: [640×400 viewport with three players](https://agent.virtualtabletop.io/reports/pr/1534579234175979562/players-tab-small.png), a [420px phone viewport](https://agent.virtualtabletop.io/reports/pr/1534579234175979562/players-invite-narrow.png) (no horizontal overflow), and `npm test` (15/15).

VTT-AI-Agent:
Fixed in 575b53ec:

> add player, link and colored avatars don't align properly

(Source: [Discord](https://discord.com/channels/770758631146782780/1534579234175979562/1534598214542692363))

The two footer icons were sized with `width/height: 2em`, but `.material-symbols` sets a 24px font, so their `2em` was a 48px box while a player's color circle is `2em` of the 16px table font = 32px. That made them larger than the avatars, off center against them, and their rows ~14px taller than the player rows.

They now sit in the same `.teamColorWrap` span the player rows use, with just the glyph size set on the icon — so all three columns line up by construction instead of by two size rules that have to agree:

| | color cell | icon/circle box | name column | row height |
|---|---|---|---|---|
| player row | x=230, w=48 | x=238, 32×32 | x=286 | 43px |
| add player | x=230, w=48 | x=238, 32×32 | x=286 | 41px |
| invite | x=230, w=48 | x=238, 32×32 | x=286 | 40.5px |

(measured in the browser; before the fix the two footer boxes were 48×48 and the name column started 16px further right)

Players table with everything aligned

Also rechecked on a 420px phone viewport — all three boxes are 32×32 at the same x there too. `npm test` 15/15.

VTT-AI-Agent:
Removed in 1f03542a:

> Remove "You are the only player at this table. Share the room URL or add a local player to play with others."

(Source: [Discord](https://discord.com/channels/770758631146782780/1534579234175979562/1534599514751631710))

The `#playersAloneHint` div, its CSS and the `classList.toggle('shown', sortedPlayers.length < 2)` call in `fillPlayerList` are all gone — nothing shows or hides below the legend any more. The two footer rows already offer both options it pointed at.

Players overlay with a single player

Checked in a browser with one player and after adding a second (no leftover element, no console errors); `npm test` 15/15.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review). Folding *Add player* and *Invite* into the table is a clear win — the overlay is now one aligned object instead of a table plus a loose block, and dropping the split button removes the action nobody could tell apart from *View*. The one thing I'd change: with the intro paragraph and the alone-hint both gone, a first-time visitor now opens **Players** and sees only a table of icon-only buttons and a *collapsed* help section — open `<details id="playersHelp">` by default while the room has a single player, so the newcomer case is explained and the veteran case stays tidy.

Reviewed live on the PR branch (`players-tab-cleanup`, 1f03542) against its base `players-tab-table`, driving the app with Playwright at **390×844 (phone portrait)**, **820×1180 (tablet)**, **1280×800 (laptop)** and **1920×1080 (desktop)**.
Overlays exercised at all four sizes: **Players** (empty room, 3 players, long name, duplicate-name, empty-name, help collapsed + expanded, share button clicked, invite link clicked), **Game Shelf / states**, **About**. I did not reach the JSON/widget editor, share/publish or confirm dialogs — nothing in this diff touches them, and the only shared rule the diff edits (`.overlay .heading`, players.css:190) has exactly one user in `client/room.html`, so the blast radius really is this one overlay; Game Shelf and About render identically to base ([states-laptop](https://agent.virtualtabletop.io/reports/ui-review/1534603729851126041/states-laptop.png), [about-tablet](https://agent.virtualtabletop.io/reports/ui-review/1534603729851126041/about-tablet.png)).

Laptop, the normal case — [collapsed](https://agent.virtualtabletop.io/reports/ui-review/1534603729851126041/players-added-laptop.png) and [with help open](https://agent.virtualtabletop.io/reports/ui-review/1534603729851126041/players-help-laptop.png):

players laptop

---

## Big layout improvements

**1. Nothing explains the overlay any more until you click something.** The intro text (`room.html`, removed line 87) and `#playersAloneHint` are both gone, and the replacement is collapsed. A first-time user in a fresh room sees: one table row for themselves, a "Sessions" column reading *connected (you)*, three unlabeled icon buttons, and a small `▶ Help: playing together`. Nothing says *how* a second person gets to the table — which is the single question that brings someone to this overlay.
Concrete fix, cheap and reversible: in `fillPlayerList()` (players.js:186, where `#playersAloneHint` used to be toggled) do `$('#playersHelp').open = sortedPlayers.length < 2` on first fill only. That restores exactly the value of the removed hint — help when you're alone, quiet once the table has people — without another block of permanent text.

**2. Two competing help affordances stacked under the table.** The overlay now ends with the icon legend (3 lines of 0.8rem blue text) *and* a help disclosure that re-explains the same icons in prose ("send them the room URL with the ⧉ button", "use the 👁 button…"). At laptop and above they read as one undifferentiated block of small text ([desktop](https://agent.virtualtabletop.io/reports/ui-review/1534603729851126041/players-help-desktop.png)). I'd fold `#playersLegend` into the `<details>` as a fourth topic ("What the buttons do") and rely on the existing `title=` tooltips in the table — the table would then end cleanly at the invite row with a single "Help: playing together" line under it. If you'd rather keep the legend visible, at least separate the two: the summary currently sits ~1rem below the legend in the same colour and weight class, so the section boundary is invisible.

**3. The invite row hides the one thing it exists to hand over.** `<a id="playerInviteURL">this room</a>` (room.html:123) points at `location.href` — the page you are already on. Clicking it reloads the room; in my tablet run the reload briefly registered a *second* session for me in the table below ([after-invite-link-click-tablet](https://agent.virtualtabletop.io/reports/ui-review/1534603729851126041/after-invite-link-click-tablet.png)). Meanwhile the URL itself — which the old block printed as text you could read out over voice chat or select and copy — is now only in `href`/`title`. Suggestion: render the URL as the link text (`text-overflow: ellipsis` inside the name cell, full value in `title`), and make the click copy rather than navigate — i.e. wire the anchor to the same `shareButton` handler and drop the navigation. "Invite players to **localhost:8323/mygame**" is self-explanatory; "Invite players to **this room**" is a link that does nothing useful.

- **Horizontal overflow at phone portrait.** With two or three players the table's min-content width is 382px starting at x=39, so it runs 31px past a 390px viewport and the Sessions column is clipped mid-word ("not connecte…"): [cmp-pr-phone](https://agent.virtualtabletop.io/reports/ui-review/1534603729851126041/cmp-pr-phone.png). To be fair this is *inherited* and this PR makes it better, not worse (base measures 419px wide / 458px scroll width vs. 382/421 here — [cmp-base-phone](https://agent.virtualtabletop.io/reports/ui-review/1534603729851126041/cmp-base-phone.png)), but since you are touching this table anyway: `#playersTable .playerSessionCell { white-space: nowrap }` (players.css:41) plus `.playerName { min-width: 6rem }` (players.css:83, now also applied to `#localPlayerName`) are what pin the width. Letting the session cell wrap, or `#addPlayerRow .playerName { min-width: 0 }`, gets it under 390.
- **The add-player field doesn't look like a field.** `#localPlayerName` inherits `.playerName` (`border: none; background: none; padding: 0`), so at rest it is indistinguishable from static grey label text in a table row — the `+` button is the only hint that the row is interactive, and a box only appears on focus. Existing player names can afford being borderless (they have a pencil button and a value); an empty input can't. A `border-bottom: 1px solid var(--VTTblueLight)` on `#addPlayerRow .playerName` would match the row rules already in the table.
- **`#playersShareButton { font-size: 0 }` (players.css:161) silences the only error message the share path has.** On failure `progressButton` writes `Could not share or copy URL.` into the button (domhelpers.js:51) — invisible here, leaving a red ⚠ icon with no explanation. And on desktop browsers where `navigator.share` is absent, the successful path copies to the clipboard with no wording at all, just a green ✓ for 2.5s; the old button at least said "Done". Consider writing the status into the invite row's name cell instead ("Copied to clipboard!" / the error), or mirroring it into `aria-label`/`title`.
- **Two icons for one concept.** The invite row uses `link` (room.html:122) while the *Remote play* help topic uses `share`, and the help prose then refers to "the ⧉ button in the table above" — pointing at a row the user identifies by a chain-link glyph. Pick one glyph for "get the room URL".
- **Legend wrapping is ragged.** With the split entry removed, `#playersLegend`'s three items wrap as 1 + 2, leaving *Remove this player* stranded to the right of the long *View the game as this player…* line at every size ≥ laptop. A 3-column `grid` (or one item per line) would align them; see [players-help-laptop](https://agent.virtualtabletop.io/reports/ui-review/1534603729851126041/players-help-laptop.png).
- **The native validation bubble covers the invite row** at every viewport, and Chrome draws a black `outline` on the invalid input — the only non-blue focus treatment in the overlay ([players-empty-laptop](https://agent.virtualtabletop.io/reports/ui-review/1534603729851126041/players-empty-laptop.png), [duplicate on phone](https://agent.virtualtabletop.io/reports/ui-review/1534603729851126041/players-dup-phone.png)). Functionally it's a clear improvement over the old red button label, so I'd keep it — just note the input gets no styled invalid state of its own (`:invalid` / `:user-invalid` in the app's red would tie it back to the rest of the UI).
- **Tablet, expanded:** the help section runs under the toolbar and needs an inner scroll to reach *Hotseat* ([clipped](https://agent.virtualtabletop.io/reports/ui-review/1534603729851126041/players-help-tablet.png), [scrolled](https://agent.virtualtabletop.io/reports/ui-review/1534603729851126041/players-help-tablet-scrolled.png)). It is reachable (`#playerOverlay` scrolls: 662 vs 513 client height) and there's no clipping trap, but nothing hints there's more below the fold — one more reason to keep the section collapsed by default for returning users.
- At 1920 the table stays pinned to `max-width: 40rem` in a 1728px-wide overlay, so the footer rows' merged action cell leaves a wide empty right third. Inherited from #3013, just noting it looks emptier now that two more rows end there.

- **"Add a local player"** — *local* is app jargon; a newcomer can't tell it apart from "a player on my local network". The button `title` ("Add a local player that shares this device") has the right words; consider putting them in the placeholder instead: `Add a player who shares this device`.
- **Nothing labels the icon-only footer buttons.** `+` and the share glyph carry only `title=` tooltips, which don't exist on touch — exactly the devices where hotseat play happens. Since these two rows have a spare merged cell to their right, they could afford visible text labels ("Add", "Copy link") without breaking the row rhythm.
- **"Sessions" is the overlay's central concept and is now defined only inside collapsed help.** A first-timer sees *connected (you)*, *not connected*, and after a reload *Session 2*, with no on-screen explanation of what a session is or why they suddenly have two. Point 1 above covers most of this; a `title` on the `Sessions` `<th>` would be a cheap extra.
- **Wording:** *"Help: playing together"* is good and unambiguous. In the Hotseat topic, *"whoever's turn it is"* reads well; *"Only your own session changes, the game itself stays untouched"* is the sentence that used to justify the split button and is now doing real work — nice.
- Contrast is fine throughout (blue on white); the smallest text is the 0.8rem legend/session labels at `opacity: 0.7–0.8`, which is borderline on the session column but unchanged by this PR.

VTT-AI-Agent:
Thanks — the two big points and most of the nitpicks are fixed in 2b31cf5b.

## Fixed

**1. Nothing explained the overlay any more.** `#playersHelp` now starts expanded while the room has a single player and stays collapsed once there is more than one, applied on the first fill only so a later meta update never re-opens what the user closed. That restores exactly what the removed alone-hint signalled.

First-time view

**2. Two competing help affordances.** Took your first option: `#playersLegend` is gone and its three entries are a fourth help topic, *What the buttons do*, one entry per line. The overlay now ends at the invite row with a single `▶ Help: playing together` line under it — which also removes the ragged 1 + 2 wrapping of the legend.

Help expanded

**3. The invite row hid the URL.** The link text is the room URL itself now (`main.js` sets `href` and `innerText`), and clicking it no longer navigates — it runs the same share/copy as the button, so it can no longer spawn a second session by reloading. The `href` is still set so the browser's context menu can copy it.

**`font-size: 0` silenced the share errors.** Dropped. The share button uses a small handler instead of `progressButton`, and the outcome — *Room URL copied to clipboard.* / *Room URL shared.* / the error in dark red — replaces the invite text of that row for 5s. `shareURL()` in `domhelpers.js` now returns `'share'`/`'clipboard'` so the wording matches what actually happened (it is exported now; `shareButton()` and the states overlay are unchanged).

Share result in the row

**The add-player field didn't look like a field.** `#addPlayerRow .playerName` has a `border-bottom: 1px solid var(--VTTblueLight)` (with `box-sizing: border-box`, so the 32px row alignment is unchanged), and `:invalid` turns that underline dark red while the entered name is rejected — a styled invalid state to go with the native bubble, which I kept as you suggested.

**Phone overflow.** Fixed rather than just improved: `#playersTable .playerSessionCell { white-space: normal }` below 530px. The `min-width: 6rem` of the name cell was not what pinned the width — the nowrapped session labels were. Measured at 390×844 with three players:

| | table x / width | overlay scrollWidth / clientWidth |
|---|---|---|
| before | 39 / 382 | 421 / 390 |
| now | 39 / 351 | 390 / 390 |

Above 530px nothing changes: *connected (you)* and *not connected* still sit on one line, and the icon boxes still measure 32×32 at x=350 (laptop) / x=47 (phone) in all three rows.

Phone

**Two icons for one concept.** `share` is now the only glyph for "get the room URL" (button, *Remote play* topic, prose). The invite row's leading icon — the column that says *who*, not *what happens* — is `group_add`, next to the add row's `person_add`.

**Wording.** Placeholder and tooltip are *Add a player who shares this device*; the `Sessions` header carries `title="Every connected browser tab is a session of a player"`.

## Not changed

- **Visible text labels on the footer buttons.** The share button sitting in the actions column at the same size as the ✏️/👁/🗑 buttons is what was explicitly asked for on this PR ([Discord](https://discord.com/channels/770758631146782780/1534579234175979562/1534593980023701666)); text labels would break that alignment again. The help topic plus the tooltips cover the meaning, and the auto-opened help now makes them visible to the person who needs them.
- **Tablet: help runs under the toolbar** and **1920: empty right third of the merged action cell** — both inherited from #3013 and untouched here; the help defaulting to collapsed for returning users takes the edge off the first one.

`npm test` 15/15 and the TestCafe editor test pass; everything above was verified in a live room with Playwright at 390×844 and 1280×800 (help auto-open with 1 vs 3 players, share status + clipboard content, link click not navigating, `:invalid` underline, column alignment, no console errors).

VTT-AI-Agent:
Implemented in 26ca2eeb, from this request:

> When adding a new local player, switch to it immediately **if** there are multiple sessions connected for the current player.

(Source: [Discord](https://discord.com/channels/770758631146782780/1534579234175979562/1534608403836698744))

While the current player has more than one session connected, the `+` row no longer just puts a name into the table — it hands this browser tab over to the new player. That is the case the removed ✂ split button covered: a second tab open on the same table is somebody who wants to be their own player, not somebody adding a hotseat player they would then have to reach with 👁.

It is done as a single `rename` of this session (`{ oldName: playerName, newName, sessionID: mySessionID }`), which is what *View* already sends and which creates the player when the name does not exist yet (`Room.renamePlayer`, `server/room.mjs:920`) — so it is one round trip, the other tabs keep their player, and the old player is not deleted (it still has a session). With a single session connected, `addLocalPlayer` is sent exactly as before.

Because the same button now does two things, it says which one: its tooltip switches to *Add a player and switch this browser tab to them* while a second session is connected, and the *Players and sessions* help topic gained a sentence — "While you have this room open in more than one tab, adding a player switches the current tab to them, so every tab can be its own player."

Verified with two tabs in one browser (both connected as `Guest922`):

| step | tab 1 | tab 2 |
|---|---|---|
| one session, add `Bob` | stays `Guest922`, `Bob` listed as not connected | – |
| tab 2 opens | `Session 1` | `Session 2 (you)` |
| tab 2 adds `Carol` | stays `Guest922`, sees `Carol` connected | **is now `Carol`**, input cleared, tooltip back to the plain wording |
| tab 2 (alone again) adds `Dave` | – | stays `Carol`, `Dave` listed as not connected |

Two sessions of one player, before
Tab 2 after adding Carol

`npm test` 15/15 and the TestCafe editor test pass.

VTT-AI-Agent:
Done in 21b940fb:

> the room link shouldn't be a link. clicking it is usually not what the player wants.

(Source: [Discord](https://discord.com/channels/770758631146782780/1534579234175979562/1534611168847597608))

`#playerInviteURL` is a `<span>` again instead of an `<a>`: the invite row prints the room URL as bold plain text that can be read out over voice chat or selected with the mouse, and there is nothing to click. Sharing or copying it is the share button's job, which is right next to it and still reports *Room URL copied to clipboard.* in that row.

This also reverts the change to `client/js/main.js` — it sets the URL as text exactly the way it did before this PR, so the file is no longer part of the diff.

Invite row with the URL as plain text

Verified in a browser: no `<a>` left in the row, the cursor stays the default one, clicking the URL neither navigates nor reloads (still one session afterwards), selecting it yields the full URL, the share button still copies it, and the table still fits a 390px viewport. `npm test` 15/15 and the TestCafe editor test pass.

VTT-AI-Agent:
Done in 15846d6d:

> make it clearer that the the help text is part of the collapsible. move button help into another "help" thingy. make the "Invite players" text including the URL the same color as the "add a player ..." placeholder text.

(Source: [Discord](https://discord.com/channels/770758631146782780/1534579234175979562/1534612800381522070))

**The collapsible is a box now.** `.playersHelp` has a border, a tinted summary row that gets a separator line while it is open, and a padded body — so the expanded text visibly hangs off the summary instead of just appearing underneath it.

**"What the buttons do" is its own collapsible**, *Help: what the buttons do*, next to *Help: playing together*. While I was moving it I added the two footer buttons to the list, since `+` and the share glyph had nothing but a tooltip explaining them.

**The invite row uses the placeholder color** (`#757575`), URL included, so the two hint rows of the table read as hints and the URL no longer competes with the player names above it. The `::placeholder` color of `#localPlayerName` is now spelled out with the same value, because browsers pick different defaults for it and the two would otherwise only match in Chrome. The URL keeps its bold weight so it still stands out inside the sentence.

Collapsed, and the new second section open:

Collapsed
Button help

Both open:

Both help sections open

Only *Help: playing together* still opens by itself while you are alone at the table. Checked in a browser that the three colors are identical (`rgb(117, 117, 117)` for the invite cell, the URL and the placeholder) and that the framed boxes do not reintroduce the phone overflow — they needed `box-sizing: border-box`, since `width: 100%` plus a 1px border pushed the overlay 2px past a 390px viewport. `npm test` 15/15 and the TestCafe editor test pass.

VTT-AI-Agent:
Fixed in 4d92a204:

> The inline icons in the text are a few pixels off for me in Android Firefox.

(Source: [Discord](https://discord.com/channels/770758631146782780/1534579234175979562/1534613538260127806))

The icons in the help text used `vertical-align: text-bottom`, which places them relative to the **descender line of the surrounding font** — and engines disagree about which vertical metrics of a font to read, so the same rule lands a couple of pixels apart in Gecko and Blink.

They are now placed by their own box instead:

```css
height: 1em;
line-height: 1em;
overflow: hidden;        /* baseline of the inline-block becomes its bottom margin edge */
vertical-align: -0.25em; /* ...so this is a plain length below the text baseline */
```

With `overflow` other than `visible`, an inline-block is aligned by its bottom margin edge rather than by the baseline of its content, so nothing in the placement depends on font metrics any more — only on the font size.

I could not run Gecko here (the sandbox has no `libgtk-3`, so Playwright's Firefox will not start), so I reproduced the same class of shift the other way round: swapping the metrics of the *surrounding* font moves a `text-bottom` icon exactly like a different engine's metrics would.

| surrounding font | icon box vs. text baseline, before | after |
|---|---|---|
| Roboto | bottom +4, top −12 | bottom +4, top −12 |
| Georgia | bottom +3, top −13 | bottom +4, top −12 |
| monospace | bottom +5, top −11 | bottom +4, top −12 |

So the rendering in Chromium is unchanged (still +4/−12 at the 16px help font) and the icons no longer drift. Checked at 3× zoom that the fixed height plus `overflow: hidden` does not clip any glyph — the widest ink of these symbols reaches 92% of the em above the baseline inside a box that spans 100% of it:

Button list at 3x
Prose icons at 3x

`npm test` 15/15 and the TestCafe editor test pass. If they still look off on your phone after this, it would be good to know whether they sit too high or too low — the offset is now a single number (`-0.25em`) that is easy to nudge.

VTT-AI-Agent:
Done in 2959ffb4:

> Remove the gear from the toolbar icon.

(Source: [Discord](https://discord.com/channels/770758631146782780/1534579234175979562/1534614112041173114))

`#playersButton::before` was `[users_settings]` — a person with a gear, which reads as a settings screen rather than as the list of people at the table. It is `[people]` now: the gear-less glyph of the same family, and plural like the tab it opens.

Toolbar icon options

The VTT-Symbols font also has `[person]` (literally the same person without the gear) and `[group]` (three people) if either fits better — it is a one-word change either way, just say which.

In the toolbar:

Players icon in the toolbar

`npm test` 15/15 and the TestCafe editor test pass.

VTT-AI-Agent:
Both help sections now stay collapsed by default — the auto-expand for a lone player is gone (64e9aaf7).

> Don't expand help for solo players by default. I don't like it.

— [arnoldsmith86 on Discord](https://discord.com/channels/770758631146782780/1534579234175979562/1534618944864915597)

The `<details>` simply start closed and only open when clicked, whatever the player count. Screenshot of a fresh room with one player: https://agent.virtualtabletop.io/reports/pr/1534579234175979562/players-first-time-collapsed.png

VTT-AI-Agent:
> Is the URL darker or does it just look that way because it's bold? I want it to be same color.

— [arnoldsmith86 on Discord](https://discord.com/channels/770758631146782780/1534579234175979562/1534619678570189010)

It just looked that way: the computed color of `Invite players to` and of the URL was the same `rgb(117, 117, 117)`, only the weight differed (400 vs 700). Dropped the `font-weight: bold` in bbdf346e so the row is uniform:

Invite row with a plain URL

VTT-AI-Agent:
> not sure if the trash icon to delete old names is working properly.  Shouldn't I be able to delete Incognito, Left, and MikeAdd?

— [lawdawg96 on Discord](https://discord.com/channels/770758631146782780/1534579234175979562/1534619950919061708)

The button was working as designed, but the design was unreadable — which is the actual bug here.

A player can only be dropped from the list when nothing in the game points at them any more: the server (`Room.removeLocalPlayer` → `playerIsReferencedInWidgets`) refuses the removal while any widget has that name in its `owner`, `player` (a seat) or `artist` property, because the removal would leave those pointing at a player that no longer exists. So Incognito, Left and MikeAdd still hold something in your room — a card or other widget they own, a seat they are sitting in, or an artist credit — while LeftAdd, Mike Phone and MikeTest do not. The client mirrored that check by *removing* the button, so the rows just silently had different numbers of buttons.

Now the button stays where it is and greys out, with a tooltip that names the reason, e.g.:

> You cannot remove Left because cards or other widgets in the game belong to them and they are seated in the game. Rename them to a player who stays, or free what belongs to them, and the button becomes available.

Remove button greyed out with an explanation

(For a *connected* player the button is still hidden — the sessions column of the same row already says why.)

Verified in a browser on a room built exactly like your screenshot: the greyed button does nothing when clicked, an available one still removes the player. `npm test` 15/15 and the TestCafe editor test pass.

If you would rather have the removal actually go through in that case — clearing the references or handing them to another player as part of the delete — say so and we can look at it, but that is a server-side behaviour change and belongs in its own PR.

VTT-AI-Agent:
> The remove button seems to stay disabled even when the player doesn't own anything anymore.
>
> Allow removing any disconnected player. Show a warning about owned stuff instead.

— [arnoldsmith86 on Discord](https://discord.com/channels/770758631146782780/1534579234175979562/1534622558320721960)

Done in 6b9e2224 — the disabled state is gone entirely.

- `Room.removeLocalPlayer` now only refuses while the player is **connected**; the `playerIsReferencedInWidgets` guard is dropped there (it stays in `renamePlayer`, where it protects a connected player's hand from being taken over).
- Every disconnected player therefore has a working trash button. When something in the game still names them, clicking it warns first instead of disabling:

  > **Remove player**
  > Left is not connected, but cards or other widgets in the game belong to them and they are seated in the game.
  >
  > Removing them only takes the name off this list - everything that belongs to them stays in the game and is picked up again by a player of the same name.
  >
  > `Keep` `Remove`

- With nothing pointing at the player, it removes them straight away as before.

Remove button on every disconnected player
Warning before removing a player who still owns widgets

That also settles the stale-tooltip part of the report: the reason text can only ever be a hint now, never something that blocks you. (The likeliest cause of "doesn't own anything anymore" was a seat still holding the name, or an `artist` credit — neither is visible on the table.)

Verified in a browser on a room with owner / seat / artist references: warn-then-confirm removes, warn-then-cancel keeps the player and returns to the Players overlay, and an unreferenced player is removed with no dialog. `npm test` 15/15 and the TestCafe editor test pass.

VTT-AI-Agent:
Renamed the **Sessions** column to **Connections** (a7719e67), following this feedback:

> Call "sessions" "connections" in the UI. Or an even better word that nin-technical users understand while still making clear what's happening.

(Source: [Discord](https://discord.com/channels/770758631146782780/1534579234175979562/1534626487326474411))

What changed in the UI:

- Column header **Connections**, tooltip *Every browser tab that has this room open is one connection of a player*.
- A player with several tabs is listed as *Connection 1* / *Connection 2* instead of *Session 1* / *Session 2*; a single tab still just says *connected*, and *not connected* is unchanged.
- The help topic is *Players and connections*, and remote play says the other person's browser tab "shows up here as a connection".
- The two places that talked about "your session" now name the thing people can see: the *View* tooltip and the hotseat help say *only this browser tab switches over, the game stays unchanged*.

On the "even better word": I went through **devices**, **screens** and **tabs** as well. *Devices* and *screens* are wrong as soon as somebody opens two tabs on one laptop — which is exactly the hotseat case this overlay explains. *Tabs* is the most literal word and it is what the help text uses to explain a connection, but it would be a strange column header for somebody who joined from a phone home-screen shortcut, and it invites confusion with the tabs of the toolbar (Game Shelf / Active Game / Players …). *Connections* survives all of those, so I stayed with your suggestion. Happy to switch if you prefer one of the others.

Players table with a Connections column

VTT-AI-Agent:
The button help now uses the same indentation as the *playing together* topics:

> Use the same indentation in button explanation as in "playing together" (keep the icons smaller though).

(Discord, [message](https://discord.com/channels/770758631146782780/1534579234175979562/1534634525273886800))

Each list item is a flex row with the same 32px icon column and 0.75rem gap as `.helpTopic`, so the text of both sections starts at exactly the same x and wrapped lines hang-indent instead of running back under the icon. The icons stay smaller (20px) and centered in that column.

Before / after:

Button help before, icons inline
Button help after, text aligned with the topics above

Still fits a 390px viewport:

Button help on a 390px viewport

VTT-AI-Agent (commented):
## TL;DR

No actionable correctness, security, or regression findings on the current head. The newly relaxed removal path remains bounded to disconnected player names and leaves widget state intact, while the add/switch and sharing flows correctly wait for or report their asynchronous outcomes.

## Detailed review

I reviewed the full diff against `players-tab-table`, including the changes added after the earlier review passes.

- `server/room.mjs:859`: `removeLocalPlayer` still refuses removal when any live connection uses the name; otherwise it deletes only `_meta.players[playerName]` and broadcasts meta. Owner/seat/artist values remain in room state as the confirmation says, so removing and later reclaiming a referenced name does not corrupt the game.
- `client/js/overlays/players.js:177-201`: reference detection covers the same three properties as the server (`owner`, `player`, `artist`), handles scalar and array values, and asks before the newly permitted removal. Cancel returns to the Players overlay without sending the request.
- `client/js/overlays/players.js:310-333`: validation is cleared before each attempt, duplicate/empty submissions stay client-side, and the multiple-connection path scopes `rename` to `mySessionID`; the completion predicates verify the intended meta/session result rather than accepting an unrelated update.
- `client/js/domhelpers.js:67-82` and `client/js/overlays/players.js:340-348`: exporting `shareURL` is additive, existing `shareButton` behavior is preserved, and the Players row now distinguishes share from clipboard success while surfacing failure text.
- No game-author UI support is required: this PR adds no widget property or routine operation. The changed files are `client/css/layout.css`, `client/css/overlays/players.css`, `client/js/domhelpers.js`, `client/js/overlays/players.js`, `client/room.html`, and `server/room.mjs`; none are designated engine paths.

I could not independently rerun Jest because this checkout has no `node_modules/.bin/jest`; the PR's reported Jest, TestCafe, and targeted browser checks cover the affected behavior.

The `beta` label was already present, and the non-engine UI/meta change is safe and useful for player testing.

VTT-AI-Agent:
Thanks — nothing to change from [automated review 3 of 4](https://github.com/ArnoldSmith86/virtualtabletop/pull/3102#pullrequestreview-4868291280): it reports no correctness, security or regression findings, and its reading of the changed code matches the intent (removal stays bounded to disconnected names and leaves `owner`/`player`/`artist` values in the room state, the add/switch path scopes the `rename` to `mySessionID`, and `shareURL` is an additive export that leaves `shareButton` untouched).

One point I can close out: the review noted it could not rerun Jest because its checkout had no `node_modules`. I installed dependencies and ran the suite on the current head (`1825a311`) — `npm test` passes, 3 suites / 15 tests.

No code changes pushed for this pass.

PR #3103, opened 2026-08-05
Close editor popups when the editor moves on to another widget

VTT-AI-Agent:
## What this changes

Popups of the widget editor (the parameter popups of the Automations routine editor, the "Add routine" popup, info popups) are now closed as soon as the editor moves on to a different widget, instead of being left hanging over an editor they no longer belong to.

## Why

A routine parameter that takes a widget — `MOVE from`/`to`, `SELECT`, the "property of a widget" builder, a `TURN` seat, … — opens a popup that offers **Pick in the room**. Because that button makes a click on a widget in the room fill the parameter, those popups deliberately ignore clicks in the play area.

If the user never presses that button and just clicks a widget in the room directly, the editor does what it normally does: it selects that widget and re-renders the whole properties sidebar for it. The popup, however, ignored the click — so it stayed on screen, anchored to a routine chip that had just been thrown away, now floating over the editor of an unrelated widget. Anything picked in it went to the widget that was no longer on screen.

The same stranding happened whenever the selection changed **without** any click at all, which is the "other circumstances" part of the report:

- the edited widget is deleted, or a pile dissolves and drops out of the selection,
- an undo removes it,
- a new state arrives from the server (loading another game clears the selection).

## How

`setSelection()` already cancels the sound library overlay when the selection really changes, for exactly the same reason ("whoever opened it edits the widget that was shown then"). Editor popups are now closed in the same place, so every route into a selection change is covered rather than just the click. The sidebar's info tips (the small "i" next to a property or section title) are a second, hand-rolled popup implementation that is not part of that list, so they are closed alongside it.

Picking widgets in the room is not the editor moving on: the picker selects what is picked and restores the selection it started from afterwards, so it keeps its popup. That is tied to the selection the picker itself makes — a band drawn in the room — rather than to a picker merely being armed: selecting another widget any other way (the JSON editor's tree, an "Edit line …" link in the sidebar) is the editor moving on, waiting picker or not, and the popup goes with it. It also only holds while the widget the picker belongs to still exists — once that one is deleted or replaced by a new state, the popup goes as well and the picker ends itself instead of waiting for a click in the room that will never come. "Still exists" is decided by identity, not by id: a new state rebuilds every widget in the room as a new object and regularly brings the same ids back (an undo, the same game loaded again), so the picker remembers the widget object it was started for — the way the rest of the editor does with `widgetStillExists()`.

Closing the popups is the last thing `setSelection()` does, after the sidebar has been told about the new selection: it is not a pure dismissal (see below), so the write it can cause must see a consistent selection rather than one being changed.

Behaviour that is unchanged: with **Pick in the room** armed, a click in the room still fills the parameter and the popup stays open.

## Behaviour worth knowing about

The pickers that only write their parameter when the popup goes away — color, icon, sound, key/value pairs and string lists — **apply** what was picked when the popup is closed this way, exactly as they do on a click outside the popup. So picking a color on widget A's routine and then clicking widget B writes that color to A's routine rather than dropping it. Where the widget is gone (deleted, undone, replaced by a new state) the write is dropped by the sidebar's existing `widgetStillExists()` guard.

Because the widget that is written to is off screen by then, the editor says so: a line in the corner of the module panel — *"CANVAS color set to #3cb44b on button"* — that fades after a few seconds, or *"… was not set …: button is gone"* where the write was dropped. Without it, having kept the pick and having thrown it away look exactly the same. An armed room picker that ends because its widget is gone says that too, since the crosshair over the room is all there is to see of it.

## Along the way (from the UI/UX review)

- A popup no longer lands on the control it hangs off: it keeps to the strip above or below it, so the routine it is about stays readable (at 390×844 the popup used to swallow the whole routine). Measured at 390×844, 820×1180, 1280×800 and 1920×1080: 0% of the chip covered, down from 100%.
- While a popup is open, the control it belongs to is outlined in that control's own color — the popup *is* that control, opened up, and the outline going away is what says the editor has moved on.
- A popup no longer reaches over the module button strip on the right (36px over it at 1280×800 before). It was placed by a size that did not include its own padding and border, and it can still change size after being placed — both are handled now.
- The picker button reads **Click a widget in the room…** while armed instead of `click a widget...`, and its tooltip says *room* where the button says room (it said *table*).
- The two hard-coded `color: gray` in `popup.css` follow `--textColor` at 0.7 opacity now, like the popup title does, so they stay readable in both themes.
- A popup no longer scrolls where one of its lists can be shorter instead: the height it is missing comes off the list of widget ids (or of property names, or of operations), which scrolls on its own anyway. On a short window that used to push the widget picker’s **Use these widgets** button out of the popup, with nothing saying it was there.
- Picking a parent **in the room** works for a multi-selection now. The facade the sidebar renders one through carries the ids of all of them as its own id (`h1,h2`), which is no widget in the room, so the picker's target looked gone: the first click only ended the picking, and the list of ids offered the selected widgets as their own parent.
- An info popup is no longer kept out of the strip its own 14px "i" sits in - that button has nothing to read on it, while the prose wants the height (475px of usable height at 1280x800, 780px now). Keeping a *parameter* popup off its chip is what that rule is for.
- The "... set to X on Y" note goes through the very check the write goes through (`widgetStillExists`), so it cannot claim a value the sidebar then drops - the window a dissolving pile spends in the room is exactly that case. A note that has faded to nothing no longer eats clicks either.
- Leaving edit mode closes the popups and ends a room pick the same way closing the properties module does: the editor is only hidden, so a popup left open lives on inside it.

## Testing

- `npm test` (jest): 580 passed.
- TestCafe `tests/testcafe/editor.js`: 36 passed, including the tests below; `routines.js` and `toolbar.js` also pass.
- *"A routine parameter popup goes away with the widget it belongs to"* — clicking another widget in the room closes the popup and selects that widget; a rubber band while a picker runs (the only way a pick reaches `setSelection()`) picks the widget and keeps the popup through the selection the picker restores; an armed room picker still takes a plain click and keeps the popup open.
- *"An editor popup does not outlive the widget it belongs to without a click either"* — the color picked on a routine is written to that routine when clicking another widget closes the popup; removing the edited widget closes the popup with no click involved, even with a room picker armed; and a new state from the server carrying the same widget ids does the same, which is the case the id-based check got wrong.
- *"A property info tip goes away with the widget it explains"* — the same for the sidebar's info tips, which have no other way of noticing.
- *“A long list of widget ids shrinks instead of pushing the apply button out of the popup”* — at 1280×500 with 30 holders in the room, the popup does not scroll and “Use these widgets” is inside it; on the old code the button sat 105px below its bottom edge.
- The two new assertions from the review: the acknowledgement after the color is written to the widget that left the screen, and the one for an armed picker that ends because its widget is gone.
- *"An armed picker does not keep a popup open when the sidebar moves the editor on"* — with **Pick in the room** armed, "Edit line …" in the widget header selects the line: the popup goes and the picker is disarmed. The same through the JSON editor, which also replaces the module the popup hangs off. Leaving edit mode altogether with the picker armed does the same.
- jest: *"a widget that is on its way out of the room does not get credit for the write"* (the note reads *"was not set ..."* for a widget with `isBeingRemoved`), *"the parent picker of a multi-selection picks in the room"* and *"a picker of a multi-selection ends when one of its widgets is gone"*.
- Each of them fails without the corresponding part of the change.
- Verified by hand in a browser as well (open `MOVE`'s `from` popup on a button, click a holder in the room → the popup is gone and the holder is being edited).

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3103/pr-test (or any other room on that server)

### DIFF
--- a/client/css/editor/controls/popup.css
+++ b/client/css/editor/controls/popup.css
@@ -30,8 +30,11 @@
   transition: opacity 1s;
 }
 
+/* a note nobody can see any more must not take the click that was aimed at what
+   is behind it - the strip hangs over the play area on a narrow window */
 .editorNote.editorNoteFading {
   opacity: 0;
+  pointer-events: none;
 }
 
 /* the control a popup is open for, in its own color (a chip of the routine is
--- a/client/css/editor/controls/popup.css
+++ b/client/css/editor/controls/popup.css
@@ -1,5 +1,53 @@
+/* A line of feedback about something that happened where the user is not
+   looking - a value written to the widget the editor has just left, a mode that
+   ended on its own. In the corner of the module panel, above the popups (whose
+   closing is what it is usually about), fading on its own so it never becomes
+   part of the layout. */
+#editorNotes {
+  position: absolute;
+  right: calc(var(--editSidebarWidth) + 10px);
+  bottom: 10px;
+  z-index: 10;
+  display: flex;
+  flex-direction: column;
+  align-items: flex-end;
+  gap: 6px;
+  pointer-events: none; /* the gaps between the notes stay clickable */
+}
+
+.editorNote {
+  pointer-events: auto;
+  cursor: pointer;
+  max-width: min(360px, 80vw);
+  padding: 6px 10px;
+  border-radius: 6px;
+  background: var(--backgroundColor);
+  color: var(--textColor);
+  border: 1px solid var(--backgroundHighlightColor1);
+  border-left: 3px solid var(--VTTblue);
+  box-shadow: 0 2px 12px 0 rgba(0, 0, 0, 0.4);
+  font-size: 13px;
+  transition: opacity 1s;
+}
+
+.editorNote.editorNoteFading {
+  opacity: 0;
+}
+
+/* the control a popup is open for, in its own color (a chip of the routine is
+   colored by what it holds): the popup is that control opened up, so it says
+   which one - and where the popup went when it goes away with it */
+.popupSource {
+  outline: 2px solid currentColor;
+  outline-offset: 1px;
+}
+
 .inline-popup {
   position: fixed;
+  /* the popup is placed by its outer size, so the max-width/max-height it is
+     placed with have to be the outer size as well - with the padding and the
+     border on top of them it overshoots the space it was fitted into by 22px */
+  box-sizing: border-box;
   background: var(--backgroundColor);
   color: var(--textColor);
   z-index: 9;
@@ -68,9 +116,13 @@
   margin-top: 0;
 }
 
+/* the name next to a value is quieter than the value, but a fixed gray is only
+   readable in one of the two themes - this is the text color turned down, the
+   way the popup title does it */
 .inline-popup .popup-property-label {
   flex: 0 0 auto;
-  color: gray;
+  color: var(--textColor);
+  opacity: 0.7;
 }
 
 .inline-popup .popup-property-row .propertyExpandButton {
@@ -424,7 +476,8 @@ body.edit.darkMode .inline-popup button.primary:hover {
 }
 
 .inline-popup .popup-property-empty {
-  color: gray;
+  color: var(--textColor);
+  opacity: 0.7;
   font-style: italic;
   margin: 4px 2px;
 }
--- a/client/js/editor/controls/popup.js
+++ b/client/js/editor/controls/popup.js
@@ -249,6 +249,19 @@ class Popup {
   }
 }
 
+// Every popup of the editor belongs to the widget that is being edited: it hangs
+// off a control (a chip of a routine, the button that adds one) that is thrown
+// away as soon as the editor moves on to another widget, so it would be left
+// floating over the new one with nothing behind it. A click outside dismisses a
+// popup, but a click that selects another widget in the room is not one for the
+// popups that read the room - the widget pickers take the clicks in there - and
+// a selection also changes without any click at all: deleting the widget, an
+// undo, or a new state arriving from the server.
+function closeEditorPopups() {
+  for(const popup of [ ...openPopups ])
+    popup.hide();
+}
+
 // what a tutorial is called in words: the title of the popup it is offered from
 // where there is one ("SELECT - Pick the widgets…" is a SELECT tutorial), and
 // otherwise the file name without the part every one of them starts with
--- a/client/js/editor/controls/popup.js
+++ b/client/js/editor/controls/popup.js
@@ -22,6 +22,14 @@ document.addEventListener('mousedown', e=>{
   popupMouseDownTarget = e.target;
 }, true);
 
+// The lists inside a popup that scroll on their own anyway (the widget ids, the
+// property names, the operations). Where the popup does not fit the room it has,
+// they are what gives way - see fitScrollAreas.
+const popupScrollAreas = '.widgetPickerList, .popup-property-list, .popup-operation-list';
+// three rows or so: a list shorter than that shows less than it takes to scroll
+// it, so the popup scrolls after all instead of shrinking it any further
+const popupScrollAreaMinHeight = 66;
+
 class Popup {
   constructor(source) {
     this.source = source;
@@ -155,6 +163,49 @@ class Popup {
     return height(strip) >= 240 ? strip : limits;
   }
 
+  // A popup that is taller than the room it has scrolls, which puts whatever is
+  // at its bottom out of sight with nothing saying it is there: the button a
+  // section is for is the last thing in it, so "Use these widgets" is exactly
+  // what a too-long list of widget ids pushes below the fold. The lists in a
+  // popup scroll on their own anyway, so the height that is missing is taken
+  // from them rather than from the popup - the list gets shorter, everything
+  // around it stays where it can be seen and reached.
+  fitScrollAreas() {
+    const lists = [ ...$a(popupScrollAreas, this.domElement) ].filter(list=>list.offsetParent !== null);
+    if(!lists.length)
+      return;
+    // start from their full height every time: what does not fit changes with
+    // the section that is open, the search term and the size of the window
+    for(const list of lists)
+      list.style.maxHeight = '';
+    // shrinking a list reflows what is around it, so ask again rather than
+    // assume the popup got shorter by exactly as much as the list did
+    for(let pass=0; pass<3; ++pass) {
+      let missing = this.domElement.scrollHeight - this.domElement.clientHeight;
+      if(missing <= 0)
+        break;
+      let shrank = false;
+      for(const list of lists) {
+        if(missing <= 0)
+          break;
+        const height = list.getBoundingClientRect().height;
+        const shrunk = Math.max(popupScrollAreaMinHeight, height-missing);
+        if(shrunk >= height)
+          continue;
+        list.style.maxHeight = `${shrunk}px`;
+        missing -= height-shrunk;
+        shrank = true;
+      }
+      if(!shrank) // nothing left to give: the popup scrolls after all
+        break;
+    }
+    // the mark that says a list is cut off (a rule and a fade below its last
+    // row) is set when the list is filled, before it is known how tall it may be
+    for(const list of lists)
+      if(list.classList.contains('popup-property-list'))
+        list.classList.toggle('popup-property-list-complete', list.scrollHeight <= list.clientHeight);
+  }
+
   moveIntoView() {
     const limits = this.limitsAroundSource(this.placementLimits());
     // shrink into the available strip instead of hanging out of it
@@ -168,6 +219,8 @@ class Popup {
     // Measured at the left end of the room it may use, it is the width it keeps
     // at every position that fits it, which is every position placed below.
     this.domElement.style.left = `${limits.left+10}px`;
+    // at the width it keeps as well: how tall the content is depends on it
+    this.fitScrollAreas();
     const rect = this.domElement.getBoundingClientRect();
     const fit = (position, size, from, to)=>Math.min(Math.max(position, from+10), Math.max(from+10, to-10-size));
     this.domElement.style.left = `${fit(wanted.left, rect.width, limits.left, limits.right)}px`;
--- a/client/js/editor/controls/popup.js
+++ b/client/js/editor/controls/popup.js
@@ -144,6 +144,15 @@ class Popup {
     return strips[0] || limits; // no strip is usable: covering the room beats being unusable
   }
 
+  // Whether the control the popup hangs off has to stay readable while it is
+  // open. It does where the popup edits that control - a routine chip is the
+  // sentence the parameter belongs to. An info tip is not about its own button:
+  // that button is a 14px glyph with nothing to read on it, while the tip is
+  // prose that wants every line of height it can get.
+  avoidsSource() {
+    return true;
+  }
+
   // The popup is the control it hangs off, opened up: a parameter popup that
   // lands on its own chip hides the sentence it is about (on a phone it swallows
   // the whole routine), and the chip going away with the popup is what says the
@@ -151,7 +160,7 @@ class Popup {
   // control - the roomier of the two - as long as one of them can hold a popup
   // at all; where neither can, being usable still beats being clear of it.
   limitsAroundSource(limits) {
-    if(!this.source || !this.source.isConnected)
+    if(!this.avoidsSource() || !this.source || !this.source.isConnected)
       return limits;
     const source = this.source.getBoundingClientRect();
     if(!source.width && !source.height)
@@ -383,9 +392,30 @@ function closeEditorPopups() {
   }
   closePropertyInfoPopup();
   for(const write of applied)
-    editorNote(write.kept
-      ? `${write.parameter} set to ${write.value} on ${write.widgetID}`
-      : `${write.parameter} was not set to ${write.value}: ${write.widgetID} is gone`);
+    editorNote(writeWords(write));
+}
+
+// Which of the widgets a write goes to are still there to take it, told apart
+// exactly the way the sidebar tells them apart before writing (widgetStillExists
+// in properties.js - the jest harness loads this file without it, so fall back
+// to the identity half of that test there).
+function writtenWidgets(widget) {
+  const exists = typeof widgetStillExists == 'function'
+    ? widgetStillExists
+    : w=>!!w && !w.isBeingRemoved && widgets.get(w.id) === w;
+  const targets = !widget ? [] : (widget.isMulti ? widget.widgets : [ widget ]);
+  return { kept: targets.filter(exists), gone: targets.filter(w=>!exists(w)) };
+}
+
+// What a write that happened off screen reads like. A multi-selection can be
+// half gone, and saying "set" for it would be as wrong as saying "not set": name
+// the widgets it reached and the ones it did not.
+function writeWords(write) {
+  if(!write.goneIDs.length)
+    return `${write.parameter} set to ${write.value} on ${write.widgetID}`;
+  if(!write.keptIDs.length)
+    return `${write.parameter} was not set to ${write.value}: ${write.widgetID} is gone`;
+  return `${write.parameter} set to ${write.value} on ${write.keptIDs.join(', ')}: ${write.goneIDs.join(', ')} is gone`;
 }
 
 // a value in a line of feedback: what was typed for text, the JSON for the rest,
@@ -437,6 +467,12 @@ class InfoPopup extends Popup {
     this.tutorialName = tutorialName;
     this.videoFilename = videoFilename;
     this.title = title;
+  }
+
+  // several paragraphs of prose about a 14px "i": covering that button costs
+  // nothing, halving the height the text may use costs a scrollbar
+  avoidsSource() {
+    return false;
   }
 
   show() {
@@ -574,16 +610,17 @@ class RoutinePopup extends Popup {
     this.applied = true;
     const parameter = this.parameterNames[0];
     const widget = this.widget;
-    // the same identity check the sidebar writes through (widgetStillExists):
-    // where the widget is gone, its write is dropped there rather than saved
-    const stillExists = !!widget && (widget.isMulti
-      ? widget.widgets.some(w=>widgets.get(w.id) === w)
-      : widgets.get(widget.id) === widget);
+    // Exactly the test the write itself goes through, so the note cannot claim
+    // a value the sidebar then drops: a pile that is dissolving is still in
+    // widgets under its own id for the three property changes it takes to get
+    // there, and widgetStillExists is the only thing that knows that.
+    const written = writtenWidgets(widget);
     this.appliedOnHide = {
       parameter: `${(this.operation && this.operation.func) || 'var'} ${parameter}`,
       value: shortValueWords(this.workingValue),
       widgetID: widget && widget.id,
-      kept: stillExists
+      keptIDs: written.kept.map(w=>w.id),
+      goneIDs: written.gone.map(w=>w.id)
     };
     this.notifyChangeListeners({ [parameter]: this.workingValue });
   }
--- a/client/js/editor/controls/popup.js
+++ b/client/js/editor/controls/popup.js
@@ -257,9 +257,15 @@ class Popup {
 // popups that read the room - the widget pickers take the clicks in there - and
 // a selection also changes without any click at all: deleting the widget, an
 // undo, or a new state arriving from the server.
+// Closing is not always a pure dismissal: the pickers that only write their
+// parameter when the popup goes away (color, icon, sound, key/value and string
+// lists) apply what was picked, exactly as they do on a click outside. That
+// write goes to the widget the popup belonged to, which is why this runs after
+// the editor has moved on rather than in the middle of it.
 function closeEditorPopups() {
   for(const popup of [ ...openPopups ])
     popup.hide();
+  closePropertyInfoPopup();
 }
 
 // what a tutorial is called in words: the title of the popup it is offered from
--- a/client/js/editor/controls/popup.js
+++ b/client/js/editor/controls/popup.js
@@ -83,6 +83,8 @@ class Popup {
   hide() {
     if(openPopups.indexOf(this) != -1)
       openPopups.splice(openPopups.indexOf(this), 1);
+    if(this.source && this.source.classList)
+      this.source.classList.remove('popupSource');
     // a popup opened from a button inside this one (the info tip of a section
     // title) belongs to it: every popup is appended to #editor rather than to
     // the one it came from, so without this it stays on screen anchored to an
@@ -94,6 +96,10 @@ class Popup {
     if(this.mutationObserver) {
       this.mutationObserver.disconnect();
       this.mutationObserver = null;
+    }
+    if(this.resizeObserver) {
+      this.resizeObserver.disconnect();
+      this.resizeObserver = null;
     }
     document.removeEventListener('click', this.boundOnOutsideClick);
     document.removeEventListener('keydown', this.boundOnKeyDown, true);
@@ -130,15 +136,45 @@ class Popup {
     return strips[0] || limits; // no strip is usable: covering the room beats being unusable
   }
 
+  // The popup is the control it hangs off, opened up: a parameter popup that
+  // lands on its own chip hides the sentence it is about (on a phone it swallows
+  // the whole routine), and the chip going away with the popup is what says the
+  // editor has moved on. So the popup keeps to the strip above or below that
+  // control - the roomier of the two - as long as one of them can hold a popup
+  // at all; where neither can, being usable still beats being clear of it.
+  limitsAroundSource(limits) {
+    if(!this.source || !this.source.isConnected)
+      return limits;
+    const source = this.source.getBoundingClientRect();
+    if(!source.width && !source.height)
+      return limits;
+    const above = Object.assign({}, limits, { bottom: Math.min(limits.bottom, source.top) });
+    const below = Object.assign({}, limits, { top: Math.max(limits.top, source.bottom) });
+    const height = strip=>strip.bottom-strip.top;
+    const strip = height(below) >= height(above) ? below : above;
+    return height(strip) >= 240 ? strip : limits;
+  }
+
   moveIntoView() {
-    const limits = this.placementLimits();
+    const limits = this.limitsAroundSource(this.placementLimits());
     // shrink into the available strip instead of hanging out of it
     this.domElement.style.maxWidth = `${Math.min(window.innerWidth/2, limits.right-limits.left-20)}px`;
     this.domElement.style.maxHeight = `${Math.min(window.innerHeight-20, limits.bottom-limits.top-20)}px`;
+    const wanted = this.domElement.getBoundingClientRect(); // where it wants to be
+    // How wide the popup lays itself out is limited by how far its left edge is
+    // from the right edge of the screen (it is fixed, with no right), so its
+    // width where it currently sits is not the width it will have where it is
+    // about to be put - and moving it would change the width again, in circles.
+    // Measured at the left end of the room it may use, it is the width it keeps
+    // at every position that fits it, which is every position placed below.
+    this.domElement.style.left = `${limits.left+10}px`;
     const rect = this.domElement.getBoundingClientRect();
     const fit = (position, size, from, to)=>Math.min(Math.max(position, from+10), Math.max(from+10, to-10-size));
-    this.domElement.style.left = `${fit(rect.left, rect.width, limits.left, limits.right)}px`;
-    this.domElement.style.top = `${fit(rect.top, rect.height, limits.top, limits.bottom)}px`;
+    this.domElement.style.left = `${fit(wanted.left, rect.width, limits.left, limits.right)}px`;
+    this.domElement.style.top = `${fit(wanted.top, rect.height, limits.top, limits.bottom)}px`;
+    // the size it was placed with, so that a later resize is only followed when
+    // the content really changed
+    this.placedSize = { width: rect.width, height: rect.height };
   }
 
   notifyChangeListeners(value) {
@@ -225,6 +261,11 @@ class Popup {
 
   show() {
     const sourceRect = this.source.getBoundingClientRect();
+    // which control this popup belongs to, in that control's own color: the
+    // outline appearing and going away is what ties the two together, in
+    // particular when the editor moves on and takes the popup with it
+    if(this.source.classList)
+      this.source.classList.add('popupSource');
     $('#editor').append(this.domElement);
     this.domElement.style.left = `${sourceRect.left}px`;
     this.domElement.style.top = `${sourceRect.bottom}px`;
@@ -246,6 +287,20 @@ class Popup {
       this.mutationObserver = new MutationObserver(_=>this.moveIntoView());
       this.mutationObserver.observe(this.domElement, { childList: true, subtree: true });
     }
+    // The popup also changes size without its content changing - a picker lays
+    // itself out, a swatch grid rewraps once the width is clamped - and the
+    // position computed for the size before that hangs it over the module button
+    // strip or over the control it belongs to. Only a real change of size is
+    // followed, so moveIntoView's own writes (which leave the size as it found
+    // it) cannot feed themselves back in here.
+    if(typeof ResizeObserver != 'undefined' && !this.resizeObserver) {
+      this.resizeObserver = new ResizeObserver(_=>{
+        const rect = this.domElement.getBoundingClientRect();
+        if(!this.placedSize || Math.abs(rect.width-this.placedSize.width) >= 1 || Math.abs(rect.height-this.placedSize.height) >= 1)
+          this.moveIntoView();
+      });
+      this.resizeObserver.observe(this.domElement);
+    }
   }
 }
 
@@ -261,11 +316,56 @@ class Popup {
 // parameter when the popup goes away (color, icon, sound, key/value and string
 // lists) apply what was picked, exactly as they do on a click outside. That
 // write goes to the widget the popup belonged to, which is why this runs after
-// the editor has moved on rather than in the middle of it.
+// the editor has moved on rather than in the middle of it - and why it is said
+// out loud: the widget that was written to is not on screen any more, so "it was
+// saved" and "I lost it" would look exactly the same.
 function closeEditorPopups() {
-  for(const popup of [ ...openPopups ])
+  const applied = [];
+  for(const popup of [ ...openPopups ]) {
     popup.hide();
+    if(popup.appliedOnHide) {
+      applied.push(popup.appliedOnHide);
+      popup.appliedOnHide = null;
+    }
+  }
   closePropertyInfoPopup();
+  for(const write of applied)
+    editorNote(write.kept
+      ? `${write.parameter} set to ${write.value} on ${write.widgetID}`
+      : `${write.parameter} was not set to ${write.value}: ${write.widgetID} is gone`);
+}
+
+// a value in a line of feedback: what was typed for text, the JSON for the rest,
+// cut off where it stops being a line
+function shortValueWords(value) {
+  const text = typeof value == 'string' ? value : JSON.stringify(value);
+  if(typeof text != 'string')
+    return 'nothing';
+  return text.length > 40 ? `${text.slice(0, 39)}…` : text;
+}
+
+// A line of feedback for something the editor did that the user cannot see
+// happening, in the corner of the module panel where the editing is: a write to
+// a widget that has just left the screen, a mode that ended on its own. It says
+// its piece and fades, so it never becomes part of the layout.
+function editorNote(text) {
+  const editor = $('#editor');
+  if(!editor)
+    return;
+  let notes = $('#editorNotes');
+  if(!notes) {
+    notes = div(editor);
+    notes.id = 'editorNotes';
+  }
+  // only the last few, so a burst of them cannot wall off the module panel
+  while(notes.children.length >= 3)
+    notes.firstChild.remove();
+  const note = div(notes, 'editorNote');
+  note.textContent = text;
+  note.title = 'Click to dismiss';
+  note.onclick = _=>note.remove();
+  setTimeout(_=>note.classList.add('editorNoteFading'), 4000);
+  setTimeout(_=>note.remove(), 5000);
 }
 
 // what a tutorial is called in words: the title of the popup it is offered from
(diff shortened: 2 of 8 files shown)


VTT-AI-Agent (commented):
## TL;DR

Solid, well-scoped fix: hooking the popup teardown into `setSelection()` covers every route into a selection change, and the extra `isWidgetPickerRestoringSelection()` guard correctly stops the restore of a single pick from cancelling the popup (and the sound library) it belongs to. Two things worth a look: the sidebar's *other* info-popup mechanism (`propertyInfoButton` in `propertyInputs.js`) is not part of `openPopups` and is therefore still stranded on the no-click routes, and `hide()` on the apply-on-close popups now *commits* a half-made edit on a selection change — an intentional-looking but undocumented and untested side effect.

## Findings

**1. `propertyInfoButton()` popups are not covered — `client/js/editor/propertyInputs.js:405-455`, `client/js/editor/controls/popup.js:260`**

`closeEditorPopups()` walks `openPopups`, which only holds `Popup` instances. The sidebar's property/CSS-editor info tips are a second, hand-rolled implementation (`propertyInfoButton`, used from `propertyInputs.js:585/592`, `cssEditor.js:108/191/213`, `sidebar/properties.js:4542/4869/…`): they build a bare `div.inline-popup` appended to `#editor` and are tracked only in the module-local `activePropertyInfoPopup`. They look identical to the user and hang off exactly the same throw-away sidebar controls.

The click route happens to be fine (their own `onOutsideClick` fires on a room click, since they don't ignore the play area). But the "other circumstances" this PR is really about are not: pin an info tip open on e.g. the `grid` header, then delete the widget / undo / load another game — the sidebar is re-rendered and the tip stays floating over the new one. Since the PR body advertises "info popups" as covered, this is worth closing in the same pass: add a small `closePropertyInfoPopup()` next to `activePropertyInfoPopup` (mirroring `isWidgetPickerRestoringSelection()`) and call it from `closeEditorPopups()`.

**2. Closing an apply-on-close popup now silently commits its working value — `client/js/editor/selection.js:191`**

`RoutineColorPopup`, `RoutineIconPopup`, `RoutineSoundPopup` (via `RoutinePickerPopup.hide()`, popup.js:1809-1820) and `RoutineKeyValuePopup`/`RoutineStringListPopup` (popup.js:1616-1624) apply `workingValue` from inside `hide()`. So `closeEditorPopups()` is not a pure dismissal: picking a color on widget A's routine chip and then clicking widget B now *writes* that color to A's routine. That is consistent with the click-outside semantics these popups already have, and the delete/undo/new-state routes are saved by `widgetStillExists()` in `inputValueUpdated()` (properties.js:1542) — but it is a real behavioural change that the PR body ("Behaviour that is unchanged…") doesn't mention and no test pins down. Please state it in the description, and ideally add an assertion (open the color popup on widget A, pick, click widget B, expect A's routine to carry the new color).

**3. The popups are closed before the new selection is applied — `client/js/editor/selection.js:189-193`**

Because of finding 2, `closeEditorPopups()` can now synchronously run `notifyChangeListeners → onNewValue → routineChanged → widget.set`, i.e. a state write and a delta, while `selectedWidgets` still holds the *old* selection and before the sidebar modules have been told anything. A delta that drops a widget re-enters `setSelection()` from `editorReceiveDelta()`, and the outer call then overwrites `selectedWidgets` again with a now-stale `previousSelectedWidgets` for the `setHighlighted`/`onSelectionChanged` round. Unlikely to bite in practice, but moving `closeEditorPopups()` to the end of `setSelection()` (after the modules have been notified) would make the write see a consistent world and cost nothing. `cancelAudioPicker()` can stay where it is; it doesn't write state.

**4. Test coverage stops at the click route — `tests/testcafe/editor.js:2014`**

The new test is good and covers the click case in both directions, but the routes the "Why" section calls out as the harder half — the widget being deleted, an undo, a new state from the server — are untested, and they are exactly the ones that now go through the apply-on-close path above. Deleting the edited widget while the `from` popup is open is a two-line addition to the same test (open popup → `#editorDragToolbar` delete → `expect(popup.exists).notOk()`). Likewise, the new `isWidgetPickerRestoringSelection()` guard is only exercised by the click path (`handleWidgetPickerClick`), which never reaches `setSelection()` — the restore it actually protects comes from `handleWidgetPickerSelection()` (properties.js:1656), i.e. a rubber-band/selection-mode pick, and nothing covers that.

**5. Minor: an armed picker still parks a stranded popup on the no-click routes — `client/js/editor/selection.js:189`**

`!isWidgetPickerActive()` is needed for the multi-pick rubber-band route, but it is unconditional: with "Pick in the room" armed, deleting the edited widget (or a new state arriving) changes the selection with the popup left on screen, and the picker itself is only cleaned up on the next click (`handleWidgetPickerClick` → target gone → `stopWidgetPicker`). Narrowing the guard to "the picker's target widget is still in the new selection" would cover that too; fine to leave as-is if you'd rather keep the condition simple.

Nothing here blocks the fix — the core mechanism (one place, all routes, with the restore explicitly excluded) is the right shape, and the comments explaining *why* each guard exists are up to this repo's standard.

I added the `beta` label: the diff only touches editor UI files (`client/js/editor/…`) plus a TestCafe test — no engine files (`client/js/widgets/`, `compute.js`, `serverstate.js`) are involved, and the change is safe to put in front of players since it can only close editor popups, never corrupt a game or crash the server.

VTT-AI-Agent:
Addressed the review findings:

1. **`propertyInfoButton()` popups** (finding 1) — `closeEditorPopups()` now also calls a new `closePropertyInfoPopup()` next to `activePropertyInfoPopup` in `propertyInputs.js`, so the sidebar's info tips go with the rest on the routes without a click. New test *"A property info tip goes away with the widget it explains"*.
2. **Apply-on-close** (finding 2) — this is intended, and it is now stated in the PR description under "Behaviour worth knowing about", explained in the comment above `closeEditorPopups()`, and pinned by a test: pick a color on the button's `CANVAS` routine, click a holder, expect the routine to carry `#3cb44b`.
3. **Ordering** (finding 3) — `closeEditorPopups()` moved to the end of `setSelection()`, after the toolbars and sidebar modules have been notified, so the write it can cause sees a consistent selection. `cancelAudioPicker()` stays at the top as suggested.
4. **Test coverage** (finding 4) — added the two routes that were missing: the widget being removed while the popup is open (via `removeWidgetLocal`, i.e. no click anywhere), and the rubber-band pick that goes through `handleWidgetPickerSelection()` and the restore it protects. Note a rectangle around a single widget is routed to `handleWidgetPickerClick()` and never reaches `setSelection()`, so the band in the test contains the edited widget as well.
5. **Armed picker on the no-click routes** (finding 5) — the guard is narrowed: `isWidgetPickerChangingSelection()` requires the picker's target widget to still exist, so deleting it takes the popup along instead of parking it. Covered by arming the room picker before removing the widget in the new test.

Each new assertion was checked to fail with the corresponding part of the change reverted. `npm test` 575 passed; TestCafe `editor.js` 34 passed, `routines.js` and `toolbar.js` pass.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-scoped fix — the three new TestCafe tests pass locally and the mechanism (close in `setSelection()`, after the sidebar has moved on) is the right place. One gap: with a widget picker armed, the "a new state arrives from the server" case that the PR explicitly claims to cover does **not** close the popup, because `isWidgetPickerChangingSelection()` checks `widgets.has(id)` and a reloaded state contains the same IDs. Reproduced locally.

## Findings

### 1. An armed picker still strands the popup across a state reload (`client/js/editor/controls/widgetselection.js:94-97`)

```js
function isWidgetPickerChangingSelection() {
  const picker = getWidgetPicker();
  return !!picker && widgets.has(picker.targetWidgetID);
}
```

The check is by **ID**, but `receiveStateFromServer()` (`client/js/serverstate.js:494`) does `widgets.clear()` and rebuilds every widget as a *new object*. A state reload almost always brings back the same IDs (undo of a state-level change, re-loading the same game, the `PUT /state` the tests use), so `widgets.has(picker.targetWidgetID)` is still `true` afterwards — `editorMovedOn` comes out `false` in `selection.js:189`, and neither `cancelAudioPicker()` nor `closeEditorPopups()` runs. Everywhere else the codebase uses identity for exactly this reason (`widgetStillExists()`, `properties.js:476-478`; `selection.js:237`).

Reproduced on this branch with a scratch TestCafe test (button + two holders, open `MOVE`'s `from` popup, click **Pick in the room**, then `setRoomState()` with the same state):

```
 | popups:1 | picking:true      # picker armed  -> popup survives, selection is empty
 | popups:0 | picking:false     # same test without arming the picker -> fixed as intended
```

So the popup is left floating over an editor with *no* selection, still anchored to a routine of a widget object that no longer exists — the exact stranding this PR is about. The picker also stays armed, and its `onPick` then writes through the dead widget and is silently dropped by `widgetStillExists()`, so a subsequent "pick" in the room does nothing at all.

The same reasoning applies more generally: `isWidgetPickerChangingSelection()` says "a picker is running", not "*this* selection change is the picker's". Selecting another widget from the JSON editor or from a sidebar link (`jsonedit.js:2446`, `properties.js:4636`, `sidebar/json.js:111`) while a picker is armed also skips the close.

Two possible fixes, either would do:
* Make the check identity-based — remember the widget object in `startWidgetPicker()` and test `widgets.get(picker.targetWidgetID) === picker.targetWidget` (this also makes `handleWidgetPickerClick`/`handleWidgetPickerSelection`, which currently resolve `widgets.get(id)` to the *new* object after a reload, behave sanely).
* Or simply `stopWidgetPicker()` in `receiveStateFromServer()` (`client/js/editor/selection.js:245-253`) before `setSelection([])` — a picker can't meaningfully outlive the room being replaced, and then the existing guard falls through on its own.

### 2. Test coverage for that route

The new tests cover the click route and the `removeWidgetLocal()` route, but not the "new state arrives from the server" route named in the PR description — which is the one that is broken. A `setRoomState(...)` (same IDs) with a popup open, with and without an armed picker, would pin both halves down.

## Things that check out

* `closeEditorPopups()` snapshots `openPopups` before iterating, and `Popup.hide()` is idempotent (`popup.js:83-104`, plus the one-shot `applied` flag in `RoutinePickerPopup.hide()`), so the nested `hide()` from `newRoutineValues`' change listener and the double-hide from the snapshot are both harmless.
* The apply-on-close write cannot synchronously re-enter `setSelection()` — `resolve(value)` at `popup.js:2048` defers the continuation to a microtask — so placing the call at the end of `setSelection()` is defensive rather than load-bearing, but it is also what keeps the picker guards evaluated against the *old* selection. Worth keeping.
* `closePropertyInfoPopup()` is safe to call when nothing is open, and `close()` nulls `activePropertyInfoPopup` / the `infoButton` closure's `popup`, so tips and info popups can be reopened afterwards.
* Verified locally on this branch: the three new tests pass (`A routine parameter popup goes away…`, `An editor popup does not outlive…`, `A property info tip goes away…`).
* No new widget property or routine function is exposed, so no editor-UI / validator / `jeCommands` follow-up is needed here.

The `beta` label was already on this PR and I left it in place: the diff touches only `client/js/editor/**` and a test file — no engine files at all — and the change is safe to try out, with the gap above only leaving a popup open in a case where it was already left open before this PR.

VTT-AI-Agent:
Both findings addressed in 9c881acb.

**1. An armed picker stranding the popup across a state reload** — confirmed, and the diagnosis was exactly right: `widgets.has(picker.targetWidgetID)` is true again after `receiveStateFromServer()` rebuilt every widget, so `editorMovedOn` came out `false` and the popup was left over an empty editor with the picker still armed.

Went with the identity check: `startWidgetPicker()` now takes the widget rather than its id and keeps the object, and a new `widgetPickerTarget()` returns it only while `widgets.get(id)` still *is* that object. `handleWidgetPickerClick()` and `handleWidgetPickerSelection()` use it too, so neither resolves to the replacement widget after a reload. `isWidgetPickerChangingSelection()` also calls `stopWidgetPicker()` when the target is gone — the picker cannot wait for a click in the room that will never come, since the popup it runs from is being closed in the same breath. That covers the more general point as well: a picker whose target is gone no longer suppresses the close, whichever route the selection change came from.

**2. Test coverage for that route** — the *"An editor popup does not outlive the widget it belongs to…"* test now ends with a `setRoomState()` carrying the same widget ids, with the room picker armed, and asserts the popup is gone, the picker disarmed, and `button` back in the room (so the same-ids premise is actually pinned down). Verified it fails on the previous commit with `expected true to be falsy` at that assertion and passes with the fix.

`npm test` 575 passed; TestCafe `editor.js` 34 passed, `routines.js` and `toolbar.js` pass.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review). The change itself is a clear UX win — a parameter popup no longer floats over an editor it has nothing to do with, and the empty state after deleting the edited widget is clean. The one thing I'd add before merging: **when the popup is closed this way it can silently write a value (colour/icon/sound) to a widget that is no longer on screen, with zero feedback** — that needs a visible acknowledgement.

**What I exercised.** Checked out the branch, ran the server locally and drove it with Playwright. Play-mode overlays: Players, Game Shelf / states, About/help. Edit mode: the whole sidebar rail (`tune` properties, `undo` history, `data_object` JSON editor, `widgets` library, `pest_control` debug, `image`, `home_repair_service`, `settings`), the routine editor, the `MOVE from` widget popup (incl. its "Widgets in the room" / "A property of a widget in the room" / "Values the routine has" sections), the `CANVAS color` popup, the "Pick in the room" armed state, and a property info tip. Each at **390×844, 820×1180, 1280×800 and 1920×1080**. The behaviour the PR is about held at all four sizes: clicking another widget in the room closed the popup and moved the editor on, and `removeWidgetLocal()` closed the popup *and* dropped the crosshair picking state. I could not meaningfully review the play-mode overlays at 390×844 portrait because the app itself puts up "Please rotate your device to landscape" there; the deck editor and share/publish dialogs I did not reach.

---

## Big layout improvements

### 1. The silent write on close needs to be visible

The PR description is upfront about this: colour/icon/sound/key-value pickers *apply* what was picked when the popup is closed. I reproduced it — pick `#3cb44b` in the `CANVAS color` popup, then click a holder in the room:

| popup with the colour picked | after clicking another widget in the room |
| --- | --- |
| picked | after |

`clickRoutine[0].color` goes from `#1f5ca6` to `#3cb44b`, and the second screenshot shows *nothing at all* about it: the popup is gone, the sidebar is now holder2's, and the routine that changed is not on screen. Keeping the value is the right call (dropping it would be worse), but from the user's side an edit happened to an invisible widget with no acknowledgement — the two readings "it saved" and "I lost it" look identical.

Suggestion: when `closeEditorPopups()` closes a popup that applied a value, say so once — a `jeLogNote`-style line ("`CANVAS color` set to `#3cb44b` on `button`") fits the editor's existing conventions and costs nothing visually. A flash of the routine chip would also work if the widget is still selected, but the log line covers the case where it isn't.

### 2. The popup covers the chip it belongs to — 100% of it, at every viewport

I measured the overlap between `.inline-popup` and the `[data-parameter=from]` chip it is anchored to: **100% covered at 390, 820, 1280 and 1920 px**. See [1280×800](https://agent.virtualtabletop.io/reports/ui-review/1534641705100513341/04-popup-from-laptop.png) — the `MOVE … from holder1` row is entirely behind the popup — and [390×844](https://agent.virtualtabletop.io/reports/ui-review/1534641705100513341/phone-popup.png), where the popup swallows the whole routine.

This is pre-existing, but it is exactly the relationship this PR is built on ("the popup belongs to that chip"). If the anchor were visible, both the opening and the closing would explain themselves. Two cheap fixes that fit the existing style:
- position the popup so it clears its own chip rather than landing on top of it (`popup.css:1` is `position: fixed` with a `max-width: 50vw`, so there is room to bias it away from the anchor);
- outline the owning chip in its `--sectionColor` while its popup is open — the same visual language `.accordion-section.open` already uses. The outline disappearing is then the feedback for "the editor moved on".

### 3. "Pick in the room" has no cue in the room

While a picker is armed the only signal is a crosshair cursor (`client/css/editor/propertyInputs.css:1297`) plus the button label inside the popup — and the popup is on the far side of the screen from the play area ([armed state](https://agent.virtualtabletop.io/reports/ui-review/1534641705100513341/05-picking-laptop.png); the left half is visually identical to the unarmed state). This PR makes that worse in one respect: `isWidgetPickerChangingSelection()` now ends the picker silently when its widget is deleted or replaced by a new state, so the one cue can vanish while the user is looking at the room.

Suggestion: while `body.editorWidgetPicking`, show a thin hint bar over the play area — "Click a holder in the room · Esc to cancel". It gives the mode a visible boundary, gives the user the cancel affordance the popup's toggle button currently hides, and gives the silent cancellation something to visibly retract.

### 4. The sidebar keeps its scroll offset when the editor moves on

`setSelection()` is now the single place that knows "the editor really moved on to another widget". Right now, after a room click, the sidebar re-renders for the new widget but keeps the old scroll position, so you land in the middle of an unrelated section — clearest at 390×844: [after clicking a widget in the room](https://agent.virtualtabletop.io/reports/ui-review/1534641705100513341/phone-after-room-click.png) opens on "Target widgets"/"Appearance" with no header in view. Scrolling `#editorModules` to the top on a real selection change would make the transition legible; the `editorMovedOn` flag this PR introduces is the natural hook.

- `client/js/editor/controls/widgetselection.js:237` — the button reads **"Pick in the room"** unarmed and **"click a widget..."** armed: different capitalisation, and an ASCII `...` where the rest of the editor uses `…`. Suggest "Click a widget in the room…" (and "Click widgets in the room…" for `multiple`).
- Same file, line 232 — the tooltip says "the widget on the **table**" while the button says "in the **room**". Two nouns for the same surface in one control; pick one (the app's overlays say "room").
- The popup can extend over the editor's right icon rail: in [color-picked.png](https://agent.virtualtabletop.io/reports/ui-review/1534641705100513341/color-picked.png) at 1280×800 the popup's right edge sits at ~1270px while `#editorSidebar` starts at 1244px, so the `image`/`home_repair_service`/`settings` icons are hidden behind it. Clamping the popup to the modules panel (or giving it an 8px right gutter) would keep the rail reachable while a popup is open.
- `.inline-popup .popup-property-label` and `.popup-property-empty` (`popup.css:71`, `:426`) hard-code `color: gray`, unlike the rest of the file which goes through `--textColor`/opacity. On the dark editor that grey is the low-contrast end of the scale; `var(--textColor)` at ~0.7 opacity — which `.popup-title-raw` already does — would match and stay legible in both themes.
- The property info tip is a second popup implementation that this PR now has to close by hand (`closePropertyInfoPopup()`). Visually it is also a different object from `.inline-popup` proper — [laptop info tip](https://agent.virtualtabletop.io/reports/ui-review/1534641705100513341/laptop-infotip.png) is a 582px-wide bar with no title row and no close affordance, versus the titled/×-ed popups. Since the two are now bound together in lifetime, it would be worth giving them the same chrome.

- **Nothing explains why the popup disappeared.** For someone who has just discovered "Pick in the room", clicking a widget in the room is a completely reasonable thing to try — and the result is that the popup vanishes and a different widget's editor appears. The behaviour is correct; a one-line note (see #1) turns it from "did I break something?" into "ah, it moved to the widget I clicked".
- **Popup title `MOVE - which widget ⓘ from`** ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1534641705100513341/04-popup-from-laptop.png)) — the plain-language half doesn't say *which side* of the MOVE is being filled in; both `from` and `to` read as "which widget", and the only thing distinguishing them is the small monospace `from`. "MOVE — take widgets from" / "MOVE — put them into" would be self-explanatory and still leave the raw name next to it.
- **"Pick in the room" is the only way to say "I don't know the id"**, but it is a plain outlined button sitting next to "Select none", while the filled primary button in the same section is "Use these widgets". For a newcomer the affordance ordering is inverted — the discovery action looks secondary. Making the pick button primary (or at least giving it the eyedropper icon a label like "Click it on the table") would match how the section is actually used.
- The section names **"Widgets in the room" / "A property of a widget in the room" / "Values the routine has"** are good plain language; the colour coding matches the routine chips. No complaints there — worth keeping as the model for the rest of the editor.
- The empty state after the edited widget is deleted is genuinely good — the popup goes, the crosshair goes, and the sidebar says what to do next: [after-delete.png](https://agent.virtualtabletop.io/reports/ui-review/1534641705100513341/after-delete.png).

VTT-AI-Agent:
Thanks — worked through the UI/UX review; five of the findings are fixed in 42c6409, three I'd rather not do here (with reasons), and one is a proposal I'd like a human opinion on before building it.

Everything below was re-measured with Playwright at the same four viewports (390×844, 820×1180, 1280×800, 1920×1080).

## Fixed

**1. The silent write on close is now visible.** Closing a picker popup by moving the editor on writes the picked value to a widget that is no longer on screen, with nothing to see. The editor now says so once, in the corner of the module panel, and fades after a few seconds:

> `CANVAS color set to #3cb44b on button`

and where the write was dropped because the widget is gone (`widgetStillExists()`), it says that instead: `CANVAS color was not set to #3cb44b: button is gone`. Same line for the other half of the finding under #3 — an armed room picker that ends because its widget went away says `picking in the room ended: button is gone`, so the crosshair never just disappears.

I did not use `jeLogNote` itself: `#jeLog` only exists while the JSON editor module is open, and the case this is about happens in the properties module. The note is a plain `.editorNote` in `#editorNotes` (dismissable by clicking it), styled from `--textColor`/`--backgroundColor` so it works in both themes.

note

**2. The popup no longer covers the chip it is anchored to.** `moveIntoView()` now keeps the popup in the strip above or below its source control — the roomier of the two, as long as it can hold a popup at all (240px); below that it falls back to today's behaviour rather than becoming unusable. Chip coverage measured **0% at all four viewports**, down from 100%. On the phone the whole routine stays readable:

| 1280×800 | 390×844 |
| --- | --- |
| laptop | phone |

I also took the second half of that suggestion: while a popup is open, the control it hangs off is outlined in its own colour (`currentColor`, so a widget chip outlines purple, a value chip orange, …). That is the visible half of what this PR is about — the outline going away is what says the editor moved on.

**3. The popup over the right icon rail.** Reproduced (36px over `#editorSidebar` at 1280×800) and fixed, with two causes:
* `.inline-popup` was sized content-box, so the `max-width`/`max-height` it is *placed* by were 22px smaller than the box that ends up on screen — it is `box-sizing: border-box` now.
* it can still change size *after* being placed (the colour picker lays itself out, the swatch grid rewraps), so a `ResizeObserver` re-places it. To keep that from oscillating, the width is now measured at the left end of the room the popup may use: with `position: fixed` and no `right`, how wide it lays itself out depends on its own `left`, so measuring it where it currently sits made the width depend on the position being computed from it.

Now 0px over the rail at every size.

**4./5. The nitpicks.** The armed picker button reads **Click a widget in the room…** (`Click widgets in the room…` for the multiple case) instead of `click a widget...`; its tooltip says *room* rather than *table*; and both `color: gray` in `popup.css` are `var(--textColor)` at 0.7 opacity now, the way `.popup-title-raw` already did it.

## A proposal, not built yet

The **hint bar over the play area** for the armed picker ("Click a holder in the room · Esc to cancel") is a new, persistent piece of UI over the room, and where it sits, what it says and whether it swallows clicks are decisions I don't want to make on my own. Concretely what I would build: a single `#editorPickingHint` bar, `pointer-events: none`, centred at the top of `#roomArea` while `body.editorWidgetPicking`, in the editor's popup colours, reading "Click a widget in the room — Esc to cancel" (`widgets` for a multiple picker); Escape would then cancel only the picker rather than closing the whole popup, which is what it does today. Happy to do it in this PR or a separate one — say the word (or say it should look different). Until then the retraction is covered by the note above.

## Not changed

* **Scrolling `#editorModules` to the top on a real selection change.** It cuts both ways: editing the same property across several widgets in a row (click A, edit `x`, click B, edit `x`) is a common flow, and a scroll reset makes you find that property again every time. It is also a behaviour change unrelated to the stranded popups, so I'd rather it were its own decision than a side effect of this fix.
* **`MOVE - which widget from` → "MOVE — take widgets from".** Agreed that "which widget" is the same sentence for `from` and `to`, but that wording is generated for every parameter of every operation (`parameterQuestion()`); rewording it is a pass over the routine editor's whole vocabulary, not a line in this PR.
* **Making "Pick in the room" the primary button, and giving the info tips the same chrome as `.inline-popup`.** Both are pre-existing and larger than this fix; worth their own PR.

`npm test` 575 passed; TestCafe `editor.js` 34 passed (including the two new assertions), `routines.js` and `toolbar.js` pass.

96LawDawg:
The scroll area for list of existing widgets should never be so high that you have to scroll to see "use this widgets. In fact, make it so the picker itself is high enough that scrolling is not needed, flexing the height of the click on id area as needed to fit.
image

VTT-AI-Agent:
Done in f17563ce — the list of ids gives up its own height so the popup does not scroll.

`moveIntoView()` now fits the popup's content into the room the popup has: whatever height is missing is taken from the lists inside it (the widget ids, the property names, the operation list), which scroll on their own anyway, instead of from the popup. The `220px` the id list used to insist on is its maximum now, not its height — so on a short window it shrinks and everything below it stays in view.

| before | after |
| --- | --- |
| before | after |

Both at 1280×620 with 30 holders in the room: before, the list runs into the bottom edge of the popup and "Use these widgets" (plus the two sections under it) is below the fold; after, the list is 97px, the popup does not scroll at all and the button is where it belongs.

One limit worth naming: the list stops shrinking at about three rows. Below that it shows less than it costs to scroll it, so a window short enough to need that (roughly under 500px of usable height here) still scrolls the popup. Everything from there up now fits.

New TestCafe assertion in `editor.js` ("A long list of widget ids shrinks instead of pushing the apply button out of the popup") measures exactly that: at 1280×500 the popup scroll is 0 and the apply button is inside it — it fails on the old code with the button 105px out of view. `npm test` 575 passed; TestCafe `editor.js` 35 passed, `routines.js` + `toolbar.js` 8 passed.

VTT-AI-Agent (commented):
## TL;DR

The earlier state-reload issue is fixed and the current CI suite is green, but one related selection-origin gap remains: while a room picker is armed, any selection change is assumed to belong to that picker as long as its source widget still exists. Selecting a different widget through another editor control can therefore strand the popup over the new editor.

## Findings

### 1. An active picker suppresses unrelated selection changes (`client/js/editor/controls/widgetselection.js:110-114`, `client/js/editor/selection.js:189-190`)

`isWidgetPickerChangingSelection()` does not inspect where the selection change came from; it returns `true` whenever a picker is active and its target object still exists. `setSelection()` consequently sets `editorMovedOn` to false for every such change, not only for the temporary selection made by `handleWidgetPickerSelection()`.

There are several non-room routes that call `setSelection()` while leaving the original widget alive. A concrete one is: open a routine widget-parameter popup, arm **Pick in the room**, switch to the JSON editor, then click another widget in its tree (`client/js/jsonedit.js:2442-2451`). The routine popup also ignores every outside click while that picker is active (`client/js/editor/controls/popup.js:635-641`), so neither the click nor `setSelection()` closes it. The result is the original widget's routine popup floating over the newly selected widget/JSON editor, with its picker still armed—the same stale-editor condition this PR is intended to eliminate. Sidebar links such as **Edit line …** (`client/js/editor/sidebar/properties.js:4631-4637`) have the same shape.

Please make the exception explicit to the picker-driven selection operation instead of deriving it solely from “a picker exists.” For example, wrap only the `setSelection()` reached from the rubber-band picker path in a short-lived flag (as the restore path already does), or pass an origin option into `setSelection()`. A TestCafe case that arms the picker and then selects another widget from the JSON tree would pin down the distinction.

No new widget property or routine operation is exposed, so the authoring-UI support rule does not require any additional editor controls here.

The `beta` label was already present and I left it in place because the diff is limited to edit-mode UI and tests, with no engine behavior changes, and this remaining popup-lifetime issue is non-destructive.

VTT-AI-Agent:
Addressed in 489b8f07 — the finding was right, and it turned out to have a second half.

**1. An active picker suppressed unrelated selection changes.** Confirmed. `isWidgetPickerChangingSelection()` derived the exception from "a picker exists", so every route into `setSelection()` was treated as the picker's. The exception is tied to the operation now: `applySelectionRectangle()` — the only route where a pick reaches `setSelection()`, a band drawn in the room — runs inside `selectWidgetsInRoom()`, and `isWidgetPickerChangingSelection()` is `picker && selectingWidgetsInRoom && widgetPickerTarget()`. Everything else (the JSON editor's tree, `Edit line …`, an undo, a new state) is the editor moving on, armed picker or not.

The second half: `handleWidgetPickerSelection()` had the same shape, and the properties sidebar calls it for *every* selection change. So an unrelated selection was not merely tolerated — it was turned into a pick and then undone by the picker's restore, which put the editor back on the widget the picker belonged to. With the popup gone but the selection restored, the sidebar could not move on at all while a picker ran. It takes the same origin guard, and the unit test *"a selection made anywhere but in the room is not a pick"* pins it.

Ending a picker whose widget is gone moved out of that question into `endWidgetPickerWithoutTarget()`, called from `setSelection()` whichever route the change came from — so `isWidgetPickerChangingSelection()` is now a plain question with no side effect.

Along the way, the concrete repro you gave found one more: `PropertiesModule` had **two** `onClose()` methods (the second added by #3034), so the later one silently replaced the `stopWidgetPicker()` of the earlier — closing the module left the picker armed and its popup hanging over whatever module replaced it. They are one method now, and it closes the popups as well: the controls they hang off go with the module.

**2. Test coverage.** New TestCafe test *"An armed picker does not keep a popup open when the sidebar moves the editor on"*: with **Pick in the room** armed, `Edit line route` in the widget header selects the line — a selection change that never touches the room, with the picker's widget still alive — and the popup is gone and the picker disarmed. It fails on the previous commit with the line never becoming selected (the picker's restore undid it). The existing popup test additionally arms the picker, switches to the JSON editor (`popup gone`, `picker disarmed` — that assertion fails without the `onClose()` fix) and then selects another widget in `#jeTree`.

`npm test` 576 passed; TestCafe `editor.js` 36 passed, `routines.js` + `toolbar.js` 8 passed.

VTT-AI-Agent (commented):
## TL;DR

The core mechanism is right and now well factored: closing in `setSelection()` after the sidebar has moved on, with the picker only owning a selection it made *in the room* (`selectWidgetsInRoom`) and identity — not id — deciding whether its widget still exists. CI is green and the earlier review gaps are closed. My main remaining point is a small one: the "…set to X on Y" note can claim a write that the sidebar then drops, because `applyWorkingValueOnHide()`'s existence check is not quite the sidebar's `widgetStillExists()`.

## Findings

**1. `kept` is not the same test the write goes through — the note can lie (`client/js/editor/controls/popup.js:577-581`)**

```js
const stillExists = !!widget && (widget.isMulti
  ? widget.widgets.some(w=>widgets.get(w.id) === w)
  : widgets.get(widget.id) === widget);
```

`widgetStillExists()` (`client/js/editor/sidebar/properties.js:476`) is `!!widget && !widget.isBeingRemoved && widgets.get(widget.id) === widget`. The `isBeingRemoved` term is missing here, and that flag exists exactly for the pile case (`client/js/widgets/pile.js:258` sets it, then `await`s three property changes before `removeWidgetLocal`). In that window the widget is still in `widgets` under its own id, so this reports `kept: true` and the user reads *"CANVAS color set to … on pile-xyz"*, while `inputValueUpdated`/`inputValueTransformed` drop the write. That is precisely the "kept it" vs "threw it away" confusion the note was added to remove.

Same for the multi branch: `some()` says kept if *any* widget survives, but `widget.widgets.filter(widgetStillExists)` writes only to the survivors — so a partially-dead multi-selection also reports an unqualified success. Cheapest fix is to call `widgetStillExists` itself (with a `typeof widgetStillExists == 'function'` guard, since the jest harness loads `popup.js` without `properties.js` — the file already does that for `stopSoundPreview` at line 2098).

**2. A note that has faded to nothing still eats clicks (`client/css/editor/controls/popup.css:23-40`)**

`.editorNote` sets `pointer-events: auto`, and `.editorNoteFading` only changes `opacity`. Between the 4s and the 5s timer the note is invisible but still hit-testable, so for a full second a click aimed at the module panel (or a widget in the room, where the strip overhangs it) lands on nothing. Add `pointer-events: none` to `.editorNoteFading`.

**3. `onEditorClose()` does not do what `onClose()` now does (`client/js/editor/sidebar/properties.js:1629-1642`)**

Nice catch on the duplicate `onClose()` — the first definition was dead, so `stopWidgetPicker()` never actually ran on module close before. But leaving edit mode goes through `closeEditor()` → `module.onEditorClose()` (`client/js/editor/layout.js:108`), which is not `onClose()`. An open popup then survives inside `#editor`, which is only `display: none`d (`client/css/editor/layout.css:100`), and an armed picker keeps `body.editorWidgetPicking`, i.e. a `cursor: crosshair !important` over the entire page in play mode (`client/css/editor/propertyInputs.css:1297`). In practice the click on the edit toggle usually dismisses the popup first, so this is a corner — but given the PR's premise ("whatever the editor has open belongs to the widget being edited"), leaving the editor altogether is the most complete form of moving on, and the two lines are the same two lines.

**4. In-room picking is dead for a multi-selection, and the new note misreports it (`client/js/editor/controls/widgetselection.js:7-13, 47-52, 138-141`)**

`renderParentWidgetInput()` is rendered for multi-selections too (it has a `— multiple —` placeholder, `client/js/editor/sidebar/properties.js:4286`), and it passes the selection widget straight into `startWidgetPicker(widget, …)`. For a `MultiWidget` that is `id = 'h1,h2'` (`client/js/editor/propertyInputs.js:91`), so `widgets.get(picker.targetWidgetID)` is `undefined` and `widgetPickerTarget()` returns null: the first click just stops the picker without picking. That is pre-existing (the old code did `widgets.get(picker.targetWidgetID)` too), but this PR is the one that reworks target identity, and it now adds a user-visible consequence — a selection change from any other route while such a picker is armed emits *"picking in the room ended: h1,h2 is gone"* about widgets that are not gone. Handling `targetWidget.isMulti` in `widgetPickerTarget()` (all/any of `widget.widgets` still identical in `widgets`) would fix both.

**5. `limitsAroundSource()` applies to every popup, including the long info tips (`client/js/editor/controls/popup.js:147-164`)**

Keeping a *parameter* popup off its own chip is clearly right. But this also confines an info tip (`infoButton`, several paragraphs of prose) to the strip above or below its little "i", so one anchored near the vertical middle of the window is now clamped to roughly half the height it could use and scrolls where it previously did not. Was that checked for the info popups, or would it be better scoped to sources that are routine chips (the thing the popup is *about*)?

**6. Minor / no action needed**

- `fitScrollAreas()` runs from the `MutationObserver` on every content change (each keystroke in the widget-id search resets `maxHeight` on every list and does up to three forced read/write layout passes). Fine at the sizes tested; just a thing to remember if a room with hundreds of widgets ever feels sticky in that search box.
- The PR bundles the actual fix with a placement rework, list shrinking, a whole new notification surface (`#editorNotes`) and CSS/wording changes — against `CLAUDE.md`'s "one PR = one concern". The notes are load-bearing for the fix, so I would not split them out now, but the placement/`box-sizing`/`gray` changes are independent and would have been easier to review (and to revert) on their own.
- `isWidgetPickerRestoringSelection()` is *not* redundant with `isWidgetPickerChangingSelection()`, despite looking like it: a single-widget picker calls `stopWidgetPicker()` before `onPick`, so `activeWidgetPicker` is already null by the time `restoreSelection()` re-enters `setSelection()`. Worth keeping that in mind before anyone "simplifies" the two guards into one.

The new jest and TestCafe tests cover the interesting routes (band-vs-click, delete, same-ids new state, sidebar-driven selection with a picker armed), which is what makes the rest easy to trust. No new widget property or routine function is exposed, so nothing is owed to the validator, `jeCommands` or the widget/routine editing UI here.

I added no label: `beta` is already on this PR, and it stays — the change touches only editor UI files (none of `client/js/widgets/`, `compute.js` or `serverstate.js`), so it is not engine-excluded and is safe for players to try.

VTT-AI-Agent:
All five findings addressed in 57f8ef3c. Three were real bugs, two were judgement calls I agreed with.

**1. `kept` is not the same test the write goes through.** Confirmed, including the multi half. `applyWorkingValueOnHide()` calls `widgetStillExists` itself now (behind the `typeof` guard the file already uses for `stopSoundPreview`, since the jest harness loads `popup.js` without the sidebar), so the `isBeingRemoved` window a dissolving pile spends in `widgets` reads as gone here too. And instead of one `kept` flag it keeps the ids the write reached and the ids it did not, so a half-dead multi-selection says *"CANVAS color set to #3cb44b on h1: h2 is gone"* rather than an unqualified success. New jest test *"a widget that is on its way out of the room does not get credit for the write"* — it reports `set to … on button` on the previous commit.

**2. A faded note still eats clicks.** Fixed: `.editorNoteFading` is `pointer-events: none`.

**3. `onEditorClose()` does not do what `onClose()` now does.** Right, and not as much of a corner as it looked: with a picker armed the routine popup ignores *every* outside click (`RoutinePopup.onOutsideClick` returns early while `routineWidgetPickerKey` is active), so the click on the exit button does not dismiss it either — the popup really did survive into play mode, crosshair and all. `onEditorClose()` calls `onClose()` now. The TestCafe test *"An armed picker does not keep a popup open when the sidebar moves the editor on"* ends with that route; it fails on the previous commit with `expected true to be falsy`.

**4. In-room picking is dead for a multi-selection.** Confirmed and fixed at the root: `widgetPickerTarget()` asks the widgets *behind* a facade (`isMulti`) whether they are still there instead of looking up its `"h1,h2"` id, and `handleWidgetPickerSelection()` compares against all of them (its restore also puts the real widgets back rather than the facade). While I was there, `renderWidgetSelectPopout()` excludes the member ids rather than the facade's id — the list used to offer the selected widgets as their own parent. Measured in a browser, two holders selected, **Pick in the room**, click a third:

```
before  offered ids: h1,h2,target   parents after the click: ,        (nothing picked)
after   offered ids: target         parents after the click: target,target
```

Two jest tests cover it.

**5. `limitsAroundSource()` also clamps long info tips.** Good catch — it was not checked for them, and it does hurt: at 1280×800 with the tip's "i" near the middle, the popup's `max-height` went from 780px to 475px. It is an overridable `avoidsSource()` now, `false` for `InfoPopup`: keeping a *parameter* popup off its chip is the point (the chip is the sentence the popup is about), while an info button is a 14px glyph with nothing to read on it. Measured back at 780px.

**6.** Noted, no change: `fitScrollAreas()`' cost in a huge room, and the PR bundling more than one concern — agreed on both, and on `isWidgetPickerRestoringSelection()` not being redundant.

`npm test` 580 passed; TestCafe `editor.js` 36 passed, `routines.js` + `toolbar.js` 8 passed.

ArnoldSmith86 (approved):

VTT-AI-Agent:
Updated to \`main\` (merge 25fba1c1, no conflicts) ahead of merging.

One follow-up the update called for: #3101 landed in the meantime, so entering edit mode now opens the **Edit Widgets** module by itself. The five tests this PR adds opened that module by clicking the sidebar's `tune` button — which now *closes* it, so all five failed on the merged tree. They state their own window size and editor state now (`setEditorState({ modules: { 'Edit Widgets': 'editorModuleTopLeft' } })`), the way `editor.js`'s other module-dependent tests do since #3101, and reset it afterwards.

Nothing else the update touched changed what this PR does: `PropertiesModule.onEditorClose()` still runs `onClose()` after the new base-class `onEditorClose()`, the reworked routine block rendering (#3106) and the modal icon picker (#3105) are both outside the popup lifetime this PR governs.

Verified on the merged tree: jest 1172 passed (28 suites), TestCafe `editor.js` 45 passed, `routines.js` + `toolbar.js` 11 passed.

PR #3105, opened 2026-08-06
Render the edit-mode "Pick icon" picker as a modal above the room instead of behind it

VTT-AI-Agent:
The edit-mode **"Pick icon"** picker rendered across the whole window with the game board painted on top of it, so the icon list was half covered and the picker looked like it was sitting *behind* the room. This makes it render as a proper modal above the room, scoped to the editor's content area, with only the symbol list scrolling — and makes that layout hold up in the small and narrow boxes the picker is also opened in.

### What was wrong

The shared symbol picker normally lives in the room overlays (`#roomArea`). Opening the deck editor moves it into `#editor` (`DeckEditor.initializeDOM`) and closing it deliberately leaves it there, so that a picker opened from the JSON editor or the properties panel covers the editor instead of being squeezed into the small room box. From that point on, every "Pick icon" in edit mode opens from `#editor`.

But nothing in `#editor` gave it a box or a stacking order of its own, so `.overlay`'s own `height/width: 100%` stretched it across the whole window, and `#roomArea` (`position: fixed; z-index: 1`) painted the board over the open picker. The sound picker next to it already had a rule for exactly this (`#editor > #audioPickerOverlay`); the symbol picker never got one.

### What changed

- `client/css/editor/deckeditor.css` — the `#editor > #audioPickerOverlay` rule now covers `#symbolPickerOverlay` as well, and both are scoped to the deck editor's own box (below the toolbar, beside the sidebar) instead of the whole window, so they are sized like the other editor overlays and float above the sidebar/toolbar/dialogs.
- `client/css/fonts.css` — the picker behaves like a dialog: the title, the search field and the hint stay put in a 1000px-wide centered column and `#symbolList` is the only part that scrolls. Previously the whole overlay scrolled, so scrolling down to an icon took the search field and the close button off the top of the screen. This applies in every host (room overlays, deck editor card column, editor), so the picker in the small room box got noticeably more usable too.
  - The list keeps a floor of about four rows so it can never be squeezed to nothing, and the card around it is never squeezed below what it holds — a box that cannot afford even that overflows the overlay, which scrolls, rather than clipping the bottom of the list away with no way to reach it.
  - In a **short or narrow** box the hint paragraph is dropped and the title shrinks so the close button stops sitting on top of it. That breakpoint is a container query on the picker's own box, because the picker is opened in three hosts of very different sizes and the window size says nothing about which one it is in.
  - The picker gets its own `padding: 20px 10px` instead of inheriting `.overlay`'s `padding: 5% 10%`. That percentage resolves against the box **width** in both directions, so the wider the host the more vertical padding the picker got — ~90px of dead band above the title in `#editor` at 1920px, while the small boxes that need the height got almost none.
- `client/room.html` + `client/js/symbols.js` — the hint now leads with what the dialog is for ("Click an icon to pick it.") instead of with the low-resolution caveat, the search placeholder gives examples, and a search that matches nothing shows `No icons match "…"` instead of a large blank card. That count follows the picker's image-only / font-only restriction, so a search whose only hits belong to the hidden family counts as empty too — every icon and category heading carries `data-family=image|font`, which is also what the restriction itself is now written against.
- `client/js/symbols.js` — the picker opens at the top of the icon list again: the reset used to be on the overlay, which is no longer the scroller.
- `client/css/jsonedit.css` — the JSON editor's widget-layers panel ("This will show you the widgets under your cursor…") is now hidden while an overlay is open. It lists what the cursor hovers in the room and already hides while the cursor is over the editor, but an open overlay takes the cursor away from the room without necessarily putting it over the editor, so it stayed up and covered the command buttons underneath it (visible on the right of the report screenshot).

No state, no widget/routine properties — CSS plus a few lines of client JS for the empty-search message and to keep the existing "open at the top" behaviour working now that the list is the scroller.

### Before / after

Same steps in both: open a room with a deck, enter edit mode, open the deck editor once and close it again, open the JSON editor and the widget properties panel, select a widget, open its Icon property and click "Show all".

| | |
|---|---|
| before | https://agent.virtualtabletop.io/reports/pr/1535013339619655913/pick-icon-before.png |
| after | https://agent.virtualtabletop.io/reports/pr/1535013339619655913/pick-icon-after.png |
| after, list scrolled (search + close stay put) | https://agent.virtualtabletop.io/reports/pr/1535013339619655913/pick-icon-after-scrolled.png |
| after, picker inside the deck editor (unchanged host, still constrained to the card column) | https://agent.virtualtabletop.io/reports/pr/1535013339619655913/pick-icon-after-deckeditor.png |

The narrow host at tablet portrait (the deck editor's card column, a 204×931 box), before and after — the icon list goes from a 77px strip under a 703px hint to the whole column:

| | |
|---|---|
| before | https://agent.virtualtabletop.io/reports/pr/1535013339619655913/ui-review-2/before-deckcol-820x1180.png |
| after | https://agent.virtualtabletop.io/reports/pr/1535013339619655913/ui-review-2/after-deckcol-820x1180.png |

`#symbolList` clientHeight measured in Chromium, all three hosts, before this PR → after:

| host | viewport | picker box | before | after |
|---|---|---|---|---|
| deck editor card column | 820×1180 | 204×931 | 120 (43px clipped) | 776 |
| deck editor card column | 1280×800 | 664×539 | 236 | 282 |
| deck editor card column | 1920×1080 | 1212×819 | 499 | 581 |
| room overlays | 390×844 | 354×221 | 120 (26px clipped) | 169 (+ overlay scrolls) |
| room overlays | 1280×800 | 640×400 | 196 | 245 |
| room overlays | 1920×1080 | 960×600 | 287 | 362 |
| `#editor` | 820×1180 | 784×1156 | 838 | 918 |
| `#editor` | 1280×800 | 1244×764 | 438 | 526 |
| `#editor` | 1920×1080 | 1792×1044 | 654 | 806 |

### Testing

- `npm test` — 590 passed.
- `npx testcafe chromium:headless tests/testcafe/editor.js` — 33 passed, including the empty-state checks added to "Deck editor: symbol pickers and JSON fallback" and the new "The symbol picker says an image-only search found nothing".
- Driven manually in a browser: picker opened from the properties panel in all three hosts (room overlays, `#editor`, deck editor card column) at four viewports, scrolled, searched (no matches / few matches / many matches), and an icon picked — the value lands on the widget.
- After merging `main` (which turned on clean-css level 2 and terser top-level mangling, #3089): re-checked that the minified client still ships the picker's `@container symbolPicker` block, its `min-height` floors and the `padding` override in the right order, then re-ran `npx testcafe chromium:headless tests/testcafe/editor.js` **against a minified client** (the combination `production-environment.yml` now runs) — 33 passed — and re-measured `#symbolList` in all three hosts (918 / 526 / 776 px, unchanged).

Reported by a maintainer with this screenshot: https://agent.virtualtabletop.io/reports/screens/pick-icon-overlay-2026-08-06.png

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3105/pr-test (or any other room on that server)

### DIFF
--- a/client/css/editor/deckeditor.css
+++ b/client/css/editor/deckeditor.css
@@ -474,12 +474,20 @@ body.deckEditorRoomVisible #deckEditorMainCol > .overlay {
   z-index: 30;
 }
 
-/* The sound picker is opened from the JSON editor, which is reachable while the deck editor covers the play
-   area, so it is moved into #editor as well (see initializeDOM). It has no smaller box to be constrained to
-   there, so float it over the whole editor like the rule above - otherwise #deckEditor (z-index 1) hides it. */
+/* Both pickers are opened from the JSON editor and the properties panel, which stay reachable while the deck
+   editor covers the play area, so they are moved into #editor (see initializeDOM) and the symbol picker is
+   left there when the deck editor closes (see close()) - from then on that is where edit mode opens it. In
+   #editor they have no box and no stacking order of their own, so .overlay's own 100%/100% makes them cover
+   the whole window and #roomArea (position: fixed, z-index 1) paints the board on top of the open picker.
+   Give them the deck editor's box instead - below the toolbar, beside the sidebar - and float them above the
+   sidebar (2), the toolbar (3) and the deck editor's dialogs (20). */
+#editor > #symbolPickerOverlay,
 #editor > #audioPickerOverlay {
   position: fixed;
-  inset: 0;
+  top: var(--editToolbarHeight);
+  right: var(--editSidebarWidth);
+  bottom: 0;
+  left: 0;
   width: auto;
   height: auto;
   z-index: 30;
--- a/client/css/editor/deckeditor.css
+++ b/client/css/editor/deckeditor.css
@@ -462,27 +462,18 @@ body.deckEditorRoomVisible #deckEditorMainCol > .overlay {
   z-index: 20;
 }
 
-/* Opening the picker from an icon property inside one of the dialogs above parks it next to that dialog
-   (pickSymbolKeepingOverlay), where the rule above no longer applies and it would render behind #deckEditor
-   (z-index 1). Float it over the whole editor instead: above the sidebar (2), the toolbar (3) and the
-   dialog it was opened from (20). */
-#symbolPickerOverlay.symbolPickerAboveEditor {
-  position: fixed;
-  inset: 0;
-  width: auto;
-  height: auto;
-  z-index: 30;
-}
-
 /* Both pickers are opened from the JSON editor and the properties panel, which stay reachable while the deck
    editor covers the play area, so they are moved into #editor (see initializeDOM) and the symbol picker is
-   left there when the deck editor closes (see close()) - from then on that is where edit mode opens it. In
-   #editor they have no box and no stacking order of their own, so .overlay's own 100%/100% makes them cover
-   the whole window and #roomArea (position: fixed, z-index 1) paints the board on top of the open picker.
-   Give them the deck editor's box instead - below the toolbar, beside the sidebar - and float them above the
-   sidebar (2), the toolbar (3) and the deck editor's dialogs (20). */
+   left there when the deck editor closes (see close()) - from then on that is where edit mode opens it.
+   Opening the picker from an icon property inside one of the dialogs above parks it next to that dialog
+   instead (pickSymbolKeepingOverlay, which marks it with .symbolPickerAboveEditor) - same situation.
+   Neither spot gives them a box or a stacking order of their own, so .overlay's own 100%/100% makes them
+   cover the whole window and #roomArea (position: fixed, z-index 1) paints the board on top of the open
+   picker. Give them the deck editor's box instead - below the toolbar, beside the sidebar - and float them
+   above the sidebar (2), the toolbar (3) and the deck editor's dialogs (20). */
 #editor > #symbolPickerOverlay,
-#editor > #audioPickerOverlay {
+#editor > #audioPickerOverlay,
+#symbolPickerOverlay.symbolPickerAboveEditor {
   position: fixed;
   top: var(--editToolbarHeight);
   right: var(--editSidebarWidth);
--- a/client/css/fonts.css
+++ b/client/css/fonts.css
@@ -395,17 +395,29 @@
    the media queries at the end of this block cannot see - so make it a container and query that instead. */
 #symbolPickerOverlay {
   container: symbolPicker / size;
+  /* .overlay's own 5% 10% is a percentage of the WIDTH, in both directions: the wider the host box the more
+     vertical padding the picker gets - ~90px of dead band above the title in #editor at 1920px, while the
+     small boxes that actually need every pixel of height get almost none. The column below is centered and
+     capped at 1000px anyway, so the horizontal padding only matters once the box is narrower than that. */
+  padding: 20px 10px;
 }
 
-/* In a short box the hint is what has to go: it explains the list but costs more height than the list is
-   then left with (114px of a 250px box). Same breakpoint as the media queries below, and the same
-   compensation .fewResults makes when it hides the hint for the same reason. */
-@container symbolPicker (max-height: 375px) {
+/* In a short or a narrow box the hint is what has to go: it explains the list but costs more height than the
+   list is then left with - 114px of a 250px-tall box, and 700px of a 931px-tall one when the box is only
+   204px wide and every line of the hint holds two words. Same breakpoints as the media queries below, and
+   the same compensation .fewResults makes when it hides the hint for the same reason. */
+@container symbolPicker (max-height: 375px), symbolPicker (max-width: 600px) {
   #symbolPickerOverlay p[icon] {
     display: none;
   }
   #symbolPickerOverlay input {
     margin-bottom: 20px;
+  }
+  /* the close button is positioned over the right edge of the title, which the centered "Pick icon" runs
+     into once the box is this narrow */
+  #symbolPickerOverlay h1 {
+    font-size: 20px;
+    margin: 0 40px 10px;
   }
 }
 
@@ -424,6 +436,15 @@
   flex-direction: column;
   flex: 1;
   min-height: 0;
+}
+
+/* div[icon] clips (overflow: hidden, for its rounded corners and its watermark glyph), so squeezing it below
+   what it holds drops the bottom of the list on the floor with nothing left to scroll. Keep it at least as
+   tall as the reduced layout needs - its own padding, the search field, and the list's floor below. The
+   plain wrapper around it does not clip, so a box that cannot afford even this overflows the overlay, which
+   scrolls (.overlay is overflow: auto), instead of losing the rest of the list. */
+#symbolPickerOverlay div[icon] {
+  min-height: 210px;
 }
 
 #symbolPickerOverlay > button {
@@ -462,6 +483,22 @@
 }
 #symbolPickerOverlay.bigPreviews.fewResults #symbolList {
   --symbolSize: 64px;
+}
+
+/* a search that matches nothing left an empty card without saying why - take the list's place instead */
+#symbolNoResults {
+  display: none;
+}
+#symbolPickerOverlay.noResults #symbolList {
+  display: none;
+}
+#symbolPickerOverlay.noResults #symbolNoResults {
+  display: flex;
+  flex: 1;
+  align-items: center;
+  justify-content: center;
+  text-align: center;
+  word-wrap: anywhere;
 }
 
 #symbolPickerOverlay h1 {
--- a/client/css/fonts.css
+++ b/client/css/fonts.css
@@ -388,9 +388,25 @@
 
 /* The picker is a dialog, not a document: the title, the search field and the hint stay put and the symbol
    list is the only part that scrolls. The whole overlay scrolling meant that scrolling down to an icon also
-   scrolled the search field and the close button off the top of the screen. */
+   scrolled the search field and the close button off the top of the screen. .overlay's own overflow: auto
+   stays as a fallback for a box too short even for the reduced layout below.
+   How much room the parts have is decided by the picker's own box, not by the window: it is opened in three
+   hosts of very different sizes (the room overlays, the deck editor's card column, #editor), which is what
+   the media queries at the end of this block cannot see - so make it a container and query that instead. */
 #symbolPickerOverlay {
-  overflow: hidden;
+  container: symbolPicker / size;
+}
+
+/* In a short box the hint is what has to go: it explains the list but costs more height than the list is
+   then left with (114px of a 250px box). Same breakpoint as the media queries below, and the same
+   compensation .fewResults makes when it hides the hint for the same reason. */
+@container symbolPicker (max-height: 375px) {
+  #symbolPickerOverlay p[icon] {
+    display: none;
+  }
+  #symbolPickerOverlay input {
+    margin-bottom: 20px;
+  }
 }
 
 /* the title and the list are separate children, so they need the same width to line up - the close button
@@ -440,7 +456,9 @@
   overflow-y: auto;
   word-wrap: anywhere;
   flex: 1;
-  min-height: 0;
+  /* the list takes what is left over, but never less than about four rows: with min-height: 0 whatever sits
+     above it could squeeze it to nothing, and since it is the only scroller that leaves no icon reachable */
+  min-height: 120px;
 }
 #symbolPickerOverlay.bigPreviews.fewResults #symbolList {
   --symbolSize: 64px;
--- a/client/css/fonts.css
+++ b/client/css/fonts.css
@@ -553,15 +553,11 @@
   background: var(--url);
 }
 
+/* every icon and category heading carries data-family=image|font, so the restriction to one of the two
+   families is a single pair of selectors - and symbols.js can count what is left with the same attribute */
 #symbolPickerOverlay .hidden,
-#symbolPickerOverlay.hideImages h2.imageCategory,
-#symbolPickerOverlay.hideImages .emoji-color,
-#symbolPickerOverlay.hideImages .gameicons,
-#symbolPickerOverlay.hideFonts h2.fontCategory,
-#symbolPickerOverlay.hideFonts .symbols,
-#symbolPickerOverlay.hideFonts .material-symbols,
-#symbolPickerOverlay.hideFonts .material-symbols-nofill,
-#symbolPickerOverlay.hideFonts .emoji-monochrome {
+#symbolPickerOverlay.hideImages [data-family=image],
+#symbolPickerOverlay.hideFonts [data-family=font] {
   display: none;
 }
 
--- a/client/css/fonts.css
+++ b/client/css/fonts.css
@@ -386,6 +386,30 @@
 
 
 
+/* The picker is a dialog, not a document: the title, the search field and the hint stay put and the symbol
+   list is the only part that scrolls. The whole overlay scrolling meant that scrolling down to an icon also
+   scrolled the search field and the close button off the top of the screen. */
+#symbolPickerOverlay {
+  overflow: hidden;
+}
+
+/* the title and the list are separate children, so they need the same width to line up - the close button
+   sits at the right edge of the title */
+#symbolPickerOverlay > .titleWithCloseButton,
+#symbolPickerOverlay > div:not(.titleWithCloseButton) {
+  width: 100%;
+  max-width: 1000px;
+  margin: 0 auto;
+}
+
+#symbolPickerOverlay > div:not(.titleWithCloseButton),
+#symbolPickerOverlay div[icon] {
+  display: flex;
+  flex-direction: column;
+  flex: 1;
+  min-height: 0;
+}
+
 #symbolPickerOverlay > button {
   position: absolute;
   right: 10px;
@@ -415,6 +439,8 @@
   overflow-x: hidden;
   overflow-y: auto;
   word-wrap: anywhere;
+  flex: 1;
+  min-height: 0;
 }
 #symbolPickerOverlay.bigPreviews.fewResults #symbolList {
   --symbolSize: 64px;
--- a/client/css/jsonedit.css
+++ b/client/css/jsonedit.css
@@ -418,7 +418,12 @@ i.error   { color: var(--textHighlightColor3); font-weight: bold; }
   white-space: normal;
 }
 
-#editor:hover #jeWidgetLayers, .editorModule:hover #jeWidgetLayers, body.usedTouch #jeWidgetLayers {
+/* This panel covers the command buttons underneath it and only lists what the cursor is hovering in the room,
+   so it hides while the cursor is over the editor. An open overlay (the symbol picker, a dialog) takes the
+   cursor away from the room in the same way, but does not necessarily put it over the editor - without
+   body.overlayActive the panel stayed up over the commands for as long as the overlay was open. */
+#editor:hover #jeWidgetLayers, .editorModule:hover #jeWidgetLayers, body.usedTouch #jeWidgetLayers,
+body.overlayActive #jeWidgetLayers {
   display: none;
 }
 
--- a/client/js/editor/propertyInputs.js
+++ b/client/js/editor/propertyInputs.js
@@ -233,8 +233,8 @@ async function pickSymbolKeepingOverlay(element, type='all') {
   const picker = $('#symbolPickerOverlay');
   const pickerParent = picker.parentNode;
   hostOverlay.parentNode.appendChild(picker);
-  // Where it lands has no stacking rule of its own, so it would end up behind the deck editor: the class
-  // floats it over the whole editor (see deckeditor.css) for as long as it is parked here.
+  // Where it lands has no box or stacking rule of its own, so it would end up behind the deck editor: the
+  // class gives it the editor's box and floats it above (see deckeditor.css) while it is parked here.
   picker.classList.add('symbolPickerAboveEditor');
 
   // Whatever happens, the user must get their dialog back: the picker's symbol list is fetched, so it can also
--- a/client/js/symbols.js
+++ b/client/js/symbols.js
@@ -129,7 +129,7 @@ export async function pickSymbol(type='all', bigPreviews=true, closeOverlay=true
     $('#symbolPickerOverlay').classList.toggle('bigPreviews', bigPreviews);
     $('#symbolPickerOverlay').classList.toggle('hideFonts',   type=='images');
     $('#symbolPickerOverlay').classList.toggle('hideImages',  type=='fonts');
-    $('#symbolPickerOverlay').scrollTop = 0;
+    $('#symbolList').scrollTop = 0; // the list is built once and is the picker's scroller, so open it at the top
     $('#symbolPickerOverlay input').value = '';
     $('#symbolPickerOverlay input').focus();
     $('#symbolPickerOverlay input').onkeyup();
@@ -226,7 +226,7 @@ export function addRichtextControls(dom) {
     const range = window.getSelection().getRangeAt(0);
 
     showStatesOverlay('symbolPickerOverlay');
-    $('#symbolPickerOverlay').scrollTop = 0;
+    $('#symbolList').scrollTop = 0;
     for(const c of [ 'bigPreviews', 'hideFonts', 'hideImages' ])
       $('#symbolPickerOverlay').classList.remove(c);
     $('#symbolPickerOverlay input').value = '';
--- a/client/js/symbols.js
+++ b/client/js/symbols.js
@@ -114,7 +114,10 @@ export async function loadSymbolPicker() {
         toggleClass(icon, 'hidden', !icon.dataset.keywords.match(text));
       for(const title of $a('#symbolList h2'))
         toggleClass(title, 'hidden', text);
-      toggleClass($('#symbolPickerOverlay'), 'fewResults', $a('#symbolList i:not(.hidden)').length < 100);
+      const matches = $a('#symbolList i:not(.hidden)').length;
+      toggleClass($('#symbolPickerOverlay'), 'fewResults', matches < 100);
+      toggleClass($('#symbolPickerOverlay'), 'noResults', !matches);
+      $('#symbolNoResults').textContent = `No icons match "${$('#symbolPickerOverlay input').value}".`;
     };
   }
 }
--- a/client/js/symbols.js
+++ b/client/js/symbols.js
@@ -72,12 +72,12 @@ export async function loadSymbolPicker() {
     for(const [ category, symbols ] of Object.entries(symbolData)) {
       if(category == 'Emoji - Flags')
         continue;
-      list += `<h2 class="${category.match(/Material|VTT|Emoji/)?'fontCategory':'imageCategory'}">${category}</h2>`;
+      list += `<h2 data-family="${category.match(/Material|VTT|Emoji/)?'font':'image'}">${category}</h2>`;
       for(let [ symbol, keywords ] of Object.entries(symbols)) {
         if(symbol.includes('/')) {
           const gameIconsIndex = keywords.shift();
           // increase resource limits in /etc/ImageMagick-6/policy.xml to 8GiB and then: montage -background none assets/game-icons.net/*/*.svg -geometry 48x48+0+0 -tile 60x assets/game-icons.net/overview.png
-          list += `<i class="gameicons" title="game-icons.net: ${symbol}" data-type="game-icons" data-symbol="${symbol}" data-keywords="${symbol.split('/')[1]},${keywords.join().toLowerCase()}" style="--x:${gameIconsIndex%60};--y:${Math.floor(gameIconsIndex/60)};--url:url('i/game-icons.net/${symbol}.svg')"></i>`;
+          list += `<i class="gameicons" data-family="image" title="game-icons.net: ${symbol}" data-type="game-icons" data-symbol="${symbol}" data-keywords="${symbol.split('/')[1]},${keywords.join().toLowerCase()}" style="--x:${gameIconsIndex%60};--y:${Math.floor(gameIconsIndex/60)};--url:url('i/game-icons.net/${symbol}.svg')"></i>`;
         } else {
           const hasNoFillVariant = symbol.match(/ \(FILL\+NOFILL\)$/);
           symbol = symbol.replace(/ \(FILL\+NOFILL\)$/, '');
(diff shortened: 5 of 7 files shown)


VTT-AI-Agent (commented):
## TL;DR

The layering fix itself is right and the `#editor`-scoped box matches the sibling rules, but making the overlay `overflow: hidden` and the list the only scroller has a real cost: in a short host box the symbol list collapses to **zero height** and nothing scrolls, so not a single icon is reachable (measured, see below). Also `symbols.js` still resets `scrollTop` on the overlay, which is no longer the scroller.

## Findings

### 1. `client/css/fonts.css:389` — in a short overlay box the list collapses to 0px and nothing can be scrolled

`#symbolPickerOverlay { overflow: hidden }` plus `#symbolList { flex: 1; min-height: 0 }` means the list is the only scroller — but `min-height: 0` also lets it shrink to nothing, and with the overlay no longer scrolling there is then no way to reach any icon.

I measured this in Chromium against a server running this branch, with the picker in its room-overlay host (`#roomArea`, i.e. the picker as it opens outside the deck editor / outside edit mode):

```
viewport   #symbolList clientHeight   scrollHeight   overlay scrollable?
400x700     0 px                       7832 px        no   <-- no icons at all
740x360     121 px                     ...            no   <-- ~2 rows, workable
1024x500    203 px                     ...            no
1280x800    450 px                     ...            no   <-- fine
```

On `main`, the same 400x700 case has the overlay scrolling through 8102px of content, so every icon is reachable. Screenshot of this branch at 400x700: title, search field and the full hint paragraph, and below them nothing — the list is 0px tall.

The 400x700 portrait case is partly masked by the "please rotate your device" state, but the mechanism isn't specific to portrait: the room-overlay host is sized to the board box, so any short box (short window, wide board, landscape phone) squeezes the list, and the hint `<p>` (114px at 400px wide) is what eats the space first. Suggested fix: give `#symbolList` a `min-height` (e.g. `min-height: 120px`) so it can never vanish, and/or keep a scroll fallback on the overlay so content above the list stays reachable when the flex layout runs out of room. `#symbolPickerOverlay.fewResults p { display: none }` already exists — hiding the hint at small heights would buy most of the space back.

### 2. `client/js/symbols.js:132` and `:229` — the "scroll back to the top" reset is now a no-op

Both call sites do `$('#symbolPickerOverlay').scrollTop = 0` when opening the picker. With `overflow: hidden` on the overlay that assignment can no longer do anything, and `#symbolList` — the new scroll container — keeps its position (the list is built once, `loadSymbolPicker` is guarded). Concretely: open the picker, scroll down to the flags near the bottom, pick an icon, open the picker again — it now opens scrolled to the bottom instead of at the top, and clearing the search field doesn't reset it either. A one-line JS follow-up (`$('#symbolList').scrollTop = 0`) belongs with this CSS change; leaving it means the PR silently drops an existing behaviour.

### 3. `client/css/editor/deckeditor.css:469` vs `:480` — the new rule now overrides `.symbolPickerAboveEditor`

`#editor > #symbolPickerOverlay` has two IDs (2,0,0); `#symbolPickerOverlay.symbolPickerAboveEditor` has one ID and a class (1,1,0). So whenever `pickSymbolKeepingOverlay` (`client/js/editor/propertyInputs.js:233`) parks the picker in `#editor` and adds that class, the new rule wins and the picker gets the editor box instead of the `inset: 0` the class intends. That is almost certainly the behaviour you want, but the older rule and its comment now describe something that no longer happens for the `#editor` case — worth folding the two together (or scoping the class rule to `:not(#editor > *)`) so the next reader doesn't trust a dead rule.

### 4. Adjacent, pre-existing: the picker parked in `#editor` is invisible outside edit mode

`DeckEditor.close()` (`client/js/editor/deckeditor.js:1058`, `:1109`) deliberately leaves `#symbolPickerOverlay` inside `#editor`, and `client/css/editor/layout.css:100` has `body:not(.edit) #editor { display: none }`. A `display: none` ancestor hides the picker no matter what the new rule says, so after the deck editor has been opened once, leaving edit mode and opening the picker from a non-edit context (the rich-text `add_reaction` path, `client/js/symbols.js:228`) would show a blank screen with an invisible close button. Not introduced here — but this PR builds further on the parking, so moving the picker back to `#roomArea` on edit-mode exit would be a good companion change.

### Notes

- No new widget property, routine function or author-facing feature, so the "settable from the editing UI" and "test room lives in `library/tutorials/`" rules don't apply here — correctly, this PR touches no tutorials and no validator/`jeCommands` entries.
- No automated coverage is added, and none of the existing suites would catch #1: the TestCafe state-hash comparison doesn't see CSS layout, and `tests/testcafe/editor.js` only checks that the pickers open. That's consistent with how CSS changes are handled in this repo, but it does mean #1 has to be caught by eye — worth a manual pass at a short window before merge.

I added the `beta` label: the diff is CSS-only (no `client/js/widgets/`, `compute.js` or `serverstate.js` involvement at all), so the PR is beta-eligible, and while finding #1 makes the picker unusable in a short host box it can neither corrupt a game nor crash the server, which is exactly the kind of thing player testing should surface.

VTT-AI-Agent:
Thanks — finding 1 reproduced exactly, and it is fixed together with 2 and 3 in c3ca659a. Details below, including why I left 4 out.

### 1. The list collapsing to 0px — fixed

Reproduced in Chromium on the branch, same host (`#roomArea`) and the same numbers you measured (400x700 → 0px, 740x360 → 121px, 1024x500 → 203px, 1280x800 → 450px). Two changes:

- **`#symbolList` now has `min-height: 120px`** instead of `min-height: 0` — roughly four rows. It is the only scroller, so it must never be squeezed to nothing; `flex: 1` still gives it everything that is left over above that floor.
- **the hint is dropped in a short box**, as you suggested. Not with a media query though: the picker is opened in three hosts of very different sizes (room overlays, the deck editor's card column, `#editor`) and the window says nothing about which one it is in — the 400x700 case is a 250px-tall box in a 700px-tall window. So `#symbolPickerOverlay` is now `container: symbolPicker / size` and the rule is `@container symbolPicker (max-height: 375px)`, i.e. it measures the picker's own box. 375px is the breakpoint the picker's existing media queries already use, and hiding the hint plus the `margin-bottom` compensation is exactly what `.fewResults` already does.
- **the overlay keeps `overflow: auto`.** The `overflow: hidden` declaration is gone entirely rather than changed — `.overlay` already sets `overflow: auto`, so dropping the override restores the scroll fallback you asked for while the flex layout still keeps the header put at every size where it fits.

Measured again after the fix, room-overlay host, `#symbolList` clientHeight (all with the picker still not scrolling the overlay, so search field and close button stay put):

| viewport | overlay box | before | after | hint |
|---|---|---|---|---|
| 400x700 | 400x250 | **0 px** | **120 px** | hidden |
| 600x400 | 556x348 | 109 px | 185 px | hidden |
| 740x360 | 576x360 | 121 px | 197 px | hidden |
| 1024x500 | 800x500 | 203 px | 203 px | shown |
| 1280x800 | 1236x773 | 450 px | 450 px | shown |
| 1920x1080 | 1728x1080 | 709 px | 709 px | shown |

Screenshots of the 400x700 case, picker in the room overlays:

| before (nothing below the hint) | after |
|---|---|
| https://agent.virtualtabletop.io/reports/pr/1535013339619655913/shortbox-400x700-before.png | https://agent.virtualtabletop.io/reports/pr/1535013339619655913/shortbox-400x700-after.png |

and 740x360: [before](https://agent.virtualtabletop.io/reports/pr/1535013339619655913/shortbox-740x360-before.png) / [after](https://agent.virtualtabletop.io/reports/pr/1535013339619655913/shortbox-740x360-after.png).

### 2. The no-op `scrollTop` reset — fixed

Correct, and it was a silent behaviour drop. Both call sites (`symbols.js:132` and `:229`) now reset `$('#symbolList').scrollTop`. Verified: scroll the list down, pick, reopen → back at the top.

### 3. `#editor > #symbolPickerOverlay` outranking `.symbolPickerAboveEditor` — fixed

You are right that the editor box is the behaviour I want, so rather than scoping one rule away from the other I folded them into a single rule: `#editor > #symbolPickerOverlay, #editor > #audioPickerOverlay, #symbolPickerOverlay.symbolPickerAboveEditor` now share one declaration block and one comment covering both ways the picker gets parked. That way the class does the same thing wherever `pickSymbolKeepingOverlay` puts the picker, instead of depending on which selector wins. Checked in the browser: both resolve to the identical box (`0,36 1244x764`, `position: fixed`, `z-index: 30`) at 1280x800. The comment in `propertyInputs.js` that described the old `inset: 0` behaviour is updated too.

### 4. The picker parked in `#editor` being invisible outside edit mode — not in this PR

Confirmed the bug (enter edit mode once, leave it, open the picker → `#symbolPickerOverlay` measures 0x0 inside a `display: none` `#editor`), but I don't think it belongs here:

- It is not specific to the symbol picker. `initializeDOM` parks **six** overlays in `#editor` and never moves any of them back: `#deckEditorExportOverlay`, `#deckEditorImportOverlay`, `#deckEditorNewDeckOverlay`, `#symbolPickerOverlay`, `#audioPickerOverlay` and `#libraryDecksOverlay`. `#libraryDecksOverlay` is shared with the non-edit UI too, so a fix that only moves the picker would leave the same trap next door.
- Doing it properly means a symmetric park/unpark on the edit-mode lifecycle (`openEditor()`/`closeEditor()` are the obvious hooks), not just a move on exit — moving the picker out on exit without putting it back on entry would send it back into the small room box in edit mode, which is exactly what the parking exists to avoid, and would undo what this PR fixes.

That is a JS lifecycle change in the deck editor with its own test surface, so I'd rather it were its own PR than bolted onto this one. Happy to open it if a maintainer wants it — just say so and I'll leave it to `!pr`.

### Testing

- `npm test` — 590 passed.
- `npx testcafe chromium:headless tests/testcafe/editor.js` — 32 passed.
- Browser: measured the table above at six viewports; checked that the search field and close button stay put while the list is scrolled, that reopening starts at the top, and that the `#editor` and `.symbolPickerAboveEditor` boxes match.

On your note about coverage: agreed, none of the suites can see this. The measurement above is a throwaway Playwright script rather than something the repo would keep.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review). The core fix is right and clearly visible: in the `#editor` host the picker is now a proper modal in the editor's own box instead of a full-window overlay with the board painted over it, and pinning the search + close button while only `#symbolList` scrolls is a genuine usability win at every size I tried. The most impactful thing left: the new `min-height: 120px` floor + `@container (max-height: 375px)` breakpoint only account for a **short** box, not a **narrow** one — in the deck-editor card column at tablet portrait (820×1180) the hint paragraph wraps to ~700px, the list is pushed to a ~77px visible strip, and because the overlay itself no longer scrolls that strip is all you get (before this PR you could scroll the whole overlay past the hint). Adding a width-based branch to the same container query would fix it.

**What I exercised.** Local server, Chromium via Playwright, PR branch vs `main` side by side, at **390×844, 820×1180, 1280×800 and 1920×1080**. Picker hosts: room overlays (edit mode, properties panel → Icon → *Show all*), `#editor` (deck editor opened once and closed, then properties panel), and the deck-editor card column (`Face` tab → add-icon). Also opened: deck editor itself, JSON editor (incl. the `#jeWidgetLayers` change), widget properties panel, Game Shelf, Players and About overlays. Not reached: the sound picker in the `#editor` host (its button sits in a collapsed section I couldn't open reliably from script) and the rich-text symbol picker — both share the changed `fonts.css`/`deckeditor.css` rules, so they are worth a manual look; I reviewed those two from the CSS only.

---

## Big layout improvements

### 1. The height-only breakpoint misses the narrow-box case — icon list unusable at 820×1180 in the deck-editor card column

`client/css/fonts.css` hides the hint at `@container symbolPicker (max-height: 375px)`, and `#symbolList` gets `flex: 1; min-height: 120px`. In the deck-editor card column at 820×1180 the picker box is **204 × 931** — tall, so the breakpoint does not fire, but so narrow that the hint paragraph wraps to one or two words per line and eats ~700px. Measured on this branch: `#symbolList` sits at `y=916, height=120` while the overlay ends at `y=993` → about 77px of list visible, the rest clipped, and `#symbolPickerOverlay` reports `scrollHeight == clientHeight` so the overlay can no longer be scrolled to compensate. On `main` the same spot reported `scrollHeight 104204 / clientHeight 931`, i.e. you could scroll down past the hint to the icons.

| this branch, 820×1180, deck-editor card column | same spot on `main` |
|---|---|
| after | before |

Suggestion, in the spirit of the existing rule — add the missing dimension to the same container query:

```css
@container symbolPicker (max-height: 375px), symbolPicker (max-width: 500px) {
  #symbolPickerOverlay p[icon] { display: none; }
  #symbolPickerOverlay input  { margin-bottom: 20px; }
}
```

Or, more robustly, stop letting the hint win the height fight at all: `#symbolPickerOverlay p[icon] { flex: 0 1 auto; overflow: hidden; max-height: 4.5em; }`. That makes the picker degrade gracefully in *any* awkward box instead of only the two shapes the breakpoints anticipate.

### 2. The 120px floor overflows a box that cannot hold it, and nothing scrolls to recover

`min-height: 120px` is a good idea, but when the box genuinely cannot afford it the excess is simply clipped, because the wrapper chain now carries `min-height: 0` and `.overlay { overflow: auto }` never sees the overflow. Measured:

- room-overlay host, 390×844: overlay `y 623 … 844`, `#symbolList` `y 750 … 870` → 26px of the list below the viewport edge, unreachable;
- deck-editor card column, 820×1180: 43px clipped, as above.

Visible as the half-cut bottom icon row here: room host 390

Cheapest way to buy back that space rather than clip it: give the picker its own padding instead of inheriting `.overlay { padding: 5% 10% }`. Percentage padding resolves against the box **width**, which is now doing something odd in the new boxes — at 1920 the `#editor` box is 1792 wide, so the picker gets ~90px of *vertical* padding it does not need, while at 390 it gets ~18px. A fixed `#symbolPickerOverlay { padding: 20px 10px; }` would both stop the clipping at small sizes and remove the large dead band above the title at 1920 (editor host 1920).

### 3. On a phone the room-overlay host is still ~1.5 rows of icons

Not a regression — it is clearly better than `main`, where the same spot showed *zero* icons (only the hint) — but at 390×844 the picker is a 221px-tall sliver at the bottom of the screen and picking an icon means scrolling a 120px window over 40 000px of content, with the properties panel still visible above it. This overlay is one of the few that genuinely wants the whole viewport. Since this PR already establishes the pattern of giving the picker a box of its own with an explicit `z-index`, consider extending it: in the room-overlay host, when the box is below some height, render it `position: fixed; inset: 0;` with a `z-index` above `#roomArea` — the same escape hatch, just applied to the host where it is now most needed.

room host 390 scrolled

---

## Nitpicks / small breakages

- **Title clipped by the close button in a narrow box.** At 820×1180 in the card column the `h1` renders as “Pick ico” with the × sitting on top of it (see screenshot in §1). `.titleWithCloseButton h1` needs `min-width: 0; overflow: hidden; text-overflow: ellipsis;` (or a smaller font-size in the narrow container branch).
- **The content column is centred on the box, not the window.** `#editor > #symbolPickerOverlay` is `left: 0; right: var(--editSidebarWidth)`, and the 1000px column is `margin: 0 auto` inside that, so at 1920 its centre lands at x≈896 vs a window centre of 960 — the modal reads as slightly left-of-centre next to the symmetric toolbar above it. Minor, but easy to fix by dropping `left: 0` in favour of matching the sidebar offset, or by accepting it deliberately.
- **`#symbolPickerOverlay > div:not(.titleWithCloseButton)`** (fonts.css) is a broad structural selector: it happens to match exactly one element today, but any future sibling (a footer, a toolbar) silently inherits `flex: 1` and starts competing with the list for height. Scoping it to `> div[icon]` or a class would keep the intent readable.
- **Layout jump when the hint disappears.** `.fewResults` hides the hint, so typing a search that narrows below 100 results shifts the whole list up by ~50px while the user is mid-scan. Reserving the space (`visibility: hidden` instead of `display: none`) — or removing the hint from the flow entirely, per §1 — would keep the list steady.
- `body.overlayActive #jeWidgetLayers` (jsonedit.css) behaves exactly as described — verified: with the JSON editor open and the cursor over the room the panel shows, and it disappears the moment the picker opens ([before](https://agent.virtualtabletop.io/reports/ui-review/1535040137854058536/je-layers-visible.png) / [after](https://agent.virtualtabletop.io/reports/ui-review/1535040137854058536/je-layers-hidden.png)). No complaints.
- Game Shelf, Players and About overlays render unchanged at all four viewports, as expected for a diff scoped to `#symbolPickerOverlay` / `#jeWidgetLayers`.

---

- **The hint is the wordiest element in the dialog and it is now the one crowding out the content.** “For performance reasons, game-icons are displayed in low resolution and emoji are displayed in your browser font. Filter to less than 100 images or hover over an icon to show the actual image that will be used.” is developer-facing (“for performance reasons”, “filter to less than 100 images”) and a newcomer mostly needs one fact from it. Something like *“Icons are previewed in low resolution — hover one to see how it will really look.”* on a single line would say the useful part, and would remove most of the pressure behind §1 and §2. Better still: move it behind the little ⓘ affordance the properties panel already uses everywhere, next to the search field.
- **No empty state.** Search for something with no matches and you get a large blank grey card with no explanation — and this PR makes the empty box more prominent, because the card now holds its height instead of collapsing to content. A centred “No icons match “zzzqqq”.” would cost one line.
  no results
- **No “what do I do here” line.** The sibling sound picker tells you (“Click the arrow to preview a sound, click its name to select it.”); the icon picker never says that clicking an icon picks it and closes the dialog. One short sentence in the same place, in the same voice, would match the app's convention.
- **Search field affordance.** The placeholder is just “Search”, in front of ~20 000 icons grouped under source-name headings (“Game-icons.net - Abstract”, “VTT Icons - Internal UI”) that mean nothing to a newcomer. A placeholder like “Search icons (sword, heart, dice…)” and a result count next to it would orient a first-time user far more than the headings do.
- **Lost context on the `#editor` host.** Now that the picker fills the editor box opaquely, nothing on screen says which property you are picking for — you left the properties panel and land on a full-width “Pick icon” page. Titling it “Pick icon for *Icon*” (or keeping the originating row visible) would keep the thread. This is a consequence of the fix, not a flaw in it — worth a follow-up rather than a blocker.

VTT-AI-Agent:
Thanks — the narrow-box case reproduced exactly (204×931 box, hint 703px, list 120px with 43px of it clipped and nothing scrollable), and it is fixed together with the clipping, the padding and most of the readability points in cd0b8671. Details below, including the three I deliberately left out.

Measured again on the branch, `#symbolList` clientHeight, all three hosts, same four viewports:

| host | viewport | picker box | before | after |
|---|---|---|---|---|
| deck editor card column | 820×1180 | 204×931 | **120** (43px clipped, unreachable) | **776** |
| deck editor card column | 1280×800 | 664×539 | 236 | 282 |
| deck editor card column | 1920×1080 | 1212×819 | 499 | 581 |
| room overlays | 390×844 | 354×221 | **120** (26px clipped, unreachable) | 169 (+ overlay scrolls) |
| room overlays | 820×1180 | 784×490 | 175 | 252 |
| room overlays | 1280×800 | 640×400 | 196 | 245 |
| room overlays | 1920×1080 | 960×600 | 287 | 362 |
| `#editor` | 390×844 | 354×772 | 483 | 641 |
| `#editor` | 820×1180 | 784×1156 | 838 | 918 |
| `#editor` | 1280×800 | 1244×764 | 438 | 526 |
| `#editor` | 1920×1080 | 1792×1044 | 654 | 806 |

No host at any of these sizes loses height, and nothing is clipped any more.

### 1. Height-only breakpoint → the card column at 820×1180 — fixed

Added the missing dimension to the same container query, as you suggested:

```css
@container symbolPicker (max-height: 375px), symbolPicker (max-width: 600px) {
```

600px is the width the media queries below already use, so it is the same pair of breakpoints in both places. In that column the hint now goes away and the list gets 776px instead of a 77px strip. The 784px-wide room-overlay box at the same viewport stays above the breakpoint and keeps its hint (76px, three lines) — which is what I wanted: the hint should only be sacrificed where it is actively in the way.

I skipped the `max-height: 4.5em; overflow: hidden` variant on purpose: it truncates the sentence mid-word rather than dropping it, and with the width branch in place there is no shape left in these three hosts where the hint wins the height fight.

| before, 820×1180, card column | after |
|---|---|
| before | after |

### 2. The 120px floor being clipped, and the percentage padding — both fixed

You had the mechanism right, and it is worth spelling out because it is not obvious: the excess was not lost to `.overlay { overflow: auto }` being absent (it is there), it was lost to `div[icon] { overflow: hidden }` in `layout.css` — the card clips, so the part of the list that did not fit never became scrollable overflow at all.

So the fix is to stop squeezing the card below what it holds:

```css
#symbolPickerOverlay div[icon] {
  min-height: 210px;   /* its own padding + the search field + #symbolList's 120px floor */
}
```

The plain wrapper around the card does *not* clip, so a box that cannot even afford that now overflows the overlay and the overlay scrolls — the degraded-but-complete fallback — instead of dropping the bottom of the list on the floor. At 390×844 in the room host that is exactly what happens: 169px of list visible and the remainder reachable by scrolling the overlay, where before it was 120px with 26px permanently off-screen.

(I first tried `overflow: visible` on the card, which is the more obvious fix — it lets the 250px `div[icon]::before` watermark glyph escape the rounded card and float over the white overlay. Not worth it.)

Padding: agreed, and the numbers you gave are the whole argument. `#symbolPickerOverlay` now has `padding: 20px 10px` instead of inheriting `.overlay { padding: 5% 10% }`, so ~90px of dead band above the title at 1920 becomes 20px and the horizontal padding stops mattering (the column is `max-width: 1000px; margin: 0 auto` anyway). That alone is +152px of list in the `#editor` host at 1920 and +88px at 1280.

| before, `#editor` 1920×1080 | after |
|---|---|
| before | after |

### 3. Phone room-overlay host as a full-viewport modal — not doing this one

Confirmed the observation (354×221 box at 390×844, and yes it is much better than `main`, which showed no icons there at all), but I don't want to special-case it in this PR:

- every overlay in `#roomArea` is constrained to the room box — Game Shelf, Players, About, the sound picker. Letting *one* of them escape to `inset: 0` makes the picker behave unlike its own siblings, including the sound picker it sits next to, and the room box is deliberately what the room's overlays are scoped to.
- edit mode at 390px wide is behind the "please rotate your device" state anyway, so the box this small is mostly a landscape-phone / very short window case, where a full-viewport modal is much less obviously right.

If a maintainer wants the picker to be the exception, that is a deliberate product call and I'd rather it were made explicitly (and applied to the sound picker at the same time) than smuggled into a layering fix.

### Nitpicks

- **Title clipped to "Pick ico"** — fixed. Inside the same narrow/short container branch the title is now `font-size: 20px; margin: 0 40px 10px`, so it keeps clear of the close button instead of running under it. At 820×1180 in the card column the `h1` box goes from 163px (38px font, overlapping the ×) to 104px at 20px, with the full "Pick icon" visible — see the after screenshot in §1.
- **Column centred on the box, not the window** — left as is, deliberately. The picker is scoped to the editor's *content* box, which is also where `#deckEditorNewDeckOverlay`, the import/export dialogs and the sound picker centre themselves; matching the window instead would make this one overlay sit differently from every other thing that opens in that box. If the editor box itself should be visually centred under the toolbar, that is a `deckeditor.css` question for all of them at once.
- **`> div:not(.titleWithCloseButton)`** — kept, because it is the house idiom for exactly this markup: `states.css:1175` and `:1220` use the identical selector on `#stateAddOverlay` / `#variantAddOverlay`, which have the same title-plus-body structure. Inventing a class for one of the three would make the picker the odd one out.
- **Layout jump when `.fewResults` hides the hint** — pre-existing (`#symbolPickerOverlay.fewResults p { display: none }` is on `main`), and `visibility: hidden` would permanently reserve the 38–133px this PR is spending its effort reclaiming. Left alone.
- **`body.overlayActive #jeWidgetLayers`** — thanks for verifying it.

### Readability

- **The hint** — rewritten, and it now leads with the useful part rather than the caveat: *"Click an icon to pick it. Emoji use your browser font and game-icons are previewed in low resolution: hover one, or narrow the search below 100 results, to see the real image."* That folds in your "no 'what do I do here' line" point in the same place and the same voice as the sound picker's, drops "for performance reasons", and is one line shorter at every width. The `<span>` still carries the game-icons half, which is hidden outside `bigPreviews` as before, and both readings are grammatical.
- **Empty state** — added. `symbols.js` now sets a `noResults` class alongside `fewResults`, and a `#symbolNoResults` element takes the list's place with `No icons match "zzzqqq".` (the actual search term, via `textContent`), instead of a blank card.
  no results
- **Search placeholder** — now `Search icons (sword, heart, dice, …)`.
- **Result count next to the search field** — skipped. It wants a line of its own next to or under the input, which is the one place in this dialog where vertical space is scarce, and the count is only interesting while typing. The empty state covers the case where the absence of a count is actually confusing.
- **"Pick icon for *Icon*"** — agreed it is the right thing, not done here. `pickSymbol()` has no idea which property it was opened for; the caller does (`pickSymbolKeepingOverlay` / `propertyInputs.js`), so it needs a title argument threaded through both entry points and a reset for the rich-text path that has no property at all. That is a small API change to a function called from four places, and you flagged it as a follow-up rather than a blocker — I'd rather it landed on its own.

### Testing

- `npm test` — 590 passed.
- `npx testcafe chromium:headless tests/testcafe/editor.js` — 32 passed (including "Deck editor: symbol pickers and JSON fallback").
- Browser: the table above, measured in all three hosts at four viewports; searched for a term with no matches, one with few matches and one with many; scrolled the list and confirmed the search field and close button stay put; picked an icon and reopened the picker (starts at the top); checked at 1920 that the watermark glyph is still clipped inside the card.

One thing your run and mine agree on that is *not* this PR: at 390×844 the deck editor's own card column measures 0px wide, so the picker parked in it has nothing to be laid out in. That is the deck editor's column layout at phone width, untouched by this diff (the box comes from `body.deckEditorRoomVisible #deckEditorMainCol > .overlay`), and it is the same before and after.

VTT-AI-Agent (commented):
## TL;DR

The latest commit fixes the short- and narrow-host regressions raised in the first two passes, but the new empty-search state does not account for the picker's existing image-only/font-only filters. As a result, some zero-result searches still show the same unexplained blank card this change is intended to eliminate.

## Findings

### 1. `client/js/symbols.js:117-120` — the empty state counts icons that the active picker type hides

`matches` counts every icon that matches the search text, but `pickSymbol()` applies the `images`/`fonts` restriction separately through the overlay's `hideFonts` and `hideImages` classes (`symbols.js:132-134`). Those classes hide entire icon families in CSS (`client/css/fonts.css:556-565`) without adding `.hidden`, so they are invisible to the new count.

Concrete examples from the current `assets/fonts/symbols.json`:

- Open an image-only picker (`pickSymbol('images')`) and search for `10k`. One Material symbol matches and zero image icons match. `matches` is still 1, so `.noResults` is not applied, while CSS hides the sole match and the card is blank.
- Open a font-only picker (`pickSymbol('fonts')`) and search for `abbot`. One game-icon matches and zero font icons match, producing the same blank state.

Please derive the count from icons that pass both the text search and the active type filter (or represent the type filter in classes/attributes that the count can exclude), then use that visible count for both `fewResults` and `noResults`. A focused check for an image-only and a font-only zero-result search would prevent this edge case from returning.

## Scope checks

This PR adds no widget property, routine operation, or other new game-author API, so no visual authoring control or `library/tutorials/` addition is required. The changed-file list contains no engine paths (`client/js/widgets/`, `client/js/compute.js`, or `client/js/serverstate.js`).

The `beta` label was already present; I left it in place because this UI-only diff does not touch engine behavior and the remaining empty-state bug is safe and useful to test on beta.

ArnoldSmith86 (approved):

VTT-AI-Agent:
Thanks — both of your examples reproduced exactly, and they are fixed in fe1734b4.

### 1. The empty state counting icons the type filter hides — fixed

Confirmed on the branch, driving the real pickers (deck editor → *Image* for `pickSymbol('images')`, the properties panel's *Pick a symbol* button for `pickSymbol('fonts')`), and your two search terms are indeed the only ones of their kind: `10k` matches exactly one Material symbol and nothing else, `abbot` exactly one game-icon.

| picker | search | text matches | actually visible | before | after |
|---|---|---|---|---|---|
| images only | `10k` | 1 | **0** | list shown, blank card | `noResults`, *No icons match "10k".* |
| fonts only | `abbot` | 1 | **0** | list shown, blank card | `noResults`, *No icons match "abbot".* |
| images only | `sword` | 152 | 60 | no `fewResults` → 24px sprite previews | `fewResults` → real images |
| fonts only | `sword` | 152 | 92 | no `fewResults` | `fewResults` |
| unrestricted | `10k` / `abbot` | 1 | 1 | one match, `fewResults` | unchanged |
| unrestricted | `zzzqqq` | 0 | 0 | `noResults` | unchanged |

So the same count was also driving `fewResults` past its threshold — a restricted search of "152 matches, 60 of them visible" kept the low-resolution sprite previews on. Both classes now come from the visible count.

| before, images-only picker, search `10k` | after |
|---|---|
| before | after |

and the fonts-only picker searching `abbot`: [before](https://agent.virtualtabletop.io/reports/pr/1535013339619655913/review-3/before-emptystate-fonts-abbot.png) / [after](https://agent.virtualtabletop.io/reports/pr/1535013339619655913/review-3/after-emptystate-fonts-abbot.png).

**How.** You offered two routes; I took the second one, because repeating the CSS family list as a selector string in `symbols.js` would leave two places to keep in sync the next time an icon family is added. Every icon and category heading now carries `data-family="image"` or `"font"`, so the restriction is one pair of selectors instead of nine:

```css
#symbolPickerOverlay .hidden,
#symbolPickerOverlay.hideImages [data-family=image],
#symbolPickerOverlay.hideFonts [data-family=font] {
  display: none;
}
```

and the count reads the same attribute, so it can only disagree with the CSS if both are changed:

```js
const hiddenFamily = $('#symbolPickerOverlay').classList.contains('hideImages') ? 'image'
                   : $('#symbolPickerOverlay').classList.contains('hideFonts')  ? 'font' : null;
const matches = $a(`#symbolList i:not(.hidden)${hiddenFamily ? `:not([data-family=${hiddenFamily}])` : ''}`).length;
```

The per-icon classes (`gameicons`, `material-symbols`, …) are untouched — they still carry the styling and `addRichtextControls` still reads them; only the two heading classes `imageCategory`/`fontCategory`, which existed for nothing but this hiding, are gone. Verified that the filtering itself is unchanged: images-only shows 57 headings / 6029 icons and no font ones, fonts-only 32 / 7259 and no image ones, unrestricted all of them.

### 2. The focused checks you asked for — added

Both fail on `cd0b8671` and pass on `fe1734b4` (I ran them against the parent commit to be sure they are not vacuous):

- `Deck editor: symbol pickers and JSON fallback` — in the images-only picker, searching `10k` now has to show `#symbolNoResults` and hide `#symbolList`; clearing the search brings the list back. The unrestricted picker right below it gets the same search as the counter-check and still finds its one match.
- `The symbol picker says an image-only search found nothing` (new) — the fonts-only picker from the properties panel: `abbot` shows the empty state with the search term in it, then `10k` shows exactly one icon again.

### Testing

- `npm test` — 590 passed.
- `npx testcafe chromium:headless tests/testcafe/editor.js` — 33 passed.
- Browser: the table above, plus the family/heading counts in all three modes.

PR #3106, opened 2026-08-06
Render each nested routine block once instead of twice per nesting level

VTT-AI-Agent:
The Automations (routine) editor rendered every nested IF/FOREACH block **twice**, so the cost of opening a routine was `2^depth` rather than proportional to the number of operations. A routine nested 13 levels deep produced **8191** operation renders instead of 13. This makes a deeply nested routine open essentially instantly again, with no visual or functional change.

## The cause

`RoutineEditor`'s constructor already renders the block into `this.domElement` (via `setRoutine` → `render`), but `renderSubroutine()` then called `render()` a second time just to get an element to append:

```js
const routineEditor = new RoutineEditor(...);   // renders the whole subtree
...
dom.append(routineEditor.render());             // renders the whole subtree again
```

Every nesting level therefore doubled the work of everything below it. Appending the element the constructor already built removes the second pass; `render()` returns that very same `domElement` after clearing it, so the result is identical.

Profiling (Playwright + CDP sampler) showed no single hot function — the time was spread evenly over `renderSentenceView`, `div`, `createElement`, `addEventListener`, `parseVarStatement` and GC, i.e. ordinary rendering work simply done ~1600× too often. Instrumented render counters confirmed it: **87 cards on screen, 142 729 operation renders**.

## Measured before/after

Headless Chromium, `performance.now()` around the click that expands a routine in the Automations panel (median of 3):

| routine | operations / cards | before | after |
|---|---|---|---|
| the reported example (`sortRoutine`, 8 levels of nesting) | 87 cards | **17 195 ms** | **31 ms** |
| a flat routine of the same length (89 `var` statements) | 89 cards | 9 ms | 12 ms |
| `Rails Before the War` – `modify_values_anchor.onlyVisibleForSeatChangeRoutine` | 80 cards | **600 ms** | **39 ms** |
| `Rails Before the War` – `1_auction_anchor.statusChangeRoutine` | 60 cards | **585 ms** | **25 ms** |
| `Celtic Knots` – `useful.validateHoldersRoutine` | 32 cards | 76 ms | 10 ms |
| all 21 routines of that test room, summed | | **1438 ms** | **124 ms** |

(The reported ~11 s reproduces as ~17 s in headless Chromium in this sandbox — same effect, slower machine.)

## Verification

- The rendered DOM is **byte-identical** before and after: dumped the full `innerHTML` of the expanded Automations panel for the example routine plus **21 real routines** from `Rails Before the War`, `Celtic Knots` and the `Functions - CALL` tutorial, and diffed — no difference.
- Editing still works from deep inside a nested block: parameter chips open their popups, "add operation" opens the operation list, and the outdent button hoists an operation out of the innermost block and writes it back to the widget.
- `npm test` (jest): 1154 passed, including a new regression test that renders a 13-level-deep routine and asserts one block render per nesting level (it reports 8191 without the fix).
- TestCafe `editor.js` (32 tests) and `routines.js` (4 tests) pass locally against Chromium headless.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3106/pr-test (or any other room on that server)

### DIFF
--- a/client/js/editor/controls/routine.js
+++ b/client/js/editor/controls/routine.js
@@ -3276,7 +3276,10 @@ class RoutineOperationEditor {
       this.notifyChangeListeners(this.operation);
     });
     this.subroutineEditors[property] = routineEditor;
-    dom.append(routineEditor.render());
+    // the constructor already rendered the block into domElement - rendering it a
+    // second time here would do the whole nested subtree twice per level, which
+    // is 2^depth renders for a deeply nested routine
+    dom.append(routineEditor.domElement);
   }
 
   resolveParameter(property) {
--- a/tests/client/routineeditor.test.js
+++ b/tests/client/routineeditor.test.js
@@ -1526,18 +1526,21 @@ describe('routine editor state handling', () => {
     let routine = [ { func: 'FLIP' } ];
     for(let level = 0; level < 12; level++)
       routine = [ { func: 'IF', operand1: 1, thenRoutine: routine } ];
-    const original = RoutineOperationEditor.prototype.render;
+    // counting on RoutineEditor rather than on the operation editors: it is the
+    // one class that renders a block, so no subclass can render without being
+    // counted here
+    const original = RoutineEditor.prototype.render;
     let renders = 0;
-    RoutineOperationEditor.prototype.render = function(...args) {
+    RoutineEditor.prototype.render = function(...args) {
       renders++;
       return original.apply(this, args);
     };
     try {
       new RoutineEditor({ state: {} }, routine);
     } finally {
-      RoutineOperationEditor.prototype.render = original;
+      RoutineEditor.prototype.render = original;
     }
-    expect(renders).toBe(13); // one per operation
+    expect(renders).toBe(13); // the routine itself plus one per nested block
   });
 
   test('ignores echoes of its own edits but applies remote changes', () => {
--- a/tests/client/routineeditor.test.js
+++ b/tests/client/routineeditor.test.js
@@ -1520,6 +1520,26 @@ describe('routine editor state handling', () => {
     expect(foreachEditor.domElement.textContent).toContain('Add operations to run for each iteration');
   });
 
+  test('a nested block is rendered once per level, not once per level per level above it', () => {
+    // a block that renders itself again after its editor built it costs 2^depth
+    // renders, which is what made a deeply nested routine take seconds to open
+    let routine = [ { func: 'FLIP' } ];
+    for(let level = 0; level < 12; level++)
+      routine = [ { func: 'IF', operand1: 1, thenRoutine: routine } ];
+    const original = RoutineOperationEditor.prototype.render;
+    let renders = 0;
+    RoutineOperationEditor.prototype.render = function(...args) {
+      renders++;
+      return original.apply(this, args);
+    };
+    try {
+      new RoutineEditor({ state: {} }, routine);
+    } finally {
+      RoutineOperationEditor.prototype.render = original;
+    }
+    expect(renders).toBe(13); // one per operation
+  });
+
   test('ignores echoes of its own edits but applies remote changes', () => {
     const routine = [ { func: 'FLIP' } ];
     const editor = new RoutineEditor({ state: {} }, routine);


VTT-AI-Agent (commented):
## TL;DR

Looks good — the one-line change is correct and the fix is the right one: `RoutineEditor.render()` returns the very `domElement` the constructor already filled, so appending `routineEditor.domElement` is exactly equivalent minus the redundant second full-subtree render. I found nothing to block on; the only note below is a robustness caveat on the new test's counting technique.

## Findings

**Correctness of the fix — verified, no issues.**
- `RoutineEditor`'s constructor unconditionally reaches `setRoutine()` (`client/js/editor/controls/routine.js:2015`), which ends in `this.render()` (`:2084`), and `render()` returns `this.domElement` (`:2194`). So the appended node is identical either way, and the nested editor's own later `render()` calls still repaint in place because the element stays parented.
- Nothing is registered between `new RoutineEditor(...)` and the old `render()` call that `render()` reads: `registerChangeListener` only pushes into `changeListeners`, and `render()` never notifies. `this.subroutineEditors[property]` is written by the *child's* render, not the parent's. So dropping the second pass loses no state.
- It also aligns the nested path with the top-level one — `client/js/editor/controls/events.js:487` already does `contentDOM.append(this.routineEditors[key].domElement)`. Good consistency.

**Regression test — genuinely guards the bug.** I ran it both ways: with the fix `renders === 13`, and after reverting the one line locally it reports `8191` (and takes ~17.7 s), so it is not vacuous. `npm test -- tests/client/routineeditor.test.js` → 310 passed on this branch.

**Minor (non-blocking): the counter patches only the base prototype.**
`tests/client/routineeditor.test.js:1529` wraps `RoutineOperationEditor.prototype.render`, but the container editors override `render()` (`routine.js:3317`, `:3362`). It works today only because those overrides call `super.render()`. A future operation editor whose `render()` does not delegate to `super` would silently not be counted, and the assertion `toBe(13)` would then be under-counting rather than failing loudly. Not worth changing on its own, but if you touch it, counting `renderSubroutine` calls (or the number of `.routine-editor-operation` elements created) would be less coupled to the override chain.

**Scope checks:** no new widget property and no new routine function/operation, so no editor-UI, `jeCommands` or `validator/` work is implied; no tutorial is needed and none was touched; no other tutorials were modified. Diff is 2 files, one concern.

I added the `beta` label: the change touches only `client/js/editor/` (routine editor UI) and a test — no engine file (`client/js/widgets/`, `compute.js`, `serverstate.js`) is involved, so the PR is beta-eligible, and a purely-equivalent DOM-append change with a passing suite is safe and useful for players to test.

VTT-AI-Agent:
Addressed the one note from the automated code review — the regression test now counts on `RoutineEditor.prototype.render` instead of `RoutineOperationEditor.prototype.render`.

`RoutineOperationEditor.render` is overridden by the IF and FOREACH editors, so the old counter only saw them because those overrides call `super.render()`; an override that stopped delegating would have made the count too low and turned the assertion into a silent under-count. `RoutineEditor.render` is the single method that renders a block and the class has no subclasses, so it measures the doubling directly. Re-checked both ways: 13 with the fix, 8191 (and ~16 s) with the one-line change reverted.

Full `npm test` on the current head: 1154 passed, 26 suites. All 41 CI checks were green on the merge of `main` that preceded this commit.

VTT-AI-Agent (commented):
## TL;DR

The fix is correct and minimal: `RoutineEditor`'s constructor already renders into `domElement` via `setRoutine` → `render()`, so `renderSubroutine` calling `render()` again was pure duplicated work that compounded per nesting level. Appending `routineEditor.domElement` matches what the other construction site (`client/js/editor/controls/events.js:487`) already does. I found no correctness problem — only a small optional test-coverage suggestion.

## Findings

**Correctness of the change — verified, no issues.**

- `client/js/editor/controls/routine.js:2015` → `setRoutine()` (`:2083`) → `render()` (`:2087`), so `domElement` is fully populated when the constructor returns. `render()` starts with `this.domElement.innerHTML = ''` and re-appends into the *same* node, so `dom.append(routineEditor.render())` and `dom.append(routineEditor.domElement)` append the identical element — the only difference is the discarded first pass. The claim of an unchanged DOM holds up on reading.
- The only side effect `render()` has beyond building DOM is `restoreActiveOperation()` (`:2234`), which is idempotent (it only reads `activeOperationByWidget`, and its one mutation — deleting a stale entry — is a no-op on a second call). So dropping the second call changes no state.
- Ordering is safe: `subroutineEditors[property]` is assigned and the change listener registered before the append; neither is read during rendering. Previously the second `render()` re-entered `renderSubroutine` and *overwrote* `subroutineEditors[property]` with a second editor instance — the attached DOM and the stored editor stayed in sync then, and they stay in sync now, so the "edit from deep inside a nested block" paths (hoist, parameter popups, add-operation) still operate on the editor whose DOM is on screen.
- All three call sites (`:3329` thenRoutine, `:3335` elseRoutine, `:3367` loopRoutine) funnel through the fixed helper, so there is no remaining doubling path.

**Regression test — genuinely pins the bug.** I re-applied the old `dom.append(routineEditor.render())` locally and the new test fails with the expected `8191` (and the suite time jumps from ~7 s to ~18 s); with the fix, `npm test` is 1154/1154 green here. Counting on `RoutineEditor.prototype.render` rather than on the operation editors is the right hook, and the comment explaining why is welcome.

**Minor (optional): the test only exercises `IF`/`thenRoutine`.** `tests/client/routineeditor.test.js:1523` builds a chain of `IF` with only a `thenRoutine`. Since `FOREACH`/`loopRoutine` and the `elseRoutine` branch go through the same `renderSubroutine`, coverage is adequate — but a variant where each level has *both* `thenRoutine` and `elseRoutine` would pin the worse-than-`2^depth` case (each level fans out to two blocks, so the old code grew as `3^depth`), and a `FOREACH` level would pin the second container type. Cheap to add, not blocking.

Nothing else to raise: no new widget property, routine function or user-facing feature is introduced, so no editor-UI surface and no tutorial are required for this change.

I added no label — `beta` was already present and I left it: this PR touches only `client/js/editor/controls/routine.js` (edit-mode routine editor) and a test, no engine file (`client/js/widgets/`, `compute.js`, `serverstate.js`), and it is a pure render-count reduction with byte-identical output, so it is safe and useful for players to test.

96LawDawg (approved):
Minor change to fix UI recursion.  Tested.  Approved.  Will merge now.

PR #3107, opened 2026-08-07
Don't treat a non-card widget's "deck" property as a deck reference

VTT-AI-Agent:
Removing a widget resolved its `deck` property through the widget map and called `removeCard()` on whatever it found. `deck` only means "the deck this card belongs to" on **card** widgets — every other widget can carry a property of that name, since routines and the JSON editor let a game set any property on any widget. Removing such a widget threw `widgets.get(...).removeCard is not a function`.

This came in through the client crash reporting from the main server (Chrome 151). In the reported state, a number of button widgets carried a `deck` property naming a holder.

The crash lands in `receiveStateFromServer()`, which tears the old room down with `applyRemoveRecursive()` outside of any try/catch (the delta path in `removeWidget()` does catch it). So the teardown loop aborts halfway: the widgets of the previous game stay in the DOM, the incoming state is never applied, and the room is stuck until the player reloads the page.

## The fix

`Card` already keeps the deck it registered with in `this.deck` — that is what `Card.applyDeltaToDOM()` and `renderReadonlyCopyRaw()` use. `applyRemove()` now uses it too instead of resolving the property again, so the deck bookkeeping only ever runs for widgets that actually joined a deck. Nothing changes for real cards: `this.deck` is the same object the lookup returned.

As a side effect this also stops a non-card widget that names a *real* deck from being deregistered from it on removal — it was never registered there, but `removeCard()` would still decrement that deck's card counter.

## Reproduction on main

Reproduced against a local checkout of `main` (commit f07aa6f) with a plain browser session:

1. Open a room.
2. Load a state containing a holder and a button whose `deck` property is set to that holder's id — e.g. `PUT /state/<room>` with
   ```json
   {
     "_meta": { "version": <version>, "gameSettings": {} },
     "aHolder": { "id": "aHolder", "type": "holder", "x": 100, "y": 100 },
     "zButton": { "id": "zButton", "type": "button", "x": 300, "y": 100, "text": "old", "deck": "aHolder" },
     "zzLabel": { "id": "zzLabel", "type": "label", "x": 500, "y": 100, "text": "old label" }
   }
   ```
   (In the UI: add a button in edit mode and give it a property called `deck` whose value is a holder's id.)
3. Load a different game / state into the same room, i.e. make the server send a new complete state.

Before this change, step 3 throws `TypeError: widgets.get(...).removeCard is not a function` in `Button.applyRemove`, and the widgets of the first state (`w_zButton`, `w_zzLabel`) are still on the table while the second state's widgets are missing. With this change the room switches cleanly and no error is raised.

## Tests

- New `tests/client/widget-remove.test.js`: removing a widget whose `deck` names a non-deck does not throw and does remove the widget, such a widget is not deregistered from a deck it names, and a real card is still deregistered from the deck it joined. The first two fail on `main`.
- `npm test` passes (1474 tests) with `main` merged in.
- TestCafe `interactions.js` passes locally.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3107/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

The fix is correct and minimal: `Card` is the only widget that ever registers with a deck, and it already stores that deck in `this.deck`, so using it in `applyRemove()` removes the crash without changing anything for real cards. My main point is about the tests — the "still deregisters a card" case never exercises a real `Card`, so the invariant the fix now relies on (`Card.applyDeltaToDOM` keeps `this.deck` in sync) is not actually covered.

_Automated review #1 of this PR, performed by Claude (claude-opus-5)._

## Findings

**1. `tests/client/widget-remove.test.js:43` — the card test doesn't test a card.**
`createWidget()` in `tests/client/client-util.js:7` always constructs a base `Widget`, whatever `type` the definition carries, and the test then sets `card.deck = deck` by hand. So the assertion only proves that `Widget.applyRemove()` reads `this.deck` — it would pass even if `Card` never set that field at all. The regression this fix newly depends on is exactly the opposite direction: if some path ever leaves a real card with a `deck` property but `this.deck === null`, the deck's card map and its counter now silently leak instead of being cleaned up. Worth constructing a real card (`new Card(id)` + `addWidget(def, card)` against a real `Deck`, the way `serverstate.js:72-85` does) so `Card.applyDeltaToDOM()` performs the `addCard()` and the test covers the register→remove round trip rather than a hand-set field. `tests/client/pile-handle.test.js` and friends already instantiate concrete widget classes, so the pattern exists.

**2. `client/js/widgets/widget.js:2903` — the sibling spot still resolves the deck the old way.**
`renderReadonlyCopyRaw()` does `if(this instanceof Card) this.deck.removeCard(this)`. That is the same bookkeeping call, guarded by the widget's class instead of by `this.deck`, and it throws on a `Card` whose `deck` is null. Today that is close to unreachable (`serverstate.js:73` refuses to instantiate a card whose deck isn't a real deck), but now that `applyRemove()` has settled on `if(this.deck)`, making this line match would be one character of churn and removes the remaining shape where the two paths can disagree.

**3. Out of scope, but the same root confusion bites elsewhere for the reported state.** Worth a follow-up issue rather than this PR:
- `client/js/serverstate.js:593` — `removeWidgetLocal()` cascades to every widget with `w.get('deck') == widgetID`. In the reported state (buttons carrying `deck: <holderID>`), deleting that holder in edit mode silently deletes those buttons too.
- `client/js/widgets/widget.js:1774` — RECALL collects `widgetFilter(w=>w.get('deck')==deck.get('id'))` with no type check, so a non-card widget naming a real deck gets recalled into the holder as if it were a card.

Both are much less severe than the crash you fixed (no exception, no stuck room), but they're the same "`deck` on a non-card" assumption.

**4. No new author-facing surface.** The PR adds no widget property and no routine function, so the editor-UI and `library/tutorials/` requirements don't apply here; a jest unit test is the right coverage for this one, and it correctly fails on `main`. Nothing to do.

Nice write-up on the reproduction and on why the crash leaves the room stuck (teardown in `receiveStateFromServer()` being outside the delta path's try/catch) — that's the part that makes the severity clear.

I did not add the `beta` label: the change is in `client/js/widgets/widget.js` and alters the widget-removal path that every game goes through, so it's a real engine change and excluded from beta auto-deployment by design.

ArnoldSmith86 (approved):

PR #3109, opened 2026-08-09
Update the color and monochrome Noto Emoji sets to Emoji 17.0

VTT-AI-Agent:

- [ ] **Choose how to handle changed emoji artwork** — Six mirrored and two replaced emoji affect existing library games; pick one of the three offered options (ship as-is / pin the six / pin all eight). ([discussion](https://github.com/ArnoldSmith86/virtualtabletop/pull/3109#issuecomment-5233869565))

Updates both emoji sets that VTT ships with to their latest upstream releases and makes the new emoji available in the icon pickers.

## What changed

**Assets**

| set | from | to |
| --- | --- | --- |
| color SVGs (`assets/noto-emoji/`) | Emoji 15.1 (`googlefonts/noto-emoji` v2.047-era) | **v2.051** (Emoji 17.0) |
| monochrome font (`assets/fonts/NotoEmoji-VariableFont_wght.ttf`) | 3.003 | **3.005** |

That is 155 new SVGs plus the flag of Sark (from `third_party/region-flags/waved-svg`), upstream redesigns of ~790 existing emoji, and the updated Syrian flag.

`assets/noto-emoji/FLAGS-README.md` was refreshed from upstream. `FLAGS-LICENSE`, `FLAGS-AUTHORS` and `FLAGS-README.third_party` were checked byte-for-byte against upstream and are already current. `assets/noto-emoji/LICENSE` was **not** current: it still carried the Apache-2.0 text, while upstream relicensed its root `LICENSE` to the SIL Open Font License 1.1 in July 2024 (`googlefonts/noto-emoji@254596e`, "Update LICENSE to OFL"), i.e. well before the v2.051 assets imported here. It now carries the OFL text that ships with v2.051.

The About dialog told users the opposite: `client/room.html` listed Noto Emoji among the Google fonts "licensed under the Apache License, Version 2.0". Noto Emoji now has its own paragraph attributing both the color images and the monochrome font (whose own `name` table declares OFL) to the SIL Open Font License 1.1, linking the OFL and the bundled `assets/noto-emoji/LICENSE`, and noting that the region flags among the images are public domain per `FLAGS-LICENSE`.

**Icon pickers** — the 16 new base emoji are added to `assets/fonts/symbols.json`, in their Unicode `emoji-test.txt` order and with keywords from [emojilib](https://github.com/muan/emojilib), so the pickers that read that file offer them:

🫩 face with bags under eyes · 🫪 distorted face · 🫯 fight cloud · 🫈 hairy creature · 🧑‍🩰 ballet dancer · 🫆 fingerprint · 🫍 orca · 🪾 leafless tree · 🫜 root vegetable · 🛘 landslide · 🪊 trombone · 🪉 harp · 🪎 treasure chest · 🪏 shovel · 🫟 splatter · 🇨🇶 flag of Sark

Each of them shows up in both flavours — the color image and the `emoji-monochrome` font glyph. The 15 non-flag entries appear in the symbol picker overlay *and* in the widget editor's icon search; 🇨🇶 appears in the symbol picker only, because the icon search deliberately skips the whole `Emoji - Flags` category (`client/js/editor/propertyInputs.js`), as it already does for the 250+ existing flags.

**Rich text** — the emoji regex in `client/js/symbols.js` (which replaces emoji in rich text with ``) only covered code point ranges up to Emoji 15.1, so the new characters were left as raw text. Its ranges are widened accordingly; no previously matched sequence loses its image.

The widened ranges are written out code point by code point rather than collapsed, so that the regex matches **only** characters that have a shipped SVG. Verified exhaustively over U+1F000–U+1FBFF: 0 matched code points without an asset (down from 122 before this PR), and 0 asset-backed code points lost. That also closes a pre-existing hole — `\udd47-\ude7c` swallowed the chess symbols block U+1FA00–U+1FA6F, which has no emoji SVGs, so a chess symbol such as 🨀 in rich text turned into a broken image; it is now left as text.

## Skin tone modifiers and friends

Unchanged on purpose, as requested: the images for all skin-tone (and the new Emoji 17.0 skin-toned "people with bunny ears" / "people wrestling") sequences are shipped so a game can reference them by file name, but they are **not** listed in `symbols.json` and therefore do not appear in the pickers. Showing them is left to a follow-up PR.

## Verification

- `npm test` — 1167 jest tests pass.
- TestCafe `editor.js` symbol-picker tests pass locally (`Deck editor: symbol pickers and JSON fallback`, `The symbol picker says an image-only search found nothing`).
- Drove the picker in Chromium: all 16 new emoji are present with the expected color/monochrome variants, none render as tofu, no 404s for the new SVGs, and the picker still lists zero skin-tone entries. Screenshot of the 16 new entries (monochrome glyphs first, then the color images): https://agent.virtualtabletop.io/reports/pr/1536103981317619814/new-emoji-picker.png
- Ran the *shipped* regex from `/js/symbols.js` in Chromium against sample text: the 16 new emoji, ❤️, 🇩🇪, 🏴󠁧󠁢󠁳󠁣󠁴󠁿 and 1️⃣ all resolve to SVGs that return 200; chess symbols and the ten still-unassigned code points in the neighbouring ranges are left as plain text. No console errors, no 404s.

**Icon pickers, part 2 — previewing color emoji** — both pickers used to preview a color emoji by printing the character and letting the *browser's* emoji font draw it. A font only knows the emoji that existed when it was built, so every one of the 15 new entries showed up as an empty box in the symbol picker and as a blank chip in the widget editor's icon search, even though picking one placed the correct artwork (auto-review 2/4, UI/UX). Both surfaces now draw the SVG that VTT ships — the same file the widget itself gets — so the preview matches the result and keeps working for every future emoji release:

- `.emoji-color` tiles get an explicit `--symbolSize` box and a `background: … center/contain`; a tile is handed its `--url` by an `IntersectionObserver` as it is scrolled to, because a browser fetches every rendered element's background-image up front and the color emoji are 1,914 files / 20 MB. Opening the picker now costs **0** emoji-SVG requests, down from the 270 the flags always triggered.
- tiles are all one size again (64×64 in the big-preview view; a character with no glyph used to collapse to 28×75px next to its 80×75px neighbours).
- `renderIconChip` renders the raw-emoji branch as ``, with the character as `alt` as a fallback.
- emoji tooltips read `orca (color emoji)` instead of `emoji-color: 🫍`, the duplicated emoji headings now say `(monochrome)` / `(color)`, and the search hint no longer claims emoji use the browser font.
- the picker's search flattens `-` `_` `/` to spaces on both sides (as `client/js/audio.js` already does), so `medical mask` finds `face_with_medical_mask` and `root_vegetable` finds `root vegetable`. The new keywords follow the file's snake_case convention and gained the synonyms their neighbours have; `fingerprint`, `splatter`, `leafless tree` and `ballet dancer` had a single keyword each.
- dropping the now-styleless `emojiFlag` class also fixes flags coming back from the picker as type `"emoji-color emojiFlag"`, which made `isImage` false and `url` null — picking a flag in an image-only picker (deck editor "Add image", `je_imagePicker`) silently did nothing.

**Font license attribution, part 2** — while checking which of the shipped binaries really are Apache-2.0, `nameID 13`/`14` of every `.ttf` in `assets/fonts/` turned out to disagree with the About dialog for five more families: Neuton ("Serif"), Bokor ("Fantasy"), Metamorphous ("Gothic"), Creepster ("Horror") and Gugi ("Tech") are each licensed by their own author under the SIL Open Font License 1.1, not Apache-2.0. At the maintainer's request they moved into an "Open Font License fonts" paragraph carrying the per-font copyright notices the OFL asks to be retained; Roboto (17 faces, Apache-2.0 in its name table) and Material Symbols keep the Apache-2.0 sentence.

No widget property, routine function or tutorial is touched by this change, so there is no new UI surface to build and no test room to add.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3109/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

Clean, well-scoped asset refresh — I verified independently that every emoji listed in `symbols.json` resolves to an existing SVG (1914/1914), that the new monochrome font renders every non-skipped picker entry (including the `🧑‍🩰` ZWJ ligature), and that the widened regex matches all 16 new emoji while matching *nothing* the old regex matched but the new one doesn't. Only real remark: the widened `\ud83e` ranges now also match ten still-unassigned code points that have no SVG, which is a small robustness regression versus the previously tight ranges.

## Findings

### 1. The widened ranges match unassigned code points with no shipped SVG — `client/js/symbols.js:8`

The old regex used tight ranges (`\ude80-\ude88`, `\ude90-\udebd\udebf-\udec5\udece-\udedb\udee0-\udee8\udef0-\udef8`) that deliberately skipped unassigned code points. The new `\ude80-\udec8\udecd-\udef8` (and `\uded5-\uded8`) collapses those gaps. Diffing the two regexes over U+1F000–U+1FBFF, 24 code points are newly matched, of which **10 have no SVG in `assets/noto-emoji/`**:

`U+1FA8B U+1FA8C U+1FA8D U+1FAC7 U+1FADD U+1FADE U+1FAEB U+1FAEC U+1FAED U+1FAEE`

Failure mode: rich text containing one of these characters is rewritten to ``, which 404s and renders as a broken image with alt text instead of just leaving the character as-is. Today these are all unassigned in Unicode 17, so it is unreachable in practice — but that changes the moment Unicode 18 assigns them, and then it fails silently *before* the corresponding assets are shipped, which is exactly the situation the tight ranges were protecting against. Suggest keeping the enumerated style, e.g. `\ude80-\ude8a\ude8e\ude8f\ude90-\udec6\udec8\udecd-\udedc\udedf\udee0-\udee9\udeef\udef0-\udef8` and `\uded5-\uded7\uded8` folded appropriately.

(Low severity — mentioned because the existing ranges show the file's convention is to enumerate assigned code points only.)

### 2. Minor: PR description vs. diff

The description says "The bundled license/README files were refreshed from upstream as well", but the only non-asset text file changed is `assets/noto-emoji/FLAGS-README.md` (a 2023 upstream note). `assets/noto-emoji/LICENSE`, `FLAGS-LICENSE` and `FLAGS-AUTHORS` are untouched — worth confirming they are genuinely still current upstream rather than just not re-copied.

### 3. FYI, not a defect: the new flag is not reachable from the widget-editor icon search

`client/js/editor/propertyInputs.js:338` skips the whole `Emoji - Flags` category when building the icon search index, so `🇨🇶` (like all 250+ existing flags) is offered only in the symbol-picker overlay, not in the widget editor's icon search. Pre-existing behaviour and consistent, but the PR text claims "both pickers" offer all 16 new entries; the flag is the exception.

## What I verified

- **No deletions/renames**: 156 additions, 793 modifications, 0 deletions — so no game referencing an emoji SVG by filename can break.
- **Asset/JSON consistency**: script over `assets/fonts/symbols.json` applying `emojiToFilename()` — all 1914 emoji entries across every `Emoji - *` category resolve to an existing file under `assets/noto-emoji/`, 0 missing.
- **Monochrome font coverage**: parsed the new `NotoEmoji-VariableFont_wght.ttf` cmap (format 12) + GSUB ligature lookups. All 14 new non-flag base emoji have glyphs, and `🧑‍🩰` (U+1F9D1 ZWJ U+1FA70) has a real ligature, so it will not render as two glyphs. Cross-checking every picker entry that `skipForNotoMonochrome()` does *not* skip: 0 would render as tofu with 3.005. This confirms the PR's browser check.
- **Regex regression**: 0 code points in U+1F000–U+1FBFF matched by the old regex but not the new one; all 16 new emoji (including the ZWJ sequence and the RI-pair flag) match completely.
- **`symbols.json`**: valid JSON, no duplicate keys per category, 0 skin-tone entries (as intended); the reindent/`’`→`’` churn is semantically neutral.
- **Caching**: `/i` is served by `express.static` with no `max-age`, so the ~790 redesigned SVGs revalidate normally — returning players won't be stuck on stale art. `i/fonts/symbols.json` and the font are on the 5-minute `cache5m` route.
- Local `npm test` could not run in this sandbox (`jest-environment-jsdom` missing from the clone's `node_modules`) — that's an environment issue, not a PR issue; CI/author report is green.

No widget property, routine function or tutorial is added, so the UI-support and tutorial rules don't apply here.

I added the `beta` label: the PR touches no engine file (`client/js/widgets/`, `compute.js`, `serverstate.js`) — only assets plus a purely additive regex widening in `client/js/symbols.js` — and it cannot corrupt game state or crash the server, so it is safe and useful for players to test on beta.

ArnoldSmith86:
Generate a report of the images that _changed_. Existing games use these images and might rely on details to convey meaning. Group them in the report by subtle rendering changes up to complete replacements and highlight which ones you think are worst for games.

VTT-AI-Agent:
Thanks — all three points addressed in cd35c51.

### 1. Widened ranges match unassigned code points with no SVG — fixed

Valid, and I took the suggestion a step further. Rather than only re-punching the ten holes, I rebuilt the `\ud83e` class from the actual contents of `assets/noto-emoji/`:

```
\ud83e[\udd0c-\udd3a\udd3c-\udd45\udd47-\uddff\ude70-\ude7c\ude80-\ude8a\ude8e-\udec6\udec8\udecd-\udedc\udedf-\udeea\udeef-\udef8]
```

Exhaustive check over U+1F000–U+1FBFF (parsing the regex literal straight out of `client/js/symbols.js` and stat-ing `emoji_u<cp>.svg`):

- code points matched with **no** shipped SVG: **122 → 0**
- asset-backed code points lost versus the current branch: **0**
- code points newly matched versus the current branch: **0**

The 122 are your ten (U+1FA8B–8D, U+1FAC7, U+1FADD–DE, U+1FAEB–EE) plus 112 that predate this PR: `\udd47-\ude7c` spans the **chess symbols** block U+1FA00–U+1FA6F, which has no emoji SVGs. So a chess symbol like 🨀 in rich text renders as a broken image on `main` today — that is now fixed too. It is strictly outside the original scope of the PR, but it is the same defect on the same line and the fix is provably lossless; happy to split it out if you'd rather keep the diff to Emoji 17.0.

Re-verified end to end in Chromium against the *shipped* `/js/symbols.js`: the 16 new emoji, ❤️, 🇩🇪, 🏴󠁧󠁢󠁳󠁣󠁴󠁿 and 1️⃣ all still become ``s whose URLs return 200; chess symbols and the ten unassigned code points stay plain text. No console errors, no 404s.

### 2. PR description vs. diff — you were right to ask, and one file was genuinely stale

I diffed all five bundled text files against upstream:

| file | upstream source | status |
| --- | --- | --- |
| `FLAGS-LICENSE` | `third_party/region-flags/LICENSE` | identical |
| `FLAGS-AUTHORS` | `third_party/region-flags/AUTHORS` | identical |
| `FLAGS-README.md` | `third_party/region-flags/README.md` | refreshed in this PR |
| `FLAGS-README.third_party` | `third_party/region-flags/README.third_party` | identical |
| `LICENSE` | root `LICENSE` | **stale — Apache-2.0** |

Upstream relicensed its root `LICENSE` to the **SIL Open Font License 1.1** in July 2024 (`googlefonts/noto-emoji@254596e`, "Update LICENSE to OFL"), which is well before the v2.051 assets this PR imports — so the bundled copy was stating the wrong license for the artwork alongside it. Replaced with the v2.051 text. The PR description now spells out exactly which files were refreshed, which were verified as already current, and why `LICENSE` changed.

### 3. Flag not in the widget-editor icon search — description corrected

Confirmed, and agreed it is pre-existing and consistent (`propertyInputs.js` skips the whole `Emoji - Flags` category, same as `symbols.js` does for the font list). The claim in the description was mine and it was sloppy; it now says the 15 non-flag entries appear in both pickers and 🇨🇶 in the symbol picker only, with the reason.

`npm test` still green (1167 passed).

VTT-AI-Agent:
Done — full report here:

### 📊 [**What actually changed in the 791 modified emoji**](https://agent.virtualtabletop.io/reports/pr/1536103981317619814/emoji-changes.html)

Every changed image rendered before → after at 96×96 and compared pixel by pixel, grouped by how much it changed, with the ones the public library actually uses flagged in amber.

> Generate a report of the images that _changed_. Existing games use these images and might rely on details to convey meaning. Group them in the report by subtle rendering changes up to complete replacements and highlight which ones you think are worst for games.

## The short version

791 files modified, 789 of them SVGs. **707 of those 789 are People & Body** — the same drawing re-rendered across 5 skin tones × gender. Collapsed to distinct artwork it is **183 emoji**, and only about 40 change in a way anyone would notice on a board.

**Nothing that games use as a *mechanic* changed.** I checked explicitly, and all of these are byte-identical to `main`: 🎲 game die · 🃏 joker · 🀄 mahjong · 🎴 flower cards · every coloured circle and square (🔴🟠🟡🟢🔵🟣🟤⚪⚫🟥🟧🟨🟩🟦🟪🟫⬛⬜) · every heart · every card suit ♠️♥️♦️♣️ · 🥇🥈🥉🏅🏆⭐ · all 24 clock faces · all keycaps · all arrows · ✅❌❗❓⚠️🚫🔒 · 💎🪙💰. Anything that colour-codes or counts with those is untouched.

| group | emoji | files |
| --- | --- | --- |
| Direction reversed (mirrored) | 6 | 11 |
| Different object depicted | 13 | 13 |
| Same object, different colour | 19 | 49 |
| Real-world flag updates | 2 | 2 |
| Redrawn in the new house style | 110 | 624 |
| Subtle rendering change | 28 | 73 |
| No visible change at all | 5 | 17 |

## Worst for games, in order

**1. Six emoji were simply mirrored.** 📢 loudspeaker, 📣 megaphone, 🎥 movie camera and 🦴 bone are **pixel-identical to the old image flipped horizontally**; 🪶 feather and 🦶 foot are within 1–5% of a pure flip. Any game that used one as a direction cue — a pointer, a "this way" marker, a left/right pair — now points the other way, and nothing in the file name or the property hints at it. This is the class I'd worry about most, because it is invisible until someone notices the board reads backwards. 🦴 is used by **Hunt & Gather** today.

**2. Colour identity replaced while the shape stayed.** The insidious one: the emoji still *is* the same object, so a game that picked it for its colour silently gets a different colour. 🧥 navy → tan · 🪼 pink → blue · 🪬 grey → blue · 📏 and 📐 yellow → pale blue · 🩹 tan → white · 🪒 teal → wood · 💂 red tunic → dark · 🧙 blue starry robe → grey · 🚌 🚍 white-and-blue → yellow-and-orange. And 🚠/🚡 effectively swapped — the cableway went blue → red while the tramway went red → yellow, so a pair that used to be visually distinct is now both warm-coloured.

**3. The picture shows something else now.** 🛞 was a wooden cart wheel and is now a **modern alloy car wheel** — a medieval or farming game using it for a wagon now shows a car tyre. 🧮 lost its five colour-coded bead rows and is now uniformly red on a dark square frame, which kills any use of it as a colour-counting prop. 🔦 flipped and lost its beam direction, 🪚 flipped and recoloured, 🐲 went from a head in profile to a front-facing mask.

**4. Framing collapsed.** 🫎 moose and 🦝 raccoon went from a large head portrait filling the frame to a small full-body animal. Same subject, but at the 32–64 px a card or token is usually rendered at, the new ones read as a small brown/grey shape instead of an instantly readable face.

The other ~110 are genuine restyles — same subject, same pose, same direction, same colours, just flatter shading and thinner outlines. The whole hand-gesture family is in there, and every gesture keeps its direction (👈 👉 👆 👇 ✌️ 👍 👎 all still point where they did). The 17 Japanese buttons (🈯🈚🈲…) only lost their background gradient, and 5 of them (17 files, including 👺 goblin and the wrestling set) render pixel-identical.

## What this actually breaks in the public library

I grepped `library/` for both `i/noto-emoji/…` paths and raw emoji characters. 16 changed emoji are in use:

| emoji | change | used by | how |
| --- | --- | --- | --- |
| 🍠 sweet potato | different object | Hunt & Gather | widget image |
| 🦴 bone | **mirrored** | Hunt & Gather | widget image |
| 🧥 coat | navy → tan | Hunt & Gather | widget image |
| 🐠 tropical fish | yellow → green, **now faces right** | Creation, Rails Before the War | widget image |
| 🦓 zebra | redrawn | Cheery Chimp, Kyoga | widget image |
| 🚍 oncoming bus | white/blue → yellow | Field Trip | widget image |
| 🚌 bus | white/blue → yellow | Star Star (image), London Detectives (*monochrome font — unaffected*) | mixed |
| 🖐️ splayed hand | restyle | Konigrufen | widget image |
| 📐 triangular ruler | yellow → pale blue | Dreamcatcher, Nexus, Word Whirl | rules text |
| 🧙‍♂️ man mage | blue → grey | Frontiers | rules text |
| 🤝 handshake | restyle | Frontiers, Marriage Routes | image + rules text |
| ✌️ victory hand | restyle | Nexus | rules text |
| ✍️ writing hand | restyle | Word Whirl | rules text |
| 🧮 abacus | colour rows → all red | tutorials *Appearance - Hexagons*, *Properties - Grid* | colour image |
| 👍 thumbs up | restyle | tutorial *121 Fundamental Widget Properties* | widget image |
| 🤞 crossed fingers | restyle | tutorial *Appearance - Holders* | colour image |

**Hunt & Gather** is the one game that takes a real hit — three of its resource images change at once, one of them by flipping. Nothing there breaks mechanically; it is a look-and-feel regression.

## If you want to limit the damage

Everything above is upstream's artwork, so the options are all "keep the old file":

1. **Ship it as is.** Simplest, and it is what every other platform's players already see.
2. **Pin the six mirrored ones** — keep `main`'s `1f4e2`, `1f4e3`, `1f3a5`, `1f9b4`, `1fab6`, `1f9b6`. Cheapest fix for the only class that actively misleads, and it keeps 🦴 stable for Hunt & Gather.
3. **Pin the mirrored six plus 🛞 and 🧮**, the two where the depicted object really changed.

Say the word and I'll push whichever of those you want; my own preference is (2), since a reversed direction is the only change here that can make a board read wrong rather than just look different.

VTT-AI-Agent (commented):
## TL;DR
UI/UX review (not a code review) of the interface this PR produces. The assets themselves are good — all 16 new emoji ship correctly in both flavours and the monochrome font glyphs render everywhere. The problem is the **previews**: because the picker and the widget editor draw color emoji with the *browser's* font rather than the shipped SVG, every one of the 15 new color-emoji entries shows up as an empty tofu box ▯ (or, in search results, a ~1/3-size image) — the one thing a user has to look at in order to choose them. One CSS/markup fix to the color-emoji tiles makes all 15 look like the rest.

**What I exercised.** Ran the app locally (Playwright/Chromium) on this branch. Symbol picker overlay at 390x844, 820x1180, 1280x800 and 1920x1080 (browse view, search view, "few results" big-preview view, and the images-only/fonts-only variants); the widget editor's icon property picker + icon search (1280x800); the edit-mode toolbar/sidebar; and the Game Shelf, Players and About overlays at all four viewports as a regression sweep. Not reached this run: the deck editor's own symbol-picker host, the states/save & variant dialogs, share/publish, the JSON editor sub-dialogs and chat — nothing in this diff touches their markup or CSS, and the shared picker CSS (`client/css/fonts.css`) is unchanged, so the risk there is low.

---

## Big layout improvements

### 1. The 15 new color emoji render as tofu boxes in the picker's browse view
This is the headline issue and it is a direct consequence of adding emoji that no shipped OS/browser font has yet. In `client/js/symbols.js:105` each color-emoji tile is `<i class="emoji-color" …>🫍</i>` — the character itself, drawn with the browser font. `client/css/fonts.css:530` only swaps in the real SVG on `:hover` or in `.fewResults`. For every emoji that already existed that's fine; for Emoji 17 characters the browser has no glyph, so the tile is a notdef box:

new color emoji is a tofu box in the browse list

*(Emoji – Animals & Nature, color section: 🫍 orca sits between the dolphins and the penguin as an empty ▯.)*

A user scrolling the picker sees ~15 blank boxes scattered through the color-emoji categories and has no way to know they are real, pickable icons — the hint text ("Emoji use your browser font … hover one … to see the real image") explains the *low-resolution* case, not the *no-glyph-at-all* case.

**Suggested change:** make the color-emoji tiles always use the shipped SVG rather than the font, e.g.

```css
#symbolPickerOverlay .emoji-color {
  display: inline-block;
  width: var(--symbolSize);
  height: var(--symbolSize);
  background: var(--url) center/contain no-repeat;
  color: transparent;
}
```
and drop the `:hover` / `.fewResults` special-casing for `.emoji-color` (keep it for `.gameicons`, whose spritesheet reason still applies). VTT already ships every one of those SVGs, so this costs nothing but the requests, and it also fixes #2 below. If always-SVG is considered too many requests for the full list, the minimum fix is to give `.emoji-color` an explicit box + `background-size` so the *hover* preview is at least the right size.

### 2. In "few results", the new emoji's real image is ~1/3 the size of every other one
Same root cause: `.emoji-color`'s box is sized by the *font's advance width* for the character, and the background has no `background-size`. With a glyph present the tile measures 80x75px; with no glyph it collapses to 28x75px, so the SVG is drawn tiny and sitting low next to its neighbours. Measured in-browser at 1280x800, search "whale":

```
emoji-color 🐳  80.0 x 75.0
emoji-color 🐋  80.0 x 75.0
emoji-color 🫍  28.0 x 75.0   <- new
```

orca color preview is a third the size of the other whales

Reproduces identically for 🪏 shovel, 🪉 harp, 🫜 root vegetable, 🫟 splatter, 🛘 landslide (all 28px) at every viewport I tried — phone through wide desktop. The fix in #1 removes it.

### 3. The widget editor's icon search shows the new emoji as blank chips
`renderIconChip` (`client/js/editor/propertyInputs.js:305`) renders a raw emoji value as `div(chip, 'emojiColorChip', …)` — again native font rendering, at `font-size: 36px` (`client/css/editor/propertyInputs.css:488`). So in the icon picker's search results the new color emoji are empty bordered squares:

icon search "whale" — the orca color chip is blank
icon search "harp" — same, bottom row

The blank chip is genuinely misleading rather than merely ugly: clicking it *does* work — `getIconDetails` (`client/js/symbols.js:284`) falls back to `/i/noto-emoji/emoji_u….svg`, so the widget on the table shows the real artwork. The preview contradicts the result. Since `imageURLFromSymbol()` already exists two functions away in the same file, rendering the `emoji-color` branch as `` (like the game-icons branch right above it) would make the chip match what gets placed, and would keep working for every future emoji release.

For reference, all 16 assets are present and both flavours are correct once actually drawn from the shipped files (color SVG top row, `emoji-monochrome` font glyph below):

the new emoji, color SVG vs monochrome font

---

- **`title` tooltips are unreadable for exactly these emoji.** `client/js/symbols.js:91` and `:105` set `title="emoji-color: 🫍"` / `"emoji-monochrome: 🫍"` — the tooltip repeats the character, which for a new emoji is another tofu box in the OS tooltip font, and the class name is developer jargon. Since `symbols.json` now carries a human name as the first keyword for these, `title="orca (color emoji)"` would cost nothing and make the whole picker self-describing.
- **Search-hint text is now slightly wrong.** `client/room.html:539` — "Emoji use your browser font and game-icons are previewed in low resolution: hover one … to see the real image." After this PR some emoji have *no* browser glyph at all, so "previewed in low resolution" undersells it. If #1 is fixed, the sentence can simply drop the emoji clause; if not, it should say emoji may show as an empty box until hovered.
- **Nothing else regressed.** The picker's container-query layout still behaves at all four viewports (phone: [browse](https://agent.virtualtabletop.io/reports/ui-review/1536124061766393867/picker-all-phone.png) / [search](https://agent.virtualtabletop.io/reports/ui-review/1536124061766393867/picker-fingerprint-phone.png), [tablet](https://agent.virtualtabletop.io/reports/ui-review/1536124061766393867/picker-orca-tablet.png), [laptop](https://agent.virtualtabletop.io/reports/ui-review/1536124061766393867/picker-all-laptop.png), [desktop](https://agent.virtualtabletop.io/reports/ui-review/1536124061766393867/picker-all-desktop.png)) — hint hidden in the narrow/short boxes, close button reachable, list the only scroller, headings hidden while searching. The 155 extra SVGs and the +93 lines in `symbols.json` add no measurable weight to opening the picker, and no 404s or console errors were logged in any of the runs. Game Shelf / Players / About are untouched at all four sizes.

---

- **Keyword style is inconsistent with the file's own convention, which changes how search behaves.** Existing emoji entries lead with a snake_case name (`face_with_medical_mask`, `person_getting_massage`) while the new ones lead with a spaced phrase (`face with bags under eyes`, `root vegetable`, `treasure chest`). The picker's search is a plain substring match on the joined keyword string (`client/js/symbols.js:114`), so a newcomer typing `medical mask` finds nothing but `root vegetable` works — the two halves of the same list now answer to different typing habits. Worth adding the snake_case form as a second keyword (or, better long term, normalising `_`/space in the search), so both spellings hit.
- **Several new entries have far fewer synonyms than their neighbours**, which makes them effectively unfindable unless you already know the Unicode name: `🫆` → only "fingerprint"; `🫟` → only "splatter"; `🪾` → only "leafless tree"; `🧑‍🩰` → only "ballet dancer". Compare the surrounding entries, which average 3–5. Cheap additions in the spirit of the existing keywords: fingerprint → `thumbprint, id, biometric, security`; splatter → `splash, stain, paint, blob`; leafless tree → `dead tree, bare tree, winter, branch`; ballet dancer → `ballet, dance, dancer, ballerina`. Same for `🪏 shovel` (`spade, dig, garden`) and `🛘 landslide` (`rockslide, mountain, avalanche`).
- **"fight cloud" and "hairy creature" are Unicode names, not player words.** A first-time user building a game will search `brawl`, `fight`, `punch` (covered) but also `bigfoot`/`yeti` — `🫈` has `sasquatch, bigfoot, treeman` (good); `🫯` would benefit from `dust cloud, scuffle, fighting`.
- **`🇨🇶` gets `cq` as a keyword** while no other flag has its two-letter code. Either harmless or the start of a nice feature — but as a one-off it means one flag answers to a code and 250 others don't. I'd drop it or file it as "add ISO codes to all flags".
- **The color/mono distinction is still unlabelled.** Every emoji now appears twice in one flat list — once in the `Emoji – …` font section and once in the `Emoji – …` image section, with identical headings. That predates this PR, but adding 16 more duplicated entries makes it a little more visible; the headings could read e.g. "Emoji – Animals & Nature (monochrome)" / "(color)" so a newcomer understands why they are seeing the same duck twice.

VTT-AI-Agent:
Thanks — this one was worth catching. All three layout findings and every nitpick are fixed in 9520138e; details and the one place I deviated from the suggested fix below.

## 1 + 2. Tofu tiles / one-third-size previews — fixed, but lazily

You're right about the cause and the fix: the tiles printed the character and let the browser's emoji font draw it, so anything newer than that font was a notdef box, and the box was sized by the font's advance width. `.emoji-color` now always draws the shipped SVG in an explicit `var(--symbolSize)` box with `background: … center/contain`, and the `:hover` / `.fewResults` special-casing for it is gone (kept for `.gameicons`, whose spritesheet reason still applies).

The one deviation: I did **not** make the background unconditional. Browsers fetch the background-image of every *rendered* element up front, not just the visible ones — I measured that today's picker already pulls all **270** flag SVGs on open, because `.emojiFlag` was unconditional. Doing that for the whole color list is 1,914 files / **20.2 MB** on every open. So the tile gets `.showImage` (and with it `--url`) from an `IntersectionObserver` on `#symbolList` with a 300px margin. Measured in Chromium on this branch:

| | before | after |
| --- | --- | --- |
| noto-emoji SVG requests when the picker opens | 270 | **0** |
| requests after scrolling the *entire* list to the bottom | 270 | 1914 |

Sizes in the "few results" view, same measurement you ran (search `whale`, 1280×800):

```
before                     after
emoji-color 🐳  80 x 75    emoji-color 🐳  64 x 64
emoji-color 🐋  80 x 75    emoji-color 🐋  64 x 64
emoji-color 🫍  28 x 75    emoji-color 🫍  64 x 64   <- new
```

64×64 is the `--symbolSize` the game-icons tiles already use, so emoji and game-icons line up now instead of emoji being slightly taller.

The character stays in the element (transparent) on purpose — the rich-text insert path reads `icon.innerText`; verified it still returns the emoji.

browse view, color emoji sections

## 3. Blank chips in the widget editor's icon search — fixed

`renderIconChip`'s raw-emoji branch now renders ``, exactly as you suggested, with the character as `alt` so a value with no shipped SVG degrades to text instead of a broken-image glyph. The three `.emojiColorChip` CSS rules are dead now and were removed.

icon search "whale" — the orca chip renders

## Nitpicks

- **Tooltips**: emoji tiles now read `orca (color emoji)` / `orca (monochrome emoji)`, built from `symbols.json`'s first keyword. Left `material-symbols: home` and `game-icons.net: …` alone — there the symbol *is* the readable name.
- **Hint text**: the emoji clause is gone; it now reads "Click an icon to pick it. The game-icons are previewed in low resolution: hover one, or narrow the search below 100 results, to see the real image."
- **Duplicate headings**: now "Emoji - Animals & Nature (monochrome)" and "… (color)".

## Readability

- **snake_case vs spaced keywords**: fixed on both sides rather than by adding second keywords. The picker's search now flattens `-`, `_` and `/` to spaces in the keyword string *and* in the query — the same trick `client/js/audio.js` already uses — so `medical mask` finds `face_with_medical_mask` (0 matches before, 4 now) and `root_vegetable` finds `root vegetable`. The new entries were also renamed to the file's snake_case convention (`face_with_bags_under_eyes`, `distorted_face`, `fight_cloud`, `hairy_creature`, `ballet_dancer`, `leafless_tree`, `root_vegetable`, `treasure_chest`). `10k` still matches exactly one icon, so the TestCafe expectations are unchanged.
- **Sparse keywords**: took your suggestions and added a few more — fingerprint → `thumbprint, id, biometric, security, identity`; splatter → `splash, stain, paint, blob, spill`; leafless tree → `dead_tree, bare_tree, winter, branch, nature`; ballet dancer → `ballet, dance, dancer, ballerina`; shovel → `spade, dig, garden`; landslide → `rockslide, mountain, avalanche`; fight cloud → `fight, dust_cloud, scuffle, punch`; hairy creature → `yeti, monster`; plus harp/trombone/orca/root vegetable/distorted face/face with bags under eyes. Every new entry now has 5–7 keywords, in line with its neighbours.
- **`cq` on the Sark flag**: dropped, and the entry lines up with the other 250 flags (`flag_sark, flag, nation, country, banner, sark`). Adding ISO codes to *all* flags is a reasonable separate idea, not a one-off.

## One thing outside your list

Removing the `emojiFlag` class (it had no style left) also removed a bug: the class was appended to `data-type`, so a flag came back from `pickSymbol` as type `"emoji-color emojiFlag"`, which made `isImage` false and `url` null. Picking a flag in an image-only picker — the deck editor's "Add image", `je_imagePicker` — silently did nothing. It works now.

## Verification

`npm test` (1167 jest tests) green. TestCafe `Deck editor: symbol pickers and JSON fallback` and `The symbol picker says an image-only search found nothing` both pass locally. Drove the picker and the icon search in Chromium at 1280×800: no console errors, no 404s, no tofu anywhere in the color sections, hover/`fewResults`/images-only/fonts-only variants all still behave.

Not re-swept: the phone/tablet/wide viewports and the Game Shelf / Players / About overlays — the only layout-affecting change since your sweep is the `.emoji-color` tile becoming a fixed `--symbolSize` box (it is now the same box the game-icons tiles have always had), so I don't expect the container-query behaviour to differ; happy to re-run it if you'd rather have it confirmed.

search "root vegetable" — both flavours at full size

VTT-AI-Agent (commented):
## TL;DR

The picker and rich-text changes at the latest head look functionally sound, and all 1,914 emoji exposed by `symbols.json` resolve to shipped SVGs. One correctness issue remains: after updating the bundled Noto Emoji license to OFL-1.1, the About dialog still tells users that Noto Emoji is Apache-2.0 licensed.

## Findings

### 1. Update the user-facing Noto Emoji license attribution — `client/room.html:598`

This paragraph still groups Noto Emoji with the fonts described as Apache-2.0 and links only to the Apache license. In the same PR, `assets/noto-emoji/LICENSE` is deliberately replaced with the SIL Open Font License 1.1 text, and the PR description explains that the imported v2.051 assets are OFL-licensed. That leaves the application presenting contradictory licensing information for the assets it ships. Please split Noto Emoji out into an OFL-1.1 sentence (and link to the OFL or the bundled license), while retaining the appropriate attribution for the other Google fonts/material symbols.

## Checks

- Parsed `assets/fonts/symbols.json` and reproduced the filename mapping: all 1,914 emoji entries have a corresponding SVG, with no deleted Noto Emoji assets.
- Scanned the SVG set for executable markup (`script`, `foreignObject`, `javascript:` and `onload`); none was found.
- No widget property, routine operation, or tutorial is added, so the creator-UI and tutorial requirements do not apply.
- GitHub CI and CodeQL are green. The local Jest command could not start because this clone has no `node_modules/.bin/jest`, so I relied on the green CI run plus the dependency-free checks above.

The `beta` label was already present; this PR touches no engine files and the remaining attribution issue does not make player testing unsafe.

VTT-AI-Agent:
Good catch — fixed in 26b5c3ee.

## The Noto Emoji attribution — fixed

`client/room.html:598` lumped Noto Emoji in with the Apache-2.0 Google fonts. Both halves of what VTT ships are in fact OFL-1.1:

- the color images — `assets/noto-emoji/LICENSE`, the OFL text that ships with upstream v2.051 (this PR);
- the monochrome font — `assets/fonts/NotoEmoji-VariableFont_wght.ttf` itself declares it. Read straight out of its `name` table:
  ```
  nameID 0  Copyright 2013 Google LLC
  nameID 13 This Font Software is licensed under the SIL Open Font License, Version 1.1. …
  nameID 14 http://scripts.sil.org/OFL
  ```

So Noto Emoji is now its own paragraph. Noto Emoji was removed from the "Google fonts" sentence, which keeps its Apache-2.0 wording for the fonts that really are Apache-2.0:

> **Noto Emoji:** The monochrome Noto Emoji font and the color Noto Emoji images are also provided by Google and are licensed under the SIL Open Font License, Version 1.1. The license and FAQ are available at [https://scripts.sil.org/OFL](https://scripts.sil.org/OFL), and a copy ships with the images in [assets/noto-emoji/LICENSE](https://github.com/ArnoldSmith86/virtualtabletop/blob/main/assets/noto-emoji/LICENSE). The country, territory and subdivision flags among those images come from the region-flags collection bundled with Noto Emoji; they were taken from Wikipedia and are in the public domain or otherwise exempt from copyright, as detailed in [assets/noto-emoji/FLAGS-LICENSE](https://github.com/ArnoldSmith86/virtualtabletop/blob/main/assets/noto-emoji/FLAGS-LICENSE).

The flag sentence is there because the flags are the one part of `assets/noto-emoji/` that is *not* OFL — `FLAGS-README.third_party` records them as public domain — and the dialog never mentioned them at all.

Rendered (About → Fonts and Other Symbols): https://agent.virtualtabletop.io/reports/pr/1536103981317619814/license-fix/about-noto-emoji-ofl.png

## One thing your finding uncovered that I did *not* change

While verifying which of the shipped binaries are actually Apache-2.0, I read `nameID 13` out of every `.ttf` in `assets/fonts/`. Five more of the six families named in that same sentence are OFL-1.1, not Apache-2.0:

| font | embedded license |
| --- | --- |
| Roboto (all 17 faces) | Apache-2.0 ✅ matches the dialog |
| Neuton ("Serif"), Bokor ("Fantasy"), Metamorphous ("Gothic"), Creepster ("Horror"), Gugi ("Tech") | **OFL-1.1** ❌ dialog says Apache-2.0 |

That is a pre-existing inaccuracy — those fonts are untouched by this PR, and correcting the licence statement for five unrelated assets felt like the wrong thing to slip into an emoji-update PR without a maintainer saying so. Happy to add it here in a one-line follow-up commit, or to leave it for a separate PR — say which and I'll do it. (Material Symbols is a `.woff2`, so I could not read its name table the same way; Google publishes it under Apache-2.0, which is what the dialog says.)

## Verification

`npm test` — 1167 jest tests pass. Opened the About overlay in Chromium at 1280×900: the new paragraph renders in the right place, all three links resolve, no console errors.

## On your other checks

Agreed on all of them, and thanks for re-verifying the 1,914-entry `symbols.json` → SVG mapping independently. Re the local jest failure you hit: `node_modules` was missing from your clone — `npm ci` at the repo root sets it up (the suite takes ~80s).

96LawDawg:
Fix the license attribution for Neuton ("Serif"), Bokor ("Fantasy"), Metamorphous ("Gothic"), Creepster ("Horror"), Gugi ("Tech")

VTT-AI-Agent:
Done — e606588e.

The five families that were listed as Apache-2.0 but declare OFL-1.1 in their own `name` table now have their own paragraph in the About dialog (About → Fonts and Other Symbols), with the per-font copyright notices the OFL asks to be retained. Roboto and Material Symbols keep the Apache-2.0 sentence, since they really are Apache-2.0:

> **Google fonts:** The Roboto font family and Google Material Symbols are provided by Google as part of the Google Fonts and icon libraries and are licensed under the Apache License, Version 2.0. A copy of this license is available at https://www.apache.org/licenses/LICENSE-2.0.html.
>
> **Open Font License fonts:** The Neuton ("Serif" in VTT), Bokor ("Fantasy" in VTT), Metamorphous ("Gothic" in VTT), Creepster ("Horror" in VTT) and Gugi ("Tech" in VTT) font families are distributed through Google Fonts as well, but each of them is licensed by its own author under the SIL Open Font License, Version 1.1: Neuton copyright 2010 The Neuton Project Authors, Bokor copyright 2020 The Bokor Project Authors, Metamorphous copyright 2011 Sorkin Type Co, Creepster copyright 2011 Font Diner, Inc, and Gugi copyright 2017 TAE System & Typefaces Co. The license and FAQ are available at https://scripts.sil.org/OFL.

The copyright holders are read straight out of `nameID 0` of each shipped `.ttf`, and the license claim out of `nameID 13`/`14`:

| file(s) in `assets/fonts/` | nameID 0 | nameID 13 |
| --- | --- | --- |
| `Neuton-{Regular,Italic,Bold}.ttf` | Copyright 2010 The Neuton Project Authors | SIL Open Font License, Version 1.1 |
| `Bokor-Regular.ttf` | Copyright 2020 The Bokor Project Authors | SIL Open Font License, Version 1.1 |
| `Metamorphous-Regular.ttf` | Copyright (c) 2011, Sorkin Type Co | SIL Open Font License, Version 1.1 |
| `Creepster-Regular.ttf` | Copyright (c) 2011, Font Diner, Inc | SIL Open Font License, Version 1.1 |
| `Gugi-Regular.ttf` | Copyright (c) 2017 by TAE System & Typefaces Co. | SIL Open Font License, Version 1.1 |
| `Roboto*.ttf` (17 faces) | Copyright 2011 Google Inc. | Apache License, Version 2.0 |

Material Symbols ships as `.woff2` so I can't read its name table the same way; Google publishes it under Apache-2.0, which is what the dialog says, so it stays in the Apache paragraph.

**Verification:** `npm test` — 1167 jest tests pass. Opened the About overlay in Chromium at 1280×950: both paragraphs render in the right place in the "Fonts and Other Symbols" section, all links resolve, no console errors. Screenshot: https://agent.virtualtabletop.io/reports/pr/1536103981317619814/license-fix/about-ofl-fonts.png

VTT-AI-Agent (commented):
## TL;DR

I could not find a correctness bug in the code part of this PR — the risky bits (the widened emoji regex, the picker's lazy SVG loading, the license claims) all check out when verified mechanically. Only two cosmetic nits below, both in new code. The `beta` label is already on the PR and stays.

## What I verified rather than took on trust

- **The regex in `client/js/symbols.js:8`.** I extracted the old and the new regex from `main` and the PR head and ran both over every code point U+1F000–U+1FBFF, cross-referencing `assets/noto-emoji/`: **0** code points lost an image they used to get, **0** code points now match without a shipped SVG, 14 newly match and all 14 have an asset. The claim that `\udd47-\ude7c` used to swallow the chess block also holds — U+1FA00–1FA6F no longer matches, and none of those has an SVG.
- **Every emoji key in `assets/fonts/symbols.json`** (all categories, not just the new ones) resolves through `emojiToFilename()` to a file that exists — important now that the picker draws the SVG instead of the font, because a missing file would silently produce a blank tile rather than a font-drawn fallback. 0 missing, 0 empty keyword arrays, 0 keywords containing `"`/`<`/`&` (the `title=`/`data-keywords=` interpolation is still unescaped, so that last one is worth keeping in mind if keywords ever gain punctuation).
- **The monochrome font really covers the new emoji.** All 14 new base code points are in the `cmap` (format 12) of the updated `NotoEmoji-VariableFont_wght.ttf`, and the ZWJ sequence for 🧑‍🩰 has a real GSUB ligature (`1F9D1 + ZWJ + 1FA70`, plus its five skin-tone variants) — so it renders as one glyph and correctly does *not* need an entry in `skipForNotoMonochrome`.
- **The license changes in `client/room.html`.** Read the `name` table of all 27 `.ttf`s under `assets/fonts/`: Bokor, Creepster, Gugi, Metamorphous, the three Neuton faces and NotoEmoji all declare the SIL Open Font License; all 20 Roboto faces declare Apache-2.0. The rewritten paragraphs match that exactly.
- **`IntersectionObserver` root.** `$('#symbolList')` is genuinely the scroller (`overflow-y: auto`, `client/css/fonts.css:473`), and `--symbolSize` is defined on it (24px, 64px under `.bigPreviews.fewResults`), so the new `.emoji-color` box sizing resolves. Tiles hidden by search or by `hideImages`/`hideFonts` are `display:none` and therefore never intersect, so the "0 requests on open" claim survives the family filters too.
- **The dropped `emojiFlag` class is a real fix,** not just cleanup: `pickSymbol` compares `icon.dataset.type` against `'emoji-color'` (`client/js/symbols.js:181`), and `"emoji-color emojiFlag"` never matched, so flags came back with `isImage:false, url:null`.

## Findings

### 1. `renderIconChip` now turns any non-emoji fallthrough into a broken image — `client/js/editor/propertyInputs.js:308`

The `else` branch is not reached only by emoji. Anything that isn't `/…/`, `[…]`, `(…)`, a `_NOFILL` name or `^[a-z0-9]` lands there — e.g. an `icon` of `"Sword"` (capital S), `"→"` (U+2192), or any stray character a game author typed. Those used to render as text in the chip; now they become ``, i.e. a 404 and a broken-image box. The `alt` helps (Chrome does draw alt text), but it's a visible downgrade for exactly the values that were already dubious.

A one-liner keeps the old behaviour for those without giving up the fix:

```js
chip.innerHTML = `<img src="${html(imageURLFromSymbol(icon))}" alt="${html(icon)}" onerror="this.replaceWith(this.alt)">`;
```

(or an `img.onerror` handler assigned after `div()`, to stay out of inline handlers).

### 2. `emojiTitle` prints `keywords[0]` verbatim — `client/js/symbols.js:64`

Two entries have a first keyword that isn't a display name: `#️⃣` and `*️⃣` both start with `"keycap_"`, so both tiles get the tooltip `keycap  (color emoji)` — a double space, no indication which keycap it is. Everything else in the file reads fine. Either `.trim()` after the underscore replacement, or fall back to the next keyword when the first ends in `_`; alternatively fix the two keyword lists in `symbols.json` to `keycap_hash` / `keycap_asterisk`, which also makes them searchable by name.

## Non-findings I checked and dismissed

- The new `searchableKeywords()` flattening is applied to `data-keywords` **and** to the query, so no existing search phrase stops matching (hyphenated game-icons names, `[vtt_symbol]` names and material symbol names all still match when typed with their original separator). The widget editor's icon search (`searchIconIndex`) doesn't flatten, but it splits the query on whitespace and substring-matches each term, so `root vegetable` / `medical mask` already worked there — the two surfaces agree in practice.
- The TestCafe expectation `#symbolList .material-symbols` count `eql(1)` for `10k` is unaffected by the flattening (no separators in that query or its keywords).
- Removing `top: -3px` is a no-op: `#symbolPickerOverlay i` is statically positioned.
- No widget property, routine function or new game-author-facing surface is added, so the UI-support and tutorial rules don't apply here.

I could not run the suites locally (this sandbox's `node_modules` is missing `jest-environment-jsdom`), so I'm relying on your `npm test` / TestCafe runs for those.

I left the `beta` label in place: it was already there and remains right — the diff touches no engine file (nothing under `client/js/widgets/`, no `compute.js`, no `serverstate.js`), and the only behavior change for existing games is the emoji regex, which I verified strictly widens image coverage onto code points that all have shipped assets.

ArnoldSmith86:
Closing this for now. #3131 added only the new icons.

PR #3110, opened 2026-08-09
Rebuild the public-library game Don't Blow Up with new art and automation

VTT-AI-Agent:
Rebuilds the public-library game **Don't Blow Up** (similar to *Exploding Kittens*) so that it looks and plays like a finished game instead of a functional prototype. Same rules, same 56-card deck, same component counts — new art, new table layout and a lot more automation.

Before / after (same room, same game):

| before | after |
| --- | --- |
| before | after |

## What the game is

Take turns drawing cards. Draw one of the exploding cats and you are out of the game — unless you can play a Laser Pointer, which lets you secretly slide the cat back into the deck anywhere you like. Skip, Attack, Peek, Shuffle, Favor and Nope let you push the danger onto everybody else; matching cat cards let you steal from other players. There is always one exploding cat fewer than there are players, so the game always ends with exactly one survivor.

Similar to **Exploding Kittens** (Matthew Inman, Elan Lee, Shane Small, 2015) — [BGG](https://boardgamegeek.com/boardgame/172225/exploding-kittens). 2–5 players. The deck composition is the faithful 56 cards: 4 exploding, 6 defuse, 5 nope, 5 peek, 4 attack, 4 skip, 4 favor, 4 shuffle and 5 × 4 cat cards.

## What changed

**Look.** Dark themed table with a subtle paw texture; a Creepster title; three labelled drop zones (Draw, Discard, Danger Zone); a combo legend on the left and a live table-status panel on the right.

The 56 cards were redrawn from scratch in CSS: a per-card gradient, a tinted icon, a name plate, the rules line, a flavour line, an `×N` copy counter, and a **corner index chip** with the card's own icon so that every card in a fanned hand stays identifiable at a glance. Powers are dark cards, the five cat cards are light cards, so the two groups read apart instantly; every card also has a unique icon and a unique name, so nothing depends on colour alone. Cards enlarge on hover.

Each seat carries a badge with that player's hand size, which turns into a red ✗ when they blow up.

**Automation (the chores — always on).**

* **New Game** collects every card, shuffles, deals 7 cards + 1 Laser Pointer to each seated player, shuffles the remaining Laser Pointers back into the deck, puts exactly *players − 1* exploding cats in, clears everybody's eliminated flag and picks a random starting player. It refuses to deal below two players and asks for confirmation first.
[...]

PR #3112, opened 2026-08-09
Give every preview widget its own css scope so the deck editor stops showing one card type's colors on all of them

VTT-AI-Agent:
Every widget the editor renders as a read-only preview now gets its own CSS scope, so a deck that tints its card types — through an `html` face object or through the `css` property — shows each card type in its own colors instead of all of them in the last card type's colors.

## The bug

Both ways of styling a card put CSS into a `<style>` element that is scoped to the card's widget id, and preview cards are created without one (`new Card()`), so every preview shared the same scope and whichever preview was rendered last restyled all the others:

* the CSS of an `html` face object is scoped to a generated class, `html-object-<card id>-<face>-<object>` — every preview got `html-object-undefined-<face>-<object>`;
* the object form of the widget-level `css` property is written into `<style id="STYLES_<card id>">` with `#w_<card id>` selectors — every preview got `<style id="STYLES_">` and `#w_`, and since all id-less previews also share the DOM id `w_`, that one surviving stylesheet painted all of them.

That hits every place the editor renders preview cards:

* the deck editor's card type strip and its main card view,
* the card type list in the widget properties (Card type / Card Count),
* the deck previews in the "add widget" picker.

It is very visible on decks that tint their faces from a card type property, e.g. `"background": "linear-gradient(..., ${PROPERTY tint1}, ${PROPERTY tint2})"`. Cards in the room itself were never affected — those have real, unique widget ids.

Reproduced with the deck of [#3110](https://github.com/ArnoldSmith86/virtualtabletop/pull/3110) (https://test.virtualtabletop.io/PR-3110/pr-test), whose 13 card types each carry their own `tint1`/`tint2`/`ink`. All 13 strip previews rendered in the yellow of `cat-yowl`, the last card type:

| before | after |
| --- | --- |
| before | after |

## The change

`Widget` now computes a `cssScope` once, in its constructor: the escaped id when the widget has one, a per-instance `unnamed_<n>` when it does not. Everything that identifies the widget in CSS is built from it — its DOM id, the id of its stylesheet element, the `#w_…` selectors in it, and the `html-object-…` class of `Card.createFaces`. Nothing changes for a widget in a room: `escapeID('card')` is `card`, so both the DOM id and the generated class stay exactly what they were, and `escapeID` never emits a lone `_`, so `unnamed_<n>` cannot collide with a real id.

Preview widgets are simply dropped from the DOM instead of going through `applyRemove`, so nobody removed their stylesheet. With one shared `STYLES_` id that did not matter — the next preview overwrote it — but one stylesheet per preview would pile up in `<head>` as the editor re-renders (measured: +4 per click in the deck editor's card type strip). Their stylesheets are now collected once their preview has left the document, which keeps the count at the number of previews actually on screen.

Running the id through `escapeID` also fixes a second, smaller bug: a card whose id contains a character that is not valid in a class name (a space, for example) made `classList.add()` throw, and its html face objects were not rendered at all. They render now.

## Testing

* `npm test` — 1167 passed.
* New TestCafe regression test in `tests/testcafe/editor.js`: a deck with three card types tinted red/green/blue from a `${PROPERTY}` — once through an `html` face object, once through `cardDefaults.css` — must show three different strip previews, and a room card whose id is `my card` must render its html face object at all. Each assertion fails without the corresponding fix and passes with it.
* Full `tests/testcafe/editor.js` suite: 40 passed. `tests/testcafe/pixelsnapshot.js`: 4 passed. `tests/testcafe/domsnapshot.js`: `html-card` and `holder-image` pass, `widget-gallery` fails on `main` in this sandbox as well (an unrelated opacity/`hidden` timing difference), so it is not caused by this change.
* Walked the deck editor of the #3110 deck end to end in a browser, plus the card type list in the widget properties, and checked that the room itself still renders identically.
* Browser check of the `css` half of the bug — a deck of five card types tinted from `cardDefaults.css`, in the deck editor's card type strip:

| before | after |
| --- | --- |
| before | after |

## Follow-up from the UI review: the card type strip

The fix makes the `CARD TYPES` strip the fastest way to find a card type by sight, which brought two things about
that strip to the surface (both pre-existing, both raised in the UI/UX review):

* **It scrolls sideways with nothing saying so.** 20 card types at 1280x800 is a 1875px strip in a 1044px box, and
  the platform's overlay scrollbar reserves no space at all (`offsetHeight - clientHeight` is 0), so the repo's
  global `::-webkit-scrollbar` styling never shows. Whichever edge still has tiles behind it now fades out - sticky
  pseudo-elements with a compensating negative margin, so `scrollWidth` is unchanged and the strip's scroll-restore
  and `scrollIntoView` logic is untouched. `updateStripOverflow()` sets the classes from `renderStrip`, the strip's
  `scroll` handler and the window `resize` handler.
* **The selected tile was only a 2px outline**, which is easy to lose between tiles that are now as saturated as the
  card types they show. It also gets the highlight plate and a bold label.

| before | after |
| --- | --- |
| before | after |

selected

Three small readability fixes came along with it: a tooltip on the per-card-type count (the `-`/`+` buttons had one,
the number did not), the "Click a face object..." hint under the card view raised from `opacity: 0.7` to `0.9`, and
the deck editor's two layout-relative strings ("to the right" / "below" / "the list to the left") replaced by the
names of the panels they point at.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3112/pr-test (or any other room on that server)

### DIFF
--- a/client/css/editor/deckeditor.css
+++ b/client/css/editor/deckeditor.css
@@ -1846,7 +1846,7 @@ body.deckEditorPanning #deckEditorMain * {
   bottom: 0;
   text-align: center;
   font-size: 12px;
-  opacity: 0.7;
+  opacity: 0.9; /* it is the only hint that the big card is interactive, so it has to be readable */
   padding: 4px 10px;
   pointer-events: none;
 }
@@ -1866,6 +1866,34 @@ body.deckEditorPanning #deckEditorMain * {
   flex-shrink: 0;
 }
 
+/* A deck of 13 (let alone 52) card types is far wider than the strip, so it scrolls sideways - and since every
+   tile now carries its own card type's colors, the strip is the fastest way to find a type by sight, which it
+   can only be if it doesn't look like it ends where its box does. The platform draws an overlay scrollbar that
+   is invisible while the strip sits still, so fade out whichever edge still has tiles behind it (the classes
+   come from updateStripOverflow). Sticky flex items with a negative margin: pinned to the visible edge without
+   adding to the scrollable width, and above the preview cards' own z-index (1000000). */
+#deckEditorStrip.deckEditorStripMoreLeft::before,
+#deckEditorStrip.deckEditorStripMoreRight::after {
+  content: '';
+  position: sticky;
+  flex: 0 0 48px;
+  align-self: stretch;
+  z-index: 1000002;
+  pointer-events: none;
+}
+
+#deckEditorStrip.deckEditorStripMoreLeft::before {
+  left: -10px; /* cancels the strip's own padding, so the fade reaches its edge */
+  margin-right: -53px; /* fade width plus the flex gap */
+  background: linear-gradient(to right, var(--backgroundColor), transparent);
+}
+
+#deckEditorStrip.deckEditorStripMoreRight::after {
+  right: -10px;
+  margin-left: -53px;
+  background: linear-gradient(to left, var(--backgroundColor), transparent);
+}
+
 .deckEditorStripCard {
   display: flex;
   flex-direction: column;
@@ -1883,8 +1911,15 @@ body.deckEditorPanning #deckEditorMain * {
   border-color: var(--backgroundHighlightColor1);
 }
 
+/* The tiles are as saturated as the card types they show, so an outline alone is easy to lose among them: the
+   selected tile also gets the plate and the bold label. */
 .deckEditorStripCard.selected {
   border-color: var(--textHighlightColor2);
+  background: var(--backgroundHighlightColor1);
+}
+
+.deckEditorStripCard.selected > span {
+  font-weight: bold;
 }
 
 .deckEditorStripCard.selectedInactive {
--- a/client/js/editor/deckeditor.js
+++ b/client/js/editor/deckeditor.js
@@ -499,10 +499,13 @@ class DeckEditor {
     }, { passive: false });
     $('#deckEditorMain').addEventListener('touchend', e=>{ if(e.touches.length < 2) pinchStartDist = 0; if(!e.touches.length) touchPanning = false; });
 
+    $('#deckEditorStrip').addEventListener('scroll', _=>this.updateStripOverflow());
+
     window.addEventListener('resize', _=>{
       if(this.isOpen()) {
         this.renderMain();
         this.updateDragToolbar();
+        this.updateStripOverflow();
       }
     });
     window.addEventListener('keydown', e=>this.onKeyDown(e));
@@ -1428,7 +1431,7 @@ class DeckEditor {
     if(!this.deck())
       return;
     if(this.deckSymbolSelected)
-      return div(container, 'deckEditorEmpty', '<p>Edit the default properties of every card in this deck to the right, or select a card type below to edit that card.</p>');
+      return div(container, 'deckEditorEmpty', '<p>Edit the default properties of every card in this deck in the properties panel, or select a card type in the "Card types" strip to edit that card.</p>');
     if(this.cardType === null)
       return div(container, 'deckEditorEmpty', '<p>This deck does not have any card types yet. Add one using the button in the bottom strip.</p>');
     if(!this.faceTemplates.length)
@@ -1904,7 +1907,7 @@ class DeckEditor {
     const prevScroll = strip.scrollLeft; // rebuilding the tiles must not snap the strip back to the start
     strip.innerHTML = '';
     if(!this.deck())
-      return;
+      return this.updateStripOverflow(); // an empty strip hides nothing behind either edge
 
     // The card-type toolbar's copy/delete need a selected card type; "± All" needs at least one card type.
     $('#deckEditorStripCopy').disabled = this.cardType === null;
@@ -1963,7 +1966,7 @@ class DeckEditor {
       };
       // Per-card-type "cards in game" count with +/- right under the tile.
       const count = widgetFilter(w=>w.get('deck') == this.deckID && w.get('type') == 'card' && w.get('cardType') == cardType).length;
-      const countRow = div(button, 'deckEditorStripCount', `<button icon=remove title="One fewer card of this type"></button><span class=deckEditorStripCountVal>${count}</span><button icon=add title="One more card of this type"></button>`);
+      const countRow = div(button, 'deckEditorStripCount', `<button icon=remove title="One fewer card of this type"></button><span class=deckEditorStripCountVal title="Cards of this type in the game">${count}</span><button icon=add title="One more card of this type"></button>`);
       countRow.draggable = false;
       countRow.onmousedown = e=>e.stopPropagation();
       countRow.ondragstart = e=>{ e.preventDefault(); e.stopPropagation(); };
@@ -1979,6 +1982,15 @@ class DeckEditor {
     if(selectedTile && selectedTile.scrollIntoView && this.stripScrolledTo !== selectionKey)
       selectedTile.scrollIntoView({ block: 'nearest', inline: 'nearest' });
     this.stripScrolledTo = selectionKey;
+    this.updateStripOverflow();
+  }
+
+  // Which side of the strip still has card types hidden behind its edge, for the fades in deckeditor.css.
+  updateStripOverflow() {
+    const strip = $('#deckEditorStrip');
+    const hidden = strip.scrollWidth - strip.clientWidth;
+    strip.classList.toggle('deckEditorStripMoreLeft',  hidden > 1 && strip.scrollLeft > 1);
+    strip.classList.toggle('deckEditorStripMoreRight', hidden > 1 && strip.scrollLeft < hidden - 1);
   }
 
   renderSidebar() {
@@ -3368,7 +3380,7 @@ class DeckEditor {
     if(!hint)
       return;
     const show = this.deck() && !this.deckSymbolSelected && this.selectedObject === null && this.faceTemplates.length;
-    hint.textContent = show ? 'Click a face object above or on the list to the left to select, edit, or drag it around.' : '';
+    hint.textContent = show ? 'Click a face object on the card, or in the "Face objects" list, to select, edit, or drag it around.' : '';
     hint.classList.toggle('active', !!show);
   }
 
--- a/client/js/widgets/card.js
+++ b/client/js/widgets/card.js
@@ -1,6 +1,16 @@
+// Counts the cards that were created without an id so each of them still gets a css scope of its own - see
+// the cssScope below.
+let unnamedCardCount = 0;
+
 class Card extends Widget {
   constructor(id) {
     super(id);
+
+    // The css of an html face object is written into a style element and scoped to this class (see
+    // createFaces). A card in a room has a unique id, but the read-only copies the editor renders (deck
+    // editor, card type list, widget picker) are all created without one - they would end up sharing a class,
+    // so the last copy rendered would restyle every other one with its own card type's properties.
+    this.cssScope = id !== undefined ? escapeID(id) : `unnamed${++unnamedCardCount}`;
 
     this.addDefaults({
       width: 103,
@@ -184,7 +194,7 @@ class Card extends Widget {
                     if (typeof object.css === 'object' && object.css !== null && !Array.isArray(object.css)) {
                         const faceIndex = faceTemplates.indexOf(face);
                         const objectIndex = face.objects.indexOf(original);
-                        const uniqueScope = `html-object-${this.id}-${faceIndex}-${objectIndex}`;
+                        const uniqueScope = `html-object-${this.cssScope}-${faceIndex}-${objectIndex}`;
                         objectDiv.classList.add(uniqueScope);
 
                         let styleString = '';
--- a/client/js/widgets/card.js
+++ b/client/js/widgets/card.js
@@ -1,16 +1,6 @@
-// Counts the cards that were created without an id so each of them still gets a css scope of its own - see
-// the cssScope below.
-let unnamedCardCount = 0;
-
 class Card extends Widget {
   constructor(id) {
     super(id);
-
-    // The css of an html face object is written into a style element and scoped to this class (see
-    // createFaces). A card in a room has a unique id, but the read-only copies the editor renders (deck
-    // editor, card type list, widget picker) are all created without one - they would end up sharing a class,
-    // so the last copy rendered would restyle every other one with its own card type's properties.
-    this.cssScope = id !== undefined ? escapeID(id) : `unnamed${++unnamedCardCount}`;
 
     this.addDefaults({
       width: 103,
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -69,12 +69,42 @@ function hasPlayerSpecificChooseField(overlay) {
 
 let lastExecutedOperation = null;
 
+// Counts the widgets that were created without an id so each of them still gets a css scope of its own - see
+// the cssScope below.
+let unnamedWidgetCount = 0;
+
+// A widget without an id is a read-only preview the editor renders: it is simply dropped from the dom instead
+// of going through applyRemove, so nobody removes the stylesheet its css property created. Every such
+// stylesheet is registered here, keyed by scope and valued by the preview it styles, and the ones whose
+// preview has left the document are collected after the render that replaced them - otherwise they would pile
+// up in head as the editor re-renders its previews.
+const unnamedWidgetStyles = new Map();
+let unnamedWidgetStyleCollection = null;
+
+function collectUnnamedWidgetStyles() {
+  unnamedWidgetStyleCollection = null;
+  for(const [ scope, element ] of unnamedWidgetStyles) {
+    if(element.isConnected)
+      continue;
+    if($(`#STYLES_${scope}`))
+      removeFromDOM($(`#STYLES_${scope}`));
+    unnamedWidgetStyles.delete(scope);
+  }
+}
+
 export class Widget extends StateManaged {
   constructor(id) {
+    // Everything that identifies this widget in css - its dom id, the id of its stylesheet element and the
+    // selectors in it - is built from this scope. A widget in a room has a unique id, but the read-only copies
+    // the editor renders (deck editor, card type list, widget picker) are created without one: they would all
+    // share the scope of an empty id, so the last copy rendered would restyle every other one with its own
+    // card type's properties. Widgets that do have an id keep using it, so nothing changes for a room.
+    const cssScope = id === undefined || id === null ? `unnamed_${++unnamedWidgetCount}` : escapeID(id);
     const div = document.createElement('div');
-    div.id = 'w_' + escapeID(id);
+    div.id = 'w_' + cssScope;
     super();
     this.id = id;
+    this.cssScope = cssScope;
     this.domElement = div;
     this.dropShadowWidget = null;
     this.targetTransform = '';
@@ -383,8 +413,8 @@ export class Widget extends StateManaged {
       widgets.get(this.get('parent')).applyChildRemove(this);
     if(this.get('deck') && widgets.has(this.get('deck')))
       widgets.get(this.get('deck')).removeCard(this);
-    if($(`#STYLES_${escapeID(this.id)}`))
-      removeFromDOM($(`#STYLES_${escapeID(this.id)}`));
+    if($(`#STYLES_${this.cssScope}`))
+      removeFromDOM($(`#STYLES_${this.cssScope}`));
     removeFromDOM(this.domElement);
     this.inheritFromUnregister();
     this.globalUpdateListenersUnregister();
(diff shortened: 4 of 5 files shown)


VTT-AI-Agent (commented):
## TL;DR

The fix is correct, minimal and well tested — I reproduced the new TestCafe test passing on this branch. The main thing worth following up: the *same* root cause still bites the widget-level object-form `css` property on preview cards (all previews share the DOM id `w_` and a single `<style id="STYLES_">`), so a deck that tints its card types via `css` instead of an `html` face object still shows every preview in the last card type's colors — I verified that in a browser with this branch checked out.

## Findings

### 1. The other half of the same bug is still there: widget-level object `css` on preview cards

`Card.cssScope` fixes the `html-object-…` class in `createFaces`, but the generic widget CSS path has the identical id-less-preview problem:

* `client/js/widgets/widget.js:75` — `div.id = 'w_' + escapeID(id)`, so every preview widget created without an id has DOM id `w_` (duplicated across all previews).
* `client/js/widgets/widget.js:694-710` — object-form `css` goes through `cssToStylesheet`, which writes selectors `#w_${escapeID(this.id)}` → `#w_` into a `<style id="STYLES_${escapeID(this.id)}">` → `STYLES_`, and `css()` (line 626) removes the previous element with that same id. Net effect: only the last preview's stylesheet survives and, because `#w_` matches all the duplicate-id elements, it paints *every* preview.

Verified on this branch with a scratch TestCafe test: a deck whose `cardDefaults.css` is `{ default: { 'background-color': '${PROPERTY tint}' } }` with card types tinted red/green/blue renders the deck-editor strip as `['rgb(0, 0, 255)', 'rgb(0, 0, 255)', 'rgb(0, 0, 255)']` — exactly the symptom this PR fixes for `html` face objects. (String-form `css` is unaffected; it becomes an inline style attribute.)

That's pre-existing, not a regression from this PR, so it needn't block it — but since `css` on a card type is at least as common a way to tint cards as an `html` face object, a user hitting it will report "you didn't fix it". The natural shape is to move `cssScope` up into `Widget`'s constructor (it is not card-specific) and use it for the `STYLES_` element id and the `#w_` selector in `cssToStylesheet`/`css()` — id-less previews are never in the `widgets` map, so nothing looks them up by DOM id via `unescapeID`.

### 2. `unnamed<n>` can collide with a real widget id

`client/js/widgets/card.js:13` — a card in a room with the id `unnamed1` yields `escapeID('unnamed1') === 'unnamed1'`, i.e. the same scope a preview can get. Vanishingly unlikely, but free to rule out: `escapeID` never emits a lone `_` (it doubles it), so `` `unnamed_${++unnamedCardCount}` `` is collision-proof by construction.

While there: the guard is `id !== undefined`, but `escapeID` also returns `''` for `null`/`''`, which would put us back to a shared empty scope. `const scope = escapeID(id); this.cssScope = scope || \`unnamed_${++unnamedCardCount}\`;` covers both cases and is shorter.

### 3. No test for the second bug the PR claims to fix

The description says running the id through `escapeID` also fixes cards whose id contains a character invalid in a class name (`classList.add()` threw, so the html face objects of that card never rendered at all). That's a real, separate behaviour change for rooms — and it's the only part of the diff that changes what an existing game renders — but the new test only covers the preview-scoping case. A three-line addition to the new test (a card with a space in its id inside the room, expecting its face object to exist / be tinted) would lock that in cheaply.

Otherwise the change looks right: the scope is computed once in the constructor, which matches how `widget.js` already fixes `div.id` at construction time (an id change re-creates the widget), the generated class is self-consistent within one render so no game breaks on ids containing `_`, and the `html-card` DOM snapshot keeps `html-object-card-0-0` because `escapeID('card') === 'card'`. The new test passes here (`✓ Deck editor: every card type preview keeps its own html face css`, 15s).

I added the `beta` label: the only engine-directory file touched is `client/js/widgets/card.js`, and the edit is incidental — it just renames an internally-generated CSS class (byte-identical for every normal card id, self-consistent for the rest), touches no state sync, widget behaviour or routine/expression semantics, and cannot corrupt a game or crash the server.

VTT-AI-Agent:
Thanks — all three findings addressed in 32693b88.

**1. Widget-level object-form `css` on previews.** Confirmed and fixed the way you suggested: `cssScope` moved from `Card` up into `Widget`'s constructor and is now used for the DOM id, the `STYLES_…` element id and the `#w_…` selectors in `cssToStylesheet`/`css()`/`applyRemove`/`showEnlarged`, on top of `Card`'s `html-object-…` class. Reproduced your case first (a deck of five card types tinted from `cardDefaults.css` rendered all five previews purple, the last type's color) and re-checked after:

| before | after |
| --- | --- |
| before | after |

One thing that fell out of it and is worth flagging: giving previews their own `STYLES_` id turns a harmless situation into a leak. Preview widgets are dropped from the DOM instead of going through `applyRemove`, so nothing removes their stylesheet — with the shared `STYLES_` id the next preview simply overwrote it, but with one per preview they pile up in `<head>`. Measured in a browser: the count grew by 4 per click in the deck editor's card type strip and never came back down (46 stylesheets after 10 clicks). So preview stylesheets are now registered in a small module-level map and collected once their preview is no longer `isConnected`; the collection is deferred with a `setTimeout` so a preview that is built into a container before that container is inserted is not collected mid-render. With that, the same 10 clicks stay at a constant 7 (4 from the room + the 3 live strip previews).

**2. `unnamed<n>` collision.** Fixed, it is `unnamed_<n>` now — as you say, `escapeID` doubles `_`, so it can never produce that.

I did *not* fold `null`/`''` into the same branch, though: the scope now also decides the DOM id, and a widget whose id is `''` (or the falsy `0`) *is* a real widget in the `widgets` map that mouse handling, zoom and the JSON editor find through `unescapeID(el.id.slice(2))`. `escapeID('')` is `''`, so it currently gets `w_` and those lookups work; giving it `w_unnamed_7` instead would silently make it unclickable. So the condition is `id === undefined || id === null` — `null` is covered (nothing constructs a widget with it today, but it is not a valid state id either), `''` deliberately keeps its old behavior.

**3. Missing test for the `escapeID` crash fix.** Added to the same test — the room now also contains a card with the id `my card`, and the test asserts its html face object renders (`.eql('rgb(0, 128, 0)')` on `#w_my_x0020_card .cardFace.active .cardFaceObject`). Verified it fails against `main`'s `card.js` with `expected null to deeply equal 'rgb(0, 128, 0)'`. The test also picked up a second assertion for the `css` half of finding 1 (border color per preview), which fails without that fix.

Re-ran on the updated branch: `npm test` 1167 passed, `tests/testcafe/editor.js` 40 passed, `tests/testcafe/pixelsnapshot.js` 4 passed, `tests/testcafe/domsnapshot.js` `html-card` + `holder-image` pass and `widget-gallery` fails identically on plain `main` in this sandbox (the unrelated `hidden`/opacity timing diff).

VTT-AI-Agent (commented):
## TL;DR

This is a **UI/UX review** (layout, visual design, first-time readability) — not a code review. The change is a clear visual win: on `main` every card-type preview in the deck editor wore the *last* card type's colors, and a card whose id contains a space rendered no face at all; on this branch each preview is correctly tinted and that card renders. I found **no visual regression** in any overlay I could reach, at 390×844 / 820×1180 / 1280×800 / 1920×1080. The one thing I'd add on top of the fix: the `CARD TYPES` strip is now genuinely informative, and it scrolls horizontally with **no visible affordance** — with 20 card types the last visible card is sliced off at the right edge and nothing hints there is more.

### What I actually exercised

I ran the server locally on this branch and on `main` with a purpose-built deck (6 card types tinted from `${PROPERTY tint}`/`tint2`/`ink`, once via an `html` face object and once via `cardDefaults.css`, plus a room card whose id is `my card`), and drove it with Playwright.

Overlays opened and looked at, at **390×844 (phone portrait), 820×1180 (tablet portrait), 1280×800 (laptop), 1920×1080 (desktop)**: Game Shelf / states, Players, Options, About/help, the Edit-widgets panel, Add-widget picker, widget properties (deck and card), the JSON editor, the deck editor (card-type strip, main card view, ALL CARDS / FACE / CARD TYPE / OBJECT tabs), and the deck editor's sub-dialogs — *Add a new deck* with all three sections expanded (New blank deck / Use an existing deck / Create a custom deck), *Import CSV*, *Export / Print*.

**Could not reach**, at phone-portrait 390×844 and tablet-portrait 820×1180: everything behind edit mode — the properties panel does not appear, so Add-widget, the deck editor and its sub-dialogs are unreachable. That is **pre-existing**, identical on `main` ([main, tablet portrait](https://agent.virtualtabletop.io/reports/ui-review/1536159582819196951/shots/MAIN-tablet-editpanel.png) vs [this branch](https://agent.virtualtabletop.io/reports/ui-review/1536159582819196951/shots/tablet-editpanel.png)) — not something this PR introduces or should fix.

### Before / after, reproduced independently

| `main` | this branch |
| --- | --- |
| [before](https://agent.virtualtabletop.io/reports/ui-review/1536159582819196951/shots/MAIN-deckeditor-type.png) | [after](https://agent.virtualtabletop.io/reports/ui-review/1536159582819196951/shots/desktop-deckeditor-type.png) |

On `main` all six strip previews *and* the big card view are cyan (the last card type) while `cat-blue` is selected — the editor was actively lying about what you were editing. On this branch the strip, the main card view and the little tree thumbnail on the left all agree with the selected type. That is a real usability fix, not just a cosmetic one.

Same for the room card with a space in its id — it is [completely invisible on `main`](https://agent.virtualtabletop.io/reports/ui-review/1536159582819196951/shots/MAIN-room.png) (only Red and Blue render) and [renders on this branch](https://agent.virtualtabletop.io/reports/ui-review/1536159582819196951/shots/00-room-1280.png) (Red, Blue, Purple).

The Add-widget picker is [byte-for-byte the same layout](https://agent.virtualtabletop.io/reports/ui-review/1536159582819196951/shots/laptop-addwidget.png) as on [`main`](https://agent.virtualtabletop.io/reports/ui-review/1536159582819196951/shots/MAIN-addwidget.png), and I confirmed the `<head>` stylesheet count stays flat (7 preview sheets) across ten strip clicks, so nothing visibly degrades as the editor re-renders.

---

## Big layout improvements

**1. The card-type strip scrolls sideways with no affordance — and this PR is exactly what makes that matter.** With the previews now individually coloured, the strip becomes the primary way to find a card type by sight, so silently hiding half of it hurts more than it used to. With 20 card types at 1280×800 the strip is `scrollWidth 1875` vs `clientWidth 1044`: card 11 is sliced vertically at the right edge and there is no scrollbar, no fade, no arrow — a newcomer will read "Type 0 … Type 10" as the whole deck.

[20 card types, laptop](https://agent.virtualtabletop.io/reports/ui-review/1536159582819196951/shots/laptop-deckeditor-20types.png)

Concretely, and in keeping with the app's existing flat/blue style, one of:
* give `#deckEditorStrip` a persistent thin scrollbar (`scrollbar-width: thin` plus a `::-webkit-scrollbar` rule matching the editor sidebar), or
* add a right-edge gradient mask that only shows while `scrollLeft + clientWidth < scrollWidth`, or
* let the strip wrap to two rows (`flex-wrap: wrap` with a `max-height` and vertical scroll) — with the counter row under each card, wrapping reads more naturally than a mile-long single line.

This is pre-existing, so it is fair to leave it out of this PR — but it is the single biggest thing standing between this fix and the strip being genuinely useful on a 13-to-52-type deck.

**2. Consider making the fix visible where the payoff is largest: the `Card type` section of card properties.** That section shows exactly one preview at a time behind a `− 1 +` stepper ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1536159582819196951/shots/laptop-cardprops.png)). Now that a preview is guaranteed to be tinted correctly, a small horizontal row of the deck's card types (the same markup as the strip, at ~60 px wide) with the current one outlined would let a user *pick* the card type by looking at it, instead of stepping blindly through an index. Again, out of scope for a bugfix PR — but this PR is the prerequisite that makes it worth doing.

I looked for regressions specifically in the surfaces the diff touches and found none. What I did notice, all pre-existing and none of it caused by this change:

* **`CARD TYPES` strip, selected state is faint.** The selected card type is marked only by a 1 px orange outline plus a lighter background (`.deckEditorStripCard` selected state) — at 1920×1080 with 6 saturated cards next to each other it is easy to miss which one you're editing. A thicker (2 px) outline or a filled label chip under the selected card would read at a glance.
* **`− 0 +` counters below each card type have no unit or label**, and at 20 card types they line up under a strip that is itself clipped, so the `CARD COUNT — ALL / + ALL` header on the far left drifts out of association with them. Worth a `title` at minimum (the individual `−`/`+` buttons do have "One fewer/more card of this type" — the number itself does not).
* **Phone portrait, Game Shelf**: the overlay is clipped by the bottom toolbar — the "Public Library" heading sits behind the toolbar and the panel does not scroll clear of it ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1536159582819196951/shots/phone-states.png)). Same on `main`; flagging only because I was asked to sweep all overlays.
* **Phone portrait, room**: the "Please rotate your device to landscape orientation and switch to fullscreen." message is itself clipped mid-word ("landscap…") ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1536159582819196951/shots/phone-editpanel.png)). Pre-existing.

The PR improves this materially without adding any text: an editor whose previews all show the same colour teaches a newcomer that card types *don't* control appearance, which is exactly backwards. After this change, clicking through the strip demonstrates the property→CSS binding visually, which is the fastest way to explain `${PROPERTY tint}` to someone who has never seen it.

Two small wording/affordance notes on the surfaces this touches:

* The empty state in the deck editor's main pane reads *"Edit the default properties of every card in this deck to the right, or select a card type below to edit that card."* — "**to the right**" and "**below**" are layout-relative, and at 820×1180 the right-hand panel isn't where it is at 1920×1080. The `noSelectionIntro` text elsewhere in the editor already avoids exactly this ("would be wrong in the layouts that put the panel above the room"); the same caution applies here — "in the panel on the right" → "in the properties panel".
* *"Click a face object above or on the list to the left to select, edit, or drag it around."* sits in 11-ish px grey on white at the very bottom of the card view — low contrast and easy to miss, and it is the only hint that the big card is interactive. Bumping it to the body text colour would help a first-timer notice it.

Neither is introduced by this PR.

---

**Verdict:** visually this is a straight improvement with no regressions I could find across four viewport sizes and every overlay reachable at each. Nothing here blocks it.

VTT-AI-Agent:
Thanks — the strip findings are all reproducible here, so I took the ones that are about the surface this PR just made informative. Pushed in 290635fe.

## Fixed

**1. The card type strip scrolls sideways with no affordance.** Reproduced exactly: 20 card types at 1280×800 is `scrollWidth 1875` vs `clientWidth 1044`, and `offsetHeight - clientHeight` is `0` — the platform draws an *overlay* scrollbar here, so the repo's global `::-webkit-scrollbar` rules never get a chance to show anything. I tried `overflow-x: scroll`, `scrollbar-width: thin` and a `#deckEditorStrip::-webkit-scrollbar` height in the browser first; none of them reserved a single pixel. So the persistent-scrollbar option isn't actually available, and I went with your second one, the fade.

Whichever edge still has tiles behind it now fades out, driven by a new `updateStripOverflow()` (called from `renderStrip`, the strip's `scroll` handler and the window `resize` handler). The fades are sticky flex pseudo-elements with a compensating negative margin, so they pin to the visible edge without adding to the scrollable width (`scrollWidth` stays 1875) — and at `z-index: 1000002`, above the preview cards' own `1000000`, which is why my first attempt at 1 was invisible behind them.

| before (main + this PR's fix) | after — at the start, right edge fades |
| --- | --- |
| before | after |

Scrolled into the middle, both edges fade; scrolled to the end, only the left one does; a deck whose types all fit gets neither (`className` is empty, verified):

scrolled

Also checked in dark mode — the fade uses `--backgroundColor`, so it goes to `#08090a` there: [screenshot](https://agent.virtualtabletop.io/reports/pr/1536134762509504622/deckeditor-strip-dark-after.png).

I did not take the wrapping option: two rows of tiles at 1280px is ~310px of vertical space taken away from the card view, and a 52-type deck would still need a scroll affordance on top.

**2. Faint selected state.** Agreed, and this PR is what makes it matter — the tiles are now as saturated as the card types they show, so a 2px outline competes with them. The selected tile now also gets the `--backgroundHighlightColor1` plate and a bold label, which reads at a glance without inventing a new color:

selected

**3. The `− n +` count has no label.** The number itself now carries `title="Cards of this type in the game"`, next to the `−`/`+` buttons' existing tooltips.

**4. Layout-relative strings.** Both replaced by the name of the thing they point at, so they survive any width:

* *"…in this deck **to the right**, or select a card type **below**…"* → *"…in this deck **in the properties panel**, or select a card type **in the "Card types" strip**…"*
* *"Click a face object **above or on the list to the left**…"* → *"Click a face object **on the card, or in the "Face objects" list**…"*

**5. Low-contrast object hint.** `#deckEditorObjectHint` went from `opacity: 0.7` to `0.9` — as you say, it is the only hint that the big card is interactive.

The whole deck editor after all of this: [screenshot](https://agent.virtualtabletop.io/reports/pr/1536134762509504622/deckeditor-full-after.png).

## Not doing here

* **A card type picker in the `Card type` section of card properties.** I agree this PR is the prerequisite that makes it worth doing, but it is a new control rather than a fix, and picking a card type by clicking a preview instead of stepping an index is a design decision for a maintainer — worth its own issue/PR rather than being slipped into a CSS-scoping bugfix.
* **Phone-portrait Game Shelf clipped by the toolbar, and the clipped "rotate your device" message.** Pre-existing (you confirmed both on `main`) and in overlays this diff doesn't touch — bundling them here would make this PR span three unrelated concerns.
* **Edit mode being unreachable at 390×844 / 820×1180.** Same, pre-existing and out of scope.

## Re-verified

`npm test` 1167 passed. `tests/testcafe/editor.js` 40 passed (includes the 19 deck editor tests and this PR's own preview-css regression test). The strip's `scrollWidth` is unchanged with the fades in place, so the scroll restore and `scrollIntoView` logic in `renderStrip` behaves exactly as before.

VTT-AI-Agent (commented):
## TL;DR

The current revision looks correct and the previously reported gaps are now covered; I found no actionable correctness, security, regression, or test-coverage issue. The important deployment note is that `client/js/widgets/card.js:184` deliberately changes how real room cards with escaped IDs render, so this is an actual engine-behavior change and is ineligible for beta auto-deployment even though most of the work serves editor previews.

## Findings

No actionable code findings.

I re-checked the complete five-file diff and the earlier code-review findings against the current head:

- `client/js/widgets/widget.js` now gives every id-less preview a collision-resistant per-instance scope and uses it consistently for the DOM id, object-form `css` selectors, stylesheet ids, removal, and enlargement. Existing non-empty widget IDs retain their prior escaped scope.
- `client/js/widgets/card.js:184` uses that same scope for non-iframe HTML face-object CSS, including IDs containing characters that cannot be placed directly in a class name.
- `tests/testcafe/editor.js:911` covers distinct HTML-object colors, distinct widget-level object-form CSS colors, and a real room card whose ID contains a space. All current GitHub checks, including both editor TestCafe jobs, DOM/pixel snapshots, build, production startup, and CodeQL, pass.
- This fixes existing `css`/HTML-face behavior and does not expose a new widget property, routine operation, or author-facing feature, so it does not require a new property-editor/block-UI control or a `library/tutorials/` addition.

## Beta eligibility

The engine-file edit is not merely incidental under the stated rule: the PR explicitly fixes a real card whose ID contains a space from throwing during `classList.add()` and failing to render its HTML face object. That changes how an existing game runs, so the engine exclusion applies; the builder should continue to filename-skip it, and a maintainer should remove the already-present label if they want the PR metadata to reflect that classification.

The `beta` label was already present and I left it untouched as required, but the PR is engine-ineligible because it changes HTML face rendering for real room cards with escaped IDs.

VTT-AI-Agent:
Thanks for the re-review — nothing to change on my side for this one, so no new commit. Confirming the three points I could act on:

* **No code findings.** Agreed with the re-check of the five-file diff; the head is still `290635fe` and all checks are green (both `editor.js` TestCafe jobs, DOM/pixel snapshots, build, production startup, CodeQL).
* **No new property / operation / author-facing feature.** Correct — the diff only changes how existing `css` and `html` face-object CSS is scoped, so there is no property-editor or routine-block UI to add and no `library/tutorials/` room to build. That is why this PR has neither.
* **Beta eligibility.** I agree with your reading: `client/js/widgets/card.js` is not incidental here, because a real room card whose id contains a character that has to be escaped (`my card`) goes from "html face object does not render at all" to "renders", and that is a behavior change for an existing game. As the PR author I deliberately do not set or clear the `beta` label myself — it was added by review 1/4 and I am leaving it exactly as it is for a maintainer to decide, rather than overriding one automated reviewer with another.

*(For anyone reading later: the branch is currently behind `main` — no conflicts, `MERGEABLE`. `main`'s newest deck editor change, #3114, touches `.deckEditorGroupHeader:hover`, which is a different element from the card type strip this PR restyles.)*

96LawDawg (approved):
Tested. Approved.

PR #3114, opened 2026-08-09
Fix unreadable deck editor property group headers on hover in light mode

VTT-AI-Agent:
## What this changes

In the deck editor's right sidebar, the property blocks (Content / Position / Size / Colors / Appearance / Custom) have collapsible headers. Hovering one of those headers turned it dark navy while the label stayed the theme's text color — in light mode that is **black on `--VTTblueDark` (#0d2f5e), ~1.6:1 contrast**, so the section title became unreadable exactly while you were pointing at it. This gives the header its own hover fill so it stays readable.

## Why it happened

`.deckEditorGroupHeader` (specificity 0,1,0) sets `background: var(--backgroundHighlightColor1)` and `color: var(--textColor)`, but the global `button:hover { background-color: var(--VTTblueDark) }` in `client/css/layout.css` is *more* specific (0,1,1) and won on hover — while the `color` stayed the theme's. Dark mode was unaffected because there `--textColor` is near-white. The Edit Widget sidebar's equivalent headers (`.editorModule .collapsibleHeader`, specificity 0,2,0) already outrank that rule, which is why only the deck editor showed it.

## The fix

One rule in `client/css/editor/deckeditor.css`:

```css
.deckEditorGroupHeader:hover {
  background: var(--backgroundHighlightColor2);
}
```

`--backgroundHighlightColor2` is one step lighter than the header's resting `--backgroundHighlightColor1` in both themes (`#dedede` on `#ccc` in light, `#565656` on `#444` in dark), so the header still visibly reacts to hover — it just does it with a fill the theme's text color is legible on (15:1 light, 7:1 dark).

## Verification

* `npm test` — 1167 passed.
* TestCafe `editor.js` → "Deck editor: add card type, dynamic object, delete face, undo" passes.
* Checked in a real browser (deck editor → face object → hovering the "Content" header), light and dark:

| | |
|---|---|
| before, light mode (bug) | before |
| after, light mode | after light |
| after, dark mode | after dark |

CSS-only, hover-state-only: no widget properties, routine functions or room state are touched, so no editing-UI, validator or `jeCommands` work applies.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3114/pr-test (or any other room on that server)

### DIFF
--- a/client/css/editor/deckeditor.css
+++ b/client/css/editor/deckeditor.css
@@ -1419,6 +1419,13 @@ body.deckEditorPanning #deckEditorMain * {
   cursor: pointer;
 }
 
+/* The global button:hover fill (layout.css) is more specific than the rule above, so without this the header
+   turned VTTblueDark on hover while its label stayed --textColor - black on navy, unreadable in light mode.
+   Highlight it with the theme's own lighter shade, which keeps the label readable in both themes. */
+.deckEditorGroupHeader:hover {
+  background: var(--backgroundHighlightColor2);
+}
+
 /* Focus ring like the Edit Widget sidebar's collapsible headers (propertyInputs.css), which this mirrors. */
 .deckEditorGroupHeader:focus-visible {
   outline: 2px solid var(--VTTblue);


96LawDawg (approved):
Approved.

VTT-AI-Agent (commented):
## TL;DR

Looks good — a 7-line, CSS-only fix that adds the missing `:hover` fill to the deck editor's property-group headers. The specificity analysis in the description checks out (`.deckEditorGroupHeader:hover` = 0,2,0 beats `button:hover` = 0,1,1 in `client/css/layout.css:971`), the chosen `--backgroundHighlightColor2` is defined for both themes in `client/css/editor/layout.css:6`/`:70`, and nothing else in the file sets a competing background on that element. I found nothing worth changing.

## Details

Things I checked and found fine:

* **No conflicting rule.** `.deckEditorGroupHeader` only appears at `client/css/editor/deckeditor.css:1406` (resting) and `:1423` (`:focus-visible`); there is no `.deckEditorGroup.collapsed .deckEditorGroupHeader` variant that would need a matching hover, so the new rule at `:1422` covers every state the header can be in.
* **Both themes.** `--backgroundHighlightColor2` is `#dedede` (light) / `#565656` (dark), one step off the resting `--backgroundHighlightColor1` (`#ccc` / `#444`), so hover feedback stays visible and the header's `color: var(--textColor)` stays legible. The `.deckEditorGroupSummary` at 0.75 opacity is still comfortably above 4.5:1 on both fills.
* **Consistency.** Using a `--backgroundHighlightColorN` step matches the sibling hover treatments in this file (e.g. `.deckEditorNewDeckGroupHeader:hover`, `client/css/editor/deckeditor.css:745`). Worth noting the Edit Widget sidebar's mirrored headers (`.editorModule .collapsibleHeader`, `client/css/editor/propertyInputs.css:702`) have *no* hover feedback at all — they just outrank `button:hover` and keep their resting fill. So this PR makes the deck editor slightly nicer than the thing it mirrors rather than identical to it, which seems like the right call; no change requested.
* **Scope.** Correct that no widget property, routine function or room state is involved, so no editing-UI, `jeCommands`, validator or tutorial work applies here. No test is expected for a hover-only CSS rule — TestCafe compares state hashes and would not see this.

Non-blocking micro-note: other button hovers in this file guard with `:hover:not(:disabled)`, but this header is never rendered disabled (`client/js/editor/deckeditor.js:3091`), so the plain `:hover` is fine as written.

I added the `beta` label: the PR touches only `client/css/editor/deckeditor.css` — no engine files (`client/js/widgets/`, `compute.js`, `serverstate.js`) — and a hover-only CSS rule cannot corrupt games or crash the server, so it is safe and useful for players to try on beta.

PR #3116, opened 2026-08-11
Let a spreading holder arrange piles instead of dissolving them

VTT-AI-Agent:
> **Closed without merging — superseded by #3117.** The `layout` property in #3117 carries this branch's pile/spread engine forward, so the work here is not lost; the editing-UI question below is answered there instead. Kept for reference.

## What this does

A holder that spreads its children out (`stackOffsetX`/`stackOffsetY`) has so far refused piles: cards dropped onto each other stayed separate, and a pile dropped into it was emptied out card by card onto the row. This adds a holder property **`allowPiles`** that turns such a holder into one that *arranges piles* — a solitaire tableau of columns, a row of stacks, a set of discard piles — while everything that reaches into the holder keeps counting the cards it holds.

Requested by @RaphaelAlvez (rapha195).

## The properties

Holder:

| property | meaning |
| --- | --- |
| `allowPiles` (bool, default `false`) | the holder arranges piles instead of dissolving them |
| `pilesOffsetX` / `pilesOffsetY` (number or `null`) | the next pile starts this far from the previous one, whatever it holds |
| `pilesGapX` / `pilesGapY` (number or `null`) | the next pile starts behind the cards of the previous one, plus this gap |
| `spreadMin` (number or `null`) | inherited by the piles (see below) |

Pile:

| property | meaning |
| --- | --- |
| `stackOffsetX` / `stackOffsetY` | how the pile spreads its own cards; inherited from the holder when it arranges piles |
| `spreadMin` | how many of the topmost cards keep that offset; everything below them is squeezed to a tenth of it |

Rules:

- A gap wins over an offset on the same axis. With neither of them given, the piles are placed flush, one right after the other, and the axis the piles are spaced out on is the axis the holder lines its children up along — its `stackOffset` then describes the cards *inside* the piles.
- What a drop lands on is what it joins: a card dropped anywhere on a stack or along a fanned column goes into that pile, a pile dropped onto another pile merges with it, and a drop on free space starts a new pile. What aims the drop is the spot the player is holding the dropped widget by. Outside such a holder nothing changes — there a pile still takes a drop within its snap range, now anywhere along its spread rather than only at its corner.
- A pile that spreads its cards is as big as its spread: its handle sits at the end of it, the holder gives it that much room, and it shrinks back and collects its cards when it stops spreading (when it is dragged out of the holder it took its offsets from, for instance).
- `MOVE` puts everything one operation brings in down as **one new pile**, instead of feeding the cards into whichever pile the holder already ends with. Counting is unchanged: `MOVE`, `COUNT` and `dropLimit` count the cards in the holder, not the piles they are arranged in — and turning `allowPiles` off again empties the piles it was arranging back onto the row, so that stays true.

Everything sits behind new properties whose defaults reproduce today's behaviour exactly, so no `fileupdater` migration and no legacy mode are needed.

## Follow-up on the first round of feedback

Two things reported after the first version, both fixed in the second commit:

- **A card could not be put into a pile that is already in the holder.** The drop only combined when the corner of the card landed within ten pixels of the corner of the pile, which in a holder full of stacks or fans meant hitting a target much smaller than what the player is aiming at. In a holder that arranges piles, the widget the drop lands on now decides where the card goes, across the whole box of that widget. Dropping a pile onto another pile merges them for the same reason — a tableau column has to be movable onto another column.
- **A pile is now sized to its spread** instead of staying the size of a single card, so a fanned column really is that column: its handle sits at its end, the holder measures it as what it is, and a drop anywhere on it is a drop on it. Piles that do not place their cards themselves — a scattered heap of tiles, say — are left alone.

## Test room

The demo room is authored to tutorial quality — three variants, **Basic** (what `allowPiles` changes), **Spacing** (`pilesOffset`/`pilesGap`) and **Spreading** (the inherited `stackOffset` and `spreadMin`), with the `_meta` fields, the wording and the square SVG library image of the tutorial family — but per the standing rule it is **not part of this PR**: no file under `library/` is touched here.

It lives on the test server instead, in the [`pr-test-ai` room](https://test.virtualtabletop.io/PR-3116/pr-test-ai) of this PR, and as a save file to load anywhere: [Properties-allowPiles.vtt](https://agent.virtualtabletop.io/reports/pr/1536770300362887199/Properties-allowPiles.vtt). Screenshots of the three rooms: [Basic](https://agent.virtualtabletop.io/reports/pr/1536770300362887199/tutorial-v2-0.png) · [Spacing](https://agent.virtualtabletop.io/reports/pr/1536770300362887199/tutorial-v2-1.png) · [Spreading](https://agent.virtualtabletop.io/reports/pr/1536770300362887199/tutorial-v2-2.png). It can be added to `library/tutorials/` on request — just say so and it goes into this branch as a tutorial.

## Follow-up on the reviews

Round two (UI/UX) reworked the three demo rooms — pile handles that stay readable, one colour per meaning, deal buttons beside the holder they fill, and the terms a first-time reader needs spelled out. Round three found a pile that jumped out from under the pointer when it was picked up by a handle at the far end of its fan (the drag anchor now keeps its place in the box when a pile changes size mid-drag) and moved the unit tests for the drop arithmetic into jest. Round four is in the last commit:

- **Turning `allowPiles` off no longer strands the piles in the holder.** A holder that spreads its children can otherwise hold no pile at all — every pile dropped into one is emptied out, one card per slot — so the piles were left in a state nothing else in the file expects: `children()` reported the four columns instead of the twenty cards in them, and `COUNT`, `dropLimit` and `MOVE ... from:` silently counted piles from then on. The piles are now emptied out the same way a dropped one is. Where the holder does not spread its children (piles spaced out by `pilesOffset`/`pilesGap` alone), a pile is a perfectly ordinary child and stays.
- **What aims a drop is where the player is holding what they dropped**, not the middle of its box. A pile with a `stackOffset` of its own keeps its fan while it is carried, so its middle can be several cards away from the pointer, and a column dropped visibly onto one column could join the next. The drop keeps the anchor the drag was taken with; where nothing is holding the widget — a routine putting one down — the middle of its box aims it as before.
- **A loose card gets the same flush slot as a pile holding just it**, so a row of `allowPiles` columns with a single card among them has no wider slot in the middle of it.

## Editing UI — pending a decision

The new properties are already reachable from the property editor through *Other properties → Add property* (they are suggested there, since that list comes from the widget defaults), and the JSON editor gets its insert buttons and the validator its entries automatically. What is **not** in this PR is a curated place for them in the holder and pile editors, because nothing in the request says how they should look. A proposal is in a comment below; it was never built, and the question moved to #3117 along with the rest of this work.

## Testing

- `npx jest` — 33 suites, 1295 tests pass, including `tests/client/pile-spread.test.js` (13 cases: the spread, `spreadMin`, the extent and the box of a pile, the drops it takes, that an ordinary pile keeps its cards where they were put, and the inheritance from the holder) and `tests/client/holder-piles.test.js` (15 cases: what a drop lands on and what it does not, what aims it, which pile wins where two overlap, the gap/offset/flush spacing rules, the axis the piles are lined up along, and what becomes of the piles when `allowPiles` is turned off).
- TestCafe in Chromium: `tests/testcafe/allowpiles.js` (8 cases driving the pointer over a tableau: what a drop lands on, a column onto a column, a fanned pile aimed by the pointer rather than by its middle, a column carried by a handle at the end of its fan, a column dragged out, turning `allowPiles` off, and a `MOVE` from a holder to itself), plus `holderevents.js`, `interactions.js`, `domsnapshot.js` and `publiclibrary-1..4.js` all passing unchanged — no state hash moved.
- Walked all three demo variants in a browser: dropping cards onto each other, dragging a pile in and out, the deal buttons, and the counting behaviour of `MOVE`.

## Wiki

Creator-facing: the wiki pages for the holder and the pile need the six new properties (`allowPiles`, `pilesOffsetX/Y`, `pilesGapX/Y`, `spreadMin`, plus `stackOffsetX/Y` on the pile).

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3116/pr-test (or any other room on that server)


VTT-AI-Agent:
## Proposal for the editing UI (feedback wanted before I build it)

The six new properties are usable from the property editor today only through *Other properties → Add property*, which gives them a generic input. The standing rule is that they get a curated place in the widget editor, but the request did not say what that should look like, so here is what I would build — please say yes / change it before I do.

**Holder editor** (`renderForHolder` in `client/js/editor/sidebar/properties.js`):

1. In the existing **Behavior** group of `editorTypeSections.holder`, right under "Prevent piles" (the two are opposites, so they belong together):
   - `{ label: 'Arrange piles', property: 'allowPiles', kind: 'checkbox' }`
   with the hint *"Let this holder hold piles while it spreads its children out: cards dropped onto each other combine, and a pile dropped in stays a pile."*
2. Two number pair rows directly below the existing "Drop offset" and "Stack offset" rows, in the same `renderNumberPairRow` style, shown only while `allowPiles` is on (`available` + `availableListenTo: [ 'allowPiles' ]`, the way conditional curation already works there):
   - **Piles offset** — X: `pilesOffsetX`, Y: `pilesOffsetY`, `nullIfEmpty: true`, placeholder empty
   - **Piles gap** — X: `pilesGapX`, Y: `pilesGapY`, `nullIfEmpty: true`, placeholder empty
   Empty = not set, which is what `null` means in the engine, and the gap row's hint says it wins over the offset on the same axis.
3. One more number input under them, also only while `allowPiles` is on:
   - **Spread minimum** — `spreadMin`, `min: 0`, `nullIfEmpty: true`, hint *"How many of the topmost cards of each pile keep the full stack offset. The rest are squeezed to a tenth of it. Inherited by the piles in this holder."*

**Pile editor** (`renderForPile`): a "Stack offset" X/Y pair plus "Spread minimum", in the **Behavior** group next to "Snap range" and "Stack cards on top of each other", with placeholders showing the values inherited from the holder so it is visible that leaving them empty follows the holder. These three would also join `templateProperties` in `renderPileTemplateSection`, so that with the pile-template switch on they are written to the deck's `onPileCreation` like the other pile properties.

Alternatives I considered and did not pick, in case one of them is preferred:

- Putting the piles spacing into the existing "Stack offset" row as two extra fields — rejected, four numbers in one row and it hides that they only apply to piles.
- A separate "Piles" sub-section in the holder editor instead of rows under the offsets — happy to do this instead if the offsets area gets too crowded.
- Showing the rows always instead of only while `allowPiles` is on — they do nothing without it, so hiding them seemed kinder, but always-on is one flag away.

VTT-AI-Agent (commented):
## TL;DR

Nice, carefully defaulted feature — every new behaviour is behind `allowPiles`/the new pile properties, and I could not move any existing behaviour (jest 1274 ✓, TestCafe `holderevents` + `interactions` 54 ✓ locally). Two real bugs though: **a pile never un-spreads** — drag a spread pile out of the tableau (or set `stackOffsetY` back to 0) and its cards stay fanned out forever — and **`MOVE` from a holder to itself now collapses that holder's loose cards into one pile**. The curated property-editor inputs are also still missing (standing maintainer rule), so the properties are only reachable through the generic "Add property" list.

## Findings

### 1. A pile that stops spreading keeps its cards fanned out (`client/js/widgets/pile.js:284-300`) — confirmed

`arrangeChildren()` only writes positions inside `if(this.spreadsCards())`. When a pile *stops* spreading, nothing puts the cards back on top of each other, so they keep the offsets of the last spread layout — while the pile box itself is still 1×1 and the handle sits at its corner.

I reproduced both paths with the harness of your own `tests/client/pile-spread.test.js`:

```js
// 1) the author turns the spread off
const pile = withCards(createPile({ id: 'p', stackOffsetY: 40 }).get('id'), 3);
await pile.arrangeChildren();              // [0, 40, 80]
await pile.set('stackOffsetY', 0);
// -> still [0, 40, 80]

// 2) the pile is dragged out of the tableau (inherited offset gone)
createHolder({ id: 'h', allowPiles: true, stackOffsetY: 40 });
const q = withCards(createPile({ id: 'q', parent: 'h' }).get('id'), 3);
await q.arrangeChildren();                 // [0, 40, 80]
await q.set('parent', null);
// -> still [0, 40, 80], while q.get('stackOffsetY') is now 0
```

Path 2 is the everyday one: picking a column up by its handle and dropping it on the table leaves a fan of cards lying loose-looking outside the holder, and it stays that way when it is dropped anywhere that does not spread. Toggling `allowPiles` off on the holder has the same effect (`holder.js:359` calls `child.arrangeChildren(false)`, which is then a no-op).

Suggested fix: bail out only when the pile does not align its children at all, and otherwise always lay the cards out, with a zero step when it does not spread:

```js
async arrangeChildren(notifyHolder=true) {
  if(this.get('alignChildren')) {
    const children = this.children().reverse();
    const spreads = this.spreadsCards();
    let x = this.get('dropOffsetX');
    let y = this.get('dropOffsetY');
    for(let i=0; i<children.length; ++i) {
      await children[i].setPosition(x, y, children[i].get('z'));
      if(spreads) { … }
    }
  }
  …
}
```

### 2. `MOVE` from a holder to itself collapses its loose cards into one pile (`client/js/widgets/widget.js:2053-2073`, `client/js/widgets/holder.js:315-341`)

`applyMove()` returns `1` for the `source == target` case as well (it only flips and brings to front), so every card that was *already* sitting loose in the target holder ends up in `moved`. `groupDroppedCards()` filters only by `parent == this.get('id')` and `type == 'card'`, which those cards satisfy — so a routine like

```json
{ "func": "MOVE", "from": "tableau", "to": "tableau", "face": 0 }
```

(a common idiom for flipping/reordering in place) turns every loose card of an `allowPiles` tableau into a single pile. Same for a collection-based `MOVE` whose collection happens to contain cards already in the target.

Only cards whose parent actually changed should be grouped — e.g. skip the call when `source == target`, and in the collection branch record `c.get('parent')` before `applyMove` and drop the ones that already were the target.

### 3. New properties have no curated editing UI (standing maintainer rule)

The PR body acknowledges this and asks for a decision, but as it stands the six properties are only reachable through *Other properties → Add property*, the JSON editor buttons (auto-generated from the defaults at `jsonedit.js:1864`) and the validator — which per the maintainer rule does not count. A concrete layout that matches how the neighbours are presented today:

- `client/js/editor/sidebar/properties.js:1421-1425` (holder `behavior`): add `{ label: 'Arrange piles', property: 'allowPiles', kind: 'checkbox' }` directly under *Prevent piles* — they are each other's opposite and belong next to each other (worth noting in the hint that `preventPiles` wins, see `holder.js:345`).
- `properties.js:9639-9642`: two more `renderNumberPairRow` rows right below the existing *Stack offset* row — `Piles offset` (`pilesOffsetX/Y`) and `Piles gap` (`pilesGapX/Y`), both with `nullIfEmpty: true` so "unset" stays expressible — plus a single `Spread min` number input. Ideally only rendered/enabled while `allowPiles` is on (`addPropertyListener` is already used a few lines up for the css state buttons).
- `properties.js:9654`: add the new names to the `renderOtherPropertiesSection` exclusion list, otherwise they show up twice.
- Pile (`properties.js:1428-1433`): a *Stack offset* X/Y pair and *Spread min* in the pile `behavior` block, next to *Snap range* and *Stack cards on top of each other*.
- `PROPERTY_HINTS` (`properties.js:1300-1320`): one line for each of the six, in the same voice as the `stackOffsetX/Y` and `preventPiles` entries above.
- Also consider adding `stackOffsetX`, `stackOffsetY` and `spreadMin` to `templateProperties` (`properties.js:8261`), so "Copy every setting of this pile into &lt;deck&gt;" carries the new pile layout into `onPileCreation` — otherwise a spread pile outside an `allowPiles` holder can only be configured by hand-editing the deck template.

### 4. A small `pilesOffset` re-merges the card the holder just slotted (`client/js/widgets/holder.js:200-211`)

The comment says that after `receiveCard()` the widget "sits a whole slot away from its neighbours and could never combine with any of them" — that only holds while the slot is wider than `pileSnapRange` (10). With e.g. `pilesOffsetX: 5` the automatic `updatePiles()` that `set('parent')` runs afterwards (`widget.js:3175`) still finds the neighbouring pile within range and merges the card into it, i.e. the drop position no longer decides. Either clamp/document the interaction or suppress that second `updatePiles()` for the drop you have already resolved.

### 5. Spacing is inconsistent between a lone card and a pile in a flush row (`holder.js:264-277`)

In the "flush" case a pile child consumes `spreadExtent()` (bare card width, no gap) while a plain card child consumes `width + 4`. A tableau that mixes single cards and piles therefore has a 4 px seam in some places and none in others. Cheap fix: use the same `+ 4` for the pile branch.

### 6. Test coverage stops at the pile

`tests/client/pile-spread.test.js` is good, but `holder.js` is not importable from jest, so `childSpacing()`, `spreadDirection()`, `groupDroppedCards()` and the drag-out path — exactly where findings 1 and 2 live — have no automated coverage at all. Since the tutorial rooms now exist, a TestCafe test that deals into `Properties - allowPiles/0.json`, forms a pile, drags it out and compares the state hash would lock the feature down cheaply (`docs/testcafe.md`).

### Tutorial

`library/tutorials/Properties - allowPiles/` looks right: three variants in one directory, all three exit 0 on `validate_gamefile_node.js`, `_meta.info` is field-for-field identical in shape to `Properties - dragLimit` (including `usesAIImagery`, differing only in `name`/`image`/`variant`/`lastUpdate`), and the image is a 530×530 SVG with a `<title>` in the neighbours' style. No other tutorial is touched. Nothing to complain about here.

I added the `beta` label: the engine files it touches are only incidentally engine — every new path is gated behind the opt-in `allowPiles` (holder) and the new pile `stackOffset`/`spreadMin` properties, the `widget.js` edits are a `pileSnapPositions()` hook whose default is byte-for-byte the old check plus a `groupDroppedCards()` call that returns immediately unless `allowPiles` is set, and I verified defaults are unchanged locally (jest 1274 passed, TestCafe `holderevents.js` + `interactions.js` 54 passed with no state hash moving) — so a maintainer can allow-list it for the beta build, and the two bugs above only affect games that opt in.

VTT-AI-Agent:
Feedback from Discord, acted on in 28777d5:

> I don't seem to be able to place a card inside a pile already in a holder.
> Also the pile should be resized to match the spread

([Discord](https://discord.com/channels/770758631146782780/1536770300362887199/1536791393152925727))

**Placing a card inside a pile.** The drop only combined when the *corner* of the dragged card came within ten pixels of the corner of the pile — that is what piles use on the open table, and it is far too small a target in a holder that arranges piles, where the piles sit next to each other in a row or run down the board as a fan. In such a holder the drop now picks the widget it lands on: a card dropped anywhere on a stack, or anywhere along a fanned column, goes into that pile. A drop on free space still starts a new pile, so nothing that worked before is lost. For the same reason a pile dropped onto another pile merges with it now instead of parking next to it — otherwise a tableau column could never be moved onto another column.

**Resizing to the spread.** A pile that spreads its cards now has the box that spread takes up, instead of staying the size of one card: the handle sits at the end of the fan, the holder gives it the room it really occupies, and everything that measures a pile agrees with what you see. A pile that stops spreading — dragged out of the holder whose offsets it was borrowing — collects its cards back onto one spot and shrinks. Piles that place no cards themselves (a scattered heap of tiles, for instance) are untouched.

While regenerating the tutorial rooms for this I noticed their saved states had been written by hand and disagreed with what the engine lays out, so all three variants are now the engine's own output, and the "Spacing" and "Spreading" rooms are a bit wider for the piles they show.

`npx jest` (1278) and the TestCafe suites `holderevents`, `interactions`, `editor`, `routines`, `domsnapshot` and `publiclibrary-1..4` pass. The [test server](https://test.virtualtabletop.io/PR-3116/pr-test) has it — the tutorial is under Tutorials → *Properties - allowPiles*.

VTT-AI-Agent:
Thanks — the review landed while the second commit (28777d5) was already pushed, so finding 1 was mostly fixed before it arrived; the rest is in 82062be. Point by point:

### 1. A pile that stops spreading — fixed (28777d5, extended in 82062be)

Both repro paths pass now, and both are jest tests:

- *the author turns the spread off* → `collects its cards and shrinks back when it stops spreading`
- *the pile is dragged out of the tableau* → `collects its cards when it is taken out of the holder` (your exact snippet)

I did not take the suggested "always lay the cards out when `alignChildren` is on", because it breaks a game in the public library: **Rummy Tiles** keeps its tiles scattered *inside one pile* (each with its own x/y), and collapsing them onto the drop offset moves the `publiclibrary-2` state hash. That is exactly what my first attempt at this did, and the test caught it. So the layout runs only when the pile actually places its cards itself — `laysOutCards()`: it spreads, or it sits in a holder that arranges piles — plus a one-shot `layOut` for the transitions where it *stops* doing so (`laidOutCardsBefore()` for `parent`/`stackOffsetX`/`stackOffsetY`, and a nudge from `Holder.onPropertyChange` when `allowPiles` itself goes off, which you rightly point out `arrangedChildren()` no longer covers).

The box follows the same rule: a spread pile is now as big as its spread (handle at the end of the fan, holder measuring it correctly), and it shrinks back to card size when it collects the cards again.

### 2. `MOVE` from a holder to itself — fixed (82062be)

Correct and nasty; `applyMove()` returning `1` for the in-place flip is exactly it. Only cards whose parent actually changed are grouped now, in both the `from` and the `collection` branch (the collection branch reads the parent before the move). Covered by `MOVE from a holder to itself leaves its cards as they are` in the new TestCafe file — I checked it goes red on the old code.

### 3. Curated editing UI — deliberately still open

Not ignored: [the proposal](https://github.com/ArnoldSmith86/virtualtabletop/pull/3116#issuecomment-5255894940) is up and waiting for a human answer, which is the process I have to follow for a UI nobody specified. Your layout matches mine almost line for line, and you added three things I had missed that I will fold in when I build it: the `renderOtherPropertiesSection` exclusion list, the `PROPERTY_HINTS` entries, and `templateProperties` for the pile properties. As soon as a maintainer nods (or overrules) I will push it.

### 4. `pilesOffset` smaller than `pileSnapRange` — fixed (82062be)

Rather than clamping or documenting it: inside a holder that arranges piles, *what a drop joins is the holder's decision*, so `isPileSnapTarget()` there requires the widget to sit exactly where the target is (which is what the holder does when it resolves a drop) instead of within the snap range. A slot that happens to be 5 px from its neighbour cannot pull anything in any more. Everywhere else the range rule is untouched.

### 5. The 4 px seam — as intended, I think

`overlap` defaults to *true*, so a card child of a spreading holder consumes `stackOffset`, not `width + 4`; the `+ 4` only applies to a game that turns `overlap` off. Flush piles are meant to be exactly that — the tutorial says "placed flush, one right after the other" — and an author who wants air between them has `pilesGapX`/`pilesGapY` (`pilesGapX: 4` is precisely that seam). Making the pile branch mirror the card branch would either put a seam where the tutorial promises none or, since a pile's own `overlap` is also true by default, make the piles overlap and hide each other. Happy to revisit if a maintainer disagrees.

### 6. Holder-side coverage — added (82062be)

New `tests/testcafe/allowpiles.js`, five cases through the pointer (`dragPath`) on a tableau fixture: a card dropped onto the middle of a fanned pile joins it, a card dropped beside the piles becomes its own column, a column dropped onto another column merges, a column dragged out of the holder collects its cards and goes back to card size, and the `MOVE`-to-itself case above. It asserts the state directly rather than a hash, so a failure says which rule broke.

### Runs

`npx jest` 1279 ✓ — and TestCafe `publiclibrary-1..4`, `holderevents`, `interactions`, `routines`, `domsnapshot` and the new `allowpiles`: 87 ✓, no state hash moved.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review) of the three `Properties - allowPiles` tutorial rooms and of what the new pile layout looks like on screen. The rooms are well written and the feature reads clearly at laptop/desktop size, but the one thing the whole feature is about — *where one pile ends and the next begins* — is the one thing the demos make hardest to see: in two of the five spacing demos the green pile handle (the only cue that a group of cards is a pile) is drawn **underneath the following pile and clipped to a sliver**. Most impactful improvement: pick spacings in the tutorial that keep every pile handle fully visible, and tell the reader in words that the green number *is* the pile.

<sub>Reviewed by loading each variant into a local room on a checkout of this branch and driving it with Playwright. Viewports exercised: **1920×1080, 1280×800, 820×1180, 390×844**. Overlays opened at all four sizes: **Game Shelf / Public Library (Games + Tutorials), Players, About, Edit Mode, and the JSON editor with a widget selected**. Not reached: the states/save-and-variant dialogs, share/publish, chat and the confirm prompts — the diff touches no CSS/HTML, so overlay risk is low, but they are untested by me. Interactions driven: card→card drop in an `allowPiles` holder, `MOVE` buttons, dragging a pile between holders.</sub>

---

### Big layout improvements

**1. A pile is invisible as a pile, and the one cue it has gets covered up.**
In a spreading holder, nothing distinguishes "three cards that are one pile" from "three loose cards" except the green count handle — and where the piles overlap, the *next* pile is painted over that handle. In the `pilesOffsetX: 60` demo, two of the three counts are cut down to a 6-pixel stripe:

pilesOffsetX

The "placed flush" demo is worse, because there the handle lands squarely on the neighbouring card, so it reads as belonging to the wrong card:

flush

Concretely: in `1.json`, `holder2` (`pilesOffsetX: 60`, cards 103 wide) and `holder3` (flush) both bury the handle. Give the overlap demo an offset that clears the handle (or add a small `pilesGapX`) so the teaching example shows a readable count, and consider whether the pile handle should stack above its later siblings so an overlapped pile can still be read and grabbed at all — that matters well beyond the tutorial, since a solitaire tableau with `pilesOffsetX` is exactly the intended use.

**2. The colour code the first room teaches is contradicted by the next two.**
`0.json` uses lavender `#d8dcf0` for the `allowPiles: false` holder and green `#d8f0dc` for the `allowPiles: true` one — an excellent, wordless signal. But in `1.json`/`2.json` *every* holder has `allowPiles: true` while the colours are green / lavender / beige (`holder1`/`holder2`/`holder3`). A reader who just learned "lavender = off" now sees a lavender holder that is on. Either use green everywhere in variants 1 and 2, or reserve lavender exclusively for "allowPiles off".

**3. Let each holder be the size of the thing it is demonstrating.**
The feature's headline is "a pile is as big as its spread, and the holder gives it that much room" — but the demo holders are hand-sized to much more. `2.json` `holder2` is 250×430 around a 103×256 pile, `holder3` is 340×200 around a 169-wide pile, `1.json` `holder5` is a 500×190 empty rectangle. The result at 1920×1080 is a top half that is dense and a bottom half that is mostly empty:

spreading desktop

Trimming each holder to roughly what its piles occupy would both tighten the page and demonstrate the point.

**4. Give each demo a visible label and put its control next to it.**
Every demo is "holder + a paragraph underneath", and the captions start with the property name in bold, which works — but the buttons that drive them are placed inconsistently: a vertical stack to the right in `0.json`, bottom-left of the target holder in `1.json`, and in `2.json` "Deal 5 cards" sits at the far right of the room while its caption says it deals into the *middle* holder. Put each button directly under (or beside) the holder it acts on, and consider a one-line heading above each holder (e.g. "pilesGapX 20") so the room can be scanned without reading four paragraphs.

- **Caption blocks are offset 20 px left of their holder.** `0.json` `text1.x` 90 vs `holder1.x` 110; `1.json` `text1.x` 40 vs `holder1.x` 60; `2.json` `text1.x` 90 vs 110. Small, but it makes every caption look slightly unaligned with the box it explains. Align the left edges.
- **The pile handle overhangs the holder's top border** in all three variants (piles sit at `y: 4`, the badge is drawn above the pile's top edge), so the green square floats outside the coloured rectangle. Consistent app behaviour, but now visible on every demo — a `dropOffsetY` of ~24 in the demo holders would keep it inside.
- **`spreadMin` renders its compressed cards as four hairlines** — correct, but at a glance it reads as one card sitting on a stack edge, not "four cards squeezed": spreadMin The caption should point at them ("the four thin edges above the 8 are the compressed cards").
- **"Recall & Shuffle" wraps to two lines and overflows its 40 px button** in all three variants (`deckHolderB`) — inherited from the deck-holder template, but visible in every screenshot; a taller button or "Recall" would fix it.
- **820×1180 (tablet portrait)** scales the room down to ~9 px body text — legible but borderline; **390×844** is unusable, though the app itself shows its "rotate your device" hint there, so this is app-level, not this PR's doing. Nothing in the diff overflows, clips or becomes unreachable at any of the four sizes, and all overlays I opened closed normally.
- **Library tile** matches the family convention well (badge, "Properties" word, graduate symbol, glyph) and, unlike its neighbour, keeps the glyph fully inside the tile: tile

- **"Drag the pile handle away"** (`0.json` `text2`) — a newcomer has no idea what a pile handle is; the only thing on screen is a small green number. Say so: "drag the pile by its green number (the pile handle)". This is the single wording change with the most effect, given finding 1.
- The captions are written for someone who already speaks the engine's vocabulary — *holder, slot, flush, spread, MOVE, offset, gap* all appear undefined. The overview of `1.json` is a five-line dense paragraph that defines four properties at once; splitting the rule ("a gap wins over an offset on the same axis") down to the demo it applies to would let the reader learn it one holder at a time.
- **"Take 2 back"** is ambiguous — back to where? "Return 2 cards to the deck" says it.
- **"open the JSON editor with ctrl-j"** works and does show `"allowPiles": true` on `holder2` (verified below) — worth adding that this opens edit mode, since a first-timer pressing ctrl-j lands in an unfamiliar full-screen editor: json editor
- **Empty holders have no in-place affordance.** `0.json` `holder3` and `1.json` `holder5` are large blank coloured rectangles; their instructions live in a paragraph off to the side. A short label inside the holder ("deal cards in here") would make them read as targets.
- Titles, per-holder captions and the bold property names are otherwise clear, and the before/after pairing in the Basic room ("nothing combines" vs "the two become a pile") teaches the feature well:
  basic desktop
  spacing desktop

VTT-AI-Agent:
> Upload the test tutorials to /pr-test-ai

([Discord](https://discord.com/channels/770758631146782780/1536770300362887199/1536824484215001222))

Done — the tutorial of this PR is now in the `pr-test-ai` room of the test server as one game with all three variants:

**https://test.virtualtabletop.io/PR-3116/pr-test-ai** → *Properties - allowPiles* → **Basic** / **Spacing** / **Spreading**

It is the exact content of `library/tutorials/Properties - allowPiles/` from this branch (`0.json`, `1.json`, `2.json` plus the library image asset), packed into a single `.vtt`; the file itself is at https://agent.virtualtabletop.io/reports/pr/1536770300362887199/Properties-allowPiles.vtt if you would rather drop it into a room of your own.

Checked before uploading: added the same file to a room on a local server running this branch, and all three variants show up under one entry and load without console errors.

No code change, so the branch is unchanged by this.

VTT-AI-Agent:
Thanks — this was a useful pass over the tutorial rooms. All three variants were regenerated in [`57d1ca7`](https://github.com/ArnoldSmith86/virtualtabletop/commit/57d1ca72); no engine code changed, so nothing about existing games moved.

**Changed**

1. **The buried pile handle.** The two demos where piles overlap (`1.json` `holder2` with `pilesOffsetX 60`, and the flush `holder3`) now give their piles `handlePosition: "bottom left"`, so every count is fully visible and no badge sits on a card's rank corner. The caption says why, which turns the problem into the lesson: *"A pile that overlaps the next one would have its handle covered by it, so these three carry handlePosition bottom left to keep their count in the open."*
   On the engine side I did not move the handle above later siblings: the handle is a child element of the pile's own widget div, so it cannot paint above a *different* widget without reordering the widgets themselves, and reordering them would break the reading order of an overlapping fan. `handlePosition` is the existing answer for exactly this and needs no new code. What could be added later is letting a holder that arranges piles pass a `handlePosition` down to the piles it creates, the way it already passes `stackOffset`/`spreadMin` — happy to add that if you want it, but it is a new property behaviour and not something I want to slip in on a UI review.
2. **The colour code.** Lavender now means `allowPiles: false` and nothing else: every holder that arranges piles is green in all three variants (`1.json` `holder2`/`holder3` and `2.json` `holder2`/`holder3` changed).
3. **Controls next to what they act on.** `2.json`'s "Deal 5 cards" moved from the far right to directly under the caption of the middle holder it deals into, which also fills the empty lower-left of that room. `0.json`'s three buttons grew to 150 px so "Return 2 to deck" stops wrapping out of them.
4. **Caption alignment.** Every caption now starts at the x of the holder it explains (`0.json` 110/640, `1.json` 60/560/1060, `2.json` 110/560).
5. **`spreadMin` reads as four cards.** The caption points straight at them: *"the four below them are squeezed together - they are the four thin edges above the 8."*
6. **Vocabulary.** "pile handle" is introduced where it first appears (*"the green number that appears is the pile's handle and counts its cards"*), `MOVE` is introduced as "the routine operation that moves cards from one holder to another", "flush" is spelled out as "each one begins exactly where the one before it ends", "Take 2 back" became **"Return 2 to deck"**, and the `ctrl-j` sentence now says it switches to edit mode and that `ctrl-j` again comes back.
7. **Empty holders label themselves.** `0.json` `holder3` and `1.json` `holder5` carry an in-place italic hint ("empty - deal cards in here with the buttons on the right"), sitting under the cards so it disappears behind them as they arrive.
8. **Decks sized to the holders.** While testing the buttons I found a real problem your point 3 pointed at from the other side: clicking a deal button often enough pushed a pile past the right edge of its holder. The three decks now hold 7 / 6 / 5 cards, so no sequence of deals can overflow the holder it fills.

**Not changed, with reasons**

- **Holder sizes.** They are deliberately larger than their current contents: every one of them is a drop target that grows — `2.json` `holder2` has to fit the second column that "Deal 5 cards" adds, `1.json` `holder5` three piles at `pilesGapX 40`, `0.json` `holder3` three dealt piles. Trimming them to today's piles would make the demos overflow the moment a reader uses them; sizing the decks (point 8 above) fixes the same lopsidedness from the other end.
- **The handle overhanging the holder's top border.** That is what every pile in every holder looks like in the app, including the tutorials that already exist. Padding the demo holders with `dropOffsetY` would make these rooms show something a reader will not see in their own game.
- **"Recall & Shuffle" wrapping.** The deck corner is copied verbatim from the tutorial family (it comes from `Types - Piles`), and all of them share that button. Fixing it in this one tutorial would make it the odd one out — it deserves its own PR across the tutorials.
- **390×844 / 820×1180.** Agreed, app-level and not this PR's doing.

Rooms as they look now: [Basic](https://agent.virtualtabletop.io/reports/pr/1536770300362887199/tutorial-v2-0.png) · [Spacing](https://agent.virtualtabletop.io/reports/pr/1536770300362887199/tutorial-v2-1.png) · [Spreading](https://agent.virtualtabletop.io/reports/pr/1536770300362887199/tutorial-v2-2.png). `npm test` (1279 tests) and `validate_gamefile_node.js` on all three variants pass, and I walked every demo and every button in a browser again — dropping a card onto a pile, dragging a pile in and out, and each deal button clicked until its deck runs dry. The updated rooms are also back in [`pr-test-ai`](https://test.virtualtabletop.io/PR-3116/pr-test-ai).

VTT-AI-Agent (commented):
## TL;DR

The earlier engine regressions are fixed and the tutorial is in good shape, but this still needs three follow-ups: long spread piles can miss the pile they are visibly dropped on, the new TestCafe coverage is not wired into CI, and the new author-facing properties still lack the required curated editing UI. The first issue comes from hit-testing with the pile origin while claiming to aim with its top card.

## Findings

### 1. Dropping a long downward-spread pile onto a short pile can miss (`client/js/widgets/holder.js:77-80`)

`arrangedChildAt()` selects `child.children()[0]` as the aiming card, but then only uses that card's width and height; it never adds `aiming.get('x')` / `aiming.get('y')` to the point. `children()[0]` is the highest-z card, and `Pile.arrangeChildren()` places that card at the largest positive offset for an ordinary downward/rightward spread. Thus, for a long column dragged by its handle, the visible top card can be over a short target column while `pointY` remains near the dragged pile's bounding-box origin, above the target. The hit test returns no target and the columns stay separate, contrary to the comment and the promised “what a drop lands on is what it joins” behavior.

The existing E2E case only moves a two-card column onto a taller three-card column, so both the incorrect origin point and the intended aiming-card point remain inside the target. Please include the aiming card's relative position in the point, e.g. `x + aiming.get('x') + aiming.get('width') / 2` (and likewise for Y), and cover a long column dropped by its handle onto a one- or two-card column.

### 2. The new TestCafe suite never runs in CI (`tests/testcafe/allowpiles.js:1`, `.github/workflows/testcafe-chrome.yml:18-51`, `.github/workflows/testcafe-firefox.yml:18-44`)

Both TestCafe workflows use an explicit `matrix.testfile` list, and neither list contains `allowpiles.js`. That is why the otherwise-green check list has no `Run TestCafe Tests (allowpiles.js)` job: all five new pointer/MOVE regression tests can rot without blocking this or later PRs. Add the file to the Chrome matrix at minimum, and to Firefox too unless there is a concrete browser-specific reason not to.

### 3. The new properties still lack curated editing UI (standing maintainer rule)

The PR body acknowledges that the properties are currently available only through the generic *Other properties → Add property* path. That does not satisfy the required real editing UI support. The proposal already posted on this PR is a good concrete layout: put **Arrange piles** beside **Prevent piles** in the holder Behavior group; put conditional **Piles offset**, **Piles gap**, and **Spread minimum** controls with the existing holder offsets; and put **Stack offset** plus **Spread minimum** in the pile Behavior group. The pile properties should also participate in the pile-template controls, and all curated properties should be removed from the generic Other-properties list to avoid duplication.

The three `library/tutorials/Properties - allowPiles/` variants, metadata, and square SVG follow the requested tutorial conventions, no unrelated tutorial is changed, and `check-library` passes.

The `beta` label was already present; the edits to `holder.js`, `pile.js`, and `widget.js` are incidental for beta eligibility because their behavioral paths are gated by the new opt-in properties and preserve existing-game defaults, and the findings above do not make player testing unsafe.

VTT-AI-Agent:
Thanks — two of the three are fixed in this push, and chasing the first one turned up a real defect that neither of us had named yet.

### 1. The long-pile hit test — the premise does not hold, but it pointed at a genuine bug

I instrumented `arrangedChildAt()` and dropped a five-card fan on a tableau to see what it actually gets handed. At that moment the dragged pile looks like this:

```
{"id":"col2","x":19,"y":45,"box":[103,160],
 "cards":[["col2c4",0,0],["col2c3",0,0],["col2c2",0,0],["col2c1",0,0],["col2c0",0,0]]}
```

A pile that is being carried has already **collected its cards**: leaving the holder makes it stop laying them out, so all five sit at (0,0) and its box is one card, not a 320-unit fan. So `children()[0]` is not "the visible top card of a spread" — it is the one card the pile is drawn as, and adding `aiming.get('y')` would add zero. There is no drop that misses because of it.

The code did read as if it meant something else, though, so it now aims with **the middle of the box that was dropped** — `x + child.get('width')/2` — which for a carried pile is the card under the pointer and is the same point the surface itself hit tests to decide *which holder* the drag ended in. One rule, one point, less code.

**What the investigation did find is worth having:** a pile picked up by a handle at the far end of its fan *jumps out from under the pointer*. The drag anchor is taken at mousedown from the box the pile had (320 units tall for a five-card column); the pile then collects its cards and shrinks to 160, and the pointer goes on holding a point 315 units from the corner of a box that no longer exists. Measured in a browser with the pointer at board (500, 500):

| | carried pile box (board Y) | where the pointer is |
| --- | --- | --- |
| before | 184 – 344 | 156 units *below* the pile |
| after | 342 – 502 | inside it, at the handle end |

`rescaleDragAnchor()` (mousehandling.js) keeps the anchor at its place in the box whenever a pile changes size mid-drag, so the pointer keeps hold of what it took hold of. It only ever fires for piles that lay their cards out, i.e. only for games that opt into this feature. This is the *default* configuration for a horizontally spread pile, whose handle sits at the right end of the fan, so it was worth fixing rather than documenting.

### 2. TestCafe not in CI — added, Chrome

`allowpiles.js` is now in the Chrome matrix, next to `holderevents.js` and under the same comment: both drive their drags with the synthetic mouse events of `interaction-util.js`, whose timing against the drop animation has only ever been measured in Chrome. I left it out of the Firefox matrix because Firefox does not start in my sandbox, so I cannot tell you whether it passes there — I would rather not add a job I have not run. If you would like it tried in Firefox anyway, say so and I will add it.

### 6 (from the first review) — holder-side coverage, now in jest as well

`holder.js` is importable from jest now (a single `export` on the class; `pile.js`, `line.js` and `button.js` already do this and the bundle concatenates either way). The new `tests/client/holder-piles.test.js` has 12 cases over `arrangedChildAt()`, `childSpacing()` and `spreadDirection()` — what a drop lands on, what it does not, which pile wins where two overlap, the gap/offset/flush rules and the axis the piles are spaced out on. `allowpiles.js` gained a sixth pointer-driven case for the anchor fix above (it fails on the old code).

### 3. Curated editing UI — still open, still waiting for a human

Unchanged from the last round, and not out of stubbornness: I have to have a person agree on a UI nobody specified before I build it. [The proposal](https://github.com/ArnoldSmith86/virtualtabletop/pull/3116#issuecomment-5255894940) is up, your layout and the previous reviewer's match it almost line for line, and the three details the earlier review added — the `renderOtherPropertiesSection` exclusion list, the `PROPERTY_HINTS` entries and `templateProperties` for the pile properties — are folded into what I will push. It is a single commit's worth of work the moment a maintainer nods or overrules.

### Runs

`npx jest` — 33 suites, 1291 tests ✓. TestCafe in Chromium: `allowpiles.js` (6), `holderevents.js`, `interactions.js`, `publiclibrary-2.js` and `domsnapshot.js` ✓, no state hash moved.

VTT-AI-Agent (commented):
## TL;DR

The engine work has held up well over three rounds — every new behaviour is genuinely gated behind `allowPiles` / the new pile properties, and I could not find a path that changes an existing game's defaults. What is still outstanding is process, not arithmetic: the demo **tutorial is still committed under `library/tutorials/`**, which the revised standing rule forbids without explicit human approval (it is already on `pr-test-ai`, so the four files should simply come out — they are ~2 700 of the 2 845 added lines), and the six new author-facing properties still have **no curated editing UI**. One genuine engine edge case as well: turning `allowPiles` back off leaves orphaned piles in the holder and silently changes what `COUNT`/`dropLimit` report.

## Findings

### 1. The tutorial must not be in `library/tutorials/` (`library/tutorials/Properties - allowPiles/*`)

`docs/tutorials.md` — *"Test rooms are tutorial-quality — but they do NOT go into the PR"*, standing maintainer rule revised 2026-08-11:

> But **do not commit it to `library/tutorials/`** (or anywhere else under `library/`) unless a human has explicitly approved adding it. By default such a PR contains **no** tutorial or library changes at all; the hands-off rule above stays fully in force.

The PR body still cites the *old* wording ("Per the standing rule, the test room is a **tutorial**: `library/tutorials/Properties - allowPiles/`"). I went through every comment and review on this PR and found no human approval for adding it to the library — the only maintainer instruction on the topic is the opposite one, *"Upload the test tutorials to /pr-test-ai"*, which you already complied with. So the room has already landed where it belongs.

Concretely: drop `library/tutorials/Properties - allowPiles/0.json`, `1.json`, `2.json` and `assets/705980845_4054` from the branch, keep the `.vtt` on `pr-test-ai`, and note in the PR body that the demo lives on the test server and can be added to `library/tutorials/` on request. That also shrinks the diff from 2 845 to roughly 150 added lines, which makes the engine change reviewable on its own.

### 2. The six new properties still have no curated editing UI (`client/js/editor/sidebar/properties.js`)

Standing maintainer rule: a new widget property has to be settable from the real property editor. I checked the two claims in the PR body:

- **JSON editor** — correct, `jeAddCommands()` builds the buttons from `new Holder()` / `new Pile()` defaults (`client/js/jsonedit.js:1391`, `:1394`), so `allowPiles`, `pilesOffset*`, `pilesGap*` and `spreadMin` get their insert buttons for free.
- **Property editor** — only through the generic `renderOtherPropertiesSection()` fallback (`properties.js:7024`), which is exactly the "raw list" the rule exists to avoid. `preventPiles` (`properties.js:1423`) and `stackOffset` (`properties.js:9641`) both have curated rows; `allowPiles`, the direct opposite of `preventPiles`, has none.

Your [proposal comment](https://github.com/ArnoldSmith86/virtualtabletop/pull/3116#issuecomment-5255894940) matches what I would ask for, and review 3 independently converged on the same layout. Three reviews have now flagged this and no human has objected to the proposal in the meantime — I would build it rather than keep blocking on silence, and let the reviewer of the resulting diff move a row if they disagree. Minimum to close the rule:

- `editorTypeSections.holder` **Behavior**, directly under *Prevent piles*: `{ label: 'Arrange piles', property: 'allowPiles', kind: 'checkbox' }`.
- Two `renderNumberPairRow`s below *Stack offset*, `available` on `allowPiles`: **Piles offset** (`pilesOffsetX`/`pilesOffsetY`) and **Piles gap** (`pilesGapX`/`pilesGapY`), both `nullIfEmpty: true` with an empty placeholder so "unset" is expressible.
- **Spread minimum** (`spreadMin`, `min: 0`, `nullIfEmpty: true`) under those.
- Pile editor: *Stack offset* pair + *Spread minimum* in **Behavior** next to *Snap range*, with the inherited holder values as placeholders, plus the matching `PROPERTY_HINTS` entries and the `renderOtherPropertiesSection` exclusion lists (otherwise each property appears twice).

### 3. Turning `allowPiles` off strands piles in the holder and changes what `COUNT` sees (`client/js/widgets/holder.js:236-239`, `:48-58`)

When `allowPiles` flips to `false` the handler collects each pile's cards back onto one spot — but leaves the *piles* in the holder. A spreading holder can otherwise never contain a pile: `onChildAddAlign` (`holder.js:188-207`) empties any dropped pile out card by card, which is the whole behaviour this PR makes optional. So the flip produces a state the engine cannot otherwise reach, and `children()` changes meaning underneath the game:

- with `allowPiles: true`, `children()` flattens the piles (`holder.js:52-57`), so a tableau of 4 columns × 5 cards reports **20**;
- the instant `allowPiles` goes false, the `else if` no longer applies and `childrenFilter(..., true)` keeps the piles as entries, so the same holder reports **4**.

Failure scenario: a game with a `SET` of `allowPiles` (or an author unticking the checkbox once the curated UI from finding 2 exists) sees `COUNT` on that holder drop from 20 to 4, `dropLimit` stop constraining what it used to, and `MOVE ... from: thatHolder` move piles rather than cards. Suggested fix: in the `!newValue` branch, dissolve the piles the way the `preventPiles` path does (reparent each card to the holder, `bringToFront()`, then `receiveCard()`), instead of only laying their cards out. Niche, but it is the one way to reach a state the rest of the file assumes is impossible.

### 4. `arrangedChildAt`'s aim point is not the pointer for a pile that spreads itself (`client/js/widgets/holder.js:70-85`)

The comment states that "a pile that is carried here has collected its cards on the way, so that middle is the card under the pointer". That only holds for a pile whose spread was *inherited* — `Pile.laysOutCards()` (`pile.js:308`) stays true for a pile with its **own** `stackOffsetX/Y`, so such a pile keeps its full fan while it is dragged, and `child.get('height')/2` is then ~160 units below the pointer for a five-card column rather than the card under it.

Failure scenario: an author sets `stackOffsetY: 40` on a pile template so the piles fan on the open table too; dragging one of those piles into a tableau aims the drop at the middle of the fan, so a column dropped visibly onto column 3 can join column 4 (or start a new pile) depending on where the player grabbed it. Not a regression — nothing behaved this way before — but the comment claims an invariant the code does not enforce. Either aim with the drag anchor when there is one, or at minimum reword the comment so a later reader does not rely on the invariant.

### 5. Cards and one-card piles get different spacing in the same holder (`client/js/widgets/holder.js:282-307`)

In an `allowPiles` holder with only a `stackOffset` (flush arrangement), `childSpacing` returns `child.spreadExtent(axis)` for a pile but falls through to `child.get('height') + 4` for a loose card. A holder that ends up holding, say, three columns and one single card that was moved in on its own therefore has one 4-unit-wider slot in the middle of an otherwise flush row. Harmless visually at these sizes, but it makes the "flush" rule in the PR description true only for piles; returning `child.spreadExtent(axis)` for cards too in the `allowPiles` branch would make the rule uniform.

## Things I checked and found fine

- `applyMove()` returns `0`/`1`, so rewriting `offset += await applyMove(...)` as `if(await applyMove(...)) ++offset` in the collection branch (`widget.js:2067-2079`) is behaviour-preserving, and the `source != target` guard is the right fix for the in-place-flip case review 1 found.
- `groupDroppedCards()` ignoring `dropLimit`/`onPileCreation.dropLimit` is consistent with the documented rule at `widget.js:3710-3714` ("a MOVE or CLONE in a routine ... ignores dropLimit elsewhere too"), so it is not a new hole.
- `supportsPiles()`, `spreadsChildren()` and `childSpacing()` all reduce to the previous expressions when `allowPiles` is false; `receiveCard()`'s new `Math.sign`-based comparator is equivalent to the old `stackOffset * (...)` one.
- The unconditional `bringToFront()`-before-`set('parent')` swap in `updatePiles` (`widget.js:3759-3761`) reorders the delta but lands on the same z values; `domsnapshot` and `publiclibrary-1..4` not moving is the right evidence for that.
- `Pile.isPileSnapTarget`'s whole-spread hit box only engages for a pile with a non-zero `stackOffsetX/Y`, and that property is new on piles, so no existing save can trigger it.
- `allowpiles.js` being restricted to the Chrome matrix alongside `holderevents.js` is the right call for synthetic-drag timing.

I could not run the new jest suites here (`jest-environment-jsdom` is not installed in this sandbox), so I am taking CI's word on `tests/client/holder-piles.test.js` and `tests/client/pile-spread.test.js`.

The `beta` label was already on this PR and I left it in place: the feature is entirely opt-in behind `allowPiles`, so it cannot corrupt existing games on beta, and the outstanding items above are process and edge cases rather than anything that would make it unsafe to play with.

VTT-AI-Agent:
Thanks — findings 1, 3, 4 and 5 are all in `9fadaaac`; finding 2 is the one I am still not building. Point by point:

### 1. The tutorial is out of `library/` — done

You are right, and the four files are gone from the branch: the diff is now 154 added / 1970 removed lines, all of it engine and tests. The room itself is unchanged and stays where a human can play it — the [`pr-test-ai` room](https://test.virtualtabletop.io/PR-3116/pr-test-ai) and [Properties-allowPiles.vtt](https://agent.virtualtabletop.io/reports/pr/1536770300362887199/Properties-allowPiles.vtt) — and the PR body now says so, and that it can be added to `library/tutorials/` if someone asks for it.

### 3. `allowPiles` off no longer strands the piles — fixed

Agreed, and it was reachable both ways you describe. The `!newValue` branch now empties each pile out the way `onChildAddAlign` empties a dropped one (`Holder.emptyPileIntoSlots()`: card by card onto where the pile stood, a hundredth of a unit apart so the order survives, then `receiveCard()` lines them up), so the 4 columns × 5 cards go back to 20 loose cards and `COUNT`/`dropLimit`/`MOVE ... from:` keep counting cards.

One narrowing: this only applies where the holder actually spreads its children. A holder that spaced its piles out through `pilesOffset`/`pilesGap` alone does not spread anything once `allowPiles` is off, and an ordinary holder may hold a pile — there the piles stay and only collect their cards, as before.

Covered by jest (`empties them out onto the row…`, `keeps them where it does not spread its children out…`) and by a TestCafe case that flips the property with a `SET` routine in a live room.

### 4. `arrangedChildAt`'s aim point — fixed, not just reworded

You are right that the invariant only held for an inherited spread; a pile with its own `stackOffsetY` keeps its fan all the way into the holder, and I could reproduce your scenario. So the drop is now aimed by the drag anchor, as you suggested first.

The anchor is not reachable from `arrangedChildAt` at that moment — `mousehandling.js` deletes `mouseStatus[target.id]` synchronously *before* it calls `moveEnd()`, so anything asking the mouse state during the drop gets nothing. It is threaded through instead: `moveEnd()` keeps `localAnchor` on the widget as `dropAnchor` for as long as the drop is being resolved (next to the existing `pileUpdateFromDrag` marker), and `arrangedChildAt()` uses it, falling back to the middle of the box where nothing is holding the widget — a routine putting one down, or the drop shadow.

The regression test is `A fanned pile carried in joins what the pointer is over, not what its middle is over`: a 4-card pile with its own offset (280 units of fan) is carried by a `bottom left` handle onto a single-card column, so the handle lands on that card while the middle of the fan is above the holder entirely. It fails on the old aim point (verified by reverting just that line) and passes now.

### 5. Cards vs. one-card piles — fixed

`childSpacing()` returns `child.spreadExtent(axis)` in the flush `allowPiles` branch for any child now, so a loose card takes exactly the room a pile holding just it would take and the row is flush throughout. (Side effect worth naming: `overlap` no longer applies to a card in that branch — in such a holder `stackOffset` describes the cards *inside* the piles, so there is nothing for it to mean there.)

### 2. Curated editing UI — still waiting for a human, deliberately

I know this is the third review to raise it, and I agree with your layout — it is the one I proposed. But the standing rule I work under says to propose the UI and then wait for a *human* answer rather than invent it, and three automated reviews converging on a proposal is not that answer. If a maintainer says "build the proposal", it is maybe an hour of work and it goes into this branch; I would rather have that word than land invented UI. The PR body says the rest of the change stands on its own meanwhile.

### Testing

`npx jest` 33 suites / 1295 tests, `allowpiles.js` 8 TestCafe cases, and `domsnapshot.js`, `holderevents.js`, `interactions.js`, `publiclibrary-1..4.js` (75 cases) all pass unchanged — no state hash moved.

VTT-AI-Agent:
Closing this without merging, on a maintainer's decision:

> Close the PR (without merging). Superseded by the new one.

The successor is #3117, which takes this branch's pile/spread engine (piles arranged inside a spreading holder, the drop resolution across a pile's whole box, the pile sized to its spread and the drag anchor that survives it) and puts it behind a single `layout` property instead of the `allowPiles` / `pilesOffset` / `pilesGap` / `spreadMin` set proposed here — so the pending editing-UI question is answered there, not here.

Nothing is lost: the branch `piles-in-spreading-holders` stays on the repo for reference, and the demo room of this PR remains available as [Properties-allowPiles.vtt](https://agent.virtualtabletop.io/reports/pr/1536770300362887199/Properties-allowPiles.vtt).

Issue #225, opened 2021-02-23
Cards do not accept children


96LawDawg:
When trying to attach widgets to a card, the widget simply disappears from the board.  See demo room at [https://virtualtabletop.io/09of](https://virtualtabletop.io/09of).  The single card at the top, King of Spades, has the ID of "cardtest."  Try setting the parent of a holder, card, or another widget type to that card and the widget simply disappears.

ArnoldSmith86:
I don't have time to fix it right now but I know what's wrong and will fix it soon (unless someone else is faster):

The CSS rules that hide the inactive card faces are too broad and actually hide any child widgets. Cards also have an `overflow: hidden` that break it; that should be on the faces instead.

PR #224, opened 2021-02-21
Improve default counters for mobile devices

96LawDawg:
Default counters are almost unusable on mobile devices because the plus and minus buttons sit within the overly large label text area.  This change shrinks the size of the text area box and slightly repositions the plus and minus buttons in relationship to the new box.  Values -999 to 999 still fit.  See screenshot for example of change; new default counter is on top and the original one is on the bottom.  The other change was needed to slightly reposition the default counter in the add widget menu to account for the change in size.

Counter

PR #3118, opened 2026-08-12
Let the icon pickers choose the skin tone of an emoji

VTT-AI-Agent:
## What this does

The icon list (`assets/fonts/symbols.json`) holds exactly one entry per emoji, so the ~2000 skin
tone files that ship with our noto-emoji artwork were unreachable from the UI: every person, hand
and couple could only be picked in its default yellow, and the only way to a toned one was typing
the raw character into the JSON editor.

Both icon pickers — the fullscreen **Pick icon** overlay and the icon chips in the editor sidebar —
now mark an emoji that has toned forms with a small blue corner triangle. **Clicking a marked icon
opens a flyout with that emoji's complete set and does nothing else**: it does not pick the icon and
does not close the picker. Clicking a tone in the flyout is what picks one.

* a row of five tones for a single person (👍 → 👍🏻 👍🏼 👍🏽 👍🏾 👍🏿),
* the full 5×5 matrix where two people can be toned separately — the couples, the kiss, people
  holding hands and the handshake, including the mixed-tone forms Unicode encodes as a completely
  different sequence (🤝 is `1F91D`+tone while both hands match, but `1FAF1`+tone ZWJ `1FAF2`+tone
  once they differ),
* the untoned form as a "No tone" cell, so the flyout is also how the plain yellow emoji is picked,
  and how a tone is taken back off again.

One click opens the chooser, one click picks — which is the same two taps on a touchscreen, where
the first version of this (hover on the desktop, long-press on touch) did not work:

> this doesn't really work on mobile. simplify it to clicking once to get the skin tone chooser
> (doing nothing else), then clicking the skin tone to select it. doesn't change the inlining if
> less than X - that should behave like any other icon: just click to use
>
> — @arnoldsmith86

Keyboard operation is deliberately out of scope here: the icon pickers have never had a tab stop
of their own (the grid icons are `<i>`, the sidebar chips `<div>`s), and giving one to just the ~330
emoji that have tones would scatter a handful of tab stops through a 13361-icon grid.
@arnoldsmith86 decided that the whole icon picker should be made keyboard-operable in a PR of its
own instead.

An icon that *already* carries a tone opens the same flyout with its own cell marked, so a tone can
be changed as well as chosen. The grid itself is untouched: no icon was added to it, nothing is
preselected, and there is no hidden per-user tone preference — this is
[concept B of the icon picker study](https://agent.virtualtabletop.io/reports/icon-picker-variants/index.html).
Material Symbols already have separate outlined and filled entries in the list and deliberately get
no flyout.

**Once a search has narrowed the "Pick icon" overlay down far enough, the forms are shown in the
list itself instead**, right behind the icon they belong to — searching `thumbs` lists the three
thumb emoji and their five tones each, `farmer` the three farmers with theirs, `handshake` the
whole 5×5 matrix. The limit is 50 icons in the result, which is the largest set a single emoji has (25) twice over, so
the one icon a search is often meant to find always shows what it offers. There every icon of the
list behaves like every other one — a single click picks it, the icon the forms belong to included:
the corner marker and the flyout stand aside, because everything they lead to is already on screen.
This is what @arnoldsmith86 asked for on Discord:

> Show the variants directly in the picker if all matching icons fit inside the flyout or if it
> results in <50 or whatever makes sense in the big popup.
>
> — [Discord](https://discord.com/channels/770758631146782780/1537078084648566824/1537380484210761758)

**The inline picker of the Edit Widgets panel does the same**, and so does the icon picker of a
deck editor property row, which is the same control: searching `thumbs` there lists the three thumb
emoji and their five tones each as chips of the result, `handshake` the whole matrix. Its limit
is the number of icons that picker shows at all (100) — the forms count towards it, so a result that has them is
never longer than one that has not, and a search already cut off at 100 keeps them behind the
flyout, because it has more to show than the tones of what happens to fit. The suggestions the
picker opens with are not a search and stay as they are; their chips, and those of "Used in this
game", keep the corner marker and the flyout. This is the follow-up @arnoldsmith86 asked for on
Discord:

> Also do it in the inline Edit Widgets panel picker (and deck editor which should use the same
> control). But only if the count will stay below the shown limit.
>
> — [Discord](https://discord.com/channels/770758631146782780/1537078084648566824/1537389801165291551)

## How it works

* `client/js/emojivariants.js` (new) works out an emoji's variants by inserting tone modifiers into
  its sequence and keeping the forms whose artwork exists, plus a six-entry table for the mixed-tone
  sequences that cannot be produced that way. It also renders and places the flyout.
* Which files exist is a property of the asset directory, so the server lists them
  (`server/emojivariants.mjs`, read at startup, handed out by `GET /api/emojiVariants`, ~45 kB)
  instead of the client carrying a second copy of the list that would go stale the next time the
  emoji are updated (e.g. #3109).
* `client/js/symbols.js` hangs the flyout on the overlay grid (for picking and for richtext
  insertion), `IconInput.decorateChip` in `client/js/editor/propertyInputs.js` on the sidebar chips.
* `expandEmojiVariants` / `collapseEmojiVariants` put the forms into a container and take them back
  out again. The picker's search handler calls them around its own filtering, because a form is an
  icon of the list like any other: it is a copy of the icon it belongs to, so the delegated click
  handling, the search and the 64px previews of a narrow result treat it as one of them, and the
  count the handler bases `fewResults` on has to be taken before they go in. All of them or none —
  a list where some of the emoji show their tones and others only hint at them says nothing about
  which is which.

The pickers hand a whole container to `enableEmojiVariantFlyouts` rather than decorating each icon:
one click handler serves the ~1600 emoji of the grid, and each emoji's flyout is built once and
reused. That handler takes the click in the capture phase, which is what keeps it from the picker
underneath — the overlay grid picks in one delegated handler on the same container and a sidebar
chip in one of its own, and both sit further along the event path. The picker's own click handling
is delegated in the same place, so opening it no longer assigns a closure to each of its 13361
icons.

A flyout is dismissed by clicking its icon again, by clicking anywhere else, by Escape (which the
picker behind it does not see), or by scrolling the list it hangs off — a nudge of a few pixels,
which is what a tap on a touchscreen does to the list underneath it, does not count.

## Screenshots

| | |
|---|---|
| [Corner markers in the grid](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/icon-picker-variant-markers.png) | [Tone row in the Pick icon overlay](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/icon-picker-skin-tone-flyout.png) |
| [Tone × tone matrix (handshake)](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/icon-picker-tone-matrix.png) | [Editor sidebar chips](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/editor-sidebar-skin-tone-flyout.png) |
| [Dark mode, with the tone in use marked](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/editor-sidebar-dark-mode.png) | |
| [`thumbs`: both thumbs and their tones in the list](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-thumbs.png) | [`farmer`: three farmers with theirs](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-farmer.png) |
| [`handshake`: the whole 5×5 matrix in the list](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-handshake.png) | |
| [`thumbs` in the Edit Widgets panel picker](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-sidebar-thumbs.png) | [`handshake` there](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-sidebar-matrix.png) |
| [`woman`, too many to expand: markers and flyout](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-sidebar-wide.png) | [the same picker in a deck editor property row](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-deckeditor-thumbs.png) |

Driven by taps on an emulated touch device, one tap per step:
[tap a marked chip](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/mobile-3118/02-sidebar-flyout.png) ·
[tap the tone](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/mobile-3118/03-sidebar-picked.png) ·
[tap a marked icon of the overlay](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/mobile-3118/04-overlay-flyout.png) ·
[a narrow search, where a form is picked by one tap like any other icon](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/mobile-3118/06-overlay-inline.png)

The flyout is placed beside the icon it belongs to whenever there is room (so it does not cover the
rows of the grid being scanned), clamped into the viewport and scrollable if it still does not fit,
light in the fullscreen picker and themed in the editor sidebar. Its cells are buttons labelled
"Light + Med-dark skin tone", the corner marker is explained by the picker's help text and by the
tooltip of every marked icon, and the 5x5 matrix is on a fixed 44px pitch:
[dark mode](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/ui-3118/01-darkmode-overlay-flyout.png) ·
[matrix](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/ui-3118/02-matrix-1280x800.png) ·
[matrix at 844x390](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/ui-3118/03-matrix-844x390.png) ·
[sidebar](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/ui-3118/04-sidebar-light.png) ·
[sidebar, dark](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/ui-3118/05-sidebar-dark.png) ·
[the marker](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/ui-3118/06-corner-marker-3x.png) ·
[the deck editor's picker in dark mode](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/ui-3118/09-deckeditor-flyout-light.png)

This is an editing-UI affordance rather than something a game can contain, so there is no demo room
to load: the screenshots above show it in both pickers. No tutorial and no library game is touched.

## Testing

* New jest tests (`tests/client/emoji-variants.test.js`) check the variant maths against the real
  asset directory — the tone row, the 5×5 matrices, the mixed-tone sequences, that a family emoji
  gets nothing because it has no toned artwork, and that every offered form exists as a file — and
  the flyout itself: that a click on a marked icon opens it while the picker underneath sees
  nothing of that click, that a click on an unmarked icon still reaches the picker, that clicking
  the icon again or anywhere else closes it, that Escape closes it without the picker behind it
  seeing the key, that clicking a cell reports the toned emoji, and that handing the same container
  over again replaces its handler instead of stacking a second one on it. The file set they check
  against comes from the server's own listing function, so the two cannot drift apart.
* More jest tests in the same file for the forms shown in the list: that they follow the icon they
  belong to in tone order, that a 5x5 matrix goes in as all 25 of them, that a result over the
  budget keeps every one of them out rather than expanding part of the list, that expanding twice
  does not stack them, that collapsing leaves the list as it was, and that an icon whose forms are
  in the list opens no flyout until they are gone again.
* `npm test`: 1521 tests pass.
* Three TestCafe cases, because which searches are narrow enough is decided against the real
  13361-icon list, and where the flyout is put and which colours it takes are decided against the
  real page — jsdom has neither. They cover the colours of the flyout in the dark sidebar and on
  the always-light overlay the deck editor parks inside `#editor`, that it lands inside the
  viewport, `thumbs` (15 forms in the list, no corner marker), `hand` and `woman` (no forms, marker
  back, and a click on a marked icon opens the flyout instead of picking the icon), and picking a
  form out of the list with a single click. `editor.js` is 86 tests and passes locally against
  chromium.
* Walked through on an emulated touch device (Chromium, `hasTouch`), one tap per step: a tap on a
  marked chip of the sidebar opens the chooser and leaves the widget's icon alone, a tap on a tone
  writes `👍🏽` to the widget, a tap on a marked icon of the "Pick icon" overlay opens the chooser
  with the overlay still open, a tap on a cell picks `👋🏽`, and in a narrow search a tap on a form
  in the list picks `✌🏿` straight away.
* Profiled in a real browser with a 4x CPU throttle: a keystroke in the picker's search field costs
  the same as with no flyout code at all, and taking the forms back out of the list costs 0.22ms
  of it.
* Checked on a 900x300 viewport, where the 5x5 matrix does not fit and the flyout scrolls itself:
  the box stays inside the viewport, scrolling it does not dismiss it, its bottom row is reachable
  and pickable, and scrolling the grid behind it still does close it.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3118/pr-test (or any other room on that server)


VTT-AI-Agent:
Addressed the CodeQL finding (`js/stored-xss`, alert 224) in ffa4d98e.

The alert traced the file names read from `assets/noto-emoji/` into `res.send()`, which express may answer with an HTML content type. It was not exploitable — only names matching `^emoji_u[0-9a-f_]*1f3f[b-f][0-9a-f_]*\.svg$` got through, so nothing but hex digits and underscores could ever reach the response, and an array is sent as JSON — but the flow is easy to remove, so I did rather than dismiss it:

* the route now answers with `res.json()`, which is what it always meant;
* no file name is passed on any more: each name is parsed into its code points and the sequence handed out is rebuilt from those numbers (`parseInt(codePoint, 16).toString(16).padStart(4, '0')`), which is exactly the naming the client computes for an emoji, so whatever ends up in that directory cannot put anything but hex digits into the response.

The list it returns is byte-for-byte the same (1895 entries, verified against the previous implementation), and the pickers still behave identically in the browser.

VTT-AI-Agent:
Addressed the second CodeQL finding (`js/missing-rate-limiting`, alert 225) in 0e572cba.

The route listed `assets/noto-emoji/` the first time it was asked and cached the result, which is what made it "a route handler that performs a file system access". That directory is checked into the repository and cannot change while the server runs, so it is now read **once at startup** — like `shares.json` and `widgets.json` above it — and the handler is a plain `res.json()` of an array that is already in memory. No file system access is left in it, and the first person to open an icon picker no longer pays for the directory listing.

For the record, this is the same finding the repository has had open on `/assets/:name` since 2020 (alert 2) and VTT has no rate limiter anywhere, so I did not want to pull in a dependency for one route — moving the read out of the handler makes the point moot instead.

VTT-AI-Agent (commented):
## TL;DR

Nice, well-scoped feature — the variant maths in `emojivariants.js` is clean, the server-side "ask the asset directory" approach is the right call, and the jest tests cover the interesting cases. The main thing to fix is the flyout's dismissal logic: **moving the pointer off a marked icon never closes the flyout** (`element.onmouseleave` only cancels the open timer), so it stays parked over the picker until a click, scroll or Escape. Escape and the touch path also don't quite do what their comments claim.

## Findings

### 1. The flyout is sticky: leaving the icon never closes it — `client/js/emojivariants.js:243`

```js
element.onmouseenter = _=>openTimer = setTimeout(open, 250);
element.onmouseleave = _=>clearTimeout(openTimer);
```

`onmouseleave` only cancels a *pending* open. Once `open()` has fired there is nothing on the anchor that schedules a close — only `dom.onmouseleave` (line 219) does, and that fires only if the pointer actually entered the flyout.

Failure scenario: in the "Pick icon" overlay, hover 👍 for 300 ms, then move the pointer straight up/left (away from the flyout, which is placed *below* the icon) and keep browsing the grid. The flyout stays open, covering ~200 px of the grid, and is only dismissed by a click, a scroll, or Escape. Hovering a *different* marked icon replaces it, but hovering an unmarked one (the overwhelming majority) does not.

The fix is the standard hover-flyout pattern: give the anchor the same 300 ms grace close the flyout already has, e.g. expose a module-level `scheduleEmojiVariantFlyoutClose()` next to `closeEmojiVariantFlyout()` that sets the active flyout's `closeTimer`, and call it from `element.onmouseleave` alongside `clearTimeout(openTimer)`. `dom.onmouseenter` already cancels it, so the gap between icon and flyout stays crossable.

### 2. Escape does not "close the flyout instead of the picker behind it" — `client/js/emojivariants.js:211-217`

The comment claims the capture-phase keydown listener protects the picker, but `stopPropagation()` on **keydown** cannot stop either of the handlers that actually close the picker:

* the fullscreen overlay is closed from `window.onkeyup` (`client/js/main.js:947`, `overlayActive` branch) — a different event entirely, untouched by a keydown listener;
* in edit mode, `InlinePopup.onKeyDown` is *also* a capture listener on `document` (`client/js/editor/controls/popup.js:340`) and was registered first; `stopPropagation()` does not stop other listeners on the *same* node, so the icon picker popout closes too.

So Escape with a flyout open tears down the flyout **and** the picker underneath in both places. If that isn't wanted, the precedent to copy is `popupHandledEscape` in `popup.js:6-12` — a flag plus a capture-phase **keyup** swallow. (If it *is* acceptable, the comment should say so instead of the opposite.)

### 3. Touch: the click-swallower can stay armed and eat an unrelated click — `client/js/emojivariants.js:244-251`

```js
element.ontouchstart = _=>openTimer = setTimeout(_=>{
  open();
  document.addEventListener('click', e=>{ e.preventDefault(); e.stopPropagation(); }, { capture: true, once: true });
}, 500);
```

The listener is an anonymous function registered with `once: true` and no way to remove it, and it is armed on the *assumption* that a click follows. When the long-press does not end in a click — finger slid off the icon before lifting, the touch cancelled, the browser's own long-press/context gesture taking over, or the grid scrolled (which `window`'s capture `scroll` listener uses to `close()` the flyout) — it stays armed and swallows the user's **next** click anywhere in the document. Concretely: long-press 🤝 in the picker, drag the finger off, lift; then tap a tone cell in the flyout — that tap is eaten and nothing is picked, the second tap works.

Hoisting the handler into a named `const` and disarming it (`removeEventListener` in `ontouchend`/`ontouchcancel` after a `setTimeout(0)`, or a short fallback timer) makes this deterministic.

### 4. Two new high-severity CodeQL alerts on `server.mjs`

The advanced-security bot filed [#224 `js/stored-xss`](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/224) (server.mjs:170) and [#225 `js/missing-rate-limiting`](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/225) (server.mjs:168-175), both `security_severity_level: high`, neither present on `main`.

Reading the code, both look like false positives: the sequence handed out is rebuilt from `parseInt(cp, 16).toString(16)` after a strict `^emoji_u[0-9a-f…]\.svg$` match, so only hex digits reach `res.json`, and the `readdirSync` runs once at startup, not per request. CodeQL just can't see either fact — for #225 it is presumably the `readdirSync` sitting lexically inside the same `MinifyHTML().then()` closure as the `router.get`. Hoisting the directory read to module scope (or into a tiny helper in `server/`) would likely clear #225 on its own; #224 probably needs a dismissal. Worth doing something about them either way, since new high alerts are what gates this out of the beta build (see the label note at the end).

### 5. Test coverage

The jest tests are good on the variant maths, but two things aren't covered:

* **The server route.** `tests/client/emoji-variants.test.js:14-17` builds `available` from the asset directory with its *own* regex (`^emoji_u[0-9a-f_]*1f3f[b-f][0-9a-f_]*\.svg$` + `file.slice(7, -4)`), which is a second implementation of what `server.mjs:167-170` does. They agree today — I checked, both produce exactly 1895 entries with no difference either way — but a future divergence (a filename shape the server's `[0-9a-f]{4,5}` rejects, say) would silently drop variants in the UI with all tests still green. Importing the server's list, or at least asserting the two agree, would close that.
* **The flyout itself.** Nothing exercises `addEmojiVariantFlyout` / `openEmojiVariantFlyout` — not the marker class, not dismissal, not that picking a tone resolves to the toned symbol and URL. `tests/testcafe/editor.js` already has symbol-picker tests; one that hovers a marked chip and clicks a tone cell would cover the whole path and would have caught #1 above.

### 6. Smaller things

* **Dark mode.** `.emojiVariantFlyout` and friends hard-code `background: white`, `#ddd`, `#eef4fb`, `#55606d` (`client/css/fonts.css:597-716`). The dark-mode screenshot shows a white card floating on the dark sidebar; it reads fine, but the editor's own popups do theme themselves (`body.edit.darkMode .inline-popup …`, `client/css/editor/controls/popup.css:322+`), so this is the one control in edit mode that doesn't follow.
* **Route namespace.** `/emojiVariants` sits at the root, i.e. in the same namespace as `router.get('/:room', gameRoomHandler)` (`server.mjs:472`) — a room called `emojiVariants` becomes unreachable. The existing convention for data endpoints is `/api/…` (`/api/widgets`, `/api/library/decks`); `/api/emojiVariants` would avoid it entirely and cost nothing.
* **Label race.** `iconDisplayName` (`client/js/editor/propertyInputs.js:380-383`) returns `''` until `loadIconSearchIndex()` resolves. It's kicked off when the picker opens, so it usually wins the 250 ms hover delay, but on a cold cache the first flyout can say just "SKIN TONE" with no icon name. Harmless, just noting it since the callback plumbing exists specifically for this.

### What I checked and found fine

* `fetch('emojiVariants')` and `` are relative, which is correct here — `client/room.html:8` sets `<base href="<urlPrefix>/">`, same as the existing `fetch('assetcheck')` / `fetch('i/fonts/symbols.json')`.
* `html()` is not imported into the new module but resolves fine: `compressJS` concatenates the room bundle into one scope and `domhelpers.js` comes first (`server/minify.mjs:118-120, 300-313`), and the new file is registered there.
* The refactor of `pickSymbol`/`addRichtextControls` into `pick(icon, symbol)` / `insert(icon, symbol)` is behaviour-preserving for the base case: for `emoji-color` entries `data-symbol` and `innerText` are the same string (`client/js/symbols.js:109-113`).
* Toned and mixed-tone sequences survive the rest of the pipeline: `getIconDetails` falls through to `emoji_u${emojiToFilename(icon)}.svg` for anything non-ASCII (`client/js/symbols.js:307`), and the `emojis2images` regex handles `base + tone (+ ZWJ base + tone)*` via its trailing `*` group, so richtext insertion of e.g. `🫱🏻‍🫲🏿` resolves to a real file.
* Chips are rebuilt from scratch on every render (`showResults` clears `results.innerHTML`), so the never-removed `hasEmojiVariants` class can't go stale on a reused element.
* Per the maintainer rules: this adds no widget property and no routine function, so there's nothing that needs editor/visual-block UI support beyond the UI this PR *is*, and nothing under `library/` or `library/tutorials/` is touched — I agree a demo room isn't the right vehicle for an edit-mode affordance, and the screenshots do the job.

---

I did **not** add the `beta` label: the PR touches no engine file (nothing under `client/js/widgets/`, no `compute.js`, no `serverstate.js`) so it is eligible in principle, but it introduces two new high-severity CodeQL alerts (`js/stored-xss` #224 and `js/missing-rate-limiting` #225 on `server.mjs`), and new high-severity alerts are the one bright line for beta readiness — clear or dismiss them and this becomes a good beta candidate.

VTT-AI-Agent:
Thanks — all three behaviour findings were real and are fixed in 90d4c573, together with the dark
mode and route-namespace points. Each of them now has a jest test that fails against the old code.

### 1. Sticky flyout — fixed

`element.onmouseleave` schedules the same 300 ms close the flyout itself has, and
`element.onmouseenter`/`dom.onmouseenter` cancel it, so the gap between icon and flyout stays
crossable in both directions. The active flyout is now an object (`{ anchor, close, scheduleClose,
cancelClose }`) instead of a bare close function, which is what lets an icon address the flyout that
belongs to *it* — hovering a different marked icon still replaces it, and re-entering the same one
does not tear it down and rebuild it.

### 2. Escape — fixed, and the comment was indeed wrong

You are right on both counts, and the fix is the pair of them:

* the keydown listener moved from `document` to **`window`**, whose capture phase runs before every
  listener on `document` — that is what actually keeps `InlinePopup.onKeyDown` from firing,
  registration order notwithstanding;
* the matching **keyup** is swallowed by a module-level capture listener guarded by an
  `emojiVariantHandledEscape` flag — exactly the `popupHandledEscape` pattern from `popup.js` — so
  neither `window.onkeyup` in `main.js` nor the editor's sidebar-module close sees it.

Verified in a real browser in both pickers: with a flyout open, Escape closes only the flyout; the
fullscreen picker stays open, and in edit mode the sidebar module and edit mode itself stay put
(before this, Escape over a sidebar chip flyout closed the whole properties module along with it).
With no flyout open the picker gets its Escape back, which the test asserts too.

### 3. Touch click-swallower — fixed

It is a named listener now, and it only swallows a click **on the icon it was armed for**
(`element.contains(e.target)`); anything else passes through. Either way it disarms itself — on the
first click, or after a 1 s fallback timer when the press ended without one. So the scenario you
described (long-press, drag off, lift, tap a tone cell) picks on the first tap.

### 4. CodeQL — both already cleared

They are gone as of the current head: `refs/pull/3118/merge` lists **224 `js/stored-xss` and 225
`js/missing-rate-limiting` as `fixed`**, not open —

```
gh api "repos/ArnoldSmith86/virtualtabletop/code-scanning/alerts?ref=refs/pull/3118/merge&state=fixed"
225 js/missing-rate-limiting
224 js/stored-xss
```

`ffa4d98e` (rebuild the sequence from parsed code points, answer with `res.json`) cleared 224 and
`0e572cba` (read the directory once at startup) cleared 225; your snapshot predates the analysis of
those two commits. Nothing is dismissed — both are actually fixed. This commit keeps them that way
and goes one step further along the line you suggested: the listing now lives in
`server/emojivariants.mjs` at module scope, so the handler is a one-line `res.json` of an array that
is already in memory.

### 5. Test coverage — both gaps closed

* The jest tests no longer re-implement the server's regex: they call `readEmojiVariants()` from
  `server/emojivariants.mjs`, i.e. the exact list the route hands out, and additionally assert that
  every entry is shaped the way `emojiToFilename` names a form. A divergence is now impossible
  rather than merely unlikely.
* Seven new tests cover the flyout itself in jsdom (marker present/absent, hover opens, leaving
  closes after the grace, crossing into it keeps it open, Escape closes it while a listener standing
  in for the picker sees neither the keydown nor the keyup, the touch swallower, and that clicking a
  cell reports `👍🏽` and closes the flyout). I checked each of the three fixes by reverting it and
  watching the matching test go red.

I did not add a TestCafe case on top: the picker path there would hinge on a real 250 ms hover, and
the jsdom tests already assert the same steps deterministically. The browser walkthrough is scripted
instead — 17 checks in both pickers, all passing, including the ones this review prompted.

### 6. Smaller things

* **Dark mode** — fixed. The flyout uses the editor's variables (`--backgroundColor`, `--textColor`,
  `--textHelpColor`, `--backgroundHighlightColor1`) with the light values as fallbacks, since
  `editor/layout.css` only ships with edit mode while `fonts.css` is part of the room bundle; the
  blue tint of the "Default" cell and the hover state has a dark counterpart. Screenshot:
  [themed flyout on the dark sidebar](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/editor-sidebar-dark-mode.png).
* **Route namespace** — fixed, it is `GET /api/emojiVariants` now, next to `/api/widgets`.
* **Label race** — left as is. The flyout title degrades to plain "SKIN TONE" for the fraction of a
  second before the index resolves, which is the reason the name is passed as a callback in the
  first place; blocking the flyout on the index would be a worse trade for a label.

`npm test` 1280 pass, TestCafe `editor.js` 45 pass locally against chromium.

VTT-AI-Agent (commented):
## TL;DR

The skin-tone flyout is a genuinely good addition and it looks like it belongs to the app in the sidebar picker at desktop sizes. Two things I'd fix before merge: **in dark edit mode the flyout renders dark on top of the always-light "Pick icon" overlay** (the light fallbacks in `fonts.css` never apply, because `--backgroundColor` is declared on plain `body` in `editor/layout.css`), and **the flyout can be placed half off-screen** — at 844×390 the bottom two tone rows of the 5×5 matrix are simply unreachable. Beyond that, the single most impactful improvement is to open the flyout from the **corner triangle** rather than from anywhere on the icon: at the moment a casual sweep across "People & Body" keeps popping a panel that hides ~30 of the icons you are scanning.

_This is the UI/UX review of the automatic review cycle for this PR, performed by claude-opus-5._

**This is a UI/UX review, not a code review** — layout, visual design, and how understandable the new affordance is to someone seeing it for the first time. Correctness/logic is covered by a separate reviewer.

### What I actually exercised

Local server on this branch (`90d4c573`), Chromium via Playwright.

* **Edit mode, widget properties sidebar → Icon picker popout** and its chips, at **1280×800, 1920×1080, 820×1180 and 844×390** (phone landscape), light **and** dark mode.
* **Fullscreen "Pick icon" overlay** (via *Show all*) at the same four sizes, both the tone row (`👍`), the 5×5 matrix (`🤝`, couples) and an already-toned icon (`👍🏽`, the "change the tone" flow).
* **Touch long-press** at 844×390 with a touch-enabled context — the 500 ms press opens the flyout, tapping a cell writes `👍🏽` to the widget, and the press does not leak a click. That part works.
* Non-edit overlays as a regression check: **Game Shelf, Players, About** at 390×844, 820×1180, 1280×800, 1920×1080 — unchanged, as expected (the diff only adds new class names to `fonts.css` and touches no existing selector).
* **Could not reach:** the whole editor UI at **390×844 portrait** — edit mode shows "Please rotate your device to landscape", so the flyout is unreachable there; and the **richtext symbol picker** (game-description editing), which I reviewed statically from `client/js/symbols.js:255+`. Add-widget / JSON editor / save dialogs were not exercised.
* The PR's `pr-test-ai` room is up and its shelf is empty — no draft content to review.

---

## Big layout improvements

### 1. Open the flyout from the triangle, not from the whole icon

`client/js/emojivariants.js:295` starts a 250 ms timer on `mouseenter` of the **entire icon**. In the dense grid every person/hand emoji is marked (326 of them), so moving the pointer across "People & Body" — which is exactly what the picker's own help text tells you to do ("hover one … to see the real image") — repeatedly drops a 300×90 or 290×266 panel over the icons you are browsing:

flyout covering grid neighbours

That single flyout hides roughly four rows × ten columns of the grid. Suggestions, in order of preference:

* bind `mouseenter` to the corner marker instead of the icon (it needs `pointer-events: auto` and a slightly larger hit area than the current 7 px triangle — a ~14 px invisible corner box works well and makes the triangle a real affordance rather than decoration);
* or keep the icon as the trigger but raise the dwell to ~450–500 ms, so a sweep does not trigger it;
* and place the flyout **beside** the anchor (left/right) rather than directly below it when there is horizontal room, so the row you are scanning stays visible.

### 2. Dark mode: the flyout is dark on top of a light overlay

`client/css/fonts.css:621` reads `background: var(--backgroundColor, white)`, and the comment says the light fallback covers "the fullscreen picker – always light". It doesn't: `client/css/editor/layout.css:2` declares `--backgroundColor` on plain `body`, and that stylesheet is loaded as soon as edit mode is entered. The fullscreen picker's flyout is appended to `$('body')` (`emojivariants.js:203`), so with dark mode on it computes to `rgb(8, 9, 10)` on white:

dark flyout on the light Pick icon overlay

Concrete fix: make the light values the *rule* and scope the themed ones to the sidebar, which is the only place they are wanted:

```css
.emojiVariantFlyout { background: white; color: black; }
#editor .emojiVariantFlyout { background: var(--backgroundColor); color: var(--textColor); }
body.edit.darkMode #editor .emojiVariantFlyout { --emojiVariantHighlightColor: #1c2c3d; }
```

(The same applies to `--backgroundHighlightColor1` on `.emojiVariantCell` and `--textHelpColor` on the title/labels.)

### 3. The flyout can be clipped off the bottom of the screen

`emojivariants.js:214-216` picks *above* only when the box fully fits above, otherwise *below* — with no clamping and no `max-height`. On a phone in landscape (844×390) with the anchor chip near the middle, neither fits, so it goes below and 142 px of a 266 px matrix are cut off. **The Med-dark and Dark rows cannot be picked at all**:

matrix clipped at 844x390

Measured: anchor at y≈228, flyout placed at y=266, height 266, viewport height 390. Fix is small — clamp the top the way the left is already clamped, and let the box shrink:

```js
const top = above ? anchor.top - box.height - 4 : anchor.bottom + 4;
dom.style.top = `${Math.max(4, Math.min(top, window.innerHeight - box.height - 4))}px`;
```

plus `max-height: calc(100vh - 8px); overflow-y: auto;` next to the existing `max-width` at `fonts.css:632`.

### 4. Give the matrix a real grid rhythm and a readable header

The 5×5 matrix is the part a first-time user has to *decode*, and right now the header row reads as one run-on line of grey capitals and the columns are visibly ragged:

matrix, zoomed

`grid-template-columns: auto repeat(5, auto)` (`fonts.css:661`) sizes each tone column to its own header, so the measured widths are **36 / 48 / 37 / 48 / 36 px** — the artwork sits on an uneven pitch. `repeat(5, 44px)` (or `minmax(40px, 1fr)`) fixes the rhythm *and* brings the cells up to a comfortable touch size. Together with that I'd:

* shorten the column headers to `L · ML · M · MD · D` (or 1–5) with the long names kept in the row headers only, so the header line stops running together;
* consider a 1 px separator or alternating column tint, so a cell in the middle of the block is mappable to its pair without counting.

---

## Nitpicks

* **Labels are 9 px** (`fonts.css:707` `.emojiVariantLabel`, `fonts.css:713` `.emojiVariantHeader`) in `--textHelpColor`. That is the smallest text anywhere in the editor; 10–11 px would match the rest of the sidebar and still fit.
* **The two flyout forms are sized differently**: row cells measure 52×47 px (30 px art), matrix cells 36×34 px (26 px art, `fonts.css:690`). Same control, two scales — and 36×34 is under the ~44 px target for the long-press users this feature explicitly supports.
* **Matrix cells have no usable tooltip.** `emojivariants.js:204` passes `label = null`, so `cell.title` (`:173`) becomes the bare emoji — hovering a cell shows `🤝🏻`, which tells you nothing. `` `${row} + ${column}` `` ("Light + Med-dark") would make the matrix self-explaining even without the headers.
* **"Default" out-shouts the actual selection.** `.emojiVariantBase` gets a filled tint (`fonts.css:696`) while the tone actually in use gets a 1 px border + inset (`fonts.css:702`). In the change-the-tone flow the tinted yellow cell reads as "current" and the real current cell is easy to miss — swap the weights (solid highlight for `.emojiVariantSelected`, a dashed/neutral outline for the base cell):
  change tone flow
* **The corner marker sits on top of the artwork.** At 0.55 opacity over a 30 px emoji it merges into the glyph on several icons (👇, ✊, 🖕 below) and reads as part of the drawing rather than as UI:
  markers, 3x
  A 1 px light outline (`border-left`/`border-bottom` on a slightly larger transparent wedge, or a `filter: drop-shadow(0 0 1px #fff)`) would separate it from the emoji, and full opacity would make it read as a control.
* **The flyout covers the picker's own search box** in the sidebar popout (it opens upward over the input and the "Libraries:" checkboxes), which is disorienting right after typing:
  tone row in the sidebar
* **At 820×1180 the flyout leaves the sidebar entirely** and floats over the tabletop, with a visible gap and no tail pointing back at the chip it belongs to — it reads as an unrelated panel. A small arrow/tail, or clamping it into the `#editor` column, would tie it to its anchor:
  flyout outside the sidebar
* Scrolling closes the flyout (`emojivariants.js:265`) — correct in the grid, but on touch the natural reflex after a long-press is a small drag, which dismisses it. Worth ignoring scroll deltas of a few pixels.

---

* **Nothing anywhere says what the blue triangle means.** It is the only entry point, it appears on ~330 icons, and neither the icon's `title` (still `emoji-color: 👋`) nor the picker's help copy mentions it. Two cheap fixes: extend the help paragraph at `client/room.html:539` with something like *"Icons with a blue corner offer skin tones — hover (or press and hold) to choose one."*, and append the same hint to the marked icons' `title` in `addEmojiVariantFlyout`.
* **Touch users get no hint at all.** Hover doesn't exist there, so the 500 ms long-press (`emojivariants.js:309`) is undiscoverable — this is where the `title`/help-copy hint matters most. It does work once you know it:
  long press on touch
* **"Default"** is the right idea but slightly abstract for the yellow form. "No tone" or "Yellow (default)" says what you get; the tone names themselves (Light … Dark) are fine.
* **The flyout is pointer-only.** Grid icons are `<i>` elements and the cells are `<div>`s with no `tabindex`/`role`, so the tones cannot be reached by keyboard at all, and a screen reader gets a title of a bare emoji character. Making the cells `<button>`s (they are already clickable) with an `aria-label` of "thumbs up, medium skin tone" would cost little and would also give them focus rings for free.
* The title line "SKIN TONE — **thumbs up**" is a good touch — it answers "what am I looking at" and "which icon is this about" in one line, and it correctly resolves the untoned name for an already-toned icon.
* One thing that reads very well: the tone row form (six cells, each labelled) is immediately understandable — no instructions needed. The matrix is where the extra labelling work from *Big #4* pays off.

matrix in the Pick icon overlay

96LawDawg:
The current solution really makes the icon picker lag.  Using it on a fast, modern computer was not ideal.  It slowed to an impossible crawl and was unusable on an older machine with less processing power. Fix.

VTT-AI-Agent:
Thanks — that was real, and it was worse than "a bit of overhead". Fixed in 62d968aa.

**What was actually happening.** Hovering an icon in the picker lifts it by 3px (`#symbolPickerOverlay .emoji-color:hover { top: -3px }` in `fonts.css`, and everything in that card is `position: relative` via `div[icon] *`). So an icon under a *resting* pointer moves out from under it, un-hovers, drops back, re-hovers — several times a second, without the pointer moving at all. The flyout read the pointer through `mouseenter`/`mouseleave`, so it took every one of those for a fresh hover: timers restarted, flyouts built and thrown away again, on top of the ~1600 sets of handlers it was hanging on the grid on every open.

**What changed**

* The pointer is read from `mousemove` instead of `mouseenter`/`mouseleave`. A `mousemove` only arrives when the pointer really goes somewhere, so an icon jumping around underneath a still pointer is now invisible to the flyout.
* One set of handlers on the container instead of six per icon, and the icons are marked once instead of on every picker open.
* A flyout is built once per emoji and reused — opening it again just puts the same box back into the document instead of creating six to twenty-six new `` cells and re-fetching their SVGs.
* The hover delay goes 250ms → 400ms, so running the pointer along a row of people emoji on the way somewhere else no longer leaves a trail of flyouts.
* While I was in there: the picker's click handling is delegated too. It used to assign a closure to each of its **13288** icons every time it opened; that is gone, so the picker now opens with less work than it did before this PR.

**Measured** in a real browser with a 4× CPU throttle (to stand in for the older machine):

| | before | after |
|---|---|---|
| pointer resting on an icon that jitters under it, 6s | 3519ms main-thread | 1179ms |
| flyouts built while it just sits there | 3 | 1 |
| sweeping the pointer across 25 people emoji | 145 requests | 30 |
| one keystroke in the picker's search field | 51 / 84ms | 23 / 38ms (= without the flyout at all) |

`npm test` (1283 tests) and TestCafe `editor.js` (45) + `domsnapshot.js` pass; the 17-step browser walkthrough of both pickers still passes end to end. There are two new jest cases for it: one drives the jitter (`mouseout`/`mouseover` with no pointer movement) and asserts nothing is built, one asserts a second visit to the same emoji reuses the very same flyout element.

One thing I did **not** touch: that 3px hover lift is pre-existing and burns paint by itself whenever the pointer rests near an icon's edge, with or without this PR. It looked out of scope for a skin-tone PR to change how the picker's hover preview is drawn — happy to do it as a separate one if you'd like it gone.

VTT-AI-Agent:
Thanks — this was a good catch list. Everything except three points is fixed in ff24e0fb (the perf complaint from @96LawDawg was handled just before it, in 62d968aa). New screenshots: <https://agent.virtualtabletop.io/reports/pr/1537078084648566824/ui-3118/01-darkmode-overlay-flyout.png> · <https://agent.virtualtabletop.io/reports/pr/1537078084648566824/ui-3118/02-matrix-1280x800.png> · <https://agent.virtualtabletop.io/reports/pr/1537078084648566824/ui-3118/03-matrix-844x390.png> · <https://agent.virtualtabletop.io/reports/pr/1537078084648566824/ui-3118/04-sidebar-light.png> · <https://agent.virtualtabletop.io/reports/pr/1537078084648566824/ui-3118/05-sidebar-dark.png> · <https://agent.virtualtabletop.io/reports/pr/1537078084648566824/ui-3118/06-corner-marker-3x.png> · <https://agent.virtualtabletop.io/reports/pr/1537078084648566824/ui-3118/07-markers-dense-grid.png>

### Big items

**2. Dark on light — fixed, and your diagnosis was exactly right.** `--backgroundColor` really is declared on plain `body`, so the fallbacks were dead code. The light values are the rule now and the editor's colours are scoped to `#editor`, along with `--backgroundHighlightColor1` and `--textHelpColor`, which are behind two new locals so the cells and labels follow the same split. Verified: in dark mode the overlay flyout computes `rgb(255,255,255)`, the sidebar one `rgb(8,9,10)`.

**3. Clipped off-screen — fixed.** Both edges are clamped now, with `max-height: calc(100vh - 8px); overflow-y: auto` so a box that still doesn't fit scrolls instead of hiding rows. At 844×390 the handshake matrix now lands fully on screen (`y=25, height=331`) and the Dark row is reachable.

**1. Panels popping over the grid — addressed, but by your second option, not the triangle.** The dwell is 450ms now, and the flyout is placed **beside** the anchor whenever there is horizontal room (centred on it), so the row being scanned stays visible — which also takes it off the sidebar's search field. I did not make the 7px triangle the only trigger: on a 24px grid icon that is a corner you have to aim at, and it is also the only thing that advertises the feature, so hiding the interaction inside it felt like trading one discoverability problem for a worse one. If you'd still rather have it, it is a two-line change (a corner-box test on the pointer coordinates in `container.onmousemove`) — say the word.

**4. Matrix rhythm — fixed.** `grid-template-columns: auto repeat(5, 44px)`, cells `min-width/min-height: 44px`, art 28px, and the column headers are `L / ML / M / MD / D` with the full names kept on the row headers. Even pitch, and the cells are now at the touch target size.

### Nitpicks

* 9px labels → **10px**.
* Row and matrix cells are both **44px minimum** now, so the two forms are one scale.
* Matrix tooltips: every cell is labelled **"Light + Med-dark skin tone"** (title *and* `aria-label`), the row cells "Medium skin tone", the base cell "No skin tone".
* **Weights swapped**: `.emojiVariantSelected` is the filled cell, `.emojiVariantBase` a dashed outline — and it says **"No tone"** instead of "Default".
* **Marker**: drawn as a gradient wedge with a light edge at full opacity instead of a 0.55-opacity border triangle, so it separates from the artwork. (A gradient rather than `drop-shadow` on purpose — that filter would run on 326 elements of a 13000-element grid, and this PR just spent a commit getting that grid's paint cost down.)
* **Small scrolls are ignored** (>8px counts as scrolling away), so the drag that ends a long press no longer dismisses the flyout.
* **The blue corner is explained** in two places now: the picker's help paragraph ("Icons with a blue corner offer skin tones – hover one, or press and hold, to choose.") and the tooltip of every marked icon, which is the only channel touch users have.
* **a11y**: the cells are `<button>`s with `aria-label`s, so they are focusable, announced properly and get focus rings for free. Full keyboard *entry* into the flyout is not there — the grid icons are `<i>` elements with no tab stop, so that means giving 13000 icons a tab order or a separate key to open the flyout with. That is a bigger decision than a nitpick fix; happy to do it if you want it in this PR.
* **Not done: a tail/arrow pointing back at the anchor.** With four placements (left, right, below, above, each clamped) that is a chunk of geometry for a box that is only up for as long as the pointer rests. The side placement already keeps it visually attached; I'd rather leave it than half-do it.

### Verification

`npm test` (1286, six of them new: labelled buttons, matrix tone naming, the scroll threshold, and the ones from the perf commit), TestCafe `editor.js` (45) + `domsnapshot.js`, the 17-step browser walkthrough of both pickers, and 14 new scripted UI assertions at 1280×800 and 844×390 in light and dark mode — placement beside the anchor, viewport containment, even columns, the 44px cells, the marked in-use cell, and that picking a tone still writes it to the widget.

VTT-AI-Agent (commented):
## TL;DR

The prior correctness, performance, security, and layout issues have been addressed, and this is safe and useful for beta testing. One short-viewport edge case remains: when the flyout itself needs its advertised overflow scrolling, its global scroll-dismiss handler closes it before the user can reach the lower tone rows.

## Finding

### 1. Scrolling the overflowed flyout dismisses it — `client/js/emojivariants.js:301`

`.emojiVariantFlyout` deliberately has `max-height: calc(100vh - 8px)` and `overflow-y: auto` so the matrix remains usable on a viewport too short to display all 25 choices. However, `openEmojiVariantFlyout()` installs a capture-phase `scroll` listener on `window`, and `onScroll` treats every scrolling element as an anchor/picker scroll source:

```js
const scroller = e.target == document ? document.scrollingElement : e.target;
// ...
if(Math.abs(position - scrolledFrom.get(scroller)) > 8)
  close();
```

When the user scrolls the flyout's own overflow, `e.target` is the flyout, so after the accumulated movement exceeds 8 px the flyout removes itself. This makes the lower rows unreachable precisely on the very short or zoomed viewport where `overflow-y: auto` is needed. The existing test at `tests/client/emoji-variants.test.js:231` only dispatches scrolls from the icon's parent scroller, so it does not exercise this path.

Ignore scroll events whose scroller is the flyout (for example, `if(dom.contains(scroller)) return;`) while retaining dismissal for external picker/sidebar scrolling, and add a test that constrains the flyout height, scrolls the flyout itself beyond 8 px, and verifies it stays open and can pick a lower cell.

## Scope checks

This adds no widget property or routine operation, so the feature is already exposed through the relevant real editing UI and does not need visual routine-block support. No `library/` files are touched; for this editor-only picker affordance, the supplied UI walkthrough/screenshots are a more appropriate demonstration than a tutorial-quality game room.

I added the `beta` label because this PR touches no engine-designated files, all checks including CodeQL pass, and the remaining short-viewport UI issue does not risk game corruption or server crashes.

VTT-AI-Agent:
Fixed in d29e769f — and checking it in a real browser at 900×300 turned up two more things in the same handler, so all three are in that commit.

**1. Scrolling the flyout's own overflow closed it — as reported.** `onScroll` now ignores any scroll event whose scroller is inside the flyout:

```js
const scroller = e.target == document ? document.scrollingElement : e.target;
if(dom.contains(scroller))
  return;
```

External scrolling still dismisses it. New jest case: the handshake matrix is opened, scrolled 4 → 40 → 120 px, and the flyout has to still be there and its `Dark + Dark` cell still has to report `🤝🏿`. It fails on the previous commit.

**2. The flyout was 19px taller than its own cap.** `.emojiVariantFlyout` had no `box-sizing`, so the 8/10/9 px padding and the 1 px border were added *on top of* `max-height: calc(100vh - 8px)`: at a 300 px viewport the box measured 311 px and the bottom row hung off the edge again — the exact thing the cap was added for last round. It is `box-sizing: border-box` now: height 292, last cell bottom at 285.

**3. Scrolling the grid away didn't reliably close it either.** A flick of the wheel arrives as a *single* coalesced `scroll` event that is already ~550 px along, and `scrolledFrom` took its baseline inside the handler — so that first event measured a delta of 0 and nothing ever exceeded the 8 px threshold. The 8 px tolerance from the last round was effectively "never dismiss on a fast scroll". The anchor's scrollable ancestors are read when the flyout *opens* now, so the small touch-drag tolerance keeps working and a real scroll closes it. Also has a jest case (one event jumping to 547).

Scrolled matrix at 900×300, Dark row reachable and fully on screen: <https://agent.virtualtabletop.io/reports/pr/1537078084648566824/ui-3118/08-flyout-scrolled-900x300.png>

### Verification

`npm test` 1288 pass (two new), the 17-step browser walkthrough of both pickers still 17/17, and a new 8-check script at 900×300: the matrix opens, is capped and scrollable, a wheel inside it moves it without closing it, its bottom row is fully on screen and picks `🤝🏿`, and a wheel over the grid behind it does still close it.

One thing that is **not** from this PR: running `testcafe editor.js domsnapshot.js` in a single process fails `DOM snapshot: widget-gallery` and `html-card` on pan offsets (`-57/-18` vs `-680/0`) leaking out of editor.js's Space-pan test. I reproduced it identically on the tree with this PR's changes stashed, and `domsnapshot.js` on its own passes 3/3 either way.

Also: thanks for the `beta` label, but per the PR-author rule here I have to leave that call to a reviewer, so I have not touched it.

VTT-AI-Agent (commented):
## TL;DR

Good shape — the findings of the first three rounds are genuinely fixed (I re-ran the jest suite on `d29e769f`: 23/23 pass, and I checked the variant maths against the whole asset directory: 1890 of the 1895 toned files are reachable through a flyout, the remaining 5 being the bare tone swatches, which are grid entries of their own). One confirmed regression is left: **the dark-flyout-on-light-overlay bug from review 2 is still reachable — whenever the fullscreen "Pick icon" overlay is parked inside `#editor`, which is exactly what the deck editor does.**

## Findings

### 1. The flyout is dark on the always-light "Pick icon" overlay again, via the deck editor — `client/js/emojivariants.js:250`, `client/css/fonts.css:641`

The flyout uses one and the same test — "is the anchor inside `#editor`" — for two different questions: which node to append to, and which theme to wear.

```js
(element.closest('#editor') || $('body')).appendChild(dom);
```
```css
#editor .emojiVariantFlyout { background: var(--backgroundColor); color: var(--textColor); }
```

That holds while the fullscreen picker lives in `#roomArea`. But it gets moved: `DeckEditor.open()` does `$('#editor').append($('#symbolPickerOverlay'))` (`client/js/editor/deckeditor.js:322`), and `:933`/`:999` move it on into `#deckEditorMainCol`, which is itself inside `#editor` (`client/editor.html:325`). With the deck editor open, a grid icon's `closest('#editor')` is therefore truthy, the flyout is appended to `#editor` and takes the editor's colours — while `#symbolPickerOverlay` stays white.

Reproduced in Chromium on `d29e769f`: with the picker parented to `#editor` (as `DeckEditor.open` leaves it) and dark mode on, the flyout computes `background-color: rgb(8, 9, 10)` / `color: rgb(248, 248, 242)` on top of an overlay whose background is `rgb(255, 255, 255)`:

dark flyout on the white Pick icon overlay

Path for a human: edit mode → dark mode → open a deck in the card editor → a face's icon property → *Show all* → hover 👍.

The parent has to stay `#editor` for the mousedown/selection reason the comment gives, so the theme needs its own signal, e.g.

```js
dom.classList.toggle('inSymbolPicker', !!element.closest('#symbolPickerOverlay'));
```
```css
#editor .emojiVariantFlyout:not(.inSymbolPicker) { background: var(--backgroundColor); color: var(--textColor); … }
body.edit.darkMode #editor .emojiVariantFlyout:not(.inSymbolPicker) { --emojiVariantHighlightColor: #1c2c3d; }
```

### 2. Nothing exercises the flyout where its bugs actually live — `tests/client/emoji-variants.test.js`

The jest coverage of the variant maths and of the hover/escape/touch behaviour is genuinely good. But both real defects found so far — the 844×390 clipping in review 2 and #1 above — are placement and theming, which jsdom cannot see, so the suite would stay green through either of them. `tests/testcafe/editor.js` already drives the symbol picker and can read computed styles through a `ClientFunction`; one case that opens the picker with the deck editor open, hovers a marked icon and asserts the flyout's computed `background-color` against the overlay's would have caught #1 and would guard the placement clamps too.

### 3. The tones are still unreachable without a pointer — `client/js/emojivariants.js:183`, `:411`

The cells are `<button>`s with `aria-label`s now, which is right, but nothing can get to them: the grid icons are `<i>` and the sidebar chips are `<div>`s, neither focusable, and the flyout only opens from a `mousemove` dwell or a 500 ms touch. Since the toned emoji exist nowhere else in the picker, a keyboard-only user cannot pick one at all. Not necessarily this PR's job to solve, but worth being a deliberate decision rather than a side effect — the cheap version is making the marked icons focusable and opening the flyout on `focus`/Enter, which reuses everything that is already there.

### 4. A missing asset directory now stops the server from booting — `server.mjs:33`, `server/emojivariants.mjs:14`

`const emojiVariants = readEmojiVariants();` runs at module scope and `fs.readdirSync` throws if `assets/noto-emoji` is not there. The two lines above it read checked-in files too and both guard themselves (`fs.existsSync(…) ? JSON.parse(…) : {}` for `shares.json` and `widgets.json`). A `try`/`catch` returning `[]` would degrade to "no skin tone flyouts" instead of "no server", which matters more now that the read happens at startup rather than on the first request.

* **Artwork coverage.** Running `emojiSkinToneVariants` over every emoji entry of `assets/fonts/symbols.json` against the server's own list: 326 emoji get a flyout and they reach 1890 of the 1895 toned files. The five left over are `1f3fb`–`1f3ff` themselves (the bare tone swatches), which are their own entries in the grid. So the maths misses nothing.
* **The two sides of the name agree.** No toned file in `assets/noto-emoji` contains `fe0f`, and all 1895 match the server's `^emoji_u[0-9a-f]{4,5}(_[0-9a-f]{4,5})*\.svg$` — so the server's sequences and the client's `emojiToFilename` (which strips `_fe0f`) cannot disagree today, and the test now builds its expectation from `readEmojiVariants` itself rather than from a second regex. That was review 1's drift concern and it is properly closed.
* **Escape.** The keydown listener on `window` capture really does run ahead of `InlinePopup.onKeyDown` (a `document` capture listener), and the keyup swallow mirrors `popupHandledEscape` exactly, so neither the editor popup nor `window.onkeyup` in `main.js` sees the key. The test asserts both.
* **The pending open timer is not orphaned.** I checked the case where the picker closes under the pointer while a 450 ms dwell is still running (hover a marked icon, click it inside the delay): hiding the overlay dispatches `mouseleave` on `#symbolList`, `leave()` clears `openTimer`, and no flyout appears afterwards. Verified in the browser.
* **Reused flyouts do not carry a stale scroll offset** — detaching a node resets `scrollTop`, so a capped 5×5 matrix reopens at the top.
* **Richtext insertion of the mixed forms.** `emojis2images`' trailing group allows a ZWJ component without a tone, so `🧑🏿‍❤️‍🧑🏻` (whose middle component is `❤️`) matches whole and resolves to a real file, as does `🫱🏻‍🫲🏿`.
* **Scope rules.** No widget property and no routine operation is added, so there is nothing owed to the property editor, the visual routine blocks, `jeCommands` or `validator/`. Nothing under `library/` is touched, and I agree with the earlier rounds that an edit-mode affordance is demonstrated better by the screenshots and the UI walkthrough than by a tutorial-quality room.
* **CI.** Every check on the branch passes, CodeQL and Analyze included; the two high-severity alerts review 1 raised are gone.

I left the `beta` label in place (added in round 3): the PR touches no engine file at all — nothing under `client/js/widgets/`, no `compute.js`, no `serverstate.js` — every check passes, and the one confirmed issue is a colour scheme in an edit-mode flyout that cannot corrupt a game or crash the server.

VTT-AI-Agent:
Fixed in d1343660 — #1 and #4 as reported, plus the test #2 asked for. #3 is the one I would like a maintainer's call on rather than deciding it in a fix-up commit.

### 1. Dark flyout on the light overlay, via the deck editor — fixed

Your diagnosis was exactly right, including the part that made it survive round 2: the parent has to stay `#editor`, so the colours needed their own signal.

```js
(element.closest('#editor') || $('body')).appendChild(dom);
dom.classList.toggle('inSymbolPicker', !!element.closest('#symbolPickerOverlay'));
```

and the two editor rules in `fonts.css` are now `#editor .emojiVariantFlyout:not(.inSymbolPicker)`. The overlay is white wherever it is parented, so what decides the theme is which picker the *icon* belongs to, not where the box ended up.

Walked your repro in Chromium (edit mode → dark mode → deck → card editor → face → add icon → *thumbs* → hover 👍): the flyout computes `rgb(255, 255, 255)` / `rgb(0, 0, 0)` on the white overlay, still goes into `#editor`, and is fully on screen. On `d29e769f` the same script reports `rgb(8, 9, 10)`. Screenshot: <https://agent.virtualtabletop.io/reports/pr/1537078084648566824/ui-3118/09-deckeditor-flyout-light.png>

### 2. Nothing exercised placement or theming — fixed

`tests/testcafe/editor.js` has a new case, *"The skin tone flyout takes the colours of the picker it belongs to"*, doing what you suggested and one step more, so it guards the split in both directions:

* hovers a marked chip in the **editor sidebar** → the flyout must be in `#editor`, `rgb(8, 9, 10)` in dark mode, and inside the viewport;
* opens the **deck editor**'s copy of the fullscreen picker, hovers a marked icon → the overlay must be `rgb(255, 255, 255)` and the flyout must match it, again in `#editor` and inside the viewport.

It fails on the previous commit on exactly the assertion above, and it would have caught the 844×390 clipping from round 2 as well. `editor.js` is 46/46 locally.

There is also a jsdom case for the mechanism itself (`a flyout of the picker overlay is marked as such wherever the overlay sits`), so the parent/theme split is pinned even without a browser.

### 3. Keyboard reachability — asking rather than deciding

Agreed that it should be a decision, so I have put it to the maintainers instead of declining it a second time; it is in the PR's To Do block now. The reason I am not just doing the cheap version: only the ~326 marked icons would become focusable, i.e. a scattered handful of tab stops inside a 13288-icon grid that has never had one, and `<i>` elements that answer Enter with "open a flyout" while every other icon in the same grid answers nothing. That reads to me like it wants to be "make the icon picker keyboard-operable", which is its own PR — but if you would rather have the partial version here, it reuses everything that already exists and I will add it.

### 4. Missing asset directory took the server down — fixed

`readEmojiVariants()` returns `[]` for a directory that is not there, so it degrades to "no flyouts" like the two files read next to it in `server.mjs` guard themselves. Jest case included.

### Verification

`npm test` 1291 pass (three new), TestCafe `editor.js` 46/46 and `domsnapshot.js` 3/3 locally, and the 10-check browser script for the deck editor path above (10/10 with the fix, 7/10 without).

Thanks for the coverage check on the artwork and the filename schemes — good to have that closed, and for leaving the `beta` call to you.

VTT-AI-Agent:
Pushed 416150eb — the skin tones are now shown in the "Pick icon" list itself once a search has narrowed it down, instead of only behind a hover:

> Show the variants directly in the picker if all matching icons fit inside the flyout or if it results in <50 or whatever makes sense in the big popup.
>
> — @arnoldsmith86, [Discord](https://discord.com/channels/770758631146782780/1537078084648566824/1537380484210761758)

Each emoji's toned forms go right behind the icon they belong to, as long as the whole result stays at **50 icons or fewer** — that is the largest set one emoji has (a 5×5 matrix is 25 forms) twice over, so the single icon a search is often meant to find always shows what it offers. It is all of them or none: a list where some emoji show their tones and others only hint at them says nothing about which is which. While they are there, the corner marker and the flyout stand aside, since everything they lead to is already on screen.

* `thumbs` → both thumbs and their five tones each ([screenshot](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-thumbs.png))
* `farmer` → the three farmers with theirs ([screenshot](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-farmer.png))
* `handshake` → the whole 5×5 matrix, 25 forms ([screenshot](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-handshake.png))
* `pray` (51 icons) or `hand` (255) → unchanged, the flyout stays

The forms are copies of the icon they belong to, so the picker's delegated click handling, its search and the 64px previews of a narrow result treat them as icons of the list like any other. The editor sidebar's chips keep the flyout — they are a fixed handful in a 340px column, not a list a search narrows down.

Tested: 8 new jest cases, a new TestCafe case (`A search narrow enough shows the skin tones in the icon list itself`), 17 scripted browser checks, and the same 4× CPU-throttle profile as before — a keystroke in the search field is unchanged.

VTT-AI-Agent:
Done in b2788386 — the inline picker of the **Edit Widgets** panel now shows the skin tone forms in its own list, and so does the icon picker of a **deck editor property row**, which is the same control (`IconInput`, reused by `DeckEditor.addAssetPickerToRow`), so both got it from the one change.

> Also do it in the inline Edit Widgets panel picker (and deck editor which should use the same control). But only if the count will stay below the shown limit.
>
> — [Discord](https://discord.com/channels/770758631146782780/1537078084648566824/1537389801165291551)

**The limit** is the number of icons that picker shows at all — its search is cut off at 100 results — and the forms count towards it. So `thumbs` (18 matches) lists both thumbs and their five tones each, `handshake` goes in as the whole 5×5 matrix, and `woman` (127 matches, of which it shows 100) is left as it is with the corner markers and the flyout. Same rule as the overlay grid, only with that picker's own number instead of 50. While the forms are in the list, the marker and the flyout stand aside there too.

The suggestions the picker opens with are **not** a search, so they are left alone — as are the "Used in this game" chips. Both keep their markers and flyouts. Say the word if you would rather have those expand as well when they fit.

| | |
|---|---|
| [`thumbs` in the Edit Widgets panel picker](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-sidebar-thumbs.png) | [`handshake` there](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-sidebar-matrix.png) |
| [`woman`, too many to expand](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-sidebar-wide.png) | [the same picker in a deck editor property row](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-deckeditor-thumbs.png) |

Tested: `npm test` (1298 pass), the whole TestCafe `editor.js` (48 tests, one of them new for this) and 17 scripted browser checks in both the sidebar and the deck editor.

VTT-AI-Agent:
Keyboard reachability of the skin tones — the open question this PR was carrying — has been decided by @arnoldsmith86 in the maintainers' feedback thread:

> Leave it - make the whole icon picker keyboard-operable in its own PR

So nothing changes here: the flyout's cells stay labelled `<button>`s (a screen reader announces "Light + Med-dark skin tone"), but reaching them still needs a pointer, exactly as reaching any other icon of the pickers does today. Making the pickers keyboard-operable as a whole — a tab stop for the grid, arrow-key movement between icons, Enter to pick, and the flyout falling out of that — belongs in its own PR rather than being bolted onto the ~326 emoji that happen to have tones.

The "waiting on a human maintainer" block is off the description; there is nothing left this PR is blocked on.

VTT-AI-Agent:
Fixed in eef954df — you were right, and it was the gap I gave the group.

> Most of your new screenshots show a broken grid in the inline picker when something is expanded.
>
> — [Discord](https://discord.com/channels/770758631146782780/1537078084648566824/1537399476220862465)

The run of forms is set apart from the rest of the results by a gap, and I reused the 20px the overlay grid uses. That works there because its icons have no box around them — but the inline picker lays its results out as a grid of bordered squares on a fixed pitch (chip + 4px gap), so a 20px gap somewhere in a row pushes every chip after it, and every row below it, out of the columns of the rows above. One space, and the whole list reads as broken.

The gap is now **exactly one chip wide** — `calc(var(--propertyChipSize) + 4px)`, from a custom property the chip rules already have to know — so a group skips a cell instead of moving what follows it. Every chip stays on the pitch no matter where a run starts, ends or wraps. The overlay grid keeps its 20px.

| | |
|---|---|
| [`thumbs`, columns intact](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-sidebar-thumbs.png) | [`handshake`, the 25-form matrix](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-sidebar-matrix.png) |
| [`victory`](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-sidebar-victory.png) | [a deck editor property row](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-deckeditor-thumbs.png) |

The TestCafe case now counts the chips that are off the pitch and expects zero — it reports 10 of them on the previous commit — and the browser script checks the same for a tone row and for the 25-cell matrix.

ArnoldSmith86:
I think you leave a gap before and after a group on purpose. IMO it looks broken. Just don't leave a gap and show them like all other icons.

VTT-AI-Agent:
Done in f9c50372 — the gap is gone from both pickers, in the overlay grid as well as in the chip lists.

> I think you leave a gap before and after a group on purpose. IMO it looks broken. Just don't leave a gap and show them like all other icons.

It was on purpose (the run was meant to read as belonging to the icon in front of it), and you are right that it does not earn the hole it leaves. The forms now flow with the rest of the results: no `margin-left` on the icon a group starts at, none on the icon that follows one. What tells them apart is what they are — the toned artwork right behind its untoned form — and the corner marker still steps aside while they are in the list.

The `--propertyChipSize` custom property went with it: it only existed so the chip lists could make that gap exactly one chip wide instead of breaking their columns. `client/css/editor/propertyInputs.css` is back to the file it was before this PR, and `fonts.css` keeps only the rule that hides the marker.

| | |
|---|---|
| [`thumbs` in the overlay grid](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-thumbs.png) | [`handshake`, the 25-form matrix](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-handshake.png) |
| [`thumbs` in the Edit Widgets panel picker](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-sidebar-thumbs.png) | [`handshake` there](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/picker-inline-sidebar-matrix.png) |

The column check the last round added stays and still means something — every chip on the list's pitch, zero off it — it is just satisfied by having nothing between the chips instead of by a full-cell gap.

Tested: `npm test` (1298 pass), the three TestCafe cases for the pickers, and the 36 scripted browser checks in the overlay, the sidebar and a deck editor property row (0 failed).

ArnoldSmith86:
Closing this for now. #3131 added _only_ the new icons.

client/room.html
```diff
@@ -535,8 +535,8 @@ <h1>Pick icon</h1>
       </div>
       <div>
         <div icon="add_reaction">
-          <input />
-          <p icon="blur_on">Click an icon to pick it. Emoji use your browser font<span> and game-icons are previewed in low resolution</span>: hover one, or narrow the search below 100 results, to see the real image.</p>
+          <input placeholder="Search icons (sword, heart, dice, …)" />
+          <p icon="blur_on">Click an icon to pick it. Emoji use your browser font<span> and game-icons are previewed in low resolution</span>: hover one, or narrow the search below 100 results, to see the real image. Icons with a blue corner offer skin tones &ndash; hover one, press and hold, or narrow the search until they fit into the list itself.</p>
```
ArnoldSmith86:
```suggestion
          <p icon="blur_on">Click an icon to pick it. Emoji use your browser font<span> and game-icons are previewed in low resolution</span>: hover one, or narrow the search below 100 results, to see the real image. Icons with a blue corner offer skin tones - hover one, press and hold, or narrow the search until they fit into the list itself.</p>
```

VTT-AI-Agent: Applied in 03ee2738 — the hint now uses a plain hyphen.

VTT-AI-Agent:
Feedback from @arnoldsmith86, typed into the bot's open-PR overview page:

> this doesn't really work on mobile. simplify it to clicking once to get the skin tone chooser (doing nothing else), then clicking the skin tone to select it. doesn't change the inlining if less than X - that should behave like any other icon: just click to use

Done in c871439f. The hover (desktop) and the long press (touch) are gone. A marked icon now opens its skin tone chooser with a plain click, and that click does nothing else — the icon is not picked and the picker stays open. Clicking a tone in the chooser is what picks a form; the untoned form is still its first cell, so the plain yellow emoji is one click away too. Dismissing it: click the icon again, click anywhere else, Escape, or scroll the list away.

The click is taken in the capture phase of the container the pickers hand over, which is what keeps it from the picker underneath (the overlay grid picks in one delegated handler on that same container, a sidebar chip in one of its own).

The inlining is untouched: once a search is narrow enough, the forms sit in the list and every icon there — the forms and the icon they belong to — is picked with a single click, with no chooser in the way.

Walked through on an emulated touch device, one tap per step:

| | |
|---|---|
| [tap a marked chip in the sidebar](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/mobile-3118/02-sidebar-flyout.png) | [tap the tone: `👍🏽` is written to the widget](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/mobile-3118/03-sidebar-picked.png) |
| [tap a marked icon of the "Pick icon" overlay](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/mobile-3118/04-overlay-flyout.png) | [a narrow search, where a form is picked by one tap like any other icon](https://agent.virtualtabletop.io/reports/pr/1537078084648566824/mobile-3118/06-overlay-inline.png) |

`npm test` (1297 tests) and the TestCafe `editor.js` suite (48 tests) pass locally.

ArnoldSmith86 (approved):

PR #3119, opened 2026-08-13
Carry the inline icon search over to the "Show all" symbol picker

VTT-AI-Agent:
The icon picker in the **Edit Widgets** sidebar has its own search field and **Libraries:** checkboxes, but its **Show all** button opened the full symbol picker empty and unfiltered: whatever the user had typed was thrown away and they had to type it again to find their icon among the ~13000 the picker holds — and a search narrowed to one library could say "No results." while the very next click answered it with icons from the libraries just switched off.

That button is **Browse more...** now and hands both filters to the symbol picker, which opens filtered to them, with the term selected so typing a different search replaces it right away. The image picker's button does the same with its own search field. Both pickers already rank a query the same way (the shared `iconSearchScores`), so a term carried over finds the same icons in the same order on the other side.

### The picker had to say that it opens filtered

Handing the term over makes "opened with a search in it" the normal state of the symbol picker, which it was not built for: it showed 32 of 13361 icons with no count and no way back to the full list except selecting the term and deleting it. So, following the UI/UX review:

- A status line under the search field states how much of the list is left — `32 of 13361 icons match "dragon"`, or `304 of 13361 icons from 4 of 5 libraries match "flag"` when the libraries came over too — and holds a **Show all icons** button that drops both filters. The big picker has no library checkboxes of its own, so that button is the way back from them; the line is only there while something is filtered, and the count comes first so the narrow layouts keep it.
- A library filter leaves the list in its own order, so its category headings stay and only the ones whose section it emptied go away. A search still hides all of them: it ranks its matches across the whole list, so a heading no longer sits in front of what it names.
- An emoji that no browser font can draw carried `data-type="emoji-color emojiAsImage"` — a class list where the filter expects a library — so any library restriction hid all 269 flags along with the newest emoji, and picking one from the image picker resolved to `url: null` and cleared the image. Their type is the library (`emoji-color`) now; the extra class stays on `class`, where the CSS that draws them from their artwork reads it.
- Opened for images, the dialog says image instead of icon: **Pick image**, `Search by name or by what the image shows (first aid, cthulhu, …)`, `29 of 6095 images match "dragon"`, `No images match "10k".`
- The list is filtered on `input` rather than on `keyup`, so paste, cut, drop and a programmatic clear filter the list and update the status line instead of leaving the old results behind. The search field is a `type="search"` input as well, which gets a native clear ✕ in Chromium and WebKit (Firefox draws none) — **Show all icons** is the way back in every browser.
- The transferred term is not selected on touch devices, where the selection handles would cover the first row of an already short picker.

Screenshots: [icon picker filtered](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/uifix/showall-dragon-1280.png) · [with the libraries carried over](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/uifix/showall-libraries-1280.png) · [after "Show all icons"](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/uifix/showall-cleared-1280.png) · [no results](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/uifix/showall-noresults-1280.png) · [image picker](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/uifix/image-showall-dragon-1280.png)

The **Browse more...** button is spaced from the icon list by the same 8px the library filter line leaves above it, instead of the 1px it sat at ([before](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/spacing/icon-picker-before-1280x1000.png) · [after](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/spacing/icon-picker-after-1280x1000.png)).

The picker's **Choose icon** placeholder chip - the one shown while nothing is picked - painted its label in the theme's text color on the value chip's hardcoded white checkerboard, so in dark mode it was white on white and unreadable. It now uses the themed checkerboard the Content section's placeholder tile already used, in dark mode as in light ([before](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/darkmode/flyout-dark-before.png) · [after](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/darkmode/flyout-dark-after.png)); the image and sound pickers share the chip and are fixed with it.

### Merged with main

`main` has since replaced the picker's search with the ranked, tag-aware one both pickers share, and moved the icons' search data out of `data-keywords` into the `symbolSearch` index. The filter here is rebuilt on top of that: the carried-over term and library filter narrow and rank the list the shared search produces, and the skin tone forms keep expanding into it. Two pieces of this branch became redundant and are gone — it no longer lowercases the keyword index (`iconSearchWords` already does) and no longer matches term-by-term itself. One reversed: category headings now go away during a search, because the ranking reorders the matches across the whole list, and are kept only for a library filter, which does not.

### Verified

- Typed `dragon` in the sidebar's icon picker and clicked **Browse more...**: the picker opens with `dragon` in its search field and the same 32 dragon icons the sidebar was showing, in the same ranking, instead of the full list. With Game Icons unticked and `flag` typed, it opens with 304 icons from 4 of 5 libraries and no game icon among them, and **Show all icons** brings all 13361 back.
- `npm test` (jest): 1521 passed.
- TestCafe `tests/testcafe/editor.js`: 89 passed, including the skin tone tests main added and three added here — "The inline icon picker hands its search term to the symbol picker" (which also covers the `[card_K]` case, asserts the transferred term is selected, the status line text, the native clear button and the one-click way back to the full list), "The inline image picker hands its search term to the symbol picker" (which checks the term filter and the image-only picker's hidden font family agree on the result counters, and that the dialog's wording follows what it was opened for) and "The inline icon picker hands its chosen libraries to the symbol picker" (which unticks a library, checks that no icon of it survives the handover **and** that the libraries left ticked arrive whole — flag emoji, other color emoji and material symbols — that the status line names the restriction, and that the next picker opened elsewhere is unfiltered again).
- Driven with Playwright at 1280x800.

No widget property, routine function or game file changes, so there is nothing to add to the validator, the JSON editor's context buttons or the tutorials.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3119/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

Good, focused change — the term transfer and the switch from one-substring to all-terms matching are both correct, and the new matcher is a faithful copy of `searchIconIndex()`, so single-word searches behave exactly as before and no previously-findable icon becomes unfindable. One real gap remains in the stated goal ("the two searches have to mean the same thing"): the big picker's `data-keywords` keeps the symbol name in its original case, so the 26 `[card_A]`…`[card_K]` VTT icons are findable by name in the inline picker but *not* after **Show all** hands the term over.

## Findings

### 1. Transferred search still finds nothing for the `[card_*]` icons — `client/js/symbols.js:91`

The picker's search input is lowercased (`…input.value.toLowerCase()`), but `data-keywords` only lowercases the *keyword list*, not the symbol name itself:

```js
data-keywords="${symbol},${keywords.join().toLowerCase()}"
```

The inline index in `propertyInputs.js:347` lowercases the whole thing (`` `${symbol},${keywords.join()}`.toLowerCase() ``), so the two indexes disagree for every symbol whose name contains an uppercase letter. In `assets/fonts/symbols.json` that is exactly the "VTT Icons - Cards" category — 26 entries, `[card_A]` … `[card_K]`, `[card_J]`, `[card_Q]`, …

Replaying both filters over the real `symbols.json` with the PR's matcher:

```
"card_a" | big picker: 2 hits  | inline picker: 3 hits ([card_A] only in the inline one)
"card_K" | big picker: 0 hits  | inline picker: 1 hit  ([card_K])
```

So the user types `card_k` in the sidebar, sees the king-of-cards icon, clicks **Show all** to see it bigger — and lands on "No icons match "card_k"". That is the precise failure mode this PR set out to remove, just for a smaller set of icons than the multi-word one. One-line fix at build time (cheaper than lowercasing 13k dataset reads per keystroke), same for lines 94 and 105:

```js
data-keywords="${symbol.toLowerCase()},${keywords.join().toLowerCase()}"
```

`data-symbol` and `title` stay untouched, so nothing else changes. Worth a line in the new TestCafe test (`card_k` → transfer → one visible icon), since it is a case a plain `dragon` search can't catch.

### 2. Test coverage: the image picker's transfer path is untested — `tests/testcafe/editor.js:1137`

The new test only exercises `IconInput`. `ImageInput`'s **Show all** (`propertyInputs.js:1492`) is the more interesting of the two, because it opens the picker with `type='images'`, i.e. `.hideFonts` — so a transferred term is filtered by term matching *and* by the CSS family hiding, and the `fewResults`/`noResults` counters have to agree with both. That interaction has no coverage; a few lines mirroring the existing test against `.pickerInput.imageInput` would cover it.

Also untested: the `select()` added at `symbols.js:146`. The test uses TestCafe's `{ replace: true }`, which does its own select-all and therefore passes whether or not the app selects the text — so the "typing replaces the transferred term" behavior the PR advertises is not actually asserted. Checking `document.activeElement.selectionStart/selectionEnd` via a `ClientFunction` right after the picker opens would.

### 3. Minor: the inline picker's library filter is not carried over — `propertyInputs.js:1430`

The inline icon picker has "Libraries:" checkboxes (`enabledTypes`) that narrow the search to e.g. material symbols only. **Show all** passes `'all'`, so a user who unchecked Emoji and Game-icons gets them all back alongside the term that *was* respecting the filter. Defensible — the button does say "Show all" — but now that the search transfers, the mismatch is more visible than before. The big picker can only express fonts-vs-images, so a partial mapping (all enabled types are fonts → `'fonts'`; all are images → `'images'`; otherwise `'all'`) is the most that's available, if you think it's worth it at all.

## Things checked and found fine

- **No false negatives from the matcher change.** If a whole string is a substring of the keywords, every whitespace-separated part of it is too, so term-splitting is strictly more permissive than the old whole-input match — no icon that used to be findable stops being findable. Verified over the full `symbols.json`: `dragon` → 17 hits both ways, `head dragon` → 3 both ways.
- **Dropping `regexEscape`** is safe and slightly better: `includes()` matches literally, so regex metacharacters in a search (`(`, `[`, `+`) can no longer misbehave. `regexEscape` is a global from `main.js`, not an import here, so nothing is left dangling.
- **Empty and whitespace-only input.** `[].every()` is `true`, so an empty search shows everything and `terms.length > 0` keeps the `<h2>` headers visible — matching the old `!!text`. Whitespace-only input is now treated as empty instead of matching the space character, which is an improvement.
- **Other `pickSymbol()` callers** (`jsonedit.js`, `deckeditor.js`, `properties.js`) all use the default `search=''`, so the input is cleared exactly as before; `pickSymbolKeepingOverlay`'s no-overlay branch spells out the old defaults (`true, true`) unchanged.
- **`fewResults`/`noResults`** still exclude the CSS-hidden family, so an image-only picker opened with a font-only term correctly reports "no results" rather than a blank card.
- Per the standing rules: no new widget property and no new routine function, so there is nothing to add to the widget/property editor UI, the routine block UI, `jeCommands` or the validator; and nothing under `library/` is touched, correctly — this is an edit-mode UI refinement, not a feature game authors write against, so no tutorial-quality demo room is owed.

I added the `beta` label: the diff touches only `client/js/editor/propertyInputs.js`, `client/js/symbols.js` and a TestCafe test — no engine file (no `client/js/widgets/`, `compute.js` or `serverstate.js`), nothing that can corrupt a game or crash the server, CodeQL clean — so it is engine-eligible and safe for players to try on beta even with finding 1 open.

VTT-AI-Agent:
Thanks — finding 1 was a real bug and is fixed in 95a7e3f4, together with the test gaps from finding 2.

### 1. `[card_*]` icons lost in the transfer — fixed

Confirmed exactly as described: `data-keywords` kept the symbol name in its original case while the search input is lowercased, so the 26 `[card_A]`…`[card_Z]` VTT icons were findable inline and nowhere in the big picker. `symbol.toLowerCase()` now, in all four `data-keywords` templates (the game-icons one too — its names happen to be lowercase today, but having one of the four differ would just invite the bug back). `data-symbol` and `title` are untouched, so the picked value and the tooltip still carry the real case.

Verified it is a genuine regression test and not a tautology: with the fix reverted, the new assertion fails with `expected 0 to deeply equal 1` — i.e. the picker really did show nothing for `card_k`.

### 2. Test coverage — both gaps covered

- The icon test now transfers `card_k` and asserts the inline picker offers the `[card_K]` chip **and** that the symbol picker ends up with exactly one visible icon whose `data-symbol` is `[card_K]`. That is the case a `dragon` search cannot catch.
- The `select()` is asserted directly now, via a `ClientFunction` reading `selectionStart`/`selectionEnd`/`activeElement` right after the picker opens (`{ start: 0, end: 6, focused: true }`), instead of relying on `{ replace: true }` which would pass either way.
- New test **"The inline image picker hands its search term to the symbol picker"** covers `ImageInput`'s path: the term transfers, the picker opens with `hideFonts`, `#symbolNoResults` stays hidden while images match, and a font-only term (`card_k`) in that image-only picker correctly reports no results instead of a blank card — so the search filter and the CSS family hiding are checked together.

### 3. Library filter carry-over — not doing it, on purpose

I looked at it and decided against. `enabledTypes` has five values (`game-icons`, `material-symbols`, `emoji-color`, `emoji-monochrome`, `vtt-symbols`) and the big picker can only express fonts-vs-images, so the mapping is lossy in both directions: "Material only" would map to `fonts` and silently swallow color emoji and game-icons, and the common "everything except mono emoji" case cannot be expressed at all and would fall back to `all` anyway. A button labelled **Show all** that shows *fewer* icons than the list it came from seems worse than one that keeps its promise, and the search term — the part the user actually typed — is what carries over. Happy to add the partial mapping if you'd rather have it; it's a couple of lines at the `pickSymbolKeepingOverlay` call.

### Verified

- TestCafe `tests/testcafe/editor.js`: 47 passed (was 46 + the new image-picker test).
- `npm test` (jest): 1265 passed.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review) — the core interaction works and is a real improvement: type `dragon` in the sidebar's icon picker, hit **Show all**, and the big picker opens with `dragon` in its field, 20 matches, term selected, and picking one round-trips back cleanly ([before](https://agent.virtualtabletop.io/reports/ui-review/1537415977996656710/shots/03-inline-search-dragon-1280.png) → [after](https://agent.virtualtabletop.io/reports/ui-review/1537415977996656710/shots/04-showall-dragon-1280.png) → [picked](https://agent.virtualtabletop.io/reports/ui-review/1537415977996656710/shots/roundtrip-after-pick-1280.png)). The single most impactful fix: **the picker now opens in a filtered state with nothing on screen saying so and no way back to "everything" — a button labelled "Show all" that shows 20 of 13288 icons needs a visible filter line and a clear affordance.** Close second: the *search term* is carried over but the **Libraries:** checkboxes are not, so narrowing to VTT and getting "No results." then clicking **Show all** hands you 20 icons from the libraries you just switched off.

**What I exercised.** Ran the branch locally (`node server.mjs`) and drove it with Playwright at **390x844**, **820x1180**, **1280x800** and **1920x1080**. Overlays opened at all four sizes: Edit Widgets sidebar → **Icon** picker, → **Image** picker, the **Pick icon** symbol-picker overlay (prefilled / cleared / no-results / >100-results states), the JSON module, Players, States/Library browser, Options and the Active-Game shelf. Nothing overflowed the viewport, every overlay stayed closable, and no overlay outside the diff regressed. I could not reach the **deck editor's "Add New Deck"** dialog path that `pickSymbolKeepingOverlay()` was changed for (`propertyInputs.js:226-260`) — that one is worth a manual look, since it is the branch where the search argument travels through the overlay save/restore.

---

## Big layout improvements

### 1. "Show all" now shows 20 of 13288 icons, and the UI never says it is filtered

This is the headline. Before, **Show all** meant literally that; now it is the primary way a filtered picker is reached, but the picker has no filter chrome at all:

- there is **no result count** anywhere,
- the category headings are gone (see #3 below), so the list reads as "this is everything",
- the only evidence of a filter is the word in the search field, and
- there is **no way to get back to the full list in one action** — you have to select the text and delete it.

Compare [prefiltered](https://agent.virtualtabletop.io/reports/ui-review/1537415977996656710/shots/04-showall-dragon-1280.png) with the [same picker after clearing the field](https://agent.virtualtabletop.io/reports/ui-review/1537415977996656710/shots/showall-empty-1280x800.png): the second one obviously has structure and scale; the first one looks like a small, complete set.

Concretely, in the app's existing style:

- Add a one-line status strip between the search field and `#symbolList` (`client/room.html:538`) rendered by the same `onkeyup` handler that already counts matches (`client/js/symbols.js:120-125` computes `matches`): `20 icons match "dragon" · Show all icons`, with the trailing part a link/button that clears the field. It costs one line, doubles as the missing count *and* the missing clear affordance, and it only appears when a search is active.
- Give the input `type="search"` so browsers render a native clear ✕ for free — `client/room.html:538`.
- Consider renaming the sidebar button (`propertyInputs.js:1429` / `1490`) to something that stays true once a term is carried over, e.g. **"Browse all…"** or **"More icons…"**. "Show all" is now the one label in the flow that is guaranteed to be wrong.

### 2. The search travels, the **Libraries:** filter doesn't

The inline icon picker has five library checkboxes (`propertyInputs.js:1401-1420`). `showAll.onclick` passes `search.value.trim()` and nothing else (`propertyInputs.js:1431`), so the two views disagree exactly where the user just told the app what they wanted:

- Inline, with only **VTT** ticked and `dragon` typed → *"No results."* — [screenshot](https://agent.virtualtabletop.io/reports/ui-review/1537415977996656710/shots/B1-inline-vtt-only-1280.png)
- Click **SHOW ALL** → 20 dragons from Game Icons and Color Emoji, i.e. the four libraries just unticked — [screenshot](https://agent.virtualtabletop.io/reports/ui-review/1537415977996656710/shots/B2-showall-ignores-libraries-1280.png)

That is the worst possible reading of a filter control: the user unchecks four boxes, is told there is nothing, and the very next click contradicts both statements. Since the PR's stated goal is that "the two searches have to mean the same thing", the checkbox state belongs in the handover too. `pickSymbol()` already takes a `type` restriction and `#symbolPickerOverlay` already hides families via `hideFonts`/`hideImages` CSS classes (`client/css/fonts.css:558-562`) and every icon already carries `data-type` — so a per-`data-type` hide class is the same mechanism one level finer. If that is out of scope for this PR, the alternative is at least to say so in the picker ("showing all libraries").

### 3. Category headings disappear the moment there is a search — which is now the default arrival state

`symbols.js:117` hides every `h2` whenever `terms.length > 0`. That was defensible when a search was something the user typed *inside* the picker; now it is how the picker normally opens. The result is an ungrouped grid where a first-time user cannot tell a Game Icon from a color emoji from a VTT card glyph — which matters, because those three render very differently on a widget.

Suggestion: hide an `h2` only when its own section has no visible icon left, rather than whenever a search is active. The 20 dragon results would then read as "Game Icons" (14) / "Emoji – Animals" (3) / "VTT Icons – Mahjong" (3), which is both orientation and free provenance labelling.

### 4. The card reserves full height for however few results there are

The list is `flex: 1` inside a card sized to its host box (`fonts.css:433-439`, `473-483`), so a 20-result opening leaves a large blank card: ~400px of empty white at **1920x1080** ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1537415977996656710/shots/showall-dragon-1920x1080.png)) and ~500px at **820x1180** ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1537415977996656710/shots/showall-dragon-820x1180.png)), with the watermark smiley left stranded in the middle of it. Full height is right when there are thousands of icons to scroll; with the prefiltered opening now the common case, letting the card shrink to its content (`max-height` on the list instead of `flex: 1`, keeping the current floor) would make the filtered picker read as a compact result popover rather than a mostly-empty dialog.

---

## Nitpicks / small breakages

- **The image path is labelled "icon" throughout.** Opened from the **Image** property, the picker is still titled `Pick icon` (`client/room.html:533`), its placeholder still reads `Search icons (sword, heart, dice, …)` (`client/room.html:538`), and its empty state says `No icons match "card_k".` (`symbols.js:126`) — even though `hideFonts` is set and the field you typed into was labelled `Search images...`. [Image picker prefiltered](https://agent.virtualtabletop.io/reports/ui-review/1537415977996656710/shots/img-showall-dragon-1280.png) · [image picker empty state](https://agent.virtualtabletop.io/reports/ui-review/1537415977996656710/shots/img-showall-noresult-1280.png). `pickSymbol()` already knows `type`; setting the `h1`, the placeholder and the noun in `#symbolNoResults` from it is a three-line fix, and this PR is what makes the mismatch visible (you now literally carry a term from a field that says "images" into a dialog that says "icons").
- **On a phone, Show all is a downgrade.** At **390x844** the overlay is 354×221 with the list only 169px tall — about **2 rows** visible and 422px of scroll, while the inline chip list you came *from* showed all 20 matches in 3 rows just above it, with the library filters. [Screenshot](https://agent.virtualtabletop.io/reports/ui-review/1537415977996656710/shots/phone-showall-scrolled-390.png). Nothing is broken, but at that size the button leads to strictly less UI than it left; the shrink-to-content change in #4 plus the count line in #1 would at least make the trade visible.
- **The one line that explains the picker is hidden exactly when the new path is used.** `p[icon=blur_on]` ("Click an icon to pick it…") is hidden by `.fewResults` (`fonts.css:465-467`), and a carried-over term almost always produces <100 matches — so the affordance hint is invisible on the flow this PR introduces. It is also hidden at plain **1280x800** with no search at all, because the container's content box (640×400 minus `20px 10px` padding → 360px) trips the `max-height: 375px` container query at `fonts.css:409`. Worth double-checking that 375px threshold; the hint currently vanishes on a very ordinary laptop window.
- **No clear button on the picker's search field** (`client/room.html:538`) — see #1. With the field now arriving pre-filled *and* pre-selected, "get me back to all icons" should not require a select-all + Delete.
- **`.select()` on touch devices.** `symbols.js:146` focuses and selects the transferred term. On a phone that both raises the on-screen keyboard over an already 221px-tall picker and pops the text-selection handles/context bar over the first row of results. Consider skipping the `select()` (and possibly the `focus()`) when `matchMedia('(pointer: coarse)')` matches, or using `focus({ preventScroll: true })`.
- **`SHOW ALL` sits below the results grid** (`propertyInputs.js:1434`, `1495`). With a 100-chip result set the user scrolls past every chip to reach the button that would show them more. Placing it next to the search input — where "search somewhere bigger" is adjacent to "search" — would put the escape hatch where the eye already is.
- At **820x1180** the picker overlay starts part-way through the sidebar's "Colors" row, so the panel behind it looks *clipped mid-control* rather than covered ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1537415977996656710/shots/showall-dragon-820x1180.png)). Pre-existing overlay behaviour rather than something this PR caused, but this PR makes the symbol picker a much more frequent visitor at that size, so a dimming backdrop would be cheap and would read a lot better.

---

- **"Show all" is the wrong promise.** A newcomer who typed a term and got few chips clicks **Show all** *because* they want the whole catalogue. Landing on 20 icons with no count and no headings is the exact opposite signal — and if they typed something with no matches at all, they land straight on `No icons match "…"`, which is a bewildering thing to get from a button called "Show all". Either the label or the state line from #1 has to give.
- **There is no result count anywhere in the picker**, filtered or not. `20` vs `13288` is the difference between "this is a small library" and "you are looking at a filtered slice", and the handler already computes the number.
- **Nothing labels the provenance of a result** once headings are hidden (#3). "Game Icons", "Material", "Color Emoji", "Mono Emoji", "VTT" are named in the sidebar's `Libraries:` row but never in the big picker — so the two views use different vocabulary for the same taxonomy.
- **`Icon` and `Image` are two adjacent preview boxes in the Content section that now behave identically and open dialogs with the identical title "Pick icon"** ([sidebar](https://agent.virtualtabletop.io/reports/ui-review/1537415977996656710/shots/img-inline-1280.png)). This PR makes them more alike, which raises the cost of the shared title. Distinct titles ("Pick icon" / "Pick image") plus the existing `i` info tooltips would carry most of the explanation.
- The transferred term shows as a **selected (blue-highlighted) string** in the field. That is the right behaviour for "type to replace", but on first sight it can read as an error/validation highlight rather than "your previous search". A count line beside it ("20 matches for your search") removes the ambiguity for free.

Everything above is layout/wording only — correctness is covered by another reviewer. The underlying interaction (term handover, selection, term-by-term matching, the lowercased index making `[card_A]…[card_Z]` findable, and the return trip that leaves the sidebar search intact) all behaved correctly in every case I drove.

VTT-AI-Agent:
Thanks — the headline finding was right and it is fixed in 0300cfc, together with the heading, wording and touch points. The two structural suggestions I did not implement are explained below, and the library-filter question is now with the maintainers.

### 1. The picker opened filtered with nothing saying so — fixed

There is a status line under the search field now, rendered by the same handler that already counted the matches. It only appears while a search is active, states the count *and* the scale, and holds the way back to the whole list:

> 20 of 13288 icons match "dragon" · **[SHOW ALL ICONS]**

- [Show all with `dragon` carried over](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/uifix/showall-dragon-1280.png) — count line, and the headings are back (see 3)
- [after clicking SHOW ALL ICONS](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/uifix/showall-cleared-1280.png) — the field is empty, the status line is gone, the picker looks exactly as it did before this PR
- [empty result](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/uifix/showall-noresults-1280.png) — the count is dropped (the "No icons match …" panel already says it), but the button stays, which is where you want it most

The count comes first in the string on purpose: at 390px the line ellipsizes to `20 of 13288 icons…`, so the part that survives is the part that tells you the list is a slice ([phone](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/uifix/showall-dragon-390x844.png)).

The search field is a `type="search"` input now as well, so the browser draws its own clear ✕ in it.

I left the button labelled **Show all** in the sidebar: with the count line up, "Show all" is no longer the only thing on screen describing the state, and the label is still literally true of what the button does with the *list* (see 2).

### 3. Category headings vanished on any search — fixed

A heading is now hidden only when its own section has nothing visible left, instead of whenever a search is active, so the 20 dragons read as *Game-icons.net – Animal / Defense / Magic / Mythology / Tech / Weapon*, and the emoji and VTT sections label themselves when they match. Same mechanism as before, but the filtering, the counting and the per-section bookkeeping now happen in one pass over the list instead of three `querySelectorAll`s over 13288 elements.

Side effect worth noting for 4: the headings also fill a good part of the blank card you measured — [1920x1080](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/uifix/showall-dragon-1920x1080.png), [820x1180](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/uifix/showall-dragon-820x1180.png).

### The image path called everything an icon — fixed

`pickSymbol()` knows its `type`, so the title, the placeholder, the status line and the empty state now say *image* when it was opened for images: **Pick image** / `Search images (sword, heart, dice, …)` / `17 of 6029 images match "dragon"` / `No images match "card_k".` — [screenshot](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/uifix/image-showall-dragon-1280.png). The wording resets for the next picker, which the image test asserts. That also answers the "two adjacent boxes open the identical dialog" point: **Icon** and **Image** now open dialogs with different titles.

### `.select()` on touch devices — fixed

Guarded with `matchMedia('(pointer: coarse)')`: on a coarse pointer the term is left unselected (the field is still focused, as before this PR), so the selection handles and their context bar do not land on the first row of results.

### 2. The Libraries: checkboxes do not travel — asked the maintainers

I have gone back and forth on this one (it was also finding 3 of the code review) and it is a design call rather than a bug, so I have put it to the maintainers rather than deciding it a third time myself. The state of it:

- The five inline types (`game-icons`, `material-symbols`, `emoji-color`, `emoji-monochrome`, `vtt-symbols`) can be carried over exactly — every icon in the big list already has `data-type`, so it is the same mechanism as `hideFonts`/`hideImages`, one level finer. It is not the lossy fonts-vs-images mapping I described in my earlier reply; that was too pessimistic.
- What it costs is the button's promise: **Show all** would then show *fewer* icons than the list it came from, which needs a rename (your "Browse all…" / "More icons…") and makes the big picker a place you can arrive at with four of five libraries switched off and no control to switch them back on — it has no Libraries row of its own.

In the meantime the restored headings make the current behaviour legible rather than mysterious: narrowing to VTT, getting "No results.", then clicking Show all now lands on results with *Game-icons.net – Animal* written over them, so it is visible where they come from and that the big list is not filtered by library.

### 4. The card reserves full height for few results — not changed

Partly addressed by the headings above. Beyond that I did not want to touch it in this PR: `#symbolList { flex: 1 }` and the `min-height` floors around it are load-bearing for the container-query layout (the list is the only scroller, and `div[icon]` clips), and the deck editor parks the same overlay in a completely different box. That is a change worth making on its own, with its own screenshots at all four sizes, not as a rider on a search-term handover.

Same for the two other layout points: **SHOW ALL sitting below the chips** in the sidebar picker is unchanged (it reads as "did not find it in these? here is everything", and moving it next to the search field changes the inline picker's layout for both pickers), and the **missing backdrop** at 820x1180 is pre-existing overlay behaviour, as you say.

### The hint hidden by `.fewResults`

Still hidden, but the space it used to occupy now carries the status line, which is the more useful thing to have there on the filtered path. The `max-height: 375px` container-query threshold that also hides it at 1280x800 is pre-existing and independent of this PR — worth a look, but not something I want to retune from inside this change.

### Verified

- TestCafe `tests/testcafe/editor.js`: 47 passed. The two handover tests now also assert the status line text (`/^\d+ of \d+ icons match "dragon"$/`), that headings survive a search while non-matching ones do not, that **Show all icons** empties the field and unhides every icon and heading, and the full image wording (`Pick image`, `… images match …`, `Show all images`, `No images match "card_k".`) plus its reset back to `Pick icon`.
- `npm test` (jest): 1265 passed.
- Driven with Playwright at 390x844, 820x1180, 1280x800 and 1920x1080 (screenshots linked above), through the icon path, the image path, the cleared state and the empty-result state.

VTT-AI-Agent (commented):
## TL;DR

The follow-up fixes make the term handoff, matching, status, category headings, and image-mode wording coherent, but the newly added native search-field clear control does not actually re-run the filter. Clicking its × leaves the picker showing stale filtered results and stale status text, so this needs one event-wiring fix.

## Findings

### 1. The native search clear button leaves the filter and status stale — `client/js/symbols.js:111`

This PR changes the field to `type="search"` in `client/room.html:538`, explicitly adding the browser's clear ×, but the picker still updates only through `onkeyup`. Activating a search input's native clear control changes its value and emits input/search events; it does not emit a keyup. As a result, after opening with `dragon` (or typing it) and clicking ×, the field becomes empty while the old `.hidden` classes, `.filtered`/`.fewResults` state, category visibility, and `… match "dragon"` status all remain in place. The separate **Show all icons** button works only because it manually calls `onkeyup()`.

Wire the filter to `input` instead (as the inline pickers already do), and call the shared handler directly from the programmatic-clear/open paths. That also fixes mouse-paste, cut, drop, and other non-keyboard edits. A regression assertion can set/clear the field and dispatch `input`, then verify the status disappears and all icons/headings return.

## Scope checks

No widget property, routine operation, engine path, or `library/` file is changed, so no authoring-control, routine-block, validator, tutorial, or demo-room addition is required. All reported CI checks, including both `editor.js` browser jobs and CodeQL, are green.

The `beta` label was already present; I left it in place because this edit-mode-only UI change touches no engine paths and the remaining clear-control issue cannot corrupt games or crash the server.

VTT-AI-Agent:
Thanks — the finding was right and it is fixed in a0255a54.

### 1. The native clear button left the filter and status stale — fixed

Reproduced exactly as described before touching anything: opened the picker with `dragon` carried over, clicked the browser's × and read the picker's state out of the DOM.

```
after Show all: value="dragon" status='20 of 13288 icons match "dragon"' visible icons=20 hidden=13268 hidden headings=79
after native × : value=""       status='20 of 13288 icons match "dragon"' visible icons=20 hidden=13268 hidden headings=79
```

An empty field, 20 icons and a status line still naming a search nobody was running any more.

The handler is a named `filterSymbolList()` now and the field listens for `input` instead of `keyup`, so the native ×, paste, cut and drop all filter — the same event the inline pickers use (`propertyInputs.js:1437`, `1498`). The three programmatic paths (`pickSymbol()`, the richtext `add_reaction` picker and the **Show all icons** button) call `filterSymbolList()` directly instead of reaching through `input.onkeyup()`, which also stops the "the button only works because it calls the handler by hand" coupling you noted. Same state after the fix:

```
after native × : value="" status="" visible icons=13288 hidden=0 hidden headings=0
```

[Screenshot after clicking the native ×](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/uifix/showall-nativeclear-1280.png) — empty field, no status line, the whole list and every heading back.

One line came with it: the status text is now only built while a search is active (`terms.length && matches`). Before, clearing the field left the invisible span holding `13288 of 13288 icons match ""` — never displayed, since the `.filtered` class is what shows the line, but it made "cleared" indistinguishable from "filtered" when read from the DOM, and it is nonsense text to leave lying in the page.

### Regression test

`tests/testcafe/editor.js` — the icon handover test now clears the field the way the browser's clear button does (`value = ''` plus a bubbling `input` event, since the × lives in the shadow DOM and TestCafe cannot click it) and asserts the status line is gone and that no icon and no heading is left hidden. Verified it is a real regression test: with `oninput` put back to `onkeyup` it fails on the first of those assertions.

### Verified

- TestCafe `tests/testcafe/editor.js`: 47 passed.
- `npm test` (jest): 1265 passed.
- Playwright against the branch at 1280x800: the native × drives the picker from 20 filtered icons back to all 13288, headings and all.

VTT-AI-Agent:
The maintainers answered the open question in the Discord "Human feedback" thread:

> **Q:** In PR #3119, should the symbol picker's "Show all" button carry the inline icon picker's "Libraries:" checkboxes over, or keep showing every library?
>
> **arnoldsmith86:** Carry the library filter over and rename the button (e.g. "Browse more…")

Done in 47bf052, and the ⏳ To Do block is off the description.

### The libraries travel with the term

The five inline types map onto the `data-type` every icon in `#symbolList` already carries (`material-symbols` covers its `-nofill` twin, `vtt-symbols` is the picker's `symbols`), so the big picker hides by library the same way it hides by search term — one pass, no extra DOM queries. `pickSymbol()` takes the chosen libraries as a fifth argument; everything else that opens the picker (JSON editor, deck editor, the routine editor's icon fields, the richtext toolbar) passes nothing and is unfiltered exactly as before. All five ticked is treated as no filter.

The reviewer's scenario, driven on the branch: with **Game Icons** and **Color Emoji** unticked and `dragon` typed, the sidebar shows 3 chips — and **Browse more...** now opens on the same 3 icons instead of 20 from the two libraries just switched off ([screenshot](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/uifix/showall-libraries-1280.png)).

### Saying so, and getting back out

The big picker has no library checkboxes of its own, so the status line names the restriction and the button next to it drops it:

> 3 of 13288 icons from 3 of 5 libraries match "dragon" · **[SHOW ALL ICONS]**

`Show all icons` now clears the search *and* the libraries, which is the whole list back in one click. With no search but a library filter the line reads `7259 of 13288 icons from 3 of 5 libraries`, and unticking every library gets `No icons in the chosen libraries.` instead of a blank card. Restored category headings do the rest: results say which library they came from.

### The button is "Browse more..."

Both sidebar buttons (icon and image) are relabelled — the big picker is now reached *with* your filters, so "Show all" was the one label in the flow guaranteed to be wrong. "Show all icons"/"Show all images" stays on the button inside the picker, where it is literally true.

### Verified

- TestCafe `tests/testcafe/editor.js`: 48 passed. New test **"The inline icon picker hands its chosen libraries to the symbol picker"** unticks Game Icons, checks no `data-type=game-icons` icon survives the handover, asserts the status line (`/^\d+ of \d+ icons from 4 of 5 libraries match "dragon"$/`), that `Show all icons` unhides all 13288, and that the *next* picker — opened from the image property — is unfiltered again. The icon handover test also asserts the new button label.
- `npm test` (jest): 1265 passed.
- Playwright at 1280x800 through the unticked-libraries flow, the status line and the way back.

VTT-AI-Agent (commented):
## TL;DR

Good change — the handover, the term-by-term matching and the lowercased keyword index all line up with what the inline picker does, and the status line is the right answer to "the picker now usually opens filtered". One real bug: the new library filter keys off `data-type`, but the 269 flag emoji carry `data-type="emoji-color emojiFlag"` (two class names in one attribute), so **every flag disappears from the picker as soon as any library checkbox is unticked** — even with "Color Emoji" still ticked.

## Findings

### 1. The library filter drops all flag emoji (`client/js/symbols.js:105`, `:165`)

`client/js/symbols.js:102-105` builds the flag category with

```js
let className = 'emoji-color';
if(category == 'Emoji - Flags')
  className += ' emojiFlag';
list += `<i class="${className}" ... data-type="${className}" ...`
```

so those 269 icons have `data-type="emoji-color emojiFlag"` — a class list, not a type. The new filter at `client/js/symbols.js:165` does an exact lookup:

```js
|| libraryFilter && !libraryFilter.types.has(element.dataset.type)
```

`symbolLibraries['emoji-color']` is `[ 'emoji-color' ]` (`:127`), so `has('emoji-color emojiFlag')` is always false.

Repro: open a widget's icon picker, untick just **Game Icons** (leave Color Emoji ticked), type `flag`, click **Browse more...**. The `Emoji - Flags` section is empty and its heading is hidden by the new per-section heading logic, while the status line says `N of 13288 icons from 4 of 5 libraries match "flag"` — with the flags counted in the 13288 denominator but silently excluded from the "4 libraries". This is the only way to reach flag emoji at all in the editor: `loadIconSearchIndex()` skips `Emoji - Flags` (`client/js/editor/propertyInputs.js:337`), so the inline picker never offers them and the big picker was the fallback.

Fix: emit `data-type="emoji-color"` and keep `emojiFlag` in `class` only — the CSS that needs it (`client/css/fonts.css:557`) keys off the class, not `data-type`.

That one-line change also fixes a pre-existing bug in the same area, which this PR makes easier to hit now that "Browse more..." carries a search over: the click handler at `client/js/symbols.js:240-243` compares `icon.dataset.type == 'emoji-color'`, so a flag resolves to `isImage: false, url: null`. Picking a flag from the **image** picker (flags are `data-family="image"`, so they survive `hideFonts`) runs `this.setValue(symbol.url)` at `client/js/editor/propertyInputs.js:1496` and clears the image instead of setting it; `client/js/jsonedit.js:401` inserts `null` the same way.

### 2. The library test can't catch the above (`tests/testcafe/editor.js:1241`)

The new library-handover test only asserts that nothing from the *unticked* library survives:

```js
.expect(Selector('#symbolList i[data-type=game-icons]:not(.hidden)').count).eql(0)
```

Nothing asserts that the *ticked* libraries came through intact, which is exactly where the bug is. Worth adding the mirror assertion — e.g. after unticking Game Icons, `Selector('#symbolList i.emojiFlag:not(.hidden)').count` > 0 (searching `flag`), plus one non-flag `emoji-color` — so a library that silently loses a whole section fails the suite.

### Minor

- `client/css/fonts.css:447`: the `min-height: 210px` on `#symbolPickerOverlay div[icon]` is documented as "its own padding, the search field, and the list's floor below". `#symbolSearchStatus` adds ~40px inside that clipped box, and filtered is now the *normal* opening state, so in the shortest layout (`@container symbolPicker (max-height: 375px)`, where `#symbolList` still has `min-height: 120px`) the bottom of the list may end up clipped. Worth a quick look at ~375px height with a term carried over; if it does clip, a `.filtered` variant of that min-height is the smallest fix.
- `type="search"` only gets a native clear ✕ in Chromium/WebKit — Firefox draws none. The `oninput` switch is right regardless (paste/cut/drop), and "Show all icons" covers the way back there, so this is only about the rationale in the PR description, not the behavior.
- `#symbolSearchStatus` has no `aria-live`, so the count that replaces the missing "this list is a slice" signal is silent for screen readers. A `aria-live="polite"` on the span would announce it on each filter.

Scope checks: no new widget property or routine operation, so validator / `jeCommands` / demo-room requirements don't apply, and the diff touches nothing under `library/`.

The `beta` label was already on this PR and I left it there — nothing here touches the engine (no `client/js/widgets/`, `client/js/compute.js` or `client/js/serverstate.js` changes; it is edit-mode picker UI, CSS and tests only), and the flag-emoji bug is a visibility regression in a picker, not something that can corrupt a game or crash the server.

VTT-AI-Agent:
Thanks — the flag bug was real, in both halves you described, and it is fixed in e370a217.

### 1. The library filter dropped all 269 flag emoji — fixed

Confirmed exactly as written: the flags carried `data-type="emoji-color emojiFlag"`, so `libraryFilter.types.has(...)` never matched them and unticking *any* library hid the whole `Emoji - Flags` section while still counting it in the 13288 denominator. `data-type` is the library now (`emoji-color`), and `emojiFlag` stays on `class`, which is where `client/css/fonts.css:557` reads it. The `title` says `emoji-color: 🇦🇩` instead of `emoji-color emojiFlag: 🇦🇩` for the same reason — the two class names were never a type.

Same change fixes the pre-existing `url: null` you spotted: picking a flag from the **image** picker used to run `setValue(null)` and clear the image. Reproduced before touching anything — with the fix reverted, the new assertion fails with `expected '' to match /^\/i\/noto-emoji\/emoji_u[0-9a-f_]+\.svg$/`, i.e. the image really was being cleared.

[Screenshot: Game Icons unticked, `flag` carried over, all 269 flags present](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/uifix/flags-with-library-filter-1280.png) — `303 of 13288 icons from 4 of 5 libraries match "flag"`, with the `Emoji - Flags` heading visible.

### 2. The library test never checked the ticked libraries — fixed

The library-handover test searches `flag` now, so it asserts both directions: nothing from the unticked Game Icons survives, **and** the libraries left ticked arrive whole —

```js
.expect(Selector('#symbolList i.emojiFlag:not(.hidden)').count).gt(0)
.expect(Selector('#symbolList i[data-type=emoji-color]:not(.emojiFlag):not(.hidden)').count).gt(0)
.expect(Selector('#symbolList i[data-type=material-symbols]:not(.hidden)').count).gt(0)
```

Verified as a real regression test: with the fix reverted it fails on the first of those with `expected 0 to be above 0`. The image test additionally picks a flag and asserts the widget's `image` ends up a noto-emoji URL, which covers the click-handler half.

### Minor

- **`min-height: 210px` vs the status line** — added a `.filtered` variant at 244px. I could not actually produce a clip: in the room-overlay host the box never dropped below 250px at any viewport I drove it to (down to 900x200 and 390x260), and the list stayed above its 120px floor. But the floor's comment budgets exactly "its own padding, the search field, and the list's floor below", and with the status line up that budget is 34px of line plus 20px of margins minus the 20px the search field gives up — 244px. Leaving the number describing a box it no longer holds seemed like the thing that bites the deck editor host later, so the floor now says what it means.
- **`aria-live`** — `#symbolSearchStatus span` has `aria-live="polite"`. The class toggle that shows the line already runs before the text is written, so the region is rendered when its content changes.
- **`type="search"` in Firefox** — right, no native ✕ there. The `oninput` switch stands on its own (paste, cut, drop) and **Show all icons** is the way back in every browser; I have taken the native-✕ claim out of the PR description's rationale so it does not read as the reason the field is a search input.

### Verified

- TestCafe `tests/testcafe/editor.js`: 48 passed.
- `npm test` (jest): 1265 passed.

ArnoldSmith86 (approved):

VTT-AI-Agent:
> make sure (doing screenshots) that the margin between icons and "browse more.." button is the same as the margin between icons and line above them (the button currently sits too close to the list)

(feedback from ArnoldSmith86, typed into the bot's open-PR overview page)

Measured it in the browser, and it was exactly that: the icon list keeps **8px** to the separator line under the **Libraries:** checkboxes, but only **1px** to the **Browse more...** button below it (the 1px is the button's default margin — nothing ever spaced it from the list). The button now carries a `propertyPickerBrowseMore` class with `margin-top: 8px`, so both gaps are 8px; the image picker's button gets the same class, so it is spaced from its grid too.

Verified with `getBoundingClientRect()` on the rendered picker at 1280x1000, 1920x1080 and 1100x900 — `line → icons` and `icons → button` both report 8px in each (before: 8px and 1px).

| before | after |
| --- | --- |
| before | after |

Also at [1920x1080 before](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/spacing/icon-picker-before-1920x1080.png) / [after](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/spacing/icon-picker-after-1920x1080.png), and the [image picker](https://agent.virtualtabletop.io/reports/pr/1537390081134956595/spacing/image-picker-after-1280x1000.png).

`npm test` (1265 passed) and TestCafe `tests/testcafe/editor.js` (48 passed) still green.

VTT-AI-Agent:
Feedback from Discord ([source](https://discord.com/channels/770758631146782780/1537390081134956595/1542863890184937533)):

> "Choose icon" inside the flyout (when no icon is picked) is always on bright background but changes font to white in dark mode, making it unreadable

Reproduced and fixed. The placeholder chip that says **Choose icon** in the picker's summary line is a `.propertyValueChip`, whose background is a hardcoded white checkerboard, while the label on it takes `color: var(--textColor)` — `#f8f8f2` in dark mode. So it really was near-white text on a white tile (measured in the browser: `color: rgb(248, 248, 242)` on `background: rgb(255, 255, 255)`).

The labelled placeholder chip now paints the same themed checkerboard the Content section's **Choose icon** / **Choose image** tile already used (`--backgroundHighlightColor1` over `--backgroundColor`), so it follows the theme like its text does — dark tile in dark mode, and unchanged in light mode. That tile's rule can therefore stop repeating the background. The same chip is used by the image and sound pickers and by icon rows outside the Content section, so those are fixed with it (checked in the browser: `Choose sound` now measures `#f8f8f2` on `rgb(8, 9, 10)`).

| | before | after |
|---|---|---|
| dark | before | after |
| light | before | after |

`npm test` 1265 passed and TestCafe `tests/testcafe/editor.js` 48 passed.

PR #3121, opened 2026-08-15
Show images that are not SVGs even when svgReplaces is set

VTT-AI-Agent:
## What this changes

A widget whose `image` is **not** an SVG (a PNG, a JPEG, a GIF …) but that also has an `svgReplaces` property showed **nothing at all**. This fixes that: such an image is now simply displayed, while SVGs keep being replaced exactly as before.

## Why it was broken

`svgReplaces` maps a string inside an SVG file (`#primaryColor`, a colour, a stroke width) to a widget property whose value is substituted for it. `getSVG()` in `client/js/main.js` implemented that by fetching the image **as text**, running the replacements over it and returning a `data:image/svg+xml,…` URL.

It did that for every image it was handed, without ever checking that the file actually is an SVG. For a PNG that means the raw bytes of the image end up inside an `image/svg+xml` data URL, which the browser cannot decode — so the widget's `background-image` resolves to nothing and the widget looks empty. Adding an `svgReplaces` entry (the JSON editor's "SVG colors" button does that for you) to a widget with a bitmap image therefore made the image disappear, with no error anywhere.

## The fix

`getSVG()` now looks for an `<svg>` element in the file it loaded. If there is none, the URL is remembered as "not an SVG" and returned as it is (mapped through `mapAssetURLs()`, like every other image URL) instead of being wrapped into a data URL — for that first call and for every later one, without fetching the file again. Everything that goes through `getSVG()` benefits: image/basic widgets, holders, dice faces, timers, card face objects and the symbol/icon colouring.

An image that cannot be read at all falls back to the same plain URL, which is all the browser needs: a CORS-blocked URL still displays fine as a `background-image`. `fetch()` **rejects** for an external URL blocked by CORS (`image: "https://…/board.png"`), and the old chain had no `.catch()`, so the queued callbacks never fired and the widget stayed blank for the rest of the session — the very symptom this PR is about — while the unhandled rejection also raised the blocking client-error overlay.

Unlike "loaded, and it is not an SVG", though, a file that did not arrive is no answer about the file: a server restart, a 503 or a hiccup in the network says nothing about what it contains. That verdict is therefore kept in its own cache and expires after 30 seconds, so the next widget asking for the image tries again — long enough not to refetch a CORS-blocked image on every CSS recomputation, short enough that one missed request no longer costs a perfectly good SVG its replacements for the rest of the session. While a retry is in flight the widget keeps displaying the URL it already has rather than blinking to nothing.

Deciding whether a file is an SVG lives in a single place, `fetchSVG()`, which the **SVG replacements editor** (`client/js/editor/sidebar/properties.js`) now calls too instead of repeating the check itself. The two used to disagree about a namespace-prefixed root element (`<svg:svg …>`), so the editor would offer replacements for a file the engine refused to replace anything in; and the editor's shortcut of believing the `Content-Type` of a bitmap rather than decoding a whole PNG as text now benefits the engine as well.

This is rendering only — no widget property changes meaning, no state is affected, so no migration or legacy mode is involved.

## Telling the user in the editor

Once the widget stops going blank, nothing else tells a user that the replacements they set up cannot fire, so the **SVG replacements** panel now says it. `fetchSvgReplaceCandidates()` reports what the image turned out to be - an SVG, a bitmap, a file that could not be loaded, or no image at all - instead of only whether it is an SVG:

- a bitmap gets a warning line (*"This image is not an SVG, so these replacements have no effect. They apply again as soon as this widget points at an SVG file."*) and a disabled `Add replacement`, which used to invent one dead `#000 -> color` mapping per click;
- an image that could not be loaded, an SVG whose values are all mapped, and an SVG with nothing replaceable in it each say so where the swatch list would be, instead of all three rendering as the same blank space;
- the row inputs are 34px like the buttons next to them (they were 21px), the two unlabelled inputs get `in the SVG` / `widget property` headers, and the decorative arrow is `aria-hidden`.

`jsonedit.js` was the third place sniffing SVGs - `/svg/i` anywhere in a decoded file, which would have offered its colour picker for a bitmap that happens to contain those bytes. Its `checkIfSVG()` calls `fetchSVG()` too now, and `fetchSVG()` memoizes its request per file, so the engine, the properties sidebar and the JSON editor share one request and cannot disagree about a file — verified in a browser: the SVG is pulled over the wire once while all three look at it.

## Before / after

The same board, on `main` and on this branch — the left panel is a PNG that also has `svgReplaces: { "#primaryColor": "color" }`:

| before (`main`) | after (this PR) |
| --- | --- |
| before | after |

## Demo board

That board is uploaded to this PR's test server as **"svgReplaces on a non-SVG image"** — open the `pr-test-ai` room, pick it from the Game Shelf and press Play. It shows a PNG with `svgReplaces`, the same PNG without it as a control, and an SVG with `svgReplaces`; the four colour buttons set the same `color` property on the PNG and the SVG, so you can see the SVG recolour while the PNG simply stays visible. Its typography was fixed after a UI review: every label carried a `fontSize` property, which neither the label widget nor the validator knows (the size belongs in `css`), so the whole board rendered at the default 16px and the title looked no bigger than the card headings. It is a demo board only — it is **not** part of this PR's diff and nothing under `library/` is touched. It can be added to `library/tutorials/` on request.

Direct download: <https://agent.virtualtabletop.io/reports/pr/1538043964047753277/svgReplaces%20on%20a%20non-SVG%20image.vtt>

## Testing

- New jest tests in `tests/client/get-svg.test.js`: an SVG still gets its replacements applied, including one with a namespace-prefixed root element; a non-SVG is returned as its own URL and is not fetched a second time; a bitmap is recognized from its content type without being read as text; an image whose fetch rejects and one the server answers with a 404 both end up with their plain URL, and an unreadable one is tried again once the retry delay is over while a file that loaded and turned out not to be an SVG is not; the JSON editor's `checkIfSVG()` recognizes exactly the files the engine can replace in.
- New jest test in `tests/client/property-inputs.test.js`: the four states an image can be in (`svg` / `notSvg` / `unreadable` / `none`) are told apart, and an unreadable file still falls back to its file name.
- `npm test` — 1276 tests pass.
- TestCafe `editor.js` (covers the `svgReplaces` editor), `domsnapshot.js` and `pixelsnapshot.js` pass locally on Chromium.
- Checked in a real browser at 1280×800 and 1180×820 with one image widget per state side by side: the warning line and disabled button on the PNG, the swatches on the SVG, and the "could not be loaded" line on a broken `.svg` URL and on the CORS-blocked one - no page errors in any of them.
- Checked in a real browser that aborting only the *first* request to an SVG leaves the widget showing the plain image, does not refetch during the retry delay, and picks the replacements up again on the next widget rebuild past it.
- Checked in a real browser, with `minifyJavascript` both off and on: a PNG, an SVG, an unreachable external URL and a plain PNG side by side all render (the SVG recoloured, the others as their own URL), no client-error overlay; and in edit mode the SVG replacements section still offers the swatches read from the SVG and none for the PNG.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3121/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

Good, well-scoped fix in the right place — `getSVG()` really is where this belongs, and the two new jest tests cover the happy paths. The one thing I'd still change before merge: the `fetch()` chain has no `.catch()`, so an image whose fetch *fails* (cross-origin URL blocked by CORS, offline asset) is still stuck showing nothing — and it now also raises the blocking client-error overlay. Second, smaller point: the new `<svg>` sniffing disagrees with the one the editor already implements for the same question.

## Findings

### 1. A failing fetch is still an invisible widget — and now an error overlay (`client/js/main.js:708-717`)

The promise chain has no `.catch()`. If `fetch()` **rejects** rather than returning a non-SVG body, none of the new code runs: `svgCache[url]` stays the pending array forever, the queued callbacks never fire, and every later `getSVG()` for that URL keeps returning `''`, so the widget renders `background-image: url("")` for the rest of the session. That is exactly the symptom the PR sets out to fix, for a case that is not rare: a game author who puts an external URL in `image` (`https://…/board.png`) gets a CORS-blocked `fetch()` even though `background-image` would have loaded that URL fine.

I checked this against this branch by mocking a rejecting `fetch()`:

```
first return:  ""
callback got:  CALLBACK NEVER FIRED
second return: ""     <- permanently stuck, no refetch
```

There's a second half to it: the rejection is unhandled, and `client/js/tracing.js:194` routes unhandled rejections into `errorHandler()` → `showClientError()`. So on `main` such a board silently shows nothing, and with this PR it shows nothing *and* interrupts the session with the "report this error" overlay. That's a behavior change in the wrong direction for the exact scenario the PR targets.

The cheapest fix reuses the branch you just added — turn a failed fetch into an empty body, which contains no `<svg` and therefore falls into the non-SVG path by itself:

```js
fetch(mapAssetURLs(url)).then(r=>r.text()).catch(_=>'').then(t=>{
```

Worth a third test next to the two new ones (rejecting fetch → callback fires with the plain URL, no second fetch).

### 2. The SVG check diverges from the one the editor already has (`client/js/main.js:710`)

`fetchSvgReplaceCandidates()` in `client/js/editor/sidebar/properties.js:260-288` answers the identical question about the identical file, and answers it differently:

* it checks `Content-Type` first and `body.cancel()`s a bitmap instead of pulling it through `r.text()` — the asset route sets that header from the first byte (`server.mjs:172-181`), so it's reliable for `/assets/*`;
* it falls back to the file name when the file can't be read (the case of finding 1);
* it tests `/<svg/i`, not `/<svg[\s>]/i`.

That last difference is a real, if uncommon, regression rather than a style nit — the two regexes disagree on a namespace-prefixed root, which some exporters produce:

```
"<svg:svg xmlns:svg=\"..\">"   PR: false   editor: true
"<svg/>"                       PR: false   editor: true
```

For such a file the editor keeps showing the "SVG replacements" section and its swatches (it considers the file an SVG), while the engine now decides it isn't one and stops substituting — i.e. a board where `svgReplaces` works today would stop recolouring after this PR. Switching to `/<svg/i` costs nothing and removes the divergence; adopting the content-type shortcut as well would also stop a multi-MB PNG being downloaded and UTF-8-decoded in full just to conclude it has no `<svg` in it (the `/assets/` route sets a long `Cache-Control`, so the *second* load is cheap, but the stringify isn't).

Best of all would be a single helper both sides call — the editor bundle already reaches main.js through the `Object.assign(window, {…})` in `loadEditMode()`, so it can live next to `getSVG()` and let `fetchSvgReplaceCandidates()` drop its own copy.

### 3. Note: the new early return is the one path that never calls the callback (`client/js/main.js:683-684`)

`if(nonSVGCache[url]) return mapAssetURLs(url);` returns before the callback is ever registered or invoked. Every current caller (`imagewidget.js:91`, `timer.js:85`, `dice.js:192`, `card.js:143`, `symbols.js:433-434`) uses the return value, so nothing breaks — but `getSVG()`'s contract is now "the callback always fires, except for a URL already known not to be an SVG", which is easy to trip over later. A one-line comment on the early return, or just calling `callback(url)`, would keep it uniform.

* `export function getSVG` is safe in the client bundle: `removeExportStatements()` in `server/minify.mjs` strips `export` before `function`, and main.js already exports plenty.
* Returning `mapAssetURLs(url)` matches what `dice.js:187` already does for a plain image, so the value is consistent with the non-`svgReplaces` path.
* Full jest suite passes locally on this branch (32 suites, 1267 tests), including the two new ones; CI is green including CodeQL.
* No new widget property or routine function is introduced, so there is nothing new that would need editor-UI or `jeCommands`/validator support.
* Nothing under `library/` is touched and the demo board lives on the PR test server — correct for this repo's rules.

I added the `beta` label: the diff touches only `client/js/main.js` (none of the engine files excluded from the beta build — `client/js/widgets/`, `compute.js`, `serverstate.js`) and is rendering-only with no state, sync or routine-semantics impact, so it is safe and useful for players to try even with the findings above open.

VTT-AI-Agent:
Thanks — both findings were real and both are fixed in fb232368.

**1. A failing fetch is still an invisible widget.** Confirmed and fixed, though not with the one-word version: the chain now goes through the shared `fetchSVG()` helper and a failed load is turned into "not an SVG", so the widget falls back to the plain URL exactly like a PNG does.

```js
fetchSVG(url).catch(_=>null).then(t=>{
```

A rejecting fetch therefore fires the callback with `mapAssetURLs(url)`, caches that decision so the file isn't fetched again, and never leaves an unhandled rejection for `tracing.js` to turn into the error overlay. New test: *"an image that cannot be loaded at all is used as it is"*, which asserts both the returned URL and that there is no second fetch. Note that `fetchSVG()` also rejects on a non-OK response now, where the old code fed the 404 body into the `<svg` test — same outcome, but no longer by accident.

**2. The `<svg>` sniff diverges from the editor's.** Agreed that this was a regression rather than a nit, and rather than aligning the regex I took your "best of all" option: the whole decision now lives in `fetchSVG()` in `client/js/main.js` (exported to edit mode through the `Object.assign(window, {…})` list), and `fetchSvgReplaceCandidates()` calls it instead of carrying its own copy. So the engine and the editor can no longer disagree about a file, `/<svg/i` is the one test, and the engine picked up the editor's `Content-Type` shortcut — a PNG is now recognized from its header and its body cancelled instead of being decoded as text. `fetchSvgReplaceCandidates()` keeps its own file-name fallback for a file it can't read, which is why `fetchSVG()` rejects on that case instead of returning `null`: "unreadable" and "not an SVG" are different answers and each caller wants a different thing from them. Tests for the namespace-prefixed root and for the content-type shortcut are in the same file.

**3. The early return never calls the callback.** True, but it is not new: the `typeof svgCache[url] == 'string'` branch right below it returns the replaced SVG the same way without touching the callback — the callback exists only to tell the widget that a *pending* fetch has arrived. Calling it from the early return would actually break things: `imagewidget.js:91` passes `_=>this.domElement.style.cssText = this.css()` and `css()` is what calls `getSVG()`, so a synchronous callback would recurse forever. I added a comment on the early return saying so instead.

**Verification.** `npm test` — 1270 tests pass (32 suites). TestCafe `editor.js` — 45/45 pass locally on Chromium. And in a real browser against a room with a PNG, an SVG, an unreachable external URL and a plain PNG side by side, with `minifyJavascript` both `false` and `true`:

```
png       url("http://localhost:8333/i/branding/android-512.png")
svg       url("data:image/svg+xml,%3Csvg%20xmlns%3D%22http%3A%2F%2F…")
external  url("https://example.invalid/board.png")
plain     url("http://localhost:8333/i/branding/android-512.png")
client error overlay: none
page errors: none
```

and in edit mode the SVG replacements section still reads its swatches out of the SVG (`#fff`) and offers none for the PNG or for the unreachable image — the editor side of the shared helper, including the minified bundle where `fetchSVG` has to survive terser as an `Object.assign` key.

Nothing under `library/` is touched and the demo board on the test server is unchanged.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review) of #3121. The visual result of the fix is right: a PNG with `svgReplaces` now renders instead of disappearing, SVGs still recolour, and nothing else in the app regressed at any viewport I tried. The one impactful gap it leaves open is in the **SVG replacements** panel: now that the widget no longer goes blank, *nothing at all* tells a user that the image they pointed the replacements at is not an SVG — the panel still shows editable rows, an "Add replacement" button and live colour pickers that silently do nothing. Add an inline "this image is not an SVG" notice there (the code already knows via `imageIsSvg`).

<details>
<summary>What I actually exercised</summary>

Ran the branch locally (`npm ci`, `minifyJavascript: false`) with a scratch room containing four image widgets side by side — a PNG **with** `svgReplaces`, the same PNG **without** it as a control, an SVG with four replacements, and an unreachable external URL with `svgReplaces`.

Overlays opened and eyeballed at **390×844 (phone portrait)**, **844×390 (phone landscape)**, **820×1180 (tablet portrait)**, **1180×820 (tablet landscape)**, **1280×800 (laptop)** and **1920×1080 (desktop)**: the room/toolbar, Game Shelf (`statesOverlay`), Players, About, edit mode (`Edit widgets` panel), the widget properties sidebar including the **SVG replacements** section, the Add-widget module, the Assets module, the symbol/icon picker, and the JSON editor with its command list. No overflow, clipping, collision or unreachable control that this PR introduces at any of those sizes; the icon picker still renders every icon (it goes through `getSVG()` via `symbols.js:433`), and no client-error overlay fires for the CORS-blocked image any more.

Not reachable, and not this PR's doing: edit mode is deliberately unavailable in **phone portrait** and **tablet portrait** ("Please rotate your device to landscape orientation"), so the SVG-replacements panel could only be judged from 1180×820 upwards. Overlays behind a second player, share/publish and the deck editor's sub-dialogs I did not open — the diff has no plausible reach into them.

Verified on the live PR server too: in `pr-test-ai` the PNG panel's computed `background-image` is a plain `url(.../assets/-330709127_12124)` while the SVG panel's is a `data:image/svg+xml,…`, i.e. exactly the intended split.

board
</details>

---

## Big layout improvements

### 1. The SVG replacements panel needs to say when the image isn't an SVG

This is the one that matters. Before this PR, pointing `svgReplaces` at a bitmap produced brutal but *informative* feedback — the widget vanished, so you knew something was wrong. After the fix the widget renders correctly, which is right, but it also means the setup is now **completely silent**: the panel is still shown (`properties.js:6938` keeps it visible whenever replacements exist), it still lists `#primaryColor → color`, it still renders a live colour picker underneath, and clicking that picker changes nothing at all, forever, with no explanation.

The two panels below are the same widget type, same section, at 1280×800 — one on a PNG, one on an SVG. There is no visual difference whatsoever:

| PNG + `svgReplaces` (replacements are dead) | SVG + `svgReplaces` (replacements work) |
| --- | --- |
| png | svg |

It gets worse if you keep going. `Add replacement` on a bitmap falls through to the "no file to read" branch (`properties.js:6884-6896`) and happily invents `#000 → color2`, then `#color2 → color3`, each with its own colour swatch — I clicked it twice and got three dead mappings and three dead pickers, while the image never changed:

add on png

Concrete proposal, in the panel's existing style:

- `renderSvgReplacesEditor()` already computes `imageIsSvg` and already re-runs `updateVisibility()` whenever the image or the map changes. In the `!imageIsSvg && stillHasReplaces` case, render one notice line at the top of the section body — same place and styling as the existing `.svgReplacesHelp` line (`properties.js:6715`), ideally with the app's warning colour rather than `--textHelpColor` — saying something like: *"This image is not an SVG, so these replacements have no effect. They will apply again if you point this widget at an SVG file."*
- In that same state, disable `Add replacement` (`properties.js:6884`) with a `title` explaining why, instead of letting it insert placeholder mappings that can never do anything.

That turns the current dead-end into a self-explaining state, and it is the piece a first-time user most needs — they will land here precisely because they clicked something that added `svgReplaces` for them.

### 2. Give the "values found in the SVG" list a real empty / error state

`renderSwatches()` bails out silently when there are no suggestions (`properties.js:6832-6834`), so three genuinely different situations render as *the exact same blank space*:

1. the image is a bitmap — nothing can ever be listed;
2. the image is an SVG but every value it uses is already mapped;
3. the image could not be read at all — `fetchSvgReplaceCandidates()` falls back to `guess`, and because a URL ending in `.svg` guesses `isSvg: true`, the section is shown for a broken/offline `.svg` with an empty body and no hint that the file 404'd.

Case 3 is easy to hit — I set `image: "/i/does-not-exist.svg"` and got a bare "SVG replacements" header with nothing behind it and no error anywhere:

missing svg

One short line per case, in the `.propertyPickerEmpty` style already used for "No SVG replacements yet.", would cover all three: *"All values in this SVG are already mapped."* / *"This image could not be loaded, so its values could not be listed."* Since `fetchSVG()` now distinguishes "loaded, not an SVG" (returns `null`) from "could not load" (rejects), the editor has everything it needs to tell these apart — it just throws the distinction away in the `catch`.

3. **`.svgReplaceRow` control heights don't line up.** `client/css/editor/propertyInputs.css:577` leaves the two text inputs at their intrinsic height while the trash button next to them and the `Add replacement` button below are 34px. Measured identically at 1180×820, 1280×800 and 1920×1080: inputs **21px**, buttons **34px**. The row reads as three different control sizes, and 21px is a poor touch target on the tablet-landscape layout where this panel first becomes reachable. A `min-height` matching the sibling buttons would fix it. (Pre-existing CSS, not introduced here — but this PR makes the panel appear in strictly more situations, so it's worth folding in.)

4. **The two inputs in a row are unlabelled once filled.** They only distinguish themselves via placeholders ("value in the SVG" / "widget property, e.g. color"), which disappear the moment there's content — so an existing row reads as `#primaryColor → color` with nothing saying which side is the file and which is the widget. A single small header row above `.svgReplacesList` ("in the SVG" / "widget property"), 12px like `.svgColorsLabel`, would survive being filled in.

5. **The `→` glyph is decorative but exposed.** `properties.js:6736` sets `arrow.textContent = '→'` on a bare `<span>` with no title and no `aria-hidden`, so it is announced as "right arrow" between two unlabelled fields. `aria-hidden="true"` at minimum.

6. **The JSON editor still sniffs SVGs its own way.** The PR's stated goal is that one place decides whether a file is an SVG, and it does unify the engine and the properties sidebar — but `jsonedit.js:90` `checkIfSVG()` is a third copy, and the loosest of the three: `/svg/i.test(text)` matches the literal string "svg" *anywhere* in a decoded file, and it has no content-type shortcut so it stringifies whole bitmaps. It gates the "Show colors in SVG image" command (`jsonedit.js:233`) — the very button the PR description blames for creating these bad `svgReplaces` maps. It behaved correctly on my test PNG (the command stayed hidden), but a bitmap that happens to contain those three bytes would get the button, a list of garbage colour chips, and a one-click dead `svgReplaces` entry. Routing it through the new `fetchSVG()` would stop the two editors from ever disagreeing about the same file.

7. **Demo board in `pr-test-ai`** (not part of the diff, but it's what a human will open): two stray widgets sit outside the demo's three-card layout on the right edge — a "Basic widget" hex piece at `1479,177` and an unlabelled blue arrow button at `1523,351`, both with leftover `svgReplaces` like `{"stuff":"stuff"}`. They look like an unfinished part of the demo. Also the board title (`fontSize: 42`) renders no larger than the card headings (`fontSize: 26`) and visually smaller than the body paragraph below it, so the board has no clear top-level heading:

demo crop

Otherwise the demo reads very well — three labelled cards, the code snippet under each, and four colour buttons that make the SVG-vs-PNG difference obvious in one click. ([full board](https://agent.virtualtabletop.io/reports/ui-review/1538071804789399582/pr-room-1280x800.png))

8. **"SVG replacements" is jargon at the point where a newcomer meets it.** The help line (`properties.js:6715`) — *"Map a value in your SVG file - a color, a stroke width, an opacity - to a widget property that supplies it."* — assumes the reader knows what a widget property is and, crucially, doesn't say that the property's value is then edited *right below in this same panel*. Something closer to *"Recolour this image: pick a value out of the SVG file and choose the widget property that will supply it. That property's value can then be set below."* explains both the mechanism and where the payoff appears, without changing the section title.

9. **Nothing explains what happens for a non-SVG.** The `info` tooltip (`editorPropertyHints.svgReplaces`, `properties.js:1282`) says a value is substituted "inside an uploaded SVG image" — after this PR the behaviour for a non-SVG is a deliberate, documented one (show the image as it is). Worth one clause: *"…An image that is not an SVG is simply displayed as it is."* That is the sentence a confused user needs, and it's now a true statement about the engine.

10. **The suggestion swatches are the good part — lean on them.** "Values found in the SVG - click to add a replacement:" with clickable, colour-filled chips and the `⚠` marker for ambiguous values is genuinely the clearest thing in this panel for someone who has never done this. The `Add replacement` button, by contrast, is the one that invents `#000` out of nothing. Once the suggestions are present, consider making them the primary path and demoting `Add replacement` to a secondary/text button — a first-time user should be clicking a real value out of their file, not typing a guess.

---

Everything else I opened looked unchanged and healthy at all six viewports, and the fix itself is a clear visual improvement over `main` — a widget that used to be invisible now shows up.

VTT-AI-Agent:
Thanks — the big one was real and is fixed, together with most of the rest. Everything below is in `c25e0b05`.

## What changed

**1. The panel now says when the image isn't an SVG.** This was the right call: the fix removed the only feedback there was (the widget vanishing) and put nothing in its place. `fetchSvgReplaceCandidates()` now returns a `status` alongside `isSvg` — `svg`, `notSvg`, `unreadable` or `none` — because "no values to show" had four causes that all rendered as the same blank space. On `notSvg` the section body renders a notice line right under the help text, in the `#cc7722` the CSS and pile editors already use for "this does nothing":

> This image is not an SVG, so these replacements have no effect. They apply again as soon as this widget points at an SVG file.

and `Add replacement` is disabled in that state, with a `title` saying why, instead of inventing `#000 → color2` mappings that can never fire.

**2. Empty / error state for the values list.** `renderSwatches()` no longer returns silently. A broken `.svg` URL (your `/i/does-not-exist.svg` case) now reads *"This image could not be loaded, so the values it uses could not be listed."*; an SVG whose values are all mapped says *"All values in this SVG are already mapped."*; an SVG with nothing replaceable in it says *"No replaceable values found in this SVG."* The `notSvg` case is covered by the notice above rather than saying it twice. That is exactly the distinction you pointed out `fetchSVG()` had made available and the editor was throwing away in its `catch`.

**3. `.svgReplaceRow` control heights.** The inputs were 21px next to 34px buttons because they are `content-box` with a 2px border, so a bare `min-height` would have made them 40px. With `box-sizing: border-box` they measure 34px, the same as the delete button and `Add replacement`, at every viewport.

**4. Header labels above the rows.** `in the SVG` / `widget property`, 12px like `.svgColorsLabel`, rendered only when there is at least one row and aligned to the two inputs (the arrow and delete-button slots are held open by a hidden `→` and a 34px spacer, so the labels sit exactly over the fields they name).

**5. The `→` is `aria-hidden="true"`** in both the rows and the header.

**6. `jsonedit.js checkIfSVG()` is gone as a separate sniff** — it calls `fetchSVG()` now, so the engine, the properties sidebar and the JSON editor are down to one answer for one file. `/svg/i` anywhere in a decoded file no longer counts, and a bitmap is recognized from its `Content-Type` instead of being stringified. Verified in the browser: `Show colors in SVG image` is enabled for the SVG widget and disabled for the PNG.

**8. The help line** is now *"Recolor this image: pick a value out of the SVG file - a color, a stroke width, an opacity - and choose the widget property that supplies it. That property's value is then set below."*, which says both what it does and where the payoff shows up.

**9. The `svgReplaces` tooltip** now ends with *"An image that is not an SVG has nothing to replace and is simply displayed as it is."* — true of the engine as of this PR.

**7. The demo board.** Not the same two problems, but a real one: every label and button carried a `fontSize` property, and neither the label widget nor the validator knows such a property — the size has to be in `css`. So the whole board rendered at the default 16px, which is why the title looked no bigger than the card headings. All 16 widgets now carry `font-size: Npx` in their `css` and the board has a proper hierarchy (42 / 26 / 22 / 19 / 15px); `_meta.info` was brought in line with the tutorials (`mode: Tutorial`, `players`, `language`, no bogus `type`) and `node validator/validate_gamefile_node.js` is now clean on it, where it previously reported 20 problems. The updated board is re-published to `pr-test-ai` under the same name.

I could not reproduce the two stray widgets: the published `.vtt` contains exactly 23 widgets, all part of the three-card layout, and nothing at `1479,177` or `1523,351` — nor any `{"stuff":"stuff"}`. Those look like they were live in the room rather than in the game file; opening the game fresh from the Game Shelf should not show them.

## Not changed

**10. Demoting `Add replacement` below the suggestion swatches.** I agree with the reading — the swatches are the good path and the button is the guessy one — but reordering the section is a design change beyond what this bugfix should carry, and disabling the button in the one state where it was actively harmful covers the concrete damage you found. Worth its own PR if a maintainer wants it.

## Verification

- `npm test` — 32 suites, 1272 tests pass, including a new one asserting that the four `status` values are told apart (`svg` / `notSvg` / `unreadable` / `none`) and that an unreadable file still falls back to its file name.
- TestCafe `editor.js` (which covers this panel) — 45/45 pass locally on Chromium, and it runs on this commit in CI too.
- In a real browser at 1280×800 and 1180×820, one image widget per state side by side:

```
png       notice shown, Add replacement disabled, row heights input:34 span:19 input:34 button:34
svg       no notice, Add replacement enabled, swatch "#fff" offered
missing   "This image could not be loaded, so the values it uses could not be listed."
external  same, and no client-error overlay
all-mapped  "All values in this SVG are already mapped."
```

no page errors in any of them, and the notice fits the sidebar without overflow at both widths.

| PNG (replacements are dead) | SVG (replacements work) | broken `.svg` URL |
| --- | --- | --- |
| png | svg | unreadable |

The demo board with its typography fixed: demo board

VTT-AI-Agent (commented):
## TL;DR

The current head addresses the earlier code/UI findings and CI is green, but the new `Content-Type` shortcut introduces a compatibility regression: an SVG served with an incorrect bitmap MIME type now becomes a blank image instead of being decoded and recoloured as it was on `main`. Preserve body sniffing for that case and add a mismatched-MIME regression test before merge.

## Findings

### 1. A mislabeled SVG now disappears (`client/js/main.js:695-699`)

`getSVG()` on `main` always reads the response body, so an SVG served by an external host as (for example) `Content-Type: image/png` is still converted to a valid `data:image/svg+xml` URL and its replacements work. The new shared `fetchSVG()` instead trusts that header, cancels the body, and returns `null`; `getSVG()` then records the URL in `nonSVGCache` and hands the original URL to CSS.

I reproduced that exact response in Chromium: the old body-to-data-URL path has valid SVG bytes to render, while loading the same SVG bytes directly under `image/png` fires `img.onerror`. Thus an existing game using a misconfigured image host can go from a visible/recoloured widget to a blank one. This is also at odds with the helper's comment that the bytes, rather than the name, decide the file type—the header is neither.

Please keep the MIME fast path only for VTT-controlled routes whose headers are reliable (`/assets/` and `/i/`), or otherwise retain body sniffing for the engine path. A test beside `a bitmap is recognized by its content type` with an SVG body and `image/png` should assert that replacements are still applied.

## Scope checks

- No new widget property or routine operation is introduced, so no additional property-editor, visual-routine UI, validator, or `jeCommands` entry is required; the PR improves the real property editor for the existing `svgReplaces` property.
- Nothing under `library/` is changed, and the PR documents its demo as a `.vtt` in the PR's `pr-test-ai` room rather than committing it to the library.
- None of the beta-excluded engine paths (`client/js/widgets/`, `client/js/compute.js`, or `client/js/serverstate.js`) are touched; `client/js/main.js` changes image rendering only, with no state-sync, default-property, routine, or expression semantics impact.
- All current GitHub checks pass. I could not rerun Jest locally because this clone has no `node_modules/.bin/jest`, but the new tests and surrounding call sites were inspected directly.

The `beta` label was already present; I left it in place because this rendering-only regression cannot corrupt game state or crash the server and the PR remains safe and useful for player testing.

VTT-AI-Agent:
Valid finding, fixed in `c936054c`.

**A mislabeled SVG is decoded again.** The content-type shortcut now only applies to `/assets/` and `/i/` — the two routes vtt serves itself, where the header is generated by `server/fileloader.mjs`/express and is the thing that made the shortcut worth having (an uploaded bitmap asset is the common case, and pulling a multi-megabyte PNG through the wire to decode it as text is exactly what it avoids). For any other URL the bytes decide, as they did on `main`, so an SVG that a foreign host hands out as `image/png` is still turned into a `data:image/svg+xml` URL with its replacements applied:

```js
const contentType = /^(assets|i)\//.test(mappedURL) ? response.headers.get('content-type') || '' : '';
```

That also settles the inconsistency you pointed out with the helper's own comment: the header is now only consulted where vtt itself wrote it, and the file's contents remain what actually decides everywhere else.

**Regression test.** `an SVG that a foreign host mislabels as a bitmap is still recognized by its bytes` sits next to `a bitmap is recognized by its content type` in `tests/client/get-svg.test.js`: an SVG body under `Content-Type: image/png` at `https://example.com/board.svg`, asserting both that the replacement is applied and that the body *was* read (`'read as text'`), while the existing `/assets/4_4` test still asserts the body is `'cancelled'` unread for a vtt-served bitmap.

**Verified in a browser** as well, with a Playwright route that answers `https://mislabel.test/board.svg` with SVG bytes under `image/png` and CORS open, next to a vtt-served PNG and a vtt-served SVG:

```
mislabeled  url("data:image/svg+xml,<svg …><rect … fill="…"/></svg>")   ← decoded, replacement applied
ownPng      url("http://localhost:8333/i/branding/android-512.png")     ← shortcut still taken
ownSvg      url("data:image/svg+xml,<svg …>")
page errors: none
```

The editor side is unchanged by this (`npm test` — 32 suites, 1272 tests pass; the SVG-replacements panel still shows the notice + disabled `Add replacement` for a PNG, the swatches for an SVG, and the "could not be loaded" line for a broken URL).

One consequence worth naming: an *external* bitmap with `svgReplaces` is read as text again instead of being cancelled on its header. In practice a cross-origin image is blocked from `fetch()` by CORS long before that and lands in the reject path, and it is what `main` did — keeping an existing game rendering is worth more than saving a fetch in that corner.

Scope checks match my reading too: no new widget property or routine operation, nothing under `library/` touched, and the demo board lives as a `.vtt` in the PR's `pr-test-ai` room rather than in the library.

VTT-AI-Agent (commented):
## TL;DR

The three earlier reviews' findings are all genuinely addressed at `c936054c` — the CORS/`.catch()` hole is closed, the SVG sniff is unified in one `fetchSVG()`, the content-type shortcut is now limited to vtt's own routes, and the editor panel finally says out loud what the image is. I ran the branch in a browser and the whole matrix behaves as advertised. One real edge case is left: `getSVG()` treats "could not load" as "not an SVG" **permanently**, so a single transient fetch failure silently strips `svgReplaces` off a perfectly good SVG for the rest of the session, and neither a state rebuild nor the editor (which uses a different cache) notices.

## Findings

### 1. A transient fetch failure permanently disables `svgReplaces` for that URL (`client/js/main.js:738-746`)

`fetchSVG()` deliberately *rejects* when the file can't be read, with the comment "that says nothing about what it is, so every caller decides for itself what to assume then". `getSVG()` then decides — and its decision is the same permanent one it makes for a definitively-not-an-SVG file:

```js
fetchSVG(url).catch(_=>null).then(t=>{
  ...
  } else {
    nonSVGCache[url] = true;   // "loaded, no <svg> in it"  AND  "didn't load at all"
```

`nonSVGCache` is never invalidated and the early return at line 710 short-circuits before any refetch, so a same-origin SVG that misses **one** fetch (server restart while a room is open, a wifi blip, a 503) renders un-replaced for the rest of the session. I reproduced it on this branch by aborting only the *first* request to `/i/game-pieces/2D/Hex-Point.svg`:

```
1) right after the transient failure: url("http://localhost:8348/i/game-pieces/2D/Hex-Point.svg")
2) after a full widget rebuild:       url("http://localhost:8348/i/game-pieces/2D/Hex-Point.svg")
```

The second line is after pushing a new state over `PUT /state/…`, i.e. after the client rebuilt every widget from scratch — it still never retries. The visible result is not a broken image but something more confusing: the raw SVG *does* display, with `fill: #primaryColor` / `stroke-width: #borderWidth` as literal invalid CSS values, so the piece renders in default black and the game just looks wrong. And because `fetchSvgReplaceCandidates()` keeps its own cache and refetches successfully, the editor at the same moment shows the swatches and the mapped rows as if everything worked.

`!response.ok` is in the same bucket: a 404'd `/assets/…` is not evidence about the file type either, and gets the same permanent verdict.

This is still better than `main` (which left the widget blank *and* fired the client-error overlay), so it's not a blocker — but the fix stops one step short. Concretely: keep `nonSVGCache` for the definitive answer only (`fetchSVG` resolved `null`), and give a *rejection* its own set that produces the same plain-URL fallback for the current render but is dropped again on the next state load, so a widget rebuild retries once. If the permanence is intentional (avoiding a refetch per `css()` recomputation for a CORS-blocked external image, which is a fair concern), the comment at line 735-737 should say that it applies to a temporary failure too, because right now it reads as being about CORS only.

### 2. `fetchSVG()` is the single decider but not a single fetch (`client/js/main.js:688`, `properties.js:265-276`, `jsonedit.js:91-99`)

Unifying the sniff is the right call, but each of the three callers still keeps its own cache in front of it — `svgCache`/`nonSVGCache`, `svgImageCandidatesCache`, `jeIsSVG` — so opening the properties sidebar and the JSON editor on the same SVG pulls it over the wire three times, and the three answers are only guaranteed to agree as long as the file doesn't change between them. Memoizing the *promise* inside `fetchSVG()` (`cache[url] ??= fetch…`) collapses all three into one request and makes "the three cannot disagree about a file" true by construction rather than by convention — which is what the new comments in `jsonedit.js:87-91` and `properties.js:255-261` claim.

Related, smaller: `fetchSvgReplaceCandidates()` caches `svg` and `notSvg` but returns `guess` uncached for `unreadable` (`properties.js:279`), so a widget pointing at a permanently-404 image re-fetches on every `image` change and every panel rebuild.

### 3. Test gaps around the branches that changed most recently

`tests/client/get-svg.test.js` covers a rejecting `fetch()` but not `response.ok === false` — a distinct branch in `fetchSVG()` (line 690-691) that produces the same permanent fallback as finding 1, and the one a broken `/assets/…` actually takes. A `{ ok: false, status: 404 }` case next to `an image that cannot be loaded at all` would be two lines.

`checkIfSVG()` in `jsonedit.js` changed semantics in this PR — from `/svg/i` anywhere in a decoded file to `<svg` plus the content-type shortcut — and has no unit test at all, so the "Show colors in SVG image" command's gating is only covered by whatever `editor.js` happens to exercise. Not blocking, but it's the one of the three call sites with zero automated coverage of the new behaviour.

## What I verified

Ran the branch at head `c936054c` locally (`npm ci`, `minifyJavascript: false`) with a room containing one image widget per state and drove it with Playwright at 1600×950:

| widget | rendered `background-image` |
| --- | --- |
| PNG + `svgReplaces` | `url(".../i/branding/android-512.png")` — the bug this PR fixes |
| SVG + 3 `svgReplaces` | `url("data:image/svg+xml,…")`, recoloured |
| `/i/does-not-exist.svg` + `svgReplaces` | plain URL |
| `https://example.invalid/board.png` + `svgReplaces` | plain URL, **no** client-error overlay |
| PNG without `svgReplaces` (control) | unchanged |

No `pageerror` in any of them. In edit mode the four editor states come out right: the PNG gets the warning line, a disabled `Add replacement` with its explanatory `title`, and no swatches; the SVG gets three rows plus "All values in this SVG are already mapped."; the broken and the CORS-blocked image both get "This image could not be loaded…"; and a PNG with no replacements hides the section entirely. The new header row lines up with the inputs (both columns within 1px) and the inputs now measure 34px like the buttons beside them.

Full jest suite green locally on this branch (32 suites, 1272 tests, including the 7 new ones), and all GitHub checks including CodeQL pass.

## Scope checks

- No new widget property and no new routine operation or parameter function is introduced, so there is nothing that needs a new property-editor control, a visual-routine block, a `jeCommands` entry or a `validator/` entry. The PR in fact *improves* the real property editor for the existing `svgReplaces` property rather than only the JSON editor.
- Nothing under `library/` is touched, and the demo board is published as a `.vtt` on the PR's own `pr-test-ai` test server rather than committed — correct for this repo's rules. (It is not tutorial-quality/tutorial-split, but as a PR demo board rather than a new-feature test room that's the right call here; there is no new authoring surface to teach.)
- `export`ing `getSVG`/`fetchSVG` from `main.js` is safe in the client bundle: `removeExportStatements()` (`server/minify.mjs:286-289`) strips `export` before `function`/`async function`, and `fetchSVG` reaches the editor bundle through the shorthand key in `Object.assign(window, {…})`, which terser preserves.
- Verified `server.mjs:160-183` really does set `Content-Type` from the first byte for `/assets/`, and `/i` is `express.static`, so restricting the content-type shortcut to `^(assets|i)/` is exactly the set of routes whose headers can be trusted — review 3's finding is properly fixed, and the mislabeled-external-SVG case has a test.

The `beta` label was already present and I left it: the diff touches none of the beta-excluded engine paths (`client/js/widgets/`, `client/js/compute.js`, `client/js/serverstate.js`), it is rendering-only with no state, sync or routine-semantics impact, and everything I exercised in a browser worked without errors, so it stays safe and useful for players to try with the findings above still open.

VTT-AI-Agent:
Good catch on all three — the transient-failure one was real and is the reason for `7d74d02b`. All of it is in that commit.

## 1. A transient fetch failure no longer sticks

You were right that the fix stopped one step short: `nonSVGCache` was answering two different questions. It now only holds the definitive one ("loaded, and there is no `<svg>` in it"), and a file that did not arrive at all goes into its own `unreadableCache` with a timestamp:

```js
// Images that could not be read at all, and when that last happened. Unlike the two caches above
// this is not an answer about the file - a server restart, a hiccup in the network or a CORS
// rejection say nothing about what the file contains - so it expires and the next widget asking
// for the image tries again.
const unreadableCache = {};
const UNREADABLE_RETRY_MS = 30000;
```

I went with an expiring entry rather than "dropped on the next state load" because it also covers the case where nothing rebuilds the widgets — a room that is simply open when the server restarts recovers on its own the next time anything recomputes that widget's CSS, without needing a state push. The 30 seconds is the compromise your own remark named: long enough that a CORS-blocked external image is not refetched on every `css()` recomputation, short enough that a blip costs the image its replacements for a moment rather than for the session.

Two details that fell out of it:

- `!response.ok` is in the *unreadable* bucket now, not the permanent one, exactly as you argued — a 404'd `/assets/…` is not evidence about a file type either.
- While a retry is in flight the widget keeps the plain URL it can already display instead of blinking to nothing, and is told through its callback once the file did arrive after all. `getSVG()` returns `mapAssetURLs(url)` rather than `''` for as long as the URL is in `unreadableCache`.

Reproduced your scenario on the new head with a Playwright route that aborts **only the first** request to `/i/game-pieces/2D/Hex-Point.svg` (the count includes the browser's own `` load of the fallback URL):

```
after the aborted first request    requests=2  url("http://localhost:8333/i/game-pieces/2D/Hex-Point.svg")
rebuild inside the retry delay     requests=2  url("http://localhost:8333/i/game-pieces/2D/Hex-Point.svg")
waiting out the 30s retry delay...
rebuild after the retry delay      requests=3  url("data:image/svg+xml,%3Csvg%20version%3D%221.2%22...")
page errors: none
```

The middle line matters as much as the last one: inside the delay it is *not* refetched, so a recomputation storm can't turn into a request storm.

## 2. One request per file

`fetchSVG()` memoizes now, so it is a single fetch as well as a single decider:

```js
export function fetchSVG(url) {
  const key = mapAssetURLs(url);
  if(!svgFetchCache[key])
    // a request that failed is not an answer worth keeping - drop it so the next caller retries
    svgFetchCache[key] = loadSVG(url).catch(e=>{ delete svgFetchCache[key]; throw e; });
  return svgFetchCache[key];
}
```

Dropping a rejected promise is what keeps this compatible with finding 1 — a memoized rejection would have been the same permanent verdict one level down. Keying on the **mapped** URL turned out to matter too: while measuring this I found the engine asking for `/i/game-pieces/2D/Hex-Point.svg` and the game-piece picker (`symbols.js`) asking for `i/game-pieces/2D/Hex-Point.svg`, i.e. the same file twice under two spellings. Measured in a browser, counting only `fetch`/`xhr` requests for that file (the `image: 1` is the sidebar's `` preview, not a `fetchSVG()` call):

```
after the engine rendered it:   1
after the properties sidebar:   1   | svgReplaces section: 2 row(s)
after the JSON editor:          1
requests by resource type:  { fetch: 1, image: 1 }
```

On `c936054c` the same run was `1 → 2 → 2`.

The smaller point — `fetchSvgReplaceCandidates()` returning `guess` uncached for `unreadable` — I left as it is on purpose, and for the same reason as finding 1: caching that answer in the editor would be the permanent-verdict bug again, just in the panel instead of in the engine, and the panel is rebuilt far less often than `css()` is recomputed. With the memoization above, a repeated look at the same broken image is one request shared with everything else that asks, not three.

## 3. Test coverage

`tests/client/get-svg.test.js` gained four tests (32 suites, **1276** pass):

- `an image the server refuses to hand out is used as it is` — the `{ ok: false, status: 404 }` branch you named, through a new `mockFetchStatus()` helper.
- `an image that could not be read is tried again once the retry delay is over` — stubs `Date.now()` and asserts all three states: no refetch inside the delay, a successful retry past it with the replacement actually applied, and that a file which loaded and turned out not to be an SVG is *not* retried even after a minute.
- `one file is asked for once, however it is spelled and whoever asks` — engine, `checkIfSVG` and a direct `fetchSVG()` on `/i/piece.svg` vs `i/piece.svg`, asserting a single `read as text`.
- `the JSON editor recognizes exactly the files the engine can replace in` — `checkIfSVG()` finally has unit coverage. It is evaluated out of `jsonedit.js`'s scope (the same trick `property-inputs.test.js` uses for the editor sources) and handed the real `fetchSVG`, so it covers both halves of the semantics change: an SVG is `true`, a bitmap containing the letters `svg` is `false`, and a rejected fetch is `false`.

## Verification

`npm test` — 32 suites, 1276 tests pass. In a browser on the new head: the four-state matrix (PNG + `svgReplaces`, SVG, missing `.svg`, CORS-blocked external) renders exactly as before with no page errors, the mislabeled-external-SVG case still decodes to a data URL, and the editor panel's four states (notice + disabled `Add replacement` / swatches / "could not be loaded" / "all mapped") are unchanged. TestCafe `editor.js` runs on this commit in CI.

96LawDawg (approved):
Tested.  Approved.  Ready for merge.

PR #3122, opened 2026-08-15
Add a record button to the routine editor that turns room gestures into operations

VTT-AI-Agent:
## What this does

The visual routine editor gets a second way to add an operation. Next to **add operation** there is now a **⏺ record** button: press it, do the thing in the room, and every press-drag-release you make is offered as the handful of operations that would do the same thing when the routine runs. Click one and it is added to the routine exactly where "add operation" would have added it.

Dragging a card out of the deck holder into a player's hand, for example, offers:

```
dragged the card card1 from the holder deckHolder to the holder hand-p1     ×
  + do exactly this                     Move 1 widget from deckHolder to hand-p1
  + deal one to seat-p1                 Move 1 widget from deckHolder to seat-p1
  + deal one to every player            Move 1 widget from deckHolder to seat-p1 and seat-p2
  + put every card back into deckHolder Gather all the cards back into deckHolder
  + card1 ended up on that face         Turn card1 face up
```

The last one is not something the pointer did: the hand flips what enters it face up, and the recorder read that out of the delta the drop produced.

Clicking a die offers rolling it and putting it on a face; clicking a canvas offers wiping it, the color it draws in and how thick the brush is; clicking a holder offers shuffling it and putting every card of its deck back into it.

## Why it is a short list and not one operation

One gesture is deliberately not one operation. The same drag is "move that card there", "deal one to that player", "deal one to everybody" or "put the deck back together" depending on what the routine is for — the room cannot know which was meant, but it knows the handful worth offering, so it offers those and writes nothing until one is picked. A gesture nobody picks anything from costs a card in the list and no operation in the routine; a `×` in its corner takes it off.

A card is answered once: picking a reading adds the operation and closes the card. A second operation out of the same gesture is had by doing the gesture again — which is also why doing the same drag five times does not write five cards. The fifth deal out of the same holder into the same hand raises the count on the card that is already there, so it reads `Move 5 widgets from deckHolder to hand-p1` rather than filling the panel with five readings of `Move 1`.

The readings come from two places:

- **what the gesture did** — which widget moved out of what into what (`MOVE`, and `MOVEXY` when it ends up on the table), where it was put (`SET x` / `SET y`), or that it was only clicked (`CLICK` for the roll of a die, `CANVAS reset` and the two `SET`s for a canvas, `FLIP` for a card, `SHUFFLE`/`RECALL` on a holder, `TURN` on a seat, `TIMER` on a timer);
- **what the room did on its own while it ran** — taken from the deltas, so a holder that turns a card face up as it enters becomes a "Turn card1 face up" of its own, and any other property the drop set becomes a `SET` for it. Positions, parents and the transient drag properties are left out: the move already says those.

Dealing is recognised: a holder that is a seat's `hand` is offered as the seat (`MOVE to: seat-p1`, because MOVE to a seat also sets the owner) and as every seat at once. `RECALL` is only offered for a holder that actually has a deck in it. A card taken off a stack inside a holder reads as coming out of that holder, not out of the pile the stack formed — no operation takes widgets out of a pile, and the pile is gone again as soon as it holds one card.

## While it is recording

The room answers the way it does for a player rather than the way it does in edit mode:

- the selection band and the drag toolbar step aside, so a left-drag reaches the widget instead of drawing a rubber band;
- a click neither selects the widget — that would take the editor off the routine being recorded — nor runs its click routine; it is only written down;
- a drag *does* move the widget for real, so the next gesture starts from the state the last one left behind, which is what makes a sequence of them read as one macro (undo works as usual).

The board carries a red frame and the pointer is the crosshair the widget picker already uses, so it is visible that a drag in there is being written down. Recording runs in one routine at a time; pressing record in another routine moves it there.

## Notes

- No widget property, routine function or parameter is added or changed, so there is nothing for the JSON editor's insert buttons to learn — this is purely the visual routine editor, which is where the standing "usable from the editing UI" rule points anyway.
- One validator fix rides along (see the review follow-ups below): `validate_gamefile.js` now reads `TURN`'s `turn` the way the engine does — the id of a seat with `turnCycle: 'seat'`, a whole number/`first`/`last` with every other cycle.
- New file `client/js/editor/controls/routinerecorder.js` (added to the editor bundle in `server/minify.mjs`); `mousehandling.js` gains two `edit`-gated hook calls, `selection.js` three.
- New jest tests in `tests/client/routineeditor.test.js` cover the suggestions per gesture kind, the repeat count, that nothing is written until a suggestion is picked and that picking closes the card, and that every suggested operation renders as a finished sentence. `npm test` passes (1297 tests); the TestCafe `editor.js`, `routines.js`, `interactions.js` and `domsnapshot.js` suites were run locally against a Chromium build and pass.
- Creator-facing behaviour is new, so the wiki (Edit-Mode) could use a paragraph about the record button.
- **Draft tutorial on the test server, not in the diff.** The feature is user-visible and no tutorial covers it, so I authored one to tutorial quality and published it as a draft rather than committing it: **`Xtras - Recording Routines`** (single variant, `variant: ""`), in this PR's `pr-test-ai` room — <https://test.virtualtabletop.io/PR-3122/pr-test-ai>. It is a deck holder with the conventional "Recall & Shuffle" button, a discard holder that flips what enters it face up (so the recorder's "the room did this too" suggestions show up), two seats with hands (so dealing suggestions show up), and a **Deal a card** button whose click routine is empty on purpose — the one to record into. Metadata template: `Properties - dragLimit` (the newest tutorial), mirrored field for field; the library image is a 530×530 square SVG in the style of its `Xtras - …` neighbours, reusing the shared graduate glyph and the "Xtras" category lettering. `node validator/validate_gamefile_node.js` exits 0 and I walked it end to end (sit down, record two gestures, pick a suggestion, leave edit mode, click the button and watch the card get dealt). `library/tutorials/` itself is untouched — move it in only if you want it.

## Review follow-up (automated review 1/4)

- **The recording no longer outlives the editor.** It is stopped when the properties module closes (next to the `stopWidgetPicker()` the review pointed at) *and* in `toggleEditMode()`, because the two ways out of edit mode are not the same: the editor's close button and Ctrl-J run `closeEditor()` → `onEditorClose()`, but clicking the **Active Game** tab leaves edit mode without ever calling it. Both were reproduced in a browser (crosshair + red frame left over play mode) and both are clean now.
- **The 8-suggestion cap no longer eats the delta-derived readings.** Gesture readings and "the room did this too" readings are collected in separate lists; half the card is reserved for the latter and anything they leave unused goes back to the gesture. The example above is the real output of the reviewer's own two-seat dealing gesture.
- **`TURN turn: <seat id>` validates.** Widened in `operationProps.TURN` — as a side effect this removes four false-positive errors in existing public-library games (Code Numbers ×2, Shell Game, Temple Trek) and turns `Property is invalid false` into a message that says what is allowed. (Review 3/4 then tied this to `turnCycle`; see below for the final rule.)
- **`MOVE to:` without a source is only offered when a `DEFAULT` collection is in scope** (`RoutineEditor.collectionsInScope()` walks the operations before the insertion point), so the recorder cannot write an operation that is an error the moment it is added.
- **`SET parent: null` keeps its place.** Dragging a widget out onto the table now also offers the `x` and the `y` it was dropped at, since those are counted from the parent and would otherwise land the widget somewhere else.
- **Foreign deltas stay out of a gesture.** `routineRecorderReceiveDelta()` drops deltas whose `delta.c` names another player.
- **Tests:** five new jest tests cover the cap, the collection gate, the position pair, the foreign-delta filter and the recording ending with its editor. `npm test` is green (1283) and the TestCafe `editor.js` suite passes locally.

## Feedback follow-up (Discord, 2/2)

> Clicking dice should show options to roll the die and set to a specific face.
>
> Style mismatch between add Operation and record.

(<https://discord.com/channels/770758631146782780/1538058046285160518/1538097648350007317>)

- **Clicking a die now offers the roll and the face.** A dice widget used to fall through to the generic readings — `Click die1` and a blank `SET`. Rolling a die is the one thing a routine does by clicking rather than with an operation of its own (`Dice.click()` rolls when nothing else claims the click), so the roll is now the first reading of the gesture: a `CLICK` told to leave out whatever would stop it rolling — `mode: ignoreClickRoutine` for a die whose click was given a routine, `ignoreClickable` for one players may not click, `ignoreAll` for both. Under it comes `SET activeFace`, filled in with the face the die is showing, for putting it on a face on purpose; the aside says which face that is, since `activeFace` counts from 0 and a die usually has something else printed on it. This is what the two library idioms for dice already look like (`CLICK … mode: ignoreClickRoutine` in *Bustin' at the Blue Diamond*, `SET activeFace` in *Types - Dice*).
- **A die the room rolled during a gesture is read as `SET activeFace` too.** The delta-derived readings turned every `activeFace` change into a `FLIP`, which does nothing on a die — `Dice` has no `flip()` for `FLIP` to call. Cards and basic widgets are unchanged.
- **`record` is the same button as `add operation` now.** It was a filled blue button standing next to the dashed lowercase outline it belongs with; it now shares that shape (the `.events-editor-add` rule in `events.css`), with the red dot as the only thing that sets it apart. Armed, only the fill changes — the box keeps its size, so pressing it no longer resizes the row — and the text on the fill is the page colour, which reads on the dark red of the light theme and the light red of the dark theme alike.

Before/after and the new suggestions:

| | |
|---|---|
| the two buttons | [light](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/record-button-light.png) · [dark](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/record-button-dark.png) |
| armed | [light](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/record-button-armed-light.png) · [dark](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/record-button-armed-dark.png) |
| clicking two dice | [suggestions](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/dice-suggestions.png) |
| in the draft tutorial | [screenshot](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/tutorial-dice-recording.png) |

Three new jest tests cover the roll, the modes that keep it rolling and the die-rolled-by-the-room reading; the dice gesture also joined the "every suggestion reads as a finished sentence" test. `npm test` is green (1286) and the TestCafe `editor.js` suite passes locally. Walked in a browser end to end as well: record a click on a die into an empty click routine, pick the roll, leave edit mode, press the button — the die rolls; pick `Set activeFace` instead and the button pulls the die back onto that face after every roll.

The **draft tutorial** on the test server was amended to match and re-published: it now has a die with a paragraph of its own next to it, so the new readings can be tried where the rest of the recorder is tried. Still a draft, still not in `library/tutorials/`.

## UI/UX review follow-up

The record button and its suggestion panel were reviewed for layout and readability. Everything that was actionable is fixed:

- **The drag toolbar really steps aside now.** `body.editorRoutineRecording #editorDragToolbar { display: none }` was outranked by `body:not(.overlayActive) #editorDragToolbar.active` in `dragtoolbar.css`, so a 299×36 edit toolbar sat inside the red "being recorded" frame. The rule now matches `.active` as well. Verified in a browser: `display: none`, `0×0`, in both themes.
- **Every line of the panel clears 4.5:1 in both themes**, measured in the running app against the actually painted background: header 5.62 / 7.18, hint 12.63 / 11.89, gesture headline 21.0 / 18.69, the reading 6.69 / 9.53, the sentence 8.52 / 9.18. Three changes get there: the panel is drawn on the page background instead of `--backgroundHighlightColor2`, the gesture headline is text-coloured instead of red (the left border already carries the colour), and a row now hovers with an outline instead of a fill — every fill light enough to read black on is dark enough to lose the blue on it in the other theme. `--routineRecordColor` itself is unchanged, so the frame and the armed button keep their identity.
- **Each suggestion leads with what the reading is FOR** and follows with the operation it would add, in the aside style the intent used to have. Six of eight rows used to start with the same eleven characters; the line that tells them apart is now the line you read first.
- **The headline names what kind of thing each widget is** — `dragged the card card_S_4K from the holder deckHolder to the holder hand-p1`, and `... to the pile o0ur` for the auto-created pile the reviewer hit. Ids in a real game are generated as often as chosen.
- **A widget the gesture created contributes no readings.** Dropping a card onto a card makes a pile, and that pile reported `it got width while you did that` / `... height` as two of its eight suggestions. Only widgets that existed when the gesture started can produce "the room did this too" readings now.
- **A gesture nobody meant can be taken off the card** with a quiet `×` in its corner. What it already added stays in the routine, like anything else added to a routine.
- **One name for one state.** The button keeps the label `record` and only its fill changes (so it no longer grows 70→96px and can no longer wrap the row), its tooltip says *Stop recording*, and the panel header says *Recording*. `done` is a quiet outlined button in the recorder's red instead of falling through to the sidebar's default filled blue, which made the way out the loudest thing in the panel. Only the header's dot blinks now.
- **The room says which routine it is being recorded into** — a small caption in the top-left of the frame, e.g. `recording → dealButton · clickRoutine`, because the panel that knows is in a sidebar that is usually scrolled elsewhere by the third gesture.
- **The empty-state hint says how to stop and what a click does**: "clicking one records the click instead of selecting it … Press done when the routine says what you meant."
- **The unfilled `Set property of X to number or text` reading is gone.** A card of three readings where one is a blank form undermines the other two.

Not changed: the record button is still offered at every width (a phone shows a 354×221 room strip where the gesture is barely performable) — worth a separate look, but hiding a feature by viewport is a bigger decision than this PR should make on its own.

Screenshots: [panel, light](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/panel-light.png) · [panel, dark](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/panel-dark.png) · [recording inside the tutorial](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/recording-in-the-tutorial.png)

Two new jest tests cover forgetting a gesture (and that what it added stays) and the created-widget filter; the button test now asserts the label does not change. `npm test` is green (1288).

### Draft tutorial, re-published

`Xtras - Recording Routines` was reworked along the review's notes and re-published to this PR's `pr-test-ai` room ([board screenshot](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/tutorial-board.png), [library tile next to its neighbour](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/tutorial-tile.png)):

- the prerequisites are a numbered **To try it** strip directly under the intro (1. sit down · 2. `ctrl-j` · 3. click **Deal a card** · 4. unfold *click* · 5. press **record** · 6. drag a card), instead of a clause buried in a paragraph below the objects;
- the finish line ("press **done** … then click **Deal a card** and watch the room do it again") is the last block on the board;
- the intro is two sentences and the full-width paragraph at the bottom is folded into a note beside the seats, where the "a seat only takes cards while somebody is in it" rule applies;
- the **discard holder** carries a `Discard` label, in the place the deck holder's "Recall & Shuffle" button occupies;
- the die is 120×120 showing five pips instead of 70×70 showing one, so it reads as a die;
- the four objects and their captions share left edges (120 / 480 / 840 / 1180) on one pitch, and the dead band between the die row and the bottom is gone;
- the library tile leads with a big red record dot on a small board frame, instead of a large frame with a play-shaped glyph.

Still a draft, still not in `library/tutorials/`. Validator exits 0 and I walked it end to end again in the running app: open from the shelf (with its preview image), sit, `ctrl-j`, **Deal a card** → *click* → **record**, drag a card into the hand, pick *deal one to every player*, **done**, leave edit mode, press the button — both seats get a card.

## Review follow-up (automated review 3/4)

- **`TURN`'s `turn` is validated against the turn cycle it is used with.** The widened predicate above accepted any existing widget id in every mode, so `{ "turnCycle": "seat", "turn": "button1" }` and `{ "turnCycle": "seat", "turn": "first" }` both validated clean although seat mode only ever looks the value up among the seats and falls back to the first one after reporting a problem. It now asks for a widget of type `seat` in seat mode and for a whole number/`first`/`last` everywhere else, with messages that name the other mode; a `turnCycle` that is only known at runtime (`${...}`) leaves `turn` alone. The `customRoutineChecks` special case for the reverse direction is gone — one predicate covers both now.
- Six regression tests in the new `tests/validator/turn.test.js`. `npm test` is green (1294).
- Running the validator over all 575 public-library game and tutorial variants before and after gives an identical problem set except for one reworded message (`Irrigation/0.json` points `turnCycle: 'seat'` at `0_seat` while its seats are `1_seat`…`5_seat` — already reported before this commit).

Not changed, and pre-existing on `main`: the visual editor declares `turn` once for all five variants (`special: [ 'first', 'last' ], widgetType: 'seat'`), so its popup offers `first`/`last` and a seat picker whichever cycle is selected. Making that variant-aware means teaching `parameterSpec()` per-variant overrides — a change to the routine editor's parameter machinery rather than to the recorder, and it looks like its own PR.

## Feedback follow-up (maintainer, open-PR overview)

> less options, less technical. options should be what an inexperienced user would want. clicking dice: roll dice or set active face. clicking canvas: reset, set color, set brush size. move card: deal cards to hands, recall if moved to deck's parent. stuff like that; nothing too fancy.
>
> also: if the user moves 5 cards between the same holders, update the suggestions to MOVE 5 instead of listing 5 new suggestion blocks
>
> when a user picks one of the options, close the box. if the user wants multiple options, they have to do the action again.
>
> hover state of the "record" button has bad font color - barely readable. the same applies to the "add X" buttons with the same style

All four, and the top of this description was rewritten to match.

**Fewer readings, and the ones an inexperienced author would want.** The cap is 5 instead of 8, and most cards now come in well under it. A drag between holders reads as the move it was, as dealing to the seat the hand belongs to and to every player, and as putting every card back into the holder the deck lies in — `MOVE count: all`, `MOVE ... face`, "take one from every player", the shuffle-afterwards and the `DEFAULT`-collection reading are gone, and with the last of them `RoutineEditor.collectionsInScope()`, which nothing else used. A drag inside a holder no longer offers sorting and shuffling it, only the position it was dropped at. Clicking a die offers exactly the roll and the face, with no "click it" underneath saying the same thing in words that never mention rolling; clicking a canvas is new and offers wiping it (`CANVAS mode: reset`), the color it draws in (`SET activeColor`, with the color that place in its map holds named in the reading) and how thick the brush is (`SET lineWidth`) — the three things the canvas's own buttons do; clicking a holder offers shuffling and recalling, not sorting and moving something out of it; and `CALL clickRoutine` is gone everywhere.

**Doing the same thing again raises the count.** A drag from one holder into another carries a key of the two holders, and a later drag with the same key adds to the card that is already there instead of writing another one: the headline becomes `dragged 5 widgets from the holder deckHolder to the holder hand-p1` and every reading of it counts 5. What the room did during the later drags is folded in as well. Only holder-to-holder moves coalesce — that is the operation with a count to raise, and the gesture that is really done over and over.

Making that work in a real room needed one more fix: the second card dealt out of a holder comes off a stack, and its parent is the pile that stack formed, which is neither something `MOVE` takes widgets out of nor something that is still there afterwards. The source of a drag is now resolved up to the nearest holder or seat, the way the destination already was — so the second deal is the same gesture as the first, and reads `Move 2 widgets from deckHolder to hand-p1` rather than `Move card2 to hand-p1`.

**Picking closes the card.** The tick that marked a used reading and the `added` bookkeeping behind it are gone; the card is answered once and disappears, and the panel's hint says so. The recording itself carries on, so the next gesture opens the next card.

**The hover was a fill nobody asked for.** `.routine-editor-record-operation:hover` and the `add …` buttons it shares its shape with set only `border-color` and `color`, so the `button:hover { background-color: var(--VTTblueDark) }` in `layout.css` filled them with a solid dark blue under text written in `--VTTblue` — 1.95:1 in the light theme, 2.21:1 in the dark one, measured in the running app. They keep no fill on hover now, and on the dark theme they are written in the teal the popup buttons already hover in, because `--VTTblue` on the near-black page is only 3:1. Measured after: **7.22:1** light, **12.32:1** dark. The armed record button and `done`, which are meant to be filled, are unchanged at 5.6/7.2 and 5.0/7.0. The same fix reaches `add routine`, the property panel's `add` and its two buttons — every button with that shape.

| | light | dark |
|---|---|---|
| hover, before | [screenshot](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/hover-before-light.png) | [screenshot](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/hover-before-dark.png) |
| hover, after | [screenshot](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/hover-after-light.png) | [screenshot](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/hover-after-dark.png) |
| the new cards | [screenshot](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/light-suggestions.png) | [screenshot](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/dark-suggestions.png) |

`npm test` is green (1297) and the TestCafe `editor.js` suite passes locally (45/45). Walked in a browser in both themes: three deals out of a deck holder into a hand collapse into one card counting 2 (one of the three drags never left the holder and became its own card, correctly), the die and the canvas read as above, and picking *deal 2 to every player* writes `{"func":"MOVE","from":"deckHolder","to":["seat-p1","seat-p2"],"count":2}` and closes the card.

The draft tutorial on the test server still walks correctly — its steps are "press record, drag a card, pick a reading, press done" — so it was not re-published.

## Feedback follow-up (open-PR overview)

> set x+y must be offered in _one_ suggestion but added as two operations

**A reading is one row, however many operations it takes to carry out.** A suggestion now carries a list of operations, and picking it adds all of them in order, where "add operation" would have added them. The drag that puts a widget at a spot is one row again — `put it exactly there`, spelling out `Set x of free1 to 715` and `Set y of free1 to 870` under it — and taking a widget out onto the table is one row of `SET parent: null`, `SET x` and `SET y` instead of three. The `and put it exactly there` / `and at that height` labels are gone with them: they only existed because the clauses of one reading had to be separate rows, and half a reading picked is a widget somewhere nobody dropped it. [Screenshot](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/one-reading-two-operations.png)

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3122/pr-test (or any other room on that server)

### DIFF
--- a/client/css/editor/controls/events.css
+++ b/client/css/editor/controls/events.css
@@ -199,12 +199,6 @@ body.edit.darkMode .events-editor-property-key {
   font-size: 12px;
 }
 
-.events-editor-property-add-button:hover:not(:disabled) {
-  border-color: var(--VTTblue);
-  color: var(--VTTblue);
-  opacity: 1;
-}
-
 /* nothing to add without a name: saying so beats a button that does nothing */
 .events-editor-property-add-button:disabled {
   opacity: 0.4;
@@ -235,12 +229,6 @@ body.edit.darkMode .events-editor-property-key {
   font-size: 14px;
 }
 
-.events-editor-property-buttons button:hover {
-  border-color: var(--VTTblue);
-  color: var(--VTTblue);
-  opacity: 1;
-}
-
 /* one shape for every "add something" in the panel: lowercase, dashed outline.
    "record" is one of them: it is the other way to get an operation into a
    routine, so it is the same button as "add operation" next to it and says
@@ -259,19 +247,42 @@ body.edit.darkMode .events-editor-property-key {
   font-size: 12px;
 }
 
+/* The hover of all of them, the property panel's two included: the outline and
+   the label turn blue and the fill stays off. Saying the fill stays off is what
+   keeps it off - a plain button hovers into a solid var(--VTTblueDark) (see
+   layout.css), which is all but the blue this writes the label in. */
 .events-editor-add:hover,
 .routine-editor-add-operation:hover,
-.routine-editor-record-operation:hover {
+.routine-editor-record-operation:not(.recording):hover,
+.events-editor-property-add-button:hover:not(:disabled),
+.events-editor-property-buttons button:hover {
+  background: none;
   border-color: var(--VTTblue);
   color: var(--VTTblue);
   opacity: 1;
 }
 
+/* that blue is a dark one, drawn on a page that is nearly black on the dark
+   theme - so there it is the teal the popup buttons hover in instead */
+body.edit.darkMode .events-editor-add:hover,
+body.edit.darkMode .routine-editor-add-operation:hover,
+body.edit.darkMode .routine-editor-record-operation:not(.recording):hover,
+body.edit.darkMode .events-editor-property-add-button:hover:not(:disabled),
+body.edit.darkMode .events-editor-property-buttons button:hover {
+  border-color: var(--textHighlightColor7);
+  color: var(--textHighlightColor7);
+}
+
 /* the only add button that sits on an operation card instead of on the page:
    the dashed outline of the others is the color of that card, so it takes the
-   page background the same way a nested routine is a hole carved into the card */
-.routine-editor-add-else {
+   page background the same way a nested routine is a hole carved into the card
+   - on hover as well, where the rule above turns every other one's fill off */
+.routine-editor-add-else,
+.routine-editor-add-else:hover {
   background: var(--backgroundColor);
+}
+
+.routine-editor-add-else {
   border-color: var(--backgroundColor);
   margin-left: 12px;
 }
--- a/client/css/editor/controls/events.css
+++ b/client/css/editor/controls/events.css
@@ -241,9 +241,13 @@ body.edit.darkMode .events-editor-property-key {
   opacity: 1;
 }
 
-/* one shape for every "add something" in the panel: lowercase, dashed outline */
+/* one shape for every "add something" in the panel: lowercase, dashed outline.
+   "record" is one of them: it is the other way to get an operation into a
+   routine, so it is the same button as "add operation" next to it and says
+   what it does with its dot rather than with a shape of its own */
 .events-editor-add,
-.routine-editor-add-operation {
+.routine-editor-add-operation,
+.routine-editor-record-operation {
   align-self: flex-start;
   border: 1px dashed var(--backgroundHighlightColor1);
   border-radius: 4px;
@@ -256,7 +260,8 @@ body.edit.darkMode .events-editor-property-key {
 }
 
 .events-editor-add:hover,
-.routine-editor-add-operation:hover {
+.routine-editor-add-operation:hover,
+.routine-editor-record-operation:hover {
   border-color: var(--VTTblue);
   color: var(--VTTblue);
   opacity: 1;
--- a/client/css/editor/controls/routine.css
+++ b/client/css/editor/controls/routine.css
@@ -495,3 +495,180 @@ body.edit.darkMode .routine-editor-parameter-missing {
 .routine-editor-fields .routine-editor-func-name {
   color: var(--routineVariableColor);
 }
+
+/* The two ways to get an operation into a routine, side by side at the end of
+   it: naming the one wanted ("add operation") and doing the thing in the room
+   and picking from what it could mean ("record"). */
+.routine-editor-buttons {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+  gap: 4px;
+}
+
+.routine-editor-record-operation .material-symbols {
+  font-size: 15px;
+  vertical-align: -3px;
+  margin-right: 3px;
+  /* the recorder's red lightened until it carries on the blue of a button - the
+     dark one this red is elsewhere disappears into it */
+  color: #ff8a80;
+}
+
+/* armed, the whole button is the red of what it is doing rather than one more
+   blue button among the others: the room is being watched, and that has to be
+   visible from wherever in the sidebar the eye happens to be */
+.routine-editor-record-operation.recording {
+  background-color: var(--routineRecordColor);
+  font-weight: bold;
+}
+
+.routine-editor-record-operation.recording:hover {
+  background-color: #8e1e1e;
+}
+
+.routine-editor-record-operation.recording .material-symbols {
+  color: white;
+  animation: routineRecordBlink 1.4s ease-in-out infinite;
+}
+
+@keyframes routineRecordBlink {
+  50% { opacity: 0.15; }
+}
+
+@media (prefers-reduced-motion: reduce) {
+  .routine-editor-record-operation.recording .material-symbols {
+    animation: none;
+  }
+}
+
+/* What the room has been doing since the button was pressed, framed in the same
+   red so it reads as one thing with the button below it rather than as another
+   part of the routine. It sits between the operations and the buttons because
+   that is where the operations it offers would land. */
+.routine-editor-recording {
+  border: 1px solid var(--routineRecordColor);
+  border-radius: 6px;
+  padding: 4px 6px;
+  background: var(--backgroundHighlightColor2);
+  font-size: 12px;
+}
+
+.routine-editor-recording-header {
+  display: flex;
+  align-items: center;
+  gap: 4px;
+  color: var(--routineRecordColor);
+  font-weight: bold;
+}
+
+.routine-editor-recording-header .material-symbols {
+  font-size: 16px;
+  animation: routineRecordBlink 1.4s ease-in-out infinite;
+}
+
+@media (prefers-reduced-motion: reduce) {
+  .routine-editor-recording-header .material-symbols {
+    animation: none;
+  }
+}
+
+.routine-editor-recording-title {
+  flex: 1;
+}
+
+.routine-editor-recording-hint {
+  color: var(--textColor);
+  opacity: 0.8;
+  margin: 2px 0 4px 0;
+}
+
+/* one card per press-drag-release: what was done, and under it the readings of
+   it the routine could mean */
+.routine-editor-gesture {
+  border-left: 3px solid var(--routineRecordColor);
+  border-radius: 4px 4px 4px 0;
+  background: var(--backgroundHighlightColor1);
+  padding: 3px 6px;
+  margin-top: 4px;
+}
+
+.routine-editor-gesture-what {
+  font-size: 11px;
+  font-weight: bold;
+  letter-spacing: 0.5px;
+  color: var(--routineRecordColor);
+}
+
+.routine-editor-suggestion {
+  display: flex;
+  align-items: baseline;
+  flex-wrap: wrap;
+  gap: 2px 4px;
+  padding: 2px 3px;
+  border-radius: 4px;
+  cursor: pointer;
+}
+
+.routine-editor-suggestion:hover {
+  background: var(--routineActiveCardColor);
+}
+
+.routine-editor-suggestion .material-symbols {
+  font-size: 15px;
+  align-self: center;
+  color: var(--routineOperationColor);
+}
+
+.routine-editor-suggestion-sentence {
+  flex: 1;
+  min-width: 0;
+  color: var(--routineOperationColor);
+}
+
+/* why the room thinks this is one of the readings, in the tone of an aside:
+   the sentence is what would be added, this is only what it is for */
+.routine-editor-suggestion-why {
+  width: 100%;
+  padding-left: 19px;
+  font-size: 11px;
+  font-style: italic;
+  color: var(--textColor);
+  opacity: 0.7;
+}
+
+/* a suggestion that was already added: it stays clickable (doing a thing twice
+   is a real thing to want), it just says it was used */
+.routine-editor-suggestion-added .routine-editor-suggestion-sentence {
+  opacity: 0.6;
+}
+
+.routine-editor-suggestion-added .material-symbols {
+  color: var(--routineCollectionColor);
+}
+
+/* While the room is being watched, the pointer says so - the same crosshair the
+   widget picker puts over the room - and the board carries a frame in the
+   recorder's red, so a drag in there is visibly being written down rather than
+   just moving a card around. */
+body.editorRoutineRecording, body.editorRoutineRecording .widget {
+  cursor: crosshair !important;
+}
+
+/* the board rather than #room: #room is only the untransformed box the widgets
+   are scaled out of, so a frame on it would sit somewhere inside the board */
+body.editorRoutineRecording #roomArea::after {
+  content: '';
+  position: absolute;
+  inset: 0;
+  border: 3px solid var(--routineRecordColor);
+  pointer-events: none;
+  z-index: 100;
+}
+
+/* the drag toolbar of the selected widget is edit-mode's way of moving things,
+   and it needs the selection band's event handling that the recording takes
+   over - so it steps aside while the room is the player's */
+body.editorRoutineRecording #editorDragToolbar {
+  display: none;
+}
--- a/client/css/editor/controls/routine.css
+++ b/client/css/editor/controls/routine.css
@@ -533,6 +533,7 @@ body.edit.darkMode .routine-editor-parameter-missing {
 
 .routine-editor-record-operation.recording:hover {
   filter: brightness(0.85);
+  background: var(--routineRecordColor);
   border-color: var(--routineRecordColor);
   color: var(--backgroundColor);
 }
@@ -690,13 +691,6 @@ body.edit.darkMode .routine-editor-parameter-missing {
   font-style: italic;
   color: var(--textColor);
   opacity: 0.7;
-}
-
-/* a suggestion that was already added stays clickable (doing a thing twice is a
-   real thing to want), so it does not go grey the way a disabled control does -
-   the tick is what says it was used */
-.routine-editor-suggestion-added .material-symbols {
-  color: var(--routineCollectionColor);
 }
 
 /* While the room is being watched, the pointer says so - the same crosshair the
--- a/client/css/editor/controls/routine.css
+++ b/client/css/editor/controls/routine.css
@@ -537,30 +537,28 @@ body.edit.darkMode .routine-editor-parameter-missing {
   color: var(--backgroundColor);
 }
 
+/* the dot does not blink here as well: the panel above the button has one that
+   does, a few pixels away, and two of them blinking at each other is noise */
 .routine-editor-record-operation.recording .material-symbols {
   color: var(--backgroundColor);
-  animation: routineRecordBlink 1.4s ease-in-out infinite;
 }
 
 @keyframes routineRecordBlink {
   50% { opacity: 0.15; }
 }
 
-@media (prefers-reduced-motion: reduce) {
-  .routine-editor-record-operation.recording .material-symbols {
-    animation: none;
-  }
-}
-
 /* What the room has been doing since the button was pressed, framed in the same
    red so it reads as one thing with the button below it rather than as another
    part of the routine. It sits between the operations and the buttons because
-   that is where the operations it offers would land. */
+   that is where the operations it offers would land.
+   It is drawn on the page background rather than on a highlight so that the red
+   of the header and the blue of the readings keep the 4.5:1 the rest of the
+   editor is tuned for - the frame around it is what sets it apart. */
 .routine-editor-recording {
   border: 1px solid var(--routineRecordColor);
   border-radius: 6px;
   padding: 4px 6px;
-  background: var(--backgroundHighlightColor2);
+  background: var(--backgroundColor);
   font-size: 12px;
 }
 
@@ -587,6 +585,24 @@ body.edit.darkMode .routine-editor-parameter-missing {
   flex: 1;
 }
 
+/* stopping is not the thing to do here, so it does not look like it: a quiet
+   outline in the recorder's red rather than the sidebar's default filled blue,
+   which made the way out the loudest thing in the panel */
+.routine-editor-recording-done {
+  background: none;
+  border: 1px solid var(--routineRecordColor);
+  border-radius: 4px;
+  color: var(--routineRecordColor);
+  font-size: 11px;
+  padding: 0 5px;
+  cursor: pointer;
+}
+
+.routine-editor-recording-done:hover {
+  background: var(--routineRecordColor);
+  color: var(--backgroundColor);
+}
+
 .routine-editor-recording-hint {
   color: var(--textColor);
   opacity: 0.8;
@@ -594,11 +610,12 @@ body.edit.darkMode .routine-editor-parameter-missing {
 }
 
 /* one card per press-drag-release: what was done, and under it the readings of
-   it the routine could mean */
+   it the routine could mean. The red edge is what makes it a card - a fill of
+   its own would take every line on it a step closer to its background. */
 .routine-editor-gesture {
+  position: relative;
   border-left: 3px solid var(--routineRecordColor);
-  border-radius: 4px 4px 4px 0;
-  background: var(--backgroundHighlightColor1);
+  border-top: 1px solid var(--backgroundHighlightColor1);
   padding: 3px 6px;
   margin-top: 4px;
 }
@@ -607,7 +624,29 @@ body.edit.darkMode .routine-editor-parameter-missing {
   font-size: 11px;
   font-weight: bold;
   letter-spacing: 0.5px;
+  padding-right: 16px; /* the × sits in this corner */
+  color: var(--textColor);
+}
+
+/* taking a gesture back off the card: quiet until the card is looked at, in the
+   red of the thing it belongs to */
+.routine-editor-gesture-forget {
+  position: absolute;
+  top: 0;
+  right: 2px;
+  background: none;
+  border: none;
+  padding: 0 2px;
+  font-size: 14px;
+  line-height: 1;
   color: var(--routineRecordColor);
+  opacity: 0.85;
+  cursor: pointer;
+}
+
+.routine-editor-gesture:hover .routine-editor-gesture-forget,
+.routine-editor-gesture-forget:focus-visible {
+  opacity: 1;
 }
 
 .routine-editor-suggestion {
@@ -620,8 +659,12 @@ body.edit.darkMode .routine-editor-parameter-missing {
   cursor: pointer;
 }
 
+/* an outline rather than a fill: a row is two lines of coloured text, and every
+   fill light enough to read black on is dark enough to lose the blue on it in
+   one of the two themes */
 .routine-editor-suggestion:hover {
-  background: var(--routineActiveCardColor);
+  outline: 1px solid var(--routineOperationColor);
+  outline-offset: -1px;
 }
 
 .routine-editor-suggestion .material-symbols {
@@ -630,15 +673,17 @@ body.edit.darkMode .routine-editor-parameter-missing {
   color: var(--routineOperationColor);
 }
 
-.routine-editor-suggestion-sentence {
+/* what this reading is FOR leads the row: the readings of one gesture nearly
+   all begin with the same words, so this is the line the eye picks one by */
+.routine-editor-suggestion-why {
   flex: 1;
   min-width: 0;
   color: var(--routineOperationColor);
 }
 
-/* why the room thinks this is one of the readings, in the tone of an aside:
-   the sentence is what would be added, this is only what it is for */
-.routine-editor-suggestion-why {
+/* and the operation it would add follows it, in the tone of an aside: it is
+   what the reading is checked against, not what it is chosen by */
+.routine-editor-suggestion-sentence {
   width: 100%;
   padding-left: 19px;
   font-size: 11px;
@@ -647,12 +692,9 @@ body.edit.darkMode .routine-editor-parameter-missing {
   opacity: 0.7;
 }
 
-/* a suggestion that was already added: it stays clickable (doing a thing twice
-   is a real thing to want), it just says it was used */
-.routine-editor-suggestion-added .routine-editor-suggestion-sentence {
-  opacity: 0.6;
-}
-
+/* a suggestion that was already added stays clickable (doing a thing twice is a
+   real thing to want), so it does not go grey the way a disabled control does -
+   the tick is what says it was used */
 .routine-editor-suggestion-added .material-symbols {
   color: var(--routineCollectionColor);
 }
@@ -676,9 +718,32 @@ body.editorRoutineRecording #roomArea::after {
   z-index: 100;
 }
 
+/* which routine the room is being recorded into, in the room: the panel that
+   says so is in a sidebar that is scrolled somewhere else by the time a few
+   gestures have been made */
+body.editorRoutineRecording #routineRecordingLabel {
+  position: absolute;
+  top: 3px;
+  left: 3px;
+  z-index: 101; /* above the frame, which is drawn at 100 */
+  max-width: calc(100% - 12px);
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  padding: 1px 6px;
+  border-radius: 0 0 4px 0;
+  background: var(--routineRecordColor);
+  color: var(--backgroundColor);
+  font-size: 12px;
+  font-weight: bold;
+  pointer-events: none;
+}
+
 /* the drag toolbar of the selected widget is edit-mode's way of moving things,
    and it needs the selection band's event handling that the recording takes
-   over - so it steps aside while the room is the player's */
-body.editorRoutineRecording #editorDragToolbar {
+   over - so it steps aside while the room is the player's. The .active is what
+   dragtoolbar.css shows it with, so hiding it takes at least as much. */
+body.editorRoutineRecording #editorDragToolbar,
+body.editorRoutineRecording #editorDragToolbar.active {
   display: none;
 }
--- a/client/css/editor/controls/routine.css
+++ b/client/css/editor/controls/routine.css
@@ -506,29 +506,39 @@ body.edit.darkMode .routine-editor-parameter-missing {
   gap: 4px;
 }
 
+/* the button is the shape of "add operation" next to it (see events.css), so
+   the dot is the only thing that makes it the record button - it carries the
+   recorder's red the panel and the room frame are drawn in */
 .routine-editor-record-operation .material-symbols {
   font-size: 15px;
   vertical-align: -3px;
   margin-right: 3px;
-  /* the recorder's red lightened until it carries on the blue of a button - the
-     dark one this red is elsewhere disappears into it */
-  color: #ff8a80;
+  color: var(--routineRecordColor);
 }
 
-/* armed, the whole button is the red of what it is doing rather than one more
-   blue button among the others: the room is being watched, and that has to be
-   visible from wherever in the sidebar the eye happens to be */
+/* armed, the whole button is filled with the red of what it is doing rather
+   than staying one more outline among the others: the room is being watched,
+   and that has to be visible from wherever in the sidebar the eye happens to
+   be. Only the fill changes - the box stays the size it was, so pressing it
+   does not shift the row it is in. */
 .routine-editor-record-operation.recording {
-  background-color: var(--routineRecordColor);
+  background: var(--routineRecordColor);
+  border-color: var(--routineRecordColor);
+  /* the record red is a dark one on the light theme and a light one on the
+     dark theme, so what reads on it is the page it is drawn on either way */
+  color: var(--backgroundColor);
+  opacity: 1;
   font-weight: bold;
 }
 
 .routine-editor-record-operation.recording:hover {
-  background-color: #8e1e1e;
+  filter: brightness(0.85);
+  border-color: var(--routineRecordColor);
+  color: var(--backgroundColor);
 }
 
 .routine-editor-record-operation.recording .material-symbols {
-  color: white;
+  color: var(--backgroundColor);
   animation: routineRecordBlink 1.4s ease-in-out infinite;
 }
 
--- a/client/css/editor/layout.css
+++ b/client/css/editor/layout.css
@@ -39,6 +39,10 @@ body {
      the 4.5:1 the rest of this file budgets for. Dark mode is fine as it is. */
   --routineBlankColor: #d32f2f;
 
+  /* Watching the room to record a routine: the red every recorder anywhere is,
+     darkened to the same 4.5:1 the blank above is tuned for. */
+  --routineRecordColor: #c62828;
+
   /* The card of the routine editor that was worked on last: a clear step lighter
      than the other cards (--backgroundHighlightColor1) so it is spotted without
      reading, while the chrome on it (see --routineChromeOpacity) still clears
@@ -89,6 +93,7 @@ body.edit.darkMode {
   --routinePropertyColor: #ff8fb1;
   --routineNumberColor: #40e0d0;
   --routineBlankColor: red;
+  --routineRecordColor: #ff6b6b;
   --routineActiveCardColor: #626262;
 
   color-scheme: dark;
--- a/client/js/editor/controls/routine.js
+++ b/client/js/editor/controls/routine.js
@@ -2205,9 +2205,15 @@ class RoutineEditor {
   }
 
   addOperation(values) {
+    this.addOperations([ values ]);
+  }
+
+  // several operations added as one: they go in together, in the order given,
+  // and the last of them is where the next one follows
+  addOperations(list) {
     const at = this.insertionIndex();
-    this.routine.splice(at, 0, typeof values == 'string' ? values : JSON.parse(JSON.stringify(values)));
-    this.setActiveOperation(at); // the new operation is where the next one follows
+    this.routine.splice(at, 0, ...list.map(values=>typeof values == 'string' ? values : JSON.parse(JSON.stringify(values))));
+    this.setActiveOperation(at + list.length - 1);
     this.routineChanged();
   }
 
@@ -2268,28 +2274,34 @@ class RoutineEditor {
     forget.title = 'Forget this gesture';
 
     for(const suggestion of gesture.suggestions) {
-      const editor = editorForOperation(suggestion.operation);
-      editor.setOperationDetails(this.widget, suggestion.operation, this.variables, this.collections);
-
       const suggestionDOM = div(gestureDOM, 'routine-editor-suggestion');
       const icon = document.createElement('span');
       icon.className = 'material-symbols';
       icon.textContent = 'add';
       suggestionDOM.append(icon);
-      // what the reading is FOR leads, the operation it would add follows: the
+      // what the reading is FOR leads, the operations it would add follow: the
       // readings of one gesture nearly all start with the same words ("Move 1
       // widget from deckHolder to ..."), so a card of them is only scannable
       // by the thing that tells them apart
       div(suggestionDOM, 'routine-editor-suggestion-why').textContent = suggestion.why;
-      div(suggestionDOM, 'routine-editor-suggestion-sentence').textContent = editor.getSentenceText();
-      suggestionDOM.title = `Add this operation to the routine (${suggestion.operation.func})`;
+      // a reading that takes more than one operation to carry out is still one
+      // line to pick: every operation it would add is spelled out under it
+      for(const operation of suggestion.operations) {
+        const editor = editorForOperation(operation);
+        editor.setOperationDetails(this.widget, operation, this.variables, this.collections);
+        div(suggestionDOM, 'routine-editor-suggestion-sentence').textContent = editor.getSentenceText();
+      }
+      const funcs = suggestion.operations.map(operation=>operation.func).join(', ');
+      suggestionDOM.title = suggestion.operations.length > 1
+        ? `Add these operations to the routine (${funcs})`
+        : `Add this operation to the routine (${funcs})`;
       focusable(suggestionDOM, _=>{
-        // one gesture is one operation: the card is answered and closes, and a
+        // one gesture is one reading: the card is answered and closes, and a
         // second reading of the same gesture is had by doing it again in the
         // room. A card that stays open after a pick invites reading the whole
         // list a second time to see whether anything else applies too.
         closeRoutineGesture(gesture);
-        this.addOperation(suggestion.operation);
+        this.addOperations(suggestion.operations);
       });
     }
   }
--- a/client/js/editor/controls/routine.js
+++ b/client/js/editor/controls/routine.js
@@ -2175,23 +2175,110 @@ class RoutineEditor {
       this.domElement.append(emptyHint);
     }
 
-    const addButton = button(this.domElement, 'add operation', async _=>{
+    this.renderRoutineRecording(this.domElement);
+
+    // the two ways to get an operation into the routine, side by side: naming
+    // the one wanted, and doing the thing in the room and picking from what that
+    // could mean
+    const buttonsDOM = div(this.domElement, 'routine-editor-buttons');
+
+    const addButton = button(buttonsDOM, 'add operation', async _=>{
       const popup = new RoutineOperationPopup();
       popup.setSource(addButton);
       popup.setOperationDetails({}, [ 'func' ], this.widget, this.variables, this.collections);
       const values = await newRoutineValues(popup);
-      if(values !== undefined) {
-        // right after the card that was worked on last, so a routine is built in
-        // the order it runs; at the end when that card is the last one anyway
-        const at = activeRoutineOperation && activeRoutineOperation.routine === this.routine ? Math.min(activeRoutineOperation.index+1, this.routine.length) : this.routine.length;
-        this.routine.splice(at, 0, typeof values == 'string' ? values : JSON.parse(JSON.stringify(values)));
-        this.setActiveOperation(at); // the new operation is where the next one follows
-        this.routineChanged();
-      }
+      if(values !== undefined)
+        this.addOperation(values);
     });
     addButton.className = 'routine-editor-add-operation';
 
+    this.renderRecordButton(buttonsDOM);
+
     return this.domElement;
+  }
+
+  // where an added operation goes: right after the card that was worked on last,
+  // so a routine is built in the order it runs; at the end when that card is the
+  // last one anyway
+  addOperation(values) {
+    const at = activeRoutineOperation && activeRoutineOperation.routine === this.routine ? Math.min(activeRoutineOperation.index+1, this.routine.length) : this.routine.length;
+    this.routine.splice(at, 0, typeof values == 'string' ? values : JSON.parse(JSON.stringify(values)));
+    this.setActiveOperation(at); // the new operation is where the next one follows
+    this.routineChanged();
+  }
+
+  // The record button, next to "add operation": armed, it says what it is doing
+  // rather than what it does, in the same words as unarmed - the way the widget
+  // picker's button does. Pressing it in another routine moves the recording
+  // there rather than refusing: one room can only be watched for one routine.
+  renderRecordButton(buttonsDOM) {
+    const recordingHere = isRoutineRecordingIn(this);
+    const recordButton = button(buttonsDOM, recordingHere ? 'recording…' : 'record', _=>{
+      if(recordingHere)
+        stopRoutineRecording();
+      else
+        startRoutineRecording(this);
+    });
+    recordButton.className = 'routine-editor-record-operation';
+    recordButton.classList.toggle('recording', recordingHere);
+    recordButton.title = recordingHere
+      ? 'Stop watching the room'
+      : 'Do something in the room and get the operations that would do the same thing';
+    recordButton.insertAdjacentHTML('afterbegin', '<span class="material-symbols">fiber_manual_record</span>');
+  }
+
+  // What the room has been doing since the record button was pressed: one card
+  // per press-drag-release, each offering the operations that would do the same
+  // thing. Nothing is added until one of them is clicked - a gesture nobody
+  // meant costs a card in the list and no operation in the routine.
+  renderRoutineRecording(dom) {
+    const recording = routineRecordingIn(this);
+    if(!recording)
+      return;
+
+    const panel = div(dom, 'routine-editor-recording');
+    const header = div(panel, 'routine-editor-recording-header');
+    header.insertAdjacentHTML('beforeend', '<span class="material-symbols">fiber_manual_record</span>');
+    div(header, 'routine-editor-recording-title').textContent = 'Watching the room';
+    const doneButton = button(header, 'done', _=>stopRoutineRecording());
+    doneButton.className = 'routine-editor-recording-done';
+    doneButton.title = 'Stop watching the room';
+
+    div(panel, 'routine-editor-recording-hint').textContent = recording.gestures.length
+      ? 'Click a suggestion to add it to this routine. Carry on in the room for the next step.'
+      : 'Drag and click widgets in the room the way the routine should. Each move becomes a list of operations to pick from - the room does not know which reading you meant, so it offers the likely ones.';
+
+    for(const gesture of recording.gestures)
+      this.renderRecordedGesture(panel, recording, gesture);
+  }
+
+  renderRecordedGesture(panel, recording, gesture) {
+    const gestureDOM = div(panel, 'routine-editor-gesture');
+    div(gestureDOM, 'routine-editor-gesture-what').textContent = gesture.label;
+    for(const [ index, suggestion ] of gesture.suggestions.entries()) {
+      const editor = editorForOperation(suggestion.operation);
+      editor.setOperationDetails(this.widget, suggestion.operation, this.variables, this.collections);
+      const key = `${gesture.key}/${index}`;
+      const added = recording.added.indexOf(key) != -1;
+
+      const suggestionDOM = div(gestureDOM, 'routine-editor-suggestion');
+      suggestionDOM.classList.toggle('routine-editor-suggestion-added', added);
+      const icon = document.createElement('span');
+      icon.className = 'material-symbols';
+      icon.textContent = added ? 'check' : 'add';
+      suggestionDOM.append(icon);
+      div(suggestionDOM, 'routine-editor-suggestion-sentence').textContent = editor.getSentenceText();
+      div(suggestionDOM, 'routine-editor-suggestion-why').textContent = suggestion.why;
+      suggestionDOM.title = `Add this operation to the routine (${suggestion.operation.func})`;
+      focusable(suggestionDOM, _=>{
+        // the same suggestion twice is a routine that does the thing twice,
+        // which is a real thing to want - the tick only says it was used once
+        if(recording.added.indexOf(key) == -1)
+          recording.added.push(key);
+        this.addOperation(suggestion.operation);
+        scrollRoutineRecordingIntoView(this);
+      });
+    }
   }
 
   // the operation cards that belong to THIS routine level (nested editors keep
@@ -2653,6 +2740,26 @@ class RoutineOperationEditor {
       })
       .replace(templatePlaceholder, (_, p)=>this.getDisplayedValue(p))
       .trim();
+  }
+
+  // the sentence the card would read, in plain words: the same parts
+  // renderSentenceView lays out as chips, with the options that are switched on
+  // included. What an operation nobody has added yet says (the recorder's
+  // suggestions) has to say what it does with the values it comes with, and
+  // those live in its options as often as in its variant.
+  getSentenceText() {
+    let text = '';
+    for(const part of this.sentenceParts()) {
+      if(part.clause) {
+        if(this.clauseIsActive(part.clause))
+          text += this.resolveTemplate(part.template);
+        else if(part.clause.whenOff)
+          text += this.resolveTemplate(part.clause.whenOff);
+        continue;
+      }
+      text += this.resolveTemplate(part.template);
+    }
+    return text.replace(templatePlaceholder, (_, p)=>this.getDisplayedValue(p)).replace(/\s+([,.])/g, '$1').replace(/\s+/g, ' ').trim();
   }
 
   // the whole sentence as one template string, optional parts in [brackets] -
--- a/client/js/editor/controls/routine.js
+++ b/client/js/editor/controls/routine.js
@@ -2200,8 +2200,24 @@ class RoutineEditor {
   // where an added operation goes: right after the card that was worked on last,
   // so a routine is built in the order it runs; at the end when that card is the
   // last one anyway
+  insertionIndex() {
+    return activeRoutineOperation && activeRoutineOperation.routine === this.routine ? Math.min(activeRoutineOperation.index+1, this.routine.length) : this.routine.length;
+  }
+
+  // The collections an operation added there could read: the ones the routine
+  // starts with plus the ones the operations before that point define. The
+  // recorder asks before offering an operation that works on "whatever an
+  // earlier operation picked" - one that reads a collection nothing filled is an
+  // error the moment it lands in the routine.
+  collectionsInScope() {
+    const collections = [ ...this.collections ];
+    for(const operation of this.operations.slice(0, this.insertionIndex()))
+      collections.push(...operation.getDefinedCollections());
+    return [ ...new Set(collections.filter(c=>typeof c == 'string')) ];
+  }
+
   addOperation(values) {
-    const at = activeRoutineOperation && activeRoutineOperation.routine === this.routine ? Math.min(activeRoutineOperation.index+1, this.routine.length) : this.routine.length;
+    const at = this.insertionIndex();
     this.routine.splice(at, 0, typeof values == 'string' ? values : JSON.parse(JSON.stringify(values)));
     this.setActiveOperation(at); // the new operation is where the next one follows
     this.routineChanged();
--- a/client/js/editor/controls/routine.js
+++ b/client/js/editor/controls/routine.js
@@ -2223,13 +2223,14 @@ class RoutineEditor {
     this.routineChanged();
   }
 
-  // The record button, next to "add operation": armed, it says what it is doing
-  // rather than what it does, in the same words as unarmed - the way the widget
-  // picker's button does. Pressing it in another routine moves the recording
-  // there rather than refusing: one room can only be watched for one routine.
+  // The record button, next to "add operation". Armed, only its fill changes:
+  // the label stays the word it was, so the button keeps the width it had and
+  // does not shove "add operation" onto a line of its own in a narrow sidebar.
+  // Pressing it in another routine moves the recording there rather than
+  // refusing: one room can only be recorded into one routine.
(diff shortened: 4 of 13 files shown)


VTT-AI-Agent (commented):
## TL;DR

Nice, well-scoped feature — no engine files, no new properties/functions, CI green, and the demo room correctly stayed out of `library/`. Two things I'd fix before merge: **the recording is never stopped when the properties module or edit mode closes**, so `body.editorRoutineRecording` (crosshair over everything + a red frame around the board) leaks into play mode; and **the hard cap of 8 suggestions silently drops exactly the delta-derived ones** ("the room did this too") in the standard dealing gesture the PR advertises.

## Findings

### 1. The recording outlives the editor: crosshair + red frame stay on while playing

`client/js/editor/controls/routinerecorder.js:94` — `stopRoutineRecording()` is only ever reached from the record button, the panel's *done* button, `startRoutineRecording()` and the `isConnected` guard in `routineRecordingState()`. Nothing hooks into the sidebar module or the editor closing.

Failure: arm the recorder, then hit the editor's close button (or Ctrl+J) without pressing *done*. `closeEditor()` (`client/js/editor/layout.js:102`) only runs `onEditorClose()` on the modules; the recorder never hears about it, and:

* `body.editorRoutineRecording` stays set, so `routine.css` keeps `cursor: crosshair !important` on the whole page and paints the 3px `--routineRecordColor` frame on `#roomArea::after` — while the author is back in play mode;
* the `isConnected` guard cannot save it, because leaving edit mode only `display:none`s the editor (see the comment at `client/js/editor/sidebar/properties.js:1879`), so the routine editor's `domElement` is still connected;
* the two `mousehandling.js` hooks that would call `routineRecordingState()` are gated on `edit`, so nothing runs while playing. The only way out is re-entering edit mode and pressing *done*, or reloading.

This is exactly the failure the neighbouring `stopWidgetPicker()` call was added for — `properties.js:1868-1886` closes the picker in `onClose()` *and* forwards `onEditorClose()` to it, with the comment "an armed picker keeps the crosshair over the whole page while playing". The recorder should ride along there (export `stopRoutineRecording`, or call it from `closeEditorPopups()`); the same call also covers switching the sidebar to another module and selecting another widget, which today only clear themselves on the next mouseup in the room.

### 2. The 8-suggestion cap eats the "what the room did on its own" suggestions

`routinerecorder.js:262` caps at 8, and `routineGestureSuggestions()` runs `propertySuggestions()` **last** — so the delta-derived readings, which are the feature's whole differentiator over watching the pointer, are the first thing thrown away.

Confirmed by running `routineGestureSuggestions()` directly on a room with a deck holder, two seats and their hands (i.e. the PR's own example gesture, `deckHolder` → `hand-p1`), with the drop flipping the card face up:

```
count 8
MOVE from deckHolder to hand-p1 count 1
MOVE from deckHolder to hand-p1 count all
MOVE from deckHolder to hand-p1 count 1 face 1
MOVE from deckHolder to seat-p1 count 1
MOVE from deckHolder to [seat-p1, seat-p2] count 1
RECALL holder deckHolder
SHUFFLE holder hand-p1
MOVE to hand-p1                      <- generic "whatever an earlier operation picked"
```

Zero property suggestions survive. With any other property the drop set (I re-ran it with `rotation: 90` plus a custom `myFlag`), exactly one gets in and the rest are dropped. Because a hand that turns what enters it face up is the common case, the recorder will routinely stay silent about precisely the thing it claims to notice — and silently, since nothing tells the user suggestions were dropped.

Cheapest fix: add `propertySuggestions()` before the generic `MOVE to` fallback, or reserve a couple of slots for them (e.g. cap the gesture-derived list at 6 and let the delta-derived ones use the rest), or just raise the cap. Worth a test with seats present — the existing "what the room did on its own" test (`tests/client/routineeditor.test.js:4055`) uses a seat-less room, which is why it stays under the cap.

### 3. "Give the turn to it" writes an operation the validator flags

`routinerecorder.js:362` suggests `{ func: 'TURN', turnCycle: 'seat', turn: 'seat-p1' }`. The engine supports that (`client/js/widgets/widget.js:2603-2609` looks the seat up by id), but `operationProps.TURN.turn` in `validator/validate_gamefile.js` only accepts `Number.isInteger(v) || v === 'first' || v === 'last'`. Adding the suggestion and validating the game gives:

```
b1[clickRoutine.0.turn]: Property is invalid false
```

So a click on a suggestion the editor itself offers produces a red validator error in edit mode. Per CLAUDE.md's validator-upkeep rule this belongs in the PR: widen `turn` to also accept a widget id (ideally only when `turnCycle == 'seat'`, mirroring how the engine reads it).

### 4. Smaller things

* `routinerecorder.js:324` — `{ func: 'MOVE', to }` with neither `from` nor `collection` validates as `no input given and collection DEFAULT is undefined` when it is added to a routine that has no earlier `SELECT` (confirmed with `validate_gamefile_node.js`). The editor already knows which collections are in scope (`this.collections` is passed to `setOperationDetails`) — only offering "move whatever an earlier operation picked" when a collection is actually defined would keep the recorder from writing an operation that is immediately an error.
* `routinerecorder.js:290` — `SET parent = null` does not preserve the position: `x`/`y` are parent-relative (`Widget.coordGlobalFromCoordParent`, `widget.js:614`), so at run time the widget jumps to wherever its old parent-relative coordinates land in room space, unlike the drag that was recorded. Either pair it with the `x`/`y` SETs or say so in the *why* line.
* `routinerecorder.js:137` — `routineRecorderReceiveDelta()` swallows **every** delta that arrives between press and release, including ones from other players and from widgets ticking on their own (a running timer, a dice animation), which then show up as "`timer1` got `secondsElapsed` while you did that". The delta already carries its cause (`delta.c`, set by `setDeltaCause` to `<player> dragged <id>` / `clicked <id>`) — filtering on that would keep foreign changes out cheaply.
* Test coverage for finding 1 is missing entirely (nothing exercises "recording stops when the routine editor goes away"), and the suggestion tests all run in rooms small enough never to reach the cap in finding 2.

### Process checks

* **UI rule:** satisfied — the PR adds no widget property and no routine function/parameter, and the feature itself *is* editor UI (the record button sits next to *add operation*), so there is nothing for `jeCommands` to learn. Finding 3 is the one validator entry the diff still owes.
* **Test room:** satisfied — the diff touches nothing under `library/`, and the draft `Xtras - Recording Routines` room is on the PR's `pr-test-ai` server as described. I could not open the test server from here, so I did not verify it against `docs/tutorials.md`. One thought: the feature has several fairly distinct facets (dealing/collecting suggestions, "the room did this too", click suggestions, repositioning), which the tutorials doc would rather see as two variants than one board.

---

I added the `beta` label: the diff touches no engine file (nothing under `client/js/widgets/`, no `compute.js`, no `serverstate.js` — the `mousehandling.js` and `selection.js` hooks are all `edit`-gated, so play mode is unchanged), all checks including CodeQL are green, and the worst problem I found is a cosmetic leftover cursor/frame that a reload clears — safe and useful for players to try.

VTT-AI-Agent:
Thanks — all four findings were real; every one is fixed in 54855822.

**1. The recording outlived the editor.** Confirmed in a browser and fixed in two places, because there are two ways out of edit mode and only one of them runs `onEditorClose()`:

* `PropertiesModule.onClose()` now calls `stopRoutineRecording()` right next to `stopWidgetPicker()` — that covers the editor's close button, Ctrl-J and switching the sidebar to another module.
* `toggleEditMode()` in `client/js/main.js` calls it as well, because clicking the **Active Game** tab while editing goes straight to `toggleEditMode()` and never through `closeEditor()`. I reproduced exactly your repro through both paths (`body.editorRoutineRecording` still set, `cursor: crosshair` on `body` after leaving edit mode) and both now come back clean. Worth noting that the widget picker has the same gap on the tab path — I left that alone, it is not this PR's.

A jest test covers a recording being stopped while its editor is already detached; the two call sites themselves were verified with Playwright rather than in jest, since `properties.js` and `main.js` are not in that test bundle.

**2. The cap ate the delta-derived suggestions.** Fixed the way you suggested, by reserving slots rather than raising the cap: `routineGestureSuggestions()` collects the gesture readings and the room readings into two lists, gives the room readings up to half the card, and hands back whatever they do not use. Your two-seat dealing gesture, run for real in the browser with a hand that flips what enters it face up, now reads:

```
dragged card1 from deckHolder to hand-p1
  + Move 1 widget from deckHolder to hand-p1                        do exactly this
  + Move all widgets from deckHolder to hand-p1                     move everything that is in there
  + Move 1 widget from deckHolder to hand-p1 and turn them face up  and turn them over on the way
  + Move 1 widget from deckHolder to seat-p1                        deal one to seat-p1
  + Move 1 widget from deckHolder to seat-p1 and seat-p2            deal one to every player
  + Gather all the cards back into deckHolder                       gather every card back where it came from
  + Shuffle hand-p1                                                 shuffle what is in there afterwards
  + Turn card1 face up                                              card1 ended up on that face
```

New test `what the room did survives a gesture that has more readings than fit` uses a three-seat room, which is well past the cap. Screenshot of the real thing: <https://agent.virtualtabletop.io/reports/pr/1538058046285160518/routine-recorder-suggestions.png>

**3. `TURN turn: <seat id>` was flagged by the validator.** Widened in `operationProps.TURN` to accept a string that is an existing widget id, with a message that says what is allowed instead of `Property is invalid false`. Two things came out of checking it against the whole library:

* it removes four **existing** false positives that have nothing to do with this PR — `Code Numbers` (×2), `Shell Game` and `Temple Trek` all use `turnCycle: 'seat'` today and were all reported as errors in edit mode;
* `Irrigation` has a genuine typo (`turn: '0_seat'`, and the game's seats are `1_seat`…`5_seat`), which now reads as `'0_seat' is neither a whole number, 'first', 'last' nor the id of a seat` instead of `Property is invalid false`.

I also took the "ideally only when `turnCycle == 'seat'`" half: a `customRoutineChecks` entry says so when a seat id is used with another cycle. Across `library/games` and `library/tutorials` that adds zero problems, and the total problem count drops from 7595 to 7591 with no new entries.

Nothing to add to `jeCommands` for this — `TURN`/`turn` is already there, and the visual routine editor has always offered it (`textHint: 'id of a seat (used with turnCycle seat)'`, `widgetType: 'seat'`). The validator was the only thing lagging behind the engine.

**4. The smaller ones**

* `{ func: 'MOVE', to }` is now only offered when a `DEFAULT` collection is actually in scope. `RoutineEditor` grew `collectionsInScope()`, which walks the operations *before the insertion point* (not just the routine's starting collections) and asks them for `getDefinedCollections()` — so it appears after a `SELECT` and nowhere else. Two tests, one per side.
* `SET parent: null` now comes with `SET x` and `SET y` for where the widget was actually dropped, worded as "and put it exactly there" / "and at that height". `MOVEXY` was not an option here — it has no `collection` parameter, so it cannot address a widget that was in something other than a holder (a pile, typically).
* `routineRecorderReceiveDelta()` drops a delta whose `delta.c` names another player. It keeps causeless deltas on purpose: the cause is cleared on every flush, so the tail of a drop that flushes twice arrives without one, and dropping those would lose real consequences of the gesture. That means a self-caused timer tick during a gesture can still show up — filtering *that* out reliably needs more than the cause string, and it seemed the wrong trade against losing what a holder did.
* Test coverage for finding 1 and 2 added, plus one for the foreign-delta filter — five new tests, `npm test` green at 1283, TestCafe `editor.js` green locally (45 passed).

**On the test room:** I have not re-authored it this turn — the sandbox is rebuilt between turns and the `Xtras - Recording Routines` draft from the first turn is already on `pr-test-ai`. Your point about splitting it into two variants is fair (dealing/collecting and "the room did this too" on one board, click and repositioning suggestions on another); say the word and I will publish it as two variants. It is a draft either way, `library/tutorials/` is untouched.

VTT-AI-Agent:
Both points are in, in `c2c8fbf`:

> Clicking dice should show options to roll the die and set to a specific face.
>
> Style mismatch between add Operation and record.

(<https://discord.com/channels/770758631146782780/1538058046285160518/1538097648350007317>)

**Clicking a die** used to produce only the generic readings — `Click die1` and a blank `SET`. Rolling a die is the one thing a routine does by clicking rather than with an operation of its own (`Dice.click()` rolls when nothing else claims the click), so the roll is now the first reading of that gesture, and putting the die on a face on purpose the second:

```
clicked die1
  + Click die1                                roll it the way clicking it does
  + Set activeFace of die1 to 3               put it on a face instead of rolling it - it is showing 4
  + Set property of die1 to number or text    change a property of it

clicked die2                                  (this one has a click routine of its own)
  + Click die2, but do not run their click routines   roll it the way clicking it does
  + Set activeFace of die2 to 1               put it on a face instead of rolling it - it is showing b
  + Click die2                                click it the way a player would
  + Run the routine clickRoutine of die2      run its click routine and wait for it
  + Set property of die2 to number or text    change a property of it
```

The roll leaves out whatever would stop it: `mode: ignoreClickRoutine` for a die whose click was given a routine, `ignoreClickable` for one players may not click, `ignoreAll` for both. The face reading is filled in with the face the die is showing, and its aside says which face that is, because `activeFace` counts from 0 while a die usually has something else printed on it. Both match what the library already does (`CLICK … mode: ignoreClickRoutine` in *Bustin' at the Blue Diamond*, `SET activeFace` in *Types - Dice*). One related fix rides along: a die the room rolled *during* a gesture was read as a `FLIP`, which does nothing on a die — `Dice` has no `flip()` — and is now a `SET activeFace` as well.

**The button** now has the shape of the `add operation` it stands next to (the shared `.events-editor-add` rule), with the red dot as the only thing that sets it apart. Armed, only the fill changes, so pressing it no longer resizes the row, and the text on that fill is the page colour so it reads on the dark red of the light theme and the light red of the dark theme alike:

| | light | dark |
|---|---|---|
| idle | idle light | idle dark |
| armed | armed light | armed dark |

Three new jest tests cover the roll, the modes that keep it rolling and the die-rolled-by-the-room reading; `npm test` is green (1286) and the TestCafe `editor.js` suite passes locally. Walked end to end in a browser too: record a click on a die into an empty click routine, pick the roll, leave edit mode, press the button — the die rolls; pick `Set activeFace` instead and the button pulls the die back onto that face after every roll.

The draft tutorial in <https://test.virtualtabletop.io/PR-3122/pr-test-ai> was amended to match and re-published — it now carries a die with a paragraph of its own, so the new readings can be tried where the rest of the recorder is tried ([screenshot](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/tutorial-dice-recording.png)). It is still a draft; `library/tutorials/` is untouched.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review) of the record button and its suggestion panel. The panel is a genuinely nice piece of design — it sits exactly where the operation will land, it reads as one thing with the red frame around the board, and in the tutorial's fixture it produces the eight-reading card the description promises. Two things hold it back: **the edit-mode drag toolbar does not actually step aside while recording** (the new `display:none` rule is outranked by an existing one, so a 300×36 toolbar sits in the middle of the board you are supposed to gesture in), and **the line that identifies each gesture — the one thing you read first — is the least readable thing in the panel** (11px bold red at 3.5:1 in both themes, naming raw widget ids). Fix the toolbar rule first.

*Manually requested review (UI/UX) — by `claude-opus-5`*

**What I exercised.** Local server on this PR's branch plus the PR test server. Viewports 390×844, 820×1180, 1280×800, 1920×1080 — at each: edit mode, the Edit Widgets (`tune`) module, the JSON module, the routine editor with the record button armed and disarmed, the Game Shelf / states overlay, the state-details view, the "Switch game" confirm, the Players overlay, the About overlay, and light + dark mode. I did **not** reach the deck editor, the share/publish overlay, the assets/debug/toolbox modules or the add-widget dialogs. Nothing in those I did open showed clipping or collision attributable to this diff — the new CSS is scoped to `.routine-editor-*` / `body.editorRoutineRecording`, and the two selectors bolted onto the shared `.events-editor-add` rule inherit its look correctly (`add operation` and `record` are indistinguishable in shape, which is the intent).

---

## Big layout improvements

### 1. The drag toolbar does not step aside — it sits on the board you are recording in

`client/css/editor/controls/routine.css:682`

```css
body.editorRoutineRecording #editorDragToolbar { display: none; }   /* 1 id, 1 class */
```

is outranked by the existing

```css
/* client/css/editor/dragtoolbar.css:9 */
body:not(.overlayActive) #editorDragToolbar.active { display: flex; } /* 1 id, 2 classes */
```

so it never applies while the toolbar is actually shown. Confirmed live: with `document.body.classList.contains('editorRoutineRecording') === true`, `#editorDragToolbar` computes to `display: flex` at `299×36`, parked over the room. It covers tutorial text at 1280 and board content at 1920:

drag toolbar over the board while recording

This matters beyond tidiness: the toolbar is an edit-mode control (duplicate, align, flip…) dropped inside the red frame that says "everything you do in here is being written down". A first-time author cannot tell whether pressing one of those buttons becomes an operation.

Smallest fix that keeps both intents: `body.editorRoutineRecording #editorDragToolbar.active { display: none; }` — or add `:not(.editorRoutineRecording)` to the dragtoolbar rule so the two live in one place.

### 2. Lead each suggestion with the intent, not with the sentence

Eight readings of one gesture is the right call, but as laid out they are very hard to scan — six of the eight rows start with the same eleven characters:

the eight-suggestion card

```
+ Move 1 widget from deckHolder to hand-p1          do exactly this
+ Move all widgets from deckHolder to hand-p1       move everything that is in there
+ Move 1 widget from deckHolder to hand-p1 and …    and turn them over on the way
+ Move 1 widget from deckHolder to seat-p1          deal one to seat-p1
+ Move 1 widget from deckHolder to seat-p1 and …    deal one to every player
```

The thing that actually distinguishes them — "deal one to every player" — is the smaller, dimmer, italic line underneath. The emphasis is inverted: the aside is what you choose by, the sentence is what you verify with. Two options that fit the panel as it is:

- swap the two lines: intent first in `--routineOperationColor` at 12px, generated sentence second in the italic aside style; or
- keep the order but group the card into the three families the code already distinguishes — **exactly this** / **or, more generally** / **the room also did this** — with a one-word subhead. That also makes the delta-derived reading ("Turn card_S_4K face up") visibly a different kind of thing rather than the eighth of a flat list.

### 3. Name the widgets the way the author sees them

Every gesture headline and most sentences are raw ids: `dragged card_S_4K from deckHolder to hand-p1`. In a purpose-built room those ids are readable; in a real game they are not. Recording the same drag in the `Functions - MOVE` tutorial produced:

```
dragged pyn6 from o0ur to hand
  + Move pyn6 to hand            do exactly this
  + Shuffle hand                 shuffle what is in there afterwards
```

`pyn6` is a generated card id and `o0ur` is an auto-created pile the author never made and cannot find anywhere in the UI — so the headline names two things that do not exist as far as the author is concerned, and the "do exactly this" suggestion hardcodes a single card id, which is almost never what a routine wants. (This is a fixture difference, not a bug: cards sitting directly in a holder give the good card; a pile in between gives this one. But the pile case is the common one in real games, so it is what most authors will see first.)

Two cheap improvements that fit the existing style:
- prefix the type in the headline (`dragged the card pyn6 out of the pile o0ur into hand`) — the panel already knows `gesture.type`;
- highlight the widget in the room while a gesture card or suggestion is hovered, the way the widget picker does. The board is right next to the panel; that single link would make ids irrelevant.

### 4. A gesture can be added but never removed

The list only grows. A mis-drag, a slip on the wrong holder, a drag you did just to reset the board — all of them stay on the card, and the only way out is `done`, which throws away everything including the good ones. That makes the panel feel expensive to experiment in, which is the opposite of what "just do the thing in the room" wants to feel like. An `×` on each `.routine-editor-gesture` (top-right, quiet, same red) would fix it.

### 5. One state, three names, and two ways to leave it

Armed, the feature calls itself three things at once: the button relabels to **recording…**, the panel header says **Watching the room**, and the button's tooltip says **Stop watching the room**. On top of that there are two stop controls (the toggled button, and `done` in the header).

And `done` is the loudest element in the panel: `routine.js:2260` sets `className = 'routine-editor-recording-done'`, but **no stylesheet matches that class**, so the button falls back to the sidebar's default filled-blue button while everything around it is quiet red and grey outlines. The exit reads as the primary action.

armed state at 820×1180 — filled red button, blue `done`

Suggest: pick one verb ("recording" everywhere, or "watching" everywhere), and either drop `done` and let the button be the single toggle, or keep `done` as a quiet text/outline button and leave the record button's label alone.

---

## Nitpicks / small breakages

- **Contrast, measured in the running app** (composited against the actual painted background, opacity included):

  | element | light | dark | size |
  |---|---|---|---|
  | `.routine-editor-recording-header` | 4.18:1 | **2.64:1** | 12px bold |
  | `.routine-editor-gesture-what` | **3.50:1** | **3.51:1** | 11px bold |
  | `.routine-editor-suggestion-added .routine-editor-suggestion-sentence` | **2.63:1** | **2.68:1** | 12px |

  `client/css/editor/layout.css:42` says the record red is "darkened to the same 4.5:1 the blank above is tuned for" — that budget is against the page background, but both of these sit on `--backgroundHighlightColor2` / `-1`, where it does not hold; and the dark-mode `#ff6b6b` (`layout.css:96`) was not tuned at all. The gesture headline is the worst case and it is the most important line on the card. Darkening `--routineRecordColor` a step in light mode and taking the dark-mode value down toward `#ff9a9a`-on-dark territory (or just not colouring `gesture-what` red — the left border already carries the colour) would clear it.

- **The added-suggestion state is a disabled-looking control that is still clickable.** 0.6 opacity puts it at 2.6:1 while the row keeps `cursor: pointer` and re-adds on click. The comment says that is deliberate; then it should not look disabled. Keep the sentence at full opacity and let the green ✓ carry the "already used" meaning — or show a `×2` count.

- **The armed button does change size**, contrary to the comment at `routine.css:519` ("the box stays the size it was, so pressing it does not shift the row it is in"). Measured 70×21 → 96×21, because the label goes `record` → `recording…`. It is in a `flex-wrap` row, so at narrow sidebar widths it can drop to a second line. Either keep the label and change only the fill (which is what the comment describes), or set a `min-width` sized for the longer label.

- **Two dots blink at once** — `.routine-editor-record-operation.recording .material-symbols` and `.routine-editor-recording-header .material-symbols` both run `routineRecordBlink 1.4s infinite`, permanently, a few pixels apart. One is enough; I'd keep the header's. (`prefers-reduced-motion` is handled for both — good.)

- **Phone (390×844):** the panel itself lays out fine and nothing overflows (checked `scrollWidth > clientWidth` across the whole subtree — clean), but the room is a 354×221 strip below the sidebar and widgets render at ~18px, so the gesture the feature depends on is barely performable. The record button is offered identically at every width with no hint of that.

  armed at 390×844

- Dark mode is otherwise fine — the frame, the panel border and the ✓ all read:

  dark mode, one suggestion added

---

- **The empty-state hint is the best-written text in the feature** and it is doing real work ("the room does not know which reading you meant, so it offers the likely ones"). One gap: it never says how to *stop*. Append "Press **done** when the routine says what you meant." — that also explains what the blue button in the corner is for.

- **Some suggestions arrive as unfilled templates.** Recording a click on a plain button offered `Set property of jkmt1 to number or text` ("change a property of it"). That reads like a form you forgot to fill in, and clicking it drops an incomplete operation into the routine. Either seed a plausible property or leave the reading out — a card of three where one is a placeholder undermines the other two.

- **Nothing in the room says which routine is being recorded into.** The red frame says "you are recording"; the destination is only visible in the sidebar, which is scrollable and often scrolled elsewhere by the time you have done three gestures. A single quiet caption inside the frame ("→ Deal a card · click") would remove the doubt at no layout cost.

- **The tooltip on each suggestion row is good** ("Add this operation to the routine (MOVE)") and keyboard focus works — `[tabindex]:focus-visible` is already covered by the generic `.routine-editor` rule, and Enter/Space activate. Nice.

- **A click in the room is silently swallowed while recording.** That is the right behaviour, but from the outside it looks like the app stopped responding to selection. The crosshair suggests "pick something", which is close to the opposite of what happens. Worth a clause in the hint: "clicking a widget records the click instead of selecting it".

---

## Draft tutorial — `Xtras - Recording Routines`

Played the draft in the PR's `pr-test-ai` room end to end as a first-time user (single variant), at 1280×800, 820×1180 and 1920×1080. (390×844 portrait is not reviewable — the app puts up its own "Please rotate your device to landscape" screen.) The flow works: sit down → `ctrl-j` → click **Deal a card** → unfold **click** → **record** → drag a card from the deck holder into a hand produced exactly the eight-reading card the PR describes, including the delta-derived `Turn card_S_4K face up`.

the tutorial board

**Ordering / pacing — the main issue.** The walkthrough is the last sentence of the ~100-word intro at the top, but the *prerequisite* is the third clause of the full-width paragraph **below** the objects: "…`deal one to every player` — **click to sit** first, a seat only takes cards while somebody is in it." A first-timer reads top-down, presses record, and never sees that until after it has gone wrong. Suggest a short numbered list right under the title — 1 sit down, 2 `ctrl-j`, 3 click **Deal a card**, 4 unfold **click**, 5 press **record** — and move the seat note beside the seats where it applies.

**The tutorial does not end where reading ends.** "Press **done** when the routine says what you meant, then click the button to try it" — the finish line — is in the middle-right caption, and the last block on the board is another paragraph of theory. Move the "now try it" instruction to the bottom so the board ends with an action.

**Density and shape.** Two full-width prose blocks bracket the objects (intro at the top, the dealing/collecting paragraph below), which makes the board read as an article with illustrations rather than a table with notes. The captions between them are already good and carry most of this material — trimming the intro to two sentences and folding the bottom paragraph into the seat/hand captions would let the objects lead.

**Board legibility for a newcomer:**
- the **discard holder** is an unlabelled white rectangle, visually identical to the two hand rectangles at the bottom of the board; only the caption underneath identifies it. The deck holder at least carries its "Recall & Shuffle" button. Give it an on-board label.
- **`die1`** is about half a card, drawn as a single centred dot — at board scale it reads as a bullet point. Card-sized, or with pips, it would read as a die without the caption having to say so.
- alignment is slightly ragged: the die caption is indented relative to the "Deal a card" caption directly above it, and the three caption columns do not share left edges with the objects they belong to. There is also a dead vertical band between the die row and the bottom paragraph.

tutorial at 820×1180

**Library tile.** Reads cleanly at shelf size and follows the `Xtras` / two-line-name convention of its neighbours. One note: the heavy black rounded box with a white triangle reads as a **play** button, while the actual record affordance — the red dot — is a small dot beside it. Swapping the emphasis (big red dot, small frame) would say "record" at a glance.

the draft on the shelf

VTT-AI-Agent:
Thanks — this was a genuinely useful review; nearly all of it is fixed in `efb2f3d8` (plus the re-published draft tutorial). Point by point:

**1. Drag toolbar doesn't step aside.** Fixed — you were exactly right about the specificity. The rule now also matches `.active`:

```css
body.editorRoutineRecording #editorDragToolbar,
body.editorRoutineRecording #editorDragToolbar.active { display: none; }
```

`routine.css` is bundled after `dragtoolbar.css` (`server/minify.mjs`), so the equal-specificity rule wins on source order. Verified live in both themes while armed: `display: none`, `0×0`.

**2. Lead each suggestion with the intent.** Done — took your first option. The reading ("deal one to every player") is now the lead line in `--routineOperationColor` at 12px; the generated sentence follows in the italic aside style. I did not add the three family subheads on top of that: with distinct lead lines the card scans fine, and the delta-derived reading already reads as a different kind of thing ("card_S_4K ended up on that face").

**3. Name the widgets the way the author sees them.** The headline now prefixes the type: `dragged the card card_S_4K from the holder deckHolder to the holder hand-p1`, and in your pile fixture `... from the pile o0ur to hand`. I did not add the hover-highlight-in-the-room link — good idea, but it wants the widget picker's highlight machinery and belongs in its own change.

While reproducing your pile case I found something worse than the ids: the pile the drop *creates* was contributing two of the eight readings (`o0ur got width while you did that`, `... height`). Only widgets that already existed when the gesture started can produce "the room did this too" readings now.

**4. A gesture can be added but never removed.** Added — a quiet `×` in the card's top-right corner, in the same red, revealed on hover. Only the reading goes; an operation already added from it stays in the routine, the way anything else added to a routine stays until it is deleted there.

**5. One state, three names, and two ways to leave it.** Took your second option, which also kills the resize nitpick: the button keeps the label `record` at every moment and only its fill changes (no more 70→96px, no more wrapping the row), its tooltip is *Stop recording*, and the header says *Recording*. `done` now has CSS — a quiet outlined button in the recorder's red rather than the sidebar's default filled blue.

**Contrast.** All measured in the running app against the actually painted background, both themes:

| element | before (light / dark) | after (light / dark) |
|---|---|---|
| `recording-header` | 4.18 / 2.64 | **5.62 / 7.18** |
| `gesture-what` | 3.50 / 3.51 | **21.0 / 18.69** |
| suggestion lead line | — | **6.69 / 9.53** |
| suggestion sentence | — | **8.52 / 9.18** |
| added row | 2.63 / 2.68 | same as above (no opacity) |

Three changes get there, and none of them touches `--routineRecordColor`, so the frame and the armed button keep their identity: the panel is drawn on the page background instead of `--backgroundHighlightColor2`; the gesture headline is text-coloured rather than red (the 3px left border already carries the colour, as you suggested); and the added row keeps full opacity, with the green ✓ carrying "already used". The row hover is now a 1px outline instead of a `--routineActiveCardColor` fill — that fill was itself putting the blue at 2.9:1 in dark mode.

**Two dots blink at once.** Fixed, kept the header's as you suggested.

**Unfilled templates.** The blank `SET` reading is gone rather than seeded — a card of three where one is a form you have to fill in undermines the other two.

**Nothing says which routine is being recorded into.** Added: a small caption in the top-left of the frame, `recording → dealButton · clickRoutine`.

**A click is silently swallowed.** The empty-state hint now says so, and how to stop: *"Drag and click widgets in the room the way the routine should — clicking one records the click instead of selecting it. … Press done when the routine says what you meant."*

**Phone (390×844).** Not changed. You're right that a 354×221 room strip makes the gesture barely performable, but hiding a feature by viewport is a bigger call than this PR should make on its own; I'd rather a maintainer decide that.

---

**Draft tutorial** — reworked and re-published to `pr-test-ai` ([board](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/tutorial-board.png), [tile beside `Xtras - Menus`](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/tutorial-tile.png)):

- the prerequisites are a numbered **To try it** strip directly under the intro (1. sit down · 2. `ctrl-j` · 3. click **Deal a card** · 4. unfold *click* · 5. press **record** · 6. drag a card) — not a clause below the objects;
- the "now try it" line is the last block on the board;
- the intro is two sentences and the bottom full-width paragraph is folded into a note beside the seats, where the seat rule applies;
- the discard holder carries a `Discard` label, in the place the deck holder's "Recall & Shuffle" button occupies;
- the die is 120×120 on a five-pip face instead of 70×70 on one pip;
- objects and captions share left edges (120 / 480 / 840 / 1180) and the dead band before the bottom is gone;
- the tile leads with a big red record dot on a small board frame instead of a play-shaped glyph.

Validator exits 0 and I walked it end to end again: shelf (with preview image) → sit → `ctrl-j` → **Deal a card** → *click* → **record** → drag a card into the hand → *deal one to every player* → **done** → leave edit mode → press the button, both seats get a card. [Screenshot of the panel mid-recording in the tutorial.](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/recording-in-the-tutorial.png)

`npm test` green (1288, two new tests for the `×` and the created-widget filter).

VTT-AI-Agent (commented):
## TL;DR

The routine recorder itself looks safe and is exposed through the real visual routine editor, with a tutorial-style demo in `pr-test-ai` and no library changes. One validator edge case remains: `TURN` in `seat` mode accepts targets that the engine cannot use as seat IDs.

## Findings

### 1. Validate the `turn` value against the selected turn cycle — `validator/validate_gamefile.js:893`

The new predicate treats `first`/`last` and the ID of **any** existing widget as valid regardless of `turnCycle`. For example, both `{ func: 'TURN', turnCycle: 'seat', turn: 'button1' }` (where `button1` is a button) and `{ func: 'TURN', turnCycle: 'seat', turn: 'first' }` currently validate with no problems. At runtime, however, seat mode searches only the active seat collection for an exact ID and, when it cannot find one, reports a problem and falls back to the first seat (`client/js/widgets/widget.js:2603-2608`). The validator therefore gives authors a clean result for routines that do something different from what they requested.

Please make this check conditional on `context.operation.turnCycle`: in `seat` mode require an existing widget whose type is `seat` (while still allowing dynamic `${...}` values through the existing generic handling); in other modes retain the integer/`first`/`last` rules. A small validator regression test for a non-seat ID and for `first` under `turnCycle: 'seat'` would cover both false negatives.

## UI, demo, and engine-impact checks

The new capability is available directly beside **add operation** in the visual routine editor, so no widget-property or routine-operation UI entry is missing. The live `pr-test-ai` room exercises recording moves, automatic holder effects, dealing, clicks, and dice, and the diff does not touch `library/`. No engine files are changed; the additions in `client/js/mousehandling.js` and `client/js/main.js` are gated to edit-mode recording/cleanup, so existing games' runtime behavior is unchanged.

The `beta` label was already present; I left it in place because this is an editor/validator-only feature whose input hooks are edit-gated and the green CI/CodeQL results make it safe and useful for player testing.

VTT-AI-Agent:
Thanks — the finding is right, and it is fixed in `3e109f50`.

**`TURN`'s `turn` is now validated against the turn cycle it is used with.** The predicate I added earlier accepted *any* existing widget id in *every* mode, and the custom check next to it only caught the reverse direction (an id outside seat mode). So both of your examples validated clean:

```json
{ "func": "TURN", "turnCycle": "seat", "turn": "button1" }
{ "func": "TURN", "turnCycle": "seat", "turn": "first" }
```

while `client/js/widgets/widget.js:2603` only ever does `c.find(w => w.get('id') == a.turn)` over the active seats and falls back to `c[0]` after pushing a problem — `first`/`last`/a number are no more usable there than a button id.

The property validator is now conditional on `context.operation.turnCycle`:

- `turnCycle: 'seat'` → the value has to be an existing widget of type `seat`. I reused `getWidgetTypeValidator([ 'seat' ])` so the wording matches the other id parameters (`DEAL`'s `from`/`to`), with the reason appended: `'button1' is not a valid widget type (found button - valid types: seat) - with turnCycle 'seat', turn gives the turn to the seat with that id`.
- every other cycle → whole number / `first` / `last`, and the message now says where a seat id *would* work: `'seat2' is neither a whole number, 'first' nor 'last' - turn names a seat by its id only with turnCycle 'seat'`.
- a `turnCycle` the routine works out while it runs (`${...}`) leaves `turn` alone rather than guessing which set of rules applies. Dynamic `turn` values were already handled generically before any property validator runs.

That made the `customRoutineChecks` entry redundant, so it is gone — the two directions are one predicate now instead of a predicate plus a special case.

**Regression tests** — `tests/validator/turn.test.js`, six cases: the counting cycles accept numbers/`first`/`last`, a seat id is rejected under `position`, a seat id is accepted under `seat`, a non-seat widget and an unknown id are rejected under `seat`, `first`/`last`/`2` are rejected under `seat`, and a `${...}` turn cycle is left alone. `npm test` green (1294).

**No new false positives on the public library.** I ran the validator over all 575 game and tutorial variants before and after: the problem set is byte-identical except for one message that changed wording — `library/games/Irrigation/0.json turn_button[NoPlayerTurnRoutine.1.turn]`, which points `turnCycle: 'seat'` at `0_seat` while the game's seats are `1_seat`…`5_seat`. That one was already reported before this commit, so it is a pre-existing problem in that game, not something the tightened rule introduced.

---

One thing I found while checking this, which I have **not** changed here because it is on `main` and predates this PR: the visual routine editor's `turn` parameter is declared once for all five variants (`client/js/editor/controls/routine.js:1338`, `special: [ 'first', 'last' ], widgetType: 'seat'`), so the popup offers `first`/`last` *and* a seat picker no matter which cycle is selected — the picker offers seats under "by its position", and the special values under "a specific seat". Making that variant-aware needs `parameterSpec()` to merge a per-variant override (and would move the `TURN.turn: 'seat'` entry in the presets test), which is a change to the routine editor's parameter machinery rather than to the recorder. Happy to do it, but it looks like its own PR to me — say the word and a maintainer can kick one off.

VTT-AI-Agent (commented):
## TL;DR

The recorder is well isolated (every hook is `edit`-gated and a no-op when nothing is being recorded), no engine file is touched, `library/` is untouched, and all checks are green. Two suggestion-quality bugs are worth fixing before merge: **a drop that lands on an auto-created pile only offers a `MOVE to: <pile>` that the validator rejects and whose target will not exist when the routine runs**, and **the recorder classifies click vs. drag differently from the engine**, so a slightly shaky click on a table widget silently loses the click readings (the die roll among them).

## Findings

### 1. A drop onto a pile suggests an operation that is invalid and points at a widget that will not exist

`describeRoutineGesture()` filters widgets the gesture created out of `changes` only (`client/js/editor/controls/routinerecorder.js:245-248`), but `gesture.to` is taken straight from the widget's new parent (`routinerecorder.js:262`) with no such filter, and `reparentSuggestions()` puts it into a `MOVE` regardless of what type it is (`routinerecorder.js:403-409`).

Concrete failure — the most ordinary "stack these two cards" gesture on the table:

1. record into a routine, drag `card1` onto `card2` (both children of the room);
2. the engine creates a pile around them and sets `card1.parent` to it (`client/js/widgets/widget.js:3681-3695`), with a generated id like `o0ur`;
3. `from` is `null` and not holder-like, so the only gesture reading offered is `{ func: 'MOVE', collection: [ 'card1' ], to: 'o0ur' }`.

That operation is broken twice over:

- the validator rejects it the moment it lands in the routine — I ran it: `'pile1' is not a valid widget type (found pile - valid types: holder, seat)` (`MOVE.to` is `getWidgetTypeValidator(['holder','seat'], true)`, `validator/validate_gamefile.js:805`). This is exactly the class of thing the `collectionsInScope()` gate from review 1 was added to prevent;
- and `o0ur` is transient — a pile removes itself as soon as it holds a single card, so the id is gone before anyone presses the button.

The same happens when the drop target is a pile that already existed. Suggested fix: reuse `raw.existed` for `gesture.to`/`gesture.from` as well, and gate the `MOVE` readings on `isHolderLike(to)`; when the target is a pile (or a widget that appeared during the gesture), fall through to the `SET parent` + `SET x` + `SET y` triple the `to === null` branch already builds (`routinerecorder.js:396-399`) — which is what a card-onto-card drop actually means for a routine anyway. No test covers a pile target today; the created-widget test (`tests/client/routineeditor.test.js`, "a widget the gesture created is not offered as something to set properties on") only asserts the `changes` side.

### 2. Click vs. drag is decided differently from the way the engine decides it

`placeChanged` (`routinerecorder.js:250-252`) calls a gesture a drag as soon as the rounded `x`/`y` differ by 1. The engine's threshold is 10px within 250ms (`client/js/mousehandling.js:182`), and every mousemove — even a 2px one — already runs `moveTarget.move()` and then `moveEnd()`, so the widget really does end up 2px away.

Concrete failure: press a die that sits on the table, wobble 3px, release inside 250ms. The engine treats it as a click (it calls `editClick()` → `handleRoutineRecorderClick()`, `client/js/editor/selection.js:241`), but the recorder reports `dragged die1 to 402, 301` and runs `repositionSuggestions()` — so the card offers `Set x` / `Set y` / `Sort` / `Shuffle` and never offers "roll it the way clicking it does", which is the whole point of the dice work in the Discord follow-up. On a trackpad or a touchscreen this is not an edge case. (Inside a holder it happens to be masked, because the holder snaps the child back and `placeChanged` comes out false.)

Since `handleRoutineRecorderClick()` is invoked from precisely the branch where the engine decided "this was a click", the cheapest fix is to have it set a flag on `openRoutineGesture` and have `routineRecorderPointerUp()` prefer that flag over the coordinate comparison (falling back to the current logic for a release that never reached `editClick`, e.g. over the sidebar).

### 3. Engine-internal counters become `SET` suggestions, and the `FLIP`/`SET activeFace` split is keyed on the type name

`routineRecorderIgnoredProperties` (`routinerecorder.js:44-48`) does not list `rollCount`. `Dice.click()` sets both `activeFace` and `rollCount` (`client/js/widgets/dice.js:179-181`), so any gesture during which the room rolls a die (a holder's enter routine that `CLICK`s one, say) produces `die1 got rollCount while you did that` → `SET rollCount: 4` next to the useful `SET activeFace`. `rollCount` only feeds `rollHash`/the roll animation; it validates, so it is noise rather than an error, but it is noise in a list that is capped at 8.

Related, and a simplification: the dice special case at `routinerecorder.js:532-535` tests `widget.get('type') == 'dice'`, but the actual rule is "does this widget have a `flip()`" — only `Card` and `BasicWidget` define one (`client/js/widgets/card.js:261`, `client/js/widgets/basicwidget.js:93`); `Dice` and `Spinner` extend `Widget` directly. `typeof widget.flip == 'function'` expresses the reason and keeps working if another widget type gains or loses faces.

- **No engine change.** `client/js/widgets/`, `compute.js` and `serverstate.js` are untouched. `mousehandling.js` (2 hooks), `selection.js` (3) and `main.js` (1) are all behind `edit &&` or return immediately when `isRoutineRecording()` is false, so a player who never opens the editor sees byte-identical behavior.
- **Delta timing.** Local deltas run through `receiveDelta()` synchronously inside `sendDelta()` (`client/js/serverstate.js:613-618`), i.e. before the `finally` in `inputHandler` fires `routineRecorderPointerUp()` — so the "the room did this too" readings really are available at describe time, and `getPlayerDetails` is on the list main.js hands to the editor bundle (`client/js/main.js:722`), so the foreign-delta filter works outside the jest stub too.
- **`TURN`'s `turn` predicate.** I ran it against the engine's implementation (`client/js/widgets/widget.js:2597-2613`): seat mode really does `c.find(w => w.get('id') == a.turn)`, `position` accepts `first`/`last`/integer, and the other cycles fall back to `Number.isFinite(a.turn) ? a.turn : 1` — the new rule matches. `${...}` values are short-circuited by the dollar-sign pre-check before the predicate runs, and a defined variable (`VAR seatId` → `TURN turnCycle: 'seat', turn: '${seatId}'`) validates clean, so the new predicate adds no false positives there.
- **`MOVE from: <holder>, count: 1` with no `to`** (the holder-click reading) errors at run time, but it is the same shape as the editor's own `newOperation: { func: 'MOVE', from: null, count: 1 }` (`client/js/editor/controls/routine.js:934`) and renders as a blank chip to fill in, so I would leave it alone.
- **No injection surface** — everything user-derived goes through `textContent`; the two `insertAdjacentHTML` calls take literal strings.
- **Demo room.** Built to tutorial quality, published as a draft to the PR's `pr-test-ai` room, `library/` untouched in the diff — exactly what the standing rule asks for.
- **UI rule.** No new widget property or routine function is added, and the feature itself is exposed through the visual routine editor, so there is nothing owed to `jeCommands` or the property editor.
- All CI checks (CodeQL, Analyze, the full TestCafe matrix) pass on the head commit.

The `beta` label was already on this PR and I left it there: the diff touches no engine file and its `mousehandling.js`/`main.js`/`selection.js` edits are edit-mode-gated no-ops for players, so it is safe for beta testing, and the two findings above only affect which suggestions the recorder offers a game author — nothing that can corrupt a game or crash the server.

VTT-AI-Agent:
Thanks — both suggestion-quality findings were real and are fixed in `ea062f4a`, along with the two minor ones.

**1. A drop onto a pile suggested an operation that is invalid and points at a widget that will not exist.** Fixed, and I went one step further than filtering the created ones out.

The root of it is that a pile is never a destination an operation can name: `MOVE.to` takes holders and seats, and the pile the drop invents is gone again as soon as it holds one card. That is just as true for a pile that already existed — dropping a third card on the stack from your example gave the same `MOVE to: 'o0ur'`. So instead of a `raw.existed` check, `to` is now resolved to the nearest place an operation *can* name, adding up the coordinates on the way (`routineRecorderDestination()`):

```
a card dropped onto a card on the table   →  to: null,   x/y: where the pile stands
a card dropped onto a card inside a hand  →  to: 'hand', x/y: inside the hand
```

`to` is therefore always `null` or holder-like when `reparentSuggestions()` sees it, and the `MOVE collection: [...] to: <pile>` reading cannot be built any more. Your exact case now reads as the `SET x` / `SET y` pair rather than as a reparent at all, because `from` and `to` are both the table — which is what "stack these two cards" means for a routine.

Walked in the running app (a real drag, not a stub), with the recorder armed on a button's click routine:

```
=== card onto card on the table ===
  gesture: dragged the card card1 to 700, 100
    - put it exactly there  |  Set x of card1 to 700
    - and at that height    |  Set y of card1 to 100
  (parent of card1 afterwards: 8ydu, type pile)

=== card onto an existing pile ===
  gesture: dragged the card card5 to 700, 100
    - put it exactly there  |  Set x of card5 to 700
    - and at that height    |  Set y of card5 to 100

=== card onto card inside a holder ===
  gesture: dragged the card card4 to 0, 0
    - put it exactly there  |  Set x of card4 to 0
    - and at that height    |  Set y of card4 to 0
    - re-order what is in there instead  |  Sort hand1
    - shuffle it instead                 |  Shuffle hand1
```

The coordinates are summed up the chain because `x`/`y` are counted from whatever the widget is in: the card sits at 0,0 inside a pile that stands at 700,100, and `Set x of card1 to 0` would have put it in the corner of the room. (`from` needs no equivalent: it is taken at the press, so it can never be a widget the gesture created. A pile it *emptied* is already handled — every `from`-based reading is behind `isHolderLike(from)`.)

New test: *stacking two cards on the table puts one where the other is, not into the pile the drop invented*. The created-widget test also asserts now that no reading names `pile1` at all, not just that its `width` is not offered.

**2. Click vs. drag is decided differently from the engine.** Fixed the way you suggested. `handleRoutineRecorderClick()` sets `openRoutineGesture.clicked`, and `describeRoutineGesture()` prefers that over the coordinate comparison — which stays for a release that never reached `editClick` (over the sidebar, over an overlay). Since the hook runs from exactly the branch where the engine decided "this was a click", and synchronously before the `finally` fires `routineRecorderPointerUp()`, the two can no longer disagree.

Live, with a 3px shake on a die inside 250ms, and the same die genuinely dragged afterwards:

```
=== a 3px shake on a die ===
  gesture: clicked the dice die1
    - roll it the way clicking it does                      |  Click die1
    - put it on a face instead of rolling it - it is showing 1  |  Set activeFace of die1 to 0

=== a real drag of the die ===
  gesture: dragged the dice die1 to 1364, 522
    - put it exactly there  |  Set x of die1 to 1364
    - and at that height    |  Set y of die1 to 522
```

New test: *a release the engine calls a click is recorded as one, however far the widget slid*.

**3. `rollCount` and the `FLIP` split.** Both taken.

- `rollCount` is in `routineRecorderIgnoredProperties` now, with a line saying why it is engine-internal: it is what makes a die that lands on the face it already showed still roll on screen. The die-rolled-by-the-room test asserts it is not offered.
- the split is `typeof widget.flip == 'function'` — which is literally what `FLIP` itself does (`c.flip && await c.flip(...)`, `widget.js:1704`), so a spinner now gets `SET activeFace` too instead of a `FLIP` that would pass over it. The test stub grew a `flip()` for the types that have one.

While checking that last one I found one more of the same kind: a die whose `faces` are still empty — one just dropped into the room — counts `activeFace` modulo zero, so `diceActiveFace()` returned `NaN` and the card offered *Set activeFace of die1 to NaN*. It falls back to face 0 now.

`npm test` green (1296).

VTT-AI-Agent:
Feedback from the maintainer, typed into the bot's open-PR overview page rather than here:

> less options, less technical. options should be what an inexperienced user would want. clicking dice: roll dice or set active face. clicking canvas: reset, set color, set brush size. move card: deal cards to hands, recall if moved to deck's parent. stuff like that; nothing too fancy.
>
> also: if the user moves 5 cards between the same holders, update the suggestions to MOVE 5 instead of listing 5 new suggestion blocks
>
> when a user picks one of the options, close the box. if the user wants multiple options, they have to do the action again.
>
> hover state of the "record" button has bad font color - barely readable. the same applies to the "add X" buttons with the same style

All four are in `1902fd63`, and the description above was rewritten around them.

- **Fewer, plainer readings.** The cap is 5 instead of 8 and most cards come in under it. A drag between holders reads as the move it was, as dealing to that seat and to every player, and as putting every card back into the holder the deck lies in; `MOVE count: all`, `MOVE … face`, "take one from every player", the shuffle-afterwards and the `DEFAULT`-collection reading are gone (and with them `collectionsInScope()`, which nothing else used). A die offers the roll and the face and nothing else. A **canvas** is new: wipe it (`CANVAS mode: reset`), the color it draws in (`SET activeColor`, with the color named in the reading) and how thick the brush is (`SET lineWidth`) — the three things its own buttons do. A holder offers shuffle and recall; `SORT`, "move something out of it" and `CALL clickRoutine` are gone.
- **Five deals are one card.** A holder-to-holder drag carries a key of its two holders; a later drag with the same key raises the count instead of adding a card — `dragged 5 widgets from the holder deckHolder to the holder hand-p1`, and every reading counts 5. This needed one more fix to work in a real room: the second card dealt out of a holder comes off a stack, so its parent is the pile that stack formed. The source of a drag is now resolved up to the nearest holder or seat the way the destination already was.
- **Picking closes the card.** The tick that marked a used reading is gone with the bookkeeping behind it; the recording carries on for the next gesture.
- **The hover was a fill nobody asked for.** Those buttons set only `border-color` and `color` on hover, so `button:hover { background-color: var(--VTTblueDark) }` from `layout.css` filled them with dark blue under `--VTTblue` text: **1.95:1** light, **2.21:1** dark, measured in the running app. They keep no fill now, and on the dark theme they take the teal the popup buttons already hover in (plain `--VTTblue` on the near-black page is 3:1). After: **7.22:1** and **12.32:1**. The fix reaches `add operation`, `add routine`, the property panel's `add` and its two buttons — every button with that shape. The armed record button and `done` are meant to be filled and are unchanged.

| | light | dark |
|---|---|---|
| hover, before | [screenshot](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/hover-before-light.png) | [screenshot](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/hover-before-dark.png) |
| hover, after | [screenshot](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/hover-after-light.png) | [screenshot](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/hover-after-dark.png) |
| the new cards | [screenshot](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/light-suggestions.png) | [screenshot](https://agent.virtualtabletop.io/reports/pr/1538058046285160518/dark-suggestions.png) |

VTT-AI-Agent:
> set x+y must be offered in _one_ suggestion but added as two operations

Done — a reading of a gesture is now a list of operations rather than a single one, and picking it adds all of them in one go, in order, right where "add operation" would have added them.

The drag that puts a widget at a spot is one row again:

```
dragged the card free1 to 715, 870                     ×
  + put it exactly there
        Set x of free1 to 715
        Set y of free1 to 870
```

Clicking it writes `SET x of free1 to 715` followed by `SET y of free1 to 870`. The row spells out every operation it would add under the line that says what the reading is for, so the aside stays the check on what is about to be written.

The same applies to taking a widget out onto the table, which was three rows (`do exactly this` / `and put it exactly there` / `and at that height`): it is now one reading of `SET parent: null`, `SET x` and `SET y`. Those "and …" labels only existed because the clauses of one reading had to be separate rows — half of a reading picked is a widget somewhere nobody dropped it, which was the bug.

Screenshot of the panel: <https://agent.virtualtabletop.io/reports/pr/1538058046285160518/one-reading-two-operations.png>

`npm test` is green (1297) and the TestCafe `routines.js` suite passes locally (7/7); the two-operation reading was also walked in a browser (free drag on the table and a card taken off a stack), which is where the screenshot comes from.

PR #3124, opened 2026-08-17
Offer each routine the values and collections its trigger hands it, in the routine editor

VTT-AI-Agent:
The routine editor in the Edit Widgets **Automations** section only ever offered two kinds of value: what an earlier operation of the routine stored, and the nine values (and three collections) that every routine has. The ones a routine gets *because of what started it* were nowhere in the UI, so the only way to find out that an `enterRoutine` is handed the collection `child` and the variable `oldParentID` was the wiki.

This adds them. Every routine now lists what its trigger hands it, in the value and collection sections of the parameter popups, as its own group named after the routine — for example "In every enterRoutine" between "From earlier operations" and "In every routine" — and what each entry holds is written out on a line at the foot of the open section, filled by the entry the pointer or the keyboard is on.

What is offered where (taken from the `evaluateRoutine` calls in `widget.js`, `statemanaged.js` and `serverstate.js`):

| routine | values | collections |
| --- | --- | --- |
| `changeRoutine` | `property`, `value`, `oldValue` | – |
| `<property>ChangeRoutine` | `value`, `oldValue` (worded with that property) | – |
| `globalUpdateRoutine` | `widgetID`, `property`, `value`, `oldValue` | `widget` |
| `<property>GlobalUpdateRoutine` | `widgetID`, `value`, `oldValue` | `widget` |
| `enterRoutine` | `oldParentID` | `child` |
| `leaveRoutine` | – | `child` |
| `gameStartRoutine` | `widgetID` | `widget` |
| any custom routine | the `arguments` of every CALL in the room that runs it on this widget, in a group of their own | `caller` (from the CALL that runs it) |
| a routine a CALL runs that a player can trigger too | the `arguments` of those CALLs, under "Only when a CALL runs it" | `caller`, same group |
| the `loopRoutine` of a FOREACH | `key`/`value` over a list, `value` over a range, `widgetID` over a collection | `DEFAULT` over a collection |

`clickRoutine` and `doubleClickRoutine` get nothing of their own from the engine, so nothing is added for them - unless a CALL in the room runs one of them by name, which hands it arguments and `caller` like any other routine. Those are listed in a group of their own and dimmed, because the player clicking the widget hands over none of them.

Which CALLs count is decided by the widget they run the routine on: a CALL without a `widget` runs it on the widget it is written on (in a deck's `cardDefaults`, on the card running it), an id written out is looked up, and an id the game computes can be any widget and is included. So two widgets with a routine of the same name are not shown each other's arguments.

Where two of those hand the same name over, the one that is established last wins and the others are not listed, so a name is never offered twice for two different things:

- a trigger hands its values over *before* the first operation, so an operation of the routine storing the same name wins: in a `changeRoutine` that starts with a GET into `value`, `value` stays under "From earlier operations" from the second operation on — and says so ("stored by an earlier operation - it replaced the value the changeRoutine hands over").
- a FOREACH establishes its round *after* everything outside the block, so inside the loop `${value}` is the loop's and neither an operation before the block nor the trigger is offered under that name. An operation *inside* the block runs later still and wins again, the same way as above. The loop groups are listed nearest first: "In this loopRoutine" is the loop the block is in, then "In the loopRoutine around it", "In the loopRoutine 2 levels out".

Three things came out of writing that table down:

- The routine cards of the Automations section claimed an `enterRoutine`/`leaveRoutine`/`globalUpdateRoutine` "starts with the widget picked". It does not — the widget is in a collection named `child` resp. `widget`, while operations default to `DEFAULT`, so following that description gave you a routine that quietly did nothing. The descriptions now name the collection, with the literal names in quotes so they do not read as prose.
- CALL hands over more than `caller`: it copies the whole variable and collection set of the calling routine into the routine it runs and merges the `arguments` on top (`widget.js`). The card of a custom routine says so now, and the argument names are collected from every CALL in the room that names the routine — each saying which CALLs hand it over, dimmed when not all of them do.
- The validator flagged `${widgetID}` and the `widget` collection in a `gameStartRoutine` as undefined, which is the same table being incomplete one file over. `validate_gamefile.js` now knows them (`enterRoutine`/`leaveRoutine`/`changeRoutine`/`globalUpdateRoutine` were already covered there), and it knows `caller` in a custom routine that no CALL reaches, which used to be reported as an undefined collection on top of being uncalled.

### The list itself

The only CSS in the diff, all in the parameter popups, from the UI/UX review:

- **A description line at the foot of the open section.** The descriptions are the point of this PR and they used to exist only as `title=` tooltips — never shown on a touch screen, never to the keyboard, a second late on the desktop. One line, filled on hover *and* on keyboard focus, with two lines kept free so the list does not move while the pointer travels along it. The tooltips stay.
- **A hairline above each group caption**, so which group a name is in is readable at a glance now that a nested FOREACH stacks four of them.
- **A long name wraps** instead of running out of the popup: an argument name comes from another widget's CALL, so the author of the routine cannot shorten it, and at 390px width it overflowed the popup by 25px.
- **The description line sticks to the foot of the popup**, so it stays on screen in a section too tall for the window instead of being scrolled out of sight exactly while the pointer - or the keyboard, which scrolls the focused name into view - is on a name near the top.

One thing outside the popups: inside a `FOREACH` block, a VAR is offered the loop's own names (`value`, `key`, ...) among the names it can write to again, and the trigger's names as well. Writing over what the round is for is exactly what an operation in the block may want to do.

### Verified

- `npm test` — 1265 jest tests pass.
- `npx testcafe chromium:headless tests/testcafe/editor.js` — including a test that walks every case of the table above through the editor UI, plus the shadowing rules, the nested-FOREACH group titles, the description line and the "not by the other operation that runs this routine" wording.
- The editor test also covers the arguments of two CALLs in one routine, a CALL in the `cardDefaults` of a deck (its cards run those routines), an argument named `constructor`, two widgets with a routine of the same name, a `clickRoutine` run by CALL, the VAR suggestions inside a FOREACH and the description line staying in view at 1280x320.
- `node validator/validate_gamefile_node.js` on a game whose only custom routine is never called: no undefined-collection problem for `caller`.
- Clicked through in a browser at 1280×800 and 390×844: [a custom routine's arguments at phone width](https://agent.virtualtabletop.io/reports/pr/1538977191579099247/uiux-390-customRoutine-values.png) (the long name wraps, the arguments not every call passes are dimmed), [a nested FOREACH](https://agent.virtualtabletop.io/reports/pr/1538977191579099247/uiux-1280-nested-foreach.png) (group rules, the description line at the foot), [a shadowed preset](https://agent.virtualtabletop.io/reports/pr/1538977191579099247/uiux-1280-shadowed-value.png).

### Try it

A demo room, **Routine presets demo**, is uploaded into this PR's `pr-test-ai` test-server room (it is not part of the diff and not in the public library; it can be added to `library/tutorials/` on request). It now has one block per row of the table, in two aligned columns: drop the token into the holder (`enterRoutine`/`leaveRoutine`, and the `rotationGlobalUpdateRoutine` that hears the rotation it causes), drag it around (its own `changeRoutine`, which is told which property changed), type in the label (`textChangeRoutine`), click "name the chips" (a FOREACH over a group of widgets), and the two buttons on the right run the same custom routine with different arguments. Its widgets are meant to be opened in edit mode: pick a routine under Automations, click one of its blanks and the group named after the routine is right there. The file is also at [Routine presets demo.vtt](https://agent.virtualtabletop.io/reports/pr/1538977191579099247/Routine%20presets%20demo.vtt).

No new widget property and no new routine function, so there is nothing to add to `jeCommands`; the change is entirely in the visual routine editor.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3124/pr-test (or any other room on that server)

### DIFF
--- a/client/css/editor/controls/popup.css
+++ b/client/css/editor/controls/popup.css
@@ -490,17 +490,51 @@ body.edit.darkMode .inline-popup button.primary:hover {
   color: var(--VTTblue);
 }
 
+/* a name is a name: one that came from somewhere else (an argument a CALL in
+   another widget hands over) can be longer than the popup is wide, and wrapping
+   it is the only way it does not run out of the popup on a phone */
 .inline-popup .popup-entry {
   display: inline-block;
+  max-width: 100%;
+}
+
+.inline-popup .popup-entry button {
+  max-width: 100%;
+  overflow-wrap: anywhere;
+  text-align: left;
+}
+
+/* one the routine only has when a particular operation is what ran it */
+.inline-popup .popup-entry button.popup-entry-partial {
+  opacity: 0.65;
+  border-style: dashed;
 }
 
 /* where the values of a section come from: a quiet line rather than a heading
    in a color of its own - the section around it already says what kind of value
-   its entries are, and they carry that color the way the sentence does */
+   its entries are, and they carry that color the way the sentence does. Which
+   group a name is in is what the section is about once there are three or four
+   of them, so each one is separated from the run of chips above it by a rule */
 .inline-popup .popup-value-group {
   font-size: 11px;
-  opacity: 0.7;
+  opacity: 0.85;
   margin: 8px 2px 2px 2px;
+}
+
+.inline-popup .popup-entry + .popup-value-group {
+  border-top: 1px solid var(--backgroundHighlightColor1);
+  padding-top: 6px;
+}
+
+/* what the name the pointer or the keyboard is on holds - see
+   renderRoutineValueKindSection. Two lines are kept free whether one is shown or
+   not, so the list above it does not move while the pointer travels along it */
+.inline-popup .popup-entry-description {
+  margin: 8px 2px 0 2px;
+  font-size: 12px;
+  opacity: 0.8;
+  min-height: 2.8em;
+  overflow-wrap: anywhere;
 }
 
 .inline-popup button[data-kind=variable] {
--- a/client/css/editor/controls/popup.css
+++ b/client/css/editor/controls/popup.css
@@ -528,11 +528,22 @@ body.edit.darkMode .inline-popup button.primary:hover {
 
 /* what the name the pointer or the keyboard is on holds - see
    renderRoutineValueKindSection. Two lines are kept free whether one is shown or
-   not, so the list above it does not move while the pointer travels along it */
+   not, so the list above it does not move while the pointer travels along it.
+   It sticks to the foot of the popup while its section is on screen: a section
+   with four or five groups in it is taller than a short window, so the line
+   would sit below the fold exactly while the pointer is on a name near the top -
+   and the keyboard scrolls the focused button into view, pushing it further out
+   still. The background is the popup's own so the chips do not show through. */
 .inline-popup .popup-entry-description {
+  position: sticky;
+  bottom: 0;
+  background: var(--backgroundColor);
   margin: 8px 2px 0 2px;
+  padding-bottom: 2px;
   font-size: 12px;
-  opacity: 0.8;
+  /* quiet through the text color rather than the opacity of the whole line: a
+     line at 0.8 lets the chips scrolling under it show through its background */
+  color: color-mix(in srgb, var(--textColor) 80%, transparent);
   min-height: 2.8em;
   overflow-wrap: anywhere;
 }
--- a/client/js/editor/controls/events.js
+++ b/client/js/editor/controls/events.js
@@ -116,15 +116,17 @@ const routinePresets = {
 };
 
 // The arguments every CALL in the room that runs this routine by name hands it,
-// as { arguments: { name: which CALLs pass it }, sources: every CALL that runs
+// as { arguments: Map(name -> which CALLs pass it), calls: every CALL that runs
 // it }. CALL copies the calling routine's values and collections into the
 // routine it runs and merges its arguments on top (widget.js), so those names
 // are values the routine has before its first operation - the same kind of thing
 // an event hands its routine, except that what they are is written in the game
 // rather than in the engine.
 function callArgumentsFor(routineName) {
-  const args = {};
-  const sources = new Set();
+  // the names are written in the game, so they are kept in a Map: a plain object
+  // answers "constructor" with what it inherits from Object instead of nothing
+  const args = new Map();
+  const calls = [];
   const scan = (routine, source)=>{
     for(const operation of Array.isArray(routine) ? routine : []) {
       if(!operation || typeof operation != 'object')
@@ -132,30 +134,46 @@ function callArgumentsFor(routineName) {
       // a CALL given a list of names runs the first of them (widget.js)
       const called = Array.isArray(operation.routine) ? operation.routine[0] : operation.routine;
       if(operation.func == 'CALL' && called === routineName) {
-        sources.add(source);
+        // one entry per CALL rather than per routine one is in: two CALLs in the
+        // same routine can hand over different arguments, and an argument only
+        // one of them passes is one the routine only sometimes has
+        const call = { source };
+        calls.push(call);
         if(operation.arguments && typeof operation.arguments == 'object')
-          for(const name of Object.keys(operation.arguments))
-            (args[name] = args[name] || new Set()).add(source);
+          for(const name of Object.keys(operation.arguments)) {
+            if(!args.has(name))
+              args.set(name, new Set());
+            args.get(name).add(call);
+          }
       }
       for(const block of [ 'thenRoutine', 'elseRoutine', 'loopRoutine' ])
         scan(operation[block], source);
     }
   };
-  for(const widget of widgets.values())
-    for(const [ key, value ] of Object.entries(widget.state))
+  const scanRoutinesIn = (source, describe)=>{
+    for(const [ key, value ] of Object.entries(source))
       if(key.match(/Routine$/) && Array.isArray(value))
-        scan(value, `${key} of ${widget.state.id}`);
-  return { arguments: args, sources };
+        scan(value, describe(key));
+  };
+  for(const widget of widgets.values()) {
+    scanRoutinesIn(widget.state, key=>`${key} of ${widget.state.id}`);
+    // the cards of a deck run the routines in its cardDefaults, so a CALL in one
+    // of them runs this routine just like one written on a widget itself
+    if(widgetTypeOf(widget) == 'deck')
+      scanRoutinesIn(cardDefaultsOf(widget), key=>`${key} of the cards of ${widget.state.id}`);
+  }
+  return { arguments: args, calls };
 }
 
 // Which operations hand an argument over, named rather than counted: a count
 // ("3 of the operations that run this routine") is nothing the reader can go and
 // look at. More than two of them would fill the line, so the rest is a count.
-function describeCallers(sources, total) {
-  const names = [ ...sources ];
+// Two CALLs in the same routine are one name in the list but two operations.
+function describeCallers(calls, total) {
+  const names = [ ...new Set([ ...calls ].map(call=>call.source)) ];
   const listed = names.length > 2 ? `${names.slice(0, 2).join(', ')} and ${names.length-2} more` : names.join(' and ');
-  const missing = total - names.length;
-  return `handed over by the ${names.length == 1 ? 'CALL' : 'CALLs'} in ${listed}`
+  const missing = total - calls.size;
+  return `handed over by the ${calls.size == 1 ? 'CALL' : 'CALLs'} in ${listed}`
        + (missing ? ` - not by the ${missing == 1 ? 'other operation that runs' : `other ${missing} operations that run`} this routine` : '');
 }
 
@@ -188,23 +206,25 @@ function routinePresetsOf(property) {
     // it is one the routine only has when that call is what ran it - so they are
     // a group of their own rather than part of "In every fooRoutine", and the
     // ones not every call hands over are shown as the exception they are
-    const { arguments: args, sources } = callArgumentsFor(property);
-    const variables = {};
-    for(const name of Object.keys(args).sort())
+    const { arguments: args, calls } = callArgumentsFor(property);
+    // a name written in the game, so the map it goes into inherits nothing (see
+    // presetEntryOf in popup.js)
+    const variables = Object.create(null);
+    for(const name of [ ...args.keys() ].sort())
       variables[name] = {
-        description: describeCallers(args[name], sources.size),
-        partial: args[name].size < sources.size
+        description: describeCallers(args.get(name), calls.length),
+        partial: args.get(name).size < calls.length
       };
     if(Object.keys(variables).length)
       groups.push({
-        title: `From the ${sources.size == 1 ? 'operation that runs' : 'operations that run'} it`,
+        title: `From the ${calls.length == 1 ? 'operation that runs' : 'operations that run'} it`,
         source: 'the CALL that runs this routine',
-        shadowedByOperations: true, variables, collections: {}
+        variables, collections: {}
       });
   }
   // handed over before the first operation runs, so an operation of the routine
   // that stores the same name wins over them (see presetSections in popup.js)
-  const group = Object.assign({ title: `In every ${property}`, source: `the ${property}`, shadowedByOperations: true, variables: {}, collections: {} }, preset);
+  const group = Object.assign({ title: `In every ${property}`, source: `the ${property}`, variables: {}, collections: {} }, preset);
   if(Object.keys(group.variables).length + Object.keys(group.collections).length)
     groups.push(group);
   return groups;
--- a/client/js/editor/controls/events.js
+++ b/client/js/editor/controls/events.js
@@ -115,29 +115,62 @@ const routinePresets = {
   editorAddToRoomRoutine: {}
 };
 
-// The arguments every CALL in the room that runs this routine by name hands it,
-// as { arguments: Map(name -> which CALLs pass it), calls: every CALL that runs
-// it }. CALL copies the calling routine's values and collections into the
-// routine it runs and merges its arguments on top (widget.js), so those names
+// Where a widget runs a routine of this name from, as the place the editor
+// shows routines in - the widget's own properties, or the cardDefaults of the
+// deck a card takes what it does not have itself from (Card.getDefaultValue).
+// null for an id that is in no widget of the room, which rules nothing out.
+function routineHomeOf(widgetID, routineName) {
+  const widget = widgets.get(widgetID);
+  if(!widget)
+    return null;
+  if(Array.isArray(widget.state[routineName]))
+    return { widgetID, cardDefaults: false };
+  if(widgetTypeOf(widget) == 'card') {
+    const deck = widgets.get(widget.state.deck);
+    if(deck && Array.isArray(cardDefaultsOf(deck)[routineName]))
+      return { widgetID: deck.state.id, cardDefaults: true };
+  }
+  return { widgetID, cardDefaults: false };
+}
+
+// Whether a CALL can reach the routine that is open in the editor. It names the
+// widget it runs the routine on, and with no widget of its own it runs it on the
+// widget it is written on (widget.js) - which, in the cardDefaults of a deck, is
+// the card running it rather than the deck. An id the game computes (${...}) or
+// hands over as anything but a string can be any widget, so it stays in.
+function callReaches(operation, where, routineName, target) {
+  if(!target)
+    return true;
+  const named = Array.isArray(operation.widget) ? operation.widget[0] : operation.widget;
+  const home = named === undefined ? where : (typeof named == 'string' && named.indexOf('${') == -1 ? routineHomeOf(named, routineName) : null);
+  return !home || (home.widgetID == target.widgetID && !home.cardDefaults == !target.cardDefaults);
+}
+
+// The arguments every CALL in the room that runs the routine open in the editor
+// hands it, as { arguments: Map(name -> which CALLs pass it), calls: every CALL
+// that runs it }. CALL copies the calling routine's values and collections into
+// the routine it runs and merges its arguments on top (widget.js), so those names
 // are values the routine has before its first operation - the same kind of thing
 // an event hands its routine, except that what they are is written in the game
-// rather than in the engine.
-function callArgumentsFor(routineName) {
+// rather than in the engine. target is the routine being edited as { widgetID,
+// cardDefaults }: another widget's routine of the same name is a different
+// routine, and its arguments are none of this one's business.
+function callArgumentsFor(routineName, target) {
   // the names are written in the game, so they are kept in a Map: a plain object
   // answers "constructor" with what it inherits from Object instead of nothing
   const args = new Map();
   const calls = [];
-  const scan = (routine, source)=>{
+  const scan = (routine, where)=>{
     for(const operation of Array.isArray(routine) ? routine : []) {
       if(!operation || typeof operation != 'object')
         continue;
       // a CALL given a list of names runs the first of them (widget.js)
       const called = Array.isArray(operation.routine) ? operation.routine[0] : operation.routine;
-      if(operation.func == 'CALL' && called === routineName) {
+      if(operation.func == 'CALL' && called === routineName && callReaches(operation, where, routineName, target)) {
         // one entry per CALL rather than per routine one is in: two CALLs in the
         // same routine can hand over different arguments, and an argument only
         // one of them passes is one the routine only sometimes has
-        const call = { source };
+        const call = { source: where.source };
         calls.push(call);
         if(operation.arguments && typeof operation.arguments == 'object')
           for(const name of Object.keys(operation.arguments)) {
@@ -147,20 +180,20 @@ function callArgumentsFor(routineName) {
           }
       }
       for(const block of [ 'thenRoutine', 'elseRoutine', 'loopRoutine' ])
-        scan(operation[block], source);
+        scan(operation[block], where);
     }
   };
-  const scanRoutinesIn = (source, describe)=>{
+  const scanRoutinesIn = (source, where)=>{
     for(const [ key, value ] of Object.entries(source))
       if(key.match(/Routine$/) && Array.isArray(value))
-        scan(value, describe(key));
+        scan(value, where(key));
   };
   for(const widget of widgets.values()) {
-    scanRoutinesIn(widget.state, key=>`${key} of ${widget.state.id}`);
+    scanRoutinesIn(widget.state, key=>({ source: `${key} of ${widget.state.id}`, widgetID: widget.state.id, cardDefaults: false }));
     // the cards of a deck run the routines in its cardDefaults, so a CALL in one
     // of them runs this routine just like one written on a widget itself
     if(widgetTypeOf(widget) == 'deck')
-      scanRoutinesIn(cardDefaultsOf(widget), key=>`${key} of the cards of ${widget.state.id}`);
+      scanRoutinesIn(cardDefaultsOf(widget), key=>({ source: `${key} of the cards of ${widget.state.id}`, widgetID: widget.state.id, cardDefaults: true }));
   }
   return { arguments: args, calls };
 }
@@ -181,8 +214,10 @@ function describeCallers(calls, total) {
 // the most specific group first (see presetSections in popup.js). The routines
 // named after a property hear about that one property only, so theirs are worded
 // with its name and have no property variable. Everything else is a custom
-// routine, which only ever runs through CALL.
-function routinePresetsOf(property) {
+// routine, which only ever runs through CALL. target says which routine of which
+// widget is open ({ widgetID, cardDefaults }), so the CALLs that run a routine of
+// the same name somewhere else are left out of it.
+function routinePresetsOf(property, target) {
   let match;
   const groups = [];
   let preset = routinePresets[property];
@@ -200,28 +235,35 @@ function routinePresetsOf(property) {
       },
       collections: { widget: `the widget whose ${match[1]} changed` }
     };
-  if(!preset && !predefinedEvents.some(event=>event.property == property)) {
+  // a routine nothing in the engine triggers is a custom one: it only ever runs
+  // through CALL, so what a CALL hands over is what it always has
+  const custom = !preset && !predefinedEvents.some(event=>event.property == property);
+  if(custom)
     preset = { collections: { caller: 'the widget whose routine used CALL to run this one' } };
-    // an argument only some of the calls hand over is still worth offering, but
-    // it is one the routine only has when that call is what ran it - so they are
-    // a group of their own rather than part of "In every fooRoutine", and the
-    // ones not every call hands over are shown as the exception they are
-    const { arguments: args, calls } = callArgumentsFor(property);
-    // a name written in the game, so the map it goes into inherits nothing (see
-    // presetEntryOf in popup.js)
-    const variables = Object.create(null);
-    for(const name of [ ...args.keys() ].sort())
-      variables[name] = {
-        description: describeCallers(args.get(name), calls.length),
-        partial: args.get(name).size < calls.length
-      };
-    if(Object.keys(variables).length)
-      groups.push({
-        title: `From the ${calls.length == 1 ? 'operation that runs' : 'operations that run'} it`,
-        source: 'the CALL that runs this routine',
-        variables, collections: {}
-      });
-  }
+  // CALL runs any routine by name, a clickRoutine as readily as a custom one, and
+  // hands it its arguments and its caller either way - so they are offered
+  // wherever a CALL in the room runs this routine. They are a group of their own
+  // rather than part of "In every fooRoutine" because the routine only has them
+  // when that CALL is what ran it: an argument not every call hands over is shown
+  // as the exception it is, and in a routine a player can trigger as well, all of
+  // them are.
+  const { arguments: args, calls } = callArgumentsFor(property, target);
+  // a name written in the game, so the map it goes into inherits nothing (see
+  // presetEntryOf in popup.js)
+  const variables = Object.create(null);
+  for(const name of [ ...args.keys() ].sort())
+    variables[name] = {
+      description: describeCallers(args.get(name), calls.length),
+      partial: !custom || args.get(name).size < calls.length
+    };
+  const collections = !custom && calls.length
+    ? { caller: { description: 'the widget whose routine used CALL to run this one', partial: true } } : {};
+  if(Object.keys(variables).length + Object.keys(collections).length)
+    groups.push({
+      title: custom ? `From the ${calls.length == 1 ? 'operation that runs' : 'operations that run'} it` : 'Only when a CALL runs it',
+      source: 'the CALL that runs this routine',
+      variables, collections
+    });
   // handed over before the first operation runs, so an operation of the routine
   // that stores the same name wins over them (see presetSections in popup.js)
   const group = Object.assign({ title: `In every ${property}`, source: `the ${property}`, variables: {}, collections: {} }, preset);
@@ -249,6 +291,10 @@ const routineTargets = {
     describe: property=>property
   }
 };
+
+function widgetIDOf(widget) {
+  return typeof widget.get == 'function' ? widget.get('id') : widget.state.id;
+}
 
 function widgetTypeOf(widget) {
   return typeof widget.get == 'function' ? widget.get('type') : widget.state.type;
@@ -649,7 +695,7 @@ class EventsEditor {
           // clone in both directions: the editor mutates its own copy, and the widget must
           // never share references with it (deltas would alias widget state to the editor's
           // arrays and make later set() calls no-op because the state already "changed")
-          this.routineEditors[key] = new RoutineEditor(this.widget, JSON.parse(JSON.stringify(this.routineSource(target)[property])), [], [], { routineKey: key, presets: routinePresetsOf(property) });
+          this.routineEditors[key] = new RoutineEditor(this.widget, JSON.parse(JSON.stringify(this.routineSource(target)[property])), [], [], { routineKey: key, presets: routinePresetsOf(property, { widgetID: widgetIDOf(this.widget), cardDefaults: target == 'cardDefaults' }) });
           this.routineEditors[key].registerChangeListener(v=>this.setRoutine(entry, JSON.parse(JSON.stringify(v))));
         }
         contentDOM.append(this.routineEditors[key].domElement);
--- a/client/js/editor/controls/events.js
+++ b/client/js/editor/controls/events.js
@@ -17,22 +17,22 @@ const predefinedEvents = [
   {
     property: 'enterRoutine',
     label: 'widget enters',
-    description: 'Runs when another widget is dropped into this widget. The routine starts with the widget that entered picked, and can use oldParentID - the widget it came from.'
+    description: 'Runs when another widget is dropped into this widget. The routine is handed the widget that entered as the collection child, and oldParentID - the id of the widget it came from.'
   },
   {
     property: 'leaveRoutine',
     label: 'widget leaves',
-    description: 'Runs when a widget is taken out of this widget. The routine starts with the widget that left picked. It can run more than once for a single move.'
+    description: 'Runs when a widget is taken out of this widget. The routine is handed the widget that left as the collection child. It can run more than once for a single move.'
   },
   {
     property: 'gameStartRoutine',
     label: 'game starts',
-    description: 'Runs once when a player loads the game fresh from the game shelf or public library, but not when loading a saved game in progress.'
+    description: 'Runs once when a player loads the game fresh from the game shelf or public library, but not when loading a saved game in progress. The routine is handed the widget it is on as the collection widget and its id as widgetID.'
   },
   {
     property: 'globalUpdateRoutine',
     label: 'any widget changed',
-    description: 'Runs when any property of any widget in the room changes. The routine starts with the changed widget picked and can use value - what the property is now - as well as oldValue, property and widgetID. To react to one property only, name the routine after it, like fooGlobalUpdateRoutine for the property foo.'
+    description: 'Runs when any property of any widget in the room changes. The routine is handed the changed widget as the collection widget and can use value - what the property is now - as well as oldValue, property and widgetID. To react to one property only, name the routine after it, like fooGlobalUpdateRoutine for the property foo.'
   }
 ];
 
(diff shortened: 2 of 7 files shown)


VTT-AI-Agent (commented):
## TL;DR

Good, well-scoped change — I checked every row of the table in the description against `widget.js`, `statemanaged.js` and `serverstate.js` and all of them are right, including the `oldParentID`-is-null-on-a-drag caveat. The one thing I'd fix before merge: the shadowing rule is applied in the wrong direction for a routine's *own* trigger presets, so a name an earlier operation stored is now listed under "In every `<routine>`" with the trigger's description instead of under "From earlier operations".

## Findings

### 1. A name an earlier operation stores is re-labelled as a trigger preset — `client/js/editor/controls/popup.js:906`, `:930`

The "From earlier operations" list now drops every name that appears in *any* preset group:

```js
const presetVariables = presetNames(this.presets, 'variables');
...
{ title: 'From earlier operations', list: [ ...this.variables ].filter(variable=>presetVariables.indexOf(variable) == -1)...
```

That is the correct direction for the FOREACH groups added by `subroutinePresets` — a loop variable is established when the block is entered, i.e. after everything outside it, so it really does shadow what the enclosing routine stored. It is the wrong direction for the groups `routinePresetsOf` builds: those are handed to `evaluateRoutine` *before the first operation runs*, so an operation that writes the same name shadows the preset, not the other way round.

Concrete case — a `changeRoutine` whose first operation is

```json
{ "func": "GET", "variable": "value", "property": "foo" }
```

From the second operation on, `${value}` holds the GET result. Every chip popup after that operation now lists `value` **only** under "In every changeRoutine", with the tooltip "what the property is now" — which is exactly the wrong explanation, and it is gone from "From earlier operations" where it correctly used to be before this PR. Same for `property`/`oldValue` in a `changeRoutine`, `widgetID` in a `gameStartRoutine`/`globalUpdateRoutine`, `oldParentID` in an `enterRoutine`.

Suggestion: mark which kind of group a preset is. E.g. have `routinePresetsOf` set `shadowedByOperations: true` on the group it returns, and in `renderRoutineValueSection` filter *the preset entry* (rather than the "From earlier operations" entry) for those groups, while keeping today's behaviour for the loop groups. `presetSections` already takes the whole group list, so the flag is available where the filtering happens.

### 2. Nested FOREACHs render two groups with the identical title — `client/js/editor/controls/routine.js:3385`

`ForeachRoutineOperationEditor.subroutinePresets` appends its group unconditionally, so a FOREACH inside a FOREACH gives the inner block two groups both titled `In every loopRoutine`. With an outer FOREACH over a list and an inner one over a range, the outer group is left with just `key` (its `value` is correctly hidden by the inner one) and the inner group has `value` — two headings with the same text and nothing saying which loop each belongs to. Over a collection it is worse: `In every loopRoutine → key, value` followed by `In every loopRoutine → widgetID, DEFAULT`.

Suggestion: title them by depth from the chip — e.g. "In this loop" for the innermost and "In the loop around it" / "In the loop two levels up" for the outer ones — or merge same-titled groups after the shadowing filter.

### 3. `caller` is offered for custom routines but the validator only knows it inside a reached CALL — `validator/validate_gamefile.js:1328`

`validateRoutine` gets `caller` added at line 604, but only along the CALL path. A custom routine that no CALL in the game file reaches is validated at line 1328 with `validCollections: {...SUPER_GLOBALS.collections}` only, so following the editor's new offer produces an "undefined collection" problem on top of the existing "Routine is not being called". Low severity (the routine is already flagged), but this PR closed exactly the same table-is-incomplete-one-file-over gap for `gameStartRoutine` at line 182, so adding `caller: 1` to that context would finish the job.

### 4. The custom-routine group understates what CALL actually hands over — `client/js/editor/controls/events.js:62` and the `caller` preset

`widget.js:1485-1489` clones the caller's *entire* variable set and *every* one of its collections into the called routine, then merges `arguments` on top:

```js
const inheritVariables = Object.assign(JSON.parse(JSON.stringify(variables)), a.arguments);
...
inheritCollections['caller'] = [ this ];
```

So a routine reached by `CALL … "arguments": { "amount": 3 }` genuinely has `${amount}`, and the new card description ("It is handed the widget whose routine ran it as the collection caller") reads as if `caller` were all of it. Not a regression — but since the point of the PR is to stop making people read the wiki, it would be worth either saying so in the group's tooltip, or (better) collecting the `arguments` keys of every `CALL` in the room that names this routine and offering them in the group, the way `routinePresetsOf` already does for the property-named routines.

## What I verified

- Every row of the table in the description against the engine: `widget.js:3149/3155` (`child`, `oldParentID`, and `moveStart` at `widget.js:2860` setting `parent` to null — so the "null when a player dragged it" note is right), `statemanaged.js:132-141`, `serverstate.js:162/560`, `holder.js:109`, and `widget.js:1745-1791` for the three FOREACH variants. All correct. The editor's variant detection (`list` on `in`, `range` only when `in` is absent) also matches `validate_gamefile.js`'s `loopRoutine` validator and the engine's in-then-range-then-collection order.
- `clickRoutine`/`doubleClickRoutine` really get `{}, {}` and `editorAddToRoomRoutine` is called with no arguments at all, so the explicit `editorAddToRoomRoutine: {}` entry (keeping it out of the `caller` fallback) is needed and correct.
- `npx testcafe chromium:headless tests/testcafe/editor.js -t "A routine offers the values and widgets that what started it hands over"` — passes here (45s).
- `npm test` — the six `tests/server/minify.test.js` failures I saw are the 60s `beforeAll` hook timeout and they reproduce on unmodified `main` on this machine, so they are not from this PR.
- No `library/` or tutorial change in the diff, and the demo room lives on the PR's `pr-test-ai` room — that is the right place for it.
- No new widget property and no new routine function, so nothing is owed to `jeCommands`; the change *is* the visual routine editor, so the UI-support rule is satisfied by construction.

I added the `beta` label: the diff touches only the edit-mode routine editor, the TestCafe suite and the validator — no engine file at all (nothing under `client/js/widgets/`, no `compute.js`, no `serverstate.js`) — so it cannot change how existing games run and it is safe and useful for players to try.

VTT-AI-Agent:
Thanks — all four addressed in 1ae4d7a.

**1. Shadowing applied backwards for the routine's own presets** — fixed, and you were right about why: the FOREACH groups and the trigger groups need opposite directions. `routinePresetsOf` now marks its group `shadowedByOperations: true`, and `presetSections`/`presetNames` in `popup.js` use that flag: a flagged group drops a name that an earlier operation already stored (and leaves it in "From earlier operations"), an unflagged one — the FOREACH groups — keeps today's behaviour and takes the name out of "From earlier operations". Your exact case is now a test assertion: a `textChangeRoutine` starting with `{ "func": "GET", "variable": "value", "property": "foo" }` lists `value` under "From earlier operations" from the second operation on, while `oldValue` stays under "In every textChangeRoutine".

**2. Nested FOREACHs render two groups both titled "In every loopRoutine"** — fixed. `subroutinePresets` now re-titles the loop groups it inherits (on a copy, they belong to the outer editor) by how far out they are: "In every loopRoutine" is the loop the block is directly in, then "In the loopRoutine around it", "In the loopRoutine 2 levels out", … Your outer-collection/inner-range example now reads `In every loopRoutine → value` and `In the loopRoutine around it → widgetID` / `DEFAULT`. Covered by a `FOREACH` over a collection with a `FOREACH` over a list inside it in the test.

**3. `caller` in a routine no CALL reaches** — added `caller: 1` to the `validCollections` of that context at `validate_gamefile.js:1328`, with a comment saying the "not being called" problem is reported right below. Checked with `validate_gamefile_node.js` on a game file whose only custom routine uses `caller` and is never called: the undefined-collection problem is gone, the uncalled one stays.

**4. The custom-routine group understates what CALL hands over** — did both halves. The card description now says the routine starts with everything the calling routine had, all of its values and collections, plus the arguments that operation hands over and the widget it ran on as `caller`. And the argument names are collected: `callArgumentsFor` walks every `Routine` array of every widget in the room (into `thenRoutine`/`elseRoutine`/`loopRoutine`) for a `CALL` naming this routine and offers its `arguments` keys. They are a group of their own, "From the operation that runs it", rather than part of "In every fooRoutine" — an argument only some of the calls pass is not something the routine has *every* time, and the tooltip says which CALL passes it ("handed over by the CALL in clickRoutine of holder", or "handed over by 3 of the operations that run this routine").

`npm test` (1265 tests) and the extended TestCafe test both pass here.

VTT-AI-Agent (commented):
## TL;DR

**UI/UX review only** (layout, visual design, readability — a separate reviewer covers code correctness). The feature is a real win: every case in the PR's table shows up in the right group, the popups keep their existing shape, and I found no layout regression anywhere in the editor at any of the four viewport sizes I tested. The one change I'd push for is that **the descriptions this PR adds are the whole point of it, and they live only in `title=` tooltips** — invisible on touch, invisible to the keyboard, and a one-second hover away on desktop. A single description line at the foot of the open section would make the new information actually reachable. Second: with 3–4 groups now stacked, the 11px / 0.7-opacity group line is too quiet to carry the distinction the PR is built on.

---

### What I exercised

Checked out the branch and ran it locally against a room with one widget per row of the PR's table (`enterRoutine`, `leaveRoutine`, `gameStartRoutine`, `globalUpdateRoutine`, `textChangeRoutine`, a custom `dealCardsRoutine` reached by three separate CALLs, a FOREACH and a FOREACH-in-a-FOREACH), then clicked every value and collection popup open.

Viewports: **390×844**, **820×1180**, **1280×800**, **1920×1080**.

Overlays opened at all four sizes: game shelf / states, players, about-help, edit-mode sidebar, Edit Widgets module, add-widget overlay, and all eight editor sidebar modules (History, JSON, Widgets, Debug, Assets, Toolbox, Game Settings) — plus the PR's own surface, the routine parameter popups (value / collection / widget / property sections) and the routine-card info tips. **No horizontal overflow anywhere in `#editorModules` at any size**, and nothing became unreachable or uncloseable. That is unsurprising: the diff touches no CSS at all, so a shared-style regression was never on the table.

Not reached: chat, share/publish, the deck editor's import/export dialogs, and the seat/player detail sub-dialogs.

| | |
| --- | --- |
| enterRoutine values | enterRoutine collections |
| custom routine values | nested FOREACH |

---

## Big layout improvements

### 1. The descriptions are the payload, and they only exist as `title=` tooltips

`popup.js:969` — `entryButton.title = entry.description || group.title;` is the only place the new text is ever shown. That means:

- **On a tablet or a phone it is never shown at all.** Edit mode is fully usable at 820×1180 (screenshot: [820 tablet](https://agent.virtualtabletop.io/reports/ui-review/1539004184060235856/v820-customRoutine-values.png)) and reachable at 390×844, and on both a native `title` never fires. A first-time user on an iPad gets a chip called `child` and no way to learn what it is — which is exactly the situation the PR set out to fix.
- **On the keyboard it is never shown.** These entries are `button`s, so they take focus, but `title` only responds to the pointer.
- **On desktop it costs a ~1s hover per chip**, and the value section now holds 10–13 of them. Reading the four `dealCardsRoutine` arguments means four separate hovers.

The `oldParentID` tip is the sharpest case: *"id of the widget the entering widget came from (null when a player dragged it - picking it up already took it out of its parent)"* — over 120 characters, with the genuinely important half in the parenthetical, rendered as an OS tooltip strip.

**Proposal, in the app's own idiom:** add one description line at the bottom of the open accordion section that fills from the hovered *or focused* entry, styled like the existing `.popup-list-settings` (12px, `opacity: .8`, `client/css/editor/controls/popup.css:517`). One line, so it never becomes the "wall of text nobody reads through" the comment at `popup.js:956` rightly warns about, and it reserves a fixed 1–2 lines so the popup doesn't jump. This is the change that turns the PR from "the information is in the DOM" into "the information is on screen".

### 2. The group line is too quiet to carry the distinction the PR is built on

`.popup-value-group` is 11px at `opacity: 0.7` with no rule, no indent and no weight (`popup.css:500`). That was fine when there were two groups. A nested FOREACH now shows **four** (`From earlier operations` → `In the loopRoutine around it` → `In every loopRoutine` → `In every routine`) and a custom routine three, all separating runs of chips that are visually identical — so the section reads as one undifferentiated chip field with faint captions in it. Which group a name is in *is the information this PR adds*; right now it is the least prominent thing in the section.

**Proposal:** keep the size and colour restraint, just give the line a boundary — `border-top: 1px solid var(--backgroundHighlightColor1)` on `.popup-value-group` (skipped for the first one), `opacity: .85`, and optionally indent the chips 6–8px under it so the caption visually owns its run. No new colours, and it matches the hairline the accordion cards already use.

---

### 3. A long CALL argument name overflows the popup horizontally at phone width

`client/css/editor/controls/popup.css:493` — `.inline-popup .popup-entry { display: inline-block; }` with no `max-width`. The popup is `min-width: 220px; max-width: 50vw` (`popup.css:48`), so at 390px wide the `min-width` wins and the popup is 220px. I put an argument named `extraLongArgumentNameForLayout` on one of the CALLs and the chip runs straight through the popup's right border: `scrollWidth: 243` against `clientWidth: 218`, i.e. a 25px horizontal scrollbar inside the popup.

long argument name overflowing at 390px

This is newly reachable *because of* this PR: the names in `From the operation that runs it` are collected from **other widgets'** CALLs (`events.js:157-171`), so the author editing this routine does not control them and cannot shorten them. Fix:

```css
.inline-popup .popup-entry { display: inline-block; max-width: 100%; }
.inline-popup .popup-entry button { max-width: 100%; overflow-wrap: anywhere; text-align: left; }
```

### 4. Nested FOREACH lists the outer loop above the inner one

`routine.js:3406` — the loop groups come out outermost-first, so the order down the section is `From earlier operations` (nearest) → `In the loopRoutine around it` (two levels away) → `In every loopRoutine` (the loop you are actually in) → `In every routine` (broadest). Specificity goes near, far, near, far. Every other list in this popup narrows from near to broad.

nested FOREACH group order

Suggest reversing the loop groups so the innermost is first, and — since the sibling title says "the loopRoutine around it" — titling the innermost **"In this loopRoutine"** rather than "In every loopRoutine" (`routine.js:3364-3367`). "In every X" next to "In the X around it" mixes two different senses of the article; "this" / "around it" / "2 levels out" reads as one scale.

### 5. "From the operation that runs it" is singular, but the list is a union over every CALL in the room

`events.js:180`. In my room three widgets CALL `dealCardsRoutine` with four different argument names between them, and all four are offered in one group under a singular heading — so it reads as "these are the arguments *the* call passes", when in fact no single call passes all four.

Suggest **"From the operations that run it"**, and marking arguments that only some calls pass — dimmed, or an appended `?`, or a trailing "(some calls)". Right now the only signal is the tooltip, which per finding #1 many users never see.

### 6. The multi-caller tooltip names none of the callers

`events.js:178` — with more than one caller the tip is *"handed over by 3 of the operations that run this routine"*, which tells the user a count and nothing actionable. The single-caller branch one line up is much better (*"handed over by the CALL in clickRoutine of holder"*). Suggest naming up to two or three and then eliding: *"handed over by the CALLs in clickRoutine of holder, clickRoutine of other and 1 more"*.

### 7. A shadowed preset loses its description entirely

In a `textChangeRoutine` whose first operation is a `GET` into `value`, the second operation's popup moves `value` into `From earlier operations` — correct — but its tooltip becomes `entry.description || group.title` → the literal string **"From earlier operations"**, which is the caption printed on the line directly above it. So the user gets *less* explanation of the same name further down the routine, and the one tooltip they do get is a restatement of the heading.

| operation 1 | operation 2 |
| --- | --- |
| before the GET | after the GET |

Two small fixes: don't fall back to `group.title` when it is the caption immediately above (leave the tooltip off), and for a name that shadows a preset, say so — *"stored by an earlier operation; it replaced the `value` the textChangeRoutine handed over"*. That sentence is the single most useful thing the popup could tell someone debugging this, and the data to build it is already in `presetSections`.

---

### 8. The reworded card descriptions read literal names as plain English

The new wording is accurate but every literal name is rendered as ordinary prose, because the description goes through `escapeHTML()` into a `<pre>` at `events.js:557`:

> "Runs when any property of any widget in the room changes. **The routine is handed the changed widget as the collection widget** and can use value - what the property is now - as well as oldValue, property and widgetID."

globalUpdateRoutine description

"the changed widget as the collection widget" is the hard case — a newcomer cannot tell that the second `widget` is a name they have to type, not a noun. Same for "as the collection child" on `enterRoutine`/`leaveRoutine` and "the widget it ran on as the collection caller" on a custom routine. Elsewhere the editor teaches these by colour (`--routineCollectionColor`, `--routineVariableColor`) and the info popup already knows how to colour names (`structureInfoHTML`). Even just quoting them — *"as the collection `child`"* — would separate name from prose. Worth doing precisely because these four sentences are the PR's correction of a description that used to send people down the wrong path.

### 9. The section info tip now points at something that isn't there

`popup.js:915` and `:937`: *"The ones below **the other lines** are there without any operation creating them…"*. The previous wording named a group the reader could see ("the ones below \"In every routine\""); "the other lines" names nothing. A newcomer opening this for the first time has no idea which lines are meant, and the sentence has to be re-read against the list to be decoded.

Suggest anchoring it to the first group, which is the one constant across every case: *"Everything below \"From earlier operations\" is there without any operation creating it: some of it in every routine, the rest because of what started this one — a changeRoutine is told what changed, and a routine another one runs gets the arguments it was called with."*

### 10. `gameStartRoutine` offers the same widget twice with nothing marking the alias

The `gameStartRoutine` popups list `widgetID` ("the same as thisID") and `widget` ("the same as thisButton") in their own group, *and* `thisID`/`thisButton` under "In every routine" — four chips, two things.

gameStartRoutine collections

The tooltips do say "the same as", which is the right call, but per finding #1 that is hover-only, so on screen it looks like four distinct choices and a newcomer will wonder which is correct. Suggest wording that makes the equivalence an instruction rather than a fact — *"another name for thisID; either works"* — and, if you want to go further, considering whether `gameStartRoutine` needs its own group at all when both entries are aliases of something already listed one group down.

---

### Side note: the demo room in `pr-test-ai`

There is no draft tutorial in the room — the shelf holds the PR's **Routine presets demo** game, which I played through (dropped the token, clicked "run a custom routine" and watched `shoutRoutine was called by caller` appear). Since the PR offers it for `library/tutorials/` on request, three notes for that eventuality:

demo room

- **The stale state reads as broken on arrival.** A freshly opened room already says "left the holder" and "token was rotated to null" under two of the sections, before the visitor has touched anything — so the first impression is of a demo that has already gone wrong. The `caller` section gets this right ("nobody has called shoutRoutine yet"); the other two want the same neutral empty state, e.g. "drop the token in to see this fill in".
- **The four blocks don't share a grid.** Their captions start at x≈113, 538, 485 and 343 and the blocks are different widths, so nothing lines up vertically and the eye has no column to follow. Two even columns at a common left edge would make it read as one table, which is what it is.
- **All the content sits in the top-left third**, leaving the bottom half and the right edge empty at 1280×800 and worse at 1920×1080. Widening the two columns to use the board would close the gap between a caption and the widget it labels, which is currently large enough that "left the holder" doesn't obviously belong to the holder above it.

The demo does not currently cover `changeRoutine`, `gameStartRoutine` or a `loopRoutine`, so three rows of the PR's table have no widget — worth adding before it goes anywhere near the tutorial library.

VTT-AI-Agent:
Thanks — this was the useful kind of review. All ten addressed in 994a856; the popup got its first CSS change of the PR.

**1. The descriptions only existed as `title=` tooltips** — fixed the way you proposed. Every value/collection section now ends in a description line that fills from the entry the pointer *or the keyboard* is on (`mouseenter`/`focus`, cleared on `mouseleave`/`blur`), styled like `.popup-list-settings` (12px, `opacity: .8`) with two lines kept free so the list does not move while the pointer travels along it. The tooltips stay. Every entry has a description now, including the ones under "From earlier operations", which had none.

One gap I could not close in the same move, so I am flagging it rather than claiming it: on a touch screen the tap that would reveal the line is also the tap that applies the value and closes the popup, so the line helps the keyboard and the desktop but a phone still gets nothing. Making the first tap explain and the second apply is a real interaction change for every chip in the editor — happy to do it if you want it, but I did not want to invent it here.

**2. The group line was too quiet** — `.popup-value-group` is now `opacity: .85` with a `1px solid var(--backgroundHighlightColor1)` rule above it (`.popup-entry + .popup-value-group`, so the first group has none) and 6px of space under the rule. Same size, same restraint, no new colours. I left the chips unindented — with the rule the caption reads as owning the run below it — but that is a one-liner if you want it.

**3. A long CALL argument name overflowed at 390px** — fixed with exactly the rules you wrote: `max-width: 100%` on `.popup-entry` and `max-width: 100%; overflow-wrap: anywhere; text-align: left` on its button. With `extraLongArgumentNameForLayout` on one of the CALLs, the popup at 390×844 is now `scrollWidth 218 / clientWidth 218` — [screenshot](https://agent.virtualtabletop.io/reports/pr/1538977191579099247/uiux-390-customRoutine-values.png).

**4. Nested FOREACH listed the outer loop above the inner one** — the preset groups are now ordered nearest-first throughout (`presetSections` drops a name that an *earlier* group has, instead of a later one), so the section reads `From earlier operations` → `In this loopRoutine` → `In the loopRoutine around it` → `In every fooRoutine` → `In every routine`. And the innermost is titled **"In this loopRoutine"**, as you suggested, so it is one scale with "around it" / "2 levels out". [Screenshot](https://agent.virtualtabletop.io/reports/pr/1538977191579099247/uiux-1280-nested-foreach.png), and the order is asserted in the TestCafe test.

**5. "From the operation that runs it" was singular over a union** — the title is now singular or plural by how many CALLs in the room run the routine, and an argument that only some of them hand over is dimmed (`opacity: .65`, dashed border) with the description saying so: *"handed over by the CALL in enterRoutine of holder - not by the other operation that runs this routine"*. Now that the description is on screen rather than hover-only, that is the signal you asked for.

**6. The multi-caller tooltip named none of the callers** — it names up to two and then counts: *"handed over by the CALLs in clickRoutine of holder, changeRoutine of deck and 2 more"*. The count-only wording is gone.

**7. A shadowed preset lost its description** — the `entry.description || group.title` fallback is gone (no tooltip is better than echoing the caption above), and a name that shadows a preset now says which: *"stored by an earlier operation - it replaced the value the textChangeRoutine hands over"*. Everything else under "From earlier operations" gets "stored by an earlier operation of this routine" / "picked out by an earlier operation of this routine". Covered by a test assertion.

**8. Literal names read as prose** — quoted throughout the automation descriptions: *"is handed the changed widget as the collection `"widget"`"*, `"child"`, `"caller"`, `"value"`, `"oldValue"`, `"property"`, `"widgetID"`. I did the quoting rather than the colouring: those descriptions go through `escapeHTML()` into a `<pre>` (`events.js:557`), so colouring means giving that path markup, which is a bigger change than this PR should carry — worth doing for all the card descriptions at once.

**9. "the other lines" named nothing** — reworded to your suggestion, anchored to the first group: *"Everything below \"From earlier operations\" is there without any operation creating it: some of it in every routine, the rest because of what started this one - a changeRoutine is told what changed, and a routine another one runs gets the arguments it was called with."* Same shape for the collections tip.

**10. `gameStartRoutine` offers the same widget twice** — reworded as an instruction: *"another name for thisID - the id of the widget this routine is on, either works"* and the same for `widget`/`thisButton`. I kept the group: the engine really does hand those two names over, and a game that uses `${widgetID}` in a `gameStartRoutine` is correct — dropping them from the list would hide that, and the validator knows them now.

### The demo room

Rebuilt and re-uploaded, all three of your points:

- **No stale state on arrival.** The old `.vtt` had been downloaded *after* I played it, which is why it greeted you with "left the holder". The published file is now a freshly built state and every log reads as an instruction: "drop the token in to see this fill in", "move the token to see this fill in", "type in the label above to see this fill in".
- **Two aligned columns.** Every block is 720px wide at a common left edge (x=60 and x=820), captions and logs all `text-align: left`, so the captions line up and each log sits directly under the widget it belongs to.
- **The three uncovered rows are covered.** `changeRoutine` (the token logs which of its own properties changed — it filters out `dragging`/`z` with a nested IF so the log reads as something a player did), `gameStartRoutine` (a label that rewrites itself through the `widget` collection when the game is loaded fresh) and a `loopRoutine` (a button that SELECTs three chips and FOREACHes over them, each round setting `DEFAULT`'s text to `I am ${widgetID}`). The custom-routine block now has *two* buttons calling the same routine with different arguments, so the "not every call passes this" case from finding #5 is visible in the demo too.

Walked it end to end again after rebuilding (drop, drag, type, both buttons, the loop) and verified the `.vtt` uploads into a room with its preview image and help text.

`npm test` (1265) passes; the extended `tests/testcafe/editor.js` preset test passes here, and I updated the jest test that asserted the description was *only* a hover tip to assert the new line instead.

VTT-AI-Agent (commented):
## TL;DR

The feature is close, but the editor still reports the wrong origin for a value overwritten inside a `FOREACH` block, and the new CALL-argument inventory has correctness gaps (including an edit-mode crash for a valid argument name). These do not affect engine execution or game state, but they undermine the guidance this PR adds and deserve another pass.

## Findings

### 1. An earlier operation *inside* a `loopRoutine` does not get to shadow the loop preset — `client/js/editor/controls/popup.js:584`

`presetNames()` removes every name supplied by a non-`shadowedByOperations` group from "From earlier operations." That is correct for values inherited from outside the `FOREACH`, because the loop establishes its `value`/`widgetID`/`DEFAULT` after those outer operations. It is not correct for operations already executed inside the block. `RoutineEditor.setRoutine()` seeds the nested editor from the outer scope and then appends locally defined names to the same arrays (`routine.js:2061-2081`), so the popup cannot currently tell those cases apart.

For example:

```json
{
  "func": "FOREACH",
  "in": [1, 2],
  "loopRoutine": [
    { "func": "GET", "variable": "value", "collection": "cards", "property": "rank" },
    { "func": "SET", "collection": "cards", "property": "text", "value": "${value}" }
  ]
}
```

At the second operation the popup labels `value` as "In this loopRoutine" and describes it as the list entry, while the engine initialized the loop value before entering the block (`widget.js:1716-1729`) and the preceding GET has since replaced it. Please preserve scope provenance (outer/inherited versus operations earlier in this block) so a loop preset shadows only the former, and add this in-block-shadowing case to the editor test.

### 2. Multiple CALLs in the same source routine are collapsed into one — `client/js/editor/controls/events.js:127-138`

Both `sources` and each argument's source set are keyed by the display string such as `clickRoutine of holder`, rather than by the CALL operation. If that one routine contains two CALLs of `dealCardsRoutine`, one with `{ "faceUp": true }` and one without it, both calls collapse to one source: `sources.size` and `args.faceUp.size` are both 1. The UI therefore presents `faceUp` as always supplied, uses the singular "operation that runs it" heading, and omits the partial styling even though one invocation receives `undefined`. Track a unique identity/path for each CALL while retaining the readable source label for the description, and cover two calls from the same routine in the test.

### 3. CALLs inherited from a deck's `cardDefaults` are absent from the inventory — `client/js/editor/controls/events.js:144-147`

The scan only examines top-level entries of `widget.state`. A deck's `cardDefaults.clickRoutine` is a real routine exposed by this same Automations UI (`events.js:484-495`), and cards execute it through their inherited defaults, but a CALL inside it is nested under `cardDefaults` and is never scanned. Consequently its arguments are not offered in the target custom routine, despite the PR promising the arguments of every CALL in the room. Please include routine arrays under `cardDefaults` (and any other supported inherited routine container) and add a deck-card-defaults case.

### 4. A valid argument named `constructor` crashes the Automations editor — `client/js/editor/controls/events.js:126-138`

`args` is a plain object, so before any assignment `args.constructor` resolves to the inherited `Object` constructor. A CALL with `"arguments": { "constructor": "value" }` reaches `(args[name] = args[name] || new Set()).add(source)` and throws `TypeError: args[name].add is not a function` while expanding the called custom routine. `constructor` matches the engine's variable identifier syntax and is accepted by the validator, so this is valid routine JSON. Use a `Map` or a null-prototype dictionary for user-controlled argument names (and likewise avoid reintroducing prototype-key problems when building `variables`), with a regression test.

No new widget property or routine operation is introduced, so the visual-UI support rule is satisfied by the feature itself; the PR also keeps its demo out of `library/`, and the latest CI checks are green.

The `beta` label was already present; I left it in place because this edit-mode-only PR remains safe and useful for beta testing and does not alter engine behavior.

VTT-AI-Agent:
Thanks — all four are real and all four are fixed in 4d39564.

**1. An earlier operation inside a `loopRoutine` did not get to shadow the loop preset** — you were right that the popup had no way to tell the two cases apart, so I stopped asking it to. There is one rule now: **every** preset group loses a name that an operation of the routine has stored (`presetSections` filters every group by `existing`, and the `shadowedByOperations` flag is gone). What makes a FOREACH win over the operations *before* the block is no longer a flag but the scope itself: `RoutineOperationEditor.subroutineScope` — overridden by the FOREACH editor — takes the names the round hands over out of the variables/collections the nested editor inherits, exactly as the engine establishes them when the block is entered. An operation inside the block puts the name back into that list and wins again, which is the case you gave.

Your example now reads, at the SET: `From earlier operations → value`, described as *"stored by an earlier operation - it replaced the value this loopRoutine hands over"*, with `key` still under "In this loopRoutine". Both directions are asserted in the editor test — a GET *before* the FOREACH is still shadowed by the loop at the first operation inside the block, a GET *inside* it is not.

**2. Multiple CALLs in the same source routine were collapsed into one** — `callArgumentsFor` now keeps one entry per CALL operation (`{ source }`, pushed into a list) and keys the argument sets by that entry, so two CALLs in one routine are two operations. The readable label is still what the description lists, deduplicated: two CALLs in `clickRoutine of holder` read *"handed over by the CALLs in clickRoutine of holder"*, and an argument only one of them passes is dimmed and says *"— not by the other operation that runs this routine"*. The test has a routine whose only two callers are two CALLs in one `clickRoutine`, and asserts both.

**3. CALLs inherited from a deck's `cardDefaults` were absent** — the scan now walks a deck's `cardDefaults` routines as well, labelled *"clickRoutine of the cards of deck"* so the description says which of the two places on the deck it is. The test's third caller is a CALL in `cardDefaults.clickRoutine` of a deck.

**4. An argument named `constructor` crashed the Automations editor** — the argument names go into a `Map` now, the group's `variables` is an `Object.create(null)`, and every lookup of a name in a preset group goes through `presetEntryOf`, which asks with `hasOwnProperty` (that one mattered too: a plain group would have claimed to hold `constructor` and swallowed the entry). The test hands over an argument named `constructor` and asserts it is offered with the right description.

`npm test` (1265) passes, and the whole `tests/testcafe/editor.js` suite (46 tests) passes here. I also checked that the new assertions fail on the previous commit — the loop-shadowing one and the CALL ones each fail for their own reason.

VTT-AI-Agent (commented):
## TL;DR

The shadowing model is now right in both directions — I re-derived every row of the table from `widget.js`/`statemanaged.js`/`serverstate.js`, walked the nested-FOREACH and in-block-overwrite cases through `presetSections`, and they all come out correct — and the new tests actually pass here (315 jest tests in `routineeditor.test.js`, and the new `editor.js` case green in 47s). Nothing left that I'd call a bug; the one thing I'd still fix is that the CALL-argument inventory matches purely on routine *name* and ignores the CALL's `widget`, so two widgets with a same-named custom routine each get shown the other's arguments and a wrong "not by the other N operations" count.

## Findings

### 1. The argument inventory ignores which widget a CALL runs the routine on — `client/js/editor/controls/events.js:125-165`

`callArgumentsFor(routineName)` walks every routine in the room and collects every `CALL` whose `routine` equals the name, with no look at `operation.widget`:

```js
const called = Array.isArray(operation.routine) ? operation.routine[0] : operation.routine;
if(operation.func == 'CALL' && called === routineName) { ... }
```

The engine resolves the target explicitly — `setDefaults(a, { widget: this.get('id'), routine: 'clickRoutine', ... })` and then `widgets.get(a.widget).evaluateRoutine(...)` (`widget.js:1466-1489`). So a room with `dealCardsRoutine` on two different decks, each CALLed with its own arguments, produces this in *both* editors:

- the union of both sets of argument names is offered, including names that routine never receives;
- `calls.length` counts CALLs that can never reach it, so `partial` is set for arguments that in fact arrive on every call that does reach it — the entry is dimmed and dashed and reads "not by the other 2 operations that run this routine" when it is really always there.

Generic routine names (`initRoutine`, `refreshRoutine`, `dealRoutine`) are common enough that this is not a corner case. The cheap and fully safe half of the fix is the CALL with no `widget` at all: it targets the widget it is written on, so if that widget is not the one whose routine is open, it provably does not run this routine. That needs `routinePresetsOf` to receive the widget id (`events.js:652` has it). A literal `widget: 'someID'` can be compared too, with the caveat that a routine in a deck's `cardDefaults` is reached through *card* ids rather than the deck's; `${...}` and arrays would stay included as unknown.

Worth noting this mirrors what `validate_gamefile.js:605-607` already does (it validates the called routine on every widget that has one by that name), so it may be a deliberate match — but the validator only decides whether a name is defined, while the editor puts the wrong name in front of the author as something to pick.

### 2. Inside a FOREACH block the loop's own names drop out of the VAR key suggestions — `client/js/editor/controls/routine.js:3443-3446`, `:2545`

`ForeachRoutineOperationEditor.subroutineScope` removes `value`/`key`/`widgetID`/`DEFAULT` from the arrays it hands the nested editor, which is exactly right for the popup lists. But those same arrays feed `parameterKeySuggestions`:

```js
parameterKeySuggestions(name) {
  ...
  return (this.variables || []).filter(variable=>typeof variable == 'string' && taken.indexOf(variable) == -1);
}
```

which is what a `keyValue` popup offers as names to write to (`routine.js:2530`, i.e. VAR). So inside a `FOREACH` over a list, a VAR that means to overwrite `value` for the rest of the round no longer gets `value` suggested — and that is a case this PR explicitly models as legitimate ("An operation *inside* the block runs later still and wins again"). It is also a small regression: before this PR, `value` was at least suggested when an operation before the FOREACH had stored it.

Fix is one line's worth: fold the preset names into the suggestion list, e.g. `[...this.variables, ...(this.presets||[]).flatMap(g=>Object.keys(g.variables||{}))]` deduped, in `parameterKeySuggestions`.

### 3. The description line can sit below the fold exactly while you are pointing at a name — `client/js/editor/controls/popup.js:984-1011`, `client/css/editor/controls/popup.css:532-538`

`content.append(descriptionDOM)` puts the line after the last group, and `.inline-popup` is `max-height: calc(100vh - 20px); overflow: auto` (`popup.css:62-63`). A custom routine inside a nested FOREACH now stacks four or five groups plus the nine "In every routine" names, so on a short window the foot of the section is scrolled out while the pointer is on a name near the top — and for the keyboard it is worse, because focusing a button scrolls *it* into view, which pushes the description further away. That defeats the two audiences the line was added for.

`position: sticky; bottom: 0` on `.popup-entry-description`, with `background: var(--backgroundColor)` so the chips do not show through, keeps it in view at no cost to the "two lines kept free" reasoning.

### 4. A CALLed `clickRoutine` gets `caller` and arguments too, and is offered neither — `client/js/editor/controls/events.js:185-215`

`routinePresetsOf` only builds the CALL group when the property is not a predefined event and matches neither the `ChangeRoutine` nor the `GlobalUpdateRoutine` pattern. But `CALL` runs *any* routine by name — `a.routine` defaults to `'clickRoutine'` (`widget.js:1467`), and `inheritCollections['caller']` plus the merged `arguments` are set regardless of which routine it is. Running another widget's `clickRoutine` from a routine is a normal VTT pattern, and there the argument names and `caller` are just as real as in a custom routine, and just as invisible.

Since `callArgumentsFor` already scans the whole room, offering the group whenever some CALL in the room names the routine (rather than gating it on "not a predefined event") would cover this without a new mechanism — `caller` would then also only appear where a CALL actually exists, which is more accurate than today's unconditional `caller` on every custom routine. At minimum the `clickRoutine` card description could mention it, the way the custom-routine one now does.

### Smaller notes

- The new testcafe case covers the group titles, the shadowing and the description text thoroughly, but nothing asserts the `popup-entry-partial` class itself (only the wording that goes with it), and `describeCallers`' third branch — more than two source routines, `"... and N more"` — has no case at all.
- `entryButton.title` is now only set when there is a description (`popup.js:999-1000`), dropping the old `|| group.title` fallback. Every entry does carry one today, so this is fine, but any future group whose entries have no description will silently lose its tooltip.

### What I checked and found correct

- Every preset against its `evaluateRoutine` call site: `statemanaged.js:130-141` (change/`<property>`Change/global/`<property>`GlobalUpdate, including that the property-specific variants get no `property`), `widget.js:3149`/`:3155` (`child`, `oldParentID` and its null-on-drag), `holder.js:109`, `serverstate.js:162`/`:560`, and the three FOREACH variants at `widget.js:1745-1791`.
- The shadowing direction in both cases, by hand: a trigger preset loses to a later operation (`existing` filter in `presetSections`), a loop preset wins over everything outside the block (`subroutineScope`) and loses again to an operation inside it, and after the block the outer `value` is correctly back — which matches the engine's per-round backup/restore in `callWithAdditionalValues`. Nested list-outside/range-inside, collection-inside/list-outside and "operation in the outer block stored `key`" all come out right.
- `presetEntryOf`/`Object.create(null)` really do handle an argument named `constructor`, and every label goes through `textContent`, so a game-authored argument name cannot inject markup.
- `scan` recurses into exactly the three block properties the engine evaluates (`thenRoutine`, `elseRoutine`, `loopRoutine` — `widget.js:1729`, `:1863`), so no nested CALL is missed.
- Ran `node --experimental-vm-modules node_modules/.bin/jest tests/client/routineeditor.test.js` (315 passed) and `npx testcafe "chromium:headless --no-sandbox" tests/testcafe/editor.js -t "A routine offers the values and widgets that what started it hands over"` (passed, 47s) on 4d39564c.
- No `library/` or tutorial change in the diff, and the demo room is on the PR's `pr-test-ai` server rather than committed — which is what the tutorial rule asks for. No new widget property and no new routine function, so there is nothing owed to `jeCommands`; the feature is itself part of the visual routine editor UI.

The `beta` label was already on this PR and I left it there — the diff touches no engine file at all (only `client/js/editor/`, `client/css/editor/`, `validator/` and tests), so nothing about how existing games run can change, and the change is safe and useful for players to try.

VTT-AI-Agent:
Thanks — all four are real, and all four are fixed in 1f5eb6b7. The smaller notes are covered too.

**1. The argument inventory ignored which widget a CALL runs the routine on** — it compares the widget now, both halves of what you described. `callArgumentsFor` is given the routine that is open as `{ widgetID, cardDefaults }` (from `events.js:695`, which has both), and every CALL is asked whether it can reach it:

- **no `widget` of its own** → it runs the routine on the widget it is written on, so it reaches this one only if that is the same widget; in a deck's `cardDefaults` the widget writing it is the *card*, so such a CALL reaches the deck's `cardDefaults` routine and not the deck's own.
- **an id written out** → looked up (`routineHomeOf`): the widget's own routine if it has one, otherwise — for a card — the `cardDefaults` of its deck, which is the caveat you named.
- **`${...}`, an array of them, anything not a string** → can be any widget, so it stays in. An id that is in no widget of the room stays in too.

So a room with `dealCardsRoutine` on two widgets now shows each editor its own arguments, and `partial` counts only the CALLs that can reach it. The test has a second `label2.dealCardsRoutine`, reached by a CALL with no `widget` at all in `label2`'s own `gameStartRoutine`, and asserts neither routine is shown the other's argument.

**2. Inside a FOREACH the loop's own names dropped out of the VAR key suggestions** — fixed as you wrote it: `parameterKeySuggestions` folds the preset names into the list, deduplicated. Inside a `FOREACH` over a list a VAR is offered `value` and `key` again (asserted in the test), and it is now also offered what the *trigger* hands over — a `changeRoutine`'s `value` was never suggested either, for the same reason.

**3. The description line could sit below the fold while you were pointing at a name** — `position: sticky; bottom: 0` with the popup's own background, as suggested. One change to your recipe: the line's `opacity: .8` had to go, because an 80% element lets the chips scrolling under it show through its background — the muting is in the text colour now (`color-mix(in srgb, var(--textColor) 80%, transparent)`, falling back to the plain text colour where `color-mix` is unsupported). [Screenshot](https://agent.virtualtabletop.io/reports/pr/1538977191579099247/review4-sticky-description.png) at 1280×320, custom routine two FOREACHes deep: the line sits over the list at the foot of the popup.

The test asserts the case rather than the rule: with the section taller than the popup and the popup scrolled to the top of it, the description is still inside the popup. It fails on the previous commit (`descriptionInPopup: false`).

**4. A CALLed `clickRoutine` got `caller` and arguments and was offered neither** — it is offered both now, exactly where a CALL in the room runs it, in a group of its own titled **"Only when a CALL runs it"** with every entry dimmed: a player clicking the widget hands over none of them, and that is the whole difference to a custom routine, where they are what it always has. [Screenshot](https://agent.virtualtabletop.io/reports/pr/1538977191579099247/review4-call-group.png). I did not make `caller` on a custom routine conditional on a CALL existing — a routine is usually written before the CALL that runs it, and losing `caller` until then would be the wrong way round.

**Smaller notes**

- `popup-entry-partial` is asserted as a class now, on the argument only one CALL hands over and on the `from` of the CALLed `clickRoutine`, and negatively on one every call passes.
- `describeCallers`' third branch has a case: `amount` is handed over from three source routines and reads *"handed over by the CALLs in enterRoutine of holder, clickRoutine of the cards of deck and 1 more"*.
- The `title` fallback: left as it is on purpose — echoing the group caption into the tooltip was what finding 7 of the previous review asked me to remove, and every entry carries a description. If a group without one ever appears, it is the missing description that wants fixing.

`npm test` (1265) passes and the whole `tests/testcafe/editor.js` suite (46 tests) passes here in 13m48s. Each new assertion was checked against the previous commit: the widget-filtering one, the VAR-suggestion one and the sticky one each fail there for their own reason.

The demo room on the test server is unchanged — it still shows the custom-routine case correctly; the newly offered `clickRoutine`-run-by-CALL case is not in it.

PR #3125, opened 2026-08-17
Give the deck editor's standard deck the same sorting properties as the add widget overlay's

VTT-AI-Agent:
## What

The **Standard playing cards** deck that the deck editor offers (*Add New Deck → Use an existing deck → Traditional deck*, defined in `assets/decks/standard.json`) only gave its cards `image`, `rank` and `suit`. The standard deck added through the **add widget overlay** (`generateCardDeckWidgets` in `client/js/editmode.js`) gives them six properties that exist purely so a game can sort a hand or a pile. This adds the missing ones, with the same values, so both standard decks sort identically.

| property | value | what sorting by it does |
|---|---|---|
| `suit` | `C` `D` `H` `S`, `T` for the jokers | groups the four suits, jokers last |
| `suitColor` | `♣` `♦` `♥` `♠` `🃏` | keeps the two black and the two red suits together |
| `suitAlt` | `1♣` `2♥` `3♠` `4♦` `5J` | alternates black and red |
| `rank` | `01`…`10`, `2J`, `3Q`, `4K`, `J1`/`J2` | ace low |
| `rankA` | same, but `5A` for the aces | ace high |
| `rankFixed` | `rank` + suit letter, e.g. `4K D` | one fixed order for the whole deck |

The readable card type names (`2 of spades`, `Joker 1`) and the card images are unchanged, so nothing that already uses this deck breaks — it is a template file for newly created decks, existing games carry their own copy of `cardTypes`.

Two things came out of the review of the data change and are in here as well:

- **The deck editor no longer offers a color picker for a color-named property that holds no color.** It put the swatch + picker on every property matching `/color/i`, so the new `suitColor: "♠"` row got one, and a single click in it would have overwritten the sort key with a hex code. Whether the picker is offered now depends on the current value alone: the value is a color, the field is empty (the picker is how a color gets in there), or a hex / `rgb()` / `hsl()` value is half-typed — shapes that can't be anything but a color on its way in, so the button doesn't flicker between two keystrokes. A word typed over a value (`red` → `S` → `Spades`) therefore loses the picker instead of keeping it, and real color rows are untouched: `color: transparent` on a face object still has its picker.
- **The card type panel says what the sorting properties are for.** A card type carrying **all six** of these properties now gives each of those rows a tooltip (`suitColor` — *"Sorting property: keeps the suits of the same color together. Despite its name it holds no CSS color here."*), so the codes are explained where they are met rather than only in the deck picker. Card type properties are author-defined, so anything short of the complete set stays silent: a deck that uses `rank`/`suit` for its own readable values (German, Spanish), or a hand-built card type with a `rank` and a `rankFixed` of its own, is not told that its `rank` sorts ace-low.
- **The deck's `description`** — shown in the Traditional deck dialog — mentions the sorting in one clause ("The cards carry extra properties so a hand or a pile can be sorted by suit or by rank."), the same shape as every other deck's entry.

## Why

Sorting was not just incomplete, it was wrong. With the old values `SORT` by `rank` produced `10, 2, 3, … 9, A, J, K, Q` (`localeCompare` on the plain strings), the jokers had no `suit` at all and therefore sorted *before* every real card, and `suitColor`, `suitAlt`, `rankA` and `rankFixed` were simply missing — so none of the sorting recipes the library card games rely on worked on a deck added this way. `library/games/Hearts/0.json` and `library/games/Spades/0.json`, for instance, carry exactly these six values in their own decks; a game built on the deck editor's standard deck could not sort the same way.

## Testing

- `npm test` (jest): 1280 tests pass, including two new suites:
  - `tests/standard-deck.test.js` derives every card type from its name and compares it against the file, then asserts the six documented sort orders — the table here is a hand-maintained copy of the one `generateCardDeckWidgets` builds, and nothing exercised either of them, so a typo in one of 324 lines would have shipped as a subtly wrong sort.
  - `tests/client/deckeditor-color-picker.test.js` covers which property rows offer the color picker and which card types get a sorting tooltip, including the keystroke-by-keystroke sequences `red` → `S` → `Spades` and `♠` → *(cleared)* → `Spades`, and a card type that carries only part of the set.
- The whole TestCafe `editor.js` suite (45 tests) passes, including the one that asserts a card type property added with type `color` *does* get the picker.
- In a real browser (Playwright/Chromium): added the deck through the deck editor's *Traditional deck* flow, confirmed all 54 cards get the six properties, that every sorting row carries its tooltip and that the `suitColor` row behaves like this, typed character by character with the button read out of the DOM after each keystroke ([card type panel](https://agent.virtualtabletop.io/reports/pr/1538991387469545602/cardtype-panel.png)):

  ```
  "♠" no picker | "" PICKER | "S" no | "Sp" no | "Spades" no | "♠" no
  "red" PICKER  | "S" no    | "Sp" no | "Spades" no
  "" PICKER | "#" PICKER | "#f" PICKER | "#ff0000" PICKER | "rgb(0, 0" PICKER | "rgb(0, 0, 255)" PICKER | "♠" no
  ```

  The field stays 206px wide through all of it. The Spanish deck's card type panel shows no sorting tooltips on its `rank`/`suit`/`suitColor` rows. `SORT` on each key against the created holder gives:

```
suit       2c 3c … Kc | 2d … Kd | 2h … Kh | 2s … Ks | Jk1 Jk2
suitColor  spades…, clubs…, hearts…, diamonds…, jokers      (black together, red together)
suitAlt    clubs…, hearts…, spades…, diamonds…, jokers      (alternating)
rank       Ac Ah As Ad 2c … Kd Jk1 Jk2                      (ace low)
rankA      2c … Kd Ac Ah As Ad Jk1 Jk2                      (ace high)
rankFixed  Ac Ad Ah As 2c 2d 2h 2s … Kc Kd Kh Ks Jk1 Jk2    (fixed order)
```

No new or renamed widget property, routine operation or parameter, so `validator/` (`cardTypes` is validated as `any`) and the `jeCommands` list in `client/js/jsonedit.js` need no change, and no UI affordance is added — the properties are edited in the deck editor's card type panel like any other card type property.

## Test room

The test room lives on this PR's test server, **not** in `library/tutorials/` — no tutorial and no library file is touched by this PR. It is authored to tutorial quality per `docs/tutorials.md` and ships as one game with two variants:

- **Properties - Card sorting — *Suits*** (`0.json`): one button each for `suit`, `suitColor` and `suitAlt`.
- **Properties - Card sorting — *Ranks*** (`1.json`): one button each for `rank`, `rankA` and `rankFixed`.

Each button shuffles the whole 54-card deck, sorts it by its property (plus a tie-breaker, so the cards inside a group are ordered instead of staying shuffled) and deals it out into a 13-wide grid — so sorting by a suit property puts exactly one suit per row and the two jokers on the last row.

Download: [Properties - Card sorting.vtt](https://agent.virtualtabletop.io/reports/pr/1538991387469545602/Properties%20-%20Card%20sorting.vtt) · screenshots: [Suits](https://agent.virtualtabletop.io/reports/pr/1538991387469545602/card-sorting-suits.png) · [Ranks](https://agent.virtualtabletop.io/reports/pr/1538991387469545602/card-sorting-ranks.png).

Tutorial-quality checks, for the record:

- **Template:** `library/tutorials/Properties - dragLimit` (the newest tutorial by `_meta.info.lastUpdate`, 2026-08-09), cross-read with `Types - Line` and `Types - Piles`.
- **`_meta.info` parity:** the same fields in the same order and shapes as the template — `name` in the `Properties - <Name>` category scheme, `mode: "Tutorial"`, `players: "1"`, `time: "0"`, `year: "2026"`, `language: "en-US"`, `showName: false`, `usesAIImagery: false`, empty `rules`/`bgg`/`attribution`/`skill`/`description`/`ruleText`/`helpText`/`similar*`, and `variant` `"Suits"` / `"Ranks"` sharing one `name` and `image`.
- **Layout:** the documented top-down bands on the 1600 x 1000 board — `title` (60px, centred), `overview` (25px paragraph), a row of three example buttons with captions directly underneath (22px), then the card grid; semantic ids (`title`, `overview`, `text1..4`, `button1..3`, `cell-<row>-<col>`, `deckHolder`/`deckHolderD`/`deckHolderP`/`deckHolderB`), explanatory text on `layer: -3` and `movable`/`movableInEdit` false, the conventional 111x40 **Recall & Shuffle** button parented to the deck holder, and no invented navigation chrome.
- **Library image:** a 530x530 square SVG in the tutorials' visual language — the same square badge, the `Properties` category word and the graduate cap taken verbatim from the template's image, plus a three-card symbol — embedded in the `.vtt` under `assets/<crc32>_<length>` and referenced as `_meta.info.image`. It sits next to the real tutorials in the game list without standing out.
- **Validated and walked:** `node validator/validate_gamefile_node.js` exits 0 with no output on both variants, and both were walked end to end in Chromium — all six buttons pressed, the dealt order read back out of the DOM (`suit` → one suit per row, jokers last; `rank` → four aces first; `rankA` → aces last; …), and the `.vtt` re-uploaded into a `pr-test-ai` room to confirm it loads with its preview image.

It can be moved into `library/tutorials/` and committed here if a maintainer wants it as a real tutorial — that is a call for a human, so the diff deliberately contains no library change.

## Out of scope

- The *custom deck* wizard (Suits and ranks) builds decks from user-defined suits and ranks, where `suitAlt`/`rankA`/`rankFixed` have no general meaning; it still emits only `suit`, `suitColor` and `rank`.
- The other seven traditional decks (`german`, `spanish`, `swiss`, `tarot`, `hanafuda`, `mahjong`, `dominoes`) use a different convention — numeric `suit` and `order`, a readable `rank` — and sorting them by `rank` still gives `1, 10, 11, 12, 2, …`. Only the standard deck has a de-facto convention to match (the library card games), so unifying the rest is a separate decision and a separate PR.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3125/pr-test (or any other room on that server)

### DIFF
--- a/assets/decks/standard.json
+++ b/assets/decks/standard.json
@@ -1,6 +1,6 @@
 {
   "name": "Standard playing cards",
-  "description": "French-suited 52-card deck with 2 jokers.",
+  "description": "French-suited 52-card deck with 2 jokers. The card types carry sorting keys: suit, suitColor and suitAlt group the suits, rank (ace low), rankA (ace high) and rankFixed order the cards.",
   "attribution": "Card images by Adrian Kennard, released under CC0 at https://www.me.uk/cards/. Jokers based on Byron Knoll's public domain playing cards.",
   "deck": {
     "type": "deck",
--- a/assets/decks/standard.json
+++ b/assets/decks/standard.json
@@ -1,6 +1,6 @@
 {
   "name": "Standard playing cards",
-  "description": "French-suited 52-card deck with 2 jokers. The card types carry sorting keys: suit, suitColor and suitAlt group the suits, rank (ace low), rankA (ace high) and rankFixed order the cards.",
+  "description": "French-suited 52-card deck with 2 jokers. The cards carry extra properties so a hand or a pile can be sorted by suit or by rank.",
   "attribution": "Card images by Adrian Kennard, released under CC0 at https://www.me.uk/cards/. Jokers based on Byron Knoll's public domain playing cards.",
   "deck": {
     "type": "deck",
--- a/assets/decks/standard.json
+++ b/assets/decks/standard.json
@@ -7,271 +7,489 @@
     "cardTypes": {
       "2 of spades": {
         "image": "/i/cards-default/2S.svg",
-        "rank": "2",
-        "suit": "spades"
+        "suit": "S",
+        "suitColor": "♠",
+        "suitAlt": "3♠",
+        "rank": "02",
+        "rankA": "02",
+        "rankFixed": "02 S"
       },
       "3 of spades": {
         "image": "/i/cards-default/3S.svg",
-        "rank": "3",
-        "suit": "spades"
+        "suit": "S",
+        "suitColor": "♠",
+        "suitAlt": "3♠",
+        "rank": "03",
+        "rankA": "03",
+        "rankFixed": "03 S"
       },
       "4 of spades": {
         "image": "/i/cards-default/4S.svg",
-        "rank": "4",
-        "suit": "spades"
+        "suit": "S",
+        "suitColor": "♠",
+        "suitAlt": "3♠",
+        "rank": "04",
+        "rankA": "04",
+        "rankFixed": "04 S"
       },
       "5 of spades": {
         "image": "/i/cards-default/5S.svg",
-        "rank": "5",
-        "suit": "spades"
+        "suit": "S",
+        "suitColor": "♠",
+        "suitAlt": "3♠",
+        "rank": "05",
+        "rankA": "05",
+        "rankFixed": "05 S"
       },
       "6 of spades": {
         "image": "/i/cards-default/6S.svg",
-        "rank": "6",
-        "suit": "spades"
+        "suit": "S",
+        "suitColor": "♠",
+        "suitAlt": "3♠",
+        "rank": "06",
+        "rankA": "06",
+        "rankFixed": "06 S"
       },
       "7 of spades": {
         "image": "/i/cards-default/7S.svg",
-        "rank": "7",
-        "suit": "spades"
+        "suit": "S",
+        "suitColor": "♠",
+        "suitAlt": "3♠",
+        "rank": "07",
+        "rankA": "07",
+        "rankFixed": "07 S"
       },
       "8 of spades": {
         "image": "/i/cards-default/8S.svg",
-        "rank": "8",
-        "suit": "spades"
+        "suit": "S",
+        "suitColor": "♠",
+        "suitAlt": "3♠",
+        "rank": "08",
+        "rankA": "08",
+        "rankFixed": "08 S"
       },
       "9 of spades": {
         "image": "/i/cards-default/9S.svg",
-        "rank": "9",
-        "suit": "spades"
+        "suit": "S",
+        "suitColor": "♠",
+        "suitAlt": "3♠",
+        "rank": "09",
+        "rankA": "09",
+        "rankFixed": "09 S"
       },
       "A of spades": {
         "image": "/i/cards-default/AS.svg",
-        "rank": "A",
-        "suit": "spades"
+        "suit": "S",
+        "suitColor": "♠",
+        "suitAlt": "3♠",
+        "rank": "01",
+        "rankA": "5A",
+        "rankFixed": "01 S"
       },
       "10 of spades": {
         "image": "/i/cards-default/TS.svg",
+        "suit": "S",
+        "suitColor": "♠",
+        "suitAlt": "3♠",
         "rank": "10",
-        "suit": "spades"
+        "rankA": "10",
+        "rankFixed": "10 S"
       },
       "J of spades": {
         "image": "/i/cards-default/JS.svg",
-        "rank": "J",
-        "suit": "spades"
+        "suit": "S",
+        "suitColor": "♠",
+        "suitAlt": "3♠",
+        "rank": "2J",
+        "rankA": "2J",
+        "rankFixed": "2J S"
       },
       "Q of spades": {
         "image": "/i/cards-default/QS.svg",
-        "rank": "Q",
-        "suit": "spades"
+        "suit": "S",
+        "suitColor": "♠",
+        "suitAlt": "3♠",
+        "rank": "3Q",
+        "rankA": "3Q",
+        "rankFixed": "3Q S"
       },
       "K of spades": {
         "image": "/i/cards-default/KS.svg",
-        "rank": "K",
-        "suit": "spades"
+        "suit": "S",
+        "suitColor": "♠",
+        "suitAlt": "3♠",
+        "rank": "4K",
+        "rankA": "4K",
+        "rankFixed": "4K S"
       },
       "2 of hearts": {
         "image": "/i/cards-default/2H.svg",
-        "rank": "2",
-        "suit": "hearts"
+        "suit": "H",
+        "suitColor": "♥",
+        "suitAlt": "2♥",
+        "rank": "02",
+        "rankA": "02",
+        "rankFixed": "02 H"
       },
       "3 of hearts": {
         "image": "/i/cards-default/3H.svg",
-        "rank": "3",
-        "suit": "hearts"
+        "suit": "H",
+        "suitColor": "♥",
+        "suitAlt": "2♥",
+        "rank": "03",
+        "rankA": "03",
+        "rankFixed": "03 H"
       },
       "4 of hearts": {
         "image": "/i/cards-default/4H.svg",
-        "rank": "4",
-        "suit": "hearts"
+        "suit": "H",
+        "suitColor": "♥",
+        "suitAlt": "2♥",
+        "rank": "04",
+        "rankA": "04",
+        "rankFixed": "04 H"
       },
       "5 of hearts": {
         "image": "/i/cards-default/5H.svg",
-        "rank": "5",
-        "suit": "hearts"
+        "suit": "H",
+        "suitColor": "♥",
+        "suitAlt": "2♥",
+        "rank": "05",
+        "rankA": "05",
+        "rankFixed": "05 H"
       },
       "6 of hearts": {
         "image": "/i/cards-default/6H.svg",
-        "rank": "6",
-        "suit": "hearts"
+        "suit": "H",
+        "suitColor": "♥",
+        "suitAlt": "2♥",
+        "rank": "06",
+        "rankA": "06",
+        "rankFixed": "06 H"
       },
       "7 of hearts": {
         "image": "/i/cards-default/7H.svg",
-        "rank": "7",
-        "suit": "hearts"
+        "suit": "H",
+        "suitColor": "♥",
+        "suitAlt": "2♥",
+        "rank": "07",
+        "rankA": "07",
+        "rankFixed": "07 H"
       },
       "8 of hearts": {
         "image": "/i/cards-default/8H.svg",
-        "rank": "8",
-        "suit": "hearts"
+        "suit": "H",
+        "suitColor": "♥",
+        "suitAlt": "2♥",
+        "rank": "08",
+        "rankA": "08",
+        "rankFixed": "08 H"
       },
       "9 of hearts": {
         "image": "/i/cards-default/9H.svg",
-        "rank": "9",
-        "suit": "hearts"
+        "suit": "H",
+        "suitColor": "♥",
+        "suitAlt": "2♥",
+        "rank": "09",
+        "rankA": "09",
+        "rankFixed": "09 H"
       },
       "A of hearts": {
         "image": "/i/cards-default/AH.svg",
-        "rank": "A",
-        "suit": "hearts"
+        "suit": "H",
+        "suitColor": "♥",
+        "suitAlt": "2♥",
+        "rank": "01",
+        "rankA": "5A",
+        "rankFixed": "01 H"
       },
       "10 of hearts": {
         "image": "/i/cards-default/TH.svg",
+        "suit": "H",
+        "suitColor": "♥",
+        "suitAlt": "2♥",
         "rank": "10",
-        "suit": "hearts"
+        "rankA": "10",
+        "rankFixed": "10 H"
       },
       "J of hearts": {
         "image": "/i/cards-default/JH.svg",
-        "rank": "J",
-        "suit": "hearts"
+        "suit": "H",
+        "suitColor": "♥",
+        "suitAlt": "2♥",
+        "rank": "2J",
+        "rankA": "2J",
+        "rankFixed": "2J H"
       },
       "Q of hearts": {
         "image": "/i/cards-default/QH.svg",
-        "rank": "Q",
-        "suit": "hearts"
+        "suit": "H",
+        "suitColor": "♥",
+        "suitAlt": "2♥",
+        "rank": "3Q",
+        "rankA": "3Q",
+        "rankFixed": "3Q H"
       },
       "K of hearts": {
         "image": "/i/cards-default/KH.svg",
-        "rank": "K",
-        "suit": "hearts"
+        "suit": "H",
+        "suitColor": "♥",
+        "suitAlt": "2♥",
+        "rank": "4K",
+        "rankA": "4K",
+        "rankFixed": "4K H"
       },
       "2 of diamonds": {
         "image": "/i/cards-default/2D.svg",
-        "rank": "2",
-        "suit": "diamonds"
+        "suit": "D",
+        "suitColor": "♦",
+        "suitAlt": "4♦",
+        "rank": "02",
+        "rankA": "02",
+        "rankFixed": "02 D"
       },
       "3 of diamonds": {
         "image": "/i/cards-default/3D.svg",
-        "rank": "3",
-        "suit": "diamonds"
+        "suit": "D",
+        "suitColor": "♦",
+        "suitAlt": "4♦",
+        "rank": "03",
+        "rankA": "03",
+        "rankFixed": "03 D"
       },
       "4 of diamonds": {
         "image": "/i/cards-default/4D.svg",
-        "rank": "4",
-        "suit": "diamonds"
+        "suit": "D",
+        "suitColor": "♦",
+        "suitAlt": "4♦",
+        "rank": "04",
+        "rankA": "04",
+        "rankFixed": "04 D"
       },
       "5 of diamonds": {
         "image": "/i/cards-default/5D.svg",
-        "rank": "5",
-        "suit": "diamonds"
+        "suit": "D",
+        "suitColor": "♦",
+        "suitAlt": "4♦",
+        "rank": "05",
+        "rankA": "05",
+        "rankFixed": "05 D"
       },
       "6 of diamonds": {
         "image": "/i/cards-default/6D.svg",
-        "rank": "6",
-        "suit": "diamonds"
+        "suit": "D",
+        "suitColor": "♦",
+        "suitAlt": "4♦",
+        "rank": "06",
+        "rankA": "06",
+        "rankFixed": "06 D"
       },
       "7 of diamonds": {
         "image": "/i/cards-default/7D.svg",
-        "rank": "7",
-        "suit": "diamonds"
+        "suit": "D",
+        "suitColor": "♦",
+        "suitAlt": "4♦",
+        "rank": "07",
+        "rankA": "07",
+        "rankFixed": "07 D"
       },
       "8 of diamonds": {
         "image": "/i/cards-default/8D.svg",
-        "rank": "8",
-        "suit": "diamonds"
+        "suit": "D",
+        "suitColor": "♦",
+        "suitAlt": "4♦",
+        "rank": "08",
+        "rankA": "08",
+        "rankFixed": "08 D"
       },
       "9 of diamonds": {
         "image": "/i/cards-default/9D.svg",
-        "rank": "9",
-        "suit": "diamonds"
+        "suit": "D",
+        "suitColor": "♦",
+        "suitAlt": "4♦",
+        "rank": "09",
+        "rankA": "09",
+        "rankFixed": "09 D"
       },
       "A of diamonds": {
         "image": "/i/cards-default/AD.svg",
-        "rank": "A",
-        "suit": "diamonds"
+        "suit": "D",
+        "suitColor": "♦",
+        "suitAlt": "4♦",
+        "rank": "01",
+        "rankA": "5A",
+        "rankFixed": "01 D"
       },
       "10 of diamonds": {
         "image": "/i/cards-default/TD.svg",
+        "suit": "D",
+        "suitColor": "♦",
+        "suitAlt": "4♦",
         "rank": "10",
-        "suit": "diamonds"
+        "rankA": "10",
+        "rankFixed": "10 D"
       },
       "J of diamonds": {
         "image": "/i/cards-default/JD.svg",
-        "rank": "J",
-        "suit": "diamonds"
+        "suit": "D",
+        "suitColor": "♦",
+        "suitAlt": "4♦",
+        "rank": "2J",
+        "rankA": "2J",
+        "rankFixed": "2J D"
       },
       "Q of diamonds": {
         "image": "/i/cards-default/QD.svg",
-        "rank": "Q",
-        "suit": "diamonds"
+        "suit": "D",
+        "suitColor": "♦",
+        "suitAlt": "4♦",
+        "rank": "3Q",
+        "rankA": "3Q",
+        "rankFixed": "3Q D"
       },
       "K of diamonds": {
         "image": "/i/cards-default/KD.svg",
-        "rank": "K",
-        "suit": "diamonds"
+        "suit": "D",
+        "suitColor": "♦",
+        "suitAlt": "4♦",
+        "rank": "4K",
+        "rankA": "4K",
+        "rankFixed": "4K D"
       },
       "2 of clubs": {
         "image": "/i/cards-default/2C.svg",
-        "rank": "2",
-        "suit": "clubs"
+        "suit": "C",
+        "suitColor": "♣",
+        "suitAlt": "1♣",
+        "rank": "02",
+        "rankA": "02",
+        "rankFixed": "02 C"
       },
       "3 of clubs": {
         "image": "/i/cards-default/3C.svg",
-        "rank": "3",
-        "suit": "clubs"
+        "suit": "C",
+        "suitColor": "♣",
+        "suitAlt": "1♣",
+        "rank": "03",
+        "rankA": "03",
+        "rankFixed": "03 C"
       },
       "4 of clubs": {
         "image": "/i/cards-default/4C.svg",
-        "rank": "4",
-        "suit": "clubs"
+        "suit": "C",
+        "suitColor": "♣",
+        "suitAlt": "1♣",
+        "rank": "04",
+        "rankA": "04",
+        "rankFixed": "04 C"
       },
       "5 of clubs": {
         "image": "/i/cards-default/5C.svg",
-        "rank": "5",
-        "suit": "clubs"
+        "suit": "C",
+        "suitColor": "♣",
+        "suitAlt": "1♣",
+        "rank": "05",
+        "rankA": "05",
+        "rankFixed": "05 C"
       },
       "6 of clubs": {
         "image": "/i/cards-default/6C.svg",
-        "rank": "6",
-        "suit": "clubs"
+        "suit": "C",
+        "suitColor": "♣",
+        "suitAlt": "1♣",
+        "rank": "06",
+        "rankA": "06",
+        "rankFixed": "06 C"
       },
       "7 of clubs": {
         "image": "/i/cards-default/7C.svg",
-        "rank": "7",
-        "suit": "clubs"
+        "suit": "C",
+        "suitColor": "♣",
+        "suitAlt": "1♣",
+        "rank": "07",
+        "rankA": "07",
+        "rankFixed": "07 C"
       },
       "8 of clubs": {
         "image": "/i/cards-default/8C.svg",
-        "rank": "8",
-        "suit": "clubs"
+        "suit": "C",
+        "suitColor": "♣",
+        "suitAlt": "1♣",
+        "rank": "08",
+        "rankA": "08",
+        "rankFixed": "08 C"
       },
       "9 of clubs": {
         "image": "/i/cards-default/9C.svg",
-        "rank": "9",
-        "suit": "clubs"
+        "suit": "C",
+        "suitColor": "♣",
+        "suitAlt": "1♣",
+        "rank": "09",
+        "rankA": "09",
+        "rankFixed": "09 C"
       },
       "A of clubs": {
         "image": "/i/cards-default/AC.svg",
-        "rank": "A",
-        "suit": "clubs"
+        "suit": "C",
+        "suitColor": "♣",
+        "suitAlt": "1♣",
+        "rank": "01",
+        "rankA": "5A",
+        "rankFixed": "01 C"
       },
       "10 of clubs": {
         "image": "/i/cards-default/TC.svg",
+        "suit": "C",
+        "suitColor": "♣",
+        "suitAlt": "1♣",
         "rank": "10",
-        "suit": "clubs"
+        "rankA": "10",
+        "rankFixed": "10 C"
       },
       "J of clubs": {
         "image": "/i/cards-default/JC.svg",
-        "rank": "J",
-        "suit": "clubs"
+        "suit": "C",
+        "suitColor": "♣",
+        "suitAlt": "1♣",
+        "rank": "2J",
+        "rankA": "2J",
+        "rankFixed": "2J C"
       },
       "Q of clubs": {
         "image": "/i/cards-default/QC.svg",
-        "rank": "Q",
-        "suit": "clubs"
+        "suit": "C",
+        "suitColor": "♣",
+        "suitAlt": "1♣",
+        "rank": "3Q",
+        "rankA": "3Q",
+        "rankFixed": "3Q C"
       },
       "K of clubs": {
         "image": "/i/cards-default/KC.svg",
-        "rank": "K",
-        "suit": "clubs"
+        "suit": "C",
+        "suitColor": "♣",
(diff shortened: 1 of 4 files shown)


VTT-AI-Agent (commented):
## TL;DR

The data is correct: I re-derived all 54 card types from `generateCardDeckWidgets` (`client/js/editmode.js:13-25`) and every one of the 6 properties matches exactly, and sorting by each of the six keys with `String.localeCompare` (what `sortWidgets` in `client/js/main.js:601` uses) reproduces the orders claimed in the description. The one thing worth a decision before merge is that `suitColor` now holds `"♠"` in the deck editor's *own* template, and the deck editor offers a **color picker** on any property whose name matches `/color/i` — one click there silently overwrites the sort key with a hex value.

## Findings

### 1. `suitColor: "♠"` gets a color picker in the deck editor (main finding)

`client/js/editor/deckeditor.js:3223`:

```js
shouldOfferColorPicker(property, value) {
  return /color/i.test(String(property)) || this.isColorValue(value);
}
```

So in the deck editor's card-type property panel, the new `suitColor` row gets the palette button. `parseColor("♠")` returns `null`, so the swatch renders empty (no color, just the palette glyph), and if the user opens the picker and clicks any color, `picker.setValue` writes `#rrggbb` straight over `"♠"` — the `suitColor` sort silently degrades to "all four suits are now the same colour except the ones I clicked". There is no hint in the UI that this field is a sort key rather than a colour.

This matters more here than for the add-widget-overlay deck, because this PR puts the value into the file the **deck editor** loads (`i/decks/standard.json`, fetched in `properties.js:2131`), i.e. the tool whose whole job is editing card-type properties.

The root cause is that `suitColor` is overloaded: in the custom-deck wizard it really *is* a CSS colour (`properties.js:2448` `suitColor: suit.color`, bound as `dynamicProperties: { color: "suitColor" }` by every design template), while in the overlay deck and the library card games it is a sort key. That collision predates this PR, and matching Hearts/Spades is clearly the right call — but a small follow-up would remove the trap, e.g. don't offer the picker on a `/color/i` property whose current value is not `isColorValue()` (only `isColorValue` decides), or special-case `suitColor` by name. Your call whether that belongs in this PR.

### 2. `rank` is no longer human-readable in the beginner-facing template

Before: `rank: "K"`, `"A"`, `"2"`. After: `rank: "4K"`, `"01"`, `"02"`. Nothing renders it today — `standard.json`'s `faceTemplates` bind only `image` — so there is no visual regression. But the deck editor is exactly where a user builds a face by binding a text object to a card-type property, and doing the obvious thing now prints `4K` / `2J` / `01` on the card instead of `K` / `J` / `A`. The card-type *key* ("K of spades") still carries the readable name, so nothing is lost, just less convenient.

Worth considering (strictly additive, no impact on parity with the library games): a readable `label`/`name` property alongside the sort keys, or a sentence in the deck's `description` field saying the cards carry sorting properties, since that text is shown in the Traditional-deck dialog.

### 3. The same bug is still present in the other seven traditional decks

`assets/decks/german.json`, `spanish.json`, `swiss.json`, `tarot.json` use a completely different convention: numeric `suit` (1..4), numeric `order`, readable `rank` (`"A"`, `"5"`, `"1"`). `sortWidgets` compares numbers numerically (`client/js/main.js:594`), so their `order` works — but sorting the Spanish deck by `rank` still yields `1, 10, 11, 12, 2, 3, …`, i.e. exactly the defect this PR fixes for the standard deck, and none of them has `suitColor`/`suitAlt`/`rankA`/`rankFixed`.

After this PR `standard.json` is the odd one out in its own folder. That is defensible (it is the only one whose values have a de-facto standard in the library games — I confirmed `library/games/Hearts/0.json` and `Spades/0.json` carry these exact six values), but it is worth a follow-up issue so a game author picking German vs. standard doesn't get two unrelated sorting conventions. Mentioning it in the PR description as known scope would help too.

### 4. No test guards the two hand-maintained copies

`assets/decks/standard.json` now duplicates, by hand, the table `generateCardDeckWidgets` generates in `client/js/editmode.js:13-25`. Nothing in `tests/` exercises either the Traditional-deck flow or this file (I grepped: no test references `decks/`, `traditional`, or `deckEditorNewDeckMode`), so a single typo in one of 324 new lines — `"3♠"` on a diamond, `"03 H"` on a club — would ship unnoticed and only show up as a subtly wrong sort in someone's game.

A pure-data jest test would be cheap and needs no DOM: load `assets/decks/standard.json`, assert 54 types each carrying the six properties, and assert that sorting the keys by each of the six with `localeCompare` gives the documented order. I ran exactly that check against this branch and it passes today (suit → C/D/H/S then jokers; suitColor → ♠♣ then ♥♦; suitAlt → ♣♥♠♦; rank → ace low; rankA → ace high; rankFixed → rank then suit), so the test would lock in a state that is already correct.

### 5. Description accuracy (minor)

The "Why" section says the recipes in the **Functions - SORT** tutorial didn't work on this deck. That tutorial's deck actually uses its own properties (`denom`, and `suit` as `"01"`.., see `library/tutorials/Functions - SORT/0.json`), not `suitAlt`/`rankA`/`rankFixed`. The real beneficiaries are the library card games, whose decks carry precisely these six values — which is the stronger argument anyway.

## Checks

- **UI support:** nothing new is exposed to game authors — no new widget property, routine operation or parameter function — and card-type properties are editable in the deck editor's card-type panel, so no new editor affordance is owed. (Finding 1 is about the *wrong* control being offered for an existing one.)
- **Test room:** a playable demo is on the PR's test server with a downloadable `.vtt` (the link resolves), and the diff touches nothing under `library/` — correct per the tutorials rule.
- CI is green on all checks, including CodeQL.

I added the `beta` label: the diff is a single data template (`assets/decks/standard.json`) that touches no engine file and only affects decks created from now on, so it cannot corrupt existing games or crash the server and is safe for players to try.

VTT-AI-Agent:
Thanks — all five points looked at, three of them are now in the branch (30f3223).

**1. `suitColor: "♠"` gets a color picker — fixed here.** You are right that this PR is what puts the trap into the deck editor's own template, so it belongs in this PR. `shouldOfferColorPicker()` no longer offers the picker on the property *name* alone: it offers it while the value can actually be a color — it already is one (`isColorValue`), the field is empty, or a color is being typed into it (`#f`, `rgb(255, 0`, `re`). That last part matters because the check re-runs on every keystroke, so a plain "must parse as a color" rule would have made the button disappear and reappear under the cursor while someone retypes a hex value, and closed an open picker mid-edit.

Verified in a browser on a freshly added Traditional standard deck: the `suitColor` row has no picker while it holds `♠`, gets one as soon as `#ff0000` is typed into it, keeps it at `#f`, and loses it again when `♠` comes back — while a face object's `color: transparent` still has its picker. The existing TestCafe assertion that a card type property added with type `color` gets the picker (`editor.js`, "create deck from scratch with color box, face and defaults") still passes, and `tests/client/deckeditor-color-picker.test.js` now covers the rule.

**2. `rank` is not human-readable — no `label` property, but the description now says so.** Adding a readable property would break the parity with `generateCardDeckWidgets` that this PR is about, and the card type key already carries the readable name. What was genuinely missing is the hint, so the deck's `description` — the text the Traditional deck dialog shows — now reads: *"French-suited 52-card deck with 2 jokers. The card types carry sorting keys: suit, suitColor and suitAlt group the suits, rank (ace low), rankA (ace high) and rankFixed order the cards."*

**3. The other seven decks — agreed, and now stated as known scope** in the PR description rather than silently left out. They use a different convention altogether (numeric `suit`/`order`), so giving them the same six keys is a decision about all traditional decks, not a follow-up to this one; happy to open it as a separate PR if you want it.

**4. No test guards the hand-copied table — added.** `tests/standard-deck.test.js` derives each card type from its own name (suit letter/color/alt, rank, `rankFixed`, image path) and compares the whole file against that, then asserts the six sort orders by grouping the `localeCompare`-sorted card names by suit and by rank. A diamond carrying the spades `suitColor`, or `03 H` on a club, fails it. `npm test` is at 1275 tests.

**5. Description inaccuracy — corrected.** The "Why" section no longer claims the *Functions - SORT* tutorial's recipes apply here (that tutorial's deck has its own `denom`/`suit` properties); it now names the library card games, whose decks do carry these exact six values.

On the `beta` label: noted, thanks — it is not mine to add. The diff is no longer data-only, though: it now touches `client/js/editor/deckeditor.js` (which property rows get a color picker) plus two test files, so that call may be worth re-making.

VTT-AI-Agent (commented):
## TL;DR

This is a **UI/UX review** (layout, visual design, first-time-user readability) — not a code review; correctness is covered elsewhere. The visible change is small and nothing regresses visually: cards render exactly as before, no overlay breaks, and the color-picker fix does the right thing on real color rows (`color: transparent` still gets its swatch). The one thing worth changing: **the six new sort keys are explained only in the Traditional-deck picker — a screen the user leaves and can never get back to — and are completely unexplained in the Card Type panel where they actually show up as `suit: S`, `suitAlt: 3♠`, `rankFixed: 02 S`.** Put a one-line tooltip on those rows and shorten the picker description.

**How I reviewed it:** checked the branch out, ran the server locally and drove Chromium with Playwright. Overlays exercised at **390×844, 820×1180, 1280×800 and 1920×1080**: Game Shelf / public-library browser, States & save, Players, About/help, the Add-widget overlay, the JSON editor, edit mode's *Edit widgets* panel + right sidebar, and the whole deck editor (tree, card canvas, card-types strip, all four sidebar tabs, *Add a new deck* with all three sections and the Traditional-deck picker). Overlays I did **not** reach in this pass: chat/seat panels, share/publish, and the confirm prompts. Below 700px wide the app itself shows "Please rotate your device to landscape orientation", so the 390×844 pass is documentary rather than a target.

---

## Big layout improvements

### 1. Explain the sort keys where they are *used*, not only where the deck is *picked*

`assets/decks/standard.json:3` puts the whole explanation into the deck's `description`. That string is rendered in exactly one place — `client/js/editor/sidebar/properties.js:2145`, the details block of the Traditional-deck picker — and it is gone the moment "Add to game" is clicked. Meanwhile the panel where a user actually meets these values, the deck editor's **Card Type properties**, shows this and says nothing:

card type panel

```
suit        S
suitColor   ♠
suitAlt     3♠
rank        02
rankA       02
rankFixed   02 S
```

For a first-time user that is six rows of unexplained codes on the first panel they open after clicking a card, and it is a step *down* from what was there before (`rank: 2`, `suit: spades` read themselves). Two of them are actively misleading at a glance: `suitColor` holds a glyph and no color, and `rank: 02` / `2J` / `4K` will render as literally `2J` if someone binds a text object to `rank` to print the rank on a card.

Concrete suggestion, in the editor's existing idiom: in `addTypeInput` (`client/js/editor/deckeditor.js:2391`) set `row.dom.title` from a small map of known card-type keys, the same way the face-object rows already get one at `deckeditor.js:2664`. Something like:

- `suit` — "Sorting key: groups the four suits (C, D, H, S; T for the jokers)."
- `suitColor` — "Sorting key: keeps the two black and the two red suits together. Not a CSS color."
- `suitAlt` — "Sorting key: alternates black and red suits."
- `rank` — "Sorting key: ace low. The digits and letters exist to make SORT order correctly — not for display."
- `rankA` — "Sorting key: ace high."
- `rankFixed` — "Sorting key: one fixed order for the whole deck."

A single hint line above the group (the editor already has that pattern — `deckGeneratorHint` in `properties.js`) would work too, and would carry the "don't retype these by hand, SORT depends on the exact strings" warning that nothing currently gives.

### 2. Trim the picker description — it is now the only deck that talks about its internals

Side by side with its seven siblings, the standard deck's description no longer describes the *deck*:

| deck | description |
|---|---|
| Spanish | "Spanish-suited 48-card deck (baraja española) with coins, cups, swords and clubs." |
| Hanafuda | "Japanese flower cards: 48 cards in 12 monthly suits for games like Koi-Koi and Hachi-Hachi." |
| **Standard** | "French-suited 52-card deck with 2 jokers. **The card types carry sorting keys: suit, suitColor and suitAlt group the suits, rank (ace low), rankA (ace high) and rankFixed order the cards.**" |

Measured in the browser, the description block goes from 19px to 57px tall at ≥820 wide and from 38px to **114px** at 390 wide (the details container: 71 → 109px, and 118 → 194px). It is also the deck-*choosing* moment: "sorting keys" is jargon a newcomer has no referent for yet, and it sits below the eight deck tiles — at 1280×800 the whole details block and the green **Add to game** button start at y≈896 in an 800px viewport, i.e. off-screen until you scroll, so this paragraph is easy to miss entirely and easy to trip over if you don't.

details block
details block at 390

Suggestion: keep the picker line game-facing and one clause long — e.g. *"French-suited 52-card deck with 2 jokers. The cards carry extra properties so a hand or pile can be sorted by suit or rank."* — and let the per-row tooltips from #1 carry the six names. That keeps this deck's entry the same shape as the other seven and puts the detail where it is actionable.

---

### The row width jumps while you are typing in it

`client/css/editor/deckeditor.css:1649` gives `.genericInput.hasColorPicker` `padding-right: 48px` instead of 24px, and `shouldOfferColorPicker` toggles that class on every `input` event. Measured on the `suitColor` row at 1280×800: value `♠` → input **230px**, no button; clear the field → input **206px**, palette button pops in under the cursor. So selecting-all and retyping a value makes the field visibly reflow mid-edit. The `isPartialColorValue` helper (`deckeditor.js:3224`) already removes the flicker *while typing a color*, but not this first jump. Cheap fix: keep `padding-right: 48px` on every color-named row and just toggle the button's `visibility` rather than `display`, so the field width never changes.

### A letters-only value still gets a picker it can destroy

`shouldOfferColorPicker` (`deckeditor.js:3235`) accepts any all-letters string as "partial". Typing `Spades` into `suitColor` therefore offers the palette button on a value that is not a color and never will be — the same shape of trap this PR set out to close, just narrower:

suitColor = Spades still offers the picker

Low severity (it takes a deliberate edit to get there), but if you want it tight, gate the letters branch on the text being a *prefix* of a CSS color keyword rather than on "is letters" — that keeps `b → bl → blu → blue` flicker-free without arming the picker on `Spades` or `clubs`.

### Right edges no longer line up while a picker is showing

Consequence of the same 24px: while `suitColor` shows its swatch its input ends 24px short of `suit`, `suitAlt`, `rank`… above and below it, so the column of field edges has one notch in it. Inherent to the current design (the asset-picker rows do the same on the `image` row), and it disappears the moment the sort key is back — noting it only because this PR is what makes a picker appear and disappear on a *card-type* row for the first time.

### Not caused by this PR — checked and cleared

At 390×844 the deck editor's sidebar overflows the viewport (measured right edge 580px in a 390px window) and the card-canvas hint text runs straight through the property labels. I verified against the **Spanish** deck, which is untouched by this PR and also renders 7 rows: it breaks identically. Pre-existing responsive gap in the deck editor, not a regression here — but with the standard deck now at 7 rows instead of 3, four of the labels land inside the collision zone where before they sat above it.

390 wide, standard deck
390 wide, Spanish deck — same breakage

---

1. **`suitColor` means two different things inside one editor.** In the custom-deck flow the deck editor binds face objects' `color` to `suitColor` (`properties.js:3203`, `:3213`, `:3256`, … — a real CSS color), and now the standard deck's `suitColor` holds `♠`. Same name, same panel, opposite meaning. The convention comes from `editmode.js:18-22` so it is not this PR's to change, but it is this PR that makes it visible in the deck editor — which is exactly why the tooltip in #1 should say "not a CSS color" out loud.
2. **"sorting keys" is developer vocabulary.** Nothing else in the deck-picking UI uses "key". "Extra properties for sorting" or "sort order values" would land with someone who has never opened a routine.
3. **Nothing warns that the exact strings matter.** `rank: 02` looks like a typo for `2`, and the panel invites editing it. One sentence — "these values are shaped so SORT orders them correctly; changing them changes the sort order" — prevents that.
4. **`rank` is no longer displayable.** If a newcomer follows the obvious path (bind a text object to `rank` to print the rank on a card) they now get `01`, `2J`, `3Q`, `4K` on their cards instead of `A`, `J`, `Q`, `K`. Consistency with the add-widget deck is the point of this PR so I'm not asking you to reverse it, but the tooltip should say the values are for sorting, not for showing, and the built-in card faces (which bind `image`, not `rank`) should stay the documented example.
5. **The deck tiles carry no visible name** — only a `title` on hover, so on touch there is none, and the name only appears in the details block *below* the grid after selection. Pre-existing, but it is the reason the new description text is so easy to miss; a small caption under each tile would make the picker self-describing and give the description something to sit under.

Everything else checked out: the created deck's cards render correctly at all four viewports (the faces bind `image`, so the changed `rank`/`suit` values are invisible on the table), the Add-widget overlay, JSON editor, States, Players, About and Game Shelf overlays are unaffected at every size and all close cleanly with Escape, and a genuine color row still gets its picker — I confirmed `color: transparent` on the front face object keeps the swatch.

object tab, color: transparent keeps its picker

---

## Draft tutorial

The PR's `pr-test-ai` room holds **"Sorting the standard deck"** — a demo/test room with six sort buttons, not a stepped tutorial (no guided steps, no "next", no end state), so there is no variant walkthrough to play through. I exercised all six buttons and all four viewports. It works and it demonstrates what it claims: `rank` gives A A A A 2 2 2 2 … K K K K with the jokers last, `suitAlt` alternates clubs/hearts/spades/diamonds, and the pressed button keeps a visible active state.

sort by rank
sort by suitAlt

Four UI notes if this ever becomes a real tutorial:

- **The intro is a five-line wall.** It mixes the editor path, the mechanic and a six-item glossary strung together with `·` separators, all in one paragraph at the same weight. Split it: one sentence saying what the room does, then the six keys as an actual list (or as a caption under each button), so the eye can match a button to its explanation instead of parsing a run-on line.
- **Inside a group the ranks stay shuffled**, so `suit`, `suitColor` and `suitAlt` look random at a glance (`10 A 8 7 2 Q 3 4 6 5 J 9 K` within clubs) even though they are correct. Either sort by the chosen key *and then* `rankFixed`, or say in the text that ranks stay shuffled inside each group — otherwise the demo undersells three of its six keys.
- **The "Recall & Shuffle" button sits on an empty white card outline** in the top right with no label, which reads as a stray blank card rather than the deck. A label or a border on that box would explain it.
- **Readable at 820×1180 and up; unusable at 390×844 portrait** (the app's own rotate-your-device state). Fine for a desktop demo — worth knowing if it is ever pitched as a tutorial.

demo room at 1280×800

VTT-AI-Agent:
Thanks — the two big points and two of the three nitpicks are in the branch (c119172).

**1. Explain the sort keys where they are used — done, as per-row tooltips.** `addTypeInput` now asks a new `cardTypePropertyHint()` what a card type property is for and puts the answer on the row and on its label, exactly the way you sketched:

| row | tooltip |
|---|---|
| `suit` | Sorting property: groups the cards of one suit together. |
| `suitColor` | Sorting property: keeps the suits of the same color together. Despite its name it holds no CSS color here. |
| `suitAlt` | Sorting property: a second suit order - the standard deck alternates black and red suits with it. |
| `rank` | Sorting property: orders the cards by rank, ace low. Shaped so that sorting gets it right, not for printing on a card. |
| `rankA` | Sorting property: like rank, but with the ace high. |
| `rankFixed` | Sorting property: one fixed order for the whole deck - by rank, then by suit. |

Two things I did differently from the sketch. The hints only appear on a card type that carries the whole set (it needs a `suitAlt`, `rankA` or `rankFixed`), because the German and Spanish decks use `rank`/`suit` for their own *readable* values — telling their author that `rank: "A"` is "shaped so that sorting gets it right" would be a lie. And `suitColor` gets a different sentence when its value really is a color (the tarot deck's `suitColor: "purple"`, and every deck the custom-deck wizard builds): *"The color of this card's suit. Sorting by it keeps the suits of the same color together."* That is the same collision you describe under Readability #1, answered per value instead of per name.

**2. The picker description is trimmed** to your wording, and it is a `description` of the deck again: *"French-suited 52-card deck with 2 jokers. The cards carry extra properties so a hand or a pile can be sorted by suit or by rank."* Measured in the same place you did: the details block goes 90px at 1280×800 (was 109) and 156px at 390 (was 194) — hanafuda's is 79/129, so it is back in the same shape as its siblings. The six names now live in the tooltips.

**3. Row width jumps mid-edit — fixed.** A color-named row keeps `hasColorPicker` (the 48px padding) whether or not it is currently showing a button, so the field width no longer changes when the value stops or starts looking like a color. Measured on `suitColor` at 1280×800: 206px with `♠`, with `Spades`, with `#f`, with `#ff0000` — one width throughout. The consequence is the one you already noted under "right edges no longer line up": that row is permanently 24px shorter than `suit`/`rank` above and below it. I took your trade (stable under the cursor beats aligned when idle); say the word if you'd rather have the alignment back.

**4. A letters-only value still arms the picker — fixed, differently than suggested.** Rather than list the CSS color keywords to test prefixes against (`isColorValue` deliberately hands letters to the browser's own parser instead of carrying a 148-name list), the half-typed grace now only *keeps* a picker that is already showing: `shouldOfferColorPicker(property, value, offeredBefore)`. A row that starts without one only gets it from a value that really parses as a color, so typing `S`, `Sp`, `Spades` into `suitColor: "♠"` never arms it, while retyping `#ff0000` or `red` over an existing color keeps the button (and an open picker) stable between keystrokes. Verified in the browser and covered in `tests/client/deckeditor-color-picker.test.js`.

**5. 390px deck-editor clipping** — agreed it is pre-existing (thanks for checking it against the Spanish deck), so I left it alone; it is a deck-editor responsive fix, not this PR's.

**Demo room — reworked and re-published** to the PR's `pr-test-ai` room ([Sorting the standard deck.vtt](https://agent.virtualtabletop.io/reports/pr/1538991387469545602/Sorting%20the%20standard%20deck.vtt), [screenshot](https://agent.virtualtabletop.io/reports/pr/1538991387469545602/demo-rankFixed.png)):

- The five-line intro is three lines and no longer carries the glossary; each button has a caption under it instead (`suit` → "groups the four suits, jokers last", `suitAlt` → "alternates black and red suits", …), so a button and its explanation are next to each other.
- **Every sort now ends with a tie-breaker** (`SORT` takes an array of keys): `suit`/`suitColor`/`suitAlt` sort by the key and then by `rankFixed`, `rank`/`rankA` by the key and then by `suit`. Inside a group the ranks are in order now, so the three suit keys actually show what they do instead of looking random. Walked all six buttons in Chromium and read the dealt order back out of the DOM.
- The blank white box has a **"The deck"** label above it, so the *Recall & Shuffle* button is visibly attached to the deck holder.
- Still a demo room, not a stepped tutorial — and still not in `library/tutorials/`; happy to author it as a real tutorial if a maintainer wants one.

`npm test` is at 1279, and the whole TestCafe `editor.js` suite (45 tests) passes.

VTT-AI-Agent (commented):
## TL;DR

The standard-deck data and its parity tests look correct, and the previous review's UI feedback is mostly addressed. Two editor-state bugs remain: a non-color word can still retain an active color picker after a realistic edit sequence, and partial/custom sort-key sets receive confidently incorrect standard-deck tooltips; the external room also still needs to meet and document the repository's tutorial-quality requirements.

## Findings

### 1. The color picker can still remain armed on a non-color value

`client/js/editor/deckeditor.js:3233-3252, 3356-3363`

`isPartialColorValue()` treats every letters-only string as indefinitely partial, and `updateSwatch()` feeds the previous `offer` state back into `shouldOfferColorPicker()`. Consequently, a row whose picker is already visible never loses it when the user replaces a valid color with an arbitrary word: starting from `suitColor: "red"`, the input sequence `S` → `Sp` → `Spades` returns `true` at every step. The same happens in the standard deck if the user first clears `♠`: the empty value arms the picker, then `Spades` keeps it. At that point the palette can still overwrite the non-color sort key, which is the destructive trap this change is intended to remove.

The new test at `tests/client/deckeditor-color-picker.test.js:43-48` misses this because it passes `offeredBefore = false` independently for every value instead of carrying the state between calls. Please add a stateful transition test and make the grace period terminate—for example, retain letters only while they are a prefix of a CSS color keyword, or allow partial text while editing but revalidate with `offeredBefore = false` on change/blur.

### 2. One custom marker property is enough to enable all standard-deck claims

`client/js/editor/deckeditor.js:3255-3275`

The comment says only a card type carrying the "whole set" is described, but line 3270 uses `.some()`. A custom card type such as `{ rank: "high", rankFixed: "custom" }` is therefore told that `rank` orders cards ace-low, and a type with only `suitAlt` is told it follows the standard deck's black/red convention. Card-type properties are author-defined, so those statements are not generally true. Gate these hints on the complete six-property signature (or another unambiguous standard-deck signature), and add a partial/custom-set test alongside the Spanish-deck case.

### 3. The test room is off-library as required, but tutorial quality is not yet established

The PR correctly links a `.vtt` in `pr-test-ai` and does not touch `library/`. However, it still describes the room as **"Sorting the standard deck"** and as a demo rather than using the documented tutorial category naming, and the PR body does not identify the newest tutorial used as its template or confirm field-for-field `_meta.info` parity and the required embedded square library image. The standing rule applies to comparable new user-visible features, not only schema-level widget properties. Please bring the room to that standard (for example, a `Properties - Standard deck sorting` room with the canonical frozen `title`/`overview`/caption structure, semantic IDs, matching metadata, and preview image), validate and walk it end-to-end, then record those checks in the PR body; keep it out of `library/` unless a human explicitly approves it.

## Other checks

- The six values on all 54 template cards match the existing generated standard-deck convention, including both jokers, and the new data tests cover the intended sort orders.
- No new widget property, routine operation, or routine parameter was added. The card-type properties are editable in the real deck-editor UI, so no `jeCommands`, validator, or visual routine-block addition is required.
- The changed-file list contains no engine path (`client/js/widgets/`, `client/js/compute.js`, or `client/js/serverstate.js`), no library file is changed, and the current CI/CodeQL checks are green.

The `beta` label was already present; this PR touches no engine file, and the remaining issues are editor/test-room quality problems rather than server-crash, game-corruption, or high-severity security risks.

VTT-AI-Agent:
Thanks — all three findings are addressed in 7142530.

**1. The picker could stay armed on a non-color word — fixed, by dropping the state instead of bounding it.** You are right about the loop: `updateSwatch()` fed its own answer back into `shouldOfferColorPicker()`, and `isPartialColorValue()` accepted any letters-only string, so `red` → `S` → `Spades` (and `♠` → *cleared* → `Spades`) kept the palette on a value that will never be a color.

Rather than time-boxing the grace period, `shouldOfferColorPicker(property, value)` is now stateless — no `offeredBefore` parameter at all, so there is no state left to get stuck: a color, an empty color-named field, or a half-typed `#…` / `rgb(` / `hsl(` value. Those three shapes can't be anything but a color on its way in, so the button still doesn't flicker while a hex value is retyped, and no sequence of keystrokes can end with a picker on `Spades`. Typing a keyword (`b` → `bl` → `blu` → `blue`) now shows the button only on the last keystroke; I took that over carrying a 148-name CSS keyword list purely to guess at prefixes, and the field width no longer changes either way, so nothing moves under the cursor.

Verified keystroke by keystroke in Chromium on a freshly added Traditional standard deck, reading the button out of the DOM after each `input` event:

```
"♠" no picker | "" PICKER | "S" no | "Sp" no | "Spades" no | "♠" no
"red" PICKER  | "S" no    | "Sp" no | "Spades" no
"" PICKER | "#" PICKER | "#f" PICKER | "#ff0000" PICKER | "rgb(0, 0" PICKER | "rgb(0, 0, 255)" PICKER | "♠" no
```

The field measures 206px throughout. `tests/client/deckeditor-color-picker.test.js` now walks exactly those sequences through one row instead of testing values independently, and the TestCafe assertion that a property added with type `color` gets its picker still passes.

**2. One marker property was enough — fixed.** `cardTypePropertyHint()` gates on the complete six-property signature (`Object.keys(hints).every(...)`), so `{ rank: "high", rankFixed: "custom" }` and a type carrying only `suitAlt` get nothing, and so does the tarot deck (`suitColor: "purple"`, no `suitAlt`/`rankA`/`rankFixed`). Covered by a new partial/custom-set test alongside the Spanish-deck one, and re-checked in the browser: the Spanish deck's `rank`/`suit`/`suitColor` rows carry no tooltip, the standard deck's six rows all do.

**3. Test room rebuilt to tutorial quality.** It is now **Properties - Card sorting**, one game with two variants in the documented category scheme, still on the test server only — the diff touches nothing under `library/`:

- *Suits* (`0.json`): `suit`, `suitColor`, `suitAlt`. *Ranks* (`1.json`): `rank`, `rankA`, `rankFixed`. Each button shuffles the whole deck, sorts by its property plus a tie-breaker and deals into a 13-wide grid, so a suit sort puts exactly one suit per row and the jokers on the last row.
- Metadata template: `library/tutorials/Properties - dragLimit` (newest by `_meta.info.lastUpdate`), field-for-field parity — `mode: "Tutorial"`, `players: "1"`, `time: "0"`, `year: "2026"`, `language: "en-US"`, `showName: false`, `usesAIImagery: false`, the `similar*`/`rules`/`bgg`/`helpText` fields empty, one shared `name` + `image` with `variant` `"Suits"`/`"Ranks"`.
- Band layout on 1600x1000 with semantic ids (`title`, `overview`, `text1..4`, `button1..3`, `cell-<row>-<col>`, `deckHolder…`), captions under their buttons, explanatory text on `layer: -3` and unmovable, the conventional 111x40 *Recall & Shuffle* button parented to the deck holder, no invented navigation.
- Library image: a 530x530 square SVG that reuses the template image's square badge, `Properties` category word and graduate cap, embedded in the `.vtt` under `assets/<crc32>_<length>`.
- `node validator/validate_gamefile_node.js` exits 0 on both variants; both walked end to end in Chromium (all six buttons, dealt order read back from the DOM) and the `.vtt` re-uploaded into a `pr-test-ai` room to confirm it loads with its preview image.

[Properties - Card sorting.vtt](https://agent.virtualtabletop.io/reports/pr/1538991387469545602/Properties%20-%20Card%20sorting.vtt) · [Suits](https://agent.virtualtabletop.io/reports/pr/1538991387469545602/card-sorting-suits.png) · [Ranks](https://agent.virtualtabletop.io/reports/pr/1538991387469545602/card-sorting-ranks.png). Whether it should become a real tutorial under `library/tutorials/` is a maintainer's call — say the word and I will move it into the PR.

`npm test` is at 1280 tests and the whole TestCafe `editor.js` suite (45 tests) passes.

96LawDawg (approved):
Tested.  Approved. This is a minor change to make the deck editor's standard deck have the same sort properties as the regular standard deck added through add widget overlay. It also adds a few minor explanatory notes to the deck editor about the properties.  Going to merge.

PR #3126, opened 2026-08-18
Give the deck wizard a card-sheet mode and size cards from the picture

VTT-AI-Agent:
The deck editor's **Add New Deck** wizard can turn one picture that holds a whole grid of cards into a deck, but that mode was hidden behind the label *"One image per card"* — which reads like the opposite — and the cards it produced were squashed. This PR makes the two picture modes unambiguous, derives the card size from the picture, lets a sheet of fronts be paired with a matching sheet of backs, and stops the Tabletop Simulator importer from writing an empty extra back-face object.

Prompted by user feedback in Discord #general on 2026-08-18 ([source](https://discord.com/channels/770758631146782780/793626590848352306/1539341014626734081)):

> **lesko:** is the function to upload one image with multiple cards on it and then chop it up available somewhere still? I can't seem to find it
>
> **lesko:** oh it's under "one image per card" 😅
>
> **ArnoldSmith86:** that sounds like it could be improved..
>
> **lesko:** any idea why some cards look squished like that?
>
> **LawDawg:** Probably just the size of the cardDefault is not lined up with the image size (ratio-wise anyway)
>
> **LawDawg:** you can just change the cardDefault to something that works. I did 300x300. Not perfect but better.
>
> **lesko:** weird, I see it as one row in the deck builder
>
> **lesko:** what might be a bug though, the imported decks often have this transparent additional (back)face
>
> **lesko:** I always delete them manually
>
> **lesko:** I said often but meant always, they always have it

## 1. The sheet-splitting mode is its own mode now

*Create a custom deck* used to list *"One image per card"* with the sheet case buried in its description (*"…or tiled images"*). It is now two separate modes, each with a one-line description and a small drawing of the input it expects (one card per file vs. one sheet with dashed cut lines):

* **One picture per card** — upload a separate image file for every card, one shared back for the whole deck
* **One picture with all the cards on it** — upload a single sheet holding a grid of cards and cut it into one card per grid cell

Both run the same flow (`deckImages(target, sheetMode)`); the row/column controls only exist in the sheet mode, so "one picture per card" no longer shows the confusing *"Rows (if multiple cards)"* fields, and the sheet mode leads with what it wants. The mode values are UI state only — nothing in a saved game refers to them.

| before | after |
| --- | --- |
| before | after |

## 2. Squished cards: the size came from the wizard's own preview, not from the picture

`deckImages` read the card ratio off the preview `` element:

```js
let cardWidth = Math.round($('.cardFrontPreview img', preview).width / $('.cardFrontPreview img', preview).height * 160);
```

`.cardFrontPreview` is `display: flex`, so that `img` was stretched to the row height and `img.width/img.height` was **the shape of the preview box, roughly 1.67, whatever the file looked like**. A 1500 × 400 sheet of 5 × 2 landscape cards came out as `cardDefaults.width: 114.8` with no height, i.e. 115 × 160 portrait cards each showing a 300 × 200 picture stretched into them. That is exactly what LawDawg diagnosed.

Now the size comes from the file: `cardSizeFromImage()` takes the natural pixel size, divides it by the grid, and scales the longer side of one cell to 160 — the same rule `server/ttsimport.mjs` already uses, so both import paths size cards identically. The example above becomes 160 × 107 (`{"width": 160, "height": 107}`, height written explicitly instead of leaving the 160 default). One image per card gets the same treatment.

| before (115 × 160) | after (160 × 107) |
| --- | --- |
| before | after |

The same stretching is why *"I see it as one row in the deck builder"*: the preview of a 10 × 2 sheet was squeezed to the height of the settings column, so only the top row was really visible. The preview keeps its aspect ratio now, draws the cut lines the row/column numbers describe straight onto it, and says what will be created (`20 cards of 200 × 300 pixels, 107 × 160 on the table`) — so a wrong grid is visible before anything is added. The row/column sliders also read their number input, so a sheet with more than the slider maximum can be typed in.

| before | after |
| --- | --- |
| before | after |

Two uploads can also carry the same file name (the same file twice, or two files of that name from different folders); the card types of the second one are named apart with a `(2)` suffix instead of overwriting the first one's.

One import can hold several pictures, and they do not have to be the same shape, so a single card size for the whole deck is not enough: the first upload sets the deck's `cardDefaults` and every card type cut out of a differently shaped upload carries its own `width`/`height` on top of them (a card type wins over the deck defaults in `card.js`). Uploading a 1500 × 400 sheet cut 5 × 2 and a 400 × 1200 sheet cut 2 × 4 into one deck gives 160 × 107 landscape cards and 107 × 160 portrait ones, each with the shape of its own cell:

two sheets, two shapes

The number next to "Add to game" and the per-upload summary count the copies field as well, so they say how many cards will really be created, and the preview is bounded in height as well as in width so the cut lines drawn on a tall sheet stay inside the dialog.

Two smaller bugs in the same handler are gone with it: `hasTiledImage` searched the whole document for `.rows input, .cols input` instead of the wizard's own uploads, and the fractional width it produced (`114.8`) is now a rounded value. Setting `cardDefaults.css` also used to throw when the computed width happened to be exactly 103, because `cardDefaults` was only created for other values.

## 3. TTS decks no longer get an extra transparent back face

This one is an importer bug, not TTS data. `addDeck()` in `server/ttsimport.mjs` always wrote **two** image objects onto the back face — one reading the card property `back`, one reading `simpleBack`. A card type only ever carries one of them: `back` when the CustomDeck entry has `UniqueBack` (a sheet of individual backs), `simpleBack` when it does not (one image for the whole deck). So every imported deck ended up with one image object that no card ever fills, which renders as an empty transparent layer — the one lesko deletes by hand.

Verified with the workshop item from the report, https://steamcommunity.com/sharedfiles/filedetails/?id=2939146975: all 16 decks it contains had two objects on the back face before, and have exactly one after (10 of them `back`, 6 `simpleBack`).

| before: `Face 0 (back)` has an empty object 1 | after |
| --- | --- |
| before | after |

A deck whose cards have no back image at all still gets the single transparent object, so it keeps a back face to flip to, and a deck that legitimately has per-card backs keeps the sheet-based object it needs. Existing saved games are untouched — only newly imported decks are affected, so no `fileupdater` migration is involved.

## 4. A sheet of fronts plus a matching sheet of backs

Follow-up on the same PR, from Discord ([source](https://discord.com/channels/770758631146782780/1539350642760360088/1539373654331236362)):

> **arnoldsmith86:** If there's no way to create a deck from a tiled sheet of front faces and a separate, matching tiled sheet of back faces, add it.

There was not: the sheet mode gave the whole deck one shared back, and the only per-card backs the wizard could produce came from *"A front and a back image per card"*, which needs one file per card. The sheet mode's **Card backs** section now asks where the backs come from:

* **The same back on every card** — the gallery and the single upload, as before
* **A second sheet with one back per card, cut with the same grid** — upload a second sheet; both are read with the same offsets, so a card's back is the cell its front came from

The second sheet only has to share the *grid*, not the resolution — the face scales whatever it gets to the card. Several sheets of fronts are paired with several sheets of backs in upload order, one each; the status line next to "Add to game" says what will be created and keeps the button off while the counts don't match. Each uploaded sheet of backs shows the grid it will be cut with and which sheet of fronts it belongs to.

A sheet of fronts and its sheet of backs also leave together: deleting a sheet of fronts deletes the sheet of backs paired with it, so the remaining backs cannot silently shift onto the wrong fronts.

The resulting deck is the same shape the Tabletop Simulator importer produces for a `UniqueBack` deck: every card type gets `backImage` next to `image` plus the shared `offsetX`/`offsetY`/`deckWidth`/`deckHeight`, and the back face has exactly one object reading `backImage` through the same tiling CSS. `backImage` is the property the front/back-pairs mode already uses, so nothing new enters the validator or `jeCommands`.

the wizard

Cut out of a 1500 × 400 sheet of fronts and a 750 × 200 sheet of backs — the top row face up, the bottom row of the same deck face down:

the cards

## 5. Reading the preview, and the deck browser at readable size

From the UI/UX review. The preview exists so the grid can be checked against the picture, and it was the first thing to break: bounded to half the panel and 220 px tall, an ordinary 2 × 10 print-and-play sheet came out as a 29 px sliver, and below ~500 px the picture and its settings both collapsed.

* A sheet taller than it is wide takes the whole panel with its settings underneath; below 500 px every upload does. Every upload's settings now start at the same edge, so two sheets read as two rows of one form.
* **Clicking a picture opens it at window size** with the same cut lines on it — the only size at which a 2 × 10 grid can really be checked. Click or Escape closes it again.
* The cut lines are dashed and 2 px (the way the mode's icon draws them), and a sheet gets its frame from the start rather than only once it is cut.
* **Add to game** stays off, with the reason in the status line, for a sheet still on its 1 × 1 default ("Say how many cards this sheet holds across and down.") and for an import of 0 copies. The status takes a row of its own instead of being squeezed into a 40 px column of single words.
* The **Upload sheet of backs…** button moved into the action bar next to **Upload sheet of fronts…**, so the red line asking for a sheet of backs no longer points straight at the fronts button (and the backs button is no longer half-hidden behind the sticky bar on a tablet).
* Card types are named from one — `fronts.png 1,1` rather than `fronts.png 0,0` — matching the counting the dialog just asked for. The front/back mode icon is hatched so it no longer reads like the *Text cards* icon, and the wizard says *picture* throughout instead of alternating with *image*.

the action bar

a picture at full size

The public library deck browser is the one overlay that is a screenful of text and preview tiles rather than something drawn on the table, and it was scaled with the board: at a board scale of 0.5 — an ordinary laptop — the filter field, the sort control and every deck name rendered at half size, at 0.22 on a phone they were unusable. It gets the editor's box (below the toolbar, beside the sidebar) in **both** places it can be opened from now, which also removes the borrowing of the node between `#roomArea` and `#editor` this PR introduced. It has to live in `#editor`: `#roomArea` is a size container and therefore the containing block a fixed child would be clipped by.

the browser from plain edit mode

## 6. The object list showed blank cards until a face was opened

Follow-up on the same PR, from Discord ([source](https://discord.com/channels/770758631146782780/793626590848352306/1539514991063867423)):

> **lesko:** the bug is gone 👍
>
> **lesko:** a thing I forgot to mention yesterday which I am not sure counts as a bug: when you open the decks, the cards looks blank, only after opening the faces does the art appear

The list of face objects on the left of the deck editor previews each object by cloning it out of the rendered card, so it stays in step with every edit. A deck opens on its **card defaults**, though, and that view deliberately shows an explanation instead of a card — `this.mainCard` is `null` there, nothing gets cloned, and every image object came out as the empty checkerboard the preview box draws when it has no content. Clicking a face rendered the card and filled the previews in, which is exactly the "only after opening the faces does the art appear" in the report.

Those views now render one card off-screen (`previewSourceCard()`, from the same working templates `refreshMainCardFaces` uses) and clone from that, so the list looks the same whichever view the editor is in.

| before | after |
| --- | --- |
| before | after |

The second half of the same box: a card cut out of a sheet showed the **whole sheet** squeezed into 46 × 62 pixels. The preview refitted the picture with `background-size: contain`, throwing away the `background-size`/`background-position` that pick the card's own cell out of the sheet — the tiling CSS both this wizard and the Tabletop Simulator import write. Those are `calc()`s over the card's CSS variables, which a cloned object no longer inherits, so the resolved values are read off the rendered object and the object is scaled down as a whole. Objects that are not tiled keep the `contain` fit they had.

| before: the whole sheet | after: this card's cell |
| --- | --- |
| before | after |

## 7. "403 - Forbidden" when a picture was over the upload limit

Reported on this PR by **96LawDawg**:

> Uploading a 15 MB png file got a warning that it was "forbidden." I compressed the file to a 2 MB jpg and it uploaded fine. If the forbidden warning is because of file size, the warning needs to be more informative.

That is exactly what it was. `PUT /asset` is parsed with `express.raw({ limit: '10mb' })`, and when the body is over that limit the parser throws `request entity too large` — which `Logging.userErrorHandler` answered with **403 "Too big."**. The client's upload does check for the too-big case, but it checks for the correct status:

```js
if(response.status == 413)
  throw 'File is too big.';
else if(!response.ok)
  throw `${response.status} - ${response.statusText}`;
```

So the 413 branch never ran, the generic one did, and the alert read `Uploading failed: 403 - Forbidden` — no mention of size, nothing to act on. The deck wizard is where this shows up most, because a whole sheet of cards is the biggest file anyone uploads here.

The handler now answers **413** and says how big the file was and how big it may be, using the size and limit the parser error already carries (so it stays right for the 500 MiB save-file routes too):

```
The file is 14.3 MiB - the limit is 10 MiB.
```

The client passes that text on and appends what to do about it, so the alert reads:

> Uploading failed: The file is 14.3 MiB - the limit is 10 MiB. Scaling a picture down or saving it as a JPEG usually gets it under the limit.

One thing came with it: a file that fails to upload no longer reaches the callback with a `null` asset path. In the wizard that added a blank picture tile for a file that never arrived, on top of the alert.

## 8. Sorting the deck browser by stars or popularity looked broken

Reported on this PR by **96LawDawg**:

> The sort feature on the public library decks does not work.

The sort works — it had nothing to sort by. *Stars* and *Popularity* order the game groups by each game's star count and recorded play time, and both are per-server statistics: virtualtabletop.io has stars for 288 of the 291 games in the deck catalog and play time for all 291, this PR's test server has neither for any of them. Every game ties at zero there, the tie-break is the name, and all three sort modes produce the same list.

The browser now says so instead of leaving a control that appears dead — *"No game on this server has been starred yet, so this is the order by name."*, and the same for play time. The line disappears as soon as the server has counted anything, and for *Name*, which is that order anyway.

the deck browser with nothing to sort by

## Demo room

A playable demo game is on this PR's test server (`pr-test-ai`), **not** in `library/`: [Card sheet import - Demo.vtt](https://agent.virtualtabletop.io/reports/pr/1539350642760360088/Card%20sheet%20import%20-%20Demo.vtt). It has three variants: *1) Cutting up a sheet* shows the uploaded picture, the deck the wizard builds from it now and the same deck at the size it used to get; *2) Fronts and backs* shows a deck cut from a sheet of fronts and a matching sheet of backs (click a card to turn it over) next to a Tabletop Simulator deck whose back face has a single object; *3) Two shapes in one import* shows a deck made of a landscape sheet and a portrait sheet, next to the same four cards as they came out before, all in the shape of the first sheet. It is authored to tutorial quality (`_meta` mirroring `Properties - dragLimit`, square SVG library image) and can be added to `library/tutorials/` on request — no tutorial or library file is part of this diff.

The layout notes from the UI/UX review are in: the three columns of variant 1 share one caption baseline and carry a bold label each, every closing note sits under a full-width divider and names the other variants and the route to them (Game Shelf, top left), the two decks of variant 2 share a baseline and both *Recall & Shuffle* buttons are the same width, and both sheets of variant 3 are captioned the same way. The on-board titles now match the shelf entries.

demo room, variant 1

demo room, variant 2

demo room, variant 3

## Testing

* `npm test` — 1308 passed, including four tests pinning the back face of an imported Tabletop Simulator deck down (one object for the kind of back its cards carry, both objects when a deck uses both kinds, and never a back face without any object) and four new ones on the error handler (an oversized upload of known and of unknown length, a UserError keeping its own code, anything else passed on).
* `npx testcafe chromium:headless tests/testcafe/editor.js` — 68 passed, including the wizard tests for text cards, image pairs and uploaded fronts plus new ones covering the sheet of fronts with a matching sheet of backs (and the guards that keep **Add to game** off for an uncut sheet and for 0 copies), the public library browser above the deck editor, the browser at unscaled size from plain edit mode, two sheets of different card shapes in one import, two uploads carrying the same file name, the object list showing a sheet card's own cell before a face is selected, a sheet of fronts being deleted together with the sheet of backs paired with it (both fail on the commit before their fix), and the deck browser naming the sort it has no data for.
* `node validator/validate_gamefile_node.js` on the demo game file — clean.
* Measured and walked the changed screens at 1280 × 800 and 390 × 844 after the UI review: the preview of a tall (2 × 10) and a wide (5 × 2) sheet, two uploads in one import, the full-size picture view, back-sheet mode, and the deck browser from both entry points.
* Uploaded a 15,000,000-byte picture through the wizard in a real browser: the alert reads *"Uploading failed: The file is 14.3 MiB - the limit is 10 MiB. Scaling a picture down or saving it as a JPEG usually gets it under the limit."* and no empty picture tile is added. A 5 KB upload still answers 200 with its asset path.
* Walked every import path in a browser end to end: a 5 × 2 grid sheet of 300 × 200 landscape cards through the new mode, the same sheet together with a matching sheet of backs (flipping a card shows the back from its own cell), a landscape and a portrait sheet in one import, one picture per card, and the workshop link above through the Tabletop Simulator mode. Also browsed the public library from both entry points (the Add Widget overlay and the deck editor), picked a deck, and cancelled with Escape and with the close button.
* Rebased onto `main` after the write face object type and the shared widget-selection bar landed there: the whole jest suite and the whole `editor.js` TestCafe suite (both the tests above and the new ones on main) were re-run on the rebased branch, and the sheet mode, the tree previews and the deck browser were walked again in a browser.

No widget property or routine function is added or renamed — the sheet of backs writes the existing `backImage` card-type property — so there is nothing to add to `validator/` or to `jeCommands`.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3126/pr-test (or any other room on that server)


96LawDawg:
The add deck option to browse from the public library is broken. It is also broken on production, but let's fix it here.

VTT-AI-Agent:
Added in [d3fd61d](https://github.com/ArnoldSmith86/virtualtabletop/commit/d3fd61d6), on top of the sheet mode this PR introduces — from Discord ([source](https://discord.com/channels/770758631146782780/1539350642760360088/1539373654331236362)):

> **arnoldsmith86:** If there's no way to create a deck from a tiled sheet of front faces and a separate, matching tiled sheet of back faces, add it.

There was not, so the sheet mode's **Card backs** section now asks where the backs come from: **the same back on every card** (the gallery and the single upload, as before) or **a second sheet with one back per card, cut with the same grid**. Both sheets are read with the same offsets, so a card's back is the cell its front came from; only the grid has to match, not the resolution. Several sheets of fronts pair with several sheets of backs in upload order, and the status line next to "Add to game" says what will be created and keeps the button off while the counts don't match.

The deck it writes is the shape the TTS importer produces for a `UniqueBack` deck — `backImage` next to `image` on every card type and a single back-face object reading it through the tiling CSS — so no new property enters the validator or `jeCommands`. Covered by a new TestCafe test (`Deck editor: a sheet of fronts with a matching sheet of backs in the new deck wizard`); the full `editor.js` suite is at 46 passed, `npm test` at 1280.

The demo game on the test server now has a second variant, *2) Fronts and backs*, where clicking a card turns it over to the back cut from its own cell — plus the TTS deck with the single back-face object next to it: <https://test.virtualtabletop.io/PR-3126/pr-test-ai>

the wizard

the cards

VTT-AI-Agent:
Fixed in [7f596c3](https://github.com/ArnoldSmith86/virtualtabletop/commit/7f596c35) — thanks, that one is a pure CSS stacking bug and it does predate this PR.

**What was wrong.** `DeckEditor.initializeDOM()` moves `#libraryDecksOverlay` into `#editor` so it shows above the full-screen deck editor. The other overlays moved along with it (`#deckEditorExportOverlay`, `#deckEditorNewDeckOverlay`, the symbol/audio pickers) each get an explicit box there — `position: fixed`, inset to the editor area, `z-index: 20`/`30`. The library browser never got one, so it kept what `editmode.css` gives it for `#roomArea`: `position: absolute` (from `.overlay`), sized to `--roomWidth`/`--roomHeight` and `transform: scale(var(--scale))`, with no `z-index`. `#deckEditor` has `z-index: 1`, so it painted straight over the browser — the overlay was open (its DOM was there, the decks had loaded) but invisible, and a click on a deck was swallowed by `#deckEditorTree`.

That is also why it is broken in production, and it is not limited to the deck editor: the move is permanent, so after the deck editor has been opened once, "Browse the public library" from the Add Widget overlay in plain edit mode was covered by the room surface in the same way. Before the deck editor is opened for the first time the overlay is still in `#roomArea` and works, which is presumably why it slipped through.

**The fix** gives `#editor > #libraryDecksOverlay` the box its neighbours already have (fixed, inset to the editor area, `transform: none`, `z-index: 20`).

While in there: closing the browser without picking a deck used to leave the *Add New Deck* dialog hidden and `pendingNewDeck` armed, dropping you into an empty deck editor. `openLibraryDecksOverlay()` now takes an `onCancel` from whoever opened it, and the deck editor uses it to bring its dialog back on the section it was opened from.

Covered by a new TestCafe test (`Deck editor: the public library deck browser opens above the deck editor`) that checks a deck entry is really the element a click at its position lands on, and that cancelling returns to the dialog — it fails on the code before this commit. Full `editor.js` suite: 47 passed; `npm test`: 1283 passed (the `minify` suite only passes here when it is not competing with TestCafe for the sandbox's CPU).

| before | after |
| --- | --- |
| before | after |

Walked it end to end afterwards: browse → filter → click a deck → the deck, its holder and its cards are added and the deck editor opens on it.

VTT-AI-Agent (commented):
## TL;DR

The mode split, the grid preview and the TTS back-face fix all look right, and I reproduced the library-browser stacking fix working from both entry points. One real bug survives: the card size is still taken from the **first** upload only, so as soon as a second sheet (or a second picture) has a different aspect ratio, its cards get squashed again — while the new per-sheet summary explicitly promises each sheet its own on-table size.

## Findings

### 1. `cardDefaults` is derived from the first upload only — a second sheet with a different ratio is squashed (the bug this PR fixes, one upload later)

`client/js/editor/sidebar/properties.js:2813-2816`:

```js
const firstImage = $('img', sheets[0].dom);
deck.cardDefaults = cardSizeFromImage(firstImage.naturalWidth, firstImage.naturalHeight, sheets[0].columns, sheets[0].rows);
```

`cardDefaults` is the only place a size is written; no card type gets one. So every card of the deck takes the shape of `sheets[0]`, no matter which picture it actually shows.

I reproduced it in a browser against this branch (deck editor → *One picture with all the cards on it*, two sheets):

* sheet A `1500 × 400`, cut `5 × 2` → summary: `10 cards of 300 × 200 pixels, 160 × 107 on the table`
* sheet B `400 × 1200`, cut `2 × 4` → summary: `8 cards of 200 × 300 pixels, 107 × 160 on the table`

Result after *Add to game*: 18 card types, `cardDefaults: { width: 160, height: 107 }`. The 8 portrait cards from sheet B are rendered 160 × 107 — a `200 × 300` picture stretched into a landscape card, i.e. exactly the squashing described in section 2 of the PR description. And the wizard told the user, in the row right above the button, that those cards would be `107 × 160`.

The same applies to *One picture per card* with mixed-ratio uploads. That mode was equally broken before, so it is not a regression — but the new per-image summary now makes a promise the import does not keep, which is arguably worse than the old silence.

`client/js/widgets/card.js:45-46` already merges `cardTypes[cardType]` over `cardDefaults`, so the fix is local: keep `cardDefaults` from the first sheet and, for every sheet whose `cardSizeFromImage` differs, write `width`/`height` onto its card types:

```js
const defaults = cardSizeFromImage(...sheets[0]...);
// per sheet, inside the loop:
const size = cardSizeFromImage(img.naturalWidth, img.naturalHeight, columns, rows);
if(size.width != defaults.width || size.height != defaults.height)
  Object.assign(cardTypes[cardType], size);
```

That also keeps the summary honest without changing it.

### 2. An image with no decoded size produces a card size derived from the grid, not from the picture

`cardSizeFromImage` (properties.js:96-101) falls back with `imageWidth || 1`. When `naturalWidth` is still 0 at click time (broken/undecodable asset, or a click before `onload` for a slow upload), a `5 × 2` grid yields cell `0.2 × 0.5` → `64 × 160`, i.e. a ratio invented from the row/column numbers.

`updateSummary` already special-cases this (`image.naturalWidth ? … : ''`, properties.js:2703-2705) — the add path does not. Either keep `addButton` disabled while `!image.naturalWidth` (there is already an `onload` to re-enable on), or fall back to the previous `103 × 160` default rather than to the grid.

### 3. Nothing caps the preview height, so a tall sheet pushes the dialog well past the viewport

`client/css/editor/sidebarModules.css:326-331` gives `.cardFrontPreviewImage` `max-width: 50%` and the image `height: auto`. A portrait sheet — `2 × 10` is a normal shape for print-and-play — renders at ~900 px tall in the wizard; with a 400 × 1200 test sheet the second preview already ran off the bottom of the dialog for me, so the cut lines it is meant to let you check are not all visible. A `max-height` on `.tune.editorModule .cardFrontPreview img` (with `width: auto`) would bound both dimensions; the `inset: 0` grid overlay still lines up because the wrapper shrink-wraps the image.

### 4. The counts in the summary and the status line ignore the copies field

`updateSummary` / `updateStatus` (properties.js:2701, 2673) count grid cells: a `5 × 2` sheet with *Copies of every card: 3* still says `10 cards` and `10 cards, all sharing the back picked above`, while 30 card widgets are created — and `addDeckWithCards` uses the copies for its >500 confirmation, so the two numbers the user sees disagree. Multiplying by the `.cards` number input would make the status line say what will actually land on the table.

### 5. Only the close button routes through the new `onCancel`

`client/js/editmode.js:1760-1767` runs `libraryDecksCancelled` from `#libraryDecksClose`. Any other dismissal still leaves the deck editor with no dialog and `pendingNewDeck` armed — in edit mode `Escape` goes to `$('#editorToolbar button[icon=close]')` (`client/js/main.js:947-951`) rather than to the overlay, so it does not hit the new path. `libraryDecksCancelled` is at least re-initialised on every `openLibraryDecksOverlay`, so it cannot leak into a later browse; this is just the cancel path being narrower than "the browser was closed".

### 6. The library browser now looks different depending on whether the deck editor has been opened

Not a defect, just worth being aware of: `DeckEditor.initializeDOM` moves `#libraryDecksOverlay` into `#editor` **permanently**, and the new `#editor > #libraryDecksOverlay` rule only applies there. I checked both states on this branch — before the deck editor is opened, the browser is still the room-scaled overlay in `#roomArea` (entry rect `28,284,151,143`); afterwards it is the fixed editor-inset box (`40,229,212,200`). Both are visible and clickable now, but the same "Browse the public library" button silently changes layout mid-session. Giving the overlay the fixed box unconditionally in `editmode.css` would remove the split.

## Checked, no finding

* **`server/ttsimport.mjs`** — the `back` / `simpleBack` split is right, including the case a single `Deck` object references several `CustomDeck` entries with different `UniqueBack` flags (both objects get pushed, and each card type only fills its own). A deck whose `BackURL` resolves to nothing falls to the transparent `simpleBack` object, so the back face is never object-less. Import-only, no `fileupdater` migration needed.
* **Validator / `jeCommands`** — no new widget property or routine operation. `backImage` is a card-type key, and `validate_gamefile.js:245` declares `cardTypes: 'any'`, so no validator or JSON-editor entry is due.
* **UI support for new author-facing features** — the whole change *is* the real editing UI (the Add New Deck wizard), so there is nothing that only exists in the JSON editor.
* **Demo room** — no `library/` path in the diff, and the demo lives on the PR's own `pr-test-ai` room with two variants, which is where it belongs.

## Verification I ran

`npm test`: 1274 passed, 6 failed — all six in `tests/server/minify.test.js` and all timeouts under sandbox CPU contention, unrelated to this diff. TestCafe on `chromium:headless`: the new `Deck editor: a sheet of fronts with a matching sheet of backs in the new deck wizard` passes, as do the neighbouring `mismatched and shared card backs` and `a few uploaded card fronts …` tests. The multi-sheet size case in finding 1 is not covered by any test, and neither is the new sizing in *One picture per card* — both would be cheap to add next to the test that already exists.

I added the `beta` label: no engine file is touched at all (nothing under `client/js/widgets/`, no `compute.js`, no `serverstate.js`), the server change only affects newly imported TTS decks, and the findings above are wrong-card-shape and layout issues that make it worth testing rather than unsafe to test.

VTT-AI-Agent:
Thanks — all six points are addressed in [d9af56d](https://github.com/ArnoldSmith86/virtualtabletop/commit/d9af56df) and [1e70c07](https://github.com/ArnoldSmith86/virtualtabletop/commit/1e70c078).

**1. `cardDefaults` from the first upload only.** Confirmed and fixed: the first upload still sets `cardDefaults`, and every card type coming from a differently shaped upload now carries its own `width`/`height` on top of them. Your exact repro (5 × 2 landscape sheet + 2 × 4 portrait sheet) now gives `cardDefaults: {width: 160, height: 107}`, the eight portrait card types `{width: 107, height: 160}`, and the cards on the table are 107 × 160 — the summary's promise is kept. The same code path covers *One picture per card* with mixed-ratio uploads (a 1500 × 400 and a 400 × 1200 file give 160 × 43 and 53 × 160 cards).

The wizard's demo game on the test server has a third variant for it, *3) Two shapes in one import* — the same four cards as they come out now next to the same four cards without the per-card-type size:

two shapes in one import

**2. `cardSizeFromImage(0, 0, …)` invents a ratio from the grid.** Fixed in `cardSizeFromImage` itself: without a decoded size there is no ratio to keep, so it returns the standard 103 × 160 card the wizard used before instead of dividing a placeholder by the grid. Keeping the button disabled would have locked out an asset that never decodes, with nothing on screen saying why.

**3. No height cap on the sheet preview.** Fixed: `.cardFrontPreview img` is bounded by `max-height: 220px` with `width: auto`, so a portrait sheet stays inside the dialog with all its cut lines visible. Your 400 × 1200 test sheet now renders 73 × 220 instead of ~900 tall.

**4. Status and summary ignore the copies field.** Fixed: both multiply by it now. A 5 × 2 sheet with *Copies of every card: 3* says `10 cards of 300 × 200 pixels, 160 × 107 on the table, 3 copies each - 30 cards in total` and the status line says `30 cards`, which is the number the >500 confirmation uses. In *One picture per card* the same field is *Cards to add*, and the summary counts it directly (`3 cards of …`).

**5. Only the close button routes through `onCancel`.** Escape now closes the browser rather than what is behind it, and it goes through `#libraryDecksClose`, so the cancel path is the same one: in plain edit mode `window.onkeyup` in `main.js` takes it before "leave edit mode", and in the deck editor `DeckEditor.onKeyDown` takes it before "close the editor" (the deck editor swallows the Escape *keyup* while it is open, so `main.js` never sees it there). Escape in the deck editor used to close the whole editor and leave the browser standing.

**6. The browser looks different depending on whether the deck editor has been opened.** You are right that this was a side effect of the permanent move, so it is gone too: `#libraryDecksOverlay` is borrowed into `#editor` in `open()`/`openEmpty()` and handed back to `#roomArea` in `close()`/`handleStateReplaced()`, the same way `#symbolPickerOverlay` is already passed back and forth. Each context keeps its own layout — the room-scaled overlay next to the Add Widget overlay it is opened from, the editor-inset box above the deck editor — no matter what was opened before.

Nothing changed about the TTS back-face split, and no validator/`jeCommands` entry is due (still no new widget property or routine operation).

**Testing:** `npm test` 1280 passed (33 suites, no failures when jest is not competing with TestCafe for the sandbox CPU). `npx testcafe chromium:headless tests/testcafe/editor.js` — 48 passed, including a new *Deck editor: sheets of different card shapes each keep their own card size*, which covers finding 1 and the copy counts of finding 4 and fails on the code before this push. Walked both wizard modes with two differently shaped uploads in a browser, plus the library browser from both entry points (open → pick a deck, open → Escape, open → close button, before and after the deck editor has been opened).

VTT-AI-Agent (commented):
## TL;DR

Splitting the sheet import into its own mode with a drawing of the input it expects, and telling the user what will be created (`10 cards of 300 × 200 pixels, 160 × 107 on the table`) before anything is added, makes this wizard far more findable and far less of a guess than the old *"One image per card"* — this is a real UX win. The single most impactful thing left is the **preview box**: fixed at `max-width: 50%` / `max-height: 220px`, it renders a normal 2 × 10 print-and-play sheet as a **29 px-wide sliver** and, below ~500 px viewport width, squeezes both the picture (116 px) and its controls (133 px) into unusable columns — so the cut-line check the preview exists for is exactly the thing that breaks first.

This is a **UI/UX review** — layout, visual design, readability and first-time-user clarity only. Correctness/code bugs are covered by another reviewer. I checked out the branch, ran the app locally and clicked through the changed screens with Playwright.

**What I actually exercised** (viewports 390×844 phone, 820×1180 tablet, 1280×800 laptop, 1920×1080 desktop):
Add New Deck dialog (all three groups) · all five *Create a custom deck* modes · *One picture with all the cards on it* end-to-end (upload → grid → back sheet → Add to game) · *One picture per card* · the public-library deck browser from **both** entry points (deck editor's Add New Deck **and** plain edit mode's Add-widget overlay) incl. Escape/close behaviour · Add-widget overlay · Game Shelf / Players / About overlays · edit-mode sidebar modules (properties, history, JSON editor, widgets, settings) · deck editor + its Print/PDF export overlay · and the PR's demo room on the test server. I did not reach the audio picker, share-link, state-save/variant-add or connection-lost overlays; nothing in this diff touches their CSS.

---

## Big layout improvements

### 1. The preview box collapses exactly when the preview matters most
`client/css/editor/sidebarModules.css:330-352` — `.cardFrontPreviewImage { max-width: 50% }` plus `img { max-height: 220px }`.

A 400 × 3000 sheet cut 2 × 10 (an ordinary print-and-play shape, and the one the code comment names) is rendered at a measured **29 × 220 px**. Ten rows of cut lines in 220 px of height and two columns in 29 px of width cannot be checked against anything — and "check the grid against the picture" is the entire justification for drawing the overlay:

tall sheet preview

The old code overflowed the dialog; the new one keeps it inside but makes it unreadable. Suggestion that fits the existing layout: let a portrait/tall picture take the **full panel width with the settings stacked underneath it** (the JS already reads `image.naturalWidth/naturalHeight` in `updateSummary` — one `classList.toggle('cardFrontPreviewTall', naturalHeight > naturalWidth)` is enough to switch the row to `flex-direction: column` and lift `max-height` to something like `min(60vh, 480px)`). A click-to-enlarge on the preview would also work, and would help the wide case on a 1920 screen too, where the dialog is 656 px wide and the picture is still capped at ~300 px while most of the screen is empty.

### 2. Below ~500 px the whole upload row stops working
At 390 × 844 the 50/50 split leaves the picture **116 × 31 px** and the settings column **133 px**: every label wraps to two lines, the number input drops below its slider instead of sitting next to it, and the sliders run to the panel edge. In the same state the sticky action bar squeezes `.imagePairStatus` into a **41 px wide, 98 px tall** seven-line column beside two buttons whose labels also wrap to three lines each.

| phone: sheet mode | phone: the upload row alone |
| --- | --- |
| phone sheet mode | phone upload row |

Two small rules fix both: `.tune.editorModule .cardFrontPreview { flex-wrap: wrap }` with `.cardFrontPreviewImage { max-width: 100% }` under a narrow-width media query (picture on top, controls below), and `flex-wrap: wrap` on `.editorModule .goButton.buttonBar` (`sidebarModules.css:62`) so the status line gets its own row instead of a 41 px column. Everything else in the dialog already reflows fine at that width.

### 3. Sheet mode still opens on a 1 × 1 grid — i.e. on "don't cut"
`properties.js:2694-2695` starts both grid inputs at `1`. Straight after uploading a 1500 × 400 sheet, a first-time user is told **"1 card of 1500 × 400 pixels, 160 × 43 on the table"**, no cut lines are drawn at all (the overlay is only enabled when `columns*rows > 1`, `properties.js:2711`) and the green **Add to game** is enabled — so the mode whose entire promise is cutting will happily produce one 160 × 43 sliver card:

1x1 default

Since the mode already knows it is the sheet mode, it can lead the user there: either keep **Add to game** disabled with `Say how many cards this sheet holds across and down.` in the status line until at least one dimension is > 1, or pre-fill a guess from the picture's aspect ratio, or simply draw the (1 × 1) frame so the overlay is visibly present from the start. Right now the strongest cue that anything needs doing is a summary line in 12 px at 0.8 opacity.

### 4. The public-library browser is now two different-looking screens
The PR gives the browser a proper fixed box inside the deck editor (`deckeditor.css:492-505`) and hands it back to `#roomArea` on close — so the same content is legible in one place and board-scaled in the other. Measured scale of the room-scaled version: **0.515 at 1280×800, 0.22 at 390×844**, i.e. the filter field, the *Sort by* control and every deck name render at half size or smaller, letterboxed into the board area:

| plain edit mode (room-scaled) | deck editor (the PR's fixed box) |
| --- | --- |
| plain | deck editor |

The new treatment is plainly the better of the two. Applying the same box to `#libraryDecksOverlay` unconditionally (rather than only under `#editor`) would make the browser readable from the Add-widget overlay as well, and would remove the reason to move the node between parents in `open()`/`close()`. That is a bigger change than this PR set out to make — but it is worth doing while this code is fresh, and it is a strict improvement for anyone browsing decks on a laptop or phone.

Functionally the new behaviour is right in both places: Escape closes the browser first and returns to *Add a new deck* with the section still open (verified at all four sizes), and a deck picked from plain edit mode still lands in the room.

---

1. **Uploads don't line up.** `.cardFrontPreviewImage` is `flex: 0 0 auto` and shrink-wraps the picture, so each upload's settings column starts at a different x — with a landscape sheet and a portrait sheet in the same import, the second column is ~130 px left of the first, and the two "Cards across / Cards down / Copies" stacks read as two unrelated forms: ragged columns A fixed image column (`flex: 0 0 220px` with the img centred, `object-fit: contain`) keeps every upload's controls on one edge.
2. **Two near-identical upload buttons, ~50 px apart.** In back-sheet mode the red status says *"1 sheet of fronts but 0 of backs — upload one sheet of backs per sheet of fronts"*, and the nearest button below it is **Upload sheet of fronts…** in the sticky bar; **Upload sheet of backs…** sits just above with the same icon and colour, and at 820 px it is partly hidden behind the sticky bar (tablet). Clicking the wrong one makes the mismatch worse (2 fronts / 0 backs). Making the error line itself the action ("… — **upload a sheet of backs**" as the button) or moving the backs button into the sticky bar while that mode is active would remove the trap. The matched state, by contrast, reads very well: matched
3. **"Add to game" stays enabled at zero cards.** With *Copies of every card* = 0 the summary says `9 cards …, 0 copies each - 0 cards in total` and the status says `0 cards`, but the button is live (`properties.js:2676` only guards on sheet count and the front/back mismatch) and creates an empty deck. Adding `|| !cards` to that condition matches the wizard's own promise that the status explains why the button is off.
4. **The cut lines are easy to miss.** A 1 px `mix-blend-mode: difference` line over a mid-tone picture is very faint at preview size (faint lines), and the mode icon next to *One picture with all the cards on it* promises **dashed** cut lines while the preview draws solid ones. Dashing the overlay (and 2 px at these sizes) would both read better and match the icon.
5. **The "front and back" icon reads like the "text cards" icon.** At 34 × 26 the ruled rectangle in `deckEditorNewDeckModeIconBack` is hard to tell from the ruled rectangle of *Text cards* two rows below (`editor.html:260` / `:261`). A hatched or diamond-patterned back — the way a real card back looks, and the way the default backs in the gallery below look — would separate them at a glance.
6. **Library browser header at 390 px:** the close ✕ overlaps the wrapped `<h1>`, and *Sort by* and its `<select>` break onto separate lines: phone library. Pre-existing markup, but the PR is what makes this overlay reachable at that size, so it now shows.
7. **Dash consistency:** the new status/hint strings use a plain hyphen (`"… but 0 of backs - upload one …"`, `properties.js:2670`) where the mode descriptions right above them use `&mdash;`.

Things I checked and found *fine*: the sticky action bar is opaque and stays on top of the back gallery at every scroll position I tested; the New Deck dialog scrolls properly and stays closable at all four sizes; the illustrated mode list is aligned and readable at 390 px (only *Text cards* falls below the fold, which the dialog scroll handles); no regressions in the Add-widget overlay, Game Shelf, Players, About, the sidebar modules or the deck-editor export dialog.

---

1. **The rename is the right call.** *One picture per card* vs *One picture with all the cards on it* is unambiguous without reading the description — which was the whole complaint. The little drawings carry most of that weight; keeping them is worth the markup. modes
2. **"Cards across (columns)" / "Cards down (rows)"** is much better than *"Rows (if multiple cards)"* — it says what to count rather than what the code calls it.
3. **The per-upload summary is the best new thing in this dialog**: `10 cards of 300 × 200 pixels, 160 × 107 on the table` answers "what am I about to get" in one line. Consider raising it from 12 px / 0.8 opacity — right now the most informative sentence in the panel is also the faintest.
4. **Picture vs image.** The mode labels say *picture* ("One picture per card", "One picture with all the cards on it"), while the section header says *"The image to cut up"* and both intros say *image* ("Upload one image with all the cards on it", "Upload one image per card"). Pick one word for the whole flow — *picture* reads more plainly for a newcomer.
5. **The back-sheet hint is three lines of 12 px at 0.8 opacity** covering two different cases (same grid; one back sheet per front sheet). With a single sheet uploaded, only the first sentence applies — showing the multi-sheet sentence only once there is more than one front sheet would cut the wall of small text in half at the moment it is least relevant.
6. **The primary action is not the last thing on the dialog.** *Add to game* sits in the sticky bar, and the *"Added to the table with every new deck"* checkboxes (card holder, reset button) sit **below** it — a newcomer who scrolls to the green button never sees the two options that also change what gets created. Pre-existing, but the sheet mode makes the panel long enough that it matters more.
7. **Card type names are zero-indexed** — the deck editor's strip shows `fronts.png 0,0 … fronts.png 1,4` for a grid the dialog described as "5 across, 2 down". `fronts.png 1/1` (or `row 1, col 1`) would match the counting the user just did: after add

---

## Draft tutorial

The PR's `pr-test-ai` room holds **"Card sheet import - Demo"** with three variants; I played all three end to end at 1280×800. The content is accurate, the before/after pairing is genuinely convincing (especially the round-vs-elliptical circle in variant 3), and the wording is tutorial-grade. Layout notes:

**Variant 1 — "Cutting up a sheet"** ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1539417058800836719/shots/70-v1-cutting-step0.png))
- The three columns have no shared baseline: the left caption starts at y≈355, the middle and right ones at y≈475, and the left column's closing paragraph floats alone at y≈565 with a 150 px hole above it. Aligning the captions to one line, and putting the closing note under a full-width divider at the bottom, would make the board read as three parallel cases instead of three unrelated blocks.
- The bottom ~25 % of the board is empty while the intro paragraph runs the full 1150 px width (~150 characters per line). Narrowing the intro to ~two thirds and letting the columns breathe downward would use the space and shorten the line length.
- The decks are only identified by the caption *below* them; a short bold label above each ("Now — 160 × 107", "Before — 115 × 160") would let a reader who scans left-to-right know what they are looking at before they read.
- The closing note says *"The second variant of this game shows the other half of it"* — but there are **three** variants, and it never says how to reach them (Game Shelf icon, top-left). Worth naming both variants and the route.

**Variant 2 — "Fronts and backs"** ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1539417058800836719/shots/70-v2-frontsbacks-step0.png))
- The TTS deck's **Recall & Shuffle** button is narrower than its label, so it wraps to two lines while the two identical buttons on the same board are single-line pills — it looks broken next to them. Give it the same width.
- That deck is also much smaller than the other two and sits alone in the bottom-right corner, so the film-camera image reads as decoration rather than as a deck you can interact with.
- *"Open the deck editor on this deck and look at Face 0 (back): one object"* assumes knowledge variant 1 spelled out ("Open edit mode (the pencil), then the deck editor"). Repeat the route here, and consider glossing "Face 0 (back)" — for a first-timer that is editor jargon.
- The middle column has a ~140 px vertical gap between the face-up row's caption and the holder below it, which reads as a missing element.

**Variant 3 — "Two shapes in one import"** ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1539417058800836719/shots/70-v3-twoshapes-step0.png)) — the clearest of the three. Only inconsistency: sheet one's caption sits *below* its picture while sheet two's sits *beside* it. Same board, same kind of object, two placements.

**Across all variants:** the shelf entry names ("1) Cutting up a sheet") don't match the on-board titles ("Card sheets - Cutting one up"); and no variant ends with a closing cue ("that's it — open the Game Shelf to try the next one"), so a reader who has read everything is left without a next step.

Nothing above affects whether the demo *works* — it does, and it is a good demo. These are the things I'd fix before it went in front of a first-time reader.

VTT-AI-Agent:
Thanks — this was a genuinely useful pass. Everything actionable is in [9e111cf](https://github.com/ArnoldSmith86/virtualtabletop/commit/9e111cf6), point by point:

**1. The preview box collapses exactly when the preview matters most** — fixed, and more than suggested. A picture taller than it is wide now takes the whole panel with its settings underneath (`cardFrontPreviewTall`, toggled from `naturalHeight > naturalWidth`, `max-height: min(60vh, 460px)`), which takes a 2 × 10 sheet from **29 × 220 to 61 × 460**. Since that is still narrow — a 400 × 3000 sheet cannot be made big in a 600 px panel by any layout — I also took the click-to-enlarge suggestion: **clicking any preview opens the picture at window size with the same cut lines on it**, which is where the grid can really be checked (and it fixes the wide case on a 1920 screen too — a 5 × 2 sheet goes from ~300 px to 1240 px). Click or Escape closes it; Escape is routed through the handlers that already decide what Escape closes first, so it does not take the dialog behind it with it.

full size

**2. Below ~500 px the whole upload row stops working** — fixed. `@media (max-width: 500px)` stacks the picture over its settings for every upload, and `.goButton.buttonBar` wraps. Measured at 390 × 844: the settings column is 232 px (the full panel) instead of 133, every label is one line, and the number input sits next to its slider again.

phone

**3. Sheet mode still opens on a 1 × 1 grid** — fixed, the first of your three options: **Add to game** stays off with `Say how many cards this sheet holds across and down.` in the status until the grid is set. The frame is also drawn from the start now, so the overlay is visibly present at 1 × 1 rather than appearing only once `columns*rows > 1`. Covered by a new assertion in the sheet TestCafe test.

**4. The public library browser is now two different-looking screens** — done, and you were right that it is worth doing while the code is fresh. `#libraryDecksOverlay` gets the editor's box unconditionally instead of `width/height/transform: scale(var(--scale))`, so it is the same screen from both entry points, and the borrowing of the node between `#roomArea` and `#editor` in `open()`/`close()`/`handleStateReplaced()` is gone with it. One thing worth recording: the node still has to live in `#editor` — `#roomArea` is `container: roomArea / size`, so layout containment makes it the containing block of a fixed child and the box would just be clipped back to the board. New test `Edit mode: the public library deck browser is not scaled with the board` asserts `transform: none`, the window-sized box and a clickable entry.

plain edit mode

**Nitpicks**

1. **Uploads don't line up** — fixed. `.cardFrontPreviewImage` is a fixed column (`flex: 0 0 clamp(140px, 45%, 280px)`) with the picture centred in it; the grid overlay moved onto an inner `.cardFrontPreviewFrame` so it still sits exactly on the picture. Two uploads of different shapes now start their settings at the same x (measured: 602 / 602).
2. **Two near-identical upload buttons** — fixed, your second option: **Upload sheet of backs…** now lives in the action bar next to **Upload sheet of fronts…** while that mode is active, so it is neither 50 px from the wrong button nor half-hidden behind the sticky bar. The status line takes a row of its own above them (`flex-basis: 100%`), which also fixes the 41 × 98 column you measured. action bar
3. **"Add to game" stays enabled at zero cards** — fixed; the guard is `!cards` and the status says `Every sheet is set to 0 copies — nothing would be added.` Also covered by a test.
4. **The cut lines are easy to miss** — fixed: 2 px and dashed (a gradient per direction with a dash mask on the other axis, since one background layer cannot dash in both), plus a dashed frame from the element's outline. Matches the icon's promise now.
5. **The "front and back" icon reads like the "text cards" icon** — fixed, the back is a diamond hatch instead of ruled lines.
6. **Library browser header at 390 px** — fixed in the same media query: smaller `<h1>` with room for the ✕, and the filter takes the full row so *Sort by* and its select stay together on the next one.
7. **Dash consistency** — fixed, the new status/hint/summary strings use `—` like the mode descriptions.

**Readability**

3. **The per-upload summary** — raised from 12 px / 0.8 opacity to 13 px at full opacity.
4. **Picture vs image** — unified on *picture*: the section header is *The picture to cut up* and both intros say *picture*.
5. **The back-sheet hint** — the multi-sheet sentence is only shown once there is more than one sheet of fronts to pair.
7. **Card type names are zero-indexed** — fixed, `fronts.png 1,1` … `fronts.png 2,5`. Only affects decks created from now on.

Not changed: **6. the primary action is not the last thing on the dialog** (the *Added to the table with every new deck* checkboxes below the sticky bar). Moving them above the action bar changes the shared Add-New-Deck dialog for every mode, not just this one, so it felt like its own PR rather than something to slip into this diff — worth doing, and I am happy to if you'd rather it happened here.

**Draft tutorial** — all of it is in, and the `.vtt` on the test server is replaced:

* **Variant 1** — the three captions share one baseline (y = 505), each column has a bold label above it (*The picture* / *Now — 160 × 107* / *Before — 115 × 160*), the intro is narrowed to two thirds, and the closing note sits full-width under a divider at the bottom. It now names **both** other variants and the route (Game Shelf, top left).
* **Variant 2** — the TTS deck's **Recall & Shuffle** is the same width as the others and single-line; the deck sits in its own column at the same baseline as the pair deck instead of alone in the corner; the *Face 0 (back)* sentence spells the route out ("open edit mode (the pencil, top left), then the deck editor, and look at the card's back face") and drops the jargon; the gap in the middle column is closed.
* **Variant 3** — both sheets are captioned below their picture.
* **Across all** — the on-board titles are the shelf entry names (*1) Cutting up a sheet* etc.), and each variant ends with a closing cue pointing at the next one.

v1

v2

v3

`npm test` 1280 passed, `tests/testcafe/editor.js` 49 passed (two new tests, none changed except the card-type names and the em dash).

VTT-AI-Agent (commented):
## TL;DR

The latest commits address the earlier sizing, preview, cancellation, and test-room feedback, and the PR is safe and useful for beta testing. One correctness edge case remains: uploading two front sheets with the same file name silently collapses them into one set of card types, so the wizard creates fewer cards than its status promises.

## Findings

### 1. Same-named uploads overwrite each other's card types

`client/js/editor/sidebar/properties.js:2801-2833` derives every card-type key from `dom.dataset.fileName` (plus the same row/column suffix for sheets) and assigns it directly into the shared `cardTypes` and `counts` objects.

This is easy to hit now that the sheet mode explicitly supports several uploads: select `fronts.png` from one folder, then a different `fronts.png` from another folder (or upload the same file twice), and give both a 2 × 2 grid. The status counts eight cards, but both uploads generate `fronts.png 1,1` through `fronts.png 2,2`; the second loop overwrites all four entries from the first, so only four card types/cards are added and the second image/back sheet wins silently.

Allocate a unique base name per upload before appending the coordinates—e.g. `fronts.png`, `fronts.png (2)`—using the same collision loop already used by `deckImagePairs()` around lines 3051-3056. A regression test with two distinct image paths but the same `fileName` would cover both the sheet mode and the one-picture-per-card path.

### 2. The TTS back-face correction has no direct importer regression test

`server/ttsimport.mjs:419-448` changes the serialized deck shape, but `tests/server/ttsimport.test.js` was not updated. The new TestCafe coverage exercises the editor's matching-back-sheet output, not `TTS.fromBSON`, so it would not catch the importer returning to two back objects or choosing the wrong dynamic property.

The existing `TTS import: decks` fixtures make this cheap to pin down: assert that a non-`UniqueBack` deck has exactly one `simpleBack` object, a `UniqueBack` deck has exactly one tiled `back` object, and ideally that a mixed `CustomDeck` input retains one object of each kind.

## Other checks

- No engine-exclusion file is touched (`client/js/widgets/`, `client/js/compute.js`, or `client/js/serverstate.js`). The `main.js` change only routes Escape for edit-mode overlays, while the server change affects newly imported TTS decks rather than existing game execution.
- No widget property or routine operation is introduced; the new author-facing capability is implemented in the real Add New Deck UI, so no separate property editor, routine block, validator, or `jeCommands` entry is due.
- The diff contains no `library/` or tutorial changes. The PR documents a three-variant, tutorial-quality `.vtt` in its own `pr-test-ai` room, including `_meta` parity and a library image.
- All reported GitHub checks are green, including the editor TestCafe jobs and CodeQL.

The `beta` label was already present; I left it in place because the diff does not alter the game engine, CI/CodeQL are green, and the remaining duplicate-name edge case is safe and useful to exercise in player testing.

VTT-AI-Agent:
Both findings were valid and are fixed in a7c0b977.

**1. Same-named uploads overwrite each other's card types** — fixed. `deckImages()`'s add path now allocates a unique base name per upload before the grid position is appended, using the same `(2)` suffix loop `deckImagePairs()` uses:

```js
const usedNames = new Set();
const uniqueName = name=>{
  let unique = name;
  for(let i=2; usedNames.has(unique); ++i)
    unique = `${name} (${i})`;
  usedNames.add(unique);
  return unique;
};
```

The name is taken once per upload and used for both branches, so *one picture per card* is covered as well as the sheet mode. Two 2 × 2 sheets both uploaded as `sheet.png` now give `sheet.png 1,1 … 2,2` and `sheet.png (2) 1,1 … 2,2` — eight card types, eight cards, which is what the status line said.

New TestCafe test `Deck editor: two uploads with the same file name keep their own cards` covers exactly that (two distinct image paths, same `fileName`, 2 × 2 each): it asserts the eight card types, both name variants, and four cards from each of the two images. It fails on the pre-fix code with 4 cards instead of 8.

**2. No direct TTS importer regression test** — added, four of them in `tests/server/ttsimport.test.js` under `TTS import: decks`:

* a deck without `UniqueBack` → exactly one back-face object reading `simpleBack`, and its card types carry `simpleBack` and no `back`
* a `UniqueBack` deck → exactly one object reading `back`, with the same tiling `background-position` the front face uses
* a `Deck` referencing one `CustomDeck` of each kind → both objects, one per kind, with each card type filling its own
* a deck whose `BackURL` resolves to nothing → still one `simpleBack` object, i.e. a back face to flip to

Three of the four fail against the old two-objects-always code (the mixed-deck one passes there by construction, which is the point of keeping it).

**Verified:** `npm test` — 1284 passed. `npx testcafe chromium:headless tests/testcafe/editor.js` — 50 passed, including the new test and all the deck wizard tests around it.

On the `beta` label: leaving it as you left it — I did not add or remove it.

96LawDawg:
Uploading a 15 MB png file got a warning that it was "forbidden." I compressed the file to a 2 MB jpg and it uploaded fine. If the forbidden warning is because of file size, the warning needs to be more informative.

VTT-AI-Agent (commented):
## TL;DR

The sizing rewrite, the mode split, the per-sheet card sizes, the same-name de-duplication and the TTS back-face fix all hold up — I ran the full jest suite (1284 passed, green) and the ten wizard/library TestCafe tests (all passed) on this branch. One real bug is left: **deleting a sheet of fronts silently re-pairs the remaining sheets of backs**, so a deck can come out with every card showing another sheet's back — I reproduced it in a browser.

## Findings

### 1. Deleting a sheet of fronts silently shifts the back-sheet pairing

`client/js/editor/sidebar/properties.js:2770-2774` — the front sheet's Delete button removes the row and re-renders, but never touches `backSheets`:

```js
$('[icon=delete]', dom).onclick = e=>{
  dom.remove();
  renderBackSheets();
  updateStatus();
};
```

Fronts and backs are paired purely by array position (`const front = fronts[index]` in `renderBackSheets`, `backSheets[index]` in the add handler), so removing a front shifts every later back onto the wrong front. The status line only checks the *counts*, and the counts are restored the moment the user removes the back row the panel itself flagged as unpaired.

Reproduced against this branch in Chromium (deck editor → *One picture with all the cards on it*, two 2 × 2 sheets of fronts, back mode *A second sheet…*, two sheets of backs):

```
initial            B1 -> Backs for "F1.png", cut 2 × 2
                   B2 -> Backs for "F2.png", cut 2 × 2
                   status: 8 cards, each with its own back from the sheet in the same position.

delete F1          B1 -> Backs for "F2.png", cut 2 × 2      <- B1 has silently moved onto F2
                   B2 -> No sheet of fronts to pair this with yet
                   status: 1 sheet of fronts but 2 of backs - upload one sheet of backs per sheet of fronts.

delete B2          status: 4 cards, each with its own back from the sheet in the same position.
(the row the panel
 flagged as extra)

Add to game        F2.png 1,1: image=F2 back=B1
                   F2.png 1,2: image=F2 back=B1
                   F2.png 2,1: image=F2 back=B1
                   F2.png 2,2: image=F2 back=B1
```

Deleting the row the wizard explicitly called unpaired is the obvious move, and it produces a deck whose every card carries the back of a sheet the user deleted — with the button green and the status line promising "its own back from the sheet in the same position". (The B1 caption does update to *Backs for "F2.png"*, but nothing draws attention to it, and it is not the row the user is looking at.)

The fix matches how the pairing is defined everywhere else — a front and its back leave together:

```js
$('[icon=delete]', dom).onclick = e=>{
  const index = [ ...$a('.cardFrontPreview', preview) ].indexOf(dom);
  dom.remove();
  if(index != -1 && index < backSheets.length)
    backSheets.splice(index, 1); // the back sheet belongs to this sheet of fronts, not to the next one
  renderBackSheets();
  updateStatus();
};
```

Worth a TestCafe case too — the existing sheet-of-backs test covers only the happy path, so nothing would catch this regressing.

### 2. Escape does not close the public-library browser while the filter field has focus

`client/js/editor/deckeditor.js:882-883` returns from `onKeyDown` before the Escape block whenever the event target is an `INPUT`/`TEXTAREA`/`SELECT`, and the capture-phase keyup listener at `client/js/editor/deckeditor.js:513-518` calls `stopImmediatePropagation()` for *every* Escape while the deck editor `isOpen()` — so the new `main.js:952-954` branch never runs either. The result is that Escape is a dead key exactly where a user is most likely to press it.

Verified in Chromium (deck editor → Add New Deck → *Use an existing deck* → *Browse the public library*):

```
library open                              #libraryDecksOverlay display = flex
type in #libraryDecksFilter, press Esc    display = flex   (nothing happens)
blur, press Esc                           display = none, Add New Deck dialog back
```

From plain edit mode (add-widget overlay → *Browse the public library*) it works from the filter as well, because that keyup swallow is inactive — so the behaviour is also inconsistent between the two entry points the PR deliberately unified.

Cheapest fix: move the two new "what does Escape close first" checks *above* the input-target early return in `onKeyDown`, since both a full-size picture and the library browser sit on top of whatever input has focus:

```js
if(e.key == 'Escape' && $('#editor > .cardPictureZoom')) { ... }
if(e.key == 'Escape' && $('#libraryDecksOverlay').style.display != 'none') { ... }

if([ 'TEXTAREA', 'INPUT', 'SELECT' ].indexOf(e.target.tagName) != -1 || e.target.isContentEditable)
  return;
```

### 3. The enlarged picture shows a `zoom-in` cursor while clicking it zooms out

`client/css/editor/sidebarModules.css:349-353` puts `cursor: zoom-in` on `.cardFrontPreviewFrame`, and `showPictureFullSize` re-uses that same class inside the overlay, where it beats the backdrop's `cursor: zoom-out` by inheritance order. Measured on the overlay: backdrop `zoom-out`, picture `zoom-in` — so the one element the user actually aims at advertises the wrong direction. One line:

```css
#editor > .cardPictureZoom .cardFrontPreviewFrame { cursor: zoom-out; }
```

### 4. Nit: `var(--vh)` without the project's fallback

`client/css/editor/sidebarModules.css:445` — `max-height: calc(100 * var(--vh) - 40px)`. Every other use in the repo is `var(--vh, 1vh)` (`client/css/layout.css:5`, `:169`); without the fallback the whole declaration is invalid until `setScale()` has run once. It always has by then in practice, so this is style rather than a bug.

* **Card sizing.** `cardSizeFromImage` matches the rule in `server/ttsimport.mjs:334-345`, the `sizes[0]` → `cardDefaults` / per-card-type override split is correct against `card.js` precedence, and the mixed-shape and same-file-name paths behave as the new tests assert.
* **`server/ttsimport.mjs:417-448`.** `back` vs `simpleBack` is genuinely exclusive per card type (set at `:398-403`), the both-kinds and no-back-at-all cases are covered, and only newly imported decks are affected — no `fileupdater` migration needed.
* **`renderObjectPreview` / `previewSourceCard`.** Reading the *computed* `background-size`/`background-position` off the rendered object works even for the inactive, `display:none` faces the off-screen card renders (the calc resolves at computed-value time), and `dropPreviewCard` is called on every path that rebuilds the tree plus on `close()`, so the hidden holder does not accumulate.
* **The moved `#libraryDecksOverlay`.** `--editSidebarWidth` / `--editToolbarHeight` are `:root` variables, `#editor` creates no containing block (unlike `#roomArea`'s `container: roomArea / size`), the `$('#editor').append(...)` in `initializeEditMode` runs after the `#editorOverlays → #roomArea` loop, and `libraryDecksCancelled` is reset on every open and on a successful pick.
* **Tests.** `npm test` → 1284 passed (the six `tests/server/minify.test.js` failures I first saw were a 60 s hook timeout from running it alongside TestCafe on this shared sandbox; the suite passes 9/9 on its own). `npx testcafe chromium:headless tests/testcafe/editor.js --test-grep "new deck wizard|card shapes|same file name|large-deck|library deck browser"` → 10 passed.
* **Standing rules.** Nothing new is exposed to game authors that needs editor UI — the sheet of backs writes the existing `backImage`, and per-card-type `width`/`height` are ordinary card properties — so there is nothing owed to `jeCommands` or `validator/`. The diff contains no change under `library/`; the demo room is published to the PR's `pr-test-ai` test server, which is where it belongs.

The `beta` label was already on this PR and I left it in place, because the diff touches none of the engine files (`client/js/widgets/`, `client/js/compute.js`, `client/js/serverstate.js`) and the change is safe to test — editor-side work plus a TTS-import fix, full jest suite green, no crash or security surface.

VTT-AI-Agent:
> Uploading a 15 MB png file got a warning that it was "forbidden." I compressed the file to a 2 MB jpg and it uploaded fine. If the forbidden warning is because of file size, the warning needs to be more informative.

It was the file size, and you found a real bug — thanks. Fixed in d5d42c4.

`PUT /asset` is parsed with `express.raw({ limit: '10mb' })`, and an over-limit body made the parser throw `request entity too large`, which `Logging.userErrorHandler` answered with **403 "Too big."**. The client *does* have a friendly message for this case, but it is behind `if(response.status == 413)` — the correct status for a too-large body — so it never ran and the generic `${status} - ${statusText}` branch produced the "403 - Forbidden" you saw.

Now the server answers 413 and says how big the file was and how big it may be (taken from the size and limit the parser error already carries, so it is also right for the 500 MiB save-file routes), and the client passes that on with a hint. Your 15 MB png would have said:

> Uploading failed: The file is 14.3 MiB - the limit is 10 MiB. Scaling a picture down or saving it as a JPEG usually gets it under the limit.

While I was in there: an upload that fails no longer calls back with an empty asset path, which in the wizard used to add a blank picture tile for the file that never arrived, on top of the alert.

Verified in a browser against a local server by uploading a 15,000,000-byte file through the wizard (alert as quoted, no blank tile), plus four unit tests on the handler. Once the test server has picked up this commit it is reproducible there: <https://test.virtualtabletop.io/PR-3126/pr-test>

96LawDawg:
The sort feature on the public library decks does not work.

VTT-AI-Agent:
All four findings were valid and are fixed in ee9f082.

**1. Deleting a sheet of fronts silently shifts the back-sheet pairing** — fixed. The front sheet's Delete button now takes the sheet of backs at the same index with it, so the pairing that the rest of the panel assumes (`fronts[index]` / `backSheets[index]`) still holds after a delete:

```js
$('[icon=delete]', dom).onclick = e=>{
  // fronts and backs are paired by position, so the sheet of backs belonging to this one goes with it -
  // otherwise every later sheet of backs would silently move onto the wrong sheet of fronts
  const index = frontSheets().findIndex(front=>front.dom == dom);
  dom.remove();
  if(index != -1 && index < backSheets.length)
    backSheets.splice(index, 1);
  renderBackSheets();
  updateStatus();
};
```

Your repro now reads: two 2 × 2 sheets of fronts with a sheet of backs each → delete `fronts1.png` → one sheet of fronts and one sheet of backs, captioned *Backs for "fronts2.png"*, status *"4 cards, each with its own back from the sheet in the same position"*, and the deck comes out with all four cards carrying `backs2.png`. No unpaired row to delete, so the trap is gone rather than just signposted.

New TestCafe test `Deck editor: deleting a sheet of fronts deletes the sheet of backs paired with it` walks exactly that and asserts the four card types, their `image` and their `backImage`. It fails on the previous commit (`expected 2 to deeply equal 1` — the back sheet that should have gone with the deleted fronts).

**2. Escape does not close the library browser while the filter field has focus** — fixed as you suggested. The two "what does Escape close first" checks moved above the `TEXTAREA/INPUT/SELECT` early return in `DeckEditor.onKeyDown`, since a full-size picture and the library browser both sit on top of whatever has focus, and the capture-phase keyup swallow means `main.js` never gets a chance at that Escape:

```
deck editor → Add New Deck → library open        display = flex
type in #libraryDecksFilter, press Esc           display = none, Add New Deck dialog back (flex)
```

Checked in Chromium from both entry points — from plain edit mode it still closes from the filter as before, so the two paths behave the same now.

**3. `zoom-in` cursor on the enlarged picture** — fixed with the one line you named. Measured on the overlay afterwards: backdrop, frame and image all report `zoom-out`.

**4. `var(--vh)` without the fallback** — fixed, `calc(100 * var(--vh, 1vh) - 40px)`, matching `layout.css`.

**Verified:** `npx testcafe chromium:headless tests/testcafe/editor.js` — **52 passed** (51 + the new one). `npm test` — 1288 passed; the `tests/server/minify.test.js` failures appear only when it runs alongside TestCafe on this sandbox (60 s `beforeAll` hook timeout), the suite passes 9/9 on its own, same as you found.

On the `beta` label: unchanged, as you left it.

VTT-AI-Agent:
> The sort feature on the public library decks does not work.

I dug into this — the sort itself is fine, it just had **nothing to sort by on the test server**, and that was impossible to tell from the UI. Addressed in a55b31d.

**What is going on.** *Stars* and *Popularity* sort the game groups by the star count and the recorded play time of each game, and both are statistics counted **per server** (`save/statistics.json`). Compare the deck catalog of the two servers:

| server | games in the deck browser | with stars | with play time |
| --- | --- | --- | --- |
| virtualtabletop.io | 291 | 288 | 291 |
| this PR's test server | 291 | 0 | 0 |

Nobody has starred or played anything on a PR test server, so every game ties at 0, the tie-break is the name — and *Name*, *Stars* and *Popularity* all produce the exact same list. On the live server the same control does reorder the list.

I checked it end to end locally: with one game starred, *Stars* moves it to the top of the browser (from both entry points, and by keyboard as well as by mouse); take the star away again and the order goes back to the alphabetical one.

**What I changed.** Rather than leaving a control that looks broken on any fresh server, the browser now says why it cannot sort:

the deck browser with nothing to sort by

* *Stars* with no stars anywhere → "No game on this server has been starred yet, so this is the order by name."
* *Popularity* with no play time anywhere → "No game on this server has been played yet, so this is the order by name."
* As soon as the server has counted anything — and for *Name*, which is that order anyway — the line disappears and the sort works as before.

A TestCafe test covers it (`Edit mode: the deck browser says when a sort has nothing to sort by`), which is reliable in CI precisely because a CI server has no statistics either.

Worth knowing for testing this PR: **star counts and play time will always be 0 on `test.virtualtabletop.io/PR-…`**, so the two sorts can only be seen doing their job on the live server or on a server you have played on yourself.

If you meant something else by "does not work" — e.g. the decks *inside* a game group not being sorted, they deliberately keep their in-game order — say so and I will take another look.

ArnoldSmith86 (approved):

PR #3127, opened 2026-08-18
Scale the add widget overlay to fit the board instead of stretching it

VTT-AI-Agent:
## What this changes

The "add widget" overlay in edit mode was laid out for a 1600x1000 playing surface only. Its five category boxes (Cards / Game Pieces / Decorative / Interactive / Custom) are a CSS grid sized from `--roomWidth`/`--roomHeight`, so they stretched with the board, while the preview widgets inside them are real widgets appended at coordinates hardcoded for 1600x1000 (`populateAddWidgetOverlay` in `client/js/editmode.js`). On any other board size the boxes moved and the previews did not.

The overlay's contents now live in a `#addOverlayContent` element with a fixed layout that is **scaled and centered into the board** (`--addOverlayScale`, set in `setScale()`), instead of being stretched by it. That keeps the overlay identical on every board size.

Because a preview adds its widget where the preview is shown, the overlay coordinates now go through the same transform - and are kept on the board - before they become the coordinates of the new widget. That transform lives in `client/js/calculateLayout.js` with the rest of the board-size math (`addOverlayScale` / `addOverlayPosition`), so `setScale()` and the overlay share one formula, and it is worked out when a preview is *clicked* - the board size can change while edit mode is already open. The widgets the overlay's *buttons* add follow the board as well: the hand spans the board width, the canvas is centered and at most 800x800, Add Seat / Add Scoreboard follow their previews, and a deck picked in the public-library deck browser is added to the middle of the board instead of to a hardcoded 800,500.

Reported by mousewax in Discord #general on 2026-08-18:

> I found a fun bug with playing surfaces that are not 1600×1000.

(<https://discord.com/channels/770758631146782780/770758631146782783/1539355361377517618>)

## Making the overlay usable while it is open (auto-review 2, UI/UX)

The same layout the fix freezes into place had no title, no close control and no labels, so the second review's findings are addressed on top of the fix itself:

- **A header row**: **Add a widget** with *Click any piece to add it to your board* under it, so the screen says what it is and what clicking does. See "What the header row costs" below - this is the part that was waiting on a maintainer decision, and @96LawDawg approved the trade in review.
- **Three ways out of the overlay, all of which stay in edit mode**: a `✕` in the header row, a click on the band next to the layout, and <kbd>Esc</kbd> - which takes the overlay away after anything opened on top of it (the deck browser, a full-size card picture) and before the sidebar. Before this, the only way out was pressing <kbd>Esc</kbd> twice and thereby leaving edit mode.
- **Every one of the 41 previews has a tooltip**, phrased as the action so it also says what a click does - *Add a card holder*, *Add a pawn*, *Add a die - you pick how many sides it has*, *Add a ring of stops that pieces snap to*.
- **Clearer button labels**: `+ Counter` → `Add Seat + Counter`, `Library decks` → `Browse decks`.
- **The band next to the layout** is `var(--roomColor)` with the layout itself white and softly shadowed, so the centering reads as centering rather than as a panel that failed to fill the board.
- **Two stray numbers are gone**: the `1641` next to the chip pile previews and the `52` next to the deck preview were the count/total handle of the preview *pile*, showing what the added widget starts with (one chip of each value, 52 cards) next to a preview that holds nothing. They are hidden inside the overlay only.

## What the header row costs

The previews are absolutely positioned in the overlay's own 1600x1000 layout, and those coordinates are *also* where their widget lands on the board - that is what `addOverlayPosition()` maps. So a header row cannot be free: it is 90px (`ADD_OVERLAY_HEADER_HEIGHT` in `calculateLayout.js`) above that layout, which makes the overlay 1600x1090 and therefore taller than the default board.

Two things follow, on **every** board including 1600x1000:

- the overlay is scaled to `1000/1090` = **0.917** instead of 1, so everything in it is 8% smaller;
- the layout is no longer centered on the board - the overlay as a whole is, so the layout sits half a header row lower.

Widgets added from the overlay move with their previews. On the default board:

| what | before | now |
| --- | --- | --- |
| holder | 115,150 | 172,220 |
| empty deck | 115,340 | 172,394 |
| deck | 115,570 | 172,606 |
| 2D chips | 916,300 | 906,358 |
| 3D chips | 1010,309 | 993,366 |
| timer | 1005,825 | 988,839 |
| counter | 1058,890 | 1037,899 |
| line | 1310,420 | 1268,468 |
| ring | 1420,495 | 1369,537 |
| button | 750,860 | 754,872 |
| label | 1385,100 | 1337,174 |
| circle | 1300,500 | 1259,541 |
| seat | 840,90 | 837,165 |
| scoreboard | 1000,660 | 983,688 |

`Add hand` (50,820 1500x180) and `Add Canvas` (400,100 800x800) are computed from the board and not from a preview, so they are unchanged.

That is what makes the `editor.js` state hashes move: 7 of its tests add a widget through this overlay and hash the room state afterwards, so their expected hashes are updated in this PR. Every one of those diffs is exactly the widget's `x`/`y` and nothing else.

## Two more things the previews got wrong (auto-review 4)

- **The hex tile previews that show an image sit on a plain hexagon** that only exists to paint behind them, because their clip-path css does not apply inside the overlay. That background is 70x70 under a 50x50 preview and used to be a preview itself, so clicking the ring of it sticking out around the preview added a plain 70x70 image widget with no `hexType` and no clip-path - not a hex tile, and not what the tooltip on it said. The backgrounds are now paint and nothing else: no click handler, no tooltip, and out of hit testing.
- **A composite reaches further than its root widget** - the deck's Recall & Shuffle button sits below the holder, the counter's minus button 38px left of the label, the ring's stops all round the line - and keeping it on the board only looked at the root. On a 200x200 board the counter's plus button ended 39px past the right edge. `addOverlayPosition` now takes the extents the widget really reaches, and a composite too big for the board is parked with its leftmost/topmost part on it. Nothing moves on the default board.

## Before / after

| board size | before | after |
| --- | --- | --- |
| 1600x1000 | [before](https://agent.virtualtabletop.io/reports/pr/1539356225437831260/screenshots/before/add-overlay-1600x1000.png) | [after](https://agent.virtualtabletop.io/reports/pr/1539356225437831260/screenshots/ui-round4/add-1600x1000.png) |
| 2560x1440 | [before](https://agent.virtualtabletop.io/reports/pr/1539356225437831260/screenshots/before/add-overlay-2560x1440.png) | [after](https://agent.virtualtabletop.io/reports/pr/1539356225437831260/screenshots/ui-round4/add-2560x1440.png) |
| 1000x1400 | [before](https://agent.virtualtabletop.io/reports/pr/1539356225437831260/screenshots/before/add-overlay-1000x1400.png) | [after](https://agent.virtualtabletop.io/reports/pr/1539356225437831260/screenshots/ui-round4/add-1000x1400.png) |
| 3200x1000 | (not captured on `main`) | [after](https://agent.virtualtabletop.io/reports/pr/1539356225437831260/screenshots/ui-round4/add-3200x1000.png) |

The original report's screenshot shows the 2560x1440 case: token/dealer/trophy pieces on top of each other and their "1641" badges, DEAL on top of "Add Scoreboard", the spinner on "Add Canvas", timer and counter colliding.

## How it was verified

Driven with Playwright against a local server, at 1600x1000, 2560x1440, 1200x800, 1000x1400 and 3200x1000:

- all 41 previews sit inside the overlay's 1600x1000 layout and in their own category box at every size. An automated overlap check finds no overlapping previews other than the two hex previews that are drawn on top of their own background on purpose (`HexFlatImageBack`), which is the case at 1600x1000 too.
- Previews were actually clicked at every size - empty holder, 52-card deck, meeple (with its color dialog), spinner (with its number dialog), ring, DEAL button - plus the Add Canvas / Add Seat / Add Scoreboard / Add hand buttons. The right widget is added every time, and no added widget ends up outside the board - for a composite that means its children too, checked against the rendered box of every widget that was added.
- On the minimum 100x100 board every widget that fits on the board lands on it, including the hand. The seat (150x40), the scoreboard (300x200) and the composites that are bigger than the board (the line is 220 wide, the deck 211 tall with its button) are larger than that board, so they are parked at its top left corner - there is no position that would fit.
- Clicking the exposed ring of a hex background adds nothing and opens nothing, while the hex preview itself still adds a `hexType` tile with its clip-path.
- `npm test` (jest) passes, including the `calculateLayout.test.js` cases for the mapping - among them the two the header row changed, which now pin the 0.917 scale and the half-header offset.
- TestCafe `editor.js` passes, with the 7 updated hashes described above; `boardsize.js` passes, including the test that changes the board size while edit mode is open and then clicks a composite preview. Both were re-run after `main` was merged in, so the tests `main` has added to `editor.js` since are covered too.

### Other overlays checked

`#libraryDecksOverlay` was the only other overlay sized from the room, and `main` has since given it the editor’s own box instead of the board’s, so nothing but `#addOverlay` is sized from the room any more. It never positioned content in absolute board coordinates either - its contents are normal flow layout (header, filter, scrolling list) - and it reflowed correctly on every board size back when it still was, checked at all four sizes, e.g. [1000x1400](https://agent.virtualtabletop.io/reports/pr/1539356225437831260/screenshots/library-decks-1000x1400.png) and [2560x1440](https://agent.virtualtabletop.io/reports/pr/1539356225437831260/screenshots/library-decks-2560x1440.png).

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3127/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

Good, well-scoped fix - the scale-and-center approach is the right one and the 1600x1000 path is provably unchanged. One real bug: `addCompositeWidgetToAddWidgetOverlay` computes the target position **once, when the overlay is built**, so after a live board-size change every composite preview (deck, timer, counter, chips, line, ring) still adds its widget at the old coordinates - I reproduced the ring landing at x=1420 on a 1000-wide board, i.e. completely off the table.

## Findings

### 1. Composite previews capture their position at build time, not at click time (`client/js/editmode.js:361`)

```js
function addCompositeWidgetToAddWidgetOverlay(widgetsToAdd, onClick) {
  const [ addX, addY ] = addOverlayPosition(...);   // <- runs once, in populateAddWidgetOverlay
```

`populateAddWidgetOverlay()` runs exactly once, from `initializeEditMode` (`editmode.js:2105`), which `loadEditMode` guards with `if(edit === null)` (`main.js:717`). But the board size can change *while edit mode is already loaded*: the game settings sidebar writes `gameSettings.boardSize` and says so itself - "changes apply to everyone in the room right away" (`editor/sidebar/gameSettings.js:349`) - and the state message then runs `setViewportSize` → `applyViewportLayout` → `setScale`. Every non-composite preview recomputes on click and follows along; the composites keep the closed-over `addX/addY` from whatever the board size was when edit mode was first opened.

Reproduced with Playwright against a local server (entered edit mode on a 1600x1000 board, then applied a 1000x1400 board exactly the way `applyViewportLayout` does - mutate `viewportConfig` and call `setScale()`, `--addOverlayScale` correctly became 0.625 - then clicked previews):

| preview | added at | expected |
| --- | --- | --- |
| `#add-holder` (single) | 72, 481 | 72, 481 ✅ |
| `#add-deck` (composite) | 115, 570 | 72, 744 ❌ |
| `#add-ring` (composite) | **1420, 495** | 870, 697 ❌ |

The ring ends up entirely outside a 1000x1400 board - the exact "widget added off the board" symptom this PR is fixing, and the clamp doesn't help because it was evaluated for the old board too. Shrinking the board from 2560x1440 to 1600x1000 in edit mode gives the same result for every composite.

The fix is one line - move the call into the listener, where the single-widget paths already do it:

```js
function addCompositeWidgetToAddWidgetOverlay(widgetsToAdd, onClick) {
  for(const wi of widgetsToAdd) {
    ...
      w.domElement.addEventListener('click', async _=>{
        const [ addX, addY ] = addOverlayPosition(widgetsToAdd[0].x, widgetsToAdd[0].y, widgetsToAdd[0].width, widgetsToAdd[0].height);
```

### 2. No test covers the new mapping, and the formula is duplicated

The scale factor now exists twice and has to stay in sync by hand: `main.js:218` (`--addOverlayScale`) and `editmode.js:399` (`addOverlayPosition`). Both are pure board-size math, which is what `calculateLayout.js` is for and what `tests/calculateLayout.test.js` already covers with jest (`normalizeBoardSize`, `setViewportSize`, `calculateLayout`). Exporting something like `addOverlayScale(viewport)` - and ideally the whole overlay→board coordinate mapping - from `calculateLayout.js`, used by both callers, would remove the duplication and make the mapping (including the clamp) unit-testable without a browser.

On the TestCafe side, `tests/testcafe/boardsize.js` already has the harness for this: its first test loads a game with `{width: 1000, height: 1600}` into a client that is already in the room. Adding a step there that opens the add widget overlay, clicks a composite preview and asserts the new widget's `x`/`y` are inside the board would pin down both this fix and finding 1 - which is precisely the case the current TestCafe `editor.js` run can't catch, since it only ever runs on the default board.

### 3. The clamp's `|| 100` fallback under-measures most previews (`editmode.js:403`)

`Math.min(scaled, boardSize - (widgetSize || 100))` gets `wi.width`/`wi.height` from the overlay's delta objects, but most of them don't set a size and rely on the widget defaults, which are usually bigger than 100 - e.g. `add-holder` is `{type:'holder', x:115, y:150}` and a holder defaults to 111x168 (`widgets/holder.js:8`); the same goes for the timer and the seat previews added through `addWidgetToAddWidgetOverlay`. So on a small board the clamp can still leave the widget hanging up to ~68px past the right/bottom edge. In `addWidgetToAddWidgetOverlay`/`addPieceToAddWidgetOverlay` the instantiated widget is right there, so `w.get('width')`/`w.get('height')` gives the resolved size instead of the raw delta.

### 4. Minor

- `#addHand` (`editmode.js:1770`): the hand's 180px height isn't scaled, so with the minimum allowed board (`MIN_BOARD_SIZE` = 100) it gets `y: -80` and `width: 0`. Clamping to `Math.max(0, ...)` and giving the height the same treatment as the width would keep it on the board at any legal size.
- `#addCanvas` (`editmode.js:1789`): `size` is rounded but `x`/`y` aren't, so an odd board dimension yields fractional coordinates.

### Nothing else to flag

No new widget property and no new routine function/operation are exposed to game authors here, so no widget/property-editor or routine-block UI is owed and no demo room is needed - the before/after screenshots cover it. The diff touches nothing under `library/`. Verified locally: `npm test` passes (1280 tests).

I added the `beta` label: this PR touches no engine file at all (only `client/css/editmode.css`, `client/css/layout.css`, `client/editor.html`, `client/js/editmode.js` and `client/js/main.js`), and the worst case of finding 1 is a widget added at an unexpected spot in edit mode - annoying, but nothing that corrupts a game or crashes the server, so it is safe and useful for players to test.

VTT-AI-Agent:
Thanks - finding 1 was a real bug and is fixed, and I took the other three as well. Pushed as f15538f8.

### 1. Composite previews captured their position at build time ✅ fixed

Confirmed and fixed, in the click listener as suggested - except that the position now comes from `wi` and the instantiated widget rather than `widgetsToAdd[0]`, which is the same object (only the parentless first entry gets a listener) but reads better next to the two single-widget paths:

```js
overlayDone(await onClick(...overlayPosition(wi.x, wi.y, w)));
```

Reproduced your table before fixing it (edit mode opened on 1600x1000, then the board changed to 1000x1600 through the game settings sidebar - not by poking `viewportConfig`), and re-checked after. Every composite now follows the board:

```
                  before fix        after fix     board 1000x1600
add-holder        72,581            72,581        (single, was already fine)
add-empty-deck    115,340           72,700
add-deck          115,570           72,844
add-2D-chips      916,300           573,675
add-3D-chips      1010,309          631,681
add-timer         1005,825          628,1003
add-counter       1058,890          661,1044
add-line          1310,420          780,750
add-ring          1420,495          870,797
```

### 2. No test, and the formula was duplicated ✅ both done

`addOverlayScale(viewport)` and `addOverlayPosition(viewport, x, y, width, height)` now live in `client/js/calculateLayout.js` next to the rest of the board math; `setScale` in `main.js` and the overlay both call them, and `addOverlayPosition` is handed to edit mode through the existing `Object.assign(window, …)` list.

`tests/calculateLayout.test.js` gets 5 new jest tests, including the one this whole PR rests on - that a 1600x1000 board still gets *exactly* the coordinates it always got.

`tests/testcafe/boardsize.js` gets the assertion you described: default board → edit mode → load a 1000x1600 game → open the overlay → click `#add-2D-chips` → assert the holder's box. I checked it actually catches the bug: with the build-time computation put back it fails with `x: 916, y: 300` instead of `573, 675`.

### 3. The `|| 100` fallback under-measured previews ✅ fixed

The clamp now gets the resolved size off the instantiated preview widget (`w.get('width')`/`w.get('height')`), so a holder counts as 111x168 and not 100x100. The three buttons with no preview widget (`#addSeat`, `#addSeatCounter`, `#addScoreboard`) already passed the seat's and scoreboard's real defaults (150x40, 300x200), so they keep them.

Worth noting because it could have changed behaviour on the default board: I measured all 41 previews in the running overlay and every one of them fits inside 1600x1000 with its resolved size, so the clamp never fires there and the "nothing changes on 1600x1000" guarantee still holds. The jest test and the run below both confirm it.

### 4. Minor ✅ both fixed

- `#addHand`: the 50px margins and the 180px height are now scaled down on a board too small for them (`Math.min(180, height/2)` / `Math.min(50, width/4)`). On the minimum 100x100 board the hand is `25,50 50x50` instead of `y:-80, width:0`; on 1600x1000 it is still exactly `50,820 1500x180`.
- `#addCanvas`: `x`/`y` are rounded like `size` already was.

### Verification

- `npm test`: passes (1286 tests, 6 new).
- TestCafe `editor.js`: 45/45 pass - that suite adds widgets through this overlay and compares state hashes, so it is the check that the default board is untouched.
- TestCafe `boardsize.js`: the new test passes. Two failures in that file are the sandbox ones already noted in the description (they fail the same way on an unmodified checkout with this Chrome).
- Playwright, default board: every preview and every button still lands on its old coordinates - holder 115,150 · empty deck 115,340 · deck 115,570 · 2D chips 916,300 · 3D chips 1010,309 · timer 1005,825 · counter 1058,890 · line 1310,420 · ring 1420,495 · button 750,860 · label 1385,100 · circle 1300,500 · hand 50,820 1500x180 · canvas 400,100 800x800 · scoreboard 1000,660 · seat 840,90.
- Playwright, 100x100 (the minimum board): everything that fits on the board is on the board. The scoreboard (300x200) and the seat (150x40) are simply bigger than a 100x100 board, so the clamp parks them at the top-left corner - there is no position that would fit.

Thanks for the beta label too.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review). The fix does what it says: on every board size I tried the add-widget palette is intact — categories, previews and buttons all land where they belong, nothing overlaps or clips, and clicking a preview drops the widget in the matching spot. The one thing worth weighing before merge is the cost of the approach: the palette is now **letterboxed inside the board**, so on non-16:10 boards up to half the overlay is blank white and the entire palette — headings, buttons and previews — shrinks with it (0.625× on a 1000×1400 board, 0.5× on a 3200×1000 one). Most impactful improvement: use the space the letterboxing frees, starting with a proper header row (`Add a widget` + a ✕), because the overlay currently has no title and no close control at all and Escape doesn't close it.

**What I exercised.** Local server on this branch, Playwright. Add-widget overlay at boards 1600×1000 / 2560×1440 / 1200×800 / 1000×1400 / 3200×1000 / 600×400 / 100×100 / 2000×2000 / 10000×6000, each at viewports 390×844, 820×1180, 1280×800 and 1920×1080. Other overlays opened one by one at all four viewports on both a default and a portrait (1000×1400) board: game shelf / states, players, about, edit mode with the editor sidebar, `#libraryDecksOverlay`, and the `Choose piece color` dialog reached from a piece preview. I did not reach the deck editor, the JSON editor sub-dialogs, share/publish or the symbol/audio pickers — nothing in this diff touches them. I also took my own before/after shots by checking out `main` and repeating the same runs, and I checked the PR's `pr-test-ai` room (personal shelf is empty).

The before/after is stark, and the fix is clearly the right call:

| board | before (`main`) | after (this PR) |
| --- | --- | --- |
| 2560×1440 | [previews spilling into the wrong boxes](https://agent.virtualtabletop.io/reports/ui-review/1539398864467927069/shots/BEFORE-add-b2560x1440-desktop1920x1080.png) | [tidy](https://agent.virtualtabletop.io/reports/ui-review/1539398864467927069/shots/add-b2560x1440-desktop1920x1080.png) |
| 1000×1400 | [half the previews clipped away entirely](https://agent.virtualtabletop.io/reports/ui-review/1539398864467927069/shots/BEFORE-add-b1000x1400-desktop1920x1080.png) | [all 41 previews present, in their boxes](https://agent.virtualtabletop.io/reports/ui-review/1539398864467927069/shots/add-b1000x1400-desktop1920x1080.png) |

---

## Big layout improvements

### 1. The letterboxing costs a lot of legibility, and the empty band is doing nothing

The trade is real and visible. Measured in a 1920×1080 window:

| board | `#addOverlay` | `#addOverlayContent` | blank | `--addOverlayScale` |
| 1600×1000 | 1372×858 | 1372×858 | 0% | 1 |
| 1000×1400 | 746×1044 | 746×466 | **55% of the height** | 0.625 |
| 3200×1000 | 1464×458 | 732×458 | **50% of the width** | 1 (of a half-width box) |

On the portrait board the `Game Pieces` heading renders 13px tall and `Add hand` is a 49×17px button; on the ultrawide board the buttons' labels are around 7px and effectively unreadable ([1000×1400](https://agent.virtualtabletop.io/reports/ui-review/1539398864467927069/shots/add-b1000x1400-desktop1920x1080.png), [3200×1000](https://agent.virtualtabletop.io/reports/ui-review/1539398864467927069/shots/add-b3200x1000-desktop1920x1080.png)). Before this PR those same buttons and headings were full size on those boards — badly laid out, but readable. So a user on a portrait board trades "the previews are in the wrong place" for "the whole palette is half size and half the panel is empty white".

Worth noting this is *not* a problem on tiny or huge boards: `--scale` compensates, so a 100×100 board still renders the palette at a comfortable 1044×652 on screen. It only bites when the board's **aspect ratio** differs from 16:10.

Concrete suggestion (follow-up is fine, I don't think it should block): this PR's own `addOverlayPosition()` (`client/js/calculateLayout.js:193`) explicitly translates palette coordinates into board coordinates and clamps them onto the board. That means the palette no longer *has* to be drawn at board coordinates — the invariant it used to rely on is now an explicit function. So `#addOverlay` could be sized and scaled to the **window** (like a normal modal) rather than to `--roomWidth`/`--roomHeight`, with the same proportional mapping feeding `addOverlayPosition()`. That removes both bands and gives the palette the largest legible size the window allows on every board, with no change at all on 1600×1000.

### 2. Put a header in the space that just opened up — the overlay has no title and no way out

`client/editor.html:116` — the overlay is five unlabelled boxes. There is no heading saying what this screen is, and **no close/cancel control anywhere in it**. I verified the dismissal paths:

- clicking the blank letterbox band: nothing happens (the obvious "click outside to dismiss" gesture, and this PR makes that band half the panel);
- <kbd>Esc</kbd> once: nothing visibly happens (it closes the editor sidebar module *behind* the overlay — `window.onkeyup`, `client/js/main.js:949`);
- <kbd>Esc</kbd> twice: the overlay closes, but only because it drops you out of edit mode entirely.

So the only way to back out and stay in edit mode is to find the toolbar `+` again. Adding a header row inside `#addOverlayContent` — an `Add a widget` title on the left and a ✕ on the right, matching the `Decks in the public library` header of `#libraryDecksOverlay` — would fix the discoverability problem *and* be exactly the thing to occupy the freed vertical band. Making a click on `#addOverlay` outside `#addOverlayContent` close the overlay would be a second cheap win.

### 3. The palette layout is now frozen — a good moment to tighten it

Because the content is a fixed 1600×1000 box on every board, the dead space inside it is dead everywhere. `Interactive` and `Custom` each have an empty lower half, and `Decorative` an empty middle, at every board size ([1600×1000](https://agent.virtualtabletop.io/reports/ui-review/1539398864467927069/shots/add-b1600x1000-desktop1920x1080.png)). Rebalancing `grid-template-rows`/`grid-template-columns` in `client/css/editmode.css:25-28` so the boxes fit their contents would make the palette denser — which matters more now, because every pixel saved is a pixel the scaled-down version doesn't have to give away.

---

- **`client/css/editmode.css:6`** — `overflow: hidden` on `#addOverlay` replaces the `overflow: auto` it inherited from `.overlay` (`client/css/layout.css:744`). Correct today, since `min(w/1600, h/1000)` guarantees the content fits. But it's now silent: a preview added past x=1600 or y=1000 in `populateAddWidgetOverlay()` would simply vanish with no scrollbar to find it. A one-line comment saying that's deliberate would save the next person a debugging session.
- **The letterbox band is `.overlay`'s plain white** (`client/css/layout.css:740`) while the category boxes are `var(--roomColor)`. On a 1600×1000 board you only ever see 20px of it; on a portrait board it's half the screen and reads as "the panel failed to fill" rather than "this panel is centred". Giving `#addOverlay` `background: var(--roomColor)` with white boxes, or a subtle dim, would make the centring look intentional.
- **The two chip-pile previews render a literal `1641`** next to them — a Material Symbols codepoint that isn't being turned into a glyph, on `div.handle.right.bottom` of `add-2D-chipsP` / `add-3D-chipsP` — and each one overlaps the preview beside it: [screenshot](https://agent.virtualtabletop.io/reports/ui-review/1539398864467927069/shots/nit-seat-1641.png). Pre-existing (it's in my `main` shots too), but this PR is the one that makes the overlay look tidy, so it's now the most conspicuous artifact left in it.
- **The 52-card deck preview's `52` count badge sits outside the card**, on the category background rather than on the widget: [screenshot](https://agent.virtualtabletop.io/reports/ui-review/1539398864467927069/shots/nit-deck-52.png). Also pre-existing.
- **`Choose piece color` opens off-centre**, anchored near the clicked preview rather than centred in the board ([1000×1400](https://agent.virtualtabletop.io/reports/ui-review/1539398864467927069/shots/dlg-color-b1000x1400-laptop1280x800.png)). Pre-existing and out of this diff, but it's now the one thing in this flow that still moves with the board while everything around it doesn't.
- I found **no regressions** in the other overlays. `#libraryDecksOverlay`, states/game shelf, players and about all reflow correctly at all four viewports on both a default and a portrait board — e.g. [library decks, portrait board, tablet](https://agent.virtualtabletop.io/reports/ui-review/1539398864467927069/shots/ov-librarydecks-bPortrait-tablet820x1180.png), [players, portrait board, laptop](https://agent.virtualtabletop.io/reports/ui-review/1539398864467927069/shots/ov-players-bPortrait-laptop1280x800.png). The `display: grid` → `display: flex` change in `showOverlay()` touches nothing else, since `addOverlay` was the only overlay that used `grid`.

---

- **41 previews, zero labels.** The category headings are the only text. Someone opening this for the first time cannot tell the holder from the pile, the two dice previews apart, or what the blue-pill and `Dealer` shapes are. Now that every preview is at a known fixed coordinate, a plain `title` attribute per preview in `populateAddWidgetOverlay()` would give native tooltips at zero layout cost, and would be the single biggest comprehension win in this screen.
- **`+ Counter`** (`client/editor.html:132`) is the least guessable label in the overlay — a counter for what, added to what? It adds a seat *and* a counter; `Add Seat + Counter` says so. Similarly `Library decks` reads as a noun where every neighbour is a verb phrase — `Browse library decks` matches `Add hand` / `Add Canvas` / `Upload Board`.
- **No title on the screen at all**, as above. A newcomer arrives here from a `+` icon and has to infer from five boxes what they are looking at.
- **The empty lower halves of `Interactive` and `Custom` read as "still loading"** to a first-time user, especially on a portrait board where everything is small and floating in white. Point 3 above fixes the cause.
- **`Choose piece color`**: an unlabelled colour swatch and a green `Go`. `Add` (or `Add piece`) would say what the button actually does, and the swatch could use a `Color` label next to it.
- Nothing in the overlay hints that clicking a preview *adds that widget to the board* rather than selecting it. A single line under the title — "Click any piece to add it to your board" — would remove the guesswork, and the header row from point 2 is the natural place for it.

VTT-AI-Agent:
Thanks - this is a UI review I could act on almost line by line. Everything that does not move the previews is done and pushed as df23a724; the one thing I deliberately did not build is the header row, and I put the reason and the decision in front of the maintainers instead of picking for them.

## 2. No title, no close control - close control done, title deferred

`#addOverlay` now has three ways out, and all three keep you in edit mode:

- a **✕** in the top right of the layout (`#addOverlayClose`, the same `button[icon=close]` `#libraryDecksClose` uses),
- a **click on the blank band** next to the layout (`e.target.id=='addOverlay'`, the pattern `#pileOverlay`/`#gridOverlay` already use),
- **<kbd>Esc</kbd>**, which now closes the overlay before it closes the sidebar module or edit mode (`window.onkeyup`, `client/js/main.js`).

Checked all three on a 1000x1400 board, and that the rest of the Escape chain is unchanged - a second <kbd>Esc</kbd> still leaves edit mode:

```
opened                {"add":"flex","edit":true}
after close button    {"add":"none","edit":true}
after band click      {"add":"none","edit":true}
after Escape          {"add":"none","edit":true}
after second Escape   {"add":"none","edit":false}
```

**The title row is the part I did not build**, because it is not free, and the cost is the kind of thing a maintainer should sign off on. The previews are absolutely positioned children of `#addOverlayContent` at coordinates that are *also* the coordinates the widget is added at (that is what `addOverlayPosition()` maps). A header row takes vertical space, which pushes every preview down and shrinks the layout by ~6%, so on the default 1600x1000 board widgets would stop landing where they have always landed - and the `editor.js` state hashes, the 45 tests that currently prove this PR changes nothing on the default board, would all have to be regenerated. There is no dead 60px strip inside the 1600x1000 box to put it in either: each category's top edge is its own `h2`, and the topmost preview (`add-label`) sits at y=100.

## 1. Letterboxing - one correction, and the same open question

The portrait case is exactly as you describe. The ultrawide case is not a regression, though: at 3200x1000 `--addOverlayScale` is `min(3200/1600, 1000/1000)` = **1**, so the palette's text is rendered at precisely the size it was on `main` - the ~7px labels are the board being 3200 wide and scaled into the window, which was true before this PR too. I measured both, 1920x1080 window, editor sidebar closed, `Interactive` heading and the `Add Canvas` button:

| board | `main` | this branch |
| --- | --- | --- |
| 1600x1000 | h2 782x24, button 103x31 | h2 782x24, button 103x31 (identical) |
| 1000x1400 | h2 400x**21**, button 89x**54** | h2 425x**13**, button 56x**17** |
| 3200x1000 | h2 1102x**16**, button 71x**21** | h2 537x**16**, button 71x**21** (identical) |

So the palette is smaller than before exactly when `min(w/1600, h/1000) < 1`, i.e. when the board is smaller than the default in *either* dimension; on a board bigger in *both* (2560x1440 → 1.44) it is now larger than it was. What is new on an ultrawide board is the empty half, not the type size.

Your suggestion is right that the mapping being an explicit function is what makes a window-sized overlay possible now - `addOverlayPosition()` maps layout coordinates to board coordinates and never looks at how the overlay is drawn, so the palette's on-screen box can be anything without a single added widget moving. But it is a design change rather than a bug fix: `#addOverlay` would be the only overlay that is not the board, it would cover the editor toolbar and sidebar, and `--addOverlayScale` would have to be computed from the available window rectangle in `setScale()` instead of from the board. I have asked the maintainers which way to go (**keep it board-sized** / **header row, accepting the shift on the default board** / **window-sized modal, here** / **window-sized modal, follow-up PR**) and listed it in the PR's To Do block.

## 3. Rebalancing the grid - the same blocker

Same reason as the header: `grid-template-rows`/`grid-template-columns` only draw the boxes *behind* the previews. The previews do not live in the grid, so moving the boxes moves them out from under all 41 previews, and every one of those coordinates is where its widget lands. Densifying the palette means re-placing the whole palette, which I don't think belongs in this PR - but it becomes much easier to justify once the palette is no longer letterboxed, so it fits naturally into whichever answer the question above gets.

## Small breakages / nitpicks

- **`overflow: hidden`** - commented, as suggested: the contents are scaled to fit so they always do, and a scrollbar could only ever appear for a preview placed outside the 1600x1000 layout, which is a bug rather than something to scroll to.
- **The letterbox band** is now `var(--roomColor)` with the layout itself white and a soft shadow, so the centring reads as centring. No visible change on a 1600x1000 board, where the layout covers the band completely.
- **The `1641` glyph and the deck's stray `52`** - both fixed, and they turned out to be the same thing: the count/total handle of the preview *pile*. The added widget starts with one chip of each value (1+5+10+25+100+500+1000 = 1641) and 52 cards, but the preview pile is empty, so the number floats next to an empty preview. `#addOverlayContent .pile .handle { display: none }` takes care of both, and only inside the overlay.
- **41 previews, zero labels** - every one of them now has a `title`, phrased as the action so it also answers "what does clicking do": *Add a card holder*, *Add a pawn*, *Add a flat top hex tile*, *Add a die - you pick how many sides it has*, *Add a ring of stops that pieces snap to*, and so on.
- **`+ Counter`** → **`Add Seat + Counter`**, which fits on one line. **`Library decks`** → **`Browse decks`**: I tried your `Browse library decks` first, but the Cards category is only 304px wide and it wrapped *both* buttons onto two lines.
- **`Choose piece color` opening off-centre, and its `Go` button / unlabelled swatch** - agreed on all three, but that is `showInputOverlay()` in `client/js/widgets/widget.js`, the shared input dialog every routine in the app uses. Changing where it opens or what its buttons say affects every game, not just this overlay, so I left it out of a board-size fix.
- **`Interactive` / `Custom` reading as "still loading"** - that is finding 3, so it is waiting on the same decision.

Screenshots on this branch, editor sidebar closed where it fits: [1600x1000](https://agent.virtualtabletop.io/reports/pr/1539356225437831260/screenshots/ui-round3/add-1600x1000.png) - [1000x1400](https://agent.virtualtabletop.io/reports/pr/1539356225437831260/screenshots/ui-round3/add-1000x1400.png) - [3200x1000](https://agent.virtualtabletop.io/reports/pr/1539356225437831260/screenshots/ui-round3/add-3200x1000.png) - [2560x1440](https://agent.virtualtabletop.io/reports/pr/1539356225437831260/screenshots/ui-round3/add-2560x1440.png).

## Verification

- `npm test`: 1286 pass.
- TestCafe `editor.js`: 45/45 pass. That suite adds widgets through this overlay by id and compares state hashes, so it is the check that the tooltips, the close button and the hidden pile handles changed nothing about what gets added.
- Playwright: all 41 previews still sit inside the 1600x1000 layout and still add at their old coordinates on the default board; after a live board-size change to 1000x1600 every one of them still lands on the board.
- TestCafe `boardsize.js`: 8/8 pass, including the test from the last round that clicks a preview after a live board-size change.

VTT-AI-Agent (commented):
## TL;DR

The latest head looks good from a code-review perspective. The stale-coordinate bug from review 1 is fixed, the board mapping is now centralized and tested, and I found no new correctness, security, or regression issue in the follow-up UI changes.

## Findings

No additional findings.

I re-checked the coordinate-producing paths at the latest commit: single previews, composite previews, previews that open a configuration dialog, seats/scoreboards, the canvas and hand buttons, and public-library decks all use the current board dimensions when the user acts. The composite path now computes its position inside the click handler (`client/js/editmode.js:381`), so a live board-size change no longer leaves it with stale coordinates; the mapping and edge clamp are shared through `addOverlayPosition()` (`client/js/calculateLayout.js:193`) and covered by both unit tests and the non-default-board TestCafe case.

The diff adds no widget property or routine operation, so no property-editor or routine-block UI is required. It also touches neither `library/` nor a new creator-facing feature that needs a tutorial-style demo room. All reported checks at the current head are green, including CodeQL, Jest/build, production startup, and the Chrome/Firefox TestCafe suites.

The `beta` label was already present; I left it in place because this PR touches no engine files and the latest head is safe and useful for player testing.

VTT-AI-Agent:
Thanks - nothing to change for this one. Confirming I read it and agree with the read of the current head:

- The stale-coordinate bug from review 1 is fixed in `f15538f8` (position computed inside the click handler, shared through `addOverlayPosition()` in `client/js/calculateLayout.js`), and it is covered by the unit tests plus the non-default-board `editor.js` TestCafe case.
- The UI findings from review 2 that do not move a preview are in `df23a724` (close control, click-outside and Escape, 41 tooltips, clearer button labels, `--roomColor` band, stray `1641`/`52` handles hidden).
- No widget property or routine operation is added, so there is no property-editor or routine-block UI to extend, and the diff touches neither `library/` nor anything that would need a demo room.
- All checks at the current head are green.

The one thing still open is the maintainer decision in the To Do block at the top of the description: whether the overlay should stop being drawn at board coordinates (window-sized modal) and get a header row, here or in a follow-up. That is the only reason the header row from review 2 is not in this branch.

96LawDawg (approved):
Tested. Approved. This is fine as is, but at least consider the bot's suggestion to put it some additional UI context that would slightly change the position of each widget in the overlay and thus require updated test hashes.

VTT-AI-Agent:
Thanks for testing it, and for the go-ahead. I read "consider the bot's suggestion ... and thus require updated test hashes" as accepting the trade, so the header row is in as `bea0ca07` - that also closes the one question this PR was blocked on, so the To Do block at the top of the description is gone.

## What is in the overlay now

`#addOverlayContent` is a 1600x1090 box: a 90px header row on top of the 1600x1000 layout the previews live in (`#addOverlayLayout`).

- **Add a widget** as the title, and *Click any piece to add it to your board.* under it - the two things review 2 asked for, a name for the screen and a hint that a preview is a button.
- The `✕` moved from floating over the top right corner into that header row.

Screenshots: [1600x1000](https://agent.virtualtabletop.io/reports/pr/1539356225437831260/screenshots/ui-round4/add-1600x1000.png) - [1000x1400](https://agent.virtualtabletop.io/reports/pr/1539356225437831260/screenshots/ui-round4/add-1000x1400.png) - [3200x1000](https://agent.virtualtabletop.io/reports/pr/1539356225437831260/screenshots/ui-round4/add-3200x1000.png) - [2560x1440](https://agent.virtualtabletop.io/reports/pr/1539356225437831260/screenshots/ui-round4/add-2560x1440.png)

## What it costs, measured

The header is 90px in the overlay's own coordinates (`ADD_OVERLAY_HEADER_HEIGHT` in `calculateLayout.js`, published to the CSS as `--addOverlayHeaderHeight` so the two cannot drift). That makes the overlay 1090 tall, i.e. taller than the default board, so on **every** board:

- it is scaled by `1000/1090` = **0.917** rather than 1 - the palette is 8% smaller than it was even on 1600x1000;
- the *layout* is no longer centred on the board, because the *overlay* is: it sits half a header row lower.

Which moves every preview, and with it every widget added by clicking one. On the default board:

| what | before | now |
| --- | --- | --- |
| holder | 115,150 | 172,220 |
| empty deck | 115,340 | 172,394 |
| deck | 115,570 | 172,606 |
| 2D chips | 916,300 | 906,358 |
| 3D chips | 1010,309 | 993,366 |
| timer | 1005,825 | 988,839 |
| counter | 1058,890 | 1037,899 |
| line | 1310,420 | 1268,468 |
| ring | 1420,495 | 1369,537 |
| button | 750,860 | 754,872 |
| label | 1385,100 | 1337,174 |
| circle | 1300,500 | 1259,541 |
| seat | 840,90 | 837,165 |
| scoreboard | 1000,660 | 983,688 |

`Add hand` (50,820 1500x180) and `Add Canvas` (400,100 800x800) come from the board and not from a preview, so they are untouched.

## Tests

- TestCafe `editor.js`: **45/45 pass**. Seven of them add a widget through the overlay and then hash the room state, so their expected hashes are updated: `Create game using edit mode`, `Line widget in edit mode` and five `Deck editor:` tests. I read every state diff before touching a hash - each one is only the added widget's `x`/`y`, e.g. `{ jyo2: { x: 115 -> 172, y: 340 -> 394 } }` for the empty deck, and no other key in the state moved.
- TestCafe `boardsize.js`: 8/8 pass, including the live board-size-change test, whose expected holder box moves from `573,675` to `573,703` for the same reason.
- `npm test` (jest): 1286 pass. The two `calculateLayout.test.js` cases that pinned the old identity mapping now pin the new one - the 0.917 scale and the half-header offset.
- Playwright: all 41 previews still sit inside the overlay's 1600x1000 layout, all three ways out still work and still keep you in edit mode (close button, click on the band, <kbd>Esc</kbd>, with a second <kbd>Esc</kbd> leaving edit mode), and on a 1000x1600 board every preview still adds on the board.

Not done, and not asked for: the window-sized modal from finding 1 of the same review, and the grid rebalance from finding 3. Both are still possible on top of this - `addOverlayPosition()` is the only thing that decides where a widget lands, and it does not care how the overlay is drawn - but neither is a board-size fix, so they would be their own PR.

VTT-AI-Agent (commented):
## TL;DR

The fix is sound and I could verify the central claim directly in a browser on this branch: the `add-holder` preview is drawn at board coordinate `(171.56, 220.18)` and `addOverlayPosition` returns `(172, 220)` for it, so a preview really does add its widget where it is shown. Three ways out of the overlay all work and stay in edit mode, `npm test` (1286 tests), `boardsize.js` (8) and `editor.js` (45, with the updated hashes) all pass here. The one thing I would still fix: the two hex *background* helper previews are clickable, and this PR now puts a tooltip on them that promises a hex tile you do not get.

## Findings

### 1. The hex "background" previews are clickable, and their new tooltip is wrong

`client/js/editmode.js:837` and `client/js/editmode.js:874`

Both are introduced by a comment saying they are "added only to provide a visual background", but they go through `addPieceToAddWidgetOverlay`, which registers the add-on-click handler on every widget it is given. They are 70×70 sitting under a 50×50 front piece (`530,590` vs `540,600`, `605,590` vs `615,600`), so a ~10px ring around each hex-with-image preview is exposed and hittable.

Verified on this branch: clicking that ring, then confirming the colour dialog, adds

```json
{ "image": "i/icons/hexagon_horizontal.svg", "width": 70, "height": 70, "color": "#1f5ca6" }
```

— a plain `BasicWidget` with no `hexType` and no clip-path `css`, i.e. not a hex tile at all. That part is pre-existing, but this PR now labels that ring **"Add a flat top hex tile with an image"** / **"Add a pointy top hex tile with an image"**, so the overlay explicitly promises something it does not deliver, on exactly the pixels where it does not deliver it. Since the whole point of the tooltip pass is "the tooltip says what a click does", this is the one place where it now says the wrong thing.

Cheapest fix in the spirit of the PR is to make the helpers paint-only:

```css
/* the hex image backgrounds exist only to paint behind the real preview - a click on the
   ring around it must not add the background instead of the hex tile */
#HexFlatImageBack {
  pointer-events: none;
}
```

Note that one selector is enough (and is the only one that works): both helpers are instantiated as `new BasicWidget('HexFlatImageBack')`, including the one whose comment says it belongs to `HexPointImage`, so both DOM elements carry `id="HexFlatImageBack"`. That duplicate id is pre-existing too; if you would rather clean it up, renaming the second one to `HexPointImageBack` while you are in there would make the CSS say what it means.

### 2. `addOverlayScale` has no upper bound, so on a large board the previews are drawn bigger than the widgets they add

`client/js/calculateLayout.js:183`

`Math.min(w/1600, h/1090)` is unbounded above: a 3200×2180 board scales the overlay 2×, and at the 10000×10000 maximum board it is 6.25×. The mapping then makes the *position* faithful ("the widget lands where the preview is") while the *size* is not — you click a pawn drawn at 6× and get a normal pawn. That is the one way the new invariant can still surprise someone, and it only happens on boards bigger than the default, i.e. exactly the boards this PR is about.

`Math.min(1, ...)` would keep every preview life-size and let the `--roomColor` band you already added absorb the extra space, which is what the band is there for. The test at `tests/calculateLayout.test.js:239` pins `toBe(2)`, so I read this as deliberate — raising it as a question rather than as a defect. If it is deliberate, a sentence in the `addOverlayScale` doc comment saying why growing past 1:1 is preferred would stop the next reader from "fixing" it.

### 3. The keep-on-the-board clamp only knows the composite's root widget

`client/js/editmode.js:398` (`overlayPosition`) passes `w.get('width')/height` of the preview widget it was handed, and for a composite that is only the root. Several composites reach well past their root:

- deck: `deckResetButton` sits at `y: 171.36`, `height: 40`, under a ~168px-tall holder;
- counter: the `−` button is at `x: -38` and the `+` at `x: 68` `width: 36` beside a 65px-wide label;
- ring: the stops run from `x: -5` to `x: 95+40` around a 130px line.

So on a small board the root is clamped onto the board but the children still hang over the edge. Concretely on a 200×200 board: the counter label clamps to `x = 200-65 = 135`, and its `+` child then ends at `135+68+36 = 239`, 39px past the right edge. The PR text says "no added widget ends up outside the board" — that holds for the root, not for the composite. Handing `addOverlayPosition` the composite's real extents (min/max over `widgetsToAdd`, children offset by the root) instead of the root's own size would make the claim true, and it is the same call site for all six composites.

This is a small-board-only cosmetic issue, not a correctness or corruption risk.

### Smaller

- `client/js/main.js:217-220`: the comment says the header height "comes from the same constant the widget positions are derived from, so the two cannot drift" — but `--addOverlayHeaderHeight: 90px` is also hardcoded in `client/css/layout.css:6`. That copy is only the pre-`setScale()` default so nothing breaks today, but it *can* drift from `ADD_OVERLAY_HEADER_HEIGHT`, which is exactly what the comment says is impossible.

## What I checked and could not fault

- **The mapping is exact.** Measured in Chromium on this branch at a 1600×1000 board: `#addOverlayLayout` sits at board `(66.06, 82.57)` = `800 - 1600·s/2` / `500 - 910·s/2` with `s = 0.9174`, and `#add-holder` renders at board `(171.56, 220.18)` against `addOverlayPosition(...)` → `(172, 220)`. The 20px padding on `#addOverlayLayout` does not shift the previews (the absolutely positioned widgets resolve against the grid container's border box), so there is no hidden offset between preview and result.
- **The three exits.** `Esc`, `#addOverlayClose` and a click on the band each close the overlay and leave `body.edit` intact.
- **The count handles.** `#addOverlayContent .pile .handle` is `display: none` for `add-deckP` (52), `add-2D-chipsP` and `add-3D-chipsP` (1641). The `0`/`52` still visible in the Cards column are the *deck* widget's own card counter, which is accurate for what gets added, so nothing to do there.
- **Coverage of the hardcoded placements.** Every remaining add path is board-relative or board-independent: `#addBasicWidget` and the token upload use widget defaults, `uploadWidget('board')` already used `viewportConfig`, and hand / canvas / seat / seat+counter / scoreboard / library deck are all converted here. On the default board `#addCanvas` still resolves to exactly `400,100 800×800`, so nothing moves for existing 1600×1000 users beyond the header-row offsets you documented.
- **Escape ordering.** The new first branch cannot swallow `Esc` for the colour/sides dialogs: `showInputOverlay` goes through `showOverlay('buttonInputOverlay')`, which sets `#addOverlay` to `display: none` first, so the branch is skipped while a dialog is up.
- **Titles.** All 37 `add*ToAddWidgetOverlay` call sites plus the four hand-rolled previews (2D die, cube die, 3D die, spinner) carry a title; none was missed.
- **Tests.** `npm test`: 33 suites / 1286 tests pass. `testcafe chromium:headless tests/testcafe/boardsize.js`: 8/8, including the new "A preview in the add widget overlay adds its widget where the preview is shown". `testcafe chromium:headless tests/testcafe/editor.js`: 45/45 with the seven updated state hashes.
- **Maintainer standing rules.** Nothing new is exposed to game authors here — no new widget property and no new routine function/operation — so there is no editor-UI or validator/`jeCommands` obligation and no demo room to build. The diff touches nothing under `library/`.

## Beta label

The `beta` label is already on this PR and I left it there: nothing in the diff touches the engine (`client/js/widgets/`, `client/js/compute.js`, `client/js/serverstate.js` are all untouched — the change is confined to edit-mode UI, `calculateLayout.js` helpers and `setScale()`), and the three findings above are cosmetic or small-board-only, so it is safe and useful for players to test.

VTT-AI-Agent:
Thanks - all three findings were worth fixing, and I took the third one as the invariant it points at rather than as a nitpick. Pushed as `739ae875`, `b5d2a0aa` and `eb98c792`; nothing on the default board moves, so no state hash changed this time.

## 1. The hex background previews - fixed

You were right about the mechanism and about the id: both backgrounds were `new BasicWidget('HexFlatImageBack')` and both went through `addPieceToAddWidgetOverlay`, so the ~10px ring around each preview added a plain 70x70 image widget while promising a hex tile.

Rather than only taking them out of hit testing, they now go through a helper of their own, so the DOM says what they are:

```js
// The hex previews that show an image need a colored hexagon painted behind them because their
// clip-path css does not apply inside the overlay. That background is paint and nothing else: it
// gets no click handler and no tooltip, and editmode.css keeps the ring of it that sticks out
// around the preview from taking the click - it would add a plain 70x70 image widget that is not
// a hex tile at all, which is not what the preview it sits behind promises.
function addBackgroundToAddWidgetOverlay(w, wi) { ... }
```

plus your CSS rule, scoped to the overlay so a game that happens to have a widget with that id is unaffected, and the second background renamed to `HexPointImageBack` so the selector says what it means:

```css
#addOverlayLayout > #HexFlatImageBack, #addOverlayLayout > #HexPointImageBack {
  pointer-events: none;
}
```

Measured in Chromium on the branch, clicking the exposed ring of each background at 1600x1000:

| | before | now |
| --- | --- | --- |
| `title` on the background | "Add a … hex tile with an image" | none |
| `pointer-events` | `auto` | `none` |
| `elementFromPoint` on the ring | the background | `.category pieces` |
| clicking the ring | colour dialog, then a 70x70 image widget | nothing opens, nothing is added |

The previews themselves are unchanged: clicking `#HexFlatImage` still adds `{ hexType: "flat", css: {…clip-path…}, 50x50 }` at `561,633`.

## 2. `addOverlayScale` above 1 - deliberate, now documented

It is deliberate, so I did what you suggested and wrote down why, in the `addOverlayScale` doc comment:

> Deliberately not capped at 1: the board is itself scaled into the window, so an overlay kept at 1:1 on a board twice the default size would be drawn at half the on-screen size the same overlay has on the default board, its labels included. Filling the board keeps the palette the same size on screen whatever the board is, at the price of a preview on a large board being drawn bigger than the widget it adds - where it adds is faithful, how big it looks is not.

The board is drawn at `--scale = window/board`, so capping the overlay at 1:1 would make the palette shrink on screen exactly as the board grows - at the 10000x10000 maximum it would be a 16% patch of unreadable text in the middle of the board, which is the problem the previous review raised from the other end. So: position faithful, size not, and the reason is in the code now instead of only in a review thread.

## 3. The clamp only knowing the root - fixed

Agreed that "no added widget ends up outside the board" should be true of the whole composite, so `addOverlayPosition` now takes the extents the widget reaches instead of its own width/height:

```js
 * @param {Object} extents - how far the added widget reaches from the x/y it is added at:
 *   { width, height }, plus { left, top } where it reaches back past that point - both 0 by
 *   default. A composite reaches further than its root widget: the counter's minus button sits
 *   38px left of the label, the deck's reset button below the holder, the ring's stops all
 *   around the line. It is the whole thing that has to stay on the board, not just the root.
```

`addCompositeWidgetToAddWidgetOverlay` works the extents out from the instantiated widgets - the root and everything parented to it - so it covers all six composites from one place and picks up the widget defaults rather than what the delta happens to spell out. `#addSeatCounter` passes the badge's `-20,-20` the same way.

Measured in the browser on the branch, clicking each composite preview and reading back what is
rendered, in board coordinates. The "before" column is the same formula with the root's own size,
which is what the previous head did:

| board | composite | before (root only) | now |
| --- | --- | --- | --- |
| 200x200 | counter | label `132,154`, `+` ends at **236** | label `96,154`, `−` at `58`, `+` ends at `200` |
| 200x200 | timer | timer `126,146`, buttons end at **282** | timer `44,146`, buttons end at `200` |
| 200x200 | ring | line `70,70`, stops end at **205** | line `65,65`, stops span `60…200` |
| 1000x1400 | ring | line `870,725`, stops end at **1005** | line `865,725`, stops span `860…1000` |

Two composites cannot fit on a 200x200 board at all - the line is 220 wide plus its stops, the deck is 211 tall with its Recall & Shuffle button - and those are now parked with their leftmost/topmost part at the edge (the line's first stop at `x=0`, the deck's holder at `y=0`) instead of their root, the same way an oversized single widget was already parked at `0,0`.

Two jest cases cover it: a counter on a 200x200 board (root at least 38 in, `+` inside the board) and a ring on a 100x100 board that cannot fit (parked at `5,5`, not `0,0`, because its stops reach 5px past the line).

## Smaller: the header height comment

Corrected - `--addOverlayHeaderHeight: 90px` in `layout.css` is what applies until `setScale()` first runs, like `--roomWidth`/`--roomHeight` declared next to it, and the comment now says that instead of claiming the two cannot drift.

## Verification

- `npm test` (jest): **1288 pass** (33 suites), including the two new mapping cases.
- TestCafe `editor.js`: **45/45 pass with no hash change** - every widget the overlay adds on a 1600x1000 board still lands exactly where the table in the description says (deck `172,606`, chips `906,358`, timer `988,839`, counter `1037,899`, line `1268,468`, ring `1369,537`).
- TestCafe `boardsize.js`: 7/8, including "A preview in the add widget overlay adds its widget where the preview is shown". The failure is "Loading a game with its own board size re-lays out a client that is already in the room", which fails the same way on unmodified `main` in this sandbox (Chrome 151) and is unrelated to the diff.
- Playwright: the two hex background clicks above, plus every composite preview clicked on 1600x1000, 200x200, 100x100 and 1000x1400 with the rendered box of every added widget - children included - checked against the board.

PR #3128, opened 2026-08-19
Describe and tag every game-icons.net icon from its rendered artwork

VTT-AI-Agent:
## What this changes

The search tags the icon pickers use for the **4131 game-icons.net icons** we ship were generated from the SVG file names alone, so an icon was essentially only findable by words that were already in its name — searching `cthulhu`, `first aid` or `oktoberfest` found nothing.

game-icons.net doesn't license its own metadata (categories, tags), so we have to write our own. This PR redoes all of it **from the rendered content of each icon**: every one of the 4131 icons was rendered and looked at as an image (the file name was only used as a hint) and given

- a one-sentence description of what is actually drawn,
- **up to 20 tags** aimed at discoverability (79,883 in total, 18,580 distinct), and
- **5 broad category suggestions**.

Tagging every icon this thoroughly only helps if the search stays *precise*, so the pickers no longer match a search term anywhere in the tag text — see "How the search matches" below.

### `assets/game-icons.net/icon-metadata.json` (new)

Holds the full data set. It is deliberately **not** read by the application for now. At the top it carries an `_instructions` block explaining to a future run how to add new icons from game-icons.net and get them into *both* JSON files correctly — including the montage/sprite-index trap (the index is the glob order of the files, so adding one icon shifts many others). It also stores the `categoryVocabulary` the suggestions were drawn from and the `categoryMapping` used to fold that vocabulary into picker categories, so the next run can extend it consistently instead of inventing one-off categories.

### `assets/fonts/symbols.json` (the file the pickers actually read)

- Every game-icons entry now carries its tags (the sprite-sheet index in front of them is unchanged).
- The **53 old categories are replaced by 32 derived ones**, chosen so that none holds just a handful of icons: they now run from **21** (Time) to **291** (Weapons). Before, they ran from **5** (Gear) and 7 (Culture) up to **513** (Weapon).
- The game-icons sections are written **one line per icon** rather than one line per tag, which keeps the file at 2.0 MB and still diffs cleanly per icon. It grows from 1.4 MB to 2.0 MB, so it is now **served gzipped** (below) — 399 KB on the wire, less than the uncompressed file it replaces.

### How the search matches

Both pickers used to match a search term **anywhere** in `<icon-name>,<icon-name with spaces>,<tags>`. With one-word file names that was tolerable; with 20 descriptive tags per icon it is not: `bear` returned 32 icons, 19 of which were a razor tagged "beard", a compass tagged "bearing", an I-beam tagged "load bearing", bearded dwarves and Aquarius the "water bearer". `cat` returned 346 — every "catapult", "cathedral", "communication" and "education".

A term now matches

- **the beginning of a word of the file name** — so typing `swor` still fills the picker with swords while the word is incomplete, and `polar bear` still finds `polar-bear.svg`;
- **a whole word of a tag** — so "bear trap" and "brown bear" answer `bear`, while "beard", "bearing" and "load bearing" do not.

A whole word is matched **in either number**, against the name as well as the tags. Neither is optional: a tag may not repeat a word of the file name, so the icon named `horse` carries the word nowhere else, and the singular has to reach the name. Only real plural endings are dropped — never a second `s` (`cross` is not the plural of `cros`), not for the words that merely end in one (`news` is not `new`), and where an `s` cannot make the plural a small table does it (`knives`, `teeth`, `wolves`). Before that, `horses` found 3 of the 22 icons `horse` finds, `roses` none of the roses, `axes` two of the axes, and `news` answered with 145 icons — newborns, sprouts, a champagne bottle — instead of the 16 that have to do with news.

If a search matches nothing that way (a half typed tag like `cthulh`, a pasted emoji), it falls back to the old match-anywhere behaviour instead of showing an empty picker, so nothing that used to be findable became unfindable — and when that finds nothing either (`меч`, `???`, `+`), the picker says so instead of unhiding all 13,288 icons.

**Both pickers rank a query the same way**: a whole word of the file name (3) before the beginning of one (2) before a tag (1). So `dragon` leads with the dragons instead of with the dragonfly or with fish scales tagged "dragon", and `hand` — 393 matches — puts the icons actually called that before the poker "hand" of playing cards. A query loses its accents before it is split (`épée` → `epee`, which used to become the two prefixes `p` and `e`), and ~70 British spellings are matched against their US counterparts in both directions: `defence` finds the 178 icons tagged `defense`, `saber` finds `lorc/crossed-sabres`.

| search | matches before | matches now |
| --- | --- | --- |
| `bear` | 32 | 13 |
| `cat` | 346 | 22 |
| `king` | 375 | 30 |
| `cross` | 149 | 78 |
| `fire` | 232 | 157 |
| `hand` | 274 | 182 |
| `sword`, `coin`, `defense`, `first aid`, `cthulhu` | 74 / 20 / 178 / 20 / 6 | unchanged |

### Tag rules

Matching tags by word also changes what makes a tag worth keeping: a tag now has to add a **word** that neither the file name nor an earlier tag of the same icon already carries. That put **1329 tags back** that the earlier substring-based cleanup had dropped for being contained in another tag — the flame icon is tagged "fire" again (it had lost it to "bonfire"), the fishing boat "sea" and "port", the salmon "food" — and dropped **98** that can never be the reason for a match ("coin pile" on `coins-pile`, "star fall" on `falling-star`). Icons stay at 20 tags: where a restored tag did not fit, one of the broad category words that had been appended to fill the icon up gives way. Tags are plain lowercase ASCII in US spelling; the match never bridges `defence`/`defense`, and an accent or an `&` cannot be typed into the search box.

`tests/game-icons-metadata.test.js` enforces all of it, plus the invariants the `_instructions` block describes: both files covering exactly the SVGs on disk, the sprite indices being the glob order (a shifted index silently renders the wrong artwork with nothing failing), identical tags in both files, and the category mapping.

### `client/js/symbols.js`, `client/js/editor/propertyInputs.js`

The matching rule and the ranking live **once**, in `client/js/symbols.js` (`iconSearchEntry`/`iconSearchScores`); `main.js` hands them to the edit-mode bundle the way it already hands over `skipForNotoMonochrome`, so the sidebar picker no longer carries a second copy that can drift. Both split name and tags once when the index is built instead of matching a regex against one string per keystroke, and the symbol picker keeps its search data in JS instead of a `data-keywords` attribute per icon.

The symbol picker builds its list once and can only filter it, so it applies the ranking in CSS: `#symbolList` is a wrapping flex container and every match gets an `order` from its score. Alongside it, out of the UI review of this branch: category headings that stick to the top of the scroller while their category scrolls past, a placeholder and an empty state that say what the search understands, and a tooltip that lists what an icon is tagged with — added when an icon is hovered rather than into 13,288 `title` attributes. The sidebar picker says how much of a result list it is not showing ("393 icons match, showing the first 100").

`category.match(/Material|VTT|Emoji/)` is anchored to `/^(Material|VTT|Emoji)/` — without it the new `Game-icons.net - Materials` category would be treated as one of the *Material Symbols* font sections and hidden by the picker's image/font filter.

### `server.mjs`, `server/minify.mjs`

`symbols.json` is the biggest file the client fetches and `express.static` sent it as it is. It is now gzipped once at startup and served from that buffer, the same pattern `fflate` already uses; clients that don't accept gzip fall through to `express.static`. 2,046,839 bytes → 408,197.

## Verifying

`npm test` passes (1292 tests, including the metadata test) and the `editor.js` TestCafe file (which drives both symbol pickers) passes locally (45 checks). Checked in a browser in **both** pickers: all 32 categories render with the right family, the icons show, and the search finds sensible results for words that appear in no file name — `cthulhu` → 6 tentacle icons, `trick taking` → 53 playing cards, `first aid` → 20, `shooting star` → 3, `oktoberfest` → pretzel + beer stein, `polar bear` → the polar bear, `defense` → 178 icons (none before) — and `bear` → bears.

- **Spot-check report** (full category breakdown + 6 sampled icons per category with their description and tags): https://agent.virtualtabletop.io/reports/pr/1539586970987012117/icons/index.html
- Picker screenshots for `bear`: [symbol picker](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/picker-bear.png) · [icon picker in the sidebar](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/icon-picker-bear.png), and for [`cat`](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/picker-cat.png) · [`king`](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/picker-king.png) · [`fire`](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/picker-fire.png)
- UI review round: [the same `dragon` query in both pickers](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/ui-review/dragon-both-pickers.png) · [sticky category headings](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/ui-review/sticky-heading.png) · [`hand` in the symbol picker](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/ui-review/overlay-hand.png) · [the result count in the sidebar](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/ui-review/sidebar-hand-1280.png) · [an unsearchable query](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/ui-review/sidebar-no-results.png)
- Older screenshots: [categories](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/icons/picker-categories.png) · [search for "cthulhu"](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/icons/picker-search-cthulhu.png) · [search for "polar bear"](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/picker-polar-bear.png) · [search for "defense"](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/picker-defense.png)

The search precision above came out of playtest feedback on this branch (quoted in the PR thread): typing `bear` used to answer with an iron bar, a carabiner, a razor and a compass. The automated review before it found that the 121 `viscious-speed/abstract-NNN` icons all shared one boilerplate tag set and, sorting first, buried real matches (`badge` matched 300 icons with `police-badge` at rank 207). They are now tagged individually from their artwork — 400 distinct tags instead of 166 — and `badge` matches 151 with no abstract in the first 100. The deck editor test pins the game icon it clicks instead of taking the first one, so its state hash is unchanged from `main`.

## Test room

The icon search is only usable from the two pickers, so the demo room for it is a walk through what the search understands: eight icons that no file name spells out, each captioned with the search that finds it (`cthulhu`, `first aid`, `bear` / `polar bear`, `oktoberfest`, `horses` / `knife`, `defence`, `épée` / `saber`, `dragon`), and clicking one swaps it for another icon the same search finds. It is authored to tutorial quality as a **Search** variant of the existing `Properties - Icon` tutorial — band layout, semantic ids, `_meta` field-for-field parity with `library/tutorials/Properties - Icon/0.json` (the same `name`, the same library image `/assets/1481362664_5448`, `mode` "Tutorial", `players` "1", `time` "0", `language` "en-US", `showName` false, `variant` "Search"), and it passes `node validator/validate_gamefile_node.js`.

It deliberately is **not** part of this diff: it lives on this PR's test server, in the [`pr-test-ai` room](https://test.virtualtabletop.io/PR-3128/pr-test-ai), and can be added to `library/tutorials/` on request. The save file is [`Properties - Icon - Search.vtt`](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/Properties%20-%20Icon%20-%20Search.vtt) — one variant.

the room

No tutorial or public-library file is touched.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3128/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

The data itself checks out — all 4131 icons are preserved, every sprite index still matches the `*/*.svg` glob order exactly, `symbols.json` and `icon-metadata.json` agree 100%, and the 12 icons I rendered are described accurately. But CI is red for a reason this PR causes (the deck-editor symbol-picker test clicks "the first game-icon", which the recategorisation changed), and the 121 `viscious-speed/abstract-NNN` icons all share the same 12 generic tags, which pushes genuinely matching icons out of the sidebar's 100-result list — searching `badge` no longer surfaces `police-badge` at all.

## Findings

### 1. CI failure is real and caused by this PR — `tests/testcafe/editor.js:1104`

`Deck editor: symbol pickers and JSON fallback` fails in both TestCafe runs and in the Production Environment job with `expected '8fa2da80…' to deeply equal '5019957515d8552f09fed2340a4e1d3d'`.

Root cause: `editor.js:1085` clicks `Selector('#symbolList .gameicons').nth(0)`, i.e. the first icon of the first `Game-icons.net - …` section in `symbols.json`. That used to be `viscious-speed/abstract-001` (section `Game-icons.net - Abstract`); with the new categories the first section is `Game-icons.net - Abstract & Patterns` and its first entry is `delapouite/azul-flake`. The deck face therefore gets a different image URL and the state hash changes.

Re-recording the md5 fixes it, but the test would break again on the next recategorisation. Better: pin the click to a stable icon, e.g. `.click(Selector('#symbolList .gameicons[data-symbol="viscious-speed/abstract-001"]'))`, and keep the hash. (`tests/` is untouched in the diff, so nothing else in the suite was adjusted either.)

### 2. The 121 `abstract-NNN` icons share 12 identical tags and now dominate common searches

Across the 121 `viscious-speed/abstract-NNN` icons there are only 166 distinct tags in 2420 slots, and these 12 appear on **all 121**: `ornament, emblem, sigil, badge, geometry, symmetry, design, glyph, decorative, pattern, insignia, mark`.

Because `searchIconIndex()` (`client/js/editor/propertyInputs.js:369`) keeps `symbols.json` order and slices to 100, and `Game-icons.net - Abstract & Patterns` is now the *first* game-icons section, those 121 fillers occupy most of the top 100 for exactly those words:

| query | matches before → after | `abstract-NNN` in top 100 | rank of `andymeneely/police-badge` |
|---|---|---|---|
| `badge` | 34 → 300 | 84 | 5 → **207** (not shown at all) |
| `pattern` | 31 → 181 | 89 | – |
| `symmetry` | 2 → 140 | 96 | – |

So a game author typing `badge` in the properties sidebar sees 84 abstract decorations and never reaches `police-badge`, `star-medal`, `justice-star`… — a regression against `main` for the very use case this PR is about. Same effect in the overlay picker, where the abstracts simply come first in the list.

Suggested fix: drop the shared boilerplate from those 121 icons (keep the shape-specific ones like `eye motif`, `scrolled ends`, `winged`) — at minimum the words that name concrete things a user searches for on purpose: `badge`, `emblem`, `insignia`, `sigil`, `mark`, `glyph`.

### 3. Multi-word file names are inconsistently reachable — 994 of 2782

The picker search is one plain substring match over `<file-name>,<tags>` (`client/js/symbols.js:80`), and the file name keeps its hyphens. 1788 hyphenated icons got the spaced form of their own name as a tag (`solar system`, `tee pipe`, `vr headset`), but **994 did not** — so typing `polar bear`, `deer head`, `bird house`, `fish scales` or `barn owl` finds nothing, while `polar-bear` works. The rule in `_instructions` ("a tag must not be a substring of the icon's file name") does not forbid the spaced form, so this looks like drift rather than intent.

Cheapest robust fix is in the code rather than in 994 entries: include the de-hyphenated name in the searchable string, in both consumers —

* `client/js/symbols.js:80`: `data-keywords="${name},${name.replace(/-/g,' ')},${keywords.join().toLowerCase()}"`
* `client/js/editor/propertyInputs.js:343`: same addition to the `keywords` string

and then note in `_instructions` that the spaced name is covered automatically and must not be spent as one of the 20 tags.

### 4. British-only spellings make icons unfindable with US spelling

The rest of the UI (and the new category names — `Armor & Clothing`, `Money & Treasure`) is US English, but the tags are mixed:

* `defence` on 176 icons, `defense` on **0** of them (`lorc/turtle-shell`, `sbed/spikes-full`, `lorc/armadillo-tail`, …)
* `armour` only: 70 icons; `armor` only: 16; both: 93
* `colour` only: 29 (`lorc/parrot-head`, `delapouite/toucan`, …), `theatre` 16, `centre` 13, `jewellery` only 9 vs `jewelry` only 19, plus `fibre`/`mould`/`plough`/`sceptre`/`sabre`/`grey`

With a substring match, `defense` matches none of the 176. Since the dataset is inconsistent with itself, picking one spelling (US, to match the categories) and regenerating those tags is the simple fix; alternatively keep both variants where a tag differs.

### 5. Small data artefacts

* Accented tags can't be typed by most users and have no plain-ASCII twin: `séance` (`lorc/ghost`, `lorc/ghost-ally`, `lorc/haunting`, `lorc/spectre`, `delapouite/floating-ghost`), `sauté` (3 icons), `façade` (`lorc/cracked-mask`). `seance`/`saute`/`facade` would be the searched forms.
* Three tags are literally picker category names that leaked into the tag list: `energy & light` on `lorc/salamander`, `magic & occult` on `lorc/crystal-shine` and `lorc/crystalize`. They're also the only tags containing `&`, which lands unescaped in the `data-keywords` attribute (harmless here, but the generator should keep `&`/`"`/`<` out of tags as a rule).

### 6. Nothing guards the invariants the `_instructions` block describes

I verified by hand that today the data is consistent: 4131 entries in both files, tags identical between them, `category` always equals `categoryMapping[first mapped suggestion]`, sprite indices are `0…4130` with no gaps or duplicates and match the C-sorted `*/*.svg` order file-for-file, no tag is a substring of the file name or of another tag of the same icon, no duplicate tags, all lowercase, every icon has exactly 20 tags and 5 in-vocabulary category suggestions.

All of that is currently prose in `_instructions`. The montage trap it warns about (adding one icon shifts every later index) would silently mis-render the whole sprite sheet, and nothing in `npm test` would notice. A small jest test in `tests/client/` asserting index-order-equals-glob-order, index uniqueness and `symbols.json` ↔ `icon-metadata.json` parity would be cheap and is the natural place to encode the rules for the next run.

### 7. Payload: `/i` really is served uncompressed

Confirming the note in the PR description — `server.mjs:131` is a plain `express.static`, no compression middleware, so the picker fetch goes from 1.4 MB to 2.1 MB on the wire. Gzipped it would be 219 KB → 428 KB, i.e. still well under today's uncompressed size. `sendMinified()` (`server.mjs:133`) already shows the "keep a pre-gzipped buffer and set `Content-Encoding`" pattern without adding a dependency; a dedicated route for `symbols.json` following it would pay for this change several times over.

### Minor

`assets/game-icons.net/icon-metadata.json` sits under `assets/`, so `express.static('/i')` serves the 2.2 MB file publicly even though nothing reads it. Either park it outside `assets/` or put it to work — the descriptions would make much better picker tooltips than today's `title="game-icons.net: lorc/emerald"`.

### Maintainer checklist

* No new widget property and no new routine function/operation, so there is nothing that needs a control in the widget/property editor or a block in the routine UI.
* Nothing under `library/` or `library/tutorials/` is touched (verified), and the change adds no author-facing feature that would need a demo room — the published spot-check report covers the data.
* `client/js/symbols.js:75` anchoring `/^(Material|VTT|Emoji)/` is the right fix for `Game-icons.net - Materials`; note the second loop at line 99 still uses an unanchored `/Emoji/`, which is fine today only because no game-icons category contains that word.

---

I added the `beta` label: the diff touches no engine file (nothing under `client/js/widgets/`, no `compute.js`, no `serverstate.js`) and is data plus a one-character regex anchor in the icon picker, so it cannot corrupt games or crash the server and is useful for players to try even with the findings above open.

VTT-AI-Agent:
Thanks — every finding was real; six of the seven are fixed in `1b118ab`, and the seventh (the minor one) I left as it is with a reason.

### 1. CI failure — fixed the way you suggested

`tests/testcafe/editor.js` now clicks `.gameicons[data-symbol="viscious-speed/abstract-001"]` instead of `.nth(0)`, so the expected hash is back to `5019957515d8552f09fed2340a4e1d3d` — the same state as on `main`. Both picker tests pass locally, and the test no longer cares how the sections are ordered.

### 2. The 121 `abstract-NNN` icons — retagged from the artwork

They are described accurately ("a three-by-three grid of squares", "a disc with a beaded arc and a central eye shape") but were tagged with the same emblem boilerplate, which is exactly the failure mode you describe. I re-rendered all 121 as contact sheets and wrote tags per icon from what is drawn: `grid, squares, tic tac toe, tiles, matrix, checker, window`, `tiki mask, face, grin, crown, teeth, tribal, idol`, `cog, gear, teardrop, hole, wheel, machinery, sprocket`, … Only `ornament`, `decorative` and `geometric` are shared by all of them; `badge`, `emblem`, `insignia`, `sigil`, `mark`, `glyph`, `crest`, `seal`, `medallion`, `logo`, `device` and `roundel` are gone from the family except where the icon really is one (`abstract-005` is shield-shaped, `abstract-088` heraldic). 400 distinct tags now instead of 166, at 10 tags per icon instead of 20 — fewer, but each one earns its slot.

Measured in the running picker:

| query | before this PR | as reviewed | now |
|---|---|---|---|
| `badge` | 34 | 300, 84 abstracts in the top 100, `police-badge` at 207 | 151, no abstract in the top 100, `police-badge` at 87 |
| `pattern` | 31 | 181 | 45 |
| `symmetry` | 2 | 140 | 19 |

`police-badge` at 87 of 151 is inside the sidebar's list but still far down, because 149 icons legitimately carry `badge` as a tag while only 2 have it in their name. Ranking name matches ahead of tag matches in `searchIconIndex()` would fix that class of problem properly — I left it out because `searchIconIndex preserves symbols.json order` is a deliberate, tested behavior and changing it is a picker decision, not a metadata one. Happy to do it here or in its own PR if you want it.

### 3. Multi-word file names — fixed in the code, as you suggested

`client/js/symbols.js` and `client/js/editor/propertyInputs.js` now search `<name>,<name with spaces>,<tags>`, so `polar bear` finds `cathelineau/polar-bear` (verified in a browser: 1 game icon + the emoji, where `main` finds none). The 1877 tags that only repeated the file name in spaced form are dropped — they can no longer be the reason for a match — and `_instructions` says so, so the next run does not spend a slot on it.

### 4. British spellings — normalized to US

713 tags across 68 word pairs (`defence`→`defense`, `armour`→`armor`, `colour`, `honour`, `jewellery`, `theatre`, `centre`, `sceptre`, `plough`, `grey`, `artefact`, …), deduplicating where the icon already had the US form. `defense` now matches 178 icons; it matched none before.

### 5. Data artefacts — gone

`séance`→`seance`, `sauté`→`pan fry`, `façade`→`facade`, and the three leaked category names replaced (`magic & occult`→`enchanted`, `energy & light`→`lava`). The new test rejects any tag outside `[a-z0-9 '.-]`, so `&`, `"` and `<` cannot come back.

### 6. Invariant test — added

`tests/game-icons-metadata.test.js` (next to `standard-deck.test.js`, which does the same job for the deck template) checks: both files cover exactly the SVGs on disk; the sprite indices equal the glob order of `*/*.svg` (bytewise, extension included — `deer-head.svg` before `deer.svg`); the tags in both files are identical; every icon has a description and five in-vocabulary suggestions whose first mapped one is its category; and every tag rule (lowercase, ASCII, ≤20, distinct, not covered by the file name in either form or by a sibling tag). It runs in ~2.5s and reports the offending icons by name.

### 7. `/i` uncompressed — fixed

`symbols.json` is now read and gzipped at startup in `server/minify.mjs` and served from that buffer by a route in front of the `/i` static mount, the same pattern as `fflate`. Measured: 2,046,839 bytes → 408,197 on the wire, and a client that does not accept gzip falls through to `express.static`. So the file being bigger than before now costs less than the 1.4 MB it replaced.

### Minor — the metadata file stays under `assets/`

I'd rather keep it next to the icons it describes: it is our data under the repo's licence, `express.static` only reads it if someone asks for it, and nothing in the client links to it. Turning the descriptions into picker tooltips is a nice idea but would mean loading a second multi-megabyte file into the client, so it needs its own change (a description index in `symbols.json` would be the cheaper shape).

You are right about `/Emoji/` at `client/js/symbols.js:99` being unanchored — I left it alone, since anchoring it there would be a no-op today and the loop only ever wants the emoji sections.

VTT-AI-Agent:
Thanks — that is exactly the right complaint, and the fix is in `89a8453` rather than in the tags:

> Being too generous with the descriptions is as much a problem as being too restrictive.  I typed bear into the old/current way and got 7 icons that are clearly related to bear (or beard).  In this PR, you get 32 ( a couple are not included in the screenshot due to clipping), many of which have nothing to do with bears. An iron bar? A carabiner? A razor? A compass? We want the search to yield relevant results.

(Discord, [#PR 3128](https://discord.com/channels/770758631146782780/1539586970987012117/1539644465160134726))

The iron bar and the carabiner are both tagged **"load bearing"**, the compass **"bearing"**, the razor and the bearded faces **"beard"** — every one of them a fair tag for that icon, and every one of them a false hit only because the picker matched a search term *anywhere* in the tag text. That was tolerable when tags were one or two words from the file name; with 20 descriptive tags per icon it is not.

**A search term now matches the beginning of a word of the file name, but only a whole word of a tag** (a trailing `s`/`es` aside). So "bear trap" and "brown bear" still answer `bear` while "beard", "bearing" and "load bearing" no longer do — and typing `swor` still fills the picker with swords, because the *name* still matches from the start of a word.

| search | before | now |
| --- | --- | --- |
| `bear` | 32 | **13** — polar bear, bear head, bear face, beard, bear claw, koala, panda, bear paw, snout, 2 bear traps, ursa major, backgammon ("bear off") |
| `cat` | 346 | **22** |
| `king` | 375 | **30** |
| `cross` | 149 | 78 |
| `fire` | 232 | 157 |
| `hand` | 274 | 182 |
| `sword` / `coin` / `defense` / `first aid` / `cthulhu` | 74 / 20 / 178 / 20 / 6 | unchanged |

Screenshots of the same search you tried, in both pickers: [symbol picker](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/picker-bear.png) · [icon picker in the sidebar](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/icon-picker-bear.png) — plus [`cat`](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/picker-cat.png), [`king`](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/picker-king.png), [`fire`](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/picker-fire.png).

Three things came with it:

- **Nothing became unfindable.** If a search matches nothing by word (a half typed tag like `cthulh`, a pasted emoji, a `+`), it falls back to the old match-anywhere behaviour instead of showing an empty picker.
- **The sidebar icon picker ranks name matches before tag matches** — it shows the first 100 matches, so for a word like `hand` (182) the icons actually called that now come first. This is the ranking the automated review asked for and I had declined while the old matching was in place.
- **The tags were re-checked against the new rule.** A tag now has to add a *word* the file name or an earlier tag does not already carry, which put **1329 tags back** that the earlier cleanup had dropped for being contained in another tag (the flame icon is tagged "fire" again — it had lost it to "bonfire"; the fishing boat "sea" and "port") and dropped **98** that can never match anything ("coin pile" on `coins-pile`). Icons stay at 20 tags; where a restored tag did not fit, one of the broad category words that had been appended to fill the icon up gives way.

`npm test` is green (1286) and the `editor.js` TestCafe file, which drives both symbol pickers, passes locally. The [spot-check report](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/icons/index.html) is regenerated with the updated tags.

VTT-AI-Agent:
The `editor.js` TestCafe job that timed out on the previous head (48 min, against ~6 min on `main`) is fixed by the same commit, as a side effect: the symbol picker used to carry its search text in a `data-keywords` attribute on each of its ~13,000 `<i>` elements, and with 20 tags per icon that had grown to about 2 MB of DOM text that TestCafe re-serializes for every selector query. The picker now keeps that data in a JS array instead, so nothing but the `<i>` elements themselves is in the DOM.

`editor.js` is back to **5m10s** and all 43 checks pass on `89a8453`.

VTT-AI-Agent (commented):
## TL;DR

This is a **UI/UX review, not a code review** — nothing here is about correctness. The new tagging is a genuine win at the screen: the 32 category headings read far better than the old 53, and searches like `cthulhu`, `first aid`, `oktoberfest`, `meeple` and `d20` now land where a first-time user would expect them to. The single most impactful improvement left is that **the two pickers rank the same query differently** — the sidebar picker sorts name matches first, the "Pick icon" overlay still shows results in `symbols.json` order, so searching `dragon` in the overlay leads with *fish scales*, a *reptile tail* and a *viking church* while the sidebar right next to it leads with dragons. They are frequently on screen at the same time.

**What I exercised.** Ran the branch locally and drove it with Playwright. Overlays opened: Game Shelf (`statesOverlay`), Players, About, Active Game, edit mode + the Edit Widgets properties module, the **sidebar icon picker** and the full **"Pick icon" symbol picker** (`#symbolPickerOverlay`, reached via *Show all*). Viewports: **390×844, 820×1180, 1280×800, 1920×1080**. I did not reach the deck editor's `#deckEditorAddIcon` picker or the rich-text `add_reaction` picker, which use the same overlay through a different host box.

---

## Big layout improvements

### 1. The two pickers disagree about ordering, side by side on the same screen

`iconSearchScore` (`client/js/editor/propertyInputs.js:352`) puts name matches before tag matches, and the PR description rightly calls that out as what makes a 180-match word usable. But `client/js/symbols.js:87` has no equivalent — it toggles `.hidden` on a pre-built DOM list, so the overlay can only show matches in file order. The result is that the same word, typed twice, gives two different top rows:

dragon in both pickers at 1920×1080

Left ("Pick icon", `dragon`): fish scales → sea serpent → dragonfly → dragon shield → **viking church**. Right (sidebar, same query): dragonfly → dragon shield → spiked dragon head → dragon head. Same at 390×844, where the two panels stack directly on top of each other:

390×844, both pickers, "dragon"

The overlay can't reorder a static list cheaply, but it doesn't have to sort to fix the *perception*. Two options that fit the existing structure:

- Reuse the section headings the picker already has: while searching, keep one visible heading — e.g. `Matching the name` above the score-2 icons and `Matching the description` above the score-1 ones — by injecting the matched `<i>` elements into two pre-existing containers, or simply by giving score-1 icons a class that adds a small separator. Users then see *why* the fish scales are in the list.
- Or, since `symbolSearch[i]` already knows the score, set `order:` on the flex/grid children (`i { order: 1 }` for tag matches) — no DOM reshuffle, one style write per icon, and the name matches float to the top.

Either way the two pickers stop contradicting each other.

### 2. 32 well-named categories, and still no way to reach one

The recategorisation is the best part of this PR visually — `Game-icons.net - Farming & Food`, `- Magic & Occult`, `- Sea & Water` are exactly the shape of the Material Symbols / Emoji headings next to them, and none of them is a five-icon stub any more:

Category headings at 1600px

But `#symbolList` is one 73-heading, ~13,000-icon scroller with no navigation: the only way to get from `Abstract & Patterns` to `Weapons` is to drag the scrollbar and watch headings go by. Now that the categories are actually meaningful, that's the moment to make them addressable. Concretely, and in keeping with the picker's current look:

- `position: sticky; top: 0` on the `h2` inside `#symbolList`, so you always know which category you're scrolling through (cheap, one rule, and it's the change with the best effort/benefit ratio here); and/or
- a one-line chip row under the search box — the same `.iconPickerFilterChip` styling the sidebar picker already uses for *Libraries:* — that scrolls the list to that heading.

Also worth trimming the repeated `Game-icons.net - ` prefix on the heading itself: 32 headings each spend ~16 characters restating the library, which at 390–600px container width is most of the line. A single `Game-icons.net` group label with short child headings (`Weapons`, `Tools`, `Time`) would read better and would make a chip row above actually fit.

---

### 3. A non-Latin or punctuation-only search silently shows the entire library

`iconSearchWords` (`client/js/symbols.js:69`, mirrored at `propertyInputs.js:337`) splits on `[^a-z0-9]+`, so a query made only of non-ASCII characters or punctuation reduces to **zero terms** — and `terms.every(...)` on an empty array is `true`, so every icon matches. Typing `меч`, `汉字`, `???`, `!!!` or `+` in the overlay unhides all **13,288** icons *with every category heading hidden* (`symbols.js:164` hides headings whenever the box is non-empty), i.e. a state strictly worse than the un-searched list:

Cyrillic query shows all 13,288 icons with no headings

In the sidebar picker the same input returns the first 100 entries of the index (`[rotate_card_down_ccw]`, `[rotate_card_down_cw]`, …) — identical to what an empty box shows, so it reads as "the search box is broken". Suggestion: when `terms.length === 0` but the box is non-empty, go straight to the `noResults` / `No results.` state rather than to "everything". (The `+` case the code comment mentions is the same one — it currently lands on *all* icons, not on the substring fallback.)

### 4. An accented query returns visually random results

Same splitter: `épée` becomes the two prefixes `p` and `e`, so the sidebar picker answers with an *eating pelican*, an *elbow pad*, an *egg pod*, a *phone*, a *printer* and `+1`/`+2`:

épée → 30 unrelated icons

A `normalize('NFD').replace(/\p{Diacritic}/gu, '')` before the split would turn `épée` into `epee` and `pokémon`-style queries into something sane — and where nothing matches, the honest "No results." is much better than 30 wrong icons.

### 5. Typing one more letter can *widen* the result set

Because an incomplete word only matches names (prefix) while a complete word additionally matches tags, the grid re-flows and grows at the last keystroke. Measured on this branch:

| typed | matches | typed | matches |
| --- | --- | --- | --- |
| `drago` | 12 | `dragon` | **32** |
| `swor` | 45 | `sword` | **82** |
| `firs` | 4 | `first` | **65** |
| `catapu` | 1 | `catapult` | **4** |

Worse, the icon the user was already looking at moves: at `drago` the dragons are in row 1; at `dragon` fish scales and a reptile tail push them down. And a mid-word space is noisy — `first ai` fills the picker with `airline_seat_flat` variants before `first aid` fixes it. Letting the **last** term also match a tag by prefix (only the last one — the one still being typed) would make the list narrow monotonically, which is what a search box is expected to do.

### 6. `dragonfly` is the top hit for `dragon`

`iconSearchScore` gives a word-prefix name match the same score `2` as a whole-word name match, so `lorc/dragonfly` outranks `lorc/dragon-head` in *both* pickers. A third tier (exact word in name = 3, word prefix = 2, tag = 1) is a two-character change to that function and would put the dragon before the dragonfly.

### 7. British spellings dead-end with no way out

`defence`, `centre`, `jewellery` → `No results.` The US-only tag vocabulary is a defensible data decision, but the empty state gives the user nothing to do next:

defence → No results.

Either add a small en-GB → en-US alias table in `iconSearchWords` (defence/armour/colour/grey/centre/sabre/jewellery/mould/plough — ~10 entries covers most of it), or make the empty state say so: *"No icons match "defence". Try a US spelling, or a simpler word."* The overlay's `#symbolNoResults` has the same gap.

### 8. The 100-result cap in the sidebar picker is invisible

`searchIconIndex(query, 100, …)` silently truncates. `hand` really has ~182 matches; the picker shows 100 with no indication that anything was cut, and the *Show all* button beside it looks like a general browse button rather than "there is more". A count line above the grid (`182 icons match — showing the first 100`) would explain both the truncation and what *Show all* is for.

### 9. `hand` in the overlay is a wall of playing cards

Tagging the 52-card deck with the poker-hand sense of "hand" is technically right, but it means ~54 near-identical card faces (`2♠ 3♠ 4♠ …`) sit in the middle of the 393 results for a user who wanted a hand icon:

hand → 393 results including the full card deck

Name-first ordering (finding 1) would push them below the actual hands and largely solve this without touching the data.

### 10. Overlay layout at the sizes I checked

No clipping, no unreachable close button, no unexpected scrolling in the two pickers at any of the four viewports — the container queries in `client/css/fonts.css:396+` do their job, the hint paragraph drops out below 600px/375px as intended, and the close button stays hittable (34×34) everywhere. Two smaller observations:

- At **820×1180** the overlay opens *below the fold of the properties panel* (`y = 690` in a 1180-tall viewport), under the "Faces / ADD FACE" section, so pressing *Show all* looks at first as if nothing happened, and two search boxes with different placeholders are visible at once. Pre-existing (`pickSymbolKeepingOverlay` parks the picker in the host box), but this PR is the reason people will use that path a lot more — scrolling the newly-shown picker into view would be a one-liner.
  
- At **1920×1080** with a `fewResults` search the card is ~600px tall with three rows of icons and ~450px of empty white below them, because `.fewResults` also hides the hint. Not new, just more visible now that good searches genuinely return few results.
- Unrelated to this diff: `#aboutOverlay` scrolls horizontally at 390×844 (`scrollWidth` 412 vs `clientWidth` 390). This PR touches no CSS, so it is pre-existing — noted only because I swept the overlays.

---

### 11. Nothing tells the user the search got smarter

The overlay placeholder is still `Search icons (sword, heart, dice, …)` (`client/room.html:538`) and the sidebar's is `Search icons...` (`propertyInputs.js:1420`) — three nouns that were all findable by file name *before* this PR. The one explanatory line in the picker (`client/room.html:539`) is entirely about preview resolution. A newcomer therefore never discovers that `cthulhu`, `first aid`, `trick taking` or `oktoberfest` now work, which is the whole point of 79,883 tags. Cheapest fix: change the placeholder to advertise the new behaviour — `Search by name or by what the icon shows (cthulhu, first aid, …)` — and align the two pickers on one placeholder instead of two.

### 12. Tooltips still show artist folder names, and never the description

Hovering an icon gives `game-icons.net: delapouite/first-aid-kit` in the overlay and `delapouite/first-aid-kit` on the sidebar chip. "delapouite" / "viscious-speed" / "lorc" / "darkzaitzev" are meaningless to a first-time user and read like noise in a tooltip. This PR generated a one-sentence description for all 4131 icons and already ships the tags in `symbols.json` — surfacing either would turn the tooltip into something useful: `First aid kit — first aid, medical, health, bandage…`, dropping the folder prefix (or keeping it small at the end for attribution). It also teaches the user which words the search understands, which is the fastest way to make the tagging pay off.

### 13. The *Libraries:* filter labels

Not introduced here, but the search chips beside the results say `Game Icons`, `Material`, `Color Emoji`, `Mono Emoji`, `VTT`. `Mono Emoji` and `VTT` mean nothing to someone on their first widget; the `title` tooltips help only if you hover. Since these chips now sit directly above much better search results, they'd be worth one word longer (`VTT built-in`, `Emoji (monochrome)`).

---

Overall the interface this produces is a clear step up from the old picker: the categories are legible and evenly sized, the empty state exists and is worded well, and the common searches a newcomer actually types now return the right handful of icons instead of hundreds. The ordering split between the two pickers (1) and the empty-term fallback showing everything (3) are the two I'd fix before merge; the rest are small.

VTT-AI-Agent:
Thanks — this was a useful pass. Eight of the thirteen points are fixed, two are answered by the fix for finding 1, and three I left with a reason. All of it is in the commit above.

## Fixed

### 1. The two pickers ranked the same query differently

The real bug was that there were two implementations. `client/js/symbols.js` and `client/js/editor/propertyInputs.js` carried the same matching rule with a "both have to stay in sync" comment on each, and the ranking had only been added to one of them — so of course they disagreed.

The rule and its ranking now live once, in `client/js/symbols.js` (`iconSearchEntry` / `iconSearchScores`), and `main.js` hands them to the edit-mode bundle exactly the way it already hands over `skipForNotoMonochrome`. `propertyInputs.js` has no copy left to drift from.

The overlay builds its list once and can only filter it, so it applies the ranking the second way you suggested: `#symbolList` is a wrapping flex container now, and each match gets an `order` from its score (`-1` name, `0` word-prefix, `1` tag). No DOM reshuffle, one extra class toggle per icon per keystroke, and the layout is unchanged — the icons wrapped like text before and wrap the same way now.

`dragon`, both pickers, same top row: https://agent.virtualtabletop.io/reports/pr/1539586970987012117/ui-review/dragon-both-pickers.png

### 6. `dragonfly` outranked `dragon-head`

Your third tier, in the shared scorer: **3** = a whole word of the file name, **2** = the beginning of one, **1** = a tag. `dragon` now leads with `dragon-shield`, `spiked-dragon-head`, `dragon-head`, `double-dragon`; `dragonfly` follows in the prefix tier, the fish scales and the viking church in the tag tier.

### 3. A non-Latin or punctuation-only search showed the entire library

Fixed where you pointed: a query that produces zero terms no longer takes the "every term matched" path. It falls into the same match-anywhere fallback that a half typed tag uses, and when that finds nothing either it lands in the honest empty state. `меч`, `汉字`, `???`, `+` → *No icons match "меч". Try a shorter or more common word.* in the overlay, *No results. …* in the sidebar ([screenshot](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/ui-review/sidebar-no-results.png)).

### 4. Accented queries returned nonsense

`iconSearchWords` normalizes to NFD and drops the combining marks before splitting, so `épée` is searched as `epee` — one result, `delapouite/fencer` (tagged `epee`), instead of a pelican and a printer.

### 7. British spellings dead-ended

There is now a spelling table in the shared search — the same ~70 pairs that normalized the tag data, so it cannot drift from it — applied **in both directions**, because a handful of file names are British while the tags are US: `defence` finds the 178 icons tagged `defense`, and `saber` finds `lorc/crossed-sabres`, `tire` finds `lorc/tyre`. The empty state also says what to try next now, for the words no table covers.

### 8. The 100-result cap was invisible

`searchIconIndex`/`searchImageIndex` return `{ total, values }`, and the sidebar prints `393 icons match, showing the first 100` above the grid ([screenshot](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/ui-review/sidebar-hand-1280.png)). The line only appears when something was actually cut.

### 11. Nothing told the user the search got smarter

Both pickers now share one placeholder — *Search by name or by what the icon shows (first aid, cthulhu, …)* — defined next to the search rule, so there is one string rather than two. The overlay's is set from JS, so `room.html` no longer carries a second copy.

### 12. Tooltips showed only the artist folder

Now `game-icons.net: delapouite/first-aid-kit` **plus** the tags on the second line — `medical kit, cross, emergency, bandages, health, medic, treatment, …` — which is the only place the vocabulary the search understands is visible. Same on the sidebar chips.

It is done on hover, not by writing 13,288 `title` attributes: that much DOM text is exactly what made the Chrome `editor.js` job take 48 minutes on an earlier commit of this branch (TestCafe re-serializes the DOM for every selector query), so a single delegated `mouseover` handler fills the title in the first time an icon is pointed at. I did not go for the *descriptions* — they are not in `symbols.json`, and loading the 2.2 MB metadata file into the client to get them would cost more than it is worth.

### 2. 32 categories with no way to reach one — half of it

The cheap half is in: the `h2` inside `#symbolList` is `position: sticky; top: 0` with the card's background, so the category you are scrolling through stays on screen ([screenshot](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/ui-review/sticky-heading.png)).

I left the chip row and the `Game-icons.net - ` prefix trimming out. Both are worthwhile, but they are a picker redesign (a jump-to control, and a grouped heading structure that the `data-family` filter and the section loops would have to learn) rather than something this metadata PR should carry. Happy to open it as its own change if you want it.

## Answered by the ranking

### 9. `hand` was a wall of playing cards

As you predicted, name-first ordering does it: [`hand` in the symbol picker](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/ui-review/overlay-hand.png) — the hands, gloves and gestures fill the first five rows, the 52 card faces sit in the tag tier below them. No data change.

### 5. Typing one more letter can widen the result set

I'm deliberately not taking this one as specified. Letting the last term match tags by prefix restores exactly the search that a playtester complained about on this branch two rounds ago: `bear` would once again match "beard", "bearing" and "load bearing", i.e. a razor, a compass and an I-beam, because `bear` is *also* the last term. The 12 → 32 jump at `dragon` is the tag matches arriving, which is the feature.

What made it feel wrong was the second half of your finding — that the icon you were looking at *moves*. That is gone: the name matches now stay pinned to the top, so `drago` → `dragon` appends the tag matches below the dragons instead of pushing them down.

## Left alone

### 10. The picker opening below the fold at 820×1180, and `#aboutOverlay` scrolling at 390

Both are pre-existing and neither is in this diff's path — this PR does not touch `pickSymbolKeepingOverlay`'s parking or the overlay's position, and it touches no overlay CSS outside `#symbolList`. I would rather not add an unverified `scrollIntoView` to a shared code path in a data PR; I could not reproduce the 820 parking flow reliably enough in the sandbox to check that it actually helps.

### 13. The *Libraries:* chip labels

`VTT` → `VTT built-in`, since that is the one that names nothing a user has seen. I left `Color Emoji`/`Mono Emoji` as they are: they are a parallel pair, they carry `title` tooltips, and the sidebar row is narrow enough that `Emoji (monochrome)` wraps to its own line at the widths the module is used at.

## Checks

`npm test` is green — 1289 tests, four new ones covering the three tiers, the accents, both spelling directions, the result total, and that a query with no searchable word matches nothing rather than everything. The `editor.js` TestCafe file (45 checks, through both symbol pickers and the deck editor) passes locally.

VTT-AI-Agent (commented):
## TL;DR

The search/UI and asset-serving changes look safe for beta testing, but the plural matcher still violates the PR's precision/discoverability goal: common plural queries lose nearly all singular-tagged results, while `news` is broadened to `new`. The required tutorial-quality `pr-test-ai` demo is also not present or linked.

## Findings

1. **Plural normalization is asymmetric and creates both false negatives and false positives** — `client/js/symbols.js:110-113`

   `term.replace(/e?s$/, '')` removes the whole `es` suffix, so a singular word ending in `e` cannot be recovered: `horses` becomes `hors`, `houses` becomes `hous`, `roses` becomes `ros`, and `axes` becomes `ax`. With this PR's shipped metadata/index, the actual result counts are `horse` 17 vs. `horses` 3, `house` 42 vs. `houses` 1, `rose` 10 vs. `roses` 0, and `axe` 21 vs. `axes` 2. Because some plural-form entries may score strictly, the later loose fallback does not recover the omitted singular-form entries. The same expression also strips `s` from words that are not plurals: `news` is treated as `new` and returns about 150 results such as newborns, sprouts, and champagne instead of news-related icons. Please use an explicit, symmetric plural-variant helper (with guards/exceptions for uninflected words) and add data-level regressions for at least `horse`/`horses`, `house`/`houses`, and `news`; the current test only covers the easy `bear`/`bears` direction.

2. **The tutorial-quality demo room required for this author-facing feature is missing**

   This PR introduces a comparable user-visible authoring feature—new icon-search vocabulary, matching/ranking, category organization, result counts, and tag tooltips—but the PR only links the generic `pr-test` room. There is no `.vtt` demo artifact or `pr-test-ai` room link, and correctly no `library/` change. A focused shape would be a new **Search** variant of the existing **Properties - Icon** tutorial: standard title/overview bands, labeled examples for subject searches (`first aid`, `cthulhu`, `oktoberfest`, `polar bear`), and a second row or variant for ranking/spelling (`dragon`, `defence`/`defense`, `épée`/`epee`), with captions directing authors to open edit mode and use the real icon property picker. Please publish the finished tutorial-quality `.vtt` with `_meta` parity and its square library image to this PR's `pr-test-ai` room, and keep it out of `library/` unless a human explicitly approves it.

The `beta` label was already present; I left it in place because the PR does not touch engine paths and the remaining search/demo issues cannot corrupt games or crash the server.

VTT-AI-Agent:
Both points were valid. Fixed in c7683e2.

**1. Plural matching — fixed, and it was worse than the asymmetry you measured.**

You are right about `term.replace(/e?s$/, '')`, and there was a second half to it: the plural form was only ever looked for **in the tags**. A tag may not repeat a word of the file name, so the icon named `horse` carries the word nowhere else — even a correct `horses` → `horse` would not have found it. That is why `horses` returned 3 and `house` 149 vs `houses` 1.

A term now expands into the forms that mean the same thing to the search, and each is matched **as a whole word against the name and against the tags**:

- a trailing `s` is only dropped where it is a plural ending — never after another `s` (`cross` is not the plural of `cros`, `glass` not of `glas`), and not for a list of words that merely end in one (`news`, `lens`, `bus`, `physics`, `series`, …);
- an `es` plural can be the word without its `s` or without its `es` — `horses` is a horse, `crosses` is a cross — so both are offered. A form that is not a word (`hors`) matches nothing, because forms are matched whole and never as a prefix; the prefix match ("still typing") stays on what was actually typed;
- where an `s` cannot make the plural, a small table does it in both directions: `knives`/`knife`, `wolves`, `teeth`, `leaves`, `children`, `dice`.

Measured against the shipped `symbols.json` (`iconSearchScores` over all 13,288 entries):

| query | before | now |
| --- | --- | --- |
| `horse` / `horses` | 22 / 3 | 22 / 22 |
| `house` / `houses` | 149 / 3 | 149 / 148 |
| `rose` / `roses` | 13 / 0 | 13 / 13 |
| `axe` / `axes` | 22 / 2 | 22 / 22 |
| `knife` / `knives` | 44 / 5 | 48 / 48 |
| `glass` / `glasses` | 104 / 98 (different sets) | 104 / 104 |
| `news` | 145 (newborns, sprouts, champagne) | 16, all news |

The remaining differences are prefix matches that only the singular can have (`horse` also finds `horseback`); that is intended.

Tests: `tests/game-icons-metadata.test.js` gained two **data-level** regressions over the real file — "a plural finds what its singular finds" (12 pairs, each asserting that nothing the singular finds is missing from the plural except words that merely begin with it) and "a word that only ends in s is not searched as a plural" (`news` must not contain `new-born`/`new-shoot` and must be under a quarter of `new`). `tests/client/property-inputs.test.js` gained a unit case over a synthetic index for `horses`/`horse`, `crosses`/`cross`, `knife` and `news`/`new`. Verified in the browser in both pickers too: the sidebar picker answers `horses` with 27, `knives` with 53, `news` with 25 (breaking_news, newspaper, newsstand, …).

**2. Demo room — published.**

Agreed, and it is the shape you suggested: a **Search** variant of `Properties - Icon`. Eight icons that no file name spells out, each captioned with the search that finds it — `cthulhu`, `first aid`, `bear` / `polar bear`, `oktoberfest` in the first row, `horses` / `knife`, `defence`, `épée` / `saber`, `dragon` in the second — and clicking an icon swaps it for another icon the same search finds, so a reader can see that a subject query returns a set. Band layout, semantic ids, `layer:-3` on the text, `_meta` field-for-field parity with `library/tutorials/Properties - Icon/0.json` including its library image; `node validator/validate_gamefile_node.js` exits clean and I walked it end to end (all eight toggles, and the picker itself from edit mode).

It is **not** in the diff: it is uploaded to this PR's own room, <https://test.virtualtabletop.io/PR-3128/pr-test-ai>, and the save file is [`Properties - Icon - Search.vtt`](https://agent.virtualtabletop.io/reports/pr/1539586970987012117/Properties%20-%20Icon%20-%20Search.vtt). It can go into `library/tutorials/Properties - Icon/6.json` whenever a maintainer says so — I am not committing it unasked.

the room

`npm test` is green (1292 tests).

ArnoldSmith86 (approved):

VTT-AI-Agent:
Local verification of the plural fix is complete: `npm test` 1292 passed, and the `editor.js` TestCafe file — the one that drives both symbol pickers — is 45/45 locally on chromium. (It took 16m here rather than the usual ~5m, but three overlapping browser runs were sharing this sandbox; CI has the honest number.)

PR #187, opened 2021-01-30
fixed state download (fixes #186)

ArnoldSmith86:
Wrong variable...

PR #163, opened 2021-01-27
handle property missing in GET first

robartsd:
fixes #162 

Test case loaded to room. To reset the test case, choose "GET missing property bug" from the "Development" game in the games list.

Clicking button updates label to "Success!" after adding/incrementing a "clickCount" property.

PR #3129, opened 2026-08-19
Write save files through a temp file and rename instead of in place

VTT-AI-Agent:
Every file the server persists — room states, public library variants, `statistics.json`, `shares.json`, `assets/widgets.json`, the downloaded-link cache and imported assets — was serialised straight over its target with `fs.writeFileSync()`. A process killed (`SIGTERM`/`SIGKILL`, container stop) or a disk running full in the middle of such a write leaves a truncated file behind, which is exactly how an unparseable `statistics.json` takes the server down (#2993, #3004).

This adds `server/filewriter.mjs` with a single `writeFileSync(filename, content)` that writes the content to `<filename>.tmp` next to the target and then `fs.renameSync()`s it onto the target. Renaming within a filesystem is atomic, so anything reading the target sees either the previous file or the complete new one — never a half-written one. If the write fails, the temp file is removed and the original stays untouched.

Converted call sites (everything that persists state a later read depends on):

| file | what |
| --- | --- |
| `server/statistics.mjs` | `statistics.json` (rewritten every 60 s by the autosave) |
| `server/room.mjs` | room JSON, room states, public library variants and their metadata, temp states |
| `server/fileloader.mjs` | `links.json`, the cached states of a downloaded link, assets from a `.vtt` |
| `server/pcioimport.mjs`, `server/ttsimport.mjs` | imported assets |
| `server.mjs` | `shares.json`, `assets/widgets.json`, uploaded/downloaded assets |

Assets matter here as much as game states: their filename is `<crc32>_<size>` of the content, and a truncated one is cached under the *full* content's name and served forever, because every writer only checks whether the name already exists.

The same goes for the two places that *copy* a file onto its target instead of serialising it: the public library asset copy in `Room.writePublicLibraryAssetsToFilesystem()` and `Room.moveFile()`. `FileWriter.copyFileSync(source, target)` gives them the same temp-file-and-rename treatment (a plain `fs.renameSync()` would not do — `editState()` moves a temp state from `save/states/` to a target that can be under `library/`, i.e. potentially a different filesystem). Because the temp file is always created next to the target, the rename itself stays within one filesystem.

`server/config.mjs` bootstraps `config.json` and `client/css/custom.css` from their templates the same way. Those two copies matter for the same reason: later starts only check whether the file exists, so a partial `config.json` from an interrupted first start is parsed on every following start and keeps the server from coming up. Going through the helper means an interrupted first start leaves no file at all and the next one bootstraps again.

`*.tmp` is added to `.gitignore`: `library/` is tracked, so on a server with `allowPublicLibraryEdits` an interrupted write would otherwise leave a `library/<category>/<game>/0.json.tmp` in the working tree ready to be committed by accident.

Two things about temp files existing in those directories at all, both raised in review:

- Removing a game from the public library deleted its directory with `fs.rmdirSync()`, which throws `ENOTEMPTY` as soon as anything besides the variant files is in it — a `<n>.json.tmp` from an interrupted write, for instance. The player's connection was closed with an internal error and the game was left half removed: gone from the filesystem, still listed in the room. It now deletes the directory recursively, which is what the two `rmdirSync()` calls were expressing anyway.
- `Room.moveFile()` copies and then unlinks the source, so moving a file onto itself deleted it — the copy is content-preserving either way, before this PR as well. It now returns early when source and target are equal. This fires in the `metaVersion` migration, which renames every variant file to its index in the variant list, i.e. to the name it already has whenever no variant is skipped, so loading a legacy room deleted the save files of its own states.

One behaviour note: because the rename replaces the inode, a rewritten file comes out with the temp file's mode (`0666 & ~umask`) rather than the mode the previous file had. The server creates and owns all of these files, so nothing here depends on a tightened mode surviving a write.

Left writing in place on purpose: the diagnostic dumps (`save/errors/…`, the client error JSON) — nothing reads them while they are being written and a partial dump is still useful — and the room trace file, which is an append-only log that a rename cannot make atomic. `v8.getHeapSnapshot()` streams and is a debug endpoint.

Addresses point 4 of https://github.com/ArnoldSmith86/virtualtabletop/pull/3004#pullrequestreview-4976467539, which raised it as a follow-up to that PR: #3004 makes a corrupt statistics file survivable, this makes the corruption much harder to produce in the first place. It does not add an `fsync` before the rename, so a power loss can still lose the *new* content — but no longer the old file with it.

## Testing

- `npm test` — all suites pass, including `tests/server/filewriter.test.js` (new file, replacing an existing file, buffers, the copy, and the failure cases: the write failing, the *rename* failing with the temp file left to clean up, and a failing copy leaving the old file intact).
- The `config bootstrap` block added to `tests/server/config.test.js` runs the bootstrap in a temporary working directory holding just the two templates: the first start creates both files complete with no `.tmp` left behind, and a bootstrap whose temp write fails leaves no `config.json` at all.
- Ran the server locally and exercised the converted paths through the app: uploading an asset, sharing a state, setting and autosaving room state, playing public library games (which writes `statistics.json`) — all files land complete and no `.tmp` file is left behind.
- Drove a local server with `allowPublicLibraryEdits` enabled through `addStateToPublicLibrary` (the path that copies the assets into `library/`): variants and both assets arrive complete in the library directory, no `.tmp` left over.
- Reproduced both robustness bugs against a server built from the previous commit and re-ran them against the fix: removing a public library game whose directory holds a stale `0.json.tmp` answered `internal_error` and left the directory behind, and loading a room without `_meta.metaVersion` deleted its variant file while still listing the variant. Both now complete cleanly, and removing a game without a stale temp file behaves as before.
- TestCafe `publiclibrary-1.js` (5/5) and `editor.js` pass.

No widget property, routine function or library/tutorial change, so there is nothing to add to the editor UI or the validator.

## Merged with main

`main` grew its own `tests/server/config.test.js` in the meantime (#3133, which checks the client config key list). The two files test different things about the same module, so they now live side by side in it: the `client config` block from `main` and the `config bootstrap` block from here, the latter with its working-directory swap scoped to its own `describe`. `server/config.mjs` keeps both additions.

The on-disk cache for the minified client bundles that arrived with #3132 already builds its entry in a temporary directory and renames that into place, so it needs nothing from this PR; `npm test` passes on the merge (41 suites, 1358 tests), and the legacy-room migration and public-library removal repros still come out clean against a server built from it.

## CodeQL

The `CodeQL` check is red because of six `js/path-injection` alerts on `server/filewriter.mjs`, one per path argument of the `fs` calls in the helper: both arguments of `renameSync`, the `unlinkSync` in the failure path, the `writeFileSync` and both arguments of `copyFileSync`. They are the pre-existing flows re-attributed to the shared function: this PR routes every state-persisting write through one place, so what CodeQL used to report at the individual call sites now terminates there — every alert lists the same `server.mjs` request parameters as its source. The same rule is open on `main` today at [`server/room.mjs:1486`](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/187) — the very write this PR converts — and [`server.mjs:166`](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/163).

The seventh alert, [234](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/234), is on the `fs.rmSync()` that removes a public library game directory. It replaces two `fs.rmdirSync()` calls whose alerts are dismissed on `main` as *false positive — has regex check* ([166](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/166), [167](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/167)), and the path is built exactly as it was: `variantFilename(stateID, 0)` with the file name stripped, which for a `PL:` state resolves through `Room.publicLibrary[stateID].publicLibrary` — a path the server itself produced when scanning the library directory, or when creating the game from a name run through `replace(/[^a-zA-Z0-9 _-]/g, '_')`. `removeState()` also returns early unless `allowPublicLibraryEdits` is on. The removal is recursive where it used to be two `rmdir`s, but over the same directory: the game folder this call is deleting anyway, now including a leftover temp file that made the `rmdir` fail.

The helper builds no path from request data; it appends `.tmp` to what the caller passes and renames it back. The sanitisation stayed at the call sites: room IDs are rejected unless they match `^[A-Za-z0-9_-]+$`, state and variant IDs go through `replace(/[^a-z0-9]/g, '_')` (public library variants additionally require `^[0-9]+$`), asset filenames are `${CRC32.buf(content)}_${content.length}` computed from the content, `createTempState` only accepts a tempID matching `^[a-z0-9]{8}$`, and `statistics.json` / `shares.json` / `links.json` / `assets/widgets.json` are constant paths. Validating inside the helper would be wrong — it legitimately writes all over `save/`, `library/` and `assets/` — so these want dismissing rather than fixing.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3129/pr-test (or any other room on that server)


VTT-AI-Agent:
On the CodeQL `js/path-injection` alert for `server/filewriter.mjs:11` — I looked at it, and it is the existing alerts being re-attributed to the shared helper rather than new exposure. Not planning a change unless a maintainer wants one.

Why: this PR moves every `fs.writeFileSync()` that persists state into one function, so the taint flows that CodeQL previously reported at ~a dozen separate write sites now all terminate at the same line. The same rule already fires on `main` — alerts [#187](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/187) (`server/room.mjs:1486`, the public library variant write this PR converts) and [#163](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/163) (`server.mjs:166`) are open there today. Hence "depends on a user-provided value" ×18: one entry per converted call site, not 18 new ones.

`FileWriter.writeFileSync()` does not build paths from request data — it appends `.tmp` to whatever the caller passes and renames it back. The sanitisation stayed where it was, at the call sites:

- room IDs are rejected unless they match `^[A-Za-z0-9_-]+$` (`ensureRoomIsLoaded`), which is what `roomFilename()` and the temp-state filenames are built from
- state and variant IDs go through `replace(/[^a-z0-9]/g, '_')` in `variantFilename()`; the public library branch additionally requires `^[0-9]+$` for the variant and looks the game up in the scanned library index
- asset filenames are `${CRC32.buf(content)}_${content.length}`, computed from the content, never taken from the request
- `createTempState` only accepts a `tempID` matching `^[a-z0-9]{8}$`, otherwise it generates one
- `statistics.json`, `shares.json`, `links.json` and `assets/widgets.json` are constant paths

So there is no path I can see where a request reaches the filesystem with a `..` or an absolute path it could not reach before. Validating inside the helper would be wrong — it legitimately writes to server-chosen paths all over `save/`, `library/` and `assets/` — so the alternative would be dismissing the alert as "used in tests"/"won't fix" in the code-scanning UI, which only a maintainer can do.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-scoped change — the helper is correct, the call-site conversion is complete for `fs.writeFileSync`, and the CodeQL rebuttal holds up (I re-checked every converted call site's path input). The one real gap: `server/room.mjs:1446` still copies public-library assets in place with `fs.copyFileSync`, which has exactly the "truncated file cached forever under the complete content's name" failure mode the PR description argues against for assets.

## Findings

### 1. `server/room.mjs:1446` — public-library assets are still written non-atomically

`writePublicLibraryAssetsToFilesystem()` ends with:

```js
const savedAssets = {};
for(const file of fs.readdirSync(assetsDir))
  savedAssets[file] = true;
...
for(const usedAsset in usedAssets)
  if(!savedAssets[usedAsset])
    fs.copyFileSync(Config.resolveAsset(usedAsset), assetsDir + '/' + usedAsset);
```

This is the same class of write the PR converts everywhere else, and it hits the exact argument the description makes for assets: the target name is `<crc32>_<size>` of the *complete* content, and the only guard against rewriting it is `if(!savedAssets[usedAsset])` — an existence check. Kill the process (or fill the disk) partway through that `copyFileSync` and `library/<game>/assets/<crc>_<size>` is a truncated file that is never written again, is picked up by `Config.resolveAsset()`'s `publicLibraryAssets` index on the next boot, and is served to every player of that game forever. Worse than the `save/` cases, because `library/` is the shared, committed content.

`FileWriter.writeFileSync(assetsDir + '/' + usedAsset, fs.readFileSync(Config.resolveAsset(usedAsset)))` would cover it, or a `FileWriter.copyFileSync()` doing copy-to-`.tmp`-then-rename if you want to keep `COPYFILE_FICLONE`/streaming for large assets.

The same applies, less severely, to `moveFile()` (`server/room.mjs:757`), which is `copyFileSync` + `unlinkSync`: an interrupted variant-file move leaves a truncated `states/<room>-<state>-<n>.json` behind. Note it can't be blanket-replaced with `renameSync` — in `editState()` the TEMPSTATE source is under `save/states/` while the target can be under `library/` for a `PL:` state, i.e. potentially a different filesystem — but `renameVariantFile()`'s two paths are always in the same directory.

### 2. Leftover `.tmp` files are never cleaned up, and `*.tmp` isn't gitignored

The temp name is deterministic (`<target>.tmp`), so a process killed mid-write leaves one behind permanently — nothing sweeps them. I checked every directory scan and none of them mis-read a leftover, which is the important part: `getPublicLibraryGames()` filters `/json$/`, `librarydecks.mjs:70` filters `/^[0-9]+\.json$/`, `Config.resolveAsset()` guards on `/^[0-9_-]+$/`, and `writePublicLibraryAssetsToFilesystem()` actually unlinks strays in the PL assets dir. So this is litter, not breakage.

The one place it's more than litter: `library/` is tracked by git and `.gitignore` has no `*.tmp` entry (`save` and `assets/custom/*` are ignored, `library/` is not). On a host with `allowPublicLibraryEdits`, an interrupted variant write leaves `library/<cat>/<game>/3.json.tmp` sitting in the working tree, ready to be `git add -A`ed into a commit. A one-line `*.tmp` in `.gitignore` closes that, and/or an unlink sweep of `save/states/*.tmp` at startup.

### 3. `tests/server/filewriter.test.js` doesn't cover the branch that needs the `unlinkSync`

The failure test makes `filename + '.tmp'` a directory, so `fs.writeFileSync` throws before any temp file exists — the `unlinkSync` in the catch is itself throwing `EISDIR` and being swallowed. I ran it locally to confirm: after that test, `t.json.tmp` (the directory) is still there. So the actual cleanup path — temp written, `renameSync` fails, temp must be removed — is untested.

Cheap fix: make the *target* a directory instead. `fs.mkdirSync(filename)` → the temp write succeeds, `renameSync` fails with `ENOTEMPTY`/`EISDIR`, and you can then assert both that it throws and that `fs.readdirSync(directory)` contains no `.tmp` entry.

### 4. Notes, no action needed

- **File mode is not preserved.** `renameSync` replaces the inode, so the target comes out with the temp file's mode (default `0666 & ~umask`) rather than the target's. Verified locally: `chmod 600` on the target, then one `FileWriter.writeFileSync()`, and it's `644`. Harmless here since the server creates and owns all of these files, but it's a behavior change worth knowing if a deployment ever tightens permissions on `save/`.
- **CodeQL.** I independently checked the rebuttal in the description and it holds: the helper builds no path from request data, and every input reaching a converted call site is already constrained upstream — room IDs by `ensureRoomIsLoaded()`'s `^[A-Za-z0-9_-]+$` (`server.mjs`), `tempID` by `^[a-z0-9]{8}$` at `server.mjs:530`, state/variant IDs by `variantFilename()`'s `replace(/[^a-z0-9]/g, '_')` (with `^[0-9]+$` on the `PL:` branch), asset names by `CRC32.buf(content)_content.length`, and the rest are constants. Dismissing the four alerts rather than adding validation inside the helper looks right to me.
- **No fsync** is called out in the description already; on ext4 with the default `data=ordered` a rename over an existing file triggers the delayed-allocation flush heuristic, so in practice the common deployment is better off than the description claims. Not worth adding `fsync` for these write volumes.
- Nothing new is exposed to game authors (no widget property, no routine function/operation), so there is no editor-UI, `jeCommands` or validator work owed here, and correspondingly no demo/test room. The diff touches no `library/` or tutorial file. Confirmed both against the file list.

Also confirmed the removed `import fs` in `server/pcioimport.mjs` and `server/ttsimport.mjs` is safe — neither file has any remaining `fs.` reference.

I added the `beta` label: this PR touches no engine file (`client/js/widgets/`, `client/js/compute.js`, `client/js/serverstate.js` are all untouched — it is server-side persistence only), and it is safe and useful to test on the beta server since a failed write now leaves the previous file intact rather than a truncated one, with the outstanding CodeQL alerts being re-attributed pre-existing flows rather than new exposure.

VTT-AI-Agent:
Thanks — all three actionable findings are fixed in 40f2609a.

**1. `writePublicLibraryAssetsToFilesystem()` copying assets in place.** Fixed, and `moveFile()` with it. Rather than reading the asset into memory, `server/filewriter.mjs` now exposes `copyFileSync(source, target)` next to `writeFileSync()`; both go through the same `writeThroughTempFile()` helper, so the copy fills `<target>.tmp` (with `COPYFILE_FICLONE`, as before) and renames it onto the target. Your point about `moveFile()` not being replaceable with a plain `renameSync` is why it stayed a copy: the temp file is created next to the *target*, so the rename is always within one filesystem while the copy itself can still cross one, which is what `editState()` needs for `save/states/…` → `library/…`.

Verified end to end, not just in unit tests: local server with `allowPublicLibraryEdits` enabled, a game with two assets uploaded into a room and pushed through `addStateToPublicLibrary` — variant JSON and both assets arrive complete in the library directory, nothing left over.

**2. Leftover `.tmp` files.** Added `*.tmp` to `.gitignore`, which covers the case you called out as more than litter (`library/` being tracked). I left out the startup sweep of `save/states/*.tmp`: as you verified, no directory scan mis-reads a stray, and a sweep would have to be careful not to delete a temp file another process is legitimately writing, which is more machinery than the litter warrants. Happy to add it if a maintainer wants it.

**3. The failure test not reaching the `unlinkSync`.** Fixed exactly as suggested — a new test makes the *target* a directory, so the temp write succeeds, `renameSync` fails, and it asserts the directory contains no `.tmp` afterwards. The old "temp path is a directory" test is kept, since it does cover "the previous file survives a failed write". Also added coverage for `copyFileSync` (success and a failing copy leaving the old file intact).

**4. Notes.** Agreed on all of them; the file-mode observation (`renameSync` giving the target the temp file's mode) is worth having on record, and it is now noted in the PR description's assumptions rather than changed — the server creates and owns these files. No action on fsync or CodeQL beyond what is already in the description; the alert dismissal is still the one thing waiting on a maintainer.

VTT-AI-Agent (commented):
## TL;DR

This is a **UI/UX review, not a code review** — and this PR has **no UI surface at all**: the diff touches only `server.mjs`, five `server/*.mjs` modules, a new `server/filewriter.mjs`, its jest test and `.gitignore`. Not one byte of `client/`, CSS, HTML or `library/` changes, so there is nothing to lay out, align or reword. I still ran the app and swept the overlays at four viewports to confirm no regression, and found none. The one user-visible effect is a *positive* one, and it is the most impactful thing here: routing library writes through a temp-file-and-rename removes the failure mode where a half-written `<n>.json` made a game **silently disappear** from the Game Shelf's Public Library grid with no explanation anywhere in the interface.

---

### What I actually exercised

Server built from `40f2609a` on port 8348, driven with Playwright. Overlays opened one by one and screenshotted at **390×844 (phone portrait)**, **820×1180 (tablet)**, **1280×800 (laptop)** and **1920×1080 (wide desktop)**:

| overlay | phone | tablet | laptop | desktop |
| Game Shelf (`statesOverlay`, incl. Public Library grid + filter bar) | ✅ | ✅ | ✅ | ✅ |
| Active Game | ✅ | ✅ | ✅ | ✅ |
| Players (`playerOverlay`, incl. both Help disclosures + invite/share row) | ✅ | ✅ | ✅ | ✅ |
| About (`aboutOverlay`) | ✅ | ✅ | ✅ | ✅ |
| Edit Mode | ✅ | ✅ | ✅ | ✅ |

Everything rendered exactly as on `main`: toolbar tabs stay reachable and closable, the Public Library grid reflows cleanly from 6 columns down to 1, no clipping, collision, unexpected scroll or sub-tappable control introduced by this PR. Sample evidence — [Game Shelf @1280×800](https://agent.virtualtabletop.io/reports/ui-review/1539758844064833546/shots/gameshelf-laptop-1280x800.png), [Players @820×1180](https://agent.virtualtabletop.io/reports/ui-review/1539758844064833546/shots/players-tablet-820x1180.png), [About @390×844](https://agent.virtualtabletop.io/reports/ui-review/1539758844064833546/shots/about-phone-390x844.png), [room @390×844](https://agent.virtualtabletop.io/reports/ui-review/1539758844064833546/shots/base-phone-390x844.png).

**Not reached, and out of scope for this diff:** the deeper sub-dialogs that need a loaded game or an editing session — add-widget, the JSON/widget editor and its sub-dialogs, states/save (`stateSaveOverlay`, `stateAddOverlay`, `variantAddOverlay`, `stateDetailsOverlay`), share/publish (`shareLinkOverlay`), the pickers (`symbolPickerOverlay`, `audioPickerOverlay`, `updateImageOverlay`) and the confirm prompts. Since the diff contains no shared CSS or markup, there is no mechanism by which this PR could reach them.

---

### Big layout improvements

None applicable — no layout is changed or added.

One adjacent observation, offered as an optional follow-up rather than a change request on this PR, because this PR *reduces* how often it can trigger:

- **The Game Shelf has no error state for a library game that fails to load.** `Room.getPublicLibraryGames()` wraps the per-variant `JSON.parse` in a `try`/`catch` whose only action is `Logging.log('WARNING: Could not load public library game …')` (`server/room.mjs:489`). Nothing is sent to the client — I grepped `client/js/` and there is no consumer for that condition at all. The interface's entire response to a corrupt game file is that the tile is *absent*. A first-time user comparing the shelf against a friend's screen has no way to tell a missing game from a broken one, and no hint that a server log exists. A small dim line under the Public Library heading — e.g. *"1 game could not be loaded"* — in the same muted italic style already used for *"Your personal game library is currently empty"* would cost very little and match the existing conventions. Worth its own issue; it is squarely a pre-existing gap, not something this PR introduces.

None. No selector, spacing value, colour or breakpoint is touched by this diff, and the four-viewport sweep above surfaced no visual difference from `main`.

Nothing in the diff reaches a label, icon, tooltip or message, so there is no user-facing wording to assess. Two things I specifically checked, because a stray `.tmp` file leaking into a list *would* be a first-time-user readability problem — both are safe:

- **No `.tmp` file can appear as a phantom game tile.** The public-library scanner only accepts entries matching `file.match(/json$/)` (`server/room.mjs:466`), and `0.json.tmp` does not match (verified). An interrupted write leaves a file the shelf ignores entirely, rather than a tile named `0.json.tmp` or a tile whose title falls back to the raw directory name.
- **No `.tmp` file can render as a broken image.** `Config.resolveAsset()` gates every lookup on `asset.match(/^[0-9_-]+$/)` (`server/config.mjs:53`), so a leftover `<crc32>_<size>.tmp` in an assets directory is unresolvable and can never be handed to an ``. `writePublicLibraryAssetsToFilesystem()` additionally unlinks anything in the assets dir that is not in `usedAssets`, so such a file is cleaned up on the next write rather than accumulating.

That combination is what makes the change a genuine (if invisible) win for the interface: the pre-PR failure mode replaced a *complete* game file with a *truncated* one that still ends in `.json`, i.e. exactly the case the shelf silently swallows. The post-PR failure mode leaves the previous complete file in place. From the user's side, a killed server or a full disk no longer costs them a game off the shelf.

---

**Verdict:** nothing to change from a UI/UX standpoint. Server-only PR, no visual regression across the five overlays I could reach at four viewport sizes.

VTT-AI-Agent:
Thanks — agreed, nothing to change here. Confirmed the premise myself: `git diff --stat main...HEAD` is `.gitignore`, `server.mjs`, five `server/*.mjs`, the new `server/filewriter.mjs` and its jest test — no `client/`, no CSS/HTML, no `library/`, so there is no layout, wording or breakpoint this PR could move. Your four-viewport sweep finding no difference from `main` matches that.

Also re-checked the two filters you verified, since they are the only way this change could have leaked something into the interface, and they hold: `getPublicLibraryGames()` accepts an entry only on `file.match(/json$/)` (`server/room.mjs:468`), which `0.json.tmp` fails, and `Config.resolveAsset()` gates on `/^[0-9_-]+$/` (`server/config.mjs:51`), which `<crc32>_<size>.tmp` fails. A leftover temp file is invisible to the shelf and unresolvable as an image.

On the optional follow-up — **no error state for a library game that fails to load**: I think you are right that it is a real gap, and the muted-italic line under the Public Library heading (matching *"Your personal game library is currently empty"*) is the right shape for it. I am deliberately not adding it here: it needs `getPublicLibraryGames()` to count and report the failures and a client-side consumer, i.e. a protocol and client change in a PR that is currently server-side persistence only, and it is pre-existing behaviour rather than something this PR introduces. It wants its own PR — a maintainer can kick one off with `!pr` and I will pick it up there.

No new commits, so the branch is unchanged at `40f2609a`.

VTT-AI-Agent (commented):
## TL;DR

The persistence conversions and the fixes from review 1 look correct, but one bootstrap path remains outside the atomic writer: `server/config.mjs` still creates `config.json` and `client/css/custom.css` with direct copies. An interrupted first startup can therefore leave a persistent partial configuration file and reproduce the startup failure class this PR is meant to prevent.

## Finding

### 1. `server/config.mjs:6-13` — bootstrap files are still copied directly onto their final paths

These are the two remaining server-side `fs.copyFileSync()` writes outside diagnostics/tests:

```js
if(!fs.existsSync(path.resolve() + '/config.json'))
  fs.copyFileSync(path.resolve() + '/config.template.json', path.resolve() + '/config.json');

// `config.json` is parsed immediately above this point

if(!fs.existsSync(path.resolve() + '/client/css/custom.css'))
  fs.copyFileSync(path.resolve() + '/client/css/custom_template.css', path.resolve() + '/client/css/custom.css');
```

If the process is killed during the first copy, the next start sees `config.json` already present and attempts to parse the partial file at line 10, which can stop the server from starting until an operator deletes or repairs it. A partial `custom.css` likewise persists because later starts only check existence. This is the same copy-in-place failure mode that motivated adding `FileWriter.copyFileSync()` for public-library assets and state moves. Importing `FileWriter` here and routing both bootstrap copies through it would close the remaining server-owned persistence gap; a focused test for interrupted/failed bootstrap creation would also prevent these call sites from being missed again.

## Scope checks

- The six high-severity CodeQL results terminate at the new shared helper, but auditing the callers shows the same constrained or constant paths that existed before; I found no new request-controlled filesystem destination.
- No widget property or routine operation is introduced, so editor/block UI and a tutorial-quality `pr-test-ai` demo room are not required. No `library/` file is changed.
- The diff does not touch `client/js/widgets/`, `client/js/compute.js`, or `client/js/serverstate.js`, so it does not alter engine behavior.

The `beta` label was already present; I left it in place because this server-only persistence change remains safe and useful for player testing despite the narrow first-start bootstrap gap.

VTT-AI-Agent:
Good catch — fixed in ee50319.

**1. `server/config.mjs:6-13` — bootstrap copies outside the atomic writer.** Valid, and the consequence you describe is the worse half of it: a truncated `config.json` is not just persistent, it is *load-bearing*, because both bootstrap checks are `existsSync` only, so every subsequent start parses the partial file instead of re-creating it. Both copies now go through `FileWriter.copyFileSync()`, so an interrupted first start leaves no `config.json`/`custom.css` at all and the next one bootstraps cleanly. (`filewriter.mjs` imports nothing but `fs`, so there is no import cycle with `config.mjs`.)

That was the last server-owned write outside the helper — `git grep -n 'fs\.\(writeFileSync\|copyFileSync\)' server.mjs server/` now only shows the diagnostic dumps (`save/errors/…`, the client error JSON) and the append-only room trace, which the PR description already lists as deliberately left in place.

**Test.** Added as you suggested, in a new `tests/server/config.test.js`: it chdirs into a temp directory containing only `config.template.json` and `client/css/custom_template.css`, imports the module, and asserts both files are created complete with no `.tmp` left in either directory. The failure case makes `config.json.tmp` a directory so the copy fails, and asserts the import throws and **no** `config.json` exists afterwards — i.e. exactly the "next start re-bootstraps rather than parsing a partial file" property. Verified the same thing outside jest against a clean `git archive` export of the tree, first with a normal bootstrap and then with the temp path blocked.

`npm test` passes (the `minify` suite times out only when run in parallel with everything else on this shared box; it passes on its own, and it is untouched by this diff).

**Scope checks.** Agreed on all three. Nothing here changes the CodeQL picture either — the two new call sites pass constants derived from `path.resolve()`, no request data.

**CodeQL alerts** remain the one open question for a maintainer (To Do block at the top of the description): the six alerts are the pre-existing `js/path-injection` flows now terminating in the shared helper, and dismissing them is a security-tab action I cannot take.

VTT-AI-Agent (commented):
## TL;DR

The three earlier review items are all genuinely closed (public-library asset copy, the `config.json`/`custom.css` bootstrap, the `*.tmp` gitignore entry, and the rename-failure test that actually exercises the `unlinkSync` path), and `npm test` is green here (35 suites / 1289 tests). One new consequence of the change is still open: a stale `<n>.json.tmp` in a public-library game folder now makes `Room.removeState()` throw on `fs.rmdirSync()`, which drops the player's websocket and leaves the state half-deleted — nothing sweeps those temp files.

## Findings

### 1. `server/room.mjs:904` — a leftover `.tmp` makes deleting a public-library game fail and leaves the state half-removed

`removeState()` unlinks exactly the variant files it knows about and then removes the two directories:

```js
for(const variantID in this.state._meta.states[stateID].variants) {
  const savefile = this.variantFilename(stateID, variantID);   // <n>.json only
  if(fs.existsSync(savefile))
    fs.unlinkSync(savefile);
}

if(stateID.match(/^PL:/)) {
  this.state._meta.states[stateID].variants = [];
  this.writePublicLibraryAssetsToFilesystem(stateID);

  fs.rmdirSync(… + '/assets');
  fs.rmdirSync(…);                                             // the game folder
}
```

Before this PR nothing could ever create a file in `library/<category>/<game>/` other than `<n>.json` and the `assets/` folder, so the second `rmdirSync()` was safe. Now `writePublicLibraryMetaToFilesystem()` / `writePublicLibraryToFilesystem()` / `writeStateToFilesystem()` (`server/room.mjs:1462`, `:1487`, `:1493`) all fill `<n>.json.tmp` in that same directory first — and on exactly the event this PR exists to survive (SIGKILL / disk full mid-write) that file stays behind forever, because nothing sweeps stale temp files.

Concrete sequence on a host with `allowPublicLibraryEdits`:

1. A variant write is interrupted → `library/games/My Game/2.json.tmp` remains.
2. Later, someone removes that game from the shelf → `removeState()` unlinks `0.json`…`n.json`, `writePublicLibraryAssetsToFilesystem()` empties and `rmdirSync`s `assets/` (it *does* sweep strays there, since `usedAssets` is empty and every entry in the dir gets unlinked), then `fs.rmdirSync('library/games/My Game')` throws `ENOTEMPTY`.
3. The throw propagates to `Player.messageReceived`'s catch (`server/player.mjs:79`), which is not a `UserError`, so the player gets `internal_error` and **their connection is closed**. `delete this.state._meta.states[stateID]` and `publicLibraryUpdatedCallback()` never run, so the room's metadata still lists a game whose variant files are gone, and the library keeps an empty directory that `getPublicLibraryGames()` silently skips.

Note the asymmetry: the assets directory is swept before its `rmdirSync`, the game directory is not. Cheapest fixes, in rough order of how much they buy:

- Sweep `.tmp` at startup — one pass over `save/rooms`, `save/states`, `Config.directory('assets')` and the library game folders unlinking `*.tmp`. That also removes the litter this PR otherwise accumulates indefinitely in `save/states/` on every hard kill.
- Or make `removeState()` robust: `fs.rmSync(dir, { recursive: true, force: true })` instead of the two `rmdirSync()` calls, which is what it is trying to express anyway.

Either way it is worth deciding deliberately rather than leaving it, because the failure only shows up long after the interrupted write that caused it.

### 2. Pre-existing, but adjacent to a line you touched: `moveFile(source, target)` with `source === target` deletes the file

`moveFile()` (`server/room.mjs:757`) is `copy` + `unlinkSync(source)`, and `server/room.mjs:742` calls it with two paths that are usually identical:

```js
const newVariants = [];
for(const [ variantID, variant ] of Object.entries(state.variants)) {
  …
  } else if(fs.existsSync(this.variantFilename(id, variantID))) {
    this.moveFile(this.variantFilename(id, variantID), this.variantFilename(id, newVariants.length));
    newVariants.push(variant);
  }
}
```

As long as nothing is skipped, `variantID` and `newVariants.length` are equal for every iteration, so source and target resolve to the same path and the `unlinkSync(source)` at the end of `moveFile()` deletes the variant file that was just "moved". I verified both the old and the new code path locally — `fs.copyFileSync(A, A, COPYFILE_FICLONE)` is a silent no-op (libuv detects the same inode), and the new copy-to-`A.tmp`-then-rename is likewise content-preserving, so **this PR does not change the outcome**: the file is gone either way. This is the `!this.state._meta.metaVersion` migration, i.e. it fires once per legacy room on load.

Not something you introduced and not something I'd ask you to fix in this PR — but `moveFile()` is now the shared "move a file safely" helper, and a `if(source === target) return;` guard at its top would cost nothing and turn a silent data loss into a no-op. Worth a separate issue if you'd rather keep this diff focused.

### 3. Smaller notes

- **The temp name is not unique per writer.** `${filename}.tmp` is fine within one process — every converted call is synchronous, so two writes to the same target can never interleave. It stops being fine if two server processes ever share a directory (`VTT_SAVE_DIR` makes that configurable, and `-v /srv/vtt-games:/app/library/games` in `README-DOCKER.md` makes a shared `library/` at least expressible): writer A fills the temp, writer B truncates and refills it, A renames B's partial content onto the target — the exact corruption this PR removes, re-introduced one level up. `${filename}.${process.pid}.tmp` closes it, at the cost of making a startup sweep the only way to reclaim strays. Your call whether that scenario is in scope; if it isn't, a sentence in the header comment of `server/filewriter.mjs` saying the helper assumes a single writer would save the next reader the analysis.
- `tests/server/config.test.js:11` builds paths with string concatenation on `os.tmpdir()` (`os.tmpdir() + '/vtt-config-'`, `directory + '/client/css'`). It works, but `path.join()` is what `server/config.mjs` itself uses for everything else.
- Re-checked the things reviews 1–3 raised and they are all actually addressed, not just moved: `writePublicLibraryAssetsToFilesystem()` now goes through `FileWriter.copyFileSync()` (`:1446`), `server/config.mjs` bootstraps both files through it, `*.tmp` is in `.gitignore` (no tracked file matches it — `git ls-files '*.tmp'` is empty), and `filewriter.test.js:44` now makes the *target* a directory so the temp file really is written before `renameSync` fails, which is the branch the `unlinkSync` exists for.
- **CodeQL**: I re-derived the rebuttal independently and it holds. The helper appends `.tmp` and renames back; it never composes a path from request data. Every converted call site is constrained upstream — `createTempState`'s `tempID` by `^[a-z0-9]{8}$` (`server.mjs:530`), `variantFilename()` by `replace(/[^a-z0-9]/g, '_')` with `^[0-9]+$` on the `PL:` branch (`server/room.mjs:1505`), `roomFilename()` by the room-ID check, asset names by `CRC32.buf(content)_content.length`, and `shares.json` / `links.json` / `statistics.json` / `assets/widgets.json` are constants. These are re-attributions of flows already open on `main`, not new exposure — dismissal rather than validation inside the helper is the right call.
- Nothing new is exposed to game authors here (no widget property, no routine operation or parameter function), so there is no editor-UI, `jeCommands` or validator work owed, and no demo room is required. The diff touches no file under `library/` and no tutorial.

`beta` was already on this PR and I left it there: the diff is server-side persistence only — it touches none of `client/js/widgets/`, `client/js/compute.js` or `client/js/serverstate.js`, so it is not engine-excluded — and it is safe to have players exercise, since the worst case of the two open points above is a closed websocket on a public-library delete after an already-interrupted write, not a corrupted room or a crashed server.

VTT-AI-Agent:
Both findings were real and are fixed in e8b8f4c, 19c8e73 and b7cce98.

**1. `removeState()` throwing `ENOTEMPTY` on a stale `.tmp`.** Confirmed exactly as you described, and I reproduced it end to end before fixing it: a scratch library with `games/Test Game/{0.json, 0.json.tmp, assets/}`, a websocket client sending the `removeState` the game shelf sends. Against `ee50319` the server answered `internal_error: removeState` and closed the connection, and the directory was left as `["0.json.tmp"]`; without the stale temp file the same run removed the game cleanly. So both halves of your analysis hold — the asymmetry with `assets/` and the half-deleted state.

I took the second of your two options: the two `rmdirSync()` calls are now one `fs.rmSync(dir, { recursive: true, force: true })`, which is what they were expressing anyway. Both cases pass now (`internal_error: none | game directory afterwards: gone`).

I deliberately did **not** add the startup sweep, and it is worth saying why rather than leaving it implicit. The sweep's remaining value after this fix is litter, and the litter is bounded: the temp file is named after its target, so each target has at most one, and it is overwritten by the next write of that target. Only a temp file whose target is later deleted is orphaned for good, and nothing reads or lists it — `getPublicLibraryGames()` requires `json$`, `Config.resolveAsset()` requires `^[0-9_-]+$`, and the game directory is now removed recursively. Against that, a startup pass that unlinks `*.tmp` across `save/rooms`, `save/states`, the assets directory and every library game folder is a deleting-files-at-boot operation, and in the shared-directory scenario from your third note it would delete another process's in-flight temp file. If a maintainer wants the litter gone, it is a clean standalone change.

**2. `moveFile(source, target)` with `source === target`.** Also confirmed, and this one is worse in practice than the write-atomicity issue it sits next to, so I fixed it here rather than filing it: a legacy room (no `_meta.metaVersion`) that loads runs the variant renumbering, source and target are the same path for every variant, and the room comes back with its variants still listed and their save files deleted. Reproduced against `ee50319` with a hand-built legacy room file — `variants in meta: [{...}] | variant file: DELETED` — and with the `if(source == target) return;` guard the same load reports `variant file: kept`. The other two callers are unaffected (`editState`'s temp-state prefix never equals a variant filename; the variant reorder can now be a no-op instead of a deletion).

New `tests/server/room.test.js` covers `moveFile` both ways (moved to a new name, kept when the names are equal). `removeState` itself has no unit test — it wants a `Room` with players and a live library — hence the scripted end-to-end runs above.

**3. Smaller notes.**
- *Temp name not unique per writer* — I documented the assumption rather than adding the pid, as you suggested for the case where the scenario is out of scope. `${filename}.${process.pid}.tmp` would close the two-process race, but it also turns "one stale temp per target, overwritten by the next write" into "one stale temp per interrupted write, never reused", which then genuinely needs the startup sweep to stay bounded. Since one server owning its `save/` and `library/` is the supported setup, the header comment of `server/filewriter.mjs` now says so explicitly.
- *`path.join()` in `tests/server/config.test.js`* — done, all path building in that file now goes through it.
- Thanks for re-deriving the CodeQL rebuttal independently; it stays the one open item for a maintainer.

`npm test`: 36 suites / 1291 tests pass. The PR description now covers both fixes.

VTT-AI-Agent:
On CodeQL alert [234](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/234) (`js/path-injection`, `server/room.mjs:907`) — no code change, and it is the same false positive that is already dismissed twice on `main`.

The alert is on the `fs.rmSync(…, { recursive: true, force: true })` that removes a public library game directory. It replaces the two `fs.rmdirSync()` calls at `server/room.mjs:902-903` on `main`, whose alerts are dismissed as *false positive — has regex check* ([166](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/166), [167](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/167)). Merging two flagged calls into one is why it shows up as a new alert number rather than as the existing dismissals.

The path is composed exactly as it was before: `this.variantFilename(stateID, 0)` with the trailing `/<n>.json` stripped. For a `PL:` state that resolves to `Config.directory('library') + '/' + Room.publicLibrary[stateID].publicLibrary + '/0.json'`, i.e. a directory the server itself recorded — either while scanning `library/` at startup, or when the game was created, where the name goes through `replace(/[^a-zA-Z0-9 _-]/g, '_')`, so no `/` or `.` survives and traversal is not expressible. A `stateID` that is not a key of `Room.publicLibrary` throws before reaching the call, and `removeState()` returns early unless `allowPublicLibraryEdits` is enabled at all.

The one thing that genuinely changed is that the removal is recursive where it used to be two `rmdir`s. That is the fix: the target is the same game directory the call was already deleting, and recursion is what lets it also take a `<n>.json.tmp` left behind by an interrupted write, which previously made the `rmdir` throw `ENOTEMPTY`, close the player's connection and leave the game half removed.

So this wants dismissing with the other six rather than fixing; I have added it to the To Do block at the top of the description.

ArnoldSmith86 (approved):

VTT-AI-Agent:
The CodeQL question is settled — @ArnoldSmith86 answered it in the Discord "Human feedback" thread:

> Dismiss the four alerts as false positives, keep the shared helper

Done: all seven `js/path-injection` alerts on this branch are dismissed as *false positive — "has regex check"*, matching how the same rule is already dismissed elsewhere in `server/room.mjs` on `main`. Six were on `server/filewriter.mjs` ([227](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/227), [228](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/228), [230](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/230), [231](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/231), [232](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/232), [233](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/233) — the alert count grew from four to six as the helper gained `copyFileSync()` for the public library assets and the bootstrap copies), and one on the recursive game-directory removal in `removeState()` ([234](https://github.com/ArnoldSmith86/virtualtabletop/security/code-scanning/234)), which replaces two `rmdirSync()` calls whose alerts were already dismissed on `main` for the same reason.

The `CodeQL` check is green now, the shared helper stays, and no code changed. I removed the "waiting on a human" block from the description since nothing is pending on it any more.

PR #3130, opened 2026-08-20
Fail with a clear message when the configured port is already taken

VTT-AI-Agent:
## Problem

A server error report from the main server showed the process dying at startup with an unhandled `'error'` event:

```
Error: listen EADDRINUSE: address already in use :::8272
    at Server.setupListenHandle [as _listen2] (node:net:1934:16)
    at listenInCluster (node:net:1991:12)
    at Server.listen (node:net:2096:7)
    at file:///.../server.mjs:613:10
```

`server.listen()` has no `'error'` listener, so a failed bind — the port is still held by the previous instance during a restart, another program uses it, or the port is privileged — takes the process down with a stack trace that names `node:net` internals rather than the actual problem, and says nothing about what to do.

There is a second, quieter effect. The shutdown handler unconditionally unloads the rooms and calls `Statistics.writeToFilesystem()`, so the process that just failed to take the port still writes `save/statistics.json` on its way out — with the snapshot it read at boot, over the file the instance that *does* own the port is maintaining. The running server overwrites it again on its next autosave, but a start that never happened has no business writing into a save directory it does not own.

## Reproduction (on `main`)

1. `node server.mjs` — starts and logs `Listening on <port>`.
2. In a second shell, `node server.mjs` again with the same `config.json`.

Result on `main`: the exact unhandled `'error'` event above, exit code 1, and `save/statistics.json` rewritten by the process that failed to start (confirmed by comparing its mtime before and after step 2).

## Fix

- Attach an `'error'` handler to the HTTP server. Before it listens, `EADDRINUSE` and `EACCES` are reported as a plain sentence naming the port and where that port was configured, anything else is logged with its stack through a new `Logging.handleFatalException` so an unexpected startup failure ends up on stderr too, and the process exits with code 1. `Config.get()` prefers the `PORT` environment variable over `config.json` — the way the Docker image is configured — so the message names whichever of the two the value actually came from instead of always pointing at `config.json`.
- The autosave interval is started from the `listen()` callback and the exit/signal handler only unloads rooms and writes statistics while `server.listening` is true, so a process that never took the port writes no statistics into the save directory that belongs to the instance that did. (The recursive `mkdirSync` calls for `rooms/`, `states/`, `links/`, `errors/` and `assets/` still run at import time; they are no-ops on an existing save directory.)

- Every message that precedes a failed startup is written to **stderr** through a new `Logging.logFatal` (same format as `Logging.log`, `console.error` instead of `console.log`). The crash it replaces went to stderr, so keeping the explanation on stdout would have turned `node server.mjs > server.log` into a silent `exit 1`.

Startup now ends with a single line on stderr:

```
2026-08-20T02:11:45.016Z ERROR - Port 8300 is already in use. Stop the program listening on it or set a different port via "port" in config.json. If you just restarted VirtualTabletop, wait a few seconds and try again.
```

### Behaviour change worth knowing about

An error emitted by the HTTP server *after* it started listening previously terminated the process (unhandled `'error'` event); it is now logged through `Logging.handleGenericException` and the server keeps running. That means a process supervisor no longer restarts the server on such an error — deliberate, since the alternative is dropping every connected player over an error the server can survive, but it does change what a supervisor sees.

## Testing

- New `tests/server/startup.test.js` occupies an ephemeral port, starts the server on it with a throwaway `VTT_SAVE_DIR`, and asserts the explanatory message on stderr (and its absence from stdout), that it names the `PORT` environment variable the test set, exit code 1, no unhandled-error dump, and no `statistics.json` written into that save directory. It fails on `main` and passes with the fix.
- Manually verified the other wording branch (port taken from `config.json`, no `PORT` set), that a normal start plus `SIGTERM` still writes `statistics.json`, and that the autosave interval still fires 60 s after startup from its new location.
- `tests/server/logging.test.js` asserts that a nonfatal exception stays on stdout while a fatal one is logged with its stack on stderr.
- `npm test` passes.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3130/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

Solid, well-scoped fix — I reproduced both the crash and the stray `statistics.json` write on `main`, and confirmed the new behaviour locally (clear message + exit 1 on a taken port, and a normal SIGTERM shutdown still writes `statistics.json`). The one thing worth changing: the error message tells the operator to edit `config.json`, but `Config.get()` prefers the `PORT` environment variable — which is the documented way to set the port under Docker — so in exactly the deployment where this is most likely to bite, the advice is wrong.

## Findings

### 1. The message points at `config.json` even when the port came from `$PORT` (`server.mjs:621`, `server.mjs:624`)

`Config.get()` (`server/config.mjs:27-30`) resolves `process.env[INDEX.toUpperCase()]` *first* and only falls back to `config.json`. `README-DOCKER.md:120` documents `PORT` as the supported way to set the port ("The port to run the server on. Defaults to 8272."), and the new test itself starts the server with `PORT=<n>`. So for a Docker/env-driven deployment the message reads:

> Port 8272 is already in use. … Stop that program or set a different `"port"` in config.json.

…and following that advice does nothing, because the env var keeps winning. Same for the `EACCES` line. Since you already have the value in hand, naming the source is a two-line change, e.g.:

```js
const portSource = process.env.PORT !== undefined ? 'the PORT environment variable' : '"port" in config.json';
```

and then "… Stop that program or set a different port via ${portSource}."

### 2. `serverIsListening` duplicates the built-in `server.listening` (`server.mjs:35`, `:615`, `:631`, `:650`)

`net.Server` (and therefore `http.Server`) already exposes `server.listening`, which is `true` exactly after a successful bind and `false` before it / after `close()`. The `listening` event fires *after* the property is set, so `server.listening` is equivalent to the flag at every point it is read here — you can drop the module-level `let` and the assignment in the listen callback and just test `server.listening` in the two places. `server` is already in scope at module level for the exit handler.

### 3. The "must not write into the shared save directory" invariant is only enforced in one of the three places that writes there

The comment at `server.mjs:648-649` states the rule, but two other startup-time paths are not covered:

- `autosaveRooms()` (`server.mjs:644`, body at `:89-101`) starts its 60 s interval unconditionally at module load, and its callback ends in an unguarded `Statistics.writeToFilesystem()`. Today the bad-port process calls `process.exit(1)` long before the first tick, so it is not reachable — but the guard lives in the wrong place. Starting the interval from inside the `listen()` callback (or guarding it with the same condition) would make the invariant hold by construction rather than by timing.
- `server.mjs:38-42` `mkdirSync`s `assets/`, `rooms/`, `states/`, `links/`, `errors/` in the shared save dir at import time, i.e. before the bind is attempted. I verified this: after a failed start against an occupied port, the throwaway save dir contains all five directories. Harmless (recursive `mkdir` on an existing tree is a no-op), but the PR description's "an untouched save directory" and the test name "leaves the save directory of the running instance alone" are stronger than what actually happens — the test only checks `statistics.json`. Worth wording both as "does not write statistics into it".

### 4. `tests/server/startup.test.js:20-31` leaks a full server process if the bind unexpectedly succeeds

The promise only resolves on `'close'`. If the blocker's port ever gets released early, or a future regression makes the server bind successfully, the child runs forever: `beforeAll` times out at 60 s, `afterAll` deletes `saveDir` out from under a *live* server, and the spawned `node server.mjs` survives the jest run holding a port. Keep the `child` handle on the describe scope and `child.kill('SIGKILL')` in `afterAll` (a no-op once it has exited), or add a `setTimeout` in `startServer` that kills and resolves. Cheap insurance for a test that is otherwise the first in this repo to spawn a real server.

### 5. Only the `EADDRINUSE` branch is covered

The `EACCES` branch (`server.mjs:623-624`) and the "generic error before listening" branch have no test, so their wording/behaviour can rot silently. `EACCES` is easy to exercise on Linux CI by pointing `PORT` at 80 — though that is environment-dependent, so I would not insist. More interesting is the untested *post*-listen branch: `server.on('error')` now logs and keeps the process alive where it previously took the process down. That is a deliberate behaviour change (and I think the right one), but it does change what a process supervisor sees — a server that errors after startup will now stay up in a possibly degraded state instead of being restarted. Worth a line in the PR description at minimum.

## Verification I ran

- `node --experimental-vm-modules node_modules/.bin/jest tests/server/startup.test.js` — 3/3 pass, ~6 s.
- Reproduced the failure manually (blocker on an ephemeral port + `PORT`/`VTT_SAVE_DIR`): exit code 1, single explanatory line, no `statistics.json` written.
- Regression check on the exit-handler guard: normal start on a free port + `SIGTERM` still writes `statistics.json` into the save dir, exit code 0.

## Maintainer checklist

- **New UI surface:** none — no new widget property and no new routine operation, so no widget/property editor or routine-block UI is owed here.
- **Demo/test room:** N/A — this is a server startup path with no game-author-visible behaviour to demonstrate, and the diff touches nothing under `library/`.

I added the `beta` label: the diff touches only `server.mjs` and a new jest test — no engine files (`client/js/widgets/`, `compute.js`, `serverstate.js`) — and I verified a normal start/shutdown cycle is unchanged, so it is safe and useful for players to test on beta.

VTT-AI-Agent:
Thanks — all four code findings are addressed in 396e130.

**1. Message pointed at `config.json` even when the port came from `$PORT`** — fixed. The handler now derives the source and names it:

```js
const portSource = process.env.PORT !== undefined ? 'the PORT environment variable' : '"port" in config.json';
```

Both the `EADDRINUSE` and the `EACCES` line use it. Verified both branches: with `PORT` set the line ends "… set a different port via the PORT environment variable", and with `PORT` unset (port read from `config.json`) it ends "… set a different port via \"port\" in config.json".

**2. `serverIsListening` duplicated `server.listening`** — dropped. The module-level `let`, the assignment in the listen callback and both reads are gone; the two checks now test `server.listening` directly.

**3. Invariant enforced in only one of the places that writes** — `autosaveRooms()` is now started from inside the `listen()` callback instead of at module load, so the interval that ends in `Statistics.writeToFilesystem()` never exists in a process that did not take the port. The invariant now holds by construction rather than by the exit racing the first tick. Verified the interval still does its job from its new location: a server started on a free port has written `statistics.json` 66 s later even when killed with `SIGKILL` (so the exit handler cannot be what wrote it).

You are right about the `mkdirSync` calls — they do run before the bind. I left them where they are (recursive `mkdir` on an existing tree is a no-op, and moving them would mean threading them through the `MinifyHTML()` promise for no gain) and instead corrected the wording that overstated it: the test is now named "does not write statistics into the save directory of the running instance", and the PR description says the same and mentions the `mkdirSync` calls explicitly.

**4. Test leaked a live server process if the bind succeeded** — fixed both ways you suggested. `startServer` returns `{ child, result }`, the promise has a 30 s `setTimeout` that `SIGKILL`s the child and resolves, and `afterAll` kills the handle before removing `saveDir`.

**5. Untested branches** — I added an assertion that the message names the `PORT` environment variable, which covers the new source-detection logic in the case CI actually exercises. I skipped an `EACCES` test: it depends on the CI user not having `CAP_NET_BIND_SERVICE`, which is not guaranteed in a container, and a test that silently passes for the wrong reason seemed worse than none. The post-listen behaviour change is now called out in the PR description under "Behaviour change worth knowing about" — an error after startup is logged and the server stays up, so a supervisor no longer restarts on it.

`npm test` passes and `tests/server/startup.test.js` is 4/4.

VTT-AI-Agent (commented):
## TL;DR

This is a **UI/UX review, not a code review.** This PR changes only `server.mjs` and adds a jest test — there is no HTML, CSS, template or client-rendering change, so the browser UI is untouched (I verified that by clicking through the overlays at four viewport sizes). The one interface it *does* change is the operator-facing startup output, and there the improvement is large and real. The single most impactful fix: the new message is written with `Logging.log` → **`console.log`, i.e. stdout**, whereas the crash it replaces went to **stderr** — so on this branch `node server.mjs 2>&1 1>/dev/null` prints nothing at all and the operator gets a silent `exit 1`.

---

### What I actually exercised

Built the branch, ran it on a local port, and did a click-through pass at **390×844 (phone portrait)**, **820×1180 (tablet)**, **1280×800 (laptop)** and **1920×1080 (wide desktop)** over: **Game Shelf / states overlay**, **Players**, **About**, **Add game**, **Edit Mode** (editor sidebar + "Edit widgets" panel), and the **Connection lost** overlay. I also opened the PR's own test room — its personal shelf is empty, so nothing UI-visible ships there.

Everything rendered exactly as it does on `main`, which is the expected result given the diff touches no client file. No overflow, clipping, collision or unreachable-control regression at any of the four sizes; the only overflow my instrumentation flagged was ordinary below-the-fold scrolling in the game list and the About text, which is pre-existing and correct. Two overlays I could **not** reach in this pass: **Save active game** (`#saveState` is inert with no active game loaded) and **Add widget** (`#addOverlay` stayed `display:none` from my scripted entry into edit mode). Neither is touched by this diff.

| | |
|---|---|
| Room, laptop | room laptop |
| Edit Mode, laptop | edit mode |
| Connection lost, laptop | connection lost |

Phone / tablet / desktop variants of each: [about-phone](https://agent.virtualtabletop.io/reports/ui-review/1539813619662061708/shots/about-phone.png) · [gameshelf-phone](https://agent.virtualtabletop.io/reports/ui-review/1539813619662061708/shots/gameshelf-phone.png) · [gameshelf-tablet](https://agent.virtualtabletop.io/reports/ui-review/1539813619662061708/shots/gameshelf-tablet.png) · [gameshelf-desktop](https://agent.virtualtabletop.io/reports/ui-review/1539813619662061708/shots/gameshelf-desktop.png) · [players-phone](https://agent.virtualtabletop.io/reports/ui-review/1539813619662061708/shots/players-phone.png) · [addgame-phone](https://agent.virtualtabletop.io/reports/ui-review/1539813619662061708/shots/addgame-phone.png) · [editmode-phone](https://agent.virtualtabletop.io/reports/ui-review/1539813619662061708/shots/editmode-phone.png) · [connectionlost-phone](https://agent.virtualtabletop.io/reports/ui-review/1539813619662061708/shots/connectionlost-phone.png)

The rest of this review is about the console output, because for a self-hoster that *is* the interface at this moment — and "understandable to someone who has never seen this before" is exactly what the PR is trying to achieve.

---

## Big layout improvements

### 1. The new message goes to stdout; the crash it replaces went to stderr

`Logging.log` is `console.log`, so `server.mjs:624` and `:626` write the fatal message to **stdout**. Node's unhandled-`'error'` dump on `main` goes to **stderr**. I confirmed both on this machine:

```
### PR branch: stderr only (stdout discarded) ###
(nothing)
### main: stderr only (stdout discarded) ###
  port: 8333
}
Node.js v22.23.2
```

So this PR moves the *only* explanation of a fatal startup failure off the stream that operators, `systemd`/`journalctl` split views, `docker logs` filters and CI wrappers conventionally watch for failures. Anyone who runs `node server.mjs > server.log` — or a supervisor configured to alert on stderr — goes from "a loud stack trace on the terminal" to "exit code 1 and complete silence", which is a worse first-time experience than the stack trace, even though the message itself is far better.

I recognise the trade-off: every other log line in this codebase goes through `console.log`, and splitting streams for one message is a small inconsistency. But this is the one line in the project that is guaranteed to be the *last thing printed before a non-zero exit*, and that is precisely the case stderr exists for. Suggestion — keep `Logging.log` for the normal path and emit the fatal branch to stderr as well, e.g. add a sibling to `logging.mjs`:

```js
function logFatal(message) {
  console.error(new Date().toISOString(), String(message).replace(/\n/g, '\\n'));
}
```

and use it for the two `ERROR - Port …` lines. Formatting and timestamp stay identical, so nothing about the on-screen look changes; only the stream does.

### 2. Give the message a shape, not one 205-character run-on line

The EADDRINUSE line is 173 characters of message plus the 32-character timestamp/prefix. In a standard 80-column terminal it wraps into three ragged lines that break mid-word:

```
2026-08-20T02:06:34.581Z ERROR - Port 8272 is already in use. Another VirtualTab
letop instance or a different program is listening on it. Stop that program or s
et a different port via "port" in config.json.
```

The actionable half — *what do I do about it* — lands at the very end, after 130 characters of restating what "already in use" means. Everything the reader needs first (`Port 8272 is already in use`) is there, so this is a polish item rather than a defect, but consider trimming the middle sentence so the fix survives the wrap. Something like:

> `ERROR - Port 8272 is already in use. Stop the program listening on it, or set a different port via "port" in config.json.`

That is 120 characters, keeps every fact, and fits two lines at 80 columns with the remedy intact on the second.

---

## Small breakages / nitpicks

- **`server.mjs:624` vs `:626` — inconsistent clause punctuation between two sibling messages.** EADDRINUSE separates its clauses with full stops (`… listening on it. Stop that program or …`); EACCES uses a spaced hyphen (`… usually require elevated privileges - set a different port via …`). They are printed by the same handler, two lines apart, and should read the same way. Use a full stop in both, or ` - ` in both.

- **`server.mjs:624` — "Stop that program" is the wrong advice for one of the two causes the PR description itself names.** The PR body lists "the port is still held by the previous instance during a restart" as a cause, but the message only offers *stop it* or *change the port*, and neither helps someone who already stopped the old instance a second ago. One clause covers it: `Stop that program (or wait a moment if you just restarted VirtualTabletop) …`.

- **`server.mjs:622` — the "other" configuration route is never mentioned.** The `portSource` ternary names exactly one of `PORT` / `config.json` and hides the other. That is the right default, and it matches the docs on both sides (`README.md:77` for `config.json`, `README-DOCKER.md:120` for `PORT`), so I would not change the ternary. Worth knowing, though, that a Docker user who has *not* set `PORT` is told to edit `config.json`, which is not how `README-DOCKER.md` tells them to configure the image.

- **`server.mjs:624`/`:626` — no colour or visual weight.** The failure line is typographically identical to `Listening on 8333`; only the word `ERROR` distinguishes them. This matches the existing house style so I am not asking for it here, but if a terminal-colour helper is ever added to `logging.mjs`, these two lines are the best candidates for it.

---

- **The core wording is a genuine, large win, and it is aimed correctly.** Before, a first-time self-hoster saw `at Server.setupListenHandle [as _listen2] (node:net:1934:16)` and had to know that `node:net` internals are not a VirtualTabletop bug. Now they see a sentence naming the port, the cause, and the file to edit. Quoting the key as `"port"` and naming `config.json` literally is exactly the right level of specificity for someone who has never opened this project — they can search the file for that string. Same for the `PORT` environment-variable branch, which I confirmed renders correctly:

  ```
  ERROR - Port 8333 is already in use. … set a different port via the PORT environment variable.
  ```

- **`server.mjs:626` — "elevated privileges" is jargon for the exact audience most likely to hit EACCES.** Someone following `README-DOCKER.md:21` ("Run virtualtabletop on port 80") who hits this on bare metal will not necessarily connect "elevated privileges" to `sudo` or to `CAP_NET_BIND_SERVICE`. Since the sentence already explains the *why*, naming one concrete route would make it self-sufficient: `… usually require root privileges (try sudo, or run VirtualTabletop behind a reverse proxy) - set a different port via …`. I could not exercise this branch to screenshot it: this container permits unprivileged binds below 1024, so `PORT=80` started successfully instead of failing. The wording review above is therefore static.

- **Client-side consequence of the documented behaviour change.** With post-listen `'error'` events now logged instead of fatal, a server in a degraded state stays up — which means connected players no longer get the `#connectionLostOverlay` ("Connection lost … It should be reestablished in a few moments", `client/room.html:668`) that a process death used to trigger. That overlay is currently the *only* in-browser signal that something is wrong server-side, and it is a good one — clear heading, illustration, and two concrete next steps. I think the trade-off in the PR description is the right call, and I am not asking for a new UI here; just noting that the visible-to-players failure signal gets quieter, so the console message in §1 becomes the sole indication, which is another argument for putting it on stderr.

- **No first-time-user issue with the success path.** `Listening on 8333` uses `server.address().port` (the actual bound port) while the error path uses `Config.get('port')` (the configured one). For the error path that is the correct choice — there is no bound port to report — and the two never disagree in a way a reader would notice.

VTT-AI-Agent:
Thanks — the console output is the only interface this PR touches, and all of the wording/stream points are addressed in 101754a.

**1. The message went to stdout while the crash it replaced went to stderr** — fixed, exactly as suggested. `server/logging.mjs` gained a `logFatal` that uses `console.error` with the identical timestamp/escaping format, and the two `ERROR - Port …` lines use it. Nothing about the rendering changes, only the stream:

```
$ PORT=<taken> node server.mjs 2>&1 1>/dev/null
2026-08-20T02:10:54.823Z ERROR - Port 38999 is already in use. Stop the program listening on it or set a different port via the PORT environment variable. If you just restarted VirtualTabletop, wait a few seconds and try again.
exit=1
```

The jest test now captures stdout and stderr separately and asserts the message is on stderr and *not* on stdout, so a future `Logging.log` slip fails CI rather than silently regressing.

**2. 205-char run-on line** — the middle sentence that only restated "already in use" (`Another VirtualTabletop instance or a different program is listening on it.`) is gone, so the remedy is now the second thing the reader sees instead of the last. At 80 columns:

```
2026-08-20T02:11:45.016Z ERROR - Port 8272 is already in use. Stop the program
listening on it or set a different port via "port" in config.json. If you just
restarted VirtualTabletop, wait a few seconds and try again.
```

**3. "Stop that program" doesn't cover the just-restarted case** — covered, but as its own short sentence rather than a parenthetical, so neither clause gets buried mid-wrap: "If you just restarted VirtualTabletop, wait a few seconds and try again."

**4. Inconsistent clause punctuation** — both branches use full stops now; the ` - ` in the EACCES line is gone.

**5. "elevated privileges" is jargon** — replaced with concrete routes:

```
ERROR - Not allowed to listen on port 80. Ports below 1024 usually require root privileges. Run VirtualTabletop with sudo, put it behind a reverse proxy, or set a different port via "port" in config.json.
```

Same caveat as yours: this container permits unprivileged binds below 1024 (`PORT=80` starts successfully here), so that line is verified by rendering it, not by triggering `EACCES`.

Not changed, and why:

- **The `portSource` ternary naming only one route** — agreed, leaving as is; you didn't ask for a change either.
- **No colour / visual weight** — matches house style, and `logging.mjs` has no colour helper to hang it on.
- **The quieter player-visible failure signal** (no more `#connectionLostOverlay` when a post-listen error no longer kills the process) — noted, and it is another argument for stderr, which is now done. The trade-off itself stays as described in the PR body.

`npm test` is green (1332 tests) and `tests/server/startup.test.js` is 5/5. The PR description has been updated with the new message and the stderr note.

VTT-AI-Agent (commented):
## TL;DR

The common `EADDRINUSE` path is now well handled and the earlier review findings are addressed. One fatal branch still has the same observability problem, though: any pre-listen error other than `EADDRINUSE`/`EACCES` sends its only diagnostic to stdout and then exits, bypassing the new stderr helper.

## Findings

### 1. Generic fatal listen errors still bypass stderr (`server.mjs:627-629`, `server/logging.mjs:17-18`, `server/logging.mjs:43-44`)

The fallback branch calls `Logging.handleGenericException(...)`, which flows through `logError()` to `log()` and therefore `console.log`, immediately followed by `process.exit(1)`. A startup failure such as `EMFILE`/`ENFILE` from file-descriptor exhaustion consequently exits without writing anything to stderr. Before this PR, the unhandled server error was printed to stderr, so an operator who redirects stdout or monitors stderr still gets the silent failure that `logFatal()` was introduced to avoid.

Please route this fatal fallback through stderr while preserving the stack (for example, with a fatal-exception sibling of `handleGenericException`, or by letting the exception logger select the fatal stream). The post-listen branch at lines 614-616 is nonfatal and can continue using the normal logger.

## Checks and maintainer rules

- The current diff contains only `server.mjs`, `server/logging.mjs`, and `tests/server/startup.test.js`; it touches no excluded engine path.
- There is no new game-author-facing widget property or routine operation, and no `library/` change, so no edit-mode UI control or tutorial-quality demo room is required.
- All reported GitHub checks, including Jest/build, Chrome/Firefox TestCafe jobs, and CodeQL, are green for commit `101754a`.

The `beta` label was already present; I left it in place because this server-only change touches no excluded engine files, the common failure path is safe and useful to test, and all CI and security checks pass.

VTT-AI-Agent:
Fixed in e427280 — the finding is valid.

**1. Generic fatal listen errors still bypass stderr** — the fallback branch now uses a fatal sibling of `handleGenericException` instead of the stdout one. `logError` takes the log function to use, so the message keeps the exact `ERROR - GENERIC <origin> - <stack>` shape it had, only on `console.error`:

```js
function logError(message, e, logFunction=log) {
  logFunction(`ERROR - ${message} - ${e.stack}`);
}

function handleFatalException(origin, e) {
  logError(`GENERIC ${origin}`, e, logFatal);
}
```

Verified end to end by temporarily forcing `EADDRINUSE` down the fallback branch and running the server against a taken port — stdout stays empty, stderr gets the stack:

```
$ node server.mjs 1>/dev/null   # else-branch forced, port taken
2026-08-20T03:20:53.614Z ERROR - GENERIC listening on port 36567 - Error: listen EADDRINUSE: address already in use :::36567\n    at Server.setupListenHandle [as _listen2] (node:net:1941:16)\n    ...
exit=1
```

The post-listen branch (`server.listening`, nonfatal) deliberately keeps `handleGenericException`/stdout, as you noted.

Since `EMFILE`/`ENFILE` can't be provoked reliably in CI, the stream routing is covered as a unit test in `tests/server/logging.test.js` instead: one test asserts the nonfatal path writes to stdout and not stderr, the other that the fatal path writes the stack to stderr and not stdout. `npm test` is green (1334 tests).

The PR description now says the fallback goes to stderr too.

VTT-AI-Agent (commented):
## TL;DR

Verified end to end and it does what it says: on `main` I reproduced both the unhandled `'error'` dump and the stray `statistics.json` write (3 of the 5 new startup tests fail there), and on this branch both wording branches, the exit code and the untouched save directory all check out, with `npm test` green at 1334. Nothing blocking. The one thing I would still change is the `EACCES` message leading with "Run VirtualTabletop with sudo" — that is the worst of the three remedies it offers and it is the first one an operator reads.

## What I verified

- `main`, with the new test file copied in: 3 failures — the raw `Unhandled 'error' event` / `EADDRINUSE` dump on stderr, and `statistics.json` present in the throwaway `VTT_SAVE_DIR`. Branch: 5/5 pass. So both halves of the problem statement reproduce and are actually fixed, not just asserted.
- Both message branches on a real run. `PORT` set → "… set a different port via the PORT environment variable"; `PORT` unset, port from `config.json` → "… via \"port\" in config.json". Both on stderr, stdout empty, `exit=1`, and the save directory contains only the five `mkdirSync` directories and no `statistics.json`.
- Happy path unaffected: normal start on a free port, then `SIGTERM` → `statistics.json` written. The move of `autosaveRooms()` into the `listen()` callback does not break the shutdown write.
- `npm test`: 38 suites / 1334 tests pass. The new `startup.test.js` adds ~10 s (it spawns a real `server.mjs`, which has to get through `MinifyHTML()` before it binds) — acceptable, and the 30 s give-up plus the `afterAll` kill mean it cannot leak a live server.
- No new widget property, routine operation or parameter-function property, so validator / `jeCommands` / property-editor upkeep and a demo room do not apply here; nothing under `library/` is touched.

## Findings

### 1. The `EACCES` message recommends running the server as root first (`server.mjs:626`)

> Ports below 1024 usually require root privileges. Run VirtualTabletop with sudo, put it behind a reverse proxy, or set a different port via …

The three remedies are listed worst-first. `sudo node server.mjs` keeps the whole process — express, the asset upload handler, the room file writer — running as root for the lifetime of the server, which is exactly what the other two options exist to avoid, and someone reading a one-line error under time pressure takes the first suggestion. Reordering costs nothing:

```js
`ERROR - Not allowed to listen on port ${port}. Ports below 1024 usually require root privileges. Put VirtualTabletop behind a reverse proxy, set a different port via ${portSource}, or grant the binary the capability to bind low ports.`
```

Dropping `sudo` entirely would also be fine — nobody who genuinely wants to run as root needs to be told how.

### 2. A repeating post-listen error is now an unbounded log stream instead of one crash (`server.mjs:613-616`)

The PR body covers the supervisor angle, but not the shape of the failure it replaces. `net.Server`'s `onconnection` emits `'error'` on the server **once per failed accept** (`self.emit('error', errnoException(err, 'accept'))`). Under fd exhaustion (`EMFILE`/`ENFILE`) that is not a single event: every incoming connection re-triggers it. Previously the process died once and a supervisor brought back a server that could accept connections. Now every attempt writes a full stack trace through `handleGenericException` to stdout, and the server stays up in a state where it cannot serve anybody — a log that grows as fast as clients retry, and no automatic recovery.

Not a reason to go back to crashing, but the branch is currently "log everything, survive everything". Two cheap options: keep the fatal path for the accept-level resource codes (`EMFILE`, `ENFILE`, `ENOBUFS`), or log the first occurrence of each `e.code` with its stack and subsequent ones as a bare one-liner. Either keeps the "don't drop every connected player over a survivable error" goal while bounding the bad case.

### 3. `server.listening` as the persist guard says something slightly different from the invariant (`server.mjs:648-655`)

The comment states the rule correctly — "a process that never took over the port … must not write anything back". But `server.listening` means "the socket is bound *right now*", and it goes back to `false` after `server.close()` as well as before the first bind. Today nothing calls `close()`, so the two are identical and the earlier suggestion to drop the duplicate flag was right on the facts as they stand. The trap is for later: the obvious next change in this area is a graceful shutdown that stops accepting connections before saving, and the moment someone puts a `server.close()` ahead of the exit path, this guard silently turns the room unload and the statistics write into no-ops — a data-loss bug with no error message, in the exact code that exists to protect that data.

A one-line comment on the guard ("`server.listening` is also false after `close()` — do not close the server before this runs") would be enough to stop that.

## Smaller notes

- `handleFatalException` routes through `logError('GENERIC ' + origin, …)` (`server/logging.mjs:47-49`), so an unexpected *fatal* startup failure and a routine survivable exception produce textually identical lines and differ only by stream. `FATAL` instead of `GENERIC` for the fatal sibling would make the thing an operator most wants to find greppable.
- `process.exit(1)` immediately after `console.error` (`server.mjs:631`) is safe on Linux and Windows, where writes to a pipe are synchronous — my runs confirm the message survives `2>&1 | cat`. On macOS, pipe writes are asynchronous, so a developer running `node server.mjs 2>&1 | tee server.log` on a Mac can get the exit without the explanation. If that matters, `process.exitCode = 1` plus letting the (now handle-free) process end on its own would flush it; if not, it is worth knowing the message is not guaranteed on every platform.
- The `EADDRINUSE` line is ~205 characters, so it wraps to three lines in an 80-column terminal. That is a deliberate trade-off already discussed on this PR and the remedy now comes second, so it reads fine wrapped — just noting that the "wait a few seconds" sentence lands on line three and will be the part people miss.

The `beta` label was already present and I left it: this PR touches no engine files (`server.mjs`, `server/logging.mjs` and two test files only), the failure path is verified to exit cleanly without writing into a shared save directory, and the normal start plus shutdown path is unchanged, so it is safe for players to test.

ArnoldSmith86 (approved):

PR #3131, opened 2026-08-20
Add the 16 newest Noto Emoji to the icon pickers

VTT-AI-Agent:
Ships the 16 emoji that Emoji 16.0 and 17.0 added and VTT did not have yet, lists them in `assets/fonts/symbols.json` so both icon pickers offer them, and draws them from that artwork wherever an emoji appears — the picker grid and the text of a game's name or description, where the browser's own font has no glyph for them yet. This is the additive half of #3109: it adds new files only, and deliberately leaves every emoji that is already shipped exactly as it is.

🫩 face with bags under eyes · 🫪 distorted face · 🫯 fight cloud · 🫈 hairy creature · 🧑‍🩰 ballet dancer · 🫆 fingerprint · 🫍 orca · 🪾 leafless tree · 🫜 root vegetable · 🛘 landslide · 🪊 trombone · 🪉 harp · 🪎 treasure chest · 🪏 shovel · 🫟 splatter · 🇨🇶 flag of Sark

the 16 new emoji

## What changed

**156 new SVGs in `assets/noto-emoji/`** — the 15 new base emoji plus the flag of Sark, and the 140 skin-tone and gender sequences that the same upstream release adds for emoji VTT already ships (for example the skin-toned 👯 people with bunny ears). They are taken from [`googlefonts/noto-emoji` v2.051](https://github.com/googlefonts/noto-emoji), the flag from its `third_party/region-flags/waved-svg`. Every one of them is a file that did not exist before.

**16 entries in `assets/fonts/symbols.json`**, in Unicode `emoji-test.txt` order, with keywords from [emojilib](https://github.com/muan/emojilib). The 15 non-flag emoji show up in the symbol picker overlay *and* in the widget editor's icon search; 🇨🇶 appears in the symbol picker only, because the icon search skips the whole `Emoji - Flags` category as it already does for the 250+ existing flags. Like every other flag, it carries its region code `cq` as a search term.

**8 code points added to `skipForNotoMonochrome` in `client/js/symbols.js`.** Each `symbols.json` entry is offered twice, as a color image and as a glyph of the bundled monochrome Noto Emoji font. That font is not updated here, and it has no glyph for 🫪 🫯 🫈 🫍 🛘 🪊 🪎 and no ligature for the ZWJ sequence 🧑‍🩰, so those eight would have been offered as an empty box. They are now offered as color images only; the other seven have a glyph and get both variants. The change is confined to those eight characters — the function returns the same answer as before for all 9651 symbols that were already in the file.

**The new characters are drawn from their SVG rather than from the browser's font.** Both halves of that were missing, and shipping the artwork without them means the picker shows a box where the emoji should be:

- The picker grid previews a color emoji with whatever emoji font the operating system has, which for an Emoji 16.0/17.0 character is a "not defined" box on essentially every machine today. The grid now uses the bundled SVG for these 15, the way it has always done for the flags. They also get the box of a game icon: the "not defined" glyph is a third of an emoji's width, so without a size the artwork ended up tiny between its neighbours.
- The regex that replaces emoji in text with their artwork (a game's name and description, the welcome dialog, the game list) did not know 14 of the 16, so they stayed browser text. Its ranges now cover exactly those 14 code points — checked character by character over all of Unicode, the old and the new regex differ on nothing else.

**Two jest guards** in `tests/game-icons-metadata.test.js`, in the spirit of the ones already there: every emoji the pickers offer has its SVG in `assets/noto-emoji/`, and every one of them is recognized in text as the same file. The second one fails on the 14 characters as they were before this commit, so the next batch of emoji cannot be half-added without CI saying so.

## What this does *not* change

Nothing that ships today is touched: no existing SVG is redrawn, `NotoEmoji-VariableFont_wght.ttf` stays at 3.003, and no license or About-dialog text changes. The new artwork is therefore drawn in the current upstream style while the rest of the set is the older one — bringing the whole set to v2.051, and with it the redesign of ~790 existing emoji, which changes what existing library games look like, is what #3109 is for.

## Verification

- `npm test` — 1329 jest tests pass, including the two new guards.
- TestCafe `Deck editor: symbol pickers and JSON fallback` and `The symbol picker says an image-only search found nothing` pass locally.
- Drove both pickers in Chromium: all 16 emoji are present, each with the variants listed above and no tofu; the 15 new ones and the flags render from their SVG at the same size as the emoji around them, while every other emoji keeps the browser-font preview it had. Searching `orca`, `sark`, `cq`, `fight cloud`, `ballet`, `landslide`, `splatter`, `root vegetable` and `treasure chest` finds them in the symbol picker and in the widget editor's icon search; no console errors and no failed requests for the new SVGs.
- Loaded a game whose name is `Emoji test 🫩 🪎 🛘 🇨🇶 🧑‍🩰 😀` and confirmed the game list draws all six from `assets/noto-emoji/`.
- Rendered the seven monochrome glyphs from the shipped font to confirm they really are glyphs (screenshot above, right-hand side of each tile).

No widget property, routine function or tutorial is touched, so there is no new UI surface to build and no test room to add.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3131/pr-test (or any other room on that server)


ArnoldSmith86 (approved):

VTT-AI-Agent (commented):
## TL;DR

Clean, genuinely additive PR: I re-derived the claims and they hold — the 156 new SVGs are byte-identical to `googlefonts/noto-emoji` v2.051 (repo is now a superset of upstream's `svg/`, nothing from that release is missing), the 16 `symbols.json` entries are new keys in the right categories/positions, and the `skipForNotoMonochrome` change returns the *exact* same answer for all 9651 pre-existing symbols while skipping precisely the 7 code points the bundled `NotoEmoji-VariableFont_wght.ttf` cmap has no glyph for (plus the ZWJ ballet sequence). The one thing I'd push back on is deferring the *display* half: 14 of the 16 characters are still unreachable to `emojis2images()` and the picker grid previews color emoji with the browser's font, so the new artwork ships but mostly isn't what a user sees.

## Findings

### 1. The picker grid previews the new emoji with the *browser's* font, which is exactly the case where that fails — `client/css/fonts.css:552`

`#symbolPickerOverlay .emoji-color` only swaps in `var(--url)` on `:hover`, in `.fewResults` mode, or for `.emojiFlag`. In the default grid the entry renders as the raw character in whatever emoji font the OS has. For Emoji 16.0/17.0 characters that is a tofu box on essentially every machine today (Emoji 17.0 shipped Nov 2025; OS fonts lag by a year or more) — so scrolling `Emoji - Objects` shows a blank box for 🪊 🪎 🪏 🪉 and the user has to hover each one to find out what it is. The PR notes this and defers it to #3109, but it isn't really an artwork-refresh problem: the fix is independent of redrawing the existing set, and this PR is what puts the characters into the grid.

The mechanism already exists for exactly this reason — flags get `emojiFlag` because Windows can't draw them. Concretely: in `client/js/symbols.js:274`, add a second class when the symbol is one of the new code points, e.g.

```js
let className = 'emoji-color';
if(category == 'Emoji - Flags' || emojiNewerThanBrowserFonts(symbol))
  className += ' emojiFlag';
```

with `emojiNewerThanBrowserFonts()` a sibling of `skipForNotoMonochrome()` above it (same shape, same kind of comment: "the emoji the browser's own font cannot draw yet"), listing the 15 new code points. That gets you correct artwork in the grid with no CSS change at all, since `.emojiFlag` already carries the `background: var(--url)` rule. It would want revisiting once browser fonts catch up, which is fine — that's the same maintenance the monochrome skip list already carries.

### 2. The 14 standalone new characters still render as text in labels — `client/js/symbols.js:8`

I ran the `emojis2images()` regex against all 16 additions: only 🧑‍🩰 and 🇨🇶 match; 🫩 🫪 🫯 🫈 🫆 🫍 🪾 🫜 🛘 🪊 🪉 🪎 🪏 🫟 do not. So a game author who types or pastes one of them into a rich-text label gets browser text (tofu for most players) even though the correct SVG is now sitting in `assets/noto-emoji/`. Again deferred to #3109, but the asset is *here*, and the gap is a handful of range ends in the surrogate-pair ranges the regex already has:

- `\ud83d` … `\uded5-\uded7` → `\uded5-\uded8` (1F6D8 🛘)
- `\ud83e` … `\ude80-\ude88` → `\ude8a` plus `\ude8e-\ude8f` (1FA89 1FA8A 1FA8E 1FA8F)
- `\ude90-\udebd` → `\udebe` (1FABE), `\udebf-\udec5` → `\udec6`/`\udec8` (1FAC6 1FAC8)
- `\udece-\udedb` → `\udecd-\udedb` (1FACD) plus `\udedc`/`\udedf` (1FADC 1FADF), `\udee0-\udee8` → `\udee9-\udeef` (1FAE9 1FAEA 1FAEF)

If you'd rather keep this PR strictly "new files only", that's a defensible call — but then it's worth saying in the PR body that until #3109 lands, the new emoji are picker-only in practice, because that's the part a game author will notice.

### 3. No regression guard that every `symbols.json` emoji has its SVG — `tests/game-icons-metadata.test.js`

This is the classic silent failure for this kind of change: one mistyped `_fe0f` or a sequence added to the JSON but not to `assets/noto-emoji/` gives a broken `` in the picker with nothing failing in CI. That file already exists to guard exactly this class of drift for game-icons ("the kind a later run adding a few icons silently breaks", line 7), and the check is a few lines in its style:

```js
test('every emoji in symbols.json has its noto-emoji SVG', () => {
  for(const [ category, entries ] of Object.entries(symbols))
    if(category.startsWith('Emoji'))
      for(const symbol of Object.keys(entries))
        expect(fs.existsSync(path.join(assets, `noto-emoji/emoji_u${filename(symbol)}.svg`))).toBe(true);
});
```

(`filename()` being the same one-liner as `emojiToFilename()`/`imageURLFromSymbol()`.) I ran that check against this branch: all 9667 entries resolve, so it passes today and locks the invariant in for the next batch. A second cheap one, given how much of this PR rests on it: assert `skipForNotoMonochrome()` agrees with the shipped font's cmap, so bumping `NotoEmoji-VariableFont_wght.ttf` (which #3109 may do) can't silently leave a stale skip list behind.

### 4. Nitpick — `🇨🇶` is missing its region code — `assets/fonts/symbols.json`

Its Channel-Island neighbours carry theirs (`🇬🇬` has `"gg"`, `🇯🇪` has `"je"`), so `"cq"` would match the convention and make the flag findable by code like the others.

- **Provenance.** sha256 of `emoji_u1fac6/1fadf/1f6d8/1f46f_1f3fb/1f9d1_200d_1fa70.svg` matches `googlefonts/noto-emoji@v2.051` `svg/` byte for byte, and the flag matches `third_party/region-flags/waved-svg/emoji_u1f1e8_1f1f6.svg` byte for byte (and its 1000×1000 waved shape matches the existing flags). `comm` of upstream's full `svg/` listing against `assets/noto-emoji/` is empty in both directions of interest — the repo is a superset, so no sequence from that release was left out, including the same-tone/gender forms (👯🏻 is the condensed `1f46f_1f3fb`, mixed tones the expanded `1f9d1_1f3fb_200d_1f430_200d_1f9d1_1f3fc`; both complete).
- **Licensing.** v2.051's `svg/LICENSE` is Apache 2.0 and region-flags is public domain, so the About text in `client/room.html:598` stays accurate and needs no edit — despite the root `LICENSE` at that tag being OFL 1.1 (that one covers `fonts/`, which this PR doesn't touch).
- **`skipForNotoMonochrome` really is a no-op for what exists.** Ran old and new regex over all 9651 previous symbols: 0 differences. And the 7 newly skipped code points (1F6D8 1FA8A 1FA8E 1FAC8 1FACD 1FAEA 1FAEF) are exactly the ones absent from the font's format-12 cmap, while the 7 that stay (1FAE9 1FAC6 1FABE 1FADC 1FA89 1FA8F 1FADF) all have glyph IDs 1875–1881. The new character class is unanchored like its neighbours, which is the right behavior here: a future 🧑‍🫯‍🧑 entry gets skipped too, and the font has no glyph for that either.
- **Placement and shape of the JSON entries.** All 16 are new keys, none collide, `🫩` lands between 😴 and 😷 per `emoji-test.txt`, `🇨🇶` sorts CO→CP→**CQ**→CR→CU, and the keyword arrays follow the existing snake_case-name-first convention.
- **Scope rules.** No widget property and no routine function are added, so `validate_gamefile.js`/`jeCommands` need nothing and there's no new editor UI to build; nothing under `library/` is touched, and a tutorial-style demo room wouldn't add anything over the PR test server for a picker-content change. No existing file's meaning changes, so no FileUpdater migration and no state-hash impact — consistent with all 45 checks passing.

Added the `beta` label: this PR touches no engine file (`client/js/widgets/`, `compute.js`, `serverstate.js` are all untouched — it is new assets plus a picker-only skip list), it cannot affect how existing games run or sync, and CodeQL/TestCafe are green, so it is safe and useful for players to try on beta.

VTT-AI-Agent:
Thanks — all four findings were valid and all four are fixed in b3bbb198. Pushing the artwork without the two display halves really did mean the emoji ship but mostly are not what a user sees, so this no longer defers them to #3109.

**1. The picker grid previewed the new emoji with the browser's font.** Fixed the way you suggested: a `tooNewForBrowserEmojiFont()` sibling of `skipForNotoMonochrome()` lists the 15 new code points, and the picker adds the always-image class for them as well as for the flags.

Two deviations from the sketch. The class is now called `emojiAsImage` rather than `emojiFlag` — it is the same rule for the same reason (the browser cannot draw the character), and it is used by both, so naming it after the flags would have been misleading. And it needed a size: a character with no glyph gets the width of the "not defined" box, which is a third of an emoji's, so the treasure chest rendered at 11×28 next to 30×28 neighbours. `.emojiAsImage` now takes the game icons' `width/height: var(--symbolSize)`, which leaves the flags looking exactly as they did and puts the new emoji in line with the rest of the grid.

**2. 14 of the 16 were not matched by the emoji-in-text regex.** Extended, using your range list as the starting point but adding only the 14 code points that exist rather than the unassigned neighbours inside those ranges — `\ude80-\ude8a\ude8e\ude8f\ude90-\udec6\udec8\udecd-\udedc\udedf-\udeea\udeef` and `\uded5-\uded8`. Ran the old and the new regex over every code point from U+0000 to U+10FFFF: they differ on exactly 1F6D8 1FA89 1FA8A 1FA8E 1FA8F 1FABE 1FAC6 1FAC8 1FACD 1FADC 1FADF 1FAE9 1FAEA 1FAEF and nothing else, and all 14 have their SVG. Verified end to end by loading a game named `Emoji test 🫩 🪎 🛘 🇨🇶 🧑‍🩰 😀` — the game list draws all six from `assets/noto-emoji/`.

The regex moved out of `emojis2images()` to module scope as `emojiRegex` so a test can reach it; it was being recompiled on every call anyway.

**3. No regression guard.** Added both, in `tests/game-icons-metadata.test.js`: every emoji the pickers offer has its SVG, and every one of them is recognized in text as the same file (the match may drop a variation selector, which addresses the same file — 207 entries rely on that). The second one fails on exactly the 14 characters as they were before this commit, so it would have caught this gap. Skipped the cmap check: parsing `NotoEmoji-VariableFont_wght.ttf` in jest is a lot of machinery for one list, and #3109 bumping the font is the moment to reconsider the skip list anyway.

**4. `🇨🇶` missing its region code.** Added `"cq"` after `flag_sark`, matching `gg` and `je`.

`npm test` is at 1329 passing, and both TestCafe symbol-picker tests pass locally.

PR #3132, opened 2026-08-20
Cache the minified client bundles on disk and prebuild them before a restart

VTT-AI-Agent:

- [ ] **An approving review.** The branch was rebased onto current `main` (`08659629`), all 45 checks are green and a squash merge is queued with auto-merge — GitHub's branch policy only still wants one approval before it runs.

Starting the server spends almost all of its time minifying the client bundles, and the result only lives in memory — so every restart pays for it again. On a production checkout a cold `node server.mjs` needs about 27 s before it logs `Listening on` (3.9 s with `MINIFYJAVASCRIPT=false`): terser with `passes: 2` over the ~2.4 MB client bundle (twice — once in `compressJS` and again inside html-minifier-terser), clean-css level 2 over ~440 KB of CSS, and four `gzip` calls at `Z_BEST_COMPRESSION`. That is long enough for a deploy to give up waiting for the new server and kill it, which on 2026-08-20 turned one deploy into ~95 restarts and about an hour of downtime.

This adds a disk cache for that build plus a prebuild step, so the expensive work happens **before** the restart, while the old server is still serving.

## What changed

**`server/minify.mjs`** — `minifyHTML()` is now a thin wrapper around the build:

- **Where:** `<save dir>/minify-cache/`, with the save dir taken from `Config.directory('save')`. `VTT_SAVE_DIR` is honoured, so MAIN, beta and every PR server get their own cache. Created on demand.
- **Key:** a sha256 over the content of every input file (all CSS and JS lists, `client/room.html`, `client/editor.html`, `client/css/custom.css`, `validator/validate_gamefile.js`, dompurify, fflate, `assets/fonts/symbols.json`), everything that gets injected into the output (`serverName`, `externalURL`, `urlPrefix`, the full `Config.getClientConfig()` and the normalised `minifyJavascript` flag), the versions of terser, clean-css and html-minifier-terser, a `CACHE_FORMAT` constant — and `server/minify.mjs` itself, so that editing the build code invalidates the cache the same way editing an input file does. The file lists moved into module-level constants that both the build and the hash read, so the two cannot drift apart; the order is unchanged.
- **Stored shape:** one directory per key holding exactly what `minifyHTML()` returns (`min`, `gzipped`, `editorJSmin`, `editorJSgzipped`, `fflateMin`, `fflateGzipped`, `symbolsGzipped`) as plain files — no base64 inflation — plus an `entry.json` with the format, the key and the byte size of every file. It is written to a temporary directory and renamed into place, so a crash mid-write can leave a stale temp directory (cleaned up later) but never half an entry. The five most recently used entries are kept.
- **Anything unexpected is a miss:** a missing, truncated, foreign or unparsable entry is discarded and the bundles are built as before. Same for a cache directory that cannot be created or written — that logs a warning and builds in memory. A cache problem can never stop the server from starting.
- **One log line** says which happened: `Client bundles: cache hit 9c9fd077` or `Client bundles: cache miss, built in 25.5 s, stored as 4e102254`.

**`server/prebuild.mjs`** (new) — builds and stores the bundles without opening a port or touching anything besides the cache directory. It loads Config exactly like the server does (`config.json` from the current directory, environment overrides), so the deploy scripts run it in the checkout the new server will start from, with the same config and environment. Exits 0 once the bundles were built, non-zero if the build failed; a cache that cannot be written is reported but is not an error, since the server still starts.

**Second commit** runs the gzip calls for the editor bundle, fflate and `symbols.json` concurrently instead of one after another. They do not depend on each other and zlib compresses in the libuv threadpool. The output is byte-for-byte identical; measured here it takes 361 ms instead of 484 ms.

No client-side change, and nothing about what is served changes.

## Measured

On this checkout with `MINIFYJAVASCRIPT=true`:

```
node server/prebuild.mjs   ->  cache miss, built in 25.5 s, stored as 4e102254   (26.1 s total)
node server/prebuild.mjs   ->  cache hit 4e102254                                 (0.5 s total)
node server.mjs            ->  cache hit 9c9fd077 ... Listening on 8333          (1.4 s to first response)
```

Changing `serverName` (or any input file, or `config.json`) and restarting without a prebuild logs a miss and builds as before. The bytes served from a cache hit were compared against a fresh uncached build — `min`, `gzipped`, `editorJSmin`, `editorJSgzipped`, `fflateMin`, `fflateGzipped` and `symbolsGzipped` are identical, and the `room.html`, `/edit.js` and `/scripts/fflate` the running server sends match the cached files exactly.

## Tests

`tests/server/minify-cache.test.js` covers: the same checkout produces the same key; a changed input file, a changed client config value and a changed `minifyJavascript` flag each produce a different one; the key does not depend on where the checkout lives; an entry round-trips every field byte-identically with strings coming back as strings and buffers as buffers; a truncated file or an unreadable `entry.json` is refused; `minifyHTML()` rebuilds instead of crashing when a stored entry is corrupted, and the rebuild is identical to the original; `prebuild.mjs` exits 0 and fills the cache on the first run, hits it on the second, and exits non-zero when a client file does not parse. The tests that need a checkout of their own build one out of symlinks in a temporary directory, so no file of the real one is ever modified, and they use a temporary `VTT_SAVE_DIR`.

The build hooks in the existing `tests/server/minify.test.js` got a longer timeout: two full builds already took most of the old minute, and they now compete for the same cores with the new test file.

`npm test` is green (38 suites, 1344 tests).

## For the deploy scripts

`misc/server-scripts/puppeteer/main-update.sh` and `beta-update.sh` already call `node server/prebuild.mjs` if the file exists, between updating the checkout and stopping the old server. This PR is the other half of that.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3132/pr-test (or any other room on that server)

### DIFF
--- a/server/minify.mjs
+++ b/server/minify.mjs
@@ -1,6 +1,8 @@
+import crypto from 'crypto';
 import fs from 'fs';
 import os from 'os';
 import path from 'path';
+import url from 'url';
 import util from 'util';
 import zlib from 'zlib';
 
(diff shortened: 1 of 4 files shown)


VTT-AI-Agent (commented):
## TL;DR

Solid, well-guarded change: every cache failure path falls back to a normal in-memory build, and the key is derived from the same file lists the build itself uses, so it cannot drift. I reproduced the whole thing locally (prebuild → 4.2 s miss, server → cache hit, listening in ~0.3 s) and byte-compared what the running server sends against the stored entry — `room.html`, `room.html.gz`, `/edit.js`, `/scripts/fflate` and `symbols.json.gz` are all identical, and touching `client/js/audio.js` correctly produced a miss on the next start. The one thing I'd change before merge: the `fs.utimesSync()` bookkeeping call sits *inside* the cache-hit `try`, so an entry the current user can read but doesn't own turns a good hit into a full 27 s rebuild plus a misleading warning — exactly the deploy shape (prebuild as the deploy user, server as the service user) this PR is written for.

## Findings

### 1. A failing `utimesSync` throws away a perfectly good cache hit — `server/minify.mjs:511-519`

```js
const build = loadFromCache(directory, key);
fs.utimesSync(`${directory}/${key}`, new Date(), new Date());  // youngest by use, not by build
Logging.log(`Client bundles: cache hit ${shortKey}`);
return build;
```

`utimes(2)` needs ownership of the target (or `CAP_FOWNER`), not write permission on it — and the `fs.accessSync(directory, W_OK)` guard above only covers the *directory*, not the entry inside it. So in the deploy layout this PR is built for — `node server/prebuild.mjs` run by the deploy user into a `minify-cache/` directory the service user created and can write, with the new entry owned by the deploy user — the load succeeds, `utimesSync` throws `EPERM`, control falls into the `catch`, and the server logs

```
WARNING - Client bundles: discarding cache entry 4e102254 (EPERM: operation not permitted, utime ...)
```

and rebuilds for ~27 s. That is both the failure this PR exists to prevent and a message that points at entry corruption rather than at file ownership, so it would be genuinely hard to diagnose from a deploy log. The LRU stamp is a nicety; it shouldn't be able to reject a build that already loaded cleanly:

```js
const build = loadFromCache(directory, key);
try {
  fs.utimesSync(`${directory}/${key}`, new Date(), new Date());  // youngest by use, not by build
} catch(e) {
  // an entry written by another user can still be read, it just cannot be re-stamped
}
```

(Same reasoning applies to `pruneCache()` deleting entries it doesn't own, but there the `rmSync` failure only mislabels the log line — see 3b.)

### 2. `tests/server/minify.test.js` now asserts against the cache instead of against the build — `tests/server/minify.test.js:4`, `server/minify.mjs:246`

`buildWith()` still calls the default export, which after this PR is the *cached* path, and that file does not set `VTT_SAVE_DIR` — so it reads and writes the checkout's real `save/minify-cache/`. Running it here:

```
Client bundles: cache miss, built in ...      <- describe #1, buildWith('false')
Client bundles: cache hit 68039543            <- describe #2, buildWith('false') again
Client bundles: cache miss, built in 27.6 s   <- describe #2, buildWith('true')
$ du -sh save/minify-cache  ->  6.3M
```

Two consequences: (a) `npm test` now leaves ~6 MB in the developer's `save/` (gitignored and `npm run clobber` clears it, so cosmetic), and (b) more importantly the minifier tests can now be satisfied by bytes that were produced by an *earlier* process. Today that is safe because the key covers every input, but these are precisely the tests that would catch a regression in `compressJS`/`removeExportStatements` — and if the key ever gains a blind spot, they'd pass against a stale entry and hide it. The PR already exports `buildHTML()` (currently exported but imported by nothing), so the fix is a one-word import change:

```js
import { buildHTML } from '../../server/minify.mjs';
...
return await buildHTML();
```

That keeps those tests hermetic and fast to reason about, gives the new export a caller, and leaves `minify-cache.test.js` as the only place that exercises the cached path. It also removes the ordering coupling where the second `describe` gets a hit only because the first one ran.

### 3. Smaller things

**a) A partially deleted entry is a silent miss** — `server/minify.mjs:516-518`. `loadFromCache` reads `entry.json` first and each payload file after it, and the caller suppresses *any* `ENOENT`. So "no entry at all" (normal, silent) and "entry.json is there but `symbols.json.gz` was removed" (abnormal, e.g. a `pruneCache` race or a half-cleaned directory) log identically. Checking `fs.existsSync(`${directory}/${key}/entry.json`)` before the load, and warning on every failure after that point, would make the abnormal case visible without adding noise to the normal one.

**b) A `pruneCache` failure is reported as "could not be stored"** — `server/minify.mjs:525-531`. `pruneCache()` runs inside the same `try` as `storeInCache()`, and it `statSync`s names it read from `readdirSync` — a concurrent prebuild removing an entry in between (or an entry owned by another user that `rmSync` cannot remove) throws, and the operator sees `cache miss, built in 25.5 s, could not be stored` even though the entry landed fine. Its own `try`/`catch`, or moving it after the success log, keeps the message honest.

**c) Integrity is byte-length only** — `server/minify.mjs:443-446`. With the write-to-temp-then-rename design a torn write is already impossible, so the size check mostly guards against manual meddling and disk corruption — and it misses any corruption that preserves length. A sha256 per file in `entry.json` would close that at ~20 ms for a 6 MB entry; your call whether that's worth it given the rename already rules out the realistic case.

**d) The temp directory name can collide across containers** — `server/minify.mjs:453-455`. `.tmp-${process.pid}-${key.slice(0,8)}` is unique per host, but two containers sharing one save volume can hold the same PID, and the unconditional `fs.rmSync(temporary, ...)` on entry would then delete the other process's in-flight directory. Appending `crypto.randomUUID().slice(0,8)` removes the question entirely.

- **Key completeness.** I walked every read in `buildHTML()` — `ROOM_HTML`, `CLIENT_CSS`, `CLIENT_JS`, `EDITOR_HTML`, `EDITOR_CSS`, `EDITOR_JS`, `FFLATE_JS`, `SYMBOLS_JSON`, plus the injected `serverName`/`externalURL`/`urlPrefix`/`getClientConfig()`/`minifyJavascript()` and `minify.mjs` itself — and everything the build depends on is hashed. Hashing the module's own source via `import.meta.url` rather than a cwd-relative path is the right call given prebuild may run from a different directory. Hoisting the lists into module constants so the build and the hash cannot diverge is the part that makes this maintainable.
- **Atomicity.** Write-to-temp + `renameSync` within the same filesystem, `entry.format`/`entry.key` self-check on load, `EEXIST`/`ENOTEMPTY` from a parallel build treated as success. All correct.
- **No engine or client change.** `server/minify.mjs` only reorganises the build and adds the cache in front of it; `compressCSS`/`compressJS`/`compress` are unchanged apart from `readInputFile()`, and `readInputFile(...).toString('utf8')` is equivalent to the old `readFileSync(..., {encoding:'utf8'})`. The types the server relies on survive the round trip (`min`/`editorJSmin` come back as strings, everything else as Buffers) and `minify-cache.test.js` asserts exactly that.
- **Concurrent gzip (second commit).** `editorJS`, `fflate` and `symbols` have no dependency on each other and zlib runs in the libuv threadpool; output is unchanged.
- **Cache location.** `save/minify-cache/` is not enumerated by anything else in the save directory (only `rooms/`, `states/`, `links/`, `errors/`, `shares.json`, `statistics.json` are), is not reachable through `/i` (that maps to `save/assets`), and `save/` is gitignored. ~3 MB per entry × 5 ≈ 16 MB per server.
- **Local verification.** `npx jest tests/server/minify-cache.test.js` → 13/13 green; `tests/server/minify.test.js` → 9/9 green. `node server/prebuild.mjs` → `cache miss, built in 4.2 s`; `node server.mjs` → `cache hit 12445fc5` then `Listening on 8350` ~20 ms later; `cmp` of the served `room.html`, `room.html.gz`, `/edit.js`, `/scripts/fflate` and gzipped `symbols.json` against the stored files → identical; appending a comment to `client/js/audio.js` and restarting → `cache miss, built in 3.3 s`. All GitHub checks are green, including CodeQL.
- **Maintainer rules that don't apply here.** This PR exposes nothing new to game authors — no widget property and no routine function/operation — so there is no editor/property-UI or routine-block-UI requirement and no demo room to build, and the diff contains no change under `library/`.

The `beta` label was added: the diff touches only `server/minify.mjs`, the new `server/prebuild.mjs` and two test files — no engine file (`client/js/widgets/`, `client/js/compute.js`, `client/js/serverstate.js`) at all, so the engine exclusion does not apply, and the change is safe to test on beta because every cache failure falls back to the normal in-memory build and I confirmed the bytes served from a hit are identical to a fresh build.

VTT-AI-Agent:
Thanks — all four points were valid, and 3a turned out to hide a second problem. Pushed as 1ea0a8c2.

**1. `utimesSync` throwing away a good hit** — fixed. The stamp now has its own `try`/`catch` and a comment saying why an entry another user wrote can be read but not re-stamped: it only loses its place in the pruning order, it never invalidates a build that already loaded. `tests/server/minify-cache.test.js` covers it (`serves an entry it cannot stamp as used`) by making `fs.utimesSync` throw `EPERM` and asserting the returned build is still the cached one.

**2. `tests/server/minify.test.js` going through the cache** — fixed exactly as suggested: `buildWith()` calls the now-imported `buildHTML()`, so those tests assert against what the minifiers produce, run hermetically, and no longer leave `save/minify-cache/` behind (verified: after `npx jest tests/server/minify.test.js` the save directory is untouched). `minify-cache.test.js` remains the only test of the cached path.

**3a. Partial entry was a silent miss** — fixed, and it exposed more than a logging gap. `loadFromCache` is now only attempted when `entry.json` exists, and every failure past that point warns. Walking that through on a real checkout showed the damaged entry could never be repaired: `storeInCache` renames onto `${directory}/${key}`, the rename fails with `ENOTEMPTY`, and the existing `existsSync` check treats that as "a parallel build got there first" — so the rebuilt bundles were dropped and the broken entry stayed, warning and rebuilding on every single start. Storing now checks whether an entry already under that key actually loads and replaces it if it does not. Reproduced end to end: deleting `fflate.js.gz` from an entry made the next `node server/prebuild.mjs` warn, rebuild and repair it, and the run after that hit again.

**3b. `pruneCache` failure mislabelled as "could not be stored"** — fixed; pruning moved out of the store `try` into its own, with its own message. It also no longer throws when another process removes an entry between the `readdir` and the `stat` (`statSync(..., { throwIfNoEntry: false })`), which is the likeliest way it would have failed.

**3d. PID-based temp directory name** — fixed with a `crypto.randomUUID().slice(0, 8)` suffix. The unconditional `rmSync` of the temp directory on entry is gone with it, since a fresh random name cannot collide with anything.

**3c. sha256 per file** — not doing it, for the reason you gave yourself: with write-to-temp + rename a torn write is impossible, so the length check plus the `format`/`key` self-check already cover truncation and foreign entries, and what is left is bit rot that happens to preserve the byte length — at which point the save directory has bigger problems. It also sits on the hit path, which is the one thing this PR is spending its complexity budget on. Happy to add it if you'd rather have the belt and braces.

Verified after the change: `npm test` → 38/38 suites, 1342 tests green; `node server/prebuild.mjs` → `cache miss, built in 3.7 s`, second run `cache hit`; server start → `cache hit 6647cc32` then `Listening on 8333` 35 ms later, and the served `/edit.js` is byte-identical to the stored entry; an unwritable cache directory still falls back to `building in memory`.

VTT-AI-Agent:
Updated to current `main` and re-checked.

- Rebased the three commits onto `main` (`08659629`, the Noto Emoji PR #3131) — no conflicts, the branch is linear again.
- That update is a good exercise of the cache itself: `client/js/symbols.js`, `client/css/fonts.css` and `assets/fonts/symbols.json` are all hashed inputs, so the old entries stopped matching and a fresh `node server/prebuild.mjs` built and stored a new one (27.3 s). The second prebuild hit it (0.0 s) and the server started off that hit; the `symbols.json` it serves gzipped is byte-identical to the file on disk, new emoji included.
- `npm test`: 38 suites, 1344 tests, green. `tests/testcafe/editor.js` (including the symbol-picker tests) passes locally against a server running off the cached, minified bundles.
- All 45 GitHub checks are green on `43fadaf`.

Nothing in the change became redundant or wrong through the update; the only edit was the test count in the description.

A squash merge is queued with auto-merge, so it lands as soon as someone approves.

ArnoldSmith86 (approved):

Issue #1212, opened 2022-05-13
Holder Events


robartsd:
We know that holder events don't work as we want them to, but we haven't really defined how we want them to work. This is a draft of what I think should happen.

### Widget dragged from HolderA to HolderB

1. mousedown/touchstart followed by mousemove/touchmove triggers beginning of drag
2. Widget.`parent` set to null
3. call HolderA.`beforeLeaveRoutine` (mitigate that leave routine no longer fires twice by adding `beforeLeaveRoutine` that calls `leaveRoutine` via fileupdater)
4. mousemove/touchmove events move widget - HolderA is no longer hoverTarget
5. change Widget properties based on HolderA.`childrenPerOwner` and HolderA.`onLeave`
6. call HolderA.`leaveRoutine` 
7. mousemove/touchmove events move widget - HolderB becomes hoverTarget
8. mouseup/touchend event completes the move
9. Widget.`parent` set to HolderB.`id`
10. change Widget properties based on HolderB.`childrenPerOwner`, HolderB.`onEnter`, HolderB.`alignChildren`
11. resolve piles
12. call HolderA.`afterLeaveRoutine`
13. call HolderB.`enterRoutine`

### Widget dragged into HolderB

1. mousedown/touchstart followed by mousemove/touchmove triggers beginning of drag
7. mousemove/touchmove events move widget - HolderB becomes hoverTarget
8. mouseup/touchend event completes the move
9. Widget.`parent` set to HolderB.`id`
10. change Widget properties based on HolderB.`childrenPerOwner`, HolderB.`onEnter`, HolderB.`alignChildren`
11. resolve piles
12. call HolderB.`enterRoutine`

### Widget dragged out of HolderA

1. mousedown/touchstart followed by mousemove/touchmove triggers beginning of drag
2. Widget.`parent` set to null
3. call HolderA.`beforeLeaveRoutine`
4. mousemove/touchmove events move widget - HolderA is no longer hoverTarget
5. change Widget properties based on HolderA.`childrenPerOwner` and HolderA.`onLeave`
6. call HolderA.`leaveRoutine` 
8. mouseup/touchend event completes the move
9. resolve piles
12. call HolderA.`afterLeaveRoutine`

### Widget dragged within HolderA (Holder is always hoverTarget during the move)

1. mousedown/touchstart followed by mousemove/touchmove triggers beginning of drag
2. Widget.`parent` set to null
3. call HolderA.`beforeLeaveRoutine`
[...]

PR #3134, opened 2026-08-20
Fire holder enter and leave events once per move, whichever way the widget moved

### LINKED ISSUE #480: leaveRoutine executes twice
I'm just adding this as an issue so it isn't forgotten with everything else being worked on.

leaveRoutine is activated once when a widget is picked up, and again when it leaves the area of the holder.  Considering that enterRoutine is activated when the widget is dropped, it would probably be more consistent if leaveRoutine was only activated when the widget is picked up.

This room is set up to demonstrate this issue.  The number increments when either leaveRoutine or enterRoutine is activated.

https://virtualtabletop.io/rgp2

### LINKED ISSUE #1094: enterRoutine or leaveRoutine and dropLimit strange behavior
Here's an odd one.  Demo in https://virtualtabletop.io/dgax.

This is part of a room from one of my new solitaire games.  All the cards have little holders at the bottom so they stack vertically and can be moved around the room as a sort of group (but not really a pile widget).  That works fine.  But @ArnoldSmith86 noticed an issue when you try to stack them on top of one another in a holder where you could accidentally stack onto the holder to create a new vertical stack instead of making an actual pile in the holder.  So to fix that, I added an enterRoutine and leaveRoutine that makes the dropLimit for the small holder on the card = 0.  That sort of works fine if there is just one card in the holder.

By that I mean if you just try to drag a card in there, the small holder is deactivated because the dropLimit is 0.  But after you drop that card in there, if you pick it up again, the holder reactivates to dropLimit 1 (probably because of the leaveRoutine).  But the leaveRoutine shouldn't fire on the card that is in the holder because the routine uses the pre-defined collection "child" which shouldn't involve the card already in the holder.

Also, if the holder has more than one card it in, some (or maybe all; I don't know) of the cards have their holders active with a dropLimit of 1.

On the right side of the demo room is a holder where I've manually set the card dropLimit to 0.  That is how this is supposed to work.  I'm sure the problem has something to do with either leaveRoutine firing twice or when it isnt' suppoed to.  Or maybe its an issue with the pre-defined collection "child" not working right.  It's probably the leaveRoutine and related to #495 or #614.  But I can't figure it out and just wanted to document it.

### LINKED ISSUE #1371: MOVEXY should respect onLeave
at the moment MOVEXY doesn't respect the onLeave property of a holder. I can set the face, but I think the default behaviour should be to behave according to onLeave

### LINKED ISSUE #1836: Inconsistent `onEnter` and `onLeave` with `SET`
Assume a room with holder1 and holder2,  holder1 has `onEnter` and `onLeave` properties established.  When you drag and drop things in holder1, the `onEnter` and `onLeave` work properly.  When you progamatically use `MOVE` to move back and forth between the holders, everything works.

But, when you use `SET` parent to or from holder1, that is where things break down.  `SET` parent to holder1 triggers the `onEnter`, but when you take a widget that is in holder1 and `SET` its parent to holder2, `onLeave` does not get triggered.

This is sort of related to #1212, but in that issue, there is no discussion of `SET` parents.

VTT-AI-Agent:
A widget moving into or out of a holder now raises **at most one leave and one enter**, and each of them applies its properties before it calls its routine — whichever way the widget was moved. Until now two different places raised the same leave event, and the operations that only went through one of them applied half of it, which is what the five issues below are about.

## What was wrong

`moveStart()` cleared the widget's `parent`, and `onPropertyChange('parent')` called the old parent's `leaveRoutine` right away — before anything else about the departure had happened. A moment later `checkParent()` detached the widget for real and `Holder.dispenseCard()` applied `onLeave` and called `leaveRoutine` **again**. Everything else was the same table read from a different column:

- a drag and a `MOVE` called `leaveRoutine` twice (#480),
- `SET parent` and `MOVEXY` never reached `dispenseCard()`, so they applied no `onLeave` at all (#1836, #1371),
- a card that joined a pile inside the holder it was already in raised a leave with no matching enter (#1094),
- and a card dropped back into the holder it was picked up from ran `enterRoutine` without applying `onEnter`.

## The model that replaces it

1. **The parent change is the only firing site.** A drag, `MOVE`, `MOVEXY`, `SET parent`, `CLONE`, `RECALL`, `DELETE` and the JSON editor all end up writing `parent`, and that write *is* the event.
2. **Properties first, routine second.** A leave applies the `onLeave` properties, re-compacts a stacked holder, then calls `leaveRoutine`; an enter applies `childrenPerOwner`'s owner and the `onEnter` properties, then calls `enterRoutine`. A routine always reads the values its own event wrote.
3. **A leave is complete before the matching enter starts.**
4. **A pile is transparent inside whatever holds it.** The container an event belongs to is the parent, or the parent's parent when the parent is a pile — so forming, joining, leaving or dissolving a pile inside a holder raises nothing, and taking a card out of one leaves the holder exactly once. A pile *on the table* has nothing behind it and stays a container of its own, so its own `enterRoutine`/`leaveRoutine` fire as before.
5. **A widget on its way out of the room gets no properties.** `DELETE` still tells the holder that it lost a card — and a stacked holder still closes the gap it left — there is just nothing left to apply `onLeave` to.
6. **A per-owner hand lets go of a dragged card only when the card is outside it.** `childrenPerOwner` is the one part of a leave that is visible to the *other* players: a card without an owner loses the `foreign` class, which is what hides everyone else's hand. A drag detaches the widget at the pickup, long before it is anywhere else, so the hand keeps the card until `checkParent()` decides it is really out of the box — rearranging a hand never uncovers a card. Every other way of taking a card out (`SET parent`, `MOVEXY`, `MOVE`) releases it with the rest of the leave — except when `MOVEXY` says `resetOwner: false`, which is the operation's explicit "put this on the table but keep it owned".
7. **A drop-shadow preview is not a move.** A holder with `dropShadow` paints a clone of the dragged widget where the card would land, and that clone is parented into the holder and taken out again while the drag is still going on. It used to run `enterRoutine`, `leaveRoutine`, `onEnter` and `onLeave` for a drop the player had not made — so merely hovering over a holder could score, deal or delete. A parent change of a widget carrying `dropShadowOwner` now raises nothing; the preview still takes its slot in a stacked holder, and that slot closes again from `onChildRemove()` when the pointer moves on.

Unchanged on purpose: a *holder* changing parent still does not run the parent's `leaveRoutine`, a `deck` is still exempt from `onEnter` and `childrenPerOwner`, `ignoreOnLeave` still skips the property half but not the routine, and a pile is still one widget for the routines and a stack of cards for the properties.

Closes #480. Closes #1094. Closes #1371. Closes #1836.

This adopts the parts of #1212 that need no new API. The `beforeLeaveRoutine`, `afterLeaveRoutine` and `leaveCancelledRoutine` it proposes are new routine properties and belong in their own PR with their own editor and validator support, so #1212 stays open. The one remaining difference to its description is that `leaveRoutine` runs after the `parent` write rather than before it — which is what makes a single hook possible at all.

## Backwards compatibility

New legacy mode **`legacyHolderEnterLeaveEvents`** (file version 22 → 23). The file updater switches it on for every save older than version 23 that either mentions `enterRoutine`, `leaveRoutine`, `onEnter`, `onLeave` or `childrenPerOwner`, or stacks children anywhere — a widget carrying a non-zero `stackOffsetX`/`stackOffsetY` (which may be the template a holder inherits from or clones rather than the holder itself), or a routine that sets one while the game runs. A stacked holder notices the re-compaction even when it uses none of the event names. Existing games therefore keep the pipeline they were built against. The detector is deliberately over-inclusive, as `docs/compat.md` asks: `childrenPerOwner` is in the list because the owner half of the change is not gated by any of the four routine/property names. The Game Settings sidebar picks the checkbox up from the registry automatically. Its tile is labelled **Fire holder leave events twice** and lists the seven differences as bullets — [1280×800](https://agent.virtualtabletop.io/reports/pr/1539994146897399808/holder-events-legacy-tile.png), [390×844](https://agent.virtualtabletop.io/reports/pr/1539994146897399808/holder-events-legacy-tile-phone.png).

A game author who wants the new pipeline unchecks it and looks for seven things:

- a `leaveRoutine` that counted or dealt now runs once per departure instead of twice;
- `SET parent` and `MOVEXY` now apply `onLeave`;
- picking a card up and dropping it back into the *same* holder is a real leave and a real enter, so both apply their properties (previously neither did);
- a `childrenPerOwner` holder now releases the owner for `SET parent` and `MOVEXY` as well, not only for drags and `MOVE` — `MOVEXY`'s own `resetOwner: false` still keeps it, in both pipelines;
- a stacked holder now closes the gap a card left however it left — `SET parent`, `MOVEXY` and `DELETE` included, where the old pipeline left the hole open;
- hovering a `dropShadow` holder no longer runs its `enterRoutine`/`leaveRoutine` or applies `onEnter`/`onLeave` for the preview;
- and the `_ancestor` workaround for #1094 is dead code.

The drop-shadow half is the one entry on that list that is a plain bug rather than a design decision, and it is gated like the rest: `legacyHolderEnterLeaveEvents` restores the old engine as a whole, so a migrated game keeps the preview's events until its author unchecks the mode. A game *can* observe them — a holder whose `enterRoutine` lights something up while a card hovers over it is the plausible case — so it is not something the migration takes away silently.

## Tests

- `tests/client/engine/enterleave.test.js` — 76 jsdom assertions over dragging, `MOVE` (holder / collection / seat), `MOVEXY`, `SET parent`, `CLONE`, `DELETE`, `childrenPerOwner`, `ignoreOnLeave`, drop shadows, multi-property `onEnter`, decks, holders inside holders, pile formation and dissolution and the ordering invariants — each with the mode off *and* on.
- `tests/client/engine/bundle-widgets.js` — new: evaluates the real `Holder`, `ImageWidget` and `Pile` classes inside jsdom the way the shipped bundle concatenates them, so a holder assertion costs a few lines instead of a browser fixture.
- `tests/testcafe/holderevents.js` — the existing fixture that pinned the old behaviour, updated to the new traces. Its diff is the review of this change. It also pins that hovering a `dropShadow` holder and then dropping somewhere else leaves the room exactly as it was.
- `tests/testcafe/enterleave.js` — new: every legacy trace with the mode on (proving old games keep working), the pile cases that need a real pointer, stacked and `preventPiles` holders, seats and per-owner hands — including a rearrange inside a per-owner hand, which is the case where the owner must survive the whole drag and where the drag must leave nothing behind that keeps the *next* card owned, `MOVEXY`'s `resetOwner: false`, and the stacked-holder layout with the mode on and off.
- `tests/server/fileupdater.test.js` — the stacked-holder half of the detector: the bare stacked holder, the offset written as an expression, the offset a routine sets at runtime, the template a holder inherits it from, and the two that must *not* trigger it (`stackOffsetY: 0`, `alignChildren: false`).
- `tests/testcafe/holderevent-util.js` — new: the probe game both browser fixtures drive.
- `tests/testcafe/multiclient.js` — one case added: a second browser window watches Alice rearrange her per-owner hand, and the card stays `foreign` there for the whole gesture. `dragPath()` gained a `hold` option so a fixture can assert in the middle of a drag.

Locally: `npm test` green (1569); TestCafe green on `enterleave.js`, `holderevents.js`, `legacymatrix.js`, `interactions.js`, `multiclient.js`, `card.js`, `draglimit.js` and `routines.js`, plus a `corpusreplay.js` sample of the public library.

## Report and demo rooms

A detailed HTML report — the model, the before/after trace of every case, and what each of the five issues turns into — is at
https://agent.virtualtabletop.io/reports/pr/1539994146897399808/lm1-holder-enter-leave.html

Three manual test rooms ship as **Properties - Holder events** in this PR's `pr-test-ai` test-server room, with a live event log and buttons that toggle the properties while you play:

- **Dragging** — `childrenPerOwner`, `preventPiles`, `stackOffsetY`, `alignChildren`, `dropShadow`, `ignoreOnLeave` ([screenshot](https://agent.virtualtabletop.io/reports/pr/1539994146897399808/holder-events-dragging.png))
- **Routines** — one button per operation: `MOVE`, `MOVEXY`, `SET parent`, `CLONE`, `DELETE`, `RECALL`, `FLIP` ([screenshot](https://agent.virtualtabletop.io/reports/pr/1539994146897399808/holder-events-routines.png))
- **Piles & hands** — a holder that takes piles, one that refuses them, a per-seat hand and a seat ([screenshot](https://agent.virtualtabletop.io/reports/pr/1539994146897399808/holder-events-piles-and-hands.png))

They are authored to tutorial quality but are deliberately **not** part of the diff: the public library stays untouched. They can be added to `library/tutorials/` on request. Each room's overview says that switching *Fire holder leave events twice* on in Game settings shows the pipeline this replaced, which is the fastest way to see the difference.

The wiki's *Legacy-Mode* page and the holder pages will want a line about the new mode.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3134/pr-test (or any other room on that server)

### DIFF
--- a/client/css/editor/sidebarModules.css
+++ b/client/css/editor/sidebarModules.css
@@ -1932,6 +1932,13 @@ div.widget-name {
   color: var(--textColor);
 }
 
+/* a bulleted list in a tile's description text - the browser default indents it far past
+   the 21px column the rest of the text sits in */
+.settings-tile ul {
+  margin: 6px 0;
+  padding-left: 20px;
+}
+
 .boardSizeSetting .boardSizeHeader {
   font-weight: bold;
   margin-bottom: 8px;
--- a/client/js/legacymoderegistry.js
+++ b/client/js/legacymoderegistry.js
@@ -95,7 +95,7 @@ export const LEGACY_MODES = {
   disableHolderImageWidget: {
     since: 21,
     pr: 2634,
-    interactsWith: [],
+    interactsWith: [ 'legacyHolderEnterLeaveEvents' ],
     detect: function(state) {
       for(const id in state) {
         const properties = state[id];
@@ -113,6 +113,26 @@ export const LEGACY_MODES = {
       <b>New behavior</b>: Holders display image, icon and text properties directly. Games that built those manual workarounds can look broken because both are drawn at once.
       <br><br>
       This legacy mode disables the native image/icon/text support for holders, restoring the old behavior.
+      `
+  },
+  legacyHolderEnterLeaveEvents: {
+    since: 23,
+    pr: 3134,
+    interactsWith: [ 'disableHolderImageWidget' ],
+    detect: state => /"(enterRoutine|leaveRoutine|onEnter|onLeave)"\s*:/.test(JSON.stringify(state)),
+    label: 'Legacy holder enter/leave events',
+    summary: 'leaveRoutine fires twice per move, onLeave only for drags and MOVE.',
+    description: `
+      <b>Old behavior</b>: Two different places raised the leave event, so every drag and every <code>MOVE</code> called <code>leaveRoutine</code> twice - once when the widget was detached and once when it was really out of the holder. The <code>onLeave</code> properties were only applied on the second of those, which meant <code>SET parent</code> and <code>MOVEXY</code> never applied them at all. A card that joined a pile inside the holder it was already in raised a leave event with no matching enter, and a card dropped back into the holder it was picked up from ran <code>enterRoutine</code> without applying <code>onEnter</code>.
+      <br><br>
+      <b>New behavior</b>: A parent change raises at most one leave and one enter, whichever way the widget was moved. Each event applies its properties first and calls its routine afterwards, so the routine always sees the values the event wrote. Piles are transparent: joining, leaving or dissolving a pile inside a holder is not a move between containers and raises no event.
+      <br><br>
+      <b>Example:</b> a holder whose <code>leaveRoutine</code> counts the cards that left it.
+      <br><br>
+      Old result: dragging one card out counts <code>2</code><br>
+      New result: dragging one card out counts <code>1</code>
+      <br><br>
+      Enable this mode for a game that was built around the double call - or around <code>MOVEXY</code> not applying <code>onLeave</code>.
       `
   }
 };
--- a/client/js/legacymoderegistry.js
+++ b/client/js/legacymoderegistry.js
@@ -120,19 +120,23 @@ export const LEGACY_MODES = {
     pr: 3134,
     interactsWith: [ 'disableHolderImageWidget' ],
     detect: state => /"(enterRoutine|leaveRoutine|onEnter|onLeave|childrenPerOwner)"\s*:/.test(JSON.stringify(state)),
-    label: 'Legacy holder enter/leave events',
-    summary: 'leaveRoutine fires twice per move, onLeave only for drags and MOVE.',
+    label: 'Fire holder leave events twice',
+    summary: 'Holders raise the leave event twice per move, and apply onLeave only for drags and MOVE.',
     description: `
-      <b>Old behavior</b>: Two different places raised the leave event, so every drag and every <code>MOVE</code> called <code>leaveRoutine</code> twice - once when the widget was detached and once when it was really out of the holder. The <code>onLeave</code> properties were only applied on the second of those, which meant <code>SET parent</code> and <code>MOVEXY</code> never applied them at all. A card that joined a pile inside the holder it was already in raised a leave event with no matching enter, and a card dropped back into the holder it was picked up from ran <code>enterRoutine</code> without applying <code>onEnter</code>. A card taken out of a <code>childrenPerOwner</code> holder by <code>SET parent</code> or <code>MOVEXY</code> kept its owner.
+      <b>Old behavior</b>: A holder raises a <i>leave</i> when a widget stops being its child and an <i>enter</i> when one becomes its child. Two places raised the leave, so a drag or a <code>MOVE</code> called <code>leaveRoutine</code> twice, and <code>onLeave</code> was applied only on the second call.
       <br><br>
-      <b>New behavior</b>: A parent change raises at most one leave and one enter, whichever way the widget was moved. Each event applies its properties first and calls its routine afterwards, so the routine always sees the values the event wrote. Piles are transparent: joining, leaving or dissolving a pile inside another container is not a move between containers and raises no event. Picking a card up and dropping it back into the same holder is a real leave and a real enter, so both apply their properties. A <code>childrenPerOwner</code> holder releases the owner however the card left it, dragged or not - but a dragged card only once it is outside the holder's box, so rearranging a hand never uncovers a card.
-      <br><br>
+      <b>New behavior</b>: one leave and one enter per move, properties before routine. In detail:
+      <ul>
+        <li><code>leaveRoutine</code> runs once per departure, not twice</li>
+        <li><code>SET parent</code> and <code>MOVEXY</code> now apply <code>onLeave</code></li>
+        <li>dropping a card back into the holder it came from is a real leave and a real enter</li>
+        <li>a <code>childrenPerOwner</code> holder releases the owner however the card left it - a dragged card once it is outside the box, so rearranging a hand never uncovers it</li>
+        <li>joining or leaving a pile inside a holder raises nothing</li>
+      </ul>
       <b>Example:</b> a holder whose <code>leaveRoutine</code> counts the cards that left it.
       <br><br>
       Old result: dragging one card out counts <code>2</code><br>
       New result: dragging one card out counts <code>1</code>
-      <br><br>
-      Enable this mode for a game that was built around the double call - or around <code>MOVEXY</code> not applying <code>onLeave</code>.
       `
   }
 };
--- a/client/js/legacymoderegistry.js
+++ b/client/js/legacymoderegistry.js
@@ -120,13 +120,18 @@ export const LEGACY_MODES = {
     pr: 3134,
     interactsWith: [ 'disableHolderImageWidget' ],
     detect: function(state) {
-      if(/"(enterRoutine|leaveRoutine|onEnter|onLeave|childrenPerOwner)"\s*:/.test(JSON.stringify(state)))
+      const json = JSON.stringify(state);
+      if(/"(enterRoutine|leaveRoutine|onEnter|onLeave|childrenPerOwner)"\s*:/.test(json))
         return true;
-      // a stacked holder re-compacts on every departure now, which a game notices even when it
-      // uses none of the event names above
+      // A stacked holder re-compacts on every departure now, which a game notices even when it
+      // uses none of the event names above. A stack offset counts wherever it comes from: set
+      // by a routine while the game runs, or written on a widget - which does not have to be
+      // the holder itself, since it can be the template another widget inherits from or clones.
+      if(/"property"\s*:\s*"stackOffset[XY]"/.test(json))
+        return true;
       for(const id in state) {
         const properties = state[id];
-        if(!properties || properties.type != 'holder' || properties.alignChildren === false)
+        if(!properties || properties.alignChildren === false)
           continue;
         if([ properties.stackOffsetX, properties.stackOffsetY ].some(offset=>offset !== undefined && offset !== null && parseFloat(offset) != 0))
           return true;
@@ -146,6 +151,7 @@ export const LEGACY_MODES = {
         <li>a <code>childrenPerOwner</code> holder releases the owner however the card left it - a dragged card once it is outside the box, so rearranging a hand never uncovers it</li>
         <li>joining or leaving a pile inside a holder raises nothing</li>
         <li>a stacked holder closes the gap a card left however it left, <code>DELETE</code> included</li>
+        <li>the preview a <code>dropShadow</code> holder paints under the pointer raises nothing, so hovering over a holder no longer runs its routines</li>
       </ul>
       <b>Example:</b> a holder whose <code>leaveRoutine</code> counts the cards that left it.
       <br><br>
--- a/client/js/legacymoderegistry.js
+++ b/client/js/legacymoderegistry.js
@@ -119,7 +119,20 @@ export const LEGACY_MODES = {
     since: 23,
     pr: 3134,
     interactsWith: [ 'disableHolderImageWidget' ],
-    detect: state => /"(enterRoutine|leaveRoutine|onEnter|onLeave|childrenPerOwner)"\s*:/.test(JSON.stringify(state)),
+    detect: function(state) {
+      if(/"(enterRoutine|leaveRoutine|onEnter|onLeave|childrenPerOwner)"\s*:/.test(JSON.stringify(state)))
+        return true;
+      // a stacked holder re-compacts on every departure now, which a game notices even when it
+      // uses none of the event names above
+      for(const id in state) {
+        const properties = state[id];
+        if(!properties || properties.type != 'holder' || properties.alignChildren === false)
+          continue;
+        if([ properties.stackOffsetX, properties.stackOffsetY ].some(offset=>offset !== undefined && offset !== null && parseFloat(offset) != 0))
+          return true;
+      }
+      return false;
+    },
     label: 'Fire holder leave events twice',
     summary: 'Holders raise the leave event twice per move, and apply onLeave only for drags and MOVE.',
     description: `
@@ -132,6 +145,7 @@ export const LEGACY_MODES = {
         <li>dropping a card back into the holder it came from is a real leave and a real enter</li>
         <li>a <code>childrenPerOwner</code> holder releases the owner however the card left it - a dragged card once it is outside the box, so rearranging a hand never uncovers it</li>
         <li>joining or leaving a pile inside a holder raises nothing</li>
+        <li>a stacked holder closes the gap a card left however it left, <code>DELETE</code> included</li>
       </ul>
       <b>Example:</b> a holder whose <code>leaveRoutine</code> counts the cards that left it.
       <br><br>
--- a/client/js/legacymoderegistry.js
+++ b/client/js/legacymoderegistry.js
@@ -119,13 +119,13 @@ export const LEGACY_MODES = {
     since: 23,
     pr: 3134,
     interactsWith: [ 'disableHolderImageWidget' ],
-    detect: state => /"(enterRoutine|leaveRoutine|onEnter|onLeave)"\s*:/.test(JSON.stringify(state)),
+    detect: state => /"(enterRoutine|leaveRoutine|onEnter|onLeave|childrenPerOwner)"\s*:/.test(JSON.stringify(state)),
     label: 'Legacy holder enter/leave events',
     summary: 'leaveRoutine fires twice per move, onLeave only for drags and MOVE.',
     description: `
-      <b>Old behavior</b>: Two different places raised the leave event, so every drag and every <code>MOVE</code> called <code>leaveRoutine</code> twice - once when the widget was detached and once when it was really out of the holder. The <code>onLeave</code> properties were only applied on the second of those, which meant <code>SET parent</code> and <code>MOVEXY</code> never applied them at all. A card that joined a pile inside the holder it was already in raised a leave event with no matching enter, and a card dropped back into the holder it was picked up from ran <code>enterRoutine</code> without applying <code>onEnter</code>.
+      <b>Old behavior</b>: Two different places raised the leave event, so every drag and every <code>MOVE</code> called <code>leaveRoutine</code> twice - once when the widget was detached and once when it was really out of the holder. The <code>onLeave</code> properties were only applied on the second of those, which meant <code>SET parent</code> and <code>MOVEXY</code> never applied them at all. A card that joined a pile inside the holder it was already in raised a leave event with no matching enter, and a card dropped back into the holder it was picked up from ran <code>enterRoutine</code> without applying <code>onEnter</code>. A card taken out of a <code>childrenPerOwner</code> holder by <code>SET parent</code> or <code>MOVEXY</code> kept its owner.
       <br><br>
-      <b>New behavior</b>: A parent change raises at most one leave and one enter, whichever way the widget was moved. Each event applies its properties first and calls its routine afterwards, so the routine always sees the values the event wrote. Piles are transparent: joining, leaving or dissolving a pile inside a holder is not a move between containers and raises no event.
+      <b>New behavior</b>: A parent change raises at most one leave and one enter, whichever way the widget was moved. Each event applies its properties first and calls its routine afterwards, so the routine always sees the values the event wrote. Piles are transparent: joining, leaving or dissolving a pile inside another container is not a move between containers and raises no event. Picking a card up and dropping it back into the same holder is a real leave and a real enter, so both apply their properties. A <code>childrenPerOwner</code> holder releases the owner however the card left it, dragged or not - but a dragged card only once it is outside the holder's box, so rearranging a hand never uncovers a card.
       <br><br>
       <b>Example:</b> a holder whose <code>leaveRoutine</code> counts the cards that left it.
       <br><br>
--- a/client/js/widgets/holder.js
+++ b/client/js/widgets/holder.js
@@ -90,14 +90,9 @@ class Holder extends ImageWidget {
     return p;
   }
 
-  // childrenPerOwner makes the holder a per-player hand: what it receives becomes the receiving
-  // player's, and what leaves belongs to nobody again.
-  async applyEnter(child, oldParentID) {
-    if(child.get('type') != 'deck' && this.get('childrenPerOwner'))
-      await child.set('owner', child.targetPlayer||playerName);
-    await super.applyEnter(child, oldParentID);
-  }
-
+  // childrenPerOwner makes the holder a per-player hand, so what leaves it belongs to nobody
+  // again. The other half of that is in onChildAdd(), because becoming a child is what claims
+  // a widget, not the arrival event.
   async applyLeave(child) {
     if(child.get('type') != 'deck' && !child.isBeingRemoved && this.get('childrenPerOwner'))
       await child.set('owner', null);
@@ -158,16 +153,19 @@ class Holder extends ImageWidget {
     if(child.get('type') == 'deck')
       return;
 
-    // the legacy pipeline runs the enter half from here, which is why it applied onEnter to
-    // every parent change into the holder except one: a drop back into the holder the drag
-    // started in, which it recognised by the widget still remembering it as currentParent
-    if(legacyMode('legacyHolderEnterLeaveEvents')) {
-      if(this.get('childrenPerOwner'))
-        await child.set('owner', child.targetPlayer||playerName);
+    // A per-player hand claims every widget that becomes a direct child of it, which is not the
+    // same thing as an arrival: a hand that unpacks a dropped pile hands the cards over one by
+    // one, and none of those is a move between containers. The pile the cards came from is gone
+    // by the time the last one is out, so it is not claimed along with them.
+    if(this.get('childrenPerOwner') && !child.isBeingRemoved)
+      await child.set('owner', child.targetPlayer||playerName);
 
-      if(this != child.currentParent) // FIXME: this isn't exactly pretty
-        await this.applyEnterProperties(child);
-    }
+    // the legacy pipeline runs the property half of the arrival from here too, which is why it
+    // applied onEnter to every parent change into the holder except one: a drop back into the
+    // holder the drag started in, which it recognised by the widget still remembering it as
+    // currentParent
+    if(legacyMode('legacyHolderEnterLeaveEvents') && this != child.currentParent)
+      await this.applyEnterProperties(child);
   }
 
   async onChildAddAlign(child, oldParentID) {
--- a/client/js/widgets/holder.js
+++ b/client/js/widgets/holder.js
@@ -90,21 +90,57 @@ class Holder extends ImageWidget {
     return p;
   }
 
-  async dispenseCard(card) {
-    let toProcess = [ card ];
-    if(card.get('type') == 'pile')
-      toProcess = card.children();
+  // childrenPerOwner makes the holder a per-player hand: what it receives becomes the receiving
+  // player's, and what leaves belongs to nobody again.
+  async applyEnter(child, oldParentID) {
+    if(child.get('type') != 'deck' && this.get('childrenPerOwner'))
+      await child.set('owner', child.targetPlayer||playerName);
+    await super.applyEnter(child, oldParentID);
+  }
+
+  async applyLeave(child) {
+    if(child.get('type') != 'deck' && !child.isBeingRemoved && this.get('childrenPerOwner'))
+      await child.set('owner', null);
+    await super.applyLeave(child);
+  }
+
+  // onEnter and onLeave hold properties to apply to what entered or left. A pile is one widget
+  // for the event and a stack of cards for the properties, so they reach the cards it holds
+  // rather than the pile itself.
+  async applyEnterProperties(child) {
+    if(child.get('type') == 'deck')
+      return;
+
+    const toProcess = child.get('type') == 'pile' ? child.children() : [ child ];
+    for(const property in this.get('onEnter')) {
+      for(const w of toProcess) {
+        if(tracingEnabled)
+          sendTraceEvent('onEnter', { w: w.get('id'), child: child.get('id'), property, value: this.get('onEnter')[property], toProcess: toProcess.map(w=>w.get('id')) });
+        await w.set(property, this.get('onEnter')[property]);
+      }
+    }
+  }
+
+  async applyLeaveProperties(child) {
+    const toProcess = child.get('type') == 'pile' ? child.children() : [ child ];
     for(const w of toProcess) {
       if(!w.get('ignoreOnLeave')) {
         for(const property in this.get('onLeave')) {
           if(tracingEnabled)
-            sendTraceEvent('onLeave', { w: w.get('id'), child: card.get('id'), property, value: this.get('onLeave')[property], toProcess: toProcess.map(w=>w.get('id')) });
+            sendTraceEvent('onLeave', { w: w.get('id'), child: child.get('id'), property, value: this.get('onLeave')[property], toProcess: toProcess.map(w=>w.get('id')) });
           await w.set(property, this.get('onLeave')[property]);
         }
       }
     }
+    // a gap in a stacked holder closes as soon as the card that left it is gone
     if(this.get('alignChildren') && (this.get('stackOffsetX') || this.get('stackOffsetY')))
       await this.receiveCard(null);
+  }
+
+  // The leave half as the legacy pipeline ran it: from checkParent() once the widget really is
+  // outside the holder, and from a pile that handed a card back while sitting in one.
+  async dispenseCard(card) {
+    await this.applyLeaveProperties(card);
     if(Array.isArray(this.get('leaveRoutine')))
       await this.evaluateRoutine('leaveRoutine', {}, { child: [ card ] });
   }
@@ -122,20 +158,15 @@ class Holder extends ImageWidget {
     if(child.get('type') == 'deck')
       return;
 
-    if(this.get('childrenPerOwner'))
-      await child.set('owner', child.targetPlayer||playerName);
+    // the legacy pipeline runs the enter half from here, which is why it applied onEnter to
+    // every parent change into the holder except one: a drop back into the holder the drag
+    // started in, which it recognised by the widget still remembering it as currentParent
+    if(legacyMode('legacyHolderEnterLeaveEvents')) {
+      if(this.get('childrenPerOwner'))
+        await child.set('owner', child.targetPlayer||playerName);
 
-    if(this != child.currentParent) { // FIXME: this isn't exactly pretty
-      let toProcess = [ child ];
-      if(child.get('type') == 'pile')
-        toProcess = child.children();
-      for(const property in this.get('onEnter')) {
-        for(const w of toProcess) {
-          if(tracingEnabled)
-            sendTraceEvent('onEnter', { w: w.get('id'), child: child.get('id'), property, value: this.get('onEnter')[property], toProcess: toProcess.map(w=>w.get('id')) });
-          await w.set(property, this.get('onEnter')[property]);
-        }
-      }
+      if(this != child.currentParent) // FIXME: this isn't exactly pretty
+        await this.applyEnterProperties(child);
     }
   }
 
--- a/client/js/widgets/holder.js
+++ b/client/js/widgets/holder.js
@@ -139,6 +139,15 @@ class Holder extends ImageWidget {
     await this.closeStackGap();
   }
 
+  // A drop shadow is a preview and raises no leave, but it does take a slot in a stacked holder
+  // while it is there, and that slot has to close again when the pointer moves on. The legacy
+  // pipeline closes it from dispenseCard() instead, along with the rest of the shadow's leave.
+  async onChildRemove(child) {
+    await super.onChildRemove(child);
+    if(child.get('dropShadowOwner') && !legacyMode('legacyHolderEnterLeaveEvents'))
+      await this.closeStackGap();
+  }
+
   // a gap in a stacked holder closes as soon as the card that left it is gone
   async closeStackGap() {
     if(this.get('alignChildren') && (this.get('stackOffsetX') || this.get('stackOffsetY')))
--- a/client/js/widgets/holder.js
+++ b/client/js/widgets/holder.js
@@ -93,9 +93,18 @@ class Holder extends ImageWidget {
   // childrenPerOwner makes the holder a per-player hand, so what leaves it belongs to nobody
   // again. The other half of that is in onChildAdd(), because becoming a child is what claims
   // a widget, not the arrival event.
+  //
+  // A drag is the exception: it detaches the card at the pickup, long before it is anywhere
+  // else, and a card without an owner is one every other player can see. So a hand holds on to
+  // a dragged card until it is really out of its box, which is what checkParent() decides -
+  // rearranging a hand never exposes a card, because that drag never gets there.
   async applyLeave(child) {
-    if(child.get('type') != 'deck' && !child.isBeingRemoved && this.get('childrenPerOwner'))
+    if(child.get('type') != 'deck' && !child.isBeingRemoved && this.get('childrenPerOwner') && child.currentParent !== this)
       await child.set('owner', null);
+    // a widget on its way out of the room has no properties left to apply, but the gap it
+    // leaves in a stacked holder closes all the same
+    if(child.isBeingRemoved)
+      await this.closeStackGap();
     await super.applyLeave(child);
   }
 
@@ -127,13 +136,18 @@ class Holder extends ImageWidget {
         }
       }
     }
-    // a gap in a stacked holder closes as soon as the card that left it is gone
+    await this.closeStackGap();
+  }
+
+  // a gap in a stacked holder closes as soon as the card that left it is gone
+  async closeStackGap() {
     if(this.get('alignChildren') && (this.get('stackOffsetX') || this.get('stackOffsetY')))
       await this.receiveCard(null);
   }
 
   // The leave half as the legacy pipeline ran it: from checkParent() once the widget really is
-  // outside the holder, and from a pile that handed a card back while sitting in one.
+  // outside the holder, and from a pile that handed a card back while sitting in one. Nothing
+  // outside legacy mode calls this - applyLeave() is where a leave happens now.
   async dispenseCard(card) {
     await this.applyLeaveProperties(card);
     if(Array.isArray(this.get('leaveRoutine')))
--- a/client/js/widgets/holder.js
+++ b/client/js/widgets/holder.js
@@ -94,12 +94,13 @@ class Holder extends ImageWidget {
   // again. The other half of that is in onChildAdd(), because becoming a child is what claims
   // a widget, not the arrival event.
   //
-  // A drag is the exception: it detaches the card at the pickup, long before it is anywhere
-  // else, and a card without an owner is one every other player can see. So a hand holds on to
-  // a dragged card until it is really out of its box, which is what checkParent() decides -
-  // rearranging a hand never exposes a card, because that drag never gets there.
+  // Two exceptions. A drag detaches the card at the pickup, long before it is anywhere else,
+  // and a card without an owner is one every other player can see - so a hand holds on to a
+  // card that is still on the cursor until checkParent() decides it is really out of the box,
+  // which is why rearranging a hand never exposes a card. And MOVEXY has resetOwner, an
+  // explicit "put this on the table but keep it owned", which the leave must not override.
   async applyLeave(child) {
-    if(child.get('type') != 'deck' && !child.isBeingRemoved && this.get('childrenPerOwner') && child.currentParent !== this)
+    if(child.get('type') != 'deck' && !child.isBeingRemoved && this.get('childrenPerOwner') && !child.get('dragging') && !child.keepOwner)
       await child.set('owner', null);
     // a widget on its way out of the room has no properties left to apply, but the gap it
     // leaves in a stacked holder closes all the same
--- a/client/js/widgets/line.js
+++ b/client/js/widgets/line.js
@@ -673,7 +673,7 @@ export class Line extends Widget {
   // of the line, gets the onEnter properties applied and triggers enterRoutine
   // (which Widget.onPropertyChange does for every parent change).
   async onChildAdd(child, oldParentID) {
-    const entering = this != child.currentParent;
+    const enteringInLegacyMode = legacyMode('legacyHolderEnterLeaveEvents') && this != child.currentParent;
     await super.onChildAdd(child, oldParentID);
     // a rename re-adds the same stop under a new id; it carries no positioning
     // change, so skip the layout pass a real add/remove would trigger
@@ -685,7 +685,7 @@ export class Line extends Widget {
       await this.addStop(child.id, this.positionAtPoint(this.childCenter(child)));
     if(this.stopList().some(entry=>entry.widget == child.id))
       await this.layoutStops();
-    if(legacyMode('legacyHolderEnterLeaveEvents') && entering)
+    if(enteringInLegacyMode)
       await this.applyEnterLeave(child, 'onEnter');
   }
 
@@ -708,10 +708,10 @@ export class Line extends Widget {
   // Leaving the line again is the mirror of that: checkParent calls this hook -
   // named after the holder method it stands in for - once a widget is really
   // out, so the stop comes off the list and onLeave is applied. leaveRoutine is
-  // triggered by the parent change itself.
+  // triggered by the parent change itself. Nothing outside legacy mode calls
+  // this - applyLeaveProperties() is where a leave happens now.
   async dispenseCard(child) {
-    await this.removeStop(child.id);
-    await this.applyEnterLeave(child, 'onLeave');
+    await this.applyLeaveProperties(child);
   }
 
   // onEnter / onLeave hold properties to apply to the widget that entered or
--- a/client/js/widgets/line.js
+++ b/client/js/widgets/line.js
@@ -685,8 +685,19 @@ export class Line extends Widget {
       await this.addStop(child.id, this.positionAtPoint(this.childCenter(child)));
     if(this.stopList().some(entry=>entry.widget == child.id))
       await this.layoutStops();
-    if(entering)
+    if(legacyMode('legacyHolderEnterLeaveEvents') && entering)
       await this.applyEnterLeave(child, 'onEnter');
+  }
+
+  // The property half of the consolidated enter/leave pipeline - a line applies the same
+  // onEnter/onLeave a holder does, and drops the widget from its stop list on the way out.
+  async applyEnterProperties(child) {
+    await this.applyEnterLeave(child, 'onEnter');
+  }
+
+  async applyLeaveProperties(child) {
+    await this.removeStop(child.id);
+    await this.applyEnterLeave(child, 'onLeave');
   }
 
   // the middle of a child in the line's own coordinate frame
--- a/client/js/widgets/pile.js
+++ b/client/js/widgets/pile.js
@@ -267,7 +267,9 @@ export class Pile extends Widget {
       await removeWidgetLocal(this.get('id'));
     }
 
-    if(this.parent && this.parent.get('type') == 'holder')
+    // a card handed back by a pile inside a holder counts as leaving that holder, which the
+    // consolidated pipeline derives from the parent change itself (see enterLeaveContainer)
+    if(legacyMode('legacyHolderEnterLeaveEvents') && this.parent && this.parent.get('type') == 'holder')
       await this.parent.dispenseCard(child);
   }
 
--- a/client/js/widgets/widget.js
+++ b/client/js/widgets/widget.js
@@ -39,13 +39,14 @@ const readOnlyProperties = new Set([
   '_localOriginAbsoluteY'
 ]);
 
-// The widget whose enter and leave events a child of `parent` belongs to. A pile is transparent
-// here: it is a stack of cards inside whatever holds it, not a container a card can enter or
-// leave on its own, so a card that joins or leaves a pile inside its holder stays in that
-// holder. Returns null for a widget that sits on the table.
+// The widget whose enter and leave events a child of `parent` belongs to. A pile inside a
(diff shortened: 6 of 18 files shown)


VTT-AI-Agent:
> The test game should put its legacy mode to false so it isn't hidden in the ui and can easily be changed.

Done — all three variants of **Properties - Holder events** now carry

```json
"gameSettings": { "legacyModes": { "legacyHolderEnterLeaveEvents": false } }
```

in their `_meta`. The sidebar only renders a tile for a mode the game actually carries (`legacyMode(name) !== undefined`), so naming it explicitly as `false` is what makes it appear. Game Settings now shows *"0 of 1 legacy modes active for this game"* with an unchecked, toggleable **Legacy holder enter/leave events** tile, which is what each room's overview text tells the reader to flip.

The updated rooms are already in this PR's [`pr-test-ai` room](https://test.virtualtabletop.io/PR-3134/pr-test-ai) (they are not part of the diff — the public library stays untouched).

VTT-AI-Agent (commented):
## TL;DR

Solid, well-argued consolidation with unusually good test coverage — the model ("the parent write *is* the event, properties before routine") is the right one and the jest/TestCafe pairs make the change reviewable. One real regression though: because a drag now raises the leave at `moveStart()`, a card picked up inside a `childrenPerOwner` hand loses its `owner` for the whole drag, so **every other player sees that card while you reorder your own hand**. Second: the file-updater detector doesn't look at `childrenPerOwner`, so games that use a per-owner hand without any enter/leave routine or property get the new owner semantics with the legacy flag off.

## Findings

### 1. Reordering a card inside a per-owner hand exposes it to every other player (`client/js/widgets/holder.js:96`)

`moveStart()` writes `parent: null` (`client/js/widgets/widget.js:2876`), which is now a full leave: `Holder.applyLeave()` clears `owner` right there. That happens at *pickup*, before the drag has gone anywhere and regardless of whether the card ever leaves the holder's box. `owner: null` drops the `foreign` class in `Widget.classes()` (`client/js/widgets/widget.js:492-496`), and `body:not(.edit) .widget.foreign { display: none !important }` (`client/css/layout.css:892`) is the only thing that hides other players' hand cards — so for the whole duration of the drag the card (and its face) is rendered on every other client.

The legacy pipeline released the owner in `checkParent()`, i.e. only once the widget really was outside the holder, so a drag that stayed inside the hand never touched `owner`. Verified against this branch with the PR's own jest harness:

```
MODERN owner during an in-hand reorder drag: null    (after the drop: "jestPlayer")
LEGACY owner during an in-hand reorder drag: "jestPlayer"
```

(same room as the `childrenPerOwner` scenario at `tests/client/engine/enterleave.test.js:362`, but with `dragWidget('c1', { to: 'handA', leavesBox: false })`.)

Rearranging your hand is one of the most common interactions in a card game, so this is player-visible. It compounds with the other half of the same change: an in-holder reorder now also applies `onLeave` and then `onEnter` (your own `tests/client/engine/enterleave.test.js:153-168` records exactly that), so a hand with `onLeave: { activeFace: 0 }` will flip the card face-up *while it is visible to everyone*.

Suggestion: keep the `childrenPerOwner` release tied to actually leaving rather than to the `parent` write — e.g. skip it in `Holder.applyLeave()` while `child.currentParent === this` (the drag has not committed yet) and do it from `checkParent()` for that case, or have `moveStart()` mark the leave as provisional and settle it in `checkParent()`/`moveEnd()`. Whichever way, please add the missing case to both suites: none of the 65 jest scenarios or the TestCafe fixtures cover `childrenPerOwner` + a drag with `leavesBox: false`, which is why this slipped through.

### 2. The legacy-mode detector can't see `childrenPerOwner` games (`client/js/legacymoderegistry.js:122`)

```js
detect: state => /"(enterRoutine|leaveRoutine|onEnter|onLeave)"\s*:/.test(JSON.stringify(state)),
```

The owner half of the new pipeline is not gated by any of those four properties. A game with a per-owner hand and no enter/leave routine or property gets real behavior changes with the flag left off:

- `SET parent` and `MOVEXY` out of a per-owner hand now clear `owner`; they did not before (your own `tests/client/engine/enterleave.test.js:379-386` shows `owner=jestPlayer` in legacy vs `owner=null` now). A game that plays a card from a hand to the table with `MOVEXY` and then selects it with `SELECT owner` — or simply relies on it staying hidden — changes behavior on load with no opt-out flag set.
- finding 1 applies to it too.

`docs/compat.md` says to be deliberately over-inclusive here ("a false positive just keeps the old (working) behavior; a false negative breaks someone's game"), so adding `childrenPerOwner` to the alternation looks like the cheap fix.

### 3. The mode's description omits the two changes an author is most likely to hit (`client/js/legacymoderegistry.js:124-134`)

The Game Settings text covers the doubled `leaveRoutine` and the missing `onLeave` for `SET parent` / `MOVEXY`, and the PR body's "looks for three things" list matches it. Neither mentions:

- picking a card up and dropping it back into the *same* holder now applies `onLeave` and then `onEnter` (previously: neither), so an in-holder rearrange runs the property half of both events;
- `childrenPerOwner` now releases the owner at the parent write rather than when the widget leaves the holder's box, and for `SET parent` / `MOVEXY` as well.

Both are behavior an author has to check for, and the second is what makes finding 1 visible. Worth a line each in `description` and in the PR body's migration list.

### 4. `applyLeaveProperties()` bundles the stacked-holder re-compaction with the property half, so `DELETE` never closes the gap (`client/js/widgets/holder.js:130-132`, `client/js/widgets/widget.js:3213`)

`Widget.applyLeave()` skips `applyLeaveProperties()` entirely for `isBeingRemoved`, and the `receiveCard(null)` re-layout lives inside it. Deleting a card out of a `stackOffsetY` holder therefore leaves a hole — measured on this branch, a three-card stack at y=4/34/64 with the first card deleted ends up at y=34/64 rather than 4/34.

This is **pre-existing** (legacy never reached `dispenseCard()` on `DELETE` either), so it is not a regression — but the new structure is exactly where it becomes fixable, and the new fixture `tests/testcafe/enterleave.js:255` ("A stacked holder closes the gap when a card is taken out") reads as if it were covered when only the drag/`MOVEXY` paths are. Moving the `receiveCard(null)` call out of `applyLeaveProperties()` into `applyLeave()` (it is a layout consequence of losing a child, not a property applied to the departing widget) would make it run for removals too. Fine to leave for a follow-up, but the current split is a small trap.

### 5. Pile transparency is wider than the description says (`client/js/widgets/widget.js:47-51`)

`enterLeaveContainer()` makes a pile transparent whenever its parent exists in `widgets`, not only when that parent is a holder. So a pile sitting inside a *line*, a seat, or any plain container widget also loses its own `enterRoutine`/`leaveRoutine` for cards joining and leaving it — the PR body only claims this for "a pile inside a holder", and "a pile on the table has nothing behind it" is the only exception in the code. Games that put `enterRoutine`/`leaveRoutine` on piles via `onPileCreation` are caught by the detector, so old games are safe; it's the documented model that is narrower than the implementation. Either widen the comment/description or restrict the check to `parent.get('type') == 'holder'`.

### 6. Small things

- `Line.dispenseCard()` (`client/js/widgets/line.js:712-715`) is now byte-identical to `Line.applyLeaveProperties()` — let it delegate (`return this.applyLeaveProperties(child)`), so the legacy path can't drift from the modern one.
- `const entering = this != child.currentParent` in `Line.onChildAdd()` (`client/js/widgets/line.js:676`) is only read inside the legacy branch now; fold it into the `if`.
- `dispenseCard()` on both `Holder` and `Line` is dead outside legacy mode after this PR. Worth a one-line comment saying so (the stale "which lets the line dispense it" comment at `client/js/widgets/widget.js:3044` now describes only the legacy path).

## What looks good

- `applyParentChange()` guards `oldValue`/`newValue` with `widgets.has()`, where the old code did a bare `widgets.get(oldValue).onChildRemove(...)` and would throw on a dangling parent id.
- The legacy path is a faithful copy of the old code rather than a re-derivation, and every jest scenario asserts both combinations — including the "unchanged on purpose" ones, which turns "this didn't change" into a real assertion.
- `tests/client/engine/bundle-widgets.js` is a genuinely useful piece of infrastructure: getting the real `Holder`/`Pile` into jsdom is what makes 65 assertions affordable here instead of 65 browser fixtures.
- No `library/` changes, and the three demo rooms live on this PR's `pr-test-ai` server as required. The new legacy mode needs no `jeCommands`/validator entry (it isn't a widget or operation property) and the Game Settings sidebar picks it up from the registry, so the UI-support rule is satisfied.
- `npm test` is green on this branch here as well (1553 tests).

I did not add the `beta` label: this PR rewrites the core parent-change/enter-leave pipeline in `client/js/widgets/widget.js`, `holder.js`, `pile.js` and `line.js` and changes how existing games run, which is engine work and therefore excluded from beta auto-deployment by design.

VTT-AI-Agent:
Thanks — finding 1 was a real regression and it is fixed, along with 2, 3, 4 and the three small things. Pushed as b0a5349 and e77a8b7.

### 1. Reordering a card inside a per-owner hand exposed it — fixed

Confirmed and fixed. `Holder.applyLeave()` now skips the `childrenPerOwner` release while `child.currentParent === this` — i.e. while a drag still remembers this holder as the one it was picked up in — and `checkParent()` does it once the card is really outside the box, which is where it lived before this PR. Everything else about the leave still happens at the parent change.

The consequence you predicted for the trace is real and is now written down: a `leaveRoutine` on a per-owner hand reads the owner *still set* for a drag and for `MOVE`/`moveToHolder` (both detach through `checkParent()`), and reads it already cleared for `SET parent`/`MOVEXY`. That is the one place where the "properties before routine" rule bends, so the model line in the description now spells it out: a hand lets go of a *dragged* card only when the card is outside it; every other way of taking one out releases it with the rest of the leave.

Three new tests, at the three levels the claim lives at:

- jest, both combinations: `a card rearranged inside a per-owner hand stays owned for the whole drag` and `a card dragged out of a per-owner hand is released when it leaves the box` — `dragWidget()` gained a `whileDragging` hook so the owner can be read at the moment the card hangs on the cursor.
- TestCafe: `Rearranging a card inside a per-owner hand never takes its owner away` (the leave and the enter both read `owner=Alice`), and the existing per-owner drag-out test renamed to `… releases it once it is out of the box`.
- TestCafe multiclient: `Rearranging a per-owner hand keeps the card hidden from the other player` — `dragPath()` gained `hold: true`, so the fixture stops mid-gesture, switches to Bob's window and asserts `foreign` is still on the card, then releases. This is the assertion the bug was actually about, and it fails on the pre-fix code with exactly your symptom:

  ```
  AssertionError: Bob does not see the card Alice is moving inside her hand: expected false to deeply equal true
  ```

### 2. The detector can't see `childrenPerOwner` games — fixed

`childrenPerOwner` added to the alternation, and the PR body now says why the detector is wider than the four routine/property names.

### 3. The description omits two author-visible changes — fixed

Both added to the mode's `description` and to the migration list in the PR body: the in-holder rearrange now applying `onLeave` → `onEnter`, and the owner release reaching `SET parent`/`MOVEXY`. The migration list is five bullets now.

### 4. `DELETE` never closed a stacked holder's gap — fixed

You are right that the new structure is where it becomes fixable, so I did rather than leave the trap. The re-compaction moved into `Holder.closeStackGap()`, called from `applyLeaveProperties()` as before *and* from `applyLeave()` for `child.isBeingRemoved`, where there are no properties to apply but the holder is still one card shorter. Legacy mode keeps the old behavior (it never reaches `applyLeave()`). New fixture `DELETE closes the gap in a stacked holder as well`, next to the drag and `MOVEXY` ones — so the section no longer over-claims.

### 5. Pile transparency is wider than the description says — documentation widened

Deliberate: what is behind a pile is the container the cards are in, whatever its type. A card joining a pile that sits in a line has not left the line's area, so the line should not see a departure — restricting the check to holders would make that case wrong rather than narrower. The comment on `enterLeaveContainer()`, the mode description and the PR body now all say "inside whatever holds it" instead of "inside a holder".

### 6. Small things — all three done

- `Line.dispenseCard()` delegates to `applyLeaveProperties()`.
- `entering` folded into the legacy condition as `enteringInLegacyMode`.
- `dispenseCard()` on `Holder` and `Line` carries a line saying nothing outside legacy mode calls it, and the stale "which lets the line dispense it" comment in `applyLineStopDrop()` now describes the leave event instead.

### Verification

`npm test` green (1557, up from 1553 — the four new jest cases). TestCafe on this branch, all green: `enterleave.js` (25), `holderevents.js` (22), `multiclient.js` (6, with `--disable-native-automation`), `interactions.js` (32), `legacymatrix.js` (9, including `only-legacyHolderEnterLeaveEvents` and the `disableHolderImageWidget` pair).

The **Piles & hands** demo room on the test server got a line about the new timing and is re-uploaded to [`pr-test-ai`](https://test.virtualtabletop.io/PR-3134/pr-test-ai).

On the `beta` label: agreed, and I have not added it.

VTT-AI-Agent:
Feedback from Discord on the manual test rooms:

> The test game should have human readable cards that act like cards. You have 3 cards with text on a transparent background. Stack those on top of each other and it is impossible to see them well. Make the background white on face 1.

— [lawdawg96, #🔵 PR 3134](https://discord.com/channels/770758631146782780/1539994146897399808/1540095842382258176)

Done — the deck in all three variants of **Properties - Holder events** now has an opaque, card-shaped face instead of bare text on a transparent rectangle.

| before | after |
| --- | --- |
| three stacked cards, their ids printed over each other | the same stack, one legible card on top |

Face 1 is off-white with a black edge and a hairline inset, and it is laid out the way a playing card is:

- the card's **id in the top-left corner** and mirrored upside-down in the bottom-right, so a stacked or fanned hand can be read card by card — which is exactly what the log lines (`child=card2`) refer to;
- a large **suit pip** in the middle, red or black, so the cards are told apart at a glance;
- a small line under the pip showing the card's current **`mark`** — the property `onEnter`/`onLeave` writes. The card now says `enter-handB` on its own face at the same moment the log line appears, so the demo can be followed without reading the log at all.

a hand of three cards fanned out with stackOffsetY, each id legible

Face 2 is a proper card back (diagonal weave with a light inset border) rather than a flat grey fill.

The rooms' library preview image was regenerated to match, all three variants pass `validator/validate_gamefile_node.js` with no warnings, and the updated file is live in this PR's test-server room: <https://test.virtualtabletop.io/PR-3134/pr-test-ai>. No repository files changed — the demo rooms stay out of the diff.

VTT-AI-Agent (commented):
## TL;DR

**UI/UX review only** — layout, visual design and first-time readability. Correctness is another reviewer's job; I flag nothing about the engine logic here.

The interface this PR ships is essentially **one new Game Settings tile** (`legacyHolderEnterLeaveEvents`), and it renders correctly at every viewport I tried. The single most impactful improvement: **its description is 285 words — 3× longer than the longest of the four existing legacy-mode tiles (47–92 words) — and the migration detector fires on 328 of the 401 public-library game files (82%)**, so this is the legacy tile most authors will ever open. Break the two behaviour paragraphs into the five-bullet list your own PR body already has, and shorten the label to the verb-phrase form its four siblings use.

The `pr-test-ai` room also holds a three-variant draft tutorial; it is reviewed in its own section at the bottom, and it has the more actionable layout problems of the two surfaces.

**What I exercised.** Local server on the PR branch, Chromium via Playwright, at **390×844, 820×1180, 1280×800 and 1920×1080**: the board, the Game Shelf / library browser, Players, About, the Options menu, Edit Mode, all eight editor sidebar modules (Edit Widgets, History, JSON, Widgets, Debug, Assets, Toolbox, Game Settings), the game-details + variants overlay, and all three tutorial variants. I could **not** reach the Add Widget overlay, the zoom popup, the save/share/publish dialogs or the chat panel in automation. Nothing outside the Legacy Modes section changed appearance — expected, since the diff touches no CSS or HTML.

---

## Big layout improvements

### 1. The new tile's description is far longer than any of its neighbours, and 82% of games will see it

`client/js/legacymoderegistry.js:117-137`. Word counts of the rendered description text:

| mode | words | longest sentence |
|---|---|---|
| `convertNumericVarParametersToNumbers` | 92 | 38 |
| `useOneAsDefaultForVarParameters` | 50 | 21 |
| `useIframeForHtmlCards` | 47 | 21 |
| `disableHolderImageWidget` | 65 | 25 |
| **`legacyHolderEnterLeaveEvents`** | **285** | **40** |

The registry's own header comment asks for "the vocabulary of the existing entries … so the panel reads as one document" — and the panel currently reads as four index cards plus an essay:

tile open, 1280×800

This matters more than usual because of the detector's reach. I ran the `detect` regex over the shipped library:

```
library/games/*.json: 401 files, 328 match (81.8%)
```

That is the intended over-inclusiveness per `docs/compat.md`, no argument — but it means four out of five authors who open Game Settings will meet *this* tile, checked, with a "1 of N legacy modes active" banner above it.

**Concrete proposal:** keep "Old behavior" to one sentence, then use a `<ul>` for the specifics. Your PR body already contains exactly the right list ("A game author who wants the new pipeline unchecks it and looks for five things") and it is dramatically more scannable than the 90-word "New behavior" sentence-chain that shipped. Something like:

```html
<b>Old behavior</b>: Two places raised the leave event, so a drag or a <code>MOVE</code>
called <code>leaveRoutine</code> twice and <code>onLeave</code> was applied only on the
second call.
<br><br>
<b>New behavior</b>: one leave and one enter per move, properties before routine. In detail:
<ul>
  <li><code>leaveRoutine</code> runs once per departure, not twice</li>
  <li><code>SET parent</code> and <code>MOVEXY</code> now apply <code>onLeave</code></li>
  <li>dropping a card back into the holder it came from is a real leave and enter</li>
  <li>a <code>childrenPerOwner</code> holder releases the owner however the card left</li>
  <li>joining or leaving a pile inside a holder raises nothing</li>
</ul>
```

The Example block at the end is good and matches the sibling style — keep it as is.

### 2. On a phone the opened tile is taller than the whole screen

At 390×844 the expanded tile does not fit in one viewport — the reader scrolls a full screen-height of prose inside a single settings row before reaching the next tile:

tile open, 390×844

The bulleted rewrite above roughly halves this. No other tile has this problem.

### 3. Label doesn't follow the section's naming convention

Every existing label is an imperative verb phrase, and none of them repeats the word "Legacy" (the section header already says *Legacy Modes*):

- `Convert numeric var parameters to numbers`
- `Use 1 as default for var parameters`
- `Use iframes for card face HTML objects`
- `Disable holder image support`
- **`Legacy holder enter/leave events`** ← noun phrase, and the only one containing "Legacy"

Suggest **`Fire holder leave events twice per move`** or **`Raise holder enter/leave events the old way`**. As a bonus this fixes the wrap noted below.

---

- **`client/js/legacymoderegistry.js:123` — label wraps at phone width.** "Legacy holder enter/leave events" is the only one of the five labels that wraps to two lines at 390×844; the four siblings all stay on one, so the tile header is a line taller than its neighbours. Side by side with `Disable holder image support`: two tiles, 390×844
- **`client/js/legacymoderegistry.js:124` — the summary line is rendered with `textContent`** (`gameSettings.js:65`, `summary.textContent = mode.summary`), so `<code>` is not available there. `leaveRoutine fires twice per move, onLeave only for drags and MOVE.` therefore renders as plain prose starting with a lowercase identifier, and the second clause has no verb — the four sibling summaries are all ordinary sentences with normal capitalisation. Suggest: `Holders raise the leave event twice per move, and apply onLeave only for drags and MOVE.`
- **`interactsWith` pairing renders fine** — with both holder modes defined, the two tiles stack cleanly and the "N of M legacy modes active" counter is correct at all four viewports. No issue, just confirming I checked it.
- Verified at 1920×1080 in the unchecked state, including the `REMOVE` affordance and the Details disclosure: Game Settings, 1920×1080

### Two in-room visual changes worth noting (both good)

**A stacked holder now re-compacts after `DELETE`.** Same room, same click, mode off vs on — with the legacy mode the remaining cards stay piled on one slot, with the new pipeline the row re-lays out:

| new pipeline | `legacyHolderEnterLeaveEvents` on |
|---|---|
| gap closed | gap not closed |

**Rearranging a per-owner hand still doesn't uncover a card.** Two browser contexts, Alice mid-drag inside her own `childrenPerOwner` hand; Bob sees only her cursor, in both modes — no flicker regression:

| Alice, mid-drag | Bob, same moment |
|---|---|
| Alice | Bob |

---

- **The summary reads as a statement of fact even when the box is unchecked.** "leaveRoutine fires twice per move" is present tense, so on an *inactive* tile it describes behaviour the game does not have. The sibling summaries share the pattern, but this is the one that names an observable count, so it's the one most likely to be read as "my game does this". Prefixing the sentence — or keeping the count out of the summary and leaving it to the Example — would remove the ambiguity.
- **The description opens on jargon.** "Two different places raised the leave event" assumes the reader already knows what a leave event is. The audience is game authors, so this is defensible — but one clause up front ("a holder raises a *leave* when a widget stops being its child") would let a newcomer read the rest.
- `SET parent`, `MOVEXY`, `childrenPerOwner`, `ignoreOnLeave`, "property bag" and "re-compacts" all appear without gloss. Fine for the wiki-reading author; worth one link to the Legacy-Mode wiki page you already append at the bottom.
- Everything else in the tile is clear: the `Active` chip, the accent border and the Example/Old result/New result block all read correctly at every viewport.

---

## Draft tutorial

The `pr-test-ai` room holds a draft: **"Properties - Holder events"**, mode `Tutorial`, three variants — **Dragging**, **Routines**, **Piles and hands**. I played all three end to end at 1280×800 and 1920×1080, and checked them at 820×1180 and 390×844 (phone portrait shows the app's own "rotate your device" prompt, so I did not judge layout there).

`_meta.info` parity against the canonical block in `docs/tutorials.md` is **good** — `mode: "Tutorial"`, `players: "1"`, `time: "0"`, `language: "en-US"`, `showName: false`, `variant` set per file, nothing invented or omitted. The preview is still the draft screenshot, which is what the draft rule expects; the square-SVG badge is the step for after an Accept.

It also *works*: dragging a card from handA to handB produced exactly the two-line trace the intro promises.

Dragging, 1280×800

### Big

**1. The event log — the whole point of all three rooms — is an unlabelled grey slab that is ~95% empty.** `log` is 1440×360 in "Dragging" and "Piles and hands" and 1440×220 in "Routines", it carries no caption, and after a normal interaction it holds two lines at the top of it:

after a drag

The intro says "watch the log", but nothing on the board says which grey rectangle that is. Three fixes, all cheap: add a caption widget above it the way every other example here has one ("Event log — one line per enter/leave"), cut the height to ~180–200 so it stops dominating the lower half of the board, and give `text` a placeholder so it isn't a blank slab on load (`(no events yet — drag a card out of handA)`).

**2. "Routines": ten buttons in one column overflow into the log, while the left half of the board below the caption is empty.** `button1..button10` run x810 y250 → y762 in a single 480-wide column; `button10` (y718 + h44 = **762**) crosses the log's top edge (**y760**). Meanwhile x80–740 / y580–980 is dead space.

Routines, 1280×800

Splitting them into two columns of five — exactly the 2-column grid "Dragging" already uses for its toggles — removes the collision, fills the gap, and makes the three variants look like one set. While you're there: `clearLog` sits alone at x1330 y540, detached from the group it belongs to; in "Dragging" it lives inside the button grid.

### Nitpicks

- **`text3` in "Piles and hands" overflows its box and renders on top of the log.** x800 y460 w300 **h130**, but its content needs 216px (measured `scrollHeight` 216 vs `clientHeight` 130) — the last two lines ("Click the seat to sit down before you use the seat button.") land on the grey slab. Visible at both 1280×800 and 820×1180: Piles, 1280×800 Piles, 820×1180 Raise `height` to ~220, or move the seat sentence next to the seat where it applies.
- **`intro` in "Dragging" overflows too** — `scrollHeight` 140 vs `height` 130. `docs/tutorials.md` specifies `height: 150` for `overview`; using it fixes this.
- **The three cards overprint into illegible text.** `card1`, `card2` and `card3` all sit at `x:4, y:4` in handA (and in handB in "Piles and hands") with `stackOffsetY: 0`, so their labels render on top of each other as a smudge at every viewport — it reads as one card, not three, which undercuts "Drag *a* card out of handA". A small `stackOffsetX` (~30) on handA would fan them out and show the reader there is more than one to drag.
- **The log's width changes between variants** — 1440 in "Dragging"/"Routines" but 1040 in "Piles and hands", so the bottom band jumps when you switch variant.
- **"Recall & Shuffle" wraps to two lines** inside its 111×40 pill in all three variants; it's the documented size, so either shorten the label or widen the button in this room.
- **In "Piles and hands" the seat and the deck holder are two identical plain white rounded rectangles side by side** — only one of them says "click to sit", so a newcomer can't tell what the other one is. A `text` on `deckHolder` ("deck") would do it.
- **Button widths differ per variant** (240 / 330 / 480), so identical-looking pills render their labels at three noticeably different sizes across the three rooms.
- **Minor deviations from the structural-widget spec** in `docs/tutorials.md`: `title` is x200 w1200 (doc: x400 w800), `intro` is x60 y62 w1480 h130 `layer:-1` (doc: `overview` at x108 y80 w1400 h150 `layer:-3`). Harmless visually, but the doc asks for parity with the newest tutorials.

### Readability

- **`mark` is never explained.** Captions say things like "so **onLeave** writes *mark* before **leaveRoutine** reads it" — a first-time reader will hunt for a `mark` property in the wiki. One clause fixes it: "…writes *mark* (an arbitrary property this room sets through `onLeave`)…".
- **Pointing the reader at Game settings → *Legacy holder enter/leave events* to see the old pipeline is genuinely excellent** — I followed that instruction end to end and it does what the intro promises. Worth keeping in all three variants (it is).
- The captions are otherwise in the right register: second person, one or two sentences, property names in `<b>` and values in `<i>`, in visual order.
- The variants list in the library details shows all three with the same preview thumbnail, so they are only distinguishable by name — fine per the metadata convention, just noting it since the names ("Dragging" / "Routines" / "Piles and hands") are doing all the work: variants

VTT-AI-Agent:
Thanks — this was a useful pass. Everything actionable is fixed; details below, in the order you raised them. Reply to [pullrequestreview-4987320444](https://github.com/ArnoldSmith86/virtualtabletop/pull/3134#pullrequestreview-4987320444).

## The Game Settings tile (`dfe6558d`)

**Description length / phone height.** Rewritten around your five-bullet proposal: one sentence of "Old behavior", one sentence of "New behavior", then the five specifics as a `<ul>`, then the Example block. 285 → 135 words. Measured in Chromium at 390×844 the opened tile is now **701px tall against an 844px viewport**, so it fits on one phone screen; it was 890px before. A `.settings-tile ul` rule in `sidebarModules.css` pulls the list back into the 21px text column the rest of the tile uses, since the browser default indents it much further.

- [tile at 1280×800](https://agent.virtualtabletop.io/reports/pr/1539994146897399808/holder-events-legacy-tile.png)
- [tile at 390×844](https://agent.virtualtabletop.io/reports/pr/1539994146897399808/holder-events-legacy-tile-phone.png)

**Label.** Now **`Fire holder leave events twice`** — imperative verb phrase, no "Legacy", and it describes what checking the box makes the engine do, like `Disable holder image support` does.

One correction on the wrap measurement: I measured the rendered `<label>` at 390×844 and the label column is ~280px wide there. The old `Legacy holder enter/leave events` fitted on one line (241px); the one that wraps is `Convert numeric var parameters to numbers` (280px, 2 lines). The new label is 215px, so it stays on one line whether or not the `Active` chip is showing — which was the point either way.

**Summary line.** Now `Holders raise the leave event twice per move, and apply onLeave only for drags and MOVE.` — an ordinary sentence, as you suggested, since `textContent` gives it no `<code>`.

**Jargon.** The description now opens by saying what the events are ("A holder raises a *leave* when a widget stops being its child and an *enter* when one becomes its child") before it uses the words. The wiki link the sidebar appends to every tile covers the rest.

**82% detector reach.** Intentional and staying — `docs/compat.md` asks for over-inclusion, and `childrenPerOwner` had to go into the alternation because the owner half of the change is not gated by the four routine/property names. Your point was that this makes the tile the one authors actually read, which is exactly why the rewrite was worth doing.

## Draft rooms — *Properties - Holder events*

All three variants rebuilt and replayed end to end; they are back in the [`pr-test-ai` room](https://test.virtualtabletop.io/PR-3134/pr-test-ai), still not part of the diff.

| | |
|---|---|
| [Dragging](https://agent.virtualtabletop.io/reports/pr/1539994146897399808/holder-events-dragging.png) | [Routines](https://agent.virtualtabletop.io/reports/pr/1539994146897399808/holder-events-routines.png) |
| [Piles & hands](https://agent.virtualtabletop.io/reports/pr/1539994146897399808/holder-events-piles-and-hands.png) | |

**The log.** All three fixes. It has a caption above it (`Event log - one line per enter or leave, in the order they happened.`), it is 220 high instead of 360 (~8 lines, enough for the longest trace any button produces), it is 1440 wide at the same `y` in every variant so the band no longer jumps, and it carries a placeholder — `(no events yet - drag a card out of handA)` and the equivalent per room — which the first event replaces and *Clear the log* restores. The placeholder needed a single place to decide "first line or append", so the six holder routines now hand their line to one `appendRoutine` on the `log` widget through `CALL`; that also took a copy of the same six-op block out of each holder.

**"Routines" button column.** Two columns of five at 240×44, the same grid and the same pill size the other two variants use, ending at y502 — nothing reaches the log band any more. *Clear the log* sits under the grid instead of alone at x1330, and the `MOVEXY` landing spot moved out of the log's band.

**`text3` overflow in "Piles and hands".** Fixed — the seat sentence moved to its own caption next to the seat it applies to (`text4`), and the caption row is 170 high.

**`intro` overflow.** Fixed by adopting the documented structural geometry you also flagged: `intro` is now `overview` at `x:108,y:80,width:1400,height:150,layer:-3` and `title` is `x:400,width:800`.

**Cards overprinting.** The three cards are now authored at `x: 4 / 68 / 132`, so a hand reads as a hand. Two notes: the card faces are opaque, so what you saw was one card covering two rather than a smudge; and in "Piles and hands" the fan needed `alignChildren: false` plus `onPileCreation.pileSnapRange` on the cards, because with `alignChildren` on, every drop snaps back to `dropOffset` and the pile demo would only have worked from the exact stacked position. `stackOffsetX` was not an option — it turns `supportsPiles()` off, which would have silently killed the `preventPiles` toggle.

**"Recall & Shuffle" wrap.** Kept the documented 111×40 button and the conventional label — every tutorial shares them — and gave it `font-size: 14px` in this room so it renders on one line.

**Deck holder vs seat.** `deckHolder` now says `deck`.

**Button widths.** 240 in all three variants; the labels that were too long for it were shortened (`MOVE a card to seat 1`, `MOVEXY out of handA`, …).

**`mark` gloss.** Added: "so **onLeave** writes *mark* — an ordinary property this room uses as a marker — before **leaveRoutine** reads it".

**Two captions were also wrong**, which the replay turned up: "Drop a card onto another card here. A pile forms and the log stays quiet" is not what happens — the drag's own leave of `handA` and re-enter of `handA` are logged, and it is the *joining* that is silent. Both that caption and the `preventPiles` one now say what the log actually shows. The `preventPiles` claim that unpacking a pile is still one arrival did check out: dragging a two-card pile into `handB` logs exactly one `ENTER handB` naming the pile.

The preview image is a fresh screenshot of the reworked "Dragging" board, mid-trace.

`npm test` green (1557). Nothing in the diff touches engine behaviour — it is the registry text plus one CSS rule.

VTT-AI-Agent (commented):
## TL;DR

The core event consolidation and the fixes from the first code review are present, but two correctness gaps remain. Most importantly, temporary drop-shadow clones still fire author routines and property hooks merely while a card hovers over a holder; the PR's own test currently locks in that unwanted behavior.

## Findings

### 1. Drop-shadow previews still execute real enter/leave behavior (`client/js/widgets/widget.js:3174`)

The consolidated path does not distinguish a real widget from the temporary clone marked by `dropShadowOwner`. During a drag, that clone is parented into the hovered holder (`widget.js:2974-2979`) and later parented back to `null` (`widget.js:3121`), so `applyParentChange()` calls `applyEnter()` and `applyLeave()` for it. That applies `onEnter`/`onLeave` and runs `enterRoutine`/`leaveRoutine` even if the player only hovers over the holder and never drops the real card.

This is not just theoretical: `tests/testcafe/holderevents.js:137-149` explicitly expects the shadow's synthetic `enter`, its synthetic `leave`, and then the real card's `enter`. A holder routine may score, deal, delete, or move widgets, so previewing a drop can mutate the game twice before the actual drop; it also contradicts the PR's "at most one leave and one enter" model and the open PR TODO that drop-shadow widgets should trigger nothing.

Please suppress the event halves for widgets with `dropShadowOwner` while retaining the child add/remove work needed to lay out the preview. The browser test should then expect only the real card's final enter, with an additional cancelled-hover case asserting that neither routines nor `onEnter`/`onLeave` produce state changes.

### 2. Plain stacked holders do not receive the legacy flag for the new re-compaction semantics (`client/js/legacymoderegistry.js:122`)

The new pipeline now calls `closeStackGap()` for every modern leave (`client/js/widgets/holder.js:128-145`) and explicitly for removal at `holder.js:104-108`. Consequently, `SET parent`, `MOVEXY`, and `DELETE` now re-position the remaining children of a stacked holder where old code did not; `tests/testcafe/enterleave.js:270-301` demonstrates the latter two changes.

However, the updater detector only looks for `enterRoutine`, `leaveRoutine`, `onEnter`, `onLeave`, or `childrenPerOwner`. A pre-v23 game with an otherwise plain `{ "type": "holder", "stackOffsetY": 40 }` matches none of those, so it is upgraded without `legacyHolderEnterLeaveEvents: true` and silently receives the new layout behavior. That is a false negative under `docs/compat.md`, even though the implementation preserves the old behavior when the mode is actually enabled.

Please either include stacked holders (nonzero `stackOffsetX`/`stackOffsetY`, with effective `alignChildren`) in the detector or leave the additional re-compaction change for a separately migrated change. If retained here, add an updater test for a bare stacked holder, cover the legacy layout in the browser fixture, and mention this sixth migration difference in the Game Settings description.

## Standing checks

No new widget property or routine operation is introduced, so no new property-editor or visual routine-block control is required; the new compatibility switch is exposed through Game Settings. The three tutorial-quality demo variants are on the PR's `pr-test-ai` room, no `library/` path is changed, and the current CI checks are green.

I did not add the `beta` label because the changes to `client/js/widgets/widget.js`, `holder.js`, `line.js`, and `pile.js` genuinely alter default engine event and layout behavior, making this PR engine-ineligible for beta auto-deployment.

VTT-AI-Agent:
Both findings were valid and both are fixed in `4a375f5a`. Reply to [pullrequestreview-4988127318](https://github.com/ArnoldSmith86/virtualtabletop/pull/3134#pullrequestreview-4988127318).

## 1. Drop-shadow previews no longer execute enter/leave behavior

Fixed as you suggested, and it also clears the open TODO that was at the top of the description.

`Widget.applyParentChange()` now derives `raisesEvents` from `leaveTarget !== enterTarget && !this.get('dropShadowOwner')`, so the clone still goes through `onChildRemove()`/`onChildAdd()` — the preview needs that to be laid out — but neither `applyLeave()` nor `applyEnter()` runs for it. No routine, no `onEnter`, no `onLeave`, so hovering over a holder cannot score, deal, delete or move anything any more.

One consequence worth naming, because dropping the leave half dropped a piece of layout with it: while the shadow sits in a stacked holder it takes a slot, and the old pipeline closed that slot again from `dispenseCard()`. `Holder.onChildRemove()` now closes it for a `dropShadowOwner` child instead, so the cards that were there go back up as soon as the pointer moves on. I checked that this is load-bearing by stubbing the hook out — the new browser case goes red without it.

Tests, as you asked:

- `tests/testcafe/holderevents.js` — the case that pinned the unwanted behavior now expects only the real card's `enter handB[parent=handB]`, and a second case drags onto the holder and then away again, asserting an empty trace, no `onEnter` mark on the card and the card not in the holder.
- `tests/testcafe/enterleave.js` — *A stacked holder with dropShadow closes the preview gap again when the pointer moves on*.
- `tests/client/engine/enterleave.test.js` — *a drop shadow passes through a holder without raising anything*, run in both combinations: empty in modern, and the legacy trace unchanged.

The legacy pipeline deliberately keeps firing them: `legacyHolderEnterLeaveEvents` is a byte-for-byte restoration of the old engine, and *Legacy: a dropShadow holder runs a doubled leave for the shadow widget too* is one of the traces that proves it. The description says this is a plain bug fix rather than one of the mode's documented differences.

## 2. The detector now catches plain stacked holders

Also valid — a bare `{ "type": "holder", "stackOffsetY": 40 }` matched none of the five names and would have picked up the re-compaction silently. The detector keeps the regex and adds a structural pass: any holder with a non-zero `stackOffsetX`/`stackOffsetY` and `alignChildren` not switched off turns the mode on. An offset written as an expression counts as stacked (`parseFloat` gives `NaN`, which is not `0`), which is the over-inclusive direction `docs/compat.md` asks for.

The extra coverage:

- `tests/server/fileupdater.test.js` — a v22 save with a bare stacked holder gets exactly `{ legacyHolderEnterLeaveEvents: true }`; an offset written as `${PROPERTY offset OF board}` counts; `stackOffsetY: 0` and `alignChildren: false` do not.
- `tests/testcafe/enterleave.js` — *Legacy: DELETE leaves the hole in a stacked holder open* and *Legacy: SET parent leaves the hole in a stacked holder open as well*, both deleting the top card of a three-card column and asserting the two below it stay at y 44 and 84. The three modern counterparts assert 4 and 44, and all five now share one `stackedHolder()` fixture.
- The Game Settings tile has a sixth bullet — *a stacked holder closes the gap a card left however it left, `DELETE` included* — and the description above lists it as the fifth thing an author unchecking the mode should look for. The tile is still 735px tall at 390×844, so it still fits one phone screen: [1280×800](https://agent.virtualtabletop.io/reports/pr/1539994146897399808/holder-events-legacy-tile.png), [390×844](https://agent.virtualtabletop.io/reports/pr/1539994146897399808/holder-events-legacy-tile-phone.png).

## Verification

`npm test` green (1563). TestCafe green on `holderevents.js` (23), `enterleave.js` (28), `legacymatrix.js` (9) and `interactions.js` (32) against a local server.

Agreed on `beta` — the diff changes default engine behavior, so that call is not the author's to make either way.

VTT-AI-Agent (commented):
## TL;DR

The consolidation looks right and is now genuinely well pinned — I re-ran the suites on `4a375f5a` (jest 1563 green, `holderevents.js` 23 green, `enterleave.js` 28 green) and the two findings from review 3 (drop-shadow previews, stacked-holder detection) are both fixed. One new bug survived: `Holder.applyLeave()` uses the transient `currentParent` field as its "a drag is in progress" marker, but that field is never cleared when a drag ends *inside* the holder it started in — so after a player rearranges a card in their own `childrenPerOwner` hand, a later `SET parent` out of that hand silently keeps `owner`, and the card lands on the table invisible to every other player.

## Findings

### 1. A rearrange inside a per-owner hand poisons the next `SET parent` out of it (`client/js/widgets/holder.js:102`)

```js
if(child.get('type') != 'deck' && !child.isBeingRemoved && this.get('childrenPerOwner') && child.currentParent !== this)
  await child.set('owner', null);
```

`currentParent` is only ever deleted inside `checkParent()` (`client/js/widgets/widget.js:485`), i.e. at the moment the widget stops overlapping the container it was picked up in. A drag that never leaves the holder's box never reaches that branch: `moveStart()` sets `this.currentParent` (`widget.js:2871`), `move()`/`moveEnd()` call `checkParent()` without `forceDetach`, the overlap test keeps returning true, and `moveToHolder(this.hoverTarget)` skips its own `checkParent(true)` because `this.currentParent == holder` (`widget.js:2841`). Nothing in `moveEnd()` deletes the field afterwards. So on that client the card carries `currentParent === handA` for the rest of the session.

The next departure that is *not* a drag then reads that stale field and concludes a drag is in progress, so it skips the release the new pipeline promises ("a `childrenPerOwner` holder releases the owner however the card left it"). Confirmed on this branch with a TestCafe probe next to your own per-owner fixtures — same room, `handA: { childrenPerOwner: true }`, `card1` owned by Alice, `go` = `SELECT card1` + `SET parent null`:

```
✓ control: SET parent out of a per-owner hand releases the owner
✖ probe:   SET parent after an in-hand rearrange drag
    AssertionError: owner released after the earlier rearrange: expected 'Alice' to be falsy
```

Consequences:

- the card sits on the table with `owner: "Alice"`, which is `foreign` on every other client (`widget.js:492-496` → `body:not(.edit) .widget.foreign { display: none }`), so the played card is invisible to the opponents — and nothing later clears it, because only `checkParent()`/`moveToHolder()`/`MOVEXY resetOwner` ever reset an owner;
- the result depends on which client ran the routine and on what that player happened to drag earlier, so two clients disagree about whether the release happens at all. That is the one property a state-sync engine should not have.

Suggested fix: stop using `currentParent` as the drag marker, or make it honest. `delete this.currentParent` at the end of `moveEnd()` fixes the leak in one line; keying the check on the child's own `dragging` property instead would remove the reliance on a field whose lifetime is this subtle (note `moveEnd()` clears `dragging` before the drop's `checkParent()`/`moveToHolder()`, so the drop still releases the owner exactly as it does today).

Worth a fixture for it — a rearrange inside the hand followed by a routine-driven `SET parent` out of it. This case cannot be caught in Layer A as it stands, because `dragWidget()` ends with `delete widget.currentParent` (`tests/client/engine/enterleave.test.js:95`) — the jsdom replay of a drag cleans up something the real drag does not.

### 2. `MOVEXY … resetOwner: false` out of a per-owner hand now clears the owner anyway (`client/js/widgets/holder.js:102`, `client/js/widgets/widget.js:2098`)

`MOVEXY` has an explicit opt-out for exactly this:

```js
if(a.resetOwner)
  await c.set('owner', null);
...
await c.set('parent', null);
```

With the new pipeline that last line releases the owner regardless, because the parent write *is* the leave. Verified both ways on this branch:

```
✖ MOVEXY resetOwner:false out of a per-owner hand - modern   (expected 'Alice', got no owner)
✓ MOVEXY resetOwner:false out of a per-owner hand - legacy
```

So a game that deals a card out of a hand onto the table while deliberately keeping it owned (a face-down "played" card that only its owner may see, a staging area) has no way to express that any more once the legacy flag is off — `resetOwner: false` became a no-op for `childrenPerOwner` holders. Migration bullet 4 covers this only implicitly ("releases the owner however the card left it"), and the operation's own parameter now contradicts the engine. Either honour the opt-out (e.g. a transient marker on the widget the way `movedByButton` works, checked in `applyLeave()`), or say plainly in the mode description and the PR body that `resetOwner: false` no longer preserves a per-owner hand's owner.

### 3. The drop-shadow fix *is* gated by the legacy mode, contrary to the description

The PR body says: "A drop-shadow preview raising nothing is **not** one of the six: it is a plain bug fix and the legacy mode does not bring it back." It does. The `dropShadowOwner` check lives in `applyParentChange()` (`widget.js:3171`), which only runs when the mode is off; `applyLegacyParentChange()` has no such check, and your own fixture pins the old behaviour — `tests/testcafe/enterleave.js:132`, *"Legacy: a dropShadow holder runs a doubled leave for the shadow widget too"*, expecting `enter/leave/leave/enter`.

Preserving it under the flag is defensible as legacy-mode policy; the sentence in the PR body is what is wrong, and it matters because the migration flags a very large share of existing games — so in practice almost every published game with a `dropShadow` holder keeps a preview that can score, deal or delete on a mere hover until its author unchecks the mode. Please either fix the sentence and add a seventh bullet to the tile's list, or move the `dropShadowOwner` suppression out of the gate (a preview firing routines is not a semantic a game can meaningfully depend on).

### 4. The detector cannot see holders that become stacked at runtime (`client/js/legacymoderegistry.js:122-131`)

Both branches read the *static* JSON: the regex needs `"stackOffsetY"` as a key (a routine carries it as a value — `{"func":"SET","property":"stackOffsetY"}` — and the alternation does not list it anyway), and the widget scan reads `properties.stackOffsetX/Y` off the top-level widget only. A holder that gets its stack offset from a `SET` in a routine, from `inheritFrom`, or from a `CLONE` of a widget that is not itself a holder therefore migrates without the flag and silently gets the new re-compaction on `SET parent`/`MOVEXY`/`DELETE`. `docs/compat.md` asks for over-inclusiveness here, and adding `stackOffsetX|stackOffsetY` to the alternation without the `\s*:` anchor would close it for a handful of extra false positives — cheap insurance, given the tile already appears for most games.

## Verified on this branch

- `npm test`: 1563 passed, 40 suites.
- TestCafe `holderevents.js`: 23 passed. `enterleave.js`: 28 passed.
- Review 3's finding 1 (drop-shadow previews executing real enter/leave) is fixed and covered by two fixtures, one of which asserts the *whole room* is unchanged after a cancelled hover — that is the right assertion for it.
- Review 3's finding 2 (plain stacked holders missing the flag) is fixed: the detector's holder scan, four updater tests, the legacy layout fixtures (`DELETE`/`SET parent` leave the hole open) and the sixth bullet in the tile are all there.
- Standing checks: no new widget property and no new routine operation, so no property-editor or routine-block UI is owed; the new switch surfaces through Game Settings straight from the registry. No `library/` path is touched and the three demo variants live on this PR's `pr-test-ai` room, as required.

I did not add the `beta` label: this PR rewrites the parent-change/enter-leave pipeline in `client/js/widgets/widget.js`, `holder.js`, `line.js` and `pile.js` and changes how existing games run by default, which is engine work and therefore excluded from beta auto-deployment by design.

VTT-AI-Agent:
All four findings were valid and all four are addressed in `734ee9d8`. Reply to [pullrequestreview-3676697698](https://github.com/ArnoldSmith86/virtualtabletop/pull/3134#pullrequestreview-3676697698).

## 1. A rearrange inside a per-owner hand poisoning the next `SET parent`

Real bug, and the diagnosis is exact — thank you for the probe, it reproduced immediately.

Fixed both ways you suggested, because they fix different things:

- **`Holder.applyLeave()` now asks the card**, not the field: `!child.get('dragging')`. That is the property the question was really about — "is this leave the pickup of a drag that is still going on" — and `moveEnd()` clears `dragging` before the drop's `checkParent()`/`moveToHolder()`, so the drop still releases the owner exactly as it did.
- **`moveEnd()` now ends with `delete this.currentParent`.** The field means "the container this drag started in" and nothing else, so the drag has to give it back. Leaving it set was also feeding `moveToHolder()` (`widget.js:2839-2841`), where a stale value skips the `checkParent(true)` a move out of a *different* holder needs — and in legacy mode that reaches `dispenseCard()` on a holder the card is no longer in.

`MOVEXY … resetOwner: false` is the same question asked by a routine rather than by a pointer, so it uses the same shape as `movedByButton`: a transient `keepOwner` marker set around the `parent` write. See finding 2.

Tests:

- `tests/testcafe/enterleave.js` — *A rearrange inside a per-owner hand leaves nothing behind that keeps the next card owned*: Alice drags a card 150px inside her own `childrenPerOwner` hand, then a button runs `SELECT card1` + `SET parent null`, and the card on the table must have no owner. Verified load-bearing: with the two lines reverted it fails with `expected 'Alice' to be falsy`, which is your probe.
- `tests/client/engine/enterleave.test.js` — the same sequence in both combinations, and `dragWidget()` now sets and clears `dragging` where `moveStart()`/`moveEnd()` do. You were right that the replay was cleaning up something the real drag did not; it is faithful on this point now, and the `currentParent` delete it always did is what the engine does too.

## 2. `MOVEXY … resetOwner: false`

Also valid, and worse than a documentation gap: the parameter had become a no-op for exactly the holder it exists for. `MOVEXY` now marks the widget with a transient `keepOwner` for the duration of its `set('parent', null)`, and `applyLeave()` honours it, so the two pipelines agree again — a face-down "played" card that only its owner may see is expressible with the mode off.

Covered by `MOVEXY with resetOwner off keeps the owner of a card taken out of a per-owner hand` (TestCafe, asserts both the trace and the resulting state) and by the Layer A scenario in both combinations. Also verified load-bearing.

## 3. The drop-shadow sentence in the PR body

You are right; the sentence was wrong and I have replaced it. The suppression lives in `applyParentChange()`, the legacy pipeline has no equivalent, and `tests/testcafe/enterleave.js:132` pins the old behaviour on purpose.

I kept the gate rather than moving the check out of it. A holder whose `enterRoutine` lights something up while a card hovers over it is a game that *can* observe the preview's events, so taking them away from every migrated save silently is the one thing the legacy mode exists to prevent. What was wrong was claiming it wasn't gated. The body now says so, the author checklist has it as a seventh item, and the Game Settings tile has a seventh bullet — *the preview a `dropShadow` holder paints under the pointer raises nothing, so hovering over a holder no longer runs its routines*. The tile is 803px at 390×844, so it still fits one phone screen: [1280×800](https://agent.virtualtabletop.io/reports/pr/1539994146897399808/holder-events-legacy-tile.png), [390×844](https://agent.virtualtabletop.io/reports/pr/1539994146897399808/holder-events-legacy-tile-phone.png).

## 4. Holders that become stacked at runtime

Valid. The detector now has three passes:

- the event-name regex, unchanged;
- `"property"\s*:\s*"stackOffset[XY]"` — a routine that sets the offset while the game runs;
- the structural scan, with the `type == 'holder'` restriction dropped, so a template a holder `inheritFrom`s or clones counts even though the template is not a holder itself.

I did not add the bare names to the first alternation. `"stackOffsetX"` matches a holder that explicitly writes `stackOffsetY: 0` or `alignChildren: false`, and those two are the cases review 3 asked to keep out — the three passes above cover the routes you listed without giving that back. Two new updater tests: the routine-driven offset and the inherited one.

## Verification

`npm test` green (1569). TestCafe against a local server: `enterleave.js` 30, `holderevents.js` 23, `legacymatrix.js` 9, `interactions.js` 32, `multiclient.js` 6.

Agreed on `beta` — the diff changes default engine behaviour, so that is not the author's call either way.

PR #3135, opened 2026-08-20
Add the 50 icons game-icons.net gained since the last import

VTT-AI-Agent:
Brings `assets/game-icons.net/` up to date with the upstream [game-icons/icons](https://github.com/game-icons/icons) set: 50 icons have been published there since the last import and were missing from both icon pickers. This is purely additive — no existing icon is renamed, moved or removed, so every game that references one keeps working.

**Contact sheet of all 50, with the description, category and tags each one got:** https://agent.virtualtabletop.io/reports/pr/1540029655619473418/new-game-icons.html

### What is new

| artist | icons |
| --- | --- |
| caro-asercion (29) | the 22 Major Arcana (`tarot-00-the-fool` … `tarot-21-the-world`), `schrodingers-cat-alive`/`-dead`, `armchair`, `bucket-wheel-excavator`, `dodo`, `medieval-village-01`, `sink` |
| delapouite (16) | `aero-bike`, `armor-cuisses`, `attack-gauge`, `bird-scepter`, `encrypted-channel`, `extra-vision`, `hook-swords`, `lambda`, `manual-water-pump`, `mid-summit`, `plum`, `ringed-tentacle`, `sea-urchin`, `soul`, `warehouse`, `waterskin` |
| seregacthtuf (5) | `acid-shield`, `armor-blueprint`, `pouch-with-beads`, `sake-bottle`, `thought-bubble` — a new artist folder, so SeregaCthtuf is now named in the CC BY credits in `client/room.html` |

### How it was done

`assets/game-icons.net/icon-metadata.json` documents the procedure and it was followed step by step:

- The SVGs are the bare artwork the repo ships (upstream's black background rect dropped, `fill="#fff"` turned into `fill="#000"`).
- Every icon was **rendered and looked at** before it got its one-sentence description, its five category suggestions and up to 20 search tags — the file name was only a hint. The tags follow the rules in `_instructions`: lowercase ASCII, US spelling, each one adding a word the name and the earlier tags do not already carry.
- The same tags went into `assets/fonts/symbols.json`, which is what the two pickers actually read, under the category the suggestions map to.
- Both sprite sheets were regenerated. `symbols.json` addresses them by position in the montage, so **every** game-icons sprite index was recomputed from the new glob order, not just the new ones. The sheet grew from 69 to 70 rows, so `--spritesheetY` in `client/css/fonts.css` follows — without that the whole picker would show the wrong artwork. A new test ties that row count and both sheets' dimensions to the number of icons.
- The sheets are stored as PNG color type 4 (grayscale + alpha). The artwork is black on transparent, so an RGBA sheet spends two channels repeating the same gray: `overview48.png`, which the widget and deck-editor pickers fetch, is 2.7 MB instead of 4.4 MB, and `overview24.png`, the sheet the rich-text picker draws from, 1.2 MB instead of 1.8 MB — both *smaller than before this import* despite carrying 50 more icons. The pixels are identical either way.

Three additions to the `_instructions` block record what was not written down before: the form the SVGs are stored in, that the row count in `fonts.css` has to be updated with the sheet, and that the sheets are written as grayscale.

### Renamed and skipped upstream

game-icons.net also renamed three icons (`lorc/eskimo` → `person-in-blizzard`, `lorc/sattelite` → `satellite`, `lorc/star-sattelites` → `star-satellites`) and dropped `delapouite/female-vampire`. Renaming a file here breaks every game that references it by URL, so the files stay as they are — but the name the site shows today has to find them, so it is one of the icon's tags:

- searching **`person in blizzard`** now answers with `lorc/eskimo` and nothing else. It has to be one multi-word tag rather than separate words, because every word of a query has to match, down to the `in` that no single-word tag carries.
- `satellite` and `star satellites` already found the two misspelled satellites, which are tagged with the correct spelling.

A test covers all three, and `_instructions` records the rule for the next import.

Upstream's `badges/` folder (59 icons) is still not imported — those are small overlay badges of the other icons, a different visual family. That is now written down in `_instructions` too, instead of being rediscovered by the next run.

### Searching and browsing the new icons

The set of 22 is only worth importing if it can be found as a set, so a UI/UX pass over the picker led to three follow-ups:

- **The tarot cards are found by the name of the set.** `major arcana` answered with nothing and `tarot card` with the joker, while all 22 cards sat in the picker — their file names carry `tarot` and a number and nothing else. Each card now carries the name of its set in its tags, so both queries answer with the whole set, in numeric order, at the 64 px preview size. The name of a set stops there: the arcana are deliberately not tagged `deck`, because 22 more matches in a query the picker already fills twice over push the icons that are really about decks past the first 100 it shows. `_instructions` records both halves of the rule for the next import.
- **The two Schrödinger cats can be told apart.** They shared 16 of their 18 tags, so they always surfaced as a pair with the same tooltip, and `live cat` missed the live one altogether. Each of them now leads with the tags that distinguish it (`awake`, `live` / `limp`, `slumped`).
- **The icon that is called exactly what was typed leads the results.** A whole-word match said nothing about how much of the name it left over, so `soul` led with `delapouite/soul-vessel` and `shield` with `delapouite/american-shield` — the icon that owns the word ranked wherever the montage order happened to put it. Both pickers rank an exact name match above them now (`#symbolList .exactMatch` in the room picker, the existing score sort in the sidebar), in whichever number or spelling the query was typed — `souls` leads with `delapouite/soul` just like `soul` does.

### Verification

- `npm test` — 1354 passed, including `tests/game-icons-metadata.test.js`, which checks that both files cover exactly the SVGs on disk, that the sprite indices are the glob order, that the tags in the two files are identical, every tag rule, the sheet dimensions against `--spritesheetY`, that both sheets hold artwork in exactly the first 4181 cells and nothing after them (read out of their alpha channel, which catches a montage that stopped short or started at another offset), that a renamed icon is found by its current name, that the tarot set is found by the name of the set and not by `deck`, that the two cats are told apart, and that an exact name match ranks first.
- TestCafe `Deck editor: symbol pickers and JSON fallback` and `The symbol picker says an image-only search found nothing` pass locally.
- Walked the icon picker in a browser, in both of its preview modes: `major arcana` fills the picker with the 22 cards in order, and after the sheets were re-encoded the new icons render from the sprite sheet at the right position (the 22 tarot cards side by side), `person in blizzard` finds the parka figure, and the browser decodes the grayscale sheet with its 256 alpha levels intact.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3135/pr-test (or any other room on that server)


96LawDawg:
Leaving the renamed icons is fine, but for lorc/eskimo → person-in-blizzard, the user should be able to find the icon using the tags/file name currently on the game-icons.net website.  So "person in blizzard" must also find eskimo and it currently does not.

VTT-AI-Agent (commented):
## TL;DR

Solid, well-verified import — I re-ran the whole `game-icons-metadata` test logic standalone and independently decoded `overview48.png` to check that the artwork at each index really is the icon `symbols.json` claims, including the row-69 tail (exactly 41 filled cells = 4181 − 69·60), so the reindexing and `--spritesheetY: 70` are correct. Two things to fix: the new artist folder `seregacthtuf` is missing from the CC BY credit list in `client/room.html`, and both sprite sheets were re-encoded from grayscale+alpha to RGBA even though every pixel is gray, adding ~1.6 MB to `overview48.png` and ~590 KB to `overview24.png` for nothing.

## Findings

### 1. `seregacthtuf` is not in the CC BY attribution list — `client/room.html:594`

The credits paragraph names every game-icons.net artist folder individually (all 35 of them, down to one-icon contributors like `catsu` and `sparker`). This PR adds a 36th folder, `seregacthtuf` (5 icons), without adding it there. The trailing "and other various artists" arguably covers it, but the pattern in that file is clearly "name each artist", and CC BY 3.0 attribution is the reason the list exists. Add the artist's handle (`SeregaCTHTuF`) to that sentence — sensibly next to the other individual handles.

This is the only thing in the diff that is actually *missing* rather than optional.

### 2. The sprite sheets are now RGBA although every pixel is grayscale — `assets/game-icons.net/overview24.png`, `overview48.png`

The pre-PR sheets were PNG color type **4** (grayscale + alpha, 2 bytes/px); the regenerated ones are color type **6** (RGBA, 4 bytes/px). I decoded both new sheets and counted **0** pixels where R≠G≠B — the extra two channels are pure padding. What that costs:

```
                    before      committed    same pixels as
                    (type 4)    (type 6)     type 4, deflate -9
overview48.png      3694 KB     4371 KB      2687 KB
overview24.png      1646 KB     1786 KB      1197 KB
```

So a straight grayscale+alpha re-encode of the *identical* pixels is ~1.65 MB / ~590 KB smaller than what is committed — and smaller than the pre-PR sheets despite carrying 50 more icons. That estimate is plain `zlib.deflateSync(level 9)` with no filter search, so a real encoder does better still. `overview24.png` is fetched by every user who opens the icon picker and is served straight from `express.static` with no gzip (`server.mjs:143`), so this is a live bandwidth regression, not a repo-size nitpick.

Fix is a flag on the montage step — `-define png:color-type=4` (or `-colorspace gray -type GrayscaleMatte`), or run the output through `optipng`/`pngcrush`. Worth adding to the `_instructions` block in `icon-metadata.json` next to the new row-count note, since the same trap catches the next import. (The new sheets also dropped the old `gAMA`/`cHRM`/`bKGD`/`tIME` chunks, which is fine and slightly helps.)

### 3. Nothing tests that `--spritesheetY` matches the icon count — `tests/game-icons-metadata.test.js`

The PR description names this exactly right: "without that the whole picker would show the wrong artwork". That is the same class of silent breakage the sprite-index test in this file was written to catch, but the row count is still only prose in `_instructions` (and now in `client/css/fonts.css` where nothing checks it). It's three lines in a file that already computes `svgFiles`:

```js
test('fonts.css has a sprite sheet row for every icon', () => {
  const css = fs.readFileSync(path.join(dir, '../client/css/fonts.css'), 'utf8');
  expect(+css.match(/--spritesheetY:\s*(\d+)/)[1]).toBe(Math.ceil(svgFiles.length / 60));
});
```

Optionally the same test can read the two PNGs' IHDR (bytes 16–24) and assert `width == 60 * size` and `height == rows * size`, which would also have caught a sheet that was regenerated at the wrong tile width or not regenerated at all.

### 4. Two icons land in a surprising picker section

`categoryMapping` picks the first suggestion that maps, so:

- `caro-asercion/sink` — suggestions `Water, Furniture, …` → **Sea & Water**. A kitchen/bathroom sink next to sea urchins and waterskins is odd; swapping to `Furniture, Water, …` puts it with `armchair` and `waterskin` in **Furniture & Containers**.
- `seregacthtuf/armor-blueprint` — suggestions `Writing, Armor, …` → **Writing & Communication**. Defensible (it is a plan sheet), but a reader browsing armor won't find it; `Armor, Writing, …` reads better.

Both are found by their tags either way, so this is cosmetic — just the two placements I'd second-guess out of 50.

### 5. Stale duplicate of the montage procedure — `client/js/symbols.js:253`

That comment is a second copy of the rebuild instructions and is now further out of date than before: it names `overview.png` (the file is `overview48.png`), doesn't mention the 24px sheet, and doesn't mention `--spritesheetY`. Since this PR made `icon-metadata.json`'s `_instructions` the canonical, complete version, the code comment is better off pointing at it than repeating a stale subset.

Related, while that file is open: `tests/game-icons-metadata.test.js:33` was correctly updated to "4181 icons and 80000 tags" (actual: 80 783), but line 109 still says "search the real 13288 icons" — the index it builds actually holds 9 447 entries. That number was already wrong before this PR, so purely an FYI.

### 6. FYI — what is still missing from upstream

I diffed the local set against the upstream `game-icons/icons` tree. After this PR the delta is exactly: the 3 renames the PR already explains, plus the **59 icons in upstream's `badges/` folder**, which VTT has never imported. Presumably deliberate (they're small overlay badges, a different visual family), but a line in "Not included" would save the next import from rediscovering them.

## What I verified

- Re-implemented every assertion in `tests/game-icons-metadata.test.js` as a standalone script (jest isn't installed in this sandbox): 4181 SVGs, both files cover exactly them, sprite indices == glob order, tags byte-identical between the two files, all tag rules (count, dedup, ASCII, "adds a new searchable word"), 5 known-vocabulary categories each, `category` == first mapped suggestion, picker section agrees → **0 problems**. Plural/`news`/`crosses`/`lens` search tests still pass against the real `client/js/symbols.js`.
- Decoded `overview48.png` and rendered 32 sampled cells — index 0/1/2, the last two, and the icon before/after every insertion point — the artwork matches the name at that index in every case. Row 69 has exactly 41 non-empty cells and the first empty one is column 41, i.e. 4181 icons, confirming `--spritesheetY: 70` and the 1440×1680 / 2880×3360 dimensions.
- Compared unchanged icons' cells in the old vs new sheet at 4× — visually identical; the byte differences are sub-pixel AA from a different rasterizer, no shift or crop.
- Rendered all 50 new icons from the rebuilt sheet and read them against their descriptions — every description matches what is actually drawn (`sink` really is a basin with a mixer tap, `extra-vision` an eye in a disc casting beams, `soul` a figure between wisps, etc.). All 50 are a single `<path fill="#000">` on `viewBox="0 0 512 512"`, no `<script>`, `<use>`, `xlink:href`, `<image>` or external references.
- Spot-checked 5 of the new SVGs against `raw.githubusercontent.com/game-icons/icons` — byte-identical to upstream after exactly the documented transform (drop `<path d="M0 0h512v512H0z"/>`, `#fff` → `#000`).
- No engine file, `library/` file or tutorial is touched, and the new icons reach game authors through the existing pickers (`client/js/symbols.js` and the property editor's icon search both build their index from `symbols.json`), so no editor-UI work or demo room is needed for this one.

I added the `beta` label: the diff touches no engine file (no `client/js/widgets/`, `compute.js` or `serverstate.js`) — only assets, one CSS variable and a test comment — every check including CodeQL is green, and the change is purely additive, so it is safe and useful for players to try on the beta server.

VTT-AI-Agent:
Pushed 94636e77 — it fixes @96LawDawg's point and findings 1–5 of the review.

**The renamed icon has to be findable by its current name (@96LawDawg)** — right, and it was not: `person in blizzard` answered with nothing at all. The file keeps the name it has (renaming it breaks every game that references it by URL), so `lorc/eskimo` carries the current name as a tag now, and `person in blizzard` answers with that one icon. It has to be *one* multi-word tag: a query only matches if every one of its words matches something, and the `in` is carried by no single-word tag — separate `person` and `blizzard` tags would still have found nothing. `satellite` and `star satellites` already answered with the two misspelled satellites, which are tagged with the correct spelling. A test now covers all three, and `_instructions` records the rule so the next import handles a rename the same way.

**1. `seregacthtuf` missing from the CC BY credits** — fixed, `SeregaCthtuf` (the spelling the site uses) is in the list in `client/room.html`, next to the other artists with a handful of icons.

**2. The sheets were RGBA although every pixel is gray** — confirmed and fixed: 0 pixels with R≠G≠B in either sheet. Both are re-encoded as color type 4, pixel for pixel identical (decoded both, 0 differing pixels), and the browser reads them back with all 256 alpha levels.

| | before this PR | as committed before | now |
| --- | --- | --- | --- |
| `overview48.png` | 3694 KB | 4371 KB | **2672 KB** |
| `overview24.png` | 1646 KB | 1786 KB | **1187 KB** |

So both sheets are now smaller than they were before the import, with 50 more icons in them. `_instructions` says to write color type 4 and why.

**3. Nothing tested `--spritesheetY`** — added, close to your sketch, plus the IHDR check you suggested:

```js
test('the sprite sheets cover every icon and fonts.css has a row for each of them', () => {
  const rows = Math.ceil(svgFiles.length / 60);
  …
  expect(+fonts.match(/--spritesheetY:\s*(\d+)/)[1]).toBe(rows);
  for(const size of [ 24, 48 ]) {
    const ihdr = …
    expect([ ihdr.readUInt32BE(0), ihdr.readUInt32BE(4) ]).toEqual([ 60 * size, rows * size ]);
    expect(ihdr[9]).toBe(4);   // grayscale + alpha
  }
});
```

It fails with `Expected: 70, Received: 69` when the row count is put back to what it was.

**4. The two odd category placements** — both moved: `caro-asercion/sink` to Furniture & Containers (next to `armchair` and `waterskin` rather than the sea urchins) and `seregacthtuf/armor-blueprint` to Armor & Clothing. Their tags find them either way, but that is where someone browsing would look.

**5. The stale montage comment in `symbols.js`** — replaced with what the line actually does (`--x`/`--y` address the icon by its montage position) and a pointer to `_instructions`, which is the one complete copy of the procedure. The `13288` in the test file's comment is gone too — the number was wrong and would go stale again with every import, so it now just says "the real picker index".

**6. The 59 `badges/` icons** — still deliberately out (small overlay badges of the other icons, a different visual family), but that is written down now, in `_instructions` and in the PR description, so the next import does not have to rediscover them.

`npm test` is at 1349 passing, and the two symbol-picker TestCafe tests pass locally against the re-encoded sheets.

96LawDawg (approved):
Tested.  Approved.

VTT-AI-Agent (commented):
## TL;DR

**UI/UX review, not a code review.** Visually this import lands clean: I rendered the sprite sheet against the real SVGs across the whole montage (first cell, middle, the new last row, and nine of the 50 new icons) and all 22 sampled tiles match their artwork exactly, so the `--spritesheetY: 69 → 70` bump is right and the grayscale re-encode is visually indistinguishable ([proof sheet](https://agent.virtualtabletop.io/reports/ui-review/1540075273582547044/pr3135/01-sprite-vs-real-svg.png)). The most impactful improvement is in the **search tags, not the layout**: `major arcana`, `tarot card` and `tarot deck` all return 0-or-1 results while 22 perfect matches sit in the picker — the same multi-word-tag trick the PR already invented for `person in blizzard` would fix it.

---

### What I actually exercised

Local server on a checkout of `add-new-game-icons`, driven with Playwright. Overlays opened one by one:

* **Game Shelf**, **Players**, **About** (incl. the changed credits paragraph) — at **390×844**, **820×1180**, **1280×800**, **1920×1080**.
* **Add-a-widget overlay**, **Edit-widgets sidebar** and its inline icon picker — 820, 1280, 1920.
* **`#symbolPickerOverlay` ("Pick icon") through all three of its entry points**: the sidebar picker's *Show all*, the deck editor's suit-icon chip (`symbolPickerAboveEditor`), and the rich-text editor's `add_reaction` button (the only path that is *not* `bigPreviews`, i.e. the only one that actually renders `overview24.png`) — 820, 1280, 1920, plus 390 wide by resizing an open picker.
* **Deck editor**, **Add a new deck** dialog, its *Create a custom deck → Suits and ranks* section, and the return-to-dialog behaviour after closing the picker — 820, 1280, 1920.
* **Game/state details editor** (the *Save room → new game* flow) at 1280.

**Could not reach:** at **390×844 portrait** Edit Mode replaces the board with *"Please rotate your device to landscape orientation"* and the editor sidebar (and therefore both icon pickers) is unreachable — a pre-existing gate, not something this PR causes ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1540075273582547044/pr3135/10-editmode-390x844.png)). I did not open the chat panel, share/publish or the state save/variant dialogs; the diff touches no CSS they share (the only style change is one custom property scoped to `#symbolPickerOverlay .gameicons`).

I also opened this PR's `pr-test-ai` room — the personal shelf there is empty, so there is no draft tutorial to review.

---

## Big layout improvements

**1. A 22-icon numbered sequence is invisible when you browse instead of search.**
The Major Arcana land inside *Game-icons.net — Magic & Occult*, a ~350-icon block rendered at 24 px. At that size they are 22 near-identical dense figures with no labels and no visual grouping ([grid at 1280](https://agent.virtualtabletop.io/reports/ui-review/1540075273582547044/pr3135/02-picker-grid-1280x800.png), [at 1920](https://agent.virtualtabletop.io/reports/ui-review/1540075273582547044/pr3135/11-picker-grid-1920x1080.png)). The picker's own escape hatch works beautifully — searching `tarot` drops under 100 matches and `#symbolPickerOverlay.bigPreviews.fewResults #symbolList` re-renders everything at 64 px from the real SVGs, and there the set reads perfectly and in numeric order 00→21 ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1540075273582547044/pr3135/03-search-tarot-1280x800.png)). The problem is that you only get there if you already know the word to type. Since the whole value of this import is a *coherent set*, the cheapest fix is the tag one in the Readability section below, so `tarot card` / `major arcana` land you in the 64 px view.

**2. The full picker opened from the sidebar renders *below* the sidebar, not over the screen.** (Pre-existing, not introduced here — but this PR is what makes the picker worth opening.)
`pickSymbolKeepingOverlay` gives the deck-editor path a proper floating overlay (`symbolPickerAboveEditor`), and it looks great: centred card, whole viewport, obvious close button ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1540075273582547044/pr3135/07-deckeditor-picker-1280x800.png)). The sidebar's *Show all* has no host `.overlay`, so it falls through to plain `pickSymbol()` and the "Pick icon" card is laid out *after* the sidebar's own picker, Faces and Appearance sections — at 820×1180 you must scroll past three panels to reach the picker you just asked for, and the small inline picker stays visible above it, so two icon pickers are on screen at once ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1540075273582547044/pr3135/06-picker-below-sidebar-820x1180.png)). Proposal: route the sidebar's *Show all* through the same `symbolPickerAboveEditor` treatment so both entry points present the picker identically.

**3. Opening the full picker blocks the page for ~8–10 s with no feedback.** (Also pre-existing; this PR adds ~1.2 % to it.)
`pickSymbol` awaits `loadSymbolPicker()` *before* `showOverlay()`, and that call builds 13 361 `<i>` elements in one synchronous pass. I instrumented it: a `setInterval` registered right before the click fires **zero** times until the overlay appears — the main thread is blocked end to end, so the *Show all* button just looks dead. `#symbolList` already ships the string `Loading...`; showing the overlay first with that state and building the list afterwards would turn a frozen tab into an honest wait.

**4. `fonts.css:583` — hovering an icon does not enlarge it, but the help text promises it will.**
`room.html:539` says *"…hover one, or narrow the search below 100 results, to see the real image."* Narrowing does exactly that (64 px, real SVG). Hover does not: `#symbolPickerOverlay.bigPreviews .gameicons:hover { background: var(--url) }` swaps the sprite for the SVG **inside the same 24 × 24 box** — I measured the bounding box before and after hover, both `24×24`. So the two headline additions that differ only in fine detail stay unreadable on hover: `schrodingers-cat-alive` and `-dead` sit side by side and are indistinguishable ([screenshot, cats are 3rd and 4th in *Science & Space*](https://agent.virtualtabletop.io/reports/ui-review/1540075273582547044/pr3135/04-schrodinger-cats-24px.png)). Adding `transform: scale(2.5)` + `position: relative; z-index: 1` to that hover rule would make the sentence true without changing the grid.

**5. `room.html:594` — the CC BY credit list is a 35-name run-on sentence.**
`SeregaCthtuf` is inserted between `Felbrigg` and `Pierre Leducq`, which follows no order a reader can see (the list is neither alphabetical nor by icon count), so nobody checking whether an artist is credited can scan it ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1540075273582547044/pr3135/05-about-credits-1280x800.png)). Correct as-is and fine to merge, but the *Icons, Symbols, and Visual Assets* section would read much better with this one paragraph as a wrapped/multi-column `<ul>` in alphabetical order — the neighbouring paragraphs are short enough that this one is the only wall of text in the section.

**6. `overview24.png` is not what the widget/deck pickers fetch.**
The PR body justifies the re-encode partly with *"`overview24.png`, which every opened icon picker fetches"*. In practice `#symbolPickerOverlay.bigPreviews .gameicons` wins the cascade for both the widget-property picker and the deck-editor picker, so the only sheet that goes over the wire there is **`overview48.png` (2.61 MB)** — I confirmed with a response listener: one `overview48.png`, zero `overview24.png`. `overview24.png` is only fetched by the rich-text `add_reaction` picker, which drops `bigPreviews` ([that path renders correctly too](https://agent.virtualtabletop.io/reports/ui-review/1540075273582547044/pr3135/08-richtext-picker-overview24.png)). Nothing is broken — just noting the framing, since the perceived-load argument rests on the wrong file.

**7. Everything else held up at every size I tried.** No overflow, clipping, collision or unreachable close button in the picker at 820, 1280 or 1920; the `noResults` card ("No icons match "qqqzzz". Try a shorter or more common word.") sits centred and readable; closing the deck-editor picker restores the *Add a new deck* dialog correctly. At 390 the picker is usable but cramped ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1540075273582547044/pr3135/09-picker-390x844.png)) — same as before this PR.

**8. The obvious queries for the new tarot set return nothing.** This is the one finding I'd act on before merge.

| what a newcomer types | results |
| --- | --- |
| `arcana` | all 22, in order ✅ |
| `tarot` | 25 (the 22 + joker/sun) ✅ |
| `major arcana` | **0** |
| `tarot card` | **1** (`delapouite/card-joker`) |
| `tarot deck` | **1** (`delapouite/card-joker`) |

Every word of a query has to match, and none of the 22 carries `major`, `card` or `deck`. Someone who knows these images as "the major arcana" or who wants "tarot cards" is told the icons do not exist while 22 of them are sitting in the picker. The PR already solved exactly this shape of problem for `lorc/eskimo` by giving it the single multi-word tag `person in blizzard` (verified: that query returns that one icon and nothing else). Applying the same technique — `major arcana` and `tarot card` on all 22 — closes the gap with two tags per icon.

**9. The renamed icons are findable but rank low, which is most of the point of tagging them.**
`satellite` returns 20 matches and puts `lorc/sattelite` **15th**, behind `delapouite/beam-satellite`, `defense-satellite`, `satellite-communication` and four Material `satellite_alt*` glyphs. `star satellites` puts `lorc/star-sattelites` **3rd**, behind `caro-asercion/orrery` and `delapouite/defense-satellite`. A user who typed the name game-icons.net shows them today has to hunt for the icon that name belongs to. Consider ranking a match against a *complete* tag or file name above a word-inside-a-longer-name match, so the icon that literally owns the query leads. (`person in blizzard` is unaffected — it is unique.)

**10. Exact-name matches can rank below longer names.** `soul` lists `delapouite/soul-vessel` before the new `delapouite/soul`; same root cause as #9.

**11. `live cat` misses `schrodingers-cat-alive`.** `dead cat` correctly returns only `schrodingers-cat-dead` (nice), but its twin has no `alive`/`awake` tag and matching appears to be prefix- rather than substring-based, so the natural opposite query finds two unrelated cats instead. Worth one tag.

**12. The two Schrödinger icons carry near-identical tag lists** (16 of 18 tags shared) so they always surface as a pair, and at 24 px with no label a newcomer cannot tell which is which without resting the pointer on one and reading the tooltip. Combined with #4, "which cat did I just pick?" is a real question. One or two distinguishing leading tags per icon would fix it more cheaply than any layout change.

**13. Category placement of the new icons is well judged** — `sink` and `armchair` under *Furniture & Containers*, `hook-swords` under *Weapons*, `soul`/`extra-vision`/`bird-scepter` under *Magic & Occult*, `lambda` under *Symbols & Signs*. `armchair` also surfaces `delapouite/sofa` and `excavator` uniquely finds `bucket-wheel-excavator`, which is the behaviour a first-time user would expect.

VTT-AI-Agent:
Thanks — this one was worth reading twice. Pushed f4b7ed71: findings **1, 4, 8, 10, 11 and 12** are fixed, **6** is corrected in the description, and **2, 3, 5 and 9** are deliberately left alone, with reasons below.

### Fixed

**8 + 1 — the tarot set is found by the name of the set.** Exactly the diagnosis: the file names carry `tarot` and a number and nothing else, so nothing answered `major arcana`. Each of the 22 cards carries the set in its tags now (`major arcana` replaces the bare `arcana`, plus `card` and `deck`), so:

| query | before | now |
| --- | --- | --- |
| `major arcana` | 0 | 22, in numeric order |
| `tarot card` | 1 (the joker) | 23 |
| `tarot deck` | 1 (the joker) | 23 |

All three land under 100 results, so they arrive in the 64 px view you described, which is the whole point — the browse path is unusable for a set of 22 dense figures and the search path is the escape hatch. `_instructions` now carries the rule ("an icon that is part of a set carries the name of the set as a tag, as one phrase") so the next import does not rediscover it.

major arcana in the picker

**11 + 12 — the two Schrödinger cats.** Both fixed with tags, as you suggested, and both lead with what tells them apart: the live one starts with `awake, live`, the dead one with `limp, slumped`. `live cat` finds the live one, `limp cat` finds only the dead one, `dead cat` is unchanged, and the two tooltips no longer read alike.

**10 — an exact name match now leads.** A whole-word match said nothing about how much of the name it left over, so the icon that owns the query ranked wherever the montage order put it. `iconSearchScores` gives 4 to an entry whose *whole* name is what was typed, `#symbolList .exactMatch` gets `order: -2`, and the sidebar picker sorts by the same score, so both pickers agree as before. `soul` leads with `delapouite/soul`, `anvil` with `lorc/anvil`, `shield` with `sbed/shield`. Covered by tests in both suites.

**4 — hovering now magnifies.** `transform: scale(2.5)` + `z-index`, exactly your proposal, but only where the tiles are 24 px (a narrowed search already draws them at 64 px). While I was there: the hover swap itself was scoped to `.bigPreviews`, so in the rich-text picker — the one path that actually renders `overview24.png` — hovering did nothing at all; it now draws the real SVG there too. Measured 24×24 → 60×60, artwork from the icon's own file, and the grid does not move:

hover magnifies the icon out of the grid
the same in the rich text picker

**6 — the framing was wrong and is corrected.** You are right that the widget and deck-editor pickers fetch `overview48.png`; `overview24.png` is the rich-text picker's sheet. The description says that now. The re-encode still pays for itself on both sheets — it is the same 1.7 MB on the sheet that the two common pickers do fetch.

### Not changed

**9 — the renamed icons still rank as tag matches.** `lorc/sattelite` cannot become a name match without renaming the file, which breaks every game that references it by URL, and the new exact-name tier does not reach it either. Promoting *whole tag* matches above name matches is the version I did not want to ship: 54 icons carry the tag `fire` exactly, against 27 with `fire` in their name, so `fire` would answer with 54 icons that merely have flames in them before the ones that are called fire — a much worse trade than a satellite sitting 15th. `satellite` returns 16 icons, so all of them are on screen at once in the 64 px view, and `star satellites` returns 3.

**2 (the sidebar's *Show all* lays out below the sidebar) and 3 (the ~8 s main-thread block while the list is built).** Both real, both pre-existing, and neither is about this import — routing *Show all* through `symbolPickerAboveEditor` and showing the overlay before building 13 361 elements are each a change to the picker's own plumbing with their own regression surface. They deserve their own PR rather than riding along with an asset import; I have not touched them here.

**5 — the CC BY credits paragraph.** This PR only adds one name to it. Turning 36 run-on names into an alphabetical multi-column list is an About-page change, not an import change, so it is left for one.

**7, 13** — nothing to do, thank you for checking.

### Verification

`npm test` is at 1353 passing (4 new tests: the set-name search, the two cats, the exact-match ranking in the room picker, and the same ranking in the sidebar picker). The two symbol-picker TestCafe tests still pass. Walked the picker in a browser at 1280×800 in both preview modes for the screenshots above.

96LawDawg:
From the previous update, revert 4 — hovering now magnifies. This does not work well, does not look good, and goes beyond a simple import of new files.

VTT-AI-Agent:
Reverted in 393e1c46 — hovering an icon in the picker is back to what it did before this branch.

> From the previous update, revert 4 — hovering now magnifies. This does not work well, does not look good, and goes beyond a simple import of new files.

I took "revert 4" to mean the whole hover item, not just the scaling, since the second half of it (making the rich-text picker swap in the real artwork on hover too) is the same kind of change to the picker rather than to the import. So `client/css/fonts.css` is byte-identical to `main` in that block again: `#symbolPickerOverlay.bigPreviews .gameicons:hover` swaps the sprite for the icon's own file inside the 24 px tile, and nothing scales or lifts out of the grid. Measured in a browser: the tile stays 24×24 on hover, `transform` is `none`, and the background still switches from `overview48.png` to the icon's SVG. The description no longer advertises the magnifier.

The other three findings from that round stay in — they are tags and search ranking, no visual change: the tarot set found by `major arcana`/`tarot card`/`tarot deck`, the two Schrödinger cats told apart, and an exact name match leading the results.

Say the word if you'd rather the rich-text picker kept its hover swap; it is two characters in the selector.

`npm test` — 1353 passed.

VTT-AI-Agent (commented):
## TL;DR

PR #3135's icon import is safe and well covered, but the new exact-name ranking has one user-visible normalization gap: plural queries can still put a longer compound icon name ahead of the icon that owns the word. There is also one stale network-behavior claim in the newly expanded import instructions.

## Findings

### 1. Exact-name priority is lost for normalized plural/spelling queries — `client/js/symbols.js:213-216`

The search already treats singular/plural forms and US/UK spellings as equivalent through `iconSearchTerm()`, but the new score-4 check bypasses those term forms and compares `entry.name` to the literal normalized query words. Against the real picker index, this produces:

- `soul` → `delapouite/soul` first, but `souls` → `delapouite/soul-vessel` first and `delapouite/soul` second.
- `anvil` → `lorc/anvil` first, but `anvils` → `lorc/anvil-impact` first and `lorc/anvil` second.

That leaves the exact-name fix dependent on which grammatical form the user typed, even though the rest of this search deliberately makes those forms equivalent. Consider awarding score 4 when the entry has the same number of name words as the query and each name word matches the corresponding term's normalized forms, then add plural (and ideally spelling-alias) cases beside the new exact-name test in `tests/client/property-inputs.test.js`.

### 2. The import instructions still misstate which picker downloads `overview24.png` — `assets/game-icons.net/icon-metadata.json:19-20`

The PR description was corrected after the UI review, but the new canonical instructions still say that `overview24.png` is fetched every time a picker opens. `client/css/fonts.css:583-585` overrides it with `overview48.png` for the `bigPreviews` widget/deck-editor paths; the 24 px sheet is used by the rich-text picker. This does not undermine the grayscale optimization, but it should describe the two consumers accurately so the next import does not repeat the already-resolved misconception.

## Verification and scope

- All hosted build, TestCafe, production-start, and CodeQL checks pass at head `393e1c46`.
- I executed the current search functions standalone against the real picker index to reproduce the ranking examples above. The full local Jest command could not start because this clone does not contain the Jest binary.
- No engine path or `library/` path is touched. These additions use the existing author-facing icon pickers and add no widget property or routine operation, so no separate property/block UI or tutorial-quality demo room is required.

The `beta` label was already present; I left it in place because this diff touches no engine paths, all hosted checks including CodeQL are green, and the remaining findings do not make player testing unsafe.

VTT-AI-Agent:
Both findings were valid; fixed in 049f2d5b.

### 1 — exact-name priority now uses the same word forms as the rest of the search

Right diagnosis: the score-4 test compared `entry.name.join(' ')` to the literal normalized query, while every other tier goes through `iconSearchTerm()`. So the tier that is supposed to say "this icon *is* what you typed" was the one place where the number of the query mattered.

`iconSearchExactName(entry, terms)` replaces the string comparison: as many words in the name as there are terms, and each name word is one of the forms its term stands for. Against the real picker index:

| query | before | now |
| --- | --- | --- |
| `soul` | `delapouite/soul` | `delapouite/soul` |
| `souls` | `delapouite/soul-vessel` | `delapouite/soul` |
| `anvil` | `lorc/anvil` | `lorc/anvil` |
| `anvils` | `lorc/anvil-impact` | `lorc/anvil` |
| `shields` | `delapouite/american-shield` | `sbed/shield` |

The forms carry the spelling aliases too, so this also covers the case the index does not currently contain: an icon named `sabre` leads a search for `saber`, and vice versa.

Tests as you suggested, in both suites: the plural queries beside the existing exact-name test in `tests/game-icons-metadata.test.js` (against the real index), and `dragons` plus the `saber`/`sabre` pair beside it in `tests/client/property-inputs.test.js` (the sidebar picker, synthetic index). `npm test` — 1353 passed.

### 2 — the import instructions no longer misstate which picker fetches which sheet

Corrected in `_instructions`, and it now names both consumers rather than dropping the claim, so the next import has the whole picture: the widget and deck-editor pickers draw their grid from `overview48.png` (`.bigPreviews` overrides the sheet, the cells stay 24 px), the rich text editor's picker from `overview24.png`. The grayscale re-encode is still justified there, just on the right file — it saves ~1.7 MB on `overview48.png` alone, which is the sheet the two common pickers download.

VTT-AI-Agent (commented):
## TL;DR

The three findings from the earlier rounds are all fixed (credits, grayscale sheets, plural exact-name), and I re-verified the data end to end myself: 4181 SVGs, indices exactly the glob order, tags identical in both files, both sheets 70 rows / PNG color type 4, and — rasterizing 18 sampled SVGs and comparing masks against the sheet cells — cell *i* really is `svgs[i]`, including the new icons and the last row. One thing left worth a decision: giving all 22 arcana the bare tag word `deck` makes `deck` return 124 icons instead of 102 and pushes 22 genuine card icons (`faithtoken/card-draw`, `lorc/ace`, `delapouite/up-card`, …) out of the sidebar picker's 100-result window.

## Findings

### 1. The generic `deck` tag on all 22 arcana costs the `deck` query 22 of its visible results — `assets/fonts/symbols.json`

Each arcanum ends its tag list with the two bare words `card, deck`. `card` is harmless (the arcana rank 155–176 for that query, far past what anyone scrolls), but `deck` is a small query: it matched 102 icons before this PR and matches 124 now, and the 22 arcana land at ranks 18–39 because a tag match is a tag match. The sidebar icon picker shows `searchIconIndex(query, 100, …)` (`client/js/editor/propertyInputs.js:1422`), so exactly 22 icons fall off the end — and they are the ones a board-game author typing "deck" is looking for:

```
delapouite/snowboard, delapouite/up-card, faithtoken/card-discard, faithtoken/card-draw,
faithtoken/card-pick, faithtoken/card-random, lorc/ace, lorc/fire-ace, … (22 total)
```

(measured by running the pre-PR and post-PR `iconSearchScores` over the pre-PR and post-PR `symbols.json`; the room overlay is unaffected, it hides rather than truncates.)

This is precisely the case the PR's own new `_instructions` paragraph warns about — *"a word carried by a whole section buries the icons that are really about it"* — so it reads as an oversight rather than a decision. The word only exists to make the query `tarot deck` work, and `tarot`, `tarot card` and `major arcana` all answer with the full set without it. Dropping `deck` from the 22 (keeping `card`) restores the `deck` results and loses only that one phrasing; keeping it is defensible too, but then it is worth saying so in `_instructions` next to the rule it contradicts.

### 2. Nothing pins the sheet *content* to the SVGs, which is the one thing the reindex can get wrong — `tests/game-icons-metadata.test.js:51`

The new test is a good addition, but it only asserts the sheets' dimensions and color type. A montage that rendered the same 4181 files in a different order, or stopped short, produces a sheet of exactly the right dimensions and color type with every icon drawn at the wrong offset — the failure the test is there to prevent. A cheap deterministic check that needs no rasterizer: decode the alpha channel of `overview24.png` (plain `zlib.inflateSync` + un-filter, ~0.5 s for the whole sheet here) and assert that cells `0 … 4180` are all non-empty and the 19 trailing cells of the last row are all empty. That catches a short or shifted montage; it is what I ran here, and both sheets pass (4181 filled, contiguous, 19 empty, and no pixel with alpha > 0 has gray > 16, so the artwork really is black-on-transparent).

Not blocking — I confirmed the alignment out-of-band instead: rendering 18 SVGs at 48 px in Chromium and comparing binary masks against cells, `andymeneely/police-badge`→0, `delapouite/cable-stayed-bridge`→500, `lorc/candle-flame`→2500, `viscious-speed/abstract-091/092`→4139/4140, `zeromancer/heart-plus`→4180, `caro-asercion/tarot-00-the-fool`→151, `delapouite/soul`→1883, `seregacthtuf/thought-bubble`→3871 and the rest all match their expected cell at IoU 0.99–1.00, and always beat both neighbours by a wide margin.

## What I checked and found nothing wrong with

- **Purely additive to existing games.** Diffing `symbols.json` against `main`: 0 icons removed, 0 moved category, 50 added, and exactly one pre-existing entry changed at all — `lorc/eskimo`, which gained the `person in blizzard` phrase. Every existing icon URL and every existing tag still resolves.
- **The exact-name tier.** `iconSearchExactName()` (`client/js/symbols.js:194`) can only promote entries `iconSearchScore()` would already have given 3 — every name word whole-matches its term by construction — so score 4 is a strict re-ranking with no new matches and no way to reorder the 3/2/1 tiers below it. The `forms`-based comparison from 049f2d5b closes the plural/spelling gap the previous round found: `souls`, `anvils`, `shields`, `saber` now lead with the same icon as their other form. Emoji entries have an empty `name` array and can never be exact, so nothing odd happens there. `.exactMatch { order: -2 }` slots cleanly above `.nameMatch (-1)`, prefix matches (0) and `.tagMatch (1)`.
- **The hover-preview revert (393e1c46)** leaves `client/css/fonts.css` byte-identical to `main` for that rule, so the 24 px rich-text picker keeps its previous hover behaviour — no leftover from the magnifier experiment.
- **The 50 new SVGs** are all `viewBox="0 0 512 512"`, `fill="#000"`, no background rect, no `#fff`, same one-line shape as the existing files. Rendered as a contact sheet, every one matches its file name and its `icon-metadata.json` description (`attack-gauge` really is a sword over a nearly-full bar, `mid-summit` a flag partway up a peak, `medieval-village-01` a half-timbered house beside a spired tower).
- **Category balance** after the import: 292 (Weapons) down to 62 (Arrows & Direction) — the 29 new caro-asercion icons did not lump into one bucket.
- No engine file, no `library/` path, no widget property and no routine operation is touched, so `validate_gamefile.js` / `jeCommands` upkeep and a tutorial-quality demo room do not apply here; the new icons are reachable from the real editing UI through the existing widget/deck/rich-text pickers.
- All hosted checks are green at `049f2d5b` (Analyze, CodeQL, Production Environment, the full TestCafe matrix). I could not run `npm test` in this sandbox (`jest-environment-jsdom` is not installed here), so I re-implemented the metadata and search assertions standalone in node instead; all of them hold.

The `beta` label was already on the PR and I left it there: the diff touches no engine path (no `client/js/widgets/`, `compute.js` or `serverstate.js`), is additive for every existing game, and both remaining points are search-ranking polish that cannot corrupt a room or crash the server.

VTT-AI-Agent:
Both findings were valid; fixed in `cc4d128d`.

### 1 — the arcana no longer carry `deck`

Your diagnosis and my own measurement agree, and it was an oversight rather than a decision: `deck` was on the 22 only to make the one phrasing `tarot deck` work, and it cost that query a fifth of what it is for. Running the search over the branch against `main`:

| query | main | with `deck` | without `deck` |
| --- | --- | --- | --- |
| `deck` | 102 | 124 (arcana at 17–38, 24 icons past the 100 shown) | **102, in the same order as `main`** |
| `tarot` | 3 | 25 | 25 |
| `tarot card` | 1 | 23 | 23 |
| `major arcana` | 0 | 22 | 22 |
| `tarot deck` | 1 | 23 | 1 |
| `card` | 244 | 266 (arcana at 155–176) | 266, first 100 unchanged |

So the `deck` result list is now byte-identical to `main`'s, including its order, and the set is still found by `tarot`, `tarot card` and `major arcana`. `tarot deck` goes back to answering with the joker, which is the one thing lost.

`_instructions` now carries the second half of the rule next to the first, so the next import does not re-add it: the words of a set-name phrase are still tags of their own, so a set name may not smuggle in a broad word another family owns — with the numbers above as the example. The test for the set asserts it too: `major arcana` and `tarot card` return all 22, `deck` returns none of them.

### 2 — the sheets are now pinned by their content

Also right, and worth doing: dimensions plus color type is exactly what a short or offset montage still satisfies. `decodeGrayAlphaPNG()` inflates the IDAT chunks and un-filters the scanlines (~60 lines, no new dependency), and the test asserts on **both** sheets that the first 4181 cells hold artwork, that the 19 trailing cells of the last row are empty, and — since the alpha scan is already walking the pixels — that no opaque pixel is lighter than black, which catches a montage built from game-icons.net's own downloads (white on a black background rect) rather than from the files here.

I checked that it actually bites rather than just passing: re-running the same cell scan over the decoded sheet shifted by one cell fails it, and over the sheet with its last row cleared fails it too. 1.6 s for both sheets.

It does not catch a *reordering* of the same 4181 files, which your out-of-band mask matching does — the test name and its comment say what it covers rather than implying alignment.

### Verification

`npm test` — 1354 passed (the new sheet test; the set test now covers the `deck` case as well). The diff is 3 files: the `deck` tag off 22 entries in both metadata files, the instructions paragraph, and the test.

PR #3137, opened 2026-08-20
Write imported PCIO and TTS games at the current file version so they get no legacy modes

VTT-AI-Agent:
Games imported from a PlayingCards.io or Tabletop Simulator file started out with legacy modes switched on, so brand new imports behaved like games from before those bugs were fixed. This makes both importers write their output at the current file version, which is what they actually produce - and, where dropping a migration raised the question of what the imported game *should* do, answers it from what the source platform does.

### The bug

`server/pcioimport.mjs` stamped its output with `_meta.version: 4` and `server/ttsimport.mjs` with `5` - the file versions they were written against, never updated since. Loading such a file runs every `FileUpdater` migration from there to `VERSION`, and those include the steps that enable a legacy mode for games that predate a behavior change:

* every PCIO import contains `var` expressions (counters, spinners, turn buttons), so it tripped the detectors of `convertNumericVarParametersToNumbers` and `useOneAsDefaultForVarParameters` and started with **both var legacy modes active** - old var semantics in a game written by today's importer, which is what the report below ran into,
* every TTS import with a hand zone carries a holder with `text: 'Hand'`, so it tripped `disableHolderImageWidget` and started with **holder image support disabled** - hiding the very caption the importer had just written.

Legacy modes exist so that games saved *before* a change keep working. A file an importer wrote a second ago was never played under the old semantics, so it must not get any of them.

### The change

Both importers write `_meta.version: VERSION`, so a fresh import is loaded as what it is: a new file.

Stamping the current version also means the migrations that silently corrected importer output on load no longer run, so those corrections are written by the importers themselves:

* **face objects** that follow a card type property become `dynamicProperties` instead of PCIO's `valueType`/`value` (v5),
* **hands** get `dropShadow` and `hidePlayerCursors` - what v12/v14 used to add and what edit mode's "add hand" button sets,
* a **classic PCIO game piece** is written at 56x84 at x+17/y+3, the box its pawn shape actually fills, instead of the 90x90 PCIO box that v6 used to shrink. A 90x90 box stretches the `classicPiece` mask (`viewBox='0 0 60 90' preserveAspectRatio='none'`) about 1.5x horizontally: classic piece geometry
* **every label** the PCIO importer emits is grown to at least one line of its own font size (v13), not just the striped placeholder,
* a **TTS notecard** gets `white-space: pre-wrap` (v20) so the runs of spaces and the line breaks its author typed survive.

Two of the dropped migrations turned out to have been correcting the importer *towards* something PlayingCards.io does not do, so the importer writes what PlayingCards.io does instead - see below.

### Measured against what PlayingCards.io actually does

Their client ships as one bundle (`https://playingcards.io/build/bundle.js`), so the questions below are answered by reading and running their code rather than by comparing the import with its previous output.

* **Moving a number of objects that is only known when the button is pressed** ("deal as many cards as this counter says") keeps its plain `MOVE`/`MOVEXY`/`ROTATE`. v16 wrapped exactly that shape into an `IF` whose `else` branch moved **everything** when the number came out as 0, so a counter at 0 dumped the whole pile onto the table. PlayingCards.io declares the quantity as `numberMinimum: 1` with "all objects" as the constant 1300 - a 0 can only ever arrive from a counter or an automation question at run time, and `MOVE_CARDS_BETWEEN_HOLDERS.run` then loops `while(sum(quota) > 0)` and moves nothing at all. Lifting their function out of the bundle and running it confirms it: 0 → nothing, 2 → two, 1300 → everything. The plain operation the importer writes does the same, and there is a test for the counter-driven count and for a 0 spelled out in the file.
* **A recall hands the objects back to everyone.** v11 used to add `resetOwner: false` to the `MOVEXY` a recall of a deck without a holder emits, so recalled objects stayed the property of whoever had them in their hand. PlayingCards.io clears the owner of every object it recalls, in a holder or on the deck's own position, so the import does too.
* **A move into the hand ends up in the hand.** A move whose destination is the hand widget was written as a `MOVEXY` without coordinates, which drops the objects in the top left corner of the table. PlayingCards.io accepts only holders and seats as destinations - a hand is dropped from the list, so such a step moves nothing there. The import now moves the objects into the hand, where they become the private property of whoever pressed the button, which is what a hand means on both platforms; a hand that PlayingCards.io does not show is no destination at all and the step is left out. (Moves to a *seat* already matched: PCIO sets `owner` to that seat's player and parks the object in the hand, which is what VTT's `MOVE` to a seat does.)
* **A counter is laid out the way PlayingCards.io lays it out.** A counter has no stored size there: the box is computed from the size of its value - room for the widest number its bounds allow (at least 52px) plus 44px for each of the two buttons, one line tall but never shorter than the buttons - and the value is drawn a quarter larger while it is shorter than four characters. The `width`/`height` a `.pcio` file carries for a counter are numbers PCIO ignores, so the importer ignores them too and computes the same box, with a 32px button in each end. An imported counter is therefore never too short for its own value and never too narrow for its buttons, whatever the file says.

the counters of the demo table

### The hand zone is captioned

A PCIO import ended in an unlabeled 1400x180 white band across the bottom of the table, and a hand's own PlayingCards.io label was dropped on the way in. The hand now carries that label as its caption, falling back to *"Your hand"* for the table's main hand and *"Private"* for a further private zone. PlayingCards.io does not name its hand either, but its empty hand is not blank: it shows grey centred text - *"Cards placed here are only visible to you."* - which fades out as soon as a card is in it. The TTS hand says *"Your hand"* as well now instead of *"Hand"*, which reads as the name of a zone rather than as "this band is private to you". The caption only became visible at all with this PR, because `disableHolderImageWidget` used to hide it.

A demo table is on the test server as **PCIO import demo** ([download](https://agent.virtualtabletop.io/reports/pr/1540040772802584646/PCIO%20import%20demo.vtt)) - the raw output of importing a `.pcio` with four counters (one whose file box is 40x10, one bounded, one styled at 40px with a 1.6 line height, one without buttons), a deck, a hand and three buttons: deal as many cards as the counter says, move one into the hand, recall everything. It is a demo, not a library game and not a tutorial. Set the counter to 0 and press the deal button: nothing is dealt, where the migrated form dealt the whole pile.

the demo table

Existing rooms are unaffected: files imported before this change are still at version 4/5 and still get their legacy modes when loaded.

### Verified

`npm test` (1362 tests, including the new ones). Both importer test suites share one helper that checks every import they do against the migrations: the imported state must come out of `FileUpdater` unchanged when it is handed in stamped at `IMPORT_MIGRATION_BASE`, the oldest version from which every migration is known to leave importer output alone. (Handing it in at the current version proves nothing - `FileUpdater` returns such a state untouched.) Every migration added later therefore runs against every importer fixture, however many versions a release adds at once, and fails the tests if the importer does not already write what it produces.

Beyond the suites, both importers were run over a fixture covering every widget type and automation step they know, and their output diffed against `FileUpdater(<the same output stamped back to version 4/5>)`, i.e. against exactly what `main` hands to the client today, so that every difference is a deliberate one rather than an accident.

In the browser, on the demo table: the counter at 3 deals three cards and at 0 deals none, "one into my hand" puts the card in the player's own hand, "recall everything" brings all thirteen cards back to the draw pile including the one that was in the hand, and the console stays clean. Manually, by importing the same `.pcio` file into a room on `main` and on this branch: on `main` Game Settings shows *"2 of 2 legacy modes active for this game"*, on this branch *"No legacy modes are active for this game"*.

### Reported on Discord

> Ok yeah, there are two set on the game... I imported this from a playingcards.io file, maybe it set those on import?

— moshek, [#Gin Rummy](https://discord.com/channels/770758631146782780/1538606143000158248/1540040452294709349), after `x.trim is not a function` errors in a game that worked identically in another room.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3137/pr-test (or any other room on that server)

### DIFF
--- a/server/pcioimport.mjs
+++ b/server/pcioimport.mjs
@@ -267,10 +267,13 @@ export default async function convertPCIO(content) {
 
     const text = options.noTextStyle ? null : widget.mainTextStyle;
     if(text && typeof text == 'object') {
-      if(text.size)
-        textCSS['font-size'] = `${Math.round(text.size)}px`;
-      if(text.size && text.lineHeight)
-        textCSS['line-height'] = `${Math.round(text.size*text.lineHeight)}px`;
+      // options.textSize is the size the text is actually drawn at where that is
+      // not the size of the style itself - a counter enlarges its value
+      const size = options.textSize || text.size;
+      if(size)
+        textCSS['font-size'] = `${Math.round(size)}px`;
+      if(size && text.lineHeight)
+        textCSS['line-height'] = `${Math.round(size*text.lineHeight)}px`;
       if(text.align)
         textCSS['text-align'] = text.align;
       Object.assign(textCSS, textFill(text.mainFill, !!options.textSelector, widget));
@@ -499,6 +502,47 @@ export default async function convertPCIO(content) {
   function roundCounter(variable, step) {
     const decimals = typeof step == 'number' ? (String(step).split('.')[1] || '').length : 0;
     return decimals ? [ `var ${variable} = \${${variable}} toFixed ${decimals}` ] : [];
+  }
+
+  // PCIO does not lay a counter out from the width and height in the file - it
+  // ignores both and builds the box from the size of the value: room for the
+  // widest number the bounds allow (at least 52px) plus a 44px button at each
+  // end, and a line of the value's own font, never shorter than those buttons.
+  // The value itself is drawn a quarter larger while it is shorter than four
+  // characters.
+  function counterLayout(widget) {
+    const style   = widget.mainTextStyle || {};
+    const size    = +style.size || 21;
+    const buttons = widget.counterShowButtons !== false;
+
+    // the bounds every PCIO counter has, which are what its box is wide enough for
+    const min = Math.max(-1e10, Math.min(1e10, isNaN(parseFloat(widget.counterMin)) ? -9999 : parseFloat(widget.counterMin)));
+    let   max = Math.max(-1e10, Math.min(1e10, isNaN(parseFloat(widget.counterMax)) ?  9999 : parseFloat(widget.counterMax)));
+    if(min >= max)
+      max = min + 1;
+    const characters = Math.max(String(Math.abs(min)).length + (min < 0 ? 0.5 : 0), String(max).length);
+
+    const value      = counterValue(widget);
+    const boosted    = widget.counterBoostFontSize !== false && String(value).length < 4 ? Math.min(size*1.25, 150) : size;
+    const fontSize   = Math.round(boosted);
+    const lineHeight = Math.round(fontSize*(style.lineHeight || 1));
+    return {
+      buttons,
+      buttonSize: 32,
+      buttonMargin: 6,
+      fontSize,
+      lineHeight,
+      width:  Math.max(52, characters*0.62*size) + (buttons ? 88 : 0),
+      // PCIO lets a value taller than the box overflow it while a VTT label cuts
+      // it off, so the box is never shorter than the line the value takes
+      height: Math.max(buttons ? 44 : 0, size, lineHeight + 2)
+    };
+  }
+
+  // the value a counter starts at, kept inside its bounds the way PCIO does
+  function counterValue(widget) {
+    const bounds = counterBounds[widget.id];
+    return bounds ? Math.min(Math.max(+widget.counterValue || 0, bounds.min), bounds.max) : widget.counterValue;
   }
 
   // the operations that keep a variable within the bounds of a counter
@@ -1086,40 +1130,34 @@ export default async function convertPCIO(content) {
     } else if(widget.type == 'counter') {
       w.type = 'label';
       w.y = (w.y || 0) + 5;
-      w.width = widget.width || 140;
-      w.height = widget.height || 44;
+      const counter = counterLayout(widget);
+      w.width = counter.width;
+      w.height = counter.height;
       const bounds = counterBounds[widget.id];
-      w.text = bounds ? Math.min(Math.max(+widget.counterValue || 0, bounds.min), bounds.max) : widget.counterValue;
+      w.text = counterValue(widget);
       if(widget.allowPlayerEditValue !== false)
         w.editable = true;
       // the text style belongs on the value, not on the caption and +/- buttons
       pcioStyle(widget, w, [], {
         textSelector: ' > textarea',
-        text: (widget.mainTextStyle || {}).size ? {} : { 'font-size': '30px' }
+        textSize: counter.fontSize,
+        text: { 'font-size': `${counter.fontSize}px`, 'line-height': `${counter.lineHeight}px` }
       });
       if(bounds && !isNaN(parseFloat(widget.counterValue)) && w.text != parseFloat(widget.counterValue))
         warn(`Counter ${widgetName(widget)} was outside its ${boundsText(bounds)} and starts at ${w.text} instead of ${widget.counterValue}.`);
       if(bounds && widget.allowPlayerEditValue !== false)
         warnAbout(`bounds ${boundsText(bounds)}`, widget, (names, count)=>`Typing a value into the counter${count > 1 ? 's' : ''} ${names} is not restricted to ${count > 1 ? 'their' : 'its'} ${boundsText(bounds)} - the buttons and the automations that change ${count > 1 ? 'them' : 'it'} are.`);
 
-      // the +/- buttons are sized from the counter, so its box has to have room
-      // for one line of the value before they are created
-      w.height = Math.max(w.height, labelMinHeight(w));
-
       const counterStep = Math.abs(+widget.counterStep) || 1;
-
-      // square buttons in the corners of the counter, shrunk to fit next to each
-      // other on a counter too narrow for two of them at the height's size
-      const buttonSize = Math.max(0, Math.min(w.height - 8, Math.floor((w.width - 12)/2)));
 
       function addCounterButton(suffix, x, text, value) {
         output[widget.id + suffix] = {
           id: widget.id + suffix,
           parent: widget.id,
-          x: 4,
-          y: -2,
-          width: buttonSize,
-          height: buttonSize,
+          x: counter.buttonMargin,
+          y: Math.round((counter.height - counter.buttonSize)/2),
+          width: counter.buttonSize,
+          height: counter.buttonSize,
           type: 'button',
           movableInEdit: false,
           text,
@@ -1137,9 +1175,9 @@ export default async function convertPCIO(content) {
         if(x)
           output[widget.id + suffix].x += x;
       }
-      if(widget.counterShowButtons !== false) {
-        addCounterButton('_decrementButton', 0,                        '-', -counterStep);
-        addCounterButton('_incrementButton', w.width - 8 - buttonSize, '+',  counterStep);
+      if(counter.buttons) {
+        addCounterButton('_decrementButton', 0, '-', -counterStep);
+        addCounterButton('_incrementButton', counter.width - 2*counter.buttonMargin - counter.buttonSize, '+', counterStep);
       }
 
       if(widget.label) {
@@ -1715,10 +1753,13 @@ export default async function convertPCIO(content) {
           const quantity = numberArgument(routine, args.quantity, 1);
           const fill = args.fillAdd && args.fillAdd.value == 'fill' && quantity !== 'all';
 
+          // a hand PCIO does not show is nowhere to move objects to, and without
+          // any destination left there is nothing to move
+          const destinations = args.to.value.filter(id=>!byID[id] || byID[id].type != 'hand' || byID[id].enabled !== false);
+          if(!destinations.length)
+            return;
           // PCIO starts dealing at the n-th of its destinations
-          const destinations = args.to.value.slice();
-          if(destinations.length)
-            destinations.push(...destinations.splice(0, (+(args.startingOffset || {}).value || 0) % destinations.length));
+          destinations.push(...destinations.splice(0, (+(args.startingOffset || {}).value || 0) % destinations.length));
 
           c = importWidgetQuery(routine, args, 'from', 'from', 'collection', {
             func:  'MOVE',
@@ -1738,12 +1779,6 @@ export default async function convertPCIO(content) {
             delete c.count;
           if(c.to.length == 1)
             c.to = c.to[0];
-          if(c.to == 'hand') {
-            delete c.to;
-            c.func = 'MOVEXY';
-            // the objects land on the table but stay the property of whoever held them
-            c.resetOwner = false;
-          }
           if(moveFlip && moveFlip != 'none')
             c.face = moveFlip == 'faceDown' ? 0 : 1;
 
@@ -1821,7 +1856,8 @@ export default async function convertPCIO(content) {
             return;
 
           // a deck that sits on the table has no holder to recall into, so its cards are
-          // gathered on its position instead - the ones a player owns stay that player's
+          // gathered on its position instead - a recall takes them back from their owner
+          // as much as it takes them out of a holder
           for(const deckID of decks) {
             if(!byID[deckID].parent) {
               output.tempHolderForDeckRecall = {
@@ -1844,8 +1880,7 @@ export default async function convertPCIO(content) {
                 from: 'tempHolderForDeckRecall',
                 x: byID[deckID].x + (86-(byID[deckID].cardWidth ||103))/2,
                 y: byID[deckID].y + (86-(byID[deckID].cardHeight||160))/2,
-                count: 'all',
-                resetOwner: false
+                count: 'all'
               });
             }
           }
--- a/server/pcioimport.mjs
+++ b/server/pcioimport.mjs
@@ -629,8 +629,11 @@ export default async function convertPCIO(content) {
       if(widget.kinged)
         w.activeFace = 1;
     } else if(widget.type == 'gamePiece' && widget.pieceType == 'classic') {
-      w.width  = 90;
-      w.height = 90;
+      // the PCIO piece is a 90x90 box while the pawn shape only fills 56x84 of it
+      w.width  = 56;
+      w.height = 84;
+      w.x = (w.x || 0) + 17;
+      w.y = (w.y || 0) + 3;
       w.classes = 'classicPiece';
       w.color = pieceColors[widget.color] || pieceColors.default;
     } else if(widget.type == 'gamePiece' && widget.pieceType == 'pin') {
@@ -1719,6 +1722,8 @@ export default async function convertPCIO(content) {
           if(c.to == 'hand') {
             delete c.to;
             c.func = 'MOVEXY';
+            // the objects land on the table but stay the property of whoever held them
+            c.resetOwner = false;
           }
           if(moveFlip && moveFlip != 'none')
             c.face = moveFlip == 'faceDown' ? 0 : 1;
@@ -1796,6 +1801,8 @@ export default async function convertPCIO(content) {
           if(!decks.length)
             return;
 
+          // a deck that sits on the table has no holder to recall into, so its cards are
+          // gathered on its position instead - the ones a player owns stay that player's
           for(const deckID of decks) {
             if(!byID[deckID].parent) {
               output.tempHolderForDeckRecall = {
@@ -1818,7 +1825,8 @@ export default async function convertPCIO(content) {
                 from: 'tempHolderForDeckRecall',
                 x: byID[deckID].x + (86-(byID[deckID].cardWidth ||103))/2,
                 y: byID[deckID].y + (86-(byID[deckID].cardHeight||160))/2,
-                count: 'all'
+                count: 'all',
+                resetOwner: false
               });
             }
           }
@@ -2218,8 +2226,7 @@ export default async function convertPCIO(content) {
       // striped block where the widget used to be
       const typeName = pcioTypeNames[widget.type] || widget.type;
       w.width = widget.width || 100;
-      // a label shorter than one line of text shows nothing at all
-      w.height = Math.max(widget.height || 100, 18);
+      w.height = widget.height || 100;
       w.type = 'label';
       w.text = `${typeName} not imported`;
       w.css ='background: repeating-linear-gradient(45deg, red, red 10px, darkred 10px, darkred 20px); color: white; text-shadow: 0 0 4px black;';
@@ -2236,6 +2243,16 @@ export default async function convertPCIO(content) {
   for(const seatID in turnAtSeat)
     if(output[seatID])
       output[seatID].turn = true;
+
+  // a label whose box is shorter than one line of its text shows nothing at all
+  for(const w of Object.values(output)) {
+    if(w.type != 'label')
+      continue;
+    const match = JSON.stringify(w.css || '').match(/font-size"?:"? *([0-9]+) *px/);
+    const minHeight = (match ? +match[1] : 16) + 2;
+    if((w.height || 20) < minHeight)
+      w.height = minHeight;
+  }
 
   for(const group of Object.values(groupedWarnings))
     warn(group.message(widgetNames(group.names), group.names.length));
--- a/server/pcioimport.mjs
+++ b/server/pcioimport.mjs
@@ -319,6 +319,13 @@ export default async function convertPCIO(content) {
       w.height = widget.height;
   }
 
+  // a label whose box is shorter than one line of its text shows nothing at all,
+  // so this is the smallest height that one line of its own font still fits into
+  function labelMinHeight(w) {
+    const sizes = [ ...JSON.stringify(w.css || '').matchAll(/(?:font-size|line-height)"?:"? *([0-9.]+) *px/g) ].map(match=>+match[1]);
+    return Math.max(16, ...sizes) + 2;
+  }
+
   // The objects an automation works on: either a holder parameter on the
   // operation itself or a SELECT into a collection. Returns null for a query
   // that matches nothing so that the caller can skip the whole step, which is
@@ -733,6 +740,8 @@ export default async function convertPCIO(content) {
       w.childrenPerOwner = true;
       w.dropShadow = true;
       w.hidePlayerCursors = true;
+      // an empty holder is a blank band otherwise
+      w.text = widget.label || 'Your hand';
       w.width = widget.width || 1500;
       w.height = widget.height || 180;
       if(widget.allowedDecks && widget.allowedDecks.length)
@@ -1090,6 +1099,10 @@ export default async function convertPCIO(content) {
         warn(`Counter ${widgetName(widget)} was outside its ${boundsText(bounds)} and starts at ${w.text} instead of ${widget.counterValue}.`);
       if(bounds && widget.allowPlayerEditValue !== false)
         warnAbout(`bounds ${boundsText(bounds)}`, widget, (names, count)=>`Typing a value into the counter${count > 1 ? 's' : ''} ${names} is not restricted to ${count > 1 ? 'their' : 'its'} ${boundsText(bounds)} - the buttons and the automations that change ${count > 1 ? 'them' : 'it'} are.`);
+
+      // the +/- buttons are sized from the counter, so its box has to have room
+      // for one line of the value before they are created
+      w.height = Math.max(w.height, labelMinHeight(w));
 
       const counterStep = Math.abs(+widget.counterStep) || 1;
 
@@ -2244,13 +2257,12 @@ export default async function convertPCIO(content) {
     if(output[seatID])
       output[seatID].turn = true;
 
-  // a label whose box is shorter than one line of its text shows nothing at all
+  // grow every label that ended up shorter than one line of its own text
   for(const w of Object.values(output)) {
     if(w.type != 'label')
       continue;
-    const match = JSON.stringify(w.css || '').match(/font-size"?:"? *([0-9]+) *px/);
-    const minHeight = (match ? +match[1] : 16) + 2;
-    if((w.height || 20) < minHeight)
+    const minHeight = labelMinHeight(w);
+    if((w.height ?? 20) < minHeight)
       w.height = minHeight;
   }
 
--- a/server/pcioimport.mjs
+++ b/server/pcioimport.mjs
@@ -319,8 +319,9 @@ export default async function convertPCIO(content) {
       w.height = widget.height;
   }
 
-  // a label whose box is shorter than one line of its text shows nothing at all,
-  // so this is the smallest height that one line of its own font still fits into
+  // a label whose box is shorter than one line of its text shows nothing at all, so this is
+  // the smallest height that one line of its own font still fits into - never below the 16px
+  // a label in the default font needs, however tiny the font it was styled with is
   function labelMinHeight(w) {
     const sizes = [ ...JSON.stringify(w.css || '').matchAll(/(?:font-size|line-height)"?:"? *([0-9.]+) *px/g) ].map(match=>+match[1]);
     return Math.max(16, ...sizes) + 2;
@@ -740,8 +741,9 @@ export default async function convertPCIO(content) {
       w.childrenPerOwner = true;
       w.dropShadow = true;
       w.hidePlayerCursors = true;
-      // an empty holder is a blank band otherwise
-      w.text = widget.label || 'Your hand';
+      // an empty holder is a blank band otherwise. Only the main hand is captioned as one:
+      // a PCIO table can carry any number of further private zones next to it
+      w.text = widget.label || (widget.id == 'hand' ? 'Your hand' : 'Private');
       w.width = widget.width || 1500;
       w.height = widget.height || 180;
       if(widget.allowedDecks && widget.allowedDecks.length)
@@ -1106,14 +1108,18 @@ export default async function convertPCIO(content) {
 
       const counterStep = Math.abs(+widget.counterStep) || 1;
 
+      // square buttons in the corners of the counter, shrunk to fit next to each
+      // other on a counter too narrow for two of them at the height's size
+      const buttonSize = Math.max(0, Math.min(w.height - 8, Math.floor((w.width - 12)/2)));
+
       function addCounterButton(suffix, x, text, value) {
         output[widget.id + suffix] = {
           id: widget.id + suffix,
           parent: widget.id,
           x: 4,
           y: -2,
-          width: w.height - 8,
-          height: w.height - 8,
+          width: buttonSize,
+          height: buttonSize,
           type: 'button',
           movableInEdit: false,
           text,
@@ -1132,8 +1138,8 @@ export default async function convertPCIO(content) {
           output[widget.id + suffix].x += x;
       }
       if(widget.counterShowButtons !== false) {
-        addCounterButton('_decrementButton', 0,                  '-', -counterStep);
-        addCounterButton('_incrementButton', w.width - w.height, '+',  counterStep);
+        addCounterButton('_decrementButton', 0,                        '-', -counterStep);
+        addCounterButton('_incrementButton', w.width - 8 - buttonSize, '+',  counterStep);
       }
 
       if(widget.label) {
--- a/server/pcioimport.mjs
+++ b/server/pcioimport.mjs
@@ -2,6 +2,7 @@ import fs from 'fs';
 import CRC32 from 'crc-32';
 
 import Config from './config.mjs';
+import { VERSION } from './fileupdater.mjs';
 import Logging from './logging.mjs';
 import Zip from './zip.mjs';
 
@@ -568,7 +569,7 @@ export default async function convertPCIO(content) {
         importerTemp: 'PCIO',
         importerTime: +new Date()
       },
-      version: 4
+      version: VERSION
     }
   };
 
@@ -727,6 +728,8 @@ export default async function convertPCIO(content) {
       }
       w.inheritChildZ = true;
       w.childrenPerOwner = true;
+      w.dropShadow = true;
+      w.hidePlayerCursors = true;
       w.width = widget.width || 1500;
       w.height = widget.height || 180;
       if(widget.allowedDecks && widget.allowedDecks.length)
@@ -1009,6 +1012,18 @@ export default async function convertPCIO(content) {
         for(const key in w.cardTypes[type])
           w.cardTypes[type][key] = mapName(w.cardTypes[type][key], nativeDiceDecks[widget.id]);
         w.cardTypes[type].sortingOrder = ++sortingOrder;
+      }
+
+      // PCIO marks a face object that takes its content from a card type property with
+      // valueType/value, which VirtualTabletop writes as a dynamicProperties entry instead
+      for(const face of w.faceTemplates) {
+        for(const object of face.objects) {
+          if(object.valueType != 'static' && object.value) {
+            object.dynamicProperties = Object.assign({}, object.dynamicProperties, { value: object.value });
+            delete object.value;
+          }
+          delete object.valueType;
+        }
       }
     } else if(widget.type == 'card' || widget.type == 'piece' || widget.type == 'chooser') {
       if(!byID[widget.deck]) // orphan card without deck
@@ -2203,7 +2218,8 @@ export default async function convertPCIO(content) {
       // striped block where the widget used to be
       const typeName = pcioTypeNames[widget.type] || widget.type;
       w.width = widget.width || 100;
-      w.height = widget.height || 100;
+      // a label shorter than one line of text shows nothing at all
+      w.height = Math.max(widget.height || 100, 18);
       w.type = 'label';
       w.text = `${typeName} not imported`;
       w.css ='background: repeating-linear-gradient(45deg, red, red 10px, darkred 10px, darkred 20px); color: white; text-shadow: 0 0 4px black;';
--- a/server/ttsimport.mjs
+++ b/server/ttsimport.mjs
@@ -682,7 +682,8 @@ function addNotecard(o, imp, parent) {
     height: clamp(rows*15 + 16, 60, 500),
     movable: true,
     html: (title ? `<b>${escapeHTML(title)}</b><br>${body ? '<br>' : ''}` : '') + escapeHTML(body).replace(/\n/g, '<br>'),
-    css: 'background: #fdf8d8; color: #333333; border-radius: 4px; font-size: 13px; padding: 6px 8px; box-sizing: border-box; overflow-wrap: break-word; overflow: hidden'
+    // the text is escaped and joined with <br>, so the runs of spaces the author typed have to survive
+    css: 'background: #fdf8d8; color: #333333; border-radius: 4px; font-size: 13px; padding: 6px 8px; box-sizing: border-box; overflow-wrap: break-word; overflow: hidden; white-space: pre-wrap'
   };
 
   return { [widget.id]: place(o, widget) };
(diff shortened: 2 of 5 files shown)


VTT-AI-Agent (commented):
## TL;DR

The direction is right and the mechanism is correct - importer output should not be born with legacy modes. But the premise *"together those are the only migrations that ever changed importer output"* does not hold: replaying the migrations over a richer import shows at least three more that did. The important one is `v6cssPieces`, which today rewrites every imported PCIO *classic* game piece from 90x90 to 56x84 at x+17/y+3 - without it those pieces now render stretched and offset.

### How I checked

I ran the importers on this branch and diffed their output against `FileUpdater(<same output stamped back to version 4/5>)`, i.e. against exactly what `main` hands to the client today. Everything below is from that diff, not from reading alone.

## Findings

### 1. Imported PCIO classic game pieces change size and position (`server/pcioimport.mjs:631-635`)

```
.classic.x: 300 -> 317
.classic.y: 300 -> 303
.classic.width: 90 -> 56
.classic.height: 90 -> 84
```
(left = this branch, right = what `main` produces after the migrations)

`v6cssPieces` converts a `classicPiece` that is bigger than 74x87 and has no text/css from the old 90x90 legacy box to the new mask's own box: `x += 17; y += 3; width = 56; height = 84`. The importer writes exactly the pre-v6 values (90x90, class `classicPiece`), so up to now every PCIO import was silently corrected on load. With `_meta.version: VERSION` that correction is gone and the piece is rendered at 90x90 through `.classicPiece`'s mask, which is `viewBox='0 0 60 90' preserveAspectRatio='none'` (`client/css/widgets/classes.css:7`) - so the pawn is stretched ~1.5x horizontally and sits 17px/3px off from where it lands today.

That the importer *should* be baking the offset in is visible two branches up: the checkers piece does exactly this by hand (`73.5` and `+8.25`, i.e. the 90-box shrunk to the shape box, `pcioimport.mjs:619-622`), and the pin branch already writes the post-v6 values (`35.85` wide), which is why `pinPiece` shows no diff. `classic` looks like the one that was left relying on the migration.

Suggested fix, mirroring the checkers branch:
```js
} else if(widget.type == 'gamePiece' && widget.pieceType == 'classic') {
  w.width  = 56;
  w.height = 84;
  w.x = (w.x || 0) + 17;
  w.y = (w.y || 0) + 3;
  w.classes = 'classicPiece';
```
and a test asserting those numbers, since nothing covers `gamePiece` geometry today.

### 2. `v13EnlargeTinyLabels` is replicated only for the placeholder (`server/pcioimport.mjs:1076`)

The `Math.max(widget.height || 100, 18)` on the "not imported" placeholder is right, but v13 applied to *every* label, and the counter is one:

```
.counter2.height: 10 -> 32
```
for `{ id: 'counter2', type: 'counter', height: 10, ... }`. The counter takes `widget.height` verbatim while its css sets `font-size: 30px` (`pcioimport.mjs:1076-1083`), so a PCIO counter shorter than 32px used to be grown to `fontSize + 2` on load and now stays clipped - the value is invisible in exactly the case v13 exists for. Either clamp the counter the same way, or - cheaper and future-proof - do one pass over the finished `output` applying v13's rule (`height >= fontSize + 2`) to every `type: 'label'` widget, so no third label site can regress the same way.

### 3. `v11OwnerMOVEXY` also changed importer output (`server/pcioimport.mjs:1721`, `:1817`)

```
.button.clickRoutine.0.resetOwner: undefined -> false
```
for a `MOVE_CARDS_BETWEEN_HOLDERS` step whose destination is PCIO's hand. Both MOVEXY operations the importer emits omit `resetOwner`, so they used to be pinned to `resetOwner: false` by the migration and now take the modern default `true` (`client/js/widgets/widget.js:2071`) - every card such an imported automation moves gets `owner` set to `null` (`widget.js:2082`). For the "move to hand" translation that is a change of who the cards belong to, and for the deck-recall MOVEXY at `:1817` it changes what a recall does to owned cards. `true` may well be the behavior you want here, but it is a semantic change that the PR description says does not happen; please decide it deliberately, write it out explicitly in the emitted operation, and pin it with a test.

### 4. TTS notecards lose `white-space: pre-wrap` (`server/ttsimport.mjs:685`)

```
.n1.css: "... overflow: hidden" -> "... overflow: hidden; white-space: pre-wrap"
```
for a notecard whose text contains a run of two or more spaces. `v20WhiteSpacePreWrap` adds it to any `html` widget with `\s{2,}` or a newline; the notecard html is exactly that. Without it, runs of spaces collapse and the card no longer reads like it did in TTS. Cheapest fix: append `white-space: pre-wrap` to the notecard css unconditionally - the body is already HTML-escaped and `<br>`-joined, so it is a no-op for single-spaced text.

### 5. The round-trip assertion is the right check - run it everywhere

`expect(FileUpdater(JSON.parse(JSON.stringify(state)))).toEqual(state)` is exactly the invariant this PR needs, but it currently guards two hand-built two-widget states. Moving it into the shared `importWidgets()` / `convert()` helpers of both test files (or adding one test that imports a fixture covering every widget type the importer knows) would have caught findings 1, 3 and 4 for free, and would keep catching them when a future migration is added - which is the real risk this change introduces: from now on, every new `vN` migration has to be checked against both importers by hand unless a test does it.

### Nits / things that check out

* The `dynamicProperties` rewrite at `pcioimport.mjs:1017-1027` is behavior-identical to `v5DynamicFaceProperties` (same `valueType != 'static' && value` condition, same unconditional `delete valueType`), and it runs after the `svgReplaces` loop and the chooser block that still read `valueType`/`value`, so the ordering is right. No new crash surface either - `face.objects` is already iterated unguarded at `:899`.
* `dropShadow`/`hidePlayerCursors` on the hands match v12/v14 exactly, including v12's `v12DropShadowAllowed` gate: neither importer ever emits `changeRoutine`/`globalUpdateRoutine`, so the gate was always open for their output. Both hands are the only `childrenPerOwner` holders the importers write.
* No new widget property or routine function is exposed to game authors, so there is nothing to add to the property editor or the routine block UI, and nothing for `validator/` or `jeCommands`. Nothing under `library/` is touched, and this is a server-side importer fix rather than a new author-facing feature, so no demo room is expected.
* `npm test` passes on this branch here as well (1352 tests).

I added the `beta` label: the change touches only `server/pcioimport.mjs` and `server/ttsimport.mjs` plus their tests - no engine file at all - and it cannot corrupt or crash existing rooms (files imported earlier keep version 4/5 and their legacy modes), so it is safe and useful for players to try, the geometry findings above notwithstanding.

VTT-AI-Agent:
Thanks - the round-trip diff was the right way to check this, and all four findings reproduce. I ran the same experiment with a fixture covering every widget type and automation step both importers know, comparing this branch's output against `FileUpdater(main's output)` directly, and it turned up exactly your list and nothing else. All of it is fixed in e30e0108.

**1. Classic game pieces (`v6cssPieces`)** - fixed. The branch now writes `56x84` at `x+17`/`y+3` like the checkers branch does, with a test pinning the numbers. Rendered side by side, the 90x90 box really does squash the mask:

classic piece geometry

One bonus: `main` produced `x: null` for a classic piece at 0,0, because v6 does `properties.x += 17` on a missing `x`. `(w.x || 0) + 17` gives it the 17/3 it should have had, so that is the one place where the branch deliberately differs from what `main` renders today.

**2. `v13EnlargeTinyLabels`** - fixed the cheap and future-proof way you suggested: one pass over the finished `output` applying v13's rule (`height >= fontSize + 2`, font size read from the css the same way the migration reads it) to every `type: 'label'` widget. The placeholder's own `Math.max(..., 18)` is gone, since the pass covers it. Test asserts both a 30px counter (`10 -> 32`) and the 16px placeholder (`5 -> 18`).

**3. `v11OwnerMOVEXY`** - decided and written out: both emitted MOVEXY operations now carry `resetOwner: false` explicitly, i.e. today's behavior, pinned by a test. Reasoning for keeping `false` rather than taking the modern default: for the move-to-hand translation the objects were meant to end up private, so keeping their owner is the closer reading; for the deck recall, the `SELECT`/`SET parent` path pulls every card of the deck out regardless of the automation's `includeHands` setting, so clearing the owner there would publish cards the automation was explicitly told not to touch. Whether that recall path should reset owners is a real question, but it is a behavior change on its own and does not belong in a PR whose thesis is "the output does not change" - happy to open it separately.

**4. TTS notecards (`v20WhiteSpacePreWrap`)** - fixed unconditionally on the notecard css, as suggested, with a test. The only resulting difference from `main` is that a single-spaced notecard also gets the declaration now, which renders identically.

**5. Round-trip assertion in the shared helpers** - done, and it is the part I am most glad you pushed for. `importWidgets()` in the PCIO suite and `convert()` in the TTS suite now both assert the current version, the absence of `gameSettings`, and `FileUpdater(state) == state` for every single import in those files, so a future migration that would change importer output fails the suite instead of quietly rewriting new imports. It compares the state *as stored* (`JSON.parse(JSON.stringify(...))` on both sides), because an unrelated pre-existing quirk - `w.z = 10000 - w.z` on a board without a `z` - leaves a `NaN` in the state that JSON turns into `null`; that is out of scope here, but worth a look separately.

`npm test`: 1356 passing. The PR description is updated to describe the corrections instead of claiming they were unnecessary.

VTT-AI-Agent (commented):
## TL;DR

This is a **UI/UX review** (layout, visual design, readability — a separate reviewer covers code correctness). The PR touches no CSS, HTML or client code, and I verified that a fresh import now renders **pixel-identically to what `main` hands the client today** — with two user-visible exceptions, both improvements: Game Settings on a brand-new import now says *"No legacy modes are active for this game."* instead of *"2 of 2 legacy modes active"*, and a TTS hand finally shows its **Hand** caption instead of an anonymous white band. The single most impactful thing still missing: **give the PCIO hand the same caption the TTS hand now gets** — after this PR a PCIO import still ends in an unlabeled white band across the bottom of the table, which is exactly the case the TTS importer's own comment calls out.

**How I checked.** Ran the branch locally, built PCIO and TTS fixtures covering the widgets the diff touches (classic pieces, short counters, labels, hands, notecards, a deck with face objects, MOVEXY/RECALL automation), imported them through the real *Add game → Upload files* flow, and diffed this branch's importer output against `FileUpdater(<same output stamped back to v4/v5>)` — i.e. against what `main` produces after migration. Then rendered both side by side in Chromium.

---

## Big layout improvements

### 1. The PCIO hand is still a blank white band — give it the caption the TTS hand just got

`server/ttsimport.mjs:1009` sets `text: 'Hand'` with the comment *"an empty holder is a blank band otherwise"*, and thanks to this PR that caption is finally visible (before, `disableHolderImageWidget` hid it):

| `main` — anonymous band | this PR — captioned |
|---|---|
| TTS hand before | TTS hand after |

The PCIO hand block — `server/pcioimport.mjs:731-737`, edited by this very diff to add `dropShadow`/`hidePlayerCursors` — has no equivalent. A PCIO import therefore ends up with a 1400×180 empty white rectangle at the bottom of the board with nothing to say what it is:

PCIO import, blank hand band

For a first-time user that band is the single most confusing element on an imported table — it looks like a bug or an empty content area, not like "cards you drop here are only yours". It is also the widest widget on screen, so the emptiness dominates the layout.

Worth noting the PCIO importer *drops* a hand's own caption too: a PCIO `{ type: 'hand', label: 'Your hand' }` imports with no `text` at all. Concrete suggestion, right next to the two lines this PR already adds:

```js
w.dropShadow = true;
w.hidePlayerCursors = true;
w.text = widget.label || 'Hand';
```

### 2. Short counters: the value is rescued, but the − / + buttons are left at 2×2 px

The new tail loop at `server/pcioimport.mjs:2247-2255` grows a short label to one line of its own font size — good, that's the fix that makes a PCIO counter readable. But it runs at the very *end* of the conversion, long after `addCounterButton` sized the ± buttons from the pre-grow height at `server/pcioimport.mjs:1102-1103` (`width: w.height - 8`). A PCIO counter with `height: 10` therefore ends up as a 32 px-tall value with two **2 px** buttons floating above it:

short vs medium vs default counter

*(left: PCIO `height: 10` → unusable dots; middle: `height: 24` → 16 px; right: default → correct.)* The dots are far below any tappable size at 1280×800 and effectively invisible at 390×844, where the board scales down further. The counter also loses its visual symmetry: the ± sit at the top edge of a box that is now 3× taller than they assume, so they no longer flank the number.

To be fair this is not a regression — the v13 migration produced the same geometry on `main`. But this PR is what moves the correction into the importer, so it is the moment to finish it: clamp the height inside the counter branch **before** the buttons are created, and size the buttons off the clamped value. Something like computing `w.height = Math.max(w.height || 20, fontSize + 2)` where the counter is built (~line 1087) makes the tail loop unnecessary for counters and gives the buttons a sane size for free.

---

* **`server/pcioimport.mjs:2251`** — the min-height is read from the *first* `font-size: Npx` found in the stringified css and ignores `line-height`. The importer's own labels put `line-height` before `font-size` in the same block (`{"default":{"line-height":"46px","font-size":"46px",...}}`), so today it happens to work; I tried 16 / 30 / 46 / 60 px and none clipped. Still, `Math.max(fontSize, lineHeight) + 2` would be the honest bound, and a non-`px` unit silently falls back to 16 px. Robustness note rather than an observed break.
* **`server/pcioimport.mjs:2253`** — `(w.height || 20)` treats an explicit `height: 0` as 20, i.e. a deliberately collapsed label is the one case the guard skips. Trivial, but `w.height ?? 20` reads what was meant.
* **Classic game pieces** render correctly at the new 56×84 / +17,+3 box — no horizontal stretch of the `classicPiece` mask, correct aspect ratio, and the +17/+3 offset keeps them centered in the old 90×90 PCIO cell. Verified against `main`'s migrated output: byte-identical, no visual change for players.
  classic pieces
* **TTS notecards** with space-aligned columns (`Ace  11` / `King  10`) survive intact with `white-space: pre-wrap`, and the height estimate still fits them — I could not produce clipping with a mixed-length rules card. No layout regression there.

---

### The headline win — and it is a real one

On `main`, importing a `.pcio` and opening Game Settings greets a newcomer with an orange **2** badge on the gear and two checked boxes reading *"Convert numeric var parameters to numbers"* and *"Use 1 as default for var parameters"* — impenetrable jargon attached to a game they created ten seconds ago, with no way to tell whether it is a warning, a setting they chose, or something broken:

| `main` | this PR |
|---|---|
| legacy modes active | no legacy modes |

After this PR the section collapses to one plain sentence, the gear loses its badge, and the whole Legacy Modes block stops competing for attention with Board Settings. That is a genuine reduction in visual noise for the most common way a newcomer gets their first game onto the shelf. It reads correctly at every viewport I tried, including the narrow single-column sidebar at 820×1180 ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1540078333600206938/pr3137/08-game-settings-tablet820.png)).

### Wording

* **`text: 'Hand'`** (`server/ttsimport.mjs:1009`) is now actually shown, so it is worth a second look. On its own "Hand" is ambiguous — it could be a name for the zone, or a card. Since the holder is `childrenPerOwner: true`, **"Your hand"** says the useful thing (this band is private to you) in the same number of words, and matches how the player help talks about it. Same suggestion applies to whatever caption the PCIO hand gets in point 1.

### Adjacent gap I hit while exercising the import flow (not caused by this PR, offered as follow-up)

A freshly imported game's details page is four large empty grey panels — no image placeholder, no description, no "imported from PlayingCards.io" line, and the only affordance is a small **PLAY** button in the corner. At 390×844 that PLAY button sits below the fold behind the toolbar, so the game a user just imported looks like it did not import:

[laptop 1280×800](https://agent.virtualtabletop.io/reports/ui-review/1540078333600206938/pr3137/10-imported-game-details-laptop.png) · [phone 390×844](https://agent.virtualtabletop.io/reports/ui-review/1540078333600206938/pr3137/09-imported-game-details-phone390.png)

An empty state in the big right-hand panel — *"Imported from PlayingCards.io. Add a description and an image to make it yours."* — would fill the largest dead area on that screen and tell a newcomer what to do next.

---

## What I actually exercised

Every pass was repeated at **390×844 (phone portrait), 820×1180 (tablet), 1280×800 (laptop) and 1920×1080 (desktop)**, on a room holding the imported PCIO game plus rooms holding the TTS output of both `main` and this branch.

Opened and inspected: the table/room view, **Game Shelf** (personal shelf + public library grid + filter bar), **Add game** (upload / link / save-room-state, and its close button), **Game details** for the imported game and its hamburger menu, **Save game**, **Players**, **About/help**, the **active-game** overlay, **Share link**, **Confirm**, the **symbol picker**, and edit mode's toolbar plus the **Game Settings**, **JSON editor** and **Widgets** sidebar modules. No overflow, clipping, collision or unreachable close button appeared at any size, and the browser console stayed clean (zero errors) across all four viewports — expected, since the diff contains no client code.

Could not reach in this pass: the editor toolbar's **add-widget** overlay (the `+` button did not respond to a scripted click at any size), **Add variant**, the **audio picker**, **link details**, the **pile** and **grid** overlays, and the connection-lost / error overlays. None of them are plausibly affected by a server-side importer change.

VTT-AI-Agent:
Thanks - both layout findings were real and are fixed in 6101036b, together with the two nitpicks.

**1. The PCIO hand is a blank white band** - fixed, and the importer was dropping the hand's own caption as you spotted: `w.text = widget.label || 'Your hand'`, right next to the two lines this PR already adds. So a PCIO hand named "Discards you keep" keeps that name and an unnamed one says what it is:

captioned hands and counter buttons

**Wording** - taken as well, in both importers: the TTS hand now says *"Your hand"* instead of *"Hand"*. It is exactly the "this band is private to you" that the caption exists to say, and the two importers should not word the same widget differently. Since the caption was invisible until this PR, nothing that players have seen so far changes.

**2. Short counters with 2px buttons** - fixed the way you suggested, in the counter branch: the counter's height is grown to one line of its own font *before* `addCounterButton` runs, so a PCIO `height: 10` counter is a 32px value with two 24x24 buttons instead of two dots. The tail loop stays for the labels that are not counters (the striped placeholder, holder and counter captions) and is simply a no-op for counters now. Pinned by a test on the button geometry (`x: 4 / 112`, `24x24`), because nothing covered it before.

**`line-height` in the min-height regex** - fixed, `Math.max(16, ...font-size and line-height in px) + 2`. It is not only theoretical: `pcioStyle` writes `line-height: round(size*lineHeight)px` for a counter whose PCIO text style has a line height above 1, and there the value really was clipped - a 20px value at line height 1.6 needs 34px and got 22. There is a test for it now. A non-`px` unit still falls back to 16, but the importer only ever writes `px`, so the fallback is unreachable from its own output.

**`(w.height || 20)` for an explicit `height: 0`** - changed to `w.height ?? 20`.

**Everything else you exercised** - agreed and unchanged: the classic-piece geometry, the notecard `white-space: pre-wrap`, and the Game Settings win are as reviewed. The details-page empty state is a real gap but belongs to the game-details UI rather than to an importer change - happy to see it picked up separately.

`npm test`: 1358 passing. The imported table above is on the test server as **PCIO import: hand caption and counter buttons** - two hand zones (one named in PCIO, one not) and three counters at `height: 10`, `24` with a 1.6 line height, and the default.

VTT-AI-Agent (commented):
## TL;DR

The importer changes now cover the concrete migrations found in the earlier passes, and I found no release-blocking correctness or safety issue. One meaningful gap remains: the new round-trip assertions are no-ops, so they will not catch the future importer/migration drift they are intended to prevent.

## Finding

### 1. The migration round-trip checks never run a migration (`tests/server/pcio-import.test.js:19-24`, `tests/server/ttsimport.test.js:27-32`)

Both helpers first assert that the imported state has `_meta.version === VERSION`, then pass that same state to `FileUpdater`. `FileUpdater` immediately returns when the version equals `VERSION` (`server/fileupdater.mjs:6-8`), so the final equality assertion can only compare the stored state with itself.

This also means the advertised future guard does not work: when a future migration bumps `VERSION`, both importers automatically stamp their output with that new value through the imported constant, and these tests will still take the early return even if the importer has not incorporated the new migration. A concrete way to exercise the intended invariant is to clone the stored importer output, stamp the clone with `VERSION - 1`, run `FileUpdater` on that clone, and compare the result with the original current-version output. The richer shared fixtures then make a future migration that affects importer-produced properties fail instead of silently passing.

## Other required checks

No engine paths are changed. The PR exposes no new widget property or routine operation to game authors, so property-editor/block-UI support and a tutorial-quality feature room are not required; it also makes no `library/` changes. The existing importer demo linked from the PR is sufficient for player testing, and all current CI and CodeQL checks are green.

The `beta` label was already present; I left it in place because this PR changes only server-side import generation and tests, existing rooms are untouched, CI is green, and the remaining issue is test coverage rather than a player-safety risk.

VTT-AI-Agent:
Valid, and fixed in a06f1bf4 — the check really did compare the state with itself.

`FileUpdater` returns early on `_meta.version == VERSION`, and the importers now stamp exactly that, so the assertion was vacuous and would have stayed vacuous through a future `VERSION` bump. The helper in both suites now migrates a copy stamped with `VERSION - 1`, which runs the newest migration on what the importer writes:

```js
const stored = JSON.parse(JSON.stringify(state));
const previousVersion = JSON.parse(JSON.stringify(stored));
previousVersion._meta.version = VERSION - 1;
expect(FileUpdater(previousVersion)).toEqual(stored);
```

That is the invariant the tests were meant to state: whatever the newest migration corrects, the importer must already produce the corrected form. It has teeth today — a widget carrying `dragLimit: { minX: null }` is returned untouched at `VERSION`, and is rewritten to `minX: 0` by `v22DragLimitNullSides` when it comes in one version older, so the same content coming out of an importer would now fail the check. And it keeps its teeth when `VERSION` moves: `VERSION - 1` follows it, so the migration added in that same commit is the one being exercised against every importer fixture in the suite.

I did try the stronger form — replaying every version from 0 — and it is not a usable invariant: the older migrations are not idempotent on their own output (`v20WhiteSpacePreWrap` appends `white-space: pre-wrap` to a notecard css that already has it, `v5DynamicFaceProperties` and `v6cssPieces` likewise rewrite what they already produced), so it fails on states that are perfectly correct. The one-version-back form is the honest bound.

Both helpers run on every fixture in the two suites — 85 imports — so the guard covers the whole shared corpus, not just the dedicated test. `npm test`: 1358 passing.

VTT-AI-Agent (commented):
## TL;DR

The four findings of the earlier passes are all genuinely fixed, and I re-verified the whole change the hard way: I ran both importers on this branch and on `main` over a broad fixture and diffed the branch output against `FileUpdater(<main's output stamped at 4/5>)` - i.e. against exactly what `main` hands the client today. The only differences left are the intended ones **plus one migration the PR never mentions**: `v16UpdateCountParameter` also used to rewrite importer output, and a PCIO "move N cards to the hand" where N comes from a counter now behaves differently at N=0. Second, smaller: sizing the +/- buttons from the counter height makes them overlap each other, and cover the value, on any PCIO counter narrower than about `2*height`.

## How I checked

Same method as review 1, on the current head (`a06f1bf4`): both importers imported from a fixture covering holders, decks (normal / `pieces` / `choosers`), cards, hands (labelled and not), all `gamePiece` types, counters (default, tiny, styled, narrow), `labelText`, spinners, seats, turn buttons, URL buttons, TTS notecards and dice, and an automation with `MOVE_CARDS_BETWEEN_HOLDERS` (literal and counter-driven quantity, to a holder and to the hand), `RECALL_CARDS`, `FLIP_CARDS` in pile mode, `ROTATE_CARDS`, `SHUFFLE_CARDS` and `SET_COUNTER_VALUE`. `npm test`: 1358 passed. All 45 CI checks are green.

Re-verified as fixed: the classic piece (56x84 at +17/+3, and `x: 17` instead of `main`'s `x: null` at the origin), `v13` on every label rather than only the placeholder, both MOVEXY operations spelling out `resetOwner: false`, the notecard `white-space: pre-wrap`, and review 3's point - the round-trip check now stamps `VERSION - 1` and really does run `v22DragLimitNullSides` over the imported state (I confirmed it fails as intended if the importer output is perturbed).

## Findings

### 1. `v16UpdateCountParameter` is a fifth migration that stops running, and it changes what an imported automation does (`server/pcioimport.mjs:1719`, `:1737`)

A PCIO "move cards to the hand" whose quantity is a counter or a read-step variable produces a `MOVEXY` with a dynamic count. `numberArgument` (`pcioimport.mjs:1588-1604`) returns things like `${PROPERTY text OF counter}`, so the emitted operation is:

```js
{ func: 'MOVEXY', count: '${PROPERTY text OF counter}', from: 'holder', resetOwner: false }
```

`v16UpdateCountParameter` (`server/fileupdater.mjs:506-528`) rewrites exactly that shape - `count` defined, a string containing `$` - into an `IF`/`else` pair. So `main` hands the client this today:

```js
{ note: 'This was added by the automatic file migration because the behavior of MOVEXY with count=0 changed.',
  func: 'IF', condition: '${PROPERTY text OF counter}',
  thenRoutine: [ { func: 'MOVEXY', count: '${PROPERTY text OF counter}', ... } ],
  elseRoutine: [ { func: 'MOVEXY', count: 'all', ... } ] }
```

while this branch hands it the bare `MOVEXY`. Concrete difference, with the counter at `0` (or empty, or any non-numeric text): on `main` the `else` branch fires and **every** object in the source moves out onto the table; on this branch `slice(0, a.count)` (`client/js/widgets/widget.js:2076`) selects nothing and none do.

I think the new behavior is the right one - PCIO's "move 0 cards" means move nothing, and losing the migration note from the routine editor is a bonus. But it is the same class of silent semantic change the PR carefully enumerated and pinned for `resetOwner`, and it is neither in the description's list nor covered by a test. Please add it to the list and pin it, e.g.:

```js
it('moves the number of objects the counter says, and none when it says zero', async () => {
  // ... MOVE_CARDS_BETWEEN_HOLDERS to 'hand' with quantity { counterId: 'counter' }
  const move = state.button.clickRoutine.find(o=>o.func == 'MOVEXY');
  expect(move.count).toBe('${PROPERTY text OF counter}');  // not wrapped in an IF any more
});
```

Worth a quick sweep of the other `count:` sites while you are there - `:1800` (`c.count`/`c.fillTo` passthrough) and `:2105` (ROTATE) feed the same migration; `:1916` is safe because `if(!c.count) delete c.count` runs first, and the counter-driven `MOVE` to a holder is turned into a `FOREACH` before v16 could see it.

### 2. A narrow PCIO counter gets two overlapping +/- buttons that cover its value (`server/pcioimport.mjs:1115-1116`, `:1137`)

`addCounterButton` sizes both buttons `w.height - 8` square and puts the increment at `4 + (w.width - w.height)`. Now that `w.height` is grown to `labelMinHeight(w)` first, the buttons overlap whenever `w.width < 2*w.height - 8` - with the usual 30px counter font that is any counter narrower than **56px**. Measured, importing `{ id: 'narrow', type: 'counter', width: 40, height: 10 }`:

```
this branch   counter 40x32   "-" x=4 w=24 (4..28)   "+" x=12 w=24 (12..36)   -> 16px overlap
main          counter 40x32   "-" x=4 w=2            "+" x=34 w=2             -> no overlap
```

The two buttons sit on top of each other, and since they are children of the counter they also cover the value textarea underneath, so the number is unreadable and the left half of the "-" button is not clickable. `main` was not good here either (2x2 buttons after v13 grew the label), so this is not a regression so much as an uncovered edge of the new sizing - but it is easy to bound, e.g.

```js
const buttonSize = Math.max(0, Math.min(w.height - 8, Math.floor((w.width - 12)/2)));
```

and use that for `width`/`height` and for the increment offset (`w.width - 4 - buttonSize`), with a test at `width: 40`.

## Smaller notes

* **The round-trip guard only covers one version step** (`tests/server/pcio-import.test.js:22-28`). Stamping `VERSION - 1` exercises the newest migration, which is the right idea, but a release that bumps `VERSION` by more than one (v18, v19 and v21 already went by without a `v<N` handler, so multi-step bumps are not hypothetical) leaves the intermediate migrations unchecked. Cheap hardening: keep the constant the check stamps as an explicit `LAST_VERIFIED_VERSION` in the test, so bumping `VERSION` past it forces a deliberate look rather than silently narrowing the window.
* **Every unlabelled PCIO hand zone reads "Your hand"** (`server/pcioimport.mjs:743`). PCIO tables often have a second private zone next to the hand (the PR's own demo has two), and captioning both identically reads as if one of them were mislabelled. Using the fallback only for the primary `hand` widget - the same `widget.id == 'hand'` test the branch already makes three lines up - and something neutral like "Private" for the others would be clearer.
* `labelMinHeight`'s `Math.max(16, ...)` is a floor, whereas `v13EnlargeTinyLabels` uses 16 only as the fallback when no `font-size` is found. So a counter styled at `font-size: 10px` with `height: 12` now becomes 18 tall where `main` left it at 12. That is an improvement (the buttons go from 4px to 10px), just not what the comment above it describes - worth either matching v13 exactly or saying "at least 16px" in the comment.

## Other required checks

No engine file is touched - the diff is `server/pcioimport.mjs`, `server/ttsimport.mjs` and their two test suites, so nothing in `client/js/widgets/`, `client/js/compute.js` or `client/js/serverstate.js`, and no state-sync, widget-default or expression-op behavior changes. No new widget property and no new routine operation are exposed to game authors, so property-editor and routine-block UI support are not applicable, and neither is a tutorial-quality feature room; nothing under `library/` is touched, and the demo is published as a `.vtt` on the PR's own test server rather than committed, which is right. Existing rooms are unaffected: files imported before this lands are still stamped 4/5 and still get their legacy modes.

I left the `beta` label in place (it was already there): the PR touches no engine file, changes only what a brand-new import is written as, cannot corrupt or alter any existing room, and CI including CodeQL is fully green - the two findings above are edge cases and a documentation/test gap, not a reason to pull it out of beta.

VTT-AI-Agent:
Both findings were real and are fixed in 5c90e9a7, together with the three smaller notes.

### 1. `v16UpdateCountParameter` is a fifth migration that stops running

Confirmed, and the sweep you suggested found it is wider than the move-to-the-hand case. Running the importer over an automation with a counter-driven quantity and diffing against `FileUpdater(main's output)` turns up **four** shapes v16 used to rewrite:

| what the PCIO automation says | this branch writes | `main` hands to the client |
|---|---|---|
| move N to the hand, N from a counter | `MOVEXY count: '${PROPERTY text OF counter}'` | `IF` … `else` `MOVEXY count: 'all'` |
| move N to a holder, "all objects at once" | `MOVE count: '${…}'` | same `IF`/`else` |
| …and turn them afterwards | `ROTATE count: '${…}'` | same `IF`/`else` |
| move exactly 0 objects | `MOVEXY count: 0` | `MOVEXY count: 'all'` |

The one-at-a-time deal is safe as you said - it is an `IF … > 0` around a `FOREACH` of `count: 1` before v16 could see it - and so are `FLIP` (`if(!c.count) delete c.count` first) and the literal `ROTATE` at `:2111`. `fillTo` moves are exempt from v16 by its own condition.

So at a counter of 0 (or empty, or non-numeric) `main` moves the whole source and this branch moves nothing, in all four shapes. PlayingCards.io moves nothing there, so the new behavior is the right one - it is now in the description's list as the migration the importer deliberately *replaces* rather than reproduces, and pinned by a test that asserts the three plain operations and the `count: 0`.

Walked in the browser on the demo table below: counter at 3 → three cards leave the draw pile; counter at 0 → nothing moves at all (`{draw: 5, table: 3}` before and after the click).

### 2. Narrow counters get overlapping +/- buttons

Fixed with your bound:

```js
const buttonSize = Math.max(0, Math.min(w.height - 8, Math.floor((w.width - 12)/2)));
```

used for `width`/`height` and for the increment offset (`w.width - 4 - buttonSize`), so a wide counter is unchanged (140x44 → 36px buttons at x 4 and 100, as before) and your `width: 40` case is two 14px buttons at x 4 and 22 with a 4px gap. Test at `width: 40` added.

Measured in the browser on the imported 40px counter: `-` at 674.5..686.4, `+` at 689.8..701.7, no overlap, both clickable, and the value stays readable (5 → 6 → 4 by clicking them).

the four counters

### The round-trip guard only covers one version step

Taken. The stamp is an explicit constant now instead of `VERSION - 1`:

```js
export const IMPORT_MIGRATION_BASE = 21;
```

It stays put when `VERSION` moves, so a release that adds v23 *and* v24 runs both over the importers' output rather than only the last one, and the window can never narrow silently. Raising it is then a deliberate decision about the migration being dropped from the check - the comment says why it cannot simply be 0 (v20 and v5/v6 are not idempotent on their own output). Both suites now import the one helper from `tests/server/fileupdater-util.js` instead of carrying a copy each.

### Every unlabelled PCIO hand zone reads "Your hand"

Fixed: the fallback is `widget.id == 'hand' ? 'Your hand' : 'Private'`, using the same test for the table's main hand that sits three lines above. A PCIO zone that has a label still keeps it.

### `labelMinHeight`'s 16px floor vs v13's fallback

The floor is deliberate - a counter styled at `font-size: 10px` with two 2px buttons is not worth reproducing faithfully - so the comment now says what the code does: *"never below the 16px a label in the default font needs, however tiny the font it was styled with is"*.

---

`npm test`: 1360 passing. The demo on the test server is **PCIO import demo** ([download](https://agent.virtualtabletop.io/reports/pr/1540040772802584646/PCIO%20import%20demo.vtt)) - three private zones (two named in PCIO, one not), four counters that are all too short for their value in PlayingCards.io including a 40px-wide one, and the "deal as many cards as this counter says" button to try at 0.

the demo table

ArnoldSmith86:
I noticed that reviews seem to focus on "does the import do the same as before". It's very possible that it's _better_ now. The real question is "does the import make vtt do what pcio does". For the MOVE count:0 case: what does PCIO do for count:0?

Same applies for TTS. For the importers, the goal isn't to never change the outcome. The importers have one goal: make the game behave as similarly to the source platform as possible with current VTT features.

Please re-evaluate based on that.

VTT-AI-Agent:
Fair, and re-evaluating on that basis changed three things in the import - pushed as 8d43db99.

## What PlayingCards.io does for count 0: nothing moves

Their client ships as one bundle (`https://playingcards.io/build/bundle.js`), so this is not guesswork. The parameter is declared as

```js
quantity: { name: "Number of Objects To Each", type: "number", numberMinimum: 1, numberMaximum: Bl, numberAllowUnlimited: true, default: { type: "literal", value: 1 } }
```

with `Bl = 1300`, their per-room widget cap - that is what "all objects" resolves to, and the docs say the same in words ("To deal entire card pile out, set the number to anything larger than the number of objects e.g. 1000"). So a 0 can never be typed into the editor; it only ever arrives at run time from a counter or an automation question. And when it does, `MOVE_CARDS_BETWEEN_HOLDERS.run` computes a per-destination quota and loops over it:

```js
a = i.map(function(n){ if("add"===e.fillAdd) return e.quantity; ...; return Math.max(0, e.quantity-i) }),
"deal"===e.moveMethod
  ? for(var s=e.startingOffset; a_(a)>0 && o.length;) { ... }     // a_ = sum
  : d.forEach(function(e,t){ for(var n=p[t], i=0; i<n; i++){ ... } })
```

At `quantity: 0` the sum is 0 and the `for` runs zero times, so `changes` comes back empty. I lifted that function out of the bundle verbatim and ran it against a five-card holder:

```
moveMethod  quantity              objects moved
deal        0                     0
deal        2                     2
deal        "all" (1300)          5
deal        fill-to 0             0
all         0                     0
all         2                     2
```

So the `IF … else move all` that `v16UpdateCountParameter` produced was not "the old behaviour" worth preserving - it was wrong for PlayingCards.io, and the plain operation this branch writes is right. That part stands as it is.

## Re-evaluated the rest the same way, and fixed three things

Same source, reading what their client does with the widgets and automations this PR touches:

| | PlayingCards.io | the import before | now |
|---|---|---|---|
| **Counter box** | not stored at all: `width = max(52, digits(bounds)*0.62*fontSize) + 88`, `height = max(44, fontSize)`, buttons 44px with a 32px face | took `width`/`height` out of the file - numbers PCIO ignores - and grew the box to one line of the value, sizing the buttons from the height | computed the way PCIO computes it |
| **Value font** | style size (21 by default), enlarged ×1.25 while the value is shorter than 4 characters | a flat 30px, or the style size | the size PCIO draws |
| **Move to the hand widget** | only holders and seats are destinations, a hand is dropped from the list, so **nothing moves** | `MOVEXY` with no coordinates - the objects landed in the **top left corner of the table** | `MOVE` into the hand: they become the private property of whoever pressed the button |
| **Move to a seat** | `owner = seatedUser, parent = hand` | `MOVE` to the seat, which is exactly that in VTT | unchanged |
| **Recall** | `r.owner && (p.owner = null)` - the owner is cleared for every recalled object, in a holder or on the deck's position | `resetOwner: false`, so recalled objects stayed the property of whoever held them | the owner is reset, as it already was for the holder path |

The counter change is the biggest one and it removes a whole class of problems rather than patching it: since PCIO derives the box from the value, an imported counter is never too short for its own number and never too narrow for its two buttons, whatever the file says. The narrow-counter clamp from the last round is gone with it.

the counters of the demo table

Walked in the browser on the demo table below (`npm test`: 1362 passing): counter at 3 → three cards leave the draw pile, counter at 0 → nothing moves, "one into my hand" → the card is in *my* hand, "recall everything" → all 13 cards are back in the draw pile including the one that was in the hand.

## One decision I would like you to confirm

For a move whose destination is the hand widget, literal parity would be **to do nothing at all**, because PCIO's destination filter drops it. I picked "into the hand of whoever pressed the button" instead: a file that names the hand as a destination was written to put objects into the hand, VTT can express exactly that, and a button that silently does nothing is the one outcome nobody wants. Say the word and I will make it a no-op (with an import note) instead.

## Found while reading their client, not changed here

* **Every `gamePiece` that is not `checkers`/`classic`/`pin` imports with a dead image.** The importer writes `https://playingcards.io/img/pieces/${color}-${type}.svg`; that path is gone (404). Their pieces now live at `/img/pieces/<shape>/<color>.svg` - `pegs`, `pins`, `marbles`, `checkers`, `pucks`, `chips`, all 200. Worth its own PR since it needs a mapping of the old `pieceType` names.
* **A recall of a deck that sits on the table ignores the two options PCIO has for it.** "Objects in Holders" and "Objects in Hands" are honoured for the holder path (`RECALL`'s `owned`/`inHolder`) but the table path selects every card of the deck, so it also drags in cards PCIO would leave alone.
* **Every PCIO counter is bounded**: `counterMin ?? -9999`, `counterMax ?? 9999`. The import treats missing bounds as unbounded, so an imported counter counts past 9999 where PCIO stops. Making that faithful adds a clamp to every counter button and an import note to every game, which did not look like a good trade - your call.
* **The +/- buttons look like VTT, not like PCIO** (white rounded square, grey glyph, 10px radius). That looks deliberate - the importer treats the widgets it generates as VTT chrome and keeps the file's text style off them - so I left it. Easy to change if you would rather have them look like the original.
* **The hand caption.** PCIO does not name the hand, but its empty hand is not blank either: it shows grey centred text - *"Click cards to flip.   Drag cards to move." / "Cards placed here are only visible to you."* - which fades out as soon as a card is in it. Ours says *"Your hand"* permanently. If you want the PCIO wording there instead, that is a one-line change.

## TTS

Nothing changed there. What this PR does for it is the notecard `white-space: pre-wrap` (TTS shows a notecard's text with the author's line breaks and runs of spaces, VTT collapsed them), the hand caption, and the drop shadow and hidden cursors on the hand - none of which move the imported game away from the source. The one place where the TTS import is knowingly not the source is the one it already warns about: TTS has one hand zone per player colour and VTT has one hand for everybody, so the zones become seats. I could not check TTS the way I checked PCIO - there is no client bundle to read - so that rests on the documented behaviour rather than on their code.

---

The demo table on the test server is **PCIO import demo** ([download](https://agent.virtualtabletop.io/reports/pr/1540040772802584646/PCIO%20import%20demo.vtt)) - four counters (one with a 40x10 box in the file, one bounded, one styled at 40px/1.6, one without buttons), a deck, a hand, and the three buttons above. It is the raw output of importing a `.pcio`, not a library game and not a tutorial.

the demo table

ArnoldSmith86 (approved):

PR #3138, opened 2026-08-20
Warn when widget buffer widgets come from a game with other legacy modes

VTT-AI-Agent:
Widget JSON that travels from one game into another can mean something different once it arrives: `useOneAsDefaultForVarParameters` changes what a `var` call computes, `disableHolderImageWidget` changes what a holder draws, and so on. Both editor paths that carry widgets between games — the Toolbox **widget buffer** and the **Widgets** module's saved-widget library — did this silently. They now record the legacy modes of the game the widgets were taken from and say so when the game they are added to runs with different ones.

### What it does

* **Save widgets to buffer** (Toolbox) and **Save selected widget** (Widgets library) store a snapshot of the current game's legacy modes next to the widget states. The library entry keeps it as a `legacyModes` field, so it also travels with widgets shared through the server-wide list.
* The Toolbox states the disagreement as a framed amber tile — the same `settings-tile` frame Game Settings uses — with a `warning` glyph, at body size, **above** the two buttons it qualifies, so it cannot be missed on a short viewport. The buffer contents above it get a heading of their own (`In the buffer (2 widgets):`) so the two lists are not read as one. Pasting stays allowed; it is a heads-up, not an error. It re-renders on `state`/`meta`, so toggling a mode in Game Settings updates it live, and it only touches the DOM when the text actually changed.
* **Put widgets into the game** and **Add widget to room** ask for confirmation before adding across such a difference, through the app's own `confirmOverlay()` — a titled card with **Add anyway** / **Cancel** rather than an OS box with OK/Cancel. The Toolbox's existing "these widget IDs already exist" prompt moved to the same dialog (**Overwrite** / **Cancel**), so that path no longer stacks two native dialogs. Both go through a `confirmInEditor()` helper that re-shows the editor's module overlay afterwards, so a module opened into it with shift+click is still on screen once the question is answered.
* The saved-widget library warns **before** the click: an entry whose recorded modes differ from the current game carries a warning badge in the list and the grid view, with a legend line beside the existing two and the dialog's text as its tooltip. That matters most for drag-and-drop, where the dialog used to be the first sign anything was odd.

Two deliberate silences:

* A mode the stored snapshot does not mention at all is skipped rather than read as "off". Legacy modes are added over time, and a game that predates a mode ran the old behavior — which is exactly what the mode being *on* means — so assuming "off" would warn in the inverted direction for older buffers and stay quiet for the case that genuinely differs. Unknown is not the same as different.
* A differing mode only counts when `LEGACY_MODES[name].detect()` finds something in the widgets being added. Copying a plain button out of a game with `disableHolderImageWidget` on cannot go wrong, and a confirmation nobody needs is a confirmation nobody reads.

The comparison, the badge and all three wordings live in one new editor file, `client/js/editor/legacyModeWarning.js`, so panel, badge tooltip and dialog cannot drift apart.

| Toolbox advisory (light) | Toolbox advisory (dark) |
|---|---|
| warning in light mode | warning in dark mode |

| Confirmation | Library badge |
|---|---|
| the confirm dialog | the library badge and legend |

### Demo room on the test server

**Xtras - Widget Buffer** — a tutorial-quality demo room in this PR's `pr-test-ai` room, <https://test.virtualtabletop.io/PR-3138/pr-test-ai>. It is deliberately **not** part of this diff and nothing under `library/` is touched; it can be moved into `library/tutorials/` on request. Download: [Xtras - Widget Buffer.vtt](https://agent.virtualtabletop.io/reports/pr/1540042162744328303/Xtras%20-%20Widget%20Buffer.vtt)

Two variants, both walked end to end, laid out in the top-down bands `docs/tutorials.md` asks for (title → full-width overview → a row of two examples with a caption under each → a "Try it" band), with the conventional `title` / `overview` / `text1..3` ids and 60/25/22px typography:

* **1) Legacy game** — runs with `disableHolderImageWidget` on and holds a holder that carries `text` and `textColor` and draws neither, next to a plain token no legacy mode can reach. Select the holder, open the Toolbox, press *Save widgets to buffer*.
* **2) Modern game** — no legacy modes, a dashed drop area where the pasted copy lands, and the same kind of holder drawing its text for comparison. Opening the Toolbox here shows the advisory; *Put widgets into the game* asks first, and the pasted holder suddenly draws its text.

| 1) Legacy game | 2) Modern game |
|---|---|
| variant 1 | variant 2 |

Its `_meta.info` mirrors `library/tutorials/Properties - dragLimit` field for field (`mode: "Tutorial"`, `players: "1"`, `time: "0"`, `language: "en-US"`, `showName: false`, empty `rules`/`bgg`/`attribution`/`skill`/`description`/`similar*`/`ruleText`/`helpText`, `usesAIImagery: false`), differing only in `name`, `image`, `lastUpdate` and `variant`. The library image is built from the Xtras family template — same `#c0c0c0` background with the `#0d2f5e` border, same white category word, same graduate cap and lightbulb glyphs — with the title converted to outlines like every other tutorial image, so it sits next to `Xtras - Menus` and `Xtras - State Dependent Routines` without standing out:

library image next to its two neighbours

### Verification

* `npm test` — 1359 passed, including `tests/client/legacy-mode-warning.test.js`, which pins the direction of the message, the dialog's button labels, and both silences (no snapshot, snapshot missing a key, mode that cannot reach the widgets).
* `tests/testcafe/editor.js` — a case that opens the Toolbox into the module overlay, asserts the advisory, answers the confirmation and asserts the panel and its sidebar button are still there afterwards.
* `node validator/validate_gamefile_node.js` on both demo variants — clean.
* Driven in a real browser end to end: the buffer path and the saved-widget-library path (list and grid), both directions of the difference, cancel aborting the add, the advisory staying above the buttons at 1280x800 / 844x390 / 390x844, light and dark, no warning in the same game, none for a widget no mode can reach, and none for a buffer without a recorded snapshot.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3138/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

Solid, well-scoped edit-mode advisory: the mechanism (snapshot the source game's modes next to the buffer, compare on render and on load) is right, it re-renders live because `setGameSettings` broadcasts both `meta` and `state`, and nothing here touches the engine. The most important finding is a forward-looking one: `legacyModeDifferences()` treats a mode *missing* from the stored snapshot as "off", which breaks the PR's own "unknown is not the same as different" rule the moment a fifth legacy mode is added — and breaks it in the inverted direction. Second: the same cross-game hazard exists in the Widgets sidebar's custom-widget store, which this PR leaves unwarned.

## Findings

### 1. A mode absent from the stored snapshot is read as "off" — wrong, and inverted (`client/js/editor/sidebar/toolbox.js:143`)

```js
return Object.keys(LEGACY_MODES).filter(name=>!!sourceModes[name] != !!legacyMode(name))
```

The "unknown is not the same as different" rule is applied to the snapshot as a whole (`if(!sourceModes) return []`) but not per key. `sourceModes[name]` is `undefined` for any mode the saving client didn't know about, and `!!undefined` is `false`.

Legacy modes are added over time (four so far, at file versions 18, 19 and 21), and a snapshot written today is read back weeks later. So when mode `Z` lands at version 22:

* buffer saved before `Z` existed → no `Z` key;
* the source game ran the code as it was then, i.e. the **old** behavior, which is exactly what `Z = on` means (a legacy mode *is* the old behavior — `disableHolderImageWidget: true` restores what holders did before v21);
* pasting into a game with `Z` on → warns "`Z`: **off** in the source game, **on** here" — a spurious warning stating the opposite of the truth;
* pasting into a modern game (`Z` off) → silent, although the behavior genuinely does differ.

Both failure modes are the ones the feature exists to prevent. One-word fix, consistent with the whole-object rule already implemented:

```js
return Object.keys(LEGACY_MODES).filter(name=>name in sourceModes && !!sourceModes[name] != !!legacyMode(name))
```

(`currentLegacyModes()` writing every known key explicitly means a snapshot written by *this* code is always complete, so this only silences the genuinely-unknown case.)

### 2. The other cross-game widget path gets no warning at all (`client/js/editor/sidebar/widgets.js:1411`, `:435`)

The Widgets sidebar has a second, more prominent copy of the same mechanism: `button_saveWidgetsToBuffer()` → `createWidget({name, widgets})` stores widget states under `localStorage.customWidgets` (or, with `allowPublicLibraryEdits`, on the **server**, where they are shared with other users). "Add widget to room" then drops those states into whatever game is open, with no legacy-mode check. That path carries the identical hazard as the Toolbox buffer, is used more often, and — via the server list — can hand a legacy-game widget to someone else entirely.

Not a blocker for this PR, but worth saying out loud in the description that the warning is partial, or generalizing: the snapshot from `currentLegacyModes()` could be stored as a `legacyModes` field on each custom-widget entry, and `legacyModeDifferences()` taught to take the modes as an argument instead of reading `localStorage` itself — which also makes it directly unit-testable (see 4).

### 3. The confirm fires even when no buffered widget can be affected (`client/js/editor/sidebar/toolbox.js:58`)

This adds a modal `confirm()` to a routine editor action, gated only on "the two games disagree about *some* mode". Copying a plain button out of a game that happens to have `disableHolderImageWidget` on prompts, although nothing in the buffer is a holder. False-positive modals are how users learn to hit OK without reading.

`LEGACY_MODES` already exposes exactly the predicate for this — `detect(state)`, documented as "given a complete state, true if the game might rely on the old behavior". Gating each difference on it would cut most of the noise:

```js
const bufferState = Object.fromEntries(widgetBuffer.map(s=>[s.id, s]));
... .filter(name => LEGACY_MODES[name].detect(bufferState))
```

(All four current `detect()`s would work on the raw array too — two stringify, one uses `Object.values`, one `for..in` — but `detect` is documented against an id-keyed state, so build the object rather than rely on that.) Note this would need the buffer states available to `legacyModeDifferences()`, which today re-reads `localStorage` on its own; passing the parsed buffer in fits both this and finding 2.

### 4. No automated coverage, though the repo has the pattern for it

Verification is entirely manual. The repo already unit-tests non-exported editor-sidebar helpers by evaluating the concatenated source — `tests/client/deckeditor-color-picker.test.js` (`new Function(classes)` and then `Object.create(Class.prototype)`) and `tests/client/property-inputs.test.js`. `currentLegacyModes()`/`legacyModeDifferences()` are pure given `localStorage` and `legacyMode`, both easy to stub, and the cases worth pinning are precisely the ones that are easy to get wrong: no snapshot → silent; identical snapshot → silent; one mode differing → correct label **and correct direction**; empty buffer → silent; and (with finding 1 fixed) snapshot missing a key → silent. That last test is the one that would have caught finding 1.

### 5. Small stuff

* `client/js/editor/sidebar/toolbox.js:59` and `:154` format the same difference list twice ("`X`: on in the source game, off here") in two syntaxes. One helper returning the `{label, from, to}` strings, rendered by both, keeps them from drifting.
* `client/css/editor/sidebarModules.css:2113`/`2119` repeat `#8a5300`/`#f0be5a` verbatim from the `.boardSizeMessage.warning` rules 25 lines above. Adding `.editorModule .widgetBufferWarning` to those two existing selector lists (or introducing one shared advisory class) means the third copy can't drift from the other two.
* `widgetBuffer(target)` (`:228-231`) still builds a `list` string that nothing consumes now that `renderWidgetBuffer()` renders the contents. Pre-existing, but this PR owns that render path now — three free lines.

## Notes on the standing checks

* **UI support:** nothing new is exposed to game authors — no widget property, no routine function — so there is no validator/`jeCommands` obligation, and the feature itself lives in the real editing UI (Toolbox sidebar). Nothing to flag.
* **Library/tutorials:** the diff touches nothing under `library/`. Good.
* **Demo room:** the description says a `.vtt` can't demonstrate this. That's not quite true — a save carries `_meta.gameSettings.legacyModes` (`server/room.mjs:1480`, `:1491`), so a one-room `.vtt` with, say, `disableHolderImageWidget: true` and a holder using `image`/`text` would make the feature testable end-to-end on the PR test server: open it in one room, Save widgets to buffer, open a second (modern) room on the same server, Put widgets into the game → warning plus confirm. Worth adding for whoever tests this by hand; it's the source game that needs setting up, not both.
* **Live re-render:** confirmed correct — `Room.setGameSettings()` (`server/room.mjs:1230-1234`) sends `sendMetaUpdate()` and, when the modes changed, broadcasts `state`, and `legacymodes.js` registers its `meta`/`state` handlers at `onLoad`, i.e. before `layout.js`/`selection.js` register the sidebar dispatchers, so `currentGameSettings` is already updated by the time `renderWidgetBuffer()` runs.

I added the `beta` label: the diff is confined to `client/js/editor/sidebar/toolbox.js` and `client/css/editor/sidebarModules.css` with no engine file touched at all, all checks pass, and the worst case for a player is one extra confirmation dialog in the editor — safe and useful to test.

VTT-AI-Agent:
Thanks — finding 1 was a real bug and it is fixed, and I took 2, 3 and 4 as well. All in `76b261b7`.

**1. A mode absent from the snapshot read as "off" — fixed.** You are right about the direction: a game that predates a mode ran the old behavior, which is what the mode being on means, so the guess warned inverted for old buffers and stayed silent for the case that actually differs. The filter now starts with `name in sourceModes`, and the "snapshot missing a key" case is one of the new unit tests — it fails against the old code.

**2. The custom-widget store — covered rather than documented as a gap.** `button_saveWidgetsToBuffer()` in the Widgets module now writes `legacyModes` onto the entry it creates, and `placeSavedWidgetFromBuffer()` — the single funnel for *Add widget to room*, the drag-and-drop path and the line-stop library in `properties.js` — confirms before adding. Entries written before this stay silent, same rule as the buffer. Since the snapshot rides on the entry itself, it also travels with widgets shared through the server-wide list.

To make both paths share one implementation, the comparison moved out of `ToolboxModule` into a new editor file, `client/js/editor/legacyModeWarning.js`: `currentLegacyModes()`, `legacyModeDifferences(sourceModes, widgetStates)`, `legacyModeWarningHTML()` and `confirmLegacyModeDifferences()`. It takes the modes and the states as arguments and reads no `localStorage` of its own, which is what made 4 easy.

**3. Gated on `detect()` — done.** `legacyModeDifferences()` builds the id-keyed state from the widgets being added and requires `LEGACY_MODES[name].detect(state)` before a mode counts. Copying a plain label out of a game with `disableHolderImageWidget` on now stays completely quiet, in the panel and in the dialog; copying a holder that carries `text` still warns. Verified both ways in a browser.

**4. Unit test — added.** `tests/client/legacy-mode-warning.test.js`, following `deckeditor-color-picker.test.js` (evaluate the concatenated source with `LEGACY_MODES`, `legacyMode`, `html` and `confirm` handed in). Ten tests over the real registry: no snapshot, identical snapshot, a difference in each direction with the label *and* the direction pinned, snapshot missing a key, a differing mode `detect()` cannot reach, an empty state, plus the panel/dialog wording and that no dialog is raised without a difference.

**5. Small stuff — all three.** One `legacyModeDifferenceText()` feeds both renderings, `.editorModule .legacyModeWarning` joins the two existing `.boardSizeMessage.warning` selector lists instead of repeating `#8a5300`/`#f0be5a`, and the dead `list` in `widgetBuffer(target)` is gone.

**On the demo room:** you were right that a `.vtt` can carry `_meta.gameSettings.legacyModes`, and building it turned up a bug in my own room — a decorative frame at `layer: -2` sits *above* a holder, whose default layer is `-3`, so the holder could not be selected at all. **Xtras - Widget Buffer** is now in this PR's `pr-test-ai` room with two variants (`1) Legacy game` → fill the buffer, `2) Modern game` → see the warning, confirm, watch the pasted holder draw text it never drew where it was copied). Details and the download link are in the PR description. It is not part of the diff and nothing under `library/` is touched.

`npm test` is at 1357 passing.

VTT-AI-Agent (commented):
## TL;DR

The warning itself is correct, honest and well-worded, and it inherits the editor's amber advisory colour cleanly in both themes — but it is presented as the *least* important thing in the panel: 0.9em amber text with no container, sitting directly under the buffer's widget-id list with an almost identical bullet style, and its key explanation sentence falls below the fold on small viewports. **Most impactful improvement: give the advisory a bounded block (like `.boardSizeSetting`'s bordered card in Game Settings) with a `warning` icon, at body size, placed *above* the two buttons — and replace the native `confirm()` with the app's own `confirmOverlay()`, which already exists and already has "Add anyway / Cancel" style buttons.**

This is a **UI/UX review only** — layout, visual hierarchy, and how understandable the interface is for someone who has never used VTT. Code correctness is covered by another reviewer.

**What I actually exercised.** Ran the branch locally and drove it with Playwright. Real end-to-end flow: legacy game (`disableHolderImageWidget` on, a holder carrying `text`/`textColor`) → select → Toolbox → *Save widgets to buffer* → open a modern game → Toolbox panel warning → *Put widgets into the game* → confirm. Also the Widgets-module path (Save selected widget in the legacy game → *Add widget to room* in the modern game). Both a single difference and a three-difference buffer. Light and dark mode. Viewports: **390x844 (phone portrait), 844x390 (phone landscape), 820x1180 (tablet), 1280x800 (laptop), 1920x1080 (desktop)**. Overlay/panel sweep at all four sizes: Game Shelf, Players, About, and all eight editor modules (Edit Widgets, History, JSON, Widgets, Debug, Assets, Toolbox, Game Settings). Nothing outside the Toolbox/Widgets modules changed — the one shared-CSS line this PR adds is scoped to `.editorModule .legacyModeWarning`, and no other overlay or module regressed at any size. I did not reach the deck editor, share/publish or the JSON editor's sub-dialogs (nothing in the diff touches them).

| light | dark |
|---|---|

---

## Big layout improvements

### 1. The advisory has no container, so it reads as a footnote to the id list rather than a warning about it

`legacyModeWarningHTML()` (`client/js/editor/legacyModeWarning.js:47`) emits a bare `<div>` — bold line, `<ul>`, trailing sentence — and the CSS (`client/css/editor/sidebarModules.css:2114`) only sets `font-size: 0.9em; margin-top: 8px`. Rendered, the result is:

* the warning's `<ul>` uses the **same bullet, indent and list style** as the buffer-contents `<ul>` immediately above it, so `discardHolder` / `scoreButton` and `Convert numeric var parameters to numbers: …` look like one list in two colours;
* the trailing explanation sits *outside* the bulleted block with a blank line above it, so it reads as an unrelated stray sentence rather than the payoff of the warning;
* nothing separates the whole thing from the Search-and-Replace section that follows.

The editor already has the right pattern one module over: **Game Settings → Board Settings** wraps its explanation + fields + amber `boardSizeMessage` in a bordered `.boardSizeSetting` card ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1540090683820351709/ui-review-3138/game-settings-card-convention-1280.png)). Suggest the same here: a bordered/tinted block around the whole advisory, an `icon="warning"` glyph on the heading line (the app's own `confirmOverlay` markup uses `<div icon="warning">` — `client/room.html:457`), and a plain-text label above the buffer list (`In the buffer:`) so the two lists stop competing.

### 2. Size the warning *up*, not down, and put it where the decision is made

At 1920 the buffer-id list renders at 16px and the warning at **14.4px** — the more important message is the smaller one ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1540090683820351709/ui-review-3138/toolbox-warning-1920.png)). More seriously, it is rendered *below* both buttons, so on a short viewport you press *Put widgets into the game* without ever having seen it. At **844x390** (phone landscape) only the heading and the first of three modes are visible; the remaining modes and the whole explanation are below the fold, while both buttons are fully visible above:

clipped on a landscape phone

Suggest dropping `font-size: 0.9em` (1em, like the rest of the module) and moving the advisory **between the intro paragraph and the button bar** in `widgetBuffer()` / `renderWidgetBuffer()` (`client/js/editor/sidebar/toolbox.js:150`) — i.e. render it into its own `div` created before the `buttonBar`, so it is above the action it qualifies at every viewport size. It also renders fine at 390x844 and 820x1180 that way ([phone](https://agent.virtualtabletop.io/reports/ui-review/1540090683820351709/ui-review-3138/toolbox-warning-phone-390x844.png), [tablet](https://agent.virtualtabletop.io/reports/ui-review/1540090683820351709/ui-review-3138/toolbox-warning-tablet-820x1180.png)).

### 3. Use the app's own confirm dialog instead of `confirm()`

`confirmLegacyModeDifferences()` (`client/js/editor/legacyModeWarning.js:56`) uses the browser's native `confirm()`. VTT has a styled equivalent, `confirmOverlay(title, text, confirmLabel, cancelLabel, confirmIcon, cancelIcon, color)`, used for every comparable "are you sure" in the app (`client/js/overlays/states.js:617,888,1045`). Side by side, the house dialog is a titled card with a warning icon and two *labelled* buttons; the native one is a grey OS box captioned with the server's host name ("… says"), with generic **OK/Cancel** — which is why the message has to end with the instruction "*Press OK to add them anyway, or Cancel to abort.*", a sentence that disappears entirely if the buttons simply read **Add anyway** / **Cancel**.

the house confirm dialog

I know the Toolbox's existing duplicate-ids prompt is also a native `confirm()`, so the new one is locally consistent — but this PR doubles the native dialogs on that path (legacy-mode confirm at `toolbox.js:57`, then immediately the duplicate-ids confirm at `toolbox.js:62`), and two stacked OS boxes for one button press is a lot. If both moved to `confirmOverlay`, they could even be one dialog with both lists.

### 4. The Widgets library shows nothing until you have already clicked

`placeSavedWidgetFromBuffer()` (`client/js/editor/sidebar/widgets.js:200`) is the *only* place the library path says anything, and it is a modal that appears after the click. The saved-widget entry itself gives no hint at all:

library entry with no marker

The module already has exactly the affordance for this — a legend at its top ("⊙: Has a routine that runs when added to a room." / "☀: Is a unique widget.") and matching per-entry badges rendered in `renderListWidget()` (`widgets.js:1373`). A third badge on entries whose `legacyModes` differ from the current game, with a matching legend line and a tooltip carrying the same text as the dialog, would warn *before* the click and make the two editor paths consistent (Toolbox: passive panel note; Widgets: passive badge; both: confirm on add).

This matters more than it looks, because `placeSavedWidgetFromBuffer` is also reached by **dragging a saved widget onto the board** (`onRoomAreaDrop` → `placeSavedWidget`, `widgets.js:1497`/`:159`). Today that means: drag, drop, and *then* an OS modal appears — and if you cancel it, the widget silently does not appear where you dropped it, with no feedback that anything was declined. A badge you can see while dragging, plus (if it stays a confirm) a short "Not added." confirmation after a cancel, would close that loop.

---

## Nitpicks / small breakages

* `client/css/editor/sidebarModules.css:2114-2118` — `font-size: 0.9em` is the only sizing; there is no `margin-bottom`, so the advisory is 8px from the list above it and 0px from the `Search and Replace` header below. Add symmetric spacing (or the container from #1).
* `client/js/editor/legacyModeWarning.js:47-51` — the `<ul>` inherits the module's default list style. Give it its own class (e.g. `.legacyModeWarning ul { margin: 4px 0 6px; }`) so it is not pixel-identical to the buffer-contents list directly above it.
* `legacyModeWarning.js:39` — the explanation is emitted as a bare text node after `</ul>`, so it has no `<p>` and no spacing of its own; in the panel it looks like a sentence that fell out of the block. Wrap it.
* Colour/contrast is fine — `#8a5300` on white is 6.33:1, `#f0be5a` on the dark panel 10.99:1 — but colour is currently the *only* signal that this text is a warning. An icon (see #1) would carry it for colour-blind readers and in a screenshot-grey printout.
* `renderWidgetBuffer()` is now called from `onStateReceivedWhileActive` (`toolbox.js:133`), which rewrites `this.currentContents.innerHTML` on every received state. In a busy room that re-renders the buffer list under the reader's cursor (and drops any text selection in it) even when nothing about the buffer changed. Cheap guard: only re-render when the computed difference list actually differs from the one currently shown.
* Wording nit: the panel says "**These widgets come from a game with different legacy modes**" while the dialog says "**These widgets were saved in a game with different legacy modes**". Same fact, two phrasings, two lines apart in one flow — pick one (I'd take the dialog's).

---

* **"legacy modes" is never explained where it is used.** The warning names modes ("Disable holder image support: on in the source game, off here") but a newcomer has no idea what a legacy mode is or that it is a per-game switch. Game Settings has the full explanation, complete with per-mode summaries and Details disclosures. Suggest ending the advisory with a pointer — "See **Game settings** in the editor sidebar." — or, better, making the block's heading a link that opens the Game Settings module (the sidebar buttons are already programmatically clickable), so the reader can go read the mode they were just warned about.
* **"the source game" / "here" is guessy.** The reader may have filled the buffer days ago in a game they can no longer name. If the snapshot can carry it, "on in *Chess* (where these were copied), off in this game" would be unambiguous; failing that, "on where these widgets were copied from, off in this game" reads better than "in the source game, off here".
* **Nothing tells the reader what to do about it.** The advisory correctly stays non-blocking, but ends on a worry ("the widgets can work differently in this game") with no next step. One clause would fix it: "…so check them after pasting, or switch the mode on in Game settings."
* **The buffer contents list is unlabelled** (pre-existing, but the new list under it makes it worse) — a first-time user sees `• discardHolder` with no heading. "In the buffer (2 widgets):" costs one line and makes the second list obviously different in kind.
* The dialog's "*Press OK to add them anyway, or Cancel to abort*" is only needed because the native buttons are unlabelled — see big-improvement #3.

---

## Draft tutorial — "Xtras - Widget Buffer" (2 variants, in the [`pr-test-ai` room](https://test.virtualtabletop.io/PR-3138/pr-test-ai))

I played both variants end to end as a first-time reader: **1) Legacy game** → save to buffer → **2) Modern game** → Toolbox note → *Put widgets into the game* → confirm → the pasted holder draws its text. **The teaching works** — the two-room split is the right idea, the "Nothing warns you while you are still here" note in variant 1 pre-empts exactly the question the reader has, and variant 2's "the JSON did not change, the game around it did" is the sentence the whole feature turns on. What it doesn't do is *look* like the tutorials it would sit next to.

| variant 1 | variant 2 |
|---|---|
| variant 1 | variant 2 |

**1. It uses the left/right text-vs-demo split that the tutorial conventions rule out.** Both variants put the demo in a left column (`boardFrame`/`legacyHolder`/`dropFrame` at x 110–730) and a wall of prose in a right column (`steps`/`stepsNote`/`captionA` at x 810–1500). `docs/tutorials.md` § *Room layout & structure* asks for **top-down bands** — title, full-width overview, then a *row* of examples with **captions directly under each example** — and says "never a left/right text-vs-demo split". Compare with `Types - Line`, one of the reference rooms the doc names:

house layout: Types - Line

Reflowed into bands, variant 2 would read: title → overview → a row of *two* examples (the drop area, and the sample holder) with a caption under each → the "Try it" steps as the last band. That also fills the empty bottom-right quadrant both variants currently leave.

**2. Structural ids and geometry are off the convention** — `intro` (x60, y62, w1480, 24px) where the doc specifies `overview` at x108, y80, w1400, 25px; `captionA`/`steps`/`stepsNote` where the doc specifies `text1`, `text2`, … at 22px. Cheap to align, and it's what keeps the set consistent for the next author.

**3. The library image is in the wrong family colours.** Both existing Xtras tutorials use background `#c0c0c0` with a `#0d2f5e` border, white category text, black title and black glyphs — and both use the same lightbulb glyph bottom-right. The draft uses `#2f4858` with an invisible border, white title and a clipboard glyph, i.e. the *Foundations* palette in a shade no family uses:

library image comparison

In the Tutorials grid it would read as a different category from the two neighbours it sorts next to. Fix: `.s0 { fill:#c0c0c0; stroke:#0d2f5e }`, category `#ffffff`, title + both glyphs `#000000`, and reuse the family's lightbulb as the right glyph. Also, the existing images convert their text to `<path>` outlines; the draft keeps live `<text font-family="Arial, Helvetica, sans-serif">`, so its title will render with whatever the reader's system substitutes.

**4. Two first-time-reader gaps in the instructions.**
* Variant 1 step 3 says "Open variant **2) Modern game** of this tutorial" without saying how — a newcomer doesn't know that means Game Shelf → the tutorial's tile → *Play* on the second row (and that a "Switch game" confirm will appear). One clause fixes it.
* Both variants open with "Enter edit mode" with no hint where that is. Other tutorials tend to name the gesture; "press the ✎ edit button in the top toolbar" would match.

**5. Smaller things.** The title uses `Widget buffer - Legacy game` where the convention is `"<Subject>: <Variant>"` with a colon. The variant-2 drop area is a plain grey rectangle with a sentence in its top-left corner — a dashed outline and a centred label would make it read as a target (the Widgets module's own "+ New Group" zone is the in-app precedent). The intro paragraphs carry five bold spans in three lines, which flattens the emphasis; bold the two term names and leave the rest plain. And the details page shows "Time: 0 minutes" ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1540090683820351709/ui-review-3138/tutorial-details-1280.png)) — `time: "0"` matches `Properties - dragLimit`, so this is consistent, but most tutorials use `""` and then the line is simply absent, which reads better.

VTT-AI-Agent:
Thanks — this was a good read and almost all of it applied. Everything below is in `7c863a77`, plus a rebuilt demo room on the test server.

## Big layout improvements

**1. No container → a framed tile with an icon.** `legacyModeWarningHTML()` now emits `<div class="settings-tile legacyModeWarning">` — the same `.settings-tile` frame Game Settings gives `.boardSizeSetting` — with a bold heading carrying a `warning` glyph (a `::before` in the Material Symbols face, so it needs no markup of its own), the mode list in a `<ul>` with its own margins, and the explanation wrapped in a `<p>`. The buffer contents above it now have a heading, `In the buffer (2 widgets):`, so the two lists are visibly different in kind. I extended the `.settings-tile` comment to say it is no longer only a Game Settings thing.

**2. Sized up and moved above the buttons.** `font-size: 0.9em` is gone — it renders at the module's body size. `widgetBuffer()` now builds the intro paragraph, then the contents div, then the button bar, so the buffer list and the advisory are always above the two buttons. At 844x390 that inverts your screenshot: the advisory is what you see, and you have to scroll past it to reach *Put widgets into the game*.

**3. `confirmOverlay()` instead of `confirm()`.** `confirmLegacyModeDifferences()` is now `async` and raises the house dialog: title *Different legacy modes*, warning glyph, **Add anyway** / **Cancel**. The "Press OK to add them anyway" sentence is gone with it. I took the duplicate-ids prompt in the same function along (**Overwrite** / **Cancel**, title *Widget IDs already exist*), so that button press no longer stacks two OS boxes. I left them as two dialogs rather than merging the lists: they ask different questions, only the second one is about overwriting, and the merge would have to bypass the helper `widgets.js` shares. `confirmOverlay` had to be added to the whitelist `loadEditMode()` hands the editor bundle — `showOverlay` was already there.

**4. The library warns before the click.** `renderListWidget()` and `renderGridWidget()` compute `legacyModeDifferences(state.legacyModes, state.widgets)` and add a third badge next to `hasRoutine`/`isUnique`, with a matching legend line ("Was saved in a game with legacy modes that differ from this one.") and the dialog's exact text as its `title`. So dragging an entry onto the board now shows the mismatch while you drag, and the confirm is the second time you hear it, not the first. On cancel-feedback: with a card titled *Different legacy modes* and a button reading **Cancel** I think the loop is closed well enough without a second "Not added." dialog — say the word if you disagree.

## Nitpicks

* Spacing, `<ul>` class, `<p>` around the explanation — all folded into the tile (its `margin: 8px 0` gives the symmetric gap).
* Colour is no longer the only signal: the heading glyph carries it, and the badge is a warning triangle.
* `onStateReceivedWhileActive` → `renderWidgetBuffer()` now compares the computed HTML against what is currently shown and returns early when it is identical, so a busy room no longer rewrites the list under the cursor.
* Wording: one `legacyModeWarningHeading` constant feeds panel, badge tooltip and dialog — the dialog's phrasing won, as you suggested.
* "the source game" / "here" → "on **where these widgets were copied from**, off **in this game**". I did not add the game's name: the snapshot stores only the modes, and adding a name would mean a second field that older entries do not have.
* The advisory now ends with a next step — "Check them after adding, or switch the mode on under **Game settings**." I left it as plain text rather than a link into the Game Settings module; that is a cross-module jump the sidebar does not do anywhere else, and I would rather not invent the pattern in this PR.

## Demo room

Rebuilt, same two variants, still only on the test server (<https://test.virtualtabletop.io/PR-3138/pr-test-ai>, download in the PR description) and still not in the diff.

* **Bands, not columns.** Both variants are now title → full-width `overview` → a row of two examples at the same `y` with a caption directly under each → a full-width "Try it" band. The empty bottom-right quadrant is gone. Variant 1's second example is a plain token that no legacy mode can reach, which gives the `detect()` gating something to teach.
* **Ids and geometry on convention**: `title` (x400, y-2, 60px), `overview` (x108, y80, w1400, 25px), `text1`/`text2`/`text3` at 22px, `layer: -3` on the text and `-4` on the drop area.
* **Library image** rebuilt from the `Xtras - Menus` file: same `#c0c0c0`/`#0d2f5e` background, same white category word, same graduate cap and lightbulb, and the title converted to outlines — no live `<text>` left, matching all 67 existing tutorial images. Side by side with both neighbours: <https://agent.virtualtabletop.io/reports/pr/1540042162744328303/widget-buffer-library-image.png>
* **Instructions**: variant 1 now says how to switch ("Open the game shelf, pick this tutorial and press **Play** on variant **2) Modern game** - confirm the switch") and both open with "Press the **edit** button (the pencil) in the top toolbar".
* **Smaller things**: title is `Widget buffer: Legacy game` / `Widget buffer: Modern game` with the colon; the drop area is a dashed outline with a centred label; the bold spans in the overviews are down to the terms being introduced. `time: "0"` I kept, since that is what `Properties - dragLimit` does and what `docs/tutorials.md` specifies.

`npm test` is at 1358 passing (the dialog's button labels and the "closed without choosing" case are new tests). Verified in Chromium at 1280x800, 844x390 and 390x844 in both themes, over both editor paths and both list and grid view.

VTT-AI-Agent (commented):
## TL;DR

PR #3138's core warning logic is now sound and the earlier review findings are addressed, but two UI correctness gaps remain. Most importantly, the new `confirmOverlay()` presentation collapses the intended mode list into a single run-on paragraph; saved-widget warning badges can also become stale after changing the current game's legacy modes.

## Findings

### 1. The confirmation loses the mode-list formatting (`client/js/editor/legacyModeWarning.js:44-46`)

`legacyModeWarningText()` separates the heading, each difference, and the explanation with newlines, but `confirmOverlay()` assigns that string via `innerText` to `<p data-field="text">`, and the confirm overlay has no `white-space: pre-line`/`pre-wrap` rule. Normal paragraph whitespace therefore collapses every newline and the leading spaces. With multiple differences, the dialog renders the entries directly next to one another without bullets, line breaks, or even terminal punctuation, making the final safety prompt substantially harder to parse than the framed Toolbox list.

Please either give `#confirmOverlay [data-field="text"]` a suitable `white-space` rule or render the differences as structured list markup through an overlay API that sanitizes HTML. A DOM-level test for two differences would catch this; the current helper test only inspects the raw string passed to the stub.

### 2. Widgets-library badges do not follow live legacy-mode changes (`client/js/editor/sidebar/widgets.js:1380`, `client/js/editor/sidebar/widgets.js:1404`)

Both list and grid badges calculate `legacyModeDifferences()` only during `renderWidgetBuffer()`, but `WidgetsModule` has no `onMetaReceivedWhileActive` or `onStateReceivedWhileActive` hook. If an author leaves the Widgets module open and toggles a legacy mode under Game Settings, the Toolbox advisory updates but the Widgets entry keeps its old badge state until some unrelated action rerenders the library. The add confirmation recalculates at click/drop time, so users can see no badge and then get a warning, or see a warning badge and then add without one.

Refresh the library when the legacy-mode snapshot changes, preserving the current filter and view state. Since rendering fetches the server-wide widget list, tracking the last current-mode signature before rerendering would avoid doing that work for every unrelated `meta`/`state` event.

## Standing checks

- This exposes no new widget property or routine operation, so no property-editor, visual-routine, validator, or `jeCommands` entry is required; the feature itself is implemented in the real Toolbox and Widgets editing UI.
- The diff touches nothing under `library/`. I inspected the linked **Xtras - Widget Buffer** `.vtt`: it contains two focused variants, matching tutorial metadata fields, the conventional title/overview/caption geometry, explicit legacy/modern game settings, and a 530×530 embedded SVG library image.
- No engine paths are touched (`client/js/widgets/`, `client/js/compute.js`, and `client/js/serverstate.js` are absent), so this changes editor-only transfer guidance rather than how existing games execute. All reported GitHub checks, including CodeQL and the TestCafe matrix, are green.

The `beta` label was already present; I left it in place because this editor-only change remains safe and useful for player testing despite the two presentation issues above.

VTT-AI-Agent:
Thanks — finding 2 was real and is fixed. Finding 1 was half right: the mechanism you describe does not happen, but the entries did run together for another reason, and that is fixed too. Both in `4ca4f846`.

## 1. The confirmation's mode list

The newlines survive. `applyValuesToDOM()` assigns the text with `innerText`, and the `innerText` **setter** turns each `\n` into a `<br>` element — it is not a text node whose whitespace CSS then collapses. Measured in the running app, the `<p data-field="text">` for three differences holds:

```html
These widgets were saved in a game with different legacy modes:<br><br>
Convert numeric var parameters to numbers: on where these widgets were copied from, off in this game<br>
Use 1 as default for var parameters: off where these widgets were copied from, on in this game<br>
Disable holder image support: on where these widgets were copied from, off in this game<br><br>
Legacy modes change how widgets behave, so these widgets can work differently here. …
```

So no `white-space` rule is needed, and adding one would change every other `confirmOverlay` caller that relies on the same behavior (`players.js:82`, `symbols.js:420`).

What *does* get lost is the two-space indent I used to set the entries off: leading spaces collapse at the start of a line box. With entries long enough to wrap — all of them are — line two of one entry and line one of the next looked the same, which is the run-together effect you saw. Each entry now starts with a `•`, so a wrapped entry is still one visible item:

the dialog with three differing modes

I kept it plain text rather than moving to HTML markup through another overlay API: the same string is the badge's `title` attribute, where markup would show up literally, and plain text keeps a label out of the HTML sanitizer's way entirely. The unit test now pins the multi-difference shape (each entry its own bulleted line, in order) instead of only checking that a label appears somewhere in the string.

## 2. Stale badges in the Widgets library

Confirmed and fixed. `WidgetsModule` now has `onMetaReceivedWhileActive`/`onStateReceivedWhileActive`, both going through `rerenderOnLegacyModeChange()`, which compares `JSON.stringify(currentLegacyModes())` against the signature stored by the last `renderWidgetBuffer()` and returns without doing anything when they match — so the server-wide widget fetch does not happen for unrelated `meta`/`state` traffic. The rerender passes the value currently in `#widgetFilter`, and view mode, sort and the editing flag were already instance/DOM state that survives it.

Verified in Chromium with the Widgets module open the whole time, in both list and grid view: badge count 0 → toggle *Disable holder image support* under Game settings → 1, toggle back → 0, filter still `legacyHolder` and still one entry throughout, and an unrelated widget added to the room leaves the library DOM untouched.

the badge appearing after a live mode change

`npm test` is at 1359 passing.

VTT-AI-Agent (commented):
## TL;DR

The logic is in good shape now — I ran the 12 new jest tests (all pass) and drove both editor paths in a real browser: the advisory renders as intended, the confirmation reads well, and the library badge really does follow a live legacy-mode toggle (badge count 1 → 2 the moment I flipped `disableHolderImageWidget`). One real regression came out of the switch from `confirm()` to `confirmOverlay()`: **answering either new dialog closes an editor module that is open in the module overlay and never brings it back**, leaving the sidebar button stuck `active`. Also, review 3's "the dialog collapses the mode list" finding is not real — `innerText` converts the newlines to `<br>` — so please don't "fix" it with a `white-space` rule.

## Findings

### 1. Answering a dialog kills a module opened in the overlay (`client/js/editor/sidebar/toolbox.js:57`, `:62`, `client/js/editor/sidebar/widgets.js:200`)

`confirmOverlay()` starts with `showOverlay('confirmOverlay')`, which sets `display:none` on **every** `.overlay` — including `#editorModuleOverlay` (`client/editor.html:285`) — and every one of its exits calls `showOverlay()` with no id, which hides everything and leaves `overlayActive` false. Nothing puts the module overlay back. The native `confirm()` this replaced did not touch overlays, so this is new.

Verified end to end on the branch (1280x800, Chromium):

1. edit mode → **shift-click** the `TOOLBOX` sidebar button — `SidebarModule.click()` (`client/js/editor/sidebarModule.js:92`) opens the module in `#editorModuleInOverlay`;
2. buffer holds a holder saved with `disableHolderImageWidget: true`, current game has it off, so the advisory shows: toolbox in the overlay
3. *Put widgets into the game* → the dialog
4. **Add anyway** → the widget is added correctly, and the panel is gone: panel gone

Measured after the click: `getComputedStyle($('#editorModuleOverlay')).display == 'none'`, while the sidebar reports `['EDIT WIDGETS', 'TOOLBOX']` as its active buttons. The module still believes it is open, so **one** click on `TOOLBOX` does not bring it back — `openInTarget()` takes the "already open" branch and closes it; you need a second click. **Cancel** behaves the same way (I checked both), which is worse: you decline the paste and the panel you were reading disappears with it.

This hits all three new call sites — both dialogs in `button_loadWidgetsFromBuffer()` and the one in `placeSavedWidgetFromBuffer()` (the Widgets module can be shift-clicked into the overlay too).

The house pattern for this is already in the tree — `client/js/overlays/players.js:83` re-shows its own overlay right after awaiting `confirmOverlay`. The Toolbox/Widgets equivalent is roughly:

```js
const reopen = this.moduleDOM && this.moduleDOM.id == 'editorModuleInOverlay';
...
if(reopen) showOverlay('editorModuleOverlay');
```

`placeSavedWidgetFromBuffer()` is a free function with no module reference, so either the confirm moves up into `placeWidget()`/`onRoomAreaDrop()`, or the restore becomes a small shared helper next to `confirmLegacyModeDifferences()` (`if($('#editorModuleInOverlay').children.length) showOverlay('editorModuleOverlay')`).

### 2. Cancelling on the line-stop path closes the picker as though it had placed something (`client/js/editor/sidebar/properties.js:10088`)

`placeSavedWidgetFromBuffer()` now returns `[]` when the user cancels, but the library stop picker ignores the result:

```js
const place = async entry=>{
  await lineEdit(`added stop from the library to line ${line.id}`, _=>placeSavedWidget(entry.id, entry.source, null, { line }));
  popout.style.display = 'none';
  onPlaced();
};
```

Press **Cancel** and the popout still closes and `onPlaced()` still runs, so the flow looks exactly like a successful placement with a stop that silently isn't there. Thread the ids out and bail:

```js
let placed = [];
await lineEdit(`added stop from the library to line ${line.id}`, async _=>{ placed = await placeSavedWidget(entry.id, entry.source, null, { line }); });
if(!placed.length) return;
```

(The drag-onto-the-board path, `widgets.js:1518`, has the same silence, but there "nothing happens" at least matches "Cancel". No toast helper exists in the client, so I would not invent one for it.)

Related, lower priority: `lineEdit` (`properties.js:10433`) is `batchStart()` → `await change()` → `batchEnd()`, so this confirm is the first modal the codebase awaits **inside an open delta batch**. The batch is empty while the dialog waits and `sendDelta()` skips empty deltas, so there is no phantom undo entry today — but a confirm inside a batch is a shape worth avoiding; doing the confirm before `lineEdit()` would keep batches synchronous.

### 3. The buffer's mode snapshot lives in a second localStorage key that can drift (`client/js/editor/sidebar/toolbox.js:51`, `:121`)

`widgetBuffer` and `widgetBufferLegacyModes` are two independent keys, only kept in step because the same function writes both. Anything that writes `widgetBuffer` without the companion key — most plausibly an older tab still open across a deploy, since localStorage is shared per origin — leaves a snapshot describing a **different** game behind, and `legacyModeDifferences()` then confidently names a difference that does not exist (or stays quiet about one that does). That is precisely the failure the "unknown is not the same as different" rule was written to avoid, except here the data looks known.

Cheap and permanent fix: keep the snapshot inside the stored value rather than beside it, e.g. write `{ legacyModes, widgets }` and read `Array.isArray(parsed) ? { widgets: parsed, legacyModes: null } : parsed` so an old array-shaped buffer degrades to "no snapshot". The library entries already do it this way (`legacyModes` as a field on the entry), so it would also make the two paths the same shape.

### 4. Not a bug — do not apply review 3's finding 1 (`client/js/editor/legacyModeWarning.js:44`)

Review 3 asked for `white-space: pre-line` on `#confirmOverlay [data-field="text"]` because the newlines "collapse". They don't: `applyValuesToDOM()` (`client/js/domhelpers.js:449`) assigns through `innerText`, whose **setter** turns `\n` into `<br>` elements. I dumped the result in the browser:

```
some text<br><br>• line one<br>• line two<br><br>end
```

and the screenshot in finding 1 shows the bullet on its own line. Adding a `white-space` rule now would double every blank line. The comment at `legacyModeWarning.js:41-43` already describes this correctly — worth keeping so the next pass doesn't undo it.

The *other* half of that comment does apply to the new duplicate-ids dialog, though: `toolbox.js:62` writes `\n\n  ${duplicatesList}`, and the two-space indent is dropped exactly as the comment predicts (visible in my run — `legacyHolder` renders flush left). Either drop the indent or bullet the ids the way the legacy-mode dialog bullets its modes.

### 5. Test coverage sits entirely below the wiring

`tests/client/legacy-mode-warning.test.js` is good — I ran it locally, 12/12 pass, and it pins the two silences and the direction of the wording. But every problem above is in the wiring, not the helper: which DOM the dialog disturbs, what the caller does with `[]`, which localStorage keys move together. `tests/testcafe/editor.js` already drives the sidebar; a short case (seed the buffer, open the Toolbox, assert the `.legacyModeWarning` tile, press the button, assert the dialog, assert the module is still on screen afterwards) would have caught finding 1 and would guard the panel placement that review 2 asked for.

## Standing checks

* **UI support for anything new:** nothing new is exposed to game authors — no widget property, no routine operation or parameter function — so there is no `operationProps`/`jeCommands` obligation, and the feature itself lives in the real editing UI (Toolbox and Widgets sidebar modules), not in the JSON editor. The new `legacyModes` field is on *saved-widget library entries* (localStorage and `/api/widgets`), not in game JSON, so `validate_gamefile.js` is untouched by it; I checked `server.mjs:349` — the PUT handler stores `data.widgets` verbatim, so the field rides along to the server list with no server change needed. Nothing to flag.
* **Test room:** the diff touches nothing under `library/` — correct. I unpacked the published `Xtras - Widget Buffer.vtt`: two variants, `_meta.info` complete and matching the tutorial fields, `_meta.gameSettings.legacyModes.disableHolderImageWidget: true` on variant 1, an `/assets/...` library image in the zip, and top-down band geometry with the conventional ids (`title` 60px, `overview` at x108/y80/w1400, `text1..3` at 22px). That matches `docs/tutorials.md`, and it is correctly kept out of the diff.
* **Verification I ran:** `npx jest tests/client/legacy-mode-warning.test.js` → 12 passed. Branch served on a local server and driven with Playwright: the Toolbox advisory renders as a framed amber tile above the buttons; the confirmation renders its bullets on separate lines; the widget is added on **Add anyway** and not added on **Cancel**; and the Widgets-library badge count went 1 → 2 the instant I called `legacyMode('disableHolderImageWidget', true)` with the module open, so review 3's finding 2 is genuinely fixed.
* **Engine:** none of `client/js/widgets/`, `client/js/compute.js` or `client/js/serverstate.js` is touched. `client/js/main.js` gains one name in the `Object.assign(window, …)` list that hands globals to the editor bundle, and `server/minify.mjs` one filename — neither changes how existing games run.

The `beta` label was already there and I left it: nothing in the diff touches the engine, all checks are green, and the worst case for a player is an editor panel that has to be reopened after answering a dialog — annoying, but it corrupts nothing and crashes nothing, so it stays useful to test.

VTT-AI-Agent:
Review 4/4 addressed in 5b0a25d — all three findings were real; the fourth is left alone as you asked.

**1. Answering a dialog killed a module opened in the overlay — fixed.** Confirmations raised from the editor now go through a small `confirmInEditor()` helper next to the other module helpers in `client/js/editor/sidebarModule.js`:

```js
async function confirmInEditor(...args) {
  const confirmed = await confirmOverlay(...args);
  const moduleInOverlay = $('#editorModuleInOverlay');
  if(moduleInOverlay && moduleInOverlay.classList.contains('active'))
    showOverlay('editorModuleOverlay');
  return confirmed;
}
```

It wraps rather than repeats the `players.js:83` pattern because all three new call sites need it, and `placeSavedWidgetFromBuffer()` has no module to ask. Both dialogs use it, so `button_loadWidgetsFromBuffer()` and the Widgets module are covered — I drove Cancel and Add anyway on both, and the panel, its `active` class and the sidebar button all survive: [after Cancel](https://agent.virtualtabletop.io/reports/pr/1540042162744328303/review4-4-after-cancel.png). Closing the module afterwards with one shift+click still hides the overlay properly.

**2. The line-stop picker ignored the `[]` — fixed.** `place()` now threads the ids out of `lineEdit()` and bails before closing:

```js
let placed = [];
await lineEdit(`added stop from the library to line ${line.id}`, async _=>{
  placed = await placeSavedWidget(entry.id, entry.source, null, { line });
});
if(!placed.length)
  return;
```

Cancel leaves the picker open with the room untouched (`theLine.stops` still `null`, no widget added) and the accept path still closes it and places the stop: [picker after Cancel](https://agent.virtualtabletop.io/reports/pr/1540042162744328303/review4-6-stop-picker-after-cancel.png). I left the confirm inside `lineEdit()` — the batch stays empty while the dialog waits, so there is nothing to undo, and hoisting it would mean duplicating the difference check outside `placeSavedWidgetFromBuffer()`.

**3. The second localStorage key — gone.** `widgetBuffer` now stores `{ legacyModes, widgets }` as one value and `widgetBufferLegacyModes` is removed, so the snapshot cannot describe a different game than the widgets beside it. A bare array — what any browser that filled the buffer before this change has — reads back as "widgets, no snapshot" and stays silent, which I checked in the browser: the old-shape buffer lists, pastes and asks nothing.

**4. `white-space: pre-line` — agreed, not applied.** Thanks for chasing the `innerText` setter down; the comment at `legacyModeWarning.js:41-43` stays. The other half does apply and is fixed: the duplicate-ids dialog bullets its ids instead of indenting them — [dialog](https://agent.virtualtabletop.io/reports/pr/1540042162744328303/review4-3-duplicate-dialog.png).

**5. Coverage above the helper — added.** `tests/testcafe/editor.js` gained *"Pasting the widget buffer across legacy modes asks, and keeps the module it asks from"*: it seeds the buffer, shift-clicks the Toolbox into the module overlay, asserts the advisory tile, presses **Put widgets into the game**, answers the dialog, and then asserts the widget arrived *and* `#editorModuleOverlay` is still visible with both the module and its sidebar button `active`. I confirmed it fails on the branch with the `showOverlay('editorModuleOverlay')` line removed (`expected false to be truthy` at the overlay assertion) and passes with it.

Verified locally: `npm test` 1359 passed, and `npx testcafe chromium:headless tests/testcafe/editor.js -T "Line (widget|stops)"` plus the new case all green.

ArnoldSmith86 (approved):

PR #3139, opened 2026-08-20
Tick timers from the wall clock and hand the ticking on when a client leaves

### LINKED ISSUE #1655: Timer stops ticking if first joining player leaves
When the first player joins a room with a paused timer, they start a timer on their system to tick it:
https://github.com/ArnoldSmith86/virtualtabletop/blob/17a17ae00d774a0f1e7fde4334e66abdd6e90c80/client/js/widgets/timer.js#L39

However, if that player leaves the timer stops.

VTT-AI-Agent:

- **Should a timer a routine follows catch up on the intervals a backgrounded tab skipped?** The clock faces in the *Appearance* variant of the timer tutorial stand still while their window is minimized, because a timer with a `millisecondsChangeRoutine` is given every value it passes through and so counts only the ticks its browser delivered — unchanged by this PR, and the same on `main` ([measurements and the options](https://github.com/ArnoldSmith86/virtualtabletop/pull/3139#issuecomment-5416974468)). Leave it as it is, replay the skipped intervals when the tab wakes up, or put it under a property the game sets?

A running timer widget is counted up by one of the clients in the room, which sends every new `milliseconds` value to the others. That client was picked once — whoever started the timer, or the first player in the room — and each tick simply added one interval to the previous value. Both halves of that made timers lose time or stop altogether:

- **The timer stopped when that client left the room** (#1655). Nobody took over, so it stood still while still showing itself as running, until somebody paused and restarted it.
- **A background tab made the timer run slow.** Browsers throttle `setInterval` in tabs that are not visible (down to once a minute) and suspend them completely on a sleeping device. Every interval that was skipped was a second the timer never counted, and it never caught up.
- **Two connections of the same player counted twice.** The check that started the ticking on load counted *players*, not connections, so a second tab or a second device under the same name could end up with two clients ticking and the timer running at double speed.

## What changed

- **The ticking belongs to the client that started the timer** — the one whose player pressed the button or ran the `TIMER` operation — so the timer's `millisecondsChangeRoutine` keeps running with that player's name, seats and inputs. Everybody else only writes once the value has stood still for a whole interval plus three seconds: the primary session (the oldest connection in the room) first, the rest of the room after a second such period. So a timer whose writer has left, and one that was already running when the room was loaded, is taken over by exactly one client, and even a stalled primary session gets replaced.
- **A timer nobody watches is derived from the wall clock** instead of being added to the previous value: a tick works out how many whole intervals have passed since the value it last measured from. A tab that was throttled or asleep therefore lands on the time that really passed on its next tick instead of falling behind.
- **A timer something watches passes on every value it takes.** If the timer has a `millisecondsChangeRoutine` or a `changeRoutine`, or another widget listens for updates, it advances by exactly one interval per tick as it always did — a countdown that acts on `milliseconds == 100` is still given that value when a throttled tab wakes up or another client takes the writing over. The intervals its browser skipped are time it did not count, as before; it just no longer stops for good. A client that takes such a timer over carries on from the value the previous one last wrote.
- **Time a routine adds while the timer runs is kept.** Any update a client did not write itself — `TIMER` with `inc`/`set`, another client's tick — becomes the new base the following ticks are measured from. An update that says exactly what the client would have written itself leaves the base alone, so the part of the interval that has not elapsed yet is not lost on every update.
- **A client that is not writing still shows the time.** The value is derived from the wall clock, so every client knows it without being told. While a stalled writer is given its grace period, the others keep their own display moving instead of standing still and then jumping. A timer something watches is excluded, because it does not catch up — its display would run ahead of the value it is going to take.
- **An hour or more gets an hours field.** The display was `m:ss` with an unbounded minute count, so an hour read as `60:00` and a tab that slept for three hours as `193:07`. From an hour on the value now reads `1:00:00` / `3:13:07`; below an hour nothing changes. The Start / End / Current time fields of the widget editor format and accept the same `h:mm:ss` form.
- **A tick that writes nothing leaves nothing behind.** Consecutive ticks share one undo entry, labelled `timer ticked`. A tick that finds the value already written — for the moment two clients write, the other one got there first — no longer leaves that cause on the delta for the player's next action to inherit and be merged into the timer's undo entry.
- **A widget preview outside the room no longer starts a timer.** The property editor and the widget sidebar render copies of a widget that are not part of the room; for a running timer such a copy used to start ticking and try to write to the room state under an id the room does not have. The sidebar's drag image now marks its previews as the read-only copies they are and takes them down through `applyRemove` when the drag is over, which also releases the styles, listeners and parent bookkeeping every other widget type left behind. A timer additionally refuses to write a tick for a widget the room does not hold.
- **A frozen tab hands the timer on instead of pushing a tick of its own.** The client that started a timer keeps writing it for as long as it runs, but it now gives that claim up once it has both stopped writing long enough to be taken over and been overtaken since. Without that, a tab coming back from being frozen wrote again straight away, which on a timer a routine follows put an extra interval on the clock and ran the routine an extra time. A one-off change from somebody else — a routine adding a minute — leaves the writing where it is.
- **An hours field is understood wherever a game writes a time, not just where one is shown.** `timeToMS`, the parser behind `start`, `end` and `TIMER`'s `value` / `seconds`, only knew `m:ss`, so the form the widget and the property editor display meant something else entirely in the JSON editor: `"start": "1:00:00"` reset a timer to 1 ms, an `"end"` of `"1:00:00"` never turned the alert on, and `TIMER seconds: "1:00:00"` set 0 — all three silently. It now takes an optional hours field. Nothing that parsed before changes, and a string with two colons could not mean anything until now.

No widget property or routine parameter is added, renamed or removed, so there is nothing to add to the validator or to `jeCommands` — the timer section of the property editor keeps exactly the fields it had, its time fields just render and accept an hours field as well.

## User-visible effects

- A running timer keeps running when the player who started it closes the tab; the room picks it up after about four seconds.
- Coming back to a tab whose plain display timer was throttled or asleep now shows the time that really passed — the value can jump forward by the time the tab was away, instead of being permanently behind.
- A timer that drives a routine still counts one tick per interval and skips no value, so games that act on an exact `milliseconds` value keep working unchanged.
- Time still does not pass while nobody is in the room; the timer continues from where it was when the first client comes back, as before.
- In edit mode the undo protocol no longer fills with one entry per tick: consecutive ticks share an entry, labelled `timer ticked`.
- A timer past an hour reads `1:00:00` instead of `60:00` ([before/after](https://agent.virtualtabletop.io/reports/pr/1540126994505666660/timer-hours-format.png)), which matters more now that a tab coming back from the background catches up instead of losing the time.

## Verified

- New jest fixture `tests/client/timer-tick.test.js` (28 tests) covering the interval, pausing and resuming, counting down, a frozen tab catching up, time added by a routine, a removed timer, a widget preview, following another client, keeping the display moving while waiting for it, taking the writing over in the right order, the starter keeping the writing even when it is not the primary session (including a start a routine only sends when it ends, and a restart that hands the writing on), a watched timer passing on every value it takes and carrying on from the value it was last told, a no-op tick leaving no delta cause behind, a dropped drag preview writing nothing, a `globalUpdateRoutine` in the room putting every timer on the value-by-value path, and a frozen starter dropping its claim after a takeover while keeping it when somebody else only adds time. `tests/client/primary-session.test.js` covers picking the session that takes over.
- `npm test`: 1383 tests pass. TestCafe `multiclient.js` passes locally; `editor.js` passed on the previous commit. (`routines.js` has one failure, `moving the mouse while a DELAY is running does not drag the clicked widget`, which reproduces identically with `main`'s `timer.js` in place, so it is not from this change. No fixture renders a timer — `render-fixtures.js` leaves it out on purpose.)
- Manually in Chromium with three clients in one room: the client that pressed start does all the writing and the other two send nothing, whether or not it is the oldest connection; when it closes its tab the oldest of the rest takes over after the grace period and the others stay quiet; when it freezes without disconnecting for nine seconds, the primary session's turn passes and the third client takes over; a timer with a `millisecondsChangeRoutine` was watched value by value through a takeover and produced `1000, 2000, …, 8000` with nothing skipped, the routine running under the starting player's name until the takeover. A four-timer tutorial room with precisions of 1000, 5000 and 500 ms was walked end to end and all of them keep step with each other. Pause, resume and reset behave as before; the undo protocol still collapses consecutive ticks into one entry and a player action after them keeps its own. The hours format was checked in the browser at the widget's default size and in the editor: `59:59` renders as before, `1:00:00` / `3:13:07` / `-3:13:07` render with hours, and the editor's Current time field round-trips `3:13:07`.
- The hours parsing was walked in a room in Chromium: with `"start": "1:00:00"` and `"end": "1:00:05"`, a `TIMER reset` puts the timer at `1:00:00` (it used to be 1 ms), `TIMER set seconds: "1:00:00"` does the same (it used to be 0), and the alert turns on five seconds later (it never did). Typing `2:30:15` into the editor's Current time field writes 9 015 000 ms and reads back as `2:30:15`.
- The drag previews were checked in edit mode in Chromium by dragging a saved widget that holds a running timer: before the fix the preview kept a one-second interval running for the rest of the session and threw once a second; now no interval is started at all, and dropping the widget still places it and it ticks in the room.

## Try it

A demo room is in this PR's test room [pr-test-ai](https://test.virtualtabletop.io/PR-3139/pr-test-ai), where it shows up in the game shelf once the test server has finished starting. It is written as a tutorial - **Types - Timer**, variant **Multi-client**, one room - with four examples: a plain count-up, a countdown from 2:00 with the alert, a `TIMER inc` button that adds a minute while the clock runs, and a timer whose `millisecondsChangeRoutine` counts its own ticks beside it, so you can watch the routine run once per interval however many people are watching and see that it skips no value. A `+ 1 hour` button next to the `+ 1 minute` one pushes a timer past the hour so the `1:00:00` display can be tried out. Things worth trying are written on the board: start it in one tab and close that tab, leave a tab in the background for a while, leave the room entirely and come back. The file itself is at [Timer check.vtt](https://agent.virtualtabletop.io/reports/pr/1540126994505666660/Timer%20check.vtt). It lives on the test server only - it is not part of this PR and nothing under `library/` is touched. It can be added to `library/tutorials/` on request.

## The timer tutorial, amended

`library/tutorials/Types - Timer/0.json` (variant **Basics**) told the reader that "one important limitation of the timer is that it is controlled by the player that started the timer. If that person leaves the room or the player's internet connection is disrupted while the timer is running, the timer will stop." The first half stays true - the player who starts a timer is still the one whose client counts it - but the second half stops being true with this PR.

The amended variant was published as a draft on this PR's test server and **accepted there by a maintainer (lawdawg96)**, who has since confirmed the amendment on this pull request, so it is part of this diff. Only `0.json` of that tutorial changes; variants `1.json` and `2.json`, its `assets/` and every other tutorial are untouched, and its `_meta.info` keeps the file's own fields unchanged apart from `lastUpdate` - the library image stays the existing square SVG `assets/1382176810_3946`. Five passages changed:

- the overview now says the timer belongs to the room, that another player carries it on when the one who started it closes their tab, and that a tab left in the background shows the time that really passed;
- the last example, the one already driving a `changeRoutine` every 500 ms, gains a sentence saying that a timer a routine follows is given every step it takes and so carries on where it was instead of catching up;
- the first example's "Add Widget button on the left side of the screen" becomes the `+` button in the top toolbar (there is no add control on the left any more);
- the first example mentions that an hour or more reads `1:00:00`;
- the overview's sentence about what the JSON editor accepts names all three forms a time can take now that `timeToMS` understands an hours field: "timers take milliseconds, minutes:seconds or hours:minutes:seconds for start and end as well as for TIMER's value and seconds parameters".

`node validator/validate_gamefile_node.js` exits 0 with no output for all three variants, and the room was walked end to end in the browser: all four timers start and pause from their own controls, the countdown reaches `0:00`, the 5000 ms example updates every five seconds, and the 500 ms one keeps step with the others while recolouring the holder below it. The hours sentence renders in the same six lines as the sentence it replaces, at the same line breaks, so the overview still fills the band exactly as before. [Types - Timer 0.vtt](https://agent.virtualtabletop.io/reports/pr/1540126994505666660/tutorials/Types%20-%20Timer%200.vtt) on the test server is packed straight from the committed file, so that room shows exactly what the repository holds.

Closes #1655.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3139/pr-test (or any other room on that server)

### DIFF
--- a/client/js/domhelpers.js
+++ b/client/js/domhelpers.js
@@ -113,12 +113,12 @@ export function rand() {
   return number;
 }
 
-// converts "minutes:seconds" or "minutes:seconds.fraction" strings to milliseconds
-// (rounded to the nearest millisecond); returns any other value unchanged
+// converts "minutes:seconds" and "hours:minutes:seconds" strings, with or without a fraction of a
+// second, to milliseconds (rounded to the nearest millisecond); returns any other value unchanged
 export function timeToMS(value) {
-  const match = typeof value == 'string' && value.match(/^(-?)(\d+):(\d+(?:\.\d+)?)$/);
+  const match = typeof value == 'string' && value.match(/^(-?)(?:(\d+):)?(\d+):(\d+(?:\.\d+)?)$/);
   if(match)
-    return (match[1] ? -1 : 1) * Math.round((+match[2]*60 + +match[3])*1000);
+    return (match[1] ? -1 : 1) * Math.round(((+match[2] || 0)*3600 + +match[3]*60 + +match[4])*1000);
   return value;
 }
 
--- a/client/js/editor/sidebar/properties.js
+++ b/client/js/editor/sidebar/properties.js
@@ -9126,7 +9126,7 @@ class PropertiesModule extends SidebarModule {
     });
   }
 
-  // a text input showing a milliseconds property as mm:ss
+  // a text input showing a milliseconds property as mm:ss, with an hours field from an hour on
   renderTimerTimeInput(widget, labelText, property, target, options = {}) {
     const wrap = div(target, 'propertyInput timeInput');
     const label = document.createElement('label');
@@ -9137,7 +9137,7 @@ class PropertiesModule extends SidebarModule {
 
     const input = document.createElement('input');
     input.type = 'text';
-    input.placeholder = 'mm:ss';
+    input.placeholder = '[h:]mm:ss';
     input.onchange = () => {
       const ms = parseTimerInput(input.value);
       if(ms === undefined || (ms === null && !options.nullable)) {
--- a/client/js/editor/sidebar/properties.js
+++ b/client/js/editor/sidebar/properties.js
@@ -1038,19 +1038,24 @@ function formatTimerMs(ms) {
   const minutes = Math.floor(totalSeconds / 60);
   const seconds = totalSeconds - minutes * 60;
   const secondsString = (seconds < 10 ? '0' : '') + (Number.isInteger(seconds) ? seconds : +seconds.toFixed(3));
-  return `${negative ? '-' : ''}${minutes}:${secondsString}`;
+  // an hour or more gets an hours field, the same way the timer widget shows it
+  const hours = Math.floor(minutes / 60);
+  const minutesString = hours ? `${hours}:${String(minutes - hours*60).padStart(2, '0')}` : `${minutes}`;
+  return `${negative ? '-' : ''}${minutesString}:${secondsString}`;
 }
 
-// accepts "mm:ss", "m:ss.s" or plain seconds; returns milliseconds,
+// accepts "mm:ss", "m:ss.s", "h:mm:ss" or plain seconds; returns milliseconds,
 // null for empty input and undefined for unparseable input
 function parseTimerInput(text) {
   const trimmed = String(text).trim();
   if(trimmed === '')
     return null;
-  const match = trimmed.match(/^(-)?(?:(\d+):)?(\d+(?:\.\d+)?)$/);
+  const match = trimmed.match(/^(-)?(?:(\d+):)?(?:(\d+):)?(\d+(?:\.\d+)?)$/);
   if(!match)
     return undefined;
-  return Math.round(((+match[2] || 0) * 60 + +match[3]) * 1000) * (match[1] ? -1 : 1);
+  // one number in front of the seconds is the minutes, two are hours and minutes
+  const [ hours, minutes ] = match[3] === undefined ? [ 0, +match[2] || 0 ] : [ +match[2], +match[3] ];
+  return Math.round(((hours * 60 + minutes) * 60 + +match[4]) * 1000) * (match[1] ? -1 : 1);
 }
 
 function parseFontSize(fontSize) {
--- a/client/js/editor/sidebar/widgets.js
+++ b/client/js/editor/sidebar/widgets.js
@@ -894,6 +894,8 @@ class WidgetsModule extends SidebarModule {
                 case 'timer': previewWidget = new Timer(tempId); break;
                 default: previewWidget = new BasicWidget(tempId); break;
               }
+              // the preview is not part of the room, so nothing it renders may be written back
+              previewWidget.isReadonlyCopy = true;
               previewWidget.domElement.style.transformOrigin = 'top left';
               tempWidgetInstances.set(tempId, previewWidget);
             }
@@ -928,7 +930,10 @@ class WidgetsModule extends SidebarModule {
           e.dataTransfer.setDragImage(previewContainer, 0, 0);
 
           setTimeout(() => {
-            for (const tempId of tempWidgetInstances.keys()) {
+            for (const [ tempId, previewWidget ] of tempWidgetInstances) {
+              // the preview is torn down the way a widget leaving the room is, so that whatever it
+              // registered while it rendered - a running timer, a listener - is released with it
+              previewWidget.applyRemove();
               widgets.delete(tempId);
             }
             document.body.removeChild(previewContainer);
@@ -1132,6 +1137,8 @@ class WidgetsModule extends SidebarModule {
               case 'timer': previewWidget = new Timer(tempId); break;
               default: previewWidget = new BasicWidget(tempId); break;
             }
+            // the preview is not part of the room, so nothing it renders may be written back
+            previewWidget.isReadonlyCopy = true;
             previewWidget.domElement.style.transformOrigin = 'top left';
             tempWidgetInstances.set(tempId, previewWidget);
           }
@@ -1166,7 +1173,10 @@ class WidgetsModule extends SidebarModule {
         e.dataTransfer.setDragImage(previewContainer, 0, 0);
 
         setTimeout(() => {
-          for (const tempId of tempWidgetInstances.keys()) {
+          for (const [ tempId, previewWidget ] of tempWidgetInstances) {
+            // the preview is torn down the way a widget leaving the room is, so that whatever it
+            // registered while it rendered - a running timer, a listener - is released with it
+            previewWidget.applyRemove();
             widgets.delete(tempId);
           }
           document.body.removeChild(previewContainer);
--- a/client/js/overlays/players.js
+++ b/client/js/overlays/players.js
@@ -24,9 +24,14 @@ export {
 // Work that exactly one client in the room may do - ticking a running timer - falls to the primary
 // session. The server numbers sessions in connection order, so every client picks the same one
 // without talking to the others, and the next session inherits the job as soon as that one
-// disconnects. While the session list is unknown, every client considers itself primary: doing the
-// work twice is something its caller has to cope with anyway (during a handover two clients hold
-// the job for a moment), not doing it at all is not.
+// disconnects.
+export function primarySessionOf(sessions) {
+  return (sessions || []).reduce((oldest, s)=>oldest === null || s.sessionID < oldest ? s.sessionID : oldest, null);
+}
+
+// While the session list is unknown, every client considers itself primary: doing the work twice is
+// something its caller has to cope with anyway (during a handover two clients hold the job for a
+// moment), not doing it at all is not.
 export function isPrimarySession() {
   return primarySessionID === null || primarySessionID === mySessionID;
 }
@@ -130,7 +135,7 @@ function showInviteStatus(text, isError) {
 
 function fillPlayerList(players, active, sessions) {
   activePlayers = [...new Set(active)];
-  primarySessionID = (sessions || []).reduce((oldest, s)=>oldest === null || s.sessionID < oldest ? s.sessionID : oldest, null);
+  primarySessionID = primarySessionOf(sessions);
   activeColors = activePlayers.map(playerName=>players[playerName]);
   removeFromDOM('#playersTable tbody > tr, #playerCursors > .cursor');
 
--- a/client/js/overlays/players.js
+++ b/client/js/overlays/players.js
@@ -8,6 +8,7 @@ let activePlayers = [];
 let activeColors = [];
 let mouseCoords = [];
 let mySessionID = null;
+let primarySessionID = null;
 let metaUpdateResolves = [];
 let inviteStatusTimeout = null;
 localStorage.setItem('playerName', playerName);
@@ -18,6 +19,16 @@ export {
   activePlayers,
   activeColors,
   mouseCoords
+}
+
+// Work that exactly one client in the room may do - ticking a running timer - falls to the primary
+// session. The server numbers sessions in connection order, so every client picks the same one
+// without talking to the others, and the next session inherits the job as soon as that one
+// disconnects. While the session list is unknown, every client considers itself primary: doing the
+// work twice is something its caller has to cope with anyway (during a handover two clients hold
+// the job for a moment), not doing it at all is not.
+export function isPrimarySession() {
+  return primarySessionID === null || primarySessionID === mySessionID;
 }
 
 function getPlayerDetails() {
@@ -119,6 +130,7 @@ function showInviteStatus(text, isError) {
 
 function fillPlayerList(players, active, sessions) {
   activePlayers = [...new Set(active)];
+  primarySessionID = (sessions || []).reduce((oldest, s)=>oldest === null || s.sessionID < oldest ? s.sessionID : oldest, null);
   activeColors = activePlayers.map(playerName=>players[playerName]);
   removeFromDOM('#playersTable tbody > tr, #playerCursors > .cursor');
 
--- a/client/js/widgets/timer.js
+++ b/client/js/widgets/timer.js
@@ -1,3 +1,9 @@
+// A running timer is written by one client only so that its shared millisecond count moves on once
+// per interval no matter how many people watch it. Normally that is the primary session, but a tab
+// that its browser has frozen or throttled stops writing without disconnecting, so every other
+// client takes over once the value has not moved for a whole interval plus this grace period.
+const timerTakeoverGrace = 3000;
+
 export class Timer extends Widget {
   constructor(id) {
     super(id);
@@ -20,32 +26,45 @@ export class Timer extends Widget {
     });
   }
 
+  // the value the elapsed time is measured from and the moment it was current. The last update
+  // this client saw doubles as the sign of life of whoever is writing the timer.
+  anchorTicking(milliseconds) {
+    this.tickTime = this.lastMillisecondsUpdate = Date.now();
+    this.tickMilliseconds = milliseconds;
+    delete this.tickedMilliseconds;
+  }
+
   applyDeltaToDOM(delta) {
     super.applyDeltaToDOM(delta);
     if(delta.milliseconds !== undefined) {
       const s = Math.floor(Math.abs(delta.milliseconds)/1000);
       setText(this.domElement, `${delta.milliseconds < 0 ? '-' : ''}${Math.floor(s/60)}:${Math.floor(s%60)}`.replace(/:(\d)$/, ':0$1'));
+
+      if(delta.milliseconds !== this.tickedMilliseconds) {
+        const now = Date.now();
+        this.lastMillisecondsUpdate = now;
+        // an update that says what this client would have written itself leaves its base alone, so
+        // that the part of the interval that has not elapsed yet survives; anything else - a routine
+        // adding time, another client with a clock of its own - becomes the new base
+        if(delta.milliseconds !== this.millisecondsAt(now))
+          this.anchorTicking(delta.milliseconds);
+      }
     }
 
-    if(this.interval && (delta.paused !== undefined || delta.precision !== undefined)) {
+    if(delta.paused !== undefined || delta.precision !== undefined || delta.countdown !== undefined) {
       this.stopTimer();
-      if(!this.get('paused'))
-        this.startTimer();
+      this.updateTicking();
     }
   }
 
   applyInitialDelta(delta) {
     super.applyInitialDelta(delta);
-    if(delta.paused === false && activePlayers.length == 1)
-      this.startTimer();
+    this.updateTicking();
   }
 
   applyRemove() {
     super.applyRemove();
-    if(this.interval) {
-      console.log('remove clear');
-      this.stopTimer();
-    }
+    this.stopTimer();
   }
 
   classes(includeTemporary=true) {
@@ -85,17 +104,24 @@ export class Timer extends Widget {
     return getSVG(this.get('image'), replaces, _=>this.domElement.style.cssText = this.css());
   }
 
+  // whole intervals that have passed since the base, and the value they add up to - deriving it
+  // from the wall clock instead of adding one interval per tick is what keeps a throttled or
+  // suspended tab from falling behind: however many intervals its browser skipped, the next tick
+  // lands on the time that really passed
+  intervalsSince(now) {
+    return Math.floor((now - this.tickTime)/this.getPrecision());
+  }
+
+  millisecondsAt(now) {
+    return this.tickMilliseconds + Math.max(this.intervalsSince(now), 0)*this.getPrecision()*(this.get('countdown') ? -1 : 1);
+  }
+
   async onPropertyChange(property, oldValue, newValue) {
     await super.onPropertyChange(property, oldValue, newValue);
 
     if(property == 'milliseconds') {
       const end = timeToMS(this.get('end'));
       await this.set('alert', end !== null && ((this.get('countdown') && newValue<=end) || (!this.get('countdown') && newValue>=end)));
-    }
-
-    if(property == 'paused' && newValue !== true) {
-      this.stopTimer();
-      this.startTimer();
     }
   }
 
@@ -118,7 +144,19 @@ export class Timer extends Widget {
   }
 
   async tick() {
-    await this.set('milliseconds', this.get('milliseconds') + this.getPrecision()*(this.get('countdown') ? -1 : 1));
+    const now = Date.now();
+    const intervals = this.intervalsSince(now);
+    if(intervals < 1)
+      return;
+    // lastMillisecondsUpdate only moves for values somebody else wrote, so a client that had to
+    // take over keeps the timer running until the one that should be writing shows up again
+    if(!isPrimarySession() && now - this.lastMillisecondsUpdate < this.getPrecision() + timerTakeoverGrace)
+      return;
+
+    this.tickedMilliseconds = this.millisecondsAt(now);
+    this.tickTime += intervals*this.getPrecision();
+    this.tickMilliseconds = this.tickedMilliseconds;
+    await this.set('milliseconds', this.tickedMilliseconds);
   }
 
   async setPaused(mode) {
@@ -131,11 +169,24 @@ export class Timer extends Widget {
   }
 
   startTimer() {
-    this.interval = setInterval(_=>this.tick(), this.getPrecision());
+    // a preview of a widget outside the room (the editor renders those) must not write state
+    if(this.isReadonlyCopy)
+      return;
+    this.anchorTicking(this.get('milliseconds'));
+    // checking at least once a second keeps a takeover from a stalled client from being delayed by
+    // a long precision - the value itself only ever moves in whole intervals
+    this.interval = setInterval(_=>this.tick(), Math.min(this.getPrecision(), 1000));
   }
 
   stopTimer() {
     clearInterval(this.interval);
     delete this.interval;
+  }
+
+  updateTicking() {
+    if(this.get('paused'))
+      this.stopTimer();
+    else if(!this.interval)
+      this.startTimer();
   }
 }
--- a/client/js/widgets/timer.js
+++ b/client/js/widgets/timer.js
@@ -242,6 +242,10 @@ export class Timer extends Widget {
   // tick that writes nothing - for the moment two clients write, the other one got there first -
   // must not leave that cause behind for whatever the player does next.
   async writeTick(milliseconds) {
+    // a widget of this id that is not this one is a copy the editor renders - a drag preview builds
+    // itself as a widget of its own - and nothing it derives may be written to the room
+    if(widgets.get(this.id) !== this)
+      return;
     this.tickMilliseconds = milliseconds;
     if(milliseconds === this.get('milliseconds'))
       return;
--- a/client/js/widgets/timer.js
+++ b/client/js/widgets/timer.js
@@ -195,10 +195,14 @@ export class Timer extends Widget {
       // "milliseconds == 100" has to be given that value - so a timer something watches moves on by
       // one interval per tick however many intervals its browser skipped. The ones it skipped are
       // time the timer did not count, as they have always been, rather than values it jumps over.
-      // Its base never trails the clock by more than the interval just taken, so that the tick after
-      // a frozen tab wakes up is paced by the precision again instead of racing through a backlog.
-      this.tickTime = Math.max(this.tickTime, now - this.getPrecision()) + this.getPrecision();
-      await this.writeTick(this.tickMilliseconds + this.intervalStep());
+      // The base moves on by exactly the interval that was taken, so that the jitter of the
+      // browser's callbacks - which are never early, only late - can not drop a tick. It may trail
+      // the clock by two whole intervals before it is pulled along, which is far beyond that jitter
+      // and keeps a tab that has just woken up from racing through the backlog.
+      this.tickTime = Math.max(this.tickTime, now - 2*this.getPrecision()) + this.getPrecision();
+      // measured from the value the timer actually holds, not from the base the wall clock is
+      // derived from: taking over from another client means carrying on from what it last wrote
+      await this.writeTick(this.get('milliseconds') + this.intervalStep());
     } else {
       // nothing in the room can tell which values it passed through, so it lands straight on the
       // time that really passed instead of falling behind by every interval the browser skipped
--- a/client/js/widgets/timer.js
+++ b/client/js/widgets/timer.js
@@ -1,9 +1,11 @@
 // A running timer is written by one client only so that its shared millisecond count moves on once
-// per interval no matter how many people watch it. Normally that is the primary session, but a tab
-// that its browser has frozen or throttled stops writing without disconnecting, so every other
-// client takes over once the value has not moved for a whole interval plus this grace period. For
-// the moment two of them write, both derive the same number from the same wall clock, so the count
-// stays true - but each of them runs the timer's routines for the tick it writes.
+// per interval no matter how many people watch it. That client is the one whose player started the
+// timer, so that the timer's routines keep running for that player. Anybody else steps in only once
+// the value has stood still for a whole interval plus this grace period - a tab that its browser has
+// frozen or throttled stops writing without disconnecting - and a timer nobody in the room started
+// goes to the primary session straight away. For the moment two clients write, both derive the same
+// number from the same wall clock, so the count stays true - but each of them runs the timer's
+// routines for the tick it writes.
 const timerTakeoverGrace = 3000;
 
 export class Timer extends Widget {
@@ -56,6 +58,14 @@ export class Timer extends Widget {
       }
     }
 
+    // a timer that stops or starts puts the writing up for election again, and a start goes to the
+    // client whose own change it is - the one whose player started the timer
+    if(delta.paused !== undefined) {
+      this.startedTicking = delta.paused === false && this.locallyPausedTo === false;
+      this.startedWhileHere = delta.paused === false;
+      delete this.locallyPausedTo;
+    }
+
     if(delta.paused !== undefined || delta.precision !== undefined || delta.countdown !== undefined) {
       this.stopTimer();
       this.updateTicking();
@@ -64,6 +74,9 @@ export class Timer extends Widget {
 
   applyInitialDelta(delta) {
     super.applyInitialDelta(delta);
+    // a timer that was already running when this client got the room was started by nobody who is
+    // here, so there is no client to leave the writing to
+    delete this.startedWhileHere;
     this.updateTicking();
   }
 
@@ -117,8 +130,13 @@ export class Timer extends Widget {
     return Math.floor((now - this.tickTime)/this.getPrecision());
   }
 
+  // what one interval adds to the value - a countdown subtracts it
+  intervalStep() {
+    return this.getPrecision()*(this.get('countdown') ? -1 : 1);
+  }
+
   millisecondsAt(now) {
-    return this.tickMilliseconds + Math.max(this.intervalsSince(now), 0)*this.getPrecision()*(this.get('countdown') ? -1 : 1);
+    return this.tickMilliseconds + Math.max(this.intervalsSince(now), 0)*this.intervalStep();
   }
 
   async onPropertyChange(property, oldValue, newValue) {
@@ -162,30 +180,83 @@ export class Timer extends Widget {
     const intervals = this.intervalsSince(now);
     if(intervals < 1)
       return;
-    // lastMillisecondsUpdate only moves for values somebody else wrote, so a client that had to
-    // take over keeps the timer running until it leaves and the primary session picks the job up
-    // again. Waiting out the grace period only keeps this client from writing, not from showing the
-    // time - the value comes from the wall clock, so a stalled writer freezes nobody's display
-    if(!isPrimarySession() && now - this.lastMillisecondsUpdate < this.getPrecision() + timerTakeoverGrace) {
-      this.renderMilliseconds(this.millisecondsAt(now));
+    if(!this.writesTicks(now)) {
+      // leaving the writing to somebody else does not keep this client from showing the time: the
+      // value comes from the wall clock, so a stalled writer freezes nobody's display. A timer
+      // something watches does not catch up, though, so the clock would run its display ahead of the
+      // value it is going to take - that one shows what it was last told.
+      if(!this.valueIsWatched())
+        this.renderMilliseconds(this.millisecondsAt(now));
       return;
     }
 
-    // consecutive ticks share one undo entry instead of filling the protocol one second at a time
+    if(this.valueIsWatched()) {
+      // a routine can be watching for an exact value - a countdown that changes the turn at
+      // "milliseconds == 100" has to be given that value - so a timer something watches moves on by
+      // one interval per tick however many intervals its browser skipped. The ones it skipped are
+      // time the timer did not count, as they have always been, rather than values it jumps over.
+      // Its base never trails the clock by more than the interval just taken, so that the tick after
+      // a frozen tab wakes up is paced by the precision again instead of racing through a backlog.
+      this.tickTime = Math.max(this.tickTime, now - this.getPrecision()) + this.getPrecision();
+      await this.writeTick(this.tickMilliseconds + this.intervalStep());
+    } else {
+      // nothing in the room can tell which values it passed through, so it lands straight on the
+      // time that really passed instead of falling behind by every interval the browser skipped
+      this.tickTime += intervals*this.getPrecision();
+      await this.writeTick(this.tickMilliseconds + intervals*this.intervalStep());
+    }
+  }
+
+  // whether anything in the room is given every value the timer passes through: a routine on the
+  // timer itself, or another widget listening for updates
+  valueIsWatched() {
+    return Array.isArray(this.get('millisecondsChangeRoutine')) || Array.isArray(this.get('changeRoutine'))
+        || (StateManaged.globalUpdateListeners.milliseconds || []).length > 0
+        || (StateManaged.globalUpdateListeners['*'] || []).length > 0;
+  }
+
+  // Which client writes the timer. The one whose player started it does, so that the timer's
+  // routines keep running for that player.
+  writesTicks(now) {
+    if(this.startedTicking)
+      return true;
+
+    const grace = this.getPrecision() + timerTakeoverGrace;
+    const silence = now - this.lastMillisecondsUpdate;
+    // a timer that was started while this client was in the room belongs to the client that started
+    // it, and as long as that one keeps writing nobody else does
+    if(this.startedWhileHere && silence < grace)
+      return false;
+    // a timer nobody here started - one that was already running when this client got the room - and
+    // one whose writer has fallen silent go to the primary session, so that exactly one client takes
+    // them over. The rest of the room waits out a second grace period, which only ever elapses if
+    // the primary session is itself the one that stalled.
+    return isPrimarySession() || silence >= 2*grace;
+  }
+
+  // Consecutive ticks share one undo entry instead of filling the protocol one second at a time. A
+  // tick that writes nothing - for the moment two clients write, the other one got there first -
+  // must not leave that cause behind for whatever the player does next.
+  async writeTick(milliseconds) {
+    this.tickMilliseconds = milliseconds;
+    if(milliseconds === this.get('milliseconds'))
+      return;
+    this.tickedMilliseconds = milliseconds;
     setDeltaCause('timer ticked');
-    this.tickedMilliseconds = this.millisecondsAt(now);
-    this.tickTime += intervals*this.getPrecision();
-    this.tickMilliseconds = this.tickedMilliseconds;
-    await this.set('milliseconds', this.tickedMilliseconds);
+    await this.set('milliseconds', milliseconds);
   }
 
   async setPaused(mode) {
-    if(mode == 'pause' || mode == 'reset')
-      await this.set('paused',  true);
-    else if(mode == 'start')
-      await this.set('paused',  false);
-    else
-      await this.set('paused',  !this.get('paused'));
+    const paused = mode == 'pause' || mode == 'reset' ? true : mode == 'start' ? false : !this.get('paused');
+
+    // Starting a timer here makes this the client that writes it, so that its routines run for the
+    // player who started it rather than for whoever the room's oldest connection belongs to. The
+    // claim is noted for applyDeltaToDOM to pick up rather than made here, because the change comes
+    // back either from inside set() or, for a routine, only when its batch of changes ends - a
+    // change that is already the current value sends nothing, so it must not leave a claim behind.
+    if(this.get('paused') !== paused)
+      this.locallyPausedTo = paused;
+    await this.set('paused', paused);
   }
 
   startTimer() {
--- a/client/js/widgets/timer.js
+++ b/client/js/widgets/timer.js
@@ -1,7 +1,9 @@
 // A running timer is written by one client only so that its shared millisecond count moves on once
 // per interval no matter how many people watch it. Normally that is the primary session, but a tab
 // that its browser has frozen or throttled stops writing without disconnecting, so every other
-// client takes over once the value has not moved for a whole interval plus this grace period.
+// client takes over once the value has not moved for a whole interval plus this grace period. For
+// the moment two of them write, both derive the same number from the same wall clock, so the count
+// stays true - but each of them runs the timer's routines for the tick it writes.
 const timerTakeoverGrace = 3000;
 
 export class Timer extends Widget {
@@ -24,6 +26,10 @@ export class Timer extends Widget {
       start: 0,
       end: null
     });
+
+    // the wall clock base exists from the start, so that the value can be derived before the first
+    // update anchors it
+    this.anchorTicking(this.get('milliseconds'));
   }
 
   // the value the elapsed time is measured from and the moment it was current. The last update
@@ -37,8 +43,7 @@ export class Timer extends Widget {
   applyDeltaToDOM(delta) {
     super.applyDeltaToDOM(delta);
     if(delta.milliseconds !== undefined) {
-      const s = Math.floor(Math.abs(delta.milliseconds)/1000);
-      setText(this.domElement, `${delta.milliseconds < 0 ? '-' : ''}${Math.floor(s/60)}:${Math.floor(s%60)}`.replace(/:(\d)$/, ':0$1'));
+      this.renderMilliseconds(delta.milliseconds);
 
       if(delta.milliseconds !== this.tickedMilliseconds) {
         const now = Date.now();
@@ -125,6 +130,11 @@ export class Timer extends Widget {
     }
   }
 
+  renderMilliseconds(milliseconds) {
+    const s = Math.floor(Math.abs(milliseconds)/1000);
+    setText(this.domElement, `${milliseconds < 0 ? '-' : ''}${Math.floor(s/60)}:${Math.floor(s%60)}`.replace(/:(\d)$/, ':0$1'));
+  }
+
   async setMilliseconds(value, mode) {
     let ms = timeToMS(this.get('start'));
 
@@ -149,10 +159,16 @@ export class Timer extends Widget {
     if(intervals < 1)
       return;
     // lastMillisecondsUpdate only moves for values somebody else wrote, so a client that had to
-    // take over keeps the timer running until the one that should be writing shows up again
-    if(!isPrimarySession() && now - this.lastMillisecondsUpdate < this.getPrecision() + timerTakeoverGrace)
+    // take over keeps the timer running until it leaves and the primary session picks the job up
+    // again. Waiting out the grace period only keeps this client from writing, not from showing the
+    // time - the value comes from the wall clock, so a stalled writer freezes nobody's display
+    if(!isPrimarySession() && now - this.lastMillisecondsUpdate < this.getPrecision() + timerTakeoverGrace) {
+      this.renderMilliseconds(this.millisecondsAt(now));
       return;
+    }
 
+    // consecutive ticks share one undo entry instead of filling the protocol one second at a time
+    setDeltaCause('timer ticked');
     this.tickedMilliseconds = this.millisecondsAt(now);
     this.tickTime += intervals*this.getPrecision();
     this.tickMilliseconds = this.tickedMilliseconds;
--- a/client/js/widgets/timer.js
+++ b/client/js/widgets/timer.js
@@ -212,7 +212,11 @@ export class Timer extends Widget {
   }
 
   // whether anything in the room is given every value the timer passes through: a routine on the
-  // timer itself, or another widget listening for updates
+  // timer itself, or another widget listening for updates. The listener registries are room-wide -
+  // a globalUpdateRoutine is told about every widget - so one of them anywhere in the game puts
+  // every timer on the one-interval-per-tick path. That is deliberate: no value may be skipped for
+  // a routine that might be looking for it, and the registries say nothing about which widget an
+  // update came from.
   valueIsWatched() {
     return Array.isArray(this.get('millisecondsChangeRoutine')) || Array.isArray(this.get('changeRoutine'))
         || (StateManaged.globalUpdateListeners.milliseconds || []).length > 0
@@ -222,11 +226,18 @@ export class Timer extends Widget {
   // Which client writes the timer. The one whose player started it does, so that the timer's
   // routines keep running for that player.
   writesTicks(now) {
+    const grace = this.getPrecision() + timerTakeoverGrace;
+    const silence = now - this.lastMillisecondsUpdate;
+    // a tab that comes back from being frozen to find the value moving without it gives up its
+    // claim, so that it does not put another interval on top of what the client that took over
+    // writes. It takes both to lose the claim - having stopped writing for long enough to be taken
+    // over, and having been overtaken since - so a one-off change from somebody else, a routine
+    // adding time, leaves the writing where it is.
+    if(this.startedTicking && now - this.lastTickWrite >= grace && this.lastMillisecondsUpdate > this.lastTickWrite)
+      delete this.startedTicking;
     if(this.startedTicking)
       return true;
 
-    const grace = this.getPrecision() + timerTakeoverGrace;
-    const silence = now - this.lastMillisecondsUpdate;
     // a timer that was started while this client was in the room belongs to the client that started
     // it, and as long as that one keeps writing nobody else does
     if(this.startedWhileHere && silence < grace)
@@ -247,6 +258,7 @@ export class Timer extends Widget {
     if(widgets.get(this.id) !== this)
       return;
     this.tickMilliseconds = milliseconds;
+    this.lastTickWrite = Date.now();
     if(milliseconds === this.get('milliseconds'))
       return;
     this.tickedMilliseconds = milliseconds;
@@ -272,6 +284,9 @@ export class Timer extends Widget {
     if(this.isReadonlyCopy)
       return;
     this.anchorTicking(this.get('milliseconds'));
+    // the moment this client last wrote a tick itself - it starts out as the moment it started
+    // ticking, so that a claim on the writing is not dropped before the first tick
+    this.lastTickWrite = Date.now();
     // checking at least once a second keeps a takeover from a stalled client from being delayed by
     // a long precision - the value itself only ever moves in whole intervals
     this.interval = setInterval(_=>this.tick(), Math.min(this.getPrecision(), 1000));
--- a/client/js/widgets/timer.js
+++ b/client/js/widgets/timer.js
@@ -130,9 +130,13 @@ export class Timer extends Widget {
     }
   }
 
+  // below an hour the value reads as it always has, m:ss with unbounded minutes; from an hour on it
+  // gets an hours field, so that a timer somebody left running says 1:00:00 instead of 60:00
   renderMilliseconds(milliseconds) {
     const s = Math.floor(Math.abs(milliseconds)/1000);
-    setText(this.domElement, `${milliseconds < 0 ? '-' : ''}${Math.floor(s/60)}:${Math.floor(s%60)}`.replace(/:(\d)$/, ':0$1'));
+    const pad = value => String(value).padStart(2, '0');
+    const time = s < 3600 ? `${Math.floor(s/60)}:${pad(s%60)}` : `${Math.floor(s/3600)}:${pad(Math.floor(s/60)%60)}:${pad(s%60)}`;
+    setText(this.domElement, `${milliseconds < 0 ? '-' : ''}${time}`);
   }
 
   async setMilliseconds(value, mode) {
--- a/library/tutorials/Types - Timer/0.json
+++ b/library/tutorials/Types - Timer/0.json
@@ -25,7 +25,7 @@
       "time": "",
       "attribution": "The clock face, by temas, is in the Public Domain under CC0, and is available at https://openclipart.org/detail/134335/diy-clock-face-by-temas-134335.\n\nThe stopwatch face is the based on the clock face, but the numbers were modified by LawDawg96.  It is released to the Public Domain under CC0.\n\nThe hour and minute hands are in the Public Domain under CC0, and are available at https://freesvg.org/minute-hand-mitchell-joh.",
       "showName": false,
-      "lastUpdate": 1787287344505,
+      "lastUpdate": 1787323839668,
       "skill": "",
       "description": "",
       "similarImage": "",
@@ -42,7 +42,7 @@
     }
   },
   "ekxj": {
-    "text": "This tutorial demonstrates the basic operations of the timer widget. A running timer belongs to the room, not to the player who started it: it keeps running when that player closes their tab, and another player carries it on. Time passes only while somebody is in the room; a tab left in the background shows the time that really passed when you come back. In the JSON editor, timers take either milliseconds or minutes:seconds for start and end as well as for the value and seconds parameters of the TIMER function.",
+    "text": "This tutorial demonstrates the basic operations of the timer widget. A running timer belongs to the room, not to the player who started it: it keeps running when that player closes their tab, and another player carries it on. Time passes only while somebody is in the room; a tab left in the background shows the time that really passed when you come back. In the JSON editor, timers take milliseconds, minutes:seconds or hours:minutes:seconds for start and end as well as for TIMER's value and seconds parameters.",
     "css": "font-size: 20px",
     "height": 50,
     "width": 900,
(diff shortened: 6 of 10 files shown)


gitran:
I've been waiting on this one! Doing playtests with a session clock that isn't reliable is pretty annoying!

VTT-AI-Agent (commented):
## TL;DR

Solid fix — the wall-clock derivation plus "lowest sessionID writes" is the right shape, and I reproduced all three headline claims in a real browser (handover on tab close, one writer for two tabs of the same player, no time passing in an empty room). Nothing here looks like a correctness bug. The one thing I'd want changed before merge is the **visible 4-second freeze during a takeover**: with the primary tab blocked, every client's clock sat at `0:27` for 4.4 s and then jumped to `0:31`, even though every client already knew the right value locally.

## Findings

### 1. The takeover grace is visible as a frozen clock on every client (`client/js/widgets/timer.js:146-160`)

`tick()` returns early on a non-primary client, so a client that is not writing only ever refreshes its display from an incoming `milliseconds` delta (`applyDeltaToDOM`). While the writer is stalled, nobody updates the DOM, so the timer *looks* stopped for `getPrecision() + timerTakeoverGrace` before it jumps.

Measured (two Chromium contexts on this branch, primary's main thread blocked for 9 s, timer delta frames captured off the websocket):

```
A  7.4s  ms=27000     <- primary's last write, then blocked
   ...   (4.4s with no write at all; both clients display 0:27)
B 11.8s  ms=31000     <- B takes over, display jumps 0:27 -> 0:31
```

Four seconds of a dead clock followed by a jump is exactly the symptom #1655 complains about, just shorter. It's avoidable: every client already computes the authoritative value with `millisecondsAt(now)`, so the non-writing branch can refresh its own `domElement` text and leave the `set()` to the writer — e.g. factor the `setText(...)` out of `applyDeltaToDOM` into a `renderMilliseconds(ms)` and call it from the early-return paths in `tick()`. The single-writer rule and the state stay exactly as they are; only the pixels stop lying. (Trade-off worth stating explicitly: the display would briefly show a value that is not yet in the room state. Given the alternative is showing a value that is *wrong*, I think that's the better side to err on, but it's your call.)

### 2. `isPrimarySession()` — the half of the fix that decides *who* ticks — has no test (`client/js/overlays/players.js:22-30,133`)

`tests/client/timer-tick.test.js:19` stubs `globalThis.isPrimarySession`, so the new `primarySessionID` reduce in `fillPlayerList` is never exercised by anything. That line carries the whole correctness argument for "exactly one client writes", and it has two silent failure modes:

- if `sessions` ever stops being `[{sessionID, player}]` (or `sessionID` stops being the numeric `Player.nextSessionID++` — `<` would become a lexicographic compare, `"10" < "9"`), the picked session changes without any test noticing;
- the `primarySessionID === null` fallback makes *every* client consider itself primary, so a regression that leaves it null degrades to N writers, which is precisely the double-count bug the PR is fixing — and it degrades quietly, since the derived-value design keeps the number correct.

A ~15-line jest test over `fillPlayerList` (lowest sessionID wins; the null fallback; the "same player, two sessions" list picks one) would cover it. Related gap: nothing covers the "two connections of one player" case the PR body claims to fix — I verified it manually (two tabs as `Same`: 9 timer deltas from tab 1, 0 from tab 2), but it isn't pinned down anywhere.

### 3. During a writer overlap the *value* is idempotent, but the *routines* are not (`client/js/widgets/timer.js:159`)

The comment at line 3 and the PR body both argue that an overlap "writes the same number twice rather than counting it twice". True for `milliseconds` — `StateManaged.set()` short-circuits on an equal value (`client/js/statemanaged.js:110`). But that short-circuit only fires if the other client's delta has already landed. If the two clients' tick phases happen to fall within one RTT of each other, both `set()` calls are real writes, and `set()` runs `millisecondsChangeRoutine`, `changeRoutine` and any `globalUpdateRoutine` listener locally on each writer — so a timer whose `millisecondsChangeRoutine` decrements a counter, moves a piece or plays a sound would fire it twice for that tick, on every tick, for as long as the phases stay aligned.

In my run the phases happened to fall the other way (after unfreezing, the primary's writes were all no-ops and it never resumed writing at all), so I did not observe it — hence "possible" rather than "confirmed". It's narrow, but the comment currently promises more safety than the code delivers; at minimum I'd reword it to say the *value* is idempotent and side-effecting routines are not.

Related, and worth a line in the comment: once a non-primary client has taken over it effectively never hands back, because the primary's update matching its own `tickedMilliseconds` suppresses the `lastMillisecondsUpdate` refresh (line 43). Harmless — the values agree and the loser's `set()` is a no-op — and when the taker-over leaves the primary resumes at once, but it isn't the "the primary takes the job back" behaviour a reader would infer from the comment.

### 4. Demo room isn't at the quality bar the standing rule asks for (`Timer check.vtt`)

Right call on the important half: nothing under `library/` is touched, and the room lives on the PR's `pr-test-ai` server. Thanks for that. But `docs/tutorials.md` ("Test rooms are tutorial-quality") also asks the room itself to match tutorial conventions, and this one doesn't:

- **`_meta.info` parity is a hard requirement and this misses ~10 fields.** It has `mode: "coop"`, `time: "2"`, `players: "1-8"`, `language: "EN"`, and omits `showName`, `skill`, `description`, `similarImage/Name/Designer/Awards`, `ruleText`, `helpText`, `variantImage`. Tutorial parity is `mode: "Tutorial"`, `time: "0"`, `players: "1"`, `language: "en-US"`, `showName: false`, and the full field list.
- **Naming isn't in the category scheme.** `library/tutorials/Types - Timer` already exists — the natural home is a variant of it (`_meta.info.variant: "Multi-client"` or similar) rather than a standalone `"Timer check"`.
- **Structure deviates from the newest tutorials' band layout**: `title` at `y:60` (convention is `x:400,y:-2,width:800,height:60`, 60px centered), `subtitle` at `layer:-2` where the convention is an `overview` widget at `x:108,y:80,width:1400,height:150,layer:-3` / 25px, and `howto1..5` where captions are `text1..n` at `layer:-3`. The `what to try` checklist is genuinely useful content — it just wants to be the `overview` paragraph plus per-example captions.
- The library image is fine: a roughly-square PNG screenshot is exactly what the doc prescribes for a test-server room (the square-SVG badge rule only applies once something moves into `library/tutorials/`).

Content-wise the room is well chosen (count-up, countdown with an alert, `+ 1 minute` while running) and 16 widgets is comfortably inside the one-room budget, so this is presentation, not scope.

### 5. Minor

- **Undo protocol** (`client/js/serverstate.js:394-425`): every tick delta becomes its own undo entry because `delta.c` is unset, so with a timer running, edit-mode undo fills with timer ticks. Pre-existing, but this PR makes ticking happen reliably on every client where it previously often didn't start at all after a load, so it will be more noticeable. A `setDeltaCause('timer')` in `tick()` would collapse consecutive ticks into one entry via the existing same-cause merge at line 414.
- **`millisecondsAt()` returns `NaN` before the first anchor** (`timer.js:115`), because `tickTime`/`tickMilliseconds` are undefined. Today that's benign — it just forces a re-anchor, and `applyInitialDelta`'s defaults pass anchors immediately — but it's an implicit ordering dependency. Initialising both in the constructor next to `addDefaults` would make it explicit.
- Flipping `countdown` mid-run goes through `stopTimer()` + `startTimer()`, which re-anchors from `Date.now()` and so silently drops the un-elapsed part of the current interval. Sub-second, almost certainly fine, just noting it's a deliberate consequence of the restart.

## What I verified

- `npm test` on this branch: **1357 pass**, including the 10 new `tests/client/timer-tick.test.js` cases. All 45 CI checks (incl. CodeQL) are green.
- Real Chromium, multi-client, against a local server on this branch:
  - **Handover on close** — two clients, closed the older one: `0:08` → `0:14` over the next 6 s, no gap, no lost second.
  - **Two tabs, same player name** — 9 timer deltas sent by tab 1, **0** by tab 2 over 8 s; display correct at `0:09`.
  - **Empty room** — everyone left for 7 s, a fresh client found the timer still at `0:09`.
  - **Blocked primary** — takeover works and the value stays true to the wall clock (no double count: 19.3 s of wall time produced exactly 19000 ms), but with the 4.4 s display freeze described in finding 1.
- Confirmed the bundling assumption holds: `client/js/overlays/players.js` is concatenated before `client/js/widgets/timer.js` in `server/minify.mjs`, `export` is stripped for the client bundle, and `isPrimarySession` is a hoisted function declaration — same pattern as the `activePlayers` reference it replaces.
- Confirmed every widget-preview path (`domhelpers.js:383`, the ~10 `properties.js`/`deckeditor.js` call sites) funnels through `renderReadonlyCopyRaw`, which sets `isReadonlyCopy` *before* `applyInitialDelta` — so the new `startTimer()` guard covers all of them.
- Confirmed no validator / `jeCommands` / widget-editor work is owed: no widget property or routine parameter is added, renamed or removed, so the PR body's claim holds and there is no new-UI-support gap.

I did not add the `beta` label: `client/js/widgets/timer.js` changes how the timer widget behaves by default for every existing game — who writes it, how the value is derived and when ticking starts — which is a real engine change, not an incidental one, so this PR is excluded from the beta auto-deployment by design.

VTT-AI-Agent:
Thanks - the frozen-clock measurement was the useful one, that is fixed now. Everything below is in `fc13cbb3`.

### 1. Frozen clock during a takeover - fixed

The `setText` half of `applyDeltaToDOM` is now a `renderMilliseconds(ms)` of its own, and the grace-period branch of `tick()` calls it with `millisecondsAt(now)` before returning. A client that is waiting to see whether the writer comes back keeps showing the time; only the writing stays with whoever owns it.

Measured the same way as your run (two Chromium contexts, the writer's main thread blocked for 9 s, sampling what the other one displays):

```
before   0.0s 0:03 ... 5.8s 0:08     longest the display stood still: 5728 ms
after    0.0s 0:03  1.1s 0:04  2.1s 0:05  3.2s 0:06  4.2s 0:07  5.4s 0:08 ...
                                     longest the display stood still: 1213 ms  (= one sampling interval)
```

The trade-off you named is real and I took the side you suggested: for up to `precision + 3 s` a non-writing client shows a value that is not in the room state yet. It is the value the state is about to get, and the alternative was showing one that is wrong. New jest case `its display follows the clock while it waits for the updates to come back` pins it down: `set()` is not called, `milliseconds` is still 3000, and the DOM reads `0:05`.

### 2. `isPrimarySession()` untested - partly fixed

The reduce is now `primarySessionOf(sessions)` in `players.js`, and `tests/client/primary-session.test.js` covers what you named: lowest `sessionID` wins regardless of order, `10` vs `9` compares numerically (a lexicographic compare would fail it), two sessions of the same player pick the lower one, and an empty or missing list yields `null` - plus that `isPrimarySession()` is `true` while the list is unknown, which is the "everyone writes" fallback.

Not covered: the `primarySessionID === mySessionID` comparison itself, because `mySessionID` is module state that only the server's `sessionID` message sets, and `fillPlayerList` needs the whole player-list DOM. I did not want to add test-only exports for a one-line equality. The multi-client behaviour it produces (two tabs of one player, one writer) is still only covered by hand - reproduced again here: 11 timer deltas from the first tab, 0 from the second.

### 3. Routines are not idempotent during an overlap - comment corrected

You are right, and the code cannot honestly promise more without the clients coordinating. The header comment now says what actually holds:

> For the moment two of them write, both derive the same number from the same wall clock, so the count stays true - but each of them runs the timer's routines for the tick it writes.

The comment in `tick()` no longer suggests the primary takes the job back on its own either; it says the client that took over keeps it until it leaves.

### 4. Demo room - rebuilt to the tutorial conventions

Same file name, so it replaces the old upload in `pr-test-ai`. It is now `Types - Timer`, variant `Multi-client`, i.e. a variant of the existing tutorial rather than a game of its own, with `_meta.info` mirroring `library/tutorials/Types - Timer/0.json` field for field (`mode: "Tutorial"`, `time: "0"`, `players: "1"`, `language: "en-US"`, `showName: false`, and the `skill`/`description`/`similar*`/`ruleText`/`helpText`/`variantImage` fields that were missing). Structure follows the band layout: `title` at `x:400,y:-2,width:800,height:60` in 60px centered, an `overview` at `x:108,y:80,width:1400,height:150,layer:-3` in 25px, and `text1..5` captions at `layer:-3` in 22px under their examples. `node validator/validate_gamefile_node.js` exits 0, and I loaded it from the Game Shelf and played it.

The examples are a plain count-up, a countdown from 2:00 with the alert, a `TIMER inc` button - and a new fourth one that is the interesting one for this PR: a timer whose `millisecondsChangeRoutine` counts its own ticks next to it, so you can see the routine run once per interval and not once per client. `../publish/`, still nothing under `library/`.

### 5. Minor

- **Undo protocol**: `tick()` now calls `setDeltaCause('timer ticked')`. Five ticks became one undo entry labelled `timer ticked` instead of five `unknown` ones - verified in edit mode against `getUndoProtocol()`.
- **`millisecondsAt()` returning `NaN`**: the constructor now calls `anchorTicking(this.get('milliseconds'))`, so the base exists from the start and the ordering is no longer implicit.
- **`countdown` flipped mid-run** dropping the un-elapsed part of the interval: left as is. It is sub-second and it falls out of restarting the timer, which is what changing the direction has always done.

### While I was in there

`library/tutorials/Types - Timer/0.json` tells the reader today: *"one important limitation of the timer is that it is controlled by the player that started the timer. If that person leaves the room or the player's internet connection is disrupted while the timer is running, the timer will stop."* This PR makes that false. Following the draft-tutorial rule I did **not** touch the file; an amended copy is on the test server for a maintainer to accept or discard: [`Types - Timer 0.vtt`](https://agent.virtualtabletop.io/reports/pr/1540126994505666660/tutorials/Types%20-%20Timer%200.vtt).

`npm test` is at 1363 (6 new), and TestCafe `multiclient.js` passes locally.

VTT-AI-Agent (commented):
## TL;DR

UI/UX review (not a code review) of #3139. The in-room UI is essentially untouched and nothing regressed: I ran the timer tutorial with two clients, watched the handover, and swept the overlays at four viewport sizes without finding a break caused by this diff. The one thing worth acting on is a direct consequence of the change: the timer's on-screen format is `M:SS` with no hours and no overflow guard, and this PR is exactly what makes long values common — a tab that slept for three hours now renders `193:07`, and a countdown that ran negative renders `-193:07`, which spills outside the widget's own box. The draft tutorial also needs a spacing fix: its new intro paragraph now sits 2px above the first example and reads as one block with it.

---

### What I actually exercised

Checked out the branch, ran the server locally, loaded `library/tutorials/Types - Timer/0.json` into a room, and drove it with Playwright.

**Viewports:** 390×844 (phone portrait), 844×390 (phone landscape), 820×1180 (tablet), 1280×800 (laptop), 1920×1080 (wide).

**Overlays opened and looked at:** Game Shelf, Players, About, Game details, Add game, Save game, Add variant, Share link, Confirm, Switch game, Symbol picker, Audio picker, Button input, Connection lost — and in edit mode: the editor sidebar / widget property editor (with the Timer section), the "Add a widget" overlay, and the JSON view.
**Could not reach:** the deck-editor sub-dialogs (`libraryDecksOverlay`, `deckEditorExportOverlay`, `deckEditorImportOverlay`, `deckEditorNewDeckOverlay`), `updateImageOverlay`, `linkDetailsOverlay`, `askIDoverlay`, `inputBlockOverlay` — they need content or flows the tutorial room doesn't provide. Caveat: `showOverlay()` isn't reachable from the page scope, so a few of the modal overlays were opened by setting `display:flex` directly, which is faithful for layout but not for the show/hide bookkeeping.

Multi-client behaviour renders correctly: with two tabs the non-primary tab started the timer, both showed the same second, and closing the primary tab left the survivor counting on without a visible gap or a frozen display. The Players overlay is unchanged by the new `primarySessionID` bookkeeping ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1540170854648909834/pr3139/players-overlay-two-connections.png)) — keeping the "who is counting" detail out of the UI is the right call, no need to surface it.

---

## Big layout improvements

### 1. The timer needs an hours format — this PR is what makes it matter

`renderMilliseconds()` (`client/js/widgets/timer.js:133-136`) only ever produces `M:SS`, and minutes are unbounded:

| value | rendered |
|---|---|
| 59 min 59 s | `59:59` |
| 1 hour | `60:00` |
| 3 h 13 m 07 s (a tab that slept) | `193:07` |
| countdown gone negative, same length | `-193:07` |

timer long values overflowing the widget box
*(red dashes = the widget's own 74×30 box)*

Two problems, both now much easier to hit:

- **`60:00` is unreadable.** A first-time user cannot tell whether that is an hour or a minute, and there is nothing on the widget to disambiguate.
- **It overflows.** At the default `width: 74`, the box is 77px of content and `-193:07` measures 81px. Because `.timer` is `text-align: right` with no `overflow` rule (`client/css/widgets/timer.css:1-8`), the extra characters spill out to the **left**, outside the widget — and when `alert` is on, the red top/bottom border rules stop where the box stops, so the underline no longer runs under the whole number.

Before this PR a throttled or sleeping tab simply lost that time, so the number stayed small. The PR's stated user-visible effect is the opposite — "the value can jump forward by the time the tab was away" — so `>1 h` values go from a curiosity to the normal outcome of closing a laptop lid.

Concrete proposal, minimal and non-breaking:

```js
renderMilliseconds(milliseconds) {
  const s = Math.floor(Math.abs(milliseconds)/1000);
  const mmss = `${Math.floor(s/60)%60}:${s%60}`.replace(/:(\d)$/, ':0$1');
  const text = s < 3600 ? mmss : `${Math.floor(s/3600)}:${mmss.replace(/^(\d):/, '0$1:')}`;
  setText(this.domElement, `${milliseconds < 0 ? '-' : ''}${text}`);
}
```

Everything under an hour renders exactly as it does today, so no existing game changes; an hour or more becomes `1:00:00` / `3:13:07`. Pair it with `overflow: hidden; white-space: nowrap;` on `.timer` so nothing can ever paint outside the widget, or let the widget grow — but the hard clip is the safer default given widgets sit in fixed layouts.

---

- **`client/css/widgets/timer.css:1-8`** — `.timer` has `text-align: right` and `padding-right: 3px` but no `overflow`/`white-space`, so any string wider than the widget escapes it (see above). Even the positive `193:07` fills the box literally edge to edge with 0px of breathing room on the left, while `0:00` sits with 30px, so a row of timers at different magnitudes reads as ragged.
- **Property editor, "Current time" field** — while the timer is running, the field is rewritten once a second, so anything you type into it gets shuffled: focusing it and typing `12:3` on a running timer leaves `0:1112:3` in the box. I checked this on `main` too and it is **pre-existing**, not introduced here — but this PR deliberately keeps running timers ticking in more situations (including edit mode with the widget selected), so it will be hit more often. Cheap fix in the same area: skip the write-back while the input has focus, or grey the field out while `paused` is false.
- **Add-widget overlay at 844×390** — the toolbar tooltip *"Add a new widget. Hotkey: a"* stays over the overlay it just opened and covers the overlay's own **"Add a widget"** header ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1540170854648909834/pr3139/add-widget-844x390.png)). Pre-existing, unrelated to this diff, but it's the first thing you see at that height.
- **390×844 portrait** — the Game Shelf's filter bar overflows ~500px to the right with no scroll affordance, and the "Public Library" heading is clipped by the bottom toolbar. Also pre-existing and arguably intentional (the app tells you to rotate at that aspect), just recording what I saw at the size.

No overflow, clipping, collision or unreachable close button found in any overlay at 820×1180, 1280×800 or 1920×1080.

---

- **The jump is unexplained.** The headline behaviour change is that coming back to a backgrounded tab makes the number leap forward. On screen there is no signal that this is catch-up rather than a bug — the timer just shows a different, much larger number than the one you left. Nothing heavyweight is warranted, but this is precisely why the tutorial wording (below) matters: it is the only place a user will learn it.
- **`60:00` reads as ambiguous** for someone who has never seen this widget — the hours format above is the fix.
- The countdown alert state reads well: red digits plus the blinking red rules above and below are unmistakable, and it survives negative values.

---

## Draft tutorial

The `pr-test-ai` room holds two entries, both named **Types - Timer**: the amended `Types - Timer 0.vtt` draft (variant *Basics*, the one this PR proposes) and the `Timer check.vtt` demo (variant *Multi-client*). Note the shelf shows only the name, not the variant, so a maintainer opening the room sees two identical tiles and has to click both to tell them apart.

### `Types - Timer 0.vtt` — variant *Basics* (the draft)

Played end to end at 1280×800; it ships one variant. Only the intro paragraph differs from `library/tutorials/Types - Timer/0.json`, and the new wording is factually right. Three things I'd change before it is accepted:

**1. The new intro collides with the first example (layout).** The label widget's box is 39px tall and the text already overflows it; the rewrite grows the paragraph from 11 rendered lines to 13. Measured on the board:

| | intro text bottom | first example text top | gap |
|---|---|---|---|
| current `0.json` | 189px | 210px | **21px** |
| draft | 208px | 210px | **2px** |

Current:
current tutorial intro

Draft:
draft tutorial intro

The intro and the first example now read as a single run-on paragraph, and the `0:00` timer on the left no longer lines up with the start of a visually separate block. Fix either way: move the first example row and everything below it down by ~25 board px, or trim the intro back to the 11 lines it used to occupy. Trimming is better — see 2.

**2. "Client" is developer jargon in the first paragraph of a Basics tutorial.** *"one of the clients in the room counts it and sends the new value to everybody else"* — "client" appears nowhere else in this tutorial, and the reader doesn't need the sync protocol to use a timer; they need to know it doesn't stop. A shorter version that keeps the meaning, fits the tutorial's voice and lands back at ~11 lines:

> A running timer belongs to the room, not to the player who started it: it keeps running when that player closes their tab, and another player in the room carries the count on. Time only passes while somebody is in the room — the timer waits where it is and continues when the first person comes back.

**3. The other user-visible change is missing.** The draft says nothing about a tab that was in the background or asleep catching up when you return, which is the behaviour a reader most likely to notice on their own. One sentence, e.g. *"A tab that was in the background or asleep shows the time that really passed once you come back to it."*

**4. Stale text the draft leaves alone but the maintainer will be in this file anyway:** *"which can be added from the Add Widget button on the left side of the screen"*. In the current edit-mode UI there is nothing on the left — the add control is the `+` in the **top** toolbar and the **"ADD A NEW WIDGET"** button in the right-hand editor panel. Worth correcting in the same pass. ([draft board](https://agent.virtualtabletop.io/reports/ui-review/1540170854648909834/pr3139/draft-tutorial-board-1280.png))

**5. Merge, don't replace.** The library tutorial has three variants (`0/1/2.json`); the draft carries only *Basics*, so the intended action is to lift the paragraph into `0.json`, not to import the draft as a game.

### `Timer check.vtt` — variant *Multi-client* (the demo the PR links as "Try it")

Not part of the PR, but it is what a reviewer opens first, so: the layout is genuinely good — a 3-up top row and 2-up bottom row, each example with its own caption directly under it, and a separate "Things worth trying" panel. ([screenshot](https://agent.virtualtabletop.io/reports/ui-review/1540170854648909834/pr3139/demo-room-multi-client.png)) Four small things:

- **The intro names a section that doesn't exist.** *"A timer - the 'Functional' section of the Add Widget overlay"*. The overlay's sections are **Cards / Game Pieces / Interactive / Decorative / Custom**, and the timer sample sits under **Interactive**. A first-time user following that sentence will not find it.
- **The bare `0` beside the fourth timer has no label** — until you've read the caption it looks like part of the timer widget. A tiny caption ("ticks") next to it would carry the point on its own.
- Same **"one client counts it"** jargon in the intro as in the draft tutorial.
- The **"Things worth trying with a second tab open"** list is four unbulleted lines that wrap into each other and read as a paragraph. A `•` prefix or a blank line between items would make it scannable, which matters because it's the instruction list the whole demo exists for.

VTT-AI-Agent:
Thanks - the hours finding was the right one to lead with. The format is fixed in `f1c892c5`, the tutorial and the demo room are republished. Point by point.

## 1. Hours format - done

`renderMilliseconds()` keeps `m:ss` below an hour and adds an hours field from an hour on: `60:00` is now `1:00:00`, `193:07` is `3:13:07`, `-193:07` is `-3:13:07`. Below an hour not a pixel changes, so no existing game renders differently.

before and after

The widget editor moved with it: `formatTimerMs`/`parseTimerInput` render and accept `h:mm:ss`, so the **Current time** field of a timer that has run for three hours reads `3:13:07` rather than `193:07`, and Start / End / Current time take `90`, `1:30` or `1:30:00`. New jest cases cover the widget's display (below an hour, from an hour on, negative) and the editor helpers including the round-trip; `npm test` is at 1368.

## 2. The overflow guard - deliberately not done

`overflow: hidden` on `.timer` would not clip the *end* of the value, it would clip the *front*. The element is `text-align: right`, so a line wider than the box overflows past its **start** edge, and start-edge overflow is unreachable and simply cut off. `3:13:07` in a default-width box would render as `13:07` - a wrong time that looks like a right one. For a widget whose entire job is to say what time it is, sticking out of its box is the better failure.

Measured on the branch at the default `74 x 30`, in board pixels (content box 71 after the 3px padding, monospace at 23px):

```
              before        after
 59:59         55           55      fits
 1 hour        55  (60:00)  78      +7 outside
 3:13:07       66 (193:07)  78      +7 outside
 -3:13:07      78 (-193:07) 89     +18 outside
```

So the hours form is one to two characters wider in exactly the range that already did not fit, and what it buys is that the number is readable at all. A game that gives the timer room has no issue - the demo room's timers are 180 wide and show `2:00:02` with air to spare - and a game that wants a hard clip can already put `overflow: hidden` in the widget's own `css`.

If you would still rather have the clip in the stylesheet I will add it (one line, plus `white-space: nowrap`); I just did not want to ship a timer that can show the wrong number. Same answer for the ragged-left-edge nitpick: it comes from right-aligning a value whose width varies, and every fix for it (centring, clipping, a fixed-width box) trades away either the alignment or the digits.

## 3. "Current time" rewritten while typing - could not reproduce

`renderTimerTimeInput`'s property listener already skips the write-back while the input has focus (`client/js/editor/sidebar/properties.js:9147-9153`, on `main` as well). On this branch I clicked into **Current time** on a timer running at 1 s precision and typed `12:3` with 400 ms between keystrokes, so four ticks landed mid-typing: the field held exactly `12:3` throughout and the timer took `12:03` on Enter. If you have a repro (multi-selection, or a re-render coming from another module?) I will chase it.

## 4. Add-widget tooltip at 844x390, Game Shelf at 390x844

Both reproduce on `main`, neither is touched by this diff, and fixing them means going into the toolbar and shelf layout - out of scope here. Worth their own issue.

## 5. The jump is unexplained

Agreed that the tutorial is the only place a player learns it, which is what the wording below is for.

## Draft tutorial - all five points

Republished as [Types - Timer 0.vtt](https://agent.virtualtabletop.io/reports/pr/1540126994505666660/tutorials/Types%20-%20Timer%200.vtt) (same file name, so it replaces the previous draft in `pr-test-ai`).

1. **Spacing.** The new intro is 516 characters against the library file's 515, so it wraps to the same 11 lines. Measured on the board the same way for both: gap under the intro **27px in the library file, 27px in the draft**; gap under the first example caption **40px and 40px**. The draft is line-for-line the original's layout again (the previous one was 649 characters).
2. **Jargon.** Gone, close to your wording: *"A running timer belongs to the room, not to the player who started it: it keeps running when that player closes their tab, and another player carries it on."*
3. **Catch-up.** Added: *"Time passes only while somebody is in the room; a tab left in the background shows the time that really passed when you come back."*
4. **Stale add-widget text.** Now *"which can be added with the + button in the toolbar at the top of the screen while edit mode is on"*. The other half of that sentence stays as it is, though: outside edit mode there **is** a left toolbar and the Edit Mode (wrench) button is on it, so *"using the Edit Mode button to the left"* is still accurate.
5. **Merge, don't replace.** Yes - it is variant `0` only and meant to be lifted into `0.json`, not imported as a game.

To pay for the added sentence I trimmed the closing JSON sentence, and the first example caption gained one short line about the hours format this PR adds (*"An hour or more reads 1:00:00 rather than 60:00."*) without gaining a line of layout.

draft tutorial board

## Demo room - all four

[Timer check.vtt](https://agent.virtualtabletop.io/reports/pr/1540126994505666660/Timer%20check.vtt), same file name, already in `pr-test-ai`:

- *"Functional"* -> **Interactive**.
- The bare `0` has a **ticks** caption above it.
- *"one client counts it"* -> *"one of the open tabs counts it and sends every new milliseconds value to everybody else"*; the fourth caption lost its "client" too.
- The **Things worth trying** list is bulleted.
- New: a **+ 1 hour** button beside **+ 1 minute**, so the hours display can be tried in the room (two presses give `2:00:00` and it keeps counting from there).

demo room

On the two tiles that look alike: both files carry the game name `Types - Timer` because review 1 asked for the demo to be a variant of the existing tutorial rather than a game of its own, and the shelf tile shows the name only. They are told apart by their preview image - the blue tutorial badge for the *Basics* draft, a screenshot of the board for *Multi-client*. Say the word and I will give the demo a standalone name instead.

VTT-AI-Agent (commented):
## TL;DR

PR #3139 fixes timer drift and handover, but it also changes two established routine semantics in ways that break existing games: timer routines now run as the oldest connected player rather than the player who started the timer, and catch-up jumps skip intermediate values/routine invocations. Those compatibility issues need resolving before merge.

## Findings

### 1. Timer routines now run in the wrong player's context (`client/js/widgets/timer.js:59-61,160-179`)

The remote `paused` delta now starts an interval on every client, and `tick()` allows only the room's primary session to perform the `set('milliseconds', ...)`. Because `StateManaged.set()` executes `millisecondsChangeRoutine` locally, `${playerName}`, `${seatID}`, `playerSeats`, and every other player-local routine value now describe the oldest connected player—not the player whose click/TIMER operation started this timer. Previously, a newly started timer's interval belonged to the client that unpaused it, so existing games can and do rely on that context.

This is already concrete in the public library. `library/games/Ultra/0.json:7729-7739` assigns the generated code cards' `owner` to `${playerName}` when `codeTimer1` finishes; if Bob clicks **Draw Code** while Alice is the oldest session, this branch runs on Alice and makes Bob's private code visible to Alice instead. `library/games/Speedy Fruits/0.json:3203-3221` likewise adds `${playerName}` to `clickCover.owner` after a player's enforcer timer, so it would block the primary player rather than the player whose timer was started. INPUTs, sounds, and seat-relative routines have the same problem.

Please preserve the initiating player's routine context (including across a handover), or protect existing games with the repository's legacy-mode/migration mechanism. The tests currently stub only writer election and never attach a `millisecondsChangeRoutine`, so a two-player test where the non-primary player starts a timer and the routine consumes `${playerName}` would pin this down.

### 2. Catch-up skips timer values that existing routines use as events (`client/js/widgets/timer.js:160-179`)

When several intervals elapsed, `tick()` computes the final wall-clock value and performs one `set()`. Thus a timer advancing from 400 ms to 3,500 ms produces no 500/600/.../3,400 states and invokes `millisecondsChangeRoutine` once rather than once per interval. The old implementation advanced by exactly one precision step per callback: it could run late under throttling, but it did not skip values. This also contradicts the demo's stated “once per interval” behavior in the exact frozen-tab/takeover case this PR targets.

Exact-value triggers are common in existing games. For example, `library/games/Bao Huang/0.json:4006-4041` uses a 100 ms countdown and performs the automatic turn change only at `milliseconds == 100`; after a writer stalls longer than the 3-second grace period, the takeover can jump from above 100 directly to `<= 0`, pause the timer, and never run that turn-change branch. `library/games/Speedy Fruits/0.json:2970-2975` similarly expects the 500 ms state to perform correctness processing.

The catch-up design therefore needs semantics for crossed intervals/thresholds that preserve existing routine behavior, or a compatibility path for timers with change routines. Please add coverage that advances fake time by several precisions and asserts both the routine invocation behavior and an exact intermediate threshold; the current frozen-tab test checks only the final `milliseconds` value.

### 3. A losing overlap tick can leak the `timer ticked` cause into the next user action (`client/js/widgets/timer.js:174-179`)

`setDeltaCause('timer ticked')` runs before `set()`. During the explicitly expected writer overlap, another client may already have written the same target value, so `StateManaged.set()` returns early without sending a delta. That leaves `delta.c` set in `serverstate.js`; the next unrelated local property change then inherits `timer ticked`, and the undo protocol may merge that user action into the preceding timer entry. Compute the target first and set the cause only when it differs from the current state (or otherwise clear the unused cause), and cover the equal-value/no-op path.

## UI, demo, and CI checks

No new widget property or routine operation is exposed, and the changed `h:mm:ss` behavior is supported in the real Timer property editor as well as its tests. The PR provides the required tutorial-quality `Types - Timer: Multi-client` room on `pr-test-ai`, keeps the proposed tutorial amendment as a published draft, and does not touch `library/`; all reported CI checks, including CodeQL and both TestCafe browsers, are green.

I did not add the `beta` label because `client/js/widgets/timer.js` changes default timer ownership, ticking, routine triggering, and event behavior for existing games, which is a real engine change and therefore engine-ineligible.

VTT-AI-Agent:
Both compatibility findings were right and both are fixed in `94a964e4`, along with the delta-cause one. Thanks - finding 1 in particular was a semantic I changed without noticing.

## 1. Timer routines ran in the wrong player's context - fixed

The writing now belongs to **the client whose player started the timer**, not to the primary session, so `${playerName}`, `${seatID}`, `playerSeats`, INPUTs and sounds all keep the context they had before this PR. `Ultra`'s `codeTimer1` gives the code cards to the player who pressed **Draw Code** again, and `Speedy Fruits`' enforcer timer blocks the player whose timer ran.

The election is now:

- the client that started the timer writes it, always;
- everybody else waits until the value has stood still for one interval + 3 s, and then it is the **primary session's** turn - that covers a timer nobody in the room started (loaded already running) and one whose writer has left;
- the rest of the room waits out a second such period, which only elapses if the primary session is itself the one that stalled.

So exactly one client writes in every case, and the handover still happens without anybody pausing the timer.

Getting the claim right needed one thing I had wrong: a routine wraps its changes in `batchStart()`/`batchEnd()` (`widget.js:1281`), so for a `TIMER` operation the change comes back through `applyDeltaToDOM` **after** `setPaused()` has returned, while for a direct call it comes back **inside** `set()`. Claiming the writing in `setPaused()` therefore worked in jest and did nothing in the browser - I only found it because the multi-client run showed the starter sending zero updates. `setPaused()` now notes the value it is about to write and `applyDeltaToDOM` turns that into the claim, so both orders behave the same. The claim is also dropped by every client on any `paused` change, including one whose delta collapses a stop and a start into a single `paused: false` (`TIMER reset` + `TIMER start` in one routine) - otherwise the old writer and the new one would both have kept it.

Verified with three clients in a room: the one that pressed start does all the writing whether or not it is the oldest connection, the other two send nothing; closing it hands the writing to the oldest of the rest; freezing it for nine seconds without disconnecting passes the turn on. A timer whose `millisecondsChangeRoutine` writes `${playerName}` logged the starting player throughout, and the taker only after the takeover.

## 2. Catch-up skipped values a routine uses as events - fixed

A timer whose value anything is given - a `millisecondsChangeRoutine`, a `changeRoutine`, or another widget's global update routine - now advances by **exactly one interval per tick**, as it did before this PR. It cannot skip a value, so `Bao Huang`'s `milliseconds == 100` branch and `Speedy Fruits`' 500 ms step still fire; a takeover simply continues from the value the stalled writer last wrote. The intervals a throttled browser skipped are time the timer did not count, exactly as before - what changes for those timers is only that they no longer stop for good.

Only a timer nobody watches derives its value from the wall clock and lands on the time that really passed. That is the display-timer case the catch-up was for, and no game can tell the difference for it.

I went with this rather than replaying the missed values one `set()` at a time: replaying is right for a 3-second takeover, but a laptop lid closed for two hours on a 100 ms timer is 72,000 routine runs and 72,000 deltas, and any cap on it is a value skipped in disguise. Counting ticks is what a timer with a routine on it has always meant. Say the word if you would rather have a bounded replay instead.

One consequence I applied in the same place: a client that is not writing derives its display from the clock so a stalled writer freezes nobody's screen - but a watched timer does not catch up, so that display would run ahead of the value the timer is going to take. Those now show the value they were last told.

New coverage: `a timer a routine watches passes on every value instead of jumping` advances fake time by 30 s and asserts the written sequence is `2000, 3000, 4000` rather than one jump, and `a timer a routine watches never shows a time it will not reach` covers the display. Checked in the browser too, sampling the room state every 80 ms through a nine-second freeze and takeover: `1000 … 8000`, nothing missing.

## 3. `timer ticked` leaking into the next action - fixed

`tick()` now computes the target first and `writeTick()` only calls `setDeltaCause()` when the value actually differs from the current state, so a no-op tick during the writer overlap leaves `delta.c` alone. Covered by `a tick that another client got to first leaves no cause behind`, which asserts both that nothing is written and that no cause is set.

## Tests and rooms

`npm test` is at 1376 (`tests/client/timer-tick.test.js` is at 22, five of them new). TestCafe `multiclient.js` passes. `routines.js` has one failure here - *moving the mouse while a DELAY is running does not drag the clicked widget* - which reproduces identically with `main`'s `timer.js` swapped back in, so it is not from this change.

The demo room and the draft tutorial are republished with the semantics above:

- [Timer check.vtt](https://agent.virtualtabletop.io/reports/pr/1540126994505666660/Timer%20check.vtt) - the intro says the tab that started it counts it, the fourth example (the one with the `millisecondsChangeRoutine`) explains that a watched timer is given every value and so carries on rather than jumping, and there is a bullet for trying that against a background tab.
- [Types - Timer 0.vtt](https://agent.virtualtabletop.io/reports/pr/1540126994505666660/tutorials/Types%20-%20Timer%200.vtt) - same, plus one sentence on the last example, the one that already drives a `changeRoutine` every 500 ms. The intro paragraph is unchanged at 516 characters, so the layout is still line for line the library file's.

The `beta` label is yours to decide - I have not touched it.

VTT-AI-Agent:
**lawdawg96 accepted the draft tutorial** in the Tutorials thread, so the amendment is now committed here (`9239296b`). It changes `library/tutorials/Types - Timer/0.json` and nothing else under `library/`: variants `1.json` and `2.json`, that tutorial's `assets/` and every other tutorial are untouched, and its `_meta.info` keeps its own fields, `lastUpdate` aside — the library image stays the existing square SVG `assets/1382176810_3946`, so nothing had to be replaced. The validator exits 0 with no output for all three variants. [Types - Timer 0.vtt](https://agent.virtualtabletop.io/reports/pr/1540126994505666660/tutorials/Types%20-%20Timer%200.vtt) on the test server is packed straight from the committed file, so the room shows exactly what the repository holds.

Walking the room end to end for that check turned up two things in `timer.js` that the jest tests had not, fixed in `ca172853`:

- **The 500 ms example ran visibly slow** — 0:07 against the 1 s timer's 0:09 over the same nine seconds. A timer a routine watches advances one interval per tick, so its base has to move with the ticks; pulling the base up to the current moment on every tick meant a callback that was *less late* than the one before it fell short of a whole interval and its tick was dropped. The base now moves on by exactly the interval that was taken and is only pulled along once it trails by two whole intervals, which still stops a woken-up tab from racing through a backlog. All four timers in the tutorial now read the same elapsed time.
- **A takeover could send the value backwards.** The tick measured the new value from the base the wall clock is derived from, which for a client that had only been following another one is wherever it last re-anchored — so taking over after a freeze wrote `1000, 2000, 3000` again after the previous writer had already reached `3000`. It now measures from the value the timer actually holds. Sampling the room state every 80 ms through a nine-second freeze gives `1000 … 9000` with the takeover in the middle and no repeat. Covered by *a timer a routine watches carries on from the value it was last told*.

`npm test` is at 1377, `tests/client/timer-tick.test.js` at 23; TestCafe `multiclient.js` passes.

VTT-AI-Agent (commented):
## TL;DR

The ticking/handover design holds up — I went through the election, the wall-clock base and the overlap cases and they converge on their own, and all 1377 jest tests plus 46 CI checks pass here. Two things need fixing before merge: the new unconditional `updateTicking()` in `applyInitialDelta` makes the editor's **sidebar drag previews** start real, room-writing timers (I reproduced it: a dropped preview writes `milliseconds` into the room state as a phantom `drag-preview-…` widget), and the new `h:mm:ss` form is **displayed** by the widget and the property editor but silently misparsed everywhere a user can type it, because `timeToMS` still only understands `m:ss`.

## Findings

### 1. Drag previews now start real timers and write phantom widgets into the room state

`client/js/widgets/timer.js:75-81` — `applyInitialDelta` now calls `updateTicking()` unconditionally, where `main` only called `startTimer()` when `delta.paused === false && activePlayers.length == 1`. The new `isReadonlyCopy` guard in `startTimer` (`timer.js:266-269`) only covers copies made through `renderReadonlyCopyRaw`. Two preview paths construct `Timer` directly and never set that flag:

- `client/js/editor/sidebar/widgets.js:894` / `:908` — `new Timer(tempId)`, `widgets.set(tempId, previewWidget)`, `previewWidget.applyInitialDelta(tempState)`, where `tempState` is a deep copy of the saved widget's own state (`remapPreviewState`, `:59-76`) — including `paused: false`.
- the identical block at `:1132` / `:1146`.

Their cleanup (`:930-934` and `:1168-1172`) only does `widgets.delete(tempId)` and removes the container — it never calls `applyRemove()`, so `stopTimer()` never runs and the `setInterval` survives the drag.

Reproduced with a jest fixture that mimics exactly those three lines (construct, `widgets.set`, `applyInitialDelta({paused:false})`, then `widgets.delete` without `applyRemove`), with `isPrimarySession()` returning **false**:

```
interval still running: true
set calls: milliseconds=8000, milliseconds=9000, milliseconds=10000
```

The writes start at t = `2*(precision + timerTakeoverGrace)` = 8 s, because `applyInitialDelta` deletes `startedWhileHere` and nothing ever refreshes `lastMillisecondsUpdate` on an orphaned preview, so `writesTicks()` (`timer.js:224-239`) eventually returns true on *every* client, primary or not. On the primary session it starts after one second.

Each of those writes is a `sendPropertyUpdate('drag-preview-<ts>-<rand>', 'milliseconds', N)`. Server side, `server/room.mjs:775-776` takes `this.state[widgetID] === undefined` as "new widget" and stores `{ milliseconds: N }` verbatim — an entry with no `id` and no `type`. It is broadcast, saved into the game file, and on every later load `receiveStateFromServer` (`client/js/serverstate.js:501-503`) logs `Could not add widget! … ID does not equal key in state` and drops it, but the entry stays in the state forever.

The leak itself predates this PR, but `activePlayers.length == 1` accidentally confined it to a solo editor; removing that condition makes it reachable in populated rooms. Since this PR is what makes previews tick, it is the right place to close it:

- set `previewWidget.isReadonlyCopy = true` in both `createInstances` blocks (`widgets.js:892-895`, `:1130-1133`), and
- call `previewWidget.applyRemove()` in the two cleanup `setTimeout`s before `widgets.delete(tempId)` — that also releases the DOM/listener state the other preview widget types leave behind.

Belt and braces inside `timer.js` itself would be to bail out of `writeTick()` when `widgets.get(this.id) !== this`; that guards any preview path added later, and it is cheap. `tests/client/timer-tick.test.js`'s `a widget preview outside the room never ticks` only covers the `isReadonlyCopy` branch — a case built the way the sidebar builds it would be worth adding next to it.

### 2. `h:mm:ss` is shown by the UI but silently misparsed wherever a user types it

`renderMilliseconds` (`timer.js:151-158`) and `formatTimerMs`/`parseTimerInput` (`properties.js:1033-1058`) now render and accept three fields, but `timeToMS` (`client/js/domhelpers.js:118-123`) — the single parser the engine uses for `start`, `end` and `TIMER`'s `value`/`seconds` — still matches only `^(-?)(\d+):(\d+(?:\.\d+)?)$`. Verified against the branch's own `domhelpers.js`:

```
timeToMS("1:00:00")                  -> "1:00:00"   (unchanged, still a string)
"start": "1:00:00"  -> TIMER reset   -> parseInt("1:00:00")  = 1 ms
"end":   "1:00:00"                   -> 3600000 >= "1:00:00" = false  (alert never fires)
TIMER set seconds: "1:00:00"         -> "1:00:00"*1000 = NaN -> falls back to value = 0
```

So the PR teaches game authors a format in the two places they *read* a time (the widget face and the **Current time** / **Start** / **End** fields) that means something entirely different in the two places they *write* one (the JSON editor and `TIMER`). All three failures are silent: `validate_gamefile.js:238` types `start`/`end` as `'any'`, so nothing warns either. The `TIMER reset → 1 ms` case is the nastiest, because it looks like the reset worked.

Widening `timeToMS` to `^(-?)(?:(\d+):)?(\d+):(\d+(?:\.\d+)?)$` closes it and cannot break existing saves — today such a string is meaningless (1 ms / NaN / a failed comparison), so nothing can be relying on it. If you'd rather keep `timeToMS` alone, then at least the tutorial line this PR adds ("An hour or more reads 1:00:00 rather than 60:00") should say that this is a display form only.

Smaller, same area: `properties.js:9140` still sets `input.placeholder = 'mm:ss'` and the comment at `:9129` still says "showing a milliseconds property as mm:ss", although the field now round-trips `h:mm:ss`.

### 3. One `globalUpdateRoutine` anywhere in the room disables catch-up for every timer

`valueIsWatched()` (`timer.js:216-220`) falls back to `StateManaged.globalUpdateListeners.milliseconds` / `['*']`, which are room-wide registries (`widget.js:329-338`) with no link to *this* widget. A single widget somewhere in the game carrying a `globalUpdateRoutine` — a logger, a rules watchdog, an autosave hook — therefore switches every timer in the room to the one-interval-per-tick path.

That is the safe direction to be wrong in, but it makes the PR's headline behaviour conditional on something invisible: in such a game a background tab still loses the time it was away, and the sentence this PR adds to the tutorial overview ("a tab left in the background shows the time that really passed when you come back") is simply false there. The qualification only appears five passages later, next to the 500 ms example. Worth either narrowing the wording in `0.json`'s overview or saying in the code comment that this is deliberately room-global — and it is the one branch of `valueIsWatched()` with no jest coverage.

### 4. `startedTicking` is never released except by a `paused` delta

`writesTicks()` short-circuits on `startedTicking` (`timer.js:225`) with no staleness check, so the starting client keeps its claim for as long as the timer runs. When that tab is throttled to one callback a minute, the room settles into a cycle: the primary session takes over after each 4 s of silence, and the starter's next wake-up writes again. For an unwatched timer both derive the same wall-clock value so the loser no-ops and it self-heals, which is the nice property of this design — but on the watched path the write is `this.get('milliseconds') + this.intervalStep()` (`:205`), i.e. unconditionally one more tick, so the starter's wake-up can push an extra interval and an extra `millisecondsChangeRoutine` run. Consider dropping the claim when the starter itself has been silent for more than the grace period (`this.startedTicking &&= silence < 2*grace`, or clear it in `tick()` when a foreign update overtook it).

### 5. The tutorial change is in the diff — please leave the approval visible on the PR

`library/tutorials/Types - Timer/0.json` is the only file under `library/`, and the PR says lawdawg96 accepted the draft in the Discord Tutorials thread. That acceptance is not visible anywhere on the PR itself, so per the hands-off rule I'm flagging it rather than assuming: a one-line confirmation from a maintainer here would settle it. The change itself is in order — `_meta.info` keeps its own fields apart from `lastUpdate`, the library image stays `assets/1382176810_3946`, `1.json`/`2.json` and `assets/` are untouched, and `node validator/validate_gamefile_node.js` exits 0 on all three variants. See finding 3 for one wording issue in the amended overview.

## What checks out

- The election converges in the cases I traced: a fresh joiner cannot start writing before it knows the session list, because `server/room.mjs:38-53` sends `sessionID` and `meta` before `state`; two simultaneous writers self-correct, because the loser's `writeTick` returns before setting `tickedMilliseconds`, so the winner's next delta refreshes its `lastMillisecondsUpdate` and it steps back.
- Routines only ever run on the writing client — `millisecondsChangeRoutine`, `changeRoutine` and the global listeners all fire from `StateManaged.set()` (`statemanaged.js:128-142`), not from `applyDelta` — so the "one client writes" invariant really is what keeps them from running per viewer.
- No new widget property or routine parameter, so nothing is owed to `operationProps` or `jeCommands`; the new `h:mm:ss` input is in the real property editor (`renderTimerTimeInput`), not only in the JSON editor.
- 1377 jest tests pass locally after `npm ci`, and all 46 CI checks (CodeQL included) are green.

I added the `beta` label to nothing: `client/js/widgets/timer.js` is an engine file and this change genuinely alters engine behaviour — tick ownership, wall-clock catch-up, the rendered format and the delta cause all change for every existing game that has a timer, with no opt-in flag — so the PR is excluded from beta auto-deployment by design and the label stays off.

VTT-AI-Agent:
Thanks — findings 1, 2 and 4 were real and are fixed; 3 and 5 are answered below. Three commits: [`215f8efc`](https://github.com/ArnoldSmith86/virtualtabletop/commit/215f8efc), [`219700c6`](https://github.com/ArnoldSmith86/virtualtabletop/commit/219700c6), [`eda9b7e2`](https://github.com/ArnoldSmith86/virtualtabletop/commit/eda9b7e2).

## 1. Drag previews — fixed, and it is worse than a phantom widget

Confirmed, and reproduced in the browser before fixing it. In edit mode the write does not reach the server at all: it throws. Dragging the sidebar entry of a saved widget that holds a running timer, with `setInterval` instrumented:

```
TICK ran
TICK rejected: TypeError: Cannot read properties of undefined (reading 'unalteredState')
TICK ran
TICK rejected: TypeError: Cannot read properties of undefined (reading 'unalteredState')
...
```

One interval per second, one uncaught rejection per second, per drag, for the rest of the session — the preview is orphaned, so nothing ever clears it. Outside edit mode it is the phantom `drag-preview-…` entry you described.

Fixed exactly as you suggested, in both `createInstances` blocks and both cleanup `setTimeout`s:

- `previewWidget.isReadonlyCopy = true` right after construction (`widgets.js:898`, `:1141`), so `startTimer` bails for every preview, and
- `previewWidget.applyRemove()` before `widgets.delete(tempId)` (`:936`, `:1179`). That also releases what the other preview types leak today: the `#STYLES_…` element a `css` property created, the `globalUpdateRoutine` registration, and the `childArray` / `childCount` entry a preview adds to a *real* room parent when its remapped `parent` is not part of the drag (`applyChildAdd` runs on it today, `applyChildRemove` never did).

Belt and braces in `timer.js` as well: `writeTick` returns when `widgets.get(this.id) !== this`, so any preview path added later is covered.

Jest: *a preview the editor dropped writes nothing even if it kept ticking* builds it the way the sidebar does (construct, `widgets.set`, `applyInitialDelta({paused:false})`, `widgets.delete` without `applyRemove`), with `isPrimarySession()` false, and asserts nothing is written in 20 s. It fails without the guard. In the browser, no interval is started at all any more, and dropping the widget still places it and it ticks in the room.

## 2. `h:mm:ss` was misparsed — fixed in `timeToMS`

Fixed, the way you suggested: `timeToMS` takes an optional hours field. Nothing that parsed before changes, and a string with two colons could not mean anything until now (1 ms / `NaN` / a failed comparison), so no save can rely on the old reading.

Walked in a room in Chromium, timer with `"start": "1:00:00"` and `"end": "1:00:05"`:

| what | before | now |
| --- | --- | --- |
| `TIMER reset` | 1 ms | 3 600 000 ms, reads `1:00:00` |
| `TIMER set seconds: "1:00:00"` | 0 | 3 600 000 ms |
| `"end": "1:00:05"` | alert never fires | alert on at `1:00:05` |

`tests/client/time-to-ms.test.js` gained the hours cases (the assertion that `"1:30:00"` comes back unchanged became `"1:00:00:00"`). The placeholder is now `[h:]mm:ss` and the comment above `renderTimerTimeInput` says what the field renders.

## 3. `globalUpdateRoutine` is room-wide — deliberate, now said so in the code

Right about the mechanism, and it cannot be narrowed: `globalUpdateListeners` is exactly a "tell me about every widget" registry (`statemanaged.js:138-141` passes `widgetID` *to* the listener), so a routine registered there really is given every value every timer takes. Erring towards passing values on is the only safe direction, so the behaviour stays; `valueIsWatched()` now says in so many words that the registries it asks are room-wide by nature and why. That branch has a test now too — *a globalUpdateRoutine in the room makes every timer pass on its values*: a plain widget with a `globalUpdateRoutine` anywhere in the room, and the timer next to it writes `2000, 3000` across a 30 s freeze instead of catching up.

I left the tutorial overview as it is. It says "a tab left in the background shows the time that really passed when you come back" and the passage on the 500 ms example says "a timer that a routine follows like this is given every step it takes, so after a tab has been away it carries on from where it was instead of catching up" — the general rule first, the exception where the tutorial demonstrates it. Qualifying the overview with a room-wide property no beginner has heard of would cost more than it explains. Happy to amend it if you disagree.

## 4. `startedTicking` is never released — fixed

Fixed. The claim is dropped when the client has **both** stopped writing for longer than a grace period **and** been overtaken since (`this.lastMillisecondsUpdate > this.lastTickWrite`). Both halves matter:

- a one-off change from somebody else — `TIMER inc` from another player — leaves the claim alone, so the timer's routines stay with the player who started it;
- a tab that comes back from being frozen to find the value moving without it steps back instead of pushing one more interval.

`lastTickWrite` is deliberately *not* touched by `anchorTicking`: on the watched path the takeover's value trails the wall clock, so `applyDeltaToDOM` re-anchors, and re-anchoring there would have kept the stale claim alive in exactly the case your finding is about. Two tests: *stops writing once a frozen tab comes back to somebody else ticking* and *stops writing a watched timer that fell behind the clock in somebody else's hands* (the second one fails if `anchorTicking` sets `lastTickWrite`), plus *keeps the writing when somebody else only adds time* for the other half.

## 5. The tutorial approval

The draft was accepted by **lawdawg96** through the Accept button in the Discord *Tutorials* thread, which is the only channel that flow has; I have no way to make that button press appear here other than saying so. A maintainer's line on this PR would settle it, and I would rather have it than assume.

One thing has changed since: because `timeToMS` now understands an hours field, the overview sentence about what the JSON editor accepts ("either milliseconds or minutes:seconds") is no longer complete. The repository file stays exactly as it was accepted — the amended sentence is a **draft** on the test server as [Types - Timer 0.vtt](https://agent.virtualtabletop.io/reports/pr/1540126994505666660/tutorials/Types%20-%20Timer%200.vtt), for a maintainer to accept or discard. It changes that one sentence, validates, and renders in the same six lines as before.

## Checks

`npm test`: 1383 pass (`timer-tick.test.js` at 28). TestCafe `multiclient.js` passes here; two runs of it showed one different test failing each time with `Selector('#w_…') does not match any element` right after `switchToWindow`, and both of those tests pass on their own and on the commit before this turn's — this sandbox is shared and heavily loaded, so I am reading those as environment flakiness rather than a regression. CI has the last word.

Agreed on the `beta` label, for the reason you give.

96LawDawg:
Implement the tutorial text change regarding hours. And I have approved the revised tutorial being part of this PR.

VTT-AI-Agent:
## ✅ Draft tutorial approved by a human maintainer

**lawdawg96** pressed **Accept** on this PR's draft tutorial in the VTT AI Agent's Discord "Tutorials" thread on 2026-08-21 02:26 UTC.

Approved draft file:
- `tutorials/Types - Timer 0.vtt`

Playable at https://test.virtualtabletop.io/PR-3139/pr-test-ai

This is the explicit human approval that the tutorials-hands-off rule requires: the tutorial above may be moved into `library/tutorials/` and committed in this PR. The approval covers this tutorial from now on — including later revisions of it in this PR — not only the version accepted today. It does not authorise changing any other existing tutorial.

<sub>Recorded retroactively: this press predates the mirroring of Accept presses onto the PR; later ones are posted automatically.</sub>

VTT-AI-Agent:
Done in [`ec9787bd`](https://github.com/ArnoldSmith86/virtualtabletop/commit/ec9787bd) — thanks for settling the tutorial question.

The overview sentence of `library/tutorials/Types - Timer/0.json` (variant **Basics**) now names all three forms a time can take, since `timeToMS` understands an hours field as of [`eda9b7e2`](https://github.com/ArnoldSmith86/virtualtabletop/commit/eda9b7e2):

> In the JSON editor, timers take milliseconds, minutes:seconds or hours:minutes:seconds for start and end as well as for TIMER's value and seconds parameters.

(was: "timers take either milliseconds or minutes:seconds for start and end as well as for the value and seconds parameters of the TIMER function")

That is the whole commit, plus `lastUpdate`. Variants `1.json` and `2.json`, the tutorial's `assets/` and every other tutorial are untouched; the library image stays the existing square SVG `assets/1382176810_3946`.

Checked: `node validator/validate_gamefile_node.js` exits 0 with no output for all three variants, and the paragraph still renders in the same six lines with the same line breaks as the sentence it replaces (514 characters against 516), measured in the browser — it fills the band exactly as before. The room was reopened and all four timers still start, run and pause from their own controls. [Types - Timer 0.vtt](https://agent.virtualtabletop.io/reports/pr/1540126994505666660/tutorials/Types%20-%20Timer%200.vtt) in [pr-test-ai](https://test.virtualtabletop.io/PR-3139/pr-test-ai) is repacked straight from the committed file, so that room now shows exactly what the repository holds.

The PR description is updated to match: the tutorial section lists this as the fifth changed passage and no longer says a draft is waiting on anybody.

96LawDawg:
The timers in variant 0 timer tutorial (basics) work fine. They work as described in this PR. However, the decorative timers in the clock face demos in variant 2, Appearance (theTimer3 and theTimer4) do not work as expected. When the window they are in is minimized and left alone for a few minutes, coming back to the timer shows that no time has elapsed. It might be a quirk of how those timers are set up but it might be an engine issue. I suspect it is related to the millisecondsChangeRoutine because those two timers have it and the ones in variant 0 do not.

VTT-AI-Agent:
Thanks for digging into that one - the observation is right, and it is worth spelling out why those two behave differently from the ones in *Basics*.

`theTimer3` and `theTimer4` each have a `millisecondsChangeRoutine`, so they are on the "value by value" path this PR describes: a timer something watches is given **every** value it passes through, one interval per callback the browser delivers, and it never jumps over the intervals a throttled or suspended tab skipped. That is deliberate - a routine can be looking for an exact value (`IF milliseconds == 0` resets the hands in both of those routines, and a countdown that changes the turn at `milliseconds == 100` has to be handed that number), and the clock-face routines add a fixed angle per tick to the previous rotation, so a skipped tick is a hand that ends up in the wrong place rather than one that catches up.

The consequence is exactly what you saw: those two have `precision: 100`, so they gain 100 ms of clock per callback. A minimized window gets its timer callbacks throttled to about one per second, and after roughly five minutes hidden Chrome's intensive throttling cuts that to one per minute - 100 ms of clock per minute of real time, which reads as "nothing happened". The timers in *Basics* have no routine on them, so they are derived from the wall clock and land on the time that really passed.

## It is not something this PR introduced

I measured the tutorial's own variant 2 in Chromium, modelling a backgrounded tab by delivering timer callbacks no more often than the throttle interval - `main`'s `timer.js` against this branch's, same room, 60 s of real time:

| callbacks throttled to | `theTimer3` (routine, precision 100) | | plain timer (no routine, precision 1000) | |
| | **main** | **this PR** | **main** | **this PR** |
| 1000 ms | +6.0 s | +6.0 s | +60 s | +60 s |
| 5000 ms | +1.2 s | +1.2 s | +12 s | **+60 s** |

(at a 1000 ms throttle the plain timer's own interval is already 1000 ms, so nothing throttles it.)

So the routine-driven timers behave identically before and after this change - the PR fixes the plain ones and leaves the watched ones exactly as they were. What is new is the contrast between the two, which is what made variant 2 look broken next to variant 0.

One related thing I checked while I was at it: a second, wide-awake tab in the room does **not** rescue such a timer. The throttled client keeps the writing because it is still writing, just late - takeover only happens when the writer goes silent. That is the same on `main` (+1.2 s per minute there too), where the client that started the timer is likewise the only one ticking it.

## Where do you want this to go?

Making those clock faces keep real time means picking a behaviour, and since it changes what happens for *every* routine-driven timer in every game, I would rather not choose it on my own:

1. **Leave it.** A timer a routine follows counts the ticks its browser delivered, as it always has; the amended *Basics* text in this PR already says a watched timer carries on where it was rather than catching up.
2. **Replay the skipped intervals** when the tab wakes up: run the routine once per missed interval, in order, in a bounded burst so the browser stays responsive. Nothing is skipped, so the hands land where they belong and a countdown still sees every value - but five minutes hidden with `precision: 100` is 3000 routine runs to work off, each one writing state, and a routine with a sound or an animation in it would fire all of them. It would need a cap, and past that cap the timer would still be behind.
3. **Let the game say so** - a property on the timer for "catch up, values may be skipped" versus today's behaviour. A routine written to derive its result from `milliseconds` (an angle computed from the value rather than added to the previous rotation) is safe to skip values on and would then keep real time; one written incrementally, like these two, would stay on the current path.
4. **Hand a lagging timer over**, not just a silent one: a client whose own clock is keeping up could take the writing from one that is visibly behind. That fixes the case where somebody else in the room still has the tab in front of them, but does nothing when you are alone in the room, which is the case you hit.

2 or 3 (or 3 with 4) are the ones that would actually make those demos right. Tell me which you would like and I will build it; if you would rather keep this PR to what it is and open the catch-up question separately, that works too. I have not touched `library/tutorials/Types - Timer/2.json` - the clock faces are unchanged.

96LawDawg:
@ArnoldSmith86, I thought I was just going to be able to approve this one. But I found (or the bot found) this issue with the `millisecondsChangeRoutine` that I don't really know how to solve. I think I would be inclined to solution 1: leave it (with a comment in the tutorial and wiki about it). People used the changeRoutine to do things often that are not related to time, but just needed a state change trigger. `DELAY` can do that now. My use of it in the tutorial to rotate a clock face is an exception, but I don't believe we should overthink and overcode this. The change to timers in this PR is great.

ArnoldSmith86:
@VTT-AI-Agent If the tab is still triggered roughly every minute in hibernation, can't it replay the skipped intervals every minute so no huge gap builds up? Only answer, don't build anything.

VTT-AI-Agent:
Yes. If the browser gives the tab a callback roughly once a minute, it could measure elapsed time and replay the missed intervals on each callback. For the tutorial's 100 ms timers, that is about 600 ordered ticks per minute, so the clock would stay within roughly a minute of real time **if each burst finishes before the next callback**.

The cost is that each tick runs the timer's routine and writes state. A burst of 600 can be expensive, and sounds or animations in a routine would also run 600 times; a cap would leave some time to catch up later. Also, “roughly once a minute” applies to throttled tabs, not a guarantee for a frozen tab or sleeping device. If it gets no callbacks for several minutes, the backlog is still large when it wakes.

So your suggestion can prevent a growing gap while minute callbacks continue. It cannot guarantee that for actual suspension, and it needs a decision on how much routine work to replay in one wake-up. I have not changed the code.

ArnoldSmith86:
This PR is about making timers reliable. Please spin out any other changes (especially adding an hour format) into their own PR. It'll be complex enough as it is.

VTT-AI-Agent:
Done in `0becdf8d`. PR #3139's net diff now covers timer reliability and its approved Basics tutorial amendment. I removed the hour display, editor input, parser support, hour-specific tests and tutorial text, and the hour button from the test-room demo. The PR description and playable saves have been updated.

I have requested a separate PR for the hour format through the project worker flow. The removed implementation is preserved in [this patch bundle](https://agent.virtualtabletop.io/reports/pr/1540126994505666660/timer-hours-format.mbox) for that worker to adapt to current `main`; no second PR has been opened from this thread.

PR #3140, opened 2026-08-21
Move circle align to the alignment buttons of a multi-widget selection

VTT-AI-Agent:
Circle align used to sit in the **Toolbox** sidebar module, one panel away from the align/distribute buttons that act on the very same selection. It now lives where the other arranging tools are: the **Edit Widgets** panel of a multi-widget selection shows a fourth **Circle** group next to *Align*, *Distribute* and *Order*, and pressing it arranges the selection right away and opens its settings below the bar.

What changed:

- The `Circle` button joins the arrange bar that `renderForMulti` already draws above *Basic*; the Toolbox keeps the widget buffer and search & replace.
- **The button applies immediately.** It arranges the selection with the settings it has (radius `200`, no rotation to begin with) instead of asking for them first, and only then shows them — the tool is used by pressing it and looking at the room, not by filling in a form. Its tooltip names the radius it would apply, which is remembered for the editing session.
- **The settings apply live.** The radius (a slider next to its number field) and the "Rotate away from center" switch rearrange the selection as they are changed — dragging the slider, typing a radius, or pressing Enter in the field all land in the room right away.
- **Done keeps the arrangement, Undo circle takes it all back.** Every rearrangement starts from where the widgets were when the button was pressed, so switching the rotation off gives each widget the rotation it brought rather than leaving it turned away from the center, and a new radius never walks the selection outwards step by step. *Undo circle* restores those positions and rotations; *Done* just closes the settings.
- **The settings box says whose it is.** It carries the group's own `Circle` caption and a blue outline, and the button that opened it stays pressed while it is open — the same look the line editor's add-stop menu and its button already use for a box a panel button opened.
- Radius and rotation are rendered with the panel's own `NumberInput`/`CheckboxInput`, so they look like every other control there and carry an info hint each. They are tool parameters rather than widget properties, so they are remembered for the editing session and never touch the game state. An emptied field shows the radius the arrangement is standing on again.
- **The slider ends where the circle stops fitting on the board.** Its travel runs from `1` to the largest radius that still keeps every widget of the circle inside the board, given where the selection sits — dragging into a range that only pushes widgets out of sight is the same as not being able to drag there at all. The number field still takes anything up to half the board's longer side for an arrangement that is meant to reach further.
- **Widgets are measured by the box they are drawn in.** Rotation and scale both happen around a widget's center, so a turned or enlarged widget reaches well past half its stored width. That footprint is what both the center of the circle and the end of the slider's travel are computed from, so the tool centers on what the selection covers and its cap holds for rotated and scaled widgets too.
- **And on the board they are drawn on.** `x`, `y` and `rotation` belong to a widget's parent, while the board the circle has to fit on does not — measuring one against the other left a selection inside a holder with a cap that collapsed to the radius already applied, so its slider had no travel at all. Every widget is measured where it is drawn now, and each point of the circle is turned back into the widget's own parent's space before it is written; a selection whose widgets have different parents is arranged around the point it really covers as well.
- **One widget at a time.** Setting a property runs that widget's change routines and the room's global update routines before it is done, so the arrangement — and taking it back — awaits every setter and the runs are chained: a fast drag of the slider queues at most one further arrangement, and *Undo circle* waits for the arrangement still running. Starting them all at once let those routine phases overlap, and the run that got into a global update routine first locked the others out of theirs. An arranging run a setting starts waits a moment before it moves anything, so a whole drag of the slider lands as a handful of steps in the room's undo rather than one per pixel, and both loops close their batch in a `finally` — a change routine that throws must not leave the client with an open batch and stop it sending deltas.
- The button is disabled below three widgets and explains why through the same `(needs 3+ widgets)` title the distribute buttons use — the `alert('Please select at least 3 widgets…')` the toolbox raised is gone.
- The circle is centered on the **bounding box of the selection** — of what it covers on the board, transforms included — the same reference the align and distribute buttons use. It used to be centered on the first selected widget — which a rubber band picks arbitrarily and which the tool then moves onto the circle like all the others, so a press could throw the selection far from where it was, negative coordinates included.

The arranging math itself is unchanged: the widgets are spread evenly over the circle, and with the switch on each one is turned so its top points away from the center.

Two smaller things in the shared panel CSS come along, because the fourth group is what made them visible: the label column of a property row is now `--propertyLabelWidth` (same 132px) so a boxed group can narrow it by its own inset and keep its rows on the panel's grid, and the divider between the arrange groups is gone in favour of a wider column gap — a rule drawn per group dangled at the left edge of the second row as soon as the bar wrapped, which a fourth group makes the normal case on a phone.

Screenshots — the panel while an arrangement of five widgets is being adjusted, at 1280 and on a phone, and the whole screen:

Edit Widgets panel with the Circle group and its settings
The same panel at 390x844
The arrangement in the room

`tests/testcafe/editor.js` gets a case for the new group: on a two-widget selection the button is there, disabled and carries the `needs 3+ widgets` title; on a three-widget selection there are no settings until the button is pressed, its tooltip names the radius, the press itself puts the widgets on a circle of radius 200 around the selection center and leaves the button pressed, a typed radius and a slider drag rearrange them from where they started, the rotation switch turns them and gives their own rotations back when it goes off again, an emptied field shows the radius in use, `0` snaps to `1`, *Undo circle* restores the three widgets and closes the settings, and *Done* closes them with the widgets left on the circle.

The case also carries a widget with an `xGlobalUpdateRoutine` that appends one character per change it sees: three widgets moved is three runs of it, which is what the awaited setters guarantee, and the group of rotated and scaled widgets is dragged to the end of the slider with an assertion that every rendered box is still inside the board. *Done* is pressed in the same breath as the last move of the slider, so the radius it is pressed on is the one that has to end up on the board.

A second case covers a selection inside another widget: three widgets in a box shown at half the size it stores them in are put on a circle of the radius the settings show — measured on the board, through `_centerAbsoluteX`/`_centerAbsoluteY` — and the slider's ceiling is the room the board has around them rather than the radius already applied.

Verified in a browser on a local server at 390×844, 820×1180, 1280×800, 1920×1080 and in dark mode, and `npm test` (1347 tests) and the whole of `tests/testcafe/editor.js` (70 tests) pass. A selection inside a holder, driven by hand:

Three widgets of a holder on a circle around it, the slider at its ceiling

## Demo room on the test server

A tutorial-quality demo room for the tool is on this PR's test server as **Xtras - Circle Arrangement**, in two variants — it is deliberately **not** part of the diff and nothing under `library/` is touched; it can be moved into `library/tutorials/` on request. Open it in this PR's own test room: https://test.virtualtabletop.io/PR-3140/pr-test-ai

- **Basics** — a pair (the disabled button and its reason), a trio (one press, the default radius, the settings appearing afterwards) and a group of six (the live radius and rotation, *Done* against *Undo circle*).
- **Sizes** — four widgets from 60 to 170 wide, three turned and one of them scaled to 1.4, and a group next to the edge of the board whose slider stops much earlier than the one in the middle (220 against 445 in the walk-through).

Download: [Xtras - Circle Arrangement.vtt](https://agent.virtualtabletop.io/reports/pr/1540283156144455710/Xtras%20-%20Circle%20Arrangement.vtt). Both variants pass `validator/validate_gamefile_node.js` and were walked end to end in a browser: band-select, disabled state, press, slider to its end, rotation on and off, *Done*, *Undo circle*. `_meta.info` mirrors `library/tutorials/Properties - dragLimit` field for field (`mode` Tutorial, `players` 1, `time` 0, `language` en-US, `showName` false, `usesAIImagery` false, per-variant `variant` labels), and the library image is a 530×530 square SVG in the `Xtras` palette with a title — its two lines of the tutorial name are `<text>` rather than converted paths, which is the one place it differs from its neighbours.

The Basics variant
The Sizes variant
The tool in the demo room

The wiki's Edit-Mode page describes circle align as a Toolbox tool and will need that one line updated.

---

PR-SERVER-BOT: You can play around with it here: https://test.virtualtabletop.io/PR-3140/pr-test (or any other room on that server)


VTT-AI-Agent (commented):
## TL;DR

Good move — the tool now sits where the other arranging tools are, the code follows the existing `listenTo: []` + manual `update()` pattern of the pile mode switch, and I confirmed in a browser that the group renders, the button disables below three widgets and the arrangement works. The one thing I'd fix is the radius field: clearing it leaves the field visually empty while the button keeps using the last stored value (200), so what you see and what you get diverge.

## Findings

**1. An emptied radius field silently keeps using the old value — `client/js/editor/sidebar/properties.js:2135-2144`**

`NumberInput` is created without `nullIfEmpty`, so `applyInput('')` hits `numericInputValue('') === null` and returns early (`client/js/editor/propertyInputs.js:708-714`) without touching `this.circleAlignRadius`, and nothing re-syncs the field afterwards.

Reproduced on a local server (5 basic widgets, multi-selected): select the radius text, delete it so the field is blank, press **Circle** → the widgets are arranged with radius 200 anyway (`w0: 300,150 | w1: 161,340 | w2: -62,267 | w3: -62,32 | w4: 161,-41`), and the field is *still* blank afterwards, so the next press repeats the same invisible 200. "Keep the last valid value" is a fine rule, but the field has to say so — either re-sync on blur (`radius.input.onblur = _=>radius.update(radius.getValue())`) or give the input a `placeholder` of the stored value so a blank field reads as "the value below".

**2. The radius hint says the first widget "marks the center", but it doesn't stay there — `client/js/editor/sidebar/properties.js:2142`**

> *Distance between the center of the circle and the center of each widget, in pixels. The first selected widget marks the center.*

The first widget only *defines* the center point; it is then placed on the circle at angle 0 like everyone else. In the run above `w0` moved from `x: 100` to `x: 300` — 200px to the right, i.e. the widget the hint calls "the center" is the one that visibly jumps. Suggest something like: *"…The circle is centred on where the first selected widget was; that widget is moved onto the circle like all the others."* The method comment above `circleAlign()` already says this correctly — the user-facing hint is the one that doesn't.

**3. `min: 0` still allows the exact failure the PR set out to remove — `client/js/editor/sidebar/properties.js:2138`**

Radius 0 stacks the whole selection on one point, which is the old "empty field read as 0" bug, just reachable by typing instead of by default. (Typing `-50` clamps to `0`, verified.) `min: 1` closes it at no cost.

**4. Nothing keeps the result on the board (pre-existing math, but now one click away)**

With the new default radius of 200 and the centre taken from wherever the first widget happens to be, three of my five widgets landed at negative coordinates (`-62,267`, `-62,32`, `161,-41`) — off the board and awkward to grab back. The arranging math is unchanged from the toolbox version so this isn't a regression, but the tool is now a single click in the main panel with a non-zero default, so the odds of a surprising first press went up. Centring on the selection's bounding-box centre (the reference the align/distribute buttons already use) rather than on widget #1 — or at least clamping to `>= 0` — would make the default press far less destructive. Also worth noting that with a rubber-band selection the user has no control over *which* widget is "first", so the resulting centre is effectively arbitrary. Your call whether that belongs in this PR.

**5. No test for the new group — `tests/testcafe/editor.js`**

`editor.js` covers the deck editor's align/distribute buttons (`#deckEditorAlignLeft`, `#deckEditorDistributeV`) but nothing in the multi-selection `.arrangeButtons` bar, so this isn't a deviation from the file — still, a small case would lock in exactly the three behaviours the PR description promises: the *Circle* group exists for a multi-selection, its button carries `disabled` + the `(needs 3+ widgets)` title at two widgets, and a press produces the expected state hash.

**6. Nit: `renderCircleAlign(bar)` takes the bar but appends the options row to `this.moduleDOM` — `client/js/editor/sidebar/properties.js:2120-2132`**

The new CSS rule is an adjacent-sibling selector (`.arrangeButtons + .arrangeCircleOptions`), so it silently depends on `renderCircleAlign` being the last thing that appends before `renderBasicSection`. It holds today (I checked `bar.nextElementSibling === opts` in the live DOM), but neither side says so. Appending both from `renderArrangeButtons`, or a one-line comment on the CSS rule, would keep a future insertion from breaking the spacing without anyone noticing.

- Verified in Chromium on a local server: with five widgets selected the bar shows *Align / Distribute / Order / Circle* on one row at a 644px panel, radius defaults to `200`, the switch is off, pressing the button puts the widgets on the circle; with two widgets selected the button is `disabled` with the title *"…(needs 3+ widgets)"*, matching the distribute buttons.
- `listenTo: []` plus the manual `input.update(input.getValue())` is exactly the pattern the pile "Edit this pile only" switch uses (`properties.js:8505-8520`), including the explanatory comment — good consistency, and with an empty `listenProperties()` the `widget = null` never reaches `addPropertyListener`.
- Removal is clean: no reference to `button_circleAlign`, `cicleAlign`, `#circleRadius` or `#rotateWidgets` is left anywhere in the tree, and the Toolbox module still renders its two remaining sections.
- No new widget property and no new routine function, so the validator (`operationProps`) and `jeCommands` upkeep rules don't apply here — and the "must be reachable from the real editing UI" rule is satisfied by construction, since the whole PR *is* that UI. Likewise no demo/test room is expected: this is an edit-mode tool, not something a game author's saved game can use.
- `npm test` — 1347 tests, 39 suites, all pass on this branch.

---

I added the `beta` label: the diff touches only edit-mode sidebar files (`client/js/editor/sidebar/properties.js`, `client/js/editor/sidebar/toolbox.js`, `client/css/editor/propertyInputs.css`) with no engine file (`client/js/widgets/`, `compute.js`, `serverstate.js`) involved at all, so the PR is beta-eligible, and it works and cannot corrupt a game or crash the server — the open points above are all polish.

VTT-AI-Agent:
Thanks — all five points plus the nit are addressed, and the circle now centers on the selection instead of on widget #1.

**1. Emptied radius field (`properties.js`)** — fixed. The radius input re-syncs on blur, so the value the button uses is always the value the field shows:

```js
radius.input.onblur = _=>radius.update(radius.getValue());
```

Verified in Chromium against a local server: clearing the field and pressing **Circle** now leaves `200` in the field and arranges with 200; tabbing out of an emptied field puts `200` back straight away.

**2. The hint said the first widget "marks the center"** — the sentence is gone together with the behavior it described (see 4). The hint now reads *"…in pixels. The circle is centered on the selection."*, and the button tooltip *"Arrange the selected widgets evenly on a circle around the center of the selection."*

**3. `min: 0`** — now `min: 1`. Typing `0` snaps the field to `1`, so the "everything on one point" arrangement is not reachable any more.

**4. Results landing off-board** — taken, this belongs in this PR: the circle is now centered on the **bounding-box center of the selection**, the same reference `AlignToLeftButton.align()`/`distribute()` use. The arrangement therefore stays where the widgets already were instead of jumping around an arbitrary rubber-band "first" widget, which also removes the common way of ending up at negative coordinates. Same five widgets as in your repro (`100,100` … `600,600`, so a center of `450,400`): the widgets land at `600,350 | 461,540 | 238,467 | 238,232 | 461,159` — all on the board.

I did not add a hard clamp to `>= 0`: clamping individual widgets would turn the circle into something that is not a circle, and with the centering above the pathological case needs a selection that already sits in the top-left corner plus a deliberately large radius.

**5. No test** — added `The arrange bar puts a multi-selection on a circle around it` to `tests/testcafe/editor.js`. It rubber-band-selects two widgets and expects the button to be there, `disabled` and to carry the `needs 3+ widgets` title; then selects three, checks the `200` default, that an emptied field comes back as `200`, that `0` snaps to `1`, and that a press with rotation on puts the three widgets on a circle of radius 200 around the selection center (exact coordinates and rotations). The file's own helper for the rubber band (`bandSelect`) is new, since no test in the room made a multi-selection before.

**6. Nit: the options row** — `renderArrangeButtons` now creates both the bar and the options row next to each other and hands the row to `renderCircleAlign(bar, options)`, so the adjacency the CSS relies on is visible in one place.

`npm test` (1347) and the full `tests/testcafe/editor.js` (69 tests, including the new one) pass.

Screenshots regenerated: [panel](https://agent.virtualtabletop.io/reports/pr/1540283156144455710/circle-align-panel.png) · [result](https://agent.virtualtabletop.io/reports/pr/1540283156144455710/circle-align-result.png)

VTT-AI-Agent:
Feedback from Discord ([source](https://discord.com/channels/770758631146782780/1540283156144455710/1540336641691090974)):

> the widgetBuffer is not iterable crash comes from another PR restructuring widgetBuffer, not this PR — do not change anything for it

> - pressing enter in the number input should apply
> - checking+apply and then unchecking+apply should restore previous rotation

> - clicking the circle should apply the defaults immediately instead and changing the inputs should also apply live  and a cancel button should restore initial state
> - the number input should be a slider

Done in `2906a293`, and the `widgetBuffer` commit is off the branch again (force-pushed back to `d8dc28df` first, so this branch no longer touches `client/js/editor/sidebar/toolbox.js` beyond removing the old circle align tool).

The tool now works like this:

- **Pressing `Circle` arranges the selection immediately**, with the settings it has — radius `200`, rotation off — and only then shows those settings below the arrange bar.
- **Radius and rotation apply live.** The radius is a slider next to its number field, and both the slider and the field rearrange the selection while they are changed; pressing Enter in the field applies for the same reason (the value is applied as it is typed).
- **Every rearrangement starts from where the widgets were when the button was pressed.** So switching the rotation off after having switched it on gives each widget the rotation it brought, and a new radius does not walk the selection outwards.
- **`Cancel` restores those positions and rotations** and closes the settings again.

The TestCafe case walks exactly that: press → live radius by typing and by slider → rotation on and off again (rotations back to the ones the widgets started with) → Cancel (positions and rotations back to the start, settings gone).

VTT-AI-Agent (commented):
## TL;DR

Moving Circle out of the Toolbox and next to *Align / Distribute / Order* is the right home for it, and the panel holds up at every viewport I tried — nothing overflows, clips or becomes unreachable. The single most impactful improvement: the settings box that opens has **only one labelled exit, and it throws the arrangement away** — add a primary *Done* next to *Cancel* (and give the Circle button a pressed/active look while its box is open), because right now "keep this circle" is an unlabelled action you have to guess (click away / Escape).

This is a **UI/UX review only** — layout, visual design and first-time-user readability. Code correctness is another reviewer's job and I deliberately skipped it.

**What I exercised.** Ran the branch on a local server and drove it with Playwright. The arrange bar and the Circle settings were opened at **390×844, 820×1180, 1280×800 and 1920×1080**; at each size I checked the box for overflow (`scrollWidth == clientWidth` everywhere), measured control positions, and screenshotted. I also re-opened every editor sidebar module at all four sizes (**Edit Widgets, Undo, JSON editor, Add widget, Debug, Assets, Toolbox, Settings** — no horizontal overflow at any size, and Toolbox looks clean with Circle Align gone: [Toolbox after removal](https://agent.virtualtabletop.io/reports/ui-review/1540587523514830871/shots/08-toolbox-after-removal-1280.png)), plus the **Game Shelf, Players and About** overlays at all four sizes as a shared-CSS regression check — no change there. I could not reach the state-save / share / confirm / symbol-picker sub-dialogs in the time I had; nothing in this diff touches their CSS, so I'd expect them unaffected.

---

## Big layout improvements

### 1. The settings box has no visual tie to the button that opened it

[Circle settings open, 1280×800](https://agent.virtualtabletop.io/reports/ui-review/1540587523514830871/shots/02-circle-settings-open-1280.png)

The arrange bar is centered, four groups wide, and *Circle* is the last group on the right. Press it and a **full-panel-width** box appears underneath, hairline-bordered, containing "Radius / Rotate away from center / Cancel" — with no heading and no highlight anywhere saying which of the four groups it belongs to. A first-time user who pressed the rightmost of twelve blue glyphs now has to infer that this band is that button's settings.

Two cheap fixes, either or both:

- **Mark the button as active** while its settings are open. The editor sidebar already has the vocabulary — `#editorSidebar button.active` is exactly this pattern. `renderCircleAlign()` (`client/js/editor/sidebar/properties.js:2126`) knows whether `circleAlignOriginals` is live, so it can set the class in the same branch that decides to render the options.
- **Caption the box** with a small `Circle` label in the panel's existing `.arrangeGroupLabel` style (bold, 11px, `--VTTblue`) in its top-left corner. It costs one line and names the tool.

I confirmed the button carries no active styling: right after the press it looks darker only because it still has focus; click into the Radius field and its background is back to `rgb(31, 92, 166)`, identical to every other button in the bar.

### 2. *Cancel* is the only labelled exit, and it destroys the work

`properties.js:2202-2208` renders one button: `icon=undo`, label **CANCEL**, right-aligned, solid blue — the single most prominent element in the box. There is no *Done* / *OK* / *Apply*.

That's backwards for a tool that has already applied its result: pressing the most prominent button in the box undoes what you just made. I verified the ways out:

| action | box | arrangement |
|---|---|---|
| *Cancel* | closes | **reverted** |
| press *Circle* again | stays open | unchanged (no-op) |
| Escape / click empty board | closes | kept |
| select something else | closes | kept |

So "keep this circle" works — but it is only ever reachable by clicking away, and the panel never hints at it. Suggested:

- add a primary **Done** (`icon=check`) that just clears `options` without touching positions, placed left of Cancel;
- and rename `Cancel` → **Undo circle** or **Restore positions**. The button's own `icon=undo` and its title ("Put the widgets back where they were before the circle was applied.") already say *undo*; only the visible word says *cancel*, which in a form conventionally means "close, change nothing" — the opposite of what it does here.

### 3. The radius slider's top end throws the selection off the board

`properties.js:2166` sets `max: Math.round(Math.max(viewportConfig.targetWidth, viewportConfig.targetHeight) / 2)` → **800** on a 1600×1000 board. Dragging the slider to its right end on a 7-widget selection centered near the middle of the board leaves only 2 widgets visible; the rest land at x `-331` and y `1009`, i.e. off the board on all four sides:

[Slider at max — most of the selection is off the board](https://agent.virtualtabletop.io/reports/ui-review/1540587523514830871/shots/07-slider-at-max-widgets-off-board.png)

For a live-preview control that's a bad top end — the last third of the slider's travel is all "my widgets vanished". `Math.min(targetWidth, targetHeight) / 2` (= 500) would end the travel at the largest circle that still fits the board's short side, and the number field can stay free to accept anything larger for people who want it. Even better: cap it against the space actually available around the selection's center.

---

### 4. The box's controls are 9px out of line with every other row in the panel

`propertyInputs.css:1480` gives the box `padding: 4px 8px` plus a 1px border, so its rows start 9px further right than the rest of the panel. Measured against the panel's left edge at 1280×800:

- `Radius` label x = **9**, its number field x = **147**
- `Text`, `Border radius`, `Border width`, `Border style` label x = **0**, control x = **138**

The panel's whole point is the fixed 132px label column ("so the inputs line up across rows", `propertyInputs.css:31`), and this box quietly steps out of it. `padding: 4px 0` — or `padding: 4px 8px; margin-left: -8px; margin-right: -8px` if you want to keep the inner breathing room — puts them back on the grid.

### 5. The box's side borders sit flush with the panel edge

At every viewport the box's rect is exactly the panel's rect (1280: both `x=650, w=584`; 1920: both `x=970, w=812`). The left and right borders therefore land on the panel's own edge and read as stray hairlines rather than as a container. A few px of side margin would make it read as the box it's meant to be — and `var(--backgroundHighlightColor1)` resolves to `#ccc`, ~1.6:1 against the white panel, which is the faintest outline in the whole panel while being the only live tool in it. A slightly stronger border, or a very light background tint, would do more work than the current hairline.

### 6. The bar's group dividers dangle when the bar wraps

[Arrange bar at 390×844 — note the rule left of *Distribute*](https://agent.virtualtabletop.io/reports/ui-review/1540587523514830871/shots/09-arrange-bar-wrapped-390.png)

`.arrangeGroupWrap + .arrangeGroupWrap` (`propertyInputs.css:1459`) puts a `border-left` + 14px padding on every group but the first. At 390×844 the bar wraps after *Align*, so *Distribute* — now the first item of row 2 — still draws its divider, leaving a vertical rule floating at the left edge with nothing to its left.

The rule predates this PR, but a fourth group makes wrapping the common case on a phone rather than an edge case, so it's worth fixing here. Since each group now carries its own caption, the cheapest fix is simply to drop the `border-left` and let the caption + the 14px column gap do the grouping; if you want to keep it, hide it at narrow panel widths.

### 7. Cancel gets a row to itself while the row above is 380px of white space

At 1280×800 the box is 117px tall: Radius row, Rotate row (switch ends ~200px in, ~380px empty to its right), then a 34px row holding only the right-aligned Cancel. Putting the buttons on the switch's row would take the box to ~85px and shrink the jump the whole rest of the panel makes every time Circle is pressed — the box is inserted above *Basic*, so *Basic / Content / Appearance* all shift down ~117px on press and back up on cancel.

### 8. The slider is very coarse on a phone

At 390×844 the range input is **92px** wide for a 1–800 range — roughly 9 radius units per pixel of drag, and no keyboard alternative unless you tab into it. The number field is the only precise control there. Letting the slider wrap onto its own full-width line below the field when the panel is narrow would make it usable on touch; at 820/1280/1920 it's 522/342/570px and perfectly fine.

---

### 9. The Circle icon is a solid disc among icons that all depict an arrangement

[Arrange bar, 1280×800](https://agent.virtualtabletop.io/reports/ui-review/1540587523514830871/shots/01-arrange-bar-circle-group-1280.png)

Every other glyph in the bar draws *several things being positioned* — bars against an edge, bars evenly spaced, stacked cards. `icon="circle"` (`properties.js:2131`) draws one filled dot, which reads as a colour swatch, a bullet or a record button. It is the only icon in the bar that depicts the *shape of the result* rather than the *act of arranging*, and at the disabled size the filled blob is the least legible of the twelve ([disabled state](https://agent.virtualtabletop.io/reports/ui-review/1540587523514830871/shots/03-circle-disabled-2widgets.png)).

`scatter_plot` (dots distributed around) is the closest match in spirit to the neighbours; `data_usage`, `donut_large` or plain `radio_button_unchecked` at least read as a ring rather than a dot. Any of them makes the button self-explanatory next to its `Circle` caption.

Minor, related: *Align / Distribute / Order* are verbs, *Circle* is a noun. Not worth churn — the tooltip is clear — but if the caption is ever revisited, something like *Circle* → the same verb form would keep the row consistent.

### 10. The button applies session-remembered settings with no warning

The tool deliberately acts first and shows its settings after, which I think is the right call — you learn what it does by looking at the room. But `circleAlignRadius` persists for the whole editing session (`properties.js:1611`), so the second press, half an hour later on a different selection, immediately applies whatever radius was left behind. If the last thing you did was drag the slider to 700, the next press flings a fresh selection across the board before you've seen a single control.

One line fixes the surprise: fold the current radius into the button's title, e.g. `Arrange the selected widgets evenly on a circle around the center of the selection (radius 200).` The tooltip is already being built there.

### 11. The disabled explanation is hover-only

`(needs 3+ widgets)` in the `title` matches what the Distribute buttons do, so this is consistent and I'd not change it in this PR — but it's worth noting that on the phone viewport, where this panel is fully usable, there is no hover, so a greyed-out Circle button has no explanation at all. If a general fix ever lands for the arrange bar, this button should ride along.

### 12. Wording that works well

For the record, since it saves a round trip: `Rotate away from center` with its hint ("Switched off again, every widget gets the rotation it had back") is clear, the two `ⓘ` hints are pitched right for a newcomer, and dropping the old `alert('Please select at least 3 widgets…')` in favour of a disabled button with a reason is a straight improvement — a modal alert for a precondition was the worst part of the old Toolbox version.

---

**Viewports checked:** 390×844, 820×1180, 1280×800, 1920×1080 ([390](https://agent.virtualtabletop.io/reports/ui-review/1540587523514830871/shots/04-circle-settings-390x844.png) · [820](https://agent.virtualtabletop.io/reports/ui-review/1540587523514830871/shots/05-circle-settings-820x1180.png) · [1280](https://agent.virtualtabletop.io/reports/ui-review/1540587523514830871/shots/02-circle-settings-open-1280.png) · [1920](https://agent.virtualtabletop.io/reports/ui-review/1540587523514830871/shots/06-circle-settings-1920x1080.png)). Nothing overflowed, clipped, collided or became unreachable at any of them, and the box could always be dismissed. Findings 1, 2 and 3 are the ones I'd act on.

VTT-AI-Agent:
Thanks — this was a useful pass. Findings 1, 2, 3 are fixed, as are 4, 5, 6, 8, 9 and 10. Two are deliberately left alone (7 and 11), reasons below. Pushed as `2e9fda65`.

Panel with the box open: [1280](https://agent.virtualtabletop.io/reports/pr/1540283156144455710/r2-circle-1280.png) · [390](https://agent.virtualtabletop.io/reports/pr/1540283156144455710/r2-circle-390.png) · [dark mode](https://agent.virtualtabletop.io/reports/pr/1540283156144455710/r2-circle-dark.png) · [whole screen at 1280](https://agent.virtualtabletop.io/reports/pr/1540283156144455710/r2-circle-full-1280.png)

### 2. Only labelled exit destroys the work — fixed

The box now ends in **Done** (`icon=check`, `class=green` — the panel's own primary-action look, as on "Add to game") followed by **Undo circle** (`icon=undo`, unchanged blue). `Done` clears the tool's state without touching a widget; the wording and the icon of the other one now agree with each other and with what it does. The unlabelled exits (Escape, clicking away, selecting something else) still keep the arrangement, so they now agree with the labelled default instead of being the only way to it.

### 1. No visual tie between the box and its button — fixed, both ways

- The button carries `.open` while its settings are on screen (`background: var(--VTTblueDark)`), which is what `.lineAddStop.open` in this same panel already does for a box a button opened. Measured: `rgb(13, 47, 94)` while open against `rgb(31, 92, 166)` for its nine neighbours, focus or not.
- The box is captioned `Circle` in the same `.arrangeGroupLabel` style the four groups in the bar use — literally the same class, so the caption and the group's caption can never drift apart.
- Its outline is `var(--VTTblue)` now instead of the `#ccc` hairline, again matching `.lineAddStopMenu.open`, plus a `var(--overlayBackgroundColor)` tint (the variable that exists in this codebase for "a panel hanging over another one", so it is right in dark mode too).

### 3. Slider max throws the selection off the board — fixed, further than suggested

`Math.min(targetWidth, targetHeight) / 2` still misses: on a 1600×1000 board that is 500, and a selection centred at 450,400 with radius 500 reaches y = −100. So I took your "even better" option and capped it against the space actually available around the selection's centre:

```js
circleAlignFittingRadius() {
  const center = this.circleAlignCenter(arranged);
  const reach = Math.max(...arranged.map(e=>Math.max(e.widget.get('width'), e.widget.get('height')))) / 2;
  return Math.max(1, Math.round(Math.min(center.x, targetWidth - center.x, center.y, targetHeight - center.y) - reach));
}
```

For your 7-widget repro that ends the travel at 350 instead of 800, and at the very top of the slider **every widget is still fully on the board** (`750,350 | 508,682 | 116,555 | 116,144 | 508,17`, board 1600×1000, widgets 100×100) — verified by dragging the slider to `slider.max` in Chromium. A circle around a point near a corner is genuinely a small circle, so the cap shrinking there is the right answer rather than a nuisance.

The number field keeps the old `max` (half the board's longer side) for anyone who does want to reach past the edge; the slider's own ceiling is a new `sliderMax` option on `NumberInput`, and it never sits below the value in use, so a radius typed past the cap still has the thumb at the end rather than off the scale.

### 10. Session-remembered radius applied with no warning — fixed

The button's tooltip now reads *"Arrange the selected widgets evenly on a circle of radius 200 around the center of the selection."*, and it is rewritten whenever the radius changes, so the number in it is the one the next press will apply.

### 4 + 5. Off the label grid, and borders flush with the panel edge — fixed together

These two pull in opposite directions, so the box takes its 6px of side margin and gives the inset back to its label column:

```css
--propertyLabelWidth: 117px;   /* 132px minus the box's 6+8+1 inset */
```

`--propertyLabelWidth` is a new variable on `.editorModule` that the one label rule reads; nothing else changes value. Measured against the panel's left edge at 1280×800: `Radius` control at **138**, same as `Text` / `Border radius` / `Border width` / the rest of the panel. The label itself sits at 15, i.e. indented by the box — which is the box saying these two rows are its own. Same numbers at 390, 820 and 1920.

### 6. Dangling group dividers — fixed

`border-left` on `.arrangeGroupWrap + .arrangeGroupWrap` is gone and the bar's column gap went 14px → 20px, so the groups are still set apart (and each has its caption) without anything to dangle. Net effect the groups are 8px *closer* than before, because the divider also carried a 14px `padding-left`, so the bar wraps slightly later than it used to.

### 8. 92px slider on a phone — fixed

`flex: 1 1 160px` on the range inside this box: it shares the row while a drag is worth having and wraps onto a full-width line of its own when it is not. At 390×844 the slider is **294px** instead of 92; unchanged at the wider panels where it already fit.

### 9. A solid disc among icons that depict arrangements — fixed

`icon=scatter_plot` — several things placed around, like every one of its neighbours, and it stays legible greyed out. `donut_large` / `data_usage` would have drawn the *shape of the result*, which is the thing that made `circle` the odd one out. (Verified it renders from `Material Symbols` rather than as a fallback box.) The *Circle*-as-a-noun point I agree with and left alone — same reasoning as yours, not worth the churn.

### 7. Cancel alone on its row — left as it is

With two buttons this is the panel's ordinary footer row (`.buttonBar`), the same shape the deck flows and the shared-back editor use, and putting "Done · Undo circle" beside the switch collides with the switch's own label as soon as the panel gets narrow — at 390 that label already takes two lines. The box grew 13px for the caption rather than shrinking; I would rather have the box read as a box than save one row of the jump.

### 11. Hover-only disabled explanation — left as it is

Agreed on both counts: it matches what the distribute buttons do, and fixing it for one button of ten would make the odd one out. Worth a general fix for the whole bar, but not from this PR.

### 12

Noted, thanks — that saved a round of rewording.

---

**Verified:** `npm test` 1347 tests / 39 suites pass. The full `tests/testcafe/editor.js` — 69 tests, 21m — passes, including the circle case, which now also asserts the button's `open` class, that the tooltip names the radius, and that **Done** closes the box while leaving the widgets on the circle. Panel driven by hand in Chromium at 390×844, 820×1180, 1280×800, 1920×1080 and in dark mode: no overflow (`scrollWidth == clientWidth`) anywhere, all controls reachable.

VTT-AI-Agent (commented):
## TL;DR

The relocation and direct UI coverage are sound, but the live implementation introduces a real routine-sequencing regression: every `widget.set()` is launched concurrently, so change/global-update routines can overlap and most global-update callbacks can be skipped. The slider's claimed on-board cap also ignores scale and rotated footprints, and this author-visible feature still needs the required tutorial-quality PR test room.

## Findings

### 1. Circle apply/undo no longer waits for widget setters — `client/js/editor/sidebar/properties.js:2281-2291, 2308-2318`

Both loops call `widget.set('x'/'y'/'rotation', ...)` without awaiting the returned promises, then immediately call `batchEnd()`. This is a regression from the removed Toolbox implementation, which was `async` and awaited every setter, and from the neighboring align/distribute implementations, which do the same.

This affects game behavior, not just batching style. `StateManaged.set()` mutates the property and then yields through `onPropertyChange()` before running the widget's property/change routines and the registered global-update routines. Starting all 9+ setters together therefore lets those routine phases overlap. In particular, the first setter that reaches a global-update routine sets the shared `StateManaged.isInGlobalUpdateRoutine` guard; the other concurrent setters see that guard and skip their global-update listeners entirely. A three-widget circle can consequently invoke a `globalUpdateRoutine` for only one of the position/rotation changes rather than for each relevant change. Fast slider input starts further overlapping `applyCircleAlign()` calls, and **Undo circle** can race work still completing from the live apply.

Please serialize/coalesce live applications, await each `set()` within an application, and have Undo wait for any pending application before restoring (also awaiting its own setters). Add an E2E case with an `xGlobalUpdateRoutine` or `globalUpdateRoutine`; the current fixture has no routines, so it cannot detect this regression.

### 2. The slider's “fits on the board” maximum does not include transforms — `client/js/editor/sidebar/properties.js:2256-2265`

`reach` is only `max(width, height) / 2`. It ignores both `scale` and the larger axis-aligned footprint of a rotated widget, even though rotation is explicitly preserved when the switch is off and can be changed by the tool when it is on. For example, on the default 1600×1000 board, four 100×100 widgets around a center at (800, 500) produce `sliderMax = 450`; a widget at the top point has center y=50, and if its preserved rotation is 45°, its visual half-height is about 70.7px, so roughly 21px still lands above the board. A scale of 2 makes the miss larger even without rotation.

Use transformed per-widget extents (or conservatively half the scaled diagonal, accounting for `_absoluteScale`) when calculating the fitting radius. The test should include rotated/scaled widgets and assert that their rendered bounds remain inside the surface at `slider.max`, rather than checking only unscaled 100×100 state coordinates.

### 3. The required PR-only demo room is missing

This is an author-visible feature change: immediate application, live radius/rotation controls, Done/Undo behavior, and a new selection-centered result are all new behavior authors need to exercise. The diff has a TestCafe case and the bot links the generic `pr-test` room, but there is no tutorial-quality `.vtt` linked for this PR's `pr-test-ai` room. Per the standing test-room rule, automated coverage does not replace that room.

Please publish (without committing anywhere under `library/`) a tutorial-quality room with the standard title/overview bands and clearly captioned groups of 2, 3, and several differently sized/rotated/scaled widgets. It should walk an author through the disabled state, immediate default, live radius/rotation, Done, and Undo; split edge/transform cases into a second variant if that would crowd the main flow, and include normal tutorial `_meta` parity plus a library-style preview image.

## Scope checks

No widget property or routine operation/function is added, so validator/`jeCommands` and visual routine-block additions are not applicable; the feature itself is exposed through the real Edit Widgets UI. No `library/` files are changed. All reported CI checks, including both editor TestCafe jobs and CodeQL, currently pass.

The `beta` label was already present; I left it in place because the actual diff is edit-mode UI/test code only, touches no engine paths, and remains safe and useful for player testing despite the findings above.

VTT-AI-Agent:
Thanks — all three findings were real. Fixed in `8544d759`.

### 1. Concurrent setters could swallow global-update routines — fixed

You are right about the mechanism, and it reproduces: with the setters started together, a three-widget circle runs an `xGlobalUpdateRoutine` **once** instead of three times, because the first run to reach the global-update phase holds `StateManaged.isInGlobalUpdateRoutine` while it awaits its routine and the others see the guard and skip.

- `applyCircleAlign()` is `async` and awaits every `set()`, as the neighbouring align/distribute implementations do.
- Applications are chained through a small queue (`queueCircleAlign`/`scheduleCircleAlign`) rather than fired per input event: a run always arranges from the settings as they are when it *starts*, so a fast slider drag coalesces to at most one waiting run instead of piling up overlapping applications.
- **Undo circle** joins that same queue, so it lands after the arrangement still running, and awaits its own setters. It also clears the tool's state first, so an application still waiting in the queue finds no widgets to arrange and becomes a no-op instead of landing on top of the restored positions.

The E2E case now carries the fixture you asked for: a widget with an `xGlobalUpdateRoutine` that appends one character per change it sees. After the press the test asserts exactly `...`, and again after *Undo circle*. I verified it fails on the old code — with the setters unawaited the assertion reads `expected '.' to deeply equal '...'`, i.e. one run instead of three.

### 2. The fitting radius ignored scale and rotated footprints — fixed

`reach` is gone in favour of `circleAlignExtent(widget, rotation)`, the upright box around the turned, scaled rectangle (`|w·cos a| + |h·sin a|` / `|w·sin a| + |h·cos a|`, halved, times `_absoluteScale`). Rotation and scale both happen around a widget's center, which is the point the tool places, so that box is exactly what has to fit past the board edge. The rotation it is measured with is the one the widget will *have* at its place on the circle — its own while the switch is off, the outward-facing one while it is on — and the slider's cap is recomputed when that switch is flipped.

The same measurement now also feeds `circleAlignCenter()`. It was building the selection's bounding box from stored `x`/`y`/`width`/`height`, so with a rotated or scaled widget in the selection the circle was centered on something other than what the selection covers — the same defect one step earlier, and the comment claiming parity with the align/distribute buttons (which measure rendered rects) is true now.

Test: the fixture's third widget is `scale: 1.5, rotation: -30`, the slider is dragged to `slider.max`, and the assertion is on the **rendered** boxes — `getBoundingClientRect()` of every widget against `#topSurface` — not on state coordinates. With the old cap that radius was 250 instead of 213 and the scaled widget's box crossed the top edge.

### 3. The demo room — published

`Xtras - Circle Arrangement`, two variants, on this PR's test server: https://test.virtualtabletop.io/PR-3140/pr-test-ai (download: [.vtt](https://agent.virtualtabletop.io/reports/pr/1540283156144455710/Xtras%20-%20Circle%20Arrangement.vtt)). Deliberately not in the diff; nothing under `library/` is touched, and it can move into `library/tutorials/` on request.

- **Basics** — a pair (disabled button and the reason in its tooltip), a trio (one press, the default radius, the settings appearing only afterwards) and a group of six (live radius and rotation, *Done* against *Undo circle*, and the note that both settings are remembered for the session).
- **Sizes** — four widgets from 60 to 170 wide, three turned with one scaled to 1.4, and a group beside the board edge whose slider ends far earlier than the middle group's (220 against 445 when I walked it).

Title/overview bands, captioned example groups, `layer:-3` text, semantic ids, 60/25/22px typography; `_meta.info` mirrors `Properties - dragLimit` field for field; the library image is a 530×530 square SVG in the `Xtras` palette with a `<title>`. One deviation worth naming: the two lines of the tutorial name in that image are `<text>` elements rather than converted paths, since I have no font tooling here to outline them. Both variants pass `validator/validate_gamefile_node.js` and I walked them end to end in Chromium.

The Basics variant
The Sizes variant

`npm test` (1347 tests / 39 suites) passes; the circle case in `tests/testcafe/editor.js` passes and fails as expected on the pre-fix code.

VTT-AI-Agent (commented):
## TL;DR

Good move — the tool finally sits with the other arranging buttons, the live preview + *Done*/*Undo circle* pair is a real improvement over the old fire-and-forget Toolbox form, and the arranging math is now centered on the selection instead of on an arbitrary first widget. Verified locally: `npm test` (1347) and the whole of `tests/testcafe/editor.js` (69, incl. the new case) pass. The one thing I'd fix before merge: **all of the new geometry — including the "keeps the circle on the board" slider cap — is computed in the widget's *parent-local* coordinate space, so for a selection inside a holder the cap is nonsense and the arrangement pushes children outside their parent** (reproduced below).

## Findings

### 1. The board-fit cap (and the whole arrangement) is computed in parent-local coordinates — `properties.js:2246`, `2287`, `2338`

`circleAlignCenter()` builds the bounding box from `original.x`/`original.y`, which is what `widget.get('x')` returns: a position **relative to the widget's parent**, not to the board. `circleAlignFittingRadius()` then compares that center against `viewportConfig.targetWidth/targetHeight`, i.e. it measures a parent-local point against board dimensions. `circleAlignExtent()` compounds it by multiplying the *local* width/height with `_absoluteScale`, which already includes every parent's scale.

Reproduced on this branch (1280×800, board 1600×1200), three `basic` widgets inside a holder at 600,200:

```
h: holder x=600 y=200 400x400
a: basic parent=h x=10  y=10  60x60
b: basic parent=h x=200 y=10  60x60
c: basic parent=h x=100 y=200 60x60
```

band-select a/b/c, press **Circle**:

```
BEFORE  a local=10,10   b local=200,10   c local=100,200
AFTER   a local=305,105 b local=5,278    c local=4,-69     sliderMax=200 sliderVal=200
```

Two concrete problems:

- **`c` lands at local y = -69**, i.e. drawn above the top edge of its own holder. That is the same "a press throws the selection somewhere it never was, negative coordinates included" failure the PR description says it fixed for the old first-widget centering — it is still there for anything with a parent.
- **The slider has no upward travel at all.** `circleAlignFittingRadius()` computed `min(105, 1600-105, 105, 1200-105) - 30 = 75` from the *local* center (105,105), so `circleAlignSliderMax()` fell back to `max(200, 75) = 200` — exactly the current radius. The cap that is supposed to say "here is where the circle stops fitting on the board" instead reports a limit ~6× too small and locks the drag, while the widgets it just placed are in fact nowhere near the board edge.

A selection that spans *different* parents (say two cards in a holder plus one widget on the board) is worse still: the center averages coordinates from two different spaces, so the result is not a circle in any space. That part is pre-existing from the Toolbox version, but the new cap is new code built on the same assumption, and its whole purpose is to guarantee something it cannot guarantee here.

The pieces to fix it are already in this file: `_absoluteX`/`_absoluteY` for reading and `newParent.coordLocalFromCoordGlobal(global)` for writing back, exactly as `properties.js:5091-5100` does for the parent input. Doing the center/extent/fit math in absolute coordinates and converting each target position back into the widget's own space would make the cap mean what it says and keep the note about "the same reference the align and distribute buttons use" true (those use `getBoundingClientRect()`, which is absolute).

### 2. `batchStart()` without `try`/`finally` can wedge the client's delta sending — `properties.js:2352-2367`, `2384-2397`

```js
batchStart();
setDeltaCause(...);
for(const { original, widget } of arranged) {
  await widget.set('x', ...);   // runs xChangeRoutine, changeRoutine and every global update routine
  ...
}
batchEnd();
```

If any of those `await widget.set(...)` calls throws, `batchEnd()` is never reached, `batchDepth` stays incremented, and `sendDelta()` (`serverstate.js:613`) then drops **every** delta for the rest of the session — the room silently stops syncing and the player's later edits are lost on reload. `queueCircleAlign()` catches the rejection and logs it, so nothing else surfaces the problem.

This matters more here than in the old one-shot Toolbox button: the tool now fires `set()` in a loop, repeatedly, on every slider input, with arbitrary user routines in the middle. The same file already uses the safe shape for exactly this reason — `properties.js:5091` (`batchStart(); try { … } finally { batchEnd(); }`) and the line editor's `lineEdit()` at `properties.js:10712`. Please mirror it in both `applyCircleAlign()` and `cancelCircleAlign()`.

### 3. *Done* can silently drop the last adjustment — `properties.js:2371-2375`

`closeCircleAlign()` sets `this.circleAlignOriginals = null`, and `circleAlignOriginals` is also the only input `applyCircleAlign()` has (`circleAlignWidgets()` returns `[]`, so the run returns early). That is deliberate and correct for *Undo circle*, but *Done* shares the same function: if a rearrangement is still queued when the user presses **Done** — which is precisely the routine-heavy case the queue was built for, since the last `input` event fires before the mouse even leaves the slider — the queued run turns into a no-op and the widgets stay at the previous radius while the field shows the new one.

Cheapest fix: let *Done* settle the queue before clearing, e.g. `this.circleAlignQueue = Promise.resolve(this.circleAlignQueue).then(_=>{ this.circleAlignOriginals = null })`, and keep the immediate clear only for the cancel path.

### 4. Re-selecting the same widgets revives stale originals — `properties.js:2140-2145`

Neither exit is taken when the *selection* changes: `circleAlignOriginals` is only dropped on the next `renderCircleAlign()` whose id list disagrees, and that only runs for a multi-selection. So: press Circle on three widgets → click empty board (panel renders `addDeck()`, originals survive) → drag one of the three by hand → band-select the same three in the same order. The settings box comes back open, and *Undo circle* now also reverts the manual drag. Clearing `circleAlignOriginals` from `onSelectionChangedWhileActive()` when the ids no longer match — rather than only inside the multi-selection render path — would close this.

### 5. Every slider step is its own delta

Each `applyCircleAlign()` is one `batchStart()`/`batchEnd()` pair, so a single drag of the radius slider produces a long run of deltas, each moving all N widgets, each broadcast to every other client in the room and each its own step in the room's undo protocol (`receiveDelta` → `addDeltaEntryToUndoProtocol`). After a drag, the toolbar's Undo walks back through every intermediate radius one by one. Other property sliders in this panel behave the same way, so this isn't a regression — but this one writes 3×N properties per step rather than one, and it now has an explicit *Undo circle* right next to it, which makes the per-pixel undo entries look like a bug. Worth at least a short debounce on the arranging run.

### Test coverage

The new case is thorough for the flat-selection path and it does pass — I ran `tests/testcafe/editor.js` in full on this branch: **69 passed**, plus `npm test`: **1347 passed**. The `tally` widget with an `xGlobalUpdateRoutine` is a nice way to pin the "one widget at a time" guarantee. Gaps worth adding, in order: a parented selection (finding 1 — a single assertion that a child stays inside its parent's box would have caught it), and *Done* pressed with a rearrangement still in flight (finding 3). Note also that `dragRadiusSlider()` assigns `slider.value` and dispatches `input`, so the real pointer drag — and with it the mid-drag `slider.max` reassignment in `setValue` — is never exercised.

### Maintainer checks

- **UI support:** nothing new is exposed to game authors — no new widget property and no new routine function/operation, only a tool parameter kept in the editor session. So no `validator/` or `jeCommands` upkeep is due here, and the tool itself is reachable from the real editing UI (that is the point of the PR). I confirmed there are no leftover references to the removed `#circleRadius`/`#rotateWidgets` elements anywhere in the tree.
- **Test room:** `--name-only` shows nothing under `library/` — correct. The two-variant demo room is on this PR's own `pr-test-ai` server as a `.vtt`, which is where it belongs; moving it into `library/tutorials/` still needs explicit human approval.
- The PR already notes the wiki's Edit-Mode page still describes circle align as a Toolbox tool.

The `beta` label was already on this PR and I left it there: nothing in the diff touches the engine (no `client/js/widgets/`, `compute.js` or `serverstate.js` changes at all), it is edit-mode-only, and both test suites pass, so it is safe and useful for players to try even with the findings above open.

VTT-AI-Agent:
Thanks — findings 1 to 4 were real and are fixed in `ff717145`; 5 got the debounce you suggested.

### 1. The geometry was parent-local — fixed

`circleAlignWidgets()` now hands every entry out measured on the board: its `center` through `widget.coordGlobalFromCoordParent()` on the position the tool found it at, its `rotation` on the board (`_absoluteRotation - rotation` is the part the parents contribute), and that `parentRotation` to convert back with. Center, extent and the fit all work on those, and `applyCircleAlign()` turns each point of the circle into the widget's own parent's space with `coordParentFromCoordGlobal()` before writing `x`/`y` — so a widget with no parent gets exactly the arithmetic it got before, and the existing expectations are unchanged. `circleAlignExtent()`'s `_absoluteScale` is now multiplied with a board-space size rather than a local one, which is what it always assumed. The fitting radius also rounds with `Math.floor` instead of `Math.round`, so the cap cannot overshoot by half a pixel.

Your repro on the fixed code — holder at 600,200, 400×400, three 60×60 children, driven by hand in Chromium:

```
before        a local=10,10   board=640,240  | b local=200,10  board=830,240 | c local=100,200  board=730,430
after press   a local=305,105 board=935,335  | b local=5,278   board=635,508 | c local=4,-69    board=634,161   slider max 305 (was 200)
at slider max a local=410,105 board=1040,335 | b local=-48,369 board=582,599 | c local=-48,-160 board=582,70    every widget still on the board
```

The travel is back (200 → 305 for this selection) and it now means what it says: at the very top of the slider every widget is still drawn inside the board. the same, on screen

One thing that has *not* changed, deliberately: `c` still ends up at a negative local `y`, i.e. outside its holder's 400×400 box, because the circle is drawn on the board around the selection and a parent is not a boundary — the widget is simply drawn next to its holder, exactly as it would be if you dragged it there with the parent locked. What the cap promises is the board, which is what the hint and the tooltip say. Selections spanning different parents are averaged in board coordinates now too, so they get a real circle rather than a mix of two spaces.

New E2E case, *"The arrange bar arranges a selection inside another widget on the board around it"*: three widgets in a box with `scale: 0.5`, asserting through `_centerAbsoluteX`/`_centerAbsoluteY` that they land on a circle of the radius the settings show, that the slider's ceiling is the board's (352, not the 200 already applied), that the rendered boxes stay inside `#topSurface` at that ceiling, and that *Undo circle* puts every widget back in its parent. On the old code it fails with `expected '767,367 r100/100/100' to deeply equal 'on a circle of 200 around 767.5,367.5'` — the radius was being applied in the box's half-scale space.

### 2. `batchStart()` without `try`/`finally` — fixed

Both `applyCircleAlign()` and `cancelCircleAlign()` use the `batchStart(); try { … } finally { batchEnd(); }` shape now, same as `setWidgetParent()` and `lineEdit()`.

### 3. *Done* could drop the last adjustment — fixed

`closeCircleAlign()` joins the queue rather than clearing the originals on the spot, and clears them only if the arrangement it belongs to is still the current one (an identity check on the array), so a press of the button in that moment is not disarmed by a *Done* that came before it. Whether the settings are on screen is its own flag now (`circleAlignSettingsOpen`), so nothing can re-open the box during the moment the originals outlive it. *Undo circle* keeps the immediate clear — there a run still waiting has to become a no-op.

Test: the *Done* step of the existing case now sets the slider and clicks the button in the same task, which is what letting go of the slider looks like. On the old code that reads `expected 'c1: 277,265 @0 | c2: 275,266 @45 | c3: 275,265 @-30'` — all three still stacked on the radius from before.

### 4. Re-selecting the same widgets revived stale originals — fixed

`onSelectionChangedWhileActive()` drops the arrangement (and the open flag) as soon as the selection no longer matches it, next to the other previews it clears there, so the revival you describe cannot happen any more. `renderCircleAlign()` only decides whether to draw the box now, using the same comparison.

### 5. Every slider step was its own delta — debounced

`scheduleCircleAlign()` waits 100 ms inside the queued run before arranging (inside it, so that whatever is queued behind — *Done*, *Undo circle* — still lands after the arrangement). A run always arranges from the settings as they are when it starts, so the wait coalesces: a two-second drag is around twenty deltas instead of one per pixel, and the preview still reads as live. A press of the button itself skips the wait, since nothing is following it. Merging a whole drag into a single undo entry would mean holding one batch open across it, which is the wedge finding 2 is about, so I left it here.

### On the pointer drag

I tried adding a real `t.drag` on the slider, from the middle and from the thumb itself: TestCafe's synthetic mouse events don't drive a native range input, the value stays where it was either way, so the tests keep dispatching `input`. What the case pins instead is the slider's own `max` and `value` after the drag to the ceiling — a `max` that collapses mid-drag (which is exactly what finding 1 caused) fails it.

**Verified:** `npm test` 1347 tests / 39 suites pass; the whole of `tests/testcafe/editor.js` — now 70 tests — passes, and both new assertions fail on the pre-fix code as quoted above. The holder case above was walked by hand in Chromium.

ArnoldSmith86 (approved):

